Issue #2197, opened 2019-12-01
7.6 support


pebrc:
Test and support Elasticsearch version 7.6. I did a first test with a recent snapshot build and 
#2177 came out of that. But there are other changes we know of already that we have to support: 
- [x] https://github.com/elastic/elasticsearch/pull/49223 support enterprise license type. This will make it necessary in our license selection logic to prefer stack level enterprise licenses over platinum licenses when both are contained in a orchestration level enterprise license. 
- [ ] https://github.com/elastic/elasticsearch/pull/49418 support `xpack.license.upload.types` this should be set to `trial, enterprise` for ECK managed clusters as of version 7.6

tvernum:
> `xpack.license.upload.types` this should be set to `basic, trial, enterprise`

It actually needs to be `trial, enterprise`.
You are never allowed to _upload_ `basic` licenses, so the latest version of that PR prevents you from including it in the allowed types.

PR #2242, opened 2019-12-11
Support 7.6 and new stack level enterprise license type.

pebrc:
* support 7.6 and stack level enterprise licenses by preferring enterprise stack license over platinum for 7.6 and upwards
* moved some code around because of import cycles (I think we may need a larger rethink of the dependencies with in the project)
* does not yet include for the new settings (will raise separate PR) 
* does not yet include 8.x support
* testing is possible only with current master snapshots of Elasticsearch

Relates to #2197


pkg/controller/elasticsearch/client/model.go
```diff
@@ -264,6 +286,14 @@ func (l License) IsValid(instant time.Time) bool {
 		l.ExpiryTime().After(instant)
 }
 
+// IsSupported return true if the current license type is supported by the given version of Elasticsearch.
+func (l License) IsSupported(v *version.Version) bool {
+	if l.Type == string(ElasticsearchLicenseTypeEnterprise) && !v.IsSameOrAfter(version.MustParse("7.6.0")) {
+		return false
+	}
+	return true
```
pebrc: I know this statement can be collapsed into the just returning directly from the comparison, but I found the intent much easier to understand by being a bit redundant here.

pkg/controller/elasticsearch/label/label_test.go
```diff
@@ -87,3 +87,84 @@ func TestExtractVersion(t *testing.T) {
 		})
 	}
 }
+
+func TestCurrentVersions(t *testing.T) {
```
barkbay: Not really from your PR but should not it be called `TestMinVersion`?

pkg/controller/common/license/match.go
```diff
@@ -21,14 +22,16 @@ type licenseWithTimeLeft struct {
 // desired license type and the remaining validity period of the license.
 // Returns the license, parent license UID, a bool indicating a match was found and an optional error.
 func BestMatch(
+	minVersion *version.Version,
```
barkbay: Should we update the godoc to mention why there is an additional parameter ? (not 💯 % it's really worth it but it can help when we read the code)

Issue #2259, opened 2019-12-12
Readyness probe is timing out


l0x-c0d3z:
## Bug Report

**What did you do?**

I'm running a three node cluster. I have write-heavy loads. I run sometimes queries on a pretty large dataset (~600GB).

**What did you expect to see?**

The nodes are all up and running, and kibana should work.

**What did you see instead? Under which circumstances?**

Soon after the initial query, kibana stops working and every request starts to fail. I get the following in kibana log:

```
{"type":"log","@timestamp":"2019-12-12T17:56:27Z","tags":["security","error"],"pid":6,"message":"Error registering Kibana Privileges with Elasticsearch for kibana-.kibana: No Living connections"}
```

When I check the pods, they are marked as not ready.
```
pod/noscit-production-logs-es-basic-0         0/1     Running   0          28h
pod/noscit-production-logs-es-basic-1         0/1     Running   0          28h
pod/noscit-production-logs-es-basic-2         0/1     Running   0          31h
```

I have stuff like this in my pods events:
```
  Warning  Unhealthy  19s (x15444 over 28h)  kubelet, node-002  Readiness probe failed:
```

However, all my curl requests to the node seem to work fine.

Because no nodes is considered ready, the service reports no endpoints:

```
kubectl describe svc noscit-production-logs-es-http
Name:              noscit-production-logs-es-http
Namespace:         noscit-production
Labels:            common.k8s.elastic.co/type=elasticsearch
                   elasticsearch.k8s.elastic.co/cluster-name=noscit-production-logs
Annotations:       <none>
Selector:          common.k8s.elastic.co/type=elasticsearch,elasticsearch.k8s.elastic.co/cluster-name=noscit-production-logs
Type:              ClusterIP
IP:                x.x.x.x
Port:              https  9200/TCP
TargetPort:        9200/TCP
Endpoints:         
Session Affinity:  None
Events:            <none>
```

I tracked it down to the probe script (`/mnt/elastic-internal/scripts/readiness-probe-script.sh` ). It performs a curl request with a 3 seconds timeout. And it seems pretty obvious I'm hitting the timeout.

```
[root@noscit-production-logs-es-basic-1 elasticsearch]# time /mnt/elastic-internal/scripts/readiness-probe-script.sh ; echo $?

real	0m3.006s
user	0m0.024s
sys	0m0.000s
1
```

I quickly confirmed that this was indeed the case:

```
[...]

PR #2260, opened 2019-12-12
Allow user to override readiness timeout

anyasabo:
Relates to https://github.com/elastic/cloud-on-k8s/issues/2259

Default the readiness timeout 3s, but allow the user to set an env var to override the timeout for clusters that are under load. Currently it is hardcoded so changing the timeout means you need to change the whole readiness probe

### DIFF
--- a/pkg/controller/elasticsearch/nodespec/readiness_probe.go
+++ b/pkg/controller/elasticsearch/nodespec/readiness_probe.go
@@ -30,7 +30,7 @@ const ReadinessProbeScriptConfigKey = "readiness-probe-script.sh"
 const ReadinessProbeScript string = `
 #!/usr/bin/env bash
 # Consider a node to be healthy if it responds to a simple GET on "/_cat/nodes?local"
-${CURL_TIMEOUT:=3}
+CURL_TIMEOUT=${CURL_TIMEOUT:=3}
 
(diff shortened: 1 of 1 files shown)


pkg/controller/elasticsearch/nodespec/readiness_probe.go
```diff
@@ -30,7 +30,7 @@ const ReadinessProbeScriptConfigKey = "readiness-probe-script.sh"
 const ReadinessProbeScript string = `
 #!/usr/bin/env bash
 # Consider a node to be healthy if it responds to a simple GET on "/_cat/nodes?local"
-CURL_TIMEOUT=3
+CURL_TIMEOUT=${CURL_TIMEOUT:=3}
```
sebgl: I feel like this env var name should be more explicit: `READINESS_CURL_TIMEOUT`

sebgl: I feel like this env var name should be more explicit: `READINESS_CURL_TIMEOUT`

sebgl: I feel like this env var name should be more explicit: `READINESS_CURL_TIMEOUT`

thbkrkr: `READINESS_PROBE_TIMEOUT`?

anyasabo: Agreed, I'll open a new PR

Issue #2207, opened 2019-12-03
Empty limits are not take into account


thbkrkr:
Defining an empty limits section in an ES manifest so that the operator does not apply the default limits does not work as explained [here](https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-managing-compute-resources.html).

#### Steps to reproduce

Deploy an Elasticsearch cluster with this manifest:

```yaml
apiVersion: elasticsearch.k8s.elastic.co/v1beta1
kind: Elasticsearch
metadata:
  name: nolimitx
spec:
  version: 7.5.0
  nodeSets:
  - name: default
    count: 1
    config:
      node.master: true
      node.data: true
      node.ingest: true
      node.store.allow_mmap: false
    podTemplate:
      spec:
        containers:
        - name: elasticsearch
          resources:
            limits: {}
```

If we look at the ES custom resource, we can see that the `limits` field has been removed:
```json
> k get elasticsearch nolimit -o json | jq '.spec.nodeSets[0].podTemplate.spec'
{
  "containers": [
    {
      "name": "elasticsearch",
      "resources": {}
    }
  ]
}
```

So the pod inherits from the default limit:
```json
> k get pod nolimit-es-default-0 -o json | jq '.spec.containers[0].resources.limits'
{
  "memory": "2Gi"
}
```

This is due to the `omitempty` tag that decorates the `Limits` field: `Limits ResourceList 'json:“limits,omitempty”'`.

barkbay:
I don't understand how this was supposed to work, and embarrassed because I thought I did some tests.

`omitempty` does not seem to be a recent change, the only idea I have in mind so far is to delete the relevant part in the documentation and discuss an other option to let the user explicitly disable the default resources (an annotation ?).

thbkrkr:
I tested with the latest operator (in master) and it works!

So I double-checked that it does not work with the old operators (0.9, 1.0.0-beta1, 1.0.0) and that it is independent of the k8s version. Confirmed.

Now I have to figure out what has changed to make this work again.

thbkrkr:
01437745 is responsable of the change (if `kubectl apply`, not `kubectl create`).

This is because with this code we mutate the `dest` struct (~ the custom resource) to validate, by using it to decode the JSON coming from the annotation where the object has been stored without modification (`resources:{limits:{}}` is there).

[...]

PR #2264, opened 2019-12-13
[DOC] Empty limits are ineffective

barkbay:
This PR updates the documentation re #2207


docs/managing-compute-resources.asciidoc
```diff
@@ -106,17 +106,4 @@ With the above restriction in place, if you create an Elasticsearch object witho
...
-  nodeSets:
-  - podTemplate:
-      spec:
-        containers:
-        - name: elasticsearch
-          resources:
-            # specify empty limits
-            limits: {}
-----
+To avoid this, you must define some resources in the specification so that the operator does not apply the default limits to the object.
```
charith-elastic:
```suggestion
To avoid this, explicitly define the requests and limits mandated by your environment in the resource specification. It will prevent the operator from applying the built-in defaults.
```

PR #2233, opened 2019-12-10
Minimize downtime during pod recycling

david-kow:
This PR fixes https://github.com/elastic/cloud-on-k8s/issues/1927 and adds:
- SERVICE_NAME env variable in ES pods
- pre-stop-hook-script.sh to scripts ConfigMap
- prestop lifecycle hook in ES pods
- e2e test with watcher doing high rps against ES during pod recycle

PR #2268, opened 2019-12-13
Use PodTemplate that contains correct SecurityContext

david-kow:
Fixes the fail reason of https://devops-ci.elastic.co/job/cloud-on-k8s-e2e-tests/603/ introduced in #2233.

When using `.WithPodTemplate()` one has to provide correct `SecurityContext`. To avoid that and use the default pod template, label and env var were set using new `With.*()` funcs.

### DIFF
--- a/test/e2e/es/mutation_test.go
+++ b/test/e2e/es/mutation_test.go
@@ -236,38 +236,15 @@ func TestMutationWithLargerMaxUnavailable(t *testing.T) {
 }
 
 func TestMutationWhileLoadTesting(t *testing.T) {
-	pt := corev1.PodTemplateSpec{
-		Spec: corev1.PodSpec{
-			Containers: []corev1.Container{
-				{
-					Name: esv1.ElasticsearchContainerName,
-					Env: []corev1.EnvVar{
-						{
-							// Use much higher value to greatly reduce chance of race condition.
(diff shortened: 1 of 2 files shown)


test/e2e/es/mutation_test.go
```diff
@@ -236,38 +236,15 @@ func TestMutationWithLargerMaxUnavailable(t *testing.T) {
...
-		WithPodTemplate(pt)
+		WithEnvironmentVariable("ADDITIONAL_WAIT_SECONDS", "10")
 
 	// force a rolling upgrade through label change
-	pt.Labels = map[string]string{"some_label_name": "some_new_value"}
 	mutated := b.WithNoESTopology().
 		WithESMasterDataNodes(3, elasticsearch.DefaultResources).
-		WithPodTemplate(pt)
+		WithEnvironmentVariable("ADDITIONAL_WAIT_SECONDS", "10").
+		WithPodLabel("some_label_name", "some_new_value")
```
barkbay: nit: why this label ?

david-kow: Not sure I got it: why this label in particular? No specific reason. Why label at all? `// force a rolling upgrade through label change`.

barkbay: my bad, I missed the comment 😐

PR #2269, opened 2019-12-13
Change readiness timeout var

anyasabo:
Follow up to https://github.com/elastic/cloud-on-k8s/pull/2260

Changing the name of the env var and adding some docs

### DIFF
--- a/docs/elasticsearch-specification.asciidoc
+++ b/docs/elasticsearch-specification.asciidoc
@@ -29,6 +29,7 @@ Before you deploy and run ECK, take some time to look at the basic and advanced 
 - <<{p}-advanced-node-scheduling,Advanced Elasticsearch node scheduling>>
 - <<{p}-orchestration>>
 - <<{p}-snapshots,Create automated snapshots>>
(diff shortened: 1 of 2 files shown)


docs/elasticsearch-specification.asciidoc
```diff
@@ -506,3 +507,26 @@ spec:
...
+  nodeSets:
+  - name: default
+    count: 1
+    podTemplate:
+      spec:
+        containers:
+        - name: elasticsearch
+          env:
+          - name: READINESS_PROBE_TIMEOUT
+            value: 10
```
thbkrkr: I fear that this is not enough because the probe has also a timeout defined at the container level in the spec (set to 5sec).

https://github.com/elastic/cloud-on-k8s/blob/0728ae29afce0b79d9e04ef7d0d26a6597900d7c/pkg/controller/elasticsearch/nodespec/readiness_probe.go#L14-L20

anyasabo: Good catch, I updated the docs with an example and tested it out

PR #2272, opened 2019-12-16
Fix readiness probe

### LINKED ISSUE #2248: Elasticsearch readiness probe might fail if a single node is stuck
It looks like we run `GET _cat/nodes?local` in order to check a pod's readiness:

https://github.com/elastic/cloud-on-k8s/blob/master/pkg/controller/elasticsearch/nodespec/readiness_probe.go#L32

This API call does two things: it gets a list of nodes from the local cluster state, and then it reaches out to all the nodes in the cluster and obtains some extra info from them. If a single node in the cluster fails to respond in this second step within a few seconds then I think this means we consider all the pods to be failing their readiness checks.

I think it's a bug to accept `?local` on this API at all given this gotcha (see https://github.com/elastic/elasticsearch/issues/50088) but also think we should be using a different API for our readiness checks. For instance, `GET _cluster/health?timeout=0s` returns `200 OK` iff the node is part of a cluster with a master, and `GET /` returns (something) if the node is actually alive and responding.

(I say "something" because in certain versions `GET /` return `503 Service Unavailable` if there is no master, but more recent versions always return `200 OK` here).

barkbay:
Fix #2248 

This PR updates the readiness probe by:
* Calling the root of the web server /
* Expecting a 200 response code unless Elasticsearch version is 6.x in which case 503 errors are tolerated

### DIFF
--- a/pkg/controller/common/volume/downward_api.go
+++ b/pkg/controller/common/volume/downward_api.go
@@ -1,0 +1,46 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License;
+// you may not use this file except in compliance with the Elastic License.
+
+package volume
+
+import (
+	"github.com/elastic/cloud-on-k8s/pkg/controller/elasticsearch/volume"
+	corev1 "k8s.io/api/core/v1"
+)
+
+var downwardApiVolume = corev1.Volume{
+	Name: volume.DownwardApiVolumeName,
+	VolumeSource: corev1.VolumeSource{
+		DownwardAPI: &corev1.DownwardAPIVolumeSource{
+			Items: []corev1.DownwardAPIVolumeFile{
+				{
+					Path: volume.LabelsFile,
+					FieldRef: &corev1.ObjectFieldSelector{
+						FieldPath: "metadata.labels",
+					},
+				},
+			},
+		},
+	},
+}
+
+var downwardApiVolumeMount = corev1.VolumeMount{
+	Name:      volume.DownwardApiVolumeName,
+	MountPath: volume.DownwardApiMountPath,
+	ReadOnly:  true,
+}
+
+type DownwardApi struct{}
+
+func (DownwardApi) Name() string {
+	return volume.DownwardApiVolumeName
+}
+
+func (DownwardApi) Volume() corev1.Volume {
+	return downwardApiVolume
+}
+
+func (DownwardApi) VolumeMount() corev1.VolumeMount {
+	return downwardApiVolumeMount
+}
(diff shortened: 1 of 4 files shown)


pkg/controller/elasticsearch/nodespec/readiness_probe.go
```diff
@@ -28,7 +28,25 @@ func NewReadinessProbe() *corev1.Probe {
 
 const ReadinessProbeScriptConfigKey = "readiness-probe-script.sh"
 const ReadinessProbeScript = `#!/usr/bin/env bash
-# Consider a node to be healthy if it responds to a simple GET on "/_cat/nodes?local"
+
+# fail should be called as a last resort to help the user to understand why the probe failed
+function fail {
+  timestamp=$(date --iso-8601=seconds)
+  echo "{\"timestamp\": \"${timestamp}\", \"message\": \"readiness probe failed\", "$1"}" | tee /proc/1/fd/1
```
pebrc: That's clever :-) But should it be `/fd/2` for stderr?

pkg/controller/common/volume/downward_api.go
```diff
@@ -0,0 +1,46 @@
...
+			},
+		},
+	},
+}
+
+var downwardApiVolumeMount = corev1.VolumeMount{
+	Name:      volume.DownwardApiVolumeName,
+	MountPath: volume.DownwardApiMountPath,
+	ReadOnly:  true,
+}
```
charith-elastic: Nit: I feel like these don't need to be global variables (even though they're only package-private).  Maybe they can be directly returned from the method calls of the `DownwardApi` object?

pkg/controller/common/volume/downward_api.go
```diff
@@ -0,0 +1,46 @@
...
+
+type DownwardApi struct{}
```
charith-elastic: Nit: I think the Go convention would be to call this `DownwardAPI`

pkg/controller/common/volume/downward_api.go
barkbay: The idea here is to avoid to recreate what must be a constant everytime a Pod spec is created. I understand that it may sound like a very tiny optimization though 😐

charith-elastic: :+1:  I don't feel strongly about it. Happy to leave it as is.

pkg/controller/elasticsearch/nodespec/readiness_probe.go
```diff
@@ -28,7 +28,25 @@ func NewReadinessProbe() *corev1.Probe {
...
+}
+
+labels="` + volume.DownwardAPIMountPath + "/" + volume.LabelsFile + `"
+
+if [[ ! -f "${labels}" ]]; then
+  fail "\"reason\": \"${labels} does not exist\""
+fi
+
+# get Elasticsearch version from the downward API
+version=$(grep "elasticsearch.k8s.elastic.co/version" ${labels} | cut -d '=' -f 2)
```
thbkrkr: Nit: could be safer to use the `VersionLabelName` constant:

https://github.com/elastic/cloud-on-k8s/blob/d212d415a0651066ad2f6fd506929c0e3fa976e8/pkg/controller/elasticsearch/label/label.go#L27

PR #2277, opened 2019-12-17
Report total memory under management for licensing

thbkrkr:
Start a reporter in a goroutine that repeatedly (every 2 minutes) aggregates the total memory of
all Elastic managed components by the operator and reports it in the form of a licensing information
in a config map.

```sh
> kubectl -n elastic-system get cm elastic-licensing -o json | jq .data
{
  "eck_license_level": "enterprise",
  "enterprise_resource_units": "1",
  "timestamp": "2019-12-20T18:18:31+01:00",
  "total_managed_memory": "12.88GB"
}
```

I left `cmd/licensing-info/main.go` to show another use:

```sh
> go run cmd/licensing-info/main.go | jq . 
{
  "eck_license_level": "enterprise",
  "enterprise_resource_units": "1",
  "timestamp": "2019-12-20T18:18:31+01:00",
  "total_managed_memory": "12.88GB"
}
```


pkg/resource/license.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+	return err
+}
+
+// getOperatorLicenseLevel gets the level of the operator license.
+// If no license is found, the defaultOperatorLevel is returned.
+func (r LicensingResolver) getOperatorLicenseLevel() string {
+	checker := commonlicense.NewLicenseChecker(r.client, r.operatorNs)
+	minVersion, _ := version.Parse(minSupportedVersion)
+	lic, _, found := license.FindLicense(r.client, checker, minVersion)
```
pebrc: This is incorrect as `findLicense` returns a stack level license not an orchestration level license. 

I would suggest to refactor 
https://github.com/elastic/cloud-on-k8s/blob/571a78aaf8e2dbc49d839e12c48c15dde7ccb3f2/pkg/controller/common/license/check.go#L49

to something along those lines (Pseudo-code!)
```go
// EnterpriseFeaturesEnabled returns true if a valid enterprise license is installed.
func (lc *checker) CurrentLIcense() (*EnterpriseLicense, error) {
	licenses, err := EnterpriseLicenses(lc.k8sClient)
	if err != nil {
		return nil, errors.Wrap(err, "failed to list enterprise licenses")
	}

       sort.Slice(licenses, func(i, j int) bool {
		t1, t2 := EnterpriseLicenseTypeOrder[licenses[i].Type],
			 EnterpriseLicenseTypeOrder[licenses[j].Type
		if t1 != t2 { // sort by type
			return t1 < t2
		}
		// and by remaining time
		return licenses[i].remaining < licenses[j].remaining
	})
	for _, l := range licenses {
		valid, err := lc.Valid(l)
		if err != nil {
			return nil, err
		}
		if valid {
			return l, nil
		}
	}
	return nil, nil
}

// EnterpriseFeaturesEnabled returns true if a valid enterprise license is installed.
func (lc *checker) EnterpriseFeaturesEnabled() (bool, error) {
	license, err := lc.CurrentLicense(lc.k8sClient) // or similar name
	if err != nil {
		return false, err
	}
       return license != nil 
}

```

pkg/controller/license/license_controller.go
```diff
@@ -154,7 +154,7 @@ type ReconcileLicenses struct {
 }
 
 // findLicense tries to find the best license available.
-func findLicense(c k8s.Client, checker license.Checker, minVersion *version.Version) (esclient.License, string, bool) {
+func FindLicense(c k8s.Client, checker license.Checker, minVersion *version.Version) (esclient.License, string, bool) {
```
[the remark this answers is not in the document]
thbkrkr: @#! :pensive:

I'm going to update the comment as suggested, even if it is enough to be woken up.

pkg/resource/license.go
```diff
@@ -0,0 +1,119 @@
...
+	// licensingCfgMapName is the name of the config map used to store licensing information
+	licensingCfgMapName = "elastic-licensing"
+)
+
+// LicensingInfo represents information about the operator license including the total memory of all Elastic managed
+// components
+type LicensingInfo struct {
+	Timestamp               string `json:"timestamp"`
+	LicenseLevel            string `json:"license_level"`
+	MemoryInGiga            string `json:"memory_in_giga"`
```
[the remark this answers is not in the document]
thbkrkr: I did this thinking that it could ease potential integrations by external systems (inspired by prometheus, e.g.: `go_memstats_sys_bytes: 7.3335032e+07`), but
```json
"memory": "5.37GB"
```
also looks fine to me.

pkg/resource/aggregator.go
```diff
@@ -0,0 +1,215 @@
...
+	// capitalize the suffix unless it's a K to have a surjection of [g|G|m|M|k|K] in [G|M|k]
+	suffix := match[2]
+	switch suffix {
+	case "K", "k":
+		suffix = "k"
+	default:
+		suffix = strings.ToUpper(suffix)
+	}
+	// multiply by 2 and convert it in a quantity using the suffix
+	return resource.ParseQuantity(fmt.Sprintf("%d%s", value*2, suffix))
```
david-kow: Aren't xmx/xms using binary prefixes? I.e.: `-Xmx1g` would translate to 1GiB (1,073,741,824 bytes) and here we will treat it as 1GB (1,000,000,000 bytes)?

pkg/resource/aggregator.go
```diff
@@ -0,0 +1,215 @@
...
+		return resource.Quantity{}, err
+	}
+
+	var total resource.Quantity
+	for _, es := range esList.Items {
+		for _, nodeSet := range es.Spec.NodeSets {
+			var mem resource.Quantity
+
+			// read the container memory limits
+			for _, container := range nodeSet.PodTemplate.Spec.Containers {
```
david-kow: Would it make sense to move this loop to separate function taking ([]container, containerName, defaultLimit) and use it in all three aggregations?

pkg/resource/aggregator.go
```diff
@@ -0,0 +1,215 @@
...
+
+		total.Add(multiply(mem, as.Spec.Count))
+		log.V(1).Info("Collecting", "as_name", as.Name, "memory", mem.String(), "count", as.Spec.Count)
+	}
+
+	return total, nil
+}
+
+// maxHeapSizePattern is the pattern to extract the max Java heap size (-Xmx<size>[g|G|m|M|k|K])
+const maxHeapSizePattern = "-Xmx([0-9]*)([gGmMkK]*)"
```
david-kow:
```suggestion
const maxHeapSizePattern = "-Xmx([0-9]*)([gGmMkK]?)$"
```

Otherwise we would match and try parse things like `5MM`.

pkg/resource/aggregator.go
thbkrkr: Yes!

pkg/resource/aggregator.go
thbkrkr: Yes for the `?` but no for the `$`. 
We don't know if -Xmx is the latest option declared in the JAVA_OPTS env var.

pkg/resource/aggregator.go
```diff
@@ -0,0 +1,215 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
```
sebgl: `pkg/resource` feels a bit generic.
I think I would move that to `pkg/license`. So in `main.go` we call `license.NewResourceReporter` instead of `resource.NewLicenseReporter`. WDYT?

pkg/resource/aggregator.go
```diff
@@ -0,0 +1,215 @@
...
+
+package resource
+
+import (
+	"fmt"
+	"regexp"
+	"strconv"
+	"strings"
+
+	asv1 "github.com/elastic/cloud-on-k8s/pkg/apis/apm/v1"
```
sebgl: I think we use `apmv1` in other places.

pkg/resource/license.go
```diff
@@ -0,0 +1,119 @@
...
+// inGiga converts a resource.Quantity in gigabytes
+func inGiga(q resource.Quantity) string {
+	// divide the value (in bytes) per 1 billion (1GB)
+	return fmt.Sprintf("%0.2f", float32(q.Value())/1000000000)
+}
+
+// inEnterpriseResourceUnits converts a resource.Quantity in Elastic Enterprise resource units
+func inEnterpriseResourceUnits(q resource.Quantity) string {
+	// divide the value in bytes per 64 billion (64 GB)
+	return fmt.Sprintf("%0.2f", float32(q.Value())/64000000000)
```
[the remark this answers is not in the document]
thbkrkr: Not really. It's up to us to suggest something.

pkg/resource/aggregator.go
```diff
@@ -0,0 +1,215 @@
...
+
+	var total resource.Quantity
+	for _, es := range esList.Items {
+		for _, nodeSet := range es.Spec.NodeSets {
+			var mem resource.Quantity
+
+			// read the container memory limits
+			for _, container := range nodeSet.PodTemplate.Spec.Containers {
+				if container.Name == esv1.ElasticsearchContainerName {
+					mem = *container.Resources.Limits.Memory()
```
sebgl: can this be nil? would panic?

pkg/resource/aggregator.go
```diff
@@ -0,0 +1,215 @@
...
+	var total resource.Quantity
+	for _, kb := range kbList.Items {
+		var mem resource.Quantity
+
+		// read the container memory limits
+		for _, container := range kb.Spec.PodTemplate.Spec.Containers {
+			if container.Name == kbv1.KibanaContainerName {
+				mem = *container.Resources.Limits.Memory()
+
+				// if not, fallback to the max JVM heap size
```
sebgl: `node VM`

pkg/resource/aggregator.go
```diff
@@ -0,0 +1,215 @@
...
+		log.V(1).Info("Collecting", "as_name", as.Name, "memory", mem.String(), "count", as.Spec.Count)
+	}
+
+	return total, nil
+}
+
+// maxHeapSizePattern is the pattern to extract the max Java heap size (-Xmx<size>[g|G|m|M|k|K])
+const maxHeapSizePattern = "-Xmx([0-9]*)([gGmMkK]*)"
+
+var maxHeapSizeRe = regexp.MustCompile(maxHeapSizePattern)
```
sebgl: nit: I think it's fine to have a single variable here:
```
var maxHeapSizeRe = regexp.MustCompile("var maxHeapSizeRe = regexp.MustCompile")
```
but totally fine if you prefer having 2 👍

pkg/resource/aggregator.go
```diff
@@ -0,0 +1,215 @@
...
+		suffix = "k"
+	default:
+		suffix = strings.ToUpper(suffix)
+	}
+	// multiply by 2 and convert it in a quantity using the suffix
+	return resource.ParseQuantity(fmt.Sprintf("%d%s", value*2, suffix))
+}
+
+// nodeHeapSizePattern is the pattern to extract the max heap size of the node memory (--max-old-space-size=<mb_size>)
+const nodeHeapSizePattern = "--max-old-space-size=([0-9]*)"
```
sebgl: this is so weird 😄 :javascript:

pkg/resource/aggregator.go
thbkrkr: -Xmx1g would translate to 1GB (0.93 GiB).

`ParseQuantity` detects the format to use from the suffix:
```
//  Ki | Mi | Gi | Ti | Pi | Ei         => binarySI
//  m | "" | k | M | G | T | P | E      => decimalSI
```

pkg/resource/license.go
```diff
@@ -0,0 +1,121 @@
...
+
+	if lic == nil {
+		return defaultOperatorLicenseLevel, nil
+	}
+
+	return string(lic.License.Type), nil
+}
+
+// inGiga converts a resource.Quantity in gigabytes
+func inGiga(q resource.Quantity) string {
```
sebgl: `inGB` would be a bit more precise (vs. GiB)

pkg/resource/license.go
```diff
@@ -0,0 +1,121 @@
...
+	// divide the value (in bytes) per 1 billion (1GB)
+	return fmt.Sprintf("%0.2fGB", float32(q.Value())/1000000000)
+}
+
+// inEnterpriseResourceUnits converts a resource.Quantity in Elastic Enterprise resource units
+func inEnterpriseResourceUnits(q resource.Quantity) string {
+	// divide the value in bytes per 64 billion (64 GB)
```
sebgl: I think we say `divide by`

pkg/resource/license.go
```diff
@@ -0,0 +1,121 @@
...
+	return fmt.Sprintf("%0.2f", float32(q.Value())/64000000000)
+}
+
+// toMap transforms a LicensingInfo in a map of string, in order to fill in the data of a config map
```
sebgl: in -> to

pkg/resource/license.go
```diff
@@ -0,0 +1,121 @@
...
+	if err != nil {
+		return nil, err
+	}
+	var m map[string]string
+	err = json.Unmarshal(bytes, &m)
+	if err != nil {
+		return nil, err
+	}
+	return m, nil
+}
```
sebgl: let's unit test that just in case

pkg/resource/reporter.go
```diff
@@ -0,0 +1,73 @@
...
+import (
+	"time"
+
+	"github.com/elastic/cloud-on-k8s/pkg/utils/k8s"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+const (
+	// refreshPeriod defines how long the licensing information should be refreshed
```
sebgl: how long -> how often

pkg/resource/aggregator.go
thbkrkr: The returned memory? No.

https://github.com/kubernetes/api/blob/ca1d04f8b0d36b14ac1225fbc30504ee3c229f04/core/v1/resource.go#L37-L42

pkg/resource/aggregator.go
thbkrkr: Looks better.

pkg/controller/common/license/check.go
```diff
@@ -45,27 +47,49 @@ func (lc *checker) publicKeyFor(l EnterpriseLicense) ([]byte, error) {
...
-		return false, errors.Wrap(err, "failed to list enterprise licenses")
+		return nil, errors.Wrap(err, "failed to list enterprise licenses")
 	}
 
+	sort.Slice(licenses, func(i, j int) bool {
+		t1, t2 := EnterpriseLicenseTypeOrder[licenses[i].License.Type], EnterpriseLicenseTypeOrder[licenses[j].License.Type]
+		if t1 != t2 { // sort by type (first the most features)
+			return t1 > t2
+		}
+		// and by expiry date (first which expires in a long time)
```
anyasabo:
```suggestion
		// and by expiry date (first which expires last)
```

pkg/controller/common/license/model.go
```diff
@@ -43,6 +43,12 @@ type LicenseSpec struct {
 	ClusterLicenses    []ElasticsearchLicense `json:"cluster_licenses"`
 }
 
+// EnterpriseLicenseTypeOrder license types mapped to ints in increasing order of feature sets for sorting purposes.
+var EnterpriseLicenseTypeOrder = map[OperatorLicenseType]int{
```
anyasabo: I always struggle with the iota semantics but I wonder if that might be e better fit here

pkg/controller/common/license/check.go
```diff
@@ -45,27 +47,49 @@ func (lc *checker) publicKeyFor(l EnterpriseLicense) ([]byte, error) {
 	}, &signatureSec)
 }
 
-// EnterpriseFeaturesEnabled returns true if a valid enterprise license is installed.
-func (lc *checker) EnterpriseFeaturesEnabled() (bool, error) {
+// CurrentEnterpriseLicense returns the currently valid Enterprise license if installed.
+func (lc *checker) CurrentEnterpriseLicense() (*EnterpriseLicense, error) {
```
anyasabo: It may be worth adding some unit tests for this new logic

pkg/controller/kibana/config/settings.go
```diff
@@ -18,7 +18,11 @@ import (
 )
 
 // Kibana configuration settings file
-const SettingsFilename = "kibana.yml"
+const (
+	SettingsFilename = "kibana.yml"
+	EnvNodeOpts = "NODE_OPTS"
```
anyasabo: Can you add a comment describing what this is?

itsmed:
I cancelled this build due to the jenkins upgrade, but devops-ci is back online now.
jenkins test this please

pkg/resource/aggregator.go
david-kow: >

pkg/resource/aggregator.go
david-kow: Ah, got it. But then we should include space at the end, right? Also, should we do `[0-9]+` instead of `[0-9]*`? For cases: `-Xmx5GMoresettings` and `-XmxM`.

pkg/resource/license.go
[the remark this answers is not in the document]
david-kow: Actually, on the similar topic - should we use GB or GiB?

pkg/license/aggregator.go
```diff
@@ -0,0 +1,216 @@
...
+	}
+
+	return totalMemory, nil
+}
+
+func (a Aggregator) aggregateElasticsearchMemory() (resource.Quantity, error) {
+	var esList esv1.ElasticsearchList
+	err := a.client.List(&esList)
+	if err != nil {
+		return resource.Quantity{}, err
```
anyasabo: nit: might be worth using errors.Wrap() here (and in the other funcs) so we can have a stack trace and know where it's coming from

pkg/license/aggregator.go
```diff
@@ -0,0 +1,216 @@
...
+	}
+
+	return total, nil
+}
+
+// containerMemLimits reads the container memory limits from the resource specification with fallback
+// on the environment variable and on the default limits
+func containerMemLimits(
+	containers []corev1.Container,
+	containerName string,
```
anyasabo: Would it be simpler if we just pass it the specific container we want to examine rather than the array and the name of the item in the array?

pkg/controller/common/license/check.go
david-kow:
```suggestion
		// and by expiry date (first which expires later)
```

Or what @anyasabo said :)

pkg/license/aggregator.go
```diff
@@ -0,0 +1,216 @@
...
+
+// maxHeapSizePattern is the pattern to extract the max Java heap size (-Xmx<size>[g|G|m|M|k|K])
+var maxHeapSizeRe = regexp.MustCompile("-Xmx([0-9]*)([gGmMkK]?)")
+
+// memFromJavaOpts extracts the maximum Java heap size from a Java options string, multiplies the value by 2
+// and converts it to a resource.Quantity
+func memFromJavaOpts(javaOpts string) (resource.Quantity, error) {
+	match := maxHeapSizeRe.FindStringSubmatch(javaOpts)
+	if len(match) != 3 {
+		return resource.Quantity{}, fmt.Errorf("cannot extract max jvm heap size from %s", javaOpts)
```
anyasabo: nit: may make sense to use `errors.New()` here for the same reason as above, though this is probably the only place with that error message so we could grep for it too

pkg/license/license.go
```diff
@@ -0,0 +1,124 @@
...
+	// divide the value (in bytes) per 1 billion (1GB)
+	return fmt.Sprintf("%0.2fGB", float32(q.Value())/1000000000)
+}
+
+// inEnterpriseResourceUnits converts a resource.Quantity in Elastic Enterprise resource units
+func inEnterpriseResourceUnits(q resource.Quantity) string {
+	// divide by the value (in bytes) per 64 billion (64 GB)
+	eru := float64(q.Value()) / 64000000000
+	// round to the nearest superior integer
+	return fmt.Sprintf("%f", math.Round(eru+0.5))
```
david-kow: nit: could we use `math.Ceil` instead?

pkg/controller/common/license/model.go
anyasabo: I do see that we use a map here for ES licenses though
https://github.com/elastic/cloud-on-k8s/blob/master/pkg/controller/elasticsearch/client/model.go#L236

pkg/license/license.go
```diff
@@ -0,0 +1,124 @@
...
+	return string(lic.License.Type), nil
+}
+
+// inGB converts a resource.Quantity in gigabytes
+func inGB(q resource.Quantity) string {
+	// divide the value (in bytes) per 1 billion (1GB)
+	return fmt.Sprintf("%0.2fGB", float32(q.Value())/1000000000)
+}
+
+// inEnterpriseResourceUnits converts a resource.Quantity in Elastic Enterprise resource units
```
anyasabo:
```suggestion
// inEnterpriseResourceUnits converts a resource.Quantity to Elastic Enterprise resource units
```

pkg/license/license.go
```diff
@@ -0,0 +1,124 @@
...
+	eru := float64(q.Value()) / 64000000000
+	// round to the nearest superior integer
+	return fmt.Sprintf("%f", math.Round(eru+0.5))
+}
+
+// toMap transforms a LicensingInfo to a map of string, in order to fill in the data of a config map
+func (i LicensingInfo) toMap() (map[string]string, error) {
+	bytes, err := json.Marshal(&i)
+	if err != nil {
+		return nil, err
```
anyasabo: ditto on the errors.wrap here

pkg/license/reporter.go
```diff
@@ -0,0 +1,73 @@
...
+import (
+	"time"
+
+	"github.com/elastic/cloud-on-k8s/pkg/utils/k8s"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+const (
+	// refreshPeriod defines how often the licensing information should be refreshed
```
anyasabo: Do we maybe want to make this configurable with a default? I can see leaving that for later though.

pkg/license/license.go
```diff
@@ -0,0 +1,124 @@
...
+// Save updates or creates licensing information in a config map
+func (r LicensingResolver) Save(info LicensingInfo, operatorNs string) error {
+	data, err := info.toMap()
+	if err != nil {
+		return err
+	}
+
+	log.V(1).Info("Saving", "license_info", info)
+	cm := corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
```
anyasabo: Should we maybe have a label on this so that users with multiple operators can use a label selector across namespaces to query for them?

I also wonder if it should be owned by the operator sset so that if the operator is deleted, the config map is deleted as well.

anyasabo:
We probably want to open up a docs issue targeting 1.0 for this as well since I don't see one on the board.

cmd/manager/main.go
```diff
@@ -321,6 +321,9 @@ func execute() {
 			log.Error(err, "unable to create controller", "controller", "LicenseTrial")
 			os.Exit(1)
 		}
+
+		r := licensing.NewResourceReporter(mgr.GetClient())
+		go r.Start(operatorNamespace)
```
anyasabo: Since we're using the manager's client we should be using its cache, but the manager also waits for the cache to be synced before starting the controllers.

I need to look at it more but I wonder if we want to run this to be managed by the Manager. It looks like we just need to implement `Start(<-chan struct{}) error`. I also think we can explicitly wait for the cache sync with `mgr.GetCache().WaitForCacheSync()` if we have to.

But that might not be necessary. For operators we really don't want to be operating on an incomplete cache, but for this I think it just means that the first run might be inaccurate. After three minutes the cache should be synced and the next run will be correct.

pkg/resource/aggregator.go
thbkrkr: I misread the question. You are right. -Xmx1g must be translated into 1GiB.

pkg/license/license.go
thbkrkr: Oh yes.

pkg/license/aggregator.go
thbkrkr: Not really.

cmd/manager/main.go
sebgl: Calling `WaitForCacheSync ` sounds like a good idea to me
```
go func() {
    mgr.GetCache().WaitForCacheSync()
    r := licensing.NewResourceReporter(mgr.GetClient())
    r.Start(operatorNamespace)
}()
```

pkg/license/reporter.go
sebgl: I'm +1 on keeping that for later. Too many flags :smile:

pkg/license/license.go
```diff
@@ -0,0 +1,124 @@
...
+			Namespace: operatorNs,
+			Name:      licensingCfgMapName,
+		},
+		Data: data,
+	}
+	err = r.client.Update(&cm)
```
sebgl: I guess we could technically check if an update is required to avoid the unnecessary call to the apiserver but I don't think that would be a big issue anyway.

thbkrkr: I think this version is more optimal since we will be updating the config map most of the time.

pkg/controller/common/license/model.go
thbkrkr: Like you I'm not a big fan of the iota semantics. 

The stack license order is defined in the same way.

https://github.com/elastic/cloud-on-k8s/blob/d212d415a0651066ad2f6fd506929c0e3fa976e8/pkg/controller/elasticsearch/client/model.go#L249-L255

pkg/controller/common/license/check.go
thbkrkr: Yes!

pkg/license/reporter.go
thbkrkr: Good. I also thought that we will add it later if we really feel the need.

cmd/manager/main.go
thbkrkr: Looks cleaner to wait for the cache sync. Let's do that.

thbkrkr:
> How easy would it be to have an E2E test check this so we make sure we don't break it in the future?

I added unit tests for the 2 important methods: `Checker.CurrentEnterpriseLicense` and `ResourceReporter.Get()`.

An E2E test does not seem very complicated:
- spawn some Elastic resources
- eventually get the config map

The main drawback I see is that we have to set a timeout at least equal to the refresh period (3min) which is huge. This could be a good argument to have the refresh period configurable with a flag.

pkg/resource/aggregator.go
thbkrkr: > But then we should include space at the end, right?

No?! But we can be more precise and wait for only 1 character.

> Also, should we do `[0-9]+` instead of `[0-9]*`?

Yes!

Which gives: `"-Xmx([0-9]+)([gGmMkK])"`.

pkg/license/license.go
thbkrkr: > Should we maybe have a label on this so that users with multiple operators can use a label selector across namespaces to query for them?

Seems a good idea but which one? `common.k8s.elastic.co/type: licensing-info`

> I also wonder if it should be owned by the operator sset so that if the operator is deleted, the config map is deleted as well.

Sounds like another good idea.
We need to know the name of the elastic operator statefulset.
That doesn't seem so trivial to me.

```go
cm := corev1.ConfigMap{
		OwnerReferences: []metav1.OwnerReference{
			{
				APIVersion:         "apps/v1",
				Kind:               "StatefulSet",
				Name:               "elastic-operator",
				UID:                "",
				Controller:         &true,
				BlockOwnerDeletion: &true,
			},
		},
```

pkg/resource/aggregator.go
david-kow: > No?! But we can be more precise and wait for exactly 1 character.

What I meant is that in case of some badly formatted options (I'm not even sure that could be the case) we might match the pattern successfully and then fail on parsing. And the example could be `-Xmx5GMoresettings`, where we would match `GM` in second group. 

As to the `-Xmx([0-9]+)([gGmMkK])` I think the default unit is bytes, which won't have any letter, so we probably have to support that :)

I think the right one would be `-Xmx([0-9]+)([gGmMkK]?(?:\ .+|$))`. At the end we either want to see space followed by any other characters or just the end. Or we can accept failing on parsing in rare cases and just do simpler `-Xmx([0-9]+)([gGmMkK]?)`.

thbkrkr:
Thanks for all the feedbacks.
We are good for another pass.

pkg/license/license.go
anyasabo: >Seems a good idea but which one? common.k8s.elastic.co/type: licensing-info

That works for me. Maybe `type: elastic-usage`?

>We need to know the name of the elastic operator statefulset.

We could extract the sset name from `HOSTNAME` or if we are worried something will change that, the downward API can give us the pod name. We could leave both of these for a future PR though.

cmd/licensing-info/main.go
```diff
@@ -0,0 +1,51 @@
...
+
+	eckscheme "github.com/elastic/cloud-on-k8s/pkg/controller/common/scheme"
+	"github.com/elastic/cloud-on-k8s/pkg/license"
+	"k8s.io/client-go/kubernetes/scheme"
+	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp" // auth on gke
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
+)
+
+func main() {
```
anyasabo: I think it may make sense to add a comment to his file on why it exists and its usage -- basically what you did in the PR. Otherwise I'd be confused since nothing calls this directly and we have it running in the operator already

pkg/license/aggregator.go
```diff
@@ -0,0 +1,208 @@
...
+	envVarName string,
+	envLookup func(envVar string) (resource.Quantity, error),
+	defaultLimit resource.Quantity,
+) (resource.Quantity, error) {
+	var mem resource.Quantity
+	for _, container := range containers {
+		if container.Name == containerName {
+			mem = *container.Resources.Limits.Memory()
+
+			// if not, maybe fallback to the environment variable
```
sebgl: not sure what "if not" refers to here

pkg/license/aggregator.go
```diff
@@ -0,0 +1,208 @@
...
+						if err != nil {
+							return resource.Quantity{}, err
+						}
+					}
+				}
+			}
+		}
+	}
+
+	// if not, fallback to the default limits
```
sebgl: same: not sure what "if not" refers to here

pkg/license/aggregator.go
```diff
@@ -0,0 +1,208 @@
...
+	if mem.IsZero() {
+		mem = defaultLimit
+	}
+
+	return mem, nil
+}
+
+// maxHeapSizePattern is the pattern to extract the max Java heap size (-Xmx<size>[g|G|m|M|k|K] in binary units)
+var maxHeapSizeRe = regexp.MustCompile(`-Xmx([0-9]+)([gGmMkK]?)(?:\s.+|$)`)
```
sebgl: variable/comment mismatch

pkg/license/aggregator.go
```diff
@@ -0,0 +1,208 @@
...
+
+// memFromJavaOpts extracts the maximum Java heap size from a Java options string, multiplies the value by 2
+// and converts it to a resource.Quantity
+func memFromJavaOpts(javaOpts string) (resource.Quantity, error) {
```
sebgl: I'd add a bit more context to the "multiply by 2" thing, explaining how giving twice the JVM memory to the container is a common thing people do

pkg/license/aggregator.go
```diff
@@ -0,0 +1,208 @@
...
+	match := maxHeapSizeRe.FindStringSubmatch(javaOpts)
+	if len(match) != 3 {
+		return resource.Quantity{}, fmt.Errorf("cannot extract max jvm heap size from %s", javaOpts)
+	}
+	value, err := strconv.Atoi(match[1])
+	if err != nil {
+		return resource.Quantity{}, err
+	}
+	// capitalize the suffix unless to have a surjection of [g|G|m|M|k|K] in [Gi|Mi|Ki]
```
sebgl: ~unless~

pkg/license/aggregator.go
```diff
@@ -0,0 +1,208 @@
...
+	suffix := strings.ToUpper(match[2]) + "i"
+	// multiply by 2 and convert it in a quantity using the suffix
```
sebgl: convert it to

pkg/license/aggregator.go
```diff
@@ -0,0 +1,208 @@
...
+				esv1.ElasticsearchContainerName,
+				essettings.EnvEsJavaOpts, memFromJavaOpts,
+				multiply(nodespec.DefaultMemoryLimits, 2),
+			)
+			if err != nil {
+				return resource.Quantity{}, errors.Wrap(err, "failed to aggregate Elasticsearch memory")
+			}
+
+			total.Add(multiply(mem, nodeSet.Count))
+			log.V(1).Info("Collecting", "es_name", es.Name, "memory", mem.String(), "count", nodeSet.Count)
```
sebgl: let's also log `es.Namespace`, for consistency with other logs

pkg/license/aggregator.go
```diff
@@ -0,0 +1,208 @@
...
+		return resource.Quantity{}, err
+	}
+	// capitalize the suffix unless to have a surjection of [g|G|m|M|k|K] in [Gi|Mi|Ki]
+	suffix := strings.ToUpper(match[2]) + "i"
+	// multiply by 2 and convert it in a quantity using the suffix
+	return resource.ParseQuantity(fmt.Sprintf("%d%s", value*2, suffix))
+}
+
+// nodeHeapSizePattern is the pattern to extract the max heap size of the node memory (--max-old-space-size=<mb_size>)
+var nodeHeapSizeRe = regexp.MustCompile("--max-old-space-size=([0-9]*)")
```
sebgl: variable/comment mismatch

pkg/license/license.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+// Save updates or creates licensing information in a config map
+func (r LicensingResolver) Save(info LicensingInfo, operatorNs string) error {
+	data, err := info.toMap()
+	if err != nil {
+		return err
+	}
+
+	log.V(1).Info("Saving", "license_info", info)
```
sebgl: Let's also log the config map name and namespace

pkg/license/license.go
sebgl: We'd also need to retrieve StatefulSets in the operator namespace so we get its UID, which is not something we currently have RBAC permissions for.

pkg/license/reporter.go
```diff
@@ -0,0 +1,73 @@
...
+		},
+		licensingResolver: LicensingResolver{
+			client: c,
+		},
+	}
+}
+
+// Start starts to report the licensing information repeatedly at regular intervals
+func (r ResourceReporter) Start(operatorNs string) {
+	ticker := time.NewTicker(refreshPeriod)
```
sebgl: would be nice to call `r.Report()` once as soon as possible so we don't wait the first 2 minutes if we just want to test it out. Cache should be sync already once this is called.

pkg/license/reporter_test.go
```diff
@@ -0,0 +1,162 @@
...
+	kbv1 "github.com/elastic/cloud-on-k8s/pkg/apis/kibana/v1"
+	essettings "github.com/elastic/cloud-on-k8s/pkg/controller/elasticsearch/settings"
+	kbconfig "github.com/elastic/cloud-on-k8s/pkg/controller/kibana/config"
+	"github.com/elastic/cloud-on-k8s/pkg/utils/k8s"
+	"github.com/stretchr/testify/assert"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+)
+
+func Test_Get(t *testing.T) {
```
sebgl: This is the "integration test" I was looking for in my previous comment: seems god enough for me to keep it as a unit test here 👍 

It would be nice if we could add another one to test the configmap reporting?
We could have ResourceReporter.Start accept the `refreshPeriod ` as a parameter, call it in the test with a super small `refreshPeriod`, and have the test check:
- a configmap is eventually created with the right content
- modify the ES spec
- the configmap should eventually be updated with the right content

pkg/license/aggregator.go
thbkrkr:
```
// if memory is defined at the container level, maybe fallback to the environment variable
```

pkg/license/aggregator.go
thbkrkr:
```
// if still no memory found, fallback to the default limits
```

pkg/license/aggregator.go
```diff
@@ -0,0 +1,208 @@
...
+	match := maxHeapSizeRe.FindStringSubmatch(javaOpts)
+	if len(match) != 3 {
+		return resource.Quantity{}, fmt.Errorf("cannot extract max jvm heap size from %s", javaOpts)
+	}
+	value, err := strconv.Atoi(match[1])
+	if err != nil {
+		return resource.Quantity{}, err
+	}
+	// capitalize the suffix unless to have a surjection of [g|G|m|M|k|K] in [Gi|Mi|Ki]
+	suffix := strings.ToUpper(match[2]) + "i"
```
david-kow: I think if it's specified in plain bytes `match[2]` will be empty and adding `i` will cause parsing error.

pkg/license/reporter_test.go
thbkrkr: +100

pkg/license/license.go
thbkrkr: I added the label.
For the owner's reference, this does not seem trivial to achieve.. Considering the timing and the complexity, let's see that later.

thbkrkr:
I just found an issue with the use of `WaitForCacheSync` and the recent modification to call `r.Report()` once as soon as possible to not wait the first 2 minutes.

Current error:

```json
2019-12-19T17:36:51.740+0100    ERROR  resource Failed to report licensing information { "ver": "1.0.0-4c5b4f42",  
  "error": "failed to aggregate Elasticsearch memory: the cache is not started, can not read objects"
}
```

I think the cache needs to be started (so the manager) before calling it. I don't (yet) see a simple way to fix this.

thbkrkr:
I had a chat with @sebgl and we ended up thinking that a `time.Sleep()` seems to be a simple acceptable solution to let the manager start before waiting for the cache sync.

pkg/license/reporter_test.go
```diff
@@ -0,0 +1,221 @@
...
+			Name:      licensingCfgMapName,
+		}, &cm)
+		if err != nil {
+			return false
+		}
+		return cm.Data["timestamp"] != "" &&
+			cm.Data["eck_license_level"] == defaultOperatorLicenseLevel &&
+			cm.Data["enterprise_resource_units"] == "2" &&
+			cm.Data["total_managed_memory"] == "89.12GB"
+	}, refreshPeriod*2, refreshPeriod/2)
```
sebgl: `refreshPeriod*2 ` should be enough unless the system is under heavy load in CI. I would suggest we make it a bit higher than that.
We can also use our own `test.RetryUntilSuccess` which has its own defaults (100ms period & 5 sec timeout).

Issue #1738, opened 2019-09-17
Kibana: set xpack.security.encryptionKey


jordansissel:
[Per the docs:](https://www.elastic.co/guide/en/kibana/7.3/security-settings-kb.html)

> By default, a value is automatically generated in memory. If you use that default behavior, all sessions are invalidated when Kibana restarts. 

The unspoken note here is that with 2 or more instances, each Kibana will generate its own key. This causes all kibana instances to reject eachother's login cookies.

The user experience when this happens goes like this:
* Visit kibana, receive login prompt.
* Login correctly
* Get redirected to whatever landing page is appropriate
* Get a 403/redirected back to the login prompt because your 2nd request actually went to a different Kibana which rejected the previous Kibana instance's session cookie, insisting that you need to login.

To resolve this, set xpack.security.encryptionKey the same on all Kibana instances.

Recommend ECK sets this value for the Kibana resource.

PR #2278, opened 2019-12-17
Ensure Kibana encryption key is specified

anyasabo:
Fixes https://github.com/elastic/cloud-on-k8s/issues/1738

Reads the existing Kibana config secret (if any) and re-uses that, otherwise we generate a new encryption key.

I tested and it does not seem to break our current guidance of using a secure settings secret. Kibana seems a lot looser than Elasticsearch in this regard. So users following the current standard should be able to upgrade seamlessly (though with a restart since the config changed). We should add something to the release notes mentioning it is not necessary.


pkg/controller/kibana/config/settings.go
```diff
@@ -15,10 +15,20 @@ import (
...
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/rand"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
 )
 
 // Kibana configuration settings file
-const SettingsFilename = "kibana.yml"
+const (
+	SettingsFilename = "kibana.yml"
+	name             = "kibana-config"
```
sebgl: took me some time this is used for logging 😄 
I think I would either hardcode it in `logf.Log.WithName(name)` or try to find a more explicit name for the const?

pkg/controller/kibana/config/settings.go
```diff
@@ -42,7 +54,11 @@ func NewConfigSettings(client k8s.Client, kb kbv1.Kibana, versionSpecificCfg *se
 	kibanaTLSCfg := settings.MustCanonicalConfig(kibanaTLSSettings(kb))
 
 	if !kb.RequiresAssociation() {
-		if err := cfg.MergeWith(versionSpecificCfg, kibanaTLSCfg, userSettings); err != nil {
+		if err := cfg.MergeWith(
```
sebgl: Can the following happen?
1. user sets some `foo:bar` setting in their Kibana manifest
2. that setting is now part of the secret
3. they remove it from their Kibana manifest
4. we still inherit it because we merge from the existing secret

pkg/controller/kibana/config/settings.go
```diff
@@ -73,12 +90,38 @@ func NewConfigSettings(client k8s.Client, kb kbv1.Kibana, versionSpecificCfg *se
...
+}
+
 func baseSettings(kb kbv1.Kibana) map[string]interface{} {
 	return map[string]interface{}{
 		ServerName:         kb.Name,
 		ServerHost:         "0",
 		ElasticSearchHosts: []string{kb.AssociationConf().GetURL()},
 		XpackMonitoringUiContainerElasticsearchEnabled: true,
+		// this will get overriden if one already exists or is specified by the user
+		XpackSecurityEncryptionKey: rand.String(64),
```
sebgl: Based on what you said in the PR top post , I would also add something like:
`// if the user specifies one in secure settings, we still generate one in the Kibana configuration file, but the secure settings one takes precedence`

anyasabo: It's actually not clear to me which one takes precedence -- from what I could see there's no way to introspect what encryption key it decides to use, but I could definitely be missing it. I was able to verify that it didn't error out at least and did use _an_ encryption key.

What I meant is that it doesn't get overridden even if they just put it in spec. I'll add a test for that case though

pkg/controller/kibana/config/settings.go
anyasabo: Yes, but that seems okay? Deleting the secret will work to re-generate the secret with a new encryption key, so there at least is a way to do that if that is the goal.

pkg/controller/kibana/config/settings.go
anyasabo: Ah I see in other places we just hardcode it as you suggested, I think the first example I looked at had this pattern so mimicked that. It looks like it is about 50/50 which we do.

pkg/controller/kibana/config/settings.go
sebgl: What I meant was this also apply to any other settings.
If I set:

```yaml
kind: Kibana
metadata:
  name: kibana-sample
spec:
  version: 7.5.0
  count: 1
  elasticsearchRef:
    name: "elasticsearch-sample"
  config:
    logging.verbose: true # temporarily enable verbose logging
```

Then 1 hour later once I'm done with looking at Kibana logs I remove my extra config and apply that yaml:

```yaml
kind: Kibana
metadata:
  name: kibana-sample
spec:
  version: 7.5.0
  count: 1
  elasticsearchRef:
    name: "elasticsearch-sample"
```

I think the code would still merge the new empty user-provided config with the current config which does have `logging.verbose: true` and that's not what I would expect.

Please correct me if I'm wrong, I may be misreading the way we do the config merge.

pkg/controller/kibana/config/settings.go
sebgl: I think we need to make sure if the user sets an encryption key in their secure settings, that's the one that will be used and not our randomly generated one. We may need to inspect the secure settings content?

pkg/controller/kibana/config/settings.go
david-kow: Seems like we have two setting types: for some we want to always flow from the resource (like `logging.verbose`) and apply our default to it and for others we want to persist whatever was generated the first time (like the `xpack.security.encryptionKey`).

Should we just get the specific settings that we know should be persisted from the secret?

pkg/controller/kibana/config/settings.go
```diff
@@ -73,12 +87,38 @@ func NewConfigSettings(client k8s.Client, kb kbv1.Kibana, versionSpecificCfg *se
...
 }
 
+// getExistingConfig retrieves the canonical config for a given Kibana, if one exists
+func getExistingConfig(client k8s.Client, kb kbv1.Kibana) *settings.CanonicalConfig {
+	var secret corev1.Secret
+	err := client.Get(types.NamespacedName{Name: SecretName(kb), Namespace: kb.Namespace}, &secret)
+	if err != nil && apierrors.IsNotFound(err) {
+		log.V(1).Info("Kibana config secret does not exist", "kibana_namespace", kb.Namespace, "kibana_name", kb.Name)
+		return nil
+	} else if err != nil && !apierrors.IsNotFound(err) {
```
david-kow: nit: second condition is not required here (unless for readability).

pkg/controller/kibana/config/settings.go
anyasabo: Ah that's a good point and a big miss on my part Seb, good catch.

>Should we just get the specific settings that we know should be persisted from the secret?

I think this makes sense.

pkg/controller/kibana/config/settings.go
anyasabo: Let me see if I can find out which takes precedence. If the secure setting does not take precedence we may want to do that for all values defined there.

anyasabo:
Confirmed via experimentation and conversations with the Kibana team that the priority order for settings is CLI > secure settings > yaml. For example, a user specifying a secure setting as currently recommended overrides whatever we set in the kibana.yml. So we should be good there as is.

I will think about how to solve the issue with not overwriting the config even when we want to (probably by only extracting specific keys from the existing config) and update the PR.

anyasabo:
This is ready for another look please

pkg/controller/kibana/config/settings.go
```diff
@@ -73,12 +89,57 @@ func NewConfigSettings(client k8s.Client, kb kbv1.Kibana, versionSpecificCfg *se
 	return CanonicalConfig{cfg}, nil
 }
 
+// getExistingConfig retrieves the canonical config for a given Kibana, if one exists
+func getExistingConfig(client k8s.Client, kb kbv1.Kibana) *settings.CanonicalConfig {
+	var secret corev1.Secret
+	err := client.Get(types.NamespacedName{Name: SecretName(kb), Namespace: kb.Namespace}, &secret)
+	if err != nil && apierrors.IsNotFound(err) {
+		log.V(1).Info("Kibana config secret does not exist", "kibana_namespace", kb.Namespace, "kibana_name", kb.Name)
```
thbkrkr: Nit: just use `"namespace"` for the key because that's what we've mainly done so far.

```sh
> grep -r Info | grep -o '"[a-z_]*"' | sort | uniq -c | sort -n | grep namespace
      1 "namespaces"
      3 "owner_namespace"
     58 "namespace"
```

pkg/controller/common/settings/canonical_config.go
```diff
@@ -62,14 +62,24 @@ func MustCanonicalConfig(cfg interface{}) *CanonicalConfig {
...
 // Convenience constructor, will panic in the unlikely event of errors.
 func MustNewSingleValue(k string, v string) *CanonicalConfig {
-	cfg := fromConfig(ucfg.New())
+	cfg := NewCanonicalConfig()
 	err := cfg.asUCfg().SetString(k, -1, v, Options...)
 	if err != nil {
 		panic(err)
 	}
 	return cfg
 }
```
sebgl: Looks like `MustNewSingleValue ` can now just be a call to `NewSingleValue` + panic on error?

pkg/controller/kibana/config/settings.go
```diff
@@ -73,12 +89,57 @@ func NewConfigSettings(client k8s.Client, kb kbv1.Kibana, versionSpecificCfg *se
...
+// getExistingConfig retrieves the canonical config for a given Kibana, if one exists
+func getExistingConfig(client k8s.Client, kb kbv1.Kibana) *settings.CanonicalConfig {
+	var secret corev1.Secret
+	err := client.Get(types.NamespacedName{Name: SecretName(kb), Namespace: kb.Namespace}, &secret)
+	if err != nil && apierrors.IsNotFound(err) {
+		log.V(1).Info("Kibana config secret does not exist", "namespace", kb.Namespace, "kibana_name", kb.Name)
+		return nil
+	} else if err != nil {
+		log.Error(err, "Error retrieving kibana config secret", "namespace", kb.Namespace, "kibana_name", kb.Name)
+		return nil
```
sebgl: I would expect we return the error here. Otherwise it means a failing `Get` call here (network reason or whatever) could end up with a successful `Update` later, and unexpected settings.

pkg/controller/kibana/config/settings.go
```diff
@@ -73,12 +89,57 @@ func NewConfigSettings(client k8s.Client, kb kbv1.Kibana, versionSpecificCfg *se
...
+	rawCfg, exists := secret.Data[SettingsFilename]
+	if !exists {
+		log.Error(nil, "No kibana config file in secret", "namespace", secret.Namespace, "secret_name", secret.Name, "key", SettingsFilename)
+		return nil
+	}
+	cfg, err := settings.ParseConfig(rawCfg)
+	if err != nil {
+		log.Error(err, "Error parsing existing kibana config in secret", "namespace", secret.Namespace, "secret_name", secret.Name, "key", SettingsFilename)
+		return nil
+	}
```
sebgl: If we end up erroring out above, we might also error out on those 2.

pkg/controller/kibana/config/settings.go
anyasabo: Hmmm that makes sense. I'll change it to return an error here and exit the reconcile loop

PR #2290, opened 2019-12-18
Enable debug log level on the operator for E2E tests

sebgl:
This should make it easier to debug failing E2E tests if the operator
gives us a bit more logs.

### DIFF
--- a/config/e2e/global_operator.yaml
+++ b/config/e2e/global_operator.yaml
@@ -166,7 +166,7 @@ spec:
       - image: {{ .OperatorImage }}
         imagePullPolicy: IfNotPresent
         name: manager
-        args: ["manager", "--operator-roles", "global,webhook"]
+        args: ["manager", "--operator-roles", "global,webhook", "--log-verbosity=1"]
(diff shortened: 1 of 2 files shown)


charith-elastic:
Should we do this on a case-by-case basis (by making it configurable). This change would also affect matrix jobs like the stack versions job where the logs are already very large and hard to parse.

sebgl:
Oh you're right @charith-elastic I did not think about it.

Now that we have tgz archives updated by Jenkins with the full logs I'm wondering whether we should remove the operator logs entirely from the Jenkins job output?
If we need the logs, we can look into the tgz archive. WDYT?

charith-elastic:
I wasn't thinking clearly. This change would only change the operator log level -- which doesn't get captured by Jenkins anyway. It should be fine to go ahead with this PR. Sorry!

sebgl:
@charith-elastic oh you're right we're only grabbing the tests logs + k8s events, not the operator logs.
I wasn't thinking clearly either 😄

PR #2292, opened 2019-12-18
Update operator config doc

charith-elastic:
- Document how to correctly specify duration values and environment variables
- Sort flags and simplify the table

[Preview](http://cloud-on-k8s_2292.docs-preview.app.elstc.co/guide/en/cloud-on-k8s/master/k8s-operator-config.html)

### DIFF
--- a/docs/operator-config.asciidoc
+++ b/docs/operator-config.asciidoc
@@ -9,7 +9,7 @@ endif::[]
 ECK can be configured using either command line flags or environment variables.
 
 
-[width="100%",valign="middle",halign="center",cols="~m,~m,d",options="header"]
+[width="100%",cols=".^35m,.^25m,.^40d",options="header"]
 |===
 |Flag |Default|Description
 |auto-port-forward |false |Enables automatic port forwarding to allow running the operator outside the cluster. For dev use only as it exposes k8s resources on ephemeral ports to localhost.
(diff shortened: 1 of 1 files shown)


anyasabo:
I wonder if we should increase the width, I notice that the new version wraps for instance localhost even on my big ol monitor

in this pr
Screen Shot 2019-12-18 at 3 41 31 PM


current
Screen Shot 2019-12-18 at 3 37 48 PM

In general I'm a fan of switching to the monospaced font for the defaults

docs/operator-config.asciidoc
```diff
@@ -6,53 +6,56 @@ endif::[]
...
+|ms | Milliseconds
+|s  | Seconds
+|m  | Minutes
+|h  | Hours
+|===
 
-and change the following lines from:
 
-[source,yaml]
+Edit the operator deployment to change any of the flag values. For example, to enable `DEBUG` level logs, the operator needs to be restarted with the `--enable-debug-logs` flag set to `true`. This can be done as follows:
```
anyasabo: I think we can clarify that it just needs to be edited and then the sset controller will handle restarting

charith-elastic:
> I wonder if we should increase the width, I notice that the new version wraps for instance localhost even on my big ol monitor

I tried a bunch of things but I think the custom CSS in our docs site does not fully implement all the Asciidoctor table formatting rules. Apart from the entry for `localhost:6060`, everything else looks OK though. If you feel strongly about it, I think the only solution is to revert back to non-monospaced fonts.

PR #2295, opened 2019-12-18
Update kind setup

graphaelli:


hack/kind/kind.sh
```diff
@@ -86,17 +86,18 @@ function setup_kind_cluster() {
...
   # Create Kind cluster
-  if ! (kind create cluster --name=${CLUSTER_NAME} ${config_opts} --loglevel "${KIND_LOG_LEVEL}" --retain --image "${NODE_IMAGE}"); then
+  if ! (kind ${KIND_LOGGING} create cluster --name=${CLUSTER_NAME} ${config_opts} --retain --image "${NODE_IMAGE}"); then
     echo "Could not setup Kind environment. Something wrong with Kind setup."
     exit 1
   fi
 
-  KUBECONFIG="$(kind get kubeconfig-path --name="${CLUSTER_NAME}")"
-  export KUBECONFIG
+  # persist kubeconfig for reliabililty in following kubectl commands
```
anyasabo: does the path change or what is the issue with the original variable? it's not clear to me what is unreliable about it

graphaelli: good point. `kind` sets the default context when creating a cluster so if not running kind elswhere in parallel, these `kubectl` commands would just work without this `kubeconfig`.  If that's not something to worry about we can remove this altogether.

anyasabo: ah got it thank you. I'm not sure when we'll be doing that but I haven't worked with it much so it's definitely possible I'm missing something obvious.

graphaelli:
Thanks for reviewing - just a reminder that I'm not authorized to merge so please go ahead if you're ok with these changes.

PR #2301, opened 2019-12-19
Remove enable-debug-logs flag

charith-elastic:
The `enable-debug-logs` flag was an alias for `log-verbosity=0` and was kept for backward compatibility reasons. This PR removes `enable-debug-logs` so that there's only one flag available for changing the log level of the operator.

### DIFF
--- a/config/operator/all-in-one/operator.template.yaml
+++ b/config/operator/all-in-one/operator.template.yaml
@@ -21,7 +21,7 @@ spec:
       containers:
       - image: <OPERATOR_IMAGE>
         name: manager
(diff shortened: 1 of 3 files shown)


thbkrkr:
[`docs/operator-config.asciidoc`](https://github.com/elastic/cloud-on-k8s/blob/master/docs/operator-config.asciidoc) also has some references to this flag.

charith-elastic:
I have #2292 open to update that document

charith-elastic:
Jenkins test this please

PR #2313, opened 2019-12-20
Increase e2e test timeout

david-kow:
We seem to be hitting 2h time limit of `go test` and we seem to be very close to hitting 3h limit of entire Jenkins pipeline for AKS. This PR raises both those limits.

### DIFF
--- a/test/e2e/run.sh
+++ b/test/e2e/run.sh
@@ -12,8 +12,8 @@ set -euo pipefail
 for PKG in $(go list github.com/elastic/cloud-on-k8s/test/e2e/...); do
     if [ "$E2E_JSON" == "true" ]
     then
-        go test -v -failfast -timeout=2h -tags=e2e -p=1 --json $PKG $@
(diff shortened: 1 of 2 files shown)


test/e2e/run.sh
```diff
@@ -12,8 +12,8 @@ set -euo pipefail
 for PKG in $(go list github.com/elastic/cloud-on-k8s/test/e2e/...); do
     if [ "$E2E_JSON" == "true" ]
     then
-        go test -v -failfast -timeout=2h -tags=e2e -p=1 --json $PKG $@
+        go test -v -failfast -timeout=4h -tags=e2e -p=1 --json $PKG $@
```
sebgl: Should we reach go test timeout before Jenkins timeout to get better test output?

david-kow: :+1: I did that, but it's still not guaranteed as `go test` timeout is per package and Jenkins one is overall.

Issue #2291, opened 2019-12-18
Single node cluster upgrade from 6.X to 7.X does not have cluster.initial_master_nodes set


sebgl:
Spotted in an E2E test: https://github.com/elastic/cloud-on-k8s/issues/2288

A single-node cluster upgrade from 6.X to 7.X seems to fail because the restarted 7.X node does not have `cluster.initial_master_nodes` set:

```
{"type": "server", "timestamp": "2019-12-18T02:16:51,944+0000", "level": "WARN", "component": "o.e.c.c.ClusterFormationFailureHelper", "cluster.name": "test-version-up-1-685-to-732-grq9", "node.name": "test-version-up-1-685-to-732-grq9-es-masterdata-0",  "message": "master not discovered yet, this node has not previously joined a bootstrapped (v7+) cluster, and [cluster.initial_master_nodes] is empty on this node: have discovered [{test-version-up-1-685-to-732-grq9-es-masterdata-0}{qrttZLBcQRifqkZmtte6xg}{Kr-otzquS2mllFoK02HECA}{10.64.1.70}{10.64.1.70:9300}{dim}{ml.machine_memory=2147483648, xpack.installed=true, ml.max_open_jobs=20}]; discovery will continue using [127.0.0.1:9300, 127.0.0.1:9301, 127.0.0.1:9302, 127.0.0.1:9303, 127.0.0.1:9304, 127.0.0.1:9305] from hosts providers and [{test-version-up-1-685-to-732-grq9-es-masterdata-0}{qrttZLBcQRifqkZmtte6xg}{Kr-otzquS2mllFoK02HECA}{10.64.1.70}{10.64.1.70:9300}{dim}{ml.machine_memory=2147483648, xpack.installed=true, ml.max_open_jobs=20}] from last-known cluster state; node term 0, last-accepted version 13 in term 0"  }
```

This was supposed to be fixed in https://github.com/elastic/cloud-on-k8s/pull/1681 but it looks like we might have a bug or race condition somewhere.
[Elasticsearch 7.3 pod logs.txt](https://github.com/elastic/cloud-on-k8s/files/3978778/logs.txt)
[Operator pod logs.txt](https://github.com/elastic/cloud-on-k8s/files/3978782/logs.txt)

sebgl:
Investigating the logs.

Elasticsearch instance reports `cluster.initial_master_nodes` not set:

```
[...]

PR #2315, opened 2019-12-20
Fix how cluster.initial_master_nodes is set

sebgl:
Fixes https://github.com/elastic/cloud-on-k8s/issues/2291.

We had a bug where the value of `cluster.initial_master_nodes` would not
be set while upgrading a single v6 cluster to v7. This commit fixes it
by handling `cluster.initial_master_nodes` differently.

:: Annotation logic

Instead of relyin on the ClusterUUID annotation to know whether a
cluster is bootstrapped, and manually removing on the special case of a
single v6 -> v7 annotation, we now rely on a dedicated annotation:
`elasticsearch.k8s.elastic.co/initial-master-nodes: node-0,node-1,node-2`.

When that annotation is set at the cluster level, it means the cluster
is currently bootstrapping for zen2. Any v7+ master node must have
`cluster.initial_master_nodes` in its configuration set to the value in
the annotation. This value is not supposed to vary over time, we make
sure that is the case here.

Once we detect the cluster has finished its bootstrap (the current
master of the cluster is a v7+ master node), we remove the annotation,
and remove the configuration setting from master nodes.

:: When should cluster.initial_master_nodes be set?

There are 2 cases where this setting must be set:
- when we create a v7+ cluster for the first time
- when we upgrade/restart a single v6 master to v7: that new master must
identify itself as a legit master

We don't want to set this setting when:
- we do a regular rolling-upgrade of multiple master nodes from v6 to v7
- Elasticsearch is not running in v7+

:: Edge cases

The following cases are quite tricky and covered by unit & E2E tests:

* upgrade from a single v6 master to a single v7 master **in a different
NodeSet**. The new master will be created before the old one gets
removed, hence the setting should not be set.
* upgrade from a single v6 master to more v7 masters. The new v7 masters
will be created before the v6 master is upgraded, hence the setting
should not be set.
* upgrade from two v6 masters to two v7 masters. This is a regular
rolling upgrade, hence the setting should not be set. However when the
first master goes down for upgrade, the cluster becomes unavailable
since minimum_master_nodes=2.


pkg/controller/elasticsearch/client/model.go
```diff
@@ -66,16 +68,25 @@ func (n Nodes) Names() []string {
...
 		} `json:"mem"`
 	} `json:"jvm"`
 }
 
+func (n Node) isV7OrAbove() (bool, error) {
+	v, err := version.Parse(n.Version)
+	if err != nil {
+		return false, errors.Wrap(err, fmt.Sprintf("unable to parse node version %s", n.Version))
+	}
+	return v.Major > 7, nil
```
barkbay: I think there's a bug here, `isV7OrAbove` actually returns `false` if the major version of the cluster is 7.

pkg/controller/elasticsearch/version/zen2/initial_master_nodes.go
```diff
@@ -5,32 +5,182 @@
...
+
+	// if the cluster is annotated with `cluster.initial_master_nodes` (zen2 bootstrap in progress),
+	// make sure we reuse that value since it is not supposed to vary over time
+	if initialMasterNodes := getInitialMasterNodesAnnotation(es); initialMasterNodes != nil {
+		return patchInitialMasterNodesConfig(nodeSpecResources, initialMasterNodes)
+	}
+
+	// we want to set `cluster.initial_master_nodes` if either:
+	// - a new cluster is getting created
+	isInitialClusterBootstrap := !bootstrap.AnnotatedForBootstrap(es)
```
barkbay: nit: we later use `!isInitialClusterBootstrap`, there is a double negation which make the code a little bit tricky to understand at first :)

pkg/controller/elasticsearch/version/zen2/initial_master_nodes.go
```diff
@@ -5,32 +5,182 @@
...
+// Rolling upgrades from eg. v6 to v7 do not need that setting.
+// It ensures `cluster.initial_master_nodes` does not vary over time, when this function gets called multiple times.
+func SetupInitialMasterNodes(es esv1.Elasticsearch, k8sClient k8s.Client, nodeSpecResources nodespec.ResourcesList) error {
+	if v, err := version.Parse(es.Spec.Version); err != nil || !versionCompatibleWithZen2(*v) {
+		// we only care about zen2-compatible clusters here
+		return err
+	}
+
+	// if the cluster is annotated with `cluster.initial_master_nodes` (zen2 bootstrap in progress),
+	// make sure we reuse that value since it is not supposed to vary over time
```
barkbay: I'm actually wondering if it is really the case, I think the user may want to downscale or change the topology before the cluster is bootstrapped.

pkg/controller/elasticsearch/client/v6.go
```diff
@@ -112,6 +113,24 @@ func (c *clientV6) DeleteVotingConfigExclusions(ctx context.Context, waitForRemo
...
+func (c *clientV6) ClusterBootstrappedForZen2(ctx context.Context) (bool, error) {
+	// Look at the current master node of the cluster: if it's running version 7.x.x or above,
+	// the cluster has been bootstrapped.
+	// Even though c is a clientV6, it may be targeting a mixed v6/v7 having a v7 master.
+	var response Nodes
+	if err := c.get(ctx, "/_nodes/_master", &response); err != nil {
+		return false, err
+	}
+	if len(response.Nodes) != 1 {
+		return false, fmt.Errorf("GET /_nodes/_master returned %d nodes", len(response.Nodes))
```
barkbay: I'm trying to understand if returning an error is the right thing to do here. If there's no master it could be because election did not happen yet, imho it's not an error from the operator POV. My fear is that the error here prevents a downscale or an upgrade of the nodes.

pkg/controller/elasticsearch/client/model.go
sebgl: Wow, indeed !

sebgl: Thanks for fixing it!

pkg/controller/elasticsearch/version/zen2/initial_master_nodes.go
sebgl: Agreed 👍 
I'll change to the opposite: `isAlreadyBootstrapped`

pkg/controller/elasticsearch/client/v6.go
sebgl: If there's no master I'm pretty sure we will not be able to downscale or upgrade nodes anyway? Also I think we don't get an HTTP response 200 to the call, but 503 if there's no master.
I think it's OK to leave as is?

pkg/controller/elasticsearch/version/zen2/initial_master_nodes.go
sebgl: Yeah I was wondering the same.
@DaveCTurner @ywelsch any opinion here maybe? Some context to understand what this is about:

We set `cluster.initial_master_nodes` when:
* a v7+ cluster is created for the first time
* we upgrade a single v6 master to v7 through a restart

In both cases, we keep track of the value of `cluster.initial_master_nodes` we set, so we always  reuse the same, until not necessary anymore because the v7 cluster has formed.

While the v7 cluster is starting, the user could apply a different specification for the ES cluster, leading to the reused `cluster.initial_master_nodes` value to not match the new cluster specification.
An example:
1. Apply a 1-node v7 cluster spec. ECK sets `cluster.initial_master_nodes: onenode-0`.
2. Change the spec to a 3-nodes v7 cluster while the cluster hasn't even started yet. ECK will reuse `cluster.initial_master_nodes: onenode-0` matching a node that does not exist.

If this is a fresh new cluster it's maybe not a big deal (no cluster ever existed, no data loss): the cluster will never form but a way out for the user is to remove the cluster and recreate it.
If this is a v6->v7 upgrade, ECK will not remove the existing master node from the cluster until it is safe to do so (at least another master running).

DaveCTurner: I think supporting downscaling a cluster before it's formed is unreasonably heroic. As you say, with a brand-new cluster you've always got the option to safely start again from scratch.

If you want to be heroic anyway, since bootstrapping is majority-based it should be ok to drop <½ of the nodes before it's bootstrapped (without changing `cluster.initial_master_nodes`) but note that you can't add voting config exclusions if the cluster hasn't formed so you have to wait for bootstrapping before removing ≥½ of the existing nodes anyway.

Downscaling doesn't make sense in the context of a one-node cluster since it's only got one node. More generally, whatever topology change you might make must still involve the node that existed before the upgrade, so setting `cluster.initial_master_nodes: unique-existing-node` will still work.

pkg/controller/elasticsearch/client/v6.go
DaveCTurner: If there's no master then `GET /_nodes` doesn't respond immediately, but if it times out while waiting for the master (default `30s` I think) then it returns 503.

DaveCTurner: `GET _nodes/_master` isn't a master-node action, it uses the node's local cluster state to work out which nodes are involved. This means it returns `200 OK` (but no nodes) if there is no master.

barkbay: > we will not be able to downscale or upgrade nodes anyway

Regarding the upgrade I think we should. That said, today, if there is no master (because of a bug or a bad configuration) user will have to delete Pods manually today because `ClearVotingConfigExclusions` will fail _(I think I will open an issue at least to document it)_

pkg/controller/elasticsearch/version/zen2/initial_master_nodes.go
barkbay: 👍 Makes sense. The use case I had in mind was a user hitting a quota, decreasing the number of nodes in order to sort out the quota issue.
I agree that in such a situation the user can delete and create the cluster again.

pkg/controller/elasticsearch/client/v6.go
sebgl: Thanks for chiming in @DaveCTurner. Seems better to not return an error if the list of nodes is empty as you suggested @barkbay 👍 
I'm doing the change.

Issue #2320, opened 2019-12-20
Pods are force-upgraded on initial StatefulSet creation


sebgl:
It looks like we're triggering a force-upgrade on Pods that were just created.

To reproduce:

- deploy an Elasticsearch cluster
- look at the operator logs: it's doing a force upgrade of Pods that were juste created

There is no data loss here or impact on the cluster since it only concerns newly created Pods (not even started yet).

PR #2321, opened 2019-12-20
Don't upgrade pods if empty StatefulSet UpdateRevision

sebgl:
We had a bug were some Pods of a newly created StatefulSet would be
force upgraded because they are Pending, and we consider they need to be
upgraded because their revision does not match the StatefulSet update
revision.
That's because the StatefulSet updateRevision is empty on initial
StatefulSet creation.

This commit fixes it by not considering empty StatefulSet
updateRevisions when looking for Pods to upgrade.

Fixes https://github.com/elastic/cloud-on-k8s/issues/2320.

### DIFF
--- a/pkg/controller/elasticsearch/driver/upgrade.go
+++ b/pkg/controller/elasticsearch/driver/upgrade.go
@@ -174,6 +174,10 @@ func podsToUpgrade(
 ) ([]corev1.Pod, error) {
 	var toUpgrade []corev1.Pod
 	for _, statefulSet := range statefulSets {
+		if statefulSet.Status.UpdateRevision == "" {
+			// no upgrade scheduled
+			continue
+		}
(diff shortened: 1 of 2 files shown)


pkg/controller/elasticsearch/driver/upgrade.go
```diff
@@ -190,28 +194,14 @@ func podsToUpgrade(
 				// Pod does not exist, continue the loop as the absence will be accounted by the deletion driver
 				continue
 			}
-			alreadyUpgraded := podUpgradeDone(pod, statefulSet.Status.UpdateRevision)
-			if !alreadyUpgraded {
+			if sset.PodRevision(pod) != statefulSet.Status.UpdateRevision {
```
sebgl: this is just extracting the important bit from the removed function below

PR #2324, opened 2019-12-23
Add cert-manager example

charith-elastic:
Adds an example section on how to create custom certificates for Elasticsearch using cert-manager. If users forget to set `isCA` in the certificate resource, Kibana refuses to connect to Elasticsearch with the unhelpful generic error message `No living connections` -- which is not easy to debug.

[Preview](http://cloud-on-k8s_2324.docs-preview.app.elstc.co/guide/en/cloud-on-k8s/master/k8s-custom-http-certificate.html)

### DIFF
--- a/docs/elasticsearch-specification.asciidoc
+++ b/docs/elasticsearch-specification.asciidoc
@@ -258,12 +258,46 @@ spec:
         secretName: my-cert
 ----
 
-This is an example on how to create a Kubernetes TLS secret with a self-signed certificate:
+[float]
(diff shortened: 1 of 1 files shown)


docs/elasticsearch-specification.asciidoc
```diff
@@ -258,12 +258,46 @@ spec:
         secretName: my-cert
```
anyasabo: Is it worth updating this to `quickstart-es-cert` so that someone can just combine both the ES spec example and the cert manager example?

PR #2326, opened 2019-12-24
Add version awareness to TestKibanaStandalone E2E test

### LINKED ISSUE #2325: TestKibanaStandalone E2E test fails on Kibana 6.8.5
In Kibana 6.x the association configuration should be `elasticsearch.url` but the test is not version aware and uses the newer `elasticsearch.hosts` setting which causes Kibana to fail with a config validation error.

```
FATAL  ValidationError: child "elasticsearch" fails because [child "hosts" fails because [single value of "hosts" fails because ["hosts" is not allowed to be empty, "hosts" must be a valid uri with a scheme matching the http|https pattern]]]
```

charith-elastic:
`TestKibanaStandalone` E2E test fails on 6.x stack versions because the configuration setting for specifying the Elasticsearch URL should be `elasticsearch.url` instead of `elasticsearch.hosts`. This PR changes the test to pick the correct manifest based on the stack version being tested.

Fixes #2325

### DIFF
--- a/test/e2e/kb/standalone_test.go
+++ b/test/e2e/kb/standalone_test.go
@@ -10,6 +10,7 @@ import (
 	"testing"
 	"text/template"
 
+	"github.com/elastic/cloud-on-k8s/pkg/controller/common/version"
 	"github.com/elastic/cloud-on-k8s/test/e2e/test"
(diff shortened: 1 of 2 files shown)


charith-elastic:
Closing in favour of #2331

PR #2330, opened 2019-12-30
Allow node restart even if cluster health is yellow

### LINKED ISSUE #1643: Version upgrades could stop making progress if replicas need to be placed on not upgraded nodes
When we upgraded less nodes than there are replicas for a primary than is on one of the upgraded nodes, the cluster could stay yellow. Replicas can't be placed if new version contains internal representation changes that are incompatible with older version.

barkbay:
This PR fixes #1643 by adding two predicates to the rolling upgrade process in order to allow some Pods to be restarted even if the cluster health is yellow.
The first predicate only allow for a Pod to be restarted in the following conditions:

* The version of the Pod does not match the version of the cluster
* All primaries are assigned, only replicas are actually not assigned
* There are no `initializing` or `relocating` shards

A second predicate ensures that the Pod with the last remaining started primary Shard is not deleted

This PR, in a nutshell, merely implements the recommendations found in the [documentation](https://www.elastic.co/guide/en/elasticsearch/reference/current/rolling-upgrades.html#_upgrading_your_cluster):

>If it is not possible to assign the replica shards to another node (there is only one upgraded node in the cluster), the replica shards remain unassigned and status stays yellow.
> In this case, you can proceed once there are no initializing or relocating shards (check the init and relo columns).

As a side note I think I will open an other PR to refactor the predicates array which is becoming quite ... large.

### DIFF
--- a/pkg/controller/elasticsearch/driver/upgrade_predicates.go
+++ b/pkg/controller/elasticsearch/driver/upgrade_predicates.go
@@ -240,9 +240,9 @@ var predicates = [...]Predicate{
 					continue
 				}
 				shardKey := shard.Key()
-				replicas[shardKey] += 1
+				replicas[shardKey]++
 				if shard.State == client.STARTED {
-					startedReplicas[shardKey] += 1
+					startedReplicas[shardKey]++
 				}
 			}
 
--- a/pkg/controller/elasticsearch/driver/upgrade_predicates.go
+++ b/pkg/controller/elasticsearch/driver/upgrade_predicates.go
@@ -217,7 +217,9 @@ var predicates = [...]Predicate{
 		},
 	},
 	{
-		name: "do_not_delete_last_started_shard",
+		// We may need to delete nodes in a yellow cluster, but not if they contain the only replica
+		// of a shard since it would make the cluster go red.
+		name: "require_started_replica",
 		fn: func(
 			context PredicateContext,
 			candidate corev1.Pod,
@@ -238,15 +240,13 @@ var predicates = [...]Predicate{
 					continue
 				}
 				shardKey := shard.Key()
-				shardReplica := replicas[shardKey]
-				replicas[shardKey] = shardReplica + 1
+				replicas[shardKey] += 1
 				if shard.State == client.STARTED {
-					assignedReplica := startedReplicas[shardKey]
-					startedReplicas[shardKey] = assignedReplica + 1
+					startedReplicas[shardKey] += 1
 				}
 			}
 
-			// Do not delete a node with a Primary if at least one replica is not STARTED
+			// Do not delete a node with a Primary if there is not at least one STARTED replica
 			shardsByNode := allShards.GetShardsByNode()
 			shardsOnCandidate := shardsByNode[candidate.Name]
 			for _, shard := range shardsOnCandidate {
@@ -254,9 +254,10 @@ var predicates = [...]Predicate{
 					continue
 				}
 				shardKey := shard.Key()
-				replicas := replicas[shardKey]
+				numReplicas := replicas[shardKey]
 				assignedReplica := startedReplicas[shardKey]
-				if replicas > 0 && assignedReplica == 0 {
+				// We accept here that there will be some unavailability if an index is configured with zero replicas
+				if numReplicas > 0 && assignedReplica == 0 {
 					// If this node is deleted there will be no more shards available
 					return false, nil
 				}
--- a/pkg/controller/elasticsearch/driver/upgrade_predicates.go
+++ b/pkg/controller/elasticsearch/driver/upgrade_predicates.go
@@ -142,23 +142,23 @@ var predicates = [...]Predicate{
 		},
 	},
 	{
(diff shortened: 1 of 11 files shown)


pkg/controller/elasticsearch/driver/upgrade_predicates.go
```diff
@@ -168,6 +168,102 @@ var predicates = [...]Predicate{
...
+			// * How many of them are STARTED
+			startedReplicas := make(map[string]int)
+			replicas := make(map[string]int)
+			for _, shard := range allShards {
+				if shard.NodeName == candidate.Name {
+					continue
+				}
+				shardKey := shard.Key()
+				shardReplica := replicas[shardKey]
+				replicas[shardKey] = shardReplica + 1
```
charith-elastic: Nit: It would be easier to read if this was written as `replicas[shardKey] += 1`. I think the compiler can optimize it as well even though that's not very important in this context.

pkg/controller/elasticsearch/driver/upgrade_predicates.go
```diff
@@ -168,6 +168,102 @@ var predicates = [...]Predicate{
...
+				if shard.State == client.STARTED {
+					assignedReplica := startedReplicas[shardKey]
+					startedReplicas[shardKey] = assignedReplica + 1
```
charith-elastic: Same nit as above: `startedReplicas[shardKey] += 1`

pkg/controller/elasticsearch/driver/upgrade_predicates.go
```diff
@@ -168,6 +168,102 @@ var predicates = [...]Predicate{
...
+					if !shard.IsReplica() {
+						return false, nil
+					}
+				}
+			}
+			return true, nil
+		},
+	},
+	{
+		name: "do_not_delete_last_started_shard",
```
anyasabo: Maybe this could be rephrased to `require_started_replica`, `do_not_delete_only_started_shard` or something similar? At least when I read this the first time I read it as "do not delete the most recently started shard" rather than "do not delete the only started shard". Maybe a comment explaining that we still want to delete nodes in a yellow cluster if we have to, but not if they contain the only replica of a shard and thus would make the cluster go red? English is weird.

pkg/controller/elasticsearch/driver/upgrade_predicates.go
```diff
@@ -168,6 +168,102 @@ var predicates = [...]Predicate{
...
+				shardKey := shard.Key()
+				shardReplica := replicas[shardKey]
+				replicas[shardKey] = shardReplica + 1
+				if shard.State == client.STARTED {
+					assignedReplica := startedReplicas[shardKey]
+					startedReplicas[shardKey] = assignedReplica + 1
+				}
+			}
+
+			// Do not delete a node with a Primary if at least one replica is not STARTED
```
anyasabo:
```suggestion
			// Do not delete a node with a Primary if there is not at least one STARTED replica
```

pkg/controller/elasticsearch/driver/upgrade_predicates.go
```diff
@@ -168,6 +168,102 @@ var predicates = [...]Predicate{
...
+			shardsByNode := allShards.GetShardsByNode()
+			shardsOnCandidate := shardsByNode[candidate.Name]
+			for _, shard := range shardsOnCandidate {
+				if !shard.IsPrimary() {
+					continue
+				}
+				shardKey := shard.Key()
+				replicas := replicas[shardKey]
```
anyasabo: nit: changing the type of replicas from a map to an int was confusing for me at least, maybe this can be `numReplicas`?

pkg/controller/elasticsearch/driver/upgrade_predicates.go
```diff
@@ -168,6 +168,102 @@ var predicates = [...]Predicate{
...
+				assignedReplica := startedReplicas[shardKey]
+				if replicas > 0 && assignedReplica == 0 {
```
anyasabo: It may be worthwhile to add a comment indicating that we accept there will be unavailability in the event that an index is configured with zero replicas

pkg/controller/elasticsearch/driver/upgrade_predicates.go
```diff
@@ -168,6 +168,103 @@ var predicates = [...]Predicate{
...
+		// We may need to delete nodes in a yellow cluster, but not if they contain the only replica
+		// of a shard since it would make the cluster go red.
+		name: "require_started_replica",
+		fn: func(
+			context PredicateContext,
+			candidate corev1.Pod,
+			deletedPods []corev1.Pod,
+			maxUnavailableReached bool,
+		) (b bool, e error) {
+			allShards, err := context.shardLister.GetShards()
```
sebgl: Should we execute this only if the cluster is yellow? So we don't have to list shards if the cluster is already green.

sebgl: We already do as part of the `"do_not_delete_pods_with_same_shards"` predicate so it does not really matter.

PR #2331, opened 2019-12-30
Remove Kibana 6.x settings factory

### LINKED ISSUE #2325: TestKibanaStandalone E2E test fails on Kibana 6.8.5

charith-elastic:
The version 6 settings factory was only used to override the Elasticsearch connection string. Kibana 6.8.x (minimum supported version in ECK), has support for the equivalent 7.x setting `elasticsearch.hosts` and does not require a special settings factory any more.

Fixes #2325


charith-elastic:
Since we already have some specialisation for 7.6.x, I assumed that we will need to keep the factory concept around to support upcoming stack releases. Happy to get rid of it completely if you think that's unlikely.

sebgl:
Would it make sense to remove both `version6` and `version7` packages, and move the current `version/version7/settings.go` to `version/settings.go`?

Issue #2338, opened 2020-01-03
Cannot create master nodes if not in the first StatefulSet (E2E test TestReversalIllegalConfig fails)


sebgl:
The test TestReversalIllegalConfig fails in CI.

After investigating a bit, it looks like `cluster.initial_master_nodes` cannot be computed since there is no master node in the expected replicas.
Debugging further, I think we have a bug when limiting the number of master nodes to create in the upscale state.

PR #2339, opened 2020-01-03
Reuse the same upscaleState across StatefulSets

sebgl:
We had a bug where the second StatefulSet containing master nodes would
have its replicas set to 0, because it is not allowed to add additional
masters.

Turns out that's because we pass the upscaleState by value, so any mutation
we do on it is not preserved at the next call for the next StatefulSet. Moreover
it contains a sync.Once initialization not reused the second time. The first
StatefulSet is allowed master nodes creation thanks to the upscale state
initialization, but the second StatefulSet isn't because its
upscaleState is zeroed-out and not re-initialized.

This commit fixes it by passing around a StatefulSet pointer, and adds a
unit test that fails without the fix.

Fixes https://github.com/elastic/cloud-on-k8s/issues/2338.

### DIFF
--- a/pkg/controller/elasticsearch/driver/upscale.go
+++ b/pkg/controller/elasticsearch/driver/upscale.go
@@ -99,6 +99,8 @@ func adjustZenConfig(k8sClient k8s.Client, es esv1.Elasticsearch, resources node
 	return nil
 }
 
+// adjustStatefulSetReplicas updates the replicas count in expected according to
+// what is allowed by the upscaleState, that may be mutated as a result.
 func adjustStatefulSetReplicas(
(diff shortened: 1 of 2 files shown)


pkg/controller/elasticsearch/driver/upscale.go
```diff
@@ -100,7 +100,7 @@ func adjustZenConfig(k8sClient k8s.Client, es esv1.Elasticsearch, resources node
 }
 
 func adjustStatefulSetReplicas(
```
nkvoll: Can we have a small doc/comment here clarifying why `upscaleState` is a pointer specifically?

Issue #2319, opened 2019-12-20
Document how to retrieve the Elastic usage config map for licensing


thbkrkr:
Follow-up #2277.

Draft:

You can check your current operator license and how much total memory under management is available by retrieving the `elastic-licensing` ConfigMap located in the operator namespace:

```sh
> kubectl -n elastic-system get configmap elastic-licensing -o json | jq .data
{
  "enterprise_resource_units": "1",
  "license_level": "basic",
  "memory": "5.37GB",
  "timestamp": "2019-12-17T11:42:15+01:00"
}
```
or?
```sh
> kubectl -n elastic-system get configmap elastic-licensing -o=yaml | grep -A4 ^data: 
data:
  eck_license_level: basic
  enterprise_resource_units: "1"
  timestamp: "2019-12-20T16:16:08+01:00"
  total_managed_memory: 5.37GB

```

PR #2346, opened 2020-01-03
Add license config map docs

anyasabo:
Fixes https://github.com/elastic/cloud-on-k8s/issues/2319

### DIFF
--- a/docs/licensing.asciidoc
+++ b/docs/licensing.asciidoc
@@ -72,3 +72,18 @@ we recommend you install the new enterprise license as a new Kubernetes secret n
 Just replace `eck-license` with a different name in the examples above. ECK will pick the right license automatically.
 
 As soon as you have installed the new license you can safely delete the old license if you want.
+
+[float]
+=== Getting usage data
+The operator periodically writes the total amount of Elastic resources under management to a config map. It is named `elastic-licensing` in the same namespace as the operator. For an example of retrieving the data:
+
+[source,shell]
+----
+> kubectl -n elastic-system get configmap elastic-licensing -o json | jq .data
+{
+  "eck_license_level": "basic",
+  "enterprise_resource_units": "1",
+  "timestamp": "2020-01-03T23:38:20Z",
+  "total_managed_memory": "3.22GB"
+}
+----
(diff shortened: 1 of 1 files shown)


docs/licensing.asciidoc
```diff
@@ -71,4 +71,19 @@ To avoid any unintended downgrade of individual Elasticsearch clusters to a basi
...
\ No newline at end of file
+As soon as you have installed the new license you can safely delete the old license if you want.
+
+[float]
+=== Getting usage data
+The operator periodically writes the total amount of Elastic resources under management to a config map. It is named `elastic-licensing` in the same namespace as the operator. Here is an example of retrieving the data:
+
+[source,shell]
+----
+> kubectl -n elastic-system get configmap elastic-licensing -o json | jq .data
```
pebrc: Would it make sense to avoid an external tool here and just do `-o go-template='{{.data}}`?

docs/licensing.asciidoc
```diff
@@ -71,4 +71,19 @@ To avoid any unintended downgrade of individual Elasticsearch clusters to a basi
 we recommend you install the new enterprise license as a new Kubernetes secret next to your existing enterprise license.
 Just replace `eck-license` with a different name in the examples above. ECK will pick the right license automatically.
 
-As soon as you have installed the new license you can safely delete the old license if you want.
\ No newline at end of file
+As soon as you have installed the new license you can safely delete the old license if you want.
```
alaudazzi:
```suggestion
As soon as you have installed the new license, you can safely delete the old license if you want.
```

docs/licensing.asciidoc
anyasabo: It doesn't pretty print it though and this is really just for illustrative purposes so people can see the contents. For example
```
09:58 $ kubectl -n elastic-system get configmap elastic-licensing -o go-template='{{.data}}'
map[eck_license_level:basic enterprise_resource_units:1 timestamp:2020-01-07T15:58:20Z total_managed_memory:3.22GB]
```

is a little bit harder to read in the docs IMO, but I see your point

pebrc: No you are right, I did not think about pretty printing

Issue #2355, opened 2020-01-06
Webhook does not return any error for resources deployed in version v1beta1


sebgl:
To reproduce: 

* Deploy ECK 1.0.0-beta1 (not sure if necessary)
* Deploy ECK 1.0.0-rc4

Apply the following manifest:

```yaml
apiVersion: elasticsearch.k8s.elastic.co/v1beta1
kind: Elasticsearch
metadata:
  name: es-apm-sample-invalid
spec:
  version: 7.4.0
  nodeSets:
    - name: default
      count: 3
      config:
        # This setting could have performance implications for production clusters.
        # See: https://www.elastic.co/guide/en/cloud-on-k8s/master/k8s-virtual-memory.html
        node.store.allow_mmap: false
        node.master: false
```

The apply should fail, because the webhook detects no master node, but it does not.
Changing `apiVersion: elasticsearch.k8s.elastic.co/v1beta1` to `apiVersion: elasticsearch.k8s.elastic.co/v1` does trigger the webhook correctly.

sebgl:
Looks like the v1beta1 endpoint is not setup correctly in the operator:

```
[elastic@elastic-operator-0 ~]$ curl -k https://localhost:9443/validate-elasticsearch-k8s-elastic-co-v1beta1-elasticsearch                                                                                │
404 page not found                                                                                                                                                                                        │
[elastic@elastic-operator-0 ~]$ curl -k https://localhost:9443/validate-elasticsearch-k8s-elastic-co-v1-elasticsearch                                                                                     │
{"response":{"uid":"","allowed":false,"status":{"metadata":{},"message":"contentType=, expected application/json","code":400}}}
```

PR #2358, opened 2020-01-07
Fix v1beta1 webhook

sebgl:
The v1 webhook did work as expected, but not the v1beta1 webhook before
this commit.

To fix this, it:

* creates 2 distinct webhooks with distinct names
(`elastic-es-validation-v1beta1.k8s.elastic.co` and
`elastic-es-validation-v1.k8s.elastic.co`) in the
ValidatingWebhookConfiguration
* ensures the v1beta1 webhook is created along with the v1 webhook at
operator startup
* registers v1beta1 resources in the schema, required for the v1beta1
webhook to run

To test it, I deployed the operator and webhook configurations, then:
- tried applying a v1beta1 resource: the v1beta1 webhook was triggered,
returning an error mentioning v1beta1.NodeSet
- tried applying a v1 resource: the v1 webhook was triggered, returning
an error mentioning v1.NodeSet

Fixes https://github.com/elastic/cloud-on-k8s/issues/2355.

### DIFF
--- a/docs/webhook.asciidoc
+++ b/docs/webhook.asciidoc
@@ -7,7 +7,7 @@ A validating webhook provides additional validation of Elasticsearch resources: 
 === Architecture
 The webhook is composed of 4 main components. Here is a brief description of each of them to understand how they interact, their naming, and how they are managed.
 
-. A `ValidatingWebhookConfiguration` object that defines the validating webhook, named `elastic-es-validation-{eck_crd_version}.k8s.elastic.co`. It must be created before starting the operator. The `caBundle` field can be automatically managed as part of the automatic certificate management _(see below)_.
+. A `ValidatingWebhookConfiguration` object that defines the validating webhook, targeting the right webhook path and resource. It must be created before starting the operator. The `caBundle` field can be automatically managed as part of the automatic certificate management _(see below)_.
 . A Kubernetes Service is used to expose the validating server, named `elastic-webhook-server`. It is in the same Namespace as the webhook server.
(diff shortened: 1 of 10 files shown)


docs/webhook.asciidoc
```diff
@@ -7,7 +7,7 @@ A validating webhook provides additional validation of Elasticsearch resources:
 === Architecture
 The webhook is composed of 4 main components. Here is a brief description of each of them to understand how they interact, their naming, and how they are managed.
 
-. A `ValidatingWebhookConfiguration` object that defines the validating webhook, named `elastic-es-validation.k8s.elastic.co`. It must be created before starting the operator. The `caBundle` field can be automatically managed as part of the automatic certificate management _(see below)_.
+. A `ValidatingWebhookConfiguration` object that defines the validating webhook, named `elastic-es-validation-{eck_crd_version}.k8s.elastic.co`. It must be created before starting the operator. The `caBundle` field can be automatically managed as part of the automatic certificate management _(see below)_.
```
barkbay: The `ValidatingWebhookConfiguration` object is named `elastic-webhook.k8s.elastic.co`
It is the `clientConfig` that is named `elastic-es-validation-{eck_crd_version}.k8s.elastic.co`

sebgl: The `ValidatingWebhookConfiguration` contains a `webhooks` array of `Webhook`, for which we define the name. That's not the name of the `clientConfig`.

I think in "A `ValidatingWebhookConfiguration` object that defines the validating webhook, named `elastic-es-validation-{eck_crd_version}.k8s.elastic.co`",  `named` refers to the inner validating webhook, not to the ValidatingWebhookConfiguration.

WDYT?

sebgl: We can maybe simplify that sentence to `A ValidatingWebhookConfiguration` object that defines the validating webhooks`(without any name reference) so we avoid the confusion?

Issue #2318, opened 2019-12-20
E2E test Cluster_UUID_should_have_been_preserved is flaky


sebgl:
This happens fairly often, for any mutation where we check the ClusterUUID when done:

```
    --- FAIL: TestMutationMdiToDedicated/Cluster_UUID_should_have_been_preserved (64.35s)
        cluster_uuid.go:58: 
            	Error Trace:	cluster_uuid.go:58
            	Error:      	Received unexpected error:
            	            	Get https://test-mutation-mdi-to-dedicated-9j57-es-http.e2e-hbq8i-mercury.svc:9200/: dial tcp 10.91.240.2:9200: connect: no route to host
            	Test:       	TestMutationMdiToDedicated/Cluster_UUID_should_have_been_preserved
```

We do check the ClusterUUID in other places of the code, but we wrap the code in a `test.Eventually`.
Here we don't, because in theory if the mutation is over we should be able to reach the cluster.
`dial tcp 10.91.240.2:9200: connect: no route to host` looks like we're trying to request a Pod which doesn't exist anymore (was removed/replaced during the mutation).

We also get some errors like `dial tcp 10.112.0.26:9200: connect: connection refused `.

I was hoping https://github.com/elastic/cloud-on-k8s/pull/2233 would fix it but it looks like it's not enough.

sebgl:
Looking at the logs of [this test run](https://devops-ci.elastic.co/view/cloud-on-k8s/job/cloud-on-k8s-stack/130/testReport/junit/github/com_elastic_cloud-on-k8s_test_e2e_es/Run_tests_for_different_ELK_stack_versions_in_GKE___7_3_2___TestMutationHTTPSToHTTP_Cluster_UUID_should_have_been_preserved/) where `dial tcp 10.112.0.26:9200: connect: connection refused ` happens: `10.112.0.26` is not any of the mutated cluster Pod IPs.

It seems to match the IP address of node `test-mutation-https-to-http-jdcb-es-masterdata-1` before it was upgraded according to the logs of the Elasticsearch node 0:

```
[...]

PR #2360, opened 2020-01-07
Wait 30sec in the preStop hook to account for kube-proxy refresh

sebgl:
Kube-proxy defaults to a 30sec iptables-sync-period (overridden to 10sec
on GKE).

For that reason, when an Elasticsearch Pod is gracefully shutdown, some
clients in the k8s cluster might still target the Pod IP, because the
local kube-proxy iptables (or ipvs) rules are not updated yet.

Let's increase our initial 1 sec additional wait time to 30 sec so we
account for those changes. It has the side-effect of slowing down
Pods deletion (hence rolling upgrades) by 30sec per Pod, but should help
keeping Elasticsearch available during rolling upgrades.

This commit also increases the Elasticsearch container gracePeriod to
account for the increased wait time. I changed it from 120 to 180
seconds (technically could have been 150, but 180 rounds up to 3
minutes).

And renames `ADDITIONAL_WAIT_TIME` & `MAX_WAIT_TIME` to
`PRE_STOP_ADDITIONAL_WAIT_TIME` & `PRE_STOP_MAX_WAIT_TIME`
so the names are more explicit when overridden through env vars.

Fixes https://github.com/elastic/cloud-on-k8s/issues/2318.


pkg/controller/elasticsearch/nodespec/lifecycle_hook.go
```diff
@@ -29,14 +29,13 @@ set -eux
 # contains IP that is already inactive. Assumes $HEADLESS_SERVICE_NAME and $POD_IP env variables are defined.
 
 # max time to wait for pods IP to disappear from DNS. As this runs in parallel to grace period
-# (defaulting to 30s) after which process is SIGKILLed, it should be set to allow enough time
```
sebgl: We did already override the grace period to 120 sec (now 180), I think the 30sec mention here was a bit misleading.

docs/elasticsearch-specification.asciidoc
```diff
@@ -585,7 +585,7 @@ Note that this will require restarting the pods.
...
+To address this issue and guarantee no downtime for new connections, ECK relies on a link:https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/[PreStop lifecycle hook]. This waits to terminate the Elasticsearch process until the `Service` DNS record does not contain the IP of the `Pod`. First, the PreStop lifecycle hook will keep querying DNS for `PRE_STOP_MAX_WAIT_SECONDS` (defaulting to 20) and then it will wait for `PRE_STOP_ADDITIONAL_WAIT_SECONDS` (defaulting to 30). Additional wait is used to give clients time to use an IP resolved just before DNS record was updated. The exact behavior is configurable using environment variables, for example:
```
david-kow:
```suggestion
To address this issue and guarantee no downtime for new connections, ECK relies on a link:https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/[PreStop lifecycle hook]. This waits to terminate the Elasticsearch process until the `Service` DNS record does not contain the IP of the `Pod`. First, the PreStop lifecycle hook will keep querying DNS for `PRE_STOP_MAX_WAIT_SECONDS` (defaulting to 20) and then it will wait for `PRE_STOP_ADDITIONAL_WAIT_SECONDS` (defaulting to 30). Additional wait is used to:
1. give clients time to use an IP resolved just before DNS record was updated, and
1. give kube-proxy time to refresh iptables rules on all nodes.
The exact behavior is configurable using environment variables, for example:
```

Also, I think we should mention somehow that additional wait should be at least min-sync-interval, so someone running with higher knows to adjust this and someone running with lower (for example, on GKE) can lower it.

pkg/controller/elasticsearch/nodespec/lifecycle_hook.go
```diff
@@ -23,31 +23,30 @@ const PreStopHookScript = `#!/usr/bin/env bash
...
-MAX_WAIT_SECONDS=${MAX_WAIT_SECONDS:=20}
+# after which process is SIGKILLed, it should be set to allow enough time for the process to gracefully terminate.
+PRE_STOP_MAX_WAIT_SECONDS=${PRE_STOP_MAX_WAIT_SECONDS:=20}
 
-# additional wait, allows queries to successfully use IP from DNS from before pod termination
-# this gives a little bit more time for clients that resolved DNS just before DNS record
-# was updated.
-ADDITIONAL_WAIT_SECONDS=${ADDITIONAL_WAIT_SECONDS:=1}
+# additional wait before shutting down Elasticsearch.
+# It allows kube-proxy to refresh its routes (defaults to every 30 seconds) and remove that Pod IP once Terminating.
```
david-kow: I'd add this to docs (as suggested above), also:
```suggestion
# It allows kube-proxy to refresh its rules (defaults to every 30 seconds) and remove that Pod IP once Terminating.
```

pkg/controller/elasticsearch/nodespec/lifecycle_hook.go
```diff
@@ -23,31 +23,30 @@ const PreStopHookScript = `#!/usr/bin/env bash
...
+# Also gives some additional bonus time to in-flight requests to terminate.
```
david-kow: Not a strong opinion on this one, so I'm fine if you leave it like that, but I think it's a little bit imprecise, as this is not only about requests in-flight, but also new requests using IP from earlier DNS resolution.

docs/elasticsearch-specification.asciidoc
barkbay: > 1. give kube-proxy time to refresh iptables rules on all nodes

Correct me if I'm wrong but should it be "give kube-proxy time **to start refreshing** iptables rules on all nodes" ?
I'm not sure we have any guarantee about when iptables rules will be actually refreshed. IIRC until `ipvs` is more widely used [it can take some time to apply the rules](https://youtu.be/4-pawkiazEg?t=862).

david-kow: The intention here is to allow refresh to start and finish (that's what we are waiting for), ie. we don't care about only starting the refresh. Having said that, you are right that the refresh time is unbounded. Maybe we should say instead that additional time is supposed to be larger than min-sync + the rules refresh?

sebgl: That's correct.
@barkbay @david-kow do you think we should set a higher threshold (more than 30sec) to account for a "slow" iptables refresh?  I could not find any numbers about iptables refresh performance under reasonably normal conditions (ie. not 5k services).

barkbay: I must confess that having fighting some iptables/SDN/DNS problems in production I'm not comfortable with the term "guarantee no downtime for new connections". We are adding some tools and parameters to minimize downtime, which is definitely great, and I'm 👍 with the changes here. My point was more that given the complexity of some deployments (sdn providers, dns caches, ipvs VS iptables) I would not give such guarantee. 
I understand it is a personal feeling though :)

david-kow: > @barkbay @david-kow do you think we should set a higher threshold (more than 30sec) to account for a "slow" iptables refresh?

I think we shouldn't, because:
1) we've are already built in some worst case assumptions (worst possible timing, burst depleted)
2) there is already some "hidden", added time as refresh is on the path happening in parallel to prestop. So when we run prestop, wait for DNS update, es is shutting down, etc. the clock is ticking for iptables refresh.
3) I don't know good value to set it for as it can vary from ~0s to hours as @barkbay noted.

I think I'd document it and assume that if someone has a cluster where routes need 10 mins to refresh for some reason, they can adjust the settings themselves.

pkg/controller/elasticsearch/nodespec/lifecycle_hook.go
sebgl: 👍

docs/elasticsearch-specification.asciidoc
david-kow: > I must confess that having fighting some iptables/SDN/DNS problems in production I'm not comfortable with the term "guarantee no downtime for new connections".

The message I tried to convey here is that orchestration done by ECK is not introducing anything that would cause ES downtime (which was the case before). That guarantee doesn't cover parts we are not responsible for (network, other hw, power, etc.), but I'm not sure how to word it differently.

pebrc: > but I'm not sure how to word it differently.

"minimise downtime"?

david-kow: > "minimise downtime"?

At least to me, it implicates the inevitable existence of downtime for all upgrades. But I don't feel that strongly about it :)

sebgl: I agree with all the above 👍 
I kept the 30sec but changed comments and docs to be more explicit about what we're trying to achieve here: minimise unavailability by waiting long enough for a "default" kube-proxy to refresh its rules, which may still take longer in some situations.

PR #2371, opened 2020-01-08
Report max ERUs in the licensing info configmap

thbkrkr:
This commit includes the `max_enterprise_resource_units` in the licensing info config map to make it easier for users to see whether they are in compliance with their license.

For old style enterprise orchestration licenses which only have `max_instances`, the `max_enterprise_resource_units` by dividing `max_instances` by 2.

For trial enterprise licenses, the field is omitted.

### DIFF
--- a/pkg/license/license.go
+++ b/pkg/license/license.go
@@ -65,7 +65,7 @@ func (r LicensingResolver) ToInfo(totalMemory resource.Quantity) (LicensingInfo,
 	}
 
 	// include the max ERUs only for a non trial license
-	if !operatorLicense.IsTrial() {
+	if operatorLicense != nil && !operatorLicense.IsTrial() {
 		licensingInfo.MaxEnterpriseResourceUnits = strconv.Itoa(maxERUs)
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/license/license.go
```diff
@@ -45,19 +47,29 @@ type LicensingResolver struct {
...
 		Timestamp:               time.Now().Format(time.RFC3339),
 		EckLicenseLevel:         licenseLevel,
 		TotalManagedMemory:      memoryInGB,
-		EnterpriseResourceUnits: eru,
-	}, nil
+		EnterpriseResourceUnits: ERUs,
+	}
+
+	// include the max ERUs only for a non trial license
+	if !operatorLicense.IsTrial() {
```
pebrc: Operator license can be nil

pebrc: or `maxERUs > 0`

thbkrkr: I opted to check the nullity to keep the `IsTrial()` call.

anyasabo:
Do we want to update the example in the `docs/licensing.asciidoc` with the extra data as well?

PR #2375, opened 2020-01-08
Update licensing doc with max_enterprise_resource_units

thbkrkr:
Add the extra data `max_enterprise_resource_units` in the licensing documentation.

### DIFF
--- a/docs/licensing.asciidoc
+++ b/docs/licensing.asciidoc
@@ -77,6 +77,7 @@ The operator periodically writes the total amount of Elastic resources under man
 {
   "eck_license_level": "basic",
   "enterprise_resource_units": "1",
+  "max_enterprise_resource_units": "10",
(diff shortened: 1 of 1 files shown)


docs/licensing.asciidoc
```diff
@@ -77,6 +77,7 @@ The operator periodically writes the total amount of Elastic resources under man
 {
   "eck_license_level": "basic",
   "enterprise_resource_units": "1",
+  "max_enterprise_resource_units": "10",
```
anyasabo: Is this added for basic licenses? I thought from https://github.com/elastic/cloud-on-k8s/pull/2371/files#diff-fcbf74a837a779a0d548ba5b3b57ef5aR67
that it was only for legit enterprise licenses, not basic or enterprise trial.

thbkrkr: Good point. I forgot to think about basic licenses.

PR #2383, opened 2020-01-09
Add note that version field is required even when custom image is provided

david-kow:
Users might think that version is only used to pick an image, while we do need it, for example, to identify how to talk to ES.


sebgl:
We should actually make it mandatory in the OpenAPI spec (remove the `omitempty` tag).

But that breaks the CRD generation (`v1` not exactly the same as `v1beta1` - needs https://github.com/kubernetes-sigs/controller-tools/pull/350) so I would say let's maybe wait for after the release to do that?

PR #1594, opened 2019-08-19
Use deployer in operators/Makefile dev targets

david-kow:
This moves operators/Makefile targets to use deployer instead of gke-cluster.sh, which can be now deleted.

Deployer now uses cobra for cli as the interface got bigger.

PR #2386, opened 2020-01-09
Rework Makefile targets to not be cloud specific

david-kow:
While deployer supports targeting both GKE and AKS our Makefile targets are specific to GKE, which (after modifying deployer-config.yml manually) forced us to run `bootstrap-gke` to bootstrap AKS cluster. Now the targets where reworked similarly to what was mentioned in https://github.com/elastic/cloud-on-k8s/pull/1594#discussion_r316018647:
- each provider will have it's own file (`hack/deployer/config/deployer-config-PROVIDER.yml`)
- `make switch-gke` and `make switch-aks` can be used to switch between them
- `make bootstrap-cloud` and `make delete-cloud` can be used to create and delete a cluster in currently chosen provider
- deployer will supply a sane default config file for each provider if the file is not already present

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,7 @@ SNAPSHOT   	?= true
 LOG_VERBOSITY ?= 1
 
 # Get the currently used golang install path (in GOPATH/bin, unless GOBIN is set)
-ifeq (,$(shell go env GOBIN))
+ifeq ($(shell go env GOBIN),)
 GOBIN=$(shell go env GOPATH)/bin
 else
 GOBIN=$(shell go env GOBIN)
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -262,41 +262,49 @@ bootstrap-minikube:
...
+DEPLOYER=./hack/deployer/deployer --plans-file=hack/deployer/config/plans.yml --config-file=hack/deployer/config/deployer-config-$(PROVIDER).yml
 
 build-deployer:
 	@ go build -mod=readonly -o ./hack/deployer/deployer ./hack/deployer/main.go
 
-setup-deployer-for-gke-once: require-gcloud-project build-deployer
-ifeq (,$(wildcard hack/deployer/config/deployer-config.yml))
-	@ ./hack/deployer/deployer create defaultConfig --path=hack/deployer/config/deployer-config.yml
+create-default-config:
+ifeq (,$(wildcard hack/deployer/config/deployer-config-$(PROVIDER).yml))
```
pebrc: Nit? we seem to be a bit inconsistent with the order of `ifeq` args here and elsewhere

david-kow:
jenkins test this please (CLA check not starting)

anyasabo:
jenkins test this please

david-kow:
jenkins test this please

PR #2388, opened 2020-01-09
Use a 15min RollingUpgradeTimeout for keystore checks in E2E tests

sebgl:
Since we added a 30sec preStop wait, rolling upgrades take longer than
before. We recently updated the rolling upgrade timeout to 15 minutes,
but did not do it for the keystore rolling upgrade test which is written
differently.

### DIFF
--- a/test/e2e/test/elasticsearch/checks_k8s.go
+++ b/test/e2e/test/elasticsearch/checks_k8s.go
@@ -26,7 +26,7 @@ import (
 const (
 	// RollingUpgradeTimeout is used for checking a rolling upgrade is complete.
 	// Most tests require less than 5 minutes for all Pods to be running and ready,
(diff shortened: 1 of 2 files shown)


test/e2e/test/elasticsearch/checks_k8s.go
```diff
@@ -23,6 +23,15 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 )
 
+const (
+	// RollingUpgradeTimeout is used for checking a rolling upgrade is complete.
+	// Most tests require less than 5 minutes for all Pods to be running and ready,
+	// but it occasionally takes longer for various reasons (long pod Pod creation time, long volume binding, etc.).
```
thbkrkr: Nit: `s|pod Pod|Pod|`

Issue #2393, opened 2020-01-09
Rolling upgrade order seems inconsistent


sebgl:
I think we have a bug with the order in which we do rolling upgrades.

To reproduce:
* Deploy a 6.8 cluster with 5 data nods and 3 master nodes
* Upgrade it to 7.1

According to the comment in the `sortCandidates ` function, we want to upgrade master nodes last:

```
// sortCandidates is the default sort function, masters have lower priority as
// we want to update the data nodes first.
```

In my environment, I observed:
- master-2 being upgraded first
- then data-4, data-3
- then master-1
- then data-2, data-1, data-0
- then master-0

barkbay:
An other option would be that the list of Pods to be updated only contain the masters during the first reconciliation: https://gist.github.com/barkbay/d5ccd3ae9cc3bf649678b16ecfb6e83f

sebgl:
@barkbay that could make sense, when retrieving the pods to upgrade we check the StatefulSet updateRevision: 
```
if statefulSet.Status.UpdateRevision == "" {
			// no upgrade scheduled
			continue
		}
```
We may be in a situation where the data sset does not have its update revision set yet, but the masters one does.

That would explain processing `master-2` first in the first post.

barkbay:
Re `master-2` : 
I added some logs:

```
{"level":"info","@timestamp":"2020-01-10T09:26:42.309Z","logger":"generic-reconciler","message":"Updating resource","ver":"1.0.0-021fc6ca","kind":"Secret","namespace":"elasticsearch-namespace","name":"es-apm-sample-es-masters-es-config"}
{"level":"info","@timestamp":"2020-01-10T09:26:42.318Z","logger":"statefulset","message":"ReconcileStatefulSet","ver":"1.0.0-021fc6ca","name":"es-apm-sample-es-masters"}
{"level":"info","@timestamp":"2020-01-10T09:26:42.318Z","logger":"generic-reconciler","message":"Updating resource","ver":"1.0.0-021fc6ca","kind":"StatefulSet","namespace":"elasticsearch-namespace","name":"es-apm-sample-es-masters"}
{"level":"info","@timestamp":"2020-01-10T09:26:42.331Z","logger":"generic-reconciler","message":"Updating resource","ver":"1.0.0-021fc6ca","kind":"Secret","namespace":"elasticsearch-namespace","name":"es-apm-sample-es-nodes-es-config"}
[...]

PR #2394, opened 2020-01-09
Update deletion order sort test

anyasabo:
Edited*

Related https://github.com/elastic/cloud-on-k8s/issues/2393

Updated test cases and changed the names to make sure we are not falling back on the name sort. Some slight sort changes to clarify intent but logic is the same.

### DIFF
--- a/pkg/controller/elasticsearch/driver/upgrade_predicates_test.go
+++ b/pkg/controller/elasticsearch/driver/upgrade_predicates_test.go
@@ -15,6 +15,7 @@ import (
 	"github.com/elastic/cloud-on-k8s/pkg/controller/elasticsearch/reconcile"
 	"github.com/elastic/cloud-on-k8s/pkg/utils/k8s"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 // These tests are focused on "type changes", i.e. when the type of a nodeSet is changed.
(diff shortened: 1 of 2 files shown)


pkg/controller/elasticsearch/driver/upgrade_predicates_test.go
```diff
@@ -675,48 +676,50 @@ func TestDeletionStrategy_SortFunction(t *testing.T) {
...
-					newTestPod("data-0").needsUpgrade(true),
-					newTestPod("masters-1").needsUpgrade(true),
-					newTestPod("data-1").needsUpgrade(true),
-					newTestPod("masters-2").needsUpgrade(true),
+					// use "amasters" rather than "masters" to ensure we are not relying on the name sort accidentally
+					newTestPod("amasters-0").isMaster(true).needsUpgrade(true),
+					newTestPod("data-0").isData(true).needsUpgrade(true),
+					newTestPod("amasters-1").isMaster(true).needsUpgrade(true),
+					newTestPod("data-1").isData(true).needsUpgrade(true),
+					newTestPod("amasters-2").isMaster(true).needsUpgrade(true),
```
david-kow: Could we add a pod here from another master sset with `masters-0` name? Just to be sure we cover both cases.

anyasabo:
I got a little myopic. Both @barkbay and @sebgl are correct, I had an issue on my end that made it look like it was failing with the past version. Updated PR accordingly.

Issue #2397, opened 2020-01-10
Cluster UUID changed when upgrading single master 6.8.x to new nodeSet 7.3.x


sebgl:
We got this in E2E tests: https://devops-ci.elastic.co/job/cloud-on-k8s-stack/148/testReport/github/com_elastic_cloud-on-k8s_test_e2e_es/Run_tests_for_different_ELK_stack_versions_in_GKE___7_4_2___TestVersionUpgradeSingleMaster68xToNewNodeSet73x_Cluster_UUID_should_have_been_preserved/

```
=== RUN   TestVersionUpgradeSingleMaster68xToNewNodeSet73x/Cluster_UUID_should_have_been_preserved
{"level":"error","@timestamp":"2020-01-10T02:42:11.083Z","message":"stopping early","ver":"0.0.0-00000000","error":"test failure","stacktrace":"github.com/go-logr/zapr.(*zapLogger).Error\n\t/go/pkg/mod/github.com/go-logr/zapr@v0.1.0/zapr.go:128\ngithub.com/elastic/cloud-on-k8s/test/e2e/test.StepList.RunSequential\n\t/go/src/github.com/elastic/cloud-on-k8s/test/e2e/test/step.go:42\ngithub.com/elastic/cloud-on-k8s/test/e2e/test.RunMutations\n\t/go/src/github.com/elastic/cloud-on-k8s/test/e2e/test/run_mutation.go:37\ngithub.com/elastic/cloud-on-k8s/test/e2e/test.RunMutation\n\t/go/src/github.com/elastic/cloud-on-k8s/test/e2e/test/run_mutation.go:78\ngithub.com/elastic/cloud-on-k8s/test/e2e/es.RunESMutation\n\t/go/src/github.com/elastic/cloud-on-k8s/test/e2e/es/mutation_test.go:299\ngithub.com/elastic/cloud-on-k8s/test/e2e/es.TestVersionUpgradeSingleMaster68xToNewNodeSet73x\n\t/go/src/github.com/elastic/cloud-on-k8s/test/e2e/es/version_upgrade_test.go:73\ntesting.tRunner\n\t/usr/local/go/src/testing/testing.go:909"}
    --- FAIL: TestVersionUpgradeSingleMaster68xToNewNodeSet73x/Cluster_UUID_should_have_been_preserved (0.03s)
        cluster_uuid.go:60: 
            	Error Trace:	cluster_uuid.go:60
            	Error:      	Not equal: 
[...]

PR #2399, opened 2020-01-10
Synchronously request the Cluster UUID

sebgl:
Since we only perform that request when:
- the cluster is not marked as bootstrapped yet
- es is supposed to be reachable

It's fine to do it synchronously, and not rely on state observed every
10 seconds asynchronously.

Doing a sync requests reduces a bit the time window where we could wrongly
assume the cluster has not been bootstrapped yet, and removes some
complexity in the code.

This commit also refactors a bit the unit tests with a single function
that should cover all cases.

And removes the unused `ObservedState.ClusterInfo`.

Relates https://github.com/elastic/cloud-on-k8s/issues/2397.

### DIFF
--- a/pkg/controller/elasticsearch/bootstrap/bootstrap.go
+++ b/pkg/controller/elasticsearch/bootstrap/bootstrap.go
@@ -6,8 +6,11 @@ package bootstrap
 
 import (
 	esv1 "github.com/elastic/cloud-on-k8s/pkg/apis/elasticsearch/v1"
-	"github.com/elastic/cloud-on-k8s/pkg/controller/elasticsearch/observer"
+	"github.com/elastic/cloud-on-k8s/pkg/controller/elasticsearch/client"
 	"github.com/elastic/cloud-on-k8s/pkg/utils/k8s"
+
+	"context"
+
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 )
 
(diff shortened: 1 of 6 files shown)


pkg/controller/elasticsearch/driver/driver.go
```diff
@@ -166,9 +166,7 @@ func (d *defaultDriver) Reconcile() *reconciler.Results {
 		))
 
 	// always update the elasticsearch state bits
-	if observedState.ClusterInfo != nil && observedState.ClusterHealth != nil {
-		d.ReconcileState.UpdateElasticsearchState(*resourcesState, observedState)
-	}
+	d.ReconcileState.UpdateElasticsearchState(*resourcesState, observedState)
```
pebrc: Can we remove the same call on L246 then?

pkg/controller/elasticsearch/bootstrap/bootstrap.go
```diff
@@ -6,8 +6,11 @@ package bootstrap
 
 import (
 	esv1 "github.com/elastic/cloud-on-k8s/pkg/apis/elasticsearch/v1"
-	"github.com/elastic/cloud-on-k8s/pkg/controller/elasticsearch/observer"
+	"github.com/elastic/cloud-on-k8s/pkg/controller/elasticsearch/client"
 	"github.com/elastic/cloud-on-k8s/pkg/utils/k8s"
+
+	"context"
+
```
barkbay: nit: spaces in imports

pkg/controller/elasticsearch/driver/driver.go
sebgl: Good question. I'm looking into it.

sebgl: I think we have a spaghetti plate to handle here. We seem to update the ES status at multiple places (sometimes deep down in nodes reconciliation), multiple times.
I think it's better to extract it to its own issue & PR.

pebrc: Works for me.

sebgl: https://github.com/elastic/cloud-on-k8s/issues/2403

pkg/controller/elasticsearch/bootstrap/bootstrap.go
sebgl: This is what IntelliJ produces when clicking on "Optimize imports". I tried sticking all imported lines together then reapplying the optimization. This is what I get:

```go
import (
	"context"

	logf "sigs.k8s.io/controller-runtime/pkg/log"

	"github.com/elastic/cloud-on-k8s/pkg/controller/elasticsearch/client"
)

```

PR #2410, opened 2020-01-11
Editing "Update strategy" section

alaudazzi:
Suggested some rewording.

### DIFF
--- a/docs/elasticsearch-specification.asciidoc
+++ b/docs/elasticsearch-specification.asciidoc
@@ -447,7 +447,10 @@ The init container inherits the image of the main container image if one is not 
 [id="{p}-update-strategy"]
 === Update strategy
 
-Sometimes the operator must take a Pod down, for example, to restart a node and apply a new configuration value. Other times the operator must spin up pods above what is currently in the specification, say to migrate to a new node set. You can limit the number of these simultaneous changes  using an `updateStrategy` as follows:
+You can use the `updateStrategy` specification to limit the number of simultaneous changes, like for example in the following cases:
+
+* The operator takes a Pod down to restart a node and applies a new configuration value.
+* The operator must spin up Pods above what is currently in the specification to migrate to a new node set.
 
 [source,yaml]
 ----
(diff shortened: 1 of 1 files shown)


docs/elasticsearch-specification.asciidoc
```diff
@@ -457,20 +460,20 @@ spec:
       maxSurge: 3
       maxUnavailable: 1
 ----
-`maxSurge` refers to the number of Pods that can be scheduled above the total number of Pods in the currently applied specification. This setting might be useful when the new specification has similar number of Pods, but different node sets (for example, when migrating Pods to a new node set). It will prevent the operator from creating all new Pods before removing the old ones thus preventing having (temporarily) as many as two times more pods than the specification defines.
+`maxSurge`: Refers to the number of Pods that can be scheduled out of the total number of Pods in the currently applied specification. This setting is useful when the new specification has similar number of Pods, but different node sets. For example, when migrating Pods to a new node set. This setting prevents the operator from creating all new Pods before removing the old ones, by temporarily avoiding two times more Pods than the ones defined in the specification.
```
charith-elastic: The original text was a bit confusing so the meaning got lost in the re-write. I would try to explain `MaxSurge` as follows.  
```suggestion
`maxSurge`: Refers to the number of extra Pods that can be temporarily scheduled exceeding the number of Pods  defined in the specification. This setting is useful for controlling the resource usage of the Kubernetes cluster when new Pods need to be spun up to replace existing Pods. For example, when applying a configuration change to a Node Set, ECK creates a clone of the Node Set with the desired configuration and shuts down the old Pods. Without the `MaxSurge` restriction, the cluster would temporarily contain twice as many Pods as usual while this operation is in progress.
```

docs/elasticsearch-specification.asciidoc
```diff
@@ -457,20 +460,20 @@ spec:
...
+`maxUnavailable`: Refers to the number of Pods that can be unavailable out of the total number of Pods in the currently applied specification. A Pod is defined unavailable when it is not ready from a Kubernetes perspective.
 
-As these constraints are enforced immediately after the specification is applied, it is possible the the actual state of the cluster might violate these constraints. When the cluster is outside these constraints, the operator can add or remove nodes only with the purpose of getting closer to these constraints. When the cluster is within these constraints, the operator can either add or remove nodes, as long as the cluster remains within the predefined constraints.
+As these constraints are enforced immediately after the specification is applied, it is possible that the actual state of the cluster might violate these constraints. When the cluster is outside these constraints, the operator can add or remove nodes only with the purpose of getting closer to these constraints. When the cluster is within these constraints, the operator can either add or remove nodes, as long as the cluster remains within the predefined constraints.
```
charith-elastic:
```suggestion
The operator only tries to apply these constraints when a new specification is being applied. It is possible that the cluster state does not conform to the constraints at the beginning of the operation due to external factors. The operator will attempt to get to the desired state by adding or removing Pods as necessary while ensuring that the constraints are still satisfied.
```

docs/elasticsearch-specification.asciidoc
```diff
@@ -484,13 +487,13 @@ spec:
       maxUnavailable: 1
 ----
 
-`maxSurge` is unbounded: this means that all the missing Pods are immediately created according to the best practices at any given time.
-`maxUnavailable` defaults to `1`: this assures that clusters have at most one unavailable Pod at any given time.
+`maxSurge` is unbounded: This means that all the missing Pods are immediately created according to the best practices at any given time.
```
charith-elastic:
```suggestion
`maxSurge` is unbounded: This means that all the required Pods are created immediately.
```

docs/elasticsearch-specification.asciidoc
```diff
@@ -484,13 +487,13 @@ spec:
...
+`maxUnavailable` defaults to `1`: This assures that clusters have at most one unavailable Pod at any given time.
```
charith-elastic:
```suggestion
`maxUnavailable` defaults to `1`: This ensures that the cluster has no more than one unavailable Pod at any given point in time.
```

docs/elasticsearch-specification.asciidoc
```diff
@@ -484,13 +487,13 @@ spec:
...
-* Due to the safety measures employed by the operator, a particular changeBudget might be insufficient to perform the requested update. For example, with `maxSurge` set to 0 you cannot remove the last data node from one nodeSet and add a data node to a different nodeSet. In this case the operator cannot remove the old node and migrate data as other nodes are not available, and can't add the new node because `maxSurge` is set to 0.
-* Currently, the operator has a limited understanding of the cluster topology and during the update might not pick the optimal order of Pods being added or removed. This can block the update process and you might need to specify a higher `maxSurge` than theoretically needed.
+* Due to the safety measures employed by the operator, a particular `changeBudget` might be insufficient to perform the requested update. For example, with `maxSurge` set to 0 you cannot remove the last data node from one `nodeSet` and add a data node to a different `nodeSet`. In this case, the operator cannot remove the old node and migrate data as other nodes are not available, and can't add the new node because `maxSurge` is set to 0.
```
charith-elastic:
```suggestion
* Due to the safety measures employed by the operator, certain `changeBudget`s might prevent the operator from making any progress . For example, with `maxSurge` set to 0, you cannot remove the last data node from one `nodeSet` and add a data node to a different `nodeSet`. In this case, the operator cannot create the new node because `maxSurge` is 0, and it cannot remove the old node because there are no other data nodes to migrate the data to.
```

docs/elasticsearch-specification.asciidoc
```diff
@@ -484,13 +487,13 @@ spec:
...
+* Currently, the operator has a limited understanding of the cluster topology and during the update it might not pick the optimal order of Pods being added or removed. This can block the update process and you need to specify a higher `maxSurge` than theoretically needed.
```
charith-elastic:
```suggestion
* For certain complex configurations, the operator might not be able to deduce the optimal order of operations necessary to achieve the desired outcome. If progress is blocked,  you may need to update the `maxSurge` setting to a higher value than the theoretical best to help the operator make progress in that case.
```

charith-elastic:
Also, perhaps this PR should go into `master` first and then be backported to `1.0`? The other way around should be fine too. Happy to help if needed.

alaudazzi:
Thank you for your feedback @charith-elastic!

PR #2411, opened 2020-01-12
Ensure that the list of nodes scheduled for upgrade is consistent in case of version change

barkbay:
This PR fixes an issue where a single dedicated master can be restarted while remaining data nodes are still running an old version of Elasticsearch.

It could be the case if there are:
* Only 1 master node in a nodeSet
* *n* data nodes in an other nodeSet

In the above situation, if the nodeSet that contains the data nodes is not selected for an upgrade *(as it is the case in #2393)* then the master node will be upgraded while the data nodes are still running the previous version.

This is possible because the predicates are relying on a consistent list of the Pod that needs to be upgraded.

Note that this PR does not completely solve #2393 but mitigates the worst case scenario _(i.e. having all Elasticsearch masters running a newer version than the data nodes)_

I have decided to implement here what I think is the most simple fix because:
* I didn't want to introduce an important change in the code base while we are very close from a release.
* [podsToUpgrade](https://github.com/elastic/cloud-on-k8s/blob/c0155249e999a0fd52d3bc76351dd526f3e0c7f8/pkg/controller/elasticsearch/driver/upgrade.go#L171) is used at several places in the code, if we add some code to detect an inconsistency we need to figure out where it must be used and how other uses are impacted. It is ofc not impossible, just would require more work.
* This change will  trigger a new reconciliation anyway because of the use of [podsToUpgrade](https://github.com/elastic/cloud-on-k8s/blob/c0155249e999a0fd52d3bc76351dd526f3e0c7f8/pkg/controller/elasticsearch/driver/upgrade.go#L171) in [driver.Reconciled](https://github.com/elastic/cloud-on-k8s/blob/master/pkg/controller/elasticsearch/driver/nodes.go#L166)

*That said I'm happy to close this PR if a best solution is found in a reasonable timeframe :)*

### DIFF
--- a/pkg/controller/elasticsearch/driver/upgrade.go
+++ b/pkg/controller/elasticsearch/driver/upgrade.go
@@ -203,8 +203,10 @@ func podsToUpgrade(
 				continue
 			}
 			// We consider that a Pod needs an update if at least one the following condition is met:
-			// 1. The update revision of the Pod does not match the one of the StatefulSet
+			// 1. The update revision of the Pod does not match the one in the status of the StatefulSet
 			// 2. The Elasticsearch version run by the Pod does not match the expected one in the Elasticsearch object
+			// Using the Pod revision might not be enough since it is not propagated consistently across all the StatefulSets.
+			// See https://github.com/elastic/cloud-on-k8s/issues/2393#issuecomment-572951884
 			podVersion, err := label.ExtractVersion(pod.Labels)
 			if err != nil {
 				return toUpgrade, err
--- a/pkg/controller/elasticsearch/driver/upgrade.go
+++ b/pkg/controller/elasticsearch/driver/upgrade.go
@@ -10,7 +10,9 @@ import (
 	esv1 "github.com/elastic/cloud-on-k8s/pkg/apis/elasticsearch/v1"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common/expectations"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common/reconciler"
+	"github.com/elastic/cloud-on-k8s/pkg/controller/common/version"
 	esclient "github.com/elastic/cloud-on-k8s/pkg/controller/elasticsearch/client"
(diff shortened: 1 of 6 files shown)


pkg/controller/elasticsearch/driver/upgrade.go
```diff
@@ -194,7 +202,16 @@ func podsToUpgrade(
...
+			// We consider that a Pod needs an update if at least one the following condition is met:
+			// 1. The update revision of the Pod does not match the one in the status of the StatefulSet
+			// 2. The Elasticsearch version run by the Pod does not match the expected one in the Elasticsearch object
+			// Using the Pod revision might not be enough since it is not propagated consistently across all the StatefulSets.
+			// See https://github.com/elastic/cloud-on-k8s/issues/2393#issuecomment-572951884
+			podVersion, err := label.ExtractVersion(pod.Labels)
+			if err != nil {
+				return toUpgrade, err
+			}
+			if sset.PodRevision(pod) != statefulSet.Status.UpdateRevision || !podVersion.IsSame(*esVersion) {
```
charith-elastic: I am just curious to know whether we can be certain that the version label would be up-to-date when the revision label isn't? Is it due to a difference in timing between the the Statefulset controller and ECK? (i.e. StatefuleSet controller is more relaxed about updating the revision than ECK is about updating the version label)

barkbay: This is the revision in the status that seems to not be up to date. I didn't have the time to investigate why it can take some time to update it while the generation has been updated.

pkg/controller/elasticsearch/driver/upgrade.go
```diff
@@ -194,7 +202,16 @@ func podsToUpgrade(
 				// Pod does not exist, continue the loop as the absence will be accounted by the deletion driver
 				continue
 			}
-			if sset.PodRevision(pod) != statefulSet.Status.UpdateRevision {
+			// We consider a Pod for an upgrade if at least one the following conditions is met:
```
thbkrkr: Nit: one **of** the following?

PR #2414, opened 2020-01-13
(DOCS) Adding alternative to reaching es in k8s

mjmbischoff:
.. without having to modify the loadbalancer k8s resource if curl is used

Small addition to the docs to alternatively make use of --resolve this could be useful for CI and other temporary hosts/use.

### DIFF
--- a/docs/accessing-elastic-services.asciidoc
+++ b/docs/accessing-elastic-services.asciidoc
@@ -230,11 +230,18 @@ curl: (51) SSL: no alternative certificate subject name matches target host name
 ----
 
 As the certificate subject by default is service name we can resolve this by using the service name 
-and specify the resolve information through an additional argument.  
+and specify the resolve information through an additional argument.
 
 [source,sh]
 ----
-curl --cacert tls.crt -u elastic:$PW https://$NAME-es-http:9200/ --resolve $NAME-es-http:9200:$IP
+> curl --cacert tls.crt -u elastic:$PW https://$NAME-es-http:9200/ --resolve $NAME-es-http:9200:$IP
+{
+  "name" : "hulk-es-4qk62zd928",
+  "cluster_name" : "hulk",
+  "cluster_uuid" : "q6itjqFqRqW576FXF0uohg",
+  "version" : {...},
+  "tagline" : "You Know, for Search"
+}
 ----
 
 Alternatively, and for other clients where the resolve information can't be specified, or when it is more convenient modify the server side, you can add the external IP of the service to the SANs of the certificate in the same Elasticsearch resource YAML manifest used to create the cluster and apply it again using `kubectl`.
@@ -252,21 +259,7 @@ spec:
         - ip: 35.198.131.115
 ----
 
-You can now reach Elasticsearch:
-
-[source,sh]
-----
-> curl --cacert tls.crt -u elastic:$PW https://$NAME-es-http:9200/ --resolve $NAME-es-http:9200:$IP
-{
-  "name" : "hulk-es-4qk62zd928",
-  "cluster_name" : "hulk",
-  "cluster_uuid" : "q6itjqFqRqW576FXF0uohg",
-  "version" : {...},
-  "tagline" : "You Know, for Search"
-}
-----
-
-If IP was added to the list of SANs: 
++You can now reach Elasticsearch without `--resolve`:
 
 [source,sh]
 ----
--- a/docs/accessing-elastic-services.asciidoc
+++ b/docs/accessing-elastic-services.asciidoc
@@ -229,7 +229,15 @@ Now you should get this message:
 curl: (51) SSL: no alternative certificate subject name matches target host name '35.198.131.115'
 ----
 
-Add the external IP of the service to the SANs of the certificate in the same Elasticsearch resource YAML manifest used to create the cluster and apply it again using `kubectl`.
+As the certificate subject by default is service name we can resolve this by using the service name 
+and specify the resolve information through an additional argument.  
+
+[source,sh]
+----
+curl --cacert tls.crt -u elastic:$PW https://$NAME-es-http:9200/ --resolve $NAME-es-http:9200:$IP
+----
+
+Alternatively, and for other clients where the resolve information can't be specified, or when it is more convenient modify the server side, you can add the external IP of the service to the SANs of the certificate in the same Elasticsearch resource YAML manifest used to create the cluster and apply it again using `kubectl`.
 
 [source,yaml]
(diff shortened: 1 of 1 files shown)


thbkrkr:
I'm +1 to add this nice addition. Nevertheless I think we can find a better place for it to be more explicit about the fact that it gives a way to not modify the loadbalancer k8s resource. `if you only use curl,` is not enough clear for me.

Maybe use this curl command in the first snippet of the "Outside the Kubernetes cluster" section. Then, explain that to use the IP directly (`curl --cacert tls.crt -u elastic:$PW https://$IP:9200/`) you have to modify the LB to avoid the `curl: (51) SSL: no alternative` error.

mjmbischoff:
sorry, my git-fu failed me there.

mjmbischoff:
While going over this I noticed the last example in the doc is using $PASSWORD vs $PW. it's something we probably want to fix regardless.

Moved the things up which with additional text which hopefully makes things better. The trade off is that the last section now has two blocks which might confuse someone who doesn't fully grasp what's going on.

Lastly maybe it's also a consideration to cover 'by DNS name' since instructions for that are provided earlier on the page.

thbkrkr:
> While going over this I noticed the last example in the doc is using $PASSWORD vs $PW. it's something we probably want to fix regardless.

Yes, thank you for fixing it.

> Moved the things up which with additional text which hopefully makes things better.

I think so.

> Lastly maybe it's also a consideration to cover 'by DNS name' since instructions for that are provided earlier on the page.

You are right. WDYT if we keep only one curl command with the `--resolve`. Repeating it at the end doesn't do much.

```diff
diff --git a/docs/accessing-elastic-services.asciidoc b/docs/accessing-elastic-services.asciidoc
index d9381002..5494e359 100644
--- a/docs/accessing-elastic-services.asciidoc
+++ b/docs/accessing-elastic-services.asciidoc
@@ -229,12 +229,19 @@ Now you should get this message:
 curl: (51) SSL: no alternative certificate subject name matches target host name '35.198.131.115'
 ----
 
-As the certificate subject by default is service name we can resolve this by using the service name 
-and specify the resolve information through an additional argument.  
+As the certificate subject by default is service name we can resolve this by using the service name
+and specify the resolve information through an additional argument.
 
 [source,sh]
 ----
 curl --cacert tls.crt -u elastic:$PW https://$NAME-es-http:9200/ --resolve $NAME-es-http:9200:$IP
+{
+  "name" : "hulk-es-4qk62zd928",
+  "cluster_name" : "hulk",
+  "cluster_uuid" : "q6itjqFqRqW576FXF0uohg",
+  "version" : {...},
+  "tagline" : "You Know, for Search"
+}
 ----
 
 Alternatively, and for other clients where the resolve information can't be specified, or when it is more convenient modify the server side, you can add the external IP of the service to the SANs of the certificate in the same Elasticsearch resource YAML manifest used to create the cluster and apply it again using `kubectl`.
@@ -252,21 +259,7 @@ spec:
         - ip: 35.198.131.115
 ----
 
-You can now reach Elasticsearch:
-
-[source,sh]
-----
-> curl --cacert tls.crt -u elastic:$PW https://$NAME-es-http:9200/ --resolve $NAME-es-http:9200:$IP
-{
-  "name" : "hulk-es-4qk62zd928",
-  "cluster_name" : "hulk",
-  "cluster_uuid" : "q6itjqFqRqW576FXF0uohg",
-  "version" : {...},
-  "tagline" : "You Know, for Search"
-}
-----
-
-If IP was added to the list of SANs: 
+You can now reach Elasticsearch without `--resolve`:
```

mjmbischoff:
> WDYT if we keep only one curl command

Yeah looks good.

thbkrkr:
LGTM.

Just a leftover `+` to delete and we are good.

docs/accessing-elastic-services.asciidoc
```diff
@@ -229,7 +229,22 @@ Now you should get this message:
 curl: (51) SSL: no alternative certificate subject name matches target host name '35.198.131.115'
 ----
 
-Add the external IP of the service to the SANs of the certificate in the same Elasticsearch resource YAML manifest used to create the cluster and apply it again using `kubectl`.
+As the certificate subject by default is service name we can resolve this by using the service name 
+and specify the resolve information through an additional argument.
```
anyasabo: looks like an extra line break here

docs/accessing-elastic-services.asciidoc
```diff
@@ -229,7 +229,22 @@ Now you should get this message:
...
+{
+  "name" : "hulk-es-4qk62zd928",
+  "cluster_name" : "hulk",
+  "cluster_uuid" : "q6itjqFqRqW576FXF0uohg",
+  "version" : {...},
+  "tagline" : "You Know, for Search"
+}
+----
+
+Alternatively, and for other clients where the resolve information can't be specified, or when it is more convenient modify the server side, you can add the external IP of the service to the SANs of the certificate in the same Elasticsearch resource YAML manifest used to create the cluster and apply it again using `kubectl`.
```
anyasabo: I think there may be a simpler way to phrase this. Maybe
"As an alternative, you can edit the Elasticsearch spec to include the external IP of the service to the certificate subject alternative names (SANs) as described below. This is useful for other clients where resolve information cannot be specified, or when it is more convenient to make changes on the server rather than client." ?

anyasabo:
Left some minor editing notes, thanks for bringing this up

mjmbischoff:
Think this PR is a good example of how technical writing is hard. :sweat_smile: 

Thank you for all the feedback,

docs/accessing-elastic-services.asciidoc
```diff
@@ -229,7 +229,21 @@ Now you should get this message:
 curl: (51) SSL: no alternative certificate subject name matches target host name '35.198.131.115'
 ----
 
-Add the external IP of the service to the SANs of the certificate in the same Elasticsearch resource YAML manifest used to create the cluster and apply it again using `kubectl`.
+As the certificate subject by default is service name we can resolve this by using the service name and specify the resolve information through an additional argument.
```
anyasabo: I think there might be a better way to phrase and avoid the double usage of "resolve". Maybe
>As the certificate subject by default is service name we can resolve this by using the service name and specify the resolve information through an additional argument.

to

>You can avoid this in curl by manually specifying the certificate subject and IP address to the `resolve` argument
?

thbkrkr: Heck, sorry I merged without seeing this comment.
We can bring this little improvement in another PR.

PR #2429, opened 2020-01-15
PULL_SECRET should be set for dev environments

flaper87:


hack/deployer/runner/settings.go
```diff
@@ -64,6 +64,7 @@ type OcpSettings struct {
 	Region        string `yaml:"region"`
 	AdminUsername string `yaml:"adminUsername"`
 	WorkDir       string `yaml:"workDir"`
+	PullSecret    string `yaml:"PullSecret"`
```
david-kow:
```suggestion
	PullSecret    string `yaml:"pullSecret"`
```

flaper87: argh, forgot to change the capitalization

PR #2432, opened 2020-01-15
Add CI files for OCP

### LINKED ISSUE #2170: Create OpenShift 4.x e2e testing environment
We need an automated OpenShift 4.x environment to run our e2e tests on. The idea is to integrate the OpenShift installer with our existing deployer tooling to stand up an Openshift installation
* we need to be able to create and teardown these environments from Jenkins
* we would like to be able to do the same from a local machine for debugging and development purposes

flaper87:
This PR will add all the CI files to setup an e2e job on OpenShift

Fixes #2170


build/ci/e2e/Jenkinsfile-ocp
```diff
@@ -0,0 +1,96 @@
...
+        stage('Run Checks') {
+            steps {
+                sh 'make -C build/ci TARGET=ci-check ci'
+            }
+        }
+        stage("Run E2E tests") {
+            steps {
+                sh """
+                    cat >.env <<EOF
+GCLOUD_PROJECT = $GCLOUD_PROJECT
```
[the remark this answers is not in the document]
flaper87: done, thanks!

flaper87:
@david-kow In order for `openshift-install` to work, I have to configure the CI service account (grant it specific permissions) and I'll have to create an `ocp-ci` test zone. I'll start configuring that, then I'll run tests as you recommended and get back with an updated version of this PR when those commands pass for me

Thanks :)

david-kow:
@flaper87 sounds great, thank you! Let me know when I could be of any help.

flaper87:
Small update: I'm making progress and very close to submitting an update to this PR. :)

flaper87:
retest this please

go.mod
```diff
@@ -44,18 +44,15 @@ require (
...
 	golang.org/x/sys v0.0.0-20191010194322-b09406accb47 // indirect
 	golang.org/x/time v0.0.0-20190921001708-c4c64cad1fd0 // indirect
-	golang.org/x/xerrors v0.0.0-20191011141410-1b5146add898 // indirect
 	google.golang.org/appengine v1.6.5 // indirect
 	gopkg.in/yaml.v2 v2.2.4
 	gopkg.in/yaml.v3 v3.0.0-20191120175047-4206685974f2
-	k8s.io/api v0.0.0-20191114100352-16d7abae0d2a
-	k8s.io/apimachinery v0.0.0-20191028221656-72ed19daf4bb
+	k8s.io/api v0.17.1
+	k8s.io/apimachinery v0.17.1
```
sebgl: Did you bump those on purpose? I think we use the ones coming with kubebuilder/controller-runtime by default.

hack/deployer/runner/ocp.go
```diff
@@ -78,23 +78,24 @@ type OcpDriver struct {
 func (gdf *OcpDriverFactory) Create(plan Plan) (Driver, error) {
 	baseDomain := plan.Ocp.BaseDomain
 	if baseDomain == "" {
-		baseDomain = "ocp.elastic.dev"
+		baseDomain = "eck-ocp.elastic.dev"
```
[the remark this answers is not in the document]
flaper87: I'm torn but I'd like to hear your thoughts about it.

This is not a setting that can be 'just changed'. It requires an actual zone to exist in GCP (which I have configured). It is already possible to overwrite this value but I think that having this default in the code is ok. We've created this Zone under the default project and if we ever change it I think it's ok to require a change here for that.

hack/deployer/runner/ocp.go
```diff
@@ -166,6 +171,8 @@ func (d *OcpDriver) Execute() error {
 }
 
 func (d *OcpDriver) auth() error {
+	_ = NewCommand(fmt.Sprintf("gcloud config set project %s", d.ctx["GCloudProject"])).Run()
```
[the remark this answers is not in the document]
flaper87: It is, but I can triple check if needed. Without it, the OCP deployment may fail since it'll try to use the default configured account /project. This was not set for me when I ran it.

hack/deployer/runner/ocp.go
```diff
@@ -271,19 +283,37 @@ func (d *OcpDriver) uploadCredentials() error {
 }
 
 func (d *OcpDriver) GetCredentials() error {
+	log.Printf("Getting credentials")
 	kubeConfig := filepath.Join(d.ctx["ClusterStateDir"].(string), "auth", "kubeconfig")
+	log.Printf("%s", kubeConfig)
```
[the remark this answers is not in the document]
flaper87: no... leftover. I'll remove this

hack/deployer/runner/ocp.go
```diff
@@ -271,19 +283,37 @@ func (d *OcpDriver) uploadCredentials() error {
...
+		}
+
+		return nil
+	}
 	// We do this check twice to avoid re-downloading files
 	// from the bucket when we already have them locally.
 	// The second time is further down in this function and it's
 	// done when the rsync succeeds
 	if _, err := os.Stat(kubeConfig); !os.IsNotExist(err) {
+		_ = copyKubeconfig()
```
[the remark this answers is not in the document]
flaper87: correct! We should handle an error here. Good catch.

test/e2e/cmd/run/run.go
```diff
@@ -149,6 +149,7 @@ func (h *helper) initTestContext() error {
 		TestRun:               h.testRunName,
 		TestTimeout:           h.testTimeout,
 		IgnoreWebhookFailures: h.ignoreWebhookFailures,
+		OcpCluster:            h.kubectl("get", "clusterversion") == nil,
```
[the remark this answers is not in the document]
flaper87: My goal with this was to try to not make OCP a special case. I do realize that the end result here is that in this case I'm doing just that.

My hope, however, is for this to not stay as is forever. Hopefully, all the ifs/cases required for OCP will go away eventually as they are mostly related to how restricted OCP is by default and not because of some technical limitation.

This is all to say that I'd love for this special case to be an internal thing rather than it being exposed externally so that we can remove freely when we reach that ideal state.

Curious to hear your thoughts.

go.mod
flaper87: I did, or I should say: The ocp dependency did. Do you think this will be a problem?

sebgl: I think in theory that's a problem.
But in practice that's probably not 😄 I would be surprised if that causes actual troubles.

However, the right way to deal with this IMO is to make the `deployer` have its own `go.mod` and `go.sum`, entirely decoupled from the operator.
I think there's no reason for both projects to share their dependencies (which does not mean they cannot live in the same github repo).

barkbay: I think I would lean on the safe side and keep these dependencies driven by the controller runtime if it's possible. It is already the case for the license detector so I guess we can do the same for the deployer.

flaper87: Unfortunately, I think not bumping the version may be an issue for this patch. `openshift/api` depends on these versions so I'll probably take @sebgl suggestion and add a `go.mod` and `go.sum` for the `deployer`.

Any thoughts/opinions on how you would prefer to have this done?

flaper87: I should note that this is a requirement for the `tests` only, which means we could add these module files to the `test/e2e` only.

hack/deployer/runner/ocp.go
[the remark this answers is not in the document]
david-kow: I'd say it's similar to Azure Container Registry setting or GCloudProject (require setup, not likely to change). Ideally, we can have someone outside of Elastic be able to run it if he has right things in right places. This makes someone need to recompile. I tend to think there is no harm in moving it to `plans.yml`, but if you feel strongly about this, feel free to keep it here :)

test/e2e/cmd/run/run.go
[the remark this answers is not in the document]
david-kow: Got it, I'm fine with keeping this as is.

go.mod
flaper87: > I think we can avoid to rely on the OCP client if we just need to patch the SCC , which is a CRD starting 4.x btw:
> 
> ```
> $ k get crd securitycontextconstraints.security.openshift.io 
> NAME                                               CREATED AT
> securitycontextconstraints.security.openshift.io   2020-02-07T06:52:43Z
> ```
> 
> This can be done by using the legacy K8S client, using a `PATCH` request on the right endpoint:
> 
> ```go
> user := fmt.Sprintf(`"system:serviceaccount:%s:%s"`, 
>                            b.ServiceAccount.Namespace,
>                            b.ServiceAccount.Name,
>                  )
> // The patch below adds the service account user in the 'users' fields of a SCC
> patch := []byte(`{ "users": [` + user + `]}`)
> 
> // We want to patch the anyuid SCC. In term of url it means that we need to send a patch request to:
> // https://<Openshift URL>/apis/security.openshift.io/v1/securitycontextconstraints/anyuid
> patchClient := k8sClient.RESTClient().
>     Patch(types.MergePatchType).
>     Prefix("apis", "security.openshift.io", "v1").
>     Resource("securitycontextconstraints").
>     Name("anyuid").
>     Body(patch)
> 
> result := patchClient.Do()
> ```
> 
> See a full patch [here](https://github.com/elastic/cloud-on-k8s/compare/openshift...barkbay:openshift-patch-with-native-client#diff-b7b20c2a1a27a46b32159937c7485845R33)
> 
> It would avoid to rely on those dependencies and deal with new go module files. On other benefit is that it seems to work with both OCP 4.x and 3.11, I'm not sure if the official OCP client for 4.x is compatible with 3.11.

Oh, nice! Thanks. I had tried something like this but I think I did something wrong 'cause it didn't work for me. I'll go with this just to unblock this PR. 

As a side note, it feels a bit hacky in the sense that we are avoiding a, perhaps more important, issue that is handling SCCs. I understand we are just talking about our test suite here and not the operator itself, however, I think it would be better to deal with this properly by using the actual clients.

Thank you for taking the time to write this patch

flaper87: @barkbay patch applied perfectly. Just launched tests now, will report back soon :)

flaper87:
I ran a a full test throughout the night and it succeeded with the currently pushed version. I think this is ready for another review. Thanks for all the feedback @barkbay @david-kow @sebgl

config/e2e/rbac.yaml
```diff
@@ -86,13 +86,22 @@ rules:
...
       - delete
       - create
       - update
+  - apiGroups:
+      - "security.openshift.io"
+    resources:
+      - securitycontextconstraints
+    verbs:
+      - get
+      - update
```
flaper87: We should be able to do away with the `update` permission. Should be harmless to keep it for now.

hack/deployer/config/plans.yml
```diff
@@ -64,6 +64,7 @@ plans:
   ocp:
     region: europe-west2
     nodeCount: 3
+    overwriteDefaultKubeconfig: true
```
david-kow: Should we be explicit in `ocp-dev` and set `overwriteDefaultKubeconfig: false`?

hack/deployer/runner/ocp.go
```diff
@@ -237,31 +245,35 @@ func (d *OcpDriver) create() error {
...
+	// this way we can run a delete operation even on failed
+	// deployments to clean all the resources on GCP.
+	_ = d.uploadCredentials()
 
 	if err != nil {
 		return err
 	}
 
-	return d.uploadCredentials()
+	return nil
```
david-kow: This can be shorten to `return err`.

hack/deployer/config/plans.yml
flaper87: not a bad idea, I'll do it if we need another push to this PR. Otherwise, I'll send a follow-up patch :)

flaper87:
Ok, there are enough nits to make an updated push worth it :smile: 

I'll address these comments between today and tomorrow.

hack/deployer/config/plans.yml
flaper87: done!

Makefile
```diff
@@ -324,6 +324,9 @@ docker-push:
 ifeq ($(REGISTRY), docker.elastic.co)
 	@ docker login -u $(ELASTIC_DOCKER_LOGIN) -p $(ELASTIC_DOCKER_PASSWORD) push.docker.elastic.co
 endif
+ifeq ($(REGISTRY), eu.gcr.io)
+	@ gcloud auth configure-docker --quiet
```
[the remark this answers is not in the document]
flaper87: This wasn't done automatically for me and after debugging the authentication failure with @david-kow we found out that the local docker configs/auths hooks were not set properly.

I tried running this command on environments that were already configured and it succeeded. This is to say that it should be harmless to run it multiple times from here.

flaper87:
Alrighty, all comments addressed! Thank y'all for your reviews :)

thbkrkr:
LGTM.
Nice contribution, thank you.

flaper87:
okidoki, let's see how this works once merged! thank y'all

PR #2433, opened 2020-01-15
Add config validation for unknown fields

david-kow:
I'm not sure if this was considered, but this PR uses kubectls last applied config annotation to catch misaligned specs:

```
$ cat bad.yaml
kind: Elasticsearch
metadata:
  name: elasticsearch-sample
spec:
  version: 7.5.0
  nodeSets:
  - name: default
    count: 3
    config:
      node.store.allow_mmap: false
podTemplate:
(...)
$ k apply -f bad.yaml
```
`Error from server (Elasticsearch.elasticsearch.k8s.elastic.co "elasticsearch-sample" is invalid: "podTemplate": Invalid value: "elasticsearch-sample": json: unknown field "podTemplate"): error when creating "bad.yaml": admission webhook "elastic-es-validation-v1.k8s.elastic.co" denied the request: Elasticsearch.elasticsearch.k8s.elastic.co "elasticsearch-sample" is invalid: "podTemplate": Invalid value: "elasticsearch-sample": json: unknown field "podTemplate"`

I can expand to Kibana/APM and add tests, but first: is this something we would like to add? While the method is not the cleanest, it does address a really painful category of issues. One more thing to note is that the error message is not perfect, it will only mention the key even if it's nested deeper. WDYT?


anyasabo:
I think I'm good with this approach. The only time I can see it being an issue is that if an object is created with kubectl, but then later edited outside of `kubectl apply`, so the annotation does not get updated. `kubectl edit` doesn't appear to update the annotation (verified via hand and hinted at in the [docs](https://kubectl.docs.kubernetes.io/pages/app_management/field_merge_semantics.html), and anything that's not kubectl will not either.

That said, it doesn't leave us in a worse position than we're currently in (the object was already admitted successfully by kubectl, so we would potentially admit an invalid object), and at least catches the errors with `kubectl apply`.

References from our slack discussion for the rest:

A bunch of discussion on pruning/defaulting/rejecting unknown fields here: https://github.com/kubernetes/kubernetes/issues/59151

the overall "strict mode" issue: https://github.com/kubernetes/kubernetes/issues/5889

and it seems like they wanted to implement pruning and defaulting before implementing a strict mode. See the defaulting KEP here: https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/20190426-crd-defaulting.md

It will be a while before we have similar functionality generally available in the api server directly, so I think it can make sense for us to have a bandaid solution for now. Thanks for raising this David.

anyasabo:
@barkbay raised an interesting point: this does cause issues where a user currently has a broken manifest that they successfully `kubectl apply`ed. If they upgrade to include this validation, they can no longer `kubectl edit` their resource to fix it. They would need to either remove the annotation or fix it via `apply`. This only applies to existing users though -- once a resource has been `kubectl apply`ed successfully with this validation enabled, it should never happen again.

I think I'm still okay with this change -- we can make it very clear in the release notes for the next version that this is the case. Since the pod template wasn't doing anything before, I hope we do not have a lot of people in that situation out there 🤞

david-kow:
@barkbay @anyasabo these are good points.

A way to go around getting stuck would be to limit this validation to `kubectl apply` only. We could do this by comparing the spec from the annotation with the spec object we get. We would only perform the validation if they match. IIUC, this should only be the case if `kubectl apply` was used. In all other cases we would ignore annotation (and not perform this validation) as we just wouldn't have a way to do it.

But I'd be fine if we stick to what it's there now and just document it. Like @anyasabo said, this should be rare circumstance.

sebgl:
What I dislike with this PR initially is that it may lead us to some inconsistent situations, where a resource update succeeds when done in one way (eg. update API call from my CI tooling) but not when done another way (eg. `kubectl apply` was done by a human).
I don't see a better way to do it right now though, it would probably involve hacking into our webhook mechanism so we grab the raw http request json payload before we deserialize it.

Any case where a user would get stuck because the kubectl annotation is wrong is a bit scary.
I think your last proposal gives an extra layer of safety which makes me a bit more comfortable: the spec vs. annotation comparison to make sure we are in a `kubectl apply` case is worth a try IMO.

david-kow:
@pebrc Yes, I wanted to get consensus on the approach before doing the doc part, I should've been more clear about this.

> I don't see a better way to do it right now though, it would probably involve hacking into our webhook mechanism so we grab the raw http request json payload before we deserialize it.

That's the first thing I tried :) IIUC that's not possible because the marshalling into the object is done in the api-server which then sends it over to our hook (ie. hook never sees raw payload sent by the user).

I'll go ahead and add the additional check.

david-kow:
After changing the code as proposed above it turned out there are some complications. The object contains things that annotation doesn't:
```
$ diff obj.json annot.json
7,11c7
<     "selfLink": "/apis/elasticsearch.k8s.elastic.co/v1/namespaces/default/elasticsearches/quickstart",
<     "uid": "4a3e12c5-6f79-4cbd-9494-029ec140d8f8",
<     "resourceVersion": "5159851",
<     "generation": 3,
<     "creationTimestamp": "2020-01-27T15:37:45Z",
---
>     "creationTimestamp": null,
13,14d8
<       "common.k8s.elastic.co/controller-version": "1.0.0",
<       "elasticsearch.k8s.elastic.co/cluster-uuid": "Gr2iGoiAR-aXkuHKnS_9GA",
61,65c55
<   "status": {
<     "availableNodes": 1,
<     "health": "unknown",
<     "phase": "Ready"
<   }
---
>   "status": {}
```

We could limit the comparison to the `spec` part, but it would mean that can't detect if someone used kubectl or not when something outside of spec (e.g. metadata) was modified. If the other-than-kubectl-apply tool modifies something outside the spec, we would run the validation and potentially fail it. Example would be incorrect metadata in annotation being fixed by a direct API call - validation would fail based on the fact that a) specs match, b) metadata in annotation is invalid.

Given the above is not ideal and starts to be difficult to reason about, I'd propose to go back to the original idea and validate always if the annotation is present. If we fail validation based on that, it means that whatever was intended at some point, was not achieved. Users still have two ways out: removing annotation and turning off validation.

pkg/apis/common/v1/validations.go
```diff
@@ -0,0 +1,34 @@
...
+	if cfg, ok := meta.Annotations[v1.LastAppliedConfigAnnotation]; ok {
+		d := json.NewDecoder(strings.NewReader(cfg))
+		d.DisallowUnknownFields()
+		if err := d.Decode(&dest); err != nil {
+			errString := err.Error()
+			unknownPrefix := "json: unknown field "
+			if strings.HasPrefix(errString, unknownPrefix) {
+				fld := strings.TrimPrefix(errString, unknownPrefix)
+				fld = fld[1 : len(fld)-1] // removes quotes from fld
+				errs = append(errs, field.Invalid(field.NewPath(fld), fld, fmt.Sprintf("%s field is unknown", fld)))
```
sebgl: I think we'd like a hint in the error message that fields were parsed from the `v1.LastAppliedConfigAnnotation` annotation.

pkg/apis/common/v1/validations.go
```diff
@@ -0,0 +1,34 @@
...
+	var errs field.ErrorList
+	if cfg, ok := meta.Annotations[v1.LastAppliedConfigAnnotation]; ok {
+		d := json.NewDecoder(strings.NewReader(cfg))
+		d.DisallowUnknownFields()
+		if err := d.Decode(&dest); err != nil {
+			errString := err.Error()
+			unknownPrefix := "json: unknown field "
+			if strings.HasPrefix(errString, unknownPrefix) {
+				fld := strings.TrimPrefix(errString, unknownPrefix)
+				fld = fld[1 : len(fld)-1] // removes quotes from fld
```
sebgl: Should we check the slice length first?

david-kow: It's `%q`, so it should be fine, but no harm checking.

docs/troubleshooting.asciidoc
```diff
@@ -248,6 +248,15 @@ This can also be done for Kibana and APM Server.
...
 
 You can also change the https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#failure-policy[`failurePolicy`] of the webhook configuration to `Fail`, which will cause creations and updates to error out if there is an error contacting the webhook.
+
+=== Validation failures
+If validation webhook is preventing you from doing changes due to the unknown fields validation like below, you can force the webhook to ignore it by removing `kubectl.kubernetes.io/last-applied-configuration` annotation from your resource.
```
pebrc:
```suggestion
If the validation webhook is preventing you from making changes due to the unknown fields validation like below, you can force the webhook to ignore it by removing the`kubectl.kubernetes.io/last-applied-configuration` annotation from your resource.

File: docs/troubleshooting.asciidoc [hunk 4]
```diff
@@ -248,6 +248,15 @@ This can also be done for Kibana and APM Server.
...
+
+```
+admission webhook "elastic-es-validation-v1.k8s.elastic.co" denied the request: Elasticsearch.elasticsearch.k8s.elastic.co "quickstart" is invalid: some-misspelled-field: Invalid value: "some-misspelled-field": some-misspelled-field field is unknown
```
sebgl: should be updated to match the new error message

docs/troubleshooting.asciidoc
```diff
@@ -248,6 +248,15 @@ This can also be done for Kibana and APM Server.
...
 
 You can also change the https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#failure-policy[`failurePolicy`] of the webhook configuration to `Fail`, which will cause creations and updates to error out if there is an error contacting the webhook.
+
+=== Validation failures
+If the validation webhook is preventing you from making changes due to the unknown fields validation like below, you can force the webhook to ignore it by removing the`kubectl.kubernetes.io/last-applied-configuration` annotation from your resource.
+
+```
+admission webhook "elastic-es-validation-v1.k8s.elastic.co" denied the request: Elasticsearch.elasticsearch.k8s.elastic.co "quickstart" is invalid: some-misspelled-field: Invalid value: "some-misspelled-field": some-misspelled-field field found in kubectl.kubernetes.io/last-applied-configuration annotation is unknown
```
sebgl: `in the kubectl.kubernetes.io/last-applied-configuration annotation`?

PR #1812, opened 2019-09-27
Add support for MaxSurge and MaxUnavailable during scaling

david-kow:
This change will cause operator to prevent creating or removing nodes that would cause the pod counts to violate MaxSurge/MaxUnavailable settings in the change budget.

Some notes:
- MaxSurge (ms) is only checked during scaling up,
- MaxUnavailable (mu) is only checked during scaling down, 
- allowed counts are calculated based on desired, not current, state,
- ms defaults to math.MaxInt32, mu defaults to 1 (to align with default during rolling upgrade),
- only ready pods are counted towards available podes for maxUnavailable calculation,
- all existing pods are counted towards surge.

Fixes #1292.

PR #2437, opened 2020-01-16
Refactor invariant check during downscale

david-kow:
Tries to address past comment: https://github.com/elastic/cloud-on-k8s/pull/1812#discussion_r330513119. While two functions are not merged (I feel that `getMaxNodesToRemove` should be a separate func), `checkDownscaleInvariants` no longer inspects `removalsAllowed` and the call site is cleaned up.


pkg/controller/elasticsearch/driver/downscale_invariants_test.go
```diff
@@ -227,7 +227,8 @@ func Test_checkDownscaleInvariants(t *testing.T) {
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			canDownscale, reason := checkDownscaleInvariants(*tt.state, tt.statefulSet)
+			toDelete, reason := checkDownscaleInvariants(*tt.state, tt.statefulSet, 1)
+			canDownscale := toDelete > 0
```
sebgl: let's maybe test the exact value here to avoid any unpleasant surprise and make sure we test the contract

PR #2439, opened 2020-01-17
Facilitate filebeat autodiscovery with hints annotation

### LINKED ISSUE #1978: Facilitate filebeat autodiscovery with hints annotation
When creating Elasticsearch pods we could add a default filebeat annotation 

```yaml
    podTemplate:
      metadata:
        annotations:
          "co.elastic.logs/module": elasticsearch
```
which would allow filebeat [autodiscovery](https://www.elastic.co/guide/en/beats/filebeat/master/configuration-autodiscover-hints.html) to configure the Elasticsearch filebeat module for the Elasticsearch clusters created by ECK which in turn would give us nicely parsed logs:

Screenshot 2019-10-11 at 14 36 19

and integration with the Stack Monitoring feature in Kibana

Screenshot 2019-10-11 at 14 38 25

I see no harm in the additional annotation for users who don't use filebeat or don't use autodiscovery.

A similar annotation exists for Kibana:
```yaml
 podTemplate:                                                                                                                                         
    metadata:                                                                                                                                          
      annotations:                                                                                                                                     
        co.elastic.logs/module: kibana     
```

thbkrkr:
Add a default annotation to the ES/Kibana pods, 
taking care not to replace user-provided annotations.

Resolves #1978.


thbkrkr:
> I'm wondering if we should not also update the [beats recipe](https://github.com/elastic/cloud-on-k8s/blob/master/config/recipes/beats/1_monitor.yaml#L16) 
> and advertise that the annotation is automatically added by ECK.

I would say yes, to keep the latest recipes up to date with the latest operator.

pebrc:
Did we investigate if we can do something similar for APM server? E.g. https://www.elastic.co/guide/en/apm/server/current/monitoring-metricbeat-collection.html

pkg/controller/common/defaults/metadata.go
```diff
@@ -4,11 +4,11 @@
...
 package defaults
 
-// SetDefaultLabels append labels from defaults into existing. If a label already exists,
+// AppendMetadata append labels/annotations from defaults into existing. If a label/annotation already exists,
 // its value is not overridden from the one in defaults.
-// One use case here is to inherit user-provided labels, and append our own only if not already
+// One use case here is to inherit user-provided labels/annotations, and append our own only if not already
 // set by the user.
-func SetDefaultLabels(existing map[string]string, defaults map[string]string) map[string]string {
+func AppendMetadata(existing map[string]string, defaults map[string]string) map[string]string {
```
sebgl: `Metadata` feels a bit weird to me since it is a superset of labels + annotations with name, namespace, UID, etc.
Also `Append` feels a bit strange for a map (was there before this PR).

Maybe something like `MergeKeyValue(a, b)`?

thbkrkr: +1 for `Merge` instead of `Append` but not completely convinced by `KeyValue` which sounds a little too abstract.
Nevermind, while trying to move it to another package, I found:

https://github.com/elastic/cloud-on-k8s/blob/35bc1ba07779dcf5a32eaff54890462caffa6fe2/pkg/utils/maps/maps.go#L38-L54

:D

I'm going to reuse that and delete `AppendMetadata`.

PR #2449, opened 2020-01-20
Fix PDB documentation

sebgl:
The PDB doc gave an example of a custom PDB with `maxUnavailable: 2`.
Unfortunately this can only work with a builtin controller selector (eg.
matching a StatefulSet or a deployment). It does not do anything with
our custom cluster name label selector.

This commit changes it by relying on `minAvailable` instead, and gives
more details about how the PDB is set to allow one Pod to be taken down
only if the cluster health is green.


sebgl:
This should be backported to the 1.0 branch for documentation purposes.

Issue #2402, opened 2020-01-10
A performance issue


ftyuuu:
I use eck to get an es cluster with 5 nodes:

```
NAME                               READY   STATUS    RESTARTS   AGE   IP              NODE         NOMINATED NODE   READINESS GATES
elasticsearch-sample-es-client-0   1/1     Running   0          22m   10.16.215.128   172.24.5.5   <none>           <none>
elasticsearch-sample-es-data-0     1/1     Running   0          22m   10.16.184.218   172.24.5.4   <none>           <none>
elasticsearch-sample-es-data-1     1/1     Running   0          22m   10.16.33.114    172.24.5.7   <none>           <none>
elasticsearch-sample-es-data-2     1/1     Running   0          22m   10.16.215.172   172.24.5.5   <none>           <none>
elasticsearch-sample-es-master-0   1/1     Running   0          22m   10.16.33.108    172.24.5.7   <none>           <none>
```
Then i use es-rally (https://github.com/elastic/rally) to test this cluster, i use the http_logs dataset and only test with the operation of index-append:

```
{
      "name": "index-append",
      "operation-type": "bulk",
      "bulk-size": {{bulk_size | default(5000)}},
      "ingest-percentage": {{ingest_percentage | default(100)}},
      "corpora": "http_logs"
}
```
The challenges is:
```
"schedule": [
        {
          "operation": "delete-index"
        },
        {
          "operation": {
            "operation-type": "create-index",
            "settings": {{index_settings | default({}) | tojson}}
          }
        },
        {
          "name": "check-cluster-health",
          "operation": {
            "operation-type": "cluster-health",
            "index": "logs-*",
            "request-params": {
              "wait_for_status": "{{cluster_health | default('green')}}",
              "wait_for_no_relocating_shards": "true"
            }
          }
        },
        {
          "operation": "index-append",
          "warmup-time-period": 240,
          "clients": {{bulk_indexing_clients | default(30)}}
        }
      ]
```

I performed 2 sets of tests
1、es-cluster-1, set memory request to 8g and set memory limit to 16g, the cr of es cluster is:
```
apiVersion: elasticsearch.k8s.elastic.co/v1beta1
kind: Elasticsearch
metadata:
  name: elasticsearch-sample
  namespace: nes-elasticsearch
spec:
  version: 6.6.1
  http:
    tls:
      selfSignedCertificate:
        disabled: true
  nodeSets:
    - name: master
      config:
        node.master: true
        node.data: false
[...]

PR #2454, opened 2020-01-21
Mention the Kubernetes issue with CPU limits throttling in the docs

sebgl:
Some users may experience poor performance correlated with the usage of
cpu limits. This is due to a bug in the Kernel that has been fixed in
recent Kernel versions.
A workaround is to not use limits at all, or tweak the CFS quota period.

---

I'm wondering whether this PR is worth the trouble: it's hard to get into this level of details without saying too much. Also depends on various factors (Kernel version, bugfixed or not, can the user tweak the CFS period, etc.).
I'd be happy to hear other people thoughts, we may want to close this PR if it seems too complicated to explain.

### DIFF
--- a/docs/managing-compute-resources.asciidoc
+++ b/docs/managing-compute-resources.asciidoc
@@ -24,9 +24,15 @@ A good rule of thumb is to size it to half the size of RAM allocated to the Pod.
 To minimize disruption caused by Pod evictions due to resource contention, you should run Elasticsearch pods at the "Guaranteed" QoS level by setting both `requests` and `limits` to the same value.
 
 Consider also that Kubernetes throttles containers exceeding the CPU limit defined in the `limits` section. Do not set this value too low or it would affect the performance of Elasticsearch, even if you have enough resources available in the Kubernetes cluster.
-A link:https://github.com/kubernetes/kubernetes/issues/51135[known Kubernetes issue] may lead to over-aggressive CPU limits throttling. If the host Linux Kernel does not include link:https://github.com/kubernetes/kubernetes/issues/67577[this CFS quota fix], you may want to keep CPU limits unset, or link:https://github.com/kubernetes/kubernetes/pull/63437[tweak the CFS quota period].
 
+[NOTE]
+===============================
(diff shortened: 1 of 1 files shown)


thbkrkr:
I'm in favor to mention this issue with CPU limits throttling which seems to me very important to be aware.

Reading your proposal, I find it's a bit hard to understand that 'do not set any CPU limit' is also a good option. I liked the way you presented the 3 options in #2402.

> Users can either:
> * (k8s admin level) disable CFS quotas
> * (k8s admin level) reduce CFQ quota period
> * (ECK user level) do not set any CPU limit

docs/managing-compute-resources.asciidoc
```diff
@@ -18,7 +18,14 @@ You can set compute resource constraints in the `podTemplate` of objects managed
...
+For Elasticsearch objects, make sure to consider the heap size when you set resource requirements.
+A good rule of thumb is to size it to half the size of RAM allocated to the Pod.
+
+To minimize disruption caused by Pod evictions due to resource contention, you should run Elasticsearch pods at the "Guaranteed" QoS level by setting both `requests` and `limits` to the same value.
+
+Consider also that Kubernetes throttles containers exceeding the CPU limit defined in the `limits` section. Do not set this value too low or it would affect the performance of Elasticsearch, even if you have enough resources available in the Kubernetes cluster.
```
barkbay: nit: should we make it a `Note:` right below the example since it looks like an hint or an advice ? (cc @alaudazzi)

anyasabo:
+1 the PR is worth the trouble IMO. Agree it would be better served in official k8s documentation, and we may want to make a PR to add that. But in the meantime we should have a mention somewhere in our docs.

>I feel less comfortable to give some options knowing that some of them may have some side effects and must be considered in the context of the use case

I definitely understand the concern, but I think at least mentioning them as options is okay. It gives people a starting point so they can do some reading and experimenting to decide if it is appropriate for them. For that reason I like @thbkrkr 's suggestion.

I can also see the argument to split this out to a separate section of the docs, and link to it from the limits section. I think if we add more detail (which I am in favor of) then we do want to split that out into a new section.

sebgl:
I added a `NOTE` section with the 3 options.
I don't think it's worth moving to a different page yet.

Issue #2002, opened 2019-10-16
Support ECS in operator logs


pebrc:
Currently our logging output does not play nicely with Elastic observability tooling in Kibana for example because we don't follow [ECS](https://github.com/elastic/ecs) in the JSON logs we produce.

Example: ECK log  output not fully parsed in Kibana logs view:
<img width="1616" alt="Screenshot 2019-10-16 at 16 55 58" src="https://user-images.githubusercontent.com/697790/66931303-30415900-f036-11e9-9d79-fd3391c18c75.png">

There are two options to solve this: 
1. transformation on ingest to conform to ECS
2. adjust our logging format to produce ECS compliant JSON

\2. seems easier as we have full control over what we produce and it would simplify ingestion (no transformation needed at ingest)

anyasabo:
How do things like name and namespace map to ECS? Or does it make sense to map the things we can (looks like `level`, `timestamp`, `logger`, and `message` at least), and then leave the rest as is?

pebrc:
> How do things like name and namespace map to ECS?
As they don't map to any ECS fields they would just stay in their original form.

barkbay:
Just to share an issue I'm facing:

In case of an error in ECK the following log is generated by the controller runtime:

```json
{
[...]

Issue #2406, opened 2020-01-10
Add tooling to customize the all-in-one manifests


sebgl:
Some users would like to run the operator in a particular namespace, with restricted RBAC access:
* https://discuss.elastic.co/t/deploy-eck-to-a-single-namespace/211495/5
* https://github.com/elastic/cloud-on-k8s/issues/374#issuecomment-572602235

Issue https://github.com/elastic/cloud-on-k8s/issues/374 and [this design doc](https://github.com/elastic/cloud-on-k8s/blob/master/docs/design/0005-configurable-operator.md) mentions a CLI tool to help generate the correct manifest according to one's need:
* operator deployed cluster-wide vs. in a single namespace
* reduced RBAC privileges according to the above choice
* list of namespaces for the operator to monitor

Detailing every manual patch to be made in a documentation page seems quite complicated, so a CLI/script/helm/kustomize seems more appropriate to me. We could also rely on it to customize the E2E tests manifests.

barkbay:
The tool would need to take into account the cluster role needed to create `subjectaccessreviews` in the context of #2468 , see https://github.com/elastic/cloud-on-k8s/pull/2482#discussion_r372829525

pebrc:
We discussed offline:
* let's explore how far we can get with `kustomize` 
* it  has the benefit of being integrated into recent versions of `kubectl` 
* it would allow use to keep a similar installation experience as we have it now

sebgl:
A good first step could be to document how to tweak the existing manifests. It came as a question again: https://discuss.elastic.co/t/difficulty-installing-eck-in-a-single-namespace/225891.

pebrc:
Regarding this discuss forum comment. If the configuration the Makefile in `config` spits out is invalid we should maybe just fix that?

sebgl:
> Regarding this discuss forum comment. If the configuration the Makefile in config spits out when you run it with make generate-namespace is invalid we should maybe just fix that?

That's also an option yes.

I would be in favor of removing all those files that we have a hard time maintaining in sync with each other. I thought we already did that actually when [removing the namespace/global operators concept](https://github.com/elastic/cloud-on-k8s/issues/2254#issue-537074434).

pebrc:
Just tested on a fresh k8s cluster and our single namespace make target seems to work fine

```
[...]

Issue #2254, opened 2019-12-12
Deprecate the "global + namespace" operator concept


sebgl:
Relates https://github.com/elastic/cloud-on-k8s/issues/374.

We are moving towards not requiring the concepts of "global" and "namespace" operators anymore. The enterprise license can be applied to "any" operator that has the `license` role.

Without this concept, the operator can still:
- be configured with multiple roles
- be deployed in a particular namespace
- manage resources in one namespace, all namespaces, N namespaces

Maintaining all-in-one + global + namespace manifests in our repository is confusing.
We should remove those manifests to only have the all-in-one version (that users can customize to their needs).
Both are used in E2E tests, that should be adapted accordingly.

It should have no particular impact on backward-compatibility, since these are just yaml manifests pre-built in a particular way. The same configuration can still be achieved by customizing the all-in-one manifest.

Also make sure READMEs, documentation and ADRs are updated accordingly.

We need to keep a toggle for the webhook as we need to turn it off in environments where multiple instances of the operator are deployed (OLM with per namespace operators)

pebrc:
I just got a case of a user inadvertently configuring the licensing functionality away by running with the `namespace` role only. While this issue seems to mainly cover the generated manifest we should also consider removing the role flags and keep just a separate flag for the webhook.  It is worth noting that this would introduce an incompatible change with regards to the operator startup parameters which we would need to call out in documentation for users using handcrafted manifests.

PR #2457, opened 2020-01-22
Logging: add minimal support for ECS

barkbay:
This PR adds a primitive support for ECS by adding the following fields:

* log.level
* log.logger
* service.type
* service.version

It also adds the required processor as an annotation to avoid a conflict on the `error` key/document and generates the following ECS fields:

* error.message
* error.stack_trace

I think that this PR can be considered as a fix for #2002, however as pointed out in https://github.com/elastic/cloud-on-k8s/issues/2002#issuecomment-576700627 I would open at least 2 others issues to track some additional improvements that can be done to improve our support of ECS.

### DIFF
--- a/config/e2e/global_operator.yaml
+++ b/config/e2e/global_operator.yaml
@@ -158,6 +158,8 @@ spec:
   template:
     metadata:
       annotations:
+        # Rename the fields "error" to "error.message" and "source" to "event.source"
+        # This is to avoid a conflict with the ECS "error" and "service" documents.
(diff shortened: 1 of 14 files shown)


config/e2e/global_operator.yaml
```diff
@@ -157,6 +157,8 @@ spec:
   serviceName: {{ .GlobalOperator.Name }}
   template:
     metadata:
+      annotations:
+        "co.elastic.logs/raw": "[{\"type\":\"container\",\"json.keys_under_root\":true,\"paths\":[\"/var/log/containers/*${data.kubernetes.container.id}.log\"],\"processors\":[{\"convert\":{\"mode\":\"rename\",\"ignore_missing\":true,\"fields\":[{\"from\":\"error\",\"to\":\"_error\"}]}},{\"convert\":{\"mode\":\"rename\",\"ignore_missing\":true,\"fields\":[{\"from\":\"_error\",\"to\":\"error.message\"}]}},{\"convert\":{\"mode\":\"rename\",\"ignore_missing\":true,\"fields\":[{\"from\":\"source\",\"to\":\"_source\"}]}},{\"convert\":{\"mode\":\"rename\",\"ignore_missing\":true,\"fields\":[{\"from\":\"_source\",\"to\":\"event.source\"}]}}]}]"
```
david-kow: Is there a way we could have this in a single place?

barkbay: I was wondering the same. I'm not sure it is easy since the manifests are handled by different templating systems.

sebgl: This can be delayed to when we decide to handle https://github.com/elastic/cloud-on-k8s/issues/2254 and/or https://github.com/elastic/cloud-on-k8s/issues/2406.

sebgl: Can you comment (on all 5 files) what this is doing?

PR #2460, opened 2020-01-22
Reuse the release job for nightly builds

### LINKED ISSUE #1873: Use the same job for release and nightly builds
Currently we are using 2 different jobs for release and nightly builds. 
https://devops-ci.elastic.co/view/cloud-on-k8s/job/cloud-on-k8s-nightly/
https://devops-ci.elastic.co/view/cloud-on-k8s/job/cloud-on-k8s-release/

As result we may have issues related only to release job because it's triggered rarely. To fix that it make sense to have 1 job for both release and nightly builds

thbkrkr:
I tried to simplify the existing a bit without making a big-bang. I feel we should avoid having too much logic in the Jenkinsfile because there are too many layers and indirections between it and the root Makefile (Jenkinsfile -> build/ci/Makefile ci -> build/ci/Makefile ci-internal (to be in a container) -> and finally the root Makefile).

Here is my proposal:
- Add a `setenv` script to prepare the environment variables expected to run the `ci-release` make target. This is the first step to simplify the Jenkinsfiles. In the long run all jobs could use it. 
- Move the check of the existence of the Docker image which is about to be built from the Jenkinsfile to the Makefile

Some questions:
- Do we want to report the step `Run unit and integration tests` in the release job (this takes about 5-6min and this makes the release dependent on gke)?
- Do we want to upload the YAML all-in-one manifest for the nightly builds? If yes, probably in another directory than `.../downloads/eck/`?

Resolves #1873


build/ci/release/Jenkinsfile
```diff
@@ -12,53 +12,41 @@ pipeline {
...
                     steps {
                         script {
-                            def msg = "${OPERATOR_IMAGE} was pushed \r\n" +
-                                "https://download.elastic.co/downloads/eck/${TAG_NAME}/all-in-one.yaml was uploaded \r\n" +
-                                "Congratulations!"
                             slackSend botUser: true,
                                 channel: '#cloud-k8s',
                                 color: 'good',
-                                message: msg,
+                                message: "`${TAG_NAME}` was pushed \r\n" +
```
barkbay:
```suggestion
                                message: "`${TAG_NAME}` was released \r\n" +
```

build/ci/release/Jenkinsfile
```diff
@@ -12,53 +12,41 @@ pipeline {
...
+                        buildingTag()
+                    }
                     steps {
                         sh 'make -C build/ci yaml-upload'
                     }
                 }
-                stage('Send message to Slack') {
+                stage('Notify successful release build') {
+                    when {
+                        buildingTag()
```
david-kow: Is this built in? I don't see where it is defined.

build/ci/release/Jenkinsfile
```diff
@@ -12,53 +12,41 @@ pipeline {
...
                         script {
-                            def msg = "${OPERATOR_IMAGE} was pushed \r\n" +
-                                "https://download.elastic.co/downloads/eck/${TAG_NAME}/all-in-one.yaml was uploaded \r\n" +
-                                "Congratulations!"
                             slackSend botUser: true,
                                 channel: '#cloud-k8s',
                                 color: 'good',
-                                message: msg,
+                                message: "`${TAG_NAME}` was pushed \r\n" +
+                                    "https://download.elastic.co/downloads/eck/" + TAG_NAME + "/all-in-one.yaml was uploaded \r\n" +
```
david-kow: I'm not sure, but:
```suggestion
                                    "https://download.elastic.co/downloads/eck/" + $TAG_NAME + "/all-in-one.yaml was uploaded \r\n" +
```

build/ci/release/Jenkinsfile
```diff
@@ -12,53 +12,41 @@ pipeline {
         VAULT_ADDR = credentials('vault-addr')
         VAULT_ROLE_ID = credentials('vault-role-id')
         VAULT_SECRET_ID = credentials('vault-secret-id')
-        REPOSITORY = "eck"
-        IMG_NAME = "eck-operator"
-        REGISTRY = "docker.elastic.co"
-        OPERATOR_IMAGE = "${REGISTRY}/${REPOSITORY}/${IMG_NAME}:${TAG_NAME}"
+
+        TAG_NAME = sh(script: 'echo -n $TAG_NAME', returnStdout: true)
```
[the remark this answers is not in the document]
thbkrkr: This is a way to try to read the variable in case it is not defined.

The `TAG_NAME` environment variable is present only when the job is built from a commit corresponding to a tag.
We use this TAG_NAME when we build a release as the version.
In the case of a nightly build, this variable does not exist. If we try to directly use the variable in this context, we got: `groovy.lang.MissingPropertyException: No such property: TAGNAME for class: groovy.lang.Binding at groovy.lang.Binding.getVariable(Binding.java:63)`.

build/ci/release/Jenkinsfile
thbkrkr: Yes, it's a built-in condition of the declarative pipeline: https://jenkins.io/doc/book/pipeline/syntax/#built-in-conditions

build/ci/release/Jenkinsfile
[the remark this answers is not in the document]
david-kow: Ah, got it. I think this might deserve a comment.

build/ci/release/Jenkinsfile
thbkrkr: It will not work. Here, `TAG_NAME` is the groovy variable defined in the above environment directive.
On the other hand, we can use string interpolation to be consistent with the line above `"${TAG_NAME} was released \r\n"`:

` "https://download.elastic.co/downloads/eck/${TAG_NAME}/all-in-one.yaml was uploaded \r\n"`

thbkrkr:
I adressed almost all your feedbacks. Important changes:
- I brought back the `Run checks` and `Run unit and integration tests` stages that where in the `nightly/Jenkinsfile`. For that, `setenv` now takes a parameter (`test` or `release`).
- We are now always uploading the YAML manifest to S3. To make this works for a nightly build, we are now using the `VERSION` env var written previously in the `.env` file by `setenv`.
- `setenv` works in local. Note: it is sensitive to the path from which it is executed.

> It would be also great to have something idempotent that can be run again if needed (e.g. docker image was built and uploaded but upload of the manifest failed).

Yes, that would be great. To make the yaml upload idempotent, I think we can try to download the yaml file first, compute its sha1 and compare it with the original before uploading it.

.ci/jobs/nightly-build.yml
```diff
@@ -12,5 +12,5 @@
             branches:
               - master
             credentials-id: 'f6c7695a-671e-4f4f-a331-acdce44ff9ba'
-      script-path: build/ci/nightly/Jenkinsfile
+      script-path: build/ci/release/Jenkinsfile
```
barkbay: I'm wondering if we should not add a comment here (or somewhere else) to explain that this job is using the "release" Jenkinsfile on purpose and that the nightly and the release jobs should be as much as possible kept "in sync".

build/ci/setenv
```diff
@@ -0,0 +1,84 @@
...
+
+# Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+# or more contributor license agreements. Licensed under the Elastic License;
+# you may not use this file except in compliance with the Elastic License.
+
+# Script to set environment variables to run CI tasks.
+#
+# Usage: setenv CONTEXT
+#
+#   CONTEXT      test or release (test by default)
```
barkbay: What is the purpose of the `test` context ? When should it be used ?

.ci/jobs/nightly-build.yml
thbkrkr: +1 to add a note.

It may also be an indication that the naming is not good. It could be `build/ci/build/Jenkinsfile` used by `nightly-build.yml` and `release-build.yml`.

build/ci/setenv
thbkrkr: `setenv test` should be called to run (unit or integ) tests related ci tasks and `setenv release` to run build (or e2e tests) related ci tasks. Having slept on this, it sounds like a potential source of error. I will look for a better solution.

thbkrkr: `setenv test` sets `GCLOUD_PROJECT` (create/delete k8s clusters).
`setenv release` sets `GO_TAGS`, `LICENSE_PUBKEY` (build the operator binary with the public license key) and `ELASTIC_DOCKER_LOGIN` (docker push ci/e2e/operator images).

We have 9 Jenkinsfile:
* build/ci/support/cleanup.jenkinsfile
* build/ci/pr/Jenkinsfile
* build/ci/e2e/custom_operator_image.jenkinsfile
* build/ci/e2e/GKE_k8s_versions.jenkinsfile
* build/ci/e2e/Jenkinsfile-aks
* build/ci/e2e/Jenkinsfile
* build/ci/e2e/stack_versions.jenkinsfile
* build/ci/e2e/vanilla_k8s.jenkinsfile
* build/ci/release/Jenkinsfile

8/9 `Jenkinsfile` use `GCLOUD_PROJECT`. The only one which doesn't need it is `e2e/Jenkinsfile-aks` for now.

7/9 build the operator with the pub key to run e2e tests or do a release.

.ci/jobs/nightly-build.yml
thbkrkr: I renamed and added a note.

build/ci/setenv
thbkrkr: I removed the notion of context by decreasing the number of variable to set in 659f280.

thbkrkr: @pebrc Can you take a look in particular at 659f280 as it is a little sensitive? 

In a effort to simplify the Jenkinsfile, I think we can:
* move the declaration of `GO_TAGS ?= release` in the make targets `ci-e2e` and `ci-release`
* move the writing of the `export LICENSE_PUBKEY` in the `.env` file in the make target `get-elastic-public-key` that fetchs the license pub key (same for the test licenses)

thbkrkr: `GO_TAGS ?= release` in the make targets `ci-e2e` breaks the pr job.

The pr job is the only place where `ci-e2e` is used without the need to set `GO_TAGS` to `release` because to run the TestSmoke we don't need anything related to the licenses.

thbkrkr: In the same spirit of moving some logic outside the Jenkinsfile, we can add a dedicated make target `ci-e2e-smoketest` that will not have `GO_TAGS ?= release` and instead `TESTS_MATCH = TestSmoke`.

77c044f

.ci/jobs/nightly-build.yml
```diff
@@ -1,7 +1,7 @@
 ---
 - job:
-    description: Job that build nightly images of ECK. This Job is managed by JJB.
-    name: cloud-on-k8s-nightly
+    description: Job that builds nightly version of ECK. This Job is managed by JJB.
```
pebrc:
```suggestion
    description: Job that builds a nightly version of ECK. This Job is managed by JJB.
```

build/ci/pr/Jenkinsfile
```diff
@@ -127,8 +130,6 @@ void createConfig() {
         cat >.env <<EOF
 REGISTRY = eu.gcr.io
 REPOSITORY = $GCLOUD_PROJECT
-TESTS_MATCH = TestSmoke
-SKIP_DOCKER_COMMAND = false
```
david-kow: Can we just remove this everywhere?

build/ci/pr/Jenkinsfile
```diff
@@ -13,6 +13,9 @@ pipeline {
         VAULT_ROLE_ID = credentials('vault-role-id')
         VAULT_SECRET_ID = credentials('vault-secret-id')
         GCLOUD_PROJECT = credentials('k8s-operators-gcloud-project')
+        REGISTRY = "eu.gcr.io"
```
david-kow: My problem with that is that to repro this job I need to copy each of those vars, export them and run `./setenv`. Then find deployer-config snipper and run it too. 
Now it's only copy paste a single snippet (.env + deployer-config) from the code. 

Could we just move this to `createConfig` and invoke `./setenv` there too?

build/ci/Makefile
```diff
@@ -22,6 +23,9 @@ endif
 
 CI_IMAGE ?= docker.elastic.co/eck/eck-ci:$(shell md5sum $(ROOT_DIR)/go.mod $(ROOT_DIR)/build/ci/Dockerfile | awk '{print $$1}' | md5sum | awk '{print $$1}')
 
+# Set environment variables in a .env file for the root Makefile
+$(shell ./setenv)
```
david-kow: I think I'd keep that in each Jenkins file, so that entire setup is confined to a single file.

build/ci/Makefile
```diff
@@ -41,31 +45,40 @@ ci-internal: ci-build-image
...
 		DOCKER_OPTS="-e AWS_ACCESS_KEY_ID=$(AWS_ACCESS_KEY_ID) -e AWS_SECRET_ACCESS_KEY=$(AWS_SECRET_ACCESS_KEY)" \
-		DOCKER_CMD="aws s3 cp $(GO_MOUNT_PATH)/$(ALL_IN_ONE_OUTPUT_FILE) \
-		s3://download.elasticsearch.org/downloads/eck/$(TAG_NAME)/all-in-one.yaml" ci-internal
+		DOCKER_CMD="aws s3 ls $(YAML_DST) && echo OK: $(YAML_DST) already uploaded || aws s3 cp $(YAML_SRC) $(YAML_DST)" \
+		ci-internal
 
 # reads Elastic public key from Vault into license.key
 get-elastic-public-key:
 	@ VAULT_TOKEN=$(VAULT_TOKEN) vault read -address=$(VAULT_ADDR) -field=pubkey secret/devops-ci/cloud-on-k8s/license | base64 --decode > license.key
+	@ echo export LICENSE_PUBKEY = /go/src/github.com/elastic/cloud-on-k8s/build/ci/license.key >> .env
```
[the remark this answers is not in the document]
thbkrkr: No. Good catch.

build/ci/pr/Jenkinsfile
thbkrkr: Yes.

thbkrkr:
These modifications are good to merge the release and the nightly Jenkinsfile but a little problematic for the others. As soon as you pull on a wire, it moves on the other side.

I said in the top description of this PR: "I tried to simplify the existing a bit without making a big-bang". 
I tried very hard. Now I think it's not possible to have a simple and clean declarative CI without a more deep refactoring.

I also realized that to improve the existing, there is lots of names to standardize and things to homogenize and align. 
Doing that highlighted unused variables or small inconsistencies.
I showed a quick preview to @david-kow and @sebgl  and they are +1.

I think it would be a shame to stop there. So, I'm gonna close this PR and create a new one to list the different changes and ease the review.

Sorry to have used some of your brain cycles without knowing exactly where I was going.

PR #2462, opened 2020-01-23
Add basic APM agent instrumentation

pebrc:
* Trace all requests via the Elasticsearch client. 
* Create spans for significant pieces of logic in the controllers.
* Error capture is very basic and centralised in the `Results` abstraction we use.  I used `github.com/pkg/errors` where possible to capture stack traces which facilitate error tracing without the need to have tracing information/contexts at every call site. For controllers that don't use the `Results` struct I tried to keep the error capture at the top level. This might be somewhat counter intuitive as we want to capture the errors when they are created and not somewhere up to keep as much context as possible. My rationale was that for errors we create with `githubcom/pkg/errors` we already have the context so the place of capture is secondary, for errors coming from other libraries we don't have full context anyway (unless they use a similar form of stack trace capture) so reducing the number of places to instrument by keeping the capture statements on the top level seemed a good compromise (happy to reconsider if people feel strongly about it)
* k8s client requests are not traced because we use a cached client, which does not do any requests when we interact with it on reads (it ignores any context passed). We can look into it for writes but that would mean undoing at least partially our own wrapping layer which hides contexts from calling code.

The APM agent API is build in such a way that it assumes all interactions to originate from a `context.Context` object, I had to pass context objects from calling code into the our reconciliation logic in a few places and replaced various context creations (`context.Background()`) with those contexts, which seems like the right thing to do anyway even though it is fairly intrusive

To discuss: 
* granularity of the spans
* less intrusive options for tracing 
* some choices around passing the tracer with the context object (following the APM agent practice here) 

Screenshot 2020-01-23 at 11 28 26


cmd/manager/main.go
```diff
@@ -13,13 +13,13 @@ import (
...
-	// allow gcp authentication
-	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
-
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 	"k8s.io/client-go/kubernetes"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
+
+	// allow gcp authentication
+	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
```
pebrc: My editor thinks this should sit here 🤷‍♂

pkg/controller/elasticsearch/observer/observer.go
```diff
@@ -131,6 +135,12 @@ func (o *Observer) retrieveState(ctx context.Context) {
 	timeoutCtx, cancel := context.WithTimeout(ctx, o.settings.RequestTimeout)
 	defer cancel()
 
+	if o.tracer != nil {
+		tx := o.tracer.StartTransaction(o.cluster.String(), "elasticsearch_async")
```
barkbay: nit: s/elasticsearch_async/elasticsearch_observer ?

pkg/controller/kibanaassociation/association_controller.go
```diff
@@ -127,41 +130,53 @@ func (r *ReconcileAssociation) onDelete(obj types.NamespacedName) error {
...
 	}
 
-	newStatus, err := r.reconcileInternal(&kibana)
+	newStatus, err := r.reconcileInternal(ctx, &kibana)
 	if err != nil {
 		k8s.EmitErrorEvent(r.recorder, err, &kibana, events.EventReconciliationError, "Reconciliation error: %v", err)
 	}
 
 	// maybe update status
+	result, err2 := r.updateStatus(ctx, kibana, newStatus)
```
barkbay:
```suggestion
	if result, err := r.updateStatus(ctx, kibana, newStatus); err != nil || !reflect.DeepEqual(result, reconcile.Result{}) {
		return result, commonapm.CaptureError(ctx, err)
	}
```

pkg/controller/apmserverelasticsearchassociation/apmserverelasticsearchassociation_controller.go
```diff
@@ -135,17 +138,19 @@ func (r *ReconcileApmServerElasticsearchAssociation) onDelete(obj types.Namespac
 // and what is in the ApmServerElasticsearchAssociation.Spec
 func (r *ReconcileApmServerElasticsearchAssociation) Reconcile(request reconcile.Request) (reconcile.Result, error) {
 	defer common.LogReconciliationRun(log, request, "as_name", &r.iteration)()
+	tx, ctx := commonapm.NewTransaction(r.Tracer, request.NamespacedName, "apm_association")
```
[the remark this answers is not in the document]
pebrc: I did not do that because the controller names are quiet long and would look weird in the APM UI. Also the word `controller` is somewhat incorrect for an APM transaction, if anything it should be called something like `*-reconciliation` or so . But I harmonized the naming to follow the controller names minus `-controller`. Let me know if you have concerns with that.

barkbay:
> k8s client requests are not traced because we use a cached client, which does not do any requests when we interact with it on reads (it ignores any context passed). We can look into it for writes but that would mean undoing at least partially our own wrapping layer which hides contexts from calling code.

I guess it's fine as long as we want some "trends", e.g. if `reconcile_deployment` increases the duration of the transactions then we don't know:
* if it is a specific resources that took some time to reconcile
* if it's related to the API server or not

Maybe a dedicated span for [ReconcileResource](https://github.com/elastic/cloud-on-k8s/blob/master/pkg/controller/common/reconciler/reconciler.go#L70) would help without being too intrusive re. our client wrapper ? It would not cover all the cases but would help to detect when the API Server becomes too slow to respond _(even if we are using a cache)_?

No strong opinion though, I'm unfamiliar with APM tools  _(it makes me wondering it would make sense to have this PR also reviewed by someone from the monitoring team)_

pkg/controller/apmserver/certificates/reconcile.go
```diff
@@ -20,12 +23,15 @@ import (
...
+	ctx context.Context,
 	driver driver.Interface,
 	apm *apmv1.ApmServer,
 	services []corev1.Service,
 	rotation certificates.RotationParams,
-) reconciler.Results {
-	results := reconciler.Results{}
+) *reconciler.Results {
+	span, _ := apmagent.StartSpan(ctx, "reconcile_certs", commonapm.SpanTypeApp)
+	defer span.End()
```
sebgl: As a general readability improvement, I think we should include a newline here by default.

pkg/controller/common/apm/tracer.go
```diff
@@ -0,0 +1,23 @@
...
+	"go.elastic.co/apm"
+)
+
+// NewTracer returns a new APM tracer with the logger in log configured.
+func NewTracer(serviceName string, log logr.Logger) *apm.Tracer {
+	build := about.GetBuildInfo()
+	tracer, err := apm.NewTracer(serviceName, build.Version+"-"+build.Hash)
+	if err != nil {
+		// don't fail the application because tracing fails
+		log.Error(err, "failed to created tracer for "+serviceName)
```
david-kow: What is the way of surfacing any other issues with tracing, ie. bad/missing config? If I never received or stop receiving apm data - how I can detect it and debug it?

cmd/manager/main.go
```diff
@@ -13,13 +13,15 @@ import (
 	"strings"
 	"time"
 
-	// allow gcp authentication
-	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
-
+	commonapm "github.com/elastic/cloud-on-k8s/pkg/controller/common/apm"
```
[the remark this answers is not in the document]
pebrc: Good idea. I like `tracing`

cmd/manager/main.go
```diff
@@ -151,6 +155,10 @@ func init() {
 		"localhost:6060",
 		"Listen address for debug HTTP server (only available in development mode)",
 	)
+	Cmd.Flags().Bool(
+		EnableAPMFlag,
+		false,
+		"Enable APM tracing in the operator. Configure endpoint, token etc per environment variables")
```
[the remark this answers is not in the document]
pebrc: Rephrased and added a link to the go-agent docs

pkg/controller/common/apm/tracer.go
pebrc: Assuming we get past this line of code, one line below we are setting a logger into the tracer which reports when data is transmitted to the APM server but will  also report failures: 

```
{"log.level":"info","@timestamp":"2020-01-30T12:46:08.617Z","log.logger":"manager","message":"request failed: request failed with 401 Unauthorized: {\"error\":\"invalid token\"} (next request in ~1s)","service.version":"1.0.0-5791afdd","service.type":"eck","ecs.version":"1.4.0"}                                                                                                                                                   
```

I will change the logger though so that it becomes obvious this error relates to tracing.

pkg/controller/common/tracing/log.go
```diff
@@ -0,0 +1,34 @@
...
+
+	"github.com/go-logr/logr"
+	pkgerrors "github.com/pkg/errors"
+	"go.elastic.co/apm"
+)
+
+// NewLogAdapter returns an implementation of the log interface expected by the APM agent.
+func NewLogAdapter(log logr.Logger) apm.Logger {
+	return &logAdapter{
+		log: log,
```
axw: maybe `log.WithName("tracing")`?

pkg/controller/apmserver/apmserver_controller.go
```diff
@@ -416,9 +421,13 @@ func (r *ReconcileApmServer) deploymentParams(
 }
 
 func (r *ReconcileApmServer) reconcileApmServerDeployment(
+	ctx context.Context,
 	state State,
 	as *apmv1.ApmServer,
 ) (State, error) {
+	span, _ := apm.StartSpan(ctx, "reconcile_deployment", tracing.SpanTypeApp)
```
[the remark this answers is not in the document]
pebrc: Our linter is quite strict about assigning stuff to variables and then not using them. But will change the places where it is possible.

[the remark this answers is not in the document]
axw: Ah, makes sense now! Fair enough.

pkg/controller/common/tracing/log.go
pebrc: I would address that probably in a follow up PR with https://github.com/elastic/cloud-on-k8s/issues/2491

axw: OK, but just to be clear I'm just suggesting that the tracer's internal logging should probably have a different logger name associated. Nothing to do with log correlation.

pkg/controller/elasticsearch/driver/esstate.go
```diff
@@ -33,12 +33,12 @@ type MemoizingESState struct {
 }
 
 // NewMemoizingESState returns an initialized MemoizingESState.
-func NewMemoizingESState(esClient esclient.Client) ESState {
+func NewMemoizingESState(ctx context.Context, esClient esclient.Client) ESState {
```
axw: Again I think the context should not be stored, but passed into the `NodesInCluster`, `ShardAllocationsEnabled`, and `Health` methods. Those would pass the context along to initialize by creating a closure.

It probably doesn't matter greatly, but it'll allow you to attribute the memoized ES requests to the correct parent span. Depends on how much you care vs. how much work is involved in updating call sites.

sebgl: I think I agree that would be a more common thing to do.

pebrc: I can change it.

pkg/controller/common/tracing/log.go
pebrc: I use a package level logger called "tracing" for it at the moment.

axw: :man_facepalming: so you do, sorry. I made the comment based on the output in https://github.com/elastic/cloud-on-k8s/pull/2462#discussion_r372928203 (which shows `"log.logger": "manager"`). I suppose that's out of date.

pkg/controller/elasticsearch/driver/esstate.go
pebrc: I am having second thoughts, because this is a cached client abstraction we are using here, so repeated calls with different context would not have any effect

Issue #2203, opened 2019-12-02
Support for remote ES clusters within a single k8s cluster


pebrc:
https://github.com/elastic/cloud-on-k8s/issues/963 removed support for two CRDs we used initially to facilitate remote cluster support. We want to reinstate support for automatic setup of remote cluster relationships within one k8s cluster (setup of remote clusters spanning multiple k8s clusters or ES clusters hosted outside of k8s are beyond the scope of what ECK can do)

```yaml
apiVersion: elasticsearch.k8s.elastic.co/v1beta1
kind: Elasticsearch
metadata:
  name: elasticsearch-sample
spec:
  version: 7.5.0
  nodeSets:
  - name: default
    config:
      node.master: true
      node.data: true
      node.ingest: true
      node.ml: true
    count: 3
  remoteClusters:
  - name: my-other-cluster
    namespace: my-custom-ns
  - name: yet-another-cluster
    # namespace: default # can be omitted
```

* ECK would configure the CA of `elasticsearch-sample` as one of the CAs in `xpack.security.transport.ssl.certificate_authorities` for `my-other-cluster` and `yet-another-cluster`
* ECK would configure the CA's of `my-other-cluster` and `yet-another-cluster` as two of the trusted CAs for `elasticsearch-sample` in `xpack.security.transport.ssl.certificate_authorities`
* ECK would create `my-other-cluster` and `yet-another-cluster` as [remote clusters](https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-remote-clusters.html) in `elasticsearch-sample`

Questions:
* responsibilities and conflicts: are we re-introducing a remote cluster controller for this, that watches subsections of the Elasticseach resource (the `remoteClusters` attribute effectively), or is this going to be reconciled by the regular Elasticsearch controller? 
* where do we persist the additional CA's on "the other side" (in this example  `my-other-cluster` and `yet-another-cluster`)? Annotation, config map or another attribute on the CRD? If we just edit the other CRD to add the CA to the `config` section how do we distinguish between user created manual config and operator created config. We have some precedence for an [annotation driven config ](https://github.com/elastic/cloud-on-k8s/blob/3326d27df95d2393eadcfa068a016f01cff18ae4/pkg/controller/common/association/conf.go)mechnism in APM and Kibana controllers, which seems most consistent to follow here too.

anyasabo:
[...]

PR #2466, opened 2020-01-27
Add note about Kibana accessing ESs in other namespaces

david-kow:
Came up when discussing https://github.com/elastic/cloud-on-k8s/issues/2203 RBAC concerns.

### DIFF
--- a/docs/kibana.asciidoc
+++ b/docs/kibana.asciidoc
@@ -38,7 +38,7 @@ spec:
 
 NOTE: `namespace` is optional if the Elasticsearch cluster is running in the same namespace as Kibana.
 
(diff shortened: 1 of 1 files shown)


docs/kibana.asciidoc
```diff
@@ -38,6 +38,8 @@ spec:
 
 NOTE: `namespace` is optional if the Elasticsearch cluster is running in the same namespace as Kibana.
 
+NOTE: Any Kibana can reference (and thus access) any Elasticsearch instance as long as they both are in namespaces that are watched by the same ECK instance. ECK will copy the required `Secret` from Elasticsearch to Kibana namespace. If this behavior is not desired, more than one ECK instance can be deployed. Kibana won't be able to automatically (through `elasticsearchRef`) connect to Elasticsearch in a namespace managed by different ECK instance.
```
pebrc:
```suggestion
NOTE: Any Kibana can reference (and thus access) any Elasticsearch instance as long as they both are in namespaces that are watched by the same ECK instance. ECK will copy the required `Secret` from Elasticsearch to Kibana namespace. If this behavior is not desired, more than one ECK instance can be deployed. Kibana won't be able to automatically connect to Elasticsearch (through `elasticsearchRef`) in a namespace managed by a different ECK instance.
```

Issue #2461, opened 2020-01-23
E2E Cluster creation is unstable


barkbay:
https://devops-ci.elastic.co/blue/organizations/jenkins/cloud-on-k8s-stack/detail/cloud-on-k8s-stack/162/pipeline/36

![image](https://user-images.githubusercontent.com/976373/72966548-7fc6bc00-3dbf-11ea-834f-ca88bfad672d.png)

Only the creation of the `6.8.x` cluster has succeeded. All others have failed with the following errors:

* 7.1.1
```
[2020-01-23T00:50:00.464Z] ERROR: (gcloud.beta.container.clusters.create) Operation [<Operation
[2020-01-23T00:50:00.465Z]  clusterConditions: []
[2020-01-23T00:50:00.465Z]  detail: u'Retry budget exhausted (80 attempts): The network "default" does not have available private IP space in 10.0.0.0/8 to reserve a /14 block for pods for cluster {europe-west2, 1014491842772, ****, eck-71-162-e2e, 83cb6be5dc25bfe91a9e468b8975ebc9dbc63b3c6635d033d3aca7156ea6570a}.'
[2020-01-23T00:50:00.465Z]  endTime: u'2020-01-23T00:49:54.093175723Z'
[2020-01-23T00:50:00.465Z]  name: u'operation-1579738810624-f03fae70'
[2020-01-23T00:50:00.465Z]  nodepoolConditions: []
[2020-01-23T00:50:00.465Z]  operationType: OperationTypeValueValuesEnum(CREATE_CLUSTER, 1)
[2020-01-23T00:50:00.465Z]  progress: <OperationProgress
[2020-01-23T00:50:00.465Z]  metrics: [<Metric
[2020-01-23T00:50:00.465Z]  intValue: 12
[2020-01-23T00:50:00.465Z]  name: u'CLUSTER_CONFIGURING'>, <Metric
[2020-01-23T00:50:00.465Z]  intValue: 12
[2020-01-23T00:50:00.465Z]  name: u'CLUSTER_CONFIGURING_TOTAL'>]
[2020-01-23T00:50:00.465Z]  stages: []>
[2020-01-23T00:50:00.465Z]  selfLink: u'https://container.googleapis.com/v1beta1/projects/1014491842772/locations/europe-west2/operations/operation-1579738810624-f03fae70'
[2020-01-23T00:50:00.465Z]  startTime: u'2020-01-23T00:20:10.624807528Z'
[...]

PR #2473, opened 2020-01-27
Override GKE Pods CIDR IP range to allow more clusters to be created

sebgl:
For each k8s cluster, GKE creates 2 additional secondary IP ranges:
- a /14 range for Pods (262k Pods per cluster)
- a /20 range for Services (4094 Services per cluster)

The downside of using the default /14 range is that there are only 62
possible /64 ranges that can be created.
Overriding it to /20 allows up to 4094 of those subnets to be defined instead,
with up to 4094 Pods per cluster.

I hope this mitigates the IP range hard limit we sometimes hit against
our default GKE VPC.

Both Pods CIDR range and Services CIDR range can now be overridden through
the gke settings in the deployer.

Should help with https://github.com/elastic/cloud-on-k8s/issues/2461.

### DIFF
--- a/hack/deployer/runner/gke.go
+++ b/hack/deployer/runner/gke.go
@@ -60,6 +60,8 @@ func (gdf *GkeDriverFactory) Create(plan Plan) (Driver, error) {
 			"LocalSsdCount":     plan.Gke.LocalSsdCount,
 			"GcpScopes":         plan.Gke.GcpScopes,
 			"NodeCountPerZone":  plan.Gke.NodeCountPerZone,
+			"ClusterIPv4CIDR":   plan.Gke.ClusterIPv4CIDR,
+			"ServicesIPv4CIDR":  plan.Gke.ServicesIPv4CIDR,
 		},
 	}, nil
 }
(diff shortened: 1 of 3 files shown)


hack/deployer/runner/gke.go
```diff
@@ -167,7 +169,15 @@ func (d *GkeDriver) create() error {
...
-		pspOption = " --enable-pod-security-policy"
+		pspOption = "--enable-pod-security-policy "
+	}
+
+	cidrOptions := ""
+	if d.plan.Gke.ClusterIPv4CIDR != "" {
+		cidrOptions += fmt.Sprintf("--cluster-ipv4-cidr=%s ", d.plan.Gke.ClusterIPv4CIDR)
+	}
+	if d.plan.Gke.ServicesIPv4CIDR != "" {
+		cidrOptions += fmt.Sprintf("--services-ipv4-cidr=%s ", d.plan.Gke.ServicesIPv4CIDR)
```
thbkrkr: Nit: `strings.Join` may be nicer and safer to handle spaces in this string concatenation.

```go
baseCmd := "x new --a --b=c"

opts := []string{}
opts = append(opts, "--enable-pod-security-policy")
if true {
	opts = append(opts, "--yy=zz")
}

cmd := fmt.Sprintf("%s %s", baseCmd, strings.Join(opts, " "))
```

PR #2475, opened 2020-01-28
Upgrade k8s client to avoid overriding requeue after

racevedoo:
This PR upgrades the k8s client to include @barkbay's fix(https://github.com/kubernetes/kubernetes/pull/83945), removing a hard override of `RequeueAfter`.

Not sure we are safe to use `0.17.2` yet, for compatibility, but I guess you guys can confirm this.


sebgl:
Closing as per the comments above.
I hope that makes sense to you too @racevedoo. Thanks for opening this PR though!

Issue #2463, opened 2020-01-24
Add the --cert-dir flag to change the directory for Webhook certificate.


dulltz:
## Proposal
Add the `--cert-dir` flag to pass to the controller-runtime.

In PodSecurityPolicy enabled K8s cluster, read-only root filesystem is common.
I want to to use `/tmp` path as writable empty-dir because sometimes ECK writes its logs to under the /tmp/.

However, ECK v1.0.0 uses `/tmp/k8s-webhook-server/serving-certs` for the elastic-webhook-server-cert secret.

It is not good to overwrap mount paths, so I hope to add `--cert-dir` flag to ECK for changing the cert-dir.

It seems to be changeable by passing to controller-runtime.
https://github.com/kubernetes-sigs/controller-runtime/blob/a7c8a93c1cf395974911ef0ece977a7d27b2bf4b/pkg/manager/manager.go#L175

### Environment

- ECK version: v1.0.0
- Kubernetes information:
    - v1.16
    - On premise
    - PodSecurityPolicy enabled

```
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"16+", GitVersion:"v1.16.6-beta.0", GitCommit:"e7f962ba86f4ce7033828210ca3556393c377bcc", GitTreeState:"clean", BuildDate:"2020-01-15T08:26:26Z", GoVersion:"go1.13.5", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"16+", GitVersion:"v1.16.6-beta.0", GitCommit:"e7f962ba86f4ce7033828210ca3556393c377bcc", GitTreeState:"archive", BuildDate:"2020-01-17T02:54:55Z", GoVersion:"go1.13.4", Compiler:"gc", Platform:"linux/amd64"}
```

sebgl:
It sounds like a legitimate request 👍 
We probably want to make the flag name a bit more explicit to contrast with other flags and other certificates managed by ECK: `--webhook-cert-dir`.

I have a few remarks and questions, to make sure I understand completely your use case:

* If overriding the `--webhook-cert-dir` flag, you would also have to tweak the secret volume mount path in the ECK manifest. It is currently set to `mountPath: /tmp/k8s-webhook-server/serving-certs` by default.

* `sometimes  ECK writes its logs to under the /tmp/.`. The this surprises me a little bit. Can you give more details? When does ECK write its logs to `/tmp`?

[...]

PR #2476, opened 2020-01-28
Allow webhook cert directory to be overridden

sebgl:
Set a new `--webhook-cert-dir` flag that allows overriding the path to
the webhook certificates, most likely mounted from a Secret.

It's intended to be used this way:

```
args: ["manager", "--operator-roles", "all",
"--webhook-cert-dir=/tmp/whatever"]
```

(or by overriding the equivalent environment variable)

And the corresponding secret mount:

```
volumeMounts:
  - mountPath: /tmp/whatever
    name: cert
    readOnly: true
```

Fixes https://github.com/elastic/cloud-on-k8s/issues/2463.

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -145,7 +145,7 @@ func init() {
 	Cmd.Flags().String(
 		WebhookSecretFlag,
 		"",
-		fmt.Sprintf("K8s secret mounted into to the path designed by %s to be used for webhook certificates", WebhookCertDirFlag),
+		fmt.Sprintf("K8s secret mounted into the path designated by %s to be used for webhook certificates", WebhookCertDirFlag),
 	)
 	Cmd.Flags().String(
 		WebhookCertDirFlag,
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -67,6 +67,7 @@ const (
 
 	ManageWebhookCertsFlag   = "manage-webhook-certs"
 	WebhookSecretFlag        = "webhook-secret"
+	WebhookCertDirFlag       = "webhook-cert-dir"
 	WebhookConfigurationName = "elastic-webhook.k8s.elastic.co"
 	WebhookPort              = 9443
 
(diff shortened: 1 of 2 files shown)


cmd/manager/main.go
```diff
@@ -134,17 +135,23 @@ func init() {
...
 		OperatorNamespaceFlag,
 		"",
-		"k8s namespace the operator runs in",
+		"K8s namespace the operator runs in",
 	)
 	Cmd.Flags().String(
 		WebhookSecretFlag,
 		"",
-		"k8s secret mounted into /tmp/cert to be used for webhook certificates",
+		fmt.Sprintf("K8s secret mounted into to the path designed by %s to be used for webhook certificates", WebhookCertDirFlag),
```
thbkrkr: Nit: "into to" :dog: 
and not sure but: `s|designed|designated|`

docs/operator-config.asciidoc
```diff
@@ -26,7 +26,8 @@ ECK can be configured using either command line flags or environment variables.
 |operator-namespace |"" |Namespace the operator runs in. Required.
 |operator-roles |all |Roles this operator should assume. Valid values are `namespace`, `global`, `webhook` or `all`. Accepts multiple comma separated values.
 |webhook-pods-label |"" |Label used to select pods running the webhook server.
-|webhook-secret |"" | Kubernetes secret containing the webhook certificates.
+|webhook-secret |"" | K8s secret mounted into the path designated by webhook-cert-dir to be used for webhook certificates.
+|webhook-cert-dir |"{TempDir}/k8s-webhook-server/serving-certs" |Path to the directory that contains the webhook server key and certificate.
```
anyasabo: I was curious if the leading space differences in the last item (`| K8s secret mounted` vs `|Path to the directory`) changes how it renders. It looks like it does not so I don't think we need to be nitpicky here

thbkrkr: Out of curiosity, I wonder what can be the default value of "TempDir" other than `/tmp`?

sebgl: https://golang.org/pkg/os/#TempDir
> On Unix systems, it returns $TMPDIR if non-empty, else /tmp. On Windows, it uses GetTempPath, returning the first non-empty value from %TMP%, %TEMP%, %USERPROFILE%, or the Windows directory.

Issue #2479, opened 2020-01-29
Multi-version CRDs with trivialVersions don't have validation


sebgl:
When we have several apiVersions of the same CRD, but different content, we hit this bug in controller-tools: https://github.com/kubernetes-sigs/controller-tools/issues/349

Until https://github.com/kubernetes-sigs/controller-tools/pull/350 gets merged and released, we need to find a workaround in order to correctly generate the multi-version CRD with validation.

I think the easiest way is to generate the CRDs without `trivialVersions`, then manually patch the yaml file to get the expected `trivialVersions` behaviour.

Issue #2395, opened 2020-01-10
Mark version field as required


anyasabo:
As discussed in https://github.com/elastic/cloud-on-k8s/pull/2383 we should mark the `version` field of the spec as required. Currently this is challenging because v1 and v1beta1 need to be the same. I wonder if it is worthwhile to upgrade both before release? It's not a great practice but we rely on having a `version` in the spec right now anyway. This would just be making it explicit. I think we might want to do it with the v1.0.0 release?

Issue #1526, opened 2019-08-08
no matches for kind "Elasticsearch" in version "elasticsearch.k8s.elastic.co/v1alpha1"


asc-adean:
## Bug Report
**What did you do?**

Installed the all-in-one.yaml on AKS 1.14.0
Attempted to run step 2 at https://www.elastic.co/elasticsearch-kubernetes (Deploy Elasticsearch)

**What did you expect to see?**
Kubernetes accepted the input and start up elasticsearch Statefulset

**What did you see instead? Under which circumstances?**
`error: unable to recognize "STDIN": no matches for kind "Elasticsearch" in version "elasticsearch.k8s.elastic.co/v1alpha1"`

**Environment**
* Version information:
Azure AKS w/ RBAC enabled
Kubernetes server version 1.14.0
Kubernetes client version 1.15.1

* Resource definition:
I do not see the API in `api-versions`
```
 admissionregistration.k8s.io/v1beta1
apiextensions.k8s.io/v1beta1
apiregistration.k8s.io/v1
apiregistration.k8s.io/v1beta1
apps/v1
apps/v1beta1
apps/v1beta2
authentication.k8s.io/v1
authentication.k8s.io/v1beta1
authorization.k8s.io/v1
authorization.k8s.io/v1beta1
autoscaling/v1
autoscaling/v2beta1
autoscaling/v2beta2
batch/v1
batch/v1beta1
certificates.k8s.io/v1beta1
coordination.k8s.io/v1
coordination.k8s.io/v1beta1
events.k8s.io/v1beta1
extensions/v1beta1
metrics.k8s.io/v1beta1
networking.k8s.io/v1
networking.k8s.io/v1beta1
node.k8s.io/v1beta1
policy/v1beta1
rbac.authorization.k8s.io/v1
rbac.authorization.k8s.io/v1beta1
scheduling.k8s.io/v1
scheduling.k8s.io/v1beta1
storage.k8s.io/v1
storage.k8s.io/v1beta1
v1

```

* Logs:
```
k apply -f all-in-one.yaml
customresourcedefinition.apiextensions.k8s.io/apmservers.apm.k8s.elastic.co created
customresourcedefinition.apiextensions.k8s.io/elasticsearches.elasticsearch.k8s.elastic.co created
customresourcedefinition.apiextensions.k8s.io/trustrelationships.elasticsearch.k8s.elastic.co created
[...]

PR #2480, opened 2020-01-29
Mark the Version field required & adapt CRD generation for trivialVersions

sebgl:
This PR should have been split into 2:
1. Mark the `Version` field as mandatory
2. Fix CRD generation so multiple versions can be generated even though their schema differ

However they both rely on each other:
- CRD generation fails if we make the version field mandatory
- fixing CRD generation requires some JsonPath that do not exist if there is no change in the version schema
CI would fail building manifests in both individual cases. Hence a single PR with 2 commits.

## 1. Mark the Version field as required in all CRD

Currently if the Version field is not set, we reject the resource in the
validation webhook.
If the webhook is disabled, the operator returns an error early in the
reconciliation.

This could be considered a breaking API change, but the Version field
was already required implicitly, just not caught by the OpenAPI
validation.

Fixes https://github.com/elastic/cloud-on-k8s/issues/2395.

## 2. Customize CRDs to work around trivialVersions gen bug

A bug in controller-tools prevent validation to appear in the generated
CRDs if trivialVersions is set and multiple versions do not share the
exact same schema.
See kubernetes-sigs/controller-tools#349.

This commit removes the usage of `crd:trivialVersions=true`, to
"manually" (through Kustomize) apply the CRD modifications so we end up
with the trivialVersion scheme. That is: a single `validation` field
matching the OpenAPISchema of the currently stored version (v1).

Fixes https://github.com/elastic/cloud-on-k8s/issues/2479.


config/crds/all-crds.yaml
```diff
@@ -391,6 +393,7 @@ spec:
                 connect to.
               type: string
           type: object
+      type: object
```
sebgl: This appears when not using the trivialVersions controller-tools flag.
I think it's harmless, but I will double-check the CRD can be applied on k8s 1.11, just in case.

config/crds/patches/all-patches.yaml
```diff
@@ -1,9 +1,3 @@
-# Remove validation.Type that causes failures on k8s 1.11.
-# This should have been fixed with https://github.com/kubernetes-sigs/controller-tools/pull/72, but it looks like
-# this commit has been lost in history. See https://github.com/kubernetes-sigs/controller-tools/issues/296.
-# TODO: remove when fixed in controller-tools
-- op: remove
-  path: /spec/validation/openAPIV3Schema/type
```
sebgl: This bug does not exist when not using controller-tools with trivialVersions.

config/crds/patches/elasticsearch-patches.yaml
```diff
@@ -1,3 +1,13 @@
+# Move validation of the v1 version to the top-level, and remove validation from the v1beta1 version.
+# This is the behaviour intended from using controller-tools crd:trivialVersions=true flag,
+# which currently does not work as expected.
+# See https://github.com/kubernetes-sigs/controller-tools/issues/349.
+# TODO: remove once fixed in controller-tools
+- op: move
+  from: /spec/versions/0/schema
+  path: /spec/validation
+- op: remove
+  path: /spec/versions/1/schema
```
sebgl: Since ES, Kibana & APM share this same exact block I initially moved it to `patches/all-patches.yaml` but it didn't work well.
Is it because patches in the same file are applied sequentially vs. concurrently in different files, or something along those lines?

config/crds/all-crds.yaml
thbkrkr: Interesting. This reminds me that a user had to add this to fix its issue: https://github.com/elastic/cloud-on-k8s/issues/1526#issuecomment-534573668.

sebgl: Actually this comes from me removing this: https://github.com/elastic/cloud-on-k8s/pull/2480#discussion_r372430582
And applying CRDs fail on Minikube 1.11.
So I'll reintroduce the fix.

config/crds/patches/all-patches.yaml
sebgl: Actually, it does. I tested against Minikube 1.11. Reintroducing it.

config/crds/patches/elasticsearch-patches.yaml
anyasabo: Can you clarify the issue with it in `all-patches.yaml`? I was curious and so I checked it out and moved it there. It looks like it generates the same output for me on controller-tools v0.2.1.

anyasabo:
One thing I noticed while testing is that we will want to update the controller-gen check in the makefile:
https://github.com/elastic/cloud-on-k8s/blob/master/Makefile#L36
since it downloads 0.2.1 if it does not exist but we use 0.2.4 in go.mod. `controller-gen` doesn't have a command to expose the version, so I wonder if we just always want to `go get` the specific version. It takes ~5s on my machine though so maybe not.

I think any changes should probably be a separate PR though, because 0.2.4 also adds some of the markers required for structural schema (`x-int-or-string`).

config/crds/patches/elasticsearch-patches.yaml
sebgl: Since this is kustomize running after initial CRD generation is done it should not have anything to do with controller-tools version?
I updated kustomize to the latest version (3.5.4) and gave it another try.
When moving
```
# Move validation of the v1 version to the top-level, and remove validation from the v1beta1 version.
# This is the behaviour intended from using controller-tools crd:trivialVersions=true flag,
# which currently does not work as expected.
# See https://github.com/kubernetes-sigs/controller-tools/issues/349.
# TODO: remove once fixed in controller-tools
- op: move
  from: /spec/versions/0/schema
  path: /spec/validation
- op: remove
  path: /spec/versions/1/schema
# Remove validation.openAPIV3Schema.type that causes failures on k8s 1.11.
# This should have been fixed with https://github.com/kubernetes-sigs/controller-tools/pull/72, but it looks like
# this commit has been lost in history. See https://github.com/kubernetes-sigs/controller-tools/issues/296.
# TODO: remove once fixed in controller-tools
- op: remove
  path: /spec/validation/openAPIV3Schema/type
```
from `apm-kibana-patches.yaml` and `elasticsearch-patches.yaml` to `all-patches.yaml`, then run `make generate` I get the following error:

```
failed to apply json patch '
# Using `kubectl apply` stores the complete CRD file as an annotation,
# which may be too big for the annotations size limit.
# One way to mitigate this problem is to remove the (huge) podTemplate properties from the CRD.
# It also avoids the problem of having any k8s-version specific field in the Pod schema,
# that would maybe not match the user's k8s version.
- op: remove
  path: /spec/validation/openAPIV3Schema/properties/spec/properties/podTemplate/properties
': jsonpatch remove operation does not apply: doc is missing path: "/spec/validation/openAPIV3Schema/properties/spec/properties/podTemplate/properties"
make: *** [generate-crds] Error 1
```

I think that's because the patches in `elasticsearch-patches.yaml` rely on the patches in `all-patches.yaml`. Switching the order of items in `patchesJson6902` in `kustomization.yaml` does not seem to make any difference.

anyasabo: As discussed out of band I was only trying to move
```
- op: remove
  path: /spec/versions/1/schema
```
and it's not worth moving just that command to the `all-patches.yaml`, as we would also need to copy and paste the comment. I ran into the same error you did if I try to move the
```
- op: move
  from: /spec/versions/0/schema
  path: /spec/validation
```
stanza

PR #2482, opened 2020-01-29
Control associations across namespaces with ServiceAccount and RBAC

### LINKED ISSUE #2468: Access control on referenced objects
As raised in #2203 we should decide if we want to improve access control between objects.
The mechanism could be applied to any relationship between 2 objects of the stack, for example:
* Elasticsearch to Elasticsearch (#2203)
* Kibana or APM Server to Elasticsearch
* APM Server to Kibana (#1264)
* [Beats to Kibana](https://www.elastic.co/guide/en/beats/metricbeat/current/metricbeat-configuration.html)

So far 2 options have been identified:

## Option 1: An annotation or a new field in the spec

For example:
```yaml
kind: Elasticsearch
metadata:
  name: quickstart
spec:
  # allow ECK to configure resources in other namespaces to access this cluster
  # defaults to Nil (allow all), can be set to `allow: []` (only allow resources in the current namespace)
  crossNamespaceReferences:
      allow: ["namespaceA", "namespaceB"]
  version: 7.5.2
  nodeSets: []
```
Pro:
* Lightweight and easy to setup

Cons:
* Harder to figure out the permissions in case of an audit (it would need some dedicated tooling)
* ~How to manage granularity (namespace ? how to enforce only a given type ? fully qualified object name ?)~ _Edit: not sure it makes sense_ 

## Option 2: Using Kubernetes built-in RBAC mechanism

See https://github.com/elastic/cloud-on-k8s/compare/master...barkbay:rbac-controlled-references

Pro:
* Security is fully delegated to the API server, _(almost)_ no additional security code to maintain.
* Integrate with existing audit tools, for example:

```
$ kubectl who-can update elasticsearch -n elasticsearch-ns
ROLEBINDING                         NAMESPACE         SUBJECT      TYPE            SA-NAMESPACE
allow-kibana-from-remote-namespace  elasticsearch-ns  kibana-user  ServiceAccount  kibana-ns

CLUSTERROLEBINDING                                    SUBJECT                             TYPE            SA-NAMESPACE
cluster-admin                                         system:masters                      Group
cluster-admin-binding                                 michael.morello@elastic.co          User
elastic-operator                                      elastic-operator                    ServiceAccount  elastic-system
system:controller:clusterrole-aggregation-controller  clusterrole-aggregation-controller  ServiceAccount  kube-system
system:controller:generic-garbage-collector           generic-garbage-collector           ServiceAccount  kube-system
```

```
$ kubectl access-matrix for elasticsearch --verbs 'update' -n elasticsearch-ns
NAME                                KIND            SA-NAMESPACE    UPDATE
clusterrole-aggregation-controller  ServiceAccount  kube-system     ✔
elastic-operator                    ServiceAccount  elastic-system  ✔
generic-garbage-collector           ServiceAccount  kube-system     ✔
kibana-user                         ServiceAccount  kibana-ns       ✔
michael.morello@elastic.co          User                            ✔
system:masters                      Group                           ✔
```

Cons:

* Less trivial to setup than a annotation or an additional field

We should also decide if we want to allow references in a same "local" namespace by default

barkbay:
Fix #2468 

TODO:

- [x] rebase on #2480 
- [x] update all the `cluster_role.template.yaml`


cmd/manager/main.go
```diff
@@ -63,6 +64,8 @@ const (
 	CertValidityFlag       = "cert-validity"
 	CertRotateBeforeFlag   = "cert-rotate-before"
 
+	RbacControlledReferencesFlag = "rbac-controlled-references"
```
barkbay: _disclaimer: I'm bad at naming_

config/operator/all-in-one/cluster_role.template.yaml
```diff
@@ -6,6 +6,12 @@ kind: ClusterRole
 metadata:
   name: elastic-operator
 rules:
+- apiGroups:
```
barkbay: Only the cluster role of the all-in-one manifest is updated since this feature only makes sense when the operator is watching more than one namespace.

pkg/utils/rbac/access_review.go
```diff
@@ -0,0 +1,142 @@
...
+
+func (s *permissiveAccessReviewer) AccessAllowed(_ string, _ string, _ runtime.Object) (bool, error) {
+	return true, nil
+}
+
+// NextReconciliation returns a reconcile result depending on the implementation of the AccessReviewer.
+// It is mostly used when using the subjectAccessReviewer implementation in which case a next reconcile loop should be
+// triggered later to keep the association in sync with the RBAC roles and bindings.
+// See https://github.com/elastic/cloud-on-k8s/issues/2468#issuecomment-579157063
+func NextReconciliation(accessReviewer AccessReviewer) reconcile.Result {
```
sebgl: Having this in the `rbac` package feels a bit strange to me, it does not keep a clear separation of concerns.
Maybe in `common/association/rbac.go`? We could call it as `association.RequeueRbacCheck()`.
It would force `subjectAccessReviewer ` to become public though. Maybe that's ok? 

(I don't have a strong opinion on this, I'm fine if you prefer keeping the current approach)

pkg/utils/rbac/access_review.go
```diff
@@ -0,0 +1,142 @@
...
+				Namespace: metaObject.GetNamespace(),
+				Verb:      "get",
+				Resource:  plural,
+				Group:     strings.ToLower(object.GetObjectKind().GroupVersionKind().Group),
+				Version:   strings.ToLower(object.GetObjectKind().GroupVersionKind().Version),
+				Name:      metaObject.GetName(),
+			},
+			User: ServiceAccountUsernamePrefix + sourceNamespace + ":" + serviceAccount,
+		},
+	}
```
sebgl: Can we extract this to its own function and unit test it? Just because of the string manipulation.

config/samples/associations-rbac/apm_es_kibana_rbac.yaml
```diff
@@ -0,0 +1,120 @@
...
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: apmserver-ns
+---
+# Create a Role at the cluster level to access some Elasticsearch clusters.
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: elasticsearch-association
```
sebgl: Do you think we should pre-create this role as part of the operator manifest?

config/samples/associations-rbac/apm_es_kibana_rbac.yaml
```diff
@@ -0,0 +1,120 @@
...
+    name: kibana-user
+    namespace: kibana-ns
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: apmserver-user
+  namespace: apmserver-ns
+---
+# This RoleBinding give the permission to Kibana to access the Elasticsearch cluster
```
sebgl: gives

config/samples/associations-rbac/apm_es_kibana_rbac.yaml
```diff
@@ -0,0 +1,120 @@
...
+      - get # association is allowed if a resource can "get" the remote one
+---
+# This is the service account used by Kibana
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: kibana-user
+  namespace: kibana-ns
+---
+# This RoleBinding give the permission to Kibana to access the Elasticsearch cluster
```
sebgl: gives

config/samples/associations-rbac/apm_es_kibana_rbac.yaml
```diff
@@ -0,0 +1,120 @@
...
+metadata:
+  name: apm-apm-sample
+  namespace: apmserver-ns
+spec:
+  version: 7.5.2
+  count: 1
+  elasticsearchRef:
+    name: "elasticsearch-sample"
+    namespace: "elasticsearch-ns"
+  # Service account used the APM Server to get access to the Elasticsearch cluster
```
sebgl: used by

pkg/controller/apmserverelasticsearchassociation/apmserverelasticsearchassociation_controller.go
```diff
@@ -252,6 +264,23 @@ func (r *ReconcileApmServerElasticsearchAssociation) reconcileInternal(apmServer
...
+	if !allowed {
+		log.Info("Association not allowed",
+			"as_name", apmServer.Name,
+			"as_namespace", apmServer.Namespace,
+			"serviceAccount", apmServer.Spec.ServiceAccountName,
+			"es_name", es.Name,
+			"es_namespace", es.Namespace,
+		)
+		// Ensure that user in Elasticsearch is deleted to prevent illegitimate access
+		err := user.DeleteUser(r.Client, NewUserLabelSelector(assocKey))
```
sebgl: Should we also call association.RemoveAssociationConf?

sebgl:
It would be nice to have this verified in an E2E test, but it does not really match the current E2E test framework :/
Let's make sure to at least create an issue for it?

pkg/controller/apmserverelasticsearchassociation/apmserverelasticsearchassociation_controller.go
sebgl: We should maybe extract that block and unit test it does the right thing when the association is not allowed.

pkg/utils/rbac/access_review.go
barkbay: I was wondering exactly the same. The problem is that this package is supposed to be reused for remote clusters (CCR/CCS) which is not strictly speaking considered (so far) as an association but more as an internal feature of the elasticsearch controller.

config/samples/associations-rbac/apm_es_kibana_rbac.yaml
barkbay: My gut feeling is that we should create roles only if there are needed but add it as an example in our documentation if the user wants to enable access control. That being said I'm happy to pre-create it if you think it's worth it.

barkbay:
> Let's make sure to at least create an issue for it?

Agreed, I'll create a follow-up issue to track the e2e test effort for this feature.

pkg/utils/rbac/access_review.go
barkbay: I have addressed your concern and moved this part of the code to the association package. I will check later if it does fit or not with remote clusters.

pkg/apis/apm/v1/apmserver_types.go
```diff
@@ -40,6 +40,11 @@ type ApmServerSpec struct {
 	// SecureSettings is a list of references to Kubernetes secrets containing sensitive configuration options for APM Server.
 	// See: https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-apm-server.html#k8s-apm-secure-settings
 	SecureSettings []commonv1.SecretSource `json:"secureSettings,omitempty"`
+
+	// ServiceAccountName is used to check access from the current resource to a resource (eg. Elasticsearch) in a different namespace.
+	// Can only be used if ECK is configured for checking references RBAC.
```
pebrc: This wording strikes me as a bit odd: should this say "check references with RBAC" or "enforcing RBAC on references"

pkg/controller/common/association/rbac.go
```diff
@@ -0,0 +1,62 @@
...
+	if !allowed {
+		metaObject, err := meta.Accessor(object)
+		if err != nil {
+			return false, nil
+		}
+		log.Info("Association not allowed",
+			"associated_kind", associated.GetObjectKind().GroupVersionKind().Kind,
+			"associated_name", associated.GetName(),
+			"associated_namespace", associated.GetNamespace(),
+			"serviceAccount", associated.ServiceAccountName(),
```
pebrc: `serviceAccount` seems to only one not using `_`

cmd/manager/main.go
```diff
@@ -63,6 +64,8 @@ const (
 	CertValidityFlag       = "cert-validity"
 	CertRotateBeforeFlag   = "cert-rotate-before"
 
+	CheckReferencesRbac = "check-references-rbac"
```
pebrc: I know this has already been renamed once and I don't want to bikeshed too much but what about `check-rbac-on-refs` or `enforce-rbac-on-refs` somthing that makes it clear that we are not checking the references but RBAC rules when working with references

sebgl: +1 that's better

config/samples/associations-rbac/apm_es_kibana_rbac.yaml
sebgl: 👍  let's not include it for now and maybe do it later depending on users feedback.

pkg/controller/common/association/rbac.go
```diff
@@ -0,0 +1,62 @@
...
+
+type Unbinder interface {
+	Unbind(associated commonv1.Associated) error
+}
+
+// CheckAndUnbind checks if a reference is allowed and unbind the association if it is not the case
+func CheckAndUnbind(
+	accessReviewer rbac.AccessReviewer,
+	associated commonv1.Associated,
+	object runtime.Object,
```
sebgl: `object` is not very explicit: is it the src object holding the reference, or the referenced one?
Let's maybe rename it to `remoteObject` or `referencedObject`.

pkg/controller/common/association/rbac.go
```diff
@@ -0,0 +1,62 @@
...
+
+	commonv1 "github.com/elastic/cloud-on-k8s/pkg/apis/common/v1"
+	"github.com/elastic/cloud-on-k8s/pkg/utils/rbac"
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+type Unbinder interface {
+	Unbind(associated commonv1.Associated) error
```
sebgl: I like this 👍

sebgl:
We should have a dedicated page in the documentation (in this PR or in a follow-up one).

pkg/controller/common/association/rbac.go
```diff
@@ -0,0 +1,62 @@
...
+		}
+		log.Info("Association not allowed",
+			"associated_kind", associated.GetObjectKind().GroupVersionKind().Kind,
+			"associated_name", associated.GetName(),
+			"associated_namespace", associated.GetNamespace(),
+			"service_account", associated.ServiceAccountName(),
+			"remote_type", object.GetObjectKind().GroupVersionKind().Kind,
+			"remote_name", metaObject.GetNamespace(),
+			"remote_namespace", metaObject.GetName(),
+		)
```
sebgl: Let's emit an event here?

pkg/utils/rbac/access_review.go
```diff
@@ -0,0 +1,117 @@
...
+
+	sar := newSubjectAccessReview(metaObject, object, serviceAccount, sourceNamespace)
+
+	sar, err = s.client.AuthorizationV1().SubjectAccessReviews().Create(sar)
+	if err != nil {
+		return false, err
+	}
+	log.V(1).Info(
+		"Access review", "result",
+		sar.Status, "service_account", serviceAccount,
```
sebgl: Let's put:
```
"result", sar.Status,
```
on the same line.

config/samples/associations-rbac/apm_es_kibana_rbac.yaml
```diff
@@ -0,0 +1,120 @@
...
+    name: kibana-user
+    namespace: kibana-ns
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: apmserver-user
+  namespace: apmserver-ns
+---
+# This RoleBinding gives the permission to Kibana to access the Elasticsearch cluster
```
sebgl: s/Kibana/ApmServer/

PR #2484, opened 2020-01-30
Add gcloud component installation commands to dev setup

wandergeek:
I missed this when I was setting up and don't want someone else to do the same. :)


sebgl:
Thanks @wandergeek!

Issue #2270, opened 2019-12-16
Instrument e2e tests and feed results into a permanent ES cluster


pebrc:
As discussed previously we should instrument our e2e test environments with beats and feed the data into a long lived Elasticsearch cluster managed by ECK. This should help us
* investigate issues with our e2e test suite
* identify potential feature gaps in ECK wrt to monitoring and metrics
* create a long running ECK environment that might be useful in itself (memory leaks, version upgrades, etc) 

We should set up an ILM policy to configure the data retention in that cluster to limit the overall amount of data stored.

sebgl:
Having Elasticsearch logs available would have helped to troubleshoot https://github.com/elastic/cloud-on-k8s/issues/2380.

PR #2486, opened 2020-01-30
Setup e2e tests to run with filebeat logging to e2e monitor cluster

david-kow:
This PR:
- allows instrumenting in-cluster E2E test runs with Filebeat
- enables this instrumentation for the main E2E test

Data needed by Filebeat to output the logs is stored in Vault and accessed when preparing the run.

Resolves https://github.com/elastic/cloud-on-k8s/issues/2270.

### DIFF
--- a/build/ci/Makefile
+++ b/build/ci/Makefile
@@ -69,3 +69,6 @@ get-test-license:
 
 get-docker-creds:
 	@ echo "ELASTIC_DOCKER_PASSWORD = $(shell VAULT_TOKEN=$(VAULT_TOKEN) vault read -address=$(VAULT_ADDR) -field=value secret/devops-ci/cloud-on-k8s/eckadmin)" >> ../../.env
+
+get-monitoring-secrets:
+	@ VAULT_TOKEN=$(VAULT_TOKEN) vault read -address=$(VAULT_ADDR) -field=data -format=json secret/devops-ci/cloud-on-k8s/monitoring-cluster > monitoring-secrets.json
--- a/config/dev/elastic-psp.yaml
+++ b/config/dev/elastic-psp.yaml
@@ -46,3 +46,54 @@
         - min: 1
           max: 65535
     readOnlyRootFilesystem: false
+---
+  apiVersion: policy/v1beta1
+  kind: PodSecurityPolicy
+  metadata:
+    name: elastic.filebeat.restricted
+    annotations:
+      seccomp.security.alpha.kubernetes.io/allowedProfileNames: 'docker/default,runtime/default'
+      apparmor.security.beta.kubernetes.io/allowedProfileNames: 'runtime/default'
+      seccomp.security.alpha.kubernetes.io/defaultProfileName:  'runtime/default'
+      apparmor.security.beta.kubernetes.io/defaultProfileName:  'runtime/default'
+  spec:
+    privileged: false
+    # Required to prevent escalations to root.
+    allowPrivilegeEscalation: false
+    # This is redundant with non-root + disallow privilege escalation,
+    # but we can provide it for defense in depth.
+    requiredDropCapabilities:
+      - ALL
+    # Allow core volume types.
+    volumes:
+      - 'configMap'
+      - 'emptyDir'
+      - 'projected'
+      - 'secret'
+      - 'downwardAPI'
+      # Assume that persistentVolumes set up by the cluster admin are safe to use.
+      - 'persistentVolumeClaim'
+      # Allow access to the host for log files.
+      - 'hostPath'
+    hostNetwork: true
+    hostIPC: false
+    hostPID: false
+    runAsUser:
+      # Allow the container to run with root privileges.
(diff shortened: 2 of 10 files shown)


config/dev/elastic-psp.yaml
```diff
@@ -46,3 +46,54 @@
         - min: 1
           max: 65535
     readOnlyRootFilesystem: false
+---
```
barkbay: Maybe add a comment to explain why we need a dedicated psp and what privileges are needed ?

config/e2e/filebeat.yaml
```diff
@@ -0,0 +1,191 @@
...
+      labels:
+        k8s-app: filebeat
+    spec:
+      serviceAccountName: filebeat
+      terminationGracePeriodSeconds: 30
+      hostNetwork: true
+      dnsPolicy: ClusterFirstWithHostNet
+      containers:
+        - name: filebeat
+          image: docker.elastic.co/beats/filebeat:7.5.1
```
barkbay:
```suggestion
          image: docker.elastic.co/beats/filebeat:7.5.2
```

test/e2e/cmd/run/run.go
```diff
@@ -220,6 +246,16 @@ func (h *helper) deployNamespaceOperator() error {
 	return h.kubectlApplyTemplateWithCleanup("config/e2e/namespace_operator.yaml", h.testContext)
 }
 
+func (h *helper) deployFilebeat() error {
+	if h.monitoringSecrets == "" {
+		log.Info("Not deploying filebeat")
```
barkbay: nit: maybe explain why it is not deployed ? 
```suggestion
		log.Info("No monitoring secrets provided, filebeat is not deployed")
```

build/ci/Makefile
```diff
@@ -69,3 +69,6 @@ get-test-license:
 
 get-docker-creds:
 	@ echo "ELASTIC_DOCKER_PASSWORD = $(shell VAULT_TOKEN=$(VAULT_TOKEN) vault read -address=$(VAULT_ADDR) -field=value secret/devops-ci/cloud-on-k8s/eckadmin)" >> ../../.env
+
+get-monitoring-secrets:
+	@ VAULT_TOKEN=$(VAULT_TOKEN) vault read -address=$(VAULT_ADDR) -field=data -format=json secret/devops-ci/cloud-on-k8s/monitoring-cluster > monitoring-secrets.json
```
barkbay: Should we add `monitoring-secrets.json` to `.gitignore` to avoid any mistake ?

test/e2e/cmd/run/command.go
```diff
@@ -74,6 +75,7 @@ func Command() *cobra.Command {
 	cmd.Flags().BoolVar(&flags.skipCleanup, "skip-cleanup", false, "Do not run cleanup actions after test run")
 	cmd.Flags().StringVar(&flags.testContextOutPath, "test-context-out", "", "Write the test context to the given path")
 	cmd.Flags().StringVar(&flags.testLicense, "test-license", "", "Test license to apply")
+	cmd.Flags().StringVar(&flags.monitoringSecrets, "monitoring-secrets", "", "Monitoring secrets to use")
```
thbkrkr: Trying to take a step back, I'm wondering why we don't let the e2e program fetchs directly the secrets it needs from Vault?

test/e2e/cmd/run/run.go
```diff
@@ -181,6 +183,30 @@ func (h *helper) initTestSecrets() error {
...
 	}
+
+	if h.monitoringSecrets != "" {
+		bytes, err := ioutil.ReadFile(h.monitoringSecrets)
+		if err != nil {
+			return err
+		}
+
+		monitoringSecrets := struct {
+			MonitoringIp   string `json:"monitoringIp"` //nolint
```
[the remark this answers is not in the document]
david-kow: Got some issues with that as some libraries expect camelCase for struct name. But this is irrelevant here, so changing to `MonitoringIP`.

test/e2e/cmd/run/command.go
david-kow: I would have nothing against that. That's what we do in deployer too.

david-kow:
@pebrc @barkbay I wanted to box this task, so we end up with a working cluster that stores our e2e logs. I think there is plenty of follow up work that we can potentially do to improve it (dns, ca, kubelet/api-server logs, metrics, traces, dashboards, alerts, ES sizing/tuning, probably much more), but I felt it's a value add already. As for the other work, I think that we can track/prioritize separately, WDYT?

test/e2e/cmd/run/command.go
thbkrkr: So, I would advocate for doing that.
This would avoid adding another make target and env vars, and also to have the secret written in a file on the disk.

Then, it would be very nice to do the same with the test licenses which are only used by the e2e runner.
Yes, I'm on a quest to reduce the number of env vars we are using to run our e2e tests :)

PR #2494, opened 2020-01-31
Test for up2date kubectl version

ThoTischner:
Signed-off-by: Thomas Tischner <tti@bitsbeats.com>

If kubectl is installed without kustomize support make integration will break.
This PR will extend the makefile to check for up2date kubectl version and fail with a good error message if requirements are not met.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -79,6 +79,9 @@ MANAGED_NAMESPACES ?=
 # TODO: only relevant on GKE for e2e tests for the moment
 PSP ?= 0
 
+## -- Tools
+KUBECTL_VERSION=$(shell [[ `kubectl --client=true version | grep 'Minor:"[0-9]*' -o| grep -o [0-9]*` -lt 14 ]] && echo notok || echo ok)
+
 #####################################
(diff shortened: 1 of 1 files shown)


ThoTischner:
CLA signed and commit got signature.

thbkrkr:
> I'm a bit on the fence regarding adding additional checks in a Makefile that is already fairly complicated. I think I'd rather not.

Yes, let's not make the `Makefile` more complicated.

However, [hack/check-requisites.sh](https://github.com/elastic/cloud-on-k8s/blob/master/hack/check-requisites.sh) seems a good place for this check.

charith-elastic:
I am closing this PR for now because the preference seems to be having this check in `hack/check-requisites.sh` instead. Please feel free to re-open if you have an alternative suggestion or wish to make the change to `hack/check-requisites.sh` yourself.

ThoTischner:
@thbkrkr So you will implement the check or should I open a new PR ?

thbkrkr:
@ThoTischner I'll take care of it, thank you.

PR #2495, opened 2020-01-31
Extend Clusterwide rbac roles for elastic crds

ThoTischner:
Signed-off-by: Thomas Tischner <tti@bitsbeats.com>

Without this PR the clusterwide edit or view roles do not include permissions  for the elastic CRs.
For this some RBAC resources with labels are needed.

CertManager for example also includes this for it`s quick start manifest: https://github.com/jetstack/cert-manager/blob/29389b19d515a167866f1c77278a1fe70456723b/deploy/charts/cert-manager/templates/rbac.yaml#L396

### DIFF
--- a/config/operator/all-in-one/rbac.yaml
+++ b/config/operator/all-in-one/rbac.yaml
@@ -1,0 +1,38 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: elastic-operator-view
+  labels:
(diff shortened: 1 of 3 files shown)


charith-elastic:
Jenkins test this please

sebgl:
I wasn't aware this was a thing, but it looks useful.
The [Prometheus operator](https://github.com/coreos/prometheus-operator/blob/master/Documentation/rbac-crd.md) describes something similar.

Issue #2490, opened 2020-01-30
Bump controller-tools to 0.2.5


anyasabo:
Opening an issue because it is not as straightforward as it seemed at first. You can see the updated CRD here: https://github.com/anyasabo/cloud-on-k8s/commit/5d68e0cd40d27a2dd70c35d8de0eded6c4597c23

tl;dr openshift 3.11 is still a pain to support

Controller-tools 0.2.5's CRD generation adds some validation properties that appear to force it to be evaluated as a CRD with structural schema, and there's no way to disable adding those properties. This is the error I receive
>CustomResourceDefinition.apiextensions.k8s.io "kibanas.kibana.k8s.elastic.co" is invalid: spec.validation.openAPIV3Schema.type: Required value: must not be empty at the root

The new `x-kubernetes-int-or-string: true` properties added to the CRD cause this -- it works if I remove those properties. I'm working on a PR now to add kustomize rules removing them all.

For what it's worth in 1.15, if we remove the kustomize rule removing the `type` property, it also works (though you need to set `--validate=false` in newer kubectl versions). So this is just related to the Openshift 3.11 workaround. It doesn't seem necessary to update controller-tools right now, so I think we are good. When we want to support k8s 1.19 (where structural schema are required) we will need to revisit.

This is the validation error if you do have the `type: object` property for reference (bypassable with `--validate=false`):
>error: error validating "./config/crds/all-crds.yaml": error validating data: [ValidationError(CustomResourceDefinition.spec.validation.openAPIV3Schema.properties.spec.properties.http.properties.service.properties.spec.properties.ports): unknown field "x-kubernetes-list-map-keys" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1beta1.JSONSchemaProps, ValidationError(CustomResourceDefinition.spec.validation.openAPIV3Schema.properties.spec.properties.http.properties.service.properties.spec.properties.ports): unknown field "x-kubernetes-list-type" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1beta1.JSONSchemaProps]; if you choose to ignore these errors, turn validation off with --validate=false

PR #2497, opened 2020-01-31
Bump controller-tools

anyasabo:
Fix https://github.com/elastic/cloud-on-k8s/issues/2490

Controller tools 0.2.5 includes k8s 1.17 deps, and also includes some of the new openapi properties to support new CRD features. This causes a few issues:
- `x-kubernetes-int-or-string` seems to cause the 1.15 api server to interpret the schema as structural, and so is sad that there is no `type: object` property (which we intentionally remove for Openshift 3.11 support). Removing the `x-kubernetes-int-or-string` property allows the API server to accept the manifests on both 1.11 and 1.15, though kubectl validation fails so you need to pass it `--validation=false`.

- Kubectl validation fails because there are other new properties that are added (`x-kubernetes-list-type` and `x-kubernetes-list-map-keys`). I removed these properties too to make applying the CRDs easier, even though they are not strictly required by the API server.

I believe these properties will be required in k8s 1.19 when v1beta1 CRD support is removed and structural schema is required, but that is a bit away. All of this stems from supporting 1.11 which has a bug that requires us to remove `type:object` from the root. Once we stop supporting 1.11 this all becomes much simpler.

Other options than the approach in this PR that may be worthwhile:
- Add a flag to crd-gen to disable adding these markers, though since it is only an issue for <=1.11 support I am not sure how much support it would get
- Pin an older version of controller-tools (looks like 0.2.2 is the latest we could use just browsing release notes)

Also added the namespace to some unit tests that were passing but don't seem like they should have been in older versions (and were failing now).

It's also not clear to me if we should target v1.0.1 with this.


pkg/controller/elasticsearch/driver/upgrade_test.go
```diff
@@ -36,11 +36,11 @@ func Test_podsToUpgrade(t *testing.T) {
 			args: args{
 				statefulSets: sset.StatefulSetList{
 					sset.TestSset{
-						Name: "masters", Replicas: 2, Master: true,
+						Name: "masters", Namespace: TestEsNamespace, Replicas: 2, Master: true,
```
sebgl: I'm surprised to see that in the PR.
Does it fail if not set?

anyasabo: Yeah, it was surprising to me that this ever passed. The NS was added in 
9d3aaa8 so if you check out any of the ones before that you should be able to see.

sebgl:
We may want to wait for https://github.com/kubernetes-sigs/controller-tools/issues/392 to be resolved first?

anyasabo:
Ah I missed that issue somehow, yes that would definitely make the kustomize part of this PR simpler. I could see this going either way though -- we could merge this as is then use the flag once it is available, or we could just close this and wait for the flag.

config/crds/patches/apm-kibana-patches.yaml
```diff
@@ -21,3 +21,13 @@
 # that would maybe not match the user's k8s version.
 - op: remove
   path: /spec/validation/openAPIV3Schema/properties/spec/properties/podTemplate/properties
+
+# without this the API server complains with "The CustomResourceDefinition "elasticsearches.elasticsearch.k8s.elastic.co" is invalid: spec.validation.openAPIV3Schema.type: Required value: must not be empty at the root" on 1.15 at least. When we stop supporting 1.11 we should remove these as well.
```
sebgl: We should include something like
```
# TODO: remove once https://github.com/kubernetes-sigs/controller-tools/issues/392 is fixed
```
IMO this is a bug in controller-tools.
Also it looks like this errors out in 1.14 as well, so we probably want to remove the 1.11 mention.

sebgl: Nit: can you reformat the comments to multiple lines so they read better on a small-ish screen?

charith-elastic:
Jenkins test this please

anyasabo:
This is pending https://github.com/elastic/cloud-on-k8s/pull/2611 as there are some oddities in the license generator on OSX vs Linux that should be resolved in that PR. Right now the CI check is failing for the notice file

Issue #2483, opened 2020-01-30
Default port names do not reflect https


anyasabo:
Edited for clarity*

Follow up to https://github.com/elastic/cloud-on-k8s/pull/2083

I followed the quickstart, which creates pods listening on https only by default. The port name was still `http` though. I was confused until I noticed the original PR was only targeting the service. This seems to be sufficient for istio, but was confusing when I was looking into https://github.com/elastic/cloud-on-k8s/issues/2489 where you need to provide a mapping of protocols to container port names for ingress. It looks odd to be entering
`cloud.google.com/app-protocols: '{"http":"HTTPS"}'`

I think we ought to name the service and the container port the same, and they should match the actual protocol being used.

For reference here's the returned ES config with the default TLS settings:
```
apiVersion: elasticsearch.k8s.elastic.co/v1
kind: Elasticsearch
metadata:
  annotations:
    common.k8s.elastic.co/controller-version: 1.0.0
    elasticsearch.k8s.elastic.co/cluster-uuid: acVisZ48RE6-3pUgjARbMw
  creationTimestamp: "2020-01-30T03:11:46Z"
  generation: 2
  name: quickstart
  namespace: default
  resourceVersion: "488760"
  selfLink: /apis/elasticsearch.k8s.elastic.co/v1/namespaces/default/elasticsearches/quickstart
  uid: 4bf61ed6-d09c-4ca4-994a-730c57a87f44
spec:
  http:
    service:
      metadata:
        creationTimestamp: null
      spec: {}
    tls:
      certificate: {}
  nodeSets:
  - config:
      node.data: true
      node.ingest: true
      node.master: true
      node.store.allow_mmap: false
    count: 1
    name: default
    podTemplate:
      metadata:
        creationTimestamp: null
      spec:
        containers: null
  updateStrategy:
    changeBudget: {}
  version: 7.5.2
status:
  availableNodes: 1
  health: green
  phase: Ready
```

but the actual statefulset:
```
        name: elasticsearch
        ports:
        - containerPort: 9200
          name: http
          protocol: TCP
        - containerPort: 9300
          name: transport
          protocol: TCP
```

PR #2498, opened 2020-01-31
Name container ports according to protocol

anyasabo:
Fix https://github.com/elastic/cloud-on-k8s/issues/2483

Previously we were only naming the service ports after the protocol, but we were hardcoding the container ports "http" even if they were listening only on https. This names the container ports according to the protocol.

This will break services that were referring to named ports rather than the port number. Internally we only use the port number though.


pkg/controller/elasticsearch/nodespec/podspec.go
```diff
@@ -68,6 +70,14 @@ func BuildPodTemplateSpec(
 	return builder.PodTemplate, nil
 }
 
+func getDefaultContainerPorts(es esv1.Elasticsearch) []corev1.ContainerPort {
+	return []corev1.ContainerPort{
+		{Name: es.Spec.HTTP.Protocol(), ContainerPort: network.HTTPPort, Protocol: corev1.ProtocolTCP},
+		{Name: "transport", ContainerPort: network.TransportPort, Protocol: corev1.ProtocolTCP},
+	}
+
```
thbkrkr: Nit: empty line

anyasabo:
@sebgl that seems to make sense to me too. For me it was three things that made the old implementation odd:
- seeing the container port be `http`, but not actually listen for bare http requests
- the service port being `https` but the container port it maps to being `http`
- when providing the protocol -> container port mappings in the [GKE ingress annotation](https://cloud.google.com/kubernetes-engine/docs/concepts/ingress#disabling_http), it also seemed strange to map the https protocol to the http port

So it wasn't so much fitting the k8s ecosystem expectations -- I think we already did that with changing the service port name -- but more for consistency. To my knowledge nothing keys automatically off the container port name like they do for the service port name.

PR #2501, opened 2020-02-03
some more details to gcloud beta component

jihun-cho:
While I was trying to execute `make bootstrap-cloud`, I encountered an error message like below.
```
$ make bootstrap-k8s
...
2020/02/03 14:51:57 Authenticating as user...
2020/02/03 14:51:58 Checking if cluster exists...
Error: exit status 1
exit status 1
make: *** [bootstrap-k8s] Error 1
```

I was able to find the reason by looking into gke.go file.
I think I tried the install command already, but well it looks like it did not go well or I make some mistake during the installation.

Anyway, I try this PR if it could be helpful to diagnose a situation I experienced easily for some others.

I think this additional comment might be too verbose, if so it is okay discard this PR.
So please feel free to review this. :)

### DIFF
--- a/dev-setup.md
+++ b/dev-setup.md
@@ -15,17 +15,7 @@ Before you start, install the following tools and packages:
 * sha1sum (for Mac `brew install md5sha1sum`)
 * [gcloud](https://cloud.google.com/sdk/gcloud/)
   * Install beta and docker-credential-gcr components:  `gcloud components install beta docker-credential-gcr`
-  * If it were not installed successfully, you'll get below error
-    ```
-    $ make bootstrap-k8s
-    ...
-    2020/02/03 14:56:10 Checking if cluster exists...
-    Error: exit status 1
-    exit status 1
-    make: *** [bootstrap-k8s] Error 1
-    ```
-  * Check if `beta` component installed successfully: `gcloud beta container clusters --project {GCloudProject} describe {ClusterName} --region {Region}`
-    * ex: gcloud beta container clusters --project elastic-cloud-dev describe jihuncho-dev-cluster --region europe-west1
+  * Run `gcloud components list --filter="id:(beta,docker-credential-gcr)"` and ensure that both are marked as Installed
 
 
 ### Get sources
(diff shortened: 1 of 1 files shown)


dev-setup.md
```diff
@@ -13,11 +13,22 @@ Before you start, install the following tools and packages:
...
+    $ make bootstrap-k8s
+    ...
+    2020/02/03 14:56:10 Checking if cluster exists...
+    Error: exit status 1
+    exit status 1
+    make: *** [bootstrap-k8s] Error 1
+    ```
+  * Check if `beta` component installed successfully: `gcloud beta container clusters --project {GCloudProject} describe {ClusterName} --region {Region}`
+    * ex: gcloud beta container clusters --project elastic-cloud-dev describe jihuncho-dev-cluster --region europe-west1
+
```
charith-elastic: I think we can condense this to the following line:

Run `gcloud components list --filter="id:(beta,docker-credential-gcr)"` and ensure that both are marked as `Installed`

jihun-cho: Thank you @charith-elastic , I think that would be far better.

charith-elastic:
Jenkins test this please

thbkrkr:
I am not against this change as information is missing, but it may not be the one to recommend.
I do not have `docker-credential-gcr` installed on my machine. 
Instead I use gcloud as credential helper as [it is the recommended auth method](https://cloud.google.com/container-registry/docs/advanced-authentication#gcloud-helper).

charith-elastic:
That's a good catch @thbkrkr 
I feel like the whole document needs a closer review because there are a few inconsistencies and out-dated information there. I have created #2505 to track it.

jihun-cho:
If we need to review the document in wider view, may I close this PR?

charith-elastic:
@jihun-cho Sure, thanks for bringing this to our attention. We'll make sure your suggestions are incorporated into the new text.

jihun-cho:
Thank you all!

PR #2506, opened 2020-02-03
Exclude associations-rbac from e2e samples

barkbay:
I can't find a way to exclude a directory [with a glob](https://golang.org/pkg/path/filepath/#Match). So I added a regexp to filter RBAC samples.


thbkrkr:
The last time we encountered this kind of problem, we had decided to move the sample in the `recipes` directory (https://github.com/elastic/cloud-on-k8s/pull/2071). It seems to me simpler?

barkbay:
> The last time we encountered this kind of problem, we had decided to move the sample in the `recipes` directory (#2071). It seems to me simpler?

It feels a little bit counterintuitive to me to place this one in the recipes directory, it is not very complex and only rely on "core" resources. But happy to move it there.

PR #2520, opened 2020-02-05
Make use of secure port when accessing Kubelet API

ChrsMark:
## What does this PR do?

This PR switches Metricbeat k8s manifests and docs to point to Kubelet secure port over https instead of the insecure port.

## Why is it important?

Insecure port of Kubelet (10255/TCP) is now less common and discouraged and also in most cases it is not enabled by default (requiring to restart `kubelet` with `--read-only-port` flag)

Related to https://github.com/elastic/beats/pull/16063


ChrsMark:
Thanks for reviewing @charith-elastic ! Keep in mind that I'm not authorised to merge PRs for the project, so one of you should push the button :)

PR #2524, opened 2020-02-05
Add webhook network policy troubleshooting information

### LINKED ISSUE #1634: Validate webhook fails open
From a report on the discuss forums:
https://discuss.elastic.co/t/error-from-server-notfound-services-quickstart-es-http-not-found/193658/15

It appears that if GKE has network policies enforced, but no network policy allowing the API server to communicate with the webhook, that admission of new elastic resources will time out. The goal of setting the failure policy to `Ignore` (see https://github.com/elastic/cloud-on-k8s/pull/1386) was to allow creation to still progress even if the webhook was unavailable, but it appears we may be hitting a combination of timeouts that complicates that. We should investigate and ensure that our webhook does indeed fail open in this environment.

It may also be worth updating the docs with an example network policy to allow it to function in GKE with network policies enforced as well.

Similar k/k issue:
https://github.com/kubernetes/kubernetes/issues/71508

charith-elastic:
Adds documentation about troubleshooting connectivity issues to the webhook server and includes some sample network policies that can be used when network policy enforcement is in place.

Fixes #1634

### DIFF
--- a/docs/webhook.asciidoc
+++ b/docs/webhook.asciidoc
@@ -112,7 +112,14 @@ Webhooks require network connectivity between the Kubernetes API server and the 
 Error from server (Timeout): error when creating "elasticsearch.yaml": Timeout: request did not complete within requested timeout 30s
 ....
 
-One possible reason for the above error is that a link:https://kubernetes.io/docs/concepts/services-networking/network-policies/[network policy] might be blocking any incoming requests to the webhook server. Consult your system administrator to determine whether that is the case and create an appropriate policy to allow communication between the Kubernetes API server and the webhook server. For example, the following network policy simply opens up the webhook port to the world:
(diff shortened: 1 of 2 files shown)


docs/webhook.asciidoc
```diff
@@ -102,8 +102,73 @@ EOF
...
+----
+
+You may want to restrict webhook access to just the Kubernetes API server. Currently this requires knowing the IP address of the API server -- which can be obtained through the command:
+
+[source,sh,subs="attributes"]
+----
+kubectl cluster-info | grep master
+----
+
+Assuming that the API server IP address is `10.1.0.1`, the following policy restrticts webhook access to just the API server.
```
thbkrkr: typo: s|restrticts|restricts|

PR #2528, opened 2020-02-06
Make readiness probes independent of the host/container network

### LINKED ISSUE #1819: Use a command for Kibana readiness probe due to security constraints with httpGet
## Proposal

**Use case. Why is this important?**

It is required for reliably checking for kibana pod containers readiness.

The current readiness seems very basic and not sure when it is expect to get the actual readiness.

Current readiness:
```
readinessProbe:
      failureThreshold: 3
      httpGet:
        path: /login
        port: 5601
        scheme: HTTPS
      initialDelaySeconds: 10
      periodSeconds: 10
      successThreshold: 1
      timeoutSeconds: 5
```

The above readiness probe just probes on the /login but may hide few other problems with the kibana container readiness itself. 

I feel it could be something to reliably to check the readiness, as done for elasticsearch containers.

thbkrkr:
Switch from HttpGet action to an Exec action for the readiness probes
of Kibana and APM Server to support Kubernetes environments where
network traffic is blocked from hosts to containers for security
reasons.

Resolves #1819.

### DIFF
--- a/pkg/controller/kibana/pod/pod.go
+++ b/pkg/controller/kibana/pod/pod.go
@@ -5,6 +5,9 @@
 package pod
 
 import (
+	"fmt"
+	"strings"
+
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common/annotation"
 	"k8s.io/apimachinery/pkg/api/resource"
 
@@ -17,7 +20,6 @@ import (
 	"github.com/elastic/cloud-on-k8s/pkg/utils/stringsutil"
 
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
 const (
@@ -49,6 +51,7 @@ func readinessProbe(useTLS bool) corev1.Probe {
 	if useTLS {
 		scheme = corev1.URISchemeHTTPS
 	}
+	protocol := strings.ToLower(fmt.Sprintf("%s", scheme))
 	return corev1.Probe{
 		FailureThreshold:    3,
 		InitialDelaySeconds: 10,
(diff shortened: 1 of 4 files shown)


pkg/controller/kibana/pod/pod.go
```diff
@@ -56,10 +57,10 @@ func readinessProbe(useTLS bool) corev1.Probe {
 		SuccessThreshold:    1,
 		TimeoutSeconds:      5,
 		Handler: corev1.Handler{
-			HTTPGet: &corev1.HTTPGetAction{
-				Port:   intstr.FromInt(HTTPPort),
-				Path:   "/login",
-				Scheme: scheme,
+			Exec: &corev1.ExecAction{
+				Command: []string{"bash", "-c",
+					fmt.Sprintf(`curl -o /dev/null -w "%%{http_code}" %s://127.0.0.1:%d/login -k -s`, scheme, HTTPPort),
```
sebgl: Why is there 2 `%` here vs. only one in the apm case?

thbkrkr: `%%` is for `%` in a `fmt.Sprintf` template.

```sh
# fmt.Sprintf
pkg/controller/kibana/pod/pod.go: curl -o /dev/null -w "%%{http_code}" ...
pkg/controller/apmserver/pod.go:  curl -o /dev/null -w "%%{http_code}" ...
```

```sh
# string
pkg/controller/kibana/driver_test.go: curl -o /dev/null -w "%{http_code}" ...
pkg/controller/kibana/driver_test.go: curl -o /dev/null -w "%{http_code}" ...
pkg/controller/apmserver/apmserver_controller_test.go: curl -o /dev/null -w "%{http_code}" ...
```

sebgl: 👍 I misread the code: there is the same for APM (I was looking at the test)

PR #2530, opened 2020-02-06
Remove operator roles

### LINKED ISSUE #2254: Deprecate the "global + namespace" operator concept

pebrc:
What this PR does:
* remove the concept of operator roles as flags
* remove the global operator manifests 
* keep/modify the namespace operator manifests to allow deployments restricted to one or more namespaces
* allow the parameterization of the operator name to allow multiple (namespaced) operators to be deployed in a single namespace
* make the operator logs target work for both variants
* run the e2e tests with a all-in-one operator
* add a dedicated webhook toggle flag
* disable the webhook by default in the namespaced variant, enable it in the all-in-one variant

What this PR does **not**: 
* introduce new CLI/kustomize/Helm tooling to craft different versions of the manifests
* replace the various manifest variants with just one that is customized as need 
* remove the need to update versions and RBAC permissions in various places
* remove the separate e2e manifests

Fixes #2254


pebrc:
Jenkins test this please

test/e2e/cmd/run/run.go
```diff
@@ -236,14 +231,9 @@ func (h *helper) createManagedNamespaces() error {
 	return h.kubectlApplyTemplateWithCleanup("config/e2e/managed_namespaces.yaml", h.testContext)
 }
 
-func (h *helper) deployGlobalOperator() error {
+func (h *helper) deployOperator() error {
 	log.Info("Deploying global operator")
```
barkbay:
```suggestion
	log.Info("Deploying operator")
```

test/e2e/test/context.go
```diff
@@ -81,8 +77,7 @@ func defaultContext() Context {
 
 // Context encapsulates data about a specific test run
 type Context struct {
-	GlobalOperator        ClusterResource   `json:"global_operator"`
-	NamespaceOperator     NamespaceOperator `json:"namespace_operator"`
+	Operator              NamespaceOperator `json:"global_operator"`
```
barkbay: Should we also rename `global_operator` ?

Makefile
```diff
@@ -66,11 +66,11 @@ SKIP_DOCKER_COMMAND ?= false
 
 ## -- Namespaces
 
-# namespace in which the global operator is deployed (see config/global-operator)
-GLOBAL_OPERATOR_NAMESPACE ?= elastic-system
-# namespace in which the namespace operator is deployed (see config/namespace-operator)
-NAMESPACE_OPERATOR_NAMESPACE ?= elastic-namespace-operators
-# comma separated list of namespaces in which the namespace operator should watch resources
+# namespace in which the operator is deployed (see config/global-operator)
```
barkbay: nit: not from your PR but I'm not sure where `config/global-operator` comes from

test/e2e/test/context.go
pebrc: Yes 💯 I relied on the refactoring support of my editor and missed that.

Makefile
```diff
@@ -191,15 +191,13 @@ endif
 endif
 
 # Deploy both the global and namespace operators against the current k8s cluster
-deploy: check-gke install-crds build-operator-image apply-operators
+deploy: check-gke install-crds build-operator-image apply-namespaced-operator
 
-apply-operators:
+apply-namespaced-operator:
```
thbkrkr: We could just name it `apply-operator`?

cmd/manager/main.go
```diff
@@ -332,48 +325,46 @@ func execute() {
...
-		}
 
-		go func() {
-			time.Sleep(10 * time.Second)         // wait some arbitrary time for the manager to start
-			mgr.GetCache().WaitForCacheSync(nil) // wait until k8s client cache is initialized
-			r := licensing.NewResourceReporter(mgr.GetClient())
-			r.Start(operatorNamespace, licensing.ResourceReporterFrequency)
-		}()
+	// Garbage collect any orphaned user Secrets leftover from deleted resources while the operator was not running.
+	garbageCollectUsers(cfg, managedNamespaces)
```
thbkrkr: Nit: move the call to `garbageCollectUsers` after adding of all the controllers.

Makefile
thbkrkr: There are still some references to `global` to delete in this Makefile.

```sh
153: # Run the operator locally with ... and operator namespace for a GLOBAL operator
159:  --development --operator-roles=GLOBAL,namespace \
172:  --operator-roles=GLOBAL,namespace \
193: # Deploy both the GLOBAL and namespace operators against the current k8s cluster
```

config/operator/namespace/cluster_role.template.yaml
```diff
@@ -90,19 +94,6 @@ rules:
...
-  - apmserverelasticsearchassociations
-  - apmserverelasticsearchassociations/status
-  verbs:
-  - get
-  - list
-  - watch
-  - create
-  - update
-  - patch
-  - delete
```
thbkrkr: Why can we delete this RBAC rule?

pebrc: Because the corresponding CRD does not exist anymore?

thbkrkr: Of course. :confused:

config/operator/README.md
```diff
@@ -16,25 +16,12 @@ A single operator with all roles, that manages resources in all namespaces.
...
-
-```bash
-OPERATOR_IMAGE=<?> NAMESPACE=<?> make generate-global | kubectl apply -f -
-```
-
-#### Namespace
+### Namespaced
 
-A namespace operator that manages resources in a given namespace.
+One or more operators managing resources in a given set of of namespaces.
```
thbkrkr: Nit: `s/set of of namespaces/set of namespaces/`

Issue #2467, opened 2020-01-27
Improve ECS support in operator logs


barkbay:
Following up #2002, #2457 added some support for ECS but some improvements can be done:

1. Should we consider a "reconciliation loop" as an [event](https://www.elastic.co/guide/en/ecs/current/ecs-event.html)?

* `event.sequence` looks like our `iteration` field _(number of times a controller has run its Reconcile method)_
* `event.duration` looks like our current `took`
* `event.type` could be [ `change`| `creation` | `deletion` ]
* `event.dataset` could help to separate the logs from the different controller (elasticsearch-controller, kibana-controller...)

2. The presence of some fields is inconsistent

It is worth to note that it is not easy to be consistent re. the key/values added to the log.
The most problematic case is when an error is managed by the [`controller` itself](https://github.com/kubernetes-sigs/controller-runtime/blob/8c39906c77cdb574482c01fcd74a76b105b06522/pkg/internal/controller/controller.go#L256) _(i.e. not by our code)_ :

 ```golang
if result, err := c.Do.Reconcile(req); err != nil {
	c.Queue.AddRateLimited(req)
	log.Error(err, "Reconciler error", "controller", c.Name, "request", req)
	[...]
	return false
}
```
This would produce:
```json
{
	"message": "Reconciler error",
	"log.logger": "controller-runtime.controller",
	"log.level": "error",
	"controller": "kibana-association-controller",
	"error.message": "An error message about kibanas.kibana.k8s.elastic.co \"kb-apm-sample\": something bad happened",
[...]

PR #2534, opened 2020-02-07
Rename log.logger to event.dataset as recommended in ECS

pebrc:
Instead of 

Screenshot 2020-02-07 at 11 16 57

we get

Screenshot 2020-02-07 at 11 15 20

in Kibanas log UI by following the ECS recommendation for `event.dataset`

### DIFF
--- a/pkg/utils/log/log.go
+++ b/pkg/utils/log/log.go
@@ -74,7 +74,7 @@ func setLogger(v *int) {
 		encoderConf.MessageKey = "message"
 		encoderConf.TimeKey = "@timestamp"
 		encoderConf.LevelKey = "log.level"
-		encoderConf.NameKey = "log.logger"
+		encoderConf.NameKey = "event.dataset"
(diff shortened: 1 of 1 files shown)


thbkrkr:
I'm a little confused by the amount of fields that can be used to describe the source of events and when one should be used more than another (https://github.com/elastic/ecs/issues/281#issuecomment-520567252).

But this change doesn't look dangerous. I admit that I trust you on this one with slightly closed eyes.

How is it ok to not set `log.logger`?

pebrc:
I have second thoughts now that I see that ECS seems to revise its position on where to put the logger in the discussion you linked to @thbkrkr

barkbay:
I didn't set this field for this reason and created an issue to discuss improvements re. ECS : https://github.com/elastic/cloud-on-k8s/issues/2467#issue-555433021

pebrc:
Let's not do this for now.

PR #2536, opened 2020-02-07
Do not mutate object when validating unknown fields

thbkrkr:
This commit takes care not to reuse the object to be validated during the validation to detect unknown fields in order to avoid any mutation.
Indeed the object stored in this annotation can be slightly different from the object after it has been stored and retrieved by the operator (e.g.: `resources:{limits:{}}` became `resources:{}`, see #2207).

Relates to #2433 and #2207.

### DIFF
--- a/pkg/apis/common/v1/validations.go
+++ b/pkg/apis/common/v1/validations.go
@@ -16,13 +16,13 @@ import (
 )
 
 // NoUnknownFields checks whether the last applied config annotation contains json with unknown fields.
-func NoUnknownFields(dest interface{}, meta metav1.ObjectMeta) field.ErrorList {
+func NoUnknownFields(dest runtime.Object, meta metav1.ObjectMeta) field.ErrorList {
 	var errs field.ErrorList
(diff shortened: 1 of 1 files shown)


pkg/apis/common/v1/validations.go
```diff
@@ -20,7 +21,9 @@ func NoUnknownFields(dest interface{}, meta metav1.ObjectMeta) field.ErrorList {
 	if cfg, ok := meta.Annotations[v1.LastAppliedConfigAnnotation]; ok {
 		d := json.NewDecoder(strings.NewReader(cfg))
 		d.DisallowUnknownFields()
-		if err := d.Decode(&dest); err != nil {
+		// copy the resource to be validated to avoid mutation if the object in the annotation is different
+		destCopy := dest.(runtime.Object).DeepCopyObject()
```
sebgl: How about we change the signature of `NoUnknownFields ` to accept a `runtime.Object` instead of an `interface{}`?
This would avoid the need for casting.

PR #2537, opened 2020-02-10
Add operator flag to define default container registry

### LINKED ISSUE #2223: Support custom registry address in operator command line flags
In private network eviroment, I can't download docker images from `docker.elastic.co` .
if we could set operator command line flags such as: `--default-registry=xxx.xxx.xxx`, would be more convenience to use eck where can not access internet directly.

I knew docker image can be specified, but sometimes would get confused when using image with version field, so if we can specify registry, the yaml config would be more concise.

charith-elastic:
Adds an operator flag named `--container-registry` that allows users to specify the default container registry to use when `spec.image` is not defined in the CR manifest. This is useful for users in air-gapped installations as they will no longer have to explicitly define the container image in every manifest they deploy. New users will have a better experience following the examples in the documentation as well because those would now work out of the box. 

Also includes a minor refactoring to move the operator flag definitions to their own package for readability and consistency.

For testing, an image exists at `eu.gcr.io/elastic-cloud-dev/elasticsearch/elasticsearch:7.5.0`. Start the operator with `--container-registry=eu.gcr.io/elastic-cloud-dev`.

Fixes #2223


cmd/manager/main.go
```diff
@@ -102,6 +103,11 @@ func init() {
 		certificates.DefaultCertValidity,
 		"Duration representing how long before a newly created TLS certificate expires",
 	)
+	Cmd.Flags().String(
+		operator.ContainerRegistryFlag,
+		container.DefaultContainerRegistry,
+		"Container registry to use when downloading Elastic stack container images",
```
thbkrkr: Nit: Elastic **S**tack

```sh
> curl -s https://www.elastic.co/products/| grep -oi "elastic stack" | sort | uniq -c
     15 Elastic Stack
```

docs/custom-images.asciidoc
```diff
@@ -55,3 +55,16 @@ For more information, you can check the following references:
 - https://docs.microsoft.com/en-us/azure/container-registry/[Azure Container Registry docs]
 - https://docs.aws.amazon.com/AmazonECR/latest/userguide/what-is-ecr.html[Amazon Elastic Container Registry docs]
 - https://docs.openshift.com/container-platform/4.1/registry/architecture-component-imageregistry.html[OpenShift Container Platform registry docs]
+
+
+[float]
+[id="{p}-container-registry-override"]
+=== Overriding the default container registry
+
+When creating Elasticsearch, Kibana, or APM server resources, the operator defaults to using container images pulled from the `docker.elastic.co` registry. If you are in an environment where external network access is restricted, you could configure the operator to use a different default container registry by starting the operator with the `--container-registry` command-line flag. (See <<{p}-operator-config>> for more information about how to configure the operator using command-line flags and environment variables).
```
thbkrkr: Nit: APM **S**erver

```sh
> grep -ioh 'APM server' -r docs | sort | uniq -c
      4 APM server
     94 APM Server
```

docs/operator-config.asciidoc
```diff
@@ -17,6 +17,7 @@ ECK can be configured using either command line flags or environment variables.
 |ca-cert-validity |8760h |Duration representing the validity period of a generated CA certificate.
 |cert-rotate-before |24h |Duration representing how long before expiration TLS certificates should be re-issued.
 |cert-validity |8760h |Duration representing the validity period of a generated TLS certificate.
+|container-registry |docker.elastic.co | Container registry to use for pulling Elastic stack container images.
```
thbkrkr: Same: Elastic **S**tack

Issue #2540, opened 2020-02-10
cert-validity is ignored


anyasabo:
From a discuss topic: https://discuss.elastic.co/t/cert-validity-isnt-working/218631

It looks like we do not actually use the cert-validity parameters. [This func](https://github.com/elastic/cloud-on-k8s/blob/master/pkg/controller/elasticsearch/certificates/ca_reconcile.go#L80) seems to be responsible for generating the certificates, but we only pass it the CA validity params.

Attached logs below (with some new debug logs added to make sure the params are parsed correctly). You can see the actual certificate is valid for 4h instead of 2h.

```
12:32 $ make go-run
# Run the operator locally with role All, with debug logs, operator image set to latest and operator namespace for a global operator
AUTO_PORT_FORWARD=true \
		go run \
			-ldflags "-X github.com/elastic/cloud-on-k8s/pkg/about.version=1.0.0 -X github.com/elastic/cloud-on-k8s/pkg/about.buildHash=688635ea -X github.com/elastic/cloud-on-k8s/pkg/about.buildDate=2020-02-10T18:33:06Z -X github.com/elastic/cloud-on-k8s/pkg/about.buildSnapshot=true" \
			-tags "" \
			./cmd/main.go manager \
				--development --operator-roles=global,namespace \
				--log-verbosity=1 \
				--ca-cert-validity=4h --ca-cert-rotate-before=1h \
				--cert-validity=2h --cert-rotate-before=1h \
				--operator-namespace=default \
				--namespaces= \
				--manage-webhook-certs=false
...
2020-02-10T12:33:14.728-0600	DEBUG	manager	Using certificate authority rotation parameters	{"service.version": "1.0.0-688635ea", "ca-cert-validity": "4h0m0s", "ca-cert-rotate-before": "1h0m0s"}
2020-02-10T12:33:14.728-0600	DEBUG	manager	Using certificate rotation parameters	{"service.version": "1.0.0-688635ea", "cert-validity": "2h0m0s", "cert-rotate-before": "1h0m0s"}
```

```
[root@quickstart-es-default-0 elasticsearch]# curl --insecure -v https://localhost:9200 2>&1 | awk 'BEGIN { cert=0 } /^\* SSL connection/ { cert=1 } /^\*/ { if (cert) print }'
* SSL connection using TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384
* Server certificate:
[...]

PR #2541, opened 2020-02-10
Use cert rotate parameter

anyasabo:
Fix https://github.com/elastic/cloud-on-k8s/issues/2540

Previously we did not actually use the certificate rotation parameters, and only used the CA rotation parameters.

Also increases the NotBefore buffer from 1 minute to 10 minutes for the CA to match the certificate buffer. If clock skews are that far out there's probably other issues, but it definitely happens.

I didn't add tests for the new behavior because currently we do not really have tests at this level. I can def see how adding integration tests for this would be helpful though, so if you want me to add them I'm happy to do so.


pkg/controller/apmserver/certificates/reconcile.go
```diff
@@ -48,15 +49,15 @@ func Reconcile(
...
+		caRotation,
 	)
 	if err != nil {
 		return results.WithError(err)
 	}
 
 	// handle CA expiry via requeue
 	results.WithResult(reconcile.Result{
-		RequeueAfter: certificates.ShouldRotateIn(time.Now(), httpCa.Cert.NotAfter, rotation.RotateBefore),
+		RequeueAfter: certificates.ShouldRotateIn(time.Now(), httpCa.Cert.NotAfter, caRotation.RotateBefore),
```
sebgl: Should we also make that call for the other certs?

pebrc: I am not sure I follow your comment here @sebgl. Do you mean we should requeue within the validity period of the individual certificates and not just before the CA expires?

sebgl: Yeah I mean also requeueing after we reconcile HTTP certificates below. Since we have `caRotation` and `certRotation` that can be different, it would make sense to make sure we requeue before any of these are reached? Maybe I'm missing something here.

pebrc: Should we not requeue on the minimum of the two?

sebgl: Yes! That's why I thought we would call `results.WithResult ()` twice (one for CA, one for cert), and then we let the results aggregation do its job of picking the most appropriate requeue?

pebrc: SGTM

anyasabo:
Yep! That was the plan, sorry I was on call so I let this sit. I should have put in an update though indicating the changes @sebgl suggested out of band

pkg/controller/kibana/certificates/reconcile.go
```diff
@@ -74,6 +74,15 @@ func Reconcile(
 	if err != nil {
 		return results.WithError(err)
 	}
+
+	primaryCert, err := certificates.GetPrimaryCertificate(httpCertificates.CertPem())
```
pebrc: It is a bit unfortunate that we have to parse the PEM again that we just parsed and or encoded in the reconcile function. But I guess changing this would be a larger refactoring of our certificate generation logic, so happy to track this in a follow up task.

PR #2543, opened 2020-02-11
Add local K8S remote cluster support

### LINKED ISSUE #2203: Support for remote ES clusters within a single k8s cluster

barkbay:
Fixes #2203

What this PR does:

* It adds the `remoteClusters` field _(as usual naming can be discussed)_ to the `Elasticsearch` schema:

```yaml
kind: Elasticsearch
metadata:
  name: cluster-1
  namespace: ns1
spec:
  version: 7.5.2
  remoteClusters:
  - name: ns2-cluster-2
    elasticsearchRef:
      name: cluster-2
      namespace: ns2
    #config:
    #  transport.ping_schedule: 30s
    #  transport.compress: true
    #  skip_unavailable: true
```

* Adds a new controller to copy certificate authorities between clusters. Each remote certificate is copied into the namespace of the peer cluster.
* The elasticsearch controller then concatenates all the remote certificates into an new `Secret` called `<cluster-name>-es-remote-ca`

Here is a example of the new `Secrets`:

```
$ k get es,secrets -n ns1                    
NAME                                                   HEALTH   NODES   VERSION   PHASE   AGE
elasticsearch.elasticsearch.k8s.elastic.co/cluster-1   green    3       7.5.2     Ready   2m

NAME                                          TYPE                                  DATA   AGE
[...]
secret/cluster-1-es-remote-ca                 Opaque                                1      2m21s
secret/cluster-1-ns2-cluster-2-es-remote-ca   Opaque                                1      2m18s

$ k get es,secrets -n ns2
NAME                                                   HEALTH   NODES   VERSION   PHASE   AGE
elasticsearch.elasticsearch.k8s.elastic.co/cluster-2   green    3       7.5.2     Ready   1m

NAME                                          TYPE                                  DATA   AGE
[...]
secret/cluster-2-es-remote-ca                 Opaque                                1      81s
secret/cluster-2-ns1-cluster-1-es-remote-ca   Opaque                                1      80s
```

* The elasticsearch controller also does the right API calls to setup the remote cluster in the settings. The last configuration is stored as an annotation in the Elasticsearch resource in order to avoid to do unnecessary calls:

```yaml
apiVersion: v1
items:
- apiVersion: elasticsearch.k8s.elastic.co/v1
  kind: Elasticsearch
  metadata:
    annotations:
      elasticsearch.k8s.elastic.co/remote-clusters: '[{"name":"ns2-cluster-2","configHash":"3795207740"}]'
```

* This PR also creates a transport service:

```
$ k get svc -n ns1
NAME                     TYPE           CLUSTER-IP    EXTERNAL-IP      PORT(S)          AGE
cluster-1-es-default     ClusterIP      None          <none>           <none>           10m
cluster-1-es-http        LoadBalancer   10.28.XX.XX   35.195.XX.XX     9200:32076/TCP   10m
cluster-1-es-transport   ClusterIP      None          <none>           9300/TCP         10m
```

Side notes:

* I have added some spans for performance monitoring but I still need to check if they are relevant.
* Don't forget to start a trial or use an Enterprise license if you want to do some tests


barkbay:
jenkins test this please

barkbay:
jenkins test this please

pkg/controller/elasticsearch/remotecluster/elasticsearch.go
```diff
@@ -0,0 +1,147 @@
...
+	// RemoteClusters to remove
+	for name := range currentRemoteClusters {
+		if _, ok := expectedRemoteClusters[name]; !ok {
+			persistentSettings := newRemoteClusterSetting(name, nil)
+			log.Info("Remove remote cluster",
+				"namespace", es.Namespace,
+				"es_name", es.Name,
+				"remote_cluster", name,
+			)
+			err := updateRemoteCluster(esClient, persistentSettings)
```
pebrc: Should we not try to make this just one HTTP call for all added/removed remote clusters?

pkg/controller/elasticsearch/remotecluster/remoteca/controller.go
```diff
@@ -0,0 +1,474 @@
...
+const (
+	name = "remoteca-controller"
+
+	EventReasonLocalCaCertNotFound = "LocalClusterCaNotFound"
+	EventReasonRemoteCaCertMissing = "RemoteClusterCaNotFound"
+	CaCertMissingError             = "Cannot find CA certificate for %s cluster %s/%s"
+)
+
+var (
+	//log            = logf.Log.WithName(name)
```
pebrc: left over?

pkg/controller/elasticsearch/remotecluster/remoteca/controller.go
```diff
@@ -0,0 +1,474 @@
...
+}
+
+// getCurrentRemoteCertificateAuthorities returns all the remote cluster keys for which a remote ca exists
+func getCurrentRemoteCertificateAuthorities(
+	ctx context.Context,
+	c k8s.Client,
+	associatedEs types.NamespacedName,
+) (map[types.NamespacedName]struct{}, error) {
+	span, _ := apm.StartSpan(ctx, "get_current_remote_ca", tracing.SpanTypeApp)
+	defer span.End()
```
pebrc: newline?

pkg/controller/elasticsearch/remotecluster/remoteca/labels.go
```diff
@@ -0,0 +1,62 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	// RemoteClusterNamespaceLabelName used to represent the namespace of the RemoteCluster in a TrustRelationship.
+	RemoteClusterNamespaceLabelName = "remotecluster.k8s.elastic.co/namespace"
+	// RemoteClusterNameLabelName used to represent the name of the RemoteCluster in a TrustRelationship.
+	RemoteClusterNameLabelName = "remotecluster.k8s.elastic.co/name"
+	// TypeLabelValue is a type used to .
```
pebrc: used to ...?

pkg/controller/elasticsearch/remotecluster/remoteca/watches.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+
+	// Watch Secrets that contain remote certificate authorities managed by this controller
+	if err := c.Watch(&source.Kind{Type: &v1.Secret{}}, &handler.EnqueueRequestsFromMapFunc{
+		ToRequests: newToRequestsFuncFromSecret(),
+	}); err != nil {
+		return err
+	}
+
+	// Dynamically the certificate authorities involved in a cluster relationship
```
pebrc: Dynamically watches?

pkg/controller/elasticsearch/services/services.go
```diff
@@ -9,27 +9,66 @@ import (
...
 const (
 	globalServiceSuffix = ".svc"
 )
 
+// TransportServiceName returns the name for the transport service associated to this cluster
+func TransportServiceName(esName string) string {
+	return esv1.TransportService(esName)
+}
+
+// NewTransportService returns the transport service associated to the given cluster
```
pebrc:
```suggestion
// NewTransportService returns the transport service associated with the given cluster
```

pkg/controller/elasticsearch/services/services.go
```diff
@@ -9,27 +9,66 @@ import (
...
+// It is used by nodes to talk to each other.
```
pebrc: Elasticsearch nodes maybe?

pkg/controller/elasticsearch/services/services.go
```diff
@@ -9,27 +9,66 @@ import (
...
+// It is used by Elasticsearch nodes to talk to each other.
+func NewTransportService(es esv1.Elasticsearch) *corev1.Service {
+	nsn := k8s.ExtractNamespacedName(&es)
+	return &corev1.Service{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: es.Namespace,
+			Name:      TransportServiceName(es.Name),
+			Labels:    label.NewLabels(nsn),
+		},
+		Spec: corev1.ServiceSpec{
```
pebrc: Should we allow users to control the spec like we do for the HTTP service? For remoting outside of the k8s cluster users probably want this to be of type `LoadBalancer` for example.

barkbay: It makes sense. That said this PR is already pretty large and is targeting clusters inside a same k8s cluster. I would do that in a dedicated PR, I can create an issue to track this effort.

pebrc: 👍

pkg/controller/elasticsearch/remotecluster/remoteca/watches.go
```diff
@@ -0,0 +1,108 @@
...
+	return fmt.Sprintf(
+		"%s-%s-%s-%s",
+		local.Namespace,
+		local.Name,
+		remote.Namespace,
+		remote.Name,
+	)
+}
+
+// addCertificatesAuthorityWatches sets some watches on all secrets containing the certificate of a CA involved in a association.
```
sebgl: Because I had to think about it twice (local vs. remote, what is watched?), maybe worth a comment like:

```
// watch the remote cluster CA secret to reconcile the local cluster CA secrets
```

pkg/controller/elasticsearch/remotecluster/remoteca/controller.go
```diff
@@ -0,0 +1,474 @@
...
+	}
+	if !enabled {
+		log.Info(
+			"Remote cluster controller is an enterprise feature. Enterprise features are disabled",
+			"namespace", es.Namespace, "es_name", es.Name,
+		)
+		return reconcile.Result{}, nil
+	}
+
+	// Use the driver to create the remote cluster
```
sebgl: let's remove this comment, there's no driver here

pkg/controller/elasticsearch/remotecluster/remoteca/controller.go
```diff
@@ -0,0 +1,474 @@
...
+				expectedRemoteClusters[k8s.ExtractNamespacedName(&es)] = struct{}{}
+			}
+		}
+	}
+
+	return expectedRemoteClusters, nil
+}
+
+// getCurrentRemoteCertificateAuthorities returns all the remote cluster keys for which a remote ca exists
+func getCurrentRemoteCertificateAuthorities(
```
sebgl: Wordking: I think we generally use `actual` instead of `current`.

pkg/controller/elasticsearch/remotecluster/remoteca/controller.go
```diff
@@ -0,0 +1,474 @@
...
+				esRef.Name == associatedEs.Name {
+				expectedRemoteClusters[k8s.ExtractNamespacedName(&es)] = struct{}{}
+			}
+		}
+	}
+
+	return expectedRemoteClusters, nil
+}
+
+// getCurrentRemoteCertificateAuthorities returns all the remote cluster keys for which a remote ca exists
```
sebgl: `all the Elasticsearch keys for which a remote CA exists`?

pkg/controller/elasticsearch/remotecluster/remoteca/controller.go
```diff
@@ -0,0 +1,474 @@
...
+func getCurrentRemoteCertificateAuthorities(
+	ctx context.Context,
+	c k8s.Client,
+	associatedEs types.NamespacedName,
```
sebgl: What's associatedEs in this context? A local cluster or a remote cluster?

pkg/controller/elasticsearch/remotecluster/remoteca/controller.go
```diff
@@ -0,0 +1,474 @@
...
+	if err := c.List(
+		&remoteCAList,
+		client.InNamespace(associatedEs.Namespace),
+		GetRemoteCaMatchingLabel(associatedEs.Name),
+	); err != nil {
+		return nil, err
+	}
+	for _, remoteCA := range remoteCAList.Items {
+		if remoteCA.Labels == nil {
+			continue
```
sebgl: Can this happen? We set a label selector just above?

pkg/controller/elasticsearch/remotecluster/remoteca/controller.go
```diff
@@ -0,0 +1,474 @@
...
+		}
+		remoteEs := remoteCA.Labels[label.ClusterNameLabelName]
+		currentRemoteClusters[types.NamespacedName{
+			Namespace: remoteCA.Namespace,
+			Name:      remoteEs,
+		}] = struct{}{}
+	}
+
+	return currentRemoteClusters, nil
```
sebgl: I'm a bit lost in what is remote vs. what is local in this function.
Might be worth splitting into 2 different functions with more comments.

pkg/controller/elasticsearch/remotecluster/remoteca/controller.go
```diff
@@ -0,0 +1,474 @@
...
+}
+
+func doReconcile(
+	ctx context.Context,
+	r *ReconcileRemoteCa,
+	localEs *esv1.Elasticsearch,
+) (reconcile.Result, error) {
+	localClusterKey := k8s.ExtractNamespacedName(localEs)
+	// Get current clusters according to the existing remote CAs
+	currentRemoteCertificateAuthorities, err := getCurrentRemoteCertificateAuthorities(ctx, r.Client, localClusterKey)
```
sebgl: I think we use `actual` instead of `current` in most places

pkg/controller/elasticsearch/remotecluster/remoteca/controller.go
sebgl: func name suggestion: `func clustersWithRemoteCAs()`

pkg/controller/elasticsearch/remotecluster/remoteca/controller.go
```diff
@@ -0,0 +1,474 @@
...
+	// Create or update expected remote CA
+	for remoteEsKey := range expectedRemoteCertificateAuthorities {
+		// Get the remote Elasticsearch cluster associated with this remote CA
+		remoteEs := &esv1.Elasticsearch{}
+		if err := r.Client.Get(remoteEsKey, remoteEs); err != nil {
+			if !errors.IsNotFound(err) {
+				return reconcile.Result{}, err
+			}
+			// Remote cluster does not exist, skip it
+			continue
```
sebgl: The other way around reads IMHO better because of the dual negation (not is not found):
```go
if errors.IsNotFound(err) {
    // Remote cluster does not exist, skip it
    continue
}
```

pkg/controller/elasticsearch/remotecluster/remoteca/controller.go
```diff
@@ -0,0 +1,474 @@
...
+		}
+		accessAllowed, err := isRemoteClusterAssociationAllowed(r.accessReviewer, localEs, remoteEs, r.recorder)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+		if !accessAllowed {
```
sebgl: `// if the remote CA exists but isn't allowed anymore, it will be deleted next`

pkg/controller/elasticsearch/remotecluster/remoteca/controller.go
```diff
@@ -0,0 +1,474 @@
...
+		log.V(1).Info("Delete remote CA",
+			"localNamespace", localEs.Namespace,
+			"localName", localEs.Name,
+			"remoteNamespace", toDelete.Namespace,
+			"remoteName", toDelete.Name,
+		)
+		results.WithError(deleteCertificateAuthorities(ctx, r, localClusterKey, toDelete))
+	}
+	return results.WithResult(association.RequeueRbacCheck(r.accessReviewer)).Aggregate()
+
```
sebgl: nit: newline

pkg/controller/elasticsearch/remotecluster/remoteca/controller.go
```diff
@@ -0,0 +1,474 @@
...
+		}
+	}
+
+	// Delete existing but not expected remote CA
+	for toDelete := range currentRemoteCertificateAuthorities {
+		log.V(1).Info("Delete remote CA",
+			"localNamespace", localEs.Namespace,
+			"localName", localEs.Name,
+			"remoteNamespace", toDelete.Namespace,
+			"remoteName", toDelete.Name,
```
sebgl: let's respect the usual `log_key_formatting`

pkg/controller/elasticsearch/remotecluster/remoteca/controller.go
```diff
@@ -0,0 +1,474 @@
...
+		delete(currentRemoteCertificateAuthorities, remoteEsKey)
+		results.WithResults(createOrUpdateCertificateAuthorities(ctx, r, localEs, remoteEs))
+		if results.HasError() {
+			return results.Aggregate()
+		}
+	}
+
+	// Delete existing but not expected remote CA
+	for toDelete := range currentRemoteCertificateAuthorities {
+		log.V(1).Info("Delete remote CA",
```
sebgl: `Deleting`

pkg/controller/elasticsearch/remotecluster/remoteca/controller.go
```diff
@@ -0,0 +1,474 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
```
sebgl: Generally I'm a bit lost in what's remote and what's local in this entire file.
I think that's because I initially misunderstood what we want to do here.
I thought since we only deal with unidirectionnal remote clusters we only had to copy the remote cluster CA into the local cluster namespace. But IIUC even though the remote cluster is unidirectionnal from ES point of view, our certificate handling needs to be bidirectionnal?

I don't know if we can arrange the code in such a way that this is a bit more explicit.

pkg/controller/elasticsearch/remotecluster/remoteca/controller.go
```diff
@@ -0,0 +1,474 @@
...
+	ctx context.Context,
+	c k8s.Client,
+	owner *esv1.Elasticsearch,
+	remote types.NamespacedName,
+	remoteCA []byte,
+) error {
+	span, _ := apm.StartSpan(ctx, "reconcile_remote_ca", tracing.SpanTypeApp)
+	defer span.End()
+
+	// Define the desired remote CA object, it lives in the remote namespace.
```
sebgl: I'm a bit lost here: that's the CA of the local cluster that we copy into the remote namespace?

pkg/controller/elasticsearch/remotecluster/remoteca/controller.go
```diff
@@ -0,0 +1,474 @@
...
+		UpdateReconciled: func() {
+			reconciled.Labels = maps.Merge(reconciled.Labels, expected.Labels)
+			reconciled.Data = expected.Data
+		},
+	})
+}
+
+func caCertMissingError(location string, cluster types.NamespacedName) string {
+	return fmt.Sprintf(
+		CaCertMissingError,
```
sebgl: I think I would be fine with hardcoding the constant here to make the function easier to read

pkg/controller/elasticsearch/remotecluster/remoteca/labels.go
```diff
@@ -0,0 +1,62 @@
...
+	"github.com/elastic/cloud-on-k8s/pkg/controller/common"
+	"github.com/elastic/cloud-on-k8s/pkg/controller/elasticsearch/label"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	// RemoteClusterNamespaceLabelName used to represent the namespace of the RemoteCluster in a TrustRelationship.
+	RemoteClusterNamespaceLabelName = "remotecluster.k8s.elastic.co/namespace"
```
sebgl: this should probably have `elasticsearch.k8s.elastic.co` in the name.
Can we do `elasticsearch.k8s.elastic.co/remote-cluster: {"namespace": "ns", "name": "cluster-name"}`?

pkg/controller/elasticsearch/remotecluster/remoteca/labels.go
```diff
@@ -0,0 +1,62 @@
...
+	// RemoteClusterNameLabelName used to represent the name of the RemoteCluster in a TrustRelationship.
+	RemoteClusterNameLabelName = "remotecluster.k8s.elastic.co/name"
+	// TypeLabelValue is a type used to identify a Secret which contains the CA of a remote cluster.
+	TypeLabelValue = "remote-ca"
+	// certsInternalSecretName is the suffix added to the aforementioned Secret.
+	certsInternalSecretName = "remote-ca"
```
sebgl: `remoteCASecretSuffix` is maybe a better better

pkg/controller/elasticsearch/remotecluster/remoteca/labels.go
```diff
@@ -0,0 +1,62 @@
...
+	owner *esv1.Elasticsearch,
+	remote types.NamespacedName,
+) metav1.ObjectMeta {
+	return metav1.ObjectMeta{
+		Name:      name,
+		Namespace: owner.Namespace,
+		Labels: map[string]string{
+			RemoteClusterNamespaceLabelName: remote.Namespace,
+			RemoteClusterNameLabelName:      remote.Name,
+			label.ClusterNameLabelName:      owner.Name,
```
sebgl: is the CA in this secret the CA of the remote cluster, or the CA of the owner cluster?

pkg/controller/elasticsearch/services/services.go
```diff
@@ -9,27 +9,66 @@ import (
...
 	globalServiceSuffix = ".svc"
 )
 
+// TransportServiceName returns the name for the transport service associated to this cluster
+func TransportServiceName(esName string) string {
+	return esv1.TransportService(esName)
+}
+
+// NewTransportService returns the transport service associated with the given cluster.
+// It is used by Elasticsearch nodes to talk to each other.
```
sebgl: Only in the remote cluster case?
Otherwise elasticsearch nodes discover each other through the seed hosts configmap.

pkg/controller/elasticsearch/services/services.go
```diff
@@ -9,27 +9,66 @@ import (
...
+
 // ExternalServiceName returns the name for the external service
 // associated to this cluster
 func ExternalServiceName(esName string) string {
 	return esv1.HTTPService(esName)
 }
 
+// ExternalTransportServiceHostname returns the hostname used to reach Elasticsearch's transport endpoint.
+func ExternalTransportServiceHostname(es types.NamespacedName) string {
+	return stringsutil.Concat(TransportServiceName(es.Name), ".", es.Namespace, globalServiceSuffix, ":", strconv.Itoa(network.TransportPort))
```
sebgl: The function says `hostname` but the returned URL also contains a port

pkg/controller/elasticsearch/certificates/remoteca/reconcile.go
```diff
@@ -0,0 +1,86 @@
...
+	// Get all the remote certificate authorities
+	var remoteCAList v1.SecretList
+	if err := c.List(
+		&remoteCAList,
+		client.InNamespace(es.Namespace),
+		remoteca.GetRemoteCaMatchingLabel(es.Name),
+	); err != nil {
+		return err
+	}
+	remoteCertificateAuthorities := []byte{}
```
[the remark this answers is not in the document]
pebrc: also, could this not just be `var remoteCertificateAuthorities []byte`

pkg/controller/elasticsearch/remotecluster/annotations.go
```diff
@@ -0,0 +1,73 @@
...
+	serializedRemoteClusters, ok := es.Annotations[RemoteClustersAnnotationName]
+	if !ok {
+		return nil, nil
+	}
+	var remoteClustersArray []remoteClusterState
+	if err := json.Unmarshal([]byte(serializedRemoteClusters), &remoteClustersArray); err != nil {
+		return nil, err
+	}
+
+	remoteClusters := make(map[string]string)
```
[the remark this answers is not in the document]
barkbay: 👍 on storing the map.
The idea with the hash is to not store all the configuration in the annotation: a hash is cheap to compute, collision probabilities are very low and it eventually ends up in `etcd` so I would avoid to duplicate and store the configuration of all the remote clusters in an annotation if we can.

No strong opinion and happy to reconsider if you think it worth it.

pkg/controller/elasticsearch/remotecluster/elasticsearch.go
```diff
@@ -0,0 +1,134 @@
...
+	es esv1.Elasticsearch,
+) error {
+	span, _ := apm.StartSpan(ctx, "update_remote_clusters", tracing.SpanTypeApp)
+	defer span.End()
+
+	enabled, err := licenseChecker.EnterpriseFeaturesEnabled()
+	if err != nil {
+		return err
+	}
+	if !enabled {
```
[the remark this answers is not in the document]
barkbay: I didn't do that on purpose because I don't know what would be the technical consequences for the follower indices if it's only a temporary situation (i.e.  license expired but may be renewed shortly).

pkg/controller/elasticsearch/remotecluster/remoteca/labels.go
barkbay: We copy the CA of the remote clusters, but the owner is the local one.

pkg/controller/elasticsearch/remotecluster/remoteca/controller.go
barkbay: `getCurrentRemoteCertificateAuthorities` is the counter part of `getExpectedRemoteCertificateAuthorities`. I think it is more explicit if we keep the current naming (I will rename it to getActualRemoteCertificateAuthorities).

pkg/controller/elasticsearch/remotecluster/remoteca/controller.go
barkbay: Even if the connection is unidirectional both certificate authorities must be copied.  So, Yes, the local cluster must become the remote one at some point.

I'll try to add more comments.

pkg/controller/elasticsearch/remotecluster/remoteca/watches.go
barkbay: Both are watched:
* The local CA (from the ES of the Spec) is watched to update the trusted certificates in the remote clusters.
* The remote CAs (from the ES clusters referenced in `Spec.RemoteCluster`) are watched to update the trusted certificates of the local cluster.

pkg/controller/elasticsearch/remotecluster/remoteca/labels.go
barkbay: I'm not sure we can put some json in a label (also I'm not sure it's easy to create a label selector against it)

pkg/controller/elasticsearch/remotecluster/remoteca/watches.go
```diff
@@ -0,0 +1,108 @@
...
+			return nil
+		}
+		clusterAssociationName, ok := labels[RemoteClusterNameLabelName]
+		if !ok {
+			return nil
+		}
+		clusterAssociationNamespace, ok := labels[RemoteClusterNamespaceLabelName]
+		if !ok {
+			return nil
+		}
```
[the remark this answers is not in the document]
barkbay: 👍 but since it would be useful elsewhere (e.g. in the Kibana assoc. controller) I would be in favour of doing that in a dedicated follow-up pr.

barkbay:
I have updated the PR description with the new spec format, ICYMI here is an example:

```yaml
kind: Elasticsearch
metadata:
  name: cluster-1
  namespace: ns1
spec:
  version: 7.5.2
  remoteClusters:
  - name: ns2-cluster-2
    elasticsearchRef:
      name: cluster-2
      namespace: ns2
  - elasticsearchRef:
      name: cluster-3
      namespace: ns3
    name: ns3-cluster-3
  nodeSets:
    - name: default
      count: 3
      config:
        node.store.allow_mmap: false
```

barkbay:
jenkins test this please

barkbay:
@sebgl I split the remote CA reconciler and add some comments, I also tried to address most of your comments.

pkg/controller/remoteca/remoteca_controller.go
```diff
@@ -0,0 +1,38 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+var (
+	name = "remoteca-controller"
+)
+
+// Add creates a new RemoteCa Controller and adds it to the manager with default RBAC.
+func Add(mgr manager.Manager, accessReviewer rbac.AccessReviewer, params operator.Parameters) error {
```
pebrc: Any particular reason why you did not move everything from `elasticsearch/remotecluster/remoteca` into this package?

pkg/controller/elasticsearch/remotecluster/elasticsearch.go
```diff
@@ -0,0 +1,129 @@
...
+		}
+	}
+	return remoteClusters
+}
+
+// updateSettings makes a call to an Elasticsearch cluster to apply a persistent setting.
+func updateSettings(esClient esclient.Client, remoteClusters map[string]esclient.RemoteCluster) error {
+	ctx, cancel := context.WithTimeout(context.Background(), esclient.DefaultReqTimeout)
+	defer cancel()
+	return esClient.UpdateSettings(ctx, esclient.Settings{
```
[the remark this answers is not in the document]
barkbay: I didn't know that. It does not seem dynamic, a cluster restart would be needed ?

pkg/controller/remoteca/remoteca_controller.go
barkbay: At first it was only in `elasticsearch/remotecluster/remoteca` The reason is that this controller only makes sense in the context of Elasticsearch.

I moved a part of it to make it "[visible](https://github.com/elastic/cloud-on-k8s/pull/2543#discussion_r382588005)", I'm not sure I would go further because this controller  is dependent on the Elasticsearch one, but I'm happy to move everything if you think it makes sense.

pkg/controller/elasticsearch/remotecluster/elasticsearch.go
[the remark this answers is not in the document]
pebrc: Yes, I think we would restart the nodes every time a user would change the remote cluster configuration. So it's a trade-off between avoiding blocking API calls from the operator and potential disruption of production workloads managed by ECK because we restart nodes too often. 

So +1 on keeping it the way it is now.

pkg/controller/elasticsearch/remotecluster/remoteca/controller.go
```diff
@@ -0,0 +1,289 @@
...
+		esRef := remoteCluster.ElasticsearchRef.WithDefaultNamespace(associatedEs.Namespace)
+		expectedRemoteClusters[esRef.NamespacedName()] = struct{}{}
+	}
+
+	var list esv1.ElasticsearchList
+	if err := c.List(&list, &client.ListOptions{}); err != nil {
+		return nil, err
+	}
+
+	// Seek for Elasticsearch resources where this cluster is declared as a remote cluster
```
sebgl: I'm not sure I understand this bit. Why do we need to retrieve other references?

I thought it would be enough, when doing a reconciliation for `clusterA`, to reconcile all certificate secrets for `clusterA`. That is: certificates of remote clusters in the namespace of `clusterA` and certificates of `clusterA` in the remote clusters namespaces.

If `clusterB` declares `clusterA` as remote, secrets are reconciled in the `clusterB` association reconciliation. Why does reconciliation for `clusterA` need to care?
Aren't we reconciling the same secrets multiple times for different clusters?

pkg/controller/elasticsearch/remotecluster/remoteca/secret.go
```diff
@@ -0,0 +1,185 @@
...
+		if !errors.IsNotFound(err) {
+			return results.WithError(err)
+		}
+		results.WithResult(defaultRequeue)
+	}
+
+	return nil
+}
+
+// copyCertificateAuthority create a copy of the CA from a source cluster to an other one
```
sebgl: `creates`
`to a target cluster`

sebgl:
Thanks for making the changes I understood the code much better now regarding what's remote vs. local and why we reconcile 2 secrets per association.

Most of my comments above are nitpicks. Except [this one](https://github.com/elastic/cloud-on-k8s/pull/2543#discussion_r385243070) that I'd like to understand. I think we are reconciling the same secret multiple times in different reconciliations targeting different clusters and we may not have to?

pkg/controller/elasticsearch/remotecluster/remoteca/controller.go
```diff
@@ -0,0 +1,289 @@
...
+	return results.WithResult(association.RequeueRbacCheck(r.accessReviewer)).Aggregate()
+}
+
+func caCertMissingError(cluster types.NamespacedName) string {
+	return fmt.Sprintf("Cannot find CA certificate cluster %s/%s", cluster.Namespace, cluster.Name)
+}
+
+// getExpectedRemoteCertificateAuthorities returns all the remote cluster keys for which a remote ca should created
+// The CA certificates must be copied from the remote cluster to the local one and vice versa
+func getExpectedRemoteCertificateAuthorities(
```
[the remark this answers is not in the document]
barkbay: `CA` is singular, `createOrUpdateCertificateAuthorities` makes it explicit that we are manipulating several CAs

pkg/controller/elasticsearch/remotecluster/remoteca/controller.go
barkbay: The reason is that `remoteClustersInvolvedWith` returns all the clusters for which an association has been established (This is needed to ensure that we remove all the secrets which are not needed anymore, not only the ones that are in the local namespace)

If we don't do the same here the remote CA will be deleted when the remote cluster is reconciled because it will not be expected: 

```golang
expectedRemoteClusters, err := getExpectedRemoteClusters(ctx, r.Client, localEs)
...
remoteClustersInvolved, err := remoteClustersInvolvedWith(ctx, r.Client, localClusterKey)
...
// Create or update expected remote CA
for remoteEsKey := range expectedRemoteClusters {
         ...
        delete(remoteClustersInvolved, remoteEsKey
}
// Delete existing but not expected remote CA
for toDelete := range remoteClustersInvolved { // Here
		results.WithError(deleteCertificateAuthorities(ctx, r, localClusterKey, toDelete))
}
```

If both clusters are reciprocally declared as "remote", it will induce an extra reconcile loop but since everything is already set and in cache I'm not sure that we should care too much.

pkg/controller/elasticsearch/remotecluster/remoteca/controller.go
```diff
@@ -0,0 +1,289 @@
...
+
+	return expectedRemoteClusters, nil
+}
+
+// getActualRemoteCertificateAuthorities returns for a given Elasticsearch cluster all the Elasticsearch keys for which
+// the remote certificate authorities have been copied, i.e. all the other Elasticsearch clusters for which this cluster
+// has been involved in a remote cluster association.
+// In order to get all of them we:
+// 1. List all the remote CA copied locally.
+// 2. List all the other Elasticsearch clusters for which the CA of the given cluster has been copied.
```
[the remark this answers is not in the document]
barkbay: Absolutely, I want to cover the case where a Secret is left over in a remote namespace (see my other comment)

pkg/controller/elasticsearch/remotecluster/remoteca/controller.go
sebgl: Got it. I guess the only way around would to label differently a CA that is a remote in another namespace and a CA that is a remote in the current namespace (it's very hard to explain this concept with words :()
Anyway I'm fine with the current approach, as you said an extra reconciliation on secrets we already have in the cache should not cost much.

barkbay:
@pebrc @sebgl It was a large PR, thank you both for your thorough code review ! ❤️

PR #2546, opened 2020-02-12
Make stack version 7.6.0 the new default

sebgl:
Upgrade references to the stack version from 7.5.0 to 7.6.0.

### DIFF
--- a/build/ci/e2e/stack_versions.jenkinsfile
+++ b/build/ci/e2e/stack_versions.jenkinsfile
@@ -80,14 +80,14 @@ pipeline {
                         }
                     }
                 }
-                stage("7.5.0") {
+                stage("7.6.0") {
                     agent {
                         label 'linux'
                     }
                     steps {
                         checkout scm
                         script {
-                            runWith(lib, failedTests, "eck-75-${BUILD_NUMBER}-e2e", "7.5.0")
+                            runWith(lib, failedTests, "eck-76-${BUILD_NUMBER}-e2e", "7.6.0")
                         }
                     }
                 }
--- a/.ci/packer_cache.sh
+++ b/.ci/packer_cache.sh
@@ -17,6 +17,6 @@ docker pull kindest/node:v1.16.4
 docker pull kindest/node:v1.17.0
 
 # Elastic Stack images
-docker pull docker.elastic.co/elasticsearch/elasticsearch:7.4.0
-docker pull docker.elastic.co/kibana/kibana:7.4.0
-docker pull docker.elastic.co/apm/apm-server:7.4.0
+docker pull docker.elastic.co/elasticsearch/elasticsearch:7.6.0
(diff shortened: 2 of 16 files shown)


build/ci/e2e/stack_versions.jenkinsfile
```diff
@@ -80,14 +80,14 @@ pipeline {
...
-                stage("7.5.0") {
+                stage("7.6.0") {
                     agent {
                         label 'linux'
                     }
                     steps {
                         checkout scm
                         script {
-                            runWith(lib, failedTests, "eck-75-${BUILD_NUMBER}-e2e", "7.5.0")
+                            runWith(lib, failedTests, "eck-76-${BUILD_NUMBER}-e2e", "7.6.0")
```
sebgl: This replaces our 7.5.0 E2E test by a 7.6.0 test.
We don't have a very structured way of choosing which versions should be automatically tested, in the meantime I think it's OK to not test them all?

thbkrkr:
Some remaining occurrences:
```sh
> grep '7.5' -r config 
config/e2e/filebeat.yaml:          image: docker.elastic.co/beats/filebeat:7.5.2
config/recipes/associations-rbac/apm_es_kibana_rbac.yaml:  version: 7.5.2
config/recipes/associations-rbac/apm_es_kibana_rbac.yaml:  version: 7.5.2
config/recipes/associations-rbac/apm_es_kibana_rbac.yaml:  version: 7.5.2
```

thbkrkr:
I did not look well and found a few more versions that could be updated.

2 default versions which should never be used since we are always supposed to specify one. But does it seems safer and cleaner to update them?

https://github.com/elastic/cloud-on-k8s/blob/09be60c9731520cf329d390877e73a0a2bd96221/test/e2e/cmd/run/command.go#L70

https://github.com/elastic/cloud-on-k8s/blob/09be60c9731520cf329d390877e73a0a2bd96221/test/e2e/test/context.go#L20

2 used in the E2E tests that we probably forgot to update in 7.5.2.

https://github.com/elastic/cloud-on-k8s/blob/09be60c9731520cf329d390877e73a0a2bd96221/test/e2e/kb/testdata/kibana_standalone.yaml#L7

https://github.com/elastic/cloud-on-k8s/blob/09be60c9731520cf329d390877e73a0a2bd96221/test/e2e/kb/testdata/kibana_standalone.yaml#L21

The version used to cache Docker images on the Jenkins workers.

https://github.com/elastic/cloud-on-k8s/blob/09be60c9731520cf329d390877e73a0a2bd96221/.ci/packer_cache.sh#L20-L22

A last one which is a little out of our scope but which I hope is not a problem if we update it.

https://github.com/elastic/cloud-on-k8s/blob/09be60c9731520cf329d390877e73a0a2bd96221/config/recipes/beats/2_filebeat-kubernetes.yaml#L54

PR #2553, opened 2020-02-12
Example fix: dynamic heap size is not recommend

mjmbischoff:
Elastic recommends setting the same value for xms and xmx. The example therefor might be confusing. 

See: https://www.elastic.co/guide/en/elasticsearch/reference/current/heap-size.html

> Elasticsearch will assign the entire heap specified in jvm.options via the Xms (minimum heap size) and Xmx (maximum heap size) settings. _You should set these two settings to be equal to each other._


sebgl:
@thbkrkr it's actually mentioned there: https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-managing-compute-resources.html#k8s-compute-resources-elasticsearch
I'm creating another issue to improve our current docs: the heap size and memory reqs doc is spread over 3 different pages.

PR #2566, opened 2020-02-14
Always include operator namespace in cached k8s client namespaces

pebrc:
The trial license e2e test was failing since #2530 because we are now running the e2e with a operator restricted to the managed namespaces via `--namespaces`, but we did not include the operator namespace itself in the list of namespaces. 

As the list of namespaces controls which resources will be cached by the k8s client the operator was not able to access the license related secrets in the operator namespace and trial mode could not be activated.

This PR makes sure the operator namespace is always included in the client cache even when running in namespace restricted mode.


thbkrkr:
I wonder if there is a good reason for not always doing it directly durint the operator initialization:

```diff
diff --git a/cmd/manager/main.go b/cmd/manager/main.go
index 62ede14a..b79277eb 100644
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -238,16 +238,19 @@ func execute() {
        }
 
        // configure the manager cache based on the number of managed namespaces
-       managedNamespaces := viper.GetStringSlice(operator.NamespacesFlag)
-       switch len(managedNamespaces) {
-       case 0:
+       managedNs := viper.GetStringSlice(operator.NamespacesFlag)
+       nbManagedNs := len(managedNs)
+       if nbManagedNs == 0 {
                log.Info("Operator configured to manage all namespaces")
-       case 1:
-               log.Info("Operator configured to manage a single namespace", "namespace", managedNamespaces[0])
-               opts.Namespace = managedNamespaces[0]
-       default:
-               log.Info("Operator configured to manage multiple namespaces", "namespaces", managedNamespaces)
-               opts.NewCache = cache.MultiNamespacedCacheBuilder(managedNamespaces)
+       } else {
+               if nbManagedNs == 1 {
+                       log.Info("Operator configured to manage a single namespace", "namespace", managedNs[0])
+               } else {
+                       log.Info("Operator configured to manage multiple namespaces", "namespaces", managedNs)
+               }
+               //  namespaces the k8s client can access objects in should include the managed namespaces and the namespace the
+               //  operator resides in, to be able to fetch internal secrets
+               opts.NewCache = cache.MultiNamespacedCacheBuilder(append(managedNs, operatorNamespace))
        }
 
        // only expose prometheus metrics if provided a non-zero port

```

thbkrkr:
Jenkins test this please

anyasabo:
I was wondering if we needed to update https://github.com/elastic/cloud-on-k8s/blob/master/docs/operator-config.asciidoc indicating that no matter what, the operator always needs to be monitoring its own namespace. After thinking about it though I don't think this will be surprising to people, even if they do not know about the licensing configmap.

PR #2569, opened 2020-02-14
Make licence file discovery deterministic

charith-elastic:
The directory traversal code to detect licences used the [`Unsorted` option](https://godoc.org/github.com/karrick/godirwalk#Options) to make the traversal faster but it makes results OS-dependent (most likely the cause for CI failures in #2497). 

This PR sets the `Unsorted` flag to `false` and adds test data to mimic the failure cause in #2497.


charith-elastic:
Jenkins test this please

PR #2578, opened 2020-02-18
Switch API doc generation to crd-ref-docs

### LINKED ISSUE #2109: API doc generation fails when project not in GOPATH
If the project is not checked out in `GOPATH` the api doc generation step fails with 
```
>> Generating API reference documentation...
F1111 15:20:43.219462    8904 main.go:134] no API packages found in ./pkg/apis
goroutine 1
```

After discussing with @charith-elastic it seems that 

>When outside the GOPATH, the generator only works if both the following conditions are met:
> 1. cloud-on-k8s and gen-crd-api-reference-docs must have a common parent directory 
> 2. The generator executable must be called from the gen-crd-api-reference-docs directory

charith-elastic:
Switches to using [crd-ref-docs](https://github.com/elastic/crd-ref-docs) for API documentation generation. The build process should now work outside the GOPATH and be much faster as well.

Fixes #2109


hack/api-docs/build.sh
```diff
@@ -5,56 +5,43 @@
 # you may not use this file except in compliance with the Elastic License.
 
 # Script to generate API reference documentation from the source code.
+# To test with a different version of crd-ref-docs while retaining the binary, invoke the script as follows:
+# SCRATCH_DIR=/tmp/crd-ref-docs-tmp CLEANUP=false REFDOCS_VER=e36d311 ./build.sh
```
barkbay: nit: `SCRATCH_DIR` must exist otherwise the script fails at line 33

PR #2579, opened 2020-02-18
Add GKE private cluster docs

anyasabo:
Following up on https://discuss.elastic.co/t/elastic-operator-requests-regularly-time-out/219096/

We can be a little more explicit in our docs since it is unintuitive, not documented with GKE, and a pain to troubleshoot.

### DIFF
--- a/docs/troubleshooting.asciidoc
+++ b/docs/troubleshooting.asciidoc
@@ -247,6 +247,8 @@ On startup, the operator deploys an https://kubernetes.io/docs/reference/access-
 
 For troubleshooting, you can change the https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#failure-policy[`failurePolicy`] of the webhook configuration to `Fail`, which will cause creations and updates to error out if there is an error contacting the webhook.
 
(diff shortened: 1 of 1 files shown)


docs/troubleshooting.asciidoc
```diff
@@ -247,6 +247,8 @@ On startup, the operator deploys an https://kubernetes.io/docs/reference/access-
 
 For troubleshooting, you can change the https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#failure-policy[`failurePolicy`] of the webhook configuration to `Fail`, which will cause creations and updates to error out if there is an error contacting the webhook.
 
+On link:https://cloud.google.com/kubernetes-engine/docs/concepts/private-cluster-concept[GKE private clusters], you will need to add a firewall rule allowing port 9443 from the API server so that it can contact the webhook. See the link:https://cloud.google.com/kubernetes-engine/docs/how-to/private-clusters#add_firewall_rules[GKE docs on adding rules] and the link:https://github.com/kubernetes/kubernetes/issues/79739[Kubernetes issue] for more detail. Otherwise, you may see requests modifying Elastic resources timeout or take a very long time to complete.
```
charith-elastic:
```suggestion
If your requests take too long or timeout on link:https://cloud.google.com/kubernetes-engine/docs/concepts/private-cluster-concept[GKE private clusters], you will need to add a firewall rule allowing the API server access to port 9443 so that it can contact the webhook. See the link:https://cloud.google.com/kubernetes-engine/docs/how-to/private-clusters#add_firewall_rules[GKE documentation on adding firewall rules] and the link:https://github.com/kubernetes/kubernetes/issues/79739[Kubernetes issue] for more information.
```

anyasabo: I like having the "On GKE private clusters" first so that people not on it can skip that section. I did reword it based on your suggestion though, WDYT?

charith-elastic: That's fine. It feels a bit awkward to read "...requests modifying Elastic resources..." and might confuse some people because they may not see the creation of a resource as an act of modification. 
Finally, I'd say "GKE documentation" instead of "GKE docs" because the latter feels a bit too informal.

PR #2577, opened 2020-02-18
Update deployer docs

david-kow:
An attempt to make deployer docs clearer - moved deployer prerequisites description to deployer readme and separated out non-Makefile deployer usage.

PR #2585, opened 2020-02-20
Small fixes to CI docs

david-kow:
- point from CI docs to deployer as CI bits in deployer readme were removed in https://github.com/elastic/cloud-on-k8s/pull/2577
- fix `.ci/Makefile` to properly inject credentials to `deployer-config.yml` created by `.ci/setenvconfig`

### DIFF
--- a/.ci/Makefile
+++ b/.ci/Makefile
@@ -19,7 +19,7 @@ VAULT_TOKEN = $(shell vault write -address=$(VAULT_ADDR) -field=token auth/appro
 else
 VAULT_TOKEN = $(shell vault write -address=$(VAULT_ADDR) -field=token auth/github/login token=$(GITHUB_TOKEN))
 # we use roleId as a string that has to be there for authn/z for CI, but it's empty and not needed for local execution
-NOT_USED := $(shell test -e $(ROOT_DIR)/deployer-config.yml && sed -i '' -e "s;roleId: dev;token: $(GITHUB_TOKEN);g" $(ROOT_DIR)/deployer-config.yml)
+NOT_USED := $(shell test -e $(ROOT_DIR)/deployer-config.yml && sed "s;roleId:.*;token: $(GITHUB_TOKEN);g" $(ROOT_DIR)/deployer-config.yml > tmp && mv tmp deployer-config.yml)
 endif
 endif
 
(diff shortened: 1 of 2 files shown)


.ci/Makefile
```diff
@@ -19,7 +19,7 @@ VAULT_TOKEN = $(shell vault write -address=$(VAULT_ADDR) -field=token auth/appro
 else
 VAULT_TOKEN = $(shell vault write -address=$(VAULT_ADDR) -field=token auth/github/login token=$(GITHUB_TOKEN))
 # we use roleId as a string that has to be there for authn/z for CI, but it's empty and not needed for local execution
-NOT_USED := $(shell test -e $(ROOT_DIR)/deployer-config.yml && sed -i -e "s;roleId:;token: $(GITHUB_TOKEN);g" $(ROOT_DIR)/deployer-config.yml)
+NOT_USED := $(shell test -e $(ROOT_DIR)/deployer-config.yml && sed -i '' -e "s;roleId: dev;token: $(GITHUB_TOKEN);g" $(ROOT_DIR)/deployer-config.yml)
```
thbkrkr: I think `''` makes `sed` believe this is the path of an empty file for me.

```sh
> sed --version | head -1
sed (GNU sed) 4.4
```

```suggestion
NOT_USED := $(shell test -e $(ROOT_DIR)/deployer-config.yml && sed -i -e "s;roleId:.*;token: $(GITHUB_TOKEN);g" $(ROOT_DIR)/deployer-config.yml)
```

thbkrkr: According to [stackoverflow](https://stackoverflow.com/questions/5694228/sed-in-place-flag-that-works-both-on-mac-bsd-and-linux), the only way to have a sed in-place command that works both for the GNU and the BSD version of `sed` is to set a dot prefix without space:

```suggestion
NOT_USED := $(shell test -e $(ROOT_DIR)/deployer-config.yml && sed -i.bak "s;roleId:.*;token: $(GITHUB_TOKEN);g" $(ROOT_DIR)/deployer-config.yml)
```

Otherwise, I'm also fine with using an intermediate temporary file:
 `sed ... > tmp.yml && mv tmp.yml > deployer-config.yml`.

```suggestion
NOT_USED := $(shell test -e $(ROOT_DIR)/deployer-config.yml && sed "s;roleId:.*;token: $(GITHUB_TOKEN);g" > tmp.yml && mv tmp.yml > $(ROOT_DIR)/deployer-config.yml)
```


Also, I think we can safely remove the `-e` as we only use one expression and search for `roleId: .*` instead of `roleId: dev` to be more radical.

david-kow: Behavior differs between GNU and BSD seds, so moving to simpler sed invocation and using temp file.

PR #2588, opened 2020-02-20
New make target to ease job chaining in CI

thbkrkr:
This commit fixes how the name of the operator image is retrieved in the
nightly and the release build jobs in order to trigger the different e2e
tests jobs.

Since #2502, we now rely on the `OPERATOR_IMAGE` variable computed
in the `Makefile` and stop to write it manually in the `.env` file.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,11 +18,7 @@ KUBECTL_CLUSTER := $(shell kubectl config current-context 2> /dev/null)
 LOG_VERBOSITY ?= 1
 
 # Get the currently used golang install path (in GOPATH/bin, unless GOBIN is set)
-ifeq ($(shell go env GOBIN 2>/dev/null),)
-GOBIN=$(shell go env GOPATH)/bin
-else
-GOBIN=$(shell go env GOBIN)
-endif
+GOBIN := $(or $(shell go env GOBIN 2>/dev/null), $(shell go env GOPATH 2>/dev/null)/bin)
 
 # find or download controller-gen
 # note this does not validate the version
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -18,7 +18,7 @@ KUBECTL_CLUSTER := $(shell kubectl config current-context 2> /dev/null)
 LOG_VERBOSITY ?= 1
 
 # Get the currently used golang install path (in GOPATH/bin, unless GOBIN is set)
-ifeq ($(shell go env GOBIN),)
+ifeq ($(shell go env GOBIN 2>/dev/null),)
```
charith-elastic: I know this is not strictly related to your PR but what do you think about replacing these four lines with:

```sh
GOBIN:=$(or $(shell go env GOBIN), $(shell go env GOPATH)/bin)
```

Makefile
```diff
@@ -58,6 +58,8 @@ IMG_VERSION ?= $(VERSION)-$(TAG)
 BASE_IMG       := $(REGISTRY)/$(REPOSITORY)/$(IMG_NAME)
 OPERATOR_IMAGE ?= $(BASE_IMG):$(IMG_VERSION)
 
+show-operator-image:
```
charith-elastic: Nit: `print-operator-image-name`  might be more descriptive of what this target does.

thbkrkr: I tried to be consistent with the `show-image` target of the `.ci/Makefile` but I prefer `print-`. Let's rename the two targets. 
I feel that the `name` suffix is a little noisy. We could also argue that the full image name is composed by a registry name, a repository name and an image name, so it's always a little ambiguous. 
I'm going to keep `print-ci-image` and `print-operator-image`.

Makefile
thbkrkr: I like this more concise syntax.
However, I want to make it silent if go is not installed because this is what happens when we call `make show-operator-image` outside the ci container to trigger other jobs here:

https://github.com/elastic/cloud-on-k8s/blob/ba1ea9820af67efaa95c77e5ab03191a94d64f15/.ci/pipelines/build.Jenkinsfile#L70-L77

PR #2596, opened 2020-02-21
Clean up unnecessary env var for e2e/aks

thbkrkr:
In 7684771, I have badly move the config for the deployer from the `Jenkinsfile` to the `setenvconfig` script. 

AKS does not use and does not need GKE. :confounded:


thbkrkr:
jenkins test this please

Issue #2602, opened 2020-02-24
Flaky E2E test TestDeleteServices


sebgl:
This test failed twice recently:

```
=== RUN   TestDeleteServices/ES_services_should_have_endpoints#01
Retries (5m0s timeout): ....................................................................................................
{"log.level":"error","@timestamp":"2020-02-24T01:06:57.832Z","message":"stopping early","service.version":"0.0.0-00000000","service.type":"eck","ecs.version":"1.4.0","error":"test failure","error.stack_trace":"github.com/go-logr/zapr.(*zapLogger).Error\n\t/go/pkg/mod/github.com/go-logr/zapr@v0.1.0/zapr.go:128\ngithub.com/elastic/cloud-on-k8s/test/e2e/test.StepList.RunSequential\n\t/go/src/github.com/elastic/cloud-on-k8s/test/e2e/test/step.go:42\ngithub.com/elastic/cloud-on-k8s/test/e2e/test.runFailureScenario\n\t/go/src/github.com/elastic/cloud-on-k8s/test/e2e/test/run_failure.go:56\ngithub.com/elastic/cloud-on-k8s/test/e2e/test.RunRecoverableFailureScenario\n\t/go/src/github.com/elastic/cloud-on-k8s/test/e2e/test/run_failure.go:19\ngithub.com/elastic/cloud-on-k8s/test/e2e/es.TestDeleteServices\n\t/go/src/github.com/elastic/cloud-on-k8s/test/e2e/es/failure_test.go:66\ntesting.tRunner\n\t/usr/local/go/src/testing/testing.go:909"}
    --- FAIL: TestDeleteServices/ES_services_should_have_endpoints#01 (300.00s)
        utils.go:83: 
            	Error Trace:	utils.go:83
            	Error:      	Received unexpected error:
            	            	endpoints "test-failure-delete-services-5zrb-es-http" not found
            	Test:       	TestDeleteServices/ES_services_should_have_endpoints#01
```

Looking at the diagnostic bundle for that E2E test run, I can see that a service `test-failure-delete-services-5zrb-es-http` correctly exists.
According to the operator logs it does not have any endpoint available (which is also reported by the E2E test itself):

```
[...]

PR #2603, opened 2020-02-24
Get endpoints as part of the diagnostics bundle

sebgl:
Some flaky E2E tests report no endpoints exist for the Elasticsearch
HTTP service. This may help to understand what is happening.

May help with https://github.com/elastic/cloud-on-k8s/issues/2602.

### DIFF
--- a/hack/diagnostics/eck-dump.sh
+++ b/hack/diagnostics/eck-dump.sh
@@ -110,6 +110,7 @@ main() {
     get_resources $ns persistentvolumes
     get_resources $ns persistentvolumeclaims
     get_resources $ns services
+    get_resources $ns endpoints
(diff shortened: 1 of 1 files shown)


hack/diagnostics/eck-dump.sh
```diff
@@ -110,6 +110,7 @@ main() {
     get_resources $ns persistentvolumes
     get_resources $ns persistentvolumeclaims
     get_resources $ns services
+    get_resources $ns endpoints
```
pebrc: I wonder if dumping this can become a problem on large installations.

sebgl: You mean because of the high number of items?
It's probably the same cardinality as for Pods.

PR #2611, opened 2020-02-24
Generate documentation page for third-party dependencies 

charith-elastic:
Generates a documentation page containing third-party dependency information.

This change refactors the licence-detector to add the following functionality:
- Detect licence type of each dependency.
- Allow manual overrides to provide licence and URL information for dependencies that cannot be scanned automatically.
- Add template for dependency list page.
- Force refresh of Go module cache to obtain the latest information about dependencies
- Add script to generate dependency information about the ECK container image
- Add validation option to ensure that all dependency URLs are valid

Dependency page preview: http://cloud-on-k8s_2611.docs-preview.app.elstc.co/guide/en/cloud-on-k8s/master/k8s-dependencies.html

Fixes elastic/k8s-dev#103


charith-elastic:
Jenkins test this please

charith-elastic:
Jenkins test this please

hack/licence-detector/README.md
```diff
@@ -0,0 +1,76 @@
+Licence Detector
+================
+
+This directory contains the scripts to generate licence notice and dependency information documentation.
+
+- `generate-notice.sh`: Invoked by `make generate-notice-file` to automatically generate `NOTICE.txt` and `docs/dependencies.asciidoc`.
+- `generate-image-deps.sh`: Manually invoked script to update the container image dependency information in `docs/container-image-dependencies.csv`.
```
anyasabo: Can you clarify what the CSV is intended for? It doesn't look like it's used as an input to anything else. I was also a little surprised when I was testing it out and it needed to sudo

hack/licence-detector/dependency/dependency.go
```diff
@@ -0,0 +1,80 @@
...
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+
+	securejoin "github.com/cyphar/filepath-securejoin"
```
anyasabo: I learned about a new lib, nice

hack/licence-detector/dependency/dependency.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+	defer f.Close()
+
+	rootDir, err := filepath.Abs(filepath.Dir(file))
+	if err != nil {
+		return nil, fmt.Errorf("failed to determine absolute path of overrides file: %w", err)
+	}
+
+	decoder := json.NewDecoder(f)
+	for {
```
anyasabo: This might be me overthinking it, but we mostly parse it this way because we want to end up with a map for `mkDepInfo()`, correct? If we wanted a slice we could just store overrides.json as a json array and unmarshal it right into a slice (and also get to prettify the JSON file rather than making each item be on one line).

hack/licence-detector/detector/detector.go
```diff
@@ -9,54 +9,66 @@ import (
...
 	Dir      string     // directory holding files for this module, if any
-	Replace  *Module    // replace directive
+	Replace  *module    // replace directive
+}
+
+// MkClassifier creates a new instance of the licence classifier.
+func MkClassifier(dataPath string) (*licenseclassifier.License, error) {
+	absPath, err := filepath.Abs(dataPath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to determine absolute path of licence data file: %w", err)
```
anyasabo: nit: it might be worth considering using pkg/errors so we get those sweet, sweet stack traces

hack/licence-detector/README.md
charith-elastic: The CSV is `include`d by `docs/dependencies.asciidoc` and contains the licence information for the Docker image. The sudo request by Tern is indeed a bit worrying but apparently it's needed to mount the procfs file system:

> Tern requires root privileges to run because it needs to mount procfs in order to run commands within a chroot environment and call the Docker CLI. It is enough if you have configured sudo; Tern will ask for your password before running any priviledged commands.

I wish they supported OCI images rather than relying so heavily on Docker daemon and CLI but it's the only reliable tool I have managed to find. The alternative is to [manually maintain a list of dependencies in the Docker image](https://github.com/elastic/cloud/blob/f418f0ff778977078b827669c629c22983924a42/found-deployment/cloud-enterprise/baseimages/enterprise-base/Dockerfile#L15)

hack/licence-detector/dependency/dependency.go
charith-elastic: I chose the JSON lines format because it is well-known and Go supports it out of the box.
The amount of data stored on each line is pretty minimal and I don't expect it to be edited too often. If it starts to become too complex and requires pretty formatting, I think we can consider switching to a more human-friendly file format at that point.

hack/licence-detector/detector/detector.go
charith-elastic: I think `pkg/errors` is now in maintenance mode in favour of the changes to errors introduced by Go 1.13. I am using the error wrapping method introduced by 1.13 here. I prefer error messages with added context like this over stacktraces as they are easier to read and search for in the code.

hack/licence-detector/README.md
anyasabo: Got it, thanks for the clarification. I wonder if there's a clearer phrase than "container image dependency" since our direct dependencies are also included in the image (just inside the binary). Maybe changing to "container base image dependency..."?

hack/licence-detector/detector/detector.go
anyasabo: I think it's been in maintenance mode for a bit now, but you are correct 1.13 adds the error wrapping functionality that pkg/errors had. There's still no easy support afaik for including stack traces as well though. Like I said it was just a personal style suggestion, definitely not a blocker

anyasabo:
It's a little odd the base image deps are formatted differently than the module deps, but it matches the ECE doc and I expect this will probably be the least visited page in our entire docs, so it is definitely not worth changing.

hack/licence-detector/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+The licence database file `licence.db` contains all the currently known licence types found in https://github.com/google/licenseclassifier/tree/master/licenses. In the rare case that entirely new licence types have been introduced to the codebase, follow the instructions at https://github.com/google/licenseclassifier to execute the `license_serializer` tool.
+
+
+generate-image-deps.sh
+-----------------------
+
+This script generates licence information for the contents of the ECK container image. As the container base image is rarely ever changed and the tool used ([Tern](https://github.com/vmware/tern)) is slow to run, this script is not invoked automatically by the build process.
+
+To generate the dependency list (`docs/container-image-dependencies.csv`) for a particular image tag, invoke the script as follows:
```
anyasabo:
```suggestion
To generate the dependency list (`docs/container-image-dependencies.csv`) for a particular image tag, invoke the script as follows. Note that this will prompt for root and [is necessary](https://github.com/vmware/tern/blob/master/docs/releases/v0_1_0.md#notes):
```

hack/licence-detector/detector/detector.go
```diff
@@ -9,54 +9,66 @@ import (
...
 	"time"
 
+	"github.com/elastic/cloud-on-k8s/hack/licence-detector/dependency"
+	"github.com/google/licenseclassifier"
 	"github.com/karrick/godirwalk"
 )
 
-var errLicenceNotFound = errors.New("failed to detect licence")
+// detectionThreshold is the minimum confidence score required from the licence classifier.
+const detectionThreshold = 0.85
```
anyasabo: Just curious, was the 0.8 here: https://github.com/google/licenseclassifier/blob/master/classifier.go#L39
too loose?

charith-elastic: It's mainly for self-documentation here so that someone reading the code would be aware of the existence of a confidence score. Also, all of our current dependencies had scores higher than 0.85 so I thought it would be good to have a higher threshold to reduce the potential for a false positive slipping through.

charith-elastic:
> It's a little odd the base image deps are formatted differently than the module deps

Since Go dependencies have long names, having the URL as a separate column makes the table too wide for the content area -- causing the text to get squished and unreadable.

hack/licence-detector/validate/validate.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	if count := atomic.LoadUint32(&errorCount); count > 0 {
+		return fmt.Errorf("encountered %d validation errors", count)
+	}
+
+	return nil
+}
+
+func buildHTTPClient() *http.Client {
```
pebrc: naming nits: it's not clear to me when you use `build...` vs `mk...` vs `new...`. In other parts `mk...` seems just to initialise a struct or create a new instance but in this file you are actually triggering a side effect e.g. in `mkRequest`

Similarily it's not clear if this particular func could also be called `newHTTPClient`

hack/licence-detector/validate/validate.go
```diff
@@ -0,0 +1,135 @@
...
+	"net/http"
+	"runtime"
+	"sync/atomic"
+	"time"
+
+	"github.com/elastic/cloud-on-k8s/hack/licence-detector/dependency"
+	"golang.org/x/sync/errgroup"
+)
+
+const maxErrors = 12
```
pebrc: Why do we tolerate 12 errors? Maybe add a comment.

hack/licence-detector/validate/validate.go
```diff
@@ -0,0 +1,135 @@
...
+	return validateURLs(deps)
+}
+
+func validateURLs(deps *dependency.List) error {
+	ctx, cancelFunc := context.WithCancel(context.Background())
+	defer cancelFunc()
+
+	var errorCount uint32
+	depsChan := make(chan dependency.Info, 64)
+	numWorkers := runtime.NumCPU() + 2
```
pebrc: Same here: I would be interested in the rationale behind these choices

hack/licence-detector/validate/validate.go
charith-elastic: It's for circuit-breaking because in practice we should only ever see a small number of errors. I have added a comment to clarify.

PR #2613, opened 2020-02-25
Support Elastic stack version 8.0

### LINKED ISSUE #2607: Support major 8 in code
There are few places where we assume only major 6 and 7 exist (for instance [here](https://github.com/elastic/cloud-on-k8s/blob/master/pkg/controller/elasticsearch/version/supported_versions.go) and [here](https://github.com/elastic/cloud-on-k8s/blob/master/pkg/controller/elasticsearch/client/base.go)). We should start supporting 8.

pebrc:
This commit adds version 8.x to the supported versions and add a stub implementation of a `v8` Elasticsearch client that currently delegates all calls to the `v7` version.

Fixes #2607

### DIFF
--- a/pkg/apis/elasticsearch/v1/validations.go
+++ b/pkg/apis/elasticsearch/v1/validations.go
@@ -194,7 +194,7 @@ func validUpgradePath(current, proposed *Elasticsearch) field.ErrorList {
 		// this should not happen, since this is the already persisted version
 		errs = append(errs, field.Invalid(field.NewPath("spec").Child("version"), current.Spec.Version, parseStoredVersionErrMsg))
 	}
-	currVer, err := version.Parse(proposed.Spec.Version)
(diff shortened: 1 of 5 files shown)


anyasabo:
jenkins test this please

pkg/apis/elasticsearch/v1/validations.go
```diff
@@ -194,15 +194,15 @@ func validUpgradePath(current, proposed *Elasticsearch) field.ErrorList {
 		// this should not happen, since this is the already persisted version
 		errs = append(errs, field.Invalid(field.NewPath("spec").Child("version"), current.Spec.Version, parseStoredVersionErrMsg))
 	}
-	currVer, err := version.Parse(proposed.Spec.Version)
+	proposedVer, err := version.Parse(proposed.Spec.Version)
```
anyasabo: doh! good catch

Issue #2608, opened 2020-02-24
Add E2E tests for Elastic Stack 8.0.0-SNAPSHOT


david-kow:
This depends on https://github.com/elastic/cloud-on-k8s/issues/2607.

pebrc:
I ran a quick smoke test when working on #2607. While APM Server and Elasticsearch tests where succeeding, I ran into trouble with Kibana due to an optimize run that was triggered (and I don't know yet what triggered that). Will focus on #2565 to get the necessary changes merged before returning to this.

sebgl:
Depends on https://github.com/elastic/cloud-on-k8s/issues/2791.

PR #2614, opened 2020-02-25
Test snapshot versions of the stack 

### LINKED ISSUE #2565: E2E Tests: test snapshot releases of the stack
In order to improve our confidence in the compatibility of ECK with future releases of the stack we should have some e2e tests to check that ECK can deploy and manage (e.g. upgrade) current snapshot releases of the stack.

Related to:
https://github.com/elastic/cloud-on-k8s/issues/2560
https://github.com/elastic/cloud-on-k8s/issues/2253

Example of issue that should be catch in advance:
https://github.com/elastic/cloud-on-k8s/issues/2445

pebrc:
* Adds a new build stage for `7.7.0-SNAPSHOT` builds to the stack versions test run
* Parameterizes the `get-elastic-public-key` and `get-test-license` make targets to be dependent on the target version of our e2e tests i.e. use a dev key and license for snapshot builds
* APM Server does not report SNAPSHOT builds as such in the server info structs, adjust tests to not expect that and expect the version without the suffix instead

Fixes #2565 

What's not included:
* running 8.0 snapshots, I ran into some trouble with Kibana optimize runs in the current dev builds for 8.0 so I excluded it for now given that we have #2608

### DIFF
--- a/.ci/Makefile
+++ b/.ci/Makefile
@@ -56,19 +56,21 @@ ci-build-image:
 		docker push push.$(CI_IMAGE) \
 	)
 
+ifneq (,$(findstring -SNAPSHOT,$(STACK_VERSION)))
+       SECRET_FIELD_PREFIX ?= dev-
+endif
+
 ##  Build
 
 # read Elastic public key from Vault into license.key, to build the operator for E2E tests or for a release
-get-elastic-public-key: FIELD = $(if $(SNAPSHOT_TEST_TARGET),dev-pubkey,pubkey)
 get-elastic-public-key:
-	@ VAULT_TOKEN=$(VAULT_TOKEN) vault read -address=$(VAULT_ADDR) -field=$(FIELD) secret/devops-ci/cloud-on-k8s/license | base64 --decode > license.key
+	@ VAULT_TOKEN=$(VAULT_TOKEN) vault read -address=$(VAULT_ADDR) -field=$(SECRET_FIELD_PREFIX)pubkey secret/devops-ci/cloud-on-k8s/license | base64 --decode > license.key
 
 ##  Test
 
 # read some test licenses from Vault for E2E license tests
-get-test-license: FIELD = $(if $(SNAPSHOT_TEST_TARGET),dev-enterprise,enterprise)
 get-test-license:
-	@ VAULT_TOKEN=$(VAULT_TOKEN) vault read -address=$(VAULT_ADDR) -field=$(FIELD)  secret/devops-ci/cloud-on-k8s/test-licenses > test-license.json
+	@ VAULT_TOKEN=$(VAULT_TOKEN) vault read -address=$(VAULT_ADDR) -field=$(SECRET_FIELD_PREFIX)enterprise  secret/devops-ci/cloud-on-k8s/test-licenses > test-license.json
 
 # read connection info and credentials to the E2E tests monitoring Elasticsearch cluster, to be used during E2E tests
 get-monitoring-secrets:
--- a/.ci/Makefile
+++ b/.ci/Makefile
@@ -9,6 +9,8 @@ GO_MOUNT_PATH ?= /go/src/github.com/elastic/cloud-on-k8s
 ALL_IN_ONE_OUTPUT_FILE ?= config/all-in-one.yaml
 S3_ECK_DIR             ?= s3://download.elasticsearch.org/downloads/eck
 
+-include $(ROOT_DIR)/.env
+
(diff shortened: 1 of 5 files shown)


.ci/pipelines/e2e-tests-stack-versions-gke.Jenkinsfile
```diff
@@ -155,7 +167,7 @@ pipeline {
 def runWith(lib, failedTests, clusterName, stackVersion) {
     sh ".ci/setenvconfig e2e/stack-versions $clusterName $stackVersion"
     script {
-        env.SHELL_EXIT_CODE = sh(returnStatus: true, script: 'make -C .ci get-test-license get-elastic-public-key TARGET=ci-e2e ci')
+        env.SHELL_EXIT_CODE = sh(returnStatus: true, script: "make -C .ci STACK_VERSION=$stackVersion get-test-license get-elastic-public-key TARGET=ci-e2e ci")
```
thbkrkr: `setenvconfig` already sets the `STACK_VERSION` in the `.env` but the `.ci/Makefile` don't include the `.env`.
It's a bit annoying and prone to be forgotten having to propagate the stack version here.

.ci/Makefile
```diff
@@ -23,6 +23,9 @@ NOT_USED := $(shell test -e $(ROOT_DIR)/deployer-config.yml && sed "s;roleId:.*;
 endif
 endif
 
+# are we running any e2e tests in this build against SNAPSHOT versions of the stack?
+SNAPSHOT_TEST_TARGET ?= $(findstring -SNAPSHOT,$(STACK_VERSION))
```
thbkrkr: It works but I would try not to add another env var and keep the Makefile simple.

pebrc:
@thbkrkr thanks for your review. I agree that adding more complexity to the Makefiles is problematic. On the other hand introducing a hard dependency on vault into our e2e runner is also problematic IMO. I run local e2e tests quite often and like that I can inject and override all external dependencies via the `.env` file without the need to have vault around.

thbkrkr:
It is better like that.

I wonder if we should go further by not modifying `setenvconfig` and removing the `SNAPSHOT_TEST_TARGET` and check the presence of the SNAPSHOT directly in `.ci/Makefile`
since by including the `.env`, it has now access to the `STACK_VERSION`.

It may seem like a nitpick but after #2502, I feel that each new env var has a real cost in our tech debt.

```diff
diff --git a/.ci/Makefile b/.ci/Makefile
index 32fb9b9d..68804d0b 100644
--- a/.ci/Makefile
+++ b/.ci/Makefile
@@ -11,6 +11,11 @@ S3_ECK_DIR             ?= s3://download.elasticsearch.org/downloads/eck
 
 -include $(ROOT_DIR)/.env
 
+SECRET_FIELD_PREFIX := ""
+ifneq (,$(findstring -SNAPSHOT,$(STACK_VERSION)))
+       SECRET_FIELD_PREFIX = "dev-"
+endif
+
 # This is set to avoid the issue described in https://github.com/hashicorp/vault/issues/6710
 VAULT_CLIENT_TIMEOUT = 120
 
@@ -59,16 +64,14 @@ ci-build-image:
 ##  Build
 
 # read Elastic public key from Vault into license.key, to build the operator for E2E tests or for a release
-get-elastic-public-key: FIELD = $(if $(SNAPSHOT_TEST_TARGET),dev-pubkey,pubkey)
 get-elastic-public-key:
-       @ VAULT_TOKEN=$(VAULT_TOKEN) vault read -address=$(VAULT_ADDR) -field=$(FIELD) secret/devops-ci/cloud-on-k8s/license | base64 --decode > license.key
+       @ VAULT_TOKEN=$(VAULT_TOKEN) vault read -address=$(VAULT_ADDR) -field=$(SECRET_FIELD_PREFIX)pubkey secret/devops-ci/cloud-on-k8s/license | base64 --decode > license.key
 
 ##  Test
 
 # read some test licenses from Vault for E2E license tests
-get-test-license: FIELD = $(if $(SNAPSHOT_TEST_TARGET),dev-enterprise,enterprise)
 get-test-license:
-       @ VAULT_TOKEN=$(VAULT_TOKEN) vault read -address=$(VAULT_ADDR) -field=$(FIELD)  secret/devops-ci/cloud-on-k8s/test-licenses > test-license.json
+       @ VAULT_TOKEN=$(VAULT_TOKEN) vault read -address=$(VAULT_ADDR) -field=$(SECRET_FIELD_PREFIX)enterprise  secret/devops-ci/cloud-on-k8s/test-licenses > test-license.json
```

Issue #1096, opened 2019-06-17
Running e2e tests on different versions of vanilla k8s


artemnikitin:
We will use https://github.com/kubernetes-sigs/kind to test ECK against different versions of vanilla k8s.

If we will go with `kind`:
- [x] Setup new build agent on CI, approx `n1-standard-16` with 16 CPU and 60 GB RAM should be enough (details for setup: https://github.com/elastic/infra/blob/master/terraform/providers/gcp/env/elastic-apps/helm/values/gobld/devops-ci.yml)
- [x] Build new image or simply download `kind` binary to existed one?
- [x] Create new CI job (both JJBB yml and Jenkins pipeline)

If we will go with Ansible:
- Evaluate possibility of using Ansible playbook for bootstrapping K8s cluster
- Build Docker image to run Ansible
- How to handle SSH keys?
- Create new CI job (both JJBB yml and Jenkins pipeline)

barkbay:
I'm doing some tests with `Kind` and it does not support a `LoadBalancer` service with a public IP (at least for the moment: https://github.com/kubernetes-sigs/kind/issues/702)
[TestUpdateHTTPCertSAN](https://github.com/elastic/cloud-on-k8s/blob/3c021e430f6568f14bcee86b481f8af9d6043f64/operators/test/e2e/es/certs_test.go#L45) is failing on `Kind` because of this.

I think that `Services` with a public IP are something that we can mostly find in the cloud, I'm not sure that we will be able to get it in all k8s deployments.

barkbay:
Did some tests yesterday with Kind, just want to quickly share the caveats I had:

* I had some different behaviors between MacOS and Linux depending on the version of the node image.
For example the node image `v1.13.4` works fine on my MacBook but is failing on Linux. On the opposite `v1.15.0` works fine on Linux but is failing on my Mac 😞  Would be great if someone with a Mac can try to start Kind with a 1.15 node image.

* I/O problems, if you want to start 3 nodes on one server _(or your laptop)_ it means that there are 3 hosts on the same physical host and every time Kind is started:
   * All the needed Docker images (Elasticsearch, Kibana...) must be either downloaded (network I/O) or imported (disk I/O) x3 times. Maybe we can improve that by building our own "Kind node" images.
   * At some point my local Docker daemon was "stuck" for several seconds (I have a dedicated disk for Docker but it was overloaded)

[...]

PR #2623, opened 2020-02-26
Enable JSON output for E2E tests with kind

thbkrkr:
This commit enables the JSON output for E2E tests with kind to avoid to fail the `e2e-generate-xml` target.

Relates to https://github.com/elastic/cloud-on-k8s/issues/1096#issuecomment-591486860.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -484,6 +484,7 @@ else
 	$(MAKE) go-generate docker-build
 endif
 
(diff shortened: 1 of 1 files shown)


thbkrkr:
Enable the JSON output for E2E tests with kind to avoid to fail the e2e-generate-xml target.

Issue #2351, opened 2020-01-06
Review documentation topic hierarchy


charith-elastic:
<details>
<summary>The set of help topics we have in the documentation has become a bit fragmented.</summary>

- Overview
- Quickstart
- Upgrading the Elastic stack version
- Deploying ECK on OpenShift
- Accessing Elastic Stack services
- Managing compute resources
- Running Elasticsearch on ECK
	- Pod Template
	- JVM heap size
	- Node configuration
	- Volume claim templates
	- HTTP settings & TLS SANs
	- Virtual memory
	- Custom HTTP certificate
	- Settings managed by ECK
	- Secure settings
	- Custom configuration files and plugins
	- Init containers for plugin downloads
	- Update strategy
	- Pod disruption budget
	- Nodes orchestration
	- Advanced Elasticsearch node scheduling
	- Create automated snapshots
	- Readiness probe
- Running Kibana on ECK
- Running APM Server on ECK
- Creating custom images
- Configuring ECK
	- Validating webhook
- Managing licenses in ECK
- Troubleshooting
- Upgrading ECK
- Uninstalling ECK
- Glossary
</details>

We can probably group some of the related topics together for easier comprehension and sort the Elasticsearch sections by name to make them easier to find.

For example, one possible hierarchy might be:

- Overview
- Quickstart
- Deploying ECK on OpenShift
- Operating ECK    
    - Configuring ECK
    - Validating webhook
    - Creating custom images
    - Managing licenses    
    - Upgrading ECK
    - Uninstalling ECK
    - Troubleshooting
- Orchestrating Elastic stack applications
    - Accessing Elastic Stack services
    - Managing compute resources
    - Upgrading the Elastic stack version
    - Running Elasticsearch on ECK
        - Advanced Elasticsearch node scheduling
        - Create automated snapshots
        - Custom configuration files and plugins
        - Custom HTTP certificate
        - HTTP settings & TLS SANs
        - Init containers for plugin downloads
        - JVM heap size
        - Node configuration
        - Nodes orchestration
        - Pod disruption budget
        - Pod Template
        - Readiness probe
        - Secure settings
        - Settings managed by ECK
        - Update strategy
        - Virtual memory
        - Volume claim templates
    - Running Kibana on ECK
    - Running APM Server on ECK
- Glossary

alaudazzi:
[...]

PR #2634, opened 2020-02-28
Refactor top-level documentation topic hierarchy

charith-elastic:
Part of #2351 

This PR changes the structure of documentation to get rid of the clutter and simplify navigation by grouping similar topics together. 

- Each source document is now located under a directory structure that matches the rendered topic tree.
- Each document is self-contained (no need to figure out the correct heading level or use `float`s).
- Document prelude (document ID and GitHub banner) is standardised and easier to process with scripts.
- Gerunds in secondary topics have been removed.

Preview: http://cloud-on-k8s_2634.docs-preview.app.elstc.co/guide/en/cloud-on-k8s/master/index.html


pebrc:
I like the new structure. One thing I noticed though is that we don't have a navigation on the lower level e.g. for the `Run Elasticsearch` section which is really long it is just one menu item on the rhs. I think we should have a sub-menu here as it becomes really difficult to navigate otherwise

Screenshot 2020-02-28 at 15 27 08

Something structurally similar to this:
Screenshot 2020-02-28 at 15 29 37

charith-elastic:
Our docs were structured as [an _article_ instead of a  _book_](https://asciidoctor.org/docs/user-manual/#document-types) like the rest of Elastic products -- hence why the navigation bar displayed only two levels. It has been fixed now.

http://cloud-on-k8s_2634.docs-preview.app.elstc.co/guide/en/cloud-on-k8s/master/index.html

alaudazzi:
@charith-elastic The new structure looks much better. Are we going to handle "Running Elasticsearch on ECK" section in a separate PR?

pebrc:
This looks much better! 

One thing I noticed though is that in the current version of the docs we sometimes forced shorter sections (not sure if this is the correct terminology) on to one page by using `float`, especially if sections where very short and if breaking them into separate page would feel disruptive to the flow when reading. The new `book` format does away with those joined sections, is this deliberate, an oversight or just not possible when using `book`

Example: http://cloud-on-k8s_2634.docs-preview.app.elstc.co/guide/en/cloud-on-k8s/master/k8s-licensing.html

charith-elastic:
> Are we going to handle "Running Elasticsearch on ECK" section in a separate PR?

@alaudazzi Yes, I think it would be better to handle that in a separate PR to avoid making this one too large. I also need your help with the "Running Elasticsearch" section as I am struggling to find a good way to organise it.

@pebrc We can certainly add `float`s to avoid breaking some sections into their own page. I am just struggling to find a consistent system to apply to all of our docs -- which is why I didn't attempt to do anything about it yet. I will discuss with @alaudazzi  and try to fix those short sections in a different PR.

pebrc:
@charith-elastic I am +1 on merging as is and fixing later

PR #2636, opened 2020-02-28
User-provided config take precedence over operator config

### LINKED ISSUE #2573: User-provided Elasticsearch config overrides should have precedence over ECK config
Currently, users get a warning if they specify an Elasticsearch config field already handled by ECK.
However, our philosophy is to still allow power users to achieve what they want.

When reconciling the user-provided configuration with the ECK-generated configuration, we do not currently give priority to the user-provided configuration:

https://github.com/elastic/cloud-on-k8s/blob/f63396e905a2fbaa5f5746fc9655044bddf400f5/pkg/controller/elasticsearch/settings/merged_config.go#L29-L36

This impacts the way users can set multiple volumes for `path.data`: https://discuss.elastic.co/t/how-to-define-multiple-path-data-pointing-to-separate-volumes-or-pvc-using-eck/218644

thbkrkr:
This commit implements the same behaviour that we have for Kibana and APM Server where
the user-provided config overrides have precedence over ECK config.

Resolves #2573.

### DIFF
--- a/pkg/controller/elasticsearch/settings/merged_config_test.go
+++ b/pkg/controller/elasticsearch/settings/merged_config_test.go
@@ -147,7 +147,7 @@ func TestNewMergedESConfig(t *testing.T) {
 			name:    "user-provided Elasticsearch config overrides should have precedence over ECK config",
 			version: "7.6.0",
 			cfgData: map[string]interface{}{
-				esv1.NetworkHost:        "1.2.3.4",
+				esv1.NetworkHost: "1.2.3.4",
 			},
 			assert: func(cfg CanonicalConfig) {
 				cfgBytes, err := cfg.Render()
--- a/pkg/controller/elasticsearch/settings/merged_config_test.go
+++ b/pkg/controller/elasticsearch/settings/merged_config_test.go
@@ -5,6 +5,8 @@
 package settings
 
 import (
+	"bytes"
+	"fmt"
 	"testing"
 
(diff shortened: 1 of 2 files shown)


pkg/controller/elasticsearch/settings/merged_config_test.go
```diff
@@ -141,6 +143,20 @@ func TestNewMergedESConfig(t *testing.T) {
...
+		{
+			name:    "user-provided Elasticsearch config overrides should have precedence over ECK config",
+			version: "7.6.0",
+			cfgData: map[string]interface{}{
+				esv1.NetworkHost: "1.2.3.4",
+			},
+			assert: func(cfg CanonicalConfig) {
+				cfgBytes, err := cfg.Render()
+				require.NoError(t, err)
+				fmt.Println(string(cfgBytes))
```
sebgl: leftover println?

pkg/controller/elasticsearch/settings/merged_config_test.go
```diff
@@ -141,6 +143,20 @@ func TestNewMergedESConfig(t *testing.T) {
...
+				require.True(t, bytes.Contains(cfgBytes, []byte("publish_host: ${POD_IP}")))
+				require.True(t, bytes.Contains(cfgBytes, []byte("host: 1.2.3.4")))
```
sebgl: should we also check the value is correctly overridden and not duplicated?

Issue #2605, opened 2020-02-24
Don't reset voting config exclusions at every single reconciliation


sebgl:
Closely related to https://github.com/elastic/cloud-on-k8s/issues/1522. We should apply the same fix for both.

We currently call `DELETE /_cluster/voting_config_exclusions` at every reconciliation. It is harmless but sounds a bit useless and impacts performance.
We only need to do it if some nodes are excluded from voting.

PR #2642, opened 2020-03-02
Don't call the voting config exclusions API at every single reconciliation

sebgl:
This PR sets up an annotation containing the last applied value for
voting config exclusions. The annotation can either be:
- nil: it has not been set yet
- empty: the last setting applied was to reset voting config exclusions
- not empty: it contains the last applied voting config exclusions

Before making the ES API call to either add nodes to voting config exclusions
or reset its value, we check the value of the annotation.

Users can remove the annotation at anytime, it will be set again after
the next API call.
Note we ignore cases where users would manipulate voting config
exclusions behind our back here.

Fixes https://github.com/elastic/cloud-on-k8s/issues/2605.

### DIFF
--- a/pkg/controller/elasticsearch/version/zen2/voting_exclusions.go
+++ b/pkg/controller/elasticsearch/version/zen2/voting_exclusions.go
@@ -6,6 +6,7 @@ package zen2
 
 import (
 	"context"
+	"strings"
 
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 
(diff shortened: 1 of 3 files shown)


pkg/controller/elasticsearch/version/zen2/voting_exclusions.go
```diff
@@ -19,6 +20,35 @@ var (
 	log = logf.Log.WithName("zen2")
 )
 
+const (
+	// VotingConfigExclusionsAnnotationName is an annotation that stores the last applied voting config exclusions.
+	// An empty value means no voting config exclusions is set.
```
pebrc:
```suggestion
	// An empty value means no voting config exclusions are set.
```

pkg/controller/elasticsearch/version/zen2/voting_exclusions.go
```diff
@@ -19,6 +20,35 @@ var (
...
+	VotingConfigExclusionsAnnotationName = "elasticsearch.k8s.elastic.co/voting-config-exclusions"
+)
+
+func serializeExcludedNodesForAnnotation(excludedNodes []string) string {
+	return strings.Join(excludedNodes, ",")
```
pebrc: Should this be sorted to ensure a stable comparison? I know we have an implicitly stable ordering due to the way we calculate downscales. So there would be no difference in practice as long as the downscale calculation stays the same.

sebgl: good point, it feels safer to sort here even though currently useless

pkg/controller/elasticsearch/version/zen2/voting_exclusions.go
```diff
@@ -19,6 +21,40 @@ var (
...
+)
+
+// serializeExcludedNodesForAnnotation returns a sorted comma-separated representation of the given slice.
+func serializeExcludedNodesForAnnotation(excludedNodes []string) string {
+	// sort a copy to not mutate the given slice
+	sliceCopy := make([]string, len(excludedNodes))
+	copy(sliceCopy, excludedNodes)
+	sort.Strings(sliceCopy)
+	return strings.Join(sliceCopy, ",")
+}
```
sebgl: I decided to sort a copy of the slice instead of mutating the slice itself. That's an extra allocation but I would not expect such a side effect to occur in that function.

pebrc:
reverted in #2883 (thus `>non-issue`)

PR #2648, opened 2020-03-02
Add EKS support to deployer

pebrc:
Adds the ability to create and delete clusters in EKS via `eksctl` to deployer.
* refactored the GKE NVMe setup script to be reusable in EKS (with the addition of one more init container)
* add an EKS Jenkins job/pipeline

Did a test run on a local Jenkins instance: 

Screenshot 2020-04-05 at 19 49 27


pebrc:
Jenkins test this please

hack/deployer/runner/settings.go
```diff
@@ -26,15 +26,16 @@ type Plan struct {
 	Gke               *GkeSettings `yaml:"gke,omitempty"`
 	Aks               *AksSettings `yaml:"aks,omitempty"`
 	Ocp               *OcpSettings `yaml:"ocp,omitempty"`
+	EKS               *EKSSettings `yaml:"eks,omitempty"`
```
pebrc: Abbreviations should be all caps afaik. I think we should refactor that, but maybe in a separate PR?

hack/deployer/runner/eks.go
```diff
@@ -0,0 +1,186 @@
...
+	return &EKSDriver{
+		plan: plan,
+		ctx: map[string]interface{}{
+			"ClusterName":       plan.ClusterName,
+			"Region":            plan.EKS.Region,
+			"KubernetesVersion": plan.KubernetesVersion,
+			"NodeCount":         plan.EKS.NodeCount,
+			"MachineType":       plan.MachineType,
+			"NodeAMI":           plan.EKS.NodeAMI,
+			"WorkDir":           plan.EKS.WorkDir,
```
pebrc: I am following the pattern established in other provider implementations here, but I was wondering whether it would be better to simply pass the plan struct to the template. It would requite encoding aspects of the struct hierarchy into the templates but it would save us from enumerating all used keys here.

hack/deployer/runner/settings.go
```diff
@@ -72,9 +73,17 @@ type OcpSettings struct {
 	NodeCount                  int    `yaml:"nodeCount"`
 }
 
+// EKSSettings are specific to Amazon EKS.
+type EKSSettings struct {
+	NodeAMI   string `yaml:"nodeAMI"`
+	NodeCount int    `yaml:"nodeCount"`
+	Region    string `yaml:"region"`
+	WorkDir   string `yaml:"string"`
```
thbkrkr, line 81:
```suggestion
	WorkDir   string `yaml:"workDir"`
```

hack/deployer/runner/eks.go
```diff
@@ -0,0 +1,189 @@
...
+	})
+	return nil
+}
+
+func (e *EKSDriver) GetCredentials() error {
+	if err := e.fetchSecrets(); err != nil {
+		return fmt.Errorf("while fetching secrets %w", err)
+	}
+	log.Printf("writing kubeconfig")
+	return e.newCmd("eksctl utils write-kubeconfig --name {{.ClusterName}} --region {{.Region}}").Run()
```
thbkrkr, line 155: Nit: eksctl indicates that `Flag --name has been deprecated, use --cluster`.

.ci/Dockerfile
```diff
@@ -9,6 +9,7 @@ ENV DOCKER_VERSION=18.03.1-ce
 ENV GOTESTSUM_VERSION=0.3.5
 ENV KIND_VERSION=0.7.0
 ENV OPENSHIFT_TOOLS_VERSION=4.3.0
+ENV EKSCTL_VERSION=0.14.0
```
thbkrkr, line 12: Since you opened this PR, version 0.16.0 is available :)

thbkrkr:
Looks good!

I was surprised that the kube config depends on the aws-cli:

```yaml
...
kind: Config
preferences: {}
users:
- name: eks_devops_ci@dev-thb.eu-west-2.eksctl.io
  user:
    exec:
      apiVersion: client.authentication.k8s.io/v1alpha1
      args:
      - eks
      - get-token
      - --cluster-name
      - dev-thb
      - --region
      - eu-west-2
      command: aws
      env: null
```

I met an issue with the version of the aws cli. I had version 1.16 that does not support `get-token` in the command `aws eks get-token --cluster-name $n --region $r`.

I upgraded to the 1.18 and now I have:

```sh
> k get pods
error: You must be logged in to the server (Unauthorized)
```

Worse, the build-ci image gets the 1.11 version that does not even have the `eks` command.

Maybe we should verify that the aws-cli is installed and that its version is sufficient?

hack/deployer/runner/settings.go
thbkrkr: Yes, we should. I have nothing against doing it in this PR :)

pebrc:
> Worse, the build-ci image gets the 1.11 version that does not even have the eks command.

I updated that in a commit I had not pushed.

pebrc:
I realised that I had removed the draft status a bit to early from this PR. There is more follow up work to be done: for a successful e2e test run we also need a e2e-default storage class. I might split this up into a separate PR though

pebrc:
> Maybe we should verify that the aws-cli is installed and that its version is sufficient?

Yes and typically you also need to specify a profile/assume-role so there is some docs missing here as well I believe

hack/deployer/runner/settings.go
pebrc: Let's keep things separate https://github.com/elastic/cloud-on-k8s/issues/2815

Makefile
```diff
@@ -370,7 +373,13 @@ ifeq ($(REGISTRY), eu.gcr.io)
 	# this is used by the cloud-on-k8s-e2e-tests-ocp job
 	@ gcloud auth configure-docker --quiet
 endif
+ifeq ($(REGISTRY), docker.elastic.co)
+	@ docker login -u $(ELASTIC_DOCKER_LOGIN) -p $(ELASTIC_DOCKER_PASSWORD) push.docker.elastic.co
+	@ docker tag $(E2E_IMG) push.$(E2E_IMG)
+	@ docker push push.$(E2E_IMG)
+else
```
pebrc, line 380: We have an issue to "Consider using single, public container registry for ci and dev workflows" which will address this for all build jobs.

hack/deployer/runner/eks.go
david-kow: I think the current approach has the advantage of adding keyvalues that are not in the plan and providing defaults. We do that for OCP and GKE.

hack/deployer/config/plans.yml
```diff
@@ -76,6 +80,30 @@ plans:
...
+  clusterName: ci
+  provider: eks
+  machineType: c5d.2xlarge
+  serviceAccount: false
+  kubernetesVersion: 1.14
+  eks:
+    region: eu-central-1
+    nodeCount: 3
+    nodeAMI: static
+    diskSetup: kubectl apply -f hack/deployer/config/local-disks/ssd-provisioner.yaml
```
thbkrkr, line 94: This makes it mandatory to run the deployer from the root path.
It doesn't really bother me but I got caught. I'm used to using it from `hack/deployer` to avoid specifying plans and configuration files via the flags.

FWIW, my workflow is:
```sh
cd hack/deployer
go build -o deployer main.go
./deployer execute
```

PR #2655, opened 2020-03-03
add more details for custom apm-server configuration

IzekChen:
add more details for custom apm-server configuration.
It will be more easily for users to understand


charith-elastic:
Thanks for the PR. Jaeger support in APM Server is still experimental so we probably don't want to include it in ECK documentation just yet. However, we would be happy to accept a PR to include this as a [recipe](https://github.com/elastic/cloud-on-k8s/tree/master/config/recipes).

PR #2656, opened 2020-03-03
Include samples in the documentation

charith-elastic:
Includes the samples in the documentation to make them easier to discover.

Part of #2351


charith-elastic:
I assumed that the whole repository will be checked out by the docs build but unless explicitly added to the config, only the `docs` sub-directory is available at build time.

PR #2659, opened 2020-03-03
Update transport cert verification to full

anyasabo:
We can use [full certificate verification](https://www.elastic.co/guide/en/elasticsearch/reference/current/security-settings.html#ssl-tls-settings) for the transport layer since users cannot modify the transport certs, and we [issue certificates for each pod](https://github.com/elastic/cloud-on-k8s/blob/master/pkg/controller/elasticsearch/certificates/transport/pod_secret.go#L34). We do not have any open issues to allow users to modify the transport certs, and if we decide to change course later we can roll back this change pretty easily.

Also clarifies why we use our own cert validation for the ES HTTP client, since we do not need it when we manage certificates and it was confusing for me until dear @nkvoll pointed it out.

### DIFF
--- a/pkg/controller/elasticsearch/client/client.go
+++ b/pkg/controller/elasticsearch/client/client.go
@@ -124,10 +124,13 @@ func NewElasticsearchClient(
 		TLSClientConfig: &tls.Config{
 			RootCAs: certPool,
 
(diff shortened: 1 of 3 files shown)


pebrc:
Reverted in #2831  (thus `>non-issue`)

PR #2663, opened 2020-03-04
Adding apm config recipes

IzekChen:
Adding example for enabling jaeger function. Also for people to have a better understanding of the configuration structure

### DIFF
--- a/config/recipes/beats/4_apm-server.yaml
+++ b/config/recipes/beats/4_apm-server.yaml
@@ -1,0 +1,32 @@
+apiVersion: apm.k8s.elastic.co/v1
+kind: ApmServer
+metadata:
+  name: apm-server-quickstart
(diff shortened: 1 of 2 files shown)


charith-elastic:
Hi, thanks again for the PR. There are a few things to address here though:

- Since Jaeger support is still experimental (and optional) including it in the main documentation will be confusing to some users. This is why I suggested that it should be a recipe instead. 
- The recipe should probably go into its own directory such as `config/recipes/apm-server-jaeger` and not under `beats` as it is not a required element of the Beats recipe.
- It would be good to have the PR against the master branch. We can then decide to backport it to 1.0 if necessary.

IzekChen:
Hi @charith-elastic,
Thank you for the information.
I will commit it again.
:)

PR #2669, opened 2020-03-05
Add a JJBB job for snapshot e2e tests

### LINKED ISSUE #2565: E2E Tests: test snapshot releases of the stack

pebrc:
* Adds a new JJBB job and Jenkins pipeline to test dev builds of the
Elastic stack against ECK. 
* Builds a custom dev version of ECK off of
master to do that.
* Adds a new environment to `.setenvconfig` (sadly) and tweaks the existing one for the stack versions tests to be able to deal with snapshot builds as well.

Fixes #2565


.ci/setenvconfig
```diff
@@ -310,6 +310,23 @@ ENV
...
 
+######################################
+  dev/build)
+######################################
+
+write_env <<ENV
+REGISTRY = docker.elastic.co
+REPOSITORY = eck-snapshots
+IMG_SUFFIX = -dev-ci
+SNAPSHOT_TEST_TARGET = yes
```
barkbay: Any reason to override what has been "guessed" at line 204 ?

.ci/setenvconfig
```diff
@@ -201,17 +201,17 @@ CFG
 
 clusterName=$2
 stackVersion=$3
+snapshotVersion=$(case $stackVersion in (*-SNAPSHOT)  echo Yes;; esac)
```
barkbay: This is a kind of boolean _(even this could be misleading since if we only check if it is set)_, maybe rename it to `isSnapshotVersion=...` ?

.ci/pipelines/e2e-tests-snapshot-versions-gke.Jenkinsfile
```diff
@@ -0,0 +1,131 @@
...
+            environment {
+               // use the image we just built
+               OPERATOR_IMAGE = "${sh(returnStdout: true, script: 'make print-operator-image').trim()}"
+            }
+            parallel {
+                stage("7.6.1-SNAPSHOT") {
+                    steps {
+                        checkout scm
+                        script {
+                            runWith(lib, failedTests, "eck-76-${BUILD_NUMBER}-e2e", "7.6.1-SNAPSHOT")
```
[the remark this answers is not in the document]
pebrc: Very good point! Will do.

.ci/setenvconfig
pebrc: Yes. L204 is for a different environment. Basically there are two environments where it is relevant whether we are building or testing against snapshot builds:
1. when building the operator image. Here we know we want to build a dev version that can deal with snapshot stack versions so we can hard-code `SNAPSHOT_TEST_TARGET` to pseudo-boolean `yes`. This is what this line is about
2. when running the e2e tests, here I wanted to use the same environment config for both variants so we do some auto-detection based on the stack version. That's line 204

pebrc: But I realise now that `SNAPSHOT_TEST_TARGET` is a bad name for case 1.

Issue #2674, opened 2020-03-06
Association resource leak


barkbay:
I think we have an association resource leak. It happens when Elasticsearch and the associated resource are not in the same namespace.
In this case, when the Elasticsearch reference is changed or deleted, then the user Secret in the Elasticsearch namespace is not deleted ([see this unit test](https://github.com/elastic/cloud-on-k8s/commit/c37b09ae6db21ebf7ac5158affd0745bfb4fef8e)).

It could have been introduce in https://github.com/elastic/cloud-on-k8s/commit/4ee6a23dc17fa4a6c055a06b941f51108f5755b3 but I think we also have the same issue with the `APMServer`.

barkbay:
Closed via #2679.

PR #2679, opened 2020-03-09
Association controller bug fixes

### LINKED ISSUE #2675: Annotation 'association.k8s.elastic.co/es-conf' should be removed for unmanaged ES clusters
In Kibana _(I think it's also the case with the APMServer)_, when a user removes the reference of a managed Elasticsearch cluster and set `config.elasticsearch.*` manually then the `association.k8s.elastic.co/es-conf` annotation is not removed.

This is a problem since the resources referenced in this annotation _(CA and user Secrets)_, which have been generated by the association controller, are considered "orphaned" and should be deleted. 

Related to #2674

barkbay:
This PR fixes #2675 #2674 #2694 #2695 #2692 

About the unit test: I only added a unit test for #2674. My feeling is that either we are missing some tooling to properly create unit tests against controllers or adding a unit test to cover `reconcileInternal` would be out the scope of this PR. I can, however, create an issue to track this.


pkg/controller/common/association/resources.go
```diff
@@ -0,0 +1,73 @@
...
+	"github.com/elastic/cloud-on-k8s/pkg/utils/k8s"
+	"go.elastic.co/apm"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// CreatedBy returns true if an object has been created by an other one
+type CreatedBy func(created, creator metav1.Object) bool
```
sebgl: It sounds like the APMServer would created a Secret for example, which is not exactly the case.
Maybe:

```go
// ServesAssociation returns true if the given resource has been created to serve the association of associatedResource
type ServesAssociation func(resource metav1.Object, associatedResource metav1.Object) bool
```

pkg/controller/common/association/resources.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+	return nil
+}
+
+func deleteIfOrphaned(
+	c k8s.Client,
+	secret *corev1.Secret,
+	associated commonv1.Associated,
+	createdBy CreatedBy,
+	esRef commonv1.ObjectSelector,
```
sebgl: can we remove `esRef` since it can be extracted from `associated`?

pkg/controller/common/association/resources.go
```diff
@@ -0,0 +1,73 @@
...
+) error {
+	if metav1.IsControlledBy(secret, associated) || createdBy(secret, associated) {
```
sebgl: Isn't `createdBy()` sufficient? Why/when do we need `metav1.IsControlledBy `?

pkg/controller/common/association/resources.go
```diff
@@ -0,0 +1,73 @@
...
+		if !esRef.IsDefined() {
+			// look for association secrets owned by this associated instance
```
sebgl: Reading `look for association` I would somehow expect the next piece of code to be looking for stuff, whereas it's actually deleting the given secret.

pkg/controller/common/association/resources.go
```diff
@@ -0,0 +1,73 @@
...
+			// which should not exist since no ES referenced in the spec
+			return deleteExistingSecret(c, secret, associated)
+		} else if value, ok := secret.Labels[common.TypeLabelName]; ok && value == user.UserType && esRef.Namespace != secret.Namespace {
+			// User secret may live in an other namespace, check if it has changed
```
sebgl: same here: this is not checking anything but just deleting a secret?

sebgl: I'm not sure I understand why `user` is a special case here, compared to any secret serving an association purpose.

pkg/controller/common/association/resources.go
```diff
@@ -0,0 +1,73 @@
...
+			return deleteExistingSecret(c, secret, associated)
+		}
+	}
+	return nil
+}
+
+func deleteExistingSecret(c k8s.Client, secret *corev1.Secret, associated commonv1.Associated) error {
```
sebgl: `deleteSecret` is probably good enough, `existing` feels redundant and is actually ignored with `!apierrors.IsNotFound(err)`?

pkg/controller/apmserverelasticsearchassociation/apmserverelasticsearchassociation_controller.go
```diff
@@ -406,24 +409,5 @@ func (r *ReconcileApmServerElasticsearchAssociation) reconcileElasticsearchCA(ct
 // now redundant old user object/secret. This function lists all resources that don't match the current name/namespace
 // combinations and deletes them.
 func deleteOrphanedResources(ctx context.Context, c k8s.Client, as *apmv1.ApmServer) error {
```
sebgl: Looks like we could remove this function and call `association.DeleteOrphanedResources(ctx, c, as, NewResourceLabels(as.Name), createdByApmServer) ` directly?

pkg/controller/kibanaassociation/association_controller_test.go
```diff
@@ -290,6 +291,80 @@ func Test_deleteOrphanedResources(t *testing.T) {
...
+							AssociationLabelName: "kibana-foo",
+						},
+						OwnerReferences: []metav1.OwnerReference{
+							ownerRefFixture,
+						},
+					},
+				},
+			},
+			postCondition: func(c k8s.Client) {
+				// This works even without labels because mock client currently ignores labels
```
sebgl: I thought the fake client considers labels in `.List()` now?

pkg/controller/kibanaassociation/association_controller.go
```diff
@@ -403,43 +401,10 @@ func (r *ReconcileAssociation) reconcileElasticsearchCA(ctx context.Context, kib
...
-			}
-		}
-	}
-	return nil
+func deleteOrphanedResources(
+	ctx context.Context,
+	c k8s.Client,
+	kibana *kbv1.Kibana,
+) error {
+	return association.DeleteOrphanedResources(ctx, c, kibana, NewResourceSelector(kibana.Name), createdByKibana)
```
sebgl: IIUC `NewResourceSelector ` selects based on `AssociationLabelName` whereas `createdByKibana ` ultimately selects based on `AssociationLabelName` + `AssociationLabelNamespace`. It feels a bit redundant and not exactly clear to me what is the different role of each.

pkg/controller/kibanaassociation/labels.go
```diff
@@ -21,24 +20,24 @@ const (
...
+	labels := created.GetLabels()
+	if name, ok := labels[AssociationLabelName]; !ok || name != creator.GetName() {
 		return false
 	}
-	if ns, ok := labels[AssociationLabelNamespace]; !ok || ns != kibana.Namespace {
+	if ns, ok := labels[AssociationLabelNamespace]; !ok || ns != creator.GetNamespace() {
 		return false
 	}
 	return true
 }
```
sebgl: Since it's basically the same function for Kibana, APMServer and EnterpriseSearch, could we make a single generic one?

```go
// -- in common/association

// ServesAssociationPurpose returns true if the given resource has been created to serve an association
// based on the given labels.
func ServesAssociationPurpose(resource metav1.Object, associationLabels map[string]string) bool {
    return maps.IsSubset(associationLabels, resource.GetLabels())
}

// -- in kibanaassociation

// KibanaAssociationSelector returns labels to set on resources created for association purposes.
func KibanaAssociationSelector(kb kbv1.Kibana) map[string]string{
    return map[string]string{
        AssociationLabelName: kb.Name,
        AssociationLabelNamespace: kb.Namespace,
    }
}

association.DeleteOrphanedResources(ctx, c, kibana, NewResourceSelector(kibana.Name), KibanaAssociationSelector(kb))
```

The difference between `NewResourceSelector` and `KibanaAssociationSelector` is not very clear to me though, I must be missing something.

pkg/controller/common/association/resources.go
barkbay: IIUC objects that lives in the `Kibana|APMServer` namespace don't have the `[kibana|apmserver]association.k8s.elastic.co/namespace` label. But they have `Kibana|APMServer` as an owner and so are caught by `metav1.IsControlledBy`.

Since we can't set an owner that lives in a different namespace we use `createdBy` to catch the users created in the Elasticsearch namespace.

If we decide to set the same labels on all the objects I _think_ it should be possible to only use `createdBy`. It should be noted that in this case there is a (very) small chance to not garbage collect immediately some objects during an ECK upgrade of ECK. The reason is that `[kibana|apmserver]association.k8s.elastic.co/namespace` would be missing until the first reconciliation _(but they would be removed eventually when Kibana or the APM server is deleted, so definitely not a big deal imho)_

pkg/controller/apmserverelasticsearchassociation/apmserverelasticsearchassociation_controller.go
barkbay: I would be happy to do that but there are some unit tests that rely on it _(See my comment in the description)_.

pkg/controller/kibanaassociation/association_controller_test.go
barkbay: Maybe, I didn't try to refactor the existing code so much. But I'll try to have a look.

barkbay:
This PR adds a consistent way to detect `Secrets` that servers an association purpose.
They now all have the following labels:

* `[kibana|apm]association.k8s.elastic.co/name`
* `[kibana|apm]association.k8s.elastic.co/namespace`

This simplifies the code, we don't need to rely on ownership for some of them or labels for some other ones.

It's also fixes some issues found while reviewing the code.

I'll update the title and the description before merging.

pkg/controller/apmserverelasticsearchassociation/apmserverelasticsearchassociation_controller.go
```diff
@@ -239,28 +240,46 @@ func (r *ReconcileApmServerElasticsearchAssociation) isCompatible(ctx context.Co
...
+	if err := deleteOrphanedResources(ctx, r, apmServer); err != nil {
+		log.Error(err, "Error while trying to delete orphaned resources. Continuing.", "namespace", apmServer.Namespace, "as_name", apmServer.Name)
+	}
+	apmServerKey := k8s.ExtractNamespacedName(apmServer)
 	// no auto-association nothing to do
 	elasticsearchRef := apmServer.Spec.ElasticsearchRef
 	if !elasticsearchRef.IsDefined() {
-		return commonv1.AssociationUnknown, nil
+		// stop watching any ES cluster previously referenced for this Kibana resource
+		r.watches.ElasticsearchClusters.RemoveHandlerForKey(elasticsearchWatchName(apmServerKey))
```
sebgl: Should we also remove the watch attached to the user for the previously referenced ES secret? (eg. `r.watches.Secrets.RemoveHandlerForKey(elasticsearchWatchName(obj))`)
It's tricky since we don't have that reference anymore.

pkg/controller/apmserverelasticsearchassociation/apmserverelasticsearchassociation_controller.go
```diff
@@ -239,28 +240,46 @@ func (r *ReconcileApmServerElasticsearchAssociation) isCompatible(ctx context.Co
...
 	}
 	if elasticsearchRef.Namespace == "" {
 		// no namespace provided: default to the APM server namespace
 		elasticsearchRef.Namespace = apmServer.Namespace
 	}
-	assocKey := k8s.ExtractNamespacedName(apmServer)
+
 	// Make sure we see events from Elasticsearch using a dynamic watch
 	// will become more relevant once we refactor user handling to CRDs and implement
 	// syncing of user credentials across namespaces
```
sebgl: Can we use this PR as an opportunity to remove this deprecated comment block?

pkg/controller/common/association/ca.go
```diff
@@ -68,10 +69,12 @@ func ReconcileCASecret(
 		Expected:   &expectedSecret,
 		Reconciled: &reconciledSecret,
 		NeedsUpdate: func() bool {
-			return !reflect.DeepEqual(expectedSecret.Data, reconciledSecret.Data)
+			return !reflect.DeepEqual(expectedSecret.Data, reconciledSecret.Data) ||
+				!maps.IsSubset(expectedSecret.Labels, reconciledSecret.Labels)
```
sebgl: https://github.com/elastic/cloud-on-k8s/pull/2707 should help.
I'll do a second pass in a few days for in-flight PRs that did not benefit from it so we can keep this PR as is.

pkg/controller/common/association/resources.go
```diff
@@ -0,0 +1,74 @@
...
+)
+
+// DeleteOrphanedResources deletes resources created by an association that are left over from previous reconciliation
+// attempts. Common use case is an Elasticsearch reference in Kibana or APMServer spec that was removed.
+func DeleteOrphanedResources(
+	ctx context.Context,
+	c k8s.Client,
+	associated commonv1.Associated,
+	matchLabels client.MatchingLabels,
+
```
sebgl: nit: empty line

pkg/controller/common/association/resources.go
```diff
@@ -0,0 +1,74 @@
...
+// deleteIfOrphaned deletes a Secret if it is not needed anymore, either because the association has been removed
+// or the target namespace has changed.
+func deleteIfOrphaned(
+	c k8s.Client,
+	secret *corev1.Secret,
+	associated commonv1.Associated,
+) error {
+	esRef := associated.ElasticsearchRef().WithDefaultNamespace(associated.GetNamespace())
+
+	// Secret should not exist if there is ES referenced in the spec or if the resource is deleted
```
sebgl: `if there is ES referenced` -> `if there is no ES referenced`

pkg/controller/common/association/resources.go
```diff
@@ -0,0 +1,74 @@
+package association
```
sebgl: This whole file is much easier to digest since you changed the labelling logic 🙏

pkg/controller/kibana/driver.go
```diff
@@ -122,48 +118,28 @@ func (d *driver) deploymentParams(kb *kbv1.Kibana) (deployment.Params, error) {
...
 	// we need to deref the secret here (if any) to include it in the checksum otherwise Kibana will not be rolled on contents changes
 	if kb.AssociationConf().AuthIsConfigured() {
-		esAuthSecret := types.NamespacedName{Name: kb.AssociationConf().GetAuthSecretName(), Namespace: kb.Namespace}
-		if err := d.dynamicWatches.Secrets.AddHandler(watches.NamedWatch{
-			Name:    secretWatchKey(kbNamespacedName),
-			Watched: []types.NamespacedName{esAuthSecret},
-			Watcher: kbNamespacedName,
-		}); err != nil {
-			return deployment.Params{}, err
-		}
```
sebgl: Shouldn't we watch this secret to reconcile Kibana if its ES user changes?

sebgl: Ah, we don't since we already watch all secrets in the namespace!

pkg/controller/apmserverelasticsearchassociation/apmserverelasticsearchassociation_controller.go
barkbay: The watch name does not depend on the Es name.
I kept the watch on the Secrets on purpose here. It does not cost that much and it has the benefit to delete any unnecessary secret that would be created "externally" (e.g. by the end user or restored from a backup). A reconciliation would occur and they would be deleted by `deleteOrphanedResources`

All the watchers are cleaned during the `onDelete` call.

barkbay: Discussed out of band: this is an implicit mechanism that would be hard to understand later. We reuse the `onDelete` function to clean up the watches instead.

pkg/controller/kibanaassociation/association_controller_test.go
```diff
@@ -290,6 +251,69 @@ func Test_deleteOrphanedResources(t *testing.T) {
 			},
 			wantErr: false,
 		},
+		{
+			name: "No more es ref in Kibana, orphan user for previous es ref in a different namespace still exist",
```
pebrc: exists

PR #2681, opened 2020-03-09
Fix controller-gen installation

barkbay:
It seems that `ifneq ($(shell controller-gen --version), Version: v0.2.5)` fails if `controller-gen` is not already installed:
 https://devops-ci.elastic.co/job/cloud-on-k8s-pr/2883/console


barkbay:
jenkins test this please

barkbay:
Unrelated and harmless.

TIL: 
> Running error: context loading failed: no go files to analyze 

Actually means that the code des not compile.

Issue #728, opened 2019-05-03
Support file realm in the Elasticsearch spec


sebgl:
This would allow users to inject their own users into the file realm.
The ES spec would contain the username and a reference to the password secret.
(or maybe just a reference to the secret that contains both usernames and password hashes?)

___

This issue explores a lot of options, we finally agreed on providing a way to setup secrets containing the "raw" file realm content as [described in Elasticsearch docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/file-realm.html#file-realm-configuration):

```yaml
apiVersion: elasticsearch.k8s.elastic.co/v1
kind: Elasticsearch
metadata:
  name: elasticsearch-sample
spec:
  version: 7.6.0
  auth:
    roles:
    - secretName: my-roles-1
    fileRealm:
    - secretName: my-filerealm-secret

---

# File realm in ES format (from the CLI or manually assembled)
kind: Secret
apiVersion: v1
metadata:
  name: my-filerealm-secret
stringData:
  users: |-
    rdeniro:$2a$10$BBJ/ILiyJ1eBTYoRKxkqbuDEdYECplvxnqQ47uiowE7yGqvCEgj9W
    alpacino:$2a$10$cNwHnElYiMYZ/T3K4PvzGeJ1KbpXZp2PfoQD.gfaVdImnHOwIuBKS
    jacknich:{PBKDF2}50000$z1CLJt0MEFjkIK5iEfgvfnA6xq7lF25uasspsTKSo5Q=$XxCVLbaKDimOdyWgLCLJiyoiWpA/XDMe/xtVgn1r5Sg=
  users_roles: |-
    admin:rdeniro
    power_user:alpacino,jacknich
    user:jacknich

---

# Roles in ES format
kind: Secret
apiVersion: v1
metadata:
  name: my-roles-1
stringData:
  roles.yml: |-
    click_admins:
      run_as: [ 'clicks_watcher_1' ]
      cluster: [ 'monitor' ]
      indices:
      - names: [ 'events-*' ]
        privileges: [ 'read' ]
        field_security:
          grant: ['category', '@timestamp', 'message' ]
        query: '{"match": {"category": "click"}}'
    another_role:
      cluster: [ 'all' ]
```

Let's keep the door open for higher-level abstractions (eg. one secret per user with username and password fields) in the future.

nkvoll:
Do we have an overview over the use cases where we think the file realm will be required, and the native realm (or other realms) would not be suitable?

[...]

PR #2682, opened 2020-03-09
Support user-provided file realm and roles & refactor ES user management

sebgl:
Closes https://github.com/elastic/cloud-on-k8s/issues/728.

## Defining users and roles in Kubernetes secrets

This PR adds support for user-provided file realm and roles, through Kubernetes secrets.
File realm and roles secrets must be referenced in the Elasticsearch CRD. The secret content must match the file realm and file-based roles specification from the Elasticsearch documentation.

Example:

```yaml
kind: Elasticsearch
metadata:
  name: elasticsearch-sample
spec:
  version: 7.6.0
  auth:
    roles:
    - secretName: my-roles-1
    fileRealm:
    - secretName: my-filerealm-secret

---

# File realm in ES format (from the CLI or manually assembled)
kind: Secret
apiVersion: v1
metadata:
  name: my-filerealm-secret
stringData:
  users: |-
    rdeniro:$2a$10$BBJ/ILiyJ1eBTYoRKxkqbuDEdYECplvxnqQ47uiowE7yGqvCEgj9W
    alpacino:$2a$10$cNwHnElYiMYZ/T3K4PvzGeJ1KbpXZp2PfoQD.gfaVdImnHOwIuBKS
    jacknich:{PBKDF2}50000$z1CLJt0MEFjkIK5iEfgvfnA6xq7lF25uasspsTKSo5Q=$XxCVLbaKDimOdyWgLCLJiyoiWpA/XDMe/xtVgn1r5Sg=
  users_roles: |-
    admin:rdeniro
    power_user:alpacino,jacknich
    user:jacknich

---

# Roles in ES format
kind: Secret
apiVersion: v1
metadata:
  name: my-roles-1
stringData:
  roles.yml: |-
    click_admins:
      run_as: [ 'clicks_watcher_1' ]
      cluster: [ 'monitor' ]
      indices:
      - names: [ 'events-*' ]
        privileges: [ 'read' ]
        field_security:
          grant: ['category', '@timestamp', 'message' ]
        query: '{"match": {"category": "click"}}'
    another_role:
      cluster: [ 'all' ]
```

An aggregate secret is built from:
- predefined "internal" users (the `elastic` user, the readiness probe user, the controller user)
- associated users (from Kibana or APMServer associations)
- the user-provided file realm
- internal roles (the readiness probe custom role)
- user-provided roles
and mounted into all Elasticsearch Pods.

User-provided file realm and roles get precedence over internal ones.

This PR also adds an E2E test for user-provided file realm and roles.

## Refactoring

I decided to refactor our Elasticsearch user management code, because I had a hard time figuring out how the existing User and Credentials abstraction mapped to Kubernetes resources, and fitting the new file realm concept into it.

As a result of this refactoring:
- there is no more `Credentials` concept (hashed credentials vs. cleartext credentials)
- predefined users, associated users and user-provided users are separated in their own file
- there is no more common `User` pkg: it was actually mostly useful for Elasticsearch
- it is replaced by an internal private `user` type that can easily transform to a file realm
- association code may relies on the `elasticsearch/user/associated.go` code, which described an `AssociatedUser` and holds all the associated consts
- users reconciliations consists in building file realms for all sort of users and merging those together
- we don't inspect password vs. hash when updating secrets in the reconciler, through `credentials.NeedsUpdate()` (I think that part of the code was hard to grasp). Instead we just reuse existing passwords and hashes when building predefined users in the first place. The secret reconciliation is simpler imo.

Sorry for the super big PR including both the feature & refactoring. Doing one without the other did not make much sense.

___ 

Follow-up PRs:
- add the user-facing documentation


sebgl:
I think we should not error out and block the reconciliation when we fail to parse a user-provided secret (either because secret not found or contains invalid data). Let's log and emit an event instead, similar to what we do for secure settings secrets. I'll add a new commit.

pkg/controller/elasticsearch/user/filerealm/role_users.go
```diff
@@ -0,0 +1,85 @@
...
+import (
+	"fmt"
+	"strings"
+
+	"github.com/elastic/cloud-on-k8s/pkg/utils/set"
+	"github.com/elastic/cloud-on-k8s/pkg/utils/stringsutil"
+)
+
+// roleUsersMapping is a map {role name -> [] user names}
+type roleUsersMapping map[string][]string
```
pebrc: There is a concept of role mapping in Elasticsearch (which you don't mean here) but the naming choices are  somewhat confusing close to that here https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-roles.html

pkg/controller/elasticsearch/user/associated.go
```diff
@@ -0,0 +1,98 @@
...
+	UserNameField = "name"
+	// PasswordHashField is the field in the secret that contains the hash of the password.
+	PasswordHashField = "passwordHash"
+	// UserRolesField is the field in the secret that contains the roles for the user as a comma separated list of strings.
+	UserRolesField = "userRoles"
+
+	fieldNotFound = "field %s not found in secret %s/%s"
+)
+
+// AssociatedUser represents an Elasticsearch user associated for another resource (eg. Kibana, APMServer, etc.).
```
pebrc: associated with?

pkg/controller/elasticsearch/user/filerealm/role_users.go
```diff
@@ -0,0 +1,85 @@
...
+		return other
+	}
+	if len(other) == 0 {
+		return r
+	}
+	for otherRole, otherUsers := range other {
+		currentUsers, exists := r[otherRole]
+		if !exists {
+			// role does not exist yet, create it
+			r[otherRole] = otherUsers
```
pebrc: Should we sort here too?

pkg/controller/elasticsearch/user/filerealm/role_users.go
```diff
@@ -0,0 +1,85 @@
...
+// fileBytes serializes the roleUsersMapping into a file with format:
+// ```
+// role1:user1,user2,user3
+// role2:user1
+// ```
+// Rows are sorted for easier comparison.
+func (r roleUsersMapping) fileBytes() []byte {
+	rows := make([]string, 0, len(r))
+	for role, users := range r {
+		stringsutil.SortStringSlice(users)
```
pebrc: Just for my understanding: you are sorting here again in case the `roleUsersMapping` was manually constructed and not via `mergeWith` which already sorts the users slices?

pkg/controller/elasticsearch/user/reconcile.go
```diff
@@ -0,0 +1,193 @@
...
+
+func getExistingFileRealm(c k8s.Client, es esv1.Elasticsearch) (filerealm.Realm, error) {
+	var secret corev1.Secret
+	if err := c.Get(RolesFileRealmSecretKey(es), &secret); err != nil {
+		return filerealm.Realm{}, err
+	}
+	return filerealm.FromSecret(secret)
+}
+
+// aggregateFileRealm aggregates the various file realms into a single one, and returns the controller user auth.
```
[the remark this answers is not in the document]
sebgl: Do you mean we would rename `esclient.UserAuth` into `esclient.UserCredentials`?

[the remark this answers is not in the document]
barkbay: I just had to check the implementation of `esclient.UserAuth` to understand what was a "user auth" and I was wondering if there is not a better name.
 `returns the controller user credential` seems more familiar to me when speaking about what is used by a user to prove that it is the one it pretends to be.

It could be `UserCredential` or `UsernamePasswordCredential` to make it clear that we are using a username and a password (I _think_ singular is fine).

No strong opinion though.

pkg/controller/elasticsearch/user/filerealm/role_users.go
sebgl: ah indeed! I should rename it.
`usersRoles` would be closer to the `users_roles` file, but I don't like how it seems to express a {user -> roles} relationship whereas we actually have {role -> users}.
Any naming suggestion?

pkg/controller/elasticsearch/user/filerealm/role_users.go
sebgl: I would expect `other` to be already made of sorted users.
Whereas we need to sort when merging the 2 users slices.

pkg/controller/elasticsearch/user/filerealm/role_users.go
sebgl: Hmm you're right, sorting is not very consistent in this file.
I think the following would make sense:
- the internal representation should always contain sorted users
- when we parse the raw file, we sort users
- when we merge 2 `roleUsersMapping`, we should expect both already have sorted users
- when we serialize back to bytes, we don't have to re-sort users since they're already sorted

I'll change the code accordingly.

pkg/controller/elasticsearch/user/filerealm/role_users.go
sebgl: see https://github.com/elastic/cloud-on-k8s/pull/2682#discussion_r390846007

pkg/controller/elasticsearch/user/reconcile.go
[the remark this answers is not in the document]
pebrc: I think I called it `Auth` because it will be passed on as an `Authorization` header in the client.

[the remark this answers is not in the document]
sebgl: Would it make sense to name it `esclient.BasicAuth`?
We could also move it to something like `utils/http/basicauth.go`. I don't think there's a generic `BasicAuth` type we can reuse from go stdlib.

pkg/controller/elasticsearch/user/associated.go
```diff
@@ -0,0 +1,98 @@
...
+		users = append(users, u)
+	}
+	return fromAssociatedUsers(users), nil
+}
+
+// parseAssociatedUserSecret reads an associated user from a secret.
+func parseAssociatedUserSecret(secret v1.Secret) (AssociatedUser, error) {
+	user := AssociatedUser{}
+	if len(secret.Data) == 0 {
+		return user, pkgerrors.Errorf("user secret %s/%s is empty", secret.Namespace, secret.Name)
```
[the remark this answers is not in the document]
sebgl: You're right. Probably something we should agree on for the entire codebase.
In the meantime I'll switch those to `fmt.Errorf` since they are clearly the only exceptions in that package.

pkg/controller/elasticsearch/user/filerealm/role_users.go
pebrc: Not really, your original name was actually good. I see your point about the inverse logic of the `users_roles` naming, but maybe it's still worth considering given that it would be recognisable to readers who are familiar with the files generated.

pkg/controller/elasticsearch/user/reconcile.go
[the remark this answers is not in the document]
sebgl: I renamed the type to `esclient.BasicAuth`, and changed the user pkg function names to better match "credentials".

pkg/controller/elasticsearch/user/filerealm/role_users.go
sebgl: Let's do `usersRoles` but make sure we explain the reverse mapping in comments.

pkg/controller/common/watches/secrets.go
```diff
@@ -0,0 +1,37 @@
...
+	"k8s.io/apimachinery/pkg/types"
+)
+
+// WatchUserProvidedSecrets registers a watch for user-provided secrets.
+// Only one watch per watcher is registered:
+// - if it already exists with different secrets, it is replaced to watch the new secrets.
+// - if there is non secrets provided by the user, remove the watch.
+func WatchUserProvidedSecrets(
+	watcher types.NamespacedName, // resources to which the watches are attached (eg. an Elasticsearch object)
+	watched DynamicWatches, // resources already watched by watched
```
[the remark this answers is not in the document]
sebgl: haha yes this doesn't make sense 😄

pkg/controller/common/watches/secrets.go
```diff
@@ -0,0 +1,37 @@
...
+		return nil
+	}
+	userSecretNsns := make([]types.NamespacedName, 0, len(secrets))
+	for _, secretName := range secrets {
+		userSecretNsns = append(userSecretNsns, types.NamespacedName{
+			Namespace: watcher.Namespace,
+			Name:      secretName,
+		})
+	}
+	return watched.Secrets.AddHandler(NamedWatch{
```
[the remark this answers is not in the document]
sebgl: Good catch! Will add to `(r *ReconcileElasticsearch) onDelete`.

pkg/controller/elasticsearch/user/filerealm/users_roles.go
```diff
@@ -0,0 +1,89 @@
...
+	return rolesMapping, forEachRow(data, func(row []byte) error {
+		roleUsers := strings.Split(string(row), ":")
+		if len(roleUsers) != 2 {
+			return fmt.Errorf("invalid entry in users_roles")
+		}
+		role := roleUsers[0]
+		users := strings.Split(roleUsers[1], ",")
+		// sort users for consistent comparison
+		stringsutil.SortStringSlice(users)
+		if len(users) == 1 && users[0] == "" {
```
barkbay: Why not splitting only if there are some roles ?

```golang
		var users []string
		if len(roleUsers[1]) > 0 {
			users = strings.Split(roleUsers[1], ",")
			// sort users for consistent comparison
			stringsutil.SortStringSlice(users)
		}
```

pkg/controller/elasticsearch/user/reconcile.go
```diff
@@ -0,0 +1,189 @@
...
+	"github.com/elastic/cloud-on-k8s/pkg/utils/k8s"
+	"github.com/elastic/cloud-on-k8s/pkg/utils/maps"
+)
+
+var log = logf.Log.WithName("elasticsearch-user")
+
+// ReconcileUsersAndRoles fetches all users and roles and aggregates them into a single
+// Kubernetes secret mounted in the Elasticsearch Pods.
+// That secret contains the file realm files (`users` and `users_roles`) and the file roles (`roles.yml`).
+// users are aggregated from various sources:
```
barkbay:
```suggestion
// Users are aggregated from various sources:
```

pkg/controller/elasticsearch/user/user.go
```diff
@@ -5,76 +5,60 @@
...
-
-// Password returns the password of this user.
-func (u User) Password() string {
-	return u.password
-}
-
-// Roles returns any roles of this user.
-func (u User) Roles() []string {
-	return u.roles
+// basicAuth returns basic auth credentials for the given user.
```
barkbay:
```suggestion
// credentialsFor returns basic auth credentials for the given user.
```

pkg/controller/elasticsearch/user/user_provided.go
```diff
@@ -0,0 +1,155 @@
...
+		watched,
+		UserProvidedFileRealmWatchName(esKey),
+		secretNames,
+	); err != nil {
+		return filerealm.Realm{}, err
+	}
+	return retrieveUserProvidedFileRealm(c, es, recorder)
+}
+
+// reconcileUserProvidedRoles returns aggregate roles from the referenced sources in the es spec.
```
barkbay: "returns aggregated roles" ?

sebgl: I asked myself the same question when writing this 😄 
From https://english.stackexchange.com/questions/241413/is-there-a-difference-in-meaning-between-aggregate-and-aggregated, it seems "aggregate" can/should be used as an adjective.

I can also see `aggregated` being used in many places, so I guess both can be used.

pkg/controller/kibanaassociation/association_controller.go
```diff
@@ -66,6 +65,8 @@ const (
 	kibanaUserSuffix = "kibana-user"
 	// ElasticsearchCASecretSuffix is used as suffix for CAPublicCertSecretName
 	ElasticsearchCASecretSuffix = "kb-es-ca" // nolint
+	// KibanaSystemUserBuiltinRole is the name of the built-in role for the Kibana system user
```
pebrc: .

test/e2e/es/auth_test.go
```diff
@@ -0,0 +1,225 @@
...
+	ctx, cancel := context.WithTimeout(context.Background(), esclient.DefaultReqTimeout)
+	defer cancel()
+	doc := bytes.NewBufferString(`{"foo": "bar"}`)
+	req, err := http.NewRequest(http.MethodPost, fmt.Sprintf("/%s/_doc", index), doc)
+	if err != nil {
+		return err
+	}
+	resp, err := esClient.Request(ctx, req)
+	if err != nil {
+		// the client wraps unexpected status code in an APIError, but still returns the correct resp
```
pebrc: codes?

pkg/controller/elasticsearch/initcontainer/prepare_fs.go
```diff
@@ -7,15 +7,17 @@ package initcontainer
 import (
 	"path"
 
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+
```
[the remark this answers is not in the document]
sebgl: we discussed offline: it may be related to IntelliJ settings. We agreed that it's OK to have several import groups as long as they make sense (eg. no `github.com/elastic` in different groups).

jsenon:
Thanks for adding this functionality. 
We are using applications that needs to have different login/password and roles, these credentials are part of the application and are deployed on the K8s cluster by CI. If I'm right, with this process, we have to update a single secret yaml files with credentials for all our applications. 
Do you plan to have a merge of multiple secret yaml file into a realmfile ?

sebgl:
@jsenon yes! Several secrets can be referenced, and will be merged together. See https://github.com/elastic/cloud-on-k8s/pull/2703 which contains documentation for this.

Feedback welcome (in https://github.com/elastic/cloud-on-k8s/issues/728) on this feature!

maxdanilov:
Has this feature been made available already? We're running eck 1.1.2 and can't seem to be able to configure users like specified in this PR (as if `spec.auth` is ignored)

sebgl:
@maxdanilov yes it is there in ECK 1.1.2. See https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-users-and-roles.html.

If you can't make it work please create a thread in https://discuss.elastic.co/c/eck with the details of your yaml manifests.

PR #2687, opened 2020-03-10
Add the transport service DNS name to the CSR

pebrc:
We initially connect through the transport service to the remote
cluster. This should  ensure that the transport service is one of the trusted
DNS names.

I ran into errors like this one when testing remote cluster connection setup

```
[elasticsearch.server][WARN] failed to establish trust with server at [remote-src-es-transport.default.svc]; the server provided a certificate with subject name [CN=remote-src-es-default-2.node.remote-src.default.es.local,OU=remote-src] and fingerprint [effb883baf24e32229edbe17361034c23c6c1a61]; the certificate has subject alternative names [DNS:remote-src-es-default-2.node.remote-src.default.es.local,IP:10.73.0.8,IP:127.0.0.1]; the certificate is issued by [CN=remote-src-transport,OU=remote-src]; the certificate is signed by (subject [CN=remote-src-transport,OU=remote-src] fingerprint [1743d054ed76a38e7a212781318bacfdc3ea0748] {trusted issuer}) which is self-issued; the [CN=remote-src-transport,OU=remote-src] certificate is trusted in this ssl context ([xpack.security.transport.ssl])   
```

### DIFF
--- a/pkg/controller/elasticsearch/certificates/transport/csr.go
+++ b/pkg/controller/elasticsearch/certificates/transport/csr.go
@@ -88,6 +88,7 @@ func buildGeneralNames(
 		{OtherName: *commonNameOtherName},
 		{DNSName: commonName},
 		// add the transport service name for remote cluster connections initially connecting through the service
+		// the DNS name has to match the seed hosts configured in the remote cluster settings
 		{DNSName: fmt.Sprintf("%s.%s.svc", esv1.TransportService(cluster.Name), cluster.Namespace)},
 		{IPAddress: netutil.MaybeIPTo4(podIP)},
 		{IPAddress: net.ParseIP("127.0.0.1").To4()},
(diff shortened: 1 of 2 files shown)


pkg/controller/elasticsearch/certificates/transport/csr.go
```diff
@@ -87,6 +87,8 @@ func buildGeneralNames(
 	generalNames := []certificates.GeneralName{
 		{OtherName: *commonNameOtherName},
 		{DNSName: commonName},
+		// add the transport service name for remote cluster connections initially connecting through the service
+		{DNSName: fmt.Sprintf("%s.%s.svc", esv1.TransportService(cluster.Name), cluster.Namespace)},
```
anyasabo: We have `GetServiceDNSName()` in pkg/util/k8s which might be better to use here, and also includes the name without `.svc` (which also works)

barkbay: It will not work unless we also change the [seed hosts](https://github.com/elastic/cloud-on-k8s/blob/master/pkg/controller/elasticsearch/remotecluster/elasticsearch.go#L63)

anyasabo: Makes sense, would a comment be helpful here to explain why we're not? It seems like a gotcha since the name resolves. Though to be fair, the fully qualified domain name with the default domain suffix (e.g. cluster.local) also resolves but would fail.

PR #2688, opened 2020-03-10
Add the EnterpriseSearch CRD and controller

sebgl:
Introduce the EnterpriseSearch CRD and controller in ECK.

It seems to be working fine so far but has not been extensively tested.
To be used, it requires the `image` field to be set with a valid Docker image.

This is obviously work in progress, since:
- parts of the controller should be refactored to share common code with APMServer and Kibana
- most of the association controller should be refactored to share common code with APMServer and Kibana
- there are 0 unit tests (mostly because I wanted to go fast in the prototype phase, and because many unit tests should be written along with the refactoring above)
- there are 0 E2E tests
- it is missing a few features (podDisruptionBudget, secret config refs, version upgrade handling, etc.)


config/crds/patches/kustomization.yaml
```diff
@@ -42,3 +42,10 @@ patchesJson6902:
       kind: CustomResourceDefinition
       name: kibanas.kibana.k8s.elastic.co
     path: apm-kibana-patches.yaml
+  # custom patches for EnterpriseSearch
```
jonasll: Proper name here is Enterprise Search, two words.

docs/reference/api-docs.asciidoc
```diff
@@ -796,6 +800,83 @@ UpdateStrategy specifies how updates to the cluster should be performed.
...
+.Resource Types
+- xref:{anchor_prefix}-github-com-elastic-cloud-on-k8s-pkg-apis-enterprisesearch-v1beta1-enterprisesearch[$$EnterpriseSearch$$]
+- xref:{anchor_prefix}-github-com-elastic-cloud-on-k8s-pkg-apis-enterprisesearch-v1beta1-enterprisesearchlist[$$EnterpriseSearchList$$]
+
+
+
+[id="{anchor_prefix}-github-com-elastic-cloud-on-k8s-pkg-apis-enterprisesearch-v1beta1-enterprisesearch"]
+=== EnterpriseSearch 
+
+EnterpriseSearch represents an Enterprise Search resource in a Kubernetes cluster.
```
jonasll: Can we make sure to outline that EnterpriseSearch here is a technical way of referring to Enterprise Search, the product? Naming is extra sensitive and we need make sure that we perhaps wrap it in `EnterpriseSearch` a `pre` wrap or something.

docs/reference/api-docs.asciidoc
```diff
@@ -796,6 +800,83 @@ UpdateStrategy specifies how updates to the cluster should be performed.
...
+| *`metadata`* __link:https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.17/#objectmeta-v1-meta[$$ObjectMeta$$]__ | Refer to Kubernetes API documentation for fields of `metadata`.
+
+| *`spec`* __xref:{anchor_prefix}-github-com-elastic-cloud-on-k8s-pkg-apis-enterprisesearch-v1beta1-enterprisesearchspec[$$EnterpriseSearchSpec$$]__ | 
+|===
+
+
+[id="{anchor_prefix}-github-com-elastic-cloud-on-k8s-pkg-apis-enterprisesearch-v1beta1-enterprisesearchlist"]
+=== EnterpriseSearchList 
+
+EnterpriseSearchList contains a list of EnterpriseSearch
```
jonasll: Same comment on communicating tech-level items vs. product name.

config/crds/patches/kustomization.yaml
sebgl: This really refers to the `EnterpriseSearch CRD` (the technical implementation), I'd be in favor of keeping it as is. Also this is def not a user-facing file.

docs/reference/api-docs.asciidoc
sebgl: This doc is auto-generated from the following go code comment:
```go
// EnterpriseSearch represents an Enterprise Search resource in a Kubernetes cluster.
type EnterpriseSearch struct {
```
There's a go linter rule that specifies each commented struct must start with `// <structName> ...`, so it's not easy to change it to eg. `<pre>EnterpriseSearch</pre>`.

I can change it to something like `EnterpriseSearch is a Kubernetes CRD to represent Enterprise Search.` if that makes more sense?

This is the CRD API doc, we'll definitely have more user-facing docs in the future in which it's important we make that distinction clear 👍

pkg/controller/enterprisesearch/enterprisesearch_controller.go
```diff
@@ -0,0 +1,330 @@
...
+	}
+
+	state.UpdateEnterpriseSearchExternalService(*svc)
+
+	//// update status
+	//err = r.updateStatus(ctx, state)
+	//if err != nil && errors.IsConflict(err) {
+	//	log.V(1).Info("Conflict while updating status", "namespace", as.Namespace, "as", as.Name)
+	//	return reconcile.Result{Requeue: true}, nil
+	//}
```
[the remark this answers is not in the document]
sebgl: Yeah the whole "update status" part is missing. I left it commented to not forget about it and keep track of the general code flow, but I'll just replace it with a `// TODO`

pkg/controller/enterprisesearch/state.go
```diff
@@ -0,0 +1,50 @@
...
+// UpdateApmServerState updates the ApmServer status based on the given deployment.
+func (s State) UpdateEnterpriseSearchState(deployment v1.Deployment) {
+	s.EnterpriseSearch.Status.AvailableNodes = deployment.Status.AvailableReplicas
+	// TODO
+	//s.EnterpriseSearch.Status.Health = apmv1.ApmServerRed
+	//for _, c := range deployment.Status.Conditions {
+	//	if c.Type == v1.DeploymentAvailable && c.Status == corev1.ConditionTrue {
+	//		s.EnterpriseSearch.Status.Health = apmv1.ApmServerGreen
+	//	}
+	//}
```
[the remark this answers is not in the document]
sebgl: 👍 replaced the commented code with a cleaner `// TODO update status health`

PR #2691, opened 2020-03-11
Make transport service customizable

### LINKED ISSUE #2644: Users should be able to customize the transport service
#2543 introduced a transport service to let Elasticsearch nodes communicate to remote Elasticsearch clusters.

A user should be able to control the spec of this service like we do for the HTTP service (e.g. set the type to `LoadBalancer`)

pebrc:
* adds transport service template to Elasticsearch spec. Mirroring what we have for the HTTP layer
* adds the ability to change the immutable  `clusterIP` field by deleting and re-creating the service if needed
* in order to allow re-creation of resource adds an optional `NeedsDelete` callback to `ReconcileResource` 

```
kind: Elasticsearch
metadata:
  name: with-custom-transport
spec:
  version: 7.6.1
  transport:
    service:
      metadata:
        labels:
          foo: bar
      spec:
        type: LoadBalancer
  nodeSets:
    count: 3
    name: default
```

This is useful for example if users want to either influence the metadata on the transport service or if they want to expose it via `LoadBalancer` for CCR/CCS across k8s boundaries. 

Fixes #2644


pkg/apis/elasticsearch/v1/elasticsearch_types.go
```diff
@@ -27,6 +27,10 @@ type ElasticsearchSpec struct {
 	// +kubebuilder:validation:Optional
 	HTTP commonv1.HTTPConfig `json:"http,omitempty"`
 
+	// HTTP holds Transport layer settings for Elasticsearch.
+	// +kubebuilder:validation:Optional
+	Transport TransportConfig `json:"transport,omitempty"`
```
sebgl: ~HTTP~

pkg/controller/common/reconciler/reconciler.go
```diff
@@ -39,6 +39,8 @@ type Params struct {
 	Reconciled runtime.Object
 	// NeedsUpdate returns true when the object to be reconciled has changes that are not persisted remotely.
 	NeedsUpdate func() bool
+	// NeedsDelete returns true when the object to be reconciled needs to be deleted and re-created because it cannot be updated.
+	NeedsDelete func() bool
```
sebgl: Would `NeedsRecreate` be more explicit?

pkg/controller/common/reconciler/reconciler.go
```diff
@@ -91,9 +93,7 @@ func ReconcileResource(params Params) error {
 		}
 	}
 
-	// Check if already exists
-	err = params.Client.Get(types.NamespacedName{Name: name, Namespace: namespace}, params.Reconciled)
-	if err != nil && apierrors.IsNotFound(err) {
+	create := func() error {
 		// Create if needed
```
sebgl: we should probably remove that comment now

pkg/controller/common/reconciler/reconciler.go
```diff
@@ -112,9 +112,24 @@ func ReconcileResource(params Params) error {
...
+	if err != nil && apierrors.IsNotFound(err) {
+		return create()
 	} else if err != nil {
 		log.Error(err, fmt.Sprintf("Generic GET for %s %s/%s failed with error", kind, namespace, name))
-		return errors.Wrapf(err, "failed to get %s %s/%s", kind, namespace, name)
+		return fmt.Errorf("failed to get %s %s/%s: %w", kind, namespace, name, err)
+	}
+
+	if params.NeedsDelete != nil && params.NeedsDelete() {
+		log.Info("Deleting resource", "kind", kind, "namespace", namespace, "name", name)
```
sebgl: I'm wondering if we should add an additional log here:

```
log.Info("Resource cannot be updated, hence will be deleted then recreated", ...)
```
Otherwise looking at logs could be surprising ("why is my resource deleted then recreated again?)

pkg/controller/common/reconciler/reconciler.go
```diff
@@ -112,9 +112,24 @@ func ReconcileResource(params Params) error {
...
+		err := params.Client.Delete(params.Expected)
```
sebgl: We should apply delete preconditions here to ensure we don't re-delete the object we just recreated at the previous reconciliation.

```go

opt := client.Preconditions{
		UID:             reconciled.UID,
		ResourceVersion: reconciled.ResourceVersion,
}
err := k8sClient.Delete(reconciled, opt)
```

sebgl: Should we ignore the `apierrors.IsNotFound(err)` error here?

pkg/controller/common/service_control.go
```diff
@@ -53,18 +57,32 @@ func ReconcileService(
...
-	if expected.Spec.ClusterIP == "" {
-		expected.Spec.ClusterIP = reconciled.Spec.ClusterIP
-	}
+	applyServerSideValues(expected, reconciled)
+	// if the specs, labels, or annotations differ, the object should be updated
+	return !(reflect.DeepEqual(expected.Spec, reconciled.Spec) &&
+		compare.LabelsAndAnnotationsAreEqual(expected.ObjectMeta, reconciled.ObjectMeta))
+}
 
+// applyServerSideValues applies any default the master might have set from the reconciled version to the expected one.
```
sebgl: `the master`?

pkg/controller/elasticsearch/services/services.go
```diff
@@ -33,29 +32,30 @@ func TransportServiceName(esName string) string {
...
-			ClusterIP:       "None",
-			SessionAffinity: corev1.ServiceAffinityNone,
-			Type:            corev1.ServiceTypeClusterIP,
-			// Nodes need to discover themselves before the pod is considered ready,
-			// otherwise minimum master nodes would never be reached
-			PublishNotReadyAddresses: true,
+	svc := corev1.Service{
+		ObjectMeta: es.Spec.Transport.Service.ObjectMeta,
+		Spec:       es.Spec.Transport.Service.Spec,
+	}
```
sebgl: `svc := es.Spec.Transport.Service.DeepCopy()`?

pkg/controller/common/reconciler/reconciler.go
pebrc: Funny I had that originally.

pkg/controller/common/reconciler/reconciler.go
pebrc: Excellent point! I did not think of that.

pebrc: Yes

pkg/controller/common/service_control.go
pebrc: This is inspired by the language used in the k8s API docs

pkg/controller/elasticsearch/services/services.go
pebrc: No, because that's a `ServiceTemplate` and we want a `Service` here

sebgl: Got it 👍

pkg/controller/common/service_control.go
sebgl: ok 🤷‍♂ 

Still wondering if it could be worth replacing in that comment? In Elasticsearch context `master` can be confusing.
`// applyServerSideValues applies any default that may have been set from the reconciled version...` feels explicit enough and avoids ambiguity?

pkg/apis/elasticsearch/v1/elasticsearch_types.go
```diff
@@ -57,6 +61,12 @@ type ElasticsearchSpec struct {
 	RemoteClusters []RemoteCluster `json:"remoteClusters,omitempty"`
 }
 
+// TransportConfig holds the Transport layer configuration for Elasticsearch.
```
sebgl: Should we make this the same as line 30, for consistency?

pkg/controller/common/reconciler/reconciler.go
```diff
@@ -111,9 +111,38 @@ func ReconcileResource(params Params) error {
...
+	}
+
+	if params.NeedsRecreate != nil && params.NeedsRecreate() {
+		log.Info("Resource cannot be updated, hence will be deleted and then recreated", "kind", kind, "namespace", namespace, "name", name)
+		log.Info("Deleting resource", "kind", kind, "namespace", namespace, "name", name)
+		reconciledMeta, err := meta.Accessor(params.Reconciled)
+		if err != nil {
+			return err
+		}
+		// Using a precondition here to make sure we delete the version of the resource we intend ot delete and
```
sebgl: `ot delete` -> `to delete`

pkg/controller/common/service_control.go
```diff
@@ -50,18 +54,32 @@ func ReconcileService(
 	return reconciled, err
 }
 
+func needsDelete(expected, reconciled *corev1.Service) bool {
```
sebgl: maybe also `needsRecreate` for consistency?

anyasabo:
jenkins test this please

pkg/controller/common/reconciler/reconciler.go
```diff
@@ -111,9 +111,38 @@ func ReconcileResource(params Params) error {
...
+	if err != nil && apierrors.IsNotFound(err) {
+		return create()
 	} else if err != nil {
 		log.Error(err, fmt.Sprintf("Generic GET for %s %s/%s failed with error", kind, namespace, name))
-		return errors.Wrapf(err, "failed to get %s %s/%s", kind, namespace, name)
+		return fmt.Errorf("failed to get %s %s/%s: %w", kind, namespace, name, err)
+	}
+
+	if params.NeedsRecreate != nil && params.NeedsRecreate() {
+		log.Info("Resource cannot be updated, hence will be deleted and then recreated", "kind", kind, "namespace", namespace, "name", name)
```
anyasabo: Nit: it seems like we can consolidate these two info messages

anyasabo:
Jenkins test this please

pkg/controller/common/reconciler/reconciler.go
sebgl: I think it's fine to keep it separate, so we can search for `"Deleting resource"` or `"Creating resource"` in logs to see all created resources, this one not being an exception.

pebrc:
I realised (thanks to @anyasabo ) that we should update the docs to reflect this change. I will push another commit to that end before merging this.

pebrc:
Jenkins test this please

PR #2702, opened 2020-03-13
Add Istio gateway recipe

charith-elastic:
This recipe illustrates how to use Istio to split traffic (reads -> coordinator nodes, writes -> ingest nodes) and how to expose Elasticsearch and Kibana deployments using the Istio ingress gateway.

We have had some user questions about splitting traffic and this is one way of achieving it if Istio is installed on their cluster.


charith-elastic:
Jenkins test this please

config/recipes/istio-gateway/03-elasticsearch-kibana.yaml
```diff
@@ -0,0 +1,113 @@
...
+        disabled: true
+  nodeSets:
+  - name: master
+    count: 1
+    config:
+      node.master: true
+      node.data: false
+      node.ingest: false
+      node.ml: false
+      node.store.allow_mmap: false
```
pebrc: I wonder if we should not use this setting everywhere. We added it to the quickstart to avoid that users have a bad experience when they try out ECK for the first time, but I am bit concerned that we are adding it everywhere now without comment.

charith-elastic: I added a README to the root of the recipes directory as you suggested and included a warning on it about using this setting. Do you think that's sufficient?

I could get rid of the setting from the manifests but that could increase the likelihood of these recipes not working out of the box -- leading to a bad user experience.

pebrc: 👍

config/recipes/istio-gateway/README.asciidoc
```diff
@@ -0,0 +1,134 @@
+= Expose Elasticsearch and Kibana using Istio ingress gateway
```
sebgl: Any reason for picking asciidoc instead of Markdown?
Is asciidoc correctly rendered by Github?

charith-elastic: GitHub has support for AsciiDoc but not the entire spec. I chose AsciiDoc because it's the format we use in our documentation and also because it supports includes and syntax highlighting. With includes, it's easy to write a short explanation about each manifest without having to duplicate the content in two places.

Issue #2704, opened 2020-03-13
Test GKE clusters not using pod network parameter


anyasabo:
Newly created e2e test clusters are being created with the default network size for the pod network (/14), but the service network is being correctly set (to /20). They should both be /20. This used to work and my initial suspicion was that it was likely a regression from https://github.com/elastic/cloud-on-k8s/issues/2615

But I can reproduce it with
```
gcloud beta container clusters  --project elastic-cloud-dev create sabomanualtest  \
--create-subnetwork range=/21 --region europe-west1 --enable-ip-alias
```

The [docs](https://cloud.google.com/sdk/gcloud/reference/beta/container/clusters/create#--create-subnetwork) say range can be just the mask, so I wonder if this is a GKE bug or user error. Experimenting more with it now.

Example test failure: https://devops-ci.elastic.co/job/cloud-on-k8s-e2e-tests-stack-versions/21/consoleFull

Ref: https://github.com/elastic/cloud-on-k8s/pull/2473

Issue #2615, opened 2020-02-25
CI failing: exceeded secondary ranges per subnetwork


anyasabo:
CI is failing to spin up GKE environments with

>Google Compute Engine: Exceeded maximum supported number of secondary ranges per subnetwork: 30.

I thiiiink we can alter the deployer to create a new subnetwork for each cluster:
https://cloud.google.com/sdk/gcloud/reference/container/clusters/create#--create-subnetwork
but I'm not sure of the trade offs right now.

anyasabo:
Did some more reading:

So each subnet has a primary (the actual addr that gets assigned to the VM) range and [secondary (alias) range](https://cloud.google.com/vpc/docs/alias-ip), used for pods and similar things.

We're currently using 30/50 VPC networks, and 123/275 subnets. These limits are quotas and we can request increases. I have no idea how high they can go.

Each subnet has a hard limit of 30 secondary ranges, which is what we are hitting. This cannot be increased.

[Limits doc](https://cloud.google.com/vpc/docs/quota#per_network)

Each (VPC-native, which is the newer and recommended kind) GKE cluster as it stands is created in the default VPC network and in the default VPC subnet. The nodes get put in that subnet, then it creates [two alias IP ranges](https://cloud.google.com/kubernetes-engine/docs/how-to/alias-ips#cluster_sizing) -- one for services and one for pods. We can adjust which subnet GKE uses, but it will always create two alias IP ranges for that subnet. So I think currently, we have a hard limit of 15 GKE instances created using our deployment tool in the current configuration.

And you can see that we have almost that number right now, all from CI

![Screen Shot 2020-02-25 at 4 21 02 PM](https://user-images.githubusercontent.com/1638148/75293018-ef7afd00-57ea-11ea-9b5c-81eecc6add0b.png)


We last touched this with Seb's PR that reduced the size of the subnets: https://github.com/elastic/cloud-on-k8s/pull/2473
But now we are just hitting a limit in the number of subnets rather than the size of them.

If you use the [`--create-subnetwork`](https://cloud.google.com/sdk/gcloud/reference/container/clusters/create#--create-subnetwork) option when creating a cluster, it uses the VPC you specify (currently hardcoded to `default` in our case) but creates a new subnet in it with a default name (like `gke-sabo-dev-cluster-subnet-8c20dc23` for my `sabo-dev-cluster` cluster). I think using this param is the best path forward and don't see any significant downsides.


[...]

PR #2705, opened 2020-03-13
Correct GKE subnet size

anyasabo:
Fix https://github.com/elastic/cloud-on-k8s/issues/2704

It looks like GKE changed its behavior with [`create-subnetwork`](https://cloud.google.com/sdk/gcloud/reference/beta/container/clusters/create#--create-subnetwork). Some of the older clusters that were created since https://github.com/elastic/cloud-on-k8s/issues/2615 was merged (you can tell because the subnets are something like `gke-PERSONSNAME-dev-cluster-subnet-635849b0`) have the /20 (rather than the default /14) pod address range set correctly. But the newer ones spun up for E2E testing do not. You can also reproduce it by using

```
gcloud beta container clusters  --project yourproject create yourtest \
  --create-subnetwork range=/20 --region europe-west1 --enable-ip-alias \
  --cluster-ipv4-cidr=/20
```

And removing either the `range=/20` or `cidr=/20`. Before, we were specifying the cluster cidr but not the subnetwork range, but ti would still set it to /14. Setting just the subnetwork range parameter also ended up just using the /14 default. Setting both together (as in this PR) worked. I opened a docs issue with them because I'm a little bit grumpy about it.

### DIFF
--- a/hack/deployer/runner/gke.go
+++ b/hack/deployer/runner/gke.go
@@ -26,8 +26,10 @@ overrides:
 
 var (
 	// GKE uses 18 chars to prefix the pvc created by a cluster
-	pvcPrefixMaxLength    = 18
-	GkeStorageProvisioner = "kubernetes.io/no-provisioner"
+	pvcPrefixMaxLength      = 18
+	GkeStorageProvisioner   = "kubernetes.io/no-provisioner"
+	defaultClusterIPv4CIDR  = "/20"
+	defaultServicesIPv4CIDR = "/20"
 )
 
 func init() {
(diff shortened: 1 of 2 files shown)


hack/deployer/runner/gke.go
```diff
@@ -173,21 +186,14 @@ func (d *GkeDriver) create() error {
...
-
 	return NewCommand(`gcloud beta container --project {{.GCloudProject}} clusters create {{.ClusterName}} ` +
 		`--region {{.Region}} --username {{.AdminUsername}} --cluster-version {{.KubernetesVersion}} ` +
 		`--machine-type {{.MachineType}} --image-type COS --disk-type pd-ssd --disk-size 30 ` +
 		`--local-ssd-count {{.LocalSsdCount}} --scopes {{.GcpScopes}} --num-nodes {{.NodeCountPerZone}} ` +
 		`--enable-stackdriver-kubernetes --addons HorizontalPodAutoscaling,HttpLoadBalancing ` +
 		`--no-enable-autoupgrade --no-enable-autorepair --enable-ip-alias --metadata disable-legacy-endpoints=true ` +
 		`--network projects/{{.GCloudProject}}/global/networks/default ` +
-		`--create-subnetwork "" ` +
+		`--create-subnetwork range={{.ClusterIPv4CIDR}} --cluster-ipv4-cidr={{.ClusterIPv4CIDR}} --services-ipv4-cidr={{.ServicesIPv4CIDR}} ` +
```
sebgl: Just to make sure I'm not missing something: the big change in this PR is adding `range={{.ClusterIPv4CIDR}}`, the rest is just refactoring?

anyasabo: Correct, and making the /20 a default in the code rather than in the yaml (just to make it simpler since we always want to use that parameter).

Issue #2689, opened 2020-03-11
Create a recipe for ingress


david-kow:
Create a recipe under https://github.com/elastic/cloud-on-k8s/tree/master/config/recipes with example ingress configuration.

sebgl:
An example from a while back (to be tested again): https://github.com/elastic/cloud-on-k8s/issues/1010#issuecomment-499795679

PR #2711, opened 2020-03-17
Add Google Cloud Load Balancer recipe

### LINKED ISSUE #2489: Document how to set up Ingress with a Google Cloud HTTP load balancer
https://discuss.elastic.co/t/possible-problem-with-elasticsearch-readinessprobe/217094

highlights issues setting up an Ingress with Google Cloud's HTTP load balancer. We should write documentation that gives an example of how to do that and addresses issues around: 
* ECK's self-signed TLS certificates (if run in default configuration)
* HTTP vs. HTTP health checks run by the load balancer

charith-elastic:
Illustrates how to expose Elasticsearch and Kibana using a Google Cloud Load Balancer (GCLB).  Also illustrates how to use custom certificates for external traffic while preserving the ECK-generated certificates for internal traffic.

Fixes #2489, #2689


config/recipes/gclb/01-elastic-stack.yaml
```diff
@@ -0,0 +1,74 @@
+---
+apiVersion: elasticsearch.k8s.elastic.co/v1
+kind: Elasticsearch
+metadata:
+  name: hulk
+  labels:
+    app: hulk
```
sebgl: Do we need this label?

charith-elastic: For me personally, it's just convenient to have a single short label that can be used to delete everything belonging to a specific deployment. 
It's also an informal convention used by some tools as well.

config/recipes/gclb/01-elastic-stack.yaml
```diff
@@ -0,0 +1,74 @@
...
+    service:
+      metadata:
+        labels:
+          app: hulk
+        annotations:
+          # Enable TLS between GCLB and the application
+          cloud.google.com/app-protocols: '{"https":"HTTPS"}'
+          service.alpha.kubernetes.io/app-protocols: '{"https":"HTTPS"}'
+          # Uncomment the following line to enable container-native load balancing.
+          #cloud.google.com/neg: '{"ingress": true}'
```
anyasabo: It seems like we probably want this to be the default, correct? If I'm reading the [docs](https://cloud.google.com/kubernetes-engine/docs/concepts/container-native-load-balancing) right there don't seem to be any substantial downsides

config/recipes/gclb/01-elastic-stack.yaml
```diff
@@ -0,0 +1,74 @@
...
+      spec:
+        type: NodePort
```
anyasabo: Does this need to be NodePort? It's not clear reading the GKE docs but the examples show ClusterIP. My gut feeling is that you pretty much always want to use ClusterIP unless there's a good reason to not.

config/recipes/gclb/01-elastic-stack.yaml
charith-elastic: If you forget to add the IP aliasing flag when creating the Kubernetes cluster, this will silently fail.  I assume that most people are still using the legacy cluster type so I kept this commented out for the time being.

config/recipes/gclb/01-elastic-stack.yaml
charith-elastic: If you are not using container-native load balancing, the service has to be a `NodePort` because the load balancer sends traffic to the node which then gets forwarded by the kube proxy to the container.

config/recipes/gclb/01-elastic-stack.yaml
anyasabo: Ah the default in the GUI is to use vpc native clusters. I can see it either way. If you don't create it as VPC native in the CLI it yells at you at least:
WARNING: Currently VPC-native is not the default mode during cluster creation. In the future, this will become the default mode and can be disabled using --no-enable-ip-alias flag. Use --[no-]enable-ip-alias flag to suppress this warning.

But I can see how the silent failure is a pain to debug 🤷‍♀

config/recipes/gclb/01-elastic-stack.yaml
anyasabo: 👍 with a comment to that effect I'm good to merge this in. I think in general I lean towards having the "recommended" way in the recipes also be the GKE recommended way (i.e. vpc native clusters), but as long as we document how to use them for VPC native clusters that should be enough to get people going

config/recipes/gclb/01-elastic-stack.yaml
```diff
@@ -0,0 +1,74 @@
...
+          username: anon
+          roles: monitoring_user
+    podTemplate:
+      spec:
+        containers:
+          - name: elasticsearch
+            readinessProbe:
+              httpGet:
+                scheme: HTTPS
+                port: 9200
```
pebrc: Maybe add a comment explaining why we need to change the readiness probe here.

config/recipes/gclb/01-elastic-stack.yaml
charith-elastic: The Google example uses a NodePort: https://cloud.google.com/kubernetes-engine/docs/how-to/container-native-load-balancing#create_service

But, I agree with you that there's no need to make the service do more than it really needs to.

PR #2715, opened 2020-03-17
Prevent pod deletion while ES node still contains shards

### LINKED ISSUE #2710: Prevent pod deletion if ES node still contains shards
Part of #2627 

Currently we are allowing pod and thereby Elasticsearch node deletion as soon as every shard on the tainted node has at least one copy on another node not nominated for deletion.  #2627 explains why this is not sufficient to prevent data loss. 

A first step to fix this issue is to change the data evacuation logic to only allow pod deletion once no more shards reside on a given node.

pebrc:
Removes the logic that would declare a node evacuation as completed as soon as there was another copy of every shard on the node to be deleted on at least one other node. 

Instead we now require nodes to be completely evacuated before deletion can proceed. 

This opens up the possibility for downscale operations that cannot make progress. For now we only adjust the documentation to address this and advise users to adjust their index settings. 

Fixes #2710 

See the linked issue for more context and follow-up tasks.

### DIFF
--- a/pkg/controller/elasticsearch/migration/migrate_data.go
+++ b/pkg/controller/elasticsearch/migration/migrate_data.go
@@ -24,68 +24,28 @@ const (
 	AllocationExcludeAnnotationName = "elasticsearch.k8s.elastic.co/allocation-exclude"
 )
 
-func shardIsMigrating(toMigrate client.Shard, others []client.Shard) bool {
-	if toMigrate.IsRelocating() || toMigrate.IsInitializing() {
-		return true // being migrated away or weirdly just initializing
-	}
-	if !toMigrate.IsStarted() {
-		return false // early return as we are interested only in started shards for migration purposes
-	}
-	for _, otherCopy := range others {
-		if otherCopy.IsStarted() {
-			return false // found another shard copy
-		}
-	}
-	return true // we assume other copies are initializing or there are no other copies
-}
-
-// nodeIsMigratingData is the core of IsMigratingData just with any I/O
-// removed to facilitate testing. See IsMigratingData for a high-level description.
-func nodeIsMigratingData(nodeName string, shards client.Shards, exclusions map[string]struct{}) bool {
-	// all other shards not living on the node that is about to go away mapped to their corresponding shard keys
-	othersByShard := make(map[string][]client.Shard)
+// nodeIsMigratingData returns true if there are any shards left on the node to remove.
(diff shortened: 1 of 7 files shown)


anyasabo:
Looks like a unit test `TestState_UpdateElasticsearchMigrating` needs a quick fix but otherwise LGTM.

I was wondering if we have the same concern about the rolling upgrade process (where we kill pods even if they have the only shard copy in some circumstances), but I think it's okay. Any restarts there just cause unavailability, rather than permanent data loss as in a downscale.

pkg/controller/elasticsearch/migration/migrate_data.go
```diff
@@ -24,68 +24,28 @@ const (
...
-	return true // we assume other copies are initializing or there are no other copies
-}
-
-// nodeIsMigratingData is the core of IsMigratingData just with any I/O
-// removed to facilitate testing. See IsMigratingData for a high-level description.
-func nodeIsMigratingData(nodeName string, shards client.Shards, exclusions map[string]struct{}) bool {
-	// all other shards not living on the node that is about to go away mapped to their corresponding shard keys
-	othersByShard := make(map[string][]client.Shard)
+// nodeIsMigratingData returns true if there are any shards left on the node to remove.
+func nodeIsMigratingData(nodeName string, shards client.Shards) bool {
```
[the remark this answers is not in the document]
pebrc: agreed

pkg/controller/elasticsearch/migration/migrate_data.go
```diff
@@ -24,68 +24,24 @@ const (
...
+	// all shard copies currently living on the node leaving the cluster
+	shardsToMigrate := make([]client.Shard, 0)
+	// filter shards affected by node removal
+	for _, shard := range shards {
+		if shard.NodeName == podName {
+			shardsToMigrate = append(shardsToMigrate, shard)
+		}
 	}
-	return nodeIsMigratingData(podName, shards, excludedNodes), nil
+	return len(shardsToMigrate) > 0, nil
```
barkbay: I think we don't need to maintain a slice. We can return `true` as soon as one shard is hosted on this node.

```golang
	for _, shard := range shards {
		if shard.NodeName == podName {
			return true, nil
		}
	}
	return false, nil
```

pebrc: Absolutely, sloppy me :-(

PR #2719, opened 2020-03-18
Add Traefik recipe

charith-elastic:
This recipe illustrates how to use [Traefik](https://containo.us/traefik/) to expose Elastic Stack applications to the outside world. 
It demonstrates:
- Splitting traffic by deployment type and node type
- Implementing security and resiliency using Traefik middleware

### DIFF
--- a/config/recipes/traefik/02-elastic-stack.yaml
+++ b/config/recipes/traefik/02-elastic-stack.yaml
@@ -1,0 +1,72 @@
+---
+apiVersion: elasticsearch.k8s.elastic.co/v1
+kind: Elasticsearch
+metadata:
+  name: hulk
+  labels:
+    app: hulk
+spec:
+  version: 7.6.1
+  nodeSets:
+  - name: master
+    count: 1
+    config:
+      node.master: true
+      node.data: false
+      node.ingest: false
+      node.ml: false
+      node.store.allow_mmap: false
+  - name: data
+    count: 1
+    config:
+      node.master: false
+      node.data: true
+      node.ingest: false
+      node.ml: false
+      node.store.allow_mmap: false
+  - name: ingest
+    count: 1
+    config:
+      node.master: false
+      node.data: false
+      node.ingest: true
+      node.ml: false
+      node.store.allow_mmap: false
+  - name: coordinator
+    count: 1
+    config:
+      node.master: false
+      node.data: false
+      node.ingest: false
+      node.ml: false
+      node.store.allow_mmap: false
+  - name: ml
+    count: 1
+    config:
+      node.master: false
+      node.data: false
+      node.ingest: false
+      node.ml: true
+      node.store.allow_mmap: false
+---
+apiVersion: kibana.k8s.elastic.co/v1
+kind: Kibana
+metadata:
+  name: hulk
+  labels:
+    app: hulk
+spec:
+  version: 7.6.1
+  count: 1
+  elasticsearchRef:
+    name: hulk
+---
+apiVersion: apm.k8s.elastic.co/v1
+kind: ApmServer
+metadata:
+  name: hulk
+spec:
+  version: 7.6.1
+  count: 1
+  elasticsearchRef:
+    name: hulk
--- a/config/recipes/traefik/02-elastic-stack.yaml
+++ b/config/recipes/traefik/02-elastic-stack.yaml
@@ -65,6 +65,8 @@ apiVersion: apm.k8s.elastic.co/v1
 kind: ApmServer
 metadata:
   name: hulk
+  labels:
+    app: hulk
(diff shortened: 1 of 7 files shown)


charith-elastic:
Jenkins test this please

config/recipes/traefik/README.asciidoc
```diff
@@ -0,0 +1,75 @@
+= Use Traefik to expose Elastic Stack applications
+
+This recipe illustrates how to use link:https://containo.us/traefik/[Traefik] to expose Elasticsearch, Kibana and APM Server resources to the outside world. It demonstrates how to route Elasticsearch traffic by node type and how to use Traefik middleware to add security and resiliency features.
+
+== Prerequisites
+
+- Traefik 2.1 or higher installed with the Kubernetes CRD provider. See: https://docs.traefik.io/providers/kubernetes-crd/
++
+NOTE: This recipe uses TLS to secure in-cluster communications. As the ECK generated certificates are self-signed, you will need to configure Traefik to trust those certificates. One way of doing this is to add the `--serverstransport.insecureskipverify=true` flag to the Traefik deployment.
```
sebgl: I guess having Traefik trust ECK self-signed CA makes things pretty hard to configure?

config/recipes/traefik/README.asciidoc
```diff
@@ -0,0 +1,75 @@
...
+[source,sh]
+----
+ELASTICSEARCH_PASSWORD=$(kubectl get secret hulk-es-elastic-user -o=go-template='{{ .data.elastic | base64decode}}')
+----
+
+Access Elasticsearch.
+
+[source,sh]
+----
+curl --resolve "elasticsearch.hulk:443:$INGRESS_HOST" -k -u "elastic:$ELASTICSEARCH_PASSWORD" -XGET 'https://elasticsearch.hulk/_cat/health?v'
```
sebgl: not sure I understand why we access it through `elasticsearch.hulk:443 ` instead of `$INGRESS_HOST`?

config/recipes/traefik/02-elastic-stack.yaml
```diff
@@ -0,0 +1,72 @@
...
+spec:
+  version: 7.6.1
+  count: 1
+  elasticsearchRef:
+    name: hulk
+---
+apiVersion: apm.k8s.elastic.co/v1
+kind: ApmServer
+metadata:
+  name: hulk
```
sebgl: This one does not have

```yaml
 labels:
    app: hulk
```

But I guess we don't need those labels at all?

config/recipes/traefik/03-node-type-svc.yaml
```diff
@@ -0,0 +1,33 @@
...
+  name: hulk-elasticsearch-read
+  labels:
+    app: hulk
+spec:
+  ports:
+    - name: https
+      port: 9200
+      targetPort: 9200
+  selector:
+    elasticsearch.k8s.elastic.co/cluster-name: "hulk"
```
sebgl: A comment may be nice to have here:
`# target coordinating nodes`

config/recipes/traefik/README.asciidoc
charith-elastic: Yes, it would be a complicated setup to configure it to trust specific certificates. Also, each cluster created by ECK has its own CA which makes it impractical if you have more than one cluster that you want to expose.

config/recipes/traefik/README.asciidoc
pebrc: Because traefik's IngressRoutes need a hostname to route?

charith-elastic: The proxy decides where to route requests based on the host name. If you send a request with `$INGRESS_HOST` as the host, the proxy wouldn't be able to figure out where to send that request to. The alternative to this method is to manually set the `Host` header with curl by using the `-H "Host: elasticsearch.hulk"` flag.

config/recipes/traefik/02-elastic-stack.yaml
charith-elastic: Good spot. It was a late addition and I forgot to add the label.

charith-elastic:
> I am little worried about long term maintenance. 

That's a good point. My thinking was that since I already did the work for research, someone else might find it useful. If having these in the main repo sets the implicit expectation that they are actively maintained, perhaps we could move them to a new repository. WDYT?

> One other thing that occurred to me is whether we should expose the transport service as well in these examples for remote cluster setups

I didn't include the transport service because most users don't need it and I assumed that those who do need it will be able to figure out the rest after looking at the basic example. Happy to add it if you think it would be useful.

charith-elastic:
Jenkins test this please

PR #2721, opened 2020-03-18
Check licence compatibility

anyasabo:
Checks that our dependencies' licences are compatible and approved for use as part of CI. We currently have one indirect dependency that does not have a licence. I've opened a PR against it to add a licence. If a licence is not added we will need to find a different solution.

I considered waiting to merge this, but think it would be better to have it sooner than later to ensure we do not add any incompatible dependencies in the meantime.


hack/licence-detector/whitelist/whitelist.go
```diff
@@ -0,0 +1,53 @@
...
+		if !(inWhitelist(dep.LicenceType)) {
+			return fmt.Errorf(err, dep.Name, dep.LicenceType)
+		}
+	}
+	return nil
+}
+
+func inWhitelist(depName string) bool {
+	// this is not an exhaustive list of Elastic-approved licences, but includes all the ones we use to date
+	whitelist := []string{
```
charith-elastic: Maybe pass this array to this function from `CheckWhitelist` so that it doesn't need to be initialized for every dependency?

hack/licence-detector/main.go
```diff
@@ -27,6 +28,7 @@ var (
 	noticeOutFlag       = flag.String("noticeOut", "NOTICE.txt", "Path to output the notice")
 	overridesFlag       = flag.String("overrides", "", "Path to the file containing override directives")
 	validateFlag        = flag.Bool("validate", false, "Validate results (slow)")
+	whitelistFlag       = flag.Bool("checkWhitelist", true, "Checks that all dependencies are whitelisted, returns an error if any are not")
```
charith-elastic: I wonder whether the whitelist check should just be a part of the validation check.

anyasabo: The whitelist check is much faster so it seems to make sense to run it by default on every build.

hack/licence-detector/whitelist/whitelist.go
charith-elastic: It didn't occur to me before: what if we simply make this a `map[string]struct{}`? Then we don't need the `inSlice` function and the lookup should be O(1) instead of O(n).

hack/licence-detector/main.go
charith-elastic: Then perhaps we could move the check to the licence detection stage and remove the extra loop here. It's a pretty minor optimisation + stylistic preference though. I don't feel too strongly about it.

anyasabo: Nah you're right, this doesn't really need to be behind a flag and makes it much smaller

hack/licence-detector/detector/detector.go
```diff
@@ -107,6 +107,23 @@ func doDetectLicences(licenceRegex *regexp.Regexp, classifier *licenseclassifier
 		return nil, nil
 	}
 
+	// this is not an exhaustive list of Elastic-approved licences, but includes all the ones we use to date
+	whitelist := map[string]interface{}{
```
charith-elastic: Nit: `struct{}` is zero-sized but `interface{}` is not. It's a pretty minor optimization though.

anyasabo:
The PR to licence go-gk was merged, so used a `replace` to adopt it and removed the override we were using for it.

PR #2722, opened 2020-03-18
Factorize HTTP certs reconciliation in a single function

sebgl:
Add a single common function to deal with HTTP certificates reconciliation.
It consists in reconciling 3 secrets:

- internal ca certs
- internal http certs
- public http certs (no private key)

The immediate benefit is to remove ~80 lines of somewhat duplicated code in our 4 main controllers.

This is mostly arranging code around. However, there are a few important changes to notice:

- Internal and CA secrets were created with labels matching the resource (eg. ES cluster name), but not public certs. This PR applies the same labels to all 3 secrets.
- The 3 secrets are removed automatically if TLS is disabled for Kibana, APMServer and Enterprise Search. But not for Elasticsearch, which may benefit from TLS certs hot-reload in case TLS is toggled on/off/on quickly.


pkg/controller/common/certificates/http/reconcile.go
```diff
@@ -14,48 +14,51 @@ import (
...
 
 var (
-	log = logf.Log.WithName("http")
+	log = logf.Log.WithName("http_certs")
 )
 
 // ReconcileHTTPCertificates reconciles the internal resources for the HTTP certificate.
 func ReconcileHTTPCertificates(
-	driver driver.Interface,
+	k8sClient k8s.Client,
```
charith-elastic: I know it is not strictly a part of your PR, but, this function has just too many parameters. Maybe this refactoring is a good chance to change it.  It might be easier to read if a struct was used to pass the arguments instead. Of course, the downside is that it will make it harder for the compiler to catch problems if the function signature changes in the future. WDYT?

pkg/controller/common/certificates/http/reconcile_test.go
```diff
@@ -147,24 +148,24 @@ func TestReconcileHTTPCertificates(t *testing.T) {
...
 		t.Run(tt.name, func(t *testing.T) {
 			w := watches.NewDynamicWatches()
-			testDriver := driver.TestDriver{
-				Client:  tt.args.c,
-				Watches: w,
-			}
+			//testDriver := driver.TestDriver{
+			//	Client:  tt.args.c,
+			//	Watches: w,
+			//}
```
charith-elastic: Let's delete this entirely because it's no longer used.

pkg/controller/common/certificates/reconcile.go
```diff
@@ -0,0 +1,92 @@
...
+	"github.com/elastic/cloud-on-k8s/pkg/controller/common/watches"
+	"github.com/elastic/cloud-on-k8s/pkg/utils/k8s"
+)
+
+// ReconcileCAAndHTTPCerts reconciles 3 TLS-related secrets for the given object:
+// - a Secret containing the Certificate Authority generated for the object
+// - a Secret containing the HTTP certificates and key (for internal use by the object), returned by this function
+// - a Secret containing the public-facing HTTP certificates (same as the internal one, but without the key)
+// If TLS is disabled, self-signed certificates are still reconciled, for simplicity/consistency, but not used.
+func ReconcileCAAndHTTPCerts(
```
charith-elastic: Another monster function with a lot of parameters :smile:

pkg/controller/apmserver/apmserver_controller.go
```diff
@@ -235,7 +235,18 @@ func (r *ReconcileApmServer) doReconcile(ctx context.Context, request reconcile.
 	if err != nil {
 		return reconcile.Result{}, err
 	}
-	results := apmcerts.Reconcile(ctx, r, as, []corev1.Service{*svc}, r.CACertRotation, r.CertRotation)
+	_, results := certificates.ReconcileCAAndHTTPCerts(
```
charith-elastic: Previously this was a noop because `apmcerts.Reconcile` returned immediately if the self-signed certificate was disabled. Can't we just move that check up here and avoid calling `ReconcileCAAndHTTPCerts` if TLS is disabled?

pkg/controller/kibana/driver.go
```diff
@@ -219,7 +221,18 @@ func (d *driver) Reconcile(
 		return results.WithError(err)
 	}
 
-	results.WithResults(kbcerts.Reconcile(ctx, d, *kb, []corev1.Service{*svc}, params.CACertRotation, params.CertRotation))
+	_, results = certificates.ReconcileCAAndHTTPCerts(
```
charith-elastic: Again, we can probably wrap this in an `if` clause and invoke it only if TLS is enabled.

pkg/controller/common/certificates/http/reconcile.go
sebgl: yeah it's a trade-off I often struggle with.
On one hand the struct approach makes the function signature cleaner.
On the other hand, it's just as verbose anyway (actually more since we need to declare the struct), and it's fairly easy to forget initializing some of the fields. Whereas it's not possible through function arguments. I'll still try to see if I can do something about it.

pkg/controller/apmserver/apmserver_controller.go
sebgl: we can do that, but existing secrets are not removed, which is a bug IMO.
I think it's still worth exploring though 👍

sebgl:
> I think we can still keep the old behaviour and avoid unnecessary work for Kibana and APM Server
I think you're right, it's probably not that hard. Will update the PR.

sebgl:
> I am also not fond of the certutils name (especially when referring to constants like certutils.CAFileName). Not sure what can be done though. IMO, the common/certificates package is very Java-like with too many tiny Go files (one file just for the constants, one file for the log etc.).

I agree, I had a hard time with this one 😄 
Initially we had `certificates/` and `certificates/ca`, so I just introduced `certificates/reconcile.go`. Which led me to a circular import issue since `certificates/ca` already depended on `certificates/` and `certificates/reconcile.go` now depends on `certificates/ca`. I thought there was 2 paths forward:

1. split in smaller packages so the dependencies are explicit (which I did). I think splitting `ca` and `http` in sub-packages makes sense, but agree `certutils` is not great (it's basically all certificates-related tools and constants, needed by both `ca`, `http` and other files in the codebase). I'm fine with merging eg. `certutils/fields.go`, `certutils/keys.go`, `certutils/log.go` etc. in a single file. It's pretty hard to pick a name for that one though 😄 
2. merge everything into a single pkg (`common/certificates`) - we could probably prefix the original `ca` pkg files eg. `ca_reconcile.go`. 

Do you think 2) is a cleaner approach, or should we try to find a better structure for 1)?

charith-elastic:
I am in favour of option 2. The `ca` package does not contain a lot of code so we can probably just merge them into the appropriate files within the `certificates` package. For instance, `ca/reconcile.go` can probably be merged in to the existing `certificates/reconcile.go` file. We can try to arrange the functions in such a way so that if you read a file from top to bottom, you are mostly reading the code in the order those functions are used. That should make it a lot easier to navigate even if we end up with fairly large source files as a result of this.

pebrc:
I think I would also prefer option 2 and a single `certificates` namespace

sebgl:
@charith-elastic @pebrc thanks for the feedback, let's do that.

pkg/controller/apmserver/apmserver_controller.go
sebgl: Added an option to garbage collect secrets if TLS is disabled. Used by the Kibana, APMServer and Enterprise Search controllers.

pkg/controller/common/certificates/http/reconcile.go
sebgl: Wrapped them all in a struct. I think it looks cleaner 👍

sebgl:
Summary of the recent commits:

- The 3 secrets are removed automatically if TLS is disabled for Kibana, APMServer and Enterprise Search. But not for Elasticsearch, which may benefit from TLS certs hot-reload in case TLS is toggled on/off/on quickly.
- Flatten the `certificates` pkg and merge some files together
- Wrap functions long list of args into a struct

pkg/controller/elasticsearch/certificates/transport/public_secret.go
```diff
@@ -38,7 +39,7 @@ func ReconcileTransportCertsPublicSecret(
 // PublicCertsSecretRef returns the NamespacedName for the Secret containing the publicly available transport CA.
 func PublicCertsSecretRef(es types.NamespacedName) types.NamespacedName {
 	return types.NamespacedName{
-		Name:      certificates.PublicSecretName(esv1.ESNamer, es.Name, certificates.TransportCAType),
+		Name:      esv1.ESNamer.Suffix(es.Name, "transport", "certs-public"),
```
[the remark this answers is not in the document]
sebgl: Hm not sure what happened there 😄 
Let's use a helper function from the `certificates` package here.

pkg/controller/elasticsearch/certificates/reconcile.go
```diff
@@ -54,58 +54,35 @@ func Reconcile(
...
+		K8sClient:      driver.K8sClient(),
+		DynamicWatches: driver.DynamicWatches(),
+		Object:         &es,
+		TLSOptions:     es.Spec.HTTP.TLS,
+		Namer:          esv1.ESNamer,
+		Labels:         certsLabels,
+		Services:       services,
+		CACertRotation: caRotation,
+		CertRotation:   certRotation,
+		// ES is able to hot-reload TLS certificates: let's keep secrets around even though TLS is disabled.
```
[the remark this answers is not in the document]
sebgl: That's a good question, I'm scratching my head a bit over it.

For other resources (APMServer, Kibana, Enterprise Search), we rotate Pods automatically when there's a change in TLS certs (enabled, disabled, or different content - we hash it). So we don't care much about keeping secrets around.

For ES, we rotate Pods on config change (TLS enabled/disabled), but not on certs content change. Pods rotation can take a long time (hours for very big clusters) since we wait for the cluster to be green in-between each Pod.
If we remove certs secrets, their content remain in the Pods that had them mounted, until those Pods get deleted. However I think if the Secrets then get recreated, their content does not propagate to the Pods that had the old secret mounted.
So I think the following may happen:
- user disables TLS
- we remove all certs (their old content stay there in the Pods)
- we start a 2 hours long rolling upgrade where we rotate Pods which do not have secrets mounted anymore
- halfway through users change their mind and re-enable TLS
- at this point half Pods don't have TLS secrets mounted (the ones that have been rotated), the other half have what remains of the old deleted secrets (the ones that have not been rotated)
- the first half will get recreated again (rolling upgrade) to have the certs secrets mounted
- the second half will not get recreated since their configuration is ok (TLS enabled), however their certs content will never be updated with the new ones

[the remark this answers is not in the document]
sebgl: Hmm actually after doing some test what I said above is wrong: secret that are recreated still seem to propagate to Pods that have them mounted.
Will do more tests.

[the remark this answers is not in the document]
sebgl: Just realized while playing around that we *need* the certs secrets to be available as long as Pods using them still exist, because ECK needs to be able to make https requests to the cluster, even though a rolling upgrade to disable TLS might be scheduled.
So if we remove certs secrets it must be after the rolling upgrade from non-TLS to TLS is over.
@charith-elastic how about we scope this out of this PR, and I create a dedicated low priority issue?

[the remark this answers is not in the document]
charith-elastic: Of course, I was simply curious why Elasticsearch needed a special case. Perfectly happy with deferring it to another issue.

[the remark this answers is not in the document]
sebgl: Created https://github.com/elastic/cloud-on-k8s/issues/2766.

PR #2727, opened 2020-03-19
Remove GKE 1.13 and switch to 1.14

pebrc:
https://cloud.google.com/kubernetes-engine/docs/release-notes#march_16_2020

The latest GKE release seems to have removed support for 1.13 masters which means we cannot create new 1.13 k8s clusters for testing anymore. 

This PR should removes 1.13 clusters from our k8s versions build job and replaces 1.13 with 1.14 in other jobs.

### DIFF
--- a/.ci/jobs/e2e-tests-custom-operator-image-gke.yml
+++ b/.ci/jobs/e2e-tests-custom-operator-image-gke.yml
@@ -9,8 +9,8 @@
           description: "ECK Docker image"
       - string:
           name: JKS_PARAM_K8S_VERSION
-          default: 1.13
-          description: "Kubernetes version, default is 1.13"
+          default: 1.14
+          description: "Kubernetes version, default is 1.14"
       - bool:
           name: SEND_NOTIFICATIONS
           default: true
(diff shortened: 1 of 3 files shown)


docs/reference/api-docs.asciidoc
```diff
@@ -882,8 +882,8 @@ UpdateStrategy specifies how updates to the cluster should be performed.
 Package v1beta1 contains API schema definitions for managing Enterprise Search resources.
 
 .Resource Types
-- xref:{anchor_prefix}-github-com-elastic-cloud-on-k8s-pkg-apis-enterprisesearch-v1beta1-enterprisesearch[$$EnterpriseSearch$$]
 - xref:{anchor_prefix}-github-com-elastic-cloud-on-k8s-pkg-apis-enterprisesearch-v1beta1-enterprisesearchlist[$$EnterpriseSearchList$$]
+- xref:{anchor_prefix}-github-com-elastic-cloud-on-k8s-pkg-apis-enterprisesearch-v1beta1-enterprisesearch[$$EnterpriseSearch$$]
```
pebrc: IDK about this change

pebrc: @charith-elastic is this some platform specific sorting problem maybe? I removed this change manually for now

charith-elastic: I am not sure what happened there. Will take a look.

PR #2729, opened 2020-03-19
Allow referencing K8s Secrets holding Enterprise Search configuration

sebgl:
Add a new `configRef` section in the CRD to specify a list of Kubernetes
secrets holding configuration settings.
This is to avoid cleartext credentials in the Enterprise Search
specification.

The various configuration sources are merged into a single internal
secret. User-provided configuration takes precedence. User-provided
configuration in secrets takes precedence over cleartext configuration.


pkg/controller/enterprisesearch/config.go
```diff
@@ -81,25 +83,36 @@ func ReconcileConfig(client k8s.Client, ents entsv1beta1.EnterpriseSearch) (*cor
...
-	cfg := defaultConfig(ents)
-
+// newConfig builds a single merged config from:
+// - ECK-managed default configuration
+// - association configuration (eg. ES credentials)
+// - TLS settings configuration
+// - user-provided plaintext configuration
+// - user-provided secret configuration
+// In case of duplicate settings, the last one takes precedence.
+func newConfig(c k8s.Client, dynamicWatches watches.DynamicWatches, ents entsv1beta1.EnterpriseSearch) (*settings.CanonicalConfig, error) {
```
sebgl: Will follow-up with a PR with unit tests for the config reconciliation once it's complete.
cc @barkbay

pebrc:
Would it not make sense to introduce this in all our CRDs for consistency's sake?

sebgl:
> Would it not make sense to introduce this in all our CRDs for consistency's sake?

We had a discussion and decided we may eventually do it if it makes sense in the future, but not now.

pkg/controller/enterprisesearch/config.go
```diff
@@ -135,3 +148,46 @@ func tlsConfig(ents entsv1beta1.EnterpriseSearch) *settings.CanonicalConfig {
...
+		if err := c.Get(types.NamespacedName{Namespace: ents.Namespace, Name: secretName}, &secret); err != nil {
+			// the secret may not exist (yet) in the cache
+			// it may contain important settings such as encryption keys, that we don't want to generate ourselves
+			// let's explicitly error out
+			return nil, err
+		}
+		if data, exists := secret.Data[ConfigFilename]; exists {
+			parsed, err := settings.ParseConfig(data)
+			if err != nil {
+				log.Error(err, "unable to parse configuration from secret",
```
pebrc: Should we also create an event here to raise visibility in case of misconfiguration?

sebgl: Good idea!

config/crds/all-crds.yaml
```diff
@@ -1542,6 +1542,20 @@ spec:
             config:
               description: Config holds the Enterprise Search configuration.
               type: object
+            configRef:
+              description: 'ConfigRef contains references to Kubernetes Secrets holding
```
barkbay: Maybe we should explain that it is complementary to the `config` field and explain which one takes precedence.

pkg/controller/enterprisesearch/config.go
```diff
@@ -148,6 +158,50 @@ func getExistingConfig(client k8s.Client, ents entsv1beta1.EnterpriseSearch) (*s
...
+			// the secret may not exist (yet) in the cache
+			// it may contain important settings such as encryption keys, that we don't want to generate ourselves
+			// let's explicitly error out
+			return nil, err
+		}
+		if data, exists := secret.Data[ConfigFilename]; exists {
+			parsed, err := settings.ParseConfig(data)
+			if err != nil {
+				msg := "unable to parse configuration from secret"
+				log.Error(err, "msg", "namespace", ents.Namespace, "ents_name", ents.Name, "secret_name", secretName)
```
barkbay:
```suggestion
				log.Error(err, msg, "namespace", ents.Namespace, "ents_name", ents.Name, "secret_name", secretName)
```

pkg/controller/enterprisesearch/config.go
```diff
@@ -148,6 +158,50 @@ func getExistingConfig(client k8s.Client, ents entsv1beta1.EnterpriseSearch) (*s
 	return cfg, nil
 }
 
+// configRefWatchName returns the name of the watch registered on
```
barkbay: _on Kubernetes Secrets referenced in configRef_ ?

pkg/controller/enterprisesearch/config.go
```diff
@@ -148,6 +158,50 @@ func getExistingConfig(client k8s.Client, ents entsv1beta1.EnterpriseSearch) (*s
...
+func configRefWatchName(ents types.NamespacedName) string {
+	return fmt.Sprintf("%s-%s-configref", ents.Namespace, ents.Name)
+}
+
+// parseConfigRef builds a single merged CanonicalConfig from the secrets referenced in configReg,
```
barkbay:
```suggestion
// parseConfigRef builds a single merged CanonicalConfig from the secrets referenced in configRef,
```

sebgl:
Jenkins test this please

PR #2730, opened 2020-03-19
Switch to /api/status for Kibana e2e tests

### LINKED ISSUE #2717: Kibana e2e test broken due to login page changes as of 7.7.0
```
utils.go:83
            	            	initial Kibana UI state forbids login which indicates an error in the ES/Kibana setup
            	            	github.com/elastic/cloud-on-k8s/test/e2e/test/kibana.(*kbChecks).CheckKbLoginHealthy.func1
            	            		/go/src/github.com/elastic/cloud-on-k8s/test/e2e/test/kibana/checks_kb.go:73
            	            	github.com/elastic/cloud-on-k8s/test/e2e/test.UntilSuccess.func1.1
            	            		/go/src/github.com/elastic/cloud-on-k8s/test/e2e/test/utils.go:80
            	            	github.com/elastic/cloud-on-k8s/pkg/utils/retry.UntilSuccess.func2
            	            		/go/src/github.com/elastic/cloud-on-k8s/pkg/utils/retry/retry.go:41
            	            	runtime.goexit
            	            		/usr/local/go/src/runtime/asm_amd64.s:1357
            	Test:       	TestCrossNSAssociation/Kibana_should_be_able_to_connect_to_Elasticsearch
```

This is due to implementation changes in the Kibana login page. We are currently relying on an implementation detail to check whether login was available. 

https://github.com/elastic/cloud-on-k8s/blob/da1fb962faa7b84b58038a716f8e6c8b137f80c8/test/e2e/test/kibana/checks_kb.go#L72-L73

What we want to check whether Kibana came up successfully and has connected to Elasticsearch. @liza-mae maybe there is a more robust way of doing that?

pebrc:
Instead of checking for the presence of a string in the login page make an authenticated request to the Kibana status API. 

This should be slightly more resilient against changes between different stack versions.
The expected behaviour is as follows:
> -  api/status will respond with 503 until the server is ready and able to talk to Elasticsearch and run migrations
> - If Kibana loses communication with ES and the status.allowAnonymous is not set to true then you will get 401 from Kibana on this endpoint
> - If Kibana loses communication with ES and the status.allowAnonymous is set to true then you will get 200 from Kibana on this endpoint with a status.overall.state property set to red (edited) 

In practice I have seen it simply timing out when Elasticsearch is not available which is a bit worse than the current behaviour on the login page, but I guess the API check is less brittle.

Fixes #2717

### DIFF
--- a/test/e2e/kb/telemetry_test.go
+++ b/test/e2e/kb/telemetry_test.go
@@ -30,7 +30,7 @@ func TestTelemetry(t *testing.T) {
 				Test: func(t *testing.T) {
 					uri := "/api/telemetry/v1/clusters/_stats"
 					payload := `{"timeRange":{"min":"0","max":"0"}}`
-					body, err := kibana.DoKibanaReq(k, kbBuilder, "POST", uri, []byte(payload))
+					body, err := kibana.DoKibanaReq(k, kbBuilder.Kibana, "POST", uri, []byte(payload))
 					require.NoError(t, err)
 
 					var stats ClusterStats
--- a/test/e2e/test/kibana/checks_kb.go
+++ b/test/e2e/test/kibana/checks_kb.go
@@ -5,21 +5,23 @@
 package kibana
 
 import (
+	"encoding/json"
 	"fmt"
-	"io/ioutil"
-	"net/http"
-	"strings"
-	"testing"
 
 	kbv1 "github.com/elastic/cloud-on-k8s/pkg/apis/kibana/v1"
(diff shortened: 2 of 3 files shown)


pebrc:
Jenkins test this please

pebrc:
> Would it help if we set elasticsearch.requestTimeout in Kibana to a smaller value (default is 30 seconds) and set a request timeout on the Kibana client that is slightly larger than that value? 

I tried removing the Elasticsearch pods and curl'ing the Kibana API and timed out only after 2 minutes, but I did not investigate further.

PR #2740, opened 2020-03-20
Fix transport certificates reconciliation

sebgl:
Commit 174e09b26b4feeed88104a75c5f56d82e0807b7f introduced a regression
by using the generic secret reconciler when we first create the
transport certs secret.

This secret reconciliation is rather special, since it ensures the
secret exists (before Pods are created), but does not care about its
data (since we don't have Pods IPs yet).
If an existing secret has some data already, we must make sure we don't
clear it. Which is what the generic reconciler would do.

This commit basically restores the code as it existed before the
refactoring in 174e09b26b4feeed88104a75c5f56d82e0807b7f, and adds a unit
test to catch the regression.

### DIFF
--- a/pkg/controller/elasticsearch/certificates/transport/reconcile.go
+++ b/pkg/controller/elasticsearch/certificates/transport/reconcile.go
@@ -18,6 +18,8 @@ import (
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common/reconciler"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/elasticsearch/label"
 	"github.com/elastic/cloud-on-k8s/pkg/utils/k8s"
+	"github.com/elastic/cloud-on-k8s/pkg/utils/maps"
+
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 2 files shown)


pkg/controller/elasticsearch/certificates/transport/reconcile.go
```diff
@@ -120,7 +122,7 @@ func ReconcileTransportCertificatesSecrets(
 func ensureTransportCertificatesSecretExists(
 	c k8s.Client,
 	es esv1.Elasticsearch,
-) (corev1.Secret, error) {
+) (*corev1.Secret, error) {
```
sebgl: I'm not very fond of returning pointers when not required (it introduces doubt: should we check for `nil`?), but for the sake of "just fixing the regression" I re-introduced the code as it was before https://github.com/elastic/cloud-on-k8s/commit/174e09b26b4feeed88104a75c5f56d82e0807b7f.

pkg/controller/elasticsearch/certificates/transport/reconcile.go
```diff
@@ -104,7 +106,7 @@ func ReconcileTransportCertificatesSecrets(
 	}
 
 	if !reflect.DeepEqual(secret, currentTransportCertificatesSecret) {
-		if err := c.Update(&secret); err != nil {
+		if err := c.Update(secret); err != nil {
```
sebgl: I think this is the place where we could potentially benefit from the generic secret reconciler, but it requires a bit more work. I'll open a follow-up issue.

sebgl:
Jenkins test this please

sebgl:
Jenkins test this please

PR #2742, opened 2020-03-20
Surface EULA validation in annotation for trials

pebrc:
* removes validation package
* inlines EULA validation into trial controller
* adds result of validation as trial secret annotation (as we do it for other problems during trial activation)
* adds additional tests for the trial controller


pkg/controller/license/trial/trial_controller_test.go
```diff
@@ -0,0 +1,208 @@
...
+					Name:      licensing.TrialStatusSecretKey,
+				}, &sec))
+				require.NoError(t, c.Get(trialSecretNsn, &sec))
+				_, lic, err := licensing.TrialLicense(c, trialSecretNsn)
+				require.NoError(t, err)
+				require.NoError(t, lic.IsMissingFields())
+			},
+		},
+		{
+			name: "valid trial after operator restart",
```
barkbay: I'm not sure I understand this test, the name and the assertion seems a little bit confusing to me. IIUC the license check is failing here because there is no valid license in `trialLicenseBytes` , not because it is "restarted" ?

pebrc: Yes I am not happy with this either. Let me think if there is a better way to test this.

pebrc: I re-did the test, with a valid simulated trial.

PR #2746, opened 2020-03-20
License check: update remote cluster logs and events

### LINKED ISSUE #2738: Remote cluster controller logs about missing enterprise license even though there's no remote configuration
When enterprise license is not enabled and there is an existing Elasticsearch resource with no remote cluster configuration, the remote cluster controller spams logs with:

```
2020-03-20T10:47:13.259+0100    INFO    remotecluster-remoteca  Remote cluster controller is an enterprise feature. Enterprise features are disabled    {"service.version": "1.1.0-82bdfe64", "namespace": "default", "es_name": "elasticsearch-sample"}
```

We probably want to only log this if there's a configured remote cluster.

barkbay:
Fixes #2738 

Note that with this PR remote clusters can be removed even if the license is no more valid. It allows the user to gently clean up the resources even if the trial has expired for example.

### DIFF
--- a/pkg/controller/elasticsearch/remotecluster/remoteca/controller.go
+++ b/pkg/controller/elasticsearch/remotecluster/remoteca/controller.go
@@ -90,18 +90,6 @@ func (r *ReconcileRemoteCa) Reconcile(request reconcile.Request) (reconcile.Resu
 		return common.PauseRequeue, nil
 	}
 
-	enabled, err := r.licenseChecker.EnterpriseFeaturesEnabled()
-	if err != nil {
-		return defaultRequeue, err
-	}
-	if !enabled {
-		log.Info(
-			"Remote cluster controller is an enterprise feature. Enterprise features are disabled",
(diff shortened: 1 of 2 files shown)


pkg/controller/elasticsearch/remotecluster/remoteca/controller.go
```diff
@@ -135,6 +123,18 @@ func doReconcile(
 		return reconcile.Result{}, err
 	}
 
+	enabled, err := r.licenseChecker.EnterpriseFeaturesEnabled()
+	if err != nil {
+		return defaultRequeue, err
+	}
+	if !enabled && len(expectedRemoteClusters) > 0 {
+		log.V(1).Info(
```
charith-elastic: Is it intentional to log this at debug level?

barkbay: Yes, we already have the same log during the API call

PR #2750, opened 2020-03-20
Add retry library

anyasabo:
Adds a shared library that overrides the default checkout behavior to include a sleep+retry, which should help with our checkout/github connection failures. Related PR is needed to add this library to the devops-ci jenkins


anyasabo:
jenkins test this please

anyasabo:
jenkins test this please

anyasabo:
The library has been added to the jenkins server so we should be good to use this. It will hopefully make our CI runs a little more stable.

.ci/pipelines/build.Jenkinsfile
```diff
@@ -1,3 +1,5 @@
+@Library('apm@current') _
```
sebgl: can we add a comment to explain what this does?

PR #2757, opened 2020-03-23
Toggle read-only mode on Enterprise Search version upgrade

sebgl:
Handle Enterprise Search version upgrades by toggling read-only mode.

The logic is the following:
* if a version upgrade is detected (one of the actual Pods runs a version older than the expected one)
    * perform an API call to enable read-only mode
    * store the read-only mode state in an annotation on the Enterprise Search resource so we don't do the API call over and over again
* if all Pods are running the expected version
    * if the read-only mode annotation is set, perform an API call to disable read-only mode

In practice, the following can be observed during a version upgrade:

* 3 Pods running in the initial version
* Apply a manifest to upgrade the version
* ECK enables read-only mode automatically (confirmed in Enterprise Search logs)
* Pods get slowly rotated, one by one
* The first Pod starting in the new version does a data migration (confirmed in Enterprise Search logs)
* Once no more Pods of the old version are running, ECK disables read-only mode (confirmed in Enterprise Search logs)

Since we need to connect to the Enterprise Search API from the controller, this PR extracts the HTTP client we used for Elasticsearch in a common place.


sebgl:
Jenkins test this please.

pkg/controller/enterprisesearch/version_upgrade.go
```diff
@@ -0,0 +1,240 @@
...
+}
+
+// isVersionUpgrade returns true if at least one Pod runs a version prior to the one
+// specified in the EnterpriseSearch resource.
+func (r *VersionUpgrade) isVersionUpgrade() (bool, error) {
+	expectedVersion, err := version.Parse(r.ents.Spec.Version)
+	if err != nil {
+		return false, err
+	}
+	pods, err := r.getActualPods()
```
barkbay, line 199: I'm wondering if we may have a race condition between the actual `ReplicaSets` and ECK getting the list of the Pods. Is it possible to be in a situation where there is no Pods but we have a `ReplicasSet` with a desired count of Pods greater than 0  that is about to create them ? For example something like this:

```
> k get rs
NAME                                      DESIRED   CURRENT   READY   AGE
entsearch-sample-ents-server-7f5c4b665b   3         0         0       17m
```

No Pods yet, but one second later 3 Pods are created with an older version:

```
> k get rs
NAME                                      DESIRED   CURRENT   READY   AGE
entsearch-sample-ents-server-7f5c4b665b   3         3         0       17m
```

pkg/controller/enterprisesearch/version_upgrade.go
```diff
@@ -0,0 +1,240 @@
...
+	"github.com/elastic/cloud-on-k8s/pkg/utils/k8s"
+	"github.com/elastic/cloud-on-k8s/pkg/utils/net"
+	"github.com/elastic/cloud-on-k8s/pkg/utils/stringsutil"
+)
+
+const (
+	// ReadOnlyModeAnnotationName stores "true" when read-only mode is enabled.
+	ReadOnlyModeAnnotationName = "enterprisesearch.k8s.elastic.co/read-only"
+	// ReadOnlyModeAPIPath is the HTTP path of the read-only mode API.
+	ReadOnlyModeAPIPath = "/api/ent/v1/internal/read_only_mode"
```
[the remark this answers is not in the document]
sebgl: Yeah I also realized that in this PR!
Let's create a follow up issue to use `ent` as short name everywhere.

pkg/controller/enterprisesearch/version_upgrade.go
```diff
@@ -0,0 +1,240 @@
...
+	isUpgrade, err := r.isVersionUpgrade()
+	if err != nil {
+		return err
+	}
+
+	if isUpgrade && !r.ents.AssociationConf().AuthIsConfigured() {
+		// A version upgrade is scheduled, but we don't know how to reach the Enterprise Search API
+		// since we don't have any Elasticsearch user available.
+		// Move on with the upgrade: this will cause the Pod in the new version to crash at startup with explicit logs.
+		log.Info("Detected version upgrade with no association to Elasticsearch, "+
```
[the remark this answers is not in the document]
sebgl: 👍

pkg/controller/enterprisesearch/version_upgrade.go
sebgl: Look like an edge case but still possible 👍 

Also we need at least one Pod running to do the API call.
Maybe we can change this function to check the version in the Deployment resource (instead of the Pods), and if the version is lower, requeue until at least one Pod is running?
That would not work well for cases where no Pod ever ran (invalid manifest or Pods pending forever), and the user decides to change the version in the yaml manifest. Maybe we can consider this an edge case since in this scenario a simple way out is to remove then recreate the resource?

barkbay: > Also we need at least one Pod running to do the API call.

It didn't realize that if the first Pod never come up (e.g. too much resources requested)  it is not possible to upgrade until the issue is fixed. It might not be obvious for the user how to fix this situation.
But I think it would be the same with a self-managed deployment, so we can keep this as a possible improvement for later.

> Maybe we can change this function to check the version in the Deployment resource (instead of the Pods) first, and if the version is lower, requeue until at least one Pod is running? Then also check the existing Pods to not be done with the version upgrade until no more Pods run in the old version.

👍

pkg/controller/enterprisesearch/version_upgrade.go
```diff
@@ -193,13 +225,29 @@ func (r *VersionUpgrade) readOnlyModeRequest(enabled bool) (*http.Request, error
...
+
+	podVersion, err := version.FromLabels(deployment.Spec.Template.Labels, VersionLabelName)
 	if err != nil {
 		return false, err
 	}
+	return expectedVersion.IsAfter(*podVersion), nil
+}
+
+// priorVersionStillRunning returns true if at least one Pod runs a version prior to the expected one.
+func (r *VersionUpgrade) priorVersionStillRunning(expectedVersion version.Version) (bool, error) {
```
barkbay, line 250: Maybe `isPriorVersionStillRunning` to make it clear that it returns a boolean ?

pkg/controller/enterprisesearch/version_upgrade_test.go
```diff
@@ -312,3 +342,43 @@ func TestVersionUpgrade_readOnlyModeRequest(t *testing.T) {
 		})
 	}
 }
+
+func TestVersionUpgrade_isVersionUpgrade1(t *testing.T) {
```
barkbay, line 346: The existing `TestVersionUpgrade_isVersionUpgrade` should be renamed `TestVersionUpgrade_isPriorVersionStillRunning`, this would allow to rename this one `TestVersionUpgrade_isVersionUpgrade`

pkg/controller/enterprisesearch/version_upgrade.go
```diff
@@ -66,9 +74,33 @@ func (r *VersionUpgrade) Handle(ctx context.Context) error {
...
+			log.Info(msg, "namespace", r.ents.Namespace, "ents_name", r.ents.Name)
+			r.recorder.Event(&r.ents, corev1.EventTypeWarning, events.EventReasonDelayed, msg)
+			// surface this as an error, since rather unexpected, and abort reconciliation
+			return errors.New(msg)
+		}
+		// enable read-only mode before moving on with the deployment upgrade
 		return r.enableReadOnlyMode(ctx)
 	}
+
+	oldVersionStillRunning, err := r.priorVersionStillRunning(*expectedVersion)
```
barkbay, line 95: nit: can be simplified with:

```golang
	if oldVersionStillRunning, err := r.priorVersionStillRunning(*expectedVersion); oldVersionStillRunning || err != nil {
		return err
	}
```

sebgl:
Thanks for the review @barkbay!

PR #2758, opened 2020-03-23
Add log adapter for automaxprocs

pebrc:
Looks like automaxprocs expects a `printf` style interface for logging while our logger expects balanced key-value pairs. 

This adds a simple adapter function to translate between the two

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -183,6 +183,8 @@ func init() {
 func execute() {
 	// update GOMAXPROCS to container cpu limit if necessary
 	_, err := maxprocs.Set(maxprocs.Logger(func(s string, i ...interface{}) {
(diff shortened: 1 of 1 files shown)


cmd/manager/main.go
```diff
@@ -182,7 +182,9 @@ func init() {
 
 func execute() {
 	// update GOMAXPROCS to container cpu limit if necessary
-	_, err := maxprocs.Set(maxprocs.Logger(log.Info))
+	_, err := maxprocs.Set(maxprocs.Logger(func(s string, i ...interface{}) {
```
anyasabo: Maybe also include
```
	// maxprocs needs an sprintf format string with args, but our logger needs a string with optional key value pairs,
	// so we need to do this translation
```
because it's not clear just from the signatures why we need it?

pebrc: Will add

PR #2764, opened 2020-03-24
Fix remote cluster e2e test with 6.x

### LINKED ISSUE #2761: e2e: remote cluster test on 6.x needs soft deletes enabled
CCR on 6.x Elasticsearch versions needs explicit enablement of soft-deletes
```
400 Bad Request: leader index [data-integrity-check] does not have soft deletes enabled. soft deletes must be enabled when the index is created by setting index.soft_deletes.enabled to true
```

See https://www.elastic.co/guide/en/elasticsearch/reference/6.8/ccr-requirements.html

cc @barkbay

barkbay:
Fixes #2761 by creating test index with `soft_deletes.enabled` when testing on Elastic stack 6.x

Tested on 6.8.5 and 7.1.1

### DIFF
--- a/test/e2e/es/remote_cluster_test.go
+++ b/test/e2e/es/remote_cluster_test.go
@@ -10,7 +10,6 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
-	"strings"
 	"testing"
 
 	"github.com/elastic/cloud-on-k8s/pkg/controller/elasticsearch/client"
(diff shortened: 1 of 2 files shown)


test/e2e/es/remote_cluster_test.go
```diff
@@ -69,7 +70,11 @@ func TestRemoteCluster(t *testing.T) {
 			test.Step{
 				Name: "Add some data to the first cluster",
 				Test: func(t *testing.T) {
-					require.NoError(t, elasticsearch.NewDataIntegrityCheck(k, es1Builder).Init())
+					dataIntegrityCheck := elasticsearch.NewDataIntegrityCheck(k, es1Builder)
+					if strings.HasPrefix(es1Builder.Elasticsearch.Spec.Version, "6") {
+						dataIntegrityCheck.WithSoftDeletesEnabled(true)
```
pebrc: I wonder whether we could just always explicitly enable soft deletes and avoid the conditional branch here? We could add a comment explaining that we are doing it because of older versions of Elasticsearch that did not have it enabled by default.

Issue #2415, opened 2020-01-13
Stack monitoring with Metricbeat


pebrc:
Related to #2051 

Three possible approaches to collecting monitoring data from Elastic stack containers come to mind
* via a centrally managed Metricbeat instance 
* via a process running inside the stack (e.g. Elasticsearch) container
* via a sidecar container next to the stack (e.g. Elasticsearch) container

## Central management
* Basically extending on the `DaemonSet` advertised in https://www.elastic.co/guide/en/beats/metricbeat/current/running-on-kubernetes.html
* Most resource effective (one Metricbeat process per k8s node)
* Difficult to configure: 
    * we need to add configuration stanzas to the Metricbeat config files as pods are scheduled on a specific k8s node and remove them again if they get scheduled somewhere else
    * unclear ownership of the configuration files 
    * multiple processes writing concurrently to same `ConfigMap` to add or remove configuration files if we share a `ConfigMap` between e.g. Elasticsearch clusters
    * or alternatively a potentially high number of volume mounts to manage on the Metricbeat pod if we don't share a k8s resource but every stack pod needs mount its own monitoring configuraiton into the Metricbeat pod with a restart of the Metricbeat pod at each such occasion
* Only one output Elasticsearch cluster possible

The disadvantages seem to outweigh the benefits of saving resources.

## In-container Metricbeat process
While this feels like a container anti-pattern at first, the stack seems to be going into that direction starting with version 8.0.0 see https://github.com/elastic/elasticsearch/issues/49399 and https://github.com/elastic/kibana/issues/51224
* Very hacky to do this before the official images support it (we had a similar approach with the shortlived process manager in one of the ECK alpha versions) 
* Would require introduction of a process manager to handle stray child processes
* I would assume that this could be disruptive to existing deployments if we introduce this in a future version of ECK because sizing and resource requests/limits would have been done without this extra process in mind (we could work around this of course by bumping the resources on rollout, but that in itself is also problematic as we might cause rescheduling of nodes etc)

Unless we have support for this in the official Docker containers we should avoid this route.

## Sidecar Metricbeat process

* would the be the most k8s-native way of a running a Metricbeat process alongside each Elastic stack container
* localhost network access from Metricbeat to the stack process (e.g. Elasticsearch)
* Metricbeat shares the same 'fate' as the stack process and follows it around as it is rescheduled etc
* possible for users to configure this today with hand-crafted sidecar 
[...]

PR #2777, opened 2020-03-26
Restrict APM Server user role

### LINKED ISSUE #2661: APM / Elasticsearch association: restrict the user privileges
We are currently binding the ApmServer user to the `superuser` role. We should create a custom role with less privileges.

IIUC the [documentation](https://www.elastic.co/guide/en/apm/server/7.6/feature-roles.html) the default user should have the following roles:

* [setup role](https://www.elastic.co/guide/en/apm/server/7.6/feature-roles.html#privileges-to-setup-beats) for setting up index templates and other dependencies
* [writer role](https://www.elastic.co/guide/en/apm/server/7.6/feature-roles.html#privileges-to-publish-events) for publishing events collected by APM Server

barkbay:
Fixes #2661 

This PR:
 * Adds a role called `default_apm_user_role` _(name can be discussed ofc)_ to allow APM Server instances to setup and manage `apm-*` indices:
```yaml
default_apm_user_role:
      cluster:
      - monitor
      - manage_ilm
      indices:
      - names:
        - apm-*
        privileges:
        - manage
        - create_doc
        - create_index

```
* Adds the APM users to the role described above and also to the built-in roles `kibana_user` and `ingest_admin`

I did some tests and it seems to work fine but I would be glad to have a quick review of the role by @simitt or @graphaelli. Thanks 🙇

### DIFF
--- a/pkg/controller/apmserverelasticsearchassociation/apmserverelasticsearchassociation_controller.go
+++ b/pkg/controller/apmserverelasticsearchassociation/apmserverelasticsearchassociation_controller.go
@@ -7,6 +7,7 @@ package apmserverelasticsearchassociation
 import (
 	"context"
 	"reflect"
+	"strings"
 	"time"
 
 	"go.elastic.co/apm"
@@ -35,6 +36,7 @@ import (
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common/tracing"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common/watches"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/elasticsearch/services"
+	"github.com/elastic/cloud-on-k8s/pkg/controller/elasticsearch/user"
 	"github.com/elastic/cloud-on-k8s/pkg/utils/k8s"
 	"github.com/elastic/cloud-on-k8s/pkg/utils/maps"
 	"github.com/elastic/cloud-on-k8s/pkg/utils/rbac"
@@ -49,6 +51,13 @@ const (
 var (
 	log            = logf.Log.WithName(name)
 	defaultRequeue = reconcile.Result{Requeue: true, RequeueAfter: 10 * time.Second}
+
+	// apmDefaultRoles are the default roles used by APM Server instances
+	apmDefaultRoles = strings.Join([]string{
+		user.ApmDefaultUserRole, // Retrieve cluster details (e.g. version) and manage apm-* indices
+		"kibana_user",           // Load dependencies, such as example dashboards, if available, into Kibana
+		"ingest_admin",          // Set up index templates
+	}, ",")
 )
 
 // Add creates a new ApmServerElasticsearchAssociation Controller and adds it to the Manager with default RBAC. The Manager will set fields on the Controller
@@ -291,7 +300,7 @@ func (r *ReconcileApmServerElasticsearchAssociation) reconcileInternal(ctx conte
 		r.Client,
 		apmServer,
 		associationLabels(apmServer),
-		"superuser",
+		apmDefaultRoles,
 		apmUserSuffix,
 		es,
 	); err != nil { // TODO distinguish conflicts and non-recoverable errors here
--- a/pkg/controller/apmserverelasticsearchassociation/apmserverelasticsearchassociation_controller.go
+++ b/pkg/controller/apmserverelasticsearchassociation/apmserverelasticsearchassociation_controller.go
@@ -61,6 +61,7 @@ func getRoles(v version.Version) string {
 		return strings.Join([]string{
 			user.ApmUserRoleV75, // Retrieve cluster details (e.g. version) and manage apm-* indices
 			"ingest_admin",      // Set up index templates
+			"apm_system",        // To collect metrics about APM Server
 		}, ",")
 	}
 
@@ -69,11 +70,15 @@ func getRoles(v version.Version) string {
 		return strings.Join([]string{
 			user.ApmUserRoleV7, // Retrieve cluster details (e.g. version) and manage apm-* indices
 			"ingest_admin",     // Set up index templates
+			"apm_system",       // To collect metrics about APM Server
 		}, ",")
 	}
 
 	// 6.8
-	return user.ApmUserRoleV6
+	return strings.Join([]string{
+		user.ApmUserRoleV6, // Retrieve cluster details (e.g. version) and manage apm-* indices
+		"apm_system",       // To collect metrics about APM Server
+	}, ",")
 }
 
 // Add creates a new ApmServerElasticsearchAssociation Controller and adds it to the Manager with default RBAC. The Manager will set fields on the Controller
--- a/pkg/controller/apmserverelasticsearchassociation/apmserverelasticsearchassociation_controller.go
+++ b/pkg/controller/apmserverelasticsearchassociation/apmserverelasticsearchassociation_controller.go
@@ -34,6 +34,7 @@ import (
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common/operator"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common/reconciler"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common/tracing"
+	"github.com/elastic/cloud-on-k8s/pkg/controller/common/version"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common/watches"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/elasticsearch/services"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/elasticsearch/user"
@@ -51,14 +52,29 @@ const (
 var (
 	log            = logf.Log.WithName(name)
 	defaultRequeue = reconcile.Result{Requeue: true, RequeueAfter: 10 * time.Second}
-
-	// apmDefaultRoles are the default roles used by APM Server instances
-	apmDefaultRoles = strings.Join([]string{
-		user.ApmDefaultUserRole, // Retrieve cluster details (e.g. version) and manage apm-* indices
-		"apm_user",              // Load dependencies, such as example dashboards, if available, into Kibana
-		"ingest_admin",          // Set up index templates
-	}, ",")
 )
+
+// getRoles returns for a given version of the APM Server the set of required roles.
+func getRoles(v version.Version) string {
+	// 7.5.x and above
+	if v.IsSameOrAfter(version.From(7, 5, 0)) {
+		return strings.Join([]string{
+			user.ApmUserRoleV75, // Retrieve cluster details (e.g. version) and manage apm-* indices
+			"ingest_admin",      // Set up index templates
+		}, ",")
+	}
+
+	// 7.1.x to 7.4.x
+	if v.IsSameOrAfter(version.From(7, 1, 0)) {
+		return strings.Join([]string{
+			user.ApmUserRoleV7, // Retrieve cluster details (e.g. version) and manage apm-* indices
+			"ingest_admin",     // Set up index templates
+		}, ",")
+	}
+
+	// 6.8
+	return user.ApmUserRoleV6
+}
 
 // Add creates a new ApmServerElasticsearchAssociation Controller and adds it to the Manager with default RBAC. The Manager will set fields on the Controller
 // and Start it when the Manager is Started.
@@ -300,7 +316,7 @@ func (r *ReconcileApmServerElasticsearchAssociation) reconcileInternal(ctx conte
 		r.Client,
 		apmServer,
 		associationLabels(apmServer),
-		apmDefaultRoles,
+		getRoles(version.MustParse(apmServer.Spec.Version)),
 		apmUserSuffix,
 		es,
 	); err != nil { // TODO distinguish conflicts and non-recoverable errors here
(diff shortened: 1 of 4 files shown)


sebgl:
@simitt @graphaelli should this role be builtin Elasticsearch, similar to the kibana user role?

simitt:
@elastic/apm-server for visibility

pkg/controller/elasticsearch/user/roles.go
```diff
@@ -18,12 +18,23 @@ const (
...
 	// PredefinedRoles to create for internal needs.
 	PredefinedRoles = RolesFileContent{
 		ProbeUserRole: esclient.Role{Cluster: []string{"monitor"}},
+		ApmDefaultUserRole: esclient.Role{
+			Cluster: []string{"monitor", "manage_ilm"},
+			Indices: []esclient.IndexRole{
+				{
+					Names:      []string{"apm-*"},
+					Privileges: []string{"manage", "create_doc", "create_index"},
+				},
```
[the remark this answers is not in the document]
barkbay: `default_apm_user_role` is only meant to be used for the username set in the `output.elasticsearch` configuration:

```yaml
apm-server:
  host: :8200
output:
  elasticsearch: # here
    hosts:
    - https://es-apm-sample-es-http.flux.svc:9200
    password: 2zgfoobarn
    username: default-apm-sample-apm-user
```

I would expect the end user to use a different Elasticsearch user to configure monitoring in `monitoring.elasticsearch`. In that case I think we don't need the `apm_system` role or the privileges described in [monitoring role](https://www.elastic.co/guide/en/apm/server/7.6/feature-roles.html#privileges-to-publish-monitoring) ?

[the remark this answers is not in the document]
simitt: As soon as monitoring gets enabled by setting `monitoring.enabled: true`, the elasticsearch configuration from the output section is reused for monitoring by default. That's why I assumed the role should also include monitoring rights. 
But it is a fair decision to only build the roles with the minimal required permissions, excluding monitoring. Please just disregard my first comment then.

pkg/controller/apmserverelasticsearchassociation/apmserverelasticsearchassociation_controller.go
```diff
@@ -49,6 +51,13 @@ const (
 var (
 	log            = logf.Log.WithName(name)
 	defaultRequeue = reconcile.Result{Requeue: true, RequeueAfter: 10 * time.Second}
+
+	// apmDefaultRoles are the default roles used by APM Server instances
+	apmDefaultRoles = strings.Join([]string{
+		user.ApmDefaultUserRole, // Retrieve cluster details (e.g. version) and manage apm-* indices
+		"kibana_user",           // Load dependencies, such as example dashboards, if available, into Kibana
```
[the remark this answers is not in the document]
simitt: The privileges required for agent configuration management I was referring to above, need to be set for the user talking to Kibana (`apm-server.kibana`), but not for the `elasticsearch` user. 
Based on the explanation in https://github.com/elastic/cloud-on-k8s/pull/2777/files#r400078619 the `apmDefaultRoles` will only be used for writing to ES. In this case no Kibana privileges are required for any APM Server >= 7.0, as dashboards are no longer shipped with the APM Server.

pkg/controller/elasticsearch/user/roles.go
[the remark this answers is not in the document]
anyasabo: To add more color on why we would exclude monitoring here:

We do not have any explicit support for self-monitoring right now, anyone enabling it should also be able to create their own user I think. We have a [separate issue](https://github.com/elastic/cloud-on-k8s/issues/2415) for stack monitoring. We haven't done much investigation into APM specifically, but so far we are planning to use metricbeat and it seems likely we would continue that for APM

pkg/controller/elasticsearch/client/client.go
```diff
@@ -33,14 +33,16 @@ type BasicAuth struct {
...
 type Role struct {
-	Cluster []string `json:"cluster,omitempty"`
-	/*Indices []struct {
-		Names      []string `json:"names,omitempty"`
-		Privileges []string `json:",omitempty"`
-	} `json:"indices,omitempty"`
-	Applications []struct {
+	Cluster []string    `json:"cluster,omitempty"`
+	Indices []IndexRole `json:"indices,omitempty"`
+	/*Applications []struct {
```
anyasabo: should we just remove all of these? it's not clear to me from the code or the last PR that touched them why they're here

pkg/controller/apmserverelasticsearchassociation/apmserverelasticsearchassociation_controller.go
```diff
@@ -49,6 +51,13 @@ const (
...
+		"ingest_admin",          // Set up index templates
```
anyasabo: Can you help me understand the difference between `ApmDefaultUserRole` and this? It seems like we could put `ingest_admin` in the cluster role def there along with `monitor` and `manage_ilm`, right?

barkbay: `apmDefaultRoles` is a collection of roles which contain the role `ApmDefaultUserRole`
`monitor` and `manage_ilm` are privileges, while `ingest_admin` is a built-in role.

I'm not sure we can mix a built-in role along with privileges ?

pkg/controller/elasticsearch/user/roles.go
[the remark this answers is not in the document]
simitt: Thank you for the additional information and linking to the monitoring issue, I wasn't aware of the ongoing discussion. 

The reuse of the configured ES user for other features, such as fetching sourcemaps from ES, verifying API Keys and also monitoring is designed for user convenience and flexibility. 
Excluding monitoring privileges will lead to monitoring data simply not showing up in ES, which I expect to be potentially hard to debug and identify. From my perspective we should generally aim for as little setup and config requirements as possible. I can definitely see though why you are in favor of postponing the out-of-the box monitoring support until there is a clear decision in how to support it for the whole stack.

barkbay:
> Just as information: with 7.6 we released experimental support of API Key auth for communication between APM agents and APM Server. The APM Server provides a CLI tool to manage such API Keys. The required privileges for that are the manage_api_key on clusters, and adding specific APM application privileges.

`manage_api_key` seems to exist only in 7.5.x, I'm still a little bit new re. roles in Elasticsearch but I guess it means that we would have to manage 2 different roles: one for ES < 7.5 and an other for ES >= 7.5.

> In this case no Kibana privileges are required for any APM Server >= 7.0, as dashboards are no longer shipped with the APM Server.

Same here, we still support 6.8. I would keep `apm_user` in the list of roles for backward compatibility.

barkbay:
It is a little bit more complicated that what I thought :) 

Any attempt to start the APM server with a privilege that does not exist makes the request fail with something like:
```
Failed to connect to backoff(elasticsearch(https://es-apm-sample-es-http.default.svc:9200)): 400 Bad Request: {"error":{"root_cause":[{"type":"illegal_argument_exception","reason":"unknown index privilege [create_doc]. a privilege must be either one of the predefined fixed indices privileges
```

If I understand correctly:

* For 6.8 we would need this kind of role ([doc ref](https://www.elastic.co/guide/en/apm/server/6.8/beats-basic-auth.html#beats-basic-auth)):

```json
{
  "cluster": ["manage_index_templates","monitor"],
  "indices": [
    {
      "names": [ "apm-*" ], 
      "privileges": ["write","create_index"]
    }
  ]
}
```

* From 7.x to 7.4 (incl.) we can use the setup role described [here](https://www.elastic.co/guide/en/apm/server/7.2/feature-roles.html) (which seems to be the one described in this PR) but we can't use `create_doc` or `manage_api_key` which have been introduced in `7.5`.

* Starting from 7.5 we can introduce `create_doc` and `manage_api_key`

pkg/controller/elasticsearch/user/roles.go
```diff
@@ -18,12 +18,46 @@ const (
...
+					Privileges: []string{"write", "create_index"},
+				},
+			},
+		},
+		ApmUserRoleV7: esclient.Role{
+			Cluster: []string{"monitor", "manage_ilm", "manage_index_templates"},
+			Indices: []esclient.IndexRole{
+				{
+					Names:      []string{"apm-*"},
+					Privileges: []string{"manage", "write", "create_index"},
```
barkbay: `"manage"` is actually needed only from 7.3 but I'm not sure it's worth it to introduce a specific role for 7.1 and 7.2

simitt: I'd keep it simple if it doesn't break anything for 7.1 and 7.2

simitt:
Not sure if that helps, but in our integration testing where we can start different versions inkl. 6.8 and early 7.x versions we use a user that has following custom role:
```
apm_server:
  cluster: ['manage_ilm', 'manage_security', 'manage_api_key'] 
  indices:
    - names: ['apm-*']
      privileges: ['write', 'create_index', 'manage', 'manage_ilm']
```
and also built-in roles `apm_server` and `apm_system` 
(see [roles](https://github.com/elastic/apm-integration-testing/blob/master/docker/elasticsearch/roles.yml) and [users_roles](https://github.com/elastic/apm-integration-testing/blob/master/docker/elasticsearch/users_roles))
Note: the `manage_security` role is only necessary when using the `apm-server apikey` CLI, which I don't think is necessary on ECK for now.

barkbay:
> Not sure if that helps, but in our integration testing where we can start different versions inkl. 6.8 and early 7.x versions we use a user that has following custom role:

This does not work for me. If I try to start an 6.8 APM Server with the `"manage_api_key"` privilege  I have the following error: 

```
2020-03-31T14:38:28.406Z	ERROR	pipeline/output.go:100	Failed to connect to backoff(elasticsearch(https://es-apm-sample-es-http.default.svc:9200)): 400 Bad Request:
{
	"error": {
		"root_cause": [{
			"type": "illegal_argument_exception",
			"reason": "unknown cluster privilege [[manage_index_templates, monitor, manage_api_key]]. a privilege must be either one of the predefined fixed cluster privileges [all=[all],monitor_ml=[monitor_ml],manage_ilm=[manage_ilm],manage_ingest_pipelines=[manage_ingest_pipelines],read_ccr=[read_ccr],monitor_watcher=[monitor_watcher],manage_index_templates=[manage_index_templates],manage_ccr=[manage_ccr],monitor=[monitor],manage_rollup=[manage_rollup],none=[none],manage_watcher=[manage_watcher],read_ilm=[read_ilm],manage=[manage],manage_token=[manage_token],manage_ml=[manage_ml],manage_pipeline=[manage_pipeline],monitor_rollup=[monitor_rollup],transport_client=[transport_client],manage_security=[manage_security],manage_saml=[manage_saml],create_snapshot=[create_snapshot]] or a pattern over one of the available cluster actions"
		}],
		"type": "illegal_argument_exception",
		"reason": "unknown cluster privilege [[manage_index_templates, monitor, manage_api_key]]. a privilege must be either one of the predefined fixed cluster privileges [all=[all],monitor_ml=[monitor_ml],manage_ilm=[manage_ilm],manage_ingest_pipelines=[manage_ingest_pipelines],read_ccr=[read_ccr],monitor_watcher=[monitor_watcher],manage_index_templates=[manage_index_templates],manage_ccr=[manage_ccr],monitor=[monitor],manage_rollup=[manage_rollup],none=[none],manage_watcher=[manage_watcher],read_ilm=[read_ilm],manage=[manage],manage_token=[manage_token],manage_ml=[manage_ml],manage_pipeline=[manage_pipeline],monitor_rollup=[monitor_rollup],transport_client=[transport_client],manage_security=[manage_security],manage_saml=[manage_saml],create_snapshot=[create_snapshot]] or a pattern over one of the available cluster actions"
	},
	"status": 400
}
```

simitt:
@barkbay you are right, I only checked that the server can start up. Since API Key handling is only an experimental feature and it seems to be the last blocking bit, I suggest to go ahead without this privilege for now.
Is the APM Server the only component where different privileges and roles available and required for different versions becomes an issue? I could imagine that this is not an APM specific problem.

barkbay:
My understanding is that we need at least 2 different roles: one for v6 and an other v7, because of the `manage_ilm` privilege for example which does not exist in v6 but is required in v7.

The benefit of a third one for > 7.5 would be to add the `manage_api_key` privilege but also replace `write` by `create_doc` which seems [more restrictive](https://www.elastic.co/guide/en/elasticsearch/reference/current/security-privileges.html#privileges-list-indices) from a security pov.

I have no strong opinion but I think that if we have to manage 2 roles we can add a third one.

pkg/controller/elasticsearch/user/roles.go
```diff
@@ -18,12 +18,46 @@ const (
...
+				},
+			},
+		},
+		ApmUserRoleV75: esclient.Role{
+			Cluster: []string{"monitor", "manage_ilm", "manage_api_key"}, // manage_api_key has been introduced in 7.5
+			Indices: []esclient.IndexRole{
+				{
+					Names:      []string{"apm-*", ".monitoring-beats-*"},
+					Privileges: []string{"manage", "create_doc", "create_index"},
```
simitt, line 57: Not trying to be pedantic here, but you also added the `apm_system` role now, which is the built-in role for monitoring, so from my understanding you can remove the dedicated `.monitoring-beats-*` index  from the custom roles now.

barkbay: My bad. Reading https://www.elastic.co/guide/en/elasticsearch/reference/current/built-in-roles.html I was thinking that `apm_system` is enough to send system-level data but not to setup the required indices.

pkg/controller/elasticsearch/user/roles.go
```diff
@@ -18,12 +18,46 @@ const (
 	SuperUserBuiltinRole = "superuser"
 	// ProbeUserRole is the name of the role used by the internal probe user.
 	ProbeUserRole = "elastic_internal_probe_user"
+
+	// ApmUserRoleV6 is the name of the role used by the association user for 6.8.x APMServer instances.
```
pebrc, line 22: Why are we calling it the `association user`? Is it not just the user APM Server will use to connect to Elasticsearch?

barkbay:
Thanks all for your help and the review  🙇 !

PR #2781, opened 2020-03-27
Add validation webhook configurations for all resource types

### LINKED ISSUE #2584: Setup Kibana and APM validation webhooks to catch unknown fields
We inspect unknown fields since https://github.com/elastic/cloud-on-k8s/pull/2433 on Elasticsearch manifests. For consistency we should do the same with Kibana and APMServer.

charith-elastic:
Adds validation webhook configurations for APM Server, Kibana, and Enterprise Search resources. This PR sets up the framework and basic checks (unknown fields, long names, invalid versions, and downgrades). 

Fixes #2584


pkg/controller/common/version/version.go
```diff
@@ -13,6 +13,33 @@ import (
 	"github.com/pkg/errors"
 )
 
+// supported Stack versions. See https://www.elastic.co/support/matrix#matrix_compatibility
+var (
+	SupportedAPMServerVersions        = MinMaxVersion{Min: From(6, 2, 0), Max: From(8, 99, 99)}
+	SupportedElasticsearchVersions    = MinMaxVersion{Min: From(6, 8, 0), Max: From(8, 99, 99)}
```
pebrc: We don't support 7.0 but only > 7.1 because this is the version that made TLS available under the basic license. So because we also support 6.8 this needs to be a non-contiguous range (not sure if that's the right word)

pkg/apis/apm/v1/webhook.go
```diff
@@ -0,0 +1,106 @@
...
+
+func checkNameLength(as *ApmServer) field.ErrorList {
+	return commonv1.CheckNameLength(as)
+}
+
+func checkSupportedVersion(as *ApmServer) field.ErrorList {
+	return commonv1.CheckSupportedStackVersion(as.Spec.Version, version.SupportedAPMServerVersions)
+}
+
+func checkNoDowngrade(prev, curr *ApmServer) field.ErrorList {
```
anyasabo: Can we downgrade APM? I'd have thought that we could, but I couldn't find an answer in the docs.

pkg/controller/common/version/version.go
charith-elastic: Good catch! This is a leftover line that I forgot to remove. I didn't touch the Elasticsearch validation code in this PR because I noticed this problem as well. It's a refactoring that I wanted to do in a subsequent PR but I forgot to mention it. The [logic for Elasticsearch](https://github.com/elastic/cloud-on-k8s/blob/5ae992381ac44e31dc4c0bc72823acfd081ae09d/pkg/controller/elasticsearch/version/supported_versions.go#L19) is unchanged.

pkg/apis/apm/v1/webhook.go
pebrc: Maybe @simitt or @graphaelli can advise? But my hunch would be that downgrades are not advisable as I think there is some version specific bootstrapping happening in Elasticsearch with index templates etc.

charith-elastic: As you say, it's not documented; so I am not sure either. I was following the example of Elasticsearch where even a patch version downgrade is not allowed by the validation check. 

I do want to address that issue in another PR later. In the mean time, my feeling is that it's OK to keep the somewhat rigid check in place for now. People who are confident that a downgrade is not going to affect them can disable the webhook and still go through with it if they want to. WDYT?

pkg/apis/apm/v1/webhook_test.go
```diff
@@ -0,0 +1,152 @@
...
+)
+
+func TestWebhook(t *testing.T) {
+	testCases := []test.ValidationWebhookTestCase{
+		{
+			Name:      "create-valid",
+			Operation: admissionv1beta1.Create,
+			Object: func(t *testing.T, uid string) []byte {
+				apm := mkApmServer(uid)
+				return serialize(t, apm)
```
anyasabo: Does it maybe make sense to offload the serialization calls to RunValidationWebhookTests, and change the ValidationWebhookTestCase.Object fields to a generic object? As it is there's a lot of boiler plate in TestCase.Object. That said, even if it does make sense I'm still okay with merging this in today as is and leaving that for a future refactoring.

anyasabo:
jenkins test this please

pkg/apis/apm/v1/webhook.go
simitt: That's an interesting question, I honestly never tried to downgrade an installation, and I have never seen a request for it. As long as there is no real good reason I would not actively support downgrading.

anyasabo:
cla/check

PR #2783, opened 2020-03-27
Rename pause annotation

### LINKED ISSUE #2594: Rename pause annotation
`common.k8s.elastic.co/pause: "true"` is somewhat misleading as it might indicate to users that it pauses the managed containers i.e. the Elasticsearch processes which is of course not the case. 

* Let's find a better name. A few ideas:
  * `common.k8s.elastic.co/suspend-reconciliation` (or `pause-reconciliation`)
  * `common.k8s.elastic.co/freeze` (or `freeze-config`)
  * `common.k8s.elastic.co/cordon` (maybe problematic as you can cordon k8s nodes as well)
  * `common.k8s.elastic.co/lock-config`  
* Rename and document the new annotation

pebrc:
New `eck.k8s.elastic.co/managed` annotation, with inverted boolean
semantics: false stops reconciliation. True is implied if not
specified by the user.

Legacy pause annotation is still supported for existing deployments.

Fixes #2594


docs/operating-eck/troubleshooting.asciidoc
```diff
@@ -170,11 +170,11 @@ Due to link:https://github.com/eBay/Kubernetes/blob/master/docs/devel/api-conven
...
 
-[id="{p}-pause-controllers"]
-== Pause ECK controllers
+[id="{p}-exclude-resource"]
+== Exclude resources from reconciliation
 
-When debugging Elasticsearch, you night need to "pause" the operator reconciliations, so that no resource gets modified or created in the meantime.
-To do this, set the annotation `common.k8s.elastic.co/pause` to `true` to any resource controlled by the operator:
+When debugging, you might need to temporarily exclude an Elastic stack resource from management by ECK. When you do that no Kubernetes object depending on that stack resource is modified while you investigate.
+To do this, set the annotation `eck.k8s.elastic.co/managed` to `false` on any resource controlled by the operator:
```
charith-elastic:
```suggestion
For debugging purposes, you might want to temporarily prevent ECK from modifying Kubernetes resources belonging to a particular Elastic Stack resource. To do this, annotate the Elastic object with `eck.k8s.elastic.co/managed=false`. This annotation can be added to any of the following types of objects:
```

pkg/controller/common/unmanaged.go
```diff
@@ -0,0 +1,53 @@
...
+
+	test := func(ann string, defaultValue bool) bool {
+		stateAsString, exists := annotations[ann]
+		if !exists {
+			return defaultValue
+		}
+
+		expectedState, err := strconv.ParseBool(stateAsString)
+		if err != nil {
+			log.Error(err, "Cannot parse %s as a bool, defaulting to %s: \"%v\"", annotations[ann], ann, defaultValue)
```
charith-elastic:
```suggestion
			log.Error(err, "Cannot parse %s=%s as a bool, defaulting to %t", ann, stateAsString, defaultValue)
```

pkg/controller/common/unmanaged.go
```diff
@@ -0,0 +1,53 @@
...
+			return defaultValue
+		}
+		return expectedState
+	}
+	return test(LegacyPauseAnnoation, false) || !test(ManagedAnnotation, true)
```
charith-elastic: What should happen if both annotations are present? Here we are ignoring the `managed` annotation if the `pause` annotation exists. Should it be the other way around?

pebrc: I was not sure myself. I was mainly thinking about not disrupting any existing clusters with the `pause` annotation so I gave that preference. But I can see your point, and I guess it would not hurt to prefer the new annotation.

charith-elastic: Maybe we can log which annotation was used to make the decision? That way if users are inadvertently using both annotations, they could at least see that only one of them is having an effect.

pebrc: OK I have changed it to 
1. prefer the new annotation
2. log if a user is still using the old one and if the user uses both. 

Logic is a bit convoluted but hopefully clearer for the user.

pkg/controller/common/unmanaged.go
```diff
@@ -0,0 +1,63 @@
...
+
+var (
+
+	// CheckManagedRequeue is the default requeue to re-check if a resource is still unmanaged.
+	CheckManagedRequeue = reconcile.Result{Requeue: true, RequeueAfter: 10 * time.Second}
+)
+
+// IsUnmanaged checks if a given resource is currently unmanaged.
+func IsUnmanaged(meta metav1.ObjectMeta) bool {
+	return getBoolFromAnnotation(meta.Annotations)
```
charith-elastic: Nit: I think `getBoolFromAnnotation` function body could be inlined here because `IsUnmanaged` is more descriptive of the purpose of the function than `getBoolFromAnnotation`

pkg/controller/common/unmanaged.go
```diff
@@ -0,0 +1,58 @@
...
+			log.Error(err, fmt.Sprintf("Cannot parse %s=%s as a bool, defaulting to %t", ann, stateAsString, defaultValue), "namespace", meta.Namespace, "name", meta.Name)
+			return defaultValue, true
+		}
+		return expectedState, true
+	}
+
+	managed, newExists := parse(ManagedAnnotation, true)
+	paused, legacyExists := parse(LegacyPauseAnnoation, false)
+	if legacyExists && !newExists {
+		log.V(-1).Info(fmt.Sprintf("Using legacy annotation %s. Please consider moving to %s", LegacyPauseAnnoation, ManagedAnnotation), "namespace", meta.Namespace, "name", meta.Name)
```
anyasabo: We currently don't use the -1 level for anything else that I saw. I'm a little bit skeptical about starting to use it for this? I'm open to it but could use some persuading. My gut feeling is that no one reads warning messages and that they should be errors or info messages.

anyasabo:
This should have the `>breaking` tag right? We don't have a label to indicate deprecated at this time it looks like.

pkg/controller/common/unmanaged.go
pebrc: Good point will change back to info.

pebrc:
It’s not a breaking change imo because we still support the old annotation.

anyasabo:
>It’s not a breaking change imo because we still support the old annotation.

Yeah that makes sense, but I wonder what else we put it under so that it can be found in release notes or closed PRs. Is it time maybe to add support for a `>deprecated` tag or something similar?

pebrc:
> Is it time maybe to add support for a >deprecated tag or something similar?

I add the `>enhancement` tag as we are improving existing functionality (bit of a stretch, but at least it will show up in release notes now)

pkg/controller/common/unmanaged.go
```diff
@@ -0,0 +1,58 @@
...
+const (
+	// ManagedAnnotation annotation
+	LegacyPauseAnnoation = "common.k8s.elastic.co/pause"
+	ManagedAnnotation    = "eck.k8s.elastic.co/managed"
+)
+
+var (
+
+	// CheckManagedRequeue is the default requeue to re-check if a resource is still unmanaged.
+	CheckManagedRequeue = reconcile.Result{Requeue: true, RequeueAfter: 10 * time.Second}
```
sebgl: I don't think we need that since we already watch the resource?

pkg/controller/common/unmanaged.go
```diff
@@ -0,0 +1,58 @@
...
+	paused, legacyExists := parse(LegacyPauseAnnoation, false)
+	if legacyExists && !newExists {
+		log.Info(fmt.Sprintf("Using legacy annotation %s. Please consider moving to %s", LegacyPauseAnnoation, ManagedAnnotation), "namespace", meta.Namespace, "name", meta.Name)
+		return paused
+	}
+	if legacyExists && newExists {
+		log.Info(fmt.Sprintf("Both %s and %s are set, preferring %s", LegacyPauseAnnoation, ManagedAnnotation, ManagedAnnotation), "namespace", meta.Namespace, "name", meta.Name)
+	}
+	return !managed
+}
```
sebgl: It feels a bit complicated vs. something like below?

```go
func isUnmanaged(meta metav1.ObjectMeta) bool {
    // resource is unmanaged if its annotation is set to "false"
    managed, exists := meta.Annotations[ManagedAnnotation]
    if exists && managed == "false" {
        return true
    }
    
    // handle legacy "paused" annotation
    paused, exists := meta.Annotations[LegacyPauseAnnoation]
    if exists {
        log.Info(fmt.Sprintf("%s is deprecated, please use %s instead", LegacyPauseAnnoation, ManagedAnnotation))
    }
    return exists && paused == "true"
}

pebrc: Simpler is better, will change it.

pebrc: One note though, this breaks current behavour which also accepts 1/0 as boolean values in the annotation but I guess that's OK

pkg/controller/common/unmanaged.go
pebrc: True, I did not question the necessity of this and just carried it over from the existing code. Good thinking!

PR #2784, opened 2020-03-27
Fix kind script

anyasabo:
Fixes some errors introduced in https://github.com/elastic/cloud-on-k8s/pull/2752

You can see the error [here](
https://devops-ci.elastic.co/job/cloud-on-k8s-e2e-tests-kind-k8s-versions/33/console).

Shellcheck is upset that both `${KIND_LOGGING}` and `${config_opts}` are unquoted, but quoting them also causes problems. [This](https://github.com/koalaman/shellcheck/wiki/SC2086) is the article for that error. If I understand it correctly, we want to split on spaces, so it suggests we read from an array. But then I run into other issues that I don't quite understand. 

For these I just disabled the lint check to get it up and running, and added a TODO for someone better at bash to figure out if there's a better way.

Also fixes [SC2206](https://github.com/koalaman/shellcheck/wiki/SC2206) by switching to `read -ra`.

To verify it works locally, you can use
```
LOAD_IMAGES=docker.elastic.co/k8s/build-tools,gcr.io/distroless/base  NODE_IMAGE="kindest/node:v1.15.3" ./hack/kind/kind.sh
```
or substitute other images you have downloaded locally.

### DIFF
--- a/hack/kind/kind.sh
+++ b/hack/kind/kind.sh
@@ -14,7 +14,12 @@
 # Exit immediately for non zero status
 set -e
 
-KIND_LOGGING=${KIND_LOGGING:-"-v 1"}
+if [ -n "${KIND_LOGGING}" ]; then
+  IFS=' ' read -r -a log_lvl <<< "${KIND_LOGGING}"
+else
+  log_lvl=("-v" "1")
+fi
+
 CLUSTER_NAME=${KIND_CLUSTER_NAME:-eck-e2e}
 NODES=3
 MANIFEST=/tmp/cluster.yml
@@ -25,7 +30,7 @@ scriptpath="$( cd "$(dirname "$0")" ; pwd -P )"
 
 function check_kind() {
   echo "Check if Kind is installed..."
-  if ! kind --help > /dev/null; then
+  if ! command -v kind ; then
     echo "Looks like Kind is not installed."
     exit 1
   fi
@@ -79,14 +84,13 @@ function setup_kind_cluster() {
     echo "No existing kind cluster with name ${CLUSTER_NAME}. Continue..."
   fi
 
-  config_opts=""
+  config_opts=()
   if [[ ${NODES} -gt 0 ]]; then
-    config_opts="--config ${MANIFEST}"
+    config_opts+=("--config" "${MANIFEST}")
   fi
   # Create Kind cluster
   # TODO: see if this lint rule can be re-enabled
-  # shellcheck disable=SC2086
-  if ! (kind ${KIND_LOGGING} create cluster --name="${CLUSTER_NAME}" ${config_opts} --retain --image "${NODE_IMAGE}"); then
+  if ! (kind "${log_lvl[@]}" create cluster --name="${CLUSTER_NAME}" "${config_opts[@]}" --retain --image "${NODE_IMAGE}"); then
     echo "Could not setup Kind environment. Something wrong with Kind setup."
     exit 1
   fi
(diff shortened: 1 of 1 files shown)


charith-elastic:
This seems to work for me (not exhaustively tested though)

```patch
diff --git a/hack/kind/kind.sh b/hack/kind/kind.sh
index f779c488..057d6f92 100755
--- a/hack/kind/kind.sh
+++ b/hack/kind/kind.sh
@@ -13,10 +13,8 @@
 
 # Exit immediately for non zero status
 set -e
-# Print commands
-set -x
 
-KIND_LOGGING=${KIND_LOGGING:-"-v 1"}
+KIND_LOGGING=("-v" "1")
 CLUSTER_NAME=${KIND_CLUSTER_NAME:-eck-e2e}
 NODES=3
 MANIFEST=/tmp/cluster.yml
@@ -27,7 +25,7 @@ scriptpath="$( cd "$(dirname "$0")" ; pwd -P )"
 
 function check_kind() {
   echo "Check if Kind is installed..."
-  if ! kind --help > /dev/null; then
+  if ! command -v kind ; then
     echo "Looks like Kind is not installed."
     exit 1
   fi
@@ -81,12 +79,12 @@ function setup_kind_cluster() {
     echo "No existing kind cluster with name ${CLUSTER_NAME}. Continue..."
   fi
 
-  config_opts=""
+  config_opts=()
   if [[ ${NODES} -gt 0 ]]; then
-    config_opts="--config ${MANIFEST}"
+	config_opts+=("--config" "${MANIFEST}")
   fi
   # Create Kind cluster
-  if ! (kind "${KIND_LOGGING}" create cluster --name="${CLUSTER_NAME}" "${config_opts}" --retain --image "${NODE_IMAGE}"); then
+  if ! (kind "${KIND_LOGGING[@]}" create cluster --name="${CLUSTER_NAME}" "${config_opts[@]}" --retain --image "${NODE_IMAGE}"); then
     echo "Could not setup Kind environment. Something wrong with Kind setup."
     exit 1
   fi
@@ -139,7 +137,7 @@ fi
 if [[ -n "${LOAD_IMAGES}" ]]; then
   IMAGES=("${LOAD_IMAGES//,/ }")
   for image in "${IMAGES[@]}"; do
-          kind "${KIND_LOGGING}" --name "${CLUSTER_NAME}" load docker-image --nodes "${workers}" "${image}"
+          kind "${KIND_LOGGING[@]}" --name "${CLUSTER_NAME}" load docker-image --nodes "${workers}" "${image}"
   done
 fi
 ```

**Comment by anyasabo** [comment]
Thanks @charith-elastic , that helped a lot. I made some modifications to allow users to override the log level still, but otherwise used what you provided.

File: hack/kind/kind.sh [hunk 1]
```diff
@@ -13,10 +13,13 @@
...
-# Print commands
-set -x
 
-KIND_LOGGING=${KIND_LOGGING:-"-v 1"}
+if [ -n "${KIND_LOGGING}" ]; then
+  IFS=' ' read -r -a log_lvl <<< "${KIND_LOGGING}"
+else
+  log_lvl=("-v" "1")
+fi
+
```
charith-elastic: I think it would be simpler this way:

```suggestion
KIND_LOG_LEVEL=${KIND_LOG_LEVEL:-"1"}
log_lvl=("-v" "$KIND_LOG_LEVEL")
```

WDYT?

anyasabo: 👍  good call

PR #2794, opened 2020-03-31
Improve secure string generation

barkbay:
This PR improves password and token generation by relying on [github.com/sethvargo/go-password](github.com/sethvargo/go-password)

It also improves security by moving from `match/rand` to `crypto/rand`


anyasabo:
Can you also open a follow up issue for the string length constants please?

barkbay:
> Can you also open a follow up issue for the string length constants please?

Sure: https://github.com/elastic/cloud-on-k8s/issues/2795

Issue #2595, opened 2020-02-21
Document upgrade process for production ECK deployments


pebrc:
While the current upgrade guide explicitly mentions the fact that all clusters will go through a rolling restart for example when upgrading from a beta version to the 1.0 release, this is still very subtle and can be missed. In addition just rolling out the ECK upgrade maybe problematic if the fleet of managed clusters is large as it might lead to resource contention if all clusters are restarted at the same time etc.

Suggestion: create additional documentation around the upgrade process that outlines mitigating strategies. For example:
* pausing all reconciliations before the upgrade (but see #2594 )
* rolling out the operator upgrade
* successively un-pausing the reconciliation of the managed resources starting with the lowest priority resources and moving to the higher priority resources after that

PR #2797, opened 2020-03-31
Expand upgrading ECK doc + add annotator script

charith-elastic:
Adds more information to the "Upgrading ECK" document to inform users about the rolling restart of all resources and how to manage them effectively.

Also adds a script to help with annotating all Elastic resources in a cluster.

Fixes: #2595
Depends on: #2783

Preview: http://cloud-on-k8s_2797.docs-preview.app.elstc.co/guide/en/cloud-on-k8s/master/k8s-upgrading-eck.html

### DIFF
--- a/docs/operating-eck/upgrading-eck.asciidoc
+++ b/docs/operating-eck/upgrading-eck.asciidoc
@@ -53,7 +53,7 @@ Once the operator has been upgraded and you are ready to let the resource become
 
 
 [source,shell,subs="attributes,callouts"]
-.Resume Elastic resource managment by the operator
+.Resume Elastic resource management by the operator
 ----
 RM_ANNOTATION='common.k8s.elastic.co/pause-' <1>
 
--- a/docs/operating-eck/upgrading-eck.asciidoc
+++ b/docs/operating-eck/upgrading-eck.asciidoc
@@ -9,13 +9,14 @@ endif::[]
 
 [float]
 [id="{p}-ga-upgrade"]
-== Upgrade to newer releases of ECK
+== Upgrade to ECK {eck_version}
 
-ECK reached general availability (GA) status with the link:https://www.elastic.co/blog/elastic-cloud-on-kubernetes-ECK-is-now-generally-available[release of version 1.0.0]. It is compatible with the previous beta release (1.0.0-beta1) and can be upgraded in-place (<<{p}-ga-openshift, with a few exceptions>>). Previous alpha releases, up to and including version 0.9.0, are not compatible with the GA and beta releases and link:https://www.elastic.co/guide/en/cloud-on-k8s/1.0-beta/k8s-upgrading-eck.html[require extra work to upgrade].
+ECK reached general availability (GA) status with the link:https://www.elastic.co/blog/elastic-cloud-on-kubernetes-ECK-is-now-generally-available[release of version 1.0.0]. The latest available GA version is {eck_version}. It is compatible with the previous beta release (1.0.0-beta1) and can be upgraded in-place (<<{p}-ga-openshift, with a few exceptions>>). Previous alpha releases, up to and including version 0.9.0, are not compatible with the GA and beta releases and link:https://www.elastic.co/guide/en/cloud-on-k8s/1.0-beta/k8s-upgrading-eck.html[require extra work to upgrade].
+
 
 [float]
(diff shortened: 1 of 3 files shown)


docs/operating-eck/upgrading-eck.asciidoc
```diff
@@ -9,24 +9,65 @@ endif::[]
...
-- Kubernetes versions < 1.13
-- OpenShift 3.11
+<1> Since ECK 1.1.0, this annotation has been superseded by `eck.k8s.elastic.co/managed=false`
 
-In this case, ECK must be uninstalled and reinstalled (due to link:https://github.com/kubernetes/kubernetes/issues/73752[an upstream bug] that was not backported). To uninstall ECK, see <<{p}-uninstalling-eck>>. Then you can follow the <<{p}-quickstart>> to install anew. You can save your existing `v1beta1` manifests, update them to the `v1` API version, <<{p}-snapshots,take snapshots>> and restore them as well.
+Once the operator has been upgraded and you are ready to let the resource become managed again (triggering a rolling restart of pods in the process), remove the annotation.
+
+
+[source,shell,subs="attributes,callouts"]
+.Resume Elastic resource managment by the operator
```
thbkrkr, line 56:
```suggestion
.Resume Elastic resource management by the operator
```

docs/operating-eck/upgrading-eck.asciidoc
```diff
@@ -9,24 +9,65 @@ endif::[]
...
+----
+RM_ANNOTATION='common.k8s.elastic.co/pause-' <1>
+
+# Resume management of a single Elasticsearch cluster named "quickstart"
+kubectl annotate elasticsearch quickstart $RM_ANNOTATION
+----
+
+<1> Since ECK 1.1.0, this annotation has been superseded by `eck.k8s.elastic.co/managed-`
```
thbkrkr, line 64: The final dash seems unexpected.

hack/annotator/annotator.sh
```diff
@@ -0,0 +1,78 @@
...
+
+# Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+# or more contributor license agreements. Licensed under the Elastic License;
+# you may not use this file except in compliance with the Elastic License.
+
+# Script to automate the task of adding or removing the "exclude" annotation from/to all Elastic resources in a Kubernetes cluster.
+
+set -euo pipefail
+
+ANN_KEY=${ANN_KEY:-"common.k8s.elastic.co/pause"}
```
thbkrkr, line 11: Let's use the latest annotation `eck.k8s.elastic.co/managed`?

docs/operating-eck/upgrading-eck.asciidoc
charith-elastic: It is the way to remove an annotation though. Any suggestions to make it look like it's not a typo?

hack/annotator/annotator.sh
charith-elastic: The reason I went with the old annotation is because the documentation is targeted at people who are trying to upgrade from 1.0.0-beta1. The new annotation is only available from 1.1.0 onwards.

hack/annotator/annotator.sh
```diff
@@ -0,0 +1,78 @@
...
+    remove)
+        remove_all
+        ;;
+    help)
+        usage
+        ;;
+    *)
+        echo "Unknown action '$1'"
+        usage
+        ;;
```
pebrc, line 77: I wonder if for a production deployment you would want to have more control, i.e. create certain cohorts of Elastic stack apps and remove labels not all at once by cohort by cohort.

charith-elastic: That would be pretty hard for us to generalise and put into a script, wouldn't it? I reckon the grouping logic would be unique to each user (grouping based on custom label combinations for example). My intention was to provide a starting point for users that can then be adapted to fit their particular requirements. WDYT?

docs/operating-eck/upgrading-eck.asciidoc
thbkrkr: In this case, it is legitimate and I think it's okay.

I'm not sure it will be more clear, we could keep the dash in the command like this:
```kubectl annotate elasticsearch quickstart ${RM_ANNOTATION}-```.

I'm fine with keeping it like this.

Issue #2445, opened 2020-01-18
Replace sync flush with flush for ES 8.0+


sebgl:
See https://github.com/elastic/elasticsearch/pull/50882

PR #2798, opened 2020-03-31
Perform a normal flush instead of a sync flush when targeting Elasticsearch 8.0

sebgl:
Fixes https://github.com/elastic/cloud-on-k8s/issues/2445.

Synced flush is deprecated starting 7.6, and removed starting 8.0.
Instead, we should perform a "normal" flush during rolling upgrades.

That's what this commit does.

One important implementation detail: it only inspects the expected ES
version, not the version of the currently running Pods, for simplicity.
When upgrading from 7.x to 8.x, a normal flush will be done.
In theory, we could do a synced flush for the first
pod to upgrade (since there are no 8.0 pods yet), then a normal flush
for the remaining pods (since the request may target the already
upgraded 8.0 pod).  We must make sure we don't do a synced flush if
an 8.x pod is running, or should be running.
Since synced flush is a best effort operation, and
7.x does also support normal flush, I think it's not worth the trouble.
Also, the [7.x to 8.x upgrade documentation](https://www.elastic.co/guide/en/elasticsearch/reference/master/rolling-upgrades.html) does not mention synced flush.

I noticed the current 8.0.0-SNAPSHOT still accepts `/_flush/synced` requests,
but that may disappear anytime.

### DIFF
--- a/pkg/controller/elasticsearch/client/client.go
+++ b/pkg/controller/elasticsearch/client/client.go
@@ -71,9 +71,11 @@ type Client interface {
 	DisableReplicaShardsAllocation(ctx context.Context) error
 	// EnableShardAllocation enables shards allocation on the cluster.
 	EnableShardAllocation(ctx context.Context) error
-	// SyncedFlush requests a synced flush on the cluster.
+	// SyncedFlush requests a synced flush on the cluster. Deprecated in 7.6, removed in 8.0.
 	// This is "best-effort", see https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-synced-flush.html.
 	SyncedFlush(ctx context.Context) error
+	// Flush requests a flush on the cluster.
+	Flush(ctx context.Context) error
 	// GetClusterHealth calls the _cluster/health api.
 	GetClusterHealth(ctx context.Context) (Health, error)
 	// SetMinimumMasterNodes sets the transient and persistent setting of the same name in cluster settings.
--- a/pkg/controller/elasticsearch/client/v6.go
+++ b/pkg/controller/elasticsearch/client/v6.go
@@ -55,6 +55,10 @@ func (c *clientV6) SyncedFlush(ctx context.Context) error {
 	return c.post(ctx, "/_flush/synced", nil, nil)
 }
 
+func (c *clientV6) Flush(ctx context.Context) error {
+	return c.post(ctx, "/_flush", nil, nil)
+}
+
 func (c *clientV6) GetClusterHealth(ctx context.Context) (Health, error) {
 	var result Health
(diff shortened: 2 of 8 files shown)


pkg/controller/elasticsearch/client/client.go
```diff
@@ -71,9 +71,11 @@ type Client interface {
 	DisableReplicaShardsAllocation(ctx context.Context) error
 	// EnableShardAllocation enables shards allocation on the cluster.
 	EnableShardAllocation(ctx context.Context) error
-	// SyncedFlush requests a synced flush on the cluster.
+	// SyncedFlush requests a synced flush on the cluster. Deprecated in 7.6, removed in 8.0.
 	// This is "best-effort", see https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-synced-flush.html.
 	SyncedFlush(ctx context.Context) error
+	// Flush requests a flush on the cluster.
```
sebgl: If you're bored you can try to repeat this comment several times as quickly as possible. It's very hard to pronounce!
(I'm getting ready for April 1st)

anyasabo:
LGTM. I wonder if we also need to update the supported upgrade path to min 7.6 as well?
https://github.com/elastic/cloud-on-k8s/blob/master/pkg/controller/elasticsearch/version/supported_versions.go#L35

The [current upgrade docs ](https://www.elastic.co/guide/en/elasticsearch/reference/current/rolling-upgrades.html#_upgrading_your_cluster)still indicate a synced flush for 7.x and the [docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-synced-flush-api.html) also say that flush == synced flush only in 7.6, not 7.4.

sebgl:
> I wonder if we also need to update the supported upgrade path to min 7.6 as well?
https://github.com/elastic/cloud-on-k8s/blob/master/pkg/controller/elasticsearch/version/supported_versions.go#L35

I think this is the official version upgrade doc: https://www.elastic.co/guide/en/elasticsearch/reference/master/setup-upgrade.html
The matrix seems to indicate that the upgrade path `is 7.[0-3] -> 7.4 -> 8.x`.

> The current upgrade docs still indicate a synced flush for 7.x and the docs also say that flush == synced flush only in 7.6, not 7.4.

Any (non-version change) rolling upgrade on 7.x will still use synced flush, including 7.6 (even though deprecated).
Only version upgrades to 8.x will do a non-synced flush.

anyasabo:
Ah yeah I see the -> 8.x docs mention that now. The mismatch between being able to upgrade from 7.4 -> 8.0, but synced flush only being equivalent to flush in 7.6 confused me.

PR #2799, opened 2020-03-31
Remove enterprise search CRDs

anyasabo:
Removes enterprise search CRDs for 1.1 release. This seemed to be the least invasive way to  make the change.


charith-elastic:
I think the webhook configuration for Enterprise Search has to be removed from `config/operator/all-in-one/webhook.yaml.template` as well.

PR #2803, opened 2020-04-01
Compatibility check: add Service ownership test

### LINKED ISSUE #2693: Kibana resource fails to create in 1.0.1
## Bug Report
**What did you do?**
Create a Kibana instance using the Operator
A working Kibana instance
The custom resource is present, but the operator fails to create the required resources for the Kibana instance
**Environment**

* ECK version:

	1.0.1

* Kubernetes information:

	* AKS version: 1.15.7
```
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.1", GitCommit:"d224476cd0730baca2b6e357d144171ed74192d6", GitTreeState:"clean", BuildDate:"2020-01-14T21:04:32Z", GoVersion:"go1.13.5", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.7", GitCommit:"c83d931fb9bece427bc63a02349755e0f8696d3e", GitTreeState:"clean", BuildDate:"2020-01-31T20:09:49Z", GoVersion:"go1.12.12", Compiler:"gc", Platform:"linux/amd64"}
```

* Resource definition:
```
apiVersion: kibana.k8s.elastic.co/v1beta1
kind: Kibana
metadata:
  name: kibana
spec:
  image: <obfuscated>.azurecr.io/kibana/kibana:7.5.1
  version: 7.5.1
  count: 1
  elasticsearchRef:
    name: elasticsearch
  podTemplate:
    metadata:
      labels:
        managed-by: kustomize
        app: logging
    spec:
      containers:
        - name: kibana
          resources:
            requests:
              memory: 1Gi
              cpu: 0.5
            limits:
              memory: 2Gi
              cpu: 1
  http:
    tls:
      selfSignedCertificate:
        disabled: true
```

* Logs:
```
{"level":"info","@timestamp":"2020-03-12T11:35:45.974Z","logger":"kibana-association-controller","message":"Starting reconciliation run","ver":"1.0.1-bcb74688","iteration":19,"namespace":"operator-logging","name":"kibana"}
{"level":"info","@timestamp":"2020-03-12T11:35:45.974Z","logger":"annotation","message":"Resource was created with older version of operator, will not take action","ver":"1.0.1-bcb74688","controller_version":"1.0.1","resource_controller_version":"0.0.0-UNKNOWN","namespace":"operator-logging","name":"kibana"}
```

We noticed that the annotation that's being set on the Kibana resource is 0.0.0-UNKNOWN when the resource is created. When we (manually) update this to 1.0.1, the operator kicks in and creates the resources. 

```
kind: Kibana
metadata:  
    annotations:    
        common.k8s.elastic.co/controller-version: 0.0.0-UNKNOWN
....
```

barkbay:
Fix #2693 by checking that existing Services actually belong to the reconciled resource. They may have the right labels but belong to a previous parent which has been deleted.

### DIFF
--- a/pkg/controller/common/annotation/controller_version.go
+++ b/pkg/controller/common/annotation/controller_version.go
@@ -14,6 +14,7 @@ import (
 	"go.elastic.co/apm"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
@@ -138,21 +139,25 @@ func ReconcileCompatibility(ctx context.Context, client k8s.Client, obj runtime.
 // The labels provided must exactly match.
 func checkExistingResources(client k8s.Client, obj runtime.Object, labels map[string]string) (bool, error) {
 
-	accessor := meta.NewAccessor()
-	namespace, err := accessor.Namespace(obj)
+	metaObject, err := meta.Accessor(obj)
 	if err != nil {
-		log.Error(err, "error getting namespace", "kind", obj.GetObjectKind().GroupVersionKind().Kind)
 		return false, err
 	}
+
 	labelSelector := ctrlclient.MatchingLabels(labels)
-	nsSelector := ctrlclient.InNamespace(namespace)
-	// if there's no controller version annotation on the object, then we need to see maybe the object has been reconciled by an older, incompatible controller version
+	nsSelector := ctrlclient.InNamespace(metaObject.GetNamespace())
+
 	var svcs corev1.ServiceList
 	err = client.List(&svcs, labelSelector, nsSelector)
 	if err != nil {
 		return false, err
 	}
-	// if we listed any services successfully, then we know this cluster was reconciled by an old version since any objects reconciled by a 0.9.0+ operator would have a label
-	return len(svcs.Items) != 0, nil
 
+	// We only want Services managed by this particular instance of the parent object
+	for _, svc := range svcs.Items {
+		if v1.IsControlledBy(&svc, metaObject) {
+			return true, nil
+		}
+	}
+	return false, nil
 }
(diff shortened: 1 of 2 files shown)


pkg/controller/common/annotation/controller_version.go
```diff
@@ -138,21 +139,25 @@ func ReconcileCompatibility(ctx context.Context, client k8s.Client, obj runtime.
...
-	nsSelector := ctrlclient.InNamespace(namespace)
-	// if there's no controller version annotation on the object, then we need to see maybe the object has been reconciled by an older, incompatible controller version
+	nsSelector := ctrlclient.InNamespace(metaObject.GetNamespace())
+
 	var svcs corev1.ServiceList
 	err = client.List(&svcs, labelSelector, nsSelector)
 	if err != nil {
 		return false, err
 	}
-	// if we listed any services successfully, then we know this cluster was reconciled by an old version since any objects reconciled by a 0.9.0+ operator would have a label
```
sebgl, line 155: I think this comment was pretty useful

pkg/controller/common/annotation/controller_version.go
```diff
@@ -14,6 +14,7 @@ import (
 	"go.elastic.co/apm"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
sebgl, line 17: let's try to be consistent with `metav1`

pkg/controller/common/annotation/controller_version.go
anyasabo: It could definitely be better written though, I think I wrote it and it is still confusing to me. Maybe

>// if we listed any services successfully, then we know this owner resource was reconciled by an old version since any owner resources reconciled by a 0.9.0+ operator would have a label already

But it is kind of leaking abstractions and belongs in the caller func IMO. This func just checks for existing resources, so it doesn't necessarily need to only be used for old objects (but it's also unexported so ehhhhhh).

PR #2805, opened 2020-04-02
Fix Filebeat in OCP for E2E tests monitoring

david-kow:
OpenShift needs SecurityContextConstraints correctly configured to deploy privileged Filebeat pods.

### DIFF
--- a/config/e2e/filebeat.yaml
+++ b/config/e2e/filebeat.yaml
@@ -249,9 +249,6 @@ allowedCapabilities: []
 defaultAddCapabilities: []
 fsGroup:
   type: RunAsAny
-groups:
-  - system:cluster-admins
-  - system:nodes
 priority: 0
 readOnlyRootFilesystem: false
 requiredDropCapabilities: []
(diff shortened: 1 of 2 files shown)


config/e2e/filebeat.yaml
```diff
@@ -231,3 +232,39 @@ roleRef:
...
+allowHostPID: false
+allowHostPorts: false
+allowPrivilegeEscalation: true
+allowPrivilegedContainer: true
+allowedCapabilities: []
+defaultAddCapabilities: []
+fsGroup:
+  type: RunAsAny
+groups:
+  - system:cluster-admins
```
barkbay, line 253: nit: does these 2 groups really need to belong to this SCC ?

david-kow: They do not. Fixed.

david-kow:
Privileged is more permissive than needed I think. In addition to that it might be cleaner to have a separate resource that we create and clean up rather than patch existing one and unpatch it later. I don't believe we have a mechanism to do it in our E2E runner right now.

Issue #2823, opened 2020-04-06
TLS connection error on the transport protocol during rolling upgrades


sebgl:
I observed several E2E tests failing during a rolling upgrade (example: https://devops-ci.elastic.co/job/cloud-on-k8s-e2e-tests-stack-versions/38/).

I think it randomly impacts all rolling upgrades where nodes IPs are replaced.

Symptoms:
* one of the node never joins the cluster, it cannot connect to the other nodes:

ES 7.1.1 logs:
```
{"type": "server", "timestamp": "2020-04-06T02:26:55,898+0000", "level": "INFO", "component": "o.e.c.c.JoinHelper", "cluster.name": "test-reverted-mutation-x9g8", "node.name": "test-reverted-mutation-x9g8-es-masterdata-2",  "message": "failed to join {test-reverted-mutation-x9g8-es-masterdata-0}{kZoqIftrTdyNcZkZ3HziSA}{l3XsMzFhSwWyCAiRwZpDhQ}{10.1.178.50}{10.1.178.50:9300}{ml.machine_memory=2147483648, ml.max_open_jobs=20, xpack.installed=true} with JoinRequest{sourceNode={test-reverted-mutation-x9g8-es-masterdata-2}{5ac1-E9KQ4OsnLsszGdbrg}{XCCcVXlpRwKtdRjcMj67_Q}{10.1.177.44}{10.1.177.44:9300}{ml.machine_memory=2147483648, xpack.installed=true, box_type=mixed, ml.max_open_jobs=20}, optionalJoin=Optional[Join{term=2, lastAcceptedTerm=1, lastAcceptedVersion=29, sourceNode={test-reverted-mutation-x9g8-es-masterdata-2}{5ac1-E9KQ4OsnLsszGdbrg}{XCCcVXlpRwKtdRjcMj67_Q}{10.1.177.44}{10.1.177.44:9300}{ml.machine_memory=2147483648, xpack.installed=true, box_type=mixed, ml.max_open_jobs=20}, targetNode={test-reverted-mutation-x9g8-es-masterdata-0}{kZoqIftrTdyNcZkZ3HziSA}{l3XsMzFhSwWyCAiRwZpDhQ}{10.1.178.50}{10.1.178.50:9300}{ml.machine_memory=2147483648, ml.max_open_jobs=20, xpack.installed=true}}]}" , 
[...]

PR #2831, opened 2020-04-08
Revert transport TLS certs verification from "full" to "certificate"

sebgl:
A bug in Elasticsearch
(https://github.com/elastic/elasticsearch/issues/54867) leads to TLS
certificates not being picked up during rolling upgrades. Impacted ES
nodes present a certificate with the wrong (old) Pod IP address.

Full TLS verification in those conditions lead to nodes not being able
to trust each other during (and after) rolling upgrades.

Let's switch TLS verification back to "certificate" (IP address is
ignored in the certificate but we still validate the cert signature),
until we figure out the proper way to reintroduce "full" verification.

Fixes https://github.com/elastic/cloud-on-k8s/issues/2823.


sebgl:
Issue to eventually reintroduce full TLS verification: https://github.com/elastic/cloud-on-k8s/issues/2833.

PR #2848, opened 2020-04-09
Add some temporary debugging to diagnose e2e test failures

pebrc:
* Print result of cat shards when health check tests fail.
* Set master service logging to trace to dianose cluster health issues.

### DIFF
--- a/test/e2e/test/elasticsearch/debug.go
+++ b/test/e2e/test/elasticsearch/debug.go
@@ -26,7 +26,7 @@ func printShardsAndAllocation(clientFactory func() (client.Client, error)) func(
 }
 
 func printResponse(esClient client.Client, url string) {
-	catShards, err := http.NewRequest(
+	request, err := http.NewRequest(
 		http.MethodGet,
 		url,
 		nil,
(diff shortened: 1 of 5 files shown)


test/e2e/test/elasticsearch/debug.go
```diff
@@ -0,0 +1,46 @@
...
+	"net/http"
+
+	"github.com/elastic/cloud-on-k8s/pkg/controller/elasticsearch/client"
+)
+
+func printShards(clientFactory func() (client.Client, error)) func() {
+	return func() {
+		esClient, err := clientFactory()
+		if err != nil {
+			fmt.Printf("while creating es client %v", err)
```
anyasabo, line 20:
```suggestion
			fmt.Printf("error while creating es client: %v", err)
```

test/e2e/test/elasticsearch/debug.go
```diff
@@ -0,0 +1,46 @@
...
+			return
+		}
+		catShards, err := http.NewRequest(
+			http.MethodGet,
+			"/_cat/shards",
```
anyasabo, line 25: might be worth adding a `?v` to include the header

test/e2e/test/elasticsearch/debug.go
```diff
@@ -0,0 +1,46 @@
...
+			return
+		}
+		defer shards.Body.Close()
+		bytes, err := ioutil.ReadAll(shards.Body)
+		if err != nil {
+			fmt.Printf("error while reading response body: %v \n", err)
+			return
+		}
+		fmt.Println("GET _cat/shards")
+		fmt.Printf("%s \n", bytes)
```
sebgl, line 44: Should we also get `/_cluster/allocation/explain`?

pebrc: I would assume we have that already in the logs due to the trace logging, but I can add it

Issue #2856, opened 2020-04-10
Volume claim templates validation fails if the storage unit is different


sebgl:
Observed in https://discuss.elastic.co/t/volume-claim-templates-cannot-be-modified/219493/7.

If we apply the following manifest:

```yaml
apiVersion: elasticsearch.k8s.elastic.co/v1
kind: Elasticsearch
metadata:
  name: elasticsearch-sample
spec:
  version: 7.6.0
  nodeSets:
  - name: default
    count: 3
    config:
      node.store.allow_mmap: false
    volumeClaimTemplates:
      - metadata:
          name: elasticsearch-data
        spec:
          accessModes:
            - ReadWriteOnce
          resources:
            requests:
              storage: 1024Gi
```

The underlying StatefulSet is created with `1Ti` storage (not `1024Gi`).
Any update on the above manifest, while keeping `1024Gi`, fails because the webhook rejects the volumeClaimTemplates update.
I guess because it's comparing `1024Gi` to `1Ti`.

PR #2857, opened 2020-04-10
Don't reject PVC update when a different unit is used

sebgl:
Fixes https://github.com/elastic/cloud-on-k8s/issues/2856.

When we apply a manifest with a 1024Gi storage requirement, it may be
internally persisted with a different storage unit (eg. 1Ti). When doing
a strict deep equal comparison, 1Ti and 1024Gi are different.
A user applying such a manifest will not have any update allowed, even
though he keeps using 1024Gi.

This fixes it by relying on `apiequality.Semantic.DeepEqual`, which is
used in Kubernetes itself:
https://github.com/kubernetes/kubernetes/blob/c369cf187ea765c0a2387f2b39abe6ed18c8e6a8/pkg/apis/apps/validation/validation.go#L156

### DIFF
--- a/pkg/apis/elasticsearch/v1/validations.go
+++ b/pkg/apis/elasticsearch/v1/validations.go
@@ -7,13 +7,14 @@ package v1
 import (
 	"fmt"
 	"net"
-	"reflect"
+
+	apiequality "k8s.io/apimachinery/pkg/api/equality"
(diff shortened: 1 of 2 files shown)


sebgl:
I'd like to include this fix in the 1.1 release.

PR #2858, opened 2020-04-10
Backport - Don't reject PVC update when different unit is used (#2857)

sebgl:
Backport https://github.com/elastic/cloud-on-k8s/pull/2857 on 1.1.

When we apply a manifest with a 1024Gi storage requirement, it may be
internally persisted with a different storage unit (eg. 1Ti). When doing
a strict deep equal comparison, 1Ti and 1024Gi are different.
A user applying such a manifest will not have any update allowed, even
though he keeps using 1024Gi.

This fixes it by relying on `apiequality.Semantic.DeepEqual`, which is
used in Kubernetes itself:
https://github.com/kubernetes/kubernetes/blob/c369cf187ea765c0a2387f2b39abe6ed18c8e6a8/pkg/apis/apps/validation/validation.go#L156

### DIFF
--- a/pkg/apis/elasticsearch/v1/validations.go
+++ b/pkg/apis/elasticsearch/v1/validations.go
@@ -7,13 +7,14 @@ package v1
 import (
 	"fmt"
 	"net"
-	"reflect"
+
+	apiequality "k8s.io/apimachinery/pkg/api/equality"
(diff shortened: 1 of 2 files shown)


sebgl:
Jenkins test this please.

PR #2859, opened 2020-04-10
Set Enterprise Search sample version to 7.7.0

sebgl:
We only support 7.7.0+, and now have a validation webhook that prevents using 7.6.0.

### DIFF
--- a/config/samples/enterprisesearch/ent_es.yml
+++ b/config/samples/enterprisesearch/ent_es.yml
@@ -18,7 +18,7 @@ kind: EnterpriseSearch
 metadata:
   name: ent-sample
 spec:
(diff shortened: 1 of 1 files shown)


thbkrkr:
This seems important enough to be documented?

We maintain the supported versions of the Elastic Stack in two places:
* https://github.com/elastic/cloud-on-k8s/blob/master/docs/supported-versions.asciidoc
* https://github.com/elastic/cloud-on-k8s/blob/master/README.md

sebgl:
Enterprise Search does simply not exist before 7.7.0.
Do you think it's worth calling out in the docs? I think it's ok to not do it, to keep things simple.

Issue #2864, opened 2020-04-11
Do not cache Elasticsearch settings in annotations


barkbay:
`attemptDownscale` may delete some nodes which are not excluded.

For example, let's consider the following situation where 2 `StatefulSets` with an initial `replicas: 2` are downscaled to `replicas: 1`:

## Initial state

* sset1:
  * sset1-0
  * sset1-1
* sset2:
  * sset2-0
  * sset2-1

## Iteration 1

* `actualStatefulSets=[sset1,sset2]`
* `sset1-1` is scheduled to be deleted:
  * `allocationSetter.ExcludeFromShardAllocation(ctx, ["sset1-1"])` is successful
  * `updateAllocationExcludeAnnotation(es, ["sset1-1"])` is successful
* `sset1-1` is not deleted yet because it is holding some data

## Iteration 2

* `actualStatefulSets=[sset2,sset1]` - note the inconsistent order, this is because there is no guarantee regarding the order of the StatefulSets retrieved from the K8S API
* `sset2-1` is scheduled to be deleted:
  * `allocationSetter.ExcludeFromShardAllocation(ctx, ["sset2-1"])` is successful
  * `updateAllocationExcludeAnnotation(es, ["sset2-1"])` **fails** _(e.g.: Operation cannot be fulfilled on elasticsearches.elasticsearch.k8s.elastic.co \"test-mutation-2nd-master-set-dtvx\": the object has been modified;...)_
* _at this point it is possible that `sset1-1` does not hold some data anymore_

At the end of this iteration there is an inconsistency between the excluded nodes in ES and the annotation set on the `Elasticsearch`object: `sset1-1` is not excluded any more even if the annotation on ES is saying the opposite 💣 💣💣

## Iteration 3
* `actualStatefulSets=[sset1,sset2]` - note that now the order is the same than the one at iteration 1
* `sset1-1` is scheduled to be deleted:
  * `allocationSetter.ExcludeFromShardAllocation(ctx, ["sset1-1"])` is NOT called because of the stale annotation:
```golang
	if exclusions == allocationExcludeFromAnnotation(es) {
		return nil
	}
```

* `calculatePerformableDownscale` does not detect any shards on `sset1-1` because it has been excluded in the past.
* `sset1-1` is deleted ... but since it is not excluded anymore there is no guarantee that we have not delete some data with it 💥 

Could be the root cause of #2788 in which 2 `StatefulSets` are downscaled/deleted.

[...]

Issue #2788, opened 2020-03-31
Red health during E2E test mutation


sebgl:
https://devops-ci.elastic.co/job/cloud-on-k8s-e2e-tests-stack-versions/32/testReport/github/com_elastic_cloud-on-k8s_test_e2e_es/Run_tests_for_different_ELK_stack_versions_in_GKE___7_6_0___TestMutationSecondMasterSetDown_ES_cluster_health_should_eventually_be_green_01/

```
=== RUN   TestMutationSecondMasterSetDown/ES_cluster_health_should_eventually_be_green#01
Retries (5m0s timeout): ....................................................................................................
{"log.level":"error","@timestamp":"2020-03-31T02:13:11.533Z","message":"stopping early","service.version":"0.0.0-00000000","service.type":"eck","ecs.version":"1.4.0","error":"test failure","error.stack_trace":"github.com/go-logr/zapr.(*zapLogger).Error\n\t/go/pkg/mod/github.com/go-logr/zapr@v0.1.0/zapr.go:128\ngithub.com/elastic/cloud-on-k8s/test/e2e/test.StepList.RunSequential\n\t/go/src/github.com/elastic/cloud-on-k8s/test/e2e/test/step.go:42\ngithub.com/elastic/cloud-on-k8s/test/e2e/test.RunMutations\n\t/go/src/github.com/elastic/cloud-on-k8s/test/e2e/test/run_mutation.go:37\ngithub.com/elastic/cloud-on-k8s/test/e2e/test.RunMutation\n\t/go/src/github.com/elastic/cloud-on-k8s/test/e2e/test/run_mutation.go:78\ngithub.com/elastic/cloud-on-k8s/test/e2e/es.RunESMutation\n\t/go/src/github.com/elastic/cloud-on-k8s/test/e2e/es/mutation_test.go:299\ngithub.com/elastic/cloud-on-k8s/test/e2e/es.TestMutationSecondMasterSetDown\n\t/go/src/github.com/elastic/cloud-on-k8s/test/e2e/es/mutation_test.go:168\ntesting.tRunner\n\t/usr/local/go/src/testing/testing.go:909"}
    --- FAIL: TestMutationSecondMasterSetDown/ES_cluster_health_should_eventually_be_green#01 (300.00s)
        utils.go:84: 
            	Error Trace:	utils.go:84
            	Error:      	Received unexpected error:
            	            	health is red
[...]

PR #2869, opened 2020-04-13
Check node excluded from shard allocation before deleting

pebrc:
Related to #2864 

I wrote this when experimenting with the bug described in #2788 and #2864 While this is not a fix for the bug itself, this amended check at least should have prevented the data loss we saw. Happy to close this PR if this is already part of the more comprehensive fix @barkbay  is working on for #2864 

This amends the checks we do to verify node evacuation is complete. Previously we only checked that no shards are present on the node in question.
This however was not good enough as a missing allocation filter for that node could mean that shards move back to the node just before deletion.
The new check looks at allocation filters and shards. While this check is not perfect as allocation filters can also be removed at the last moment before deletion, it narrows the
window for that to happen somewhat.

### DIFF
--- a/pkg/controller/elasticsearch/client/client.go
+++ b/pkg/controller/elasticsearch/client/client.go
@@ -43,7 +43,7 @@ type Role struct {
 
 // Client captures the information needed to interact with an Elasticsearch cluster via HTTP
 type Client interface {
-	AllocationSetter
+	AllocationFilter
 	ShardLister
 	LicenseClient
 	// Close idle connections in the underlying http client.
--- a/pkg/controller/elasticsearch/client/shard.go
+++ b/pkg/controller/elasticsearch/client/shard.go
@@ -8,11 +8,14 @@ import (
 	"context"
 )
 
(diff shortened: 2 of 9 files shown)


anyasabo:
jenkins test this please

pebrc:
>I think this algorithm does not cover the case where the last successful call to ExcludeFromShardAllocation was not done with the node we want to remove (because this node is in the annotation, but it is not the one used in the last API call) Edit: and assuming that ECK now only wants to remove that particular node, not the previous one

> In that case ExcludedFromShardAllocation may never contains the expected node and NodeEvacuated may never return true

Yes, that's correct. But I thought not making progress would be preferable to data loss. So we need another PR that fixes the actual problem of annotations being out of sync with reality and probably sorting the stateful sets so that we don't change the order in which we do downscales unnecessarily.

pebrc:
We discussed offline and decided to include a fix as part of  #2864 and not do an additional allocation filter check. There is a clean up of a unit test in this PR which is worth keeping. I will probably raise this in a separate PR.

PR #2871, opened 2020-04-14
Unit test Enterprise Search reconciliation

sebgl:
Add unit tests to cover Enterprise Search reconciliation.

This PR also fixes a few things I noticed along the way:

* don't use a `-server` suffix for the deployment (comes from apmserver code)
* we already call `onDelete()` on resource deletion, no need to also call it when a resources is about to be deleted
* remove the watch on ES TLS secret (already watched since owned)

### DIFF
--- a/pkg/controller/enterprisesearch/enterprisesearch_controller.go
+++ b/pkg/controller/enterprisesearch/enterprisesearch_controller.go
@@ -199,7 +199,6 @@ func (r *ReconcileEnterpriseSearch) Reconcile(request reconcile.Request) (reconc
 func (r *ReconcileEnterpriseSearch) onDelete(obj types.NamespacedName) {
 	// Clean up watches
 	r.dynamicWatches.Secrets.RemoveHandlerForKey(configRefWatchName(obj))
-	r.dynamicWatches.Secrets.RemoveHandlerForKey(tlsSecretWatchName(obj))
 }
 
 func (r *ReconcileEnterpriseSearch) isCompatible(ctx context.Context, ent *entv1beta1.EnterpriseSearch) (bool, error) {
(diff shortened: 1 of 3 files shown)


pkg/controller/enterprisesearch/enterprisesearch_controller.go
```diff
@@ -185,12 +185,6 @@ func (r *ReconcileEnterpriseSearch) Reconcile(request reconcile.Request) (reconc
 		return reconcile.Result{}, tracing.CaptureError(ctx, err)
 	}
 
-	if ent.IsMarkedForDeletion() {
```
barkbay, line 188: 👍

pkg/controller/enterprisesearch/enterprisesearch_controller.go
```diff
@@ -363,7 +358,7 @@ func watchEsTLSCertsSecret(ent entv1beta1.EnterpriseSearch, watched watches.Dyna
 		return nil
 	}
 	return watched.Secrets.AddHandler(watches.NamedWatch{
-		Name:    tlsSecretWatchName(ent),
+		Name:    tlsSecretWatchName(k8s.ExtractNamespacedName(&ent)),
```
barkbay, line 361: It does not come from this PR but I'm wondering if we really need this watch since the `Secret` that contains the Elasticsearch CA is owned by the `entsearch` resource ? Owned resources are already watched: https://github.com/elastic/cloud-on-k8s/blob/883f45d16f8f5f4f09474fb2ada0cafac33d797c/pkg/controller/enterprisesearch/enterprisesearch_controller.go#L118-L124

sebgl: Yes! Let's remove it.

sebgl:
> Small nit: I would add a test that checks that dynamic watches are cleared when apierrors.IsNotFound(err) is true (right after association.FetchWithAssociation)

Isn't it what this test does? https://github.com/elastic/cloud-on-k8s/pull/2871/files#diff-1c253eb2f7ab39297e0d74855ad5f220R90

barkbay:
> Isn't it what this test does? https://github.com/elastic/cloud-on-k8s/pull/2871/files#diff-1c253eb2f7ab39297e0d74855ad5f220R90

My bad, I was not on the last commit of your branch when I reviewed and ran the tests locally 😐

PR #2875, opened 2020-04-14
Add Elasticsearch client cache

barkbay:
ECK attempts to reduce the number of API calls to Elasticsearch by storing the api call arguments in annotations.
Unfortunately when ECK fails to update the annotation it may lead to some consistencies: see #2788 and #2864 

This PR adds an in memory cache to keep track of the last arguments used when calling the ES API for the following operations:
- [X] Allocation settings
- [X] Voting config exclusions
- [X] Zen1 minimum master nodes
- [ ] Remote clusters

_still a draft but ready for review re. the overall architecture_

### DIFF
--- a/pkg/controller/elasticsearch/client/cache.go
+++ b/pkg/controller/elasticsearch/client/cache.go
@@ -1,0 +1,137 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License;
+// you may not use this file except in compliance with the Elastic License.
+
+package client
+
+import (
+	"context"
+	"sync"
+
+	"k8s.io/apimachinery/pkg/types"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+var log = logf.Log.WithName("cached-client")
+
+type CachedClientBuilder interface {
+
+	// NewElasticsearchCachedClient returns an Elasticsearch client that relies on a cache for some operations.
+	// Important: do not use the client concurrently for a same operation. Concurrent calls of a given operation is not thread-safe.
+	NewElasticsearchCachedClient(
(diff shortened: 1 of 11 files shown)


pkg/controller/elasticsearch/client/cache.go
```diff
@@ -0,0 +1,185 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+var log = logf.Log.WithName("cached-client")
+
+type CachedClientBuilder interface {
+
+	// NewElasticsearchCachedClient returns an Elasticsearch client that relies on a cache for some operations.
+	// Do not use the resulting client concurrently for a same operation. Concurrent calls of a given operation is not thread-safe.
```
pebrc, line 22:
```suggestion
	// Do not use the resulting client concurrently for the same operation. Concurrent calls of a given operation are not thread-safe.
```

pkg/controller/elasticsearch/client/cache_test.go
```diff
@@ -0,0 +1,308 @@
...
+	assert.Equal(t, 4, es1FakeClient.addVotingConfigExclusionsCalled)
+
+}
+
+func Test_cachedClient_ExcludeFromShardAllocation(t *testing.T) {
+
+	steps := []struct {
+		es1Value, es2Value                   *string // value to set when calling the api
+		es1ExpectedValue, es2ExpectedValue   *string // expected value stored in the fake client
+		es1ExpectedCalled, es2ExpectedCalled int     // number of time an api is called
```
pebrc, line 140:
```suggestion
		es1ExpectedCalled, es2ExpectedCalled int     // number of times an api is called
```

barkbay:
Discussed out of band: we decided to not rely on any form of cache and always perform sync. calls to Elasticsearch API

PR #2877, opened 2020-04-14
Reconcile Enterprise Search status

sebgl:
Similar to what we do for other CRDs, update the resource status
at the end of the reconciliation, with:

- red/green health
- number of available pods in the deployment
- service name

### DIFF
--- a/pkg/controller/enterprisesearch/enterprisesearch_controller.go
+++ b/pkg/controller/enterprisesearch/enterprisesearch_controller.go
@@ -8,6 +8,8 @@ import (
 	"context"
 	"crypto/sha256"
 	"fmt"
+	"reflect"
+	"sync/atomic"
 
 	"go.elastic.co/apm"
 	appsv1 "k8s.io/api/apps/v1"
@@ -22,6 +24,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
+	commonv1 "github.com/elastic/cloud-on-k8s/pkg/apis/common/v1"
 	entv1beta1 "github.com/elastic/cloud-on-k8s/pkg/apis/enterprisesearch/v1beta1"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/association"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common"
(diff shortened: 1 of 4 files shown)


sebgl:
Jenkins test this please.

pkg/controller/enterprisesearch/enterprisesearch_controller.go
```diff
@@ -290,6 +285,37 @@ func (r *ReconcileEnterpriseSearch) validate(ctx context.Context, ent *entv1beta
...
+		Association:     ent.Status.Association,
+	}
+	for _, c := range deploy.Status.Conditions {
+		if c.Type == appsv1.DeploymentAvailable && c.Status == corev1.ConditionTrue {
+			newStatus.Health = entv1beta1.EnterpriseSearchGreen
+		}
+	}
+
+	if reflect.DeepEqual(newStatus, ent.Status) {
+		return nil // nothing to do
```
barkbay, line 304: I was wondering if this branch does not deserve a test since it protects the API Server from unnecessary calls.

sebgl: You're right. Do you have an idea for how we could test this? (mocking apiserver calls etc. feels a bit overkill)

barkbay: May be we can add a wrapper around the `Status()` receiver ?

```golang
type fakeClient struct {
	updateCalled bool
	k8s.Client
}

func newFakeClient(initObjs ...runtime.Object) k8s.Client {
	return &fakeClient{
		false,
		k8s.WrappedFakeClient(initObjs...),
	}

}

func (f *fakeClient) Status() k8s.StatusWriter {
	f.updateCalled = true // Status() has been requested for update
	return f.Client.Status()
}
```

sebgl: Yeah let's do that!

pkg/controller/enterprisesearch/enterprisesearch_controller_test.go
```diff
@@ -277,3 +277,245 @@ func TestReconcileEnterpriseSearch_Reconcile_Create_Update_Resources(t *testing.
...
+				default:
+					// ok
+				}
+			}
+		})
+	}
+}
+
+//
+//func TestReconcileEnterpriseSearch_noUpdate(t *testing.T) {
```
barkbay, line 492: I think this one is part of TestReconcileEnterpriseSearch_updateStatus now ?

sebgl: 🤦

sebgl:
Jenkins test this please.

PR #2881, opened 2020-04-15
Retry k8s api calls in the test init phase

sebgl:
Any network blip would cause the entire E2E test suite to fail in the
init phase where we just perform a few basic api calls that are ok to
retry.

Fixes https://github.com/elastic/cloud-on-k8s/issues/2775.


sebgl:
Jenkins test this please.

thbkrkr:
Looks eventually good to me ;-)

PR #2885, opened 2020-04-15
Revert "Disable debug e2e job (#2865)"

anyasabo:
This reverts commit e242d414c9475f85752da53590f85124907d18e0.

Re-enables the test to make sure https://github.com/elastic/cloud-on-k8s/pull/2880 entirely fixes the issue and there are not any further bugs. We were able to reproduce the bug pretty soon last time we had it enabled, so we should only need it for the rest of the week to feel confident it's resolved.

### DIFF
--- a/.ci/pipelines/e2e-tests-testmutationsecondmastersetdown.Jenkinsfile
+++ b/.ci/pipelines/e2e-tests-testmutationsecondmastersetdown.Jenkinsfile
@@ -91,7 +91,7 @@ pipeline {
         }
         cleanup {
             script {
-                if (notOnlyDocs() && failedTests.size == 0) {
+                if (notOnlyDocs() && failedTests.size() == 0) {
                     build job: 'cloud-on-k8s-e2e-cleanup',
                         parameters: [string(name: 'JKS_PARAM_GKE_CLUSTER', value: "eck-e2e-${BUILD_NUMBER}")],
                         wait: false
--- a/.ci/pipelines/e2e-tests-testmutationsecondmastersetdown.Jenkinsfile
+++ b/.ci/pipelines/e2e-tests-testmutationsecondmastersetdown.Jenkinsfile
@@ -1,0 +1,112 @@
+// This library overrides the default checkout behavior to enable sleep+retries if there are errors
+// Added to help overcome some recurring github connection issues
+@Library('apm@current') _
+
+def failedTests = []
+def lib
+
+pipeline {
+
+    agent {
+        label 'linux'
+    }
+
+    options {
+        timeout(time: 300, unit: 'MINUTES')
+    }
+
+    environment {
+        VAULT_ADDR = credentials('vault-addr')
+        VAULT_ROLE_ID = credentials('vault-role-id')
+        VAULT_SECRET_ID = credentials('vault-secret-id')
+        GCLOUD_PROJECT = credentials('k8s-operators-gcloud-project')
+    }
+
+    stages {
+        stage('Checkout from GitHub') {
+            steps {
+                checkout scm
+            }
+        }
+        stage('Load common scripts') {
+            steps {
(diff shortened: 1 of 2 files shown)


.ci/pipelines/e2e-tests-testmutationsecondmastersetdown.Jenkinsfile
```diff
@@ -0,0 +1,112 @@
...
+                      message: msg,
+                    tokenCredentialId: 'cloud-ci-slack-integration-token',
+                    botUser: true,
+                    failOnError: true
+                )
+            }
+        }
+        cleanup {
+            script {
+                if (notOnlyDocs() && failedTests.size == 0) {
```
pebrc, line 94: This does not work. I am not 100% sure why, but when we ran this clusters were never cleaned up. I deleted the ones for successful runs manually once we had exceeded our quota.

.ci/pipelines/e2e-tests-testmutationsecondmastersetdown.Jenkinsfile
```diff
@@ -0,0 +1,112 @@
...
+                expression {
+                    notOnlyDocs()
+                }
+            }
+            steps {
+                sh '.ci/setenvconfig e2e/master'
+                
+                sh 'echo TESTS_MATCH = TestMutationSecondMasterSetDown >> .env'
+                sh 'echo E2E_SKIP_CLEANUP = true >> .env'
+                sh 'sed "s/CLUSTER_NAME=eck-e2e/CLUSTER_NAME=eck-debug-e2e/" -i .env'
```
pebrc, line 59: This does not seem to work ^

.ci/pipelines/e2e-tests-testmutationsecondmastersetdown.Jenkinsfile
```diff
@@ -0,0 +1,112 @@
...
+                    botUser: true,
+                    failOnError: true
+                )
+            }
+        }
+        cleanup {
+            script {
+                if (notOnlyDocs() && failedTests.size == 0) {
+                    build job: 'cloud-on-k8s-e2e-cleanup',
+                        parameters: [string(name: 'JKS_PARAM_GKE_CLUSTER', value: "eck-e2e-${BUILD_NUMBER}")],
```
pebrc, line 96: The cluster name here is not aligned with the rename attempt on L59

.ci/pipelines/e2e-tests-testmutationsecondmastersetdown.Jenkinsfile
thbkrkr: The replacement works well but I forgot how it works. We also need to rename the `clusterName` defined in the `deployer-config.yml`:

`sh 'sed "s/clusterName: eck-e2e/clusterName: eck-debug-e2e/" -i deployer-config.yml'`

.ci/pipelines/e2e-tests-testmutationsecondmastersetdown.Jenkinsfile
thbkrkr: In Jenkins, a scripted pipeline is a groovy-based DSL and in this context, we can't omit the parentheses when we call `size`:
```groovy
if (notOnlyDocs() && failedTests.size() == 0) {
```

pebrc: Also will `failedTests.size()` not always be 0 because we have commented out the junit bits that report failed test into Jenkins?

anyasabo: it looks like you're probably right? We can try and update it if we end up needing it to debug in the future

PR #2887, opened 2020-04-15
Clarify upgrade instructions

charith-elastic:
Some changes to the text to clarify the upgrade process from previous beta or GA releases to the latest release.

### DIFF
--- a/docs/operating-eck/upgrading-eck.asciidoc
+++ b/docs/operating-eck/upgrading-eck.asciidoc
@@ -13,6 +13,8 @@ endif::[]
 
 ECK reached general availability (GA) status with the link:https://www.elastic.co/blog/elastic-cloud-on-kubernetes-ECK-is-now-generally-available[release of version 1.0.0]. The latest available GA version is {eck_version}. It is compatible with the previous GA releases (1.0.0 and 1.0.1) and the beta release (1.0.0-beta1), and can be upgraded in-place <<{p}-ga-openshift, with a few exceptions>>. Previous alpha releases, up to and including version 0.9.0, are not compatible with the GA and beta releases and link:https://www.elastic.co/guide/en/cloud-on-k8s/1.0-beta/k8s-upgrading-eck.html[require extra work to upgrade].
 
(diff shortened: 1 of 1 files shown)


docs/operating-eck/upgrading-eck.asciidoc
```diff
@@ -11,18 +11,17 @@ endif::[]
...
-
+ECK reached general availability (GA) status with the link:https://www.elastic.co/blog/elastic-cloud-on-kubernetes-ECK-is-now-generally-available[release of version 1.0.0]. The latest available GA version is {eck_version}. It is compatible with the previous GA releases (1.0.0 and 1.0.1) and the beta release (1.0.0-beta1), and can be upgraded in-place <<{p}-ga-openshift, with a few exceptions>>. Previous alpha releases, up to and including version 0.9.0, are not compatible with the GA and beta releases and link:https://www.elastic.co/guide/en/cloud-on-k8s/1.0-beta/k8s-upgrading-eck.html[require extra work to upgrade].
```
anyasabo, line 14: Should we clarify that "an in place upgrade" is just "kubectl apply"ing the new all-in-one?

Also I think we also want to call out the breaking changes here. I'm not sure if just a link is appropriate or if we should discuss how to remediate them in this doc:
https://www.elastic.co/guide/en/cloud-on-k8s/master/release-notes-1.1.0.html

charith-elastic: > Should we clarify that "an in place upgrade" is just "kubectl apply"ing the new all-in-one?

I thought it was obvious and didn't go into specifics. Happy to reconsider if it isn't.

I think the mitigation steps (if any) for breaking changes should go into release highlights. Since it's possible to upgrade by skipping a few intermediate versions, listing the breaking changes for each version here might be a bit too confusing.

anyasabo: >I thought it was obvious and didn't go into specifics. Happy to reconsider if it isn't.

I just could also see it be updating the version tag in the image rather than applying the whole manifest again, but maybe that is just me.

For the breaking changes documentation, I think from here we should at least link/direct people to them? As a user I would expect that I just need to read the "upgrading ECK" page to find out all the things I need to know for upgrading ECK.

Maybe something along the lines of "Please review the breaking changes section of the release highlights before upgrading"?

charith-elastic:
http://cloud-on-k8s_2887.docs-preview.app.elstc.co/guide/en/cloud-on-k8s/master/k8s-upgrading-eck.html

PR #2891, opened 2020-04-16
Use annotation to track created remote clusters

### LINKED ISSUE #2864: Do not cache Elasticsearch settings in annotations

barkbay:
The PR fixes an issue where the annotation used to cache the Elasticsearch state wrt remote clusters may become inconsistent and could prevent some updates to be applied.

The annotation is kept but only to track the remote clusters that have been previously managed by ECK. In order to be consistent with the other API calls in ECK the algorithm never tries to optimize the number of calls performed on the ES cluster.

It can be noted that it still exists an edge case where ECK may delete a cluster managed by the user:
1. Remote cluster `remote-foo` is declared as remote in the Spec
2. User decides to manage this cluster itself, `remote-foo` is removed from the spec
3. ECK is stopped or is to slow to update the annotation
4. User creates `remote-foo` manually
5. ECK is started or is finally proceeding the cluster removal from the spec, deleting `remote-foo` 💥 

At each iteration a log is printed with the upserted and deleted remote clusters.

Last PR to completely fix #2864 , related to #2880 

@anyasabo I think it should be included in `1.1.0` to avoid a mix of "remote clusters related" labels on Elasticsearch resources in the future.

### DIFF
--- a/pkg/controller/elasticsearch/remotecluster/annotations.go
+++ b/pkg/controller/elasticsearch/remotecluster/annotations.go
@@ -13,8 +13,8 @@ import (
 )
 
 const (
-	// CreatedRemoteClustersAnnotationName holds the list of the remote clusters which have been created
-	CreatedRemoteClustersAnnotationName = "elasticsearch.k8s.elastic.co/managed-remote-clusters"
+	// ManagedRemoteClustersAnnotationName holds the list of the remote clusters which have been created
+	ManagedRemoteClustersAnnotationName = "elasticsearch.k8s.elastic.co/managed-remote-clusters"
 )
 
 // getRemoteClustersInAnnotation returns a set that contains a list of remote clusters that may have been declared in Elasticsearch.
@@ -22,7 +22,7 @@ const (
 // If there's no remote clusters the map is empty but not nil.
 func getRemoteClustersInAnnotation(es esv1.Elasticsearch) map[string]struct{} {
 	remoteClusters := make(map[string]struct{})
-	serializedRemoteClusters, ok := es.Annotations[CreatedRemoteClustersAnnotationName]
+	serializedRemoteClusters, ok := es.Annotations[ManagedRemoteClustersAnnotationName]
 	if !ok {
 		return remoteClusters
 	}
@@ -41,6 +41,6 @@ func annotateWithCreatedRemoteClusters(c k8s.Client, es esv1.Elasticsearch, remo
 		annotation = append(annotation, remoteCluster)
 	}
 	sort.Strings(annotation)
-	es.Annotations[CreatedRemoteClustersAnnotationName] = strings.Join(annotation, ",")
+	es.Annotations[ManagedRemoteClustersAnnotationName] = strings.Join(annotation, ",")
 	return c.Update(&es)
 }
--- a/pkg/controller/elasticsearch/remotecluster/annotations.go
+++ b/pkg/controller/elasticsearch/remotecluster/annotations.go
@@ -5,58 +5,42 @@
 package remotecluster
 
 import (
-	"encoding/json"
+	"sort"
+	"strings"
 
 	esv1 "github.com/elastic/cloud-on-k8s/pkg/apis/elasticsearch/v1"
 	"github.com/elastic/cloud-on-k8s/pkg/utils/k8s"
-	"github.com/pkg/errors"
 )
 
 const (
-	RemoteClustersAnnotationName = "elasticsearch.k8s.elastic.co/remote-clusters"
+	// CreatedRemoteClustersAnnotationName holds the list of the remote clusters which have been created
+	CreatedRemoteClustersAnnotationName = "elasticsearch.k8s.elastic.co/managed-remote-clusters"
 )
 
-type expectedRemoteClusterConfiguration struct {
-	esv1.RemoteCluster
-
-	// ConfigHash is the hash of the remote cluster configuration. It is used to detect when the settings must be updated.
-	ConfigHash string `json:"configHash"`
-}
-
-// getCurrentRemoteClusters returns a map with the current configuration hash of the remote clusters declared in Elasticsearch.
+// getRemoteClustersInAnnotation returns a set that contains a list of remote clusters that may have been declared in Elasticsearch.
 // A map is returned here to quickly compare with the ones that are new or missing.
 // If there's no remote clusters the map is empty but not nil.
-func getCurrentRemoteClusters(es esv1.Elasticsearch) (map[string]string, error) {
-	serializedRemoteClusters, ok := es.Annotations[RemoteClustersAnnotationName]
-	remoteClusters := make(map[string]string)
+func getRemoteClustersInAnnotation(es esv1.Elasticsearch) map[string]struct{} {
+	remoteClusters := make(map[string]struct{})
+	serializedRemoteClusters, ok := es.Annotations[CreatedRemoteClustersAnnotationName]
 	if !ok {
-		return remoteClusters, nil
+		return remoteClusters
 	}
-	if err := json.Unmarshal([]byte(serializedRemoteClusters), &remoteClusters); err != nil {
-		return nil, err
+	for _, remoteClusterInAnnotation := range strings.Split(serializedRemoteClusters, ",") {
+		remoteClusters[remoteClusterInAnnotation] = struct{}{}
 	}
-
-	return remoteClusters, nil
+	return remoteClusters
 }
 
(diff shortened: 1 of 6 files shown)


pkg/controller/elasticsearch/remotecluster/elasticsearch.go
```diff
@@ -25,88 +26,143 @@ var log = logf.Log.WithName("remotecluster")
 const enterpriseFeaturesDisabledMsg = "Remote cluster is an enterprise feature. Enterprise features are disabled"
 
 // UpdateSettings updates the remote clusters in the persistent settings by calling the Elasticsearch API.
+// See the documentation of updateSettingsInternal for more information about the algorithm.
```
pebrc, line 29: Maybe document that the bool return value means `requeue`?

pkg/controller/elasticsearch/remotecluster/elasticsearch.go
```diff
@@ -25,88 +26,143 @@ var log = logf.Log.WithName("remotecluster")
...
+// getRemoteClustersInElasticsearch returns all the remote clusters currently declared in Elasticsearch
+func getRemoteClustersInElasticsearch(esClient esclient.Client) (map[string]struct{}, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), esclient.DefaultReqTimeout)
+	defer cancel()
+	remoteClustersInEs := make(map[string]struct{})
+	remoteClusterSettings, err := esClient.GetRemoteClusterSettings(ctx)
+	if err != nil {
+		return remoteClustersInEs, err
+	}
+	for remoteClusterName := range remoteClusterSettings.PersistentSettings.Cluster.RemoteClusters {
```
pebrc, line 150: I guess we can ignore transient settings for the purposes of this function?

barkbay: This is the assumption I made since we are only interested in clusters created by ECK.

pkg/controller/elasticsearch/remotecluster/annotations.go
```diff
@@ -5,58 +5,42 @@
...
 
 	esv1 "github.com/elastic/cloud-on-k8s/pkg/apis/elasticsearch/v1"
 	"github.com/elastic/cloud-on-k8s/pkg/utils/k8s"
-	"github.com/pkg/errors"
 )
 
 const (
-	RemoteClustersAnnotationName = "elasticsearch.k8s.elastic.co/remote-clusters"
+	// CreatedRemoteClustersAnnotationName holds the list of the remote clusters which have been created
+	CreatedRemoteClustersAnnotationName = "elasticsearch.k8s.elastic.co/managed-remote-clusters"
```
sebgl, line 17: `CreatedRemoteClustersAnnotationName ` -> `ManagedRemoteClustersAnnotationName ` to better match the annotation name?

pkg/controller/elasticsearch/remotecluster/elasticsearch.go
```diff
@@ -25,88 +26,143 @@ var log = logf.Log.WithName("remotecluster")
...
+	return updateSettingsInternal(remoteClustersInSpec, c, esClient, es)
+}
+
+// updateSettingsInternal updates remote clusters in Elasticsearch. It also keeps track of any remote clusters which
+// have been declared in the Elasticsearch spec. The purpose is to delete remote clusters which were managed by
+// the operator but are not desired anymore, without removing the ones which have been added by the user.
+// The following algorithm is used:
+// 1. Get the list of the previously declared remote clusters from the annotation
+// 2. Ensure that all remote clusters in the Elasticsearch spec are present in the annotation
+// 3. For each remote cluster in the annotation which is not the Spec, either:
```
sebgl, line 66: `not in the Spec`

pkg/controller/elasticsearch/remotecluster/elasticsearch.go
```diff
@@ -25,88 +26,143 @@ var log = logf.Log.WithName("remotecluster")
...
+//   3.1 Schedule its deletion from the Elasticsearch settings
+//   3.2 Otherwise remove it from the annotation
+// 4. Update the annotation on the Elasticsearch object
+// 5. Apply the settings through the Elasticsearch API
```
sebgl, line 70: I love that sort of comment 👍

pkg/controller/elasticsearch/remotecluster/elasticsearch.go
```diff
@@ -25,88 +26,143 @@ var log = logf.Log.WithName("remotecluster")
...
+func updateSettingsInternal(
+	remoteClustersInSpec map[string]esv1.RemoteCluster,
+	c k8s.Client,
+	esClient esclient.Client,
+	es esv1.Elasticsearch,
+) (bool, error) {
```
sebgl, line 76: I would also appreciate a comment on that bool here (or maybe just use a named result eg. `(requeue bool, err error)`)

pkg/controller/elasticsearch/remotecluster/elasticsearch.go
```diff
@@ -25,88 +26,143 @@ var log = logf.Log.WithName("remotecluster")
...
-			seedHosts := []string{services.ExternalTransportServiceHost(remoteCluster.ElasticsearchRef.NamespacedName())}
-			log.Info("Adding or updating remote cluster",
-				"namespace", es.Namespace,
-				"es_name", es.Name,
-				"remote_cluster", remoteCluster.Name,
-				"seeds", seedHosts,
-			)
-			remoteClusters[name] = esclient.RemoteCluster{Seeds: seedHosts}
+	var remoteClustersToDelete []string
+	// For each remote cluster in the annotation, either:
```
sebgl, line 86: `For each remote cluster in the annotation but not in the spec, either:`

pkg/controller/elasticsearch/remotecluster/elasticsearch.go
```diff
@@ -25,88 +26,143 @@ var log = logf.Log.WithName("remotecluster")
...
+	for _, name := range remoteClustersToDelete {
+		remoteClustersToApply[name] = esclient.RemoteCluster{Seeds: nil}
+	}
+
+	// Update the annotation
+	if err := annotateWithCreatedRemoteClusters(c, es, remoteClustersInAnnotation); err != nil {
+		return true, err
+	}
+
+	// Since the annotation is update before Elasticsearch we should requeue to sync the annotation
```
sebgl, line 125: `is updated`

barkbay:
Jenkins test this please

PR #2893, opened 2020-04-16
Authenticate to GCP using an env var in the ocp deployer

sebgl:
The openshift-installer accepts a `$GCLOUD_KEYFILE_JSON` environment
variable to login to GCP.

Using it instead of the default gcloud auth avoids messing up my local
gcloud auth when using a gcp service account retrieved from Vault.

### DIFF
--- a/hack/deployer/runner/ocp.go
+++ b/hack/deployer/runner/ocp.go
@@ -182,7 +182,6 @@ func (d *OcpDriver) Execute() error {
 }
 
 func (d *OcpDriver) auth() error {
-	_ = NewCommand(fmt.Sprintf("gcloud config set project %s", d.ctx["GCloudProject"])).Run()
 
 	if d.plan.ServiceAccount {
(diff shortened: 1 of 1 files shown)


hack/deployer/runner/ocp.go
```diff
@@ -182,7 +182,6 @@ func (d *OcpDriver) Execute() error {
 }
 
 func (d *OcpDriver) auth() error {
-	_ = NewCommand(fmt.Sprintf("gcloud config set project %s", d.ctx["GCloudProject"])).Run()
```
flaper87, line 185: Curious why you moved this to the bottom of the function.

sebgl: The json cred file already contains the region, so we don't want to temper the existing gcloud config on the host if we're using the JSON creds file.

sebgl:
I'll merge then trigger the CI job. Will revert if it does not work as expected.

PR #2900, opened 2020-04-17
Fix GCloud auth to work on CI

sebgl:
* Revert #2893, which worked fine for `openshift-install` but not for `gsutil`
* Instead, set `$CLOUDSDK_CONFIG`, pointing to a directory used by `gcloud auth` to store credentials files, and by `gcloud` and `gsutil` commands to authenticate to gcp. This is done in order to not tamper the default gcloud credentials set on the system (eg. my own). It can be enabled or disabled with the `useNonDefaultCloudSdkPath` plan setting.


hack/deployer/runner/ocp.go
```diff
@@ -182,38 +182,30 @@ func (d *OcpDriver) Execute() error {
...
 	}
-
-	if len(accounts) > 0 {
-		return nil
+	if err := NewCommand(fmt.Sprintf("gcloud config set project %s", d.ctx["GCloudProject"])).Run(); err != nil {
+		return err
 	}
-
-	_ = NewCommand(fmt.Sprintf("gcloud config set project %s", d.ctx["GCloudProject"])).Run()
-	return NewCommand("gcloud auth login").Run()
```
flaper87, line 216: Is the login action redundant? I remember it being required regardless. Did it cause any issues for you?

hack/deployer/runner/ocp.go
```diff
@@ -182,38 +182,30 @@ func (d *OcpDriver) Execute() error {
 }
 
 func (d *OcpDriver) auth() error {
+	log.Println("Authenticating as service account...")
 
-	if d.plan.ServiceAccount {
```
flaper87, line 186: Removing this line makes accessing vault (and having the service account stored in vault) always required, which means everyone will be using `OcpVaultPath` unless it's configured/overwritten.

hack/deployer/runner/ocp.go
sebgl: `gcloud auth activate-service-account ` seems good enough when using a service account

hack/deployer/runner/ocp.go
sebgl: Yeah I think you have a point here, I'll revert this commit. The switch on `plan.ServiceAccount ` is confusing since using a service account is mandatory so it's a bit confusing. But indeed I can see how it could be useful to use the local gcloud auth if set up correctly.

sebgl:
Jenkins test this please.

hack/deployer/runner/ocp.go
sebgl: Hmm actually we require gcloud auth in a different place: https://github.com/elastic/cloud-on-k8s/blob/master/Makefile#L376-L378  :(
It's a bit weird to have a dependency from this Makefile to gcloud auth in the deployer.

sebgl:
Hmm actually we require gcloud auth in a different place: https://github.com/elastic/cloud-on-k8s/blob/master/Makefile#L376-L378 :(
It's a bit weird to have a dependency from this Makefile to gcloud auth in the deployer. I'll see if I can find a way to fix this :/

sebgl:
I added an additional plan setting: `useNonDefaultCloudSdkPath`. If `true`, we'll set the `CLOUDSDK_CONFIG` env var to a custom value. If false (or unset), we'll use gcloud defaults.
The goal is to not mess up with existing gcloud auth when used locally (`ocp-dev`), and use defaults when used in CI (`ocp-ci`).

hack/deployer/config/plans.yml
```diff
@@ -75,11 +75,12 @@ plans:
...
   provider: ocp
   machineType: n1-standard-8
-  serviceAccount: false
+  serviceAccount: true
   ocp:
     region: europe-west1
     nodeCount: 3
-    overwriteDefaultKubeconfig: false
+    overwriteDefaultKubeconfig: true
+    useNonDefaultCloudSdkPath: true
```
sebgl, line 83: I think those are the defaults we'll want when deploying ocp clusters for dev most of the times.

hack/deployer/runner/ocp.go
```diff
@@ -198,8 +198,19 @@ func (d *OcpDriver) auth() error {
...
+			// ensure gcloud & gsutil rely on credentials stored in gcpDir instead of using the default
+			// directory (~/.config/gcloud), to not tamper any default gcloud auth already set on the system
+			log.Printf(fmt.Sprintf("Setting $CLOUDSDK_CONFIG=%s", gcpDir))
+			if err := os.Setenv("CLOUDSDK_CONFIG", gcpDir); err != nil {
+				return err
+			}
+		}
+		if err := NewCommand(fmt.Sprintf("gcloud config set project %s", d.ctx["GCloudProject"])).Run(); err != nil {
+			return err
+		}
```
flaper87, line 212: I think we can move this to the top of the function and remove the call in L226. I believe this is how it was before the previous PR.

sebgl: It should only be done after we set `$CLOUDSDK_CONFIG` otherwise we're modifying the system defaults gcloud config?

flaper87: A-ha, fair!

Would you mind adding a comment explaining why this call is here? Otherwise, someone may try to deduplicate this by doing what I suggested without knowing the consequences. :smile:

sebgl: Sure!

sebgl:
I think we still have a problem with registry credentials for the ocp deployer: https://github.com/elastic/cloud-on-k8s/blob/42b622f2466ad9bb213d0e46d9093da00d7e450e/Makefile#L335-L338
But that should be fixed with @thbkrkr work on the registries: https://github.com/elastic/cloud-on-k8s/pull/2901

PR #2901, opened 2020-04-17
Use a single Docker registry

### LINKED ISSUE #2622: cloud-on-k8s-e2e-tests-ocp can no longer push images
The merge of #2502 causes a regression that prevents the`cloud-on-k8s-e2e-tests-ocp` job from pushing Docker images.

It is 07824ff64 that breaks it up by splitting the `ci-e2e` target (note: `docker-push` is no longer a subtask of it).

I had identified that `docker-push` and `e2e-docker-push` implicitly obtains access to the GCR docker registry via the `run-deployer`. I knew it was bad. I did not dare to question it at that time because there were already many other changes in the PR.

For OCP, it is currently `docker-push` that takes care of configuring the Docker auth (c17a127e7b3cd). I found this suspicious but [didn't really understand why](https://github.com/elastic/cloud-on-k8s/pull/2432#discussion_r377648088).

thbkrkr:
For the CI, we use 3 different Docker images: for the [operator](https://github.com/elastic/cloud-on-k8s/blob/master/Dockerfile), the [E2E tests](https://github.com/elastic/cloud-on-k8s/blob/master/test/e2e/Dockerfile) and the [CI build tools](https://github.com/elastic/cloud-on-k8s/blob/master/.ci/Dockerfile). So far we have tried to use the Docker registry provided by the cloud provider which we are testing as much as possible. With expanding the testing to more providers we have to handle the specifities of each one especially for authentication.

This commit simplifies this by using a single Docker registry for all of our uses.

Resolves #2622.


thbkrkr:
jenkins test this please

thbkrkr:
jenkins test this please

thbkrkr:
jenkins test this please

thbkrkr:
jenkins test this please

thbkrkr:
jenkins test this please

thbkrkr:
jenkins test this please

Makefile
```diff
@@ -322,33 +316,18 @@ switch-eks:
 ##  --    Docker images    --  ##
 #################################
 
+docker-login:
+	@ docker login -u $(ELASTIC_DOCKER_LOGIN) -p $(ELASTIC_DOCKER_PASSWORD) push.docker.elastic.co 2> /dev/null
```
pebrc, line 320: For local dev use: how is the `ELASTIC_DOCKER_PASSWORD` being set up? If this is a manual step each dev has do upfront we should document it. But I wonder if we can have something automated?

thbkrkr: Good point. `ELASTIC_DOCKER_PASSWORD` is defined in the file `.registry.env` 
when `make -C .ci get-docker-creds` is executed.

It's fairly easy to automate:
```diff
diff --git a/Makefile b/Makefile
index b6356b8e..5cc5fc63 100644
--- a/Makefile
+++ b/Makefile
@@ -316,7 +316,10 @@ switch-eks:
 ##  --    Docker images    --  ##
 #################################
 
-docker-login:
+.registry.env:
+       make -sC .ci get-docker-creds
+
+docker-login: .registry.env
```

thbkrkr: Otherwise, another way of looking at that is to keep the registry configurable to be able to easily use at least GCE as today for dev needs.

I seem to introduce a kind of regression by removing this feature.

So far I did not manage a nice way to do this by reconciling simplicity given the specifics of each registry, which is the purpose of this task: simplify.
- the authentication is different: docker login or gcloud auth
- Elastic registry needs a retag with the push prefix
- different namespaces for the images (eck, eck-snapshots, eck-ci) vs gcloud_project for GCE

I might have something reasonable this morning.

pebrc: Would it make sense to have a `.sentenvconfig` for local dev use against the Elastic docker registry, that does the necessary setup?

thbkrkr:
We discussed with the team that there is a lack of good support for the development workflow as pointed in this [discussion](https://github.com/elastic/cloud-on-k8s/pull/2901#discussion_r414391632).

We should still be able to use GCR or a different namespace from that used by the CI in the Elastic Docker registry.

Latest commits address that.

thbkrkr:
I found several weaknesses in the implemented solution. Do not do another code review right away.

thbkrkr:
I think this is ready for a final review.

Since the last review, there is no fundamental change:

- Rename REPOSITORY in NAMESPACE
- Introduce E2E_NAMESPACE and use REGISTRY in E2E_IMG
- Renamings for the propagation of the CI docker creds
- Use IMG_SUFFIX instead of IMG_NAME override
- Add targets to switch between dev and gcr registries
- Move docker login/push logic in a bash script - not ideal but not worse that we had in the Makefile

One weakness that already exists and that remains is that the content of the `.registry.env` (e.g.: `REGISTRY = eu.gcr.io`) can be ignored. If the operator image is defined in the `.env` (e.g.: OPERATOR_IMAGE = docker.elastic.co/eck-snapshots/eck-operator:1.1.0-2020-05-07-7df50bca`), it takes the precedence over and this is the one that is used.

Makefile
thbkrkr: Yes it makes sense but I think it's better to keep the two things orthogonal.

You might want to reproduce something that happens in a given job (pr, e2e/master, e2e/stack-versions, etc) using a given registry (elastic/ci, elastic/dev, gcr).

Makefile
```diff
@@ -330,61 +324,43 @@ docker-build: go-generate
...
-endif
-endif
+	@ hack/ci/docker-push.sh $(OPERATOR_IMAGE)
 
 purge-gcr-images:
 	@ for i in $(gcloud container images list-tags $(BASE_IMG) | tail +3 | awk '{print $$2}'); \
 		do gcloud container images untag $(BASE_IMG):$$i; \
 	done
 
+switch-registry-gcr:
```
sebgl, line 334: Should we mention the `.registry.env` file in `dev-setup.md`?

hack/ci/docker-push.sh
```diff
@@ -0,0 +1,70 @@
...
+# Log in to push.docker.elastic.co if the namespace eck, eck-ci or eck-snapshots is used
+# Log in to gcloud if GCR is used
+# Add a 'push.' prefix when using docker.elastic.co
+
+set -euo pipefail
+
+# source variables if present
+if [[ -f .registry.env ]]; then
+    # shellcheck disable=SC2046
+    export $(sed "s|\s*=\s*|=|g" .registry.env)
```
sebgl, line 18: This seems to break for me:

```
hack/ci/docker-push.sh: line 18: export: `=': not a valid identifier
hack/ci/docker-push.sh: line 18: export: `eu.gcr.io': not a valid identifier
hack/ci/docker-push.sh: line 18: export: `=': not a valid identifier
hack/ci/docker-push.sh: line 18: export: `elastic-cloud-dev': not a valid identifier
hack/ci/docker-push.sh: line 18: export: `=': not a valid identifier
hack/ci/docker-push.sh: line 18: export: `elastic-cloud-dev': not a valid identifier
```

could be a Mac OS thing :(

sebgl: `sed "s|[[:space:]]*=[[:space:]]*|=|g" .registry.env` seems to work for me

hack/ci/docker-push.sh
```diff
@@ -0,0 +1,70 @@
+#!/usr/bin/env bash
```
sebgl, line 1: since this does not only apply to CI I would move it to `hack/docker-push.sh`

Makefile
```diff
@@ -31,26 +32,19 @@ endif
...
-	# default to local registry
-	REGISTRY ?= localhost:5000
-endif
-
 # for dev, suffix image name with current user name
 IMG_SUFFIX ?= -$(subst _,,$(shell whoami))
 
-REPOSITORY  ?= eck
+REGISTRY    ?= docker.elastic.co
+NAMESPACE   ?= eck-snapshots
```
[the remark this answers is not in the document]
thbkrkr: I was very hesitant about this. 
Let's make `eck-dev` the default.
A fresh git clone should be ready for development.

Makefile
```diff
@@ -31,26 +32,19 @@ endif
...
-	# default to local registry
-	REGISTRY ?= localhost:5000
-endif
-
 # for dev, suffix image name with current user name
 IMG_SUFFIX ?= -$(subst _,,$(shell whoami))
 
-REPOSITORY  ?= eck
+REGISTRY    ?= docker.elastic.co
+NAMESPACE   ?= eck-dev
```
sebgl, line 39: Should we name it `REGISTRY_NAMESPACE` (and `E2E_REGISTRY_NAMESPACE`)  to avoid confusion with k8s namespaces?

thbkrkr: Good idea.

PR #2889, opened 2020-04-16
Fix trial license validation issues

pebrc:
Fixes #2878 

What this PR does:
* introduce concept of trial activation where we keep the trial key around until trial successfully activated
* fixes erroneous validation messages as seen in #2878 
* fixes a situation where an operator crash or a k8s API error would prevent users from ever successfully starting a trial (not yet observed in the wild but possible)

PR #2907, opened 2020-04-17
Fix trial license validation issues (#2889)

pebrc:
* introduce concept of trial activation where we keep the trial key around until trial successfully activated
 * fixes erroneous validation messages as seen in e2e tests
 * fixes a situation where an operator crash or a k8s API error would prevent users from ever successfully starting a trial (not yet observed in the wild but possible)

Backport of #2889


pebrc:
Jenkins test this please

PR #2916, opened 2020-04-20
Close client after noop observer comparisons

### LINKED ISSUE #2911: Elasticsearch client leak
There is an Elasticsearch client leak in the Elasticsearch driver:

https://github.com/elastic/cloud-on-k8s/blob/e0773830c8ab5a40a2881e6c7bc1226d89da1520/pkg/controller/elasticsearch/driver/driver.go#L165-L173

In most cases the client created for the observer is never closed, leading to a memory leak.

pebrc:
Fixes #2911 

Close newly created Elasticsearch client if observer comparison results in the new client not being used.

### DIFF
--- a/pkg/controller/elasticsearch/observer/manager.go
+++ b/pkg/controller/elasticsearch/observer/manager.go
@@ -49,6 +49,7 @@ func (m *Manager) Observe(cluster types.NamespacedName, esClient client.Client) 
 		m.StopObserving(cluster)
 		return m.createObserver(cluster, esClient)
 	default:
+		esClient.Close()
 		return observer
 	}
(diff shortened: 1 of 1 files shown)


charith-elastic:
I think that instead of passing in a fully instantiated client that gets thrown away most of the time, we should pass a function that creates the client if needed only. This is because creating the client still requires some work like adding the certificates to a trust chain and creating a new HTTP client etc.

pebrc:
@charith-elastic should we still merge this as a first remediation for 1.1 and follow up with your proposed fix for 1.2 as this would require us to rethink how we do the client comparison?

PR #2920, opened 2020-04-21
Copy apmelasticsearch round_tripper to work around goroutine leak

pebrc:
Copy the apmelasticsearch roundtripper to work around a resource leak due to the missing propagation of `CloseIdleConnections`


pkg/controller/common/tracing/round_tripper.go
```diff
@@ -0,0 +1,252 @@
...
+	"path"
+	"strings"
+	"sync/atomic"
+	"unsafe"
+
+	"go.elastic.co/apm"
+	"go.elastic.co/apm/module/apmhttp"
+)
+
+// This purely exists as workaround until the issue around closing connections in apm-go-agent is fixed.
```
sebgl, line 23: Can you highlight what is the original code and what is the fix?

pebrc: Yes absolutely.

pebrc:
The APM fix got already merged so we have alternatively the option to upgrade the APM libraries in #2921

pebrc:
Closing in favor of #2921

PR #2923, opened 2020-04-21
Upgrade apm agent to latest master commit (#2921)

pebrc:
This upgrades the APM agent to an unreleased but fixed version which propagates CloseIdleConnection calls to the underlying transport. Integrating this change should fix the goroutine leak we have observed.

Backport of #2921


barkbay:
jenkins test this please

PR #2924, opened 2020-04-21
Cleanup e2e debug clusters if successful only

sebgl:
A quickfix to only clean up k8s clusters when the job was succesful,
since our logic of checking failed tests seems to fail. For temporary e2e debug tests only.

### DIFF
--- a/.ci/pipelines/e2e-tests-testdeleteservices.Jenkinsfile
+++ b/.ci/pipelines/e2e-tests-testdeleteservices.Jenkinsfile
@@ -72,7 +72,7 @@ pipeline {
     post {
         success {
             build job: 'cloud-on-k8s-e2e-cleanup',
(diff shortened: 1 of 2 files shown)


.ci/pipelines/e2e-tests-testdeleteservices.Jenkinsfile
```diff
@@ -70,6 +70,11 @@ pipeline {
     }
 
     post {
+        success {
+            build job: 'cloud-on-k8s-e2e-cleanup',
+                parameters: [string(name: 'JKS_PARAM_GKE_CLUSTER', value: "eck-debug-e2e-${BUILD_NUMBER}")],
```
anyasabo, line 75: this needs to be `eck-debug-endpoints-e2e-${BUILD_NUMBER}"` right?

PR #2927, opened 2020-04-22
Avoid unnecessary bcrypt computation

charith-elastic:
Noticed during profiling that the association controller was spending a lot of time on a bcrypt computation that gets thrown away most of the time. This patch avoids the computation if it is not necessary.

Also includes linter fixes to remove duplicate imports in `cmd/manager/main.go` and `pkg/controller/enterprisesearch/config_test.go`.

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -24,7 +24,6 @@ import (
 	"github.com/elastic/cloud-on-k8s/pkg/controller/apmserver"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/association"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/association/controller"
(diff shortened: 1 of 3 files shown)


cmd/manager/main.go
```diff
@@ -347,7 +346,7 @@ func execute() {
 		log.Error(err, "unable to create controller", "controller", "EnterpriseSearch")
 		os.Exit(1)
 	}
-	if err = associationctl.AddApmES(mgr, accessReviewer, params); err != nil {
+	if err = controller.AddApmES(mgr, accessReviewer, params); err != nil {
```
pebrc, line 349: I wonder if the `associationctl` alias was not better and we should instead remove the other import.

PR #2930, opened 2020-04-22
Avoid unnecessary bcrypt computation

charith-elastic:
Backport of #2927


charith-elastic:
After discussing with @anyasabo, I am closing this backport because it's a logic change too close to the release date.

PR #2934, opened 2020-04-22
Use enterprise_trial as the trial license type

### LINKED ISSUE #2915: Inconsistent naming of trial licenses in ECK/ECE
ECE uses `enterprise_trial` 
ECK uses `enterprise-trial` 😞 

With the current implementation this naming inconsistency actually allows externally generated (trial) licenses to work with ECK because we don't recognise them as trial licenses and treat them as regular Enterprise licenses. 

We should: 
* adopt the naming scheme `enterprise_trial` 
* make sure self-generated trials as well as externally generated trials both work 
* support `enterprise-trial` for backwards compatibility (but deprecate it)

pebrc:
Fixes #2915 

* use and expect `enterprise_trial` as the correct license type for trial licenses both auto-generated ones as well as externally generated ones 
* support `enterprise-trial` for backwards compatibility in auto-generated trial licenses
* fix an issue with the license checker which excluded trials from being checked

Labeled 1.1.0, but open for discussion. Happy to spin off a smaller PR if desired for 1.1.0


pkg/controller/common/license/model.go
```diff
@@ -69,7 +72,12 @@ func (l EnterpriseLicense) IsValid(instant time.Time) bool {
 
 // IsTrial returns true if this is a self-generated trial license.
 func (l EnterpriseLicense) IsTrial() bool {
-	return l.License.Type == LicenseTypeEnterpriseTrial
+	return l.License.Type == LicenseTypeEnterpriseTrial || l.License.Type == LicenseTypeLegacyTrial
+}
+
+// IsAutoGenerated returns true if this license has been generated by trial controller or is eligible to be turned into one.
+func (l EnterpriseLicense) IsAutoGenerated() bool {
```
sebgl, line 79: could we rename this to `IsTrialCandidate`?

pkg/controller/common/license/trial.go
```diff
@@ -21,6 +21,8 @@ import (
 )
 
 const (
+	AutogenLicenseIssuer = "Elastic k8s operator"
```
sebgl, line 24: Maybe `ECKLicenseIssuer`? `autogen` feels like there is some kind of magic I need to think about.

pkg/controller/common/license/check_test.go
```diff
@@ -39,27 +56,48 @@ func TestChecker_EnterpriseFeaturesEnabled(t *testing.T) {
...
+			wantErr: false,
+		},
+		{
+			name: "invalid trial: FAIL",
+			fields: fields{
+				initialObjects: []runtime.Object{asRuntimeObject(emptyTrialLicenseFixture), &statusSecret},
+			},
+			want:    false,
+			wantErr: false,
+		},
```
sebgl, line 87: Can we add a test for expired trial licenses?

pkg/controller/common/license/model.go
pebrc: Hm I wanted to capture that these trials originate from inside the operator. There are externally generated trials as well the we are not interested here that are also `TrialCandidates` or maybe I am misunderstandig.

sebgl: Got it. I think `IsAutoGenerated` does not reflect the function is interested in trial licenses.
Maybe something like `IsECKManagedTrial`?

pebrc: Yes that's good!

docs/operating-eck/licensing.asciidoc
```diff
@@ -36,7 +36,7 @@ At the end of the trial period, the Platinum and Enterprise features operate in
 
 [float]
 == Add a license
-If you have a valid Enterprise subscription you will receive a license as a JSON file.
+If you have a valid Enterprise subscription or you have been granted a trial license extension you will receive a license as a JSON file.
```
anyasabo, line 39:
```suggestion
If you have a valid Enterprise subscription or you have been granted a trial license extension, you will receive a license as a JSON file.
```

cmd/licensing-info/main.go
```diff
@@ -35,7 +36,10 @@ import (
 //
 
 func main() {
-	licensingInfo, err := license.NewResourceReporter(newK8sClient()).Get()
+	var operatorNamespace string
+	flag.StringVar(&operatorNamespace, "operator-namespace", "elastic-system", "indicates the namespace where the operator is deployed")
+	flag.Parse()
+	licensingInfo, err := license.NewResourceReporter(newK8sClient(), operatorNamespace).Get()
```
pebrc, line 42: I am not sure we need still need this tool.

pkg/license/reporter.go
```diff
@@ -25,43 +25,44 @@ type ResourceReporter struct {
...
+func NewResourceReporter(client client.Client, operatorNs string) ResourceReporter {
 	c := k8s.WrapClient(client)
 	return ResourceReporter{
 		aggregator: Aggregator{
 			client: c,
 		},
 		licensingResolver: LicensingResolver{
-			client: c,
+			client:     c,
+			operatorNs: operatorNs,
```
pebrc, line 36: This was always "" up until now, which lead to errors not that we actually to trial license checks.

cmd/licensing-info/main.go
anyasabo: Agreed, maybe open an issue to revisit keeping it in there? I'm not sure how valuable it is

pkg/controller/common/license/model.go
```diff
@@ -69,7 +72,12 @@ func (l EnterpriseLicense) IsValid(instant time.Time) bool {
 
 // IsTrial returns true if this is a self-generated trial license.
 func (l EnterpriseLicense) IsTrial() bool {
-	return l.License.Type == LicenseTypeEnterpriseTrial
+	return l.License.Type == LicenseTypeEnterpriseTrial || l.License.Type == LicenseTypeLegacyTrial
+}
+
+// IsECKManagedTrial returns true if this license has been generated by trial controller or is eligible to be turned into one.
```
anyasabo, line 78: Nit: "or is eligible to be turned into one." is kind of confusing, but I'm not sure what would be better. Maybe "or is eligible to have its license generated by the trial controller" but that still reads weird. Maybe just "or is eligible to be"? All sound a little bit off

cmd/licensing-info/main.go
thbkrkr: I left it because it was handy for testing the license reporting and in case someone wanted to get it in real time but it is not documented. So I don't really see any good reason to keep it.

PR #2935, opened 2020-04-22
Set kubernetes version for kind e2e test results reporting

david-kow:
Start passing version information about kubernetes used in kind runs. Patch part omitted for consistency with GKE where we only provider major and minor.


david-kow:
jenkins test this please

PR #2942, opened 2020-04-23
Check for ECK managed trials when matching licenses to clusters

pebrc:
Fixes a regression introduced with the notion of ECK-managed trials in #2934 😞 

The license matching code would always issue a trial license to clusters even if the orchestration level trial license contained short lived platinum or gold licenses.

For users this would mean if they received a trial extension from Elastic this trial extension would not propagate to the individual Elasticsearch clusters (which would still be on the expired trial)

With this PR we  check for ECK managed trials when handing out licenses to clusters and treat trial extensions just like regular licenses (which they are). 

I did not catch this in the original PR because the trial was applied correctly on the operator level and a unit test we had in place for testing the license matching used trial examples that don't exist in that form in reality. This PR also fixes that unit test. 

We don't have an e2e test for the trial extension functionality (yet) which could have caught it, because trial extensions are hard to set up and expire quickly in a CI environment.


pebrc:
Jenkins test this please

anyasabo:
>We don't have an e2e test for the trial extension functionality (yet) which could have caught it, because trial extensions are hard to set up and expire quickly in a CI environment.

I'd think we could work around this with a pre-generated trial and fake clock, but not necessary to merge this PR in I don't think.

PR #2960, opened 2020-04-26
Add e2e test for trial extensions

pebrc:
This closes a gap in the e2e test coverage simulating the trial extension use case.
This test can only run successfully against development builds of the Elastic Stack
due to the fact that a test license is generated on the fly as part of the test.

### More details about the changes in this PR: 

* refactor the license signer code to be able to also sign Elasticsearch licenses
* write license signing code for Elasticsearch (this is in the e2e tests to clearly indicate that is purely intended for testing)
   * why not use the tooling Elasticsearch has for generating licenses? Because it would force us to build parts of the Elasticsearch repo, run the tool which is JVM based in a pod somewhere to generate the cluster license. The chosen solution works with minimal modification to our existing signing code instead and allows us to generate a working license in Go.
* wire up an e2e test to pull a development private key for license signing from vault (needs to be same the Elasticsearch team uses for all of this to work, an encrypted version of the key can be found [here](https://github.com/elastic/elasticsearch/blob/master/x-pack/license-tools/src/test/resources/private.key))
* generate a development version of an Enterprise trial license with a development version of a cluster license in it.
* test that we can
   * apply this license 
   * propagate the cluster license to a test cluster
   * that we prefer the trial extension over the ECK managed trial license if both are valid.
* run this test only as part of our Elastic stack snapshot tests


test/e2e/cmd/run/command.go
```diff
@@ -81,6 +82,7 @@ func Command() *cobra.Command {
 	cmd.Flags().BoolVar(&flags.skipCleanup, "skip-cleanup", false, "Do not run cleanup actions after test run")
 	cmd.Flags().StringVar(&flags.testContextOutPath, "test-context-out", "", "Write the test context to the given path")
 	cmd.Flags().StringVar(&flags.testLicense, "test-license", "", "Test license to apply")
+	cmd.Flags().StringVar(&flags.privateKeyPath, "private-key-path", "", "Path to private key to generate test licenses")
```
sebgl, line 85: I would make it a bit more explicit, I think: `test-license-pkey-path`

test/e2e/es/license_test.go
```diff
@@ -127,3 +129,61 @@ func TestEnterpriseTrialLicense(t *testing.T) {
...
 }
+
+// TestEnterpriseTrialExtension tests that trial extensions can be successfully applied and take effect.
+func TestEnterpriseTrialExtension(t *testing.T) {
+	if test.Ctx().PrivateKeyPath == "" {
+		// skip this test if the dev private key is not configured e.g. because we are testing a production build
+		t.SkipNow()
+	}
+	privateKeyBytes, err := ioutil.ReadFile(test.Ctx().PrivateKeyPath)
+	require.NoError(t, err)
```
sebgl, line 140: aren't we reading the content of that file in `e2e/cmd/run/run.go` already? Can we pass it in the context directly?

test/e2e/es/license_test.go
```diff
@@ -127,3 +129,61 @@ func TestEnterpriseTrialLicense(t *testing.T) {
 
 	test.Sequence(initStepsFn, stepsFn, esBuilder).RunSequential(t)
 }
+
+// TestEnterpriseTrialExtension tests that trial extensions can be successfully applied and take effect.
+func TestEnterpriseTrialExtension(t *testing.T) {
```
sebgl, line 134: Can you add some comments describing what this test is doing? (it manipulates both ECK & ES licenses IIUC, which is rather complicated)

test/e2e/es/license_test.go
pebrc: Remember this runs in a k8s job. `run.go` is settting up that job, copying the key into a secret that is mounted in the job's pod. 

Also context is dumped in logs, it's a test key but still we should not put sensitive data in the context.

sebgl: Oh right, I missed that 👍

test/e2e/es/license_test.go
```diff
@@ -127,3 +129,65 @@ func TestEnterpriseTrialLicense(t *testing.T) {
 
 	test.Sequence(initStepsFn, stepsFn, esBuilder).RunSequential(t)
 }
+
+// TestEnterpriseTrialExtension tests that trial extensions can be successfully applied and take effect.
+// Generates a development version of an Enterprise trial extension license with a development Elasticsearch license inside.
+// Then tests that ECK accepts this license and propagates the Elasticsearch license to the test Elasticsearch cluster.
+// Also tests that this trial extension license is preferred by ECK over any ECK-managed trial.
+// Finally tests that trial extensions can applied repeatedly as opposed to ECK-managed trials which are one-offs.
```
sebgl, line 137: `can be applied`

PR #2985, opened 2020-04-28
Whitelist renovate

anyasabo:
This should allow renovate's PRs to run automatically rather than us telling jenkins to run it every time. I'm going by the syntax in the Kibana CI config.

Checked the user in the github API:

```
{
  "url": "https://api.github.com/repos/elastic/cloud-on-k8s/pulls/2982",
  "id": 410304773,
  "node_id": "MDExOlB1bGxSZXF1ZXN0NDEwMzA0Nzcz",
  "html_url": "https://github.com/elastic/cloud-on-k8s/pull/2982",
  "diff_url": "https://github.com/elastic/cloud-on-k8s/pull/2982.diff",
  "patch_url": "https://github.com/elastic/cloud-on-k8s/pull/2982.patch",
  "issue_url": "https://api.github.com/repos/elastic/cloud-on-k8s/issues/2982",
  "number": 2982,
  "state": "open",
  "locked": false,
  "title": "Update golang.org/x/crypto commit hash to 729f1e8",
  "user": {
    "login": "renovate[bot]",
    "id": 29139614,

```


anyasabo:
I'm actually not sure if this is very useful. Because we include the version in the notice file, but renovate cannot update the notice file, the first PR renovate opens up will always fail CI. We need to pull down the branch, regen the notice file, and push a commit (and I think at that point jenkins will test the PR automatically?).

charith-elastic:
Renovate should be able to generate the notice file itself: https://docs.renovatebot.com/configuration-options/#postupgradetasks

anyasabo:
It looks like we would need to switch from the app to self hosted to be able to use post upgrade tasks, there's no way to set the trustLevel in the github app:
https://github.com/renovatebot/renovate/issues/838#issuecomment-581821049

charith-elastic:
That's a shame. :disappointed: 

I am a bit disappointed by how Renovate seems to operate. I was expecting it to raise PRs only when there are known vulnerabilities. While it's nice to have always up-to-date dependencies, the PR-per-dependency approach feels a bit too intense. Is it possible to group all updates into one PR instead? I think that will make it easier to review, test, and bisect in case of a regression.  Also, Kubernetes dependencies are notoriously finicky -- so always getting the latest and greatest might not be that simple nor desired.

anyasabo:
I think I actually prefer the one PR per dependency as it makes debugging easier if there are issues. I think it's just especially noisy right now because we have so many dependencies that need updates. It might make sense to do one big dependency update PR, and then let renovate trickle in from there?

We can exclude upgrades that we don't want by closing the PRs it opens. For k8s dependencies I think it may make sense to use their packageRules config option and limit it to only patch releases. We can manually upgrade those (or maybe the kubebuilder upgrades will catch them). WDYT?

charith-elastic:
:+1: on K8S dependency policy.

I am not strictly opposed to PR-per-dependency but my personal preference would be for a more structured approach<sup>++</sup> so as to avoid cluttering the history with spurious dependency update commits. Also, I think a grouped PR makes it slightly easier to discover surprising interactions between two dependencies sooner rather than later. 
If post upgrade tasks are not configurable with the current setup (:disappointed:), then at least the manual work involved would be limited to a single PR as well.

<sup>++</sup> Automated PR at a regular interval (once a week, once a fortnight) or a  PR check that prompts the author to upgrade dependencies.

anyasabo:
I updated this PR to also ignore major k8s dependencies. The config docs are [here](https://docs.renovatebot.com/configuration-options/#packagerules), and I ran the config validator they provide, so I think it's right?

PR #3020, opened 2020-04-30
Add PodTemplate semantic validation for Elasticsearch

### LINKED ISSUE #2266: Reconciliation is blocked when a Pod can't be created
I got a situation where:
1. Pod could not be created because of some invalid values in the spec
2. `ok, err := d.expectationsSatisfied()` always returns `false`
3. It is impossible for the user to recover from that situation

The following manifest helps to recreate the issue:

```yaml
apiVersion: elasticsearch.k8s.elastic.co/v1beta1
kind: Elasticsearch
metadata:
  name: blocked
spec:
  version: 7.5.0
  nodeSets:
    - name: default
      count: 1
      config:
        node.store.allow_mmap: false
      podTemplate:
        spec:
          volumes:
            - name: BaDVolumeName
              emptyDir: {}
```

The problem is that we assume that a Pod is at least in a `Pending` state, which is not the case.

barkbay:
This PR adds a semantic validation for the content of the `podTemplate` field of the Elasticsearch resources.

It is achieved by performing a dry-run request which attempts to create a Pod reusing the content of the `podTemplate`. The validation is only performed when a `StatefulSet` is created or updated.

Fixes #2266

### DIFF
--- a/pkg/controller/elasticsearch/sset/validation.go
+++ b/pkg/controller/elasticsearch/sset/validation.go
@@ -33,7 +33,10 @@ func (e *PodTemplateError) Error() string {
 	)
 }
 
-// validatePodTemplate validates a Pod Template by issuing a dry API request.
+// validatePodTemplate validates a Pod Template by issuing a dry-run API request.
+// This check is performed as "best-effort" for the following reasons:
+// * It is only supported by the API server starting 1.13
+// * There might be some admission webhooks on the validation path that are not compatible with dry-run requests.
 func validatePodTemplate(
 	c k8s.Client,
 	parent metav1.Object,
--- a/pkg/controller/elasticsearch/sset/validation.go
+++ b/pkg/controller/elasticsearch/sset/validation.go
@@ -69,12 +69,7 @@ func toPodTemplateError(
 		return err
 	}
 
-	switch statusErr.ErrStatus.Reason {
-	case metav1.StatusReasonBadRequest:
-		// Dry run is beta and available since Kubernetes 1.13
-		// Openshift 3.11 and K8S 1.12 don't support dryRun but returns "400 - BadRequest" in that case
-		return nil
-	case metav1.StatusReasonInvalid:
+	if statusErr.ErrStatus.Reason == metav1.StatusReasonInvalid {
 		// If the Pod spec is invalid the expected error is 422.
 		// Since "details" is a pointer let's check that it's not nil before going further.
 		if statusErr.ErrStatus.Details == nil {
@@ -87,6 +82,9 @@ func toPodTemplateError(
 			Causes:      statusErr.ErrStatus.Details.Causes,
 		}
 	}
-	// The Kubernetes API returns an error which is not related to the spec. of the Pod, let's retry again later
-	return statusErr
+
+	// The Kubernetes API returns an error which is not related to the spec. of the Pod. In order to not block
+	// the reconciliation loop we skip the validation.
+	log.Info("Pod validation skipped", "namespace", parent.GetNamespace(), "es_name", parent.GetName(), "error", statusErr)
+	return nil
 }
--- a/pkg/controller/elasticsearch/sset/validation.go
+++ b/pkg/controller/elasticsearch/sset/validation.go
@@ -1,0 +1,92 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License;
+// you may not use this file except in compliance with the Elastic License.
+
+package sset
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/elastic/cloud-on-k8s/pkg/utils/k8s"
+	appsv1 "k8s.io/api/apps/v1"
+	v1 "k8s.io/api/core/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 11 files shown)


barkbay:
The validation fails with the following error on GKE 1.14: `Internal error occurred: admission webhook "pod-ready.common-webhooks.networking.gke.io" does not support dry run`

I think it means that the validation proposed here should only be a "best effort" validation: as long as the error is not explicitly a `metav1.StatusReasonInvalid` (422) we should not block the reconciliation loop.

pkg/controller/elasticsearch/sset/validation.go
```diff
@@ -0,0 +1,90 @@
...
+	return fmt.Sprintf(
+		"Validation of PodTemplate for StatefulSet %s in Elasticsearch %s/%s failed for the following reasons: %v",
+		e.StatefulSet.Name,
+		e.Parent.GetNamespace(),
+		e.Parent.GetName(),
+		e.Causes,
+	)
+}
+
+// validatePodTemplate validates a Pod Template by issuing a dry API request.
```
sebgl, line 36: dry -> dry-run?

test/e2e/es/podtemplate_test.go
```diff
@@ -7,20 +7,110 @@ package es
...
+// TestPodTemplateValidation simulates a cluster update with a temporary invalid pod template.
+// The invalid pod template should not prevent the update to be eventually applied.
+func TestPodTemplateValidation(t *testing.T) {
+
+	kubernetesVersion, err := semver.ParseTolerant(test.Ctx().KubernetesVersion)
+	if err != nil {
+		t.Fatalf(
+			"Failed to parse the Kubernetes version: %v, err is %v", test.Ctx().KubernetesVersion, err)
+	}
+	if kubernetesVersion.LT(semver.MustParse("1.15.0")) {
```
sebgl, line 37:
```
// Pod creation dry run is only supported starting 1.15
```

pkg/controller/elasticsearch/sset/validation.go
sebgl: `// This check is performed as "best-effort", since the dry-run API is only available starting K8s 1.15.0`

test/e2e/es/podtemplate_test.go
```diff
@@ -7,20 +7,110 @@ package es
...
+								},
+							}),
+						)
+					},
+				}).
+			WithSteps(
+				// Mutate Elasticsearch with an invalid PodTemplate
+				b.UpgradeTestSteps(k),
+			).WithStep(
+			test.Step{
```
sebgl, line 74: indentation feels a bit off here?

pkg/controller/elasticsearch/sset/validation.go
```diff
@@ -0,0 +1,90 @@
...
+// toPodTemplateError attempts to extract the meaningful information from the dry run API call by converting the original
+// error into a PodTemplateError.
+func toPodTemplateError(
+	parent metav1.Object,
+	sset appsv1.StatefulSet,
+	err error,
+) error {
+	var statusErr *k8serrors.StatusError
+	if !errors.As(err, &statusErr) {
+		// Not a Kubernetes API error (e.g. timeout)
```
sebgl, line 68: I'm wondering if we could always reach a timeout in some situations (eg. particular proxy configuration). Hard to know for sure, so I'm fine moving on with this.
Should we add some debug logs here just in case?

barkbay:
To explain a little bit more the issue with the admission webhook: as explained in the [documentation](https://kubernetes.io/docs/reference/using-api/api-concepts/#make-a-dry-run-request):

>  If the request would trigger an admission controller which would have side effects, the request will be failed rather than risk an unwanted side effect. [...] admission webhooks can declare in their configuration object that they do not have side effects by setting the sideEffects field to “None”. If a webhook actually does have side effects, then the sideEffects field should be set to “NoneOnDryRun”

In order for an admission webhook to state if it supports dry-run its `sideEffects` field must be set accordingly. If this field is not set the default value is `Unknown`. In this case the request fails with something like this: `admission webhook "iam-for-pods.amazonaws.com" does not support dry run`

```
$ k get MutatingWebhookConfiguration -o yaml
```
```yaml
apiVersion: v1
items:
- apiVersion: admissionregistration.k8s.io/v1
  kind: MutatingWebhookConfiguration
  metadata:
    name: pod-identity-webhook
  webhooks:
  - admissionReviewVersions:
    - v1beta1
    failurePolicy: Ignore
    matchPolicy: Exact
    name: iam-for-pods.amazonaws.com
     ....
    sideEffects: Unknown
```

Unfortunately on EKS it seems that the `pod-identity-webhook` has not been updated to support dry run request (tested on EKS 1.16 ).

Regarding GKE it seems that built-in admission webhooks are compatible since 1.15.

I will update the e2e test to ensure that it is only run on GKE for K8S > 1.15.

TL;DR: even if dry run requests have been promoted to beta in K8S 1.13 the admission webhooks are not all compatible with that feature 😢

test/e2e/es/podtemplate_test.go
barkbay: Agreed but this file is formatted with `go fmt` 🤷‍♂️

pkg/controller/elasticsearch/sset/validation.go
barkbay: I think that if we reach a timeout here it is either:
* a transient error, in this case it is fair to try again in the next reconciliation loop
* a permanent error, in which case the StatefulSet update/creation that should happen right after will also fail

The error is already reported in the controller log, I'm not sure we should duplicate it.

Issue #3034, opened 2020-05-11
Transport certs public secret name is not expected


sebgl:
Looks like I accidentally changed the name of the transport certs public certificates in https://github.com/elastic/cloud-on-k8s/commit/608642b3e35dc00fa394ae50df732ce486c50848.

It is now (as of `1.1.0`) named `<cluster>-es-ca-certs-public` whereas it used to be `<cluster-name>-es-transport-certs-public` before this commit (in `1.0.0`).

The docs for `1.1.0` mention the (now incorrect) `<cluster>-es-transport-certs-public` naming scheme.

I see 2 options here:
1. We change it back to `<cluster-name>-es-transport-certs-public` in a `1.1.1` version.
2. We change the docs to use `<cluster>-es-ca-certs-public` starting 1.1.0.

Considering we also have `<cluster-name>-es-http-certs-public`, I think the first option makes more sense.

barkbay:
I think I would be in favour of a rename to `<cluster-name>-es-transport-certs-public`
I don't think it is widely used yet and it would make more sense re. http certs.

PR #3035, opened 2020-05-11
Restore transport certs public secret naming scheme from 1.0

sebgl:
The name of the secret holding transport public certs was accidentally changed to
`<cluster>-es-ca-certs-public` in 608642b, but should have been kept
`<cluster-name>-es-transport-certs-public`.

Fixes https://github.com/elastic/cloud-on-k8s/issues/3034.

### DIFF
--- a/pkg/controller/common/certificates/secret.go
+++ b/pkg/controller/common/certificates/secret.go
@@ -33,15 +33,15 @@ const (
 )
 
 func InternalCertsSecretName(namer name.Namer, ownerName string) string {
-	return namer.Suffix(ownerName, "http", certsInternalSecretName)
+	return namer.Suffix(ownerName, string(HTTPCAType), certsInternalSecretName)
 }
 
 func PublicCertsSecretName(namer name.Namer, ownerName string) string {
-	return namer.Suffix(ownerName, "http", certsPublicSecretName)
+	return namer.Suffix(ownerName, string(HTTPCAType), certsPublicSecretName)
 }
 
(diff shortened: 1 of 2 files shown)


sebgl:
Users upgrading from 1.1.0 to 1.1.0 would still have a leftover `<cluster>-es-ca-certs-public ` secret. I don't know if it's worth explicitly looking for it and deleting it?

pkg/controller/common/certificates/secret.go
```diff
@@ -33,15 +33,15 @@ const (
...
 
 func PublicCertsSecretName(namer name.Namer, ownerName string) string {
-	return namer.Suffix(ownerName, "http", certsPublicSecretName)
+	return namer.Suffix(ownerName, string(HTTPCAType), certsPublicSecretName)
 }
 
-func PublicCASecretName(namer name.Namer, ownerName string) string {
-	return namer.Suffix(ownerName, "ca", certsPublicSecretName)
+func PublicTransportCertsSecretName(namer name.Namer, ownerName string) string {
+	return namer.Suffix(ownerName, string(TransportCAType), certsPublicSecretName)
```
sebgl, line 44: this is the actual fix (other changes are mostly cosmetic)

anyasabo:
>Regarding the leftover I think it's fine to keep the obsolete Secret as long as the documentation mentions the correct name (which is the case).

👍 , and the owner ref is still set so it will still be cleaned up

Issue #3040, opened 2020-05-12
[META] Beat CRD


david-kow:
ECK will support a new Beat CRD that enables to deploy and manage any Beat.

### CRD

Spec (WIP):

```yaml
apiVersion: beat.k8s.elastic.co/v1beta1
kind: Beat
metadata: ...
spec:
  type: filebeat

  elasticsearchRef:
    name: quickstart
    namespace: default
  version: 7.6.0
  image: ...
  serviceAccountName: sa-name
  
  preset: filebeat-k8s-autodiscover

  config: {}

  secureSettings:
  - secretName: my-secret

  daemonSet:
    podTemplate:
      containers:
      - name: filebeat

  # Alternatively to daemonSet:
  #deployment:
  #  podTemplate:
  #    containers:
  #    - name: metricbeat
```

### Work tracking

[Existing issues](https://github.com/elastic/cloud-on-k8s/issues?q=is%3Aissue+is%3Aopen+label%3A%3Abeats)

david-kow:
As main work items for Beat CRD are closed, I'm closing this meta issue. Bugs and future improvements will be tracked in separate issues.

Issue #1786, opened 2019-09-25
ReconcilerStatus attribute naming


pebrc:
Part of #1141 
```go
// ReconcilerStatus represents status information about desired/available nodes.
type ReconcilerStatus struct {
	AvailableNodes int `json:"availableNodes,omitempty"`
}
```

It was suggested by @nkvoll to rename this attribute to `ReadyPods` instead of `AvailableNodes` because the latter is imprecise as it is not clear what is meant with nodes and what available means. 

The suggestion is also to add an additional attribute representing the desired pods e.g. `pods` so that we can generate a print column similar to what built-in resources have: 

```
NAME                   HEALTH    READY     VERSION   PHASE               AGE
elasticsearch-sample   green     2/3         7.3.0   ApplyingChanges     12h
```

pebrc:
`additionalPrinterColumns` can only be simple JSON path expressions so the suggested fractional syntax is not possible it seems.

thbkrkr:
> `additionalPrinterColumns` can only be simple JSON path expressions so the suggested fractional syntax is not possible it seems.

What if we maintain a `string` field in the `ReconcilerStatus` computed from the number of ready pods and desired pods:

```go
type ReconcilerStatus struct {
	AvailableNodes int32  `json:"availableNodes,omitempty"`
	ReadyPods      int32  `json:"readyPods,omitempty"`
	DesiredPods    int32  `json:"desiredPods,omitempty"`
	PodCounter     string `json:"podCounter,omitempty"` # <---
}```

PR #3041, opened 2020-05-12
Introduce Beat CRD

david-kow:
This PR introduces first iteration of CRD for Beats and it's updated to reflect most recent changes.

#### Filebeat

```console
$ kubectl get beat filebeat-sample
NAME              HEALTH   AVAILABLE   EXPECTED   TYPE       VERSION   AGE
filebeat-sample   green    3           3          filebeat   7.6.2     2m7s

$ kubectl tree beat filebeat-sample
NAMESPACE  NAME                                                             READY  REASON  AGE
default    Beat/filebeat-sample                                             -              19s
default    ├─DaemonSet/filebeat-sample-beat-filebeat                        -              16s
default    │ ├─ControllerRevision/filebeat-sample-beat-filebeat-7f7667d58b  -              16s
default    │ ├─Pod/filebeat-sample-beat-filebeat-pnx85                      True           16s
default    │ ├─Pod/filebeat-sample-beat-filebeat-wmnvb                      True           16s
default    │ └─Pod/filebeat-sample-beat-filebeat-x5rtf                      True           16s
default    ├─Secret/filebeat-sample-beat-es-ca                              -              17s
default    ├─Secret/filebeat-sample-beat-filebeat-config                    -              17s
default    ├─Secret/filebeat-sample-beat-user                               -              17s
default    └─ServiceAccount/elastic-operator-beat-filebeat-sample-d7fb      -              17s
```

#### Heartbeat

```console
$ kubectl get beat heartbeat-sample
NAME               HEALTH   AVAILABLE   EXPECTED   TYPE        VERSION   AGE
heartbeat-sample   green    1           1          heartbeat   7.6.2     10h

$ kubectl tree beat heartbeat-sample
default    Beat/heartbeat-sample                                         -              10h
default    ├─Deployment/heartbeat-sample-beat-otherbeat                  -              10h
default    │ └─ReplicaSet/heartbeat-sample-beat-otherbeat-c8645b56       -              10h
default    │   └─Pod/heartbeat-sample-beat-otherbeat-c8645b56-cr4f8      True           10h
default    ├─Secret/heartbeat-sample-beat-es-ca                          -              10h
default    ├─Secret/heartbeat-sample-beat-otherbeat-config               -              10h
default    ├─Secret/heartbeat-sample-beat-user                           -              10h
default    └─ServiceAccount/elastic-operator-beat-heartbeat-sample-8d7e  -              10h
```

### Features

- native support for Filebeat and Metricbeat
- initial support for other Beats
- support for daemonsets and deployments
- providing Beat status
- basic webhook/non-webhook validation
- rollout for config/version changes
- labelling for all created resources
- support for pod template/config overriding
- basic apm/events/logging
- RBAC for autodiscovery setup (can be disabled via operator flag, separate issue at https://github.com/elastic/cloud-on-k8s/issues/3134)

#### State

This is the first pass, meta-issue was created (https://github.com/elastic/cloud-on-k8s/issues/3040) for additional context and work tracking. There is couple of open issues to discuss PSP/SCC treatment (https://github.com/elastic/cloud-on-k8s/issues/3133), default/merging config (https://github.com/elastic/cloud-on-k8s/issues/3042) and autodiscover RBAC (https://github.com/elastic/cloud-on-k8s/issues/3134).

#### Testing

I was able to run Filebeat, Metricbeat and Heartbeat successfully, change config, upgrade versions. Some E2E are added for couple basic scenarios and SmokeTest runs with Filebeat now. Tested on OpenShift as well. There is a facility introduced that helps with querying ES and checking whether Beats events really appear in ES. In E2E tests autodiscover RBAC is managed by the operator, while the correct PSP is granted to the operator similarly to other stack products. This will be improved to grant additional permissions only to Beat pods and avoid escalating operator privileges unnecessarily.

Also, at some point, tested manually with:
- Filebeat config targeting specific namespaces, labels
- output overriden
- multiple Beats, multiple Beats per namespace

#### Beat support

Only filebeat and metricbeat are supported natively. Other beats can be deployed, but only  when both `image` and `config` are provided.

#### Config handling

As this deserves an issue of it's own to discuss (https://github.com/elastic/cloud-on-k8s/issues/3042), I'll just sum up what is currently in the code: if no `spec.config` is provided we use a static, default config from `controller/common/beat/filebeat/config.go`, otherwise we use the provided config. The only merging we do is the `output` if `elasticsearchRef` was provided.


david-kow:
jenkins test this please

cmd/manager/main.go
```diff
@@ -166,6 +168,11 @@ func init() {
 		"",
 		fmt.Sprintf("K8s secret mounted into the path designated by %s to be used for webhook certificates", operator.WebhookCertDirFlag),
 	)
+	Cmd.Flags().Bool(
+		operator.DisableAutodiscoveryRBACSetup,
```
anyasabo, line 172: Nit: I wonder if we should go the opposite way since the other flags are positive (i.e. `EnableAutodiscovery`), so that we can stay consistent between flags

david-kow: I wanted it to be enabled by default, so we would need to modify args in all manifests. Alternatively set the default to true, but this feels off to me. Are we fine with adding another arg?

pkg/controller/common/beat/autodiscovery.go
```diff
@@ -0,0 +1,145 @@
...
+		Owner:      owner,
+		Expected:   obj,
+		Reconciled: obj2,
+		NeedsUpdate: func() bool {
+			objMeta2, err := meta.Accessor(obj2)
+
+			// compare hash of the deployment at the time it was built
+			return err != nil || hash.GetTemplateHashLabel(objMeta.GetLabels()) != hash.GetTemplateHashLabel(objMeta2.GetLabels())
+		},
+		UpdateReconciled: func() {
```
barkbay, line 142: `UpdateReconciled` is empty, which means it is the previous state of the object (`obj2 := obj.DeepCopyObject()`) that is used in the update call

cmd/manager/main.go
anyasabo: We default `ManageWebhookCertsFlag` to true currently, but I see your point.

>Are we fine with adding another arg?

Do you mean are we fine with adding this arg in particular? I thought we had come to consensus in a previous meeting that we were okay with it, and I still am, but could be persuaded otherwise.

config/operator/global/rbac.yaml
```diff
@@ -39,6 +42,6 @@ rules:
   - apiGroups: ["kibana.k8s.elastic.co"]
     resources: ["kibanas"]
     verbs: ["create", "delete", "deletecollection", "patch", "update"]
-  - apiGroups: ["enterprisesearch.k8s.elastic.co"]
```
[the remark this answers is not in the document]
david-kow: Just a mistake on my part.

pkg/controller/association/controller/beat_es.go
```diff
@@ -0,0 +1,42 @@
...
+		AssociatedShortName:   "beat",
+		AssociationLabels: func(associated types.NamespacedName) map[string]string {
+			return map[string]string{
+				BeatESAssociationLabelName:      associated.Name,
+				BeatESAssociationLabelNamespace: associated.Namespace,
+			}
+		},
+		UserSecretSuffix: "beat-user",
+		ESUserRole: func(commonv1.Associated) (s string, e error) {
+			return "superuser", nil
```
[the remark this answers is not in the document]
david-kow: IIUC, there is no built-in role that would have all necessary permissions (https://www.elastic.co/guide/en/beats/filebeat/master/feature-roles.html). This is tracked in https://github.com/elastic/cloud-on-k8s/pull/3040 (Use less privileged user to connect to ES") as probably requires more discussion.

pkg/controller/common/daemonset/reconcile.go
```diff
@@ -0,0 +1,63 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License;
+// you may not use this file except in compliance with the Elastic License.
+
+package daemonset
```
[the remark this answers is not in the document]
david-kow: It's consistent with deployment package (also `New` and `Reconcile` funcs only).

pkg/controller/common/beat/filebeat/filebeat.go
```diff
@@ -0,0 +1,208 @@
...
+		},
+	}
+)
+
+type Driver struct {
+	commonbeat.DriverParams
+	commonbeat.Driver
+}
+
+func NewDriver(params commonbeat.DriverParams) commonbeat.Driver {
```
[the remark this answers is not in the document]
barkbay: Closing this comment. In the process of the review I forgot your mention re. the package org.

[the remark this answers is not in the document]
david-kow: It's in the PR description :) This is the building block that is prepared to be consumed by other controllers (like future Stack Monitoring controller) that wish to deploy Beats.

pkg/controller/common/beat/otherbeat/otherbeat.go
```diff
@@ -0,0 +1,221 @@
...
+	// reconcile
+	if _, err = reconciler.ReconcileSecret(fd.Client, expected, fd.Owner); err != nil {
+		results.WithError(err)
+	}
+
+	checksum := sha256.New224()
+	_, _ = checksum.Write(cfgBytes)
+
+	// reconcile pod vehicle
+	if driverStatus, err := doReconcile(
```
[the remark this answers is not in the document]
david-kow: I'm passing entire DriverParams now to a common func.

pkg/apis/beat/v1beta1/webhook.go
```diff
@@ -0,0 +1,121 @@
...
+	"fmt"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+	ctrl "sigs.k8s.io/controller-runtime"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
+
```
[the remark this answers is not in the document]
david-kow: That's on purpose - stdlib/external/internal packages are grouped separately.

pkg/controller/common/beat/filebeat/filebeat.go
```diff
@@ -0,0 +1,63 @@
...
+			containersVolume,
+			containersLogsVolume,
+			podsLogsVolume,
+		} {
+			builder.WithVolumes(volume.Volume()).WithVolumeMounts(volume.VolumeMount())
+		}
+	}
+
+	if d.DaemonSet == nil && d.Deployment == nil {
+		d.DaemonSet = &commonbeat.DaemonSetSpec{}
```
pebrc, line 55: Feels slightly out of place here. Should this be done during Driver initialisation?

pkg/controller/common/beat/health/health.go
```diff
@@ -0,0 +1,42 @@
...
+package health
+
+import (
+	v1 "github.com/elastic/cloud-on-k8s/pkg/apis/common/v1"
+)
+
+type BeatHealth string
+
+const (
+	// BeatRedHealth means that the health is neither yellow nor green
```
pebrc, line 14:
```suggestion
	// BeatRedHealth means that the health is neither yellow nor green.
```
same below

pkg/controller/common/beat/health/health.go
```diff
@@ -0,0 +1,42 @@
...
+	// 2) association is not configured or configured and established
+	BeatYellowHealth BeatHealth = "yellow"
+
+	// BeatGreenHealth means that:
+	// 1) all Pods are Ready, and
+	// 2) association is not configured or configured and established
+	BeatGreenHealth BeatHealth = "green"
+)
+
+// CalculateHealth returns health of the Beat calculated based on association status, desired count and ready count.
```
pebrc, line 28:
```suggestion
// CalculateHealth returns health of the Beat based on association status, desired count and ready count.
```

pkg/controller/common/beat/metricbeat/metricbeat.go
```diff
@@ -0,0 +1,66 @@
...
+		} {
+			builder.WithVolumes(volume.Volume()).WithVolumeMounts(volume.VolumeMount())
+		}
+
+		builder.WithArgs("-e", "-c", commonbeat.ConfigMountPath, "-system.hostfs=/hostfs")
+	}
+
+	if d.DaemonSet == nil && d.Deployment == nil {
+		d.DaemonSet = &commonbeat.DaemonSetSpec{}
+	}
```
pebrc, line 59: Same as above: should this be in `NewDriver` instead?

pkg/controller/common/beat/otherbeat/otherbeat.go
```diff
@@ -0,0 +1,33 @@
...
+
+package otherbeat
+
+import (
+	commonbeat "github.com/elastic/cloud-on-k8s/pkg/controller/common/beat"
+	"github.com/elastic/cloud-on-k8s/pkg/utils/pointer"
+)
+
+const (
+	Type commonbeat.Type = "otherbeat"
```
pebrc, line 13: Unused? Also I think we should not surface the name `otherbeat` anywhere to the user (but I think the name is OK for internal use)

pkg/controller/common/beat/pod.go
```diff
@@ -0,0 +1,122 @@
...
+
+	ConfigVolumeName = "config"
+	ConfigMountPath  = "/etc/beat.yml"
+	ConfigFileName   = "beat.yml"
+
+	DataVolumeName        = "data"
+	DataMountPathTemplate = "/var/lib/%s/%s/%s-data"
+	DataPathTemplate      = "/usr/share/%s/data"
+
+	// ConfigChecksumLabel is a label used to store beats config checksum.
```
pebrc, line 33:
```suggestion
	// ConfigChecksumLabel is a label used to store a Beats config checksum.
```

pkg/controller/common/beat/pod.go
```diff
@@ -0,0 +1,122 @@
...
+)
+
+func buildPodTemplate(params DriverParams, defaultImage container.Image, f func(builder *defaults.PodTemplateBuilder), checksum hash.Hash) corev1.PodTemplateSpec {
+	podTemplate := params.GetPodTemplate()
+
+	// Token mounting gets defaulted to false, which prevents from detecting whether user set it.
+	// Instead, checking that here, before the default is applied.
+	if podTemplate.Spec.AutomountServiceAccountToken == nil {
+		t := true
+		podTemplate.Spec.AutomountServiceAccountToken = &t
```
pebrc, line 47: You are changing the default to `true` for Beats here IIUC, is it always required? Maybe add a comment with more explanation for folks like me.

pkg/controller/common/beat/pod.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+
+	builder := defaults.NewPodTemplateBuilder(podTemplate, params.Type)
+
+	if f != nil {
```
pebrc, line 52: Maybe add a comment and explain when this modifier func is nil (in case of the generic beat iiuc)

pkg/controller/common/beat/pod.go
```diff
@@ -0,0 +1,122 @@
...
+	DataPathTemplate      = "/usr/share/%s/data"
+
+	// ConfigChecksumLabel is a label used to store beats config checksum.
+	ConfigChecksumLabel = "beat.k8s.elastic.co/config-checksum"
+
+	// VersionLabelName is a label used to track the version of a Beat Pod.
+	VersionLabelName = "beat.k8s.elastic.co/version"
+)
+
+func buildPodTemplate(params DriverParams, defaultImage container.Image, f func(builder *defaults.PodTemplateBuilder), checksum hash.Hash) corev1.PodTemplateSpec {
```
pebrc, line 40: I would probably have spent two more letters and called `f` `mod` or something like that. Even if it's an internal function. Just a personal preference, feel free to ignore it.

pkg/controller/common/beat/reconcile.go
```diff
@@ -0,0 +1,121 @@
...
+	case params.Deployment != nil:
+		{
+			reconciliationFunc = reconcileDeployment
+			toDelete = &v1.DaemonSet{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      name,
+					Namespace: params.Owner.GetNamespace(),
+				},
+			}
+		}
```
pebrc, line 48: I _think_ the extra parens in the `case`s are redundant?

pkg/controller/common/daemonset/reconcile.go
```diff
@@ -0,0 +1,64 @@
...
+	expected = WithTemplateHash(expected)
+
+	reconciled := &appsv1.DaemonSet{}
+	err := reconciler.ReconcileResource(reconciler.Params{
+		Client:     k8sClient,
+		Owner:      owner,
+		Expected:   &expected,
+		Reconciled: reconciled,
+		NeedsUpdate: func() bool {
+			// compare hash of the daemon set at the time it was built
```
pebrc, line 49:
```suggestion
			// compare hash of the DaemonSet at the time it was built
```
also not sure I fully understand the comment

pkg/controller/common/operator/flags.go
```diff
@@ -12,6 +12,7 @@ const (
 	CertValidityFlag            = "cert-validity"
 	ContainerRegistryFlag       = "container-registry"
 	DebugHTTPListenFlag         = "debug-http-listen"
+	EnableAutodiscoverRBACSetup = "disable-autodiscover-rbac-setup"
```
pebrc, line 15: Var and value don't line up: enable or disable

pkg/controller/common/settings/canonical_config.go
```diff
@@ -89,6 +89,16 @@ func ParseConfig(yml []byte) (*CanonicalConfig, error) {
 	return fromConfig(config), nil
 }
 
+// MustParseConfig parses the given configuration content into a CanonicalConfig.
+// Expects content to be in YAML format. Panis on error.
```
pebrc, line 93:
```suggestion
// Expects content to be in YAML format. Panics on error.
```

pkg/controller/common/volume/host.go
```diff
@@ -0,0 +1,63 @@
...
+// or more contributor license agreements. Licensed under the Elastic License;
+// you may not use this file except in compliance with the Elastic License.
+
+package volume
+
+import (
+	corev1 "k8s.io/api/core/v1"
+)
+
+// NewReadOnlyHostVolume creates a new HostVolume struct
```
pebrc, line 11:
```suggestion
// NewReadOnlyHostVolume creates a new HostVolume struct.
```
and below

config/samples/beat/filebeat_es_kibana_apm.yaml
```diff
@@ -0,0 +1,74 @@
...
+  name: kibana-sample
+spec:
+  version: 7.7.0
+  count: 1
+  elasticsearchRef:
+    name: elasticsearch-sample
+  #http:
+  #  service:
+  #    spec:
+  #      type: LoadBalancer
```
pebrc, line 28: Do we need this here?

pkg/controller/beat/controller.go
```diff
@@ -0,0 +1,282 @@
...
+		return otherbeat.NewDriver(dp)
+	}
+}
+
+func newDriverParams(ctx context.Context, client k8s.Client, beat beatv1beta1.Beat) commonbeat.DriverParams {
+	spec := beat.Spec
+
+	var ds *commonbeat.DaemonSetSpec
+	if spec.DaemonSet != nil {
+		ds = &commonbeat.DaemonSetSpec{PodTemplate: spec.DaemonSet.PodTemplate}
```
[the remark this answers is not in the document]
david-kow: I didn't like that either, but if we use the one from `controller/beat` then the common code uses Beat CRD specific structs (ie. another controller will need to import Beat controller to do use it). If we use the one from `controller/common/beat` there is an import cycle:
```
import cycle not allowed
package main
	imports github.com/elastic/cloud-on-k8s/cmd/manager
	imports github.com/elastic/cloud-on-k8s/pkg/controller/apmserver
	imports github.com/elastic/cloud-on-k8s/pkg/controller/association
	imports github.com/elastic/cloud-on-k8s/pkg/controller/common
	imports github.com/elastic/cloud-on-k8s/pkg/controller/common/operator
	imports github.com/elastic/cloud-on-k8s/pkg/controller/common/certificates
	imports github.com/elastic/cloud-on-k8s/pkg/controller/common/reconciler
	imports github.com/elastic/cloud-on-k8s/pkg/utils/k8s
	imports github.com/elastic/cloud-on-k8s/pkg/controller/common/scheme
	imports github.com/elastic/cloud-on-k8s/pkg/apis/beat/v1beta1
	imports github.com/elastic/cloud-on-k8s/pkg/controller/common/beat
	imports github.com/elastic/cloud-on-k8s/pkg/controller/association
```

pkg/controller/beat/name.go
```diff
@@ -0,0 +1,26 @@
...
+import (
+	commonbeat "github.com/elastic/cloud-on-k8s/pkg/controller/common/beat"
+	common_name "github.com/elastic/cloud-on-k8s/pkg/controller/common/name"
+)
+
+// namer is a Namer that is configured with the defaults for resources related to a Beat resource.
+var namer = common_name.NewNamer("beat")
+
+type Namer struct {
+}
```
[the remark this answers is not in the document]
david-kow: This is to allow it to implement `pkg/controller/common/beat/name.go`. And that interface is there so that if a different controller wants to deploy Beats, they can use a different naming scheme. E.g. suffix names with `stackmonitoring` instead of `beat`.

pkg/controller/beat/name_test.go
```diff
@@ -0,0 +1,57 @@
...
+
+func TestNamer_ConfigSecretName(t *testing.T) {
+	namer := Namer{}
+	for _, typeName := range []struct {
+		name string
+		typ  string
+		want string
+	}{
+		{
+			name: "my",
```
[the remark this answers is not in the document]
david-kow: Hurry :) Fixed.

pkg/controller/common/beat/otherbeat/otherbeat.go
david-kow: Yes, unused. I'll remove for now, we can reintroduce it when there is a need.

pkg/controller/common/beat/pod.go
david-kow: Changing the default, yes. I've expanded the comment.

As to whether it's always required - no. We need it by default as we enable autodiscover in default configs. To rule out whether it's needed we might need to check for `*beat.autodiscover` path presence in the config, but I'm not sure if this is enough and whether it's worth it to add a condition here. We might be missing a case and disable it when it's actually needed.

david-kow:
Thanks for the review @pebrc!

> One thing I am bit on the fence about is the package structure. (...) So for example have we ruled out that stack monitoring will not simply create specially configured Beats resources for the controller you are introducing here (as opposed to code reuse through another controller)?

I don't think we discussed this. My thinking was that we want to hide the reuse in the code rather than expose it via resources that are visible to the user. It seems to me it could cause confusion and I don't see any advantages in doing so. So unless something is really blocking code reuse I'd be for doing it. @anyasabo, any thoughts on that?

pkg/controller/common/beat/pod.go
```diff
@@ -0,0 +1,124 @@
...
+		WithResources(defaultResources).
+		WithHostNetwork().
+		WithLabels(map[string]string{
+			ConfigChecksumLabel: fmt.Sprintf("%x", checksum.Sum(nil)),
+			VersionLabelName:    params.Version}).
+		WithDockerImage(params.Image, container.ImageRepository(defaultImage, params.Version)).
+		WithArgs("-e", "-c", ConfigMountPath).
+		WithDNSPolicy(corev1.DNSClusterFirstWithHostNet).
+		WithSecurityContext(corev1.SecurityContext{
+			RunAsUser: pointer.Int64(0),
```
sebgl, line 76: This is because the Beats image runs with root user by default? May be worth calling out explicitly. It also means it won't run on Openshift?

anyasabo:
>     I don't think we discussed this. My thinking was that we want to hide the reuse in the code rather than expose it via resources that are visible to the user. It seems to me it could cause confusion and I don't see any advantages in doing so. So unless something is really blocking code reuse I'd be for doing it.

I don't feel strongly either way at this time to be honest. The more I think about it though the more I like the idea of the StackMonitoring controller just creating a beat CR. It simplifies the stack monitoring controller even more than code reuse would, since it just reconciles one resource (and is clearer that it is just some syntactic sugar).

config/samples/beat/filebeat_es_kibana_apm.yaml
```diff
@@ -0,0 +1,68 @@
...
+  #  - node: ${NODE_NAME}
+  #    type: kubernetes
+  #    templates:
+  #    - condition.equals.kubernetes.namespace: log-this
+  #      config:
+  #      - paths: ["/var/log/containers/*${data.kubernetes.container.id}.log"]
+  #        type: container
+  #  processors:
+  #    - add_cloud_metadata: null
+  #    - add_host_metadata: null
```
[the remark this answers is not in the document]
david-kow: Can't repro :( With:

```yaml
kind: Beat
metadata:
  name: filebeat-sample
spec:
  type: filebeat
  version: 7.7.0
  elasticsearchRef:
    name: elasticsearch-sample
  config:
    filebeat.autodiscover.providers:
    - node: ${NODE_NAME}
      type: kubernetes
      templates:
      - config:
        - paths: ["/var/log/containers/*${data.kubernetes.container.id}.log"]
          type: container
    processors:
    - add_cloud_metadata: null
    - add_host_metadata: null
```

I'm getting:
```yaml
$ k get secret filebeat-sample-beat-filebeat-config -o json | jq -r '.data["beat.yml"]' | base64 -D
filebeat:
  autodiscover:
    providers:
    - node: ${NODE_NAME}
      templates:
      - config:
        - paths:
          - /var/log/containers/*${data.kubernetes.container.id}.log
          type: container
      type: kubernetes
output:
  elasticsearch:
    hosts:
    - https://elasticsearch-sample-es-http.default.svc:9200
    password: J7247ZyS2ws24KNNh7zdB62x
    ssl:
      certificate_authorities: /mnt/elastic-internal/es-certs/ca.crt
    username: default-filebeat-sample-beat-user
processors:
- add_cloud_metadata: null
- add_host_metadata: null
``` 

What's cluster k8s version? The above is on `v1.15.9`.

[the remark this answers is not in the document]
david-kow: Can't repro :( With:

```yaml
kind: Beat
metadata:
  name: filebeat-sample
spec:
  type: filebeat
  version: 7.7.0
  elasticsearchRef:
    name: elasticsearch-sample
  config:
    filebeat.autodiscover.providers:
    - node: ${NODE_NAME}
      type: kubernetes
      templates:
      - config:
        - paths: ["/var/log/containers/*${data.kubernetes.container.id}.log"]
          type: container
    processors:
    - add_cloud_metadata: null
    - add_host_metadata: null
```

I'm getting:
```yaml
$ k get secret filebeat-sample-beat-filebeat-config -o json | jq -r '.data["beat.yml"]' | base64 -D
filebeat:
  autodiscover:
    providers:
    - node: ${NODE_NAME}
      templates:
      - config:
        - paths:
          - /var/log/containers/*${data.kubernetes.container.id}.log
          type: container
      type: kubernetes
output:
  elasticsearch:
    hosts:
    - https://elasticsearch-sample-es-http.default.svc:9200
    password: J7247ZyS2ws24KNNh7zdB62x
    ssl:
      certificate_authorities: /mnt/elastic-internal/es-certs/ca.crt
    username: default-filebeat-sample-beat-user
processors:
- add_cloud_metadata: null
- add_host_metadata: null
``` 

What's cluster k8s version? The above is on `v1.15.9`.

config/e2e/operator.yaml
```diff
@@ -273,3 +308,26 @@ spec:
...
+  - watch
+  - list
+- apiGroups:
+  - policy
+  resources:
+  - podsecuritypolicies
+  resourceNames:
+  - elastic.beat.restricted
+  verbs:
+  - use
```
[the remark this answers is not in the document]
david-kow: That change was "in flight" - initially operator created this itself, now it will be installed together with ECK. Added support for `make run`.

cmd/manager/main.go
```diff
@@ -166,6 +168,11 @@ func init() {
 		"",
 		fmt.Sprintf("K8s secret mounted into the path designated by %s to be used for webhook certificates", operator.WebhookCertDirFlag),
 	)
+	Cmd.Flags().Bool(
+		operator.EnableAutodiscoverRBACSetupFlag,
+		true,
+		"Determines whether the operator should set up role, binding and service account for the Beats autodiscover feature",
```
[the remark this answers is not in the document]
david-kow: It's not entire rbac so I'll do `manage-beat-autodiscover-rbac`, let me know if it doesn't work.

config/samples/beat/filebeat_es_kibana_apm.yaml
```diff
@@ -0,0 +1,68 @@
+# This sample sets up an Elasticsearch cluster and a Kibana instance preconfigured for that cluster
```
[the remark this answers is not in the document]
david-kow: It's used as a smoke test config so I wanted to include all, similarly as previous one did es+kibana+apm.

config/crds/all-crds.yaml
```diff
@@ -438,6 +438,153 @@ status:
...
+        spec:
+          description: BeatSpec defines the desired state of a Beat.
+          properties:
+            config:
+              description: Config holds the Beat configuration. If provided, it will
+                override the default configuration.
+              type: object
+            daemonSet:
+              description: 'DaemonSet allows to: 1. Indicate whether the Beat should
+                be deployed as DaemonSet or Deployment, if both are absent, a default
```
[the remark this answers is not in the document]
david-kow: This doesn't convey this part from the current version: `if both are absent, a default for the Type is used`. I've changed `Cannot be used along with ``deployment`` ` part, I'm open to further suggestions :)

pkg/controller/beat/controller.go
```diff
@@ -0,0 +1,278 @@
...
+		Namer:      &Namer{},
+		Owner:      &beat,
+		Associated: &beat,
+
+		Type:               spec.Type,
+		Version:            spec.Version,
+		ElasticsearchRef:   spec.ElasticsearchRef,
+		Image:              spec.Image,
+		Config:             spec.Config,
+		ServiceAccountName: spec.ServiceAccountName,
```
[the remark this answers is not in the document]
david-kow: So that `controller/common` doesn't depend on `controller/beat`.

pkg/controller/beat/controller.go
```diff
@@ -0,0 +1,278 @@
...
+	spec := beat.Spec
+
+	var ds *commonbeat.DaemonSetSpec
+	if spec.DaemonSet != nil {
+		ds = &commonbeat.DaemonSetSpec{PodTemplate: spec.DaemonSet.PodTemplate}
+	}
+	var d *commonbeat.DeploymentSpec
+	if spec.Deployment != nil {
+		d = &commonbeat.DeploymentSpec{Replicas: spec.Deployment.Replicas, PodTemplate: spec.Deployment.PodTemplate}
+	}
```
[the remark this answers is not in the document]
david-kow: This is to properly initialize `commonbeat.DriverParams`. We can't just assign here as `DeploymentSpec` in `beat_types.go` is not the same as `commonbeat.DeploymentSpec`. We can't reuse types as it leads to import cycles.

pkg/controller/beat/controller.go
```diff
@@ -0,0 +1,278 @@
...
+
+	return commonbeat.DriverParams{
+		Client:  client,
+		Context: ctx,
+		Logger:  log,
+
+		Namer:      &Namer{},
```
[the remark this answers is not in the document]
david-kow: This is (again:)) about making `controller/common/beat` reusable. A different controller may wish to name resources differently, which having a struct implementing an interface allows.

pkg/controller/common/beat/autodiscover.go
```diff
@@ -0,0 +1,149 @@
...
+				"namespace", owner.GetNamespace(),
+				"beat_name", owner.GetName())
+			return err
+		}
+	}
+	return nil
+}
+
+func setupAutodiscoverRBAC(ctx context.Context, client k8s.Client, owner metav1.Object, labels map[string]string) error {
+	// this is the source of truth and should be respected at all times
```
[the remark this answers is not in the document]
david-kow: It means that even if (by accident) we call this when autodiscover management is disabled, we won't create those resources. Any suggestions to reword? To me it sounds better than the explanation I just gave :)

pkg/controller/common/beat/autodiscover.go
```diff
@@ -0,0 +1,149 @@
...
+			Name:     clusterRoleName,
+		},
+	}
+
+	return doReconcile(client, binding, nil)
+}
+
+func doReconcile(client k8s.Client, obj runtime.Object, owner metav1.Object) error {
+	// labels set here must be exactly the same for all callers in a particular namespace
+	// otherwise they'll just keep trying to override each other
```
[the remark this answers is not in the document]
david-kow: Since one of the resources (clusterrole) reconciled is shared between many Beats, it can't have different labels depending on which Beat resource is currently being reconciled. This is no longer needed after recent changes.

pkg/controller/common/beat/autodiscover.go
```diff
@@ -0,0 +1,149 @@
...
+		Reconciled: reconciled,
+		NeedsUpdate: func() bool {
+			objMeta2, err := meta.Accessor(reconciled)
+
+			// compare hash of the deployment at the time it was built
+			return err != nil || hash.GetTemplateHashLabel(objMeta.GetLabels()) != hash.GetTemplateHashLabel(objMeta2.GetLabels())
+		},
+		UpdateReconciled: func() {
+			reflect.ValueOf(reconciled).Elem().Set(reflect.ValueOf(obj).Elem())
+		},
```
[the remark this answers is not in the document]
david-kow: Now we only reconcile two (binding and service account). I'll make separate funcs.

pkg/controller/common/beat/common.go
```diff
@@ -0,0 +1,140 @@
...
+	if (dp.DaemonSet == nil && dp.Deployment == nil) || (dp.DaemonSet != nil && dp.Deployment != nil) {
+		return fmt.Errorf("either daemonset or deployment has to be specified")
+	}
+	return nil
+}
+
+type DriverResults struct {
+	*reconciler.Results
+	Status *DriverStatus
+}
```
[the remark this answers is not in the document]
david-kow: Yes. It can go.

pkg/controller/common/beat/common.go
```diff
@@ -0,0 +1,140 @@
...
+		Results: reconciler.NewResult(ctx),
+	}
+}
+
+type DriverStatus struct {
+	ExpectedNodes  int32
+	AvailableNodes int32
+	Health         health.BeatHealth
+	Association    commonv1.AssociationStatus
+}
```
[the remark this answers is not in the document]
david-kow: Import cycles :(

pkg/controller/common/beat/common.go
```diff
@@ -0,0 +1,140 @@
...
+		return results
+	}
+
+	podTemplate := buildPodTemplate(params, defaultImage, modifyPodFunc, checksum)
+	if driverStatus, err := reconcilePodVehicle(podTemplate, params); err != nil {
+		if apierrors.IsConflict(err) {
+			params.Logger.V(1).Info("Conflict while updating")
+			results.WithResult(reconcile.Result{Requeue: true})
+		}
+		results.WithError(err)
```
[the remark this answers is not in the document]
david-kow: Yes, good catch.

pkg/controller/common/beat/config.go
```diff
@@ -0,0 +1,141 @@
...
+var (
+	defaultResources = corev1.ResourceRequirements{
+		Limits: map[corev1.ResourceName]resource.Quantity{
+			corev1.ResourceMemory: resource.MustParse("200Mi"),
+			corev1.ResourceCPU:    resource.MustParse("100m"),
+		},
+		Requests: map[corev1.ResourceName]resource.Quantity{
+			corev1.ResourceMemory: resource.MustParse("200Mi"),
+			corev1.ResourceCPU:    resource.MustParse("100m"),
+		},
```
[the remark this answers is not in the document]
david-kow: From Filebeat/Metricbeat examples (https://raw.githubusercontent.com/elastic/beats/7.7/deploy/kubernetes/filebeat-kubernetes.yaml) although I've increased requests to match limits. Open to suggestions here.

pkg/controller/common/beat/pod.go
```diff
@@ -0,0 +1,124 @@
...
+		WithTerminationGracePeriod(30).
+		WithEnv(corev1.EnvVar{
+			Name: "NODE_NAME",
+			ValueFrom: &corev1.EnvVarSource{
+				FieldRef: &corev1.ObjectFieldSelector{
+					FieldPath: "spec.nodeName",
+				},
+			}}).
+		WithResources(defaultResources).
+		WithHostNetwork().
```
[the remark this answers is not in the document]
david-kow: I didn't test without it, but:

```yaml
$ curl https://raw.githubusercontent.com/elastic/beats/7.7/deploy/kubernetes/filebeat-kubernetes.yaml | grep hostNetwork
      hostNetwork: true
```

For now I assume these as a baseline.

pkg/controller/common/beat/config.go
```diff
@@ -0,0 +1,141 @@
...
+	client k8s.Client,
+	associated commonv1.Associated,
+	defaultConfig *settings.CanonicalConfig,
+	userConfig *commonv1.Config,
+) ([]byte, error) {
+	cfg := settings.NewCanonicalConfig()
+
+	if defaultConfig == nil && userConfig == nil {
+		return nil, fmt.Errorf("both default and user configs are nil")
+	}
```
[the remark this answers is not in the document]
david-kow: Ah, that's true. Nice catch.

pkg/controller/common/beat/common.go
```diff
@@ -0,0 +1,140 @@
...
+	if err := params.Validate(); err != nil {
+		results.WithError(err)
+		return results
+	}
+
+	if err := SetupAutodiscoverRBAC(params.Context, params.Logger, params.Client, params.Owner, params.Labels); err != nil {
+		results.WithError(err)
+	}
+
+	checksum := sha256.New224()
```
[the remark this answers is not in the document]
david-kow: Changed to `configHash`, similar to entsearch.

pkg/controller/common/beat/pod.go
david-kow: I took this from Beats examples, but after looking more carefully I think we can do a little better.

Beat containers run as uid 1000 by default, so for most of the mounted paths it's enough if we set security context with correct fsgroup.

The issue is with hostpath volume that we use for data folder. The directory gets created with 0755 (as documented [here](https://github.com/kubernetes/kubernetes/blob/b98d9407cfdc24c477ab8066d1fb69ea0983a2f9/pkg/apis/core/types.go#L534)) and it ignores fsgroup, so uid 1000 can't write to it. If we can run init container as root we can chown it to the correct user (just like we do with elasticsearch), but if we can't then I believe the directory has to be prepared outside of k8s. [PR](https://github.com/kubernetes/kubernetes/pull/39438) to fix this was already declined.

config/samples/beat/filebeat_es_kibana_apm.yaml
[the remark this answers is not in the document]
sebgl: I'm running `v1.14.10`.

I spent some time trying to reproduce and here is what I can observe:

* when deploying the manifest from scratch with the custom config from the samples, I correctly get:

```yaml
processors:
- add_cloud_metadata: null
- add_host_metadata: null
```

* when deploying the manifest with no config, I also correctly get:

```yaml
processors:
- add_cloud_metadata: null
- add_host_metadata: null
```

* if I update the existing manifest with no config to add my custom config (the one from samples), I get:

```yaml
processors:
- null
- null
```

config/operator/all-in-one/roles.yaml
```diff
@@ -0,0 +1,14 @@
+# permissions needed for Beat cluster-wide autodiscover feature
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: elastic-operator-beat-autodiscover
+rules:
+- apiGroups: [""] # "" indicates the core API group
```
barkbay, line 7: This line is deleted by https://github.com/elastic/cloud-on-k8s/blob/57269eca4c64c24599998d0406e720bd305b7a5b/config/operator/Makefile#L25-L26

You should either remove the comment or update the `Makefile`

pkg/controller/beat/controller.go
```diff
@@ -0,0 +1,278 @@
...
+
+	if commonbeat.ShouldSetupAutodiscoverRBAC() {
+		if err := c.Watch(&source.Kind{Type: &corev1.ServiceAccount{}}, &handler.EnqueueRequestForOwner{
+			IsController: true,
+			OwnerType:    &beatv1beta1.Beat{},
+		}); err != nil {
+			return err
+		}
+
+		if err := c.Watch(&source.Kind{Type: &rbacv1.ClusterRoleBinding{}}, &handler.EnqueueRequestForObject{}); err != nil {
```
barkbay, line 109: This will trigger a reconcile loop for each `ClusterRoleBinding`, I understand that the idea is to ensure that `ServiceAccounts` and `ClusterRoleBinding` are reconciled but it is a little bit "surprising" to see all those messages about some `beat_name` that don't exist during the startup:

```
2020-06-02T15:01:49.397+0200    INFO    beat-controller Starting reconciliation run     {"service.version": "1.1.0-SNAPSHOT-mmo", "iteration": 1, "namespace": "", "beat_name": "system:openshift:controller:unidling-controller"}
2020-06-02T15:01:49.397+0200    INFO    beat-controller Starting reconciliation run     {"service.version": "1.1.0-SNAPSHOT-mmo", "iteration": 2, "namespace": "", "beat_name": "csi-snapshot-controller-operator-role"}
2020-06-02T15:03:58.111+0200    INFO    beat-controller Ending reconciliation run       {"service.version": "1.1.0-SNAPSHOT-mmo", "iteration": 2, "namespace": "", "beat_name": "csi-snapshot-controller-operator-role", "took": "2m8.710463019s"}
2020-06-02T15:03:58.111+0200    INFO    beat-controller Ending reconciliation run       {"service.version": "1.1.0-SNAPSHOT-mmo", "iteration": 1, "namespace": "", "beat_name": "system:openshift:controller:unidling-controller", "took": "2m8.710491152s"}
2020-06-02T15:03:58.111+0200    INFO    beat-controller Starting reconciliation run     {"service.version": "1.1.0-SNAPSHOT-mmo", "iteration": 3, "namespace": "", "beat_name": "system:openshift:controller:build-config-change-controller"}
2020-06-02T15:03:58.112+0200    INFO    beat-controller Starting reconciliation run     {"service.version": "1.1.0-SNAPSHOT-mmo", "iteration": 4, "namespace": "", "beat_name": "cluster-autoscaler"}
2020-06-02T15:03:59.292+0200    INFO    beat-controller Ending reconciliation run       {"service.version": "1.1.0-SNAPSHOT-mmo", "iteration": 3, "namespace": "", "beat_name": "system:openshift:controller:build-config-change-controller", "took": "1.181247897s"}
2020-06-02T15:03:59.292+0200    INFO    beat-controller Starting reconciliation run     {"service.version": "1.1.0-SNAPSHOT-mmo", "iteration": 5, "namespace": "", "beat_name": "system:openshift:operator:kube-apiserver-operator"}
2020-06-02T15:03:59.836+0200    INFO    beat-controller Ending reconciliation run       {"service.version": "1.1.0-SNAPSHOT-mmo", "iteration": 4, "namespace": "", "beat_name": "cluster-autoscaler", "took": "1.72457487s"}
2020-06-02T15:04:00.380+0200    INFO    beat-controller Starting reconciliation run     {"service.version": "1.1.0-SNAPSHOT-mmo", "iteration": 6, "namespace": "", "beat_name": "system:controller:disruption-controller"}
2020-06-02T15:04:00.380+0200    INFO    beat-controller Starting reconciliation run     {"service.version": "1.1.0-SNAPSHOT-mmo", "iteration": 7, "namespace": "", "beat_name": "kube-apiserver"}
2020-06-02T15:04:00.380+0200    INFO    beat-controller Ending reconciliation run       {"service.version": "1.1.0-SNAPSHOT-mmo", "iteration": 5, "namespace": "", "beat_name": "system:openshift:operator:kube-apiserver-operator", "took": "1.087855122s"}
2020-06-02T15:04:00.827+0200    INFO    beat-controller Ending reconciliation run       {"service.version": "1.1.0-SNAPSHOT-mmo", "iteration": 7, "namespace": "", "beat_name": "kube-apiserver", "took": "447.076399ms"}
2020-06-02T15:04:00.827+0200    INFO    beat-controller Starting reconciliation run     {"service.version": "1.1.0-SNAPSHOT-mmo", "iteration": 8, "namespace": "", "beat_name": "alertmanager-main"}
2020-06-02T15:04:01.301+0200    INFO    beat-controller Ending reconciliation run       {"service.version": "1.1.0-SNAPSHOT-mmo", "iteration": 8, "namespace": "", "beat_name": "alertmanager-main", "took": "473.845307ms"}
2020-06-02T15:04:01.301+0200    INFO    beat-controller Starting reconciliation run     {"service.version": "1.1.0-SNAPSHOT-mmo", "iteration": 9, "namespace": "", "beat_name": "default-account-cluster-network-operator"}
2020-06-02T15:04:00.827+0200    INFO    beat-controller Ending reconciliation run       {"service.version": "1.1.0-SNAPSHOT-mmo", "iteration": 6, "namespace": "", "beat_name": "system:controller:disruption-controller", "took": "447.267043ms"}
2020-06-02T15:04:01.739+0200    INFO    beat-controller Starting reconciliation run     {"service.version": "1.1.0-SNAPSHOT-mmo", "iteration": 10, "namespace": "", "beat_name": "system:controller:generic-garbage-collector"}
2020-06-02T15:04:09.500+0200    INFO    beat-controller Ending reconciliation run       {"service.version": "1.1.0-SNAPSHOT-mmo", "iteration": 9, "namespace": "", "beat_name": "default-account-cluster-network-operator", "took": "8.198567717s"}
2020-06-02T15:04:09.500+0200    INFO    beat-controller Starting reconciliation run     {"service.version": "1.1.0-SNAPSHOT-mmo", "iteration": 11, "namespace": "", "beat_name": "system:scope-impersonation"}
```

I'm wondering if it's worth it to introduce a dedicated controller to reconcile these resources.

sebgl: We could also filter on labels here so we only reconcile on ECK-managed `ClusterRoleBinding`?

config/samples/beat/filebeat_es_kibana_apm.yaml
```diff
@@ -0,0 +1,67 @@
...
+    name: elasticsearch-sample
+  # to target only a specific namespace for logging, use:
+  config:
+    filebeat.autodiscover.providers:
+    - node: ${NODE_NAME}
+      type: kubernetes
+      templates:
+      - config:
+        - paths: ["/var/log/containers/*${data.kubernetes.container.id}.log"]
+          type: container
```
sebgl, line 64: I don't get how this config targets a specific namespace?

pkg/controller/common/beat/autodiscover.go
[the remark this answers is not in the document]
sebgl: OK. I think I would remove the comment, the code explains itself better imo 😄

pkg/controller/common/beat/autodiscover.go
```diff
@@ -0,0 +1,141 @@
...
+	expected.Labels = hash.SetTemplateHashLabel(nil, expected)
+
+	reconciled := &corev1.ServiceAccount{}
+	return reconciler.ReconcileResource(reconciler.Params{
+		Client:     client,
+		Owner:      owner,
+		Expected:   expected,
+		Reconciled: reconciled,
+		NeedsUpdate: func() bool {
+			// compare hash of the deployment at the time it was built
```
sebgl, line 92: deployment -> service account

config/samples/beat/filebeat_es_kibana_apm.yaml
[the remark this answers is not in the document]
david-kow: I was able to repro and investigate. Follow up in a separate issue (https://github.com/elastic/cloud-on-k8s/issues/3177). For this PR I will fix the samples.

config/samples/beat/filebeat_es_kibana_apm.yaml
david-kow: The config shouldn't be here at all, removing. I'll create samples for different setup separately.

pkg/controller/beat/controller.go
david-kow: I did the filtering that @sebgl suggested.

pkg/controller/common/beat/metricbeat/metricbeat.go
```diff
@@ -0,0 +1,68 @@
...
+
+	return &Driver{DriverParams: params}
+}
+
+func (d *Driver) Reconcile() (*commonbeat.DriverStatus, *reconciler.Results) {
+	f := func(builder *defaults.PodTemplateBuilder) {
+		dockerSockVolume := volume.NewHostVolume(DockerSockVolumeName, DockerSockPath, DockerSockMountPath, false, corev1.HostPathUnset)
+		procVolume := volume.NewReadOnlyHostVolume(ProcVolumeName, ProcPath, ProcMountPath)
+		cgroupVolume := volume.NewReadOnlyHostVolume(CGroupVolumeName, CGroupPath, CGroupMountPath)
```
sebgl, line 50: I think it's not possible currently to override the `podTemplate` so those volumes are not included. For example if you don't want to mount `/proc` because you don't have permissions to do so.
Let's not delay this PR for that reason though, I'll open another issue.

pkg/controller/common/beat/pod.go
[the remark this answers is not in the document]
david-kow: @sebgl I dug a bit more and found the explanation for needing this: https://github.com/elastic/beats/issues/15013#issuecomment-596428967.

pkg/controller/beat/common/autodiscover.go
```diff
@@ -0,0 +1,205 @@
...
+	}
+
+	reconciled := &rbacv1.ClusterRoleBinding{}
+	return reconciler.ReconcileResource(reconciler.Params{
+		Client:     client,
+		Expected:   expected,
+		Reconciled: reconciled,
+		NeedsUpdate: func() bool {
+			// compare hash of the cluster role binding at the time it was built
+			return hash.GetTemplateHashLabel(expected.Labels) != hash.GetTemplateHashLabel(reconciled.Labels)
```
sebgl, line 184: where do we set this hash?

pkg/controller/beat/common/common.go
```diff
@@ -0,0 +1,86 @@
...
+	case spec.DaemonSet != nil:
+		return spec.DaemonSet.PodTemplate
+	case spec.Deployment != nil:
+		return spec.Deployment.PodTemplate
+	}
+
+	return corev1.PodTemplateSpec{}
+}
+
+func (dp *DriverParams) Validate() error {
```
sebgl, line 50: it feels weird to me to have this function attached to `DriverParams`. I think I would make it a standalone function: `func ValidateBeatSpec(b Beat) error`

pkg/controller/beat/common/common.go
```diff
@@ -0,0 +1,86 @@
...
+		return fmt.Errorf("either daemonset or deployment has to be specified")
+	}
+	return nil
+}
+
+func Reconcile(
+	params DriverParams,
+	defaultConfig *settings.CanonicalConfig,
+	defaultImage container.Image,
+	modifyPodFunc func(builder *defaults.PodTemplateBuilder)) *reconciler.Results {
```
sebgl, line 62: I think we usually indent this way:

```go
func Reconcile(
	params DriverParams,
	defaultConfig *settings.CanonicalConfig,
	defaultImage container.Image,
	modifyPodFunc func(builder *defaults.PodTemplateBuilder),
) *reconciler.Results {
```

but def not a big deal.

pkg/controller/beat/common/common.go
```diff
@@ -0,0 +1,86 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
```
sebgl, line 1: This could be named `driver.go`

pkg/controller/beat/common/pod.go
```diff
@@ -0,0 +1,143 @@
...
+			params.Beat.AssociationConf().GetCASecretName(),
+			CAVolumeName,
+			CAMountPath))
+	}
+
+	for _, v := range volumes {
+		builder = builder.WithVolumes(v.Volume()).WithVolumeMounts(v.VolumeMount())
+	}
+
+	builder = builder.WithLabels(commonhash.SetTemplateHashLabel(NewLabels(params.Beat), builder.PodTemplate))
```
sebgl, line 128: Why do we need the hash label at the pod level? If the pod spec changes the Deployment/DaemonSet controller handles, if the config changes we have the configchecksum label set.
I understand why we'd have it at the DaemonSet or Deployment level though, so we can easily compare and reconcile.

pkg/controller/beat/common/reconcile.go
```diff
@@ -0,0 +1,138 @@
...
+		client:      params.Client,
+		beat:        params.Beat,
+		podTemplate: podTemplate,
+	})
+	if err != nil {
+		return results.WithError(err)
+	}
+
+	// clean up the other one
+	if err := params.Client.Delete(toDelete); err != nil && !apierrors.IsNotFound(err) {
```
sebgl, line 65: We could get (from the cache) then delete if necessary to avoid an unnecessary call to the apiserver.

pkg/controller/beat/common/reconcile.go
```diff
@@ -0,0 +1,138 @@
...
+
+	beatv1beta1 "github.com/elastic/cloud-on-k8s/pkg/apis/beat/v1beta1"
+	"github.com/elastic/cloud-on-k8s/pkg/controller/common/daemonset"
+	"github.com/elastic/cloud-on-k8s/pkg/controller/common/deployment"
+	"github.com/elastic/cloud-on-k8s/pkg/controller/common/reconciler"
+	"github.com/elastic/cloud-on-k8s/pkg/utils/k8s"
+	"github.com/elastic/cloud-on-k8s/pkg/utils/pointer"
+)
+
+func reconcilePodVehicle(podTemplate corev1.PodTemplateSpec, params DriverParams) *reconciler.Results {
```
sebgl, line 25: it would be nice to unit test this function (I'm fine with a follow-up issue)

pkg/controller/beat/controller.go
```diff
@@ -0,0 +1,240 @@
...
+	}); err != nil {
+		return err
+	}
+
+	if err := c.Watch(&source.Kind{Type: &appsv1.Deployment{}}, &handler.EnqueueRequestForOwner{
+		IsController: true,
+		OwnerType:    &beatv1beta1.Beat{},
+	}); err != nil {
+		return err
+	}
```
sebgl, line 87: Maybe a small `// Watch DaemonSets and Deployments` for consistency

pkg/controller/common/container/container.go
```diff
@@ -24,6 +24,8 @@ const (
 	ElasticsearchImage    Image = "elasticsearch/elasticsearch"
 	KibanaImage           Image = "kibana/kibana"
 	EnterpriseSearchImage Image = "enterprise-search/enterprise-search"
+	FilebeatImage         Image = "beats/filebeat"
+	MetricbeatImage       Image = "beats/metricbeat"
```
sebgl, line 28: I was thinking we would have a generic

```go
func BeatImage(beatType beatv1beta1.Type) string {
    return fmt.Sprintf("beats/%s", beatType)
}
```

to be compatible with any official beat image directly.

pkg/controller/common/association/association.go
```diff
@@ -0,0 +1,47 @@
...
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+
+	commonv1 "github.com/elastic/cloud-on-k8s/pkg/apis/common/v1"
+	"github.com/elastic/cloud-on-k8s/pkg/controller/common/certificates"
+	"github.com/elastic/cloud-on-k8s/pkg/utils/k8s"
+)
+
+// WriteAssocSecretToConfigHash dereferences auth secret (if any) to include it in the configHash.
+func WriteAssocSecretToConfigHash(client k8s.Client, assoc commonv1.Association, configHash hash.Hash) error {
```
sebgl, line 19: can we unit test this to catch future regressions?

test/e2e/test/beat/pod_builder.go
```diff
@@ -0,0 +1,238 @@
...
+					}
+				},
+			},
+			test.Step{
+				Name: "Pod should be created",
+				Test: func(t *testing.T) {
+					var createdPod corev1.Pod
+					err := k.Client.Get(k8s.ExtractNamespacedName(&pb.Pod), &createdPod)
+					require.NoError(t, err)
+					// TODO this is incomplete
```
sebgl, line 129: what's missing here?

test/e2e/test/beat/pod_builder.go
```diff
@@ -0,0 +1,238 @@
...
+			},
+		},
+	}
+}
+
+func (pb PodBuilder) RuntimeObjects() []runtime.Object {
+	return []runtime.Object{&pb.Pod}
+}
+
+func (pb PodBuilder) InitTestSteps(k *test.K8sClient) test.StepList {
```
sebgl, line 62: shouldn't all those test steps be attached to the Beat builder? I'm not sure I understand the PodBuilder concept.

test/e2e/test/beat/steps.go
```diff
@@ -0,0 +1,223 @@
...
+						return err
+					}
+				}
+				// wait for Beat pods to disappear
+				if err := k.CheckPodCount(0, test.BeatPodListOptions(b.Beat.Namespace, b.Beat.Name, b.Beat.Spec.Type)...); err != nil {
+					return err
+				}
+
+				// it may take some extra time for Elasticsearch to be fully deleted
+				var beat beatv1beta1.Beat
```
sebgl, line 77: for Beat?

test/e2e/test/beat/steps.go
```diff
@@ -0,0 +1,223 @@
...
+				},
+			},
+			test.Step{
+				Name: "Beat should be created",
+				Test: func(t *testing.T) {
+					var createdBeat beatv1beta1.Beat
+					err := k.Client.Get(k8s.ExtractNamespacedName(&b.Beat), &createdBeat)
+					require.NoError(t, err)
+					require.Equal(t, b.Beat.Spec.Version, createdBeat.Spec.Version)
+					// TODO this is incomplete
```
sebgl, line 110: What's incomplete here?

config/crds/all-crds.yaml
```diff
@@ -457,6 +457,153 @@ status:
...
+    name: health
+    type: string
+  - JSONPath: .status.availableNodes
+    description: Available nodes
+    name: available
+    type: integer
+  - JSONPath: .status.expectedNodes
+    description: Expected nodes
+    name: expected
+    type: integer
```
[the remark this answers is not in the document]
david-kow: I think we did want to change that in the past (https://github.com/elastic/cloud-on-k8s/issues/1786). I'd be for keeping both and improving the status on other resources. It's following the pattern of built-in resources, gives the users more information and it comes free. I'd leave it as is unless you feel strongly about it. (I'm open to naming changes.)

anyasabo:
Just to track from our out of band conversation yesterday, I think we do want to release this as v1alpha1 at first

config/operator/all-in-one/roles.yaml
```diff
@@ -0,0 +1,14 @@
+# permissions needed for Beat cluster-wide autodiscover feature
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: elastic-operator-beat-autodiscover
```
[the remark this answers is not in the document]
david-kow: The intent was to indicate that it came with the operator, but I can see how it can be confusing, I'll change it.

pkg/apis/beat/v1beta1/webhook.go
```diff
@@ -0,0 +1,135 @@
...
+func checkImageIfTypeUnknown(b *Beat) field.ErrorList {
+	knownTypes := []string{"filebeat", "metricbeat"}
+	if !stringsutil.StringInSlice(b.Spec.Type, knownTypes) &&
+		b.Spec.Image == "" {
+		return field.ErrorList{
+			field.Required(
+				field.NewPath("spec").Child("image"),
+				"Image is required if Beat type is not one of [filebeat, metricbeat]"),
+		}
+	}
```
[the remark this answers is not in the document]
david-kow: I think it makes feedback loop much longer, doesn't it? Instead of failing immediately (webhook) or before we do anything we create the deployment, all the pods, all their resources and fail on image pull. We do know that if type is not known the image has to be there. I think it's ok to list few official beats here. WDYT?

pkg/controller/common/container/container.go
david-kow: I'd rather keep all images we use defined in a single place and be explicit. And referencing beat from common seems off to me. WDYT?

test/e2e/test/beat/pod_builder.go
david-kow: PodBuilder is the builder that can be used to build a Pod that will be deployed during the test. Since for some test scenarios we need an actual Pod doing something to validate our Beat deployment.

Not all Beat tests will require a pod (e.g. heartbeat test doesn't) and we may need a different behavior of this pod depending on the test. It may need to be deployed to a different namespace, with different labels, we might need more than a single pod, it might need to log in particular way or expose an endpoint. I'd rather not overload Beat Builder with all of those. I think we have a nice (-ish) abstraction for creating resources and PodBuilder uses it.

config/dev/elastic-psp.yaml
```diff
@@ -51,20 +51,15 @@
   apiVersion: policy/v1beta1
```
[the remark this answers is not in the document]
david-kow: Yes, I'll leave it for a separate PR if that's ok.

pkg/controller/beat/common/reconcile.go
david-kow: Will be done in a separate PR.

config/dev/elastic-psp.yaml
```diff
@@ -51,20 +51,15 @@
...
   spec:
-    privileged: false
-    # Required to prevent escalations to root.
-    allowPrivilegeEscalation: false
-    # This is redundant with non-root + disallow privilege escalation,
-    # but we can provide it for defense in depth.
-    requiredDropCapabilities:
-      - ALL
+    privileged: true
+    allowPrivilegeEscalation: true
```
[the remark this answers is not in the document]
david-kow: Roughly the same (they differ by capabilities, but default ones are enough) is required for all, privileged is not needed. Fixing that.

pkg/apis/beat/v1beta1/webhook.go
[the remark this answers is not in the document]
sebgl: I'm fine with that. Let's list them all, then? (ok in another PR)

pkg/controller/common/container/container.go
sebgl: Relates https://github.com/elastic/cloud-on-k8s/pull/3041#discussion_r435826407.
Let's list all the official Beats here.

pkg/controller/beat/metricbeat/config.go
```diff
@@ -37,36 +37,7 @@ var (
...
-    - pod
-    - container
-    - volume
-  - module: kubernetes
-    period: 10s
-    host: ${NODE_NAME}
-    hosts:
-    - https://${HOSTNAME}:10249
-    metricsets:
-    - proxy
```
sebgl: Can you clarify why we remove those? Is it moving us away from https://raw.githubusercontent.com/elastic/beats/7.7/deploy/kubernetes/metricbeat-kubernetes.yaml?

pkg/apis/beat/v1beta1/webhook.go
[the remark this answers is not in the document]
sebgl: Discussed on Slack: it's about what we support or not. If users specify `type: packetbeat` they may feel like ECK supports running packetbeat, which is not true.

pkg/controller/common/container/container.go
sebgl: Discussed on Slack: it's about what we "support". If the user is allowed to specify `type: packetbeat` they may feel like we officially support it, which isn't true.

pebrc: +1 to supporting the images of all official Beats and not having to manually spec the Docker image for those that don't have explicit support.

pebrc: We could at least support those that work on k8s like auditbeat and heartbeat.

david-kow: I'd like to address it with a separate PR. Current Heartbeat e2e test is more about testing the "otherbeat" deployed than the Heartbeat itself.

david-kow:
Thanks for reviews everyone!

PR #3043, opened 2020-05-12
Add support for the APM Agent Config Management

### LINKED ISSUE #1264: Support APM Agent Config Management:
Starting with 7.3 APM server includes agent config management which requires a Kibana connection. We could facilitate the establishment of that connection in a similar way as we do for the Elasticsearch output by allowing users to simply specify a reference to a Kibana resource. 

We would also need to specify a Kibana user with the correct permissions

barkbay:
This PR adds support for the APM Agent Config Management

I still consider this PR as a draft since it needs some improvements, but I want to share the global design of the implementation:

## Association abstraction

Since a resource may now reference several services `ElasticsearchRef()` and `AssociationConf()` are removed from the `Associated` interface and replaced by a new interface called `Association`:

```golang
type Associated interface {
	metav1.Object
	runtime.Object
	ServiceAccountName() string
	GetAssociations() []Association
}

type Association interface {
       // Remote service reference
	AssociationRef() ObjectSelector

	// Configuration
	AssociationConf() *AssociationConf
	SetAssociationConf(*AssociationConf)

	// Status
	AssociationStatus() AssociationStatus
	SetAssociationStatus(status AssociationStatus)
}
```

## A new controller

The required objects needed to establish the association between the APMServer, Kibana and the Elasticsearch cluster _(the one referenced by Kibana)_ are managed by a new controller. This allows to reuse most of the existing logic.

### Association status

The `associationStatus` field of the `APMServer` is replaced by 2 separate status

```golang
    // ElasticsearchAssociationStatus is the status of any auto-linking to Elasticsearch clusters.
	ElasticsearchAssociationStatus commonv1.AssociationStatus `json:"elasticsearchAssociationStatus,omitempty"`
	// KibanaAssociationStatus is the status of any auto-linking to Elasticsearch clusters.
	KibanaAssociationStatus commonv1.AssociationStatus `json:"kibanaAssociationStatus,omitempty"`
```

### KibanaRef

User can reference a Kibana instance using the `kibanaRef` field:

```yaml
apiVersion: apm.k8s.elastic.co/v1
kind: ApmServer
metadata:
  name: apm-sample
spec:
  version: 7.6.2
  count: 1
  http:
    service:
      spec:
        type: LoadBalancer
  elasticsearchRef:
    name: "elasticsearch-sample"
  kibanaRef:
    name: "kibana-sample"
```

The Kibana instance must reference an Elasticsearch instance managed by ECK. I'm not sure it makes sense to support the case where the Kibana configuration is manually managed (in that case ECK would not be able to create the APM-Kibana user).

## Additional notes

It is important to keep in mind that Elasticsearch can be a transitive dependency: `APM → Kibana → ES`
In the controller logic `associationRef` may be different from `esRef`.

It means that:
* The URL and the CA to be trusted are computed from the `associationRef` _(can be Kibana or Elasticsearch in the case of the APMServer)_ 
* User must be created in `esRef`

TODO:
- [x] Add user garbage collection
- [x] We should support this feature only starting from 7.3
- [x] Check if some unit tests should be improved
- [x] Check the [role to use for the user](https://github.com/elastic/cloud-on-k8s/issues/1264#issuecomment-594008908)
- [x] Update samples
- [ ] Create issue to update the doc
- [ ] Create issue to add some e2e tests.

Fixes #1264


barkbay:
I still have some work to do on the unit tests by I consider this one ready for a review.

pkg/apis/common/v1/association.go
```diff
@@ -27,10 +27,35 @@ const (
...
+	// AssociatedServiceType returns a string describing the type of the target service (elasticsearch most of the time)
+	// It is mostly used to build some other strings depending on the type of the targeted service.
+	AssociatedServiceType() string
+
+	// AnnotationName is the name of the annotation used to define the config for the associated resource.
+	// It is used by the association controller to store the configuration and by the controller which is
+	// managing the associated resource to build the appropriate configuration.
+	AnnotationName() string
+
+	// Remote service reference
```
sebgl, line 51: I would replace `// Remote service reference` to `// Reference to the associated resource`

pkg/apis/apm/v1/apmserver_types.go
```diff
@@ -112,37 +118,118 @@ func (as *ApmServer) IsMarkedForDeletion() bool {
...
+func (aes *ApmEsAssociation) AnnotationName() string {
+	return "association.k8s.elastic.co/es-conf"
+}
+
+func (aes *ApmEsAssociation) AssociatedServiceType() string {
+	return "elasticsearch"
+}
+
+func (aes *ApmEsAssociation) AssociationRef() commonv1.ObjectSelector {
+	return aes.Spec.ElasticsearchRef
```
sebgl, line 167: We seem to call `AssociationRef().WithDefaultNamespace()` in many places, which seems a bit error-prone. Would it be better to have this function return the ref with the correct namespace (also in other associations)?

pkg/controller/association/ca.go
```diff
@@ -24,40 +23,40 @@ type CASecret struct {
...
-// ReconcileCASecret keeps in sync a copy of the Elasticsearch CA.
-// It is the responsibility of the controller to set a watch on the ES CA.
-func (r *Reconciler) ReconcileCASecret(associated commonv1.Associated, es types.NamespacedName) (CASecret, error) {
-	publicESHTTPCertificatesNSN := certificates.PublicCertsSecretRef(esv1.ESNamer, es)
+// ReconcileCASecret keeps in sync a copy of the target service CA.
+// It is the responsibility of the association controller to set a watch on this CA.
+func (r *Reconciler) ReconcileCASecret(association commonv1.Association, namer name.Namer, service types.NamespacedName) (CASecret, error) {
+	servicePublicHTTPCertificatesNSN := certificates.PublicCertsSecretRef(namer, service)
 
 	// retrieve the HTTP certificates from ES namespace
```
sebgl, line 37: we probably want to remove mentions of the `ES` namespace

pkg/controller/association/ca.go
```diff
@@ -24,40 +23,40 @@ type CASecret struct {
...
 			return CASecret{}, nil // probably not created yet, we'll be notified to reconcile later
 		}
 		return CASecret{}, err
 	}
 
-	labels := r.AssociationLabels(k8s.ExtractNamespacedName(associated))
+	labels := r.AssociationLabels(k8s.ExtractNamespacedName(association))
 	// Add the Elasticsearch name, this is only intended to help the user to filter on these resources
-	labels[eslabel.ClusterNameLabelName] = es.Name
+	labels[eslabel.ClusterNameLabelName] = service.Name
```
sebgl, line 48: would this set the ES cluster label to the Kibana TLS cert secret copy?

barkbay:
I'm still working on the role to be used, `apm_user` does not seem to work, see https://github.com/elastic/cloud-on-k8s/issues/1264#issuecomment-630040245

test/e2e/samples_test.go
```diff
@@ -91,7 +92,7 @@ func createBuilders(t *testing.T, decoder *helper.YAMLDecoder, sampleFile, testN
 	return builders
 }
 
-func tweakElasticsearchRef(ref commonv1.ObjectSelector, suffix string) commonv1.ObjectSelector {
+func tweakServiceRef(ref commonv1.ObjectSelector, suffix string) commonv1.ObjectSelector {
```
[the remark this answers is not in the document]
barkbay: because this function is also used to update the name of the Kibana reference

pkg/apis/apm/v1/apmserver_types.go
```diff
@@ -66,8 +69,10 @@ type ApmServerStatus struct {
 	ExternalService string `json:"service,omitempty"`
 	// SecretTokenSecretName is the name of the Secret that contains the secret token
 	SecretTokenSecretName string `json:"secretTokenSecret,omitempty"`
-	// Association is the status of any auto-linking to Elasticsearch clusters.
-	Association commonv1.AssociationStatus `json:"associationStatus,omitempty"`
+	// ElasticsearchAssociationStatus is the status of any auto-linking to Elasticsearch clusters.
+	ElasticsearchAssociationStatus commonv1.AssociationStatus `json:"elasticsearchAssociationStatus,omitempty"`
+	// KibanaAssociationStatus is the status of any auto-linking to a Kibana instance.
```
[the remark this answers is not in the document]
barkbay: Maybe I'm missing your point but this is because the APM Server is only associated to one Kibana instance.

pkg/apis/apm/v1/webhook.go
```diff
@@ -104,3 +109,23 @@ func checkSupportedVersion(as *ApmServer) field.ErrorList {
 func checkNoDowngrade(prev, curr *ApmServer) field.ErrorList {
 	return commonv1.CheckNoDowngrade(prev.Spec.Version, curr.Spec.Version)
 }
+
+func checkAgentConfigurationMinVersion(as *ApmServer) field.ErrorList {
```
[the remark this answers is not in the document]
barkbay: This is already done in the APM controller: https://github.com/elastic/cloud-on-k8s/blob/f2e4ecb670df03ac2a23e91f03d4de389496213f/pkg/controller/apmserver/controller.go#L267-L278

pkg/controller/association/ca.go
barkbay: Yes, good catch. I'm not sure it makes sense to add this label since it is already set on the original Secret. I would be in favor of removing it, wdyt ?

pkg/apis/apm/v1/apmserver_types.go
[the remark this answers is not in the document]
sebgl: I meant that "a Kibana instance" sounds like a single Kibana Pod.

pkg/apis/apm/v1/webhook.go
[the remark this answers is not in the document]
sebgl: Right, I missed that!

pkg/controller/association/ca.go
barkbay: I'm not sure myself actually. If we want to keep this label then it needs to be provided by a new field in the `AssociationInfo`. Not sure it is a bid deal, I can also do that.

sebgl: The label still feels useful to me? I think I'd be ok with adding a new field for it explicitly.

pkg/apis/apm/v1/apmserver_types.go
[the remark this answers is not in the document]
barkbay: ok, got it 😃

pkg/controller/association/resources.go
```diff
@@ -51,20 +52,22 @@ func deleteOrphanedResources(
...
+	esRef commonv1.ObjectSelector,
+	association commonv1.Association,
 ) error {
-	esRef := associated.ElasticsearchRef().WithDefaultNamespace(associated.GetNamespace())
-
-	// Secret should not exist if there is no ES referenced in the spec or if the resource is deleted
-	if !esRef.IsDefined() {
-		return deleteSecret(c, secret, associated)
+	// Secret should not exist if there is no service referenced in the spec or if the resource is deleted
+	serviceRef := association.AssociationRef().WithDefaultNamespace(association.GetNamespace())
```
pebrc, line 59: `serviceRef` threw me off here as I initially thought you meant a k8s Service.

pkg/apis/kibana/v1/kibana_types.go
```diff
@@ -73,7 +72,19 @@ func (k Kibana) IsMarkedForDeletion() bool {
...
+
+func (k *Kibana) AnnotationName() string {
+	return "association.k8s.elastic.co/es-conf"
+}
+
+func (k *Kibana) AssociatedServiceType() string {
+	return "elasticsearch"
+}
+
+func (k *Kibana) AssociationRef() commonv1.ObjectSelector {
```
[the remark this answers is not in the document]
barkbay: This is not an easy change because it would also introduce some changes in the [Kibana driver](https://github.com/elastic/cloud-on-k8s/blob/2841f5eb9002b21c4c73cc2987741768efa27c56/pkg/controller/kibana/driver.go#L202) and in the [configuration generation](https://github.com/elastic/cloud-on-k8s/blob/master/pkg/controller/kibana/config_settings.go#L90) _(same for EnterpriseSearch I guess)_.

My feeling is that iterating over the APM associations in the APMServer controller makes sense. But since there is only one association here I think it's fair for `Kibana` _(or EnterpriseSearch)_ to implement the `Association` interface. It also helps to keep this PR smaller.

pkg/controller/association/reconciler.go
```diff
@@ -52,8 +59,14 @@ type AssociationInfo struct {
 	AssociationLabels func(associated types.NamespacedName) map[string]string
 	// UserSecretSuffix is used as a suffix in the name of the secret holding user data in the associated namespace.
 	UserSecretSuffix string
+	// CASecretServiceLabelName is the name of the label added on the Secret that contains the CA of the remote service.
+	CASecretServiceLabelName string
```
pebrc, line 63: I had to read this a couple of times (CIA Secret, secret service, CIA secret service were among the parse errors my brain made :-) ) when I encountered the name in the code and had to go to the Go doc to understand what the name means ...

pkg/controller/association/ca.go
```diff
@@ -24,40 +22,40 @@ type CASecret struct {
...
+	if err := r.Get(servicePublicHTTPCertificatesNSN, &servicePublicHTTPCertificatesSecret); err != nil {
 		if errors.IsNotFound(err) {
 			return CASecret{}, nil // probably not created yet, we'll be notified to reconcile later
 		}
 		return CASecret{}, err
 	}
 
-	labels := r.AssociationLabels(k8s.ExtractNamespacedName(associated))
+	labels := r.AssociationLabels(k8s.ExtractNamespacedName(association))
 	// Add the Elasticsearch name, this is only intended to help the user to filter on these resources
```
pebrc, line 46: This is not the Elasticsearch name in case of the Kibana association but the Kibana name

pkg/controller/association/ca.go
```diff
@@ -24,40 +22,40 @@ type CASecret struct {
...
-	labels[eslabel.ClusterNameLabelName] = es.Name
+	labels[caSecretServiceName] = service.Name
```
pebrc, line 47: As mentioned above I would be prefer we reserve the name service for actual k8s services. This is the ES cluster name or Kibana name IIUC and not the name of the corresponding services.

How about something like "associated [component|resource] name"

pkg/controller/association/ca.go
```diff
@@ -24,40 +22,40 @@ type CASecret struct {
...
 }
 
-// ReconcileCASecret keeps in sync a copy of the Elasticsearch CA.
-// It is the responsibility of the controller to set a watch on the ES CA.
-func (r *Reconciler) ReconcileCASecret(associated commonv1.Associated, es types.NamespacedName) (CASecret, error) {
-	publicESHTTPCertificatesNSN := certificates.PublicCertsSecretRef(esv1.ESNamer, es)
+// ReconcileCASecret keeps in sync a copy of the target service CA.
+// It is the responsibility of the association controller to set a watch on this CA.
+func (r *Reconciler) ReconcileCASecret(association commonv1.Association, namer name.Namer, service types.NamespacedName, caSecretServiceName string) (CASecret, error) {
+	servicePublicHTTPCertificatesNSN := certificates.PublicCertsSecretRef(namer, service)
```
pebrc, line 34: Not introduced in this PR but I wonder whether we ever considered not copying the secret around if both resources are in the same namespace to reduce the number of secrets we create (and given that this will probably be the common case)

barkbay:
> One thing I am wondering though (inspired by the transitive ES ref code in the association controller) is whether a kibanaRef does not always imply (transitively) an elasticsearchRef and whether we should allow/disallow spec'ing both. What if a user specifies different ES clusters via the two references?

This has been partially discussed [here](https://github.com/elastic/cloud-on-k8s/issues/1264#issuecomment-574624543). IIUC we should not implicitly set an `elasticsearchRef` because a `kibanaRef` is set. I'm not sure there is a reliable way to guess that it is not what the user want (except adding an option to let the user explicitly opt-out from this feature).

An other option would be to implicitly set the `kibanaRef` if the `Elasticsearch` resource is only referenced by one `Kibana`.

Note that it seems that both approaches are optimization and we can add a clever mechanism later.

pebrc:
> This has been partially discussed here. IIUC we should not implicitly set an elasticsearchRef because a kibanaRef is set. I'm not sure there is a reliable way to guess that it is not what the user want (except adding an option to let the user explicitly opt-out from this feature).

Ah, my bad. I completely forgot about that discussion and the fact the APM Server can have an intermediate Logstash, Kafka etc. output. 👍 on the chosen approach.

pkg/controller/association/ca.go
barkbay: 👍 Since this PR is pretty large I'd rather work on it in a dedicated PR.

pebrc: Absolutely!

pkg/controller/apmserver/config.go
```diff
@@ -38,11 +34,15 @@ const (
 	ApmCfgSecretKey = "apm-server.yml" // nolint:gosec
 )
 
+func certificatesDir(serviceType string) string {
```
sebgl, line 37: maybe `certificatesDir(associatedType string)`

pkg/controller/apmserver/config.go
```diff
@@ -66,47 +66,34 @@ func reconcileApmServerConfig(client k8s.Client, as *apmv1.ApmServer) (corev1.Se
 	return reconciler.ReconcileSecret(client, expectedConfigSecret, as)
 }
 
-func NewConfigFromSpec(c k8s.Client, as *apmv1.ApmServer) (*settings.CanonicalConfig, error) {
-	specConfig := as.Spec.Config
-	if specConfig == nil {
-		specConfig = &commonv1.Config{}
+func newConfigFromSpec(c k8s.Client, as *apmv1.ApmServer) (*settings.CanonicalConfig, error) {
+	// Create a base configuration.
```
sebgl, line 70: this comment feels like the first step of something bigger that never comes 😄 
I would either remove it or add other comments for other blocks in this function.

pkg/controller/apmserver/deployment.go
```diff
@@ -86,48 +86,42 @@ func (r *ReconcileApmServer) deploymentParams(
 	podSpec := newPodSpec(as, params)
 	podLabels := NewLabels(as.Name)
 
-	// Build a checksum of the configuration, add it to the pod labels so a change triggers a rolling update
+	// Build a checksum of the configuration and of the cert files used by ES and Kibana, add it to the pod labels so a
+	// change triggers a rolling update. This is done because Apm Server does not support updating its configuration file
+	// or the CA file contents without restarting the process.
```
sebgl, line 91: should probably mention the keystore also

pkg/controller/apmserver/deployment.go
```diff
@@ -86,48 +86,42 @@ func (r *ReconcileApmServer) deploymentParams(
...
-		for i := range podSpec.Spec.Containers {
-			podSpec.Spec.Containers[i].VolumeMounts = append(podSpec.Spec.Containers[i].VolumeMounts, esCAVolume.VolumeMount())
+	for _, association := range as.GetAssociations() {
+		if association.AssociationConf().CAIsConfigured() {
+			caSecretName := association.AssociationConf().GetCASecretName()
+			caVolume := volume.NewSecretVolumeWithMountPath(
+				caSecretName,
+				association.AssociatedType()+"-certs",
+				filepath.Join(ApmBaseDir, certificatesDir(association.AssociatedType())),
+			)
```
sebgl, line 105: not from your PR but since you're at it: can we move the `caVolume` declaration closer to where it's needed a few lines below?

pkg/controller/apmserver/deployment_test.go
```diff
@@ -177,14 +182,24 @@ func expectedDeploymentParams() testParams {
...
 			},
 			Replicas: 0,
 		},
 	}
 }
 
+func withAssociations(as *apmv1.ApmServer, esAssocConf, kbAssocConf *commonv1.AssociationConf) *apmv1.ApmServer {
+	(&apmv1.ApmEsAssociation{
+		ApmServer: as,
+	}).SetAssociationConf(esAssocConf)
```
sebgl, line 196: feels like a constructor could improve readability a bit:
```
apmv1.NewApmEsAssociation(as).SetAssociationConf(esAssocConf)
```

but def not a big deal if we need it only in unit tests.

pkg/controller/association/conf.go
```diff
@@ -23,38 +23,57 @@ import (
...
+func IsConfiguredIfSet(association commonv1.Association, r record.EventRecorder) bool {
+	ref := association.AssociationRef()
+	if (&ref).IsDefined() && !association.AssociationConf().IsConfigured() {
+		r.Event(
+			association,
+			v1.EventTypeWarning,
+			events.EventAssociationError,
+			"Association backend for "+association.AssociatedType()+" is not configured",
+		)
+		log.Info("Association not established: skipping association resource deployment reconciliation",
```
sebgl, line 73: maybe remove `deployment` which can be confusing here

pkg/controller/association/controller/apm_es.go
```diff
@@ -9,39 +9,64 @@ import (
...
+		ExternalServiceURL: getElasticsearchExternalURL,
+		AssociatedNamer:    esv1.ESNamer,
+		AssociationName:    "apm-es",
 		AssociationLabels: func(associated types.NamespacedName) map[string]string {
 			return map[string]string{
-				ApmESAssociationLabelName:      associated.Name,
-				ApmESAssociationLabelNamespace: associated.Namespace,
+				ApmAssociationLabelName:          associated.Name,
+				ApmAssociationLabelNamespace:     associated.Namespace,
+				ApmAssociationTypeLabelNamespace: "elasticsearch",
```
sebgl, line 49: We should probably make `"elasticsearch"` a constant somewhere in `apis/elasticsearch/v1`

pkg/controller/association/controller/ent_es.go
```diff
@@ -25,16 +27,22 @@ const (
...
+		AssociatedShortName: "ent",
 		AssociationLabels: func(associated types.NamespacedName) map[string]string {
 			return map[string]string{
 				EntESAssociationLabelName:      associated.Name,
 				EntESAssociationLabelNamespace: associated.Namespace,
 			}
 		},
-		UserSecretSuffix: "ent-user",
+		UserSecretSuffix:         "ent-user",
+		CASecretServiceLabelName: eslabel.ClusterNameLabelName,
```
sebgl, line 45: `CASecretLabelName` is probably good enough? (with proper doc in the `AssociationInfo` type)

pkg/controller/association/reconciler.go
sebgl: `CASecretLabelName` is probably good enough and does not involve the FBI?

pkg/controller/association/resources.go
```diff
@@ -51,20 +52,22 @@ func deleteOrphanedResources(
...
-	associated commonv1.Associated,
+	esRef commonv1.ObjectSelector,
+	association commonv1.Association,
 ) error {
-	esRef := associated.ElasticsearchRef().WithDefaultNamespace(associated.GetNamespace())
-
-	// Secret should not exist if there is no ES referenced in the spec or if the resource is deleted
-	if !esRef.IsDefined() {
-		return deleteSecret(c, secret, associated)
+	// Secret should not exist if there is no service referenced in the spec or if the resource is deleted
```
sebgl, line 58: no service -> no association

pkg/apis/common/v1/association.go
```diff
@@ -13,24 +13,53 @@ import (
...
 
-// Associated interface represents a Elastic stack application that is associated with an Elasticsearch cluster.
-// An associated object needs some credentials to establish a connection to the Elasticsearch cluster and usually it
-// offers a keystore which in ECK is represented with an underlying Secret.
-// Kibana and the APM server are two examples of associated objects.
+// Associated interface represents an Elastic stack resource that is associated with some other ones.
+// Examples:
+// - Kibana can be associated to Elasticsearch
+// - APMServer can be associated to Elasticsearch and Kibana
+// - EnterpriseSearch can be associated to Elasticsearch
```
pebrc, line 29:
```suggestion
// - Kibana can be associated with Elasticsearch
// - APMServer can be associated with Elasticsearch and Kibana
// - EnterpriseSearch can be associated with Elasticsearch
```

pkg/apis/common/v1/association.go
```diff
@@ -13,24 +13,53 @@ import (
...
 	AssociationPending     AssociationStatus = "Pending"
 	AssociationEstablished AssociationStatus = "Established"
 	AssociationFailed      AssociationStatus = "Failed"
 )
 
-// Associated interface represents a Elastic stack application that is associated with an Elasticsearch cluster.
-// An associated object needs some credentials to establish a connection to the Elasticsearch cluster and usually it
-// offers a keystore which in ECK is represented with an underlying Secret.
-// Kibana and the APM server are two examples of associated objects.
+// Associated interface represents an Elastic stack resource that is associated with some other ones.
```
pebrc, line 25:
```suggestion
// Associated represents an Elastic stack resource that is associated with other stack resources.
```

pkg/controller/association/ca.go
```diff
@@ -24,40 +22,40 @@ type CASecret struct {
...
 	return associated.GetName() + "-" + associationName + "-ca"
 }
 
-// ReconcileCASecret keeps in sync a copy of the Elasticsearch CA.
-// It is the responsibility of the controller to set a watch on the ES CA.
-func (r *Reconciler) ReconcileCASecret(associated commonv1.Associated, es types.NamespacedName) (CASecret, error) {
-	publicESHTTPCertificatesNSN := certificates.PublicCertsSecretRef(esv1.ESNamer, es)
+// ReconcileCASecret keeps in sync a copy of the target service CA.
+// It is the responsibility of the association controller to set a watch on this CA.
+func (r *Reconciler) ReconcileCASecret(association commonv1.Association, namer name.Namer, associatedResource types.NamespacedName, caSecretServiceName string) (CASecret, error) {
```
pebrc, line 33:
```suggestion
func (r *Reconciler) ReconcileCASecret(association commonv1.Association, namer name.Namer, associatedResource types.NamespacedName, associatedResourceLabel string) (CASecret, error) {
```

pkg/controller/association/ca.go
```diff
@@ -24,40 +22,40 @@ type CASecret struct {
...
 		}
 		return CASecret{}, err
 	}
 
-	labels := r.AssociationLabels(k8s.ExtractNamespacedName(associated))
-	// Add the Elasticsearch name, this is only intended to help the user to filter on these resources
-	labels[eslabel.ClusterNameLabelName] = es.Name
+	labels := r.AssociationLabels(k8s.ExtractNamespacedName(association))
+	// Add the associated resource name, this is only intended to help the user to filter on these resources
+	labels[caSecretServiceName] = associatedResource.Name
```
pebrc, line 47:
```suggestion
	labels[associatedResourceLabel] = associatedResource.Name
```

barkbay:
@pebrc @sebgl Thank you for the review !

I'll will create the follow-up issues (doc/e2e tests/optimizations around the `kibanaRef` and the `Secrets`) shortly.

Issue #3054, opened 2020-05-13
Kibana readiness probe must be overridden to work with ingresses


charith-elastic:
We switched the Kibana readiness probe to an `exec` action in #2528. This would affect [users who want to expose Kibana through an ingress](https://discuss.elastic.co/t/kibana-ingress-not-working-after-upgrading-to-eck-1-1-0/232316/2) as some ingress controllers reuse the readiness probe in their own health checks. 

We should document this requirement in the docs and mark it as a breaking change in the 1.1 release notes.

Given that most typical users won't have such strict network controls in place, perhaps we should apply the principle of least astonishment and default to the old behaviour instead?

sebgl:
Fixed in 1.1.1, which reverted to using the same readiness probe (httpGet) as 1.0.x.

PR #3061, opened 2020-05-13
Update breaking changes in 1.1.0

charith-elastic:
Related to #3054
 
Updates the release notes and release highlights to point out that accessing Kibana or APM Server through ingresses might be broken.


docs/release-notes/highlights-1.1.0.asciidoc
```diff
@@ -65,6 +65,58 @@ The concept of `global` and `namespace` operator roles no longer exists, and as
...
 
+[float]
+[id="{p}-110-breaking-readiness-probe"]
+==== Readiness probe change for Kibana and APM Server
+
+Default readiness probes generated by the operator for Kibana and APM Server resources have changed from HTTP GET actions to exec actions. This is to accommodate environments where host-to-pod communication is restricted for security reasons. This change will affect ingress resources that are used to expose Kibana or APM Server instances as some ingress controllers re-purpose the pod readiness probe as a health check. If you are unable to access Kibana or APM Server instances through the ingress after upgrading to ECK 1.1.0, override the readiness probe definitions as follows:
```
anyasabo, line 72:
```suggestion
Default readiness probes generated by the operator for Kibana and APM Server resources have changed from HTTP GET actions to exec actions. This is to accommodate environments where host-to-pod communication is restricted for security reasons. This change will affect ingress resources that are used to expose Kibana or APM Server, as some ingress controllers re-purpose the pod readiness probe as a health check. If you are unable to access Kibana or APM Server through the ingress after upgrading to ECK 1.1.0, override the readiness probe definitions as follows:
```

anyasabo: Small suggestions: adding a `,` and removing "instances" since I'm not sure it is necessary.

I think we want to recommend upgrading to 1.1.1 as the first option, the second option is to override it as described below

charith-elastic: Do we want to refer to a future release from a past release? It feels a bit strange to me. Furthermore, I think we want to get this doc change published immediately so that means the 1.1.1 upgrade option would not be available for a little while longer.

anyasabo: That's a good point. @alaudazzi do you know how we handle this in other parts of the docs? If not then I am good with your plan Charith

charith-elastic:
I am going to merge this for now. If the text needs to be changed, we can always raise another PR.

PR #3067, opened 2020-05-13
Regenerate release notes

anyasabo:


docs/release-notes/1.1.1.asciidoc
```diff
@@ -4,6 +4,14 @@
...
 == {n} version 1.1.1
 
+[[breaking-1.1.1]]
+[float]
+=== Breaking changes
+
+* Revert readiness probes independent of the host/container... (#2528) {pull}3060[#3060]
+
+
+
```
sebgl, line 14: Should we manually remove some empty lines here?

anyasabo: When I looked at it in the past it didn't change the rendering at all so I didn't even open an issue 🤷‍♀️

PR #2926, opened 2020-04-22
Removed unused dependency on godog

pebrc:
@anyasabo made me aware that I missed to run `go mod tidy` after bumping the apm-go-agent dependency. 

Should this be part of our CI checks maybe, so people like me don't forget to run it?

PR #3068, opened 2020-05-13
Run go mod tidy before generating notice

anyasabo:
Ensures we are pruning any unused dependencies as part of PR checks / notice generation. Should catch issues like this https://github.com/elastic/cloud-on-k8s/pull/2926 where we were still shipping unused dependencies

>Tidy makes sure go.mod matches the source code in the module. It adds any missing modules necessary to build the current module's packages and dependencies, and it removes unused modules that don't provide any relevant packages. It also adds any missing entries to go.sum and removes any unnecessary ones.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -86,7 +86,10 @@ dependencies:
 
 # Generate code, CRDs and documentation
 ALL_CRDS=config/crds/all-crds.yaml
-generate: generate-crds generate-api-docs generate-notice-file
+generate: tidy generate-crds generate-api-docs generate-notice-file
+
+tidy:
+	go mod tidy
 
 go-generate:
 	# we use this in pkg/controller/common/license


Makefile
```diff
@@ -86,7 +86,10 @@ dependencies:
 
 # Generate code, CRDs and documentation
 ALL_CRDS=config/crds/all-crds.yaml
-generate: generate-crds generate-api-docs generate-notice-file
+generate: tidy generate-crds generate-api-docs generate-notice-file
+
+tidy:
```
thbkrkr, line 91: I'm fine with `tidy` but wonder if `sync-go-deps` would be better to describe what `go mod tidy` does.

anyasabo: 🤷‍♀️ at least for me i already know `go mod tidy` and would have to think about it if i read sync-go-deps. Maybe let's leave it as is for now and if we add more things like this we will change it? I don't feel strongly either way though

PR #3084, opened 2020-05-15
Avoid overwriting default IPFamily of services

### LINKED ISSUE #3074: spec.ipFamily has to be set in a dual-stack k8s cluster
## Bug Report
**What did you do?**
I deployed the quickstart Elasticsearch ECK example in a dual-stack k8s cluster.

A running quickstart Elasticsearch cluster.

The service resources are not created, because spec.ipFamily is not set explicitly:

```
"error":"Service \"<cluster-name>-es-<node-set-name>\" is invalid: [spec.ipFamily: Invalid value: \"null\": field is immutable, spec.ipFamily: Required value]","errorCauses":[{"error":"Service \"<cluster-name>-es-<node-set-name>\" is invalid: [spec.ipFamily: Invalid value: \"null\": field is immutable, spec.ipFamily: Required value]"}]
```
The elastic-operator creates 3 services: 
* [cluster-name]-es-http
* [cluster-name]-es-transport
* [cluster-name]-es-[node-set-name]

While it is documented how I can set the spec.ipFamily value for the first two (doing which makes it possible to at least run the cluster) the error above remains.

Is there a way to set this value? Can I maybe define the ipFamily for the whole cluster/operator/nodeset?

Not setting spec.ipFamily for the http/transport services prevents the cluster from running. The transport secrets are not created and the node pods are stuck in the init-container waiting for them. Sometimes the nodes manage to start, but still can't see each other.

**Environment**

* ECK version:

	Tried with 1.0.1 and 1.1

* Kubernetes information:
	* On premise 1.17
        * with dual stack 
```
$ kubectl version 1.17
```

* Logs:
```
{"log.level":"error","@timestamp":"2020-05-14T13:22:48.770Z","log.logger":"controller-runtime.controller","message":"Reconciler error","service.version":"1.1.0-29e7447f","service.type":"eck","ecs.version":"1.4.0","controller":"elasticsearch-controller","request":"services/<cluster>","error":"Service \"<cluster>-es-<node-set>\" is invalid: [spec.ipFamily: Invalid value: \"null\": field is immutable, spec.ipFamily: Required value]","errorCauses":[{"error":"Service \"<cluster>-es-<node-set>\" is invalid: [spec.ipFamily: Invalid value: \"null\": field is immutable, spec.ipFamily: Required value]"}],"error.stack_trace":"github.com/go-logr/zapr.(*zapLogger).Error\n\t/go/pkg/mod/github.com/go-logr/zapr@v0.1.0/zapr.go:128\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.5.0/pkg/internal/controller/controller.go:258\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.5.0/pkg/internal/controller/controller.go:232\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).worker\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.5.0/pkg/internal/controller/controller.go:211\nk8s.io/apimachinery/pkg/util/wait.JitterUntil.func1\n\t/go/pkg/mod/k8s.io/apimachinery@v0.17.2/pkg/util/wait/wait.go:152\nk8s.io/apimachinery/pkg/util/wait.JitterUntil\n\t/go/pkg/mod/k8s.io/apimachinery@v0.17.2/pkg/util/wait/wait.go:153\nk8s.io/apimachinery/pkg/util/wait.Until\n\t/go/pkg/mod/k8s.io/apimachinery@v0.17.2/pkg/util/wait/wait.go:88"}
```

charith-elastic:
In dual-stack clusters, the `IPFamily` of the service is not empty. Our reconciliation function compares the `Spec` of the expected against reconciled and considers this to be a difference that requires an update. However, as the `ipFamily` is an immutable field, the update operation will always fail with the following error message:

```
{"service.version": "1.2.0-12a254e1", "controller": "elasticsearch-controller", "request": "default/hulk", "error": "Service \"hulk-es-transport\" is invalid: [spec.ipFamily: Invalid value: \"null\": field is immutable, spec.ipFamily: Required value]", "errorCauses": [{"error": "Service \"hulk-es-transport\" is invalid: [spec.ipFamily: Invalid value: \"null\": field is immutable, spec.ipFamily: Required value]"}]}
```

This PR does the following:
- If the `ipFamily` is not defined in the service spec, it will be defaulted to the `ipFamily` of the reconciled service. This prevents spurious updates that result in the above error.
- If the `ipFamily` is explicitly specified by the user in the service spec and it differs from the current value, it will trigger a re-create operation to delete and create the service again.

Testing this is difficult as currently there are no cloud providers with support for dual-stack clusters. Kind is yet to support it as well. However, as the behaviour is triggered by the mere presence of the dual-stack feature gate, it is possible to configure Minikube to trigger the bug as follows:

```
minikube start --vm-driver=kvm2 --kubernetes-version=1.17.0 --extra-config=apiserver.feature-gates="IPv6DualStack=true" --extra-config=controller-manager.feature-gates="IPv6DualStack=true" --extra-config=kubelet.feature-gates="IPv6DualStack=true" --feature-gates="IPv6DualStack=true" -p k8s117
```

Fixes #3074

### DIFF
--- a/pkg/controller/common/service_control.go
+++ b/pkg/controller/common/service_control.go
@@ -56,8 +56,20 @@ func ReconcileService(
 
 func needsRecreate(expected, reconciled *corev1.Service) bool {
 	applyServerSideValues(expected, reconciled)
-	// ClusterIP is an immutable field
-	return expected.Spec.ClusterIP != reconciled.Spec.ClusterIP
+
+	shouldRecreate := false
+
+	// IPFamily is immutable
+	if expected.Spec.IPFamily != nil && expected.Spec.IPFamily != reconciled.Spec.IPFamily {
+		shouldRecreate = true
+	}
+
+	// ClusterIP is immutable
+	if expected.Spec.ClusterIP != reconciled.Spec.ClusterIP {
+		shouldRecreate = true
+	}
+
+	return shouldRecreate
 }
 
 func needsUpdate(expected *corev1.Service, reconciled *corev1.Service) bool {
@@ -105,6 +117,11 @@ func applyServerSideValues(expected, reconciled *corev1.Service) {
 
 	expected.Annotations = maps.MergePreservingExistingKeys(expected.Annotations, reconciled.Annotations)
 	expected.Labels = maps.MergePreservingExistingKeys(expected.Labels, reconciled.Labels)
+
+	// IPFamily is immutable and cannot be modified so we should retain the existing value from the server if there's no explicit override.
+	if expected.Spec.IPFamily == nil {
+		expected.Spec.IPFamily = reconciled.Spec.IPFamily
+	}
 }
 
 // hasNodePort returns for a given service type, if the service ports have a NodePort or not.
--- a/pkg/controller/common/service_control_test.go
+++ b/pkg/controller/common/service_control_test.go
@@ -62,6 +62,8 @@ func mkService() *corev1.Service {
 }
 
 func Test_needsUpdate(t *testing.T) {
+	ipv4Family := corev1.IPv4Protocol
+
 	type args struct {
 		expected   corev1.Service
 		reconciled corev1.Service
@@ -114,7 +116,6 @@ func Test_needsUpdate(t *testing.T) {
 						Annotations: map[string]string{"annotation1": "annotation1val"},
 					},
 					Spec: corev1.ServiceSpec{
-
 						Type: corev1.ServiceTypeClusterIP,
 					}},
 				reconciled: corev1.Service{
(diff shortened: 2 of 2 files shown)


charith-elastic:
Yes, this just fixes the reconciliation bug reported in #3074. There's no guarantee that IPv6 would work  OOTB with ECK as it is hard to test that easily with existing tools.

Issue #3052, opened 2020-05-12
make check-requisites produces garbled output


jhalterman:
The `make check-requisites` target seems to work well but produces somewhat garbled output currently:

<img width="411" alt="Screen Shot 2020-05-12 at 3 41 13 PM" src="https://user-images.githubusercontent.com/228804/81754556-5794c300-946b-11ea-8bdd-f52182df0828.png">

anyasabo:
I messed this up when fixing all the shellcheck warnings. That said, this is only used in `dev-setup.md` that I can see. I think it may make more sense to just link to installers rather than maintain this? I am not sure how useful it is. If other people think it's useful though I'll go ahead and fix it since it was my mistake.

jhalterman:
It's a useful little util, but I suspect linking to the installers is fine if presented in a concise list of prereqs.

PR #3092, opened 2020-05-15
Bump CI deps

anyasabo:
Bumps some CI dependencies (mostly the linter) and fixes a few annoyances I found along the way

- Unzipping awscli outputs 1k+ lines every CI run for little value, so quiet that down
- Fix https://github.com/elastic/cloud-on-k8s/issues/3052 by using `%b` which interprets escape chars
- Fix some potentially unsafe int parsing (though I don't think we would ever run into it) as described here: https://github.com/kubernetes/kubernetes/issues/81121


.golangci.yml
```diff
@@ -32,11 +37,14 @@ issues:
...
-    # currently it checks the receiver names of even generated files, but is fixed in a later version
     - linters:
-        - stylecheck
-      text: 'ST1016: methods on the same type should have the same receiver name .*'
+        - gosec
+      # this was added in https://github.com/securego/gosec/issues/438 but for our use case is more or less entirely false positives
+      text: 'G601: Implicit memory aliasing in for loop.'
+    - linters:
+        - gosec
+      # TODO: we set some files to 0644, but I think we read them in the same process so 0600 might work?
```
anyasabo, line 46: I could use some advice here, we use 644 in the deployer but I'm not sure if it's necessary

pkg/license/license.go
```diff
@@ -73,6 +73,7 @@ func (r LicensingResolver) ToInfo(totalMemory resource.Quantity) (LicensingInfo,
 }
 
 // Save updates or creates licensing information in a config map
+// This relies on UnconditionalUpdates being supported configmaps and may change in k8s v2: https://github.com/kubernetes/kubernetes/issues/21330
```
anyasabo, line 76: This is just a note for the future since it is currently failing on the new version of controller-runtime (since we are not passing an RV in the update call)

.golangci.yml
thbkrkr: `0600` should be fine as for CI everything is executed in the ci container as root.

pkg/license/license.go
sebgl: Should we just fix the function to do a get + create or update instead? (maybe in a follow-up issue)

anyasabo: I wondered about that, but we really don't care about the current state when updating the configmap so I'm not sure that would be a useful change (except that it would let us use the current fake client and update controller-runtime)

PR #3093, opened 2020-05-15
Add E2E tests to cover breaking changes on secrets and services

sebgl:
Renaming a service, a secret or a secret key is a breaking change since
users may already rely on the naming convention.
Let's catch those in E2E tests.

Fixes https://github.com/elastic/cloud-on-k8s/issues/3036.

PR #3098, opened 2020-05-18
Fix Kibana certs secret labels

sebgl:
We were accidentally using (and importing) APMServer labels instead of
using Kibana labels...

Caught as part of https://github.com/elastic/cloud-on-k8s/pull/3093,
which will include additional tests for those labels names and values.


sebgl:
I think this should be included in `v1.1.2` if we release one.

PR #3101, opened 2020-05-18
Fix remote cluster annotation handling

### LINKED ISSUE #3090: Remote cluster reconciliation results in needless updates
The [`annotateWithCreatedRemoteClusters`](https://github.com/elastic/cloud-on-k8s/blob/02bcd43ecb5d5380bb24bc979f5322d71c380e54/pkg/controller/elasticsearch/remotecluster/annotations.go#L35) function does not check whether the passed in cluster map is empty. That results in the annotation being  always created with its value set to the empty string.  Then, the function immediately updates the Elasticsearch object on the server with this empty annotation. 

Above behaviour of always updating the object seems to result in many conflicts:

```
Could not update remote clusters in Elasticsearch settings	{"service.version": "1.2.0-6a16712c", "namespace": "default", "es_name": "ekesn", "error": "Operation cannot be fulfilled on elasticsearches.elasticsearch.k8s.elastic.co \"ekesn\": the object has been modified; please apply your changes to the latest version and try again"}
```

As the remote cluster reconciliation happens somewhere in the middle of the main reconciliation logic of Elasticsearch, I am not sure whether just fixing the bug in `annotateWithCreatedRemoteClusters` would be enough. When there are legitimate reasons to update the annotation, wouldn't that still cause a conflict if another update is generated by the reconciliation code after this method? This needs further investigation.

charith-elastic:
This fixes two issues with the remote cluster annotation:
- Don't apply the annotation if there are no remote clusters configured. Previously, the annotation was always applied to the Elasticsearch object with the value set to the empty string.
- Ignore empty strings in the annotation value. Previously, parsing the empty string resulted in a cluster named `""` (https://play.golang.org/p/MFB9yfZxxQ1)

Fixes #3090

### DIFF
--- a/pkg/controller/elasticsearch/remotecluster/annotations.go
+++ b/pkg/controller/elasticsearch/remotecluster/annotations.go
@@ -23,7 +23,7 @@ const (
 func getRemoteClustersInAnnotation(es esv1.Elasticsearch) map[string]struct{} {
 	remoteClusters := make(map[string]struct{})
 	serializedRemoteClusters, ok := es.Annotations[ManagedRemoteClustersAnnotationName]
-	if !ok {
+	if !ok || strings.TrimSpace(serializedRemoteClusters) == "" {
 		return remoteClusters
 	}
 	for _, remoteClusterInAnnotation := range strings.Split(serializedRemoteClusters, ",") {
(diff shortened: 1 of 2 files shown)


charith-elastic:
Jenkins test this please

charith-elastic:
Jenkins test this please

pkg/controller/elasticsearch/remotecluster/annotations.go
```diff
@@ -33,14 +33,32 @@ func getRemoteClustersInAnnotation(es esv1.Elasticsearch) map[string]struct{} {
...
+	if len(remoteClusters) == 0 {
+		// if there are no annotations, there's nothing to do
+		if len(es.Annotations) == 0 {
+			return nil
+		}
+
+		// if the annotation exists, delete it
+		if _, ok := es.Annotations[ManagedRemoteClustersAnnotationName]; ok {
+			delete(es.Annotations, ManagedRemoteClustersAnnotationName)
+			return c.Update(&es)
```
barkbay, line 45: nit: I'm wondering if it's worth it to cover this part of the code with a unit test (the current unit tests does not check that the annotation has been actually removed)

charith-elastic:
Jenkins test this please

charith-elastic:
Jenkins test this please

PR #3105, opened 2020-05-18
Use docker.elastic.co rather than push.docker.elastic.co

anyasabo:
It was recommended by the infra team that we use docker.elastic.co rather than push.


fxdgear:
+1 to this. :)

anyasabo:
@thbkrkr do you think it's worth removing entirely? We can do it in this PR or a follow up

anyasabo:
cla/check

thbkrkr:
> @thbkrkr do you think it's worth removing entirely? We can do it in this PR or a follow up

Not really, we still need to handle the authentication to the registry and the `if`s in the Makefile were no better.

PR #3107, opened 2020-05-19
Remove warning about running APMServer and ES in different namespaces

sebgl:
Running APMServer and Elasticsearch in different namespaces is
definitely supported.
We even have an E2E test to cover it.


sebgl:
To backport in the 1.0 and 1.1 branches.

PR #2943, opened 2020-04-23
Add retries for Vault in CI

thbkrkr:
Add a tiny bash function to retry the vault command for a specified
number of times until the command exits successfully with an exponential
backoff.

Resolves #2821.

PR #3117, opened 2020-05-20
Add retries for docker push

### LINKED ISSUE #3106: Retry docker push
`docker push` has failed repeatedly in CI. We may want to wrap it in retries as we did [vault](https://github.com/elastic/cloud-on-k8s/blob/master/.ci/Makefile#L15). We updated the registry host name in https://github.com/elastic/cloud-on-k8s/pull/3105 which may help, but if it persists I think it's worth picking this up.

Example failures:
https://devops-ci.elastic.co/job/cloud-on-k8s-e2e-tests-master/249//testReport
https://devops-ci.elastic.co/job/cloud-on-k8s-e2e-tests-master/251//testReport

thbkrkr:
Reuse the [`retry.sh`](https://github.com/elastic/cloud-on-k8s/blob/master/.ci/common/retry.sh) script added in #2943 to add retries for `docker push`.

Resolves #3106.


thbkrkr:
Good point! I do not remember either but I think it is better to make all of our `docker push` more robust.

anyasabo:
Looks like shellcheck is unhappy atm but when it's resolved LGTM

PR #3121, opened 2020-05-20
Group renovate PRs

anyasabo:
The intent was to only bump k8s dependencies within a major, but I messed up originally because they version them as [`0.18.3`](https://github.com/elastic/cloud-on-k8s/pull/3119) etc. This changes it to:

- only bump k8s dependencies if they are a patch. minors can be handled with the controller-runtime bumps
- group k8s dependencies together into one PR
- ignore digest dependencies for k8s.io/utils. Currently it will open a PR for every commit, because the repo does not include any tags. I asked [in slack](https://kubernetes.slack.com/archives/C09R23FHP/p1590000806367800), if that will change, but no response yet. This means we will likely need to manually update this dependency if we want to, but I think that's okay since our use of this module is pretty limited at the moment (parsing ports safely [here](https://github.com/elastic/cloud-on-k8s/blob/master/pkg/dev/portforward/service_forwarder.go#L97)).

Config options described [here](https://docs.renovatebot.com/configuration-options/#digest)

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -16,18 +16,26 @@
         "digest"
       ],
       "enabled": false
-    }
+    },
     {
       "packageNames": [
         "k8s.io/apimachinery",
         "k8s.io/client-go",
-        "k8s.io/api",
-        "k8s.io/utils"
+        "k8s.io/api"
       ],
       "updateTypes": [
         "patch"
       ],
       "groupName": "k8s"
+    },
+    {
+      "packageNames": [
+        "k8s.io/utils"
+      ],
+      "updateTypes": [
+        "digest"
+      ],
+      "enabled": false
     }
   ]
 }
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -11,9 +11,23 @@
         "k8s.io/api"
       ],
       "updateTypes": [
-        "major"
+        "major",
+        "minor",
+        "digest"
       ],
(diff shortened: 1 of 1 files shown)


.github/renovate.json
```diff
@@ -11,9 +11,23 @@
...
+        "digest"
       ],
       "enabled": false
     }
+    {
+      "packageNames": [
+        "k8s.io/apimachinery",
+        "k8s.io/client-go",
+        "k8s.io/api",
+        "k8s.io/utils"
```
pebrc, line 25: Not sure I understand the config options fully, but why it `k8s.io/utils` not in the disabled list above?

anyasabo: I messed up. It doesn't need to be part of that list because it is versioned independently (or rather, not at all), but it did need to have its digest patches disabled. If they ever start releasing versioned updates then we will get them

anyasabo:
I wonder if we want to add x/crypto to the digest exclude list to avoid PRs like https://github.com/elastic/cloud-on-k8s/pull/3176

But I wonder if we do want to get crypto updates

pebrc:
> wonder if we want to add x/crypto to the digest exclude list to avoid PRs like #3176
> 
> But I wonder if we do want to get crypto updates

Yes I would think we would. But maybe let's discuss on this other PR?

PR #3124, opened 2020-05-21
Add manifest generator

### LINKED ISSUE #2406: Add tooling to customize the all-in-one manifests

charith-elastic:
Adds tooling to generate installation manifests for supported configurations of the operator.

Fixes #2406


hack/manifest-gen/assets/charts/eck/templates/_helpers.tpl
```diff
@@ -1,11 +1,9 @@
-# The operator has cluster-wide permissions on all required resources.
```
sebgl, line 1: could this file be named `_rbac_rules.tpl`? (is `_` a convention in helm?)

hack/manifest-gen/assets/charts/eck/templates/operator-role-binding.yaml
```diff
@@ -0,0 +1,17 @@
...
+  namespace: {{ .Values.operator.namespace }}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: {{ .Values.operator.name }}
+subjects:
+- kind: ServiceAccount
+  name: {{ .Values.operator.name }}
+  namespace: {{ .Values.operator.namespace }}
+{{- end -}}
```
sebgl, line 17: isn't it done already in `managed-ns-role-bindings.yaml`? Why is this a special case?

hack/manifest-gen/assets/charts/eck/values.yaml
```diff
@@ -0,0 +1,92 @@
...
+
+  # Labels to be added to the operator. Useful if multiple operators are to be deployed to the cluster.
+  # These labels are used by the webhook service if it is enabled as well.
+  selectorLabels:
+    control-plane: elastic-operator
+
+  # Container image options.
+  image:
+    # Full path to the operator container image excluding the tag. The tag value is derived from the operator.version field.
+    repository: docker.elastic.co/eck/eck-operator
```
sebgl, line 18: `image.repository` sounds like the docker registry to me.
How about:
```yaml
image: docker.elastic.co/eck/eck-operator
imagePullPolicy: IfNotPresent
```

I can imagine people search for `imaagePullPolicy` directly since it's the k8s name already.

hack/manifest-gen/assets/charts/eck/values.yaml
```diff
@@ -0,0 +1,92 @@
...
+  resources:
+    limits:
+      cpu: 1
+      memory: 150Mi
+    requests:
+      cpu: 100m
+      memory: 50Mi
+
+# Optional configuration values
+config:
```
sebgl, line 32: shouldn't this be under `operator` since it is the operator config (its flags, basically)?
We may need it if there's more than an operator at some point.

hack/manifest-gen/assets/charts/eck/values.yaml
```diff
@@ -0,0 +1,92 @@
...
+    pullPolicy: IfNotPresent
+
+  # Resource and request limits for the operator pod.
+  resources:
+    limits:
+      cpu: 1
+      memory: 150Mi
+    requests:
+      cpu: 100m
+      memory: 50Mi
```
sebgl, line 29: We changed those defaults recently: https://github.com/elastic/cloud-on-k8s/pull/3046

hack/manifest-gen/assets/charts/eck/values.yaml
```diff
@@ -0,0 +1,92 @@
...
+
+  # Parallelism of the operator.
+  maxConcurrentReconciles: 3
+
+  # Experimental APM tracing configuration for the operator.
+  tracing:
+    enabled: false
+    config:
+      ELASTIC_APM_SERVER_URL: http://localhost:8200
+      ELASTIC_APM_SERVER_TIMEOUT: 30s
```
sebgl, line 60: would be nice to also include/document `ELASTIC_APM_SERVER_TOKEN` here. Maybe keep it commented?

```yaml
      ELASTIC_APM_SERVER_URL: http://localhost:8200
      ELASTIC_APM_SERVER_TIMEOUT: 30s
      # ELASTIC_APM_SERVER_TOKEN: <token>
```

hack/manifest-gen/assets/charts/eck/values.yaml
```diff
@@ -0,0 +1,92 @@
...
+    validity: 8760h
+    # How long before expiry should the certificates be renewed.
+    rotateBefore: 24h
+
+  # Enable RBAC enforcement for cross-namespace references between Elastic Stack applications.
+  refs:
+    enforceRBAC: false
+
+  webhook:
+    # Enable the webhook.
```
sebgl, line 81: The flag comment is a bit more explicit: `Enables a validating webhook server in the operator process` - I think we at least want to keep "validating" here.

hack/manifest-gen/assets/charts/eck/values.yaml
```diff
@@ -0,0 +1,92 @@
...
+    enabled: true
+    # Allow the operator to create and manage the certificates for the webhook.
+    manageCerts: true
+    # Location of the webhook certificates. Should only be overridden if config.webhook.manageCerts is false and you are providing your own webhook configuration.
+    certsDir: "/tmp/k8s-webhook-server/serving-certs"
+    # Name of the Kubernetes secret containing the webhook certificates. Should only be overridden if config.webhook.manageCerts is false and you are providing your own webhook configuration.
+    secret: elastic-webhook-server-cert
```
sebgl, line 88: maybe `certsSecret`?

hack/manifest-gen/assets/charts/eck/values.yaml
```diff
@@ -0,0 +1,92 @@
...
+    # Name of the webhook.
+    name: elastic-webhook.k8s.elastic.co
+    # Name of the service created for the webhook.
+    serviceName: elastic-webhook-server
```
sebgl, line 92: should this be `service` (since we have `secret` and not `secretName`)? Or the other way around 🙃

hack/manifest-gen/go.mod
```diff
@@ -0,0 +1,9 @@
+module github.com/elastic/cloud-on-k8s/hack/manifest-gen
+
+go 1.14
+
+require (
+	github.com/spf13/cobra v1.0.0
+	helm.sh/helm/v3 v3.2.1
+	rsc.io/letsencrypt v0.0.3 // indirect
```
sebgl, line 8: 😮 transitive dependency from helm?

hack/manifest-gen/go.sum
```diff
@@ -0,0 +1,745 @@
...
+sigs.k8s.io/apiserver-network-proxy/konnectivity-client v0.0.7/go.mod h1:PHgbrJT7lCHcxMU+mDHEm+nx46H4zuuHZkDP6icnhu0=
+sigs.k8s.io/kustomize v2.0.3+incompatible h1:JUufWFNlI44MdtnjUqVnvh29rR37PQFzPbLXqhyOyX0=
+sigs.k8s.io/kustomize v2.0.3+incompatible/go.mod h1:MkjgH3RdOWrievjo6c9T245dYlB5QeXV4WCbnt/PEpU=
+sigs.k8s.io/structured-merge-diff/v3 v3.0.0-20200116222232-67a7b8c61874/go.mod h1:PlARxl6Hbt/+BC80dRLi1qAmnMqwqDg62YvvVkZjemw=
+sigs.k8s.io/structured-merge-diff/v3 v3.0.0 h1:dOmIZBMfhcHS09XZkMyUgkq5trg3/jRyJYFZUiaOp8E=
+sigs.k8s.io/structured-merge-diff/v3 v3.0.0/go.mod h1:PlARxl6Hbt/+BC80dRLi1qAmnMqwqDg62YvvVkZjemw=
+sigs.k8s.io/yaml v1.1.0/go.mod h1:UJmg0vDUVViEyp3mgSv9WPwZCDxu4rQW1olrI1uml+o=
+sigs.k8s.io/yaml v1.2.0 h1:kr/MCeFWJWTwyaHoR9c8EjH9OumOmoF9YGiZd7lFm/Q=
+sigs.k8s.io/yaml v1.2.0/go.mod h1:yfXDCHCao9+ENCvLSE62v9VSji2MKu5jeNfTrofGhJc=
+vbom.ml/util v0.0.0-20160121211510-db5cfe13f5cc/go.mod h1:so/NYdZXCz+E3ZpW0uAoCj6uzU2+8OWDFv/HxUSs7kI=
```
sebgl, line 745: So many deps it looks like Javascript 😄

hack/manifest-gen/main.go
```diff
@@ -0,0 +1,145 @@
...
+	cmd := buildCmd()
+	if err := cmd.Execute(); err != nil {
+		log.Printf("Error: %v", err)
+		os.Exit(1)
+	}
+}
+
+func buildCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:           "manifest-gen",
```
sebgl, line 32: I feel like `manifest` would be good enough? So we run `manifest generate` instead of `manifest-gen generate`

hack/manifest-gen/main.go
```diff
@@ -0,0 +1,145 @@
...
+Restricted operator managing "elastic-system", "nsa" and "nsb":
+    $ manifest-gen generate --profile=restricted --set=config.managedNamespaces='{elastic-system, nsa, nsb}'
+
+Restricted operator with tracing configured:
+    $ manifest-gen generate --profile=restricted --set=config.tracing.enabled=true --set=config.tracing.config.ELASTIC_APM_SERVER_URL=http://apm:8200
+`
+
+	cmd := &cobra.Command{
+		Use:           "generate",
+		Short:         "Generate ECK manifests",
```
sebgl, line 107: We're a bit inconsistent in using `manifests` vs. `manifest`. Not sure what's best? Let's try to settle on one.

hack/manifest-gen/main.go
```diff
@@ -0,0 +1,145 @@
...
+		SilenceErrors: true,
+		SilenceUsage:  true,
+	}
+
+	sourceDir, err := findSourceDir()
+	if err != nil {
+		log.Fatalf("Failed to locate assets: %v", err)
+	}
+
+	cmd.PersistentFlags().StringVar(&sourceFlag, "source", sourceDir, "Source directory")
```
sebgl, line 44: would it be more explicit to call it `charts`?

hack/manifest-gen/main.go
```diff
@@ -0,0 +1,145 @@
...
+    $ manifest-gen generate --set=operator.resources.limits.cpu=2,operator.resources.limits.memory=300Mi
+
+Restricted operator without CRDs, managing the "elastic-system" namespace:
+    $ manifest-gen generate --profile=restricted --exclude-crds
+
+Restricted operator installed to and managing the single namespace named "namespacex":
+    $ manifest-gen generate --profile=restricted --set=operator.namespace=namespacex --set=config.managedNamespaces='{namespacex}'
+
+Restricted operator managing "elastic-system", "nsa" and "nsb":
+    $ manifest-gen generate --profile=restricted --set=config.managedNamespaces='{elastic-system, nsa, nsb}'
```
sebgl, line 99: is it helm convention use `{a, b, c}` instead of `[a, b, c]` for lists?

hack/manifest-gen/manifest-gen.sh
```diff
@@ -0,0 +1,58 @@
...
+
+    cp -f "$ALL_CRDS" "${CHART_DIR}/crds/all-crds.yaml"
+    sed -i -E "s#version: [0-9]+\.[0-9]+\.[0-9]+#version: $VERSION#" "${CHART_DIR}/values.yaml"
+    sed -i -E "s#appVersion: [0-9]+\.[0-9]+\.[0-9]+#appVersion: $VERSION#" "${CHART_DIR}/Chart.yaml"
+}
+
+usage() {
+    echo "Usage: $0 [-u | -g <args>]"
+    echo "    '-u'"
+    echo "         Update the chart and exit"
```
sebgl, line 27: it's immediately clear to me what "update the chart" means, can we clarify a bit?

hack/manifest-gen/internal/manifest.go
```diff
@@ -0,0 +1,117 @@
...
+	ExcludeCRDs bool
+}
+
+// OptionsFlags holds flag values for the options operation.
+type OptionsFlags struct {
+	Source string
+}
+
+// Generate produces a manifest for installing ECK using an embedded Helm chart.
+func Generate(opts *GenerateFlags) error {
```
sebgl, line 37: It'd be nice to have a few comments explaining how this relies on helm internally and differs (or not?) from what helm would normally do if you invoke it directly.

hack/manifest-gen/README.md
```diff
@@ -0,0 +1,44 @@
+ECK Manifest Generator
```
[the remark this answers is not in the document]
pebrc: I agree with the general direction you are outlining @sebgl 
But does it change much for this PR, I don't think so.

hack/manifest-gen/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+```shell
+make docker-gen-global | kubectl apply -f -
+```
+
+Installation restricted to a single namespace named "myns":
+
+```shell
+make docker-gen-restricted RESTRICTED_NS=myns | kubectl apply -f -
+```
```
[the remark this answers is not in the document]
charith-elastic: It is more likely that users will have Docker installed on their machines rather than Go. Asking them to install Go and compile a binary is not a great user experience. That is one of the reasons why I was keen on the distribution model enabled by my initial proposal.

With the code staying inside the cloud-on-k8s repo,  the options as I see them are:
1. Tell users to build the manifest-gen container image locally
2. Make the manifest-gen container image available in the public registry
3. Bundle the manifest-gen binary with the ECK container image

hack/manifest-gen/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+Quickstart
+----------
+
+NOTE: Requires Docker and Make to be installed on your computer.
+
+Cluster-wide installation with access to all namespaces:
+
+```shell
+make docker-gen-global | kubectl apply -f -
```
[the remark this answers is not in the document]
charith-elastic: The script is simply there to update the chart with the latest CRDs and version. People can use the CLI directly if they wish but I think the main audience for the tool is non-developers who would prefer a simpler command to run rather than a complicated compilation + run recipe.

hack/manifest-gen/README.md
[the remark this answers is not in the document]
charith-elastic: That was my original proposal too :)

hack/manifest-gen/assets/charts/eck/.helmignore
```diff
@@ -0,0 +1,23 @@
+# Patterns to ignore when building packages.
```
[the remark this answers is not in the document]
charith-elastic: Both. It's a proper Helm chart that can be used by itself if someone is brave enough.

hack/manifest-gen/assets/charts/eck/profile-restricted.yaml
```diff
@@ -0,0 +1,7 @@
+config:
```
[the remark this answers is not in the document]
charith-elastic: `namespaced` is pretty obtuse compared to `restricted` IMO.

hack/manifest-gen/assets/charts/eck/templates/_helpers.tpl
```diff
@@ -1,11 +1,9 @@
...
-# Same resources as the namespace operator, except for the addition of:
-# - validating|mutatingwebhookconfigurations
-apiVersion: rbac.authorization.k8s.io/v1
-kind: ClusterRole
-metadata:
-  name: elastic-operator
-rules:
+{{/* vim: set filetype=mustache: */}}
```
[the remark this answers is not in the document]
charith-elastic: It's only a comment. Your editor can ignore it if it doesn't understand it.

hack/manifest-gen/README.md
```diff
@@ -0,0 +1,44 @@
...
+```shell
+make docker-build DOCKER_IMAGE=manifest-gen
+```
+
+Run the following command to view documentation and examples of customizing manifests:
+
+```shell
+docker run manifest-gen generate --help
+```
+
```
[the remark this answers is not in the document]
charith-elastic: I thought we were not officially supporting Helm -- hence the need for this in the first place. Otherwise we could just publish an official Helm chart, can't we?

hack/manifest-gen/assets/charts/eck/values.yaml
charith-elastic: It's normal Helm convention.

hack/manifest-gen/README.md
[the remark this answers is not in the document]
sebgl: I think I'd be fine with releasing the manifest-gen along with the ECK image, especially if they are versioned together (your option 3)?

However I think we should also consider:

4. Release the manifest-gen as a standalone binary (windows/osx/linux) in our github releases

I imagine most users would either run it from the ECK Docker image, or from the standalone binary.
I would consider that users who run things from the source code after a git clone can manage installing Go?

I'm fine keeping things as they are if you think it makes more sense for now, and revisit once we settle on the release/distribution mechanism.

hack/manifest-gen/README.md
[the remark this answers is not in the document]
sebgl: > the main audience for the tool is non-developers

But in that case they would probably run either form a released docker image or released binary directly, not from cloning the github repo and running a make command?

hack/manifest-gen/assets/charts/eck/profile-restricted.yaml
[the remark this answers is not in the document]
sebgl: @elastic/cloud-k8s curious what other think here? `restricted` feels very misleading to me but that could be because english isn't my 1st language.

hack/manifest-gen/README.md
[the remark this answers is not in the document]
sebgl: Yeah you're right.

hack/manifest-gen/assets/charts/eck/profile-restricted.yaml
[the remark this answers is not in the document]
david-kow: I'm -1 on `global` and +1 on `cluster-wide`. I'm not sure about a `namespaced` as it doesn't map well to k8s usage where it means belonging to a single namespace. For us it's one or more namespaces.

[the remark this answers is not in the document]
barkbay: Should we really care if we can make it an implementation detail and hide this wording to the user:

```bash
> # Generate and install a "global", "cluster-wide" operator
> make | kubectl apply -f -
```

```bash
> # Generate and install an operator which only manages some namespaces
> make MANAGED_NAMESPACES="foo,bar" | kubectl apply -f -
```

I'm not sure there is a right wording, OLM has [installation mode types](https://docs.openshift.com/container-platform/4.2/operators/understanding_olm/olm-understanding-operatorgroups.html#olm-operatorgroups-membership_olm-understanding-operatorgroups), but as long as it is not exposed to the user we can change our mind later.

_by the way it not clear what the default target actually does_

hack/manifest-gen/README.md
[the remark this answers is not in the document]
barkbay: +1 to discuss an other release mechanism later (`curl`, `brew install elastic/tap/eckctl`, `yum install eckctl` ...)

hack/manifest-gen/assets/charts/eck/Chart.yaml
```diff
@@ -0,0 +1,12 @@
...
+description: A Helm chart for deploying Elastic Cloud on Kubernetes
+type: application
+
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
+version: 0.0.1
+
+# This is the version number of the application being deployed. This version number should be
+# incremented each time you make changes to the application.
+appVersion: 1.2.0
```
barkbay, line 12: IIUC this field should be kept in sync with https://github.com/elastic/cloud-on-k8s/blob/master/VERSION ?

hack/manifest-gen/assets/charts/eck/profile-restricted.yaml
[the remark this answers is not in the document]
sebgl: Great point @barkbay! Either we have managed namespaces, either we don't (default), which seems good enough to switch on what yaml is required. So we could remove the `profile` variable entirely?

hack/manifest-gen/assets/charts/eck/values.yaml
```diff
@@ -0,0 +1,92 @@
+# Core configuration values
+operator:
+  # Name of the operator installation.
+  name: elastic-operator
+  # Namespace to install the operator to.
+  namespace: elastic-system
+  # Version of the operator. Should be set to the desired container image tag.
+  version: 1.2.0
```
barkbay, line 8: Could we not reuse `{{ .Chart.AppVersion }}` ?

hack/manifest-gen/assets/charts/eck/Chart.yaml
charith-elastic: It is: https://github.com/charith-elastic/cloud-on-k8s/blob/6f3f2bee22fff6742d29b9d5e5b0685cad3a6186/hack/manifest-gen/manifest-gen.sh#L14

hack/manifest-gen/assets/charts/eck/values.yaml
charith-elastic: That makes it tricky to test unreleased images. With an explicit setting, I can test a locally built snapshot by passing something like `--set=operator.version=1.2.0-48b3bc4e --set=operator.image.repository=docker.elastic.co/eck-dev/eck-operator-cell`

hack/manifest-gen/assets/charts/eck/profile-restricted.yaml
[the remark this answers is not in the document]
pebrc: I think `global` and `restricted` are OK names in my ears. But I am no native speaker. The question whether we need a profile feature which as I understand it also controls whether we have a webhook or not and sets up the managed namespaces at the moment is almost separate from that. 

Right now it seems that we predicate that decision indeed on if and how many managed namespace the user defines. But I wonder if some form of configuration preset will be useful once this thing generates additional config /RBAC for Beats for example (we can of course add it then when we need it)

hack/manifest-gen/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM golang as builder
```
thbkrkr, line 1: It might be good to pin a version to avoid surprises.
```suggestion
FROM golang:1.14 as builder
```

hack/manifest-gen/Makefile
```diff
@@ -0,0 +1,23 @@
+RESTRICTED_NS:=elastic-system
+DOCKER_IMAGE:=docker.elastic.co/eck/manifest-gen:$(shell cat ../../VERSION)
```
thbkrkr, line 2: Nit: add spaces and align values.
```suggestion
RESTRICTED_NS := elastic-system
DOCKER_IMAGE  := docker.elastic.co/eck/manifest-gen:$(shell cat ../../VERSION)
```

hack/manifest-gen/Makefile
```diff
@@ -0,0 +1,23 @@
...
+	@ ./manifest-gen.sh -g --profile=restricted --set=operator.namespace=$(RESTRICTED_NS) --set=config.managedNamespaces="{$(RESTRICTED_NS)}"
+
+.PHONY: docker-build
+docker-build:
+	@ ./manifest-gen.sh -u
+	@ docker build -t $(DOCKER_IMAGE) .
+
+.PHONY: docker-gen-global
+docker-gen-global: docker-build
+	@ docker run $(DOCKER_IMAGE) generate --profile=global
```
thbkrkr, line 19:
```suggestion
	@ docker run --rm $(DOCKER_IMAGE) generate --profile=global
```

hack/manifest-gen/Makefile
```diff
@@ -0,0 +1,23 @@
...
+
+.PHONY: docker-gen-restricted
+docker-gen-restricted: docker-build
+	@ docker run $(DOCKER_IMAGE) generate --profile=restricted --set=operator.namespace=$(RESTRICTED_NS) --set=config.managedNamespaces="{$(RESTRICTED_NS)}"
```
thbkrkr, line 23:
```suggestion
	@ docker run --rm $(DOCKER_IMAGE) generate --profile=restricted --set=operator.namespace=$(RESTRICTED_NS) --set=config.managedNamespaces="{$(RESTRICTED_NS)}"
```

hack/manifest-gen/manifest-gen.sh
```diff
@@ -0,0 +1,58 @@
...
+CHART_DIR="${SCRIPT_DIR}/assets/charts/eck"
+
+update_chart() {
+    local ALL_CRDS="${SCRIPT_DIR}/../../config/crds/all-crds.yaml"
+    local VERSION
+    VERSION=$(cat "${SCRIPT_DIR}/../../VERSION")
+
+    cp -f "$ALL_CRDS" "${CHART_DIR}/crds/all-crds.yaml"
+    sed -i -E "s#version: [0-9]+\.[0-9]+\.[0-9]+#version: $VERSION#" "${CHART_DIR}/values.yaml"
+    sed -i -E "s#appVersion: [0-9]+\.[0-9]+\.[0-9]+#appVersion: $VERSION#" "${CHART_DIR}/Chart.yaml"
```
david-kow, line 21: This ends up with the below on MacOSX.
```
$ ls -a assets/charts/eck/Chart*
assets/charts/eck/Chart.yaml    assets/charts/eck/Chart.yaml-E
$ ls assets/charts/eck/values*
assets/charts/eck/values.yaml   assets/charts/eck/values.yaml-E
```

hack/manifest-gen/main.go
```diff
@@ -0,0 +1,144 @@
...
+available configuration options. The behaviour of these flags is identical to the similarly named flags used by Helm.
+See https://helm.sh/docs/intro/using_helm/ for more information.
+
+By default, the operator is installed into the "elastic-system" namespace. This can be overridden by setting the
+"operator.namespace" option.
+
+`
+	examples := `
+Global operator:
+    $ eck manifest generate
```
david-kow, line 83: Should these be `manifest-gen generate`?

hack/manifest-gen/manifest-gen.sh
charith-elastic: Interesting... looks like `-i` is consuming `-E` as the backup suffix. It shouldn't do that though. What is the version of sed you have on your machine?

david-kow: Can't check the version unfortunately. Apparently it's [FreeBSD sed from 2005](https://unix.stackexchange.com/a/131940). For [.ci/Makefile](https://github.com/elastic/cloud-on-k8s/blob/b5dc5166aa35a079dd01b7ba9a0954589cc17605/.ci/Makefile#L27) we've ended up with `sed ... > tmp && mv tmp ...`.

hack/manifest-gen/assets/charts/eck/values.yaml
anyasabo: I see there's this related issue about allowing to set appversion without editing the chart yaml https://github.com/helm/helm/issues/8194

hack/manifest-gen/assets/charts/eck/profile-restricted.yaml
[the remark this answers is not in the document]
anyasabo: I think the different OLM install mode types are pretty clear what they do so could see us adopting them, but I think we can leave that as something for later

hack/manifest-gen/main.go
anyasabo: yes, https://helm.sh/docs/intro/using_helm/#the-format-and-limitations-of---set

hack/manifest-gen/manifest-gen.sh
anyasabo: As far as I know in place edits that work on both gnu sed and freebsd aren't possible without something like that Makefile david linked

hack/manifest-gen/assets/charts/eck/profile-restricted.yaml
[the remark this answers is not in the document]
anyasabo: As discussed out of band today it might be worth merging this in as is and opening a separate issue to discuss this as a refinement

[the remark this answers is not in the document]
charith-elastic: Although currently the profiles seems like installation modes, they are more than that. The idea is to provide a set of defaults than can then be individually customised. They also serve as documentation/examples for what can be achieved with the operator. For example, someone might want to install the operator to manage just a single namespace but with the webhook. Without profiles, that person would need to read and understand the whole Helm chart and construct a complicated command line with the "right" combination of parameters. However, with a profile, all that person has to do is to look at the profile definition that closely matches what they want and simply override the flag that disables the webhook.

hack/manifest-gen/assets/charts/eck/values.yaml
charith-elastic: It's the name of the service that would be created and not the actual service itself so I think `serviceName` is more appropriate.

hack/manifest-gen/main.go
charith-elastic: I guess it's the awkwardness of using the plural form for names. `manifests-gen` doesn't sound right -- hence why sometimes `manifest` is used in places.

charith-elastic:
Jenkins test this please

PR #3125, opened 2020-05-21
Add 1.1.2 release notes

anyasabo:
Also updates 1.1.0 release notes since I forgot to tag a specific PR for that release when I backported it 🤦‍♀️

### DIFF
--- a/docs/release-notes/highlights-1.1.2.asciidoc
+++ b/docs/release-notes/highlights-1.1.2.asciidoc
@@ -1,0 +1,20 @@
+[[release-highlights-1.1.2]]
+== 1.1.2 release highlights
+
+[float]
+[id="{p}-112-new-and-notable"]
+=== New and notable
+This release contains bug fixes especially relevant for
+
+* users with custom certificates configured
+* users deploying ECK into dual-stack Kubernetes clusters
+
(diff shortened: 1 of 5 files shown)


docs/release-notes/highlights-1.1.2.asciidoc
```diff
@@ -0,0 +1,20 @@
...
+* users with custom certificates configured
+* users deploying ECK into dual-stack Kubernetes clusters
+
+as well as some minor bug fixes.
+
+[float]
+[id="{p}-112-breaking-changes"]
+=== Upgrade notes
+
+In ECK 1.1.0, users using custom certificates would be required to specify a cerificate authority (CA), even if the CA was present in the system certificates. This was a regression in 1.1.0 and has been fixed in 1.1.2. Users with custom certificates that are signed by a well-known CA will now function without the user specifying the CA explicitly.
```
pebrc, line 18:
```suggestion
In ECK 1.1.0, users using custom certificates would be required to specify a certificate authority (CA), even if the CA was present in the system certificates. This was a regression in 1.1.0 and has been fixed in 1.1.2. Users with custom certificates that are signed by a well-known CA will now function without the user specifying the CA explicitly.
```

PR #3144, opened 2020-05-26
Generic association proposal

barkbay:
The goal of this PR is to describe and discuss a way to make it easier for a user application (i.e. not managed by ECK) to establish a trusted connection with an Elastic stack application.

Sorry for reviving this directory, but I was not sure about the right way to discuss the ideas I have in mind for a "generic" association controller.

I used to find the PR format convenient to discuss some architectural choices, but I'm happy to create an issue if you think it is more convenient (or once we are done with this one).

### DIFF
--- a/docs/design/0012-generic-association.md
+++ b/docs/design/0012-generic-association.md
@@ -1,0 +1,275 @@
+# Generic Application Association
+
+## Definitions
+
+  - In the rest of this document an **Elastic stack application** is an application managed by ECK. For instance, it could be Elasticsearch, the APM Server, Kibana or EnterpriseSearch.
+
+  - An **associated application** describes any user application, running in a `Pod`, in a namespace managed by ECK.
+
+## Context and Problem Statement
+
+Establishing a trusted connection to an Elastic stack application requires gathering and keeping in sync multiple moving parts. Each Elastic stack application comes with its certificates and different kind of credentials. A credential could be a token, in the case of the APM Server, or a set of user/password. Manually copying certificates, credentials and tokens [can be error prone](https://github.com/elastic/cloud-on-k8s/issues/3032). Moreover, if they are not kept in sync the association could be broken.
+
+This document describes how ECK could make available to an associated application the information needed to establish a trusted association with an Elastic stack application.
+
+It is worth to note that the problem could be divided into 2 different sub-problems:
+
+  - A first problem is to let a user specify the association information required by one or several of its applications, and make it available in the application namespace.
+  - A second problem is how ECK can automatically inject this information, either by managing a part of the `Spec` of a top-level deployment resource _(`Deployment`, `DaemonSet`, `StatefulSet`, `Job`)_ or by mutating the `Pod` spec before its creation.
+
+## Decision Drivers
+
+### Security
+
+A user should not be able to create an association without the agreement of the Elastic stack application owner. It should not be possible to create an Elastic user with an arbitrary role if a user does not have the ownership of the Elastic stack application. Similarly, it should not be possible for an Elastic stack application owner to modify the spec of a `Pod` without the agreement of the associated application owner.
+
+### Pod spec management
+
+Ideally the solution would be able to update the `Pod` specification, injecting the association data automatically and making it available through some "well-known" environment variables or files in the `Pod`. This would allow the user to delegate the binding of the association variables to ECK. This behavior is similar to what is described in the issue [Facilitate ECK integration with MutatingAdmissionWebhook](https://github.com/elastic/cloud-on-k8s/issues/2418).
+
+## Artifacts involved in an association
+
+### Certificates
+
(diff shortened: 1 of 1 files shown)


docs/design/0012-generic-association.md
```diff
@@ -0,0 +1,275 @@
...
+        image: ...
+      - name: container2
+        image: ...
+```
+
+## Making association variables available
+
+### The Association Secret
+
+The proposal here is to define the content of the association secret. This `Secret` is managed by a generic association controller. According to what the user specify in the generic association custom resource this `Secret` may be filled with the following items:
```
sebgl, line 149: This only works in situations where ECK is managing the target namespace.
vs. a situation where ECK manages resources in namespaceA but the end-user application runs in namespaceB.
In that case, a cluster-wide secret-sync operator, decoupled from ECK, could make more sense? (something like https://github.com/IBM-Cloud/kube-samples/tree/master/secret-sync-operator)

docs/design/0012-generic-association.md
```diff
@@ -0,0 +1,275 @@
...
+      - name: elastic-volume
+        secret:
+          secretName: <automatic-generated-name>
+```
+
+  * Pros:
+    * The configuration checksum can be injected in the pod template to allow an automatic restart of the `Pods` 
+    * ECK can inject and manage some "well-known" environment variables, some good examples are the ones used in the context of the APM Server (`ELASTIC_APM_SERVER_URLS`, `ELASTIC_APM_SECRET_TOKEN`...) 
+  * Cons:
+    * It could be considered as "intrusive"
```
sebgl, line 239: ++ I don't like much the idea of ECK updating the spec of other Pods/Deployments/StatefulSets outside the scope of a MutatingAdmissionWebhook.

docs/design/0012-generic-association.md
```diff
@@ -0,0 +1,275 @@
...
+    * It could be a lot of resources to watch
+
+#### Option #2: Using a MutatingAdmissionWebhook
+
+  * Pros:
+    * Less intrusive than updating the parent resource
+  * Cons:
+    * Cannot handle the lifecycle of the Secret content. The associated application must detect than something has changed in the `Secret`. In case of an environment variable the application must handle a `Pod` restart itself.
```
sebgl, line 247: Would it make sense to optionally (opt-in annotation) inject a liveness probe that would fail if the secret gets updated? Based on an additional "checksum" entry in the secret maybe.

sebgl, line 247: I was looking at how  https://www.vaultproject.io/docs/platform/k8s/injector deals with this problem. Looks like they inject files, not environment variables. Maybe for that particular reason 😄 
It's a general problem with k8s secrets.

docs/design/0012-generic-association.md
```diff
@@ -0,0 +1,275 @@
...
+  ## targetNamespaces are the namespaces where the association information are copied and reconciled
+  targetNamespaces:
+    - ns1
+
+  ## These are the Elastic applications.
+  ## The namespace is automatically inherited from this custom resource.
+  elasticsearchName: monitor
+  kibanaName: monitor
+  # apmServerName: monitor
+
```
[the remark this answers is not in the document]
barkbay: The rational here was to allow a quick review of the existing associations (for example in case of a security audit) and to be allow to have an alpha release and improve this feature over the time (decoupled lifecycle with the stack applications). But this could be discussed.

docs/design/0012-generic-association.md
```diff
@@ -0,0 +1,275 @@
...
+The new CRD allows creating and maintaining a copy of the information needed to establish an association. But, it might be worth to propose a mechanism to inject the variables automatically. Optionally, it would be possible to add some annotations to automatically inject the required data into the `Pods` _(in the targeted namespaces)_ :
+
+```yaml
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: my-application
+  annotations:
+    association.k8s.elastic.co/name: "monitor-ns/monitor" # namespaced name of the association CR
```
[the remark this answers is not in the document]
barkbay: In the original approach an association is a top level component which already allows to select which components are accessed by the remote application. That being said your idea is still relevant in that case because it would allow to have a better granularity from the application point of view.

docs/design/0012-generic-association.md
barkbay: True if we only want to share a Secret (and I agree it could be a first step). But if we want to update/mutate some resources we must manage the target namespace anyway.

docs/design/0012-generic-association.md
barkbay: I have the same feeling, it sounds odd. But I'm not sure I have a technical argument against it 🤷‍♂️

barkbay:
@sebgl thanks for your feedback

I'll try to split this proposal into smaller issues.

PR #3145, opened 2020-05-26
Add licence usage metrics

### LINKED ISSUE #3140: Expose elastic-licensing related usage data as a prometheus metric
Usage data about elastic-licensing might be also worth exposing as Prometheus metrics for admins.
This will prevent users from inadvertently violating limits.
https://www.elastic.co/guide/en/cloud-on-k8s/master/k8s-licensing.html#k8s_getting_usage_data

_Originally posted by @dulltz in https://github.com/elastic/cloud-on-k8s/issues/212#issuecomment-572850413_

charith-elastic:
Exposes licence usage information as Prometheus metrics.

Metrics output:

```
# HELP elastic_licensing_enterprise_resource_units_total Total enterprise resource units used
# TYPE elastic_licensing_enterprise_resource_units_total gauge
elastic_licensing_enterprise_resource_units_total{license_level="enterprise_trial"} 1
# HELP elastic_licensing_memory_gigabytes_total Total memory used in GB
# TYPE elastic_licensing_memory_gigabytes_total gauge
elastic_licensing_memory_gigabytes_total{license_level="enterprise_trial"} 7.516192768
```

ConfigMap output:

```
data:
  eck_license_level: enterprise_trial
  enterprise_resource_units: "1"
  timestamp: "2020-05-26T12:40:10Z"
  total_managed_memory: 7.52GB
```

Fixes #3140

### DIFF
--- a/pkg/license/reporter.go
+++ b/pkg/license/reporter.go
@@ -55,7 +55,7 @@ func (r ResourceReporter) Start(refreshPeriod time.Duration) {
 	}
 }
 
-// Report reports the licensing information in a config map
+// Report licensing information by publishing metrics and updating the config map.
 func (r ResourceReporter) Report() error {
 	licensingInfo, err := r.Get()
 	if err != nil {
--- a/pkg/license/reporter.go
+++ b/pkg/license/reporter.go
@@ -62,6 +62,7 @@ func (r ResourceReporter) Report() error {
 		return err
 	}
 
+	licensingInfo.ReportAsMetrics()
 	return r.licensingResolver.Save(licensingInfo)
 }
 
--- a/pkg/license/reporter_test.go
+++ b/pkg/license/reporter_test.go
@@ -21,155 +21,191 @@ import (
 	esv1 "github.com/elastic/cloud-on-k8s/pkg/apis/elasticsearch/v1"
 	kbv1 "github.com/elastic/cloud-on-k8s/pkg/apis/kibana/v1"
 	commonlicense "github.com/elastic/cloud-on-k8s/pkg/controller/common/license"
-	essettings "github.com/elastic/cloud-on-k8s/pkg/controller/elasticsearch/settings"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/kibana"
 	"github.com/elastic/cloud-on-k8s/pkg/utils/k8s"
 )
 
 const operatorNs = "test-system"
 
-func Test_Get(t *testing.T) {
-	es := esv1.Elasticsearch{
-		Spec: esv1.ElasticsearchSpec{
-			NodeSets: []esv1.NodeSet{{
-				Count: 10,
-			}},
-		},
-	}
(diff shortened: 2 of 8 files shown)


anyasabo:
jenkins test this please

pkg/utils/metrics/metrics.go
```diff
@@ -0,0 +1,58 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
```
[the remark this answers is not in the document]
anyasabo: I was wondering that too, but it also might be my personal twitch I get for anything in a `utils` directory

pkg/license/reporter.go
```diff
@@ -62,6 +62,7 @@ func (r ResourceReporter) Report() error {
 		return err
 	}
 
+	licensingInfo.ReportAsMetrics()
```
anyasabo, line 65: The comment on this func could use an update since it's doing more now

pkg/license/reporter_test.go
```diff
@@ -21,155 +21,191 @@ import (
...
-			NodeSets: []esv1.NodeSet{{
-				Count: 10,
-			}},
-		},
-	}
-	licensingInfo, err := NewResourceReporter(k8s.FakeClient(&es), operatorNs).Get()
-	assert.NoError(t, err)
-	assert.Equal(t, "21.47GB", licensingInfo.TotalManagedMemory)
-	assert.Equal(t, "1", licensingInfo.EnterpriseResourceUnits)
+func TestGet(t *testing.T) {
```
anyasabo, line 30: I realize both are not directly from this PR, but:

It doesn't look like we have a test exercising the maxERUs branch, correct? It might be worth adding one if possible

Nit: Since they don't share anything, at this point it might be simpler to break each case out into separate tests just to make it easier to run a specific test if you need to

anyasabo:
LGTM, nice work. Looks like most recent CI failure should be fixed with https://github.com/elastic/cloud-on-k8s/pull/3157

pkg/utils/metrics/metrics.go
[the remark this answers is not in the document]
charith-elastic: I am planning on adding more metrics soon. Other metrics are a bit harder to integrate because the code is not organised with instrumentation in mind so I'll be doing that incrementally in a series of PRs. 

I prefer the all-metrics-in-a-single-place approach as it is easier to audit and helps to implicitly enforce conventions on new metrics -- which is harder to do when they are strewn all over the place.

pkg/license/reporter_test.go
charith-elastic: There are tests for maxERUs in the licence tests.

I am not sure I understand your comment about individual tests. You can already run specific tests by invoking something like `go test -run='TestGet/elasticsearch_defaults' ./pkg/license/...`

charith-elastic:
Jenkins test this please

pkg/license/reporter_test.go
anyasabo: You're right, I was being goofy 🤦‍♀️

Issue #3147, opened 2020-05-27
Statefulset metadata Annotations removed by operator


oliveagle:
## Bug Report
**What did you do?**

Create a ElasticSearch cluster with 3 nodes.   add some annotations in sts metadata.  then scale up the cluster.   Annotation we added in sts metadata disappered. 

Add arbitrary annotations in sts metadata

![image](https://user-images.githubusercontent.com/2270205/82979678-3fe83d80-a01a-11ea-8c9d-5f4f04a2e272.png)

Do these following operations.  they all will remove annotations in sts metadata
* scale up/down
* replace image version of ElasticSearch

**What did you expect to see?**

We need those annotations to be persistent.(operator don't clean customized annotations in sts metadata). 

**Environment**

* ECK version:

    version: 1.1.1   
    hash:   b6f69fc4262838547cc4fb782048c8d415403730

We added 3 line of code that solve this problem. Here are the changes. 

![image](https://user-images.githubusercontent.com/2270205/82979446-a3be3680-a019-11ea-8de4-1558829afbde.png)

sebgl:
Related: https://github.com/elastic/cloud-on-k8s/issues/2652 to include the sset annotations in the Elasticsearch resource

PR #3149, opened 2020-05-27
Preserve extra annotations and labels on StatefulSet updates

sebgl:
Some users may "manually" set additional labels and annotations on existing
Elasticsearch StatefulSets.
When we update those it makes sense to preserve the existing labels and
annotations, and only override the ones we manage. This is a logic we
already have in place for other resource kinds.

Fixes https://github.com/elastic/cloud-on-k8s/issues/3147.


pkg/controller/elasticsearch/sset/reconcile_test.go
```diff
@@ -13,6 +13,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/utils/pointer"
```
barkbay, line 16: TIL

Issue #3154, opened 2020-05-28
Add e2e tests for APM / Kibana association


barkbay:
Following-up #3043 we should add some end to end tests to validate that an APM Server is correctly associated with Kibana.

barkbay:
I think we can implement the following scenario:

1. Create an agent configuration using the Kibana API, the API [is officially documented](https://www.elastic.co/guide/en/kibana/7.7/agent-config-api.html#apm-update-config) from 7.7.0 but is available since 7.5.0

    `PUT https://x.x.x.x:5601/api/apm/settings/agent-configuration`

    The following HTTP body creates the default agent configuration _(a.k.a. `All`)_ :

    ```json
    {
        "service": {},
        "settings": {
            "transaction_sample_rate": 1,
            "capture_body": "errors",
            "transaction_max_spans": 100
        }
    }
    ```

2. Fetch the configuration from the APM Server with the [Agent configuration API](https://www.elastic.co/guide/en/apm/server/current/agent-configuration-api.html)

    `GET https://y.y.y.y:8200/config/v1/agents?service.name=All`
    
    Compare the results:
    
    ````json
    {
      "capture_body": "errors",
      "transaction_max_spans": "100",
      "transaction_sample_rate": "1"
    }
    ````

pebrc:
Very good idea!

PR #3155, opened 2020-05-28
Introduce a restricted role for agent configuration

barkbay:
While doing some manual testing with the APM Agent configuration I realized that the `kibana_user` role used in #3043  does not work for 7.3 and 7.4

This PR: 
* Set 7.5.1 as the minimum version required to use agent configuration. This feature is GA since 7.5 and supporting older versions would [require to setup](https://github.com/elastic/apm-server/issues/3596#issuecomment-634872770) some dedicated roles.
* Introduce a restricted role to be used by the agent configuration user:

```yaml
eck_apm_agent_user_role:
  cluster: []
  indices: []
  applications:
  - application: kibana-.kibana
    privileges:
    - feature_apm.read
    resources:
    - space:default
```

This PR has been tested with 7.5.2 , 7.6.2 and 7.7.0

### DIFF
--- a/pkg/apis/apm/v1/webhook_test.go
+++ b/pkg/apis/apm/v1/webhook_test.go
@@ -97,20 +97,20 @@ func TestWebhook(t *testing.T) {
 			Operation: admissionv1beta1.Create,
 			Object: func(t *testing.T, uid string) []byte {
 				apm := mkApmServer(uid)
-				apm.Spec.Version = "7.2.0"
+				apm.Spec.Version = "7.4.0"
 				apm.Spec.KibanaRef = commonv1.ObjectSelector{Name: "kbname", Namespace: "kbns"}
 				return serialize(t, apm)
 			},
 			Check: test.ValidationWebhookFailed(
-				`spec.kibanaRef: Forbidden: required version for Kibana association is 7.3.0 but desired version is 7.2.0`,
+				`spec.kibanaRef: Forbidden: required version for Kibana association is 7.5.0 but desired version is 7.4.0`,
 			),
 		},
 		{
-			Name:      "support-72-if-kibana-ref-not-set",
+			Name:      "support-74-if-kibana-ref-not-set",
 			Operation: admissionv1beta1.Create,
 			Object: func(t *testing.T, uid string) []byte {
 				apm := mkApmServer(uid)
-				apm.Spec.Version = "7.2.0"
+				apm.Spec.Version = "7.4.0"
 				return serialize(t, apm)
 			},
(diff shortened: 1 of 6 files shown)


pkg/controller/elasticsearch/user/roles.go
```diff
@@ -58,6 +61,17 @@ var (
 				},
 			},
 		},
+		ApmAgentUserRole: esclient.Role{
+			Cluster: []string{},
+			Indices: []esclient.IndexRole{},
+			Applications: []esclient.ApplicationRole{
```
barkbay, line 67: @simitt @sqren would be glad to have your review on this role. Thanks 🙏

sqren: This lgtm.
fyi: In parallel I'm working on improving this by making this privilege default for the  `apm_user` role: https://github.com/elastic/elasticsearch/issues/57201.

It might not be available until 8.0 (so still quite a long way out) but better than never.

barkbay:
While working on the e2e tests I noticed that something seems to be broken in `7.5.0` While I had no issues with `7.5.1` and `7.5.2`. (could be related to https://github.com/elastic/apm-server/pull/3031 ?)

pkg/apis/apm/v1/webhook_test.go
```diff
@@ -97,20 +97,20 @@ func TestWebhook(t *testing.T) {
...
 			Object: func(t *testing.T, uid string) []byte {
 				apm := mkApmServer(uid)
-				apm.Spec.Version = "7.2.0"
+				apm.Spec.Version = "7.4.0"
 				apm.Spec.KibanaRef = commonv1.ObjectSelector{Name: "kbname", Namespace: "kbns"}
 				return serialize(t, apm)
 			},
 			Check: test.ValidationWebhookFailed(
-				`spec.kibanaRef: Forbidden: required version for Kibana association is 7.3.0 but desired version is 7.2.0`,
+				`spec.kibanaRef: Forbidden: required version for Kibana association is 7.5.1 but desired version is 7.4.0`,
```
anyasabo, line 105: Late nit: should read "minimum required version..."

anyasabo:
Late comment: I wonder if we want to update our compatibility docs. I think we might not have to since it was only GA in 7.5

pkg/apis/apm/v1/webhook_test.go
barkbay: Thanks for the feedback, I'll fix this as part of https://github.com/elastic/cloud-on-k8s/issues/3154

PR #3158, opened 2020-05-28
Add an info log on dev builds of ECK

pebrc:
Creates a log message when an ECK build is neither licensed with a dev public key nor a production key to facilitate debugging of license issues.

I ran more than once into the issue that I was testing license related functionality with a PR build or local build that did not have any kind of public key. This should make this condition more obvious than 

```
"asn1: syntax error: sequence truncated",
```

instead we get
```
This is an unlicensed development build of ECK. License functionality and Enterprise features are disabled
```


anyasabo:
jenkins test this please

anyasabo:
Looks like that unit test does need to be updated

PR #3161, opened 2020-05-28
Switch to distroless

anyasabo:
We had previously decided on centos to standardize with the rest of elastic products and to set us up for OpenShift certification. Neither of those apply at this time, but we do semi-frequently field issues about vulnerabilities in our image that show up on scans. These largely (entirely?) do not affect us, but there is still a maintenance burden in reviewing the results and responding. There is also a build time burden of needing to update all of the packages on every image build. Since we do not use any of the packages this change would make both our lives and our security team's lives easier.

Note: if we merge this we will also want to update the image dependencies documentation. It's kind of a pain though because it doesn't run on OSX so I did not do it yet since I wasn't sure if we would want to merge this.


pcsanwald:
It's hard to argue with this change given your points on k8s itself switching, but, if we wanted to err on the side of caution, it could be worth a broader email just saying "hey we intend to do this, please raise any concerns" just to give field/infra teams a heads up, as this will effect scanning (in a good way hopefully). That could be unnecessary, though, would defer to y'alls judgement.

Issue #3163, opened 2020-05-29
E2E TestEnterpriseTrialLicense fails


sebgl:
```
=== RUN   TestEnterpriseTrialLicense/Elasticsearch_license_should_be_[basic]
Retries (5m0s timeout): ....................................................................................................
    TestEnterpriseTrialLicense/Elasticsearch_license_should_be_[basic]: utils.go:84: 
        	Error Trace:	utils.go:84
        	Error:      	Received unexpected error:
        	            	expectedTypes license type [ basic] got platinum
        	Test:       	TestEnterpriseTrialLicense/Elasticsearch_license_should_be_[basic]
{"log.level":"error","@timestamp":"2020-05-29T01:38:11.724Z","message":"stopping early","service.version":"0.0.0-00000000","service.type":"eck","ecs.version":"1.4.0","error":"test failure","error.stack_trace":"github.com/go-logr/zapr.(*zapLogger).Error\n\t/go/pkg/mod/github.com/go-logr/zapr@v0.1.0/zapr.go:128\ngithub.com/elastic/cloud-on-k8s/test/e2e/test.StepList.RunSequential\n\t/go/src/github.com/elastic/cloud-on-k8s/test/e2e/test/step.go:43\ngithub.com/elastic/cloud-on-k8s/test/e2e/es.TestEnterpriseTrialLicense\n\t/go/src/github.com/elastic/cloud-on-k8s/test/e2e/es/license_test.go:130\ntesting.tRunner\n\t/usr/local/go/src/testing/testing.go:991"}
    --- FAIL: TestEnterpriseTrialLicense/Elasticsearch_license_should_be_[basic] (300.00s)
```

https://devops-ci.elastic.co/job/cloud-on-k8s-e2e-tests-stack-versions/101/testReport/

sebgl:
So I managed to run the test locally (after spending some time figuring out which pubkey/privkey/license I need to extract), and... it succeeded! :|
It seemed to reliably fail in CI though.
Still investigating.

sebgl:
From the operator logs:

```
{"log.level":"info","@timestamp":"2020-05-29T01:38:35.699Z","log.logger":"license-controller","message":"Starting reconciliation run","service.version":"1.2.0-57fdf901","service.type":"eck","ecs.version":"1.4.0","iteration":279,"namespace":"e2e-7uo4r-mercury","es_name":"test-es-trial-license-wk7b"}
[...]

PR #3165, opened 2020-05-29
Replace the license observer by a sync call in the reconciliation

sebgl:
Fixes https://github.com/elastic/cloud-on-k8s/issues/3163.
Fixes a bug introduced in https://github.com/elastic/cloud-on-k8s/pull/3150.

We used to rely on the license fetched by the observers on a regular
basis, but this can lead to a race condition as described in
#3163 (comment).

Instead, let's keep things simple, and do a synchronous request to
get the license as part of the normal reconciliation, to then decide
whether or not it should be updated.

In case of error (eg. Elasticsearch req timeout), enqueue a new
reconciliation, but don't abort the current reconciliation.

This PR also adds some minor refactoring of the code, and avoids
a call to `start_basic` if the current license is already basic.

### DIFF
--- a/pkg/controller/elasticsearch/license/apply.go
+++ b/pkg/controller/elasticsearch/license/apply.go
@@ -54,23 +54,26 @@ func applyLinkedLicense(
 		},
 		&license,
 	)
-	if err != nil {
-		if apierrors.IsNotFound(err) {
-			if isTrial(current) {
-				// Elasticsearch reports a trial license, but there's no ECK enterprise trial requested.
-				// This can be the case if:
-				// - an ECK trial was started previously, then stopped (secret removed)
-				// - the user manually started a trial at the stack level (eg. by clicking a button in Kibana when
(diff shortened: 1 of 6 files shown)


pkg/controller/elasticsearch/license/apply.go
```diff
@@ -23,42 +23,65 @@ import (
...
+			// nothing to do
+			return nil
+		case isTrial(current):
+			// Elasticsearch reports a trial license, but there's no ECK enterprise trial requested.
+			// This can be the case if:
+			// - an ECK trial was started previously, then stopped (secret removed)
+			// - the user manually started a trial at the stack level (eg. by clicking a button in Kibana when
+			// trying to access a commercial feature). While this is not a supported use case,
+			// we tolerate it to avoid a bad user experience because trials can only be started once.
+			log.V(1).Info("Preserving existing stack-level trial license",
```
pebrc, line 78: Somehow 'preserving' sounded odd in my ears, but I have no better suggestion :-(

Issue #2936, opened 2020-04-22
PDB per Node type (Master, Data, Injest)


akrzos-bw:
## PodDisruptionBudget per node type (master, data, injest, ...)

If we are performing maintenance on a kubernetes cluster (cordon, drain, updates/upgrades, then reboot) an ECK cluster can end up blocking the upgrade for a lengthy period of time while the ECK cluster goes into yellow status while a data node is removed from service but replaced on a separate node.  This disruption often causes re-balancing of shards and now the Elastic PDB prevents other ECK cluster pods from being evicted even though they may not have an impact on data loss such as a master pod.  If we had a PDB per node type we could mitigate some of the nodes that would be blocked while waiting for the ECK cluster to return to healthy green state.

It also appears that regardless of the underlying data redundancy the cluster is limited to a single disruption at this point.

anyasabo:
Just to address your last point, the `maxUnavailable` setting can be tuned to modify that behavior: https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-update-strategy.html

sebgl:
`updateStrategy.maxUnavailable` only impacts rolling upgrades of the cluster, not disruptions related to the PodDisruptionBudget.

Note that you can also [set your own PodDisruptionBudget](https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-pod-disruption-budget.html), or override the default one. For your use case I think you may want something similar to the following:

```yaml
apiVersion: elasticsearch.k8s.elastic.co/v1
kind: Elasticsearch
metadata:
  name: quickstart
spec:
  version: 7.6.2
  nodeSets:
  - name: default
    count: 3
  podDisruptionBudget: {} # disable the default PDB
---
apiVersion: policy/v1beta1
kind: PodDisruptionBudget
metadata:
  name: data-nodes-pdb
spec:
  minAvailable: 2 # should probably be (number of data nodes - 1), I'm considering you have 3 data nodes here
  selector:
    matchLabels:
      elasticsearch.k8s.elastic.co/cluster-name: your-es-cluster
      elasticsearch.k8s.elastic.co/node-data: true
      elasticsearch.k8s.elastic.co/node-master: false
---
apiVersion: policy/v1beta1
kind: PodDisruptionBudget
metadata:
  name: master-nodes-pdb
spec:
  minAvailable: 2 # should probably be (number of master nodes - 1), I'm considering you have 3 master nodes here
  selector:
    matchLabels:
[...]

PR #3167, opened 2020-05-30
Allow single-node clusters disruption in the default PDB

sanderma:
See issue [PDB per Node type (Master, Data, Ingest) #2936](https://github.com/elastic/cloud-on-k8s/issues/2936)

This PR will make sure that when someone deploys a single node cluster, the pdb does not get set to 1.

If someone deploys a single node cluster now, it can make sure the k8s cluster cannot reboot.
The default setting for single node, IMHO, should not set the pdb to 1.
This change makes sure that when it is single node, the pdb is 0. When the elastic cluster is scaled up, the pdb is calculated as "normal".

> If you set maxUnavailable to 0% or 0, or you set minAvailable to 100% or the number of replicas, you are requiring zero voluntary evictions. When you set zero voluntary evictions for a workload object such as ReplicaSet, then you cannot successfully drain a Node running one of those Pods. If you try to drain a Node where an unevictable Pod is running, the drain never completes. This is permitted as per the semantics of PodDisruptionBudget.
[source](https://kubernetes.io/docs/tasks/run-application/configure-pdb/#specifying-a-poddisruptionbudget)

### DIFF
--- a/pkg/controller/elasticsearch/pdb/reconcile.go
+++ b/pkg/controller/elasticsearch/pdb/reconcile.go
@@ -141,6 +141,11 @@ func buildPDBSpec(es esv1.Elasticsearch, statefulSets sset.StatefulSetList) v1be
 
 // allowedDisruptions returns the number of Pods that we allow to be disrupted while keeping the cluster healthy.
 func allowedDisruptions(es esv1.Elasticsearch, actualSsets sset.StatefulSetList) int32 {
+	if actualSsets.ExpectedNodeCount() == 1 {
+		// single node cluster (not highly-available)
+		// allow the node to be disrupted to ensure K8s nodes operations can be performed
+		return 1
+	}
 	if es.Status.Health != esv1.ElasticsearchGreenHealth {
 		// A non-green cluster may become red if we disrupt one node, don't allow it.
 		// The health information we're using here may be out-of-date, that's best effort.
--- a/pkg/controller/elasticsearch/pdb/reconcile_test.go
+++ b/pkg/controller/elasticsearch/pdb/reconcile_test.go
@@ -308,12 +308,12 @@ func Test_allowedDisruptions(t *testing.T) {
 			want: 1,
 		},
 		{
-			name: "green health but single-node cluster: 0 disruption allowed",
+			name: "single-node cluster (not high-available): 1 disruption allowed",
 			args: args{
 				es:          esv1.Elasticsearch{Status: esv1.ElasticsearchStatus{Health: esv1.ElasticsearchGreenHealth}},
 				actualSsets: sset.StatefulSetList{sset.TestSset{Replicas: 1, Master: true, Data: true}.Build()},
 			},
-			want: 0,
+			want: 1,
 		},
 		{
 			name: "green health but only 1 master: 0 disruption allowed",


pkg/controller/elasticsearch/pdb/reconcile.go
```diff
@@ -116,12 +116,20 @@ func expectedPDB(es esv1beta1.Elasticsearch, statefulSets sset.StatefulSetList,
...
-	// maybe allow some Pods to be disrupted
-	minAvailable := nodeCount - allowedDisruptions(es, statefulSets)
+	var minAvailable int32
+	if nodeCount == 1 {
+		// when only single node, allow the pod to be rebooted, to ensure Kubernetes nodes do not get hung up by the pdb.
+		minAvailable = 0
+	} else {
+		// maybe allow some Pods to be disrupted
+		minAvailable = nodeCount - allowedDisruptions(es, statefulSets)
+	}
```
sebgl, line 132: I think it would make sense to have this block as part of the `allowedDisruptions` function where the logic around checking health and number of nodes exists.
This function returns the number of nodes that can safely be disrupted. In this situation I think it could check:

```go
if actualSsets.ExpectedNodeCount() == 1 {
    // single node cluster (not highly-available)
    // allow the node to be disrupted to ensure K8s nodes operations can be performed
    return 1
}
```

Also this function is fairly easy to unit test. I think it has an existing unit test that you probably want to modify:

```go
{
    name: "green health but single-node cluster: 0 disruption allowed",
    args: args{
        es:          esv1.Elasticsearch{Status: esv1.ElasticsearchStatus{Health: esv1.ElasticsearchGreenHealth}},
				actualSsets: sset.StatefulSetList{sset.TestSset{Replicas: 1, Master: true, Data: true}.Build()},
        },
        want: 0,
},
```

sanderma: Good point, in my haste I read over it.
I moved the logic there.

There still is the state that if someone creates a nodeset with a single node, that pod will still be blocked from rebooting. But perhaps that is to the user to fix.

pkg/controller/elasticsearch/pdb/reconcile.go
```diff
@@ -116,11 +116,13 @@ func expectedPDB(es esv1beta1.Elasticsearch, statefulSets sset.StatefulSetList,
 }
 
 // buildPDBSpec returns a PDBSpec computed from the current StatefulSets,
-// considering the cluster health and topology.
+// considering the cluster health and topology, when cluster is high available.
+// it sets MinAvailable to 0 when nodecount is 1 (not high available)
+//   when the PDB is 1, and the amount of pods is 1, Kubernetes will not be able to reboot the node.
```
sebgl, line 121: I think I would remove the newly added comment here since the logic is now moved to another function.
The content of `allowedDisruptions()` is already very well documented and self-explanatory. WDYT?

sebgl:
> There still is the state that if someone creates a nodeset with a single node, that pod will still be blocked from rebooting. But perhaps that is to the user to fix.

Yeah I would consider it the user responsibility at this point.

pkg/controller/elasticsearch/pdb/reconcile.go
sanderma: You are totally right, reverted to the previous code.
Apologies for the hasty job.

sebgl: No worries!

sanderma:
> Changes LGTM!
> 
> @sanderma it looks like you need to resolve some conflicts for this PR to be merged.

Ooof, I was running on an old branch as it seems (used v1beta1, not v1) so changed that.

> Also, can you sign the CLA? https://www.elastic.co/contributor-agreement

Signed it (I did in the past btw, but perhaps this needs to be done regularly?)

sebgl:
Jenkins test this please.

sebgl:
cla/check

sebgl:
@sanderma it looks like your commits were signed by different users (I see `sanderma` and `sanderMa`). I guess that's why the cla check is complaining.

An easy way out could be to rebase and squash all your commits into a single one signed by the user you want.

sebgl:
Jenkins test this please

Issue #2670, opened 2020-03-05
Review error messages


anyasabo:
Validation errors and events are some of the main ways we communicate information back to users. We should take a look and make sure they're appropriately helpful. For example, the validation error for resizing PVCs should probably give people an indication of what they can do to resolve it (spin up a new nodeSet, go to the docs, etc.). Some tips from our friendly writing team when reviewing:

![Screen Shot 2020-03-05 at 11 41 13 AM](https://user-images.githubusercontent.com/1638148/76009130-5331b400-5ed6-11ea-96a4-943073bdca7a.png)
![Screen Shot 2020-03-05 at 11 41 25 AM](https://user-images.githubusercontent.com/1638148/76009140-562ca480-5ed6-11ea-848e-ddaaf9c3d3d1.png)

alaudazzi:
Some of the error messages that need to be reviewed:
https://github.com/elastic/cloud-on-k8s/blob/master/pkg/apis/elasticsearch/v1/validations.go#L20

alaudazzi:
@anyasabo 

Back to our meeting on this topic, we might want to refine/complete the error messages as discussed:

cfgInvalidMsg = "Configuration invalid"
=> Can we explain why? How do I get a valid configuration?

masterRequiredMsg = "Elasticsearch needs to have at least one master node"
=> OK

parseVersionErrMsg = "Cannot parse Elasticsearch version"
=> Can we explain why?

parseStoredVersionErrMsg = "Cannot parse current Elasticsearch version"
=> Can we explain why?

invalidSanIPErrMsg = "Invalid SAN IP address"
=> Explain what a valid SAN IP address is.

pvcImmutableMsg = "Volume claim templates cannot be modified"
=> OK

invalidNamesErrMsg = "Elasticsearch configuration would generate resources with invalid names"
=> OK

unsupportedVersionErrMsg = "Unsupported version"
=> Is it worth mentioning what the supported versions are? Might have a maintenance issue.

unsupportedConfigErrMsg  = "Configuration setting is reserved for internal use. User-configured use is unsupported"
=> Probably OK

duplicateNodeSets = "NodeSet names must be unique"
=> OK

noDowngradesMsg = "Downgrades are not supported"
=> OK

unsupportedVersionMsg = "Unsupported version"
=> Is it worth mentioning what the supported versions are? Might have a maintenance issue.

unsupportedUpgradeMsg = "Unsupported version upgrade path"
=> Explain what is supported?

[...]

PR #3168, opened 2020-06-01
Make some error messages more descriptive

alaudazzi:
This PRs improves the text of some error messages in 
`cloud-on-k8s/pkg/apis/elasticsearch/v1/validations.go` 

Relates to #2670.

### DIFF
--- a/pkg/apis/elasticsearch/v1/validations.go
+++ b/pkg/apis/elasticsearch/v1/validations.go
@@ -20,9 +20,9 @@ import (
 const (
 	cfgInvalidMsg            = "Configuration invalid"
 	masterRequiredMsg        = "Elasticsearch needs to have at least one master node"
-	parseVersionErrMsg       = "Cannot parse Elasticsearch version"
-	parseStoredVersionErrMsg = "Cannot parse current Elasticsearch version"
-	invalidSanIPErrMsg       = "Invalid SAN IP address"
+	parseVersionErrMsg       = "Cannot parse Elasticsearch version. String format must be {major}.{minor}.{patch}[-{label}"
+	parseStoredVersionErrMsg = "Cannot parse current Elasticsearch version. String format must be {major}.{minor}.{patch}[-{label}"
+	invalidSanIPErrMsg       = "Invalid SAN IP address. Must be a valid IPv4 address"
 	pvcImmutableMsg          = "Volume claim templates cannot be modified"
 	invalidNamesErrMsg       = "Elasticsearch configuration would generate resources with invalid names"
 	unsupportedVersionErrMsg = "Unsupported version"
@@ -30,7 +30,7 @@ const (
 	duplicateNodeSets        = "NodeSet names must be unique"
 	noDowngradesMsg          = "Downgrades are not supported"
 	unsupportedVersionMsg    = "Unsupported version"
-	unsupportedUpgradeMsg    = "Unsupported version upgrade path"
+	unsupportedUpgradeMsg    = "Unsupported version upgrade path. Check the Elasticsearch documentation."
 )
 
 type validation func(*Elasticsearch) field.ErrorList


anyasabo:
These also need to be updated in `pkg/apis/elasticsearch/v1beta1/validations.go`.

It looks like the validation added for the other resources is already probably okay but is in `pkg/common/v1/validations.go` if you want to take a look. I wonder if we should reword the errors from `NoUnknownFields()` in particular

pkg/apis/elasticsearch/v1/validations.go
```diff
@@ -20,17 +20,17 @@ import (
...
+	invalidSanIPErrMsg       = "Invalid SAN IP address. Must be a valid IPv4 address"
 	pvcImmutableMsg          = "Volume claim templates cannot be modified"
 	invalidNamesErrMsg       = "Elasticsearch configuration would generate resources with invalid names"
 	unsupportedVersionErrMsg = "Unsupported version"
 	unsupportedConfigErrMsg  = "Configuration setting is reserved for internal use. User-configured use is unsupported"
 	duplicateNodeSets        = "NodeSet names must be unique"
 	noDowngradesMsg          = "Downgrades are not supported"
 	unsupportedVersionMsg    = "Unsupported version"
-	unsupportedUpgradeMsg    = "Unsupported version upgrade path"
+	unsupportedUpgradeMsg    = "Unsupported version upgrade path. Check the Elasticsearch documentation."
```
anyasabo, line 33:
```suggestion
	unsupportedUpgradeMsg    = "Unsupported version upgrade path. Check the Elasticsearch documentation for supported upgrade paths."
```

I don't know if this is too obvious/verbose

alaudazzi: Thank you Anya, I guess it's OK to be more explicit.

alaudazzi:
@anyasabo 
Error messages are now updated also in `pkg/apis/elasticsearch/v1beta1/validations.go`.

For your second comment, `pkg/common/v1/validations.go` is probably `pkg/apis/common/v1/validations.go`?

From `NoUnknownFields()`, the error messages I can see are the following ones:
`unknownPrefix := "json: unknown field "`
`"%s field found in the %s annotation is unknown"`

How do you want to handle them?

anyasabo:
>For your second comment, pkg/common/v1/validations.go is probably pkg/apis/common/v1/validations.go?

Yes sorry.

For the `NoUnknownFields()` error message, what that means is that there were extraneous fields that we could not parse into an object. This happens 99% of the time because people's indentation is off, so their config is part of a different object than what it needs to be. So I wonder if we should tell just add something like

>This is often due to incorrect indentation in the manifest.

would make it clearer.


I also wonder if including "in the %s annotation" is useful information to provide. It seems like an implementation detail since the user isn't setting that annotation, kubectl is behind their back. @david-kow what do you think?

david-kow:
I think I'd leave it. While user is not involved in setting it, it gives at least some path to follow if they don't know where the issue coming from. I'd guess they be looking at the resource itself, not the annotation if validation fails. This is especially useful if resource is actually fine, but annotation isn't (validation fails -> `kubectl edit` resource -> validation still fails because annotation is wrong).

pkg/apis/elasticsearch/v1/validations.go
```diff
@@ -20,17 +20,17 @@ import (
 const (
 	cfgInvalidMsg            = "Configuration invalid"
 	masterRequiredMsg        = "Elasticsearch needs to have at least one master node"
-	parseVersionErrMsg       = "Cannot parse Elasticsearch version"
-	parseStoredVersionErrMsg = "Cannot parse current Elasticsearch version"
-	invalidSanIPErrMsg       = "Invalid SAN IP address"
+	parseVersionErrMsg       = "Cannot parse Elasticsearch version. String format must be {major}.{minor}.{patch}[-{label}"
```
david-kow, line 23:
```suggestion
	parseVersionErrMsg       = "Cannot parse Elasticsearch version. String format must be {major}.{minor}.{patch}[-{label}]"
```

pkg/apis/elasticsearch/v1/validations.go
```diff
@@ -20,17 +20,17 @@ import (
...
+	parseStoredVersionErrMsg = "Cannot parse current Elasticsearch version. String format must be {major}.{minor}.{patch}[-{label}"
```
david-kow, line 24:
```suggestion
	parseStoredVersionErrMsg = "Cannot parse current Elasticsearch version. String format must be {major}.{minor}.{patch}[-{label}]"
```

pkg/apis/elasticsearch/v1beta1/validations.go
```diff
@@ -17,19 +17,19 @@ import (
...
-	invalidNamesErrMsg       = "Elasticsearch configuration would generate resources with invalid names"
-	unsupportedVersionErrMsg = "Unsupported version"
-	unsupportedConfigErrMsg  = "Configuration setting is reserved for internal use. User-configured use is unsupported"
-	duplicateNodeSets        = "NodeSet names must be unique"
-	noDowngradesMsg          = "Downgrades are not supported"
-	unsupportedVersionMsg    = "Unsupported version"
-	unsupportedUpgradeMsg    = "Unsupported version upgrade path"
+			cfgInvalidMsg            = "Configuration invalid"
+			masterRequiredMsg        = "Elasticsearch needs to have at least one master node"
+			parseVersionErrMsg       = "Cannot parse Elasticsearch version. String format must be {major}.{minor}.{patch}[-{label}"
```
david-kow, line 22:
```suggestion
			parseVersionErrMsg       = "Cannot parse Elasticsearch version. String format must be {major}.{minor}.{patch}[-{label}]"
```

pkg/apis/elasticsearch/v1beta1/validations.go
```diff
@@ -17,19 +17,19 @@ import (
...
+			parseStoredVersionErrMsg = "Cannot parse current Elasticsearch version. String format must be {major}.{minor}.{patch}[-{label}"
```
david-kow, line 23:
```suggestion
			parseStoredVersionErrMsg = "Cannot parse current Elasticsearch version. String format must be {major}.{minor}.{patch}[-{label}]"
```

alaudazzi:
@david-kow thank you for your review!

alaudazzi:
@anyasabo thank you for your input! I expanded the error message for `NoUnknownFields()` as you suggested.

Issue #1233, opened 2019-07-13
admission webhook "validation.elasticsearch.elastic.co" does not support dry run


donbowman:
## Proposal

**Use case. Why is this important?**

## Bug Report
**What did you do?**

running kubectl diff we get:
`Error from server (BadRequest): admission webhook "validation.elasticsearch.elastic.co" does not support dry run`

It would be nice if this could support dry-run, its part of our workflow.

**What did you expect to see?**

expected to see a diff

**What did you see instead? Under which circumstances?**

error message
`Error from server (BadRequest): admission webhook "validation.elasticsearch.elastic.co" does not support dry run`

**Environment**

* Version information:

	insert version or git commit id here

0.8.1 of operator,

```
kubectl version
Client Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.0", GitCommit:"e8462b5b5dc2584fdcd18e6bcfe9f1e4d970a529", GitTreeState:"clean", BuildDate:"2019-06-19T16:40:16Z", GoVersion:"go1.12.5", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"13+", GitVersion:"v1.13.7-gke.8", GitCommit:"7d3d6f113e933ed1b44b78dff4baf649258415e5", GitTreeState:"clean", BuildDate:"2019-06-19T16:37:16Z", GoVersion:"go1.11.5b4", Compiler:"gc", Platform:"linux/amd64"}
```

* Kubernetes information:

	insert any information about your Kubernetes environment that could help us:
	* On premise ?
	* Cloud: GKE / EKS / AKS ?
	* Kubernetes distribution: Openshift / Rancher / PKS ?
	
	for each of them please give us the version you are using
```
$ kubectl version
```

* Resource definition:
```
if relevant insert the resource definition
```

* Logs:
```
insert operator logs or any relevant message to the issue here
```

jonsabo:
Thanks for bringing this up. I can definitely see how this would be useful, and in theory shouldn't be too much work to enable.

From the docs here:
https://kubernetes.io/blog/2019/01/14/apiserver-dry-run-and-kubectl-diff

looks like we would need to do this
>    Remove any side-effects when the dry-run parameter is specified on the webhook request,
>    Specify in the sideEffects field of the admissionregistration.k8s.io/v1beta1.Webhook object to indicate that the object doesn’t have side-effects on dry-run (or at all).


[...]

PR #3181, opened 2020-06-03
Specify webhook has no side effects

anyasabo:
Fix https://github.com/elastic/cloud-on-k8s/issues/1233

This allows users to dry run resources to see if the API server will accept the request without persisting it to storage. Dry run was [promoted to beta](https://kubernetes.io/blog/2019/01/14/apiserver-dry-run-and-kubectl-diff/) in 1.13 so will not work on older versions without the flag enabled, but does not cause errors applying the webhook in my testing.

### DIFF
--- a/config/operator/all-in-one/webhook.template.yaml
+++ b/config/operator/all-in-one/webhook.template.yaml
@@ -12,6 +12,7 @@ webhooks:
       path: /validate-apm-k8s-elastic-co-v1-apmserver
   failurePolicy: Ignore
   name: elastic-apm-validation-v1.k8s.elastic.co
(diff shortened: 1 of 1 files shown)


anyasabo:
Ah odd, 1.12 on minikube allowed it. I think I'm good to not merge this then until we drop support for OpenShift 3.11.

cmur2:
Can you estimate when support for OpenShift 3.11 will be dropped?

The issue https://github.com/elastic/cloud-on-k8s/issues/1233 is open for almost a year now and dry-run capable webhooks have become a standard so it would be great to enable this possibility by default for all the people running modern clusters.

Issue #3173, opened 2020-06-02
TestEnterpriseTrialExtension is flaky


pebrc:
https://devops-ci.elastic.co/job/cloud-on-k8s-e2e-tests-snapshot-versions/67/consoleFull

```
=== RUN   TestEnterpriseTrialExtension/Elasticsearch_license_should_be_[platinum_enterprise]
Retries (5m0s timeout): ....................................................................................................
    TestEnterpriseTrialExtension/Elasticsearch_license_should_be_[platinum_enterprise]: utils.go:84: 
        	Error Trace:	utils.go:84
        	Error:      	Received unexpected error:
        	            	expectedTypes license type [  platinum enterprise] got basic
        	Test:       	TestEnterpriseTrialExtension/Elasticsearch_license_should_be_[platinum_enterprise]
{"log.level":"error","@timestamp":"2020-06-02T01:39:45.975Z","message":"stopping early","service.version":"0.0.0-00000000","service.type":"eck","ecs.version":"1.4.0","error":"test failure","error.stack_trace":"github.com/go-logr/zapr.(*zapLogger).Error\n\t/go/pkg/mod/github.com/go-logr/zapr@v0.1.0/zapr.go:128\ngithub.com/elastic/cloud-on-k8s/test/e2e/test.StepList.RunSequential\n\t/go/src/github.com/elastic/cloud-on-k8s/test/e2e/test/step.go:43\ngithub.com/elastic/cloud-on-k8s/test/e2e/es.TestEnterpriseTrialExtension\n\t/go/src/github.com/elastic/cloud-on-k8s/test/e2e/es/license_test.go:191\ntesting.tRunner\n\t/usr/local/go/src/testing/testing.go:991"}
    --- FAIL: TestEnterpriseTrialExtension/Elasticsearch_license_should_be_[platinum_enterprise] (300.00s)
```

Looks like trial is started on operator level 

```
{"log.level":"debug","@timestamp":"2020-06-02T01:36:46.503Z","log.logger":"resource","message":"Saving","service.version":"1.2.0-fcc109b9","service.type":"eck","ecs.version":"1.4.0","namespace":"e2e-boyeh-elastic-system","configmap_name":"elastic-licensing","license_info":{"Timestamp":"2020-06-02T01:36:46Z","EckLicenseLevel":"enterprise_trial","TotalManagedMemory":2.147483648,"MaxEnterpriseResourceUnits":100,"EnterpriseResourceUnits":1}}
```
[...]

PR #3183, opened 2020-06-04
Improve trial extension e2e test

pebrc:
Make the two tests that execute trial license functionality mutually exclusive:
  *  on production builds we just run `TestEnterpriseTrialLicense` (would test only ECK-managed trial)
  * on snapshot builds we run `TestEnterpriseTrialExtension` (would test both ECK-managed trial and extension)

Add verification of ECK-managed trial to `TestEnterpriseTrialExtension`

Move the start of of the trial extension validity one hour into the past to make sure the test works even in the presence of clock skew between test runner and operator.

Related #3173


test/e2e/es/license_test.go
```diff
@@ -160,7 +159,6 @@ func TestEnterpriseTrialExtension(t *testing.T) {
 				},
 			},
 			licenseTestContext.DeleteAllEnterpriseLicenseSecrets(),
-			licenseTestContext.CreateEnterpriseTrialLicenseSecret(trialSecretName),
```
sebgl, line 163: isn't the main objective of this test to verify that an existing trial license can be extended with a non-trial short-lived license?
If we remove the trial license don't we remove half the purpose of this test?

pebrc: It was not testing this anyway. ECK-managed trials can only be started once. `TestEnterpriseTrialLicense` which runs just before this test, already starts an ECK-managed trial. So this statement was just a left-over piece of code an if you check the assertions it was never asserted that it had any effect (otherwise the test would fail)

What we could do instead to test that the trial extension is actually extending the trial is make both tests mutually exclusive:
* on production builds we just run the `TestEnterpriseTrialLicense` (would test only ECK-managed trial)
* on snapshot builds we run `TestEnterpriseTrialExtension` (would test both ECK-managed trial and extension)

test/e2e/test/elasticsearch/license_fixture.go
```diff
@@ -58,12 +58,12 @@ func GenerateTestLicense(signer *license.Signer) (client.License, error) {
 	licenseSpec := client.License{
 		UID:                uuid.String(),
 		Type:               "enterprise",
-		IssueDateInMillis:  chrono.ToMillis(time.Now()),
+		IssueDateInMillis:  chrono.ToMillis(time.Now().Add(-1 * time.Hour)),
```
sebgl, line 61: Should we take this a step further and set the issue date to eg. 3 days earlier (which also seems to be a realistic thing we'd do when generating the license)? So we avoid any potential timezone inconsistency.

PR #3195, opened 2020-06-05
Use _cluster/health API during rolling upgrades

### LINKED ISSUE #3070: Shards from _cat/shards may contain stale information
We regularly call `_cat/shards` to retrieve the list of shards in order to:

- ensure a node is safe to remove (no more shards assigned to it)
- ensure there are STARTED replica on other nodes when restarting a node containing primary shards
- ensure there is no initializing or relocating shards when dealing with upgrades in a [version upgrade scenario](https://github.com/elastic/cloud-on-k8s/blob/e35950f1ad2ae5964149c85a5bca00fcf7051c26/pkg/controller/elasticsearch/driver/upgrade_predicates.go#L177-L222)

`_cat/shards` does not provide realtime information and may contain stale information.

We should investigate what can be used instead. 

Example:
```
GET _cluster/health?wait_for_no_initializing_shards&wait_for_events=languid&wait_for_nodes=N
```
seems to be a candidate for waiting for shards to initialize.

To be discussed with the es-distrib and cloud-orchestration teams.

pebrc:
This is to make sure there are no pending or ongoing shard fetches,
initializations or relocations when rolling the next node. Otherwise
the cluster status might go red if we roll forward before the indices
on the previous node have recovered. 

* this applies to upgrades only (i.e. version upgrades) see #3070 for rationale
* adds a parameterized Health API call to ESState cache to be used during predicate evaluation 
* adds a `SafeToRoll` (naming open for discussion, I was thinking also: `NoMovingShards` or `StableShards` ) method on Health results, which checks for the conditions mentioned above

Fixes #3070

### DIFF
--- a/pkg/controller/elasticsearch/client/model.go
+++ b/pkg/controller/elasticsearch/client/model.go
@@ -25,12 +25,6 @@ type Info struct {
 	} `json:"version"`
 }
 
-// Query params for _cluster/health see https://www.elastic.co/guide/en/elasticsearch/reference/current/cluster-health.html
-const (
-	WaitForEvents string = "wait_for_events"
-	Timeout       string = "timeout"
-)
-
 // Health represents the response from _cluster/health
 type Health struct {
 	ClusterName                 string                   `json:"cluster_name"`
@@ -50,12 +44,12 @@ type Health struct {
 	ActiveShardsPercentAsNumber float32                  `json:"active_shards_percent_as_number"`
 }
 
-// SafeToRoll indicates that a rolling update can continue with the next node if
+// IsSafeToRoll indicates that a rolling update can continue with the next node if
 // - no relocating or initializing shards or shards being fetched
 // - all primaries allocated
 // only reliable if Health result was created with wait_for_events=languid
 // so that there are no pending initialisations in the task queue
-func (h Health) SafeToRoll() bool {
+func (h Health) IsSafeToRoll() bool {
 	return !h.TimedOut && // make sure request did not time out (i.e. no pending events)
 		h.Status != esv1.ElasticsearchRedHealth && // all primaries allocated
 		h.NumberOfInFlightFetch == 0 && // no shards being fetched
--- a/pkg/controller/elasticsearch/client/model.go
+++ b/pkg/controller/elasticsearch/client/model.go
@@ -56,7 +56,7 @@ type Health struct {
 // only reliable if Health result was created with wait_for_events=languid
 // so that there are no pending initialisations in the task queue
 func (h Health) SafeToRoll() bool {
-	return h.TimedOut == false && // make sure request did not time out (i.e. no pending events)
+	return !h.TimedOut && // make sure request did not time out (i.e. no pending events)
 		h.Status != esv1.ElasticsearchRedHealth && // all primaries allocated
 		h.NumberOfInFlightFetch == 0 && // no shards being fetched
 		h.InitializingShards == 0 && // no shards initializing
--- a/pkg/controller/elasticsearch/client/model.go
+++ b/pkg/controller/elasticsearch/client/model.go
@@ -25,6 +25,13 @@ type Info struct {
 	} `json:"version"`
 }
 
+// Query params for _cluster/health see https://www.elastic.co/guide/en/elasticsearch/reference/current/cluster-health.html
+const (
+	WaitForNoInitializingShards string = "wait_for_no_initializing_shards"
+	WaitForEvents               string = "wait_for_events"
+	Timeout                     string = "timeout"
+)
+
 // Health represents the response from _cluster/health
 type Health struct {
 	ClusterName                 string                   `json:"cluster_name"`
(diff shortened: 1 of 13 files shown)


pkg/controller/elasticsearch/driver/driver.go
```diff
@@ -201,7 +201,7 @@ func (d *defaultDriver) Reconcile(ctx context.Context) *reconciler.Results {
 				events.EventReasonUnexpected,
 				fmt.Sprintf("Could not update cluster license: %s", err.Error()),
 			)
-			log.Error(err, "Could not update cluster license", "namespace", d.ES.Namespace, "es_name", d.ES.Name)
+			log.Info("Could not update cluster license", "err", err, "namespace", d.ES.Namespace, "es_name", d.ES.Name)
```
pebrc, line 204: Happy to back this out, this change is unrelated, but the license errors were distracting during testing.

pkg/controller/elasticsearch/client/v6.go
```diff
@@ -59,9 +59,15 @@ func (c *clientV6) Flush(ctx context.Context) error {
 	return c.post(ctx, "/_flush", nil, nil)
 }
 
-func (c *clientV6) GetClusterHealth(ctx context.Context) (Health, error) {
+func (c *clientV6) GetClusterHealth(ctx context.Context, params url.Values) (Health, error) {
 	var result Health
-	return result, c.get(ctx, "/_cluster/health", &result)
+	err := c.get(ctx, "/_cluster/health?"+params.Encode(), &result)
+	if IsTimeout(err) {
+		// ignore timeout errors as they are communicated in the returned payload so we can reserve error handling
```
anyasabo, line 66: it's in the payload, but do we ever want to treat a timeout as *not* an error condition? I see some of the other callers we have don't check for timeouts but probably should? example: https://github.com/elastic/cloud-on-k8s/pull/3195/files#diff-972e1e28f90c08fd73595db4b03025d2L73

pkg/controller/elasticsearch/driver/driver.go
anyasabo: 🤷‍♀️ I could go either way on this, it is indeed an error and normally indicates the cluster is fubar. But is misleading to users a lot as you said

pkg/controller/elasticsearch/license/apply.go
```diff
@@ -43,7 +43,7 @@ func applyLinkedLicense(
 	defer cancel()
 	current, err := updater.GetLicense(ctx)
 	if err != nil {
-		return err
+		return fmt.Errorf("while getting current license level %w", err)
```
anyasabo, line 46: nittiest of nits: personal preference for `errors.wrap` here but obviously this is very much personal pref

anyasabo:
LGTM overall, this was easy to read. `SafeToRoll` sounds fine. If you're unsure about it I also think we could go with the `Is...()` naming convention for things that return bools, so maybe `IsSafeToRoll()` or `IsRollingUpgradeSafe()` to be extra verbose

pkg/controller/elasticsearch/client/v6.go
pebrc: I believe those would happen on an idle connection in the pool and are handled by the Go http transport thingy IIUC https://go-review.googlesource.com/c/go/+/179457/4/src/net/http/transport.go#1931

But I like the idea of a separate API, my only concern is maybe that we are mixing levels of abstraction. So far all client APIs have been direct representations of Elasticsearch REST APIs this would be the first 'higher-level' API. My gut feeling is that we should do this in layers: have a low-level client that just exposes ES APIs and then a higher-level client interface on top that offers these more convenient abstractions. But maybe I am over-complicating things. wdyt?

sebgl: You're right about the abstraction level. `ClusterBootstrappedForZen2()` is another example of a function doing a bit more than API calls.

Splitting lower vs. higher level functions attached to different structs may make things a bit cleaner but I'm not 100% sure it would reduce complexity.

Getting the cluster health with additional params is still a low-level API call?
I think I'd be fine with something like these 2 functions if it simplifies our life:

```go
func (c *clientV6) GetClusterHealth(ctx context.Context) (Health, error) {
	var result Health
	return result, c.get(ctx, "/_cluster/health", &result)
}

func (c *clientV6) GetClusterHealthWaitForLanguid(ctx context.Context, timeout time.Time) (Health, error) {
	var result Health
	// wait for all "languid" events (lowest priority) to be processed, or time out
	url := fmt.Sprintf("/_cluster/health?wait_for_events=languid&timeout=%ds", timeout.Second())
	err := c.get(ctx, url, &result)
	if IsTimeout(err) && result.TimedOut{
		// timeout reached, which is not an unexpected error
		// we still have a response body to work with
		err = nil
	}
	return result, err
}
```

pkg/controller/elasticsearch/client/model.go
```diff
@@ -44,6 +44,19 @@ type Health struct {
...
+// - all primaries allocated
+// only reliable if Health result was created with wait_for_events=languid
+// so that there are no pending initialisations in the task queue
+func (h Health) IsSafeToRoll() bool {
+	return !h.TimedOut && // make sure request did not time out (i.e. no pending events)
+		h.Status != esv1.ElasticsearchRedHealth && // all primaries allocated
+		h.NumberOfInFlightFetch == 0 && // no shards being fetched
+		h.InitializingShards == 0 && // no shards initializing
+		h.RelocatingShards == 0 // no shards relocating
+}
```
sebgl, line 58: I think I would not have attached that function to the `Health` struct, and maybe instead move it to a `func IsSafeToRoll(h Health)` close to the rolling upgrade code. It's more a rolling upgrade predicate than something that belongs to the Health domain in my mind.

But that's really not a big deal, I'm fine with keeping the current code.

pkg/controller/elasticsearch/driver/esstate.go
```diff
@@ -115,21 +117,26 @@ func (s *memoizingShardsAllocationEnabled) ShardAllocationsEnabled() (bool, erro
...
 }
 
 // initialize requests Elasticsearch for cluster health, only once.
 func (h *memoizingHealth) initialize() error {
 	ctx, cancel := context.WithTimeout(h.ctx, esclient.DefaultReqTimeout)
 	defer cancel()
-	health, err := h.esClient.GetClusterHealth(ctx)
+
+	// get cluster health but make sure we have no pending shard initializations
+	// by requiring the event queue to be empty
```
sebgl, line 133: > by requiring the event queue to be empty

This is slightly misleading as I would think we return an error here if the event queue is not empty.
Maybe:

```
// get cluster health and status of the event queue (are there pending events?)
```

pkg/controller/elasticsearch/driver/esstate.go
```diff
@@ -140,3 +147,11 @@ func (h *memoizingHealth) Health() (esv1.ElasticsearchHealth, error) {
 	}
 	return h.health, nil
 }
+
+// IsSafeToRoll returns true if shards a not moving and primaries are allocated
```
idanmo, line 151: typo? should be "if shards **are** not moving"

Issue #3190, opened 2020-06-05
Proposal: Beat "presets"


sebgl:
I'd like to discuss here a proposal to change a bit the logic of the [current Beat CRD](https://github.com/elastic/cloud-on-k8s/pull/3041).

## Concerns

I want to address the following concerns:

* if you want to just grab logs from a mounted `/data/logs` directory with filebeat, you end up by default with many additional things and permissions you don't need:
    * a dedicated service account whose token is mounted in the pod
    * a cluster role binding for that service account that can access the k8s API
    * HostNetwork = true
    * a security context to run with the root user (id 0)
    * system volumes mounted in:
        * /var/lib/docker/containers
        * /var/log/containers
        * /var/log/pods
* similarly with metricbeat, if you just want to grab metrics from a prometheus http endpoint, you still end up with:
    * similar RBAC permissions, service account, root user, host network
    * system volumes mounted in:
        * /var/run/docker.sock
        * /proc
        * /sys/fs/cgroup
* some of those defaults can be overridden in the podTemplate, but some of them are harder to override. For example: the [default volumes](https://github.com/elastic/cloud-on-k8s/issues/3187) we create for Filebeat and Metricbeat (eg. `/proc`) if they don't want them.
* with metricbeat we default to a daemonset collecting metrics from all hosts, but we could also have chosen to default to a deployment collecting kube-state-metrics (both are listed in [metricbeat doc](https://www.elastic.co/guide/en/beats/metricbeat/current/running-on-kubernetes.html)), or default to a deployment [collecting prometheus endpoints](https://www.elastic.co/guide/en/beats/metricbeat/current/metricbeat-module-prometheus.html) 
* the story around [config merging](https://github.com/elastic/cloud-on-k8s/issues/3042) is complicated

## Proposal

* keep defaults very light (no mounted volumes, no privilege, no RBAC, no config) for filebeat and metricbeat, similar to what we do for all generic "other beat", treating all beats the same. We only set the ES output config iff `elasticsearchRef` is specified, but don't set any other config setting:

```yaml
apiVersion: beat.k8s.elastic.co/v1beta1
kind: Beat
metadata:
  name: 
spec:
  type: filebeat
  version: 7.6.2
  elasticsearchRef:
    name: elasticsearch-sample
  daemonSet:
    podTemplate:
      # mount whatever volume you need here and add privileges if needed
  config:
    # whatever you need here, we just add the es output to it
```

[...]

PR #3199, opened 2020-06-08
Beat CRD presets

david-kow:
This PR implements presets proposal (https://github.com/elastic/cloud-on-k8s/issues/3190).

### Preset field

Preset is an optional, enum field that allows the operator to setup a Beat with predefined configuration. That configuration consists of:
- Beat config
- deployment or daemonset spec with podTemplate
- appropriate roles and rolebindings

Preset can be used together with other CRD fields:
- if `config` is specified, it replaces config from preset entirely (`output` is still set though)
- if `podTemplate` is specified, the preset podTemplate is merged with it
- if operator has `manage-beat-rbac` flag set to `false` or `spec.ServiceAccountName` is set, the operator won't create service account or bindings

### Role management

Each preset might need different API permissions. As we know these upfront, we can precreate correct ClusterRoles during ECK installation and allow ECK to create bindings to these roles only (verb `bind`) which doesn't require us to expand operator permissions dramatically.

In addition to that, autodiscover permissions might be needed for a Beat. We detect this by looking for `*.autodiscover` config path.

Role management can be disabled:
- on an operator level by setting `manage-beat-rbac` flag to `false`, and
- on resource level by setting `ServiceAccountName` in the `podTemplate`

### PSP

Beat e2e testing is improved by separating the permissions given to the operator and permissions given to Beat Pods.

Beat PSP and ClusterRole allowing to use it are created as a part of e2e test setup. RoleBinding to the right ServiceAccount is done during e2e test run.


david-kow:
We've decided to go with no defaults approach (https://github.com/elastic/cloud-on-k8s/pull/3214).

PR #3204, opened 2020-06-09
Add E2E tests for Enterprise Search

sebgl:
The tests cover:
- the Enterprise Search sample manifest
- configuration upgrades (through config and configRef)
- cross-namespace associations

They perform a few functional AppSearch tests: create an engine, index documents, search those documents.

Workplace Search tests are much harder to perform, so there are none so far:
- we need a trial license (already manipulated once in other tests)
- connecting to an external source (eg. Google Drive) is probably too much work

Both the Healthcheck and AppSearch tests already validated the Enterprise Search instance can correctly communicate with Elasticsearch.

Follow-up issues:
- [Setup a single http helper for all resources](https://github.com/elastic/cloud-on-k8s/issues/3205)


sebgl:
Jenkins test this please

PR #3207, opened 2020-06-09
Skip Beat test if StackVersion is not supported

david-kow:
Skip Beat tests if stack version is not supported.

### DIFF
--- a/test/e2e/beat/config_test.go
+++ b/test/e2e/beat/config_test.go
@@ -9,6 +9,7 @@ import (
 	"testing"
 
 	v1 "github.com/elastic/cloud-on-k8s/pkg/apis/elasticsearch/v1"
+	"github.com/elastic/cloud-on-k8s/pkg/controller/common/version"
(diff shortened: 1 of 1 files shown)


sebgl:
Don't we also need to deal with the version in `TestSamples`? For that reason I took a different approach in https://github.com/elastic/cloud-on-k8s/pull/3204/commits/390dcfb3832ef646eae31f7fa7f93644c7f968bc.

david-kow:
> Don't we also need to deal with the version in `TestSamples`? For that reason I took a different approach in [390dcfb](https://github.com/elastic/cloud-on-k8s/commit/390dcfb3832ef646eae31f7fa7f93644c7f968bc).

I do like your approach better @sebgl, but for `TestSamples` specifically, I'd assume we can always run the version from the samples and we take the raw spec, not the `ElasticStackVersion` from the context.

sebgl:
Oh you're right the samples do contain a fixed version, I missed that!

Issue #3200, opened 2020-06-08
Support keystore settings in Beat CRD


david-kow:
Beat CRD should allow to provide a list of secrets containing secure settings to add to the keystore and use in the Beat config as per [docs](https://www.elastic.co/guide/en/beats/filebeat/current/keystore.html). The operator would create an init container that populates the keystore.

PR #3209, opened 2020-06-09
Add SecureSettings to Beat CRD

david-kow:
This PR adds `secureSettings` field to Beat CRD, resolves https://github.com/elastic/cloud-on-k8s/issues/3200. 

The feature is similar to secure settings in other CRDs, but users have to reference variables in their config (they are not pulled out automatically), see `AGENT_NAME_VAR` in the e2e test.

Differently from other keystore package callers, Beat controller already provides the volume that contains the keystore, so I've added an option to ignore setting that up.


test/e2e/beat/config_test.go
```diff
@@ -100,6 +102,90 @@ heartbeat.monitors:
...
+		return test.StepList{
+			{
+				Name: "Create secure settings secret",
+				Test: func(t *testing.T) {
+					_ = k.Client.Delete(secret)
+					require.NoError(t, k.Client.Create(secret))
+				},
+			},
+			// Keystore should be empty
+			test.CheckKeystoreEntries(
```
barkbay, line 167: This `Step` is called before any Pod is created. There is not keystore entries because there is no Pod. I think this is not what we we want to test ?

pkg/controller/common/keystore/initcontainer.go
```diff
@@ -68,20 +68,26 @@ func initContainer(
...
 
+	volumeMounts := []corev1.VolumeMount{
+		// access secure settings
+		secureSettingsSecret.VolumeMount(),
+	}
+
+	// caller might be already taking care of the right mount and volume
+	if parameters.DataVolumePath != "" {
+		// volume mount to write the keystore in the data volume
+		volumeMounts = append(volumeMounts, DataVolume(volumePrefix, parameters.DataVolumePath).VolumeMount())
```
pebrc, line 79: I am -1 on adding additional conditional logic here that is just used for one controller. Either we always rely on `WithInitContainerDefaults` to copy the data volume mount into the keystone init container, then we should just remove this parameter. Or we do it consistently for all invocations via the parameter. 

Let me know if I am missing something here ofc.

test/e2e/beat/config_test.go
```diff
@@ -100,6 +102,84 @@ heartbeat.monitors:
...
+	}
+
+	fbBuilder := beat.NewBuilder(name, filebeat.Type).
+		WithElasticsearchRef(esBuilder.Ref()).
+		WithSecureSettings(secretName).
+		WithESValidations(
+			beat.HasEventFromBeat(filebeat.Type),
+			beat.HasEventFromPod(testPodBuilder.Pod.Name),
+			beat.HasMessageContaining(testPodBuilder.Logged),
+			beat.HasEvent("agent.name:"+agentName),
```
pebrc, line 134: Clever!

pkg/apis/beat/v1beta1/beat_types.go
```diff
@@ -33,6 +33,11 @@ type BeatSpec struct {
 	// +kubebuilder:validation:Optional
 	Config *commonv1.Config `json:"config,omitempty"`
 
+	// SecureSettings is a list of references to Kubernetes Secrets containing sensitive configuration options for the Beat.
+	// Secrets data can be then referenced in the Beat config using Secret keys or as specified in each SecureSetting Entries.
```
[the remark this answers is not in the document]
david-kow: `Entries` is a field that I was referencing to, I'll reword a bit.

pkg/controller/beat/common/driver.go
```diff
@@ -63,6 +85,7 @@ func Reconcile(
 	results := reconciler.NewResult(params.Context)
 
 	if err := ValidateBeatSpec(params.Beat.Spec); err != nil {
+		k8s.EmitErrorEvent(params.EventRecorder, err, &params.Beat, events.EventReasonValidation, err.Error())
```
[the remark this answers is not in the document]
david-kow: That's a leftover from the time when driver was in `controller/common` and this code couldn't be sure whether the caller validated it properly. I'll fix separately.

pkg/controller/common/keystore/initcontainer.go
david-kow: I've reverted that change. I renamed the data volume to match the volume name in the init container. If that feels off we can follow up separately, as this is what we are already doing in ES and I'd like to avoid expanding scope of this PR.

pkg/apis/beat/v1beta1/beat_types.go
```diff
@@ -37,6 +37,12 @@ type BeatSpec struct {
 	// +kubebuilder:validation:Optional
 	Config *commonv1.Config `json:"config,omitempty"`
 
+	// SecureSettings is a list of references to Kubernetes Secrets containing sensitive configuration options for the Beat.
+	// Secrets data can be then referenced in the Beat config using Secret keys or as specified in `Entries` field of
```
pebrc, line 41:
```suggestion
	// Secrets data can be then referenced in the Beat config using the Secret's keys or as specified in `Entries` field of
```

I don't want sound super nit picky but this sentence still sounds like something is missing in front of Secret?

pkg/controller/beat/common/driver.go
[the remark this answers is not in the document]
pebrc: Can we track it in an issue so that we don't forget?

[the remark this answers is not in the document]
david-kow: Sure, created an [issue](https://github.com/elastic/cloud-on-k8s/issues/3255).

Issue #3203, opened 2020-06-08
Add configRef to Beat CRD


david-kow:
As per the conversation in https://github.com/elastic/cloud-on-k8s/issues/3177, `configRef` gives us the ultimate escape hatch for issues around `key: null` configs.

In addition to that, I really like the [idea](https://github.com/elastic/cloud-on-k8s/issues/3042#issuecomment-640430566) that @pebrc had about prepopulating the Secret with default if it doesn't exist. I think it's worth implementing.

Issue #3042, opened 2020-05-12
Default configuration in Beat CR


david-kow:
There are few ways a config can be provided in Beats. The most basic one is through `inputs`. Some of the Beats (Filebeat, Metricbeat, Heartbeat, Auditbeat) also offer an alternative - autodiscover. There are two ways a config can be provided there - via `autodiscover.providers[*].hints.default_config` and/or `autodiscover.providers[*].templates`. This means that there are at least three ways users can write their config. This raises a question about our defaults - which way do our default settings use and how they get overriden/replaced by user configs.

I think we should go with `autodiscover` config as a default, due to:
1. it has better experience on k8s (you can use hints on pods)
2. it offers templates that allow to conditionally choose a config to use (for namespace/label x use config y), inputs don’t have such capability
3. it has kubernetes metadata processor configured correctly already

There is an additional difficulty - `autodiscover.providers` and `autodiscover.providers[*].templates` are lists which means that merging them is different (lists items are added together to form a longer list instead of being merged position by position).

I think our goals for good defaults should be to be easily understandable, easy to overwrite and good starting experience. 

We have a number of paths we can take:
1. No default - we provide no default whatsoever and user is responsible for providing the config. We only add `output` part based on `elasticsearchRef`.
2. We provide a default, but it's replaced entirely if user provides any config. We only add `output` part based on `elasticsearchRef`. This is the current implementation.
3. We merge with append the default and user config in the same way we merge all the other configs today.
4. We implement a "smarter" merging mechanism. This could be one or combination of:
- replace position by position in lists
- for top level items provided by the user, replace them entirely in default config
- allow user to choose the behavior

I think we should cross out:
- 1 - because it's a bad starting experience
- 3 - because it's just doing wrong things (our default config can't be overriden, so if someone wants to disable hints, they won't be able to)

So being left with 2 and 4:
- 2 - easy to understand and implement, but it does mean that even slight change requires user to provide the entire config and it's different behavior than our current configs. I'd be in favor of starting with this.
- 4 - might be more powerful, but will get tricky to understand and use. Arguably, our default configs won't be providing enough value to justify the work.

david-kow:
To make it easier to make a call, some examples below. As above, ignoring option 3. as won't work for some of the use cases at all.

There are at least two scenarios where this would be used:

1. overriding default config
1. overriding config provided via `configRef`


### Get logs only from Pods in `my-app-ns-1` namespace

[...]

PR #3210, opened 2020-06-10
Add ConfigRef to Beat CRD

david-kow:
This PR adds `configRef` field to Beat CRD, resolves https://github.com/elastic/cloud-on-k8s/issues/3203.

This enables two use cases:

- user is storing Beat config in a Secret and wants to use it
- user wants to deploy default config, but have a Secret created for editing later (this currently only works for metricbeat and filebeat as we have no defaults for others)


pkg/controller/beat/common/config.go
```diff
@@ -58,44 +67,137 @@ func setOutput(cfg *settings.CanonicalConfig, client k8s.Client, associated comm
...
+}
 
-		if err = cfg.MergeWith(userCfg); err != nil {
-			return nil, err
+// getUserConfig extracts the config either from the spec `config` field or from the Secret referenced by spec
+// `configRef` field. If a Secret is referenced, but doesn't exist, it will be created and populated with the
+// default config. From that point on it will be treated the same as a user provided config - it will be watched
+// and changes will be propagated to the Beat config.
+// Returns the results config, bool to indicate whether config was created from default and error.
+func getUserConfig(params DriverParams, defaultConfig *settings.CanonicalConfig) (*settings.CanonicalConfig, bool, error) {
```
barkbay, line 106: nit: I think the name of this function should also reflect that something could be created (`getOrCreateUserConfig` ?)

pkg/apis/beat/v1beta1/beat_types.go
```diff
@@ -33,6 +33,14 @@ type BeatSpec struct {
 	// +kubebuilder:validation:Optional
 	Config *commonv1.Config `json:"config,omitempty"`
 
+	// ConfigRef contains a reference to Kubernetes Secret holding the Beat configuration. At most one
+	// of [`config`, `configRef`] can be specified. If the referenced Secret doesn't exist, but a default
+	// config exists, the operator will create a Secret with that default config. If that Secret is later
+	// modified, the updated configuration will be used by the Beat. Output section of the Beat config will be
+	// populated by the operator if `elasticsearchRef` is specified.
+	// +kubebuilder:validation:Optional
+	ConfigRef *ConfigSource `json:"configRef,omitempty"`
```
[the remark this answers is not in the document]
david-kow: We have an [issue](https://github.com/elastic/cloud-on-k8s/issues/3042) with config merging in general, so I wouldn't introduce it here. 

That's a good suggestion IMO, @sebgl, WDYT?

pkg/controller/beat/common/config.go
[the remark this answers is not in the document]
david-kow: I'm holding with that, because if we give up on presets/defaults (and it's likely that we do) we won't be creating Secrets on our own any more and this code will change.

pkg/controller/beat/common/config.go
```diff
@@ -58,44 +67,137 @@ func setOutput(cfg *settings.CanonicalConfig, client k8s.Client, associated comm
...
+			Namespace: beat.Namespace,
+			Name:      configRef.SecretName,
+			Labels:    common.AddCredentialsLabel(NewLabels(beat)),
+		},
+		Data: map[string][]byte{
+			ConfigFileName: cfgBytes,
+		},
+	}
+
+	if _, err := reconciler.ReconcileSecret(client, expected, &beat); err != nil {
```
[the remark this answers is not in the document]
david-kow: That's an interesting point, but we might wait with handling that scenario until we are sure we support some form of default config.

pkg/apis/beat/v1beta1/beat_types.go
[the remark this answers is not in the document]
sebgl: `configRefs` (plural) in the EnterpriseSearch CRD does make more sense anyway IMO, let's make this change!

Is there a future where we may solve the config merging problem in Beat? In which case we may be happy to be able to specify multiple configs secrets there. Maybe it's OK to also use `configRefs` (array of secrets) in Beat, but reject the resource at validation level if it contains more than one config (similar to what we have to do if it contains config + configRef).

pkg/apis/beat/v1beta1/beat_types.go
```diff
@@ -49,6 +57,37 @@ type BeatSpec struct {
 	Deployment *DeploymentSpec `json:"deployment,omitempty"`
 }
 
+// ConfigSource references configuration settings used as a Beat config.
+type ConfigSource struct {
```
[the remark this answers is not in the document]
david-kow: Good idea. I've put it in the common, changes (use common struct and rename the filed to plural) on Enterprise Search side to be done separately. [Issue](https://github.com/elastic/cloud-on-k8s/issues/3221) for tracking.

pkg/apis/beat/v1beta1/beat_types.go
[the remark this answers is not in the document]
david-kow: I think I'd keep it singular for now. We don't know if we are going to change it for sure and it's super weird UX (like you said).

pkg/controller/beat/common/config.go
```diff
@@ -58,44 +67,137 @@ func setOutput(cfg *settings.CanonicalConfig, client k8s.Client, associated comm
...
+		// can't be sure if the secret exists or not, error out to be safe
+		return nil, false, err
+	}
+
+	if defaultConfig == nil {
+		// we don't have a default config to use, no point in creating a secret
+		return nil, false, fmt.Errorf("secret %s not found and no default config to use", configRef.SecretName)
+	}
+
+	// create the secret on behalf of the user and use default config as a starting point
```
sebgl, line 171: if we move on with no defaults, I guess we can remove that feature?

david-kow: [Yes!](https://github.com/elastic/cloud-on-k8s/pull/3210#discussion_r438687518)

pkg/controller/beat/common/config.go
```diff
@@ -89,34 +96,95 @@ func buildBeatConfig(
...
+			Namespace: namespace,
+		}, secret); err == nil {
+		// user provided an existing secret, just use it
+		return secret, nil
+	} else if !apierrors.IsNotFound(err) {
+		// can't be sure if the secret exists or not, error out to be safe
+		return nil, err
+	}
+
+	return nil, fmt.Errorf("secret %s not found in %s namespace", name, namespace)
```
barkbay, line 179: I would keep the original error, it would keep the code simpler and the original message is quite explicit: `Secret "foo" not found`

pkg/controller/beat/common/config.go
```diff
@@ -89,34 +96,95 @@ func buildBeatConfig(
...
-	log.V(1).Info(
-		"Replacing ECK-managed configuration by user-provided configuration",
-		"beat_name", beat.Name,
-		"namespace", beat.Namespace,
-	)
 
-	return cfg.Render()
+	data, exists := secret.Data[ConfigFileName]
+	if !exists {
+		msg := fmt.Sprintf("no %s key in secret", ConfigFileName)
```
barkbay, line 143:
```suggestion
		msg := fmt.Sprintf("no key %s in secret %s/%s", ConfigFileName, secret.Namespace, secret.Name)
```

pkg/controller/beat/common/config.go
```diff
@@ -89,34 +96,95 @@ func buildBeatConfig(
...
+		params.Logger.Error(err, msg, "namespace", params.Beat.Namespace, "beat_name", params.Beat.Name, "secret_name", secret.Name)
+
+		msg = fmt.Sprintf("no key %s in secret %s in namespace %s", ConfigFileName, secret.Name, params.Beat.Namespace)
```
barkbay, line 146:
```suggestion

```

test/e2e/test/beat/builder.go
```diff
@@ -31,9 +31,9 @@ const (
 
 // Builder to create a Beat
 type Builder struct {
-	Beat        beatv1beta1.Beat
-	Validations []ValidationFunc
-	RBACObjects []runtime.Object
+	Beat              beatv1beta1.Beat
+	Validations       []ValidationFunc
+	AdditionalObjects []runtime.Object
```
barkbay, line 36: Maybe I'm missing something but I can't find where these `AdditionalObjects` are created/deleted.

david-kow: Through `append` in `WithObjects` and `bind` funcs, is that what you were asking?

barkbay: I'm wondering where the k8s client is called for these objects. I'm not able to run the e2e tests locally:

```
 {"service.version": "1.1.0-SNAPSHOT-mmo", "controller": "beat-controller", "request": "e2e-mercury/test-beat-configref-g6t2", "error": "secret fb-config not found in e2e-mercury namespace", "errorCauses": [{"error": "secret fb-config not found in e2e-mercury namespace"}]}
```

david-kow: We should return them in `RuntimeObjects` func and they should be created/deleted in the cluster with `CreationTestSteps` and `DeletionTestSteps`. I'll try running locally to see if I can repro.

pkg/controller/beat/common/config.go
```diff
@@ -89,34 +95,81 @@ func buildBeatConfig(
...
+
+// getUserConfig extracts the config either from the spec `config` field or from the Secret referenced by spec
+// `configRef` field.
+func getUserConfig(params DriverParams) (*settings.CanonicalConfig, error) {
+	if params.Beat.Spec.Config != nil {
+		userCfg, err := settings.NewCanonicalConfigFrom(params.Beat.Spec.Config.Data)
+		if err != nil {
+			return nil, err
+		}
+		return userCfg, nil
```
sebgl, line 123: I think we can `return settings.NewCanonicalConfigFrom(params.Beat.Spec.Config.Data)` directly.

pkg/controller/beat/common/config.go
```diff
@@ -89,34 +95,81 @@ func buildBeatConfig(
...
+	}
+
+	if params.Beat.Spec.ConfigRef == nil {
```
sebgl, line 126: maybe also `|| params.Beat.Spec.ConfigRef.SecretName == ""`

pkg/apis/beat/v1beta1/beat_types.go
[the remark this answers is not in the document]
sebgl: Sounds good, I think it's not too bad even if we eventually introduce `configRefs`, which, as you said, could never happen.

Issue #3192, opened 2020-06-05
Auto-create dashboards in the default metricbeat/filebeat config


sebgl:
Depends on https://github.com/elastic/cloud-on-k8s/issues/3191.

If the users specified a `kibanaRef` in metricbeat & filebeat spec, I think we should consider setting `setup.dashboards.enabled: true`, which gives the best quickstart experience.

david-kow:
This was implemented in #3211, hence closing.

Issue #3197, opened 2020-06-08
Beat readiness probe


sebgl:
We probably want to introduce a readiness probe for Beats.
It's a bit surprising right now to see filebeat "ready" while Elasticsearch is unavailable.

It looks like we could execute a `filebeat test output` command. To investigate.

david-kow:
What `ready` should indicate though? If Beat can start getting logs/metrics in, I'd consider it ready even if the output is not ready itself. I'd think that's what outputs (ES for instance) `ready` is for.

anyasabo:
For filebeat that is at least what the helm chart uses:
https://github.com/elastic/helm-charts/blob/master/filebeat/values.yaml#L72

anyasabo:
https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#when-should-you-use-a-readiness-probe

>If you'd like to start sending traffic to a Pod only when a probe succeeds, specify a readiness probe. In this case, the readiness probe might be the same as the liveness probe, but the existence of the readiness probe in the spec means that the Pod will start without receiving any traffic and only start receiving traffic after the probe starts succeeding. If your Container needs to work on loading large data, configuration files, or migrations during startup, specify a readiness probe.

>If you want your Container to be able to take itself down for maintenance, you can specify a readiness probe that checks an endpoint specific to readiness that is different from the liveness probe.

The main reason I can think we would want to define a readiness probe is if you were using beats to monitor your other beats. In that case I think you would want to know if the beat was up but the output was down (and so it should be ready even if the output is down).

"Is the output responding" seems more of a question of health in the beats status. I'm not sure there's a good way for ECK to retrieve that though. We currently define beat health as:

```golang
const (
	// BeatRedHealth means that the health is neither yellow nor green.
	BeatRedHealth BeatHealth = "red"

	// BeatYellowHealth means that:
	// 1) at least one Pod is Ready, and
	// 2) association is not configured, or configured and established
	BeatYellowHealth BeatHealth = "yellow"

	// BeatGreenHealth means that:
	// 1) all Pods are Ready, and
	// 2) association is not configured, or configured and established
	BeatGreenHealth BeatHealth = "green"
)
```

david-kow:
> In that case I think you would want to know if the beat was up but the output was down (and so it should be ready even if the output is down).

I'm not sure I'm getting what do you mean here. If we have:
```
[...]

PR #3211, opened 2020-06-10
Add KibanaRef to Beats and support setup.kibana

### LINKED ISSUE #3191: Allow Kibana-Beat association for dashboard setup
`setup.dashboards.enabled: true` auto-creates Kibana dashboards in filebeat/metricbeat. This only works if [connection details to Kibana](https://www.elastic.co/guide/en/beats/filebeat/current/setup-kibana-endpoint.html) are specified.

pebrc:
Adds a `kibanaRef` to the Beats CRD
* wires up the CAs mostly fashioned after the way we do it for APM
* creates a separate user for the Kibana ref for cases where Beats are not directly connected to Elasticsearch
* uses `kibana_admin` role to talk to Kibana

Adds support for `setup.kibana` (sorry for folding this into the same PR but testing was easier this way)
* only Metricbeat, Filebeat and Auditbeat support the feature afaik, so I added the config only to Metricbeat and Filebeat
* I did not introduce a validation that prevents the usage of `kibanaRef` for generic Beats (needs documentation how to use the credentials)

TODO:
   *  e2e test would be nice, but maybe can be done in a follow up to keep the scope small for this PR

Fixes #3191 and #3192


pkg/controller/beat/common/config.go
```diff
@@ -13,66 +13,92 @@ import (
...
-			"output.elasticsearch": map[string]interface{}{
-				"hosts":    []string{associated.AssociationConf().GetURL()},
-				"username": username,
-				"password": password,
-			},
-		})
-
-	if err := cfg.MergeWith(esOutput); err != nil {
-		return err
+	esOutput := map[string]interface{}{
```
anyasabo, line 41: Doesn't need to be resolved in your change, but this made me realize we need to add documentation on how beats and ssl verification interacts with custom certs. Opened this issue:
https://github.com/elastic/cloud-on-k8s/issues/3216

The only way I can think of being smarter here is if we check that the current associated certificate is valid for the host name the beat will use, and if not, setting ssl verification mode to `none`. I'm not sure we want to do that silently though, and probably want the users to know they are disabling ssl.

I do wonder if we want to fail the association if we know that the beat cannot actually associate successfully with the current certificate.

pkg/controller/beat/filebeat/config.go
```diff
@@ -24,3 +28,14 @@ processors:
 - add_host_metadata: {}
 `))
 )
+
+func (d *Driver) configParams() (common.DefaultConfigs, error) {
+	kibanaConfig, err := common.BuildKibanaConfig(d.Client, v1beta1.BeatKibanaAssociation{Beat: &d.Beat})
```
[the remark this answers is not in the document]
anyasabo: The heartbeat error you encountered seems like a bug that should be fixed in the reference yaml, correct?

anyasabo:
LGTM on first pass, will take another look tomorrow for 👍 unless someone gets to it first. I'm good with waiting for the e2e test until a future PR, but can you open an issue so we don't forget?

pkg/controller/beat/common/config.go
```diff
@@ -13,66 +13,92 @@ import (
...
-			return err
-		}
+		esOutput["output.elasticsearch.ssl.certificate_authorities"] = []string{path.Join(certificatesDir(&associated), CAFileName)}
 	}
 
-	return nil
+	return settings.NewCanonicalConfigFrom(esOutput)
+}
+
+// BuildKibanaConfig builds on optional Kibana configuration for dashboard setup and visualizations.
```
anyasabo, line 56:
```suggestion
// BuildKibanaConfig builds an optional Kibana configuration for dashboard setup and visualizations.
```

pkg/controller/beat/common/config.go
```diff
@@ -13,66 +13,92 @@ import (
...
+	err = cfg.MergeWith(outputCfg, defaultConfig.Managed)
+	if err != nil {
 		return nil, err
 	}
-
 	// use only the default config or only the provided config - no overriding, no merging
 	userConfig := beat.Spec.Config
 	if userConfig == nil {
-		if err := cfg.MergeWith(defaultConfig); err != nil {
+		if err := cfg.MergeWith(defaultConfig.Unmanaged); err != nil {
```
anyasabo, line 101: nit: it's not in your change but I just realized that in the debug log below we probably want to include the beat name/ns

pkg/controller/association/controller/beat_kibana.go
```diff
@@ -0,0 +1,60 @@
...
+			return map[string]string{
+				BeatAssociationLabelName:      associated.Name,
+				BeatAssociationLabelNamespace: associated.Namespace,
+				BeatAssociationLabelType:      commonv1.KibanaAssociationType,
+			}
+		},
+		UserSecretSuffix:  "beat-kb-user",
+		CASecretLabelName: kibana.KibanaNameLabelName,
+		ESUserRole: func(commonv1.Associated) (string, error) {
+			return user.KibanaAdminBuiltinRole, nil
```
david-kow, line 39: Will this work with <7.5 version of Kibana? I don't see it in the [docs](https://www.elastic.co/guide/en/elasticsearch/reference/7.4/built-in-roles.html).

pkg/controller/beat/common/config.go
```diff
@@ -13,66 +13,92 @@ import (
...
+	}
+
+	kibanaCfg := map[string]interface{}{
+		"setup.dashboards.enabled": true,
+		"setup.kibana": map[string]interface{}{
+			"host":     associated.AssociationConf().GetURL(),
+			"username": username,
+			"password": password,
+		},
+	}
```
david-kow, line 74: Should we consider adding`setup.dashboards.retry.enabled` ([docs](https://www.elastic.co/guide/en/beats/filebeat/7.0/configuration-dashboards.html#_setup_dashboards_retry_enabled)) to avoid crashing?

pkg/controller/beat/filebeat/config.go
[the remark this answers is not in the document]
pebrc: I am not sure. I read it as "you asked me to import dashboards but there are no dashboards to import so I give up"

pkg/controller/beat/common/config.go
pebrc: Not sure: it would hide this unpleasant startup behaviour but it could hide an actual misconfiguration by never failing. Having said that given the low priority of the auto-install dashboards feature I am happy to include it.

pkg/controller/association/controller/beat_kibana.go
pebrc: Good point. I forgot about that.

pkg/controller/beat/common/config.go
david-kow: That's true, but with this change it will be a common behavior between ES and Kibana - if someone configures output incorrectly Beat won't fail either. We can try to come up with a common solution as a part of https://github.com/elastic/cloud-on-k8s/issues/3197.

pebrc: Turns out there is a bug in libbeat that prevents the retry from actually working https://github.com/elastic/beats/issues/18217 because a Kibana Client is created before the retry logic starts and does a first request that fails.

https://github.com/elastic/beats/blob/0c15394a85134e1797d30592cd08789c4c1a8e1d/libbeat/cmd/instance/beat.go#L782

PR #3212, opened 2020-06-10
Deploy metricbeat on e2e clusters

barkbay:
This PR deploys `metricbeat` on e2e clusters.

A few notes for the reviewers:

* Creation of the `Secret` has been moved at the beginning of the process since it is needed for both filebeats and metricbeats
* I'm not grabbing metrics from `kube-state-metrics`, I'm not sure there are some interesting metrics for our case vs the complexity of implementing the [compatibility matrix](https://github.com/kubernetes/kube-state-metrics#compatibility-matrix).

I'm still working on some relevant visualizations.

### DIFF
--- a/config/e2e/metricbeat.yaml
+++ b/config/e2e/metricbeat.yaml
@@ -1,0 +1,294 @@
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: metricbeat
+  namespace: {{ .E2ENamespace }}
+  labels:
+    k8s-app: metricbeat
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: metricbeat-daemonset-config
+  namespace: {{ .E2ENamespace }}
+  labels:
+    k8s-app: metricbeat
+data:
+  metricbeat.yml: |-
+    metricbeat.config.modules:
+      # Mounted `metricbeat-daemonset-modules` configmap:
+      path: ${path.config}/modules.d/*.yml
+      # Reload module configs as they change:
+      reload.enabled: false
+
+    metricbeat.autodiscover:
+      providers:
+        - type: kubernetes
+          node: ${NODE_NAME}
+          hints.enabled: true
+    processors:
+      - add_cloud_metadata:
+      - add_fields:
+          target: ''
+          fields:
+            pipeline: {{ .Pipeline }}
+            build_number: {{ .BuildNumber }}
+            provider: {{ .Provider }}
+            clusterName: {{ .ClusterName }}
+            kubernetes_version: {{ .KubernetesVersion }}
+            stack_version: {{ .ElasticStackVersion }}
+            e2e_test_id: {{ .Pipeline }}-{{ .BuildNumber }}-{{ .Provider }}-{{ .ClusterName }}-{{ .KubernetesVersion }}-{{ .ElasticStackVersion }}
+
+    setup.template.overwrite: true
+    setup.template.append_fields:
+    - name: stack_version
+      type: keyword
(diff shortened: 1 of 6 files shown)


config/e2e/metricbeat.yaml
```diff
@@ -0,0 +1,294 @@
...
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: metricbeat-daemonset-modules
+  namespace: {{ .E2ENamespace }}
+  labels:
+    k8s-app: metricbeat
+data:
+  kubernetes.yml: |-
```
anyasabo, line 75: This doesn't need to run as a dset since it's just querying the k8s api, but we can probably leave that for later?

That said, I wonder if we also want to scrape the prom metrics of the operator, I think we might. But that doesn't necessarily need to be in this PR. We can move these to a deployment in the future and also add the prometheus metrics at that point

anyasabo: Ah I see later on we are enabling metrics in the operator. I think we also need to add the prometheus module to metricbeat to actually scrape it

pebrc: I _think_ that's happening auto-magically through he auto-discover hints on the operator resource.

config/e2e/metricbeat.yaml
```diff
@@ -0,0 +1,294 @@
...
+            path: /proc
+        - name: cgroup
+          hostPath:
+            path: /sys/fs/cgroup
+        - name: dockersock
+          hostPath:
+            path: /var/run/docker.sock
+        - name: config
+          configMap:
+            defaultMode: 0600
```
pebrc, line 206: I think this does not work unless you run the container as root

barkbay: > I think this does not work unless you run the container as root, otherwise the metricbeat user won't be able to read the config files

Good catch, it is working in our e2e environments because the psp is overriding the default.

config/e2e/metricbeat.yaml
barkbay: > This doesn't need to run as a dset since it's just querying the k8s api, but we can probably leave that for later?

I'm not sure, IIRC autodiscover is per node, [this is also how we deploy it in the recipes](https://github.com/elastic/cloud-on-k8s/blob/master/config/recipes/beats/3_metricbeat-kubernetes.yaml#L18).
I agree it is odd for someone who has been used to prometheus.

config/e2e/metricbeat.yaml
```diff
@@ -0,0 +1,295 @@
...
+            name: metricbeat-daemonset-config
+        - name: modules
+          configMap:
+            defaultMode: 0600
+            name: metricbeat-daemonset-modules
+        - name: data
```
pebrc, line 213: This is something that's already wrong in our Metricbeat recipe: the data volume is not used in any mount afaik and if it were I believe the permissions would be incorrect for a non-root user.

barkbay: I'll fix this and raise a PR against the recipe.

config/e2e/metricbeat.yaml
anyasabo: [It can be either](https://www.elastic.co/guide/en/beats/filebeat/current/configuration-autodiscover.html), though it defaults to node local. This drove me batty before I discovered it was the default:

> scope
>    (Optional) Specify at what level autodiscover needs to be done at. scope can either take node or cluster as values. node scope allows discovery of resources in the specified node. cluster scope allows cluster wide discovery. Only pod and node resources can be discovered at node scope.

PR #3214, opened 2020-06-10
Remove config defaults and Beat RBAC management

david-kow:
This PR removes role management and defaulting configs. The only things we still default are on the podTemplate:

- data and config volumes
- default resources
- resource labelling
- docker image
- default args 

For example, previous default Filebeat deployment would be now:

```yaml
kind: Beat
metadata:
  name: filebeat-sample
spec:
  type: filebeat
  version: 7.7.0
  elasticsearchRef:
    name: elasticsearch-sample
  config:
    filebeat:
      autodiscover:
        providers:
        - hints:
            default_config:
              paths:
              - /var/log/containers/*${data.kubernetes.container.id}.log
              type: container
            enabled: true
          node: ${HOSTNAME}
          type: kubernetes
    processors:
    - add_cloud_metadata: {}
    - add_host_metadata: {}
  daemonSet:
    podTemplate:
      spec:
        containers:
        - name: filebeat
          volumeMounts:
          - mountPath: /var/lib/docker/containers
            name: varlibdockercontainers
          - mountPath: /var/log/containers
            name: varlogcontainers
          - mountPath: /var/log/pods
            name: varlogpods
        hostNetwork: true
        securityContext:
          runAsUser: 0
        serviceAccount: elastic-beat-filebeat-sample
        terminationGracePeriodSeconds: 30
        volumes:
        - hostPath:
            path: /var/lib/docker/containers
          name: varlibdockercontainers
        - hostPath:
            path: /var/log/containers
          name: varlogcontainers
        - hostPath:
            path: /var/log/pods
          name: varlogpods
```

In addition to that, setting up RBAC is needed. For full Filebeat and Metricbeat examples, see [gist](https://gist.github.com/david-kow/c80542317050ad41df55bfcbb9cc1969).

This PR turns off Beat E2E tests as this requires more changes and the removal is not final yet.


anyasabo:
We want to hold this until #3211 is merged, correct?

david-kow:
@anyasabo correct. I'll rebase when #3211 is in.

david-kow:
jenkins test this please

pkg/controller/beat/common/config.go
```diff
@@ -23,8 +23,6 @@ import (
 type DefaultConfig struct {
 	// Managed config is handled by ECK and will not be nullified by user provided config.
 	Managed *settings.CanonicalConfig
-	// Unmanaged config is default config that will be overridden by any user provided config.
-	Unmanaged *settings.CanonicalConfig
 }
```
sebgl, line 26: Should we remove that struct? It feels a bit confusing now.

pkg/controller/beat/common/config.go
```diff
@@ -95,27 +93,25 @@ func buildBeatConfig(
...
+	}
+
+	if err = cfg.MergeWith(userCfg); err != nil {
+		return nil, err
 	}
+	log.V(1).Info(
+		"Replacing ECK-managed configuration by user-provided configuration",
+		"beat_name", beat.Name,
+		"namespace", beat.Namespace,
+	)
```
sebgl, line 114: We should remove that log? Since now we just append/merge similar to all CRDs.

config/e2e/beat-roles.yaml
```diff
@@ -0,0 +1,77 @@
...
+  name: elastic-beat-restricted
+rules:
+- apiGroups:
+  - policy
+  resources:
+  - podsecuritypolicies
+  resourceNames:
+  - elastic.beat.restricted
+  verbs:
+  - use
```
sebgl, line 77: Should we include all the rbac stuff in `config/samples/beat`?

david-kow: I'm not sure. I didn't do that since we don't do it for other resources - no PSP, roles, bindings etc. I was thinking to have full examples under `config/recipes`, but we can solve that when doing docs work.

Issue #3219, opened 2020-06-11
Use less privileged user for Beats


david-kow:
Beat [docs](https://www.elastic.co/guide/en/beats/filebeat/current/feature-roles.html) describe the roles and privileges that are required for Beats to work.

They divide it into four separate roles:
- setup - create indices, dashboards, index-patterns
- monitoring - stack monitoring
- writer - push Beat data
- reader - use Data

We need privileges of the first three roles, but I think we'd want to use them at the same time. That is, we would not give up setup role permissions even after we are done with the setup. Orchestrating to switch roles at the right time would be quite complicated (multiple associations, Beat Pods restarts with different roles and adjusted config to skip setup, detecting used modules to know when to enable/disable stack monitoring related permissions, likely more). 

I think having a sum of permissions for those three roles this is an acceptable tradeoff of limiting privileges vs complexity.

Each Beat requires permissions for their respective index, `beatname-*`. This means we need to at least precreate right roles for all official Beats we support. The question is what we want to do with other Beats - dynamically create roles? Leave it creation up to the user, but at least try to use the right role name? This has some security concerns as we would base what roles we want to use of user provided string (`spec.Type`).

PR #3220, opened 2020-06-11
Use a user with minimal set of permissions for Beat

david-kow:
Resolves https://github.com/elastic/cloud-on-k8s/issues/3219.

With this PR we will get automatic user and role creation for all official Beats. Used user has minimal set of permissions that is needed for Beats setup, monitoring and publishing.

We won't get automatic role creation for other beats, so users will need to create a role with a well known format beforehand for it to work.

### DIFF
--- a/pkg/controller/association/controller/beat_es.go
+++ b/pkg/controller/association/controller/beat_es.go
@@ -5,6 +5,10 @@
 package controller
 
 import (
+	"fmt"
+	"strings"
+
+	pkgerrors "github.com/pkg/errors"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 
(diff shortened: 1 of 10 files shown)


pkg/controller/association/controller/beat_es.go
```diff
@@ -45,8 +50,65 @@ func AddBeatES(mgr manager.Manager, accessReviewer rbac.AccessReviewer, params o
...
+		}, ","), nil
+	default:
+		return strings.Join([]string{
+			"kibana_user",
+			"ingest_admin",
+			"beats_admin",
+			"monitoring_user",
+			esuser.BeatRoleName(esuser.V70, beat.Spec.Type),
+		}, ","), nil
+	}
```
pebrc, line 113: I noticed these roles accumulate if you upgrade from one ES version to another, not something this PR introduced I think but maybe worth following up on.

pkg/controller/association/controller/beat_es.go
```diff
@@ -45,8 +50,65 @@ func AddBeatES(mgr manager.Manager, accessReviewer rbac.AccessReviewer, params o
...
+	if !ok {
+		return "", pkgerrors.Errorf(
+			"Beat expected, got %s/%s",
+			assoc.GetObjectKind().GroupVersionKind().Group,
+			assoc.GetObjectKind().GroupVersionKind().Kind,
+		)
+	}
+
+	if strings.Contains(beat.Spec.Type, ",") {
+		return "", fmt.Errorf("beat type %s should not contain a comma", beat.Spec.Type)
```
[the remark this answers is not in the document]
david-kow: Defense in depth I guess :) I'll add it to validations.

pkg/controller/association/controller/beat_es.go
david-kow: Mind creating an issue?

pebrc: Hm I realise now that this might be unavoidable given that we do not know in the ES controller which Beat versions are connecting to a particular ES cluster.

pebrc: https://github.com/elastic/cloud-on-k8s/issues/3254

PR #3222, opened 2020-06-11
Update Kubernetes version in API docs

charith-elastic:
Switching to the latest available Kubernetes API version as some links seem to be broken.


charith-elastic:
Yes, a check would certainly be helpful. However it's probably not very nice to hit the Kubernetes website on every CI build. Maybe we could make it a part of the nightly e2e job?

PR #3233, opened 2020-06-12
Name transport service port

charith-elastic:
Names the port in the transport service so that it passes Istio checks.

### DIFF
--- a/pkg/controller/elasticsearch/services/services.go
+++ b/pkg/controller/elasticsearch/services/services.go
@@ -50,6 +50,7 @@ func NewTransportService(es esv1.Elasticsearch) *corev1.Service {
 	labels := label.NewLabels(nsn)
 	ports := []corev1.ServicePort{
 		{
+			Name:     "tls-transport",
 			Protocol: corev1.ProtocolTCP,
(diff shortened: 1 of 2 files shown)


pkg/controller/elasticsearch/services/services.go
```diff
@@ -50,6 +50,7 @@ func NewTransportService(es esv1.Elasticsearch) *corev1.Service {
 	labels := label.NewLabels(nsn)
 	ports := []corev1.ServicePort{
 		{
+			Name:     "tls-transport",
```
anyasabo, line 53: A note for any other reviewers: istio [requires](https://istio.io/latest/docs/ops/configuration/traffic-management/protocol-selection/#manual-protocol-selection) it be prefixed with the protocol

thbkrkr, line 53: This may deserve a little comment?
```suggestion
			Name:     "tls-transport", // name the service with the protocol for Istio
```

anyasabo:
Added to `>enhancements` since it is user facing and should help our compatibility story

PR #3235, opened 2020-06-12
Fix race condition when getting trial-status secret

### LINKED ISSUE #3234: Race condition when retrieving license public key
By working on #3115, I discovered that there is a race condition that can fail the license unit tests.

It is reproducible in local:
```sh
> go test ./pkg/license ./pkg/controller/common/license  -race 
--- FAIL: Test_Start (12.01s)
    reporter_test.go:266: 
                Error Trace:    reporter_test.go:266
                Error:          Condition never satisfied
                Test:           Test_Start
FAIL
FAIL    github.com/elastic/cloud-on-k8s/pkg/license     12.144s
--- FAIL: TestChecker_EnterpriseFeaturesEnabled (0.55s)
    --- FAIL: TestChecker_EnterpriseFeaturesEnabled/valid_trial:_OK (0.00s)
        check_test.go:131: Checker.EnterpriseFeaturesEnabled() = false, want true
    --- FAIL: TestChecker_EnterpriseFeaturesEnabled/valid_legacy_trial:_OK (0.00s)
        check_test.go:131: Checker.EnterpriseFeaturesEnabled() = false, want true
--- FAIL: Test_CurrentEnterpriseLicense (0.48s)
    --- FAIL: Test_CurrentEnterpriseLicense/get_valid_trial_enterprise_license:_OK (0.00s)
        check_test.go:234: Checker.CurrentEnterpriseLicense() = <nil>, want true
FAIL
FAIL    github.com/elastic/cloud-on-k8s/pkg/controller/common/license   4.112s
FAIL
```

The issue is in `publicKeyFor` where we return the data of the secret and we get the secret at the same time:
https://github.com/elastic/cloud-on-k8s/blob/e5b63347c25117e27b2a43100ec6a2e7f6e7562a/pkg/controller/common/license/check.go#L43-L47

It is not clear to me why it always fails locally but so far it has worked in CI.

When the tests fail, `publicKeyFor` returns a `nil` secret which is then resolved to "ECK is using a basic license". This could explain the [flakiness of TestEnterpriseTrialExtension](https://github.com/elastic/cloud-on-k8s/issues/3173).

thbkrkr:
Be careful to read the secret data only after retrieving the secret.

Resolves #3234.

### DIFF
--- a/pkg/controller/common/license/check.go
+++ b/pkg/controller/common/license/check.go
@@ -40,11 +40,14 @@ func (lc *checker) publicKeyFor(l EnterpriseLicense) ([]byte, error) {
 	if !l.IsECKManagedTrial() {
 		return lc.publicKey, nil
 	}
+
 	var signatureSec corev1.Secret
-	return signatureSec.Data[TrialPubkeyKey], lc.k8sClient.Get(types.NamespacedName{
+	err := lc.k8sClient.Get(types.NamespacedName{
 		Namespace: lc.operatorNamespace,
 		Name:      TrialStatusSecretKey,
 	}, &signatureSec)
+	// read secret data only after retrieving the secret
+	return signatureSec.Data[TrialPubkeyKey], err
 }
 
 // CurrentEnterpriseLicense returns the currently valid Enterprise license if installed.


anyasabo:
I am confused why the original code ever worked as it seems like it would always return nothing for `signatureSec.Data[TrialPubkeyKey]`, but your change LGTM. It's not clear to me that this should be labeled `>test` or `>bug` though.

sebgl:
Hm I'm not exactly sure this PR does something? (but 👍 since it makes the code a bit easier to parse)

See this go playground: https://play.golang.org/p/bK-htX2r0VN

pebrc:
Related https://github.com/elastic/cloud-on-k8s/issues/3239

PR #3262, opened 2020-06-16
Minor edits on HTTP settings and TLS SANs

alaudazzi:
Fixed a typo and slightly reworded the note.

### DIFF
--- a/docs/orchestrating-elastic-stack-applications/elasticsearch/http-settings-tls-sans.asciidoc
+++ b/docs/orchestrating-elastic-stack-applications/elasticsearch/http-settings-tls-sans.asciidoc
@@ -21,7 +21,7 @@ spec:
 
 Check the https://kubernetes.io/docs/concepts/services-networking/service/#publishing-services-service-types[Kubernetes Publishing Services (ServiceTypes)] that are currently available.
 
-NOTE: When you change the `clusterIP` setting of the service, ECK will delete and re-create the service, as `clusterIP` is an immutable field. Depending on your client implementation, this might have an impact on the ability to resolve the service's DNS name to endpoints.
+NOTE: When you change the `clusterIP` setting of the service, ECK will delete and re-create the service, as `clusterIP` is an immutable field. Depending on your client implementation, this might have a temporary impact on the ability to resolve the service's DNS name to endpoints.
 
 You can add an IP or a DNS name in the SAN of the self-signed certificate configured by default to secure the HTTP layer with TLS in the `spec.http.tls.selfSignedCertificate` section.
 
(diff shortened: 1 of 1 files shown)


docs/orchestrating-elastic-stack-applications/elasticsearch/http-settings-tls-sans.asciidoc
```diff
@@ -21,7 +21,7 @@ spec:
 
 Check the https://kubernetes.io/docs/concepts/services-networking/service/#publishing-services-service-types[Kubernetes Publishing Services (ServiceTypes)] that are currently available.
 
-NOTE: Please note that when you change the `clusterIP` setting of the service, ECK will delete and re-create the service as `clusterIP` is an immutable field. Depending on your client implementation this might have an a short impact on the ability to resolve the service's DNS name to endpoints.
+NOTE: When you change the `clusterIP` setting of the service, ECK will delete and re-create the service, as `clusterIP` is an immutable field. Depending on your client implementation, this might have an impact on the ability to resolve the service's DNS name to endpoints.
```
anyasabo, line 24: I think we do want the "short" qualifier or something else there to indicate that it is a temporary condition, right? Otherwise I could read this now and think that it might permanently break it

Also I realize that this part
>NOTE: When you change the `clusterIP` setting of the service,

Is a little confusing. What we are changing is the `spec.http.service.spec.type` field from its previous value of `ClusterIP`

docs/orchestrating-elastic-stack-applications/elasticsearch/http-settings-tls-sans.asciidoc
```diff
@@ -8,7 +8,7 @@ endif::[]
 [id="{p}-{page_id}"]
 = HTTP settings and TLS SANs
 
-In the `spec.http.service.spec` section, you can change the Kubernetes service used to expose Elasticsearch:
+In the `spec.http.service.spec.type` section, you can change the Kubernetes service used to expose Elasticsearch:
```
anyasabo, line 11:
```suggestion
In the `spec.http.service.spec` field, you can change the Kubernetes service type used to expose Elasticsearch:
```

docs/orchestrating-elastic-stack-applications/elasticsearch/http-settings-tls-sans.asciidoc
alaudazzi: How about _temporary impact_ tout court?

sebgl: > What we are changing is the spec.http.service.spec.type field from its previous value of ClusterIP

I think this was about users providing a fixed ClusterIP that is not the one that currently exists. In which case we recreate the service entirely to get rid of the previous ClusterIP.

alaudazzi: Possible rewording:

> When you change the `ClusterIP` value in the `spec.http.service.spec.type` field,  ECK will delete and re-create the service, as `clusterIP` is an immutable field.

PR #3264, opened 2020-06-16
Add heartbeat driver

david-kow:
Adds heartbeat as a well-known type. This makes `spec.image` not required and creates appropriate roles in ES. The latter fixes current E2E test failure.

### DIFF
--- a/pkg/apis/beat/v1beta1/beat_types.go
+++ b/pkg/apis/beat/v1beta1/beat_types.go
@@ -13,7 +13,7 @@ import (
 
 // BeatSpec defines the desired state of a Beat.
 type BeatSpec struct {
-	// Type is the type of the Beat to deploy (filebeat, metricbeat, etc.). Any string can be used,
+	// Type is the type of the Beat to deploy (filebeat, metricbeat, heartbeat, etc.). Any string can be used,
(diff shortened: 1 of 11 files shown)


pkg/apis/beat/v1beta1/beat_types.go
```diff
@@ -13,7 +13,7 @@ import (
 
 // BeatSpec defines the desired state of a Beat.
 type BeatSpec struct {
-	// Type is the type of the Beat to deploy (filebeat, metricbeat, etc.). Any string can be used,
+	// Type is the type of the Beat to deploy (filebeat, metricbeat, heartbeat, etc.). Any string can be used,
 	// but well-known types will be recognized and will allow to provide sane default configurations.
```
anyasabo, line 17:
```suggestion
	// but well-known types will be provided with default configurations
```

not in this PR but noticed this could be clarified

david-kow: Good catch. I think we should reword it even more, so we don't mention default configurations at all.

anyasabo: I think you're right, I was trying to think of how to say "you don't need to specify the image"

PR #3273, opened 2020-06-17
Only provision Enterprise licenses as of 7.8.1

### LINKED ISSUE #3272: Enterprise license breaks incompatible clients
As of version 7.6 of Elasticsearch [we prefer](https://github.com/elastic/cloud-on-k8s/blob/eefc8afae3cd10cad5cfd49b74a4eed456b00cf6/pkg/controller/elasticsearch/client/model.go#L321) an Enterprise stack level license. 

But older clients e.g. Beats up to and including version 7.5 do not use the new `_license` API but call the  `_xpack` API which is not backwards compatible and exposes the concept of Enterprise license to client. The client is not expecting this value and errors out.

A fix for Elasticsearch [is underway](https://github.com/elastic/elasticsearch/pull/58217) but even if it lands in 7.8.1 and 7.9.0 we still have versions 7.6.x and 7.7.x that potentially break if users combine them with older versions of Beats. 

Two options come to mind: 
* document the problem and ask users to upgrade the clients to version > 7.5
* try to fix it by preferring a platinum license which comes with its own set of problems
   1. we could potentially detect if any ECK managed beat is running an incompatible version and provision a platinum license instead of an Enterprise license but we cannot easily detect the version of clients that are not managed by ECK
      - impossible to get correct given that clients outside of ECK's control exist
   2. we could just generally prefer a platinum license for any Elasticsearch version that is less than 7.8.1 or 7.9.0 (this would only be a an issue if we wanted to support Endpoint security for those versions)
      - simplest solution, my favourite
  3. we could make it configurable per cluster for users that use older clients e.g. via label but default to Enterprise
     -  comparatively complex implementation and would require us to maintain this for many versions of ECK to come, needs documentation and customers could still run into the issue if they don't read the docs.

pebrc:
This is to avoid breakage in older clients that call the deprecated
'_xpack' API to query the license level which is not backwards compatible.

* Do no longer provision Enterprise licenses to clusters < 7.8.1
* Do not attempt to install platinum or gold licenses into cluster >= 7.8.1 (will lead to errors in the logs as they get rejected, but maybe that is a helpful diagnostic tool? Happy to back that last commit out again)

This is optimistically assuming that[ the fix ](https://github.com/elastic/elasticsearch/pull/58217) is landing in 7.8.1

Fixes #3272

### DIFF
--- a/pkg/controller/elasticsearch/settings/merged_config.go
+++ b/pkg/controller/elasticsearch/settings/merged_config.go
@@ -111,7 +111,7 @@ func xpackConfig(ver version.Version, httpCfg commonv1.HTTPConfig) *CanonicalCon
 		cfg[esv1.XPackSecurityAuthcRealmsNativeNative1Order] = -99
 	}
 
-	if ver.IsSameOrAfter(version.MustParse("7.6.0")) {
+	if ver.IsSameOrAfter(version.MustParse("7.8.1")) {
 		cfg[esv1.XPackLicenseUploadTypes] = []string{
 			string(client.ElasticsearchLicenseTypeTrial), string(client.ElasticsearchLicenseTypeEnterprise),
 		}
--- a/pkg/apis/elasticsearch/v1/fields.go
+++ b/pkg/apis/elasticsearch/v1/fields.go
@@ -40,7 +40,7 @@ const (
 	XPackSecurityTransportSslKey                    = "xpack.security.transport.ssl.key"
 	XPackSecurityTransportSslVerificationMode       = "xpack.security.transport.ssl.verification_mode"
 
-	XPackLicenseUploadTypes = "xpack.license.upload.types" // >= 7.6.0
+	XPackLicenseUploadTypes = "xpack.license.upload.types" // supported >= 7.6.0 used as of 7.8.1
 )
 
 var UnsupportedSettings = []string{
--- a/pkg/controller/common/license/match_test.go
+++ b/pkg/controller/common/license/match_test.go
@@ -297,25 +297,6 @@ func Test_bestMatchAt(t *testing.T) {
 			wantFound: true,
 		},
 		{
-			name: "no result: platinum post 7.8.1",
-			args: args{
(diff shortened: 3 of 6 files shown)


anyasabo:
I realized we should update or remove the comment here
https://github.com/elastic/cloud-on-k8s/blob/master/pkg/apis/elasticsearch/v1/fields.go#L43

>	XPackLicenseUploadTypes = "xpack.license.upload.types" // >= 7.6.0

pkg/controller/elasticsearch/settings/merged_config.go
```diff
@@ -111,7 +111,7 @@ func xpackConfig(ver version.Version, httpCfg commonv1.HTTPConfig) *CanonicalCon
 		cfg[esv1.XPackSecurityAuthcRealmsNativeNative1Order] = -99
 	}
 
-	if ver.IsSameOrAfter(version.MustParse("7.6.0")) {
+	if ver.IsSameOrAfter(version.MustParse("7.8.1")) {
 		cfg[esv1.XPackLicenseUploadTypes] = []string{
```
sebgl, line 115: I'm wondering what will happen if people do a version upgrade from 7.7.0 (platinum license) to 7.8.1 (requires an Enterprise license). Can they just continue with the existing platinum license, and only a call to refresh the license would be impacted?

pebrc: This could potentially lead to unpleasant errors in the logs. If the license was still valid everything would be fine. But once it expires or gets close to expire the operator would still try to apply a platinum license to the 7.8.1 cluster which would fail because of the restriction.  

I think we should at least not try to apply a platinum license.

pebrc:
Apologies for the back and forth but after sleeping over this I think the last commit which I just reverted might cause more problems than it solves. Leaving the matching logic unchanged gives us a bit more maneuvering space in case things go wrong at the cost of an unhelpful error message. 

I think we should think about a more comprehensive approach that would give us seamless bwc compatibility with older licenses but that should be a separate PR.

Issue #3274, opened 2020-06-17
Volume permission errors when running Enterprise Search on OCP (E2E)


david-kow:
`TestSamples/enterprisesearch-ent_es/Enterprise_Search_Pods_should_eventually_be_read` is failing consistantly since being introduced (all runs between [140](https://devops-ci.elastic.co/view/cloud-on-k8s/job/cloud-on-k8s-e2e-tests-ocp/140/) and [144](https://devops-ci.elastic.co/view/cloud-on-k8s/job/cloud-on-k8s-e2e-tests-ocp/144/), but [142](https://devops-ci.elastic.co/view/cloud-on-k8s/job/cloud-on-k8s-e2e-tests-ocp/142/) which fails earlier).

sebgl:
```
"containerStatuses": [
                    {
                        "containerID": "cri-o://3b77c37c467897306840f57a2507f8e1be3a7f557256f4fd6cf12f35283cfe80",
                        "image": "docker.elastic.co/enterprise-search/enterprise-search:7.7.1",
                        "imageID": "docker.elastic.co/enterprise-search/enterprise-search@sha256:96a6484cc077576c1554cad1e537c3ed524634202e97478d617aca5bc0562a6d",
                        "lastState": {
                            "terminated": {
                                "containerID": "cri-o://3b77c37c467897306840f57a2507f8e1be3a7f557256f4fd6cf12f35283cfe80",
                                "exitCode": 1,
                                "finishedAt": "2020-06-17T01:12:01Z",
                                "reason": "Error",
                                "startedAt": "2020-06-17T01:12:01Z"
                            }
                        },
                        "name": "enterprise-search",
                        "ready": false,
                        "restartCount": 6,
                        "started": false,
                        "state": {
                            "waiting": {
                                "message": "back-off 5m0s restarting failed container=enterprise-search pod=ent-sample-j2bt-ent-9dbdbdc8b-tf9j9_e2e-qj18m-mercury(15453832-0fc8-4964-a45e-5476ee5a87b4)",
                                "reason": "CrashLoopBackOff"
                            }
                        }
                    }
```

sebgl:
Reproduced locally:

```
❯ k logs -p ent-sample-9v6s-ent-77cc976cd-szs25 -n e2e-mercury                                       
/usr/local/bin/docker-entrypoint.sh: line 24: /usr/share/enterprise-search/config/enterprise-search.yml: Permission denied
```

[...]

PR #3277, opened 2020-06-17
Fix Enterprise Search mounted volumes for restricted environments

sebgl:
Running Enterprise Search on a restricted environment (Openshift) failed for a few reasons.
This PR fixes it.

## 1. Permission to write the configuration file

Enterprise Search attempts to overwrite the content of `/usr/share/enterprise-search/enterprise-search.yml` with configuration settings from environment variables, unless that file is already not the default one. Even though we mounted the configuration file in a different place. When that happens, the Enterprise Search process may not have permissions to write in that folder that is restricted to the `enterprise-search` user.

To deal with this, we can simply mount the configuration file directly in `/usr/share/enterprise-search/enterprise-search.yml`. Enterprise Search detects this file is not the default one and does not attempt to modify it.

To keep the number of secrets small, we keep the readiness probe in the same secret as the configuration file. For that reason, the PR introduces a second volume dedicated to the readiness probe script (mounted in `/mnt/elastic-internal/scripts/readiness-probe.sh`), that relies on the same Secret as the config volume.

## 2. Permission to write logs

Enterprise Search writes logs to `/var/log/enterprise-search` by default, but may not have permissions to create that directory. To fix it, we mount an emptyDir in `/var/log/enterprise-search`. Which is a good idea anyway in order to not have the container disk usage grow with logs. We already do the same for Elasticsearch.

Fixes https://github.com/elastic/cloud-on-k8s/issues/3274.

### DIFF
--- a/pkg/controller/enterprisesearch/config.go
+++ b/pkg/controller/enterprisesearch/config.go
@@ -32,6 +32,7 @@ const (
 	ESCertsPath              = "/mnt/elastic-internal/es-certs"
 	ConfigMountPath          = "/usr/share/enterprise-search/config/enterprise-search.yml"
 	ConfigFilename           = "enterprise-search.yml"
+	ReadinessProbeMountPath  = "/mnt/elastic-internal/scripts/readiness-probe.sh"
 	ReadinessProbeFilename   = "readiness-probe.sh"
 	ReadinessProbeTimeoutSec = 5
 
@@ -42,6 +43,10 @@ const (
 func ConfigSecretVolume(ent entv1beta1.EnterpriseSearch) volume.SecretVolume {
 	return volume.NewSecretVolume(name.Config(ent.Name), "config", ConfigMountPath, ConfigFilename, 0644)
 }
+
+func ReadinessProbeSecretVolume(ent entv1beta1.EnterpriseSearch) volume.SecretVolume {
+	// reuse the config secret
+	return volume.NewSecretVolume(name.Config(ent.Name), "readiness-probe", ReadinessProbeMountPath, ReadinessProbeFilename, 0644)
 }
 
 // Reconcile reconciles the configuration of Enterprise Search: it generates the right configuration and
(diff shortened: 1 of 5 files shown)


pkg/controller/enterprisesearch/name/name.go
```diff
@@ -27,3 +28,7 @@ func Deployment(entName string) string {
 func Config(entName string) string {
 	return EntNamer.Suffix(entName, configSuffix)
 }
+
+func Scripts(entName string) string {
```
anyasabo, line 32: It looks like this is not called anywhere, correct?

anyasabo:
For point 2, it looked like the default was not to write any logs to files, and to write to stdout instead:
https://www.elastic.co/guide/en/enterprise-search/current/configuration.html

But then I see in the dockerfile they set these two env vars:

```
ENV log_directory=/var/log/enterprise-search
ENV filebeat_log_directory=/var/log/enterprise-search
```

Writing logs to files by default in containers feels like the wrong thing and something we should request be changed. But obviously we need to do something for now so I think I am good with your approach.

For point 1, that also feels like odd behavior we should probably request be changed. Writing over config files is not great.

pkg/controller/enterprisesearch/config.go
```diff
@@ -40,7 +41,12 @@ const (
 )
 
 func ConfigSecretVolume(ent entv1beta1.EnterpriseSearch) volume.SecretVolume {
-	return volume.NewSecretVolumeWithMountPath(name.Config(ent.Name), "config", ConfigMountPath)
+	return volume.NewSecretVolume(name.Config(ent.Name), "config", ConfigMountPath, ConfigFilename, 0644)
```
thbkrkr, line 44: If the config file already exists, entsearch does not attempt to modify it. So, should we be as restrictive as possible and set `0444` or even `0400`?

pkg/controller/enterprisesearch/name/name.go
sebgl: Right! Good catch.

pkg/controller/enterprisesearch/config.go
sebgl: `0444` sounds good, yes 👍 
I'm a bit afraid of `0400` because of the various ways the container may be ran with a different user.

sebgl:
> For point 2, it looked like the default was not to write any logs to files, and to write to stdout instead

On startup, Enterprise Search also creates `/usr/share/enterprise-search/log` if no log paths are specified. Agreed that the experience feels a bit strange. I'll open an issue in the (private) Enterprise Search Github repo. For now I think we need to deal with it.

> For point 1, that also feels like odd behavior we should probably request be changed. Writing over config files is not great.

Agreed, I'm also opening an issue in the Enterprise Search Github repo.

Issue #3221, opened 2020-06-11
Use a single configRef secret in the Enterprise Search CRD


david-kow:
Makes more sense as it's a list. Also, use the `ConfigSource` from `common`.

sebgl:
Discussed out of band: let's not make this a list and be a single secret reference instead, to better match the Beat CRD.

PR #3279, opened 2020-06-17
Allow a single configRef to be specified in Enterprise Search

sebgl:
Let's be consistent with the Beat CRD, and accept a single secret reference in `configRef`.

Chances that people would like to use multiple secrets are rather small, and we could still add a plural `configRefs` later.

Fixes https://github.com/elastic/cloud-on-k8s/issues/3221.


pkg/apis/enterprisesearch/v1beta1/enterprisesearch_types.go
```diff
@@ -26,11 +26,10 @@ type EnterpriseSearchSpec struct {
 	// Config holds the Enterprise Search configuration.
 	Config *commonv1.Config `json:"config,omitempty"`
 
-	// ConfigRef contains references to Kubernetes Secrets holding the Enterprise Search configuration.
-	// Configuration settings are merged and have prcedence over plain text settings specified in  `config`.
-	// Multiple secrets can be referenced: if duplicate settings exist in multiple secrets,
-	// the last one takes precedence.
-	ConfigRef []ConfigSource `json:"configRef,omitempty"`
+	// ConfigRef contains references to an existing Kubernetes Secret holding the Enterprise Search configuration.
```
david-kow, line 29:
```suggestion
	// ConfigRef contains a reference to an existing Kubernetes Secret holding the Enterprise Search configuration.
```

pkg/apis/enterprisesearch/v1beta1/enterprisesearch_types.go
```diff
@@ -26,11 +26,10 @@ type EnterpriseSearchSpec struct {
...
+	// Configuration settings are merged and have precedence over plain text settings specified in `config`.
```
david-kow, line 30:
```suggestion
	// Configuration settings are merged with and have precedence over settings specified in `config`.
```
Without "with" it seems that they are many of them that are being merged together. 
I think I'd remove "plain text" since the one in configRef are fairly plain texty to me :) Even the example we have is plain text.

pkg/controller/enterprisesearch/config.go
```diff
@@ -224,43 +224,51 @@ func configRefWatchName(ent types.NamespacedName) string {
 	return fmt.Sprintf("%s-%s-configref", ent.Namespace, ent.Name)
 }
 
-// parseConfigRef builds a single merged CanonicalConfig from the secrets referenced in configRef,
-// and ensures watches are correctly set on those secrets.
+// parseConfigRef builds a CanonicalConfig from the secret referenced in configRef,
+// and ensures a watch is set on this secret.
 func parseConfigRef(driver driver.Interface, ent entv1beta1.EnterpriseSearch) (*settings.CanonicalConfig, error) {
```
david-kow, line 229: Maybe it would be worth to pull it out to common at some point?

pkg/controller/enterprisesearch/config.go
```diff
@@ -224,43 +224,51 @@ func configRefWatchName(ent types.NamespacedName) string {
...
-		if secretRef.SecretName == "" {
-			continue
-		}
-		secretNames = append(secretNames, secretRef.SecretName)
+	var secretNames []string
+	if ent.Spec.ConfigRef != nil && ent.Spec.ConfigRef.SecretName != "" {
+		secretNames = append(secretNames, ent.Spec.ConfigRef.SecretName)
 	}
 	nsn := k8s.ExtractNamespacedName(&ent)
+	// will reset any existing watch if secretNames is empty
```
thbkrkr, line 235: Looks right but seems like a side-effect of the function IMO.
I noticed that we are doing almost the same thing for beats. Could we factorize the code?
I also see that for beats we [return early](https://github.com/elastic/cloud-on-k8s/blob/29d7abca1e4a8f60f8712b2f64814d9aa51d489e/pkg/controller/beat/common/config.go#L117-L138) so it looks like we could leak a watch.

sebgl: Yes! I think it will be easy enough to factorize, and fix the Beat watch leak at the same time 👍

pkg/controller/enterprisesearch/config.go
sebgl: done in 80e410a

pkg/controller/common/configref.go
```diff
@@ -0,0 +1,75 @@
...
+	if configRef != nil && configRef.SecretName != "" {
+		secretNames = append(secretNames, configRef.SecretName)
+	}
+	if err := watches.WatchUserProvidedSecrets(resourceNsn, driver.DynamicWatches(), ConfigRefWatchName(resourceNsn), secretNames); err != nil {
+		return nil, err
+	}
+
+	if len(secretNames) == 0 {
+		// no secret referenced, nothing to do
+		return settings.NewCanonicalConfig(), nil
```
david-kow, line 54: Shouldn't this return `nil, nil`?

sebgl: In the enterprise search reconciliation we're doing:

```
// merge with user settings last so they take precedence
err = cfg.MergeWith(reusedCfg, tlsCfg, associationCfg, userProvidedCfg, userProvidedSecretCfg)
```

where any of those config can just be empty.  I think it feels clean that if there's no config to parse, and no error, you still get a valid empty result, for which you don't have to check nil pointers.

david-kow: I think it's nice to be able to rely on the fact that `nil` means nothing was there (vs something was there but was empty).
If we don't want to return nils maybe we should return struct instead of a pointer?
If `.MergeWith` is the only concern then I'd go with `nil, nil` since it handles nils just fine:
```
func (c *CanonicalConfig) MergeWith(cfgs ...*CanonicalConfig) error {
	for _, c2 := range cfgs {
		if c2 == nil {
			continue
		}
...
```

sebgl: 👍 I'm fine with that, I'll make the change to `nil`.

PR #3282, opened 2020-06-17
Suggest starting an ECK trial license in Enterprise Search docs

sebgl:
In order to start Workplace Search users will need an Elastic license. A
great way to try it is to start a trial license.
While a trial can be started by clicking a button on the UI, we want to
encourage users to start a trial at ECK level, so that:
* they also benefit from ECK licensed features
* they don't hit a bug in Enterprise Search 7.7.0 (fixed in 7.8.0) where
  the start trial call returns an error 500 (once)

### DIFF
--- a/docs/orchestrating-elastic-stack-applications/enterprise-search.asciidoc
+++ b/docs/orchestrating-elastic-stack-applications/enterprise-search.asciidoc
@@ -15,7 +15,7 @@ This section describes how to deploy, configure and access Enterprise Search wit
 [id="{p}-enterprise-search-quickstart"]
 == Quickstart
 
-NOTE: Workplace Search requires an Elastic license. Give it a try by enabling an link:k8s-licensing.html[ECK trial license].
+NOTE: Workplace Search requires an Enterprise license on ECK. Give it a try with a link:k8s-licensing.html[trial license].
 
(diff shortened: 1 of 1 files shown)


docs/orchestrating-elastic-stack-applications/enterprise-search.asciidoc
```diff
@@ -15,6 +15,8 @@ This section describes how to deploy, configure and access Enterprise Search wit
 [id="{p}-enterprise-search-quickstart"]
 == Quickstart
 
+NOTE: Workplace Search requires an Elastic license. Give it a try by enabling an link:k8s-licensing.html[ECK trial license].
```
anyasabo, line 18:
```suggestion
NOTE: Workplace Search requires an Enterprise license. Give it a try with a link:k8s-licensing.html[trial license].
```

Just a suggestion, I'm definitely in favor of adding something here and don't feel too opinionated

I don't feel strongly either way

sebgl: > Workplace Search requires an Enterprise license

It is not _technically_ correct so I'm on the fence here.

pebrc: I would not introduce `Platinum` though in the context of ECK which would lead to more confusion.  Maybe we could have 
```suggestion
NOTE: Workplace Search requires an Enterprise license on ECK. Give it a try with a link:k8s-licensing.html[trial license].
```

Issue #2279, opened 2019-12-17
Set kibana reporting encryption key


anyasabo:
Similar to [the session encryption key](https://github.com/elastic/cloud-on-k8s/issues/1738), we should set the reporting key `xpack.reporting.encryptionKey` by default so that in progress reports can be resumed:
>Set to any text string. By default, Kibana will generate a random key when it starts, which will cause pending reports to fail after restart. Configure this setting to preserve the same key across multiple restarts and multiple instances of Kibana.

I noticed this warning log message when verifying the other encryption key functionality:
>{"type":"log","@timestamp":"2019-12-17T16:12:58Z","tags":["reporting","warning"],"pid":6,"message":"Generating a random key for xpack.reporting.encryptionKey. To prevent pending reports from failing on restart, please set xpack.reporting.encryptionKey in kibana.yml"}

Docs:
https://www.elastic.co/guide/en/kibana/current/reporting-settings-kb.html#general-reporting-settings

anyasabo:
Ditto `xpack.encrypted_saved_objects.encryptionKey`

PR #3285, opened 2020-06-17
Add additional Kibana encryption keys

anyasabo:
Fix https://github.com/elastic/cloud-on-k8s/issues/2279 by setting the additional encryption keys if they are not explicitly set already

Also changes `server.host` from `0` to `0.0.0.0` because this was being thrown in the logs:
>{"type":"log","@timestamp":"2020-06-17T19:39:10Z","tags":["error","reporting"],"pid":6,"message":"Error: Found 'server.host: \"0\"' in settings. This is incompatible with Reporting. To enable Reporting to work, 'xpack.reporting.kibanaServer.hostname: 0.0.0.0' is being automatically to the configuration. You can change to 'server.host: 0.0.0.0' or add 'xpack.reporting.kibanaServer.hostname: 0.0.0.0' in kibana.yml to prevent this message.\n    at validateServerHost (/usr/share/kibana/x-pack/legacy/plugins/reporting/server/lib/validate/validate_server_host.js:24:11)\n    at runValidations (/usr/share/kibana/x-pack/legacy/plugins/reporting/server/lib/validate/index.js:25:256)\n    at ReportingPlugin.setup (/usr/share/kibana/x-pack/legacy/plugins/reporting/server/plugin.js:40:5)"}

And there wasn't any mention in the [Kibana config docs](https://www.elastic.co/guide/en/kibana/current/settings.html), but the [ES docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-network.html) indicate 0 should be semantically equivalent to 0.0.0.0

This resolved most of the startup warnings for Kibana, another one that stands out is:

>{"type":"log","@timestamp":"2020-06-17T19:39:07Z","tags":["warning","reporting"],"pid":6,"message":"Enabling the Chromium sandbox provides an additional layer of protection."}

Which seems like a false alarm since the [docs](https://www.elastic.co/guide/en/kibana/current/reporting-chromium-sandbox.html#_docker) recommend disabling it in docker. Reached out to the kibana team for clarity there

### DIFF
--- a/pkg/controller/kibana/config_settings.go
+++ b/pkg/controller/kibana/config_settings.go
@@ -189,7 +189,13 @@ func getOrCreateReusableSettings(c k8s.Client, kb kbv1.Kibana) (*settings.Canoni
 	if len(r.ReportingKey) == 0 {
 		r.ReportingKey = string(common.RandomBytes(64))
 	}
-	if len(r.SavedObjectsKey) == 0 {
+
+	kbVer, err := version.Parse(kb.Spec.Version)
+	if err != nil {
+		return nil, err
+	}
+	// xpack.encryptedSavedObjects.encryptionKey was only added in 7.6.0 and earlier versions error out
+	if len(r.SavedObjectsKey) == 0 && kbVer.IsSameOrAfter(version.MustParse("7.6.0")) {
 		r.SavedObjectsKey = string(common.RandomBytes(64))
 	}
 	return settings.MustCanonicalConfig(r), nil
--- a/pkg/controller/kibana/config_settings.go
+++ b/pkg/controller/kibana/config_settings.go
@@ -43,6 +43,8 @@ const (
 	XpackMonitoringUIContainerElasticsearchEnabled = "xpack.monitoring.ui.container.elasticsearch.enabled"
 	XpackLicenseManagementUIEnabled                = "xpack.license_management.ui.enabled" // >= 7.6
 	XpackSecurityEncryptionKey                     = "xpack.security.encryptionKey"
+	XpackReportingEncryptionKey                    = "xpack.reporting.encryptionKey"
+	XpackEncryptedSavedObjectsEncryptionKey        = "xpack.encrypted_saved_objects.encryptionKey"
 
 	ElasticsearchSslCertificateAuthorities = "elasticsearch.ssl.certificateAuthorities"
 	ElasticsearchSslVerificationMode       = "elasticsearch.ssl.verificationMode"
@@ -137,7 +139,9 @@ func VersionDefaults(kb *kbv1.Kibana, v version.Version) *settings.CanonicalConf
 
 // reusableSettings captures secrets settings in the Kibana configuration that we want to reuse.
 type reusableSettings struct {
-	EncryptionKey string `config:"xpack.security.encryptionKey"`
+	EncryptionKey   string `config:"xpack.security.encryptionKey"`
+	ReportingKey    string `config:"xpack.reporting.encryptionKey"`
+	SavedObjectsKey string `config:"xpack.encrypted_saved_objects.encryptionKey"`
 }
 
 // getExistingConfig retrieves the canonical config for a given Kibana, if one exists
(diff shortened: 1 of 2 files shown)


pkg/controller/kibana/config_settings.go
```diff
@@ -43,6 +43,8 @@ const (
 	XpackMonitoringUIContainerElasticsearchEnabled = "xpack.monitoring.ui.container.elasticsearch.enabled"
 	XpackLicenseManagementUIEnabled                = "xpack.license_management.ui.enabled" // >= 7.6
 	XpackSecurityEncryptionKey                     = "xpack.security.encryptionKey"
+	XpackReportingEncryptionKey                    = "xpack.reporting.encryptionKey"
+	XpackEncryptedSavedObjectsEncryptionKey        = "xpack.encryptedSavedObjects.encryptionKey"
```
pebrc, line 47: I believe this setting is not available in all version of Kibana we support. I am trying to figure out when it was added but it is definitely not in 6.8

anyasabo: Yep and it errors out too :/

>{"type":"log","@timestamp":"2020-06-18T13:45:18Z","tags":["fatal","root"],"pid":1,"message":"{ Error: \"xpack.encryptedSavedObjects.encryptionKey\" setting was not applied. Check for spelling errors and ensure that expected plugins are installed.\n    at KbnServer.exports.default (/usr/share/kibana/src/server/config/complete.js:88:17) code: 'InvalidConfig', processExitCode: 64 }"}

> FATAL  Error: "xpack.encryptedSavedObjects.encryptionKey" setting was not applied. Check for spelling errors and ensure that expected plugins are installed.

anyasabo:
Changed it to only add that key for 7.6+ and tweaked some of the tests. I'll be out on Tuesday so feel free to merge or push commits to this if necessary to merge it in.

pkg/controller/kibana/config_settings.go
```diff
@@ -182,13 +186,25 @@ func getOrCreateReusableSettings(c k8s.Client, kb kbv1.Kibana) (*settings.Canoni
...
+	if len(r.ReportingKey) == 0 {
+		r.ReportingKey = string(common.RandomBytes(64))
+	}
+
+	kbVer, err := version.Parse(kb.Spec.Version)
+	if err != nil {
+		return nil, err
+	}
+	// xpack.encryptedSavedObjects.encryptionKey was only added in 7.6.0 and earlier versions error out
+	if len(r.SavedObjectsKey) == 0 && kbVer.IsSameOrAfter(version.MustParse("7.6.0")) {
```
pebrc, line 198:
```suggestion
	if len(r.SavedObjectsKey) == 0 && kbVer.IsSameOrAfter(version.From(7,6,0)) {
```

pkg/controller/kibana/config_settings_test.go
```diff
@@ -346,9 +379,18 @@ func TestNewConfigSettingsExistingEncryptionKey(t *testing.T) {
 	require.NoError(t, err)
 	var gotCfg map[string]interface{}
 	require.NoError(t, got.Unpack(&gotCfg))
+
```
pebrc, line 382: I wonder if it would make the test easier to read if we were to unpack into `reusableSettings`

pkg/controller/kibana/config_settings.go
charith-elastic:
```suggestion
	if len(r.SavedObjectsKey) == 0 && kbVer.IsSameOrAfter(version.From(7, 6, 0)) {
```

pkg/controller/kibana/config_settings.go
```diff
@@ -182,13 +186,25 @@ func getOrCreateReusableSettings(c k8s.Client, kb kbv1.Kibana) (*settings.Canoni
...
+	if len(r.ReportingKey) == 0 {
+		r.ReportingKey = string(common.RandomBytes(64))
+	}
+
+	kbVer, err := version.Parse(kb.Spec.Version)
+	if err != nil {
+		return nil, err
+	}
+	// xpack.encryptedSavedObjects.encryptionKey was only added in 7.6.0 and earlier versions error out
+	if len(r.SavedObjectsKey) == 0 && kbVer.IsSameOrAfter(version.From(7,6,0)) {
```
thbkrkr, line 198:
```suggestion
	if len(r.SavedObjectsKey) == 0 && kbVer.IsSameOrAfter(version.From(7, 6, 0)) {
```

PR #3292, opened 2020-06-18
Add drivers for packetbeat, journalbeat and auditbeat

david-kow:
This PR adds packetbeat, journalbeat and auditbeat as well-known Beat types. This allows to setup dashboards when `kibanaRef` is provided (when Beat supports it), precreate roles for those Beat types and choose the right image.

All three Beats have E2E tests added with exemplary configs together with required PSPs.

### DIFF
--- a/config/e2e/beat-roles.yaml
+++ b/config/e2e/beat-roles.yaml
@@ -75,3 +75,18 @@ rules:
   - elastic.beat.restricted
   verbs:
   - use
+---
+# ClusterRole allowing use of Auditbeat-specific PSP
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: elastic-auditbeat-restricted
+rules:
+- apiGroups:
+  - policy
+  resources:
+  - podsecuritypolicies
+  resourceNames:
+  - elastic.auditbeat.restricted
+  verbs:
+  - use
--- a/config/e2e/beat-roles.yaml
+++ b/config/e2e/beat-roles.yaml
@@ -90,3 +90,33 @@ rules:
   - elastic.auditbeat.restricted
   verbs:
   - use
+---
+# ClusterRole allowing use of Packetbeat-specific PSP
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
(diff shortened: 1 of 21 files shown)


config/e2e/beat-roles.yaml
```diff
@@ -75,3 +75,48 @@ rules:
...
+- apiGroups:
+  - policy
+  resources:
+  - podsecuritypolicies
+  resourceNames:
+  - elastic.packetbeat.restricted
+  verbs:
+  - use
+---
+# ClusterRole allowing use of Packetbeat-specific PSP
```
thbkrkr, line 109:
```suggestion
# ClusterRole allowing use of Journalbeat-specific PSP
```

pkg/apis/beat/v1beta1/beat_types.go
```diff
@@ -13,8 +13,9 @@ import (
 
 // BeatSpec defines the desired state of a Beat.
 type BeatSpec struct {
-	// Type is the type of the Beat to deploy (filebeat, metricbeat, heartbeat, etc.). Any string can be used,
-	// but well-known types will be recognized and will allow to provide sane default configurations.
+	// Type is the type of the Beat to deploy (filebeat, metricbeat, heartbeat, auditbeat, journalbeat, packetbeat, etc.).
+	// Any string can be used, but well-known types will be recognized and will allow to use the correct image and
```
anyasabo, line 17: maybe "but well-known types..."

can be "but well-known types will have the image field defaulted and have the appropriate Elasticsearch roles created automatically"?

pkg/apis/beat/v1beta1/beat_types.go
```diff
@@ -13,8 +13,9 @@ import (
...
+	// Elasticsearch roles. It will also allow for the initial setup if the type supports it and `kibanaRef` field is specified.
```
anyasabo, line 18: >It will also allow for the initial setup if the type supports it and `kibanaRef` field is specified.

this isn't clear to me, I think maybe just allows for dashboard setup when combined with a kibanaRef?

david-kow:
> LGTM. I have no idea if those configs are valid but what could go wrong?

We do have E2E tests with some validations, so hopefully this gives us enough confidence.

PR #3295, opened 2020-06-18
Do not manage keystore if already initialized

### LINKED ISSUE #3294: Init containers should tolerate restarts
It may happen that an init container is restarted even if the Pod is already running.

See the following issues:
* https://github.com/kubernetes/kubernetes/issues/86531
* https://github.com/kubernetes/kubernetes/issues/67261#issuecomment-414152720

The main issue comes from the init containers that initialize the keystores. For example the Elasticsearch keystore can only be created once, any additional run expect an input from stdin:

```
echo 'Initializing keystore.'
+ /usr/share/elasticsearch/bin/elasticsearch-keystore create
Initializing keystore.
Exception in thread "main" java.lang.IllegalStateException: unable to read from standard input; is standard input open and a tty attached?
        at org.elasticsearch.cli.Terminal$SystemTerminal.readText(Terminal.java:212)
```

Another issue is when we use the `--force` flag, as we do for the APM Server for example: we should not execute blindly `--force` since this is (re)creating a temporary empty keystore while the APM Server is running.

I think we can solve this problem by using a "flag" on the filesystem. If we want to do some "fine grain recovery" in order to recover from various, more complicated, situations then we need to dig into any keystore specific behaviour, maybe deal with some implementation details, and keep in mind that the keystore might be in used when the init container is running.

barkbay:
Fix #3294 by creating a file to record if the keystore has been initialized.

### DIFF
--- a/pkg/controller/beat/common/pod.go
+++ b/pkg/controller/beat/common/pod.go
@@ -61,7 +61,7 @@ func initContainerParameters(typ string) keystore.InitContainerParameters {
 		KeystoreCreateCommand:         fmt.Sprintf("%s keystore create --force", typ),
 		KeystoreAddCommand:            fmt.Sprintf(`cat "$filename" | %s keystore add "$key" --stdin --force`, typ),
 		SecureSettingsVolumeMountPath: keystore.SecureSettingsVolumeMountPath,
-		DataVolumePath:                fmt.Sprintf(DataPathTemplate, typ),
+		KeystoreVolumePath:            fmt.Sprintf(DataPathTemplate, typ),
 		Resources:                     defaultResources,
 	}
 }
--- a/pkg/controller/common/keystore/initcontainer.go
+++ b/pkg/controller/common/keystore/initcontainer.go
@@ -30,7 +30,7 @@ type InitContainerParameters struct {
 	Resources corev1.ResourceRequirements
 }
 
-// script is a small bash script to create a the keystore,
+// script is a small bash script to create an Elastic Stack keystore,
 // then add all entries from the secure settings secret volume into it.
 const script = `#!/usr/bin/env bash
 
--- a/pkg/controller/common/keystore/initcontainer.go
+++ b/pkg/controller/common/keystore/initcontainer.go
@@ -16,12 +16,12 @@ const (
 	InitContainerName = "elastic-internal-init-keystore"
 )
 
-// InitContainerParameters helps to create a valid keystore init script for Kibana or the APM server.
+// InitContainerParameters helps to create a valid keystore init script.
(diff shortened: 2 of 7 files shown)


pkg/controller/common/keystore/resources_test.go
```diff
@@ -101,6 +101,13 @@ func TestResources(t *testing.T) {
 
 set -eux
 
+keystore_initialized_flag=/bar/data/elastic-internal-init-keystore.ok
```
anyasabo, line 104: This isn't necessarily your change but it seems odd we're not reusing the `script` constant here

pkg/controller/beat/common/pod.go
```diff
@@ -61,7 +61,7 @@ func initContainerParameters(typ string) keystore.InitContainerParameters {
 		KeystoreCreateCommand:         fmt.Sprintf("%s keystore create --force", typ),
 		KeystoreAddCommand:            fmt.Sprintf(`cat "$filename" | %s keystore add "$key" --stdin --force`, typ),
 		SecureSettingsVolumeMountPath: keystore.SecureSettingsVolumeMountPath,
-		DataVolumePath:                fmt.Sprintf(DataPathTemplate, typ),
+		KeystoreVolumePath:            fmt.Sprintf(DataPathTemplate, typ),
```
anyasabo, line 64: Also not your change, but `typ` is a typo of `type`, correct?

barkbay:
> LGTM, though I think we may also want to raise an issue with ES so that this call can be idempotent. We would still need this fix for old versions but would be nice for future ones.

We would also need to raise some issues for all the other stack components to ensure that there is a consistent behaviour across the stack. I agree that it would be nice, but I'm not sure we want to coordinate this effort.

pkg/controller/beat/common/pod.go
pebrc: `type` is a reserved word though, that's probably why it's `typ`.

thbkrkr: `typ` is acceptable even I'm agree it's disturbing at the first glance. Alternative ideas: `kind`, `beatType`, `specType`, `beatSpecType`.

pkg/controller/common/keystore/initcontainer.go
```diff
@@ -30,12 +30,19 @@ type InitContainerParameters struct {
 	Resources corev1.ResourceRequirements
 }
 
-// script is a small bash script to create a Kibana or APM keystore,
+// script is a small bash script to create a the keystore,
```
thbkrkr, line 33:
```suggestion
// script is a small bash script to create an Elastic Stack keystore,
```

PR #3311, opened 2020-06-24
Add upgrade test harness

### LINKED ISSUE #2218: Create a test to cover operator upgrades
Relates https://github.com/elastic/cloud-on-k8s/issues/479.

We should test the operator upgrade, and make sure it does not break existing resources.

A potential scenario:

* spin up a k8s cluster
* apply https://download.elastic.co/downloads/eck/0.9.0/all-in-one.yaml (alpha version)
* apply v1alpha1 Elasticsearch, Kibana & APM resources
* apply https://download.elastic.co/downloads/eck/1.0.0-beta1/all-in-one.yaml (beta version) - should not error out
* make sure the existing alpha cluster is still there untouched
* apply v1beta1 Elasticsearch, Kibana & APM resources
* delete the beta operator statefulset
* apply the current master operator & crd manifests - should not error out
* make sure the existing v1beta1 cluster still works fine
* apply a mutation to the existing `v1beta1` cluster, using CRDs `v1`

charith-elastic:
Adds an automated test harness to run through the process of upgrading ECK releases starting from a particular version. By default, it will start from the alpha release (0.9) and keep upgrading to subsequent releases until it ends up with a deployment of the current master version.

This PR lays the foundation for more complex testing scenarios. Currently the tests are structured around assumptions that are true for ECK versions released so far. Further development may be necessary to add more complicated tests and to cater for future releases that deviate from current assumptions.

Fixes #2218


hack/upgrade-test-harness/testdata/alpha/install.yaml
```diff
@@ -0,0 +1,970 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
thbkrkr, line 1: Have you considered downloading the manifest from https://download.elastic.co/downloads/eck/0.9.0/all-in-one.yaml?

charith-elastic: Yes. I didn't do so for several reasons:
- We'll lose the installation file if the content is removed from the website or if the URLs are shuffled around
- Downloading over the internet increases the chances of flakiness due to network issues
- If we run the test regularly, our download stats will be artificially inflated

thbkrkr: > Yes. I didn't do so for several reasons:

That seems reasonable. 
 
> * We'll lose the installation file if the content is removed from the website or if the URLs are shuffled around

I think that if we want to prevent that, this is not the right place to backup the yaml. If the URLS are shuffled around, we will have to update the doc anyway.

> * Downloading over the internet increases the chances of flakiness due to network issues

Yes, but I tend to hope that hitting S3 is very very stable.

> * If we run the test regularly, our download stats will be artificially inflated

Yes, but there is a trick to avoid that :).

Maybe I'm nitpicking but it seems wrong for me. I understand these reasons and I find them relatively reasonable. I will not oppose putting the yaml in git but I admit that it bothers me a little.

hack/upgrade-test-harness/fixture/eck.go
```diff
@@ -0,0 +1,257 @@
...
+	wantAPMServerNodes     = 1
+	wantElasticsearchNodes = 3
+	wantHealth             = "green"
+	wantKibanaNodes        = 1
+)
+
+// TestParam holds parameters for a test.
+type TestParam struct {
+	Name            string `json:"name"`
+	OperatorVersion string `json:"operatorVersion"`
```
pebrc, line 33: Not sure what the purpose of this attribute is as the actual operator version is determined by the YAML in `testdata` and this is not even used for logging

hack/upgrade-test-harness/main.go
```diff
@@ -0,0 +1,186 @@
...
+			if err := f.Execute(ctx); err != nil {
+				return err
+			}
+		}
+
+		prevTestParam = &currTestParam
+	}
+
+	// try to scale the last deployed release
+	if prevTestParam != nil {
```
pebrc, line 129: It's not clear to me why scaling is only happening at the end.

hack/upgrade-test-harness/main.go
```diff
@@ -0,0 +1,186 @@
...
+	pos, err := conf.ReleasePos(name)
+	if err != nil {
+		panic(err)
+	}
+
+	return pos
+}
+
+func setupUpcomingRelease(installYAML string) error {
+	in, err := os.Open(installYAML)
```
pebrc, line 146: Should this be validated early on to fail fast and not only after running through all the other tests given that `upcoming` will be the last test?

hack/upgrade-test-harness/fixture/eck.go
charith-elastic: Yes, it's a leftover from an earlier attempt. However, I kind of like the fact that it acts as documentation when looking at the config file, so I am inclined to leave it as is unless you feel very strongly about it.

hack/upgrade-test-harness/main.go
charith-elastic: I think the purpose of the test is to make sure that the operator works correctly after going through several in-place upgrades so the scaling is done only at the end. Scaling on every upgrade step doesn't add much value because they are past releases and it's a bit too late to fix any problems they might have anyway.

charith-elastic:
This is an almost near faithful reproduction of the manual test process we used to do in the past. In my mind, the main purpose of the test is to ensure that going through a several rounds of upgrades doesn't leave the cluster in an unstable state and that the operator can start up and be functional.

I didn't try to integrate this into the E2E framework for several reasons:
- It'll require a fairly significant refactoring of the existing framework -- which feels like a lot of unnecessary work to accommodate a single, very specific test.
- It wouldn't be practical or meaningful to run E2E tests for every upgrade step anyway. Hopefully those releases were already extensively tested at release time. Some of the existing E2E tests probably even won't work with older operator versions and will require introducing even more complexity to manage a compatibility matrix.
- As iterated above, the expectation is that the operator itself is well tested already by the existing E2E tests. I can't imagine there being a lot of tests that we want to run specifically within an upgrade context. As long as the operator is up and running after an upgrade, I think it's fair to assume that the rest of the functionality is going to work as expected.

hack/upgrade-test-harness/fixture/eck.go
pebrc: No not very strong at all.

Issue #3196, opened 2020-06-05
EKS E2E tests job is broken


thbkrkr:
Since May 18, the job `cloud-on-k8s-e2e-tests-eks/` fails when creating the k8s cluster.

```
02:28:54  2020/05/19 00:28:45 Creating cluster ...
03:06:18  Error: timed out (after 25m0s) waiting for at least 3 nodes to join the cluster and become ready in "ng-1"
03:06:18  Error: exit status 1
```

pebrc:
We need a Slack notification for this job

thbkrkr:
The output is purposely hidden to avoid any leakage of sensitive data, so we are quite blind here.

I executed the cluster creation with a more detailed log level (`eksctl create cluster -v 4 -f eks.yml`):

```sh
...
2020-06-05T15:17:14+02:00 [▶]  start waiting for CloudFormation stack "eksctl-test-eks-thb-cluster"
2020-06-05T15:17:14+02:00 [▶]  waiting for CloudFormation stack "eksctl-test-eks-thb-cluster"
2020-06-05T15:17:31+02:00 [▶]  waiting for CloudFormation stack "eksctl-test-eks-thb-cluster"
2020-06-05T15:17:51+02:00 [▶]  waiting for CloudFormation stack "eksctl-test-eks-thb-cluster"
2020-06-05T15:18:10+02:00 [▶]  waiting for CloudFormation stack "eksctl-test-eks-thb-cluster"
2020-06-05T15:18:11+02:00 [✖]  unexpected status "ROLLBACK_IN_PROGRESS" while waiting for CloudFormation stack "eksctl-test-eks-thb-cluster"
...
2020-06-05T15:18:11+02:00 [✖]  AWS::EC2::SubnetRouteTableAssociation/RouteTableAssociationPublicEUWEST2A: CREATE_FAILED – "Resource creation cancelled"
2020-06-05T15:18:11+02:00 [✖]  AWS::EC2::NatGateway/NATGateway: CREATE_FAILED – "Resource creation cancelled"
2020-06-05T15:18:11+02:00 [✖]  AWS::EC2::SubnetRouteTableAssociation/RouteTableAssociationPrivateEUWEST2A: CREATE_FAILED – "Resource creation cancelled"
2020-06-05T15:18:11+02:00 [✖]  AWS::EC2::SubnetRouteTableAssociation/RouteTableAssociationPublicEUWEST2C: CREATE_FAILED – "Resource creation cancelled"
[...]

PR #3314, opened 2020-06-24
Use the NodeAMI variable when creating an EKS cluster

thbkrkr:
This commit properly uses the configurable parameter `NodeAMI` when creating an EKS cluster instead of using statically `static`.
This does not change the current behaviour as the `NodeAMI` was already set to `static`.

https://github.com/elastic/cloud-on-k8s/blob/6c3f5eb9d50999b51ee63d7f9c19b8297ce80d42/hack/deployer/runner/settings.go#L80-L81

https://github.com/elastic/cloud-on-k8s/blob/24335cac199874cea86872e5a18a7830cb3ee758/hack/deployer/config/plans.yml#L82-L92

### DIFF
--- a/hack/deployer/runner/eks.go
+++ b/hack/deployer/runner/eks.go
@@ -35,7 +35,7 @@ nodeGroups:
   - name: ng-1
     instanceType: {{.MachineType}}
     desiredCapacity: {{.NodeCount}}
-    ami: static
+    ami: {{.NodeAMI}}
     iam:
       instanceProfileARN: {{.InstanceProfileARN}}
       instanceRoleARN: {{.InstanceRoleARN}}


hack/deployer/runner/eks.go
```diff
@@ -35,7 +35,7 @@ nodeGroups:
   - name: ng-1
     instanceType: {{.MachineType}}
     desiredCapacity: {{.NodeCount}}
-    ami: static
+    ami: {{.NodeAMI}}
```
barkbay, line 38: Maybe I'm missing something but what is the benefit of keeping the `nodeAMI` field if it is always set to static ?
In which situations should we set an other value ?

thbkrkr: 2 reasons: 
* just for correctness
* using `auto` solves #3196 but I'm trying to make it work with `static`

The other option, to be logical, is to delete the value that appears in plan.yml. As it is already in place, I thought that we could keep it.

Issue #3312, opened 2020-06-24
The operator inserts xpack.encryptedSavedObjects.encryptionKey config key for all versions of Kibana


david-kow:
The operator inserts empty `xpack.encryptedSavedObjects.encryptionKey` key into Kibana config for versions <7.6 while this setting is only supported in 7.6 and up. This results in Kibana not being able to start:

```
{ Error: "xpack.encryptedSavedObjects.encryptionKey" setting was not applied. Check for spelling errors and ensure that expected plugins are installed.
    at KbnServer._default (/usr/share/kibana/src/legacy/server/config/complete.js:90:17) code: 'InvalidConfig', processExitCode: 64 }
```

```
$ k get kibana test-ab-cfg-4p65
NAME               HEALTH   NODES   VERSION   AGE
test-ab-cfg-4p65   red              7.3.2     2m19s

$ k tree kibana test-ab-cfg-4p65
NAMESPACE    NAME                                              READY  REASON              AGE
e2e-mercury  Kibana/test-ab-cfg-4p65                           -                          2m10s
e2e-mercury  ├─Deployment/test-ab-cfg-4p65-kb                  -                          2m8s
e2e-mercury  │ └─ReplicaSet/test-ab-cfg-4p65-kb-8697f6b976     -                          2m8s
e2e-mercury  │   └─Pod/test-ab-cfg-4p65-kb-8697f6b976-48mj5    False  ContainersNotReady  2m7s
e2e-mercury  ├─Secret/test-ab-cfg-4p65-kb-config               -                          2m8s
...


$ k view-secret test-ab-cfg-4p65-kb-config kibana.yml
...
xpack:
  encryptedSavedObjects:
    encryptionKey: ""
...
```

PR #3315, opened 2020-06-24
Fix kibana saved object encryption keys pre-7.6.0

anyasabo:
Fix https://github.com/elastic/cloud-on-k8s/issues/3312

Removing just the value does not work as Kibana still chokes on the parent:

>{"type":"log","@timestamp":"2020-06-24T16:32:13Z","tags":["fatal","root"],"pid":1,"message":"{ Error: \"xpack.encryptedSavedObjects\" setting was not applied. Check for spelling errors and ensure that expected plugins are installed.\n    at KbnServer.exports.default (/usr/share/kibana/src/server/config/complete.js:88:17) code: 'InvalidConfig', processExitCode: 64 }"}

> FATAL  Error: "xpack.encryptedSavedObjects" setting was not applied. Check for spelling errors and ensure that expected plugins are installed.

Checking with the beats folks if they would be open to us adding an `omitempty` like tag for go-ucfg so we can avoid doing this in the future. From what I could see there was not a way to do something like that when unpacking currently, thus this method.

### DIFF
--- a/pkg/controller/kibana/config_settings.go
+++ b/pkg/controller/kibana/config_settings.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"path"
 
+	ucfg "github.com/elastic/go-ucfg"
 	"github.com/pkg/errors"
 	"go.elastic.co/apm"
 	corev1 "k8s.io/api/core/v1"
@@ -44,6 +45,7 @@ const (
 	XpackLicenseManagementUIEnabled                = "xpack.license_management.ui.enabled" // >= 7.6
 	XpackSecurityEncryptionKey                     = "xpack.security.encryptionKey"
 	XpackReportingEncryptionKey                    = "xpack.reporting.encryptionKey"
+	XpackEncryptedSavedObjects                     = "xpack.encryptedSavedObjects"
 	XpackEncryptedSavedObjectsEncryptionKey        = "xpack.encryptedSavedObjects.encryptionKey"
 
 	ElasticsearchSslCertificateAuthorities = "elasticsearch.ssl.certificateAuthorities"
@@ -71,6 +73,12 @@ func NewConfigSettings(ctx context.Context, client k8s.Client, kb kbv1.Kibana, v
 	defer span.End()
 
 	reusableSettings, err := getOrCreateReusableSettings(client, kb)
+	if err != nil {
+		return CanonicalConfig{}, err
+	}
+
+	// hack to support pre-7.6.0 Kibana configs as it errors out with unsupported keys, ideally we would not unpack empty values and could skip this
+	filteredReusableSettings, err := filterConfigSettings(kb, reusableSettings)
 	if err != nil {
 		return CanonicalConfig{}, err
 	}
@@ -108,7 +116,7 @@ func NewConfigSettings(ctx context.Context, client k8s.Client, kb kbv1.Kibana, v
 
 	// merge the configuration with userSettings last so they take precedence
 	err = cfg.MergeWith(
-		reusableSettings,
+		filteredReusableSettings,
 		versionSpecificCfg,
 		kibanaTLSCfg,
 		settings.MustCanonicalConfig(elasticsearchTLSSettings(kb)),
@@ -125,6 +133,19 @@ func NewConfigSettings(ctx context.Context, client k8s.Client, kb kbv1.Kibana, v
 	}
 
 	return CanonicalConfig{cfg}, nil
+}
+
+// Some previously-unsupported keys cause Kibana to error out even if the values are empty. ucfg cannot ignore fields easily so this is necessary to
+// support older versions
+func filterConfigSettings(kb kbv1.Kibana, cfg *settings.CanonicalConfig) (*settings.CanonicalConfig, error) {
+	ver, err := version.Parse(kb.Spec.Version)
+	if err != nil {
+		return cfg, err
+	}
+	if !ver.IsSameOrAfter(version.From(7, 6, 0)) {
+		_, err = (*ucfg.Config)(cfg).Remove(XpackEncryptedSavedObjects, -1, settings.Options...)
+	}
+	return cfg, err
 }
 
 // VersionDefaults generates any version specific settings that should exist by default.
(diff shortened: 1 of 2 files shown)


pkg/controller/kibana/config_settings.go
```diff
@@ -127,6 +135,19 @@ func NewConfigSettings(ctx context.Context, client k8s.Client, kb kbv1.Kibana, v
...
 
+// Some previously-unsupported keys cause Kibana to error out even if the values are empty. ucfg cannot ignore fields easily so this is necessary to
+// support older versions
+func filterConfigSettings(kb kbv1.Kibana, cfg *settings.CanonicalConfig) (*settings.CanonicalConfig, error) {
+	ver, err := version.Parse(kb.Spec.Version)
+	if err != nil {
+		return cfg, err
+	}
+	if !ver.IsSameOrAfter(version.From(7, 6, 0)) {
+		_, err = (*ucfg.Config)(cfg).Remove(XpackEncryptedSavedObjects, -1, settings.Options...)
```
pebrc, line 146: ucfg supports Initialisers and with a separate struct for > 7.6.0 we could avoid the type conversion and the angling for values here but I am not sure the result is any better:

```diff
diff --git a/pkg/controller/kibana/config_settings.go b/pkg/controller/kibana/config_settings.go
index 5f14f3f4..6075691a 100644
--- a/pkg/controller/kibana/config_settings.go
+++ b/pkg/controller/kibana/config_settings.go
@@ -160,11 +160,27 @@ func VersionDefaults(kb *kbv1.Kibana, v version.Version) *settings.CanonicalConf
 
 // reusableSettings captures secrets settings in the Kibana configuration that we want to reuse.
 type reusableSettings struct {
+	EncryptionKey string `config:"xpack.security.encryptionKey"`
+	ReportingKey  string `config:"xpack.reporting.encryptionKey"`
+}
+
+func (r *reusableSettings) InitDefaults() {
+	r.EncryptionKey = string(common.RandomBytes(64))
+	r.ReportingKey = string(common.RandomBytes(64))
+}
+
+type reusableSettingsV76 struct {
 	EncryptionKey   string `config:"xpack.security.encryptionKey"`
 	ReportingKey    string `config:"xpack.reporting.encryptionKey"`
 	SavedObjectsKey string `config:"xpack.encryptedSavedObjects.encryptionKey"`
 }
 
+func (r *reusableSettingsV76) InitDefaults() {
+	r.EncryptionKey = string(common.RandomBytes(64))
+	r.ReportingKey = string(common.RandomBytes(64))
+	r.SavedObjectsKey = string(common.RandomBytes(64))
+}
+
 // getExistingConfig retrieves the canonical config for a given Kibana, if one exists
 func getExistingConfig(client k8s.Client, kb kbv1.Kibana) (*settings.CanonicalConfig, error) {
 	var secret corev1.Secret
@@ -197,18 +213,8 @@ func getOrCreateReusableSettings(c k8s.Client, kb kbv1.Kibana) (*settings.Canoni
 	if err != nil {
 		return nil, err
 	}
-
-	var r reusableSettings
 	if cfg == nil {
-		r = reusableSettings{}
-	} else if err := cfg.Unpack(&r); err != nil {
-		return nil, err
-	}
-	if len(r.EncryptionKey) == 0 {
-		r.EncryptionKey = string(common.RandomBytes(64))
-	}
-	if len(r.ReportingKey) == 0 {
-		r.ReportingKey = string(common.RandomBytes(64))
+		cfg = settings.NewCanonicalConfig()
 	}
 
 	kbVer, err := version.Parse(kb.Spec.Version)
@@ -216,8 +222,16 @@ func getOrCreateReusableSettings(c k8s.Client, kb kbv1.Kibana) (*settings.Canoni
 		return nil, err
 	}
 	// xpack.encryptedSavedObjects.encryptionKey was only added in 7.6.0 and earlier versions error out
-	if len(r.SavedObjectsKey) == 0 && kbVer.IsSameOrAfter(version.From(7, 6, 0)) {
-		r.SavedObjectsKey = string(common.RandomBytes(64))
+	if kbVer.IsSameOrAfter(version.From(7, 6, 0)) {
+		var r reusableSettingsV76
+		if err := cfg.Unpack(&r); err != nil {
+			return nil, errors.Wrap(err, "on unpack")
+		}
+		return settings.MustCanonicalConfig(r), nil
+	}
+	var r reusableSettings
+	if err := cfg.Unpack(&r); err != nil {
+		return nil, errors.Wrap(err, "on unpack")
 	}
 	return settings.MustCanonicalConfig(r), nil
 }
diff --git a/pkg/controller/kibana/config_settings_test.go b/pkg/controller/kibana/config_settings_test.go
index 33f4e970..d0f4fb84 100644
--- a/pkg/controller/kibana/config_settings_test.go
+++ b/pkg/controller/kibana/config_settings_test.go
@@ -66,6 +66,12 @@ func Test_reuseOrGenerateSecrets(t *testing.T) {
 	kb75 := mkKibana()
 	kb75.Spec.Version = "7.5.0"
 
+	type configVerifier struct {
+		EncryptionKey   string `config:"xpack.security.encryptionKey"`
+		ReportingKey    string `config:"xpack.reporting.encryptionKey"`
+		SavedObjectsKey string `config:"xpack.encryptedSavedObjects.encryptionKey"`
+	}
+
 	tests := []struct {
 		name      string
 		args      args
@@ -109,8 +115,8 @@ func Test_reuseOrGenerateSecrets(t *testing.T) {
 			},
 			assertion: func(t *testing.T, got *settings.CanonicalConfig, err error) {
 				// Unpack the configuration to check that some default reusable settings have been generated
-				var r reusableSettings
-				assert.NoError(t, got.Unpack(&r))
+				var r configVerifier
+				assert.NoError(t, got.Unpack(&r), "on unpack")
 				assert.Equal(t, len(r.EncryptionKey), 64) // key length should be 64
 				assert.Equal(t, len(r.ReportingKey), 64)
 				assert.Equal(t, len(r.SavedObjectsKey), 64)
@@ -132,11 +138,11 @@ func Test_reuseOrGenerateSecrets(t *testing.T) {
 			},
 			assertion: func(t *testing.T, got *settings.CanonicalConfig, err error) {
 				// Unpack the configuration to check that some default reusable settings have been generated
-				var r reusableSettings
+				var r configVerifier
 				assert.NoError(t, got.Unpack(&r))
 				assert.Equal(t, len(r.EncryptionKey), 64) // key length should be 64
 				assert.Equal(t, len(r.ReportingKey), 64)
-				assert.Equal(t, len(r.SavedObjectsKey), 0) // is only introduced in 7.6.0
+				assert.Equal(t, len(r.SavedObjectsKey), 0) // only introduced in 7.6.0
 			},
 		},
 	}
```

anyasabo: Oh nice, I missed `InitDefaults`. I'm not sure either is really that much clearer than the other, so unless someone else has a different opinion I'll probably merge it as is.

PR #3316, opened 2020-06-24
Use eksctl AMIs auto resolver

### LINKED ISSUE #3196: EKS E2E tests job is broken

thbkrkr:
The AMIs static resolver still uses an old AMI that uses an old version of
the AWS CLI that doesnet have the `aws eks get-token` required since eksctl
`0.18.0`.

Resolves #3196.

### DIFF
--- a/hack/deployer/config/plans.yml
+++ b/hack/deployer/config/plans.yml
@@ -89,7 +89,7 @@ plans:
   eks:
     region: eu-central-1
     nodeCount: 3
-    nodeAMI: static
+    nodeAMI: auto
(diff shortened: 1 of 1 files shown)


thbkrkr:
> my only concern is that I am not sure if the user has the right privileges to list AMIs. Have we tested this in CI?

Good point. I just did it and it seems ok.

PR #3317, opened 2020-06-24
[1.2] Fix kibana saved object encryption keys pre-7.6.0 (#3315)

anyasabo:
 - Fix kibana saved object encryption keys pre-7.6.0 (#3315)


anyasabo:
This was my first time using https://github.com/sqren/backport to generate it, it went okay

Issue #3290, opened 2020-06-18
Kubebuilder CRD pattern annotation does not work


david-kow:
I can create ES resource with nodeset name that doesn't comply with the pattern we expect. We catch this through our own validation though.

naemono:
Is this still valid?  The referenced issue of #3325 mentions having commas in names, but the below manifest is throwing validation errors, and I've ensured that the validating webhooks were removed before attempting this.

Manifest
```
apiVersion: beat.k8s.elastic.co/v1beta1
kind: Beat
metadata:
  name: heartbeat,test
spec:
  type: heartbeat
  version: 8.2.3
  elasticsearchRef:
    name: elasticsearch,test
  config:
    heartbeat.monitors:
    - type: tcp
      schedule: '@every 5s'
      hosts: ["elasticsearch-es-http.default.svc:9200"]
    - type: tcp
      schedule: '@every 5s'
      hosts: ["kibana-kb-http.default.svc:5601"]
  deployment:
    replicas: 1
    podTemplate:
      spec:
        securityContext:
          runAsUser: 0
---
apiVersion: elasticsearch.k8s.elastic.co/v1
kind: Elasticsearch
metadata:
  name: elasticsearch,test
spec:
  version: 8.2.3
  nodeSets:
  - name: default
    count: 3
    config:
      node.store.allow_mmap: false
---
apiVersion: elasticsearch.k8s.elastic.co/v1
kind: Elasticsearch
metadata:
  name: monitoring,test
spec:
  version: 8.2.3
  nodeSets:
  - name: default
    count: 3
    config:
      node.store.allow_mmap: false
---
apiVersion: kibana.k8s.elastic.co/v1
kind: Kibana
metadata:
  name: kibana,test
spec:
  version: 8.2.3
  count: 1
  elasticsearchRef:
    name: monitoring,test
  monitoring:
    metrics:
      elasticsearchRefs:
      - name: elasticsearch,test
```

Results
```
❯ kc apply -f my-invalid-manifest.yaml -n default
[...]

PR #3325, opened 2020-06-26
Fix Beat E2E tests issues

david-kow:
This PR fixes three Beats E2E tests issues:

1. SCC for OCP has capabilities added allowing for Auditbeat/Packetbeat/Journalbeat to run. This is a bit different than our non-OCP approach where we currently have a separate PSP for each. I think this is fine.

```
[2020-06-26T01:55:35.054Z] {"reason":"FailedCreate","message":"Error creating: pods \"test-ab-cfg-bzhr-beat-auditbeat-\" is forbidden: unable to validate against any security context constraint: [provider restricted: .spec.securityContext.hostNetwork: Invalid value: true: Host network is not allowed to be used provider restricted: .spec.securityContext.hostPID: Invalid value: true: Host PID is not allowed to be used spec.volumes[0]: Invalid value: \"hostPath\": hostPath volumes are not allowed to be used spec.volumes[1]: Invalid value: \"hostPath\": hostPath volumes are not allowed to be used spec.volumes[4]: Invalid value: \"hostPath\": hostPath volumes are not allowed to be used spec.volumes[6]: Invalid value: \"hostPath\": hostPath volumes are not allowed to be used spec.volumes[7]: Invalid value: \"hostPath\": hostPath volumes are not allowed to be used spec.volumes[8]: Invalid value: \"hostPath\": hostPath volumes are not allowed to be used spec.volumes[9]: Invalid value: \"hostPath\": hostPath volumes are not allowed to be used spec.containers[0].securityContext.securityContext.runAsUser: Invalid value: 0: must be in the ranges: [1000550000, 1000559999] capabilities.add: Invalid value: \"AUDIT_CONTROL\": capability may not be added capabilities.add: Invalid value: \"AUDIT_READ\": capability may not be added capabilities.add: Invalid value: \"AUDIT_WRITE\": capability may not be added spec.containers[0].securityContext.hostNetwork: Invalid value: true: Host network is not allowed to be used spec.containers[0].securityContext.hostPID: Invalid value: true: Host PID is not allowed to be used provider beat: .spec.securityContext.hostPID: Invalid value: true: Host PID is not allowed to be used capabilities.add: Invalid value: \"AUDIT_READ\": capability may not be added capabilities.add: Invalid value: \"AUDIT_WRITE\": capability may not be added capabilities.add: Invalid value: \"AUDIT_CONTROL\": capability may not be added spec.containers[0].securityContext.hostPID: Invalid value: true: Host PID is not allowed to be used]","kind":"DaemonSet","name":"test-ab-cfg-bzhr-beat-auditbeat","namespace":"e2e-ftrer-mercury"}
```

2. Add `automountServiceAccountToken` to all Beat configs. Lower version Beats [depended](https://github.com/elastic/beats/pull/6146) on `/var/run/secrets/kubernetes.io/serviceaccount/namespace` to exists in the pod.

```
Exiting: error initializing processors: Unable to get in cluster configuration: open /var/run/secrets/kubernetes.io/serviceaccount/namespace: no such file or directory
```

3. Beats need more permissions than just `kibana_admin` or `kibana_role` to do the dashboard setup below certain Beat version.

```
{"type":"response","@timestamp":"2020-06-26T07:22:01Z","tags":["api"],"pid":6,"method":"post","statusCode":403,"req":{"url":"/api/kibana/dashboards/import?exclude=index-pattern&force=true","method":"post","headers":{"host":"test-ab-cfg-wftc-kb-http.e2e-mercury.svc:5601","user-agent":"Go-http-client/1.1","content-length":"37053","accept":"application/json","content-type":"application/json","kbn-xsrf":"1","accept-encoding":"gzip"},"remoteAddress":"10.0.88.2","userAgent":"10.0.88.2"},"res":{"statusCode":403,"responseTime":46,"contentLength":9},"message":"POST /api/kibana/dashboards/import?exclude=index-pattern&force=true 403 46ms - 9.0B"}
{"type":"log","@timestamp":"2020-06-26T07:22:01Z","tags":["debug","http","server","Kibana","cookie-session-storage"],"pid":6,"message":"Error: Unauthorized"}
```

### DIFF
--- a/config/e2e/filebeat.yaml
+++ b/config/e2e/filebeat.yaml
@@ -248,16 +248,16 @@ metadata:
 allowHostDirVolumePlugin: true
 allowHostIPC: false
 allowHostNetwork: true
-allowHostPID: true
+allowHostPID: true # required for Auditbeat
 allowHostPorts: false
 allowPrivilegeEscalation: true
 allowPrivilegedContainer: true
 allowedCapabilities:
-  - AUDIT_READ
+  - AUDIT_READ     # required for Auditbeat
   - AUDIT_WRITE
   - AUDIT_CONTROL
-  - NET_ADMIN
-  - KILL
+  - NET_ADMIN      # required for Packetbeat
+  - KILL           # required for Journalbeat
   - CHOWN
   - FSETID
   - FOWNER
--- a/config/e2e/filebeat.yaml
+++ b/config/e2e/filebeat.yaml
@@ -248,11 +248,25 @@ metadata:
 allowHostDirVolumePlugin: true
 allowHostIPC: false
 allowHostNetwork: true
-allowHostPID: false
+allowHostPID: true
 allowHostPorts: false
 allowPrivilegeEscalation: true
 allowPrivilegedContainer: true
-allowedCapabilities: []
+allowedCapabilities:
+  - AUDIT_READ
+  - AUDIT_WRITE
+  - AUDIT_CONTROL
+  - NET_ADMIN'
+  - KILL
+  - CHOWN
+  - FSETID
+  - FOWNER
+  - SETGID
+  - SETUID
+  - SETFCAP
+  - SETPCAP
+  - AUDIT_WRITE
+  - NET_BIND_SERVICE
 defaultAddCapabilities: []
 fsGroup:
   type: RunAsAny
--- a/config/e2e/filebeat.yaml
+++ b/config/e2e/filebeat.yaml
@@ -256,7 +256,7 @@ allowedCapabilities:
   - AUDIT_READ
   - AUDIT_WRITE
   - AUDIT_CONTROL
-  - NET_ADMIN'
+  - NET_ADMIN
   - KILL
   - CHOWN
   - FSETID
--- a/pkg/controller/association/controller/beat_kibana.go
+++ b/pkg/controller/association/controller/beat_kibana.go
@@ -82,7 +82,7 @@ func getBeatKibanaRoles(associated commonv1.Associated) (string, error) {
 	}
 
 	// Roles for supported Beats are based on:
-	// https://www.elastic.co/guide/en/beats/filebeat/7.6/feature-roles.html#privileges-to-setup-beats
+	// https://www.elastic.co/guide/en/beats/filebeat/current/feature-roles.html#privileges-to-setup-beats
 	// Docs are the same for all Beats. For a specific version docs change "current" to major.minor, eg:
 	// https://www.elastic.co/guide/en/beats/filebeat/7.1/feature-roles.html#privileges-to-setup-beats
 	switch {
--- a/pkg/controller/association/controller/beat_kibana.go
+++ b/pkg/controller/association/controller/beat_kibana.go
@@ -5,12 +5,16 @@
 package controller
 
 import (
+	"fmt"
+	"strings"
+
 	beatv1beta1 "github.com/elastic/cloud-on-k8s/pkg/apis/beat/v1beta1"
 	commonv1 "github.com/elastic/cloud-on-k8s/pkg/apis/common/v1"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/association"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common/operator"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common/version"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common/watches"
+	esuser "github.com/elastic/cloud-on-k8s/pkg/controller/elasticsearch/user"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/kibana"
 	"github.com/elastic/cloud-on-k8s/pkg/utils/k8s"
 	"github.com/elastic/cloud-on-k8s/pkg/utils/rbac"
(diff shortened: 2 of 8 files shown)


config/e2e/filebeat.yaml
```diff
@@ -248,11 +248,25 @@ metadata:
 allowHostDirVolumePlugin: true
 allowHostIPC: false
 allowHostNetwork: true
-allowHostPID: false
+allowHostPID: true
 allowHostPorts: false
 allowPrivilegeEscalation: true
 allowPrivilegedContainer: true
-allowedCapabilities: []
+allowedCapabilities:
```
sebgl, line 255: Can we add a comment to explain why these are necessary?

pkg/controller/association/controller/beat_kibana.go
```diff
@@ -68,12 +72,40 @@ func getBeatKibanaRoles(associated commonv1.Associated) (string, error) {
 		)
 	}
 
+	if strings.Contains(beat.Spec.Type, ",") {
+		return "", fmt.Errorf("beat type %s should not contain a comma", beat.Spec.Type)
```
sebgl, line 76: If it's a general thing about the Beats CRD it should probably live outside the association controller, and be also enforced by the Beat controller & webhook?

pkg/controller/association/controller/beat_kibana.go
```diff
@@ -68,12 +72,40 @@ func getBeatKibanaRoles(associated commonv1.Associated) (string, error) {
...
 	v, err := version.Parse(beat.Spec.Version)
 	if err != nil {
 		return "", err
 	}
-	if v.IsSameOrAfter(version.From(7, 5, 0)) {
-		return "kibana_admin", nil
+
+	// Roles for supported Beats are based on:
+	// https://www.elastic.co/guide/en/beats/filebeat/7.6/feature-roles.html#privileges-to-setup-beats
+	// Docs are the same for all Beats. For a specific version docs change "current" to major.minor, eg:
```
anyasabo, line 86: The example above does not contain current, so maybe just change this text to "change the major.minor version in the URL"?

test/e2e/beat/config.go
```diff
@@ -196,6 +196,7 @@ processors:
   hostPID: true  # Required by auditd module
   dnsPolicy: ClusterFirstWithHostNet
   hostNetwork: true
+  automountServiceAccountToken: true
```
anyasabo, line 199: This is only necessary on some versions from your description, correct? Is it worth adding a comment here and updating the recipes/docs as well with this info? We call out other OpenShift specific configs

config/e2e/filebeat.yaml
anyasabo: Ditto for allowHostPID because it's not clear to me why it needs that, and also ditto for do we need to update the openshift docs to include these changes (my guess is yes)

pkg/controller/association/controller/beat_kibana.go
david-kow: It's already forbidden by the Type regex requirements. One here is just double checking. Do you think it would be worth to check it in a validation too?

pkg/controller/association/controller/beat_kibana.go
david-kow: Replaced with current as this is what we have in `beat_es.go`.

config/e2e/filebeat.yaml
```diff
@@ -248,11 +248,25 @@ metadata:
...
+  - AUDIT_READ
+  - AUDIT_WRITE
+  - AUDIT_CONTROL
+  - NET_ADMIN'
```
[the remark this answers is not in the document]
david-kow: I remember fixing that...

pkg/controller/association/controller/beat_kibana.go
anyasabo: If it's in the OpenAPI validation it should never get through, so I think we can omit the check

test/e2e/beat/config.go
david-kow: @anyasabo, @sebgl: This is not OpenShift specific actually, only Beat version specific. I've added a comment, but I'd rather not add too much in the docs/samples for couple of reasons:
- if we do it for one setting, why we don't do it for all of them? it would a bit of work
- our docs start becoming Beats on k8s docs instead of just ECK docs
- we use those configs to test across all versions/environments, so we might end up with if statements in the comments :) we already turn off some things for old kind version, because it doesn't work there

In general I'd like to think about us testing Beats as a proof we can deploy them correctly and they work to some degree. I believe high fidelity documentation about what and why could be a rabbit hole for us and should live in Beat docs. WDYT?

pkg/controller/association/controller/beat_kibana.go
david-kow: We seem to have an [issue](https://github.com/elastic/cloud-on-k8s/issues/3290) with that. But even without it, I'd like to keep another check close to the code just to make sure we are safe in case it'll get copied over somewhere not behind that regex check.

config/e2e/filebeat.yaml
```diff
@@ -248,11 +248,25 @@ metadata:
...
+  - CHOWN
+  - FSETID
+  - FOWNER
+  - SETGID
+  - SETUID
+  - SETFCAP
+  - SETPCAP
+  - AUDIT_WRITE
+  - NET_BIND_SERVICE
 defaultAddCapabilities: []
```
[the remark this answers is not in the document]
david-kow: It's actually both. So far it was identical, hence one copy. I'd like to merge this PR asap to unblock tests, so how about a separate PR to fix this file?

david-kow:
jenkins test this please

config/e2e/filebeat.yaml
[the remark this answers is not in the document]
pebrc: Sure if you prefer, can you create an issue?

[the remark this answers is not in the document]
david-kow: [Yes](https://github.com/elastic/cloud-on-k8s/issues/3353).

david-kow:
@barkbay It shouldn't happen. Logs from the Pod would be the most helpful I think.

Crashing is due to two factors, but both should be temporary:
- Kibana not being up
- Beat user for Kibana not created in ES yet

Maybe create a separate issue for flaky test?

pkg/controller/association/controller/beat_kibana.go
anyasabo: Ah if the OpenAPI validation doesn't work we should still just check for it in the validation code we run at the beginning of the reconcile/webhook IMO rather than having type validation code scattered about

test/e2e/beat/config.go
anyasabo: Ah gotcha my reading comprehension is poor. I think it's a blurry line for when it is useful to have beats documentation in our repo, but am good with your decision here 👍

Issue #3326, opened 2020-06-26
Beat webhook is not registered in webhook server


pebrc:
Running docker.elastic.co/eck/eck-operator:1.2.0-bc2

```
curl -k -i https://localhost:9090/validate-beat-k8s-elastic-co-v1beta1-beat
HTTP/2 404 
content-type: text/plain; charset=utf-8
x-content-type-options: nosniff
content-length: 19
date: Fri, 26 Jun 2020 12:49:04 GMT

404 page not found
```

vs. :
```
 curl -k -i https://localhost:9090/validate-enterprisesearch-k8s-elastic-co-v1beta1-enterprisesearch
HTTP/2 200 
content-type: text/plain; charset=utf-8
content-length: 128
date: Fri, 26 Jun 2020 12:58:22 GMT

{"response":{"uid":"","allowed":false,"status":{"metadata":{},"message":"contentType=, expected application/json","code":400}}}
```

Also it seems we are missing an automated test here.

PR #3327, opened 2020-06-26
Fix beat webhook setup

david-kow:
Fixes https://github.com/elastic/cloud-on-k8s/issues/3326 and adds E2E test.

```
$ curl -k -i https://localhost:9090/validate-beat-k8s-elastic-co-v1beta1-beat
HTTP/2 200
content-type: text/plain; charset=utf-8
content-length: 128
date: Fri, 26 Jun 2020 13:53:31 GMT

{"response":{"uid":"","allowed":false,"status":{"metadata":{},"message":"contentType=, expected application/json","code":400}}}
```

### DIFF
--- a/test/e2e/beat/webhook_test.go
+++ b/test/e2e/beat/webhook_test.go
@@ -1,0 +1,36 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License;
+// you may not use this file except in compliance with the Elastic License.
+
+package beat
+
+import (
+	"testing"
+
+	beatv1beta1 "github.com/elastic/cloud-on-k8s/pkg/apis/beat/v1beta1"
+	"github.com/elastic/cloud-on-k8s/test/e2e/test"
+	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestWebhook(t *testing.T) {
+	beat := beatv1beta1.Beat{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "test-webhook",
+			Namespace: test.Ctx().ManagedNamespace(0),
+		},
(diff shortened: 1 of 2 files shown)


test/e2e/beat/webhook_test.go
```diff
@@ -0,0 +1,36 @@
...
+	"testing"
+
+	beatv1beta1 "github.com/elastic/cloud-on-k8s/pkg/apis/beat/v1beta1"
+	"github.com/elastic/cloud-on-k8s/test/e2e/test"
+	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestWebhook(t *testing.T) {
+	beat := beatv1beta1.Beat{
```
pebrc, line 17: Would it make sense to follow the established pattern here: https://github.com/elastic/cloud-on-k8s/blob/ccaa19f04956e52ec543f46806640822c4440e57/pkg/apis/enterprisesearch/v1beta1/webhook_test.go#L21

pebrc: I realise after talking to @david-kow that this is not actually testing correct deployment of the webhook. So this can be a different PR.

test/e2e/beat/webhook_test.go
```diff
@@ -0,0 +1,36 @@
...
+		Spec: beatv1beta1.BeatSpec{
+			Type:    "filebeat",
+			Version: "7.8.0",
+		},
+	}
+
+	err := test.NewK8sClientOrFatal().Client.Create(&beat)
+
+	require.Error(t, err)
+	require.Contains(
```
pebrc, line 31:
```golang
	require.True(t, apierrors.IsInvalid(err))
```

david-kow: @pebrc I've actually blindly added this, but it doesn't pass. I think `IsInvalid` will only work for openapi validations, not webhook validations.

pebrc: My bad :-(

PR #3336, opened 2020-06-29
Allow init containers modifications

### LINKED ISSUE #2306: Allow init container modifications
Currently, if a user wants to change a small aspect of an init container (e.g. [setting custom resource limits](https://discuss.elastic.co/t/init-container-mount-problem-in-eck/212382)), they have to provide the full definition of the init container in the `podTemplate`.

Presumably we do this to give users full control over how init containers are executed. However, there are a couple of UX issues worth considering:
- Most users are probably only interested in minor tweaks such as changing the resource limits of the container
- The way the operator treats `initContainers` of the `podTemplate` is confusing. (Other sections in the `podTemplate` get _merged_ while `initContainers` simply get _replaced_ if the container name matches a built-in container)

barkbay:
This PR allows a user to customize the built-in init containers.

For example in the case of Elasticsearch:

```yaml
  nodeSets:
    - name: my-nodeset
      podTemplate:
        spec:
          initContainers:
          - name: elastic-internal-init-filesystem
            env:
              - name: MY_INIT_FS_ENV
                value: "MY_INIT_FS_ENV_VALUE"
          - name: elastic-internal-init-keystore
            env:
              - name: MY_KEYSTORE_ENV
                value: "MY_KEYSTORE_ENV_VALUE"
            resources:
              limits:
                cpu: 1000m
                memory: 368Mi
              requests:
                cpu: 1000m
                memory: 368Mi
```

Also note that this PR fixes a bug where some env variables are set twice on some Pods, e.g. :

```yaml
  initContainers:
  - command:
    - bash
    - -c
    - /mnt/elastic-internal/scripts/prepare-fs.sh
    env:
    - name: POD_IP
      valueFrom:
        fieldRef:
          apiVersion: v1
          fieldPath: status.podIP
    - name: POD_NAME
      valueFrom:
        fieldRef:
          apiVersion: v1
          fieldPath: metadata.name
    - name: POD_IP
      valueFrom:
        fieldRef:
          apiVersion: v1
          fieldPath: status.podIP
    - name: POD_NAME
      valueFrom:
        fieldRef:
          apiVersion: v1
          fieldPath: metadata.name
```

This triggers a rolling restart of the cluster when upgrading.

Follow-up issue:
- [ ] Update documentation

Fixes #2306

### DIFF
--- a/pkg/controller/common/container/defaulter.go
+++ b/pkg/controller/common/container/defaulter.go
@@ -64,6 +64,10 @@ func (d Defaulter) WithPorts(ports []corev1.ContainerPort) Defaulter {
 			d.base.Ports = append(d.base.Ports, p)
 		}
 	}
+	// order ports by name to ensure stable pod spec comparison
+	sort.SliceStable(d.base.Ports, func(i, j int) bool {
+		return d.base.Ports[i].Name < d.base.Ports[j].Name
+	})
 	return d
 }
 
--- a/pkg/controller/common/container/defaulter.go
+++ b/pkg/controller/common/container/defaulter.go
@@ -1,0 +1,175 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License;
+// you may not use this file except in compliance with the Elastic License.
+
+package container
+
+import (
+	"sort"
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+// Defaulter ensures that values are set if none exists in the base container.
+type Defaulter interface {
+	WithImage(image string) Defaulter
+	WithCommand(command []string) Defaulter
+	WithArgs(args []string) Defaulter
+	WithPorts(ports []corev1.ContainerPort) Defaulter
+	WithEnv(vars []corev1.EnvVar) Defaulter
+	WithResources(resources corev1.ResourceRequirements) Defaulter
+	WithVolumeMounts(volumeMounts []corev1.VolumeMount) Defaulter
+	WithReadinessProbe(readinessProbe *corev1.Probe) Defaulter
+	WithPreStopHook(handler *corev1.Handler) Defaulter
+
+	// From inherits default values from an other container.
+	From(other corev1.Container) Defaulter
+
+	// Container return a copy of the resulting container.
+	Container() corev1.Container
+}
+
+var _ Defaulter = &defaulter{}
+
+type defaulter struct {
+	base *corev1.Container
+}
+
+func (d defaulter) Container() corev1.Container {
+	return *d.base.DeepCopy()
+}
+
+func NewDefaulter(base *corev1.Container) Defaulter {
+	return &defaulter{
+		base: base,
+	}
+}
+
+func (d defaulter) From(other corev1.Container) Defaulter {
+
+	if other.Lifecycle != nil {
+		d.WithPreStopHook(other.Lifecycle.PreStop)
+	}
+
+	return d.
+		WithImage(other.Image).
+		WithCommand(other.Command).
+		WithArgs(other.Args).
+		WithPorts(other.Ports).
+		WithEnv(other.Env).
+		WithResources(other.Resources).
+		WithVolumeMounts(other.VolumeMounts).
+		WithReadinessProbe(other.ReadinessProbe)
+}
+
+func (d defaulter) WithCommand(command []string) Defaulter {
+	if len(d.base.Command) == 0 {
+		d.base.Command = command
+	}
+	return d
+}
+
+func (d defaulter) WithArgs(args []string) Defaulter {
+	if len(d.base.Args) == 0 {
+		d.base.Args = args
+	}
+	return d
+}
+
+func (d defaulter) WithPorts(ports []corev1.ContainerPort) Defaulter {
+	for _, p := range ports {
+		if !d.portExists(p.Name) {
+			d.base.Ports = append(d.base.Ports, p)
+		}
+	}
+	return d
+}
+
+// portExists checks if a port with the given name already exists in the Container.
+func (d defaulter) portExists(name string) bool {
+	for _, p := range d.base.Ports {
+		if p.Name == name {
+			return true
+		}
+	}
+	return false
+}
+
+// WithImage sets up the Container Docker image, unless already provided.
+// The default image will be used unless customImage is not empty.
+func (d defaulter) WithImage(image string) Defaulter {
+	if d.base.Image == "" {
+		d.base.Image = image
+	}
+	return d
+}
+
+func (d defaulter) WithReadinessProbe(readinessProbe *corev1.Probe) Defaulter {
+	if d.base.ReadinessProbe == nil {
+		d.base.ReadinessProbe = readinessProbe
+	}
+	return d
+}
+
+// envExists checks if an env var with the given name already exists in the provided slice.
+func (d defaulter) envExists(name string) bool {
+	for _, v := range d.base.Env {
+		if v.Name == name {
+			return true
+		}
+	}
+	return false
+}
+
+func (d defaulter) WithEnv(vars []corev1.EnvVar) Defaulter {
+	for _, v := range vars {
+		if !d.envExists(v.Name) {
+			d.base.Env = append(d.base.Env, v)
+		}
+	}
+	return d
+}
+
+func (d defaulter) WithResources(resources corev1.ResourceRequirements) Defaulter {
+	// Ensure resources are set
+	if d.base.Resources.Requests == nil && d.base.Resources.Limits == nil {
+		d.base.Resources = resources
+	}
+	return d
+}
+
+// volumeExists checks if a volume mount with the given name already exists in the Container.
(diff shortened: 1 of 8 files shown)


pkg/controller/common/container/defaulter.go
```diff
@@ -0,0 +1,175 @@
...
+package container
+
+import (
+	"sort"
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+// Defaulter ensures that values are set if none exists in the base container.
+type Defaulter interface {
```
sebgl, line 14: is there a use case for this interface vs. just using a public `Defaulter` struct directly with no interface defined?

pkg/controller/common/defaults/pod_template.go
```diff
@@ -90,28 +91,20 @@ func (b *PodTemplateBuilder) WithAnnotations(annotations map[string]string) *Pod
 	return b
 }
 
-// WithDockerImage sets up the Container Docker image, unless already provided.
+// WithImage sets up the Container Docker image, unless already provided.
 // The default image will be used unless customImage is not empty.
 func (b *PodTemplateBuilder) WithDockerImage(customImage string, defaultImage string) *PodTemplateBuilder {
```
sebgl, line 96: comment / function name mismatch

pkg/controller/common/defaults/pod_template.go
```diff
@@ -319,29 +249,17 @@ func (b *PodTemplateBuilder) WithInitContainers(initContainers ...corev1.Contain
...
 	return b
 }
 
 func (b *PodTemplateBuilder) WithArgs(args ...string) *PodTemplateBuilder {
-	if b.Container.Args == nil {
-		b.Container.Args = args
-	}
+	b.containerDefaulter.WithArgs(args)
 	return b
 }
```
sebgl, line 265: Should the `PodTemplateBuilder` just embed a Defaulter to inherit all its functions, so we avoid the extra hop for most functions (one function in PodTemplateBuilder that just calls the corresponding function in Defaulter)?

```
type PodTemplateBuilder struct {
    container.Defaulter
    PodTemplate        corev1.PodTemplateSpec
    containerName      string
}
```

pkg/controller/common/defaults/pod_template_test.go
```diff
@@ -1122,7 +1122,7 @@ func TestPodTemplateBuilder_WithPreStopHook(t *testing.T) {
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			b := NewPodTemplateBuilder(tt.podTemplate, "mycontainer")
-			got := b.WithPreStopHook(tt.preStopHook).Container.Lifecycle
+			got := b.WithPreStopHook(tt.preStopHook).containerDefaulter.Container().Lifecycle
 			if !reflect.DeepEqual(got.PreStop, &tt.wantPreStop) {
 				t.Errorf("PreStop after PodTemplateBuilder.WithPreStopHook() = %v, want %v", got.PreStop, tt.wantPreStop)
 			}
```
sebgl, line 1128: Should most of those unit tests live in `common/defaults/defaulter_test.go` now?

pkg/controller/common/defaults/pod_template.go
```diff
@@ -228,47 +173,22 @@ func (b *PodTemplateBuilder) WithTerminationGracePeriod(period int64) *PodTempla
...
-	}
-	return -1
-}
-
 // WithInitContainerDefaults sets default values for the current init containers.
 //
 // Defaults:
 // - If the init container contains an empty image field, it's inherited from the main container.
 // - VolumeMounts from the main container are added to the init container VolumeMounts, unless they would conflict
 //   with a specified VolumeMount (by having the same VolumeMount.Name or VolumeMount.MountPath)
```
sebgl, line 181: Let's also mention environment vars here.

pkg/controller/common/defaults/pod_template.go
```diff
@@ -290,20 +210,30 @@ func (b *PodTemplateBuilder) findInitContainerByName(name string) int {
 // - Provided init containers are prepended to the existing ones in the template.
 // - If an init container by the same name already exists in the template, the init container in the template
 // takes its place, and the provided init container is discarded.
```
sebgl, line 212: Let's adapt the comment to the new behaviour.

pkg/controller/common/defaults/pod_template.go
```diff
@@ -290,20 +210,30 @@ func (b *PodTemplateBuilder) findInitContainerByName(name string) int {
...
+	initContainers ...corev1.Container,
+) *PodTemplateBuilder {
 	var containers []corev1.Container
 
 	for _, c := range initContainers {
 		if index := b.findInitContainerByName(c.Name); index != -1 {
-			container := b.PodTemplate.Spec.InitContainers[index]
+			userContainer := b.PodTemplate.Spec.InitContainers[index]
 
 			// remove it from the podTemplate:
```
sebgl, line 222: let's maybe remove the extra `:` at the end

pkg/controller/common/container/defaulter.go
```diff
@@ -0,0 +1,175 @@
...
+	return false
+}
+
+func (d defaulter) WithVolumeMounts(volumeMounts []corev1.VolumeMount) Defaulter {
+	for _, v := range volumeMounts {
+		if !d.volumeMountExists(v) {
+			d.base.VolumeMounts = append(d.base.VolumeMounts, v)
+		}
+	}
+	// order volume mounts by name to ensure stable pod spec comparison
```
sebgl, line 157: not from your PR but since you're at it 😄 
we are sorting for stable comparison here, but we don't do it for environment variables or ports for example. I'm wondering if that sort is still necessary, or if we should also do it for other arrays.

pkg/controller/common/defaults/pod_template.go
barkbay: If `container.Defaulter` is embedded then the return type of the "containers related" functions becomes `Defaulter`. While all the builders are expecting a `PodTemplateBuilder` to chain the calls.

For instance it breaks [the following code](https://github.com/elastic/cloud-on-k8s/blob/master/pkg/controller/apmserver/pod.go#L106): 

```golang
	builder := defaults.NewPodTemplateBuilder(
		p.PodTemplate, apmv1.ApmServerContainerName).
		WithResources(DefaultResources). # error here 
		WithDockerImage(p.CustomImageName, container.ImageRepository(container.APMServerImage, p.Version)).
		WithReadinessProbe(readinessProbe(as.Spec.HTTP.TLS.Enabled())).
		WithPorts(ports).
		WithCommand(command).
		WithVolumes(configVolume.Volume(), configSecretVolume.Volume()).
		WithVolumeMounts(configVolume.VolumeMount(), configSecretVolume.VolumeMount()).
		WithEnv(env...)


	if p.keystoreResources != nil {
		dataVolume := keystore.DataVolume(
			strings.ToLower(as.Kind),
			DataVolumePath,
		)
		builder.WithInitContainers(p.keystoreResources.InitContainer).
			WithVolumes(p.keystoreResources.Volume, dataVolume.Volume()).
			WithVolumeMounts(dataVolume.VolumeMount()).
			WithInitContainerDefaults()
	}

	return builder.PodTemplate
```

IMHO a `podTemplateBuilder` should not expose those methods at all, but delegate the container setup to a dedicated builder/defaulter.
E.g. maybe something along those lines :

```golang
	builder := defaults.NewPodTemplateBuilder(p.PodTemplate, apmv1.ApmServerContainerName).
		WithVolumes(...). // Pod specific methods
                WithTerminationGracePeriod(...).
                WithServiceAccount(...)
                WithContainer(
                          NewContainerDefaulter("apm-server").
		          WithResources(DefaultResources). // Container specific methods
		          WithDockerImage(.....).
		          WithReadinessProbe(readinessProbe(as.Spec.HTTP.TLS.Enabled())).
		          WithPorts(ports).
		          WithCommand(command).
		          WithVolumeMounts(configVolume.VolumeMount(), configSecretVolume.VolumeMount()).
		          WithEnv(env...).
                          Container()
                )

```

I can create a separate issue to track this refactoring effort.

pkg/controller/common/container/defaulter.go
barkbay: The interface helped me to drive the development, happy to remove it.

sebgl: I think I would remove it, but it's up to you, I really don't mind keeping it.

pkg/controller/common/defaults/pod_template.go
sebgl: I understand, that makes sense 👍

pkg/controller/common/defaults/pod_template_test.go
barkbay: The `Defaulter` is tested through the use of some `PodTemplate`, I guess it's fine to let them here ?

pkg/controller/common/container/defaulter.go
barkbay: I think we should not sort env variable since they may reference each other, 👍 for the ports

pkg/controller/common/defaults/pod_template_test.go
sebgl: Yeah I'm fine with that.

Issue #2791, opened 2020-03-31
Elasticsearch 8.0.0-SNAPSHOT fails at startup due to volume permissions


sebgl:
Exception raised at startup:

```
["org.elasticsearch.bootstrap.StartupException: java.lang.IllegalStateException: failed to obtain node locks, tried [/usr/share/elasticsearch/d
ata]; maybe these locations are not writable or multiple nodes were started on the same data path?",
```

I think that's because the Docker image runs with user `elasticsearch` by default, whereas it was using user `root` before (even though the elasticsearch process itself runs as user `elasticsearch`):

```
⟩ docker run -ti docker.elastic.co/elasticsearch/elasticsearch:7.6.0 id
uid=0(root) gid=0(root) groups=0(root)
⟩ docker run -ti docker.elastic.co/elasticsearch/elasticsearch:8.0.0-SNAPSHOT id
uid=1000(elasticsearch) gid=0(root) groups=0(root)
```

sebgl:
We run an init container to change the owner of the data volume to `elasticsearch`, but only if the init container runs with the `root` user:

```bash
        # chown the data and logs volume to the elasticsearch user
	# only done when running as root, other cases should be handled
	# with a proper security context
	chown_start=$(date +%s)
	if [[ $EUID -eq 0 ]]; then
		{{range .ChownToElasticsearch}}
			echo "chowning {{.}} to elasticsearch:elasticsearch"
			chown -v elasticsearch:elasticsearch {{.}}
		{{end}}
	fi
```

In 8.0.0-SNAPSHOT the init container runs with the `elasticsearch` user, hence does not have permission to chown the volume.
If I comment the `if` condition above, the init container fails with:

```
chowning /usr/share/elasticsearch/data to elasticsearch:elasticsearch
chown: changing ownership of '/usr/share/elasticsearch/data': Operation not permitted
failed to change ownership of '/usr/share/elasticsearch/data' from root:root to elasticsearch:elasticsearch
```

pebrc:
Related to https://github.com/elastic/cloud-on-k8s/issues/2599

sebgl:
I think the way we currently deal with volumes permissions is not great: we run an init container to `chown` the mounted volumes (with write access to user `root`) so they belong to the `elasticsearch` user instead.

[...]

PR #3342, opened 2020-06-29
Add set-default-security-context flag to the operator

david-kow:
Implements @sebgl proposal from https://github.com/elastic/cloud-on-k8s/issues/2791#issuecomment-620003970. We should make a final call whether we want to only do it for 8.0+ and I can add that. 

Also, for Kibana to work we also need a separate fix for 8.0 default resources [issue](https://github.com/elastic/cloud-on-k8s/issues/3344) - Kibana OOMs with 1Gi.


david-kow:
jenkins test this please

cmd/manager/main.go
```diff
@@ -174,6 +174,11 @@ func init() {
 		DefaultWebhookConfigurationName,
 		fmt.Sprintf("Name of the Kubernetes ValidatingWebhookConfiguration resource (defaults to %s). Only used when enable-webhook is true.", DefaultWebhookConfigurationName),
 	)
+	Cmd.Flags().Bool(
+		operator.SetDefaultFsGroupFlag,
+		true,
+		"Enables setting the default filesystem group in Pods security context on ES 8.0+",
```
sebgl, line 180: Maybe `ES -> Elasticsearch` to be completely explicit.

pkg/controller/common/operator/parameters.go
```diff
@@ -26,6 +26,9 @@ type Parameters struct {
 	CertRotation certificates.RotationParams
 	// MaxConcurrentReconciles controls the number of goroutines per controller.
 	MaxConcurrentReconciles int
+	// SetDefaultFsGroup determines whether the operator should set the default
+	// filesystem group for Pod security context on ES 8.0+. Ignored pre-8.0.
```
sebgl, line 30: Let's make this comment consistent with the flag description?

pkg/controller/common/defaults/pod_template.go
```diff
@@ -300,3 +301,13 @@ func (b *PodTemplateBuilder) WithAutomountServiceAccountToken() *PodTemplateBuil
...
+
+func (b *PodTemplateBuilder) WithFsGroup(defaultFsGroup int64) *PodTemplateBuilder {
+	if b.PodTemplate.Spec.SecurityContext == nil {
+		b.PodTemplate.Spec.SecurityContext = &corev1.PodSecurityContext{}
+	}
+	if b.PodTemplate.Spec.SecurityContext.FSGroup == nil {
+		b.PodTemplate.Spec.SecurityContext.FSGroup = pointer.Int64(defaultFsGroup)
+	}
+	return b
+}
```
sebgl, line 313: I think we'd want an escape hatch from that default in the podTemplate. Users should be allowed to express: "I want to override my fsGroup to `nothing`".

It's probably not easy to do at the `fsGroup` level, since it's an int pointer.
However I was thinking we could do it at the `securityContext` level:

```yaml
kind: Elasticsearch
metadata:
  name: elasticsearch-sample
spec:
  version: 8.0.0-SNAPSHOT
  nodeSets:
  - name: default
    count: 1
    podTemplate:
      spec:
        securityContext: {}
```

When I specify the above, I would like ECK to not set the fsGroup (even though the operator has the flag enabled).

So the `WithFsGroup ` function behaviour would be:

- if the securityContext is nil: set a default one with the default fsGroup
- if the securityContext is not nil (could be empty): don't set the fsGroup, consider the user want their own securityContext applied.

Given this, I'm wondering whether we should name the flag `set-default-security-context` instead.

sebgl: We discussed this offline with @david-kow, the use cases for the above are not super convincing:

- the operator is misconfigured, maybe you cannot reconfigure it for some reasons (eg. installed from operatorhub, or you don't have access to the operator namespace), but you'd still want to escape the fsGroup problem
- there are multiple PSPs on the cluster, you may want fsGroup set for some resources in some namespaces, but not for other resources in other namespaces

I still think it's a nice-to-have escape (also probably easier for just running the quickstart on Openshift without altering the operator config). Curious about other people's thought though @elastic/cloud-k8s.

anyasabo:
>TIL there seems to be a way to customize operators deployed via OLM through the subscription CRD https://github.com/operator-framework/operator-lifecycle-manager/blob/master/doc/design/subscription-config.md

This seems like a good argument to move to using env vars by default rather than cli flags (which take precedence): https://github.com/elastic/cloud-on-k8s/blob/master/hack/manifest-gen/assets/charts/eck/templates/statefulset.yaml#L58

barkbay:
> TIL there seems to be a way to customize operators deployed via OLM through the subscription CRD

This can't be done through the web ui though. Even if primary interest of OLM is no lost (managing the deployment and the lifecycle of the operator) it will be hard for users using the Openshift console to figure out what is happening.

pebrc:
> This can't be done through the web ui though.


Technically it can be done even in the OpenShift console if you count editing raw-YAML as "through the web ui"

Screenshot 2020-08-06 at 09 57 41

barkbay:
> Technically it can be done even in the OpenShift console if you count editing raw-YAML as "through the web ui"

Agreed, but my point was more about the fact that it can't be done when the user clicks on the "Install" button. I think that most of the users are expecting an up and running operator when doing that, editing the raw content of the `Subscription` in a second step sounds like a broken process to me.

pkg/controller/common/defaults/pod_template.go
pebrc: I have a slight preference for Seb's suggestion as it seems to give users an additional way out for any case we have not considered and it seems also reasonable to default the whole security context instead of just a single attribute.

pebrc (approved):
LGTM!

pkg/controller/common/version/version.go
```diff
@@ -23,6 +23,11 @@ var (
 	SupportedBeatVersions             = MinMaxVersion{Min: From(7, 0, 0), Max: From(8, 99, 99)}
 )
 
+var (
+	// MinDefaultSecurityContextVersion is the minimal version for which the operator will inspect set-default-security-context flag.
+	MinDefaultSecurityContextVersion = MustParse("8.0.0")
```
sebgl, line 28: I think I would move that variable to `pkg/controller/elasticsearch/nodespec/podspec.go, very close to the `defaultFsGroup` const. This way we have everything at a single place, and we can write a comment with all the details that may be lost in space otherwise:

```go
// Starting 8.0.0, the Elasticsearch container does not run with the root user anymore. As a result,
// we cannot chown the mounted volumes to the right user (id 1000) in an init container.
// Instead, we can rely on Kubernetes `securityContext.fsGroup` feature: by setting it to 1000
// mounted volumes can correctly be accessed by the default container user.
// On some restricted environments (custom PSPs or Openshift), setting the Pod security context
// is forbidden: the user can either set `--set-default-security-context=false`, or override the
// podTemplate securityContext to an empty value.
var minDefaultSecurityContextVersion = version.MustParse("8.0.0")
const defaultFsGroup = 1000
```

cmd/manager/main.go
```diff
@@ -218,6 +218,11 @@ func Command() *cobra.Command {
 		DefaultWebhookName,
 		"Name of the Kubernetes ValidatingWebhookConfiguration resource. Only used when enable-webhook is true.",
 	)
+	cmd.Flags().Bool(
+		operator.SetDefaultSecurityContextFlag,
+		true,
+		"Enables setting the default security context of Elasticsearch 8.0+ Pods, ignored pre-8.0",
```
sebgl, line 224: We can make this even more explicit:

```
Enables setting the default security context with fsGroup=1000 for Elasticsearh 8.0+ Pods. Ignored pre-8.0.
```

sebgl (commented):
I'm good with the approach 👍 

I would expect we also change the init container script to get rid of this block starting 8.0: https://github.com/elastic/cloud-on-k8s/blob/090d2aa689e012d351ccf7f3720547a9156060be/pkg/controller/elasticsearch/initcontainer/prepare_fs_script.go#L114-L128

Other than that we probably need some E2E tests tweak for ES 8.0 on Openshift, and the right documentation for this new behaviour, but that can be done as part of different PRs.

cmd/manager/main.go
```diff
@@ -218,6 +218,11 @@ func Command() *cobra.Command {
 		DefaultWebhookName,
 		"Name of the Kubernetes ValidatingWebhookConfiguration resource. Only used when enable-webhook is true.",
 	)
+	cmd.Flags().Bool(
+		operator.SetDefaultSecurityContextFlag,
+		true,
+		"Enables setting the default security context with fsGroup=1000 for Elasticsearch 8.0+ Pods. Ignored pre-8.0.\n",
```
sebgl, line 224: Why `\n`?

sebgl (commented):
Can we also add this flag to the list of explicit values in our manifest-gen: https://github.com/elastic/cloud-on-k8s/blob/master/hack/manifest-gen/assets/charts/eck/values.yaml
And to the default configmap: https://github.com/elastic/cloud-on-k8s/blob/master/hack/manifest-gen/assets/charts/eck/templates/configmap.yaml

Otherwise LGTM

sebgl (approved):
LGTM

Issue #2706, opened 2020-03-16
Migrate away from common.k8s.elastic.co prefix in annotations


pebrc:
The `common.k8s.elastic.co`, which is derived from the code structure has no meaning to users and is confusing at best. 

We should consider migrating away from it to a more self-explanatory generic namespace like: 

`eck.k8s.elastic.co`

We are currently using this prefix in the following annotations: 

* `common.k8s.elastic.co/type`
* `common.k8s.elastic.co/controller-version`
* `common.k8s.elastic.co/template-hash`
* `common.k8s.elastic.co/pause`

Related #2594

milanage:
@pebrc 

Not sure what's the progress of this issue but it seems at least `common.k8s.elastic.co/controller-version`  annotation is not maintained since ECK 2.0.

Should we document it?

trungcle:
another case - 01338506: customer raised a confusion of `common.k8s.elastic.co/controller-version` annotation which should be documented.
Thanks

PR #3360, opened 2020-06-30
Fix enterprise-search and apm-server label filter in the diagnostics script

sebgl:
We are fetching pod logs using their `common.k8s.elastic.co/type` label,
but its value does not always match the resource Kind.
Examples:
- `apm-server` vs. `apmserver`
- `enterprise-search` vs. `enterprisesearch`

Logs of the corresponding Pods are therefore missing from the
diagnostics.

This commit distinguishes resource kind from type label. The code is a
bit more verbose but I think that's fine.

### DIFF
--- a/hack/diagnostics/eck-dump.sh
+++ b/hack/diagnostics/eck-dump.sh
@@ -116,12 +116,18 @@ main() {
     get_resources "$ns" networkpolicies
     get_resources "$ns" controllerrevisions
     list_resources "$ns" secrets
-    
-    local types="kibana,elasticsearch,apmserver,enterprisesearch,beat"
-    for t in $types; do
-      get_resources "$ns" $t
-      get_logs "$ns" common.k8s.elastic.co/type=$t
-    done
+
+    # get all managed resources and their logs
+    get_resources "$ns" kibana
+    get_logs "$ns" common.k8s.elastic.co/type=kibana
+    get_resources "$ns" elasticsearch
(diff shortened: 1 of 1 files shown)


sebgl:
Jenkins test this please.

anyasabo:
Good catch. Do we maybe want to adjust the type label instead? It seems odd that the `common.k8s.elastic.co/type` value does not equal the kubernetes type name.

sebgl:
> Good catch. Do we maybe want to adjust the type label instead? It seems odd that the common.k8s.elastic.co/type value does not equal the kubernetes type name.

I was feeling the same initially, but it's a breaking change for people relying on those labels. Dealing with that breaking change is not worth it imo.
We could consider this `type` represents the "product name" (eg. `Enterprise Search`, normalized to `enterprise-search`), and not the k8s resource kind? 🤷

anyasabo:
It's only breaking for people who are using APM server since ent search isn't released yet . I think I'm okay with that breaking change? Keeping names as uniform as possible is a good enough reason IMO.

It's hard for me to imagine a use case for services where you would want to select on the `type` label, rather than the `apm.k8s.elastic.co/name` label.

sebgl:
After discussion with the team: let's keep the hyphens for now. We can introduce a breaking change the day we decide to rename the label entirely: https://github.com/elastic/cloud-on-k8s/issues/2706.

PR #3374, opened 2020-07-01
Fix panic on invalid version

barkbay:
This PR fixes a panic that may occur if an invalid version is submitted by the user.


barkbay:
The change which introduced the problem has been reverted in https://github.com/elastic/cloud-on-k8s/pull/3378

I just left the unit test which, I think, may help to prevent the same problem to occur later.

anyasabo:
jenkins test this please

PR #3379, opened 2020-07-01
Add Kibana plugin docs

anyasabo:
I noticed we did not have docs on installing Kibana plugins. It is a little different since we do not default the image in Kibana (though maybe should?), and the Kibana plugin installer takes different flags, but is otherwise very similar.

I didn't substitute the version in because the example uses the specific plugin version that is compatible with that Kibana version. I would have used a first party plugin but I don't think we have any listed? Xpack errors out since it is already included so just picked the first one in the [list](https://www.elastic.co/guide/en/kibana/current/known-plugins.html).


anyasabo:
Ah yeah I was just checking that it did not error out, I missed that we did not share mounts. Updated it just to point people in the right direction

pebrc:
Should we mention in the create your custom image doc that you have to run the optimize step at image creation time for Kibana otherwise it will not work and just OOM on startup?

Issue #2827, opened 2020-04-06
Default node allocation awareness


anyasabo:
We should consider adding [allocation awareness](https://www.elastic.co/guide/en/elasticsearch/reference/current/allocation-awareness.html) for the node the Elasticsearch pod is on by default. We have some documentation [here](https://www.elastic.co/guide/en/cloud-on-k8s/1.0/k8s-advanced-node-scheduling.html#k8s-availability-zone-awareness), and we set the default antiaffinity on the pods to prefer to not be scheduled on the same node. But pods still might be scheduled on the same pod, and this would give Elasticsearch the information it needs to try and spread shards out across nodes in that situation.

We can set it to just try and spread out shards (i.e. not force), and users can override it, so I do not see a downside to adding this by default.

PR #3380, opened 2020-07-01
Add node shard allocation awareness by default

anyasabo:
Fix https://github.com/elastic/cloud-on-k8s/issues/2827

Should fix issues where multiple pods in the same ES cluster get scheduled to the same k8s node, but ES does not know so it allocates both the primary and replica shards to pods on the same k8s node. We already have docs for setting up AZ awareness, but that would require extra information that ECK does not have. We can depend on having the node name available, so we can do this by default.


pkg/controller/elasticsearch/settings/environment.go
```diff
@@ -13,8 +13,9 @@ const (
...
 	HeadlessServiceName       = "HEADLESS_SERVICE_NAME"
 
-	// EnvPodName and EnvPodIP are injected as env var into the ES pod at runtime,
+	// These are injected as env var into the ES pod at runtime,
 	// to be referenced in ES configuration file
-	EnvPodName = "POD_NAME"
-	EnvPodIP   = "POD_IP"
+	EnvPodName  = "POD_NAME"
+	EnvPodIP    = "POD_IP"
+	EnvNodeName = "node.attr.NODE_NAME"
```
barkbay, line 20: I'm not a big fan of env variables with dots as it is not supported by bash (which is the default shell in the ES image).
At first you may think that the env variable is empty if you use `export` :

```
bash-1.2$ export
...
declare -x READINESS_PROBE_PROTOCOL="https"
declare -x SHLVL="1"
declare -x TERM="xterm"
declare -x node.attr.NODE_NAME # empty ?
bash-4.2$
```

Than you will hit this kind of error if you try to print it:

```
bash-4.2$ echo ${node.attr.NODE_NAME}
bash: ${node.attr.NODE_NAME}: bad substitution
```

Eventually you'll find that you need something like awk to print the value:

```
bash-4.2$ awk 'BEGIN {print ENVIRON["node.attr.NODE_NAME"]}'
gke-michael-dev-2-default-pool-1cc419ff-nl3d
```

Is there a disadvantage to setting this value in the configuration file ?
I think your proposal does the job, so I'm also fine to keep it as is.

anyasabo: No that's a good point. I just had tunnel vision after I saw that ES does not do the `_` -> `.` conversion like Kibana does. Fixing

pkg/apis/elasticsearch/v1/fields.go
```diff
@@ -18,8 +18,10 @@ const (
 
 	NodeName = "node.name"
 
-	PathData = "path.data"
-	PathLogs = "path.logs"
+	PathData                 = "path.data"
+	PathLogs                 = "path.logs"
+	ShardAwarenessAttributes = "cluster.routing.allocation.awareness.attributes"
```
barkbay, line 23: cosmetic nit: maybe add a new line between those new fields and the existing since they address different purposes in the configuration ?

anyasabo:
@alaudazzi if you can take a look at the docs changes I'd appreciate it

pkg/controller/elasticsearch/settings/merged_config.go
```diff
@@ -16,6 +17,8 @@ import (
 	"github.com/elastic/cloud-on-k8s/pkg/controller/elasticsearch/volume"
 )
 
+var nodeAttrNodeName = fmt.Sprintf("%s.%s", esv1.NodeAttr, EnvNodeName)
```
pebrc, line 20: I believe typically these `node.attr`s are lowercase (just a nit). Also should we try to make it clear that `NODE` in this context does not mean the Elasticsearch node but the k8s node by maybe using a slightly different name?

anyasabo: Do you think just `k8s_node_name`? I went this way to minimize the amount of mappings you have to keep in mind between ES attr name -> env var name -> downward api value. I can see that that might not be a big advantage over your proposed approach though

sebgl: I think I'd be +1 with `k8s_node_name`.

pebrc: Same

Issue #3324, opened 2020-06-26
Add support for the "node.roles" setting


barkbay:
With Elasticsearch 7.9 the individual settings `node.*` are deprecated in favor of a `node.roles` setting.

Related ES PR:
* main branch: https://github.com/elastic/elasticsearch/pull/54998
* For 7.x https://github.com/elastic/elasticsearch/pull/58512

We need to update several places in the code where ECK relies on the `node.*` attributes to detect the nature of a node:

* Validation function
* When we set the labels on the Pods

Note that it is not possible for the user to use both the legacy method and the new one:

e.g.

```yaml
config:
node.master: true
node.roles:
    - data
```

produces the following error:

```
java.lang.IllegalArgumentException: can not explicitly configure node roles and use legacy role setting [node.master]=[true]
```

_See https://github.com/elastic/elasticsearch/blob/master/server/src/main/java/org/elasticsearch/cluster/node/DiscoveryNode.java#L57 for implementation details_

A concern I have is that it does not seem possible to render an empty array with ucfg, I did a few tests and the following manifest:

```
apiVersion: elasticsearch.k8s.elastic.co/v1
kind: Elasticsearch
metadata:
  name: quickstart
spec:
  version: 8.0.0
  image: docker.elastic.co/elasticsearch/elasticsearch:8.0.0-SNAPSHOT
  nodeSets:
    - name: coordinating
      podTemplate:
        spec:
          securityContext:
            runAsUser: 1000
            fsGroup: 1000
      count: 1
      config:
        node.roles: []
        node.store.allow_mmap: false
...
```

is rendered as:

```
node:
  roles: null
  store:
    allow_mmap: false
```

Which produces the following error: 

```
Exception in thread "main" org.elasticsearch.ElasticsearchParseException: null-valued setting found for key [node.roles] found at line number [10], column number [10]
```

pebrc:
I believe #3387 fixes this issue. Closing.

PR #3387, opened 2020-07-02
Add support for the node.roles setting

barkbay:
This PR adds a preliminary support for the `node.roles` setting.

See https://github.com/elastic/cloud-on-k8s/issues/3324


pkg/apis/elasticsearch/v1/elasticsearch_config.go
```diff
@@ -38,6 +73,7 @@ type ElasticsearchSettings struct {
 
 // DefaultCfg is an instance of ElasticsearchSettings with defaults set as they are in Elasticsearch.
 var DefaultCfg = ElasticsearchSettings{
+	// Values below only make senses if there is no "node.roles" in the configuration provided by the user
```
sebgl, line 76: What happens if the user specifies `node.roles`? Do we still append our own `node.data`, `node.master`, etc. defaults in the Elasticsearch config?

pebrc: I don't think so because we apply the defaults only when reading the current Elasticsearch config but not when writing it.

pkg/apis/elasticsearch/v1/elasticsearch_config.go
```diff
@@ -24,10 +30,39 @@ type ClusterSettings struct {
...
 type Node struct {
-	Master bool `config:"master"`
-	Data   bool `config:"data"`
-	Ingest bool `config:"ingest"`
-	ML     bool `config:"ml"`
+	Master bool     `config:"master"`
+	Data   bool     `config:"data"`
+	Ingest bool     `config:"ingest"`
+	ML     bool     `config:"ml"`
+	Roles  []string `config:"roles"` // available as of 7.9.0
```
sebgl, line 37: I would also add: `// takes priority over the other fields if non-nil`

anyasabo:
Should we update the docs here too? https://www.elastic.co/guide/en/cloud-on-k8s/master/k8s-node-configuration.html
since the `node.role` settings described there are deprecated in 7.9?

I also wonder if we want to add some validation around it (maybe making `node.role` and `roles` mutually exclusive, and maybe only allowing `roles` >= 7.9.0), but that could be left to a future issue or not at all since we have so far tried to avoid validating the config.

barkbay:
> Should we update the docs here too? https://www.elastic.co/guide/en/cloud-on-k8s/master/k8s-node-configuration.html
> since the `node.role` settings described there are deprecated in 7.9?

Yes, I would just wait for 7.9 to be released. I'll create an issue.

> I also wonder if we want to add some validation around it (maybe making `node.role` and `roles` mutually exclusive, and maybe only allowing `roles` >= 7.9.0), but that could be left to a future issue or not at all since we have so far tried to avoid validating the config.

I'm 👍 for checking that `roles` is only set for ES >= 7.9.0. Once our quickstart is updated some users might blindly downgrade the Elasticsearch version, I think it makes sense to immediately surface this error. I'm not sure I would not add any additional validation though.

Issue #3386, opened 2020-07-02
E2E tests log streaming may end before the tests complete


sebgl:
Forked from https://github.com/elastic/cloud-on-k8s/issues/3130.

Our OCP E2E test job fails because we did not manage to get the entire list of logs for the E2E test job:

```
04:02:15  {"log.level":"info","@timestamp":"2020-07-02T02:02:05.175Z","log.logger":"e2e-ih34m","message":"Log stream ended","service.version":"0.0.0-00000000","service.type":"eck","ecs.version":"1.4.0"}
07:11:51  {"log.level":"info","@timestamp":"2020-07-02T05:11:47.261Z","log.logger":"e2e-ih34m","message":"Test Job succeeded, waiting for log stream to be over","service.version":"0.0.0-00000000","service.type":"eck","ecs.version":"1.4.0","name":"eck-e2e-ih34m-v8bh7"}
```

When running it myself locally I also get the log streaming to end with no error way before (eg. after 1 hour) the Pod terminates.

I see 2 solutions to that problem:

1. Retry the log stream request when it ends. I found this [code snippet](https://gitlab.com/gitlab-org/gitlab-runner/-/blob/1ea8d059a297c0c33025a2f56cabf4fc0807b3c6/executors/kubernetes/log_processor.go#L184-198) which seems to cache the timestamp of the last received log line. It says `it's impossible to get two log lines with the same timestamp`, which does not seem to match what I'm observing. In doubt I guess we could also cache the log lines themselves for comparison purposes.
2. Grab the logs file differently. It could be stored in eg. a PersistentVolume or an S3 bucket at the end of the E2E tests, and retrieved by the test runner.

2nd solution feels more robust but also more complicated to implement. If it relies on an external system (eg. S3) then it makes local runs a bit more complicated. I think I'd like to give a try to option 1 first.

david-kow:
+1 on trying 1. first. Streaming logs plays nicely with running E2E locally.

PR #3389, opened 2020-07-02
Retry getting the E2E tests Pod logs when the stream ends

sebgl:
In some environments (OCP mostly), we don't manage to get a full 6 hours
long Pod log stream with no interruptions. More generally, network isn't
reliable so this could happen in any environments.

Let's retry the log stream apiserver request when that happens.

The side-effect is that we now get duplicate log entries. In CI we have a
process that parses the logs to turn them into formatted test results.
In order to not make that process fail, we ask Kubernetes to
provide a timestamp with each log, so we can keep track of the last timestamp.
On a retry, we can therefore ignore logs whose timestamp is lower than
the last one we found.
This isn't a perfect system: if the stream drops in between several logs
that share the same timestamp, the second half will be lost. In our case,
for test results, I think that's unlikely to happen.

Fixes https://github.com/elastic/cloud-on-k8s/issues/3386.

### DIFF
--- a/test/e2e/cmd/run/run.go
+++ b/test/e2e/cmd/run/run.go
@@ -5,6 +5,7 @@
 package run
 
 import (
+	"bufio"
 	"context"
 	"encoding/json"
 	"fmt"
@@ -403,8 +404,11 @@ func (h *helper) monitorTestJob(client *kubernetes.Clientset) error {
 
 	informer := factory.Core().V1().Pods().Informer()
 
-	jobStarted := false
-	streamStatus := make(chan error, 1)
+	jobStarted := false                  // keep track of the first Pod running event
+	podSucceeded := false                // keep track of when we're done
+	streamStatus := make(chan error, 1)  // receive log errors or EOF (nil)
+	stopLogStream := make(chan struct{}) // notify the log stream it can stop when EOF
+
 	var err error
 
 	informer.AddEventHandler(cache.ResourceEventHandlerFuncs{
@@ -419,7 +423,7 @@ func (h *helper) monitorTestJob(client *kubernetes.Clientset) error {
 				if !jobStarted {
 					jobStarted = true
 					log.Info("Pod started", "name", newPod.Name)
-					go h.streamTestJobOutput(streamStatus, client, newPod.Name)
+					go h.streamTestJobOutput(streamStatus, stopLogStream, client, newPod.Name)
 				} else {
 					select {
 					case streamErr := <-streamStatus:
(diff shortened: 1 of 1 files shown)


sebgl:
This is rather hard to test in order to ensure it fixes https://github.com/elastic/cloud-on-k8s/issues/3386...
What I did is a manual test where I spin up a Pod that logs numbers to stdout, invoke the `monitorTestJob` function for that Pod, and simulate breaks in the stream buffer once in a while: https://github.com/sebgl/cloud-on-k8s/compare/retry-e2e-logstream...sebgl:retry-e2e-logstream-manual-testing?expand=1. I then manually check that we don't process logs twice, and that the missing logs that share the same (nanosecond) timestamp are the expected ones.

test/e2e/cmd/run/run.go
```diff
@@ -486,15 +481,84 @@ func (h *helper) streamTestJobOutput(streamStatus chan<- error, client *kubernet
...
+
+				// don't re-write logs that have been already written in a previous log stream attempt
+				if !pastPreviousLogStream && !timestamp.After(lastTimestamp) {
+					continue
+				}
+
+				// new log to process
+				pastPreviousLogStream = true
+				lastTimestamp = timestamp
+				if _, err := writer.Write([]byte(logLine + "\n")); err != nil {
```
pebrc, line 524: Did we use to print the timestamp as well, or am I misremembering?

sebgl: We did not, I'm adding it in https://github.com/elastic/cloud-on-k8s/pull/3389/files#diff-b48621daf1e1c944a38d2c599b1d11c4R543.

sebgl:
@pebrc I did some manual tests but not a full E2E test run. I suggest we merge, trigger a nightly build and see what happens! 🤞

test/e2e/cmd/run/run.go
pebrc: Makes sense. I missed that.

test/e2e/cmd/run/run.go
```diff
@@ -486,15 +481,84 @@ func (h *helper) streamTestJobOutput(streamStatus chan<- error, client *kubernet
...
-	var buffer [logBufferSize]byte
-	if _, err := io.CopyBuffer(writer, stream, buffer[:]); err != nil && err != io.EOF {
-		streamStatus <- err
-		return
+
+	// The log stream may end abruptly in some environments where the network isn't reliable.
+	// Let's retry when that happens. To avoid duplicate log entries, keep track of the last
+	// Kubernetes log timestamp: we don't want to reprocess entries with a lower timestamp.
+	// If the stream drops in between several logs that share the same timestamp, the second half will be lost.
+	lastTimestamp := time.Time{}
```
charith-elastic, line 490:
```suggestion
	lastTimestamp := time.Now()
```

charith-elastic:
Oops, submitted the review too late :disappointed:

sebgl:
@charith-elastic thanks for the review! I merged a bit too soon :) I suggest we wait and see if this PR fixed https://github.com/elastic/cloud-on-k8s/issues/3407, and then decide whether we want to improve/optimize this code or not?

charith-elastic:
:+1:  If the fix works we can optimize it in a different PR.

PR #3400, opened 2020-07-06
Update support info

anyasabo:
Fleshes out our support description. I wonder if we want to include more information to let people know they can still open discuss / github issues even if unsupported. I think we were considering calling that "best effort"?

cc @shubhaat

### DIFF
--- a/docs/supported-versions.asciidoc
+++ b/docs/supported-versions.asciidoc
@@ -3,7 +3,7 @@
 * Google Kubernetes Engine (GKE), Azure Kubernetes Service (AKS), and Amazon Elastic Kubernetes Service (EKS)
 * Elastic Stack: 6.8+, 7.1+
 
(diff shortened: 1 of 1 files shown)


docs/supported-versions.asciidoc
```diff
@@ -1,5 +1,10 @@
 * link:https://kubernetes.io/docs/tasks/tools/install-kubectl/[kubectl] 1.11+
 * Kubernetes 1.12+ or OpenShift 3.11+
+* Google Kubernetes Engine (GKE), Azure Kubernetes Service (AKS), and Amazon Elastic Kubernetes Service (EKS)
 * Elastic Stack: 6.8+, 7.1+
 
+ECK should work with all conformant installers as listed in the spreadsheet link:https://github.com/cncf/k8s-conformance/blob/master/faq.md#what-is-a-distribution-hosted-platform-and-an-installer[here]. Distributions include source patches and so may not work as-is with ECK.
```
alaudazzi, line 6:
```suggestion
ECK should work with all conformant installers as listed in these link:https://github.com/cncf/k8s-conformance/blob/master/faq.md#what-is-a-distribution-hosted-platform-and-an-installer[FAQs]. Distributions include source patches and so may not work as-is with ECK.
```
It's recommended to spell the links where possible to avoid the "here" link.

Issue #3319, opened 2020-06-25
Create a library of Beat config examples


david-kow:
We want to provide a library of Beat config examples that cover common scenarios. They should be reachable from docs and included into E2E testing.

Some examples of use cases:
- Filebeat config for logging only from Pods with particular label(s)
- Filebeat config for logging only from particular namespace(s)
- Metricbeat config for kube-state-metrics
- Metricbeat config for an app with Prometheus metrics
- Heartbeat config for ES health monitoring

Related to https://github.com/elastic/cloud-on-k8s/issues/3216.

david-kow:
Closing as this has been already delivered.

PR #3406, opened 2020-07-08
Add Beat config library

david-kow:
This PR introduces a configuration library for Beat CRD. Yaml files in `config/recipes/beats` are self-contained and should be fully functional in non-secured cluster (PSP/SCC is out of scope).

```
$ kubectl apply -f config/recipes/beats # due to conflicting naming not all examples will be deployed
```

```
$ kubectl get es,kb,apm,ent,beat -A
NAMESPACE   NAME                                                       HEALTH   NODES   VERSION   PHASE   AGE
default     elasticsearch.elasticsearch.k8s.elastic.co/elasticsearch   green    3       7.8.0     Ready   2m39s

NAMESPACE   NAME                                  HEALTH   NODES   VERSION   AGE
default     kibana.kibana.k8s.elastic.co/kibana   green    1       7.8.0     2m39s

NAMESPACE   NAME                                   HEALTH   AVAILABLE   EXPECTED   TYPE          VERSION   AGE
default     beat.beat.k8s.elastic.co/auditbeat     green    3           3          auditbeat     7.8.0     2m39s
default     beat.beat.k8s.elastic.co/filebeat      green    3           3          filebeat      7.8.0     2m39s
default     beat.beat.k8s.elastic.co/heartbeat     green    1           1          heartbeat     7.8.0     2m38s
default     beat.beat.k8s.elastic.co/journalbeat   green    3           3          journalbeat   7.8.0     2m38s
default     beat.beat.k8s.elastic.co/metricbeat    green    1           1          metricbeat    7.8.0     2m38s
default     beat.beat.k8s.elastic.co/packetbeat    green    3           3          packetbeat    7.8.0     2m37s
```

Each recipe is tested as an E2E test with custom (per recipe) setup and checks. Docs linking to those recipes will come as a separate PR. Related: https://github.com/elastic/cloud-on-k8s/issues/3319.


config/recipes/beats/filebeat_no_autodiscover.yaml
```diff
@@ -0,0 +1,71 @@
...
+    podTemplate:
+      spec:
+        automountServiceAccountToken: true
+        terminationGracePeriodSeconds: 30
+        dnsPolicy: ClusterFirstWithHostNet
+        hostNetwork: true
+        securityContext:
+          runAsUser: 0
+          # If using Red Hat OpenShift uncomment this:
+          #privileged: true
```
barkbay, line 30: The `privileged` field does not exist in the [`PodSecurityContext`](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/api/core/v1/types.go#L3166). It is only available in the container's [`SecurityContext`](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/api/core/v1/types.go#L5848).

Also I'm wondering if it is required for all versions of Openshift since it's seems to work out of the box on Openshift 4 according to our e2e tests ? I *think* it is required on Openshift 3.11 to bypass SELinux on the `hostPath` volume.

david-kow: I've fixed the placement, I'll try to test at some point to see if only 3.11 is affected. FYI, the comment was taken from Beats examples, but you are right, it works in our E2E tests.

config/recipes/beats/filebeat_autodiscover_by_metadata.yaml
```diff
@@ -0,0 +1,115 @@
+apiVersion: beat.k8s.elastic.co/v1beta1
```
pebrc, line 1: I feel like one or two sentences explaining what each config does would be helpful, especially for those of us who are not experts in the details of Beats configuration 

Maybe either have a subdirectory with README for each or alternatively a shared README with short explanation for each wdyt?

test/e2e/beat/recipes_test.go
```diff
@@ -0,0 +1,197 @@
...
+func TestFilebeatAutodiscoverByMetadataRecipe(t *testing.T) {
+	name := "fb-autodiscover-meta"
+	podBad, badLog := loggingTestPod(name + "-bad")
+	podLabel, goodLog := loggingTestPod(name + "-label")
+	podLabel.Labels["log-label"] = "true"
+
+	customize := func(builder beat.Builder) beat.Builder {
+		return builder.
+			WithRoles(beat.PSPClusterRoleName, beat.AutodiscoverClusterRoleName).
+			WithESValidations(
```
pebrc, line 65: I think it's a bit odd that we define test validations as part of the builder, that builds the test subjects, but I guess this is a general problem in our e2e setup at the moment.

test/e2e/test/helper/yaml.go
```diff
@@ -104,3 +108,212 @@ func (yd *YAMLDecoder) ToBuilders(reader *bufio.Reader, transform BuilderTransfo
...
+			return nil, fmt.Errorf("failed to decode YAML: %w", err)
+		}
+
+		objects = append(objects, obj)
+	}
+
+	return objects, nil
+}
+
+// RunFile runs builder workflow for all known resources in a yaml file, all other objects are created before and deleted
```
pebrc, line 135:
```suggestion
// RunFile runs the builder workflow for all known resources in a yaml file, all other objects are created before and deleted
```
"builder workflow" made me think twice to understand what you mean

test/e2e/test/helper/yaml.go
```diff
@@ -104,3 +108,212 @@ func (yd *YAMLDecoder) ToBuilders(reader *bufio.Reader, transform BuilderTransfo
...
+							require.NoError(t, err)
+						}
+					},
+				})
+			}
+			return steps
+		}
+}
+
+func transform(namespace, fullTestName, suffix string, transformers []BuilderTransform, objects []runtime.Object) ([]test.Builder, []runtime.Object) {
```
pebrc, line 215: `transformToE2E`  to be a bit more specific maybe?

config/recipes/beats/packetbeat_dns_http.yaml
```diff
@@ -0,0 +1,69 @@
...
+      period: 10s
+    processors:
+    - add_cloud_metadata:
+    - add_kubernetes_metadata:
+        host: ${HOSTNAME}
+        indexers:
+        - ip_port:
+        matchers:
+        - field_format:
+            format: '%{[ip]}:%{[port]}'
```
pebrc, line 32:
```
 E0714 12:34:56.169527       1 reflector.go:125] github.com/elastic/beats/libbeat/common/kubernetes/watcher.go:146: Failed to list *v1.Pod: pods is forbidden: User "system:serviceaccount:default:default" cannot  │
│ list resource "pods" in API group "" at the cluster scope       
```

config/recipes/beats/auditbeat_hosts.yaml
```diff
@@ -0,0 +1,133 @@
...
+    - add_process_metadata:
+        match_pids: ['process.pid']
+    - add_kubernetes_metadata:
+        host: ${HOSTNAME}
+        default_indexers.enabled: false
+        default_matchers.enabled: false
+        indexers:
+        - container:
+        matchers:
+        - fields.lookup_fields: ['container.id']
```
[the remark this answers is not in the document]
david-kow: Yes.

david-kow:
I went and added missing RBAC, but then I realized that those kubernetes processors are not adding anything to the events, so I removed them. 

In case of auditbeat and journalbeat I'm not sure if kubernetes metadata makes sense at all as they interact with hosts. For packetbeat, I think it would make sense to annotate events with src/dst containers, but it doesn't seem to happen and I'm not sure if this is a Beat limitation or something is wrong with the config. I think that I'd leave that for a future improvement.

config/recipes/beats/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+This directory contains yaml manifests with example configurations for Beats. These manifests are self-contained and work out-of-the-box on any non-secured Kubernetes cluster. All of them contain three-node Elasticsearch cluster and single Kibana instance. All Beat configurations set up Kibana dashboards if they are available for a given Beat and all required RBAC resources. 
+
+#### Metricbeat for Kubernetes monitoring - `metricbeat_hosts.yaml`
+
+Deploys Metricbeat as a DaemonSet that monitors the host resource usage (cpu, memory, network, filesystem) and Kubernetes resources (nodes, pods, containers, volumes).
+
+#### Filebeat with autodiscover - `filebeat_autodiscover.yaml`
+
+Deploys Filebeat as DaemonSet with autodiscover feature enabled. All pods in all namespace will have logs shipped to a Elasticsearch cluster.
```
anyasabo, line 11:
```suggestion
Deploys Filebeat as DaemonSet with autodiscover feature enabled. All pods in all namespaces will have logs shipped to a Elasticsearch cluster.
```

config/recipes/beats/README.md
```diff
@@ -0,0 +1,39 @@
+# Beats Configuration Library
+
+This directory contains yaml manifests with example configurations for Beats. These manifests are self-contained and work out-of-the-box on any non-secured Kubernetes cluster. All of them contain three-node Elasticsearch cluster and single Kibana instance. All Beat configurations set up Kibana dashboards if they are available for a given Beat and all required RBAC resources. 
+
+#### Metricbeat for Kubernetes monitoring - `metricbeat_hosts.yaml`
+
+Deploys Metricbeat as a DaemonSet that monitors the host resource usage (cpu, memory, network, filesystem) and Kubernetes resources (nodes, pods, containers, volumes).
```
anyasabo, line 7:
```suggestion
Deploys Metricbeat as a DaemonSet that monitors the host resource usage (cpu, memory, network, filesystem) and Kubernetes resources (Nodes, Pods, Contains, Volumes).
```

config/recipes/beats/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+#### Filebeat with autodiscover - `filebeat_autodiscover.yaml`
+
+Deploys Filebeat as DaemonSet with autodiscover feature enabled. All pods in all namespace will have logs shipped to a Elasticsearch cluster.
+
+#### Filebeat with autodiscover for metadata - `filebeat_autodiscover_by_metadata.yaml`
+
+Deploys Filebeat as a DaemonSet with autodiscover feature enabled. Fullfilling any of the two conditions below will cause a given Pod logs to be shipped to a Elasticsearch cluster:
```
anyasabo, line 15:
```suggestion
Deploys Filebeat as a DaemonSet with autodiscover feature enabled. Fullfilling any of the two conditions below will cause a given Pod's logs to be shipped to an Elasticsearch cluster:
```

config/recipes/beats/filebeat_autodiscover_by_metadata.yaml
```diff
@@ -0,0 +1,115 @@
...
+    processors:
+    - add_cloud_metadata: {}
+    - add_host_metadata: {}
+  daemonSet:
+    podTemplate:
+      spec:
+        serviceAccountName: filebeat
+        automountServiceAccountToken: true
+        terminationGracePeriodSeconds: 30
+        dnsPolicy: ClusterFirstWithHostNet
```
anyasabo, line 35: I realize this is in the Beats-provided examples, but it's not clear to me why filebeat would need access to the host network namespace. I didn't get a chance to test it without it yet though

config/recipes/beats/filebeat_autodiscover_by_metadata.yaml
```diff
@@ -0,0 +1,115 @@
...
+apiVersion: kibana.k8s.elastic.co/v1
+kind: Kibana
+metadata:
+  name: kibana
+spec:
+  version: 7.8.0
+  count: 1
+  elasticsearchRef:
+    name: elasticsearch
+---
```
anyasabo, line 115:
```suggestion
...
```

`---` indicates the beginning of a yaml document and `...` the ending of one

config/recipes/beats/metricbeat_hosts.yaml
```diff
@@ -0,0 +1,188 @@
...
+        - filesystem
+        - fsstat
+        processors:
+        - drop_event:
+            when:
+              regexp:
+                system:
+                  filesystem:
+                    mount_point: ^/(sys|cgroup|proc|dev|etc|host|lib)($|/)
+      - module: kubernetes
```
anyasabo, line 49: Not necessary for this PR but I think it would be interesting to have this as a deployment example using autodiscover to scrape the kubelet metrics. There's not really any need to have this as a dset

anyasabo:
>In case of auditbeat and journalbeat I'm not sure if kubernetes metadata makes sense at all as they interact with hosts.

k8s nodes still have labels and annotations, so I could see how it *might* happen, but going by your message it does not 🤷‍♀️ maybe worth opening a docs issue just to clarify what resources/beats actually add metadata

anyasabo:
> Yaml files in config/recipes/beats are self-contained

The three filebeat examples share a beat name so overlap each other, I was a bit confused at first when the k8s metadata was not being added.

config/recipes/beats/README.md
```diff
@@ -0,0 +1,39 @@
+# Beats Configuration Library
+
+This directory contains yaml manifests with example configurations for Beats. These manifests are self-contained and work out-of-the-box on any non-secured Kubernetes cluster. All of them contain three-node Elasticsearch cluster and single Kibana instance. All Beat configurations set up Kibana dashboards if they are available for a given Beat and all required RBAC resources.
```
anyasabo, line 3: Should we add a note that none of these set resource requests/limits and they may want to set these in a production deployment?

anyasabo:
I'm at least having trouble with the auditbeat recipe, the only events in the dashboard are 
>2020-07-14T22:11:16.890Z	ERROR	[auditd]	auditd/audit_linux.go:155	Failure receiving audit events	{"error": "failed to set audit PID. An audit process is already running (PID 152)"}

Even after I deleted the dset and let ECK recreate it. So hopped on and saw

>sabo@gke-sabo-dev-cluster-default-pool-c24f3bc2-ggb1 ~ $ ps aux | grep 152
root         152  0.0  0.8 340984 262828 ?       Ss   Jul08   0:35 /usr/lib/systemd/systemd-journald

And of course @jordansissel is forever in our hearts and already ran into the same issue:
https://github.com/elastic/beats/issues/8523

I wonder if we want to just link to that as a known COS issue. I can also see the argument for including the fix in a recipe though since we have a lot of GKE customers.

config/recipes/beats/README.md
```diff
@@ -0,0 +1,39 @@
+# Beats Configuration Library
```
sebgl, line 1: Can "library" be a confusing term to use? To me it sounds more like a programmatic thing than a collection of examples. I'd appreciate thought from better english speakers though (cc. @anyasabo @charith-elastic)

config/recipes/beats/README.md
sebgl: > none of these set resource requests/limits

should we set default reasonable ones in the examples directly?

sebgl, line 3: Should we mention how the Elasticsearch spec is not suitable to a production environment?

config/recipes/beats/README.md
sebgl, line 1: I'm wondering how those examples will be referenced from the documentation on the website. I think there's value in listing all the examples explicitly in the docs (ie. copying this markdown file content, basically).

config/recipes/beats/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+#### Auditbeat - `auditbeat_hosts.yaml`
+
+Deploys Auditbeat as a DaemonSet that checks file integrity and audits file operations on the host system.
+
+#### Journalbeat - `journalbeat_hosts.yaml`
+
+Deploys Journalbeat as a DaemonSet that ships data from systemd journals.
+
+
```
sebgl, line 36: nit: new line

pebrc:
> Don't think this is a blocker but: I'm at least having trouble with the auditbeat recipe, the only events in the dashboard are

I forgot to mention I tested on AKS where I did not run into the auditbeat issue @anyasabo found.

config/recipes/beats/README.md
pebrc: Good point! Maybe we should just go with "Beats Configuration Examples"?

charith-elastic: _The Beats Zoo_?  :stuck_out_tongue: 
+1 for _Beats Configuration Examples_

charith-elastic: [I tried to include the samples in the documentation a long time ago](https://github.com/elastic/cloud-on-k8s/pull/2656) but the docs build only checks out the `docs` directory so that wasn't successful. We'll either have to move the samples inside the docs directory or figure out the magic incantation required by the docs build to checkout more directories.

anyasabo: 👍 Examples or Cookbook I think both avoid the ambiguity seb mentioned

config/recipes/beats/metricbeat_hosts.yaml
david-kow: Aren't these scoped to a node? I don't see all nodes/pods when I query kubelet endpoint.

config/recipes/beats/README.md
david-kow: > > none of these set resource requests/limits
> 
> should we set default reasonable ones in the examples directly?

For Beats, I'd say our defaults are already reasonable. I think we shouldn't overload those examples.

david-kow: This is already there in https://github.com/elastic/cloud-on-k8s/blob/master/config/recipes/README.asciidoc, should we just repeat it here?

config/recipes/beats/README.md
david-kow: I planned to copy it to the docs almost 1:1.

david-kow:
> > Yaml files in config/recipes/beats are self-contained
> 
> The three filebeat examples share a beat name so overlap each other, I was a bit confused at first when the k8s metadata was not being added. Maybe this is fine, but we probably want to update the initial PR if so. Otherwise we may want to change the name of the beats

But is "self-contained" implying they won't overlap? I didn't mean to imply the latter. I'd be reluctant to rename, because we'd also need to rename RBAC pieces and ES/Kibana for consistency. Right now they are fairly easy to ready, but if we start encoding more things into the names they might get more complicated.

config/recipes/beats/filebeat_autodiscover_by_metadata.yaml
david-kow: I checked with @exekias and while it works without it, host metadata will be richer - `host.ip`, `host.mac` will be present. I'd be fine leaving it as it is.

exekias:
@ChrsMark FYI

config/recipes/beats/README.md
anyasabo: Ah I wondered because even on a cluster with very little traffic packetbeat was using up a lot of its CPU, but I'm good with not mentioning it

config/recipes/beats/filebeat_autodiscover_by_metadata.yaml
anyasabo: I think we at least want to leave a comment, I can't be the only one who sees this and it rings alarm bells. I'm also skeptical of how useful the host network info is for container logs.

config/recipes/beats/heartbeat_es_kb_health.yaml
```diff
@@ -0,0 +1,50 @@
...
+      schedule: '@every 5s'
+      hosts: ["elasticsearch-es-http.default.svc:9200"]
+    - type: tcp
+      schedule: '@every 5s'
+      hosts: ["kibana-kb-http.default.svc:5601"]
+  deployment:
+    replicas: 1
+    podTemplate:
+      spec:
+        dnsPolicy: ClusterFirstWithHostNet
```
anyasabo, line 24: Same here for heartbeat, I don't see the advantage of running in the host network and would break in clusters where the nodes can't communicate with the pods (albeit a rare situation). I think a comment justifying it makes sense

config/recipes/beats/metricbeat_hosts.yaml
anyasabo: Correct, you could autodiscover each node and hit the kubelet on it

anyasabo:
>But is "self-contained" implying they won't overlap? I didn't mean to imply the latter.

I think that combined with the suggestion to just:

>kubectl apply -f config/recipes/beats

to review is what threw me off. I'm good with leaving the names as is, just a note for other reviewers 👍

config/recipes/beats/metricbeat_hosts.yaml
david-kow: Hm, we could, but we already have a Metricbeat running everywhere for system metricsets, this would just add another one. I'll create an issue to discuss.

config/recipes/beats/README.md
sebgl: I'm not sure people will notice the root README, it's worth repeating here imo.

Issue #3139, opened 2020-05-25
Support transform node type as of 7.7


pebrc:
Elasticseach 7.7 introduces a new node type called `transform`. We should support this new node type by adding a pod labels in the same way we do it for the older node types to facilitate selection via services 

ES docs https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-node.html#transform-node

pebrc:
Closed with #3411

PR #3411, opened 2020-07-08
Add metadata for transform node role 

pebrc:
Second attempt at supporting the transform node role Elasticsearch 7.7.0 introduces.

* add this new node type to the pod metadata when running a version of Elasticsearch that supports transform. 
* adjust test cases and default values.

The complication here is that we now have version specific default configurations. I chose to just toggle the `transform` value off in the default config to be able to keep using just one config struct. Technically the `transform` value does not even exist prior to 7.7.0. The default behaviour differs depending on whether the node is a data node or not. On nodes with `node.data: false` `node.transform` is also `false` unless specified otherwise explicitly.


thbkrkr:
I wonder if we should add `node.transform` in some examples used in the documentation where there are already all the other roles, for the sake of completeness.

pkg/apis/elasticsearch/v1/elasticsearch_config.go
```diff
@@ -65,34 +69,57 @@ func (n *Node) HasMLRole() bool {
...
+		}
+		dataNode, err := cfg.Bool(NodeData, -1, commonv1.CfgOptions...)
+		if err == nil && !dataNode {
+			settings.Node.Transform = false
+		}
+	}
+	return settings
 }
 
 // Unpack unpacks Config into a typed subset.
```
thbkrkr, line 112:
```suggestion
// UnpackConfig unpacks Config into a typed subset.
```

thbkrkr:
Relates to https://github.com/elastic/cloud-on-k8s/issues/3139.

PR #3419, opened 2020-07-13
Add support for APM tracing to e2e test runs

### LINKED ISSUE #3031: Enable APM tracing in e2e tests
We should enable the APM tracing with the `--enable-tracing` operator flag and configure the APM go library we use via environment variables to output the tracing data into our shared monitoring cluster.

The intent behind this change is to 
* exercise the APM tracing code in tests
* create data in the test monitoring cluster to evaluate the usefulness of this feature in order to decide whether we want to support it in the long run

pebrc:
Adds support to enable APM tracing during e2e test runs. 
* introduces a separate secret for the APM server certificate and secret token to be mounted in the operator
* when present configures the operator to enable tracing and configures the APM Go Agent via environment variables.

Not to be merged before the necessary secrets are in place.

Fixes #3031


test/e2e/cmd/run/run.go
```diff
@@ -208,24 +209,32 @@ func (h *helper) initTestSecrets() error {
...
 		}
 
 		monitoringSecrets := struct {
 			MonitoringIP   string `json:"monitoringIp"`
 			MonitoringUser string `json:"monitoringUser"`
 			MonitoringPass string `json:"monitoringPass"`
 			MonitoringCa   string `json:"monitoringCa"`
+			APMServerCert  string `json:"apm_server_cert"`
+			APMSecretToken string `json:"apm_secret_token"`
+			APMServerURL   string `json:"apm_server_url"`
```
pebrc, line 222: I know this is inconsistent with what we already have, but I found the chosen naming convention a bit unorthodox. Happy to normalize to the established scheme

sebgl: I'm fine with any of `monitoringIp`, `monitoring-ip`, `monitoring_ip` but we should indeed choose a single one 😄

I'd be +1 with keeping your suggestion in this PR (lower_camelcase everywhere), and use the same naming scheme for the older existing `monitoring*` fields (in a follow-up PR or at least an issue?).

pebrc:
Screenshot 2020-07-17 at 18 49 45

I tested both against a local environment and against the monitoring cluster we use for CI to 🤞

test/e2e/cmd/run/run.go
pebrc: I created #3419

PR #3422, opened 2020-07-13
Add Operator Hub release generator

charith-elastic:
Adding a new release to Operator Hub requires a lot of copy-pasting and updating the version in multiple places -- which can be quite error prone. This utility automates most of the manual work involved. It downloads the release YAML, extracts the CRDs and RBAC definitions from it and generates the set of files required to publish a new release.


hack/operatorhub/templates/csv.tpl
```diff
@@ -0,0 +1,233 @@
...
+    {{- range .CRDList }}
+    - description: {{ .Description }}
+      displayName: {{ .DisplayName }}
+      group: {{ .Group }}
+      kind: {{ .Kind }}
+      name: {{ .Name }}
+      version: {{ .Version }}
+    {{- end }}
+  description: 'Elastic Cloud on Kubernetes automates the deployment, provisioning,
+    management, and orchestration of Elasticsearch, Kibana and the APM Server on Kubernetes.
```
anyasabo, line 128: I wonder if we want to simplify this to just "Elastic stack?"

hack/operatorhub/README.md
```diff
@@ -0,0 +1,17 @@
+Operator Hub Release Generator
+===============================
+
+Extracts CRDs and RBAC definitions from `all-in-one.yaml` and generates the files required to publish a new release to Operator Hub.
```
anyasabo, line 4: It may also be worth nothing that we still need to update `elastic-cloud-eck.package.yaml` as well. And maybe also the examples of how to test the generated config (`make operator.test OP_PATH=upstream-community-operators/elastic-cloud`)? It would be nice as a release manager to just come here and have the readme tell me everything I need to do.

charith-elastic: The package file is generated by the tool as well. I plan to update the release checklist to point to this tool so hopefully that addresses your second concern.

anyasabo: Ah I switched dirs right into the 1.1.2 dir 🤦‍♀️ . and yep having it in the release checklist works for me, though I still think having instructions how to run the test/cleanup directives is useful so you don't need to rummage through the operator hub docs

Issue #3424, opened 2020-07-13
 WebhookConfigurationName is hardcoded and not configurable


wmellouli:
## Proposal

Actually, we have many environments in the same kubernetes cluster, so we need to create multiple instances of eck operator in the same k8s cluster (test, dev, ...). 

To do that, we are using kustomize with `nameSuffix` feature to add a suffix on all k8s name ressources, e.g. : `nameSuffix: -dev`
so with kustomize we create a `ValidatingWebhookConfiguration` with name: `elastic-webhook.k8s.elastic.co-dev`

When we install eck operator, we get this error` validatingwebhookconfigurations.admissionregistration.k8s.io elastic-webhook.k8s.elastic.co not found`

The cause of this error is that `WebhookConfigurationName` is hardcoded as a constant and there is no way to override this value: https://github.com/elastic/cloud-on-k8s/blob/4c359ad33c32d801d26c6283209713e337f98162/cmd/manager/main.go#L65

I suggest to create a new optional flag: `WebhookConfigurationNameFlag` (with a default value `elastic-webhook.k8s.elastic.co`).
The new flag will allow users using `kustomize` with `nameSuffix` or `namePrefix` to be able to patch the `WebhookConfigurationName` using manager env variables. For example with a `patchesStrategicMerge`:

```
---
apiVersion: apps/v1
kind: StatefulSet
metadata:
  name: elastic-operator
  namespace: elastic-system
spec:
  template:
    spec:
      containers:
        - name: manager
          env:
            - name: WEBHOOK_CONFIGURATION_NAME
              value: elastic-webhook.k8s.elastic.co-dev
```

wmellouli:
I created this pull request: https://github.com/elastic/cloud-on-k8s/pull/3425

wmellouli:
FYI: related issue https://github.com/elastic/cloud-on-k8s/issues/2581

anyasabo:
This is related to https://github.com/elastic/cloud-on-k8s/issues/3431

Currently we really only support one webhook/k8s cluster. I think this makes sense as a stepping stone to supporting this use case.

wmellouli:
Related PR was merged https://github.com/elastic/cloud-on-k8s/pull/3425. I close this issue.

PR #3425, opened 2020-07-13
Add WebhookConfigurationName Flag (#3424)

wmellouli:
Add a new optional flag `webhook-configuration-name` (env `WEBHOOK_CONFIGURATION_NAME`) to allow users to change the default `WebhookConfigurationName` value `elastic-webhook.k8s.elastic.co` when needed.

`make unit` => OK
`make integration` =>  OK


thbkrkr:
jenkins test this please

cmd/manager/main.go
```diff
@@ -168,6 +168,11 @@ func init() {
 		"",
 		fmt.Sprintf("K8s secret mounted into the path designated by %s to be used for webhook certificates", operator.WebhookCertDirFlag),
 	)
+	Cmd.Flags().String(
+		operator.WebhookConfigurationNameFlag,
+		DefaultWebhookConfigurationName,
+		fmt.Sprintf("Name of K8s ValidatingWebhookConfiguration to be used for webhook params (defaults to %s)", DefaultWebhookConfigurationName),
```
sebgl, line 174: I would maybe rephrase to `Name of the Kubernetes ValidatingWebhookConfiguration resource` to be more explicit as a flag. People don't need to know about the "webhook params" which is our own internal struct in the code.

anyasabo: Also maybe a note along the lines of "only used when `enable-webhook` is `true`"

anyasabo:
Same, LGTM just one nit about docs. I also considered changing the flag to `webhook-config-name` since we tend to use abbreviations for flags, but it might be worth being explicit here.

cmd/manager/main.go
wmellouli: @sebgl , @anyasabo thanks for your PR review.
I rephrased as suggested: `Name of the Kubernetes ValidatingWebhookConfiguration resource (defaults to %s). Only used when enable-webhook is true.`

anyasabo:
jenkins test this please

sfesfizh:
When the release 1.3.0 with this fix will be released? Do you have a plan?

Issue #3363, opened 2020-07-01
Use a static role name for community Beats


david-kow:
For official Beats that we support we distinguish 4 version ranges to pick right permissions. Other Beats might not follow the same version numbering and should not need to be concerned with versions at all since user is responsible for creating correct role anyway.

We should use a static role name (eg. `fmt.Sprintf("eck_beat_%s_role", beatType)`) and omit the version if the Beat type is not known.

PR #3432, opened 2020-07-14
Make Beat role name version-independent for Community Beats

david-kow:
1) Defines well-known Beats list in a single place.
2) For community Beats:
    - the role name used no longer depends on the Beat version. Community Beats can be versioned differently than official Beats so we shouldn't expect users to follow version changes in role names that come from official Beats changes.
    - don't include other role names (`kibana_admin`, etc.). I think we could go either way on this one, but I wouldn't make any assumptions about needed permissions.

Note that the alternative option to creating right role is to create a user in ES and pass username/password in the config while still providing `elasticsearchRef` - we will take credentials from user config, but still mount/use the right CA in Beat pods.

Fixes https://github.com/elastic/cloud-on-k8s/issues/3363.

### DIFF
--- a/docs/orchestrating-elastic-stack-applications/beat.asciidoc
+++ b/docs/orchestrating-elastic-stack-applications/beat.asciidoc
@@ -203,8 +203,12 @@ You can specify the Beat to deploy and its version through `type` and `version` 
 
 ECK supports the deployment of any Community Beat. `type` and `version` specification elements have to be provided. In addition:
 
-- `image` element in the specification must point to the image to be deployed
--  a role must exist in Elasticsearch and have the permissions required by the Beat. The role name must be `eck_beat_$type_role`, where `$type` is the Beat type. For example, when deploying `type: kafkabeat`, the role name would be `eck_beat_kafkabeat_role`. See the link:https://www.elastic.co/guide/en/elasticsearch/reference/current/defining-roles.html[Elasticsearch documentation] for more details.
+1. `image` element in the specification must point to the image to be deployed
+2. the following roles must exist in Elasticsearch:ave the permissions required by the Bea
+- if `elasticsearchRef` is provided, a role with `eck_beat_es_$type_role` name must exist, where `$type` is the Beat type. For example, when deploying `type: kafkabeat`, the role name would be `eck_beat_es_kafkabeat_role`. This role must have the permissions required by the Beat. See the link:https://www.elastic.co/guide/en/elasticsearch/reference/current/defining-roles.html[Elasticsearch documentation] for more details.
+- if `kibanaRef` is provided, then, similarily to the above, a role with `eck_beat_kibana_$type_role` name must exist. This role must have the permissions required by the Beat for the dashboard setup purposes.
+
+Alternatively to the 2. above, a user can be created in Elasticsearch and its credentials included in Beats `config` specification element  for Elasticsearch output, Kibana setup or both. If `elasticsearchRef` and `kibanaRef` are also used, ECK will still manage CA mounting and rollover, while using the provided user credentials.
 
 
 [id="{p}-beat-set-up-kibana-dashboards"]
--- a/docs/orchestrating-elastic-stack-applications/beat.asciidoc
+++ b/docs/orchestrating-elastic-stack-applications/beat.asciidoc
@@ -205,11 +205,10 @@ ECK supports the deployment of any Community Beat. `type` and `version` specific
 
 1. `image` element in the specification must point to the image to be deployed
 2. the following roles must exist in Elasticsearch:ave the permissions required by the Bea
-- if `elasticsearchRef` is provided, a role with `eck_beat_es_$type_role` name must exist, where `$type` is the Beat type. For example, when deploying `type: kafkabeat`, the role name would be `eck_beat_es_kafkabeat_role`. This role must have the permissions required by the Beat. See the link:https://www.elastic.co/guide/en/elasticsearch/reference/current/defining-roles.html[Elasticsearch documentation] for more details.
-- if `kibanaRef` is provided, then, similarily to the above, a role with `eck_beat_kibana_$type_role` name must exist. This role must have the permissions required by the Beat for the dashboard setup purposes.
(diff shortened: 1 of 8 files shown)


docs/orchestrating-elastic-stack-applications/beat.asciidoc
```diff
@@ -203,8 +203,12 @@ You can specify the Beat to deploy and its version through `type` and `version`
 
 ECK supports the deployment of any Community Beat. `type` and `version` specification elements have to be provided. In addition:
 
-- `image` element in the specification must point to the image to be deployed
--  a role must exist in Elasticsearch and have the permissions required by the Beat. The role name must be `eck_beat_$type_role`, where `$type` is the Beat type. For example, when deploying `type: kafkabeat`, the role name would be `eck_beat_kafkabeat_role`. See the link:https://www.elastic.co/guide/en/elasticsearch/reference/current/defining-roles.html[Elasticsearch documentation] for more details.
+1. `image` element in the specification must point to the image to be deployed
+2. the following roles must exist in Elasticsearch:ave the permissions required by the Bea
```
charith-elastic, line 207: Editing error

docs/orchestrating-elastic-stack-applications/beat.asciidoc
```diff
@@ -203,8 +203,12 @@ You can specify the Beat to deploy and its version through `type` and `version`
...
+- if `elasticsearchRef` is provided, a role with `eck_beat_es_$type_role` name must exist, where `$type` is the Beat type. For example, when deploying `type: kafkabeat`, the role name would be `eck_beat_es_kafkabeat_role`. This role must have the permissions required by the Beat. See the link:https://www.elastic.co/guide/en/elasticsearch/reference/current/defining-roles.html[Elasticsearch documentation] for more details.
```
charith-elastic, line 208:
```suggestion
- if `elasticsearchRef` is provided, a role with `eck_beat_es_$type_role` name must exist, where `$type` is the Beat type. For example, when deploying `kafkabeat`, the role name would be `eck_beat_es_kafkabeat_role`. This role must have the permissions required by the Beat. See the link:https://www.elastic.co/guide/en/elasticsearch/reference/current/defining-roles.html[Elasticsearch documentation] for more details.
```

docs/orchestrating-elastic-stack-applications/beat.asciidoc
```diff
@@ -203,8 +203,12 @@ You can specify the Beat to deploy and its version through `type` and `version`
...
+- if `kibanaRef` is provided, then, similarily to the above, a role with `eck_beat_kibana_$type_role` name must exist. This role must have the permissions required by the Beat for the dashboard setup purposes.
```
charith-elastic, line 209:
```suggestion
- if `kibanaRef` is provided, then, as above, a role named `eck_beat_kibana_$type_role` must exist with the permissions required to setup Kibana dashboards.
```

docs/orchestrating-elastic-stack-applications/beat.asciidoc
```diff
@@ -203,8 +203,12 @@ You can specify the Beat to deploy and its version through `type` and `version`
...
+
+Alternatively to the 2. above, a user can be created in Elasticsearch and its credentials included in Beats `config` specification element  for Elasticsearch output, Kibana setup or both. If `elasticsearchRef` and `kibanaRef` are also used, ECK will still manage CA mounting and rollover, while using the provided user credentials.
```
charith-elastic, line 211:
```suggestion
Alternatively,  create a user in Elasticsearch and include the credentials in the Beats `config` for Elasticsearch output, Kibana setup or both. If `elasticsearchRef` and `kibanaRef` are also defined, ECK will use the provided user credentials when setting up the connections.
```

pkg/apis/beat/v1beta1/beat_types.go
```diff
@@ -11,6 +11,10 @@ import (
 	commonv1 "github.com/elastic/cloud-on-k8s/pkg/apis/common/v1"
 )
 
+var (
+	KnownTypes = []string{"filebeat", "metricbeat", "heartbeat", "auditbeat", "journalbeat", "packetbeat"}
```
charith-elastic, line 15: Can we make this a set (`map[string]struct{}`) as it is searched through several times in the code below.

docs/orchestrating-elastic-stack-applications/beat.asciidoc
david-kow: I wanted to to hint that this has to be the same as the type in the spec. Wouldn't this change hide that?

pkg/apis/beat/v1beta1/beat_types.go
david-kow: I'd guess that with these 6 elements the slice will be smaller and faster, but I don't think it's worth an experiment and I've made the change :)

docs/orchestrating-elastic-stack-applications/beat.asciidoc
charith-elastic: I think it's implied. Reading "...when deploying `type: kafkabeat`..." doesn't feel right to me but maybe it's fine. Perhaps @alaudazzi can chime in here.

docs/orchestrating-elastic-stack-applications/beat.asciidoc
```diff
@@ -203,9 +203,12 @@ You can specify the Beat to deploy and its version through `type` and `version`
 
 ECK supports the deployment of any Community Beat. `type` and `version` specification elements have to be provided. In addition:
 
-- `image` element in the specification must point to the image to be deployed
--  a role must exist in Elasticsearch and have the permissions required by the Beat. The role name must be `eck_beat_$type_role`, where `$type` is the Beat type. For example, when deploying `type: kafkabeat`, the role name would be `eck_beat_kafkabeat_role`. See the link:https://www.elastic.co/guide/en/elasticsearch/reference/current/defining-roles.html[Elasticsearch documentation] for more details.
+1. `image` element in the specification must point to the image to be deployed
+2. the following roles must exist in Elasticsearch:ave the permissions required by the Bea
```
charith-elastic, line 207:
```suggestion
2. the following roles must exist in Elasticsearch and have the permissions required by the Beat:
```

david-kow: Thanks! For some reason it didn't get to the first fix commit, fixed now.

PR #3436, opened 2020-07-15
Update quickstart.asciidoc

DanRoscigno:
- Have you followed the [contributor guidelines](https://github.com/elastic/cloud-on-k8s/tree/master/CONTRIBUTING.md)?
- If you submit code, is your pull request against master? We recommend pull requests against master. We will backport them as needed.

### DIFF
--- a/docs/quickstart.asciidoc
+++ b/docs/quickstart.asciidoc
@@ -42,7 +42,7 @@ NOTE: If you are using Amazon EKS, make sure the Kubernetes control plane is all
 kubectl apply -f https://download.elastic.co/downloads/eck/{eck_version}/all-in-one.yaml
 ----
 
-. Monitor the operator logs:
+. Monitor the operator logs (watch for `Starting workers`):
 +
 [source,sh]
 ----


DanRoscigno:
> @DanRoscigno I'm a bit curious about the reason for this. There could be other log lines to look for but I guess `Starting workers` is a good example.
> Is it because you experienced an error in those logs that you should not expect and didn't know what to look at?
> 
> I think in general people should not care much about `Starting workers` (we may as well rename this or remove it from the log output entirely at some point 🤷), here we want them to know how to grab the operator logs in order to look for errors.

Hi @sebgl , because the original text said "Monitor the operator logs" I immediately thought "What am I watching for?".  I think it is a good idea to look at the logs, and by knowing what the log will say when the operator is ready the reader knows when to stop tailing the log and move on.  In the first alpha I told people to look for "starting the web hook server" as that indicated to me that the operator was ready.  

I could change it like below to give the reader the information they need (this is the namespace and pod to look at for the logs), but not suggest that they need to monitor them for a particular state:

To monitor the operator logs:
```
kubectl -n elastic-system logs -f statefulset.apps/elastic-operator
```

BTW, this is still a work in progress as I work through an end to end deployment for an observability use case.

thbkrkr:
Closing for inactivity.

PR #3446, opened 2020-07-16
Add docs note about Beats version support

david-kow:
Open to suggestion about phrasing it differently (Official Beats? Elastic Beats?) or whether we should list the products (as technically Beats are part of the Stack).

### DIFF
--- a/docs/supported-versions.asciidoc
+++ b/docs/supported-versions.asciidoc
@@ -1,7 +1,8 @@
 * link:https://kubernetes.io/docs/tasks/tools/install-kubectl/[kubectl] 1.11+
 * Kubernetes 1.12+ or OpenShift 3.11+
 * Google Kubernetes Engine (GKE), Azure Kubernetes Service (AKS), and Amazon Elastic Kubernetes Service (EKS)
-* Elasticsearch, Kibana, APM Server, Enterprise Search: 6.8+, 7.1+
+* Elasticsearch, Kibana, APM Server: 6.8+, 7.1+
+* Enterprise Search: 7.6+
 * Beats: 7.0+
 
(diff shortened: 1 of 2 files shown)


anyasabo:
I'm +1 to listing individual items of the stack

docs/supported-versions.asciidoc
```diff
@@ -1,7 +1,7 @@
 * link:https://kubernetes.io/docs/tasks/tools/install-kubectl/[kubectl] 1.11+
 * Kubernetes 1.12+ or OpenShift 3.11+
 * Google Kubernetes Engine (GKE), Azure Kubernetes Service (AKS), and Amazon Elastic Kubernetes Service (EKS)
-* Elastic Stack: 6.8+, 7.1+
+* Elasticsearch, Kibana, APM Server, Enterprise Search: 6.8+, 7.1+
```
pebrc, line 4: I don't think there is an Enterprise Search in 6.8. I believe the first version we support is 7.6

docs/supported-versions.asciidoc
```diff
@@ -1,7 +1,9 @@
 * link:https://kubernetes.io/docs/tasks/tools/install-kubectl/[kubectl] 1.11+
 * Kubernetes 1.12+ or OpenShift 3.11+
 * Google Kubernetes Engine (GKE), Azure Kubernetes Service (AKS), and Amazon Elastic Kubernetes Service (EKS)
-* Elastic Stack: 6.8+, 7.1+
+* Elasticsearch, Kibana, APM Server: 6.8+, 7.1+
+* Enterprise Search: 7.6+
```
sebgl, line 5: 7.7+, actually (doesn't exist before)

thbkrkr:
We duplicate this information in the root README of the repository. It might be a good idea to update it as well.

anyasabo:
Should we remove the supported versions from the main readme and make it a link?

david-kow:
I think it's nice to be able see right away (without clicks) what is supported, I'd be fine with leaving it as is. We shouldn't need to update it too often.

Issue #3441, opened 2020-07-16
Error during compatibility check: context deadline exceeded


laverya:
## Bug Report
**What did you do?**
Created an elasticsearch resource.

```
apiVersion: elasticsearch.k8s.elastic.co/v1
kind: Elasticsearch
metadata:
  name: myname
  namespace: mynamespace
spec:
  version: 7.8.0
  nodeSets:
  - name: zone-b
    count: 1
    config:
      node.attr.zone: us-east-1b
      cluster.routing.allocation.awareness.attributes: zone
    podTemplate:
      spec:
        affinity:
          nodeAffinity:
            requiredDuringSchedulingIgnoredDuringExecution:
              nodeSelectorTerms:
              - matchExpressions:
                - key: failure-domain.beta.kubernetes.io/zone
                  operator: In
                  values:
                  - us-east-1b
                - key: <an additional scheduling constraint>
                  operator: In
                  values:
                  - elasticsearch
        containers:
        - name: elasticsearch
          env:
          - name: ES_JAVA_OPTS
            value: -Xms2g -Xmx2g
          resources:
            requests:
              memory: 4Gi
              cpu: 0.5
            limits:
              memory: 4Gi
              cpu: 2
    volumeClaimTemplates:
      - metadata:
          name: elasticsearch-data-east1b
        spec:
          accessModes:
          - ReadWriteOnce
          resources:
            requests:
              storage: 5Gi
          storageClassName: gp2
  - name: zone-d
    count: 1
    config:
      node.attr.zone: us-east-1d
      cluster.routing.allocation.awareness.attributes: zone
    podTemplate:
      spec:
        affinity:
          nodeAffinity:
            requiredDuringSchedulingIgnoredDuringExecution:
              nodeSelectorTerms:
              - matchExpressions:
                - key: failure-domain.beta.kubernetes.io/zone
                  operator: In
                  values:
                  - us-east-1d
                - key: <an additional scheduling constraint>
                  operator: In
                  values:
                  - elasticsearch
        containers:
        - name: elasticsearch
          env:
          - name: ES_JAVA_OPTS
            value: -Xms2g -Xmx2g
          resources:
            requests:
              memory: 4Gi
              cpu: 0.5
            limits:
              memory: 4Gi
              cpu: 2
    volumeClaimTemplates:
      - metadata:
          name: elasticsearch-data-east1d
        spec:
          accessModes:
          - ReadWriteOnce
          resources:
            requests:
[...]

PR #3450, opened 2020-07-16
Add webhook removal docs

anyasabo:
Related to https://github.com/elastic/cloud-on-k8s/issues/3441

Edit: moving the wraps to a separate PR
~There's still a few places where we just return the output of `client.Update()` without wrapping any errors. If yall think it makes sense I can add it to the rest. This should cover a lot of them. It is hard to troubleshoot otherwise as can be seen in the related issue.~

~Another approach could be to add the stack traces in our client wrapper. Wrapping it inside of `pkg/utils/k8s/client.go` is challenging because adding the stack breaks things like `apierrors.IsNotFound()` which we rely on in some places, though I suppose we could just do it for everything but `Get`?~

### DIFF
--- a/docs/operating-eck/webhook.asciidoc
+++ b/docs/operating-eck/webhook.asciidoc
@@ -187,3 +187,13 @@ spec:
     ports:
       - port: 9443
 ----
+
+[id="{p}-disable-webhook"]
+== Disabling the webhook
+
+To disable the webhook, set the <<{p}-operator-config, `enable-webhook`>> operator configuration flag to `false` and remove the `ValidatingWebhookConfiguration` named `elastic-webhook.k8s.elastic.co`:
+
+[source,sh]
+----
+kubectl delete validatingwebhookconfigurations.admissionregistration.k8s.io elastic-webhook.k8s.elastic.co
+----
--- a/docs/operating-eck/webhook.asciidoc
+++ b/docs/operating-eck/webhook.asciidoc
@@ -189,7 +189,7 @@ spec:
 ----
 
 [id="{p}-disable-webhook"]
-== Disabling the webhook
+== Disable the webhook
 
 To disable the webhook, set the <<{p}-operator-config, `enable-webhook`>> operator configuration flag to `false` and remove the `ValidatingWebhookConfiguration` named `elastic-webhook.k8s.elastic.co`:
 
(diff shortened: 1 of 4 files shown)


pkg/controller/common/annotation/controller_version.go
```diff
@@ -65,7 +65,7 @@ func UpdateControllerVersion(ctx context.Context, client k8s.Client, obj runtime
 		return err
 	}
 	log.V(1).Info("updating controller version annotation", "namespace", namespace, "name", name, "kind", obj.GetObjectKind())
-	return client.Update(obj)
+	return errors.WithStack(client.Update(obj))
```
sebgl, line 68: Can't we call `WithStack()` at a higher level (ie. in the top-level Reconcile() function) to automatically add the stack for all errors?

This seems to be doing what I would expect:

```go

package main

import (
	"fmt"

	"github.com/pkg/errors"
)

func main() {
	err := myfuncWithStack()
	fmt.Printf("%+v", errors.WithStack(err))
}

func myfuncWithStack() error {
	return myfunc()
}

func myfunc() error {
	return returnerr()
}

func returnerr() error {
	return errors.New("function error")
}
```

```
❯ go run tmp.go
function error
main.returnerr
        /Users/seb/work/go/src/github.com/elastic/cloud-on-k8s/tmp.go:27
main.myfunc
        /Users/seb/work/go/src/github.com/elastic/cloud-on-k8s/tmp.go:23
main.myfuncWithStack
        /Users/seb/work/go/src/github.com/elastic/cloud-on-k8s/tmp.go:19
main.main
        /Users/seb/work/go/src/github.com/elastic/cloud-on-k8s/tmp.go:14
runtime.main
        /usr/local/Cellar/go/1.14.3/libexec/src/runtime/proc.go:203
```

anyasabo:
>Splitting this PR into 2 different unrelated PRs would make more sense to me 

I think you're right, I saw it all as "have better troubleshooting info for webhook failures" but one part is safe to backport and one isn't so I'll split them

pkg/controller/common/annotation/controller_version.go
pebrc: But that will only work for errors created with `github.com/pkg/errors`or similar libraries because they capture a stack trace in the call to `New`?

anyasabo:
Preview here: https://cloud-on-k8s_3450.docs-preview.app.elstc.co/guide/en/cloud-on-k8s/master/k8s-disable-webhook.html

pkg/controller/common/annotation/controller_version.go
sebgl: Oh yes @pebrc you are right!

docs/operating-eck/webhook.asciidoc
```diff
@@ -187,3 +187,13 @@ spec:
     ports:
       - port: 9443
 ----
+
+[id="{p}-disable-webhook"]
+== Disabling the webhook
```
thbkrkr, line 192: I think we prefer the imperative way for titles.
```suggestion
== Disable the webhook
```

PR #3459, opened 2020-07-17
[1.2] Add Beat config library (#3406) and Fix recipes E2E tests (#3462)

david-kow:
 - Add Beat config library (#3406)
 - Fix recipes E2E tests (#3462)


charith-elastic:
Doesn't this require changes to fix the permission issue in the E2E test run?

david-kow:
Would you prefer a PR that includes both this and the fix? I thought separating might be clearer - 1 to 1 PRs (and that's what my backporting tool supports :)). I'm fine both ways, let me know.

david-kow:
jenkins test this please

charith-elastic:
I just want to make sure this doesn't get merged into 1.2 before the fix is available. 

Since this PR is technically incomplete without that fix, I think I would prefer it if both were combined into a single backport PR.

Issue #3464, opened 2020-07-17
Enable stack monitoring for beats


anyasabo:
Our current beats roles only provide access to the `metricbeat-*` indices, but stack monitoring via metricbeat uses `.monitoring-*`. The docs recommend just using the `remote_monitoring_agent` role on the server you send metrics to (near the end of step 6 [here](https://www.elastic.co/guide/en/elasticsearch/reference/current/configuring-metricbeat.html)).

Current role def:

```go
		PredefinedRoles[BeatEsRoleName(V77, beat)] = esclient.Role{
			Cluster: []string{"monitor", "manage_ilm", "manage_ml", "read_ilm", "cluster:admin/ingest/pipeline/get"},
			Indices: []esclient.IndexRole{
				{
					Names:      []string{fmt.Sprintf("%s-*", beat)},
					Privileges: []string{"manage", "read", "create_doc", "view_index_metadata", "create_index"},
				},
			},
		}
```

The `remote_monitoring_agent` role on ES 7.8:

```json
{"remote_monitoring_agent": {
    "cluster": [
      "manage_index_templates",
      "manage_ingest_pipelines",
      "monitor",
      "cluster:monitor/xpack/watcher/watch/get",
      "cluster:admin/xpack/watcher/watch/put",
      "cluster:admin/xpack/watcher/watch/delete"
    ],
    "indices": [
      {
        "names": [
          ".monitoring-*"
        ],
        "privileges": [
          "all"
        ],
        "allow_restricted_indices": false
      },
      {
        "names": [
          "metricbeat-*"
        ],
        "privileges": [
          "index",
          "create_index"
        ],
        "allow_restricted_indices": false
      }
    ],
    "applications": [],
    "run_as": [],
    "metadata": {
      "_reserved": true
    },
    "transient_metadata": {
      "enabled": true
    }
  }
```

If you try to enable it it errors out with:

```json
{
    "level": "warn",
    "timestamp": "2020-07-17T16:32:11.965Z",
    "logger": "elasticsearch",
    "caller": "elasticsearch/client.go:407",
[...]

PR #3465, opened 2020-07-17
Add stack monitoring

anyasabo:
Fix https://github.com/elastic/cloud-on-k8s/issues/3464

To test it, you can apply this yaml, go to the monitoring Kibana, then choose the stack monitoring tab. You can also apply the filebeat autodiscover yaml, but edited to point to `elasticsearch-monitoring` and `kibana-monitoring` and it should also show you logs in the stack monitoring app.

One known issue:

When I tried to use `kubernetes.labels.elasticsearch_k8s_elastic_co/cluster-name: elasticsearch` and `kubernetes.labels.kibana_k8s_elastic_co/name=kibana` for the conditions it seemed to work, but metricbeat was intermittently logging errors:

```json
{"level":"info","timestamp":"2020-07-17T18:09:21.343Z","caller":"module/wrapper.go:259","message":"Error fetching data for metricset elasticsearch.index_recovery: error determining if connected Elasticsearch node is master: HTTP error 401 in : 401 Unauthorized"}
```

and

```json
{"level":"error","timestamp":"2020-07-17T18:02:16.832Z","logger":"kibana.stats","caller":"stats/stats.go:82","message":"HTTP error 401 in : 401 Unauthorized"}
```

I checked and there weren't any other pods with those labels, so for now it is using those extra labels I added and not logging any errors. It seems like it should work just using the default labels though.


config/recipes/beats/README.asciidoc
```diff
@@ -26,6 +26,10 @@ Deploys Filebeat as a DaemonSet with the autodiscover feature enabled. Logs from
 
 Deploys Filebeat as a DaemonSet with the autodiscover feature disabled. Uses the entire logs directory on the host as the input source. This configuration does not require any RBAC resources as no Kubernetes APIs are used.
 
+==== Metricbeat for Elasticsearch and Kibana Stack Monitoring - `stack_monitoring.yaml`
+
+Deploys Metricbeat configured for Elasticsearch and Kibana link:https://www.elastic.co/guide/en/kibana/current/xpack-monitoring.html[Stack Monitoring]. Deploys one monitored Elasticsearch cluster and one monitoring Elasticsearch cluster. You can access the Stack Monitoring app in the monitoring cluster's Kibana. Combine with one of the Filebeat recipes but update the `elasticsearchRef` to point to the `elasticsearch-monitoring` cluster for the full Stack Monitoring experience.
```
david-kow, line 31: Should we just add that Filebeat and be done?

anyasabo: I don't feel strongly either way. The duplication kind of bothers me but not too much

anyasabo:
This is ready for a review. To run it locally you can `make TESTS_MATCH=TestMetricbeatStackMonitoringRecipe clean docker-build docker-push e2e-docker-build e2e-docker-push e2e-run`.

config/recipes/beats/README.asciidoc
david-kow: I played with it a bit and I'm more on adding it side. Some reasons:
- we say it's stack monitoring, but even the Stack Monitoring UI is incomplete without it:
Screenshot 2020-07-20 at 07 33 55

- I applied the example and then I thought I'm combining it with a Filebeat examples, by applying one too. Took me a while to figure out ES that's being monitored has different labels than in other examples, so Stack Monitoring yaml has to be applied last

- when doing the above I also hit what seems like a [bug](https://github.com/elastic/cloud-on-k8s/issues/3470)

- we already duplicate parts of the config between three Filebeat examples, I think that's ok

It's not that many lines and brings full Stack Monitoring experience, I'd do it :)

config/recipes/beats/README.asciidoc
```diff
@@ -26,6 +26,10 @@ Deploys Filebeat as a DaemonSet with the autodiscover feature enabled. Logs from
 
 Deploys Filebeat as a DaemonSet with the autodiscover feature disabled. Uses the entire logs directory on the host as the input source. This configuration does not require any RBAC resources as no Kubernetes APIs are used.
 
+==== Metricbeat for Elasticsearch and Kibana Stack Monitoring - `stack_monitoring.yaml`
```
david-kow, line 29: We should add it to the docs too.

pkg/controller/association/controller/beat_es.go
```diff
@@ -88,23 +88,23 @@ func getBeatRoles(assoc commonv1.Associated) (string, error) {
 			"kibana_admin",
 			"ingest_admin",
 			"beats_admin",
-			"remote_monitoring_user",
+			"remote_monitoring_agent",
```
pebrc, line 91: Should this bugfix be a separate PR?

anyasabo: The rest of the PR validates the bugfix, no other test is utilizing that role which is why it was only discovered now.

Issue #3338, opened 2020-06-29
Filebeat sample doesn't work out of the box


charith-elastic:
The sample is missing the service account definition which allows it to work correctly. Ideally, the sample should "just work" when applied to a cluster.

PR #3472, opened 2020-07-20
Remove Beat sample

david-kow:
Beat sample would need to be a duplicate of one of Beat examples that we have in `config/recipes/beats`. In addition to that, we would need to alter the way we invoke SmokeTest to accommodate for Service Account, Cluster Role and Binding that sample would need to have. Due to the above, I'm removing the sample altogether.

Fixes https://github.com/elastic/cloud-on-k8s/issues/3338.


david-kow:
jenkins test this please

david-kow:
jenkins test this please

Issue #2600, opened 2020-02-24
Enforce Elastic stack upgrade order


pebrc:
Supersedes #2426 and #2353 (see there for more context and discussion)

Currently each stack application is reconciled independently of each other. When a user upgrades the version of a group of linked resources (Kibana, ES, APMServer) this is violating our own documented stack upgrade procedure which clearly defines an upgrade order. The problems described in #2426 and #2353 would be avoided if the upgrade order would be enforced by ECK:

1. upgrade Elasticsearch
2. upgrade Kibana
3. upgrade APMServer

For that the Kibana and APM controllers would have to 
* inspect the linked Elasticsearch resource, and compare the specified version with their own versions
* find a way of determining whether or not the version/generation specified is also the version/generation currently rolled out
* delay their own rollout of the version upgrade until ES and all predecessors in the upgrade graph  have completed their upgrade

Problems/Questions:
* how do we express the upgrade order? APMServer does not hold a reference to the related Kibana instance. How does it know the state of any related Kibana node? 
* how do we know that a given spec is rolled out without looking at individual pods. Can we rely on an extended status subresource that somehow expresses the reconciliation state?

sebgl:
One thing that was mentioned on https://github.com/elastic/cloud-on-k8s/issues/2353#issuecomment-574358217 is to reject (as opposed to delay) the version upgrade at the validation webhook level. I think we should not go with that approach since it makes the user experience a bit painful. As a user I would expect I can update both Elasticsearch and Kibana version at the same time in the YAML manifest, then let ECK deal with it.

---

A while ago we tried to design the association controllers in such a way that they may run with different managed namespaces contexts. For example the Kibana controller may not have access to the Elasticsearch namespace, so it should not deal with Elasticsearch resources at all. I'm not sure where we stand now regarding this "constraint", but I think we should aim at keeping the various controllers responsibilities decoupled so that, for example:
- the Kibana-ES association controller does not have to deal with Elasticsearch Pods
- the APM controller does not have to inspect a Kibana resource (nor Kibana Pods)

Here is what I have in mind, I think it's pretty close to ideas @pebrc expressed in the first post:

**Currently running version reflected in the status**

[...]

PR #3489, opened 2020-07-20
Report the lowest running version in the status of each resource

sebgl:
Report the currently running version in the `status.version` field of
each resource. In case a version upgrade is in progress, report the
lowest running version.

For Elasticsearch, we check both the Pods and the StatefulSets to report
the lowest version. A StatefulSet might report a higher version whereas
the Pods haven't been upgraded yet. It may also report a lower version
and no Pods are running any more with that version.
For other resources, we just check the Pods.

Parts of the Status is refactored into a common `DeploymentStatus` for
Kibana, ApmServer and EnterpriseSearch. Elasticsearch and Beat have
different requirements which cannot fit this common place.

This is a pre-req for strict stack version upgrade ordering
(https://github.com/elastic/cloud-on-k8s/issues/2600).


pkg/controller/elasticsearch/reconcile/resources_state.go
```diff
@@ -25,6 +27,8 @@ type ResourcesState struct {
 	CurrentPodsByPhase map[corev1.PodPhase][]corev1.Pod
 	// DeletingPods are all deleted Elasticsearch pods.
 	DeletingPods []corev1.Pod
+	// StatefulSets are all existing StatefulSets for the cluster.
+	StatefulSets sset.StatefulSetList
```
sebgl, line 31: It may feel a bit strange to retrieve StatefulSets here for the only purpose of reporting the version in the status, but I think it's still way cleaner than propagating the k8s client to the function that updates the status data, that did not really need it.
I'm also thinking we should invest more in this struct, instead of retrieving the Pods and StatefulSets several times in a single reconciliation. See https://github.com/elastic/cloud-on-k8s/issues/3490.

docs/reference/api-docs.asciidoc
```diff
@@ -265,6 +265,8 @@ ConfigSource references configuration settings.
 |===
 
 
+
+
```
david-kow, line 269: nit: remove?

hack/manifest-gen/assets/charts/eck/crds/all-crds.yaml
```diff
@@ -2570,10 +2591,16 @@ spec:
...
             availableNodes:
+              description: AvailableNodes is the number of available instances.
               format: int32
               type: integer
             health:
-              description: KibanaHealth expresses the status of the Kibana instances.
+              description: Health of the deployment.
+              type: string
+            version:
+              description: 'Version of the stack resource currently running. During
```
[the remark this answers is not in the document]
sebgl: I think because it is spread over more than 1 line. I tested with a single line in the comment: quotes are removed.
So I think it's fine.

docs/reference/api-docs.asciidoc
sebgl: I don't know where it comes from, it auto-generates this way :|
I see other places in this file with a lot of empty lines.
@charith-elastic any idea?

anyasabo: They don't get rendered IIRC so I'm not sure it's worth spending too much time on

pkg/controller/common/version/version.go
```diff
@@ -130,6 +132,32 @@ func (v *Version) IsAfter(other Version) bool {
 		(v.Major == other.Major && v.Minor == other.Minor && v.Patch > other.Patch)
 }
 
+// MinInPods returns the lowest version parsed from labels in the given Pods.
+func MinInPods(pods []corev1.Pod, labelName string) (*Version, error) {
```
pebrc, line 136: Should this also replace the code here: 
https://github.com/elastic/cloud-on-k8s/blob/764c4fb41cf8721d911d2e83aee481bb2a9095f0/pkg/controller/elasticsearch/label/label.go#L85-L101

?

pkg/controller/common/version/version_test.go
```diff
@@ -373,3 +376,182 @@ func TestFromLabels(t *testing.T) {
...
+			name: "empty list of StatefulSets",
+			args: args{
+				ssets:     nil,
+				labelName: "version-label",
+			},
+			want:    nil,
+			wantErr: false,
+		},
+	}
+	for _, tt := range tests {
```
pebrc, line 545: Nice, very easy to read. We should probably clean up the verbose ones I wrote above at some point (not necessarily this PR)

sebgl:
> For Elasticsearch, we check both the Pods and the StatefulSets to report
the lowest version. A StatefulSet might report a higher version whereas
the Pods haven't been upgraded yet. It may also report a lower version
and no Pods are running any more with that version.

> Isn't this the case for Deployments/DaemonSets too?

I think it is slightly different because we reconcile a single DaemonSet/Deployment per resource, whereas we have multiple StatefulSets per Elasticsearch resource:

* We may be in a situation where a Deployment reports version N+1 and no Pods are running. In this case we keep the last observed version (eg. N). Once the Pods are running again they should be created in version N+1 so we can just use it at this point in time. I don't see a case where inspecting the Pods only would cause a version mismatch.
* When dealing with StatefulSets we may have 2/3 StatefulSets updated to version N+1. If Pods of the third StatefulSet are temporarily not running, all Pods may report version N+1, but Pods of that 3rd StatefulSet may be recreated in version N anytime soon. So we should not report version N+1 until all Pods and StatefulSets report N+1.

pkg/utils/k8s/k8sutils.go
```diff
@@ -104,3 +104,12 @@ func DeleteSecretMatching(c Client, opts ...client.ListOption) error {
 	}
 	return nil
 }
+
+// PodsMatchingLabels returns Pod from the given namespace matching the given labels.
```
thbkrkr, line 108:
```suggestion
// PodsMatchingLabels returns Pods from the given namespace matching the given labels.
```

pkg/controller/common/version/version.go
sebgl: Oh right! I totally missed this.

pkg/apis/common/v1/common.go
```diff
@@ -14,9 +14,27 @@ import (
...
+type DeploymentHealth string
+
+const (
+	GreenHealth DeploymentHealth = "green"
+	RedHealth   DeploymentHealth = "red"
+)
+
+// DeploymentStatus represents status information about a deployment.
+type DeploymentStatus struct {
+	// AvailableNodes is the number of available instances.
```
thbkrkr, line 26: Not a strong opinion on this but I find it a bit suspicious and confusing that we are mixing `nodes` and `instances` here. Note that a deployment officially uses the term `replicas`. Wouldn't it be more accurate?

pkg/controller/common/version/version_test.go
sebgl: Do you have one (or several) in mind specifically?

pkg/apis/common/v1/common.go
sebgl: 👍 
I changed the comment to `// AvailableNodes is the number of replicas in the deployment.`
I think we want to keep `AvailableNodes` for backward-compatibility reasons (and consistency with Elasticsearch).

pkg/controller/common/version/version_test.go
pebrc: The `TestMin` one that uses a lot of vertical space for the `Version` declarations which could be expressed in a more compact form for readability. But it is really not that important.

Issue #3413, opened 2020-07-09
E2E test failure: unexpected end of JSON input in e2e-generate-xml


sebgl:
https://devops-ci.elastic.co/job/cloud-on-k8s-e2e-tests-stack-versions/140//testReport

```
[2020-07-09T04:37:49.597Z] ERROR failed to parse test output: {"Time":"2020-07-09T04:35:15.722666919Z","Action":"output","Package":"github.com/elastic/cloud-on-k8s/test/e2e/kb","Test":"TestKillKibanaPod/ES_endpoint_should_eventually_be_reachable#01","Output":"    --- PASS: TestKillKibanaPod/ES_endpoint_should_eventually_be_reacha: unexpected end of JSON input
[2020-07-09T04:37:49.597Z] make: *** [Makefile:400: e2e-generate-xml] Error 3
[2020-07-09T04:37:49.597Z] Makefile:45: recipe for target 'ci-internal' failed
```

Looking at the actual tests logs I don't see anything wrong with the JSON formatting around the log line mentioned above:

```
[2020-07-09T04:35:23.252Z] {"Time":"2020-07-09T04:35:15.722666919Z","Action":"output","Package":"github.com/elastic/cloud-on-k8s/test/e2e/kb","Test":"TestKillKibanaPod/ES_endpoint_should_eventually_be_reachable#01","Output":"    --- PASS: TestKillKibanaPod/ES_endpoint_should_eventually_be_reachable#01 (0.04s)\n"}
[2020-07-09T04:35:23.252Z] {"Time":"2020-07-09T04:35:15.72267163Z","Action":"pass","Package":"github.com/elastic/cloud-on-k8s/test/e2e/kb","Test":"TestKillKibanaPod/ES_endpoint_should_eventually_be_reachable#01","Elapsed":0.04}
[2020-07-09T04:35:23.253Z] {"Time":"2020-07-09T04:35:15.722675148Z","Action":"output","Package":"github.com/elastic/cloud-on-k8s/test/e2e/kb","Test":"TestKillKibanaPod/Kibana_deployment_be_rolled_out#01","Output":"    --- FAIL: TestKillKibanaPod/Kibana_deployment_be_rolled_out#01 (300.00s)\n"}
[...]

PR #3499, opened 2020-07-21
Fix log stream flush in E2E tests

sebgl:
This commit fixes several problems:
- the main e2e tests process could return before the log stream
  goroutine has enough time to run `defer jsonLog.Close()`, leading to a
  truncated json file
- the informer was stopped (stopping the tests) on any stream error, but
  we retry in case the stream stops so we should keep things running
- we did not wait for the end of the log stream in case the Pod is in
  failed state

We now rely on a `logStreamStopped` channel, and wait for it to be
closed before returning from the `monitorTestJob` function, so we know
all stream writers have been properly closed.
Also streamStatus is now only used for reporting errors, it does not
also indicate the end of the stream, which was a bit confusing.

I did not test this fixes our current tests results parsing issues
entirely (because not easy to reproduce), but I think it should help. We'll see
on the next set of e2e tests :)

Relates https://github.com/elastic/cloud-on-k8s/issues/3413.

### DIFF
--- a/test/e2e/cmd/run/run.go
+++ b/test/e2e/cmd/run/run.go
@@ -484,7 +484,7 @@ func (h *helper) monitorTestJob(client *kubernetes.Clientset) error {
 	return err
 }
 
-func (h *helper) streamTestJobOutput(streamStatus chan<- error, stop <-chan struct{}, client *kubernetes.Clientset, pod string) {
+func (h *helper) streamTestJobOutput(streamErrors chan<- error, stop <-chan struct{}, client *kubernetes.Clientset, pod string) {
 	outputs := []io.Writer{os.Stdout}
 	if h.logToFile {
 		jl, err := newJSONLogToFile(testsLogFile)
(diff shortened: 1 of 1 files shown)


test/e2e/cmd/run/run.go
```diff
@@ -433,14 +435,16 @@ func (h *helper) monitorTestJob(client *kubernetes.Clientset) error {
 				if !jobStarted {
 					jobStarted = true
 					log.Info("Pod started", "name", newPod.Name)
-					go h.streamTestJobOutput(streamStatus, stopLogStream, client, newPod.Name)
+					go func() {
+						h.streamTestJobOutput(streamErrors, stopLogStream, client, newPod.Name)
```
pebrc, line 439: Maybe rename the arg in `streamTestJobOutput` to `streamErrors` too?

test/e2e/cmd/run/run.go
```diff
@@ -478,6 +479,8 @@ func (h *helper) monitorTestJob(client *kubernetes.Clientset) error {
 	})
 
 	informer.Run(ctx.Done())
+	// wait for the log stream to be fully flushed
+	<-logStreamStopped
```
pebrc, line 483: I am curious why did you use a channel instead of a `WaitGroup` here?

sebgl: No reason. I can totally use a `WaitGroup`, it probably makes the intent clearer 👍

PR #3505, opened 2020-07-22
Add EnterpriseSearch example to OLM template

charith-elastic:
Add EnterpriseSearch to the examples list and change the config file to point to 1.2.0.


charith-elastic:
Jenkins test this please

Issue #3426, opened 2020-07-14
E2E tests should not fail fast


david-kow:
Failing fast impacts test history and hides all but the first test failure. We should disable failing fast.

barkbay:
Closing this one as we finally concluded that disabling failing fast may raise some other problems, see https://github.com/elastic/cloud-on-k8s/pull/3507#pullrequestreview-455616569

PR #3507, opened 2020-07-22
Stop failing fast in E2E tests

david-kow:
Fixes https://github.com/elastic/cloud-on-k8s/issues/3426.

### DIFF
--- a/test/e2e/run.sh
+++ b/test/e2e/run.sh
@@ -4,19 +4,16 @@
 # or more contributor license agreements. Licensed under the Elastic License;
 # you may not use this file except in compliance with the Elastic License.
 
-# Go's -failfast flag does not prevent tests from other packages from being executed if the package
-# list is specified using the "./..." form. This script works around that limitation so that CI jobs
-# fail faster.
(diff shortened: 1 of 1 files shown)


david-kow:
I don't expect dozens of tests failing, but even if they do fail then we'll have more information about the run than today.

sebgl:
I think @barkbay raises a good point here... If one or two tests fail at the beginning of the suite, then it's likely for all following tests to fail for lack of resources, leading to I don't know how many hundreds of tests failures in the suite, making it hard to find the real failure.
Does it mean we should cleanup deployments for failed tests? It's pretty useful to have a dump of the current state of the cluster when a test fails.

Based on that I think I would be -1 for this change under the current conditions.

david-kow:
@sebgl by my napkin math we would need to see ~10 test failures before we start running out of resources, so I wouldn't consider it as a big concern. But I do agree that even with a single failure we don't have a good way to get to the cluster state as we do the `eck-dump.sh` only after Job pod is done. I also don't think it's very simple to fix, so I'll close this PR.

Issue #3344, opened 2020-06-29
Kibana 8.0 gets OOMKilled during initialization with default (1Gi) memory limit 


david-kow:
With 2Gi it runs fine. We should consider changing the default only for 8.0+ to avoid recreating Kibana pods without user specified resources.

anyasabo:
This appears to not be intended behavior, so I opened https://github.com/elastic/kibana/issues/72987

sebgl:
This doesn't seem to be a problem anymore.
See https://github.com/elastic/kibana/issues/72987#issuecomment-682445267.

I think we ran some weird tests setup with a combined Elasticsearch 7.8.0 + Kibana 8.0.0 (which ECK refuses now, and Kibana as well).
Running ES 8.0.0-SNAPSHOT + Kibana 8.0.0-SNAPSHOT seems to work as expected, I don't observe any OOM on Kibana.

PR #3514, opened 2020-07-22
Set Kibana v8 default ram to 2Gi

anyasabo:
Fix https://github.com/elastic/cloud-on-k8s/issues/3344


anyasabo:
> Have we double-checked this with Kibana team first? It could be a bug in Kibana 8.0.0-SNAPSHOT.

I didn't at least, I'll repro the original issue and ask

anyasabo:
Closing this for now as I misunderstood the original issue. Kibana team let me know this is not expected. Opened https://github.com/elastic/kibana/issues/72987

PR #3524, opened 2020-07-23
Fix memory corruption from pod template validation

### LINKED ISSUE #3421: Elasticsearch init container conflict with Istio CNI init container
When Istio CNI is enabled, Elasticsearch fails to start with the following error:

```
/usr/local/bin/docker-entrypoint.sh: line 49: /usr/share/elasticsearch/bin/elasticsearch-env-from-file: No such file or directory
```

This appears to be a conflict between the Istio CNI validation init container and the Elasticsearch file system prep init container. Requires further investigation to ascertain the root cause.

charith-elastic:
Full description of the bug: https://github.com/elastic/cloud-on-k8s/issues/3421#issuecomment-663057459

Fixes #3421


charith-elastic:
Jenkins test this please

anyasabo:
Nit: does it maybe make sense to move the deepcopy into `validatePodTemplate` so callers don't need to know the sset will be modified? Right now there's only one caller though so I'm not sure it matters.

charith-elastic:
Moving it to `validatePodTemplate` would cause the deep copy to be called several times -- which seems wasteful.

PR #3525, opened 2020-07-23
Init containers image defaulting

### LINKED ISSUE #3453: InitContainer image defaulting is not consistent
Elasticsearch resource will default init containers images to the main container image, while all other resources will only do that when (unrelated) secure settings are used. This is due to the fact that `WithInitContainerDefaults()` is always called only on ES reconciliation path. In all others, it's called conditionally when secure settings are present.

thbkrkr:
This commit makes sure that user provided init containers inherit from the default values including the main container image.

I refactored a bit the functions that build the PodTemplateSpec to try to make them boring and always have the same pattern by preparing the volumes, volumeMounts, initContainers and then only call:

```go
		WithVolumes(volumes...).
		WithInitContainerDefaults()
```

Resolves #3453.

### DIFF
--- a/pkg/controller/apmserver/pod.go
+++ b/pkg/controller/apmserver/pod.go
@@ -106,28 +106,32 @@ func newPodSpec(as *apmv1.ApmServer, p PodSpecParams) corev1.PodTemplateSpec {
 
 	ports := getDefaultContainerPorts(*as)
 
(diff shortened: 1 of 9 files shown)


thbkrkr:
jenkins test this please

thbkrkr:
jenkins test this please

thbkrkr:
jenkins test this please

PR #3536, opened 2020-07-27
[1.2] Check if CA Secret is expected before validating its name (#3527)

david-kow:
 - Check if CA Secret is expected before validating its name (#3527)


david-kow:
jenkins test this please

Issue #3535, opened 2020-07-27
Misleading log statements in association controller


david-kow:
The controller handling both Kibana and ES associations has logging statements that are confusing. E.g.:

https://github.com/elastic/cloud-on-k8s/blob/0d31c4ae18ac652d237c451d8b73758ce4c94002/pkg/controller/association/reconciler.go#L315

https://github.com/elastic/cloud-on-k8s/blob/0d31c4ae18ac652d237c451d8b73758ce4c94002/pkg/controller/association/reconciler.go#L320

This leads to logs similar to the below. Notice Kibana assoc logging about ES config.
`{"log.level":"info","@timestamp":"2020-07-23T19:27:17.235Z","log.logger":"association.beat-kibana-association-controller","message":"Updating spec with Elasticsearch association configuration","service.version":"1.2.0-51ae5fd4","service.type":"eck","ecs.version":"1.4
.0","namespace":"monitoring","beat_name":"log"}`

sebgl:
Will be fixed as part of https://github.com/elastic/cloud-on-k8s/pull/3537.

Issue #2862, opened 2020-04-10
Reduce the number of fields used in log messages


pebrc:
We use different fields for `es_name`, `apm_name`, `kb_name`, in other log messages we use additional fields like `stateful_set` or just `value`. 

I would suggest to settle on a few generic fields, for the following reasons:

* avoid mapping explosion on ES in on ingest 
* make it easier to search and display relevant fields in Kibana when looking at logs. It is really hard at moment to look at operator logs in Kibana and decide which fields to include. 

We could use something like `name`, `value` and `type` to express the the name and type of resource and additional relevant values. Also double check in ECS if there are fields we can use.

anyasabo:
We considered that when we were first starting it, but then it became complicated when there were multiple types in the same message. For example:

`log.Error(err, "failed to delete cluster license secret", "secret_name", secretName, "namespace", cluster.Namespace, "es_name", cluster.Name)`

That said overall I'm hugely in favor of this, and maybe there's a way to work around those specific cases.

sebgl:
Relates https://github.com/elastic/cloud-on-k8s/issues/2491.

I think our long-term goal could be to call something like `log.Ctx(ctx).Info()`, where the context already holds the reconciled resource details (namespace, name, kind, etc.) and the iteration number.

PR #3537, opened 2020-07-27
Enforce a strict upgrade order between stack components

sebgl:
Ensure individual stack resources can only have their version upgraded if referenced resources are also upgraded.
Implicitly, this implements the following upgrade order:
- Elasticsearch > Kibana > Beats, APM
- Elasticsearch > EnterpriseSearch

Example: if a user upgrades both Elasticsearch and Kibana from 7.7.0 to 7.8.0, the Kibana upgrade is delayed until Elasticsearch runs 7.8.0.

This restriction only applies to major and minor versions. Patch versions are ignored: Kibana 7.8.1 can be deployed alongside Elasticsearch 7.8.0.

In order to do so, the association controller adds a `version` field to the existing associationConf annotation, in the resource holding associations.
Example: the ApmServer resource that specifies a reference to Kibana and Elasticsearch will have the (lowest) currently running Kibana version and currently running Elasticsearch version in its annotations, set by both association controllers.

Each resource controller can then inspect its own association configurations annotations, and requeue in case the referenced resource is running an older version. Or if its version is not reported yet (resource in the process of being created): this slows down the initial resource creation slightly (Kibana is not created before the first Elasticsearch Pod is created), but I think that's short enough to not bother much.

We do not try to be smart in decoupling version upgrades from other upgrades (for example: decide to apply a config change but not a version change yet even though the spec specifies both): this would be tricky to implement.

**Breaking change**: we used to allow several associated resources to coexist with different versions, and apply any spec change immediately. We now enforce a strict version upgrade ordering that delays any spec changes until the resource version allows it. For example: when updating both Elasticsearch and Kibana from 7.7.0 to 7.8.0, any change to the Kibana spec will be delayed until Elasticsearch has been successfully updated to 7.8.0. Only `major.minor` in the version is considered by this.

Fixes https://github.com/elastic/cloud-on-k8s/issues/2600.
Fixes https://github.com/elastic/cloud-on-k8s/issues/3535.
Depends on https://github.com/elastic/cloud-on-k8s/pull/3534 for the E2E test to pass.

### DIFF
--- a/pkg/apis/common/v1/association.go
+++ b/pkg/apis/common/v1/association.go
@@ -67,13 +67,16 @@ type Association interface {
 	SetAssociationStatus(status AssociationStatus)
 }
 
-// AssociationConf holds the association configuration of an Elasticsearch cluster.
+// AssociationConf holds the association configuration of a referenced resource in an association.
 type AssociationConf struct {
 	AuthSecretName string `json:"authSecretName"`
 	AuthSecretKey  string `json:"authSecretKey"`
 	CACertProvided bool   `json:"caCertProvided"`
 	CASecretName   string `json:"caSecretName"`
 	URL            string `json:"url"`
+	// Version of the referenced resource. If a version upgrade is in progress,
+	// matches the lowest running version. May be empty if unknown.
+	Version string `json:"version"`
 }
 
 // IsConfigured returns true if all the fields are set.
@@ -138,4 +141,11 @@ func (ac *AssociationConf) GetURL() string {
 		return ""
 	}
 	return ac.URL
+}
+
+func (ac *AssociationConf) GetVersion() string {
+	if ac == nil {
+		return ""
+	}
+	return ac.Version
 }
--- a/pkg/controller/association/conf.go
+++ b/pkg/controller/association/conf.go
@@ -107,7 +107,7 @@ func ElasticsearchAuthSettings(c k8s.Client, association commonv1.Association) (
 	return assocConf.AuthSecretKey, string(data), nil
 }
 
-// AllowVersion returns true if the given resourceVersion is lower or equal to the associations version.
+// AllowVersion returns true if the given resourceVersion is lower or equal to the associations' versions.
 // For example: Kibana in version 7.8.0 cannot be deployed if its Elasticsearch association reports version 7.7.0.
 // Referenced resources version is parsed from the association conf annotation.
 func AllowVersion(resourceVersion version.Version, associated commonv1.Associated, logger logr.Logger) bool {
--- a/pkg/controller/association/conf.go
+++ b/pkg/controller/association/conf.go
@@ -76,6 +76,8 @@ func IsConfiguredIfSet(association commonv1.Association, r record.EventRecorder)
 			"kind", association.GetObjectKind().GroupVersionKind().Kind,
 			"namespace", association.GetNamespace(),
 			"name", association.GetName(),
+			"ref_namespace", ref.Namespace,
+			"ref_name", ref.Name,
 		)
 		return false
 	}
--- a/pkg/controller/association/conf.go
+++ b/pkg/controller/association/conf.go
@@ -110,7 +110,7 @@ func ElasticsearchAuthSettings(c k8s.Client, association commonv1.Association) (
 
 // AllowVersion returns true if the given resourceVersion is lower or equal to the associations' versions.
 // For example: Kibana in version 7.8.0 cannot be deployed if its Elasticsearch association reports version 7.7.0.
-// A difference in the patch version is ignored: Kibana 7.8.1 can be deployed alongside Elasticsearch 7.8.0.
+// A difference in the patch version is ignored: Kibana 7.8.1+ can be deployed alongside Elasticsearch 7.8.0.
 // Referenced resources version is parsed from the association conf annotation.
 func AllowVersion(resourceVersion version.Version, associated commonv1.Associated, logger logr.Logger, recorder record.EventRecorder) bool {
 	for _, assoc := range associated.GetAssociations() {
--- a/pkg/controller/association/conf.go
+++ b/pkg/controller/association/conf.go
@@ -126,7 +126,7 @@ func AllowVersion(resourceVersion version.Version, associated commonv1.Associate
 		}
 		refVer, err := version.Parse(assoc.AssociationConf().Version)
 		if err != nil {
-			logger.Error(err, "Invalid version found in association conf", "association_version", assoc.AssociationConf().Version)
+			logger.Error(err, "Invalid version found in association configuration", "association_version", assoc.AssociationConf().Version)
 			return false
 		}
 		if !refVer.IsSameOrAfter(resourceVersion) {
--- a/pkg/controller/association/conf.go
+++ b/pkg/controller/association/conf.go
@@ -14,7 +14,6 @@ import (
 	"github.com/pkg/errors"
 	"go.elastic.co/apm"
 	corev1 "k8s.io/api/core/v1"
-	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
@@ -69,7 +68,7 @@ func IsConfiguredIfSet(association commonv1.Association, r record.EventRecorder)
 	if (&ref).IsDefined() && !association.AssociationConf().IsConfigured() {
 		r.Event(
 			association,
-			v1.EventTypeWarning,
+			corev1.EventTypeWarning,
 			events.EventAssociationError,
 			"Association backend for "+association.AssociatedType()+" is not configured",
 		)
@@ -95,7 +94,7 @@ func ElasticsearchAuthSettings(c k8s.Client, association commonv1.Association) (
 	}
 
 	secretObjKey := types.NamespacedName{Namespace: association.GetNamespace(), Name: assocConf.AuthSecretName}
-	var secret v1.Secret
+	var secret corev1.Secret
 	if err := c.Get(secretObjKey, &secret); err != nil {
 		return "", "", err
 	}
--- a/pkg/controller/association/conf.go
+++ b/pkg/controller/association/conf.go
@@ -7,6 +7,7 @@ package association
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"reflect"
 	"unsafe"
 
@@ -131,9 +132,11 @@ func AllowVersion(resourceVersion version.Version, associated commonv1.Associate
 		if !refVer.IsSameOrAfter(resourceVersion) {
 			// the version of the referenced resource (example: Elasticsearch) is lower than
 			// the desired version of the reconciled resource (example: Kibana)
-			msg := "Delaying version deployment since an associated resource is not upgraded yet"
-			logger.Info(msg, "version", resourceVersion, "ref_namespace", assocRef.Namespace, "ref_name", assocRef.Name, "ref_version", refVer)
-			recorder.Event(associated, corev1.EventTypeWarning, events.EventReasonDelayed, msg)
+			logger.Info("Delaying version deployment since a referenced resource is not upgraded yet",
+				"version", resourceVersion, "ref_version", refVer,
+				"ref_type", assoc.AssociatedType(), "ref_namespace", assocRef.Namespace, "ref_name", assocRef.Name)
+			recorder.Event(associated, corev1.EventTypeWarning, events.EventReasonDelayed,
+				fmt.Sprintf("Delaying version deployment since the referenced %s is not upgraded yet", assoc.AssociatedType()))
 			return false
 		}
 	}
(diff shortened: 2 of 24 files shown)


pkg/controller/enterprisesearch/enterprisesearch_controller_test.go
```diff
@@ -280,6 +282,57 @@ func TestReconcileEnterpriseSearch_Reconcile_Create_Update_Resources(t *testing.
 	checkResources()
 }
 
+func TestReconcileEnterpriseSearch_doReconcile_AssociationDelaysVersionUpgrade(t *testing.T) {
```
sebgl, line 285: Enterprise Search is the only driver with great unit test coverage, so I also added a test there.
Other controllers don't really have unit tests for their main reconciliation algorithm, but E2E tests do cover it in a different way.

sebgl:
Jenkins test this please

test/e2e/stack_test.go
```diff
@@ -0,0 +1,219 @@
...
+	apm := apmserver.NewBuilder("apm").
+		WithNodeCount(1).
+		WithVersion(initialVersion).
+		WithElasticsearchRef(esRef).
+		WithKibanaRef(kbRef)
+	apmUpdated := apm.WithVersion(updatedVersion)
+	ent := enterprisesearch.NewBuilder("ent").
+		WithNodeCount(1).
+		WithVersion(initialVersion).
+		WithElasticsearchRef(esRef)
```
barkbay, line 67: `WithRestrictedSecurityContext` might be needed in a few places:

```suggestion
		WithElasticsearchRef(esRef).
		WithRestrictedSecurityContext()
```

pkg/controller/association/conf.go
```diff
@@ -103,6 +105,38 @@ func ElasticsearchAuthSettings(c k8s.Client, association commonv1.Association) (
...
+	for _, assoc := range associated.GetAssociations() {
+		assocRef := assoc.AssociationRef()
+		if !assocRef.IsDefined() {
+			// no association specified, move on
+			continue
+		}
+		if assoc.AssociationConf() == nil || assoc.AssociationConf().Version == "" {
+			// no conf reported yet, this may be the initial resource creation
+			logger.Info("Delaying version deployment since the version of an associated resource is not reported yet",
+				"version", resourceVersion, "ref_namespace", assocRef.Namespace, "ref_name", assocRef.Name)
```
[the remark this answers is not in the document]
sebgl: I think it's better to tackle this completely in https://github.com/elastic/cloud-on-k8s/issues/2862, so we agree on the right fields and schema there?

I'm adding the extra fields in `IsConfiguredIfSet` 👍

pkg/apis/common/v1/association.go
```diff
@@ -67,13 +67,16 @@ type Association interface {
 	SetAssociationStatus(status AssociationStatus)
 }
 
-// AssociationConf holds the association configuration of an Elasticsearch cluster.
+// AssociationConf holds the association configuration of a referenced resource in an association.
```
david-kow, line 70: test

pkg/controller/association/conf.go
```diff
@@ -103,6 +107,38 @@ func ElasticsearchAuthSettings(c k8s.Client, association commonv1.Association) (
...
+		}
+		refVer, err := version.Parse(assoc.AssociationConf().Version)
+		if err != nil {
+			logger.Error(err, "Invalid version found in association conf", "association_version", refVer)
+			return false
+		}
+		if !refVer.IsSameOrAfter(resourceVersion) {
+			// the version of the referenced resource (example: Elasticsearch) is lower than
+			// the desired version of the reconciled resource (example: Kibana)
+			logger.Info("Delaying version deployment since an associated resource is not upgraded yet",
```
david-kow, line 134: I think we should we also post events here too. If the user just applied higher version on ref holder, then nothing is going to happen. I think the first place they're going to look at for explanation is resource events.

pkg/controller/association/conf.go
```diff
@@ -103,6 +107,38 @@ func ElasticsearchAuthSettings(c k8s.Client, association commonv1.Association) (
 	return assocConf.AuthSecretKey, string(data), nil
 }
 
+// AllowVersion returns true if the given resourceVersion is lower or equal to the associations version.
```
pebrc, line 110:
```suggestion
// AllowVersion returns true if the given resourceVersion is lower or equal to the association's version.
```
or 
```suggestion
// AllowVersion returns true if the given resourceVersion is lower or equal to the associations' versions.
```

pkg/controller/association/conf.go
```diff
@@ -103,6 +107,38 @@ func ElasticsearchAuthSettings(c k8s.Client, association commonv1.Association) (
...
+		}
+		if assoc.AssociationConf() == nil || assoc.AssociationConf().Version == "" {
+			// no conf reported yet, this may be the initial resource creation
+			logger.Info("Delaying version deployment since the version of an associated resource is not reported yet",
+				"version", resourceVersion, "ref_namespace", assocRef.Namespace, "ref_name", assocRef.Name)
+			return false
+		}
+		refVer, err := version.Parse(assoc.AssociationConf().Version)
+		if err != nil {
+			logger.Error(err, "Invalid version found in association conf", "association_version", refVer)
```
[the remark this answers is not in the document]
sebgl: Great catch!

pkg/controller/association/conf.go
```diff
@@ -103,6 +107,38 @@ func ElasticsearchAuthSettings(c k8s.Client, association commonv1.Association) (
 	return assocConf.AuthSecretKey, string(data), nil
 }
 
+// AllowVersion returns true if the given resourceVersion is lower or equal to the associations version.
+// For example: Kibana in version 7.8.0 cannot be deployed if its Elasticsearch association reports version 7.7.0.
+// Referenced resources version is parsed from the association conf annotation.
+func AllowVersion(resourceVersion version.Version, associated commonv1.Associated, logger logr.Logger) bool {
```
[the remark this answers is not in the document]
sebgl: Hm same as https://github.com/elastic/cloud-on-k8s/pull/3537 I would consider this to be an improvement/fix about something that should not have worked initially.

pkg/controller/association/conf.go
sebgl: https://github.com/elastic/cloud-on-k8s/issues/3538 should also help, but I agree. I'll add the event 👍

pkg/controller/association/conf.go
```diff
@@ -103,6 +107,39 @@ func ElasticsearchAuthSettings(c k8s.Client, association commonv1.Association) (
...
+			logger.Info("Delaying version deployment since the version of an associated resource is not reported yet",
+				"version", resourceVersion, "ref_namespace", assocRef.Namespace, "ref_name", assocRef.Name)
+			return false
+		}
+		refVer, err := version.Parse(assoc.AssociationConf().Version)
+		if err != nil {
+			logger.Error(err, "Invalid version found in association conf", "association_version", assoc.AssociationConf().Version)
+			return false
+		}
+		if !refVer.IsSameOrAfter(resourceVersion) {
```
anyasabo, line 131: This is stricter than required, correct? https://www.elastic.co/support/matrix#matrix_compatibility

They're forwards and backwards compatible within the same minor (i.e. Kibana 7.8.1 is compatible with ES 7.8.0), but here we are checking the patch version too.

pkg/controller/association/conf.go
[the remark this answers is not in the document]
david-kow: > Hm same as #3537 I would consider this to be an improvement/fix about something that should not have worked initially. 

You linked to this PR :) 
It seems to me that this is a technicality and while I agree we could argue that this is a bug fix and this was always the intended behaviour the user experience will be bad. 

As soon as this is merged and without being marked as `>breaking` it's no longer discoverable in any way. So user that's running Kibana/APM higher than ES has changes to Kibana/APM effectively disabled (including config, cert, securesettings changes) until they either rollback ECK (I'm not even sure if this just works) or update ES (might not be possible at all for their own reasons). 

I think it doesn't cost us anything to document this change, but it can damage the experience for some users significantly.

pkg/controller/association/conf.go
```diff
@@ -125,14 +125,15 @@ func AllowVersion(resourceVersion version.Version, associated commonv1.Associate
...
 			return false
 		}
 		if !refVer.IsSameOrAfter(resourceVersion) {
 			// the version of the referenced resource (example: Elasticsearch) is lower than
 			// the desired version of the reconciled resource (example: Kibana)
-			logger.Info("Delaying version deployment since an associated resource is not upgraded yet",
-				"version", resourceVersion, "ref_namespace", assocRef.Namespace, "ref_name", assocRef.Name, "ref_version", refVer)
+			msg := "Delaying version deployment since an associated resource is not upgraded yet"
+			logger.Info(msg, "version", resourceVersion, "ref_namespace", assocRef.Namespace, "ref_name", assocRef.Name, "ref_version", refVer)
+			recorder.Event(associated, corev1.EventTypeWarning, events.EventReasonDelayed, msg)
```
david-kow, line 136: Isn't this lacking details? I'm not sure if users would immediately jump from `associated resource` to looking at `elasticsearchRef` or `kibanaRef`.

pkg/controller/association/conf.go
sebgl: That's a good question. I was mostly following the [stack upgrade process](https://www.elastic.co/guide/en/elastic-stack/current/upgrading-elastic-stack.html#upgrade-order-elastic-stack), but the matrix indeed shows backward compatibility with the same minor (and with the same major in most cases).
I'll try to get more details about the right thing to do.

pkg/controller/association/conf.go
sebgl: 👍 changing the event msg to `fmt.Sprintf("Delaying version deployment since the referenced %s is not upgraded yet", assoc.AssociatedType()))`.

pkg/controller/association/reconciler.go
```diff
@@ -282,7 +293,7 @@ func (r *Reconciler) getElasticsearch(
 		if apierrors.IsNotFound(err) {
 			// ES is not found, remove any existing backend configuration and retry in a bit.
 			if err := RemoveAssociationConf(r.Client, association.Associated(), association.AssociationConfAnnotationName()); err != nil && !apierrors.IsConflict(err) {
-				r.log(association).Error(err, "Failed to remove Elasticsearch output from EnterpriseSearch object")
+				r.log(association).Error(err, "Failed to remove Elasticsearch association conf")
```
thbkrkr, line 296: Nit: for clarity and consistency.
```suggestion
				r.log(association).Error(err, "Failed to remove Elasticsearch association configuration")
```

pkg/controller/association/conf.go
```diff
@@ -103,6 +108,41 @@ func ElasticsearchAuthSettings(c k8s.Client, association commonv1.Association) (
...
+		}
+		if assoc.AssociationConf() == nil || assoc.AssociationConf().Version == "" {
+			// no conf reported yet, this may be the initial resource creation
+			logger.Info("Delaying version deployment since the version of an associated resource is not reported yet",
+				"version", resourceVersion, "ref_namespace", assocRef.Namespace, "ref_name", assocRef.Name)
+			return false
+		}
+		refVer, err := version.Parse(assoc.AssociationConf().Version)
+		if err != nil {
+			logger.Error(err, "Invalid version found in association conf", "association_version", assoc.AssociationConf().Version)
```
thbkrkr, line 129: Nit: for clarity and consistency.
```suggestion
			logger.Error(err, "Invalid version found in association configuration", "association_version", assoc.AssociationConf().Version)
```

pkg/controller/association/conf.go
sebgl: I asked around and we discussed this in a different channel: it makes sense to restrict the upgrade ordering to major and minors only, and relax it for different patch versions. In other terms: allow Kibana v7.8.1 alongside Elasticsearch 7.8.0, but not Kibana 7.9.0 along side Elasticsearch 7.8.1.

fdd6d99 fixes it.

pkg/controller/association/conf.go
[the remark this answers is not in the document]
sebgl: 👍  let's mark this as breaking.

sebgl:
Recent changes:
- marked the PR as `breaking`, and added a short description of the breaking change in the PR description
- emit an event when a resource upgrade is delayed
- ignore any difference in patch versions (example: allow Kibana 7.8.1 to exist alongside Elasticsearch 7.8.0)

This is ready for another review.

pkg/controller/association/conf.go
```diff
@@ -109,6 +110,7 @@ func ElasticsearchAuthSettings(c k8s.Client, association commonv1.Association) (
 
 // AllowVersion returns true if the given resourceVersion is lower or equal to the associations' versions.
 // For example: Kibana in version 7.8.0 cannot be deployed if its Elasticsearch association reports version 7.7.0.
+// A difference in the patch version is ignored: Kibana 7.8.1 can be deployed alongside Elasticsearch 7.8.0.
```
david-kow, line 113:
```suggestion
// A difference in the patch version is ignored: Kibana 7.8.1+ can be deployed alongside Elasticsearch 7.8.0.
```

Just to show it's not only +-1 patch.

PR #3543, opened 2020-07-28
Skip manifest upload on nightly build job

charith-elastic:
Skips uploading the `all-in-one.yaml` file during the nightly build. 
Without this change, any updates to the manifests on `master` will alter the installation file for the currently released version as well.

### DIFF
--- a/.ci/pipelines/build.Jenkinsfile
+++ b/.ci/pipelines/build.Jenkinsfile
@@ -41,6 +41,9 @@ pipeline {
                     }
                 }
                 stage('Upload YAML manifest to S3') {
+                    when {
+                        buildingTag()
+                    }
                     steps {
                         script {
                             env.VERSION = readFromEnvFile("VERSION")


.ci/pipelines/build.Jenkinsfile
```diff
@@ -41,6 +41,9 @@ pipeline {
                     }
                 }
                 stage('Upload YAML manifest to S3') {
+                    when {
+                        buildingTag()
+                    }
                     steps {
                         script {
                             env.VERSION = readFromEnvFile("VERSION")
```
pebrc, line 49: Should `$VERSION`  not point to `$CURRENT_RELEASE + 1` as soon as we branch off for the release and avoid the problem you described?

charith-elastic: Ideally, yes. However it's a manual process that's not clearly documented so we have not been very good at doing that promptly. 
Even if the `VERSION` file is correct, the generated manifest doesn't point to the correct container image so uploading that to S3 currently doesn't achieve much. I was in the process of creating an issue about publishing the nightly build so I will add this problem to the list of things to fix in that issue.

PR #3544, opened 2020-07-28
Update VERSION to 1.3.0-SNAPSHOT

charith-elastic:
Update the `VERSION` file to contain `1.3.0-SNAPSHOT`.

Adding a suffix to indicate that the code in `master` is pre-release quality would help avoid confusion in places where we use the value of `VERSION` to tag builds and upload artefacts etc.


VERSION
```diff
@@ -1 +1 @@
-1.2.0
+1.3.0-prerelease
```
pebrc, line 1: I believe the rest of Elastic is using `-SNAPSHOT` for this purpose. I wonder if we should do the same even though this is not a Java project.

PR #3549, opened 2020-07-28
Reflect currently running version from status.version in additionalPrinterColumns

sebgl:
When running for example `kubectl get elastic`, we get the version
displayed from `spec.version`. It represents the desired version but not
the version currently running.

Now that we have `status.version` available to represent the version
currently running, let's use it instead!

### DIFF
--- a/config/crds/all-crds.yaml
+++ b/config/crds/all-crds.yaml
@@ -6,21 +6,6 @@ metadata:
   creationTimestamp: null
   name: apmservers.apm.k8s.elastic.co
 spec:
-  additionalPrinterColumns:
-  - JSONPath: .status.health
-    name: health
-    type: string
-  - JSONPath: .status.availableNodes
-    description: Available nodes
-    name: nodes
-    type: integer
-  - JSONPath: .spec.version
-    description: APM version
-    name: version
-    type: string
-  - JSONPath: .metadata.creationTimestamp
-    name: age
-    type: date
   group: apm.k8s.elastic.co
   names:
     categories:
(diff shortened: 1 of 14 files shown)


config/crds/all-crds.yaml
```diff
@@ -6,21 +6,6 @@ metadata:
...
-    description: Available nodes
-    name: nodes
-    type: integer
-  - JSONPath: .spec.version
-    description: APM version
-    name: version
-    type: string
-  - JSONPath: .metadata.creationTimestamp
-    name: age
-    type: date
```
sebgl, line 23: `additionalPrinterColumns` from v1beta1 does not match `additionalPrinterColumns` from v1 anymore with this change.
I need to double-check this does not break crd upgrade on some environments... will try the `ECK 0.9 -> ECK from this PR` upgrade path with Openshift 3.11 (minishift) and report here.

sebgl:
> Should we mark this as breaking? It does change the behavior of already existing field.

Hm that's a good question. We could argue it should have been this way to start with and the previous version value was incorrect (so more like a bugfix/enhancement). @elastic/cloud-k8s any other opinion?

barkbay:
> > Should we mark this as breaking? It does change the behavior of already existing field.
> 
> Hm that's a good question. We could argue it should have been this way to start with and the previous version value was incorrect (so more like a bugfix/enhancement). @elastic/cloud-k8s any other opinion?

👍 , sounds like a fix to me

config/crds/all-crds.yaml
sebgl: Applying the new all-in-one on Openshift 3.11:

```
error: error validating "config/all-in-one.yaml": error validating data: [ValidationError(CustomResourceDefinition.spec.versions[0]): unknown field "additionalPrinterColumns" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1beta1.CustomResourceDefinitionVersion, ValidationError(CustomResourceDefinition.spec.versions[1]): unknown field "additionalPrinterColumns" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1beta1.CustomResourceDefinitionVersion]; if you choose to ignore these errors, turn validation off with --validate=false
```

:(

sebgl: I think some kustomize tweaks to ensure only the newest `additionalPrinterColumns` is set at the root level would fix this.

sebgl: 3c57b29 sets up the required kustomize patches in order to have a single top-level `additionalPrinterColumns `.

If this new CRD is used with an operator running an older version (whichis unlikely to happen), or during the first few seconds before a reconciliation happens with ECK 1.3, then the `version` field in `kubectl get elastic` has an empty value, which is an OK behaviour IMO.

I can correctly apply the all-in-one manifest on Openshift 1.11 now.

PR #3566, opened 2020-07-30
Add support for keystore with Kibana 7.9.0

### LINKED ISSUE #3528: TestUpdateKibanaSecureSettings is flaky
https://devops-ci.elastic.co/job/cloud-on-k8s-e2e-tests-snapshot-versions/106/testReport/

```
=== RUN   TestUpdateKibanaSecureSettings/secure_settings_should_eventually_be_set_in_all_nodes_keystore
Retries (5m0s timeout): ................................................................................
    utils.go:84: 
        	Error Trace:	utils.go:84
        	            	invalid keystore entries. Expected: [logging.verbose]. Actual: []
        	Test:       	TestUpdateKibanaSecureSettings/secure_settings_should_eventually_be_set_in_all_nodes_keystore
```

barkbay:
Starting Kibana 7.9.0 the keystore is located in the configuration directory instead of the data directory.

This PR allows the keystore init container and the Kibana container to share the `config` directory through an empty directory.

For consistency with what is done with Elasticsearch this PR creates symbolic links from the generated configuration to the shared configuration volume.

Fixes #3528

### DIFF
--- a/pkg/controller/common/volume/shared_volumes.go
+++ b/pkg/controller/common/volume/shared_volumes.go
@@ -6,7 +6,7 @@ package volume
 
 import corev1 "k8s.io/api/core/v1"
 
-// SharedVolume between the init container and the ES container.
+// SharedVolume between the init container and the main container.
 type SharedVolume struct {
 	VolumeName             string // Volume name
 	InitContainerMountPath string // Mount path in the init container
--- a/pkg/controller/common/volume/shared_volumes.go
+++ b/pkg/controller/common/volume/shared_volumes.go
@@ -2,34 +2,38 @@
 // or more contributor license agreements. Licensed under the Elastic License;
 // you may not use this file except in compliance with the Elastic License.
 
-package initcontainer
+package volume
 
 import corev1 "k8s.io/api/core/v1"
 
 // SharedVolume between the init container and the ES container.
 type SharedVolume struct {
-	Name                   string // Volume name
+	VolumeName             string // Volume name
 	InitContainerMountPath string // Mount path in the init container
-	EsContainerMountPath   string // Mount path in the Elasticsearch container
+	ContainerMountPath     string // Mount path in the main container (e.g. Elasticsearch or Kibana)
 }
 
 func (v SharedVolume) InitContainerVolumeMount() corev1.VolumeMount {
 	return corev1.VolumeMount{
 		MountPath: v.InitContainerMountPath,
-		Name:      v.Name,
+		Name:      v.VolumeName,
 	}
 }
 
-func (v SharedVolume) EsContainerVolumeMount() corev1.VolumeMount {
+func (v SharedVolume) Name() string {
+	return v.VolumeName
+}
+
+func (v SharedVolume) VolumeMount() corev1.VolumeMount {
 	return corev1.VolumeMount{
-		MountPath: v.EsContainerMountPath,
-		Name:      v.Name,
+		MountPath: v.ContainerMountPath,
+		Name:      v.VolumeName,
 	}
 }
 
 func (v SharedVolume) Volume() corev1.Volume {
 	return corev1.Volume{
-		Name: v.Name,
+		Name: v.VolumeName,
 		VolumeSource: corev1.VolumeSource{
 			EmptyDir: &corev1.EmptyDirVolumeSource{},
 		},
@@ -49,10 +53,10 @@ func (v SharedVolumeArray) InitContainerVolumeMounts() []corev1.VolumeMount {
 	return mounts
 }
 
-func (v SharedVolumeArray) EsContainerVolumeMounts() []corev1.VolumeMount {
+func (v SharedVolumeArray) ContainerVolumeMounts() []corev1.VolumeMount {
 	mounts := make([]corev1.VolumeMount, len(v.Array))
 	for i, v := range v.Array {
-		mounts[i] = v.EsContainerVolumeMount()
+		mounts[i] = v.VolumeMount()
 	}
 	return mounts
 }
@@ -61,7 +65,7 @@ func (v SharedVolumeArray) Volumes() []corev1.Volume {
 	volumes := make([]corev1.Volume, len(v.Array))
 	for i, v := range v.Array {
 		volumes[i] = corev1.Volume{
-			Name: v.Name,
+			Name: v.VolumeName,
 			VolumeSource: corev1.VolumeSource{
 				EmptyDir: &corev1.EmptyDirVolumeSource{},
 			},
--- a/pkg/controller/kibana/config_reconcile.go
+++ b/pkg/controller/kibana/config_reconcile.go
@@ -27,6 +27,7 @@ const (
 	ConfigVolumeMountPath              = "/usr/share/kibana/config"
 	InitContainerConfigVolumeMountPath = "/mnt/elastic-internal/kibana-config-local"
 
+	// InternalConfigVolumeName is a volume which contains the generated configuration.
 	InternalConfigVolumeName      = "elastic-internal-kibana-config"
 	InternalConfigVolumeMountPath = "/mnt/elastic-internal/kibana-config"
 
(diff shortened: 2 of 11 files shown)


barkbay:
jenkins test this please

pkg/controller/common/volume/shared_volumes.go
```diff
@@ -2,34 +2,38 @@
 // or more contributor license agreements. Licensed under the Elastic License;
 // you may not use this file except in compliance with the Elastic License.
 
-package initcontainer
+package volume
 
 import corev1 "k8s.io/api/core/v1"
 
 // SharedVolume between the init container and the ES container.
```
sebgl, line 9: `between the init container and the main container`

pkg/controller/kibana/init_configuration.go
```diff
@@ -0,0 +1,65 @@
...
+// you may not use this file except in compliance with the Elastic License.
+
+package kibana
+
+import (
+	kbv1 "github.com/elastic/cloud-on-k8s/pkg/apis/kibana/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+)
+
```
sebgl, line 12: I think a few lines of comment to explain the general approach in cleartext would be nice

pkg/controller/kibana/config_reconcile.go
```diff
@@ -23,28 +23,43 @@ import (
 
 // Constants to use for the config files in a Kibana pod.
 const (
-	VolumeName        = "config"
-	VolumeMountPath   = "/usr/share/kibana/" + VolumeName
+	ConfigVolumeName                   = "elastic-internal-kibana-config-local"
+	ConfigVolumeMountPath              = "/usr/share/kibana/config"
+	InitContainerConfigVolumeMountPath = "/mnt/elastic-internal/kibana-config-local"
+
+	InternalConfigVolumeName      = "elastic-internal-elasticsearch-config"
```
sebgl, line 30: shouldn't it be `elastic-internal-kibana-config`?

pkg/controller/kibana/config_reconcile.go
```diff
@@ -23,28 +23,43 @@ import (
 
 // Constants to use for the config files in a Kibana pod.
 const (
-	VolumeName        = "config"
-	VolumeMountPath   = "/usr/share/kibana/" + VolumeName
+	ConfigVolumeName                   = "elastic-internal-kibana-config-local"
```
sebgl, line 26: a few lines of comments explaining the general logic and why we have several volumes would be nice

pkg/controller/kibana/config_reconcile.go
```diff
@@ -23,28 +23,43 @@ import (
...
+	ConfigVolumeMountPath              = "/usr/share/kibana/config"
+	InitContainerConfigVolumeMountPath = "/mnt/elastic-internal/kibana-config-local"
+
+	InternalConfigVolumeName      = "elastic-internal-elasticsearch-config"
+	InternalConfigVolumeMountPath = "/mnt/elastic-internal/kibana-config"
```
sebgl, line 31: `internal` is a bit confusing: why is `/mnt/elastic-internal/kibana-config` "internal" compared to `/mnt/elastic-internal/kibana-config-local`? Both living in `elastic-internal` sound like "internal" stuff.

pkg/controller/kibana/config_reconcile.go
sebgl, line 26: Would `shared` be more explicit than `local` to express how the volume is shared between init and main container?

pkg/controller/kibana/driver.go
```diff
@@ -287,6 +251,21 @@ func (d *driver) deploymentParams(kb *kbv1.Kibana) (deployment.Params, error) {
 	}, nil
 }
 
+func (d *driver) buildVolumes(kb *kbv1.Kibana) []commonvolume.VolumeLike {
+	volumes := []commonvolume.VolumeLike{DataVolume, ConfigSharedVolume, ConfigVolume(*kb)}
```
sebgl, line 255: Are we mounting `ConfigVolume()` (if I'm not mistaken: the "raw" config from the secret resource) into the main Kibana container? Isn't it only required for the init container?

pkg/controller/kibana/keystore.go
```diff
@@ -0,0 +1,47 @@
...
+
+import (
+	kbv1 "github.com/elastic/cloud-on-k8s/pkg/apis/kibana/v1"
+	"github.com/elastic/cloud-on-k8s/pkg/controller/common/keystore"
+	"github.com/elastic/cloud-on-k8s/pkg/controller/common/version"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+)
+
+// keystoreInConfigDirVersion is the version in which the keystore is no longer stored in the data directory but in the configuration one.
```
sebgl, line 15: maybe "in the config one" to reuse the exact directory name

pkg/controller/kibana/init_configuration.go
```diff
@@ -0,0 +1,65 @@
...
+const (
+	InitConfigContainerName = "elastic-internal-init-config"
+
+	// InitConfigScript is a small bash script to prepare the Kibana configuration directory
+	InitConfigScript = `#!/usr/bin/env bash
+set -eux
+
+init_config_initialized_flag=/usr/share/kibana/config/elastic-internal-init-config.ok
```
sebgl, line 20: is it only to avoid the cost of re-doing `ln -sf`?

pkg/controller/kibana/init_configuration.go
```diff
@@ -0,0 +1,65 @@
...
+
+if [[ -f "${init_config_initialized_flag}" ]]; then
+    echo "Kibana configuration already initialized."
+	exit 0
+fi
+
+echo "Setup Kibana configuration"
+
+ln -sf /mnt/elastic-internal/kibana-config/* /mnt/elastic-internal/kibana-config-local/
```
sebgl, line 29: this should probably reuse 2 constants instead of being hardcoded?

pkg/controller/common/volume/shared_volumes.go
```diff
@@ -2,34 +2,38 @@
 // or more contributor license agreements. Licensed under the Elastic License;
 // you may not use this file except in compliance with the Elastic License.
 
-package initcontainer
```
sebgl, line 5: Nice to see the refactoring here 👍

pkg/controller/kibana/config_reconcile.go
barkbay: Naming has been chosen to be consistent with what is done with Elasticsearch:

https://github.com/elastic/cloud-on-k8s/blob/b6cb04d65574f18843dc3fcd93d50e5232009177/pkg/controller/elasticsearch/initcontainer/prepare_fs.go#L36-L40

I agree that `shared` would be more explicit but I'm not sure we want to do this refactoring for a patch release ?

pkg/controller/kibana/driver.go
barkbay: The Secret with the configuration is needed because I eventually decided to create some symbolic links. Even if it is not strictly required for Kibana I think that it is easier to reason about something that is close to what is done with Elasticsearch.

pkg/controller/kibana/init_configuration.go
barkbay: Initially I though it was less readable, but let's avoid duplication.

pkg/controller/kibana/init_configuration.go
barkbay: This is to avoid the content of the script to be run twice, I agree it should be safe to remove it here. But I also want to avoid `Setup Kibana configuration` to be printed several times in the logs. It's harmless indeed, but I think it can be confusing for the user/support to see those messages while the container is already running for a while.

pkg/controller/kibana/config_reconcile.go
barkbay: This is also to be consistent with what is done with Elasticsearch: 

```
    - mountPath: /mnt/elastic-internal/elasticsearch-config
      name: elastic-internal-elasticsearch-config
      readOnly: true
    - mountPath: /usr/share/elasticsearch/config
      name: elastic-internal-elasticsearch-config-local
```

PR #3568, opened 2020-07-31
Move to a UBI base image

### LINKED ISSUE #3551: Support a UBI base image
We want to create a version of ECK that is based on `registry.access.redhat.com/ubi8/ubi-minimal` while keeping the current `distroless` image for users who do not want a ubi baseimage.

* Separate build pipeline for UBI
* Adjust test pipelines for OpenShift to use the UBI base image
* Keep current `distroless` testpipeline for OpenShift as well but optionally reduce tests run to a smoke test.

pebrc:
Fixes #3551 

Move ECK to a UBI base image. I opted for simply replacing `distroless` with `ubi-minimal`. The benefit of a unified build and CI environment seems much more compelling after looking into what it would take to have two parallel distributions of ECK with much deeper interdependent stacking of build pipelines and even more complex Makefiles than we already have. 

The original move to distroless was driven mainly by our desire to address security issues in the centos image we were using.  But with ubi we will have to address any potential future vulnerabilities anyway by upgrading to a newer image which will lead to the same end result. 

Sample image: `docker.elastic.co/eck-snapshots/eck-operator-ci:1.3.0-SNAPSHOT-5763bdb5`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -37,9 +37,13 @@ LABEL name="Elastic Cloud on Kubernetes" \
       description="Elastic Cloud on Kubernetes automates the deployment, provisioning, management, and orchestration of Elasticsearch, Kibana, APM Server, Beats, and Enterprise Search on Kubernetes" \
       io.k8s.description="Elastic Cloud on Kubernetes automates the deployment, provisioning, management, and orchestration of Elasticsearch, Kibana, APM Server, Beats, and Enterprise Search on Kubernetes"
 
+# Update the base image packages to the latest versions
 RUN microdnf update --setopt=tsflags=nodocs && microdnf clean all
 
 COPY --from=builder /go/src/github.com/elastic/cloud-on-k8s/elastic-operator .
+
+# Copy NOTICE.txt and LICENSE.txt into the image
+COPY *.txt /licenses/
 
 USER 1001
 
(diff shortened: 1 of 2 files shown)


anyasabo:
I think we want to go back to also updating packages, correct? I'm not sure how often they update the UBI image, but at least centos wasn't updated frequently. See how we did it before the distroless switch here: https://github.com/elastic/cloud-on-k8s/pull/3161/files

Dockerfile
```diff
@@ -22,7 +22,13 @@ RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 \
 			-o elastic-operator github.com/elastic/cloud-on-k8s/cmd
 
 # Copy the operator binary into a lighter image
-FROM gcr.io/distroless/static:nonroot
+FROM registry.access.redhat.com/ubi8/ubi-minimal:8.2
+
+RUN microdnf update --setopt=tsflags=nodocs && microdnf clean all
```
thbkrkr, line 27:
```suggestion
RUN microdnf update --setopt=tsflags=nodocs && microdnf clean all
```

anyasabo:
Out of curiousity are the labels required for OpenShift certification? Or what makes them required?

pebrc:
> Out of curiousity are the labels required for OpenShift certification? Or what makes them required?

Yes https://redhat-connect.gitbook.io/partner-guide-for-red-hat-openshift-and-container/program-on-boarding/technical-prerequisites

manojah99:
Just have 2 questions @pebrc  if you can answer them :
1. Why did eck operator moved from destroless to UBI, any particular reason apart one given in PR?
2. What happens to destroless version in 1.3?

Issue #475, opened 2019-03-06
Replace fsnotify with a simpler polling mechanism


sebgl:
We've observed some inconsistencies in our usage of fsnotify:

* sometimes getting a CREATE event for a file that does not seem to be there when reading it, resulting on the file never being reconciled
* sometimes getting no event when something did happen on the filesystem (see https://github.com/elastic/k8s-operators/issues/451)

If we can't exactly understand why fsnotify seems to sometimes fail for us, we could get rid of it and replace it by a simpler "polling" mechanism:

1. poll the fs every 1 sec to read the files we're interested in
2. do whatever we need to check on their content

We can optimize step 2. by keeping the file content in memory, and only execute 2. if the content has changed.

PR #3570, opened 2020-07-31
Allow the operator to be configured from a file

### LINKED ISSUE #3401: Rely on a configMap for operator-level settings
Forked from https://github.com/elastic/cloud-on-k8s/issues/2791.

The number of flags that can be passed to the operator binary is becoming quite large. It may feel more natural for users to tweak them through a configMap rather than modifying the operator StatefulSet spec. Providing a default config with comments explaining each setting would also make the possible options easier to understand.

Should we setup a mechanism to auto-reload the config, or auto-kill/restart the operator when the configuration changes, or should we let users deal with it?

charith-elastic:
Adds the ability to configure the operator from a file. This is a backward-compatible change that retains the existing configuration methods using flags and environment variables. Users can provide the configuration file using the `--config` flag. If there are any conflicts, the order of precedence for resolving them is:

- Flag
- Environment variable
- File

This PR also includes:
- Update to the `all-in-one.yaml` to use a ConfigMap for operator configuration
- Update to the operator configuration documentation
- A minor refactoring of the command code 

Fixes #3401

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -70,8 +70,7 @@ const (
 	DefaultWebhookConfigurationName = "elastic-webhook.k8s.elastic.co"
 	WebhookPort                     = 9443
 
-	restartDelay             = 10 * time.Second // time to wait before restarting the operator when configuration changes
-	debugHTTPShutdownTimeout = 5 * time.Second  // time to allow for the debug HTTP server to shutdown
+	debugHTTPShutdownTimeout = 5 * time.Second // time to allow for the debug HTTP server to shutdown
 )
 
 var (
@@ -150,6 +149,11 @@ func Command() *cobra.Command {
 		"Listen address for debug HTTP server (only available in development mode)",
 	)
 	cmd.Flags().Bool(
+		operator.DisableConfigWatch,
+		false,
+		"Disable watching the config file for changes",
+	)
+	cmd.Flags().Bool(
 		operator.EnforceRBACOnRefsFlag,
 		false, // Set to false for backward compatibility
 		"Restrict cross-namespace resource association through RBAC (eg. referencing Elasticsearch from Kibana)",
@@ -223,13 +227,17 @@ func doRun(_ *cobra.Command, _ []string) error {
 	// receive config file update events over a channel
 	confUpdateChan := make(chan struct{}, 1)
 
-	viper.WatchConfig()
-	viper.OnConfigChange(func(evt fsnotify.Event) {
-		switch evt.Op {
-		case fsnotify.Create, fsnotify.Write:
-			confUpdateChan <- struct{}{}
-		}
-	})
+	// start watching the file for changes unless the DisableConfigWatch flag is set
+	if !viper.GetBool(operator.DisableConfigWatch) {
+		log.Info("Watching config file for changes", "file", configFile)
+		viper.WatchConfig()
+		viper.OnConfigChange(func(evt fsnotify.Event) {
+			switch evt.Op {
+			case fsnotify.Create, fsnotify.Write:
+				confUpdateChan <- struct{}{}
+			}
+		})
+	}
 
 	// start the operator in a goroutine
 	errChan := make(chan error, 1)
@@ -242,26 +250,23 @@ func doRun(_ *cobra.Command, _ []string) error {
 		case err := <-errChan: // operator failed
 			log.Error(err, "Shutting down due to error")
 			close(stopChan)
+
 			return err
 		case <-signalChan: // signal received
 			log.Info("Signal received: shutting down")
 			close(stopChan)
-			err := <-errChan
 
-			return err
+			return <-errChan
 		case <-confUpdateChan: // config file updated
-			log.Info("Restarting to apply updated configuration")
+			log.Info("Shutting down to apply updated configuration")
 			close(stopChan)
 
 			if err := <-errChan; err != nil {
 				log.Error(err, "Encountered error from previous operator run")
+				return err
 			}
 
-			// sleep to allow time for resource cleanup from the previous run
-			time.Sleep(restartDelay)
-
-			stopChan = make(chan struct{})
-			launchOperator(stopChan, errChan)
+			return nil
 		}
 	}
 }
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -232,7 +232,9 @@ func doRun(_ *cobra.Command, _ []string) error {
 		log.Info("Watching config file for changes", "file", configFile)
 		viper.WatchConfig()
 		viper.OnConfigChange(func(evt fsnotify.Event) {
-			switch evt.Op {
+			if evt.Op&fsnotify.Write == fsnotify.Write || evt.Op&fsnotify.Create == fsnotify.Create {
+				confUpdateChan <- struct{}{}
+			}
 			case fsnotify.Create, fsnotify.Write:
 				confUpdateChan <- struct{}{}
 			}
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -83,13 +83,6 @@ func Command() *cobra.Command {
 		Use:   "manager",
 		Short: "Start the ECK operator",
 		PreRunE: func(cmd *cobra.Command, _ []string) error {
-			if configFile != "" {
-				viper.SetConfigFile(configFile)
-				if err := viper.ReadInConfig(); err != nil {
-					return fmt.Errorf("failed to read config file %s: %w", configFile, err)
-				}
-			}
-
 			// enable using dashed notation in flags and underscores in env
 			viper.SetEnvKeyReplacer(strings.NewReplacer("-", "_"))
 
@@ -98,6 +91,18 @@ func Command() *cobra.Command {
 			}
 
 			viper.AutomaticEnv()
+
+			if configFile != "" {
+				viper.SetConfigFile(configFile)
+				if err := viper.ReadInConfig(); err != nil {
+					return fmt.Errorf("failed to read config file %s: %w", configFile, err)
+				}
+
+				if !viper.GetBool(operator.DisableConfigWatch) {
+					viper.WatchConfig()
+				}
+			}
+
 			logconf.ChangeVerbosity(viper.GetInt(logconf.FlagName))
 			log = logf.Log.WithName("manager")
 
@@ -218,25 +223,21 @@ func Command() *cobra.Command {
 
 func doRun(_ *cobra.Command, _ []string) error {
 	signalChan := signals.SetupSignalHandler()
+	disableConfigWatch := viper.GetBool(operator.DisableConfigWatch)
 
-	// no config file to watch so start the operator
-	if configFile == "" {
+	// no config file to watch so start the operator directly
+	if configFile == "" || disableConfigWatch {
 		return startOperator(signalChan)
 	}
 
 	// receive config file update events over a channel
 	confUpdateChan := make(chan struct{}, 1)
 
-	// start watching the file for changes unless the DisableConfigWatch flag is set
-	if !viper.GetBool(operator.DisableConfigWatch) {
-		log.Info("Watching config file for changes", "file", configFile)
-		viper.WatchConfig()
-		viper.OnConfigChange(func(evt fsnotify.Event) {
-			if evt.Op&fsnotify.Write == fsnotify.Write || evt.Op&fsnotify.Create == fsnotify.Create {
-				confUpdateChan <- struct{}{}
-			}
-		})
-	}
+	viper.OnConfigChange(func(evt fsnotify.Event) {
+		if evt.Op&fsnotify.Write == fsnotify.Write || evt.Op&fsnotify.Create == fsnotify.Create {
+			confUpdateChan <- struct{}{}
+		}
+	})
 
 	// start the operator in a goroutine
 	errChan := make(chan error, 1)
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -6,6 +6,7 @@ package manager
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net/http"
 	"net/http/pprof"
@@ -42,8 +43,11 @@ import (
 	"github.com/elastic/cloud-on-k8s/pkg/dev"
 	"github.com/elastic/cloud-on-k8s/pkg/dev/portforward"
 	licensing "github.com/elastic/cloud-on-k8s/pkg/license"
+	logconf "github.com/elastic/cloud-on-k8s/pkg/utils/log"
 	"github.com/elastic/cloud-on-k8s/pkg/utils/net"
 	"github.com/elastic/cloud-on-k8s/pkg/utils/rbac"
+	"github.com/fsnotify/fsnotify"
+	"github.com/go-logr/logr"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 	"go.elastic.co/apm"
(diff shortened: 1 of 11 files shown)


cmd/manager/main.go
```diff
@@ -68,125 +72,153 @@ const (
...
+			if configFile != "" {
+				viper.SetConfigFile(configFile)
+				if err := viper.ReadInConfig(); err != nil {
+					return fmt.Errorf("failed to read config file %s: %w", configFile, err)
+				}
+
+				viper.WatchConfig()
+				viper.OnConfigChange(func(_ fsnotify.Event) {
+					if log != nil {
+						log.Info("Detected an update to the configuration file. Restart the operator for the changes to take effect.")
```
pebrc, line 93: This is likely to be missed by users though? Is it worth having it? Or can we make it more prominent with an event maybe?

docs/operating-eck/operator-config.asciidoc
```diff
@@ -51,4 +52,41 @@ Duration values should be specified as numeric values suffixed by the time unit.
...
+.Environment variables method
+[source,sh]
+----
+LOG_VERBOSITY=2 METRICS_PORT=6060 NAMESPACES="ns1,ns2,ns3" ./elastic-operator manager
+----
+
+If you use a combination of all or some of the methods listed above, the descending order of precedence in case of a conflict is as follows:
+
+- Flag
+- Environment variable
```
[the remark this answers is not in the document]
charith-elastic: Unfortunately, Viper does not allow configuring the order of precedence. OLM is an interesting problem. As far as I can tell, they don't yet support config maps to be bundled with the CSV. However, [they do seem to allow the user to mount volumes using the Subscription object](https://docs.openshift.com/container-platform/4.4/rest_api/operatorhub_apis/subscription-operators-coreos-com-v1alpha1.html). I haven't tested whether that actually works but assuming that it does, then I propose the following:
- We bundle a default config file in the operator container image at a well known path (say `/conf/eck.yaml`).
- In the CSV deployment section, we only set the `--config` flag which will point to the default config file above.

With this approach, in addition to environment variables, the user has the option of using the `Subscription` to mount a config map at `/conf` that will shadow the default file bundled with the operator. That ought to take care of the configuration problem for OLM. WDYT?

cmd/manager/main.go
charith-elastic: I mainly added it as an aid to jog the users memory. Hopefully, most users will look at the logs when they notice that the config change hasn't taken effect. Even if they don't, someone providing support will :crossed_fingers: 

I'll look into adding an event too but I suspect that it'll be tricky to do at such an early stage of the operator life cycle because the controller runtime hasn't been initialized yet.

docs/operating-eck/operator-config.asciidoc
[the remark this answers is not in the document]
pebrc: I think that sounds interesting. We could do that in a separate PR if you want.

cmd/manager/main.go
barkbay, line 93: I think as a user I would expect ECK to automatically restart. This is the current behaviour for all the stack components so far (and more generally what happen in K8S with the reconcile loop pattern).

An other argument is that I'm not sure how restarting manually fits with a "gitops" approach.

docs/operating-eck/operator-config.asciidoc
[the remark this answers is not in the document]
charith-elastic: I created #3591 to track it.

cmd/manager/main.go
charith-elastic: With GitOps, you'd create a dependency between the config map and the pod (trivially, this could be an annotation in the pod that gets updated with the hash of the config file) and that would cause the pod to be restarted automatically.

Unfortunately, the way the operator is structured, I don't see a way to "hot reload" the configuration. So then the question is how to force a restart. One way is to call `os.Exit` but that seems too crude and assumes that there's a restart mechanism available. The other one I can think of is having a liveness probe that starts failing when the config changes.  

Do you have any thoughts on achieving this nicely without complicating the operator too much?

barkbay: > Do you have any thoughts on achieving this nicely without complicating the operator too much?

The manager can be stopped through a [channel](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/manager/internal.go#L438), I would try to explore this solution.

charith-elastic: I had a [go at it](https://github.com/elastic/cloud-on-k8s/pull/3570/commits/e3bed8b1acfb26ddd0ae80e14c4274fb723a6b11) and it kind of works. The issues with it are:
- Log level cannot be changed by that method (See #3593)
- Unmanaged goroutines in the code could start leaking because they don't have a proper shutdown mechanism

barkbay: I think we can just let the operator stop completely and delegate the restart of the container to the statefulset controller.
I think this is what would happen if the user stopped it itself.

charith-elastic:
As discussed during the team meeting, currently there is no effective way to cleanly stop the goroutines that are spawned in the code. Therefore, instead of trying to restart eveything in-process, the operator will simply exit and rely on an external mechanism (Statefulset/Pod controller) to restart it.

cmd/manager/main.go
```diff
@@ -65,128 +69,218 @@ const (
...
+
+	// receive config file update events over a channel
+	confUpdateChan := make(chan struct{}, 1)
+
+	// start watching the file for changes unless the DisableConfigWatch flag is set
+	if !viper.GetBool(operator.DisableConfigWatch) {
+		log.Info("Watching config file for changes", "file", configFile)
+		viper.WatchConfig()
+		viper.OnConfigChange(func(evt fsnotify.Event) {
+			switch evt.Op {
```
barkbay, line 235: An event may contain several operations, e.g. `fsnotify.Write | fsnotify.Chmod` (bitwise OR)
```suggestion
			if evt.Op&fsnotify.Write == fsnotify.Write || evt.Op&fsnotify.Create == fsnotify.Create {
				confUpdateChan <- struct{}{}
			}
```

cmd/manager/main.go
```diff
@@ -65,128 +69,218 @@ const (
...
+		return startOperator(signalChan)
+	}
+
+	// receive config file update events over a channel
+	confUpdateChan := make(chan struct{}, 1)
+
+	// start watching the file for changes unless the DisableConfigWatch flag is set
+	if !viper.GetBool(operator.DisableConfigWatch) {
+		log.Info("Watching config file for changes", "file", configFile)
+		viper.WatchConfig()
```
barkbay, line 233: I think it might be too late to watch the configuration file now, changes may have happened between `viper.ReadInConfig();` and this line. Some changes might not be detected.

barkbay:
Regarding `fsnotify` we used it in the past [but replaced it](https://github.com/elastic/cloud-on-k8s/issues/475) with a simpler mechanism. I'm wondering if the same concerns may apply here.

charith-elastic:
[The watch mechanism in Viper seems to be fairly well implemented (including accounting for config map updates)](https://github.com/spf13/viper/blob/3826be313591f83193f048520482a7b3cf17d506/viper.go#L340) so I think I would go with `fsnotify` for the time being. It's not the end of the world if a certain config update is not detected in time -- therefore it's a fairly low risk to take. If it turns out that there are fundamental problems with it, we can revisit and consider an alternative approach.

charith-elastic:
Jenkins test this please

cmd/manager/main.go
```diff
@@ -411,20 +473,68 @@ func execute() {
 		"namespace", operatorNamespace, "version", operatorInfo.BuildInfo.Version,
 		"build_hash", operatorInfo.BuildInfo.Hash, "build_date", operatorInfo.BuildInfo.Date,
 		"build_snapshot", operatorInfo.BuildInfo.Snapshot)
-	if err := mgr.Start(signals.SetupSignalHandler()); err != nil {
-		log.Error(err, "unable to run the manager")
-		os.Exit(1)
+
+	if err := mgr.Start(stopChan); err != nil {
+		log.Error(err, "Unable to run the manager")
```
thbkrkr, line 478:
```suggestion
		log.Error(err, "Failed to start the controller manager")
```

cmd/manager/main.go
```diff
@@ -282,28 +387,40 @@ func execute() {
...
+	}
+
 	log.V(1).Info("Using certificate rotation parameters", operator.CertValidityFlag, certValidity, operator.CertRotateBeforeFlag, certRotateBefore)
 
 	// Setup a client to set the operator uuid config map
 	clientset, err := kubernetes.NewForConfig(cfg)
 	if err != nil {
-		log.Error(err, "unable to create k8s clientset")
-		os.Exit(1)
+		log.Error(err, "Unable to create Kubernetes client")
```
thbkrkr, line 414:
```suggestion
		log.Error(err, "Failed to create Kubernetes client")
```

cmd/manager/main.go
```diff
@@ -282,28 +387,40 @@ func execute() {
 	opts.Port = WebhookPort
 	mgr, err := ctrl.NewManager(cfg, opts)
 	if err != nil {
-		log.Error(err, "unable to create controller manager")
-		os.Exit(1)
+		log.Error(err, "Unable to create controller manager")
```
thbkrkr, line 390:
```suggestion
		log.Error(err, "Failed to create controller manager")
```

cmd/manager/main.go
```diff
@@ -65,128 +69,215 @@ const (
...
+	cmd.Flags().String(
 		operator.DebugHTTPListenFlag,
 		"localhost:6060",
 		"Listen address for debug HTTP server (only available in development mode)",
 	)
-	Cmd.Flags().Bool(
+	cmd.Flags().Bool(
+		operator.DisableConfigWatch,
+		false,
+		"Disable watching the config file for changes",
```
thbkrkr, line 159:
```suggestion
		"Disable watching the configuration file for changes",
```

cmd/manager/main.go
```diff
@@ -65,128 +69,215 @@ const (
...
+	cmd.Flags().Int(
 		operator.MetricsPortFlag,
 		DefaultMetricPort,
 		"Port to use for exposing metrics in the Prometheus format (set 0 to disable)",
 	)
-	Cmd.Flags().StringSlice(
+	cmd.Flags().StringSlice(
 		operator.NamespacesFlag,
 		nil,
 		"comma-separated list of namespaces in which this operator should manage resources (defaults to all namespaces)",
```
thbkrkr, line 193:
```suggestion
		"Comma-separated list of namespaces in which this operator should manage resources (defaults to all namespaces)",
```

cmd/manager/main.go
```diff
@@ -65,128 +69,215 @@ const (
...
 	)
-	Cmd.Flags().String(
+	cmd.Flags().String(
 		operator.OperatorNamespaceFlag,
 		"",
 		"K8s namespace the operator runs in",
```
thbkrkr, line 198:
```suggestion
		"Kubernetes namespace the operator runs in",
```

cmd/manager/main.go
```diff
@@ -65,128 +69,215 @@ const (
...
 		operator.WebhookCertDirFlag,
 		// this is controller-runtime's own default, copied here for making the default explicit when using `--help`
 		filepath.Join(os.TempDir(), "k8s-webhook-server", "serving-certs"),
 		"Path to the directory that contains the webhook server key and certificate",
 	)
-	Cmd.Flags().String(
+	cmd.Flags().String(
 		operator.WebhookSecretFlag,
 		"",
 		fmt.Sprintf("K8s secret mounted into the path designated by %s to be used for webhook certificates", operator.WebhookCertDirFlag),
```
thbkrkr, line 209:
```suggestion
		fmt.Sprintf("Kubernetes secret mounted into the path designated by %s to be used for webhook certificates", operator.WebhookCertDirFlag),
```

cmd/manager/main.go
```diff
@@ -65,128 +69,215 @@ const (
...
 		operator.CertValidityFlag,
 		certificates.DefaultCertValidity,
 		"Duration representing how long before a newly created TLS certificate expires",
 	)
-	Cmd.Flags().String(
+	cmd.Flags().StringVar(
+		&configFile,
+		operator.ConfigFlag,
+		"",
+		"Path to the ECK configuration file",
```
thbkrkr, line 144:
```suggestion
		"Path to the configuration file",
```

or 

```suggestion
		"Path to the file containing the operator configuration",
```

cmd/manager/main.go
```diff
@@ -65,128 +69,215 @@ const (
...
 
-	log = logf.Log.WithName("manager")
-)
-
-func init() {
-	Cmd.Flags().Bool(
+	cmd.Flags().Bool(
 		operator.AutoPortForwardFlag,
 		false,
 		"enables automatic port-forwarding "+
```
thbkrkr, line 117:
```suggestion
		"Enables automatic port-forwarding "+
```

cmd/manager/main.go
```diff
@@ -65,128 +69,215 @@ const (
...
+
+func Command() *cobra.Command {
+	cmd := &cobra.Command{
 		Use:   "manager",
-		Short: "Start the operator manager",
-		Long: `manager starts the manager for this operator,
- which will in turn create the necessary controller.`,
-		Run: func(cmd *cobra.Command, args []string) {
-			execute()
+		Short: "Start the ECK operator",
```
thbkrkr, line 84:
```suggestion
		Short: "Start the Elastic Cloud on Kubernetes operator",
```

hack/manifest-gen/assets/charts/eck/templates/configmap.yaml
```diff
@@ -0,0 +1,35 @@
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ .Values.operator.name }}
```
barkbay, line 5:
```
> k get cm -n elastic-system 
NAME                    DATA   AGE
elastic-operator        1      5m
elastic-operator-uuid   1      19s
```

Should we call the ConfigMap `elastic-operator-configuration` to make its purpose more explicit ?

docs/operating-eck/operator-config.asciidoc
```diff
@@ -51,4 +52,41 @@ Duration values should be specified as numeric values suffixed by the time unit.
...
+- File
+
+
+You can edit the `elastic-operator` ConfigMap to change the operator configuration. A manual restart of the operator is required for the changes to take effect. Alternatively, you can edit the `elastic-operator` StatefulSet  and add flags to the `args` section -- which will trigger an automatic restart of the operator pod by the StatefulSet controller.
```
barkbay, line 92: > A manual restart of the operator is required for the changes to take effect

This should not be the case anymore ?

docs/operating-eck/operator-config.asciidoc
```diff
@@ -18,6 +18,7 @@ ECK can be configured using either command line flags or environment variables.
 |ca-cert-validity |8760h |Duration representing the validity period of a generated CA certificate.
 |cert-rotate-before |24h |Duration representing how long before expiration TLS certificates should be re-issued.
 |cert-validity |8760h |Duration representing the validity period of a generated TLS certificate.
+|config |"" | Path to a file containing the operator configuration.
```
barkbay, line 21: Should we also document `disable-config-watch` ?

pkg/controller/common/operator/flags.go
```diff
@@ -10,8 +10,10 @@ const (
 	CACertValidityFlag           = "ca-cert-validity"
 	CertRotateBeforeFlag         = "cert-rotate-before"
 	CertValidityFlag             = "cert-validity"
+	ConfigFlag                   = "config"
 	ContainerRegistryFlag        = "container-registry"
 	DebugHTTPListenFlag          = "debug-http-listen"
+	DisableConfigWatch           = "disable-config-watch"
```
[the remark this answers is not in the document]
charith-elastic: Thanks for pointing out the inconsistency. I don't think expanding `config` to `configuration` achieves much (there's no ambiguity) other than making the flags comically long. Since the webhook flag hasn't been released yet, I think I will rename that instead. Even Kubernetes uses `ConfigMap` instead of `ConfigurationMap` :smile:.

hack/manifest-gen/assets/charts/eck/templates/configmap.yaml
charith-elastic: I have seen quite a fair amount of applications simply using the name of the application and I think that convention makes sense. It is a config map after all so the purpose is obvious without the additional `-configuration` suffix. If there are strong opinions against that, I'd be happy to reconsider.

barkbay: > It is a config map after all so the purpose is obvious

I'm not sure I would agree with that, we already have 2 ConfigMaps which are not really related to the configuration and we might have a third one for the election mechanism:

```
k get cm -n elastic-system
NAME                      DATA   AGE
elastic-licensing         4      18h
elastic-operator-leader   0      18h
elastic-operator-uuid     1      18h
```

I would be curious what other people's thoughts are, but I don't want to block this PR.

charith-elastic: I see your point. I am probably building up the convention in my mind because I have been conditioned by it but that probably doesn't apply to everybody. I'll merge this for now and if it turns out to be confusing, we still have time to change it before the next release goes out. 

Simply because I have it in hand, Istio config maps look like the following. 

```
istio                                  2      63d
istio-ca-root-cert                     1      63d
istio-leader                           0      63d
istio-namespace-controller-election    0      63d
istio-security                         1      63d
istio-sidecar-injector                 2      63d
istio-validation-controller-election   0      63d
kiali                                  1      58d
prometheus                             1      63d
```

I am not suggesting that this is a canonical example. However, I think it's quite obvious what each config map does. I reckon that ours would look similar.

PR #3574, opened 2020-08-03
Update owner references in reconcile resource util function

### LINKED ISSUE #3470: User is not propagated to new ES if reference is changed
Seems like it's affecting all references (I tested Beats and Kibana).

Repro:
1. Create `elasticsearch1` cluster and `kibana1` instance that references `elasticsearch1`.
2. Create `elasticsearch2` cluster.
3. Edit `kibana1` to reference `elasticsearch2`.

Kibana won't come up because it can't auth with `elasticsearch2`, because the user doesn't exist:
```sh
$ k exec elasticsearch2-es-default-0 cat /mnt/elastic-internal/xpack-file-realm/users
elastic-internal-probe:$2a$10$zjpmRWC1Ld5yXHOIvUwOkeZNppT6ohujkfaoqN0RUMpNw1N0KIOoS
elastic-internal:$2a$10$e8VJ2lhoiGbcpfIQGwbTDOsLrpCxgIfHsfNaME0wBdRsyvxgKvoKG
elastic:$2a$10$1MgThAe6jhfjPcvLZotVM.yhiNMDPXwWrtzqWEX6FaiYcZ5eFy6ma
```

pebrc:
We previously set owner references only on creation as they usually don't change. With the association model in ECK ownership can change though. The referenced resource can change and the controller e.g. the Elasticsearch controller for a Kibana -> Elasticsearch relationship relies on the owner referrence in its watch on secrets.

This PR takes this into account by resetting the owner reference on updates.

Fixes #3470

### DIFF
--- a/pkg/controller/common/reconciler/reconciler.go
+++ b/pkg/controller/common/reconciler/reconciler.go
@@ -161,16 +161,16 @@ func ReconcileResource(params Params) error {
 			return err
 		}
 
-		expectedMeta, err := meta.Accessor(params.Expected)
-		if err != nil {
-			return err
-		}
 		// retain the resource version to avoid unconditional updates
 		resourceVersion := reconciledMeta.GetResourceVersion()
 		params.UpdateReconciled()
 		// and set the resource version back into the resource to indicate the state we are basing the update off of
 		reconciledMeta.SetResourceVersion(resourceVersion)
 		// also keep the owner references up to date
+		expectedMeta, err := meta.Accessor(params.Expected)
+		if err != nil {
+			return err
+		}
 		expectedOwners := expectedMeta.GetOwnerReferences()
 		if expectedOwners != nil {
 			// we can safely assume we have just one reference here given that it was created just above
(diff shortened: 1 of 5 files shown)


pebrc:
> I was wondering why we are not reusing the logic of the ReconcileResource function:

I think that is a fair point. I guess the only problem with that approach is that we would have to remember at each use of the `ReconcileResource` function to include the `OwnerReference` in both the check whether an update needs to happen and in the actual update. My thinking was that thats somewhat inconsistent with the current signature of the function as it takes and `owner` as an argument which sort of indicates that it will handle the `OwnerReference` for the caller. 

Regarding the question of whether we would miss the owner upgrade if it is not included in the `NeedsUpdate` function: I was kind of relying on the fact that if the owner changes this typically has an effect on other attributes as well (labels for example) which would make sure that the update happens. 

I don't feel super strongly about both points and am happy to reconsider if we think an explicit inclusion of the owner reference in `NeedsUpdate` and `UpdateReconciled` is preferable.

PR #3582, opened 2020-08-04
Add a container suffix flag to the operator

### LINKED ISSUE #3552: Support UBI based stack images
We want to support UBI based stack images in ECK. This affects mainly the defaulting mechanism which takes effect if a user does not explicitly specify a Docker image inside the `podTemplate` section of the manifest. 

The following options come to mind: 
- an additional operator flag (flag inflation!) to control the image default (suffix) e.g. `docker.elastic.co/elasticsearch/elasticsearch-ubi7:7.6.0` instead of `docker.elastic.co/elasticsearch/elasticsearch:7.6.0`
   * adds a bit of complexity around what the effective docker image is that ECK will pull given that we already allow an override of the registry.
   * probably the most pragmatic solution
- a configuration file mounted which has a value for each of the applications we support 
   * would add additional value for users that want to override the default images with their own custom images without the need to specify them fully in every manifst
   * more generic solution more work

pebrc:
Fixes #3552 

Support a suffix to container names to allow ECK to default to a different set of images (unless the user specifices the image name explicitly)


charith-elastic (approved):
LGTM

PR #1989, opened 2019-10-14
Mitigate memory leaks from long RequeueAfter periods

sebgl:
Fixes #1984.

We have an issue where the underlying timer used by client-go worker
queue implementation stays in memory until it expires.
Since one gets created at every reconciliation attempt, we end up with a
big bunch of timers in memory that will expire in 365 days by default.

To mitigate the memory leak, let's wait for no more than 10 hours to
reconcile. This does not prevent timers to leak, but restricts it to the next
10 hours.

This is done at the level of the aggregated results, to decouple this
workaround from any business logic like certs expiration. As opposed
to https://github.com/elastic/cloud-on-k8s/pull/1988 (closed).

The PR also refactors slightly the license controller to rely on the aggregated
Results structure. Hence benefit from the max 10 hour restriction.

PR #3585, opened 2020-08-05
Remove memory leak mitigation

barkbay:
Revert #1989 as [the fix for the memory leak is included since the version 0.17.0 ](https://github.com/kubernetes/client-go/blob/v0.17.0/util/workqueue/delaying_queue.go) of the K8S go client.

### DIFF
--- a/pkg/controller/common/reconciler/results.go
+++ b/pkg/controller/common/reconciler/results.go
@@ -6,16 +6,12 @@ package reconciler
 
 import (
 	"context"
-	"time"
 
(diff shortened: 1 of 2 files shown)


pebrc:
I believe there is little reason for scheduling a requeue for more than 10 hours in the future as the Go client cache refresh interval is by default 10 hours IIRC

barkbay:
> I believe there is little reason for scheduling a requeue for more than 10 hours in the future as the Go client cache refresh interval is by default 10 hours IIRC

I don't know if waiting for the cache to be refreshed is enough, it would also need to generate an event which would not be the case if the cache is in sync ?

PR #3589, opened 2020-08-05
Garbage collect orphaned Beat users

barkbay:
This PR ensures that orphaned Beat users are deleted on startup.


barkbay:
jenkins test this please

Issue #3586, opened 2020-08-05
e2e-tests-aks job fails while creating aks cluster


thbkrkr:
Last 3 executions of the [`cloud-on-k8s-e2e-tests-aks`](https://devops-ci.elastic.co/view/cloud-on-k8s/job/cloud-on-k8s-e2e-tests-aks/) job failed while creating the k8s cluster with the deployer
 and we had no notifications despite #3508.

We have a cascading 3 errors:
- the deployer failed to create the AKS cluster
- the e2e-tests xml report generation failed because there was no e2e-tests.json input file, due to the previous failure
- the Slack notification failed because the global variable `lib` used to load reusable function like `generateSlackMessage()` did not have time to be loaded, due to the previous failure

```
02:23:03  + make -C .ci get-test-artifacts TARGET=ci-e2e ci
..
02:24:59  ./hack/deployer/deployer execute --plans-file hack/deployer/config/plans.yml --config-file deployer-config.yml
02:24:59  2020/08/04 00:24:53 Authenticating as service account...
02:24:59  Error: exit status 1
02:24:59  exit status 1
02:24:59  make: *** [Makefile:440: run-deployer] Error 1
...
02:24:56  + make -C .ci TARGET=e2e-generate-xml ci
..
02:24:58  cat: e2e-tests.json: No such file or directory
02:24:58  
02:24:58  DONE 0 tests, 1 error in 0.001s
02:24:58  Failed to generate xml report
02:24:58  cat: e2e-tests.json: No such file or directory
02:24:58  make: *** [Makefile:400: e2e-generate-xml] Error 1
...
02:25:01  [Pipeline] { (Declarative: Post Actions)
..
02:25:02  Error when executing unsuccessful post condition:
02:25:02  java.lang.NullPointerException: Cannot invoke method generateSlackMessage() on null object
02:25:02  	at org.codehaus.groovy.runtime.NullObject.invokeMethod(NullObject.java:91)
...
```

thbkrkr:
The deployer fails authentication due to expired client secret keys:
- `az login --service-principal -u {{.AppId}} -p {{.TenantSecret}} --tenant {{.TenantId}}"` returns:

```json
Get Token request returned http error: 401 and server response: 
{
    "error": "invalid_client",
    "error_description": "AADSTS7000222: The provided client secret keys are expired. 
                          Visit the Azure Portal to create new keys for your app, or consider using 
[...]

Issue #3560, opened 2020-07-29
cloud-on-k8s-e2e-tests-ocp unstable


thbkrkr:
Since build # 183 (23 juil. 2020 00:19:59), we had 3 times the job aborted after 10h ([600 minutes](https://github.com/elastic/cloud-on-k8s/blob/6be658ee4883dffa31355ea5cd8b3bb5d4698541/.ci/pipelines/e2e-tests-ocp.Jenkinsfile#L15), the job timeout). 

![2020-07-29-152025_250x197_scrot](https://user-images.githubusercontent.com/582883/88805127-0fa25400-d1af-11ea-8b96-9fde9001bb25.png)

The `make -C .ci TARGET=e2e-generate-xml ci` stage is blocked and times out. Normally, this stage takes 2 seconds.

![2020-07-29-150923_771x366_scrot](https://user-images.githubusercontent.com/582883/88804092-b8e84a80-d1ad-11ea-9b50-1e3079c66735.png)

Log of a failed run:
```java
[2020-07-29T05:25:21.868Z] + make -C .ci TARGET=e2e-generate-xml ci
[2020-07-29T05:25:21.868Z] make: Entering directory '/var/lib/jenkins/workspace/cloud-on-k8s-e2e-tests-ocp/.ci'
[2020-07-29T05:25:21.868Z] make[1]: Entering directory '/var/lib/jenkins/workspace/cloud-on-k8s-e2e-tests-ocp/.ci'
[2020-07-29T05:25:22.917Z] docker.elastic.co/eck-ci/eck-ci-tools:e514e78a
[2020-07-29T05:25:23.880Z] ✓  test/e2e (15m25.855s)
[2020-07-29T05:25:23.880Z] ✓  test/e2e/apm (11m37.496s)
[2020-07-29T05:25:23.880Z] ✓  test/e2e/beat (25m6.075s)
[2020-07-29T05:25:23.880Z] ∅  test/e2e/cmd
[2020-07-29T05:25:23.880Z] ✓  test/e2e/cmd/run (26ms)
[2020-07-29T05:25:23.880Z] ✓  test/e2e/ent (17m56.864s)
// >>>>> The job is blocked here <<<<<
[2020-07-29T10:24:52.239Z] Sending interrupt signal to process
[2020-07-29T10:25:02.432Z] make: *** [Makefile:400: e2e-generate-xml] Terminated
[...]

PR #3590, opened 2020-08-05
Improve JUnit E2E tests XML report generation

thbkrkr:
Add a script to call gotestsum for the JUnit E2E tests XML report
generation that will check that an input file exists and also filter
out invalid lines containing a space in the timestamp.

I think a bash script is a better place than a Makefile to implement this kind of logic even if it is quite small.

Relates to #3560 for the temporary fix with the filter.
Relates to #3586, avoiding generation when there is no input file will clean up our jobs log of the somewhat disturbing error
 `cat: e2e-tests.json: No such file or directory` due to a previous failure during the e2e tests execution.

### DIFF
--- a/hack/ci/generate-junit-xml-report.sh
+++ b/hack/ci/generate-junit-xml-report.sh
@@ -1,0 +1,37 @@
+#!/usr/bin/env bash
+
+# Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+# or more contributor license agreements. Licensed under the Elastic License;
+# you may not use this file except in compliance with the Elastic License.
+
+# Script to generate a JUnit XML report from a JSON go test output.
+#
+
+set -euo pipefail
+
+main() {
+    local input_file=${1:-"input file"}
+    local xml_report=${input_file%.*}.xml
+
+    # exit without error when there is no input file
+    if [[ ! -f $input_file ]]; then
+        echo "No $input_file to generate a JUnit XML report."
+        exit 0
+    fi
+  
+    # temporary filter out lines containing a space in the timestamp,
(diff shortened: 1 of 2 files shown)


hack/ci/generate-junit-xml-report.sh
```diff
@@ -0,0 +1,37 @@
...
+    if [[ ! -f $input_file ]]; then
+        echo "No $input_file to generate a JUnit XML report."
+        exit 0
+    fi
+  
+    # temporary filter out lines containing a space in the timestamp,
+    # see https://github.com/elastic/cloud-on-k8s/issues/3560.
+    gotestsum \
+        --junitfile "$xml_report" \
+        --raw-command grep -v '"Time":"[^"]*\s[^"]*"' "$input_file" || \
```
barkbay, line 26: I'm wondering if it's worth parsing the whole timestamp, maybe a bit less readable but it would also prevent some other kinds of timestamp "corruption":

```suggestion
        --raw-command grep -P '"Time":"(?:[1-9]\d{3}-(?:(?:0[1-9]|1[0-2])-(?:0[1-9]|1\d|2[0-8])|(?:0[13-9]|1[0-2])-(?:29|30)|(?:0[13578]|1[02])-31)|(?:[1-9]\d(?:0[48]|[2468][048]|[13579][26])|(?:[2468][048]|[13579][26])00)-02-29)T(?:[01]\d|2[0-3]):[0-5]\d:[0-5]\d(?:\.\d{1,9})?(?:Z|[+-][01]\d:[1-5]\d)"' "$input_file" || \
```

thbkrkr: Yes but I think it would be interesting to detect different timestamp corruption for now. So I'm in favor of keeping the regexp simpler.

PR #3620, opened 2020-08-13
[DOC] Volume claim templates

alaudazzi:
Minor editing suggestions.

### DIFF
--- a/docs/orchestrating-elastic-stack-applications/elasticsearch/volume-claim-templates.asciidoc
+++ b/docs/orchestrating-elastic-stack-applications/elasticsearch/volume-claim-templates.asciidoc
@@ -47,4 +47,4 @@ spec:
           emptyDir: {}
 ----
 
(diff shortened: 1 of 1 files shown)


docs/orchestrating-elastic-stack-applications/elasticsearch/volume-claim-templates.asciidoc
```diff
@@ -47,4 +47,4 @@ spec:
           emptyDir: {}
 ----
 
-CAUTION: Using `emptyDir` is not recommended due to the high likelihood of permanent data loss.
+CAUTION: Using `emptyDir` is not recommended as it might lead to permanent data loss.
```
charith-elastic, line 50: Is there a way to make this sound scarier? :smile: 
When I read "might lead to", my feeling is that it is not that dangerous. Is there a nicer way of saying "don't do this unless you absolutely know what you are doing"?

alaudazzi:
Relates to [#63068](https://github.com/elastic/cloud/issues/63068).

PR #3622, opened 2020-08-13
[DOC] Transport settings

alaudazzi:
Editing.
Relates to [#63068](https://github.com/elastic/cloud/issues/63068)

### DIFF
--- a/docs/orchestrating-elastic-stack-applications/elasticsearch/transport-settings.asciidoc
+++ b/docs/orchestrating-elastic-stack-applications/elasticsearch/transport-settings.asciidoc
@@ -26,4 +26,4 @@ spec:
 
 Check the https://kubernetes.io/docs/concepts/services-networking/service/#publishing-services-service-types[Kubernetes Publishing Services (ServiceTypes)] that are currently available.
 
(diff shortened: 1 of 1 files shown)


docs/orchestrating-elastic-stack-applications/elasticsearch/transport-settings.asciidoc
```diff
@@ -26,4 +26,4 @@ spec:
 
 Check the https://kubernetes.io/docs/concepts/services-networking/service/#publishing-services-service-types[Kubernetes Publishing Services (ServiceTypes)] that are currently available.
 
-NOTE: Please note that when you change the `clusterIP` setting of the service, ECK will delete and re-create the service as `clusterIP` is an immutable field. This does not typically have an impact on connectivity as the transport module uses long-lived TCP connections, but may cause a small network disruption between Elasticsearch nodes.
+NOTE: When you change the `clusterIP` setting of the service, ECK deletes and re-creates the service, as `clusterIP` is an immutable field. Typically this does not affect connectivity, as the transport module uses long-lived TCP connections, but may cause a small network disruption between Elasticsearch nodes.
```
charith-elastic, line 29:
```suggestion
NOTE: When you change the `clusterIP` setting of the service, ECK deletes and re-creates the service, as `clusterIP` is an immutable field. This will cause a short network disruption, but in most cases it should not affect existing connections as the transport module uses long-lived TCP connections. 
```

Issue #2032, opened 2019-10-21
Operator and cluster licensing as a unit


nkvoll:
### Background requirements / constraints / input:
- Operator and clusters have a 1:1 relationship when it comes to licensing. E.g enterprise operator = enterprise cluster.
- Need to make sure each cluster is only managed by a single operator at a time (no conflicts)
- Be able to go from basic <-> enterprise license. (soft requirement on the operator itself, required for clusters, being able to change the license on a cluster without re-indexing/downtime etc is important)
- Don't have to think about licensing by default before deploying a cluster (soft requirement, but would be nice from a UX perspective): you get what the operator is licensed with.

### Proposal:
 1. An optional setting which gives the operator an identifier. This defaults to `default`.

To note: It would be a *semantic error* to deploy two *overlapping* operators in the same cluster with the same operator identifier, defaulted or not -- this results in conflicts.

To note 2: Having leader election between these operators are not exactly what we'd like because it's a semantic error, and having one of them run, the other one on hold is not desired (e.g re-election potentially changes behavior). Should both of them be stopped somehow, then require an admin action to be taken? What would the procedure to resolve a conflict be in order to resume operation?

 2. An optional annotation that can be used to specify which operator should be managing the resource. This defaults to `default`.

## Example:

Two operators running with two different licenses, watching the same (or overlapping) namespaces.

```yaml
## ECK Operator, Enterprice licensed (uses default identifier: default)
apiVersion: apps/v1
kind: StatefulSet
metadata:
  name: elastic-operator
#...
---
## ECK Operator, Basic licensed (custom identifier: my-basic)
apiVersion: apps/v1
kind: StatefulSet
metadata:
  name: elastic-operator-basic
#...
spec:
  template:
    spec:
      containers:
      - image: <OPERATOR_IMAGE>
        name: manager
        env:
          - name: OPERATOR_IDENTIFIER
            value: my-basic
        #...
      #...
    #...
  #...
#...
---
## Elasticsearch cluster
apiVersion: elasticsearch.k8s.elastic.co/v1beta1
kind: Elasticsearch
metadata:
  name: elasticsearch-sample
  annotations:
    ## if not set, is "default"
    #operator.k8s.elastic.co/managed-by: default
    ## can specify which operator:
    operator.k8s.elastic.co/managed-by: my-basic
spec:
  version: 7.4.0
```

uric:
[...]

Issue #2012, opened 2019-10-17
Test operator behavior after failover


david-kow:
As a stateless service we expect the operator to be resilient to downtime, but we don't currently test for it. We should at least have a sanity test around correctness of what operator is doing when being randomly restarted. Somewhat related to general chaos testing.

anyasabo:
Related: https://github.com/elastic/cloud-on-k8s/issues/709

david-kow:
We should create a separate test pipeline where we would randomly delete the operator pod while the entire E2E test suite is running. This way we don't introduce noise into existing pipelines and we have a good coverage of different test scenarios. We should log and preserve timestamps when the operator pod is being removed for potential issues investigation.

barkbay:
Fixed by #3706

PR #3632, opened 2020-08-14
Add operator election support

### LINKED ISSUE #709: Support leader election to be more flexible about operator deployment
We currently rely on stateful sets which is the default in kubebuilder. Leader election could allow us to be a bit more flexible wrt to the k8s resource and allow the use of the operator in environments that only support deployments for operators.

barkbay:
This PR adds and enables an election mechanism between several running operators. It allows the user to scale up the number of operator instances that are running at the same time.

Fixes #709


docs/operating-eck/operator-config.asciidoc
```diff
@@ -23,6 +23,7 @@ ECK can be configured using either command line flags or environment variables.
 |debug-http-listen |localhost:6060 |Listen address for the debug HTTP server. Only available in development mode.
 |development |false |Enable development mode. Only available as a CLI flag.
 |disable-config-watch| false| Watch the configuration file for changes and restart to apply them. Only effective when the `--config` flag is used to set the configuration file.
+|enable-leader-election | true | Enable leader election. Enabling this will ensure there is only one active operator.
```
anyasabo, line 26:
```suggestion
|enable-leader-election | true | Enable leader election. Must be set to true if using multiple replicas of the operator
```

anyasabo:
Should we change the defaults to increase the number of replicas, or at least change it for e2e tests to make sure this gets used? It doesn't look like we hardcode references to `elastic-operator-0` outside of one doc

anyasabo:
I'd also argue this counts as a `>feature` not `>enhancement`

barkbay:
> Should we change the defaults to increase the number of replicas, or at least change it for e2e tests to make sure this gets used?

Regarding the defaults in the all-in-one I was wondering the same. I think I would be 👍 to increase it to something like 2 and add a variable in the manifest template.

Regarding our e2e tests I was wondering if https://github.com/elastic/cloud-on-k8s/issues/2012 would not be a better place to do that because I'm not sure that our current e2e tests will give us any insight from this feature.

> I'd also argue this counts as a >feature not >enhancement

If you already tried to increase the number of replicas, wondering why some reconciliations look "weird" ... then I think it looks like an enhancement 😄 . I'm also fine with the  `>feature` flag.

anyasabo:
>Regarding the defaults in the all-in-one I was wondering the same. I think I would be 👍 to increase it to something like 2 and add a variable in the manifest template.

👍 and point taken on the e2e testing bit, it would still eventually work without this so we would really only be catching the case where theyre deadlocked.

david-kow:
I have some questions regarding this change:
- Why someone would use this feature? IIUC, when a primary fails, a secondary will be promoted and will take over. What is the improvement with that behaviour over just starting a new Pod? If it's just time to start, do we know the actual improvement? (I'm asking since it seems to me that it would be minimal - ~0s vs few s). 
- Did we get any requests for this?
- What are the failure scenarios that this mechanism handles on it's own? Are operator pods communicating directly? What if network partition happens and both new and old primary has access to k8s APIs and are still running?

I think that if we want to add this (especially as the default), we should have it extensively tested.

barkbay:
> * Why someone would use this feature? IIUC, when a primary fails, a secondary will be promoted and will take over. What is the improvement with that behaviour over just starting a new Pod? If it's just time to start, do we know the actual improvement? (I'm asking since it seems to me that it would be minimal - ~0s vs few s).

I'm not sure what you mean by "starting a new Pod", if a container inside a Pod is crashing because of a container runtime failure, a config. error or some network issues then the kubelet will try to restart the container, but the Pod itself will stay in a `CrashLoopBackOff` state.

> * Did we get any requests for this?

Not to my knowledge, the idea here is to prevent a user to start simultaneously several operators by increasing the number of replicas. I think it should also cover the case where a Pod is in an `Unknown` state because the kubelet is partitioned and a controller attempts to start a new instance.

> * What are the failure scenarios that this mechanism handles on it's own? Are operator pods communicating directly? What if network partition happens and both new and old primary has access to k8s APIs and are still running?
> 
> I think that if we want to add this (especially as the default), we should have it extensively tested.

The leader election mechanism is based on the [ResourceLock](https://github.com/kubernetes-sigs/controller-runtime/blob/v0.6.2/pkg/leaderelection/leader_election.go#L51), [it already has its own unit tests](https://github.com/kubernetes/client-go/blob/master/tools/leaderelection/leaderelection_test.go). The overall mechanism is used by the Kubernetes project itself for the [election mechanism of the K8S scheduler](https://github.com/kubernetes/kubernetes/blob/master/cmd/kube-scheduler/app/options/options.go#L307) or for some [built-in](https://github.com/kubernetes/kubernetes/blob/master/cmd/kube-controller-manager/app/controllermanager.go#L272) [controllers](https://github.com/kubernetes/kubernetes/blob/master/cmd/cloud-controller-manager/app/controllermanager.go#L198). I agree that it should be tested but as mentioned above I think it should be done as part of #2012

There is no communication between the Pods, they are competing to apply an annotation on a `ConfigMap`, e.g.: 

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  annotations:
    control-plane.alpha.kubernetes.io/leader: '{"holderIdentity":"fcf2a873-8018-43c0-9196-13293a65d9ef","leaseDurationSeconds":15,"acquireTime":"2020-08-19T09:29:49Z","renewTime":"2020-08-19T09:30:15Z","leaderTransitions":0}'
  creationTimestamp: "2020-08-19T09:29:49Z"
  name: elastic-operator-leader
  namespace: elastic-system
  resourceVersion: "2331024"
  selfLink: /api/v1/namespaces/elastic-system/configmaps/elastic-operator-leader
  uid: 9eddb203-16c1-4751-9990-a3f61894cb5c
```

david-kow:
> I'm not sure what you mean by "starting a new Pod"

I didn't phrase it well, I meant what you described. 

It all does sound convincing, but it seems to be more a preventive measure for users that added more replicas rather than a reliability improvement, correct? I initially thought the goal is increasing the reliability. 

> I agree that it should be tested but as mentioned above I think it should be done as part of #2012

Should we wait with this until #2012 is in then?

barkbay:
> it seems to be more a preventive measure for users that added more replicas rather than a reliability improvement, correct?

It prevents the operator to start if not elected. By doing that a user can safely run several instances at the same time. It improve the availability, so I would say it is an improvement ?

> Should we wait with this until #2012 is in then?

Current e2e tests will ensure that an election happens in normal situations. As mentioned previously we are relying on a feature of the controller runtime, so my feeling was that it is a nice improvement and there is no urgency to add some additional e2e tests.

The only point I'm not sure is that should we keep the default number of replicas to 1 for now, it would make it easier for the user to get the logs. If we want to set the default value to 2 I think it makes sense to also do that for e2e tests as suggested by @anyasabo

pebrc:
> The only point I'm not sure is that should we keep the default number of replicas to 1 for now, it would make it easier for the user to get the logs. If we want to set the default value to 2 I think it makes sense to also do that for e2e tests as suggested by @anyasabo

I agree that we should probably keep the number of replicas at 1 but excercise the new functionality in an e2e test.

cmd/manager/main.go
```diff
@@ -163,6 +165,11 @@ func Command() *cobra.Command {
 		false, // Set to false for backward compatibility
 		"Restrict cross-namespace resource association through RBAC (eg. referencing Elasticsearch from Kibana)",
 	)
+	cmd.Flags().Bool(
+		operator.EnableLeaderElection,
+		true,
```
sebgl, line 170: Do you see any reason for disabling it? (other than: we discovered it's broken/buggy)

cmd/manager/main.go
```diff
@@ -359,8 +366,11 @@ func startOperator(stopChan <-chan struct{}) error {
 
 	// Create a new Cmd to provide shared dependencies and start components
 	opts := ctrl.Options{
-		Scheme:  clientgoscheme.Scheme,
-		CertDir: viper.GetString(operator.WebhookCertDirFlag),
+		Scheme:                  clientgoscheme.Scheme,
+		CertDir:                 viper.GetString(operator.WebhookCertDirFlag),
+		LeaderElection:          viper.GetBool(operator.EnableLeaderElection),
```
sebgl, line 371: I'm trying to think about the problem having 2 operators with leader election enabled might cause.

The major (only?) one I see is that I think there's a possible small time window where 2 instances of the operators may run concurrently:
- the old leader, after it lost the lease but before it shuts down completely (it is supposed to shut down almost immediately but well, you never know which goroutine gets priority) - at this point it may still be creating Pods or interacting with Elasticsearch cluster settings
- the new leader, right after it got the lease

IIUC there's an [extra 5sec](https://github.com/kubernetes-sigs/controller-runtime/blob/2d69b1e99c8bf01ab8dfe9b45474f06e0129894e/pkg/manager/internal.go#L49-L52) to prevent this from happening but I don't think there's any stronger guarantee.

In that small time period, expectations from the 1st operator may not be satisfied in the 2nd one: 1st operator creates a master and 2nd one does not see it yet. As part of startup the 2nd operator (new leader) waits for its cache to be synced before starting to reconcile but we may still hit the race condition described above, I think.

Anyway, it's better to have leader election enabled than not have anything at all.

I'm wondering if this race condition is unlikely enough to happen so we can ignore it, and switch to a Deployment instead of a StatefulSet for the operator.

hack/manifest-gen/assets/charts/eck/values.yaml
```diff
@@ -10,9 +10,11 @@ operator:
     repository: docker.elastic.co/eck/eck-operator
     pullPolicy: IfNotPresent
 
+  replicas: 2
```
sebgl, line 13: let's default to a single replica?

cmd/manager/main.go
sebgl: Related issue in operator-sdk: https://github.com/operator-framework/operator-sdk/issues/136
It looks like they implemented an additional election mechanism with different constraints: https://github.com/operator-framework/operator-sdk/pull/1052/files

cmd/manager/main.go
```diff
@@ -359,8 +366,11 @@ func startOperator(stopChan <-chan struct{}) error {
...
+		LeaderElectionID:        LeaderElectionConfigMapName,
+		LeaderElectionNamespace: operatorNamespace,
```
pebrc, line 373: I wonder if it is possible to deploy multiple independent instances of the operator, responsible for different namespaces or sets of namespaces via OLM that would all run in the `operators` namespace by default IIRC which would then all try to use the same `ConfigMap`. Is this possible or are there additional safeguards in place in the leader election algorithm, e.g. some form of unique identifier per operator id. If not, should we try to include the operator UUID as an ID to prevent that?

sebgl: 👍  I think it's a good argument for making the leader election config map name configurable.

anyasabo: I don't think we support multiple operators in the same namespace today, but you're right there's an argument that we should and maybe we should start here. Off the top of my head we need to make this variable:
```
	// licensingCfgMapName is the name of the config map used to store licensing information
	licensingCfgMapName = "elastic-licensing"
```

barkbay:
Thanks for all the inputs ! I have tried to sum my thoughts on the different topics expressed here:

## Making the leader election config map name configurable

IIUC the intention here is to be able to run several operators with distinct responsibilities in a shared namespace.
I don't think it is possible right now:
* The ConfigMap which contains the UUID is generated as a per namespace basis. I think I would expect to have distinct UUIDs for each group of operators to be consistent with what would happen if they were deployed in different namespaces.
* I believe that the licensing artifacts are also tied to the namespace in which Pods are running, not to a group of operator instances.

Regarding OLM: I'm not sure it is supported, or at least it does not seem to be the first intention. The documentation seems to indicate that an operator group is tied to a namespace:

> An Operator is considered a member of an OperatorGroup if the following conditions are true:
>  * The Operator’s CSV exists in the same namespace as the OperatorGroup.
>  * The Operator’s CSV’s InstallModes support the set of namespaces targeted by the OperatorGroup.

Maybe we should discuss if we want this feature in a dedicated issue ? (_Kind of related to https://github.com/elastic/cloud-on-k8s/issues/2032_)

## Concurrent execution

While it is possible to have an algorithm which ensures that there is at most one leader at a given point in time, it is not possible to enforce this condition at the workload level (i.e. threads/go routines may still temporarily live in a process which is not the leader anymore). Even the "leader-for-life" election mechanism has its own [edge cases](https://github.com/operator-framework/operator-sdk/pull/481#discussion_r218259908).

I would consider that:
* An election mechanism based on leases is enough for now.
* The manager will quickly shut down the controllers and [the workqueues](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/internal/controller/controller.go#L131). I don't think any slow I/O would interfere here.

## Testing

I agree that we should test this feature to some extent, but I'm not sure we should expect to have it extensively tested. Election mechanisms are mostly disrupted by slow I/O, high latencies, clock screw [rate] inconsistencies, partition conditions on the host or on the process ... I'm not sure it is easy to simulate those conditions in an e2e environment _(at least without adding some noise to our e2e tests suite)_. Killing Pods randomly would be a first step but my feeling is that it would just be happy path testing. Let's follow up this discussion in https://github.com/elastic/cloud-on-k8s/issues/2012 for which I'll try to outline a proposal.

cmd/manager/main.go
sebgl: I think I agree with what @barkbay said in https://github.com/elastic/cloud-on-k8s/pull/3632#issuecomment-680764147: perfect leader election is hard and we probably don't need it here. The current mechanism seems good enough for what we want to achieve.

cmd/manager/main.go
sebgl: Discussed through another channel: it's aimed to be an escape hatch in case we reach a deadlock in the leader election process, for some reasons.

sebgl (approved):
LGTM once we set back the number of replicas to 1 by default in the manifest.
I think I see this feature as an additional safeguard in case multiple operators run in parallel, more than a real feature that allows you to achieve HA.

About the points mentioned in the comments:
* 👍 for considering making the config map name configurable in a later stage, along with other hardcoded resource names
* I would vote for not spending too much time writing complicated tests to cover the leader election process, since we just reuse it from client-go and controller-runtime where it is already tested. I think I'm actually OK with no tests at all on our side.

sebgl (approved):

barkbay:
I'm merging this one and will quickly follow up with some e2e tests.

PR #3633, opened 2020-08-14
Add DeploymentStrategy to Beats CRDs

### LINKED ISSUE #3485: Beat might not be able to start after update
When:
- Beat is already running, and
- Beat is deployed as Deployment, and
- Beat is updated, and
- New Pod lands on the same Node as old Pod, and
- the timing is right, then

new Beat will keep crashing with the below.
```
ERROR   instance/beat.go:958    Exiting: data path already locked by another beat. Please make sure that multiple beats are not sharing the same data path (path.data).
```

(We want the new Pod to use the same path to preserve Beat identity.)

What _could_ be a solution is to allow modifying `maxUnavailable` for deployment update strategy, but we don't expose that currently. Right now, the paths to unblock would be:
- remove ReplicaSet that contains old Pods (requires manual intervention), or
- use emptyDir for `beat-data` volume (Beat won't preserve its identity):
```
volumes:
- name: beat-data
  emptyDir: {}
```

pebrc:
Fixes #3485 

* adds and exposes `appsv1.DeploymentStrategy` to the `DeploymentSpec` in the Beats CRD
* adds an e2e test to verify that only one version of Beats is running when using the `Recreate` deployment strategy

Note:  we could make this test even closer to the scenario #3485 describes by using `nodeAffinity` but it would complicate the test setup quite a bit and the test seems to reliably fail when not using `Recreate`


pebrc:
Jenkins test this please

test/e2e/beat/upgrade_test.go
```diff
@@ -0,0 +1,56 @@
...
+
+	fbBuilder = beat.ApplyYamls(t, fbBuilder, E2EFilebeatConfig, E2EFilebeatPodTemplate)
+
+	test.RunMutationsWhileWatching(
+		t,
+		[]test.Builder{esBuilder, fbBuilder},
+		[]test.Builder{esBuilder, fbBuilder.WithVersion(dstVersion)},
+		[]test.Watcher{test.NewVersionWatcher(beatcommon.VersionLabelName, opts...)},
+	)
+}
```
sebgl, line 56: I'm wondering if this test is necessary. Are we testing that the k8s Deployment `Recreate` strategy works as expected?
Instead we could just check the created Deployment resource and ensure it contains the expected DeploymentStrategy (as part of beats k8s checks)?

pebrc: Yes that is much simpler. I guess I did it this way because we did not have a single test that ran an upgrade with Beats.

sebgl:
I think I would consider this a `feature` more than a bug. It's a nice-to-have feature for other resources: EnterpriseSearch, APM, etc.
It happens to be a workaround to manually solve https://github.com/elastic/cloud-on-k8s/issues/3485 which may be considered a bug, but does not really fix the bug by itself.

pebrc:
@sebgl are you suggesting to extend this to `EnterpriseSearch` and `APM` in this PR as well?

sebgl:
> @sebgl are you suggesting to extend this to EnterpriseSearch and APM in this PR as well? (Happy to, just making sure I understand what you mean)

Oh sorry I thought it was the case already by quickly looking at the modified files. 
I think it would be a good idea? Kibana is a bit tricky though since we manipulate it during version upgrades. Let's keep this PR as is and open a different issue maybe?

test/e2e/beat/upgrade_test.go
```diff
@@ -0,0 +1,56 @@
...
+
+	name := "test-beat-upgrade-to-7x"
+	esBuilder := elasticsearch.NewBuilder(name).
+		WithESMasterDataNodes(1, elasticsearch.DefaultResources).
+		WithVersion(dstVersion)
+
+	fbBuilder := beat.NewBuilder(name).
+		WithRoles(beat.PSPClusterRoleName, beat.AutodiscoverClusterRoleName).
+		WithType(filebeat.Type).
+		WithDeployment().
```
barkbay, line 34:
```suggestion
```
`WithDeploymentStrategy()` already implies `WithDeployment()`

barkbay (approved):
LGTM

test/e2e/beat/upgrade_test.go
```diff
@@ -0,0 +1,56 @@
...
+		WithDeploymentStrategy(appsv1.DeploymentStrategy{
```
sebgl, line 35: Can we add a comment explaining why we're doing that?

test/e2e/beat/upgrade_test.go
```diff
@@ -0,0 +1,56 @@
...
+		}),
+	}
+
+	fbBuilder = beat.ApplyYamls(t, fbBuilder, E2EFilebeatConfig, E2EFilebeatPodTemplate)
+
+	test.RunMutationsWhileWatching(
+		t,
+		[]test.Builder{esBuilder, fbBuilder},
+		[]test.Builder{esBuilder, fbBuilder.WithVersion(dstVersion)},
+		[]test.Watcher{test.NewVersionWatcher(beatcommon.VersionLabelName, opts...)},
```
sebgl, line 54: It could be nice to add a comment explaining why we're checking this.

sebgl (approved):
LGTM, I think a few comments would help me remember what this is about in 6 months from now.

test/e2e/beat/upgrade_test.go
pebrc: Added comment on top to explain rationale of this test.

PR #3635, opened 2020-08-17
Hide development-mode flags

charith-elastic:
Development mode flags should not be documented as we do not expect/want them to used by end users. We have had at least one case of a user trying to set a development mode flag because it was documented in the operator configuration page.


anyasabo:
I think we may want to leave these documented so that people hacking on it can know of their existence. I'd prefer us have it be as easy as possible for new contributors to get a dev environment up and running. Maybe we just need to make sure the workflow is documented in `CONTRIBUTING.md`?

alaudazzi (approved):
LGTM

PR #3642, opened 2020-08-20
Bump test versions after 7.9 release

pebrc:
Adjusts the max 7.x version and the SNAPSHOT version we test against.


barkbay (approved):
LGTM

Should we also add a pipeline for `7.9.0` in `e2e-tests-stack-versions-gke` ?

Based on the [EOL policy](https://www.elastic.co/support/eol) I think we'll be able to remove `7.1` in november.

Issue #3431, opened 2020-07-14
Webhook ignores namespace restriction


pebrc:
ECK allows restricting its responsibilities to a set of namespaces or even a single namespace. The ValidatingAdmissionWebhook we deploy is unaware of this restriction. That means in environments where multiple instances of ECK are running all instances will validate all requests in parallel. If the user is running different (minor) versions of ECK at the same time (e.g. during an upgrade process of the operator fleet) they might come to different conclusions as to what is admissible.

There is currently no mechanism in Kubernetes to restrict a webhook to just a subset of resources in a way that is useful for ECK: 
* `namespaceSelector` allows limiting the namespaces a webhook is responsible for which sounds like what we would want. But it  requires the namespaces to be annotated with labels, which would require additional permissions in the operator process
* `objectSelector` allows limiting the objects to be validated using labels on the resources themselves, we could introduce an additional operator identifier label and restrict the webhook based on that.

A third option that comes to mind is just implementing the restriction in the webhook itself: as we cannot easily prevent all webhooks from being called we could short-circuit that call in all those webhooks belonging to ECK instances that are not responsible for the namespace in question.

PR #3645, opened 2020-08-20
Update webhook documentation and manifests

### LINKED ISSUE #3638: Improve Webhook server documentation
The [current documentation](https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-webhook.html) does not clearly describe the different ways to setup and configure the webhook server:

* Let's make it clear that by default ECK automatically manages the certificates
* [Using the cert manager](https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-webhook-cert-manager.html) is only an option
* Let's also add a sample to show how to manually manage the certs:
  * add the CA to the `ValidatingWebhookConfig`
  * put the key and the certificate signed by the CA in the secret that the operator mounts

charith-elastic:
Restructures the webhook documentation to make it easier to understand. 

- Explain the defaults and available configuration options
- Add information about using own certificates 
- Update manifest generator to support custom certificate scenarios
- Some sections have been removed and could result in broken links but I think the impact would be pretty low -- if any.
- As the full definition of the `ValidatingWebhookConfiguration` is quite long, I chose not to include it in the code samples and provided pointers to the manifest generator instead. 

Preview: https://cloud-on-k8s_3645.docs-preview.app.elstc.co/guide/en/cloud-on-k8s/master/k8s-webhook.html

Fixes #3638

### DIFF
--- a/docs/operating-eck/webhook.asciidoc
+++ b/docs/operating-eck/webhook.asciidoc
@@ -50,9 +50,9 @@ You can customise almost all aspects of the webhook setup by changing the <<{p}-
 |Configuration option | Default value |Description
 |enable-webhook       | false   | This must be set to `true` to enable the webhook server.
 |manage-webhook-certs | true    | Set to `false` to disable auto-generating the certificate for the webhook. If disabled, you must provide your own certificates using one of the methods described later in this document.
-|webhook-cert-dir     | /tmp/k8s-webhook-server/serving-certs | Path to mount the certificate
-|webhook-name         | elastic-webhook.k8s.elastic.co | Name of the `ValidatingWebhookConfiguration`. As webhook configurations are cluster resources (not restricted to a single namespace), this allows you to run multiple copies of the operator with each instance having its own webhook.
-|webhook-secret       | elastic-webhook-server-cert | Name of the secret containing the certificate for the webhook server
+|webhook-cert-dir     | /tmp/k8s-webhook-server/serving-certs | Path to mount the certificate.
+|webhook-name         | elastic-webhook.k8s.elastic.co | Name of the `ValidatingWebhookConfiguration` resource.
+|webhook-secret       | elastic-webhook-server-cert | Name of the secret containing the certificate for the webhook server.
 |===
 
 
--- a/docs/operating-eck/webhook.asciidoc
+++ b/docs/operating-eck/webhook.asciidoc
@@ -1,131 +1,191 @@
 :page_id: webhook
-[id="{p}-{page_id}"]
-= Configure validating webhook
+ifdef::env-github[]
+****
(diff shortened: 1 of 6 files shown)


docs/operating-eck/webhook.asciidoc
```diff
@@ -1,131 +1,191 @@
...
+You can customise almost all aspects of the webhook setup by changing the <<{p}-operator-config,operator configuration>>.
 
-Then, install cert-manager v0.11+ as described in the https://docs.cert-manager.io/en/latest/getting-started/install/[cert-manager documentation].
+[width="100%",cols=".^10m,.^30d,.^60d",options="header",grid=rows]
+|===
+|Configuration option | Default value |Description
+|enable-webhook       | false   | This must be set to `true` to enable the webhook server.
+|manage-webhook-certs | true    | Set to `false` to disable auto-generating the certificate for the webhook. If disabled, you must provide your own certificates using one of the methods described later in this document.
+|webhook-cert-dir     | /tmp/k8s-webhook-server/serving-certs | Path to mount the certificate
+|webhook-name         | elastic-webhook.k8s.elastic.co | Name of the `ValidatingWebhookConfiguration`. As webhook configurations are cluster resources (not restricted to a single namespace), this allows you to run multiple copies of the operator with each instance having its own webhook.
```
pebrc, line 54: I don't think this is fully correct. I a user were to do that would not all registered webhooks of all ECK instances be called for each submitted manifest?

I opened https://github.com/elastic/cloud-on-k8s/issues/3431 a while ago to improve this.

docs/operating-eck/webhook.asciidoc
```diff
@@ -1,131 +1,191 @@
...
+----
+
+- Create a secret in the namespace the operator would be deployed to. This secret must contain the certificate under the `tls.crt` key and the private key under the `tls.key` key.
++
+[source,sh]
+----
+kubectl create secret -n elastic-system generic elastic-webhook-server-custom-cert --from-file=tls.crt=tls.crt --from-file=tls.key=tls.key
+----
+
+- Encode your CA trust chain as base64 and set it as the value of `caBundle` fields in the `ValidatingWebhookConfiguration`. Note that there are multiple `caBundle` fields in the webhook configuration.
```
pebrc, line 88:
```suggestion
- Encode your CA trust chain as base64 and set it as the value of the `caBundle` fields in the `ValidatingWebhookConfiguration`. Note that there are multiple `caBundle` fields in the webhook configuration.
```
Not 100% sure about this but it feels like a word is missing here.

docs/operating-eck/webhook.asciidoc
charith-elastic: Good point!

pebrc (approved):

PR #3648, opened 2020-08-20
Update OCP deployer env vars

anyasabo:
We ran into some trouble setting @liladler up with an OpenShift cluster. `make switch-ocp bootstrap-cloud` was not using the env vars to populate the config file, even though the default plan requires it. This adds those settings (and fixes a nil ref panic if you don't have those settings). I think maybe the rest of us just created the `deployer-config-ocp.yml` manually?

I also wonder if we want to mark `OCP_PULL_SECRET` as optional (similar to how I set the Vault vars in this PR). Right now it is required even though I just have it set to empty and it works. Let me know what you think and I can update this PR to either change it to optional or update the readme to indicate it's required.

It's also possible I'm totally misunderstanding how this should be working.

### DIFF
--- a/hack/deployer/README.md
+++ b/hack/deployer/README.md
@@ -28,6 +28,15 @@ Deployer is the provisioning tool that aims to be the interface to multiple Kube
     make switch-aks bootstrap-cloud
     ```
 
+* OpenShift
+
+  * Set `VAULT_ADDR` and `GITHUB_TOKEN` appropriately
+  * Run from the [project root](/):
+
+    ```bash
+    make switch-ocp bootstrap-cloud
+    ```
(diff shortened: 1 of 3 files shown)


sebgl (approved):
LGTM

sebgl:
> I also wonder if we want to mark OCP_PULL_SECRET as optional (similar to how I set the Vault vars in this PR). Right now it is required even though I just have it set to empty and it works

I thought it was mandatory to correctly bootstrap the Openshift cluster. Are you able to deploy a cluster without it?

charith-elastic:
I'll close this because it has been inactive for a long time.

PR #3651, opened 2020-08-21
Support user provided encrypted keys

### LINKED ISSUE #3650: Add support for encrypted private keys
ECK attempts to validate the certificates provided by the user, including the private key:

https://github.com/elastic/cloud-on-k8s/blob/185b1680523226cc2f2e1d94b64a87b9a03ba200/pkg/controller/common/certificates/pem.go#L66-L72

But the [private key of the certificate might be encrypted](https://www.elastic.co/guide/en/kibana/current/configuring-tls.html#configuring-tls-browser-kib), in that case ECK should not attempt to parse it.

We should add a `case` statement to detect this situation and not attempt to decode the key. [`IsEncryptedPEMBlock`](https://golang.org/pkg/crypto/x509/#IsEncryptedPEMBlock) might be a good candidate to detect such a situation, with something along those lines for instance:

```golang
	switch {
	case x509.IsEncryptedPEMBlock(block):
		// Private key is encrypted, do not attempt to parse it
		return nil, nil
	case block.Type == "PRIVATE KEY":
		return parsePKCS8PrivateKey(block.Bytes)
	case block.Type == "RSA PRIVATE KEY" && len(block.Headers) == 0:
		return x509.ParsePKCS1PrivateKey(block.Bytes)
	default:
		return nil, errors.New("expected PEM block to contain an RSA private key")
	}
```

barkbay:
Fixes #3650 

I chose to return an error instead of `nil` to make it explicit that the key is encrypted and let the caller decides what is the best to manage this situation.


pkg/controller/common/certificates/secret.go
```diff
@@ -90,7 +90,7 @@ func (s CertificatesSecret) Validate() error {
 		return pkgerrors.Errorf("can't find private key %s in %s/%s", KeyFileName, s.Namespace, s.Name)
 	}
 	_, err := ParsePEMPrivateKey(key)
-	if err != nil {
+	if err != nil && err != ErrEncryptedPrivateKey {
```
charith-elastic, line 93:
```suggestion
	if err != nil && !errors.Is(err, ErrEncryptedPrivateKey) {
```

charith-elastic (approved):
LGTM

Issue #2830, opened 2020-04-08
Use Pod DNS names instead of IP addresses for discovery and TLS certificates


sebgl:
Related to https://github.com/elastic/cloud-on-k8s/issues/2823.

We configure nodes to advertise and discover themselves through their IP addresses.
For that reason, we are likely to regenerate TLS certificates when a Pod is restarted since its IP address changed.

It does not help with [this bug](https://github.com/elastic/cloud-on-k8s/issues/2823) where Elasticsearch may serve a certificate with the wrong IP address associated.

Things would be simpler if we only manipulated Pod DNS names instead of IP addresses, and make more sense for TLS certificates which are usually bound to a DNS name, not an IP address.

Questions (to investigate):
- can DNS caching be an issue for nodes to discover themselves?
- are there cases (security constraints) where Pods are not allowed/not supposed to rely on DNS?
- any issue with the DNS server will lead to Elasticsearch not behaving correctly: are we ok with that?
- what's the correct DNS name to use? `<pod-name>`, `<pod-name>.<namespace>`

anyasabo:
I started poking at this but won't be picking this up any time soon at least, leaving notes here in case it's helpful:
https://github.com/elastic/cloud-on-k8s/compare/master...anyasabo:pubhost

One thing I think we will need to account for is that sset pods are given a name [`pod-name.sset-service-name`](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#stable-network-id), and we use separate services for each sset (rather than one single headless service that all of the ssets use). I think this is fine, and makes it easier for people who want to target only a specific set of nodes (e.g. ingress), but we will need to make some plumbing changes on our side so that the service name is available.

The other thing I noticed is that when ES starts up, it verifies it can resolve the name you tell it to publish. Currently our services do not publish not-ready pods, so the lookup fails and ES fails to start. We can change this on the service though and I do not see a downside to doing so. If people want different behavior they can create their own service.

pebrc:
Just to update the discussion with the current state: When we moved to DNS based `publish_host` for IPv6 support we ran into issues with a stale DNS entry that causes Elasticsearch nodes to startup with on old IP for their Pod which is not routable and subsequently that condition permanently prevents that node from joining the Elasticsearch cluster see https://github.com/elastic/cloud-on-k8s/issues/3723#issuecomment-690249624

[...]

Issue #2833, opened 2020-04-08
Switch transport TLS certs verification to full


sebgl:
Related to https://github.com/elastic/cloud-on-k8s/issues/2823.

We experimented "full" TLS verification in https://github.com/elastic/cloud-on-k8s/pull/2659, and switched it back to "certificate" in https://github.com/elastic/cloud-on-k8s/pull/2831.

A [bug in Elasticsearch](https://github.com/elastic/elasticsearch/issues/54867) leads to Elasticsearch presenting an outdated certificate with the wrong IP address during and after rolling upgrades.
As a result, full TLS verification leads to nodes not being able to trust each other when the bug happens.

We eventually want to reintroduce full verification, but we need to make sure we correctly handle the bug above. One option is to rely on [DNS names instead of IP addresses](https://github.com/elastic/cloud-on-k8s/issues/2830), another one is to [check the IP address in certificates before starting Elasticsearch](https://github.com/elastic/cloud-on-k8s/issues/2832).

Once we reintroduce full verification, we must make sure we don't break existing clusters impacted by the bug, even though we have a workaround for that bug. What may happen (that we want to avoid) is the following:
- the change to full TLS verifications (and the TLS cert bugfix) leads to a rolling upgrade of the cluster
- first Pod starts with full verification enabled: it cannot contact other nodes in the cluster impacted by the TLS cert bug
- rolling upgrade can never progress

We may want to perform a double rolling upgrade instead, where we first upgrade all nodes to include the TLS certs bugfix, then upgrade all nodes to enable full TLS verification:
* a cluster was created with ECK 1.0, it may or may not include Pods impacted by the bug, serving a certificate with the wrong IP
* ECK version upgrade including the init container fix
* the cluster gets a rolling upgrade, once done we are fairly confident no Pod is impacted by the TLS certs bug. At this point TLS verification is still set to "partial".
* the cluster gets a second rolling upgrade to set TLS verification to "full", because we know no Pod is impacted by the TLS certs bug.

PR #3654, opened 2020-08-24
Make ECK IPv6 compatible

### LINKED ISSUE #3649: Make ECK IPv6 compatible
IPv6 addresses [need to be bracketed](https://www.elastic.co/guide/en/elasticsearch/reference/7.x/modules-discovery-hosts-providers.html#file-based-hosts-provider) when used in resource identifiers. I believe this affects mostly ECK's rendering of configuration files where we use IP addresses for seed hosts and are currently not using brackets for IPv6.

```
 ["java.lang.IllegalArgumentException: IPv6 addresses must be bracketed: fd00:10:244:0:3::9:9300"
```

_Originally posted by @pebrc in https://github.com/elastic/cloud-on-k8s/issues/3408#issuecomment-678102380_

pebrc:
Fixes #3649 

## Summary of changes

* put literal IPv6 addresses in brackets when used in URLs or literal form (e.g in readiness probes)
* switch to correct variant for well known IPs like loopback or zero-valued IP 
* allow IPv6 in SAN validation and CSR
* switch from IP base `network.publish_host` to DNS based:
   * Reason: we used to inject the Pod IP via the downward API, this API does not correctly bracket IPv6 addresses leading to invalid configuration 
   * we were discussing to [switch to DNS based](https://github.com/elastic/cloud-on-k8s/issues/2830) `publish_host` anyway
   * sticking to IPs would require us to change the `publish_host` once the Pod gets its IP and force another delete/create cycle for every node,  which seems worse than just using a DNS name to begin with
* Kibana and Enterprise Search are sensitive about the correct specification of the 'unspecified' ie. all-zero IP address  (`INADDR_ANY`) to bind to all addresses on the local machine
   * this PR introduces some form of IPv6 autodetection on operator startup by injecting the operator IP into the environment
   * this information is then passed on into the configuration generation code for both applications to insert the correct IP
   * we could amend this mechanism in the future by for example introducing an `IPFamily` attribute on the corresponding CRDs to allow users to express a preference in dual-stack environments or add a global flag on the operator level to do the same. 
* https://github.com/elastic/cloud-on-k8s/pull/3654/commits/59c217f2869b6d97e2d850739154668653222d0e changes the discovery mechanism from IPs to DNS names. This change is not strictly necessary to support IPv6 I only changed it for consistency (see also Anya's comment https://github.com/elastic/cloud-on-k8s/pull/3654#issuecomment-680928987) It might carry some risk for slow cluster formation due to cached negative DNS lookups of not (yet) existing Pods and **I am more than happy to revert it**. We discussed out-of-band and the only potential benefit is that we might be able to remove the IP addresses from the certificates we generate and reduce the number of times we have to generate the certificates. At the same time this benefit is probably limited as certificates are one of the things Elasticsearch can reload on the fly.  @nkvoll what are your thoughts in favour or against DNS based discovery?

## Implications of switch to DNS names

@sebgl 's questions from https://github.com/elastic/cloud-on-k8s/issues/2830 still apply: 

>  can DNS caching be an issue for nodes to discover themselves?

yes, negative lookup caching can be an issue: if a Pod was just created we might start seeing a delay until its name can be resolved if some entity did a negative lookup just before it was created

>  are there cases (security constraints) where Pods are not allowed/not supposed to rely on DNS?

Idk

> any issue with the DNS server will lead to Elasticsearch not behaving correctly: are we ok with that?

I would assume that any issues with k8s DNS server will have more widespread effects than just Elasticsearch not behaving correctly, so I would say yes

>  what's the correct DNS name to use? <pod-name>, <pod-name>.<namespace>

I believe it is `<pod-name>.<statefulset-name>` based on [this](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#stable-network-id)

## Breaking?

Should we mark this as breaking because we now include non-ready Pods in the headless service. I am on the fence because I believe the headless service should be considered an implementation detail and it is not meant for "public consumption". Having said that I am sure that there are users already relying on it, so at the very least we should highlight it in the release notes.

## Testing

You need to run this against kind under Linux to test the IPv6 part (or set up your own vanilla IPv6 k8s from scratch). 
I used the following spec: 
```yaml
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
networking:
  ipFamily: ipv6
nodes:
- role: control-plane
- role: worker
- role: worker
- role: worker
```
after reconfiguring Docker to use IPv6
`cat /etc/docker/daemon.json`
```json
{
  "ipv6": true,
  "fixed-cidr-v6": "2001:db8:1::/64"
}
```

I am thinking we should introduce another test job to run the e2e tests against a IPv6 kind cluster (or change the existing job to IPv6) Will be a separate PR though


charith-elastic:
> I believe it is `<pod-name>.<statefulset-name>` based on [this](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#stable-network-id)

I believe it's `<pod-name>.<headless-svc-name>.svc`. That particular bit of documentation is not written well.

pkg/controller/elasticsearch/nodespec/statefulset.go
```diff
@@ -54,6 +54,8 @@ func HeadlessService(es *esv1.Elasticsearch, ssetName string) corev1.Service {
 					Port:     network.HTTPPort,
 				},
 			},
+			// allow nodes to discover themselves via DNS while they are booting up ie. are not ready yet
+			PublishNotReadyAddresses: true,
```
anyasabo, line 58: I think we will want to flag this in the release notes since it could cause issues if people were relying on the old behavior

pkg/controller/elasticsearch/settings/masters_test.go
```diff
@@ -129,6 +129,20 @@ func TestUpdateSeedHostsConfigMap(t *testing.T) {
 			wantErr:         false,
 			expectedContent: "10.0.3.3:9300\n10.0.6.5:9300\n10.0.9.2:9300",
 		},
+		{
+			name: "Can handle IPv6 addresses",
+			args: args{
+				pods: []corev1.Pod{ //
+					newPodWithIP("master2", "fd00:10:244:0:2::3", true),
+					newPodWithIP("master3", "fd00:10:244:0:2::5", true),
+					newPodWithIP("master1", "fd00:10:244:0:2::2", true),
```
anyasabo, line 138: nit: maybe don't use shorthand for one of them to exercise that path too? I wouldn't expect issues based on our implementation though

anyasabo:
This approach LGTM 👍 

>Question: Should we use DNS names in discovery as well to be consistent?
I think yes? But I haven't thought about it too much.

This PR also allows us to pick up https://github.com/elastic/cloud-on-k8s/issues/2833

anyasabo:
I was curious what the potential effects of negative caching might have and it became a bit complicated. CoreDNS [by default](https://coredns.io/plugins/kubernetes/) has a 5s TTL. BUT the [default config](https://kubernetes.io/docs/tasks/administer-cluster/dns-custom-nameservers/) shipped with k8s has a 30s TTL (to match the dnsmasq based kubedns).

If you run coredns in [cache mode](https://coredns.io/plugins/cache/) (as you would for [node local DNS](https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/dns/nodelocaldns/nodelocaldns.yaml)), the default is 30 minutes for nxdomain responses. But in [GKE](https://cloud.google.com/kubernetes-engine/docs/how-to/nodelocal-dns-cache#details), they configure node local DNS to have a 5s TTL for nxdomain responses.

Openshift (v4 at least) also runs with a [30s TTL](https://rcarrata.com/openshift/dns-deep-dive-in-openshift/).

It appears like most configs would have a 5-30s TTL, which seems fine to me.

pebrc:
Elasticsearch/the JVM caches negative lookups for 10 seconds and positive ones for 60 in recent versions https://www.elastic.co/guide/en/elasticsearch/reference/current/networkaddress-cache-ttl.html

pkg/utils/net/ip_test.go
```diff
@@ -0,0 +1,209 @@
...
+				ipStr: "0.0.0.0",
+			},
+			want: corev1.IPv4Protocol,
+		},
+		{
+			name: "IPv4-mapped IPv6",
+			args: args{
+				ipStr: "::FFFF:129.144.52.38",
+			},
+			want: corev1.IPv4Protocol,
```
barkbay, line 192: nit: I must confess I spent some time trying to understand the differences between ` IPv4-embedded IPv6` and `IPv4-mapped IPv6`, and wondering if this is what I would expect from `ToIPFamily` since `::FFFF:129.144.52.38` is an ipv6 address.

But I think I'm 👍 with the actual code.

pkg/utils/net/ip.go
```diff
@@ -6,12 +6,45 @@ package net
...
+	"strconv"
+
+	corev1 "k8s.io/api/core/v1"
 )
 
-// MaybeIPTo4 attempts to convert the provided net.IP to a 4-byte representation if possible, otherwise does nothing.
-func MaybeIPTo4(ipAddress net.IP) net.IP {
-	if ip := ipAddress.To4(); ip != nil {
+// IPToRFCForm normalizes the IP address given to fit the expected network byte order octet form described in
+// https://tools.ietf.org/html/rfc5280#section-4.2.1.6
```
barkbay, line 15: IIUC in addition to the byte ordering requirement we also want to avoid compressed zero form of IPv6 addresses using "::" ?
Maybe we can also explain this in the comment ?

barkbay (commented):
I'm still doing some tests. I just left a minor comment but feel free to ignore it as I'm not an IPv6 expert.

pkg/utils/net/ip_test.go
pebrc: Yes I found that confusing too, but I am following `net.ParseIP` 's behaviour here.

pkg/utils/net/ip.go
pebrc: Well it is not removing the compressed form actually. Which is fine , with the exception of Enterprise Search which has currently an issue with compressed IPv6 IPs. I missed this because I was testing with manually configured overrides 😞 So thanks for pointing this out, will fix in another commit.

nkvoll:
I think DNS based discovery makes sense. ES wants to discover a node with a name (which also usually has a PVC attached). The Pod is ephemeral and doesn't matter, and thus the same ephemeralness applies to its IP. Having bad TTLs for positive and negative hits seem like an operator error that also would cause issues for other services.

I would vote for creating another specific issue/PR for the switch to DNS-based discovery so it gets a little more (deserved) visibility.

cmd/manager/main.go
```diff
@@ -440,6 +441,7 @@ func startOperator(stopChan <-chan struct{}) error {
 	}
 	params := operator.Parameters{
 		Dialer:            dialer,
+		IPFamily:          net.ToIPFamily(os.Getenv(settings.EnvPodIP)),
```
sebgl, line 444: I'm wondering if we should include that as an operator config setting, that defaults to autodetect if empty.
eg. `--ip-family=ipv4|ipv6|""`
So users can override it if for some reasons the IP family of the operator is not the ip family they want to use for the managed workload.

pkg/controller/enterprisesearch/config.go
```diff
@@ -227,10 +229,18 @@ func parseConfigRef(driver driver.Interface, ent entv1beta1.EnterpriseSearch) (*
 	return common.ParseConfigRef(driver, &ent, ent.Spec.ConfigRef, ConfigFilename)
 }
 
-func defaultConfig(ent entv1beta1.EnterpriseSearch) *settings.CanonicalConfig {
+func inAddrAnyFor(ipFamily corev1.IPFamily) string {
+	if ipFamily == corev1.IPv4Protocol {
+		return net.IPv4zero.String()
+	}
+	// Enterprise Search even in its most recent version 7.9.0 cannot properly handle contracted IPv6 addresses like "::"
```
sebgl, line 236: sounds like we should open an issue in the enterprise search repo

pebrc: Already done.

cmd/manager/main.go
pebrc: I thought this would be a future addition. But I think you are right it would be good to have an escape hatch right from the start. Will add it.

pkg/controller/kibana/config_settings.go
```diff
@@ -8,6 +8,8 @@ import (
 	"context"
 	"path"
 
+	"github.com/elastic/cloud-on-k8s/pkg/utils/net"
```
barkbay, line 11: nit: some files have new lines in imports, [not sure if we still care about that](https://github.com/elastic/cloud-on-k8s/blob/master/CONTRIBUTING.md#format-your-code-and-manage-imports).

barkbay (approved):
I did some tests, deployed `Elasticsearch`,`Kibana`,`APM Server`, `Enterprise Search` and `filebeats` on  Kind as suggested, it worked as expected. 

Also  👍 to add a dedicated e2e test pipeline

Regarding questions around DNS caching I'm not sure I have a good answer. It might happen at the infra level, I guess understanding how it would disrupt Pods being restarted requires to understand how `publish_host` is used by the stack and the clients ?
Should we do some tests on an infra which makes an aggressive use of DNS caching to understand the edge cases ?

image

PR #3660, opened 2020-08-25
Set 1.15 as the default version on GKE

barkbay:
GKE 1.14 seems to be deprecated, update GKE versions to 1.15 and 1.16.


sebgl (approved):
LGTM

Issue #3659, opened 2020-08-25
TestEnterpriseTrialLicense failing 


pebrc:
Interesting test failure in https://devops-ci.elastic.co/job/cloud-on-k8s-e2e-tests-stack-versions/185/

Symptom:
```
[2020-08-25T02:59:04.665Z] === FAIL: test/e2e/es TestEnterpriseTrialLicense/Elasticsearch_license_should_be_[trial] (300.00s)
[2020-08-25T02:59:04.666Z]         	Test:       	TestEnterpriseTrialLicense/Elasticsearch_license_should_be_[trial]
```

Looking a the logs: 

1. trial controller starts trial as normal

```
{"log.level":"info","@timestamp":"2020-08-25T02:50:53.500Z","log.logger":"trial-controller","message":"Start reconcile iteration","service.version":"1.3.0-SNAPSHOT+771e9b16","service.type":"eck","ecs.version":"1.4.0","iteration":1,"namespace":"e2e-wnqbg-mercury","secret_name":"eck-trial"}
{"log.level":"info","@timestamp":"2020-08-25T02:50:54.096Z","log.logger":"license","message":"Starting enterprise trial","service.version":"1.3.0-SNAPSHOT+771e9b16","service.type":"eck","ecs.version":"1.4.0","start":"2020-08-25T02:50:54.096Z","end":"2020-09-24T02:50:54.096Z"}
{"log.level":"info","@timestamp":"2020-08-25T02:50:54.156Z","log.logger":"trial-controller","message":"End reconcile iteration","service.version":"1.3.0-SNAPSHOT+771e9b16","service.type":"eck","ecs.version":"1.4.0","iteration":1,"took":0.656566376,"namespace":"e2e-wnqbg-mercury","secret_name":"eck-trial"}
```

2. license check fails to find `trial-status`. I am not surprised, we expect some delay until the client cache is up to date

```
{"log.level":"error","@timestamp":"2020-08-25T02:50:54.157Z","log.logger":"license","message":"while checking license validity","service.version":"1.3.0-SNAPSHOT+771e9b16","service.type":"eck","ecs.version":"1.4.0","error":"while loading signature secret: Secret \"trial-status\" not found","errorVerbose":"Secret \"trial-status\" not found\nwhile loading signature secret\ngithub.com/elastic/cloud-on-k8s/pkg/controller/com
[...]

PR #3662, opened 2020-08-25
Do not reset trial state once a license has been issued.

pebrc:
Fixes https://github.com/elastic/cloud-on-k8s/issues/3659

Addresses an issue observed when the trial status secret would not be found due to an outdated client cache and the trial controller would re-initialise the trial state from scratch invalidating an already issued license.

This PR checks therefore for `trialState.IsEmpty` instead of `trialState.IsTrialStarted`.

### DIFF
--- a/pkg/controller/license/trial/trial_controller.go
+++ b/pkg/controller/license/trial/trial_controller.go
@@ -14,6 +14,7 @@ import (
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common"
 	licensing "github.com/elastic/cloud-on-k8s/pkg/controller/common/license"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common/operator"
+	"github.com/elastic/cloud-on-k8s/pkg/controller/common/reconciler"
 	"github.com/elastic/cloud-on-k8s/pkg/utils/k8s"
 	pkgerrors "github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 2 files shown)


pkg/controller/license/trial/trial_controller.go
```diff
@@ -179,11 +179,16 @@ func (r *ReconcileTrials) startTrialActivation() error {
...
 func (r *ReconcileTrials) completeTrialActivation(license types.NamespacedName) (reconcile.Result, error) {
 	if r.trialState.CompleteTrialActivation() {
-		status, err := licensing.ExpectedTrialStatus(r.operatorNamespace, license, r.trialState)
+		expectedStatus, err := licensing.ExpectedTrialStatus(r.operatorNamespace, license, r.trialState)
 		if err != nil {
 			return reconcile.Result{}, err
 		}
-		return reconcile.Result{}, r.Update(&status)
+		var existingStatus corev1.Secret
+		if err := r.Get(k8s.ExtractNamespacedName(&expectedStatus), &existingStatus); err != nil {
```
pebrc, line 187: This is mainly for the overeager fake client which requires a resource version on every update. The fact that we are hitting this issue only now tells me that none of the other tests executed this code path 🤷

sebgl: How about calling `ReconcileSecret()` instead? Manipulating the resource version here feels a bit weird.

sebgl (commented):
LGTM, I left one comment worth addressing imo

sebgl (approved):
LGTM

PR #3673, opened 2020-08-28
Set default stack version to 7.9.0 in the Makefile

sebgl:
It helps running E2E tests localy with the latest version.


Makefile
```diff
@@ -369,7 +369,7 @@ switch-registry-dev: # just use the default values of variables
 E2E_REGISTRY_NAMESPACE ?= eck-dev
 E2E_IMG                ?= $(REGISTRY)/$(E2E_REGISTRY_NAMESPACE)/eck-e2e-tests:$(TAG)
 TESTS_MATCH            ?= "^Test" # can be overriden to eg. TESTS_MATCH=TestMutationMoreNodes to match a single test
-STACK_VERSION          ?= 7.8.1
+STACK_VERSION          ?= 7.9.0
```
barkbay, line 372: Is it only used for e2e tests ? I was wondering if it would not make sense to rename it to something more explicit ? (e.g. `E2E_STACK_VERSION`)

barkbay (approved):
LGTM

Makefile
sebgl: Good idea!

Issue #3656, opened 2020-08-25
Enterprise Search 7.9.x does not run on Openshift


barkbay:
[TestEnterpriseSearchVersionUpgradeToLatest7x](https://github.com/elastic/cloud-on-k8s/blob/1a9bf120a21c79077322831dc0f3ce76feba4864/test/e2e/ent/ent_test.go#L49) is failing [on](https://devops-ci.elastic.co/job/cloud-on-k8s-e2e-tests-ocp/213/testReport/) [Openshift](https://devops-ci.elastic.co/job/cloud-on-k8s-e2e-tests-ocp/214//testReport).

This test upgrades EnterpriseSearch from 7.8.1 to 7.9.0. But deploying a brand new 7.9.0 Enterprise Search resource (i.e. without upgrading) seems to fail the same way, so I don't think it is related to an upgrade from 7.8.1 to 7.9.0

```
NAME                                                                       HEALTH   NODES   VERSION   PHASE   AGE
elasticsearch.elasticsearch.k8s.elastic.co/test-ent-version-upgrade-584t   green    1       7.9.0     Ready   15m

NAME                                                                             HEALTH   NODES   VERSION   AGE
enterprisesearch.enterprisesearch.k8s.elastic.co/test-ent-version-upgrade-w82m   green    2       7.9.0     15m

NAME                                                     READY   STATUS             RESTARTS   AGE   IP            NODE                                                 NOMINATED NODE   READINESS GATES
pod/test-ent-version-upgrade-584t-es-masterdata-0        1/1     Running            0          15m   10.129.2.15   local-lv4nr-w-b-zg7mh.c.elastic-cloud-dev.internal   <none>           <none>
pod/test-ent-version-upgrade-w82m-ent-6fccc7bc98-55s4k   1/1     Running            0          15m   10.128.2.18   local-lv4nr-w-d-4cgwj.c.elastic-cloud-dev.internal   <none>           <none>
pod/test-ent-version-upgrade-w82m-ent-6fccc7bc98-6m5zs   1/1     Running            0          15m   10.129.2.14   local-lv4nr-w-b-zg7mh.c.elastic-cloud-dev.internal   <none>           <none>
pod/test-ent-version-upgrade-w82m-ent-bb56cdbd6-x5p9z    0/1     CrashLoopBackOff   6          12m   10.131.0.21   local-lv4nr-w-c-4cdmr.c.elastic-cloud-dev.internal   <none>           <none>
```

```
[...]

PR #3675, opened 2020-08-28
Skip Enterprise Search 7.9.0 E2E tests on OCP

sebgl:
Enterprise Search 7.9.0 fails to start on Openshift due to
file permission issues. See https://github.com/elastic/cloud-on-k8s/issues/3656.

Let's disable this test combination in our ocp e2e tests.

### DIFF
--- a/test/e2e/ent/ent_test.go
+++ b/test/e2e/ent/ent_test.go
@@ -63,10 +63,6 @@ func TestEnterpriseSearchVersionUpgradeToLatest7x(t *testing.T) {
 		WithVersion(srcVersion).
 		WithRestrictedSecurityContext()
 
-	if ent.SkipTest() {
-		t.SkipNow() // invalid version
-	}
-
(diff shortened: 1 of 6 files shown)


barkbay (changes requested):
Only the source version is validated, not the target one. I think this change requires something along those lines in the tests:

```golang
	entUpgraded := ent.WithVersion(dstVersion).WithMutatedFrom(&ent)

	if entUpgraded.SkipTest() {
		t.SkipNow() // invalid target version
	}
```

sebgl:
@barkbay great catch!
I extended the way we check for incompatible builders to skip in `test.Sequence` to other testing methods (`test. RunMutations `, etc.), checking both initial + mutation builders; so we keep this mechanism sort of generic with exceptions only in the builder `SkipTest()` method.

barkbay (approved):
LGTM

PR #3669, opened 2020-08-27
Add additional k8s deps to renovate

anyasabo:
We added some more k8s deps. This adds them to the renovate k8s group so a) the PRs are only for patch versions b) the k8s PRs are grouped together

PR #3678, opened 2020-08-31
Bump hack tool deps to k8s 1.19

barkbay:
Supersedes #3666 #3667 #3668 

Should be replaced by a single PR from renovate when #3669 is merged.


sebgl (approved):
LGTM

Issue #3676, opened 2020-08-28
Flaky e2e test Add_some_data_to_the_cluster_before_starting_the_mutation


sebgl:
https://devops-ci.elastic.co/job/cloud-on-k8s-e2e-tests-stack-versions/188/testReport/

```
=== RUN   TestMutationSecondMasterSet/Add_some_data_to_the_cluster_before_starting_the_mutation
    steps_mutation.go:80: 
        	Error Trace:	steps_mutation.go:80
        	Error:      	Received unexpected error:
        	            	429 Too Many Requests: [parent] Data too large, data for [<transport_request>] would be [1051180086/1002.4mb], which is larger than the limit of [1011774259/964.9mb], real usage: [1051179048/1002.4mb], new bytes reserved: [1038/1kb], usages [request=0/0b, fielddata=0/0b, in_flight_requests=1038/1kb, accounting=3768/3.6kb]
        	Test:       	TestMutationSecondMasterSet/Add_some_data_to_the_cluster_before_starting_the_mutation
```

barkbay:
It also happened today on [a different test](https://devops-ci.elastic.co/job/cloud-on-k8s-e2e-tests-stack-versions/189/):

```
{"Time":"2020-08-31T02:32:01.137797801Z","Action":"output","Package":"github.com/elastic/cloud-on-k8s/test/e2e/es","Test":"TestMutationRollingDownscaleCombination/Add_some_data_to_the_cluster_before_starting_the_mutation","Output":"=== RUN   TestMutationRollingDownscaleCombination/Add_some_data_to_the_cluster_before_starting_the_mutation\n"}^M
{"Time":"2020-08-31T02:32:02.669811262Z","Action":"output","Package":"github.com/elastic/cloud-on-k8s/test/e2e/es","Test":"TestMutationRollingDownscaleCombination/Add_some_data_to_the_cluster_before_starting_the_mutation","Output":"    steps_mutation.go:80: \n"}^M
{"Time":"2020-08-31T02:32:02.66984715Z","Action":"output","Package":"github.com/elastic/cloud-on-k8s/test/e2e/es","Test":"TestMutationRollingDownscaleCombination/Add_some_data_to_the_cluster_before_starting_the_mutation","Output":"        \tError Trace:\tsteps_mutation.go:80\n"}^M
[...]

PR #3679, opened 2020-08-31
Add a bit more context in data integrity check errors

sebgl:
In order to investigate https://github.com/elastic/cloud-on-k8s/issues/3676,
this commit adds a bit more context to the api errors returned by Elasticsearch
in DataIntegrityCheck.Init().

The goal is to understand if there's a common pattern in these errors (eg. always
returned when the first document is indexed).


barkbay (approved):
LGTM

PR #3711, opened 2020-09-04
Refactor Helm chart

charith-elastic:
This lays the ground work for potentially making the internal Helm chart public.

- Make the chart production ready by allowing common pod customizations such as defining node affinity rules, security contexts, image pull secrets etc.
- Follow the [best practices guide](https://helm.sh/docs/chart_best_practices/).
- Introduce the manifest-generator mode to produce the `all-in-one.yaml` file. (Creates the namespace and removes Helm specific labels from the output.)
- Add tests to validate the generated output for various profiles.

A recent version of Helm 3 (3.3.1 preferred) and a recent build of the operator is required for testing. For example:

```shell
helm install elastic-operator ./hack/manifest-gen/assets/charts/eck -n elastic-system --create-namespace --set=image.repository=docker.elastic.co/eck-dev/eck-operator-cell --set=image.tag=1.3.0-SNAPSHOT-41932662
```


hack/manifest-gen/assets/charts/eck/profile-restricted.yaml
```diff
@@ -1,7 +1,6 @@
-config:
-  createClusterScopedResources: false
+managedNamespaces: ["elastic-system"]
 
-  managedNamespaces: ["elastic-system"]
```
sebgl, line 4: I'm curious about the rational behind moving things that were in `config` at an upper-level. In my mind it feels natural to have all operator config items that can be customized though a flag or config map at the same level?

charith-elastic: It's a tough balance to strike. Managed namespaces are fundamental to how the operator works and the value is used in quite a few places to change how the templates are rendered. Therefore I thought it belonged at the top level.

The problem with moving all operator flags under `config` is that  sometimes just setting the config value is not enough. We need to do other things based on the value of that. Webhook is a good example of this. Just setting `enable-webhook` is not enough because you then have to make decisions about whether to use user-provided certificates, cert-manager, or operator-managed certificates and generate the appropriate manifests that require further customisation. That raises the question of where those additional customisation options should go. If they also go under `config`, we dilute the meaning of it. If we move the flag under `config` and the rest of the customisation somewhere else, the relationship between them won't be obvious to users.

Helm best practices document recommends not deeply nesting the values so the approach I used in this refactoring is to think in terms of major behavioural changes. Setting managed namespaces changes how the operator behaves in a significant way, so it merited being at the top level. The same logic applies to the webhook and so on. Hope that makes sense.

hack/manifest-gen/assets/charts/eck/templates/NOTES.txt
```diff
@@ -0,0 +1,2 @@
+1. Inspect the operator logs by running the following command:
+   kubectl logs -n {{ .Release.Namespace }} $(kubectl get pods --namespace {{ .Release.Namespace }} -l "app.kubernetes.io/name={{ include "eck.name" . }},app.kubernetes.io/instance={{ .Release.Name }}" -o jsonpath="{.items[0].metadata.name}")
```
pebrc, line 2:
```suggestion
   kubectl logs -n {{ .Release.Namespace }} sts/{{ .Release.Name }}
```

Should work too and does make any assumptions about the shell in use, or am I missing something?

hack/manifest-gen/assets/charts/eck/templates/_helpers.tpl
```diff
@@ -1,9 +1,72 @@
...
+{{/*
+Expand the name of the chart.
+*/}}
+{{- define "eck.name" -}}
+{{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" }}
+{{- end }}
+
+{{/*
+Create a default fully qualified app name.
+We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
```
pebrc, line 10: Feels like this comment applies above as well?

hack/manifest-gen/assets/charts/eck/templates/_helpers.tpl
```diff
@@ -1,9 +1,72 @@
...
+Common labels
+*/}}
+{{- define "eck.labels" -}}
+{{ include "eck.selectorLabels" . }}
+{{- if .Chart.AppVersion }}
+app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
+{{- end }}
+{{- if not .Values.internal.manifestGen }}
+helm.sh/chart: {{ include "eck.chart" . }}
+app.kubernetes.io/managed-by: {{ .Release.Service }}
```
pebrc, line 43: Interesting. Is this not always going to be `Helm`? What other services are possible?

hack/manifest-gen/assets/charts/eck/values.yaml
```diff
@@ -1,70 +1,138 @@
...
-  tracing:
-    enabled: false
-    config:
-      ELASTIC_APM_SERVER_URL: http://localhost:8200
-      ELASTIC_APM_SERVER_TIMEOUT: 30s
-
-  ca:
-    validity: 8760h
-    rotateBefore: 24h
+  # maxConcurrentReconciles is the number of concurrent reconciliation operations to perform.
```
pebrc, line 118: per controller

pebrc (commented):
Need to do a bit more testing but looks good to me. 👍 for the flattened value file.

hack/manifest-gen/assets/charts/eck/templates/_helpers.tpl
charith-elastic: It is always `Helm` for Helm charts. I think they just made the constant available from `Release.Service` for consistency so that people don't mess up the spelling or capitalization.

hack/manifest-gen/assets/charts/eck/profile-restricted.yaml
sebgl: Sounds good, thanks for the clarification!

pebrc (approved):
I did a bit of admittedly non-exhaustive testing and so far everything looked good to me. 👍

sebgl (approved):
LGTM

hack/manifest-gen/assets/charts/eck/templates/_helpers.tpl
```diff
@@ -1,9 +1,74 @@
...
+{{- end }}
+
+{{/*
+Create a default fully qualified app name.
+We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
+If release name contains chart name it will be used as a full name.
+*/}}
+{{- define "eck.fullname" -}}
+{{- if .Values.fullnameOverride }}
+{{- .Values.fullnameOverride | trunc 63 | trimSuffix "-" }}
```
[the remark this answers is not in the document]
charith-elastic: Installation would fail in that case:

```
helm install elastic-operator ./hack/manifest-gen/assets/charts/eck -n elastic-system --create-namespace --set=fullnameOverride="my elastic operator" --set=image.repository=docker.elastic.co/eck-dev/eck-operator-cell --set=image.tag=1.3.0-SNAPSHOT-41932662 
Error: ServiceAccount "my elastic operator" is invalid: metadata.name: Invalid value: "my elastic operator": a DNS-1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')
```

I haven't come across Helm charts that do validation of names and such. In the end, everything gets turned into Kubernetes resources anyway -- so if Kubernetes lets something through, it's probably valid.

I did introduce a few checks to catch impossible or non-sensical configurations. They are not exhaustive by any means because I am biased by insider knowledge and have some blind spots. We'll probably have to add more of those checks when we start seeing bug reports from users in the field.

Issue #2832, opened 2020-04-08
Wait until Pod TLS certificate are correct in the prepare-fs init container


sebgl:
Related to https://github.com/elastic/cloud-on-k8s/issues/2823.

[bug in Elasticsearch](https://github.com/elastic/elasticsearch/issues/54867) leads to Elasticsearch presenting an outdated certificate with the wrong IP address during and after rolling upgrades.

To mitigate this, we can improve the existing prepare-fs init container to not only wait for certificates to be present on disk, but also to ensure those certificates have the correct IP address, before we move on with the Elasticsearch container.

`openssl` is not present in the Elasticsearch Docker image so we cannot easily inspect the certificate content. Instead we may need to clearly identify Pods IP addresses in the certs secret content. For example, by adding new files that associate a Pod name to a Pod IP address. In the script, we would wait until the IP address presented in the secret files matches the `$POD_IP` environment variable.

We must make sure this new way of doing things does not break existing clusters which rely on the existing certificate files, named from the pod name.

https://github.com/elastic/cloud-on-k8s/issues/2830 proposes an other alternative where we would rely on Pod DNS names instead of IP addresses, which would make this issue irrelevant.

pebrc:
The underlying issue in Elasticsearch has been addressed in https://github.com/elastic/elasticsearch/pull/54999 I am closing this as well.

PR #3712, opened 2020-09-04
Move discovery to DNS names

### LINKED ISSUE #2830: Use Pod DNS names instead of IP addresses for discovery and TLS certificates

pebrc:
Fixes #2830
Related #3654 

This PR changes Elasticsearch discovery to use DNS hostnames instead of IP addresses. To replicate the discussion on #3654 : we think the potential negative impact of DNS caching is manageable and outweighs the advantage of simplifying discovery.

We could in theory remove the IP address from the ECK-generated transport certificates now and save a few certificate regenerations when a node is scheduled on another Pod. I left it in place because I believe the frequency of these regenerations if fairly low but open to reconsider.

### DIFF
--- a/pkg/controller/elasticsearch/sset/pod.go
+++ b/pkg/controller/elasticsearch/sset/pod.go
@@ -20,6 +20,18 @@ import (
 	"github.com/elastic/cloud-on-k8s/pkg/utils/stringsutil"
 )
 
+// HeadlessServiceName returns the name of the headless service for the given StatefulSet.
+func HeadlessServiceName(ssetName string) string {
+	// just use the sset name
+	return ssetName
+}
+
+//PodDNSName returns the DNS resolvable name of the given pod resolvable within the namespace.
+func PodDNSName(pod corev1.Pod) string {
+	ssetName := pod.Labels[label.StatefulSetNameLabelName]
+	return fmt.Sprintf("%s.%s", pod.Name, HeadlessServiceName(ssetName))
+}
+
 // PodName returns the name of the pod with the given ordinal for this StatefulSet.
 func PodName(ssetName string, ordinal int32) string {
 	return fmt.Sprintf("%s-%d", ssetName, ordinal)
--- a/pkg/controller/elasticsearch/sset/pod.go
+++ b/pkg/controller/elasticsearch/sset/pod.go
@@ -26,7 +26,7 @@ func HeadlessServiceName(ssetName string) string {
 	return ssetName
 }
 
-//PodDNSName returns the DNS resolvable name of the given pod resolvable within the namespace.
+// PodDNSName returns the DNS resolvable name of the given pod resolvable within the namespace.
 func PodDNSName(pod corev1.Pod) string {
 	ssetName := pod.Labels[label.StatefulSetNameLabelName]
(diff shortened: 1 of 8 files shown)


pkg/controller/elasticsearch/sset/pod.go
```diff
@@ -20,6 +20,18 @@ import (
...
+// HeadlessServiceName returns the name of the headless service for the given StatefulSet.
+func HeadlessServiceName(ssetName string) string {
+	// just use the sset name
+	return ssetName
+}
+
+//PodDNSName returns the DNS resolvable name of the given pod resolvable within the namespace.
+func PodDNSName(pod corev1.Pod) string {
+	ssetName := pod.Labels[label.StatefulSetNameLabelName]
+	return fmt.Sprintf("%s.%s", pod.Name, HeadlessServiceName(ssetName))
```
sebgl, line 32: I'm wondering if we should consider how a user may override the Pod hostname and subdomain fields: https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-s-hostname-and-subdomain-fields
But I think that's really a corner case that is probably safe to ignore for now. It probably has other impacts.

sebgl (approved):
LGTM

Side note: using DNS names only in the certificates (https://github.com/elastic/cloud-on-k8s/issues/2830) would help fixing https://github.com/elastic/cloud-on-k8s/issues/2832 and https://github.com/elastic/cloud-on-k8s/issues/2833.

pkg/controller/elasticsearch/sset/pod.go
```diff
@@ -20,6 +20,18 @@ import (
 	"github.com/elastic/cloud-on-k8s/pkg/utils/stringsutil"
 )
 
+// HeadlessServiceName returns the name of the headless service for the given StatefulSet.
+func HeadlessServiceName(ssetName string) string {
+	// just use the sset name
+	return ssetName
+}
+
+//PodDNSName returns the DNS resolvable name of the given pod resolvable within the namespace.
```
thbkrkr, line 29:
```suggestion
// PodDNSName returns the DNS resolvable name of the given pod resolvable within the namespace.
```

pebrc:
Closing in the light of https://github.com/elastic/cloud-on-k8s/issues/3723#issuecomment-690249624

PR #3714, opened 2020-09-05
Add IPv6 e2e test stage

pebrc:
Adds
*  an additional test stage to the kind e2e test suite for IPv6. 
*  a IP family parameter to the kind script

Tested with a local Jenkins setup and it seems the `eck` workers have Docker already set up to run IPv6, so this should work without further changes 🤞


thbkrkr (approved):
LGTM

PR #3717, opened 2020-09-07
Add E2E tests for Elastic Stack 8.0.0-SNAPSHOT

### LINKED ISSUE #2608: Add E2E tests for Elastic Stack 8.0.0-SNAPSHOT

thbkrkr:
Resolved #2608.

### DIFF
--- a/.ci/jobs/e2e-tests-snapshot-versions-8x-gke.yml
+++ b/.ci/jobs/e2e-tests-snapshot-versions-8x-gke.yml
@@ -8,7 +8,7 @@
     parameters:
       - bool:
           name: SEND_NOTIFICATIONS
-          default: true
+          default: false
           description: "Specified if job should send notifications to Slack. Enabled by default."
     pipeline-scm:
       scm:
--- a/.ci/jobs/e2e-tests-snapshot-versions-8x-gke.yml
+++ b/.ci/jobs/e2e-tests-snapshot-versions-8x-gke.yml
@@ -1,0 +1,21 @@
+---
+- job:
+    description: Run ECK E2E tests on GKE every night to verify ECK against Elastic Stack dev 8.x versions
+    name: cloud-on-k8s-e2e-tests-snapshot-versions
+    project-type: pipeline
+    triggers:
+      - timed: '0 0 * * 1-5'
+    parameters:
+      - bool:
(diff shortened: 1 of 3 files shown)


thbkrkr:
I saw instabilities on EnterpriseSearch and License tests but it's a bit painful to repro so I don't know if it's flakyness or a real bug.

Proposal: create a separate job only for the 8.0.0-SNAPSHOT with notifications turned off in order not to add noise until it seems stable enough.

thbkrkr:
The last commit creates a new pipeline, duplicated from `e2e-tests-snapshot-versions-gke`, to run E2E tests with Elastic Stack 8.0.0-SNAPSHOT with notifications disabled by default.

thbkrkr:
jenkins test this please

.ci/jobs/e2e-tests-snapshot-versions-8x-gke.yml
```diff
@@ -0,0 +1,21 @@
+---
+- job:
+    description: Run ECK E2E tests on GKE every night to verify ECK against Elastic Stack dev 8.x versions
+    name: cloud-on-k8s-e2e-tests-snapshot-versions
```
barkbay, line 4: Should we change the name here ?

Also I'm wondering if it makes sense to have only two jobs and avoid to set the version in the job/file name:

* One for stable versions of the stack : `cloud-on-k8s-e2e-tests-stable-versions` , which would test 6.8 to 7.9 in parallel 
* One for dev/snapshot versions: `cloud-on-k8s-e2e-tests-snapshot-versions`, which would test 7.10 and 8 in parallel 

wdyt ?

thbkrkr: Yes, for the missing name change.

Interesting idea for the two jobs but I'm afraid I won't have time to delve into this now.

barkbay (approved):
LGTM !

thbkrkr:
Creates a new pipeline, duplicated from e2e-tests-snapshot-versions-gke,
to run E2E tests with Elastic Stack 8.0.0-SNAPSHOT with notifications disabled by default.

Issue #3723, opened 2020-09-07
TestMutationHTTPToHTTPS is flaky 


pebrc:
https://devops-ci.elastic.co/job/cloud-on-k8s-e2e-tests-master/575/consoleFull

```
=== RUN   TestMutationHTTPSToHTTP/All_expected_Pods_should_eventually_be_ready#01
Retries (30m0s timeout): ...................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................
    utils.go:84: 
        	Error Trace:	utils.go:84
        	Error:      	Received unexpected error:
        	            	pod test-mutation-https-to-http-vslx-es-masterdata-0 was not upgraded (yet?) to match the expected specification
        	Test:       	TestMutationHTTPSToHTTP/All_expected_Pods_should_eventually_be_ready#01
{"log.level":"error","@timestamp":"2020-09-07T12:03:12.939Z","message":"stopping early","service.version":"0.0.0-SNAPSHOT+00000000","service.type":"eck","ecs.version":"1.4.0","error":"test failure","error.stack_trace":"github.com/elastic/cloud-on-k8s/test/e2e/test.StepList.RunSequential\n\t/go/src/github.com/elastic/cloud-on-k8s/test/e2e/test/step.go:43\ngithub.com/elastic/cloud-on-k8s/test/e2e/test.RunMutations\n\t/go/src/github.com/elastic/cloud-on-k8s/test/e2e/test/run_mutation.go:38\ngithub.com/elastic/cloud-on-k8s/test/e2e/test.RunMutation\n\t/go/src/github.com/elastic/cloud-on-k8s/test/e2e/test/run_mutation.go:80\ngithub.com/elastic/cloud-on-k8s/test/e2e/es.RunESMutation\n\t/go/src/github.com/elastic/cloud-on-k8s/test/e2e/es/mutation_test.go:299\ngithub.com/elastic/cloud-on-k8s/test/e2e/es.TestMutationHTTPSToHTTP\n\t/go/src/github.com/elastic/cloud-on-k8s/test/e2e/es/mutation_test.go:48\ntesting.tRunner\n\t/usr/local/go/src/testing/testing.go:1108"}
    --- FAIL: TestMutationHTTPSToHTTP/All_expected_Pods_should_eventually_be_ready#01 (1800.00s)

```

Looks like we have the first case of DNS related issues at hand, after changing to DNS based `publish_host`. 

```
[...]

PR #3726, opened 2020-09-08
Revert network.publish_host back to IP addresses

pebrc:
Fixes https://github.com/elastic/cloud-on-k8s/issues/3723

Using DNS names turned out to be problematic due to DNS resolving to previous Pod IPs,
which in combination with a bug in Elasticsearch (https://github.com/elastic/elasticsearch/issues/49795)
lead to nodes not being able to re-join the cluster on updates.

While the [proposal](https://github.com/elastic/cloud-on-k8s/issues/3723#issuecomment-688695436) to use `<pod>.<service>.svc.<cluster-domain>` seems tempting and I am happy to look into it if we think sticking with a DNS name is strictly better.

This PR goes back to an IP based `publish_host` which is bracketed as needed in the case of IPv6, exploiting the fact that we now have a notion of an IP family in the operator parameters.

### DIFF
--- a/pkg/utils/net/ip.go
+++ b/pkg/utils/net/ip.go
@@ -59,6 +59,6 @@ func IPLiteralFor(ipOrPlaceholder string, ipFamily corev1.IPFamily) string {
 		// IPv6: return a bracketed version of the IP
 		return fmt.Sprintf("[%s]", ipOrPlaceholder)
 	}
-	// IPv4: leave the place holder as is
+	// IPv4: leave the placeholder as is
 	return ipOrPlaceholder
 }
--- a/pkg/utils/net/ip.go
+++ b/pkg/utils/net/ip.go
@@ -55,10 +55,10 @@ func ToIPFamily(ipStr string) corev1.IPFamily {
 // The difference to net.JoinHostPort is that it also allows IP to be a placeholder that will be resolved
 // to the actual IP at a later time.
 func IPLiteralFor(ipOrPlaceholder string, ipFamily corev1.IPFamily) string {
-	var prefix, suffix string
 	if ipFamily == corev1.IPv6Protocol {
-		prefix = "["
-		suffix = "]"
+		// IPv6: return a bracketed version of the IP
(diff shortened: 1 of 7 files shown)


pkg/utils/net/ip.go
```diff
@@ -48,3 +49,16 @@ func ToIPFamily(ipStr string) corev1.IPFamily {
 	}
 	return corev1.IPv6Protocol
 }
+
+// IPLiteralFor returns the given IP as a literal that can be used in a resource identifier.
+// For IPv6 that means returning a bracketed version of the IP.
+// The difference to net.JoinHostPort is that it also allows IP be to be a placeholder that will be resolved
```
thbkrkr, line 55:
```suggestion
// The difference to net.JoinHostPort is that it also allows IP to be a placeholder that will be resolved
```

thbkrkr (approved):
LGTM while waiting to decide if it's really worth to use `<pod>.<service>.svc.<cluster-domain>`.

My feeling is that it is simple and reasonable compared to the problem caused by the DNS name approach.

pkg/utils/net/ip.go
```diff
@@ -48,3 +49,16 @@ func ToIPFamily(ipStr string) corev1.IPFamily {
...
+// to the actual IP at a later time.
+func IPLiteralFor(ipOrPlaceholder string, ipFamily corev1.IPFamily) string {
+	var prefix, suffix string
```
barkbay, line 58: nit: I was wondering if we really need those variables since they are used only for IPv6.
Any reason to not avoid them and use something more explicit ?

```golang
func IPLiteralFor(ipOrPlaceholder string, ipFamily corev1.IPFamily) string {
	if ipFamily == corev1.IPv6Protocol {
                 // IPv6: return a bracketed version of the IP
		return fmt.Sprintf("[%s]", ipOrPlaceholder)
	}
        // IPv4: leave the place holder as is
	return ipOrPlaceholder
}
```

barkbay (approved):
LGTM

PR #3727, opened 2020-09-08
Editing - Run Beats on ECK

alaudazzi:
This PR:

- Reorganizes some ordered lists
- Reworks a few sentences
- Merges CAUTION with IMPORTANT as related to the same topic (examples)
- Relates to [#65638](https://github.com/elastic/cloud/issues/65638)

## Notes for the reviewer:

> Change `some_key: null` to `some_key: {}`, if it doesn't change config semantics.

Can you help clarify this sentence?

### DIFF
--- a/docs/orchestrating-elastic-stack-applications/beat.asciidoc
+++ b/docs/orchestrating-elastic-stack-applications/beat.asciidoc
@@ -116,7 +116,7 @@ You have two options:
 curl -u "elastic:$PASSWORD" -k "https://localhost:9200/filebeat-*/_search"
 ----
 +
-- Follow the Kibana deployment link:k8s-deploy-kibana.html[guide], log in and go to *Kibana* > *Discover*.
+- Follow the Kibana deployment <<{p}-deploy-kibana,guide>>, log in and go to *Kibana* > *Discover*.
 
 [id="{p}-beat-configuration"]
 == Configuration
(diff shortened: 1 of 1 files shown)


docs/orchestrating-elastic-stack-applications/beat.asciidoc
```diff
@@ -484,16 +493,16 @@ Deploys Packetbeat as a DaemonSet that monitors DNS on port `53` and HTTP(S) tra
...
+When `kubectl` is used to modify a resource, it calculates a diff between the user applied and the existing configuration. This diff has a special  link:https://tools.ietf.org/html/rfc7396#section-1[semantics], that forces the removal of a particular key by setting its value to `null`. When the user-applied configuration contains `some_key: null` (or equivalent `some_key: ~`), this is interpreted as an instruction to remove `some_key`, often resulting in a malformed configuration. In Beat configurations, this is often the case with link:https://www.elastic.co/guide/en/beats/filebeat/current/add-cloud-metadata.html[processors]. To avoid this:
 
-- change `some_key: null` to `some_key: {}` - if it doesn't change config semantics
-- instead of `config` use `configRef` and reference a Secret containing the config - Secret data doesn't go through the same diffing process
+- Change `some_key: null` to `some_key: {}`, if it doesn't change config semantics.
```
charith-elastic, line 504:
```suggestion
- Use `some_key: {}` (empty map) or `some_key: []` (empty array) instead of `some_key: null` if doing so does not affect the behaviour. This might not be possible in all cases as some applications distinguish between null values and empty values and behave differently.
```

docs/orchestrating-elastic-stack-applications/beat.asciidoc
```diff
@@ -484,16 +493,16 @@ Deploys Packetbeat as a DaemonSet that monitors DNS on port `53` and HTTP(S) tra
...
+- Instead of `config` use `configRef` and reference a Secret containing the configuration. Secret data doesn't go through the same diff process.
```
charith-elastic, line 505:
```suggestion
- Instead of using `config` to define configuration inline, use `configRef` and store the configuration in a Secret.
```

charith-elastic (commented):
I don't know whether this is the right time to address this: the use of singular "Beat" has always bothered me quite a bit. It does make sense in some places where the usage is definitely singular. However, in most places, using "Beats" would be more natural. Should we address that in this PR or in a different one?

alaudazzi:
I was wondering the same. Let's complete this PR and address Beat vs Beats in a separate one.

charith-elastic (approved):

PR #1698, opened 2019-09-09
Always enable file based user auth

thbkrkr:
The operator relies on file realm for creating all the internal users
including the built-in elastic user. This commit makes sure the file
realm is always enabled in case a user wants to configure additional
realms.
The order of the file realm is set to -100 to favor the fact that it is
used first but there is no complete guarantee about that. The user can
still configure a realm with an order less than -100 but it is very
unlikely given this is not suggested in the documentation (order starts to 0).
The main known issue could be to configure a realm that relies on a
system (e.g. Active Directory) that has a temporary lock-out period
after several successive failed login attempts, as being checked first
(https://www.elastic.co/guide/en/elastic-stack-overview/current/trouble-shoot-active-directory.html).

Resolves #1629.

PR #3728, opened 2020-09-08
Support new auth config format starting EnterpriseSearch 8.x

### LINKED ISSUE #3725: Use the new auth config format starting EnterpriseSearch 8.X
EnterpriseSearch 8.0.0-SNAPSHOT pods don't start due to the new auth config format. This is a breaking change from 8.X introduced to support multiple auth sources.

```
> k -n e2e-mercury logs test-ent-config-ref-4rln-ent-6c766cf98-5m4bf 

Invalid config file (/usr/share/enterprise-search/config/enterprise-search.yml):
The setting '#/ent_search/auth/source' was completely deprecated in version '8.0.0'. 
Please use the new auth config format ent_search.auth.<auth_name>.source.
```

Relates to #2599.

thbkrkr:
Resolves #3725.

### DIFF
--- a/pkg/controller/enterprisesearch/config.go
+++ b/pkg/controller/enterprisesearch/config.go
@@ -9,6 +9,11 @@ import (
 	"net"
 	"path/filepath"
 
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+
 	commonv1 "github.com/elastic/cloud-on-k8s/pkg/apis/common/v1"
 	entv1beta1 "github.com/elastic/cloud-on-k8s/pkg/apis/enterprisesearch/v1beta1"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/association"
@@ -17,14 +22,11 @@ import (
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common/driver"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common/reconciler"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common/settings"
+	"github.com/elastic/cloud-on-k8s/pkg/controller/common/version"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common/volume"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/enterprisesearch/name"
 	"github.com/elastic/cloud-on-k8s/pkg/utils/k8s"
 	netutil "github.com/elastic/cloud-on-k8s/pkg/utils/net"
-	corev1 "k8s.io/api/core/v1"
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/types"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


pkg/controller/enterprisesearch/config.go
```diff
@@ -251,16 +253,32 @@ func associationConfig(c k8s.Client, ent entv1beta1.EnterpriseSearch) (*settings
...
 
+	// origin of authenticated ent users setting changed starting 8.x
+	cfg := settings.MustCanonicalConfig(map[string]string{
+		"ent_search.auth.source": "elasticsearch-native",
+	})
+	ver, err := version.Parse(ent.Spec.Version)
+	if err != nil {
+		return nil, err
+	}
+	if ver.IsSameOrAfter(version.From(8, 0, 0)) {
```
barkbay, line 264: Should we add a unit test to validate that the config is updated for versions > 8 ?

pkg/controller/enterprisesearch/config.go
```diff
@@ -251,16 +253,32 @@ func associationConfig(c k8s.Client, ent entv1beta1.EnterpriseSearch) (*settings
 		return settings.NewCanonicalConfig(), nil
 	}
 
+	// origin of authenticated ent users setting changed starting 8.x
```
sebgl, line 256: should this comment be placed right above `if ver.IsSameOrAfter(version.From(8, 0, 0)) { `?

pkg/controller/enterprisesearch/config.go
```diff
@@ -251,16 +253,32 @@ func associationConfig(c k8s.Client, ent entv1beta1.EnterpriseSearch) (*settings
...
+		"ent_search.auth.source": "elasticsearch-native",
+	})
+	ver, err := version.Parse(ent.Spec.Version)
+	if err != nil {
+		return nil, err
+	}
+	if ver.IsSameOrAfter(version.From(8, 0, 0)) {
+		cfg = settings.MustCanonicalConfig(map[string]string{
+			"ent_search.auth.native1.source": "elasticsearch-native",
+			"ent_search.auth.native1.order": "-100",
```
sebgl, line 267: I'm curious why `-100` and not eg. `0`?

pkg/controller/enterprisesearch/config.go
```diff
@@ -251,16 +253,32 @@ func associationConfig(c k8s.Client, ent entv1beta1.EnterpriseSearch) (*settings
...
+		})
+	}
+
```
sebgl, line 270: A unit test would be nice to cover this.

pkg/controller/enterprisesearch/config.go
thbkrkr: The order of the realm is set to -100 to favor the fact that it is
used first but there is no complete guarantee about that. The user can
still configure a realm with an order less than -100 but it is very
unlikely given this is not suggested in the documentation (order starts to 0).

(as in https://github.com/elastic/cloud-on-k8s/pull/1698)

pkg/controller/enterprisesearch/config.go
thbkrkr: I hesitated. My reasoning was to put the comment at the beginning of the section that handles this: we prepare the default value, get the version and then use the new format. But I agree that it can be weird when we read the code flow. Let's move the comment.

barkbay (approved):
I did a test with `*:8.0.0-SNAPSHOT` docker images, I have the following errors in the logs:

```
[2020-09-09T13:45:29.604+00:00][1][2422][worker][ERROR]: Error in processing loop; Sleeping for 4s before trying again
Elasticsearch::Transport::Transport::Errors::BadRequest ([400] {"error":{"root_cause":[{"type":"illegal_argument_exception","reason":"invalid parameter [size], use [max_docs] instead"}],"type":"illegal_argument_exception","reason":"invalid parameter [size], use [max_docs] instead"},"status":400})
	/usr/share/enterprise-search/lib/war/gems/gems/elasticsearch-transport-7.0.0/lib/elasticsearch/transport/transport/base.rb:218:in `__raise_transport_error'
	/usr/share/enterprise-search/lib/war/gems/gems/elasticsearch-transport-7.0.0/lib/elasticsearch/transport/transport/base.rb:338:in `perform_request'
	/usr/share/enterprise-search/lib/war/gems/gems/elasticsearch-transport-7.0.0/lib/elasticsearch/transport/transport/http/faraday.rb:37:in `perform_request'
	/usr/share/enterprise-search/lib/war/gems/gems/elasticsearch-transport-7.0.0/lib/elasticsearch/transport/client.rb:160:in `perform_request'
	/usr/share/enterprise-search/lib/war/gems/gems/elasticsearch-api-7.0.0/lib/elasticsearch/api/actions/update_by_query.rb:88:in `update_by_query'
	/usr/share/enterprise-search/lib/war/vendor/esqueues-me/lib/esqueues_me/worker.rb:123:in `claim_work'
	/usr/share/enterprise-search/lib/war/vendor/esqueues-me/lib/esqueues_me/worker.rb:98:in `work'
	/usr/share/enterprise-search/lib/war/vendor/esqueues-me/lib/esqueues_me/worker.rb:69:in `block in start_processing_loop!'
	org/jruby/RubyKernel.java:1446:in `loop'
	/usr/share/enterprise-search/lib/war/vendor/esqueues-me/lib/esqueues_me/worker.rb:63:in `start_processing_loop!'
	/usr/share/enterprise-search/lib/war/vendor/esqueues-me/lib/esqueues_me.rb:106:in `start!'
	/usr/share/enterprise-search/lib/war/shared_togo/lib/shared_togo.class:732:in `start_esqueues_me!'
	/usr/share/enterprise-search/lib/war/shared_togo/lib/shared_togo.class:145:in `start_worker!'
	<script>:1:in `<main>'
```

But it does not seem related to an authentication problem, so LGTM.

thbkrkr:
I did not see that but it might be worth contacting the ENT team about this if we know how to reproduce it. It seems related to https://github.com/elastic/elasticsearch/pull/43373. The log indicates `elasticsearch-api-7.0.0` is still in use.

Issue #3110, opened 2020-05-19
Update Openshift documentation re. APM Server deployment


barkbay:
With the merge of https://github.com/elastic/beats/pull/12905 Openshift users should be able to run the APM Server with the default, restricted, SCC and without any additional privileges.

Let's update our documentation when the PR is included in a release (should be 7.9).

SeanPlacchetti:
@barkbay Just making sure I'm reading this correctly, after the 7.9 drop of Elasticsearch, we'll be able to deploy the beats from the ECK pattern without the block that's in the current examples for the beats:
```
    podTemplate:
      spec:
        securityContext:
          runAsUser: 0
```
But for now, the beats need to run as `user:0` and therefore will not run in OpenShift 3.11

barkbay:
@SeanPlacchetti this issue only applies to the APM Server, before 7.9 it is not possible to run the APM Server Docker image with a random user ID.
Beats may still require some extended privileges to be able to access the filesystem of the host for example.

PR #3742, opened 2020-09-10
Editing - Deploy ECK on OpenShift

alaudazzi:
This PR:

- Suggests some minor edits
- Relates to [#65639](https://github.com/elastic/cloud/issues/65639)

## Notes for the reviewer

Would it be better to rename [Advanced topics](https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-advanced-topics.html) into something like "Deep dive"?
This change would impact title and file ID. If you are OK, I'll handle that in a separate PR.

### DIFF
--- a/docs/advanced-topics/openshift.asciidoc
+++ b/docs/advanced-topics/openshift.asciidoc
@@ -16,7 +16,7 @@ This page shows how to run ECK on OpenShift.
 * <<{p}-openshift-apm,Deploy an APM Server instance with a route>>
 * <<{p}-openshift-beats,Grant privileged permissions to Beats>>
 
-NOTE: Only Elasticsearch and Kibana are compatible with the `restricted` https://docs.openshift.com/container-platform/4.1/authentication/managing-security-context-constraints.html[Security Context Constraint]. To run the APM Server on OpenShift you must allow the Pod to run with the `anyuid` SCC as described in <<{p}-openshift-apm,Deploy an APM Server instance with a route>>.
+NOTE: Only Elasticsearch and Kibana are compatible with the `restricted` https://docs.openshift.com/container-platform/4.1/authentication/managing-security-context-constraints.html[Security Context Constraint]. To run the APM Server on OpenShift you must allow the Pod to run with the `anyuid` SCC as described in <<{p}-openshift-apm>>.
 
 [float]
 [id="{p}-openshift-before-you-begin"]
--- a/docs/advanced-topics/openshift.asciidoc
+++ b/docs/advanced-topics/openshift.asciidoc
@@ -16,7 +16,7 @@ This page shows how to run ECK on OpenShift.
 * <<{p}-openshift-apm,Deploy an APM Server instance with a route>>
 * <<{p}-openshift-beats,Grant privileged permissions to Beats>>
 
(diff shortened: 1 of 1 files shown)


docs/advanced-topics/openshift.asciidoc
```diff
@@ -16,7 +16,7 @@ This page shows how to run ECK on OpenShift.
 * <<{p}-openshift-apm,Deploy an APM Server instance with a route>>
 * <<{p}-openshift-beats,Grant privileged permissions to Beats>>
 
-NOTE: Only Elasticsearch and Kibana are compatible with the `restricted` https://docs.openshift.com/container-platform/4.1/authentication/managing-security-context-constraints.html[Security Context Constraint]. To run the APM Server on OpenShift you must allow the Pod to run with the `anyuid` SCC as described in <<{p}-openshift-apm,Deploy an APM Server instance with a route>>
+NOTE: Only Elasticsearch and Kibana are compatible with the `restricted` https://docs.openshift.com/container-platform/4.1/authentication/managing-security-context-constraints.html[Security Context Constraint]. To run the APM Server on OpenShift you must allow the Pod to run with the `anyuid` SCC as described in <<{p}-openshift-apm,Deploy an APM Server instance with a route>>.
```
charith-elastic, line 19:
```suggestion
NOTE: Only Elasticsearch and Kibana are compatible with the `restricted` https://docs.openshift.com/container-platform/4.1/authentication/managing-security-context-constraints.html[Security Context Constraint]. To run the APM Server on OpenShift you must allow the Pod to run with the `anyuid` SCC as described in <<{p}-openshift-apm>>.
```

docs/advanced-topics/openshift.asciidoc
```diff
@@ -292,7 +292,7 @@ chcon -Rt svirt_sandbox_file_t /var/lib/elastic/heartbeat/heartbeat-data/
 
 Repeat this step on all the hosts where the heartbeat Pod can be deployed.
 
-Some Beats may require additional permissions. For example `Filebeat` needs additional privileges to read other containers logs on the host. This can be done using the `privileged` field in the security context of the container spec:
+Some Beats may require additional permissions. For example `Filebeat` needs additional privileges to read other containers logs on the host. In this case, you can use the `privileged` field in the security context of the container spec:
```
charith-elastic, line 295:
```suggestion
Some Beats may require additional permissions. For example, `Filebeat` needs additional privileges to read other container logs on the host. In this case, you can use the `privileged` field in the security context of the container spec:
```

docs/advanced-topics/openshift.asciidoc
barkbay, line 19: @alaudazzi  Do you think we can take the opportunity of this PR to address https://github.com/elastic/cloud-on-k8s/issues/3110 ? This statement should not be true anymore since APM 7.9.

Also I'm wondering if we should include the following cases:
* Beats for which a [dedicated section already exists](https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-openshift-beats.html)
* Enterprise Search 7.9 [which is known to be incompatible with Openshift](https://github.com/elastic/cloud-on-k8s/issues/3656).

Maybe we should remove "Only Elasticsearch and Kibana are compatible" to be more explicit about the particular cases.

alaudazzi: Thank you @barkbay for pointing this out. In this case we should also update the section https://www.elastic.co/guide/en/cloud-on-k8s/1.2/k8s-openshift-apm.html, right?
Would it be better to address this in a separare PR?

charith-elastic (approved):

PR #3743, opened 2020-09-10
Default to IPv4 if no valid IP is given in IP family auto-detect

pebrc:
This mostly improves the situation for local development use of the operator via `make run` when no valid IP is available.


barkbay (approved):
LGTM

Thanks !

sebgl (approved):
LGTM

PR #3747, opened 2020-09-14
Skip Enterprise Search 7.9.1 e2e tests on OCP

sebgl:
Enterprise Search 7.9.1 was released and has the same problem as 7.9.0,
hence cannot run on Openshfit. See https://github.com/elastic/cloud-on-k8s/issues/3656.


barkbay (approved):
LGTM

PR #3748, opened 2020-09-14
Add brackets around address in kind IPv6 E2E test

### LINKED ISSUE #3737: Port is interpreted as a part of the address in kind IPv6 E2E test
In the [test failure](https://devops-ci.elastic.co/job/cloud-on-k8s-e2e-tests-kind-k8s-versions/207/testReport/github/com_elastic_cloud-on-k8s_test_e2e_beat/Run_tests_on_different_versions_of_vanilla_K8s___1_17_5_IPv6___TestMetricbeatStackMonitoringRecipe_ES_data_should_pass_validations/) we can see:

```
won't start runner: 1 error: host parsing failed for kibana-stats: error parsing URL: address fd00:10:244:0:3::51:5601: too many colons in address"}
```

Which is caused by appending the port (5601) to the IPv6 address of the host. The port is interpreted as the last group, making the group count 7. The `::` can't be used to omit a single group, hence the error.

This adress is constructed from Metricbeat config:

https://github.com/elastic/cloud-on-k8s/blob/15a5fd79d4b96dadfa45bbc6cf1d116b7da97e03/config/recipes/beats/stack_monitoring.yaml#L52

It works for IPv4, but for IPv6, the proper format would be `hosts: "https://[${data.host}]:${data.port}"`. This would require switching somewhere between IPv4 and IPv6, so I'm wondering if there is a better solution.

barkbay:
This PR adds some brackets around `${data.host}` when running in a IPv6 environment.

Fix #3737

~_Still a draft as I'm not done with the tests_~

### DIFF
--- a/.ci/setenvconfig
+++ b/.ci/setenvconfig
@@ -122,6 +122,7 @@ TEST_LICENSE = /go/src/github.com/elastic/cloud-on-k8s/.ci/test-license.json
 MONITORING_SECRETS = /go/src/github.com/elastic/cloud-on-k8s/.ci/monitoring-secrets.json
 
 OPERATOR_IMAGE = $JKS_PARAM_OPERATOR_IMAGE
+TEST_TIMEOUT = 10m
 KIND_NODE_IMAGE = ${kindNodeImage}
 IP_FAMILY=$4
 
--- a/test/e2e/beat/recipes_test.go
+++ b/test/e2e/beat/recipes_test.go
@@ -108,7 +108,7 @@ func TestMetricbeatStackMonitoringRecipe(t *testing.T) {
 			newSecretName := strings.Replace(currSecretName, "elasticsearch", fmt.Sprintf("elasticsearch-%s", builder.Suffix), 1)
 			builder.Beat.Spec.Deployment.PodTemplate.Spec.Containers[0].Env[1].ValueFrom.SecretKeyRef.Name = newSecretName
 
-			// We are using the pod's IP address exposed though the downward API to detect if the test is running in an IPv6 environment.
+			// We are using the pod's IP address exposed through the downward API to detect if the test is running in an IPv6 environment.
 			if net.ToIPFamily(os.Getenv(EnvPodIP)) == corev1.IPv6Protocol {
 				// In an IPv6 environment we need to patch the configuration to add some brackets around the data.host variable.
 				json, err := builder.Beat.Spec.Config.MarshalJSON()
--- a/test/e2e/beat/recipes_test.go
+++ b/test/e2e/beat/recipes_test.go
@@ -6,6 +6,7 @@ package beat
 
 import (
 	"fmt"
+	"os"
 	"path"
 	"strings"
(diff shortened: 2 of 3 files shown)


.ci/setenvconfig
```diff
@@ -122,6 +122,7 @@ TEST_LICENSE = /go/src/github.com/elastic/cloud-on-k8s/.ci/test-license.json
 MONITORING_SECRETS = /go/src/github.com/elastic/cloud-on-k8s/.ci/monitoring-secrets.json
 
 OPERATOR_IMAGE = $JKS_PARAM_OPERATOR_IMAGE
+TEST_TIMEOUT = 10m
```
sebgl, line 125: why?

test/e2e/beat/recipes_test.go
```diff
@@ -101,6 +104,17 @@ func TestMetricbeatStackMonitoringRecipe(t *testing.T) {
 			currSecretName := builder.Beat.Spec.Deployment.PodTemplate.Spec.Containers[0].Env[1].ValueFrom.SecretKeyRef.Name
 			newSecretName := strings.Replace(currSecretName, "elasticsearch", fmt.Sprintf("elasticsearch-%s", builder.Suffix), 1)
 			builder.Beat.Spec.Deployment.PodTemplate.Spec.Containers[0].Env[1].ValueFrom.SecretKeyRef.Name = newSecretName
+
+			if net.ToIPFamily(os.Getenv(essettings.EnvPodIP)) == corev1.IPv6Protocol {
```
sebgl, line 108: A comment would be nice to explain the trick

test/e2e/beat/recipes_test.go
```diff
@@ -16,7 +17,9 @@ import (
 	beatcommon "github.com/elastic/cloud-on-k8s/pkg/controller/beat/common"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common/settings"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common/version"
+	essettings "github.com/elastic/cloud-on-k8s/pkg/controller/elasticsearch/settings"
```
sebgl, line 20: nit: it's a bit strange to import this package here to grab `$POD_IP`, which is not related to es settings in that context. I think I would maybe re-declare the env var here.

sebgl (approved):
LGTM
Longer term I hope this gets fixed in Beats instead: https://github.com/elastic/cloud-on-k8s/issues/3737#issuecomment-692524901

.ci/setenvconfig
barkbay: Because when I ran the test manually I had to increase the timeout (with Kind everything is running on the same host, it can be very slow, there's a lot of I/O, and some beats spend some random times in a `CrashloopBackoff` state).
I suspect the test is almost reaching that limit and may hit it at some point. Increasing the timeout is harmless and will prevent to have some false positive. (I guess the timeout has been already increased to 10m for eks/aks and ocp for the same reason).

PR #3751, opened 2020-09-15
Update the OpenShift compatibility with 7.9

### LINKED ISSUE #3110: Update Openshift documentation re. APM Server deployment

alaudazzi:
This PR updates the OpenShift compatibility for APM  and Enterprise Search 7.9

Fixes #3110
Relates to #3742

### DIFF
--- a/docs/advanced-topics/openshift.asciidoc
+++ b/docs/advanced-topics/openshift.asciidoc
@@ -17,7 +17,6 @@ This page shows how to run ECK on OpenShift.
 * <<{p}-openshift-beats>>
 
 WARNING: For versions older than 7.9, APM Server and Enterprise Search are not compatible with `restricted` SCC. Use this <<{p}-openshift-apm,workaround>> to overcome this limitation.
-With version 7.9, the APM Server is compatible with `restricted` SCC, while for Enterprise Search the incompatibility remains and can be addressed with the same workaround.
 
 [float]
 [id="{p}-openshift-before-you-begin"]
--- a/docs/advanced-topics/openshift.asciidoc
+++ b/docs/advanced-topics/openshift.asciidoc
@@ -16,7 +16,7 @@ This page shows how to run ECK on OpenShift.
 * <<{p}-openshift-anyuid-workaround>>
 * <<{p}-openshift-beats>>
 
-WARNING: For versions older than 7.9, APM Server and Enterprise Search are not compatible with `restricted` SCC. Use this <<{p}-openshift-apm,workaround>> to overcome this limitation.
+WARNING: Some Docker images are incompatible with the `restricted` SCC. This is the case for the *APM Server before 7.9* and for the *Enterprise Search 7.9*. You can use this <<{p}-openshift-anyuid-workaround,workaround>> to run those images with the `anyuid` SCC.
 
 [float]
 [id="{p}-openshift-before-you-begin"]
--- a/docs/advanced-topics/openshift.asciidoc
+++ b/docs/advanced-topics/openshift.asciidoc
@@ -16,7 +16,7 @@ This page shows how to run ECK on OpenShift.
 * <<{p}-openshift-apm,Deploy an APM Server instance with a route>>
 * <<{p}-openshift-beats,Grant privileged permissions to Beats>>
 
-NOTE: Only Elasticsearch and Kibana are compatible with the `restricted` https://docs.openshift.com/container-platform/4.1/authentication/managing-security-context-constraints.html[Security Context Constraint]. To run the APM Server on OpenShift you must allow the Pod to run with the `anyuid` SCC as described in <<{p}-openshift-apm>>.
+WARNING: With Elastic Stack 7.9, note that APM Server is now compatible with OpenShift, see <<{p}-openshift-apm>>. Enterprise Search 7.9 is not yet compatible with Openshift.
 
 [float]
 [id="{p}-openshift-before-you-begin"]
@@ -157,7 +157,7 @@ oc get route -n elastic
 [id="{p}-openshift-apm"]
 == Deploy an APM Server instance with a route
 
-It is currently not possible to run the APM Server with the `restricted` SCC. A possible workaround is to allow the Pod to run with the default `uid 1000` by assigning it to the `anyuid` SCC:
+With Elastic Stack 7.9, it is possible to run the APM Server with the `restricted` SCC. For versions older than 7.9, you can use this workaround to allow the Pod to run with the default `uid 1000` by assigning it to the `anyuid` SCC:
 
 . Create a service account to run the APM Server:
 +
--- a/docs/advanced-topics/openshift.asciidoc
+++ b/docs/advanced-topics/openshift.asciidoc
@@ -157,7 +157,7 @@ oc get route -n elastic
 [id="{p}-openshift-apm"]
 == Deploy an APM Server instance with a route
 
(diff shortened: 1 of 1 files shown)


docs/advanced-topics/openshift.asciidoc
```diff
@@ -16,7 +16,7 @@ This page shows how to run ECK on OpenShift.
 * <<{p}-openshift-apm,Deploy an APM Server instance with a route>>
 * <<{p}-openshift-beats,Grant privileged permissions to Beats>>
 
-NOTE: Only Elasticsearch and Kibana are compatible with the `restricted` https://docs.openshift.com/container-platform/4.1/authentication/managing-security-context-constraints.html[Security Context Constraint]. To run the APM Server on OpenShift you must allow the Pod to run with the `anyuid` SCC as described in <<{p}-openshift-apm>>.
+WARNING: With Elastic Stack 7.9, note that APM Server is now compatible with OpenShift, see <<{p}-openshift-apm>>. Enterprise Search 7.9 is not yet compatible with Openshift.
```
barkbay, line 19: > Enterprise Search 7.9 is not yet compatible with Openshift

Enterprise Search before 7.9 was compatible with the `restricted` SCC . We have an incompatibility starting with 7.9.x  On a second thought I'm wondering if the same workaround used for APM < 7.9 can be used for Enterprise Search = 7.9.x.

I need to do the test, will update my review with my findings.

docs/advanced-topics/openshift.asciidoc
```diff
@@ -157,7 +157,7 @@ oc get route -n elastic
 [id="{p}-openshift-apm"]
 == Deploy an APM Server instance with a route
 
-It is currently not possible to run the APM Server with the `restricted` SCC. A possible workaround is to allow the Pod to run with the default `uid 1000` by assigning it to the `anyuid` SCC:
+With Elastic Stack 7.9, it is possible to run the APM Server with the `restricted` SCC. For versions older than 7.9, you can use this workaround to allow the Pod to run with the default `uid 1000` by assigning it to the `anyuid` SCC:
```
barkbay, line 160:
```suggestion
Starting with version 7.9, it is possible to run the APM Server with the `restricted` SCC. For versions older than 7.9, you can use this workaround to allow the Pod to run with the default `uid 1000` by assigning it to the `anyuid` SCC:
```

docs/advanced-topics/openshift.asciidoc
barkbay: I just did the test and I can confirm that if a user really want to deploy Enterprise Search 7.9.x on Openshift then the same workaround as the one described for the APM Server < 7.9 can be applied.

The only thing I'm not sure about is how to reflect those 2 two particular cases and the workaround in the doc. I guess that if we want to broaden the scope of the workaround we have to:
* Update the title `Deploy an APM Server instance with a route` to something more generic like `Deploy Docker images with the anyuid SCC` 
* Highlight that the workaround only aims to circumvent incompatibilities with APM < 7.9 and Enterprise Search = 7.9.x

alaudazzi: OK, thank you @barkbay!

alaudazzi: @barkbay I reworked the Warning and turned the APM title to something more generic (as you suggested). Let me know if this looks any better.

docs/advanced-topics/openshift.asciidoc
```diff
@@ -9,14 +9,15 @@ endif::[]
...
-* <<{p}-openshift-beats,Grant privileged permissions to Beats>>
+* <<{p}-openshift-before-you-begin>>
+* <<{p}-openshift-deploy-the-operator>>
+* <<{p}-openshift-deploy-elasticsearch>>
+* <<{p}-openshift-deploy-kibana>>
+* <<{p}-openshift-apm>>
+* <<{p}-openshift-beats>>
 
-NOTE: Only Elasticsearch and Kibana are compatible with the `restricted` https://docs.openshift.com/container-platform/4.1/authentication/managing-security-context-constraints.html[Security Context Constraint]. To run the APM Server on OpenShift you must allow the Pod to run with the `anyuid` SCC as described in <<{p}-openshift-apm>>.
+WARNING: For versions older than 7.9, APM Server and Enterprise Search are not compatible with `restricted` SCC. Use this <<{p}-openshift-apm,workaround>> to overcome this limitation.
```
barkbay, line 19: I think we should let the user know that Enterprise Search 7.8 was compatible with Openshift. This is a specific bug in the 7.9 release. Otherwise they might be surprised to have something working with 7.8 and not with 7.9 ?

```suggestion
WARNING: Some Docker images are known to be incompatible with the `restricted` SCC. This is the case for the *APM Server before 7.9* and for the *7.9 release of Enterprise Search*. A <<{p}-openshift-anyuid-workaround,workaround>> is to run those images with the `anyuid` SCC.
```

docs/advanced-topics/openshift.asciidoc
```diff
@@ -9,14 +9,15 @@ endif::[]
...
-* <<{p}-openshift-deploy-the-operator,Deploy the operator>>
-* <<{p}-openshift-deploy-elasticsearch,Deploy an Elasticsearch instance with a route>>
-* <<{p}-openshift-deploy-kibana,Deploy a Kibana instance with a route>>
-* <<{p}-openshift-apm,Deploy an APM Server instance with a route>>
-* <<{p}-openshift-beats,Grant privileged permissions to Beats>>
+* <<{p}-openshift-before-you-begin>>
+* <<{p}-openshift-deploy-the-operator>>
+* <<{p}-openshift-deploy-elasticsearch>>
+* <<{p}-openshift-deploy-kibana>>
+* <<{p}-openshift-apm>>
```
barkbay, line 16:
```suggestion
* <<{p}-openshift-anyuid-workaround>>
```

docs/advanced-topics/openshift.asciidoc
```diff
@@ -155,9 +156,9 @@ oc get route -n elastic
 ----
 
 [id="{p}-openshift-apm"]
```
barkbay, line 158:
```suggestion
[id="{p}-openshift-anyuid-workaround"]
```

docs/advanced-topics/openshift.asciidoc
```diff
@@ -9,14 +9,15 @@ endif::[]
...
+* <<{p}-openshift-before-you-begin>>
+* <<{p}-openshift-deploy-the-operator>>
+* <<{p}-openshift-deploy-elasticsearch>>
+* <<{p}-openshift-deploy-kibana>>
+* <<{p}-openshift-apm>>
+* <<{p}-openshift-beats>>
 
-NOTE: Only Elasticsearch and Kibana are compatible with the `restricted` https://docs.openshift.com/container-platform/4.1/authentication/managing-security-context-constraints.html[Security Context Constraint]. To run the APM Server on OpenShift you must allow the Pod to run with the `anyuid` SCC as described in <<{p}-openshift-apm>>.
+WARNING: For versions older than 7.9, APM Server and Enterprise Search are not compatible with `restricted` SCC. Use this <<{p}-openshift-apm,workaround>> to overcome this limitation.
+With version 7.9, the APM Server is compatible with `restricted` SCC, while for Enterprise Search the incompatibility remains and can be addressed with the same workaround.
```
barkbay, line 20:
```suggestion
```

Issue #325, opened 2019-01-29
Support dynamic PVC resize


sebgl:
If the storageClass supports it, we should be able to resize a PVC at runtime, without the need to create a new pod with a new larger PVC.

Edit: we decided to workaround the StatefulSet limitations and implement pvc resize ourselves. See [this post](https://github.com/elastic/cloud-on-k8s/issues/325#issuecomment-692591862).

___

For users landing there because they want to resize their existing volumes: the [current workaround](https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-orchestration.html#k8s-orchestration-limitations) is to rename the nodeSet to resize and give it new storage reqs. This triggers data to be copied over to new volumes by ECK.

For example, if we have:

```yaml
spec:
  nodeSets:
  - name: default
    count: 3
    volumeClaimTemplates:
    - metadata:
        name: elasticsearch-data
      spec:
        accessModes:
        - ReadWriteOnce
        resources:
          requests:
            storage: 5Gi
        storageClassName: standard
```

And we want to resize from 5GB to 10GB, we can do the following:

* Rename the nodeSet `default` to `default2`
* Increase the storage request
* Apply the yaml manifest

```yaml
spec:
  nodeSets:
  - name: default2
    count: 3
    volumeClaimTemplates:
    - metadata:
        name: elasticsearch-data
      spec:
        accessModes:
        - ReadWriteOnce
        resources:
          requests:
            storage: 10Gi
        storageClassName: standard
```

ECK will create Pods for the new nodeSet `default2` with the requested storage size (10Gi). Then, it will migrate data away from Pods in `default`. Once data migration is over, all Pods from `default` are removed automatically. Migrating data may take some time depending on how large it is.

sebgl:
This gets a bit trickier with the StatefulSets implementation, since we cannot edit the VolumeClaim section of the StatefulSets spec.

What we can technically do is:
* resize the PVC/PV
* (maybe) restart the pod (not sure that's required, to be checked)

But any new replica added to the StatefulSet would still use the initial volume size.
This could be [handled by the StatefulSet controller at some point](https://github.com/kubernetes/kubernetes/issues/68737), we may not want to rush into our own hacky implementation.

anyasabo:
A comment near the end of the issue mentioned a potential workaround to still use stateful sets and update PVCs:
https://github.com/kubernetes/kubernetes/issues/68737#issuecomment-498470138


[...]

PR #3752, opened 2020-09-15
Support Elasticsearch volumes expansion

sebgl:
Add support for resizing Elasticsearch volumes, by simply editing the storage requests in the `volumeClaimTemplates` section of the spec.

**Demo**

assembled

**Constraints**

* Volume size can only be increased, not decreased (k8s limitation, not likely to change anytime soon)
* Only possible if the storage class specifies `allowVolumeExpansion: true`
* Hot filesystem resize without re-creating the Pod is [possible in many cases](https://kubernetes.io/blog/2018/07/12/resizing-persistent-volumes-using-kubernetes/#online-file-system-expansion) (GCE-PD, AWS-EBS, Azure, TopoLVM, etc.). If the storage driver does not allow it, it is necessary to *manually* delete the Pods. ECK does not do it.

**Implementation overview**

Resizing PVCs is [not supported by the StatefulSet controller](https://github.com/kubernetes/kubernetes/issues/68737) at this time. The volume section of the StatefulSet spec is immutable.
However, it is possible to work around the StatefulSet limitations this way:
1. Update the storage requests spec of all existing PVCs: they are immediately resized by the storage driver, if inline expansion is supported. Otherwise Pods need to be recreated.
2. We now have resized volumes, but the StatefulSet still specifies the old volume size. Any new replica will end up with a smaller PVC, that in turn needs to be resized.
3. Delete the StatefulSet (but not the Pods it owns), then recreate it with the new volume spec.
4. Address the case where Pods may become orphans if something happens while the StatefulSet is getting recreated, by ensuring we always recreate deleted StatefulSets, and ensure we set a temporary ownerRef on the Pods while they are not owned by the deleted StatefulSet anymore.

**Follow-up (in other PRs)**

- [ ] Add E2E tests to cover volume expansion: https://github.com/elastic/cloud-on-k8s/pull/3769
- [ ] Write documentation
- (later) Figure out a way to automatically recreate Pods if inline filesystem expansion is not supported: https://github.com/elastic/cloud-on-k8s/issues/3757
- Check whether storage class is compatible with volume expansion: https://github.com/elastic/cloud-on-k8s/issues/3767

Relates https://github.com/elastic/cloud-on-k8s/issues/325.


hack/manifest-gen/assets/charts/eck/templates/_helpers.tpl
```diff
@@ -94,6 +94,14 @@ RBAC permissions
...
   - patch
   - delete
+- apiGroups:
+  - storage.k8s.io
+  resources:
+  - storageclasses
+  verbs:
+  - get
+  - list
+  - watch
```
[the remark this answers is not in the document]
sebgl: Hm now that I think about it again: since storage classes are a cluster-wide resource, they need to live in a ClusterRole with a ClusterRoleBinding, not in a Role with a RoleBinding?

pkg/controller/elasticsearch/driver/pvc_expansion.go
```diff
@@ -0,0 +1,207 @@
...
+func isStorageExpansion(expectedSize *resource.Quantity, actualSize *resource.Quantity) (bool, error) {
+	if expectedSize == nil || actualSize == nil {
+		// not much to compare if storage size is unspecified
+		return false, nil
+	}
+	switch expectedSize.Cmp(*actualSize) {
+	case 0: // same size
+		return false, nil
+	case -1: // decrease
+		return false, fmt.Errorf("storage size cannot be decreased from %s to %s", actualSize.String(), expectedSize.String())
```
david-kow, line 158:
```suggestion
		return false, fmt.Errorf("storage size cannot be decreased, but an attempt was made to change it from %s to %s", actualSize.String(), expectedSize.String())
```

Or some other wording that clearly states that decreasing is not possible. Now it sounds like it is, but not to this particular size.

david-kow (commented):
Nice work @sebgl. I did some testing and left some comments, nothing major.

pkg/apis/elasticsearch/v1/validations.go
```diff
@@ -207,29 +204,6 @@ func checkNodeSetNameUniqueness(es *Elasticsearch) field.ErrorList {
...
-
-		// ssets do not allow modifications to fields other than 'replicas', 'template', and 'updateStrategy'
-		// reflection isn't ideal, but okay here since the ES object does not have the status of the claims.
-		// Checking semantic equality here allows providing PVC storage size with different units (eg. 1Ti vs. 1024Gi).
-		if !apiequality.Semantic.DeepEqual(node.VolumeClaimTemplates, currNode.VolumeClaimTemplates) {
-			errs = append(errs, field.Invalid(field.NewPath("spec").Child("nodeSet").Index(i).Child("volumeClaimTemplates"), node.VolumeClaimTemplates, pvcImmutableMsg))
-		}
-	}
-	return errs
-}
```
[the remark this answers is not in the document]
sebgl: It would allow:
- exact same volumeClaimTemplates content
- increased storage requests only if the underlying storage class allows it

It would disallow everything else, for example:
- any other change in the volumeClaimTemplates
- storage requests decrease

pkg/apis/elasticsearch/v1/validations.go
```diff
@@ -52,7 +50,6 @@ type updateValidation func(*Elasticsearch, *Elasticsearch) field.ErrorList
 var updateValidations = []updateValidation{
 	noDowngrades,
 	validUpgradePath,
-	pvcModification,
```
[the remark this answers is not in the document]
sebgl: Yes! https://github.com/elastic/cloud-on-k8s/pull/3752#discussion_r488674940

pkg/controller/elasticsearch/driver/pvc_expansion.go
```diff
@@ -0,0 +1,207 @@
...
+func deleteSsetForClaimResize(k8sClient k8s.Client, expectedSset appsv1.StatefulSet, actualSset appsv1.StatefulSet) (bool, error) {
+	shouldRecreate, err := needsRecreate(k8sClient, expectedSset, actualSset)
+	if err != nil {
+		return false, err
+	}
+	if !shouldRecreate {
+		return false, nil
+	}
+
+	log.Info("Deleting StatefulSet to account for resized PVCs, it will be recreated automatically",
```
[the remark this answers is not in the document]
sebgl: Yeah I think the 2nd time the cache is not up-to-date: the StatefulSet still appears there, or may be in the process of being deleted. I don't know if it's worth fixing (probably doable through expectations but adds complexity) - I guess it's also true for other reconciliation steps.

sebgl:
There's a concern that some users may not be allowed to deploy ECK if it needs read access to storage classes (cluster-wide) resource.
I'll update this PR to remove any storage class checks, and created another issue to deal with that specific point: https://github.com/elastic/cloud-on-k8s/issues/3767

hack/manifest-gen/assets/charts/eck/templates/_helpers.tpl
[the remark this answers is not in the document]
sebgl: https://github.com/elastic/cloud-on-k8s/pull/3752#issuecomment-694339216

sebgl:
Jenkins test this please.

barkbay (commented):
The PR works well, great job 👍 
  
  I'm trying to understand how deleting a StatefulSet could be considered as "unsafe" or lead to some odd edge cases.
  I think we are assuming that, when the StatefulSet is deleted, a new one is recreated briefly. My concern is that there is no guarantee about how long is "briefly", there're might be several reasons for the creation of the new StaetfulSet to be delayed (bug, network partition, overload...)
  
  I just want to be sure we all share and document some potential edge cases which might lead to resource leaks or data loss.

When the StatefulSet is deleted:
  
  * Pods are left without any owner
  * there is no strong guarantee that a StatefulSet with the same name will be recreated

In this context we may hit the following situations:
 
### User deletes the Elasticsearch resource while the new StatefulSet is not yet created
  
  Pods that don't have an owner are not garbage collected. It leads to a resource leak, users must delete them manually. A workaround would be add the Elasticsearch resource as a second owner, but I'm not sure if it can be done through the `PodTemplate`.

### User eventually decides to rename the nodeSet

  1. If the previous nodeSet is removed/renamed while the operator has just deleted the StatefulSet then `GarbageCollectPVCs` decides to delete the PVC (because no StatefulSet owns them anymore): 

```
NAME                                                              HEALTH   NODES   VERSION   PHASE   AGE
elasticsearch.elasticsearch.k8s.elastic.co/elasticsearch-sample   green    4       7.9.1     Ready   15m

NAME                                              READY   AGE     CONTAINERS      IMAGES
statefulset.apps/elasticsearch-sample-es-data-2   3/3     9m24s   elasticsearch   docker.elastic.co/elasticsearch/elasticsearch:7.9.1
statefulset.apps/elasticsearch-sample-es-master   1/1     15m     elasticsearch   docker.elastic.co/elasticsearch/elasticsearch:7.9.1

NAME                                                                        STATUS        VOLUME                                     CAPACITY   ACCESS MODES   STORAGECLASS   AGE     VOLUMEMODE
persistentvolumeclaim/elasticsearch-data-elasticsearch-sample-es-data-1-0   Terminating   pvc-5400c0e2-819d-4505-a78b-def432aa4c87   2Gi        RWO            gp2            15m     Filesystem
persistentvolumeclaim/elasticsearch-data-elasticsearch-sample-es-data-1-1   Terminating   pvc-d3ce7e0e-b84a-4023-8bf6-b95dedc84a0b   2Gi        RWO            gp2            15m     Filesystem
persistentvolumeclaim/elasticsearch-data-elasticsearch-sample-es-data-1-2   Terminating   pvc-447d17b7-ffc3-4ba7-ad27-7c0b6facae5b   2Gi        RWO            gp2            15m     Filesystem
persistentvolumeclaim/elasticsearch-data-elasticsearch-sample-es-data-2-0   Bound         pvc-c5365b9f-12a5-4cb7-882b-1bd19f719036   2Gi        RWO            gp2            9m24s   Filesystem
persistentvolumeclaim/elasticsearch-data-elasticsearch-sample-es-data-2-1   Bound         pvc-7ee9d880-75ac-4c7c-b181-2cceeba582f2   2Gi        RWO            gp2            9m24s   Filesystem
persistentvolumeclaim/elasticsearch-data-elasticsearch-sample-es-data-2-2   Bound         pvc-46d43fdb-fdc6-4b74-898b-7e069cea9cad   2Gi        RWO            gp2            9m24s   Filesystem
persistentvolumeclaim/elasticsearch-data-elasticsearch-sample-es-master-0   Bound         pvc-8f2957dc-7cc1-4987-b9fd-f8dcc345fff8   1Gi        RWO            gp2            15m     Filesystem
```

  2. Shards are not migrated to new nodes, when the previous ones are deleted (by a human or as part of a Pod eviction) it may lead to data loss.

```
NAME                                                              HEALTH   NODES   VERSION   PHASE   AGE
elasticsearch.elasticsearch.k8s.elastic.co/elasticsearch-sample   red      4       7.9.1     Ready   25m

NAME                                              READY   AGE   CONTAINERS      IMAGES
statefulset.apps/elasticsearch-sample-es-data-2   3/3     19m   elasticsearch   docker.elastic.co/elasticsearch/elasticsearch:7.9.1
statefulset.apps/elasticsearch-sample-es-master   1/1     25m   elasticsearch   docker.elastic.co/elasticsearch/elasticsearch:7.9.1

NAME                                                                        STATUS   VOLUME                                     CAPACITY   ACCESS MODES   STORAGECLASS   AGE   VOLUMEMODE
persistentvolumeclaim/elasticsearch-data-elasticsearch-sample-es-data-2-0   Bound    pvc-c5365b9f-12a5-4cb7-882b-1bd19f719036   2Gi        RWO            gp2            19m   Filesystem
persistentvolumeclaim/elasticsearch-data-elasticsearch-sample-es-data-2-1   Bound    pvc-7ee9d880-75ac-4c7c-b181-2cceeba582f2   2Gi        RWO            gp2            19m   Filesystem
persistentvolumeclaim/elasticsearch-data-elasticsearch-sample-es-data-2-2   Bound    pvc-46d43fdb-fdc6-4b74-898b-7e069cea9cad   2Gi        RWO            gp2            19m   Filesystem
persistentvolumeclaim/elasticsearch-data-elasticsearch-sample-es-master-0   Bound    pvc-8f2957dc-7cc1-4987-b9fd-f8dcc345fff8   1Gi        RWO            gp2            25m   Filesystem
```

I just want to be sure that we are aligned on the tradeoff of this feature or if we want to consider something that could handle the situations described above.

sebgl:
@barkbay thanks, I think your concern, while being an edge case, is very valid.

I'm thinking about a different implementation where, before deleting a StatefulSet, we copy its entire spec in an annotation of the Elasticsearch resource. Very early in the Elasticsearch reconciliation the controller could recreate any deleted StatefulSet from that annotation, whether it is supposed to exist or not (before even computing the expected StatefulSets). So we always end up recreating deleted StatefulSet with their exact same spec (even if that means deleting them again at next reconciliation if they're not supposed to exist).
So far I think it's a much safer solution, it also simplifies the reconciliation logic. I need to spend more time thinking about any drawbacks it may bring, then I'll update this PR accordingly.

pkg/apis/elasticsearch/v1/validations.go
```diff
@@ -28,6 +30,7 @@ const (
 	nodeRolesInOldVersionMsg = "node.roles setting is not available in this version of Elasticsearch"
 	parseStoredVersionErrMsg = "Cannot parse current Elasticsearch version. String format must be {major}.{minor}.{patch}[-{label}]"
 	parseVersionErrMsg       = "Cannot parse Elasticsearch version. String format must be {major}.{minor}.{patch}[-{label}]"
+	pvcImmutableMsg          = "Volume claim templates cannot be modified. Only storage requests may be increased, if the storage class allows volume expansion."
```
david-kow, line 33:
```suggestion
	pvcImmutableMsg          = "Volume claim templates can only have their storage request increased (if the storage class allows volume expansion), any other modifications are forbidden."
```

Or some other rewording - right now the first sentence is contrary to the second one :)

sebgl:
Potential annotation-based flow to avoid orphan Pods if the Elasticsearch spec changes in the middle of restart:

```go
const StatefulSetsToRecreatePrefix = "elasticsearch.k8s.elastic.co/recreate-"

// 1. At the very beginning of nodes reconciliation, before any other change: recreate any sset that needs to
ssetsToRecreate := getAnnotationsWithPrefix(es, StatefulSetsToRecreatePrefix)
for name, value := range ssetsToRecreate {
    ssetName := strings.TrimLeft(value, StatefulSetsToRecreatePrefix)
    ssetToRecreate := json.Parse(value) // simplified
    actualSset, exists := k8sClient.Get(ssetName) // simplified
    switch {
        case !exists:
            // deleted before, create now
            // remove metadata we don't care about (uuid, selfLink, resourceVersion, etc.)
            k8sClient.Create(ssetToRecreate)
        case exists && actualSset.UUID == ssetToRecreate.UUID:
            // not deleted yet, delete
            k8sClient.Delete(actualSset) // with delete options to double-check uuid
        default:
            // exists but different UUID: already recreated, remove the annotation
            removeAnnotation(es, name)
            k8sClient.Update(es)
    }
}
if len(ssetsToRecreate)>0 {
    // requeue.
}

// 2. When computing changes as part of the upscale phase
if shouldBeRecreatedForPVCResize(actualSset, expectedSset) {
    es.Annotations[StatefulSetsToRecreatePrefix+actualSset.Name] = json.Marshal(actualSset)
    k8sClient.Update(es)
    // requeue
}
```

I think it would address the concerns above by always recreating the deleted StatefulSet with the exact same spec no matter what happens.
One concern with annotations (besides making sure there's no flaw in the logic) is their limited size: 256Kb. When testing locally a "normal" ES statefulset is around 10-15Kb. PodTemplate, affinity settings etc. can make that bigger though. We could span the value over multiple annotations if we exceed 256Kb, or compress+base64-encore the value to restrict the size?

pebrc:
> One concern with annotations (besides making sure there's no flaw in the logic) is their limited size: 256Kb. When testing locally a "normal" ES statefulset is around 10-15Kb. PodTemplate, affinity settings etc. can make that bigger though. We could span the value over multiple annotations if we exceed 256Kb, or compress+base64-encode the value to restrict the size?

Would it help to just store the `nodeSet` to recreate instead of the `StatefulSet`? The `StatefulSet` should be a derivative of the `nodeSet` and it should be less verbose especially if you only store the relevant parts of the `nodeSet`?

sebgl:
> Would it help to just store the nodeSet to recreate instead of the StatefulSet? The StatefulSet should be a derivative of the nodeSet and it should be less verbose especially if you only store the relevant parts of the nodeSet?

It sounds feasible. For a minimal nodeSet (only the relevant parts) this shrinks the annotation size to a few byes instead of ~12Kb for the equivalent StatefulSet (out of the available 256Kb). From a user perspective it means you have 255Kb available to define your nodeSet, instead of 244Kb. Overall I would argue this optimization does not make a big difference? (considering it may not fit the current code flow nicely)

I think I may be overthinking this: reaching 256Kb may not be likely to happen anyway. You really need a gigantic nodeSet specification for it to happen? I'm tempted to just error-out early if we reach that case, and eventually implement a multipart-annotation system if someone ever complains.

sebgl:
Last commit addresses the [orphan pods problem](https://github.com/elastic/cloud-on-k8s/pull/3752#pullrequestreview-492552664). Before deleting a StatefulSet, its entire spec is stored in an annotation of the Elasticsearch resource. A separate mechanism ensures StatefulSets that exist in those annotations are recreated before moving on with the reconciliation.
I ignored the max 256Kb annotation limit here, considering it's very unlikely to happen as explained above. If it does happen, the reconciliation will error out and it s still possible to delete the StatefulSet manually.

barkbay (commented):
I'm still doing some tests but overall it looks good 👍 
Also I just want to be sure we don't want to handle the case where the Elasticsearch resource is deleted before the statefulset is re-created, leaving some Pods orphaned ?

pkg/controller/elasticsearch/driver/upscale.go
```diff
@@ -104,22 +123,50 @@ func adjustZenConfig(k8sClient k8s.Client, es esv1.Elasticsearch, resources node
...
-	actual, alreadyExists := actualStatefulSets.GetByName(expected.Name)
 	expectedReplicas := sset.GetReplicas(expected)
-	actualReplicas := sset.GetReplicas(actual)
+
+	actual, alreadyExists := actualStatefulSets.GetByName(expected.Name)
+	actualReplicas := sset.GetReplicas(actual) // 0 if actual is empty
+	if !alreadyExists {
+		// In most cases, if the StatefulSet does not exist we're creating it for the first time (actualReplicas = 0).
+		// However if some Pods matching that missing StatefulSet exist we are likely in a situation where the
+		// StatefulSet is being re-created for volume expansion. In which case we want to take into account
```
barkbay, line 138: I'm wondering if it is still true: my understanding is that if a statefulset has been deleted this part of the code is never reached. The reason is that before `HandleUpscaleAndSpecChanges` we are checking if some recreations are in progress:

```golang
	// recreate any StatefulSet that needs to account for PVC expansion
	recreations, err := recreateStatefulSets(d.K8sClient(), d.ES)
	if err != nil {
		return results.WithError(fmt.Errorf("StatefulSet recreation: %w", err))
	}
	if recreations > 0 {
		// Some StatefulSets are in the process of being recreated to handle PVC expansion:
		// it is safer to requeue until the re-creation is done.
		// Otherwise, some operation could be performed with wrong assumptions:
		// the sset doesn't exist (was just deleted), but the Pods do actually exist.
		log.V(1).Info("StatefulSets recreation in progress, re-queueing.",
			"namespace", d.ES.Namespace, "es_name", d.ES.Name, "recreations", recreations)
		return results.WithResult(defaultRequeue)
	}
```

pkg/controller/elasticsearch/driver/pvc_expansion.go
```diff
@@ -0,0 +1,251 @@
...
+			if err := recreateStatefulSet(k8sClient, toRecreate); err != nil {
+				return recreations, err
+			}
+
+		// already recreated (existing.UID != toRecreate.UID): we're done
+		default:
+			// remove the annotation
+			delete(es.Annotations, annotation)
+			if err := k8sClient.Update(&es); err != nil {
+				return recreations, err
```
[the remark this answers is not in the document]
sebgl: Hm you're right, I think it's not necessary 👍 
Let's do that.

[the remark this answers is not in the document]
sebgl:
```go
// already recreated (existing.UID != toRecreate.UID): we're done
		default:
			// remove the annotation
			delete(es.Annotations, annotation)
			if err := k8sClient.Update(&es); err != nil {
				return recreations, err
			}
			// one less recreation
			recreations--
```

pkg/controller/elasticsearch/driver/upscale.go
sebgl: Good catch, you're right! We don't need that complicated logic anymore, which is great.

pkg/controller/elasticsearch/driver/upscale.go
```diff
@@ -104,22 +124,39 @@ func adjustZenConfig(k8sClient k8s.Client, es esv1.Elasticsearch, resources node
...
+		pods, err := sset.GetActualPodsForStatefulSet(k8sClient, k8s.ExtractNamespacedName(&expected))
+		if err != nil {
+			return appsv1.StatefulSet{}, err
+		}
+		actualReplicas = int32(len(pods))
+		if actualReplicas > 0 {
+			log.Info("Adjusting StatefulSet replicas based on orphan Pods from volume expansion",
+				"namespace", expected.Namespace, "statefulset_name", expected.Name, "replicas", actualReplicas)
+		}
+	}
```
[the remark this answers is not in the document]
sebgl: Not required anymore with the new annotation logic.

sebgl:
> Also I just want to be sure we don't want to handle the case where the Elasticsearch resource is deleted before the statefulset is re-created, leaving some Pods orphaned ?

@barkbay thanks for pointing that case again!
I tried setting an ownerRef to the ES resource in the podTemplate but according to my tests it looks like the StatefulSet controller applies its own ownerRef (only) and discards ours.
So instead I decided to manually add the Pod ownerRef right before deleting the StatefulSet, then remove it right after the StatefulSet was recreated (before removing the annotation).
It seems to work well and address the orphan Pods problem when the ES resource is deleted in the middle of a StatefulSet re-creation.

pkg/controller/elasticsearch/driver/pvc_expansion_test.go
```diff
@@ -0,0 +1,632 @@
...
+		if es.Annotations == nil {
+			es.Annotations = map[string]string{}
+		}
+		es.Annotations[key] = value
+		return es
+	}
+
+	sset1 := &appsv1.StatefulSet{ObjectMeta: metav1.ObjectMeta{Namespace: "ns", Name: "sset1", UID: "sset1-uid"}}
+	sset1Bytes, _ := json.Marshal(sset1)
+	sset1Json := string(sset1Bytes)
```
barkbay, line 351: cosmetic nit: I [think](https://github.com/golang/go/wiki/CodeReviewComments#initialisms) it is either `JSON` or `json`
```suggestion
	sset1JSON := string(sset1Bytes)
```

pkg/controller/elasticsearch/driver/pvc_expansion_test.go
```diff
@@ -0,0 +1,632 @@
...
+					"elasticsearch.k8s.elastic.co/recreate-sset1", sset1Json),
+					"elasticsearch.k8s.elastic.co/recreate-sset2", sset2Json),
+			},
+			wantES: *withAnnotation(withAnnotation(es(),
+				"elasticsearch.k8s.elastic.co/recreate-sset1", sset1Json),
+				"elasticsearch.k8s.elastic.co/recreate-sset2", sset2Json),
+			wantSsets:       nil,
+			wantPods:        []corev1.Pod{*pod1WithOwnerRef}, // ownerRef removed
+			wantRecreations: 2,
+		},
```
barkbay, line 433: Maybe add a test with an annotation which does not contain a serialized sset to test the `!strings.HasPrefix(key, RecreateStatefulSetAnnotationPrefix)` condition ?
```suggestion
		},
				{
			name: "multiple statefulsets to handle",
			args: args{
				runtimeObjs: []runtime.Object{sset1, sset2, pod1},
				es: *withAnnotation(withAnnotation(withAnnotation(es(),
					"elasticsearch.k8s.elastic.co/recreate-sset1", sset1Json),
					"elasticsearch.k8s.elastic.co/recreate-sset2", sset2Json),
					"foo", "bar"),
			},
			wantES: *withAnnotation(withAnnotation(withAnnotation(es(),
				"elasticsearch.k8s.elastic.co/recreate-sset1", sset1Json),
				"elasticsearch.k8s.elastic.co/recreate-sset2", sset2Json),
				"foo", "bar"),
			wantSsets:       nil,
			wantPods:        []corev1.Pod{*pod1WithOwnerRef}, // ownerRef removed
			wantRecreations: 2,
		},
```

pkg/controller/elasticsearch/driver/pvc_expansion_test.go
```diff
@@ -0,0 +1,632 @@
...
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			err := updatePodOwners(tt.args.k8sClient, tt.args.es, tt.args.statefulSet)
+			require.NoError(t, err)
+
+			var retrievedPods corev1.PodList
+			err = tt.args.k8sClient.List(&retrievedPods)
+			require.NoError(t, err)
```
barkbay, line 542: `wantPods` is never used, I think we want to compare it with `retrievedPods` ?

pkg/controller/elasticsearch/driver/pvc_expansion_test.go
```diff
@@ -0,0 +1,632 @@
...
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			err := removeESPodOwner(tt.args.k8sClient, tt.args.es, tt.args.statefulSet)
+			require.NoError(t, err)
+
+			var retrievedPods corev1.PodList
+			err = tt.args.k8sClient.List(&retrievedPods)
+			require.NoError(t, err)
```
barkbay, line 629: Same here, `wantPods` is not used.

pkg/controller/elasticsearch/driver/pvc_expansion_test.go
```diff
@@ -0,0 +1,632 @@
...
+		})
+	}
+}
+
+func withOwnerRef(pod corev1.Pod, ownerRef metav1.OwnerReference) *corev1.Pod {
```
barkbay, line 547: `ownerRef` is not used

barkbay (changes requested):
I left some comments re. the unit tests, I think they are not complete.

pkg/controller/elasticsearch/driver/pvc_expansion.go
```diff
@@ -0,0 +1,310 @@
...
+// is re-created), we also want the Pods to be deleted automatically.
+func updatePodOwners(k8sClient k8s.Client, es esv1.Elasticsearch, statefulSet appsv1.StatefulSet) error {
+	log.V(1).Info("Setting an owner ref to the Elasticsearch resource on the future orphan Pods",
+		"namespace", es.Namespace, "es_name", es.Name, "statefulset_name", statefulSet.Name)
+	return updatePods(k8sClient, statefulSet, func(p *corev1.Pod) error {
+		return controllerutil.SetOwnerReference(&es, p, scheme.Scheme)
+	})
+}
+
+// removeESPodOwner removes any reference to the ES resource from the Pods, that was set in updatePodOwners.
```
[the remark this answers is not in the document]
sebgl: I think it's a bit weird to have some Pods with an ES ownerRef only if they were subject to a volume resize, while the other Pods don't (also if the Pod is recreated the ownerRef disappears). I think it's a bit "cleaner" to keep this additional ownerRef temporary to only match the volume resize case? But indeed in practice it's not a big deal, the additional ownerRef should not hurt. Do you think we should not bother and keep the ownerRef?

pkg/controller/elasticsearch/driver/pvc_expansion_test.go
sebgl: Good idea.

sebgl: Actually I'll write a complete unit tests set for `ssetsToRecreate` that covers all the annotation cases.

pkg/controller/elasticsearch/driver/pvc_expansion_test.go
sebgl: 🤦 👍

pkg/controller/elasticsearch/driver/pvc_expansion.go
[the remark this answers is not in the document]
barkbay: No strong opinion at all, it was just for the sake of brevity of code and logic. I'm fine with your proposal which also makes sense. I think the real issue here is that the `ownerReferences` in the `podTemplate` are ignored by the controller.

Let's keep it as is and clean up the references 👍

pkg/controller/elasticsearch/driver/pvc_expansion_test.go
sebgl: 🤦 👍

pkg/controller/elasticsearch/driver/upscale_state_test.go
```diff
@@ -80,7 +80,7 @@ func Test_upscaleState_limitNodesCreation(t *testing.T) {
 		{
 			name:        "upscale masters from 3 to 4, but no creates allowed: should limit to 0",
 			state:       &upscaleState{allowMasterCreation: true, isBootstrapped: true, createsAllowed: pointer.Int32(0)},
-			actual:      sset.TestSset{Name: "sset", Replicas: 3, Master: true}.Build(),
+			actual:      sset.TestSset{Name: "sset", Replicas: 3, Master: false}.Build(),
```
barkbay, line 83: Do we need to change `Master` to `false` here ? Is it something left from a previous change ?

barkbay (approved):
LGTM 👍 
Great work !

pkg/controller/elasticsearch/driver/upscale_state_test.go
sebgl: :o indeed, no reason for that change!

david-kow (commented):
It worked well through all my attempts to break it :) nice work. Just one scenario I'm concerned about, the rest are nits.

pkg/apis/elasticsearch/v1/validations_test.go
```diff
@@ -326,7 +326,7 @@ func Test_pvcModified(t *testing.T) {
 		expectErrors bool
 	}{
 		{
-			name:    "resize fails",
+			name:    "resize succeeds",
```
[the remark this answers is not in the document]
sebgl: Yes! I have refactored that function and test for the needs of the validation webhook, and include the missing decrease there. I'll keep this PR as is and handle it in the other one.

pkg/controller/elasticsearch/driver/nodes.go
```diff
@@ -49,6 +49,21 @@ func (d *defaultDriver) reconcileNodeSpecs(
 		return results.WithResult(defaultRequeue)
 	}
 
+	// recreate any StatefulSet that needs to account for PVC expansion
+	recreations, err := recreateStatefulSets(d.K8sClient(), d.ES)
+	if err != nil {
+		return results.WithError(fmt.Errorf("StatefulSet recreation: %w", err))
+	}
+	if recreations > 0 {
```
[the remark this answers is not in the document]
sebgl: What we set in the annotation always matches an "actual StatefulSet", ie. a StatefulSet that exists in the apiserver, for which an update (or create) operation was successful in the past. We don't set the annotation based on the "expected StatefulSet".
I think in the case you mention where the StatefulSet cannot possibly be created, the update call would probably have failed beforehand? I don't know if there are some cases where a spec would lead to a successful update but an invalid creation. It seems weird.
WDYT?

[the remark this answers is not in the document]
david-kow: I was convinced that we set the expected sset in the annotation, which I see now isn't true. Yes, I think we can dismiss the case of successful update and failed create.

david-kow (approved):
LGTM! 🚀

PR #3753, opened 2020-09-15
Use Beats CRD for E2E monitoring

david-kow:
This PR changes deploying Filebeat/Metricbeat for our E2E runs to using Beat CRD. It also ups the version to `7.9.1`.

All RBAC resources stayed the same, `secureSettings` are used for credentials.

I'm marking this as a draft as there is one change I'd like to discuss. As we need to have an operator that "owns" those Beats I made the operator under test to manage an additional namespace where Beats are being deployed. I think it's less complicated than deploying a separate operator and the possible negative impact from reusing the operator is low. WDYT?

Logs from my tests available under `"e2e_test_id": "testpipeline-1-local-local-1.16-7.9.0"` - I'll delete them when this will be merged.


barkbay:
> I think it's less complicated than deploying a separate operator and the possible negative impact from reusing the operator is low. WDYT?

I think it depends on our expectations re. the quality of the logs/metrics uploaded in the monitoring cluster:
* If we accept to lose some information (or not having them at all) then it's fine to use the operator under test 
* If not it means that the beats must be managed by a version of the operator which is considered as already tested and stable.

I would be in favor of the second solution because investigating issues without knowing the quality of the logs is hard and can be confusing. But if it is really challenging we can also start with the first one and see how it goes.

config/e2e/monitoring.yaml
```diff
@@ -0,0 +1,711 @@
...
+        # This is to avoid a conflict with the ECS "error" and "source" documents.
+        co.elastic.logs/raw: "[{\"type\":\"container\",\"json.keys_under_root\":true,\"paths\":[\"/var/log/containers/*${data.kubernetes.container.id}.log\"],\"processors\":[{\"convert\":{\"mode\":\"rename\",\"ignore_missing\":true,\"fields\":[{\"from\":\"error\",\"to\":\"_error\"}]}},{\"convert\":{\"mode\":\"rename\",\"ignore_missing\":true,\"fields\":[{\"from\":\"_error\",\"to\":\"error.message\"}]}},{\"convert\":{\"mode\":\"rename\",\"ignore_missing\":true,\"fields\":[{\"from\":\"source\",\"to\":\"_source\"}]}},{\"convert\":{\"mode\":\"rename\",\"ignore_missing\":true,\"fields\":[{\"from\":\"_source\",\"to\":\"event.source\"}]}}]}]"
+        {{end}}
+      labels:
+        control-plane: monitoring
+        test-run: {{ .TestRun }}
+    spec:
+      serviceAccountName: monitoring-elastic-operator
+      containers:
+      - image: docker.elastic.co/eck-snapshots/eck-operator:1.3.0-SNAPSHOT-2020-09-23-199f4aab
```
david-kow, line 681: This is temporary as one of the fixes from `1.3` (modifying init containers) is needed here.

barkbay (commented):
It almost looks good to me :)
There is a small nit with the `stream-logs` label which must be set on the chaos job.
Also I don't get all the logs on my test cluster, I think the problem is mine but I'm investigating. I'll update this review asap.

barkbay (approved):
LGTM 🚀

PR #3754, opened 2020-09-15
Deployer: add options to create GKE private/net. policy enforced clusters

barkbay:
While investigating private cluster or network policy related issues I found myself not able to create corresponding clusters with the deployer

This PR adds 2 dedicated options, `private` and `networkPolicy` to the deployer in order to create private/net. policy enforced clusters:

```yaml
id: gke-dev
overrides:
  gke:
    gCloudProject: ...
    private: true
    networkPolicy: true
```

Options are both `false` by default. Current behaviour is preserved.

### DIFF
--- a/hack/deployer/runner/gke.go
+++ b/hack/deployer/runner/gke.go
@@ -112,7 +112,7 @@ func (d *GkeDriver) Execute() error {
 		if d.plan.Gke.Private {
 			log.Printf("a private cluster has been created, please retrieve credentials manually and create storage class and provider if needed")
 			log.Printf("to authorize a VM to access this cluster run the following command:\n"+
-				"$ cloud container clusters update %s"+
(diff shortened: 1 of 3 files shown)


hack/deployer/runner/gke.go
```diff
@@ -70,6 +70,7 @@ func (gdf *GkeDriverFactory) Create(plan Plan) (Driver, error) {
 			"KubernetesVersion": plan.KubernetesVersion,
 			"MachineType":       plan.MachineType,
 			"LocalSsdCount":     plan.Gke.LocalSsdCount,
+			"Private":           plan.Gke.Private,
```
david-kow, line 73: Is this needed if you are not using the value in any command?

david-kow (approved):
LGTM, one nit.

Issue #3762, opened 2020-09-16
Flaky E2E test: TestEnterpriseSearchVersionUpgradeToLatest7x


sebgl:
Happened on GKE:

```
=== RUN   TestEnterpriseSearchVersionUpgradeToLatest7x/Enterprise_Search_Pods_should_eventually_be_ready#01
Retries (5m0s timeout): ....................................................................................................
    utils.go:84: 
        	Error Trace:	utils.go:84
        	Error:      	Received unexpected error:
        	            	pod test-ent-version-upgrade-rvtz-ent-c9ffd75f4-zwd4g was not upgraded (yet?) to match the expected specification
        	Test:       	TestEnterpriseSearchVersionUpgradeToLatest7x/Enterprise_Search_Pods_should_eventually_be_ready#01
{"log.level":"error","@timestamp":"2020-09-16T02:18:14.457Z","message":"stopping early","service.version":"0.0.0-SNAPSHOT+00000000","service.type":"eck","ecs.version":"1.4.0","error":"test failure","error.stack_trace":"github.com/elastic/cloud-on-k8s/test/e2e/test.StepList.RunSequential\n\t/go/src/github.com/elastic/cloud-on-k8s/test/e2e/test/step.go:43\ngithub.com/elastic/cloud-on-k8s/test/e2e/test.RunMutations\n\t/go/src/github.com/elastic/cloud-on-k8s/test/e2e/test/run_mutation.go:38\ngithub.com/elastic/cloud-on-k8s/test/e2e/ent.TestEnterpriseSearchVersionUpgradeToLatest7x\n\t/go/src/github.com/elastic/cloud-on-k8s/test/e2e/ent/ent_test.go:71\ntesting.tRunner\n\t/usr/local/go/src/testing/testing.go:1127"}
    --- FAIL: TestEnterpriseSearchVersionUpgradeToLatest7x/Enterprise_Search_Pods_should_eventually_be_ready#01 (300.00s)
```

https://devops-ci.elastic.co/job/cloud-on-k8s-e2e-tests-stack-versions/202/testReport/

sebgl:
Looking at the diag bundle I can see Pod was successfully deleted a few seconds before the tests failed:

```
{
            "apiVersion": "v1",
            "count": 1,
            "eventTime": null,
            "firstTimestamp": "2020-09-16T02:17:39Z",
            "involvedObject": {
                "apiVersion": "apps/v1",
                "kind": "ReplicaSet",
[...]

PR #3763, opened 2020-09-16
Increase the timeout for Enterprise Search Pods to be ready in E2E tests

sebgl:
We seem to reach the 5 mins timeout for all Ent Search Pods to
be upgraded to a newer version:
https://github.com/elastic/cloud-on-k8s/issues/3762

This commit raises the timeout for that particular step to twice
the general timeout coming from the test context (eg. 5min for GKE ->
10 min for that particular step).

Fixes https://github.com/elastic/cloud-on-k8s/issues/3762


thbkrkr (approved):
Sounds reasonable.

barkbay (approved):
lgtm

Issue #3733, opened 2020-09-08
resources.requests.CPU limit Elasticsearch thread pools


pebrc:
`resources.requests.cpu` seem to be reported to the Pods as the available processors and therefore determine the sizing of the Elasticsearch [thread pools](https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-threadpool.html). This seems somewhat counter-intuitive especially if users do not set any limits. Without limits they might be under the impression that their pods could use more than the requested CPUs if enough resources are available on the k8s node but the due to the thread pool sizing this burstable capacity does not translate to better Elasticsearch performance. 

There are a few workarounds:
* request more resources to begin with
* set the `node.processors` property directly  to affect thread pool sizing (of course this can become a problem on a crowded k8s node where there are actually not more than `requests` CPUs available and Elasticsearch is running with oversized thread pools when overriding `node.processors`)

I am tagging this as a doc issue for now because I don't see currently what we could do about this other than stating the current behaviour in the docs.

anyasabo:
I was curious about how it detected it and so found this too: https://medium.com/@christopher.batey/cpu-considerations-for-java-applications-running-in-docker-and-kubernetes-7925865235b7

Agreed on this being a docs issue more than anything else

Imroath:
This is exactly what I needed... You can't even imagine how much time I spent investigating ES OOM issues on our ECK.
And it was all due to CPU not being able to do GC on time - because we had too little resources.requests.cpu, even though we had resources.limits.cpu quite high. We upgraded resources.requests.cpu yesterday - everything works like a charm since then.

PR #3765, opened 2020-09-17
Improve Elasticsearch CPU req doc

sebgl:
* Make it clear that CPU requests directly impacts Elasticsearch nodes.processors.
* Replace "should run ... at Guaranteed QoS level", by "can run ...", considering
the various performance problems with cpu limits throttling
* Modify the example to require 8 cpus instead of 0.5, which makes more sense
* Modify the example to not set cpu limits by default

Fixes https://github.com/elastic/cloud-on-k8s/issues/3733.


barkbay (approved):
lgtm

PR #3768, opened 2020-09-18
Create the e2e-default storage class with the static provisioner

sebgl:
Before this PR, we did not create the `e2e-default` storage class to
match PVCs created by the static provisioner as part of deploying the provisioner itself.

Instead, the `e2e-default` storage class was created as part of the e2e
tests bootstrap, by retrieving the default storage class (eg. gke standard),
then patching both the provisioner (to an empty value so gke provisioner
does not get triggered), and setting `volumeBindingMode: waitForFirstConsumer`.

This leads to the `e2e-default` storage class getting created as:

```yaml
allowVolumeExpansion: true
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: e2e-default
parameters:
  type: pd-standard
provisioner: kubernetes.io/no-provisioner
reclaimPolicy: Delete
volumeBindingMode: WaitForFirstConsumer
```

A few things do not make sense here:
- the parameters still match gke provisioner parameters
- `allowVolumeExpansion: true`, whereas it's not supported on the local provisioner

Instead of doing that, I think we can simply pre-create the e2e-default storage
class with the right parameters when deploying the provisioner. The e2e bootstrap
will detect that storage class is already there, and just use it.

If the static provisioner is not deployed, the e2e bootstrap will create an `e2e-default`
storage class based on the default storage class available.

Additionally, this PR removes the need for the `provisioner` argument in `createStorageClass`:
Either we rely on the pre-existing e2e-default storage class, either we create it based on the default storage class, which
already has the provisioner set.

### DIFF
--- a/hack/deployer/config/local-disks/ssd-provisioner.yaml
+++ b/hack/deployer/config/local-disks/ssd-provisioner.yaml
@@ -1,3 +1,10 @@
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: e2e-default
+provisioner: kubernetes.io/no-provisioner
+volumeBindingMode: WaitForFirstConsumer
+reclaimPolicy: Delete
 ---
 apiVersion: v1
 kind: ConfigMap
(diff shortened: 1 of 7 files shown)


barkbay (approved):
lgtm

PR #3708, opened 2020-09-03
Add a doc page on storage recommendations

sebgl:
This doc attempts to explain the trade-off between network-attached
and local volumes, and gives some details about storage class settings.

The main goal is for users to realize that there is no perfect solution,
but there are many interesting options:
- network-attached volumes, with easy operations, if performance is good enough
- local volumes, with more complicated operations, with either static provisioning
(mentioning the static provisioner as a good example), or dynamic provisioning
(mentioning TopoLVM as a good example)

I think we cannot reasonably mention all existing volume provisioners. There are many open source and commercial ones out there. For now I decided that the static volume provisioner and TopoLVM are 2 important ones worth looking at. We can revisit that decision in the future.

Fixes https://github.com/elastic/cloud-on-k8s/issues/2456.

PR #3772, opened 2020-09-21
Backport 1.2: Add a doc page on storage recommendations (#3708)

sebgl:
Backports on 1.2.

This doc attempts to explain the trade-off between network-attached
and local volumes, and gives some details about storage class settings.

The main goal is for users to realize that there is no perfect solution,
but there are many interesting options:
- network-attached, with easy operations, if performance is good enough
- local volume, with more complicated operations, with either static provisioning
(mentioning the static provisioner as a good example), or dynamic provisioning
(mentioning TopoLVM as a good example)


barkbay (approved):
lgtm

Issue #3598, opened 2020-08-10
API call controller 


anyasabo:
This issue will be updated with current status as it progresses.

### Goals

We would like to be able to declaratively configure (create/update) common configurations operations in Elasticsearch that are currently API-only, such as ILM, index settings, and index templates. This is helpful for organizations that have more than one Elasticsearch cluster, but want them to be configured identically. For instance, having separate dev, staging, and production environments, or having an environment per geography. It is helpful to be able to store these settings in configuration management to ensure all environments are identical (see “infrastructure as code” philosophy). Similar functionality exists for users and roles today.

Today to have a "fully functional" cluster, users need to create their ES cluster, wait for it to become available, then interact with the API to create the extra settings. If this could be handled by ECK it would make the process much simpler.

### Initial thoughts

We may want to have generic support for essentially any endpoint that supports a PUT. Users could provide a URL and the body, and we could get/set the URL as necessary. Even idempotent PUTs cause extra load on Elasticsearch, so it would be better for us to do a comparably cheaper GET first rather than blindly sending PUTs.

This resource could be modeled as a configmap, a CRD, or as part of the Elasticsearch manifest. My gut leans towards a CRD to start. Adding fields to a CRD that is already v1 is challenging because you cannot revise them easily. A configmap works well for a key/value list, which is all we really need. But it cannot be versioned easily, where at least CRDs we have the option to use webhook conversion (though we do not use the feature today and there are challenges with deploying it in all environments).

Caveat: some PUTs are not idempotent (e.g. [indexing documents](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-index_.html) revs the version). [SLM policies](https://www.elastic.co/guide/en/elasticsearch/reference/current/getting-started-snapshot-lifecycle-management.html) include execution status information. This may be something where we do not do a strict comparison before setting -- for instance, if a user specifies a doc to be indexed, as long as all the fields the user specified match, then we do not consider it as needing an update. This has similar problems that we have experienced in the past in that it makes it difficult to remove fields previously set by the user.

Another option would be to simply leave this as a potential footgun, and to document that users shouldn't use non-idempotent API operations.

Caveat cont'd: some operations have dependencies. For instance, a snapshot repository must be created before an SLM policy referencing it can be created. This can likely be supported by executing operations sequentially within a given resource.

### Proposed CRD example

```yaml
apiVersion: elasticsearchconfig.k8s.elastic.co/v1alpha1
[...]

PR #3775, opened 2020-09-22
Draft of ES declarative config

anyasabo:
Fix https://github.com/elastic/cloud-on-k8s/issues/3598

Opening up a draft to get feedback on the approach, still needs a lot of shoring up to do. But it "works". 

## An overview

- users provide us a list of URLs and an optional JSON body that URL should contain, and the elasticsearch resource it should apply to
- we do a get against that URL
  - if it's a 200 and the response body is a superset of the user-specified JSON body, we consider this operation successful and move on to the next operation in the list
  - if it is a 404 or a 200 but the response is not a superset of the user-specified body, we send a put to that URL with the user-specified body

## Example manual test

```yaml
---
kind: Elasticsearch
metadata:
  name: quickstart
spec:
  version: 7.9.1
  nodeSets:
  - name: default
    count: 1
    config:
      node.master: true
      node.data: true
      node.ingest: true
      node.store.allow_mmap: false
---
apiVersion: elasticsearchconfig.k8s.elastic.co/v1alpha1
kind: ElasticsearchConfig
metadata:
  name: quickstart-config
spec:
  operations:
  - url: "/_component_template/component_template1"
    body: |-
      {
        "template": {
          "mappings": {
            "properties": {
              "@timestamp": {
                "type": "date"
              }
            }
          }
        }
      }
  - url: "/_component_template/other_component_template"
    body: |-
      {
        "template": {
          "mappings": {
            "properties": {
              "ip_address": {
                "type": "ip"
              }
            }
          }
        }
      }
  elasticsearchRef:
    name: quickstart
```

```
PASSWORD=$(kubectl get secret quickstart-es-elastic-user -o go-template='{{.data.elastic | base64decode}}')

kubectl port-forward service/quickstart-es-http 9200

 curl -s -u "elastic:$PASSWORD" -k "https://localhost:9200/_component_template/component_template1" | jq

 curl -s -u "elastic:$PASSWORD" -k "https://localhost:9200/_component_template/other_component_template" | jq
```

## Open questions

- does the association interface make sense to use here? it has a meaning in other resources because the operator is configuring a resource (e.g. kibana) to create an association. here, the operator is making the calls itself, so it has a slightly different meaning. I cargo culted it in because I was used to it, but thinking about it more I'm not sure it's necessary

- ES sometimes supports additional arguments in PUTs that it does not support for GETs (e.g. [`verify=false`](https://www.elastic.co/guide/en/elasticsearch/reference/current/snapshots-register-repository.html#snapshots-repository-verification) for snapshot repo creation), but we provide no mechanism for that here. It's not clear to me how important that might be though.

- does checking specifically for 404s/200s make sense? I'm not sure if there's other status codes ES use that make sense for us to proceed on

- now is a good time for naming/structure feedback too


Makefile
```diff
@@ -171,6 +171,7 @@ go-run:
 			-tags "$(GO_TAGS)" \
 			./cmd/main.go manager \
 				--development \
+				--enable-leader-election=false \
```
anyasabo, line 174: This is just to enable quicker iteration when running locally and could be moved to a separate PR

sebgl:
I added a few comments about the general approach in the issue: https://github.com/elastic/cloud-on-k8s/issues/3598#issuecomment-702783723

pkg/controller/esconfig/esconfig_controller.go
```diff
@@ -0,0 +1,293 @@
...
+	getResp, _ := client.Request(ctx, get)
+
+	// nothing exists at this url yet, time to create it
+	if getResp.StatusCode == http.StatusNotFound {
+		logger.V(1).Info("resource does not exist yet", "url", opURL)
+		return true, nil
+	}
+
+	// TODO checking for a 200 might be wrong, really any status code not 200 might mean we need to update. but a 200 indicates we can and should compare the bodies
+	// TODO should bodies always be required to be specified? I think probably? I don't know of any ES API that you can send an empty PUT. maybe we require at a minimum "{}"?
```
pebrc, line 253: https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-create-data-stream.html

anyasabo:
I'll go ahead and close this out while we investigate an alternative approach (a single fire and forget job) that I'll detail in the issue

PR #3782, opened 2020-09-24
Configurable timeouts

### LINKED ISSUE #684: Make timeouts configurable
In the operator code we manipulate quite a few timeouts that could be configurable by the user:
- ES cluster timeout
- ~process manager timeout~
- ~process restart timeout~
- k8s apiserver timeout
- csr init container timeout

In most cases it makes sense to configure them at the operator level. Should some of them be configurable at the ES cluster spec level?

charith-elastic:
Makes timeouts configurable.

Elasticsearch timeouts are configurable for each cluster by annotating the associated `Elasticsearch` resource.
- `eck.k8s.elastic.co/es-client-timeout`: Elasticsearch client request timeout.
- ~`eck.k8s.elastic.co/es-voting-config-exclusion-timeout`: Timeout for voting config exclusion API calls.~
- `eck.k8s.elastic.co/es-observer-interval`: Elasticsearch observation interval.
- ~`eck.k8s.elastic.co/es-observer-request-timeout`: Elasticsearch observer request timeout.~

Kubernetes client timeout is configurable by setting the `kube-client-timeout` flag. I have chosen to make this a global timeout (affects controller-runtime as well) because the `WithTimeout` and `WithContext` methods provided by the wrapped `Client` API are not used anywhere to set custom request timeouts. I think that the whole `Client` abstraction can be removed from the codebase and replaced with the default controller-runtime `Client`, but, I did not do that in this PR because it would make the diff too large.

Fixes #684


cmd/manager/main.go
```diff
@@ -15,6 +15,25 @@ import (
...
+	"k8s.io/client-go/kubernetes"
+	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
+	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp" // allow gcp authentication
+	"k8s.io/client-go/rest"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/cache"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/manager/signals"
+
```
pebrc, line 36: nit: according to https://github.com/elastic/cloud-on-k8s/blob/master/CONTRIBUTING.md#user-content-format-your-code-and-manage-imports we only want two groups of imports

cmd/manager/main.go
```diff
@@ -192,6 +193,11 @@ func Command() *cobra.Command {
 		"",
 		"Set the IP family to use. Possible values: IPv4, IPv6, \"\" (= auto-detect) ",
 	)
+	cmd.Flags().Duration(
+		operator.KubeClientTimeout,
+		60*time.Second,
```
pebrc, line 198: Am I right that this default of 60 seconds changes the current behaviour for some usages (discovery has 32s apparently)? Don't think it is a big issue, though.

pkg/controller/common/annotation/timeout.go
```diff
@@ -0,0 +1,44 @@
...
+
+	metaAcc := meta.NewAccessor()
+
+	ann, err := metaAcc.Annotations(obj)
+	if err != nil {
+		log.V(1).Info("Failed to extract annotations from object", "error", err, "object", obj)
+		return defaultVal
+	}
+
+	if len(ann) == 0 {
```
pebrc, line 28: Just curious: what is the advantage of this explicit handling of an empty annotation map over just relying on the lookup?

pebrc (commented):
Looks good to me. I think I am OK with the annotation you propose if we assume that these timeout tweaks are meant as an escape hatch for exceptional cases. Are you planning to add docs in a separate PR?

pkg/controller/elasticsearch/client/client.go
```diff
@@ -9,19 +9,28 @@ import (
...
+
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common"
+	"github.com/elastic/cloud-on-k8s/pkg/controller/common/annotation"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common/version"
 	"github.com/elastic/cloud-on-k8s/pkg/utils/net"
 )
 
 const (
-	// DefaultVotingConfigExclusionsTimeout is the default timeout for setting voting exclusions.
-	DefaultVotingConfigExclusionsTimeout = "30s"
```
sebgl, line 22: Somewhat unrelated to that PR: I could not remember what that timeout was for so I double-checked the code.
It is used as a query param in the voting config exclusions request: https://www.elastic.co/guide/en/elasticsearch/reference/master/voting-config-exclusions.html#voting-config-exclusions-api-query-params.
If the timeout is reached, the request fails and returns an error.

Other API calls, such as `GET _cluster/settings` (https://www.elastic.co/guide/en/elasticsearch/reference/master/cluster-get-settings.html) also have this timeout query param, but we don't use it (we should!).

For the sake of simplicity (one less thing to configure), I feel like we could get rid of that `DefaultVotingConfigExclusionsTimeout ` and use the `defaultClientTimeout` instead (which moves the timeout to 3min instead of the default 30sec). Since we'd also have a 3min timeout to the http call, we may reach it before Elasticsearch reaches its own 3min timeout to set the voting config exclusions. Which is not a big deal, I think: this request is idempotent and we do it at every reconciliation anyway.

To summarize:
- keep a single Elasticsearch http timeout for all api calls
- ensure we set the `?timeout=` query param to that timeout for URLs that support it

@elastic/cloud-k8s any thoughts?

pkg/controller/elasticsearch/client/shard.go
```diff
@@ -21,7 +21,7 @@ type ShardLister interface {
 }
 
 func (c *clientV6) ExcludeFromShardAllocation(ctx context.Context, nodes string) error {
-	ctx, cancel := context.WithTimeout(ctx, DefaultReqTimeout)
+	ctx, cancel := context.WithTimeout(ctx, defaultClientTimeout)
```
sebgl, line 24: do we need to set the timeout in the context if the http client already has a timeout set internally?

pkg/controller/elasticsearch/driver/esstate.go
```diff
@@ -62,9 +62,7 @@ type memoizingNodes struct {
 
 // initialize requests Elasticsearch for nodes information, only once.
 func (n *memoizingNodes) initialize() error {
-	ctx, cancel := context.WithTimeout(n.ctx, esclient.DefaultReqTimeout)
-	defer cancel()
-	nodes, err := n.esClient.GetNodes(ctx)
+	nodes, err := n.esClient.GetNodes(context.Background())
```
sebgl, line 65: Why not reusing `n.ctx` here instead of creating a new context? Could be useful if later on we want to limit the total duration of a whole chain of calls.

pkg/utils/k8s/client.go
```diff
@@ -29,9 +24,6 @@ type Client interface {
 	// WithContext returns a client configured to use the provided context on
 	// subsequent requests, instead of one created from the preconfigured timeout.
```
sebgl, line 25: we should probably fix that comment since the default context does not have any timeout set

cmd/manager/main.go
```diff
@@ -233,11 +244,6 @@ func Command() *cobra.Command {
 		DefaultWebhookName,
 		"Name of the Kubernetes ValidatingWebhookConfiguration resource. Only used when enable-webhook is true.",
 	)
```
sebgl, line 246: I was also expecting a flag to control the default timeout for ES API calls (which can also be overridden per cluster).

pkg/controller/elasticsearch/client/client.go
pebrc: I am in favour of your proposal. I have to admit I did not question the need for this timeout in the first place.

cmd/manager/main.go
charith-elastic: That slipped my mind completely. Is there a particular reason for that though? Personally, I find that a separate group for "project imports" makes things a little bit easier to read -- especially when there are lots of imports like we do in `main.go`.

pebrc: I am not sure how the two-groups rule came to be. Maybe @barkbay recalls I believe he wrote the guide.

barkbay: Sorry, the rule was already existing when I joined, I don't know where it comes from. Imports are sorted alphabetically so I'm not sure there are strong arguments to use either 2 or 3 groups _(maybe when there is an alias ?)_. I think the code is already organized with 2 groups so that would an argument to stick with it.

If we want to move to 3 groups my understanding is that using `goimports` can automatically _(with some IDE integration)_ do that for us with `goimports --local "github.com/elastic/cloud-on-k8s"`

cmd/manager/main.go
charith-elastic: Going by the [comment](https://github.com/kubernetes/client-go/blob/becbabb360023e1825a48b4db85f454e452ae249/discovery/discovery_client.go#L48-L49) in `DiscoveryClient` it looks like the 32s timeout is an arbitrary value used to distinguish that particular instance from among many different timeouts. It is only used when the REST client doesn't have its own timeout set (which we are setting now with this change). I don't think it will have a serious effect, except maybe taking twice as long to detect a problem with discovery. 

Of course, there may be other unforeseen side effects from changing the global timeout. I am not too worried about that because we just need to set `--kube-client-timeout=0` to get back to the old behaviour.

pkg/controller/common/annotation/timeout.go
charith-elastic: The main purpose is to check that the map is not nil. It has the added bonus of not requiring a lookup when the map is empty as well.

pkg/controller/elasticsearch/client/client.go
charith-elastic: Currently we don't use the `timeout` override in the API call at all (we just default to the Elasticsearch default) so I have opted to remove that parameter completely from the function for now. If we find a reason to use an explicit timeout at  a later date, we can easily add that back in.

cmd/manager/main.go
charith-elastic: Yes, I refreshed my setup with my personal configuration and accidentally pulled in the `goimports` config that automatically formatted my PRs in the new way. I have reverted it. 

I think we can easily run `goimports` on the whole code base to use a new convention if we wanted to. However, this is just my personal preference, and it's not fair to ask everybody to change their settings. So let's keep things as they are.

pkg/controller/common/annotation/timeout.go
pebrc: I think lookups work even on `nil` maps afaik.

charith-elastic: Yes, they do. Interestingly, the nil check seems to make things faster -- which kind of makes sense because the length is a type attribute that can be checked efficiently.  It doesn't matter much in this context though.

```
BenchmarkNil/direct-8          	456756687	        2.56 ns/op	      0 B/op	      0 allocs/op
BenchmarkNil/nilcheck-8        	1000000000	        0.662 ns/op	      0 B/op	      0 allocs/op
```

pebrc (approved):
LGTM

deploy/eck-operator/values.yaml
```diff
@@ -140,6 +140,9 @@ config:
   # setDefaultSecurityContext determines whether a default security context is set on application containers created by the operator.
   setDefaultSecurityContext: true
 
+  # kubeClientTimeout sets the request timeout for Kubernetes API calls made by the operator.
+  kubeClientTimeout: 60s
+
```
sebgl, line 145: should the default es client timeout also appear here?

cmd/manager/main.go
```diff
@@ -389,6 +400,12 @@ func startOperator(stopChan <-chan struct{}) error {
 		return err
 	}
 
+	// set the timeout for API client
+	cfg.Timeout = viper.GetDuration(operator.KubeClientTimeout)
+
+	// set the timeout for Elasticsearch requests
+	esclient.DefaultESClientTimeout = viper.GetDuration(operator.ElasticsearchClientTimeout)
```
sebgl, line 407: For consistency the default observer timeout & interval could also be configurable through flags?

charith-elastic: I feel like that'll just clutter the already long list of configuration options. Under normal circumstances, observer timeout and frequency are not things that users need to be concerned with. They only need to change those settings in exceptional circumstances, so leaving those to annotations feels right to me. WDYT?

sebgl: I think I'm fine with that.
My gut feeling says the observer timeout should match the es client timeout though (whose value can be specified in the annotation). So basically the observer timeout setting could disappear, I think?
We can open a different issue to better refine those.

charith-elastic: I tried to stick to what we already had in the code without changing the behaviour too much. The observer is sort of a health check if you squint hard enough :smile: So, it's probably fine to be more aggressive than normal. Of course, we can discuss it later in a different issue if you feel differently.

pkg/controller/elasticsearch/observer/manager.go
```diff
@@ -7,47 +7,61 @@ package observer
...
 		lock:      sync.RWMutex{},
-		settings:  settings,
+		tracer:    tracer,
 	}
 }
 
 // ObservedStateResolver returns the last known state of the given cluster,
 // as expected by the main reconciliation driver
-func (m *Manager) ObservedStateResolver(cluster types.NamespacedName, esClient client.Client) State {
+func (m *Manager) ObservedStateResolver(cluster *esv1.Elasticsearch, esClient client.Client) State {
```
sebgl, line 44: nit: do we need a pointer here? Not passing pointers would give a better hint that this cluster object won't be mutated (even though technically you could also mutate arrays of a non-pointer but well...). So I tend to prefer not using pointers unless there are strictly required.
Same for `Observe()` and `extractObserverSettings()`.

pkg/controller/common/annotation/timeout.go
```diff
@@ -0,0 +1,44 @@
...
+		return defaultVal
+	}
+
+	metaAcc := meta.NewAccessor()
+
+	ann, err := metaAcc.Annotations(obj)
+	if err != nil {
+		log.V(1).Info("Failed to extract annotations from object", "error", err, "object", obj)
+		return defaultVal
+	}
```
sebgl, line 26: we could avoid all the meta accessor stuff if the function was passed a `metav1.ObjectMeta` instead of a `runtime.Object`?

pkg/controller/elasticsearch/observer/manager.go
```diff
@@ -118,3 +132,20 @@ func (m *Manager) notifyListeners(cluster types.NamespacedName, previousState St
...
+		ObservationInterval: annotation.ExtractTimeout(cluster, ObserverIntervalAnnotation, defaultObservationInterval),
+		RequestTimeout:      annotation.ExtractTimeout(cluster, ObserverRequestTimeoutAnnotation, defaultRequestTimeout),
+		Tracer:              m.tracer,
+	}
+
+	clientTimeout := client.Timeout(cluster)
+	if settings.RequestTimeout > clientTimeout {
+		log.Info("Ignoring observer request timeout annotation because it is larger than the client request timeout", "namespace", cluster.Namespace, "es_name", cluster.Name)
+		settings.RequestTimeout = clientTimeout
+	}
```
sebgl, line 148: I would not expect this. If I set a specific observer timeout I want it to be used whatever the reconciliation es client timeout is?

sebgl (commented):
Overall LGTM, but I'm not sure about the logic for not overriding the observer timeout if larger than the client timeout used for reconciliations.

pkg/controller/elasticsearch/observer/manager.go
charith-elastic: Observers get an already constructed Elasticsearch client passed to them -- which makes it impossible to change the timeout in the Observer.  Of course, we can create two different constructor functions for the Elasticsearch client to get over that problem. But, I didn't think that was necessary because what is a legitimate reason for having an Observer request timeout that is longer than the standard Elasticsearch client timeout? They are fetching the health information from the cluster and if that doesn't return in a sensible amount of time, then the chances of other API calls succeeding in a shorter time are pretty low as well.

pkg/controller/elasticsearch/observer/manager.go
charith-elastic: I wanted to avoid a potentially large copy of the Elasticsearch object; but, you're right, it probably doesn't matter here.

pkg/controller/common/annotation/timeout.go
charith-elastic: Good point.

pkg/controller/elasticsearch/observer/manager.go
sebgl: Ah - looking at the code I understand technically that's because we use a timeout in the http client, but also an additional context for the observer http calls? 
Having timeouts done one way in the reconciliation (http client timeout) vs. another way in the observer (through a context) has a bit of a smell to me.
Also while I'm not opposed to the argument that you generally want the observer timeout to be shorter than the reconciliation es timeout, it feels a bit like we're imposing a constraint due to the way the code is organized internally.
How about we refactor slightly the observer code to set the timeout on the http client instead of using a context? If needed I think we could add an additional `SetTimeout(timeout time.Duration)` function to the Elasticsearch Client interface so we can still pass the esClient around, and override its timeout at the observer level? But it looks like we could also pass the right timeout at the observer es client creation time?

charith-elastic: To summarise an offline conversation I had with @sebgl: we both agreed that the special timeout for Observer requests doesn't make much sense and could just be removed.

sebgl (approved):
LGTM

PR #3784, opened 2020-09-28
Add CRD usage reporting

david-kow:
This PR adds CRD usage reporting.

On an interval (currently 1h), we will scan all managed namespaces for all ECK resources (es, kb, apm, beats, ent) and update all Kibana config secrets with our findings under `telemetry.yml` key:
```
$ k view-secret kibana-kb-config    telemetry.yml
eck:
  build:
    date: "2019-09-20T07:00:00Z"
    hash: dkowalski_latest
    snapshot: "true"
    version: 1.1.0
  custom_operator_namespace: false
  distribution: v1.16.13-gke.1
  operator_uuid: 15039433-f873-41bd-b6e7-10ee3665cafa
  stats:
    apms:
      pod_count: 0
      resource_count: 0
    beats:
      auditbeat_count: 0
      filebeat_count: 0
      heartbeat_count: 0
      journalbeat_count: 0
      metricbeat_count: 0
      packetbeat_count: 0
      pod_count: 0
      resource_count: 0
    elasticsearches:
      pod_count: 3
      resource_count: 1
    enterprisesearches:
      pod_count: 0
      resource_count: 0
    kibanas:
      pod_count: 1
      resource_count: 1
```

Draft as tests are not yet added.

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -170,6 +170,11 @@ func Command() *cobra.Command {
 		"Disable watching the configuration file for changes",
 	)
 	cmd.Flags().Bool(
+		operator.DisableTelemetryFlag,
+		false,
+		"Disable periodically updating ECK telemetry data for Kibana to consume.",
+	)
+	cmd.Flags().Bool(
 		operator.EnforceRBACOnRefsFlag,
 		false, // Set to false for backward compatibility
 		"Restrict cross-namespace resource association through RBAC (eg. referencing Elasticsearch from Kibana)",
@@ -508,7 +513,8 @@ func startOperator(stopChan <-chan struct{}) error {
 		return err
 	}
 
-	go asyncTasks(mgr, cfg, managedNamespaces, operatorNamespace, operatorInfo)
+	disableTelemetry := viper.GetBool(operator.DisableTelemetryFlag)
+	go asyncTasks(mgr, cfg, managedNamespaces, operatorNamespace, operatorInfo, disableTelemetry)
 
 	log.Info("Starting the manager", "uuid", operatorInfo.OperatorUUID,
 		"namespace", operatorNamespace, "version", operatorInfo.BuildInfo.Version,
@@ -524,7 +530,13 @@ func startOperator(stopChan <-chan struct{}) error {
 }
 
 // asyncTasks schedules some tasks to be started when this instance of the operator is elected
-func asyncTasks(mgr manager.Manager, cfg *rest.Config, managedNamespaces []string, operatorNamespace string, operatorInfo about.OperatorInfo) {
+func asyncTasks(
+	mgr manager.Manager,
+	cfg *rest.Config,
+	managedNamespaces []string,
+	operatorNamespace string,
+	operatorInfo about.OperatorInfo,
+	disableTelemetry bool) {
 	<-mgr.Elected() // wait for this operator instance to be elected
 
 	// Report this instance as elected through Prometheus
@@ -539,11 +551,13 @@ func asyncTasks(mgr manager.Manager, cfg *rest.Config, managedNamespaces []strin
 		r.Start(licensing.ResourceReporterFrequency)
 	}()
 
-	// Start the telemetry reporter
-	go func() {
-		tr := telemetry.NewReporter(operatorInfo, mgr.GetClient(), managedNamespaces)
-		tr.Start()
-	}()
+	if !disableTelemetry {
+		// Start the telemetry reporter
+		go func() {
+			tr := telemetry.NewReporter(operatorInfo, mgr.GetClient(), managedNamespaces)
+			tr.Start()
+		}()
+	}
 
 	// Garbage collect any orphaned user Secrets leftover from deleted resources while the operator was not running.
 	garbageCollectUsers(cfg, managedNamespaces)
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -530,17 +530,17 @@ func asyncTasks(mgr manager.Manager, cfg *rest.Config, managedNamespaces []strin
 	// Report this instance as elected through Prometheus
 	metrics.Leader.WithLabelValues(string(operatorInfo.OperatorUUID), operatorNamespace).Set(1)
 
+	time.Sleep(10 * time.Second)         // wait some arbitrary time for the manager to start
+	mgr.GetCache().WaitForCacheSync(nil) // wait until k8s client cache is initialized
+
 	// Start the resource reporter
 	go func() {
-		time.Sleep(10 * time.Second)         // wait some arbitrary time for the manager to start
-		mgr.GetCache().WaitForCacheSync(nil) // wait until k8s client cache is initialized
 		r := licensing.NewResourceReporter(mgr.GetClient(), operatorNamespace)
 		r.Start(licensing.ResourceReporterFrequency)
 	}()
 
 	// Start the telemetry reporter
 	go func() {
-		mgr.GetCache().WaitForCacheSync(nil) // wait until k8s client cache is initialized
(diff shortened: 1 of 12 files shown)


cmd/manager/main.go
```diff
@@ -537,6 +538,13 @@ func asyncTasks(mgr manager.Manager, cfg *rest.Config, managedNamespaces []strin
 		r.Start(licensing.ResourceReporterFrequency)
 	}()
 
+	// Start the telemetry reporter
+	go func() {
+		mgr.GetCache().WaitForCacheSync(nil) // wait until k8s client cache is initialized
```
pebrc, line 543: Curious if we could do the cache sync in front of both go blocks (telemetry and resource reporter). Also, even if not introduced here, I wonder why we are ignoring the return value of `WaitForCacheSync`. It returns `false` if it was not able to sync the cache.

pkg/telemetry/telemetry.go
```diff
@@ -0,0 +1,208 @@
...
+	for range ticker.C {
+		usage, err := r.getResourceUsage()
+		if err != nil {
+			log.Error(err, "failed to get resource usage")
+			continue
+		}
+
+		telemetryBytes, err := yaml.Marshal(ECK{ECK: r.operatorInfo, Usage: usage})
+		if err != nil {
+			log.Error(err, "failed to marshal telemetry data")
```
pebrc, line 72: Why are we not `continue`ing here?

pkg/telemetry/telemetry.go
```diff
@@ -0,0 +1,208 @@
...
+			}
+			for _, kb := range kibanaList.Items {
+				var secret corev1.Secret
+				nsName := types.NamespacedName{Namespace: kb.Namespace, Name: kibana.SecretName(kb)}
+				if err := r.client.Get(nsName, &secret); err != nil {
+					log.Error(err, "failed to get kibana secret")
+					continue
+				}
+
+				secret.Data[kibana.TelemetryFilename] = telemetryBytes
```
pebrc, line 89: Can we be smart here and only update if there is a change in data. I am bit worried about the number of updates we will be pushing to the API server in larger ECK operator installations with many Kibanas.

pkg/telemetry/telemetry.go
```diff
@@ -0,0 +1,208 @@
...
+)
+
+var log = logf.Log.WithName("usage")
+
+type getUsageFn func(k8s.Client, []string) (string, interface{}, error)
+
+// ECK is a helper struct to marshal telemetry information.
+type ECK struct {
+	ECK   about.OperatorInfo     `json:"eck"`
+	Usage map[string]interface{} `json:"eck_usage"`
```
pebrc, line 39: Should this struct be called `Telemetry` instead? Should `Usage` be `Stats` (a term which seems be used in ES and other stack applications)? Should we have a hierarchy like so: 
`eck.build.date: "blah"` and `eck.stats.apms.pod_count: 1`

pkg/telemetry/telemetry.go
```diff
@@ -0,0 +1,208 @@
...
+			usage[typeToName(beat.Spec.Type)]++
+			usage[podCount] += beat.Status.AvailableNodes
+		}
+	}
+
+	return "beats", usage, nil
+}
+
+func entUsage(k8sClient k8s.Client, managedNamespaces []string) (string, interface{}, error) {
+	usage := map[string]int32{resourceCount: 0, podCount: 0}
```
pebrc, line 194: I was trying to think if we could have a common type for all of these maps so that we can avoid the `interface{}` but could not think of something good rn.

pebrc (commented):
This looks good overall. I am little bit worried about the impact of re-writing every Kibana secret every 10s especially in larger installations. Not sure if we can improve. Also I don't like that two processes are writing to the same secret. I don't think anything bad will happen because we are relying on the resource version on updates via the reconciler utility but it has a bit of a smell to it. 

So I think we should try to minimise the updates on the secrets. I am thinking that the usage stats and eck telemetry is actually the same for all Kibanas so we would need only one additional secret for each namespace that has a Kibana that all of these Kibana pods could mount. If this seems to complicated or has other drawbacks, I would at least try to be clever about the updates we do.

cmd/manager/main.go
david-kow: I left just one above, but I'm not sure what is the proper behavior if it keeps returning `false`. Crash the operator?

pkg/telemetry/telemetry.go
david-kow: I felt this is fairly isolated and allows nesting that might be useful.

pkg/telemetry/telemetry.go
david-kow: Updated issue description.

david-kow:
Thanks for comments @pebrc. I'm not that worried about using the same Secret. The keys being updated are different which already separates concerns a bit. I've changed to updating trough `ReconcileSecret` which should reduce the number of actual updates greatly. Also, 10 seconds was leftover from testing, I've intended this to be on hour scale. IIUC, Kibana only sends this data up every 24h, so we could relax this even more if we want to.

cmd/manager/main.go
pebrc: Good point and probably the explanation why the code is as it is right now.

pkg/telemetry/telemetry.go
```diff
@@ -0,0 +1,231 @@
...
+}
+
+func marshalTelemetry(info about.OperatorInfo, stats map[string]interface{}) ([]byte, error) {
+	return yaml.Marshal(struct {
+		ECK struct {
+			about.OperatorInfo
+			Stats map[string]interface{} `json:"stats"`
+		} `json:"eck"`
+	}{
+		ECK: struct {
```
pebrc, line 70: anonymous structs are a bit verbose, but I guess your rationale is that it is used only once?

pebrc (commented):
LGTM. Just one concern about the adverse effects of updating potentially many secrets in a loop. Maybe someone else from @elastic/cloud-k8s can chime in? Could be that I am a bit paranoid.

pkg/telemetry/telemetry.go
david-kow: Yes. We can create types if we use it in more than a single place.

pkg/telemetry/telemetry.go
```diff
@@ -0,0 +1,231 @@
...
+			}
+
+			if secret.Data == nil {
+				// should not happen, but just to be safe
+				secret.Data = make(map[string][]byte)
+			}
+
+			secret.Data[kibana.TelemetryFilename] = telemetryBytes
+
+			if _, err := reconciler.ReconcileSecret(r.client, secret, nil); err != nil {
```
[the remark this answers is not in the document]
david-kow: I don't think I'm that worried. Updating hundreds would require hundreds of Kibanas deployed, which I don't think is a common case and if it is, it should just work. Nevertheless, I've added flag to completely disable this code.

pkg/telemetry/telemetry_test.go
```diff
@@ -0,0 +1,236 @@
...
+	"github.com/elastic/cloud-on-k8s/pkg/controller/kibana"
+	"github.com/elastic/cloud-on-k8s/pkg/utils/k8s"
+	"github.com/stretchr/testify/require"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var testOperatorInfo = about.OperatorInfo{
+	OperatorUUID:            "15039433-f873-41bd-b6e7-10ee3665cafa",
+	CustomOperatorNamespace: true,
```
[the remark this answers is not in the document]
david-kow: I don't have a strong opinion either way. I'll leave it in this PR, but I'm open to removing it separately, if that's ok.

pebrc (approved):
LGTM!

PR #3787, opened 2020-09-30
Add name label to namespaces

david-kow:
Fixes recent failures of `TestWebhook` tests. Only the operator namespace was labelled with its name, so the webhook was not choosing test namespaces to run for correctly.


barkbay (approved):
lgtm

PR #3791, opened 2020-10-01
Add external IP addresses to certificate

### LINKED ISSUE #910: Automatically inject service external IP to the certs SAN
When users expose the ES cluster through `http.service.type`, they're hitting a cert validity issue while requesting the cluster:

```
Failed to connect to backoff(elasticsearch(https://34.66.153.170:9200)): Get https://34.66.153.170:9200: x509: certificate is valid for 10.60.1.15, 127.0.0.1, 10.63.247.8, not 34.66.153.179
```

It's expected, since the external IP is never injected into ES certs SANs.

The workaround currently is for users to manually inject that IP into the Elasticsearch Spec:
https://github.com/elastic/cloud-on-k8s/blob/5a8a8a340b19403d75659595fcf96ad7c95a167f/operators/config/samples/elasticsearch/elasticsearch.yaml#L46-L51

We should be able to inject that IP automatically, and reconcile on any IP change.
Do we need a way for this to be an opt-in/opt-out feature, or is it OK to just enforce it?

charith-elastic:
Adds the external IP address and/or hostname of the service to the self-signed certificate.

Fixes #910


pkg/controller/common/certificates/http_reconcile.go
```diff
@@ -14,17 +14,18 @@ import (
...
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes/scheme"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+
+	commonv1 "github.com/elastic/cloud-on-k8s/pkg/apis/common/v1"
+	"github.com/elastic/cloud-on-k8s/pkg/controller/common/name"
+	"github.com/elastic/cloud-on-k8s/pkg/controller/common/reconciler"
+	"github.com/elastic/cloud-on-k8s/pkg/utils/k8s"
+	netutil "github.com/elastic/cloud-on-k8s/pkg/utils/net"
```
pebrc, line 28: Nit: as mentioned elsewhere, I believe we have "two import blocks" convention.

pebrc (approved):
Looks good and worked in my test 👍

pkg/utils/k8s/k8sutils.go
```diff
@@ -63,10 +66,41 @@ func PodNames(pods []corev1.Pod) []string {
 
 // GetServiceDNSName returns the fully qualified DNS name for a service
```
sebgl, line 67: Let's maybe update the comment to reflect the ingress hostname thing?

sebgl (approved):
LGTM

Issue #3612, opened 2020-08-12
Improve how to bump the Elastic Stack version


thbkrkr:
Updating the Elastic Stack version is a manual task prone to omissions. We have a script `hack/update-samples-stack-version.sh` that handles updating the Elastic Stack version only in the samples. Since this script was created, the Elastic Stack version is present in several other places. 

It would be good to improve the existing script to handle updating all Elastic Stack versions present in the repository. We may need to change the current approach and consider adding the old version as a parameter.
The current way is to match `version: [0-9]\.[0-9]\.[0-9]` or `version: "[0-9]\.[0-9]\.[0-9]"` and it does not work for example for `hack/upgrade-test-harness/conf.yaml` which has other old stack versions that should not be updated.

Here are the versions to update:

```
Makefile:368:STACK_VERSION          ?= 7.8.1

test/e2e/test/version.go:19:	LatestVersion7x = "7.8.1" // version to synchronize with the latest release of the Elastic Stack

test/e2e/kb/testdata/kibana_standalone.yaml:7:  version: 7.8.1
test/e2e/kb/testdata/kibana_standalone.yaml:21:  version: 7.8.1

config/e2e/filebeat.yaml:91:          image: docker.elastic.co/beats/filebeat:7.8.1
config/e2e/metricbeat.yaml:134:          image: docker.elastic.co/beats/metricbeat:7.8.1

config/samples/elasticsearch/elasticsearch.yaml:7:  version: 7.8.1
config/samples/apm/apmserver.yaml:6:  version: 7.8.1
config/samples/enterprisesearch/ent_es.yaml:7:  version: 7.8.1
config/samples/enterprisesearch/ent_es.yaml:21:  version: 7.8.1
config/samples/kibana/kibana_es.yaml:7:  version: 7.8.1
config/samples/kibana/kibana_es.yaml:21:  version: 7.8.1
config/samples/apm/apm_es_kibana.yaml:8:  version: 7.8.1
config/samples/apm/apm_es_kibana.yaml:22:  version: 7.8.1
config/samples/apm/apm_es_kibana.yaml:32:  version: 7.8.1

config/recipes/associations-rbac/apm_es_kibana_rbac.yaml:87:  version: 7.8.1
config/recipes/associations-rbac/apm_es_kibana_rbac.yaml:100:  version: 7.8.1
config/recipes/associations-rbac/apm_es_kibana_rbac.yaml:114:  version: 7.8.1
[...]

PR #3793, opened 2020-10-01
Update stack to 7.9.2

anyasabo:
Also updates the `hack/update-samples-stack-version.sh` script to cover a few more files. We could also extend this to cover `hack/operatorhub/config.yaml` but I opted to just leave a note about it for now


david-kow (approved):

thbkrkr:
We've missed some versions in the [latest](https://github.com/elastic/cloud-on-k8s/pull/3642) [updates](https://github.com/elastic/cloud-on-k8s/pull/3610). This stack version update is definitively very prone to oversights.

To be complete:
```sh
> ag '7.[89].[01]' --ignore pkg --ignore go.sum --ignore hack/release-notes --ignore docs --ignore "*_test.go" --ignore NOTICE.txt --ignore stream.json --ignore builder.go

test/e2e/test/version.go
19:     LatestVersion7x = "7.9.0" // version to synchronize with the latest release of the Elastic Stack

test/e2e/kb/testdata/kibana_standalone.yaml
7:  version: 7.8.1
21:  version: 7.8.1

config/recipes/beats/auditbeat_hosts.yaml
79:        #  image: docker.elastic.co/beats/auditbeat:7.8.1

Makefile
377:E2E_STACK_VERSION          ?= 7.9.0

hack/upgrade-test-harness/conf.yaml
16:    stackVersion: 7.8.0
20:    stackVersion: 7.8.0

hack/upgrade-test-harness/testdata/upcoming/stack.yaml
7:  version: 7.8.1
19:  version: 7.8.1
29:  version: 7.8.1

hack/upgrade-test-harness/testdata/v112/stack.yaml
7:  version: 7.8.1
19:  version: 7.8.1
29:  version: 7.8.1
```

I created https://github.com/elastic/cloud-on-k8s/issues/3612 for that purpose. Maybe we should just be more declarative in the script and not just rely on a `find`. It's a small task but I feel like we waste a few minutes every month. The Elastic release train is going fast :]

anyasabo:
Good catch @thbkrkr . I'm not sure the `hack/upgrade-test-harness` is easily update-able and is only manually run anyway, so I left that out of this. I manually updated `Makefile`, `config/recipes/beats/auditbeat_hosts.yaml`, and `test/e2e/test/version.go`. I added `test/e2e` to the paths to search. To get those three I updated manually I think we would need to change our approach as you mentioned.

I also wonder if just outputting the `ag`/`grep` search at the end of the auto-update script would be a step forward to at least let people know maybe there's something that still needs attention.

thbkrkr (approved):

PR #3800, opened 2020-10-06
Add distribution-channel flag

david-kow:
This PR adds a new flag to allow the telemetry report how was the operator installed. There are three ways this can happen today: directly through `all-in-one.yaml` file, through helm and through one of the OLM based ways.

As we don't want to create a separate image for each of those, we can't embed this information there. Instead, we pass the distribution channel as a parameter to the operator. To set a good default for each channel the helm chart uses the value of `telemetry.distributionChannel` as an argument in the invocation of `manager`.
1. By default, that value is set to `helm`, so users using helm directly have this set for them.
2. When generating `all-in-one.yaml` file, we set it to `all-in-one` when we invoke `manifest-gen`. We want to avoid setting it in ConfigMap with operator config as this ends up in the image as `config/eck.yaml` file. 
3. `operatorhub` tool now allows to set the arg value used in csv.

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -176,6 +176,11 @@ func Command() *cobra.Command {
 		false,
 		"Disable periodically updating ECK telemetry data for Kibana to consume.",
 	)
+	cmd.Flags().String(
+		operator.DistributionChannelFlag,
+		"",
+		"Set the distribution channel to report through the telemetry.",
+	)
 	cmd.Flags().Bool(
 		operator.EnforceRBACOnRefsFlag,
 		false, // Set to false for backward compatibility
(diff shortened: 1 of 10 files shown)


cmd/manager/main.go
```diff
@@ -176,6 +176,11 @@ func Command() *cobra.Command {
 		false,
 		"Disable periodically updating ECK telemetry data for Kibana to consume.",
 	)
+	cmd.Flags().String(
+		operator.DistributionChannelFlag,
+		"",
+		"Set the distribution channel to report through the telemetry.",
```
pebrc, line 182:
```suggestion
		"Set the distribution channel to report through telemetry.",
```

pebrc (approved):
LGTM

deploy/eck-operator/templates/statefulset.yaml
```diff
@@ -42,6 +42,7 @@ spec:
           args:
             - "manager"
             - "--config=/conf/eck.yaml"
+            - "--distributionChannel={{ .Values.telemetry.distributionChannel }}"
```
charith-elastic, line 45: Shouldn't this be `distribution-channel`?

hack/operatorhub/main.go
```diff
@@ -291,6 +292,10 @@ func buildRenderParams(conf *config, extracts *yamlExtracts) (*RenderParams, err
 		additionalArgs = append(additionalArgs, "--ubi-only")
 	}
 
+	if conf.DistributionChannel != "" {
+		additionalArgs = append(additionalArgs, "--distributionChannel=" + conf.DistributionChannel)
```
charith-elastic, line 296:
```suggestion
		additionalArgs = append(additionalArgs, "--distribution-channel=" + conf.DistributionChannel)
```

PR #3802, opened 2020-10-07
Add flag to configure telemetry data updates interval

david-kow:
Introducing a flag to allow changing telemetry updates interval from the default 1 hour for E2E tests. Specifically, for `TestTelemetry` that depends on telemetry data being present.

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -230,6 +230,11 @@ func Command() *cobra.Command {
 		"",
 		"Kubernetes namespace the operator runs in",
 	)
+	cmd.Flags().Duration(
+		operator.TelemetryIntervalFlag,
+		1*time.Hour,
+		"Interval between ECK telemetry data updates",
+	)
 	cmd.Flags().Bool(
 		operator.UBIOnlyFlag,
 		false,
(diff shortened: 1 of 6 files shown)


cmd/manager/main.go
```diff
@@ -266,6 +271,9 @@ func Command() *cobra.Command {
 	_ = cmd.Flags().MarkHidden(operator.AutoPortForwardFlag)
 	_ = cmd.Flags().MarkHidden(operator.DebugHTTPListenFlag)
 
+	// hide the flag used for E2E test only
+	_ = cmd.Flags().MarkHidden(operator.TelemetryIntervalFlag)
```
sebgl, line 275: I'm wondering if this is necessary. I don't see a particular case where users would like to tweak this, but I also don't see any problem in having it exposed so we keep all flags explicit and there's no hidden secret?

david-kow: My motivation was that we already know the right value for it and if not for the E2E test, this wouldn't be configurable at all, just another constant. I also don't see a use case for changing it (there is already a flag to disable it completely), so I don't think there is a compelling reason to add it to the list of configuration options that users need to read/know about/understand.

sebgl (approved):
LGTM

cmd/manager/main.go
sebgl: I'm fine with keeping it hidden.

PR #3803, opened 2020-10-07
Allow the e2e operator to read storage classes

sebgl:
I missed that permission in the e2e operator ClusterRole. As a result
the e2e operator fails when trying to list storage classes.


barkbay (approved):
lgtm

PR #3805, opened 2020-10-07
Update Helm documentation to include storage class validation flag

charith-elastic:
Update the Helm documentation to show that storage class validation should be turned off when installing in a restricted environment.

Ideally, setting `createClusterScopedResources` to `false` should just switch off all cluster-scoped features; however, there may be cases where users may want to only disable a particular feature as well. We end up with a flag soup because of this.


sebgl (approved):
LGTM. Thanks for catching this!

Issue #3495, opened 2020-07-21
TestReconcile is flaky


david-kow:
[Failed with](https://devops-ci.elastic.co/job/cloud-on-k8s-pr/4146/testReport/junit/github/com_elastic_cloud-on-k8s_pkg_controller_license_trial/Run_tests_in_parallel___Run_unit_and_integration_tests___TestReconcile/):
```
=== RUN   TestReconcile
    retry.go:24: 
        	Error Trace:	retry.go:24
        	            				trial_controller_integration_test.go:94
        	            				trial_controller_integration_test.go:54
        	Error:      	Received unexpected error:
        	            	expected licensed features to be enabled [true] but was [false]
        	Test:       	TestReconcile
--- FAIL: TestReconcile (5.21s)
```

barkbay:
It doesn't seem to be the case anymore. Closing, let's reopen if needed.

sebgl:
Got it today: https://devops-ci.elastic.co/job/cloud-on-k8s-pr/4457/testReport/junit/github/com_elastic_cloud-on-k8s_pkg_controller_license_trial/Run_tests_in_parallel___Run_unit_and_integration_tests___TestReconcile/

sebgl:
And again: https://devops-ci.elastic.co/job/cloud-on-k8s-pr/4886/console.

```
09:11:29  === [31mFailed[0m
09:11:29  === [31mFAIL[0m: pkg/controller/license/trial TestReconcile (5.30s)
09:11:29      retry.go:24: 
09:11:29          	Error Trace:	retry.go:24
09:11:29          	            				trial_controller_integration_test.go:89
09:11:29          	            				trial_controller_integration_test.go:49
09:11:29          	Error:      	Received unexpected error:
09:11:29          	            	expected licensed features to be enabled [true] but was [false]
09:11:29          	Test:       	TestReconcile
09:11:29  
```

PR #3806, opened 2020-10-07
Do not set the default elasticsearch-data volume if claims are provided in the ES manifest

### LINKED ISSUE #2574: Do not set the default elasticsearch-data volume if some other claims are provided in the ES manifest
Discovered while investigating https://discuss.elastic.co/t/how-to-define-multiple-path-data-pointing-to-separate-volumes-or-pvc-using-eck/218644.

If I provide my custom `VolumeClaimTemplates`, decide to give them my own name, and handle the volume and mount point stuff myself, ECK still triggers the creation of a default `elasticsearch-data` PersistentVolume that I don't want to use.

It means that if I want to use multiple data volumes as described in https://discuss.elastic.co/t/how-to-define-multiple-path-data-pointing-to-separate-volumes-or-pvc-using-eck/218644, I need to make sure one of them is named `elasticsearch-data`, whereas I would probably have prefered `elasticsearch-data1`, `elasticsearch-data2`, etc.

I think if any VolumeClaimTemplates is provided, we should not set the default one.

pebrc:
Remove the default data volume template if the user specified volume claim templates.
If there is no default data volume also remove the corresponding volume mount from main container and init containers.

Fixes #2574 

Not urgent for v1.3 imo


pebrc:
Jenkins test this please

pebrc:
Jenkins test this please

Ran into https://github.com/elastic/cloud-on-k8s/issues/3495

pkg/controller/elasticsearch/initcontainer/prepare_fs.go
```diff
@@ -126,12 +127,13 @@ func NewPrepareFSInitContainer(
 			PluginVolumes.InitContainerVolumeMounts(),
 			certificatesVolumeMount,
 			scriptsVolume.VolumeMount(),
-			esvolume.DefaultDataVolumeMount,
 			esvolume.DefaultLogsVolumeMount,
 		),
 		Resources: defaultResources,
 	}
 
+	container.VolumeMounts = esvolume.AppendDefaultDataVolumeMount(container.VolumeMounts, volumes)
```
barkbay, line 135: I'm wondering if it is needed since init containers eventually inherit volume mounts from the "main" container: 
https://github.com/elastic/cloud-on-k8s/blob/5915eb125799fc5653d4d2465944f6a8bf382c9d/pkg/controller/common/defaults/pod_template.go#L179-L200

pebrc: I think you are absolutely right. And I believe the same is true for the scripts volume mount and the logs volume mount.

barkbay (approved):
LGTM 👍

pkg/controller/elasticsearch/volume/defaults.go
```diff
@@ -49,3 +49,13 @@ var (
 		MountPath: ElasticsearchLogsMountPath,
 	}
 )
+
+//AppendDefaultDataVolumeMount appends a volume mount for the default data volume if the slice of volumes contains the default data volume.
+func AppendDefaultDataVolumeMount(mounts []corev1.VolumeMount, volumes []corev1.Volume) []corev1.VolumeMount {
```
sebgl, line 54: nit: `//<whitespace>comment`

pkg/controller/elasticsearch/initcontainer/prepare_fs.go
```diff
@@ -125,9 +115,6 @@ func NewPrepareFSInitContainer(
 		VolumeMounts: append(
 			PluginVolumes.InitContainerVolumeMounts(),
 			certificatesVolumeMount,
-			scriptsVolume.VolumeMount(),
-			esvolume.DefaultDataVolumeMount,
-			esvolume.DefaultLogsVolumeMount,
 		),
```
sebgl, line 118: A comment repeating that some volumes are inherited from the main container may be useful here?

sebgl (approved):
LGTM

PR #3823, opened 2020-10-12
Add Logstash recipe

charith-elastic:
Shows how to use Logstash alongside ECK. There have been several questions about Logstash and this might be useful for those users to figure out what they need to do.

### DIFF
--- a/config/recipes/logstash/README.asciidoc
+++ b/config/recipes/logstash/README.asciidoc
@@ -1,0 +1,4 @@
+= Using Logstash with ECK
+
+This recipe demonstrates how to run the link:https://www.elastic.co/guide/en/logstash/current/advanced-pipeline.html[Logstash log parsing example] on Kubernetes with Elasticsearch, Kibana and Filebeat deployed via ECK.
+
--- a/config/recipes/logstash/README.asciidoc
+++ b/config/recipes/logstash/README.asciidoc
@@ -1,4 +1,14 @@
+ifdef::env-github[]
+:tip-caption: :bulb:
+:note-caption: :information_source:
+:important-caption: :heavy_exclamation_mark:
+:caution-caption: :fire:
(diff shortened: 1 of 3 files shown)


config/recipes/logstash/README.asciidoc
```diff
@@ -0,0 +1,4 @@
+= Using Logstash with ECK
+
+This recipe demonstrates how to run the link:https://www.elastic.co/guide/en/logstash/current/advanced-pipeline.html[Logstash log parsing example] on Kubernetes with Elasticsearch, Kibana and Filebeat deployed via ECK.
```
pebrc, line 3: Let's add the `node.store.allow_mmap` disclaimer here for consistency.

config/recipes/logstash/logstash.yaml
```diff
@@ -0,0 +1,185 @@
...
+        containers:
+          - name: filebeat
+            securityContext:
+              runAsUser: 0
+            volumeMounts:
+              - name: data
+                mountPath: /data
+        volumes:
+          - name: data
+            emptydir: {}
```
pebrc, line 185: Don't we need a writeable data volume for filebeat (at least on OCP) with something like 
```diff
--- a/config/recipes/logstash/logstash.yaml
+++ b/config/recipes/logstash/logstash.yaml
@@ -180,6 +180,10 @@ spec:
             volumeMounts:
               - name: data
                 mountPath: /data
+              - name: beat-data
+                mountPath: /usr/share/filebeat/data
         volumes:
           - name: data
             emptydir: {}
+          - name: beat-data
+            emptydir: {}
```

pebrc (commented):
Looks good. Great idea to use the logstash tutorial. I had to tweak it a bit to make it work on OCP (see my comment)

pebrc (approved):
LGTM

Issue #3822, opened 2020-10-12
E2E tests on GKE are failing


barkbay:
E2E [tests on GKE are failing](https://devops-ci.elastic.co/job/cloud-on-k8s-e2e-tests-stack-versions/222/console). But it does not seem to be related to the tests themselves.

After after a few moments _(1 hour ?)_  there is this error in the logs:

```
[2020-10-09T00:17:27.834Z] 2020/10/09 00:17:21 Authenticating as service account...
[2020-10-09T00:17:27.834Z] 2020/10/09 00:17:21 Setting CLOUDSDK_CONFIG=/root/.gcp
[2020-10-09T00:17:27.834Z] Updated property [core/project].
....
[2020-10-09T00:30:20.154Z] {"Time":"2020-10-09T00:30:10.910703292Z","Action":"run","Package":"github.com/elastic/cloud-on-k8s/test/e2e","Test":"TestSamples"}^M
[2020-10-09T00:30:20.154Z] {"Time":"2020-10-09T00:30:10.910872081Z","Action":"output","Package":"github.com/elastic/cloud-on-k8s/test/e2e","Test":"TestSamples","Output":"=== RUN   TestSamples\n"}^M
[...]

PR #3825, opened 2020-10-12
Don't set CLOUDSDK_CONFIG

david-kow:
Removes `CLOUDSDK_CONFIG` env variable from deployer and sets it in `setenvconfig`, only for `ocp` and `ocp3`. This should resolve the issue with e2e runner not knowing the correct path to gcloud config.

Fixes https://github.com/elastic/cloud-on-k8s/issues/3822.

### DIFF
--- a/hack/deployer/runner/ocp3.go
+++ b/hack/deployer/runner/ocp3.go
@@ -66,7 +66,7 @@ func (d Ocp3Driver) Execute() error {
 
 	if err := authToGCP(
 		d.plan.VaultInfo, OcpVaultPath, OcpServiceAccountVaultFieldName,
-		d.plan.ServiceAccount, true, true, d.plan.Ocp3.GCloudProject,
+		d.plan.ServiceAccount, true, d.plan.Ocp3.GCloudProject,
 	); err != nil {
 		return err
 	}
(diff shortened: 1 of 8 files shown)


hack/deployer/runner/ocp3.go
```diff
@@ -180,7 +180,7 @@ func (d Ocp3Driver) runAnsibleDockerContainer(action string) error {
 		-e USER={{.User}} \
 		-e USER_HOME={{.HomeVolumeMountPath}} \
 		-v {{.HomeVolumeName}}:{{.HomeVolumeMountPath}} \
-		-e CLOUDSDK_CONFIG={{.GCloudSDKPath}} \
+		-e CLOUDSDK_CONFIG \
```
pebrc, line 183: I am not sure I understand why this is needed.

david-kow:
run full pr build (testing new pr job)

david-kow:
run full pr build (testing new pr job)

david-kow:
run full pr build (testing new pr job)

hack/deployer/runner/ocp3.go
david-kow: This is to pass that env var to the container if it's set.

david-kow:
After chatting with @thbkrkr and @sebgl I'm removing `CLOUDSDK_CONFIG` from our settings.

For CI, this setting wouldn't be used because we don't care about overriding the default config there.

For dev box, when running deployer and then e2e runner it would need to be set twice. Once for deployer and separately as `export CLOUDSDK_CONFIG=...` for e2e runner to pick it up.

The only time this setting would be useful is when trying to repro CI run on the dev box (because then `setenvconfig` is used). I don't think we should introduce a setting to handle such a specific case, especially given the fact that the same setup (`export CLOUDSDK_CONFIG=...`) that works in other cases will work here as well.

hack/deployer/runner/ocp3.go
thbkrkr: I hesitate to remove `CLOUDSDK_CONFIG` completely as it seems useful to me mainly when we want to run a CI thing without Docker and without altering our existing gcloud config dir. When we run the CI container the gcloud config dir lives in the named Docker volume.

If we keep the variable, I think it would be good to add a comment to explain what this is for to avoid any ambiguity. Something along these lines:
```go
// Propagate the variable that could have been set to not alter an existing gcloud config directory, 
// in the case of a local test outside docker
```

sebgl: Let's add a comment as Thibault suggested here?

sebgl (approved):
LGTM
Let's make sure we document the `$CLOUDSDK_CONFIG` trick in our developer docs.

Issue #3821, opened 2020-10-12
Enterprise Search part of TestSamples E2E test fails


david-kow:
In the last three builds:
https://devops-ci.elastic.co/job/cloud-on-k8s-e2e-tests-ocp/252/#showFailuresLink
https://devops-ci.elastic.co/job/cloud-on-k8s-e2e-tests-ocp/253/#showFailuresLink
https://devops-ci.elastic.co/job/cloud-on-k8s-e2e-tests-ocp/254/#showFailuresLink

barkbay:
I didn't investigate this one but given [this known issue ](https://github.com/elastic/cloud-on-k8s/issues/3656) I guess it's failing because the stack version has been upgraded to `7.9.2`:

https://github.com/elastic/cloud-on-k8s/blob/d2fe6c63dfa6193885366553595537e343fa74c5/test/e2e/test/enterprisesearch/builder.go#L19-L24

PR #3831, opened 2020-10-13
Set more versions as incompatibile with OCP

david-kow:
Fixes https://github.com/elastic/cloud-on-k8s/issues/3821.


sebgl (approved):
LGTM. We need to make sure that once this is fixed in Enterprise Search we relax this check (I'm personally watching the Enterprise Search issue so will get notified about it).

PR #3837, opened 2020-10-14
Allow automatic Elasticsearch nodes discovery

### LINKED ISSUE #3182: Support client nodes discovery (a.k.a. sniffing)
Sniffing allows an Elasticsearch client to automatically discover the nodes in an Elasticsearch cluster. This feature is supported in some of the official clients: [.Net](https://www.elastic.co/guide/en/elasticsearch/client/net-api/current/sniffing-role-detection.html#sniffing-role-detection), [Js](https://www.elastic.co/guide/en/elasticsearch/client/javascript-api/current/client-configuration.html#_basic_options) or [Java](https://www.elastic.co/guide/en/elasticsearch/client/java-rest/master/sniffer.html)

The client uses the `publish_address` fields from the `_nodes/_all/http` endpoint to get the address of each nodes in the cluster, and then updates its own internal connection pool.

`publish_address` may contain both the hostname and the IP address _(in the form `hostname/ip:port`)_: `"publish_address" : "elasticsearch-sample-es-default-0.elasticsearch-sample-es-default.default.svc/10.28.32.40:9200"` The clients use the hostname if it's present.

ECK does not set [`http.publish_host`](https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-http.html#_http_settings). In this case Elasticsearch falls back to `network.publish_host` which is set to the Pod's IP address.
This IP address does not appears in the HTTP certificate which leads to the [situation described by this user](https://discuss.elastic.co/t/accessing-eck-through-nodejs-sdk/235454).

A potential solution would be to manage `http.publish_host`, [I tried to work on a small POC and it seems to be working as expected](https://github.com/elastic/cloud-on-k8s/compare/master...barkbay:sniff).

A few questions though:

* All the Pod IP may change at some point, for example during a rolling restart. Is it something the clients may handle ?
* It needs to publish "not ready" pods in the headless `Service` because Elasticsearch attempts to resolve the name during startup, this is what @anyasabo experienced in https://github.com/elastic/cloud-on-k8s/issues/2830#issuecomment-612641197

This last point is the most important I think and should be discussed.

Related to https://github.com/elastic/elasticsearch/issues/22029

barkbay:
This PR allows automatic discovery of Elasticsearch nodes by clients which support "sniffing" mode.

More precisely it:

* Sets [`http.publish_host`](https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-http.html#_http_settings) to the Pod hostname _(in the form `$(podname).$(governing service domain).$(namespace).svc`)_. Note that it is different from `network.publish_host`, therefore I _think_ we should not hit https://github.com/elastic/cloud-on-k8s/issues/3723 again.
* Adds the DNS wildcards `*.$(governing service domain).$(namespace).svc` in the HTTP self signed certificate, for example:

```
        X509v3 extensions:
            X509v3 Key Usage: critical
                Digital Signature, Key Encipherment
            X509v3 Extended Key Usage:
                TLS Web Server Authentication, TLS Web Client Authentication
            X509v3 Subject Alternative Name:
                DNS:elasticsearch-sample-es-http.default.es.local,
                DNS:elasticsearch-sample-es-http,
                DNS:elasticsearch-sample-es-http.default.svc,
                DNS:elasticsearch-sample-es-http.default,
                DNS:*.elasticsearch-sample-es-default.default.svc
```

Autodiscovery mostly makes sense if the client has an interface in the Kubernetes network. Therefore DNS wildcards are added automatically in the SANs only in the self signed certificate.

IPs are ephemeral in a Kubernetes cluster, I think it is fair to assume that DNS issues should be handled by the client. During my tests with the [Node.js client](https://github.com/elastic/elasticsearch-js) it has always been able to gently reconnect to the cluster while I was randomly deleting the Elasticsearch Pods.

A sample based on the official ones is available [here](https://github.com/barkbay/eck-elasticsearch-js) if you want to test that feature.

Fixes #3182

### DIFF
--- a/pkg/controller/common/certificates/reconcile.go
+++ b/pkg/controller/common/certificates/reconcile.go
@@ -32,8 +32,9 @@ type Reconciler struct {
 	K8sClient      k8s.Client
 	DynamicWatches watches.DynamicWatches
 
-	Object     metav1.Object       // owner for the TLS certificates (eg. Elasticsearch, Kibana)
-	TLSOptions commonv1.TLSOptions // TLS options of the object
+	Object         metav1.Object                     // owner for the TLS certificates (eg. Elasticsearch, Kibana)
+	TLSOptions     commonv1.TLSOptions               // TLS options of the object
+	ControllerSANs []commonv1.SubjectAlternativeName // SANs dynamically set by a controller, only used in the self signed cert
 
 	Namer    commonname.Namer  // helps naming the reconciled secrets
 	Labels   map[string]string // to set on the reconciled cert secrets
--- a/pkg/controller/common/certificates/reconcile.go
+++ b/pkg/controller/common/certificates/reconcile.go
@@ -32,9 +32,9 @@ type Reconciler struct {
 	K8sClient      k8s.Client
 	DynamicWatches watches.DynamicWatches
 
-	Object         metav1.Object                     // owner for the TLS certificates (eg. Elasticsearch, Kibana)
-	TLSOptions     commonv1.TLSOptions               // TLS options of the object
-	ControllerSANs []commonv1.SubjectAlternativeName // SANs dynamically set by a controller, only used in the self signed cert
+	Object        metav1.Object                     // owner for the TLS certificates (eg. Elasticsearch, Kibana)
+	TLSOptions    commonv1.TLSOptions               // TLS options of the object
+	ExtraHTTPSANs []commonv1.SubjectAlternativeName // SANs dynamically set by a controller, only used in the self signed cert
 
 	Namer    commonname.Namer  // helps naming the reconciled secrets
 	Labels   map[string]string // to set on the reconciled cert secrets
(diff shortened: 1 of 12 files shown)


barkbay:
jenkins test this please

pkg/controller/common/certificates/reconcile.go
```diff
@@ -32,8 +32,9 @@ type Reconciler struct {
 	K8sClient      k8s.Client
 	DynamicWatches watches.DynamicWatches
 
-	Object     metav1.Object       // owner for the TLS certificates (eg. Elasticsearch, Kibana)
-	TLSOptions commonv1.TLSOptions // TLS options of the object
+	Object         metav1.Object                     // owner for the TLS certificates (eg. Elasticsearch, Kibana)
+	TLSOptions     commonv1.TLSOptions               // TLS options of the object
+	ControllerSANs []commonv1.SubjectAlternativeName // SANs dynamically set by a controller, only used in the self signed cert
```
sebgl, line 37: Naming suggestion: I'd go for a more generic `ExtraHTTPSANs`. I feel like `ControllerSANs` is misleading: in practice this will contain a name based on the StatefulSet name, not anything related to the controller itself.

sebgl (approved):
LGTM, just one naming suggestion.

PR #3848, opened 2020-10-15
Reinstate the non-pvc volume handling during defaulting of data volumes.

pebrc:
d576c01055cbe6be48b336242c004a900b69805f had inadvertently removed the handling of non-pvc template based volumes when deciding whether to default a data volume pvc template. This PR reverts that aspect of defaulting behviour. 

This cover the case where users have added an `emptyDir` volume with the name `elasticsearch-data` with the intent to use it as the main data volume. We will now again honor this and not add a default data volume. If they added however  some other statically provisioned volume directly to the Pod spec, we will still provision the default data volume through a PVC template.

### DIFF
--- a/pkg/controller/common/defaults/pvc.go
+++ b/pkg/controller/common/defaults/pvc.go
@@ -1,0 +1,41 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License;
+// you may not use this file except in compliance with the Elastic License.
+
+package defaults
+
+import corev1 "k8s.io/api/core/v1"
+
+// AppendDefaultPVCs appends defaults PVCs to a set of existing ones.
+//
+// The default PVC is not appended if:
+// - a Volume with the same .Name is found in podSpec.Volumes, and that volume is not a PVC volume
+// - any PVC has been defined by the user
+func AppendDefaultPVCs(
+	existing []corev1.PersistentVolumeClaim,
+	podSpec corev1.PodSpec,
+	defaults ...corev1.PersistentVolumeClaim,
+) []corev1.PersistentVolumeClaim {
+	// create a set of volume names that are not PVC-volumes for efficient testing
+	nonPVCvolumes := map[string]struct{}{}
+
+	for _, volume := range podSpec.Volumes {
+		if volume.PersistentVolumeClaim == nil {
+			// this volume is not a PVC
+			nonPVCvolumes[volume.Name] = struct{}{}
+		}
+	}
+	// any user defined PVC shortcuts the defaulting attempt
+	if len(existing) > 0 {
+		return existing
+	}
+
+	for _, defaultPVC := range defaults {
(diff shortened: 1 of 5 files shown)


pkg/controller/common/defaults/pvc.go
```diff
@@ -0,0 +1,41 @@
...
+// or more contributor license agreements. Licensed under the Elastic License;
+// you may not use this file except in compliance with the Elastic License.
+
+package defaults
+
+import corev1 "k8s.io/api/core/v1"
+
+// AppendDefaultPVCs appends defaults PVCs to a set of existing ones.
+//
+// The default PVC is not appended if:
```
sebgl, line 11: default PVcs are not appended (plural)?

pkg/controller/common/defaults/pvc.go
```diff
@@ -0,0 +1,41 @@
...
+	for _, volume := range podSpec.Volumes {
+		if volume.PersistentVolumeClaim == nil {
+			// this volume is not a PVC
+			nonPVCvolumes[volume.Name] = struct{}{}
+		}
+	}
+	// any user defined PVC shortcuts the defaulting attempt
+	if len(existing) > 0 {
+		return existing
+	}
```
sebgl, line 31: nit: I think I would check that first (before grabbing non-PVCs)?

pkg/controller/common/defaults/pvc.go
```diff
@@ -0,0 +1,41 @@
...
+
+package defaults
+
+import corev1 "k8s.io/api/core/v1"
+
+// AppendDefaultPVCs appends defaults PVCs to a set of existing ones.
+//
+// The default PVC is not appended if:
+// - a Volume with the same .Name is found in podSpec.Volumes, and that volume is not a PVC volume
+// - any PVC has been defined by the user
```
sebgl, line 13: nit: I feel like it's easier to understand by switching the order of these 2 statements.

```
Defaults are not appended if the user already provided some PVCs, or if the user already provided the default volumes (same name) in the podSpec.
```

pkg/controller/common/defaults/pvc.go
```diff
@@ -0,0 +1,41 @@
...
+func AppendDefaultPVCs(
+	existing []corev1.PersistentVolumeClaim,
+	podSpec corev1.PodSpec,
+	defaults ...corev1.PersistentVolumeClaim,
+) []corev1.PersistentVolumeClaim {
+	// create a set of volume names that are not PVC-volumes for efficient testing
```
sebgl, line 19: not sure I understand how this impacts testing: we don't return this map to be checked in unit tests?

sebgl (approved):
LGTM. Left a few readability nits.

barkbay (approved):
lgtm

pkg/controller/common/defaults/pvc.go
barkbay: I think this comment is about this test: 
```golang
if _, isNonPVCVolume := nonPVCvolumes[defaultPVC.Name]; isNonPVCVolume
```

pebrc: Michael is right, I think that's what @\nkvoll meant originally.

PR #3850, opened 2020-10-16
Create transport certs for existing StatefulSets/Pods

### LINKED ISSUE #3846: Pods certificates are not reconciled if not part of the expected nodeSets
We recently changed the way we reconcile nodes certificates by [creating a secret per StatefulSet](https://github.com/elastic/cloud-on-k8s/pull/3828).

The new reconciliation logic only considers the expected StatefulSets, based on nodeSets spec, when reconciling Pods certificates (`reconcileNodeSetTransportCertificatesSecrets `).
Whereas in the past we used to reconcile certificates regardless of the StatefulSets or NodeSet they belong to.

I think we want to:
- reconcile existing Pods certificates before reconciling nodeSets/StatefulSets: ensures all existing Pods have a cert
- create the Secrets holding those certs before creating the StatefulSets: we want to make sure a Pod doesn't get created while the secret doesn't exist (yet). It would report an error in that case.

Relates https://github.com/elastic/cloud-on-k8s/issues/3844 (not exactly the root cause but explains why the test fails now and not before).

barkbay:
Fixes #3846 by creating transport certificates according to the Elasticsearch spec, but also for existing StatefulSets/Pods which might not be part of the spec anymore.

### DIFF
--- a/pkg/controller/elasticsearch/certificates/transport/reconcile.go
+++ b/pkg/controller/elasticsearch/certificates/transport/reconcile.go
@@ -26,6 +26,7 @@ import (
 	"github.com/elastic/cloud-on-k8s/pkg/controller/elasticsearch/sset"
 	"github.com/elastic/cloud-on-k8s/pkg/utils/k8s"
 	"github.com/elastic/cloud-on-k8s/pkg/utils/maps"
+	"github.com/elastic/cloud-on-k8s/pkg/utils/set"
 )
 
 var log = logf.Log.WithName("transport")
@@ -48,12 +49,12 @@ func ReconcileTransportCertificatesSecrets(
 	if err != nil {
 		return results.WithError(err)
 	}
-	ssets := make(map[string]struct{})
+	ssets := set.Make()
 	for _, actualStatefulSet := range actualStatefulSets {
-		ssets[actualStatefulSet.Name] = struct{}{}
+		ssets.Add(actualStatefulSet.Name)
 	}
 	for _, nodeSet := range es.Spec.NodeSets {
-		ssets[esv1.StatefulSet(es.Name, nodeSet.Name)] = struct{}{}
+		ssets.Add(esv1.StatefulSet(es.Name, nodeSet.Name))
 	}
 
 	for ssetName := range ssets {
--- a/pkg/controller/elasticsearch/certificates/transport/reconcile.go
+++ b/pkg/controller/elasticsearch/certificates/transport/reconcile.go
@@ -23,6 +23,7 @@ import (
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common/certificates"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common/reconciler"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/elasticsearch/label"
+	"github.com/elastic/cloud-on-k8s/pkg/controller/elasticsearch/sset"
 	"github.com/elastic/cloud-on-k8s/pkg/utils/k8s"
 	"github.com/elastic/cloud-on-k8s/pkg/utils/maps"
 )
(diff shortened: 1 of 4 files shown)


pkg/controller/elasticsearch/certificates/transport/reconcile_test.go
```diff
@@ -105,12 +105,16 @@ func TestReconcileTransportCertificatesSecrets(t *testing.T) {
 					// Create 3 Secrets but create transport certs only for the first Pods in each StatefulSet
 					newtransportCertsSecretBuilder(testEsName, "sset1").forPodIndices(0).build(),
 					newtransportCertsSecretBuilder(testEsName, "sset2").forPodIndices(0).build(),
+					// Add an existing statefulSet which is not par of the Spec
```
pebrc, line 108:
```suggestion
					// Add an existing statefulSet which is not part of the Spec
```

barkbay:
Jenkins test this please

Build failed with following error:
```
11:17:56  stderr: fatal: Couldn't find remote ref refs/heads/${sha1}
11:17:56  
11:17:56  	at org.jenkinsci.plugins.gitclient.CliGitAPIImpl.launchCommandIn(CliGitAPIImpl.java:2450)
11:17:56  	at org.jenkinsci.plugins.gitclient.CliGitAPIImpl.launchCommandWithCredentials(CliGitAPIImpl.java:2051)
11:17:56  	at org.jenkinsci.plugins.gitclient.CliGitAPIImpl.access$500(CliGitAPIImpl.java:84)
11:17:56  	at org.jenkinsci.plugins.gitclient.CliGitAPIImpl$1.execute(CliGitAPIImpl.java:573)
11:17:56  	at jenkins.plugins.git.GitSCMFileSystem$BuilderImpl.build(GitSCMFileSystem.java:364)
11:17:56  	at jenkins.scm.api.SCMFileSystem.of(SCMFileSystem.java:197)
11:17:56  	at jenkins.scm.api.SCMFileSystem.of(SCMFileSystem.java:173)
11:17:56  	at org.jenkinsci.plugins.workflow.cps.CpsScmFlowDefinition.create(CpsScmFlowDefinition.java:115)
11:17:56  	at org.jenkinsci.plugins.workflow.cps.CpsScmFlowDefinition.create(CpsScmFlowDefinition.java:69)
11:17:56  	at org.jenkinsci.plugins.workflow.job.WorkflowRun.run(WorkflowRun.java:309)
11:17:56  	at hudson.model.ResourceController.execute(ResourceController.java:97)
11:17:56  	at hudson.model.Executor.run(Executor.java:428)
11:17:56  Finished: FAILURE
```

barkbay:
Jenkins test this please

barkbay:
Jenkins test this please

pkg/controller/elasticsearch/certificates/transport/reconcile.go
```diff
@@ -39,8 +40,24 @@ func ReconcileTransportCertificatesSecrets(
...
 	results := &reconciler.Results{}
+
+	// We must create transport certificates for the following StatefulSets:
+	// - the ones that still exist, even if they have been removed from the Spec
+	// - the ones that do not exist yet, but will be created in a later step of the reconciliation
+	actualStatefulSets, err := sset.RetrieveActualStatefulSets(c, k8s.ExtractNamespacedName(&es))
+	if err != nil {
+		return results.WithError(err)
+	}
+	ssets := make(map[string]struct{})
```
pebrc, line 51: We have a `StringSet` utility, which makes this a little bit better to read. I recently changed it to support an empty set to start with.

pkg/controller/elasticsearch/certificates/transport/reconcile.go
```diff
@@ -39,8 +40,24 @@ func ReconcileTransportCertificatesSecrets(
...
+	for _, actualStatefulSet := range actualStatefulSets {
+		ssets[actualStatefulSet.Name] = struct{}{}
+	}
 	for _, nodeSet := range es.Spec.NodeSets {
-		if err := reconcileNodeSetTransportCertificatesSecrets(c, ca, es, nodeSet.Name, rotationParams); err != nil {
+		ssets[esv1.StatefulSet(es.Name, nodeSet.Name)] = struct{}{}
```
pebrc, line 56: The one thing I wonder is whether we could go one step further an anticipate the Pods that are about to be created and generate their certificates ahead of time. Because if I understand the flow correctly currently we will just create a secret with the CA cert in it and it takes at least another reconciliation iteration to then provision the certificates for the Pods which are hanging in the init step until the certs appear. 

This would be a bit of a bigger refactoring and if we think it is useful, it would be OK imo to do it in a separate PR and now just fix the broken behaviour.

pebrc (commented):
Looks good. Not sure what is going on with Jenkins though.

pkg/controller/elasticsearch/certificates/transport/reconcile.go
sebgl: We need their IP, I think, so we need to wait for the Pods to be created anyway.

pebrc: Oh you are right we reverted that change that used DNS names in the `network.publish_host`. I forgot about that.

barkbay:
run full pr build

pebrc (approved):
LGTM

pkg/controller/elasticsearch/certificates/transport/transport_fixtures_test.go
```diff
@@ -233,3 +234,17 @@ func getSecret(list corev1.SecretList, name string) *corev1.Secret {
 	}
 	return nil
 }
+
+func newStatfulSet(esName, ssetName string) *v1.StatefulSet {
```
sebgl, line 238: typo: StatefulSet

sebgl (approved):
LGTM

PR #3853, opened 2020-10-16
Change status-context to separate two PR jobs in PR status

david-kow:
IIUC, [status-context](https://docs.openstack.org/infra/jenkins-job-builder/triggers.html#triggers.github-pull-request) is responsible for the name used to identify job in PR status (see `devops-ci` in the image below). Right now both PR jobs have the same name and they keep overriding each other. Changing the name for full job should allow us to see status for both jobs at the same time.

Screenshot 2020-10-16 at 12 38 48

### DIFF
--- a/.ci/jobs/pr-gke-full.yml
+++ b/.ci/jobs/pr-gke-full.yml
@@ -20,7 +20,7 @@
             - elastic
           allow-whitelist-orgs-as-admins: true
           github-hooks: true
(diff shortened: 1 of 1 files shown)


pebrc (approved):
Are slashes OK in these names?

david-kow:
@pebrc, I think so:
Screenshot 2020-10-16 at 12 43 58

PR #3855, opened 2020-10-19
Add note about devops-ci/full build status

david-kow:
Green checkmark might be interpreted as all tests passed which is not the case here.


barkbay (approved):
lgtm

PR #3858, opened 2020-10-20
Include secret token in APM Server config checksum

### LINKED ISSUE #523: APM: support changing the secret token
Currently the deployment is not rolled when the secret token changes.

pebrc:
The secret token is injected via an environment variable into the config file which is why changes to the token were not taken into account when building the config checksum. By adding the token value explicitly to the checksum APM Server pods will be rotated when the secret token changes.

Fixes #523

### DIFF
--- a/pkg/controller/apmserver/deployment_test.go
+++ b/pkg/controller/apmserver/deployment_test.go
@@ -447,25 +447,28 @@ func TestReconcileApmServer_deploymentParams(t *testing.T) {
 		{
 			name: "secret token influences checksum",
 			args: args{
-				as:            apmFixture,
-				podSpecParams: defaultPodSpecParams,
-				initialObjects: []runtime.Object{
-					&corev1.Secret{
-						ObjectMeta: metav1.ObjectMeta{
-							Name: certSecretName,
-						},
-					},
-					&corev1.Secret{
+				as: apmFixture,
+				podSpecParams: func() PodSpecParams {
+					params := defaultPodSpecParams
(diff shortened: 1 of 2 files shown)


pkg/controller/apmserver/deployment_test.go
```diff
@@ -444,6 +444,30 @@ func TestReconcileApmServer_deploymentParams(t *testing.T) {
...
+			args: args{
+				as:            apmFixture,
+				podSpecParams: defaultPodSpecParams,
+				initialObjects: []runtime.Object{
+					&corev1.Secret{
+						ObjectMeta: metav1.ObjectMeta{
+							Name: certSecretName,
+						},
+					},
+					&corev1.Secret{
```
barkbay, line 458: This `Secret` is not used. This is because the token Secret is expected to be read before the call to` deploymentParams( as *apmv1.ApmServer, params PodSpecParams)` and provided as part of `params PodSpecParams`:

```golang
type PodSpecParams struct {
	Version         string
	CustomImageName string

	PodTemplate corev1.PodTemplateSpec

	TokenSecret  corev1.Secret // Here
	ConfigSecret corev1.Secret

	keystoreResources *keystore.Resources
}
```

barkbay (commented):
Overall the change lgtm and work as expected. I left a comment on the unit test which does not actually test it.

sebgl (approved):
LGTM

barkbay (approved):
lgtm

PR #3860, opened 2020-10-20
Remove unnecessary metrics from _nodes ES API request

### LINKED ISSUE #3249: No need for `?jvm` when retrieving the list of nodes
We call `GET /_nodes/_all/jvm,settings` to retrieve the list of nodes in the cluster, and their roles.
The response has a lot of details we don't need within a reconciliation, and seems only used inside E2E tests.
We should probably split this into 2 different functions:
- one that simply lists the nodes (what's the best API call for that?) for the operator
- another one that retrieves jvm settings for E2E tests purposes

pebrc:
The operator is not using any information from the jvm or settings metrics filter. And even the e2e tests are only interested in the node roles which are already included in the standard payload. Any previously existing use of JVM metrics seems to have been removed/replaced. 

Fixes #3249

### DIFF
--- a/pkg/controller/elasticsearch/client/v6.go
+++ b/pkg/controller/elasticsearch/client/v6.go
@@ -94,7 +94,7 @@ func (c *clientV6) ReloadSecureSettings(ctx context.Context) error {
 
 func (c *clientV6) GetNodes(ctx context.Context) (Nodes, error) {
 	var nodes Nodes
-	// restrict call to minimal node information with a non-existing metric filter
+	// restrict call to minimal node information with a non-existent metric filter
 	err := c.get(ctx, "/_nodes/_all/no-metrics", &nodes)
 	return nodes, err
 }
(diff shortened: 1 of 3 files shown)


pkg/controller/elasticsearch/client/v6.go
```diff
@@ -94,8 +94,8 @@ func (c *clientV6) ReloadSecureSettings(ctx context.Context) error {
 
 func (c *clientV6) GetNodes(ctx context.Context) (Nodes, error) {
 	var nodes Nodes
-	// restrict call to basic node info only
-	err := c.get(ctx, "/_nodes/_all/jvm,settings", &nodes)
+	// restrict call to minimal node information with a non-existing metric filter
```
anyasabo, line 97:
```suggestion
	// restrict call to minimal node information with a non-existent metric filter
```

anyasabo (approved):

sebgl (approved):
LGTM

PR #3865, opened 2020-10-21
Lint cleanup

anyasabo:
Bumps the linter to 1.31 and cleans up our config. Some of the linters are re-enabled, and the disabled ones have comments explaining why. Where applicable I also moved the `// nolint` from the config into the code, which I feel mixed about. But it does let the `nolintlint` checker work and catch no longer applicable exclusions, which I think is a net positive compared to having a long list of potentially no longer applicable nolint rules in the config.

The biggest changes here are running gofumpt, gci, and godot across the repo.

Gufumpt and gci actually seem useful for formatting consistently. We have had code review comments about newlines and import order etc that could be caught ahead of time. godot I was less sure about. The [reasoning](https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences) is decent and from upstream though so I went ahead and kept it. It makes this already large PR even larger though.

We might also want to add contributor docs on how to set up common editors to say, use gofumpt and gci on save.

This is a pretty large PR but I think there's a decent argument for it. I can break them up into different PRs though if that makes sense. I tried to comment the ones that aren't just automated changes.


pkg/apis/elasticsearch/v1/name.go
```diff
@@ -29,10 +29,10 @@ const (
 	statefulSetTransportCertificatesSecretSuffix = "transport-certs"
 
 	// calling this secret "xpack-file-realm" is conceptually wrong since it also holds the file-based roles which
-	// are not part of the file realm - let's still keep this legacy name for convenience
-	rolesAndFileRealmSecretSuffix = "xpack-file-realm"
+	// are not part of the file realm - let's still keep this legacy name for convenience.
+	rolesAndFileRealmSecretSuffix = "xpack-file-realm" // nolint:gosec
```
anyasabo, line 33: I wonder if we want to disable these "hardcoded secret" checks entirely as I cannot imagine we really have any place where we might accidentally put in a hardcoded secret

pkg/controller/elasticsearch/services/services.go
```diff
@@ -146,7 +146,7 @@ func ElasticsearchURL(es esv1.Elasticsearch, pods []corev1.Pod) string {
 	}
 	if schemeChange {
 		// switch to sending requests directly to a random pod instead of going through the service
-		randomPod := pods[rand.Intn(len(pods))]
+		randomPod := pods[rand.Intn(len(pods))] // nolint:gosec
```
anyasabo, line 149: These I think are actually useful since there are places where we might want to make sure we use a csprng, and we made that mistake in the past IIRC

pkg/controller/enterprisesearch/version_upgrade.go
```diff
@@ -201,7 +198,7 @@ func (r *VersionUpgrade) serviceURL() string {
 }
 
 // readOnlyModeRequest builds the HTTP request to toggle the read-only mode on Enterprise Search.
-func (r *VersionUpgrade) readOnlyModeRequest(enabled bool) (*http.Request, error) {
+func (r *VersionUpgrade) readOnlyModeRequest(ctx context.Context, enabled bool) (*http.Request, error) {
```
anyasabo, line 201: This is the only substantial change I think. It also uncovered a couple places where we kind of take extra steps with the context passing both in entsrch and ES that could be simplified. In some place we build the request without a context, then we pass the context and request to a wrapper function that then adds the context and some headers to the request.

In a lot of cases we already have the context when we build the request, so we might as well build it with the context instead of adding it later. It makes it easier to tell if we are propagating contexts everywhere. Definitely a nitpicky thing but something we will probably want to do as we add context aware logging to existing controllers.

test/e2e/test/http_client.go
```diff
@@ -29,7 +29,8 @@ func NewHTTPClient(caCerts []*x509.Certificate) *http.Client {
 		certPool.AddCert(c)
 	}
 	transport.TLSClientConfig = &tls.Config{
-		RootCAs: certPool,
+		MinVersion: tls.VersionTLS12, // ES 6.8 only supports 1.2
```
anyasabo, line 32: I think this is also useful just to make sure we aren't doing something silly, but is definitely not necessary. The default is TLS 1.0.

anyasabo:
jenkins test this please

barkbay:
run full pr build

barkbay (commented):
I _tried_ to go through all the files, overall I think I'm 👍 for those changes.

I only have one concern re. `godot` which makes some comments a bit odd sometimes, for example when a period is added at the end of a URL, or a version (`// V77 indicates version 7.7.`) or a type (`newReconciler returns a new reconcile.Reconciler.`)

I triggered [a full pr job](https://devops-ci.elastic.co/job/cloud-on-k8s-pr-full/23/console) and got these 2 errors:

```
12:22:02  golangci-lint run
12:24:24  [1mtest/e2e/beat/recipes_test.go:83[0m:1: [31mcommentFormatting: put a space between `//` and comment text[0m (gocritic)
12:24:24  // TODO: flaky test, investigate then re-enable
12:24:24  [33m^[0m
12:24:24  [1mtest/e2e/beat/recipes_test.go:197[0m:1: [31mcommentFormatting: put a space between `//` and comment text[0m (gocritic)
12:24:24  // TODO: flaky test, investigate then re-enable
12:24:24  [33m^[0m
12:24:24  make: *** [Makefile:146: lint] Error 1
12:24:24  Makefile:48: recipe for target 'ci-internal' failed
```

I think it comes from https://github.com/elastic/cloud-on-k8s/pull/3888, which means that we should update and commit this one sooner than later I guess.

anyasabo:
Thanks for the feedback @barkbay I think a lot of that is fair. I will still break this out into separate PRs I think to make it easier to weigh each individual change, and hopefully you don't need to spend much more time reviewing.

PR #3872, opened 2020-10-22
Add Helm migration instructions

charith-elastic:
Describes how to migrate an existing installation to Helm. 

Thanks to Thibault for raising the question and Anya for pointing out the solution.

### DIFF
--- a/docs/operating-eck/installing-eck.asciidoc
+++ b/docs/operating-eck/installing-eck.asciidoc
@@ -112,7 +112,9 @@ helm show values elastic/eck-operator
 
 CAUTION: Migrating an existing installation to Helm is essentially an upgrade operation and any <<{p}-beta-to-ga-rolling-restart,caveats associated with normal operator upgrades>> are applicable. Read the <<{p}-ga-upgrade,upgrade documentation>> before proceeding.
 
-You can migrate an existing operator installation to Helm by doing the following:
+You can migrate an existing operator installation to Helm by adding the `meta.helm.sh/release-name`, `meta.helm.sh/release-namespace` annotations and the `app.kubernetes.io/managed-by` label to all the resources you want to be adopted by Helm. You _must_ do this for the Elastic Custom Resource Definitions (CRD) because deleting them would trigger the deletion of all deployed Elastic applications as well. All other resources are optional and can be deleted. 
+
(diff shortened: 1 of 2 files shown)


docs/operating-eck/installing-eck.asciidoc
```diff
@@ -104,3 +104,57 @@ You can view all configurable values by running the following:
 ----
 helm show values elastic/eck-operator
 ----
+
+
+[float]
+[id="{p}-migrate-to-helm"]
+=== Migrate an existing installation to Helm
+
+If you have an operator installed using the `all-in-one.yaml` method, you can migrate to Helm by running the following script, which is also available for download from the link:{eck_github}/tree/{eck_release_branch}/deploy[ECK source repository]. It should preserve your existing Elastic applications deployed using the old operator.
```
alaudazzi, line 113:
```suggestion
If you have an operator installed using the `all-in-one.yaml` method, you can migrate to Helm by running the following script, that you can also download from the link:{eck_github}/tree/{eck_release_branch}/deploy[ECK source repository]. It should preserve your existing Elastic applications deployed using the old operator. 
```

alaudazzi (approved):
Minor suggestion. LGTM.

anyasabo:
The script is super useful. I'm worried a little bit about maintaining it in two places though. I also wonder if we want to describe the essential pieces in prose -- label the CRDs, delete the other resources, install the chart -- rather than have users parse out what they can keep if their situation does not directly apply to the script.

charith-elastic:
I did try to describe the steps but the resulting text was confusing even for me. I feel that a script is much easier to understand than a wall of text.

What parts do you think users will need to parse out? This script covers the installation method we have in our docs. Presumably the advanced users who have customized it are capable of modifying the script as well.

anyasabo:
We have had difficulty in the past with users parsing out shell scripts (for example the diagnostic). It unfortunately has not been as self documenting as we might have hoped. Maybe something like:


>To adopt existing resources into a new Helm chart deployment. Add these annotations on each resource (substituting the appropriate Helm release name and namespace you will use):
meta.helm.sh/release-name="$RELEASE_NAME"
meta.helm.sh/release-namespace="$RELEASE_NAMESPACE"

>And this label
app.kubernetes.io/managed-by=Helm

>This is primarily useful for the custom resource definitions, where deleting the CRD also removes all instances of that resource type. The other resources are stateless and can be deleted. The following script illustrates the process:

would work?

sebgl (approved):
LGTM

PR #3862, opened 2020-10-20
Fail CI on lint failures

anyasabo:
Fixes https://github.com/elastic/cloud-on-k8s/issues/3815

Previous versions of this PR reverted 1fe23a45837ed41509e51c55c94b18119295d899 and so should have failed the lint check before this fix was in place.

PR #3873, opened 2020-10-22
[1.3] Fail CI on lint failures (#3862)

anyasabo:
 - Fail CI on lint failures (#3862)


barkbay (approved):
lgtm

PR #3876, opened 2020-10-26
Add support for webhook constraints

### LINKED ISSUE #3431: Webhook ignores namespace restriction

charith-elastic:
This PR is a minor refactoring of the Helm chart to allow users to set `namespaceSelector` and `objectSelector` on the webhook. The webhook name is now generated as `<operator-name>.<operator-namespace>.k8s.elastic.co` so that multiple operators can be installed side-by-side in the same cluster. This can be useful in a very large and busy cluster where dividing the set of namespaces between several operators can speed up reconciliation times and reduce the amount of resources required per operator.

Since webhooks are cluster-scoped resources, currently the chart requires `createClusterScopedResources` to be `true` as well. This means that each operator gets a `ClusterRole` not just restricted to the set of namespaces they are managing. It avoids over-complicating the chart at this point and should be fine for a load-splitting scenario like above. In an environment where operators are run by untrusted users and need to be locked down tightly, deploying webhooks is probably not an option anyway.

<details>
<summary>How to test</summary>

```sh
# Create managed namespaces and label them
for N in a b c d; do kubectl create ns "ns-$N"; kubectl label ns "ns-$N" name="ns-$N"; done

# Install the first operator to manage ns-a and ns-b namepsaces
 helm install operator-a ./deploy/eck-operator -n operator-a --create-namespace --values=<(cat <<EOF
nameOverride: operator-a
fullnameOverride: operator-a
installCRDs: true
managedNamespaces: ["ns-a", "ns-b"]

image:
  repository: docker.elastic.co/eck-dev/eck-operator-cell
  tag: 1.4.0-SNAPSHOT-d4c12355

webhook:
  enabled: true
  failurePolicy: Fail
  namespaceSelector:
    matchExpressions:
      - key: "name"
        operator: In
        values: ["ns-a", "ns-b"]
EOF
)

# Install the second operator to manage ns-c and ns-d namepsaces
 helm install operator-b ./deploy/eck-operator -n operator-b --create-namespace --values=<(cat <<EOF
nameOverride: operator-b
fullnameOverride: operator-b
installCRDs: false
managedNamespaces: ["ns-c", "ns-d"]

image:
  repository: docker.elastic.co/eck-dev/eck-operator-cell
  tag: 1.4.0-SNAPSHOT-d4c12355

webhook:
  enabled: true
  failurePolicy: Fail
  namespaceSelector:
    matchExpressions:
      - key: "name"
        operator: In
        values: ["ns-c", "ns-d"]
EOF
)
```

Resources deployed to `ns-a` or `ns-b`  will be validated by `https://operator-a-webhook.operator-a.svc`. You can verify this by simply scaling down the operator (`kubectl scale --replicas=0 sts/operator-a -n operator-a`) so that the webhook call fails. Since `failurePolicy` is set to `Fail`  the webhook request would ultimately fail and produce an error message indicating the endpoint it was trying to reach. 

```sh
# Cleanup
helm delete operator-a -n operator-a
helm delete operator-b -n operator-b
```

</details>

Fixes #3431

### DIFF
--- a/deploy/eck-operator/profile-soft-multi-tenancy.yaml
+++ b/deploy/eck-operator/profile-soft-multi-tenancy.yaml
@@ -7,6 +7,12 @@ refs:
 
 webhook:
   enabled: true
+  namespaceSelector:
+    matchExpressions:
+      - key: "eck.k8s.elastic.co/tenant"
+        operator: In
+        values: ["team-a", "team-b"]
+
 
 softMultiTenancy:
   enabled: true
--- a/deploy/eck-operator/templates/_helpers.tpl
+++ b/deploy/eck-operator/templates/_helpers.tpl
@@ -68,6 +68,42 @@ Create the name of the service account to use
 {{- end }}
 
 {{/*
+Determine the name for the webhook 
+*/}}
+{{- define "eck-operator.webhookName" -}}
+{{- if .Values.internal.manifestGen -}}
+elastic-webhook.k8s.elastic.co
+{{- else -}}
+{{- $name := include "eck-operator.name" . -}}
+{{ printf "%s.%s.k8s.elastic.co" $name .Release.Namespace }}
+{{- end -}}
+{{- end -}}
+
+{{/*
+Determine the name for the webhook secret 
+*/}}
+{{- define "eck-operator.webhookSecretName" -}}
+{{- if .Values.internal.manifestGen -}}
+elastic-webhook-server-cert
+{{- else -}}
+{{- $name := include "eck-operator.name" . -}}
+{{ printf "%s-webhook-cert" $name | trunc 63 }}
+{{- end -}}
+{{- end -}}
+
+{{/*
+Determine the name for the webhook service 
+*/}}
+{{- define "eck-operator.webhookServiceName" -}}
+{{- if .Values.internal.manifestGen -}}
+elastic-webhook-server
+{{- else -}}
+{{- $name := include "eck-operator.name" . -}}
+{{ printf "%s-webhook" $name | trunc 63 }}
+{{- end -}}
+{{- end -}}
+
+{{/*
 Add the webhook sideEffects field on supported Kubernetes versions
 */}}
 {{- define "eck-operator.webhookSideEffects" -}}
@@ -77,6 +113,7 @@ Add the webhook sideEffects field on supported Kubernetes versions
 sideEffects: "None"
 {{- end }}
 {{- end }}
+
 
 {{/*
 RBAC permissions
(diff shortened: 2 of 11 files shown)


pebrc (commented):
Sorry for the late review. I am in favour of merging this as the web hooks add a lot of usability, so it is good that we now have a way of supporting them when multiple operators are installed. 

The only two questions I have are: 
* how are we going to advertise this feature. Were you planning to add something to our documentation?
* do we need a warning in the docs for the case when users install both restricted operators and a global operator or do you think the errors will be obvious (this problem of exists of course independently of this change) ?

charith-elastic:
> * how are we going to advertise this feature. Were you planning to add something to our documentation?

I was not quite sure how to explain the use case. Because it's a Helm-only feature, I was thinking that the users would inspect the `values.yaml` file and decide whether they want to use it or not. Do you think we should document it?

> * do we need a warning in the docs for the case when users install both restricted operators and a global operator or do you think the errors will be obvious (this problem of exists of course independently of this change) ?

I haven't tried doing this but I assume the two operators would compete with each other for reconciling all resources? The webhook is named after the operator name and namespace, so there wouldn't be a clash there at least.

charith-elastic:
I tried installing a restricted operator and a global operator. It needs to be a conscious decision because one of the installations needs to skip installing CRDs. If they still go ahead with it, the two operators will compete with each other for reconciliation. I don't think we can do much to avoid that situation.

pebrc (approved):
LGTM I still think docs would be good but we can add them separately if we are unsure right now how to best document it.

charith-elastic:
Thanks. I created #4028 to follow up with the documentation.

Issue #3867, opened 2020-10-22
Cluster can be downscaled while shards are not migrated


barkbay:
When a Pod is deleted/restarting:
* the shards on the node enter the `UNASSIGNED` state
* the `node` field is set to an empty string

```json
{
		"index": "logs-191998",
		"shard": "4",
		"state": "UNASSIGNED",
		"node": "",
		"prirep": "p"
}
```

The empty string prevents [ECK](https://github.com/elastic/cloud-on-k8s/blob/ecb2f41cd5d38e2149e5e91649bc8b0a31856449/pkg/controller/elasticsearch/migration/migrate_data.go#L19) to detect that shards might actually still be assigned to a node which is downscaled (because `shard.NodeName` is empty) :

```golang
// NodeHasShard returns true if the given ES Pod is holding at least one shard (primary or replica).
func NodeHasShard(ctx context.Context, shardLister esclient.ShardLister, podName string) (bool, error) {
	[...]
	// filter shards affected by node removal
	for _, shard := range shards {
		if shard.NodeName == podName { // <--- Here
			return true, nil
		}
	}
	return false, nil
}
```

If Pods are deleted (by the controller itself while upgrading as described in https://github.com/elastic/cloud-on-k8s/issues/3861) or evicted (K8S node maintenance), ECK might then downscale and delete a set of nodes while they are still hosting some shards.

Note that it is not possible to know to which node "name" a shard was assigned, only its `uid` is stored in the shard metadata:
```json
{
	"shard": "0",
	"prirep": "p",
	"state": "UNASSIGNED",
	"unassigned.reason": "NODE_LEFT",
	"node": null,
	"id": null,
	"ud": "node_left [fMeKGb_PSi-DIGR_qPTgQQ]"
}
```

anyasabo:
Is the right fix for this to add another predicate to not scale down if there are unassigned shards? It seems like it might be too conservative but I'm not sure. Maybe it just makes sense to filter for unassigned.reason == NODE_LEFT? But I don't know if there's other states where a shard is unassigned and removing a node might cause data loss.

sebgl:
I think I'm +1 with your suggestion @anyasabo: don't downscale if there is any `UNASSIGNED` shard.

To be less conservative we could filter on `NODE_LEFT, keep a `node name -> node id` mapping in memory and parse the node ID from `unassigned.reason`, but I feel like that's not needed. It feels also a bit unreliable since there's no schema guarantee on that API.
[...]

PR #3883, opened 2020-10-28
Don't allow downscales if some shards are unassigned

sebgl:
In some conditions, for example when a Pod gets killed/restarted right
before a downscale happens, the shards on that Pod is reported as UNASSIGNED.
At this point it is dangerous to definitely remove the Pod (downscale the cluster)
since we can't know for sure the Pod to remove isn't supposed to hold any of the
unassigned shards.

To avoid that situation, this commit disallows any downscale to happen if
some of the shards don't have a node assigned to them (regardless of their status -
unassigned or not). The logic to allow a node to be downscaled is rather simple:
- all shards must have a node assigned to them
- the pod to remove must not have a shard assigned to it

This is a rather conservative/safe approach that could be optimized in the future.

I tried implementing an e2e test for this, but it's a bit tricky to setup the right way
so it consistently fails without this commit. I'm considering the unit test is good enough.

Fixes https://github.com/elastic/cloud-on-k8s/issues/3867.

### DIFF
--- a/pkg/controller/elasticsearch/migration/migrate_data.go
+++ b/pkg/controller/elasticsearch/migration/migrate_data.go
@@ -17,7 +17,7 @@ var log = logf.Log.WithName("migrate-data")
 
 // NodeMayHaveShard returns true if one of those condition is met:
 // - the given ES Pod is holding at least one shard (primary or replica)
-// - some shards in the cluster don't have a node assigned, in which case we can't make sure about the 1st condition
+// - some shards in the cluster don't have a node assigned, in which case we can't be sure about the 1st condition
 //   this may happen if the node was just restarted: the shards it is holding appear unassigned
 func NodeMayHaveShard(ctx context.Context, es esv1.Elasticsearch, shardLister esclient.ShardLister, podName string) (bool, error) {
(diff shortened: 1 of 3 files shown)


pkg/controller/elasticsearch/migration/migrate_data.go
```diff
@@ -15,17 +15,27 @@ import (
 
 var log = logf.Log.WithName("migrate-data")
 
-// NodeHasShard returns true if the given ES Pod is holding at least one shard (primary or replica).
-func NodeHasShard(ctx context.Context, shardLister esclient.ShardLister, podName string) (bool, error) {
+// NodeMayHaveShard returns true if one of those condition is met:
+// - the given ES Pod is holding at least one shard (primary or replica)
+// - some shards in the cluster don't have a node assigned, in which case we can't make sure about the 1st condition
```
anyasabo, line 20:
```suggestion
// - some shards in the cluster don't have a node assigned, in which case we can't be sure about the 1st condition
```

anyasabo (approved):

sebgl:
run full pr build

PR #3904, opened 2020-11-04
Fix broken link to elasticsearch heap size docs (v1.0)

kilfoyle:
This pre-emptively fixes a link to the Elasticsearch heap size docs that will break with the stack 7.10 release, and will prevent docs builds from passing CI checks.


charith-elastic (commented):
I think this is changing more than it should. Furthermore, isn't this a duplicate of #3902?

Issue #3909, opened 2020-11-06
Misleading E2E failure (ES_data_should_pass_validations) when Beat is down, because Kibana isn't available, because Elasticsearch isn't started, because AKS is slow


sebgl:
https://devops-ci.elastic.co/job/cloud-on-k8s-e2e-tests-aks/518/testReport/ reports an `ES_data_should_pass_validations` Beat E2E test failure:

```
Retries (10m0s timeout): ...........................................................................................................................................................
    utils.go:84: 
        	Error Trace:	utils.go:84
        	Error:      	Received unexpected error:
        	            	hit count should be more than 0 for /*beat*/_search?q=agent.type:metricbeat
        	Test:       	TestMetricbeatHostsRecipe/ES_data_should_pass_validations
{"log.level":"error","@timestamp":"2020-10-30T02:12:54.670Z","message":"stopping early",
```

The test stopped at 02:12.

Further investigation reveals that:

1. All 3 Metricbeat Pods are actually down. They stopped themselves at 02:09:00 because they could not reach Kibana:
```
2020-10-30T02:09:00.277Z	INFO	instance/beat.go:447	metricbeat stopped.
2020-10-30T02:09:00.277Z	ERROR	instance/beat.go:951	Exiting: error connecting to Kibana: fail to get the Kibana version: HTTP GET request to https://kibana-k5sv-kb-http.e2e-hrbd3-mercury.svc:5601/api/status fails: fail to execute the HTTP GET request: Get "https://kibana-k5sv-kb-http.e2e-hrbd3-mercury.svc:5601/api/status": dial tcp 10.0.62.48:5601: connect: connection refused. Response: .
```
They restarted 6 times already, and entered a 5m back-off restart period:

```json
    "restartCount": 6,
                        "started": false,
                        "state": {
                            "waiting": {
                                "message": "back-off 5m0s restarting failed container=metricbeat pod=metricbeat-k5sv-beat-metricbeat-2w2g5_e2e-hrbd3-mercury(2e70f41d-3b94-48bd-b0ca-656afec5f213)",
                                "reason": "CrashLoopBackOff"
                            }
````

2. Why can't Beat reach Kibana? Because Kibana itself could not reach Elasticsearch from 02:03 to 02:11:

```
[...]

PR #3910, opened 2020-11-06
Sort builders by test priority in Beat recipes E2E tests

sebgl:
In those tests we processed the builders in the order they are parsed
(Beat first, then Elasticsearch & Kibana).

In some cases a Beats test fails because Beats is waiting for Elasticsearch
to be available. The test failure reports a Beats problem, but the real
problem is on Elasticsearch side, which is misleading.

This fixes it by reordering builders by priority after parsing them
from recipe files.

Fixes https://github.com/elastic/cloud-on-k8s/issues/3909.

### DIFF
--- a/test/e2e/test/helper/yaml.go
+++ b/test/e2e/test/helper/yaml.go
@@ -10,6 +10,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"sort"
 	"strings"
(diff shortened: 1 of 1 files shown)


barkbay (approved):
lgtm

david-kow (approved):
LGTM

Issue #3917, opened 2020-11-06
Flaky TestMutationSecondMasterSet


sebgl:
https://devops-ci.elastic.co/job/cloud-on-k8s-e2e-tests-stack-versions/251/testReport/

```
=== RUN   TestMutationSecondMasterSet/ES_nodes_topology_should_eventually_be_the_expected_one#01
Retries (5m0s timeout): ..................................................................................................
    utils.go:84: 
        	Error Trace:	utils.go:84
        	Error:      	Received unexpected error:
[...]

PR #3919, opened 2020-11-06
Better error message on E2E CheckESNodesTopology errors

sebgl:
There are 2 reasons why  some nodes reported from the Elasticsearch cluster cannot be
matched to expected nodes in the spec:

1. Those nodes are being downscaled (not in the spec anymore), should be removed soon
2. Those nodes do not match the spec (different config, name, memory limit)

In the 2nd case, the error we report is rather generic. We print the expected topology
that hasn't been matched to the actual nodes: `expected elements missing from cluster`.

This commit adds an additional error message earlier in the process to report more details
about the actual node that cannot be matched to the spec (name, roles, memory limit).

Would have helped troubleshooting https://github.com/elastic/cloud-on-k8s/issues/3917.


barkbay (approved):
lgtm

PR #3922, opened 2020-11-07
Capitalize operator display name in OLM template

david-kow:


barkbay (approved):
I didn't find a guideline but looking at the other operators it seems to be common on the Operator Hub ("[Title case](https://en.wikipedia.org/wiki/Title_case)" I guess ?)

PR #3941, opened 2020-11-16
Add a note about upgrading from beta in upgrade instructions

david-kow:
Without this addition, one might think that even if beta version does not change the upgrade is affected by the bug mentioned.


sebgl (approved):
LGTM

Issue #3935, opened 2020-11-12
Flaky TestVersionUpgradeToLatest7x/Kibana_deployment_be_rolled_out


sebgl:
Happened twice in 2 days: https://devops-ci.elastic.co/job/cloud-on-k8s-e2e-tests-stack-versions/258/testReport/

```
=== RUN   TestVersionUpgradeToLatest7x/Kibana_deployment_be_rolled_out
Retries (5m0s timeout): ....................................................................................................
    utils.go:84: 
        	Error Trace:	utils.go:84
        	Error:      	Received unexpected error:
        	            	pod not running yet
        	Test:       	TestVersionUpgradeToLatest7x/Kibana_deployment_be_rolled_out
```

1/3 Kibana Pods are ready, the other 2 are still Pending:

```
 "status": {
                "availableReplicas": 1,
                "conditions": [
                    {
                        "lastTransitionTime": "2020-11-12T04:27:40Z",
                        "lastUpdateTime": "2020-11-12T04:27:40Z",
                        "message": "Deployment does not have minimum availability.",
                        "reason": "MinimumReplicasUnavailable",
                        "status": "False",
                        "type": "Available"
                    },
                    {
                        "lastTransitionTime": "2020-11-12T04:27:40Z",
                        "lastUpdateTime": "2020-11-12T04:28:27Z",
                        "message": "ReplicaSet \"test-version-upgrade-to-7x-6mhl-kb-69d4cc5459\" is progressing.",
                        "reason": "ReplicaSetUpdated",
                        "status": "True",
                        "type": "Progressing"
                    }
                ],
                "observedGeneration": 1,
                "readyReplicas": 1,
                "replicas": 3,
                "unavailableReplicas": 2,
                "updatedReplicas": 3
            }
```

```
"phase": "Pending",
                "qosClass": "Burstable",
                "startTime": "2020-11-12T04:27:40Z"
```
```
               "phase": "Pending",
                "qosClass": "Burstable",
                "startTime": "2020-11-12T04:27:40Z"
```

I think this is related to a slow image pull. Looking at the events for one of those Pods:

```
   "kind": "Event",
            "lastTimestamp": "2020-11-12T04:27:41Z",
            "message": "Pulling image \"docker.elastic.co/kibana/kibana:7.2.1\"",
            "metadata": {
                "creationTimestamp": "2020-11-12T04:27:41Z",
[...]

PR #3943, opened 2020-11-16
Set default e2e test timeout to 30 min

sebgl:
Increase the default timeout from 5min to 30min.
In some environment, that timeout was already overridden to 10min. This commit removes
the environment-specific overrides to use 30min everywhere.

Fixes https://github.com/elastic/cloud-on-k8s/issues/3935.


sebgl:
Actually I'm thinking we should make this 30min directly. (Otherwise I feel like we'd do it at the next flakiness 😓).

pebrc (approved):

PR #3950, opened 2020-11-17
Use new node_names query param for voting exclusions as of 7.8.0

### LINKED ISSUE #2951: Set node names in the voting config exclusions API call starting 7.8.0
https://github.com/elastic/elasticsearch/pull/55291 introduces a way to specify the node names we want to exclude: `POST /_cluster/voting_config_exclusions?node_names=.... `.

We can rely on it starting 7.8.0, and we must use it starting 8.0.0.

Relates https://github.com/elastic/cloud-on-k8s/issues/1986.

pebrc:
The new query parameter tolerates non-existing nodes and covers therefore the case where we try to remove a master node that has never successfully joined the cluster. This prevents an orchestration dead lock where the operator was previously unable to make progress due to an API error returned by Elasticsearch for nodes that were not part of the cluster.

Fixes #2951

### DIFF
--- a/pkg/controller/elasticsearch/client/v7.go
+++ b/pkg/controller/elasticsearch/client/v7.go
@@ -47,6 +47,7 @@ func (c *clientV7) AddVotingConfigExclusions(ctx context.Context, nodeNames []st
 	if c.version.IsSameOrAfter(version.From(7, 8, 0)) {
 		path = fmt.Sprintf("/_cluster/voting_config_exclusions?node_names=%s", strings.Join(nodeNames, ","))
 	} else {
+		// versions < 7.8.0 or unversioned clients which is OK as this deprecated API will be supported until 8.0
 		path = fmt.Sprintf("/_cluster/voting_config_exclusions/%s", strings.Join(nodeNames, ","))
 	}
 
--- a/pkg/controller/elasticsearch/client/v7.go
+++ b/pkg/controller/elasticsearch/client/v7.go
@@ -10,6 +10,7 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/elastic/cloud-on-k8s/pkg/controller/common/version"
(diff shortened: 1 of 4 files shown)


pkg/controller/elasticsearch/client/v7.go
```diff
@@ -42,7 +43,12 @@ func (c *clientV7) StartTrial(ctx context.Context) (StartTrialResponse, error) {
 }
 
 func (c *clientV7) AddVotingConfigExclusions(ctx context.Context, nodeNames []string) error {
-	path := fmt.Sprintf("/_cluster/voting_config_exclusions/%s", strings.Join(nodeNames, ","))
+	var path string
+	if c.version.IsSameOrAfter(version.From(7, 8, 0)) {
+		path = fmt.Sprintf("/_cluster/voting_config_exclusions?node_names=%s", strings.Join(nodeNames, ","))
+	} else {
+		path = fmt.Sprintf("/_cluster/voting_config_exclusions/%s", strings.Join(nodeNames, ","))
```
sebgl, line 50: If `c.version` is empty (client constructed from `client.NewElasticsearchClient()`), we end up in this `else` path, which is fine.
Maybe worth explicitly pointing out in a comment though?

sebgl (approved):
LGTM

pkg/controller/elasticsearch/client/v7.go
pebrc: Not sure I follow your argument wrt to `client.NewElasticsearchClient()` but I added a comment.

sebgl: I meant client which do not have a version set, which I think is the case when built from `client.NewElasticsearchClient()`.
I'm 👍 with the comment.

sebgl (approved):

PR #3951, opened 2020-11-17
Update Openshift compatibility note

barkbay:
Docker compatibility issues in the context of Openshift are mentioned in a [dedicated OpenShift section](https://github.com/elastic/cloud-on-k8s/blob/master/docs/advanced-topics/openshift.asciidoc).

This PR also adds a note about Enterprise Search 7.10


sebgl (approved):
LGTM

PR #3954, opened 2020-11-19
[1.3] Use new node_names query param for voting exclusions as of 7.8.0 (#3950)

pebrc:
 - Use new node_names query param for voting exclusions as of 7.8.0 (#3950)


david-kow:
Should we create a temporary `1.3.1` branch and merge it later? Some changes (docs) shouldn't be merged to `1.3` just yet so defaulting to a new branch could be easier. Having said that, we might not have any more changes for `1.3.1` so I'm fine with leaving it here as well.

david-kow (approved):

Issue #3956, opened 2020-11-19
Panic at the end of e2e tests preventing diag bundle to be created


sebgl:
We don't have diagnostic bundles in https://devops-ci.elastic.co/blue/organizations/jenkins/cloud-on-k8s-e2e-tests-stack-versions/detail/cloud-on-k8s-e2e-tests-stack-versions/262/pipeline.

I think the reason is that the e2e test program panics before we get to the bundle part:

```
[2020-11-19T02:00:49.272Z] E1119 02:00:48.954564   15944 runtime.go:78] Observed a panic: "close of closed channel" (close of closed channel)

[2020-11-19T02:00:49.272Z] goroutine 76 [running]:

[2020-11-19T02:00:49.272Z] k8s.io/apimachinery/pkg/util/runtime.logPanic(0x17ee780, 0x1bceb30)

[2020-11-19T02:00:49.272Z] 	/go/pkg/mod/k8s.io/apimachinery@v0.18.6/pkg/util/runtime/runtime.go:74 +0xa6

[2020-11-19T02:00:49.272Z] k8s.io/apimachinery/pkg/util/runtime.HandleCrash(0x0, 0x0, 0x0)

[2020-11-19T02:00:49.272Z] 	/go/pkg/mod/k8s.io/apimachinery@v0.18.6/pkg/util/runtime/runtime.go:48 +0x89

[2020-11-19T02:00:49.272Z] panic(0x17ee780, 0x1bceb30)

[2020-11-19T02:00:49.272Z] 	/usr/local/go/src/runtime/panic.go:969 +0x1b9

[2020-11-19T02:00:49.272Z] github.com/elastic/cloud-on-k8s/test/e2e/cmd/run.(*Job).WaitForLogs(0xc0003cf260)

[2020-11-19T02:00:49.272Z] 	/go/src/github.com/elastic/cloud-on-k8s/test/e2e/cmd/run/job.go:58 +0x37

[2020-11-19T02:00:49.272Z] github.com/elastic/cloud-on-k8s/test/e2e/cmd/run.(*JobsManager).Start.func3(0x19cee00, 0xc0003c8c00, 0x19cee00, 0xc0003c8c00)

[2020-11-19T02:00:49.272Z] 	/go/src/github.com/elastic/cloud-on-k8s/test/e2e/cmd/run/job_manager.go:140 +0x3fa

[...]

PR #3957, opened 2020-11-19
Don't close the log stream channel twice in e2e tests

sebgl:
We may receive multiple time the PodSucceeded (or PodFailed)  event, which leads us to wait
for Pod logs multiple times. However this step involves closing a channel (no need to stream
logs anymore), and causing a channel twice causes a Panic().
This commit adds an early check in the waitForLogs() function to just return if it has
been executed already. The Stop() function already did the same thing.

Fixes https://github.com/elastic/cloud-on-k8s/issues/3956.


barkbay (approved):
lgtm

PR #3959, opened 2020-11-19
Update stack versions in test pipelines

barkbay:
This PR updates the stack versions in our e2e test pipelines. It also removes the pipeline for 7.1.x, according to our [EOL policy](https://www.elastic.co/support/eol) that version is no more supported since 2020-11-20.


sebgl (approved):
LGTM

Issue #3958, opened 2020-11-19
All-in-one installation using k8s 1.19 fails because of a removed k8s API (v1beta1/ValidatingWebhookConfiguration)


maorkh4:
Deploying ECK operator 1.3.0 using the all-in-one YAML with k8s 1.19 fails because it tries to use the removed `admissionregistration.k8s.io/v1beta1/ValidatingWebhookConfiguration` API. The error message is:
```
kubernetes:admissionregistration.k8s.io/v1beta1:ValidatingWebhookConfiguration (elastic-webhook.k8s.elastic.co):
  error: apiVersion "admissionregistration.k8s.io/v1beta1/ValidatingWebhookConfiguration" was removed in Kubernetes 1.19. Use "admissionregistration.k8s.io/v1/ValidatingWebhookConfiguration" instead.
  See https://git.k8s.io/kubernetes/CHANGELOG/CHANGELOG-1.19.md#deprecation-1 for more information.
```

`kubectl version` returns:
```
Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.3", GitCommit:"1e11e4a2108024935ecfcb2912226cedeafd99df", GitTreeState:"clean", BuildDate:"2020-10-14T12:50:19Z", GoVersion:"go1.15.2", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.3", GitCommit:"1e11e4a2108024935ecfcb2912226cedeafd99df", GitTreeState:"clean", BuildDate:"2020-10-14T12:41:49Z", GoVersion:"go1.15.2", Compiler:"gc", Platform:"linux/amd64"}
```

Downgrading to k8s 1.18.8 fixes the problem but produces the warning:
```
kubernetes:admissionregistration.k8s.io/v1beta1:ValidatingWebhookConfiguration (elastic-webhook.k8s.elastic.co):
  warning: admissionregistration.k8s.io/v1beta1/ValidatingWebhookConfiguration is deprecated by admissionregistration.k8s.io/v1/ValidatingWebhookConfiguration and not supported by Kubernetes v1.19+ clusters.
```
An issue was already opened regarding the warning on k8s 1.16+ (#3886).

alongubkin:
This error prevents us from using K8s 1.19

sebgl:
[...]

Issue #3886, opened 2020-10-30
Deprecation warnings shown when installing all-in-one (for K8s 1.16+)


nkvoll:
Not quite sure which at which version the deprecations show up as warnings.

Example (first and last lines showing the warnings and migration path forward).

```
$ kubectl apply -f https://download.elastic.co/downloads/eck/1.2.1/all-in-one.yaml
Warning: apiextensions.k8s.io/v1beta1 CustomResourceDefinition is deprecated in v1.16+, unavailable in v1.22+; use apiextensions.k8s.io/v1 CustomResourceDefinition
customresourcedefinition.apiextensions.k8s.io/apmservers.apm.k8s.elastic.co configured
customresourcedefinition.apiextensions.k8s.io/beats.beat.k8s.elastic.co configured
customresourcedefinition.apiextensions.k8s.io/elasticsearches.elasticsearch.k8s.elastic.co configured
customresourcedefinition.apiextensions.k8s.io/enterprisesearches.enterprisesearch.k8s.elastic.co configured
customresourcedefinition.apiextensions.k8s.io/kibanas.kibana.k8s.elastic.co configured
namespace/elastic-system unchanged
serviceaccount/elastic-operator unchanged
secret/elastic-webhook-server-cert unchanged
clusterrole.rbac.authorization.k8s.io/elastic-operator unchanged
clusterrole.rbac.authorization.k8s.io/elastic-operator-view unchanged
clusterrole.rbac.authorization.k8s.io/elastic-operator-edit unchanged
clusterrolebinding.rbac.authorization.k8s.io/elastic-operator unchanged
rolebinding.rbac.authorization.k8s.io/elastic-operator unchanged
service/elastic-webhook-server unchanged
statefulset.apps/elastic-operator configured
Warning: admissionregistration.k8s.io/v1beta1 ValidatingWebhookConfiguration is deprecated in v1.16+, unavailable in v1.22+; use admissionregistration.k8s.io/v1 ValidatingWebhookConfiguration
validatingwebhookconfiguration.admissionregistration.k8s.io/elastic-webhook.k8s.elastic.co configured
```

jamiechapmanbrn:
Next kubernetes version to release outright removes apiextensions.k8s.io/v1beta1 CustomResourceDefinitions, has there been any progress on this?

pebrc:
[...]

PR #3963, opened 2020-11-19
[Helm] Use ValidatingWebhookConfiguration v1 on supported Kubernetes versions

charith-elastic:
Uses `v1` of `ValidatingWebhookConfiguration` if Kubernetes version is >= 1.16.

Includes a change to explicitly set the Kubernetes version in manifest-gen mode so that we can generate `all-in-one.yaml` files targeting different minimum Kubernetes versions by setting the `internal.kubeVersion` value.

Partial fix for #3958 and #3886.


charith-elastic:
Jenkins test this please

deploy/eck-operator/templates/_helpers.tpl
```diff
@@ -67,17 +67,53 @@ Create the name of the service account to use
...
+
+
+{{/*
+Define admissionReviewVersions based on Kubernetes version
+*/}}
+{{- define "eck-operator.webhookAdmissionReviewVersions" -}}
+{{- $kubeVersion := (include "eck-operator.effectiveKubeVersion" .) -}}
+{{- $kubeVersionSupported := semverCompare ">=1.16.0-0" $kubeVersion -}}
+{{- if $kubeVersionSupported  }}
+admissionReviewVersions: [v1beta1, v1]
```
pebrc, line 113: Maybe I am misunderstanding how this works but the operator does not currently offer a v1 webhook. Should we not just list `v1beta1` here?

pebrc (commented):
LGTM!

deploy/eck-operator/templates/_helpers.tpl
charith-elastic: I kept it that way because that's what Kubebuilder generates (even though it doesn't v1 either). But, you're right. It's probably confusing. I'll remove it.

pebrc (approved):

PR #3965, opened 2020-11-20
Allow operatorhub tool to generate multiple packages

david-kow:
Moving to a single config and single invocation to create all the packages. This avoids error-prone, manual changes to `config.yaml`. Added support for ubi-only, distribution-channel and output path to the config file.


charith-elastic (approved):
LGTM

Issue #3970, opened 2020-11-23
K8s API network errors during E2E Elasticsearch health checks cause test failure


sebgl:
https://devops-ci.elastic.co/job/cloud-on-k8s-e2e-tests-aks/540/testReport/

```
steps_mutation.go:119: Elasticsearch cluster health check failure at 2020-11-23 16:19:46.968009788 +0000 UTC m=+5665.163135094: Get "https://10.0.0.1:443/api/v1/namespaces/e2e-z2i7j-mercury/secrets/test-mutation-2nd-master-set-pld5-es-elastic-user": dial tcp 10.0.0.1:443: connect: connection refused
```

We run continuous health checks during a mutation to ensure the cluster is still available and not red.
In that case one call to the k8s API failed, which caused the entire test to fail.

sebgl:
```go
			case <-ticker.C:
				// recreate the Elasticsearch client at each iteration, since we may have switched protocol from http to https during the mutation
				client, err := hc.esClientFactory()
				if err != nil {
					// treat client creation failure same as unavailable cluster
					hc.AppendErr(err)
					continue
				}
```

We treat client creation failures as unavailable cluster, but client creation only involves interacting with k8s.
I'm wondering if it's fine to just ignore client creation errors and `continue` here.

PR #3971, opened 2020-11-23
Ignore network errors in e2e healthcheck apiserver calls

sebgl:
During the healthcheck function, we register an error when we fail to create
the Elasticsearch client. That error can either be:
- an explicit apiserver error (e.g. secret containing password does not exist)
- a network error (dns resolving problem, timeout, connection refused, etc.)

In the 2nd case, I think we just want to retry at the next healthcheck iteration,
and don't fail the entire test.

Fixes https://github.com/elastic/cloud-on-k8s/issues/3970.

### DIFF
--- a/test/e2e/test/elasticsearch/steps_mutation.go
+++ b/test/e2e/test/elasticsearch/steps_mutation.go
@@ -7,6 +7,7 @@ package elasticsearch
 import (
 	"context"
 	"errors"
+	"fmt"
 	"testing"
 	"time"
 
@@ -17,6 +18,7 @@ import (
 	"github.com/elastic/cloud-on-k8s/test/e2e/test"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


test/e2e/test/elasticsearch/steps_mutation.go
```diff
@@ -192,8 +194,16 @@ func (hc *ContinuousHealthCheck) Start() {
 				// recreate the Elasticsearch client at each iteration, since we may have switched protocol from http to https during the mutation
 				client, err := hc.esClientFactory()
 				if err != nil {
-					// treat client creation failure same as unavailable cluster
-					hc.AppendErr(err)
+					// according to https://github.com/kubernetes/client-go/blob/fb61a7c88cb9f599363919a34b7c54a605455ffc/rest/request.go#L959-L960,
+					// client-go requests may return *errors.StatusError or *errors.UnexpectedObjectError, or http client errors.
```
barkbay, line 198: Intuitively I think we would like to catch only a small set of known errors _(mostly the one you mentioned in the description: timeout, connection refused and dns related errors)_ but letting all the other ones bubble up. Could you add a comment to explain why we are doing the opposite here ?

sebgl: I think the only reason I did it this way is that those errors [are not trivial to catch](https://stackoverflow.com/questions/22761562/portable-way-to-detect-different-kinds-of-network-error-in-golang), catching expected api errors is easier. 

@barkbay do you think we should change the approach to indeed catch those network errors explicitly, or is adding a comment explaining the reverse approach enough?

test/e2e/test/elasticsearch/steps_mutation.go
```diff
@@ -192,8 +194,16 @@ func (hc *ContinuousHealthCheck) Start() {
...
+					switch err.(type) {
+					case *k8serrors.StatusError, *k8serrors.UnexpectedObjectError:
+						// explicit apiserver error, consider as healthcheck failure
+						hc.AppendErr(err)
+					default:
+						// likely a network error, log and ignore
+						fmt.Printf("error while creating the Elasticsearch client: %s", err.Error())
```
barkbay, line 205: nit: I _think_ `.Error()` is not required here, according to https://golang.org/pkg/fmt/

> 4. If an operand implements the error interface, the Error method will be invoked to convert the object to a string, which will then be formatted as required by the verb (if any).

test/e2e/test/elasticsearch/steps_mutation.go
barkbay: >  or is adding a comment explaining the reverse approach enough

👍 , I think I'm fine with your proposal, I would just add a comment to understand later why we made this choice.

barkbay (approved):
LGTM

david-kow (approved):

PR #3972, opened 2020-11-24
Update accessing-elastic-services.asciidoc

Leaf-Lin:
Change data loss `note` to `warning`. As this could cause data loss, I think it requires a little bit more attention.

### DIFF
--- a/docs/orchestrating-elastic-stack-applications/accessing-elastic-services.asciidoc
+++ b/docs/orchestrating-elastic-stack-applications/accessing-elastic-services.asciidoc
@@ -29,7 +29,7 @@ To access Elastic resources, the operator manages a default user named `elastic`
 42xyz42citsale42xyz42
 ----
 
-NOTE: Beware of copying this Secret as-is into a different namespace. See <<{p}-common-problems-owner-refs, Common Problems: Owner References>> for more information.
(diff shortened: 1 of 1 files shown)


barkbay (commented):
Should we clearly state that copying this Secret as-is may result in data loss ?

Also I think there are 2 other places in the documentation where a user might think it is a good idea to duplicate a Secret:

* [Retrieve the CA certificate](https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-request-elasticsearch-endpoint.html#k8s-request-elasticsearch-endpoint) within the same cluster.

* [Default self-signed certificate](https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-tls-certificates.html#k8s-default-self-signed-certificate): "The CA, the certificate and the private key are each stored in a separate Secret.", which could also be interpreted as an invitation to copy this `Secret`

I'm not sure about the later, but I think the former would deserve the same warning.

eedugon:
@barkbay , besides saying that copying the secret `as-is` may result in data loss I would share an example of how exactly to copy it (reuse the content, not the secret itself) in the recommended way to avoid the k8s bug.

charith-elastic:
The issue with copying secrets has been mitigated in ECK 1.3.1 so I am closing this PR.

Issue #3942, opened 2020-11-16
Doc overview page is outdated


david-kow:
At least [it](https://www.elastic.co/guide/en/cloud-on-k8s/1.3/k8s-overview.html)'s not listing all supported products, maybe the feature list could also use a refresh.

david-kow:
Fixed in #3974.

PR #3974, opened 2020-11-25
Update ECK overview

alaudazzi:
This PR proposes a slight refresh of the overview page.

Relates: #3942

### DIFF
--- a/docs/overview.asciidoc
+++ b/docs/overview.asciidoc
@@ -15,7 +15,6 @@ With Elastic Cloud on Kubernetes you can streamline critical operations, such as
 
 . Managing and monitoring multiple clusters
 . Scaling cluster capacity and storage
-. Configure backups using snapshots
 . Securing clusters with TLS certificates
 . Setting up hot-warm-cold architectures with availability zone awareness
 --
--- a/docs/overview.asciidoc
+++ b/docs/overview.asciidoc
@@ -15,6 +15,7 @@ With Elastic Cloud on Kubernetes you can streamline critical operations, such as
 
 . Managing and monitoring multiple clusters
 . Scaling cluster capacity and storage
+. Performing safe configuration changes through rolling upgrades
 . Securing clusters with TLS certificates
 . Setting up hot-warm-cold architectures with availability zone awareness
 --
--- a/docs/overview.asciidoc
+++ b/docs/overview.asciidoc
@@ -14,7 +14,7 @@ Built on the Kubernetes Operator pattern, Elastic Cloud on Kubernetes (ECK) exte
 With Elastic Cloud on Kubernetes you can streamline critical operations, such as:
 
 . Managing and monitoring multiple clusters
-. Scaling cluster capacity and volume
+. Scaling cluster capacity and storage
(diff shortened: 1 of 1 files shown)


docs/overview.asciidoc
```diff
@@ -9,14 +9,13 @@ endif::[]
...
+Built on the Kubernetes Operator pattern, Elastic Cloud on Kubernetes (ECK) extends the basic Kubernetes orchestration capabilities to support the setup and management of Elasticsearch, Kibana, APM Server, Enterprise Search, and Beats on Kubernetes.
 
-With Elastic Cloud on Kubernetes you can streamline all those critical operations, such as:
+With Elastic Cloud on Kubernetes you can streamline critical operations, such as:
 
 . Managing and monitoring multiple clusters
-. Scaling cluster capacity up and down
-. Changing cluster configuration
-. Scheduling backups
+. Scaling cluster capacity and volume
```
sebgl, line 17: maybe `capacity and storage`?

docs/overview.asciidoc
```diff
@@ -9,14 +9,13 @@ endif::[]
...
+. Configure backups using snapshots
```
sebgl, line 18: I think I would maybe remove that item since we don't really have first-class support for snapshots. @elastic/cloud-k8s @shubhaat any different opinion?

docs/overview.asciidoc
```diff
@@ -9,14 +9,13 @@ endif::[]
...
 --
-Built on the Kubernetes Operator pattern, Elastic Cloud on Kubernetes (ECK) extends the basic Kubernetes orchestration capabilities to support the setup and management of Elasticsearch, Kibana and APM Server on Kubernetes.
+Built on the Kubernetes Operator pattern, Elastic Cloud on Kubernetes (ECK) extends the basic Kubernetes orchestration capabilities to support the setup and management of Elasticsearch, Kibana, APM Server, Enterprise Search, and Beats on Kubernetes.
 
-With Elastic Cloud on Kubernetes you can streamline all those critical operations, such as:
+With Elastic Cloud on Kubernetes you can streamline critical operations, such as:
 
 . Managing and monitoring multiple clusters
-. Scaling cluster capacity up and down
-. Changing cluster configuration
```
sebgl, line 18: I think we should maybe keep that one, slightly transformed into something like: `performing rolling upgrades for configuration change`

docs/overview.asciidoc
david-kow: Here we just picked a single feature of ES and we emphasise that we support it. I think we could either:
1. make a broader statement listing more "big" features (but I'm not sure what should go into that list, we would need to refresh it etc.), or
2. make a general statement that we support all Stack features (but I'm not sure if we can say that about every single feature/setting that ES supports), or
3. as @sebgl suggests, remove this

docs/overview.asciidoc
david-kow: or `performing safe configuration changes through rolling upgrades`?

sebgl (approved):
LGTM

Issue #3617, opened 2020-08-13
Flaky integration test: pkg/controller/license/TestReconcile


barkbay:
We got this issue at least twice, last time [here](https://devops-ci.elastic.co/job/cloud-on-k8s-pr/4313/console).

```
15:24:40  [31m✖[0m  pkg/controller/license (28.072s) (coverage: 61.2% of statements)
15:24:48  [32m✓[0m  pkg/controller/common/certificates (45.41s) (coverage: 80.7% of statements)
15:24:48  [33m∅[0m  pkg/controller/common/daemonset
15:24:48  [33m∅[0m  pkg/controller/common/driver
15:24:48  [33m∅[0m  pkg/controller/common/events
15:24:48  [33m∅[0m  pkg/controller/common/operator
15:24:48  [33m∅[0m  pkg/controller/common/pod
15:24:48  [33m∅[0m  pkg/controller/common/scheme
15:24:48  [33m∅[0m  pkg/controller/common/tracing
15:24:48  [33m∅[0m  pkg/controller/elasticsearch
15:24:48  [33m∅[0m  pkg/controller/elasticsearch/certificates
15:24:48  [33m∅[0m  pkg/controller/elasticsearch/client/test_fixtures
15:24:48  [33m∅[0m  pkg/controller/elasticsearch/configmap
15:24:48  [33m∅[0m  pkg/controller/elasticsearch/network
15:24:48  [33m∅[0m  pkg/controller/elasticsearch/volume
15:24:48  [33m∅[0m  pkg/controller/enterprisesearch/name
15:25:10  [32m✓[0m  pkg/controller/license/trial (55.977s) (coverage: 85.6% of statements)
15:25:10  [33m∅[0m  pkg/dev
15:25:10  [33m∅[0m  pkg/utils/compare
15:25:10  [33m∅[0m  pkg/utils/cryptutil
15:25:10  [33m∅[0m  pkg/utils/diff
15:25:10  [33m∅[0m  pkg/utils/log
15:25:10  [33m∅[0m  pkg/utils/metrics
15:25:10  [33m∅[0m  pkg/utils/net
15:25:10  [33m∅[0m  pkg/utils/pointer
15:25:10  [33m∅[0m  pkg/utils/test
15:25:10  [33m∅[0m  cmd
15:25:10  [33m∅[0m  cmd/licensing-info
15:25:10  [33m∅[0m  cmd/manager
15:25:10  
15:25:10  === [31mFailed[0m
[...]

PR #3977, opened 2020-11-25
Use a custom version of JSON marshalling for license verification

pebrc:
Standard Go json.Marshal escapes "problematic" characters >, <, and & which breaks license verification.

I also generated a test license with external tooling using the key pair we use for unit testing to create a unit test that verifies this works now.


pkg/controller/common/license/verifier.go
```diff
@@ -229,6 +229,22 @@ func (l EnterpriseLicense) SignableContentBytes() ([]byte, error) {
...
+	encoder := json.NewEncoder(buffer)
+	encoder.SetEscapeHTML(false)
+	err := encoder.Encode(t)
+	if err != nil {
+		return nil, err
+	}
+	marshaledBytes := buffer.Bytes()
+	// json.Encoder adds an additional newline between objects which we do not want here
+	//as it is not part of the signature, so we are removing the last byte
+	return marshaledBytes[:len(marshaledBytes)-1], err
```
sebgl, line 245: I'm pretty convinced the slice manipulation will almost never panic (unless empty content in the license?), however I think it'd still be safer to rely on `bytes.TrimSuffix()` here.

pkg/controller/common/license/verifier.go
```diff
@@ -229,6 +229,22 @@ func (l EnterpriseLicense) SignableContentBytes() ([]byte, error) {
...
+	buffer := &bytes.Buffer{}
+	encoder := json.NewEncoder(buffer)
+	encoder.SetEscapeHTML(false)
+	err := encoder.Encode(t)
+	if err != nil {
+		return nil, err
+	}
+	marshaledBytes := buffer.Bytes()
+	// json.Encoder adds an additional newline between objects which we do not want here
+	//as it is not part of the signature, so we are removing the last byte
```
sebgl, line 244: nit: missing whitespace after `//`

pkg/controller/common/license/verifier.go
pebrc: Ha I had that initially and though it was overkill. But you are right.

sebgl (approved):
LGTM

pebrc:
Jenkins test this please. 

https://github.com/elastic/cloud-on-k8s/issues/3617 hit again.

PR #3981, opened 2020-11-26
eck-dump: print Secrets metadata

barkbay:
If the logs of the operator are lost having the metadata on Secrets can be useful to understand when they have been (re)created or updated.


sebgl (approved):
LGTM

Issue #3983, opened 2020-11-26
Create/update apiserver requests network issues lead to e2e tests failures


sebgl:
https://devops-ci.elastic.co/job/cloud-on-k8s-e2e-tests-stack-versions/268/testReport/

This is a timeout on the requests to create an Elasticsearch resource in the apiserver.
I think we should just retry those.

PR #3984, opened 2020-11-26
Retry E2E test steps that involve network IO with the apiserver

sebgl:
Retry test steps involving network calls to the APIServer.
We occasionally get E2E tests failures for a network error when calling the apiserver.
One way to solve that is to retry those test steps, which is what this PR does.

Some bits are a bit tricky: for example when we create a resource, and that call fails for a network error, the resource could still very well be created in the apiserver. Which means the retried creation will fail (resource already exists) - which is something we need to deal with explicitly (accept `apierrors.IsAlreadyExists(err)`).

Fixes https://github.com/elastic/cloud-on-k8s/issues/3983.


sebgl:
run full pr job

sebgl:
run full pr build

barkbay (commented):
`partialAPMConfiguration` involves some API calls, I think we should also retry the following steps:
* https://github.com/elastic/cloud-on-k8s/blob/459d005f53eb500a11d237f262c1e891560f7e90/test/e2e/apm/configuration_test.go#L94
* https://github.com/elastic/cloud-on-k8s/blob/459d005f53eb500a11d237f262c1e891560f7e90/test/e2e/apm/configuration_test.go#L178

thbkrkr:
run full pr build

thbkrkr:
run full pr build

charith-elastic:
I wonder if we can avoid changing so many tests (making them longer and a bit harder to read) by simply changing the Kubernetes client to be a retrying client. As far as I can see, these tests are almost always using the client injected by the test framework --  which makes it easier to do that change in a single place.

sebgl:
@charith-elastic that's a great question, I asked myself the same while working on this PR.
Tests that consist in a single call to the apiserver which may fail for some random network issues could indeed definitely be solved by having retries handled by the client.

However there are several cases which are slightly more complicated:
* get + update pattern ([example](https://github.com/elastic/cloud-on-k8s/pull/3984/files#diff-3fdbd0dec0cd42f28661cf8d70846b7816addef79b8d4d9ec3b697ac0dcb8949R131-R138)). The call may fail thanks to optimistic concurrency (we're updating a resource with a resourceVersion older than what's in the apiserver), in which case retrying the update is useless, we want to re-get + re-update (retry the whole thing).
* eventual result pattern ([example](https://github.com/elastic/cloud-on-k8s/pull/3984/files#diff-9b3733d5ea2e91242031d9efb339331d820b0548e3251145d06600bf4cfe62a7R114-R120)). We want to retry until the apiserver result contains eg. only 1 Pod. If we were to use a client that handles the retries we'd have to pass it a condition on the result (doable, but not simpler?).
* retry all errors except error X pattern ([example](https://github.com/elastic/cloud-on-k8s/pull/3984/files#diff-9b3733d5ea2e91242031d9efb339331d820b0548e3251145d06600bf4cfe62a7R214)). Here we'd have to instruct the client to retry on most errors, except this particular one (doable, but not simpler)?
* create or update pattern ([example](https://github.com/elastic/cloud-on-k8s/pull/3984/files#diff-6c7fb869829501d769d5ff596b13a856578cdb3ec7f0bb192e4749fabf4802b5R332-R347)). Here we want to catch a specific error on creations, turn it into an update, but retry on other errors.

Overall my feeling was that, while a few of those could definitely be dealt with a (smart enough) retrying client, there are many edge cases that require some thinking. It felt simpler to use a higher-level retry scoping the entire test step.

charith-elastic:
Thanks for the detailed explanation @sebgl. That makes sense and 👍🏼 for consistency. 

Just out of curiosity, did you consider using the [builtin client-go retry helpers](https://pkg.go.dev/k8s.io/client-go/util/retry) to address some of those problems? Of course, they can't fix all of the problems you mentioned. But it feels like some could be, and I am just curious to know if you discovered any issues that are not quite obvious.

sebgl:
> Just out of curiosity, did you consider using the builtin client-go retry helpers to address some of those problems? Of course, they can't fix all of the problems you mentioned. But it feels like some could be, and I am just curious to know if you discovered any issues that are not quite obvious.

I didn't, thanks for mentioning I forgot about that one! It looks like both `OnError` and `RetryOnConflict` could deal with most of the patterns highlighted above.

Do you think we should modify that PR and try to use them? I think I'd be fine with keeping the current approach (less accurate, but maybe simpler and more consistent?).

charith-elastic:
My personal preference would be to see if these common chores can be built into the testing framework so that writing tests is as simple as possible without requiring a lot of scaffolding that detract from the test logic. However, I do understand that it's a lot of work so I am not opposed to merging the PR as-is.

pebrc (approved):
LGTM

PR #3991, opened 2020-11-30
Fix broken telemetry e2e test for 7.11.0-SNAPSHOT

pebrc:
We are relying on an internal debug API here. So breakage even between minor versions is to be expected:
Adjust the payload as of 7.11.0 to [no longer contain the timeRange attribute](https://github.com/elastic/kibana/pull/83791/files#diff-58fe00aa6374eb9c41c65f1799c779d71dc5d607bd425a782f8f9a8f66d8a9afL14).


david-kow (approved):
LGTM

Issue #3986, opened 2020-11-26
Work around Kubernetes bug causing garbage collection across namespaces to happen when ownerReferences are set


sebgl:
This [Kubernetes bug](https://github.com/kubernetes/kubernetes/issues/65200) leads to very dangerous situations for ECK users currently.

As we explained in [ECK documentation](https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-common-problems.html#k8s-common-problems-owner-refs), the following may happen:

- a user copies the Kubernetes secret holding the `elastic` user password from one namespace to another, in order to use it in other applications
- suddenly, Kubernetes garbage collection mechanism detects this copied secret, and decides to garbage collect the entire owner resource tree
- this means all PersistentVolumesClaims suddenly get deleted. If their storage class specify a `delete` retention policy, all Elasticsearch data gets immediately deleted. Even if they use the `retain` detention policy, it is likely that new PVCs will get created along with empty fresh PVs

This bug will be [fixed in Kubernetes 1.20](https://github.com/kubernetes/kubernetes/pull/92743).

In general, we try to not work around Kubernetes bugs in ECK, and instead encourage people to upgrade their Kubernetes version, but here the impact is dangerous enough to deserve being addressed in ECK directly.

I see several solutions to this problem:

**1. Don't set ownerReferences on Secrets user are likely to copy around**

I think there are 2 Secrets users can be tempted to copy across namespaces:
- the `elastic` user password
- the public-facing CA of Elasticsearch HTTP certificates

We could also potentially add secrets of other stack components to that list:
- Kibana public CA
- APM token
- etc.

Those additional secrets are also impacted by the GC bug, but will not cause Elasticsearch data loss.

In those secrets, we could decide to not set an ownerReference. Which means they will remain orphan Kubernetes resources if the user decides to delete the Elasticsearch cluster.
Overall it means replacing the ownerReference bug with another bug (not cleaning secrets on deletion), but that second bug is much less dangerous.

**2. Set the ownerReference of those Secrets to target a different, harmless resource**

To be experimented, not sure if it works.
We could decide to bind the ownerReference of those secrets to a different resource. And make that different resource have an ownerReference to the Elasticsearch resource. So the parent-child tree looks as follow:

```
Elasticsearch <-- PlaceHolderWorkaroundResource <-- Secret
```

[...]

PR #3992, opened 2020-11-30
Don't set an ownerRef on secrets users are susceptible to copy around

sebgl:
A k8s bug (https://github.com/kubernetes/kubernetes/issues/65200) may cause the
k8s garbage collection to delete undesired resources in case users manually
copy an operator-managed secret to another namespace.

To avoid that situation, this commit ensures no ownerRef is set on a subset of
managed secrets users are susceptible to copy around:
- the elastic user password secret
- elasticsearch public transport certs
- elasticsearch, kibana, enterprise search, apm server public http certs

Existing ownerReferences set with earlier ECK versions will be removed when
reconciled.

Since they do not have an ownerRef anymore, those secrets are not automatically
deleted when the Elasticsearch resource is deleted. To work around that situation,
the secret reconciliation logic adds an additional set of labels to the reconciled
secrets that don't have an ownerRef specified. These labels reference the "soft"
owner ("soft" as in handled through some custom code and not through k8s builtin
garbage collection logic).
Once a controller receives a deletion event for the resource it manages, it will
automatically remove the soft-owned secrets.

Additionally, garbage collection runs at operator startup to cover the case where
owners were deleted while the operator was down.

TODO:
- [x] Unit tests
- [x] E2E test
- [x] GC secrets on operator startup to cover the case where the operator is down during the owner deletion
- [x] Major bug to fix: the soft owner reference needs to include the object Kind, otherwise we may end up mismatching e.g. Elasticsearch stuff with Kibana stuff if they have the same name.

Fixes https://github.com/elastic/cloud-on-k8s/issues/3986.

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -614,7 +614,7 @@ func asyncTasks(
 		}()
 	}
 
-	// Garbage collect orphan secrets leftover from deleted resources while the operator was not running
+	// Garbage collect orphaned secrets leftover from deleted resources while the operator was not running
 	// - association user secrets
 	garbageCollectUsers(cfg, managedNamespaces)
 	// - soft-owned secrets
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -618,16 +618,7 @@ func asyncTasks(
 	// - association user secrets
 	garbageCollectUsers(cfg, managedNamespaces)
 	// - soft-owned secrets
-	if err := reconciler.GarbageCollectAllOrphanSecrets(k8s.WrapClient(mgr.GetClient()), map[string]runtime.Object{
-		esv1.Kind:        &esv1.Elasticsearch{},
-		apmv1.Kind:       &apmv1.ApmServer{},
-		kbv1.Kind:        &kbv1.Kibana{},
-		entv1beta1.Kind:  &entv1beta1.EnterpriseSearch{},
-		beatv1beta1.Kind: &beatv1beta1.Beat{},
-	}); err != nil {
-		log.Error(err, "Orphan secrets garbage collection failed, will be attempted again at next operator restart.")
-	}
-	log.Info("Orphan secrets garbage collection complete")
+	garbageCollectSoftOwnedSecrets(k8s.WrapClient(mgr.GetClient()))
 }
 
 func chooseAndValidateIPFamily(ipFamilyStr string, ipFamilyDefault corev1.IPFamily) (corev1.IPFamily, error) {
@@ -714,6 +705,20 @@ func garbageCollectUsers(cfg *rest.Config, managedNamespaces []string) {
 		log.Error(err, "user garbage collector failed")
 		os.Exit(1)
 	}
+}
+
+func garbageCollectSoftOwnedSecrets(client k8s.Client) {
+	if err := reconciler.GarbageCollectAllSoftOwnedOrphanSecrets(client, map[string]runtime.Object{
+		esv1.Kind:        &esv1.Elasticsearch{},
+		apmv1.Kind:       &apmv1.ApmServer{},
+		kbv1.Kind:        &kbv1.Kibana{},
+		entv1beta1.Kind:  &entv1beta1.EnterpriseSearch{},
+		beatv1beta1.Kind: &beatv1beta1.Beat{},
+	}); err != nil {
+		log.Error(err, "Orphan secrets garbage collection failed, will be attempted again at next operator restart.")
+		return
+	}
+	log.Info("Orphan secrets garbage collection complete")
 }
 
 func setupWebhook(mgr manager.Manager, certRotation certificates.RotationParams, validateStorageClass bool, clientset kubernetes.Interface) {
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -19,6 +19,7 @@ import (
 	apmv1 "github.com/elastic/cloud-on-k8s/pkg/apis/apm/v1"
 	apmv1beta1 "github.com/elastic/cloud-on-k8s/pkg/apis/apm/v1beta1"
 	beatv1beta1 "github.com/elastic/cloud-on-k8s/pkg/apis/beat/v1beta1"
+	esv1 "github.com/elastic/cloud-on-k8s/pkg/apis/elasticsearch/v1"
 	esv1beta1 "github.com/elastic/cloud-on-k8s/pkg/apis/elasticsearch/v1beta1"
 	entv1beta1 "github.com/elastic/cloud-on-k8s/pkg/apis/enterprisesearch/v1beta1"
 	kbv1 "github.com/elastic/cloud-on-k8s/pkg/apis/kibana/v1"
@@ -30,6 +31,7 @@ import (
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common/certificates"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common/container"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common/operator"
+	"github.com/elastic/cloud-on-k8s/pkg/controller/common/reconciler"
 	controllerscheme "github.com/elastic/cloud-on-k8s/pkg/controller/common/scheme"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common/tracing"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common/version"
@@ -47,6 +49,7 @@ import (
 	"github.com/elastic/cloud-on-k8s/pkg/dev/portforward"
 	licensing "github.com/elastic/cloud-on-k8s/pkg/license"
 	"github.com/elastic/cloud-on-k8s/pkg/telemetry"
+	"github.com/elastic/cloud-on-k8s/pkg/utils/k8s"
 	logconf "github.com/elastic/cloud-on-k8s/pkg/utils/log"
 	"github.com/elastic/cloud-on-k8s/pkg/utils/metrics"
 	"github.com/elastic/cloud-on-k8s/pkg/utils/net"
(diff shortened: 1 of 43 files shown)


barkbay (commented):
I just gave it a try, lgtm so far. I left a few nits and one question about the use of the UID instead of the object name.

pkg/controller/common/reconciler/secret.go
```diff
@@ -41,3 +51,120 @@ func ReconcileSecret(c k8s.Client, expected corev1.Secret, owner metav1.Object)
...
+//
+// Since they won't have an ownerRef specified, reconciled secrets will not be deleted automatically on parent deletion.
+// To account for that, we add a label for best-effort garbage collection by the operator on parent resource deletion.
+func ReconcileSecretNoOwnerRef(c k8s.Client, expected corev1.Secret, theoreticalOwner metav1.Object) (corev1.Secret, error) {
+	// this function is similar to "ReconcileSecret", but:
+	// - we don't pass an owner
+	// - we remove the existing owner
+	// - we set additional labels to perform garbage collection on owner deletion (best-effort)
+	expected.Labels[SoftOwnerNamespaceLabel] = theoreticalOwner.GetNamespace()
+	expected.Labels[SoftOwnerNameLabel] = theoreticalOwner.GetName()
```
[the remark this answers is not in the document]
sebgl: Hm I think the problem is that the garbage collection part of the code only knows about the deleted resource name + namespace (as part of the `reconcile.Request`). We don't know anything about the parent ID since it doesn't exist anymore? (Maybe I'm misunderstanding what you mean).

sebgl:
- [ ] Major bug to fix: the soft owner reference needs to include the object Kind, otherwise we may end up mismatching e.g. Elasticsearch stuff with Kibana stuff if they have the same name.

cmd/manager/main.go
```diff
@@ -611,8 +614,20 @@ func asyncTasks(
...
+	// Garbage collect orphan secrets leftover from deleted resources while the operator was not running
+	// - association user secrets
 	garbageCollectUsers(cfg, managedNamespaces)
+	// - soft-owned secrets
+	if err := reconciler.GarbageCollectAllOrphanSecrets(k8s.WrapClient(mgr.GetClient()), map[string]runtime.Object{
+		esv1.Kind:        &esv1.Elasticsearch{},
+		apmv1.Kind:       &apmv1.ApmServer{},
+		kbv1.Kind:        &kbv1.Kibana{},
+		entv1beta1.Kind:  &entv1beta1.EnterpriseSearch{},
+		beatv1beta1.Kind: &beatv1beta1.Beat{},
```
sebgl, line 626: I'm not super happy with that :( 
If anyone has another suggestion.

Reasoning behind this:
- when reconciling soft-owned secrets, we need to set a label representing the Kind ("Elasticsearch"), to not mismatch Elasticsearch with Kibana for example. The Kind can just be retrieved from the given owner `runtime.Object`, which is great.
- when running the garbage collection at startup, we need to `client.Get()` owners, which requires a typed object in parameter (e.g. `&esv1.Elasticsearch{}`)
- this map matches the soft owner Kind label with an actual object type for the garbage collection function
- retrieving the Kind from an apiserver object works fine, but `&esv1.Elasticsearch{}.Kind` is empty because the object did not go through the apiserver. I realized the operator bootstrap code actually performs reflection in `SchemeBuilder.Register()` to come up with that value...

pkg/apis/apm/v1/apmserver_types.go
```diff
@@ -13,7 +13,12 @@ import (
 	commonv1 "github.com/elastic/cloud-on-k8s/pkg/apis/common/v1"
 )
 
-const ApmServerContainerName = "apm-server"
+const (
+	ApmServerContainerName = "apm-server"
+	// Kind is inferred from the struct name using reflection in SchemeBuilder.Register()
+	// we duplicate it as a constant here for practical purposes.
+	Kind = "ApmServer"
```
sebgl, line 20: I did not find a better alternative to that. Using reflection similar to what SchemeBuilder.Register() is a bit more accurate but not sure if better? (we know those Kinds cannot change anyway? otherwise we'll run into all sorts of troubles)

sebgl:
Current status: this PR is missing unit tests, but otherwise I don't expect any code change.

cmd/manager/main.go
pebrc: What are you not happy with? The fact that you have to spec both `Kind` and the corresponding struct somewhat over-specifying even though one could be derived from the other? 

The only other idea how to potentially avoid this was not to use the controller-runtime client but using the [dynamic client from client-go directly](https://github.com/kubernetes/client-go/blob/v0.18.1/dynamic/interface.go#L39), in which case `Kind` should suffice and as we only care about exists/does not exist we would not need to worry about REST mapping etc. But it feels like a lot of trouble for little gain.

sebgl: Yeah 👍 
Using our usual k8s.Client makes it easier to test, so I think I'd rather keep things as they are.

pkg/controller/common/reconciler/secret.go
```diff
@@ -10,7 +10,20 @@ import (
...
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// Labels set on secrets which cannot rely on owner references due to https://github.com/kubernetes/kubernetes/issues/65200,
+// but should still be garbage-collected (best-effort) by the operator upon owner deletion.
+const (
+	SoftOwnerNamespaceLabel = "eck.k8s.elastic.co/soft-owner-namespace"
+	SoftOwnerNameLabel      = "eck.k8s.elastic.co/soft-owner-name"
+	SoftOwnerKindLabel      = "eck.k8s.elastic.co/soft-owner-kind"
```
pebrc, line 26: I think we could lose the `soft` bit and it would still be clear that it is an alternative to the built-in owner reference mechanism. Don't feel strongly though.

pkg/controller/common/reconciler/secret.go
```diff
@@ -41,3 +54,185 @@ func ReconcileSecret(c k8s.Client, expected corev1.Secret, owner metav1.Object)
...
 }
+
+// ReconcileSecretNoOwnerRef should be called to reconcile a Secret for which we explicitly don't want
+// an owner reference to be set, and want existing ownerReferences from previous operator versions to be removed,
+// because of this k8s bug: https://github.com/kubernetes/kubernetes/issues/65200 (fixed in k8s 1.20).
+//
+// It makes sense to use this function for secrets which are likely to be manually
+// copied into other namespaces by the end user.
+// Because of the k8s bug mentioned above, the ownerReference could trigger a racy garbage collection
+// that deletes all children resources, potentially resulting in data loss.
```
pebrc, line 65:
```suggestion
// that deletes all child resources, potentially resulting in data loss.
```

pebrc (approved):
LGTM, did a few simple tests that all seemed to work. Quirk: If you delete your es cluster while the operator is not running and recreate another cluster with the same name it adopts the left over resources from the previous incarnation which means password stays the same etc.

pkg/controller/common/certificates/reconcile.go
```diff
@@ -32,7 +33,8 @@ type Reconciler struct {
 	K8sClient      k8s.Client
 	DynamicWatches watches.DynamicWatches
 
-	Object        metav1.Object                     // owner for the TLS certificates (eg. Elasticsearch, Kibana)
+	Owner runtime.Object // owner for the TLS certificates (eg. Elasticsearch, Kibana)
```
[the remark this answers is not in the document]
sebgl: Indeed, sorry I should have made it clear in a PR comment.
The reason is that we need to access the `Kind`, which is not available in a `metav1.Object`.

I feel like a comment in the Github PR would have been useful, but a comment in the code would maybe be strange and not bring much value (`// we need a runtime.Object and not a metav1.Object here because we use the Kind deep down`).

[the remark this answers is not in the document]
pebrc: Absolutely, a comment would feel out of place.

pkg/controller/common/reconciler/secret.go
sebgl: Do you mean in the label name, or throughout the code?
I think I agree for the label part; however in the code having `Soft` makes things more explicit I think (this is not the namespace of the owner in k8s sense, it's the namespace of our soft owner in eck sense).
I think I'll remove `soft` from the label name but keep it in the code structures. WDYT?

sebgl:
run full pr build

sebgl:
We're missing a watch on soft-owned secrets to automatically trigger reconciliations on any change (e.g. deleting the elastic user secret).
I'll add another commit.

pebrc:
> We're missing a watch on soft-owned secrets to automatically trigger reconciliations on any change (e.g. deleting the elastic user secret).
> I'll add another commit.

Good catch! I totally forgot that we are basing our watches on the owner reference :-(

pkg/controller/common/watches/secrets_test.go
```diff
@@ -0,0 +1,91 @@
...
+
+	kbv1 "github.com/elastic/cloud-on-k8s/pkg/apis/kibana/v1"
+
+	"k8s.io/apimachinery/pkg/types"
+
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/elastic/cloud-on-k8s/pkg/controller/common/reconciler"
```
pebrc, line 14: import formatting

pkg/controller/common/watches/secrets_test.go
```diff
@@ -0,0 +1,91 @@
...
+			requests := toRequestsFunc(mapObject)
+			require.Equal(t, tt.wantReconcileRequests, requests)
+		})
+	}
+}
+
+func Test_reconcileReqForSoftOwner1(t *testing.T) {
+	type args struct {
+		kind string
+	}
```
pebrc, line 89: empty test can be deleted?

pebrc (approved):
LGTM, so happy we have these e2e tests...

pebrc:
Jenkins test this please 

My bad still not fixed https://github.com/elastic/cloud-on-k8s/issues/3617

sebgl:
run full pr build

jberger:
I guess I understand why this is done however it does mean that those generated resources do not show up in tools like Argo. When lots of resources are generated being visible from tooling is helpful so as to understand what has even been generated. Would there be any appetite to revisit this?

Issue #3881, opened 2020-10-28
Remote CA cert is created empty when remote clusters are not configured for an ES cluster


Navbar:
Upgrade ECK 1.0.1 operator to 1.2.1 over K8s ended with empty transport ca.crt.

I would like to check the ca rotation feature before expiration date - new parameters added:
       - --ca-cert-validity=80h
        - --ca-cert-rotate-before=2h
        - --cert-validity=80h
        - --cert-rotate-before=2h

How to reproduce:
1. Install ECK 1.0.1 (as quickstart) - certificate valid for 1y, possible to query _ssl/certificate over API
2. Upgrade Operator to 1.2.1 (apply all 1.2.1quickstart yaml files)

Logs show:
"message": "path: /_ssl/certificates, params: {}", "cluster.uuid": "TZIF4kdNTo2cadQReCI_yw", "node.id": "a63gze16SX21ZrqI0DVSbw" ,
"stacktrace": ["java.security.cert.CertificateException: failed to parse any certificates from [/usr/share/elasticsearch/config/transport-remote-certs/ca.crt]",

curl -u "elastic:password" -s -k "https://quickstart-es-http:9200/_ssl/certificates"|jq .
{
  "error": {
    "root_cause": [
      {
        "type": "certificate_exception",
        "reason": "failed to parse any certificates from [/usr/share/elasticsearch/config/transport-remote-certs/ca.crt]"
      }
    ],
    "type": "certificate_exception",
    "reason": "failed to parse any certificates from [/usr/share/elasticsearch/config/transport-remote-certs/ca.crt]"
  },
  "status": 500
}

empty cert file:
[root@quickstat-es-master-nodes-0 transport-remote-certs]# cat ca.crt
[root@quickstart-es-master-nodes-0 transport-remote-certs]# pwd
/usr/share/elasticsearch/config/transport-remote-certs

Am i doing something wrong?

sebgl:
Can you provide the ECK operator logs? I'm wondering if anything particular is logged about those certs.
Please also provide your Elasticsearch manifest.

Navbar:
kubectl logs elastic-operator-0 -nelastic-system|grep cert
{"log.level":"info","@timestamp":"2020-10-27T12:03:49.152Z","log.logger":"manager","message":"Automatic management of the webhook certificates enabled","service.version":"1.2.1-b5316231","service.type":"eck","ecs.version":"1.4.0"}
[...]

PR #3993, opened 2020-11-30
Use the public transport CA as remote CA if the remote CA list is empty

idanmo:
If remote clusters are not set for an ES cluster, the remote CA `ca.crt` mounted on the ES container is empty, which causes the `_ssl/certificates` ES REST API call to fail. In this PR, if remote clusters are not set for an ES cluster, the remote CA is populated with the public transport CA as a "dummy" CA. 
This fix preserves the behavior where adding/setting remote clusters for ES does not require restarting the ES cluster.

Fixes [#3881](https://github.com/elastic/cloud-on-k8s/issues/3881)


pkg/controller/elasticsearch/certificates/remoteca/reconcile_test.go
```diff
@@ -101,14 +104,24 @@ func TestReconcile(t *testing.T) {
 						Data: map[string][]byte{certificates.CAFileName: []byte("cert2\n")},
 					},
 				},
+				ca: nil,
```
pebrc, line 107: I think we should pass in the `testCA` here and on L64 as well to make sure it does not get included when there are remote CA certs.

pebrc (approved):
LGTM. I did a quick test and seems to work. The only thing I noticed is that the remote connection is initially disconnected because the CA certificate was not available yet when the remote cluster was set up in ES via the API, but I believe the same could happen before if the CA propagation was too slow.

pkg/controller/elasticsearch/certificates/remoteca/reconcile_test.go
idanmo: Yes, that makes sense.

pkg/controller/elasticsearch/certificates/remoteca/reconcile.go
```diff
@@ -36,6 +36,7 @@ func Labels(esName string) client.MatchingLabels {
 func Reconcile(
 	c k8s.Client,
 	es esv1.Elasticsearch,
+	ca *certificates.CA,
```
sebgl, line 39: Let's not make this a pointer so we don't have to check for `nil` in the function.
(In practice we know it's not `nil` in this case (until further refactored :)) but it's not very explicit).

pkg/controller/elasticsearch/certificates/remoteca/reconcile.go
```diff
@@ -36,6 +36,7 @@ func Labels(esName string) client.MatchingLabels {
 func Reconcile(
 	c k8s.Client,
 	es esv1.Elasticsearch,
+	ca certificates.CA,
```
sebgl, line 39: nit: `transportCA` would make things even more explicit

pkg/controller/elasticsearch/certificates/remoteca/reconcile_test.go
```diff
@@ -24,7 +24,9 @@ func TestReconcile(t *testing.T) {
 	type args struct {
 		es      esv1.Elasticsearch
 		secrets []runtime.Object
+		ca      *certificates.CA
```
sebgl, line 27: shouldn't be a pointer

sebgl (approved):
LGTM, I also did a few tests 👍

idanmo:
run full pr build

PR #3996, opened 2020-12-01
Bring license signer in line with external tooling

pebrc:
Remove a subtle difference in how a public key is incorporated into the signature compared to other Elastic(search) tooling.

### DIFF
--- a/pkg/controller/common/license/crypto.go
+++ b/pkg/controller/common/license/crypto.go
@@ -15,12 +15,12 @@ const customPadding = 20
 
 var (
 	// aesKey is effectively constant
-	// see https://github.com/elastic/elasticsearch/blob/master/x-pack/plugin/core/src/main/java/org/elasticsearch/license/CryptUtils.java#L162-L179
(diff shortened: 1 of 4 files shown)


pkg/controller/common/license/crypto.go
```diff
@@ -0,0 +1,64 @@
...
+	"crypto/aes"
+	"crypto/cipher"
+	"crypto/rand"
+)
+
+const customPadding = 20
+
+var (
+	// aesKey is effectively constant
+	// see https://github.com/elastic/elasticsearch/blob/master/x-pack/plugin/core/src/main/java/org/elasticsearch/license/CryptUtils.java#L162-L179
```
barkbay, line 18: Should we use [permalinks](https://github.com/elastic/elasticsearch/blob/69cebcd516c208c8e563c0d4e6e9f99e71bf4a3d/x-pack/plugin/core/src/main/java/org/elasticsearch/license/CryptUtils.java#L162-L179) when referring to files that may change ?

barkbay (approved):
LGTM, even if it's a bit to hard to understand the change in the signer tbh  😄 

Since `Sign` is used to sign the trial licence I did a quick test to check the backward compatibility: a previously signed trial licence is still considered as valid with this branch (not sure it is a relevant test but it helped to remind me the workflow of this part of the code 😄 )

Issue #3504, opened 2020-07-22
ECK on ARM architectures


jeffspahr:
## Support running ECK on ARM architectures

Elastic is starting to provide multi-architecture docker images to support ARM architectures:
https://www.elastic.co/blog/elasticsearch-on-arm

It would be useful to be able to run Elastic Stack on ARM hardware via ECK.

Use cases:
Home Lab - Raspberry Pi Kubernetes clusters
Enterprise - ARM servers are gaining popularity in public clouds.  https://docs.aws.amazon.com/eks/latest/userguide/arm-support.html

jeffspahr:
@charith-elastic I wanted to pull your comments on CI testing for ARM into this open issue.

https://github.com/elastic/cloud-on-k8s/issues/922#issuecomment-691962564 

> I had a brief look at this out of interest. From the point of view of ECK, the main issue is how to run the automated test suites on ARM architectures. ~AFAIK, there are no major cloud providers that fully support ARM, so it is not currently possible to set up the test infrastructure to automatically test ECK. (Testing on RPi's is not an automated option, unfortunately.)~

I'm guessing you came across the link above for ARM nodes on ECK in AWS, but I figured it'd be helpful to be a bit more explicit.

AWS seems to be investing pretty heavily in arm64 instance types:
- https://aws.amazon.com/ec2/graviton/
- https://docs.aws.amazon.com/eks/latest/userguide/arm-support.html

[These GA'd on June 11, 2020](https://aws.amazon.com/about-aws/whats-new/2020/07/announcing-preview-for-amazon-rds-m6g-and-r6g-instance-types/#:~:text=AWS%20announced%20the%20general%20availability,processors%20on%20June%2011%2C%202020.) and they are already seeing production use:
- https://twitter.com/lizthegrey/status/1296837123195260928
- https://twitter.com/_msw_/status/1280395618519674882

I was hoping that some of the road is paved for us already based on what [Elasticsearch is doing around arm64 support](https://www.elastic.co/blog/elasticsearch-on-arm).

charith-elastic:
Yes, AWS ARM machines are now GA and EKS seems to support them so that's one hurdle down.

[...]

Issue #3906, opened 2020-11-04
ci-build-image invalid credentials


anyasabo:
Looks like https://github.com/elastic/cloud-on-k8s/pull/3893 is failing because `ci-build-image` is using invalid docker credentials:

https://devops-ci.elastic.co/job/cloud-on-k8s-pr/4872/console

```
11:38:45  Successfully built bbd1c698f6aa
11:38:45  Successfully tagged docker.elastic.co/eck-ci/eck-ci-tools:290d60ed
11:38:45  Push docker.elastic.co/eck-ci/eck-ci-tools:290d60ed...
11:38:45  denied: requested access to the resource is denied
11:38:45  Retry 1/5 exited 1, retrying in 1 seconds...
11:38:45  denied: requested access to the resource is denied
11:38:45  Retry 2/5 exited 1, retrying in 2 seconds...
11:38:45  denied: requested access to the resource is denied
11:38:45  Retry 3/5 exited 1, retrying in 4 seconds...
11:38:48  denied: requested access to the resource is denied
11:38:48  Retry 4/5 exited 1, retrying in 8 seconds...
11:38:56  denied: requested access to the resource is denied
11:38:56  Retry 5/5 exited 1, no more retries left.
11:38:56  The push refers to repository [docker.elastic.co/eck-ci/eck-ci-tools]
11:38:56  The push refers to repository [docker.elastic.co/eck-ci/eck-ci-tools]
11:38:56  The push refers to repository [docker.elastic.co/eck-ci/eck-ci-tools]
11:38:56  The push refers to repository [docker.elastic.co/eck-ci/eck-ci-tools]
11:38:56  The push refers to repository [docker.elastic.co/eck-ci/eck-ci-tools]
11:38:56  Makefile:63: recipe for target 'ci-build-image' failed
11:38:56  make[1]: *** [ci-build-image] Error 1
11:38:56  make[1]: Leaving directory '/var/lib/jenkins/workspace/cloud-on-k8s-pr/.ci'
11:38:56  Makefile:42: recipe for target 'ci' failed
```

This doesn't crop up in other PRs because we only push when something changes in the CI image. I didn't get a chance to look into it so wanted to create the issue just to make sure it's not forgotten.

david-kow:
I took a look at it and it seems to me that we are not trying to log in with admin creds, ie. we just exit here:
[...]

PR #4000, opened 2020-12-02
Add EKS ARM e2e job

charith-elastic:
Add an E2E job to run on ARM machines. This PR only introduces the job. After it is tested manually, it will be added to the nightly job in a separate PR.

Related to #3504

### DIFF
--- a/.ci/setenvconfig
+++ b/.ci/setenvconfig
@@ -335,7 +335,6 @@ write_deployer_config <<CFG
 id: eks-arm-ci
 overrides:
   clusterName: ${BUILD_TAG}-arm
-  machineType: m6gd.2xlarge
   vaultInfo:
     address: $VAULT_ADDR
     roleId: $VAULT_ROLE_ID
--- a/.ci/setenvconfig
+++ b/.ci/setenvconfig
@@ -310,6 +310,40 @@ CFG
 ;;
 
 ######################################
+  e2e/eks-arm)
+######################################
+
+write_env <<ENV
+IMG_SUFFIX = -ci
+E2E_REGISTRY_NAMESPACE = eck-ci
+
+GO_TAGS = release
+export LICENSE_PUBKEY = /go/src/github.com/elastic/cloud-on-k8s/.ci/license.key
+TEST_LICENSE = /go/src/github.com/elastic/cloud-on-k8s/.ci/test-license.json
+MONITORING_SECRETS = /go/src/github.com/elastic/cloud-on-k8s/.ci/monitoring-secrets.json
+
+OPERATOR_IMAGE = $JKS_PARAM_OPERATOR_IMAGE
+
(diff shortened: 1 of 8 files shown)


charith-elastic:
run full pr build

charith-elastic:
Jenkins test this please

charith-elastic:
run full pr build

hack/docker.sh
```diff
@@ -26,11 +26,6 @@ docker-login() {
     local image=$1
     local registry=${image%%"/"*}
 
-    if grep -q "$registry" ~/.docker/config.json; then
```
pebrc, line 29: Does this cause problems with incorrect push permissions or why are we removing it? Curious because is came up in the context of https://github.com/elastic/cloud-on-k8s/issues/3906

charith-elastic: I couldn't pinpoint the exact cause, but, removing this check is the way to make the CI image build succeed. My theory is that we are getting a very short-lived Docker credential from Vault. Because it takes a while to build the CI image, by the time it is done the credential has already expired.

I couldn't see a reason for having this check in here. It's not a particularly expensive operation to do. Given that the build randomly fails because of it, I thought it would be better to remove it.

pebrc: Agreed!

pebrc:
I did a quick test just with deployer and I think we need to update the `eksctl` version to at least v30 to get support for k8s 1.18

charith-elastic:
Jenkins test this please

pebrc:
I tested the deployer config part of the PR which seems to work fine. I noticed https://github.com/elastic/cloud-on-k8s/issues/4038 though which we need to look into.

charith-elastic:
#4081 should fix #4038. I managed to run the container on a `linux/aarch64` (Graviton 2) machine after that change. I also tried deploying the operator on ARM using Kind. 
- Kind base images don't have an ARM build, so I had to use a custom image: `kind create cluster --image rossgeorgiev/kind-node-arm64:v1.20.0`
- There wasn't enough disk space in the machine to run the full E2E suite on Kind. I did manage to deploy Elasticsearch 7.10.0 using the operator though.

.ci/setenvconfig
```diff
@@ -309,6 +309,40 @@ overrides:
...
+E2E_PROVIDER=eks
+CLUSTER_NAME=${BUILD_TAG}-arm
+KUBERNETES_VERSION=default
+TEST_OPTS=-race
+ENV
+write_deployer_config <<CFG
+id: eks-arm-ci
+overrides:
+  clusterName: ${BUILD_TAG}-arm
+  machineType: m6gd.2xlarge
```
pebrc, line 338: Do we need this override? The same machine type is specified in plans.yaml

pebrc (approved):
LGTM

Issue #3997, opened 2020-12-01
[META] Elastic Agent CRD (standalone)


david-kow:
### Elastic Agent
ECK will support a new Elastic Agent CRD that enables to deploy and manage Elastic Agent. The current goal is to only provide support for the Agent running in the [standalone](https://www.elastic.co/guide/en/fleet/current/run-elastic-agent-standalone.html) mode.

### CRD
The approach I propose to take is based on our experience with Beats. There are only few changes between those CRDs, as below. I would not include any indication of `mode` (standalone vs fleet) yet, because we will only support one for now.

#### Remove
- `type` - there are no types of Elastic Agent to deploy, as it's driven by the config
- `kibanaRef` - as of now, Elastic Agent doesn't have any direct ties to Kibana, hence no ref
- `elasticsearchRef` - see "Add" below

#### Add
- `elasticsearchRefs` - multiple outputs can be defined for a single Elastic Agent, different inputs can use different outputs to send the data. Given that, we need:
  1. a way to reference multiple ES clusters (hence list, not a single item)
  2. an ability to name them, as this is how users will refer to them in other parts of the config (alternatively, we could derive the name from namespace and name, but this would force users to use those names instead of defining their own - I'd be against that).

#### Spec example

```yaml
apiVersion: agent.k8s.elastic.co/v1alpha1
kind: Agent
metadata:
  name: elastic-agent
spec:
  version: 7.10.0
  image: docker.elastic.co/beats/elastic-agent:7.10.0

  configRef: ... # optional, instead of `config`
  secureSettings: ... # optional
  serviceAccountName: ... # optional
  daemonSet: ... # optional, one of `daemonSet, deployment`
  deployment: ... # optional, one of `daemonSet, deployment`

  elasticsearchRefs:
  - name: elasticsearch
    outputName: default
  - name: k8s-monitoring
    namespace: monitoring
    outputName: monitoringOut

  config:
    id: 2d70a6f0-33a5-11eb-bb2f-418d0388a8cf
    revision: 2
    agent:
      monitoring:
        enabled: true
        use_output: monitoringOut
        logs: true
        metrics: true
    inputs:
      - id: 2e187fb0-33a5-11eb-bb2f-418d0388a8cf
        name: system-1
        revision: 1
        type: logfile
        use_output: default
        meta:
          package:
            name: system
            version: 0.9.1
        data_stream:
          namespace: default
        streams:
...
```

### Not in scope
[...]

PR #4010, opened 2020-12-04
Add Elastic Agent CRD (standalone mode)

david-kow:
#### In this PR

Added:

- Elastic Agent CRD 
- Elastic Agent controller and its association controller
- required RBAC changes for Helm/E2E tests
- required PSP for E2E tests
- E2E tests utilities for Elastic Agent
- E2E test of Elastic Agent with System integration enabled
- improved logging utilities

#### Additional notes

- CRD is complete (contains all fields forseen right now), but controller does not support all of the features yet. These gaps will be [tracked](https://github.com/elastic/cloud-on-k8s/issues/3997) separately and are out-of-scope for this PR.
- code is heavily based on the Beats code, but it's not shared. My assessment was that the work required to make the code reusable was not worth it and we could end up with it being fairly convoluted. Also, Elastic Agent most likely will evolve differently than Beats so having the code separated seems to be desirable.

#### Try it out yourself!

- an example manifest that works OOTB on non-PSP cluster:

<details>
  <summary>Click to see Elasticsearch, Kibana, Elastic Agent manifest</summary>

```yaml
cat <<EOF | kubectl apply -f -
kind: Elasticsearch
metadata:
  name: elasticsearch
spec:
  version: 7.10.0
  nodeSets:
  - name: default
    count: 3
    config:
      node.store.allow_mmap: false
---
kind: Kibana
metadata:
  name: kibana
spec:
  version: 7.10.0
  count: 1
  elasticsearchRef:
    name: elasticsearch
---
kind: Agent
metadata:
  name: elastic-agent
spec:
  version: 7.10.0
  elasticsearchRefs:
  - name: elasticsearch
  daemonSet:
    podTemplate:
      spec:
        securityContext:
          runAsUser: 0
  config:
    id: 2d70a6f0-33a5-11eb-bb2f-418d0388a8cf
    revision: 2
    agent:
      monitoring:
        enabled: true
        use_output: default
        logs: true
        metrics: true
    inputs:
      - id: 2e187fb0-33a5-11eb-bb2f-418d0388a8cf
        name: system-1
        revision: 1
        type: logfile
        use_output: default
        meta:
          package:
            name: system
            version: 0.9.1
        data_stream:
          namespace: default
        streams:
          - id: logfile-system.auth
            data_stream:
              dataset: system.auth
              type: logs
            paths:
              - /var/log/auth.log*
              - /var/log/secure*
            exclude_files:
              - .gz$
            multiline:
              pattern: ^\s
              match: after
            processors:
              - add_locale: null
              - add_fields:
                  target: ''
                  fields:
                    ecs.version: 1.5.0
          - id: logfile-system.syslog
            data_stream:
              dataset: system.syslog
              type: logs
            paths:
              - /var/log/messages*
              - /var/log/syslog*
            exclude_files:
              - .gz$
            multiline:
              pattern: ^\s
              match: after
            processors:
              - add_locale: null
              - add_fields:
                  target: ''
                  fields:
                    ecs.version: 1.5.0
      - id: 2e187fb0-33a5-11eb-bb2f-418d0388a8cf
        name: system-1
        revision: 1
        type: system/metrics
        use_output: default
        meta:
          package:
            name: system
            version: 0.9.1
        data_stream:
          namespace: default
        streams:
          - id: system/metrics-system.cpu
            data_stream:
              dataset: system.cpu
              type: metrics
            metricsets:
              - cpu
            cpu.metrics:
              - percentages
              - normalized_percentages
            period: 10s
          - id: system/metrics-system.diskio
            data_stream:
              dataset: system.diskio
              type: metrics
            metricsets:
              - diskio
            diskio.include_devices: null
            period: 10s
          - id: system/metrics-system.filesystem
            data_stream:
              dataset: system.filesystem
              type: metrics
            metricsets:
              - filesystem
            period: 1m
            processors:
              - drop_event.when.regexp:
                  system.filesystem.mount_point: ^/(sys|cgroup|proc|dev|etc|host|lib|snap)($|/)
          - id: system/metrics-system.fsstat
            data_stream:
              dataset: system.fsstat
              type: metrics
            metricsets:
              - fsstat
            period: 1m
            processors:
              - drop_event.when.regexp:
                  system.fsstat.mount_point: ^/(sys|cgroup|proc|dev|etc|host|lib|snap)($|/)
          - id: system/metrics-system.load
            data_stream:
              dataset: system.load
              type: metrics
            metricsets:
              - load
            period: 10s
          - id: system/metrics-system.memory
            data_stream:
              dataset: system.memory
              type: metrics
            metricsets:
              - memory
            period: 10s
          - id: system/metrics-system.network
            data_stream:
              dataset: system.network
              type: metrics
            metricsets:
              - network
            period: 10s
            network.interfaces: null
          - id: system/metrics-system.process
            data_stream:
              dataset: system.process
              type: metrics
            metricsets:
              - process
            period: 10s
            process.include_top_n.by_cpu: 5
            process.include_top_n.by_memory: 5
            process.cmdline.cache.enabled: true
            process.cgroups.enabled: false
            process.include_cpu_ticks: false
            processes:
              - .*
          - id: system/metrics-system.process_summary
            data_stream:
              dataset: system.process_summary
              type: metrics
            metricsets:
              - process_summary
            period: 10s
          - id: system/metrics-system.socket_summary
            data_stream:
              dataset: system.socket_summary
              type: metrics
            metricsets:
              - socket_summary
            period: 10s
          - id: system/metrics-system.uptime
            data_stream:
              dataset: system.uptime
              type: metrics
            metricsets:
              - uptime
            period: 10s
EOF
```

</details>

- kubectl tree output for the above manifest:

```shell
$ kubectl tree agents elastic-agent
NAMESPACE  NAME                                                   READY  REASON  AGE
default    Agent/elastic-agent                                    -              59s
default    ├─DaemonSet/elastic-agent-agent                        -              55s
default    │ ├─ControllerRevision/elastic-agent-agent-7b97cddc98  -              55s
default    │ ├─Pod/elastic-agent-agent-dcvsb                      True           55s
default    │ ├─Pod/elastic-agent-agent-lmwkf                      True           55s
default    │ └─Pod/elastic-agent-agent-pwxhf                      True           55s
default    ├─Secret/elastic-agent-agent-config                    -              55s
default    ├─Secret/elastic-agent-agent-es-ca                     -              57s
default    └─Secret/elastic-agent-agent-user                      -              59s
```

- data streams visible in Fleet UI:

Screenshot 2020-12-04 at 10 07 34

- to create a manifest on your own:
  - create the desired policy in Fleet UI
  - in `Policies` tab go to `Actions` (three dots) > `Add agent` > `Run standalone`
  - remove `outputs` and use the config as the value of `config` in Agent spec

Screenshot 2020-12-04 at 11 24 57

#### Testing

- E2E test utilities for Elastic Agent were added
- basic, example E2E test is added (passing on PSP-enabled cluster when manually tested)
- UT coverage will be added gradually

#### Logging and APM

As per previous discussions, our next CRD controller should pass tracing context around to allow log correlation. It turned out there is not that much of logging in the Agent controller, but the facilities were added (based on @charith-elastic work) and can be reused in the future. The logic I followed is that every function that returns `reconciler.Results` will have a matching APM `Span` and will capture errors, before bubbling them up. This way errors visible in APM provide more context. The changes in Agent controller compared to other controllers manifest as:
- transaction metadata is passed in context
- logger with transaction metadata included is passed in context:

Screenshot 2020-12-04 at 08 03 14

- iteration is now a label on the transaction allowing to see a particular iteration execution in the APM UI with stacks, logs and errors attached:

Screenshot 2020-12-03 at 16 00 28

- more spans in the code:

Screenshot 2020-12-03 at 21 10 26


pebrc (commented):
I did a very quick first pass. Looks good and works at least the default setup you provided. I want to find more time to look more closely. The only thing the stuck out a bit is the setup for the tracing where I think we need to streamline the API a bit and separate concerns between the generic log package and reconciliation a bit more maybe.

pkg/controller/common/tracing/transaction.go
```diff
@@ -28,3 +28,26 @@ func EndTransaction(tx *apm.Transaction) {
 		tx.End()
 	}
 }
+
+// NewContextTransaction starts a new transaction and sets up a new context with that transaction that also contains the related
+// APM agent's tracer.
+func NewContextTransaction(t *apm.Tracer, txType, txName string, labels map[string]string) context.Context {
```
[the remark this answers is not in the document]
david-kow: I wanted to keep them both until we have the improvements (setting label) that were added to Agent controller elsewhere. What I want to avoid is to bundle the rework of other controllers into this PR. Similar to `LogReconciliationRunNoSideEffects` function.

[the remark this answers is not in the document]
pebrc: OK makes sense!

pkg/controller/agent/config.go
```diff
@@ -0,0 +1,127 @@
...
+			"type":     "elasticsearch",
+			"username": username,
+			"password": password,
+			"hosts":    []string{assoc.AssociationConf().GetURL()},
+		}
+		if assoc.AssociationConf().GetCACertProvided() {
+			output["ssl.certificate_authorities"] = []string{path.Join(certificatesDir(assoc), CAFileName)}
+		}
+
+		agentAssoc := (assoc).(*v1alpha1.AgentESAssociation)
```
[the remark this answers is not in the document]
david-kow: I wanted to depend on loop variable, but this won't make any effective difference. Changed as you suggested. This part will likely change when https://github.com/elastic/cloud-on-k8s/issues/4018 is done.

test/e2e/test/agent/builder.go
```diff
@@ -0,0 +1,309 @@
...
+
+func (b Builder) SkipTest() bool {
+	ver := version.MustParse(b.Agent.Spec.Version)
+	return version.SupportedAgentVersions.WithinRange(ver) != nil
+}
+
+// NewBuilderFromAgent creates an Agent builder from an existing Agent config. Sets all additional Builder fields
+// appropriately.
+func NewBuilderFromAgent(agent *agentv1alpha1.Agent) Builder {
+	var podTemplate *corev1.PodTemplateSpec
```
pebrc, line 54: Seems unused?

test/e2e/test/agent/builder.go
```diff
@@ -0,0 +1,309 @@
...
+		Agent:       *agent,
+		PodTemplate: podTemplate,
+	}
+}
+
+func NewBuilder(name string) Builder {
+	return newBuilder(name, rand.String(4))
+}
+
+func newBuilder(name string, suffix string) Builder {
```
pebrc, line 71: why do we need this function? Should it not be just `NewBuilder`?

test/e2e/test/agent/builder.go
```diff
@@ -0,0 +1,309 @@
...
+
+func (b Builder) WithPodTemplateServiceAccount(name string) Builder {
+	b.PodTemplate.Spec.ServiceAccountName = name
+
+	return b
+}
+
+func (b Builder) WithRoles(clusterRoleNames ...string) Builder {
+	for _, clusterRoleName := range clusterRoleNames {
+		b = bind(b, clusterRoleName)
```
pebrc, line 213: Not sure what happens if you assign to the method receiver, my IDE does not seem to like it and I am actually unsure if the `b` in the the `bind` invocation will be propagated? Maybe introduce a variable just to avoid confusion?

pkg/apis/agent/v1alpha1/agent_types.go
```diff
@@ -0,0 +1,223 @@
...
+	commonv1 "github.com/elastic/cloud-on-k8s/pkg/apis/common/v1"
+)
+
+// AgentSpec defines the desired state of the Agent
+type AgentSpec struct {
+	// Version of the Agent.
+	Version string `json:"version"`
+
+	// ElasticsearchRefs is a reference to a list of Elasticsearch clusters running in the same Kubernetes cluster.
+	// Due to existing limitations, only a single ES cluster is currently supported.
```
charith-elastic, line 21: Are you planning on adding a validation check to enforce this?

charith-elastic (commented):
LGTM. I am a bit torn about `elasticsearchRefs`. If we support only one sink at the moment, maybe we should stick with `elasticsearchRef` because the CRD is still in alpha and can always be changed later. On the other hand, I can see that it'd be a painful migration to do.

pebrc (approved):
LGTM I think I am in favour of merging. Nice work! I have not tested the tracing bits. Implementationwise it looks indeed very similar to our Beats implementation and I wonder if we can merge some of the code once we have the full picture of what Agent will look like.

pkg/apis/agent/v1alpha1/agent_types.go
david-kow: Yes, as part of https://github.com/elastic/cloud-on-k8s/issues/4012. As to the `elasticsearchRefs`, I'd like to take a long shot at supporting it on the initial release, already chatted with @barkbay about this a bit.

test/e2e/test/agent/builder.go
david-kow: I think it warns because it thinks, that I think `b` will be replaced at the call site, while I'm just using it as a variable to carry the result. Anyway, probably better to rename 👍

david-kow:
run full pr build

charith-elastic (approved):

PR #4026, opened 2020-12-09
Update Kubernetes API version used in docs

charith-elastic:
Preemptively update the Kubernets API version we link to from the API docs. Links to older Kubernetes versions tend to disappear silently from the site.


pebrc (approved):

sebgl (approved):
LGTM

PR #4035, opened 2020-12-14
Add webhook for Agent controller

pebrc:
Add a webhook for the agent controller with a similar validations to what we have in Beats.
* add webhook definition to Helm charts
* call create validations as a fallback from the controller
* add a very simple e2e test to verify functionality


pkg/apis/agent/v1alpha1/validations.go
```diff
@@ -0,0 +1,75 @@
...
+)
+
+var (
+	defaultChecks = []func(*Agent) field.ErrorList{
+		checkNoUnknownFields,
+		checkNameLength,
+		checkSupportedVersion,
+		checkAtMostOneDeploymentOption,
+		checkSingleConfigSource,
+		checkSpec,
```
david-kow, line 20: Can we add one more to validate whether output is specified correctly? Agent expects exactly one output to be named `default`. To facilitate this for ECK users, they don't need to provide `outputName` if there is only one item in `elasticsearchRefs`. But they have to specify it for all outputs if there is more than one item. I think it would be good to validate user input here.

david-kow (commented):
I'd add one more validation, otherwise looks good.

pebrc:
Jenkins test this please

pkg/apis/agent/v1alpha1/validations.go
```diff
@@ -49,6 +53,41 @@ func checkAtMostOneDeploymentOption(b *Agent) field.ErrorList {
...
+func checkAtMostOneDefaultESRef(b *Agent) field.ErrorList {
+	var found int
+	for _, o := range b.Spec.ElasticsearchRefs {
+		if o.OutputName == "default" {
+			found++
+		}
+	}
+	if found > 1 {
+		return field.ErrorList{
+			field.Forbidden(field.NewPath("spec").Child("elasticsearchRefs"), "only one elasticsearchRef may be named 'default'"),
```
david-kow, line 65: nit: the field name that we are complaining about is `outputName`, maybe message could hint that?

david-kow (approved):
One nit, other than that, LGTM!

Issue #3718, opened 2020-09-07
Coordinator nodes cannot be defined using node.roles


charith-elastic:
This was mentioned in #3324 but is worth raising as a separate issue.

`ucfg` seems to treat empty arrays as null values so the [suggested method of defining a coordinator node](https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-node.html#coordinating-only-node) cannot be done on ECK by setting `node.roles: [ ]`.

```
ElasticsearchParseException[null-valued setting found for key [node.roles] found at line number [16], column number [10]]
```

charith-elastic:
Workaround is to use YAML type tags:

```yaml
node.roles: !!seq ""
```

This is probably not going to be obvious to the users and slightly deviates from the Elasticsearch documentation (even though they are semantically equivalent.)

Other options that I could think of are:
- Make `config` a strongly-typed struct to help ucfg figure out how to deserialize each field.
   - Con: we'll have to keep on top of any changes to the Elasticsearch configuration schema.
- Ask the Elasticsearch team to consider adding `coordinator` as a valid role name. 
   - Con: as the empty array is already an accepted value in a released version, they will have to keep supporting it as well but ECK won't be able to.
- Ditch ucfg and use standard YAML parsing libraries
   - Con: Handling dot separated setting keys is going to be tricky.
- Ask ucfg to support external schemas/type hints for parsing certain paths.

charith-elastic:
Added a note to `docs/orchestrating-elastic-stack-applications/elasticsearch/node-configuration.asciidoc` in #3719 about the YAML type tag workaround. If a different solution is found, that note should be removed.

pebrc:
I don't think the workaround works as expected. On 7.10.0 with `node.roles: !!seq ""` all node roles seem to be configured. The rendered configuration on the node has `node.roles: ""` 

```
10.73.50.49 41 65  4 1.21 0.41 0.27 cdhilmrstw - elasticsearch-es-coordinating-0
```

pebrc:
The only workaround <8.0 that I could find is to use the legacy node role format as in: 
```yaml
  - config:
      node.master: false
      node.data: false
      node.ingest: false
      node.ml: false
      node.remote_cluster_client: false
```

pebrc:
[...]

PR #4041, opened 2020-12-14
Add a replacement mechanism to work around nil handling in ucfg

pebrc:
Stopping short of replacing our use of ucfg alltogether this tries to provide a short term fix to the rendering problem that currently prevents users specifying Elasticsearch coordinator nodes using the new `node.roles: []` syntax. 

This introduces a simple (hacky) replacement mechanism into the configuration rendering process to handle the cases where we know that null is not equivalent to [] (empty seq) in the rendered Yaml configuration.

If we think this is viable as short term fix for https://github.com/elastic/cloud-on-k8s/issues/3718 I would add a few more tests and see to get this merged for 1.4.0


sebgl (commented):
I think I'm good with the approach 👍 
Let's make sure we add a comment to explain why we do this.

(I fiddle a bit with patching the raw user input instead of the config rendering but it does not really make things easier...)

pkg/controller/common/settings/replace.go
```diff
@@ -0,0 +1,35 @@
...
+				replaceTarget[p] = r.Replacement
+				return
+			}
+		}
+
+		v, exists := replaceTarget[p].(untypedDict)
+		if !exists {
+			return
+		}
+		replaceTarget = v
```
sebgl, line 33: The 4 lines above give me a headache for some reason 😄 
we mutate `replaceTarget` by the inner path value here, right? A comment would help, but also variable names a bit more explicit than `v` and `p`.

pebrc: You are right, it is too terse. I revised the naming.

pkg/apis/elasticsearch/v1/elasticsearch_config.go
```diff
@@ -37,13 +39,14 @@ type ClusterSettings struct {
...
-	ML         *bool    `config:"ml"`
-	Transform  *bool    `config:"transform"` // available as of 7.7.0
-	Roles      []string `config:"roles"`     // available as of 7.9.0, takes priority over the other fields if non-nil
-	VotingOnly *bool    `config:"voting_only"`
+	Master              *bool    `config:"master"`
+	Data                *bool    `config:"data"`
+	Ingest              *bool    `config:"ingest"`
+	ML                  *bool    `config:"ml"`
+	Transform           *bool    `config:"transform"`             // available as of 7.7.0
+	RemoteClusterClient *bool    `config:"remote_cluster_client"` // available as of 7.7.0
```
pebrc, line 47: We have an inconsistency with our sset pod labelling. We label the `transform` role but we don't label the `voting_only` or the `remote_cluster_role` (which we ignored until this PR). I am not sure what to do about it. Are there node roles that are more important than others? If we consider them all equals we should add labels for all of them on the pods.

pkg/controller/common/settings/replace_test.go
```diff
@@ -0,0 +1,71 @@
...
+					"something": "else",
+				},
+			},
+			expected: []byte(`node:
+  something: else
+`),
+		},
+	}
+
+	for _, tt := range tests {
```
david-kow, line 64: Maybe invoke inside `t.Run`?

pkg/controller/common/settings/replace_test.go
```diff
@@ -0,0 +1,71 @@
...
+	tests := []struct {
+		name     string
+		input    map[string]interface{}
+		expected []byte
+	}{
+		{
+			name: "nil roles replacement",
+			input: map[string]interface{}{
+				"node": map[string]interface{}{
+					"roles": make([]string, 0),
```
david-kow, line 32: Should this be `roles: nil`? Or maybe add a separate test to validate if the "[] to nil" behavior is still there in ucfg?

test/e2e/test/elasticsearch/checks_es.go
```diff
@@ -186,24 +186,33 @@ func (e *esClusterChecks) CheckESNodesTopology(es esv1.Elasticsearch) test.Step
 }
 
 func compareRoles(expected *esv1.Node, actualRoles []string) bool {
```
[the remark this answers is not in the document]
pebrc: Well spotted!

pkg/apis/elasticsearch/v1/elasticsearch_config.go
sebgl: I think I'm +1 for labelling them all.

pebrc: OK will add another commit here.

pebrc: I added a commit only to realise that there is more amiss #4054

pkg/controller/common/settings/replace_test.go
```diff
@@ -0,0 +1,85 @@
...
+			expected: []byte(`node:
+  roles:
+  - master
+`),
+		},
+		{
+			name: "does not touch other paths",
+			input: map[string]interface{}{
+				"node": map[string]interface{}{
+					"something": "else",
```
sebgl, line 68: would be more interesting if it was `"something": nil`

pkg/controller/elasticsearch/label/label.go
```diff
@@ -42,6 +42,10 @@ const (
 	NodeTypesMLLabelName common.TrueFalseLabel = "elasticsearch.k8s.elastic.co/node-ml"
 	// NodeTypesTransformLabelName is a label set to true on nodes with the transform role
 	NodeTypesTransformLabelName common.TrueFalseLabel = "elasticsearch.k8s.elastic.co/node-transform"
+	//NodeTypesRemoteClusterClientLabelName is a label set to true on nodes with the remote_cluster_client role
+	NodeTypesRemoteClusterClientLabelName common.TrueFalseLabel = "elasticsearch.k8s.elastic.co/node-remote_cluster_client"
+	//NodeTypesVotingOnlyLabelName is a label set to true on nodes with voting_only master-eligible node
```
sebgl, line 47: nit: `//<whitespace>`

pkg/controller/elasticsearch/label/label.go
```diff
@@ -116,9 +120,14 @@ func NewPodLabels(
 	NodeTypesDataLabelName.Set(nodeRoles.HasDataRole(), labels)
 	NodeTypesIngestLabelName.Set(nodeRoles.HasIngestRole(), labels)
 	NodeTypesMLLabelName.Set(nodeRoles.HasMLRole(), labels)
-	// transform nodes were only added in 7.7.0 so we should not annotate previous versions with them
+	// transform and remote_cluster_client roles were only added in 7.7.0 so we should not annotate previous versions with them
 	if ver.IsSameOrAfter(version.From(7, 7, 0)) {
 		NodeTypesTransformLabelName.Set(nodeRoles.HasTransformRole(), labels)
+		NodeTypesRemoteClusterClientLabelName.Set(nodeRoles.HasRemoteClusterClientRole(), labels)
+	}
+	// voting_only master eligible nodes were added only in 7.3.0 so we don't want label prior versions with it
```
sebgl, line 128: `we don't want to`?

pkg/controller/elasticsearch/settings/canonical_config.go
```diff
@@ -10,6 +10,18 @@ import (
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common/version"
 )
 
+// replaceNilNodeRoles works around a rendering issue with 0-sized string arrays/slices
+// that have gone through ucfg and have been reduced to nil.
+// Elasticsearch's new node.roles syntax relies on empty string arrays in YAML to express
+// a coordinating only node.
```
sebgl, line 16: let's link to https://github.com/elastic/cloud-on-k8s/issues/3718 here maybe?

sebgl (approved):
LGTM

test/e2e/es/topology_test.go
```diff
@@ -0,0 +1,21 @@
...
+	"github.com/elastic/cloud-on-k8s/test/e2e/test/elasticsearch"
+)
+
+//TestCoordinatorNodes tests a cluster with coordinator nodes.
+func TestCoordinatorNodes(t *testing.T) {
+	b := elasticsearch.NewBuilder("test-es-coord").
+		WithESMasterDataNodes(1, elasticsearch.DefaultResources).
+		WithESCoordinatorNodes(1, elasticsearch.DefaultResources)
+
+	test.Sequence(nil, test.EmptySteps, b).RunSequential(t)
```
pebrc, line 20: I forgot to `git add` this one

david-kow (approved):

sebgl (approved):
LGTM

pebrc:
Jenkins test this please

PR #4044, opened 2020-12-15
Add agent webhook to e2e tests

pebrc:
Missed this in https://github.com/elastic/cloud-on-k8s/pull/4035


david-kow (approved):
Ah, forgot about this as well.

PR #4046, opened 2020-12-16
Add Elastic Agent configRef E2E test

### LINKED ISSUE #4014: Add ConfigRef support to Elastic Agent controller
`configRef` should be support in the similar fashion that Beats supports it. Either `config` or `configRef` should be allowed, not both.

david-kow:
Adds missing E2E test for `configRef` functionality in Elastic Agent CRD.

Resolves #4014.

### DIFF
--- a/test/e2e/agent/config_test.go
+++ b/test/e2e/agent/config_test.go
@@ -7,13 +7,12 @@ package agent
 import (
 	"testing"
 
-	corev1 "k8s.io/api/core/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-
 	"github.com/elastic/cloud-on-k8s/test/e2e/test"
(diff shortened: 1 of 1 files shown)


test/e2e/agent/config_test.go
```diff
@@ -7,6 +7,9 @@ package agent
 import (
 	"testing"
 
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
```
pebrc, line 12: nit: imports formatting

pebrc (approved):
👍

test/e2e/agent/config_test.go
david-kow: I thought this is the right formatting (three groups) 🙂 Changing to two.

PR #4048, opened 2020-12-16
Fetch k8s version from cluster under E2E test

david-kow:
If we will need to release two `all-in-one` manifests we need to be able to test them first. To do that, we need to be able to determine which Kubernetes version we are testing against. This information is not available to the E2E test runner as it lives either in `setenvconfig` script or in deployer `config/plans.yaml`.

To avoid rewiring CI to pass it around we can dynamically fetch the Kubernetes version from the cluster at test. This PR introduces this behavior.


test/e2e/cmd/run/run.go
```diff
@@ -185,14 +186,42 @@ func (h *helper) initTestContext() error {
...
+	if err != nil {
+		panic(fmt.Sprintf("can't determine kubernetes version, err %s", err))
+	}
+
+	kubectlVersionResponse := struct {
+		ServerVersion map[string]string `json:"serverVersion"`
+	}{}
+
+	if err := json.Unmarshal([]byte(out), &kubectlVersionResponse); err != nil {
+		panic(fmt.Sprintf("can't determine kubernetes version, err %s", err))
```
pebrc, line 200:
```suggestion
		panic(fmt.Sprintf("can't determine Kubernetes version, err %s", err))
```
```suggestion
		panic(fmt.Sprintf("can't determine kubernetes version, err %s", err))
```

test/e2e/cmd/run/run.go
```diff
@@ -185,14 +186,42 @@ func (h *helper) initTestContext() error {
...
+	}
+
+	serverVersion, ok := kubectlVersionResponse.ServerVersion["gitVersion"]
+	if !ok {
+		panic("can't determine kubernetes version, gitVersion missing from kubectl response")
```
pebrc, line 205:
```suggestion
		panic("can't determine Kubernetes version, gitVersion missing from kubectl response")
```
```suggestion
		panic("can't determine kubernetes version, gitVersion missing from kubectl response")
```

pebrc (approved):
LGTM

PR #4049, opened 2020-12-16
Add support for DaemonSetUpdateStrategy

### LINKED ISSUE #3839: updateStrategy support for Beat CRD
## Proposal

Add support for updateStrategy on Beat CRD. Currently, only the PodTemplate (And replicas on a Deployment) can be modified.
https://github.com/elastic/cloud-on-k8s/blob/e1aa625567a7a8909ab01a9b7c0d3b22c46aca58/pkg/controller/beat/common/reconcile.go#L116

**Use case. Why is this important?**
When rolling out a Beat (Filebeat in our case) on large clusters (especially when using DaemonSet), the default updateStrategy's `maxUnavailable: 1` increases rollout times. In the case of Filebeat, unless the configuration is invalid, we can actually afford to rollover most of the pods (as long as the burst can be ingested by the ES after rollout).

**Note**
At first, I thought that `daemonSet` was mostly passed to the daemonSet itself as the fields description is:
```
Daemon Set:
  Description:  DaemonSet specifies the Beat should be deployed as a DaemonSet, and allows providing its spec. Cannot be used along with `deployment`. If both are absent a default for the Type is used.
```
However, as I searched in the code, it does not seem like we are actually allowed to provide any spec for the DaemonSet itself and also had a hard time finding documentation on what fields are allowed inside this `daemonSet` field. Is there any need for help on this subject?

pebrc:
Allow users to specify the DaemonSetUpdateStrategy when specifying a Beat to be deployed as a Daemonset.

Agent already supports it in master. Beat only allowed the specify the DeploymentStrategy but not the equivalent for DaemonSets. 

Fixes #3839


pkg/apis/beat/v1beta1/beat_types.go
```diff
@@ -79,7 +79,8 @@ type BeatSpec struct {
 }
 
 type DaemonSetSpec struct {
-	PodTemplate corev1.PodTemplateSpec `json:"podTemplate,omitempty"`
+	PodTemplate corev1.PodTemplateSpec         `json:"podTemplate,omitempty"`
+	Strategy    appsv1.DaemonSetUpdateStrategy `json:"strategy,omitempty"`
```
david-kow, line 83: I guess it's not required, but maybe add `// +kubebuilder:validation:Optional` for consistency with other fields.

david-kow (approved):
IIUC, `Test_reconcileDaemonSet` assertions are not combined into single test case for some more failure granularity? Seems each is testing different part, so they could be merged? LGTM either way 🙂

sebgl (approved):
LGTM

Issue #2328, opened 2019-12-25
Do not remove PVC when removing Elasticsearch resource with Operator


andrii-litvinov:
Relates to #1288. When removing Elasticsearch I don't necessary want to remove PVC, because it will lead to a data loss.

Expected behavior from k8s is to [keep PVCs](https://kubernetes.io/docs/tasks/run-application/delete-stateful-set/#persistent-volumes):
>Deleting the Pods in a StatefulSet will not delete the associated volumes. This is to ensure that you have the chance to copy data off the volume before deleting it. Deleting the PVC after the pods have left the terminating state might trigger deletion of the backing Persistent Volumes depending on the storage class and reclaim policy. You should never assume ability to access a volume after claim deletion.

It may be convenient for development, but in production there will be scenarios when users would want to remove the Elasticsearch and reinstall it agaian. Maybe there will be issues with upgrading the Opeartor or other Elasticsearch resources to new version. I am working with MongoDB Operator and there are plenty of scenarios when I would want to keep PVCs, for example when I convert sharded cluster to a replica set. And it is also expected behavior for other k8s users who worked with StatfulSets before.

Here is an open issue with the discussion in k8s repository [#55045](https://github.com/kubernetes/kubernetes/issues/55045). I would suggest leaving it up to a StatefulSet or at least introduce a flag whether to remove the PVCs or not. Or maybe make a decision about removing PVCs depending on there is it is scale-down or removal of a ReplicaSet.

anyasabo:
Since the reclaim policy is decided by the storage class, and you can specify the storage class for the PVCs, it is possible to retain volumes on scale down or delete. Can you go into more details on use cases when this is not sufficient?

There's an upstream issue here for allowing ssets to delete PVCs as well (also linked in the original issue you mentioned): https://github.com/kubernetes/kubernetes/issues/55045#issuecomment-559524385

andrii-litvinov:
In the documentation of StatefulSet they explicitly mention that one shouldn't rely on volumes:
>You should never assume ability to access a volume after claim deletion.

When PVC gets removed and new PVC is created with same name it gets bound to a new volume, not the existing one. So in this case I cannot have Elasticsearch attached to existing data automatically. I use https://github.com/rancher/local-path-provisioner.

barkbay:
[...]

PR #4050, opened 2020-12-16
Add support for VolumeClaimDeletePolicies for Elasticsearch clusters

pebrc:
First draft of a possible approach inspired by https://github.com/kubernetes/enhancements/pull/1915

Adds an new `volumeClaimDeletePolicy` to the Elasticsearch Spec on the cluster level (not per NodeSet)
```yaml
kind: Elasticsearch
metadata:
  name: es
spec:
  version: 7.10.1
  volumeClaimDeletePolicy: DeleteOnScaledownAndClusterDeletion
  nodeSets:
  - name: default
    count: 2
```

Possible values are `DeleteOnScaledownAndClusterDeletion` (default), `DeleteOnScaledownOnly`. 

 `RemoveOnScaledownAndClusterDeletion` relies on an owner reference pointing to the Elasticsearch resource to garbage collect PVCs once the Elasticsearch cluster has been deleted (existing behaviour). 

It also runs additional garbage collection to remove PVCs on each reconciliation that are no longer in use because either the whole node set has been removed or individual nodes have been scaled down (existing behaviour).

`RemoveOnScaledownOnly` means the PVCs are kept around after the cluster has been deleted. This is implemented by removing the owner reference. Removal of PVCs on scale down happens as before.

Switching from one to the other strategy is allowed and is implemented by avoiding the StatefulSet templating mechanism. This is mainly because the PVC template in StatefulSets are considered immutable and it would require StatefulSets to be recreated in order to change the PVC ownership. Instead the operator edits the PVCs after they have been created by the StatefulSet controller.

**Why not a per NodeSet setting?** I initially started with a setting per NodeSet, but it becomes tricky when users remove a whole node set as we then have no trace of the chosen policy anymore. We would need to add the policy choice to the PVC as an annotation or persist the policy choice in some other place so that we can prevent garbage collection if so desired by the user.

Fixes https://github.com/elastic/cloud-on-k8s/issues/2328


pkg/controller/elasticsearch/validation/claim_validation.go
```diff
@@ -0,0 +1,52 @@
...
+
+// noIllegalVolumeClaimDeletePolicyChange validates that the user is not changing VolumeClaimDeletePolicy on an existing
+// Elasticsearch cluster. This is implemented as a create validation and not as an update validation to allow usage from
+// within the controller as well where we don't have the context of the previous version of the spec anymore.
+// But we can infer a policy change by checking the ownerReferences on the PVCs in the knowledge that a Retain policy
+// must mean no owner reference on the PVCs and vice versa for the Remove* policies
+func noIllegalVolumeClaimDeletePolicyChange(k8sClient k8s.Client, es esv1.Elasticsearch) field.ErrorList {
+	var errs field.ErrorList
+	matchingLabels := label.NewLabelSelectorForElasticsearch(es)
+	var pvcs v1.PersistentVolumeClaimList
```
sebgl, line 23: We should inspect StatefulSets here, not PersistentVolumeClaims.
(PVCs _are_ mutable, but not sset spec claims)

sebgl: This also means we could hack around sset spec immutability using a similar mechanism as we use for pvc resize, and remove pvc ownerRefs. But it's maybe not worth it...

pkg/controller/elasticsearch/nodespec/statefulset.go
```diff
@@ -161,6 +162,16 @@ func setVolumeClaimsControllerReference(
 			continue
 		}
 
+		// TODO we can skip the whole setting of ownerReferences for the retain case if we can make sure via validation
```
sebgl, line 165: I'm wondering if the whole ownerRef on volume claim trick is not a mistake to start with.
As in: we could stop doing that (but keep existing ownerrefs for bwc), and gc pvcs "manually" instead.
By doing so users would be free to change their delete policy at any time (unless they rely on the bwc mode).

pebrc: So I have been thinking about that as well. The suggestion are IMO two different implementation variants: 

* we would not use the template mechanism of the stateful set to set `ownerReferences`, but instead set them directly on the PVC. This makes the creating of the Pod and its volume a multi step process with more potential for failure (what if we fail to set/remove the `ownerReference`).  The main advantage would be that users would not be constrained to chose one of the policies at cluster creation and could change their mind afterwards. 

* the other suggestion here is to manually garbage collect PVCs. We would need to annotate the PVCs with the current deletion policy so that we know what to garbage collect after the corresponding cluster has been deleted. This of course means we cannot use the sset template mechanism to place these annotations if we want to allow users to change the policy (same as when setting the `owerReference` directly). It therefore has similar failure cases (failing to set or update the annotation before garbage collection kicks in) The arguments in favour of using our own garbage collection that I can see are only that we are already doing a limited form of garbage collection during scale down of individual nodes and we could therefore standardize on one mechanism. The arguments against this are that `ownerReference`s are already implementing exactly this feature which we would duplicate. 

We could also add the 'going behind the sset's back' approach as a follow up PR to this work as it would be API compatible and we would only lift restrictions not introduce new ones.

pebrc: We discussed out of band and I will change this PR to set the `ownerReferences` without relying on the templating in the StatefulSets. This will free us to change policy as the user desires. I need to verify that this will not cause problems if the effective `ownerReference` in the PVC contradicts the volumeClaimTemplate e.g. if the StatefulSet controller will try to reconcile the `ownerReference`. If that turns out to be the case we would introduce an exception for existing clusters which could not change their `VolumeClaimDeletePolicy`

pebrc:
@sebgl I think this ready for review, with one exception: I am thinking about adding an e2e test following roughly the following playbook:
1. create ES cluster with retain policy, wait for it to healthy etc as per our standard test steps
2. add some data
3. delete the cluster
4. recreate the cluster
5. check if data still present
6. change policy to Remove*
7. delete cluster
8. verify pvcs are deleted (again a standard e2e test step we already have)

sebgl (changes requested):
There's one thing I don't understand well. The main use case for this feature as I understood it is to be able to do this combination:
- delete pvcs on scale down AND do not delete pvcs on cluster deletion

As in: if I scale my cluster down I don't care about old pvcs of old nodes (likely no data left in there anyway since we migrate it away in the downscale process), but if I remove the Elasticsearch resource entirely I want to be able to restore it as it was later.

Can I achieve this with one of the 3 options?

pkg/controller/elasticsearch/nodespec/statefulset.go
```diff
@@ -156,30 +145,10 @@ func setVolumeClaimsControllerReference(
...
-		// set finalizers on the resource.
-		// See https://github.com/elastic/cloud-on-k8s/issues/1884
-		refs := claim.OwnerReferences
-		for i := range refs {
-			refs[i].BlockOwnerDeletion = &f
 		}
 		claims = append(claims, claim)
 	}
-	return claims, nil
+	return claims
```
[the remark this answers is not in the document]
pebrc: > We'd want the controller: true bit from the sset controller when implemented to supersede our own controller bits?

I would say so and I would even say what we are doing rn is incorrect as the controller for the PVCs is the StatefulSet and not the Elasticsearch resource. Not sure if it has much practical relevance.

pebrc:
> delete pvcs on scale down AND do not delete pvcs on cluster deletion

No that's not possible with this implementation. The closest you get is using `Retain` which means that the more or less empty PVCs from nodes you scaled down are retained and if and when the cluster is recreated and scaled up again to its previous size they would rejoin. 

I tried to keep the behaviour aligned with the k8s KEP but if we think the use case you mentioned is a strong enough we could add another option.

sebgl:
I left a comment on https://github.com/kubernetes/enhancements/pull/2440 to advocate for the "preserve on deletion, remove on downscale" use case.

pebrc:
I think our naming is different enough from the KEP to allow us deviate also in behaviour. It seems that for Elasticsearch there is indeed little benefit in keeping the volume around on scale down. If anything this might cause trouble when rejoining if > 500 index deletions have happened while the node was offline. (Thanks to @\DaveCTurner for pointing this out). The indices will be considered dangling (index metadata on the old node is enough for that, so master eligible nodes are affected by that).

We could drop the `RemoveOnScaleDown` option and rename `Retain` to `RetainOnClusterDeletion` and change its behaviour to only affect cluster deletion and not scale downs.

This will make it a little bit harder to integrate with the KEP feature once it becomes available.  `RetainOnClusterDeletion` would require us to keep our GC approach in place. This is because in the KEP `apps.DeleteOnStatefulSetScaledown` implies `apps.DeleteOnStatefulSetDeletion` and sets a owner reference on the StatefulSet. For our `RemoveOnClusterDeletion` we could use the KEP feature.

pkg/controller/elasticsearch/driver/pvc_owner_test.go
```diff
@@ -0,0 +1,119 @@
...
+		return pvc
+	}
+
+	pvcFixturePtr := func(name string, ownerRefs ...string) *corev1.PersistentVolumeClaim {
+		pvc := pvcFixture(name, ownerRefs...)
+		return &pvc
+	}
+
+	updated := func(pvc corev1.PersistentVolumeClaim) corev1.PersistentVolumeClaim {
+		pvc.ResourceVersion = "1000" // fake client starts at 999
```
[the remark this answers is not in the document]
pebrc: I forgot that we had that. It works on single `RuntimeObjects` though so we need a version that takes slices.

pkg/apis/elasticsearch/v1/elasticsearch_types.go
```diff
@@ -66,6 +66,30 @@ type ElasticsearchSpec struct {
 	// RemoteClusters enables you to establish uni-directional connections to a remote Elasticsearch cluster.
 	// +optional
 	RemoteClusters []RemoteCluster `json:"remoteClusters,omitempty"`
+
+	// VolumeClaimDeletePolicy sets the policy for handling deletion of PersistentVolumeClaims for all NodeSets.
+	// Defaults to RemoveOnScaleDownPolicy.
+	// +kubebuilder:validation:Optional
+	VolumeClaimDeletePolicy VolumeClaimDeletePolicy `json:"volumeClaimDeletePolicy,omitempty"`
```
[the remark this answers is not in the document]
pebrc: TIL

pkg/controller/elasticsearch/driver/pvc_owner.go
```diff
@@ -0,0 +1,47 @@
...
+	ns := client.InNamespace(es.Namespace)
+	labelSelector := label.NewLabelSelectorForElasticsearch(es)
+	if err := c.List(context.Background(), &pvcs, ns, labelSelector); err != nil {
+		return fmt.Errorf("while listing pvcs to reconcile owner refs: %w", err)
+	}
+
+	for _, pvc := range pvcs.Items {
+		switch es.Spec.VolumeClaimDeletePolicyOrDefault() {
+		case esv1.RetainPolicy:
+			k8s.RemoveOwner(&pvc, &es)
```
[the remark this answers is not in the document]
pebrc: I don't think there is log statement by default in the client (I assume you were referring to our own `ReconcileREsource` implementation which we are not using here.). But I could add one here at debug level to be sure. I guess audit logs might help if necessary.

Makefile
```diff
@@ -460,6 +460,7 @@ e2e-local: LOCAL_E2E_CTX := /tmp/e2e-local.json
 e2e-local:
 	@go run test/e2e/cmd/main.go run \
 		--test-run-name=e2e \
+		--operator-image=$(OPERATOR_IMAGE) \
```
pebrc, line 463: https://github.com/elastic/cloud-on-k8s/issues/4285

sebgl (changes requested):
I think the API is a bit confusing.
Using `RetainOnClusterDeletion` instead of  `RemoveOnScaleDown` somewhat suggests we don't remove on scale down, but IIUC we do?
It's strange to offer `RemoveOnScaleDown` while other options are actually a superset of it (rather than something different).

I'm wondering if we should move away further from the StatefulSet KEP, and propose something a bit simpler that fits our needs:

```
spec:
    # retainVolumeClaims specifies whether ECK should retain PersistentVolumeClaims when the Elasticsearch resource is removed, allowing to recreate a cluster from existing volumes. If false, ECK deletes PersistentVolumeClaims upon Elasticsearch resource deletion. Defaults to false.
    retainVolumeClaimsOnDeletion: false
```

pebrc:
> I'm wondering if we should move away further from the StatefulSet KEP, and propose something a bit simpler that fits our needs:

I think the only argument against it is that a boolean flag is closed against extension vs the enum style attribute which is open to extension by simple addition of values. Now we could say we simply do not see another use case that would require us to add another policy that encodes different behaviour, ever.

thbkrkr:
[Seb's proposal](https://github.com/kubernetes/enhancements/pull/2440#issuecomment-785813242) to have "_PVCs automatically deleted on StatefulSet scaledown, but not on StatefulSet deletion_" led to the [update of the KEP](https://github.com/kubernetes/enhancements/pull/2545) (that has already received 2 lgtm).
Possible values for `PersistentVolumeClaimDeletePolicy` are now:
- Retain
- DeleteOnStatefulSetDeletionOnly
- DeleteOnScaledownOnly
- DeleteOnScaledownAndStatefulSetDeletion

I find these names rather clear and it covers all the usecases. 
What do you think about following this to facilitate the integration with the KEP once it becomes available?

pebrc:
@thbkrkr yes, I talked to @sebgl out of band and we agreed on following the KEP naming. Specifically we want to support two values in the initial implementation: 
`DeleteOnScaleDownOnly`
`DeleteOnScaledownAndClusterDeletion`

I will update this PR accordingly.

pkg/controller/elasticsearch/nodespec/statefulset.go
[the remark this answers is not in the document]
pebrc: It seems the `controller: true` property has relevance for adoption of orphaned resources but not for garbage collection. https://github.com/kubernetes/community/blob/master/contributors/design-proposals/api-machinery/controller-ref.md So for the purposes of this PR we are still good I believe.

pkg/controller/elasticsearch/driver/nodes.go
```diff
@@ -124,6 +124,10 @@ func (d *defaultDriver) reconcileNodeSpecs(
 		return results.WithError(err)
 	}
 
+	if err := reconcilePVCOwnerRefs(d.K8sClient(), d.ES); err != nil {
+		return results.WithError(err)
+	}
```
sebgl, line 129: let's reconcile owner refs before we garbage collect pvcs?
Otherwise I get funny error messages because we update owner refs after the pvc was deleted:

```
2021-03-02T16:58:56.354+0100	INFO	driver	Deleting PVC	{"service.version": "1.5.0-SNAPSHOT+45c993aa", "namespace": "default", "pvc_name": "elasticsearch-data-elasticsearch-sample-es-default-1"}
2021-03-02T16:58:56.438+0100	INFO	elasticsearch-controller	Ending reconciliation run	{"service.version": "1.5.0-SNAPSHOT+45c993aa", "iteration": 21, "namespace": "default", "es_name": "elasticsearch-sample", "took": "773.970731ms"}
2021-03-02T16:58:56.438+0100	ERROR	manager.eck-operator.controller.elasticsearch-controller	Reconciler error	{"service.version": "1.5.0-SNAPSHOT+45c993aa", "name": "elasticsearch-sample", "namespace": "default", "error": "while updating pvc during owner ref reconciliation: Operation cannot be fulfilled on persistentvolumeclaims \"elasticsearch-data-elasticsearch-sample-es-default-1\": StorageError: invalid object, Code: 4, Key: /registry/persistentvolumeclaims/default/elasticsearch-data-elasticsearch-sample-es-default-1, ResourceVersion: 0, AdditionalErrorMsg: Precondition failed: UID in precondition: cd93c74e-ed72-4661-8617-7f4f8d7b8cc4, UID in object meta: ", "errorCauses": [{"error": "while updating pvc during owner ref reconciliation: Operation cannot be fulfilled on persistentvolumeclaims \"elasticsearch-data-elasticsearch-sample-es-default-1\": StorageError: invalid object, Code: 4, Key: /registry/persistentvolumeclaims/default/elasticsearch-data-elasticsearch-sample-es-default-1, ResourceVersion: 0, AdditionalErrorMsg: Precondition failed: UID in precondition: cd93c74e-ed72-4661-8617-7f4f8d7b8cc4, UID in object meta: "}]}
```

sebgl (approved):
I did some manual testing (including backward compatibility) and it all seems to work as expected 👍 

I left a few nits, also let's not forget about user-facing docs on the website. Other than that, LGTM :)

pkg/controller/elasticsearch/driver/nodes.go
pebrc: Good point. I initially thought GC first made sense as it would whittle down the list of PVCs that need ownerRef updates but with a cached client that does not make sense and actually causes the problem you saw.

pkg/controller/elasticsearch/driver/pvc_owner_test.go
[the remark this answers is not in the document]
pebrc: I think its usefulness would be somewhat limited by the fact that most (all?) k8s API objects implement `runtime.Object` with pointer receivers, so callers will typically have to take pointers of the elements of the slice. 

But I removed to the custom `ResourceVersion` code and reused `comparison.AssertEqual`

pebrc:
> I left a few nits, also let's not forget about user-facing docs on the website. 

I raised https://github.com/elastic/cloud-on-k8s/issues/4287 as a reminder.

Issue #4051, opened 2020-12-16
Enterprise search revision slam when custom encryption key provided


MightyDjinn:
## Proposal

**Use case. Why is this important?**

## Bug Report
**What did you do?**
Deployed an enterprise search like so: 
```yaml
apiVersion: enterprisesearch.k8s.elastic.co/v1beta1
kind: EnterpriseSearch
metadata:
  name: enterprise-search-nonprod
  namespace: default
spec:
  version: 7.10.1
  image: 249300375295.dkr.ecr.us-east-1.amazonaws.com/vs-enterprisesearch:7.10.1
  count: 1
  elasticsearchRef:
    name: elasticsearch-nonprod
  http:
    service:
      spec:
        type: NodePort
    tls:
      selfSignedCertificate:
        disabled: true
  configRef:
    secretName: elastic-nonprod
```

this references the secret here:

```yaml
kind: Secret
apiVersion: v1
metadata:
  name: elastic-nonprod
stringData:
  enterprise-search.yml: |-
    ent_search.external_url: https://search.nonprod.compass.viasat.io
    workplace_search.custom_api_source.document_size.limit: 100000kb
    workplace_search.custom_api_source.total_fields.limit: 1000
    secret_management.encryption_keys: [8867d936149479963c1ed3713f327ee8272eebfc7748226197a3047796b6fa07]
```

The image is from the following dockerfile:

```Dockerfile
FROM docker.elastic.co/enterprise-search/enterprise-search:7.10.1
USER root
COPY ViasatCaBundle/viasatca/*.crt /etc/pki/ca-trust/source/anchors/
RUN update-ca-trust check > /dev/null |grep ENABLED || update-ca-trust force-enable
RUN update-ca-trust extract
```

**What did you expect to see?**
A deployed Enterprise search instance

**What did you see instead? Under which circumstances?**

Extremely rapid revision spew. Thousands per minute. A few pods come up and fail and crashloop/get progressed into next revision slam. 

**Environment**

* ECK version: 

	1.3

* Kubernetes information:

	insert any information about your Kubernetes environment that could help us:

	EKS (1.18) w/ cillium networking
	
	for each of them please give us the version you are using
```
[...]

PR #4052, opened 2020-12-17
Don't duplicate user-provided Enterprise Search encryption keys when reusing them

sebgl:
This commit fixes a bug in the operator when users provide their own
encryption keys.

When we retrieved existing encryption keys (to ensure we don't generate a new
one), we retrieved *all of them* including user-provided ones.
We then merge the user-provided configuration, which also includes user-provided
encryption keys. So everytime we reconcile the configuration, we duplicate
the user-provided encryption keys into the reconciled config secret. Because this
secret changes, all Pods get rotated automatically.
This results in a never-ending config update (growing forever) and pod rotation.

This commit fixes it by ensuring the operator always manages the first encryption key.
This first key is the only one that gets reused. User-provided keys are then appended
to the list of keys.

This allows the following use case:
- deploy ent search with no encryption key: you get a generated one
- add your own custom encryption keys: it is appended to the array after
the generated one. Transition is smooth. If we were to replace the generated
one in that case, Enterprise Search would not be able to decrypt data.

Fixes https://github.com/elastic/cloud-on-k8s/issues/4051.

### DIFF
--- a/pkg/controller/enterprisesearch/config.go
+++ b/pkg/controller/enterprisesearch/config.go
@@ -192,11 +192,26 @@ func getOrCreateReusableSettings(c k8s.Client, ent entv1beta1.EnterpriseSearch) 
 	} else if err := cfg.Unpack(&e); err != nil {
 		return nil, err
 	}
+
+	// generate a random secret session key, or reuse the existing one
 	if len(e.SecretSession) == 0 {
 		e.SecretSession = string(common.RandomBytes(32))
 	}
+
+	// generate a random encryption key, or reuse the existing one
+	// Encryption keys are stored in an array, so they can be rotated.
+	// When Enterprise Search decrypts a secret, it tries all encryption keys in the array, in order.
+	// When Enterprise Search rewrites a secret, it uses the latest encryption key in the array.
+	// We manage the first item of that array: it is randomly generated once, then reused.
+	// Users are free to provide their own encryption keys through the configuration:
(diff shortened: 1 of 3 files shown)


pkg/controller/enterprisesearch/config.go
```diff
@@ -192,11 +192,26 @@ func getOrCreateReusableSettings(c k8s.Client, ent entv1beta1.EnterpriseSearch)
...
+	// We manage the first item of that array: it is randomly generated once, then reused.
+	// Users are free to provide their own encryption keys through the configuration:
+	// in that case we still keep the first item we manage, user-provided keys will be appended to the array.
+	// This allows users to go from no custom key provided (use operator's generated one), to providing their own.
 	if len(e.EncryptionKeys) == 0 {
+		// no encryption key, generate a new one
 		e.EncryptionKeys = []string{string(common.RandomBytes(32))}
+	} else {
+		// encryption keys already exist, reuse the first ECK-managed one
+		e.EncryptionKeys = []string{e.EncryptionKeys[0]}
```
pebrc, line 214: The mental leap I had to make to understand this was that the user provided config is merged later into the `reusableSettings` outside of this function.

pebrc (approved):
LGTM

pkg/controller/enterprisesearch/config.go
sebgl: 👍 definitely deserves an additional comment here. Doing that.

Issue #2812, opened 2020-04-02
Support providing just a CA and key for transport certs


dbason:
## Proposal
The operator should support the option to just provide a CA cert and key and then generate the self signed transport certs from that.
**Use case. Why is this important?**
This is primarily to support cross cluster connections.  It would greatly simplify work of connecting up multiple clusters if we could provide a CA that would then be trusted by all of the clusters.  ~AFAIK the elasticsearch keygen tool supports this so hopefully it is just a matter of wiring this up?~

dbason:
I'm having a poke around the code now and I _think_ this looks doable.  Full disclosure I'm not a developer in any way shape or form, but I'll see if I can put together something to start this.

sebgl:
Another way of handling this (not sure if better) is to use the established naming convention.
I think if you pre-create the `<es-name>-es-transport-ca-internal` secret, the operator will detect it's already there and, if valid, will use it to sign certificates.

More generally I don't know if this feature should be ECK's responsibility, vs. using something such as CertManager. Happy to hear other people thoughts about it.

dbason:
@sebgl the comment about CertManager seems fair, but there would still need to be a change to allow customer transport certs.  At the moment there is only the option to provide custom http certs.

I'll change the title and initial post to reflect that this is explicitly for transport certs

MHenn1g:
This feature would be a great enhancement, if not to say necessity for one of our use-cases, in which we would need to link an external (kind of static) application to elasticsearch via the transport interface.
Not only would it enhance the automation, but at the same time the stability of operations.
At the moment, we are able to link the software, however, if the operator handles the ca management, this connection may break if the operator decides to rotate the ca unexpectedly.
Therefore I would like to ask, if there is any update regarding this request?

sebgl:
@dbason @MHenn1g We discussed this with the team and decided we should do it. No release date we could give yet, however.

deveshk0:
is there any plan to support the same with http layer?

pebrc:
> is there any plan to support the same with http layer?

Please see https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-tls-certificates.html#k8s-setting-up-your-own-certificate

PR #4053, opened 2020-12-18
Add support for a user-defined transport CA certificate and key

pebrc:
Fixes https://github.com/elastic/cloud-on-k8s/issues/2812

Introduces a new `certificate` attribute in the transport TLS options section of the Elasticsearch spec:

```yaml
kind: Elasticsearch
metadata:
  name: certs-test
spec:
  version: 7.10.1
  transport:
    tls:
      certificate:
        secretName: custom-ca
  nodeSets:
  - name: default
    count: 3    
    config:
      node.store.allow_mmap: false
```

Structure of the CA secret is analogous to the HTTP certificate secret we already support and expects a `tls.crt` and a `tls.key` entry contain the certificate as well as the private key in PEM format. 

If specified the user provided CA certificate will be used instead of a self-signed certificate authority generated by the operator. 

When a custom CA is specified the operator will remove the self-signed CA secret and not reconcile it anymore. The custom certificate will be parsed and validated in the reconciliation loop and not at admission. Any parsing errors will be surfaced as events (in addition to appearing in the logs) 

If we agree on the approach in this PR I can add the necessary user facing documentation either as an additional commit her or in a separate PR.


pkg/apis/elasticsearch/v1/elasticsearch_types.go
```diff
@@ -72,6 +72,22 @@ type ElasticsearchSpec struct {
...
+	TLS TransportTLSOptions `json:"tls,omitempty"`
+}
+
+type TransportTLSOptions struct {
+	// Certificate is a reference to a Kubernetes secret that contains the CA certificate
+	// and private key for generating node certificates.
+	// The referenced secret should contain the following:
+	//
+	// - `tls.crt`: The CA certificate.
+	// - `tls.key`: The private key for the CA certificate.
```
sebgl, line 85: We should also specify we expect PEM format.

pkg/controller/common/certificates/dynamic_watches.go
```diff
@@ -18,18 +18,27 @@ func CertificateWatchKey(namer name.Namer, ownerName string) string {
 
 // reconcileDynamicWatches reconciles the dynamic watches needed by the HTTP certificates.
```
sebgl, line 19: that's not only the http cert now, right?

pkg/controller/common/certificates/dynamic_watches.go
```diff
@@ -18,18 +18,27 @@ func CertificateWatchKey(namer name.Namer, ownerName string) string {
...
+	return ReconcileCustomCertWatch(dynamicWatches, CertificateWatchKey(namer, owner.Name), owner, tls.Certificate)
+}
+
+func ReconcileCustomCertWatch(
+	dynamicWatches watches.DynamicWatches,
+	watchKey string,
+	owner types.NamespacedName,
+	tlsSecret commonv1.SecretRef,
+) error {
 	// watch the Secret specified in es.Spec.HTTP.TLS.Certificate because if it changes we should reconcile the new
```
sebgl, line 30: same comment re. not only http

pkg/controller/elasticsearch/certificates/transport/ca.go
```diff
@@ -0,0 +1,79 @@
...
+
+	// 2. Assuming from here on the user wants to use custom certs and has configured a secret with them.
+
+	// Garbage collect the self-signed CA secret which might be left over from an earlier revision on a best effort basis.
+	_ = driver.K8sClient().Delete(&corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      certificates.CAInternalSecretName(esv1.ESNamer, esNSN.Name, certificates.TransportCAType),
+			Namespace: esNSN.Namespace,
+		},
+	})
```
sebgl, line 68: We should at least log the error here (if not `NotFound`). I'm fine with ignoring it, though.

pkg/controller/elasticsearch/certificates/transport/ca.go
```diff
@@ -0,0 +1,79 @@
...
+
+	// Try to parse the provided secret to get to the CA and to report any validation errors to the user.
+	ca, err := certificates.ParseCustomCASecret(*customCASecret)
+	if err != nil {
+		// Surface validation/parsing errors to the user via an event otherwise they might be hard to spot
+		// validation at admission would also be an alternative but seems quite costly and secret contents might change
+		// in the time between admission and reading the secret contents so we need to re-run validation here anyway.
+		driver.Recorder().Eventf(&es, corev1.EventTypeWarning, events.EventReasonValidation, err.Error())
+	}
```
sebgl, line 77: When this happens we move on and don't fail the reconciliation? Meaning something bad will likely happen down the line when we generate certs from that CA.
I think I would return an error here.

sebgl: Also we could move this code block before the self-signed CA deletion: if we fail to validate the user provided CA let's not delete our self-signed CA yet (slightly optimizing for the case where you start your cluster with no custom CA, it runs, then you add a wrong custom CA).

test/e2e/es/certs_test.go
```diff
@@ -7,24 +7,112 @@ package es
...
 	"github.com/elastic/cloud-on-k8s/pkg/utils/k8s"
 	"github.com/elastic/cloud-on-k8s/test/e2e/test"
 	"github.com/elastic/cloud-on-k8s/test/e2e/test/elasticsearch"
+	"github.com/stretchr/testify/require"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 )
 
+func TestCustomTransportCA(t *testing.T) {
```
sebgl, line 32: I think it would be nice to test the following flow:
- no custom CA provided: cluster available with generated CA
- then modify to provide a custom CA: cluster available with new CA
- remove the custom CA: cluster available with new generated CA

pkg/apis/elasticsearch/v1/elasticsearch_types.go
```diff
@@ -72,6 +72,22 @@ type ElasticsearchSpec struct {
...
+	Certificate commonv1.SecretRef `json:"certificate,omitempty"`
```
sebgl, line 86: I was expecting a `name + namespace` combination here, with us copying the content over another secret in the ES namespace.
But we don't do that for the custom http cert, I don't think anyone complained.
So 👍 with only a name to keep things simple.

sebgl (commented):
I'm 👍 with the approach. It's very nice that you could reuse a lot of the code constructs we already have!
I left a few comments worth considering.

pkg/controller/elasticsearch/certificates/transport/ca.go
pebrc: We do return the error already, don't we just one line below. It is just that I am recording the event before returning.  

💯 agreed on reordering the code blocks though 👍

pkg/controller/common/certificates/dynamic_watches.go
pebrc: No it still is. I use the new function `ReconcileCustomCertWatch` for transport (which this function also calls). The latter needs a godoc though, so 👍

test/e2e/es/certs_test.go
pebrc: I gave this another try and I think I have got something that is close to the flow you described.

pkg/controller/elasticsearch/certificates/transport/ca.go
```diff
@@ -0,0 +1,86 @@
...
+		CustomTransportCertsWatchKey(esNSN),
+		esNSN,
+		es.Spec.Transport.TLS.Certificate,
+	); err != nil {
+		return nil, err
+	}
+
+	customCASecret, err := certificates.GetSecretFromRef(driver.K8sClient(), esNSN, es.Spec.Transport.TLS.Certificate)
+	if err != nil {
+		return nil, err
```
sebgl, line 46: let's emit an event here

pkg/controller/elasticsearch/certificates/transport/ca.go
```diff
@@ -0,0 +1,86 @@
...
+	// 2. Assuming from here on the user wants to use custom certs and has configured a secret with them.
+
+	// Try to parse the provided secret to get to the CA and to report any validation errors to the user.
+	ca, err := certificates.ParseCustomCASecret(*customCASecret)
+	if err != nil {
+		// Surface validation/parsing errors to the user via an event otherwise they might be hard to spot
+		// validation at admission would also be an alternative but seems quite costly and secret contents might change
+		// in the time between admission and reading the secret contents so we need to re-run validation here anyway.
+		driver.Recorder().Eventf(&es, corev1.EventTypeWarning, events.EventReasonValidation, err.Error())
+		return nil, err
```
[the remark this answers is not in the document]
sebgl: answering myself: it's already logged somewhere up in the call chain.

pkg/controller/common/certificates/ca_secret.go
```diff
@@ -0,0 +1,42 @@
...
+	pubKeys, err := ParsePEMCerts(cert)
+	if err != nil {
+		return nil, pkgerrors.Wrapf(err, "can't parse CA certificate %s in %s/%s", CertFileName, s.Namespace, s.Name)
+	}
+	if len(pubKeys) != 1 {
+		return nil, pkgerrors.Errorf("only expected one PEM formated CA certificate in %s/%s", s.Namespace, s.Name)
+	}
+
+	certificate := pubKeys[0]
+	if !certificate.IsCA {
```
sebgl, line 38: I'm wondering if this is strictly required.
I generated a cert following a tutorial somewhere on the internet, and openssl did not set that IsCA bit in the asn1 extension, so I get an error here.
I tried removing this check: everything works fine, node transport certs are successfully generated from the given cert (even though it's not marked as CA).

I think we should rather be less restrictive here unless you can think of a good reason to keep this?

pebrc: You are right. I was actually thinking the same thing.

test/e2e/es/certs_test.go
```diff
@@ -7,24 +7,154 @@ package es
...
+			}
+		},
+	}
+
+	// tests the following sequence:
+	// 1. healthy cluster with self-signed operator provided CA
+	// 2. reconfigure to use custom certs but simulate user error on certificate setup: cluster to stay healthy
+	// 3. reconfigure with correct custom certificates
+	// 4. reconfigure back to self-signed operator provided CA
+	// TODO make TCP request to ES and verify that custom certs are actually used.
```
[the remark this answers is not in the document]
pebrc: I think I have something that works. I will add another commit.

pebrc:
Jenkins test this please

pkg/controller/elasticsearch/certificates/transport/ca.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+
+	// Garbage collect the self-signed CA secret which might be left over from an earlier revision on a best effort basis.
+	err = driver.K8sClient().Delete(&corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      certificates.CAInternalSecretName(esv1.ESNamer, esNSN.Name, certificates.TransportCAType),
+			Namespace: esNSN.Namespace,
+		},
+	})
+	if err != nil {
```
sebgl, line 81: Small improvement: I think we can do `if err != nil && !apierrors.IsNotFound(err)`

test/e2e/es/certs_test.go
```diff
@@ -7,24 +7,170 @@ package es
...
+								CommonName:         "eck-e2e-test-custom-ca",
+								OrganizationalUnit: []string{"eck-e2e"},
+							},
+						})
+						require.NoError(t, err)
+						caSecret := mkTestSecret(
+							certificates.EncodePEMCert(ca.Cert.Raw),
+							certificates.EncodePEMPrivateKey(*ca.PrivateKey),
+						)
+						_, err = reconciler.ReconcileSecret(k.Client, caSecret, nil)
```
sebgl, line 144: should likely be wrapped in `eventually()` to account for network glitches

sebgl (approved):
LGTM

sebgl (approved):

PR #4059, opened 2020-12-22
Re-enable Kibana telemetry for telemetry test 

pebrc:
Enable telemetry for the Kibana telemetry e2e tests. 

Add a new Kibana check step to verify that telemetry is off in all other cases (HTTP 404 on the telemetry API)


david-kow (approved):
I didn't run it, but it looks good.

Issue #3967, opened 2020-11-20
Upgrade-test-harness: cover operator upgrades with an existing ES cluster


sebgl:
Our upgrade test harness covers operator upgrades. After each upgrade it ensures the stack resources still have a green health (stack still ok after operator upgrade), then removes the stack and deploys a new one (creation successful with new operator).

I _think_ we may not wait long enough between operator upgrade and stack health check to detect problems that may be caused by reconciliations from the upgraded operator.

Ideas:
- Wait X minutes (arbitrary - 2/3 minutes sounds reasonable to ensure a reconciliation happened?)
- Update the stack spec (eg. `node.attr.timestamp: <timestamp>`) to force a rolling upgrade of Elasticsearch, wait until that rolling upgrade is over (all Pods recreated with a newer timestamp) + green health

PR #4060, opened 2020-12-22
Improve upgrade test harness

pebrc:
Keep one stack through all tests and upgrade it instead of checking the status of the old stack and then recreating a new one. 
This is to address the edge case where a managed resource becomes unhealthy due to an operator upgrade but that situation is not detected because the stack is deleted too quickly. 

By keeping the existing stack around we now have a single Elasticsearch cluster that lives through all upgrade tests. Any deterioration in health should be detected at the latest when the test harness is trying to upgrade the stack. 

There is one exception for the alpha version of ECK. Resources created by the alpha version are not reconciled by later versions of the operator so we delete the alpha stack after upgrading to the beta. 

A downside of this approach is that we are now testing two things together: operator upgrades and stack upgrades. 

Fixes https://github.com/elastic/cloud-on-k8s/issues/3967


pebrc:
Jenkins test this please

charith-elastic (commented):
LGTM overall. A couple of minor things I noticed during testing are:

- Alpha to Beta upgrade doesn't work. It appears to be due to: `Resource was created with older version of operator, will not take action`.  Indeed, the operator version annotation is `common.k8s.elastic.co/controller-version: 0.0.0-UNKNOWN`. I personally don't think this is a case worth testing since we have released so many other versions since the Alpha and are not obliged to support it. Maybe we should just remove the alpha from the test cases completely?
-  I now remember one of the reasons why I didn't go with in-place stack upgrades. They take ages! I had to increase the timeout to make the tests pass. It'd be good to increase the defaults defined in the flags to ensure that people don't get a bad experience the first time they try to run the suite.

pebrc:
> Alpha to Beta upgrade doesn't work. It appears to be due to: Resource was created with older version of operator, will not take action. Indeed, the operator version annotation is common.k8s.elastic.co/controller-version: 0.0.0-UNKNOWN. I personally don't think this is a case worth testing since we have released so many other versions since the Alpha and are not obliged to support it. Maybe we should just remove the alpha from the test cases completely?

But isn't that expected? According to https://www.elastic.co/guide/en/cloud-on-k8s/1.0-beta/k8s-upgrading-eck.html we don't support upgrade of v1alpha1 resources to beta or later without manual intervention. 

That is why I am removing the resources created in the alpha version explicitly here: 
https://github.com/elastic/cloud-on-k8s/pull/4060/files#diff-ffb295511c47b4830ec4bd1757a365e8bdeeb855d3fd6898ab0c6a63b5ce98c1R179

pebrc:
> I now remember one of the reasons why I didn't go with in-place stack upgrades. They take ages! I had to increase the timeout to make the tests pass. It'd be good to increase the defaults defined in the flags to ensure that people don't get a bad experience the first time they try to run the suite.

That is a good point. I have update the defaults a bit to settings that seemed to work for me. I was also wondering whether it would make sense to got down to single node test ES clusters. It would save at least a minute for each tested version...

charith-elastic:
> But isn't that expected? According to https://www.elastic.co/guide/en/cloud-on-k8s/1.0-beta/k8s-upgrading-eck.html we don't support upgrade of v1alpha1 resources to beta or later without manual intervention.

Yes, indeed. But, somehow the test failed multiple times for me. I just tried it again and now I see the following error:

```
"step": "RemoveESFinalizers", "error": "the API version in the data (elasticsearch.k8s.elastic.co/v1beta1) does not match the expected API version (elasticsearch.k8s.elastic.co/v1alpha1)"
```

I don't see an obvious reason why this update would break the finalizer removal code. Maybe it's the Kubernetes version I am using (1.18).  🤷🏼

charith-elastic (approved):
LGTM. I don't think the alpha to beta issue I experienced is worth holding up this PR for. We can look into it later.

PR #4063, opened 2020-12-23
Allow handshake erros during transport TLS cert check

pebrc:
We only want to inspect the peer certificates to verify the expected CA is among them. A successful handshake is not necessary and it seems not possible without presenting a client certificate at least on single node clusters.

There is still at bit of research to be done to fully understand why a handshake is possible on multi-node clusters without a client certificate and also why it seems to be possible when port-forwarding.


sebgl (approved):
LGTM. Please let me know if you manage to get to the bottom of the single-node/multi-node TLS handshake behaviour!

PR #4064, opened 2020-12-23
E2E test validate peer certificate chain during TLS handshake

pebrc:
Work around the issue that peer certificates are only populated in connection state after a successful handshake and that Elasticsearch 6.x is more restrictive when it comes to the presented client certificates. This is done by hooking into the TLS handshake and making sure the presented peer certificates can be validated against the expected CA certificate but do not require the full TLS handshake to succeed.

Tested with 6.x and 7.x.

### DIFF
--- a/test/e2e/es/certs_test.go
+++ b/test/e2e/es/certs_test.go
@@ -15,10 +15,9 @@ import (
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


sebgl (approved):

sebgl (approved):
👍 
(let's make sure the PR description matches what the code does re. validating the cert)

PR #4065, opened 2020-12-23
E2E adjust telemetry disablement for 6.x

pebrc:
Telemetry settings where namespaced below xpack in 6.x and had different effects than in 7.x. The telemetry API would still stay enabled even in the presence of the the setting.

I need to verify still that there no intermediate states in earlier 7.x versions (I tested 6.8 and 7.10)


pebrc:
It is more complicated: `telemetry.enabled` was only introduced in 7.5. Disabling it leads to an optimize run at least on 7.5 where I tested it and crashing Kibana on k8s. 

I will update the PR to disable the disablement of telemetry in a bit

PR #3864, opened 2020-10-21
Disable telemetry in E2E tests

david-kow:

PR #4066, opened 2020-12-23
Do not touch telemetry settings in e2e tests

pebrc:
On older versions of Kibana disabling telemetry means an optimize run will happen. While optimize can be disabled for CI with "optimize.enabled=false" it feels like this would hide other potential problems e.g. with insufficient default resource allocation.

This effectively reverts #3864 and https://github.com/elastic/cloud-on-k8s/pull/4059


sebgl (approved):

pebrc:
I raised https://github.com/elastic/cloud-on-k8s/issues/4067 to follow up on this

pebrc:
Jenkins test this please

PR #4069, opened 2020-12-29
Return the appropriate apiVersion for CustomResourceDefinition

isindir:
Fixing `helm lint` of the chart when using with k8s equal or greater then 1.18

### DIFF
--- a/deploy/eck-operator/charts/eck-operator-crds/templates/_helpers.tpl
+++ b/deploy/eck-operator/charts/eck-operator-crds/templates/_helpers.tpl
@@ -31,6 +31,17 @@ Create chart name and version as used by the chart label.
 {{- end }}
 
 {{/*
+Return the appropriate apiVersion for CustomResourceDefinition.
+*/}}
+{{- define "eck-operator-crds.CustomResourceDefinition.apiVersion" -}}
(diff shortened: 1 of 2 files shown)


charith-elastic:
AFAIK, CRD `v1beta1` is supported until Kubernetes 1.22.  Helm lint is being a bit overzealous here (see https://github.com/helm/helm/issues/8596). 

We have [implemented a solution](https://github.com/elastic/cloud-on-k8s/blob/1ee9dad656f09780a81ec7229d9a47dea0fce5aa/deploy/eck-operator/templates/_helpers.tpl#L128) similar to yours for some of the other deprecated APIs but the CRDs are a bit harder to fix that way. That is because we generate the CRDs from the code and copy that over to the chart on every build. That means we need to inject the templating code into the CRD generation process -- which is a bit more tricky.

isindir:
@charith-elastic Thanks for explanation. I actually expected CRD to be generated and thought this change may not be accepted. I'm completely fine if you close this PR without merging it. We have already introduced a hotfix in our CI to mitigate `helm lint` issue.

charith-elastic:
Thanks. I'll close this for now. The underlying problem is being tracked in #3886.

PR #4074, opened 2021-01-03
Updating kind.sh to accomodate new kind apiVersion

joshuacox:
Signed-off-by: Josh Cox <josh@webhosting.coop>

This is a small update that changes the apiVersion to follow the latest docs from kind:
https://kind.sigs.k8s.io/docs/user/configuration/

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -503,7 +503,7 @@ validate-jenkins-pipelines:
 # Kind specific targets #
 #########################
 KIND_NODES ?= 3
-KIND_NODE_IMAGE ?= kindest/node:v1.15.3
+KIND_NODE_IMAGE ?= kindest/node:v1.19.1
 KIND_CLUSTER_NAME ?= eck
 
 kind-node-variable-check:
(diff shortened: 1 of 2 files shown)


pebrc:
Hey @joshuacox thanks for your PR. The kind version is tied to the version we use in our CI pipeline: 

https://github.com/elastic/cloud-on-k8s/blob/master/.ci/Dockerfile#L12

Unfortunately we are not quite ready yet to move to kind 0.9 because we still want to run tests on k8s 1.12.

joshuacox:
@pebrc in any case the local development setup breaks on this https://github.com/elastic/cloud-on-k8s/blob/master/dev-setup.md

Should the docs be updated?  Or the Makefile itself?

pebrc:
@joshuacox yes, you are right we should mention that we currently require kind 0.8.0

pebrc:
I added support for both kind 0.9 and 0.8 in this PR https://github.com/elastic/cloud-on-k8s/pull/4122 This is mainly intended for our CI builds but if you want to use it `make bootstrap-kind` should work as long as the binaries are named accordingly e.g. kind-0.9.0 etc. I will update the developer docs as well.

PR #4086, opened 2021-01-05
Make Association status serialization backward compatible

david-kow:
E2E tests are failing on `TestAPMAssociationWhenReferencedESDisappears` (e.g. https://devops-ci.elastic.co/job/cloud-on-k8s-e2e-tests-aks/578/testReport/).

The reason for the failure is that event message changed after introducing multiple associations of the same type (#4036).

Previously, each event message contained only previous and current status (eg. `Pending`, `Established`). Now, `Status` for Agent CRD differs from others because it will also contain target namespaced name.

In the initial PR, Status serialization was generalized to support both single and multiple associations. 

Unfortunately, we should not be doing that because old CRDs don't carry the information about the reference target. Ie. we can't tell which target the Status is for. Initially, we implied that the Status is for the ref in the spec, but that's not always the case. For example, in the E2E test that's failing, we change ref namespace. Previously this produced:

`Association status changed from Established to Pending`

Now, the same message says:

`Association status changed from xxxx/es1: Established to xxxx/es1: Pending`

This is incorrect as `xxxx/es1` never existed and was never `Established`. We can't fix this without modifing old CRDs which we don't want to do for compatibility reasons. The best thing to do is to preserve the old behavior (the old message) even if it doesn't specify what was the target when status was `Established`.

For Agent, we do want to have full insight into the status. And we can have that because the `Status` itself carries the information about the target that status is for. In the similar situation as above, but for the Agent, we would have:

`Association status changed from default/es1: Established to xxxx/es1: Pending`

which gives the user all the information.

To achieve the above behavior, each Association that requires old behavior will create AssociationStatusMap that knows how to serialize itself to be backward compatible.

### DIFF
--- a/pkg/apis/common/v1/association.go
+++ b/pkg/apis/common/v1/association.go
@@ -13,6 +13,10 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 )
 
+const (
+	singleStatusKey = ""
+)
+
 // AssociationType is the type of an association resource, eg. for Kibana-ES association, AssociationType identifies ES.
 type AssociationType string
 
@@ -23,13 +27,25 @@ type AssociationStatus string
 // have a single Association of a given type (eg. single ES reference), this map will contain a single entry.
 type AssociationStatusMap map[string]AssociationStatus
 
-func NewAssociationStatusMap(selector ObjectSelector, status AssociationStatus) AssociationStatusMap {
+// NewSingleAssociationStatusMap creates an AssociationStatusMap that expects only a single Association. Using a
+// well-known key allows to keep serialization of the map backwards compatible.
+func NewSingleAssociationStatusMap(status AssociationStatus) AssociationStatusMap {
 	return map[string]AssociationStatus{
-		selector.NamespacedName().String(): status,
+		singleStatusKey: status,
 	}
 }
 
 func (asm AssociationStatusMap) String() string {
+	// check if it's single status map and return only status string if yes
+	// this allows to keep serialization backwards compatible
+	if len(asm) == 1 {
+		for key, value := range asm {
+			if key == singleStatusKey {
+				return string(value)
+			}
+		}
+	}
+
 	// sort by keys to make String() stable
 	keys := make([]string, 0, len(asm))
 	for key := range asm {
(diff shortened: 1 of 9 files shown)


pkg/apis/common/v1/association.go
```diff
@@ -23,13 +27,25 @@ type AssociationStatus string
...
 	}
 }
 
 func (asm AssociationStatusMap) String() string {
+	// check if it's single status map and return only status string if yes
+	// this allows to keep serialization backwards compatible
+	if len(asm) == 1 {
+		for key, value := range asm {
+			if key == singleStatusKey {
+				return string(value)
```
pebrc, line 44: Should we also add a unit test for this? Like 

```go
{
			name: "single established",
			statusMap: map[string]AssociationStatus{
				singleStatusKey: AssociationEstablished,
			},
			wanted: "Established",
		},
```

pkg/controller/association/reconciler.go
```diff
@@ -378,10 +378,16 @@ func (r *Reconciler) updateStatus(ctx context.Context, associated commonv1.Assoc
 	defer span.End()
 
 	oldStatus := associated.AssociationStatusMap(r.AssociationType)
+
+	// To correctly compare statuses without making the reconciler aware of singleton vs multiple associations status
+	// differences we: set new status, get it from associated and only then compare with the oldStatus. Setting the
+	// same status is harmless, setting a different status is fine as we have a copy of oldStatus above.
+	if err := associated.SetAssociationStatusMap(r.AssociationType, newStatus); err != nil {
```
pebrc, line 385: This has a smell to me. I wonder if it would be marginally better to have

```go
	SetAssociationStatusMap(typ AssociationType, statusMap AssociationStatusMap) (bool, error)
```
where `bool` would signal whether the status had changed compared to the old value, but I am OK with doing this in a later refactoring given that this PR is intended as a bug fix with minimal changeset.

pebrc (approved):
LGTM, the set-to-get thing is maybe the only thing that struck me as odd.

pkg/controller/association/reconciler.go
david-kow: It might make it a bit clearer, but the call to get still isn't avoided - to pass in the correct `newStatus` for `annotation.ForAssociationStatusChange` and `r.recorder.AnnotatedEventf` it needs to come from a call to `associated`. Otherwise it comes from the `Reconcile` func which doesn't have the knowledge of single vs multi and always supplies (sometimes incorrectly) namespaced name.

PR #4090, opened 2021-01-06
Update kind tests to close gaps in test matrix

pebrc:
Add 1.13,1.14, 1.15 and 1.18 to the tested versions. Stick with kind 0.8.0 for now

Remove stack image caching, which was out of date and I don't believe very useful given that we tests against multiple stack versions every day.


barkbay (approved):
LGTM

PR #4091, opened 2021-01-07
Increase integration/unit test timeout to 10 seconds

pebrc:
This increases the timeout used in the license and trial license integration tests as well as the observer tests to 10 seconds. My theory being that the immediately after booting the control plane for the integration tests it can take more than 5 seconds before reconciliation events reach the controllers. 

This should resolve https://github.com/elastic/cloud-on-k8s/issues/3495 and https://github.com/elastic/cloud-on-k8s/issues/3617

### DIFF
--- a/pkg/utils/test/retry.go
+++ b/pkg/utils/test/retry.go
@@ -14,7 +14,7 @@ import (
 
 // Default values to be used for testing purpose.
 const (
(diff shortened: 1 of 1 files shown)


barkbay (approved):
👍 Let's give it a try.

PR #4078, opened 2021-01-04
Add a doc section about stuck master scale down in ES <= 7.8

pebrc:
Fixes #1986 

Preview: https://cloud-on-k8s_4078.docs-preview.app.elstc.co/guide/en/cloud-on-k8s/master/k8s-common-problems.html#k8s-common-problems-scale-down

PR #4094, opened 2021-01-07
[1.4] Add a doc section about stuck master scale down in ES <= 7.8 (#4078)

pebrc:
Backports the following commits to 1.4:
 - Add a doc section about stuck master scale down in ES <= 7.8 (#4078)


barkbay (approved):
LGTM

PR #4099, opened 2021-01-11
Do not delete StatefulSets associated to empty nodeSets

### LINKED ISSUE #4077: Empty nodeSet compatibility
It is not possible to scale down a nodeSet to 0 node at the moment. It might be interesting to be able to do so in the context of autoscaling and machine learning nodes. It would allow ECK to spin up some machine learning nodes only when required.

barkbay:
This PR prevents the Elasticsearch controller to delete a StatefulSet associated to an empty nodeSet.

Fix #4077

### DIFF
--- a/config/crds/all-crds.yaml
+++ b/config/crds/all-crds.yaml
@@ -1430,7 +1430,7 @@ spec:
                   count:
                     description: Count of Elasticsearch nodes to deploy.
                     format: int32
-                    minimum: 1
+                    minimum: 0
                     type: integer
                   name:
                     description: Name of this set of nodes. Becomes a part of the
--- a/config/crds/bases/elasticsearch.k8s.elastic.co_elasticsearches.yaml
+++ b/config/crds/bases/elasticsearch.k8s.elastic.co_elasticsearches.yaml
@@ -393,7 +393,7 @@ spec:
                     count:
                       description: Count of Elasticsearch nodes to deploy.
                       format: int32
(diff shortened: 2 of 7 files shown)


charith-elastic (approved):
LGTM. 
This is not strictly related to your change, but, `StatefulSetList` seems to be used as a Set in several places ( the existing `GetByName` method, the `Names` method you introduced, and [transport certificate reconciler](https://github.com/elastic/cloud-on-k8s/blob/63064b6450272af43602de31ed2bac08672cb74e/pkg/controller/elasticsearch/certificates/transport/reconcile.go#L37), to name a few). Would it make sense to redefine `StatefulSetList` as  `map[string]appsv1.StatefulSet` or as a struct that contains both the slice and the set? I guess that normally there would only be one or two nodeSets and this optimization probably wouldn't matter. It is in the hot path though, and in extreme cases, it could reduce the amount of garbage created.  WDYT?

barkbay:
@charith-elastic Thanks ! And agreed, I think your proposal makes sense.
My only concern would be that `StatefulSetList` can't be replaced directly with a map since it is expected to have [a sorted List](https://github.com/elastic/cloud-on-k8s/blob/e52945b66f33debc67c0c1028714eec3216edfed/pkg/controller/elasticsearch/sset/list.go#L34). I would like to take more time to assess the impact of this change and make a clean refactoring. I would raise a dedicated PR. I'll create an issue.

Issue #4092, opened 2021-01-07
TestKubernetesIntegrationRecipe is flaky in IPv6 Kind cluster


pebrc:
```
        	Error Trace:	utils.go:84
        	Error:      	Received unexpected error:
        	            	404 Not Found: no such index [metrics-kubernetes.event-k8s]
        	Test:       	TestKubernetesIntegrationRecipe/ES_data_should_pass_validations
{"log.level":"error","@timestamp":"2021-01-07T01:56:27.485Z","message":"stopping early","service.version":"0.0.0-SNAPSHOT+00000000","service.type":"eck","ecs.version":"1.4.0","error":"test failure","error.stack_trace":"github.com/elastic/cloud-on-k8s/test/e2e/test.StepList.RunSequential\n\t/go/src/github.com/elastic/cloud-on-k8s/test/e2e/test/step.go:43\ngithub.com/elastic/cloud-on-k8s/test/e2e/test/helper.RunFile\n\t/go/src/github.com/elastic/cloud-on-k8s/test/e2e/test/helper/yaml.go:156\ngithub.com/elastic/cloud-on-k8s/test/e2e/agent.runBeatRecipe\n\t/go/src/github.com/elastic/cloud-on-k8s/test/e2e/agent/recipes_test.go:96\ngithub.com/elastic/cloud-on-k8s/test/e2e/agent.TestKubernetesIntegrationRecipe\n\t/go/src/github.com/elastic/cloud-on-k8s/test/e2e/agent/recipes_test.go:58\ntesting.tRunner\n\t/usr/local/go/src/testing/testing.go:1123"}
    --- FAIL: TestKubernetesIntegrationRecipe/ES_data_should_pass_validations (1800.00s)
```

or 

```
  utils.go:84: 
        	Error Trace:	utils.go:84
        	Error:      	Received unexpected error:
        	            	404 Not Found: no such index [metrics-kubernetes.apiserver-k8s]
        	Test:       	TestKubernetesIntegrationRecipe/ES_data_should_pass_validations
[...]

PR #4100, opened 2021-01-11
Disable k8s apiserver and event data stream checks for Agent e2e tests

pebrc:
This disables the `kubernetes.events` and `kubernetes.apiserver` data sets in the Agent tests. While I have a good explanation why the event data set might not produce any data, I don't fully understand the failure of the apiserver data set. There should be enough requests to the apiserver to generate some metrics data. 

We have https://github.com/elastic/cloud-on-k8s/issues/4092 to investigate this fully and I would like to disable those to checks to remove the noise until this has been better understood


david-kow (approved):
LGTM, maybe add a link to the issue in the comment for more context?

PR #4113, opened 2021-01-13
Allow non-root user to run Elastic Agent process

### LINKED ISSUE #4107: Unable to run agent as non-root user
Removing the `securityContext.runAsUser: 0` stanza leads to an agent crash bootlooping because it cannot read the configuation file we mount in `/etc/agent.yaml`  with 0600 permissions.

We should allow non-root users to run the agent even though root user privileges would probably be required for many use cases that involve accessing host path volumes etc. 

I think the correct permissions should be 0440

pebrc:
Adjust file permissions on mounted configuration file to be read-only and group readable.

Also removes the `runAsUser: 0` setting from the recipes. 

I tested on OCP and GCP with the recipes which went OK.

Fixes #4107

### DIFF
--- a/config/recipes/elastic-agent/kubernetes-integration.yaml
+++ b/config/recipes/elastic-agent/kubernetes-integration.yaml
@@ -13,10 +13,6 @@ spec:
         serviceAccountName: elastic-agent
         containers:
         - name: agent
-          securityContext:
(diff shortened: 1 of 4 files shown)


david-kow (approved):
LGTM

PR #4121, opened 2021-01-14
Introduce E2E test tags and exclude Agent recipe from Kind IPv6 tests

charith-elastic:
This commit introduces a way to tag test runs with metadata that can be used to skip tests.

It is necessitated by #4092 where the test only fails on the Kind IPv6 cluster. This PR also includes a patch to skip `TestKubernetesIntegrationRecipe` when run in that environment.

Partial workaround for #4092


pebrc (approved):
LGTM I wonder if we should unify this with other environmental information we have in the TestContext like for example `isOCPCluster`

charith-elastic:
I raised #4127 to remove other custom test attribute flags.

PR #4122, opened 2021-01-14
Support both kind 0.8 and 0.9

### LINKED ISSUE #4117: Test k8s 1.19 and 1.20
The most pragmatic approach would probably be to introduce two more kind tests for the newer k8s versions. 

The only complication is that we need kind 0.9.0 for that which does not support versions < 1.13 for single node clusters and version < 1.14 for multi-node clusters. 

One idea is to download/install both kind binaries into the our ci Docker image and have two kinds CI jobs one for the older versions using kind 0.8 and one for the newer versions using kind 0.9. We also would have to make the kind script aware of the version of kind and use a different Cluster resource for each version. And finally we need to make sure to use the correct kind node images for each version as one of the main version differences between 0.8 and 0.9 is a change in the node image format. 

* kind 0.8.x: 1.12, 1.13
* kind 0.9.x: 1.14, 1.15, 1.16, 1.17, 1.18, 1.19, 1.20

pebrc:
This is to allow testing of all supported k8s versions. Given that we still want to support k8s 1.12 and 1.13, kind 0.8 is the only option to to test against those versions. At the same time we want to test the latest version of k8s as well even before they are available on the Cloud providers, for that we need kind 0.9

I ended up rewriting `kind.sh` in Go as it seemed to offer better templating and the script was getting more complicated than I wanted to. 

Fixes #4117


.ci/pipelines/e2e-tests-kind-k8s-versions.Jenkinsfile
```diff
@@ -109,18 +109,40 @@ pipeline {
...
+                        label 'eck'
+                    }
+                    steps {
+                        unstash "source"
+                        script {
+                            runTests(lib, failedTests, "kindest/node:v1.20.0@sha256:b40ecf8bcb188f6a0d0f5d406089c48588b75edc112c6f635d26be5de1c89040", "0.9.0", "ipv4")
+                        }
+                    }
+                }
+                stage("1.20.0 IPv6") {
```
pebrc, line 138: I am not confident that IPv6 will work, in my tests the kind clusters failed to setup the CNI networking correctly, but I am still looking.

pebrc: Found the issue https://github.com/kubernetes-sigs/kind/commit/60074a9e67ddc8d35d3468ab137358b62a4cf723

hack/kind/main.go
```diff
@@ -0,0 +1,263 @@
...
+		// check that the kind executable is installed in the right version
+		binaryName := "kind-" + k.kindVersion
+		path, err := exec.LookPath(binaryName)
+		if err != nil {
+			return err
+		}
+		// populate the kind struct with version specific information
+		k.binary = path
+		k.apiVersion = "kind.sigs.k8s.io/v1alpha3"
+		if strings.HasPrefix(k.kindVersion, "0.9") {
```
charith-elastic, line 86: Should we do a semver compare here to check that the version is greater than 0.8? Otherwise, if we forget to update this line with the next release of Kind (say. 0.10.0), the API version would fall back to v1alpha3.

hack/kind/main.go
```diff
@@ -0,0 +1,263 @@
...
+  - role: control-plane
+{{- range .WorkerNames }}
+  - role: worker
+{{- end}}
+`
+
+var k kind
+
+type startArgs struct {
+	IPFamily string
```
charith-elastic, line 44: Why is this the only public field?

charith-elastic (commented):
Apart from a nit about the version check, LGTM 👍🏼

Now that the driver script is written in Go, I wonder if we can simply use Kind as a library instead of installing the binaries onto the container image. With Go modules, it should be possible to have both Kind 0.8 and 0.9 as dependencies. 🤔

hack/kind/main.go
pebrc: Good point I was cutting corners here.

hack/kind/main.go
pebrc: Because it's the only one I am using in the template ...

charith-elastic (approved):

PR #4133, opened 2021-01-18
E2E: separate SCC from monitoring configuration

pebrc:
The SCC is not only need by the monitoring configuration but also for the tests itself. Running the tests locally or without monitoring configured will therefore fail on OCP. This change separates the SCC from the monitoring configuration.


charith-elastic (approved):
LGTM

PR #4134, opened 2021-01-18
Add go build constraints to e2e tests

### LINKED ISSUE #4085: Making it easier to run E2E tests for a specific Stack application
Currently it is not possible to limit the E2E test runs to just one Stack application. Being able to do this is useful when adding application-specific features to ECK (or when each application is introducing a feature at different schedules). 

We could do this in one of the following ways;
- Follow a naming convention that makes it easier to craft a regex to match all related tests.
- Use sub tests (e.g. have a single `TestElasticsearch` function that invokes all other Elasticsearch tests as sub tests).

pebrc:
This would allow us to test just subsets of our e2e tests e.g. just specific applications. The approach chosen here has the disadvantage that the selection of tests to run is baked into the e2e Docker image and not changeable at runtime. 

A few alternatives to consider:

Filter directories/packages in run.sh
* \+ test authors do not need to take any additional action
* \- very coarse grained
* \- shell script based (tricky to read/maintain), more env variables 

Go build tags
* \+ integrated with go tool chain, works even when running tests manually outside of shell script
* \- test authors need to add the tags
* \- needs a different build/Docker image  for every test run 

Hierarchical tests by nesting inside parent tests
* \+ can be very fine grained 
* \- need to be excluded on regular runs (e.g. with a tag :-) 
* \- needs awareness and inclusion during development otherwise tests won't run 

I am happy to reconsider another approach if we think one of the others would be preferable.

Fixes #4085


charith-elastic (commented):
LGTM 👍🏼 
I think the `e2e-compile` Make target needs to be updated too. It's quite handy when refactoring things.

pebrc:
Jenkins test this please

charith-elastic (approved):
LGTM.

PR #4156, opened 2021-01-27
Fix Dockerfile platform argument for e2e image

pebrc:
It looks like we missed this Dockerfile when we fixed the main build in #4081 . The e2e test job on ARM is failing.


charith-elastic (approved):
Good catch! LGTM.

PR #4157, opened 2021-01-27
Refer to eck.k8s.elastic.co/managed in doc and annotator.sh

barkbay:
This PR updates the default label used by the annotator script from `common.k8s.elastic.co/pause` to `eck.k8s.elastic.co/managed`
The new annotation is available since 1.1.0 and it might be a bit confusing for the user to still see that legacy annotation while our documentation refers to the [new one](https://www.elastic.co/guide/en/cloud-on-k8s/1.3/k8s-troubleshooting-methods.html#k8s-exclude-resource).

The PR also updates the documentation which is supposed to cover upgrades between 2 GA releases and not only Beta to GA. I think the former should be the norm today.

### DIFF
--- a/docs/operating-eck/upgrading-eck.asciidoc
+++ b/docs/operating-eck/upgrading-eck.asciidoc
@@ -23,7 +23,7 @@ Note that the release notes and highlights only list the changes since the last 
 
 CAUTION: The upgrade process results in an update to all the existing managed resources. This triggers a rolling restart of all Elasticsearch and Kibana pods. <<{p}-beta-to-ga-rolling-restart, Guidance>> is available on controlling this process more gracefully.
 
-IMPORTANT: If you are running Openshift 3.11 or a Kubernetes version older than 1.13, skip to the <<{p}-ga-openshift, upgrade instructions for older Kubernetes releases>>.
+IMPORTANT: If you are running Openshift 3.11 or a Kubernetes version older than 1.13, and want to upgrade from beta, skip to the <<{p}-ga-openshift, upgrade instructions for older Kubernetes releases>>.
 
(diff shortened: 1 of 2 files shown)


charith-elastic (approved):
LGTM

barkbay:
Thanks @charith-elastic for the review 🙇 

@alaudazzi these are only some small adjustments but I would be glad to have your ✅ if you can, thanks !

docs/operating-eck/upgrading-eck.asciidoc
```diff
@@ -23,7 +23,7 @@ Note that the release notes and highlights only list the changes since the last
 
 CAUTION: The upgrade process results in an update to all the existing managed resources. This triggers a rolling restart of all Elasticsearch and Kibana pods. <<{p}-beta-to-ga-rolling-restart, Guidance>> is available on controlling this process more gracefully.
 
-IMPORTANT: If you are running Openshift 3.11 or a Kubernetes version older than 1.13, skip to the <<{p}-ga-openshift, upgrade instructions for older Kubernetes releases>>.
+IMPORTANT: If you are running Openshift 3.11 or a Kubernetes version older than 1.13, and want to upgrade from beta, skip to the <<{p}-ga-openshift, upgrade instructions for older Kubernetes releases>>.
```
alaudazzi, line 26:
```suggestion
IMPORTANT: If you are running Openshift 3.11 or a Kubernetes version older than 1.13, and want to upgrade from beta, jump to the <<{p}-ga-openshift, upgrade instructions for older Kubernetes releases>>.
```

alaudazzi (approved):
Left a minor suggestion, otherwise LGTM.

PR #4158, opened 2021-01-27
[Autoscaling] Add Elasticsearch API client

barkbay:
This PR updates the Elasticsearch client code so it can interact with the autoscaling API.

I'm trying to split the autoscaling feature into comprehensible chunks, this PR relies on https://github.com/elastic/cloud-on-k8s/pull/4148 but it can still be reviewed to discuss naming.


pkg/controller/elasticsearch/client/autoscaling.go
```diff
@@ -0,0 +1,133 @@
...
+	Total AutoscalingResources `yaml:"total" json:"total,omitempty"`
+}
+
+type AutoscalingNodeInfo struct {
+	Name string `json:"name"`
+}
+
+// IsEmpty returns true if all the resource values are empty (no values in the API response).
+// 0 is considered as a value since deciders are allowed to return 0 to fully scale down a tier.
+func (rc AutoscalingCapacityInfo) IsEmpty() bool {
```
charith-elastic, line 81: Nit: odd name for the receiver (`rc`)

pkg/controller/elasticsearch/client/autoscaling.go
```diff
@@ -0,0 +1,133 @@
...
+	return rc.Node.IsEmpty() && rc.Total.IsEmpty()
+}
+
+// AutoscalingCapacity models a capacity value as received by Elasticsearch.
+type AutoscalingCapacity int64
+
+// Value return the int64 value returned by Elasticsearch. It returns 0 if no value has been set by Elasticsearch.
+func (e *AutoscalingCapacity) Value() int64 {
```
charith-elastic, line 89: Same nit: odd receiver name (`e`)

pkg/controller/elasticsearch/client/autoscaling.go
```diff
@@ -0,0 +1,133 @@
...
+}
+
+// IsZero returns true if the value is greater than 0.
+func (e *AutoscalingCapacity) IsZero() bool {
+	return e.Value() == 0
+}
+
+type AutoscalingResources struct {
+	Storage *AutoscalingCapacity `yaml:"storage" json:"storage,omitempty"`
+	Memory  *AutoscalingCapacity `yaml:"memory" json:"memory,omitempty"`
```
charith-elastic, line 108: I suppose it's a quirk of the API that these are just integer values without units?

pkg/controller/elasticsearch/client/test_fixtures/autoscaling.go
```diff
@@ -0,0 +1,140 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
```
charith-elastic, line 1: Can't we put this and all the tests under the special `client_test` package because it is only needed during tests? Alternatively, it can be stored as a JSON file in the `testdata` directory and loaded at runtime?

pkg/controller/elasticsearch/client/v6.go
```diff
@@ -148,6 +149,22 @@ func (c *clientV6) DeleteVotingConfigExclusions(ctx context.Context, waitForRemo
 	return errors.New("Not supported in Elasticsearch 6.x")
 }
 
+func (c *clientV6) DeleteAutoscalingPolicies(_ context.Context) error {
+	return errors.New("Not supported in Elasticsearch 6.x")
+}
+
+func (c *clientV6) CreateAutoscalingPolicy(_ context.Context, _ string, _ esv1.AutoscalingPolicy) error {
+	return errors.New("Not supported in Elasticsearch 6.x")
```
charith-elastic, line 157: Nit: only because my linter complains; maybe we can define this error once? I don't feel strongly about it though.

charith-elastic (commented):
LGTM. There are a couple of nits but I don't feel very strongly about them.

pkg/controller/elasticsearch/client/test_fixtures/autoscaling.go
barkbay: This file belongs to the `fixtures` package, maybe I'm missing something but I can't find the `client_test` package.

I didn't want to introduce a `testdata` directory here only to be consistent with what we do [for other tests](https://github.com/elastic/cloud-on-k8s/blob/633982dd077c1d41519e9aa6995159615c06f92c/pkg/controller/elasticsearch/client/test_fixtures/nodes.go). But I agree that some of those tests would deserve it.

pkg/controller/elasticsearch/client/autoscaling.go
barkbay: Yes, Elasticsearch does not return units in the result.

pkg/controller/elasticsearch/client/test_fixtures/autoscaling.go
charith-elastic: It's a Go testing idiom that we don't use in the ECK code base much. You can change the package name of `*_test.go` files to `<package>_test` and that's legal. So, what I am suggesting is that instead of using `package client` in the test files, we can switch to `package client_test` to keep the test code separate.

We use `testdata` in several places as well: https://github.com/elastic/cloud-on-k8s/tree/master/pkg/controller/common/certificates/testdata

charith-elastic (approved):

barkbay:
jenkins test this please

> ERROR: failed commit on ref "manifest-sha256:e3712b8ba68e...86cd1c3e316": unexpected status: 400 Bad Request

:no_mouth:

PR #4170, opened 2021-01-28
E2E: Update Ansible image tag in ocp3 deployer

pebrc:
I _think_ this is one of the few occasions where we want to use `latest` but I am happy to hear dissenting voices.


charith-elastic (commented):
I wasn't aware that OCP3 could be deployed through the Deployer 🤦🏼 

I think the danger of `latest` (or any other generic tag) is that there's no guarantee that Docker will always pick the most up-to-date container image. Could we maybe force a `docker pull` in the script to ensure that `latest` actually points to latest?

pebrc:
@charith-elastic you are right. But I think in that case using the specific tag is simpler. I have changed it. We don't update this image too often.

charith-elastic (approved):

PR #4173, opened 2021-01-29
[Autoscaling] Add Elasticsearch autoscaling controller

barkbay:
This PR introduces the Elasticsearch autoscaling controller.

This new controller manages the autoscaling policies, periodically polls the `_autoscaling/capacity` API and adjusts automatically the resources in the node sets.

Some parts of the code are not optimized, this PR is large, I tried to keep something which is easy to understand.
I also tried to add comments where needed to help the reader/reviewer.

Starting by the `reconcileInternal` function in `pkg/controller/autoscaling/elasticsearch/driver.go` might be a good starting point for a review.

The following improvements will be added in dedicated PRs:

- [ ] Add a warning in case of overlapping roles.
- [ ] Manage limits using ratios, only resource requests are supported for now.
- [ ] Add a stabilization period
- [ ] Make the poll period a setting, current default is 60s

## Testing

This PR relies on the [ES autoscaling client](https://github.com/elastic/cloud-on-k8s/pull/4158) and requires to activate a trial (or a valid license)

Deciders are documented [here](https://www.elastic.co/guide/en/elasticsearch/reference/7.x/autoscaling-deciders.html).

### Data tier

Unfortunately I don't have strong advices for testing storage deciders.
You can try to use Rally, but note that it will only exercise the reactive decider:

```
esrally race --pipeline=benchmark-only --track=eventdata --track-repository=eventdata --challenge=index-logs-fixed-daily-volume \
             --track-params="bulk_size:100,number_of_replicas:1,bulk_indexing_clients:1,daily_logging_volume:1GB" \
             --target-hosts=https://${ES}:9200 \
             --client-options="use_ssl:true,verify_certs:false,basic_auth_user:'elastic',basic_auth_password:$PASSWORD"
```

You can also deploy some agents to create datastreams.

### ML

You can use [the following API call](https://gist.githubusercontent.com/barkbay/d956d408366da14e74108b8a22205091/raw/9ca9a95d6d003a0d4ceb9b25e5e5bc0f14d16cfc/Create%2520ML%2520Job) to create a ML job _(it requires the sample data flights)_. Then do a `POST _ml/data_frame/analytics/mljob/_start` to start the job.

## Logs and status

The expected resources for the node sets and some useful messages are printed in the logs but also stored in the `elasticsearch.alpha.elastic.co/autoscaling-status` annotation. You can use the following command to retrieve it:`kubectl get -n elasticsearch.elasticsearch.k8s.elastic.co/mycluster -o jsonpath='{.metadata.annotations.elasticsearch\.alpha\.elastic\.co/autoscaling-status}' | jq`


pebrc (commented):
I am still working through this but I thought I share a first set of comments. All minor.

pkg/controller/autoscaling/elasticsearch/status/status.go
```diff
@@ -0,0 +1,212 @@
...
+	HorizontalScalingLimitReached PolicyStateType = "HorizontalScalingLimitReached"
+	MemoryRequired                PolicyStateType = "MemoryRequired"
+	EmptyResponse                 PolicyStateType = "EmptyResponse"
+	StorageRequired               PolicyStateType = "StorageRequired"
+	NoNodeSet                     PolicyStateType = "NoNodeSet"
+)
+
+type Status struct {
+	// PolicyStatus is used to expose state messages to user or external system
+	AutoscalingPolicyStatuses []AutoscalingPolicyStatus `json:"policies"`
```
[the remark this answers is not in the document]
barkbay: I'll fix the comment. I'm reusing `policies` here to match the root `policies` [element in the spec](https://github.com/elastic/cloud-on-k8s/blob/fa7f098583f84888044813d257d5b0ff4500271e/config/recipes/autoscaling/elasticsearch.yaml#L9).

For example, the following spec:

```json
{
	"policies": [{
		"name": "di",
		"roles": ["data", "ingest", "transform"],
		"resources": {
			"nodeCount": { "min": 3, "max": 8},
			"cpu": {"min": 2,"max": 8},
			"memory": {"min": "2Gi","max": "16Gi"},
			"storage": {"min": "64Gi","max": "512Gi"}
		}
	}]
}
```

Might lead to the following status:

```json
{
	"policies": [{
		"name": "di",
		"nodeSets": [{ "name": "nodeset-di-a", "nodeCount": 4}, {"name": "nodeset-di-b","nodeCount": 3}],
		"resources": {
			"requests": {"cpu": "6","memory": "10Gi","storage": "4Gi"}
		},
		"state": [{
			"type": "VerticalScalingLimitReached",
			"messages": [
				"Node required storage 283970435 is greater than max allowed: 209715200"
			]
		}],
		"lastModificationTime": "2021-01-06T09:55:57Z"
	}]
}
```

Reusing `policies` helps to have a consistency between the top-level structure of the 2 documents:

```
{
	"policies": [{
		"name": "di", ...
}
```

pkg/controller/autoscaling/elasticsearch/status/status.go
```diff
@@ -0,0 +1,212 @@
...
+	// ResourcesSpecification holds the resource values common to all the nodeSet managed by a same autoscaling policy.
+	// Only the resources managed by the autoscaling controller are saved in the Status.
+	ResourcesSpecification resources.NodeResources `json:"resources"`
+	// PolicyStates may contain various messages regarding the current state of this autoscaling policy.
+	PolicyStates []PolicyState `json:"state"`
+	// LastModificationTime is the last time the resources have been updated, used by the cooldown algorithm.
+	LastModificationTime metav1.Time `json:"lastModificationTime"`
+}
+
+func (s *Status) GetNamedTierResources(policyName string) (resources.NodeSetsResources, bool) {
```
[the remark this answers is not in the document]
barkbay: I shouldn't have used the term "named tier", I agree it is confusing. I'll fix that.

pkg/controller/autoscaling/elasticsearch/autoscaler/nodesets.go
```diff
@@ -0,0 +1,49 @@
...
+		nodeSetNodeCountList: nodeSetNodeCount,
+	}
+	fnm.sort()
+	return fnm
+}
+
+// AddNode selects the nodeSet with the highest priority and increases by one the value of its NodeCount field.
+// Priority is defined as the nodeSet with the lowest NodeCount value, or the first nodeSet in the alphabetical order if
+// several node sets have the same NodeCount value.
+func (fnm *FairNodesManager) AddNode() {
```
charith-elastic, line 44: Could we sort the list once and then treat it like a circular queue? I.e. maintain the current "head" of the array and move it to the next element (wrapping around if necessary) with each call to `AddNode`? 

We could also consider using https://golang.org/pkg/container/heap/ but that's probably overkill for this use case.

pkg/controller/elasticsearch/driver/autoscaling.go
```diff
@@ -0,0 +1,61 @@
...
+		if err != nil {
+			return false, err
+		}
+		if nodeSetAutoscalingSpec == nil {
+			// This nodeSet is not managed by an autoscaling configuration
+			log.V(1).Info("NodeSet not managed by an autoscaling controller", "nodeset", nodeSet.Name)
+			continue
+		}
+
+		s, ok := autoscalingStatus.CurrentResourcesForPolicy(nodeSetAutoscalingSpec.Name)
```
pebrc, line 40: `s` is confusing me. Should it be `r` for resources? Or even `expectedResources`

pkg/controller/elasticsearch/driver/autoscaling.go
```diff
@@ -0,0 +1,61 @@
...
+		if !ok {
+			log.Info("NodeSet managed by the autoscaling controller but not found in status",
+				"nodeset", nodeSet.Name,
+			)
+			return false, nil
+		}
+		inSync, err := s.IsUsedBy(esv1.ElasticsearchContainerName, nodeSet)
```
pebrc, line 47:
```suggestion
		inSync, err := expectedResources.Match(esv1.ElasticsearchContainerName, nodeSet)
```
?

pkg/controller/autoscaling/elasticsearch/driver.go
```diff
@@ -0,0 +1,252 @@
...
+			ctx := autoscaler.Context{
+				Log:                      log,
+				AutoscalingSpec:          autoscalingPolicy,
+				NodeSets:                 nodeSetList,
+				CurrentAutoscalingStatus: currentAutoscalingStatus,
+				RequiredCapacity:         capacity.RequiredCapacity,
+				StatusBuilder:            statusBuilder,
+			}
+			nodeSetsResources = ctx.GetResources()
+		}
```
charith-elastic, line 167: Nit: Could this be an `if/else`?

pkg/controller/autoscaling/elasticsearch/autoscaler/nodesets.go
```diff
@@ -0,0 +1,49 @@
...
+	// Peak the first element, this is the one with the less nodes
```
pebrc, line 45:
```suggestion
	// Peak the first element, this is the one with fewer nodes
```

charith-elastic (commented):
It's a lot to digest and I won't claim to have understood everything. But, overall LGTM 👍🏼

pkg/controller/autoscaling/elasticsearch/autoscaler/linear_scaler.go
```diff
@@ -0,0 +1,76 @@
...
+	if memRatio == 1 {
+		return cpuRange.Max.DeepCopy()
+	}
+
+	allowedCPURange := float64(cpuRange.Max.MilliValue() - cpuRange.Min.MilliValue())
+	requiredAdditionalCPUCapacity := int64(allowedCPURange * memRatio)
+	requiredCPUCapacityAsMilli := cpuRange.Min.MilliValue() + requiredAdditionalCPUCapacity
+
+	// Round up memory to the next core
+	requiredCPUCapacityAsMilli = roundUp(requiredCPUCapacityAsMilli, 1000)
```
pebrc, line 38: Should we allow CPU capacity in smaller increments than one core?

pkg/controller/autoscaling/elasticsearch/autoscaler/linear_scaler_test.go
```diff
@@ -0,0 +1,142 @@
...
+			},
+			wantMemory: qPtr("1Gi"), // keep the min. value
+		},
+		{
+			name: "Do not allocate more memory than max allowed",
+			args: args{
+				requiredStorageCapacity: q("2Gi"),
+				autoscalingSpec:         NewAutoscalingSpecBuilder("my-autoscaling-policy").WithMemory("1Gi", "1500Mi").WithStorage("1Gi", "2Gi").Build(),
+			},
+			wantMemory: qPtr("1500Mi"), // keep the min. value
```
pebrc, line 56: max value?

pkg/controller/autoscaling/elasticsearch/autoscaler/linear_scaler_test.go
```diff
@@ -0,0 +1,142 @@
...
+		},
+		{
+			name: "Do not allocate more memory than max allowed II",
+			args: args{
+				requiredStorageCapacity: q("1800Mi"),
+				autoscalingSpec:         NewAutoscalingSpecBuilder("my-autoscaling-policy").WithMemory("1Gi", "1500Mi").WithStorage("1Gi", "2Gi").Build(),
+			},
+			wantMemory: qPtr("1500Mi"), // keep the min. value
```
pebrc, line 64: max value?

pkg/controller/autoscaling/elasticsearch/autoscaler/linear_scaler_test.go
```diff
@@ -0,0 +1,142 @@
...
+		},
+		{
+			name: "Allocate max of memory when it's possible",
+			args: args{
+				requiredStorageCapacity: q("2Gi"),
+				autoscalingSpec:         NewAutoscalingSpecBuilder("my-autoscaling-policy").WithMemory("1Gi", "2256Mi").WithStorage("1Gi", "2Gi").Build(),
+			},
+			wantMemory: qPtr("2256Mi"), // keep the min. value
```
pebrc, line 72: max value?

pkg/controller/autoscaling/elasticsearch/autoscaler/offline.go
```diff
@@ -0,0 +1,129 @@
...
+		expectedNodeCount = autoscalingSpec.NodeCount.Max
+	}
+
+	// User may have added or removed some NodeSets while the autoscaling API is not available.
+	// We distribute the nodes to reflect that change.
+	fnm := NewFairNodesManager(log, nodeSetsResources.NodeSetNodeCount)
+	for expectedNodeCount > 0 {
+		fnm.AddNode()
+		expectedNodeCount--
+	}
```
pebrc, line 55: This abstraction strikes me as a bit odd. Could this not simply be a function?
```golang
distributeFairly(nodeSetsResources, expectedNodeCount)
```

Also: the logger seems unused.

pkg/controller/autoscaling/elasticsearch/autoscaler/offline.go
```diff
@@ -0,0 +1,129 @@
...
+		"policy", autoscalingSpec.Name,
+		"nodeset", nodeSetsResources.NodeSetNodeCount,
+		"count", nodeSetsResources.NodeSetNodeCount.TotalNodeCount(),
+		"resources", nodeSetsResources.ToInt64(),
+	)
+	return nodeSetsResources
+}
+
+// nodeSetResourcesFromStatus restores NodeSetResources from the status.
+// If user removed the limits while offline we are assuming that it wants to take back control on the resources.
```
pebrc, line 69:
```suggestion
// If user removed the limits while offline we are assuming that they want to take back control of the resources.
```

pkg/controller/autoscaling/elasticsearch/autoscaler/offline.go
```diff
@@ -0,0 +1,129 @@
...
+func nodeSetResourcesFromStatus(
+	currentAutoscalingStatus status.Status,
```
pebrc, line 71: Why are we passing the status here once more we already have extracted the `currentNodeSetsResouces` from the status which should contain everything we need?

pkg/controller/autoscaling/elasticsearch/autoscaler/offline.go
```diff
@@ -0,0 +1,129 @@
...
+	if autoscalingSpec.IsMemoryDefined() {
+		nodeSetsResources.SetRequest(corev1.ResourceMemory, autoscalingSpec.Memory.Min.DeepCopy())
+	}
+	if autoscalingSpec.IsStorageDefined() {
+		nodeSetsResources.SetRequest(corev1.ResourceStorage, autoscalingSpec.Storage.Min.DeepCopy())
+	}
+	return nodeSetsResources
+}
+
+// adjustQuantity ensures that a quantity is comprised between a min and a max.
```
pebrc, line 121:
```suggestion
// adjustQuantity ensures that the Quantity in value is between min and max.
```

pebrc (commented):
I think I am 2/3 through ...

pkg/controller/autoscaling/elasticsearch/controller.go
```diff
@@ -0,0 +1,228 @@
...
+		return defaultReconcile, nil
+	}
+
+	if common.IsUnmanaged(&es) {
+		log.Info("Object is currently not managed by this controller. Skipping reconciliation", "namespace", es.Namespace, "es_name", es.Name)
+		return reconcile.Result{}, nil
+	}
+
+	selector := map[string]string{label.ClusterNameLabelName: es.Name}
+	compat, err := annotation.ReconcileCompatibility(ctx, r.Client, &es, selector, r.OperatorInfo.BuildInfo.Version)
```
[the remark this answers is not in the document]
barkbay: Good point, since I think we still want to check the compatibility I introduced a dedicated function in https://github.com/elastic/cloud-on-k8s/pull/4197

pkg/controller/autoscaling/elasticsearch/resources/resources_test.go
```diff
@@ -0,0 +1,366 @@
...
+				Requests: tt.fields.Requests,
+			}
+			rs.MaxMerge(tt.args.other, tt.args.resourceName)
+			assert.True(t, apiequality.Semantic.DeepEqual(rs.Requests, tt.args.want.Requests), "Unexpected requests")
+			assert.True(t, apiequality.Semantic.DeepEqual(rs.Limits, tt.args.want.Limits), "Unexpected limits")
+		})
+	}
+}
+
+func TestNodeSetsResources_IsUsedBy(t *testing.T) {
```
pebrc, line 116: Test name out of sync

pkg/controller/autoscaling/elasticsearch/resources/resources.go
```diff
@@ -0,0 +1,283 @@
...
+}
+
+// NodeSetNodeCount models the number of nodes expected in a given NodeSet.
+type NodeSetNodeCount struct {
+	// NodeSet name.
+	Name string `json:"name"`
+	// NodeCount is the number of nodes, as computed by the autoscaler, expected in this NodeSet.
+	NodeCount int32 `json:"nodeCount"`
+}
+type NodeSetNodeCountList []NodeSetNodeCount
```
pebrc, line 118: Would it makes sense to have this as a map[string]int32 to begin with instead of having a custom struct and converting to maps on the fly?

pkg/controller/autoscaling/elasticsearch/status/events.go
```diff
@@ -0,0 +1,29 @@
...
+func EmitEvents(elasticsearch esv1.Elasticsearch, recorder record.EventRecorder, status Status) {
+	for _, status := range status.AutoscalingPolicyStatuses {
+		emitEventForAutoscalingPolicy(elasticsearch, recorder, status)
+	}
+}
+
+func emitEventForAutoscalingPolicy(elasticsearch esv1.Elasticsearch, recorder record.EventRecorder, status AutoscalingPolicyStatus) {
+	for _, event := range status.PolicyStates {
+		switch event.Type {
+		case VerticalScalingLimitReached, HorizontalScalingLimitReached, MemoryRequired, StorageRequired:
```
pebrc, line 25: Just curious why are we dropped certain event types here?

pebrc (commented):
Impressive work! I think I went through everything once. But I haven't tested as this is depending on the API PR IIUC

pkg/controller/autoscaling/elasticsearch/autoscaler/offline.go
barkbay: You're right, done in 3f462982f

pkg/controller/autoscaling/elasticsearch/controller.go
```diff
@@ -0,0 +1,228 @@
...
+
+type EsClientProvider func(ctx context.Context, c k8s.Client, dialer net.Dialer, es esv1.Elasticsearch) (esclient.Client, error)
+
+const (
+	controllerName = "elasticsearch-autoscaling"
+
+	enterpriseFeaturesDisabledMsg = "Autoscaling is an enterprise feature. Enterprise features are disabled"
+)
+
+var defaultReconcile = reconcile.Result{
```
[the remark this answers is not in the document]
barkbay: 👍  for `defaultRequeue`, this is also what we use in various [other](https://github.com/elastic/cloud-on-k8s/blob/fa7f098583f84888044813d257d5b0ff4500271e/pkg/controller/association/reconciler.go#L40) [places](https://github.com/elastic/cloud-on-k8s/blob/fa7f098583f84888044813d257d5b0ff4500271e/pkg/controller/elasticsearch/driver/driver.go#L46).

pkg/controller/autoscaling/elasticsearch/autoscaler/nodesets.go
barkbay: I think this is included in "Some parts of the code are not optimized" 😄 
I needed to have something that "just" do the work to make some progress on others, more important, parts. As long as it works as expected I think I'm happy to let someone else optimize it in an other PR.

pkg/controller/autoscaling/elasticsearch/autoscaler/linear_scaler.go
barkbay: I don't have a strong opinion. I always have preferred rounded values because it makes it easier to resonate about the performance model (e.g. how to size thread pools in the JVM with fractional number ?) But I agree that the user might want to have more accurate resource requests.
@elastic/cloud-k8s I'm curious if there are other opinions.

pkg/controller/autoscaling/elasticsearch/reconcile.go
```diff
@@ -0,0 +1,151 @@
...
+
+func newVolumeClaimTemplate(storageQuantity resource.Quantity, nodeSet esv1.NodeSet) ([]corev1.PersistentVolumeClaim, error) {
+	onlyOneVolumeClaimTemplate, volumeClaimTemplateName := validation.HasAtMostOnePersistentVolumeClaim(nodeSet)
+	if !onlyOneVolumeClaimTemplate {
+		return nil, fmt.Errorf(validation.UnexpectedVolumeClaimError)
+	}
+	if volumeClaimTemplateName == "" {
+		volumeClaimTemplateName = volume.ElasticsearchDataVolumeName
+	}
+	return []corev1.PersistentVolumeClaim{
```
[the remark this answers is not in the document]
barkbay: Good catch ! 🤦

pkg/controller/autoscaling/elasticsearch/resources/resources.go
barkbay: The aim of the `NodeSetNodeCount` struct is to make it easier to serialize the status with the pattern `{ name: "...", .... }`, for example:

```
	"nodeSets": [
        {
		"name": "nodeset-di-a",
		"nodeCount": 4
	},
        {
		"name": "nodeset-di-b",
		"nodeCount": 3
	}]
```

I wanted to be consistent with other items in the specs and explicitly have a `name` field in the status.

pkg/controller/autoscaling/elasticsearch/status/status.go
```diff
@@ -0,0 +1,212 @@
...
+		return psb
+	}
+	psb.states[stateType] = PolicyState{
+		Type:     stateType,
+		Messages: []string{message},
+	}
+	return psb
+}
+
+type PolicyStateType string
```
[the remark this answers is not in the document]
barkbay: This is something I didn't try to optimize yet. The API has been designed to be used this way, but I agree we should try to be clever here. Since this is an optimization I would track it with a dedicated issue and a separate PR ?

pkg/controller/autoscaling/elasticsearch/policy.go
```diff
@@ -0,0 +1,39 @@
...
+func updatePolicies(
+	ctx context.Context,
+	log logr.Logger,
+	autoscalingSpec esv1.AutoscalingSpec,
+	esclient client.AutoscalingClient,
+) error {
+	span, _ := apm.StartSpan(ctx, "update_autoscaling_policies", tracing.SpanTypeApp)
+	defer span.End()
+	// Cleanup existing autoscaling policies
+	if err := esclient.DeleteAutoscalingPolicies(ctx); err != nil {
```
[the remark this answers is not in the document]
barkbay: This is something I didn't try to optimize yet. The API has been designed to be used this way, but I agree we should try to be clever here. Since this is an optimization I would track it with a dedicated issue and a separate PR ?

pkg/controller/autoscaling/elasticsearch/autoscaler/horizontal.go
```diff
@@ -0,0 +1,115 @@
...
+	totalRequiredCapacity int64, // required capacity at the tier level
+	minNodes, maxNodes int, // min and max number of nodes in this tier, as specified by the user the autoscaling spec.
+	resourceName string, // used for logging and in events
+) int {
+	// minResourceQuantity is the resource quantity in the tier before scaling horizontally.
+	minResourceQuantity := int64(minNodes) * nodeResourceCapacity
+	// resourceDelta holds the resource needed to comply with what is requested by Elasticsearch.
+	resourceDelta := totalRequiredCapacity - minResourceQuantity
+	// getNodeDelta translates resourceDelta into a number of nodes.
+	nodeToAdd := getNodeDelta(resourceDelta, nodeResourceCapacity)
```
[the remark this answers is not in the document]
barkbay: This is actually the number of nodes that must be added to fulfill the requirements, I tried to improve the comment [here](https://github.com/elastic/cloud-on-k8s/pull/4173/files#diff-e0e0ede710eb10e3e4b3e10b8aa6f25802dc7a18a4784fa66b16410613de6b4dR21).

For example, if there is a minimum number of 2 (because this is what the user specified in the spec) but we need a total of 5 nodes in the cluster then `nodeToAdd` is 3

pkg/controller/autoscaling/elasticsearch/status/events.go
barkbay: I believe my initial thinking was that there may not be a 1-1 relationship between the events we want in the resource status _(primarily targeting the user)_ and the events we want to raise at the cluster level _(targeting the user and the K8S cluster admins)_.
For example I'm not sure that an empty response from Elasticsearch (`EmptyResponse`  event), which is likely to be a temporary situation, really deserves a K8S cluster event. While reaching a limit, on the contrary, is a critical situation which might require to be surfaced more broadly.
I agree it is not a strong argument, just a feeling, so I think I'm fine removing this filter for the moment.

pkg/controller/autoscaling/elasticsearch/autoscaler/autoscaler_test.go
```diff
@@ -0,0 +1,261 @@
...
+				policy: NewAutoscalingSpecBuilder("my-autoscaling-policy").WithNodeCounts(3, 6).WithMemory("5G", "8G").Build(),
+			},
+			want: resources.NodeSetsResources{
+				Name:             "my-autoscaling-policy",
+				NodeSetNodeCount: []resources.NodeSetNodeCount{{Name: "default", NodeCount: 3}},
+				NodeResources:    resources.NodeResources{Requests: map[corev1.ResourceName]resource.Quantity{corev1.ResourceMemory: q("7Gi")}},
+			},
+		},
+		{
+			name: "Scale both vertically and horizontally",
```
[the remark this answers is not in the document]
barkbay: Done in https://github.com/elastic/cloud-on-k8s/commit/e96a25e3eb591f5e4deee999c3f6b2dac89c70e3

pkg/controller/autoscaling/elasticsearch/status/events.go
pebrc: Makes sense, we can adjust this later as well.

pkg/controller/autoscaling/elasticsearch/autoscaler/linear_scaler.go
mapdegree: For Elasticsearch or Kibana, we would round up CPU capacity to an integer, but might be too much for lighter components, for example beats. 
Also, if you just want to run the whole ECK stack on a laptop, users would prefer to have less granual option to define resource limits, just to enable them to test locally.

barkbay:
Thanks all for the time you spent on this PR, I know it is not an easy one 🙇 

Quick update:
* I have been working on a [PR to support request/limit ratio](https://github.com/barkbay/cloud-on-k8s/compare/autoscaling-pr/controller...barkbay:tmp/autoscaling-pr/limits) but I think it requires this one to be merged before I raise it.
* Also this PR is blocked by this [one](https://github.com/elastic/cloud-on-k8s/pull/4197).

pkg/controller/autoscaling/elasticsearch/resources/resources.go
```diff
@@ -0,0 +1,283 @@
...
+		Name:             name,
+		NodeSetNodeCount: newNodeSetNodeCountList(nodeSetNames),
+	}
+}
+
+// ClusterResources models the desired resources (CPU, memory, storage and number of nodes) for all the autoscaling policies in a cluster.
+type ClusterResources []NodeSetsResources
+
+// Match returns true if the resources assigned to a container in a NodeSet matches the one specified in the NodeSetsResources.
+// It returns false if the container is not found in the NodeSet.
```
pebrc, line 40:
```suggestion
// It also returns false if the container is not found in the NodeSet.
```
?

pebrc (approved):
LGTM! I did only a very few tests over night and storage autoscaling worked as expected. But I hoped to be able to do some more tests. But I believe we can merge this as is and iterate if necessary.

PR #4174, opened 2021-02-01
Use semver package for version parsing

### LINKED ISSUE #4119: E2E getKubernetesVersion cannot deal with OCP3 patch versions
```
{"log.level":"info","@timestamp":"2021-01-14T10:56:21.396+0100","log.logger":"e2e-lkzoz","message":"Cleaning up the scratch directory","service.version":"0.0.0-SNAPSHOT+00000000","service.type":"eck","ecs.version":"1.4.0","directory":"/tmp/eck-e2e/e2e-lkzoz"}
panic: invalid patch format. version: 1.11.0+d4cacc0: strconv.Atoi: parsing "0+d4cacc0": invalid syntax

goroutine 1 [running]:
github.com/elastic/cloud-on-k8s/pkg/controller/common/version.MustParse(...)
	/Users/pebrc/src/cloud-on-k8s/pkg/controller/common/version/version.go:129
github.com/elastic/cloud-on-k8s/test/e2e/cmd/run.getKubernetesVersion(0xc0002e8000, 0x2, 0x2)
	/Users/pebrc/src/cloud-on-k8s/test/e2e/cmd/run/run.go:209 +0x3a7
github.com/elastic/cloud-on-k8s/test/e2e/cmd/run.(*helper).initTestContext(0xc0002e8000, 0x0, 0x0)
	/Users/pebrc/src/cloud-on-k8s/test/e2e/cmd/run/run.go:160 +0x1cd
github.com/elastic/cloud-on-k8s/test/e2e/cmd/run.doRun(0xc000351640, 0x2, 0x2, 0x7ffeefbff631, 0x30, 0x7ffeefbff6b7, 0x6, 0x0, 0x0, 0x7ffeefbff5ee, ...)
	/Users/pebrc/src/cloud-on-k8s/test/e2e/cmd/run/run.go:85 +0x25d
github.com/elastic/cloud-on-k8s/test/e2e/cmd/run.Command.func2(0xc0001c1340, 0xc0004dc700, 0x0, 0x10, 0x0, 0x0)
	/Users/pebrc/src/cloud-on-k8s/test/e2e/cmd/run/command.go:71 +0xc4
github.com/spf13/cobra.(*Command).execute(0xc0001c1340, 0xc0004dc600, 0x10, 0x10, 0xc0001c1340, 0xc0004dc600)
	/Users/pebrc/go/pkg/mod/github.com/spf13/cobra@v1.1.1/command.go:850 +0x453
github.com/spf13/cobra.(*Command).ExecuteC(0xc0001c1080, 0xc0004bff68, 0x2, 0x2)
	/Users/pebrc/go/pkg/mod/github.com/spf13/cobra@v1.1.1/command.go:958 +0x349
github.com/spf13/cobra.(*Command).Execute(...)
	/Users/pebrc/go/pkg/mod/github.com/spf13/cobra@v1.1.1/command.go:895
main.main()
	/Users/pebrc/src/cloud-on-k8s/test/e2e/cmd/main.go:31 +0x118
exit status 2
```

It looks like OCP3 has some slightly idiosyncratic patch version format. `1.11.0+d4cacc0` 

`semver.ParseTolerant()` should work instead or we add a tolerant mode to our versioning code.

charith-elastic:
We already import the semver package as a dependency, so there is no need to maintain the complicated version parsing logic ourselves.  

Also includes some cleanup work such as removing the duplication of logic between `LowestHighestSupportedVersion` and `MinMaxVersion`.

Fixes #4119

### DIFF
--- a/pkg/controller/common/version/version.go
+++ b/pkg/controller/common/version/version.go
@@ -6,14 +6,16 @@ package version
 
 import (
 	"fmt"
-	"sort"
-	"strconv"
-	"strings"
 
+	semver "github.com/blang/semver/v4"
 	"github.com/pkg/errors"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 )
+
+// Version is an alias for semver.Version.
+// It exists purely to avoid accidentally importing the old github.com/blang/semver instead of github.com/blang/semver/v4.
+type Version = semver.Version
 
 // GlobalMinStackVersion to additional restrict the allowed min version beyond the technical requirements expressed below.
 var GlobalMinStackVersion Version
@@ -30,6 +32,7 @@ var (
 )
 
 // MinMaxVersion holds the minimum and maximum supported versions.
+// Could be replaced with semver.Range if we didn't have to support GlobalMinStackVersion.
 type MinMaxVersion struct {
 	Min Version
 	Max Version
@@ -37,11 +40,11 @@ type MinMaxVersion struct {
 
 // WithinRange returns an error if the given version is not within the range of minimum and maximum versions.
 func (mmv MinMaxVersion) WithinRange(v Version) error {
-	if !v.IsSameOrAfter(mmv.Min) {
+	if v.LT(mmv.Min) {
 		return fmt.Errorf("version %s is lower than the lowest supported version of %s", v, mmv.Min)
 	}
 
-	if !mmv.Max.IsSameOrAfter(v) {
+	if v.GT(mmv.Max) {
 		return fmt.Errorf("version %s is higher than the highest supported version of %s", v, mmv.Max)
 	}
 
@@ -49,7 +52,7 @@ func (mmv MinMaxVersion) WithinRange(v Version) error {
 }
 
 func (mmv MinMaxVersion) WithMin(min Version) MinMaxVersion {
-	if min.IsAfter(mmv.Min) {
+	if min.GT(mmv.Min) {
 		return MinMaxVersion{
 			Min: min,
 			Max: mmv.Max,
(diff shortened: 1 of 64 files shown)


pkg/controller/common/version/version.go
```diff
@@ -30,26 +32,27 @@ var (
 )
 
 // MinMaxVersion holds the minimum and maximum supported versions.
+// Could be replaced with semver.Range if we didn't have to support GlobalMinStackVersion.
```
barkbay, line 35: I think it would also no longer be possible to return detailed errors from `WithinRange` ?
Otherwise would it help to declare `GlobalMinStackVersion` as  type `semver.Range`:

```golang
	if ubiOnly {
		container.SetContainerSuffix("-ubi8")
		version.GlobalMinStackVersion = semver.MustParseRange(">7.10.0")
	}
```

And use the `AND` receiver ?

```golang
// CheckSupportedStackVersion checks that the given version is a valid Stack version supported by ECK.
func CheckSupportedStackVersion(proposed version.Version, supported semver.Range) field.ErrorList {
	if (version.GlobalMinStackVersion.AND(supported))(proposed) {
		return field.ErrorList{field.Invalid(field.NewPath("spec").Child("version"), proposed, fmt.Sprintf("Unsupported version: %s", proposed))}
	}
	return nil
}
```

barkbay (commented):
I still want to do a few tests but LGTM overall, just left a few nits.

pkg/controller/common/version/version.go
charith-elastic: That is a clever idea. As you pointed out yourself, the downside is that the error message would be less detailed because `semver.Range` only gives an yes/no answer and no explanation. Therefore, I think I'd prefer to keep the existing implementation of `MinMaxVersion` for now because it is simple enough. If we need to support non-contiguous version ranges in the future, we can consider switching to `semver.Range`. WDYT?

barkbay: 👍 , SGTM

barkbay (approved):
LGTM

PR #4186, opened 2021-02-03
Fix Kibana NODE_OPTIONS env var

### LINKED ISSUE #4176: [DOCS] Document how to set Kibana heap space on ECK
## Request
Document how to modify Kibana's heap size in ECK per https://www.elastic.co/guide/en/kibana/current/production.html#memory

Users should set the env variable NODE_OPTIONS with the desired max-old-space-size value.

Example:

```
kind: Kibana
metadata:
  name: test-kibana
spec:
  version: 7.10.2
  count: 1
  podTemplate:
    spec:
      containers:
        - name: kibana
          env:
            - name: NODE_OPTIONS
              value: "--max-old-space-size=2048"
```

Ideally there should be such an example in the more examples section here: https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-customize-pods.html#k8s_more_examples

### LINKED ISSUE #4184: Kibana NODE_OPTS vs NOTE_OPTIONS
## Potential bug

The code here:
https://github.com/elastic/cloud-on-k8s/blob/master/pkg/license/testdata/stack.yaml#L108
sets NODE_OPTS with `--max-old-space-size`.

The parameter `--max-old-space-size` refers to node.js heap size per [https://www.elastic.co/guide/en/kibana/current/production.html#memory](https://www.elastic.co/guide/en/kibana/current/production.html#memory).
The relevant env variable for this should be NODE_OPTIONS (not NODE_OPTS) - ref. https://nodejs.org/api/cli.html#cli_environment_variables.

```
kind: Kibana
metadata:
  name: test-kibana-2
spec:
  version: 7.9.0
  count: 1
  podTemplate:
    spec:
      containers:
        - name: kibana
          env:
            - name: NODE_OPTS
              value: "--max-old-space-size=2048"
```

The code here 
https://github.com/elastic/cloud-on-k8s/blob/master/pkg/controller/kibana/config_settings.go#L34
also refers to increasing Kibana maximum memory limit. It is unclear if it is referring to something else (container memory?) or incorretly referring to the node.js NODE_OPTIONS env variable.

Clarify which case each of NODE_OPTIONS and NODE_OPTS should cover: if node.js heap memory, then it should be changed to NODE_OPTIONS and ideally documented per https://github.com/elastic/cloud-on-k8s/issues/4176

charith-elastic:
We have been using the wrong environment variable name to determine the heap size of Kibana.

Fixes #4184 
Fixes #4176

### DIFF
--- a/docs/orchestrating-elastic-stack-applications/kibana.asciidoc
+++ b/docs/orchestrating-elastic-stack-applications/kibana.asciidoc
@@ -133,7 +133,7 @@ The resource definitions in ECK share the same philosophy when you want to:
 === Pod configuration
 You can <<{p}-customize-pods,customize the Kibana Pod>> using a Pod template.
 
-The following example demonstrates how to create a Kibana deployment with custom node affinity and resource limits.
+The following example demonstrates how to create a Kibana deployment with custom node affinity, increased heap size, and resource limits.
 
 [source,yaml,subs="attributes"]
 ----
@@ -150,12 +150,15 @@ spec:
     spec:
       containers:
       - name: kibana
+        env:
+          - name: NODE_OPTIONS
+            value: "--max-old-space-size=2048"
         resources:
           requests:
             memory: 1Gi
             cpu: 0.5
           limits:
-            memory: 2Gi
+            memory: 2.5Gi
             cpu: 2
       nodeSelector:
         type: frontend
--- a/docs/orchestrating-elastic-stack-applications/managing-compute-resources.asciidoc
+++ b/docs/orchestrating-elastic-stack-applications/managing-compute-resources.asciidoc
@@ -66,8 +66,7 @@ spec:
 [id="{p}-compute-resources-kibana-and-apm"]
 === Set compute resources for Kibana and APM Server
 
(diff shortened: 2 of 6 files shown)


docs/orchestrating-elastic-stack-applications/kibana.asciidoc
```diff
@@ -150,12 +150,15 @@ spec:
...
+        env:
+          - name: NODE_OPTIONS
+            value: "--max-old-space-size=2048"
         resources:
           requests:
             memory: 1Gi
             cpu: 0.5
           limits:
-            memory: 2Gi
+            memory: 2.5Gi
```
pebrc, line 161: Not familiar with Node.js but how did we arrive at this 0.5Gi overhead?

pebrc (approved):
LGTM

docs/orchestrating-elastic-stack-applications/kibana.asciidoc
charith-elastic: It was an educated guess on my part too. My reasoning was that if Node.js is allowed to grow the heap to 2Gi, there must be some slack left for the app to run properly.

PR #4191, opened 2021-02-03
Adjust init container script for Elastic License 2.0

pebrc:
License version changes with 7.11 to Elastic License 2.0


barkbay (approved):
LGTM

PR #4197, opened 2021-02-05
Controller version refactoring

barkbay:
This pull request:

* Removes unnecessary calls to `accessor` methods (because of the switch to `client.Object` in https://github.com/elastic/cloud-on-k8s/pull/4098). The accessor is now only used to update the annotation if required (`accessor.SetAnnotations(obj, annotations)`)
* Introduces a function to check the compatibility with no side effect on the resource. It can then be used by auxiliary controllers to only check the compatibility with a given resource. See [this comment](https://github.com/elastic/cloud-on-k8s/pull/4173#discussion_r569422000) for more context.

### DIFF
--- a/pkg/controller/common/annotation/controller_version.go
+++ b/pkg/controller/common/annotation/controller_version.go
@@ -56,10 +56,7 @@ func UpdateControllerVersion(ctx context.Context, client k8s.Client, obj ctrlcli
 // can be used by auxiliary controllers to check if they can process a resource.
 // First boolean is true if the compatibility cannot be determined at the moment, mostly because the main controller did not reconcile the resource yet. In that case the
 // auxiliary controller must try again later.
-func CheckCompatibility(ctx context.Context, obj ctrlclient.Object, controllerVersion string) (requeue bool, supported bool, err error) {
-	span, ctx := apm.StartSpan(ctx, "check_compatibility", tracing.SpanTypeApp)
-	defer span.End()
-
+func CheckCompatibility(obj ctrlclient.Object, controllerVersion string) (requeue bool, supported bool, err error) {
 	annotatedVersion := obj.GetAnnotations()[ControllerVersionAnnotation]
 	if annotatedVersion == "" {
 		return true, false, nil
--- a/pkg/controller/common/annotation/controller_version.go
+++ b/pkg/controller/common/annotation/controller_version.go
@@ -54,15 +54,13 @@ func UpdateControllerVersion(ctx context.Context, client k8s.Client, obj ctrlcli
 
 // CheckCompatibility determines if this controller is compatible with a given resource by examining the controller version annotation. It has no side effect and
 // can be used by auxiliary controllers to check if they can process a resource.
-// First boolean is true if the compatibility cannot be determined at the moment, mostly because the main controller did not reconcile the resource yet. In that case the
-// auxiliary controller must try again later.
-func CheckCompatibility(obj ctrlclient.Object, controllerVersion string) (requeue bool, supported bool, err error) {
+// The auxiliary controller must watch the resource to check the compatibility if the resource is updated.
+func CheckCompatibility(obj ctrlclient.Object, controllerVersion string) (supported bool, err error) {
 	annotatedVersion := obj.GetAnnotations()[ControllerVersionAnnotation]
 	if annotatedVersion == "" {
-		return true, false, nil
+		return false, nil
 	}
-	supported, err = isAnnotatedVersionSupported(annotatedVersion, controllerVersion, obj)
-	return false, supported, err
+	return isAnnotatedVersionSupported(annotatedVersion, controllerVersion, obj)
 }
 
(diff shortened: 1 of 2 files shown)


pkg/controller/common/annotation/controller_version.go
```diff
@@ -59,15 +43,28 @@ func UpdateControllerVersion(ctx context.Context, client k8s.Client, obj ctrlcli
...
-	log.V(1).Info("updating controller version annotation", "namespace", namespace, "name", name, "kind", obj.GetObjectKind())
+	log.V(1).Info("updating controller version annotation", "namespace", obj.GetNamespace(), "name", obj.GetName(), "kind", obj.GetObjectKind())
 	return client.Update(context.Background(), obj)
 }
 
+// CheckCompatibility determines if this controller is compatible with a given resource by examining the controller version annotation. It has no side effect and
+// can be used by auxiliary controllers to check if they can process a resource.
+// First boolean is true if the compatibility cannot be determined at the moment, mostly because the main controller did not reconcile the resource yet. In that case the
+// auxiliary controller must try again later.
+func CheckCompatibility(obj ctrlclient.Object, controllerVersion string) (requeue bool, supported bool, err error) {
```
charith-elastic, line 59: How about returning a specific error (e.g. `ErrNotReconciled`) if the resource is not reconciled yet?

Returning two booleans from this method feels weird to me.  Since the function is called `CheckCompatibility`, I would expect it to return a true/false result about the compatibility. Whether to requeue or not is a decision that the caller can make for themselves by inspecting the error.

barkbay: I think it does not feel like an error to me, also returning such a specific error somewhat reminds me the "coding by exception" anti-pattern.
I would also expect this function to be called by controllers, they would all have to check for this error. Given the purpose of the function I'm not sure there is a situation where we would just ignore it and return it to the caller.

I agree that retuning multiple values is not common, I usually avoid that. But in that case I would say it seems "acceptable" ?

charith-elastic: I would say that in Go checking for specific kinds of errors is a common pattern (they are not "exceptions" that cause an expensive stack unwinding like in Java). To me this function seems to have multiple responsibilities. The controllers will still need to check whether `requeue == true` which is no more different than `errors.Is(err, ErrNotReconciled)`. The latter feels more idiomatic.

barkbay: I removed the first boolean as it is the responsibility for the caller to watch for changes on the resource anyway.
I still feel like using an error while there is none (it is an expected situation) seems wrong, this what I'm referring to with the "coding by exception" anti-pattern.
It meets the current need and can be revisited later if needed.

pkg/controller/common/annotation/controller_version.go
```diff
@@ -126,37 +116,30 @@ func ReconcileCompatibility(ctx context.Context, client k8s.Client, obj ctrlclie
 	}
 
 	// if the current version is gte the minimum version then they are compatible
-	if currentVersion.GTE(minVersion) {
```
pebrc, line 129: Not sure why we parse the controller version here and then only use it for logging. (Nothing you changed)  https://github.com/elastic/cloud-on-k8s/pull/4197/files#diff-b45319baa705e2377f984ec51682b15436fdf75a4db9c04e604c443c4b200546L123-L125

pebrc (approved):
LGTM, I think I am OK with the `(bool, error)` return.

pkg/controller/common/annotation/controller_version.go
barkbay: I _think_ it is to throw an early error. We have [some unit tests](https://github.com/elastic/cloud-on-k8s/blob/a0fe4502ac74242e66d92d5af1f90db952b32cc7/pkg/controller/association/reconciler_test.go#L275) which are expecting an error to be raised if the controller version can't be parsed.

barkbay: The unit test I mentioned is odd, its result does not depend on the value of the annotation as it should be:

```golang
func TestReconciler_Reconcile_NotCompatible(t *testing.T) {
	kb := sampleKibanaWithESRef()
	// set an incompatible controller annotation
	kb.Annotations = map[string]string{
		annotation.ControllerVersionAnnotation: "0.9.0",
	}
	r := testReconciler(&kb)
	_, err := r.Reconcile(context.Background(), reconcile.Request{NamespacedName: k8s.ExtractNamespacedName(&kb)})
	// should error out
	require.Error(t, err)
}
```

If the controller version in the annotation is not compatible we should not expect an error, the resource should be skipped.
When running this test the error we have is `No Major.Minor.Patch elements found`, it's because the controller version used in the test reconciler is wrong (`BuildInfo: about.BuildInfo{Version: "unit-tests"}`).
Setting a compatible version in the annotation still results in an error, it should not be the case.
I will raise a separate issue to investigate later.

PR #4198, opened 2021-02-05
Remove license check and custom err code in init container

pebrc:
Alternative to #4191 

This one has the advantage that it will be future proof thanks to the absence of the check. 
It has the disadvantage that users will have to deal with cryptic error messages of an unsupported Docker image not starting up or bootlooping, while previously we had a clear message that was also surfaced through a Kubernetes event.


pebrc:
We are going with #4191

PR #4203, opened 2021-02-05
Use Helm in E2E tests

### LINKED ISSUE #3804: Reuse helm RBAC assets for the E2E operator
We have RBAC rules for operator running in e2e tests defined in a dedicated place: https://github.com/elastic/cloud-on-k8s/tree/master/config/e2e.
And released operator RBAC rules defined in a different place: https://github.com/elastic/cloud-on-k8s/tree/master/deploy/eck-operator/templates.

Our E2E tests don't validate the RBAC rules (nor any other k8s resource) from the helm assets we release. Instead they validate what's specified in the dedicated e2e operator manifest.

https://github.com/elastic/cloud-on-k8s/pull/3803 is an example of breaking the tests because a rule from the helm assets wasn't duplicated in the e2e manifests. It could also happen the other way around, which is scarier: we could fix the e2e manifests but forget to duplicate its content to the helm manifests.

We should find a way to reuse the helm assets for the e2e tests.

charith-elastic:
Use the Helm chart (via manifest-gen) to deploy the operator for E2E tests.

Fixes #3804


pebrc (commented):
LGTM just one thing about the telemetry test

pebrc (approved):
LGTM!

PR #4207, opened 2021-02-05
Update VERSION to 1.2.2

pebrc:


config/eck.yaml
```diff
@@ -0,0 +1,14 @@
+log-verbosity: 0
```
charith-elastic, line 1: I think this file is ignored by `.gitignore` in other branches.

hack/manifest-gen/crd_patches/all-crds.yaml
```diff
@@ -0,0 +1,2673 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
charith-elastic, line 1: Git mishap?

hack/upgrade-test-harness/testdata/upcoming/install.yaml
```diff
@@ -0,0 +1,3045 @@
+---
```
charith-elastic, line 1: Git mishap?

charith-elastic (commented):
Looks like a couple of files snuck in due to outdated `.gitignore` in this branch

hack/upgrade-test-harness/testdata/upcoming/install.yaml
pebrc: Oh yeah I did a `add -A` going to fast. Will fix.

charith-elastic (approved):

PR #4218, opened 2021-02-08
[1.4] Update Stack resources list

charith-elastic:
Add Elastic Agent to the supported Stack applications list.
Backport of #4217


pebrc (approved):
Good catch I missed that one.

PR #4223, opened 2021-02-08
Add known issue description for stuck OLM based upgrade

pebrc:
Found while testing 1.4.0 upgrades with OLM. Affects also ECK upgrades from 1.2.x to 1.3.x


barkbay (approved):
LGTM

PR #4235, opened 2021-02-11
E2E: remove botuser from slackSend invocations

pebrc:
According to https://github.com/jenkinsci/slack-plugin#bot-user-mode botUser can only be used if a custom app is used. But we are using the default Jenkins CI app in Slack.


charith-elastic (approved):
LGTM.

According to the linked doc, the default app is considered legacy and not supported anymore. Should we involve infra and move to the new "blessed" integration method?

Issue #4006, opened 2020-12-03
Autoscaling specification


barkbay:
This issue is to discuss the content of the autoscaling specification _(at least in its first version)_.

### Global autoscaling settings (All optional - sane defaults would be provided)
The user should be able to:
- [x] Set the Elasticsearch autoscaling API poll frequency 
- [ ] ~Define a stabilization window: used to restrict the flapping of resources or replicas when the values from the Elasticsearch autoscaling API keep fluctuating _(see below)_~

### Tier/autoscaling policy specific settings

At the tier level, the user should be able to:
- [x] Create N-to-1 associations between nodeSets and an autoscaling policy (already discussed in https://github.com/elastic/cloud-on-k8s/issues/3955)
- [x] Tweak deciders settings (Optional)
- [x] Describe resource constraints (min. / max. values) applied on the associated `nodeSets`
- [x] Specify a `request`/`limit` ([Requests and limits](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#requests-and-limits)) ratio to convert a resource `request` into a `limit` _(Optional: default would be 1 for memory, limit == request, and 0 for CPU, not CPU limit)_
 
 ## Proposal

_Updated on January 5th, 2021 with @sebgl proposal_
 
 ```json
{
	"policies": [{
		"name": "di",
		"roles": ["data", "ingest"],
		"deciders": {
			"fixed": {
				"nodes": "4",
				"storage": "10g",
				"memory": "5g"
			}
		},
		"resources": {
			"nodeCount": {
				"min": 3,
				"max": 8
			},
			"cpu": {
				"min": 2,
				"max": 6,
				"requestsToLimitsRatio": 0
			},
			"memory": {
				"min": "6Gi",
				"max": "10Gi",
				"requestsToLimitsRatio": 1
			},
			"storage": {
				"min": "1Gi",
				"max": "4Gi"
			}
		}
	}],
	"syncPeriod": "60s"
}
```

`requestsToLimitsRatio` and `syncPeriod` are optional.

#### resources

`resources` allows the user to set a range of resources (min and max) which would be applied to each `nodeSet` targeted by the policy.

#### requestsToLimitsRatio

`requestsToLimitsRatio` is a float, either set to 0 or greater than 1, it would be used to convert a `request` value into a `limit`, for example:
* `1`: `limit` value is the same as the `request`
* `0`: no limit 
* `1.5`: `limit` is one and a half the value of the computed `request`
[...]

PR #4239, opened 2021-02-11
[Autoscaling] Add support for request/limit ratio

barkbay:
This PR allows to user to provide a ratio to convert Elasticsearch container `Requests` into `Limits`.
The optional field named `requestsToLimitsRatio` has the following default values:
* 1 for memory, which means that by default limit and request have the same value.
* 0 for CPU, which means that by default no limit is set on CPU

Here is a example where CPU request and limit have the same value, giving a QoS class of `Guaranteed` to the Pods in the `NodeSets` managed by the autoscaling policy: 

```json
{
	"policies": [{
		"name": "di-hot",
		"roles": ["data_hot", "ingest", "transform"],
		"resources": {
			"nodeCount": {
				"min": 2,
				"max": 5
			},
			"cpu": {
				"min": 1,
				"max": 3,
				"requestsToLimitsRatio": 1
			},
			"memory": {
				"min": "2Gi",
				"max": "6Gi"
			},
			"storage": {
				"min": "1Gi",
				"max": "3Gi"
			}
		}
	}]
}
```

Related to https://github.com/elastic/cloud-on-k8s/issues/4006

### DIFF
--- a/pkg/apis/elasticsearch/v1/autoscaling.go
+++ b/pkg/apis/elasticsearch/v1/autoscaling.go
@@ -26,7 +26,7 @@ var (
 	defaultMemoryRequestsToLimitsRatio = 1.0
 
 	// defaultCPURequestsToLimitsRatio is the default ratio used to convert a CPU request to a CPU limit in the Pod
-	// resources specification. By default we don't want CPU limit, hence a default value of 0.0
+	// resources specification. By default we don't want a CPU limit, hence a default value of 0.0
 	defaultCPURequestsToLimitsRatio = 0.0
 )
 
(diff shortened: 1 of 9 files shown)


pkg/apis/elasticsearch/v1/autoscaling.go
```diff
@@ -17,7 +17,18 @@ import (
...
 const ElasticsearchAutoscalingSpecAnnotationName = "elasticsearch.alpha.elastic.co/autoscaling-spec"
 
-var errNodeRolesNotSet = errors.New("node.roles must be set")
+var (
+	errNodeRolesNotSet = errors.New("node.roles must be set")
+
+	// defaultMemoryRequestsToLimitsRatio is the default ratio used to convert a memory request to a memory limit in the
+	// Pod resources specification. By default we want to have the same value for both the memory request and the memory
+	// limit.
+	defaultMemoryRequestsToLimitsRatio = 1.0
```
barkbay, line 26: I'm not 💯 sure about the plural here and in the spec.

barkbay:
Jenkins test this please

pebrc (approved):
LGTM. I think I am fine with the naming.

PR #4245, opened 2021-02-15
E2E: fix panic in GetCredentials and broken cluster handling for OCP

pebrc:
I had a broken OCP cluster that I could not delete anymore as it was returning certificate errors on `kubectl version` 
Also `make credentials` always failed with a panic for OCP clusters. 

This tries to address these two issues, by making `delete` work even with non functional clusters and making sure that we can alwasy call `make credentials`.  But there is room for further improvement.

### DIFF
--- a/hack/deployer/runner/ocp.go
+++ b/hack/deployer/runner/ocp.go
@@ -107,32 +107,15 @@ func (gdf *OcpDriverFactory) Create(plan Plan) (Driver, error) {
 			"OcpStateBucket":             OcpStateBucket,
 			"PullSecret":                 plan.Ocp.PullSecret,
 			"OverwriteDefaultKubeconfig": plan.Ocp.OverwriteDefaultKubeconfig,
+			"Authenticated":              false,
 		},
 	}, nil
 }
 
 func (d *OcpDriver) Execute() error {
-	if d.ctx["WorkDir"] == "" {
-		dir, err := ioutil.TempDir("", d.ctx["ClusterName"].(string))
-		if err != nil {
-			log.Fatal(err)
-		}
-
-		defer os.RemoveAll(dir)
-		d.ctx["WorkDir"] = dir
-	}
-
-	log.Printf("using WorkDir: %s", d.ctx["WorkDir"])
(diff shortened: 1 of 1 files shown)


hack/deployer/runner/ocp.go
```diff
@@ -186,7 +167,54 @@ func (d *OcpDriver) Execute() error {
 	return err
 }
 
-func (d *OcpDriver) clusterExists() (bool, error) {
+func (d *OcpDriver) ensureContext() (cleanUp func(), err error) {
+	cleanUp = func() {} // NOOP
+	if d.ctx["WorkDir"].(string) == "" {
```
charith-elastic, line 172: Is this guaranteed to be not nil? Otherwise the cast would panic.

hack/deployer/runner/ocp.go
```diff
@@ -186,7 +167,54 @@ func (d *OcpDriver) Execute() error {
...
+			os.RemoveAll(dir)
+		}
+		d.ctx["WorkDir"] = dir
+		log.Printf("Created WorkDir: %s", d.ctx["WorkDir"])
+	}
+
+	if _, exists := d.ctx["ClusterStateDir"]; !exists {
+		stateDir := filepath.Join(d.ctx["WorkDir"].(string), d.ctx["ClusterName"].(string))
+		if err := os.MkdirAll(stateDir, os.ModePerm); err != nil {
+			return cleanUp, err
```
charith-elastic, line 188: Is it possible that `cleanup` could be a noop because `d.ctx["WorkDir"]` was not empty? Is that the expected behaviour?

charith-elastic (commented):
It looks OK to me. 

I know it's not your code; but, the way it  plays fast and loose with types is a bit chilling for my personal taste. But let's not pull on that thread 😉

hack/deployer/runner/ocp.go
pebrc: Yes.

hack/deployer/runner/ocp.go
pebrc: No it's not.

charith-elastic (approved):

PR #4251, opened 2021-02-17
Upgrade to Go 1.16

charith-elastic:
- Update the Go version in `go.mod` so that transitive test dependencies don't show up in our dependencies list
- Simplify tool installation scripts with `go install` now that it supports versions


charith-elastic:
Jenkins test this please

hack/api-docs/build.sh
```diff
@@ -26,14 +26,13 @@ build_docs() {
...
-    local REFDOCS_VER="${REFDOCS_VER:-v0.0.6}"
+    local REFDOCS_VER="${REFDOCS_VER:-v0.0.7}"
     local BIN_DIR=${SCRATCH_DIR}/bin
 
     (
         echo "Installing crd-ref-docs $REFDOCS_VER to $BIN_DIR"
         mkdir -p "$BIN_DIR"
-        cd "$SCRATCH_DIR"
-        go mod init github.com/elastic/cloud-on-k8s-docs && GOBIN=$BIN_DIR go get -u "${REFDOCS_REPO}@${REFDOCS_VER}"
+        GOBIN=$BIN_DIR go install "${REFDOCS_REPO}@${REFDOCS_VER}"
```
barkbay, line 35: This requires Go 1.6, otherwise it fails with the following error:

```
> hack/api-docs/build.sh

Installing crd-ref-docs v0.0.7 to /var/folders/8c/2y60vh0108x9ry0flhyzvswm0000gn/T/crd-ref-docs-XXXXX.2ZiVYt2I/bin
package github.com/elastic/crd-ref-docs@v0.0.7: can only use path@version syntax with 'go get'
Removing /var/folders/8c/2y60vh0108x9ry0flhyzvswm0000gn/T/crd-ref-docs-XXXXX.2ZiVYt2I
```

I would also update https://github.com/elastic/cloud-on-k8s/blob/master/dev-setup.md to reflect this.

barkbay (approved):
LGTM

PR #4253, opened 2021-02-17
[E2E] Fix test agent configuration

barkbay:
It seems that some end of lines in the Agent configuration used by e2e tests have been truncated.


pebrc (approved):
Good catch!

Issue #4260, opened 2021-02-19
Elastic Agent writes data into its own container


pebrc:
We mount `/usr/lib/$K8S_NS/$NAME/agent-data` to `/usr/share/data` but Elastic Agent uses `/usr/share/elastic-agent/data`

https://github.com/elastic/cloud-on-k8s/blob/072c591f69f190d3a8c30789014a517d8a648208/pkg/controller/agent/pod.go#L33

This means that Elastic Agent is writing data into its own container. That includes all binaries it install as part of any configured packages (Metricbeat and Filebeat for example). This also means that it will lose its identity and runtime state on container restarts. 

The intention behind using the `hostPath` volume was to create a persistent store for Agent identity and runtime state. We are doing the same thing for Beats.

pebrc:
This turns out to be non-trivial to fix. HostPath volumes are mounted `noexec` but Elastic Agent does try to install and  execute the managed programs (typically Filebeat and Metricbeat) from its data directory.

pebrc:
The filesystem structure is: 
`/usr/share/ealstic-agent` => home **`--path.home`**
`/usr/share/elastic-agent/data` => data directory derived from home
`/usr/share/elastic-agent/data/elastic-agent-${hash}/run` => runtime state derived from home
`/usr/share/elastic-agent/data/elastic-agent-${hash}/install` => install location for the programs to execute **`agent.download.install_path`**
`/usr/share/elastic-agent/data/elastic-agent-${hash}/download` => download location for packages to install  **`agent.download.target_directory`**

The problem is that the install path by default is a subdirectory of the data directory which we want to have on the host via `hostPath`. But all we need to persist Agent identity across restarts is the runtime state in `run`. 

@david-kow suggested to override the latter two download settings to have Elastic Agent install into a directory inside the container in which we can execute and keep the data directory on the `hostPath` volume. Unfortunately Agent still creates a tmp folder below the data directory and then tries to rename/link from there to the install directory which is now on a different filesystem and fails subsequently.

```
[...]

PR #4261, opened 2021-02-19
Adjust config volume mount permissions for Beats to allow non-root user

### LINKED ISSUE #4108: Beats config mounted with 0600 permissions
Related to #4107 

We mount the `/etc/beat.yml` with 0600 permissions which seems wrong. It should be read-only and group readable so `0440` seems more appropriate. 

A quick test on non-secured cluster with a recent version of Beats seems to indicate that `0440` is fine but we need to verify with other versions and on OCP.

pebrc:
Align the permissions we use for the Beats configuration files with what we do in Agent. This theoretically allows non-root users (but root group members) to also read the configuration file. 

Related #4113 

Fixes #4108

### DIFF
--- a/pkg/controller/beat/common/pod.go
+++ b/pkg/controller/beat/common/pod.go
@@ -87,7 +87,7 @@ func buildPodTemplate(
 			ConfigVolumeName,
 			ConfigMountPath,
 			ConfigFileName,
-			0600),
+			0440),
 		dataVolume,
 	}
 


pebrc:
I don't think it makes sense to change the permissions as proposed here. In the light of https://github.com/elastic/cloud-on-k8s/issues/4260 it becomes clear that the non-root use of was only possible because we were not actually using the `hostPath` volume we are mounting into the Pod. I am therefore closing this for now.

PR #4262, opened 2021-02-19
[Helm] Unify role bindings

charith-elastic:
I discovered a bug that I could swear I fixed a couple of months ago (a missing role binding if the operator namespace is included in the managed namespaces list) 🤷🏼 

This PR fixes the bug and also combines the role bindings into a single template. Having them spread over two files made it too difficult to reason about.


deploy/eck-operator/templates/role-bindings.yaml
```diff
@@ -0,0 +1,80 @@
...
+rules:
+{{ template "eck-operator.rbacRules" $ | toYaml | indent 2 }}
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: "{{ $fullName }}"
+  namespace: {{ $namespace }}
+  labels:
+    {{- include "eck-operator.labels" $ | nindent 4 }}
```
pebrc, line 25: I think the Selector labels include is missing a `-` as it produces a leading empty line.

deploy/eck-operator/templates/role-bindings.yaml
```diff
@@ -0,0 +1,80 @@
+{{- $operatorNSIsManaged := has .Release.Namespace .Values.managedNamespaces -}}
```
pebrc, line 1: bit odd that the file is called `role-bindings.yaml` but it also contains roles

pebrc (approved):

PR #4264, opened 2021-02-22
[Autoscaling] Add setting to customize autoscaling API polling frequency

barkbay:
This PR allows the user to customize the default autoscaling API polling frequency.

### DIFF
--- a/pkg/apis/elasticsearch/v1/autoscaling.go
+++ b/pkg/apis/elasticsearch/v1/autoscaling.go
@@ -31,8 +31,8 @@ var (
 	// resources specification. By default we don't want a CPU limit, hence a default value of 0.0
 	defaultCPURequestsToLimitsRatio = 0.0
 
-	// defaultSyncPeriod is the default period between 2 Elasticsearch autoscaling API polls.
-	defaultSyncPeriod = 60 * time.Second
+	// defaultPollingPeriod is the default period between 2 Elasticsearch autoscaling API polls.
+	defaultPollingPeriod = 60 * time.Second
 )
 
 // -- Elasticsearch Autoscaling API structures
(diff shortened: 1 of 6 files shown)


pkg/apis/elasticsearch/v1/autoscaling.go
```diff
@@ -52,6 +57,10 @@ type AutoscalingPolicy struct {
 // +kubebuilder:object:generate=false
 type AutoscalingSpec struct {
 	AutoscalingPolicySpecs AutoscalingPolicySpecs `json:"policies"`
+
+	// SyncPeriod is the period at which to synchronize and poll the Elasticsearch autoscaling API.
+	SyncPeriod *metav1.Duration `json:"syncPeriod"`
```
pebrc, line 62: Should it simply be called `pollingFrequency` or `pollingPeriod`? If I read `sync` I wonder what it is being synchronized with what.

pkg/controller/autoscaling/elasticsearch/controller.go
```diff
@@ -42,7 +42,9 @@ const (
 	enterpriseFeaturesDisabledMsg = "Autoscaling is an enterprise feature. Enterprise features are disabled"
 )
 
-var defaultRequeue = reconcile.Result{
+// licenseCheckRequeue is the default duration used to retry a licence check if the cluster is supposed to be managed by
+// the autoscaling controller and if the licence is not valid.
+var licenseCheckRequeue = reconcile.Result{
```
pebrc, line 47: What is the motivation for using a different interval on invalid licenses? To avoid a bad experience for users that set `syncPeriod` to a really high value? Initially I thought it would be to protect the operator from users that set it to a really low value because the license check is relatively expensive. But I realise that the check runs on every reconciliation loop anyway. Maybe worth a comment.

pebrc (commented):
LGTM! Maybe adjust naming.

pkg/controller/autoscaling/elasticsearch/controller.go
barkbay: > To avoid a bad experience for users that set syncPeriod to a really high value?

Yes exactly, my initial thought was that if the license is invalid it might be a user mistake, and we may want to trigger a new loop sooner than later, regardless of the `syncPeriod` value which serves an other purpose.

pebrc (approved):

PR #4265, opened 2021-02-22
[Autoscaling] Report overlapping policies

barkbay:
This PR reports overlapping policies in the autoscaling status and in the logs.

### DIFF
--- a/pkg/controller/autoscaling/elasticsearch/driver.go
+++ b/pkg/controller/autoscaling/elasticsearch/driver.go
@@ -7,6 +7,8 @@ package elasticsearch
 import (
 	"context"
 	"fmt"
+	"sort"
+	"strings"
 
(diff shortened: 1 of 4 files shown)


pkg/controller/autoscaling/elasticsearch/driver.go
```diff
@@ -42,22 +45,51 @@ func (r *ReconcileElasticsearch) reconcileInternal(
...
+	for _, role := range roles {
+		policies := policiesByRole[role]
+		if len(policies) < 2 {
+			// This role is declared in only one autoscaling policy
+			continue
+		}
+		// sort policies for consistent comparison
+		sort.Strings(policies)
+		for _, policy := range policies {
+			message := fmt.Sprintf("role %s is declared in autoscaling policies %s", role, strings.Join(policies, ","))
```
pebrc, line 85: Should this not be outside the loop?

pebrc (approved):
LGTM just one nit

PR #4266, opened 2021-02-22
Add autoscaling usage reporting

barkbay:
This PR adds autoscaling usage reporting.

### DIFF
--- a/pkg/telemetry/telemetry.go
+++ b/pkg/telemetry/telemetry.go
@@ -28,8 +28,9 @@ import (
 )
 
 const (
-	resourceCount = "resource_count"
-	podCount      = "pod_count"
+	autoscaledResourceCount = "autoscaled_resource_count"
+	resourceCount           = "resource_count"
+	podCount                = "pod_count"
 
 	timestampFieldName = "timestamp"
 )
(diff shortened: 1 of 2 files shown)


idanmo:
@barkbay Perhaps we should add to telemetry what node roles the autoscaling policies are set for?

david-kow (approved):
LGTM

barkbay:
> @barkbay Perhaps we should add to telemetry what node roles the autoscaling policies are set for?

Do you have any specific idea in mind about how the roles could be reported in the document ?
I'm not sure if something like that could do the job:

```yaml
   elasticsearches:
       autoscaling:
          resource_count: 2 # 2 clusters have autoscaling enabled
          roles:
            data_hot: 2 # Declared in 2 autoscaling policies
            ml: 1 # Declared in 1 autoscaling policy
            master: 1 # Even if there is no decider for master nodes the role could be included in autoscaling policies roles
      pod_count: 9
      resource_count: 2
```

idanmo:
> Do you have any specific idea in mind about how the roles could be reported in the document ?
After discussing with @barkbay , we don't have a strong feeling about adding these metrics so I'm +1 with the current implementation.

Issue #4267, opened 2021-02-23
Kind E2E test failure due to monitoring operator


charith-elastic:
After #4248 the Kind tests are failing with the following error:

```
error: error validating "/tmp/eck-e2e/e2e-bxqdo/monitoring-operator.yaml": error validating data: ValidationError(Namespace): unknown field "go" in io.k8s.api.core.v1.Namespace; if you choose to ignore these errors, turn validation off with --validate=false
```

This doesn't happen when the Kind E2E tests are run locally so it must be some environmental factor that's causing this.

PR #4268, opened 2021-02-23
Skip creating namespaces during E2E run

charith-elastic:
Skip creating the namespaces using Helm because the E2E test runner already creates the namepaces beforehand.

Potential fix for #4267


pebrc (approved):
LGTM, took a while to wrap my head around the difference between `--create-namespace` in Helm, `manifestGen` and `manifestGen` with `createOperatorNamespace=false` Not sure how or if we can document this, for people like me.

charith-elastic:
I have added some comments. Hopefully that helps.

PR #4274, opened 2021-02-24
Use a helper pod to figure out the mount path for Elastic Agent

pebrc:
As I have exhausted all other ideas for https://github.com/elastic/cloud-on-k8s/issues/4260, this PR illustrates a workaround that uses a helper pod to figure out the correct mount path for Elastic Agents runtime data directory, which we want to have in the a `hostPath` volume.

Drawbacks of this approach are many:
* very brittle
* might break our Elastic Agent orchestration as soon as the Agent bug gets fixed maybe even ealier

### DIFF
--- a/config/recipes/elastic-agent/kubernetes-integration.yaml
+++ b/config/recipes/elastic-agent/kubernetes-integration.yaml
@@ -13,6 +13,8 @@ spec:
         serviceAccountName: elastic-agent
         containers:
         - name: agent
+          securityContext:
+            runAsUser: 0
           env:
           - name: NODE_NAME
             valueFrom:
--- a/config/recipes/elastic-agent/system-integration.yaml
+++ b/config/recipes/elastic-agent/system-integration.yaml
@@ -11,6 +11,8 @@ spec:
       spec:
         containers:
         - name: agent
+          securityContext:
+            runAsUser: 0
   config:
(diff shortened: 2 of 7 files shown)


david-kow:
Two questions:
1. Do I read correctly that this change forbids running two different images of EA with the same Operator (eg. official and custom)?
2. How do we back out of this change if the issue is fixed in future EA version? Few ifs here and there to get to the previous behavior for new versions?

pebrc:
> Do I read correctly that this change forbids running two different images of EA with the same Operator (eg. official and custom)?

Well it depends, our requirement is that all images used are derived from the official images so they would have the same filesystem structure. But in principle you are right we are encoding assumptions about the filesystem of the managed containers into the operator here and that could turn out to be a limitation.

>    How do we back out of this change if the issue is fixed in future EA version? Few ifs here and there to get to the previous behavior for new versions?

No good answer here as well. A "few ifs here and there" it is.

david-kow:
> > Do I read correctly that this change forbids running two different images of EA with the same Operator (eg. official and custom)?
> 
> Well it depends, our requirement is that all images used are derived from the official images so they would have the same filesystem structure. But in principle you are right we are encoding assumptions about the filesystem of the managed containers into the operator here and that could turn out to be a limitation.

I misunderstood what the hash is (it is sha of the latest commit, which won't change even if there is a custom container), so my concern was invalid.

PR #4275, opened 2021-02-25
Exclude Go module download output from manifest-gen output

### LINKED ISSUE #4267: Kind E2E test failure due to monitoring operator

charith-elastic:
If the Go module cache is unpopulated, running the manifest-gen script results in a module download operation. When the E2E runner tries to capture the output of manifest-gen, it also captures the output from the Go tool and the resulting manifest file becomes invalid. 

Fixes #4267

### DIFF
--- a/hack/manifest-gen/manifest-gen.sh
+++ b/hack/manifest-gen/manifest-gen.sh
@@ -67,7 +67,9 @@ while getopts "ug" OPT; do
             ARGS=("$@")
             (
                 cd "$SCRIPT_DIR"
-                go run main.go --source="$CHART_DIR" generate "${ARGS[@]}"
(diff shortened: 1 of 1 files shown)


pebrc (approved):
LGTM! 🎉 🚀 👏

PR #4279, opened 2021-02-27
Update eck-elasticsearch ingress namespace selector

Ares3266:
The eck-elasticsearch ingress ns selector is attempting to use a label that isn't actually present on the 1.4.0 release of the `elastic-system` namespace. Propose instead to use the same label selector that you do for the pod; `control-plane: elastic-operator`


Ares3266:
Helps if I read the docs properly 🤦🏼‍♂️

PR #4271, opened 2021-02-24
Set webhook matchPolicy to Exact

charith-elastic:
`admissionregistration.k8s.io/v1` changed the default `matchPolicy` to `Equivalent` from `Exact`. This results in both `v1` and `v1beta1` webhooks firing for resources like Elasticsearch, Kibana and APMServer. 

Fixes #4270

PR #4283, opened 2021-03-01
[1.4.1] Set webhook matchPolicy to Exact (#4271)

charith-elastic:
Backport of #4271


go.sum
```diff
@@ -1,7 +1,5 @@
-cloud.google.com/go v0.26.0 h1:e0WKqKTd5BnrG8aKH3J3h+QvEIQtSUcf2n5UZ5ZgLtQ=
```
thbkrkr, line 1: Why is this file modified?

charith-elastic: Go 1.16 gets rid of those entries when you run `go mod tidy`. I don't know the reason why. The changes seem to be harmless given that the E2E tests ran successfully so I kept them on.

thbkrkr (approved):
LGTM

PR #4295, opened 2021-03-03
Increase timeout and parallelism in PR build job

### LINKED ISSUE #4294: 45min timeout reached on PR job
https://devops-ci.elastic.co/view/cloud-on-k8s/job/cloud-on-k8s-pr/6049/

This is the third time in the past 13 days.

thbkrkr:
* Increase global timeout of the job from 45min to 60min
* Run the 2 first stages `Validate Jenkins pipelines` and `Run checks` in parallel with failfast enabled

Resolves #4294

### DIFF
--- a/.ci/pipelines/pr-gke.Jenkinsfile
+++ b/.ci/pipelines/pr-gke.Jenkinsfile
@@ -20,70 +20,65 @@ pipeline {
     }
 
     stages {
-        stage('Run checks and tests in parallel') {
+        stage('Run checks in parallel') {
             failFast true
             parallel {
-                stage('Run checks in parallel') {
-                    failFast true
-                    parallel {
-                        stage('Validate Jenkins pipelines') {
-                            when {
-                                expression {
-                                    notOnlyDocs()
-                                }
-                            }
-                            steps {
-                                sh 'make -C .ci TARGET=validate-jenkins-pipelines ci'
-                            }
+                stage('Validate Jenkins pipelines') {
+                    when {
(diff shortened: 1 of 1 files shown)


thbkrkr:
Argh: `WorkflowScript: 28: "matrix" or "parallel" cannot be nested inside another "matrix" or "parallel". @ line 28, column 21.`

sebgl (approved):
LGTM

.ci/pipelines/pr-gke.Jenkinsfile
```diff
@@ -20,25 +20,30 @@ pipeline {
...
 
     stages {
-        stage('Validate Jenkins pipelines') {
-            when {
-                expression {
-                    notOnlyDocs()
+        stage('Run checks in parallel') {
+            failFast true
+            parallel {
+                stage('Validate Jenkins pipelines') {
```
pebrc, line 26: What is the benefit of running these validations every time? To get better feedback on potential errors? I wonder if we should make this conditional on a flag or something and not incur the cost of always running these.

thbkrkr: It's a best-effort validation to prevent submitting pipelines with invalid syntax. I agree that it seems a little overkill to run it every time. I think we've done it for the sake of simplicity, even it shouldn't be very hard to make this validation conditional by reusing what we do with `notOnlyDocs()`.

pebrc (approved):

PR #4312, opened 2021-03-05
Add data tier labels to Elasticsearch nodes

### LINKED ISSUE #4054: Elasticsearch pod labels ignore data tiers
Elasticsearch 7.10.0 introduces with the data tiers concept four additional node roles: 
```
data_content
data_hot
data_warm
data_cold
```

We are currently not adding labels for those new roles to the Elasticsearch pods.

charith-elastic:
Adds node labels to indicate the data tiers associated with each node.

Also refactors the node role checking function to reduce code duplication.

Fixes #4054


thbkrkr (approved):
LGTM

As we increase the number of labels, I wonder how it makes sense to keep the labels where the value is `false`.

PR #4317, opened 2021-03-08
E2E: Extract Support Diagnostics from Elasticsearch clusters under test

### LINKED ISSUE #3766: Grab data from the Elasticsearch API to better troubleshoot E2E tests
Troubleshooting https://github.com/elastic/cloud-on-k8s/issues/3764 is pretty hard, since we are lacking details about:
- the entire cluster state (truncated in the Elasticsearch logs)
- the output from _cluster/allocation/explain

It would probably be also useful to grab:
- _cat/shards
- _cluster/settings

We could perform these calls as part of our existing `eck-dump.sh` tool, which already runs on failed E2E tests.
It would also make sense to rely on the [official Elasticsearch support diagnostics tool](https://github.com/elastic/support-diagnostics), but it probably involves more plumbing.

These calls also make sense beyond the context of E2E tests, for any troubleshooting purposes.

pebrc:
This adds another post-failure action to our e2e test setup to run the official support diagnostics tools against all currently deployed Elasticsearch clusters. 

It is using a "naked" k8s Pod that runs the diagnostics in an init container and simply sleeps in the main container. This allows us to know when the diagnostics are ready (== when the Pod is ready). We then extract the results with `kubectl cp` and delete the Pod.

Fixes #3766


pebrc:
Jenkins test this please

config/e2e/diagnostics_job.yaml
```diff
@@ -0,0 +1,39 @@
...
+metadata:
+  name: {{ .PodName }}
+  namespace: {{ .ESNamespace }}
+  labels:
+    test-run: {{ .Context.TestRun }}
+spec:
+  terminationGracePeriodSeconds: 0
+  initContainers:
+    - name: diagnostics
+      image: docker.elastic.co/eck-dev/support-diagnostics:412ac7a392cf
```
barkbay, line 12: Should we add a note to explain how this image is created/managed ?

test/e2e/cmd/run/run.go
```diff
@@ -783,3 +783,117 @@ func (h *helper) dumpK8sData() {
...
+		if err != nil {
+			log.Info(string(out))
+			return err
+		}
+	}
+	return nil
+}
+
+func (h *helper) normalizeDiagnosticArchives(dirName string) error {
+	// CI piplines are configured to upload *.tgz
```
barkbay, line 885: nit: Maybe follow the godoc's approach and move this comment before the function ?
Also I'm wondering if replacing the current pattern with `*.{tgz,zip}` in the pipelines would also do the job.

barkbay (commented):
Just left a few nits, I'm still doing some tests.

test/e2e/cmd/run/run.go
pebrc: > Also I'm wondering if replacing the current pattern with *.{tgz,zip,gz} in the pipelines would also do the job.

It probably would but this normalisation also helps with name clashes which can happen if more than one cluster exist that we want to extract diagnostics from. The support-diagnostics tool just uses a timestamp with second resolution and I had in my test a couple of times the case where the names for the archive of two different clusters where identical because the the tool ran at the same time for each of them.

test/e2e/cmd/run/diagnostics.go
```diff
@@ -0,0 +1,130 @@
...
+
+func (h *helper) runEsDiagnosticsJob() {
+	output, err := h.kubectl("get", "es", "--all-namespaces", "-o", "json")
+	if err != nil {
+		log.Error(err, "failed to list Elasticsearch clusters")
+		return
+	}
+	var ess v1.ElasticsearchList
+	if err := json.Unmarshal([]byte(output), &ess); err != nil {
+		log.Error(err, "failed to unmarshal kubectl response")
```
barkbay, line 35:
```suggestion
		log.Error(err, "failed to unmarshal kubectl response")
		return
```

test/e2e/cmd/run/diagnostics.go
```diff
@@ -0,0 +1,130 @@
...
+	var wg sync.WaitGroup
+	wg.Add(len(ess.Items))
+
+	diagnosticsTimeout := 10 * time.Minute
+
+	for _, es := range ess.Items {
+		go func(es v1.Elasticsearch) {
+			defer wg.Done()
+			podName := fmt.Sprintf("diag-%s", es.Name)
+			err := h.kubectlApplyTemplateWithCleanup("config/e2e/diagnostics_job.yaml", diagnosticContext{
```
barkbay, line 52: nit: there are a few places where the returned values can be scoped to the `if` block that handles them:

```golang
			if err := h.kubectlApplyTemplateWithCleanup("config/e2e/diagnostics_job.yaml", diagnosticContext{
				Context:     h.testContext,
				ESNamespace: es.Namespace,
				ESName:      es.Name,
				PodName:     podName,
				TLS:         es.Spec.HTTP.TLS.Enabled(),
			}); err != nil {
				log.Error(err, "diagnostics pod failed to create")
				return
			}
```

test/e2e/cmd/run/diagnostics.go
```diff
@@ -0,0 +1,130 @@
...
+				return
+			}
+
+			wait := exec.Command("kubectl", "wait", //nolint:gosec
+				fmt.Sprintf("--timeout=%s", diagnosticsTimeout.String()),
+				"--for=condition=ContainersReady",
+				"-n", es.Namespace,
+				fmt.Sprintf("pod/%s", podName),
+			)
+			out, err := wait.CombinedOutput()
```
barkbay, line 70: Same here:

```suggestion
			if out, err := wait.CombinedOutput(); err != nil {
				log.Error(err, "diagnostics pod did not complete successfully", "pod", podName, "output", string(out))
				return
			}
```

test/e2e/cmd/run/diagnostics.go
```diff
@@ -0,0 +1,130 @@
...
+		out, err := fn(file)
+		if err != nil {
+			log.Info(string(out))
+			return err
+		}
+	}
+	return nil
+}
+
+// normalizeDiagnosticArchives  normalizes everything to *.tgz. This is because CI piplines are configured to upload *.tgz,
```
barkbay, line 116:
```suggestion
// normalizeDiagnosticArchives normalizes everything to *.tgz. This is because CI piplines are configured to upload *.tgz,
```

config/e2e/diagnostics_job.yaml
```diff
@@ -0,0 +1,40 @@
...
+            key: elastic
+      command: [sh, -c]
+      args:
+      - ./diagnostics.sh -h {{.ESName}}-es-http --type api --bypassDiagVerify -u elastic --passwordText $ES_PW {{if .TLS}}-s --noVerify{{end}} -o /diagnostic-output
+      volumeMounts:
+      - name: output
+        mountPath: /diagnostic-output
+  containers:
+  - name: offer-output
+    image: alpine
```
barkbay, line 29: Would it be possible to rely on the same Docker image as the one used for the `diagnostics` container ?
The purpose would be to avoid to add an additional dependency on an external infra. Also I don't know how this image is maintained.

barkbay (approved):
I only did a few tests outside of Jenkins but it seems to be working as expected 👍

config/e2e/diagnostics_job.yaml
pebrc: Yes definitely. I changed it.

test/e2e/cmd/run/diagnostics.go
pebrc: Sometimes I find it it easier to parse visually if it is a separate statement, especially if the call producing the value is already spreading over multiple lines like here. But I changed it and the others too.

PR #4323, opened 2021-03-09
Add Webhook section to Advanced topics

### LINKED ISSUE #4028: Document webhook namespace selectors
Follow up from #3876 

The use of namespace selectors should be documented. The main use case is the ability to partition a large cluster by running several operators where each operator is responsible for a set of namespaces. This is a specialised use case which does not have a good place within the current documentation structure.

alaudazzi:
This PR proposes an initial draft to document Webhook namespace selectors.

The local doc build fails. Fixing it while the content gets finalized. 

Fixes #4028


docs/advanced-topics/advanced-topics.asciidoc
```diff
@@ -14,10 +14,12 @@ endif::[]
 - <<{p}-service-meshes>>
 - <<{p}-traffic-splitting>>
 - <<{p}-network-policies>>
+- <<{p}-webhook>>
```
charith-elastic, line 17: We already have this anchor defined elsewhere: https://github.com/elastic/cloud-on-k8s/blob/master/docs/operating-eck/webhook.asciidoc

Maybe rename to something like `{p}-webhook-namespace-selectors`?

docs/advanced-topics/webhook.asciidoc
```diff
@@ -0,0 +1,19 @@
...
+ifdef::env-github[]
+****
+link:https://www.elastic.co/guide/en/cloud-on-k8s/master/k8s-{page_id}.html[View this document on the Elastic website]
+****
+endif::[]
+
+[id="{p}-{page_id}"]
+= Webhook namespace selectors 
+
+Users are now allowed to set `namespaceSelector` and `objectSelector` on the webhook. The webhook name is generated as ``<operator-name>.<operator-namespace>.k8s.elastic.co` so that multiple operators can be installed side-by-side in the same cluster.
```
charith-elastic, line 11: Users can only do this when installing the operator via the Helm chart. We should make that clear.

docs/advanced-topics/webhook.asciidoc
```diff
@@ -0,0 +1,19 @@
...
+endif::[]
+
+[id="{p}-{page_id}"]
+= Webhook namespace selectors 
+
+Users are now allowed to set `namespaceSelector` and `objectSelector` on the webhook. The webhook name is generated as ``<operator-name>.<operator-namespace>.k8s.elastic.co` so that multiple operators can be installed side-by-side in the same cluster. 
+
+This can be useful in large and busy clusters, where you might want to divide the set of namespaces across several operators to speed up reconciliation times and reduce the amount of resources required per operator.
+
+Webhook resources are cluster-scoped, therefore `createClusterScopedResources` must be true. This means that each operator gets a ClusterRole that is not restricted to the set of managed namespaces. This approach avoids over-complicating the chart and is suitable for load-splitting scenarios.
```
charith-elastic, line 15:
```suggestion
Webhook resources are cluster-scoped, therefore `createClusterScopedResources` must be set to true when installing the chart. This means that each operator gets a ClusterRole that applies to the whole cluster and not just the set of namespaces it is configured to manage. This approach is suitable only if you want to do load-splitting in a trusted cluster. 
```

docs/advanced-topics/webhook.asciidoc
```diff
@@ -0,0 +1,19 @@
...
+
+Users are now allowed to set `namespaceSelector` and `objectSelector` on the webhook. The webhook name is generated as ``<operator-name>.<operator-namespace>.k8s.elastic.co` so that multiple operators can be installed side-by-side in the same cluster. 
+
+This can be useful in large and busy clusters, where you might want to divide the set of namespaces across several operators to speed up reconciliation times and reduce the amount of resources required per operator.
+
+Webhook resources are cluster-scoped, therefore `createClusterScopedResources` must be true. This means that each operator gets a ClusterRole that is not restricted to the set of managed namespaces. This approach avoids over-complicating the chart and is suitable for load-splitting scenarios. 
+
+WARNING: It is not recommended to deploy webhook resources in environments where operators are run by untrusted users and need to be locked down tightly.
+
+For more information on webhook, see link:https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/[Dynamic Admission Control].
```
charith-elastic, line 19: Maybe point to https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-webhook.html as well?

alaudazzi:
Thank you @charith-elastic!
I renamed `webhook.asciidoc` into `webhook-namespace-selectors.asciidoc` and integrated your comments. 
The local doc builds correctly.
How can I remove the old `webhook.asciidoc` from this PR?

PR #4330, opened 2021-03-10
[Autoscaling] Fix wrong ML node count in settings

barkbay:
This PR fixes the ML node count calculation used by the autoscaling controller to set the `max_lazy_ml_nodes` setting


thbkrkr (approved):
LGTM

Issue #4314, opened 2021-03-08
ARM e2e run sometimes ignores build constraints


pebrc:
We use build constraints to only run a subset of e2e tests for Elasticsearch. But sometimes it seems the build constraints are ignored and tests are run that do no have the `es` build tag. 

Example run: 
https://devops-ci.elastic.co/job/cloud-on-k8s-e2e-tests-eks-arm/38/testReport/

charith-elastic:
```
10:35:47  docker buildx build \
10:35:47  	--progress=plain \
10:35:47  	--file test/e2e/Dockerfile \
10:35:47  	--build-arg E2E_JSON=true \
10:35:47  	--build-arg GO_TAGS=es \
10:35:47  	--platform linux/amd64,linux/arm64 \
10:35:47  	--push \
10:35:47  	-t docker.elastic.co/eck-ci/eck-e2e-tests:a1026cfa .
```

```shell
docker run -i -t --entrypoint=/bin/bash docker.elastic.co/eck-ci/eck-e2e-tests:a1026cfa 
101@bc853846abc8:~$ printenv
...
GO_TAGS=e2e
...
E2E_JSON=true
```

🤔 :rage3:

pebrc:
Is it not the case that we build the same e2e test image for multiple jobs during the nightly run with the same tag e.g. last night: 
`docker.elastic.co/eck-ci/eck-e2e-tests:7c7bd9b8` and but they are not the same content. So we have effectively created a race condition: 
* ARM job pushes image
* non-ARM job pushes same tag
* ARM job starts pulling image
* ARMT job runs "wrong" image

charith-elastic:
I think you hit the nail right on the head there Peter. We build the operator image once and share it with all the jobs but not the E2E image. I wonder why that is. Before the build tags were introduced, the image would have been the same for every job and we could have saved quite a lot of time on test runs by building it just once.

charith-elastic:
So, until the build process can be overhauled to build more targeted artifacts (#4316), a simple fix would be to append a unique value to the tag of the E2E image built by each job. WDYT?

pebrc:
SGTM as temporary fix

PR #4343, opened 2021-03-15
Fix env vars in ARM CI job

charith-elastic:
The reason why the ARM E2E job runs tests that it's not supposed to is probably because the `E2E_TAGS` variable is not set correctly. In the Bourne family of shells, there shouldn't be a space between the variable name, equals sign and the value (https://unix.stackexchange.com/a/297217).

I am not sure why the job sometimes works though. The same issue is present in the other job definitions as well and they seem to work too. 🤷🏼  

Potential fix for #4314

### DIFF
--- a/.ci/setenvconfig
+++ b/.ci/setenvconfig
@@ -35,15 +35,15 @@ case $environment in
 ######################################
 
 write_env <<ENV
-IMG_SUFFIX = -ci
-E2E_REGISTRY_NAMESPACE = eck-ci
+IMG_SUFFIX=-ci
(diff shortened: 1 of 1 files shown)


pebrc (approved):
LGTM 🤞

thbkrkr (approved):
Weird!

Just for the history, the first purpose of the `.env` file generated by `setenvconfig` is to be included by `.ci/Makefile` and the root `Makefile`. That's why we had spaces between `=` to set the variables because it is the main way to declare variables with make.

The shell syntax without spaces should work just as well. As you said, it's really a mystery why this can create a random bug.

PR #4344, opened 2021-03-15
Fix OpenShift monitoring recipe

barkbay:
Deploying Metricbeat using `config/recipes/beats/openshift_monitoring.yaml` is failing with the following error:

```
Beat.beat.k8s.elastic.co "metricbeat" is invalid: privileged: Invalid value: "privileged": privileged field found in the kubectl.kubernetes.io/last-applied-configuration annotation is unknown. This is often due to incorrect indentation in the manifest.
```

This is because the `privileged` field is only available in the container `securityContext `.


thbkrkr (approved):
Good catch!

Issue #4128, opened 2021-01-15
OCP3 testing problems


pebrc:
Assuming #4118 is fixed. A few observations on testing on OCP3 on CI or with our e2e test suite in general:

* all Beats or Agent tests need a manual adjustment of the SELinux security context with something like `sudo chcon -Rt svirt_sandbox_file_t /var/lib/e2e-{random-string}-mercury/fb-{random-string}/filebeat-data` because they use hostPath volumes. 
* Beats and Agent also need an SCC for their privileged access but we only install the SCC if monitoring is turned on. this should be separated.

thbkrkr:
The [last run of the job](https://devops-ci.elastic.co/job/cloud-on-k8s-e2e-tests-ocp3/9/testReport/) failed because of:
```
18:19:57  === [31mFAIL[0m: test/e2e/agent TestSystemIntegrationConfig/Agent_status_should_be_updated (1800.00s)
18:19:57  Retries (30m0s timeout): .......................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................
18:19:57      utils.go:85: 
18:19:57          	Error Trace:	utils.go:85
18:19:57          	Error:      	Received unexpected error:
18:19:57          	            	expected status {Version:7.11.0 ExpectedNodes:0 AvailableNodes:0 Health:green ElasticsearchAssociationsStatus:} but got {Version:7.11.0 ExpectedNodes:0 AvailableNodes:0 Health:yellow ElasticsearchAssociationsStatus:}
18:19:57          	Test:       	TestSystemIntegrationConfig/Agent_status_should_be_updated

pebrc:
One thing to look into is whether it would be possible to run the Beats and Agent tests with `emptyDir` instead of `hostPath` volumes to step around the OCP3 SELinux restrictions. 

Another workaround would be to adopt a similar approach to our current ARM testing strategy, use Go build constraints to only run the tests we know can work: Elasticsearch, Kibana and Enterprise Search.

thbkrkr:
Thanks for the hints.

I'm not yet sure about the root cause of the failure. As in CI, I can see that the agent has a yellow health and its daemonset gets only 3 replicas out of 5. I assume it's because there are 5 nodes in the k8s cluster (1 lb, 1 master and 3 workers) but only 3 accept our workload?

I'm trying to reproduce what is happening and I do not have completely the same behaviour as in CI.

[...]

PR #4348, opened 2021-03-15
Reset node selector of managed namespaces on OCP3

thbkrkr:
When the environment is OCP3, execute `oc annotate --overwrite namespace <ns> openshift.io/node-selector=""` for all managed namespaces to prevent that daemonset pods from being scheduled on some nodes, because by default there can be a node selector at the OCP project level.

Relates to #4128.

### DIFF
--- a/test/e2e/cmd/run/run.go
+++ b/test/e2e/cmd/run/run.go
@@ -377,7 +377,7 @@ func (h *helper) createOperatorNamespaces() error {
 }
 
 func (h *helper) createManagedNamespaces() error {
-	log.Info("Creating managed namespacesS")
+	log.Info("Creating managed namespaces")
 	var err error
 	// when in local mode, don't delete the namespaces on exit
 	if h.testContext.Local {
--- a/test/e2e/cmd/run/run.go
+++ b/test/e2e/cmd/run/run.go
@@ -387,7 +387,7 @@ func (h *helper) createManagedNamespaces() error {
 	}
 
 	// Reset the node selector for all managed namespaces to override any possible OCP project node selector that might
-	// prevent scheduling daemonset pods on some nodes
+	// prevent scheduling daemonset pods on some nodes.
 	if h.testContext.Ocp3Cluster {
 		log.Info("Resetting namespace node selector")
 		for _, ns := range h.testContext.Operator.ManagedNamespaces {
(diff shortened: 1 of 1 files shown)


test/e2e/cmd/run/run.go
```diff
@@ -377,14 +377,28 @@ func (h *helper) createOperatorNamespaces() error {
 }
 
 func (h *helper) createManagedNamespaces() error {
-	log.Info("Creating managed namespaces")
+	log.Info("Creating managed namespacesS")
```
barkbay, line 380:
```suggestion
	log.Info("Creating managed namespaces")
```

test/e2e/cmd/run/run.go
```diff
@@ -377,14 +377,28 @@ func (h *helper) createOperatorNamespaces() error {
...
-		_, err := h.kubectlApplyTemplate("config/e2e/managed_namespaces.yaml", h.testContext)
-		return err
+		_, err = h.kubectlApplyTemplate("config/e2e/managed_namespaces.yaml", h.testContext)
+	} else {
+		err = h.kubectlApplyTemplateWithCleanup("config/e2e/managed_namespaces.yaml", h.testContext)
 	}
 
-	return h.kubectlApplyTemplateWithCleanup("config/e2e/managed_namespaces.yaml", h.testContext)
+	// Reset the node selector for all managed namespaces to override any possible OCP project node selector that might
+	// prevent scheduling daemonset pods on some nodes
```
barkbay, line 390:
```suggestion
	// prevent scheduling daemonset pods on some nodes.
```

test/e2e/cmd/run/run.go
```diff
@@ -377,14 +377,28 @@ func (h *helper) createOperatorNamespaces() error {
...
+	if h.testContext.Ocp3Cluster {
+		log.Info("Resetting namespace node selector")
+		for _, ns := range h.testContext.Operator.ManagedNamespaces {
+			if err := exec.Command("oc", "annotate", "--overwrite", "namespace", ns, "openshift.io/node-selector=").Run(); err != nil {
+				log.Info("err:", err)
```
barkbay, line 395:
```suggestion
				log.Error(err, "Error while setting openshift.io/node-selector annotation", "namespace", ns)
```

barkbay (approved):
lgtm

test/e2e/cmd/run/run.go
thbkrkr: Sorry I forgot to remove this _debug_ logger.
Fixed with 61911e2.

PR #4352, opened 2021-03-16
E2E local: run test script from within e2e cmd

### LINKED ISSUE #4332: e2e local mode starts cleaning resources before the start of the tests
When running E2E tests with `--local`, the resources cleanup begins before the tests start. Indeed, in normal mode, the last `runTestJob` step is blocking and the program waits for its end before executing the cleanup. In local mode, there is no blocking step and the cleanup is performed in the process. One particular downside is that it removes the newly created SCC.

pebrc:
This is to ensure any cleanup functions run after the test script has terminated and not before it has started running.

Fixes #4332


thbkrkr (commented):
Almost LGTM but `CTRL+C` doesn't stop the local script. I think it's important to keep this behaviour.

test/e2e/cmd/run/run.go
```diff
@@ -459,6 +461,26 @@ func (h *helper) deployTestSecrets() error {
 	)
 }
 
+func (h *helper) runTestScript() error {
+	log.Info("Running local test script", "timeout", h.testTimeout.String())
+	ctx, cancelFunc := context.WithTimeout(context.Background(), h.testTimeout)
+	defer cancelFunc()
+	cmd := exec.Command("test/e2e/run.sh", "-run", os.Getenv("TESTS_MATCH"), "-args", "-testContextPath", h.testContextOutPath) // nolint:gosec
```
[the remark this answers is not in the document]
charith-elastic: To provide some context on that, the Go convention is to not have any spaces on machine-readable comments. See the last sentence of https://golangci-lint.run/usage/false-positives/

thbkrkr:
Almost good.  CTRL+C works well.

2 little things:
- We can delete the call of the run.sh script present in the Makefile.
- If I don't have a `.env` file, it fails with `test/e2e/run.sh: ligne 15: GO_TAGS : variable sans liaison` because previously we set `GO_TAGS` with `E2E_TAGS`.

https://github.com/elastic/cloud-on-k8s/blob/671905f88ccc14d4388262439ab366b2b48e0c51/Makefile#L481

pebrc:
@thbkrkr good catch the last [merge from master](https://github.com/elastic/cloud-on-k8s/pull/4352/commits/6fc6bf7c2156c35d6d46a24c5b75286f259e82ca) reintroduced the `run.sh` invocation and the somehow also removed the environment variable. I probably messed up a merge conflict resolution.

test/e2e/cmd/run/run.go
```diff
@@ -459,7 +462,41 @@ func (h *helper) deployTestSecrets() error {
...
+			select {
+			case s := <-sigs:
+				log.Info("caught Unix signal", "signal", s)
+				cancelFunc()
+			case <-ctx.Done():
+				// exec.Command's support for contexts does not allow sending sigkill to the whole process group
+				// so we are doing it manually here. Go sets the process group to PID and kill on Linux and BSD supports
+				// sending signals to the whole process group if number passed to kill is negative see `man 2 kill`
+				err := syscall.Kill(-cmd.Process.Pid, syscall.SIGKILL)
+				log.Info("test cancelled", "kill_error", err)
```
thbkrkr, line 491:
```suggestion
				log.Info("Test cancelled", "kill_error", err)
```

test/e2e/cmd/run/run.go
```diff
@@ -459,7 +462,41 @@ func (h *helper) deployTestSecrets() error {
...
+	// listen to Unix signals to handle user abort
+	sigs := make(chan os.Signal, 1)
+	signal.Notify(sigs, syscall.SIGINT, syscall.SIGTERM)
+
+	go func() {
+	DONE:
+		for {
+			select {
+			case s := <-sigs:
+				log.Info("caught Unix signal", "signal", s)
```
thbkrkr, line 484:
```suggestion
				log.Info("Caught Unix signal", "signal", s)
```

thbkrkr (approved):
LGTM

Nit: All log info existing message start with an uppercase except the two you added.

PR #4356, opened 2021-03-17
E2E fix errors in monitoring setup

pebrc:
* Incorrect placement of `privileged` attribute in OCP
* Missing nodes privilege for Filebeat (new in a recent version of Filebeat?)
* Broken APM config for the operator

### DIFF
--- a/config/e2e/helm-operator-under-test.yaml
+++ b/config/e2e/helm-operator-under-test.yaml
@@ -20,6 +20,7 @@ podAnnotations:
 tracing:
   enabled: true
   config:
+    ELASTIC_APM_SERVER_URL: null
     ELASTIC_APM_SERVER_CERT: "/mnt/elastic/apm.crt"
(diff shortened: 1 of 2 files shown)


config/e2e/helm-operator-under-test.yaml
```diff
@@ -20,6 +20,7 @@ podAnnotations:
 tracing:
   enabled: true
   config:
+    ELASTIC_APM_SERVER_URL: null
```
pebrc, line 23: We have ha default of localhost in `values.yaml` that takes precedence otherwise.

charith-elastic: Yes, that make sense. It was only added to the values.yaml to act as an example of how to set different tracing related environment variables.

charith-elastic (approved):
LGTM

pebrc:
Jenkins test this please

PR #4357, opened 2021-03-17
Clean up unused dependencies

thbkrkr:
While running `hack/operatorhub/main.go`, I noticed that we had unused dependencies in its `go.sum`. So I ran `go mod tidy` wherever there is a `go.sum`.


thbkrkr:
jenkins test this please

thbkrkr:
I didn't read it carefully. These are zip sums of modules updated by 'go mod download'.

PR #4358, opened 2021-03-18
Add E2E namespace to diagnostic dump

charith-elastic:
Include the E2E namespace in the diagnostic bundle created at the end of a test run. 

The E2E namespace contains the monitoring deployment and the E2E test job itself. It is sometimes useful to see the status of those components when debugging a test failure.


thbkrkr (approved):
LGTM

Issue #4334, opened 2021-03-10
Missing SYS_CHROOT capability for Elasticsearch on OCP >=4.6 


thbkrkr:
To increase security, the `NET_RAW` and `SYS_CHROOT` capabilities are no longer available in the default list of CRI-O capabilities and this impacts OpenShift from version `4.6`.

So, if you deploy Elasticsearch on OCP `4.6`, **in the default namespace**, its pods will be in the `CrashLoopBackOff` status because of this error:

```
> k logs test-es-master-0 
chroot: cannot change root directory to '/': Operation not permitted
```

A quick solution to this problem is to add the necessary capacity directly to the container:

```yaml
apiVersion: elasticsearch.k8s.elastic.co/v1
kind: Elasticsearch
metadata:
  name: test
spec:
  version: 7.11.1
  nodeSets:
  - name: master
    count: 1
    podTemplate:
      spec:
        containers:
        - name: elasticsearch
          securityContext:
            capabilities:
              add: ["SYS_CHROOT"] 
```

Another approach consists to give the namespace the missing capability using a security context.

~I think it's annoying enough to deserve a note in the 'Deploy ECK on OpenShift' [documentation](https://www.elastic.co/guide/en/cloud-on-k8s/master/k8s-openshift-deploy-elasticsearch.html)?~

When following the [Deploy ECK on OpenShift documentation](https://www.elastic.co/guide/en/cloud-on-k8s/master/k8s-openshift-deploy-the-operator.html) which asks to create a new project `oc new-project elastic`, there is no need to add this capability because new projects get more permissions.

It is visible by looking at the template used to create projects: `oc adm create-bootstrap-project-template -o yaml > template.yaml`.

thbkrkr:
Additional question: why does it work out of the box in our E2E tests?

alaudazzi:
@thbkrkr 

> I think it's annoying enough to deserve a note in the 'Deploy ECK on OpenShift' documentation?

Sorry, I'm not sure I understand what you mean here.

thbkrkr:
@alaudazzi I mean that this new setting required to run Elasticsearch on OCP> = 4.6 should surely be documented as it is difficult to find out on your own.

alaudazzi:
@thbkrkr Here is a proposal for the note to be added to [Deploy an Elasticsearch instance with a route](https://www.elastic.co/guide/en/cloud-on-k8s/master/k8s-openshift-deploy-elasticsearch.html):

[...]

PR #4359, opened 2021-03-18
Add a note on missing capabilities for OCP >=4.6

alaudazzi:
This updates the page Deploy an [Elasticsearch instance with a route](https://www.elastic.co/guide/en/cloud-on-k8s/master/k8s-openshift-deploy-elasticsearch.html) with a note on capabilities no longer available with OPC 4.6. 

Fixes: #4334

### DIFF
--- a/docs/advanced-topics/openshift.asciidoc
+++ b/docs/advanced-topics/openshift.asciidoc
@@ -101,6 +101,16 @@ spec:
 EOF
 ----
 
+On OpenShift 4.6, `NET_RAW` and `SYS_CHROOT` are no longer available in the default list of CRI-O capabilities. When you deploy Elasticsearch on OpenShift 4.6, to avoid having your pods in `CrashLoopBackOff` status, you can add the necessary capabilities directly to the container:
+ 
+[source,shell]
+----
+- name: elasticsearch
+  securityContext:
+    capabilities:
+     add: ["SYS_CHROOT"]
+----
+
 [id="{p}-openshift-es-plugins"]
 === Elasticsearch plugins
 


thbkrkr:
Thank you for taking care of this.

I'm a little worried that the quickstart experience may be degraded if the user deploys the quickstart and then reads the note.
The snippet can also be confusing (`name` is a field of `containers` but also of `nodeSets`).

So, having said that, seeing what it gives live, I wonder if it might not be better to directly update the example of manifest with a comment. We could also comment the whole section but I imagine using the lastest OCP versions (4.6 or 4.7) is a pretty common use case. 

That would give that:
```diff
diff --git a/docs/advanced-topics/openshift.asciidoc b/docs/advanced-topics/openshift.asciidoc
index 5286a624..e1022f5d 100644
--- a/docs/advanced-topics/openshift.asciidoc
+++ b/docs/advanced-topics/openshift.asciidoc
@@ -85,6 +85,14 @@ spec:
     count: 1
     config:
       node.store.allow_mmap: false
+    # required from OpenShift 4.6 as `NET_RAW` and `SYS_CHROOT` are no longer available in the default list of CRI-O capabilities
+    podTemplate:
+      spec:
+        containers:
+        - name: elasticsearch
+          securityContext:
+            capabilities:
+              add: ["SYS_CHROOT"]
 ---
```

barkbay (commented):
I just did a quick test following the [current documentation](https://www.elastic.co/guide/en/cloud-on-k8s/master/k8s-openshift-deploy-the-operator.html) on a brand new Openshift 4.6 cluster. Everything seems to be working as expected:

```
> oc version                                                                                                                                                                      
Client Version: 4.6.3
Server Version: 4.6.1
Kubernetes Version: v1.19.0+d59ce34

> oc get es,pods                                                                                                                                                                  
NAME                                                              HEALTH   NODES   VERSION   PHASE   AGE
elasticsearch.elasticsearch.k8s.elastic.co/elasticsearch-sample   green    1       7.11.2    Ready   74s

NAME                                    READY   STATUS    RESTARTS   AGE
pod/elasticsearch-sample-es-default-0   1/1     Running   0          73s
```

Do we know if this modification is necessary in all cases?

thbkrkr:
> Do we know if this modification is necessary in all cases?

This is necessary on the 4.7 that I have on hand but I may have tweaked it a bit for other purposes (debugging e2e tests). I will check again with fresh clusters.

thbkrkr:
After executing `oc new-project elastic`, it works well on OCP 4.6 and 4.7 without adding `SYS_CHROOT` capability.

thbkrkr:
Closing because deploying in the `default` namespace shouldn't be the nominal case.
If the existing documentation is followed, it works well.
Sorry for the noise.

thbkrkr:
When following the [Deploy ECK on OpenShift doc](https://www.elastic.co/guide/en/cloud-on-k8s/master/k8s-openshift-deploy-the-operator.html) which asks to create a new project `oc new-project elastic`, there is no need to add this capability because new projects get more permissions.

It is visible by looking at the template used to create projects: `oc adm create-bootstrap-project-template -o yaml > template.yaml`.

PR #4363, opened 2021-03-19
correct list of version to upgrade from

jguay:
currently 1.4 doc mentions upgrade from `It is compatible with the previous GA releases (1.0.x, 1.1.x and 1.2.x)` omitting 1.3.x, changing wording to `(1.0.x and higher)` to try to avoid same issue if/when we release a 1.5.x


alaudazzi (approved):
LGTM

barkbay:
I'll take care of the "backport" to master and 1.5

PR #4368, opened 2021-03-22
Fix case of variables in readiness probe

m1pl:
The readiness check fails in k8s because of this.

### DIFF
--- a/pkg/controller/elasticsearch/nodespec/readiness_probe.go
+++ b/pkg/controller/elasticsearch/nodespec/readiness_probe.go
@@ -57,8 +57,8 @@ else
 fi
 
 # setup basic auth if credentials are available
-if [ -n "${PROBE_USERNAME}" ] && [ -f "${probe_password_path}" ]; then
-  PROBE_PASSWORD=$(<${probe_password_path})
+if [ -n "${PROBE_USERNAME}" ] && [ -f "${PROBE_PASSWORD_PATH}" ]; then
+  PROBE_PASSWORD=$(<${PROBE_PASSWORD_PATH})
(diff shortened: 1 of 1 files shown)


pkg/controller/elasticsearch/nodespec/readiness_probe.go
```diff
@@ -57,8 +57,8 @@ else
 fi
 
 # setup basic auth if credentials are available
-if [ -n "${PROBE_USERNAME}" ] && [ -f "${probe_password_path}" ]; then
-  PROBE_PASSWORD=$(<${probe_password_path})
+if [ -n "${PROBE_USERNAME}" ] && [ -f "${PROBE_PASSWORD_PATH}" ]; then
```
barkbay, line 60: `PROBE_PASSWORD_PATH` is not used on purpose here because it was not set in beta1. `probe_password_path` should be initialized right before:

```bash
# Check if PROBE_PASSWORD_PATH is set, otherwise fall back to its former name in 1.0.0.beta-1: PROBE_PASSWORD_FILE
if [[ -z "${PROBE_PASSWORD_PATH}" ]]; then
  probe_password_path="${PROBE_PASSWORD_FILE}"
else
  probe_password_path="${PROBE_PASSWORD_PATH}"
fi
```

barkbay (commented):
Hi,

Could you give us more context about the situation in which the probe is failing and your PR fixed the problem ?

Thanks

pkg/controller/elasticsearch/nodespec/readiness_probe.go
m1pl: My bad, I messed up my configmap. The issue was apparently a different one.

PR #4371, opened 2021-03-23
Update Google Cloud LB recipe

charith-elastic:
- Update the Ingress API version as it is deprecated in all of the available GKE Kubernetes versions.
- Add an example to demonstrate serving Kibana from a path


config/recipes/gclb/README.asciidoc
```diff
@@ -56,3 +56,12 @@ Access Elasticsearch.
 ----
 curl -H "Host: elasticsearch.hulk" --resolve "elasticsearch.hulk:443:$INGRESS_HOST" -k -u "elastic:$ELASTICSEARCH_PASSWORD" 'https://elasticsearch.hulk/_cat/health?v'
 ----
+
```
pebrc, line 59: Not added in this PR but should we add a not about the minimal version of cert-manager you are expecting (I had an old one installed and need to upgrade first before this recipe worked ootb)

pebrc (approved):
LGTM

PR #4372, opened 2021-03-23
Clarify shard allocation awareness

charith-elastic:
Touch-up the sentence about why the `k8s_node_name` attribute is used to set shard allocation awareness.


barkbay (approved):
lgtm

Issue #4370, opened 2021-03-23
beat.TestFilebeatAutodiscoverRecipe is failing


thbkrkr:
`test/e2e/beat.TestFilebeatAutodiscoverRecipe/Remove_Beat_if_it_already_exists` failed in several nightly jobs:

https://devops-ci.elastic.co/job/cloud-on-k8s-e2e-tests-stack-versions/376/testReport
https://devops-ci.elastic.co/job/cloud-on-k8s-e2e-tests-eks/319/testReport/
https://devops-ci.elastic.co/job/cloud-on-k8s-e2e-tests-gke-k8s-versions/371/testReport/
https://devops-ci.elastic.co/job/cloud-on-k8s-e2e-tests-stack-versions/376/testReport/

I took a quick look and the test seems stuck at the initialization step when it tries to clean up the filebeat resource because the 
namespace is missing. The suffix is also missing from the filebeat name.

https://github.com/elastic/cloud-on-k8s/blob/ae5ab3cd5f3f1beb1e3beb2d31f8fa9064ab0fb3/test/e2e/test/beat/steps.go#L53-L58

The error is `an empty namespace may not be set when a resource name is provided`.

thbkrkr:
Closed by #4374.

PR #4339, opened 2021-03-11
Update golangci-lint

charith-elastic:
- Version of the linter used in CI is quite old and lots of improvements have been made since 
- Cleanup the config file and move exclusions inline -- which can be checked for correctness by the `nolintlint` linter.

PR #4374, opened 2021-03-23
Fix E2E builders

thbkrkr:
This fixes a bug introduced in https://github.com/elastic/cloud-on-k8s/pull/4339#discussion_r594442104.

https://github.com/elastic/cloud-on-k8s/blob/ae5ab3cd5f3f1beb1e3beb2d31f8fa9064ab0fb3/test/e2e/test/helper/yaml.go#L266-L278

```go
/*L268*/ builder = b.WithNamespace(namespace)           // ns of b value is not updated, only ns of builder
/*L277*/ builder = b.WithPodTemplateServiceAccount(sa)  // then we reset the ns of builder here
```

Relates to #4370.


barkbay (approved):
lgtm

Issue #4316, opened 2021-03-08
E2E Docker image is too big


pebrc:
Building and pushing a e2e docker image I noticed that it has grown to 2 GB in size. Most of it seems to be the Go module dependency cache. 

I wonder if we could just build the test binaries and just add the compiled binary which presumably is much much smaller in size. The disadvantage is of course that we have to build a separate image for each test run but I think that is already the case.
![Screenshot 2021-03-08 at 14 01 44](https://user-images.githubusercontent.com/697790/110325588-d02ac880-8017-11eb-88b9-d2f9bd9f4803.png)

pebrc:
I did a few experiments with `go test -c -tags e2e -json $PKG` and it looks like we could bring down the size to about 400M based on our current tests:

```
 50M	agent.test
 50M	apm.test
 50M	beat.test
 50M	e2e.test
 46M	elasticsearch.test
 50M	ent.test
 50M	es.test
 50M	kb.test
396M	total
```

One issue to keep in mind is that the we are currently relying on relative paths to checked out configs e.g. for the sample tests that break when pre-compiling the test unless we copy the samples into the image as well.

PR #4377, opened 2021-03-24
E2E: precompile tests for a smaller Docker image

pebrc:
I am not 💯 convinced that it is an improvement as we trade off slower compile time (~ 15 minutes on my machine as we have to build per package) vs smaller image size. 

TODO: 
* directory structure in the `test` package is somewhat inconsistent, we could save a couple of 100M by excluding all the test infrastructure from the build + moving the `cmd` bits into the test infra package + closing build tag gaps
* the chaos job should not be part of the e2e command imo but belongs to the test runner 
* It would be good if we could compile all tests into a single binary (something go tools don't support out of the box)
* Rebase once https://github.com/elastic/cloud-on-k8s/pull/4352 is merged

Fixes https://github.com/elastic/cloud-on-k8s/issues/4316


pebrc:
Closing this until we find time to bring this to completion.

PR #4385, opened 2021-03-26
Improve Stack version update script

thbkrkr:
The current script `hack/update-samples-stack-version.sh` is buggy:
- the pattern `[0-9]+\.[0-9]+\.[0-9]` is too open and upgrades a package version in an agent recipe.
- the script is not exhaustive either and we regularly miss versions (because at first it was created to just update the samples)

This commit changes the way things work by replacing the existing version with the new version, rather than using a pattern and also makes the change in the missing files.

Relates to #3612.


hack/update-stack-version.sh
```diff
@@ -0,0 +1,35 @@
...
+
+set -eu
+
+# Elastic Stack versions
+PREVIOUS_VERSION="$1"
+NEW_VERSION="$2"
+
+# Use the "-i.bak" trick to be compatible with both Linux and OSX
+bump_version() {
+	sed -i.bak -E "s|${PREVIOUS_VERSION}|${NEW_VERSION}|g" "$1"
```
charith-elastic, line 20: One minor issue with this is that `update-stack-version.sh 7.11.2 7.12.0` will also replace `791192` (or any similar string) with `7.12.0` as well. Not sure whether it's worth worrying about though.

It'd be nice to have a usage message when the script is invoked without any arguments. Maybe that can show how to escape the version to avoid the above problem. (`update-stack-version.sh 7\.11\.2 7.12.0`)

charith-elastic (approved):
LGTM

hack/update-stack-version.sh
thbkrkr: Good point. I will add a usage message.

I don't want to make the script to complicated by adding validation. Especially in bash, it quickly becomes a hassle to validate `\.` .

But now I'm thinking that we could derive the previous version to escape the dots. It doesn't sound so bad.

```sh
previous_version() {
	grep STACK_VERSION Makefile | grep -Eo "[0-9]+\.[0-9]+\.[0-9]+" | sed "s|\.|\\\.|g"
}
```

thbkrkr: Back from WE, I wonder: why not rewrite the script in Go?

charith-elastic: I think the shell utilities are highly optimized for this problem and anything we write ourselves will require a lot of work to deal with edge cases and performance (evidenced by "I re-implemented X in Y" articles).  So it's probably not worth the effort? The chances of there being a string like `791192` in the code is pretty unlikely, and if it does happen, hopefully it would be very obvious from a quick glance through the diff.

thbkrkr: Yes, let's stay there, it's already an improvement.

Issue #4360, opened 2021-03-18
agent.TestKubernetesIntegrationRecipe is failing


thbkrkr:
`test/e2e/agent.TestKubernetesIntegrationRecipe/ES_data_should_pass_validations` failed in all nightly jobs:

https://devops-ci.elastic.co/job/cloud-on-k8s-e2e-tests-snapshot-versions/275/testReport/
https://devops-ci.elastic.co/job/cloud-on-k8s-e2e-tests-eks/316/testReport
https://devops-ci.elastic.co/job/cloud-on-k8s-e2e-tests-aks/648/testReport
https://devops-ci.elastic.co/job/cloud-on-k8s-e2e-tests-kind-k8s-versions/383/testReport
https://devops-ci.elastic.co/job/cloud-on-k8s-e2e-tests-stack-versions/373/testReport

[`cloud-on-k8s-e2e-tests-stack-versions`](https://devops-ci.elastic.co/blue/organizations/jenkins/cloud-on-k8s-e2e-tests-stack-versions/detail/cloud-on-k8s-e2e-tests-stack-versions/373/pipeline/48) shows that the failure only occurs with stack version `7.11.2`, which was recently updated with https://github.com/elastic/cloud-on-k8s/pull/4355.
`6.8.x` failed because of another test: `kb.TestVersionUpgradeAndRespecToLatest7x`.

pebrc:
Looks like the agent is just sitting idle: 
```
==== START logs for pod e2e-mlmzr-mercury/pod/elastic-agent-jwj6-agent-6p8lg ====
2021-03-18T13:04:39.693Z        INFO    warn/warn.go:18 The Elastic Agent is currently in BETA and should not be used in production
2021-03-18T13:04:39.694Z        INFO    application/application.go:59   Detecting execution mode
2021-03-18T13:04:39.694Z        INFO    application/application.go:68   Agent is managed locally
2021-03-18T13:04:42.000Z        INFO    [composable]    composable/controller.go:44     EXPERIMENTAL - Inputs with variables are currently experimental and should not be used in production
2021-03-18T13:04:42.093Z        INFO    [composable.providers.docker]   docker/docker.go:43     Docker provider skipped, unable to connect: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
[...]

PR #4393, opened 2021-03-30
Adjust Elastic agent Kubernetes recipe

thbkrkr:
Use the correct variable syntax with the `env` prefix in the configuration of the Elastic agent in the Kubernetes recipe.

> The fact that 7.11.x worked with ${NODE_NAME} in standalone mode was a bug. The correct variable syntax is ${env.NODE_NAME} in 7.11.x.

Relates to #4360.


thbkrkr:
elastic/beats#24842 is still under investigation. I will reopen if necessary.

pebrc (approved):

PR #4404, opened 2021-04-02
Allow users to spec a serviceName in associations

### LINKED ISSUE #3732: Create the ability to target a custom service with `elasticsearchRef`
Currently `elasticsearchRef` always uses the default HTTP service ECK creates to make requests to Elasticsearch. While we offer the ability to customize that default service and adjust its selector there are cases where that is not enough: 

Consider: 
* target coordinator nodes with Kibana
* target data/ingest nodes with Beats

We should investigate whether it makes sense to allow users to specify a custom service in the `elasticsearchRef`. 

```yaml
elasticsearchRef:
  name: my-elasticsearch-cluster
  service: my-custom-service
```

pebrc:
Fixes #3732 

1000 lines of YAML say more than 100 words: 

```yaml
kind: Elasticsearch
metadata:
  annotations:
  name: test-coord
  namespace: default
spec:
  nodeSets:
  - config:
      node.roles: []
      node.store.allow_mmap: false
    count: 2
    name: coord
    podTemplate:
      metadata:
        labels:
          app: coord-node
  - config:
      node.store.allow_mmap: false
    count: 2
    name: default
  version: 7.12.0
---
kind: Kibana
metadata:
  name: kibana-sample
spec:
  version: 7.12.0
  count: 1
  elasticsearchRef:
    name: "test-coord"
    serviceName: "es-coord"
--- 
apiVersion: v1
kind: Service
metadata:
  name: es-coord
  namespace: default
spec:
  ports:
  - name: https
    port: 80
    protocol: TCP
    targetPort: 9200
  selector:
    app: coord-node
```

Add a new `serviceName` attribute to all associations (even though it makes only sense for Elasticsearch but the tradeoff here is additional complexity in the code). Query for the referenced service, extract the configured port which matches the currently configured protocol of the managed resource (e.g. Elasticsearch running with HTTPS) and use it to construct a service URL which will be configured in the associated resource.

* users need to create a service either ahead of time or once the Elastic stack resource have been deployed.
* users need to follow the port naming convention we use in ECK i.e. `https` or `http` (which is compatible with [Istio's port naming convention](https://istio.io/latest/docs/reference/config/analysis/ist0118/) btw)

### DIFF
--- a/pkg/apis/common/v1/common.go
+++ b/pkg/apis/common/v1/common.go
@@ -50,7 +50,8 @@ type ObjectSelector struct {
 	// Namespace of the Kubernetes object. If empty, defaults to the current namespace.
 	Namespace string `json:"namespace,omitempty"`
 	// ServiceName is the name of an existing Kubernetes service which will be used to make requests to the referenced
-	// object. If left empty the default HTTP service of the referenced resource will be used.
+	// object. It has to be in the same namespace as the referenced resource. If left empty the default HTTP service of
+	// the referenced resource will be used.
 	ServiceName string `json:"serviceName,omitempty"`
 }
 
--- a/pkg/apis/common/v1/common.go
+++ b/pkg/apis/common/v1/common.go
@@ -49,6 +49,9 @@ type ObjectSelector struct {
 	Name string `json:"name"`
 	// Namespace of the Kubernetes object. If empty, defaults to the current namespace.
 	Namespace string `json:"namespace,omitempty"`
+	// ServiceName is the name of an existing Kubernetes service which will be used to make requests to the referenced
+	// object. If left empty the default HTTP service of the referenced resource will be used.
+	ServiceName string `json:"serviceName,omitempty"`
 }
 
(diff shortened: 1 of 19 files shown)


pebrc:
Jenkins test this please

pkg/apis/common/v1/common.go
```diff
@@ -49,6 +49,9 @@ type ObjectSelector struct {
 	Name string `json:"name"`
 	// Namespace of the Kubernetes object. If empty, defaults to the current namespace.
 	Namespace string `json:"namespace,omitempty"`
+	// ServiceName is the name of an existing Kubernetes service which will be used to make requests to the referenced
+	// object. If left empty the default HTTP service of the referenced resource will be used.
+	ServiceName string `json:"serviceName,omitempty"`
```
sebgl, line 54: Let's make it clear in the comment and generated doc that the service must exist in the referenced object namespace?

sebgl (approved):
LGTM

PR #4406, opened 2021-04-03
Correct Kibana settings in the Stack Monitoring recipe

eedugon:
`xpack.monitoring.collection.enabled: true` is an ES setting and not kibana.
I haven't verified the rest of the recipe examples.

The only needed setting at Kibana level to enable `metricbeat based monitoring` in Kibana is `monitoring.kibana.collection.enabled: false`. The other setting should be set at ES level (although it's true by default).

I'm also fixing the right Kibana setting as the name doesn't have `xpack` prefix anymore.


pebrc:
Jenkins test this please

pebrc (approved):

PR #4416, opened 2021-04-12
Deployer: fix diskSetup in plans.yml for GKE/EKS/AKS

barkbay:
IIUC since [#3768](https://github.com/elastic/cloud-on-k8s/pull/3768/files#diff-18aecebc864709ce6c3a0569ff7eda975fffb7871a66a72c047ee404f7f0651bR33) `diskSetup` is a plan setting, not related to the K8S cluster providers.

This PR moves `diskSetup` settings from the provider sections to their respective plan sections, I think it also means that SSD disks were not used anymore.

### DIFF
--- a/hack/deployer/config/plans.yml
+++ b/hack/deployer/config/plans.yml
@@ -7,13 +7,13 @@ plans:
   machineType: n1-standard-8
   serviceAccount: true
   psp: true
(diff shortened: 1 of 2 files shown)


thbkrkr (approved):
Good catch! LGTM.

PR #4417, opened 2021-04-12
[Autoscaling] Run e2e test only if sc with volume expansion is available

### LINKED ISSUE #4414: TestAutoscaling is flaky
When using the fixed decider the value of `required_capacity.node.storage` in the response is based on the value provided in the `storage` decider parameter, which is expected. However, the returned value for `required_capacity.total.storage` seems to be based on the perceived total capacity. In the context of Kind this value is the storage capacity available on the K8S node itself. This is likely because `rancher.io/local-path` does not expose a "real" dedicated volume, but rather a directory on the host:

Here is the fixed decider parameter used in the e2e test:

```json
{"storage": "20gb", "nodes": "3"}
```

Which produces the following response on `Kind` with `rancher.io/local-path`:

```yaml
log.logger: elasticsearch-autoscaling
message: Required capacity for policy
iteration: '938'
namespace: e2e-cimgr-mercury
es_name: test-autoscaling-ntfh
policy: data-ingest
required_capacity:
  node:
    storage: 21 474 836 480 # expected because of the fixed decider set to  { "storage": "20gb" }
  total:
    storage: 249 386 926 080 # ~250Gi of required total capacity, this is not expected by the test
current_capacity:
  node:
    storage: 124 693 463 040 # seems to be the available storage on the host, not in the Pod
    memory: 2 147 483 648
  total:
    storage: 249 386 926 080 # 2 x current_capacity.node.storage
    memory: 4 294 967 296 # 2 x current_capacity.node.memory
current_capacity.count: 2
current_nodes:
- name: test-autoscaling-ntfh-es-data-ingest-0
- name: test-autoscaling-ntfh-es-data-ingest-1
```

Also the autoscaling controller attempts to scale vertically but the storage class does not support volume expansion:

```json
{
 "@timestamp": "2021-04-09T03:49:22.156Z",
 "log.logger": "elasticsearch-autoscaling",
 "message": "Horizontal autoscaler",
 "policy": "data-ingest",
 "scope": "tier",
 "count": 4,
 "required_capacity": {
  "storage": 249 386 926 080
 }
}
```

```json
{"log.level":"debug","@timestamp":"2021-04-09T03:19:54.485Z","log.logger":"elasticsearch-controller","message":"Recording event","service.version":"1.6.0-SNAPSHOT+f52ca470","service.type":"eck","ecs.version":"1.4.0","event":{"EventType":"Warning","Reason":"ReconciliationError","Message":"Failed to apply spec change: handle volume expansion: claim elasticsearch-data (storage class e2e-default) does not support volume expansion"}}
```

Given the nature of the storage class I do not think it makes sense to run this test on Kind. I'm still investigating other failures.
This issue however reveals an other situation which might happen when the available storage is larger than the one specified in the claim, I'll open a dedicated issue for this one.

barkbay:
This PR updates the autoscaling e2 test so it only runs if a storage class which supports volume expansion is available.

It should fix [the issue encountered with the `rancher.io/local-path` provider](https://github.com/elastic/cloud-on-k8s/issues/4414).

### DIFF
--- a/test/e2e/es/autoscaling_test.go
+++ b/test/e2e/es/autoscaling_test.go
@@ -15,12 +15,14 @@ import (
 	esv1 "github.com/elastic/cloud-on-k8s/pkg/apis/elasticsearch/v1"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common/version"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/elasticsearch/nodespec"
+	"github.com/elastic/cloud-on-k8s/pkg/controller/elasticsearch/volume"
 	"github.com/elastic/cloud-on-k8s/pkg/utils/stringsutil"
 	"github.com/elastic/cloud-on-k8s/test/e2e/test"
 	"github.com/elastic/cloud-on-k8s/test/e2e/test/elasticsearch"
 	"github.com/stretchr/testify/require"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 // TestAutoscaling ensures that the operator is compatible with the autoscaling Elasticsearch API.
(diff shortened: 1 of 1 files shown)


thbkrkr (approved):
LGTM

Issue #4420, opened 2021-04-14
E2E OCP cluster is not created 


barkbay:
Looks like it is related to the recent changes we made on the [OCP deployer](https://github.com/elastic/cloud-on-k8s/pull/4387) :

```
02:42:00  Error: clientVersion must not be empty
02:42:00  clientVersion must not be empty
02:42:00  make: *** [Makefile:500: run-deployer] Error 1
02:42:00  Makefile:48: recipe for target 'ci-internal' failed
```

At first glance I would say this is because `clientVersion` is not set for `ocp-ci`:

```yaml
- id: ocp-ci
  operation: create
  clusterName: ci
  provider: ocp
  machineType: n1-standard-8
  serviceAccount: true
  ocp:
    region: europe-west2
    nodeCount: 3
- id: ocp-dev
  operation: create
  clusterName: dev
  clientVersion: 4.7.0
  provider: ocp
  machineType: n1-standard-8
  serviceAccount: true
  ocp:
    region: europe-west1
    nodeCount: 3
```

Issue #4319, opened 2021-03-08
Test all supported versions of OpenShift


pebrc:
We are currently only testing OpenShift 4.7. and OpenShift 3.11. We should however test all version we claim to support through automated e2e tests. 

To discuss: 
* frequency of tests (we are already running a lot of tests on a daily basis, we could try to be a bit smarter and test with a lower frequency to minimise costs and resource consumption, for example weekly interval would still give us the necessary signal in case we create an incompatibility. At the same time a weekly interval would give us enough lead time to address the issue before the next release) 
* current structure of our e2e test image (the CI image is currently monolithic and with 2G way too big, the main reason is that we use one image for all test pipelines. We should investigate a better layering based on a common base image with dedicated images for the individual pipelines)

PR #4422, opened 2021-04-15
Make OCP deployer CI compatible

pebrc:
Fixes https://github.com/elastic/cloud-on-k8s/issues/4420
Fixes https://github.com/elastic/cloud-on-k8s/issues/4319

The first iteration of this revised OCP driver was not really CI compatible: 
* We cannot use `eck-dev` images in CI => change to `eck-ci`
* We need to login into to docker because we cannot rely on a Docker login happening outside of the deployer invocation
* We need to use a shared volume between CI container and client container, which we don't need in local mode because the Make process is not running in a container as well. 
* We need to specify the  OCP version in the Jenkins job we want to test. To conserve resources if the version is not specified it just picks one of the supported versions which are tested round robin during the course of a week

I have tested this on a local Jenkins instance. But I have not yet tested the flow where this job is called from an upstream job as a dependent job.

### DIFF
--- a/hack/deployer/runner/ocp.go
+++ b/hack/deployer/runner/ocp.go
@@ -74,8 +74,11 @@ type OcpDriverFactory struct {
 }
 
 type runtimeState struct {
-	// Authenticated tracks authentication against the GCloud API to avoid double authentication
+	// Authenticated tracks authentication against the GCloud API to avoid double authentication.
 	Authenticated bool
+	// SafeToDeleteWorkdir indicates that the installer state has been uploaded successfully to the storage bucket or is
+	// otherwise not needed anymore.
+	SafeToDeleteWorkdir bool
 	// ClusterStateDir is the effective work dir containing the OCP installer state. Derived from plan.Ocp.Workdir.
 	ClusterStateDir string
 	// ClientImage is the name of the installer client image.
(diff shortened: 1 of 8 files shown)


hack/deployer/runner/ocp.go
```diff
@@ -425,20 +446,19 @@ func (d *OcpDriver) bucketParams() map[string]interface{} {
...
+		"ClusterStateDirBase": filepath.Base(d.runtimeState.ClusterStateDir),
 		"HomeVolume":          SharedVolumeName(),
 		"GCloudCredsPath":     filepath.Join("/home", GCPDir, ServiceAccountFilename),
 		"OCPToolsDockerImage": d.runtimeState.ClientImage,
 		"Action":              action,
 	}
 	cmd := NewCommand(`docker run --rm \
-		-v {{.HomeVolume	}}:/home \
-		-v {{.ClusterStateDir}}:/config \
+		-v {{.HomeVolume}}:/home \
```
barkbay, line 456: I think this volume mount can be confusing as it is not the `/home` directory which is mounted here but the home directory of the current user.

barkbay (approved):
LGTM

hack/deployer/runner/ocp.go
pebrc: I tried to make it clearer by adding a comment and renaming the template parameter.

hack/deployer/runner/ocp.go
```diff
@@ -425,20 +446,19 @@ func (d *OcpDriver) bucketParams() map[string]interface{} {
...
 		-v /tmp:/tmp \
 		-e GOOGLE_APPLICATION_CREDENTIALS={{.GCloudCredsPath}} \
-		-e USER_HOME=/home \
+		-e HOME=/home \
```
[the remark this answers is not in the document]
pebrc: Mostly so that the GCloud tools pick up the credentials.

PR #4434, opened 2021-04-20
Remove incorrect statement about specifying coordinating roles in YAML

pebrc:
While we are not happy with the hack we added to address https://github.com/elastic/cloud-on-k8s/issues/3718 from a user point of view `node.roles: []` works just fine. Where as `!!seq ""` never worked. Let's remove this note and update any affected examples.


sebgl (approved):
LGTM

Issue #4179, opened 2021-02-02
Support Elastic Maps Service


pebrc:
We should investigate whether ECK needs to explicitly support Elastic Maps Service for air-gapped environments. 

A simple deployment of Elastic Maps Server can be done with a standard k8s Deployment resource. 

```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: ems
  labels:
     app.kubernetes.io/name: ems
spec:
  replicas: 1
  selector:
    matchLabels:
       app.kubernetes.io/name: ems
  template:
    metadata:
      labels:
         app.kubernetes.io/name: ems
    spec:
      containers:
        - name: ems
          image: docker.elastic.co/elastic-maps-service/elastic-maps-server-ubi8:7.11.0-SNAPSHOT
          ports:
            - containerPort: 8080
          env:
            - name: HOST
              value: 0.0.0.0
            - name: ELASTICSEARCH_HOST
              value: "https://elasticsearch-es-http.default.svc:9200"
            - name: ELASTICSEARCH_USERNAME
              value: "elastic" # TODO this should be a user with monitor privileges instead
            - name: ELASTICSEARCH_PASSWORD
              valueFrom:
                secretKeyRef:
                  name: elasticsearch-es-elastic-user
                  key: elastic
            - name: ELASTICSEARCH_SSL_CERTIFICATEAUTHORITIES
              value: /usr/src/app/server/certificates/ca.crt
          volumeMounts:
            - name: ca-certs
              mountPath: /usr/src/app/server/certificates
              readOnly: true
      volumes:
        - name: ca-certs
          secret:
            secretName: elasticsearch-es-http-certs-public
---
apiVersion: v1
kind: Service
metadata:
  name: ems
  labels:
    app.kubernetes.io/name: ems
spec:
  ports:
    - name: "tcp"
      port: 8080
      targetPort: 8080
  selector:
    app.kubernetes.io/name: ems
---
apiVersion: elasticsearch.k8s.elastic.co/v1
kind: Elasticsearch
metadata:
  name: elasticsearch
spec:
  version: 7.11.0-SNAPSHOT
  nodeSets:
    - name: default
      count: 3
      config:
        node.store.allow_mmap: false
---
apiVersion: kibana.k8s.elastic.co/v1
kind: Kibana
metadata:
  name: kibana
spec:
  version: 7.11.0-SNAPSHOT
  count: 1
  config:
    map.emsUrl: http://ems.default.svc:8080
[...]

PR #4439, opened 2021-04-21
Add Elastic Maps Server CRD

pebrc:
First part of #4179 split up to allow easier review. 

Adds the new CRD to support Elastic Maps Server in the ECK operator. 

```yaml
apiVersion: maps.k8s.elastic.co/v1alpha1
kind: ElasticMapsServer
metadata:
  name: ems
  labels:
    app: ems-demo
spec:
  version: 7.12.0
  podTemplate:
    spec:
      containers:
      - name: maps
        volumeMounts:
        - name: map-data
          readOnly: true
          mountPath: /usr/src/app/data
      volumes:
        - name: map-data
          persistentVolumeClaim:
            claimName: pv-claim-ems
  count: 1
  config:
    logging.level: trace
  elasticsearchRef:
    name: es
```

Compared to earlier versions I have adjusted the naming to `ElasticMapsServer`. I decided not to expose `appsv1.DeploymentStrategy`: 
* there is currently no justification to change the DeploymentStrategy default value for EMS
* to be consistent with similar CRDs like Kibana, EnterpriseSearch and APM Server
* additions are easier than removals: if we feel strongly later on that it is needed we can still add it


pebrc:
Jenkins test this please

pkg/apis/maps/v1alpha1/webhook.go
```diff
@@ -0,0 +1,88 @@
...
+	validationLog.V(1).Info("Validate update", "name", m.Name)
+	oldObj, ok := old.(*ElasticMapsServer)
+	if !ok {
+		return errors.New("cannot cast old object to ElasticMapsServer type")
+	}
+
+	return m.validate(oldObj)
+}
+
+func (m *ElasticMapsServer) validate(_ *ElasticMapsServer) error {
```
barkbay, line 62: nit: any reason for keeping this function's argument if there is no update check ? (or maybe update checks are missing ?)
```suggestion
func (m *ElasticMapsServer) validate() error {
```

barkbay (approved):
LGTM

pkg/apis/maps/v1alpha1/webhook.go
pebrc: There is no upgrade check because there is currently no constraint on upgrades (you can even downgrade if you are so inclined). I will have a look if I can not just remove the whole bit.

PR #4443, opened 2021-04-22
Add Elastic Maps Server controller

pebrc:
Second part of #4179 split up to allow easier review. 

Adds the a new controller to support Elastic Maps Server in the ECK operator. 

Depends on https://github.com/elastic/cloud-on-k8s/pull/4439 so please only review https://github.com/elastic/cloud-on-k8s/commit/829c2383dfdca345a9eed4ad30c5edbcf5419910

I have not added a sample yet because it is quite tricky to get a fully working setup without Ingress. But something like this should work: 

```yaml
kind: Elasticsearch
metadata:
  name: elasticsearch-sample
spec:
  version: 7.12.0
  nodeSets:
    - name: default
      count: 1
      config:
        # This setting could have performance implications for production clusters.
        # See: https://www.elastic.co/guide/en/cloud-on-k8s/master/k8s-virtual-memory.html
        node.store.allow_mmap: false
---
apiVersion: maps.k8s.elastic.co/v1alpha1
kind: ElasticMapsServer
metadata:
  name: ems-sample
spec:
  version: 7.12.0
  count: 1
  http:
    tls:
      selfSignedCertificate:
        disabled: true
    service:
      spec:
        type: LoadBalancer
  elasticsearchRef:
    name: elasticsearch-sample
---
kind: Kibana
metadata:
  name: kibana
spec:
  version: 7.12.0
  count: 1
  http:
    service:
      spec:
        type: LoadBalancer
  config:
    map.emsUrl: http://$LB_IP:8080 # counter intuitive but you are accessing EMS through the load balancer
  elasticsearchRef:
    name: elasticsearch-sample
```

### DIFF
--- a/pkg/controller/maps/controller.go
+++ b/pkg/controller/maps/controller.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"reflect"
 	"sync/atomic"
+	"time"
 
 	emsv1alpha1 "github.com/elastic/cloud-on-k8s/pkg/apis/maps/v1alpha1"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/association"
@@ -20,6 +21,7 @@ import (
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common/deployment"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common/driver"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common/events"
+	"github.com/elastic/cloud-on-k8s/pkg/controller/common/license"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common/operator"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common/reconciler"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common/tracing"
@@ -65,6 +67,7 @@ func newReconciler(mgr manager.Manager, params operator.Parameters) *ReconcileMa
 		Client:         client,
 		recorder:       mgr.GetEventRecorderFor(controllerName),
 		dynamicWatches: watches.NewDynamicWatches(),
+		licenseChecker: license.NewLicenseChecker(client, params.OperatorNamespace),
 		Parameters:     params,
 	}
 }
@@ -120,6 +123,7 @@ type ReconcileMapsServer struct {
 	operator.Parameters
 	recorder       record.EventRecorder
 	dynamicWatches watches.DynamicWatches
+	licenseChecker license.Checker
 	// iteration is the number of times this controller has run its Reconcile method
 	iteration uint64
 }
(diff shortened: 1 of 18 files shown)


pebrc:
Jenkins test this please

pebrc:
Jenkins test this please

pebrc:
Jenkins test this please

pkg/controller/maps/pod.go
```diff
@@ -0,0 +1,105 @@
...
+	"github.com/elastic/cloud-on-k8s/pkg/controller/common/container"
+	"github.com/elastic/cloud-on-k8s/pkg/controller/common/defaults"
+	"github.com/elastic/cloud-on-k8s/pkg/controller/common/volume"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/apimachinery/pkg/util/intstr"
+)
+
+const (
+	HTTPPort           = 8080
```
barkbay, line 19: nit and not a big deal, `8080` seems a bit odd when accessing the service with https, I think it is more common to do that on 8443 when HTTP is secured with TLS. However I don't think it's worth the complexity to dynamically change the port 🤷

pkg/controller/maps/pod.go
```diff
@@ -0,0 +1,105 @@
...
+		Requests: map[corev1.ResourceName]resource.Quantity{
+			corev1.ResourceMemory: DefaultMemoryLimits,
+		},
+		Limits: map[corev1.ResourceName]resource.Quantity{
+			corev1.ResourceMemory: DefaultMemoryLimits,
+		},
+	}
+)
+
+// readinessProbe is the readiness probe for the Kibana container
```
barkbay, line 36:
```suggestion
// readinessProbe is the readiness probe for the maps container
```

barkbay (approved):
I don't have a lot of feedback to give on this PR except it looks very clean and work as expected, nice work 👍 

I did a few tests, including:

* Using custom HTTPS certificates (using the cert-manager) for both the maps server and Elasticsearch (including some updates with some back and forth to check that all the certs are reloaded and the maps server Pods are restarted)
* Using the config reference setting.
* Running the documentation proposed in https://github.com/elastic/cloud-on-k8s/pull/4447 using a Nginx ingress controller as mentioned in [my comment](https://github.com/elastic/cloud-on-k8s/pull/4447/files#r621288921).

Even if loading the tiles in a volume seems a bit challenging everything worked fine.

While scaling horizontally and updating the EMS resource at the same time I was wondering how much the users would like to customize the default deployment strategy:

```yaml
  strategy:
    rollingUpdate:
      maxSurge: 25%
      maxUnavailable: 25%
```

But as you mentioned in https://github.com/elastic/cloud-on-k8s/pull/4439#issue-620110737 this is something we should consider for all our resources.

Also I opened https://github.com/elastic/cloud-on-k8s/issues/4451 because I think the status can be misleading, it is however also out of the scope of this PR.

pkg/controller/maps/controller.go
```diff
@@ -0,0 +1,393 @@
...
+		}
+		return reconcile.Result{}, tracing.CaptureError(ctx, err)
+	}
+
+	if common.IsUnmanaged(&ems) {
+		log.Info("Object is currently not managed by this controller. Skipping reconciliation", "namespace", ems.Namespace, "name", ems.Name)
+		return reconcile.Result{}, nil
+	}
+
+	// check for compatibility with the operator version
```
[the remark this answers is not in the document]
pebrc: I added one in https://github.com/elastic/cloud-on-k8s/pull/4443/commits/782e9f21ab85925d9990bb72625e812c795df310 if you want to take another look

pkg/controller/maps/controller.go
```diff
@@ -0,0 +1,410 @@
...
+	}
+
+	if common.IsUnmanaged(&ems) {
+		log.Info("Object is currently not managed by this controller. Skipping reconciliation", "namespace", ems.Namespace, "name", ems.Name)
+		return reconcile.Result{}, nil
+	}
+
+	enabled, err := r.licenseChecker.EnterpriseFeaturesEnabled()
+	if err != nil {
+		return reconcile.Result{Requeue: true, RequeueAfter: 10 * time.Second}, err
```
barkbay, line 171: I'm wondering if it's worth it to requeue the event after only 10s ? If the license is corrupted it might take some time to fix it and it might generate a lot of logs in the meantime  (and also some unnecessary load on the operator)?
```suggestion
		return reconcile.Result{}, err
```

barkbay (approved):
Still LGTM 👍

PR #4444, opened 2021-04-22
Add E2E tests for Elastic Maps Server

pebrc:
Third part of #4179 split up to allow easier review.

Adds a few basic e2e test for  Elastic Maps Server in the ECK operator. I tried to also refactor some common steps and checks that were more or less identical between Kibana/Enterprise Search and APM Server

Depends on #4439 and https://github.com/elastic/cloud-on-k8s/pull/4443 so please only review dff0eb478b3982fb8bf3915a229f9acdac47aa51

### DIFF
--- a/test/e2e/test/common_checks.go
+++ b/test/e2e/test/common_checks.go
@@ -19,7 +19,7 @@ import (
 // CheckDeployment checks the Deployment resource exists
 func CheckDeployment(subj Subject, k *K8sClient, deploymentName string) Step {
 	return Step{
-		Name: "Elastic Map Server deployment should be created",
+		Name: subj.Kind() + " deployment should be created",
 		Test: Eventually(func() error {
 			var dep appsv1.Deployment
 			err := k.Client.Get(context.Background(), types.NamespacedName{
@@ -40,7 +40,7 @@ func CheckDeployment(subj Subject, k *K8sClient, deploymentName string) Step {
 	}
 }
 
-// CheckPods checks test subjects expected pods are eventually ready.
+// CheckPods checks that the test subject's expected pods are eventually ready.
 func CheckPods(subj Subject, k *K8sClient) Step {
 	// This is a shared test but it is common for Enterprise Search Pods to take some time to be ready, especially
 	// during the initial bootstrap, or during version upgrades. Let's increase the timeout
@@ -86,7 +86,7 @@ func CheckPods(subj Subject, k *K8sClient) Step {
 	}
 }
 
-// CheckServices checks that all Enterprise Search services are created
+// CheckServices checks that all expected services have been created
 func CheckServices(subj Subject, k *K8sClient) Step {
 	return Step{
 		Name: subj.Kind() + " services should be created",
--- a/test/e2e/test/common_checks.go
+++ b/test/e2e/test/common_checks.go
@@ -1,0 +1,131 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License;
+// you may not use this file except in compliance with the Elastic License.
+
+package test
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/elastic/cloud-on-k8s/pkg/controller/common/hash"
+	"github.com/elastic/cloud-on-k8s/pkg/utils/k8s"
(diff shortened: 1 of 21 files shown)


test/e2e/test/common_checks.go
```diff
@@ -0,0 +1,131 @@
...
+					return fmt.Errorf("pod not ready yet")
+				}
+			}
+
+			return nil
+		}, timeout),
+	}
+}
+
+// CheckServices checks that all Enterprise Search services are created
```
thbkrkr, line 89:
```suggestion
// CheckServices checks that all services are created
```

test/e2e/test/subject.go
```diff
@@ -0,0 +1,20 @@
...
+
+package test
+
+import (
+	"k8s.io/apimachinery/pkg/types"
+	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// Subject the main test subject for a test case
+type Subject interface {
```
thbkrkr, line 13: :heart:

test/e2e/test/common_checks.go
```diff
@@ -0,0 +1,131 @@
...
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+// CheckDeployment checks the Deployment resource exists
+func CheckDeployment(subj Subject, k *K8sClient, deploymentName string) Step {
+	return Step{
+		Name: "Elastic Map Server deployment should be created",
```
thbkrkr, line 22:
```suggestion
		Name: subj.Kind() + " Deployment should be created",
```

test/e2e/test/common_checks.go
```diff
@@ -0,0 +1,131 @@
...
+			}
+			if *dep.Spec.Replicas != subj.Count() {
+				return fmt.Errorf("invalid %s replicas count: expected %d, got %d", subj.Kind(), subj.Count(), *dep.Spec.Replicas)
+			}
+			return nil
+		}),
+	}
+}
+
+// CheckPods checks test subjects expected pods are eventually ready.
```
thbkrkr, line 43:
```suggestion
// CheckPods checks test subject expected pods are eventually ready.
```

thbkrkr (approved):
Nice factorization with the introduction of the `Subject` interface.

LGTM

barkbay (commented):
Should we not skip the tests if the stack version is not at least 7.11 ?
Also should we not initialize a valid license with `NewLicenseTestContext` ?

pebrc:
> Should we not skip the tests if the stack version is not at least 7.11 ?
> Also should we not initialize a valid license with NewLicenseTestContext ?

Oh of course yes to both! Will fix.

test/e2e/test/maps/checks.go
```diff
@@ -0,0 +1,150 @@
...
+				}
+				bytes, err := DoRequest(client, b.EMS, "GET", "/status")
+				if err != nil {
+					return err
+				}
+				var status emsStatus
+				if err := json.Unmarshal(bytes, &status); err != nil {
+					return err
+				}
+				// we cannot just test for "available" because in absence of a valid license the state is "critical"
```
barkbay, line 140: nit: Is this statement still valid now that we are running the tests with a license ?

test/e2e/test/maps/checks.go
```diff
@@ -0,0 +1,150 @@
...
+				// however we consider a parseable response to be a sufficient signal for a successful deployment
```
barkbay, line 141: While running the tests I mostly got the `degraded` status because of the tiles which are not installed. I'm wondering if a test should fail for `unavailable` or `critical`, but as I don't know the real meaning of these status I think I'm fine with the current approach.

barkbay (approved):
LGTM 👍

test/e2e/test/maps/checks.go
pebrc: Hm good point. I think you are right we could test for that. You are right in 7.11-7.12 it returns `degraded` if the tile database is missing. Starting with 7.13 it will be `available` even in the absence of the tiles, because maps server will start  to support running with smaller basemaps.

PR #4447, opened 2021-04-23
Add initial documentation for Elastic Maps Server

pebrc:
First draft of a documentation page for Elastic Maps Server. The recipe included needs more work. It cannot be applied without modification and I have not figured out a way to get to a good Quickstart like recipe without compromising security. 

Preview: https://cloud-on-k8s_4447.docs-preview.app.elstc.co/diff


docs/orchestrating-elastic-stack-applications/maps.asciidoc
```diff
@@ -0,0 +1,245 @@
...
+=== Provide your own certificate
+
+If you want to use your own certificate, the required configuration is identical to Elasticsearch. See <<{p}-custom-http-certificate>>.
+
+[id="{p}-maps-http-disable-tls"]
+=== Disable TLS
+
+You can disable the generation of the self-signed certificate and hence disable TLS.
+
+[source,yaml,subs="attributes"]
```
barkbay, line 226: Same here, should we just add a link to https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-tls-certificates.html#k8s-disable-tls ?

docs/orchestrating-elastic-stack-applications/maps.asciidoc
```diff
@@ -0,0 +1,245 @@
...
+== Map data
+The {ems} Docker image contains only a few zoom levels of data. In order to get the map data up to the highest zoom level {ems} needs a basemap file mounted into its container.
+[id="{p}-maps-basemap-download"]
+=== Basemap download
+You have to download the basemap ahead of time on a machine that is not air-gapped and populate a volume that can be mounted into the {ems} Pods.   See also the link:https://www.elastic.co/guide/en/kibana/current/maps-connect-to-ems.html#elastic-maps-server[{ems} documentation.]
+
+The exact procedure of how to get a Kubernetes volume populated with that data is out of scope of this documenation as it depends on your specific Kubernetes setup and choice of volume provisioner. One possible approach that should work for most setups is as follows:
+
+. Download the basemap zip archive using the link  shown in the {ems} UI or extracted from the  `/status` endpoint.
+. Create a PersistentVolumeClaim of sufficient size (> 90G) and a temporary Pod to mount the corresponding volume
```
barkbay, line 62: I guess we don't want to load the data in the EMS Pod because of the `readOnly: true` section. I am wondering how much mandatory the `readOnly` setting is. Do we expect the EMS containers to concurrently write in the volume ? (I'm just wondering if there is a way to avoid this temporary Pod)

config/recipes/maps/README.asciidoc
```diff
@@ -0,0 +1,46 @@
+= Expose Elastic Maps Server and Kibana using a Kubernetes Ingress
+
+This recipe illustrates how to expose an Elastic Maps Server and Kibana instance to the outside world using custom TLS certificates. All communications -- including those between the load balancer and the Elasticsearch/Kibana pods -- are done over secure TLS connections. The specific configuration necessary might differ depending on your Ingress controller.
+
+
+IMPORTANT: This recipe assumes that your have valid TLS certificates for the domain you want to use.
```
barkbay, line 6: nit: small typo, also the certificates must be valid but maybe an other important point here is that they should be trusted by the clients ?
```suggestion
IMPORTANT: This recipe assumes that you have valid TLS certificates issued by a trusted certificate authority for the domain you want to use.
```

config/recipes/maps/README.asciidoc
```diff
@@ -0,0 +1,46 @@
...
+
+
+Obtain the password of the `elastic` user.
+
+[source,sh]
+----
+kubectl get secret hulk-es-elastic-user -o=go-template='{{ .data.elastic | base64decode}}'
+----
+
+Access Kibana in your browser and verify that maps data is coming from your Elastic Maps Server instance.
```
barkbay, line 44: Should we not invite the user to go through the status page  (`https://maps.{{YOUR_DOMAIN}}/status`) first ?

barkbay (commented):
Just left a few nits after a first reading.

docs/orchestrating-elastic-stack-applications/maps.asciidoc
pebrc: I think it is not mandatory at all. Let me see if we can simplify this further.

pebrc: The problem is that we mount on top of the basemap file that comes with the Docker image. So initially there will be no basemap file until we manage to copy it into the volume. However the Pod will boot loop because the container cannot run without a basemap.

barkbay: 👍 , makes sense, thank you for the clarification.

docs/orchestrating-elastic-stack-applications/maps.asciidoc
```diff
@@ -0,0 +1,211 @@
...
+You can <<{p}-customize-pods,customize the {ems} Pod>> using a Pod template.
+
+The following example demonstrates how to create a {ems} deployment which mounts a data volume with the complete basemap.
+
+[source,yaml,subs="attributes"]
+----
+apiVersion: maps.k8s.elastic.co/v1alpha1
+kind: ElasticMapsServer
+metadata:
+  name: ems-sample
```
barkbay, line 127: Maybe keep the same name in the example, when the previous steps have been followed it helps the user to proceed with `kubectl apply -f` :
```suggestion
  name: quickstart
```

docs/orchestrating-elastic-stack-applications/maps.asciidoc
```diff
@@ -0,0 +1,211 @@
...
+spec:
+  version: {version}
```
barkbay, line 129: By default `count` is 0, also if the reference to Elasticsearch is missing the Pod fails to start:

```suggestion
  version: {version}
  elasticsearchRef:
    name: "quickstart"
  count: 1
```

docs/orchestrating-elastic-stack-applications/maps.asciidoc
```diff
@@ -0,0 +1,214 @@
...
+
+* Customize the Pod configuration
+* Customize the product configuration
+* Manage HTTP settings
+
+
+
+[id="{p}-maps-configuration"]
+=== {ems} configuration
+You can add any valid {ems} setting as documented on the link:https://www.elastic.co/guide/en/kibana/current/maps-connect-to-ems.html#elastic-maps-server-configuration[product] page to the `spec.config` section.
```
barkbay, line 164: Should we also document `spec.configRef` ?

pebrc: Yes, good point. Will add it.

config/recipes/maps/03-ingress.yaml
```diff
@@ -0,0 +1,37 @@
+---
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  name: ems-demo
+  annotations:
+    # Disable HTTP traffic between the Ingress and the backend services
```
barkbay, line 7: According to [this documentation](https://cloud.google.com/kubernetes-engine/docs/concepts/ingress-xlb#disabling_http):

> If you want all traffic between the client and the HTTP(S) load balancer to use HTTPS, you can disable HTTP by including the `kubernetes.io/ingress.allow-http` annotation in your Ingress manifest. Set the value of the annotation to "false".

IIUC `kubernetes.io/ingress.allow-http` affects the traffic between the client and the ingress point, not between the ingress and the backends.

pebrc: Oh well spotted. I think we might just drop the annotation all together I think it is GKE specific.

barkbay: If it helps here is what I'm using for my test (`networking.k8s.io/v1beta1` because I'm testing on K8S 1.18):

```yaml
---
apiVersion: networking.k8s.io/v1beta1
kind: Ingress
metadata:
  name: ems-demo
  annotations:
    cert-manager.io/issuer: "letsencrypt-prod"
    kubernetes.io/ingress.class: "nginx"
    nginx.ingress.kubernetes.io/backend-protocol: HTTPS
    nginx.ingress.kubernetes.io/proxy-ssl-verify: "off"
    nginx.ingress.kubernetes.io/ssl-redirect: "true"
spec:
  tls:
    - hosts: ["kibana.{{YOUR_DOMAIN}}"]
      # secret populated with your Kibana certificates
      secretName: kibana-certs
    - hosts: ["maps.{{YOUR_DOMAIN}}"]
      # secret populate with your Elastic Maps Server certificates
      secretName: maps-certs
  rules:
    - host: "maps.{{YOUR_DOMAIN}}"
      http:
        paths:
          - path: "/"
            #pathType: Prefix
            backend:
                    #  service:
              serviceName: quickstart-ems-http
              servicePort: 8080
    - host: "kibana.{{YOUR_DOMAIN}}"
      http:
        paths:
          - path: "/"
            #pathType: Prefix
            backend:
                    #  service:
              serviceName: quickstart-kb-http
              servicePort: 5601
```

I don't know how to avoid the `nginx.ingress.kubernetes.io/proxy-ssl-verify: "off"`

pebrc: I am using a similar Ingress. Do you think we should just commit to Nginx in the recipe and add the necessary annotations? 

> I don't know how to avoid the nginx.ingress.kubernetes.io/proxy-ssl-verify: "off"

I think in theory `nginx.ingress.kubernetes.io/proxy-ssl-secret: secretName` should work but we have different certs for the different backends.

pebrc: This is what I am using basically the same just a `v1` Ingress and I manually issued the let's encrypt certs

```yaml
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: ems-in
  labels:
    app: ems-demo
  annotations:
    kubernetes.io/ingress.class: "nginx"
    nginx.ingress.kubernetes.io/backend-protocol: "HTTPS"
    nginx.ingress.kubernetes.io/proxy-ssl-verify: "off"
spec:
  tls:
    - hosts: ["dev.pebrc.ovh"]
      secretName: dev-pebrc-ovh
    - hosts: ["maps.pebrc.ovh"]
      secretName: maps-pebrc-ovh
  rules:
    - host: "maps.pebrc.ovh"
      http:
        paths:
          - path: "/"
            pathType: Prefix
            backend:
              service:
                name: ems-sample-ems-http
                port:
                  name: https
    - host: "dev.pebrc.ovh"
      http:
        paths:
          - path: "/"
            pathType: Prefix
            backend:
              service:
                name: thor-kb-http
                port:
                  name: https

```

barkbay: >  Do you think we should just commit to Nginx in the recipe and add the necessary annotations?

I'm not an ingress expert but it looks difficult to provide a working example compatible with all the ingress controllers. The Nginx implementation is developed as part of the K8S project and is available in Minikube, I think I would be inclined to provide a working example for Nginx and mention that it must be adapted for other ingress implementations.

> I think in theory `nginx.ingress.kubernetes.io/proxy-ssl-secret: secretName` should work but we have different certs for the different backends.

Yes that's the tricky part.

david-kow (approved):
I've read (but not tested) and it LGTM. One nit though: what is the idea behind our ordering of `Run x on ECK` in the menu? I can't figure out what do we sort it by :)

docs/orchestrating-elastic-stack-applications/managing-compute-resources.asciidoc
```diff
@@ -185,8 +209,9 @@ If `resources` is not defined in the specification of an object, then the operat
 |APM Server |512Mi |512Mi
 |Elasticsearch |2Gi |2Gi
 |Kibana |1Gi |1Gi
-|Beat   |200Mi | 200Mi
-|Elastic Agent | 200Mi|200Mi
+|Beat   |200Mi |200Mi
+|Elastic Agent |200Mi |200Mi
```
thbkrkr, line 213: Need a rebase here as we bumped the memory for the Agent to 300Mi in #4453.

docs/orchestrating-elastic-stack-applications/maps.asciidoc
```diff
@@ -0,0 +1,248 @@
...
+
+[id="{p}-maps-data"]
+== Map data
+
+experimental[]
+
+The {ems} Docker image contains only a few zoom levels of data. In order to get the map data up to the highest zoom level {ems} needs a basemap file mounted into its container.
+[id="{p}-maps-basemap-download"]
+=== Basemap download
+You have to download the basemap ahead of time on a machine that is not air-gapped and populate a volume that can be mounted into the {ems} Pods.  See also the link:https://www.elastic.co/guide/en/kibana/current/maps-connect-to-ems.html#elastic-maps-server[{ems} documentation.]
```
thbkrkr, line 65:
```suggestion
You have to download the basemap ahead of time on a machine that is not air-gapped and populate a volume that can be mounted into the {ems} Pods. See also the link:https://www.elastic.co/guide/en/kibana/current/maps-connect-to-ems.html#elastic-maps-server[{ems} documentation.]
```

barkbay (approved):
LGTM !

config/recipes/maps/README.asciidoc
```diff
@@ -0,0 +1,48 @@
+= Expose Elastic Maps Server and Kibana using a Kubernetes Ingress
+
+This recipe illustrates how to expose an Elastic Maps Server and Kibana instance to the outside world using custom TLS certificates. All communications -- including those between the load balancer and the Elasticsearch/Kibana pods -- are done over secure TLS connections.
```
alaudazzi, line 3:
```suggestion
This recipe illustrates how to expose an Elastic Maps Server and Kibana instance to the outside world using custom TLS certificates. All communications, including those between the load balancer and the Elasticsearch/Kibana pods , occur over secure TLS connections.
```

config/recipes/maps/README.asciidoc
```diff
@@ -0,0 +1,48 @@
...
+
+The Ingress in this example is configured for an Nginx Ingress controller. The specific configuration necessary might differ if you are using another Ingress controller.
```
alaudazzi, line 5:
```suggestion
The Ingress in this example is configured for an Nginx Ingress controller. The required specific configuration might differ if you are using another Ingress controller.
```

config/recipes/maps/README.asciidoc
```diff
@@ -0,0 +1,48 @@
...
+
+== Recipe
+
+Create an Elasticsearch cluster and an Elastic Maps Server instance with customized service definition.
+
+[source,yaml]
+----
+include::01-ems-es.yaml[]
+----
+CAUTION: The Elasticsearch configuration uses `node.store.allow_mmap: false` to avoid configuring memory mapping settings on the underlying host. This could have a significant performance impact on your Elasticsearch clusters and should not be used in production without careful consideration. See https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-virtual-memory.html for more information.
```
alaudazzi, line 19:
```suggestion
CAUTION: The Elasticsearch configuration uses `node.store.allow_mmap: false` to avoid configuring memory mapping settings on the underlying host. This could impact the performance of your Elasticsearch clusters and should not be used in production without careful consideration. See https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-virtual-memory.html for more information.
```

config/recipes/maps/README.asciidoc
```diff
@@ -0,0 +1,48 @@
...
+
+
+Create a Kibana. Replace the placeholder `{{YOUR_DOMAIN}}` in the recipe with your actual domain.
+[source,yaml]
+----
+include::02-kb.yaml[]
+----
+
+Create the Ingress. Again replace the placeholder `{{YOUR_DOMAIN}}` in the recipe with your actual domain. Create the TLS secrets with your certificates.
```
alaudazzi, line 28:
```suggestion
Create the Ingress. Replace the placeholder `{{YOUR_DOMAIN}}` in the recipe with your actual domain. Create the TLS secrets with your certificates.
```

config/recipes/maps/README.asciidoc
```diff
@@ -0,0 +1,48 @@
...
+
+[source,yaml]
+----
+include::03-ingress.yaml[]
+----
+
+
+== Test the ingress
```
alaudazzi, line 36:
```suggestion
== Test the Ingress
```

docs/orchestrating-elastic-stack-applications/maps.asciidoc
```diff
@@ -0,0 +1,248 @@
...
+NOTE: Any {ems} can reference (and thus access) any Elasticsearch instance as long as they are both in namespaces that are watched by the same ECK instance. ECK will copy the required Secret from Elasticsearch to the {ems} namespace. {ems} cannot automatically connect to Elasticsearch (through `elasticsearchRef`) in a namespace managed by a different ECK instance. For more information, see <<{p}-restrict-cross-namespace-associations,Restrict cross-namespace resource associations>>.
+
+The {ems} configuration file is automatically setup by ECK to establish a secure connection to Elasticsearch.
+
+[id="{p}-maps-data"]
+== Map data
+
+experimental[]
+
+The {ems} Docker image contains only a few zoom levels of data. In order to get the map data up to the highest zoom level {ems} needs a basemap file mounted into its container.
```
alaudazzi, line 62:
```suggestion
The {ems} Docker image contains only a few zoom levels of data. To get the map data up to the highest zoom level, {ems} needs a basemap file mounted into its container.
```

docs/orchestrating-elastic-stack-applications/maps.asciidoc
```diff
@@ -0,0 +1,248 @@
...
+[id="{p}-maps-basemap-download"]
+=== Basemap download
+You have to download the basemap ahead of time on a machine that is not air-gapped and populate a volume that can be mounted into the {ems} Pods. See also the link:https://www.elastic.co/guide/en/kibana/current/maps-connect-to-ems.html#elastic-maps-server[{ems} documentation.]
+
+The exact procedure of how to get a Kubernetes volume populated with that data is out of scope of this documenation as it depends on your specific Kubernetes setup and choice of volume provisioner. One possible approach that should work for most setups is as follows:
```
alaudazzi, line 67:
```suggestion
The procedure on how to get a Kubernetes volume populated with that data is outside the scope of this document, as it depends on your specific Kubernetes setup and choice of volume provisioner. This is a possible approach that works for most setups:
```

docs/orchestrating-elastic-stack-applications/maps.asciidoc
```diff
@@ -0,0 +1,248 @@
...
+
+. Download the basemap zip archive using the link shown in the {ems} UI or extracted from the `/status` endpoint.
+. Create a PersistentVolumeClaim of sufficient size (> 90G) and a temporary Pod to mount the corresponding volume
```
alaudazzi, line 70:
```suggestion
. Create a PersistentVolumeClaim of sufficient size (> 90G) and a temporary Pod to mount the corresponding volume.
```

docs/orchestrating-elastic-stack-applications/maps.asciidoc
```diff
@@ -0,0 +1,248 @@
...
+----
+kubectl cp planet.zip ems-data-setup:/usr/share/planet/planet.zip
+----
+. Unzip the archive on the temporary Pod
++
+[source,sh]
+----
+kubectl exec ems-data-setup -- unzip /usr/share/data/planet.zip -d /usr/share/planet
+----
+. Delete the temporary Pod and remount the volume into the {ems} Pods as shown <<{p}-maps-pod-configuration,below>>.
```
alaudazzi, line 117:
```suggestion
. Delete the temporary Pod and remount the volume into the {ems} Pods as described in <<{p}-maps-pod-configuration,Pod configuration>>.
```

docs/orchestrating-elastic-stack-applications/maps.asciidoc
```diff
@@ -0,0 +1,248 @@
...
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: maps-config
+stringData:
+  elastic-maps-server.yml: |-
+    logging.level: debug
+----
+Refer to <<{p}-compute-resources-kibana-and-apm>> for adjusting compute resources for {ems}
```
alaudazzi, line 215:
```suggestion
Refer to <<{p}-compute-resources-kibana-and-apm>> for adjusting compute resources for {ems}.
```

docs/orchestrating-elastic-stack-applications/maps.asciidoc
```diff
@@ -0,0 +1,248 @@
...
+
+[id="{p}-maps-scaling"]
+=== Scale out an {ems} deployment
+
+To deploy more than one instance of maps, all the instances must mount the data volume containing the basemap read only. Once that is the case scaling out is just a matter of increasing the `count` attribute.
```
alaudazzi, line 220:
```suggestion
To deploy more than one instance of maps, all the instances must mount the data volume containing the basemap read only. When this is done, scaling out is just a matter of increasing the `count` attribute.
```

docs/orchestrating-elastic-stack-applications/maps.asciidoc
```diff
@@ -0,0 +1,248 @@
...
+
+[id="{p}-maps-http-disable-tls"]
+=== Disable TLS
+
+You can disable the generation of the self-signed certificate and hence disable TLS. See <<{p}-disable-tls>>.
+
+
+[id="{p}-maps-ingress"]
+==== Ingress and Kibana configuration
+In order to use {ems} from your Kibana instances you need to configure Kibana to fetch maps from your {ems} instance by using the link:https://www.elastic.co/guide/en/kibana/current/maps-connect-to-ems.html#elastic-maps-server-kibana[`map.emsUrl`] configuration key. The value of this setting needs to be the URL where the {ems} instance is reachable from your browser. In order to work correctly the certificates presented by {ems} need to be trusted by the browser and the URL should have the same origin as the URL where your Kibana is hosted to avoid cross origin resource issues.  Please see the link:{eck_github}/tree/{eck_release_branch}/config/recipes/[recipe section] for an example how to set this up using an Ingress resource.
```
alaudazzi, line 248:
```suggestion
To use {ems} from your Kibana instances, you need to configure Kibana to fetch maps from your {ems} instance by using the link:https://www.elastic.co/guide/en/kibana/current/maps-connect-to-ems.html#elastic-maps-server-kibana[`map.emsUrl`] configuration key. The value of this setting needs to be the URL where the {ems} instance is reachable from your browser. The certificates presented by {ems} need to be trusted by the browser, and the URL must have the same origin as the URL where your Kibana is hosted to avoid cross origin resource issues. See the link:{eck_github}/tree/{eck_release_branch}/config/recipes/[recipe section] for an example on how to set this up using an Ingress resource.
```

alaudazzi (approved):
Left some editing suggestions. Otherwise LGTM.

PR #4460, opened 2021-04-30
E2E tests: combine GO_TAGS and E2E_TAGS

pebrc:
Combining allows tests that import generated code to run. An alternative solution would be to remove all imports of production code from the e2e tests. I think I would be in favour of that second solution: it avoids 'pollution' of test results through production code where the code under test is used to test the code under test. But it is a larger refactoring so I would like to master working again in the meantime.


thbkrkr (approved):
A bit tricky but seems reasonable to move forward.

LGTM

PR #4461, opened 2021-04-30
E2E compile requires go-generate

pebrc:
Follow-up to https://github.com/elastic/cloud-on-k8s/pull/4460

When testing locally I missed this new requirement as I had already run generate manually. https://github.com/elastic/cloud-on-k8s/pull/4460 introduces the main build constraints into the e2e compilation. As a consequence code generation needs to run before we can compile the tests now.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -463,7 +463,7 @@ e2e-generate-xml:
 	@ hack/ci/generate-junit-xml-report.sh e2e-tests.json
 
 # Verify e2e tests compile with no errors, don't run them
-e2e-compile:
+e2e-compile: go-generate
 	@go test ./test/e2e/... -run=dryrun -tags='$(E2E_TAGS)' $(TEST_OPTS) > /dev/null
 
 # Run e2e tests locally (not as a k8s job), with a custom http dialer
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -463,7 +463,7 @@ e2e-generate-xml:
 	@ hack/ci/generate-junit-xml-report.sh e2e-tests.json
 
 # Verify e2e tests compile with no errors, don't run them
-e2e-compile:
+e2e-compile: go-generate
```
thbkrkr, line 466: I think we may have the same issue with `e2e-local`.

```
> GO_TAGS=release make e2e-local
# github.com/elastic/cloud-on-k8s/pkg/controller/common/license
pkg/controller/common/license/check.go:41:22: undefined: publicKeyBytes
make: *** [Makefile:473 : e2e-local] Erreur 2
```

Although if in most cases we run `e2e-local` with `GO_TAGS` empty, which incidentally, I just saw is broken.

```
> GO_TAGS='' make e2e-local
package run is not in GOROOT (/home/thb/dev/tools/go/src/run)
make: *** [Makefile:473 : e2e-local] Erreur 1
```

We need to quote `$(GO_TAGS)` like you did for `'$(E2E_TAGS)'`?

pebrc: Excellent points!

PR #4462, opened 2021-04-30
Update how to reset namespace node selector

thbkrkr:
* Use `kubectl` instead of `oc`.
* Not related but add `--overwrite` flag, useful with `e2e-local` when
test namespaces are not deleted but just cleaned before two runs.

Relates to https://github.com/elastic/cloud-on-k8s/issues/4128#issuecomment-830053470.


barkbay (approved):
lgtm

PR #4464, opened 2021-05-01
E2E: fix nil pointer in license test initialisation

pebrc:
Fix an issue created by the recent license secret handling refactoring where the difference between step declaration time and test execution time in combination with a nil initialised license test context lead to a nil pointer.

* Always initialise test context first
* refactor reused code to simple helper functions to avoid pitfall of declaring but not calling the step function

### DIFF
--- a/test/e2e/es/license_test.go
+++ b/test/e2e/es/license_test.go
@@ -24,10 +24,11 @@ func TestEnterpriseLicenseSingle(t *testing.T) {
 	if test.Ctx().TestLicense == "" {
 		t.SkipNow()
 	}
-	k := test.NewK8sClientOrFatal()
 
 	licenseBytes, err := ioutil.ReadFile(test.Ctx().TestLicense)
 	require.NoError(t, err)
+
+	k := test.NewK8sClientOrFatal()
 
 	// create a single node cluster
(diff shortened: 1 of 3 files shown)


test/e2e/es/license_test.go
```diff
@@ -89,21 +90,17 @@ func TestEnterpriseTrialLicense(t *testing.T) {
 	licenseBytes, err := ioutil.ReadFile(test.Ctx().TestLicense)
 	require.NoError(t, err)
 
+	k := test.NewK8sClientOrFatal()
+
 	esBuilder := elasticsearch.NewBuilder("test-es-trial-license").
 		WithESMasterDataNodes(1, elasticsearch.DefaultResources)
 
-	var licenseTestContext elasticsearch.LicenseTestContext
+	licenseTestContext := elasticsearch.NewLicenseTestContext(k, esBuilder.Elasticsearch)
```
barkbay, line 98: nit:  I think we can avoid `k` as it is only used in this statement.
```suggestion
		licenseTestContext := elasticsearch.NewLicenseTestContext(test.NewK8sClientOrFatal(), esBuilder.Elasticsearch)
```

barkbay (approved):
LGTM

PR #4465, opened 2021-05-03
E2E: run generate before all relevant e2e targets

pebrc:
Fix https://devops-ci.elastic.co/view/cloud-on-k8s/job/cloud-on-k8s-e2e-tests-kind-k8s-versions/416/

This introduces a dependency on `go-generate` to the `e2e-docker*` targets as well as the `e2e-run` target. While this will run generate more than once in some cases it will on the other hand ensure that the expected generated code is present even if the targets are called independently.


thbkrkr (approved):
LGTM

PR #4471, opened 2021-05-04
Switch EKS CI to eu-west-3 to avoid quota issues

pebrc:
We are seeing VPC quota issues in `eu-central-1`. Let's try `eu-west-3` which seems less crowded.


thbkrkr (approved):
:four_leaf_clover:

PR #4473, opened 2021-05-04
Handle volumes with larger capacities than the one claimed

### LINKED ISSUE #4469: Autoscaling: handling unforeseeable volume capacities
This issue is a sub-issue of https://github.com/elastic/cloud-on-k8s/issues/4459 to focus on/discuss the case where the capacity of a persistent volume is larger than the one specified in the claim.

As an example here is a claim of `1Gi` bounded to a volume with an actual physical capacity of `368Gi`:

```yaml
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  annotations:
    pv.kubernetes.io/bind-completed: "yes"
    pv.kubernetes.io/bound-by-controller: "yes"
  creationTimestamp: "2021-04-29T11:44:01Z"
  finalizers:
  - kubernetes.io/pvc-protection
  labels:
    common.k8s.elastic.co/type: elasticsearch
    elasticsearch.k8s.elastic.co/cluster-name: storage-sample
    elasticsearch.k8s.elastic.co/statefulset-name: storage-sample-es-data
  name: elasticsearch-data-storage-sample-es-data-0
  namespace: demo
  ownerReferences:
  - apiVersion: elasticsearch.k8s.elastic.co/v1
    kind: Elasticsearch
    name: storage-sample
    uid: b0b9b12d-092d-4c9d-a73e-8c5039bfb2d9
spec:
  accessModes:
  - ReadWriteOnce
  resources:
    requests:
      storage: 1Gi # Claim is 1Gi
  storageClassName: e2e-default
  volumeMode: Filesystem
  volumeName: local-pv-6df90d02
status:
  accessModes:
  - ReadWriteOnce
  capacity:
    storage: 368Gi # Storage capacity is actually 1Gi
  phase: Bound
```

Having a capacity larger than the request may lead to the following situations:

1. Elasticsearch reports the total, observed, capacity as the required capacity for a tier. If the actual capacity is higher than the claim it can lead to some cascading scale up events, up to the limit specified by the user. It can also exceed to limit specified by the user in which case some not pertinent `HorizontalScalingLimitReached` events are generated.
2. If the actual capacity of a volume is greater than the claim, then the nodes may hold more data than the maximum one specified in the autoscaling specification. It may lead to overloaded nodes. For example, assuming the following autoscaling policy:
```json
{
    "name": "data",
    "roles": ["data", "ingest", "transform"],
    "resources": {
        "nodeCount": { "min": 2, "max": 5 },
        "memory": { "min": "2Gi", "max": "6Gi" },
        "storage": { "min": "1Gi",  "max": "3Gi" }
    }
}
```
Say that the claims of 1Gi have been bound to volumes of 1Ti of data each, then chances are that the 2Gi of memory are not enough to handle that amount of data.

Unforeseeable storage capacity makes it hard to scale vertically. It is especially true as long as there is no memory requirement in the autoscaling Elasticsearch API and as long as the operator attempts to infer memory requirement from storage (see https://github.com/elastic/cloud-on-k8s/issues/4076).

### Documentation update

It should be clearly stated in the documentation that scaling vertically data nodes assumes that the storage provider provisions physical volume with a predictable and even capacity across all the volumes managed by an autoscaling policy.

### Autoscaling controller update

If the operator detects that the capacity of a volume is greater than the one specified in the claim then:
  * Memory is computed according to the capacity of the volume, as reported in the PVC status, up to the memory limit specified by the user.
  * Print a log, update the status and emit a K8S warning.

barkbay:
Fixes #4469

* Warn the user if the physical capacity of the volumes is greater than the claims.
* Prevent unnecessary nodes creation and spurious `HorizontalScalingLimitReached` events.

### DIFF
--- a/pkg/controller/autoscaling/elasticsearch/autoscaler/horizontal.go
+++ b/pkg/controller/autoscaling/elasticsearch/autoscaler/horizontal.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/elastic/cloud-on-k8s/pkg/controller/autoscaling/elasticsearch/resources"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/autoscaling/elasticsearch/status"
+	"github.com/elastic/cloud-on-k8s/pkg/controller/elasticsearch/client"
 	corev1 "k8s.io/api/core/v1"
 )
 
@@ -26,13 +27,12 @@ func (ctx *Context) scaleHorizontally(
 	// Scale horizontally to match memory requirements
 	if !totalRequiredCapacity.Memory.IsZero() {
 		nodeMemory := nodeCapacity.GetRequest(corev1.ResourceMemory)
-		nodesToAdd = ctx.getNodesToAdd(nodeMemory.Value(), totalRequiredCapacity.Memory.Value(), ctx.AutoscalingSpec.NodeCountRange.Min, ctx.AutoscalingSpec.NodeCountRange.Max, string(corev1.ResourceMemory))
+		nodesToAdd = ctx.getNodesToAdd(nodeMemory.Value(), totalRequiredCapacity.Memory.Value(), string(corev1.ResourceMemory))
 	}
 
 	// Scale horizontally to match storage requirements
 	if !totalRequiredCapacity.Storage.IsZero() {
-		nodeStorage := nodeCapacity.GetRequest(corev1.ResourceStorage)
-		nodesToAdd = max(nodesToAdd, ctx.getNodesToAdd(nodeStorage.Value(), totalRequiredCapacity.Storage.Value(), ctx.AutoscalingSpec.NodeCountRange.Min, ctx.AutoscalingSpec.NodeCountRange.Max, string(corev1.ResourceStorage)))
+		nodesToAdd = max(nodesToAdd, ctx.getNodesToAddForStorage(nodeCapacity, totalRequiredCapacity.Storage))
 	}
 
 	totalNodes := nodesToAdd + ctx.AutoscalingSpec.NodeCountRange.Min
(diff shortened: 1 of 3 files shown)


pkg/controller/autoscaling/elasticsearch/autoscaler/horizontal.go
```diff
@@ -49,20 +49,89 @@ func (ctx *Context) scaleHorizontally(
...
+
+	// reuse the actual number of nodes from the status
+	totalExpected := currentResources.NodeSetNodeCount.TotalNodeCount()
+	// we still want to downscale if user required it.
+	if totalExpected > ctx.AutoscalingSpec.NodeCountRange.Max {
+		totalExpected = ctx.AutoscalingSpec.NodeCountRange.Max
+	}
+	if totalExpected > ctx.AutoscalingSpec.NodeCountRange.Min {
+		return totalExpected - ctx.AutoscalingSpec.NodeCountRange.Min
+	}
```
pebrc, line 115: I am struggling to understand this logic here. I suppose the intend is to ensure that we downscale if the current capacity exceeds the specified maximum but I don't understand how that is achieved: 

Assuming min 1, max 3, total expected 2 would we not return 1 here which means adding another node, why?
Or min 1, max 3, total expected 4, return 2 i.e adding 2 nodes? I am almost certain I am missing something here but maybe we need a comment to clarify?

barkbay: You're right, it is not clean and misleading.
I'm working on a refactoring.

barkbay:
I'm working on a refactoring. I'll raise an improved PR later.
Sorry for the noise.

PR #4478, opened 2021-05-06
Update versions for 1.6.0 release

idanmo:


thbkrkr (commented):
:rocket:

pebrc (approved):
LGTM I think we could also update https://github.com/elastic/cloud-on-k8s/blob/6884c66259832b63a5a55f69cc2ee9ca3be3da14/hack/operatorhub/config.yaml but we can also do that in a separate PR

PR #4480, opened 2021-05-06
E2E add tags to resilience e2e job

pebrc:
The refactoring in https://github.com/elastic/cloud-on-k8s/pull/4444 requires that build constraints are specified on invoking `go run` to include/exclude the correct generated code.

### DIFF
--- a/test/e2e/run.sh
+++ b/test/e2e/run.sh
@@ -28,7 +28,7 @@ sleep 1
 }
 
 run_chaos() {
-  go run test/e2e/cmd/main.go chaos "$@"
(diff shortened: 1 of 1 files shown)


thbkrkr (approved):

thbkrkr:
I think we should backport this in `1.6` as it is a job executed after each build.
https://github.com/elastic/cloud-on-k8s/blob/89f3bc28d74198d1c981fa20ffb8a3261332f228/.ci/pipelines/build.Jenkinsfile#L130

PR #4486, opened 2021-05-10
Add Elastic Maps Server to the telemetry

david-kow:


pebrc (approved):
LGTM, thanks for remembering that we need to add this!

PR #4489, opened 2021-05-12
Split manifest generation to produce both v1/v1beta1 CRDs

pebrc:
This PR moves all CRDs to `apiVersion: v1` but keeps `v1beta1` around for older Kubernetes versions that we still want to support (all the way back to 1.12 and OCP 3.11)

Upgrades from `v1beta1` to `v1` will be a bit tricky which is why I have marked this PR as `breaking`:
* `all-in-one.yaml` users will have to replace the CRDs and not use `kubectl apply` because a meaningful merge of `v1beta1` to `v1` is not possible. There is however another gotcha with the webhook for which a replace is an illegal operation as it would remove an already populated `clusterIP`: `failed to replace object: Service "elastic-operator-webhook" is invalid: spec.clusterIPs[0]: Invalid value: []string(nil): primary clusterIP can not be unset`. A workaround is to just reapply the manifest. I therefore added a separate manifest with just the CRDs as a download. 
* Helm: a similar situation: users have to first upgrade the CRDs with the `--force` option to replace with `v1` they should be able to follow this with a regular `helm upgrade` of the main chart.
* OLM: in my testing it seems that OLM does the "right thing" and does correctly upgrade from `v1beta1` to `v1`

## Docs preview
https://cloud-on-k8s_4489.docs-preview.app.elstc.co/diff

## Notes on the implementation:
* this PR relies on controller-tools 0.6.0-beta which brings an important bugfix for embedded k8s runtime objects metadata 
* introduced global values in the Helm chart to allow access from the dependent CRD chart, this is a potential breaking change but the old values were marked as internal before as well so I guess it is reasonable to expect users to not rely on them
* The size of the CRDs with schemas and multiple versions is significant (~ 1.6M) We can discuss whether it would make sense to drop the descriptions. So far I was able to apply the manifests with the description as the individual CRDs are still below the max limit.

## Test scenarios:

- [x]  OLM: build custom catalog image and upgrade to latest from a previously released version
- [x]  Helm: v1beta1 to v1: install current chart and upgrade from this PR 
- [x]  Helm: verify future v1 to v1 upgrade: manually install v1 manifests and [adopt them into a 1.6.0 release](https://www.elastic.co/guide/en/cloud-on-k8s/1.4/k8s-install-helm.html#k8s-migrate-to-helm) and try a regular `helm upgrade` without force
- [x] YAML: install operator 1.6.0 and upgrade to this PR 
- [x] YAML: run our [release upgrade test](https://github.com/elastic/cloud-on-k8s/tree/master/hack/upgrade-test-harness) through multiple operator upgrades and finally upgrade to this PR


pebrc:
run full pr build

pebrc:
Marking this ready for review with the caveat that we have a dependency on https://github.com/elastic/cloud-on-k8s/pull/4491

pebrc:
I am going to change the manifests to always separate CRDs and operator manifests into separate files, thus reverting to draft.

pebrc:
The plan to have just three manifests: `crds-v1`, `crds-v1beta1` and `operator` did not work out because we also have version specific differences in the webhook manifests, so we have four manifests now: `crds`, `crds-legacy`, `operator` and `operator-legacy`.

config/crds/v1/patches/ent-patches.yaml
```diff
@@ -0,0 +1,8 @@
+
```
barkbay, line 1: nit:  not from your PR, `ent-patches.yaml`, `ems-patches.yaml` and `apm-kibana-patches.yaml` have the same content, maybe we can have a single `remove-pod-template.patch.yaml`

config/crds/v1/patches/kustomization.yaml
```diff
@@ -0,0 +1,73 @@
+bases:
+  - ../bases
+
+patchesJson6902:
+  # patches that apply to all CRDS
```
barkbay, line 5: nit: can we remove the following lines ?

docs/operating-eck/upgrading-eck.asciidoc
```diff
@@ -25,7 +25,52 @@ CAUTION: The upgrade process results in an update to all the existing managed re
 
 IMPORTANT: If you are running Openshift 3.11 or a Kubernetes version older than 1.13, and want to upgrade from beta, jump to the <<{p}-ga-openshift, upgrade instructions for older Kubernetes releases>>.
 
-Follow the instructions in <<{p}-deploy-eck>> to upgrade ECK to the latest version. This will update the ECK installation to the latest binary and update the CRDs and other ECK resources in the cluster. If you are upgrading from the beta version, ensure that your Elasticsearch, Kibana, and APM Server manifests are updated to use the `v1` API version instead of `v1beta1` after the upgrade.
+This release moves the link:https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/[CustomResourceDefinitions] (CRD) used by ECK to the v1 version. If you are upgrading from a previous version of ECK the new version of the CRDs have to replace the existing CRDs. If removing the current ECK installation is not an option because you have production workloads that must not be deleted, then the following approach is recommended:
```
barkbay, line 28: Should it be `Release 1.7.0 moves` to have a constant version ?

barkbay (commented):
Impressive work 👍 
I only left a few nitpicks, I need some additional time to process and test the PR.

deploy/eck-operator/values.yaml
```diff
@@ -199,13 +199,13 @@ podMonitor:
   podMetricsEndpointConfig: {}
   # honorTimestamps: true
 
-# Internal use only
-internal:
+# Globals meant for internal use only
+global:
```
barkbay, line 203: Should we also update the following test files in `hack/manifest-gen/testdata/` ?
* [`no_defaults.yaml`](https://github.com/elastic/cloud-on-k8s/blob/5362e7317d240b6313d7cec32d2dd8f3743f36bf/hack/manifest-gen/testdata/no_defaults.yaml#L179)
* [`kube113.yaml`](https://github.com/elastic/cloud-on-k8s/blob/5362e7317d240b6313d7cec32d2dd8f3743f36bf/hack/manifest-gen/testdata/kube113.yaml#L160)
* [`kube116.yaml`](https://github.com/elastic/cloud-on-k8s/blob/5362e7317d240b6313d7cec32d2dd8f3743f36bf/hack/manifest-gen/testdata/kube116.yaml#L160)

If yes I think I would expect the tests to fail, but I think the purpose is only to validate the schemas using `garethr/kubeval`.

config/crds/v1/patches/elasticsearch-patches.yaml
```diff
@@ -0,0 +1,16 @@
+# Move validation of the v1 version to the top-level, and remove validation from the v1beta1 version.
+# This is the behaviour intended from using controller-tools crd:trivialVersions=true flag,
+# which currently does not work as expected.
+# See https://github.com/kubernetes-sigs/controller-tools/issues/349.
+# TODO: remove once fixed in controller-tools
```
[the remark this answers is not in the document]
pebrc: I think we were losing validation when using `trivialVersions` with more than one version: https://github.com/elastic/cloud-on-k8s/issues/2479

Makefile
```diff
@@ -142,11 +178,11 @@ unit-xml: clean
 	ECK_TEST_LOG_LEVEL=$(LOG_VERBOSITY) gotestsum --junitfile unit-tests.xml -- -cover ./pkg/... ./cmd/... $(TEST_OPTS)
 
 integration: GO_TAGS += integration
-integration: clean generate-crds
+integration: clean generate-crds-v1beta1
```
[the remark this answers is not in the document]
pebrc: Afaik I know we cannot until we upgrade kubebuilder. Because we are relying on the kubebuilder integration test support and kubebuilder 2.3.1 comes with a 1.15 control plane. I can add another comment to make this clear

deploy/eck-operator/values.yaml
pebrc: Good catch. Indeed we don't have a good way of verifying that the test input is what we expect. I don't have a good idea right now.

docs/operating-eck/upgrading-eck.asciidoc
```diff
@@ -25,7 +25,52 @@ CAUTION: The upgrade process results in an update to all the existing managed re
...
+.If you are running a version of Kubernetes before 1.16: use the legacy manifests
+----
+kubectl replace -f https://download.elastic.co/downloads/eck/{eck_version}/crds-legacy.yaml
+kubectl apply -f https://download.elastic.co/downloads/eck/{eck_version}/operator-legacy.yaml
+----
+
+[source,shell,subs="attributes,callouts"]
+.If you are using Helm: force upgrade the CRD chart to move to the v1 CRDs
+----
+helm upgrade eck-operator elastic/eck-operator-crds -n elastic-system --force
```
barkbay, line 62: Maybe `elastic-operator` instead of `eck-operator` to be consistent with the [Helm documentation](https://www.elastic.co/guide/en/cloud-on-k8s/master/k8s-install-helm.html#k8s-install-helm-global) ?
```suggestion
helm upgrade elastic-operator elastic/eck-operator-crds -n elastic-system --force
```

docs/operating-eck/upgrading-eck.asciidoc
```diff
@@ -25,7 +25,52 @@ CAUTION: The upgrade process results in an update to all the existing managed re
...
+----
+
+Then upgrade the main chart as usual:
+[source,shell,subs="attributes,callouts"]
+----
+helm upgrade eck-operator elastic/eck-operator -n elastic-system
```
barkbay, line 68: Same as above:
```suggestion
helm upgrade elastic-operator elastic/eck-operator -n elastic-system
```

Makefile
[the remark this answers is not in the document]
barkbay: TIL, and my bad: you already mentioned it [here](https://github.com/elastic/cloud-on-k8s/pull/4489/files#diff-e5ec31770015c79fe55d4b0deded05848e0d54c6f1ddc74cfc02e9c32854e735R31).

barkbay (commented):
nit: there is a reference to the all in one [here](https://github.com/elastic/cloud-on-k8s/blob/5362e7317d240b6313d7cec32d2dd8f3743f36bf/.ci/pipelines/build.Jenkinsfile#L65).

pebrc:
> nit: there is a reference to the all in one here.

Good catch, there is another one in the operatorhub tooling. Will fix!

barkbay (approved):
LGTM ! 🚀

docs/advanced-topics/service-meshes.asciidoc
```diff
@@ -45,7 +45,8 @@ kubectl label namespace elastic-system istio-injection=enabled
 +
 [source,sh,subs="attributes"]
 ----
-kubectl apply -f https://download.elastic.co/downloads/eck/{eck_version}/all-in-one.yaml
+kubectl create -f https://download.elastic.co/downloads/eck/{eck_version}/create.yaml
```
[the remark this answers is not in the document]
pebrc: oops

pebrc:
run full pr build

deploy/eck-operator/values.yaml
```diff
@@ -199,13 +199,13 @@ podMonitor:
   podMetricsEndpointConfig: {}
   # honorTimestamps: true
 
-# Internal use only
-internal:
+# Globals meant for internal use only
```
thbkrkr, line 202: Out of curiosity, why did you rename `internal` to `global`?

docs/operating-eck/installing-eck.asciidoc
```diff
@@ -17,9 +17,9 @@ CAUTION: Deleting CRDs will trigger deletion of all custom resources ({eck_resou
 
 
 [id="{p}-install-all-in-one"]
```
thbkrkr, line 19: Nit: rename the anchor?
```suggestion
[id="{p}-install-yaml-manifests"]
```

docs/operating-eck/upgrading-eck.asciidoc
```diff
@@ -25,7 +25,52 @@ CAUTION: The upgrade process results in an update to all the existing managed re
 
 IMPORTANT: If you are running Openshift 3.11 or a Kubernetes version older than 1.13, and want to upgrade from beta, jump to the <<{p}-ga-openshift, upgrade instructions for older Kubernetes releases>>.
 
-Follow the instructions in <<{p}-deploy-eck>> to upgrade ECK to the latest version. This will update the ECK installation to the latest binary and update the CRDs and other ECK resources in the cluster. If you are upgrading from the beta version, ensure that your Elasticsearch, Kibana, and APM Server manifests are updated to use the `v1` API version instead of `v1beta1` after the upgrade.
+Release 1.7.0 moves the link:https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/[CustomResourceDefinitions] (CRD) used by ECK to the v1 version. If you are upgrading from a previous version of ECK the new version of the CRDs have to replace the existing CRDs. If removing the current ECK installation is not an option because you have production workloads that must not be deleted, then the following approach is recommended:
+
+[source,shell,subs="attributes,callouts"]
+.If you are installing using the YAML manifests: replace existing CRDs
```
thbkrkr, line 31:
```suggestion
.If you are installing using the YAML manifests, replace existing CRDs:
```

docs/operating-eck/upgrading-eck.asciidoc
```diff
@@ -25,7 +25,52 @@ CAUTION: The upgrade process results in an update to all the existing managed re
...
+----
+kubectl replace -f https://download.elastic.co/downloads/eck/{eck_version}/crds.yaml
+----
+
+[NOTE]
+================================
+If you skipped a release in which new CRDs where introduced, you will see an error message similar to `Error from server (NotFound): error when replacing "config/crds.yaml": customresourcedefinitions.apiextensions.k8s.io ... not found`.  To add the missing CRDs run
```
thbkrkr, line 38:
```suggestion
If you skipped a release in which new CRDs where introduced, you will see an error message similar to `Error from server (NotFound): error when replacing "config/crds.yaml": customresourcedefinitions.apiextensions.k8s.io ... not found`. To add the missing CRDs run:
```

docs/operating-eck/upgrading-eck.asciidoc
```diff
@@ -25,7 +25,52 @@ CAUTION: The upgrade process results in an update to all the existing managed re
...
+
+================================
+Then upgrade the remaining objects with the operator manifest:
+[source,shell,subs="attributes,callouts"]
+----
+kubectl apply -f https://download.elastic.co/downloads/eck/{eck_version}/operator.yaml
+----
+
+[source,shell,subs="attributes,callouts"]
+.If you are running a version of Kubernetes before 1.16: use the legacy manifests
```
thbkrkr, line 53: I understand `before` as strictly less than.
```suggestion
.If you are running a version of Kubernetes less than or equal to 1.16, use the legacy manifests:
```

config/crds/v1beta1/patches/agent-patches.yaml
```diff
@@ -25,3 +25,6 @@
   path: /spec/validation/openAPIV3Schema/properties/spec/properties/deployment/properties/strategy/properties/rollingUpdate/properties/maxUnavailable/x-kubernetes-int-or-string
 - op: remove
   path: /spec/validation/openAPIV3Schema/properties/spec/properties/daemonSet/properties/updateStrategy/properties/rollingUpdate/properties/maxUnavailable/x-kubernetes-int-or-string
+# we need to generate x-kubernetes-preserve-unknown-fields for v1 CRDs but they break v1beta so we have to remove them again here
+- op: remove
+  path: /spec/validation/openAPIV3Schema/properties/spec/properties/config/x-kubernetes-preserve-unknown-fields
```
thbkrkr, line 30: Normal that we don't have the same count of `x-kubernetes-preserve-unknown-fields` and `// +kubebuilder:pruning:PreserveUnknownFields`?

Agent and Beat do not get:
```
- op: remove
  path: /spec/validation/openAPIV3Schema/properties/spec/properties/daemonSet/properties/podTemplate/x-kubernetes-preserve-unknown-fields
- op: remove
  path: /spec/validation/openAPIV3Schema/properties/spec/properties/deployment/properties/podTemplate/x-kubernetes-preserve-unknown-fields
```

thbkrkr (approved):
Great work :clap: 
I left some minor and cosmetic changes.

deploy/eck-operator/values.yaml
pebrc: Because `global` is a built-in thingy in Helm and I can share the values with the dependent chart. Otherwise we would have to set the `internal.k8sVersion` value again for the crd chart.

pkg/apis/elasticsearch/v1/elasticsearch_types.go
```diff
@@ -219,6 +219,7 @@ type NodeSet struct {
 	Name string `json:"name"`
 
 	// Config holds the Elasticsearch configuration.
+	// +kubebuilder:pruning:PreserveUnknownFields
```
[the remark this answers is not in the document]
pebrc: Which ones do you mean exactly?

hack/upgrade-test-harness/README.md
```diff
@@ -35,6 +36,7 @@ Adding a new release
 --------------------
 
 - Create a directory under `testdata` following the existing naming convention.
-- Add the `all-in-one.yaml` as `install.yaml` to the release directory.
+- Add the `crds.yaml` as `crds.yaml` to the release directory.
+- Add the `operator.yaml` as `install.yaml` to the release directory.
```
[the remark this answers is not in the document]
pebrc: I kept it as `install.yaml` to be backwards compatible with previous releases.

docs/operating-eck/installing-eck.asciidoc
pebrc: I guess I was hesitant to change this one because I thought there might be external references to that. But I guess we only link to the quickstart from the the download page (which we need to update for the next release)

docs/operating-eck/upgrading-eck.asciidoc
pebrc: But you are fine with using the new manifests with 1.16

config/crds/v1beta1/patches/agent-patches.yaml
pebrc: That's because we remove the whole `podTemplate` section for the `v1beta1` version on the top of the file.

pkg/apis/elasticsearch/v1/elasticsearch_types.go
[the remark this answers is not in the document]
thbkrkr:
```diff
diff --git a/pkg/apis/elasticsearch/v1beta1/elasticsearch_types.go b/pkg/apis/elasticsearch/v1beta1/elasticsearch_types.go
index 3dd03c28..3230e1a2 100644
--- a/pkg/apis/elasticsearch/v1beta1/elasticsearch_types.go
+++ b/pkg/apis/elasticsearch/v1beta1/elasticsearch_types.go
@@ -61,6 +61,7 @@ type NodeSet struct {
        Name string `json:"name"`
 
        // Config holds the Elasticsearch configuration.
+       // +kubebuilder:pruning:PreserveUnknownFields
        Config *commonv1beta1.Config `json:"config,omitempty"`
 
        // Count of Elasticsearch nodes to deploy.
@@ -69,12 +70,14 @@ type NodeSet struct {
 
        // PodTemplate provides customisation options (labels, annotations, affinity rules, resource requests, and so on) for the Pods belonging to this NodeSet.
        // +kubebuilder:validation:Optional
+       // +kubebuilder:pruning:PreserveUnknownFields
        PodTemplate corev1.PodTemplateSpec `json:"podTemplate,omitempty"`
 
        // VolumeClaimTemplates is a list of persistent volume claims to be used by each Pod in this NodeSet.
        // Every claim in this list must have a matching volumeMount in one of the containers defined in the PodTemplate.
        // Items defined here take precedence over any default claims added by the operator with the same name.
        // +kubebuilder:validation:Optional
+       // +kubebuilder:pruning:PreserveUnknownFields
        VolumeClaimTemplates []corev1.PersistentVolumeClaim `json:"volumeClaimTemplates,omitempty"`
 }
```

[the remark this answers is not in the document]
pebrc: Ah I understand now. I don't think we do, because we don't use structural schema in v1beta1 and the default was `true` in v1beta1 for backwards compatibility afaik

config/crds/v1beta1/patches/agent-patches.yaml
thbkrkr: Of course.
That's not important but just for my understanding, we could do the same the following patches where we also remove the whole `podTemplate`?
```
v1beta1/patches/apm-kibana-patches.yaml:  path: /spec/validation/openAPIV3Schema/properties/spec/properties/podTemplate/properties
v1beta1/patches/apm-kibana-patches.yaml:  path: /spec/validation/openAPIV3Schema/properties/spec/properties/podTemplate/x-kubernetes-preserve-unknown-fields

v1beta1/patches/elasticsearch-patches.yaml:  path: /spec/validation/openAPIV3Schema/properties/spec/properties/nodeSets/items/properties/podTemplate/properties
v1beta1/patches/elasticsearch-patches.yaml:  path: /spec/validation/openAPIV3Schema/properties/spec/properties/nodeSets/items/properties/podTemplate/x-kubernetes-preserve-unknown-fields

v1beta1/patches/ems-patches.yaml:  path: /spec/validation/openAPIV3Schema/properties/spec/properties/podTemplate/properties
v1beta1/patches/ems-patches.yaml:  path: /spec/validation/openAPIV3Schema/properties/spec/properties/podTemplate/x-kubernetes-preserve-unknown-fields

v1beta1/patches/ent-patches.yaml:  path: /spec/validation/openAPIV3Schema/properties/spec/properties/podTemplate/properties
v1beta1/patches/ent-patches.yaml:  path: /spec/validation/openAPIV3Schema/properties/spec/properties/podTemplate/x-kubernetes-preserve-unknown-fields
```

pkg/apis/elasticsearch/v1/elasticsearch_types.go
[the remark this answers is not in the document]
thbkrkr: :+1: 

I admit I just:
```

File: docs/operating-eck/upgrading-eck.asciidoc [hunk 18]
```diff
@@ -25,7 +25,52 @@ CAUTION: The upgrade process results in an update to all the existing managed re
 
 IMPORTANT: If you are running Openshift 3.11 or a Kubernetes version older than 1.13, and want to upgrade from beta, jump to the <<{p}-ga-openshift, upgrade instructions for older Kubernetes releases>>.
 
-Follow the instructions in <<{p}-deploy-eck>> to upgrade ECK to the latest version. This will update the ECK installation to the latest binary and update the CRDs and other ECK resources in the cluster. If you are upgrading from the beta version, ensure that your Elasticsearch, Kibana, and APM Server manifests are updated to use the `v1` API version instead of `v1beta1` after the upgrade.
+Release 1.7.0 moves the link:https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/[CustomResourceDefinitions] (CRD) used by ECK to the v1 version. If you are upgrading from a previous version of ECK the new version of the CRDs have to replace the existing CRDs. If removing the current ECK installation is not an option because you have production workloads that must not be deleted, then the following approach is recommended:
```
alaudazzi, line 28:
```suggestion
Release 1.7.0 moves the link:https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/[CustomResourceDefinitions] (CRD) used by ECK to the v1 version. If you upgrade from a previous version of ECK, the new version of the CRDs replace the existing CRDs. If you cannot remove the current ECK installation because you have production workloads that must not be deleted, the following approach is recommended:
```

docs/operating-eck/webhook.asciidoc
```diff
@@ -26,7 +26,7 @@ Validating webhooks are defined using a `ValidatingWebhookConfiguration` object
 [id="{p}-{page_id}-defaults"]
 == Defaults provided by ECK
 
-When installed through the default `all-in-one.yaml` manifest, ECK will be installed with a `ValidatingWebhookConfiguration` configured as follows:
+When installed through the default `operator.yaml` manifest, ECK will be installed with a `ValidatingWebhookConfiguration` configured as follows:
```
alaudazzi, line 29:
```suggestion
When using the default `operator.yaml` manifest, ECK is installed with a `ValidatingWebhookConfiguration` configured as follows:
```

alaudazzi (approved):
A few minor suggestions for the text, otherwise LGTM.

config/crds/v1beta1/patches/agent-patches.yaml
pebrc: The way we remove the `podTemplate` differs a little bit between Agent and Beats and the other CRDs. Agent and Beats remove the whole `podTemplate` tree, while the other CRDs only remove the `properties` from the `podTemplate` section. This makes the other tweaking necessary. This is nothing this PR introduced but has been in place since the introduction of Beats and Agent. I kept it that way so that the legacy CRDs stay as close to the versions we currently ship as possible.

PR #4490, opened 2021-05-14
Docs: clarify volume resize statement in quickstart

pebrc:
We had a statement in the quickstart that seems to negate the existence of the volume claim resize feature. I tried to make it more explicit and link to the relevant doc in the Elasticsearch section.


barkbay (approved):
lgtm

PR #4523, opened 2021-05-26
E2E enable Kibana, APM Server, Enterprise Search, Beats and Agent tests on ARM

pebrc:
As of 7.13 we have arm64 Docker images for those Elastic Stack apps which means we can start running more tests. Elastic Maps Server does not have an arm64 image yet.


barkbay (approved):
LGTM

Issue #4525, opened 2021-05-27
Misconfiguration of volume claim can lead to data loss


pebrc:
https://github.com/elastic/cloud-on-k8s/pull/3806 changed the defaulting behaviour so that if a a user specifies a volume claim template we no longer create the default 1G data volume. This happens because we assume that users want to replace the default data volume with their own volume. The motivation for this change was to allow users to mount multiple data volumes and give them full flexibility in naming the volumes/mounts to their liking. 

This use case is no longer a very strong one as multiple data paths are being deprecated in Elasticsearch. The change introduced however a potential for misconfiguration if a user does not follow our documentation and chooses a different name for the claim than `elasticsearch-data`. In that case a PVC is created but the volume is not mounted and the Elasticsearch process is writing into the container ie. non-persistent storage.

Mitigation: 
* validate that the volume claim template follows the naming convention or if not that a matching volume mount exists
* consider reverting https://github.com/elastic/cloud-on-k8s/pull/3806

pebrc:
After some consideration rolling back #3806 is not a good option as users might rely on it even though we did not document this "feature". Having added additional validation in #4526 I am closing this issue in the hope that validation will suffice to alert users of this condition.

PR #4526, opened 2021-05-27
Add validation for declared volume claim templates

pebrc:
Partially addresses https://github.com/elastic/cloud-on-k8s/issues/4525

Adds a validation to check that declared volume claim templates are actually mounted in any of the containers. This validation does not apply if the user configures the documented default data volume `elasticsearch-data` for which ECK manages the volume mount. 

There is still a possibility for misconfiguration. As we are currently supporting custom volume mounts and custom `data.path` settings in Elasticsearch it is quite hard to fully validate the configuration. With the proposed validation it would still be possible for a user to configure a custom volume claim template, mount it in a side car and have no persistent Elasticsearch data volume. This could be addressed in two ways:
* also parse the Elasticsearch configuration during validation and make sure each `data.path` has a volume mount
* always configure a default data volume as we had it pre-1.4.0

### DIFF
--- a/pkg/controller/elasticsearch/validation/validations.go
+++ b/pkg/controller/elasticsearch/validation/validations.go
@@ -34,7 +34,7 @@ const (
 	parseStoredVersionErrMsg = "Cannot parse current Elasticsearch version. String format must be {major}.{minor}.{patch}[-{label}]"
 	parseVersionErrMsg       = "Cannot parse Elasticsearch version. String format must be {major}.{minor}.{patch}[-{label}]"
 	pvcImmutableErrMsg       = "volume claim templates can only have their storage requests increased, if the storage class allows volume expansion. Any other change is forbidden"
-	pvcNotMounted            = "volume claim declared but volume not mounted in any container"
+	pvcNotMountedErrMsg      = "volume claim declared but volume not mounted in any container. Note that the Elasticsearch data volume should be named 'elasticsearch-data'"
 	unsupportedConfigErrMsg  = "Configuration setting is reserved for internal use. User-configured use is unsupported"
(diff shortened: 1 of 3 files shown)


pkg/controller/elasticsearch/validation/validations.go
```diff
@@ -34,6 +34,7 @@ const (
 	parseStoredVersionErrMsg = "Cannot parse current Elasticsearch version. String format must be {major}.{minor}.{patch}[-{label}]"
 	parseVersionErrMsg       = "Cannot parse Elasticsearch version. String format must be {major}.{minor}.{patch}[-{label}]"
 	pvcImmutableErrMsg       = "volume claim templates can only have their storage requests increased, if the storage class allows volume expansion. Any other change is forbidden"
+	pvcNotMounted            = "volume claim declared but volume not mounted in any container"
```
sebgl, line 37: Should we make it more explicit to cover the usual user mistake?
`volume claim declared but volume not mounted in any container. Please note the Elasticsearch data volume should be named "elasticsearch-data"`

sebgl, line 37: `pvcNotMountedErrMsg` for consistency?

sebgl (approved):
LGTM. Left a few comments for minor improvements.

One concern is that existing users with misconfigured volumes would suddenly have their cluster reconciliation error-out when they upgrade the operator to a version with that validation. I think that's OK (better point out the problem and stop reconciling rather than silently ignoring it?).

> With the proposed validation it would still be possible for a user to configure a custom volume claim template, mount it in a side car and have no persistent Elasticsearch data volume. This could be addressed in two ways

I think I'm fine with not addressing this edge case.

pebrc:
Jenkins test this please

aliciascott:
Happened again here https://github.com/elastic/sdh-elasticsearch/issues/4741

PR #4528, opened 2021-05-27
Elasticsearch monitoring with Metricbeat and Filebeat as sidecars

### LINKED ISSUE #4183: Support log and metrics delivery with restricted RBAC permissions
We added support for Beats in ECK 1.2 and have published a YAML recipe that illustrates how to set up stack monitoring with Metricbeat and log delivery with Filebeat. 

I order to run these recipes the user needs to be able to give Metricbeat access to a service account that allows queriying the k8s API and for Filebeat they need to be able to deploy a (privileged) DeamonSet.

We should investigate a simple way that allows users to extract log files and stack monitoring metrics without the need for elevated permission or the ability to deploy DaemonSets.

One approach that comes to mind is to revive the sidecar approach we discussed originally here https://github.com/elastic/cloud-on-k8s/issues/2415 for Metricbeat.

For Filebeat @david-kow has experimented with FIFOs to harvest output from the monitored applications std out 

<details>
<summary> FIFO example</summary>

```yaml
 working with stdin and pipe
apiVersion: v1
kind: Pod
metadata:
  name: test-pod8
spec:
  initContainers:
  - name: fifo
    image: docker.elastic.co/beats/filebeat:7.6.0
    command: [ 'bash', '-c', 'mkfifo -m 0666 /filebeat/sink' ]
    volumeMounts: 
    - name: sink
      mountPath: /filebeat
  containers:
  - name: app
    image: ubuntu:trusty
    command: [ 'bash', '-c', 'while [ true ]; do echo "$(date)"; sleep 1; done | tee $FILEBEAT_SINK' ]
    env:
    - name: FILEBEAT_SINK
      value: /filebeat/sink
    volumeMounts:
    - name: sink
      mountPath: /filebeat
  - name: filebeat
    image: docker.elastic.co/beats/filebeat:7.6.0
    command: [ 'bash', '-c', 'filebeat -e -c /etc/filebeat.yml <$FILEBEAT_SINK' ]
    env:
    - name: FILEBEAT_SINK
      value: /filebeat/sink
    volumeMounts:
    - name: sink
      mountPath: /filebeat
    - name: config
      mountPath: /etc/filebeat.yml
      readOnly: true
      subPath: filebeat.yml
  volumes:
  - name: sink
    emptyDir:
      medium: Memory
  - name: config
    configMap:
      defaultMode: 0744
      name: filebeat-config
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: filebeat-config
data:
  filebeat.yml: |-
    filebeat.inputs:
    - type: stdin
    processors:
      - add_cloud_metadata:
      - add_host_metadata:
    output.console:
      pretty: true
```

</details>

The goal of this feature would not only be to allow the setup in restricted k8s environments but also to make it easy to setup: 

```
kind: Elasticsearch
metadata:
  name: quickstart
spec:
  version: 7.11.0
  monitoring:
    logs:
      elasticsearchRef:
        namespace: monitoring-ns
        name: my-logging-cluster
    metrics:
      elasticsearchRef:
        namespace: monitoring-ns
        name: my-metrics-cluster
  nodeSets:
  - name: default
    count: 1
```
Of course there are a few assumptions and restrictions:
* we would not allow users to customize any aspect of the Metricbeat or Filebeat configuration
* it needs to be seen whether we can allow users to override the Pod template for these side car containers
* technically a separate stack monitoring sidecar for each Elasticsearch node is wasteful, but it is simpler to manage 
* with access to the k8s APIs we would not be able to enrich the data with k8s specific metadata but maybe we compensate with additional operator generated metadata in the Filebeat and Metricbeat configs

thbkrkr:
Adds a new `monitoring` field to the Elasticsearch resource to configure one or two different Elasticsearch references to set up stack monitoring with Metricbeat and log delivery with Filebeat. The referenced ES are used to send the data collected by the beats.

```yaml
kind: Elasticsearch
spec:
  monitoring:
    metrics:
      elasticsearchRefs: 
        - name: m1
          namespace: m1
    logs:
      elasticsearchRefs:
        - name: m2
          namespace: m2
```

This is implemented with a multiple association of ES type (1 es <-> [1|2] es).

New `stackmon` packages contains functions to:
- set the `xpack.monitoring.*` settings in the ES config
- enable stack logging for the ES container using the environment variable `ES_LOG_STYLE=file`
- add volumes to the ES pod to mount the ES CA and the beats configurations in the beats sidecar containers
- inject Metricsbeat and Filebeat as sidecar container in the ES pod than that of ES container with volumeMounts for the ES CA
- reconcile the Metricsbeat and Filebeat configurations in ConfigMaps

The beats configuration is built from a base configuration merged with the output config section which defines es info to send data. For Metricsbeat, the base config is a template to inject the es info to collect data.

A hash of the two beats config files is added in a pod label to ensure pods are rotated when es user passwords are rotated.

<details>
<summary>YAML example for testing</summary>

```yaml
apiVersion: v1
kind: Namespace
metadata:
  name: a
---
apiVersion: v1
kind: Namespace
metadata:
  name: b
---
apiVersion: v1
kind: Namespace
metadata:
  name: c
---
#######################################################################
# Monitored Elasticsearch
#######################################################################
kind: Elasticsearch
metadata:
  name: test
  namespace: a
spec:
  version: 7.14.0-SNAPSHOT
  # --------------------- #
  monitoring:
    metrics:
      elasticsearchRefs: 
        - name: m1
          namespace: b
    logs:
      elasticsearchRefs:
        - name: m1
          namespace: b
        # - name: m2
        #   namespace: c
  # --------------------- #
  nodeSets:
  - name: master
    count: 2
    config:
      node.store.allow_mmap: false
---
kind: Kibana
metadata:
  name: test
  namespace: a
spec:
  version: 7.14.0-SNAPSHOT
  count: 1
  elasticsearchRef:
    name: test
    namespace: a
---
#######################################################################
# Monitoring clusters
#######################################################################
kind: Elasticsearch
metadata:
  name: m1
  namespace: b
spec:
  version: 7.14.0-SNAPSHOT
  nodeSets:
  - name: master
    count: 2
    config:
      node.store.allow_mmap: false
---
kind: Kibana
metadata:
  name: m1
  namespace: b
spec:
  version: 7.14.0-SNAPSHOT
  count: 1
  elasticsearchRef:
    name: m1
    namespace: b
---
kind: Elasticsearch
metadata:
  name: m2
  namespace: c
spec:
  version: 7.14.0-SNAPSHOT
  nodeSets:
  - name: master
    count: 2
    config:
      node.store.allow_mmap: false
---
kind: Kibana
metadata:
  name: m2
  namespace: c
spec:
  version: 7.14.0-SNAPSHOT
  count: 1
  elasticsearchRef:
    name: m2
    namespace: c
```
</details>

### Limitations
- Minimum supported Stack version is 7.14.0 (to benefit from ES_LOG_STYLE=file)
- ~Custom Elastisearch image that don't follow the Elastic scheme (`$registry/elasticsearch/elasticsearch:$version`) are not supported~  To use custom beat images, you have to override the podTemplate. 
- The monitored Elasticsearch is not deployed until monitoring clusters are not ready yet or the association are not configured
- `monitoring.[metrics|logs].elasticsearchRefs` accepts only one Elasticsearch reference. It's a slice to future proof the API for Elastic agent.

Relates to #4183.


pkg/apis/elasticsearch/v1/elasticsearch_types.go
```diff
@@ -72,8 +77,264 @@ type ElasticsearchSpec struct {
...
+
+func (mma *MetricsMonitoringAssociation) AssociationRef() commonv1.ObjectSelector {
+	return commonv1.ObjectSelector{
+		Name:      mma.ref.Name,
+		Namespace: mma.ref.Namespace,
+	}
+}
+
+func (mma *MetricsMonitoringAssociation) AssociationConf() *commonv1.AssociationConf {
+	if mma.assocConfs == nil {
```
pebrc, line 154: nil pointer deref if metrics are not defined due to a pointer to Elasticsearch being embedded in the struct

pkg/controller/elasticsearch/stackmon/volume.go
```diff
@@ -0,0 +1,99 @@
...
+	assocConf := es.GetMonitoringMetricsAssociation().AssociationConf()
+	return volume.NewSecretVolumeWithMountPath(
+		assocConf.CASecretName,
+		MonitoringMetricsTargetEsCaCertVolumeName,
+		MonitoringMetricsTargetEsCaCertMountPath,
+	)
+}
+
+func monitoringLogsTargetCaCertSecretVolume(es esv1.Elasticsearch) volume.SecretVolume {
+	assocConf := es.GetMonitoringMetricsAssociation().AssociationConf()
```
pebrc, line 93:
```suggestion
	assocConf := es.GetMonitoringLogsAssociation().AssociationConf()
```

pkg/apis/common/v1/association.go
```diff
@@ -92,6 +92,10 @@ func (asm AssociationStatusMap) AllEstablished() bool {
 const (
 	ElasticsearchConfigAnnotationNameBase = "association.k8s.elastic.co/es-conf"
 	ElasticsearchAssociationType          = "elasticsearch"
+	MetricbeatConfigAnnotationNameBase    = "association.k8s.elastic.co/mb-conf"
+	MetricbeatAssociationType             = "metricbeat"
+	FilebeatConfigAnnotationNameBase      = "association.k8s.elastic.co/fb-conf"
+	FilebeatAssociationType               = "filebeat"
```
pebrc, line 98: Do we need two separate controllers? I would imagine that the Beats users should be compatible for both?

pebrc (commented):
Looks promising! I wonder however if we can model this with just one additional association type es->es instead of two separate ones for metricbeat and filebeat?

thbkrkr:
> I wonder however if we can model this with just one additional association type es->es instead of two separate ones for metricbeat and filebeat?

It was my first idea that I too quickly abandoned because I did not yet understand very well how associations work. I started again and you're right it's a little simpler and cleaner. I will update the PR.

pkg/apis/common/v1/association.go
thbkrkr: Yes, only one is enough.

pkg/apis/elasticsearch/v1/elasticsearch_types.go
```diff
@@ -72,8 +77,264 @@ type ElasticsearchSpec struct {
...
+
+	nsNameHash := sha256.New224()
+	// concat with dot to avoid collisions, as namespace can't contain dots
+	_, _ = nsNameHash.Write([]byte(fmt.Sprintf("%s.%s", lma.ref.Namespace, lma.ref.Name)))
+	// base32 to encode and limit the length, as using Sprintf with "%x" encodes with base16 which happens to
+	// give too long output
+	// no padding to avoid illegal '=' character in the annotation name
+	hash := base32.StdEncoding.WithPadding(base32.NoPadding).EncodeToString(nsNameHash.Sum(nil))
+
+	return commonv1.FormatNameWithID(
```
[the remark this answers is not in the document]
thbkrkr: The hash-based annotation helps avoid creating an illegal annotation name that is too long.

pkg/apis/elasticsearch/v1/elasticsearch_types.go
thbkrkr: I think it can't happen because when it's call, an association has always been created, but it doesn't hurt to add a nil check.

Note that we have the same code for the Agent.
https://github.com/elastic/cloud-on-k8s/blob/c0b259b8a29d64371a339539846bd42e2d58ddc6/pkg/apis/agent/v1alpha1/agent_types.go#L257-L258

pebrc: Yes but for agent the association is only created when needed. Just try to run it an example with log shipping but without metrics shipping. 

The problem is that you return an empty struct in `GetMonitoringMetricsAssociation` and `GetMonitoringLogsAssocation` but then access the embedded Elasticsearch pointer unconditionally here. 

```diff
--- a/pkg/apis/elasticsearch/v1/elasticsearch_types.go
+++ b/pkg/apis/elasticsearch/v1/elasticsearch_types.go
@@ -288,7 +288,7 @@ func (es *Elasticsearch) GetMonitoringMetricsAssociation() commonv1.Association
                        ref:           ref.WithDefaultNamespace(es.Namespace).NamespacedName(),
                }
        }
-       return &MetricsMonitoringAssociation{}
+       return &MetricsMonitoringAssociation{Elasticsearch: es}
 }
 
 func (es *Elasticsearch) GetMonitoringLogsAssociation() commonv1.Association {
@@ -299,7 +299,7 @@ func (es *Elasticsearch) GetMonitoringLogsAssociation() commonv1.Association {
                        ref:           ref.WithDefaultNamespace(es.Namespace).NamespacedName(),
                }
        }
-       return &LogsMonitoringAssociation{}
+       return &LogsMonitoringAssociation{Elasticsearch: es}
 }
```

barkbay (commented):
Out of curiosity I did a test with version 6.8 of the stack, Metricbeat containers do not start:

```
2021-06-03T08:59:11.655Z        INFO    instance/beat.go:280    Setup Beat: metricbeat; Version: 6.8.16
2021-06-03T08:59:11.657Z        INFO    elasticsearch/client.go:164     Elasticsearch url: https://monitoring-es-http.demo.svc:9200
2021-06-03T08:59:11.657Z        INFO    [publisher]     pipeline/module.go:110  Beat name: monitored-es-master-0
2021-06-03T08:59:11.657Z        INFO    instance/beat.go:359    metricbeat stopped.
2021-06-03T08:59:11.657Z        ERROR   instance/beat.go:906    Exiting: 1 error: The elasticsearch module with xpack.enabled: true must have metricsets: [ccr cluster_stats index index_recovery index_summary ml_job node_stats shard]
Exiting: 1 error: The elasticsearch module with xpack.enabled: true must have metricsets: [ccr cluster_stats index index_recovery index_summary ml_job node_stats shard]
```

I think it's because `enrich` is not supported before `7.5` _(not sure about the version)_. It seems that we need to generate the Metricbeat configuration according to the stack version and plan to add an e2e test to validate version dependant behaviour _(can be added in a subsequent PR)_.

Also, reading this [compatibility matrix](https://www.elastic.co/support/matrix#matrix_compatibility) I'm wondering if we should prevent association between a monitored ES/Beat 6.x and a monitoring ES 7.x ?

pkg/controller/elasticsearch/stackmon/config.go
```diff
@@ -0,0 +1,129 @@
...
+output.elasticsearch:
+  hosts: ['${ES_TARGET_URL}']
+  username: ${ES_TARGET_USERNAME}
+  password: ${ES_TARGET_PASSWORD}
+  ssl.certificate_authorities: ["/mnt/es/monitoring/metrics/target/ca.crt"]`
+
+	// FilebeatConfig is a static configuration for Filebeat to collect Elasticsearch logs
+	// Warning: environment variables and CA cert paths are hard-coded for simplicity.
+	FilebeatConfig = `filebeat.modules:
+# https://www.elastic.co/guide/en/beats/filebeat/current/filebeat-module-elasticsearch.html
```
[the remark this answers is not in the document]
thbkrkr: Good idea.

pkg/apis/elasticsearch/v1/elasticsearch_types.go
```diff
@@ -364,10 +536,12 @@ type ElasticsearchStatus struct {
 	Version string                          `json:"version,omitempty"`
 	Health  ElasticsearchHealth             `json:"health,omitempty"`
 	Phase   ElasticsearchOrchestrationPhase `json:"phase,omitempty"`
+
+	ElasticsearchAssociationsStatus commonv1.AssociationStatusMap `json:"esAssociationStatus,omitempty"`
 }
 
 type ZenDiscoveryStatus struct {
-	MinimumMasterNodes int `json:"minimumMasterNodes,omitempty"`
+	MinimueeasterNodes int `json:"minimueeasterNodes,omitempty"`
```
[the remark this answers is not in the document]
thbkrkr: :see_no_evil:

cmd/manager/main.go
```diff
@@ -677,6 +677,7 @@ func registerControllers(mgr manager.Manager, params operator.Parameters, access
 		{name: "BEAT-KB", registerFunc: associationctl.AddBeatKibana},
 		{name: "AGENT-ES", registerFunc: associationctl.AddAgentES},
 		{name: "EMS-ES", registerFunc: associationctl.AddMapsES},
+		{name: "ES-ES", registerFunc: associationctl.AddEsEs},
```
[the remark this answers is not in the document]
thbkrkr: I like your naming proposal.

pkg/apis/elasticsearch/v1/elasticsearch_types.go
```diff
@@ -72,8 +77,175 @@ type ElasticsearchSpec struct {
 	// +kubebuilder:validation:Optional
 	// +kubebuilder:validation:Enum=DeleteOnScaledownOnly;DeleteOnScaledownAndClusterDeletion
 	VolumeClaimDeletePolicy VolumeClaimDeletePolicy `json:"volumeClaimDeletePolicy,omitempty"`
+
+	// Monitoring enables you to extract log and stack monitoring metrics of this Elasticsearch cluster.
+	// Metricbeat and Filebeat are deployed in the same pod as sidecar and each one send data to one or two different
```
david-kow, line 82:
```suggestion
	// Metricbeat and Filebeat are deployed in the same Pod as sidecars and each one send data to one or two different
```

pkg/apis/elasticsearch/v1/elasticsearch_types.go
```diff
@@ -72,8 +77,175 @@ type ElasticsearchSpec struct {
 	// +kubebuilder:validation:Optional
 	// +kubebuilder:validation:Enum=DeleteOnScaledownOnly;DeleteOnScaledownAndClusterDeletion
 	VolumeClaimDeletePolicy VolumeClaimDeletePolicy `json:"volumeClaimDeletePolicy,omitempty"`
+
+	// Monitoring enables you to extract log and stack monitoring metrics of this Elasticsearch cluster.
```
david-kow, line 81: Should we link to the Stack Monitoring feature docs? This was it's clear that it's a Stack feature, not ECK feature.

pkg/apis/elasticsearch/v1/elasticsearch_types.go
```diff
@@ -72,8 +77,175 @@ type ElasticsearchSpec struct {
...
+
+// Monitoring holds Elasticsearch references to send logs and metrics of this Elasticsearch cluster.
+type Monitoring struct {
+	Metrics MetricsMonitoring `json:"metrics,omitempty"`
+	Logs    LogsMonitoring    `json:"logs,omitempty"`
+}
+
+type MetricsMonitoring struct {
+	// ElasticsearchRef is a reference to a monitoring Elasticsearch cluster running in the same Kubernetes cluster
+	// dedicated to receiving stack monitoring metrics.
```
david-kow, line 96:
```suggestion
	// dedicated to receiving Stack Monitoring metrics.
```

pkg/apis/elasticsearch/v1/elasticsearch_types.go
```diff
@@ -72,8 +77,175 @@ type ElasticsearchSpec struct {
...
+func (es *Elasticsearch) SetAssociationStatusMap(typ commonv1.AssociationType, status commonv1.AssociationStatusMap) error {
+	if typ != commonv1.ElasticsearchAssociationType {
+		return fmt.Errorf("association type %s not known", typ)
+	}
+
+	es.Status.ElasticsearchAssociationsStatus = status
+	return nil
+}
+
+/*func (es *Elasticsearch) AssociationID() string {
```
david-kow, line 245: To remove?

pkg/controller/association/controller/es_es.go
```diff
@@ -0,0 +1,59 @@
...
+const (
+	// EsEsAssociationLabelName marks resources created by this controller for easier retrieval.
+	EsEsAssociationLabelName = "esesassociation.k8s.elastic.co/name"
+	// EsEsAssociationLabelNamespace marks resources created by this controller for easier retrieval.
+	EsEsAssociationLabelNamespace = "esesassociation.k8s.elastic.co/namespace"
+	// EsEsAssociationLabelType marks the type of association.
+	EsEsAssociationLabelType = "esesassociation.k8s.elastic.co/type"
+
+	// BeatBuiltinRole is the name of the built-in role for the Metricbeat/Filebeat system user.
+	BeatBuiltinRole = "superuser" // FIXME: create a dedicated role?
```
david-kow, line 29: +1 on a dedicated role.

pkg/controller/elasticsearch/driver/driver.go
```diff
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/elastic/cloud-on-k8s/pkg/controller/elasticsearch/stackmon"
```
david-kow, line 13: nit: import order

pkg/controller/elasticsearch/nodespec/podspec.go
```diff
@@ -8,6 +8,7 @@ import (
 	"crypto/sha256"
 	"fmt"
 
+	"github.com/elastic/cloud-on-k8s/pkg/controller/elasticsearch/stackmon"
```
david-kow, line 11: nit: import ordering

david-kow (commented):
Couple small comments, I also played with it a bit and it works nicely.

thbkrkr:
Remains to be done:
- [ ] dedicated role
- [ ] min supported version (>= 7.14)

pkg/controller/common/defaults/pod_template.go
```diff
@@ -179,6 +183,15 @@ func (b *PodTemplateBuilder) WithTerminationGracePeriod(period int64) *PodTempla
 	return b
 }
 
+// WithContainer appends the given containers to the list of containers belonging to the pod.
```
david-kow, line 186:
```suggestion
// WithContainers appends the given containers to the list of containers belonging to the pod.
```

pkg/controller/common/defaults/pod_template.go
```diff
@@ -179,6 +183,15 @@ func (b *PodTemplateBuilder) WithTerminationGracePeriod(period int64) *PodTempla
...
+// It also ensures that the base container defaulter still points to the container in the list because append()
+// create a new slice.
```
david-kow, line 188:
```suggestion
// creates a new slice.
```

thbkrkr:
If we want to have a dedicated role for each beat, things get a little tough because we can have the two associations that point to the same monitoring Elasticsearch cluster. In this case, we would like to have two dedicated users in this unique ES cluster. I think our `AssociationInfo` type is not ready for that. We need a way to identify the `EsMonAssociation` for the metrics from the one for the logs in the association controller in order to differentiate the `ESUserRole` and the `UserSecretSuffix` for each association.

Since the permissions for the two beats do not differ a lot and doesn't seem to be excessive, I went with one single dedicated user.

pkg/controller/elasticsearch/nodespec/resources.go
```diff
@@ -54,7 +55,8 @@ func BuildExpectedResources(
 		if nodeSpec.Config != nil {
 			userCfg = *nodeSpec.Config
 		}
-		cfg, err := settings.NewMergedESConfig(es.Name, ver, ipFamily, es.Spec.HTTP, userCfg)
+		isMonitoringMetrics := stackmon.IsMonitoringMetricsDefined(es)
+		cfg, err := settings.NewMergedESConfig(es.Name, ver, ipFamily, es.Spec.HTTP, userCfg, isMonitoringMetrics)
```
pebrc, line 59: Feels a bit odd how the "knowledge" about stack monitoring is spread here between this place and the `NewMergedESConfig` I wonder if would be cleaner to just pass in an additional (sometimes empty) config object?

```go
cfg, err := settings.NewMergedESConfig(es.Name, ver, ipFamily, es.Spec.HTTP, userCfg, stackmon.MonitoringConfig())
```

pebrc (commented):
I did a quick pass. I will try find some time to do a more in-depth review and run some tests later this week if no one else does it in the meantime.

pkg/apis/common/v1/association.go
```diff
@@ -79,7 +79,7 @@ func (asm AssociationStatusMap) Single() (AssociationStatus, error) {
 	return result, nil
 }
 
-// AllEstablished returns true iff all Associations have AssociationEstablished status, false otherwise.
+// AllEstablished returns true if all Associations have AssociationEstablished status, false otherwise.
```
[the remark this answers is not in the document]
thbkrkr: TIL `iff`.

pkg/controller/common/volume/configmap.go
```diff
@@ -30,15 +30,20 @@ type ConfigMapVolume struct {
...
 func (cm ConfigMapVolume) VolumeMount() corev1.VolumeMount {
-	return corev1.VolumeMount{
+	vm := corev1.VolumeMount{
 		Name:      cm.name,
 		MountPath: cm.mountPath,
 		ReadOnly:  true,
 	}
+	if cm.subPath != "" {
+		vm.SubPath = cm.subPath
+	}
```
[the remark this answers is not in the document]
thbkrkr: My bad, it's a leftover that I haven't reverted.

pkg/controller/association/reconciler.go
```diff
@@ -377,7 +377,11 @@ func (r *Reconciler) updateStatus(ctx context.Context, associated commonv1.Assoc
 	span, _ := apm.StartSpan(ctx, "update_association_status", tracing.SpanTypeApp)
 	defer span.End()
 
-	oldStatus := associated.AssociationStatusMap(r.AssociationType)
+	oldStatus := commonv1.AssociationStatusMap{}
+	// Copy from the original map to the target map
+	for key, value := range associated.AssociationStatusMap(r.AssociationType) {
+		oldStatus[key] = value
+	}
```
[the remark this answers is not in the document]
thbkrkr: No good reason. I will revert. (It dates from when I used 2 different types of associations.)

pkg/controller/elasticsearch/elasticsearch_controller.go
```diff
@@ -191,6 +192,11 @@ func (r *ReconcileElasticsearch) Reconcile(ctx context.Context, request reconcil
 		return reconcile.Result{}, tracing.CaptureError(ctx, err)
 	}
 
+	// Requeue if associations are configured but not yet defined
+	if !association.AreConfiguredIfSet(es.GetAssociations(), r.recorder) {
+		return reconcile.Result{}, nil
+	}
```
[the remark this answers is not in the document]
thbkrkr: It panics in `monitoringMetricsTargetCaCertSecretVolume` when we try to get the `CASecretName` because the `assocConf` is nil.

```go
func monitoringMetricsTargetCaCertSecretVolume(es esv1.Elasticsearch) volume.SecretVolume {
	assocConf := es.GetMonitoringMetricsAssociation().AssociationConf()
	return volume.NewSecretVolumeWithMountPath(
		assocConf.CASecretName, // <----- HERE
		MonitoringMetricsTargetEsCaCertVolumeName,
		MonitoringMetricsTargetEsCaCertMountPath,
	)
}
```

```
E0616 17:41:04.241541  184548 runtime.go:78] Observed a panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory address or nil pointer dereference)                    
goroutine 801 [running]:                                                                                                                                                                                 
k8s.io/apimachinery/pkg/util/runtime.logPanic(0x1a0a280, 0x2a539d0)                                                                                                                                      
        /home/thb/dev/pkg/mod/k8s.io/apimachinery@v0.20.2/pkg/util/runtime/runtime.go:74
k8s.io/apimachinery/pkg/util/runtime.HandleCrash(0x0, 0x0, 0x0)                                                                                                                                          
        /home/thb/dev/pkg/mod/k8s.io/apimachinery@v0.20.2/pkg/util/runtime/runtime.go:48
panic(0x1a0a280, 0x2a539d0)                                                                                                                                                                              
        /home/thb/dev/tools/go/src/runtime/panic.go:965
pkg/controller/elasticsearch/stackmon.monitoringMetricsTargetCaCertSecretVolume(...)
        pkg/controller/elasticsearch/stackmon/volume.go:79
pkg/controller/elasticsearch/stackmon.monitoringVolumes(...)      
        pkg/controller/elasticsearch/stackmon/volume.go:37
pkg/controller/elasticsearch/stackmon.WithMonitoring(...)
        pkg/controller/elasticsearch/stackmon/container.go:54
pkg/controller/elasticsearch/nodespec.BuildPodTemplateSpec(...)
        pkg/controller/elasticsearch/nodespec/podspec.go:94
pkg/controller/elasticsearch/nodespec.BuildStatefulSet(...)      
        pkg/controller/elasticsearch/nodespec/statefulset.go:77
pkg/controller/elasticsearch/nodespec.BuildExpectedResources(...)                        
        pkg/controller/elasticsearch/nodespec/resources.go:64
pkg/controller/elasticsearch/driver.(*defaultDriver).reconcileNodeSpecs(...)                                                                                                                                                                                                  
        pkg/controller/elasticsearch/driver/nodes.go:81
pkg/controller/elasticsearch/driver.(*defaultDriver).Reconcile(...)                                                                
        pkg/controller/elasticsearch/driver/driver.go:261
pkg/controller/elasticsearch.(*ReconcileElasticsearch).internalReconcile(...)
        pkg/controller/elasticsearch/elasticsearch_controller.go:292
```

We do the same thing here:

```
pkg/controller/beat/controller.go
167:    if !association.AreConfiguredIfSet(beat.GetAssociations(), r.recorder) {
168-            return results

pkg/controller/agent/controller.go
156:    if !association.AreConfiguredIfSet(agent.GetAssociations(), r.recorder) {
157-            return results

pkg/controller/apmserver/controller.go
210:    if !association.AreConfiguredIfSet(as.GetAssociations(), r.recorder) {
211-            return reconcile.Result{}, nil
```

thbkrkr:
I will add unit tests.

pkg/controller/elasticsearch/elasticsearch_controller.go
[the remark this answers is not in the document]
thbkrkr: We can avoid to panic by using `GetCASecretName()`. Then without the `association.AreConfiguredIfSet` check, we create the pod with empty secret name and the reconciliation fails:

```
{FieldValueRequired Required value spec.volumes[14].secret.secretName} 
{FieldValueRequired Required value spec.volumes[16].secret.secretName} 
{FieldValueInvalid Invalid value: \"\": a DNS-1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*') spec.containers[1].env[5].valueFrom.secretKeyRef.name} 
{FieldValueRequired Required value spec.containers[1].env[5].valueFrom.secretKeyRef.key} 
{FieldValueNotFound Not found: \"es-monitoring-metrics-target-certs\" spec.containers[1].volumeMounts[2].name} 
```

> Should it not be OK to deploy Elasticsearch even if the monitoring clusters are not ready yet or the association not configured?

With or without restarting the Elasticsearch container when it's ready/configured?
I agree it would be good to have this but it doesn't seem trivial.
I see 3 options:

When monitoring clusters are not ready yet or the association not configured?
1) wait and deploy ES with the sidecars
2) deploy ES and when it's ready recreate the pod with the sidecars
3) don't wait, deploy ES with the sidecars by "precomputing" the association conf

We do (1). (2) involves restarting the Elasticsearch container to reconfigure the pod with the sidecars.

I'm not sure of the complexity of (3). Assoc conf info are `{"authSecretName":"test-c-m2-beat-mon-user","authSecretKey":"a-test-c-m2-beat-mon-user","caCertProvided":true,"caSecretName":"test-es-mon-c-m2-ca","url":"https://m2-es-http.c.svc:9200","version":"7.14.0"}`.

I do not have a strong opinion between (1) and (2).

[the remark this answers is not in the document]
pebrc: If it is complicated it is maybe not worth it. We can also revisit this later once we have the basic functionality merged.

pebrc (commented):
I ran some more tests. I was able to override the Filebeat config by declaring my own ConfigMap volume called "filebeat-config" which is nice. The only downside is that you have to change the name of the file on every config change to trigger a reconciliation (or manually delete your pods)

pkg/apis/elasticsearch/v1/elasticsearch_types.go
```diff
@@ -72,6 +77,171 @@ type ElasticsearchSpec struct {
...
+type MetricsMonitoring struct {
+	// ElasticsearchRef is a reference to a monitoring Elasticsearch cluster running in the same Kubernetes cluster
+	// dedicated to receiving Stack Monitoring metrics.
+	ElasticsearchRef commonv1.ObjectSelector `json:"elasticsearchRef,omitempty"`
+}
+
+type LogsMonitoring struct {
+	// ElasticsearchRef is a reference to a monitoring Elasticsearch cluster running in the same Kubernetes cluster
+	// dedicated to receiving Elasticsearch logs.
+	ElasticsearchRef commonv1.ObjectSelector `json:"elasticsearchRef,omitempty"`
```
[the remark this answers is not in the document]
thbkrkr: I chose to support the list of `elasticsearchRef` in the `stackmon` package with a validation to only support one `elasticsearchRef` for now. d8bd2a6

thbkrkr:
> I was able to override the Filebeat config by declaring my own ConfigMap volume called "filebeat-config" which is nice. The only downside is that you have to change the name of the file on every config change to trigger a reconciliation (or manually delete your pods)

In order to improve this, we could inject a hash of the 2 beat configs in a pod label to trigger a pod recreation when the config changes. I don't think it's possible to just recreate one container without the whole pod. For example, I tested putting the hash in an env var of each container and that doesn't allow to recreate only the container.

More broadly, we could choose to:
1. not support a custom config. If you need something that is not set, use the Beats CRD.
2. support custom config provided by overriding the ConfigMap (=> do not reconcile the ConfigMap if there is already a ConfigMap)
3. support custom config embedded in the custom resource (`monitoring.metrics.config`), which is merged with the "default" config

I've mixed feelings to do (2) or (3) in the spirit of providing an easy and pretty opiniated way to enable Stack Monitoring. On the other hand, it's not super hard to support it. If we decide to support a custom config, (3) seems more aligned with what we already do on our custom resources and is more convenient for the user?

What do you think?

pebrc:
> What do you think?

I think it is OK to ship without explicit support for custom config, given that there is a workaround through the podTemplate and manual Pod deletion if users really need it.

thbkrkr:
I did some little refactorings to move the validation bits in one file and the beat configurations in secrets as it's done for other resources. Sorry for pushing more changes, I think it's good now.

thbkrkr:

config/crds/v1/all-crds.yaml
```diff
@@ -2833,6 +2833,80 @@ spec:
...
                 type: string
+              monitoring:
+                description: Monitoring enables you to collect and ship log and monitoring
+                  data of this Elasticsearch cluster. See https://www.elastic.co/guide/en/elasticsearch/reference/current/monitor-elasticsearch-cluster.html.
+                  Metricbeat and Filebeat are deployed in the same Pod as sidecars
+                  and each one sends data to one or two different Elasticsearch monitoring
+                  clusters running in the same Kubernetes cluster.
+                properties:
+                  logs:
+                    description: Logs holds references to Elasticsearch clusters which
```
alaudazzi, line 2844:
```suggestion
                    description: Holds references to Elasticsearch clusters which
```

config/crds/v1/all-crds.yaml
```diff
@@ -2833,6 +2833,80 @@ spec:
...
+                            name:
+                              description: Name of the Kubernetes object.
+                              type: string
+                            namespace:
+                              description: Namespace of the Kubernetes object. If
+                                empty, defaults to the current namespace.
+                              type: string
+                            serviceName:
+                              description: ServiceName is the name of an existing
+                                Kubernetes service which will be used to make requests
```
alaudazzi, line 2865:
```suggestion
                                Kubernetes service which is used to make requests
```

config/crds/v1/all-crds.yaml
```diff
@@ -2833,6 +2833,80 @@ spec:
...
+                                to the referenced object. It has to be in the same
+                                namespace as the referenced resource. If left empty
+                                the default HTTP service of the referenced resource
+                                will be used.
```
alaudazzi, line 2869:
```suggestion
                                namespace as the referenced resource. If left empty,
                                the default HTTP service of the referenced resource
                                is used.
```

config/crds/v1/all-crds.yaml
```diff
@@ -2833,6 +2833,80 @@ spec:
...
+                              type: string
+                          required:
+                          - name
+                          type: object
+                        type: array
+                    type: object
+                  metrics:
+                    description: Metrics holds references to Elasticsearch clusters
```
alaudazzi, line 2877:
```suggestion
                    description: Holds references to Elasticsearch clusters
```

config/crds/v1/all-crds.yaml
```diff
@@ -2833,6 +2833,80 @@ spec:
...
+                            name:
+                              description: Name of the Kubernetes object.
+                              type: string
+                            namespace:
+                              description: Namespace of the Kubernetes object. If
+                                empty, defaults to the current namespace.
+                              type: string
+                            serviceName:
+                              description: ServiceName is the name of an existing
+                                Kubernetes service which will be used to make requests
```
alaudazzi, line 2898:
```suggestion
                              description: ServiceName is the name of an existing
                                Kubernetes service which is used to make requests
```

config/crds/v1/all-crds.yaml
```diff
@@ -2833,6 +2833,80 @@ spec:
...
+                                to the referenced object. It has to be in the same
+                                namespace as the referenced resource. If left empty
+                                the default HTTP service of the referenced resource
+                                will be used.
```
alaudazzi, line 2902:
```suggestion
                                namespace as the referenced resource. If left empty,
                                the default HTTP service of the referenced resource
                                is used.
```

pkg/controller/elasticsearch/stackmon/volume.go
```diff
@@ -0,0 +1,81 @@
...
+	return volumes
+}
+
+func monitoringLogsTargetCaCertSecretVolumes(es esv1.Elasticsearch) []volume.VolumeLike {
+	volumes := make([]volume.VolumeLike, 0)
+	for i, assoc := range es.GetMonitoringLogsAssociation() {
+		volumes = append(volumes, volume.NewSecretVolumeWithMountPath(
+			assoc.AssociationConf().GetCASecretName(),
+			fmt.Sprintf(monitoringLogsTargetEsCaCertVolumeNameFormat, i),
+			fmt.Sprintf(monitoringLogsTargetEsCaCertMountPathFormat, i),
```
pebrc, line 77: I wonder if we should adopt a similar approach here to what we do in Agent and not use the order dependent indices as part of the mount path. 

I _think_ it does not matter in this first stage where we only support one reference per type but I wonder if this could become a problem in the future. 

In Agent we use namespace/name instead.

config/crds/v1/all-crds.yaml
```diff
@@ -3724,6 +3798,15 @@ spec:
...
                   by the health API.
                 type: string
+              monitoringAssociationStatus:
+                additionalProperties:
+                  description: AssociationStatus is the status of an association resource.
+                  type: string
+                description: AssociationStatusMap is the map of association's namespaced
+                  name string to its AssociationStatus. For resources that have a
+                  single Association of a given type (eg. single ES reference), this
+                  map will contain a single entry.
```
alaudazzi, line 3808:
```suggestion
                description: AssociationStatusMap is the map of association's namespaced
                  name string to its AssociationStatus. For resources that have a
                  single Association of a given type (for ex. single ES reference), this
                  map contains a single entry.
```
This sentence is bit unclear:
AssociationStatusMap is the map of association's namespaced name string 
Not sure how to reword it...

config/crds/v1/bases/elasticsearch.k8s.elastic.co_elasticsearches.yaml
```diff
@@ -494,6 +494,80 @@ spec:
...
                 type: string
+              monitoring:
+                description: Monitoring enables you to collect and ship log and monitoring
+                  data of this Elasticsearch cluster. See https://www.elastic.co/guide/en/elasticsearch/reference/current/monitor-elasticsearch-cluster.html.
+                  Metricbeat and Filebeat are deployed in the same Pod as sidecars
+                  and each one sends data to one or two different Elasticsearch monitoring
+                  clusters running in the same Kubernetes cluster.
+                properties:
+                  logs:
+                    description: Logs holds references to Elasticsearch clusters which
```
alaudazzi, line 505:
```suggestion
                    description: Holds references to Elasticsearch clusters which
```

config/crds/v1/bases/elasticsearch.k8s.elastic.co_elasticsearches.yaml
```diff
@@ -494,6 +494,80 @@ spec:
...
+                            name:
+                              description: Name of the Kubernetes object.
+                              type: string
+                            namespace:
+                              description: Namespace of the Kubernetes object. If
+                                empty, defaults to the current namespace.
+                              type: string
+                            serviceName:
+                              description: ServiceName is the name of an existing
+                                Kubernetes service which will be used to make requests
```
alaudazzi, line 526:
```suggestion
                              description: ServiceName is the name of an existing
                                Kubernetes service which is used to make requests
```

config/crds/v1/bases/elasticsearch.k8s.elastic.co_elasticsearches.yaml
```diff
@@ -494,6 +494,80 @@ spec:
...
+                                to the referenced object. It has to be in the same
+                                namespace as the referenced resource. If left empty
+                                the default HTTP service of the referenced resource
+                                will be used.
```
alaudazzi, line 530:
```suggestion
                                namespace as the referenced resource. If left empty,
                                the default HTTP service of the referenced resource
                                is used.
```

config/crds/v1/bases/elasticsearch.k8s.elastic.co_elasticsearches.yaml
```diff
@@ -494,6 +494,80 @@ spec:
...
+                              type: string
+                          required:
+                          - name
+                          type: object
+                        type: array
+                    type: object
+                  metrics:
+                    description: Metrics holds references to Elasticsearch clusters
+                      which will receive monitoring data from this Elasticsearch cluster.
```
alaudazzi, line 539:
```suggestion
                    description: Metrics holds references to Elasticsearch clusters
                      which receive monitoring data from this Elasticsearch cluster.
```

alaudazzi (commented):
Left a few minor editing suggestions.

config/crds/v1/all-crds.yaml
thbkrkr: It is imposed by Godoc, the first word of the description in comment must correspond to the described property.

pkg/controller/elasticsearch/stackmon/volume.go
thbkrkr: Good point.
This implies to generate the `output` section in the beat configuration.
```

Comment (thbkrkr) [reply pkg/controller/elasticsearch/stackmon/volume.go:77]: Good point. I will put this aside as it's not trivial.
I think then it implies to generate the ouput section in the beat config. (I probably should have done this from the start.)

Because currently it references the indices:
```
output.elasticsearch:
  hosts: ['${ES_0_TARGET_URL}']
  username: ${ES_0_TARGET_USERNAME}
  password: ${ES_0_TARGET_PASSWORD}
  ssl.certificate_authorities: ["/mnt/es/0/monitoring/logs/target/ca.crt"]
```
Note: the env var indirection instead of the raw value is useful to consume the password through the secret as env var.

**Comment by thbkrkr** [comment]
I have a new naming proposal:
```diff
spec:
- monitoring:
+ observability:
    metrics:
      elasticsearchRefs: 
        - name: m1
          namespace: m1
-   logs:
+   logging:
      elasticsearchRefs:
        - name: m2
          namespace: m2
```

sebgl:
```
- monitoring:
+ observability:
```

This may be a bit confusing with [Elastic Observability](https://www.elastic.co/observability), and lead users to think this is about observing their k8s cluster?

pebrc:
> ```
> - monitoring:
> + observability:
> ```
> 
> 
>      
>   
> 
> This may be a bit confusing with [Elastic Observability](https://www.elastic.co/observability), and lead users to think this is about observing their k8s cluster?

Agreed. I think we should stick with the current naming.

thbkrkr:
jenkins test this please

pkg/apis/elasticsearch/v1/elasticsearch_types.go
```diff
@@ -72,6 +77,182 @@ type ElasticsearchSpec struct {
...
+	// annotation key should be stable to allow Elasticsearch Controller only pick up the ones it expects,
+	// based on ElasticsearchRefs
+
+	nsNameHash := sha256.New224()
+	// concat with dot to avoid collisions, as namespace can't contain dots
+	_, _ = nsNameHash.Write([]byte(fmt.Sprintf("%s.%s", eea.ref.Namespace, eea.ref.Name)))
+	// base32 to encode and limit the length, as using Sprintf with "%x" encodes with base16 which happens to
+	// give too long output
+	// no padding to avoid illegal '=' character in the annotation name
+	hash := base32.StdEncoding.WithPadding(base32.NoPadding).EncodeToString(nsNameHash.Sum(nil))
```
sebgl, line 142: I think this is duplicated code with agent? Can we factor this into a single function we reuse?

pkg/apis/elasticsearch/v1/elasticsearch_types.go
```diff
@@ -72,6 +77,182 @@ type ElasticsearchSpec struct {
...
+		return nil
+	}
+	if eea.Elasticsearch == nil {
+		eea.Elasticsearch = &Elasticsearch{}
+	}
+	return eea.Elasticsearch
+}
+
+func (eea *EsMonitoringAssociation) AssociationConfAnnotationName() string {
+	// annotation key should be stable to allow Elasticsearch Controller only pick up the ones it expects,
```
sebgl, line 133:
```suggestion
	// annotation key should be stable to allow the Elasticsearch Controller to only pick up the ones it expects,
```

pkg/controller/association/controller/es_monitoring.go
```diff
@@ -0,0 +1,60 @@
...
+	"github.com/elastic/cloud-on-k8s/pkg/utils/rbac"
+)
+
+const (
+	// EsMonitoringAssociationLabelName marks resources created by this controller for easier retrieval.
+	EsMonitoringAssociationLabelName = "esmonitoringassociation.k8s.elastic.co/name"
+	// EsMonitoringAssociationLabelNamespace marks resources created by this controller for easier retrieval.
+	EsMonitoringAssociationLabelNamespace = "esmonitoringassociation.k8s.elastic.co/namespace"
+	// EsMonitoringAssociationLabelType marks the type of association.
+	EsMonitoringAssociationLabelType = "esmonitoringassociation.k8s.elastic.co/type"
```
sebgl, line 27: I think the labelling logic of other association controllers is slightly different, and closer to something like:

```
// to which ES cluster is this association coming from
esassociation.k8s.elastic.co/name: es-name
esassociation.k8s.elastic.co/namespace: default
// what's the type of association here?
esassociation.k8s.elastic.co/type: es-monitoring  // associated to ES, for monitoring reasons (in case we add other association reasons later on)
```

so the label name means "this is an association to elasticsearch", and the label values mean "this is the associated elasticsearch, and the type of association for that elasticsearch".

Here it's slightly different since the type of association is reflected in the label name? WDYT? (def not a big deal, trying to be as consistent as possible)

pkg/controller/common/defaults/pod_template.go
```diff
@@ -55,23 +55,27 @@ func NewPodTemplateBuilder(base corev1.PodTemplateSpec, containerName string) *P
...
+// getContainer retrieves the existing Container from the pod template
+func (b *PodTemplateBuilder) getContainer() *corev1.Container {
+	for i, c := range b.PodTemplate.Spec.Containers {
+		if c.Name == b.containerName {
+			return &b.PodTemplate.Spec.Containers[i]
+		}
+	}
+	return nil
+}
+func (b *PodTemplateBuilder) setContainerDefaulter() {
```
sebgl, line 67: nit: new line

pkg/controller/common/defaults/pod_template.go
```diff
@@ -55,23 +55,27 @@ func NewPodTemplateBuilder(base corev1.PodTemplateSpec, containerName string) *P
 	return builder.setDefaults()
 }
 
+// getContainer retrieves the existing Container from the pod template
```
sebgl, line 58:
```suggestion
// getContainer retrieves the main Container from the pod template
```

pkg/controller/common/defaults/pod_template.go
```diff
@@ -55,23 +55,27 @@ func NewPodTemplateBuilder(base corev1.PodTemplateSpec, containerName string) *P
...
+func (b *PodTemplateBuilder) getContainer() *corev1.Container {
+	for i, c := range b.PodTemplate.Spec.Containers {
+		if c.Name == b.containerName {
+			return &b.PodTemplate.Spec.Containers[i]
+		}
+	}
+	return nil
+}
```
sebgl, line 66: can we unit test that to catch future regressions? (it involves pointers and a for loop so a unit test may not hurt 🤷)

pkg/controller/elasticsearch/stackmon/config.go
```diff
@@ -0,0 +1,130 @@
...
+// MonitoringConfig returns the Elasticsearch settings required to enable the collection of monitoring data
+func MonitoringConfig(es esv1.Elasticsearch) commonv1.Config {
+	if !IsMonitoringMetricsDefined(es) {
+		return commonv1.Config{}
+	}
+	return commonv1.Config{Data: map[string]interface{}{
+		esv1.XPackMonitoringCollectionEnabled:              true,
+		esv1.XPackMonitoringElasticsearchCollectionEnabled: false,
+	}}
+}
```
sebgl, line 56: For readability it could be nice to decouple `beat_config.go` and `es_config.go`?

pkg/controller/elasticsearch/stackmon/config.go
```diff
@@ -0,0 +1,130 @@
...
+	esTargetURLEnvVarKeyFormat      = "ES_%d_TARGET_URL"
+	esTargetUsernameEnvVarKeyFormat = "ES_%d_TARGET_USERNAME"
+	esTargetPasswordEnvVarKeyFormat = "ES_%d_TARGET_PASSWORD" //nolint:gosec
+
+	monitoringMetricsSourceEsCaCertMountPath       = "/mnt/es/monitoring/metrics/source"
+	monitoringMetricsTargetEsCaCertMountPathFormat = "/mnt/es/%d/monitoring/metrics/target"
+	monitoringLogsTargetEsCaCertMountPathFormat    = "/mnt/es/%d/monitoring/logs/target"
+
+	// metricbeatConfig is a static configuration for Metricbeat to collect monitoring data about Elasticsearch
+	//go:embed metricbeat.yml
```
sebgl, line 39: :yay:

pkg/controller/elasticsearch/stackmon/config.go
```diff
@@ -0,0 +1,130 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// Environments variables and paths to the Elasticsearch CA certificates used in the beats configuration to describe
+// how to connect to Elasticsearch.
+// Warning: environment variables and CA cert paths defined below are also used in the embedded files.
+var (
+	esSourceURLEnvVarKey      = "ES_SOURCE_URL"
+	esSourceURLEnvVarValue    = "https://localhost:9200"
```
sebgl, line 26: what if https is disabled?

pkg/controller/elasticsearch/stackmon/config.go
```diff
@@ -0,0 +1,130 @@
...
+	esSourceUsernameEnvVarKey = "ES_SOURCE_USERNAME"
+	esSourcePasswordEnvVarKey = "ES_SOURCE_PASSWORD" //nolint:gosec
+
+	esTargetURLEnvVarKeyFormat      = "ES_%d_TARGET_URL"
+	esTargetUsernameEnvVarKeyFormat = "ES_%d_TARGET_USERNAME"
+	esTargetPasswordEnvVarKeyFormat = "ES_%d_TARGET_PASSWORD" //nolint:gosec
+
+	monitoringMetricsSourceEsCaCertMountPath       = "/mnt/es/monitoring/metrics/source"
+	monitoringMetricsTargetEsCaCertMountPathFormat = "/mnt/es/%d/monitoring/metrics/target"
```
sebgl, line 35: the directory tree could be a bit more consistent with `/mnt/es/monitoring/metrics/target/%d`?
So we get:

```
# /mnt/es/monitoring/metrics
source
target/1
target/2
```

rather than:

```
# /mnt/es
monitoring
1/monitoring
2/monitoring
```

pkg/controller/elasticsearch/stackmon/container.go
```diff
@@ -0,0 +1,176 @@
...
+}
+
+// WithMonitoring updates the Elasticsearch Pod template builder to deploy Metricbeat and Filebeat in sidecar containers
+// in the Elasticsearch pod and injects volumes for Metricbeat/Filebeat configs and ES source/target CA certs.
+func WithMonitoring(builder *defaults.PodTemplateBuilder, es esv1.Elasticsearch) (*defaults.PodTemplateBuilder, error) {
+	isMonitoringMetrics := IsMonitoringMetricsDefined(es)
+	isMonitoringLogs := IsMonitoringLogsDefined(es)
+
+	// No monitoring defined, skip
+	if !isMonitoringMetrics && !isMonitoringLogs {
```
sebgl, line 60: `if !IsStackMonitoringDefined(es)`?

pkg/controller/elasticsearch/stackmon/container.go
```diff
@@ -0,0 +1,176 @@
...
+	"github.com/elastic/cloud-on-k8s/pkg/controller/common/container"
+	"github.com/elastic/cloud-on-k8s/pkg/controller/common/defaults"
+	"github.com/elastic/cloud-on-k8s/pkg/controller/common/volume"
+	esvolume "github.com/elastic/cloud-on-k8s/pkg/controller/elasticsearch/volume"
+	corev1 "k8s.io/api/core/v1"
+)
+
+const (
+	esLogStyleEnvVarKey   = "ES_LOG_STYLE"
+	esLogStyleEnvVarValue = "file"
```
sebgl, line 22: let's add a comment about how this is a configuration option for the ES container (could be hard to understand what's beat config vs. what's ES config)

pkg/controller/elasticsearch/stackmon/container.go
```diff
@@ -0,0 +1,176 @@
...
+	volumes := make([]corev1.Volume, 0)
+	for _, v := range volumeLikes {
+		volumes = append(volumes, v.Volume())
+	}
+	builder.WithVolumes(volumes...)
+
+	return builder, nil
+}
+
+func stackLoggingEnvVar() corev1.EnvVar {
```
sebgl, line 110:
```suggestion
func fileLogStyleEnvVar() corev1.EnvVar {
```

pkg/controller/elasticsearch/stackmon/container.go
```diff
@@ -0,0 +1,176 @@
...
+		Env:          append(envVars, defaults.PodDownwardEnvVars()...),
+		VolumeMounts: volumeMounts,
+	}, nil
+}
+
+// fullContainerImage returns the full Beat container image with the image registry.
+// If the Elasticsearch specification is configured with a custom image, we do best effort by trying to derive the Beat
+// image from the Elasticsearch custom image with an image name replacement
+// (<registry>/elasticsearch/elasticsearch:<version> becomes <registry>/beats/<filebeat|metricbeat>:<version>)
+func fullContainerImage(es esv1.Elasticsearch, defaultImage container.Image) (string, error) {
```
sebgl, line 164: I'm wondering if this is not a bit too much?
vs. clearly documenting how users can override the Beats image through the podTemplate?

pkg/controller/elasticsearch/stackmon/filebeat.yml
```diff
@@ -0,0 +1,47 @@
+filebeat.modules:
+  # https://www.elastic.co/guide/en/beats/filebeat/current/filebeat-module-elasticsearch.html
```
sebgl, line 2: let's link to the actual version we took this from rather than `current`?

pkg/controller/elasticsearch/stackmon/filebeat.yml
```diff
@@ -0,0 +1,47 @@
...
+        - /usr/share/elasticsearch/logs/*_deprecation.json
+      close_timeout: 2h
+      fields_under_root: true
+
+processors:
+  - add_cloud_metadata: {}
+  - add_host_metadata: {}
+
+output.elasticsearch:
+  hosts: ['${ES_0_TARGET_URL}']
```
sebgl, line 44: could deserve a comment

```
// only one target ES can be referenced so far so we take it as ordinal 0
```

pkg/controller/elasticsearch/stackmon/container.go
```diff
@@ -0,0 +1,176 @@
...
+		if !ref.IsDefined() {
+			return false
+		}
+	}
+	return len(es.Spec.Monitoring.Logs.ElasticsearchRefs) > 0
+}
+
+// WithMonitoring updates the Elasticsearch Pod template builder to deploy Metricbeat and Filebeat in sidecar containers
+// in the Elasticsearch pod and injects volumes for Metricbeat/Filebeat configs and ES source/target CA certs.
+func WithMonitoring(builder *defaults.PodTemplateBuilder, es esv1.Elasticsearch) (*defaults.PodTemplateBuilder, error) {
```
sebgl, line 55: it could be nice to unit test this function

pkg/controller/elasticsearch/stackmon/validations.go
```diff
@@ -0,0 +1,61 @@
...
+var (
+	// Minimum Stack version to enable Stack Monitoring.
+	// This requirement comes from the fact that we configure Elasticsearch to write logs to disk for Filebeat
+	// via the env var ES_LOG_STYLE available from this version.
+	MinStackVersion = version.MustParse("7.14.0")
+)
+
+// Validate validates that the Elasticsearch version is supported for Stack Monitoring and that there is exactly one
+// Elasticsearch reference defined when Stack Monitoring is defined
+func Validate(es esv1.Elasticsearch) field.ErrorList {
```
sebgl, line 30: as a general principle I think any validation function should be unit-tested, even though they may be very simple functions? At least to catch future regressions!

pkg/controller/elasticsearch/stackmon/volume.go
sebgl: I think I'd be +1 with trying to use explicit namespace & names rather than ordinals.
It means the config file needs to be templatized, I guess? Maybe not that hard?

test/e2e/es/stack_monitoring_test.go
```diff
@@ -0,0 +1,129 @@
...
+	"strconv"
+	"testing"
+
+	esClient "github.com/elastic/cloud-on-k8s/pkg/controller/elasticsearch/client"
+	"github.com/elastic/cloud-on-k8s/pkg/controller/elasticsearch/stackmon"
+	"github.com/elastic/cloud-on-k8s/test/e2e/test"
+	"github.com/elastic/cloud-on-k8s/test/e2e/test/elasticsearch"
+)
+
+// TestStackMonitoring
```
sebgl, line 22: let's describes what this test does in the comment?

test/e2e/es/stack_monitoring_test.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+// Index partially models Elasticsearch cluster index returned by /_cat/indices
+type Index struct {
+	Index     string `json:"index"`
+	DocsCount string `json:"docs.count"`
+}
+
+func AreIndexedDocs(esClient esClient.Client, indexPattern string) error {
+	req, err := http.NewRequest(http.MethodGet, fmt.Sprintf("/_cat/indices/%s?health=green&format=json", indexPattern), nil) //nolint:noctx
```
sebgl, line 96: Why `health=green `? (deserves a comment)

sebgl (commented):
Awesome work it's very easy to read and understand!
It's amazing how it "just works".

The main concern I have is about cases where TLS is disabled at ES level and Beats still tries to connect through https.

pkg/apis/elasticsearch/v1/elasticsearch_types.go
```diff
@@ -72,6 +77,182 @@ type ElasticsearchSpec struct {
...
+type EsMonitoringAssociation struct {
+	// The monitored Elasticsearch cluster from where are collected logs and monitoring metrics
+	*Elasticsearch
+	// ref is the namespaced name of the Elasticsearch referenced in the Association used to send and store monitoring data
+	ref types.NamespacedName
+}
+
+var _ commonv1.Association = &EsMonitoringAssociation{}
+
+func (eea *EsMonitoringAssociation) Associated() commonv1.Associated {
```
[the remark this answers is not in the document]
thbkrkr: Residue of `EsEsAssociation`. I fix.

pkg/controller/elasticsearch/stackmon/container.go
thbkrkr: I think you are right. I prefer to document the podTemplate override than that.

pkg/controller/association/controller/es_monitoring.go
thbkrkr: Makes sense.

pkg/controller/elasticsearch/stackmon/validations.go
thbkrkr: It's a good general principle, you're right, I'm adding unit tests.

test/e2e/es/stack_monitoring_test.go
thbkrkr: You beat me. It is not really necessary, we are not here to test the index replication. I'll remove it.

pkg/controller/elasticsearch/stackmon/config.go
thbkrkr: I forgot that. I will see to support it. It shouldn't be difficult.

thbkrkr:
Remains to be done:

- [ ] support tls disabled (both on monitored source and monitoring target ES)
- [ ] use namespace/name instead of ordinals in volume names and mount paths of the monitoring target ES
- [ ] support rotation of the ES user password (both on monitored source and monitoring target ES)

thbkrkr:
Dear reviewers, to fill in the last gaps, I pushed the version I was working on to add monitoring to Kibana. It is less declarative and more dynamic to manage the configuration of the beats and it factors code to reuse it for Kibana monitoring. I addressed the last missing points on it. The code diverges from the first version but I think it's better this way, more powerful and robust, less verbose, without duplication.

pkg/controller/elasticsearch/stackmon/config.go
thbkrkr: This is covered by the latest version using go templating.

thbkrkr:
Jenkins test this please.

pkg/controller/common/stackmon/beat.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+func NewFilebeatBuilder(client k8s.Client, resource HasMonitoring, baseConfig string, additionalVolume volume.VolumeLike) (BeatBuilder, error) {
+	return NewBeatBuilder(client, "filebeat", container.FilebeatImage, resource,
+		resource.GetMonitoringLogsAssociation(), baseConfig, additionalVolume)
+}
+
+// BeatBuilder helps with building a beat sidecar container to monitor an Elastic Stack application. It focuses on
+// building the container, the secret holding the configuration file and the associated volumes for the pod.
+type BeatBuilder struct {
```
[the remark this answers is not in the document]
thbkrkr: You are right, "Builder" is not appropriate. I went with `BeatSidecar` because there is already a` beatConfig` struct.

pkg/controller/common/stackmon/monitoring.go
```diff
@@ -0,0 +1,19 @@
...
+import (
+	commonv1 "github.com/elastic/cloud-on-k8s/pkg/apis/common/v1"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+// HasMonitoring is the interface implemented by an Elastic Stack application that supports Stack Monitoring ()
+type HasMonitoring interface {
+	GetMonitoringMetricsAssociation() []commonv1.Association
+	GetMonitoringLogsAssociation() []commonv1.Association
+	NSN() types.NamespacedName
```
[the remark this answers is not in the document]
thbkrkr: Good idea!

pkg/controller/elasticsearch/stackmon/validations.go
```diff
@@ -0,0 +1,61 @@
...
+const (
+	unsupportedVersionForStackMonitoringMsg    = "Unsupported version for Stack Monitoring. Required >= %s."
+	invalidStackMonitoringElasticsearchRefsMsg = "Only one Elasticsearch reference is supported for %s Stack Monitoring"
+)
+
+var (
+	// Minimum Stack version to enable Stack Monitoring.
+	// This requirement comes from the fact that we configure Elasticsearch to write logs to disk for Filebeat
+	// via the env var ES_LOG_STYLE available from this version.
+	MinStackVersion = version.MustParse("7.14.0")
```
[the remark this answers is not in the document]
thbkrkr: Let's do this to get CI on the `7.14.0-SNAPSHOT`.

pkg/controller/common/stackmon/monitoring.go
```diff
@@ -0,0 +1,19 @@
...
+	ShortKind() string
```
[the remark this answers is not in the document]
thbkrkr: f1a45c9d68 instead uses `assoc.AssociationType()` for the beat config volume name and directly `es` for the local ES CA config name for Metricbeat as it is ES specific.

pkg/controller/common/stackmon/monitoring.go
```diff
@@ -0,0 +1,19 @@
...
+	Version() string
```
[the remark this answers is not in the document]
thbkrkr: The association has the version of the referenced resource, not the associated, which can be different. I moved where we get the full container image to remove the need of the version in this interface in f1a45c9.

pebrc (commented):
Filebeat keeps crash looping for me with `one or more modules must be configured` I haven't quite figured out what causes it but my guess is YAML indentation is to blame somewhere. Otherwise I think we are almost ready to merge 👍

pkg/controller/elasticsearch/stackmon/validations.go
```diff
@@ -0,0 +1,61 @@
...
+		return field.ErrorList{field.Invalid(field.NewPath("spec").Child("monitoring").Child("metrics").Child("elasticsearchRefs"),
+			es.Spec.Monitoring.Metrics.ElasticsearchRefs,
+			fmt.Sprintf(invalidStackMonitoringElasticsearchRefsMsg, "Metrics"))}
+	}
+	if IsMonitoringLogsDefined(es) && len(es.Spec.Monitoring.Logs.ElasticsearchRefs) != 1 {
+		return field.ErrorList{field.Invalid(field.NewPath("spec").Child("monitoring").Child("logs").Child("elasticsearchRefs"),
+			es.Spec.Monitoring.Logs.ElasticsearchRefs,
+			fmt.Sprintf(invalidStackMonitoringElasticsearchRefsMsg, "Logs"))}
+	}
+	return field.ErrorList{}
```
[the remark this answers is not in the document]
thbkrkr: Yes, it will improve the UX. I forgot we could do that, tss.

test/e2e/es/stack_monitoring_test.go
```diff
@@ -0,0 +1,155 @@
...
+	}
+
+	// create 1 monitored and 2 monitoring clusters to collect separately metrics and logs
+	metrics := elasticsearch.NewBuilder("test-es-mon-metrics").
+		WithESMasterDataNodes(2, elasticsearch.DefaultResources)
+	logs := elasticsearch.NewBuilder("test-es-mon-logs").
+		WithESMasterDataNodes(2, elasticsearch.DefaultResources)
+	monitored := elasticsearch.NewBuilder("test-es-mon-a").
+		WithESMasterDataNodes(1, elasticsearch.DefaultResources).
+		WithMonitoring(metrics.Ref(), logs.Ref())
```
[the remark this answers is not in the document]
thbkrkr: Not only, it's also because we're adding the namespace.

<clusterName>-es-monitoring-<namespace>-<clustername>-ca


36+15+63+1+36+3=154 < 253

test-es-mon-a-8xhw-es-monitoring-e2e-mercury-test-es-mon-logs-jbh2-ca


[
	{FieldValueInvalid Invalid value: "test-es-mon-a-8xhw-es-monitoring-e2e-mercury-test-es-mon-logs-jbh2-ca": must be no more than 63 characters spec.volumes[14].name} 
	{FieldValueInvalid Invalid value: "test-es-mon-a-8xhw-es-monitoring-e2e-mercury-test-es-mon-metrics-884m-ca": must be no more than 63 characters spec.volumes[15].name}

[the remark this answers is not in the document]
thbkrkr: My bad, I've not retested and I thought that the max volume name length was like the max secret name length (253) but it's 63.

The pattern used is: `<clusterName>-es-monitoring-<namespace>-<clustername>-ca` (ex: test-es-mon-a-8xhw-es-monitoring-e2e-mercury-test-es-mon-logs-jbh2-ca).
Max cluster name length is 36. Max namespace name length is 63.
36+15+63+1+36+3 = 154 > 63

So, we can't reuse the assoc ca secret name for the CA volume name. A volume name is 'local' to a pod, so we don't need the current es cluster name but we want to differentiate volumes coming from the different ES referenced for monitoring.
We could add the beginning of the sha of the "namespace+name" and still add the es ref name for debugging: `es-monitoring-<ns/clustername hash>-<clustername>-ca`?

Same issue with the config volume name:
`fmt.Sprintf("%s-%s-%s-config", resource.GetName(), assoc.AssociationType(), beatName)`
=> `<clusterName>-es-monitoring-metricbeat-config` 36+32 = 68 > 63

thbkrkr:
> Filebeat keeps crash looping for me with `one or more modules must be configured` I haven't quite figured out what causes it but my guess is YAML indentation is to blame somewhere.

Bad copy-paste f4759fdf :man_facepalming:

test/e2e/es/stack_monitoring_test.go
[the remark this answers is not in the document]
pebrc: Sounds good to me. Even just the hash I guess you would still see where it comes from in the referenced secret's name.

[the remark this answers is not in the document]
thbkrkr: c26faa5

pebrc (approved):
LGTM! Nice work @thbkrkr 👍 I have done a few tests, ran the e2e test you added, explored the validations. There is probably room for more in-depth testing of edge cases where a resource constrained Metricbeat or Filebeat bring down a whole Pod but I haven't tried that myself.

thbkrkr:
Thanks Peter! I'm going to follow-up with Kibana monitoring. 

> There is probably room for more in-depth testing of edge cases where a resource constrained Metricbeat or Filebeat bring down a whole Pod but I haven't tried that myself.

This is a disadvantage of the sidecar pattern. If it happens, you will have to increase the compute resources.

I did a quick test be restricting memory for Metricbeat:
```
kind: Elasticsearch
metadata:
  name: alpha
  namespace: production
spec:
  version: 7.14.0-SNAPSHOT
  monitoring:
    metrics:
      elasticsearchRefs:
        - name: monitoring
          namespace: observability
    logs:
      elasticsearchRefs:
        - name: monitoring
          namespace: observability
  nodeSets:
  - name: master
    count: 2
    config:
      node.store.allow_mmap: false
    podTemplate:
      spec:
        containers:
          - name: metricbeat
            resources:
              limits:
                memory: 20M
```

Elasticsearch is green and reachable but from the operator's point of view, the ES resource is red and it blocks the deployment of the associated Kibana.

```
> k get elastic,sts,deploy,pods
NAME                                               HEALTH    NODES   VERSION           PHASE             AGE
elasticsearch.elasticsearch.k8s.elastic.co/alpha   unknown           7.14.0-SNAPSHOT   ApplyingChanges   14m

NAME                                 HEALTH   NODES   VERSION           AGE
kibana.kibana.k8s.elastic.co/alpha   red              7.14.0-SNAPSHOT   14m

NAME                               READY   AGE
statefulset.apps/alpha-es-master   0/2     13m

NAME                       READY   UP-TO-DATE   AVAILABLE   AGE
deployment.apps/alpha-kb   0/1     1            0           13m

NAME                            READY   STATUS             RESTARTS   AGE
pod/alpha-es-master-0           2/3     CrashLoopBackOff   7          13m
pod/alpha-es-master-1           2/3     CrashLoopBackOff   7          13m
pod/alpha-kb-6cf84594b9-rm4wp   0/1     Running            0          13m
```

```                                                                                                                                                                                                                                                                                                             
> eckurl production alpha '/_cat/health' 
1625568204 10:43:24 alpha green 2 2 2 1 0 0 0 0 - 100.0%
```

To increase the memory limit, you need to apply the updated manifest with a new limit and kill the pod.

PR #4533, opened 2021-05-28
[Autoscaling] Fix e2e test

barkbay:
#4493  broke e2e autoscaling test for 2 reasons:

1. The operator now adds a storage capacity margin of 5% to account for the fs reserved disk space. [Storage capacity requested using the fixed decider needs to be adjusted](https://github.com/elastic/cloud-on-k8s/commit/00cb71be9a897f2ace5bff4919d48f10de905bb5).
2. A total storage requirement of 0 explicitly means that storage recommender should not prevent a horizontal scale down. This is for example the case in the context of the ML decider.

This PR:

* Fixes the e2e tests
* Handles the case where a storage capacity request is 0
* Adds a unit test to catch this regression earlier

### DIFF
--- a/pkg/controller/autoscaling/elasticsearch/autoscaler/linear_scaler.go
+++ b/pkg/controller/autoscaling/elasticsearch/autoscaler/linear_scaler.go
@@ -68,8 +68,8 @@ func memoryFromStorage(requiredStorageCapacity resource.Quantity, storageRange, 
 	requiredMemoryCapacity := memoryRange.Min.Value() + requiredAdditionalMemoryCapacity
 
 	// Round up memory to the next GB
-	requiredMemoryCapacity = math.RoundUp(requiredMemoryCapacity, resources.GIB)
-	resourceMemoryAsGiga := resource.MustParse(fmt.Sprintf("%dGi", requiredMemoryCapacity/resources.GIB))
+	requiredMemoryCapacity = math.RoundUp(requiredMemoryCapacity, resources.GiB)
+	resourceMemoryAsGiga := resource.MustParse(fmt.Sprintf("%dGi", requiredMemoryCapacity/resources.GiB))
 
 	if resourceMemoryAsGiga.Cmp(memoryRange.Max) > 0 {
 		resourceMemoryAsGiga = memoryRange.Max.DeepCopy()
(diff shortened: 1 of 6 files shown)


pkg/controller/autoscaling/elasticsearch/resources/resources.go
```diff
@@ -15,8 +15,10 @@ import (
 )
 
 const (
-	// GIB - 1 GibiByte
+	// GIB - 1 GibiBytes
```
thbkrkr, line 18: If 1 GiB, why a final `s`?
```suggestion
	// GIB - 1 Gibibyte
```

pkg/controller/autoscaling/elasticsearch/resources/resources.go
```diff
@@ -15,8 +15,10 @@ import (
...
 	GIB = int64(1024 * 1024 * 1024)
```
thbkrkr, line 19: Nit: Unrelated to this PR but the acronym of Gibibyte is GiB with the i in lowercase.

pkg/controller/autoscaling/elasticsearch/resources/resources.go
```diff
@@ -15,8 +15,10 @@ import (
...
+	// GB - 1 Gigabytes
```
thbkrkr, line 20:
```suggestion
	// GB - 1 Gigabyte
```

pkg/controller/autoscaling/elasticsearch/autoscaler/linear_scaler.go
```diff
@@ -68,8 +68,8 @@ func memoryFromStorage(requiredStorageCapacity resource.Quantity, storageRange,
 	requiredMemoryCapacity := memoryRange.Min.Value() + requiredAdditionalMemoryCapacity
 
 	// Round up memory to the next GB
```
thbkrkr, line 70: Nit: unrelated to this commit
```suggestion
	// Round up memory to the next GiB
```

pkg/controller/autoscaling/elasticsearch/autoscaler/recommender/recommender.go
```diff
@@ -110,7 +110,7 @@ func getResourceValue(
 	}
 
 	// Try to round up the Gb value
```
thbkrkr, line 112:
```suggestion
	// Try to round up the GiB value
```

thbkrkr (approved):

barkbay:
This is because Elasticsearch translate 20gb to 20GiB, not 20GB:

```json
"fixed": {
	"required_capacity": {
		"node": {
			"storage": 21474836480
		},
		"total": {
			"storage": 64424509440
		}
	},
	"reason_summary": "fixed storage [20gb] memory [null] nodes [3]",
	"reason_details": {
		"storage": "20gb",
		"memory": null,
		"nodes": 3
	}
}
```

`20gb` leads to a request of `21474836480` bytes per node.

thbkrkr (approved):
:+1:

Issue #4479, opened 2021-05-06
Enable Elastic Maps Server E2E tests for Stack version 8.x and 7.14.x


thbkrkr:
In #4477, we disabled Elastic Maps Server E2E tests when running tests against snapshot builds of 8.x and 7.14.x

This issue is a placeholder to remember to enable these tests once a Docker image is available for Elastic Maps Server in version 8.x. and/or as soon as a snapshot build is available for 7.14.

Relates to #4476.

PR #4477, opened 2021-05-05
Skip Elastic Maps Server tests for Stack version 8.x

thbkrkr:
This commit adds a check to skip Elastic Maps Server tests when the Stack major version is equal or greater than 8.

It made me want to factorize the code that mades the skips.

Resolves #4476.

PR #4535, opened 2021-05-28
E2E: Elastic Maps Server does not offer snapshot builds at the moment

pebrc:
This disables the Elastic Maps Server E2E test in the two snapshot E2E test we run.


thbkrkr:
This is smarter and more generic than what I did in #4477. I feel like we could revert #4477 and close #4479?
A little comment might be helpful as this `!ems` may seem obscure in a few months.

pebrc:
>  I feel like we could revert #4477 and close #4479?

Agreed to revert #4477 but I would keep #4479 open otherwise we will forget to re-enable.

thbkrkr (approved):

PR #4537, opened 2021-05-31
E2E: go build constraints are OR'ed. Exclude EMS tests

pebrc:
I misunderstood how build constraints worked (and missed the EMS tests flying past my terminal when testing with `go test -test.list ...` ) https://golang.org/cmd/go/#hdr-Build_constraints 

Build constraints are either OR'ed or AND'ed (if on separate lines, or joined with a comma). Neither helps with excluding the EMS tests. This PR list all the constraints we want to build instead. It also adds a new constraint for the top level tests that test a combination of stack applications (`mixed`).

### DIFF
--- a/.ci/setenvconfig
+++ b/.ci/setenvconfig
@@ -484,7 +484,7 @@ IMG_SUFFIX=-dev-ci
 REGISTRY_NAMESPACE=eck-snapshots
 E2E_REGISTRY_NAMESPACE=eck-ci
 # Disable EMS tests for snapshot builds see https://github.com/elastic/cloud-on-k8s/issues/4479
-E2E_TAGS=!ems e2e
+E2E_TAGS=es kb apm beat agent ent mixed
 IS_SNAPSHOT_BUILD=yes
 GO_TAGS=release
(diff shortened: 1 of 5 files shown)


pebrc:
Jenkins test this please

thbkrkr (approved):
I'm fine with this approach which has the little defect of requiring us to be vigilant when adding a new tag to also add it in this list, but we add new tags very rarely so it seems to me reasonable.

Another alternative could be to add a new tag to describe the architectures in addition to the component and always required the two. It's a bit more intrusive. 

```js
# test/es/*.go
// +build e2e es
// +build arch_amd64 arch_arm

# test/kb/*.go
// +build e2e kb
// +build arch_amd64 arch_arm

# test/ems/*.go
// +build e2e ems
// +build arch_amd64
```
```sh
go build -tags 'e2e arch_amd'
> es
> kb
> ems

go build -tags 'e2e arch_arm'
> es
> kb
```

pebrc:
I am merging this for now to avoid the test failure tomorrow. We can look into your idea separately.

PR #4540, opened 2021-05-31
node with specific type has only one node

bsturg:
A customer was performing maintenance where there was a single ingest node and there was confusion in why pod disruption was going yellow.
.
There is a note about single node clusters, but not single node instance types.

- Have you followed the [contributor guidelines](https://github.com/elastic/cloud-on-k8s/tree/master/CONTRIBUTING.md)?
- If you submit code, is your pull request against master? We recommend pull requests against master. We will backport them as needed.

### DIFF
--- a/docs/orchestrating-elastic-stack-applications/elasticsearch/pod-disruption-budget.asciidoc
+++ b/docs/orchestrating-elastic-stack-applications/elasticsearch/pod-disruption-budget.asciidoc
@@ -35,6 +35,8 @@ spec:
 
 NOTE: link:https://kubernetes.io/docs/tasks/run-application/configure-pdb/#arbitrary-controllers-and-selectors[`maxUnavailable`] cannot be used with an arbitrary label selector, therefore `minAvailable` is used in this example.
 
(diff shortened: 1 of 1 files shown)


thbkrkr:
> A customer was performing maintenance where there was a single ingest node and there was confusion in why pod disruption was going yellow.

Pod disruption has no color? Do you mean "why the default pod disruption set by ECK made the cluster health yellow?

About the note you propose to add:

> If there is a node type where there is only one node instance such as ingest, then when that node is restarted, the cluster will go yello.

I do not understand why restarting a ingest node would turn the cluster health to yellow. Can you share a manifest to reproduce this?

Test a version upgrade to the 7.13.0 with this manifest and you will see that the cluster health remains green throughout the rolling upgrade.

```yaml
kind: Elasticsearch
metadata:
  name: test
spec:
  version: 7.12.0
  nodeSets:
  - name: master
    count: 2
    config:
      node.store.allow_mmap: false
  - name: ingest
    count: 1
    config:
      node.roles: ["ingest"]
      node.store.allow_mmap: false
```

thbkrkr:
Closing we have clarified this offline.

PR #4543, opened 2021-06-01
E2E: Elastic Maps Server revert to explicit snapshot check

pebrc:
After tripping over go build constraints and our `setenvconfig` script twice, this goes back to an explicit check for snapshot versions in the Elastic Maps Server builder to skip any tests on SNAPSHOT candidates. 🤞

### DIFF
--- a/test/e2e/test/version.go
+++ b/test/e2e/test/version.go
@@ -45,3 +45,7 @@ func isValidUpgrade(from string, to string) (bool, error) {
 	validMajorDigit := dstVer.Major == srcVer.Major || dstVer.Major == srcVer.Major+1
 	return validMajorDigit && !srcVer.GTE(dstVer), nil
 }
+
+func IsSnapshotVersion(v version.Version) bool {
+	return len(v.Pre) > 0
+}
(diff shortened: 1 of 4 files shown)


test/e2e/test/version.go
```diff
@@ -45,3 +45,7 @@ func isValidUpgrade(from string, to string) (bool, error) {
 	validMajorDigit := dstVer.Major == srcVer.Major || dstVer.Major == srcVer.Major+1
 	return validMajorDigit && !srcVer.GTE(dstVer), nil
 }
+
+func IsSnapshotVersion(v version.Version) bool {
+	return len(v.Pre) > 0
```
thbkrkr, line 50: I was curious to know why it's called `Pre` and it's for [pre-release](https://pkg.go.dev/github.com/blang/semver/v4?utm_source=godoc#PRVersion). Thus `Snapshot` could be semantically inaccurate (e.g.: 7.12.0-bc1) but we don't use this kind of stack version, only snapshots so I think it's clearer in our context like this.

thbkrkr (approved):
Explicit. Simple. LGTM.

test/e2e/test/version.go
pebrc: Yes that was the intention, I kept it in the test package for that reason as well as it is not universally correct. We could improve it an inspect the `Pre` bits for the word snapshot but I think it is probably overkill at this point.

pebrc:
Jenkins test this please

PR #4548, opened 2021-06-02
Fix typo in OperatorHub CSV template

thbkrkr:
`Beats` was mentionned twice in the second description and
 I updated the first description to use the same order in the product list.


barkbay (approved):
lgtm

PR #4567, opened 2021-06-12
Update upgrading-stack.asciidoc

Leaf-Lin:
Adding a description of the expected order of stack version upgrade.

The page currently talks about "Elasticsearch" upgrade only even though the page is titled "Upgrade the Elastic Stack version".

It would be good to clarify by adding the kibana, apm, enterprise-search etc should be upgraded after elasticsearch upgrade is successful. Feel free to add more description to that effect.

### DIFF
--- a/docs/orchestrating-elastic-stack-applications/upgrading-stack.asciidoc
+++ b/docs/orchestrating-elastic-stack-applications/upgrading-stack.asciidoc
@@ -11,4 +11,6 @@ The operator can safely perform upgrades to newer versions of the various Elasti
 
 Follow the instructions in the link:https://www.elastic.co/guide/en/elastic-stack/current/upgrading-elastic-stack.html[Elasticsearch documentation]. Make sure that your cluster is compatible with the target version, take backups, and follow the specific upgrade instructions for each resource type, especially the order in which the upgrade should be carried out. When you are ready, modify the `version` field in the resource spec to the desired stack version and the operator will start the upgrade process automatically.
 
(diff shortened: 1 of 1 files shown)


docs/orchestrating-elastic-stack-applications/upgrading-stack.asciidoc
```diff
@@ -11,4 +11,6 @@ The operator can safely perform upgrades to newer versions of the various Elasti
 
 Follow the instructions in the link:https://www.elastic.co/guide/en/elastic-stack/current/upgrading-elastic-stack.html[Elasticsearch documentation]. Make sure that your cluster is compatible with the target version, take backups, and follow the specific upgrade instructions for each resource type, especially the order in which the upgrade should be carried out. When you are ready, modify the `version` field in the resource spec to the desired stack version and the operator will start the upgrade process automatically.
 
+Once Elasticsearch is upgraded to the newer version with a green cluster health, you will be able to perform the rest of the stack upgrade.
```
alaudazzi, line 14:
```suggestion
When Elasticsearch is upgraded to the newer version with a green cluster health, you can perform the rest of the stack upgrade. 
```

alaudazzi (approved):
Just a minor suggestion, otherwise LGTM.

pebrc (changes requested):
I am not sure the statement added in this PR is accurate. ECK will automatically make sure that the correct upgrade order is adhered to if the Elastic stack applications are linked up via the `*Ref` mechanism (exception: patch version upgrades) This statement is only true if users chose to manage the connection between say Kibana and Elasticsearch themselves. But I am not sure how many users are doing that because the would give up on one of main benefits of using ECK in the first place.

PR #4569, opened 2021-06-15
Detect admission registration API version

### LINKED ISSUE #4555: Webhook certificate reconciliation relies on deprecated v1beta1 API
When reconciling the certificates for the webhook client we rely on the deprecated v1beta1 version of the `ValidatingWebhookConfiguration` API. This API will be unavailable in v1.22 

Moving to v1 would break compatibility with older versions of Kubernetes. As of ECK version 1.6. we only support K8s 1.16+ but we also need to support OpenShift 3.11.

We could: 
* just use v1 and document the webhook is not supported on older versions. 
* call both versions for as long as we have the 1.21 API available (controller-runtime 0.9.0 moves to 1.21)

barkbay:
Fix #4555

I rebased this PR on #4489 and tested it on the following environments:

* OCP 3.11 with and without the admission controller enabled
* EKS 1.16
* GKE 1.18
* Kind running K8S v1.22.0-alpha.3

Unit testing is limited to `v1` because of the way the fake client returns errors when a resource is not found, I opened https://github.com/kubernetes/kubernetes/pull/102872 to improve this.

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -748,13 +748,20 @@ func setupWebhook(mgr manager.Manager, certRotation certificates.RotationParams,
 			Rotation:   certRotation,
 		}
 
-		// Force a first reconciliation to create the resources before the server is started
-		if err := webhookParams.ReconcileResources(context.Background(), clientset); err != nil {
+		// retrieve the current webhook configuration interface
+		wh, err := webhookParams.NewAdmissionControllerInterface(context.Background(), clientset)
+		if err != nil {
 			log.Error(err, "unable to setup and fill the webhook certificates")
 			os.Exit(1)
 		}
 
-		if err := webhook.Add(mgr, webhookParams, clientset); err != nil {
+		// Force a first reconciliation to create the resources before the server is started
+		if err := webhookParams.ReconcileResources(context.Background(), clientset, wh); err != nil {
+			log.Error(err, "unable to setup and fill the webhook certificates")
+			os.Exit(1)
+		}
+
+		if err := webhook.Add(mgr, webhookParams, clientset, wh); err != nil {
 			log.Error(err, "unable to create controller", "controller", webhook.ControllerName)
 			os.Exit(1)
 		}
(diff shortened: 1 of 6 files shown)


cmd/manager/main.go
```diff
@@ -748,13 +748,20 @@ func setupWebhook(mgr manager.Manager, certRotation certificates.RotationParams,
 			Rotation:   certRotation,
 		}
 
+		// retrieve the current webhook configuration interface
+		wh, err := webhookParams.NewAdmissionControllerInterface(context.Background(), clientset)
+		if err != nil {
+			log.Error(err, "unable to setup and fill the webhook certificates")
```
pebrc, line 754: Not from your PR but "fill" sounds a bit odd to me, maybe setup is enough?

pkg/controller/webhook/admission_registration.go
```diff
@@ -0,0 +1,162 @@
...
+}
+
+type Services map[types.NamespacedName]struct{}
+
+// AdmissionControllerInterface helps to setup webhooks for different versions of the admissionregistration API.
+type AdmissionControllerInterface interface {
+	getType() client.Object
+	// services returns the set of services used by the Webhooks
+	services() Services
+	// webhooks returns the list of webhook in the configuration
```
pebrc, line 31:
```suggestion
	// webhooks returns the list of webhooks in the configuration
```

pkg/controller/webhook/admission_registration.go
```diff
@@ -0,0 +1,162 @@
...
+	webhooks() []webhook
+	// update ca bundle with the provided CA in all the Webhooks
```
pebrc, line 33:
```suggestion
	// update CABundle with the provided CA in all the Webhooks
```
not sure. or "CA bundle" maybe

pkg/controller/webhook/admission_registration.go
```diff
@@ -0,0 +1,162 @@
...
+		if err != nil {
+			// 404 is also considered as an error, webhook configuration is expected to be created before the operator is started
+			return nil, err
+		}
+		return &v1beta1webhookHandler{ctx: ctx, clientset: clientset, webhookConfiguration: webhookConfiguration}, nil
+	}
+	if err != nil {
+		return nil, err
+	}
+	log.V(1).Info(" using admissionregistration.k8s.io/v1 for webhook configuration")
```
pebrc, line 52:
```suggestion
	log.V(1).Info("using admissionregistration.k8s.io/v1 for webhook configuration")
```

pebrc (approved):
LGTM! I did a few simple tests with kind and it looks good 👍

Issue #4522, opened 2021-05-26
Integrate custom CA for http interface


MHenn1g:
I have a request regarding the certificate management for the http certificate management.

The basic construct is as follows.
- ECK version: 1.5.0
- Multiple Elasticsearch Deployments, each deployed automatically via Helm through a deployment tool like Argo CD
- Apps need to communicate to the Elasticsearch Deployments via the LoadBalancer IP
- A central enterprise CA is integrated in the Apps trust store.
  Customers should not need to generate their own certificates.
  Instead, this CA should be utilized during the creation of the Cluster deployments to automatically enable a trusted communication.

As I take the documentation there are currently multiple ways to utilize certificates for the http interface in order to communicate with an Elasticsearch installation based on ECK:
1. Utilize a default self signed certificate (https://www.elastic.co/guide/en/cloud-on-k8s/master/k8s-tls-certificates.html#k8s-default-self-signed-certificate)
2. Setup your own certificate (https://www.elastic.co/guide/en/cloud-on-k8s/master/k8s-tls-certificates.html#k8s-setting-up-your-own-certificate)
3. Use a third party tool like the cert-manager (https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-custom-http-certificate.html#k8s_custom_self_signed_certificate_using_cert_manager)

Since developers should not generate their own certificates, option 2 is out of the question for our use case.

When utilizing version 1, using a self signed certificate, the operator generates a certificate including a whole bunch of SANs out of the box:
	
	Subject Alternative Names:  
		quickstart-eck-es-http.ns.es.local,
		quickstart-eck-es-http,  
		quickstart-eck-es-http.ns.svc,
		quickstart-eck-es-http.ns, quickstart-eck-es-http,
		*.quickstart-eck-es-control-plane-az1.ns.svc,
		*.quickstart-eck-es-control-plane-az2.ns.svc,
		*.quickstart-eck-es-control-plane-az3.ns.svc,
		*.quickstart-eck-es-data-plane-az1.ns.svc,
		*.quickstart-eck-es-data-plane-az2.ns.svc,
		*.quickstart-eck-es-data-plane-az3.ns.svc,
		IP Address:192.168.0.121
Especially the IP Adress of the LoadBalancer gets injected directly, which is pretty handy, since this may be utilized to directly communicate through the service.

However, this only works as long as self-signed certificates are utilized.
[...]

PR #4575, opened 2021-06-16
Support custom CA on HTTP layer 

pebrc:
Fixes https://github.com/elastic/cloud-on-k8s/issues/4522

Allows users to add a full CA certificate in `ca.crt` and `ca.key` which will be used to sign the leaf certificate that ECK will use for the Elastic stack applications. 

Awkwardly the options we offer in `selfSignedCertificate` can then still be used to modify the certificate. Awkwardly, because that certificate is no longer necessarily self-signed, depending on the CA the user has provided. But I don't think we want to break the stable API we have for that. 

This also changes the naming convention for the transport layer CA user can specify for Elasticsearch to follow the new `ca.crt`, `ca.key` scheme. The old `tls.key` and `tls.crt` names are still supported for backwards compatibility.

Docs preview: https://cloud-on-k8s_4575.docs-preview.app.elstc.co/diff


pkg/controller/common/certificates/secret.go
```diff
@@ -63,29 +66,78 @@ func HTTPCertSecretVolume(namer name.Namer, name string) volume.SecretVolume {
...
+func (s *CertificatesSecret) HasLeafCertificate() bool {
+	return s != nil && !s.HasFullCA()
+}
+
+func (s *CertificatesSecret) validateCustomCA() error {
+	// flag up user error when specifying both CA certificate with key and leaf certificate
+	_, tlsKeyExists := s.Data[KeyFileName]
+	_, tlsCertExists := s.Data[CertFileName]
+	if tlsKeyExists || tlsCertExists {
+		return fmt.Errorf("cannot specify both %s and %s or %s in %s/%s",
```
david-kow, line 123: Took me a second to understand what `and` and `or` operate on :) Maybe something like below (+ change arguments order)?

```suggestion
		return fmt.Errorf("cannot specify %s or %s when %s is set in %s/%s",
```

pkg/controller/common/certificates/secret.go
```diff
@@ -63,29 +66,78 @@ func HTTPCertSecretVolume(namer name.Namer, name string) volume.SecretVolume {
...
+			CAKeyFileName, KeyFileName, CertFileName, s.Namespace, s.Name)
+	}
+	ca, err := parseCAFromSecret(s.Secret, CAKeyFileName, CAFileName)
+	if err == nil {
+		// breaking the validation contract here by remembering the results to avoid parsing everything once more
```
david-kow, line 128: Change the func name (and callers) then? ie. `parseCustomCA`?

test/e2e/es/certs_test.go
```diff
@@ -32,24 +32,133 @@ import (
...
+						toDelete := mkCertSecret(nil, nil)
+						_ = k.Client.Delete(context.Background(), &toDelete)
+					},
+				},
+			}
+		},
+	}
+
+	// The initial Cluster builder should result in a healthy cluster as verified by the standard check steps
+	// Now modify cluster to use custom HTTP CA  but simulate a user error by populating the secret with
```
david-kow, line 66:
```suggestion
	// Now modify cluster to use custom HTTP CA but simulate a user error by populating the secret with
```

test/e2e/test/elasticsearch/builder.go
```diff
@@ -448,6 +454,11 @@ func (b Builder) WithPodLabel(key, value string) Builder {
 	return b
 }
 
+func (b Builder) WithToleratedHealthCheckErrors(msgs ...string) Builder {
+	b.toleratedHealthCheckErrors = msgs
```
david-kow, line 458: Should we append instead? If one chains those only the last will be in effect which might be surprising.

pkg/controller/common/certificates/ca_secret.go
```diff
@@ -12,23 +12,36 @@ import (
 // ParseCustomCASecret checks that mandatory fields are present and returns a CA struct.
 // It does not check that the public key matches the private key.
 func ParseCustomCASecret(s corev1.Secret) (*CA, error) {
+	keyFileName := CAKeyFileName
```
david-kow, line 15: Should this function name (or at least godoc) indicate allowing legacy key naming? I'm not sure how one decides calling `ParseCustomCASecret` vs `parseCAFromSecret` (for example from `validateCustomCA`).

david-kow (approved):
LGTM, few nits here and there.

docs/orchestrating-elastic-stack-applications/accessing-elastic-services.asciidoc
```diff
@@ -148,10 +149,22 @@ Create a Kubernetes secret with:
 
 [source,sh]
 ----
-kubectl create secret generic my-cert --from-file=ca.crt=tls.crt --from-file=tls.crt=tls.crt --from-file=tls.key=tls.key
+kubectl create secret generic my-cert --from-file=ca.crt --from-file=tls.crt --from-file=tls.key
```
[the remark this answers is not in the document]
pebrc: I was honestly confused by the original example. Why would one want to specify the leaf certificate as the CA certificate? If the leaf certificate were also the CA at the same time we would not need to specify it. So I am thinking the original example was done by mistake. At the very least it would have required some additional explanation. So I made this change on purpose to reflect what I would think is a typical use case.

pkg/controller/common/certificates/ca_secret.go
pebrc: Good point. I actually want to hide the `parseCAFromSecret` one as an internal function that just avoids duplication between the different parsing functions and meant `ParseCustomCASecret` as the actual API.

pebrc:
The e2e test I added is flaky. I will need to fix that before merging this.

alaudazzi (approved):
LGTM

pkg/controller/common/certificates/secret.go
```diff
@@ -18,11 +19,13 @@ import (
 )
 
 const (
-	// CAFileName is used for the CA Certificates inside a secret
+	// CAFileName is used for the CA Certificates inside a secret.
 	CAFileName = "ca.crt"
-	// CertFileName is used for Certificates inside a secret
+	// CAKeyFileName is used for CA certficiates' private keys inside a secret.
```
barkbay, line 24:
```suggestion
	// CAKeyFileName is used for CA certificate's private keys inside a secret.
```

pkg/controller/common/certificates/secret.go
```diff
@@ -63,29 +66,78 @@ func HTTPCertSecretVolume(namer name.Namer, name string) volume.SecretVolume {
...
 	return s.Data[CertFileName]
 }
 
-func (s CertificatesSecret) KeyPem() []byte {
+func (s *CertificatesSecret) KeyPem() []byte {
 	return s.Data[KeyFileName]
 }
 
-// Validate checks that mandatory fields are present.
+func (s *CertificatesSecret) HasFullCA() bool {
```
barkbay, line 105: nit: at first I was thinking that a "full CA" is a CA with all the intermediate certificates. Should we name this function `HasPrivateKey` to make its purpose more explicit ?

docs/orchestrating-elastic-stack-applications/accessing-elastic-services.asciidoc
```diff
@@ -148,10 +149,22 @@ Create a Kubernetes secret with:
...
 ----
 
-Then, you have to reference the secret name in the `http.tls.certificate` section of the resource manifest.
+Alternatively you can also bring your own CA certificate including a private key and let ECK issue certificates with it.
```
barkbay, line 155: nit: I think I would add a comment to explain how `selfSignedCertificate` is used in the generated certificates.

barkbay (commented):
Just left a few comments after a first read, I did not test it extensively.

pkg/apis/elasticsearch/v1/elasticsearch_types.go
```diff
@@ -99,8 +99,8 @@ type TransportTLSOptions struct {
 	// and private key for generating node certificates.
 	// The referenced secret should contain the following:
 	//
-	// - `tls.crt`: The CA certificate in PEM format.
-	// - `tls.key`: The private key for the CA certificate in PEM format.
```
barkbay, line 103: Should we keep these 2 lines ?  My understanding is that `tls.crt` and `tls.key` are still required if `ca.key` is not provided.

pebrc: This is for the transport layer. I replaced the `tls.crt` `tls.key` keys (but offer backwards compatibility for existing users). Your statement is true for the HTTP layer.

docs/orchestrating-elastic-stack-applications/accessing-elastic-services.asciidoc
pebrc: This is actually one of the not so great aspects of this change. We will have a an attribute `selfSignedCertificate.subjectAltNames` that affects also certificates that are not actually self-signed.

pkg/apis/elasticsearch/v1/elasticsearch_types.go
barkbay: My bad, I missed the `TransportTLSOptions` in the diff header.

test/e2e/es/certs_test.go
```diff
@@ -32,24 +32,133 @@ import (
...
+	// A NOOP mutation but set up the CA secret correctly now (using the existing CA generation code in the operator)
+	withCustomCert := test.WrappedBuilder{
+		BuildingThis: initialCluster.
+			WithCustomHTTPCerts(caSecretName).
+			WithMutatedFrom(&initialCluster),
+
+		PreMutationSteps: func(k *test.K8sClient) test.StepList {
+			return test.StepList{
+				{
+					Name: "Create custom CA secret",
```
barkbay, line 122:
```suggestion
					Name: "Update custom CA secret with valid CA",
```

test/e2e/es/certs_test.go
```diff
@@ -32,24 +32,133 @@ import (
...
+							certificates.EncodePEMCert(ca.Cert.Raw),
+							certificates.EncodePEMPrivateKey(*ca.PrivateKey),
+						)
+						_, err = reconciler.ReconcileSecret(k.Client, caSecret, nil)
+						return err
+					}),
+				},
+			}
+		},
+		// verification that the certificate works is implicit by the standard check steps that all require an Elasticsearch
```
barkbay, line 145: How do we know from here that the operator has reconciled the cluster and the updated custom Secret is in use ?

pebrc: I added a new check step called `CheckHTTPConnectivity` which should create a test request to each Pod. But you are right that there is a gap in the logic in that this step relies on the `*-es-certs-public` secret being reconciled and then uses the CA from that secret to make the connections.

pebrc: I think could add an explicit step that verifies with the custom CA.

test/e2e/es/certs_test.go
```diff
@@ -32,24 +32,133 @@ import (
...
+				test.Step{
+					Name: "Invalid CA secret should create events",
+					Test: test.Eventually(func() error {
+						eventList, err := k.GetEvents(test.EventListOptions(esNamespace, initialCluster.Elasticsearch.Name)...)
+						if err != nil {
+							return err
+						}
+						for _, evt := range eventList {
+							if evt.Type == corev1.EventTypeWarning &&
+								evt.Reason == events.EventReconciliationError &&
```
barkbay, line 98: I would include a namespaced name comparison to check that the event has not been generated by a previous run (it happened on my dev environment), or a similar test that we may add in the future.
```suggestion
								evt.Reason == events.EventReconciliationError &&
								evt.InvolvedObject.Namespace == b.Elasticsearch.Namespace &&
								evt.InvolvedObject.Name == b.Elasticsearch.Name &&
```

test/e2e/es/certs_test.go
barkbay: Also if I slow down the operator and delay the reconciliation I can see this kind of error:

```
    cluster_uuid.go:57: 
        	Error Trace:	cluster_uuid.go:57
        	            	Get "https://test-custom-http-ca-2fxk-es-http.e2e-mercury.svc:9200/": x509: certificate signed by unknown authority
        	Test:       	TestCustomHTTPCA/Cluster_UUID_should_have_been_preserved#01
```

Not sure if it is a realistic experiment though...

pebrc: Slowing down means you also delayed the certificate secret reconciliation to run into the condition I [mentioned](https://github.com/elastic/cloud-on-k8s/pull/4575#discussion_r659671813)?

barkbay: Yes, I think `CheckHTTPConnectivity` was run before the certificate was reconciled:
 
image

You're right, it is just another manifestation of the same condition. Sorry, I'm just trying to understand if this test could become a flaky one once deployed 🙃

pebrc: No need to be sorry. Very grateful for your thorough review!

barkbay: I just checked out last commit ([5e29d98](https://github.com/elastic/cloud-on-k8s/pull/4575/commits/5e29d98ac5c11745d9ccb5649a9e08f54baa1ac6)), it looks good but I think we need a similar step to validate that at step `reconfigure back to self-signed operator provided CA` the custom CA is no more in use.

test/e2e/es/certs_test.go
```diff
@@ -32,24 +32,161 @@ import (
...
+							certificates.EncodePEMCert(customCA.Cert.Raw),
+							certificates.EncodePEMPrivateKey(*customCA.PrivateKey),
+						)
+						_, err = reconciler.ReconcileSecret(k.Client, caSecret, nil)
+						return err
+					}),
+				},
+			}
+		},
+		PostMutationSteps: func(k *test.K8sClient) test.StepList {
```
barkbay, line 165: Any reason for having this check as a `PostMutationStep` ? I think we should expect the CA to be available right after the update, something along those lines:

```golang
PreMutationSteps: func(k *test.K8sClient) test.StepList {
    return test.StepList{
        {
            Name: "Update custom CA secret",
                var err error
                customCA, err = certificates.NewSelfSignedCA(certificates.CABuilderOptions{
                    Subject: pkix.Name{
                        CommonName:         "eck-e2e-test-custom-ca",
                        OrganizationalUnit: []string{"eck-e2e"},
                    },
                })
                if err != nil {
                    return err
                }

                caSecret := mkCertSecret(
                    certificates.EncodePEMCert(customCA.Cert.Raw),
                    certificates.EncodePEMPrivateKey(*customCA.PrivateKey),
                )
                _, err = reconciler.ReconcileSecret(k.Client, caSecret, nil)
                return err
            }),
        },
        {
            // Updated CA should be exposed almost immediately 
            Name: "Verify that the custom CA is in use",
                return elasticsearch.CheckHTTPConnectivityWithCA(initialCluster.Elasticsearch, k, []*x509.Certificate{customCA.Cert})
            }),
        },
    }
},
```

pkg/controller/common/certificates/secret_test.go
```diff
@@ -83,75 +113,139 @@ func TestCertificatesSecret_Validate(t *testing.T) {
...
+					Data: map[string][]byte{
+						CertFileName: tls,
+						KeyFileName:  corruptedKey,
+					},
+				},
+			},
+			wantErr: true,
+		},
+		{
+			name: "Full custom CA",
```
barkbay, line 195:
```suggestion
			name: "With CA private key",
```

barkbay (approved):
LGTM 👍

test/e2e/es/certs_test.go
pebrc: Good point!

pebrc: I guess the only reason would be that it allows you to switch the sequence of the builders without breaking the test.

pebrc:
run full pr build

pebrc:
@barkbay I reworked the e2e test to avoid spurious test failures. It is much simpler now and should be stable  🤞 .  If you could take another quick look that would be great 🙇

test/e2e/es/certs_test.go
```diff
@@ -32,23 +32,127 @@ import (
...
+	steps = steps.WithSteps(test.StepList{
+		{
+			Name: "Ensure built-in CA is used after removing custom CA",
+			Test: test.Eventually(func() error {
+				ca, err := k.GetCA(initialCluster.Elasticsearch.Namespace, initialCluster.Elasticsearch.Name, certificates.HTTPCAType)
+				if err != nil {
+					return err
+				}
+				if customCA != nil && customCA.Cert.Equal(ca.Cert) {
+					return errors.New("Still using custom CA cert")
```
barkbay, line 135:
```suggestion
					return errors.New("still using custom CA cert")
```

test/e2e/test/elasticsearch/checks_http.go
```diff
@@ -0,0 +1,53 @@
...
+	var dialer net.Dialer
+	if test.Ctx().AutoPortForwarding {
+		dialer = portforward.NewForwardingDialer()
+	}
+	v, err := version.Parse(es.Spec.Version)
+	if err != nil {
+		return err
+	}
+
+	for _, p := range reconcile.AvailableElasticsearchNodes(pods) {
```
barkbay, line 44: While re-reading this check I was wondering if we could "miss" some Pods, either because some of them are being recreated or because they are not ready.
I guess that if we do not want to make any progress until all the expected Pods are exposing the desired CA the algorithm would be:

1. Get the actual nodes, if at least one is not available/ready return an error
2. Get all the expected nodes, if the list does not match the actual one return an error
3. For each node check the connectivity using the desired CA

This situation is very unlikely to happen though, so I'm not sure if it's worth to improve this check.

test/e2e/es/certs_test.go
```diff
@@ -32,23 +32,127 @@ import (
...
+	// reference to custom CA for comparison purposes
+	var customCA *certificates.CA
+
+	// tests the following sequence:
+	// 1. healthy cluster with self-signed operator provided CA
+	// 2. reconfigure to use custom certs but simulate user error on certificate setup: no rollout expected
+	// 3. reconfigure with correct custom certificates
+	// 4. reconfigure back to self-signed operator provided CA
+	k := test.NewK8sClientOrFatal()
+	steps := test.StepList{}
```
barkbay, line 59: cosmetic nit: the `steps` variable can be avoided

```suggestion
	test.StepList{}.
		WithSteps(initialCluster.InitTestSteps(k)).
		WithSteps(initialCluster.CreationTestSteps(k)).
		WithSteps(test.CheckTestSteps(initialCluster, k)).
		WithStep(test.Step{
			Name: "Create an invalid CA secret",
			Test: test.Eventually(func() error {
				bogusSecret := mkCertSecret([]byte("garbage"), []byte("more garbage"))
				_, err := reconciler.ReconcileSecret(k.Client, bogusSecret, nil)
				return err
			}),
		}).
		WithSteps(withCustomCA.UpgradeTestSteps(k)).
		WithStep(
			test.Step{
				Name: "Invalid CA secret should create events",
				Test: test.Eventually(func() error {
					eventList, err := k.GetEvents(test.EventListOptions(esNamespace, initialCluster.Elasticsearch.Name)...)
					if err != nil {
						return err
					}
					for _, evt := range eventList {
						if evt.Type == corev1.EventTypeWarning &&
							evt.Reason == events.EventReconciliationError &&
							evt.InvolvedObject.Namespace == withCustomCA.Elasticsearch.Namespace &&
							evt.InvolvedObject.Name == withCustomCA.Elasticsearch.Name &&
							strings.Contains(evt.Message, "can't parse") {
							return nil
						}
					}
					return fmt.Errorf("expected validation event but could not observe it")
				})}).
		WithStep(test.Step{
			Name: "Update custom CA secret",
			Test: test.Eventually(func() error {
				var err error
				customCA, err = certificates.NewSelfSignedCA(certificates.CABuilderOptions{
					Subject: pkix.Name{
						CommonName:         "eck-e2e-test-custom-ca",
						OrganizationalUnit: []string{"eck-e2e"},
					},
				})
				if err != nil {
					return err
				}

				caSecret := mkCertSecret(
					certificates.EncodePEMCert(customCA.Cert.Raw),
					certificates.EncodePEMPrivateKey(*customCA.PrivateKey),
				)
				_, err = reconciler.ReconcileSecret(k.Client, caSecret, nil)
				return err
			}),
		}).
		WithStep(test.Step{
			Name: "Verify that the custom CA is in use",
			Test: test.Eventually(func() error {
				return elasticsearch.CheckHTTPConnectivityWithCA(initialCluster.Elasticsearch, k, []*x509.Certificate{customCA.Cert})
			}),
		}).
		// "upgrade" back to the initial cluster without custom CA
		WithSteps(initialCluster.UpgradeTestSteps(k)).
		WithSteps(test.StepList{
			{
				Name: "Ensure built-in CA is used after removing custom CA",
				Test: test.Eventually(func() error {
					ca, err := k.GetCA(initialCluster.Elasticsearch.Namespace, initialCluster.Elasticsearch.Name, certificates.HTTPCAType)
					if err != nil {
						return err
					}
					if customCA != nil && customCA.Cert.Equal(ca.Cert) {
						return errors.New("still using custom CA cert")
					}
					return elasticsearch.CheckHTTPConnectivityWithCA(initialCluster.Elasticsearch, k, []*x509.Certificate{ca.Cert})
				}),
			},
			{
				Name: "Clean up custom CA secret",
				Test: func(t *testing.T) {
					// let's clean up the CA secret here.
					toDelete := mkCertSecret(nil, nil)
					_ = k.Client.Delete(context.Background(), &toDelete)
				},
			},
		}).WithSteps(initialCluster.DeletionTestSteps(k)).RunSequential(t)
```

barkbay (approved):
lgtm, checking the CA right after a related change should greatly improve the stability 👍

test/e2e/test/elasticsearch/checks_http.go
pebrc: I think you are right that is unlikely. All upgrades in this test are not leading to actual Pod rotations and we check the readiness of the expected Pods during the initial creation. So I am tempted to leave it as is for now.

PR #4593, opened 2021-06-23
OCP deployer: Remove context, user and cluster from kube config after delete

pebrc:
Repeated deployments of OCP with deployer lead to a broken setup otherwise with certificates from previous incarnations of the cluster mixed up with new configuration.


barkbay (approved):
lgtm

PR #4597, opened 2021-06-29
Use node shutdown API for rolling upgrades and downscales

pebrc:
This adds support for the new shutdown API to the operator. This comes with no escape hatch for users and fully relies on the new API as of version ~7.15.1~ 7.15.2

Tests should be covered by our regular e2e tests when run with 7.15.0-SNAPSHOT where we execute all kinds of Elasticsearch upgrade and up-/downscale scenarios.


pkg/controller/elasticsearch/driver/node_shutdown.go
```diff
@@ -0,0 +1,28 @@
...
+		if err != nil {
+			return nil, err
+		}
+		return shutdown.NewNodeShutdown(client, idLookup, esclient.Remove, es.ResourceVersion), nil
+	}
+	return migration.NewShardMigration(es, client, client), nil
+}
+
+func supportsNodeshutdown(v version.Version) bool {
+	return v.GTE(version.MustParse("7.14.0-SNAPSHOT"))
```
pebrc, line 27: This should be 7.15 but for testing it early with current snapshot builds it needs to be 7.14.x and the feature flag needs to be set in Elasticsearch.

pkg/controller/elasticsearch/driver/upgrade.go
```diff
@@ -236,8 +246,27 @@ func doFlush(ctx context.Context, es esv1.Elasticsearch, esClient esclient.Clien
...
 	}
 }
+func (d *defaultDriver) maybeCompleteNodeUpgrades(
+	ctx context.Context,
+	esClient esclient.Client,
+	esState ESState,
+	nodeShutdown *shutdown.NodeShutdown,
+) *reconciler.Results {
+	// we still have to enable shard allocation in cases where we just upgraded from
+	// a version that did not support node shutdown to a supported version.
```
barkbay, line 256: I was wondering if we can be smarter here, and only enable shard allocation if necessary. But I'm not sure there is an easy way to know that we have just restarted the last node that did not support node shutdown.
It would require to store the previous version (as in `ElasticsearchStatus.Version`) or persist somewhere the fact that shards allocation has been disabled on purpose by the operator.

barkbay (commented):
I just went through the code a first time, left a few nitpicks.
I still need to do some tests and have a better understanding of the shutdown API but it lgtm overall. It's really great to be able to rely on a dedicated API to remove a node instead of handling shards allocations 😄  !

pebrc:
One thing I realised over the last couple of days is that we probably want an operator level flag to opt out of the node shutdown API. The reasoning behind that would be that it is a very new API that may contain bugs and we want to give users an escape hatch to get back to the old behaviour if things go wrong.  Wdyt?

barkbay:
> One thing I realised over the last couple of days is that we probably want an operator level flag to opt out of the node shutdown API. The reasoning behind that would be that it is a very new API that may contain bugs and we want to give users an escape hatch to get back to the old behaviour if things go wrong. Wdyt?

My first feeling is that if something is broken it should be fixed in a new release. That being said I agree it is indeed an important change in the orchestration process and, even if a fix is provided in a new stack release we might need this flag to upgrade the clusters using the old behaviour (assuming there is a bug which prevents the upgrade). So I think I would be 👍 on the idea.
We could also introduce this feature as soon as the shutdown API is available, add the flag but keep it disabled by default to allow early adopters to do some tests or enable it in a dedicated e2e pipeline. Then, once we are more confident, enable it by default with the option for the user to disable it. I think we would still need to deprecate it at some point.

pkg/controller/elasticsearch/shutdown/node.go
```diff
@@ -0,0 +1,141 @@
...
+		}
+		ns.shutdowns = shutdowns
+	})
+	return err
+}
+
+func (ns *NodeShutdown) lookupNodeID(podName string) (string, error) {
+	nodeID, exists := ns.podToNodeID[podName]
+	if !exists {
+		return "", fmt.Errorf("node %s currently not member of the cluster", podName)
```
barkbay, line 65: This might be expected, for example say that:
1. User creates a cluster
2. User makes a change which prevents the Pod to start, for example by setting a wrong `nodeSelector` in the `podTemplate`
3. Pod is deleted by the operator but the new one is stuck in a `Pending` state
4. User makes a new change in order to fix the configuration
5. The operator attempts to delete the Pod, but since it never started and has not joined the cluster this error is raised:

```
2021-08-20T09:05:47.071+0200    ERROR   manager.eck-operator.controller.elasticsearch-controller        Reconciler error        {"service.version": "1.8.0-SNAPSHOT+mmo", "name": "quickstart", "namespace": "default", "error": "node quickstart-es-data-hot-1 currently not member of the cluster", "errorCauses": [{"error": "node quickstart-es-data-hot-1 currently not member of the cluster"}]}
```

This situation would require the user to manually delete the Pod.

I think it is fine to consider that we can delete a Pod if it has not joined the cluster. We might hit a situation where the Pod has actually started, maybe with a ML job already running on it, but as long as the operator lives in the past I'm not sure there is a great solution to this edge case.

pebrc: That's a really good find, I had not considered this case! I think we have to distinguish between upgrade and downscale here though.

During downscale operations we don't want to ignore this error, even if the Pod is currently not part of the cluster because it cannot be scheduled it might have been at some time and have data worth migrating on a (local) disk? If we just went ahead and deleted the Pod the volume claim would be deleted as well.

During upgrades going ahead with the deletion should be fine as we can assume that the upgrade/configuration change could potentially unblock the situation. 

Do you agree?

barkbay: > During downscale operations we don't want to ignore this error [...]

Good point 👍 

> Do you agree?

Yes !

pebrc:
> java.lang.IllegalStateException: JENKINS-37121: something already locked /var/lib/jenkins/workspace/cloud-on-k8s-pr

pebrc:
I am closing this as we have no immediate intention to merge this. To be re-opened when the time comes.

pkg/controller/elasticsearch/driver/upgrade.go
pebrc: We could track state in an annotation on a best-effort basis. Once we migrated a cluster fully to node shutdown we would set the annotation and when it is present we would not try to enable shard allocations anymore.

pebrc:
Jenkins test this please

pkg/controller/elasticsearch/client/model.go
```diff
@@ -401,3 +401,58 @@ type SearchResults struct {
...
+	Reason                string          `json:"reason"`
+	ShutdownStartedMillis int             `json:"shutdown_started_millis"`
+	Status                ShutdownStatus  `json:"status"`
+	ShardMigration        ShardMigration  `json:"shard_migration"`
+	PersistentTasks       PersistentTasks `json:"persistent_tasks"`
+	Plugins               Plugins         `json:"plugins"`
+}
+
+func (ns NodeShutdown) Is(t ShutdownType) bool {
+	//	 API returns type in capital letters currently
```
thbkrkr, line 447:
```suggestion
	// API returns type in capital letters currently
```

pkg/controller/elasticsearch/driver/upgrade.go
```diff
@@ -296,24 +329,56 @@ func (d *defaultDriver) MaybeEnableShardsAllocation(
 	return results
 }
 
-func (ctx *rollingUpgradeCtx) prepareClusterForNodeRestart(esClient esclient.Client, esState ESState) error {
+func (ctx *rollingUpgradeCtx) readyToDelete(pod corev1.Pod) (bool, error) {
+	if !supportsNodeShutdown(ctx.esClient.Version()) {
+		return true, nil // always OK to restart pre 7.14
```
thbkrkr, line 334: The comment may be confusing as we have decided to support node shutdown from version 7.15.1.
```suggestion
		return true, nil // always OK to restart pre node shutdown support
```

pkg/controller/elasticsearch/shutdown/node.go
```diff
@@ -0,0 +1,156 @@
...
+	podToNodeID map[string]string
+	shutdowns   map[string]esclient.NodeShutdown
+	once        sync.Once
+	log         logr.Logger
+}
+
+var _ Interface = &NodeShutdown{}
+
+// NewNodeShutdown creates a new NodeShutdown struct restricted to one type of shutdown (typ); podToNodeID is mapping from
+// K8s Pod name to Elasticsearch node ID; reason is a arbitrary bit of metadata that will be attached to each node shutdown
```
thbkrkr, line 32:
```suggestion
// K8s Pod name to Elasticsearch node ID; reason is an arbitrary bit of metadata that will be attached to each node shutdown
```

pebrc:
run full pr build

pebrc:
run full pr build

barkbay (commented):
Almost LGTM, great work 👍 I still want to do a few tests but I'm almost done.

My only concern is that it is hard to understand the flow of the events from the logs because we only have the node id in ~the message~ some messages, not the node name.
_Edit: some of the logs actually have the node names, it's missing in the logs which reflect the status from the shutdown API_

For example:

```json
2021-11-16T10:57:49.099+0100    DEBUG   driver  Existing shutdown       {"service.version": "my-snapshot", "namespace": "default", "es_name": "elasticsearch", "node_id": "T2g1TUfzSAq8BCRMLJn2XQ", "type": "REMOVE", "status": "COMPLETE"}
```

From a human point of view it's not easy to understand which node exactly has been removed here. But my understanding is that this information is not available in the shutdown API 🤷

pkg/controller/elasticsearch/shutdown/node_test.go
```diff
@@ -0,0 +1,385 @@
...
+			fixture: shutdownFixture,
+			args: args{
+				podToNodeID: map[string]string{
+					"pod-1": "txXw-Kd2Q6K0PbYMAPzH-Q",
+				},
+				podName: "pod-1",
+			},
+			want: NodeShutdownStatus{
+				Status:      esclient.ShutdownComplete,
+				Explanation: "",
```
barkbay, line 211: nit: maybe use an API response which includes an explanation if it is expected to have it reported in `NodeShutdownStatus`:

for example:

```golang
var (
	stalledFixture = `{
  "nodes" : [
    {
      "node_id" : "MT7eFuG5QliYII9K2HCgLw",
      "type" : "REMOVE",
      "reason" : "93475",
      "shutdown_startedmillis" : 1637057097778,
      "status" : "STALLED",
      "shard_migration" : {
        "status" : "STALLED",
        "shard_migrations_remaining" : 41,
        "explanation" : "shard [0] [primary] of index [.geoip_databases] cannot move, use the Cluster Allocation Explain API on this shard for details"
      },
      "persistent_tasks" : {
        "status" : "COMPLETE"
      },
      "plugins" : {
        "status" : "COMPLETE"
      }
    },
    {
      "node_id" : "rX6VerB4RUuZVzGSb8uSIw",
      "type" : "REMOVE",
      "reason" : "93475",
      "shutdown_startedmillis" : 1637057097175,
      "status" : "STALLED",
      "shard_migration" : {
        "status" : "STALLED",
        "shard_migrations_remaining" : 26,
        "explanation" : "shard [0] [primary] of index [.ds-metrics-kubernetes.apiserver-default-2021.11.16-000001] cannot move, use the Cluster Allocation Explain API on this shard for details"
      },
      "persistent_tasks" : {
        "status" : "COMPLETE"
      },
      "plugins" : {
        "status" : "COMPLETE"
      }
    }
  ]
}`
)
...
		{
			name:    "successful lookup with explanation",
			fixture: stalledFixture,
			args: args{
				podToNodeID: map[string]string{
					"pod-1": "MT7eFuG5QliYII9K2HCgLw",
				},
				podName: "pod-1",
			},
			want: NodeShutdownStatus{
				Status:      esclient.ShutdownStalled,
				Explanation: "shard [0] [primary] of index [.geoip_databases] cannot move, use the Cluster Allocation Explain API on this shard for details",
			},
			wantErr: false,
		},
```

pkg/controller/elasticsearch/shutdown/node.go
```diff
@@ -0,0 +1,156 @@
...
+
+// ReconcileShutdowns retrieves ongoing shutdowns and based on the given node names either cancels or creates new
+// shutdowns.
+func (ns *NodeShutdown) ReconcileShutdowns(ctx context.Context, leavingNodes []string) error {
+	if err := ns.initOnce(ctx); err != nil {
+		return err
+	}
+	// cancel all ongoing shutdowns
+	if len(leavingNodes) == 0 {
+		return ns.Clear(ctx, nil)
```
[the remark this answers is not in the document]
pebrc: So the concern here is that we accumulate to many completed shutdowns in cluster state before we finally clean them up?

pkg/controller/common/events/events.go
```diff
@@ -12,7 +12,10 @@ const (
 	EventReasonDeleted = "Deleted"
 	// EventReasonDelayed describes events where a requested change was delayed e.g. to prevent data loss.
 	EventReasonDelayed = "Delayed"
-	// EventReasonCreated describes events where resources are upgraded.
+	// EventReasonStalled describes events where a requested change is stalled and cannot make progress without user
```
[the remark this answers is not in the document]
pebrc: Good point and we also want to soften the message in the events we create as this can be a transient situation.

pkg/controller/elasticsearch/shutdown/node.go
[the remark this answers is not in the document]
barkbay: Yes, my first understanding reading the comment is that it would cancel the shutdown for the nodes which are not in the slice:
> [...] based on the given node names either cancels or creates new shutdowns

But maybe I'm overthinking...

pkg/controller/elasticsearch/shutdown/node.go
```diff
@@ -0,0 +1,156 @@
...
+	shutdowns   map[string]esclient.NodeShutdown
+	once        sync.Once
+	log         logr.Logger
+}
+
+var _ Interface = &NodeShutdown{}
+
+// NewNodeShutdown creates a new NodeShutdown struct restricted to one type of shutdown (typ); podToNodeID is mapping from
+// K8s Pod name to Elasticsearch node ID; reason is an arbitrary bit of metadata that will be attached to each node shutdown
+// request in Elasticsearch and can help tracking and auditing shutdown requests.
```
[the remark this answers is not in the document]
naemono: `can help with the tracking and auditing of shutdown requests` maybe?

pkg/controller/elasticsearch/client/client.go
```diff
@@ -103,10 +103,22 @@ type Client interface {
...
+	PutShutdown(ctx context.Context, nodeID string, shutdownType ShutdownType, reason string) error
+	// DeleteShutdown attempts to cancel an ongoing node shutdown.
+	// Introduced in: Elasticsearch 7.14.0
+	DeleteShutdown(ctx context.Context, nodeID string) error
 	// Request exposes a low level interface to the underlying HTTP client e.g. for testing purposes.
 	// The Elasticsearch endpoint will be added automatically to the request URL which should therefore just be the path
 	// with a leading /
 	Request(ctx context.Context, r *http.Request) (*http.Response, error)
+	// Version returns the Elasticsearch version this client is constructed for which should equal the minimal version
+	// in the cluster
```
naemono, line 120:
```suggestion
	// in the cluster.
```

pkg/controller/elasticsearch/shutdown/node.go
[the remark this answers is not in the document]
pebrc: I went with `and can help to track and audit shutdown requests.` hope that works too?

pkg/controller/elasticsearch/client/model.go
```diff
@@ -401,3 +401,58 @@ type SearchResults struct {
 	Shards json.RawMessage            // model when needed
 	Aggs   map[string]json.RawMessage // model when needed
 }
+
+type ShutdownType string
```
naemono, line 405: Are we intentionally disabling these types of checks in `golangci-lint`?

```
❯ golint pkg/controller/elasticsearch/client/model.go
---cut---
pkg/controller/elasticsearch/client/model.go:405:6: exported type ShutdownType should have comment or be unexported
```

pkg/controller/elasticsearch/driver/upgrade.go
```diff
@@ -296,24 +329,56 @@ func (d *defaultDriver) MaybeEnableShardsAllocation(
...
-func (ctx *rollingUpgradeCtx) prepareClusterForNodeRestart(esClient esclient.Client, esState ESState) error {
+func (ctx *rollingUpgradeCtx) readyToDelete(pod corev1.Pod) (bool, error) {
+	if !supportsNodeShutdown(ctx.esClient.Version()) {
+		return true, nil // always OK to restart pre node shutdown support
+	}
+	if pod.Status.Phase != corev1.PodRunning {
+		// there is no point in trying to query the shutdown status of a Pod that is not running
+		return true, nil
+	}
+	response, err := ctx.nodeShutdown.ShutdownStatus(ctx.parentCtx, pod.Name)
```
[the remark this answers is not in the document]
pebrc: I can do that I think that makes sense.

pebrc:
> My only concern is that it is hard to understand the flow of the events from the logs because we only have the node id in the message some messages, not the node name.

We could keep a reverse lookup from ES node ID to `podName` in that shutdown struct and enrich any log messages that currently don't have pod name information.

pkg/controller/elasticsearch/client/model.go
pebrc: No I don't think this is on purpose.

naemono: I'll open a separate issue on this, as it's outside of the context of this PR, and this level of changes would pollute this PR

```
❯ golint ./... | wc -l
    1252
```

pkg/controller/elasticsearch/driver/esstate.go
```diff
@@ -66,7 +67,11 @@ func (n *memoizingNodes) initialize() error {
 	if err != nil {
 		return err
 	}
-	n.nodes = nodes.Names()
+	m := map[string]string{}
+	for id, node := range nodes.Nodes {
+		m[node.Name] = id
+	}
+	n.nodeNameToID = m
```
naemono, line 74:
```suggestion
	n.nodeNameToID := map[string]string{}
	for id, node := range nodes.Nodes {
		n.nodeNameToID[node.Name] = id
	}
```

pkg/controller/elasticsearch/client/model.go
pebrc: That's a good idea 👍

pkg/controller/elasticsearch/driver/nodes.go
```diff
@@ -165,11 +165,16 @@ func (d *defaultDriver) reconcileNodeSpecs(
 	// Maybe clear zen2 voting config exclusions.
 	requeue, err = zen2.ClearVotingConfigExclusions(ctx, d.ES, d.Client, esClient, actualStatefulSets)
 	if err != nil {
-		return results.WithError(err)
+		return results.WithError(fmt.Errorf("when clearing voting exclustions: %w", err))
```
naemono, line 168:
```suggestion
		return results.WithError(fmt.Errorf("when clearing voting exclusions: %w", err))
```

pkg/controller/elasticsearch/driver/upgrade.go
```diff
@@ -296,24 +329,56 @@ func (d *defaultDriver) MaybeEnableShardsAllocation(
...
 			return err
 		}
 	}
 
 	// Request a flush to optimize indices recovery when the node restarts.
-	if err := doFlush(ctx.parentCtx, ctx.ES, esClient); err != nil {
-		return err
-	}
-
-	// TODO: halt ML jobs on that node
```
naemono, line 317: was this something we still needed to do; is there a backlog story for implementation?

pkg/controller/elasticsearch/migration/migrate_data.go
```diff
@@ -10,16 +10,54 @@ import (
...
+// ShardMigration implements the shutddown.Interface based on externally controlled shard allocation filtering.
+type ShardMigration struct {
+	es esv1.Elasticsearch
+	c  esclient.Client
+	s  esclient.ShardLister
+}
+
+var _ shutdown.Interface = &ShardMigration{}
+
+// NewShardMigration creates a new ShardMigration struct that holds no other state then the arguments to this
```
naemono, line 28:
```suggestion
// NewShardMigration creates a new ShardMigration struct that holds no other state than the arguments to this
```

pkg/controller/elasticsearch/driver/upgrade.go
pebrc: This was an unaddressed issue that the node shutdown API takes care of for all versions of Elasticsearch that support the API. We have no plans of managing ML jobs (or ILM policies which is another unsolved thing) for older versions of Elasticsearch.

pkg/controller/elasticsearch/migration/migrate_data.go
```diff
@@ -10,16 +10,54 @@ import (
...
 	esv1 "github.com/elastic/cloud-on-k8s/pkg/apis/elasticsearch/v1"
 	esclient "github.com/elastic/cloud-on-k8s/pkg/controller/elasticsearch/client"
+	"github.com/elastic/cloud-on-k8s/pkg/controller/elasticsearch/shutdown"
 	ulog "github.com/elastic/cloud-on-k8s/pkg/utils/log"
 )
 
 var log = ulog.Log.WithName("migrate-data")
 
-// NodeMayHaveShard returns true if one of those condition is met:
+// ShardMigration implements the shutddown.Interface based on externally controlled shard allocation filtering.
```
barkbay, line 19:
```suggestion
// ShardMigration implements the shutdown.Interface based on externally controlled shard allocation filtering.
```

pkg/controller/elasticsearch/migration/migrate_data.go
```diff
@@ -10,16 +10,54 @@ import (
...
+	if err != nil {
+		return shutdown.NodeShutdownStatus{}, err
+	}
+	if migrating {
+		return shutdown.NodeShutdownStatus{Status: esclient.ShutdownStarted}, nil
+	}
+	return shutdown.NodeShutdownStatus{Status: esclient.ShutdownComplete}, nil
+}
+
+// nodeMayHaveShard returns true if one of those condition is met:
```
barkbay, line 56:
```suggestion
// nodeMayHaveShard returns true if one of those conditions is met:
```

pkg/controller/elasticsearch/driver/upgrade.go
barkbay: Let's keep it simple for now, I'm not convinced that it really deserves the complexity of handling an additional annotation.

barkbay (approved):
LGTM 🚀 
Great work 👍 !

PR #4598, opened 2021-06-29
Simplify setup of Enterprise Search access through Kibana UI

sebgl:
Starting version 7.14.0, Kibana becomes the default UI for Enterprise Search.
This PR introduces an `enterpriseSearchRef` in the Kibana CRD, to simplify the Kibana -> Enterprise Search connection configuration.

Sample manifests to try it out:

```yaml
kind: Elasticsearch
metadata:
  name: es-test
spec:
  version: 7.14.0
  image: docker.elastic.co/elasticsearch/elasticsearch:7.14.0-SNAPSHOT
  nodeSets:
    - name: default
      count: 1
      config:
        node.store.allow_mmap: false
---
apiVersion: enterprisesearch.k8s.elastic.co/v1beta1
kind: EnterpriseSearch
metadata:
  name: entsearch-test
spec:
  version: 7.14.0
  image: docker.elastic.co/enterprise-search/enterprise-search:7.14.0-SNAPSHOT
  count: 1
  elasticsearchRef:
    name: es-test
---
kind: Kibana
metadata:
  name: kibana-test
spec:
  version: 7.14.0
  image: docker.elastic.co/kibana/kibana:7.14.0-SNAPSHOT
  count: 1
  elasticsearchRef:
    name: es-test
  enterpriseSearchRef:
    name: entsearch-test
```

You can then access Enterprise Search by clicking the "Enterprise Search" button in Kibana UI.

---

Some technical details:

This is the first time we introduce an association between two resources that does not involve creating an Elasticsearch user along the way. Therefore, aside from the "simple" additional Kibana -> Enterprise Search association, this PR refactors the association controller to allow associations with no transitive Elasticsearch resource, and no additional user creation.

A few noticeable changes:
- Each association's `AssociationInfo` now contains an optional `ElasticsearchUserCreation` section to embed the Elasticsearch user creation details (`nil` in the Kibana -> Enterprise Search association).
- The existing association conf annotation remains unmodified, however a special value can be set on the auth secret name to indicate no auth configuration is required: `authSecretName: "-"` (as opposed to `""` which means the value has not been set yet). 
- We now explicitly check whether the referenced resource (not always Elasticsearch!) exists at the very beginning of the association reconciliation.
- I refactored the dynamic watches to ensure we always watch the referenced resource (can be done in a generic way), and conditionally watch the es user secret. No need for `SetDynamicWatches` in the `AssociationInfo` anymore.

The PR also adds an integration test to cover the Kibana -> Enterprise Search association, and checks whether the Enterprise Search API is correctly available through Kibana.

Remaining to be done:
- Documentation (through another PR)


pkg/apis/common/v1/association.go
```diff
@@ -106,6 +109,10 @@ const (
 	// should use `SingletonAssociationID` as their `AssociationID`. On the contrary, Agent can have unbounded number
 	// of Associations so Agent-ES Associations should _not_ use `SingletonAssociationID`.
 	SingletonAssociationID = ""
+
+	// NoAuthRequiredValue is the value set for AuthSecretName if no authentication
+	// credentials are necessary for that association.
+	NoAuthRequiredValue = "-"
```
sebgl, line 115: I'm assuming here that no secret will ever be named "-". Since this is a value we set with our own secret naming scheme it should be OK.
It's somewhat nice because it doesn't break the existing schema and is quite easy to manipulate in the code.
However it also feels a bit "implicit", we could also modify the assoc conf schema to make things more explicit. WDYT?

pkg/controller/association/dynamic_watches.go
```diff
@@ -43,17 +43,39 @@ func serviceWatchName(associated types.NamespacedName) string {
...
+
+		// TODO: this is not great. This reconcileWatches function is called by each association controller.
+		// For example for Kibana, called both by the kb-es and kb-ent controllers.
+		// Both controllers override each other's watches. It's OK when they write the same thing, but in this case
+		// the kb-ent controller does not setup any es user while the kb-es controller does.
+		// Because the context is different in both controller we end up with those weird if conditions.
+		// Things would be more consistent if each association controller sets up its own watches rather than
+		// having multiple association controllers setting up all associations watches for a given resource.
+		// This way we'd just check r.ElasticsearchUserCreation != nil, rather than checking again later
+		// for each association whether it requires auth.
```
sebgl, line 64: This is the part I don't like much.
@david-kow I remember we discussed doing things that way when we introduced the agent multi-association support, for simplicity. Now it's causing some troubles since multiple association controllers (kb-es, kb-ent) reconcile the same watches with a different reconciler parameters.


I'm trying to think about a better solution.

pkg/controller/kibana/config_settings.go
```diff
@@ -263,8 +274,33 @@ func elasticsearchTLSSettings(kb kbv1.Kibana) map[string]interface{} {
...
+	assocConf := kb.EntAssociation().AssociationConf()
+	if assocConf.URLIsConfigured() {
+		cfg[EnterpriseSearchHost] = assocConf.GetURL()
+	}
+	if assocConf.CAIsConfigured() {
+		cfg[EnterpriseSearchSslCertificateAuthorities] = filepath.Join(entCertsVolumeMountPath, certificates.CAFileName)
+		// Rely on "certificate" verification mode rather than "full" to allow Kibana
+		// to connect to Enterprise Search through the k8s-internal service DNS name
+		// even though the user-provided certificate may only specify a public-facing DNS.
+		cfg[EnterpriseSearchSslVerificationMode] = "certificate"
```
sebgl, line 303: @kovyrin fyi this is what we also do for the Kibana -> Elasticsearch connection.
Less secure but more convenient when using internal k8s dns services.
Users are free to override that configuration option through the Kibana CRD if they want to. Users setting will take precedence.
Let me know if you think that's a problem!

test/e2e/test/kibana/checks_kb.go
```diff
@@ -62,3 +67,20 @@ func (check *kbChecks) CheckKbStatusHealthy(b Builder) test.Step {
...
+	return test.Step{
+		Name: "The Enterprise Search UI should be available in Kibana",
+		Test: test.Eventually(func() error {
+			password, err := check.client.GetElasticPassword(b.ElasticsearchRef().NamespacedName())
+			if err != nil {
+				return errors.Wrap(err, "while getting elastic password")
+			}
+			// returns 200 OK if accessible
+			path := "/api/enterprise_search/config_data"
+			_, err = DoRequest(check.client, b.Kibana, password, "GET", path, nil)
```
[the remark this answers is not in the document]
kovyrin: This URL looks good!

pkg/controller/kibana/config_settings.go
kovyrin: I think it makes sense to keep this in line with what we do for other products. It is definitely a lot more convenient and makes sense as a default in a world where TLS certificate names are pretty arbitrary (depend on resource names).

pkg/controller/association/dynamic_watches.go
sebgl: I ended up refactoring the dynamic watches to make things simpler and more consistent.

sebgl:
Jenkins test this please

pkg/controller/association/reconciler_test.go
```diff
@@ -456,46 +455,197 @@ func TestReconciler_Reconcile_NewAssociation(t *testing.T) {
...
+			ent := entv1.EnterpriseSearch{}
+			if err := c.Get(context.Background(), entRef.NamespacedName(), &ent); err != nil {
+				return "", err
+			}
+			serviceName := entRef.ServiceName
+			if serviceName == "" {
+				serviceName = "entname-ent-http"
+			}
+			nsn := types.NamespacedName{Namespace: ent.Namespace, Name: serviceName}
+			return ServiceURL(c, nsn, ent.Spec.HTTP.Protocol())
```
pebrc, line 545: What's the reason for not just using `getEntExternalServiceURL` here?

pkg/controller/association/reconciler_test.go
```diff
@@ -456,46 +455,197 @@ func TestReconciler_Reconcile_NewAssociation(t *testing.T) {
...
+		},
+		ReferencedResourceVersion: func(c k8s.Client, entRef types.NamespacedName) (string, error) {
+			var ent entv1.EnterpriseSearch
+			err := c.Get(context.Background(), entRef, &ent)
+			if err != nil {
+				return "", err
+			}
+			return ent.Status.Version, nil
+		},
```
pebrc, line 554: Same here, why not use the function from the production code?

pebrc (commented):
Looks good! I like the association refactoring beyond adding the required capabilities I think it is an improvement in general. 

> We now explicitly check whether the referenced resource (not always Elasticsearch!) exists at the very beginning of the association reconciliation. An additional ReferencedResourceExists() function needs to be implemented by each association controller that knows about the referenced object type.

I think you gave up on that idea and found a generic way with `k8s.ObjectExists`

pkg/controller/association/reconciler_test.go
sebgl: Circular dependencies :(
the association package is supposed to be generic to any association controller defined in `association/controllers` (which itself imports `association/`).
Here we test the "generic" reconciler logic, not the kibana<->ent association controller specifically.
Since it's for unit tests only I was thinking a bit of code duplication is not _too_ bad (we could also make this function much simpler in the unit test).

pebrc: Makes sense!

pkg/controller/enterprisesearch/labels.go
```diff
@@ -14,6 +14,8 @@ const (
 	Type = "enterprise-search"
 	// EnterpriseSearchNameLabelName used to represent an EnterpriseSearch in k8s resources.
 	EnterpriseSearchNameLabelName = "enterprisesearch.k8s.elastic.co/name"
+	// EnterpriseSearchNameLabelName used to represent an EnterpriseSearch in k8s resources.
```
thbkrkr, line 17:
```suggestion
	// EnterpriseSearchNamespaceLabelName used to represent an EnterpriseSearch in k8s resources.
```

pkg/apis/kibana/v1/kibana_types.go
```diff
@@ -57,30 +96,19 @@ type KibanaSpec struct {
 // KibanaStatus defines the observed state of Kibana
 type KibanaStatus struct {
 	commonv1.DeploymentStatus `json:",inline"`
-	AssociationStatus         commonv1.AssociationStatus `json:"associationStatus,omitempty"`
+	// AssociationStatus is the status of any auto-linking to Elasticsearch clusters.
+	// This was kept as "associationStatus" rather than a more consistent "elasticsearchAssociationStatus"
```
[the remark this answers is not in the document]
sebgl: 👍 we discussed async that we can do this but and keep the existing field marked as deprecated.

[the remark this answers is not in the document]
sebgl: hm actually not that simple since the existing field is already called `associationStatus` :(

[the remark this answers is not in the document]
sebgl: we could keep everything flat though?

```
status:
  availableNodes: 3
  associationStatus: green # deprecated, will be removed in a future release
  elasticsearchAssociationStatus: green
  enterpriseSearchAssociationStatus: green
```

Our APM CRD has this flat structure ealready:

```
	// ElasticsearchAssociationStatus is the status of any auto-linking to Elasticsearch clusters.
	ElasticsearchAssociationStatus commonv1.AssociationStatus `json:"elasticsearchAssociationStatus,omitempty"`
	// KibanaAssociationStatus is the status of any auto-linking to Kibana.
	KibanaAssociationStatus commonv1.AssociationStatus `json:"kibanaAssociationStatus,omitempty"`
```

pebrc (approved):
LGTM but I think we need to regenerate the API docs.

sebgl:
Jenkins test this please

PR #4600, opened 2021-06-29
Allow users to configure SANs on transport certificates

### LINKED ISSUE #4025: Customizable transport layer certificates
I have a request regarding the certificate management for the transport layer certificates.

The basic construct is as follows.
- ECK version: 1.3.0
-  A legacy app needs to communicate with the elasticsearch cluster on port 9300
- In order to validate the node certificates, I need to configure the app with the ca (quickstart-es-transport-ca-internal)

As long as I communicate directly with the Pods through the Pod-IPs everything works fine.
However, since these IPs are dynamic by nature I would like to implement the communication based on a LoadBalancer.

As described in https://www.elastic.co/guide/en/cloud-on-k8s/master/k8s-transport-settings.html, I deployed the LoadBalancer for the transport interface.

	spec:
	  transport:
	    service:
	      spec:
	        type: LoadBalancer

Which works...

	kubectl get svc
	NAME                             TYPE           CLUSTER-IP        EXTERNAL-IP      PORT(S)          AGE
	quickstart-es-transport          LoadBalancer   192.168.200.212   172.16.189.102   9300:30343/TCP   43h

...and registers the endpoints as expected.

	kubectl describe svc quickstart-es-transport
	Endpoints:                10.233.35.66:9300,10.233.35.68:9300,10.233.35.71:9300

However, I am not able to communicate with the cluster through the application since the communication is blocked with the following message.	

	[quickstart-es-coordinating-plane-1] client did not trust this server's certificate, closing connection Netty4TcpChannel{localAddress=0.0.0.0/0.0.0.0:9300, remoteAddress=/172.16.181.159:47597}

As expected, the certificate returned by the LoadBalancer is one of the coordinating nodes.

	openssl s_client -host 172.16.189.102 -port 9300

	Subject Alternative Names: 
		othername:<unsupported>, 
		quickstart-es-coordinating-plane-2.node.quickstart.poc.es.local, 
		quickstart-es-transport.poc.svc, 
		quickstart-es-coordinating-plane-2.quickstart-es-coordinating-plane, 
		IP Address:10.233.35.71, 
		IP Address:127.0.0.1

It seems the error arrises, because the certificates are validated in full.
Therefore, since the application connects to the LoadBalancer on 172.16.189.102 and the certificate returns only the node IP 10.233.35.71, the communication is not trusted.

For the http interface there is the option to provide custom SANs for the certificates
	https://www.elastic.co/guide/en/cloud-on-k8s/master/k8s-http-settings-tls-sans.html#k8s-elasticsearch-http-service-san

However, this does not seem to be possible for the transport interface.

Would it be possible to implement a the construct of customizing cerfiticates on the transport layer like this, to incorporate the LoadBalancer IP?

	spec:
	  transport:
	    tls:
	      selfSignedCertificate:
	        subjectAltNames:
	          - ip: 172.16.189.102
	          - dns: custom.example.com

pebrc:
Fixes #4025 

This adds the following extension to the Elasticsearch API and configures the user-specified DNS names and IPs into the certificates generated by the operator. 

```yaml
apiVersion: elasticsearch.k8s.elastic.co/{eck_crd_version}
kind: Elasticsearch
metadata:
  name: quickstart
spec:
  transport:
    tls:
      subjectAltNames:
      - ip: 1.2.3.4
      - dns: hulk.example.com
```

I tested manually with a transport client through a load balancer which worked fine. I don't see a good way of automating this though in an e2e test without introducing the deprecated transport client. Maybe the unit test I added is sufficient?


pebrc:
Jenkins test this please

pebrc:
Jenkins test this please

pkg/controller/elasticsearch/certificates/transport/csr_test.go
```diff
@@ -97,6 +98,32 @@ func Test_buildGeneralNames(t *testing.T) {
...
 		},
+		{
+			name: "optional user provided SANs",
+			args: args{
+				cluster: func() esv1.Elasticsearch {
+					es := testES
+					es.Spec.Transport.TLS.SubjectAlternativeNames = []commonv1.SubjectAlternativeName{
+						{
+							DNS: "my-custom-domain",
+							IP:  "111.222.333.444",
```
sebgl, line 109: I would argue this is an unexpected way to use the API and not the one we document.

What we test here:

```
- ip: 1.2.3.4
  dns: abcd.com
```

vs what we document:

```
- ip: 1.2.3.4
- dns: abcd.com
```

Both are valid though, let's make sure our unit test encapsulate them (i.e. add more entries to that test array 😄)?

sebgl (approved):
LGTM
I'm fine with the unit test only.

pkg/controller/elasticsearch/certificates/transport/csr_test.go
pebrc: Good point.

PR #4611, opened 2021-07-05
Add doc to configure Kibana to access Enterprise Search

sebgl:
This adds a section to the Enterprise Search quickstart to highlight how
Kibana can be configured to serve the Enterprise Search UI, following https://github.com/elastic/cloud-on-k8s/pull/4598.

Even though the impacted resource is Kibana, I think it makes more sense
to document this feature in the Enterprise Search section.

PR #4643, opened 2021-07-15
[1.3] Add doc to configure Kibana to access Enterprise Search (#4611)

sebgl:
 - Add doc to configure Kibana to access Enterprise Search (#4611)


sebgl:
Meh. Mis-backported.

PR #4664, opened 2021-07-21
Update Secrets and OwnerReference in Common problems

### LINKED ISSUE #4659: Clarify "common problems" re. Secrets and OwnerReference
We should clarify our [documentation](https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-common-problems.html#k8s-common-problems-owner-refs) to let the users know that since `1.3.1` the `OwnerReference` has been removed from `Secrets` likely to be copied across namespaces. We should however also point out that any previously copied `Secrets` may trigger this bug and it is up to to the user to ensure that all the `OwnerReference` have been removed.

alaudazzi:
This PR reworks the first paragraph of the [Copying secrets with Owner References](https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-common-problems.html#k8s-common-problems-owner-refs) section.

Fixes #4659


docs/operating-eck/troubleshooting/common-problems.asciidoc
```diff
@@ -35,7 +35,9 @@ This error is usually an indication of a problem communicating with the validati
 [id="{p}-{page_id}-owner-refs"]
 == Copying secrets with Owner References
 
-Copying the Elasticsearch Secrets generated by ECK (for instance, the certificate authority or the elastic user) into another namespace wholesale can trigger a link:https://github.com/kubernetes/kubernetes/issues/65200[Kubernetes bug] which can delete all of the Elasticsearch-related resources (for instance, the data volumes). To avoid this, when copying Secrets between namespaces, remove the `metadata.ownerReferences` section. For example, a source secret might be:
+Since ECK 1.3.1, `OwnerReference` was removed from Elasticsearch Secrets. Copying the Elasticsearch Secrets generated by ECK (for instance, the certificate authority or the elastic user) into another namespace wholesale can trigger a link:https://github.com/kubernetes/kubernetes/issues/65200[Kubernetes bug] which can delete all of the Elasticsearch-related resources (for instance, the data volumes). Make sure you remove the `metadata.ownerReferences` section, even from Elasticsearch Secrets that were previously copied.
```
barkbay, line 38: Maybe we should clarify that the `OwnerReference` is automatically removed from a subset of those Secrets, for example:

> Since ECK 1.3.1, `OwnerReference` was removed from Secrets which are likely to be copied (the Secrets which contain the public certificates and the Secret which holds the `elastic` user credentials)

I think I would also move that part after `Copying the Elasticsearch Secrets...` and mention that if some Secrets have been copied in other namespaces before `1.3.1`, then users should remove the `OwnerReference` manually because they might still be affected, even if ECK has been upgraded.

barkbay (approved):
LGTM, thanks ! 🙇

PR #4679, opened 2021-07-26
Add v1alpha1/[es|kb|apm] versions in v1 crds

barkbay:
This PR adds an empty `v1alpha1` version for Elasticsearch, Kibana and AMP Server. Without it, users might not be able to apply or replace the new CRDs if the `v1alpha1` versions are still referenced in the CRD's status.


pebrc:
I believe I removed the empty v1alpha1 versions on purpose because they failed structural schema validation but my memory is a bit hazy. I wanted to add a note about this in the upgrade instructions but maybe I forgot? However I think our current guidance for users coming from v1alpha1 versions is a clean install anyway, so maybe that is acceptable?

barkbay:
> I believe I removed the empty v1alpha1 versions on purpose because they failed structural schema validation but my memory is a bit hazy.

`openAPIV3Schema` is required in `v1`, which was indeed not the case in `v1beta1`, this is why there is an empty `properties` field:

```json
schema:
  openAPIV3Schema:
    description: to not break compatibility when upgrading from previous versions
      of the CRD
    properties: {}
    type: object
```

> However I think our current guidance for users coming from v1alpha1 versions is a clean install anyway, so maybe that is an acceptable upgrade approach here as well?

The documentation invites the user to install "ECK 1.0.0-beta1 in a fresh Kubernetes cluster". My thinking was that we may want to be conservative because we are already providing a workaround, and our test procedure includes a full upgrade from a cluster where `v1alpha1` resources have been installed.

That being said I'm fine with this approach and happy to remove `v1alpha1` from our upgrade test harness 😄

barkbay:
On a second thought my concern is that users who have upgraded from `v1alpha1` at some point, even if they are using a stable release today, may not be able to upgrade to `v1` because `v1alpha1` is referenced in the CRDs status.

pebrc:
> On a second thought my concern is that users who have upgraded from v1alpha1 at some point, even if they are using a stable release today, may not be able to upgrade to v1 because v1alpha1 is referenced in the CRDs status.

Yes unfortunately we outlined an upgrade approach back then as an alternative to a fresh install for advanced users. So maybe you are right to include the v1alpha1 reference if it does not cause other problems.  👍

pebrc (approved):

barkbay:
> openAPIV3Schema is required in v1, which was indeed not the case in v1beta1, this is why there is an empty properties field:

In the context of https://github.com/elastic/cloud-on-k8s/pull/4813 I was trying to figure out which fields are mandatory, it seems that it's only the case for `type: object`:

```yaml
# Add v1alpha1 to the list of versions, to not break compatibility when upgrading from previous versions of the CRD
# that list the v1alpha1 version in their status.storedVersions. We mark it as `served: false` so it cannot be used.
- op: add
  path: /spec/versions/-
  value:
    name: v1alpha1
    served: false
    storage: false
    schema:
      openAPIV3Schema:
        type: object
```

⬆️  `properties` and `description` seem to be optional

PR #4682, opened 2021-07-26
ECK dump: avoid sensitive fields when dumping secret metadata

### LINKED ISSUE #4435: eckdump.sh also dumps secret content through metadata
We inadvertently dump the contents of (all) secrets in the the `eckdump.sh` script. While the intent behind the `get_metadata` function was probably to avoid exactly that it is not effective. This is because the `last-applied-configuration` metadata also contains the secret data. 

We should probably be a bit smarter about that by either:
1. not dump secrets at all
2. be more selective with secrets: e.g. what we really want is the keys but not the values as this gives us clues whether the expected data is present, also avoid dumping the `last-applied-configuration` annotation.

pebrc:
Fixes #4435

This replaces the JSON dump of the K8s Secret metadata with a Go-template producing a text file that leaves out sensitive information specifically the `last-applied-configuration` annotation `kubectl` adds. 

This PR does not have a version label as we don't release the script explicitly but users just pull it from the repo.

### DIFF
--- a/support/diagnostics/eck-dump.sh
+++ b/support/diagnostics/eck-dump.sh
@@ -34,6 +34,28 @@ RESOURCES_NS="default"
 OUTPUT_DIR=""
 ZIP=""
 VERBOSE=""
+# go-template to dump metadata in textual form excluding sensitive fields like "last-applied-configuration"
+METADATA_TPL=$(cat <<'EOF'
+{{range .items}}
+{{"---" -}}
+{{range $k,$v := .metadata }}
+{{ printf `"%v": ` $k -}}
+  {{- if eq $k "annotations" "labels" -}}
+    {{- $nth := false -}}
+    {{- range $k,$v := . -}}
+      {{- if ne $k "kubectl.kubernetes.io/last-applied-configuration" -}}
+        {{- if $nth -}}{{- "," -}}{{- end -}}
+        {{- printf `"%v":"%v"` $k $v -}}
+        {{$nth = true}}
+      {{- end -}}
+    {{- end -}}
+  {{else}}
+    {{- printf `"%v"` $v -}}
+  {{- end -}}
+{{- end -}}
+{{end}}
+EOF
+)
 
 parse_args() {
   while :; do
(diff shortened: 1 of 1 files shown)


pebrc:
Jenkins test this please

support/diagnostics/eck-dump.sh
```diff
@@ -34,6 +34,28 @@ RESOURCES_NS="default"
...
+    {{- $nth := false -}}
+    {{- range $k,$v := . -}}
+      {{- if ne $k "kubectl.kubernetes.io/last-applied-configuration" -}}
+        {{- if $nth -}}{{- "," -}}{{- end -}}
+        {{- printf `"%v":"%v"` $k $v -}}
+        {{$nth = true}}
+      {{- end -}}
+    {{- end -}}
+  {{else}}
+    {{- printf `"%v"` $v -}}
```
thbkrkr, line 53: Nit: `%v` vs `$v` doesn't help to decypher the template:
```suggestion
{{range $key,$value := .metadata }}
{{ printf `"%v": ` $key -}}
  {{- if eq $key "annotations" "labels" -}}
    {{- $nth := false -}}
    {{- range $key,$value := . -}}
      {{- if ne $key "kubectl.kubernetes.io/last-applied-configuration" -}}
        {{- if $nth -}}{{- "," -}}{{- end -}}
        {{- printf `"%v":"%v"` $key $value -}}
        {{$nth = true}}
      {{- end -}}
    {{- end -}}
  {{else}}
    {{- printf `"%v"` $value -}}
```

thbkrkr (approved):
LGTM, left a nitpick about readability.

Indirectly related to this PR: I didn't remember why we don't fetch the secrets metadata from the namespaces in which operators are running in but we could now do it safely with this change.

```diff
diff --git a/support/diagnostics/eck-dump.sh b/support/diagnostics/eck-dump.sh
index 9477f3786..bcad68cc7 100755
--- a/support/diagnostics/eck-dump.sh
+++ b/support/diagnostics/eck-dump.sh
@@ -120,6 +120,7 @@ main() {
     get_resources "$ns" events
     get_resources "$ns" networkpolicies
     get_resources "$ns" controllerrevisions
+    get_metadata "$ns" secrets
     get_logs "$ns"
   done
```

thbkrkr:
Retested and still LGTM

PR #4702, opened 2021-07-28
Bump Stack version to 7.13.4

thbkrkr:
* Bump min `6.8.x` version to `6.8.17`
* Bump max `7.x` version to `7.13.4`
* Update `hack/update-stack-version.sh` to update the stack version in `test/e2e/stack_test.go` and `hack/operatorhub/config.yaml`

I think we should backport this into `1.7`.


barkbay (approved):
LGTM !

> I think we should backport this into 1.7.

👍

Issue #4710, opened 2021-07-29
ent.TestEnterpriseSearchConfigUpdate fails in 8.0.0-SNAPSHOT


thbkrkr:
`ent.TestEnterpriseSearchConfigUpdate/Enterprise_Search_Pods_should_eventually_be_ready` failed in
https://devops-ci.elastic.co/job/cloud-on-k8s-e2e-tests-snapshot-versions-8x/212/.

EnterpriseSearch pod log from the eck-dump:
```
Invalid config file (/usr/share/enterprise-search/config/enterprise-search.yml):
The setting '#/kibana/external_url' is required. The standalone Enterprise Search management UI is deprecated.
Please follow the setup instructions at https://www.elastic.co/guide/en/enterprise-search/master/user-interfaces.html
to configure and run Kibana.
```

thbkrkr:
Something is odd. This should be covered by #4598 and especially when I run the test locally I got:
```
Unsupported Elasticsearch version: 8.0.0-SNAPSHOT.
Enterprise Search requires Elasticsearch >= 8.0 and < 8.
```
which should be fixed by #4604.

I spoted the bug in #4604. We add the special env var to all containers in the pod template but there is no containers in the pod template. We need to add at least the default main container if there is none.

thbkrkr:
I went through the last 15 builds of the `cloud-on-k8s-e2e-tests-snapshot-versions-8x` job:
```
201 Unsupported Elasticsearch version: 8.0.0-SNAPSHOT. Enterprise Search requires Elasticsearch >= 8.0 and < 8.
202 github.com/elastic/cloud-on-k8s/test/e2e.TestNameValidation/longestPossibleName/All_expected_Pods_should_eventually_be_ready
203 github.com/elastic/cloud-on-k8s/test/e2e.TestNameValidation/longestPossibleName/All_expected_Pods_should_eventually_be_ready
204 Unsupported Elasticsearch version: 8.0.0-SNAPSHOT. Enterprise Search requires Elasticsearch >= 8.0 and < 8.
205 Unsupported Elasticsearch version: 8.0.0-SNAPSHOT. Enterprise Search requires Elasticsearch >= 8.0 and < 8.
206 github.com/elastic/cloud-on-k8s/test/e2e.TestNameValidation/longestPossibleName/All_expected_Pods_should_eventually_be_ready
[...]

Issue #4455, opened 2021-04-29
ent.TestEnterpriseSearchConfigUpdate fails in 8.0.0-SNAPSHOT


thbkrkr:
`ent.TestEnterpriseSearchConfigUpdate/Enterprise_Search_Pods_should_eventually_be_ready` failed in https://devops-ci.elastic.co/view/cloud-on-k8s/job/cloud-on-k8s-e2e-tests-snapshot-versions-8x/149/.

Pod crashes with `CrashLoopBackOff` due to:

```
Invalid config file (/usr/share/enterprise-search/config/enterprise-search.yml):
The setting '#/ent_search/auth/source' was completely deprecated in version '8.0.0'. Please use the new auth config format ent_search.auth.<auth_name>.source.
``` 

This was solved in #3725 where we made a change to use the new auth config format starting EnterpriseSearch 8.x.

But since we use the `semver` package for version parsing (#4174), we broke the version test:

https://github.com/elastic/cloud-on-k8s/blob/a4b7e530d9bb73a349cc11c82d73cc05778c71e4/pkg/controller/enterprisesearch/config.go#L279-L285

According to [semver](https://semver.org/#spec-item-11), `8.0.0-SNAPSHOT < 8.0.0`:

> When major, minor, and patch are equal, a pre-release version has lower precedence than a normal version:
> Example: 1.0.0-alpha < 1.0.0.

thbkrkr:
Tested: `7.999.0` < `8.0.0-0` < `8.0.0-A` < `8.0.0-SNAPSHOT` < `8.0.0-alpha1` < `8.0.0-bc1` < `8.0.0`

My thoughts on this: 
- compare strictly to 7.999.999 instead of 8.0.0, (1)
- add a prerelease (`0`, `A`, `SNAPSHOT`) to the reference version 8.0.0, (2)
- clean the source version we test to remove the prerelease, (3) (4)

```go
func TestGTWithSnapshot(t *testing.`T) {
	var v1,v2 Version

	v1, _ = Parse("8.0.0-SNAPSHOT")
	v2 = From(8, 0, 0)
	require.False(t, v1.GTE(v2))  // :-( This is what we want to fix.

	// (1) Compare with 7.999.999
	v1, _ = Parse("8.0.0-SNAPSHOT")
	v2 = From(7, 999, 0)
	require.True(t, v1.GT(v2))    // GT instead of GTE

[...]

PR #4713, opened 2021-07-29
Fix to allow to test EnterpriseSearch with Elasticsearch 8.0.0-SNAPSHOT

thbkrkr:
In #4604, I added `WithEnvVar` to the EnterpriseSearch builder to add the environment variable `ALLOW_PREVIEW_ELASTICSEARCH_8X=true`. 
This adds the environment variable passed as a parameter in all containers defined in the pod template of the EnterpriseSearch resource. But if there is no containers in the pod template, we don't add the env var.

This commit fixes that by adding the main `enterprise-search` container if there is no container defined in the pod template.

Relates to #4710 and #4455.

### DIFF
--- a/test/e2e/test/enterprisesearch/builder.go
+++ b/test/e2e/test/enterprisesearch/builder.go
@@ -177,6 +177,9 @@ func (b Builder) WithPodLabel(key, value string) Builder {
 }
 
 func (b Builder) WithEnvVar(name, value string) Builder {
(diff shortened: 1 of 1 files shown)


thbkrkr:
```
--- FAIL: TestDynamicEnqueueRequest (30.93s)
```

jenkins test this please

pebrc (approved):

PR #4721, opened 2021-08-03
Bump versions in preparation for 1.7.0 release for OLM

pebrc:
The OLM release needs to wait until 7.14.0 becomes available of course, while the ECK 1.7.0 artefacts should be available already.


thbkrkr (approved):
:rocket:

PR #4759, opened 2021-08-13
Add validation of distribution channel

david-kow:
This allows to filter out invalid values of distribution channel.


thbkrkr (approved):
LGTM

PR #4761, opened 2021-08-13
Fix documentation rendering

thbkrkr:
Remove spaces at the start of the line which disturb the rendering.

Fixes a small regression introduced in #4664.


thbkrkr:
Before:

2021-08-13-100610_851x176_scrot

After:

2021-08-13-100631_852x247_scrot

alaudazzi (approved):
LGTM

PR #4767, opened 2021-08-17
Docs: make volume naming convention more prominent

pebrc:
Users have missed this naming convention and run into trouble with Elasticsearch nodes writing data into the container filesystem. This PR makes it more prominent with an admonition paragraph.

### DIFF
--- a/config/samples/elasticsearch/elasticsearch.yaml
+++ b/config/samples/elasticsearch/elasticsearch.yaml
@@ -43,7 +43,7 @@ spec:
   #   # request 2Gi of persistent data storage for pods in this topology element
   #   volumeClaimTemplates:
   #   - metadata:
-  #       name: elasticsearch-data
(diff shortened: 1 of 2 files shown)


barkbay (approved):
I was wondering if we should also add a warning in the snippet and in the [sample](https://github.com/elastic/cloud-on-k8s/blob/master/config/samples/elasticsearch/elasticsearch.yaml#L46).
Something like:

```yaml
    volumeClaimTemplates:
    - metadata:
        name: elasticsearch-data # Do not change the name unless you setup a volume mount for the data path.
```

pebrc:
@elasticmachine, run elasticsearch-ci/docs

cawoodm:
According to the [helm docs](https://github.com/elastic/helm-charts/blob/7.17/elasticsearch/values.yaml) it should be called `volumeClaimTemplate:` and not `volumeClaimTemplates:`. Which is correct?

PR #4770, opened 2021-08-18
Add more detail and workaround to 1.7.0 known issue

pebrc:
This updates the known issue description for https://github.com/elastic/support-known-issues/issues/794 to include a workaround and more detail about which versions of Kubernetes are affected.

### DIFF
--- a/docs/release-notes/highlights-1.7.0.asciidoc
+++ b/docs/release-notes/highlights-1.7.0.asciidoc
@@ -42,5 +42,11 @@ After installing ECK 1.7.0 on Kubernetes versions 1.16/1.17 using Helm or the YA
 error: SchemaError(co.elastic.k8s.elasticsearch.v1.Elasticsearch.spec.nodeSets.volumeClaimTemplates): array should have exactly one sub-item
 ----
 
+This is due to an validation issue in `kubectl` that has been addressed in Kubernetes as of versions 1.18.12, 1.19.5 and 1.20. To work around this issue patch the Elasticsearch CRD as follows:
+
+[source,bash]
+----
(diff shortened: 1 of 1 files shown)


docs/release-notes/highlights-1.7.0.asciidoc
```diff
@@ -42,5 +42,11 @@ After installing ECK 1.7.0 on Kubernetes versions 1.16/1.17 using Helm or the YA
 error: SchemaError(co.elastic.k8s.elasticsearch.v1.Elasticsearch.spec.nodeSets.volumeClaimTemplates): array should have exactly one sub-item
 ----
 
+This is due to a validation issue in `kubectl` that has been addressed in Kubernetes as of versions 1.18.12, 1.19.5 and 1.20. To work around this issue patch the Elasticsearch CRD as follows:
```
barkbay, line 45: I'm wondering how we could make it cleat that the fix is not in the client, but in the Kubernetes API server:

```suggestion
This is due to a validation issue in `kubectl` that has been addressed in the Kubernetes API server as of versions 1.18.13, 1.19.5 and 1.20. To work around this issue patch the Elasticsearch CRD as follows:
```

My fear is that some users may wrongly assume that if the issue is in `kubectl` then upgrading the client can fix the problem.

barkbay (approved):
LGTM, I believe that for K8S `1.18` the fix is included in `1.18.13`.

docs/release-notes/highlights-1.7.0.asciidoc
pebrc: Good point.

PR #4780, opened 2021-08-19
Add yaml output to eck dump

kunisen:
- Have you signed the [contributor license agreement](https://www.elastic.co/contributor-agreement)? Yes
- Have you followed the [contributor guidelines](https://github.com/elastic/cloud-on-k8s/tree/master/CONTRIBUTING.md)? Yes
- If you submit code, is your pull request against master? We recommend pull requests against master. We will backport them as needed. Yes

### Comment

I feel it would be great if we could have YAML format output.
I don't want to bother the current directory structures so I simply added yaml output line.

### DIFF
--- a/docs/operating-eck/troubleshooting/take-eck-dump.asciidoc
+++ b/docs/operating-eck/troubleshooting/take-eck-dump.asciidoc
@@ -68,52 +68,88 @@ Directory structure of the output files. The output may vary depending on the co
 
 [source,bash]
 ----
- tree -A -L 10 eck-dump-log      
+tree -A -L 10 eck-dump-log
 eck-dump-log
 ├── clusterroles.txt
 ├── elastic-system
 │   ├── agent.json
+│   ├── agent.yaml
 │   ├── apmserver.json
+│   ├── apmserver.yaml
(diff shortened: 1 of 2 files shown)


pebrc:
I am slightly -1 on this change. It duplicates the output in the archive and we have good tooling for JSON data. Also `kubectl` accepts both JSON and YAML as input, if this were a concern.

kunisen:
Thanks @pebrc 

How do you think if we add another option to choose which format to output as? e.g. `-yaml` to be `yaml` and `-json` to be `json`, and let the default value to be `json`?

I understand your concern to prefer to use json, but I feel when doing support, viewing yaml is sometimes a bit easier, especially when the user gets used to yaml. 
At least from what I have heard from team, `yaml` option is (as you said not "mandatory" but)  a "slightly good to have" level thing.

barkbay:
Just want to mention this one-liner to convert json files into yaml using `yq`:

```
find . -name '*.json' -exec sh -c 'yq e -P "$1" > "$(dirname "$1")/$(basename "$1" .json).yaml"' sh {} \;
```

_Tested with YQ 4.12.0_

Issue #4625, opened 2021-07-09
One-node clusters with a yellow health cannot be upgraded


sebgl:
Deploy a one-node cluster, and ingest data into it so that the cluster becomes yellow (indices with one extra replica).
One quick way to do that is to deploy a cluster with stack monitoring enabled, while having the monitoring data sent to the cluster itself.

Once that single node cluster is yellow, there is no way to upgrade it anymore. Any change to the Elasticsearch spec requiring a rolling upgrade will wait until the cluster becomes green:

```
2021-07-09T11:37:03.750+0200	INFO	driver	Cannot restart some nodes for upgrade at this time	{"service.version": "1.7.0-SNAPSHOT+30579acb", "namespace": "default", "es_name": "quickstart", "failed_predicates": {"if_yellow_only_restart_upgrading_nodes_with_unassigned_replicas":["quickstart-es-default-0"]}}
```

I think we should make an exception for single-node clusters. There is no point in waiting for them to report a green health before rotating the single Pod, since we know we'll break availability anyway?

barkbay:
Sorry, I'm late to the party. IIUC a generic way to think about this problem is: there are `n` nodes in a tier but `number_of_replicas` is set to `m >= n`. For example having an index with 1 replica moving to the warm tier while there's only 1 node in that tier will end up in the same situation. I can also think about an other case where there is 1 data node and 1 dedicated ml node. Do we want to allow the upgrade as the ml node can't hold the shard ?

I believe we don't want to go into that level of details, and just allow a single, multirole, node to be upgraded. However it would be great to mention this exception in our documentation: https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-orchestration.html#k8s-orchestration-limitations as it might raise some questions about which cases are supported:

---

* Rolling upgrades can only make progress if the Elasticsearch cluster health is green. There are exceptions to this rule if the cluster health is yellow and if the following conditions are satisfied:

  * A cluster version upgrade is in progress and some Pods are not up to date.
  * There are no initializing or relocating shards.
  * _other conditions/exceptions here_
----

CamiloSierraH:
Hey @barkbay !!
> Do we want to allow the upgrade as the ml node can't hold the shard ?

I agree with you at 100% I think this will be great for monitoring clusters that most of then are using a single node and users do not change the default replica value, so most of them are yellow (this is what we can see 90% of the time in support).

[...]

PR #4787, opened 2021-08-23
Allow to upgrade a single node cluster with yellow status

CamiloSierraH:
Fixes: https://github.com/elastic/cloud-on-k8s/issues/4625

### DIFF
--- a/pkg/controller/elasticsearch/driver/upgrade_predicates.go
+++ b/pkg/controller/elasticsearch/driver/upgrade_predicates.go
@@ -24,6 +24,7 @@ type PredicateContext struct {
 	shardLister            client.ShardLister
 	masterUpdateInProgress bool
 	ctx                    context.Context
+	numberOfPods           int
 }
 
 // Predicate is a function that indicates if a Pod can be deleted (or not).
@@ -59,6 +60,7 @@ func NewPredicateContext(
 	podsToUpgrade []corev1.Pod,
 	masterNodesNames []string,
 	actualMasters []corev1.Pod,
+	numberOfPods int,
 ) PredicateContext {
 	return PredicateContext{
 		es:               es,
@@ -69,6 +71,7 @@ func NewPredicateContext(
 		esState:          state,
 		shardLister:      shardLister,
 		ctx:              ctx,
+		numberOfPods:     numberOfPods,
 	}
 }
 
@@ -198,7 +201,7 @@ var predicates = [...]Predicate{
 				// This predicate is only relevant on healthy node if cluster health is yellow
 				return true, nil
 			}
-			if health.NumberOfNodes == 1 {
+			if context.numberOfPods == 1 {
 				// If the cluster is a single node cluster, allow restart even if there is no version difference
 				return true, nil
 			}
@@ -227,8 +230,8 @@ var predicates = [...]Predicate{
 				return false, err
 			}
 			_, healthyNode := context.healthyPods[candidate.Name]
-			if health.NumberOfNodes == 1 && health.Status == esv1.ElasticsearchYellowHealth && healthyNode {
-				// If the cluster is a single node cluster, replicas can not be started, allow the upgrade
+			if context.numberOfPods == 1 && health.Status == esv1.ElasticsearchYellowHealth && healthyNode {
+				// If the cluster is a healthy single node cluster, replicas can not be started, allow the upgrade
 				return true, nil
 			}
 			allShards, err := context.shardLister.GetShards(context.ctx)
--- a/pkg/controller/elasticsearch/driver/upgrade_predicates.go
+++ b/pkg/controller/elasticsearch/driver/upgrade_predicates.go
@@ -198,6 +198,10 @@ var predicates = [...]Predicate{
 				// This predicate is only relevant on healthy node if cluster health is yellow
 				return true, nil
 			}
+			if health.NumberOfNodes == 1 {
+				// If the cluster is a single node cluster, allow restart even if there is no version difference
+				return true, nil
+			}
 			version := candidate.Labels[label.VersionLabelName]
 			if version == context.es.Spec.Version {
 				// Restart in yellow state is only allowed during version upgrade
@@ -222,12 +226,10 @@ var predicates = [...]Predicate{
 			if err != nil {
 				return false, err
 			}
+			health, err := context.esState.Health()
 			_, healthyNode := context.healthyPods[candidate.Name]
-			// Get the number of expected masters
-			expectedMasters := len(context.masterNodesNames)
-			healthyPods := len(context.healthyPods)
-			if healthyNode && healthyPods == 1 && expectedMasters == 1 && len(deletedPods) == 0 {
-				// If the cluster is a single node cluster, replicas can not be started, skip the predicate
+			if health.NumberOfNodes == 1 && health.Status == esv1.ElasticsearchYellowHealth && healthyNode {
+				// If the cluster is a single node cluster, replicas can not be started, allow the upgrade
 				return true, nil
 			}
 			// We maintain two data structures to record:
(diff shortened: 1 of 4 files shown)


pkg/controller/elasticsearch/driver/upgrade_predicates.go
```diff
@@ -222,6 +222,14 @@ var predicates = [...]Predicate{
 			if err != nil {
 				return false, err
 			}
+			_, healthyNode := context.healthyPods[candidate.Name]
+			// Get the number of expected masters
+			expectedMasters := len(context.masterNodesNames)
+			healthyPods := len(context.healthyPods)
+			if healthyNode && healthyPods == 1 && expectedMasters == 1 && len(deletedPods) == 0 {
```
pebrc, line 229: I am not sure that this condition is sufficient to detect a one node cluster, as you could have a single master cluster with a bunch of unhealthy Pods that hold data nodes.

pkg/controller/elasticsearch/driver/upgrade_predicates_test.go
```diff
@@ -367,6 +367,104 @@ func TestUpgradePodsDeletion_Delete(t *testing.T) {
...
+						NodeName: "masters-1",
+						Type:     "r",
+					},
+				}),
+				podFilter: nothing,
+			},
+			deleted:                      []string{"masters-0"},
+			wantErr:                      false,
+			wantShardsAllocationDisabled: true,
+		},
```
pebrc, line 467: We are missing the case of a regular rolling upgrade i.e no version difference here.

pebrc (changes requested):
The original issue talks about upgrade but this term is used somewhat loosely as we also mean rolling restarts to roll out configuration changes by that and not just version upgrades. This PR does currently only address the version upgrade scenario as far as I can see.

pkg/controller/elasticsearch/driver/upgrade_predicates.go
```diff
@@ -218,6 +222,15 @@ var predicates = [...]Predicate{
 			deletedPods []corev1.Pod,
 			maxUnavailableReached bool,
 		) (b bool, e error) {
+			health, err := context.esState.Health()
+			if err != nil {
+				return false, err
+			}
+			_, healthyNode := context.healthyPods[candidate.Name]
+			if health.NumberOfNodes == 1 && health.Status == esv1.ElasticsearchYellowHealth && healthyNode {
```
pebrc, line 230: What I meant was that we are basing the decision on the result of `_cat/health` or `_cluster/health` here. But the internal view of cluster membership might not be consistent with the external state of things in Kubernetes. The operator will take care to do node deletions and creations before the rolling upgrade. But Pods in a multi-node cluster might be not `Running` for other reasons in which case we would proceed with the upgrade assuming it is a single node cluster and potentially drop into `red` health. I think we need to look at the actual Pods in the Kubernetes cluster e.g. with `StatefulSetList.GetActualPods`

CamiloSierraH: understood, while reviewing the code I thought that the best option would be to break a bit the Open-Closed Principle. 

Checking `predicateContext` we have nothing that could allow us to know the total number of pods on the `StatefulSet`, so I decided to add new information there (sorry OCP). 

We can add `GetActualPods` on `upgrade_pods_deletion.go` or in `upgrade.go>handleRollingUpgrades`, upgrade.go seems a better option as in the function `handleRollingUpgrades` we already calculate information as healthyPods, actualMasters.. etc.

I know I'm taking a risk and I can change it if needed.

pebrc:
Jenkins test this please

pkg/controller/elasticsearch/driver/upgrade_predicates_test.go
```diff
@@ -367,6 +378,140 @@ func TestUpgradePodsDeletion_Delete(t *testing.T) {
...
+				esVersion: "7.5.0",
+				upgradeTestPods: newUpgradeTestPods(
+					newTestPod("masters-0").isMaster(true).isData(true).isHealthy(true).needsUpgrade(true).isInCluster(true).withVersion("7.4.0"),
+					newTestPod("masters-1").isMaster(true).isData(true).isHealthy(true).needsUpgrade(true).isInCluster(true).withVersion("7.5.0"),
+				),
+				maxUnavailable: 1,
+				numberOfPods:   2,
+				health:         client.Health{Status: esv1.ElasticsearchYellowHealth, NumberOfNodes: 2},
+				shardLister: migration.NewFakeShardLister(client.Shards{
+					// One shard is not assigned on masters-0 because its version is not the expected one
```
pebrc, line 393: This is interesting, but I am not sure I understand the scenario. You mean the version of the index requires 7.4.0, but the only other node is already 7.5.0? This would be a deadlocked upgrade then? I would expect Elasticsearch to have primary on `masters-1` and replica on `masters-0` or if it is the other way round promote the replica to primary and unassigned the replica on `masters-0` because it is now potentially no longer compatible with the upgrade version of the index. But I am not 100% sure myself how this works internally in Elasticsearch.

pebrc (commented):
Almost LGTM, a few nit picks in the tests remain.

pkg/controller/elasticsearch/driver/upgrade_predicates_test.go
CamiloSierraH: Hey @pebrc , yes, the test and the name are not really clear, let me elaborate more.
I had a few issues (ESS or on-prem) with the new "Data tiers" feature, so I was thinking about how this could have an impact on ECK.
I have seen that some users (small clusters) have a few hot nodes along with one node in warm where they keep old data with no replicas.
My use case is the following: there is a cluster with 2 hot nodes and 1 warm node, when upgrading the warm node works but we lose one of the hot nodes, then the shard could not be assigned to the warm node, so we move to yellow.
So, the cluster could not finish the upgrade, as we only would have one shard/copy of the `index_a`.

what do you think about this? I'm not sure if my explanation makes sense.

I updated the name of the test and the pods name too, to better reflect the use case

Thanks

pebrc: I see what you mean now. But I think your test is not testing for that. Because you have specified your test cluster as `masters-hot-0` and `masters-warm-1` the test code assumes one of the warm nodes is missing (`masters-warm-0`) not one of the hot nodes. That means we have one unavailable node (`masters-warm-0`) and the `do_not_restart_healthy_node_if_MaxUnavailable_reached` predicate applies. Also I would assume that `needsUpgrade` should be `false` on the node already upgraded.

I would suggest to simplify the test and remove the notion of data tiers as the predicate system is not aware of them anyway. Rather I would rewrite the test to cover the scenario where we have a multi-node cluster where one Pod is temporarily unavailable (because of host failure or some other reason) and where we want to make sure that the single node exception you added does not apply in this case.

CamiloSierraH: yes agree, I updated the code reviewed the comments and settings, and the test is now a two-node cluster one pod is unavailable/unhealthy, we are in the situation only one node on the cluster so the shard is assigned but not the replica, so that could match the single node exception (but not because we are using StatefulSetList) so this is not a really single node cluster so we can not delete the pod. 

Thanks for your time and patience Peter

pebrc (approved):
LGTM! Nice work 👍

pebrc:
https://devops-ci.elastic.co/view/cloud-on-k8s/job/cloud-on-k8s-e2e-tests-tanzu/1/ I kicked off a test job to run all e2e test with this PR otherwise it's good to merge from my POV

CamiloSierraH:
Hi @pebrc !! this is great news, I'm really happy!! 

last but not least, If you agree I will update the [documentation](https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-orchestration.html#k8s-orchestration-limitations) with this change in a new PR.

PR #4795, opened 2021-08-27
Add information regarding ECK operator update will trigger pods rolling restart

kunisen:
- Have you signed the [contributor license agreement](https://www.elastic.co/contributor-agreement)? Yes
- Have you followed the [contributor guidelines](https://github.com/elastic/cloud-on-k8s/tree/master/CONTRIBUTING.md)? Yes
- If you submit code, is your pull request against master? We recommend pull requests against master. We will backport them as needed. => It's a doc PR and not a code PR.

Side note: After modifying, it will look like this:

image

### DIFF
--- a/docs/operating-eck/upgrading-eck.asciidoc
+++ b/docs/operating-eck/upgrading-eck.asciidoc
@@ -11,6 +11,20 @@ endif::[]
 [id="{p}-ga-upgrade"]
 == Upgrade to ECK {eck_version}
 
+[IMPORTANT]
+================================
+This is the upgrade guide for ECK. For Elastic Stack upgrade, please refer to this <<{p}-upgrading-stack,guidance>> for more details.
+================================
+
+[CAUTION]
+================================
+Upgrading the ECK operator results in a one-time update to existing managed resources in the cluster. This triggers a rolling restart of pods, on which Elastic Stack components are running, by Kubernetes to apply those changes. Therefore, if you have a large Elasticsearch cluster or multiple Elastic Stack deployments, this rolling restart will cause performance degradation, which might be disruptive or inconvenient.
+
+For production use, when you plan to upgrade ECK, it's important to include this pods rolling restart into time schedule consideration.
+
+Also as described later, you could temporarily exclude the Elastic Stack components from being managed by ECK operator. After the ECK upgrade is done, you must remember to remove the exclusion to ensure that your deployment is continually monitored and managed by the operator, and let the rolling restart go through.
+================================
+
 ECK reached general availability (GA) status with the link:https://www.elastic.co/blog/elastic-cloud-on-kubernetes-ECK-is-now-generally-available[release of version 1.0.0]. The latest available GA version is {eck_version}. It is compatible with the previous GA releases (1.0.x and higher) and the beta release (1.0.0-beta1), and can be upgraded in-place (<<{p}-ga-openshift, with a few exceptions>>) by applying the new set of deployment manifests. Previous alpha releases, up to and including version 0.9.0, are not compatible with the GA and beta releases and link:https://www.elastic.co/guide/en/cloud-on-k8s/1.0-beta/k8s-upgrading-eck.html[require extra work to upgrade].
 
 Before upgrading, refer to the <<release-notes-{eck_version}, release notes>> to make sure that the release does not contain any breaking changes that could affect you. The <<release-highlights-{eck_version},release highlights document>> provides more details and possible workarounds for any breaking changes or known issues in each release.
--- a/docs/operating-eck/upgrading-eck.asciidoc
+++ b/docs/operating-eck/upgrading-eck.asciidoc
@@ -20,7 +20,7 @@ This is the upgrade guide for ECK. For Elastic Stack upgrade, please refer to th
 ================================
 Upgrading the ECK operator results in a one-time update to existing managed resources in the cluster. This triggers a rolling restart of pods, on which Elastic Stack components are running, by Kubernetes to apply those changes. Therefore, if you have a large Elasticsearch cluster or multiple Elastic Stack deployments, this rolling restart could cause a performance degradation, which might be disruptive or inconvenient.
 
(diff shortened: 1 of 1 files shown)


docs/operating-eck/upgrading-eck.asciidoc
```diff
@@ -11,6 +11,20 @@ endif::[]
...
 == Upgrade to ECK {eck_version}
 
+[IMPORTANT]
+================================
+This is the upgrade guide for ECK. For Elastic Stack upgrade, please refer to this <<{p}-upgrading-stack,guidance>> for more details.
+================================
+
+[CAUTION]
+================================
+Upgrading the ECK operator results in a one-time update to existing managed resources in the cluster. This triggers a rolling restart of pods, on which Elastic Stack components are running, by Kubernetes to apply those changes. Therefore, if you have a large Elasticsearch cluster or multiple Elastic Stack deployments, this rolling restart will cause performance degradation, which might be disruptive or inconvenient.
```
pebrc, line 21: I don't think a performance degradation is a necessary consequence of every rolling restart. I think it depends on the use case.  
```suggestion
Upgrading the ECK operator results in a one-time update to existing managed resources in the cluster. This triggers a rolling restart of pods, on which Elastic Stack components are running, by Kubernetes to apply those changes. Therefore, if you have a large Elasticsearch cluster or multiple Elastic Stack deployments, this rolling restart could cause a performance degradation, which might be disruptive or inconvenient.
```

docs/operating-eck/upgrading-eck.asciidoc
```diff
@@ -11,6 +11,20 @@ endif::[]
...
+
+For production use, when you plan to upgrade ECK, it's important to include this pods rolling restart into time schedule consideration.
```
pebrc, line 23:
```suggestion
For production use, when you plan to upgrade ECK, it's important to include this rolling restart into time schedule consideration.
```
I am a bit confused by this statement. Why are time schedule consideration a concern here? Do we mean that the overall time the upgrade takes is the time to upgrade the ECK operator plus the time to roll all managed workloads/Elasticserach clusters? If so, we should maybe be explicit.

docs/operating-eck/upgrading-eck.asciidoc
```diff
@@ -11,6 +11,20 @@ endif::[]
...
+
+Also as described later, you could temporarily exclude the Elastic Stack components from being managed by ECK operator. After the ECK upgrade is done, you must remember to remove the exclusion to ensure that your deployment is continually monitored and managed by the operator, and let the rolling restart go through.
```
pebrc, line 25: We are already mentioning this a few paragraphs down. I am a bit worried if we pepper this page with admonition paragraphs that it will not necessary lead to better reader comprehension but instead readers will just start to ignore the many exclamation marks.

pebrc (commented):
Hey thanks for raising this 👍  I think I am generally in favour of this PR and highlighting the impact of an ECK upgrade in a better way. However I think we need to do some more editing before we can merge this. I have two main concerns:
1. we are repeating content verbatim in the new admonition paragraphs that is already on this page
2. we added many admonition paragraphs and thereby diluting their overall effect

I am sure however that @alaudazzi can help with this and we can find a good balance for this page.

docs/operating-eck/upgrading-eck.asciidoc
kunisen: Thanks! Indeed it depends on the use case!

docs/operating-eck/upgrading-eck.asciidoc
kunisen: > Do we mean that the overall time the upgrade takes is the time to upgrade the ECK operator plus the time to roll all managed workloads/Elasticserach clusters

Yes, your reading is correct.
I will make it explicit.

docs/operating-eck/upgrading-eck.asciidoc
kunisen: Thanks! Totally agree with this when thinking about the better reader comprehension! I will remove the duplicated part at the top.

kunisen:
Thanks @pebrc again for the review. 

I updated the statement based on the valuable suggested comment.
Now it's more visible and clearer than before.

image

docs/operating-eck/upgrading-eck.asciidoc
```diff
@@ -11,6 +11,18 @@ endif::[]
...
+[IMPORTANT]
+================================
+This is the upgrade guide for ECK. For Elastic Stack upgrade, please refer to this <<{p}-upgrading-stack,guidance>> for more details.
+================================
+
+[CAUTION]
+================================
+Upgrading the ECK operator results in a one-time update to existing managed resources in the cluster. This triggers a rolling restart of pods, on which Elastic Stack components are running, by Kubernetes to apply those changes. Therefore, if you have a large Elasticsearch cluster or multiple Elastic Stack deployments, this rolling restart could cause a performance degradation, which might be disruptive or inconvenient.
+
+For production use, when you plan to upgrade ECK, it's important to recognize that, the overall time the upgrade takes is the time to upgrade the ECK operator plus the time to roll all managed workloads/Elasticserach clusters
```
thbkrkr, line 23:
```suggestion
For production use, when you plan to upgrade ECK, it's important to recognize that, the overall time the upgrade takes is the time to upgrade the ECK operator plus the time to roll all managed workloads/Elasticsearch clusters.
```

kunisen: Thanks for the checking!

alaudazzi:
@kunisen I added a few suggestions for text and structure.
Here is a preview of the changes:
image

kunisen:
Thanks @alaudazzi  for the advice.

I am sorry that I just noticed there's already a `Caution` banner in the doc.

image

Based on that, sorry for the last minute change, but I think we could merge the statement below into that caution section.
Do you think that sounds reasonable?

image

(I should notice it earlier before I raise PR 🤦‍♂️  I only focused on the under part during `Control rolling restarts during the upgrad`... Actually we have already mentioned that twice...)

alaudazzi (approved):
LGTM, with the suggested changes.

kunisen:
Thanks @alaudazzi 
I updated the doc as per the above chat.
Sorry for the back and forth, but could you please check again to see if this time it looks good? 🙏

PR #4801, opened 2021-08-31
Remove non-functional pre-stop hook parts

### LINKED ISSUE #4698: PreStop hook logic seems incorrect
## Bug Report
**What did you do?**
When terminating ES Pod PreStop hook script `pre-stop-hook-script.sh` attempts to wait for the `POD_IP` to be removed from the list of Endpoints of the headless service

I replicated what the PreStop hook did in an ubuntu pod in the same namespace with the following:
```
while true; do getent hosts search-test-es-client  | grep $ES_POD_IP; sleep 0.5; done
```

When modifying `PRE_STOP_MAX_WAIT_SECONDS` and `PRE_STOP_ADDITIONAL_WAIT_SECONDS` you can observe that the IP remains in the Service until `PRE_STOP_MAX_WAIT_SECONDS` when the PreStop hook terminates unsuccessfully and Kubernetes continues with the Termination. This holds true even if you set `PRE_STOP_MAX_WAIT_SECONDS` to a substantial number. I went as high as 180s which is more than enough time.

I would expect to see the POD_IP get removed from the output of `getent hosts $HEADLESS_SERVICE_NAME | grep $POD_IP` before the Pod is terminated.

It appears (although documentation on headless services is sparse) that headless services do not remove the IP from the list of Endpoint IPs until **after** the Pod is terminated, not when it enters `Terminating` status. What this results in is the PreStop hook will simply wait `PRE_STOP_MAX_WAIT_SECONDS` and then continue terminating, but the IP is still in service for some time afterwards. It seems like the logic here is not working as intended. I'm not really sure how this would work given this behavior of Headless Services

**Environment**

* ECK version:

	`1.3.0`

* Kubernetes information:
GKE 1.19
	
	for each of them please give us the version you are using
```
kubectl version
Client Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.3", GitCommit:"ca643a4d1f7bfe34773c74f79527be4afd95bf39", GitTreeState:"clean", BuildDate:"2021-07-15T20:58:09Z", GoVersion:"go1.16.5", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"19+", GitVersion:"v1.19.10-gke.1600", GitCommit:"7b8e568a7fb4c9d199c2ba29a5f7d76f6b4341c2", GitTreeState:"clean", BuildDate:"2021-05-07T09:18:53Z", GoVersion:"go1.15.10b5", Compiler:"gc", Platform:"linux/amd64"}
```

* Resource definition:
```
kind: Elasticsearch
metadata:
  name: search-test
spec:
  auth: {}
  http:
    service:
      metadata:
        creationTimestamp: null
      spec: {}
    tls:
      certificate: {}
      selfSignedCertificate:
        disabled: true
  nodeSets:
  - config:
      node.data: false
      node.ingest: false
      node.master: true
      node.store.allow_mmap: true
    count: 1
    name: master-nodes
    podTemplate:
      metadata:
      spec:
        containers:
        - name: elasticsearch
        initContainers:
        - Image: busybox:latest
          command:
          - sh
          - -c
          - |
            sysctl -w vm.max_map_count=262144
          name: sysctl
          securityContext:
            privileged: true
  - config:
      node.data: true
      node.ingest: false
      node.master: false
      node.store.allow_mmap: true
    count: 1
    name: data-nodes
    podTemplate:
      spec:
        containers:
        - name: elasticsearch
        initContainers:
        - Image: busybox:latest
          command:
          - sh
          - -c
          - |
            sysctl -w vm.max_map_count=262144
          name: sysctl
          securityContext:
            privileged: true
  - config:
      node.data: false
      node.ingest: false
      node.master: false
      node.store.allow_mmap: true
    count: 2
    name: client
    podTemplate:
      spec:
        containers:
        - name: elasticsearch
        initContainers:
        - Image: busybox:latest
          command:
          - sh
          - -c
          - |
            sysctl -w vm.max_map_count=262144
          name: sysctl
          securityContext:
            privileged: true
  transport:
    service:
      metadata:
        creationTimestamp: null
      spec: {}
  updateStrategy:
    changeBudget: {}
  version: 6.8.13
```

* Logs:
```
insert operator logs or any relevant message to the issue here
```

pebrc:
Fixes #4698 

The pre-stop hook script intended to improve availability during rolling upgrades has not been doing what it is advertising to do since ECK 1.3.0. 

https://github.com/elastic/cloud-on-k8s/blob/72bfb73cb32cc41885b900187638da63fa9f4235/pkg/controller/elasticsearch/nodespec/statefulset.go#L51

We are including non-ready Pods in the headless service which we then try to use to determine when a terminating Pod has been removed from DNS (never, in this constellation). We have since added functionality in #3837 that relies on this setting for client-side node discovery (aka sniffing). 

This means that our pre-stop hook was always running the full 20+30 secs before terminating the Pod. I see two potential fixes for this problem:
1. go back to not publishing unready Pods and revert #3837
2. simplify the pre-stop hook and turn it into a simple timeout without trying to check DNS for the Pod IP (this PR) 

I have opted for sticking with 50 seconds of wait time (but I am open to cutting this down) 

NOTE: I also kept the `terminationGracePeriod` at 180s to avoid a rolling restart on ECK upgrade.

### DIFF
--- a/docs/orchestrating-elastic-stack-applications/elasticsearch/prestop.asciidoc
+++ b/docs/orchestrating-elastic-stack-applications/elasticsearch/prestop.asciidoc
@@ -19,7 +19,7 @@ It waits for an additional `PRE_STOP_ADDITIONAL_WAIT_SECONDS` (defaulting to 50)
 2. Give clients time to use the terminating Pod IP resolved just before DNS record was updated.
 3. Give kube-proxy time to refresh ipvs or iptables rules on all nodes, depending on its sync period setting.
 
-The exact behavior is configurable using environment variables, for example:
+The exact behavior is configurable using an environment variable, for example:
 
 [source,yaml,subs="attributes"]
 ----
--- a/docs/orchestrating-elastic-stack-applications/elasticsearch/prestop.asciidoc
+++ b/docs/orchestrating-elastic-stack-applications/elasticsearch/prestop.asciidoc
@@ -13,7 +13,7 @@ When an Elasticsearch `Pod` is terminated, its `Endpoint` is removed from the `S
 Moreover, kube-proxy resynchronizes its rules link:https://kubernetes.io/docs/reference/command-line-tools-reference/kube-proxy/#options[every 30 seconds by default]. During that time window of 30 seconds, the terminating Pod IP may still be used when targeting the service. Please note the resync operation itself may take some time, especially if kube-proxy is configured to use iptables with a lot of services and rules to apply.
 
 To address this issue and minimize unavailability, ECK relies on a link:https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/[PreStop lifecycle hook].
-It waits for an additional `PRE_STOP_ADDITIONAL_WAIT_SECONDS` (defaulting to 30). The additional wait time is used to:
+It waits for an additional `PRE_STOP_ADDITIONAL_WAIT_SECONDS` (defaulting to 50). The additional wait time is used to:
 
 1. Give time to in-flight requests to be completed.
 2. Give clients time to use the terminating Pod IP resolved just before DNS record was updated.
--- a/docs/orchestrating-elastic-stack-applications/elasticsearch/prestop.asciidoc
+++ b/docs/orchestrating-elastic-stack-applications/elasticsearch/prestop.asciidoc
@@ -12,9 +12,8 @@ When an Elasticsearch `Pod` is terminated, its `Endpoint` is removed from the `S
 
 Moreover, kube-proxy resynchronizes its rules link:https://kubernetes.io/docs/reference/command-line-tools-reference/kube-proxy/#options[every 30 seconds by default]. During that time window of 30 seconds, the terminating Pod IP may still be used when targeting the service. Please note the resync operation itself may take some time, especially if kube-proxy is configured to use iptables with a lot of services and rules to apply.
 
-To address this issue and minimize unavailability, ECK relies on a link:https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/[PreStop lifecycle hook]. This waits to terminate the Elasticsearch process until the `Service` DNS record does not contain the IP of the `Pod`.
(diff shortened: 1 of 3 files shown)


sebgl (approved):
LGTM
I actually like the simplicity of this more than our initial approximation that relies on the (unrelated) StatefulSet headless service DNS.

docs/orchestrating-elastic-stack-applications/elasticsearch/prestop.asciidoc
```diff
@@ -12,9 +12,8 @@ When an Elasticsearch `Pod` is terminated, its `Endpoint` is removed from the `S
...
 
-To address this issue and minimize unavailability, ECK relies on a link:https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/[PreStop lifecycle hook]. This waits to terminate the Elasticsearch process until the `Service` DNS record does not contain the IP of the `Pod`.
-First, the PreStop lifecycle hook keeps querying DNS for `PRE_STOP_MAX_WAIT_SECONDS` (defaulting to 20) until the Pod IP is no longer referenced.
-Then, it waits for an additional `PRE_STOP_ADDITIONAL_WAIT_SECONDS` (defaulting to 30). Additional wait is used to:
+To address this issue and minimize unavailability, ECK relies on a link:https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/[PreStop lifecycle hook].
+It waits for an additional `PRE_STOP_ADDITIONAL_WAIT_SECONDS` (defaulting to 50). The additional wait time is used to:
 
 1. Give time to in-flight requests to be completed.
 2. Give clients time to use the terminating Pod IP resolved just before DNS record was updated.
```
david-kow, line 19: Line 22 should say `environment variable` instead of `environment variableS`:

`The exact behavior is configurable using environment variable, for example:`

david-kow (approved):
I'm also +1 on this change. Should we mark it as breaking though? For users that set `PRE_STOP_ADDITIONAL_WAIT_SECONDS` to `0` this change will mean going from somewhat safe 20 seconds wait to no waiting at all.

barkbay (approved):
lgtm

PR #4803, opened 2021-08-31
Improve Elasticsearch cluster health API call when timeout

### LINKED ISSUE #4507: GetClusterHealthWaitForAllEvents may return empty/default value
If the Elasticsearch endpoint `"/_cluster/health?wait_for_events=languid&timeout=0s"` returns 408 (Timeout) then `GetClusterHealthWaitForAllEvents` swallows the error and returns an empty/default value. [`TimedOut`](https://github.com/elastic/cloud-on-k8s/blob/b48b1afd26b54a5891183904bc29a9128a46f5d1/pkg/controller/elasticsearch/client/model.go#L32) may hold the default value `false` while the API actually returned `true`:

```json
{
	"cluster_name": "elasticsearch-sample",
	"status": "green",
	"timed_out": true,
	"number_of_nodes": 3,
	"number_of_data_nodes": 3,
	"active_primary_shards": 12,
	"active_shards": 24,
	"relocating_shards": 0,
	"initializing_shards": 0,
	"unassigned_shards": 0,
	"delayed_unassigned_shards": 0,
	"number_of_pending_tasks": 0,
	"number_of_in_flight_fetch": 0,
	"task_max_waiting_in_queue_millis": 0,
	"active_shards_percent_as_number": 100.0
}
```

The returned value is not initialized because [we do not attempt to parse/decode the body and return immediately](https://github.com/elastic/cloud-on-k8s/blob/cd6ce28679fd4c2534a27b12d5738c13f04a1a95/pkg/controller/elasticsearch/client/base.go#L116) when there is an error:

```golang
func (c *baseClient) request(
	...
	responseObj interface{},
) error {
	resp, err := c.doRequest(ctx, request)
	if err != nil {
		return err // <--- here
	}
    ...
	if responseObj != nil {
		if err := json.NewDecoder(resp.Body).Decode(responseObj); err != nil {
			return err
		}
	}
```

Should we attempt to decode the body as a best effort even if there is an error ?

thbkrkr:
Adds a new `unsafeRequest` function that accepts a `skipErrFunc func(error) bool` argument used to know if we should decode the response even when the status code is not 2xx.

Resolves #4507.

### DIFF
--- a/pkg/controller/elasticsearch/client/base.go
+++ b/pkg/controller/elasticsearch/client/base.go
@@ -90,8 +90,8 @@ func (c *baseClient) delete(ctx context.Context, pathWithQuery string, in, out i
 // request performs a new http request
 //
 // if requestObj is not nil, it's marshalled as JSON and used as the request body
-// if responseObj is not nil, it should be a pointer to an struct. the response body will be unmarshalled from JSON
-// into this struct.
+// if responseObj is not nil, it should be a pointer to an struct. The response body will be unmarshalled from JSON
+// into this struct if the status code of the response is 2xx.
 func (c *baseClient) request(
 	ctx context.Context,
 	method string,
@@ -114,6 +114,53 @@ func (c *baseClient) request(
 	}
 
 	resp, err := c.doRequest(ctx, request)
+	if err != nil {
+		return err
+	}
+
+	defer resp.Body.Close()
+
+	if responseObj != nil {
+		if err := json.NewDecoder(resp.Body).Decode(responseObj); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// unsafeRequest performs a new http request and decodes the payload if the status code of the response is 2xx or if the
+// given skipErrFunc function returns true.
+//
+// if requestObj is not nil, it's marshalled as JSON and used as the request body
+// if responseObj is not nil, it should be a pointer to an struct. The response body will be unmarshalled from JSON
+// into this struct if the status code of the response is 2xx or if the given skipErrFunc function returns true.
+func (c *baseClient) unsafeRequest(
+	ctx context.Context,
+	method string,
+	pathWithQuery string,
+	requestObj,
+	responseObj interface{},
+	skipErrFunc func(error) bool,
+) error {
+	var body io.Reader = http.NoBody
+	if requestObj != nil {
+		outData, err := json.Marshal(requestObj)
+		if err != nil {
+			return err
+		}
+		body = bytes.NewBuffer(outData)
+	}
+
+	request, err := http.NewRequest(method, stringsutil.Concat(c.Endpoint, pathWithQuery), body) //nolint:noctx
+	if err != nil {
+		return err
+	}
+
+	resp, err := c.doRequest(ctx, request)
+	if skipErrFunc(err) {
+		err = nil
+	}
 	if err != nil {
 		return err
 	}
--- a/pkg/controller/elasticsearch/client/base.go
+++ b/pkg/controller/elasticsearch/client/base.go
@@ -72,70 +72,28 @@ func (c *baseClient) doRequest(context context.Context, request *http.Request) (
 }
 
 func (c *baseClient) get(ctx context.Context, pathWithQuery string, out interface{}) error {
-	return c.request(ctx, http.MethodGet, pathWithQuery, nil, out)
+	return c.request(ctx, http.MethodGet, pathWithQuery, nil, out, nil)
 }
 
 func (c *baseClient) put(ctx context.Context, pathWithQuery string, in, out interface{}) error { //nolint:unparam
-	return c.request(ctx, http.MethodPut, pathWithQuery, in, out)
+	return c.request(ctx, http.MethodPut, pathWithQuery, in, out, nil)
 }
 
 func (c *baseClient) post(ctx context.Context, pathWithQuery string, in, out interface{}) error {
-	return c.request(ctx, http.MethodPost, pathWithQuery, in, out)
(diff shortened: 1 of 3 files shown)


pkg/controller/elasticsearch/client/base.go
```diff
@@ -108,12 +108,59 @@ func (c *baseClient) request(
...
+	return nil
+}
+
+// unsafeRequest performs a new http request and decodes the payload if the status code of the response is 2xx or if the
+// given skipErrFunc function returns true.
+//
+// if requestObj is not nil, it's marshalled as JSON and used as the request body
+// if responseObj is not nil, it should be a pointer to an struct. The response body will be unmarshalled from JSON
+// into this struct if the status code of the response is 2xx or if the given skipErrFunc function returns true.
+func (c *baseClient) unsafeRequest(
```
pebrc, line 138: Can we find another name for this? Is it really unsafe to parse the body in certain error cases? Idk maybe `requestIgnoringError(context.Context, func(error) bool, string, interface{}, interface{})` 

An alternative idea would be to modify the existing request function to take a predicate that by default only returns false. But I guess that would be a bit more intrusive as you would have to change all invocations.

thbkrkr: > Can we find another name for this? Is it really unsafe to parse the body in certain error cases?

Sure but I don't find a good name! I think it could be unsafe depending the `checkSkipErr`. With this name I wanted to highlight that you really need know what you are doing if you use this function.

> An alternative idea would be to modify the existing request function to take a predicate that by default only returns 
false. But I guess that would be a bit more intrusive as you would have to change all invocations.

That was my initial thought but looking at it again, given that we have little intermediates functions get/put/post/delete functions, it is not that intrusive. I'll update the PR in that direction.

thbkrkr:
Jenkins test this please.

barkbay:
FWIW here is the unit test I'm using to validate the function if you want to include it in the PR:

```golang
func TestGetClusterHealthWaitForAllEvents(t *testing.T) {
	expectedPath := "/_cluster/health"
	expectedRawQuery := "wait_for_events=languid&timeout=0s"
	type fields struct {
		client Client
	}
	tests := []struct {
		name    string
		fields  fields
		want    Health
		wantErr bool
	}{
		{
			name: "other than 408 should not be ignored",
			fields: fields{
				client: NewMockClient(version.MustParse("7.14.0"), func(req *http.Request) *http.Response {
					require.Equal(t, expectedPath, req.URL.Path)
					require.Equal(t, expectedRawQuery, req.URL.RawQuery)
					return &http.Response{
						StatusCode: 500,
					}
				}),
			},
			wantErr: true,
		},
		{
			name: "408 should be ignored",
			fields: fields{
				client: NewMockClient(version.MustParse("7.14.0"), func(req *http.Request) *http.Response {
					require.Equal(t, expectedPath, req.URL.Path)
					require.Equal(t, expectedRawQuery, req.URL.RawQuery)
					return &http.Response{
						StatusCode: 408,
						Body: ioutil.NopCloser(strings.NewReader(
							`
{
	"cluster_name": "elasticsearch-sample",
	"status": "green",
	"timed_out": true,
	"number_of_nodes": 3,
	"number_of_data_nodes": 3,
	"active_primary_shards": 12,
	"active_shards": 24,
	"relocating_shards": 1,
	"initializing_shards": 2,
	"unassigned_shards": 3,
	"delayed_unassigned_shards": 4,
	"number_of_pending_tasks": 5,
	"number_of_in_flight_fetch": 6,
	"task_max_waiting_in_queue_millis": 7,
	"active_shards_percent_as_number": 42.0
}`,
						)),
						Header:  make(http.Header),
						Request: req,
					}
				}),
			},
			want: Health{
				ClusterName:                 "elasticsearch-sample",
				Status:                      "green",
				TimedOut:                    true,
				NumberOfNodes:               3,
				NumberOfDataNodes:           3,
				ActivePrimaryShards:         12,
				ActiveShards:                24,
				RelocatingShards:            1,
				InitializingShards:          2,
				UnassignedShards:            3,
				DelayedUnassignedShards:     4,
				NumberOfPendingTasks:        5,
				NumberOfInFlightFetch:       6,
				TaskMaxWaitingInQueueMillis: 7,
				ActiveShardsPercentAsNumber: 42.0,
			},
		},
	}
	for _, tt := range tests {
			got, err := tt.fields.client.GetClusterHealthWaitForAllEvents(context.Background())
			if (err != nil) != tt.wantErr {
				t.Errorf("GetClusterHealthWaitForAllEvents() error = %v, wantErr %v", err, tt.wantErr)
				return
			}
			if !reflect.DeepEqual(got, tt.want) {
				t.Errorf("GetClusterHealthWaitForAllEvents() = %v, want %v", got, tt.want)
			}
		})
	}
}
```

pkg/controller/elasticsearch/client/base.go
```diff
@@ -114,6 +115,9 @@ func (c *baseClient) request(
 	}
 
 	resp, err := c.doRequest(ctx, request)
+	if skipErrFunc != nil && skipErrFunc(err) {
+		err = nil
```
barkbay, line 119: I think we should offer an option to print the error, either with a log at the debug level:

```golang
	if err != nil {
		log.V(1).Error(err, "...")
	}
````

or by using the `multierror` package:

```golang
	resp, requestErr := c.doRequest(ctx, request)
	if !skipErrFunc(requestErr) {
		return requestErr
	}

	defer resp.Body.Close()

	if responseObj != nil {
		if err := json.NewDecoder(resp.Body).Decode(responseObj); err != nil {
			return multierror.Append(err, requestErr)
		}
	}
```

thbkrkr: fc3940e

Issue #4790, opened 2021-08-23
fleet and elastic agent doesn't work without a ca.crt


Raboo:
## Bug Report
**What did you do?*

I have a elasticsearch + kibana deployment that uses Letsencrypt ACME certificate manged via cert-manager.

```yaml
apiVersion: cert-manager.io/v1
kind: Certificate
metadata:
  name: elasticsearch0
  namespace: elastic0
spec:
  secretName: elasticsearch0-tls
  issuerRef:
    kind: ClusterIssuer
    name: letsencrypt
  dnsNames:
  - 'elasticsearch0.k8s.mydomain.net'
--- 
apiVersion: elasticsearch.k8s.elastic.co/v1
kind: Elasticsearch
metadata:
  name: elastic0
  namespace: elastic0
spec:
...
  http:
    service:
      spec:
        type: LoadBalancer
    tls:
      certificate:
        secretName: elasticsearch0-tls
```

Deploying fleet server and elastic agent when using Letsencrypt ACME certificate manged via cert-manager doesn't work because the generated certificate secret doesn't contain a ca.crt and is expected by the current manifests. The tls.crt field in the cert-manager generated secrets contains the entire chain, from the leaf cert to the intermediate cert (not the root cert). https://github.com/jetstack/cert-manager/issues/1571#issuecomment-486661887

**What did you expect to see?**
I expect fleet server and elastic agent to work without a private ca or self-signed tls setup.
I.e. to work with letsencrypt acme certificates generated with the help of cert-manager.

**What did you see instead? Under which circumstances?**
The pods fail to start because they are trying to copy ca.crt from a secret that doesn't have a ca.crt.
The pod manifest looks something like this

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: fleet-server0-agent-76cbcdd5b5-swcdm
  namespace: elastic0
spec:
  ...
  containers:
  - command:
    - /usr/bin/env
    - bash
    - -c
    - |
      #!/usr/bin/env bash
      set -e
      cp /mnt/elastic-internal/elasticsearch-association/elastic0/elastic0/certs/ca.crt /etc/pki/ca-trust/source/anchors/
      update-ca-trust
      /usr/bin/tini -- /usr/local/bin/docker-entrypoint -e
...
```

And startup fails with this message

```
cp: cannot stat '/mnt/elastic-internal/elasticsearch-association/elastic0/elastic0/certs/ca.crt': No such file or directory 
```

**Environment**

[...]

PR #4807, opened 2021-09-01
Fix handling CAs in Agent controller

david-kow:
When Elasticsearch and/or Kibana have a TLS certificate signed by a well-known CA and the Secret used for their TLS configuration doesn't contain `ca.crt` Agent in Fleet mode fails to start.

This PR fixes two related bugs:
1. container command will not try to copy CA file if it's not present
2. `fleet-setup.yml` will not be populated with CA paths if CA files are not present

Some more context available in https://github.com/elastic/cloud-on-k8s/issues/4790#issuecomment-910463148.

Fixes https://github.com/elastic/cloud-on-k8s/issues/4790.

### DIFF
--- a/pkg/controller/agent/config_test.go
+++ b/pkg/controller/agent/config_test.go
@@ -78,7 +78,7 @@ func TestBuildFleetSetupKibanaConfig(t *testing.T) {
 			client:  client,
 		},
 		{
-			name:  "kibana ref present, kibana with ca populated",
+			name:  "kibana ref present, kibana without ca populated",
 			agent: *assocWithoutCa.Agent,
 			wantCfg: map[string]interface{}{
 				"fleet": map[string]interface{}{
@@ -92,7 +92,7 @@ func TestBuildFleetSetupKibanaConfig(t *testing.T) {
 			client:  client,
 		},
 		{
-			name:  "kibana ref present, kibana without ca populated",
+			name:  "kibana ref present, kibana with ca populated",
 			agent: *assocWithCa.Agent,
 			wantCfg: map[string]interface{}{
 				"fleet": map[string]interface{}{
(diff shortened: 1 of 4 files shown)


pkg/controller/agent/pod.go
```diff
@@ -294,10 +294,12 @@ func getAssociatedFleetServer(params Params) (commonv1.Associated, error) {
 func trustCAScript(caPath string) string {
 	return fmt.Sprintf(`#!/usr/bin/env bash
 set -e
-cp %s /etc/pki/ca-trust/source/anchors/
-update-ca-trust
+if [[ -f %s ]]; then
+  cp %s /etc/pki/ca-trust/source/anchors/
+  update-ca-trust
+fi
```
pebrc, line 300:
```suggestion
if [[ -f %[1]s ]]; then
  cp %[1]s /etc/pki/ca-trust/source/anchors/
  update-ca-trust
fi
```

pkg/controller/agent/pod.go
```diff
@@ -294,10 +294,12 @@ func getAssociatedFleetServer(params Params) (commonv1.Associated, error) {
...
 /usr/bin/tini -- /usr/local/bin/docker-entrypoint -e
-`, caPath)
+`, caPath, caPath)
```
pebrc, line 302:
```suggestion
`, caPath)
```

pkg/controller/agent/config_test.go
```diff
@@ -64,8 +78,22 @@ func TestBuildFleetSetupKibanaConfig(t *testing.T) {
 			client:  client,
 		},
 		{
-			name:  "kibana ref present",
-			agent: *assoc.Agent,
+			name:  "kibana ref present, kibana with ca populated",
+			agent: *assocWithoutCa.Agent,
```
pebrc, line 82: It's the other way round this one is the one withou CA and the one on L95 is with CA.

PR #4819, opened 2021-09-09
Update supported K8s versions for upcoming 1.8.0 release

sebgl:
Starting version 1.8.0, we support Kubernetes 1.18-1.22, and Openshift 3.11, 4.4-4.8.


pebrc (approved):
LGTM (and I guess we want an extra note in the release notes about dropping OCP 3.11 support?)

PR #4823, opened 2021-09-09
Remove permissions not used by the operator

david-kow:


thbkrkr:
Jenkins test this please

david-kow:
run full pr build

thbkrkr:
Jenkins test this please

deploy/eck-operator/templates/_helpers.tpl
```diff
@@ -223,8 +222,6 @@ RBAC permissions
   - elasticsearches
   - elasticsearches/status
   - elasticsearches/finalizers # needed for ownerReferences with blockOwnerDeletion on OCP
-  - enterpriselicenses
-  - enterpriselicenses/status
```
pebrc, line 227: 😮

pebrc (commented):
What about delete permissions on the CRDs? Is there a reason to keep those?

david-kow:
> What about delete permissions on the CRDs? Is there a reason to keep those?

Not sure if I understand, we do watch/get/update them and their statuses. Or you've meant thinning down the verbs?

pebrc:
> > What about delete permissions on the CRDs? Is there a reason to keep those?
> 
> Not sure if I understand, we do watch/get/update them and their statuses. Or you've meant thinning down the verbs?

Yes removing the delete verb was what I was thinking of.

david-kow:
> > > What about delete permissions on the CRDs? Is there a reason to keep those?
> > 
> > 
> > Not sure if I understand, we do watch/get/update them and their statuses. Or you've meant thinning down the verbs?
> 
> Yes removing the delete verb was what I was thinking of.

Oh, of course, I read it as "deleting CRD permissions", my bad.

docs/operating-eck/eck-permissions.asciidoc
```diff
@@ -0,0 +1,150 @@
+:page_id: operator-config
+ifdef::env-github[]
```
pebrc, line 2: This page seems identical to the existing `operator-config` page. Also it does not seem to be linked anywhere. Is that maybe a file you forgot to delete?

pebrc (changes requested):
I think something is off with the doc page you added. Maybe a Git mishap?

docs/operating-eck/eck-permissions.asciidoc
david-kow: Yes, git mishap, thanks for noticing.

pebrc (approved):

PR #4830, opened 2021-09-10
Skip beat tests on OCP3

thbkrkr:
We skip all Beat tests to reduce maintenance burden as Beat on OpenShift3
requires specific securityContext due to hostPath volumes.

Relates to #4128.

### DIFF
--- a/test/e2e/test/beat/builder.go
+++ b/test/e2e/test/beat/builder.go
@@ -51,6 +51,12 @@ type Builder struct {
 }
 
 func (b Builder) SkipTest() bool {
+	// Beat on OpenShift3 requires specific securityContext due to hostPath volumes.
+	// Skipping all Beat tests to reduce maintenance burden.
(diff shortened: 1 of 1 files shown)


test/e2e/test/beat/builder.go
```diff
@@ -51,6 +51,12 @@ type Builder struct {
 }
 
 func (b Builder) SkipTest() bool {
+	// Beat on OpenShift3 requires specific securityContext due to hostPath volumes.
+	// Skipping all Beat tests to reduce maintenance burden.
+	if test.Ctx().Provider == "ocp3" {
+		return true
+	}
+
```
pebrc, line 59: What about Agent won't it run into the same issues?

thbkrkr: The agent tests in `agent/recipes_test.go` are already skipped (#4388) and the others work.

pebrc (approved):

Issue #4832, opened 2021-09-13
Fleet Server does not work with Kibana TLS disabled


zikphil:
Hello,

I have attempted to run the fleet server using the documentation here: [https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-elastic-agent-fleet-configuration.html](https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-elastic-agent-fleet-configuration.html).

I am running Kibana with TLS disabled since I do SSL termination on my AWS ALB. It does not seem to detect that I have turned it off because in the fleet-server-config secret it still references a CA path which does not exists causing the container to crash with the following output:

`Error: 1 error: open /mnt/elastic-internal/kibana-association/tools/xxx/certs/ca.crt: no such file or directory reading <nil>`

Any workaround that you can think of?

thbkrkr:
Hi,

This is a bug (duplicates #4790) fixed by #4807 which will be released in the next ECK 1.8.0.

Sorry for the inconvenience.

thbkrkr:
Thanks to you, I discovered another bug that we had missed. If you disable TLS on Kibana, then the association Agent<->Kibana has no CA and it breaks the `volumeMounts` of the Agent container because we `return` instead of `continue` to populate the slice of volumes: https://github.com/elastic/cloud-on-k8s/blob/3a572ec460765a5ba66efd0f5ab0e74f5dd6632d/pkg/controller/agent/pod.go#L254-L258.

zikphil:
@thbkrkr You are very welcome :) Is there a way for me to deploy the "master branch" version of the operator?

thbkrkr:
> Is there a way for me to deploy the "master branch" version of the operator?

It should be possible if you have a Docker registry.

- Clone this repository
- Install the [requirements](https://github.com/elastic/cloud-on-k8s/blob/master/dev-setup.md#requirements)
- `export REGISTRY=<docker_image_registry>; export REGISTRY_NAMESPACE=<docker_image_repository>`
- `make docker-build docker-push apply-operator`

Or deploy this manifest that I prepared for you:
```
kubectl apply -f https://gist.githubusercontent.com/thbkrkr/a1ed93ad4b6d40a00654a7b5199f4c40/raw/98f90c07a93f4ea547ba27bc76f25d75eb87f302/eck-1.9.0-SNAPSHOT-5058f86d.yml
```

thbkrkr:
[...]

PR #4833, opened 2021-09-14
Fix Agent volumes when an association has no CA

thbkrkr:
14aa58f is the fix. In a6be319 I took the opportunity to do a little refactoring (remove the warnings due to the unnecessary `vols` slice allocation and the conflict between the package and the variable name for `association`).

Resolves https://github.com/elastic/cloud-on-k8s/issues/4832#issuecomment-919051998.


pebrc (approved):

sebgl (approved):
👍

david-kow (approved):

PR #4846, opened 2021-09-18
Add support for Tanzu to deployer

pebrc:
This adds support for Tanzu to deployer. ~This has not been tested on CI yet but only from a local machine. This PR also does not add yet any CI job (pending successful testing on CI).~ There is some room for additional refactorings which I have left out to keep the PR easier to review.  The refactoring I had in mind centres mainly around splitting up the `runner` namespace and factoring out provider specific bits into separate packages. 

Regarding the mechanics of running Tanzu: I am using a private container repository for the installer image therefore deviating from the concept of building client images on the fly as need which we adopted for `kind` and `ocp`. This is mainly because we cannot share the Tanzu installer in a publicly accessible Docker image due to T&C restrictions around redistribution. Using this image we spin up a management cluster in Azure from which we can spin up the workload cluster which is then used to run our e2e test suite. While Tanzu also supports AWS I found the resource group abstraction in Azure helpful for a simpler cleanup of the created infrastructure and I ran into IAM issues on AWS with the profiles we are using, so I followed the path of least resistance here.

### DIFF
--- a/hack/deployer/runner/tanzu.go
+++ b/hack/deployer/runner/tanzu.go
@@ -108,8 +108,9 @@ type TanzuDriver struct {
 	acrName             string
 	azureStorageAccount string
 	azureCredentials    azureCredentials
-	// runtime state
-	installerStateDir string
+	// runtime state paths relative to deployers environment
+	installerStateDirPath     string
+	installerStateDirBasename string
 }
 
 func (t *TanzuDriver) Execute() error {
@@ -149,16 +150,16 @@ func (t *TanzuDriver) GetCredentials() error {
 // copies it to the main kube config (typically $HOME/.kube/config) where it is merged with existing configs.
 func (t *TanzuDriver) copyKubeconfig() error {
 	workloadConfigFile := "workload-kubeconfig"
-	tanzuContainerKubeconfigPath := filepath.Join("/root", workloadConfigFile)
+	tanzuContainerKubeconfigPath := t.tanzuContainerPath(workloadConfigFile)
 	if err := t.dockerizedTanzuCmd("cluster", "kubeconfig", "get", t.plan.ClusterName,
 		"--admin", "--export-file", tanzuContainerKubeconfigPath).Run(); err != nil {
 		return err
 	}
-	ciContainerKubeconfigPath := filepath.Join(t.installerStateDir, workloadConfigFile)
+	ciContainerKubeconfigPath := t.deployerContainerPath(workloadConfigFile)
 	return mergeKubeconfig(ciContainerKubeconfigPath)
 }
 
-// perpareCLI prepares the tanzu CLI by installing the necessary plugins.
+// perpareCLI prepares the tanzu/az CLI by installing the necessary plugins and setting up configuration
 func (t *TanzuDriver) perpareCLI() error {
 	log.Println("Installing Tanzu CLI plugins")
 	return t.dockerizedTanzuCmd("plugin", "install", "--local", "/", "all").Run()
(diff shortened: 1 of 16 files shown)


hack/deployer/runner/tanzu.go
```diff
@@ -0,0 +1,420 @@
...
+
+// setup is a list of functions that need to run before any installer commands can run.
+func (t *TanzuDriver) setup() []func() error {
+	return []func() error{
+		t.loginToAzure,
+		t.loginToContainerRegistry,
+		t.ensureWorkDir,
+		t.ensureStorageContainer,
+		t.restoreInstallerState,
+		t.perpareCLI,
```
thbkrkr, line 302: It looks like we need to install the plugins (`prepareCLI()`) before we can call `az storage azcopy blob download` (`restoreInstallerState()`) otherwise the `az` CLI prompts you to install the storage-preview extension.
```
2021/09/20 09:57:47 Using installer state dir: /root/thb-dev-cluster3727324114   
2021/09/20 09:57:50 Reusing existing storage container to persist installer state
2021/09/20 09:57:50 Restoring installer state from storage container if any                                                                                                                                                         
exit status 2                                                                               
make: *** [Makefile:351: bootstrap-k8s] Error 1

# I manually ran the az command to get the error
root@fun:/go/src/github.com/elastic/cloud-on-k8s# az storage azcopy blob download -c ***** --account-name **** -s '*' -d "/root/*********"                     
The command requires the extension storage-preview. Do you want to install it now? The command will continue to run after the extension is installed. (Y/n):
```
```suggestion
	return []func() error{
		t.loginToAzure,
		t.loginToContainerRegistry,
		t.ensureWorkDir,
		t.ensureStorageContainer,
		t.prepareCLI,
		t.restoreInstallerState,
```

thbkrkr: I'm completely off the mark, Tanzu CLI plugins != Azure CLI extensions :smile:

thbkrkr: We need to run `az extension add --name storage-preview`.

pebrc: Yes! I realised this as well. I started testing on CI today. Not sure why I did not hit it locally proabably because I already had installed the extension during development at some point. There is also an issue with the volume mounts when using a Docker volume. I will push additional commits.

hack/deployer/runner/tanzu.go
```diff
@@ -0,0 +1,420 @@
...
+			context := fmt.Sprintf("%s@%s", user, t.plan.ClusterName)
+			return removeKubeconfig(context, t.plan.ClusterName, user)
+		},
+	})
+}
+
+// dockerizedTanzuCmd runs a tanzu CLI command inside a Docker container based off the Tanzu installer image.
+func (t *TanzuDriver) dockerizedTanzuCmd(args ...string) *Command {
+	params := map[string]interface{}{
+		"SharedVolume":        filepath.Join(SharedVolumeName(), filepath.Base(t.installerStateDir)),
```
thbkrkr, line 273: This doesn't work for me (Docker 19.03.13).
```
docker: Error response from daemon: create eck-ci-home-002657483/thb-dev-cluster2047199651: "eck-ci-home-002657483/thb-dev-cluster2047199651" includes invalid characters for a local volume name, only "[a-zA-Z0-9][a-zA-Z0-9_.-]" are allowed. If you intended to pass a host directory, use absolute path.
```

I don't think we can mount a given directory of a volume, we have to use the volume name.
```suggestion
		"SharedVolume":        SharedVolumeName(),
```

So, as an alternative we could make the `HOME` equals to `filepath.Join("\root", filepath.Base(t.installerStateDir))`?

pebrc: Yes that is the volume issue I mentioned above. Yes changing `$HOME` in the container might work, I also thought of just changing the bind mount for the case where we are not mounting a Docker volume. 

So in local mode (`make bootstrap-cloud`) we would use `$HOME/$installerStateDir` and in CI mode when we mount the named volume it would just be the root of that volume.

pebrc: I went with your proposal in the end @thbkrkr and changed the $HOME directory in the Tanzu container.

thbkrkr (approved):
:rocket:

PR #4857, opened 2021-09-21
Update Kind to 0.11.1 and test K8s 1.22

sebgl:
This commit updates Kind version to 0.11.1, along with the corresponding
node images referenced in the 0.11.1 release: https://github.com/kubernetes-sigs/kind/releases.

It also adds a stage for Kubernetes 1.22.0 (ipv4 and ipv6 variants).

Fixes https://github.com/elastic/cloud-on-k8s/issues/4855.

PR #4862, opened 2021-09-21
Use rbac.authorization.k8s.io/v1 rather than rbac.authorization.k8s.i…

sebgl:
…o/v1beta1

The v1beta1 resources are no longer served by K8s 1.22: https://kubernetes.io/docs/reference/using-api/deprecation-guide/#rbac-resources-v122.
It turns out we were already using v1 everywhere except in this particular case.
Openshift 3.11 is also using v1.

Should fix the E2E test failure in https://github.com/elastic/cloud-on-k8s/pull/4857#issuecomment-923798966.

### DIFF
--- a/config/e2e/monitoring.yaml
+++ b/config/e2e/monitoring.yaml
@@ -148,7 +148,7 @@ metadata:
   labels:
     k8s-app: metricbeat
 ---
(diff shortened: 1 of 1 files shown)


barkbay (approved):
lgtm

PR #4876, opened 2021-09-23
E2E: use eck-diagnostics instead of shell script

pebrc:
We should use the new diagnostics tool internally in our e2e test suite before sharing it with users. This replaces `eck-dump.sh` and the custom code to run stack diagnostics with the new tool. It also removes the kind binaries from the CI images as we have already moved them to dedicated images inside deployer.


thbkrkr:
As this now generates a `.zip`, I think we also need to update the `pattern: "*.tgz"` in the [pipeline definitions](https://github.com/elastic/cloud-on-k8s/tree/master/.ci/pipelines).

pebrc:
> As this now generates a `.zip`, I think we also need to update the `pattern: "*.tgz"` in the [pipeline definitions](https://github.com/elastic/cloud-on-k8s/tree/master/.ci/pipelines).

Ah good point. I forgot that we changed the archive format ...

thbkrkr (approved):
LGTM

pebrc:
I have not fully tested this on CI but only in a limited simulation locally. Going to risk the merge nonetheless.

PR #4889, opened 2021-09-28
Move to env vars based Fleet mode configuration

david-kow:
Currently, configuration of Fleet is passed to the Elastic Agent in `fleet-setup.yml` file. This is convenient because it's structured and the entire config lives in a single file. Unfortunately, it's also difficult to override values from the file as it takes precedence over environmental variables. Any override requires manipulating `fleet-setup.yml` file to either modify or remove a setting. Users might want to do it when:
- disabling Fleet setup
- using pre-generated enrollment tokens
- using different ES/FS URL to avoid TLS certificate failures (when Service URL is not on SAN list)

To make creating overrides easier, this PR removes creating `fleet-setup.yml` file and makes the operator pass the config through Pods environmental variables. This also moves us from the undocumented to [documented](https://www.elastic.co/guide/en/fleet/current/agent-environment-variables.html) way of configuring the Elastic Agent. Environment variables are easy to override using `podTemplate` spec in the CRD.

This is a breaking change and will require users that are modifying `fleet-setup.yml` file as a part of Pod start up to:
1. change their modification logic to expect `fleet-setup.yml` file/mount to be absent
1. set the modified values through environmental variables
2. upgrade the operator
3. remove their modification logic

The above steps will result in upgrade without interruptions. Having said that, I'd expect very few users to be affected by this change and to have to follow those steps.


david-kow:
`java.lang.IllegalStateException: JENKINS-37121: something already locked /var/lib/jenkins/workspace/cloud-on-k8s-pr` - I'll retry in a while.

david-kow:
jenkins test this please

pebrc:
Just to better understand: why did we go with a file based approach to begin with?
What about security implications? IIRC correctly agent has no facility for secure settings so what will now be in an environment variable used to be in a file mounted in the Pod. Both requiring roughly the same degree of access to read i.e. the ability to exec into the pod for example or the ability to access the underlying secrets. So no change here correct?

david-kow:
> Just to better understand: why did we go with a file based approach to begin with? What about security implications? IIRC correctly agent has no facility for secure settings so what will now be in an environment variable used to be in a file mounted in the Pod. Both requiring roughly the same degree of access to read i.e. the ability to exec into the pod for example or the ability to access the underlying secrets. So no change here correct?

Initially I thought it's better to have a single file so that all the settings are easily accessible/visible in a single place instead of having them spread over 10-15 environmental variables. But given that the most common use cases require to change only 0-4 variables (enrolment token, policy name, Kibana/ES CA) depending on what the customer needs I find it ok to do through env vars.

Regarding security implications, yes, I believe there is no difference with this change.

pkg/controller/agent/pod.go
```diff
@@ -323,3 +320,111 @@ func certificatesDir(association commonv1.Association) string {
...
+		fleetCfg = append(fleetCfg, corev1.EnvVar{Name: "FLEET_CA", Value: path.Join(FleetCertsMountPath, certificates.CAFileName)})
+		fleetCfg = append(fleetCfg, corev1.EnvVar{Name: "FLEET_URL", Value: fleetURL})
+	} else if agent.Spec.FleetServerRef.IsDefined() {
+		assoc, err := association.SingleAssociationOfType(agent.GetAssociations(), commonv1.FleetServerAssociationType)
+		if err != nil {
+			return nil, err
+		}
+
+		if assoc != nil {
+			fleetCfg = append(fleetCfg, corev1.EnvVar{Name: "FLEET_CA", Value: path.Join(certificatesDir(assoc), CAFileName)})
```
pebrc, line 394: Make these `FLEET_*` thingies constants?

pkg/controller/agent/pod.go
```diff
@@ -323,3 +320,111 @@ func certificatesDir(association commonv1.Association) string {
...
+			types.NamespacedName{Namespace: agent.Namespace, Name: HTTPServiceName(agent.Name)},
+			agent.Spec.HTTP.Protocol(),
+		)
+		if err != nil {
+			return nil, err
+		}
+
+		fleetCfg = append(fleetCfg, corev1.EnvVar{Name: "FLEET_CA", Value: path.Join(FleetCertsMountPath, certificates.CAFileName)})
+		fleetCfg = append(fleetCfg, corev1.EnvVar{Name: "FLEET_URL", Value: fleetURL})
+	} else if agent.Spec.FleetServerRef.IsDefined() {
```
pebrc, line 387: Maybe a comment would help the casual reader remember the different modes we have: 

* Standalone agent
* Agent acting as Fleet server
* Agent enrolled in Fleet (I think that is what `FleetServerRef` indicates

pkg/controller/agent/pod.go
```diff
@@ -323,3 +320,111 @@ func certificatesDir(association commonv1.Association) string {
...
+	esExpected := len(agent.Spec.ElasticsearchRefs) > 0 && agent.Spec.ElasticsearchRefs[0].IsDefined()
+	if esExpected {
+		esConnectionSettings, err := extractConnectionSettings(agent, client, commonv1.ElasticsearchAssociationType)
+		if err != nil {
+			return nil, err
+		}
+
+		fleetServerCfg = append(fleetServerCfg, corev1.EnvVar{Name: "FLEET_SERVER_ELASTICSEARCH_HOST", Value: esConnectionSettings.host})
+		fleetServerCfg = append(fleetServerCfg, corev1.EnvVar{Name: "FLEET_SERVER_ELASTICSEARCH_USERNAME", Value: esConnectionSettings.username})
+		fleetServerCfg = append(fleetServerCfg, corev1.EnvVar{Name: "FLEET_SERVER_ELASTICSEARCH_PASSWORD", Value: esConnectionSettings.password})
```
pebrc, line 421: This means we have passwords in clear text in the Pod spec, which I think it is not great. Should we not at least mount them from a secret?

pebrc (commented):
I think we should also have section in the ECK Fleet docs explaining the transition to env variables for users who customised their fleet config. 

Also I think when we use env variables the secret parts should be mounted from k8s secrets instead of putting them directly in the spec.

PR #4890, opened 2021-09-28
Use arbitrary user ID other than 1000 in E2E tests

### LINKED ISSUE #3664: E2E Tests: do not run containers as UID 1000
Right now we ensure that Pods can be run as non root by forcing the uid of the containers to 1000. I selected this value in the past because it was the only UID that could be used by the APM Server before 7.9.0.

https://github.com/elastic/cloud-on-k8s/blob/09be60c9731520cf329d390877e73a0a2bd96221/test/e2e/test/default.go#L19-L23

But most of the container images are designed to run as user 1000, hence using that value doesn't validate that the containers are able to run as a random uid. We should use another value.

It would help to detect situations like the one described in #3656

thbkrkr:
Resolves #3664.

### DIFF
--- a/test/e2e/test/default.go
+++ b/test/e2e/test/default.go
@@ -13,35 +13,18 @@ import (
 // Values should be inherited and checked against a PSP, but we provide some
 // default values if pods are started outside E2E tests, by a developer for example.
 func DefaultSecurityContext() *corev1.PodSecurityContext {
-	dscc := &corev1.PodSecurityContext{
-		RunAsNonRoot: BoolPtr(true),
-	}
-
-	if !Ctx().OcpCluster {
-		defaultUserID := int64(12345) // arbitrary user ID
-		// Stack images expected to run with the user ID 1000 or 0 before 7.9.0 for APM and Beats (https://github.com/elastic/beats/issues/18871)
-		// and 7.11.0 for Enterprise Search (https://github.com/elastic/enterprise-search-team/issues/285)
-		stackVersion := version.MustParse(Ctx().ElasticStackVersion)
-		if stackVersion.LT(version.MustParse("7.11.0")) {
-			defaultUserID = int64(1000)
+	// OpenShift sets the security context automatically
+	if Ctx().OcpCluster {
+		return &corev1.PodSecurityContext{
+			RunAsNonRoot: BoolPtr(true),
 		}
-		dscc.RunAsUser = &defaultUserID
-		dscc.FSGroup = &defaultUserID
 	}
 
-	return dscc
-}
-
-// It's currently not possible to run APM using OpenShift's
-// restricted SCC. Therefore, we are forcing the required UID
-// and fsGroup for APM's security context. A dedicated ServiceAccount
-// with special permissions is created by APM test's builder
-// so that this can work.
-func APMDefaultSecurityContext() *corev1.PodSecurityContext {
-	defaultUserID := int64(12345) // arbitrary user ID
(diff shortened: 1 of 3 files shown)


david-kow (approved):
LGTM

test/e2e/test/default.go
```diff
@@ -17,7 +18,12 @@ func DefaultSecurityContext() *corev1.PodSecurityContext {
 	}
 
 	if !Ctx().OcpCluster {
-		defaultUserID := int64(1000)
+		defaultUserID := int64(123456) // arbitrary user ID
+		// before 7.9.0 Stack images expected to run with the user ID 1000 or 0 (https://github.com/elastic/beats/issues/18871)
```
barkbay, line 22: Regarding EnterpriseSearch, I think https://github.com/elastic/ent-search/pull/2439 is only available starting 7.11

thbkrkr:
run full pr build

thbkrkr:
`watches.TestDynamicEnqueueRequest` :red_circle: => Jenkins test this please

thbkrkr:
`java.lang.IllegalStateException: JENKINS-37121` :red_circle: => Jenkins test this please

thbkrkr:
run full pr build

test/e2e/test/default.go
```diff
@@ -31,8 +38,12 @@ func DefaultSecurityContext() *corev1.PodSecurityContext {
 // with special permissions is created by APM test's builder
 // so that this can work.
 func APMDefaultSecurityContext() *corev1.PodSecurityContext {
```
barkbay, line 40: nit: Maybe we should update the godoc for this function as it is no longer accurate.

test/e2e/test/default.go
```diff
@@ -31,8 +38,12 @@ func DefaultSecurityContext() *corev1.PodSecurityContext {
...
-	defaultUserID := int64(1000)
-
+	defaultUserID := int64(12345) // arbitrary user ID
+	// APM image expected to run with the user ID 1000 or 0 before 7.9.0 (https://github.com/elastic/beats/issues/18871)
+	stackVersion := version.MustParse(Ctx().ElasticStackVersion)
+	if stackVersion.LT(version.MustParse("7.9.0")) {
```
barkbay, line 44: It works, but for versions > 7.9.0 we should ideally let OpenShift assigns the "UID" as it is done for the other Stack components with `DefaultSecurityContext()`:

```golang
	if !Ctx().OcpCluster {
		// if not in an OCP cluster manage the UID
	}
        // if in an OCP cluster, OpenShift sets the security context automatically
```

It would also mean that we can use the `default` `ServiceAccount` for the APM Server, instead of the one used to escape from the restricted SCC today: https://github.com/elastic/cloud-on-k8s/blob/f52ca4700f56238c56186f0fc751ada563cf6dda/test/e2e/test/apmserver/steps_creation.go#L32-L35

It might be a good opportunity to update this TODO.

That being, as I said, it works and you should have the pretty same result with your proposal.

barkbay (approved):
lgtm, I just left a proposal about using the `restricted` SCC on OpenShift for APM Server > 7.9.0, and therefore let OpenShift sets the `SecurityContext` and the UID in that case.

test/e2e/test/default.go
thbkrkr: Yes, let's clean up.

thbkrkr:
Jenkins test this please

thbkrkr:
Jenkins test this please

thbkrkr:
run full pr build

thbkrkr:
run full pr build

PR #4893, opened 2021-09-28
Decrease default Elasticsearch client timeout

### LINKED ISSUE #3496: Short-circuit reconciliation if ES is down
If Elasticsearch is not responding in time to API requests we keep trying to make requests to it from the Elasticsearch controller effectively blocking reconciliation of other clusters in that go routine for 90 seconds. 

Screenshot 2020-07-21 at 09 12 35

I believe @anyasabo also observed the same thing. We could short-circuit the reconciliation loop in these cases instead of continuing. So instead or 90 seconds we would give up after 30 seconds.

thbkrkr:
To short-circuit the reconciliation loop if Elasticsearch is down and give back to reconcile other clusters.

Resolves #3496.

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -179,7 +179,7 @@ func Command() *cobra.Command {
 	)
 	cmd.Flags().Duration(
 		operator.ElasticsearchClientTimeout,
-		3*time.Minute,
+		30*time.Second,
 		"Default timeout for requests made by the Elasticsearch client.",
 	)
 	cmd.Flags().Bool(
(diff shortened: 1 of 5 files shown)


sebgl:
I'm a bit concerned this may not play well with existing clusters that are slow to respond (for example, because they are overloaded with shards), and hence could be a breaking change for existing users. Curious about what other folks think.

thbkrkr:
This could have an impact because it can prevent a cluster from being reconciled, thus blocking any new updates for it, but this will not break or disrupt it. 

Almost all ES API have a default timeout set to 30s so we never reach this timeout.

I realize this PR is irrelevant for #3496, sorry for the noise :facepalm:.

PR #4905, opened 2021-10-01
E2E: Azure disk exception

pebrc:
Until https://github.com/Azure/AKS/issues/1477#issuecomment-901689783 is resolved exclude azure-disk explicitly. This is not needed for AKS where we have an `azurefile` based storageclass but the Tanzu e2e tests run on Azure directly and the default storage class there is provisioned by `azure-disk`.

Open to alternative solutions as well. I know this is not great.


sebgl (approved):
LGTM
Let's make sure we track the TODO in a separate issue?

pebrc:
Jenkins test this please

> JENKINS-37121: something already locked /var/lib/jenkins/workspace/cloud-on-k8s-pr

Issue #4692, opened 2021-07-27
TestDynamicEnqueueRequest is flaky


barkbay:
[TestDynamicEnqueueRequest](https://github.com/elastic/cloud-on-k8s/blob/f52ca4700f56238c56186f0fc751ada563cf6dda/pkg/controller/common/watches/handler_integration_test.go#L36) is flaky. 

It has been [first noticed](https://github.com/elastic/cloud-on-k8s/pull/4587#issuecomment-866105188) by @thbkrkr while we were upgrading controller-runtime to v0.9.1. It also failed a couple of times for these PRs:

* #4689 
* #4681
* #4667 
* #4618 ([here](https://github.com/elastic/cloud-on-k8s/pull/4618#issuecomment-878322900))

PR #4922, opened 2021-10-04
Refactor Elasticsearch HTTP error handling

### LINKED ISSUE #4918: Elasticsearch client should enrich errors returned with context information about the request
It is currently very hard to find out what specific request to Elasticsearch caused an error. Especially if there is no additional debug logging around that request and the error bubbles up all the way to the top most error handler in controller-runtime. 

We should at least decorate each error create in the Elasticsearch client code with the original request.

barkbay:
This PR fixes #4918.

It also removes `response *http.Response` from the `APIError` object, which is currently preventing `Error()` from being called multiple times:

```golang
// Error() implements the error interface.
func (e *APIError) Error() string {
	defer e.response.Body.Close() # body is closed the first
	reason := "unknown"
	// Elasticsearch has a detailed error message in the response body
	var errMsg ErrorResponse
	err := json.NewDecoder(e.response.Body).Decode(&errMsg) // subsequent calls fail
	if err == nil {
		reason = errMsg.Error.Reason
	}
	return fmt.Sprintf("%s: %s", e.response.Status, reason)
}
```

This is for example the case when an `APIError` is aggregated in [`zapcode.encodeError`](https://github.com/uber-go/zap/blob/42e70dd4e9258b142fca9fc01c639fb3c8e8f664/zapcore/error.go#L63) :

```golang
func encodeError(key string, err error, enc ObjectEncoder) (retErr error) {

        basic := err.Error() // Error() is called a first time

        switch e := err.(type) {
        case errorGroup:
                return enc.AddArray(key+"Causes", errArray(e.Errors())) // Error() is called a second time
```

This is the reason why the error message may contain `unknown` in the error cause:

```json
{
    "log.level": "error",
    "@timestamp": "2021-09-23T14:25:54.121Z",
    "log.logger": "manager.eck-operator.controller.elasticsearch-controller",
    "message": "Reconciler error",
    "service.version": "1.7.1+26876766",
    "service.type": "eck",
    "ecs.version": "1.4.0",
    "name": "my-cluster",
    "namespace": "elastic-system",
    "error": "400 Bad Request: closed",
    "errorCauses": [
        {
            "error": "400 Bad Request: unknown"
        }
    ],
    "error.stack_trace": "sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.9.2/pkg/internal/controller/controller.go:253\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.9.2/pkg/internal/controller/controller.go:214"
}
```

### DIFF
--- a/pkg/controller/elasticsearch/client/base.go
+++ b/pkg/controller/elasticsearch/client/base.go
@@ -13,9 +13,12 @@ import (
 	"net/http"
 
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common/version"
+	ulog "github.com/elastic/cloud-on-k8s/pkg/utils/log"
 	"github.com/elastic/cloud-on-k8s/pkg/utils/stringsutil"
 	"github.com/hashicorp/go-multierror"
 )
+
+var log = ulog.Log.WithName("elasticsearch-client")
 
 type baseClient struct {
 	User     BasicAuth
@@ -66,10 +69,15 @@ func (c *baseClient) doRequest(context context.Context, request *http.Request) (
 
 	response, err := c.HTTP.Do(withContext)
 	if err != nil {
-		return response, err
+		return response, newDecoratedHTTPError(request, err)
 	}
-	err = checkError(response)
-	return response, err
+
+	// Check HTTP code in Elasticsearch response.
+	if response.StatusCode < 200 || response.StatusCode >= 300 {
+		return response, newDecoratedHTTPError(request, newAPIError(response))
+	}
+
+	return response, nil
 }
 
 func (c *baseClient) get(ctx context.Context, pathWithQuery string, out interface{}) error {
@@ -156,13 +164,4 @@ func versioned(b *baseClient, v version.Version) Client {
 	default:
 		return &v6
 	}
-}
-
-func checkError(response *http.Response) error {
-	if response.StatusCode < 200 || response.StatusCode >= 300 {
-		return &APIError{
-			response: response,
-		}
-	}
-	return nil
 }
--- a/pkg/controller/elasticsearch/client/error.go
+++ b/pkg/controller/elasticsearch/client/error.go
@@ -1,0 +1,92 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License 2.0;
+// you may not use this file except in compliance with the Elastic License 2.0.
+
+package client
+
+import (
+	"bytes"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+)
+
+// APIError is a non 2xx response from the Elasticsearch API
+type APIError struct {
+	Status        string
+	StatusCode    int
+	ErrorResponse ErrorResponse
+}
+
+// newAPIError converts an HTTP response into an API error, attempting to parse the body to include the details about the error.
+func newAPIError(response *http.Response) error {
+	defer response.Body.Close()
+	apiError := &APIError{
+		Status:     response.Status,
+		StatusCode: response.StatusCode,
+	}
+	// We may need to read the body multiple times, read the full body and store it as an array of bytes.
+	body, err := ioutil.ReadAll(response.Body)
+	if err != nil {
+		// We were not able to read the body, log this I/O error and return the API error with the status.
+		log.Error(err, "Cannot read error body")
+		return apiError
+	}
+	// Reset the response body to the original unread state. It allows a caller to read again the body if necessary,
+	// for example in the case of a 408.
+	response.Body = ioutil.NopCloser(bytes.NewBuffer(body))
+
+	// Parse the body to get the details about the API error, as they are stored by Elasticsearch.
+	var errorResponse ErrorResponse
+	if err := json.Unmarshal(body, &errorResponse); err != nil {
+		// Only log at the debug level since it is expected to not be able to parse all types of errors.
+		// Some errors, like 408 on /_cluster/health may return a different body structure.
+		log.V(1).Error(err, "Cannot parse error body")
+		return apiError
+	}
+	apiError.ErrorResponse = errorResponse
+	return apiError
+}
+
+// Error() implements the error interface.
+func (a *APIError) Error() string {
+	return fmt.Sprintf("%s: %+v", a.Status, a.ErrorResponse)
+}
+
+func newDecoratedHTTPError(request *http.Request, err error) error {
+	if request == nil {
+		return err
+	}
+	return fmt.Errorf(`elasticsearch client failed for %s: %w`, request.URL.Redacted(), err)
+}
+
+// IsForbidden checks whether the error was an HTTP 403 error.
+func IsForbidden(err error) bool {
+	return isHTTPError(err, http.StatusForbidden)
+}
+
+// IsNotFound checks whether the error was an HTTP 404 error.
+func IsNotFound(err error) bool {
+	return isHTTPError(err, http.StatusNotFound)
+}
+
+// IsTimeout checks whether the error was an HTTP 408 error
+func IsTimeout(err error) bool {
+	return isHTTPError(err, http.StatusRequestTimeout)
+}
+
+// IsConflict checks whether the error was an HTTP 409 error.
+func IsConflict(err error) bool {
+	return isHTTPError(err, http.StatusConflict)
+}
+
+func isHTTPError(err error, statusCode int) bool {
+	apiErr := new(APIError)
+	if errors.As(err, &apiErr) {
+		return apiErr.StatusCode == statusCode
+	}
+
+	return false
+}
--- a/pkg/controller/elasticsearch/client/client.go
+++ b/pkg/controller/elasticsearch/client/client.go
@@ -7,8 +7,6 @@ package client
 import (
 	"context"
 	"crypto/x509"
-	"encoding/json"
-	"errors"
 	"fmt"
 	"math"
 	"net/http"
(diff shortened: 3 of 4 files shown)


barkbay:
Jenkins test this please

https://github.com/elastic/cloud-on-k8s/issues/4692

barkbay:
Jenkins test this please

 java.lang.IllegalStateException: JENKINS-37121: something already locked /var/lib/jenkins/workspace/cloud-on-k8s-pr

pkg/controller/elasticsearch/client/error.go
```diff
@@ -0,0 +1,92 @@
...
+		// Only log at the debug level since it is expected to not be able to parse all types of errors.
+		// Some errors, like 408 on /_cluster/health may return a different body structure.
+		log.V(1).Error(err, "Cannot parse error body")
+		return apiError
+	}
+	apiError.ErrorResponse = errorResponse
+	return apiError
+}
+
+// Error() implements the error interface.
```
sebgl, line 53:
```suggestion
// APIError implements the error interface.
```

pkg/controller/elasticsearch/client/error.go
```diff
@@ -0,0 +1,92 @@
...
+	defer response.Body.Close()
+	apiError := &APIError{
+		Status:     response.Status,
+		StatusCode: response.StatusCode,
+	}
+	// We may need to read the body multiple times, read the full body and store it as an array of bytes.
+	body, err := ioutil.ReadAll(response.Body)
+	if err != nil {
+		// We were not able to read the body, log this I/O error and return the API error with the status.
+		log.Error(err, "Cannot read error body")
```
sebgl, line 34:
```suggestion
		log.Error(err, "Cannot read Elasticsearch error response body")
```

pkg/controller/elasticsearch/client/error.go
```diff
@@ -0,0 +1,92 @@
...
+
+// newAPIError converts an HTTP response into an API error, attempting to parse the body to include the details about the error.
+func newAPIError(response *http.Response) error {
+	defer response.Body.Close()
+	apiError := &APIError{
+		Status:     response.Status,
+		StatusCode: response.StatusCode,
+	}
+	// We may need to read the body multiple times, read the full body and store it as an array of bytes.
+	body, err := ioutil.ReadAll(response.Body)
```
sebgl, line 31: nit: I would put the `defer response.Body.Close() ` statement here (right after we decide to use it, don't forget to close it).

pkg/controller/elasticsearch/client/error.go
```diff
@@ -0,0 +1,92 @@
...
+	if err != nil {
+		// We were not able to read the body, log this I/O error and return the API error with the status.
+		log.Error(err, "Cannot read error body")
+		return apiError
+	}
+	// Reset the response body to the original unread state. It allows a caller to read again the body if necessary,
+	// for example in the case of a 408.
+	response.Body = ioutil.NopCloser(bytes.NewBuffer(body))
```
sebgl, line 39: Do we read that body in case of 408? Or just check the status code?
The code flow here gets a bit tricky. I'm wondering if it may be simpler to always return either `response, nil` or `nil, err` for any http request we make. So that function that creates the `APIError` would read the response body, and then we don't care about that response body anymore, and there's no way to retrieve it (it is still stored in the error object if we need it).
Checking whether the api call we made returned a 408 would be a call to a helper function like `client.IsTimeout(err)`.

pkg/controller/elasticsearch/client/error.go
```diff
@@ -0,0 +1,92 @@
...
+
+	// Parse the body to get the details about the API error, as they are stored by Elasticsearch.
+	var errorResponse ErrorResponse
+	if err := json.Unmarshal(body, &errorResponse); err != nil {
+		// Only log at the debug level since it is expected to not be able to parse all types of errors.
+		// Some errors, like 408 on /_cluster/health may return a different body structure.
```
sebgl, line 45: If we know and expect 408 will return a different body, should we just not parse the error body in case of 408?

pkg/controller/elasticsearch/client/error.go
```diff
@@ -0,0 +1,92 @@
...
+		log.V(1).Error(err, "Cannot parse error body")
```
sebgl, line 46:
```suggestion
		log.V(1).Error(err, "Cannot parse Elasticsearch error response body")
```

(to give a bit more context)

pkg/controller/elasticsearch/client/base.go
```diff
@@ -66,10 +69,15 @@ func (c *baseClient) doRequest(context context.Context, request *http.Request) (
 
 	response, err := c.HTTP.Do(withContext)
 	if err != nil {
-		return response, err
+		return response, newDecoratedHTTPError(request, err)
+	}
+
+	// Check HTTP code in Elasticsearch response.
+	if response.StatusCode < 200 || response.StatusCode >= 300 {
+		return response, newDecoratedHTTPError(request, newAPIError(response))
```
sebgl, line 77: Considering this is the only usage of APIError, and we wrap it in a `newDecoratedHTTPError`, would it be better to add the request `path` to the `APIError` struct?
So the `path` belongs to a structured object rather than a raw string.

pkg/controller/elasticsearch/client/base.go
```diff
@@ -66,10 +69,15 @@ func (c *baseClient) doRequest(context context.Context, request *http.Request) (
 
 	response, err := c.HTTP.Do(withContext)
 	if err != nil {
-		return response, err
+		return response, newDecoratedHTTPError(request, err)
+	}
```
sebgl, line 73: unrelated to that PR, but since we're at it: I'm wondering if we should debug log all the requests we make to Elasticsearch here (with their path, not their body).

pkg/controller/elasticsearch/client/error.go
barkbay: It's tricky because we don't know how to parse the body in case of a "custom" error here.
For now it's the responsibility of [`baseClient.request(..,responseObj interface{},skipErrFunc func(error) bool)`](https://github.com/elastic/cloud-on-k8s/blob/dcc5a90ffc702053953b229936fe8787a54ccf87/pkg/controller/elasticsearch/client/base.go#L96) to decide:
1. If an API error can be skipped (using `skipErrFunc`) and should not prevent the body to be parsed. In the case of a `408`, the 2 conditions are:
   * path is `/_cluster/health`
   * code is 408
2. Unmarshall the response using the struct provided by the [caller](https://github.com/elastic/cloud-on-k8s/blob/dcc5a90ffc702053953b229936fe8787a54ccf87/pkg/controller/elasticsearch/client/v6.go#L79) (`responseObj interface{}`)

pkg/controller/elasticsearch/client/error.go
barkbay: Maybe I'm missing something, the response body is always read by `body, err := ioutil.ReadAll(response.Body)`, there is no condition. This is why the `defer response.Body.Close()` is right before.

pkg/controller/elasticsearch/client/error.go
barkbay: I _think_ the purpose of this comment was to mention that `Error()` allows `APIError` to implement `error`. I would rather fix it as follow to respect godoc convention:

```suggestion
// Error implements the error interface.
```

pkg/controller/elasticsearch/client/error.go
barkbay: > Do we read that body in case of 408? Or just check the status code?

In case of a 408, and if the path is `/_cluster/health`, then the error is later skipped, and `baseClient.request(..)` attempts to read again the body from the response.

I think I wanted to be conservative with this PR, keeping the responsibility of reading the "common" errors here, while delegating the more specific ones to the callers. This is the current flow.

pkg/controller/elasticsearch/client/error.go
sebgl: ++ to what you said :)

pkg/controller/elasticsearch/client/error.go
sebgl: I think by convention people tend to put the `defer` statement immediately after the line that requires it ([effective go example](https://golang.org/doc/effective_go#defer), [uber go style example](https://github.com/uber-go/guide/blob/master/style.md#defer-to-clean-up)), but I really don't mind much.

barkbay: I think the situation is a bit different here since the stream is already opened. The purpose is to be sure that it is effectively closed when we exit the function and no one else will read from it later.

pkg/controller/elasticsearch/client/base.go
barkbay: 👍 , I'm just tempted to create a dedicated issue. It requires a small change to the current code in order to have the name of the Elasticsearch resource in the log message, which sounds like a "good first issue" for someone who want to familiarize itself with the client code.

pkg/controller/elasticsearch/client/error.go
sebgl: 🤔 I tend to feel like this is more complicated than it should be, but I'm fine with keeping that PR small and focused on the problem it solves.

pkg/controller/elasticsearch/client/error.go
sebgl: 🤔 I tend to feel like this is more complicated than it should be, but I'm fine with keeping that PR small and focused on the problem it solves.

sebgl (approved):
LGTM

barkbay:
@thbkrkr I removed `FakeAPIError` from `client.go` (see [here](https://github.com/elastic/cloud-on-k8s/blob/9bb8068e94ed8ddc96189ba9e1e9fecc2506bf65/pkg/controller/elasticsearch/client/client.go#L146-L152)), it was not in a unit test file and also I don't think it was really required.
Could you check I did not break anything ? Thanks !

PR #4924, opened 2021-10-05
Skip remote clusters reconciliation if remote clusters are not defined

thbkrkr:
This is a small improvement to not start to reconcile remote clusters or update remote clusters settings in ES when remote clusters are not defined in the ES specification.

### DIFF
--- a/pkg/apis/elasticsearch/v1/elasticsearch_types.go
+++ b/pkg/apis/elasticsearch/v1/elasticsearch_types.go
@@ -175,6 +175,11 @@ func (r RemoteCluster) ConfigHash() string {
 	return hash.HashObject(r)
 }
 
+func (es Elasticsearch) IsRemoteClustersDefined() bool {
+	return len(es.Spec.RemoteClusters) > 0 &&
+		es.Spec.RemoteClusters[0].Name != "" && es.Spec.RemoteClusters[0].ElasticsearchRef.IsDefined()
+}
+
 // NodeCount returns the total number of nodes of the Elasticsearch cluster
 func (es ElasticsearchSpec) NodeCount() int32 {
 	count := int32(0)
--- a/pkg/controller/elasticsearch/remotecluster/elasticsearch.go
+++ b/pkg/controller/elasticsearch/remotecluster/elasticsearch.go
@@ -37,6 +37,10 @@ func UpdateSettings(
 	licenseChecker license.Checker,
 	es esv1.Elasticsearch,
 ) (bool, error) {
+	if !es.IsRemoteClustersDefined() {
+		return false, nil
+	}
+
 	span, _ := apm.StartSpan(ctx, "update_remote_clusters", tracing.SpanTypeApp)
 	defer span.End()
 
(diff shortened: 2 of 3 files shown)


pkg/apis/elasticsearch/v1/elasticsearch_types.go
```diff
@@ -175,6 +175,11 @@ func (r RemoteCluster) ConfigHash() string {
 	return hash.HashObject(r)
 }
 
+func (es Elasticsearch) IsRemoteClustersDefined() bool {
+	return len(es.Spec.RemoteClusters) > 0 &&
+		es.Spec.RemoteClusters[0].Name != "" && es.Spec.RemoteClusters[0].ElasticsearchRef.IsDefined()
```
david-kow, line 180: Should we iterate over all remote clusters in case the first one is not correctly defined?

pkg/controller/elasticsearch/driver/driver.go
```diff
@@ -213,7 +213,6 @@ func (d *defaultDriver) Reconcile(ctx context.Context) *reconciler.Results {
 			msg := "Could not update remote clusters in Elasticsearch settings"
 			d.ReconcileState.AddEvent(corev1.EventTypeWarning, events.EventReasonUnexpected, msg)
 			log.Error(err, msg, "namespace", d.ES.Namespace, "es_name", d.ES.Name)
-			results.WithResult(defaultRequeue)
```
david-kow, line 216: Doesn't this change the behaviour for every other error case? Eg. es client or k8s client temporary errors?

david-kow: I missed the if below. So actually, the difference in behaviour would only be there if we return `false` requeue and no error - I believe this can happen if `remoteClustersToDelete` is `0` and `updateSettings(esClient, remoteClustersToApply)` fails in `updateSettingsInternal`.

thbkrkr:
run full pr build

thbkrkr:
I forgot the case where the remote clusters were enabled and then disabled, so we need to clean up some k8s secrets.

thbkrkr:
When we update the settings for remote clusters in the elasticsearch controller, we take care to clean up the annotation.
In the remoteca controller, it is more delicate because we clean the secrets containing CA and we have to retrieve ES clusters for that. This makes the optimization only relevant for the former. I'll revert the change for the latter.

pkg/controller/elasticsearch/driver/driver.go
thbkrkr: The goal is to not change the current behaviour and just avoid to requeue twice. I also missed the edge case. It should be good if we also always requeue on error: 7c319fb.

pkg/apis/elasticsearch/v1/elasticsearch_types.go
thbkrkr: Yes, but I ended up reuse `getRemoteClustersInSpec()`.

barkbay:
👀

PR #4928, opened 2021-10-07
Do not stop reconciliation on invalid controller-version annotation

### LINKED ISSUE #4898: Reconcile resources with no controller-version annotation
In https://github.com/elastic/cloud-on-k8s/blob/20f522ff9f5fa04f8ccb22ce19edaa6d044a1aa6/pkg/controller/common/annotation/controller_version.go, we inspect the `common.k8s.elastic.co/controller-version` annotation on the custom resources we manage (Elasticsearch, Kibana, etc.).

This annotation allows us to track which version of the operator last reconciled the resource. It's intended to provide us a way to not reconcile existing resources if they were last updated by an incompatible operator.
So far we've only used that mechanism for the transition from <1.0 ECK versions to >=1.0 ECK versions, because we made significant changes to the orchestration mechanism (Pods -> StatefulSets) that were not worth the backward-compatibility trouble.

We've seen a few cases from time to time were users report ECK suddenly stopped reconciling their existing cluster, even though they were not running ECK <1.0.

This will happen if _something_ removes the controller-version annotation, in which case our implementation considers the resource was created with ECK <1.0. Chances are users did not explicitly want to remove that annotation, but rather, they updated the Elasticsearch resource through the update api endpoint, or through `kubectl replace`, rather than relying on the merge+patch mechanism of `kubectl apply`.

Given that we're far away from ECK <1.0 now, I think it would be reasonable to only log a warning if the annotation isn't set, but move on with the reconciliation. I think we've reached a point where that annotation does more harm than good.

I'll create a separate issue to discuss the risks of that annotation mechanism in general, let's focus that one on the controller-version problem only.

pebrc:
Fixes #4898 

This completely removes the code the managed the controller-version annotation. It leaves however existing annotations intact. The rationale behind that decision is that unsetting this annotation introduces technical debt into the code base while an additional annotation on existing resources should not hurt.


thbkrkr (approved):
LGTM

barkbay (approved):
LGTM

PR #4931, opened 2021-10-07
Fix 8.0.0 related issues in E2E

david-kow:
This PR fixes a number of issues that prevent E2E tests from passing on `8.0.0-alpha2`:

- ignore all suffixes (not only `-SNAPSHOT`) when checking APM Server version
- don't provide `ent_search.auth.*` settings in Enterprise Search config for `8.0.0+`
- use `monitoring.ui.container.elasticsearch.enabled` instead of `xpack.monitoring.ui.container.elasticsearch.enabled` in Kibana config for `7.16.0+`
- use `node.roles` instead of `node.data: true` etc. in Elasticsearch config for`7.9.0+` E2E tests and manifests
- set `kibana.host` setting in Enterprise Search config for `8.0.0`+ E2E tests
- inspect `metricbeat-*` instead of `.monitoring-kibana-*` index for checking Stack Monitoring data in Elasticsearch and Kibana
- check for `level` == `available` instead of `state` == `green` field when checking Kibana health
- check `/internal/workplace_search/overview` instead of `/api/enterprise_search/config_data` API when checking Enterprise Search UI availability in Kibana 
-  Temporarily disable `TestVolumeMultiDataPath` test for 8.x as the feature deprecation revert is still in progress

### DIFF
--- a/pkg/controller/common/version/version.go
+++ b/pkg/controller/common/version/version.go
@@ -84,10 +84,10 @@ func From(major, minor, patch int) Version {
 	return Version{Major: uint64(major), Minor: uint64(minor), Patch: uint64(patch)}
 }
 
-// MinFrom creates a new version from the given major, minor, patch numbers with lowest PreRelease version, ie.
+// MinFor creates a new version for the given major, minor, patch numbers with lowest PreRelease version, ie.
 // the returned Version is the lowest possible version with those major, minor and patch numbers.
 // See https://semver.org/#spec-item-11.
-func MinFrom(major, minor, patch uint64) Version {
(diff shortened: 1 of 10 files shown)


david-kow:
jenkins test this please

david-kow:
Tests look good, the only failure (below) is caused by using `nodes.roles` notion in `pre-7.9.0` ES versions.

 Screenshot 2021-10-08 at 23 21 38

This is fixed in https://github.com/elastic/cloud-on-k8s/pull/4931/commits/455e60ce695e67738d45206920fbb823db612eda.

PR #4945, opened 2021-10-12
Update golangci-lint/goimports configuration

naemono:
Update goimports settings to consider github.com/elastic/cloud-on-k8s a local prefix in imports section.

This standardizes having

1. stdlib imports at top of import section
2. 3rd party imports as the next section
3. local imports as the final section


pebrc:
If we go for this new three section import block we should also update the contributing guideline accordingly.

naemono:
@david-kow @pebrc any thoughts on the above question?  Should we just leave as-is, and merge?  lmk, thanks.

PR #4958, opened 2021-10-14
Add logging for call to ES

### LINKED ISSUE #4935: Elasticsearch HTTP client: log requests at debug level
As mentioned in this [comment](https://github.com/elastic/cloud-on-k8s/pull/4922#discussion_r722087762) we should log (at the debug level) all the requests to Elasticsearch.
We should take care to not log the body or any header as they may contain some credentials. The debug log should also contain the Elasticsearch cluster namespaced name as for the other logs (`es_name`).

koneko096:
Fixes #4935

Basically adding a debug log on each client request with only the URL path.


barkbay:
@koneko096  Thanks for your contribution !

Issue #4953, opened 2021-10-14
Add 7.16.0-SNAPSHOT to 8.0.0-SNAPSHOT E2E tests


david-kow:
We should make sure that `7.16.0-SNAPSHOT` to `8.0.0-SNAPSHOT` upgrades are working as expected, similar to tests we do between `6.8.x` and `7.x`.

PR #4961, opened 2021-10-18
Add 7.16.x to 8.x E2E tests

david-kow:
Fixes https://github.com/elastic/cloud-on-k8s/issues/4953.

### DIFF
--- a/test/e2e/ems/ems_test.go
+++ b/test/e2e/ems/ems_test.go
@@ -98,6 +98,8 @@ func TestElasticMapsServerVersionUpgradeToLatest7x(t *testing.T) {
 }
 
 func TestElasticMapsServerVersionUpgradeToLatest8x(t *testing.T) {
+	// todo skipping until https://github.com/elastic/cloud-on-k8s/issues/5251 is fixed.
+	t.SkipNow()
 	srcVersion := test.Ctx().ElasticStackVersion
 	dstVersion := test.LatestVersion8x
 
@@ -105,8 +107,8 @@ func TestElasticMapsServerVersionUpgradeToLatest8x(t *testing.T) {
 
 	name := "test-ems-version-upgrade-8x"
 	es := elasticsearch.NewBuilder(name).
-		WithESMasterDataNodes(1, elasticsearch.DefaultResources).
-		WithVersion(dstVersion)
+		WithESMasterDataNodes(3, elasticsearch.DefaultResources).
+		WithVersion(srcVersion)
 
 	ems := maps.NewBuilder(name).
 		WithElasticsearchRef(es.Ref()).
--- a/test/e2e/ems/ems_test.go
+++ b/test/e2e/ems/ems_test.go
@@ -96,3 +96,28 @@ func TestElasticMapsServerVersionUpgradeToLatest7x(t *testing.T) {
 
 	test.RunMutations(t, []test.Builder{esWithLicense, ems}, []test.Builder{esWithLicense, emsUpgraded})
 }
+
+func TestElasticMapsServerVersionUpgradeToLatest8x(t *testing.T) {
+	srcVersion := test.Ctx().ElasticStackVersion
+	dstVersion := test.LatestVersion8x
+
+	test.SkipInvalidUpgrade(t, srcVersion, dstVersion)
+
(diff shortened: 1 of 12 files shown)


david-kow:
run full pr build

david-kow:
run full pr build

pebrc (dismissed):
LGTM.

 I wonder though if we should parameterise the tests instead of more or less copying the existing ones  especially when the only difference between the 7x and 8x version of the upgrade test is the version numbers, but it does not have to be in this PR. It is more important to start testing these upgrades.

pebrc:
So I think the problem we have with these upgrade tests in general is that they won't work for any tests that requires a valid license. That means we need to make sure that tests that require licensed functionality do not run when we transition from a production build to a developed snapshot build as it is the case currently for all 8.x builds. That is also the underlying problem in https://github.com/elastic/cloud-on-k8s/issues/5251

barkbay:
I'm wondering if we could use [https://artifacts-api.elastic.co/v1/versions/](https://artifacts-api.elastic.co/v1/versions/) to automatically detect the latest versions or release candidates, and thus still be able to test the upgrade of components that require a valid license.

pebrc:
> thus still be able to test the upgrade of components that require a valid license.

You mean building a development build only test pipeline that tests upgrades? I think that might be a good idea. 

There is something odd here that confuses me about the test failure in #5251 that I want to understand first. I think this is fine in general to merge this given that problematic tests are skipped for now.

test/e2e/test/version.go
```diff
@@ -11,12 +11,14 @@ import (
...
+// Elastic Stack versions used in the E2E tests. These should be updated as new versions for each major are released.
 const (
-	// MinVersion68x minimum version for 6.8.x tested with the operator
-	MinVersion68x = "6.8.20"
-	// LatestVersion7x current latest version for 7.x
-	LatestVersion7x = "7.16.2" // version to synchronize with the latest release of the Elastic Stack
+	// LatestVersion6x is the current latest version for 6.x
+	LatestVersion6x = "6.8.20"
+	// LatestVersion7x is the current latest version for 7.x
+	LatestVersion7x = "7.17.0-SNAPSHOT"
```
pebrc, line 19: The `TestElasticMapsServerVersionUpgradeToLatest7x` test is failing because you changed the target version here to a snapshot build. So again the SNAPSHOT vs production build problem. I think you either have to change the test in `TestElasticMapsServerVersionUpgradeToLatest7x` to keep Elasticsearch on the production version as you have done already for the `TestElasticMapsServerVersionUpgradeToLatest8x` test. Or keep this variable on the latest production build.

test/e2e/ems/ems_test.go
```diff
@@ -96,3 +96,30 @@ func TestElasticMapsServerVersionUpgradeToLatest7x(t *testing.T) {
...
+	t.SkipNow()
+	srcVersion := test.Ctx().ElasticStackVersion
+	dstVersion := test.LatestVersion8x
+
+	test.SkipInvalidUpgrade(t, srcVersion, dstVersion)
+
+	name := "test-ems-version-upgrade-8x"
+	es := elasticsearch.NewBuilder(name).
+		WithESMasterDataNodes(3, elasticsearch.DefaultResources).
+		WithVersion(srcVersion)
```
pebrc, line 111: Given that you are keeping Elasticsearch on `srcVersion` here I don't think we need to skip this test.

test/e2e/test/version.go
david-kow: At least for now, `LatestVersion7x = "7.17.0-SNAPSHOT"` has to stay because `8.0.0-SNAPSHOT` already expects to be upgraded from `7.17.0` (which doesn't have a production build yet).

david-kow: Actually, the src version has to be the snapshot which doesn't mean that this variable needs to be the source. I'm making it back into latest production build and updating the comment.

david-kow:
run full pr build

david-kow:
run full pr build

pebrc (approved):
LGTM!

david-kow:
run full pr build

PR #4968, opened 2021-10-18
Allowing easy access to sub-sections of accessing Elastic Stack Services

naemono:
Finding the documentation for accessing Elasticsearch, and associated services outside of the cluster, in the public docs, is rather hard to find.  This will add direct links into the main doc, instead of just the generated TOC on the right hand side

Before
image

After (using preview in vscode)
image

Are there ways to generate what is seen on the main website to verify further locally?


naemono:
Thanks @david-kow for the info.  I've double checked both via the local server, and the public diff link, and all looks as intended.

docs/orchestrating-elastic-stack-applications/accessing-elastic-services.asciidoc
```diff
@@ -13,6 +13,13 @@ To access the Elastic Stack services, you need to:
 - Specify the IP of the service, if you want to access the service from outside the Kubernetes cluster.
 - Decide if you want to use the self-signed certificate generated by ECK, or configure your own certificate.
 
+Sub-Sections:
```
barkbay, line 16:
```suggestion
The following sections will guide you through this process:
```

barkbay (commented):
I think I'm usually fine with the autogenerated TOC, but we also have these sections repeated in other parts of our documentation (see [Run Beats on ECK](https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-beat.html) for example), so I'm 👍 overall.

I would just defer to @alaudazzi for the final ✅ 

As as side note this PR brings me to read again this part of our documentation.
I never realized that there was a bit of overlap between `Access Elastic Stack services` and `Access Elastic Stack services`:

* `Access Elastic Stack services > Security` is only about getting the `elastic user`, which is already described in the previous section `Secure the Elastic Stack > Users and roles`

* TLS certificates documentation seems to be split in both sections:
  * `Secure the Elastic Stack > Custom HTTP certificate`
  * `Access Elastic Stack services > TLS certificates`

* ` Access Elastic Stack services > Access the Elasticsearch endpoint` seems to be only relevant in the context of a quickstart given all the details provided in the previous sections.

I was wondering it is intended, or if it would be better to have only one section dealing with Security.

alaudazzi (approved):
LGTM

PR #4994, opened 2021-10-25
Also Release Images to DockerHub

naemono:
Also push to dockerhub when running ci-release target
Use secret/release/docker-hub path in vault to get dockerhub credentials
In docker-multiarch-build target, only build and tag the image
Use docker-push target to do pushing of the image to both elastic, and dockerhub registry
Update hack/docker.sh to support docker.io images

*Note* I'll be testing this, as it's CI changes, hence opening a Draft.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,10 +36,8 @@ endif
 IMG_SUFFIX ?= -$(subst _,,$(shell whoami))
 
 REGISTRY           ?= docker.elastic.co
-DOCKERHUB_REGISTRY = docker.io
 REGISTRY_NAMESPACE ?= eck-dev
 NAME               ?= eck-operator
-ORGANIZATION       ?= elastic
 SNAPSHOT           ?= true
 VERSION            ?= $(shell cat VERSION)
 TAG                ?= $(shell git rev-parse --short=8 --verify HEAD)
@@ -48,7 +46,7 @@ IMG_VERSION        ?= $(VERSION)-$(TAG)
 
 BASE_IMG                 := $(REGISTRY)/$(REGISTRY_NAMESPACE)/$(IMG_NAME)
 OPERATOR_IMAGE           ?= $(BASE_IMG):$(IMG_VERSION)
-OPERATOR_DOCKERHUB_IMAGE ?= $(DOCKERHUB_REGISTRY)/$(ORGANIZATION)/$(IMG_NAME):$(IMG_VERSION)
+OPERATOR_DOCKERHUB_IMAGE ?= docker.io/elastic/$(IMG_NAME):$(IMG_VERSION)
 
 print-operator-image:
 	@ echo $(OPERATOR_IMAGE)
@@ -244,7 +242,8 @@ build-operator-image:
 	|| $(MAKE) docker-build docker-push
 
 build-operator-multiarch-image:
-	@ docker buildx imagetools inspect $(OPERATOR_IMAGE) | grep -q 'linux/arm64' 2>&1 >/dev/null \
+	@ (docker buildx imagetools inspect $(OPERATOR_IMAGE) | grep -q 'linux/arm64' 2>&1 >/dev/null \
+	|| docker buildx imagetools inspect $(OPERATOR_DOCKERHUB_IMAGE) | grep -q 'linux/arm64' 2>&1 >/dev/null) \
 	&& echo "OK: image $(OPERATOR_IMAGE) already published" \
 	|| $(MAKE) docker-multiarch-build
 
--- a/Makefile
+++ b/Makefile
@@ -394,7 +394,8 @@ docker-multiarch-build: go-generate generate-config-file
 		--build-arg VERSION='$(VERSION)' \
 		--platform linux/amd64,linux/arm64 \
 		-t $(OPERATOR_IMAGE) \
-		-t $(OPERATOR_DOCKERHUB_IMAGE)
+		-t $(OPERATOR_DOCKERHUB_IMAGE) \
+		--load
 
 docker-build: go-generate generate-config-file 
 	DOCKER_BUILDKIT=1 docker build . \
--- a/Makefile
+++ b/Makefile
@@ -245,7 +245,7 @@ build-operator-multiarch-image:
 	@ hack/docker.sh -l -m $(OPERATOR_IMAGE)
 	@ hack/docker.sh -l -m $(OPERATOR_DOCKERHUB_IMAGE)
 	@ (docker buildx imagetools inspect $(OPERATOR_IMAGE) | grep -q 'linux/arm64' 2>&1 >/dev/null \
-	|| docker buildx imagetools inspect $(OPERATOR_DOCKERHUB_IMAGE) | grep -q 'linux/arm64' 2>&1 >/dev/null) \
+	&& docker buildx imagetools inspect $(OPERATOR_DOCKERHUB_IMAGE) | grep -q 'linux/arm64' 2>&1 >/dev/null) \
 	&& echo "OK: image $(OPERATOR_IMAGE) already published" \
 	|| $(MAKE) docker-multiarch-build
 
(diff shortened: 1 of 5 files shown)


naemono:
https://hub.docker.com/r/elastic/eck-operator/tags
Note the snapshot image tags that were released during testing.

Makefile
```diff
@@ -393,6 +401,17 @@ docker-multiarch-build: go-generate generate-config-file
...
+docker-multiarch-build-dockerhub:
+	@ hack/docker.sh -l -m $(OPERATOR_DOCKERHUB_IMAGE)
+	docker buildx build . \
+		--progress=plain \
+		--build-arg GO_LDFLAGS='$(GO_LDFLAGS)' \
+		--build-arg GO_TAGS='$(GO_TAGS)' \
+		--build-arg VERSION='$(VERSION)' \
+		--platform linux/amd64,linux/arm64 \
+		--push \
+		-t $(OPERATOR_DOCKERHUB_IMAGE)
```
thbkrkr, line 413: Why rebuild an image? Couldn't we just tag and push?
```suggestion
	docker tag $(OPERATOR_IMAGE) $(OPERATOR_DOCKERHUB_IMAGE)
	docker push $(OPERATOR_DOCKERHUB_IMAGE)
```

naemono: I initially attempted this, and ran into issues, but my approach was incorrect, and just telling buildx to tag 2x tags, and the `--push` flag makes it work without issues.  I've pushed this change.

hack/docker.sh
```diff
@@ -38,6 +39,11 @@ docker-login() {
             gcloud auth configure-docker --quiet 2> /dev/null
         ;;
 
+        docker.io/*)
```
barkbay, line 42: Not from your PR but should we not use a similar pattern with `docker.elastic.co` ? Otherwise any image with `/eck/` in its name matches the first condition, even if we actually want to push it to GCR or DockerHub.

I was thinking about something along those lines:

```bash
shopt -s extglob
...
    case "$image" in

        docker.elastic.co/@(eck|eck-ci|eck-snapshots)/*)
            echo "Authentication to ${registry}..."
            docker login -u "${DOCKER_LOGIN}" -p "${DOCKER_PASSWORD}" docker.elastic.co 2> /dev/null
        ;;

        *.gcr.io/*)
            echo "Authentication to ${registry}..."
            gcloud auth configure-docker --quiet 2> /dev/null
        ;;

        docker.io/*)
            echo "Authentication to ${registry}..."
            docker login -u "${DOCKERHUB_LOGIN}" -p "${DOCKERHUB_PASSWORD}" 2> /dev/null
        ;;

        *)
            if ! grep -q "$registry" ~/.docker/config.json; then
               echo "Please log in to $registry."
               exit 1
            fi
        ;;
    esac
```
_(not sure that `@(eck|eck-ci|eck-snapshots)` is really required)_

Makefile
```diff
@@ -384,14 +387,16 @@ switch-tanzu:
...
 	docker buildx build . \
 		--progress=plain \
 		--build-arg GO_LDFLAGS='$(GO_LDFLAGS)' \
 		--build-arg GO_TAGS='$(GO_TAGS)' \
 		--build-arg VERSION='$(VERSION)' \
 		--platform linux/amd64,linux/arm64 \
-		--push \
-		-t $(OPERATOR_IMAGE)
+		-t $(OPERATOR_IMAGE) \
+		-t $(OPERATOR_DOCKERHUB_IMAGE) \
```
barkbay, line 398: Should we also update `build-operator-multiarch-image` to check if `OPERATOR_DOCKERHUB_IMAGE` has been published ? Right now it is only checking for `OPERATOR_IMAGE`:

```Makefile
build-operator-multiarch-image:
	@ docker buildx imagetools inspect $(OPERATOR_IMAGE) | grep -q 'linux/arm64' 2>&1 >/dev/null \
	&& echo "OK: image $(OPERATOR_IMAGE) already published" \
	|| $(MAKE) docker-multiarch-build
```

hack/docker.sh
thbkrkr: +1 to use `docker.elastic.co/*)`.

Makefile
```diff
@@ -36,16 +36,19 @@ endif
...
 VERSION            ?= $(shell cat VERSION)
 TAG                ?= $(shell git rev-parse --short=8 --verify HEAD)
 IMG_NAME           ?= $(NAME)$(IMG_SUFFIX)
 IMG_VERSION        ?= $(VERSION)-$(TAG)
 
-BASE_IMG       := $(REGISTRY)/$(REGISTRY_NAMESPACE)/$(IMG_NAME)
-OPERATOR_IMAGE ?= $(BASE_IMG):$(IMG_VERSION)
+BASE_IMG                 := $(REGISTRY)/$(REGISTRY_NAMESPACE)/$(IMG_NAME)
+OPERATOR_IMAGE           ?= $(BASE_IMG):$(IMG_VERSION)
+OPERATOR_DOCKERHUB_IMAGE ?= $(DOCKERHUB_REGISTRY)/$(ORGANIZATION)/$(IMG_NAME):$(IMG_VERSION)
```
thbkrkr, line 51: Nit: Since we already have a lot of variables, I'd be tempted to add just the minimum.

```diff
-DOCKERHUB_REGISTRY = docker.io
 REGISTRY_NAMESPACE ?= eck-dev
 NAME               ?= eck-operator
-ORGANIZATION       ?= elastic
 SNAPSHOT           ?= true
 VERSION            ?= $(shell cat VERSION)
 TAG                ?= $(shell git rev-parse --short=8 --verify HEAD)
@@ -48,7 +46,7 @@ IMG_VERSION        ?= $(VERSION)-$(TAG)
 
 BASE_IMG                 := $(REGISTRY)/$(REGISTRY_NAMESPACE)/$(IMG_NAME)
 OPERATOR_IMAGE           ?= $(BASE_IMG):$(IMG_VERSION)
-OPERATOR_DOCKERHUB_IMAGE ?= $(DOCKERHUB_REGISTRY)/$(ORGANIZATION)/$(IMG_NAME):$(IMG_VERSION)
+OPERATOR_DOCKERHUB_IMAGE ?= docker.io/elastic/$(IMG_NAME):$(IMG_VERSION)
```

hack/docker.sh
naemono: makes a lot of sense.  👍 updated.

Makefile
naemono: yeah, I can get behind that.

Makefile
naemono: yeah, I initially had this in a separate target to make this cleaner, but it was removed in a suggested change.  I'm adding it back here.  Testing in local jenkins to ensure I didn't bork it.

Makefile
```diff
@@ -241,7 +242,10 @@ build-operator-image:
 	|| $(MAKE) docker-build docker-push
 
 build-operator-multiarch-image:
-	@ docker buildx imagetools inspect $(OPERATOR_IMAGE) | grep -q 'linux/arm64' 2>&1 >/dev/null \
+	@ hack/docker.sh -l -m $(OPERATOR_IMAGE)
+	@ hack/docker.sh -l -m $(OPERATOR_DOCKERHUB_IMAGE)
+	@ (docker buildx imagetools inspect $(OPERATOR_IMAGE) | grep -q 'linux/arm64' 2>&1 >/dev/null \
+	|| docker buildx imagetools inspect $(OPERATOR_DOCKERHUB_IMAGE) | grep -q 'linux/arm64' 2>&1 >/dev/null) \
```
thbkrkr, line 248: I think it's `&&` to not call `docker-multiarch-build` only if both images exist. With `||`, we will consider it's OK as soon as an image exists.
```suggestion
	&& docker buildx imagetools inspect $(OPERATOR_DOCKERHUB_IMAGE) | grep -q 'linux/arm64' 2>&1 >/dev/null) \
```

naemono: Agreed, thanks.  Something is still up with the authentication in `hack/docker.sh` I'm working through.

naemono:
Fyi, still running into issues after vault secret updates.  Working with sec/infra team to resolve.

naemono:
Ok, issues appear to be resolved.  This is G2G

thbkrkr (approved):
LGTM :rocket:

PR #5005, opened 2021-11-02
Isolate operator from http service misconfiguration

### LINKED ISSUE #4394: Isolate the operator from misconfiguration of HTTP service
Related https://github.com/elastic/cloud-on-k8s/issues/3732 and https://github.com/elastic/cloud-on-k8s/issues/2572

Currently it is possible for users to misconfigure the default HTTP service in such a way the ability of the operator to roll out further configuration changes is impeded.

Consider: 
```yaml
...
spec:
  http:
    service:
      spec:
        selector:
          app: coordinating-node
...
  - config:
       node.roles: []
    count: 1
    name: coordinating-node
    podTemplate:
      metadata:
        labels:
          app: client-node

```

There two ways for users to fix this misconfiguration: a) either change the service selector to `client-node` or b) change the labels on the `client-node` nodeSet to `coordinating-node`.  Approach a) would work while b) would not 

Once the user has rolled out the change with the incorrect selector in the service, the operator has only limited ability to roll out further changes because it lost connectivity with Elasticsearch. In such a situation any rolling upgrade of existing nodes is effectively blocked while new nodes can still be added to the cluster. Also any updates to the service would be rolled out. This is because these operations don't require API access to Elasticsearch while a rolling upgrade or a downscale does require API access. 

We could consider isolating the operator from this class of configuration errors by either not relying on the service (talking to Pods directly as we already when we switch form HTTP to HTTPS ) or creating an internal service that users cannot configure that is used by the operator.

naemono:
When constructing ES client during ES reconcile, choose random healthy pod instead of service.
When constructing ES client during autoscaling, choose random healthy pod instead of service.

resolves #4394

### DIFF
--- a/pkg/controller/autoscaling/elasticsearch/controller.go
+++ b/pkg/controller/autoscaling/elasticsearch/controller.go
@@ -192,7 +192,7 @@ func newElasticsearchClient(
 	if err != nil {
 		return nil, err
 	}
-	url, err := services.ElasticsearchURLFromRandomPod(es, resourcesState.CurrentPodsByPhase[corev1.PodRunning])
+	url, err := services.ElasticsearchURLFromRandomPod(resourcesState.CurrentPodsByPhase[corev1.PodRunning])
 	if err != nil {
 		log := logconf.FromContext(ctx)
 		log.V(1).Info("could not get a URL to communicate with elasticsearch from a random pod, failing back to using external service", "error", err)
--- a/pkg/controller/autoscaling/elasticsearch/controller.go
+++ b/pkg/controller/autoscaling/elasticsearch/controller.go
@@ -192,7 +192,12 @@ func newElasticsearchClient(
 	if err != nil {
 		return nil, err
 	}
-	url := services.AttemptRandomElasticsearchPodURL(es, resourcesState.CurrentPodsByPhase[corev1.PodPhase(corev1.PodReady)])
+	url, err := services.ElasticsearchURLFromRandomPod(es, resourcesState.CurrentPodsByPhase[corev1.PodRunning])
+	if err != nil {
+		log := logconf.FromContext(ctx)
+		log.V(1).Info("could not get a URL to communicate with elasticsearch from a random pod, failing back to using external service", "error", err)
+		url = services.ElasticsearchURL(es, resourcesState.CurrentPodsByPhase[corev1.PodRunning])
+	}
 	v, err := version.Parse(es.Spec.Version)
 	if err != nil {
 		return nil, err
--- a/pkg/controller/autoscaling/elasticsearch/controller.go
+++ b/pkg/controller/autoscaling/elasticsearch/controller.go
@@ -27,6 +27,7 @@ import (
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common/tracing"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common/version"
 	esclient "github.com/elastic/cloud-on-k8s/pkg/controller/elasticsearch/client"
+	esreconciler "github.com/elastic/cloud-on-k8s/pkg/controller/elasticsearch/reconcile"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/elasticsearch/services"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/elasticsearch/user"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/elasticsearch/validation"
@@ -187,7 +188,11 @@ func newElasticsearchClient(
 	es esv1.Elasticsearch,
 ) (esclient.Client, error) {
 	defer tracing.Span(&ctx)()
-	url := services.ExternalServiceURL(es)
+	resourcesState, err := esreconciler.NewResourcesStateFromAPI(c, es)
+	if err != nil {
+		return nil, err
+	}
+	url := services.RandomElasticsearchPodURL(es, resourcesState.CurrentPodsByPhase[corev1.PodRunning])
 	v, err := version.Parse(es.Spec.Version)
 	if err != nil {
 		return nil, err
--- a/pkg/controller/autoscaling/elasticsearch/controller.go
+++ b/pkg/controller/autoscaling/elasticsearch/controller.go
@@ -192,7 +192,7 @@ func newElasticsearchClient(
 	if err != nil {
 		return nil, err
 	}
-	url := services.RandomElasticsearchPodURL(es, resourcesState.CurrentPodsByPhase[corev1.PodRunning])
+	url := services.AttemptRandomElasticsearchPodURL(es, resourcesState.CurrentPodsByPhase[corev1.PodPhase(corev1.PodReady)])
 	v, err := version.Parse(es.Spec.Version)
 	if err != nil {
 		return nil, err
--- a/pkg/controller/autoscaling/elasticsearch/driver.go
+++ b/pkg/controller/autoscaling/elasticsearch/driver.go
@@ -23,6 +23,7 @@ import (
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common/reconciler"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common/tracing"
 	esReconcile "github.com/elastic/cloud-on-k8s/pkg/controller/elasticsearch/reconcile"
+	"github.com/elastic/cloud-on-k8s/pkg/utils/k8s"
 	logconf "github.com/elastic/cloud-on-k8s/pkg/utils/log"
 )
 
@@ -98,7 +99,12 @@ func (r *ReconcileElasticsearch) isElasticsearchReachable(ctx context.Context, e
 	if err != nil {
 		return false, tracing.CaptureError(ctx, err)
 	}
-	return len(resourcesState.CurrentPodsByPhase[corev1.PodPhase(corev1.PodReady)]) > 0, nil
+	for _, pod := range resourcesState.CurrentPods {
+		if pod.Status.Phase == corev1.PodRunning && k8s.IsPodReady(pod) {
+			return true, nil
+		}
+	}
+	return false, nil
 }
 
 // attemptOnlineReconciliation attempts an online autoscaling reconciliation with a call to the Elasticsearch autoscaling API.
--- a/pkg/controller/autoscaling/elasticsearch/driver.go
+++ b/pkg/controller/autoscaling/elasticsearch/driver.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/go-logr/logr"
 	"go.elastic.co/apm"
+	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
@@ -21,7 +22,7 @@ import (
 	"github.com/elastic/cloud-on-k8s/pkg/controller/autoscaling/elasticsearch/status"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common/reconciler"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common/tracing"
-	esclient "github.com/elastic/cloud-on-k8s/pkg/controller/elasticsearch/client"
+	esReconcile "github.com/elastic/cloud-on-k8s/pkg/controller/elasticsearch/reconcile"
 	logconf "github.com/elastic/cloud-on-k8s/pkg/utils/log"
 )
 
@@ -90,19 +91,14 @@ func newStatusBuilder(log logr.Logger, autoscalingSpec esv1.AutoscalingSpec) *st
 	return statusBuilder
 }
 
-// Check if the Service is available.
+// Check if the Elasticsearch has any pods ready to handle requests.
 func (r *ReconcileElasticsearch) isElasticsearchReachable(ctx context.Context, es esv1.Elasticsearch) (bool, error) {
 	defer tracing.Span(&ctx)()
-	esClient, err := r.esClientProvider(ctx, r.Client, r.Dialer, es)
+	resourcesState, err := esReconcile.NewResourcesStateFromAPI(r.Client, es)
 	if err != nil {
 		return false, tracing.CaptureError(ctx, err)
 	}
-	var health esclient.Health
-	health, err = esClient.GetClusterHealth(ctx)
-	if err != nil || (err == nil && health.TimedOut) {
-		return false, tracing.CaptureError(ctx, err)
-	}
-	return true, nil
+	return len(resourcesState.CurrentPodsByPhase[corev1.PodPhase(corev1.PodReady)]) > 0, nil
 }
 
 // attemptOnlineReconciliation attempts an online autoscaling reconciliation with a call to the Elasticsearch autoscaling API.
--- a/pkg/controller/autoscaling/elasticsearch/driver.go
+++ b/pkg/controller/autoscaling/elasticsearch/driver.go
@@ -21,7 +21,7 @@ import (
 	"github.com/elastic/cloud-on-k8s/pkg/controller/autoscaling/elasticsearch/status"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common/reconciler"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common/tracing"
-	"github.com/elastic/cloud-on-k8s/pkg/controller/elasticsearch/services"
+	esclient "github.com/elastic/cloud-on-k8s/pkg/controller/elasticsearch/client"
 	logconf "github.com/elastic/cloud-on-k8s/pkg/utils/log"
 )
 
@@ -93,18 +93,16 @@ func newStatusBuilder(log logr.Logger, autoscalingSpec esv1.AutoscalingSpec) *st
 // Check if the Service is available.
 func (r *ReconcileElasticsearch) isElasticsearchReachable(ctx context.Context, es esv1.Elasticsearch) (bool, error) {
 	defer tracing.Span(&ctx)()
-	externalService, err := services.GetExternalService(r.Client, es)
-	if apierrors.IsNotFound(err) {
-		return false, nil
-	}
+	esClient, err := r.esClientProvider(ctx, r.Client, r.Dialer, es)
 	if err != nil {
 		return false, tracing.CaptureError(ctx, err)
 	}
-	esReachable, err := services.IsServiceReady(r.Client, externalService)
-	if err != nil {
+	var health esclient.Health
+	health, err = esClient.GetClusterHealth(ctx)
+	if err != nil || (err == nil && health.TimedOut) {
 		return false, tracing.CaptureError(ctx, err)
 	}
-	return esReachable, nil
+	return true, nil
 }
 
 // attemptOnlineReconciliation attempts an online autoscaling reconciliation with a call to the Elasticsearch autoscaling API.
--- a/pkg/controller/autoscaling/elasticsearch/driver.go
+++ b/pkg/controller/autoscaling/elasticsearch/driver.go
@@ -92,7 +92,7 @@ func newStatusBuilder(log logr.Logger, autoscalingSpec esv1.AutoscalingSpec) *st
 	return statusBuilder
 }
 
-// Check if the Elasticsearch has any pods ready to handle requests.
+// Check if Elasticsearch has any pods ready to handle requests.
 func (r *ReconcileElasticsearch) isElasticsearchReachable(ctx context.Context, es esv1.Elasticsearch) (bool, error) {
 	defer tracing.Span(&ctx)()
 	resourcesState, err := esReconcile.NewResourcesStateFromAPI(r.Client, es)
--- a/pkg/controller/elasticsearch/driver/driver.go
+++ b/pkg/controller/elasticsearch/driver/driver.go
@@ -191,10 +191,16 @@ func (d *defaultDriver) Reconcile(ctx context.Context) *reconciler.Results {
 	)
 	defer esClient.Close()
 
-	esReachable, err := services.IsServiceReady(d.Client, *externalService)
+	healthCtx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+
+	var health esclient.Health
+	health, err = esClient.GetClusterHealth(healthCtx)
 	if err != nil {
 		return results.WithError(err)
 	}
+
+	esReachable := !health.TimedOut
 
 	var currentLicense esclient.License
 	if esReachable {
--- a/pkg/controller/elasticsearch/driver/driver.go
+++ b/pkg/controller/elasticsearch/driver/driver.go
@@ -309,7 +309,11 @@ func (d *defaultDriver) newElasticsearchClient(
 	v version.Version,
 	caCerts []*x509.Certificate,
 ) esclient.Client {
-	url := services.AttemptRandomElasticsearchPodURL(d.ES, state.CurrentPodsByPhase[corev1.PodPhase(corev1.PodReady)])
+	url, err := services.ElasticsearchURLFromRandomPod(d.ES, state.CurrentPodsByPhase[corev1.PodRunning])
+	if err != nil {
+		log.V(1).Info("could not get a URL to communicate with elasticsearch from a random pod, failing back to using external service", "error", err)
+		url = services.ElasticsearchURL(d.ES, state.CurrentPodsByPhase[corev1.PodRunning])
+	}
 	return esclient.NewElasticsearchClient(
 		d.OperatorParameters.Dialer,
 		k8s.ExtractNamespacedName(&d.ES),
--- a/pkg/controller/elasticsearch/driver/driver.go
+++ b/pkg/controller/elasticsearch/driver/driver.go
@@ -327,7 +327,12 @@ func (d *defaultDriver) isReachable() (bool, error) {
 	if err != nil {
 		return false, err
 	}
-	return len(resourcesState.CurrentPodsByPhase[corev1.PodPhase(corev1.PodReady)]) > 0, nil
+	for _, pod := range resourcesState.CurrentPods {
+		if pod.Status.Phase == corev1.PodRunning && k8s.IsPodReady(pod) {
+			return true, nil
+		}
+	}
+	return false, nil
 }
 
 // warnUnsupportedDistro sends an event of type warning if the Elasticsearch Docker image is not a supported
--- a/pkg/controller/elasticsearch/driver/driver.go
+++ b/pkg/controller/elasticsearch/driver/driver.go
@@ -195,12 +195,11 @@ func (d *defaultDriver) Reconcile(ctx context.Context) *reconciler.Results {
 	defer cancel()
 
 	var health esclient.Health
+	esReachable := false
 	health, err = esClient.GetClusterHealth(healthCtx)
-	if err != nil {
-		return results.WithError(err)
+	if err == nil && !health.TimedOut {
+		esReachable = true
 	}
-
-	esReachable := !health.TimedOut
 
 	var currentLicense esclient.License
 	if esReachable {
--- a/pkg/controller/elasticsearch/driver/driver.go
+++ b/pkg/controller/elasticsearch/driver/driver.go
@@ -191,7 +191,10 @@ func (d *defaultDriver) Reconcile(ctx context.Context) *reconciler.Results {
 	)
 	defer esClient.Close()
 
-	esReachable := d.isReachable()
+	var esReachable bool
+	if esReachable, err = d.isReachable(); err != nil {
+		return results.WithError(err)
+	}
 
 	var currentLicense esclient.License
 	if esReachable {
@@ -318,12 +321,13 @@ func (d *defaultDriver) newElasticsearchClient(
 	)
 }
 
-func (d *defaultDriver) isReachable() bool {
+// isReachable will check if Elasticsearch has any pods ready to handle requests.
+func (d *defaultDriver) isReachable() (bool, error) {
 	resourcesState, err := reconcile.NewResourcesStateFromAPI(d.Client, d.ES)
 	if err != nil {
-		return false
+		return false, err
 	}
-	return len(resourcesState.CurrentPodsByPhase[corev1.PodPhase(corev1.PodReady)]) > 0
+	return len(resourcesState.CurrentPodsByPhase[corev1.PodPhase(corev1.PodReady)]) > 0, nil
 }
 
 // warnUnsupportedDistro sends an event of type warning if the Elasticsearch Docker image is not a supported
--- a/pkg/controller/elasticsearch/driver/driver.go
+++ b/pkg/controller/elasticsearch/driver/driver.go
@@ -309,7 +309,7 @@ func (d *defaultDriver) newElasticsearchClient(
 	v version.Version,
 	caCerts []*x509.Certificate,
 ) esclient.Client {
-	url := services.ElasticsearchURL(d.ES, state.CurrentPodsByPhase[corev1.PodRunning])
+	url := services.RandomElasticsearchPodURL(d.ES, state.CurrentPodsByPhase[corev1.PodRunning])
 	return esclient.NewElasticsearchClient(
 		d.OperatorParameters.Dialer,
 		k8s.ExtractNamespacedName(&d.ES),
--- a/pkg/controller/elasticsearch/driver/driver.go
+++ b/pkg/controller/elasticsearch/driver/driver.go
@@ -309,10 +309,10 @@ func (d *defaultDriver) newElasticsearchClient(
 	v version.Version,
 	caCerts []*x509.Certificate,
 ) esclient.Client {
-	url, err := services.ElasticsearchURLFromRandomPod(d.ES, state.CurrentPodsByPhase[corev1.PodRunning])
+	url, err := services.ElasticsearchURLFromRandomPod(state.CurrentPodsByPhase[corev1.PodRunning])
 	if err != nil {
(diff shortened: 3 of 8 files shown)


pkg/controller/elasticsearch/services/services.go
```diff
@@ -103,6 +103,7 @@ func NewExternalService(es esv1.Elasticsearch) *corev1.Service {
 }
 
 // IsServiceReady checks if a service has one or more ready endpoints.
+// *Note* this is unused now.  Should we simply remove? --mmontgomery
```
thbkrkr, line 106: Yes, I would remove this function as it is now dead code. Git is our kind memory to help us if we ever want to restore it.

pebrc (commented):
Hey 👋 , I ran some tests (looked good!) and tried to think of edge cases. Some initial thoughts in the comments. I think I need to sleep over the implications of this change some more. I know we wanted to fix a real problem we observed on a users installation with this but I guess we have to make sure that we don't make things worse. One thing I realised is that the long running observers now always hit the same Pod, I know with services it depends a lot on the kube-proxy configuration how the Pod is picked but still something to think about some more.

pkg/controller/elasticsearch/driver/driver.go
```diff
@@ -191,10 +191,12 @@ func (d *defaultDriver) Reconcile(ctx context.Context) *reconciler.Results {
...
-	esReachable, err := services.IsServiceReady(d.Client, *externalService)
-	if err != nil {
-		return results.WithError(err)
-	}
+	healthCtx, cancel := context.WithTimeout(context.Background(), esclient.Timeout(d.ES))
+	defer cancel()
+
+	var health esclient.Health
+	health, err = esClient.GetClusterHealth(healthCtx)
+	esReachable := (err == nil && !health.TimedOut)
```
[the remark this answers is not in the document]
naemono: This makes sense.  Unfortunately the esClient in `pkg/controller/elasticsearch/client` has no concept of pods, or kubernetes at all, so implementing that logic in that package seems to be out of scope of that particular package.  I could add a method on the default elasticsearch driver in `pkg/controller/es/driver`, which is what I've done, which seems to work well.  Let me know your thoughts on that implementation.

pkg/controller/elasticsearch/services/services.go
```diff
@@ -148,10 +149,21 @@ func ElasticsearchURL(es esv1.Elasticsearch, pods []corev1.Pod) string {
...
-		}
+		return RandomElasticsearchPodURL(es, pods)
+	}
+	return ExternalServiceURL(es)
+}
+
+// RandomElasticsearchPodURL will return a URL to communicate with a random Elasticsearch pod from the
+// given set of pods.  If for some reason a pod URL cannot be generated, the external Elasticsearch
+// service URL will be returned.
+func RandomElasticsearchPodURL(es esv1.Elasticsearch, pods []corev1.Pod) string {
```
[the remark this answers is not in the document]
naemono: Ok, I updated this to be `AttemptRandomElasticsearchPodURL`, but I don't feel great about that name either.  Any ideas on a better one?

pkg/controller/elasticsearch/driver/driver.go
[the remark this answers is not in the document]
pebrc: Yes, you are right, let's not add anything to `elasticsearch/client`. As I said I think we don't need this additional check in the current form at all.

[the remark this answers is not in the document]
naemono: > As I said I think we don't need this additional check in the current form at all.

Right, I removed the referenced form of the check hitting `_cluster/health`, and moved it to checking whether there are any ready pods, which seems to be what you were indicating needs to happen from the below comment.

> The purpose of the esReacheable is just to be able to short-circuit in case there is not even a chance that a Pod will respond. So I think it would be OK to just rely on the readiness property of the Pods here.

Let me know if that is your intent, thanks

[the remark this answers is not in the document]
pebrc: Ah sorry, I had not seen your latest commit when I wrote that. I will take another look.

pkg/controller/elasticsearch/driver/driver.go
```diff
@@ -321,6 +321,15 @@ func (d *defaultDriver) newElasticsearchClient(
 	)
 }
 
+// isReachable will check if Elasticsearch has any pods ready to handle requests.
+func (d *defaultDriver) isReachable() (bool, error) {
+	resourcesState, err := reconcile.NewResourcesStateFromAPI(d.Client, d.ES)
+	if err != nil {
+		return false, err
+	}
+	return len(resourcesState.CurrentPodsByPhase[corev1.PodPhase(corev1.PodReady)]) > 0, nil
```
pebrc, line 330: This won't work `Ready` is a Pod condition and not a phase.

pkg/controller/elasticsearch/services/services.go
```diff
@@ -148,10 +132,21 @@ func ElasticsearchURL(es esv1.Elasticsearch, pods []corev1.Pod) string {
...
-		}
+		return AttemptRandomElasticsearchPodURL(es, pods)
+	}
+	return ExternalServiceURL(es)
+}
+
+// AttemptRandomElasticsearchPodURL will return a URL to communicate with a random Elasticsearch pod from the
+// given set of pods.  If for some reason a pod URL cannot be generated, the external Elasticsearch
+// service URL will be returned.
+func AttemptRandomElasticsearchPodURL(es esv1.Elasticsearch, pods []corev1.Pod) string {
```
pebrc, line 143: I know naming is hard: how about simply `ElasticsearchURL` and you are already describing in the Godoc that it can fall back to the service URL.

pkg/controller/elasticsearch/driver/driver.go
naemono: Good point.  I've updated this to take into account both the phase of the pod, and the conditions.  There's also a set of tests around this behavior.

pkg/controller/elasticsearch/services/services.go
naemono: https://github.com/elastic/cloud-on-k8s/blob/master/pkg/controller/elasticsearch/services/services.go#L140

Thanks for the suggestion, unfortunately we already have `ElasticsearchURL`, and it conditionally uses this func in this branch.

pkg/controller/elasticsearch/services/services.go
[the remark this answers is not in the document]
naemono: This now `ElasticsearchURLFromRandomPod` and returns an error when a URL cannot be constructed from the pod slice, per conversation.

pkg/controller/elasticsearch/services/services.go
naemono: Updated per discussion.

pkg/controller/elasticsearch/driver/driver.go
```diff
@@ -309,7 +309,11 @@ func (d *defaultDriver) newElasticsearchClient(
 	v version.Version,
 	caCerts []*x509.Certificate,
 ) esclient.Client {
-	url := services.ElasticsearchURL(d.ES, state.CurrentPodsByPhase[corev1.PodRunning])
+	url, err := services.ElasticsearchURLFromRandomPod(d.ES, state.CurrentPodsByPhase[corev1.PodRunning])
+	if err != nil {
+		log.V(1).Info("could not get a URL to communicate with elasticsearch from a random pod, failing back to using external service", "error", err)
+		url = services.ElasticsearchURL(d.ES, state.CurrentPodsByPhase[corev1.PodRunning])
```
barkbay, line 315: I'm not sure there is point to call `services.ElasticsearchURL` here as its only purpose is to call `ElasticsearchURLFromRandomPod` once again if `scheme` has changed, and we already know that `ElasticsearchURLFromRandomPod` returns an error.
```suggestion
		url = services.ExternalServiceURL(d.ES)
```

pkg/controller/autoscaling/elasticsearch/driver.go
```diff
@@ -90,21 +92,19 @@ func newStatusBuilder(log logr.Logger, autoscalingSpec esv1.AutoscalingSpec) *st
 	return statusBuilder
 }
 
-// Check if the Service is available.
+// Check if the Elasticsearch has any pods ready to handle requests.
```
barkbay, line 95:
```suggestion
// Check if Elasticsearch has any pods ready to handle requests.
```

pkg/controller/autoscaling/elasticsearch/controller.go
```diff
@@ -187,7 +188,16 @@ func newElasticsearchClient(
...
-	url := services.ExternalServiceURL(es)
+	resourcesState, err := esreconciler.NewResourcesStateFromAPI(c, es)
+	if err != nil {
+		return nil, err
+	}
+	url, err := services.ElasticsearchURLFromRandomPod(es, resourcesState.CurrentPodsByPhase[corev1.PodRunning])
+	if err != nil {
+		log := logconf.FromContext(ctx)
+		log.V(1).Info("could not get a URL to communicate with elasticsearch from a random pod, failing back to using external service", "error", err)
+		url = services.ElasticsearchURL(es, resourcesState.CurrentPodsByPhase[corev1.PodRunning])
```
barkbay, line 199: Same comment as the one on `driver.go`. Also it looks like duplicated code, which might mean that we want the same behaviour in all the cases:
* Try to get the URL from a Pod as a best effort
* Fallback on the external service if it is not possible

Should we just implement this mechanism in `ElasticsearchURLFromRandomPod` ?

barkbay (commented):
Looks good overall 👍 
Just left a few nitpicks/questions.

pkg/controller/autoscaling/elasticsearch/controller.go
naemono: This is how I initially had things setup, but upon some comments from @pebrc  https://github.com/elastic/cloud-on-k8s/pull/5005#discussion_r745055036, I updated it to do just one thing, and error if it can't.  

Do both of you prefer for this func to just fail back if it can't generate the URL inside this func?

barkbay (commented):
> One thing I realised is that the long running observers now always hit the same Pod

I have a different experience, a new observer is (almost) always created during each reconciliation loop: 

```
2021-11-24T09:55:18.602+0100    INFO    observer        Replacing observer      {"service.version": "2.0.0-SNAPSHOT+mmo", "namespace": "e2e-mercury", "es_name": "es-apm-sample"}
2021-11-24T09:55:18.602+0100    INFO    observer        Stopping observer for cluster   {"service.version": "2.0.0-SNAPSHOT+mmo", "namespace": "e2e-mercury", "es_name": "es-apm-sample"}
2021-11-24T09:55:18.602+0100    INFO    observer        Creating observer for cluster   {"service.version": "2.0.0-SNAPSHOT+mmo", "namespace": "e2e-mercury", "es_name": "es-apm-sample"}
```

This is because using a "random" URL (`Endpoint`) makes `!observer.esClient.Equal(esClient)`  likely to be always `true`: https://github.com/elastic/cloud-on-k8s/blob/7e71c1bcc56d24bb19e623f3b3ac456b0d4f8724/pkg/controller/elasticsearch/observer/manager.go#L69-L70

https://github.com/elastic/cloud-on-k8s/blob/7e71c1bcc56d24bb19e623f3b3ac456b0d4f8724/pkg/controller/elasticsearch/client/base.go#L60-L62

naemono:
> This is because using a "random" URL (`Endpoint`) makes `!observer.esClient.Equal(esClient)` likely to be always `true`:

I see this now.  The `Observer` piece of the ES driver is something I wasn't terribly familiar with.  After reviewing that package, I'm a bit more familiar with what's going on.  This is triggering the behavior now, as we are building a new ES client on every reconcile call, which is now, with this PR, choosing a random pod:

https://github.com/elastic/cloud-on-k8s/blob/2df4d0c822c82bce18a9baa7fc83e4f28aa07819/pkg/controller/elasticsearch/driver/driver.go#L170-L175

Is the ES client something we might want to persist across reconciliation calls?  We currently persist a K8s client, and some other objects on the `defaultDriver`.  During a pod/node failure, or an upgrade, where the pod to which this client is pointing is removed/replaced, we would need to build a new client, or potentially when the pod we are pointed at simply becomes `unhealthy` (what is the source of truth for this?  k8s pod health, or health endpoint of ES on this node?).  Are there hard rules we would be able to follow that could decide whether we would need to re-generate this ES client on each reconcile call?

pebrc:
> Is the ES client something we might want to persist across reconciliation calls? We currently persist a K8s client, and some other objects on the defaultDriver

Nothing is persistent on the driver level across reconciliations. We create a new driver on each reconciliation event/iteration. The only thing existing outside of reconciliations are the observers which hold state in form of an Elasticsearch client configured with a set of certificates and now a Pod-specific URL, and of course the health observation itself. 

> > One thing I realised is that the long running observers now always hit the same Pod
> I have a different experience, a new observer is (almost) always created during each reconciliation loop:

I think we have to distinguish different phases here: during reconciliation we are now recreating the observers on (almost) every iteration due to the randomness introduced in the Pod URL selection. However, once we reach the desired state we keep hitting the same Pod 

```
2021-11-24T19:38:20.951+0100	DEBUG	observer	Retrieving cluster state	{"service.version": "2.0.0-SNAPSHOT+9814f77d", "es_name": "es", "namespace": "default"}
2021-11-24T19:38:20.951+0100	DEBUG	elasticsearch-client	Elasticsearch HTTP request	{"service.version": "2.0.0-SNAPSHOT+9814f77d", "method": "GET", "url": "https://es-es-mixed-2.es-es-mixed.default:9200/_cluster/health", "namespace": "default", "es_name": "es"}
2021-11-24T19:38:30.951+0100	DEBUG	observer	Retrieving cluster state	{"service.version": "2.0.0-SNAPSHOT+9814f77d", "es_name": "es", "namespace": "default"}
2021-11-24T19:38:30.951+0100	DEBUG	elasticsearch-client	Elasticsearch HTTP request	{"service.version": "2.0.0-SNAPSHOT+9814f77d", "method": "GET", "url": "https://es-es-mixed-2.es-es-mixed.default:9200/_cluster/health", "namespace": "default", "es_name": "es"}
2021-11-24T19:38:40.950+0100	DEBUG	observer	Retrieving cluster state	{"service.version": "2.0.0-SNAPSHOT+9814f77d", "es_name": "es", "namespace": "default"}
2021-11-24T19:38:40.950+0100	DEBUG	elasticsearch-client	Elasticsearch HTTP request	{"service.version": "2.0.0-SNAPSHOT+9814f77d", "method": "GET", "url": "https://es-es-mixed-2.es-es-mixed.default:9200/_cluster/health", "namespace": "default", "es_name": "es"}
2021-11-24T19:38:50.951+0100	DEBUG	observer	Retrieving cluster state	{"service.version": "2.0.0-SNAPSHOT+9814f77d", "es_name": "es", "namespace": "default"}
2021-11-24T19:38:50.951+0100	DEBUG	elasticsearch-client	Elasticsearch HTTP request	{"service.version": "2.0.0-SNAPSHOT+9814f77d", "method": "GET", "url": "https://es-es-mixed-2.es-es-mixed.default:9200/_cluster/health", "namespace": "default", "es_name": "es"}
2021-11-24T19:39:00.950+0100	DEBUG	observer	Retrieving cluster state	{"service.version": "2.0.0-SNAPSHOT+9814f77d", "es_name": "es", "namespace": "default"}
2021-11-24T19:39:00.950+0100	DEBUG	elasticsearch-client	Elasticsearch HTTP request	{"service.version": "2.0.0-SNAPSHOT+9814f77d", "method": "GET", "url": "https://es-es-mixed-2.es-es-mixed.default:9200/_cluster/health", "namespace": "default", "es_name": "es"}
2021-11-24T19:39:10.953+0100	DEBUG	observer	Retrieving cluster state	{"service.version": "2.0.0-SNAPSHOT+9814f77d", "es_name": "es", "namespace": "default"}
2021-11-24T19:39:10.953+0100	DEBUG	elasticsearch-client	Elasticsearch HTTP request	{"service.version": "2.0.0-SNAPSHOT+9814f77d", "method": "GET", "url": "https://es-es-mixed-2.es-es-mixed.default:9200/_cluster/health", "namespace": "default", "es_name": "es"}
2021-11-24T19:39:20.953+0100	DEBUG	observer	Retrieving cluster state	{"service.version": "2.0.0-SNAPSHOT+9814f77d", "es_name": "es", "namespace": "default"}
2021-11-24T19:39:20.953+0100	DEBUG	elasticsearch-client	Elasticsearch HTTP request	{"service.version": "2.0.0-SNAPSHOT+9814f77d", "method": "GET", "url": "https://es-es-mixed-2.es-es-mixed.default:9200/_cluster/health", "namespace": "default", "es_name": "es"}
2021-11-24T19:39:30.940+0100	DEBUG	observer	Retrieving cluster state	{"service.version": "2.0.0-SNAPSHOT+9814f77d", "es_name": "es", "namespace": "default"}
2021-11-24T19:39:30.940+0100	DEBUG	elasticsearch-client	Elasticsearch HTTP request	{"service.version": "2.0.0-SNAPSHOT+9814f77d", "method": "GET", "url": "https://es-es-mixed-2.es-es-mixed.default:9200/_cluster/health", "namespace": "default", "es_name": "es"}
2021-11-24T19:39:40.927+0100	DEBUG	observer	Retrieving cluster state	{"service.version": "2.0.0-SNAPSHOT+9814f77d", "es_name": "es", "namespace": "default"}
2021-11-24T19:39:40.927+0100	DEBUG	elasticsearch-client	Elasticsearch HTTP request	{"service.version": "2.0.0-SNAPSHOT+9814f77d", "method": "GET", "url": "https://es-es-mixed-2.es-es-mixed.default:9200/_cluster/health", "namespace": "default", "es_name": "es"}
2021-11-24T19:39:50.923+0100	DEBUG	observer	Retrieving cluster state	{"service.version": "2.0.0-SNAPSHOT+9814f77d", "es_name": "es", "namespace": "default"}
2021-11-24T19:39:50.923+0100	DEBUG	elasticsearch-client	Elasticsearch HTTP request	{"service.version": "2.0.0-SNAPSHOT+9814f77d", "method": "GET", "url": "https://es-es-mixed-2.es-es-mixed.default:9200/_cluster/health", "namespace": "default", "es_name": "es"}
```

This would be a behaviour change I believe from standard kube-proxy-in-iptables-mode behaviour when using a service URL where we should get a random backend Pod. Let's make sure we understand the potential downsides here before we commit to this behaviour change.

barkbay (commented):
> I think we have to distinguish different phases here: during reconciliation we are now recreating the observers on (almost) every iteration due to the randomness introduced in the Pod URL selection. However, once we reach the desired state we keep hitting the same Pod

🤦 , you're right.

Could we use the service endpoint in the context of the observer (using `services.ExternalServiceURL(d.ES)` as the url) ? 

Assuming the observer state is only used to populate the status it would still allow the operator to recover from http service misconfiguration ? I think only the health status would be  affected.

pebrc:
I was comparing the change proposed in the PR with our current implementation to understand the impact that switching from a service URL to a randomly chosen Pod URL would have. What I realised is that we already have effectively a similar situation because the observers will keep the TCP connection to a Pod open and reuse it for subsequent requests. 

```
9:32:24.415339 IP 100.96.1.5.55510 > 100.96.1.9.9200: Flags [P.], seq 18428:18699, ack 35497, win 10682, options [nop,nop,TS val 1705480895 ecr 3723947912], length 271
19:32:24.417430 IP 100.96.1.9.9200 > 100.96.1.5.55510: Flags [P.], seq 35497:36019, ack 18699, win 5761, options [nop,nop,TS val 3723957912 ecr 1705480895], length 522
19:32:24.417479 IP 100.96.1.5.55510 > 100.96.1.9.9200: Flags [.], ack 36019, win 10680, options [nop,nop,TS val 1705480897 ecr 3723957912], length 0
19:32:34.414584 IP 100.96.1.5.55510 > 100.96.1.9.9200: Flags [P.], seq 18699:18970, ack 36019, win 10682, options [nop,nop,TS val 1705490894 ecr 3723957912], length 271
19:32:34.416683 IP 100.96.1.9.9200 > 100.96.1.5.55510: Flags [P.], seq 36019:36541, ack 18970, win 5769, options [nop,nop,TS val 3723967911 ecr 1705490894], length 522
19:32:34.416745 IP 100.96.1.5.55510 > 100.96.1.9.9200: Flags [.], ack 36541, win 10680, options [nop,nop,TS val 1705490896 ecr 3723967911], length 0
19:32:44.414723 IP 100.96.1.5.55510 > 100.96.1.9.9200: Flags [P.], seq 18970:19241, ack 36541, win 10682, options [nop,nop,TS val 1705500894 ecr 3723967911], length 271
19:32:44.416534 IP 100.96.1.9.9200 > 100.96.1.5.55510: Flags [P.], seq 36541:37063, ack 19241, win 5778, options [nop,nop,TS val 3723977911 ecr 1705500894], length 522
19:32:44.416596 IP 100.96.1.5.55510 > 100.96.1.9.9200: Flags [.], ack 37063, win 10680, options [nop,nop,TS val 1705500896 ecr 3723977911], length 0
19:32:54.415397 IP 100.96.1.5.55510 > 100.96.1.9.9200: Flags [P.], seq 19241:19512, ack 37063, win 10682, options [nop,nop,TS val 1705510895 ecr 3723977911], length 271
19:32:54.417486 IP 100.96.1.9.9200 > 100.96.1.5.55510: Flags [P.], seq 37063:37585, ack 19512, win 5786, options [nop,nop,TS val 3723987912 ecr 1705510895], length 522
19:32:54.417547 IP 100.96.1.5.55510 > 100.96.1.9.9200: Flags [.], ack 37585, win 10680, options [nop,nop,TS val 1705510897 ecr 3723987912], length 0
19:33:04.415289 IP 100.96.1.5.55510 > 100.96.1.9.9200: Flags [P.], seq 19512:19783, ack 37585, win 10682, options [nop,nop,TS val 1705520895 ecr 3723987912], length 271
19:33:04.417110 IP 100.96.1.9.9200 > 100.96.1.5.55510: Flags [P.], seq 37585:38107, ack 19783, win 5794, options [nop,nop,TS val 3723997911 ecr 1705520895], length 522
19:33:04.417154 IP 100.96.1.5.55510 > 100.96.1.9.9200: Flags [.], ack 38107, win 10680, options [nop,nop,TS val 1705520896 ecr 3723997911], length 0
19:33:14.415291 IP 100.96.1.5.55510 > 100.96.1.9.9200: Flags [P.], seq 19783:20054, ack 38107, win 10682, options [nop,nop,TS val 1705530895 ecr 3723997911], length 271
19:33:14.417071 IP 100.96.1.9.9200 > 100.96.1.5.55510: Flags [P.], seq 38107:38629, ack 20054, win 5803, options [nop,nop,TS val 3724007911 ecr 1705530895], length 522
19:33:14.417115 IP 100.96.1.5.55510 > 100.96.1.9.9200: Flags [.], ack 38629, win 10680, options [nop,nop,TS val 1705530896 ecr 3724007911], length 0
19:33:24.414326 IP 100.96.1.5.55510 > 100.96.1.9.9200: Flags [P.], seq 20054:20325, ack 38629, win 10682, options [nop,nop,TS val 1705540894 ecr 3724007911], length 271
19:33:24.416381 IP 100.96.1.9.9200 > 100.96.1.5.55510: Flags [P.], seq 38629:39151, ack 20325, win 5811, options [nop,nop,TS val 3724017911 ecr 1705540894], length 522
19:33:24.416443 IP 100.96.1.5.55510 > 100.96.1.9.9200: Flags [.], ack 39151, win 10680, options [nop,nop,TS val 1705540896 ecr 3724017911], length 0
19:33:34.414682 IP 100.96.1.5.55510 > 100.96.1.9.9200: Flags [P.], seq 20325:20596, ack 39151, win 10682, options [nop,nop,TS val 1705550894 ecr 3724017911], length 271
19:33:34.416720 IP 100.96.1.9.9200 > 100.96.1.5.55510: Flags [P.], seq 39151:39673, ack 20596, win 5820, options [nop,nop,TS val 3724027911 ecr 1705550894], length 522
19:33:34.416813 IP 100.96.1.5.55510 > 100.96.1.9.9200: Flags [.], ack 39673, win 10680, options [nop,nop,TS val 1705550896 ecr 3724027911], length 0
```

barkbay:
> the observers will keep the TCP connection to a Pod open and reuse it for subsequent requests

Great finding 👍  I think the question is then to understand what would happen if the connection to the node is lost in both implementations. I would expect the actual one to reuse the service to establish a connection to a new Pod, not sure about what would happen with what is proposed in the PR.

naemono:
Thanks @barkbay and @pebrc for further investigating the ramifications of these changes, and continuing to move this forward.  I'll test the 2x scenarios described to determine behavior, and report back.

naemono:
Here's what I'm seeing from my testing
(note this is using an ES cluster without TLS, and sniffing via ngrep to isolate only the observer health calls)

**With ECK 1.8**

Cluster Healthy, operator hitting the same pod via service
```
T 10.32.0.46:49346 -> 10.32.5.26:9200 [AP] #1550
GET /_cluster/health HTTP/1.1.
Host: testing-es-http.default.svc:9200.

T 10.32.0.46:49346 -> 10.32.5.26:9200 [AP] #1787
GET /_cluster/health HTTP/1.1.
Host: testing-es-http.default.svc:9200.
```

After Deleting Pod, and Pod fully terminates, a new pod is hit via service
```
T 10.32.0.46:58434 -> 10.32.2.227:9200 [AP] #2138
GET /_cluster/health HTTP/1.1.
Host: testing-es-http.default.svc:9200.

T 10.32.0.46:58434 -> 10.32.2.227:9200 [AP] #2603
GET /_cluster/health HTTP/1.1.
Host: testing-es-http.default.svc:9200.
```

**With Proposed Changes**

Cluster Healthy, operator hitting the same pod, NO service
```
T 10.32.0.47:58664 -> 10.32.4.30:9200 [AP] #178
GET /_cluster/health HTTP/1.1.
Host: testing-es-masters-resized-v2-2.testing-es-masters-resized-v2.default:9200.

T 10.32.0.47:58664 -> 10.32.4.30:9200 [AP] #425
GET /_cluster/health HTTP/1.1.
Host: testing-es-masters-resized-v2-2.testing-es-masters-resized-v2.default:9200.
```

While Deleting Pod, and until cluster becomes green again, target of observer is random
```
T 10.32.0.47:36096 -> 10.32.5.27:9200 [AP] #1145
GET /_cluster/health HTTP/1.1.
Host: testing-es-masters-resized-v2-1.testing-es-masters-resized-v2.default:9200.

T 10.32.0.47:33210 -> 10.32.2.227:9200 [AP] #1204
GET /_cluster/health HTTP/1.1.
Host: testing-es-masters-resized-v2-0.testing-es-masters-resized-v2.default:9200.
```

Then it stabilizes again:

```
T 10.32.0.47:36852 -> 10.32.5.27:9200 [AP] #6337
GET /_cluster/health HTTP/1.1.
Host: testing-es-masters-resized-v2-1.testing-es-masters-resized-v2.default:9200.

T 10.32.0.47:36852 -> 10.32.5.27:9200 [AP] #6673
GET /_cluster/health HTTP/1.1.
Host: testing-es-masters-resized-v2-1.testing-es-masters-resized-v2.default:9200.

T 10.32.0.47:36852 -> 10.32.5.27:9200 [AP] #6917
GET /_cluster/health HTTP/1.1.
Host: testing-es-masters-resized-v2-1.testing-es-masters-resized-v2.default:9200.
```

barkbay:
I wanted to check what would happen if the Pod used by the observer is partitioned from the rest of the cluster. This is a very edge case: only the ES node is partitioned, not the K8S node, also the Pod is still considered as `Ready` because the local loopback interface is still available.

Here is the protocol I followed:
1. I waited for the observer to report a `green` health and stabilize, hitting the same Pod (`elasticsearch-sample-es-default-1` in this example):

```
2021-12-03T12:09:30.676+0100    DEBUG   elasticsearch-client    Elasticsearch HTTP request      {"service.version": "1.5.0-SNAPSHOT+mmo", "method": "GET", "url": "https://elasticsearch-sample-es-default-1.elasticsearch-sample-es-default.default:9200/_cluster/health", "namespace": "default", "es_name": "elasticsearch-sample"}
```

2. I inserted an iptable rule in the `Forward` chain on the K8S node to drop all the packets targeting `elasticsearch-sample-es-default-1`:

```
Chain FORWARD (policy DROP 0 packets, 0 bytes)
 pkts bytes target     prot opt in     out     source               destination
   55 14680 DROP       all  --  cbr0   *       10.45.114.7          0.0.0.0/0 /* partition elasticsearch-sample-es-default-1  */
75700  173M KUBE-FORWARD  all  --  *      *       0.0.0.0/0            0.0.0.0/0            /* kubernetes forwarding rules */
  333 23606 KUBE-SERVICES  all  --  *      *       0.0.0.0/0            0.0.0.0/0            ctstate NEW /* kubernetes service portals */
  333 23606 KUBE-EXTERNAL-SERVICES  all  --  *      *       0.0.0.0/0            0.0.0.0/0            ctstate NEW /* kubernetes externally-visible service portals */
```

Note that at this point the Pod is still considered as `Ready` since `127.0.0.1` can still be accessed by the readiness probe.

3. As expected the observer was no more able to get the cluster health:

```
2021-12-03T12:09:40.676+0100    DEBUG   observer        Unable to retrieve cluster health       {"error": "elasticsearch client failed for https://elasticsearch-sample-es-default-1.elasticsearch-sample-es-default.default:9200/_cluster/health: Get \"https://elasticsearch-sample-es-default-1.elasticsearch-sample-es-default.default:9200/_cluster/health\": context deadline exceeded", "namespace": "default", "es_name": "elasticsearch-sample"}
```

4. A new reconciliation is started right after the error is raised by the observer. This is because [a generic event is created on health change](https://github.com/elastic/cloud-on-k8s/blob/5e5bbdac30dc1c81e2f6621cda76402a891011f2/pkg/controller/elasticsearch/observer/watch.go#L32-L50) (`green` -> `unknown` in this case). It means a new Pod might be selected:

```
2021-12-03T12:09:40.676+0100    INFO    elasticsearch-controller        Starting reconciliation run     {"service.version": "1.5.0-SNAPSHOT+mmo", "iteration": 32, "namespace": "default", "es_name": "elasticsearch-sample"}
```

The problem here is that if the observer still hit the partitioned Pod, then the cluster health is stuck on `unknown`. This is because [`hasHealthChanged`](https://github.com/elastic/cloud-on-k8s/blob/5e5bbdac30dc1c81e2f6621cda76402a891011f2/pkg/controller/elasticsearch/observer/watch.go#L53) would return false, preventing a new reconciliation to happen. Using the `Service` would offer a chance for the observer to report a status.
This is a (very) edge case though, I'm still not sure myself if we want to consider it.

naemono:
@barkbay thanks for the detailed testing of the changes in this PR.  This got me thinking.  Since we don't seem to adjust the default sync period for the manager, which is 10 hours:

https://github.com/kubernetes-sigs/controller-runtime/blob/f8472cea624747986e82cc950a3abd643931f1a2/pkg/manager/manager.go#L108-L133

The cluster would be stuck in this state for 10 hours, right?  And "hopefully" then it would hit another pod during that next reconcile cycle?    This is because the only thing, outside of k8s objects changing within the cluster, is the state of the cluster changing:

https://github.com/elastic/cloud-on-k8s/blob/5e5bbdac30dc1c81e2f6621cda76402a891011f2/pkg/controller/elasticsearch/elasticsearch_controller.go#L130

What about another `listener` on the ES observer that triggers a reconciliation at a much shorter interval (minutes, not hours) when the health of the cluster is not `green`?  The only other option I see is to have a much more intelligent choosing of the ES node (like reaching out and ensuring health, or at least connectivity when choosing) to which we speak to internally within the operator.

pkg/controller/autoscaling/elasticsearch/controller.go
pebrc: My comment was more about the naming of the function. Not so much about not being able to have a shared function with a fallback. 

I wonder why we are not simply changing the existing `services.ElasticsearchURL` function to have the new behaviour of picking a random Pod by default and falling back to the service URL if that fails for some reason. 

Because the flow we have right now in this PR seems a bit odd: 

We call first `services.ElasticsearchURLFromRandomPod` if that fails we call `services.ElasticsearchURL` which in case of a scheme change once more calls `services.ElasticsearchURLFromRandomPod` (which we already know that it fails) 

I believe if we default to picking a random pod we do not need the scheme change logic at all anymore and we can keep the service URL as a fallback.

sebgl:
* About the isolated Pod edge case: network partitions are not the only thing that could happen here. A Pod could also be Ready (responds to `GET /`), but not have joined the Elasticsearch cluster, I think. If misconfigured, for example. In which case reusing the TCP conn for that Pod will always lead to not being able to retrieve cluster-level information from the API, until the Pod configuration is updated.
* Let's make sure we do recreate a new Elasticsearch client at each reconciliation, not reusing TCP conns, to ensure we don't get stuck in that edge case. I think this is already what happens as we `defer esClient.Close()`  in the driver.
* The main reason for observers to exist is to report the cluster health in the resource status (something that is debatable to start with). They trigger reconciliations for that only purpose. None of the orchestration logic should depend on that. We also shouldn't depend *at all* on the 10 hours manager cache resync period. I think the observer health reporting is really some sort of best-effort thing. Worst thing that can happen if a Pod gets isolated from the cluster is that we report an `unknown` health for some time. I think that's acceptable, the extra effort of detecting that case is not worth the complexity imo. If users see an unknown health, something's not right with at least one of the Pods so that health value is not completely off.
* In the current code it looks like we select a random ready Pod, then fallback to the service URL. This seems strange to me. Are we not trusting Pod IPs to be reliable? I think we could just use Pod IPs only, and consider the Elasticsearch service is an "external" convenience for users. None of the operator code should depend on it, at all. We only use Elasticsearch Pod IPs.
* Let's make sure the reasoning behind using Pod IPs and not Service URLs is documented in the code ("because users can override the service spec in a way that would not allow use to reach out to Elasticsearch").
* The TCP conn reuse was already there before this PR, I think it should not block merging this PR that addresses a different concern.

naemono:
going to close this, as we are leaning towards #5211 instead of the approach defined in this PR.

PR #5019, opened 2021-11-05
Add cpu/mem/storage resources validation in E2E tests

### LINKED ISSUE #4411: E2E Tests: improve resources validation
When the node topology is validated [CheckESNodesTopology](https://github.com/elastic/cloud-on-k8s/blob/2a781357091e87d3a026d08bb3fa8ead43d67b9b/test/e2e/test/elasticsearch/checks_es.go#L88)  only verifies the memory limit. We should expand this validation to `requests` and other resources (cpu, storage).

Follow up of https://github.com/elastic/cloud-on-k8s/pull/4373#discussion_r602172506

thbkrkr:
- Add 3 checks to [`CheckStackTestSteps()`](https://github.com/elastic/cloud-on-k8s/pull/5019/files#diff-00080e264ff64ec6d44d82f3809be50b7aa9f2d964ef21d9c75759e54ae72cbb):
	- `compareCgroupCPULimit` compares the CPU limit specified in a nodeSet with the limit set in the CPU cgroup at the OS level
	- `compareSpecResources` compares the resources specified in the ES resource with the resources specified in the pods
	- `compareClaimedStorage` compares the requested storage specified in a nodeSet with the capacity claimed in the PVC
- Add `TestResourcesRequirements` E2E test
 
Resolves #4411.


test/e2e/test/elasticsearch/checks_es.go
```diff
@@ -211,3 +236,84 @@ func compareMemoryLimit(topologyElement esv1.NodeSet, cgroupMemoryLimitsInBytes
...
+	if expectedResources == nil {
+		expectedResources = &nodespec.DefaultResources
+	}
+	for _, pod := range pods {
+		for _, c := range pod.Spec.Containers {
+			if c.Name == esv1.ElasticsearchContainerName {
+				fmt.Println("{\"compareResources\": \"", reflect.DeepEqual(*expectedResources, c.Resources), "\"}")
+				fmt.Println("{\"compareResources\": \"", *expectedResources, "\"}")
+				fmt.Println("{\"compareResources\": \"", c.Resources, "\"}")
+				if !reflect.DeepEqual(*expectedResources, c.Resources) {
```
barkbay, line 282: It is not possible to compare 2 quantities using `DeepEqual` since a same value may have different internal representations. For example, '1000m' and '1' are semantically equal, but their representations are different.

```suggestion
				if !equality.Semantic.DeepEqual(*expectedResources, c.Resources) {
```

test/e2e/test/elasticsearch/checks_es.go
```diff
@@ -211,3 +236,84 @@ func compareMemoryLimit(topologyElement esv1.NodeSet, cgroupMemoryLimitsInBytes
...
+		}
+		fmt.Println("{\"compareStorage\": \"", reflect.DeepEqual(expectedStorage, actualStorage), "\"}")
+		fmt.Println("{\"compareStorage\": \"", actualStorage, "\"}")
+		fmt.Println("{\"compareStorage\": \"", expectedStorage, "\"}")
+		if !reflect.DeepEqual(expectedStorage, actualStorage) {
+			return false
+		}
+	}
+	return true
+}
```
barkbay, line 319: The actual capacity of a volume might be greater than the requested capacity. It means we can't really compare a PVC and its PV. What we could do instead is comparing the claim in the `nodeSet` with the capacity claimed in the `PVC`:

```suggestion
// compareClaimedStorage compares the requested storage specified in a nodeSet with the actual capacity claimed in the PVC.
func compareClaimedStorage(topologyElement esv1.NodeSet, pvcs []corev1.PersistentVolumeClaim) bool {
	var expectedStorage *resource.Quantity
	for _, v := range topologyElement.VolumeClaimTemplates {
		if v.Name == volume.ElasticsearchDataVolumeName {
			expectedStorage = v.Spec.Resources.Requests.Storage()
			continue
		}
	}
	if expectedStorage == nil {
		// no expected storage, consider it's ok
		return true
	}
	for _, pvc := range pvcs {
		actualStorage := pvc.Spec.Resources.Requests.Storage()
		if !equality.Semantic.DeepEqual(expectedStorage, actualStorage) {
			return false
		}
	}
	return true
}
```

thbkrkr:
run full pr build

thbkrkr:
run full pr build

thbkrkr:
run full pr build

thbkrkr:
run full pr build

thbkrkr:
`TestVolumeExpansion` failed with `expected claimed storage [1073741824], got [2147483648]` because we don't use the updated ES resource but the initial resource still with 1Gi for the comparison.

thbkrkr:
In [`TestVolumeExpansion`](https://github.com/elastic/cloud-on-k8s/blob/bdd6728c014de1f740ec64a1dbacb94259ea9f2b/test/e2e/es/volume_test.go#L193-L194), we update the ES resource with a new volumeClaim in a Step after the initial creation. Only at the moment, we want to update the expected NodeSets used in the ES checks.

There is [`WithExpectedNodeSets()`](https://github.com/elastic/cloud-on-k8s/blob/20f522ff9f5fa04f8ccb22ce19edaa6d044a1aa6/test/e2e/es/autoscaling_test.go#L93-L101) in combination with [`GetExpectedElasticsearch()`](https://github.com/elastic/cloud-on-k8s/blob/9be005d2a4b6e26f98ff9c520ea53ae4d6cc9455/test/e2e/test/elasticsearch/checks_es.go#L99) to handle this sort of use cases where the ES resource is modified by an external mechanism but it only works when it is called on the initial builder used to create the [`test.CheckTestSteps(builder, k)`](https://github.com/elastic/cloud-on-k8s/blob/d9bd0ee2ff37fee52d1698c61d53e4234d1fc874/test/e2e/es/volume_test.go#L232) because we don't pass a pointer and the methods (`WithExpectedNodeSets()` and [`CheckStackTestSteps()`](https://github.com/elastic/cloud-on-k8s/blob/d9bd0ee2ff37fee52d1698c61d53e4234d1fc874/test/e2e/test/checks.go#L24)) don't use a pointer receiver.

I'm looking at the refactoring to get pointers to be able to use `WithExpectedNodeSets()` in `TestVolumeExpansion`. Essentially, it's to make `*elasticsearch.Builder` implement `test.Builder`.

thbkrkr:
run full pr build :crossed_fingers:

thbkrkr:
run full pr build

thbkrkr:
The full pr build fails on `TestElasticMapsServerCrossNSAssociation` because "This is an unlicensed development build of ECK. License management and Enterprise features are disabled". I opened #5065.

thbkrkr:
run full pr build

test/e2e/es/volume_test.go
```diff
@@ -191,7 +191,13 @@ func TestVolumeExpansion(t *testing.T) {
 						return err
 					}
 					patchStorageSize(&es, resizedStorage)
-					return k.Client.Update(context.Background(), &es)
+					err := k.Client.Update(context.Background(), &es)
+					if err != nil {
+						return err
+					}
+					// update the expected nodeSets in the builder because the ES resource has just been modified by an external mechanism
+					b.WithExpectedNodeSets(es.Spec.NodeSets...)
```
barkbay, line 199: > Essentially, it's to make *elasticsearch.Builder implement test.Builder

Could we not just rely on 2 builders here ? One to create the initial deployment and a second one to describe the expected state ?
Something along those lines:

```golang
	// Create a copy of the builder with the expected storage resources.
	scaledUpStorage := b.DeepCopy()
	for i := range scaledUpStorage.Elasticsearch.Spec.NodeSets {
		scaledUpStorage.Elasticsearch.Spec.NodeSets[i].VolumeClaimTemplates[0].Spec.Resources.Requests[corev1.ResourceStorage] = resizedStorage
	}

	test.Sequence(nil, func(k *test.K8sClient) test.StepList {
		return test.StepList{
			{
				Name: "Update the Elasticsearch spec with increased storage requests",
					var es esv1.Elasticsearch
					if err := k.Client.Get(context.Background(), k8s.ExtractNamespacedName(&b.Elasticsearch), &es); err != nil {
						return err
					}
					patchStorageSize(&es, resizedStorage)
					return k.Client.Update(context.Background(), &es)
				}),
			},
			{
				Name: "PVCs should eventually be resized",
					for _, pvcName := range pvcNames {
						var pvc corev1.PersistentVolumeClaim
						if err := k.Client.Get(context.Background(), types.NamespacedName{Namespace: b.Elasticsearch.Namespace, Name: pvcName}, &pvc); err != nil {
							return err
						}
						reportedStorage := pvc.Status.Capacity.Storage()
						if reportedStorage == nil {
							return fmt.Errorf("no storage size reported in %s status", pvcName)
						}
						if !reportedStorage.Equal(resizedStorage) {
							return fmt.Errorf("expected resized capacity %s but got %s", resizedStorage.String(), reportedStorage.String())
						}
					}
					return nil
				}),
			},
			{
				Name: "StatefulSets should eventually be recreated with the right storage size",
					for _, ssetName := range []string{masterSset, dataSset} {
						var sset appsv1.StatefulSet
						if err := k.Client.Get(context.Background(), types.NamespacedName{Namespace: b.Elasticsearch.Namespace, Name: ssetName}, &sset); err != nil {
							return err
						}
						if !sset.Spec.VolumeClaimTemplates[0].Spec.Resources.Requests.Storage().Equal(resizedStorage) {
							return fmt.Errorf("StatefulSet %s has not been recreated with storage size %s", ssetName, resizedStorage.String())
						}
					}
					return nil
				}),
			},
			// re-run all the regular checks
		}.WithSteps(test.CheckTestSteps(scaledUpStorage, k))
	}, b).RunSequential(t)
```

thbkrkr: It's much smarter indeed.

thbkrkr:
run full pr build

test/e2e/test/elasticsearch/checks_es.go
```diff
@@ -185,29 +171,173 @@ func (e *esClusterChecks) CheckESNodesTopology() test.Step {
...
+	if expectedCgroupMemoryLimit != actualCgroupMemoryLimit {
+		return fmt.Errorf("expected Cgroup memory limit %d, got %d", expectedCgroupMemoryLimit, actualCgroupMemoryLimit)
+	}
+	return nil
+}
+
+// compareCgroupCPULimit compares the CPU limit specified in a nodeSet with the limit set in the CPU control group at the OS level
+func compareCgroupCPULimit(topologyElement esv1.NodeSet, cgroupCPU struct {
+	CFSPeriodMicros int `json:"cfs_period_micros"`
+	CFSQuotaMicros  int `json:"cfs_quota_micros"`
```
barkbay, line 250: I don't think the tags are required here. I guess I understand why you don't want to use `nodeStats client.NodeStats` (as we do for memory), however my feeling is that it would make it easier to read:

```golang
func compareCgroupCPULimit(topologyElement esv1.NodeSet, nodeStats client.NodeStats) error {
	var expectedCPULimit *resource.Quantity
	for _, c := range topologyElement.PodTemplate.Spec.Containers {
		if c.Name == esv1.ElasticsearchContainerName {
			expectedCPULimit = c.Resources.Limits.Cpu()
		}
	}
	if expectedCPULimit == nil || expectedCPULimit.IsZero() {
		// no expected cpu, consider it's ok
		return nil
	}

	cgroupCPU := nodeStats.OS.CGroup.CPU
	actualCgroupCPULimit := float64(cgroupCPU.CFSQuotaMicros) / float64(cgroupCPU.CFSPeriodMicros)
	if expectedCPULimit.AsApproximateFloat64() != actualCgroupCPULimit {
		return fmt.Errorf("expected Cgroup CPU limit [%f], got [%f]", expectedCPULimit.AsApproximateFloat64(), actualCgroupCPULimit)
	}
	return nil
}
```

test/e2e/test/elasticsearch/checks_es.go
```diff
@@ -185,29 +171,173 @@ func (e *esClusterChecks) CheckESNodesTopology() test.Step {
...
+		}
 	}
+	if expectedCPULimit == nil || expectedCPULimit.IsZero() {
+		// no expected cpu, consider it's ok
+		return nil
+	}
+
+	actualCgroupCPULimit := float64(cgroupCPU.CFSQuotaMicros) / float64(cgroupCPU.CFSPeriodMicros)
+	if expectedCPULimit.AsApproximateFloat64() != actualCgroupCPULimit {
+		return fmt.Errorf("expected Cgroup CPU limit [%f], got [%f]", expectedCPULimit.AsApproximateFloat64(), actualCgroupCPULimit)
```
barkbay, line 265: nit: `cgroup` should not be capitalized
```suggestion
		return fmt.Errorf("expected cgroup CPU limit [%f], got [%f]", expectedCPULimit.AsApproximateFloat64(), actualCgroupCPULimit)
```

test/e2e/test/elasticsearch/checks_es.go
```diff
@@ -185,29 +171,173 @@ func (e *esClusterChecks) CheckESNodesTopology() test.Step {
...
+
+	actualCgroupMemoryLimit, err := strconv.ParseInt(
+		nodeStats.OS.CGroup.Memory.LimitInBytes, 10, 64,
+	)
+	if err != nil {
+		return err
+	}
+	expectedCgroupMemoryLimit := memoryLimit.Value()
+	if expectedCgroupMemoryLimit != actualCgroupMemoryLimit {
+		return fmt.Errorf("expected Cgroup memory limit %d, got %d", expectedCgroupMemoryLimit, actualCgroupMemoryLimit)
```
barkbay, line 242:
```suggestion
		return fmt.Errorf("expected cgroup memory limit %d, got %d", expectedCgroupMemoryLimit, actualCgroupMemoryLimit)
```

test/e2e/test/elasticsearch/checks_es.go
```diff
@@ -185,29 +171,173 @@ func (e *esClusterChecks) CheckESNodesTopology() test.Step {
...
+func compareQuantity(msg string, expected, actual *resource.Quantity) error {
+	if !expected.IsZero() && !equality.Semantic.DeepEqual(expected, actual) {
+		return fmt.Errorf("expected %s [%d], got [%d]", msg, expected.Value(), actual.Value())
+	}
+	return nil
+}
 
-	return expectedBytes == cgroupMemoryLimitsInBytes
+// compareClaimedStorage compares the requested storage specified in a nodeSet with the actual capacity claimed in the PVC
+func compareClaimedStorage(k8sClient *test.K8sClient, topologyElement esv1.NodeSet, pods []corev1.Pod) error {
```
barkbay, line 318: I'm wondering if it would not allow us to remove [`PVCs should eventually be resized`](https://github.com/elastic/cloud-on-k8s/blob/7e71c1bcc56d24bb19e623f3b3ac456b0d4f8724/test/e2e/es/volume_test.go#L198-L213). It is also fine to keep both.

test/e2e/test/elasticsearch/checks_es.go
```diff
@@ -185,29 +171,173 @@ func (e *esClusterChecks) CheckESNodesTopology() test.Step {
...
+	if err != nil {
+		return err
+	}
+	if len(pods) != len(pvcs) {
+		return fmt.Errorf("expected PVC count %q, got %q", len(pods), len(pvcs))
+	}
+	for _, pvc := range pvcs {
+		actualStorage := pvc.Spec.Resources.Requests.Storage()
+		if !equality.Semantic.DeepEqual(expectedStorage, actualStorage) {
+			return fmt.Errorf("expected claimed storage [%d], got [%d]", expectedStorage.Value(), actualStorage.Value())
```
barkbay, line 339: nit: maybe add some context about which PVC failed:
```suggestion
			return fmt.Errorf("%s: expected claimed storage [%d], got [%d]", pvc.Name, expectedStorage.Value(), actualStorage.Value())
```

test/e2e/test/elasticsearch/checks_es.go
```diff
@@ -185,29 +171,173 @@ func (e *esClusterChecks) CheckESNodesTopology() test.Step {
...
+					}
+				}
+			}
+		}
+	}
+	return nil
+}
 
-	expectedBytes := memoryLimit.Value()
+func compareQuantity(msg string, expected, actual *resource.Quantity) error {
```
barkbay, line 310: nit: I would surface the Pod name here in the error message.

test/e2e/test/elasticsearch/checks_es.go
```diff
@@ -185,29 +171,173 @@ func (e *esClusterChecks) CheckESNodesTopology() test.Step {
...
+			return fmt.Errorf("actual role %s not expected", r)
 		}
 	}
-
-	return true
+	return nil
 }
 
-func compareMemoryLimit(topologyElement esv1.NodeSet, cgroupMemoryLimitsInBytes int64) bool {
+// compareCgroupMemoryLimit compares the memory limit specified in a nodeSet with the limit set in the CPU control group at the OS level
```
barkbay, line 221:
```suggestion
// compareCgroupMemoryLimit compares the memory limit specified in a nodeSet with the limit set in the memory cgroup controller at the OS level
```

test/e2e/test/k8s_client.go
```diff
@@ -357,6 +376,15 @@ func ESPodListOptions(esNamespace, esName string) []k8sclient.ListOption {
 	return []k8sclient.ListOption{ns, matchLabels}
 }
 
+func ESPodListOptionsByNodeSet(esNamespace, esName, nodeSetName string) []k8sclient.ListOption {
```
barkbay, line 379: nit: maybe add a comment to explain the purpose of this function

barkbay (commented):
Almost LGTM ! Still want to do some additional testing.

test/e2e/test/elasticsearch/checks_es.go
thbkrkr: Yes, I hesitated on this point, thinking it is better to have too many checks than not enough, but it's redundant so now I feel it doesn't make sense to keep it and we have to trust `compareClaimedStorage`.

barkbay (approved):
LGTM !

thbkrkr:
run full pr build

PR #5024, opened 2021-11-08
E2E: adjust Enterprise search Kibana API path as of 7.16

pebrc:
https://github.com/elastic/kibana/pull/111035 has been back ported to 7.16 so we need to adjust our expectations in the e2e test.


barkbay (approved):
lgtm

PR #5029, opened 2021-11-09
Remove legacy v1beta1 CRDs

### LINKED ISSUE #4885: Remove support for legacy CRDs (k8s < 1.16)
We will be able to remove legacy CRDs support (v1beta1) for ECK 2.0.0, which no longer supports OCP3 (k8s 1.11), which was the last reason to continue providing legacy CRDs.

thbkrkr:
Remove legacy `v1beta1` CRDs, which removes the support of OpenShift 3.11 (4ca3965) and Kubernetes <1.16 (e8d15f1, ceb9890).

Resolves #4885.


pebrc (commented):
LGTM, the manifest generator has test cases for k8s 1.13 that we could probably remove as well wdyt?

thbkrkr:
> the manifest generator has test cases for k8s 1.13 that we could probably remove as well wdyt?

Yes I missed this part and I would even go further and remove all conditional code in the templates 3e6c79b, or do you see a reason to keep it?

hack/manifest-gen/templatize/.gitignore
```diff
@@ -1,2 +1 @@
 all-crds.yaml
```
pebrc, line 1: I think this whole directory is not needed anymore.

pebrc (approved):
LGTM!

pebrc:
One more thing we still have `config/all-in-one.yaml` file, I think this is a left-over from when I split up the manifests and I believe it should be safe to remove as well.

PR #5033, opened 2021-11-09
E2E monitor: move from IP to URL

pebrc:
This allows us to move the ES and APM endpoints behind Let's encrypt certificates managed by an Ingress controller. The new URL will therefore be at 443 so we need to remove the hard-coded port here. 

This also does a few other things to get better visibility into the recent test failures:

* upgrades the monitoring operator to 1.8
* bumps Filebeat resources to 300Mi (I will follow up with an issue to bump the default in the operator code)


barkbay (approved):
LGTM

PR #5035, opened 2021-11-10
E2E: metricbeat index names do not include pre-release indicator

pebrc:
To address https://devops-ci.elastic.co/job/cloud-on-k8s-e2e-tests-snapshot-versions-8x/294//testReport


thbkrkr (approved):
Good catch!

PR #5039, opened 2021-11-10
Move Docker Hub creds fetching

thbkrkr:
Move the Docker Hub credentials fetching to do it only in CI based on whether `BUILD_ID` is set at runtime in Jenkins, but not otherwise.


naemono (approved):
seems like it solves the issue with this vault secret not being available outside of Jenkins.

PR #5044, opened 2021-11-10
Revert "Add a replacement mechanism to work around nil handling in ucfg (#4041)"

### LINKED ISSUE #3718: Coordinator nodes cannot be defined using node.roles

thbkrkr:
Since elastic/go-ucfg#188 fixes #3718, we no longer need the replacement mechanism to work around nil handling in ucfg (#4041).
Also updates the unit test `TestCanonicalConfig_Render` to cover that the configuration with an empty array works.

Resolves #3718.


pebrc (approved):
LGTM, I tested with the old ucfg version for comparison purposes and your unit test flags up the error correctly 👍

PR #5045, opened 2021-11-11
E2E: switch full-pr build to master GKE pipeline

pebrc:
As it rarely makes sense to test a PR change against all Kubernetes variants we support, this switches the build pipeline to the one we use to test the master branch. 

This also turns off Slack notifications in the case such a build fails. 

I am not sure how to test this other than trying it in the real environment.


barkbay (approved):
LGTM

thbkrkr (approved):

PR #5046, opened 2021-11-11
E2E: move 8.0 workaround to mutation steps

pebrc:
https://github.com/elastic/cloud-on-k8s/pull/5041 was ineffective because I naively put the step right after the check for an Elasticsearch clusters existence. That check only checks for the existence in the Kubernetes API. At that point in time no Elasticsearch node is booted yet and the 30 second wait time is doing nothing to improve the success rate of the test.  This PR moves the step to the mutation steps which run any kind of update to the cluster which would be affected by the issue.


pebrc:
Jenkins test this please

barkbay (approved):

PR #5048, opened 2021-11-11
E2E: Adjust Kibana keystore test to be compatible with 8.0

pebrc:
The configuration settings we were using to test the Kibana keystore do no longer exist in 8.0. I have adjusted the test to use  configuration properties of the Elasticsearch connection instead which seemed to me the best bet for things that stayed the same across all versions we want to test ie. 6.8.x through 8.x. 🤞


barkbay (approved):
LGTM

Issue #3933, opened 2020-11-11
Simplify zone awareness configuration


eedugon:
## Proposal
At the moment the only way to configure nodes with different attributes for zone awareness is to create different nodeSets, knowing in advance the specific values for the attributes. So it's purely a static configuration, as we suggest in this document:
https://www.elastic.co/guide/en/cloud-on-k8s/1.2/k8s-advanced-node-scheduling.html#k8s-availability-zone-awareness

It would be great to have, within the same nodeSet, the possibility of having certain parameters retrieved from variables or pod metadata. For example to associate a `node.attr` setting to an environment variable available in the POD.

If that we already have (configuring settings based on pod's environment variables), then we would need to find a way for a running POD to know the availability zone where it's running :)

For that, apparently there's a feature request on Kubernetes side (https://github.com/kubernetes/kubernetes/issues/62078) to allow node labels to be passed to the pods. If that's implemented then probably this would be easier.

**Use case. Why is this important?**
- Some kubernetes environments are based on Autoscaling groups on multiple AZs with dynamic zones allocation. In such cases it would be great to provide the requested mechanisms.

- The `Elasticsearch` manifest of a 3 node cluster where each node has to run in a different AZ doesn't look very nice when we need to create 3 different nodeSets with `1 replica each` (same settings in each nodeSet except of the specific attribute for the AZ). With a single nodeSet (replica 3) + affinity rules the same goal could be achieved from scheduling point of view, but in such case there's no way for the 3 instances of the nodes to know where (in what AZ) they are running. If we are able to implement the requested mechanism then we could easily build this type of configuration with a single `nodeSet`, based on some extra dynamic parameters.

sebgl:
Also related (probably the right k8s issue to track): https://github.com/kubernetes/kubernetes/issues/40610
I see Anya already commented there :)

sebgl:
A very hacky unsupported way to achieve this, through an init container that retrieves the k8s node label and injects it in the Elasticsearch configuration:

```yaml
apiVersion: elasticsearch.k8s.elastic.co/v1
kind: Elasticsearch
metadata:
  name: elasticsearch-sample
spec:
  version: 7.9.2
  nodeSets:
  - name: default
    count: 1
    config:
      cluster.routing.allocation.awareness.attributes: zone
    podTemplate:
      spec:
        serviceAccountName: node-label-fetcher
        automountServiceAccountToken: true
        containers:
        - name: elasticsearch
        initContainers:
[...]

PR #5054, opened 2021-11-18
Allow K8S node labels to be propagated as pod annotations

barkbay:
Fixes https://github.com/elastic/cloud-on-k8s/issues/3933

## How to test (tested on GKE)

* Start the operator with the following flag: `--exposed-node-labels="topology.kubernetes.io/.*,failure-domain.beta.kubernetes.io/.*"`
* Annotate an Elasticsearch deployment: `k annotate --overwrite es/elasticsearch-sample   eck.k8s.elastic.co/downward-node-labels="topology.kubernetes.io/zone,topology.kubernetes.io/region"`
* Wait for the Pods to be restarted
* Check the annotations: `k get pod -l common.k8s.elastic.co/type=elasticsearch -o go-template='{{range .items}}{{.metadata.annotations}}{{printf "\n"}}{{end}}'`

To get the logs of the init script you can run `k logs  elasticsearch-sample-es-default-0 -c elastic-internal-init-filesystem  -f`

## Misc

* I eventually decided to reuse the `elastic-internal-init-filesystem` init container to wait for the annotations to be set.
* This PR should not restart the current `RUNNING` Pods. The `elastic-internal-init-filesystem` init-container should be left untouched as long as the feature is not used on an Elasticsearch resource.
* Pods are patched early in the reconciliation loop. I'm not sure if there's an ideal place (or a place we should avoid)
* A Pod is considered as scheduled when both its `Spec.NodeName` has a non-empty value and the relevant condition `PodScheduled` is set to `True` (see K8S binding code [here](https://github.com/kubernetes/kubernetes/blob/6c357f9996070fd015e08eec24f07c82b4b8b4a6/pkg/registry/core/pod/storage/storage.go#L217-L230))
* The primary intent for this PR is to propagate _topology_ node labels, hence node labels are considered as immutable and K8S `Nodes` are not watched.
* When an Elasticsearch resource is submitted the admission controller only check that the node label is allowed. Node labels are not checked until the Pod is scheduled. This is mostly because node labels may be inconsistent.
* I'll try to think about some e2e tests in a subsequent PR.

## To be addressed in follow up PRs

- [ ] Update documentation, including `docs/operating-eck/operating-eck.asciidoc`
- [ ] Add E2E Tests
- [ ] Enable telemetry

### DIFF
--- a/pkg/controller/elasticsearch/driver/driver.go
+++ b/pkg/controller/elasticsearch/driver/driver.go
@@ -151,6 +151,9 @@ func (d *defaultDriver) Reconcile(ctx context.Context) *reconciler.Results {
 		return results.WithError(err)
 	}
 
+	// Patch the Pods to add the expected node labels as annotations
+	results.WithResults(annotatePodsWithNodeLabels(ctx, d.Client, d.ES))
+
 	resourcesState, err := reconcile.NewResourcesStateFromAPI(d.Client, d.ES)
 	if err != nil {
 		return results.WithError(err)
--- a/pkg/controller/elasticsearch/driver/driver.go
+++ b/pkg/controller/elasticsearch/driver/driver.go
@@ -151,7 +151,8 @@ func (d *defaultDriver) Reconcile(ctx context.Context) *reconciler.Results {
 		return results.WithError(err)
 	}
 
-	// Patch the Pods to add the expected node labels as annotations
+	// Patch the Pods to add the expected node labels as annotations. Record the error, if any, but do not stop the
+	// reconciliation loop as we don't want to prevent other updates from being applied to the cluster.
 	results.WithResults(annotatePodsWithNodeLabels(ctx, d.Client, d.ES))
 
 	resourcesState, err := reconcile.NewResourcesStateFromAPI(d.Client, d.ES)
--- a/pkg/controller/elasticsearch/driver/node_labels_test.go
+++ b/pkg/controller/elasticsearch/driver/node_labels_test.go
@@ -138,6 +138,37 @@ func Test_annotatePodsWithNodeLabels(t *testing.T) {
 				},
 			},
 		},
+		{
+			name: "Retain existing annotations",
+			args: args{
+				es: &esv1.Elasticsearch{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:        esName,
+						Namespace:   "ns",
+						Annotations: map[string]string{"eck.k8s.elastic.co/downward-node-labels": "topology.kubernetes.io/region,topology.kubernetes.io/zone"},
+					},
+				},
+				objects: []runtime.Object{
+					newPodBuilder("elasticsearch-sample-es-default-0").scheduledOn("k8s-node-0").withAnnotation(map[string]string{"foo": "bar", "topology.kubernetes.io/region": "existing-annotation"}).build(),
+					newPodBuilder("elasticsearch-sample-es-default-1").scheduledOn("k8s-node-1").withAnnotation(map[string]string{"foo": "bar", "topology.kubernetes.io/region": "existing-annotation"}).build(),
+					&corev1.Node{ObjectMeta: metav1.ObjectMeta{Name: "k8s-node-0", Labels: map[string]string{"topology.kubernetes.io/region": "europe-west1", "topology.kubernetes.io/zone": "europe-west1-a"}}},
+					&corev1.Node{ObjectMeta: metav1.ObjectMeta{Name: "k8s-node-1", Labels: map[string]string{"topology.kubernetes.io/region": "europe-west1", "topology.kubernetes.io/zone": "europe-west1-b"}}},
+				},
+				ctx: context.Background(),
+			},
+			expectedAnnotations: expectedPodsAnnotations{
+				"elasticsearch-sample-es-default-0": {
+					"topology.kubernetes.io/region": "existing-annotation",
+					"topology.kubernetes.io/zone":   "europe-west1-a",
+					"foo":                           "bar",
+				},
+				"elasticsearch-sample-es-default-1": {
+					"topology.kubernetes.io/region": "existing-annotation",
+					"topology.kubernetes.io/zone":   "europe-west1-b",
+					"foo":                           "bar",
+				},
+			},
+		},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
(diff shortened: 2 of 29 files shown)


pkg/controller/elasticsearch/initcontainer/prepare_fs_script_test.go
```diff
@@ -34,6 +35,14 @@ func TestRenderScriptTemplate(t *testing.T) {
 				"ln -sf /secrets/users /usr/share/elasticsearch/users",
 			},
 		},
+		{
+			name: "With expected annotations",
+			params: TemplateParams{
+				PluginVolumes:       PluginVolumes,
+				ExpectedAnnotations: &expectedAnnotations,
+			},
+			wantSubstr: []string{},
```
pebrc, line 44: I think the expectation is missing here.

pkg/controller/elasticsearch/validation/node_labels_test.go
```diff
@@ -0,0 +1,72 @@
...
+	tests := []struct {
+		name              string
+		exposedNodeLabels []string
+		args              args
+		want              bool
+		wantErr           bool
+	}{
+		{
+			name:              "Cannot parse label",
+			exposedNodeLabels: []string{"*"},
```
pebrc, line 28: ".*" would match any label, should we test that as well? And maybe one test where the `exposedNodeLabels` do not contain a wildcard? I know that's borderline testing Go's regex engine but maybe for completeness?

pkg/controller/elasticsearch/nodespec/podspec.go
```diff
@@ -100,6 +103,11 @@ func BuildPodTemplateSpec(
 		return corev1.PodTemplateSpec{}, err
 	}
 
+	// inherit the node labels list to trigger a restart when changed
+	if es.HasNodeLabelsToPodsAnnotations() {
+		builder = builder.WithAnnotations(map[string]string{esv1.NodeLabelsToPodsAnnotationsAnnotation: es.Annotations[esv1.NodeLabelsToPodsAnnotationsAnnotation]})
+	}
```
pebrc, line 109: Could we save an annotation here by incorporating the value of `esv1.NodeLabelsToPodsAnnotationsAnnotation` into the `config-hash`?

pebrc (commented):
Looks really good. 👍 I did a few simple tests and it worked as advertised. I have only a few small comments nothing major. Naming suggestions are totally optional of course.

sebgl:
(applies to anything:) how do we decide whether a feature is specified as an annotation or as a field in the spec?
We went the annotation way here but I'm not sure about the reasoning. There is an "alpha" feel to the annotation, but I suspect we'd still have to commit to maintaining this working over time anyway.

From a user pov, I'm wondering if i'd rather want to do something like this, much higher-level:

```yml
spec:
  allocationAwareness:
    enable: true # defaults to false
    attribute: "zone" # default
    nodeLabel: "topology.kubernetes.io/zone" # default
```

and have the operator automatically handle the Elasticsearch configuration part + the downward API + the environment variable propagation?

barkbay:
> I suspect we'd still have to commit to maintaining this working over time anyway.

I think one of the reason is that I have the feeling to implement something which might be part of K8S at some point. I'm still hoping for an ["official" solution](https://github.com/kubernetes/kubernetes/issues/40610). It feels then easier to deprecate and remove an annotation than a field in the spec (assuming we I don't want to maintain a feature which is natively available).
It also allows us to get some feedback early, and adjust the feature if needed. This PR might then be a building block for a more advanced feature as the one outlined in your message.
Happy to reconsider if we think it deserves an update of the Elasticsearch schema.

pkg/controller/elasticsearch/driver/node_labels_test.go
```diff
@@ -0,0 +1,208 @@
...
+					&corev1.Node{ObjectMeta: metav1.ObjectMeta{Name: "k8s-node-0", Labels: map[string]string{"topology.kubernetes.io/region": "europe-west1", "topology.kubernetes.io/zone": "europe-west1-a"}}},
+					&corev1.Node{ObjectMeta: metav1.ObjectMeta{Name: "k8s-node-1", Labels: map[string]string{"topology.kubernetes.io/region": "europe-west1", "topology.kubernetes.io/zone": "europe-west1-b"}}},
+				},
+				ctx: context.Background(),
+			},
+			expectedAnnotations: expectedPodsAnnotations{
+				"elasticsearch-sample-es-default-0": {
+					"topology.kubernetes.io/region": "europe-west1",
+					"topology.kubernetes.io/zone":   "europe-west1-a",
+					"foo":                           "bar",
```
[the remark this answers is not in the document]
barkbay: > What if a "topology.kubernetes.io/region" or "zone" annotation already exists on the Pod? I believe strategic merge patch will retain the old value. Which is probably OK.

Yes, the idea is to not override an existing annotation. 👍 re. the unit test.

pkg/controller/elasticsearch/nodespec/podspec.go
barkbay: The operator sets only one annotation on Elasticsearch Pods for now (`co.elastic.logs/module`) _(and it makes me wondering why we are using labels to store configuration hashes 🤔 )_.
But TBH this was my original intent, however the purpose of `elasticsearch.k8s.elastic.co/config-hash` is to store a sum of the configuration of the Elasticsearch node, my feeling is that it might be misleading to store something else. And I think it's the reason why we still have this TODO: https://github.com/elastic/cloud-on-k8s/blob/83b517e24593adee2108714acf6a5f0ec661f3c5/pkg/controller/elasticsearch/nodespec/podspec.go#L148-L152

pebrc: I guess if we use a wide definition of `config` it still could make sense to include keystore version or the node labels here, but I don't feel strongly about it.

barkbay: I implemented the idea in https://github.com/elastic/cloud-on-k8s/commit/c3e4a4719adca2e17277f4bb2fa9b81fa6b13c63. I have reused the same 32-bit FNV-1 hash function to have a "kind of consistent" hash format.

pkg/controller/elasticsearch/driver/driver.go
```diff
@@ -151,6 +151,9 @@ func (d *defaultDriver) Reconcile(ctx context.Context) *reconciler.Results {
 		return results.WithError(err)
 	}
 
+	// Patch the Pods to add the expected node labels as annotations
+	results.WithResults(annotatePodsWithNodeLabels(ctx, d.Client, d.ES))
```
pebrc, line 155: Should we mention that we explicitly want to continue reconciliation here even in the presence of errors?

pebrc (approved):
LGTM, I think this will be really useful 🚀

PR #5057, opened 2021-11-18
E2E: adjust Kibana telemetry test to new format for 7.16+

pebrc:
Structure of JSON document returned by the internal Kibana API we use to verify static telemetry data is picked up correctly has changed since 7.16 see https://github.com/elastic/kibana/pull/113525

This adjusts the test to expect the new structure for the relevant versions.


david-kow (approved):
LGTM

PR #5058, opened 2021-11-18
Ignore events that are outside of namespaces that the controller manages

### LINKED ISSUE #4168: unable to get: x/y because of unknown namespace for the cache
I observed this error in our e2e tests, it seems to be harmless but it makes the logs harder to read and I think we should understand where it comes from:

```json
{
	"log.level": "error",
	"@timestamp": "2021-01-28T01:42:34.794Z",
	"log.logger": "manager.eck-operator.controller.beat-es-association-controller",
	"message": "Reconciler error",
	"service.version": "1.5.0-SNAPSHOT+7257dd51",
	"service.type": "eck",
	"ecs.version": "1.4.0",
	"name": "filebeat",
	"namespace": "e2e-xdhm8",
	"error": "unable to get: e2e-xdhm8/filebeat because of unknown namespace for the cache",
	"error.stack_trace": "sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.8.1/pkg/internal/controller/controller.go:252\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func1.2\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.8.1/pkg/internal/controller/controller.go:215\nk8s.io/apimachinery/pkg/util/wait.JitterUntilWithContext.func1\n\t/go/pkg/mod/k8s.io/apimachinery@v0.20.2/pkg/util/wait/wait.go:185\nk8s.io/apimachinery/pkg/util/wait.BackoffUntil.func1\n\t/go/pkg/mod/k8s.io/apimachinery@v0.20.2/pkg/util/wait/wait.go:155\nk8s.io/apimachinery/pkg/util/wait.BackoffUntil\n\t/go/pkg/mod/k8s.io/apimachinery@v0.20.2/pkg/util/wait/wait.go:156\nk8s.io/apimachinery/pkg/util/wait.JitterUntil\n\t/go/pkg/mod/k8s.io/apimachinery@v0.20.2/pkg/util/wait/wait.go:133\nk8s.io/apimachinery/pkg/util/wait.JitterUntilWithContext\n\t/go/pkg/mod/k8s.io/apimachinery@v0.20.2/pkg/util/wait/wait.go:185\nk8s.io/apimachinery/pkg/util/wait.UntilWithContext\n\t/go/pkg/mod/k8s.io/apimachinery@v0.20.2/pkg/util/wait/wait.go:99"
}
```

naemono:
This will ignore any CRDs laid down within a Kubernetes cluster that are outside of the namespaces that ECK is set to manage (`managedNamespaces` option).

resolves #4168

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -553,6 +553,7 @@ func startOperator(ctx context.Context) error {
 		Dialer:            dialer,
 		ExposedNodeLabels: exposedNodeLabels,
 		IPFamily:          ipFamily,
+		ManagedNamespaces: managedNamespaces,
 		OperatorNamespace: operatorNamespace,
 		OperatorInfo:      operatorInfo,
 		CACertRotation: certificates.RotationParams{
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -812,7 +812,7 @@ func setupWebhook(mgr manager.Manager, params operator.Parameters, clientset kub
 			os.Exit(1)
 		}
 
-		if err := webhook.Add(mgr, webhookParams, clientset, wh, predicates.ManagedNamespacePredicate); err != nil {
+		if err := webhook.Add(mgr, webhookParams, clientset, wh, params); err != nil {
 			log.Error(err, "unable to create controller", "controller", webhook.ControllerName)
 			os.Exit(1)
 		}
@@ -842,7 +842,7 @@ func setupWebhook(mgr manager.Manager, params operator.Parameters, clientset kub
 	}
 
 	// esv1 validating webhook is wired up differently, in order to access the k8s client
-    esvalidation.RegisterWebhook(mgr, params.ValidateStorageClass, params.ExposedNodeLabels)
+	esvalidation.RegisterWebhook(mgr, params.ValidateStorageClass, params.ExposedNodeLabels)
 
 	// wait for the secret to be populated in the local filesystem before returning
 	interval := time.Second * 1
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -58,7 +58,6 @@ import (
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common/container"
 	commonlicense "github.com/elastic/cloud-on-k8s/pkg/controller/common/license"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common/operator"
-	"github.com/elastic/cloud-on-k8s/pkg/controller/common/predicates"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common/reconciler"
 	controllerscheme "github.com/elastic/cloud-on-k8s/pkg/controller/common/scheme"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common/tracing"
@@ -467,8 +466,6 @@ func startOperator(ctx context.Context) error {
 
 	// configure the manager cache based on the number of managed namespaces
 	managedNamespaces := viper.GetStringSlice(operator.NamespacesFlag)
-	// initialize the managed namespace predicate to ignore events outside of the namespaces the operator is concerned with
-	predicates.ManagedNamespacePredicate = predicates.NewManagedNamespacesPredicate(managedNamespaces)
 
 	switch {
 	case len(managedNamespaces) == 0:
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -667,7 +667,7 @@ func asyncTasks(
 	// - association user secrets
 	garbageCollectUsers(cfg, managedNamespaces)
 	// - soft-owned secrets
-	garbageCollectSoftOwnedSecrets(mgr.GetClient())
+	garbageCollectSoftOwnedSecrets(mgr.GetClient(), managedNamespaces)
 }
 
 func chooseAndValidateIPFamily(ipFamilyStr string, ipFamilyDefault corev1.IPFamily) (corev1.IPFamily, error) {
@@ -768,7 +768,7 @@ func garbageCollectUsers(cfg *rest.Config, managedNamespaces []string) {
 	}
 }
 
-func garbageCollectSoftOwnedSecrets(k8sClient k8s.Client) {
+func garbageCollectSoftOwnedSecrets(k8sClient k8s.Client, managedNamespaces []string) {
 	if err := reconciler.GarbageCollectAllSoftOwnedOrphanSecrets(k8sClient, map[string]client.Object{
 		esv1.Kind:          &esv1.Elasticsearch{},
 		apmv1.Kind:         &apmv1.ApmServer{},
@@ -777,7 +777,7 @@ func garbageCollectSoftOwnedSecrets(k8sClient k8s.Client) {
 		beatv1beta1.Kind:   &beatv1beta1.Beat{},
 		agentv1alpha1.Kind: &agentv1alpha1.Agent{},
 		emsv1alpha1.Kind:   &emsv1alpha1.ElasticMapsServer{},
-	}); err != nil {
+	}, managedNamespaces); err != nil {
 		log.Error(err, "Orphan secrets garbage collection failed, will be attempted again at next operator restart.")
 		return
 	}
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -58,6 +58,7 @@ import (
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common/container"
 	commonlicense "github.com/elastic/cloud-on-k8s/pkg/controller/common/license"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common/operator"
+	"github.com/elastic/cloud-on-k8s/pkg/controller/common/predicates"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common/reconciler"
 	controllerscheme "github.com/elastic/cloud-on-k8s/pkg/controller/common/scheme"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common/tracing"
@@ -571,7 +572,7 @@ func startOperator(ctx context.Context) error {
 	}
 
 	if viper.GetBool(operator.EnableWebhookFlag) {
-		setupWebhook(mgr, params.CertRotation, params.ValidateStorageClass, clientset, exposedNodeLabels)
+		setupWebhook(mgr, params, clientset)
 	}
 
 	enforceRbacOnRefs := viper.GetBool(operator.EnforceRBACOnRefsFlag)
@@ -783,12 +784,7 @@ func garbageCollectSoftOwnedSecrets(k8sClient k8s.Client) {
 	log.Info("Orphan secrets garbage collection complete")
 }
 
-func setupWebhook(
-	mgr manager.Manager,
-	certRotation certificates.RotationParams,
-	validateStorageClass bool,
-	clientset kubernetes.Interface,
-	exposedNodeLabels esvalidation.NodeLabels) {
+func setupWebhook(mgr manager.Manager, params operator.Parameters, clientset kubernetes.Interface) {
 	manageWebhookCerts := viper.GetBool(operator.ManageWebhookCertsFlag)
 	if manageWebhookCerts {
 		log.Info("Automatic management of the webhook certificates enabled")
@@ -797,7 +793,7 @@ func setupWebhook(
 			Name:       viper.GetString(operator.WebhookNameFlag),
 			Namespace:  viper.GetString(operator.OperatorNamespaceFlag),
 			SecretName: viper.GetString(operator.WebhookSecretFlag),
-			Rotation:   certRotation,
+			Rotation:   params.CertRotation,
 		}
 
 		// retrieve the current webhook configuration interface
@@ -813,7 +809,7 @@ func setupWebhook(
 			os.Exit(1)
 		}
 
-		if err := webhook.Add(mgr, webhookParams, clientset, wh); err != nil {
+		if err := webhook.Add(mgr, webhookParams, clientset, wh, predicates.NewManagedNamespacesPredicate(params.ManagedNamespaces)); err != nil {
 			log.Error(err, "unable to create controller", "controller", webhook.ControllerName)
 			os.Exit(1)
 		}
@@ -843,7 +839,7 @@ func setupWebhook(
 	}
 
 	// esv1 validating webhook is wired up differently, in order to access the k8s client
-	esvalidation.RegisterWebhook(mgr, validateStorageClass, exposedNodeLabels)
+    esvalidation.RegisterWebhook(mgr, params.ValidateStorageClass, params.ExposedNodeLabels)
 
 	// wait for the secret to be populated in the local filesystem before returning
 	interval := time.Second * 1
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -467,6 +467,9 @@ func startOperator(ctx context.Context) error {
 
 	// configure the manager cache based on the number of managed namespaces
 	managedNamespaces := viper.GetStringSlice(operator.NamespacesFlag)
+	// initialize the managed namespace predicate to ignore events outside of the namespaces the operator is concerned with
+	predicates.ManagedNamespacePredicate = predicates.NewManagedNamespacesPredicate(managedNamespaces)
+
 	switch {
 	case len(managedNamespaces) == 0:
 		log.Info("Operator configured to manage all namespaces")
@@ -809,7 +812,7 @@ func setupWebhook(mgr manager.Manager, params operator.Parameters, clientset kub
 			os.Exit(1)
 		}
 
-		if err := webhook.Add(mgr, webhookParams, clientset, wh, predicates.NewManagedNamespacesPredicate(params.ManagedNamespaces)); err != nil {
+		if err := webhook.Add(mgr, webhookParams, clientset, wh, predicates.ManagedNamespacePredicate); err != nil {
 			log.Error(err, "unable to create controller", "controller", webhook.ControllerName)
 			os.Exit(1)
 		}
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -479,12 +479,12 @@ func startOperator(ctx context.Context) error {
 		// The managed cache should always include the operator namespace so that we can work with operator-internal resources.
 		managedNamespaces = append(managedNamespaces, operatorNamespace)
 
+		opts.NewCache = cache.MultiNamespacedCacheBuilder(managedNamespaces)
+
 		// Add the empty namespace to allow watching cluster-scoped resources if storage class validation is enabled.
 		if viper.GetBool(operator.ValidateStorageClassFlag) {
-			managedNamespaces = append(managedNamespaces, "")
+			opts.NewCache = cache.MultiNamespacedCacheBuilder(append(managedNamespaces, ""))
 		}
-
-		opts.NewCache = cache.MultiNamespacedCacheBuilder(managedNamespaces)
 	}
 
 	// only expose prometheus metrics if provided a non-zero port
--- a/pkg/controller/common/predicates/predicates.go
+++ b/pkg/controller/common/predicates/predicates.go
@@ -45,5 +45,8 @@ func ManagedNamespacesPredicate(managedNamespaces []string) predicate.Predicate 
 			// Ignore resources that do not exist within the managed namespaces
 			return namespaceInSlice(e.Object.GetNamespace(), managedNamespaces)
 		},
+		GenericFunc: func(e event.GenericEvent) bool {
+			return namespaceInSlice(e.Object.GetNamespace(), managedNamespaces)
+		},
 	}
 }
--- a/pkg/controller/common/predicates/predicates.go
+++ b/pkg/controller/common/predicates/predicates.go
@@ -5,48 +5,44 @@
 package predicates
 
 import (
+	"k8s.io/utils/strings/slices"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 )
 
-func namespaceInSlice(namespace string, namespaces []string) bool {
-	// If the operator is managing all namespaces,
-	// never ignore any namespace.
-	if len(namespaces) == 0 {
-		return true
+// ManagedNamespacePredicate is a predicate that will ignore events that exist
+// outside of the namespaces that the operator manages.
+var ManagedNamespacePredicate predicate.Predicate
+
+// NewManagedNamespacesPredicate will return a predicate that will ignore events
+// that exist outside of the given managed namespaces,
+func NewManagedNamespacesPredicate(managedNamespaces []string) predicate.Predicate {
+	return predicate.Funcs{
+		CreateFunc: func(e event.CreateEvent) bool {
+			// Ignore resources that do not exist within the managed namespaces
+			return IsNamespaceManaged(e.Object.GetNamespace(), managedNamespaces)
+		},
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			// Ignore resources that do not exist within the managed namespaces
+			return IsNamespaceManaged(e.ObjectNew.GetNamespace(), managedNamespaces)
+		},
+		DeleteFunc: func(e event.DeleteEvent) bool {
+			// Ignore resources that do not exist within the managed namespaces
+			return IsNamespaceManaged(e.Object.GetNamespace(), managedNamespaces)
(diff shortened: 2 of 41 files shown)


naemono:
This PR takes the most minimal approach to solving the issue referenced in #4168, which references the issue where our custom CRDs themselves, from namespace `a`, are ignored by the operator which is managing a set of namespaces `b,c,d`.    I'm currently testing whether we'd also need to include this/these predicates in other watches, for objects owned by these parent objects, such as https://github.com/elastic/cloud-on-k8s/blob/master/pkg/controller/agent/controller.go#L68.  Any early 👀  on this would be welcomed.

naemono:
Yeah, my testing is showing that these will need to be ignored as well.  That change is a WIP.

naemono:
Ok, changes complete.  I'm not seeing these error messages any longer when running 2 operators within the same cluster, managing 2x separate namespaces:

operator 1
```
❯ helm get values -n elastic-system elastic-operator
USER-SUPPLIED VALUES:
managedNamespaces:
- default
```

operator 2
```
❯ helm get values -n es-system eck-operator
USER-SUPPLIED VALUES:
fullnameOverride: eck-operator
installCRDs: false
managedNamespaces:
- elastic
nameOverride: eck-operator
```

With the following resources

```
❯ kc get es,kibana,agent -n default
NAME                                                 HEALTH    NODES   VERSION   PHASE             AGE
elasticsearch.elasticsearch.k8s.elastic.co/testing   unknown   1       7.15.1    ApplyingChanges   5d20h

NAME                                          HEALTH   NODES   VERSION   AGE
kibana.kibana.k8s.elastic.co/kibana-testing   green    1       7.15.0    5d20h
```

pkg/controller/common/predicates/predicates.go
```diff
@@ -0,0 +1,49 @@
...
+			return namespaceInSlice(e.Object.GetNamespace(), managedNamespaces)
+		},
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			// Ignore resources that do not exist within the managed namespaces
+			return namespaceInSlice(e.ObjectNew.GetNamespace(), managedNamespaces)
+		},
+		DeleteFunc: func(e event.DeleteEvent) bool {
+			// Ignore resources that do not exist within the managed namespaces
+			return namespaceInSlice(e.Object.GetNamespace(), managedNamespaces)
+		},
```
barkbay, line 47: Should we also handle `GenericFunc` ?
```suggestion
		},
		GenericFunc: func(e event.GenericEvent) bool {
			return namespaceInSlice(e.Object.GetNamespace(), managedNamespaces)
		},
```

pkg/controller/common/predicates/predicates.go
```diff
@@ -0,0 +1,49 @@
...
+	if len(namespaces) == 0 {
+		return true
+	}
+	for _, ns := range namespaces {
+		if namespace == ns {
+			return true
+		}
+	}
+	return false
+}
```
barkbay, line 24:
```suggestion
// IsNamespaceManaged returns true if the namespace is managed by the operator.
func IsNamespaceManaged(namespace string, managedNamespaces []string) bool {
	return len(managedNamespaces) == 0 || slices.Contains(managedNamespaces, namespace)
}
```

pkg/controller/common/reconciler/secret.go
```diff
@@ -170,17 +172,41 @@ func GarbageCollectSoftOwnedSecrets(c k8s.Client, deletedOwner types.NamespacedN
...
-// GarbageCollectAllSoftOwnedOrphanSecrets iterates over all Secrets that reference a soft owner. If the owner
+// GarbageCollectAllSoftOwnedOrphanSecrets iterates over all Secrets in the namespaces that the operator
+// manages that reference a soft owner. If the owner
 // doesn't exist anymore, it deletes the secrets.
 // Should be called on operator startup, after cache warm-up, to cover cases where
 // the operator is down when the owner is deleted.
 // If the operator is up, garbage collection is already handled by GarbageCollectSoftOwnedSecrets on owner deletion.
-func GarbageCollectAllSoftOwnedOrphanSecrets(c k8s.Client, ownerKinds map[string]client.Object) error {
+func GarbageCollectAllSoftOwnedOrphanSecrets(c k8s.Client, ownerKinds map[string]client.Object, managedNamespaces []string) error {
+	if len(managedNamespaces) == 0 {
```
barkbay, line 182: Would it be possible to get the same behavior by adding a predicate while iterating over the SecretList:

```suggestion
	// remove any secret whose owner in the same namespace doesn't exist
	for i := range secrets.Items {
		secret := secrets.Items[i]
		if !predicates.IsNamespaceManaged(secret.Namespace, managedNamespaces) {
			continue
		}
```

barkbay (commented):
I'm wondering if we could not assume that there is only one predicate, and also that this predicate is immutable. Based on this assumption we would have the predicate initialized in only one place, and shared as a public variable.

For example, the predicate can be managed in a `predicate` package:

```golang
package predicate

import (
	"sigs.k8s.io/controller-runtime/pkg/event"
	"sigs.k8s.io/controller-runtime/pkg/predicate"
)

var ManagedNamespacePredicate predicate.Predicate

// NewManagedNamespacesPredicate will return a predicate that will ignore events
// that exist outside of the given managed namespaces,
func NewManagedNamespacesPredicate(managedNamespaces []string) predicate.Predicate {
	/* logic to create the predicate*/
}
```

The predicate would be then initialized in `main.go` :
```golang
// configure the manager cache based on the number of managed namespaces
managedNamespaces := viper.GetStringSlice(operator.NamespacesFlag)
predicate.ManagedNamespacePredicate = predicate.NewManagedNamespacesPredicate(managedNamespaces)
switch {
   case len(managedNamespaces) == 0:
....
```

The predicate can now be used from any controller, without the need of adding a new function parameter, or a new member in a struct.
For example:

```golang
// addWatches adds watches for all resources this controller cares about
func addWatches(c controller.Controller, r *ReconcileBeat) error {
	// Watch for changes to Beat
	if err := c.Watch(&source.Kind{Type: &beatv1beta1.Beat{}}, &handler.EnqueueRequestForObject{}, predicate.ManagedNamespacePredicate); err != nil {
		return err
	}

	// Watch DaemonSets
	if err := c.Watch(&source.Kind{Type: &appsv1.DaemonSet{}}, &handler.EnqueueRequestForOwner{
		IsController: true,
		OwnerType:    &beatv1beta1.Beat{},
	}, predicate.ManagedNamespacePredicate); err != nil {
		return err
	}

	// Watch Deployments
	if err := c.Watch(&source.Kind{Type: &appsv1.Deployment{}}, &handler.EnqueueRequestForOwner{
		IsController: true,
		OwnerType:    &beatv1beta1.Beat{},
	}, predicate.ManagedNamespacePredicate); err != nil {
		return err
	}

	// Watch Pods, to ensure `status.version` is correctly reconciled on any change.
	// Watching Deployments or DaemonSets only may lead to missing some events.
	if err := watches.WatchPods(c, beatcommon.NameLabelName, predicate.ManagedNamespacePredicate); err != nil {
		return err
	}

	// Watch owned and soft-owned Secrets
	if err := c.Watch(&source.Kind{Type: &corev1.Secret{}}, &handler.EnqueueRequestForOwner{
		IsController: true,
		OwnerType:    &beatv1beta1.Beat{},
	}, predicates.ManagedNamespacePredicate); err != nil {
		return err
	}
	if err := watches.WatchSoftOwnedSecrets(c, beatv1beta1.Kind, predicates.ManagedNamespacePredicate); err != nil {
		return err
	}

	// Watch dynamically referenced Secrets
	return c.Watch(&source.Kind{Type: &corev1.Secret{}}, r.dynamicWatches.Secrets, predicates.ManagedNamespacePredicate)
}
```

pkg/controller/common/reconciler/secret.go
naemono: I'm not following you here.  GarbageCollectAllSoftOwnedOrphanSecrets is called asynchronously a single time at startup, and doesn't have a `corev1.SecretList` at this time.  Are you referring to leveraging `predicates.IsNamespaceManaged` in this func?
https://github.com/elastic/cloud-on-k8s/blob/0dbe3b868cbee9cbcdb854ccae6ac688fdad6917/pkg/controller/common/reconciler/secret.go#L178

naemono:
@barkbay Yeah, those suggested changes would make things cleaner.  Thanks for the suggestion.  I'll get those changes taken care of.

naemono:
I'm investigating why integration tests are failing in `github.com/elastic/cloud-on-k8s/pkg/controller/license`, otherwise it's ready for 👀  again.

pkg/controller/common/reconciler/secret.go
barkbay: Sorry, just realized that I was not clear.
Yes, I was wondering if we could keep the existing `GarbageCollectAllSoftOwnedOrphanSecrets` and reuse the predicate. I think it would avoid to introduce `garbageCollectSecrets` (no strong opinion though as it is only a cosmetic change).

naemono: Ok, I finally understood what you were suggesting, and since I thought it was this call which was causing the namespace error

`if err := c.List(context.Background()`

upon doing a little more debugging, I noticed it was this:

`err := c.Get(context.Background(), types.NamespacedName{Namespace: softOwner.Namespace, Name: softOwner.Name}, owner)`

and it all made sense.  This greatly simplifies these changes in this package.  Thanks @barkbay

naemono:
> I'm investigating why integration tests are failing in `github.com/elastic/cloud-on-k8s/pkg/controller/license`, otherwise it's ready for 👀 again.

Resolved.

barkbay (commented):
While doing a new review I realized that the predicate is missing in the [trial controller](https://github.com/elastic/cloud-on-k8s/blob/5e5bbdac30dc1c81e2f6621cda76402a891011f2/pkg/controller/license/trial/trial_controller.go#L241).

This made me think that we should maybe try another approach, in which we don't have to pass the predicate as a parameter in all the `Watch(...)` functions. I remembered that most of the controllers are created using a custom function, `common.NewController`. For example [in the case of Elasticsearch](https://github.com/elastic/cloud-on-k8s/blob/5e5bbdac30dc1c81e2f6621cda76402a891011f2/pkg/controller/elasticsearch/elasticsearch_controller.go#L54-L64):

```golang
// Add creates a new Elasticsearch Controller and adds it to the Manager with default RBAC. The Manager will set fields
// on the Controller and Start it when the Manager is Started.
// this is also called by cmd/main.go
func Add(mgr manager.Manager, params operator.Parameters) error {
	reconciler := newReconciler(mgr, params)
	c, err := common.NewController(mgr, name, reconciler, params)
	if err != nil {
		return err
	}
	return addWatches(c, reconciler)
}
```
_(There is the exception of the [webhook certificate controller ](https://github.com/elastic/cloud-on-k8s/blob/5e5bbdac30dc1c81e2f6621cda76402a891011f2/pkg/controller/webhook/webhook_certificates_controller.go#L98), but maybe we can also use `common.NewController(...)` here ?)_

I think an alternative would be to inject the predicate in a new implementation of the `controller.Controller` interface, something along those lines in `pkg/controller/common/controller.go`:

```golang
// NewController creates a new controller with the given name, reconciler and parameters and registers it with the manager.
func NewController(mgr manager.Manager, name string, r reconcile.Reconciler, p operator.Parameters) (controller.Controller, error) {
	c, err := controller.New(name, mgr, controller.Options{Reconciler: r, MaxConcurrentReconciles: p.MaxConcurrentReconciles})
	if err != nil {
		return nil, err
	}
	return newNamespaceAwareWatchersController(c, p.ManagedNamespaces, p.OperatorNamespace), nil
}

var _ controller.Controller = &namespaceAwareController{}

// namespaceAwareController implements the controller.Controller interface and automatically include a predicate to filter events
// which are not in a managed namespace.
type namespaceAwareController struct {
	controller.Controller
	namespacePredicate predicate.Predicate
}

func newNamespaceAwareWatchersController(c controller.Controller, managedNamespaces []string, operatorNamespace string) controller.Controller {
	watchedNamespaces := managedNamespaces
	if !slices.Contains(managedNamespaces, operatorNamespace) {
		watchedNamespaces = append(watchedNamespaces, operatorNamespace)
	}
	return &namespaceAwareController{
		Controller: c,
		namespacePredicate: predicate.Funcs{
			CreateFunc: func(e event.CreateEvent) bool {
				return IsNamespaceManaged(e.Object.GetNamespace(), managedNamespaces)
			},
			UpdateFunc: func(e event.UpdateEvent) bool {
				return IsNamespaceManaged(e.ObjectNew.GetNamespace(), managedNamespaces)
			},
			DeleteFunc: func(e event.DeleteEvent) bool {
				return IsNamespaceManaged(e.Object.GetNamespace(), managedNamespaces)
			},
			GenericFunc: func(e event.GenericEvent) bool {
				return IsNamespaceManaged(e.Object.GetNamespace(), managedNamespaces)
			},
		},
	}
}

func (n *namespaceAwareController) Watch(src source.Source, eventhandler handler.EventHandler, predicates ...predicate.Predicate) error {
	return n.Controller.Watch(src, eventhandler, append(predicates, n.namespacePredicate)...)
}
```

Sorry for all those back and forth, I think that if this approach works it'll make this PR a bit smaller and reduce the chance of a missing a controller.

pkg/controller/webhook/webhook_certificates_controller.go
```diff
@@ -92,10 +92,10 @@ func newReconciler(mgr manager.Manager, webhookParams Params, clientset kubernet
 }
 
 // Add adds a new Controller to mgr with r as the reconcile.Reconciler
-func Add(mgr manager.Manager, webhookParams Params, clientset kubernetes.Interface, webhook AdmissionControllerInterface) error {
+func Add(mgr manager.Manager, webhookParams Params, clientset kubernetes.Interface, webhook AdmissionControllerInterface, operatorParams operator.Parameters) error {
 	r := newReconciler(mgr, webhookParams, clientset)
 	// Create a new controller
-	c, err := controller.New(ControllerName, mgr, controller.Options{Reconciler: r})
```
naemono, line 98: This is the primary thing I'm concerned about in these new changes.  Was there a specific reason why the webhook controller wasn't using the common `NewController` func?

cmd/manager/main.go
```diff
@@ -466,6 +467,9 @@ func startOperator(ctx context.Context) error {
 
 	// configure the manager cache based on the number of managed namespaces
 	managedNamespaces := viper.GetStringSlice(operator.NamespacesFlag)
+	// initialize the managed namespace predicate to ignore events outside of the namespaces the operator is concerned with
+	predicates.ManagedNamespacePredicate = predicates.NewManagedNamespacesPredicate(managedNamespaces)
```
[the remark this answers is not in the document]
naemono: Thanks @barkbay for all of the suggestions in this PR.  Don't worry about the back and forth at all.  The last suggestion significantly cleaned up the changes.  👍 As for this comment, the operator namespace is considered in the common controller `newNamespaceAwareWatchersController` func now, and this code is simply gone.

pkg/controller/webhook/webhook_certificates_controller.go
barkbay: I _think_ it's because we don't want/need to tune `MaxConcurrentReconciles` in the context of the webhook certificates.

Maybe we could introduce a function that accepts a predefined set of `controller.Options` to get the same behaviour: 

```golang
// NewController creates a new controller with the given name, reconciler and parameters and registers it with the manager.
func NewController(mgr manager.Manager, name string, r reconcile.Reconciler, p operator.Parameters) (controller.Controller, error) {
	options := controller.Options{Reconciler: r, MaxConcurrentReconciles: p.MaxConcurrentReconciles}
	return NewControllerWithOptions(mgr, name, p, options)
}

// NewControllerWithOptions creates a new controller with the given name, reconciler, parameters, options and registers it with the manager.
func NewControllerWithOptions(mgr manager.Manager, name string, p operator.Parameters, options controller.Options) (controller.Controller, error) {
	c, err := controller.New(name, mgr, options)
	if err != nil {
		return nil, err
	}
	return newNamespaceAwareWatchersController(c, p.ManagedNamespaces, p.OperatorNamespace), nil
}
``` 

And then use it here (`webhook_certificates_controller.go`):

```golang
// Add adds a new Controller to mgr with r as the reconcile.Reconciler
func Add(mgr manager.Manager, webhookParams Params, clientset kubernetes.Interface, webhook AdmissionControllerInterface, operatorParams operator.Parameters) error {
	r := newReconciler(mgr, webhookParams, clientset)
	// Create a new controller
	c, err := common.NewControllerWithOptions(mgr, ControllerName, operatorParams, controller.Options{Reconciler: r})
	if err != nil {
		return err
	}

	secret := types.NamespacedName{
		Namespace: webhookParams.Namespace,
		Name:      webhookParams.SecretName,
	}
...
```

naemono: this seems sensible now that you've noted the potential reasons, and I've implemented this.

barkbay (commented):
After doing some additional tests it turns out that we also need to update `telemetry.go`. First error is raised when reading Kibana secrets here: https://github.com/elastic/cloud-on-k8s/blob/7122d8db35382772eb59298a82f7b9f43f470259/pkg/telemetry/telemetry.go#L150-L153

```json
{
	"log.level": "error",
	"@timestamp": "2021-12-09T08:49:49.727Z",
	"log.logger": "usage",
	"message": "failed to get Kibana secret",
	"service.version": "2.0.0-SNAPSHOT+a25633ec",
	"service.type": "eck",
	"ecs.version": "1.4.0",
	"error": "unable to get: namespace-a/kb-apm-sample-kb-config because of unknown namespace for the cache",
	"error.stack_trace": "github.com/elastic/cloud-on-k8s/cmd/manager.asyncTasks.func2\n\t/go/src/github.com/elastic/cloud-on-k8s/cmd/manager/main.go:655"
}
```

An option might be to use once again the predicate as we do for `GarbageCollectAllSoftOwnedOrphanSecrets`. However I fear we'll have the same problem in all the `getStatsFn` functions (`esStats`, `kbStats`...). I'm then wondering if a better solution would not be to reuse the original flag, as provided by the user, when the reporter is created [here](https://github.com/elastic/cloud-on-k8s/blob/7122d8db35382772eb59298a82f7b9f43f470259/cmd/manager/main.go#L656-L662):

```golang
	if !disableTelemetry {
		// Start the telemetry reporter
		go func() {
			tr := telemetry.NewReporter(operatorInfo, mgr.GetClient(), operatorNamespace, viper.GetStringSlice(operator.NamespacesFlag), telemetryInterval)
			tr.Start()
		}()
	}
```

Overall I think we have a problem with the `managedNamespaces`, it's confusing to have this empty string. Maybe we should just make it immutable:

```golang
		// Add the empty namespace to allow watching cluster-scoped resources if storage class validation is enabled.
		if viper.GetBool(operator.ValidateStorageClassFlag) {
			opts.NewCache = cache.MultiNamespacedCacheBuilder(append(managedNamespaces, ""))
		} else {
			opts.NewCache = cache.MultiNamespacedCacheBuilder(managedNamespaces)
		}
```

naemono:
@barkbay may I ask how you generated this error?  

`GarbageCollectAllSoftOwnedOrphanSecrets` has this potential issue, as it's doing a non-namespace scoped list operation, and then iterating over the list, and running a `.Get`, which is where the error originates:

https://github.com/elastic/cloud-on-k8s/blob/7122d8db35382772eb59298a82f7b9f43f470259/pkg/controller/common/reconciler/secret.go#L181-L184

but telemetry is doing a namespace scoped list operation, and then iterating over it, which should only get Kibanas in it's set of managed namespaces, right?

https://github.com/elastic/cloud-on-k8s/blob/7122d8db35382772eb59298a82f7b9f43f470259/pkg/telemetry/telemetry.go#L143

> reuse the original flag

To which flag are you referring?  Surely not `disableTelemetry`, as this code path isn't even traversed when telemetry is disabled.  As for the managedNamespaces/operator.NamespacesFlag, that is passed along to the `NewReporter` func.

> Overall I think we have a problem with the managedNamespaces, it's confusing to have this empty string

I agree that it's confusing, but isn't that what the apimachinery uses to designate all namespaces:

https://github.com/kubernetes/apimachinery/blob/15885e270e2cefaaff9aa2b7b05eb004192bb5db/pkg/apis/meta/v1/types.go#L294
```
	// NamespaceAll is the default argument to specify on a context when you want to list or filter resources across all namespaces
	NamespaceAll = ""
```

Do you see a path to making this less confusing in this code?

barkbay:
> @barkbay may I ask how you generated this error?

I have a "restricted" operator running with `--telemetry-interval=5s` (default interval is 1 hour which might explain why it is easy to miss it)

> but telemetry is doing a namespace scoped list operation, and then iterating over it, which should only get Kibanas in it's set of managed namespaces, right?

No because `r.managedNamespaces` may contain the empty string `""`. The operator first iterates over the "real" managed namespaces, but eventually it will try to list `Kibana` instances across all of them:

image
 
> To which flag are you referring? [...] As for the managedNamespaces/operator.NamespacesFlag, that is passed along to the `NewReporter` func.

Yes, I was referring to `viper.GetStringSlice(operator.NamespacesFlag)`. What is passed to `NewReporter` is actually `append(viper.GetStringSlice(operator.NamespacesFlag), "")`: it is the original flag and the empty string `""`. It is not the original value of `viper.GetStringSlice(operator.NamespacesFlag)`, and this is the root of this issue.

> I agree that it's confusing, but isn't that what the apimachinery uses to designate all namespaces:

I don't think we want to watch all the namespaces. We want to watch `StorageClass` resources, which are cluster-scoped resources. Unfortunately it seems that the only way to watch cluster-scoped resources is to add `""` to the managed namespace. Watching all the namespaces is an unfortunate side-effect.

> Do you see a path to making this less confusing in this code?

I was wondering if adding `NamespaceAll` only when the manager is created would help:

```golang
// Add the empty namespace to allow watching cluster-scoped resources if storage class validation is enabled.
if viper.GetBool(operator.ValidateStorageClassFlag) {
   opts.NewCache = cache.MultiNamespacedCacheBuilder(append(managedNamespaces, ""))
} else {
   opts.NewCache = cache.MultiNamespacedCacheBuilder(managedNamespaces)
}
```

naemono:
My apologies.  I didn't realize all your initial comments were coming from the fact that we do this:

```
		if viper.GetBool(operator.ValidateStorageClassFlag) {
			managedNamespaces = append(managedNamespaces, "")
		}
```

now it all makes sense. 

I completely agree that it's quite confusing to have the "all namespaces" namespace in the list of managed namespaces.  I'll get these changes in.

naemono:
Fixed the integration test, which caught an issue where we weren't managing the empty slice of managed namespaces properly when initializing the controller.

barkbay (commented):
I spotted another place where some resources are listed and processed without predicates/options: https://github.com/elastic/cloud-on-k8s/blob/3016db432ee6dd8b03b6f0ad9666a8e6bb26a2cc/pkg/controller/license/list.go#L19-L23

We might have the same issue in the remote ca controller: https://github.com/elastic/cloud-on-k8s/blob/3016db432ee6dd8b03b6f0ad9666a8e6bb26a2cc/pkg/controller/remoteca/controller.go#L222-L225 _(not tested, to be confirmed)_

I'm afraid fixing all these errors / events and maintaining the code is tricky 😕

barkbay:
While having a look at the multicache implementation I noticed that cluster scope resources are now [handled](https://github.com/kubernetes-sigs/controller-runtime/blob/f236f0345ad2933912ebf34bfcf0f93620769654/pkg/cache/multi_namespace_cache.go#L90-L104): 

```golang
	// If the object is clusterscoped, get the informer from clusterCache,
	// if not use the namespaced caches.
	isNamespaced, err := objectutil.IsAPINamespaced(obj, c.Scheme, c.RESTMapper)
	if err != nil {
		return nil, err
	}
	if !isNamespaced {
		clusterCacheInf, err := c.clusterCache.GetInformer(ctx, obj)
		if err != nil {
			return nil, err
		}
		informers[globalCache] = clusterCacheInf

		return &multiNamespaceInformer{namespaceToInformer: informers}, nil
	}
```

(see this [PR](https://github.com/kubernetes-sigs/controller-runtime/pull/1418/files#diff-cf71f1c7387ae1d9a63da1f81380effac1b826d0c3c1594c4d05da4324be2159R145))

I think we can just stop adding the empty namespace: we should still be able to read `StorageClasses` and it should solve the issue without the need of a predicate.

naemono:
@barkbay great catch.   I'm testing to see if this simple fix is sufficient now in a separate branch.  Will report back.

naemono:
closing this per #5187

Issue #5040, opened 2021-11-10
E2E: Elasticsearch cluster health should not have been red during mutation process


pebrc:
https://devops-ci.elastic.co/job/cloud-on-k8s-e2e-tests-snapshot-versions/446//testReport
```
=== RUN   TestMutationResizeMemoryUp/Elasticsearch_cluster_health_should_not_have_been_red_during_mutation_process
    steps_mutation.go:128: 
        	Error Trace:	steps_mutation.go:128
        	Error:      	Not equal: 
        	            	expected: 0
        	            	actual  : 9
        	Test:       	TestMutationResizeMemoryUp/Elasticsearch_cluster_health_should_not_have_been_red_during_mutation_process
    steps_mutation.go:130: Elasticsearch cluster health check failure at 2021-11-10 05:36:06.457907754 +0000 UTC m=+5717.933233476: cluster health red
```

Variations of this error have been observed on recent 7.16 and 8.0 snapshot builds for similar mutation test cases.

The problem is that we start with a rolling upgrade  while Elasticsearch is still in the process of creating internal indices.
#4998 should fix the issue but in the mean time this issue serves to document the problem. 

![allocation-problem](https://user-images.githubusercontent.com/697790/141147066-b9aa0e9b-702d-4bce-8d3e-fd40532aa63e.png)

Sequence of events leading to the error:
1. e2e test runner checks for green cluster health OK
2. e2e test running triggers rolling upgrade
3. we disable replica shard allocation on the cluster
4. first Pod is deleted in K8s and connection draining via pre-stop hook begins ~50 secs
5. Elasticsearch starts allocating an internal index like ILM history that it had not allocated previously
6. It picks the already doomed Pod as the node to allocate to and because replica shard allocation is disabled only a primary is allocated
7. Pod is being terminated and cluster goes RED

Problems are twofold:
* we don't have a notion of Elasticsearch being ready in the sense that all internal indices have been successfully allocated. Cluster goes green even when a few internal indices have not been created yet
* Elasticsearch does not have a notion that we are going to take down a node or in fact already have lit the fuse for the destruction of a Pod (this will be addressed by the Node Shutdown API coming in #4998)

pebrc:
I did a bit more research and I think I understand better why these things happen (with some open questions remaining) 

> why are these two data streams (`logs-deprecation.elasticsearch-default` and something like `"ilm-history-5`) not there initially and only created at this unfortunate point in time? 

[...]

PR #5063, opened 2021-11-22
E2E: remove ineffective wait step

pebrc:
Related to #5040 

Removes a wait step added when I still had an incomplete understanding of the problem. This removes said step.


barkbay (approved):
LGTM

Could we also remove (or update the comment if still required) this one: https://github.com/elastic/cloud-on-k8s/blob/83b517e24593adee2108714acf6a5f0ec661f3c5/test/e2e/test/elasticsearch/builder.go#L292-L293

pebrc:
> Could we also remove (or update the comment if still required) this one:

I still find these helpful to diagnose test failures so I would keep them but I can change the comment to reflect that.

Issue #5017, opened 2021-11-04
Rename master branch to main


pebrc:
We would like to rename the current `master` branch to `main`. 

Timeline: right after the next (1.9) release (just to avoid any rename related hiccups impacting our ability to ship a release)

Things to consider: 
* Github has made it [much easier to rename branches](https://github.com/github/renaming#rename-existing)
* We are referencing `master` in our CI jobs, there is an `e2e/master` configuration for example, many if not all jobs except the release build are referencing the `master` branch
* Our docs might reference `master` but I think that is actually a bug and should be replaced with a variable pointing to the most recent release or the current branch  (so no hard-coding of the branch name)

Steps:

- [x] Plan the rename, flesh out the points I mentioned above, decide on the sequence of rename steps (when: can start now)
- [x] Rename according to the plan from the previous step (when: right after 1.9)

thbkrkr:
### (draft) Plan

#### Day before the renaming
- [ ] prepare elastic/cloud-on-k8s PR **A** to rename all occurences of `master` to `main` (see script below))
- [ ] prepare elastic/docs PR **B** to update the docs build config to point to main ([example](https://github.com/elastic/docs/pull/2266/files))

#### Day of the renaming
- [ ] rename `master` branch to `main` via the [GitHub UI](https://github.com/elastic/cloud-on-k8s/settings/branches) 
- [ ] update the pattern of the `master` branch protection rule via the [GitHub UI](https://github.com/elastic/cloud-on-k8s/settings/branches) 
- [ ] merge PR **A** & PR **B**
- [ ] send an email to the team to warn of the branch renaming and explain how to update their local clone


-----------

Script to help with the branch renaming for PR **A**:

```sh
#!/bin/bash -eux

replace() { xargs -n1 sed -i "s|$1|$2|g"; }

# rename files containing master
mv .ci/jobs/e2e-tests-master-gke.yml              .ci/jobs/e2e-tests-main-gke.yml
mv .ci/pipelines/e2e-tests-master-gke.Jenkinsfile .ci/pipelines/e2e-tests-main-gke.Jenkinsfile

# rename all occurences of `master` to `main` in `.ci`
grep master -lr .ci  | replace master main

# rename all occurences of `cloud-on-k8s/(blob|tree)/master/` to `cloud-on-k8s/(blob|tree)/main/` in the whole repo
grep master -lr .    | replace cloud-on-k8s/blob/master/  cloud-on-k8s/blob/main/
[...]

PR #5064, opened 2021-11-22
Point Elastic doc links to current instead of master

thbkrkr:
Point Elastic doc links that uses `master` to `current`. 

Relates to #5017.


barkbay (approved):
LGTM

PR #5067, opened 2021-11-23
Add new distribution channel

david-kow:
Moves specifying distribution channel from `manager` flag to config file and adds new channel.

### DIFF
--- a/deploy/eck-operator/templates/configmap.yaml
+++ b/deploy/eck-operator/templates/configmap.yaml
@@ -20,6 +20,7 @@ data:
     kube-client-timeout: {{ .Values.config.kubeClientTimeout }}
     elasticsearch-client-timeout: {{ .Values.config.elasticsearchClientTimeout }}
     disable-telemetry: {{ .Values.telemetry.disabled }}
+    distribution-channel: {{ .Values.telemetry.distributionChannel }}
     {{- if .Values.telemetry.interval }}
     telemetry-interval: {{ .Values.telemetry.interval }}
     {{- end }}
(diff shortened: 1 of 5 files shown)


deploy/eck-operator/templates/configmap.yaml
```diff
@@ -20,6 +20,7 @@ data:
     kube-client-timeout: {{ .Values.config.kubeClientTimeout }}
     elasticsearch-client-timeout: {{ .Values.config.elasticsearchClientTimeout }}
     disable-telemetry: {{ .Values.telemetry.disabled }}
+    distribution-channel: {{ .Values.telemetry.distributionChannel }}
```
pebrc, line 23: This would make `Helm` the distribution channel inside the config file that we embed in the Docker image because we are not overriding this value in `make generate-config-file`. This will have no effect in most cases because we ship another correctly templated configuration file as part of the `operator.yaml` with the correct distribution channel. But users creating their own manifests (if there are any) would be reported as `Helm` users.

pebrc (commented):
LGTM only one small concern regarding the embedded configuration file.

deploy/eck-operator/templates/configmap.yaml
david-kow: That's right, good catch. We could actually introduce another channel specifically for these by setting it in `extract.sh`. WDYT?

pebrc: Yes either that or just set the distribution channel value override to `all-in-one`  in that make target I guess, both would be ok for me.

pebrc (approved):

Issue #5066, opened 2021-11-22
Reload secure settings isn't adding plugin


perezjasonr:
## Bug Report
**What did you do?**

the following was done successfully:

installed s3 plugin, it shows up using the list command

added items to keystore, they also show up using list

then

tried to register an s3 repo

**What did you expect to see?**

a successful registration

**What did you see instead? Under which circumstances?**

```

PUT _snapshot/my_s3_repository
{
  "type": "s3",
  "settings": {
    "bucket": "my-bucket"
  }
}

RESPONSE:

{
  "error" : {
    "root_cause" : [
      {
        "type" : "repository_exception",
        "reason" : "[my_s3_repository] repository type [s3] does not exist"
      }
    ],
    "type" : "repository_exception",
    "reason" : "[my_s3_repository] repository type [s3] does not exist"
  },
  "status" : 500
}
```

also 

`_nodes/reload_secure_settings`

seems to be successful, but 

`GET _cat/plugins?v`

comes up empty. I'm not sure how its possible for it to claim the plugin is installed, then accept the keystore settings (it would complain it id didnt recognize the settings, like s3 access key, secret key, etc...but it accepted them), but then say it doesn't recognize the s3 type.

**Environment**

* ECK version:

	1.4.0

* Kubernetes information:

	disconnected environment, so the plugin was installed (successfully) using a url
	
```
$ kubectl version

Client Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.4", GitCommit:"e87da0bd6e03ec3fea7933c4b5263d151aafd07c", GitTreeState:"clean", BuildDate:"2021-02-18T16:12:00Z", GoVersion:"go1.15.8", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.4", GitCommit:"e87da0bd6e03ec3fea7933c4b5263d151aafd07c", GitTreeState:"clean", BuildDate:"2021-02-18T16:03:00Z", GoVersion:"go1.15.8", Compiler:"gc", Platform:"linux/amd64"}

```

* Resource definition:
```
n/a
```

* Logs:

from one of the nodes:

```

[...]

PR #5068, opened 2021-11-23
Add note to plugin documentation about using Istio and init containers.

naemono:
relates to #5066

### DIFF
--- a/docs/orchestrating-elastic-stack-applications/elasticsearch/bundles-plugins.asciidoc
+++ b/docs/orchestrating-elastic-stack-applications/elasticsearch/bundles-plugins.asciidoc
@@ -27,7 +27,7 @@ To run Elasticsearch with specific plugins or configuration files installed on E
 ** Easier to get started and upgrade versions.
 +
 * *Cons*
-** Requires pods to have internet access. *See <<istio-note,note>> below about using Istio*
+** Requires pods to have internet access. *Check <<istio-note,note>> below about using Istio*
 ** Adding new Elasticsearch nodes could randomly fail due to network issues or bad configuration.
 ** Each Elasticsearch node needs to repeat the download, wasting bandwidth and slowing startup.
 ** Deployment manifests are more complicated.
--- a/docs/orchestrating-elastic-stack-applications/elasticsearch/bundles-plugins.asciidoc
+++ b/docs/orchestrating-elastic-stack-applications/elasticsearch/bundles-plugins.asciidoc
@@ -25,10 +25,36 @@ This option has the following pros and cons:
 This option has the following pros and cons:
 +
 * Pro: Easier to get started and upgrade versions.
-* Con: Requires pods to have internet access.
+* Con: Requires pods to have internet access. *See <<istio-note,note>> below about using Istio*
 * Con: Adding new Elasticsearch nodes could randomly fail due to network issues or bad configuration.
 * Con: Each Elasticsearch node needs to repeat the download, wasting bandwidth and slowing startup.
 * Con: Deployment manifests are more complicated.
+
+[id="istio-note"]
+**Note when using Istio**
+
+When using Istio, init containers do *not* have network access, as the Envoy sidecar that provides network connectivity is not online.  In this scenario, custom containers are the best option.  If custom containers are simply not a viable option, then it is possible, although not recommended to adjust the startup command for the elasticsearch container itself to run the plugin installation before starting Elasticsearch, as the following example describes.
+
+[source,yaml]
+----
+spec:
+  nodeSets:
+  - name: default
+    count: 3
+    podTemplate:
+      spec:
+        containers:
+        - name: elasticsearch
+          command:
+          - /usr/bin/env
+          - bash
+          - -c
+          - |
+            #!/usr/bin/env bash
+            set -e
+            bin/elasticsearch-plugin install --batch repository-s3
+            /bin/tini -- /usr/local/bin/docker-entrypoint.sh
+----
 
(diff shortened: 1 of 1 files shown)


docs/orchestrating-elastic-stack-applications/elasticsearch/bundles-plugins.asciidoc
```diff
@@ -25,11 +25,37 @@ This option has the following pros and cons:
 This option has the following pros and cons:
 +
 * Pro: Easier to get started and upgrade versions.
-* Con: Requires pods to have internet access.
+* Con: Requires pods to have internet access. *See <<istio-note,note>> below about using Istio*
```
david-kow, line 28:
```suggestion
* Con: Requires pods to have internet access. *Check <<istio-note,note>> below about using Istio*
```

docs/orchestrating-elastic-stack-applications/elasticsearch/bundles-plugins.asciidoc
```diff
@@ -25,11 +25,37 @@ This option has the following pros and cons:
...
 * Con: Adding new Elasticsearch nodes could randomly fail due to network issues or bad configuration.
 * Con: Each Elasticsearch node needs to repeat the download, wasting bandwidth and slowing startup.
 * Con: Deployment manifests are more complicated.
 
+[id="istio-note"]
+**Note when using Istio**
+
+When using Istio, init containers do *not* have network access, as the Envoy sidecar that provides network connectivity is not online.  In this scenario, custom containers are the best option.  If custom containers are simply not a viable option, then it is possible, although not recommended to adjust the startup command for the elasticsearch container itself to run the plugin installation before starting Elasticsearch, as the following example describes.
```
david-kow, line 36:
```suggestion
When using Istio, init containers do *not* have network access, as the Envoy sidecar that provides network connectivity is not started yet.  In this scenario, custom containers are the best option. If custom containers are not a viable option, then it is possible to adjust the startup command for the elasticsearch container itself to run the plugin installation before starting Elasticsearch, as the following example describes. Note that this approach will require updating the startup command if it changes in the Elasticsearch image.
```

Assuming that user can't use custom containers I think this would be our recommended option - the wording right now sounds very discouraging though :) I'd openly mention the downside of this approach instead.

david-kow (approved):
LGTM, some nits.

docs/orchestrating-elastic-stack-applications/elasticsearch/bundles-plugins.asciidoc
```diff
@@ -25,11 +25,37 @@ This option has the following pros and cons:
...
 +
 * Pro: Easier to get started and upgrade versions.
-* Con: Requires pods to have internet access.
+* Con: Requires pods to have internet access. *See <<istio-note,note>> below about using Istio*
 * Con: Adding new Elasticsearch nodes could randomly fail due to network issues or bad configuration.
 * Con: Each Elasticsearch node needs to repeat the download, wasting bandwidth and slowing startup.
 * Con: Deployment manifests are more complicated.
 
+[id="istio-note"]
+**Note when using Istio**
```
pebrc, line 34: This makes it look like in the rendered version as if every below applies to Istio which it does not.

docs/orchestrating-elastic-stack-applications/elasticsearch/bundles-plugins.asciidoc
```diff
@@ -25,11 +25,37 @@ This option has the following pros and cons:
...
+        - name: elasticsearch
+          command:
+          - /usr/bin/env
+          - bash
+          - -c
+          - |
+            #!/usr/bin/env bash
+            set -e
+            bin/elasticsearch-plugin install --batch repository-s3
+            /bin/tini -- /usr/local/bin/docker-entrypoint.sh
```
pebrc, line 56: This approach has the downside that it will break if anything changes in the way the Elasticsearch startup script is implemented. So I wonder if we should feature it so prominently.

alaudazzi:
@pebrc, @david-kow, @naemono  
I reformatted the pros and cons list as follows:
image

WDYT?

docs/orchestrating-elastic-stack-applications/elasticsearch/bundles-plugins.asciidoc
thbkrkr: The full stop is missing.

docs/orchestrating-elastic-stack-applications/elasticsearch/bundles-plugins.asciidoc
naemono: I adjusted this wording, along with adding a note about when this could potentially affect a user.

docs/orchestrating-elastic-stack-applications/elasticsearch/bundles-plugins.asciidoc
naemono: Moved Istio note to the end of the doc.  LMK if this is acceptable.

docs/orchestrating-elastic-stack-applications/elasticsearch/bundles-plugins.asciidoc
naemono: Is moving it to the end of the doc less prominent, or is some other approach desired?

naemono:
> WDYT?

Feels more readable to me.  Thanks!

thbkrkr (changes requested):
Almost LGTM!

There are some unexpected `::` to remove and a period to add.

thbkrkr (approved):

PR #5069, opened 2021-11-24
Update Kubernetes version for EKS to 1.20

thbkrkr:
1.17 which we used is no longer supported.

### DIFF
--- a/hack/deployer/config/plans.yml
+++ b/hack/deployer/config/plans.yml
@@ -105,7 +105,7 @@ plans:
   provider: eks
   machineType: c5d.2xlarge
   serviceAccount: false
-  kubernetesVersion: 1.17
+  kubernetesVersion: 1.18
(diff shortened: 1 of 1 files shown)


hack/deployer/config/plans.yml
```diff
@@ -105,7 +105,7 @@ plans:
   provider: eks
   machineType: c5d.2xlarge
   serviceAccount: false
-  kubernetesVersion: 1.17
+  kubernetesVersion: 1.18
```
barkbay, line 108: Default version seems to be `1.20.9`, any reason to not use it ? (also I can't see `1.18` in the list)
image

thbkrkr: I was looking at https://docs.aws.amazon.com/eks/latest/userguide/kubernetes-versions.html. +1 to use `1.20`.

barkbay (approved):
LGTM

PR #5072, opened 2021-11-24
Introduce orchestration hints as a way to gate orchestration features

### LINKED ISSUE #4997: Stop using transient settings as of 7.16
Transient cluster settings are deprecated as of 7.16. https://github.com/elastic/elasticsearch/pull/78794

We should stop using them starting with certain cut off version and either consider persistent settings or see if we can remove the call to the settings API. I am specifically thinking of the settings calls for allocation filtering during data migrations that we won't need anymore once we implement the Node Shutdown API as proposed in https://github.com/elastic/cloud-on-k8s/pull/4597

There is the question how we are going to migrate from one form to another. E.g. assuming we start using persistent settings with 7.16 and we disabled allocations in 7.15 with the transient version of the setting. In this case we would have to check for both options when removing settings and only set them as persistent settings. 

We should also investigate which steps are necessary before upgrading to 8.0: e.g. do we need to remove all transient settings before an upgrade to 8.0 can succeed? Or will the upgrade work and we will just get warnings?

pebrc:
Fixes #4997 

* using a new annotation to store orchestration hints 
* use this field to make decisions about orchestration features that are relevant or not (currently enabling allocations and use of transient settings both not needed anymore when using node shutdown API)
* not using a single hint  value here as both things currently controlled can fail independently from each other and to allow future extensions
* if lost will run the logic that added the hints again and recreate the status

We could verify the absence of transient settings via an e2e test step for the relevant versions which I am happy to add. Curious to hear your opinions.


pebrc:
We discussed this offline and agreed to use an annotation instead with a full JSON document with speaking names instead of a bit mask. 

Why annotation: to make it less prominent as it would be in the status. 
Why not bit mask: concerns around debugging of a bit mask.

pkg/controller/elasticsearch/elasticsearch_controller.go
```diff
@@ -295,6 +310,30 @@ func (r *ReconcileElasticsearch) updateStatus(
...
+	ctx context.Context,
+	es esv1.Elasticsearch,
+	reconcileState *esreconcile.State,
+) error {
+	span, _ := apm.StartSpan(ctx, "update_annotations", tracing.SpanTypeApp)
+	defer span.End()
+
+	// cluster-uuid is a special case that is not treated here but through an immediate update due to the risk of data loss
+	// see bootstrap package.
+	newAnnotations, err := reconcileState.Annotations()
```
barkbay, line 323: nit: using `reconcileState.OrchestrationHints().AsAnnotation()` would help to understand what annotations are set by this part of the code.
```suggestion
	newAnnotations, err := reconcileState.OrchestrationHints().AsAnnotation()
```

pkg/controller/elasticsearch/elasticsearch_controller.go
```diff
@@ -295,6 +310,30 @@ func (r *ReconcileElasticsearch) updateStatus(
 	return common.UpdateStatus(r.Client, cluster)
 }
 
+func (r *ReconcileElasticsearch) annotateResource(
```
barkbay, line 313: nit: Maybe add a comment to explain the purpose and the scope of the annotation set by this function.

pkg/controller/elasticsearch/elasticsearch_controller.go
```diff
@@ -295,6 +310,30 @@ func (r *ReconcileElasticsearch) updateStatus(
...
+	ctx context.Context,
+	es esv1.Elasticsearch,
+	reconcileState *esreconcile.State,
+) error {
+	span, _ := apm.StartSpan(ctx, "update_annotations", tracing.SpanTypeApp)
```
barkbay, line 318:
```suggestion
	span, _ := apm.StartSpan(ctx, "update_orchestration_hints_annotation", tracing.SpanTypeApp)
```

pkg/controller/elasticsearch/elasticsearch_controller.go
```diff
@@ -295,6 +310,30 @@ func (r *ReconcileElasticsearch) updateStatus(
...
+
+	// cluster-uuid is a special case that is not treated here but through an immediate update due to the risk of data loss
+	// see bootstrap package.
+	newAnnotations, err := reconcileState.Annotations()
+	if err != nil {
+		return err
+	}
+
+	expected := maps.Merge(es.ObjectMeta.DeepCopy().Annotations, newAnnotations)
+	if reflect.DeepEqual(expected, es.Annotations) {
```
barkbay, line 329: It does not really matter for now since there is only one hint, however I think we assume here the serialization of `OrchestrationsHints` to be the same across reconciliations. However [I'm not sure there is such a guarantee](https://stackoverflow.com/questions/44755089/does-serialized-content-strictly-follow-the-order-in-definition-use-encoding-json-package).

barkbay (approved):
LGTM

PR #5076, opened 2021-11-25
Bump Stack version to 7.15.2

### LINKED ISSUE #4896: Update recipe fleet-custom-logs-integration.yaml once Stack 7.15.1 is out
See https://github.com/elastic/cloud-on-k8s/pull/4873#issuecomment-930247845 for all the details.
We haven't updated that recipe for 7.15.0 because of a bug in Kibana. Let's make sure we do it once 7.15.1 is out.

thbkrkr:
* Bump min `6.8.x` version to `6.8.20`
* Bump max `7.x` version to `7.15.2`

Resolves #4896.


pebrc (approved):
Will need to be updated soon to 7.16 but I guess that's ok

PR #5084, opened 2021-11-29
Update version to 1.10.0-SNAPSHOT

david-kow:


pebrc (changes requested):
I think we are not going to do a 1.10 but go to 2.0 directly

PR #5093, opened 2021-11-29
Improve documentation around cpu and memory resources

### LINKED ISSUE #3985: Consider the importance of setting resources.requests.cpu for Elasticsearch nodeSets
For consideration and discussion :)

We should highlight in the docs the importance of configuring `resources.requests.cpu` at nodeSet level for Elasticsearch clusters, or __even setting automatically a default of 1 or 2 processors if the user don't provide any value__ (as we are doing for memory requests and limit by default).

If that's not set, by default the pod is going to request [100m](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#how-pods-with-resource-limits-are-run) cores, which is too low (1/10th of a core). That will probably lead to a bad experience and super low performance at some point. In some cases the users might wrongly believe or expect that the performance will improve eventually if there isn't any limit defined (or if the limit is high and the node has a lot of CPUs), but that won't happen due to the nature of Elasticsearch and how the thread pools are configured during the instance restart.

In that scenario, Elasticsearch will dimension the thread pools considering a `node.processors` of 2 (which is the default value if the number of available processors detected is less than 2). So we would have an instance dimensioned for 2 processors but with a real `cpu-quota` of 1/10th of a processor (high cpu over-allocation). This could end up in very bad experience as described above. Based on my tests (which maybe are wrong) I would always suggest a minimum of `1` processor to the `resources.requests.cpu` setting (for small instances expected to handle low load actually).

We have currently the following doc and sections (which are nice btw):
- https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-managing-compute-resources.html#k8s-compute-resources-elasticsearch
- https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-managing-compute-resources.html#k8s-default-behavior --> This only affects to memory at the moment.

That doc is generic to the entire stack and not strictly related with Elasticsearch (which makes them more difficult to find when we are focused on Elasticsearch tuning). I believe that Elasticsearch CPU configuration deserves its own doc it the Elasticsearch section, the same as the "JVM Heap" doc or other docs designed specifically for ES.

My suggestions are:
- Create a doc in the Elasticsearch section, just after the "JVM Heap Size" doc for "CPU and processors" configuration, explaining everything there, explaining the defaults (__if any__) and the recommendations. Explain also the risk of not setting `requests.cpu` as the performance might be too low and even if there's no limit the instance would never be able to use all the CPU capacity of the k8s node even if this is totally idle.
- For recommendations my suggestions are:
  - Provide always `requests.cpu`  and `limits.cpu` and the same values in both places, in order to have clearer expectations in terms of the performance that the instance should provide. (__note: maybe we don't want to suggest setting `limits.cpu` due to the known issue mentioned in the docs).
  - If the user wants to set a limit higher than the `requests.cpu` the difference shouldn't be high as that would be a bit useless (the thread pools are dimensioned based on the `requests.cpu` and they will never grow. We need the users to understand that a request of 2 and a limit of 16 CPUs makes no sense, as that instance will always perform more or less like a 2 cores instance (maybe a bit better, but just a bit) and never like an instance with 16 cores.

- Change the code (__breaking change__) to set a default of 1 processor if `requests.cpu` is not set, and allow only smaller values if they are directly set in the `PodSpec`. In my view the default of `100m` is only valid for proof of concepts without almost any load, and we should default to something "usable". If for some reason we accept this change, the default should be documented in the Elasticsearch section of the docs and we should indicate that for `lower` values the setting needs to be set.

From the current doc the following sentence is very good and a nice explanation but __not enough__ in my view to prevent users having unexpected performance issues:

> The value set for cpu requests directly impacts Elasticsearch node.processors setting. For example, with resources.requests.cpu: 1, Elasticsearch effectively relies on a single core, which may significantly limit performance

So, we are telling the customers that 1 is a low number, but the default if they don't set it is a tenth of that value :)

barkbay:
This PR:

* Fixes #3985 by adding a note to highlight that default requests and limits must be adapted for any use case which goes beyond the "quick start"
* Moves the `JVM Heap Size section` into `Manage compute resources`. There was a circular dependency between the two sections and I feel it's fine to have both information (settings limits and the JVM heap) at the same place (especially since there is a direct relationship between `resources.limits.memory` and the heap settings starting with `7.11`). _Happy to reconsider this change though._
* Moves some notes which are not only relevant to ` Elasticsearch`, but to any Pod running on K8S (QoS, CPU throttling)
* Adds a table in an attempt to explain the impact of CPU limits and request on the Elasticsearch `node.processors` setting.
* Adds a section to explain how to monitor compute resources. I'm not :100:  convinced about this one as I think it goes beyond ECK, but I did not find a relevant topic in the Beat documentation. Glad to have your thoughts.

Remaining TODOs:
- [x] ~Add a note about 2 being the default value for `node.processors` if the number of available processors detected is less than 2~ _Edit: it does not seem to be the case_
- [x] ~Add CPU to https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-managing-compute-resources.html#k8s-default-behavior~ _Edit: no CPU resources is set, with exception of [Agent](https://github.com/elastic/cloud-on-k8s/blob/66a41dbfee781a04e66c76d8646be3f60b1de1b5/pkg/controller/agent/pod.go#L89) and [Beat](https://github.com/elastic/cloud-on-k8s/blob/66a41dbfee781a04e66c76d8646be3f60b1de1b5/pkg/controller/beat/common/pod.go#L48)_


pebrc (approved):
LGTM I like the examples with the screenshots as they illustrate how to use and interpret the output of our monitoring features.

PR #5097, opened 2021-11-29
Support trial licenses inside enterprise_trial orchestration licenses

pebrc:
Currently `enterprise_trial` orchestration licenses can only contain short-lived `enterprise` or `platinum` licenses (which one is applied depends on the version of Elasticsearch). This is confusing to users which expect a trial to be running if they apply a trial license. This PR prepares the ground to remedy this confusion by consistently using trial license across both the orchestrator and the Elastic stack. 

 I have adjusted the existing e2e test to cover both cases, which means we no longer test the "I can apply the same enterprise_trial multiple times (as long as it's valid)". But I would argue that this is only about being able to apply a regular `enterprise` license to Elasticsearch which we are already covering in another test. So instead we now testing once the old-style `enterprise_trial` with short-lived `enterprise` and `platinum` stack licenses and once the new-style `enterprise_trial` with a stack `trial` license inside.

### DIFF
--- a/pkg/controller/common/license/match_test.go
+++ b/pkg/controller/common/license/match_test.go
@@ -225,14 +225,13 @@ func Test_bestMatchAt(t *testing.T) {
 							Type:               LicenseTypeEnterpriseTrial,
 							Issuer:             "API",
 							ClusterLicenses: []ElasticsearchLicense{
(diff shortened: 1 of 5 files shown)


barkbay (approved):
LGTM

Issue #5105, opened 2021-12-02
TestFleetCustomLogsIntegrationRecipe is failing in 7.15.2


thbkrkr:
```
=== RUN   TestFleetCustomLogsIntegrationRecipe/ES_data_should_pass_validations
Retries (30m0s timeout): ..............................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................
    step.go:43: 
        	Error Trace:	utils.go:87
        	Error:      	Received unexpected error:
        	            	elasticsearch client failed for https://elasticsearch-d66t-es-http.e2e-saklj-mercury.svc:9200/_data_stream/logs-elastic_agent-default: 404 Not Found: {Status:404 Error:{CausedBy:{Reason: Type:} Reason:no such index [logs-elastic_agent-default] Type:index_not_found_exception RootCause:[{Reason:no such index [logs-elastic_agent-default] Type:index_not_found_exception}]}}
        	Test:       	TestFleetCustomLogsIntegrationRecipe/ES_data_should_pass_validations
{"log.level":"error","@timestamp":"2021-12-01T17:58:42.511Z","message":"stopping early","service.version":"0.0.0-SNAPSHOT+00000000","service.type":"eck","ecs.version":"1.4.0","error":"test failure","error.stack_trace":"github.com/elastic/cloud-on-k8s/test/e2e/test/helper.RunFile\n\t/go/src/github.com/elastic/cloud-on-k8s/test/e2e/test/helper/yaml.go:162\ngithub.com/elastic/cloud-on-k8s/test/e2e/agent.runAgentRecipe\n\t/go/src/github.com/elastic/cloud-on-k8s/test/e2e/agent/recipes_test.go:226\ngithub.com/elastic/cloud-on-k8s/test/e2e/agent.TestFleetCustomLogsIntegrationRecipe\n\t/go/src/github.com/elastic/cloud-on-k8s/test/e2e/agent/recipes_test.go:160\ntesting.tRunner\n\t/usr/local/go/src/testing/testing.go:1259"}
    --- FAIL: TestFleetCustomLogsIntegrationRecipe/ES_data_should_pass_validations (1800.00s)
```

Elastic Agent logs:

```
Performing setup of Fleet in Kibana                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                     

[...]

PR #5108, opened 2021-12-02
Update fleet custom logs integration recipe in 7.15.1

thbkrkr:
Specify `paths` as an array value but temporarily pin the stack version to `7.15.1` because it breaks in `7.15.2`.

Relates to #5105.


david-kow (approved):

thbkrkr:
I haven't found how to make the recipe work in `7.15.2`. I dreamed, pinning the version of the log package does not work. Let's stay in `7.15.1` for the moment.

thbkrkr:
Specify paths as an array value but temporarily pin the stack version to 7.15.1 because it breaks in 7.15.2.

PR #5120, opened 2021-12-06
[DOCS] Hardcoded broken cross-doc links for 7.16 release.

debadair:
The links to current will break when we flip to 7.16. Hardcoding them for now & will open a second PR to update them to use current instead of 7.16.


nrichers (approved):
LGTM! Thank you so much for fixing these broken links in our ECK docs. :shipit:

PR #5121, opened 2021-12-06
[DOCS] Switched hardcoded 7.16 links to current

debadair:
Updated the hardcoded 7.16 links to use current.

DO NOT MERGE UNTIL 7.16 is released.


nrichers (commented):
@alaudazzi I noticed that we hardcode links to `current` which can be problematic. Ideally, we'd use a shared version file for this stuff. It's not a perfect solution, but one step up from hardcoding links directly. I wonder if this might not be something to raise with the ECK team at some point?

nrichers (approved):
LGTM. 🐢

debadair:

nrichers:
@elasticcloudmachine update branch

debadair:

PR #5129, opened 2021-12-07
Add internal product header to requests

david-kow:
Adds internal product header to ES client and all probe requests.

### DIFF
--- a/pkg/controller/common/http.go
+++ b/pkg/controller/common/http.go
@@ -4,21 +4,8 @@
 
 package common
 
-import (
-	"net/http"
-)
-
 const (
 	InternalProductRequestHeaderString = "x-elastic-product-origin: cloud"
-	internalProductRequestHeaderKey    = "x-elastic-product-origin"
-	internalProductRequestHeaderValue  = "cloud"
+	InternalProductRequestHeaderKey    = "x-elastic-product-origin"
+	InternalProductRequestHeaderValue  = "cloud"
 )
-
-// SetInternalProductRequestHeader sets header that specifies which internal product is making the request.
-func SetInternalProductRequestHeader(req *http.Request) {
-	if req == nil || req.Header == nil {
-		return
-	}
-
-	req.Header.Set(internalProductRequestHeaderKey, internalProductRequestHeaderValue)
-}
--- a/pkg/controller/common/http.go
+++ b/pkg/controller/common/http.go
@@ -15,26 +15,19 @@ const (
 	internalProductRequestHeaderValue = "cloud"
 )
 
+// InternalProductHTTPHeaders is a slice that contains header that specifies which product is making the request.
+var InternalProductHTTPHeaders = []corev1.HTTPHeader{
+	{
+		Name:  internalProductRequestHeaderKey,
+		Value: internalProductRequestHeaderValue,
+	},
+}
+
 // SetInternalProductRequestHeader sets header that specifies which internal product is making the request.
 func SetInternalProductRequestHeader(req *http.Request) {
-	req.Header.Set(GetInternalProductRequestHeaderKey(), GetInternalProductRequestHeaderValue())
-}
-
-// CreateInternalProductHTTPHeaders creates header slice that contains header that specifies which internal
-// product is making the request.
-func CreateInternalProductHTTPHeaders() []corev1.HTTPHeader {
-	return []corev1.HTTPHeader{
-		{
-			Name:  GetInternalProductRequestHeaderKey(),
-			Value: GetInternalProductRequestHeaderValue(),
-		},
+	if req == nil || req.Header == nil {
+		return
 	}
-}
 
-func GetInternalProductRequestHeaderKey() string {
-	return internalProductRequestHeaderKey
-}
-
-func GetInternalProductRequestHeaderValue() string {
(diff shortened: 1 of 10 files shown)


pkg/controller/common/http.go
```diff
@@ -0,0 +1,40 @@
...
+)
+
+const (
+	internalProductRequestHeaderKey   = "x-elastic-product-origin"
+	internalProductRequestHeaderValue = "cloud"
+)
+
+// SetInternalProductRequestHeader sets header that specifies which internal product is making the request.
+func SetInternalProductRequestHeader(req *http.Request) {
+	req.Header.Set(GetInternalProductRequestHeaderKey(), GetInternalProductRequestHeaderValue())
```
barkbay, line 20: `req.Header` is a map, should we check it's non nil before using it ?

pkg/controller/common/http.go
```diff
@@ -0,0 +1,40 @@
...
+// CreateInternalProductHTTPHeaders creates header slice that contains header that specifies which internal
+// product is making the request.
+func CreateInternalProductHTTPHeaders() []corev1.HTTPHeader {
+	return []corev1.HTTPHeader{
+		{
+			Name:  GetInternalProductRequestHeaderKey(),
+			Value: GetInternalProductRequestHeaderValue(),
+		},
+	}
+}
```
barkbay, line 32: nit: Could we just allocate the slice once ?

```suggestion
// InternalProductHTTPHeaders is a slice that contains header that specifies which product is making the request.
var InternalProductHTTPHeaders = []corev1.HTTPHeader{
	{
		Name:  internalProductRequestHeaderKey,
		Value: internalProductRequestHeaderValue,
	},
}
```

pkg/controller/common/http.go
```diff
@@ -0,0 +1,40 @@
...
+
+func GetInternalProductRequestHeaderKey() string {
```
barkbay, line 34: nit: does not seem to be used outside of this file

pkg/controller/common/http.go
```diff
@@ -0,0 +1,40 @@
...
+	return internalProductRequestHeaderKey
+}
+
+func GetInternalProductRequestHeaderValue() string {
```
barkbay, line 38: nit: does not seem to be used outside of this file

pkg/controller/common/http.go
barkbay, line 20:
```suggestion
	req.Header.Set(internalProductRequestHeaderKey, internalProductRequestHeaderValue)
```

pkg/controller/common/http.go
```diff
@@ -0,0 +1,33 @@
...
+// InternalProductHTTPHeaders is a slice that contains header that specifies which product is making the request.
+var InternalProductHTTPHeaders = []corev1.HTTPHeader{
+	{
+		Name:  internalProductRequestHeaderKey,
+		Value: internalProductRequestHeaderValue,
+	},
+}
+
+// SetInternalProductRequestHeader sets header that specifies which internal product is making the request.
+func SetInternalProductRequestHeader(req *http.Request) {
```
barkbay, line 27: I did not realized that this function is only called once, by the ES client, I think we could inline it.
Also I would consider having a `nil` parameter a programming error.
I was thinking about something along those lines in `base.go`:

```golang
	request, err := http.NewRequest(method, stringsutil.Concat(c.Endpoint, pathWithQuery), body) //nolint:noctx
	if err != nil {
		return err
	}

	// Sets headers allowing ES to distinguish between deprecated APIs used internally and by the user
	if request.Header == nil {
		request.Header = make(http.Header)
	}
	request.Header.Set(common.InternalProductRequestHeaderKey, common.InternalProductRequestHeaderValue)
```

_(Assuming `internalProductRequestHeaderKey` and `internalProductRequestHeaderValue` are public const)_

pkg/controller/elasticsearch/client/base.go
```diff
@@ -133,6 +134,9 @@ func (c *baseClient) request(
 		return err
 	}
 
+	// Sets headers allowing ES to distinguish between deprecated APIs used internally and by the user
+	common.SetInternalProductRequestHeader(request)
```
barkbay, line 138: nit: I would update the unit test [TestClient_request](https://github.com/elastic/cloud-on-k8s/blob/b3f4ae2fd7b281f5a9a8693132772b4a347d409f/pkg/controller/elasticsearch/client/client_test.go#L236) to ensure that the header is set.

naemono (approved):
Looks straightforward to me.

pkg/controller/common/http.go
```diff
@@ -0,0 +1,11 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License 2.0;
+// you may not use this file except in compliance with the Elastic License 2.0.
+
+package common
+
+const (
+	InternalProductRequestHeaderString = "x-elastic-product-origin: cloud"
```
pebrc, line 8: Nit: why are we not just concatenating key and value with a colon at the place where we need both instead of duplicating the content in the `*String` version of the constant?

pebrc (approved):
LGTM

PR #5150, opened 2021-12-09
Add autodetect option to security context configuration

### LINKED ISSUE #5061: Add autodetect option to set-default-security-context flag
The behaviour of the autodetect flag should be that it will make the operator try to detect whether it's running on OCP. If yes, operator will behave as the `set-default-security-context` was set to `false`, otherwise the operator will behave as the `set-default-security-context` was set to `true`.

For more context check the discussion in https://github.com/elastic/cloud-on-k8s/issues/2791. 
The "warning" about this behaviour will be also added to ECK `1.9.0` release notes (check https://github.com/elastic/cloud-on-k8s/issues/5062).

naemono:
resolves #5061 

This change adds an option to the `set-default-security-context` flag to help the user with settings that pertain to [security context](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/).  It adds the `auto-detect` option ~~(which is triggered by simply not setting the `set-default-security-context` flag)~~ which does the following when set:

1. If the `set-default-security-context` flag is explicitly set to either true, or false, use this value.
2. if the `set-default-security-context` flag is set to `auto-detect` (the default value in helm charts) use openshift detection to determine whether or not we are running within an openshift cluster.  If we determine we are on an openshift cluster then default to setting this flag to `false` to allow openshift to set the security context properly, otherwise set this flag to `true` to allow the operator to set the security context appropriately.

Any other set value will throw a syntax error.

## testing
This was tested in both an openshift cluster, and vanilla cluster with the auto-detect option set and the results were as follows

### vanilla cluster

Elasticsearch stateful set had the following settings
```
spec.securityContext:
          fsGroup: 1000
```

### openshift cluster
```
container.securityContext:
      capabilities:
        drop:
        - KILL
        - MKNOD
        - SETGID
        - SETUID
      runAsUser: 1000630000

spec.securityContext:
    fsGroup: 1000630000
    seLinuxOptions:
      level: s0:c25,c15
```

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -289,8 +289,8 @@ func Command() *cobra.Command {
 	)
 	cmd.Flags().String(
 		operator.SetDefaultSecurityContextFlag,
-		"",
-		"Enables setting the default security context with fsGroup=1000 for Elasticsearch 8.0+ Pods. Ignored pre-8.0. Possible values: true, false, \"\" (= auto-detect)",
+		"auto-detect",
+		"Enables setting the default security context with fsGroup=1000 for Elasticsearch 8.0+ Pods. Ignored pre-8.0. Possible values: true, false, auto-detect",
 	)
 
 	// hide development mode flags from the usage message
@@ -690,10 +690,10 @@ func chooseAndValidateIPFamily(ipFamilyStr string, ipFamilyDefault corev1.IPFami
 //    to determine whether or not we are running within an openshift cluster.  If we determine we are on an openshift cluster,
 //    then default to false, otherwise, return true.
 func determineSetDefaultSecurityContext(setDefaultSecurityContext string, clientset kubernetes.Interface) (bool, error) {
-	if len(setDefaultSecurityContext) > 0 {
-		return strconv.ParseBool(setDefaultSecurityContext)
+	if setDefaultSecurityContext == "auto-detect" {
+		return isOpenShift(clientset)
 	}
-	return isOpenShift(clientset)
+	return strconv.ParseBool(setDefaultSecurityContext)
 }
 
 func isOpenShift(clientset kubernetes.Interface) (bool, error) {
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -693,6 +693,10 @@ func determineSetDefaultSecurityContext(setDefaultSecurityContext string, client
 	if len(setDefaultSecurityContext) > 0 {
 		return strconv.ParseBool(setDefaultSecurityContext)
 	}
+	return isOpenShift(clientset)
+}
+
+func isOpenShift(clientset kubernetes.Interface) (bool, error) {
 	openshiftSecurityGroupVersion := schema.GroupVersion{Group: "security.openshift.io", Version: "v1"}
 	apiResourceList, err := clientset.Discovery().ServerResourcesForGroupVersion(openshiftSecurityGroupVersion.String())
 	if err != nil {
@@ -720,7 +724,6 @@ func determineSetDefaultSecurityContext(setDefaultSecurityContext string, client
 
 	// we could not determine that we are running within an openshift cluster,
 	// so we will behave as if "setDefaultSecurityContext" was set to true.
-	log.V(1).Info("returning true as fallback")
 	return true, nil
 }
 
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -694,7 +694,7 @@ func determineSetDefaultSecurityContext(setDefaultSecurityContext string, client
 	if err == nil {
 		return isSet, nil
 	}
-	apis, err := clientset.Discovery().ServerResources()
+	_, apiResources, err := clientset.Discovery().ServerGroupsAndResources()
 	if err != nil && !discovery.IsGroupDiscoveryFailedError(err) {
 		return false, err
 	}
@@ -705,19 +705,21 @@ func determineSetDefaultSecurityContext(setDefaultSecurityContext string, client
 	// In case of an error, check if security.openshift.io is the reason (unlikely).
 	// If it is, we are obviously on an openshift cluster.
 	if discovery.IsGroupDiscoveryFailedError(err) {
-		e := err.(*discovery.ErrGroupDiscoveryFailed)
-		if _, exists := e.Groups[openshiftSecurityGroupVersion]; exists {
-			// since we have determined we are absolutely running within an openshift cluster,
-			// behave as if "setDefaultSecurityContext" was set to false.
-			return false, nil
+		var e *discovery.ErrGroupDiscoveryFailed
+		if ok := errors.As(err, &e); ok {
+			if _, exists := e.Groups[openshiftSecurityGroupVersion]; exists {
+				// since we have determined we are absolutely running within an openshift cluster,
+				// behave as if "setDefaultSecurityContext" was set to false.
+				return false, nil
+			}
 		}
 	}
 
 	// search for "securitycontextconstraints" within the cluster's api resources,
 	// since this is an openshift specific api resource that does not exist outside of openshift.
-	for _, api := range apis {
-		if api.GroupVersion == openshiftSecurityGroupVersion.String() {
-			for _, resource := range api.APIResources {
+	for _, apiResource := range apiResources {
+		if apiResource.GroupVersion == openshiftSecurityGroupVersion.String() {
+			for _, resource := range apiResource.APIResources {
 				if resource.Name == "securitycontextconstraints" {
 					// since we have determined we are absolutely running within an openshift cluster,
 					// behave as if "setDefaultSecurityContext" was set to false.
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -12,6 +12,7 @@ import (
 	"net/http/pprof"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"time"
 
@@ -23,7 +24,9 @@ import (
 	"go.uber.org/automaxprocs/maxprocs"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/kubernetes"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp" // allow gcp authentication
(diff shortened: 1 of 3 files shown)


cmd/manager/main.go
```diff
@@ -674,6 +683,58 @@ func chooseAndValidateIPFamily(ipFamilyStr string, ipFamilyDefault corev1.IPFami
...
+	if err != nil && !discovery.IsGroupDiscoveryFailedError(err) {
+		return false, err
+	}
+
+	openshiftSecurityGroupName := "security.openshift.io"
+	openshiftSecurityGroupVersion := schema.GroupVersion{Group: openshiftSecurityGroupName, Version: "v1"}
+
+	// In case of an error, check if security.openshift.io is the reason (unlikely).
+	// If it is, we are obviously on an openshift cluster.
+	if discovery.IsGroupDiscoveryFailedError(err) {
```
pebrc, line 707: Should we separate the logic that detects whether we are running on OpenShift into a function of its own?

deploy/eck-operator/values.yaml
```diff
@@ -159,7 +159,12 @@ config:
   exposedNodeLabels: [ "topology.kubernetes.io/.*", "failure-domain.beta.kubernetes.io/.*" ]
 
   # setDefaultSecurityContext determines whether a default security context is set on application containers created by the operator.
-  setDefaultSecurityContext: true
+  # *note* that the default option now is "", which is using "auto-detect" to attempt to set this properly automatically when both running
+  # in an openshift cluster, and a standard kubernetes cluster.  Valid values are as follows:
+  #   ""    : auto detect
```
pebrc, line 164: Should we support an additional value called `auto-detect` or `autodetect` that we could use here so we avoid the awkwardness of the empty string value.

cmd/manager/main.go
```diff
@@ -674,6 +683,58 @@ func chooseAndValidateIPFamily(ipFamilyStr string, ipFamilyDefault corev1.IPFami
...
+// 1. If the setDefaultSecurityContext is explicitly set to either true, or false, use this value.
+// 2. use openshift detection pulled from kubevirt
+//    https://github.com/kubevirt/kubevirt/blob/f71e9c9615a6c36178169d66814586a93ba515b5/pkg/util/cluster/cluster.go#L21
+//    to determine whether or not we are running within an openshift cluster.  If we determine we are on an openshift cluster,
+//    then default to false, otherwise, return true.
+func determineSetDefaultSecurityContext(setDefaultSecurityContext string, clientset kubernetes.Interface) (bool, error) {
+	isSet, err := strconv.ParseBool(setDefaultSecurityContext)
+	if err == nil {
+		return isSet, nil
+	}
```
barkbay, line 696: I think we should return the error here. If an unexpected value is set the operator will use the autodetect mechanism without any warning/error.

```suggestion
	if len(setDefaultSecurityContext) > 0 {
		isSet, err := strconv.ParseBool(setDefaultSecurityContext)
		if err != nil {
			return false, err
		}
		return isSet, nil
	}
```

cmd/manager/main.go
```diff
@@ -674,6 +683,58 @@ func chooseAndValidateIPFamily(ipFamilyStr string, ipFamilyDefault corev1.IPFami
...
+		if apiResource.GroupVersion == openshiftSecurityGroupVersion.String() {
+			for _, resource := range apiResource.APIResources {
+				if resource.Name == "securitycontextconstraints" {
+					// since we have determined we are absolutely running within an openshift cluster,
+					// behave as if "setDefaultSecurityContext" was set to false.
+					return false, nil
+				}
+			}
+		}
+	}
```
barkbay, line 730: Slightly shorter / optimized version:

```suggestion
	openshiftSecurityGroupVersion := schema.GroupVersion{Group: "security.openshift.io", Version: "v1"}
	apiResourceList, err := clientset.Discovery().ServerResourcesForGroupVersion(openshiftSecurityGroupVersion.String())
	if err != nil {
		// In case of an error, check if security.openshift.io is the reason (unlikely).
		var e *discovery.ErrGroupDiscoveryFailed
		if ok := errors.As(err, &e); ok {
			if _, exists := e.Groups[openshiftSecurityGroupVersion]; exists {
				// since we have determined we are absolutely running within an openshift cluster,
				// behave as if "setDefaultSecurityContext" was set to false.
				return false, nil
			}
		}
		return false, err
	}

	// search for "securitycontextconstraints" within the cluster's api resources,
	// since this is an openshift specific api resource that does not exist outside of openshift.
	for _, apiResource := range apiResourceList.APIResources {
		if apiResource.Name == "securitycontextconstraints" {
			// since we have determined we are absolutely running within an openshift cluster,
			// behave as if "setDefaultSecurityContext" was set to false.
			return false, nil
		}
	}
```

deploy/eck-operator/values.yaml
naemono: So, have helm values of ["on","off","auto-detect"], which translates to ["on","off",""] in the flags of the operator?  I assume this also means that if its unset, that it's inherently "auto-detect"?  I think that makes good sense.  Will do.  Let me know if my assumptions here are incorrect.

cmd/manager/main.go
naemono: this is a good point.  nice catch.  👍

cmd/manager/main.go
naemono: I think that would be a cleaner approach.  Will do.

deploy/eck-operator/values.yaml
pebrc: I was thinking more of `true`, `false`, `auto-detect` and no translation between Helm and operator parameters. The operator parameters should also be these three if that makes sense. Or is there a benefit of having the "" parameter that I am missing?

naemono: I've updated this to only allow those 3x options.   `setDefaultSecurityContext` being unset, or "" will fail with a syntax error at `strconv.ParseBool(setDefaultSecurityContext)`

cmd/manager/main.go
```diff
@@ -674,6 +684,53 @@ func chooseAndValidateIPFamily(ipFamilyStr string, ipFamilyDefault corev1.IPFami
...
 
+// determineSetDefaultSecurityContext determines what settings we need to use for security context by using the following rules:
+// 1. If the setDefaultSecurityContext is explicitly set to either true, or false, use this value.
+// 2. use openshift detection pulled from kubevirt
+//    https://github.com/kubevirt/kubevirt/blob/f71e9c9615a6c36178169d66814586a93ba515b5/pkg/util/cluster/cluster.go#L21
+//    to determine whether or not we are running within an openshift cluster.  If we determine we are on an openshift cluster,
+//    then default to false, otherwise, return true.
+func determineSetDefaultSecurityContext(setDefaultSecurityContext string, clientset kubernetes.Interface) (bool, error) {
+	if setDefaultSecurityContext == "auto-detect" {
+		return isOpenShift(clientset)
```
pebrc, line 695: I think the behaviour here is the opposite of what we want. We do not want to set the `securityContext` if it is OpenShift (because of randomised UIDs and UID 1000 being forbidden by default).

cmd/manager/main.go
```diff
@@ -674,6 +684,53 @@ func chooseAndValidateIPFamily(ipFamilyStr string, ipFamilyDefault corev1.IPFami
 	}
 }
 
+// determineSetDefaultSecurityContext determines what settings we need to use for security context by using the following rules:
+// 1. If the setDefaultSecurityContext is explicitly set to either true, or false, use this value.
+// 2. use openshift detection pulled from kubevirt
```
pebrc, line 689:
```suggestion
// 2. use OpenShift detection inspired by kubevirt
```
We have made changes to the original detection code. Also I think the attribution should go on the detection function. 

Maybe we can motivate a little bit more why we don't set the `securityContext` if it is OpenShift so that it becomes clearer to future readers why we implemented this behaviour.

cmd/manager/main.go
```diff
@@ -674,6 +684,53 @@ func chooseAndValidateIPFamily(ipFamilyStr string, ipFamilyDefault corev1.IPFami
...
+//    https://github.com/kubevirt/kubevirt/blob/f71e9c9615a6c36178169d66814586a93ba515b5/pkg/util/cluster/cluster.go#L21
+//    to determine whether or not we are running within an openshift cluster.  If we determine we are on an openshift cluster,
```
pebrc, line 691:
```suggestion
//    to determine whether or not we are running within an OpenShift cluster.  If we determine we are on an OpenShift cluster,
```

pebrc (changes requested):

cmd/manager/main.go
naemono: 👍

cmd/manager/main.go
naemono: Completely agree, and added a suite of unit tests to verify this functionality.

cmd/manager/main_test.go
```diff
@@ -192,3 +223,117 @@ func Test_garbageCollectSoftOwnedSecrets(t *testing.T) {
...
+		clientset                 kubernetes.Interface
+	}
+	tests := []struct {
+		name    string
+		args    args
+		want    bool
+		wantErr bool
+	}{
+		{
+			"auto-detect in openshift cluster does not set security context",
```
pebrc, line 250:
```suggestion
			"auto-detect in OpenShift cluster does not set security context",
```

and a couple of times below.

pebrc (commented):
Almost LGTM, a few nits and the build is failing due to the import statements in the new test.

cmd/manager/main.go
```diff
@@ -674,6 +684,54 @@ func chooseAndValidateIPFamily(ipFamilyStr string, ipFamilyDefault corev1.IPFami
...
+//    https://github.com/kubevirt/kubevirt/blob/f71e9c9615a6c36178169d66814586a93ba515b5/pkg/util/cluster/cluster.go#L21
+func isOpenShift(clientset kubernetes.Interface) (bool, error) {
+	openshiftSecurityGroupVersion := schema.GroupVersion{Group: "security.openshift.io", Version: "v1"}
+	apiResourceList, err := clientset.Discovery().ServerResourcesForGroupVersion(openshiftSecurityGroupVersion.String())
+	if err != nil {
+		// in case of an error, check if security.openshift.io is the reason (unlikely).
+		var e *discovery.ErrGroupDiscoveryFailed
+		if ok := errors.As(err, &e); ok {
+			if _, exists := e.Groups[openshiftSecurityGroupVersion]; exists {
+				// If security.openshift.io is the reason for the error, we are absolutely in openshift
```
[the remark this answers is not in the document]
naemono: Honestly, both seem to fit, but I've defaulted to `on` in all comments now.

cmd/manager/main_test.go
```diff
@@ -6,25 +6,56 @@ package manager
...
+	discovery.DiscoveryInterface
+	resources                         []*metav1.APIResourceList
+	errServerResourcesForGroupVersion error
+}
+
+func (c *fakeClientset) Discovery() discovery.DiscoveryInterface {
+	return c.discovery
+}
+
+func (d *fakeDiscovery) ServerResourcesForGroupVersion(groupVersion string) (*metav1.APIResourceList, error) {
```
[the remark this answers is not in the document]
naemono: Gotcha.  After looking through some of the tests in this repo, I'm definitely noticing these required bits for a test going after the tests that use them.  I'll follow this pattern from this point forward.

pebrc (approved):
LGTM

PR #5158, opened 2021-12-14
Resource aggregator should handle missing memory settings

barkbay:
This PR fixes a bug where a user might want to set some JVM parameters without JVM memory settings (`Xmx`). To reproduce create the following Elasticsearch resource:

```yaml
kind: Elasticsearch
metadata:
  name: es
spec:
  version: 6.8.21
  nodeSets:
    - name: master
      config:
        node.master: true
        node.data: true
        node.ingest: true
        node.store.allow_mmap: false
      podTemplate:
        spec:
          containers:
            - name: elasticsearch
              env:
                - name: ES_JAVA_OPTS
                  value: "-Dlog4j2.formatMsgNoLookups=true"
      count: 1
```

Operator logs:

```
2021-12-14T10:23:36.616+0100    ERROR   resource        Failed to report licensing information
{
  "service.version": "XXXXXXX",
     "error": "failed to aggregate Elasticsearch memory: cannot extract max jvm heap size from -Dlog4j2.formatMsgNoLookups=true", "errorVerbose": "cannot extract max jvm heap size from -Dlog4j2.formatMsgNoLookups=true\n..."
}
```

This PR also removes unit tests cases where the JVM would not be able to start (`Unrecognized option`, `Invalid maximum heap size` ...)

### DIFF
--- a/pkg/license/aggregator.go
+++ b/pkg/license/aggregator.go
@@ -208,8 +208,13 @@ var maxHeapSizeRe = regexp.MustCompile(`-Xmx([0-9]+)([gGmMkK]?)(?:\s.*|$)`)
 // memFromJavaOpts extracts the maximum Java heap size from a Java options string, multiplies the value by 2
 // (giving twice the JVM memory to the container is a common thing people do)
 // and converts it to a resource.Quantity
+// If no value is found the function returns the 0 value.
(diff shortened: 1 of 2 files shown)


thbkrkr (approved):
LGTM

PR #5185, opened 2021-12-17
Bump the operator memory limit to 1Gi for larger deployments

### LINKED ISSUE #5177: Default memory limit in operator CSV is too low
The default memory limit for Elastic operator (512Mi) is too low and as a consequence, elastic operator is OOMKilled when reconciling multiple Elastic resources.

```
apiVersion: operators.coreos.com/v1alpha1
kind: ClusterServiceVersion
metadata:
  annotations:
    certified: 'false'
...
spec:
...
                      - name: OPERATOR_IMAGE
                        value: >-
                          registry.connect.redhat.com/elastic/eck-operator@sha256:6555ff4bfd7823f2ab2557b4c44fb07e6227ca9f54a0c960d3a14329ee9462bc
                    image: >-
                      registry.connect.redhat.com/elastic/eck-operator@sha256:6555ff4bfd7823f2ab2557b4c44fb07e6227ca9f54a0c960d3a14329ee9462bc
                    name: manager
                    ports:
                      - containerPort: 9443
                        name: https-webhook
                        protocol: TCP
                    resources:
                      limits:
                        cpu: '1'
                        **memory: 2Gi**
                      requests:
                        cpu: 100m
                        memory: 150Mi
...
```

Increasing the memory limit to 1GI or 2Gi resolves the problem for us. We need to do this on each operator update via OLM, though.

This happened with 1.8.x and also with the new 1.9.0 version.

Please increase the memory limit to at least 1Gi to prevent this issue.

Thanks,

Michal Bukovjan

pebrc:
Fixes #5177

Bumps the memory limit to 1Gi to accommodate larger deployments with a large client cache. The memory limit can be changed already today by the user either directly or in case of OLM through the subscription config. But given that memory limits do not affect the scheduling decision and we know that larger k8s clusters with lots of objects will require more memory in the operator given our current use of cached clients increasing the limit to 1Gi should not hurt and save users the additional task of increasing the limit in many (most?) cases.

### DIFF
--- a/deploy/eck-operator/values.yaml
+++ b/deploy/eck-operator/values.yaml
@@ -32,7 +32,7 @@ imagePullSecrets: []
 resources:
   limits:
     cpu: 1
-    memory: 512Mi
+    memory: 1Gi
   requests:
     cpu: 100m
     memory: 150Mi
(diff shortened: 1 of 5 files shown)


barkbay (approved):
LGTM

david-kow (approved):
LGTM, but I'd also make a change to https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-common-problems.html#k8s-common-problems-operator-oom, in:

> The default memory limit for the operator pod is set to **512 MiB.**

PR #5187, opened 2021-12-20
Remove "all" namespace from managed namespaces

naemono:
Don't append the 'all' namespaces when storage class validation is enabled, as it causes 'unknown namespace' errors in controller-runtime, and cluster-scoped resources are handled properly now in ctrl-runtime.  This will supersede #5058, as it's a much simpler implementation to solve the same issue as the below PR better handles these types of situations in controller-runtime:  

https://github.com/kubernetes-sigs/controller-runtime/pull/1418

Tested with validation of storage class enabled, and successfully resized volumes from 100GB -> 200GB.

## Known Issues

Validating webhooks for things such as ES PV resizing are created "unscoped"

```
- admissionReviewVersions:
  - v1beta1
  clientConfig:
    service:
      name: elastic-operator-webhook
      namespace: elastic-system
      path: /validate-elasticsearch-k8s-elastic-co-v1-elasticsearch
      port: 443
  failurePolicy: Ignore
  matchPolicy: Exact
  name: elastic-es-validation-v1.k8s.elastic.co
  namespaceSelector: {}
  objectSelector: {}
  rules:
  - apiGroups:
    - elasticsearch.k8s.elastic.co
    apiVersions:
    - v1
    operations:
    - CREATE
    - UPDATE
    resources:
    - elasticsearches
    scope: '*'    <--- Here
```

https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/

`The scope field specifies if only cluster-scoped resources ("Cluster") or namespace-scoped resources ("Namespaced") will match this rule. "∗" means that there are no scope restrictions.`

This means that the webhook[s] get resources across the whole cluster, in namespaces that they do not manage, which generates the dreaded

```
"unable to get: elastic/testing-es-masters because of unknown namespace for the cache"
```

There seems to be 2 approaches to handles this situation:

1. Pass along `managedNamespaces` slice to webhook, and if the crud operation is outside of it's managed namespaces, allow it, ~~which brings up the question: "How are the same webhook, with different names, all cluster scoped, and with the same rules (when having the operator installed twice in a cluster, and validating ES objects) handled?  Do all have to validate, or just one?"~~ (validated all webhooks with appropriate rules receive the request, and any can allow/deny)
2. Work to get these webhooks `Namespaced` (which by looking at the operator framework docs, It wasn't clear how exactly to do this)

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -35,6 +35,7 @@ import (
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/manager/signals"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 
 	"github.com/elastic/cloud-on-k8s/pkg/about"
 	agentv1alpha1 "github.com/elastic/cloud-on-k8s/pkg/apis/agent/v1alpha1"
@@ -62,6 +63,7 @@ import (
 	controllerscheme "github.com/elastic/cloud-on-k8s/pkg/controller/common/scheme"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common/tracing"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common/version"
+	commonwebhook "github.com/elastic/cloud-on-k8s/pkg/controller/common/webhook"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/elasticsearch"
 	esclient "github.com/elastic/cloud-on-k8s/pkg/controller/elasticsearch/client"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/elasticsearch/settings"
@@ -817,7 +819,8 @@ func setupWebhook(
 	// setup webhooks for supported types
 	webhookObjects := []interface {
 		runtime.Object
-		SetupWebhookWithManager(manager.Manager) error
+		admission.Validator
+		WebhookPath() string
 	}{
 		&agentv1alpha1.Agent{},
 		&apmv1.ApmServer{},
@@ -831,7 +834,12 @@ func setupWebhook(
 		&emsv1alpha1.ElasticMapsServer{},
 	}
 	for _, obj := range webhookObjects {
-		if err := obj.SetupWebhookWithManager(mgr); err != nil {
+		if err := commonwebhook.SetupValidatingWebhookWithConfig(&commonwebhook.Config{
+			Manager:          mgr,
+			WebhookPath:      obj.WebhookPath(),
+			ManagedNamespace: managedNamespaces,
+			Validator:        obj,
+		}); err != nil {
 			gvk := obj.GetObjectKind().GroupVersionKind()
 			log.Error(err, "Failed to setup webhook", "group", gvk.Group, "version", gvk.Version, "kind", gvk.Kind)
 		}
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -478,11 +478,6 @@ func startOperator(ctx context.Context) error {
 		// The managed cache should always include the operator namespace so that we can work with operator-internal resources.
 		managedNamespaces = append(managedNamespaces, operatorNamespace)
 
-		// Add the empty namespace to allow watching cluster-scoped resources if storage class validation is enabled.
-		if viper.GetBool(operator.ValidateStorageClassFlag) {
-			managedNamespaces = append(managedNamespaces, "")
-		}
-
 		opts.NewCache = cache.MultiNamespacedCacheBuilder(managedNamespaces)
 	}
 
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -565,7 +565,7 @@ func startOperator(ctx context.Context) error {
 	}
 
 	if viper.GetBool(operator.EnableWebhookFlag) {
-		setupWebhook(mgr, params.CertRotation, params.ValidateStorageClass, clientset, exposedNodeLabels)
+		setupWebhook(mgr, params.CertRotation, params.ValidateStorageClass, clientset, exposedNodeLabels, managedNamespaces)
 	}
 
 	enforceRbacOnRefs := viper.GetBool(operator.EnforceRBACOnRefsFlag)
@@ -782,7 +782,8 @@ func setupWebhook(
 	certRotation certificates.RotationParams,
 	validateStorageClass bool,
 	clientset kubernetes.Interface,
-	exposedNodeLabels esvalidation.NodeLabels) {
+	exposedNodeLabels esvalidation.NodeLabels,
+	managedNamespaces []string) {
 	manageWebhookCerts := viper.GetBool(operator.ManageWebhookCertsFlag)
 	if manageWebhookCerts {
 		log.Info("Automatic management of the webhook certificates enabled")
@@ -837,7 +838,7 @@ func setupWebhook(
 	}
 
 	// esv1 validating webhook is wired up differently, in order to access the k8s client
-	esvalidation.RegisterWebhook(mgr, validateStorageClass, exposedNodeLabels)
+	esvalidation.RegisterWebhook(mgr, validateStorageClass, exposedNodeLabels, managedNamespaces)
 
 	// wait for the secret to be populated in the local filesystem before returning
 	interval := time.Second * 1
--- a/pkg/controller/common/webhook/webhook.go
+++ b/pkg/controller/common/webhook/webhook.go
@@ -77,7 +77,7 @@ func (v *validatingWebhook) Handle(_ context.Context, req admission.Request) adm
 
 	if req.Operation == admissionv1.Update {
 		oldObj := v.validator.DeepCopyObject()
-		err = v.decoder.DecodeRaw(req.Object, oldObj)
+		err = v.decoder.DecodeRaw(req.OldObject, oldObj)
 		if err != nil {
 			whlog.Error(err, "decoding old object from webhook request into type (%T)", oldObj)
 			return admission.Errored(http.StatusBadRequest, err)
--- a/pkg/controller/common/webhook/webhook.go
+++ b/pkg/controller/common/webhook/webhook.go
@@ -1,0 +1,87 @@
+package webhook
+
+import (
+	"context"
+	"net/http"
+
+	admissionv1 "k8s.io/api/admission/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+
+	"github.com/elastic/cloud-on-k8s/pkg/utils/set"
+)
+
+// SetupValidatingWebhookWithConfig will register a set of validation functions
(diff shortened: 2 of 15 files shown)


barkbay:
> This means that the webhook[s] get resources across the whole cluster, in namespaces that they do not manage, which generates the dreaded

I don't think it makes a lot of sense to deploy several "restricted" operators and expect one of them to validate resources at the cluster level. Using a single service to validate/mutate resources is a limitation of K8S itself.

> Pass along managedNamespaces slice to webhook, and if the crud operation is outside of it's managed namespaces

👍 we could do that as a best effort. I'm not familiar with the `scope` field, I'm not sure to understand how it works.

naemono:
> 👍 we could do that as a best effort. I'm not familiar with the `scope` field, I'm not sure to understand how it works.

The `scope` field of the webhook configuration allows a webhook to be namespace-scoped, instead of cluster scoped, which would be the right setting if we're managing a set of namespaces, but seems to indicate that the webhook would only work for the namespace that it's within....  I'll test this a bit more and update with my findings...

pebrc:
As I understand the `scope` field it only helps restricting a rule in a webhook to match only on namespaced API resources or cluster scoped API resources, it does not help with what we are trying to do here which is untangle overlapping webhook instances. 

The only mechanism the webhook API has to restrict responsibility to a namespace or multiple namespaces is the `namespaceSelector` property but that AFAIK relies on the namespaces being labeled with something we can select on which is not the case by default. 

So maybe it would make sense to implement the little `managedNamespaces` hack. Another option could be to have a different client for the webhook that is not namespace restricted but that might be a problem from a memory footprint perspective as I would assume would then cache everything twice. We could also think about having a webhook only mode where one instance of the operator would just run the webhook and other instances would be namespace restricted and do the orchestration work.

naemono:
run full pr build

naemono:
@pebrc, @barkbay I've updated this PR to include a change where the webhooks also know what namespaces are being managed, and ignore requests from namespaces that they don't manage.  Let me know if this works.  Thanks.

pkg/controller/elasticsearch/validation/webhook.go
```diff
@@ -55,13 +59,24 @@ func (wh *validatingWebhook) InjectDecoder(d *admission.Decoder) error {
 }
 
 func (wh *validatingWebhook) validateCreate(es esv1.Elasticsearch) error {
+	// If this Elasticsearch instance is not within the set of managed namespaces
+	// for this operator ignore this request.
+	if !slices.Contains(wh.managedNamespaces, es.Namespace) {
```
barkbay, line 64: We should also check that `len(wh.managedNamespaces) > 0` , otherwise the validation is skipped if the operator is watching all the namespaces.

pkg/controller/elasticsearch/validation/webhook.go
```diff
@@ -55,13 +59,24 @@ func (wh *validatingWebhook) InjectDecoder(d *admission.Decoder) error {
 }
 
 func (wh *validatingWebhook) validateCreate(es esv1.Elasticsearch) error {
+	// If this Elasticsearch instance is not within the set of managed namespaces
```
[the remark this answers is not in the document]
naemono: great catch.  change pushed.

pkg/controller/elasticsearch/validation/webhook.go
```diff
@@ -29,11 +30,13 @@ const (
 
 var eslog = ulog.Log.WithName("es-validation")
 
-func RegisterWebhook(mgr ctrl.Manager, validateStorageClass bool, exposedNodeLabels NodeLabels) {
+// RegisterWebhook will register the elasticsearch validating webhook.
+func RegisterWebhook(mgr ctrl.Manager, validateStorageClass bool, exposedNodeLabels NodeLabels, manaedNamespaces []string) {
```
pebrc, line 34:
```suggestion
func RegisterWebhook(mgr ctrl.Manager, validateStorageClass bool, exposedNodeLabels NodeLabels, managedNamespaces []string) {
```
and where used further down ofc

pkg/controller/elasticsearch/validation/webhook.go
```diff
@@ -76,13 +79,21 @@ func (wh *validatingWebhook) validateUpdate(prev esv1.Elasticsearch, curr esv1.E
...
 func (wh *validatingWebhook) Handle(_ context.Context, req admission.Request) admission.Response {
 	es := &esv1.Elasticsearch{}
 	err := wh.decoder.DecodeRaw(req.Object, es)
 	if err != nil {
 		return admission.Errored(http.StatusBadRequest, err)
 	}
 
+	// If this Elasticsearch instance is not within the set of managed namespaces
+	// for this operator ignore this request.
+	if len(wh.managedNamespaces) > 0 && !slices.Contains(wh.managedNamespaces, es.Namespace) {
```
pebrc, line 92: We have

 https://github.com/elastic/cloud-on-k8s/blob/20f522ff9f5fa04f8ccb22ce19edaa6d044a1aa6/pkg/utils/set/set.go#L9

Not that is necessary to use here but it offers a slightly easier to read containment API something like:

```golang
if managedNamespaces.Has(es.Namespace) {
```

cmd/manager/main.go
```diff
@@ -842,7 +838,7 @@ func setupWebhook(
 	}
 
 	// esv1 validating webhook is wired up differently, in order to access the k8s client
-	esvalidation.RegisterWebhook(mgr, validateStorageClass, exposedNodeLabels)
+	esvalidation.RegisterWebhook(mgr, validateStorageClass, exposedNodeLabels, managedNamespaces)
```
pebrc, line 841: We need to cover the other web hooks as well that are not wired up like the Elasticsearch webhook but are using the standard controller-runtime webhook builder. Otherwise the web hooks will still validate resources in namespaces they are not responsible for.

pebrc (commented):
I think we need to cover the other web hooks as well see my comments.   Also something is wrong with GH web UI today I have a hard time leaving review comments that are not markup garbage, I hope I got them all fixed up now.

pkg/controller/elasticsearch/validation/webhook.go
naemono: I updated it to use this package instead, as it certainly enhances the readability.

cmd/manager/main.go
naemono: @pebrc this took some work, and a new common/webooks package to accomplish, but this is now handled in a consistent fashion across all non-elasticsearch webhooks (ES is special, although with a bit of work, we probably could make it fit this new 'mold' for webhooks, let me know if you want me to go down that path)

naemono:
run full pr build

pkg/controller/common/webhook/webhook.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	// If this instance is not within the set of managed namespaces
+	// for this operator ignore this request.
+	if !v.managedNamespaces.Has(obj.GetNamespace()) {
+		return admission.Allowed("")
+	}
+
+	if req.Operation == admissionv1.Create {
+		err = v.validator.ValidateCreate()
```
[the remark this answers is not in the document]
naemono: @pebrc I believe I follow your meaning here.  I've adjusted this, and testing, and this is doing what I would expect it to.

```
{"log.level":"debug","@timestamp":"2022-02-04T18:26:39.184Z","log.logger":"common-webhook","message":"Skipping resource validation","service.version":"2.0.0-SNAPSHOT+e24fa16a","service.type":"eck","ecs.version":"1.4.0","name":"kibana-testing","namespace":"default"}
{"log.level":"debug","@timestamp":"2022-02-04T18:26:39.186Z","log.logger":"controller-runtime.webhook.webhooks","message":"wrote response","service.version":"2.0.0-SNAPSHOT+e24fa16a","service.type":"eck","ecs.version":"1.4.0","webhook":"/validate-kibana-k8s-elastic-co-v1-kibana","code":200,"reason":"object namespace (default) outside of managed namespaces: [elastic elastic-system]","UID":"0bb3b1f4-bba8-4c7b-8bb2-9936b45883a9","allowed":true}
```

Let me know if this is what you were intending.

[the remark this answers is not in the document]
pebrc: Yes sorry for the cryptic comment. That was indeed what I was after.

naemono:
I noticed the ci tests failing yesterday, and thought I resolved them all, but something is now failing in the e2e test.  I'm investigating.

pkg/controller/elasticsearch/validation/webhook.go
```diff
@@ -76,13 +79,21 @@ func (wh *validatingWebhook) validateUpdate(prev esv1.Elasticsearch, curr esv1.E
...
 func (wh *validatingWebhook) Handle(_ context.Context, req admission.Request) admission.Response {
 	es := &esv1.Elasticsearch{}
 	err := wh.decoder.DecodeRaw(req.Object, es)
 	if err != nil {
 		return admission.Errored(http.StatusBadRequest, err)
 	}
 
+	// If this Elasticsearch instance is not within the set of managed namespaces
+	// for this operator ignore this request.
+	if !wh.managedNamespaces.Has(es.Namespace) {
```
pebrc, line 92: I realise we still need the length check with `Count` on `managedNamespaces` here to cover the case where all namespaces are managed by the operator and `managedNamespaces` is the empty set.

naemono: thank you.   I've added this to both this file, and the shared webhook validation.

pkg/controller/common/webhook/webhook.go
```diff
@@ -0,0 +1,99 @@
...
+	if req.Operation == admissionv1.Create {
+		err = obj.ValidateCreate()
+		if err != nil {
+			return admission.Denied(err.Error())
+		}
+	}
+
+	if req.Operation == admissionv1.Update {
+		oldObj := v.validator.DeepCopyObject()
+		err = v.decoder.DecodeRaw(req.Object, oldObj)
```
pebrc, line 80:
```suggestion
		err = v.decoder.DecodeRaw(req.OldObject, oldObj)
```

This is currently a NOOP validation comparing the new object with itself. This tells me that we are missing a unit test or if that's too complicated an e2e test to validate the webhook is actually validating what it is supposed to.

pebrc (commented):
I think it's almost there but the upgrade validation is not working correctly.

pkg/controller/common/webhook/webhook.go
naemono: I've fixed this in the latest commit.  I will add a unit, or e2e test to ensure properly functionality next.  I misread the documentation for `decoder.Decode`, which state `If you want decode the OldObject in the AdmissionRequest, use DecodeRaw.`...

I'll update when this test has been added.

naemono: Unit test added.

pkg/controller/common/webhook/webhook_test.go
```diff
@@ -0,0 +1,227 @@
...
+		validator         admission.Validator
+	}
+	tests := []struct {
+		name   string
+		fields fields
+		req    admission.Request
+		want   admission.Response
+	}{
+		{
+			"create properly validates valid agent, and returns allowed.",
```
pebrc, line 43: I would be in favour of using keys in these structs instead of positional initialisation.

pebrc (approved):
LGTM!

pkg/controller/common/webhook/webhook_test.go
naemono: No problem, and done.  I'll merge this after ensuring all of the checks pass again.

PR #5189, opened 2021-12-21
Add support for transport TLS certificate other/common name suffix

### LINKED ISSUE #5148: Allow users to provide the name pattern of Elasticsearch's transport certs commonName & otherName
When we generate Elasticsearch transport certificates, we set [both the x509 commonName *and* the Elasticsearch otherName extension](https://github.com/elastic/cloud-on-k8s/blob/9be005d2a4b6e26f98ff9c520ea53ae4d6cc9455/pkg/controller/elasticsearch/certificates/transport/csr.go#L79) to a [value that makes sense for us](https://github.com/elastic/cloud-on-k8s/blob/9be005d2a4b6e26f98ff9c520ea53ae4d6cc9455/pkg/controller/elasticsearch/certificates/transport/csr.go#L117): `<pod_name>.node.<cluster_name>.<namespace>.es.local`.

Currently the name pattern chosen here does not have any particular impact on Elasticsearch. Elasticsearch nodes uses that name to identify other nodes, so it is however important for all nodes to have a deterministic name.

Considering we don't offer a way for users to provide their own transport certs (they can provide a CA from which we generate the certs), it would make sense to allow users to customize the name pattern, similar to how they can customize subject alternative names.

Something like:

```yml
kind: Elasticsearch
metadata:
  name: quickstart
  namespace: mynamespace
spec:
  version: 7.16.0
  transport:
    tls:
      # also used as otherName in Elasticsearch certs
      commonNamePattern: ${pod_name}.node.quickstart.cluster.mynamespace.namespace
      subjectAltNames:
      - ip: 1.2.3.4
      - dns: hulk.example.com
```

There are a few values we may consider templating  (so the user does not provide the exact value but rather a placeholder): pod name, Elasticsearch resource name, namespace name. For example: `${pod_name}.node.${elasticsearch_name}.cluster.${namespace_name}.namespace`. Meaning we'd have to figure out a nice templating mechanism: what's the format of those placeholders?

Another approach that may actually be enough for most users is to always prefix the given common name with the Pod name, since we do need one certificate per Pod anyway. In which case users could provide a suffix with no variable templating mechanism:

```yml
kind: Elasticsearch
metadata:
  name: quickstart
  namespace: mynamespace
spec:
  version: 7.16.0
  transport:
    tls:
      # also used as otherName in Elasticsearch certs
      commonNameSuffix: ".node.quickstart.cluster.mynamespace.namespace" # operator will prepend the pod name here
      subjectAltNames:
      - ip: 1.2.3.4
      - dns: hulk.example.com
```

pebrc:
Fixes #5148

### DIFF
--- a/pkg/apis/elasticsearch/v1/elasticsearch_types.go
+++ b/pkg/apis/elasticsearch/v1/elasticsearch_types.go
@@ -151,7 +151,7 @@ type TransportTLSOptions struct {
 	// CommonNameSuffix when defined will be prefixed with the Pod name and used as the common name,
 	// and the first DNSName as well an OtherName required by Elasticsearch in the Subject Alternative Name extension of
 	// each Elasticsearch node's transport TLS certificate.
-	CommonNameSuffix string
+	CommonNameSuffix string `json:"commonNameSuffix,omitempty"`
 	// SubjectAlternativeNames is a list of SANs to include in the generated node transport TLS certificates.
 	SubjectAlternativeNames []commonv1.SubjectAlternativeName `json:"subjectAltNames,omitempty"`
 	// Certificate is a reference to a Kubernetes secret that contains the CA certificate
--- a/pkg/apis/elasticsearch/v1/elasticsearch_types.go
+++ b/pkg/apis/elasticsearch/v1/elasticsearch_types.go
@@ -148,10 +148,11 @@ type TransportConfig struct {
 }
 
 type TransportTLSOptions struct {
-	// CommonNameSuffix when defined will be prefixed with the Pod name and used as the common name,
-	// and the first DNSName as well an OtherName required by Elasticsearch in the Subject Alternative Name extension of
-	// each Elasticsearch node's transport TLS certificate.
-	CommonNameSuffix string `json:"commonNameSuffix,omitempty"`
+	// OtherNameSuffix when defined will be prefixed with the Pod name and used as the common name,
+	// and the first DNSName, as well as an OtherName required by Elasticsearch in the Subject Alternative Name
+	// extension of each Elasticsearch node's transport TLS certificate.
+	// Example: if set to "node.cluster.local", the generated certificate will have its otherName set to "<pod_name>.node.cluster.local".
+	OtherNameSuffix string `json:"otherNameSuffix,omitempty"`
 	// SubjectAlternativeNames is a list of SANs to include in the generated node transport TLS certificates.
 	SubjectAlternativeNames []commonv1.SubjectAlternativeName `json:"subjectAltNames,omitempty"`
 	// Certificate is a reference to a Kubernetes secret that contains the CA certificate
--- a/pkg/apis/elasticsearch/v1/elasticsearch_types.go
+++ b/pkg/apis/elasticsearch/v1/elasticsearch_types.go
@@ -148,6 +148,10 @@ type TransportConfig struct {
 }
 
 type TransportTLSOptions struct {
+	// CommonNameSuffix when defined will be prefixed with the Pod name and used as the common name,
+	// and the first DNSName as well an OtherName required by Elasticsearch in the Subject Alternative Name extension of
+	// each Elasticsearch node's transport TLS certificate.
+	CommonNameSuffix string
 	// SubjectAlternativeNames is a list of SANs to include in the generated node transport TLS certificates.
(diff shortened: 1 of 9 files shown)


pkg/apis/elasticsearch/v1/elasticsearch_types.go
```diff
@@ -148,6 +148,10 @@ type TransportConfig struct {
 }
 
 type TransportTLSOptions struct {
+	// CommonNameSuffix when defined will be prefixed with the Pod name and used as the common name,
+	// and the first DNSName as well an OtherName required by Elasticsearch in the Subject Alternative Name extension of
+	// each Elasticsearch node's transport TLS certificate.
+	CommonNameSuffix string
```
pebrc, line 154: Naming: should this be called `OtherNameSuffix` because this is actually the relevant one (`CommonName` is only populated for legacy reasons)

pkg/controller/elasticsearch/certificates/transport/csr.go
```diff
@@ -113,6 +113,10 @@ func buildGeneralNames(
...
 
 // buildCertificateCommonName returns the CN (and ES othername) entry for a given pod within a stack
-func buildCertificateCommonName(pod corev1.Pod, clusterName, namespace string) string {
-	return fmt.Sprintf("%s.node.%s.%s.es.local", pod.Name, clusterName, namespace)
+func buildCertificateCommonName(pod corev1.Pod, es esv1.Elasticsearch) string {
+	userConfiguredSuffix := es.Spec.Transport.TLS.CommonNameSuffix
+	if userConfiguredSuffix == "" {
+		return fmt.Sprintf("%s.node.%s.%s.es.local", pod.Name, es.Name, es.Namespace)
+	}
+	return fmt.Sprintf("%s.%s", pod.Name, userConfiguredSuffix)
```
pebrc, line 121: Do we want to handle user error here: e.g. `.my.suffix` vs `my.suffix` and try to do "the right thing"?

pkg/apis/elasticsearch/v1/elasticsearch_types.go
```diff
@@ -148,6 +148,10 @@ type TransportConfig struct {
 }
 
 type TransportTLSOptions struct {
+	// CommonNameSuffix when defined will be prefixed with the Pod name and used as the common name,
+	// and the first DNSName as well an OtherName required by Elasticsearch in the Subject Alternative Name extension of
+	// each Elasticsearch node's transport TLS certificate.
+	CommonNameSuffix string `json:"commonNameSuffix,omitempty"`
```
sebgl, line 154: Let's provide an example in the doc maybe?
```
Example: if set as "node.cluster.local", the generated certificate will have their otherName set to "<pod_name>.node.cluster.local".
```

pkg/controller/elasticsearch/certificates/transport/csr.go
sebgl: Hmm good question. I think it's best to not do any magic here (if you manipulate this field you probably know what you're doing).

pkg/apis/elasticsearch/v1/elasticsearch_types.go
sebgl: Yeah I think that makes sense to go with `otherName` 👍

pkg/controller/elasticsearch/certificates/transport/csr.go
sebgl: I initially thought we would not even set the `.` in between by default.
But I don't know if there'd be any case where we wouldn't want a `.` there.

pkg/controller/elasticsearch/certificates/transport/csr.go
```diff
@@ -113,6 +113,10 @@ func buildGeneralNames(
 }
 
 // buildCertificateCommonName returns the CN (and ES othername) entry for a given pod within a stack
```
sebgl, line 115: Let's use this PR as an opportunity to improve this comment?

```suggestion
// buildCertificateCommonName returns the CN (and ES otherName) entry for a given Elasticsearch Pod.
// If the user provided an otherName suffix in the spec, it prepends the pod name to it (<pod_name>.<user-suffix).
// Otherwise, it defaults to <pod_name>.node.<es_name>.es.local.
```

sebgl (approved):
LGTM

PR #5194, opened 2021-12-21
[DOCS] Fix cross-doc links

debadair:
The repository plugins are becoming Elasticsearch modules, so the docs are moving into the ES reference. These hardcoded links to current break the build when those changes are made. Will switch each version to link to 7.17, and then switch master back to current once the changes on the ES side are merged.


nrichers (approved):
LGTM. 🚀

PR #5211, opened 2021-12-27
Isolate operator from http service misconfiguration - Use internal service

### LINKED ISSUE #4394: Isolate the operator from misconfiguration of HTTP service

naemono:
Since #5005 has been open, with active discussion for quite some time, I thought it would be useful to go back to the beginning and look at what we were trying to accomplish in the original issue #4394, which was ensuring that the user can't misconfigure the operator so it can't manage an Elasticsearch cluster.  In the original PR, we chose to go the route of choosing a random pod for communication between the operator, and the ES cluster, which caused its own set of concerns once we both saw the implementation, and tested the consequences of the implementation.  

This PR shows the changes that would be applied to the operator if we chose the other route, which was managing an internal service that does not inherit the selector defined within the custom resources's spec.  This will, at least, give us more information to make a better decision about which route we believe is best to move either solution forward to resolving the initial issue.

To be clear, only one of #5005, or this issue (or none if we so choose) will be merged.  The other will be closed as is.

resolves #4394

### DIFF
--- a/pkg/controller/autoscaling/elasticsearch/driver.go
+++ b/pkg/controller/autoscaling/elasticsearch/driver.go
@@ -93,14 +93,14 @@ func newStatusBuilder(log logr.Logger, autoscalingSpec esv1.AutoscalingSpec) *st
 // Check if the Service is available.
 func (r *ReconcileElasticsearch) isElasticsearchReachable(ctx context.Context, es esv1.Elasticsearch) (bool, error) {
 	defer tracing.Span(&ctx)()
-	externalService, err := services.GetInternalService(r.Client, es)
+	internalService, err := services.GetInternalService(r.Client, es)
 	if apierrors.IsNotFound(err) {
 		return false, nil
 	}
 	if err != nil {
 		return false, tracing.CaptureError(ctx, err)
 	}
-	esReachable, err := services.IsServiceReady(r.Client, externalService)
+	esReachable, err := services.IsServiceReady(r.Client, internalService)
 	if err != nil {
 		return false, tracing.CaptureError(ctx, err)
 	}
--- a/pkg/controller/autoscaling/elasticsearch/driver.go
+++ b/pkg/controller/autoscaling/elasticsearch/driver.go
@@ -93,7 +93,7 @@ func newStatusBuilder(log logr.Logger, autoscalingSpec esv1.AutoscalingSpec) *st
 // Check if the Service is available.
 func (r *ReconcileElasticsearch) isElasticsearchReachable(ctx context.Context, es esv1.Elasticsearch) (bool, error) {
 	defer tracing.Span(&ctx)()
-	externalService, err := services.GetExternalService(r.Client, es)
+	externalService, err := services.GetInternalService(r.Client, es)
 	if apierrors.IsNotFound(err) {
 		return false, nil
 	}
--- a/pkg/controller/elasticsearch/services/services.go
+++ b/pkg/controller/elasticsearch/services/services.go
@@ -92,7 +92,26 @@ func InternalServiceURL(es esv1.Elasticsearch) string {
 // NewExternalService returns the external service associated to the given cluster.
 // It is used by users to perform requests against one of the cluster nodes.
 func NewExternalService(es esv1.Elasticsearch) *corev1.Service {
-	return newServiceWithName(es, ExternalServiceName(es.Name))
+	nsn := k8s.ExtractNamespacedName(&es)
+
+	svc := corev1.Service{
+		ObjectMeta: es.Spec.HTTP.Service.ObjectMeta,
+		Spec:       es.Spec.HTTP.Service.Spec,
+	}
+
+	svc.ObjectMeta.Namespace = es.Namespace
+	svc.ObjectMeta.Name = ExternalServiceName(es.Name)
+
+	labels := label.NewLabels(nsn)
+	ports := []corev1.ServicePort{
+		{
+			Name:     es.Spec.HTTP.Protocol(),
+			Protocol: corev1.ProtocolTCP,
+			Port:     network.HTTPPort,
+		},
+	}
+
+	return defaults.SetServiceDefaults(&svc, labels, labels, ports)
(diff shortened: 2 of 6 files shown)


naemono:
Question for discussion:  In this solution, would we want to only target the master nodes, instead of all nodes?

naemono:
run full pr build

pkg/controller/elasticsearch/services/services.go
```diff
@@ -77,9 +83,30 @@ func ExternalServiceURL(es esv1.Elasticsearch) string {
...
+	return newServiceWithName(es, ExternalServiceName(es.Name))
+}
+
+// NewInternalService returns the internal service associated to the given cluster.
+// It is used by the operator to perform requests against the elasticsearch cluster nodes,
+// and does not inherit the spec defined within the Elasticsearch custom resource,
+// to remove the possibility of the user misconfiguring access to the ES cluster.
+func NewInternalService(es esv1.Elasticsearch) *corev1.Service {
+	svc := newServiceWithName(es, InternalServiceName(es.Name))
+	svc.Spec.Selector = nil
```
pebrc, line 103: This only covers the selector. Other aspects of the service's definition are still inherited from the service template in the custom resource: e.g. type, publishNotReadyAddresses, IPFamilies etc.

naemono: I was leaving `Type`, PublishNotReadyAddresses`, and `IPFamilies` unset as `Type` defaults to `ClusterIP`, `PublishNotReadyAddresses` defaults to false, and `IPFamilies` states:

```
//  This field is usually assigned automatically based on cluster configuration and the
// ipFamilyPolicy field
```

But I've updated all of these to be explicitly set now to remove any ambiguity.

The comments were also updated to include punctuation, and proper capitalization.  I'll be more careful in the future about these.

pebrc (commented):
Looks good to me! Just one thing ...

pkg/controller/elasticsearch/services/services.go
pebrc: Sorry my comment was maybe misleading. When I wrote "control" I did not mean to set each value but to make sure that the users cannot accidentally set them and shoot themselves in the foot.

pkg/controller/elasticsearch/services/services.go
```diff
@@ -95,15 +96,32 @@ func NewExternalService(es esv1.Elasticsearch) *corev1.Service {
...
+				{
+					Name:     es.Spec.HTTP.Protocol(),
+					Protocol: corev1.ProtocolTCP,
+					Port:     network.HTTPPort,
+				},
+			},
+			Selector:                 label.NewLabels(k8s.ExtractNamespacedName(&es)),
+			PublishNotReadyAddresses: false,
+			IPFamilies: []corev1.IPFamily{
+				corev1.IPv4Protocol,
```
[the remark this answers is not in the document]
naemono: 👍  removed.

pkg/controller/elasticsearch/services/services.go
naemono: 👍  no worries.  Explicit is better anyways.

pebrc (approved):
LGTM

pkg/controller/elasticsearch/services/services.go
```diff
@@ -67,19 +68,60 @@ func ExternalServiceName(esName string) string {
...
+					Port:     network.HTTPPort,
+				},
+			},
+			Selector:                 label.NewLabels(k8s.ExtractNamespacedName(&es)),
+			PublishNotReadyAddresses: false,
+		},
+	}
+}
+
+func newServiceWithName(es esv1.Elasticsearch, serviceName string) *corev1.Service {
```
barkbay, line 124: nit: this function seems to be used only in `NewExternalService` I would keep it inlined.

pkg/controller/autoscaling/elasticsearch/driver.go
```diff
@@ -93,7 +93,7 @@ func newStatusBuilder(log logr.Logger, autoscalingSpec esv1.AutoscalingSpec) *st
 // Check if the Service is available.
 func (r *ReconcileElasticsearch) isElasticsearchReachable(ctx context.Context, es esv1.Elasticsearch) (bool, error) {
 	defer tracing.Span(&ctx)()
-	externalService, err := services.GetExternalService(r.Client, es)
+	externalService, err := services.GetInternalService(r.Client, es)
```
barkbay, line 96:
```suggestion
	internalService, err := services.GetInternalService(r.Client, es)
```

barkbay (approved):
LGTM

pkg/controller/elasticsearch/services/services.go
naemono: makes sense. 👍

PR #5172, opened 2021-12-15
Elastic Maps Server: no Elasticsearch connection required since 7.14

pebrc:
As of 7.14 no direct connection to Elasticsearch is needed if the orchestrator can validate the license. This adjusts the documentation accordingly and the examples we ship.

PR #5225, opened 2022-01-04
Elastic Maps Server: must not use standalone in combination with association mode as of 7.14

pebrc:
#5172 missed an important test case: standalone EMS and specifying an association must not be combined or config validation will fail in EMS. This PR tries to remedy this bug and adds also an e2e test case for standalone mode as I should have done already in #5172


naemono:
Makes sense to me. 👍

naemono (approved):

PR #5242, opened 2022-01-12
Do not delete last master-eligible node if other nodes are not up-to-date

### LINKED ISSUE #5241: All master-eligible nodes may be upgraded before all master-ineligible nodes
To reproduce:
1. Create the following deployment:

```yaml
kind: Elasticsearch
metadata:
  name: elasticsearch-sample
spec:
  version: 7.16.1
  updateStrategy:
    changeBudget:
      maxUnavailable: 2
  nodeSets:
  - name: content
    config:
      node.roles: ["data_content"]
      node.store.allow_mmap: false
    count: 2
  - name: master
    config:
      node.roles: ["master"]
      node.store.allow_mmap: false
    count: 1
```

2. Create an index with the following settings:

```json
{
  "settings": {
    "index": {
      "number_of_shards": 2,
      "number_of_replicas": 1
    }
  }
}
```

3. Upgrade the deployment (to `7.16.2` for example)

The master node instance is upgraded/deleted before one of the two data node, which is not able to join the cluster anymore:

```
"stacktrace": ["org.elasticsearch.transport.RemoteTransportException: [elasticsearch-sample-es-master-0][10.48.178.44:9300][internal:cluster/coordination/join]",
"Caused by: java.lang.IllegalStateException: node version [7.16.1] may not join a cluster comprising only nodes of version [7.16.2] or greater",
```

The reason is that the predicate [`do_not_delete_last_master_if_data_nodes_are_not_upgraded`](https://github.com/elastic/cloud-on-k8s/blob/f2fcf2eb474f9dd409d3724a85a1248e02a3c1ce/pkg/controller/elasticsearch/driver/upgrade_predicates.go#L355-L386) only relies on the label `"elasticsearch.k8s.elastic.co/node-data"` to assess if a node is not a master node: https://github.com/elastic/cloud-on-k8s/blob/f2fcf2eb474f9dd409d3724a85a1248e02a3c1ce/pkg/controller/elasticsearch/driver/upgrade_predicates.go#L380

barkbay:
Fix #5241

### DIFF
--- a/pkg/controller/elasticsearch/driver/upgrade_predicates.go
+++ b/pkg/controller/elasticsearch/driver/upgrade_predicates.go
@@ -351,7 +351,7 @@ var predicates = [...]Predicate{
 		},
 	},
 	{
-		// Force an upgrade of all the data nodes before upgrading the last master
+		// Force an upgrade of all the master-ineligible nodes before upgrading the last master-eligible
 		name: "do_not_delete_last_master_if_data_nodes_are_not_upgraded",
 		fn: func(
 			context PredicateContext,
@@ -372,13 +372,13 @@ var predicates = [...]Predicate{
 					return true, nil
 				}
 			}
-			// This is the last master, check if all data nodes are up to date
+			// This is the last master, check if all master-ineligible nodes are up to date
 			for _, pod := range context.toUpdate {
 				if candidate.Name == pod.Name {
 					continue
 				}
-				if label.IsDataNode(pod) {
-					// There's still a data node to update
+				if !label.IsMasterNode(pod) {
+					// There's still some master-ineligible nodes to update
 					return false, nil
 				}
 			}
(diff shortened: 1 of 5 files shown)


pkg/controller/elasticsearch/driver/upgrade_predicates.go
```diff
@@ -351,7 +351,7 @@ var predicates = [...]Predicate{
 		},
 	},
 	{
-		// Force an upgrade of all the data nodes before upgrading the last master
+		// Force an upgrade of all the master-ineligible nodes before upgrading the last master-eligible
 		name: "do_not_delete_last_master_if_data_nodes_are_not_upgraded",
```
pebrc, line 355: Nit: the name of the predicate does not reflect the new behaviour. Non-master nodes could also be ingest only or ML nodes.

pebrc (approved):
LGTM, glad that you found this.

Issue #5228, opened 2022-01-05
Rolling upgrade should handle data tiers


barkbay:
Running a mixed version cluster may lead to some issues, for example with searchable snapshot and ILM, see an example here: https://github.com/elastic/elasticsearch/issues/77007#issuecomment-909076823

The [Elasticsearch documentation](https://www.elastic.co/guide/en/elasticsearch/reference/8.0/rolling-upgrades.html#rolling-upgrades) explicitly states that data nodes should be upgraded in the following order:

1. frozen
2. cold
3. warm
4. hot

For now ECK only ensures that master nodes are upgraded once all the data nodes are up-to-date: https://github.com/elastic/cloud-on-k8s/blob/eb46da9af31fe97348792332ae1afec938181307/pkg/controller/elasticsearch/driver/upgrade_predicates.go#L354-L355
We should introduce a mechanism to upgrade data nodes in the recommended order.

barkbay:
The operator would still allow some new nodes, with a more recent version, to be added/started while nodes in other tiers have not been upgraded yet.
For example it would be the case if the user upgrades the ES version while adding some nodes at the same time. This is because the cluster is first upscaled using the most recent specification: https://github.com/elastic/cloud-on-k8s/blob/721cccfd905f43c345c7957fbfe173ea3ebb6bbf/pkg/controller/elasticsearch/driver/nodes.go#L89-L98

I think we want the user to be able to add some nodes/capacity when requested. It would then mean that, for a given tier, only the `replicas` field in the `StatefulSet` specification can be changed if other tiers with higher priorities are not fully upgraded.

pebrc:
I wonder if it would be acceptable to really interpret this rule as a `should` rule. That means we would pick the Pods to roll respecting this recommended upgrade order in the rolling upgrade part of the code. But if a user would combine a rolling upgrade with a scale up we would accept the suboptimal upgrade order caused by the fact that we scale out first.

PR #5248, opened 2022-01-14
Handle data tiers during rolling upgrades

barkbay:
Fix https://github.com/elastic/cloud-on-k8s/issues/5228 by adding a predicate to control the upgrade order for the data tiers.

In the case of a circular dependency between 2 tiers the predicate does not block an upgrade, there is no guarantee for a particular order in such a situation.


pkg/apis/elasticsearch/v1/elasticsearch_config_test.go
```diff
@@ -175,9 +175,61 @@ func TestConfig_HasRole(t *testing.T) {
...
+		node      *Node
+		wantRoles []NodeRole
+	}{
+		{
+			name:      "master and data",
+			node:      &Node{Roles: []string{"master", "data"}},
+			wantRoles: []NodeRole{MasterRole, DataContentRole, DataRole, DataHotRole, DataWarmRole, DataColdRole, DataFrozenRole},
+		},
+		{
+			name:      "master and data",
```
pebrc, line 193:
```suggestion
			name:      "master and data_content",
```

pkg/apis/elasticsearch/v1/elasticsearch_config.go
```diff
@@ -8,17 +8,17 @@ import (
...
-
-	commonv1 "github.com/elastic/cloud-on-k8s/pkg/apis/common/v1"
 )
 
 type NodeRole string
 
 const (
 	DataColdRole            NodeRole = "data_cold"
 	DataContentRole         NodeRole = "data_content"
+	DataFrozenRole          NodeRole = "data_frozen"
```
pebrc, line 21: Not sure how we missed the frozen role and did not notice for so long. Should we also add the matching label to the pods for consistency?

pkg/controller/elasticsearch/driver/upgrade_predicates.go
```diff
@@ -10,13 +10,98 @@ import (
...
+	case candidate.HasRole(esv1.DataWarmRole):
+		// warm tier must be upgraded after cold and frozen
+		return other.HasRole(esv1.DataColdRole) || other.HasRole(esv1.DataFrozenRole)
+	case candidate.HasRole(esv1.DataColdRole):
+		// cold tier must be upgraded after frozen
+		return other.HasRole(esv1.DataFrozenRole)
+	}
+	// frozen and content have no dependency
+	return false
+}
```
pebrc, line 100: I was wondering wether the dependency logic could live closer to the domain model in the `esv1` package but no strong preference.

pebrc (approved):
Looks good! I like how clean the dependency logic turned out despite being somewhat tricky. 👍

pkg/apis/elasticsearch/v1/elasticsearch_config.go
barkbay: `elasticsearch.k8s.elastic.co/node-data_frozen` added in https://github.com/elastic/cloud-on-k8s/pull/5248/commits/ac81b733e712f96fc0dd6d601a4f4350ba784e2b

PR #5253, opened 2022-01-17
Adjust Agent startup command to Ubuntu base image

### LINKED ISSUE #5250: Elastic Agent has to trust CA of the Elasticsearch used as output
Today, there is no good mechanism offered by Kibana/Agent that would allow to trust a CA of Elasticsearch that Agents are sending data to. While [ca_sha256 setting](https://www.elastic.co/guide/en/kibana/master/fleet-settings-kb.html#_fleet_settings) exists, it's not useful due to https://github.com/elastic/kibana/issues/105400. The alternative that ECK-managed Agents are using right now, is to trust ES CA by copying it to `/etc/pki/ca-trust/source/anchors/` and running `update-ca-trust` which updates CA store for the entire container.

This worked so far, but in recent days Agent base image was changed in `8.0.0` from CentOS to Ubuntu. The path `/etc/pki/ca-trust/source/anchors/` is no longer correct. We need to update Agent container command to be able to handle this.

pebrc:
Fixes #5250 

As stated in the title this adjusts the CA trust script we are running on startup when Agent is associated with an Elasticsearch instance to the new Ubuntu base image Agent is using in 8.x

### DIFF
--- a/pkg/controller/agent/pod.go
+++ b/pkg/controller/agent/pod.go
@@ -6,7 +6,6 @@ package agent
 
 import (
 	"fmt"
-	"github.com/elastic/cloud-on-k8s/pkg/controller/common/version"
 	"hash"
 	"path"
 	"sort"
@@ -29,6 +28,7 @@ import (
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common/defaults"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common/reconciler"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common/tracing"
+	"github.com/elastic/cloud-on-k8s/pkg/controller/common/version"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common/volume"
 	"github.com/elastic/cloud-on-k8s/pkg/utils/k8s"
 	"github.com/elastic/cloud-on-k8s/pkg/utils/maps"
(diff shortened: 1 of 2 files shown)


pkg/controller/agent/pod.go
```diff
@@ -381,15 +387,22 @@ func getAssociatedFleetServer(params Params) (commonv1.Associated, error) {
 	return &fs, nil
 }
 
-func trustCAScript(caPath string) string {
+func trustCAScript(version version.Version, caPath string) string {
+	sharedCAPath := "/etc/pki/ca-trust/source/anchors/"
+	updateCmd := "update-ca-trust"
+	if version.Major == 8 {
```
naemono, line 393: Do we want `>=` 8?  Otherwise, we'd have to update this again when work began on v9.

naemono (commented):
One question.

pkg/controller/agent/pod.go
naemono: Well, I guess it's a question of whether we think that the next version is going to go back to CentOS style path/cmd, as that's how it will currently be handled.

pebrc: No, this is a good observation. Even if there is not other major after 8 we should not box ourselves in here.

naemono (approved):

PR #5261, opened 2022-01-19
Change upgrade path validation for 8.0 to only allow 7.17

### LINKED ISSUE #5258: Validate that upgrade to 8.0 is going through 7.last
Upgrades to Elastic Stack 8.0 will only be possible from the last version of the 7.x series (~ 7.17). We are currently not enforcing or validating this behaviour. The lack of any validation can lead to very tricky situations for our users as "downgrades" are not allowed in ECK. 

1.  user upgrades from a lower version of 7.x say 7.16.0 to 8.0
2. 8.0 Pods keep boot looping due to bootstrap checks in Elasticsearch
3. user tries to go back to 7.16.0
4. ECK validation forbids downgrade in the webhook
5. user disables webhook
6. ECK validation in the reconciliation loop prevents rollout of downgrade as we compare currently deployed Pods with desired version for wire compatibility 
7. user can not successfully take any other action to my knowledge, other than recreating the cluster or manually deleting Pods.

pebrc:
Fixes #5258 

This sets the minimal supported version for 8.0 to 7.17. It also changes the validation on version upgrade to take the `.status.Version` into account instead of the `spec.Version` (which it still uses as a fallback in case we do not have a valid `.status.Version`). The rationale behind that is that we want to make sure that users do update the version of the `spec` in quick succession without waiting for the rollout of the upgrade to complete thereby unwittingly circumventing the version restriction.

### DIFF
--- a/pkg/controller/elasticsearch/validation/validations.go
+++ b/pkg/controller/elasticsearch/validation/validations.go
@@ -272,16 +272,9 @@ func noDowngrades(current, proposed esv1.Elasticsearch) field.ErrorList {
 
 func validUpgradePath(current, proposed esv1.Elasticsearch) field.ErrorList {
 	var errs field.ErrorList
-	// if available use the status version which reflects the lowest version currently running in the cluster
-	currentVer, err := version.Parse(current.Status.Version)
-	if err != nil {
-		// let's swallow that error which could be caused because we do not have a version in status yet and fall back
-		// to the Spec version which is better than nothing
-		currentVer, err = version.Parse(current.Spec.Version)
-		if err != nil {
-			// this should not happen, since this is the already persisted version
-			errs = append(errs, field.Invalid(field.NewPath("spec").Child("version"), current.Spec.Version, parseStoredVersionErrMsg))
-		}
+	currentVer, ferr := currentVersion(current)
+	if ferr != nil {
+		errs = append(errs, ferr)
 	}
 
 	proposedVer, err := version.Parse(proposed.Spec.Version)
@@ -303,6 +296,26 @@ func validUpgradePath(current, proposed esv1.Elasticsearch) field.ErrorList {
 		errs = append(errs, field.Invalid(field.NewPath("spec").Child("version"), proposed.Spec.Version, unsupportedUpgradeMsg))
 	}
 	return errs
+}
+
+func currentVersion(current esv1.Elasticsearch) (version.Version, *field.Error) {
+	// if available use the status version which reflects the lowest version currently running in the cluster
+	if current.Status.Version == "" {
+		currentVer, err := version.Parse(current.Spec.Version)
+		if err != nil {
+			// this should not happen, since this is the version from the spec copied to the status by the operator
+			return version.Version{}, field.Invalid(field.NewPath("spec").Child("version"), current.Spec.Version, parseStoredVersionErrMsg)
+		}
+		return currentVer, nil
+	}
+	// we do not have a version in the status let's use the version in the current spec instead which will not reflect
+	// actually runnning Pods but which is still better than no validation. 
+	currentVer, err := version.Parse(current.Status.Version)
+	if err != nil {
+		// this should not happen, since this is the already persisted version
+		return version.Version{}, field.Invalid(field.NewPath("status").Child("version"), current.Status.Version, parseStoredVersionErrMsg)
+	}
+	return currentVer, nil
 }
 
 func validMonitoring(es esv1.Elasticsearch) field.ErrorList {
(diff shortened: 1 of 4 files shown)


pkg/controller/elasticsearch/validation/validations.go
```diff
@@ -272,11 +272,18 @@ func noDowngrades(current, proposed esv1.Elasticsearch) field.ErrorList {
...
 func validUpgradePath(current, proposed esv1.Elasticsearch) field.ErrorList {
 	var errs field.ErrorList
-	currentVer, err := version.Parse(current.Spec.Version)
+	// if available use the status version which reflects the lowest version currently running in the cluster
+	currentVer, err := version.Parse(current.Status.Version)
 	if err != nil {
-		// this should not happen, since this is the already persisted version
-		errs = append(errs, field.Invalid(field.NewPath("spec").Child("version"), current.Spec.Version, parseStoredVersionErrMsg))
+		// let's swallow that error which could be caused because we do not have a version in status yet and fall back
+		// to the Spec version which is better than nothing
```
sebgl, line 279: Should we rather check if the version is empty (`""`)? If get a parsing error I would expect we log that error explicitly as a real error.

pebrc: Absolutely. That sounds cleaner, even though I suspect it will not make much difference in practice.

pkg/controller/elasticsearch/validation/validations.go
```diff
@@ -272,11 +272,11 @@ func noDowngrades(current, proposed esv1.Elasticsearch) field.ErrorList {
 
 func validUpgradePath(current, proposed esv1.Elasticsearch) field.ErrorList {
 	var errs field.ErrorList
-	currentVer, err := version.Parse(current.Spec.Version)
-	if err != nil {
-		// this should not happen, since this is the already persisted version
-		errs = append(errs, field.Invalid(field.NewPath("spec").Child("version"), current.Spec.Version, parseStoredVersionErrMsg))
+	currentVer, ferr := currentVersion(current)
+	if ferr != nil {
```
sebgl, line 276: `ferr`?

sebgl (approved):
LGTM

pkg/controller/elasticsearch/validation/validations.go
pebrc: `fear` for `field.Error` which is what the webhook API works with, `err` is used as a variable for a regular builtin `error` below

naemono (approved):
LGTM as well, but your non-deterministic `TestDelayingQueueInvariants` tests are failing you in ci.

pebrc:
Jenkins test this please


TestDelayingQueueInvariants is flaky sometimes unfortunately

pkg/controller/elasticsearch/version/supported_versions_test.go
```diff
@@ -58,7 +58,7 @@ func TestSupportedVersions(t *testing.T) {
 				v: version.MustParse("8.0.0"),
 			},
 			supported: []version.Version{
-				version.MustParse("7.4.0"),
+				version.MustParse("7.17.0"),
```
barkbay, line 61: nit: maybe add a unit test to exercise the other code path:

```golang
		{
			name: "not yet fully upgraded",
			current: esv1.Elasticsearch{
					Namespace: "default",
					Name:      "foo",
				},
				Spec: esv1.ElasticsearchSpec{Version: "7.17.0"},
				Status: esv1.ElasticsearchStatus{
					Version: "7.16.2",
				},
			},
			proposed:     es("8.0.0"),
			expectErrors: true, // still running 7.16.2
		},
```

pkg/controller/elasticsearch/validation/validations.go
```diff
@@ -298,6 +298,26 @@ func validUpgradePath(current, proposed esv1.Elasticsearch) field.ErrorList {
 	return errs
 }
 
+func currentVersion(current esv1.Elasticsearch) (version.Version, *field.Error) {
+	// we do not have a version in the status let's use the version in the current spec instead which will not reflect
+	// actually running Pods but which is still better than no validation.
+	if current.Status.Version == "" {
+		currentVer, err := version.Parse(current.Spec.Version)
+		if err != nil {
+			// this should not happen, since this is the version from the spec copied to the status by the operator
```
barkbay, line 307: Is it not the version from the stored specification ? (which might be wrong if the validation webhook has been disabled at some point)

pebrc: I swapped the logic around and forgot to move the comment 😞

barkbay (approved):
LGTM

PR #5269, opened 2022-01-20
[2.0] Change upgrade path validation for 8.0 to only allow 7.17 (#5261)

pebrc:
 - Change upgrade path validation for 8.0 to only allow 7.17 (#5261)


barkbay (approved):
LGTM

pebrc:
cla/check

PR #5272, opened 2022-01-21
Allow users to disable version validation to downgrade clusters 

### LINKED ISSUE #5259: Allow users to force a downgrade
If users accidentally upgrade to version that does not exist or to a version that is not on a valid upgrade path from the currently running version the have no recourse to get back into a working state at the moment. 

We could introduce an annotation on the Elasticsearch resource for example `eck.k8s.elastic.co/force-downgrade` that would do the following:

1. disable the validation that prevents downgrades
2. do not run the wire format compatibility check comparing the running Pods with the desired Elasticsearch version

Concerns: 
* who is going to remove the annotation once that process is complete?
* should we built in additional safeguards if the Pods about to be downgraded are already running and it can be assumed that the data directory of the node has also been upgraded to the new version?

pebrc:
Fixes #5259 

I tried to go with a different name for the annotation `permissive-version-validation`, but I am not happy yet because we are not permissive on all version validations only on the one that prevents downgrades. So maybe we should go back to the proposal in the original issue. 

This is intended as a troubleshooting tool so I am going to only document it in the troubleshooting section. 

I am currently aware of two use cases: 

1. violation of upgrade path: user upgraded to 8.0 from < 7.17 on an older version of ECK and is now stuck because 8.0 upgrades have to go through 7.17. 
2. user specified a non-existing version. 

In both cases it should suffice to set the annotation and revert to the previous Elasticsearch version will get the user back into a working state.


pkg/controller/elasticsearch/driver/driver.go
```diff
@@ -188,7 +188,9 @@ func (d *defaultDriver) Reconcile(ctx context.Context) *reconciler.Results {
 	// always update the elasticsearch state bits
 	d.ReconcileState.UpdateElasticsearchState(*resourcesState, observedState())
 
-	if err := d.verifySupportsExistingPods(resourcesState.CurrentPods); err != nil {
+	_, permissive := d.ES.Annotations[esv1.PermissiveVersionValidation]
+
+	if err := d.verifySupportsExistingPods(resourcesState.CurrentPods); !permissive && err != nil {
```
barkbay, line 193: It feels a bit odd to me to have a validation function outside of the `validation` package. I think I would expect them to be placed in the same one. It would allow to have a consistent code path where `verifySupportsExistingPods` would be part of a common validation logic together with `noDowngrades` _(maybe with `verifySupportsExistingPods` disabled in the context of the validation webhook if we don't want to use the cached client)._

That being said I understand that it is not the purpose of this PR and I think I'm fine with the current proposal.

pebrc: I agree. I tried to minimise the changes in this PR given that we want to back port it. How about I open an issue and we fix this in main?

barkbay:
> I tried to go with a different name for the annotation `permissive-version-validation`, but I am not happy yet because we are not permissive on all version validations only on the one that prevents downgrades.

What about `disable-downgrade-validation` ?

pebrc:
> What about disable-downgrade-validation ?

Sounds good to me. I changed it.

pebrc:
I just retested on a new 8.0 snapshot and the `nodes` file a saw last week is no longer being created it seems. I am investigating what is going on but if so we can drop the last commit.

pebrc:
I removed the instructions requiring the additional suspend step which is no longer necessary. I still trying to understand what has changed in Elasticsearch but I guess it makes ECK users' life easier.

pkg/controller/elasticsearch/driver/driver.go
```diff
@@ -188,7 +188,9 @@ func (d *defaultDriver) Reconcile(ctx context.Context) *reconciler.Results {
 	// always update the elasticsearch state bits
 	d.ReconcileState.UpdateElasticsearchState(*resourcesState, observedState())
 
-	if err := d.verifySupportsExistingPods(resourcesState.CurrentPods); err != nil {
+	_, permissive := d.ES.Annotations[esv1.DisableDowngradeValidationAnnotation]
```
barkbay, line 191: I wonder if we should try to detect when the user set `false`, and expect the validation to be enabled (similar to what we do for `eck.k8s.elastic.co/managed`)

pkg/controller/elasticsearch/driver/driver.go
```diff
@@ -188,7 +188,9 @@ func (d *defaultDriver) Reconcile(ctx context.Context) *reconciler.Results {
...
+
+	if err := d.verifySupportsExistingPods(resourcesState.CurrentPods); !permissive && err != nil {
```
barkbay, line 193: If `permissive` is `true` and, if the validation fails, I think I would still log the validation result and mention that the downgrade is allowed as requested by the user. It might help to understand some situations afterwards by reading the logs.

barkbay (approved):
LGTM

pkg/controller/elasticsearch/driver/driver.go
pebrc: Good point, will change it.

docs/operating-eck/troubleshooting/common-problems.asciidoc
```diff
@@ -214,3 +214,10 @@ elastic-cloud-eck.v1.4.0       Elasticsearch (ECK) Operator   1.4.0       elasti
 If you are using one of the affected versions of OLM and upgrading OLM to a newer version is not possible then ECK
 can still be upgraded by uninstalling and reinstalling it. This can be done by removing the `Subscription` and both `ClusterServiceVersion` resources and adding them again.
 On OpenShift the same workaround can be performed in the UI by clicking on "Uninstall Operator" and then reinstalling it through OperatorHub.
+
+[id="{p}-{page_id}-version-downgrade"]
+== If you upgraded Elasticsearch to the wrong version
+If you accidentally upgrade one of your Elasticsearch clusters to a version that does not exist or a version to which a direct upgrade is not possible from your currently deployed version, a validation will prevent you from going back to the previous version.
+The reason for this validation is that ECK will not allow downgrades as this is not supported by Elasticsearch and once the data directory of Elasticsearch has been upgraded there is no way back to the old version.
```
naemono, line 221: Do we maybe want to say `there is no way back to the old version without a restore`?  Also, do we want to point to the Elasticsearch docs here?

docs/operating-eck/troubleshooting/common-problems.asciidoc
```diff
@@ -214,3 +214,10 @@ elastic-cloud-eck.v1.4.0       Elasticsearch (ECK) Operator   1.4.0       elasti
...
+
+The two scenarios described above however are exceptions because Elasticsearch never started up successfully. If you annotate the Elasticsearch resource with `k8s.eck.elastic.co/disable-downgrade-validation` ECK will allow you to go back to the old version at your own risk. Please remove the annotation afterwards to prevent accidental downgrades and reduced availability.
```
naemono, line 223: We mention annotating the ES resource with `k8s.eck.elastic.co/disable-downgrade-validation`, but the validation below also checks the `value == "true"`, and not just the existence of the key.

docs/operating-eck/troubleshooting/common-problems.asciidoc
pebrc: Good point I added this.

docs/operating-eck/troubleshooting/common-problems.asciidoc
pebrc: Yes I forgot to update the documentation when I changed the behaviour 😄

Issue #5078, opened 2021-11-25
Support secretName to define an external Elasticsearch in associations


thbkrkr:
Motivation: makes possible to associate any Stack components with an external Elasticsearch, in particular use an Elastic Cloud cluster to monitor ECK clusters via the Stack Monitoring feature but all the other components would benefit from it.

All Stack components (e.g.: Kibana, Beats, ES with Stack Monitoring) that can associated with an Elasticsearch have an `elasticsearchRef` field to define the targeted Elasticsearch. This is convenient but it requires that the Elasticsearch cluster is managed by ECK.
Some Stack components support the association to an Elasticsearch not managed by ECK (e.g.: managed by [Elastic Cloud](https://cloud.elastic.co/)), but to do so, you cannot use `elasticsearchRef` and have to follow the documentation as each component as its own configuration ([Kibana](https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-kibana-es.html#k8s-kibana-external-es), [APM Server](https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-apm-advanced-configuration.html#k8s-apm-existing-es), [EnterpriseSearch](https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-enterprise-search-configuration.html#k8s-enterprise-search-connect-non-eck-es)). And for Beats, Fleet, Agent, EMS and ES with Stack Monitoring, this is completely not supported.

To support this for all associations, we could add a new field `secretName` to `elasticsearchRef`, which supersedes all the others fields, to point to a secret that contains information to connect to the targeted Elasticsearch.

```yaml
spec:
  elasticsearchRef:
    secretName: "external-es-cluster01-endpoint"
```

```sh
> kubectl get secret external-es-cluster01-endpoint -o json | jq '.data'
{
	username: 	"xxxxxxxx",
	password: 	"yyyyyyyy",
	hosts: 		["zzzzzzzz"],
	ca_cert:    "WWWWWWWWWWWWW"
}
```

thbkrkr:
What are the impacts on the existing? It seems quite reasonable, see this ([draft](https://github.com/elastic/cloud-on-k8s/pull/5240/commits/436a683b70c23d61c67ecea8c43ac191dc8a21d0)):
- watchers: new dynamic watcher for the custom referenced secret
- assocation reconciliation loop: 
  - use a `Secret` instead of the type returned by `ReferencedObjTemplate()` to check if the ref object exists
[...]

PR #5277, opened 2022-01-24
Use only ca.crt in associations

thbkrkr:
While working on #5078, I noticed that the associations `ent-es` and `ems-es` use `tls.crt`, while all the other associations use `ca.crt` and also that we use the content of the `tls.crt` when we write the CA content to the config hash. Stop me if I'm wrong but it seems to me that there is no particular reason for this. This commit changes that so that only `ca.crt` is used in associations.


naemono:
If/when this is released, are there any issues migration from one to the other?  Are there any scenarios where this cause any downtime for users @thbkrkr ?

thbkrkr:
> If/when this is released, are there any issues migration from one to the other? Are there any scenarios where this cause any downtime for users @thbkrkr ?

Good question. ECK triggers a rolling upgrade of the EnterpriseSearch or ElasticMapServer resources without downtime.

 I tested an update from ECK `1.9.1` to this version and it was :ok_hand:.

thbkrkr:
The last commit is about the certificate file used to calculate the configuration hash to rotate the pods when the certificate changes, which I had forgotten.

I updated the description with a good side effect that I didn't think about. It avoids recreating pods for `EnterpriseSearch` and `ElasticMapsServer` resources when the http service is of type `LoadBalancer`. Currently, as soon as the LB is created, the certificate `tls.crt` is regenerated with the IP of the LB, and it was this one that was used in the config hash, so this triggered a pod recreation, while the CA `ca.crt` is kept intact.

pebrc (approved):
I think you are right that including the CA in the EMS and Enterprise Search configuration is the right thing to do. I don't think there is a need to look at the leaf certificate for the associated resource like Elasticsearch but we should prefer the CA. That should mean less restarts and less disruption of service. I think it was probably just an oversight. I tested both EMS and Enterprise Search and an upgrade from 1.9.1 and I think this is OK. 👍

pebrc:
We realised after merging this that things are not that easy especially when using custom certificates where the CA cert is optional. We will need to revisit this.

PR #5280, opened 2022-01-25
[2.0] Allow users to disable version validation to downgrade clusters  (#5272)

pebrc:
 - Allow users to disable version validation to downgrade clusters  (#5272)


barkbay (approved):
LGTM

PR #5283, opened 2022-01-25
E2E: remove noop upgrade from EMS test

pebrc:
We were trying a NOOP upgrade in the 8x test from 8.0 to 8.0 that created an impossible condition ie the test runner waiting for an upgrade that was not one. 

I removed the bit that upgrades Elasticsearch from this test as we want to test the EMS upgrade here (we test Elasticsearch upgrades in other tests already)


barkbay (approved):
Tested on my e2e environment:
* TestElasticMapsServerVersionUpgradeToLatest8x ✅ 
* TestElasticMapsServerVersionUpgradeToLatest7x ✅ 

LGTM

PR #5284, opened 2022-01-25
Allow predicates to disabled on a case-by-case basis via annotation.

### LINKED ISSUE #2092: Allow unsafe rolling upgrades if specified by the user
Currently we make sure the cluster has a green health before moving on with rolling upgrades.
There are cases where the cluster cannot be green:

* because there is a single node in the cluster, and it has indices with 1 replica (default).
* because there is a problem with some shards, from which we won't recover

In these situations it could make sense to "force" the upgrade, even though it may cause unavailability on the cluster (and a risk of data loss). At the user's own risks, basically. It could be expressed through a particular field in the `updateStrategy` of the Elasticsearch manifest.

naemono:
Closes #2092 

This change will allow ECK user to disable certain upgrade predicates, by name, via an annotation on the Elasticsearch CRD, thus allowing clusters to be "forced" to be upgraded in scenarios where the default predicates (that disallow upgrade because of a set of rules) are too restrictive.

Use Case 1:

Cluster is a "red" state because of an un-assignable shard.  This cluster would never be allowed to be upgraded with the standard set of upgrade predicates in place.

If the following annotation was added to the cluster

```yaml
kind: Elasticsearch
metadata:
  name: testing
  annotations:
    eck.k8s.elastic.co/disable-upgrade-predicates: "only_restart_healthy_node_if_green_or_yellow"
spec:
  version: 7.15.1
```

And the version was bumped to `7.15.2`, then this cluster's nodes would be allowed to be upgraded in this scenario, even with the cluster in a "red" state.

Use Case 2:

A test Elasticsearch cluster has been setup, and regardless of any state of the cluster, a user wants to force an upgrade.

With the following annotation, and a version bump to `7.15.2`, all predicates will be disabled, allowing an upgrade to proceed regardless of the state of the cluster:

```yaml
kind: Elasticsearch
metadata:
  name: testing
  annotations:
    eck.k8s.elastic.co/disable-upgrade-predicates: "*"
spec:
  version: 7.15.1
```

Discussion Points

* Do we want to allow anything other than disabling one predicate, or disabling all predicates?  Such as disabling 2 or more predicates specifically?  The way this is designed at the moment, this wouldn't be allowed.

Todo
- [ ] Write documentation

### DIFF
--- a/pkg/controller/elasticsearch/driver/fixtures.go
+++ b/pkg/controller/elasticsearch/driver/fixtures.go
@@ -116,11 +116,12 @@ func newUpgradeTestPods(pods ...testPod) upgradeTestPods {
 	return result
 }
 
-func (u upgradeTestPods) toES(version string, maxUnavailable int) esv1.Elasticsearch {
+func (u upgradeTestPods) toES(version string, maxUnavailable int, annotations map[string]string) esv1.Elasticsearch {
 	return esv1.Elasticsearch{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      TestEsName,
-			Namespace: TestEsNamespace,
+			Name:        TestEsName,
+			Namespace:   TestEsNamespace,
+			Annotations: annotations,
 		},
 		Spec: esv1.ElasticsearchSpec{
 			Version: version,
@@ -159,7 +160,7 @@ func (u upgradeTestPods) toStatefulSetList() sset.StatefulSetList {
 	return statefulSetList
 }
 
-func (u upgradeTestPods) toRuntimeObjects(version string, maxUnavailable int, f filter) []runtime.Object {
+func (u upgradeTestPods) toRuntimeObjects(version string, maxUnavailable int, f filter, annotations map[string]string) []runtime.Object {
 	var result []runtime.Object
 	for _, testPod := range u {
 		pod := testPod.toPod()
@@ -167,7 +168,7 @@ func (u upgradeTestPods) toRuntimeObjects(version string, maxUnavailable int, f 
 			result = append(result, &pod)
 		}
 	}
-	es := u.toES(version, maxUnavailable)
+	es := u.toES(version, maxUnavailable, annotations)
 	result = append(result, &es)
 	return result
 }
--- a/pkg/controller/elasticsearch/driver/upgrade_pods_deletion.go
+++ b/pkg/controller/elasticsearch/driver/upgrade_pods_deletion.go
@@ -50,7 +50,7 @@ func (ctx *rollingUpgradeCtx) Delete() ([]corev1.Pod, error) {
 		"maxUnavailableReached", maxUnavailableReached,
 		"allowedDeletions", allowedDeletions,
 	)
-	podsToDelete, err := applyPredicates(predicateContext, candidates, maxUnavailableReached, allowedDeletions)
+	podsToDelete, err := applyPredicates(predicateContext, candidates, ctx.ES.Annotations, maxUnavailableReached, allowedDeletions)
 	if err != nil {
 		return podsToDelete, err
 	}
@@ -186,6 +186,7 @@ func runPredicates(
 	ctx PredicateContext,
 	candidate corev1.Pod,
 	deletedPods []corev1.Pod,
+	annotations map[string]string,
 	maxUnavailableReached bool,
 ) (*failedPredicate, error) {
 	for _, predicate := range predicates {
@@ -194,6 +195,13 @@ func runPredicates(
 			return nil, err
 		}
 		if !canDelete {
+			// if this specific predicate name is disabled by the disable predicate annotation
+			// "eck.k8s.elastic.co/disable-upgrade-predicates", then ignore this predicate,
+			// and continue processing the remaining predicates.
+			if predicateDisabled(predicate, annotations) {
+				log.V(0).Info("disabling upgrade predicate because of annotation", "predicate", predicate.name)
+				continue
+			}
 			// Skip this Pod, it can't be deleted for the moment
 			return &failedPredicate{
 				pod:       candidate.Name,
@@ -203,4 +211,14 @@ func runPredicates(
 	}
 	// All predicates passed!
 	return nil, nil
+}
+
+// predicateDisabled determines whether a given predicate has been disabled by providing the
+// disable predicate annotation "eck.k8s.elastic.co/disable-upgrade-predicates", and either providing
+// the predicate name as the value to the annotations map, or providing "*" as the value to disable all predicates.
+func predicateDisabled(predicate Predicate, annotations map[string]string) bool {
+	if val, ok := annotations[disableUpgradePredicatesAnnotation]; ok && (predicate.name == val || val == "*") {
+		return true
+	}
+	return false
 }
--- a/pkg/controller/elasticsearch/driver/upgrade_predicates.go
+++ b/pkg/controller/elasticsearch/driver/upgrade_predicates.go
@@ -15,6 +15,22 @@ import (
 	"github.com/elastic/cloud-on-k8s/pkg/utils/stringsutil"
 )
 
+const (
+	// disableUpgradePredicatesAnnotation is the annontation that can be applied to an
+	// Elasticsearch cluster to disable certain predicates during rolling upgrades.
+	//
+	// Example:
+	//
+	//   To disable "if_yellow_only_restart_upgrading_nodes_with_unassigned_replicas" predicate
+	//
+	//   metadata:
+	//     annotations:
+	//       eck.k8s.elastic.co/disable-upgrade-predicates="if_yellow_only_restart_upgrading_nodes_with_unassigned_replicas"
+	disableUpgradePredicatesAnnotation = "eck.k8s.elastic.co/disable-upgrade-predicates"
+)
+
+// PredicateContext is the set of fields used while determining what set of pods
+// can be upgraded when performing a rolling upgrade on an Elasticsearch cluster.
 type PredicateContext struct {
 	es                     esv1.Elasticsearch
 	masterNodesNames       []string
@@ -52,6 +68,8 @@ func groupByPredicates(fp failedPredicates) map[string][]string {
 	return podsByPredicates
 }
 
+// NewPredicateContext will return a new predicate context for use when
+// processing an Elasticsearch rolling upgrade.
 func NewPredicateContext(
 	ctx context.Context,
 	es esv1.Elasticsearch,
@@ -76,12 +94,12 @@ func NewPredicateContext(
 	}
 }
 
-func applyPredicates(ctx PredicateContext, candidates []corev1.Pod, maxUnavailableReached bool, allowedDeletions int) (deletedPods []corev1.Pod, err error) {
+func applyPredicates(ctx PredicateContext, candidates []corev1.Pod, annotations map[string]string, maxUnavailableReached bool, allowedDeletions int) (deletedPods []corev1.Pod, err error) {
 	var failedPredicates failedPredicates
 
 Loop:
 	for _, candidate := range candidates {
-		switch predicateErr, err := runPredicates(ctx, candidate, deletedPods, maxUnavailableReached); {
+		switch predicateErr, err := runPredicates(ctx, candidate, deletedPods, annotations, maxUnavailableReached); {
 		case err != nil:
 			return deletedPods, err
 		case predicateErr != nil:
--- a/pkg/controller/elasticsearch/driver/upgrade_predicates_test.go
+++ b/pkg/controller/elasticsearch/driver/upgrade_predicates_test.go
@@ -27,10 +27,10 @@ func TestUpgradePodsDeletion_WithNodeTypeMutations(t *testing.T) {
 	type fields struct {
 		esVersion       string
 		upgradeTestPods upgradeTestPods
-		ES              esv1.Elasticsearch
 		health          client.Health
 		mutation        mutation
 		maxUnavailable  int
+		esAnnotations   map[string]string
 	}
 	tests := []struct {
 		name                         string
@@ -141,8 +141,8 @@ func TestUpgradePodsDeletion_WithNodeTypeMutations(t *testing.T) {
 			health:    tt.fields.health,
 		}
 		esClient := &fakeESClient{version: version.MustParse("7.13.0")}
-		es := tt.fields.upgradeTestPods.toES(tt.fields.esVersion, tt.fields.maxUnavailable)
-		k8sClient := k8s.NewFakeClient(tt.fields.upgradeTestPods.toRuntimeObjects(tt.fields.esVersion, tt.fields.maxUnavailable, nothing)...)
+		es := tt.fields.upgradeTestPods.toES(tt.fields.esVersion, tt.fields.maxUnavailable, tt.fields.esAnnotations)
+		k8sClient := k8s.NewFakeClient(tt.fields.upgradeTestPods.toRuntimeObjects(tt.fields.esVersion, tt.fields.maxUnavailable, nothing, tt.fields.esAnnotations)...)
 		ctx := rollingUpgradeCtx{
 			parentCtx:       context.Background(),
 			client:          k8sClient,
@@ -180,11 +180,11 @@ func TestUpgradePodsDeletion_Delete(t *testing.T) {
 		upgradeTestPods upgradeTestPods
 		shardLister     client.ShardLister
 		shutdowns       map[string]client.NodeShutdown
-		ES              esv1.Elasticsearch
 		health          client.Health
 		maxUnavailable  int
 		podFilter       filter
 		esVersion       string
+		esAnnotations   map[string]string
 	}
 	tests := []struct {
 		name                         string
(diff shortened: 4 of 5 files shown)


naemono:
run full pr build

pkg/controller/elasticsearch/driver/upgrade_predicates.go
```diff
@@ -15,6 +15,22 @@ import (
 	"github.com/elastic/cloud-on-k8s/pkg/utils/stringsutil"
 )
 
+const (
+	// DisableUpgradePredicatesAnnotation is the annontation that can be applied to an
```
barkbay, line 19:
```suggestion
	// DisableUpgradePredicatesAnnotation is the annotation that can be applied to an
```

pkg/controller/elasticsearch/driver/upgrade_pods_deletion.go
```diff
@@ -194,6 +195,13 @@ func runPredicates(
 			return nil, err
 		}
 		if !canDelete {
+			// if this specific predicate name is disabled by the disable predicate annotation
+			// "eck.k8s.elastic.co/disable-upgrade-predicates", then ignore this predicate,
+			// and continue processing the remaining predicates.
+			if predicateDisabled(predicate, annotations) {
+				log.V(0).Info("disabling upgrade predicate because of annotation", "predicate", predicate.name)
```
barkbay, line 202:
```suggestion
				log.Info("Warning: disabling upgrade predicate because of annotation", "predicate", predicate.name, "namespace", ctx.es.Namespace, "es_name", ctx.es.Name)
```

pkg/controller/elasticsearch/driver/fixtures.go
```diff
@@ -116,11 +116,12 @@ func newUpgradeTestPods(pods ...testPod) upgradeTestPods {
 	return result
 }
 
-func (u upgradeTestPods) toES(version string, maxUnavailable int) esv1.Elasticsearch {
+func (u upgradeTestPods) toES(version string, maxUnavailable int, annotations map[string]string) esv1.Elasticsearch {
```
barkbay, line 119: We could also get the disabled predicates from the Elasticsearch resource, using a receiver: it would avoid to manage this additional parameter (It would also move `DisableUpgradePredicatesAnnotation` closer to other annotations):

The idea would be to have something along those lines in `elasticsearch_types.go`:
```golang
func (es Elasticsearch) DisabledPredicates() set.StringSet {
	if es.Annotations == nil {
		return set.StringSet{}
	}
	disabledPredicates, exists := es.Annotations[DisableUpgradePredicatesAnnotation]
	if !exists {
		return set.StringSet{}
	}
	return set.Make(strings.Split(disabledPredicates, ",")...)
}
```

_(We could also create a "utility function" shared with `SuspendedPodNames` as I think we almost want the same result: a string set from an annotation)_

Also maybe add a relevant unit test with some "edge cases" in `elasticsearch_types_test.go`:

```golang
func TestElasticsearch_DisabledPredicates(t *testing.T) {
	tests := []struct {
		name string
		es   Elasticsearch
		want set.StringSet
	}{
		{
			name: "no annotations",
			es: Elasticsearch{ObjectMeta: metav1.ObjectMeta{
				Annotations: nil,
			}},
			want: set.StringSet{},
		},
		{
			name: "no annotation",
			es: Elasticsearch{ObjectMeta: metav1.ObjectMeta{
				Annotations: map[string]string{},
			}},
			want: set.StringSet{},
		},
		{
			name: "1 disabled predicate",
			es: Elasticsearch{ObjectMeta: metav1.ObjectMeta{
				Annotations: map[string]string{
					DisableUpgradePredicatesAnnotation: "foo",
				},
			}},
			want: set.Make("foo"),
		},
		{
			name: "2 disabled predicates",
			es: Elasticsearch{ObjectMeta: metav1.ObjectMeta{
				Annotations: map[string]string{
					DisableUpgradePredicatesAnnotation: "foo,bar",
				},
			}},
			want: set.Make("foo", "bar"),
		},
	}
	for _, tt := range tests {
			if got := tt.es.DisabledPredicates(); !reflect.DeepEqual(got, tt.want) {
				t.Errorf("Elasticsearch.DisabledPredicates() = %v, want %v", got, tt.want)
			}
		})
	}
}
```

The logic in `runPredicates` would be:

```golang
func runPredicates(
	ctx PredicateContext,
	candidate corev1.Pod,
	deletedPods []corev1.Pod,
	annotations map[string]string,
	maxUnavailableReached bool,
) (*failedPredicate, error) {

	disabledPredicates := ctx.es.DisabledPredicates()
	for _, predicate := range predicates {
		canDelete, err := predicate.fn(ctx, candidate, deletedPods, maxUnavailableReached)
		if err != nil {
			return nil, err
		}
		if !canDelete {
			// if this specific predicate name is disabled by the disable predicate annotation
			// "eck.k8s.elastic.co/disable-upgrade-predicates", then ignore this predicate,
			// and continue processing the remaining predicates.
			if disabledPredicates.Has(predicate.name) || disabledPredicates.Has("*") {
				log.Info("Warning: disabling upgrade predicate because of annotation", "predicate", predicate.name, "namespace", ctx.es.Namespace, "es_name", ctx.es.Name)
				continue
			}
...
```

docs/orchestrating-elastic-stack-applications/elasticsearch/orchestration.asciidoc
```diff
@@ -173,3 +173,68 @@ Specifically, if the Elasticsearch index settings demand a higher number of shar
...
 ** Use link:https://www.elastic.co/guide/en/elasticsearch/reference/current/index-modules.html#dynamic-index-settings[`auto_expand_replicas`] to automatically adjust the replicas to the number of data nodes in the cluster.
+
+[id="{p}-advanced-upgrade-control"]
+== Advanced control during rolling upgrades
+
+The rules (otherwise known as predicates) that the ECK operator follows during an Elasticsearch upgrade can be selectively disabled for certain scenarios where the ECK operator will not proceed with an Elasticsearch cluster upgrade because it deems it to be "unsafe".
+
+WARNING: Selectively disabling the predicates below are extremely risky, and carry a high chance of either data loss, or causing a cluster to become completely unavailable.  Use only if you are absolutely positive you are not causing permanent damage to an Elasticsearch cluster.
+
+* The following named predicates control the upgrade process
```
barkbay, line 184: Maybe also mention that all the predicates can be disabled using `"*"`

barkbay (commented):
> A comma separated list should suffice here for > 1 predicate. I'll get this change into this PR.

👍 I think it would be great to be able to disable more than one replica.

docs/orchestrating-elastic-stack-applications/elasticsearch/orchestration.asciidoc
```diff
@@ -173,3 +173,68 @@ Specifically, if the Elasticsearch index settings demand a higher number of shar
...
+** require_started_replica
++
+If a cluster is yellow, allow deleting a node, but only if they do not contain the only replica of a shard since it would make the cluster go red.
+** one_master_at_a_time
++
+Only allow a single master to be upgraded at a time.
+** do_not_delete_last_master_if_all_master_ineligible_nodes_are_not_upgraded
++
+Force an upgrade of all the master-ineligible nodes before upgrading the last master-eligible node.
+** do_not_delete_pods_with_same_shards
```
thbkrkr, line 207: As the predicate names become public with this feature, I wonder if we should write a little check to run in CI to verify that the names in the doc are in sync with the names in the code.

Something like this:
```bash
code=pkg/controller/elasticsearch/driver/upgrade_predicates.go \
doc=docs/orchestrating-elastic-stack-applications/elasticsearch/orchestration.asciidoc \
predicate_pattern='[a-z]*_[A-Za-z_]*'
diff \
	<(grep "name:" "$code" | grep -o "$predicate_pattern" ) \
	<(grep '\*\* [a-z]' "$doc" | grep -o "$predicate_pattern" )
```
If we add a comment in the two files to mark the predicate name (`// predicate`), this could make the parsing more sure. 
```
diff \
	<(grep '// predicate' $code | grep -o "$predicate_pattern" ) \
	<(grep '// predicate' -A1 $doc | grep -o "$predicate_pattern" )
```

docs/orchestrating-elastic-stack-applications/elasticsearch/orchestration.asciidoc
naemono: good point.

pkg/controller/elasticsearch/driver/fixtures.go
naemono: this makes perfect sense, done.

docs/orchestrating-elastic-stack-applications/elasticsearch/orchestration.asciidoc
naemono: This is a good thought. I've added this CI check, missing the marks.  Let's see if that's sufficient.  I'm happy to also add the marks if we feel that's necessary.

docs/orchestrating-elastic-stack-applications/elasticsearch/orchestration.asciidoc
```diff
@@ -173,3 +173,69 @@ Specifically, if the Elasticsearch index settings demand a higher number of shar
...
+metadata:
+  name: testing
+  annotations:
+    eck.k8s.elastic.co/disable-upgrade-predicates: "only_restart_healthy_node_if_green_or_yellow"
+    # Also note that eck.k8s.elastic.co/disable-upgrade-predicates: "*" would work as well, but is much less selective.
+spec:
+  version: 7.15.3 # previously set to 7.15.2, for example
+```
+
+The ECK operator would allow this upgrade to proceed, even though the cluster was in a "red" state during this upgrade process.
```
barkbay, line 241: nit: maybe add another warning: "Do not forget to remove the annotation once not needed" , and a one-liner to remove it: `kublectl annotate elasticsearch.elasticsearch.k8s.elastic.co/elasticsearch-sample eck.k8s.elastic.co/disable-upgrade-predicate-`

pkg/apis/elasticsearch/v1/elasticsearch_types.go
```diff
@@ -634,3 +638,24 @@ func (es *Elasticsearch) MonitoringAssociation(ref commonv1.ObjectSelector) comm
...
+	//
+	//   To disable "if_yellow_only_restart_upgrading_nodes_with_unassigned_replicas" predicate
+	//
+	//   metadata:
+	//     annotations:
+	//       eck.k8s.elastic.co/disable-upgrade-predicates="if_yellow_only_restart_upgrading_nodes_with_unassigned_replicas"
+	DisableUpgradePredicatesAnnotation = "eck.k8s.elastic.co/disable-upgrade-predicates"
+)
+
+// DisabledPredicates will return the set of predicates that are currently disabled by the
```
barkbay, line 657:
```suggestion
// DisabledPredicates returns the set of predicates that are currently disabled by the
```

test/e2e/test/version.go
```diff
@@ -16,7 +16,7 @@ const (
 	// LatestReleasedVersion6x is the latest released version for 6.x
 	LatestReleasedVersion6x = "6.8.20"
 	// LatestReleasedVersion7x is the latest released version for 7.x
-	LatestReleasedVersion7x = "7.16.2"
+	LatestReleasedVersion7x = "7.16.3"
```
barkbay, line 19: nit: I would keep version upgrades in dedicated PRs, they might have some side effects, it makes e2e test failures easier to track.

test/e2e/es/disable_predicates_test.go
```diff
@@ -0,0 +1,91 @@
...
+func TestRedClusterCanBeUpgradedByDisablingPredicate(t *testing.T) {
+	srcVersion := test.Ctx().ElasticStackVersion
+	dstVersion := test.LatestReleasedVersion7x
+
+	test.SkipInvalidUpgrade(t, srcVersion, dstVersion)
+
+	k := test.NewK8sClientOrFatal()
+	initial := elasticsearch.NewBuilder("test-v-up-with-red-cluster-7x").
+		WithESMasterDataNodes(1, elasticsearch.DefaultResources).
+		WithAnnotation(driver.DisableUpgradePredicatesAnnotation, "only_restart_healthy_node_if_green_or_yellow")
```
barkbay, line 34:
```suggestion
		WithAnnotation(esv1.DisableUpgradePredicatesAnnotation, "only_restart_healthy_node_if_green_or_yellow")
```

test/e2e/es/disable_predicates_test.go
```diff
@@ -0,0 +1,91 @@
...
+
+	esv1 "github.com/elastic/cloud-on-k8s/pkg/apis/elasticsearch/v1"
+	"github.com/elastic/cloud-on-k8s/pkg/controller/elasticsearch/client"
+	"github.com/elastic/cloud-on-k8s/pkg/controller/elasticsearch/driver"
+	"github.com/elastic/cloud-on-k8s/test/e2e/test"
+	"github.com/elastic/cloud-on-k8s/test/e2e/test/elasticsearch"
+)
+
+func TestRedClusterCanBeUpgradedByDisablingPredicate(t *testing.T) {
+	srcVersion := test.Ctx().ElasticStackVersion
```
barkbay, line 26: nit: instead of relying on a version upgrade, which might bring its own issues, I would use an annotation in the `podTemplate` to trigger an upgrade:

```golang
	podTemplate1 := elasticsearch.ESPodTemplate(elasticsearch.DefaultResources)
	podTemplate1.Annotations = map[string]string{"foo": "bar"}

	k := test.NewK8sClientOrFatal()
	initial := elasticsearch.NewBuilder("test-v-up-with-red-cluster").
		WithNodeSet(esv1.NodeSet{
			Name:        "default",
			Count:       1,
			PodTemplate: podTemplate1,
		}).
		WithAnnotation(esv1.DisableUpgradePredicatesAnnotation, "only_restart_healthy_node_if_green_or_yellow")

	podTemplate2 := elasticsearch.ESPodTemplate(elasticsearch.DefaultResources)
	podTemplate2.Annotations = map[string]string{"foo": "bar2"}
	mutated := initial.WithNoESTopology().
		WithNodeSet(esv1.NodeSet{
			Name:        "default",
			Count:       1,
			PodTemplate: podTemplate2,
		}).
		WithAnnotation(esv1.DisableUpgradePredicatesAnnotation, "only_restart_healthy_node_if_green_or_yellow")
```

barkbay (approved):
I left a few comments but LGTM.
Since we have some documentation here I would maybe request a review from @alaudazzi

test/e2e/test/version.go
naemono: Completely understand.  thanks

test/e2e/es/disable_predicates_test.go
naemono: Thanks.  this is a simpler path for sure.

naemono:
will merge after review by @alaudazzi, as suggested.

docs/orchestrating-elastic-stack-applications/elasticsearch/orchestration.asciidoc
```diff
@@ -173,3 +173,69 @@ Specifically, if the Elasticsearch index settings demand a higher number of shar
...
+If `maxUnavailable` is reached, only allow unhealthy Pods to be deleted.
+** skip_already_terminating_pods
++
+Do not attempt to restart pods that are already in the process of being terminated.
+** only_restart_healthy_node_if_green_or_yellow
++
+Only restart healthy Elasticsearch nodes if the health of the cluster is either green or yellow, never red.
+** if_yellow_only_restart_upgrading_nodes_with_unassigned_replicas
++
+During a rolling upgrade, primary shards assigned to a node running a new version cannot have their replicas assigned to a node with the old version. Therefore we must allow some Pods to be restarted even if cluster health is Yellow so the replicas can be assigned.
```
alaudazzi, line 197:
```suggestion
During a rolling upgrade, primary shards assigned to a node running a new version cannot have their replicas assigned to a node with the old version. Therefore we must allow some Pods to be restarted even if cluster health is yellow so the replicas can be assigned.
```

docs/orchestrating-elastic-stack-applications/elasticsearch/orchestration.asciidoc
```diff
@@ -173,3 +173,69 @@ Specifically, if the Elasticsearch index settings demand a higher number of shar
...
+If a cluster is yellow, allow deleting a node, but only if they do not contain the only replica of a shard since it would make the cluster go red.
+** one_master_at_a_time
++
+Only allow a single master to be upgraded at a time.
+** do_not_delete_last_master_if_all_master_ineligible_nodes_are_not_upgraded
++
+Force an upgrade of all the master-ineligible nodes before upgrading the last master-eligible node.
+** do_not_delete_pods_with_same_shards
++
+Do not allow 2 pods to be deleted at the same time, that contain the same shard.
```
alaudazzi, line 209:
```suggestion
Do not allow two pods containing the same shard to be deleted at the same time.
```

docs/orchestrating-elastic-stack-applications/elasticsearch/orchestration.asciidoc
```diff
@@ -173,3 +173,69 @@ Specifically, if the Elasticsearch index settings demand a higher number of shar
...
+
+Any of these predicates can be disabled by adding an annotation with the key of `eck.k8s.elastic.co/disable-upgrade-predicates` to the Elasticsearch metadata, specifically naming the predicate that is needing to be disabled.  Also, all predicates can be disabled by replacing the name of any predicatae with "*".
+
+* Example use case
+
+Assume a given Elasticsearch cluster is a "red" state because of ab un-allocatable shard setting that was applied to the cluster
```
alaudazzi, line 215:
```suggestion
Assume a given Elasticsearch cluster is a "red" state because of an un-allocatable shard setting that was applied to the cluster:
```

alaudazzi (approved):
Left a few suggestions in the Asciidoc file, otherwise LGTM.

PR #5303, opened 2022-01-28
Use Kind instead of GKE to run E2E test on PR job

thbkrkr:
Change the PR job to use a Kind cluster instead of a GKE cluster to save a few minutes and dollars.
I don't know how much we'll save, but I think it's worth a try.

### DIFF
--- a/hack/deployer/config/plans.yml
+++ b/hack/deployer/config/plans.yml
@@ -157,7 +157,7 @@ plans:
   kubernetesVersion: 1.21.1
   kind:
     nodeCount: 3
-    nodeImage: kindest/node:v1.21.1
+    nodeImage: kindest/node:v1.21.1@sha256:69860bda5563ac81e3c0057d654b5253219618a22ec3a346306239bba8cfa1a6
     ipFamily: ipv4
 - id: kind-ci
   operation: create


thbkrkr:
Creating the Kind cluster took 1min21s which is faster than creating a GKE cluster. :partying_face:

I also forgot that when we added  specific `eck` workers to run Kind in addition to a bigger disk (120GB) we chose a more powerful config (n1-standard-16 16 vCPUs 60 GiB) than that of basic `linux` workers (n1-standard-4 5 VCPUs 15 GiB). Consequence: all stages went faster! Total duration 26m 27s. :partying_face:

thbkrkr:
Let's trigger another run to see how it works today. Jenkins test this please.

thbkrkr:
Jenkins test this please.

thbkrkr:
Jenkins test this please.

.ci/pipelines/pr-gke.Jenkinsfile
```diff
@@ -71,7 +71,7 @@ pipeline {
                         }
                     }
                     steps {
-                        sh '.ci/setenvconfig pr'
+                        sh '.ci/setenvconfig pr kindest/node:v1.22.0@sha256:b8bda84bb3a190e6e028b1760d277454a72267a5454b57db34437c34a588d047 0.11.1 ipv4'
```
pebrc, line 74: Should we just use/update the defaults in the `kind-ci` item in  `hack/deployer/config/plans.yaml` instead of passing the arguments here?

pebrc (approved):
This is a great improvement! I see 24 minute runs instead of ~ 43 minutes

.ci/pipelines/pr-gke.Jenkinsfile
thbkrkr: Good idea.

PR #5304, opened 2022-01-28
Do not push non-release builds to dockerhub

naemono:


naemono:
Hold up on review of this.  Let me do some further verification.

Makefile
```diff
@@ -362,7 +362,23 @@ switch-tanzu:
 #################################
 ##  --    Docker images    --  ##
 #################################
+# This regex will only match the below, without a trailing build/release candidate
+# 1. an optional preceding 'v'
+# 2. a standard semver '1.9.0'
+tag_regex_no_build_candidates := ^v{0,1}([0-9]{1,}\.){2}[0-9]{1,}$$
+
 docker-multiarch-build: go-generate generate-config-file 
+ifeq ($(shell echo ${IMG_VERSION} | egrep "${tag_regex_no_build_candidates}"),)
```
thbkrkr, line 371: Nice regexp :) I think we could otherwise rely on the environment variable [`SNAPSHOT` or `TAG_NAME`](https://github.com/elastic/cloud-on-k8s/blob/2efb002c1dde3ba63d6f039dc983b3dc8d24b32a/.ci/setenvconfig#L473-L510).

```
ifneq ($(SNAPSHOT),true) # check for `!true` so that if never it is not set, it's nightly
 // nightly
or
ifeq ($(strip $(TAG_NAME)),)
  // nightly
```

naemono: Well @thbkrkr that shows my familiarity with the build process.  That's a lot cleaner.  Updated.  Thanks for the assist.

Makefile
```diff
@@ -362,7 +362,19 @@ switch-tanzu:
 #################################
 ##  --    Docker images    --  ##
 #################################
+
 docker-multiarch-build: go-generate generate-config-file 
+ifeq ($(SNAPSHOT),true)
```
thbkrkr, line 367: One more little thing, I think it's safer to have the `if` applied to the release build and the `else` to the nightly build, so that if `SNAPSHOT=yolo` or SNAPSHOT is unexpectedly not set, it's considered as a nightly build. So I would invert the logic.

naemono: Updated.  Let me know if this isn't the inverted logic you had in mind.  I think it's what you're referring to.

thbkrkr (approved):
LGTM :+1:

pebrc (approved):

PR #5310, opened 2022-01-31
Fix Stack Monitoring with custom certificate without CA

### LINKED ISSUE #5309: Stack monitoring doesn't work with a resource monitored with a custom certificate without a CA
Stack Monitoring incorrectly assumes that if the monitored Elastic resource has TLS enabled, there is a CA to configure in the input section of the Metricbeat config. This overlloks that you can have a certificate issued by a well-known CA, so you don't always provide a CA when TLS is enabled. This makes Stack Monitoring impossible to use with a custom certificate without a CA.

Symptom: Metricbeat is in a boot loop due to an invalid configuration.
```
2022-01-31T23:09:16.106Z ERROR [kibana.stats] stats/stats.go:79 1 error: open 
/mnt/elastic-internal/kb-monitoring/<es-namespace>/<es-name>/certs/ca.crt: 
no such file or directory reading <nil> accessing config
```

thbkrkr:
Resolves #5309.

### DIFF
--- a/pkg/controller/common/certificates/ca.go
+++ b/pkg/controller/common/certificates/ca.go
@@ -135,7 +135,7 @@ func (c *CA) CreateCertificate(
 // namespace and name.
 func HasPublicCA(client k8s.Client, namer name.Namer, namespace string, name string) (bool, error) {
 	certsNsn := types.NamespacedName{
-		Name: PublicCertsSecretName(namer, name),
+		Name:      PublicCertsSecretName(namer, name),
 		Namespace: namespace,
 	}
 	var certsSecret corev1.Secret
--- a/pkg/controller/common/certificates/ca.go
+++ b/pkg/controller/common/certificates/ca.go
@@ -5,6 +5,7 @@
 package certificates
 
 import (
+	"context"
 	"crypto"
 	cryptorand "crypto/rand"
 	"crypto/rsa"
@@ -14,6 +15,11 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+
+	"github.com/elastic/cloud-on-k8s/pkg/controller/common/name"
+	"github.com/elastic/cloud-on-k8s/pkg/utils/k8s"
 )
 
 var (
@@ -123,4 +129,19 @@ func (c *CA) CreateCertificate(
 	)
 
 	return certData, err
+}
+
+// HasPublicCA returns true if an Elastic resource has a CA (ca.crt) in its public HTTP certs secret given its namer,
+// namespace and name.
+func HasPublicCA(client k8s.Client, namer name.Namer, namespace string, name string) (bool, error) {
+	certsNsn := types.NamespacedName{
+		Name: PublicCertsSecretName(namer, name),
+		Namespace: namespace,
+	}
+	var certsSecret corev1.Secret
+	if err := client.Get(context.Background(), certsNsn, &certsSecret); err != nil {
+		return false, err
+	}
+	_, ok := certsSecret.Data[CAFileName]
+	return ok, nil
 }
--- a/pkg/controller/common/certificates/ca.go
+++ b/pkg/controller/common/certificates/ca.go
@@ -131,9 +131,9 @@ func (c *CA) CreateCertificate(
 	return certData, err
 }
 
-// HasPublicCA returns true if an Elastic resource has a CA (ca.crt) in its public HTTP certs secret given its namer,
+// PublicCertsHasCACert returns true if an Elastic resource has a CA (ca.crt) in its public HTTP certs secret given its namer,
 // namespace and name.
-func HasPublicCA(client k8s.Client, namer name.Namer, namespace string, name string) (bool, error) {
+func PublicCertsHasCACert(client k8s.Client, namer name.Namer, namespace string, name string) (bool, error) {
 	certsNsn := types.NamespacedName{
 		Name:      PublicCertsSecretName(namer, name),
 		Namespace: namespace,
--- a/pkg/controller/common/certificates/ca_test.go
+++ b/pkg/controller/common/certificates/ca_test.go
@@ -128,7 +128,7 @@ func Test_PublicCertsHasCACert(t *testing.T) {
 			want: false,
 		},
 		{
-			name: "Error if secret",
+			name: "Error if no certs secret",
 			wantErr: true,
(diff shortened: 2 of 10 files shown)


pkg/controller/common/certificates/ca_test.go
```diff
@@ -85,3 +90,36 @@ func TestNewSelfSignedCA(t *testing.T) {
...
+			CertFileName: []byte("..."),
+		},
+	}
+	hasCA, err = HasPublicCA(k8s.NewFakeClient(&certsWithoutCA), esv1.ESNamer, "ns", "c2")
+	assert.NoError(t, err)
+	assert.False(t, hasCA)
+
+	// no certs secret
+	_, err = HasPublicCA(k8s.NewFakeClient(), esv1.ESNamer, "ns", "c3")
+	assert.Error(t, err)
```
pebrc, line 124: Should we stick with the table tests for consistency?

pkg/controller/common/stackmon/config.go
```diff
@@ -159,8 +159,8 @@ type inputConfigData struct {
 	URL      string
 	Username string
 	Password string
-	IsSSL    bool
-	SSLPath  string
+	IsCA     bool
```
pebrc, line 162: I know `Is*` is the go convention but would `HasCA` not be better fitting here?

pkg/controller/kibana/stackmon/metricbeat.tpl.yml
```diff
@@ -9,8 +9,8 @@ metricbeat.modules:
     hosts: ["{{ .URL }}"]
     username: {{ .Username }}
     password: {{ .Password }}
-    {{- if .IsSSL }}
-    ssl.certificate_authorities: ["{{ .SSLPath }}"]
+    {{- if .IsCA }}
+    ssl.certificate_authorities: ["{{ .CAPath }}"]
     ssl.verification_mode: "{{ .SSLMode }}"
```
pebrc, line 14: `verification_mode` is now predicated on the presence of a CA certificate. Will this not be a problem if users are using their custom certificates without a CA and those certificates are not issued for localhost?

pkg/controller/common/certificates/ca.go
```diff
@@ -124,3 +130,18 @@ func (c *CA) CreateCertificate(
 
 	return certData, err
 }
+
+// HasPublicCA returns true if an Elastic resource has a CA (ca.crt) in its public HTTP certs secret given its namer,
+// namespace and name.
+func HasPublicCA(client k8s.Client, namer name.Namer, namespace string, name string) (bool, error) {
```
barkbay, line 136: nit, naming: `ca.crt` can be managed in the context of a private PKI, without being a public CA

```suggestion
// PublicCertsHasCACert returns true if an Elastic resource has a Certificate Authority certificate (ca.crt) in its public HTTP certs secret given its namer,
// namespace and name.
func PublicCertsHasCACert(client k8s.Client, namer name.Namer, namespace string, name string) (bool, error) {
```

barkbay (commented):
I did a quick test with Elasticsearch and Kibana pub. certs signed by a well known CA. Metricbeat is failing with the following error:

```
2022-02-01T08:26:57.702Z	ERROR	module/wrapper.go:259	Error fetching data for metricset elasticsearch.index: error determining if connected Elasticsearch node is master: error making http request: Get "https://localhost:9200/_nodes/_local/nodes": x509: certificate is valid for my-public-hostname.com, not localhost
```

I think it is related to Peter's comment [here](https://github.com/elastic/cloud-on-k8s/pull/5310#discussion_r796321746).

pkg/controller/common/certificates/ca_test.go
thbkrkr: Yes.

pkg/controller/common/stackmon/config.go
thbkrkr: ++, I was tempted but I didn't dare.

pkg/controller/kibana/stackmon/metricbeat.tpl.yml
thbkrkr: It is set to `certificate` which IIUC should verify that the provided certificate is signed by a trusted authority (CA), but does not perform any hostname verification ([doc](https://www.elastic.co/guide/en/beats/metricbeat/current/configuration-ssl.html#ssl-common-config)). But you are right, this is a problem as reported by Michael [here](https://github.com/elastic/cloud-on-k8s/pull/5310#pullrequestreview-868801747). I'm a bit confused.

thbkrkr: Yes, you are right, it also needs to be set with a custom cert without a CA.

thbkrkr: I opted to keep `IsSSL`, add `ssl.enabled: {{ .IsSSL }}` and always set `ssl.verification_mode: certificate`.

thbkrkr:
Thanks for the feedback. All have been addressed.

pkg/controller/elasticsearch/stackmon/metricbeat.tpl.yml
```diff
@@ -17,9 +17,10 @@ metricbeat.modules:
     hosts: ["{{ .URL }}"]
     username: {{ .Username }}
     password: {{ .Password }}
-    {{- if .IsSSL }}
-    ssl.certificate_authorities: ["{{ .SSLPath }}"]
+    ssl.enabled: {{ .IsSSL }}
     ssl.verification_mode: "{{ .SSLMode }}"
```
pebrc, line 21: Nit: I wonder if we should just have the verification mode in the template directly given that the value is effectively constant.

pebrc (approved):
Looks good to me now. I did a test with a Let's encrypt certificate and Elasticsearch and one with the self-signed certs ECK uses by default.

pkg/controller/elasticsearch/stackmon/metricbeat.tpl.yml
thbkrkr: Yeah, that crossed my mind too.

thbkrkr:
I found a little bug. When disabling TLS, we generate certs for ES but not for KB. This means we shouldn't get the `kb-http-public-certs` secret to resolve `hasCA` for the Kibana Metricbeat config without first checking if TLS is enabled.

pebrc (approved):
> I found a little bug

Good catch. I did not test with Kibana initially, but it looks good now.

PR #5308, opened 2022-01-31
Fix incorrect modelled node shutdown API in Elasticsearch client

pebrc:
Fixes #5307 

Excluded from release notes because node shutdown functionality has not yet been shipped in any release.

PR #5311, opened 2022-02-01
[2.0] Fix incorrectly modelled node shutdown API in Elasticsearch client (#5308)

pebrc:
 - Fix incorrectly modelled node shutdown API in Elasticsearch client (#5308)


barkbay (approved):
LGTM

PR #5314, opened 2022-02-01
Fix association with custom cert without CA

thbkrkr:
Follow-up of #5277 to adress the case of a custom cert without CA that I forgot. This commit takes into account `CACertProvided` in `CAIsConfigured()` and not longer just the presence of `CASecretName`. When using a custom certificate, the `CASecretName` is defined but `CACertProvided` is `false`. I did not rechallenge this logic.

https://github.com/elastic/cloud-on-k8s/blob/721cccfd905f43c345c7957fbfe173ea3ebb6bbf/pkg/controller/association/ca.go#L64-L65


thbkrkr:
I missed the initial intention. We mount the CA secret using a volume in the pods when the secret exists (works when we mount the secret as a directory without directly targetting the CA file) but we don't configure the CA in the configurations when the CA is not provided in the secret. This means we can keep the calls to `CAIsConfigured` when it's about configuring the pod without referencing `ca.crt`.

pkg/controller/common/association/association.go
```diff
@@ -58,7 +58,7 @@ func writeAuthSecretToConfigHash(client k8s.Client, assoc commonv1.Association,
 
 func writeCASecretToConfigHash(client k8s.Client, assoc commonv1.Association, configHash hash.Hash) error {
 	assocConf := assoc.AssociationConf()
-	if !assocConf.CAIsConfigured() {
+	if !assocConf.GetCACertProvided() {
```
pebrc, line 61: I am still confused by this. If all we care about is the CA certificate why do we need these two different methods? Why do we copy the public certificates at all into the associations namespace when they do not contain a CA? What I mean is could we not simplify this further?

pkg/controller/agent/pod.go
```diff
@@ -488,19 +488,23 @@ func getFleetSetupFleetEnvVars(agent agentv1alpha1.Agent, client k8s.Client) (ma
...
 		assoc, err := association.SingleAssociationOfType(agent.GetAssociations(), commonv1.FleetServerAssociationType)
 		if err != nil {
 			return nil, err
 		}
+		if assoc == nil {
+			return fleetCfg, nil
+		}
 
-		if assoc != nil {
+		fleetCfg[FleetURL] = assoc.AssociationConf().GetURL()
```
pebrc, line 502: I realise that part of what I have asked for in #5240 is implemented here!

pebrc (approved):
I am OK with merging this given we follow up in #5379

PR #5327, opened 2022-02-03
Do not attempt rolling upgrades for non-HA clusters

### LINKED ISSUE #5321: Rolling upgrades between major versions for 2-node Elasticsearch cluster impossible
Rolling upgrades between major versions for a 2-node Elasticsearch cluster can be impossible. If the first node that is upgraded is the master node, then the rolling upgrade gets stuck because we never reach a state where we have two nodes in the cluster. The cluster health stays yellow. The second node can't join the cluster formed by the first upgraded node.

Log of the second node log:
```
{
  "type": "server",
  "timestamp": "2022-02-03T09:29:04,631Z",
  "level": "WARN",
  "component": "o.e.c.c.JoinHelper",
  "cluster.name": "test-version-up-2-to-8x-wddt",
  "node.name": "test-version-up-2-to-8x-wddt-es-masterdata-0",
	
  "message": "last failed join attempt was 7ms ago, failed to join {test-version-up-2-to-8x-wddt-es-masterdata-1}{Dm1aKtG4QtqAWdG1qYhxrg}{66D72IvmSR2LTi77eKbd_g}{10.42.176.208}{10.42.176.208:9300}{cdfhilmrstw}{k8s_node_name=gke-thbkrkr-dev-cluster-default-pool-9292dcc0-3h08, ml.machine_memory=2147483648, xpack.installed=true, ml.max_jvm_size=1073741824} with JoinRequest{sourceNode={test-version-up-2-to-8x-wddt-es-masterdata-0}{A50qiG3DRo-dL0toCbiQxA}{_tnSXZa1QOGe97A63qoKDw}{10.42.177.120}{10.42.177.120:9300}{cdfhilmrstw}{k8s_node_name=gke-thbkrkr-dev-cluster-default-pool-a617668f-p82k, ml.machine_memory=2147483648, xpack.installed=true, transform.node=true, ml.max_open_jobs=512, ml.max_jvm_size=1073741824}, minimumTerm=2, optionalJoin=Optional[Join{term=2, lastAcceptedTerm=1, lastAcceptedVersion=65, sourceNode={test-version-up-2-to-8x-wddt-es-masterdata-0}{A50qiG3DRo-dL0toCbiQxA}{_tnSXZa1QOGe97A63qoKDw}{10.42.177.120}{10.42.177.120:9300}{cdfhilmrstw}{k8s_node_name=gke-thbkrkr-dev-cluster-default-pool-a617668f-p82k, ml.machine_memory=2147483648, xpack.installed=true, transform.node=true, ml.max_open_jobs=512, ml.max_jvm_size=1073741824}, targetNode={test-version-up-2-to-8x-wddt-es-masterdata-1}{Dm1aKtG4QtqAWdG1qYhxrg}{66D72IvmSR2LTi77eKbd_g}{10.42.176.208}{10.42.176.208:9300}{cdfhilmrstw}{k8s_node_name=gke-thbkrkr-dev-cluster-default-pool-9292dcc0-3h08, ml.machine_memory=2147483648, xpack.installed=true, ml.max_jvm_size=1073741824}}]}",

  "cluster.uuid": "En4wbZQ-Ru-J0IXK-Ysl0g",
  "node.id": "A50qiG3DRo-dL0toCbiQxA" , 
  "stacktrace": ["org.elasticsearch.transport.RemoteTransportException: [test-version-up-2-to-8x-wddt-es-masterdata-1][10.42.176.208:9300][internal:cluster/coordination/join]",
    "Caused by: java.lang.IllegalStateException:

      node version [7.17.0] may not join a cluster comprising only nodes of version [8.0.0] or greater",

    "at org.elasticsearch.cluster.coordination.JoinTaskExecutor.ensureVersionBarrier(JoinTaskExecutor.java:325) ~[elasticsearch-7.17.0-SNAPSHOT.jar:7.17.0-SNAPSHOT]",
    "at org.elasticsearch.cluster.coordination.Coordinator.validateJoinRequest(Coordinator.java:585) ~[elasticsearch-7.17.0-SNAPSHOT.jar:7.17.0-SNAPSHOT]",
    "at org.elasticsearch.cluster.coordination.Coordinator.lambda$handleJoinRequest$9(Coordinator.java:556) ~[elasticsearch-7.17.0-SNAPSHOT.jar:7.17.0-SNAPSHOT]",
    "at org.elasticsearch.action.ActionListener$1.onResponse(ActionListener.java:136) ~[elasticsearch-7.17.0-SNAPSHOT.jar:7.17.0-SNAPSHOT]",
...
```

<details>
<summary>
YAML manifest to reproduce:
</summary>

```yaml
kind: Elasticsearch
metadata:
  name: test-version-up-2-to-8x
  namespace: e2e-mercury
  version: 7.17.0-SNAPSHOT
  #version: 8.0.0-SNAPSHOT
  nodeSets:
  - config:
      logger.org.elasticsearch.cluster.service.MasterService: trace
      node.store.allow_mmap: false
    count: 2
    name: masterdata
      spec:
        containers:
        - name: elasticsearch
          resources:
            limits:
              memory: 2Gi
        securityContext:
          fsGroup: 12345
          runAsNonRoot: true
          runAsUser: 12345
    volumeClaimTemplates:
    - metadata:
        name: elasticsearch-data
      spec:
        accessModes:
        - ReadWriteOnce
        resources:
          requests:
            storage: 1Gi
```
</details>

ECK should do a full cluster restart on clusters with 1 or 2 voting nodes/master role nodes.

pebrc:
Fixes #5321 

This tries to handle the non-HA cluster upgrade problem by doing a full immediate restart of the whole cluster instead. 

I restricted this to major version upgrades only because I was not able to reproduce the problem on minor upgrades.


pkg/controller/elasticsearch/driver/upgrade.go
```diff
@@ -75,21 +76,38 @@ func (d *defaultDriver) handleRollingUpgrades(
...
 	numberOfPods := len(currentPods)
+
+	expectedMasters := expectedResources.MasterNodesNames()
+
+	isMajorVersionUpgrade, err := isMajorVersionUpgrade(d.ES)
+	if err != nil {
+		return results.WithError(err)
+	}
+	var deletedPods []corev1.Pod
+	shouldDoFullRestartUpgrade := isNonHACluster(currentPods, expectedMasters) && isMajorVersionUpgrade
```
barkbay, line 87: cosmetic nit: `shouldDoFullRestartUpgrade` is not immediately used:

```golang
	expectedMasters := expectedResources.MasterNodesNames()

	// Maybe upgrade some of the nodes.
	upgrade := newUpgrade(
		ctx,
		d,
		statefulSets,
		expectedResources,
		esClient,
		esState,
		nodeShutdown,
		expectedMasters,
		actualMasters,
		podsToUpgrade,
		healthyPods,
		numberOfPods,
	)

	isMajorVersionUpgrade, err := isMajorVersionUpgrade(d.ES)
	if err != nil {
		return results.WithError(err)
	}

	var deletedPods []corev1.Pod
	shouldDoFullRestartUpgrade := isNonHACluster(currentPods, expectedMasters) && isMajorVersionUpgrade
	if shouldDoFullRestartUpgrade {
		// unconditional full cluster upgrade
		deletedPods, err = run(upgrade.DeleteAll)
	} else {
		// regular rolling upgrade
		deletedPods, err = run(upgrade.Delete)
	}
```

pkg/controller/elasticsearch/driver/upgrade_pods_deletion.go
```diff
@@ -78,7 +78,35 @@ func (ctx *rollingUpgradeCtx) Delete() ([]corev1.Pod, error) {
...
+	if err := ctx.prepareClusterForNodeRestart(ctx.podsToUpgrade); err != nil {
+		return nil, err
+	}
+	// TODO: If master is changed into a data node (or the opposite) it must be excluded or we should update m_m_n
+	var deletedPods []corev1.Pod //nolint:prealloc
+	for _, podToDelete := range ctx.podsToUpgrade {
+		if err := ctx.handleMasterScaleChange(podToDelete); err != nil {
+			return deletedPods, err
+		}
+		if readyToDelete, err := ctx.readyToDelete(podToDelete); err != nil || !readyToDelete {
```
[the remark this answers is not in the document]
pebrc: That's a good point. I am going to look into that but I think you are right.

pkg/controller/elasticsearch/driver/upgrade.go
```diff
@@ -176,6 +195,29 @@ func (ctx rollingUpgradeCtx) run() ([]corev1.Pod, error) {
...
+	actualMasters := label.FilterMasterNodePods(actualPods)
+	return len(actualMasters) <= 2
+}
+
+func isMajorVersionUpgrade(es esv1.Elasticsearch) (bool, error) {
+	specVersion, err := version.Parse(es.Spec.Version)
+	if err != nil {
+		return false, err
+	}
+	statusVersion, err := version.Parse(es.Status.Version)
```
barkbay, line 214: This creates a dependency on the resource's `Status`, I'm wondering if we should consider the situation where `es.Status.Version` is an empty string. My feeling is that `es.Status.Version` is managed on a "[best-effort](https://github.com/elastic/cloud-on-k8s/blob/96006dec0977d9fdb84b8be2a41bd9c4c03d2670/pkg/controller/elasticsearch/reconcile/state.go#L103)" basis, and the upgrade process should not be prevented because the status is not fully up-to-date yet. If we cannot parse `es.Status.Version` maybe we should err on the safe side and assume that this is a regular update.

pkg/controller/elasticsearch/driver/upgrade_pods_deletion.go
```diff
@@ -78,7 +78,39 @@ func (ctx *rollingUpgradeCtx) Delete() ([]corev1.Pod, error) {
...
+	if err := ctx.prepareClusterForNodeRestart(ctx.podsToUpgrade); err != nil {
+		return nil, err
+	}
+
+	for _, podToDelete := range ctx.podsToUpgrade {
+		if err := ctx.handleMasterScaleChange(podToDelete); err != nil {
+			return nil, err
+		}
+		// do not delete any Pods if at least one is not ready for deletion
+		if readyToDelete, err := ctx.readyToDelete(podToDelete); err != nil || !readyToDelete {
```
barkbay, line 105: Should we surface the situation where `!readyToDelete` is true, with a log message for example ? I'll try to see how to deal with it in the context of #5328

pkg/controller/elasticsearch/driver/upgrade.go
pebrc: I understand your argument. But we are already basing upgrade decisions on the existence of this value in the association mechanism where a upgrade is not rolled out if that version is the empty string or the version indicates that an update is still pending for the associated resource. So it feels to me like we are already committed to keeping this up to date in a relatively timely manner and if anything we should make sure that these status updates succeed. 

In addition if we were to fall back to a rolling upgrade, there is a 50% chance that the upgrade will not go through and we/the user have a broken cluster afterwards.

barkbay (approved):
LGTM 👍

PR #5339, opened 2022-02-07
Only configure Stack Monitoring if association reconciled

thbkrkr:
Currently, if a referenced monitoring Elasticsearch does not exist, Stack Monitoring is still configured with empty values in beat configs causing the beats to `CrashLoopBackOff`.

This happens only for ES-ES. For KB-ES, the call to `AllowVersion()` checks that `assoc.AssociationConf() != nil`, so the reconciliation is aborted if the association is not yet configured. I fixed both for consistency.

### DIFF
--- a/pkg/controller/common/stackmon/monitoring/monitoring.go
+++ b/pkg/controller/common/stackmon/monitoring/monitoring.go
@@ -18,11 +18,19 @@ type HasMonitoring interface {
 	MonitoringAssociation(ref commonv1.ObjectSelector) commonv1.Association
 }
 
-func AreAssocConfigured(resource HasMonitoring) bool {
-	return IsMetricsAssocConfigured(resource) && IsLogsAssocConfigured(resource)
+// IsReconcilable return true if a resource has at least one association defined in its specification
+// and all defined associations are configured.
+func IsReconcilable(resource HasMonitoring) bool {
+	return IsDefined(resource) && isMetricsAssocConfigured(resource) && isLogsAssocConfigured(resource)
 }
 
-func IsMetricsAssocConfigured(resource HasMonitoring) bool {
+// IsDefined return true if a resource has at least one association for Stack Monitoring defined in its specification
+// (can be one for logs or one for metrics).
+func IsDefined(resource HasMonitoring) bool {
+	return IsMetricsDefined(resource) || IsLogsDefined(resource)
+}
+
+func isMetricsAssocConfigured(resource HasMonitoring) bool {
 	refs := resource.GetMonitoringMetricsRefs()
 	for _, ref := range refs {
 		if !resource.MonitoringAssociation(ref).AssociationConf().IsConfigured() {
@@ -32,7 +40,7 @@ func IsMetricsAssocConfigured(resource HasMonitoring) bool {
 	return true
 }
 
-func IsLogsAssocConfigured(resource HasMonitoring) bool {
+func isLogsAssocConfigured(resource HasMonitoring) bool {
 	refs := resource.GetMonitoringLogsRefs()
 	for _, ref := range refs {
 		if !resource.MonitoringAssociation(ref).AssociationConf().IsConfigured() {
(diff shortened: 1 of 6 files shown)


thbkrkr:
Jenkins test this please

pkg/controller/elasticsearch/stackmon/beat_config.go
```diff
@@ -36,7 +41,7 @@ func ReconcileConfigSecrets(client k8s.Client, es esv1.Elasticsearch) error {
 		}
 	}
 
-	if monitoring.IsLogsDefined(&es) {
+	if monitoring.IsMetricsDefined(&es) {
```
pebrc, line 44: This looks like an unintentional change?

pkg/controller/kibana/stackmon/sidecar.go
```diff
@@ -69,8 +69,8 @@ func Filebeat(client k8s.Client, kb kbv1.Kibana) (stackmon.BeatSidecar, error) {
 // WithMonitoring updates the Kibana Pod template builder to deploy Metricbeat and Filebeat in sidecar containers
 // in the Kibana pod and injects the volumes for the beat configurations and the ES CA certificates.
 func WithMonitoring(client k8s.Client, builder *defaults.PodTemplateBuilder, kb kbv1.Kibana) (*defaults.PodTemplateBuilder, error) {
-	// no monitoring defined, skip
-	if !monitoring.IsDefined(&kb) {
+	// no monitoring defined or yet configured, skip
+	if !monitoring.IsDefined(&kb) || !monitoring.AreAssocConfigured(&kb) {
```
pebrc, line 73: Given that we use this combined condition multiple times throughout the code base I wonder if it would make sense to have a single method 

```golang
func IsReconcilable(resource HasMonitoring) bool {
    return IsDefined(resource) && AreAssocConfigured(resource)
}
```

and then 
```suggestion
	if !monitoring.IsReconcilable(&kb) {
```

pkg/controller/common/stackmon/monitoring/monitoring.go
```diff
@@ -18,6 +18,30 @@ type HasMonitoring interface {
 	MonitoringAssociation(ref commonv1.ObjectSelector) commonv1.Association
 }
 
+func AreAssocConfigured(resource HasMonitoring) bool {
```
pebrc, line 21:
```suggestion
func AreAssocsConfigured(resource HasMonitoring) bool {
```

maybe?

pkg/controller/elasticsearch/stackmon/beat_config.go
thbkrkr: @#!, yes.

pkg/controller/kibana/stackmon/sidecar.go
thbkrkr: Good idea.

pkg/controller/common/stackmon/monitoring/monitoring.go
```diff
@@ -18,10 +18,38 @@ type HasMonitoring interface {
...
+	return IsDefined(resource) && isMetricsAssocConfigured(resource) && isLogsAssocConfigured(resource)
+}
+
+// IsDefined return true if a resource has at least one association for Stack Monitoring defined in its specification
+// (can be one for logs or one for metrics).
 func IsDefined(resource HasMonitoring) bool {
 	return IsMetricsDefined(resource) || IsLogsDefined(resource)
 }
 
+func isMetricsAssocConfigured(resource HasMonitoring) bool {
```
pebrc, line 33: I guess it is a little bit confusing that this check passes when there is no association configured at all! Not sure if it would help to either add a comment or to add an `Optional` somewhere in the function name. Or maybe something like
`isMetricsAssocConfiguredIfDefined` and
`isLogsAssocConfiguredIfDefined`

pebrc (approved):
LGTM and works I did a quick test, with a release version of ECK versus your PR. 👍

pkg/controller/common/stackmon/monitoring/monitoring.go
thbkrkr: I totally understand. With `IsReconcilable` these two methods are only used once, so I think we can just remove them and inline the code in `IsReconcilable`.

Issue #5330, opened 2022-02-04
ES cluster health reported by the operator might be misleading if a reconciliation error


thbkrkr:
The ES health reported by the operator in the ES resource status may never be updated if the operator encounters an error during the reconciliation loop.

#### Steps to reproduce
- Create an ES cluster
```yaml
apiVersion: v1
kind: Secret
metadata:
  name: bad-ca
stringData:
  ca.crt: bla
  ca.key: bla
---
apiVersion: elasticsearch.k8s.elastic.co/v1
kind: Elasticsearch
metadata:
  name: yellowforever
spec:
  version: 7.16.2
  # transport:                  
  #   tls:
  #     certificate:
  #       secretName: bad-ca
  nodeSets:
  - name: master
    count: 1
    config:
      node.store.allow_mmap: false
```
- Make cluster health yellow
```
# create an index with 1 replica
-XPUT /my-index-1 -d '{"settings":{"number_of_shards":1,"number_of_replicas":1}}'
```
- Configure the bad CA secret in the cluster
``` uncomment transport.tls.certificate.secretName``` and apply.
- Make cluster health green
```
# delete the index
-XDELETE /my-index-1
```

thbkrkr:
The right thing to do seems to be to update the status even when there are errors in the reconcile loop. And that's what we already do. The problem here is that we depend on the bad secret to get the cluster health. We fail to reconcile the certificates required to create the ES observer required to get the cluster health required to update the status. Precisely here, we don't really need the certificates for the transport layer to get cluster health but the certificates for HTTP and transport are reconciled together.

https://github.com/elastic/cloud-on-k8s/blob/4650e61024b3da9ebd6408a02507631b39b7a29b/pkg/controller/elasticsearch/driver/driver.go#L143-L189

I see few things to improve that:
- We could separate the certs reconciliation for HTTP from that for transport so that as soon as we reconciled the HTTP certs, we start the observer and update the status. The problem will still occur if there is an error to reconcile the HTTP certs. So this is just an optimization to reduce the situations in which the problem can occur.
- We could set the status health to `unknown` when there is a reconciliation error before the observer is started.

pebrc:
Is this bug still present given that we have made significant changes to how the status is calculated?

PR #5328, opened 2022-02-04
Improve Elasticsearch status sub-resource

barkbay:
This PR implements an improvement proposal for the status of the Elasticsearch resource.

## [Conditions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties)

The PR introduces the following [conditions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties):

* `ReconciliationComplete`: the reconciliation on this resource can be considered as complete, there is no short term operations pending.
* `RunningDesiredVersion`: all the Pods are running the desired version.
* `ElasticsearchIsReachable`: Internal service has some running Pods 

Note that the proposal here is to have all the conditions to `True` on a healthy and fully reconciled cluster. It makes it easier to check if all the conditions are in a "expected"/"healthy" state.

## In progress operations

If not reconciled, a dedicated parent field `inProgressOperations` provides more details regarding the following topology operations:

### `downscale`

List of nodes which are scheduled to be removed from the cluster. For each node a status field is maintained to surface the shutdown status.

### `upscale`

Contains the list of nodes to be created, note that `Pending` nodes (created but not scheduled yet) may not be included in that list as they may not prevent the reconciliation loop to be considered as "complete"

### `upgrade`

List of nodes that need to be _(maybe forced)_ deleted in the context of an upgrade. For each node a status field is maintained to state if deletion is in progress or if a predicate prevents the node from being deleted.

##  Timestamps

`Conditions` or `inprogressOperations` contains either a `lastUpdatedTime` field, or a `lastTransitionTime` field, to evaluate if an operation is stalled.

## This is a draft

This PR is opened as a draft to give a first idea about the implementation and the status schema. I'm still thinking about a better way to capture the state in some situations. Also, unit tests still need to be added or updated.

## Examples

In order to have a better idea of the proposal here are a few examples of the status in various situations:

<details><summary>Downscale in progress</summary>

```json
{
  "availableNodes": 12,
  "conditions": [
    {
      "lastTransitionTime": "2022-01-28T13:57:08Z",
      "message": "Service test-gcs/elasticsearch-gcs-sample-es-internal-http has endpoints",
      "status": "True",
      "type": "ElasticsearchIsReachable"
    },
    {
      "lastTransitionTime": "2022-02-04T10:14:58Z",
      "message": "Downscale in progress",
      "status": "False",
      "type": "ReconciliationComplete"
    },
    {
[...]

PR #5349, opened 2022-02-09
Improve ES status health reporting

thbkrkr:
2 first commits are more an optimization:
- Split HTTP and transport certs reconciliation in order to be able to retrieve the cluster health despite an issue with the transport certs.
- Start the observer and start it as soon as possible.

Those after could be enough:
- Initialize health state to unknown in the NewState constructor

Relates to #5330.

### Testing

- Create an ES cluster
```yaml
kind: Elasticsearch
metadata:
  name: eight
spec:
  version: 8.0.0
  # http:
  # transport:
  #   tls:
  #     certificate:
  #       secretName: ca-that-does-not-exist
  nodeSets:
  - name: master
    count: 1
    config:
      node.store.allow_mmap: false
```
- Make cluster health yellow
```
# create an index with 1 replica
-XPUT /my-index-1 -d '{"settings":{"number_of_shards":1,"number_of_replicas":1}}'
```
- Break the transport service by configuring a CA Secret that doesn't exist in the cluster

Uncomment `transport.tls.certificate.secretName` and reapply.

- Make cluster health green Again
```
# delete the index
-XDELETE /my-index-1
```

- Cluster should be reported as `green` and not stuck in the `yellow` state, with phase `ApplyingChanges`.

- Do the same by breaking the `http` service. This time the operator will report an `unknown` state as the observer depends on this service, still in phase `ApplyingChanges`.

### DIFF
--- a/pkg/controller/common/certificates/reconcile.go
+++ b/pkg/controller/common/certificates/reconcile.go
@@ -8,7 +8,6 @@ import (
 	"context"
 	"time"
 
-	"go.elastic.co/apm"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
@@ -18,7 +17,6 @@ import (
 	commonv1 "github.com/elastic/cloud-on-k8s/pkg/apis/common/v1"
 	commonname "github.com/elastic/cloud-on-k8s/pkg/controller/common/name"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common/reconciler"
-	"github.com/elastic/cloud-on-k8s/pkg/controller/common/tracing"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common/watches"
 	"github.com/elastic/cloud-on-k8s/pkg/utils/k8s"
 	ulog "github.com/elastic/cloud-on-k8s/pkg/utils/log"
@@ -53,9 +51,6 @@ type Reconciler struct {
 // - a Secret containing the public-facing HTTP certificates (same as the internal one, but without the key)
 // If TLS is disabled, self-signed certificates are still reconciled, for simplicity/consistency, but not used.
 func (r Reconciler) ReconcileCAAndHTTPCerts(ctx context.Context) (*CertificatesSecret, *reconciler.Results) {
-	span, _ := apm.StartSpan(ctx, "reconcile_certs", tracing.SpanTypeApp)
-	defer span.End()
-
 	results := reconciler.NewResult(ctx)
 
 	if !r.TLSOptions.Enabled() && r.GarbageCollectSecrets {
--- a/pkg/controller/common/certificates/reconcile.go
+++ b/pkg/controller/common/certificates/reconcile.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"time"
 
+	"go.elastic.co/apm"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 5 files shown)


mtparet:
Quick question, does it means the operator will also set the status to `unknown` when we excluding an elasticsearch from being managed by the operator using https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-upgrading-eck.html#k8s-beta-to-ga-rolling-restart ?

Because today, if we exclude temporary a ressources, the status is reported is kept also it does not reflect the reality.

thbkrkr:
> does it means the operator will also set the status to unknown when we excluding an elasticsearch from being managed by the operator 

No, it is independent.

> Because today, if we exclude temporary a ressources, the status is reported is kept also it does not reflect the reality.

Could you elaborate on that (maybe in #5330 or a dedicated issue)? What part of the status isn't updated?

mtparet:
> Could you elaborate on that (maybe in https://github.com/elastic/cloud-on-k8s/issues/5330 or a dedicated issue)? What part of the status isn't updated?

I think it's a different issue. 
If we set the annotation `eck.k8s.elastic.co/managed=false` to a resource, the operator will keep the status as previously (say "green" for example) although it does not update anymore the status whatever happens to the elasticsearch cluster.

thbkrkr:
I realized that we can reset the `State` with an empty `ResourcesState`, so before the `ResourcesState` is created and before we have reconciled the services on which it depends on. c370d1e

pkg/controller/elasticsearch/driver/driver.go
```diff
@@ -115,6 +115,10 @@ var _ commondriver.Interface = &defaultDriver{}
 func (d *defaultDriver) Reconcile(ctx context.Context) *reconciler.Results {
 	results := reconciler.NewResult(ctx)
 
+	// reset the Elasticsearch health to 'unknown' so that if reconciliation fails before the observer has had a chance to get it,
+	// we stop reporting a state that may be out of date
+	d.ReconcileState.UpdateElasticsearchPendingUnknownHealth()
```
pebrc, line 120: Would it not be easier to change the `NewState` constructor to initialise the state with `Unknown`?

pebrc (commented):
I am not sure the split in certificate reconciliation is worth the effort (but I am also not opposed to it). 

Regarding the reset to `unknown` I think we could instead just initialise the reconciliation state with `unknown` health in the reconcile state constructor function. 

Another possible improvement could be to update the status sub resource even if the cluster is unmanaged but that should probably be discussed in an issue first.

pkg/controller/elasticsearch/driver/driver.go
thbkrkr: Much better indeed! And this way, we keep the other information (number of nodes, version, ...) that are retrieved from the existing custom resource, therefore still valid.

thbkrkr:
> I am not sure the split in certificate reconciliation is worth the effort (but I am also not opposed to it).

I discussed with barkbay and he is rather in favor of it but will share his opinion after verifying that it goes well with #5328.

pkg/controller/common/certificates/reconcile.go
```diff
@@ -53,9 +51,6 @@ type Reconciler struct {
 // - a Secret containing the public-facing HTTP certificates (same as the internal one, but without the key)
 // If TLS is disabled, self-signed certificates are still reconciled, for simplicity/consistency, but not used.
 func (r Reconciler) ReconcileCAAndHTTPCerts(ctx context.Context) (*CertificatesSecret, *reconciler.Results) {
-	span, _ := apm.StartSpan(ctx, "reconcile_certs", tracing.SpanTypeApp)
```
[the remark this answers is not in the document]
thbkrkr: It is mistake, I'll revert. With the certificate reconciliation split, I start a span in `ReconcileHTTP` so I thought we could remove that one but I forgot all the other uses of `ReconcileCAAndHTTPCerts`.

PR #5352, opened 2022-02-10
Add note about auto-detecting security context

pebrc:
I forgot to add a note about the security context auto-detection complications on OpenShift


alaudazzi (approved):
LGTM

PR #5382, opened 2022-02-15
Add v2.0.0 to upgrade harness

pebrc:
This adds version 2.0.0 of the operator and Elastic Stack 8.0 to the upgrade harness. It also bumps the stack version for 1.9.1 to 7.17.0 to create a valid upgrade path. 

I also added 
```yaml 
    podTemplate:
      spec:
        containers:
        - name: elasticsearch
          env:
          - name: PRE_STOP_ADDITIONAL_WAIT_SECONDS
            value: "0"
```
to the Elasticsearch manifests to speed up the overall runtime of the upgrade test. I do not expect any negative impact on test outcomes as we do not make any requests to Elasticsearch node that could be cut short by terminating them earlier. If any requests originating in the operator would be affected I would expect the operator to be able to deal with them.

### DIFF
--- a/hack/upgrade-test-harness/conf.yaml
+++ b/hack/upgrade-test-harness/conf.yaml
@@ -43,7 +43,10 @@ testParams:
     stackVersion: 7.16.0
   - name: v191
     operatorVersion: 1.9.1
-    stackVersion: 7.16.1
-  - name: upcoming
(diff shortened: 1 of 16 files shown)


barkbay (approved):
LGTM

Issue #3392, opened 2020-07-03
Set .status.observedGeneration


benlangfeld:
## Proposal

In order for deployment tools like [krane](https://github.com/Shopify/krane#deploying-custom-resources) to deterministically monitor the rollout of custom resources, those resources [must set `.status.observedGeneration`](https://github.com/Shopify/krane/issues/472).

This is a feature request for exactly that in this operator, such that our automation may effectively monitor rollout.

sebgl:
It sounds like a good idea to me.

One important thing to discuss though:

The purpose of `status.observedGeneration`, as I understand it, is for people to compare with `metadata.generation`, so you know whether the `Status` you are looking at is up-to-date with the object you're looking at.
It does *not* mean that the underlying resources (eg. Pods) match the specification of the resource.
For example in a StatefulSet, seeing `metadata.generation == status.observedGeneration` means the controller has processed the StatefulSet resource with that generation, but it does not mean all underlying Pods are running.

I think it's an important distinction. Let's take an example with this Elasticsearch resource, where let's say we increased `count` from 1 to 3:

```yaml
apiVersion: elasticsearch.k8s.elastic.co/v1
  kind: Elasticsearch
  metadata:
    name: mycluster
    generation: 12
  spec:
    version: 7.8.0
    nodeSets:
      count: 3
      name: default
  status:
    availableNodes: 1
    health: green
    phase: ApplyingChanges
    observedGeneration: 12
```

We see here that ECK has processed the resource in generation 12 (`observedGeneration: 12`). But it does **not** mean the desired 3-nodes cluster is up and running (see how `availableNodes` is still `1`). So it does not mean the "rollout" of the new Elasticsearch specification (from 1 to 3) is complete yet.
In order for a check on `status.observedGeneration` to be useful, it must be combined with another check in the `status` section (eg. `status.phase`). It requires us to have strong consistency in all fields of the status. I'm not sure that's the case currently, but we should aim for it.

benlangfeld:
Yes, of course. If you check out the krane documentation I linked to you’ll see how that’s done. It’s easy and we’re ready to do it in our use case, but blocked on this requirement of the operator; that’s why I opened this feature request, not because I believe it’s the only requirement :)

benlangfeld:
[...]

PR #5383, opened 2022-02-15
 Elastic Agent: Set status.ObservedGeneration from metadata.Generation

naemono:
This change is Elastic Agent specific at the moment

related to https://github.com/elastic/cloud-on-k8s/issues/3392
related to https://github.com/elastic/cloud-on-k8s/pull/5331

This change allows status.ObservedGeneration to be kept in sync with metadata.Generation according to the rules described in the above issue. This change adds a set of unit tests which show more details into when observedGeneration will, and will not be updated according to any errors that may occur during reconciliation. An end-to-end test is also added to show how the change is intended to function in a true cluster.


pkg/controller/agent/controller.go
```diff
@@ -157,19 +160,54 @@ func (r *ReconcileAgent) doReconcile(ctx context.Context, agent agentv1alpha1.Ag
...
+				"agent_name", params.Agent.Name)
+			return results.WithResult(reconcile.Result{Requeue: true})
+		} else if err != nil {
+			params.Logger().Error(
+				err,
+				"Error while updating status",
+				"namespace", params.Agent.Namespace,
+				"agent_name", params.Agent.Name,
+			)
+			return results.WithError(errors.Wrapf(err, "while updating status: %s", statusErr))
```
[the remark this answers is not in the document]
pebrc:
```golang
func (r *ReconcileAgent) doReconcile(ctx context.Context, agent agentv1alpha1.Agent) (results *reconciler.Results) {
	defer tracing.Span(&ctx)()
        status := newStatus(agent)
	results = reconciler.NewResult(ctx)
	if !association.AreConfiguredIfSet(agent.GetAssociations(), r.recorder) {
		return results
	}

	params := Params{
		Context:        ctx,
		Client:         r.Client,
		EventRecorder:  r.recorder,
		Watches:        r.dynamicWatches,
		Agent:          agent,
		OperatorParams: r.Parameters,
	}	

	defer func() {
		if err := updateStatus(params, status); err != nil && apierrors.IsConflict(err) {
			params.Logger().V(1).Info(
				"Conflict while updating status",
				"namespace", params.Agent.Namespace,
				"agent_name", params.Agent.Name)
			results = results.WithResult(reconcile.Result{Requeue: true})
		} else if err != nil {
			params.Logger().Error(
				err,
				"Error while updating status",
				"namespace", params.Agent.Namespace,
				"agent_name", params.Agent.Name,
			)
			results = results.WithError(err)
		}
	}()

	// Run basic validations as a fallback in case webhook is disabled.
	if err := r.validate(ctx, agent); err != nil {
		return results.WithError(err)
	}

	return results.WithResults(internalReconcile(params, &status))
}
```

pkg/controller/agent/controller_test.go
```diff
@@ -0,0 +1,151 @@
...
+func newReconcileAgent(objs ...runtime.Object) *ReconcileAgent {
+	r := &ReconcileAgent{
+		Client:   k8s.NewFakeClient(objs...),
+		recorder: record.NewFakeRecorder(100),
+	}
+	return r
+}
+
+func TestReconcileAgent_Reconcile(t *testing.T) {
+	type k8sFields struct {
```
pebrc, line 34: Nit: struct seems redundant and/or could be named differently to highlight this is the API Server state fixture.

pkg/controller/agent/controller.go
```diff
@@ -18,6 +19,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
+	"github.com/pkg/errors"
```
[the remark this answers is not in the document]
naemono: weird my editor didn't correct this.  I adjusted my settings.

pkg/controller/agent/controller.go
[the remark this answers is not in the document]
naemono: I went with deferring, as you suggested here.  Makes sense.

test/e2e/agent/observed_generation_test.go
```diff
@@ -0,0 +1,85 @@
...
+	"testing"
+
+	agentv1alpha1 "github.com/elastic/cloud-on-k8s/pkg/apis/agent/v1alpha1"
+	"github.com/elastic/cloud-on-k8s/pkg/utils/k8s"
+	"github.com/elastic/cloud-on-k8s/test/e2e/test"
+	"github.com/elastic/cloud-on-k8s/test/e2e/test/agent"
+	"github.com/elastic/cloud-on-k8s/test/e2e/test/elasticsearch"
+)
+
+func TestAgentObservedGenerationIncrement(t *testing.T) {
```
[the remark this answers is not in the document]
naemono: I moved these Generation steps to their own e2e package as they're repetitive across all of the tests, and only need metav1.Object interface, and made the agent steps use it.  If this works, then I'll also make a follow up PR to the ES changes to just use this package.

pkg/controller/agent/reconcile.go
```diff
@@ -128,23 +134,36 @@ func reconcileDaemonSet(rp ReconciliationParams) (int32, int32, error) {
...
 	client      k8s.Client
 	agent       agentv1alpha1.Agent
 	podTemplate corev1.PodTemplateSpec
 }
 
-func updateStatus(params Params, ready, desired int32) error {
+// calculateStatus will calculate a new status from the state of the pods within the k8s cluster
+// and will return the new status, and any errors encountered.
+func calculateStatus(params *Params, ready, desired int32) (agentv1alpha1.AgentStatus, error) {
+	status := newStatus(params.Agent)
```
barkbay, line 147: nit: a new status has already been created at the beginning of the reconciliation loop and a pointer is available in the context of `reconcilePodVehicle`. Should we just use it instead of creating a new one and replacing the content ?

```suggestion
// calculateStatus updates the current status using the state of the pods within the k8s cluster.
func calculateStatus(params *Params, ready, desired int32, status *agentv1alpha1.AgentStatus) error {
```

test/e2e/test/generation/generation.go
```diff
@@ -0,0 +1,92 @@
...
+		return 0, err
+	}
+
+	switch v := obj.(type) {
+	case *agentv1alpha1.Agent:
+		return v.Status.ObservedGeneration, nil
+	default:
+		return 0, fmt.Errorf("unsupported type while retrieving status.observedGeneration: %T", v)
+	}
+}
```
barkbay, line 38: Assuming we have the following interface `HasObservedGeneration` in `pkg/apis/common/v1/common.go`:

```golang
type HasObservedGeneration interface {
	client.Object
	GetObservedGeneration() int64
}
```

And that interface implemented by both Elasticsearch and Agent, for example in `pkg/apis/elasticsearch/v1/elasticsearch_types.go`:

```golang
func (es Elasticsearch) GetObservedGeneration() int64 {
	return es.Status.ObservedGeneration
}
```

We could have:

```suggestion
func getObservedGeneration(object commonv1.HasObservedGeneration, k *test.K8sClient) (int64, error) {
	if err := k.Client.Get(context.Background(), k8s.ExtractNamespacedName(object), object); err != nil {
		return 0, err
	}
	return object.GetObservedGeneration(), nil
}
```

Which would allow to have a compile-time check.

pkg/controller/agent/controller.go
```diff
@@ -153,25 +153,44 @@ func (r *ReconcileAgent) Reconcile(ctx context.Context, request reconcile.Reques
 func (r *ReconcileAgent) doReconcile(ctx context.Context, agent agentv1alpha1.Agent) *reconciler.Results {
 	defer tracing.Span(&ctx)()
 	results := reconciler.NewResult(ctx)
+	status := newStatus(agent)
 	if !association.AreConfiguredIfSet(agent.GetAssociations(), r.recorder) {
```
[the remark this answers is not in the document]
naemono: This logic has been updated, and I'll get these steps tested.

pkg/controller/agent/reconcile.go
```diff
@@ -128,23 +134,36 @@ func reconcileDaemonSet(rp ReconciliationParams) (int32, int32, error) {
...
 
-	return params.Client.Status().Update(context.Background(), &agent)
+// updateStatus will update the Elastic Agent's status within the k8s cluster, using the Elastic Agent from the
+// given params, and the given status.
+func updateStatus(params Params, status agentv1alpha1.AgentStatus) error {
+	if reflect.DeepEqual(params.Agent.Status, status) {
+		return nil
+	}
+	params.Agent.Status = status
+	return params.Client.Status().Update(context.Background(), &params.Agent)
```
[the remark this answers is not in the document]
naemono: 👍

pkg/controller/agent/reconcile.go
naemono: 👍

test/e2e/test/generation/generation.go
naemono: I added this logic to both the ES, and Kibana PRs we merged before this one, and I updated this to have the same logic.

naemono:
Last update resolves this issue with associations @barkbay 

initial
```
❯ kc get agent.agent.k8s.elastic.co/fleet-server -o go-template='{{ printf "\nGeneration: "}}{{.metadata.generation}}{{printf "\n\nStatus:"}}{{.status}}'

Generation: 2

Status:map[availableNodes:1 elasticsearchAssociationsStatus:map[default/elasticsearch:Established] expectedNodes:1 health:green kibanaAssociationStatus:Established observedGeneration:2 version:8.0.0]
```

broken kibana assocciation
```
❯ kc get agent.agent.k8s.elastic.co/fleet-server -o go-template='{{ printf "\nGeneration: "}}{{.metadata.generation}}{{printf "\n\nStatus:"}}{{.status}}'

Generation: 3

Status:map[availableNodes:1 elasticsearchAssociationsStatus:map[default/elasticsearch:Established] expectedNodes:1 health:green kibanaAssociationStatus:Pending observedGeneration:3 version:8.0.0]
```

spec updated, observedGeneration updated:
```
❯ kc get agent.agent.k8s.elastic.co/fleet-server -o go-template='{{ printf "\nGeneration: "}}{{.metadata.generation}}{{printf "\n\nStatus:"}}{{.status}}'

Generation: 4

Status:map[availableNodes:1 elasticsearchAssociationsStatus:map[default/elasticsearch:Established] expectedNodes:1 health:green kibanaAssociationStatus:Pending observedGeneration:4 version:8.0.0]
```

pkg/controller/association/reconciler.go
```diff
@@ -209,11 +209,18 @@ func (r *Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
...
-			return results.WithResult(reconcile.Result{Requeue: true}).Aggregate()
-		}
+	if err := r.updateStatus(ctx, associated, newStatusMap); err != nil && apierrors.IsConflict(err) {
+		log.V(1).Info(
+			"Conflict while updating status",
+			"namespace", associatedKey.Namespace,
+			"name", associatedKey.Name)
+		return results.WithResult(reconcile.Result{Requeue: true}).Aggregate()
+	} else if err != nil {
+		log.Error(
```
pebrc, line 219: I think this will lead to the error being logged twice. Once here and once through controller-runtime. I think it would be preferable to maybe wrap the error instead with some context information and stack trace e.g. through `pkg/errors` and pass it on to be logged on the top level.

pkg/controller/agent/controller.go
```diff
@@ -121,21 +121,24 @@ type ReconcileAgent struct {
 
 // Reconcile reads that state of the cluster for an Agent object and makes changes based on the state read
 // and what is in the Agent.Spec
-func (r *ReconcileAgent) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
+func (r *ReconcileAgent) Reconcile(ctx context.Context, request reconcile.Request) (result reconcile.Result, err error) {
```
pebrc, line 124:
```suggestion
func (r *ReconcileAgent) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
```

Maybe left over from a previous iteration of this PR?

pkg/controller/agent/controller.go
```diff
@@ -144,22 +147,35 @@ func (r *ReconcileAgent) Reconcile(ctx context.Context, request reconcile.Reques
...
-	results := reconciler.NewResult(ctx)
+	var err error
+	results = reconciler.NewResult(ctx)
+	status := newStatus(agent)
+
+	// defer the updating of status to ensure that the status is updated regardless of the outcome of the reconciliation.
+	// note that this deferred function is modifying the return values, which are named return values, which allows this
+	// to function properly.
+	defer func() {
+		if updateStatusresults := updateStatus(ctx, agent, r.Client, status).WithError(err); updateStatusresults != nil {
```
pebrc, line 166: It is not clear to me why we want to call `WithError` here with variable `err` we closed over in the defer. Either that variable will be `nil` in which case the `WithError` call is redundant, or the variable points to an error in which case we  adding the same error twice as we already called `WithError` for example on L177 (which is the right thing to do). 

Instead I would close over the `results` variable and pass it in to `updateStatus` so you don't have to create. new `result` and you preserve all the results and errors that have potentially been produced further down the call stack. In your current implementation you will lose any error or result returned from functions from `doReconcile` and you only capture the `err` variable in this function. If you don't want to pass in the results into `updateStatus` you need to at least merge them with the results produced by the main function body.

pebrc (changes requested):

pkg/controller/agent/controller.go
```diff
@@ -121,21 +121,24 @@ type ReconcileAgent struct {
...
+	if err = association.FetchWithAssociations(ctx, r.Client, request, agent); err != nil {
 		if apierrors.IsNotFound(err) {
 			r.onDelete(request.NamespacedName)
 			return reconcile.Result{}, nil
 		}
-		return reconcile.Result{}, tracing.CaptureError(ctx, err)
+		if agent == nil {
+			return reconcile.Result{}, err
+		}
+		return updateStatus(ctx, *agent, r.Client, newStatus(*agent)).WithError(err).Aggregate()
```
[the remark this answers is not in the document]
naemono: I've just removed updating the status on association failures, as it seems to be causing more issues than it's worth.

pkg/controller/agent/reconcile.go
```diff
@@ -128,23 +134,44 @@ func reconcileDaemonSet(rp ReconciliationParams) (int32, int32, error) {
...
+	status.Health = CalculateHealth(agent.GetAssociations(), ready, desired)
+	status.Version = common.LowestVersionFromPods(status.Version, pods, VersionLabelName)
+	return nil
+}
 
-	return params.Client.Status().Update(context.Background(), &agent)
+// updateStatus will update the Elastic Agent's status within the k8s cluster, using the Elastic Agent from the
+// given params, and the given status.
+func updateStatus(ctx context.Context, agent agentv1alpha1.Agent, client client.Client, status agentv1alpha1.AgentStatus) *reconciler.Results {
+	results := reconciler.NewResult(ctx)
```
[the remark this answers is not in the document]
naemono: refactored to just return an error again.

pkg/controller/agent/reconcile.go
```diff
@@ -128,27 +131,38 @@ func reconcileDaemonSet(rp ReconciliationParams) (int32, int32, error) {
...
-	agent.Status.ExpectedNodes = desired
+	status.AvailableNodes = ready
+	status.ExpectedNodes = desired
 	health, err := CalculateHealth(agent.GetAssociations(), ready, desired)
 	if err != nil {
 		return err
 	}
 	agent.Status.Health = health
-	agent.Status.Version = common.LowestVersionFromPods(agent.Status.Version, pods, VersionLabelName)
+	status.Version = common.LowestVersionFromPods(status.Version, pods, VersionLabelName)
```
pebrc, line 157: Move this and `PodsMatchingLabels` closer together.

pkg/controller/agent/reconcile.go
```diff
@@ -128,27 +131,38 @@ func reconcileDaemonSet(rp ReconciliationParams) (int32, int32, error) {
...
 	}
-	agent.Status.AvailableNodes = ready
-	agent.Status.ExpectedNodes = desired
+	status.AvailableNodes = ready
+	status.ExpectedNodes = desired
 	health, err := CalculateHealth(agent.GetAssociations(), ready, desired)
 	if err != nil {
 		return err
 	}
 	agent.Status.Health = health
```
pebrc, line 156: This update is lost IIUC. You are not updating `status`. Missing unit test?
```suggestion
	status.Health = health
```

pkg/controller/agent/reconcile.go
```diff
@@ -75,10 +75,13 @@ func reconcilePodVehicle(params Params, podTemplate corev1.PodTemplateSpec) *rec
 		results.WithError(err)
 	}
 
-	err = updateStatus(params, ready, desired)
-	if err != nil && apierrors.IsConflict(err) {
-		params.Logger().V(1).Info("Conflict while updating status")
-		return results.WithResult(reconcile.Result{Requeue: true})
+	err = calculateStatus(&params, ready, desired, status)
+	if err != nil {
```
pebrc, line 79:
```suggestion
	if err := calculateStatus(&params, ready, desired, status); err != nil {
```

pkg/controller/agent/reconcile.go
```diff
@@ -75,10 +75,13 @@ func reconcilePodVehicle(params Params, podTemplate corev1.PodTemplateSpec) *rec
...
+		params.Logger().Error(
+			err, "Error while calculating new status",
+			"namespace", params.Agent.Namespace,
+			"agent_name", params.Agent.Name,
+		)
```
pebrc, line 84: As mentioned elsewhere this logged twice once by controller runtime and once here. Consider: 

`pkgerrors.Wrap(err, "while calculating status") `

pkg/controller/agent/controller.go
```diff
@@ -144,26 +144,39 @@ func (r *ReconcileAgent) Reconcile(ctx context.Context, request reconcile.Reques
 		return reconcile.Result{}, nil
 	}
 
-	res, err := r.doReconcile(ctx, agent).Aggregate()
-	k8s.EmitErrorEvent(r.recorder, err, &agent, events.EventReconciliationError, "Reconciliation error: %v", err)
+	results, status := r.doReconcile(ctx, *agent)
 
-	return res, err
+	statusErr := updateStatus(*agent, r.Client, status)
```
pebrc, line 149: Could just be called `err`?

pkg/controller/agent/reconcile.go
```diff
@@ -27,7 +27,7 @@ import (
 	"github.com/elastic/cloud-on-k8s/pkg/utils/pointer"
 )
 
-func reconcilePodVehicle(params Params, podTemplate corev1.PodTemplateSpec) *reconciler.Results {
+func reconcilePodVehicle(params Params, podTemplate corev1.PodTemplateSpec, status *agentv1alpha1.AgentStatus) *reconciler.Results {
```
pebrc, line 30: What's a bit messy here is that on the one hand we have a more functional approach where results are returned and not pointers manipulated:
```golang
reconcilePodVehicle(...) *reconcilerResults
```

and now we are introducing status here that is following another idiom where a pointer to a struct is passed and the underlying struct is manipulated: sort of a hidden return value if you will. 

```golang
reconcilePodVehicle(&status)
```

pkg/controller/agent/reconcile.go
naemono: You were correct.  Unit test added for status.health.

pkg/controller/agent/reconcile.go
naemono: I see your point here.  I have refactored this to return both a status and a results pointer throughout the reconciliation.  Lmk if that is what you were intending.

pkg/controller/agent/reconcile.go
```diff
@@ -128,27 +128,38 @@ func reconcileDaemonSet(rp ReconciliationParams) (int32, int32, error) {
...
 type ReconciliationParams struct {
 	client      k8s.Client
 	agent       agentv1alpha1.Agent
 	podTemplate corev1.PodTemplateSpec
 }
 
-func updateStatus(params Params, ready, desired int32) error {
+// calculateStatus will calculate a new status from the state of the pods within the k8s cluster
+// and will return any error encountered.
+func calculateStatus(params *Params, ready, desired int32, status *agentv1alpha1.AgentStatus) error {
```
pebrc, line 140: If you change all the other ones to return `status` I would probably change this one as well for consistency.

pebrc (commented):
Almost LGTM!

pkg/controller/agent/controller.go
```diff
@@ -144,38 +144,48 @@ func (r *ReconcileAgent) Reconcile(ctx context.Context, request reconcile.Reques
...
+	}
+
+	result, err := results.Aggregate()
+	k8s.EmitErrorEvent(r.recorder, err, agent, events.EventReconciliationError, "Reconciliation error: %v", err)
+
+	return result, err
 }
 
-func (r *ReconcileAgent) doReconcile(ctx context.Context, agent agentv1alpha1.Agent) *reconciler.Results {
+func (r *ReconcileAgent) doReconcile(ctx context.Context, agent agentv1alpha1.Agent) (agentv1alpha1.AgentStatus, *reconciler.Results) {
```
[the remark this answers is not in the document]
naemono: 👍

pebrc (approved):
LGTM!

Issue #4903, opened 2021-10-01
Test Elastic Agent recipes using Stack version under test


david-kow:
Currently, in our E2E tests for Elastic Agent recipes we use the version that is in the recipe. This makes us unable to test snapshot versions of the Stack. We should overwrite the version in the recipe manifest and use the Stack version under test.

PR #5388, opened 2022-02-16
Adjust validations in TestMetricbeatStackMonitoringRecipe for version 8.0.0

### LINKED ISSUE #5387: Update E2E assertions for Beats >= 8.x
https://github.com/elastic/cloud-on-k8s/pull/5375 highlighted that the structure of the documents written in Elasticsearch by Metricbeat for Stack Monitoring have changed starting 8.0.0.

https://devops-ci.elastic.co/job/cloud-on-k8s-pr-full/114/ :red_circle: :
```
Test:       	TestMetricbeatStackMonitoringRecipe/ES_data_should_pass_validations
  hit count should be more than 0 for /.monitoring-*/_search?q=type:cluster_stats
```

I reproduced and yes I see that the field `type:cluster_stats is no more in the documents.
```
GET /.monitoring-*/_search?q=type:cluster_stats" | jq '.hits.total.value'     
0                                                                                                                                                                                                                                                                                  
GET /.monitoring-*/_search?q=metricset.name:cluster_stats" | jq '.hits.total.value'
129
```

It goes further, we have to reconsider all the assertions we were doing here:

https://github.com/elastic/cloud-on-k8s/blob/721cccfd905f43c345c7957fbfe173ea3ebb6bbf/test/e2e/beat/recipes_test.go#L131-L139

thbkrkr:
Resolves #5387.

We could run queries that will work in 7x and 8x using bool filter but I prefer to have the exact queries for each version.

The Stack version in the recipes is currently fixed but it will change after #4903. This shouldn't be a problem because we have the version of the resource from the constructor that has already been transformed.

### DIFF
--- a/test/e2e/beat/recipes_test.go
+++ b/test/e2e/beat/recipes_test.go
@@ -134,7 +134,6 @@ func TestMetricbeatStackMonitoringRecipe(t *testing.T) {
 			beat.HasMonitoringEvent("kibana_stats.kibana.status:green"),
 		}
 
-		fmt.Println("DEBUG", builder.Beat.Spec.Version)
 		if version.MustParse(builder.Beat.Spec.Version).LT(version.MinFor(8, 0, 0)) {
 			// before 8.0.0, `metricset.name` was not indexed
 			metricbeatValidations = []beat.ValidationFunc{
--- a/test/e2e/beat/recipes_test.go
+++ b/test/e2e/beat/recipes_test.go
@@ -123,24 +123,39 @@ func TestMetricbeatStackMonitoringRecipe(t *testing.T) {
 			}
 		}
 
-		return builder.
-			WithRoles(beat.PSPClusterRoleName).
-			WithESValidations(
-				// metricbeat validations
-				// TODO: see if we can add validation for ccr, ml_job, and shard metricsets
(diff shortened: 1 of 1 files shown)


test/e2e/beat/recipes_test.go
```diff
@@ -123,24 +123,39 @@ func TestMetricbeatStackMonitoringRecipe(t *testing.T) {
...
+			beat.HasMonitoringEvent("metricset.name:enrich"),
+			beat.HasMonitoringEvent("metricset.name:index"),
+			beat.HasMonitoringEvent("metricset.name:index_summary"),
+			beat.HasMonitoringEvent("metricset.name:node_stats"),
+			beat.HasMonitoringEvent("metricset.name:stats"),
+			beat.HasMonitoringEvent("metricset.name:shard"),
+			beat.HasMonitoringEvent("kibana_stats.kibana.status:green"),
+		}
+
+		fmt.Println("DEBUG", builder.Beat.Spec.Version)
```
pebrc, line 137: left over debugging?

pebrc (approved):

test/e2e/beat/recipes_test.go
thbkrkr: Yep :confused:

PR #5393, opened 2022-02-17
Bug fixes to OLM generation identified during 2.0 release

pebrc:
These are small tweaks I identified as necessary during the most recent ECK release. Red Hat's certification pipeline runs a YAML linter on the submitted manifests now that failed a couple of times on the manifests generated by our tooling, due to trailing whitespace or additional newlines. 

This PR: 
* removes trailing whitespace from templates where applicable
* tries to normalize newlines add the end of YAML documents
* adds a bit more context to error messages
* removes the required config argument from the tool and instead just uses the config file we always use by default (overrides are still possible)
* removes the requirement to have a clean working directory before running the generator, instead it just wipes previous output and starts over


hack/operatorhub/config.yaml
```diff
@@ -1,6 +1,6 @@
-newVersion: 1.9.0
-prevVersion: 1.8.0
-stackVersion: 7.16.0
+newVersion: 2.0.0
+prevVersion: 1.9.1
+stackVersion: 8.0.0
```
thbkrkr, line 3: Just a thought. This file will stay like this in `main`, isn't it? I know we never did that before but we could already set the future versions.
```suggestion
newVersion: 2.1.0-SNAPSHOT
prevVersion: 2.0.0
stackVersion: 8.1.0-SNAPSHOT
```

Then when we cut a release branch, we just have to remove the `-SNAPSHOT` like in https://github.com/elastic/cloud-on-k8s/pull/5256.

PR #5396, opened 2022-02-17
remove erroneous error check in finalizers RemoveAll

naemono:
Upon looking at #3392 for APM, came upon this error check that seems to be unnecessary.  This simply removes it.


thbkrkr (commented):
:smile_cat:

thbkrkr (approved):
:smile_cat:

PR #5398, opened 2022-02-20
Extend full upgrade to any 8x version upgrade of non-HA ES

pebrc:
Follow up to https://github.com/elastic/cloud-on-k8s/pull/5327 

I originally tested #5327 with 7.x and 8.0 upgrades. I did not however re-test before merging that the upgrades also work successfully with the by then released 8.0 on the 8.0->8.1 upgrade path. As evidenced by recent CI failures this is not the case. The behaviour we observed with major version upgrades from 7.last to 8.0 is also observable during minor upgrades from 8.0 to 8.1: i.e. depending on which node is currently the master node the cluster can disintegrate in such a way that a rolling upgrade with its invariant of green or yellow cluster health between node restarts cannot proceed. 

This PR extends therefore the condition that enables upgrades through simultaneous restarts of all nodes to any version 8.x upgrade in a non-HA cluster.

For 6.8 to 7.x we want to keep the previous behaviour because we need to bootstrap the new Zen2 mechanisms correctly. 

An alternative implementation to fix this issue could be as follows (I am open to that as well):

* do full simultaneous restarts on any 1/2 node cluster upgrade
* change the mechanisms that bootstraps Zen2 to also set `initial_master_ nodes` on 2 node cluster upgrades to cover the 6.x  to 7.x case

This would give us a more uniform approach across Elastic stack versions with less version specific exceptions.

### DIFF
--- a/test/e2e/es/version_upgrade_test.go
+++ b/test/e2e/es/version_upgrade_test.go
@@ -207,7 +207,10 @@ func TestVersionUpgradeTwoNodesToLatest8x(t *testing.T) {
 
 	initial := elasticsearch.NewBuilder("test-version-up-2-to-8x").
 		WithVersion(srcVersion).
-		WithChangeBudget(2, 2). // 8x non-HA upgrades cannot honour a change budget with maxUnavailable 1
+		// 8x non-HA upgrades cannot honour a change budget with maxUnavailable 1 because we are rolling both nodes at once
+		// setting the change budget accordingly allows this test to pass as otherwise the changeBudgetWatcher step would
+		// fail as it would detect a change budget violation.
+		WithChangeBudget(2, 2).
 		WithESMasterDataNodes(2, elasticsearch.DefaultResources)
 
 	mutated := initial.WithNoESTopology().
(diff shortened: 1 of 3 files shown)


barkbay (approved):
LGTM

pebrc:
@barkbay I had to adjust an e2e test to account for the new behaviour  and wanted to run this by you before we merge.

test/e2e/es/version_upgrade_test.go
```diff
@@ -207,6 +207,7 @@ func TestVersionUpgradeTwoNodesToLatest8x(t *testing.T) {
 
 	initial := elasticsearch.NewBuilder("test-version-up-2-to-8x").
 		WithVersion(srcVersion).
+		WithChangeBudget(2, 2). // 8x non-HA upgrades cannot honour a change budget with maxUnavailable 1
```
barkbay, line 210: I would explain in the comment that this is to "circumvent" the [`changeBudgetWatcher`](https://github.com/elastic/cloud-on-k8s/blob/780831adab3c88e57666d7a1659a589a4a2b5d70/test/e2e/test/elasticsearch/steps_mutation.go#L83). Having a change budget with `maxUnavailable` set to 1 does not prevent the operator to restart all the Pods in this particular case (which should maybe stated in the [documentation](https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-update-strategy.html#k8s-update-strategy) ?)

barkbay (approved):
`TestVersion*` (including `ToLatest8x`) are ✅ locally

test/e2e/es/version_upgrade_test.go
pebrc: I am going to do the documentation update in a separate PR to get this merged today.

PR #5401, opened 2022-02-22
Small changes 

miriam-eid:
This PR removes the unused function ObjectMetas and changes the _for_ loop that adds the names in the [ReconcileTransportCertificatesSecrets](https://github.com/elastic/cloud-on-k8s/blob/9d375cc128b52cef1e6d784b95b39e8b951e0344/pkg/controller/elasticsearch/certificates/transport/reconcile.go#L37) function.


barkbay (approved):
LGTM

Issue #5047, opened 2021-11-11
TestFleetMode/ES_data_should_pass_validations is flaky


pebrc:
```
  step.go:43: 
        	Error Trace:	utils.go:87
        	Error:      	Received unexpected error:
        	            	elasticsearch client failed for https://test-agent-fleet-85gp-es-http.e2e-05uba-mercury.svc:9200/_data_stream/logs-elastic_agent.fleet_server-default: 404 Not Found: {Status:404 Error:{CausedBy:{Reason: Type:} Reason:no such index [logs-elastic_agent.fleet_server-default] Type:index_not_found_exception RootCause:[{Reason:no such index [logs-elastic_agent.fleet_server-default] Type:index_not_found_exception}]}}
        	Test:       	TestFleetMode/ES_data_should_pass_validations
```

Looks like we are hitting a known issue with initialisation of Fleet policies in Kibana as mentioned here 
https://www.elastic.co/guide/en/fleet/7.14/release-notes-7.14.0.html

PR #5406, opened 2022-02-23
[E2E] Fleet agent: use explicit policies

barkbay:
This is an attempt to fix https://github.com/elastic/cloud-on-k8s/issues/5047 (at least the error we are currently hitting: `unable to find policy named ""`)

Both `TestAgentVersionUpgradeToLatest8x` and `TestFleetMode` are ✅ locally with `8.0.0` and `8.1.0-SNAPSHOT`. Still testing with other versions...

### DIFF
--- a/test/e2e/agent/config.go
+++ b/test/e2e/agent/config.go
@@ -5,6 +5,47 @@
 package agent
 
 const (
+	E2EFleetPolicies = `
+    xpack.fleet.packages:
+    - name: system
+      version: latest
+    - name: elastic_agent
+      version: latest
+    - name: fleet_server
+      version: latest
+    - name: kubernetes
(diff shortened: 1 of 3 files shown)


pebrc (approved):
LGTM, excellent find. I think this is due to the default policies being removed from Kibana in 8x.

barkbay:
`TestFleetMode` is still failing using `7.14.2` and `7.15.2` with  ` FATAL  Error: [config validation of [xpack.fleet].agentPolicies.0.package_policies.0.id]: definition for this key is missing` 😓 

It seems to be ok starting with `7.16.3`, I added a conditional statement to fallback on the current behaviour for versions before `7.16`

pebrc (approved):
LGTM

PR #5413, opened 2022-02-25
Fix lost downscale state

pebrc:
Recent test failures point to a bug whereby we lose downscale state due to copying which becomes apparent if multiple StatefulSets with masters for example have to be processed in the downscale calculation. 

This fixes the issue by using a pointer and extends the existing test to have two master StatefulSets.

### DIFF
--- a/pkg/controller/elasticsearch/driver/downscale.go
+++ b/pkg/controller/elasticsearch/driver/downscale.go
@@ -158,7 +158,7 @@ func calculateDownscales(
 
 type downscaleFilter func(_ *downscaleState, _ appsv1.StatefulSet, _ int32) int32
 
-// noDownscaleFilter is a filter which does no remove any Pod. It can be used to compute the full list of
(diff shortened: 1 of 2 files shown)


pkg/controller/elasticsearch/driver/downscale.go
```diff
@@ -156,19 +156,19 @@ func calculateDownscales(
 	return downscales, deletions
 }
 
-type downscaleFilter func(_ downscaleState, _ appsv1.StatefulSet, _ int32) int32
+type downscaleFilter func(_ *downscaleState, _ appsv1.StatefulSet, _ int32) int32
 
 // noDownscaleFilter is a filter which does no remove any Pod. It can be used to compute the full list of
```
barkbay, line 161:
```suggestion
// noDownscaleFilter is a filter which does not remove any Pod. It can be used to compute the full list of
```

barkbay (approved):
LGTM
Thanks for fixing it 🙇

PR #5416, opened 2022-02-28
Remove unused variables

thbkrkr:
Since we migrated to use dedicated images inside the deployer for Kind, we no longer install Kind binaries in the CI Docker image. These variables which are no longer used confused me for a moment to know which version of Kind we are using (0.11.1 now).

https://github.com/elastic/cloud-on-k8s/blob/c47cb6fbcaa0143c9def253f2f161f232a6f21bb/hack/deployer/config/plans.yml#L162-L165


pebrc (approved):
Thanks for cleaning this up!

Issue #5429, opened 2022-03-03
Fleet-managed elastic agent on ECK documentation quick start YAML RBAC configuration problem


jeniawhite:
## Bug Report
**What did you do?**
I've attempted to run this quick start:
https://github.com/elastic/cloud-on-k8s/blob/main/docs/orchestrating-elastic-stack-applications/agent-fleet.asciidoc#quickstart

**What did you expect to see?**
I've expected to get a working deployment (kibana, elastic search, fleet sever, elastic agent)

**What did you see instead? Under which circumstances?**
I've drilled down into the logs of fleet server and saw that it doesn't have the necessary permissions to list namespaces
He got into an infinite loop trying to list the namespaces and failing (look at error log snippet below in the logs section)

**Environment**

* ECK version:

	8.0.1

* Kubernetes information:
        On premise kind cluster v0.11.1 go1.16.4 darwin/amd64

```
$ kubectl version 
Client Version: version.Info{Major:"1", Minor:"22", GitVersion:"v1.22.5", GitCommit:"5c99e2ac2ff9a3c549d9ca665e7bc05a3e18f07e", GitTreeState:"clean", BuildDate:"2021-12-16T08:38:33Z", GoVersion:"go1.16.12", Compiler:"gc", Platform:"darwin/arm64"}
Server Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.1", GitCommit:"5e58841cce77d4bc13713ad2b91fa0d961e69192", GitTreeState:"clean", BuildDate:"2021-05-21T23:06:30Z", GoVersion:"go1.16.4", Compiler:"gc", Platform:"linux/arm64"}
```

* Resource definition:
```
if relevant insert the resource definition
```

* Logs:
```
E0303 15:23:25.809252     936 reflector.go:138] pkg/mod/k8s.io/client-go@v0.21.1/tools/cache/reflector.go:167: Failed to watch *v1.Namespace: failed to list *v1.Namespace: namespaces is forb
idden: User "system:serviceaccount:default:elastic-agent" cannot list resource "namespaces" in API group "" at the cluster scope
E0303 15:23:27.274403     936 reflector.go:138] pkg/mod/k8s.io/client-go@v0.21.1/tools/cache/reflector.go:167: Failed to watch *v1.Namespace: failed to list *v1.Namespace: namespaces is forb
idden: User "system:serviceaccount:default:elastic-agent" cannot list resource "namespaces" in API group "" at the cluster scope
[...]

PR #5430, opened 2022-03-03
Adding necessary ClusterRole permissions for elastic-agent in fleet server quick start 

jeniawhite:
In the quick start guide of fleet-managed elastic agent on ECK, there is a YAML that deploys the environment.
The RBAC of elastic-agent inside that YAML doesn't include the namespaces resource in the core API group.
This causes the fleet agent to log failures and not spin up correctly since it cannot list the namespaces.
I've added the required permissions for the fleet agent to continue it's work
This PR resolves this issue: https://github.com/elastic/cloud-on-k8s/issues/5429


pebrc (approved):
LGTM!

PR #5435, opened 2022-03-04
moved monitoring section to right place

eedugon:
on Override beats pod template the monitoring was at the wrong place. I think monitoring has to be at spec level and not at nodeSets level.


thbkrkr (approved):
:bow:

PR #5437, opened 2022-03-04
Fix webhook generator to include matchPolicy of 'Exact'

### LINKED ISSUE #5423: OLM generator code does not set matchPolicy in webhook definitions
Introduced in https://github.com/elastic/cloud-on-k8s/pull/5043 

The webhook definitions in the cluster service version CR for Operator Lifecycle Manager omit to specify `matchPolicy`.  that means that the default `matchPolicy` comes into effect which is `Equivalent`. That however has the undesired side effect that the `v1beta1` webhook will try to validate `v1` resources and vice versa. 

As validations are not equivalent between the two API versions this can block users from using features like dynamic volume expansion that are not supported on `v1beta1` Elasticsearch CRs for example. 

We have to always use `matchPolicy: Exact` as long as we have multiple API versions. 

cc @naemono

naemono:
Fixes #5423 

This change includes `matchPolicy: Exact` in the OLM webhook generator such that v1beta1 webhooks will not try to validate v1 resources.

Test generated webhook

```
  webhookdefinitions:
    - admissionReviewVersions:
      - v1beta1
      containerPort: 443
      deploymentName: elastic-operator
      failurePolicy: Ignore
      generateName: elastic-agent-validation-v1alpha1.k8s.elastic.co
      matchPolicy: Exact
      rules:
      - apiGroups:
        - agent.k8s.elastic.co
        apiVersions:
        - v1alpha1
        operations:
        - CREATE
        - UPDATE
        resources:
        - agents
      sideEffects: None
      targetPort: 9443
      type: ValidatingAdmissionWebhook
      webhookPath: /validate-agent-k8s-elastic-co-v1alpha1-agent
```

### DIFF
--- a/hack/operatorhub/main.go
+++ b/hack/operatorhub/main.go
@@ -242,6 +242,7 @@ type WebhookDefinition struct {
 	ContainerPort           int                              `json:"containerPort"`
 	DeploymentName          string                           `json:"deploymentName"`
 	FailurePolicy           *admissionv1.FailurePolicyType   `json:"failurePolicy"`
+	MatchPolicy             admissionv1.MatchPolicyType      `json:"matchPolicy"`
 	GenerateName            string                           `json:"generateName"`
 	Rules                   []admissionv1.RuleWithOperations `json:"rules"`
 	SideEffects             *admissionv1.SideEffectClass     `json:"sideEffects"`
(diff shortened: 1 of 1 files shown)


pebrc (approved):
LGTM!

PR #5438, opened 2022-03-04
Fix Agent trust CA commands for all image variants

### LINKED ISSUE #5434: Elastic Agent/Fleet container certificate update does not work for UBI images
from website instructions
https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-elastic-agent-fleet-quickstart.html

```
---
kind: Agent
metadata:
  name: fleet-server-quickstart
  namespace: elastic
spec:
  version: 8.0.1
  kibanaRef:
    name: kibana-quickstart
  elasticsearchRefs:
  - name: elasticsearch-quickstart
  mode: fleet
  fleetServerEnabled: true
  deployment:
    replicas: 1
    podTemplate:
      spec:
        serviceAccountName: elastic-agent
        securityContext:
          runAsUser: 0
---
kind: Agent
metadata:
  name: elastic-agent-quickstart
  namespace: elastic
spec:
  version: 8.0.1
  kibanaRef:
    name: kibana-quickstart
  fleetServerRef:
    name: fleet-server-quickstart
  mode: fleet
  daemonSet:
    podTemplate:
      spec:
        serviceAccountName: elastic-agent
        securityContext:
          runAsUser: 0
---
kind: Kibana
metadata:
  name: kibana-quickstart
  namespace: elastic
spec:
  version: 8.0.1
  count: 1
  elasticsearchRef:
    name: elasticsearch-quickstart
  config:
    xpack.fleet.agents.elasticsearch.hosts: ["https://elasticsearch-quickstart-es-http.elastic.svc:9200"]
    xpack.fleet.agents.fleet_server.hosts: ["https://fleet-server-quickstart-agent-http.elastic.svc:8220"]
---
kind: Elasticsearch
metadata:
  name: elasticsearch-quickstart
  namespace: elastic
spec:
  version: 8.0.1
  nodeSets:
  - name: elastic
    count: 1
    config:
      node.store.allow_mmap: false
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: elastic-agent
rules:
- apiGroups: [""] # "" indicates the core API group
  resources:
  - pods
  - nodes
  verbs:
  - get
  - watch
  - list
- apiGroups: ["coordination.k8s.io"]
  resources:
  - leases
  verbs:
  - get
  - create
  - update
---
apiVersion: v1
kind: ServiceAccount
metadata:
  name: elastic-agent
  namespace: elastic
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: elastic-agent
subjects:
- kind: ServiceAccount
  name: elastic-agent
  namespace: elastic
roleRef:
  kind: ClusterRole
  name: elastic-agent
  apiGroup: rbac.authorization.k8s.io
```
then 
`oc adm policy add-scc-to-user anyuid -z apm-server -n elastic`

but the fleet-server and elastic-agent give errors
`bash: line 4: update-ca-certificates: command not found
`
ps: OpenShift version 4.9.10
Elasticsearch (ECK) Operator 2.0.0 provided by Elastic

naemono:
Fixes #5434

On OCP when using the certified Red Hat operator version of ECK we are using UBI based images by default, which breaks Agent >= version 7.17.0 startup, since we assume we are running within Ubuntu image, not UBI based images.  This fixes that assumption by adjusting the init container's CA trust script to work in both redhat, and debian-based images.

### DIFF
--- a/pkg/controller/agent/pod.go
+++ b/pkg/controller/agent/pod.go
@@ -76,6 +76,11 @@ const (
 	FleetServerElasticsearchUsername = "FLEET_SERVER_ELASTICSEARCH_USERNAME"
 	FleetServerElasticsearchPassword = "FLEET_SERVER_ELASTICSEARCH_PASSWORD" //nolint:gosec
 	FleetServerElasticsearchCA       = "FLEET_SERVER_ELASTICSEARCH_CA"
+
+	ubiSharedCAPath    = "/etc/pki/ca-trust/source/anchors/"
+	ubiUpdateCmd       = "/usr/bin/update-ca-trust"
+	debianSharedCAPath = "/usr/local/share/ca-certificates/"
+	debianUpdateCmd    = "/usr/sbin/update-ca-certificates"
 )
 
 var (
@@ -382,11 +387,6 @@ func getAssociatedFleetServer(params Params) (commonv1.Associated, error) {
 }
 
 func trustCAScript(caPath string) string {
-	redhatSharedCAPath := "/etc/pki/ca-trust/source/anchors/"
-	redhatUpdateCmd := "/usr/bin/update-ca-trust"
-	debianSharedCAPath := "/usr/local/share/ca-certificates/"
-	debianUpdateCmd := "/usr/sbin/update-ca-certificates"
-
 	return fmt.Sprintf(`#!/usr/bin/env bash
 set -e
 if [[ -f %[1]s ]]; then
@@ -399,7 +399,7 @@ if [[ -f %[1]s ]]; then
   fi
 fi
 /usr/bin/tini -- /usr/local/bin/docker-entrypoint -e
-`, caPath, redhatSharedCAPath, redhatUpdateCmd, debianSharedCAPath, debianUpdateCmd)
+`, caPath, ubiSharedCAPath, ubiUpdateCmd, debianSharedCAPath, debianUpdateCmd)
 }
 
 func createDataVolume(params Params) volume.VolumeLike {
--- a/pkg/controller/agent/pod.go
+++ b/pkg/controller/agent/pod.go
@@ -28,7 +28,6 @@ import (
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common/defaults"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common/reconciler"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common/tracing"
-	"github.com/elastic/cloud-on-k8s/pkg/controller/common/version"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common/volume"
 	"github.com/elastic/cloud-on-k8s/pkg/utils/k8s"
 	"github.com/elastic/cloud-on-k8s/pkg/utils/maps"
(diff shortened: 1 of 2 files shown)


pebrc (changes requested):
This is not specific to OpenShift and should not be tied to the OpenShift detection. This is specific to UBI images being in use. Users can use UBI images on other K8s flavours as well. They just happen to be the default in the Red Hat certified version of the ECK operator. 

I am thinking we should maybe check the image name in use and if it has the `-ubi8` suffix use the old version of the certificate trust logic. 

There is also the `-ubi-only` operator flag, which is what is used internally to default to UBI images. But I would also not want to rely on the presence of this flag as users might just override the image tag manually to a UBI version. 

While looking at the image suffix is far from perfect and there are still loop holes in that logic (e.g. custom images that do not use the `-ubi8` suffix) it is probably the most generic. 

There is also the problem of Ironbank which falls into that loophole category as all images there are UBI but I think they do not follow this naming convention.

naemono:
thanks @pebrc, I will investigate and develop a more focused fix based on UBI and/or Ironbank.

naemono:
@pebrc I essentially did what you are asking, but I looked for the binary instead of the existence of a directory.  Let me know what you think.  I tested in `docker.elastic.co/beats/elastic-agent:7.17.0` , `docker.elastic.co/beats/elastic-agent-ubi8:7.17.0` , and all the way back to `docker.elastic.co/beats/elastic-agent-ubi:7.10.0` and this works without fail.

pkg/controller/agent/pod.go
```diff
@@ -388,21 +388,24 @@ func getAssociatedFleetServer(params Params) (commonv1.Associated, error) {
 }
 
 func trustCAScript(ver version.Version, caPath string) string {
-	sharedCAPath := "/etc/pki/ca-trust/source/anchors/"
-	updateCmd := "update-ca-trust"
-	if ver.GTE(version.MinFor(7, 17, 0)) {
-		sharedCAPath = "/usr/local/share/ca-certificates"
-		updateCmd = "update-ca-certificates"
-	}
+	redhatSharedCAPath := "/etc/pki/ca-trust/source/anchors/"
```
pebrc, line 391: `ver` argument is unused and should be removed.

pkg/controller/agent/pod_test.go
```diff
@@ -720,7 +714,7 @@ fi
 			wantPodSpec: generatePodSpec(func(ps corev1.PodSpec) corev1.PodSpec {
 				ps.Volumes = expectedCAVolume
 				ps.Containers[0].VolumeMounts = expectedCAVolumeMount
-				ps.Containers[0].Command = expectedUbuntuCmd
+				ps.Containers[0].Command = expectedCmd
```
pebrc, line 717: There is some duplication in the unit tests now. If you go through the list you will find the same test multiple times once for 7.x and 8.x. The only difference was the different CA script. With a unified script we should remove the duplication in the test cases as those are otherwise just a maintenance burden.

pebrc (commented):
Almost LGTM! Tested and this works indeed across versions so I think that's an improvement 👍

pkg/controller/agent/pod_test.go
naemono: The duplicative tests now have been removed.

pkg/controller/agent/pod.go
naemono: This was caught in the ci tests, and I removed it earlier.

pebrc (approved):
LGTM 👍

pkg/controller/agent/pod.go
```diff
@@ -387,22 +381,25 @@ func getAssociatedFleetServer(params Params) (commonv1.Associated, error) {
...
 
-func trustCAScript(ver version.Version, caPath string) string {
-	sharedCAPath := "/etc/pki/ca-trust/source/anchors/"
-	updateCmd := "update-ca-trust"
-	if ver.GTE(version.MinFor(7, 17, 0)) {
-		sharedCAPath = "/usr/local/share/ca-certificates"
-		updateCmd = "update-ca-certificates"
-	}
+func trustCAScript(caPath string) string {
+	redhatSharedCAPath := "/etc/pki/ca-trust/source/anchors/"
```
barkbay, line 385: I would use the distro name for consistency:
```suggestion
	ubiSharedCAPath := "/etc/pki/ca-trust/source/anchors/"
```

Also those seem to be constants, so maybe:

```
const (
	ubiSharedCAPath    = "/etc/pki/ca-trust/source/anchors/"
	ubiUpdateCmd       = "/usr/bin/update-ca-trust"
	debianSharedCAPath = "/usr/local/share/ca-certificates/"
	debianUpdateCmd    = "/usr/sbin/update-ca-certificates"
)
```

pkg/controller/agent/pod.go
```diff
@@ -387,22 +381,25 @@ func getAssociatedFleetServer(params Params) (commonv1.Associated, error) {
...
+	redhatUpdateCmd := "/usr/bin/update-ca-trust"
```
barkbay, line 386:
```suggestion
	ubiUpdateCmd := "/usr/bin/update-ca-trust"
```

Issue #5306, opened 2022-01-31
Remediate accessibility issues in ECK docs


alaudazzi:
To make the ECK docs more accessible, the following a11y issues in the text have to be fixed:

- Provide an ALT text to  images
- Replace occurrences of
    - [x] see = #5315
    - [x] read more = #5317
    - [x] above = #5672
    - [x] below = #5714
    - [x] right/left = no relevant occurrences found in the docs
    - [x] click = no relevant occurrences found in the docs
    - [x] latinisms (eg., e.g., i.e., etc., via) = #5456

For reference: [How to create accessible web content](https://docs.google.com/document/d/1IGuMf16-cwxx8BX1Z9PWsFWsTPxicwTkX_r7UOhjd4E/edit#heading=h.iykpo698d18b)

PR #5456, opened 2022-03-10
[a11y] Fix latinisms occurrences

alaudazzi:
This PR replaces latinisms with more accessible content:

- eg., e.g.  =>  for ex., for example
- i.e.  =>  that is
- etc.  =>  and so on
- via  =>  through, with

Rel: #5306
The generated content in `../reference/api-docs.asciidoc` is not in scope.

For reference: [How to create accessible web content](https://docs.google.com/document/d/1IGuMf16-cwxx8BX1Z9PWsFWsTPxicwTkX_r7UOhjd4E/edit#heading=h.iykpo698d18b)

### DIFF
--- a/config/crds/v1/all-crds.yaml
+++ b/config/crds/v1/all-crds.yaml
@@ -1365,7 +1365,7 @@ spec:
                 type: array
               serviceAccountName:
                 description: ServiceAccountName is used to check access from the current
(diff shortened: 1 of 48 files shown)


thbkrkr (approved):
LGTM

PR #5459, opened 2022-03-10
Avoid unnecessary update calls when reconciling PVC owner refs

### LINKED ISSUE #5451: Avoid unnecessary PVCs updates when updating ownerRefs

in https://github.com/elastic/cloud-on-k8s/blob/9be005d2a4b6e26f98ff9c520ea53ae4d6cc9455/pkg/controller/elasticsearch/driver/pvc_owner.go#L34-L47

we do not check whether a update of the PVC is actually necessary. We should use a similar approach here as in other cases whereby we compare the current state with the desired state and only issue the update if they differ. 

This should significantly reduce the number updates on PVCs the operator does.

pebrc:
Fixes #5451


thbkrkr (approved):
LGTM

Issue #5450, opened 2022-03-09
Investigate why we do so many apiserver requests to reconcile a cluster


sebgl:
**We do ~900 apiserver requests to create a hot/warm/master 3xAZ (9 ssets) ES cluster + Kibana.**

I've been trying to grab numbers about how many requests we do to the apiserver to reconcile a "standard" cluster.
The approach I took was to patch the client go to log all requests being made:

```go
# main.go
type loggingRoundTripper struct {
	initialrt http.RoundTripper
}

var _ http.RoundTripper = loggingRoundTripper{}

func (l loggingRoundTripper) RoundTrip(r *http.Request) (*http.Response, error) {
	resp, err := l.initialrt.RoundTrip(r)
	fmt.Println("req", time.Now().Format("15:04:05"), r.Method, r.URL.RequestURI(), resp.StatusCode)
	return resp, err
}

(...)

# before creating the manager
cfg.WrapTransport = func(rt http.RoundTripper) http.RoundTripper {
		return loggingRoundTripper{initialrt: rt}
	}
```

With a manifest that looks like this:

```
# This sample sets up an Elasticsearch cluster and a Kibana instance preconfigured for that cluster
apiVersion: elasticsearch.k8s.elastic.co/v1
kind: Elasticsearch
metadata:
  name: elasticsearch-sample
spec:
  version: 8.0.0
  nodeSets:
  - name: hot-zone0
    count: 1
    config:
      node.store.allow_mmap: false
  - name: hot-zone1
    count: 1
    config:
      node.store.allow_mmap: false
  - name: hot-zone2
    count: 1
    config:
      node.store.allow_mmap: false
  - name: warm-zone0
    count: 1
    config:
      node.store.allow_mmap: false
  - name: warm-zone1
    count: 1
    config:
      node.store.allow_mmap: false
  - name: warm-zone2
    count: 1
    config:
      node.store.allow_mmap: false
  - name: master-zone1
    count: 1
    config:
      node.store.allow_mmap: false
  - name: master-zone2
    count: 1
    config:
      node.store.allow_mmap: false
  - name: master-zone3
    count: 1
    config:
      node.store.allow_mmap: false
---
apiVersion: kibana.k8s.elastic.co/v1
kind: Kibana
metadata:
  name: kibana-sample
spec:
  version: 8.0.0
  count: 2
  elasticsearchRef:
    name: "elasticsearch-sample"
```

[...]

PR #5461, opened 2022-03-10
Avoid unnecessary DELETE calls to manage legacy transport secret

pebrc:
Related to #5450 

This PR proposes to do a GET (from cache) before DELETE to minimise impact on the API server. Alternative approach I could think of is to track the deletion in the orchestration hints annotation but that seemed costly/complicated compared to the almost free GET request.

### DIFF
--- a/pkg/controller/elasticsearch/certificates/transport/reconcile.go
+++ b/pkg/controller/elasticsearch/certificates/transport/reconcile.go
@@ -13,7 +13,9 @@ import (
 
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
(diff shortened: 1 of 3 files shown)


pkg/controller/elasticsearch/certificates/transport/reconcile.go
```diff
@@ -74,11 +76,14 @@ func DeleteStatefulSetTransportCertificate(client k8s.Client, namespace string,
...
 // DeleteLegacyTransportCertificate ensures that the former Secret which used to contain the transport certificates is deleted.
 func DeleteLegacyTransportCertificate(client k8s.Client, es esv1.Elasticsearch) error {
-	secret := corev1.Secret{
-		ObjectMeta: metav1.ObjectMeta{
-			Namespace: es.Namespace,
-			Name:      esv1.LegacyTransportCertsSecretSuffix(es.Name),
-		},
+	var secret corev1.Secret
+	// do a GET from the cache first before attempting a DElETE that hits the API server
+	err := client.Get(context.Background(), types.NamespacedName{Namespace: es.Namespace, Name: esv1.LegacyTransportCertsSecretSuffix(es.Name)}, &secret)
```
naemono, line 81: Question, not an issue, but why do we use `context.Background()` throughout the codebase for kubernetes client.* calls instead of something with a default timeout?

naemono (approved):
lgtm; question though

pkg/controller/elasticsearch/certificates/transport/reconcile.go
```diff
@@ -74,11 +76,14 @@ func DeleteStatefulSetTransportCertificate(client k8s.Client, namespace string,
 
 // DeleteLegacyTransportCertificate ensures that the former Secret which used to contain the transport certificates is deleted.
 func DeleteLegacyTransportCertificate(client k8s.Client, es esv1.Elasticsearch) error {
-	secret := corev1.Secret{
-		ObjectMeta: metav1.ObjectMeta{
-			Namespace: es.Namespace,
-			Name:      esv1.LegacyTransportCertsSecretSuffix(es.Name),
-		},
+	var secret corev1.Secret
+	// do a GET from the cache first before attempting a DElETE that hits the API server
```
thbkrkr, line 80:
```suggestion
	// do a GET from the cache first before attempting a DELETE that hits the API server
```

pkg/controller/elasticsearch/certificates/transport/reconcile.go
pebrc: I think that's a valid question maybe raise a separate issue to discuss it though?

pebrc: FWIW https://github.com/elastic/cloud-on-k8s/issues/5450 indicates that a 1m timeout is applied to all requests

PR #5487, opened 2022-03-16
Avoid unnecessary DELETE when managing ES license secret

pebrc:

Removes an unconditional DELETE call when reconciling the Elasticsearch cluster license secret. This only affects installations running with a Basic license or those downgrading from an Enterprise license to basic e.g because the license expired. I added an initial GET to check if there is something to delete in the first place before we delete.


pebrc:
Closing this one as I am going to do a bunch of this one new PR instead to reuse code

PR #5488, opened 2022-03-16
Avoid DELETEs of non-existent secrets

pebrc:

This PR guards DELETE calls on secrets (cluster license, secure settings, legacy transport certs) with a GET call through a new shared utility function to remove unnecessary load on the API server. 

It also renames `keystore.NewResources`  to `keystore.ReconcileResources`, which I am happy to back out into a separate PR if that's people are uncomfortable. The reason for this rename was that while I was tracking down these DELETES, I was confused and surprised to find reconciliation activity happening in a function that appeared to be a simple constructor with `New*` in the name.

### DIFF
--- a/pkg/controller/common/certificates/reconcile.go
+++ b/pkg/controller/common/certificates/reconcile.go
@@ -114,19 +114,19 @@ func (r Reconciler) ReconcileCAAndHTTPCerts(ctx context.Context) (*CertificatesS
 func (r *Reconciler) removeCAAndHTTPCertsSecrets() error {
 	owner := k8s.ExtractNamespacedName(r.Owner)
 	// remove public certs secret
-	if err := deleteIfExists(r.K8sClient,
+	if err := k8s.DeleteSecretIfExists(r.K8sClient,
 		types.NamespacedName{Namespace: owner.Namespace, Name: PublicCertsSecretName(r.Namer, owner.Name)},
 	); err != nil {
 		return err
 	}
 	// remove internal certs secret
-	if err := deleteIfExists(r.K8sClient,
+	if err := k8s.DeleteSecretIfExists(r.K8sClient,
 		types.NamespacedName{Namespace: owner.Namespace, Name: InternalCertsSecretName(r.Namer, owner.Name)},
 	); err != nil {
 		return err
 	}
 	// remove CA secret
-	if err := deleteIfExists(r.K8sClient,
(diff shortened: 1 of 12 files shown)


pkg/controller/common/certificates/reconcile.go
```diff
@@ -140,17 +139,7 @@ func (r *Reconciler) removeCAAndHTTPCertsSecrets() error {
...
-		return err
-	}
-	log.Info("Deleting secret", "namespace", secretRef.Namespace, "secret_name", secretRef.Name)
-	err = c.Delete(context.Background(), &secret)
-	if err != nil && apierrors.IsNotFound(err) {
-		return nil
-	}
-	return err
+	return k8s.DeleteSecretIfExists(c, secretRef, func() {
+		log.Info("Deleting secret", "namespace", secretRef.Namespace, "secret_name", secretRef.Name)
```
thbkrkr, line 143: I'm a little perplexed about the usefulness of this log. If we do without it, we can delete `deleteIfExists` and use `k8s.DeleteSecretIfExists` instead.

pkg/controller/license/license_controller.go
```diff
@@ -198,19 +198,12 @@ func (r *ReconcileLicenses) reconcileClusterLicense(cluster esv1.Elasticsearch)
...
-		secretName := esv1.LicenseSecretName(cluster.Name)
-		err := r.Client.Delete(context.Background(), &corev1.Secret{
-			ObjectMeta: k8s.ToObjectMeta(types.NamespacedName{
-				Namespace: cluster.Namespace,
-				Name:      secretName,
-			}),
+		// no matching license found, delete cluster level license if it exists to revert to basic
+		clusterLicenseNSN := types.NamespacedName{Namespace: cluster.Namespace, Name: esv1.LicenseSecretName(cluster.Name)}
+		err := k8s.DeleteSecretIfExists(r.Client, clusterLicenseNSN, func() {
+			log.V(1).Info("No enterprise license found. Attempting to remove cluster license secret", "namespace", cluster.Namespace, "es_name", cluster.Name)
```
thbkrkr, line 204: Is it really important to log this between the `get` and the `delete`. Otherwise, we keep the log before `DeleteSecretIfExists` and we can remove the `onDelete` callback.

pebrc: Sure I am open to that.

thbkrkr (approved):
LGTM

Issue #4627, opened 2021-07-09
Stack monitoring and version upgrade order


sebgl:
We ensure a strict ordering of stack product version upgrades: Elasticsearch first, then Enterprise Search, then Kibana, etc.

This is done through our association mechanism: the associated resource can only be updated once all referenced resources are updated: https://github.com/elastic/cloud-on-k8s/blob/7c39ff040e1e15b3e9b41ceb9779fd9101e938cb/pkg/controller/association/conf.go#L111-L147

This function gets called as part of individual resource controllers. For example in the Kibana reconciliation: https://github.com/elastic/cloud-on-k8s/blob/7c39ff040e1e15b3e9b41ceb9779fd9101e938cb/pkg/controller/kibana/driver.go#L131-L133

We don't do that check in the Elasticsearch controller. Mostly because so far we didn't have any Elasticsearch -> X association. That's not true anymore with [stack monitoring](https://github.com/elastic/cloud-on-k8s/pull/4528).

In the Kibana controller, we do check that any referenced ES cluster for monitoring purposes is running with a higher or equal version.
For consistency, we should do the same in the Elasticsearch controller.

This is where things get a bit tricky :)

* Association to itself: Elasticsearch monitoring data can be configured to be sent to itself. In which case we need to make sure we don't get into a circular race condition where the version cannot be updated at all because the controller waits for its own resource to be updated first (🔁).

* Circular associations: we could configure 2 different clusters to monitor each other (ES A <-> ES B). In which case both would require the other one to be updated first.

(The current situation where we don't do that check in the Elasticsearch controller is maybe an OK solution to that problem, overall).

thbkrkr:
Currently, the circular associations (ES A <-> ES B) works and the association to itelf does not.

thbkrkr:
Note on version compatibility of monitored and monitoring clusters: Ideally the monitoring cluster and the production cluster run on the same Elastic Stack version. However, a monitoring cluster on the latest release of 7.x also works with production clusters that **use the same major version**. Monitoring clusters that use 7.x also work with production clusters that use the latest release of 6.x (source: https://www.elastic.co/guide/en/elasticsearch/reference/current/monitoring-production.html#monitoring-production).

thbkrkr:
Closed by:
- #5339
- #5489

PR #5489, opened 2022-03-16
Make sure to read association configuration again from annotations if it was reset

### LINKED ISSUE #4709: Cannot use Stack Monitoring to have Elasticsearch monitor itself
Configuring an Elasticsearch cluster to monitor itself is not possible due to a circular dependency issue (#4627).

The Elasticsearch monitoring cluster referenced in `monitoring.metrics` and `monitoring.logs` must be a separate cluster.

This is currently documented as a limitation:

https://github.com/elastic/cloud-on-k8s/blob/a6efd826e6ccbcf03658a007a510934923d5ceae/docs/advanced-topics/stack-monitoring.asciidoc#L65

thbkrkr:
This commit introduces two new functions `GetAndSetAssociationConf()`and `GetAndSetAssociationConfByRef()` to be called in the implementation of `AssociationConf()` for each `Association` to be sure that if there is an update of the associated resource that carries the association that resets the assocConf, then we make sure to read it back from the annotation.

This should be called very rarely because we re-read the annotation only if the current assoc conf is not set.

Due to a circular dependency, I move the functions in the `pkg/apis/common/v1` package.

Note: I couldn't reproduce the issue that resets non persisted attributes during an update with the fakeClient.

Resolves #4709.


pkg/apis/elasticsearch/v1/elasticsearch_types_test.go
```diff
@@ -314,3 +316,76 @@ func TestElasticsearch_DisabledPredicates(t *testing.T) {
...
+	// set assoc conf
+	assert.Equal(t, 0, len(esMon.AssocConfs))
+	for _, association := range esMon.GetAssociations() {
+		assocConf, err := assocutils.GetAssociationConf(association)
+		assert.NoError(t, err)
+		association.SetAssociationConf(assocConf)
+	}
+	assert.Equal(t, 2, len(esMon.AssocConfs))
+
+	// simulate the case where the assocConfs map is reset, which can happen if the resource is updated
```
barkbay, line 382: > simulate the case where the assocConfs map is reset, which can happen if the resource is updated

Sorry, I'm not sure to understand. `assocConfs` is not persisted, I think it is expected for it to be empty before [`FetchWithAssociations`](https://github.com/elastic/cloud-on-k8s/blob/721cccfd905f43c345c7957fbfe173ea3ebb6bbf/pkg/controller/association/conf.go#L50) kicks in ?

thbkrkr: The issue happens after [FetchWithAssociations](https://github.com/elastic/cloud-on-k8s/blob/721cccfd905f43c345c7957fbfe173ea3ebb6bbf/pkg/controller/association/conf.go#L50), when the resource is updated, it resets the map. This test shows that if it is reset after `SetAssociationConf()`, the map is repopulated on the fly by `AssociationConf()`. 
Maybe I should go further and delete `SetAssociationConf()` to avoid confusion. We could even delete the `FetchWithAssociations()` and just rely on `AssociationConf()` called in different places.

barkbay: > We could even delete the FetchWithAssociations() and just rely on AssociationConf() called in different places.

Thanks for the clarification ! My understanding is that it means that `FetchWithAssociations` is unreliable 😕 I wonder if it's really worth keeping it along with the unserialized annotations ? IIUC they are only here to spare a few CPU cycles by not using `json.Unmarshal` when the association config. is needed ?

barkbay (commented):
> Due to a circular dependency, I created a new utils package under association 

Could we move these new functions in `pkg/apis/common/v1` ? It would avoid the addition of a new package and solve the circular dependency.

pkg/apis/elasticsearch/v1/elasticsearch_types_test.go
```diff
@@ -314,3 +316,88 @@ func TestElasticsearch_DisabledPredicates(t *testing.T) {
...
+	for _, assoc := range esMon.GetAssociations() {
+		assert.NotNil(t, assoc.AssociationConf())
+	}
+	assert.Equal(t, 2, len(esMon.AssocConfs))
+
+	// delete just one entry in the map
+	delete(esMon.AssocConfs, metricsEsRef.NamespacedName())
+	assert.Equal(t, 1	, len(esMon.AssocConfs))
+
+	// checks that the map the missing entry is set again
```
barkbay, line 397:
```suggestion
	// checks that the missing entry is set again
```

pkg/apis/elasticsearch/v1/elasticsearch_types_test.go
```diff
@@ -314,3 +316,88 @@ func TestElasticsearch_DisabledPredicates(t *testing.T) {
 		})
 	}
 }
+
+// Test_AssociationConfs tests that if something resets the AssocConfs map, then AssociationConf() reinitializes
+// the map from the annotation.
```
barkbay, line 321: Not sure about `resets`, I would use `clear`, but it might come from my [Java background](https://docs.oracle.com/javase/7/docs/api/java/util/HashMap.html#clear()).
```suggestion
// Test_AssociationConfs tests that if the association configuration map in an associated object is cleared, then AssociationConf() is rebuilt from the annotation.
```

pkg/apis/elasticsearch/v1/elasticsearch_types_test.go
```diff
@@ -314,3 +316,88 @@ func TestElasticsearch_DisabledPredicates(t *testing.T) {
...
+		},
+	}
+
+	// set assoc conf even if no assoc
+	assert.Equal(t, 0, len(es.AssocConfs))
+	for _, association := range es.GetAssociations() {
+		assocConf, err := assocutils.GetAssociationConf(association)
+		assert.NoError(t, err)
+		association.SetAssociationConf(assocConf)
+	}
```
barkbay, line 337: nit: should we just remove this loop and ensure that there is no item returned ?
```suggestion
	assert.Equal(t, 0, len(es.GetAssociations()))
```

pkg/apis/elasticsearch/v1/elasticsearch_types_test.go
```diff
@@ -314,3 +316,88 @@ func TestElasticsearch_DisabledPredicates(t *testing.T) {
...
+	assert.Equal(t, 0, len(es.AssocConfs))
+
+	// checks that assocConfs is nil
+	for _, assoc := range es.GetAssociations() {
```
barkbay, line 341: nit: same as above, my understanding is that there should be no association.

pkg/apis/elasticsearch/v1/elasticsearch_types_test.go
```diff
@@ -314,3 +316,88 @@ func TestElasticsearch_DisabledPredicates(t *testing.T) {
...
+	}
+
+	// set assoc conf
+	assert.Equal(t, 0, len(esMon.AssocConfs))
+	for _, association := range esMon.GetAssociations() {
+		assocConf, err := assocutils.GetAssociationConf(association)
+		assert.NoError(t, err)
+		association.SetAssociationConf(assocConf)
+	}
+	assert.Equal(t, 2, len(esMon.AssocConfs))
```
barkbay, line 381: my understanding is that `AssociationConf()` is expected to also set the association configuration:
```suggestion
	// map should be initially empty
	assert.Equal(t, 0, len(esMon.AssocConfs))
	for _, association := range esMon.GetAssociations() {
		conf := association.AssociationConf()
		assert.NotNil(t, conf)
	}
	// map should have been populated by the call to AssociationConf()
	assert.Equal(t, 2, len(esMon.AssocConfs))
```

pkg/controller/association/utils/utils.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+	return extractAssocConfFromAnnotation(annotations, association.AssociationConfAnnotationName())
+}
+
+// setAssocConfFromAnnotation sets the association configuration extracted from the annotations in the given association.
+func setAssocConfFromAnnotation(assoc commonv1.Association) *commonv1.AssociationConf {
+	assocConf, err := extractAssocConfFromAnnotation(assoc.Associated().GetAnnotations(), assoc.AssociationConfAnnotationName())
+	if err != nil {
+		// ignore this unlikely unexpected error that should not happen as we control the annotation
+		return nil
```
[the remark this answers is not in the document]
thbkrkr: You're absolutely right. It was acceptable when the error was caught in `SetAssociationConf()` called in `FetchWithAssociation()` but now that they are gone it is not acceptable. This involves a refactoring so that `AssociationConf()` returns an `error.`

pkg/controller/apmserver/controller.go
```diff
@@ -241,7 +245,11 @@ func (r *ReconcileApmServer) doReconcile(ctx context.Context, request reconcile.
 		return reconcile.Result{}, err
 	}
 	logger := log.WithValues("namespace", as.Namespace, "as_name", as.Name)
-	if !association.AllowVersion(asVersion, as, logger, r.recorder) {
+	assocAllowed, err := association.AllowVersion(asVersion, as, logger, r.recorder)
+	if err != nil {
+		return reconcile.Result{}, err
```
barkbay, line 250:
```suggestion
		return reconcile.Result{}, tracing.CaptureError(ctx, err)
```

barkbay (approved):
LGTM

barkbay:
As I already mentioned I still have the feeling that these "non-persisted" annotations bring more complexity than value, but overall I think I'm fine keeping them.

thbkrkr:
As this is already in place and is now well isolated and reliable, I am in favor of keeping it.

barkbay:
@thbkrkr : just to double check, does it really resolve https://github.com/elastic/cloud-on-k8s/issues/4709 ? I'm asking because that issue is linked to this other one: https://github.com/elastic/cloud-on-k8s/issues/4627

thbkrkr:
> @thbkrkr : just to double check, does it really resolve #4709 ? I'm asking because that issue is linked to this other one: #4627

Yes it fixes #4709 and so #4627. Version upgrade for ES cluster with self monitoring was fixed by #5339.

PR #5492, opened 2022-03-17
Add known issue for Red Hat certified operator upgrades

pebrc:
This adds a known issue to the release highlights for 1.9.1 and 2.0.0 explaining the workaround for the broken upgrade graph on the Red Hat certified version of the operator

### DIFF
--- a/docs/release-notes/highlights-1.9.1.asciidoc
+++ b/docs/release-notes/highlights-1.9.1.asciidoc
@@ -18,5 +18,6 @@ This release introduces a preemptive measure to mitigate link:https://github.com
 [id="{p}-191-known-issues"]
 === Known issues
 
+- When using the Red Hat certified version of the operator automatic upgrades from previous versions of ECK do not work. To upgrade remove the old ECK operator installation and install the new version manually. Because CRDs remain in place after removal this operation should not negatively affect existing Elastic Stack deployments managed by ECK.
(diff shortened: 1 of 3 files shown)


docs/release-notes/highlights-1.9.1.asciidoc
```diff
@@ -18,5 +18,6 @@ This release introduces a preemptive measure to mitigate link:https://github.com
 [id="{p}-191-known-issues"]
 === Known issues
 
+- When using the Red Hat certified version of the operator automatic upgrades from previous versions of ECK do not work. To upgrade remove the old ECK operator installation and install the new version manually. Because CRDs remain in place after removal this operation should not negatively affect existing Elastic Stack deployments managed by ECK.
```
thbkrkr, line 21:
```suggestion
- When using the Red Hat certified version of the operator, automatic upgrades from previous versions of ECK do not work. To upgrade, delete the old ECK operator installation and install the new version manually. Because CRDs remain in place after deletion, this operation should not negatively affect existing Elastic Stack deployments managed by ECK.
```

thbkrkr: TIL https://apps.labor.ny.gov/ux/doc/v1/design-remove-vs-delete.html. which makes me think that delete is more appropriate that remove here.

pebrc: How about uninstall? I think that is actually what it says in the OCP UI.

thbkrkr (approved):
LGTM

PR #5504, opened 2022-03-22
E2E: change EKS region in CI

pebrc:
Change region where we deploy EKS for e2e test runs to avoid quota issues.


thbkrkr (approved):
は、私にはよく見えますよ

PR #5507, opened 2022-03-23
Reset phase on each reconciliation in Elasticsearch status

### LINKED ISSUE #5506: Elasticsearch Invalid phase can only be cleared through downscale
https://github.com/elastic/cloud-on-k8s/blob/16180dd961b596d15e035feadb2bbaf27a28066c/pkg/controller/elasticsearch/elasticsearch_controller.go#L208-L211

The above gives special treatment to the invalid phase. That means that once it is set it will never be overwritten unless another non-ready phase is set. But `invalid` can be a transient state e.g. a license check request that temporary fails or an error in the `podTemplate` that has been fixed.

https://github.com/elastic/cloud-on-k8s/blob/adb6ea5e434a01bc66f5f36fe2f28d761626a6aa/pkg/controller/elasticsearch/driver/driver.go#L239-L243

Because we preserve other non-ready phases when we update the phase to `ApplyingChanges` the only way to clear this status is to scale down the cluster by at least one node or have a node shutdown stall during a rolling upgrade (which is unlikely)

pebrc:
Fixes #5506 

Instead of carrying the previous phase over from persisted state this PR proposes to reset the phase on each reconciliation to a neutral value (""). Because we set a phase value even in the presence of errors this should ensure that phase values like `Invalid` that we give priority over other values do not stick around forever.  Also moved the priority logic into the state struct method.

### DIFF
--- a/pkg/controller/elasticsearch/elasticsearch_controller.go
+++ b/pkg/controller/elasticsearch/elasticsearch_controller.go
@@ -200,13 +200,9 @@ func (r *ReconcileElasticsearch) Reconcile(ctx context.Context, request reconcil
 	}
 
 	if isReconciled, message := results.IsReconciled(); !isReconciled {
-		// Do not overwrite other "non-ready" phases like MigratingData
-		if state.IsElasticsearchPhase(esv1.ElasticsearchReadyPhase) {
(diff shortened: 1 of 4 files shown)


barkbay (approved):
LGTM !

Thanks for correcting my mistake!
Also the small refactoring makes the code much more cleaner👍

PR #5526, opened 2022-03-28
EnterpriseSearch: Set status.ObservedGeneration from metadata.Generation

naemono:
related to #3392

This change allows status.ObservedGeneration to be kept in sync with metadata.Generation according to the rules described in the above issue. This change adds a set of unit tests which show more details into when observedGeneration will, and will not be updated according to any errors/situations that may occur during reconciliation. End to end tests are in place as well.

Remaining Items
- [x] E2E test


naemono:
run full pr build

PR #5533, opened 2022-03-29
Log info for service not found error when reconciling associations

thbkrkr:
When two associated Elastic resources are created, we log a few `Service not found` error from the association controller because the main controller of the resource haven't had time to create the service yet.
This commit makes error logging smarter by waiting an arbitrary 5 seconds delay after the resource is created to consider it worth logging the `Service not found` error.

Example of a `Service not found` error that is repeatedly logged during creation:
```
2022-03-29T16:39:12.572+0200    ERROR   manager.eck-operator.controller.kb-es-association-controller    Reconciler error
{
    "service.version": "2.2.0-SNAPSHOT+c6de2cfd",
    "name": "self-mon-x",
    "namespace": "default",
    "error": "while fetching referenced service: Service \"self-mon-x-es-http\" not found",
    "errorCauses": [
        {
            "error": "while fetching referenced service: Service \"self-mon-x-es-http\" not found"
        }
    ]
}
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
        /home/thb/dev/pkg/mod/sigs.k8s.io/controller-runtime@v0.11.1/pkg/internal/controller/controller.go:266
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2
        /home/thb/dev/pkg/mod/sigs.k8s.io/controller-runtime@v0.11.1/pkg/internal/controller/controller.go:227
```


pkg/controller/association/reconciler.go
```diff
@@ -263,7 +267,13 @@ func (r *Reconciler) reconcileAssociation(ctx context.Context, association commo
 
 	url, err := r.AssociationInfo.ExternalServiceURL(r.Client, association)
 	if err != nil {
-		return commonv1.AssociationPending, err // maybe not created yet
+		// don't log an error if the service is not found and its parent resource has just been created
+		timeSinceCreation := time.Since(association.Associated().GetCreationTimestamp().Time)
+		maybeNotYetCreated := timeSinceCreation < timeToExpectServiceAfterCreation
+		if apierrors.IsNotFound(err) && maybeNotYetCreated {
+			return commonv1.AssociationPending, nil
+		}
```
barkbay, line 275: If the `Service` does not exist yet, and if we think that it can be confusing for the user to see an error, should we just log this event at the info level, and maybe generate a warn event? If the `Service` cannot be created I would expect the error to be logged by the relevant controller anyway.

```suggestion
		if apierrors.IsNotFound(err) {
			// Service is expected to be created by the resource controller
			k8s.EmitErrorEvent(r.recorder, err, association, events.EventAssociationError,
				"Associated resource Service for %s does not exist yet: %v", assocRef.NamespacedName(), err)
			log.Info("Associated resource Service is not available yet", "error", err, "name", association.Associated().GetName(), "ref_name", assocRef.Name)
			return commonv1.AssociationPending, nil
		}
```

thbkrkr: You're right, the resource controller is doing its job and will report any errors. I'm good at just logging the event at the info level but I'm not in favor of generating a warn event. Otherwise, I feel like that we are moving the noise from the operator log to the k8s event stream.

thbkrkr: You're right, the resource controller is doing its job and will report any errors. I'm good at just logging the event at the info level but I'm not in favor of generating a warn event. Otherwise I feel like we're moving the noise from the operator log to the k8s event stream.

barkbay (approved):
LGTM

PR #5534, opened 2022-03-29
Fix elasticsearch status subresource is updated infinitely

### LINKED ISSUE #5532: Elasticsearch status subresource is updated infinitely
Starting with #5496 the slices of nodes in the status (`DownscaleOperation.Nodes`, `UpgradeOperation.Nodes` and `UpscaleOperation.Node`) are retrieved as `nil` when empty, while the reconcile logic always ensures that this is not the case in the expected status:

```golang
// RecordNewNodes records pending node creations.
func (u *UpscaleReporter) RecordNewNodes(nodes []string) {
	if u == nil {
		return
	}
	if u.nodes == nil {
		u.nodes = make(map[string]esv1.NewNode, len(nodes)) // Here
	}
 ```

This leads `reflect.DeepEqual(previous, current)` to always return `false`: https://github.com/elastic/cloud-on-k8s/blob/371c3248c7fecab952cf04b614229eae0da80134/pkg/controller/elasticsearch/reconcile/state.go#L179-L181
The status is therefore updated infinitely.

### PR DESCRIPTION
**Pull request by thbkrkr**
This commit makes sure that we don't convert a nil slice of nodes into an empty slice in the previous status during the merge with the current status. 

Resolves #5532.

### DIFF
--- a/pkg/controller/elasticsearch/reconcile/status.go
+++ b/pkg/controller/elasticsearch/reconcile/status.go
@@ -105,22 +105,25 @@ func (u *UpscaleReporter) HasPendingNewNodes() bool {
 // Merge creates a new upscale status using the reported upscale status and an existing upscale status.
 func (u *UpscaleReporter) Merge(other esv1.UpscaleOperation) esv1.UpscaleOperation {
 	upscaleOperation := other.DeepCopy()
-	if u == nil || (len(u.nodes)+len(other.Nodes) == 0) {
+	if u == nil {
 		return *upscaleOperation
 	}
-	nodes := make([]esv1.NewNode, 0, len(u.nodes))
-	for name, node := range u.nodes {
-		nodes = append(nodes, esv1.NewNode{
-			Name:    name,
-			Status:  node.Status,
-			Message: node.Message,
+	var nodes []esv1.NewNode
+	if len(u.nodes) != 0 {
+		nodes = make([]esv1.NewNode, 0, len(u.nodes))
+		for name, node := range u.nodes {
+			nodes = append(nodes, esv1.NewNode{
+				Name:    name,
+				Status:  node.Status,
+				Message: node.Message,
+			})
+		}
+		// Sort for stable comparison
+		sort.Slice(nodes, func(i, j int) bool {
+			return nodes[i].Name < nodes[j].Name
 		})
 	}
-	// Sort for stable comparison
-	sort.Slice(nodes, func(i, j int) bool {
-		return nodes[i].Name < nodes[j].Name
-	})
-	if (u.nodes != nil && !reflect.DeepEqual(nodes, other.Nodes)) || upscaleOperation.Nodes == nil {
+	if !reflect.DeepEqual(nodes, other.Nodes) || upscaleOperation.LastUpdatedTime.IsZero() {
 		upscaleOperation.Nodes = nodes
 		upscaleOperation.LastUpdatedTime = metav1.Now()
 	}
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: pkg/controller/elasticsearch/reconcile/status.go [hunk 1]
```diff
@@ -105,7 +105,7 @@ func (u *UpscaleReporter) HasPendingNewNodes() bool {
 // Merge creates a new upscale status using the reported upscale status and an existing upscale status.
 func (u *UpscaleReporter) Merge(other esv1.UpscaleOperation) esv1.UpscaleOperation {
 	upscaleOperation := other.DeepCopy()
-	if u == nil {
+	if u == nil || (len(u.nodes)+len(other.Nodes) == 0) {
```
barkbay, line 108: I think one of the original idea was to always update `LastUpdatedTime`, which is no more the case here. Also having an empty slice of nodes instead of `nil` was a clear message to the user that the status has been effectively processed, making it explicit that there is no nodes involved in an operation. For example after an upgrade:

* Original status:
```json
  "inProgressOperations": {
    "downscale": {
      "lastUpdatedTime": "2022-03-30T05:33:33Z",
      "nodes": []
    },
    "upgrade": {
      "lastUpdatedTime": "2022-03-30T05:37:51Z",
      "nodes": []
    },
    "upscale": {
      "lastUpdatedTime": "2022-03-30T05:33:33Z",
      "nodes": []
    }
  }
```

* With this PR:
```json
  "inProgressOperations": {
    "downscale": {},
    "upgrade": {
      "lastUpdatedTime": "2022-03-30T05:27:48Z"
    },
    "upscale": {}
  }
```

I think I'm fine with the new approach though (but I still prefer the original one tbh). In that case I think we can simplify the following code since the lengths of of the slices have already been tested:

```go
	if (d.nodes != nil && !reflect.DeepEqual(nodes, other.Nodes)) || downscaleOperation.Nodes == nil {
		downscaleOperation.Nodes = nodes
		downscaleOperation.LastUpdatedTime = metav1.Now()
	}
```

pkg/controller/elasticsearch/reconcile/status.go
```diff
@@ -240,7 +240,7 @@ func (d *DownscaleReporter) RecordNodesToBeRemoved(nodes []string) {
 // Merge creates a new downscale status using the reported downscale status and an existing downscale status.
 func (d *DownscaleReporter) Merge(other esv1.DownscaleOperation) esv1.DownscaleOperation {
 	downscaleOperation := other.DeepCopy()
-	if d == nil {
+	if d == nil || (len(d.nodes)+len(other.Nodes) == 0) {
```
barkbay, line 243: I'm wondering if there are some cases where the `LastUpdatedTime` might not been updated if the value of `stalled` is updated.

pkg/controller/elasticsearch/reconcile/status.go
thbkrkr: I too prefer the original approach. Here is another version to avoid converting the nil slices into empty slices and keeping the logic behind `LastUpdatedTime`: b99f0f9e5.

pkg/controller/elasticsearch/reconcile/status.go
```diff
@@ -191,20 +194,23 @@ func (u *UpgradeReporter) Merge(other esv1.UpgradeOperation) esv1.UpgradeOperati
...
+		sort.Slice(nodes, func(i, j int) bool {
+			return nodes[i].Name < nodes[j].Name
 		})
 	}
-	// Sort for stable comparison
-	sort.Slice(nodes, func(i, j int) bool {
-		return nodes[i].Name < nodes[j].Name
-	})
-	if (u.nodes != nil && !reflect.DeepEqual(nodes, other.Nodes)) || upgradeOperation.Nodes == nil {
+	if !reflect.DeepEqual(nodes, other.Nodes) || upgradeOperation.LastUpdatedTime.IsZero() {
```
barkbay, line 213: I think we should also test if `u.nodes` is not `nil` There are some cases where the upgrade logic may not been invoked. In that cases we don't want to clear the current status, because the reporter has not been used at all. This is why the [current condition](https://github.com/elastic/cloud-on-k8s/blob/84b13c74d7a367122190b78feaa650ffa21424ed/pkg/controller/elasticsearch/reconcile/status.go#L207) is `u.nodes != nil && !reflect.DeepEqual(nodes, other.Nodes)`

thbkrkr: Makes sense.

pkg/controller/elasticsearch/reconcile/status.go
```diff
@@ -191,20 +194,23 @@ func (u *UpgradeReporter) Merge(other esv1.UpgradeOperation) esv1.UpgradeOperati
...
+		sort.Slice(nodes, func(i, j int) bool {
+			return nodes[i].Name < nodes[j].Name
 		})
 	}
-	// Sort for stable comparison
-	sort.Slice(nodes, func(i, j int) bool {
-		return nodes[i].Name < nodes[j].Name
-	})
-	if (u.nodes != nil && !reflect.DeepEqual(nodes, other.Nodes)) || upgradeOperation.Nodes == nil {
+	if (u.nodes != nil && !reflect.DeepEqual(nodes, other.Nodes)) || upgradeOperation.LastUpdatedTime.IsZero() {
```
barkbay, line 213: I would add the following kind of unit test to ensure that the status is updated only when the reporter has been used (`u.nodes != nil`):

```golang
		{
			name: "Do not overwrite existing status if reporter is not used",
			state: func() *State {
				s := MustNewState(esv1.Elasticsearch{})
				return s
			},
			args: args{
				otherStatus: esv1.ElasticsearchStatus{
					Phase: esv1.ElasticsearchResourceInvalid,
					Conditions: esv1.Conditions{
						{
							Type:    "ReconciliationComplete",
							Status:  "False",
							Message: "Nodes upgrade in progress",
						},
					},
					InProgressOperations: esv1.InProgressOperations{
						DownscaleOperation: esv1.DownscaleOperation{
							Nodes: nil,
						},
						UpgradeOperation: esv1.UpgradeOperation{
							LastUpdatedTime: metav1.Now(),
							Nodes: []esv1.UpgradedNode{
								{
									Message:   pointer.String("Cannot restart node because of failed predicate"),
									Name:      "node-1",
									Status:    "PENDING",
									Predicate: pointer.String("if_yellow_only_restart_upgrading_nodes_with_unassigned_replicas"),
								},
							},
						},
						UpscaleOperation: esv1.UpscaleOperation{
							Nodes: nil,
						},
					},
				},
			},
			wantElasticsearchStatus: esv1.ElasticsearchStatus{
				Phase: esv1.ElasticsearchResourceInvalid,
				Conditions: esv1.Conditions{
					{
						Type:    "ReconciliationComplete",
						Status:  "False",
						Message: "Nodes upgrade in progress",
					},
				},
				InProgressOperations: esv1.InProgressOperations{
					DownscaleOperation: esv1.DownscaleOperation{
						Nodes: nil,
					},
					UpgradeOperation: esv1.UpgradeOperation{
						Nodes: []esv1.UpgradedNode{
							{
								Message:   pointer.String("Cannot restart node because of failed predicate"),
								Name:      "node-1",
								Status:    "PENDING",
								Predicate: pointer.String("if_yellow_only_restart_upgrading_nodes_with_unassigned_replicas"),
							},
						},
					},
					UpscaleOperation: esv1.UpscaleOperation{
						Nodes: nil,
					},
				},
			},
			wantPendingNewNodes: false,
		}
```

barkbay (approved):
LGTM

Issue #1230, opened 2019-07-12
Break Jenkins pipeline into more granular stages


charith-elastic:
In the current pipeline, the "Run unit and integration tests" stage performs the following operations:

- Build and test operator
    - Download dependencies
    - Run linters
    - Generate code
    - Run unit tests
    - Run integration tests
    - Compile e2e tests
    - Build container image
- Build and test local volume
    - Download dependencies
    - Build
    - Run unit tests

![Screenshot from 2019-07-12 13-15-25](https://user-images.githubusercontent.com/52399125/61127443-33f99880-a4a7-11e9-9b03-86eef850f1a5.png)

If we add more stages to the pipeline declaration to match the above set of steps, it would make it very easy to locate a failure and jump to the relevant log section. It would also provide a nice visualisation of the current progress of the build as well.

### Pros
- Easy to visualise the build steps and locate the cause of failures

### Cons
- Jenkinsfile becomes more complicated
- Makefile and Jenkinsfile should be kept in sync to get the best results

pebrc:
One idea we discussed and which would give us similar visualisation of the different steps would be to break up the current build into separate Jobs and call them as separate checks from GitHub. 

This would have the added benefit of making the steps immediately visible in each PR and remove the need to go to the Jenkins UI to see the result of a build. 

It comes with the downside of Jenkins jobs being comparatively heavy-weight. 
It also would only improve the situation for PR builds while the analysis of every other build would still mean that we have to inspect the raw build output in Jenkins.

PR #5535, opened 2022-03-29
Break Jenkins pipeline into more granular stages

thbkrkr:
Opening of the PR to test the benefit that this can bring.

Relates to #1230.


thbkrkr:
The last modification removes the parallelism at the beginning of the pipeline because we need the CI image before doing anything else (even `validate-jenkins-pipelines`).

2022-04-25-094807_2104x520_scrot

hack/manifest-gen/manifest-gen.sh
```diff
@@ -73,9 +73,10 @@ while getopts "cug" OPT; do
             shift $((OPTIND-1))
             (
                 cd "$SCRIPT_DIR"
-                go build -o manifest-gen >/dev/null 2>&1
-                ./manifest-gen --source="$EFFECTIVE_SRC_CHART_DIR" generate "$@"
-                rm manifest-gen
+                tmpBinary=$(mktemp /tmp/manifest-gen.XXXXXX)
```
pebrc, line 76: What is the motivation for this change?

.ci/pipelines/pr-gke.Jenkinsfile
```diff
@@ -20,65 +20,124 @@ pipeline {
...
-                            junit "e2e-tests.xml"
-
-                            sh 'exit $SHELL_EXIT_CODE'
+                }
+                stage('e2e-tests') {
+                    stages {
+                        stage('e2e-tests') {
+                            steps {
+                                script {
+                                    env.SHELL_EXIT_CODE = sh(returnStatus: true, script: 'make -C .ci TARGET="set-kubeconfig e2e-run" ci')
```
pebrc, line 135: `set-kubeconfig` does not deployer automatically set up the kubeconfig after bootstrapping the k8s cluster?

.ci/pipelines/pr-gke.Jenkinsfile
```diff
@@ -20,65 +20,124 @@ pipeline {
...
+                expression {
+                    notOnlyDocs()
+                }
+            }
+            failFast true
+            parallel {
+                stage("build-e2e-tests-docker-image") {
                     steps {
-                        sh 'make -C .ci TARGET=validate-jenkins-pipelines ci'
+                        sh 'make -C .ci TARGET=e2e-docker-multiarch-build ci'
```
pebrc, line 55: We used to call `e2e-compile` before creating the image to catch compile errors in e2e tests early. I am not sure if the extra time is worth the tradeoff though. 

Also I think we have a few unused Makefile targets now like `setup-e2e` that we could potentially get rid of?

pebrc (approved):
LGTM I am in favour of trying this and adjusting as needed.

hack/manifest-gen/manifest-gen.sh
thbkrkr: It is to be able to introduce parallelism. It's due to a few things:
- This script is called by the `generate` make target
- We run the CI container by mounting the current directory `-v $(ROOT_DIR):$(GO_MOUNT_PATH)`, so several runs of the CI container in parallel share the same directory
- several steps depend on the `generate` target
So, when we run in parallel  `make -C .ci TARGET=xxx` and `make -C .ci TARGET=yyy ci`, there may be a race condition where the binary generated by `xxx` is removed by `yyy`.

At one point I was massively testing parallelism and got the problem. It may not be really necessary now but I left it.

.ci/pipelines/pr-gke.Jenkinsfile
thbkrkr: Indeed, I removed `e2e-compile` because I wanted to optimize to reduce execution time and like you said I don't think it is worth it. As it is rare to introduce such error, I think we pay the extra time in 99% of the case.
With the latest pipeline layout, compile errors in e2e tests will be catched early because the pipeline starts by building the e2e-tests image.

.ci/pipelines/pr-gke.Jenkinsfile
thbkrkr: Yes, the deployer does it well but it is not persisted outside the container, it is written in the default location `~/.kube/config`. It worked in the past because, we stayed in the same container to bootstrap the k8s cluster and then run the e2e tests, which is no more the case now.

.ci/pipelines/pr-gke.Jenkinsfile
thbkrkr: > Also I think we have a few unused Makefile targets now like setup-e2e that we could potentially get rid of?

The other pipelines are still using it. The goal is to split them too and then we can clean up.

PR #5536, opened 2022-03-29
Elastic Maps: Set status.ObservedGeneration from metadata.Generation

naemono:
related to #3392

This change allows status.ObservedGeneration to be kept in sync with metadata.Generation according to the rules described in the above issue. This change adds a set of unit tests which show more details into when observedGeneration will, and will not be updated according to any errors/situations that may occur during reconciliation. End to end tests are in progress.

Remaining Items (Reason for Draft)

- [ ] E2E test

### DIFF
--- a/pkg/controller/maps/controller.go
+++ b/pkg/controller/maps/controller.go
@@ -167,9 +167,34 @@ func (r *ReconcileMapsServer) Reconcile(ctx context.Context, request reconcile.R
 		return reconcile.Result{}, nil
 	}
 
+	// MapsServer will be deleted nothing to do other than remove the watches
+	if ems.IsMarkedForDeletion() {
+		return reconcile.Result{}, r.onDelete(k8s.ExtractNamespacedName(&ems))
+	}
+
+	// main reconciliation logic
+	results, status := r.doReconcile(ctx, ems)
+	updateStatusErr := r.updateStatus(ems, status)
+	if updateStatusErr != nil {
+		if apierrors.IsConflict(updateStatusErr) {
+			log.V(1).Info(
+				"Conflict while updating status",
+				"namespace", ems.Namespace,
+				"beat_name", ems.Name)
+			return results.WithResult(reconcile.Result{Requeue: true}).Aggregate()
+		}
+		results.WithError(updateStatusErr)
+	}
+	return results.Aggregate()
+}
+
+func (r *ReconcileMapsServer) doReconcile(ctx context.Context, ems emsv1alpha1.ElasticMapsServer) (*reconciler.Results, emsv1alpha1.MapsStatus) {
+	results := reconciler.NewResult(ctx)
+	status := newStatus(ems)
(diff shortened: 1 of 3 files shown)


naemono:
run full pr build

pkg/controller/maps/controller.go
```diff
@@ -167,48 +167,63 @@ func (r *ReconcileMapsServer) Reconcile(ctx context.Context, request reconcile.R
...
+	if updateStatusErr != nil {
+		if apierrors.IsConflict(updateStatusErr) {
+			log.V(1).Info(
+				"Conflict while updating status",
+				"namespace", ems.Namespace,
+				"beat_name", ems.Name)
+			return results.WithResult(reconcile.Result{Requeue: true}).Aggregate()
+		}
+		results.WithError(updateStatusErr)
+	}
```
pebrc, line 187:
```suggestion
	if err := r.updateStatus(ems, status); err != nil {
		if apierrors.IsConflict(err) {
			log.V(1).Info(
				"Conflict while updating status",
				"namespace", ems.Namespace,
				"beat_name", ems.Name)
			return results.WithResult(reconcile.Result{Requeue: true}).Aggregate()
		}
		results.WithError(err)
	}
```

pkg/controller/maps/controller_test.go
```diff
@@ -91,6 +98,11 @@ func TestReconcileMapsServer_Reconcile(t *testing.T) {
 			post: func(r ReconcileMapsServer) {
 				// watches should have been cleared
 				require.Empty(t, r.DynamicWatches().Secrets.Registrations())
+
+				// observedGeneration should not have been updated
+				var ems v1alpha1.ElasticMapsServer
+				require.NoError(t, r.Get(context.Background(), types.NamespacedName{Name: nsnFixture.Name, Namespace: nsnFixture.Namespace}, &ems))
+				require.Equal(t, int64(1), ems.Status.ObservedGeneration)
```
pebrc, line 105: Given that this is repeated many times, you could have a little helper function defined in the scope of the test that you just call here.

At the top:
```golang
func TestReconcileMapsServer_Reconcile(t *testing.T) {
	timeFixture := metav1.Now()

	assertObservedGeneration := func(r k8s.Client, expected int) {
		var ems v1alpha1.ElasticMapsServer
		require.NoError(t, r.Get(context.Background(), types.NamespacedName{Name: nsnFixture.Name, Namespace: nsnFixture.Namespace}, &ems))
		require.Equal(t, int64(expected), ems.Status.ObservedGeneration)
	}
```
and then here
```golang
			post: func(r ReconcileMapsServer) {
				// watches should have been cleared
				require.Empty(t, r.DynamicWatches().Secrets.Registrations())
				// observedGeneration should not have been updated
				assertObservedGeneration(r, 1)
			},
```

pebrc (approved):
LGTM 👍

Issue #4812, opened 2021-09-02
Handling certificates issued by well known CAs


barkbay:
While working on https://github.com/elastic/cloud-on-k8s/issues/4790 using certificates generated by Let's Encrypt, and deployed on Elasticsearch and Kibana, I discovered that we don't handle certificate verification consistently across the stack:

Some associations work but the certificate verification is "downgraded" using `*.verificationMode: certificate`:
* `Kibana` to `Elasticsearch` and  `EnterpriseSearch`
* `Maps` to `Elasticsearch`

Associations fail, and require the user to set the url manually for:
* `APMServer` to `Elasticsearch` and `Kibana`, user must set `apm-server.kibana.host` and `output.elasticsearch.hosts` manually.
* I think it is the same for `Beat` and `EnterpriseSearch` to `Elasticsearch`, the same workaround should work (not tested)

There is a known issue for `Agent`, [investigations are in progress](https://github.com/elastic/cloud-on-k8s/issues/4790), I don't think there is workaround at the moment

I was wondering if we want to document how to use certificates issued by well known certificate authorities and/or try to have a more consistent behaviour.

PR #5539, opened 2022-03-30
Support a globally shared CA

pebrc:
This adds support for a new operator parameter that allows users to specify a shared CA that will be used for all managed resources. If users opted to specify separate CAs per resource those configurations will be left untouched, however any resources that are using the default per-resource self-signed CA will use the shared CA. 

The CA is specified as a directory via an operator param, currently `--ca`. The operator expects two files to be present in this directory `tls.crt` and `tls.key`. I am thinking of supporting `ca.crt` and `ca.key` as well to be consistent with the existing per-resource custom CA feature.

The operator watches the directory and restarts if it detects changes. This is following the behaviour we have for general operator configuration changes and simplifies the implementation IMO considerably. 

I am still trying to figure out a testing strategy (unit/integaition tests but maybe also an e2e test) but wanted to raise the PR already in case there is early feedback.


pebrc:
I am not 💯  sure about supporting the two naming conventions `ca.*` and `tls.*` (the latter being the standard in K8s TLS secrets) and therefore happy to back out that last commit https://github.com/elastic/cloud-on-k8s/pull/5539/commits/b505883cabd79b2a89e1bdd8ce7c772287b42d38

pkg/controller/common/certificates/ca_reconcile.go
```diff
@@ -232,6 +235,66 @@ func internalSecretForCA(
...
+	}
+	switch {
+	case (files[CertFileName] || files[KeyFileName]) && files[CAKeyFileName]:
+		return "", "", fmt.Errorf("both tls.* and ca.* files exist, configuration error")
+	case files[CAFileName] && files[CAKeyFileName]:
+		return filepath.Join(path, CAFileName), filepath.Join(CAKeyFileName), nil
+	case files[CertFileName] && files[KeyFileName]:
+		return filepath.Join(path, CertFileName), filepath.Join(path, KeyFileName), nil
+	}
+	return "", "", fmt.Errorf("no CA certificate files found: %+v", files)
```
barkbay, line 260: I was trying to think how we can avoid to leak the map implementation in the error message, and also exit early if the directory does not even exist.

```suggestion
func detectCAFileNames(path string) (string, string, error) {
	if !fs.FolderExists(path) {
		return "", "", fmt.Errorf("global CA directory %s does not exist", path)
	}
	caFiles := []string{CAFileName, CAKeyFileName}
	tlsFiles := []string{CertFileName, KeyFileName}
	existsInDirectory := map[string]bool{}

	for _, f := range append(caFiles, tlsFiles...) {
		exists, err := fs.FileExists(filepath.Join(path, f))
		if err != nil {
			return "", "", err
		}
		existsInDirectory[f] = exists
	}
	switch {
	case (existsInDirectory[CertFileName] || existsInDirectory[KeyFileName]) && existsInDirectory[CAKeyFileName]:
		return "", "", fmt.Errorf("both tls.* and ca.* files exist, configuration error")
	case existsInDirectory[CAFileName] && existsInDirectory[CAKeyFileName]:
		return filepath.Join(path, CAFileName), filepath.Join(CAKeyFileName), nil
	case existsInDirectory[CertFileName] && existsInDirectory[KeyFileName]:
		return filepath.Join(path, CertFileName), filepath.Join(path, KeyFileName), nil
	}
	return "", "",
		fmt.Errorf(
			"no CA certificate files found in %s, expecting one of the following key pair: (%s) or (%s)",
			path,
			strings.Join(caFiles, ","),
			strings.Join(tlsFiles, ","))
}
```

barkbay (commented):
I only did a first review, but looks good so far 👍 
Thanks for the `FileWatcher`, I'm pretty sure it'll make config file detection more reliable.

test/e2e/ca_test.go
```diff
@@ -0,0 +1,192 @@
...
+	"github.com/ghodss/yaml"
+	"github.com/stretchr/testify/require"
+	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"testing"
+	"time"
+)
+
+func TestSharedCA(t *testing.T) {
```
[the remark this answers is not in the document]
pebrc: Only if you deploy the operator through the e2e test mechanism into the k8s cluster with the empty secret mounted. Testing against a locally running operator is tricky not sure how we would do that.

pkg/utils/fs/watcher.go
```diff
@@ -0,0 +1,85 @@
...
+		if err != nil {
+			if os.IsNotExist(err) && !prev.IsZero() {
+				// file was deleted
+				updated = append(updated, f)
+				fmc[f] = time.Time{}
+			} else {
+				log.Error(err, "while getting file info", "file", f, "err", err.Error())
+			}
+			continue
+		}
```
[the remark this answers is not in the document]
pebrc: well spotted!

pkg/controller/common/certificates/ca_reconcile.go
pebrc: Good idea! Do you have an opinion on https://github.com/elastic/cloud-on-k8s/pull/5539#issuecomment-1087481662 i.e. whether we should support two naming conventions for the files to begin with?

barkbay (commented):
I have run a few other tests, including running the E2E test, or enabling/disabling the feature on existing deployments. It seems to work as expected 👍 During these tests I only noticed that the `*-ca-internal` Secrets are not removed, while my understanding is that they are not required anymore. However it's definitely not a big deal since it's completely harmless.

I'll take the time to do another review later.

pkg/controller/common/certificates/ca_reconcile.go
barkbay: > Do you have an opinion on [#5539 (comment)](https://github.com/elastic/cloud-on-k8s/pull/5539#issuecomment-1087481662) i.e. whether we should support two naming conventions for the files to begin with?

It seems fair to support both, and I don't think I have a strong argument against this proposal.

pkg/controller/common/certificates/ca_reconcile.go
```diff
@@ -232,6 +235,66 @@ func internalSecretForCA(
...
+		return nil, err
+	}
+
+	bytes, err := ioutil.ReadFile(certFile)
+	if err != nil {
+		return nil, err
+	}
+	certs, err := ParsePEMCerts(bytes)
+	if err != nil {
+		return nil, errors.Wrapf(err, "Cannot parse PEM cert from %s", certFile)
```
naemono, line 275:
```suggestion
		return nil, errors.Wrapf(err, "cannot parse PEM cert from %s", certFile)
```

pkg/controller/common/certificates/ca_reconcile.go
```diff
@@ -232,6 +235,66 @@ func internalSecretForCA(
...
+	}
+	cert := certs[0]
+
+	privateKeyBytes, err := ioutil.ReadFile(privateKeyFile)
+	if err != nil {
+		return nil, err
+	}
+	privateKey, err := ParsePEMPrivateKey(privateKeyBytes)
+	if err != nil {
+		return nil, errors.Wrapf(err, "Cannot parse private key from PEM file %s", privateKeyFile)
```
naemono, line 293:
```suggestion
		return nil, errors.Wrapf(err, "cannot parse private key from PEM file %s", privateKeyFile)
```

pkg/controller/common/certificates/ca_reconcile_integration_test.go
```diff
@@ -0,0 +1,171 @@
...
+
+import (
+	"fmt"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
```
naemono, line 17:
```suggestion
	"io/ioutil"
	"os"
	"path/filepath"
	"strings"
	"testing"

	"github.com/stretchr/testify/assert"
	"github.com/stretchr/testify/require"
```

pkg/controller/elasticsearch/certificates/transport/ca.go
```diff
@@ -51,7 +52,12 @@ func ReconcileOrRetrieveCA(
 		return nil, err
 	}
 	// 1. No custom certs are specified, reconcile our internal self-signed CA instead (probably the common case)
+	// or return the shared CA
```
naemono, line 55:
```suggestion
	// or return the global CA
```

pkg/utils/fs/watcher.go
```diff
@@ -0,0 +1,85 @@
...
+	ctx      context.Context
+	onChange func([]string)
+	period   time.Duration
+	cache    fileModTimeCache
+}
+
+// NewFileWatcher creates a new file watcher, use ctx context for cancellation, paths to specify the files to watch.
+// onChange is a callback to be invoked when changes are detected, a list of affected files will be passed as argument.
+// period determines how often the file watcher will try to detect changes to the files of interest.
+func NewFileWatcher(ctx context.Context, paths []string, onChange func([]string), period time.Duration) *FileWatcher {
```
naemono, line 28: Would `interval` better describe the argument currently named `period`?

test/e2e/test/elasticsearch/builder.go
```diff
@@ -54,6 +54,8 @@ type Builder struct {
 	// situations where the Elasticsearch resource is modified by an external mechanism, like the autoscaling controller.
 	// In such a situation the actual resources may diverge from what was originally specified in the builder.
 	expectedElasticsearch *esv1.Elasticsearch
+
+	SharedCA bool
```
naemono, line 58: I'm seeing two different naming conventions.  Is it a shared CA, or a global CA?  Or do we want to support both?

pebrc: Good question, I haven't made up my mind yet. But of course a single name for this concept would be preferable.

pebrc:
Jenkins test this please

>  Makefile:63: recipe for target 'ci-build-image' failed

cmd/manager/main.go
```diff
@@ -501,6 +511,13 @@ func startOperator(ctx context.Context) error {
 		return err
 	}
 
+	// Retrieve shared CA if any
```
naemono, line 514:
```suggestion
	// Retrieve global CA if any
```

test/e2e/global_ca_test.go
```diff
@@ -0,0 +1,194 @@
...
+	elasticagent "github.com/elastic/cloud-on-k8s/test/e2e/test/agent"
+	"github.com/elastic/cloud-on-k8s/test/e2e/test/beat"
+	"github.com/elastic/cloud-on-k8s/test/e2e/test/elasticsearch"
+	"github.com/elastic/cloud-on-k8s/test/e2e/test/enterprisesearch"
+	"github.com/elastic/cloud-on-k8s/test/e2e/test/kibana"
+)
+
+func TestGlobalCA(t *testing.T) {
+	k := test.NewK8sClientOrFatal()
+	name := "shared-ca"
```
naemono, line 37:
```suggestion
	name := "global-ca"
```

pebrc:
Unfortunately it looks like the file watcher integration test I added is flaky. Trying to improve on the current implementation.

pebrc:
Sometimes mtime changes more than once on writes. I am not entirely sure why given the test case uses probably a single write syscall. But then I again I am not an expert on how these things work on unix'ish OSs. 

```
/var/folders/1x/dzn9_byd2sn3cpn48j3d14m40000gn/T/file-watcher-test105416277/file2 prev 0001-01-01 00:00:00 +0000 UTC now 2022-04-19 20:52:24.299815768 +0200 CEST
/var/folders/1x/dzn9_byd2sn3cpn48j3d14m40000gn/T/file-watcher-test105416277/file1 prev 0001-01-01 00:00:00 +0000 UTC now 2022-04-19 20:52:24.300063142 +0200 CEST
/var/folders/1x/dzn9_byd2sn3cpn48j3d14m40000gn/T/file-watcher-test105416277/file3 prev 0001-01-01 00:00:00 +0000 UTC now 2022-04-19 20:52:24.300206579 +0200 CEST
/var/folders/1x/dzn9_byd2sn3cpn48j3d14m40000gn/T/file-watcher-test105416277/file4 prev 0001-01-01 00:00:00 +0000 UTC now 2022-04-19 20:52:24.302395466 +0200 CEST
/var/folders/1x/dzn9_byd2sn3cpn48j3d14m40000gn/T/file-watcher-test105416277/file4 prev 2022-04-19 20:52:24.302395466 +0200 CEST now 2022-04-19 20:52:24.302612857 +0200 CEST
/var/folders/1x/dzn9_byd2sn3cpn48j3d14m40000gn/T/file-watcher-test105416277/file2 prev 2022-04-19 20:52:24.299815768 +0200 CEST now 2022-04-19 20:52:24.302811553 +0200 CEST
```

I modified the test's assumptions to allow for additional events on the simulated writes.

barkbay:
> I am not entirely sure why given the test case uses probably a single write syscall.

I think that there are always at least 2 underlying syscall:
1. A first one to create an empty file, or truncate the content if it already exists: `mtime` is changed and file size is `0`
2. A second one to write the data to the empty file , `mtime` is changed once again

Running the [`WriteFile`](https://github.com/golang/go/blob/2ebe77a2fda1ee9ff6fd9a3e08933ad1ebaea039/src/os/file.go#L699-L712) in debug mode, step by step, and using `stat` at the same time we can see that `mtime` is changing at least twice:

```golang
func WriteFile(name string, data []byte, perm FileMode) error {
	f, err := OpenFile(name, O_WRONLY|O_CREATE|O_TRUNC, perm) // -- File is created or truncated, `stat -f "%Sm" file1 = Apr 20 08:56:10 2022`
	if err != nil {...}
	_, err = f.Write(data) // -- data is written (not sure if it is synced though) -- `stat -f "%Sm" file1 = Apr 20 08:56:16 2022`
	if err1 := f.Close(); err1 != nil && err == nil { // -- data is flushed on disk?
		err = err1
	}
	return err
}
```

Also if the file is mapped to several pages in memory, I'm not sure there is a guarantee that all the pages are synced at the same time and `mtime` is changed only once. I don't know if [this article](https://lwn.net/Articles/564120/) is still relevant, but it is still interesting as it describes a behavior that has existed:

> So a quick operation to make a page writable turns into a heavyweight filesystem operation, and it happens every time the application attempts to write to a clean page. If the application writes large numbers of pages that have been mapped into memory, the result will be a painful slowdown. And most of that effort is wasted; the timestamp updates overwrite each other, so only the last one will persist for any useful period of time.

pebrc:
> A first one to create an empty file, or truncate the content if it already exists: mtime is changed and file size is 0

Of course! I clearly didn't think this through. Thanks for the great explanation! 

> Also if the file is mapped to several pages in memory, I'm not sure there is a guarantee that all the pages are synced at the same time and mtime is changed only once. 

Do you mean this in the sense that the test might still be flaky and we could see yet another event coming from another mtime update?

barkbay (approved):
This looks great, I just left some nitpicks 👍 🚀

pkg/controller/common/certificates/ca_reconcile_integration_test.go
```diff
@@ -0,0 +1,178 @@
...
+			},
+			want:       nil,
+			wantErrMsg: "did not contain any certificates",
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			tempDir, err := ioutil.TempDir("testdata", "ca-from-file")
+			defer os.RemoveAll(tempDir)
+			require.NoError(t, err)
```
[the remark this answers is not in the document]
pebrc: TIL!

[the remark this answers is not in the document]
naemono: right?  Me as well.

ejsmith:
Is it possible to use this feature to get ES, Kibana and APM to use certs that are signed by a trusted authority like Let's Encrypt so that my app trusts the cert without having to configure my app to add a trusted self-signed CA?

jmp601:
> Is it possible to use this feature to get ES, Kibana and APM to use certs that are signed by a trusted authority like Let's Encrypt so that my app trusts the cert without having to configure my app to add a trusted self-signed CA?

We are an elastic enterprise customer and have the same ask. I have a hard time configuring ECK with e2e with Let's encrypt certs.

naemono:
We have some documentation around this feature:

https://www.elastic.co/guide/en/cloud-on-k8s/master/k8s-tls-certificates.html#k8s-setting-up-your-own-certificate

and here an example for cert manager (you can replace the self-signed issue of course with a publicly trusted one like Let's Encrypt)
https://www.elastic.co/guide/en/cloud-on-k8s/master/k8s-custom-http-certificate.html#k8s_custom_self_signed_certificate_using_cert_manager

We also have an issue tracking updating our documentation to be more clear https://github.com/elastic/cloud-on-k8s/issues/4812

ejsmith:
Would love to see a tutorial that sets up cert-manager with let's encrypt and then creates an ES cluster with a trusted SSL feet that apps can use without having to do anything as well as exposing a public Kibana and APM instance that an app can use to send front end APM data to.

jbouse:
Is this still available in v2.9.0? I'm unable to find mention of it in the documentation nor in the Helm chart for the operator deployment.

PR #5542, opened 2022-03-31
Fix e2e monitoring file/metricbeat configuration 

naemono:
Since e2e monitoring Elasticsearch cluster is behind a cloud ingress Load Balancer at this time, the filebeat/metricbeat configuration needs to be tweaked to work in that scenario.  See https://github.com/elastic/beats/issues/3688

### DIFF
--- a/config/e2e/monitoring.yaml
+++ b/config/e2e/monitoring.yaml
@@ -84,12 +84,8 @@ spec:
       password: ${monitoring_pass}
       # Since E2E cluster's Elasticsearch instance may lag behind the newest version of Beat, we need to allow this scenario.
       allow_older_versions: true
-      # The maximum size to send in a single Elasticsearch bulk API index request. Needed if Elasticsearch is behind Cloud Ingress.
-      bulk_max_body_size: 10M
       # The maximum number of events to bulk in a single Elasticsearch bulk API index request. Needed if Elasticsearch is behind Cloud Ingress.
       bulk_max_size: 10
-      # The maximum byte size to send in a single Elasticsearch bulk API index request, regardless of compression. Needed if Elasticsearch is behind Cloud Ingress.
-      bulk_max_bytes: 10485760
   daemonSet:
     podTemplate:
       spec:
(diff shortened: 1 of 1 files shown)


config/e2e/monitoring.yaml
```diff
@@ -82,6 +82,14 @@ spec:
       hosts: ['${monitoring_url}']
       username: ${monitoring_user}
       password: ${monitoring_pass}
+      # Since E2E cluster's Elasticsearch instance may lag behind the newest version of Beat, we need to allow this scenario.
+      allow_older_versions: true
+      # The maximum size to send in a single Elasticsearch bulk API index request. Needed if Elasticsearch is behind Cloud Ingress.
+      bulk_max_body_size: 10M
```
pebrc, line 88: I am confused:  have we not been running the monitor behind an ingress for a long time? Why is this only now a problem?

naemono: Unsure what has recently changed, or changed with what I was ingesting, but I directly ran into:

```
client.go:244: ERR Failed to perform any bulk index operations: 413 Request Entity Too Large
```

during my testing of this during a local e2e test run.  This caused no ingestion at all into the ES cluster, and everything after failed.

config/e2e/monitoring.yaml
```diff
@@ -278,6 +286,14 @@ spec:
...
+      # The maximum number of events to bulk in a single Elasticsearch bulk API index request. Needed if Elasticsearch is behind Cloud Ingress.
+      bulk_max_size: 10
+      # The maximum byte size to send in a single Elasticsearch bulk API index request, regardless of compression. Needed if Elasticsearch is behind Cloud Ingress.
+      bulk_max_bytes: 10485760
```
pebrc, line 296: I struggle to find any reference to these options in Beats documentation with the exception of `allow_older_versions` and `bulk_max_size` The only thing I could find was a user requesting for those to be added, but afaik that never happened. Do they exist?

naemono: My apologies.  I really should have referenced the [beat configuration reference](https://www.elastic.co/guide/en/beats/filebeat/current/filebeat-reference-yml.html) before adding those.  I'm testing this configuration locally now again to ensure I don't see any `413 Request Entity Too Large` errors again.

naemono:
run full pr build

pebrc (approved):

PR #5547, opened 2022-04-01
Improve E2E tests runner Docker image

### LINKED ISSUE #5546: Improve E2E tests runner Docker image
The Docker image for the E2E tests runner is built to run as non-root user `1001` to support Kubernetes clusters with restrictive PSP.

https://github.com/elastic/cloud-on-k8s/blob/c80040884e369835af38236c932341fe8186a7cc/test/e2e/Dockerfile#L17-L28

The permissions of `/go` are updated using `chgrp -R 0 /go && chmod -R g=u /go` to let the user write files under `/go/src/github.com/elastic/cloud-on-k8s`, only **after** go sources were copied and go modules were copied. 

`/go` contains more than 50k files, which makes the execution time of the `chgrp/chmod` commands very long (~2min)!

```
[2022-04-01T01:38:08.535Z] #24 [linux/amd64 9/9] RUN chgrp -R 0 /go && chmod -R g=u /go
[2022-04-01T01:38:08.535Z] #24 DONE 131.1s
[2022-04-01T01:38:08.535Z] #15 [linux/arm64 9/9] RUN chgrp -R 0 /go && chmod -R g=u /go
[2022-04-01T01:38:23.536Z] #15 DONE 110.7s
```

By setting the right permissions **before** copying sources and downloading modules we can save minutes.

I think we can also be more restrictive and just run `chown -R 1001 /go/src/github.com/elastic/cloud-on-k8s` since writing in `/go/pkg` is already possible.

The average duration of the PR job is about [25 minutes](https://devops-ci.elastic.co/job/cloud-on-k8s-pr/buildTimeTrend), so a gain of 2 minutes represents an 8% decrease :party:.

thbkrkr:
This commit sets the non-root user `1001` before downloading go modules and copying sources to avoid to update permissions on 50k files and also makes the permissions change more precise.
 
Resolves #5546.

### DIFF
--- a/test/e2e/Dockerfile
+++ b/test/e2e/Dockerfile
@@ -10,21 +10,20 @@ ENV E2E_TAGS $E2E_TAGS
 
 WORKDIR /go/src/github.com/elastic/cloud-on-k8s
 
-# Create the go test cache directory
+# create the go test cache directory
 RUN mkdir -p /.cache && chmod 777 /.cache
 
-# Download dependencies
+# non-root user to support restricted PSP
+RUN chown -R 1001 /go/src/github.com/elastic/cloud-on-k8s
+USER 1001
+
+# download go dependencies
 COPY ["go.mod", "go.sum","./"]
 RUN go mod download
 
-# Copy the source
+# copy sources
 COPY pkg/ pkg/
 COPY config/ config/
 COPY test/ test/
-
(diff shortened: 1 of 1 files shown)


thbkrkr:
Jenkins test this please

pebrc (approved):

PR #5551, opened 2022-04-04
 Add v2.1.0 to upgrade harness

naemono:
This adds version 2.1.0 of the operator and Elastic Stack 8.1 to the upgrade harness.  This also updates the default retry timeout from 5mins -> 15 mins, as the typical 3-mode Elasticsearch cluster upgrade, in every upgrade test I ran, takes this amount of time to come back to a 'Green' state, and would fail with the below error:

```
2022-04-04T11:36:29.868-0500	ERROR	eck-upgrade.fixture	[FAIL] CheckElasticsearch[v210]	{"step": "TestStatusOfResources[v210]", "step": "CheckElasticsearch[v210]", "error": "status mismatch: want={health:green nodes:3 version:8.1.0} have={health:green nodes:2 version:8.0.0}: retriable error"}
```

### DIFF
--- a/hack/upgrade-test-harness/testdata/upcoming/stack.yaml
+++ b/hack/upgrade-test-harness/testdata/upcoming/stack.yaml
@@ -4,7 +4,7 @@ kind: Elasticsearch
 metadata:
   name: es
 spec:
-  version: 8.0.0-SNAPSHOT
+  version: 8.2.0-SNAPSHOT
   nodeSets:
   - name: default
     count: 3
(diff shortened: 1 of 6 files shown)


hack/upgrade-test-harness/testdata/upcoming/stack.yaml
```diff
@@ -4,7 +4,7 @@ kind: Elasticsearch
 metadata:
   name: es
 spec:
-  version: 7.17.0-SNAPSHOT
+  version: 8.0.0-SNAPSHOT
```
thbkrkr, line 7: It looks like this hasn't been updated before, shouldn't it be `8.2.0-SNAPSHOT`?

naemono: I think you're right @thbkrkr .  I've updated all of the versions in this file.

thbkrkr (approved):
:rocket:

PR #5554, opened 2022-04-06
E2E: use pre-stop hook variable to speed up tests

pebrc:
This PR reconfigures the Elasticsearch pre-stop hook to wait for 0 seconds by default. The idea is to speed up the tests by not "draining" connections in the pre-stop hook. 

I reckon that we can be 20% faster on a typical 3-node cluster test with one mutation of every node. 

The controversial bit is of course that we do no longer test the default configuration that we ship with the operator to customers. But given that most of our e2e tests do not verify that Elasticsearch is not dropping connections thanks to the pre-stop hook I think benefit of faster e2e test runs is maybe bigger. 

I left the pre-stop hook intact for the load test we have, that verifies that all connection attempts succeed. But I am not sure if that is actually relevant because the client in this test does not use keep alive anyway.

### DIFF
--- a/test/e2e/test/elasticsearch/builder.go
+++ b/test/e2e/test/elasticsearch/builder.go
@@ -463,7 +463,8 @@ func (b Builder) WithEnvironmentVariable(name, value string) Builder {
 	return b
 }
 
-// WithPreStopAdditionalWaitSeconds allows speeding up the test by shortening the pre-stop hook runtime.
+// WithPreStopAdditionalWaitSeconds updates the PRE_STOP_ADDITIONAL_WAIT_SECONDS environment variable in the Elasticsearch container.
+// It can be used to speed up the test by shortening the pre-stop hook runtime.
 // Don't use if you want to test that Elasticsearch is not dropping connections.
 func (b Builder) WithPreStopAdditionalWaitSeconds(s int32) Builder {
 	for i := range b.Elasticsearch.Spec.NodeSets {


thbkrkr:
run/e2e-tests tags=es

pebrc:
run/e2e-tests tags=es

pebrc:
```
steps_mutation.go:142: 
        	Error Trace:	steps_mutation.go:142
        	Error:      	Not equal: 
        	            	expected: 0
        	            	actual  : 1
        	Test:       	TestMutationAndReversal/Elasticsearch_cluster_health_should_not_have_been_red_during_mutation_process#01
    steps_mutation.go:144: Elasticsearch cluster health check failure at 2022-04-07 10:19:58.403645989 +0000 UTC m=+5045.890779627: cluster health red
```
This took me a while: but this is actually a bug in our test code. We reconfigure the data integrity check to have 0 replicas and then do a rolling upgrade which cannot be done with green health like this. Going to add a fix to this PR.

pebrc:
run/e2e-tests tags=es

pebrc:
run/e2e-tests tags=es

pebrc:
run/e2e-tests match=TestMutationWhileLoadTesting

pebrc:
run/e2e-tests match=TestMutationWhileLoadTesting

pebrc:
run/e2e-tests match=TestMutationWhileLoadTesting

pebrc:
The last run shows a 20% time saving compared to using the default `PRE_STOP_WAIT_TIME` If we think that would be worth the risk of testing the delayed shutdown only during the load test e2e test then this is ready for review now.

test/e2e/test/elasticsearch/builder.go
```diff
@@ -447,13 +447,38 @@ func (b Builder) WithMutatedFrom(builder *Builder) Builder {
...
+			if !update {
+				container.Env = append(container.Env, corev1.EnvVar{Name: name, Value: value})
+			}
 			b.Elasticsearch.Spec.NodeSets[i].PodTemplate.Spec.Containers[j].Env = container.Env
 		}
 	}
 	return b
 }
 
+// WithPreStopAdditionalWaitSeconds allows speeding up the test by shortening the pre-stop hook runtime.
```
barkbay, line 466:
```suggestion
// WithPreStopAdditionalWaitSeconds updates the PRE_STOP_ADDITIONAL_WAIT_SECONDS environment variable in the Elasticsearch container.
// It can be used to speed up the test by shortening the pre-stop hook runtime.
```

barkbay (approved):
> If we think that would be worth the risk of testing the delayed shutdown only during the load test e2e test then this is ready for review now.

I think I'm 👍 on the proposal, shortening the feedback loop is always a good thing, and testing the default shutdown process only with the load test makes sense.

PR #5555, opened 2022-04-06
Use latest released version 8x instead of the latest snapshot in E2E tests

thbkrkr:
Since the release of `8.0.0`, we no longer have to test the latest `8.x` snapshot in the E2E tests, we can use the latest released version instead. Otherwise it may cause instability in all E2E test jobs instead of only impacting the `snapshot-versions` job which we have to specifically test the snapshot versions.

https://github.com/elastic/cloud-on-k8s/blob/c4af92c740f8bcc9673bd6b999f66e88181149af/.ci/pipelines/e2e-tests-snapshot-versions-gke.Jenkinsfile#L68

This commit introduces `GetUpgradePathTo8x` to calculate the upgrade path depending the current Elastic Stack version which uses the `LatestReleasedVersion8x` instead of the `LatestSnapshotVersion8x`.

Before this PR, we tested:
- 7.17 to 8.2.0-SNAPSHOT
- 8.0.1 to 8.2.0-SNAPSHOT
- 8.1.1 to 8.2.0-SNAPSHOT

With this PR, we will test:
- 7.17.1 to 8.1.1
- 8.0.1 to 8.1.1
- 8.1.1 to 8.2.0-SNAPSHOT


thbkrkr:
It's a bit more involved as we don't have upgrade tests where the `dstVersion` comes from the context and a srcVersion in 7.x (only 6.x) to have a valid upgrade path, example:
```go
  srcVersion := test.LatestReleasedVersion7x
  dstVersion := test.Ctx().ElasticStackVersion
```
With this PR in this state, we will never test an upgrade to the latest 8.x snapshot despite the `snapshot-versions` job.

pebrc:
> With this PR in this state, we will never test an upgrade to the latest 8.x snapshot despite the snapshot-versions job.

I was thinking the same thing. Maybe we should set up the upgrade test the other way round, just as you have it in your example?

thbkrkr:
> Maybe we should set up the upgrade test the other way round, just as you have it in your example?

This means that we will no longer test a minor update in 8x. Not sure we can afford that?

For all the `Test.*Upgrade.*ToLatest8x `:

Currently:
```
srcVersion := test.Ctx().ElasticStackVersion
dstVersion := test.LatestSnapshotVersion8x
```

We test:
- 7.17  to 8.2.0-SNAPSHOT
- 8.0.1 to 8.2.0-SNAPSHOT
- 8.1.0 to 8.2.0-SNAPSHOT

------------------
With the proposed change:
```
srcVersion := test.Ctx().ElasticStackVersion
dstVersion := test.LatestReleasedVersion8x
```

We will test:
- 7.17  to 8.1.0
- 8.0.1 to 8.1.0

We don't have an upgrade test with the `8.2.0-SNAPSHOT`.

------------------
If we set:
```
srcVersion := test.LatestReleasedVersion7x
dstVersion := test.Ctx().ElasticStackVersion
```

We will test:
- 7.17 to 8.0.1
- 7.17 to 8.1.0
- 7.17 to 8.2.0-SNAPSHOT

We will no longer test a minor update in 8x.

------------------

Unless we add a new test:

```
srcVersion := test.FirstReleasedVersion8x
dstVersion := test.Ctx().ElasticStackVersion
```
- 8.0.1 to 8.1.0
- 8.0.1 to 8.2.0-SNAPSHOT

thbkrkr:
Proposal with [6e31b3e](https://github.com/elastic/cloud-on-k8s/pull/5555/commits/6e31b3e0d0131d45d778787a57462823a8dfce44):

We keep the same logic as today but we replace LatestSnapshotVersion8x by LatestReleasedVersion8x:
```
srcVersion := test.Ctx().ElasticStackVersion
dstVersion := test.LatestReleasedVersion8x
```
But if `srcVersion > dstVersion`, we do the reverse.

pebrc (commented):
Looks good to me. I wonder if we should keep the upgrade to 7x test for Beats though

pebrc (approved):

PR #5557, opened 2022-04-06
E2E: remove Azure client from CI image

pebrc:
We need the Azure client in only two of our 20 build jobs. But the Azure client adds ~ 1GiB to the CI image. This PR follows the example of the kind and OCP drivers and runs the Azure commands from a dedicated Docker image. 

We do not need to create our own image for this, as we do for kind and OCP, because Microsoft offers already a ready-made image. 
Screenshot 2022-04-06 at 17 45 05

I have not tested this on CI yet. Only local tests so far.

I also refactored the package structure and naming, so apologies for the larger diff.


thbkrkr (approved):
LGTM

PR #5585, opened 2022-04-17
Fix all Azure CLI commands used in Tanzu deployer

pebrc:
This makes the necessary changes to use use the Docker based CLI also from the Tanzu deployer which was missed in the
original change that introduced the Docker based Azure CLI tooling.

### DIFF
--- a/hack/deployer/runner/tanzu.go
+++ b/hack/deployer/runner/tanzu.go
@@ -6,6 +6,7 @@ package runner
 
 import (
 	"bytes"
+	"encoding/json"
 	"fmt"
 	"io/ioutil"
 	"log"
@@ -229,6 +230,7 @@ func (t *TanzuDriver) create() error {
 	if err := t.dockerizedTanzuCmd("management-cluster", "create", "--file", cfgPathInContainer).Run(); err != nil {
 		return err
 	}
+
 	log.Println("Creating workload cluster")
 	if err := t.dockerizedTanzuCmd("cluster", "create", t.plan.ClusterName, "--file", cfgPathInContainer).Run(); err != nil {
 		return err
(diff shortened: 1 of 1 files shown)


hack/deployer/runner/tanzu.go
```diff
@@ -349,49 +351,57 @@ func (t *TanzuDriver) loginToAzure() error {
 // loginToContainerRegistry we use a private container registry to make the Tanzu CLI available in CI.
 func (t *TanzuDriver) loginToContainerRegistry() error {
 	log.Println("Logging in to container registry")
-	return exec.NewCommand("az acr login --name {{.ContainerRegistry}}").
-		AsTemplate(map[string]interface{}{
-			"ContainerRegistry": t.acrName,
-		}).Run()
+	// the Azure CLI image we use does not have a Docker client installed thus we extract a token here ...
+	jsonResp, err := azure.Cmd("acr", "login", "--name", t.acrName, "--expose-token").
```
barkbay, line 355: nit: I wonder if it is possible to avoid `--expose-token`. I would expect `$HOME/.docker/config.json` to be updated by `az acr login --name REDACTED` as it would be by `docker login -u .... -p token`?

pebrc: It does indeed update the docker config but the issue is that inside the Azure CLI container we don't have the Docker tooling so this update fails. We could now of course build our own image for the Azure tools that includes Dockers CLI. Or we could integrate the Azure tools into the Tanzu image we already maintain. 

I decided to use the method proposed in this PR because it avoids the additional maintenance burden of another Docker image and the (perceived) security disadvantage seemed acceptable.

barkbay: Makes sense, I was wrongly assuming that `az` could update the credentials without docker.
LGTM overall, I'm giving it a try in Jenkins 🤞

barkbay (approved):
LGTM 👍

PR #5586, opened 2022-04-19
[E2E] Use container image digest for ES dev. images

### LINKED ISSUE #5545: [E2E] Consider using Docker image digest instead of tag for pre-release versions
When testing snapshot releases an E2E test may end up creating nodes with different image builds for a same cluster, preventing the last restarted Elasticsearch node to join the cluster, for example:

```
remote node [..] is build [579695d032f3b28e9c1efc6c05aab596b1b1afca] of version [8.2.0] but this node is build [49dba59f288e539907098a559d163c3be2851f83] of version [8.2.0]
```

We should find a way to "pin" the image across all the tests, we could for example pull the image a first time and use the digest in the `spec.image` field.

To get/use the digest:
1. `docker inspect docker.elastic.co/elasticsearch/elasticsearch:8.2.0-SNAPSHOT`
2. Read digest in `RepoDigests`
3. `docker pull docker.elastic.co/elasticsearch/elasticsearch@sha256:digest`

barkbay:
This PR adds a utility function to retrieve a container image digest.
It is used in the context of Elasticsearch when a single and same version of a `SNAPSHOT` image must be used across all the nodes in a cluster.
Fixes #5545


test/e2e/test/registry.go
```diff
@@ -0,0 +1,152 @@
...
+	// Assert that we get a 200, otherwise attempt to parse body as a structured error.
+	if err := transport.CheckError(resp, http.StatusOK); err != nil {
+		return "", err
+	}
+
+	// Digest is not in the response body, it's in the header "Docker-Content-Digest"
+	digest := resp.Header.Get("Docker-Content-Digest")
+	if digest == "" {
+		return "", errors.New("No image digest in response")
+	}
```
pebrc, line 146: Is there a particular reason why this is not simply: 

```go
	withRegistry := fmt.Sprintf("%s/%s", container.DefaultContainerRegistry, taggedImage)
	digest, err := crane.Digest(withRegistry)
	if err != nil {
		return "", err
	}
```

barkbay: I spent too much time figuring out how the protocol works behind the scenes, I forgot to check if this was implemented elsewhere in the library 😄

Issue #4796, opened 2021-08-30
Promote Helm charts to GA


pebrc:
We currently have a beta label on the doc pages for the Helm charts. I think by now we are confident that they can be used in production environments given that we have been using them for all CI deployments for ~8 months and users have been relying on them for a while as well.

PR #5588, opened 2022-04-19
Remove "experimental" from Helm chart description

pebrc:
We missed this in #4796 there was another a textual reference to "experimental" beyond the warning.


barkbay (approved):

thbkrkr:
@pebrc Let's backport this to `2.2` as well, thanks.

PR #5596, opened 2022-04-20
Fix resources limits conversion in ToInt64() used for logging

thbkrkr:
Fixed a small bug with no real impact because the method is only used for logging.

I also renamed the `resource` variable to `res` to avoid the collision with the imported package `k8s.io/apimachinery/pkg/api/resource`.


barkbay (approved):
LGTM, thanks !

PR #5599, opened 2022-04-21
GKE: use default OS image type

### LINKED ISSUE #5466: GKE: migrate to cos_containerd image type
Starting with GKE 1.23 it is not possible to create K8S clusters using Docker as the container runtime ([see node images documentation here](https://cloud.google.com/kubernetes-engine/docs/concepts/node-images)).
We must update the deployer code to migrate from the old `cos` image that still relies on Docker: https://github.com/elastic/cloud-on-k8s/blob/1a6c22a58ec29f8624c39a162e5a3409a0edcf3e/hack/deployer/runner/gke.go#L176 to the new `cos_containerd`, which is the default, the PR would simply look like:

```diff
diff --git a/hack/deployer/runner/gke.go b/hack/deployer/runner/gke.go
index dab31756e..91ac29208 100644
--- a/hack/deployer/runner/gke.go
+++ b/hack/deployer/runner/gke.go
@@ -173,7 +173,7 @@ func (d *GkeDriver) create() error {
 
        return NewCommand(`gcloud beta container --quiet --project {{.GCloudProject}} clusters create {{.ClusterName}} ` +
                `--region {{.Region}} --no-enable-basic-auth --cluster-version {{.KubernetesVersion}} ` +
-               `--machine-type {{.MachineType}} --image-type COS --disk-type pd-ssd --disk-size 30 ` +
+               `--machine-type {{.MachineType}} --disk-type pd-ssd --disk-size 30 ` +
                `--local-ssd-count {{.LocalSsdCount}} --scopes {{.GcpScopes}} --num-nodes {{.NodeCountPerZone}} ` +
                `--enable-stackdriver-kubernetes --addons HorizontalPodAutoscaling,HttpLoadBalancing ` +
                `--no-enable-autoupgrade --no-enable-autorepair --enable-ip-alias --metadata disable-legacy-endpoints=true ` +
```

However, we should also take care of the [beats](https://github.com/elastic/cloud-on-k8s/tree/main/config/recipes/beats) and [agents](https://github.com/elastic/cloud-on-k8s/tree/main/config/recipes/elastic-agent) recipes to match the changes on the underlying host, for example:

* `/var/lib/docker/containers` does not exist anymore
* CRI endpoint is `/run/containerd/containerd.sock` (even if ` /var/run/docker.sock` is still there)

This would require to validate our recipes and run all the related E2E tests.

barkbay:
Fixes #5466

### DIFF
--- a/hack/deployer/runner/gke.go
+++ b/hack/deployer/runner/gke.go
@@ -175,7 +175,7 @@ func (d *GKEDriver) create() error {
 
 	return exec.NewCommand(`gcloud beta container --quiet --project {{.GCloudProject}} clusters create {{.ClusterName}} ` +
 		`--region {{.Region}} --no-enable-basic-auth --cluster-version {{.KubernetesVersion}} ` +
-		`--machine-type {{.MachineType}} --image-type COS --disk-type pd-ssd --disk-size 30 ` +
+		`--machine-type {{.MachineType}} --disk-type pd-ssd --disk-size 30 ` +
 		`--local-ssd-count {{.LocalSsdCount}} --scopes {{.GcpScopes}} --num-nodes {{.NodeCountPerZone}} ` +
 		`--enable-stackdriver-kubernetes --addons HorizontalPodAutoscaling,HttpLoadBalancing ` +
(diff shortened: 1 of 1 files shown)


barkbay:
run/e2e-tests tags=beat

thbkrkr (approved):

Issue #3056, opened 2020-05-13
Allow providing cleartext passwords for creating ES users


joscha-alisch:
## Proposal
Currently, to create users via the operator, you can enhance the file realm of ES by providing secrets files with the username and the salted and hashed password just as ES uses it. 

```yaml
kind: Secret
apiVersion: v1
metadata:
  name: my-filerealm-secret
stringData:
  users: |-
    rdeniro:$2a$10$BBJ/ILiyJ1eBTYoRKxkqbuDEdYECplvxnqQ47uiowE7yGqvCEgj9W
    alpacino:$2a$10$cNwHnElYiMYZ/T3K4PvzGeJ1KbpXZp2PfoQD.gfaVdImnHOwIuBKS
    jacknich:{PBKDF2}50000$z1CLJt0MEFjkIK5iEfgvfnA6xq7lF25uasspsTKSo5Q=$XxCVLbaKDimOdyWgLCLJiyoiWpA/XDMe/xtVgn1r5Sg=
```

This is very nice as a first step, but now every user of the operator needs to go through the hassle of running `elasticsearch-users` as described in the [users and roles docs](https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-users-and-roles.html). 

It would be great to be able to specify the user passwords as cleartext and have the operator handle the hashing as described by @sebgl here: https://github.com/elastic/cloud-on-k8s/issues/728#issuecomment-582810869

They provide an example of how this could look, which I think would work fine
```yaml
kind: Secret
apiVersion: v1
metadata:
  name: es-users-file-realm
stringData:
  # users_cleartext will be internally derived into 'users' by ECK, with a hash of the provided passwords
  # (the current secret is kept unmodified by ECK)
  users_cleartext: |-
    username1:cleartextpassword
```

**Use case. Why is this important?**
At the moment there are additional steps in our pipeline we need to maintain, be it the hashing of the passwords or creating users via the API. Moving this logic into the operator would allow us to simplify our setup.

krjackso:
I ran into this recently as well - we are using Vault as the source of truth for our secrets, and sync them to kubernetes secrets. We would like to be able to create ES users with plaintext passwords so we are not required to double-store the values as both plaintext (for injecting into filebeat, signalfx, other auxillary services), and encoded (for the fileRealm usage). Ideally the passwords could be specified as just a single property in a secret, not using stringData, so that we can use existing vault-secret syncing solutions going forward. 

[...]

PR #5613, opened 2022-04-25
Allow providing cleartext passwords for creating ES users

pebrc:
Fixes https://github.com/elastic/cloud-on-k8s/issues/3056

This PR adds support for an additional form of secret following the https://kubernetes.io/docs/concepts/configuration/secret/#basic-authentication-secret format. It will allow users to specify username and a cleartext password. 

I also added support for the `users_roles` file in that secret. The idea being that a user that wants to create a user and assign a role to it does not need to create two separate secrets for that (one for the user and one for the `users_role` file). The counterargument is of course that we are now mixing Elasticsearch file realm content with the basic authentication secret which might be confusing in a different way. 

We always supported setting a custom password for the `elastic` user through the file realm (because user defined ES users always take precedence) This PR also removes the default `elastic` user secret if the `elastic` user is set up through the file realm. The rationale for this decision was to avoid having to sync the contents of both secrets (the user provided one and the operator created one) or have the contents diverge which might lead to confusion. Also we should not not create resources in the API server that are not needed.


docs/orchestrating-elastic-stack-applications/security/users-and-roles.asciidoc
```diff
@@ -55,7 +56,30 @@ spec:
...
+type: kubernetes.io/basic-auth
+stringData:
+  username: rdeniro    # required field for kubernetes.io/basic-auth
+  password: mypassword # required field for kubernetes.io/basic-auth
+  users_roles: |-
+    admin:rdeniro  # optional, not part of kubernetes.io/basic-auth
+----
+
+NOTE: If you specify the password for the `elastic` user through such a basic authentication secret then the secret holding the password described in <<{p}-default-elastic-user>> will not be created by the operator.
+
```
thbkrkr, line 81: Maybe add a note that it can take some time (~1min) befire it's available?

pkg/controller/elasticsearch/user/user_provided.go
```diff
@@ -141,6 +148,37 @@ func retrieveUserProvidedFileRealm(c k8s.Client, es esv1.Elasticsearch, recorder
...
 }
 
+func realmFromBasicAuthSecret(secret corev1.Secret, existing filerealm.Realm) (filerealm.Realm, error) {
+	realm := filerealm.New()
+	nsn := k8s.ExtractNamespacedName(&secret)
+	username := k8s.GetSecretEntry(secret, corev1.BasicAuthUsernameKey)
+	if username == nil {
+		return realm, fmt.Errorf("username is required but was empty: %v", nsn)
+	}
+	password := k8s.GetSecretEntry(secret, corev1.BasicAuthPasswordKey)
```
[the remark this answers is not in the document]
pebrc: Good question. The validation would asynchronous so maybe  a bit hard to discover, but I can look into that.

docs/orchestrating-elastic-stack-applications/security/users-and-roles.asciidoc
pebrc: Is that not true for everything we orchestrate with ECK? Everything is asynchronous and available after a variable delay.

pkg/controller/elasticsearch/user/user_provided.go
```diff
@@ -141,6 +148,41 @@ func retrieveUserProvidedFileRealm(c k8s.Client, es esv1.Elasticsearch, recorder
...
+	if password == nil {
+		return realm, fmt.Errorf("password is required but was empty: %v", nsn)
+	}
+	const minPasswordLength = 6
+	if len(password) < minPasswordLength {
```
barkbay, line 163: Not sure if we want to go that far but Elasticsearch `username` validation function is [here](https://github.com/elastic/elasticsearch/blob/ce9819fa6cc1aed8d52a09ec28503487e25bb6f8/x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/support/Validation.java#L135-L159)

pkg/controller/elasticsearch/user/user_provided.go
```diff
@@ -151,5 +193,5 @@ func handleSecretNotFound(recorder record.EventRecorder, es esv1.Elasticsearch,
 func handleInvalidSecretData(recorder record.EventRecorder, es esv1.Elasticsearch, secretName string, err error) {
 	msg := "invalid data in secret"
 	log.Error(err, msg, "namespace", es.Namespace, "es_name", es.Name, "secret_name", secretName)
-	recorder.Event(&es, corev1.EventTypeWarning, events.EventReasonUnexpected, msg+": "+secretName)
+	recorder.Event(&es, corev1.EventTypeWarning, events.EventReasonUnexpected, fmt.Sprintf("%s (%s): %s/%s", msg, err.Error(), es.Namespace, secretName))
```
barkbay, line 196:
```suggestion
	recorder.Event(&es, corev1.EventTypeWarning, events.EventReasonUnexpected, fmt.Sprintf("%s %s/%s: %s", msg, es.Namespace, secretName, err.Error()))
```

pkg/controller/elasticsearch/user/user_provided.go
```diff
@@ -131,7 +132,13 @@ func retrieveUserProvidedFileRealm(c k8s.Client, es esv1.Elasticsearch, recorder
 			}
 			return filerealm.Realm{}, err
 		}
-		realm, err := filerealm.FromSecret(secret)
+		var realm filerealm.Realm
+		var err error
+		if k8s.HasSecretEntries(secret, corev1.BasicAuthPasswordKey, corev1.BasicAuthUsernameKey) {
```
barkbay, line 137: I wonder if using only the keys to decide how the `Secret` should be processed may not lead to situations which could be hard to understand. For example if there's a typo in a key: `kubectl create secret generic my-elastic2-user --from-literal=username=elastic --from-literal=pasword=alongpassword --from-literal=users_roles=role1:my-elastic2-user`, it does not generate any error.

On another hand I don't have any good alternative 🤷 (maybe generate an event if there is an unexpected key ?)

barkbay (commented):
I left a few comments, I'm still doing some tests, looks good so far 👍

pkg/controller/elasticsearch/user/user_provided.go
pebrc: I can look into that.

docs/orchestrating-elastic-stack-applications/security/users-and-roles.asciidoc
```diff
@@ -55,7 +56,30 @@ spec:
...
+apiVersion: v1
+kind: Secret
+metadata:
+  name: secret-basic-auth
+type: kubernetes.io/basic-auth
+stringData:
+  username: rdeniro    # required field for kubernetes.io/basic-auth
+  password: mypassword # required field for kubernetes.io/basic-auth
+  users_roles: |-      # optional, not part of kubernetes.io/basic-auth
+    admin:rdeniro
```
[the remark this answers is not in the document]
pebrc: I think I am coming around to your point of view here. `users_roles` would also allow the human user to define roles for other Elasticsearch users not defined in the secret which is extra confusing. I am going to change it to `roles` and update the documentation.

pkg/controller/elasticsearch/user/user_provided_test.go
```diff
@@ -241,3 +244,117 @@ func TestReconcileUserProvidedRoles(t *testing.T) {
...
+			wantErr: false,
+		},
+		{
+			name: "creates new password hash in absence of existing file realm",
+			args: args{
+				secret: basicAuthSecretFixture,
+			},
+			wantPassword: "my-user-pass",
+			wantErr:      false,
+		},
```
barkbay, line 339:
```suggestion
		},
		{
			name: "Generate new hash without error if current one is invalid",
			args: args{
				secret: basicAuthSecretFixture,
				existing: filerealm.New().
					WithUser(testUser, []byte("$2a$10$invalidhash.invalidhash")),
			},
			wantPassword: "my-user-pass",
			wantErr:      false,
		},
```

barkbay (approved):
LGTM 👍

test/e2e/es/auth_test.go
```diff
@@ -164,6 +166,30 @@ func TestESUserProvidedAuth(t *testing.T) {
...
+					}
+					return k.Client.Update(context.Background(), &existingSecret)
+				}),
+			},
+			test.Step{
+				Name: "ES API should eventually be accessible using the updated password and the updated role",
+				Test: test.Eventually(func() error {
+					esUser := esclient.BasicAuth{Name: sampleUser, Password: samplePasswordCleartext}
+					expectedStatusCode := 201
+					return postDocument(b.Elasticsearch, k, esUser, writeIndexUpdated, expectedStatusCode)
```
barkbay, line 190: 🤦 Sorry, I didn't notice that the test actually failed:

```
=== RUN   TestESUserProvidedAuth/ES_API_should_eventually_be_accessible_using_the_updated_password_and_the_updated_role#01
Retries (5m0s timeout): ..........................
    step.go:43: 
        	Error Trace:	utils.go:88
        	            	unexpected status code 403
        	Test:       	TestESUserProvidedAuth/ES_API_should_eventually_be_accessible_using_the_updated_password_and_the_updated_role#01
```

Content of the error is:

```
action [indices:data/write/index] is unauthorized for user [myuser] with roles [], this action is granted by the index privileges [create_doc,create,index,write,all]
```

pebrc: My bad I should have run the e2e test on this PR. Thanks for catching this.

pebrc:
run/e2e-tests tags=es

PR #5618, opened 2022-04-25
Do not delete restart shutdowns before node rejoins cluster

### LINKED ISSUE #5617: Node shutdown records deleted too early on restarts
As part of the analysis of https://github.com/elastic/elasticsearch/issues/85922 it became apparent that the operator is deleting the node shutdown records immediately after the Pods slated for a restart have been deleted. That means the operator does not wait until the Pods have re-joined the cluster. 

This has the unintended effect that the re-allocation delay for the shards on the ES node/Pod is re-set to its default value and a much shorter reallocation delay of 1 minute by default now applies. 

Node shutdown records for restarts should only be deleted once the Pod/Es node has rejoined the cluster to avoid disruption on large Elasticsearch clusters with lots of data.

pebrc:
Fixes #5617 

This is a minimal fix for #5617. It predicates the deletion of a node shutdown record of type `RESTART` on the presence of the corresponding node in the cluster. Presence in the cluster is established via the cached node list maintained in the operator and refreshed on each reconciliation.  

Implementation considerations: This adds flexible predicates to clear. This way we can call it multiple times: once we have observed all expected deletions to clean up all shutdowns for the happy path where the node has rejoined the cluster. We then call it a second time behind all the conditional logic that we also apply to the "should we re-enable shard allocations" question in the pre-shutdown world. This second time is only predicated on completed shutdowns so once we reach a stable state we also delete shutdowns for nodes that did not rejoin the cluster. This would be the case for example if a node lost its state during restart du to an external event e.g. a user manipulating the node's data directory or a hardware failure. It is debatable whether this optimisation is necessary as we could simply only delete shutdown once we reach a stable state (which is btw what happens during scale downs). The only downside would be that shutdown records start piling up in the Elasticsearch cluster state. I could potentially be a problem on very large clusters with hundreds of nodes. The optimisation is more or less free for us so I went for it here. 

I did a fair amount of manual testing: 
* rolling upgrade where restarted node stays pending
* rolling upgrade where node state is lost during restart
* node set downscale + rolling upgrade
* node set upscale + rolling upgrade

I also wonder if we are missing a test step in our e2e tests: we should make sure that shards don't start moving during rolling upgrades. In practice however I would not be suprised if we will find it hard to ever observe a shard moving given the small amounts of data we use in the test clusters even with the current implementation before the changes proposed in this PR.

### DIFF
--- a/pkg/controller/elasticsearch/driver/esstate_test.go
+++ b/pkg/controller/elasticsearch/driver/esstate_test.go
@@ -44,7 +44,8 @@ type fakeESClient struct { //nolint:maligned
 	clusterRoutingAllocation             esclient.ClusterRoutingAllocation
 	GetClusterRoutingAllocationCallCount int
 
-	Shutdowns map[string]esclient.NodeShutdown
+	Shutdowns            map[string]esclient.NodeShutdown
+	DeleteShutdownCalled bool
 
 	health                      esclient.Health
 	GetClusterHealthCalledCount int
@@ -124,7 +125,8 @@ func (f *fakeESClient) GetShutdown(_ context.Context, nodeID *string) (esclient.
 	return esclient.ShutdownResponse{Nodes: ns}, nil
 }
 
-func (f *fakeESClient) DeleteShutdown(ctx context.Context, nodeID string) error {
+func (f *fakeESClient) DeleteShutdown(_ context.Context, _ string) error {
+	f.DeleteShutdownCalled = true
 	return nil
 }
 
--- a/pkg/controller/elasticsearch/driver/upgrade_test.go
+++ b/pkg/controller/elasticsearch/driver/upgrade_test.go
@@ -16,6 +16,13 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 
 	esv1 "github.com/elastic/cloud-on-k8s/pkg/apis/elasticsearch/v1"
+	"github.com/elastic/cloud-on-k8s/pkg/controller/common/expectations"
+	"github.com/elastic/cloud-on-k8s/pkg/controller/common/reconciler"
+	"github.com/elastic/cloud-on-k8s/pkg/controller/common/version"
+	esclient "github.com/elastic/cloud-on-k8s/pkg/controller/elasticsearch/client"
+	"github.com/elastic/cloud-on-k8s/pkg/controller/elasticsearch/hints"
+	"github.com/elastic/cloud-on-k8s/pkg/controller/elasticsearch/reconcile"
+	"github.com/elastic/cloud-on-k8s/pkg/controller/elasticsearch/shutdown"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/elasticsearch/sset"
 	"github.com/elastic/cloud-on-k8s/pkg/utils/k8s"
 )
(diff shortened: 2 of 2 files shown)


pkg/controller/elasticsearch/shutdown/node.go
```diff
@@ -142,13 +151,17 @@ func logStatus(logger logr.Logger, podName string, shutdown esclient.NodeShutdow
 
 // Clear deletes shutdown requests matching the type of the NodeShutdown field typ and the given optional status.
 // Depending on the progress of the shutdown in question this means either a cancellation of the shutdown or a clean-up
-// after shutdown completion.
+// after shutdown completion. Restart shutdowns will only be deleted once the corresponding node is back in the cluster.
```
naemono, line 154:
```suggestion
// after shutdown completion. Restart shutdown records will only be deleted once the corresponding node is back in the cluster.
```

naemono (commented):
Going to also do some local testing as well.

pkg/controller/elasticsearch/shutdown/node.go
```diff
@@ -142,13 +151,17 @@ func logStatus(logger logr.Logger, podName string, shutdown esclient.NodeShutdow
...
 func (ns *NodeShutdown) Clear(ctx context.Context, status *esclient.ShutdownStatus) error {
 	if err := ns.initOnce(ctx); err != nil {
 		return err
 	}
 	for _, s := range ns.shutdowns {
 		if s.Is(ns.typ) && (status == nil || s.Status == *status) {
+			if ns.typ == esclient.Restart && !ns.nodeInCluster(s.NodeID) {
```
barkbay, line 161: Not related to your PR but I never realized that `Clear()` may never be called until all nodes are restarted: https://github.com/elastic/cloud-on-k8s/blob/757a07eb557b992a4d1645ce86eb8d5650907f5e/pkg/controller/elasticsearch/driver/upgrade.go#L111-L112
I managed to reproduce a situation where the 3 nodes of a cluster are restarted, but `Clear()` is never called during the upgrade, leading to a situation where restarted nodes are piling up in the cluster state:

```json
{
	"nodes": [{
		"node_id": "B_CBcBIzTlGnKrereheimw",
		"type": "RESTART",
		"reason": "610898",
		"allocation_delay": "5m",
		"shutdown_startedmillis": 1650956474657,
		"status": "COMPLETE",
		"shard_migration": {
			"status": "COMPLETE",
			"shard_migrations_remaining": 0,
			"explanation": "no shard relocation is necessary for a node restart"
		},
		"persistent_tasks": {
			"status": "COMPLETE"
		},
		"plugins": {
			"status": "COMPLETE"
		}
	}, {
		"node_id": "Htx8TgKwS5CE6Mt69SVcug",
		"type": "RESTART",
		"reason": "612171",
		"allocation_delay": "5m",
		"shutdown_startedmillis": 1650956680896,
		"status": "COMPLETE",
		"shard_migration": {
			"status": "COMPLETE",
			"shard_migrations_remaining": 0,
			"explanation": "no shard relocation is necessary for a node restart"
		},
		"persistent_tasks": {
			"status": "COMPLETE"
		},
		"plugins": {
			"status": "COMPLETE"
		}
	}, {
		"node_id": "w3R9ACQ8RseE9cEQDJf8bQ",
		"type": "RESTART",
		"reason": "611345",
		"allocation_delay": "5m",
		"shutdown_startedmillis": 1650956571498,
		"status": "COMPLETE",
		"shard_migration": {
			"status": "COMPLETE",
			"shard_migrations_remaining": 0,
			"explanation": "no shard relocation is necessary for a node restart"
		},
		"persistent_tasks": {
			"status": "COMPLETE"
		},
		"plugins": {
			"status": "COMPLETE"
		}
	}]
}
```

I'm wondering if we should consider it as a bug?

pebrc: I think potentially yes. It could become a problem if another restart is necessary before the old records have been cleared. I think keeping these records in the cluster state is probably less of a problem in terms of size. I feel like we should probably tackle this in a separate issue though and it probably has a little bit lower priority than this one wdyt?

pkg/controller/elasticsearch/shutdown/node.go
```diff
@@ -142,13 +151,17 @@ func logStatus(logger logr.Logger, podName string, shutdown esclient.NodeShutdow
...
 // Depending on the progress of the shutdown in question this means either a cancellation of the shutdown or a clean-up
-// after shutdown completion.
+// after shutdown completion. Restart shutdown records will only be deleted once the corresponding node is back in the cluster.
 func (ns *NodeShutdown) Clear(ctx context.Context, status *esclient.ShutdownStatus) error {
 	if err := ns.initOnce(ctx); err != nil {
 		return err
 	}
 	for _, s := range ns.shutdowns {
 		if s.Is(ns.typ) && (status == nil || s.Status == *status) {
+			if ns.typ == esclient.Restart && !ns.nodeInCluster(s.NodeID) {
```
barkbay, line 161: Should we compare the current time with `allocation_delay` `+` `shutdown_startedmillis` in the `GET` answer in order to garbage collect shutdown requests for nodes that may never join again the cluster? (if node state is lost before the restart for example?)

pkg/controller/elasticsearch/shutdown/node.go
barkbay: > I feel like we should probably tackle this in a separate issue though and it probably has a little bit lower priority than this one wdyt?

Yes, absolutely 👍

pkg/controller/elasticsearch/shutdown/node.go
pebrc: This could be a problem if there is clocks skew between the operator (node) and the nodes the Elasticsearch Pods run on. But I agree in general that we are creating garbage. Maybe we could look at the expected nodes in the cluster instead and if the restarted node is no longer expected to be in cluster remove it? I have to do some experiments.

pebrc: I realise that going by the expected nodes is more complicated than I thought because we cannot map the nodeID back to a Pod name unless we create additional state either in the operator or in the k8s API to remember which nodeID maps to which Pod name. With the help of Elasticsearch we can only do this mapping for nodes in the cluster.

pebrc: https://github.com/elastic/cloud-on-k8s/pull/5618/commits/d098f7910d93e77653541d328a8f7096142b4c3a implements your proposal.

pebrc:
Ignore https://github.com/elastic/cloud-on-k8s/commit/d098f7910d93e77653541d328a8f7096142b4c3a it is based on a wrong assumption. I will have to attempt another solution.

pkg/controller/elasticsearch/shutdown/node.go
pebrc: I have come up with another solution and updated the PR description accordingly.

pebrc:
Jenkins test this please

>  Error: dirty local changes
> 19:33:09   M pkg/controller/webhook/cert.go

I have been trying for half an hour to reproduce this locally without luck.

pkg/controller/elasticsearch/shutdown/node.go
```diff
@@ -140,20 +149,38 @@ func logStatus(logger logr.Logger, podName string, shutdown esclient.NodeShutdow
 	}
 }
 
+// ClearCondition predicate that can be used to limit the Clear function.
+type ClearCondition func(s esclient.NodeShutdown) bool
```
barkbay, line 153:
```suggestion
type ClearCondition func(esclient.NodeShutdown) bool
```

barkbay (approved):
LGTM 👍

pebrc:
run/e2e-tests tags=es

Issue #5621, opened 2022-04-27
Public HTTP Secret may hold expired CA


barkbay:
The CA certificate in the public HTTP `Secret` is only updated when a new HTTP certificate is issued (i.e. when [shouldIssueNewHTTPCertificate](https://github.com/elastic/cloud-on-k8s/blob/721cccfd905f43c345c7957fbfe173ea3ebb6bbf/pkg/controller/common/certificates/http_reconcile.go#L194-L228) returns `true`), but not when a new CA certificate is available:

```golang
	// check if the existing cert should be re-issued
	if shouldIssueNewHTTPCertificate(owner, namer, tls, controllerSANs, secret, svcs, ca, rotationParam.RotateBefore) {
		log.Info(
			"Issuing new HTTP certificate",
			"namespace", secret.Namespace,
			"secret_name", secret.Name,
			"owner_namespace", owner.Namespace,
			"owner_name", owner.Name,
		)
		secretWasChanged = true
		// store certificate and signed certificate in a secret mounted into the pod
		secret.Data[CAFileName] = EncodePEMCert(ca.Cert.Raw)
		secret.Data[CertFileName] = EncodePEMCert(certData, ca.Cert.Raw)
	}
       // Here is the bug: what if ca.Cert must be updated while the HTTP certificate is still valid (`shouldIssueNewHTTPCertificate == false`) ?
```

This may lead to a situation where the CA certificate is updated in `*-es-http-ca-internal`, but not propagated to `*-es-http-certs-public`. It may prevent associated applications, for example Kibana, to connect to Elasticsearch:

```
[ERROR][elasticsearch-service] Unable to retrieve version information from Elasticsearch nodes. certificate has expired
```

PR #5622, opened 2022-04-27
Ensure CA is always updated in HTTP Secret

barkbay:
Fix https://github.com/elastic/cloud-on-k8s/issues/5621

### DIFF
--- a/pkg/controller/common/certificates/http_reconcile.go
+++ b/pkg/controller/common/certificates/http_reconcile.go
@@ -191,7 +191,8 @@ func ensureInternalSelfSignedCertificateSecretContents(
 	}
 
 	// check if the existing cert should be re-issued
-	if shouldIssueNewHTTPCertificate(owner, namer, tls, controllerSANs, secret, svcs, ca, rotationParam.RotateBefore) {
+	certificate := getHTTPCertificate(owner, namer, tls, controllerSANs, secret, svcs, ca, rotationParam.RotateBefore)
+	if certificate == nil {
 		log.Info(
 			"Issuing new HTTP certificate",
 			"namespace", secret.Namespace,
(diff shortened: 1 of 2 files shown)


pkg/controller/common/certificates/http_reconcile.go
```diff
@@ -216,29 +217,45 @@ func ensureInternalSelfSignedCertificateSecretContents(
...
+		certificate, err = ca.CreateCertificate(*validatedCertificateTemplate)
 		if err != nil {
 			return secretWasChanged, err
 		}
 
 		secretWasChanged = true
 		// store certificate and signed certificate in a secret mounted into the pod
 		secret.Data[CAFileName] = EncodePEMCert(ca.Cert.Raw)
-		secret.Data[CertFileName] = EncodePEMCert(certData, ca.Cert.Raw)
+		secret.Data[CertFileName] = EncodePEMCert(certificate, ca.Cert.Raw)
```
pebrc, line 228: I wonder if we should avoid the redundancy of calling `EncodePEMCert` twice for both the CA and the certificate here and on L232-233

pebrc (approved):

pkg/controller/common/certificates/http_reconcile.go
barkbay: The idea was to have 2 different log messages according to what is generated/updated:
1. A new certificate is generated: "Issuing new HTTP certificate" is logged
2. No new certificate is generated, only the CA should be updated: "Updating CA certificate" is logged

I can remove the redundancy, and only call `EncodePEMCert` on L232-233, but it means that I should remove the "Updating CA certificate" message since it is no more accurate (it would also be logged if the certificate has been updated, not only the CA).

pebrc: I see the dilemma. I think the log message is valuable. I am sure we could fine a way but it would probably be harder to read so let's not make this micro optimisation for now.

PR #5439, opened 2022-03-04
Introduce a new custom build job

pebrc:
It allows specifying which stack images/versions should be used for each Elastic Stack application supported by ECK with the exception of Beats and APM Server. 

It is currently fixed to the operator main branch but we could make that configurable too if we believe that would be of benefit. 

I am open to suggestions how to solve the parameterisation on the Jenkins level in a better way. I am not a Jenkins expert. 

Fixes https://github.com/elastic/cloud-on-k8s/issues/5301

PR #5639, opened 2022-05-04
E2E avoid implicitly resetting configured stack version for Fleet

pebrc:
This aims to fix recent test failures of `TestAgentVersionUpgradeToLatest8x`. The cause of the failure is unfortunate interaction between two recently made tweaks to the e2e runner:
* https://github.com/elastic/cloud-on-k8s/pull/5439
* https://github.com/elastic/cloud-on-k8s/pull/5555

#5555 flips the version under test if the latest released version is lower than the version under test. e.g. if we test `8.2.0-SNAPSHOT` and the latest version release was (until yesterday) `8.1.1` then `8.1.1` would be used as the start version in the test and `8.2.0-SNAPSHOT` as the target version we try to upgrade to. 

#5439 however added some implicit version switching logic that tries to take the version under test from a config file. Now if that file does not exist as it is the case most of the time it should fall back to the version and image configured globally in the test context. Unfortunately when applying this fallback it reset the flipped version from #5555 

The fix attempts to respect any version set in the builder unless said configuration file exists with specific version overrides.

### DIFF
--- a/test/e2e/test/agent/builder.go
+++ b/test/e2e/test/agent/builder.go
@@ -305,9 +305,12 @@ func (b Builder) WithFleetServer() Builder {
 }
 
 func (b Builder) WithFleetImage() Builder {
-	def := test.Ctx().ImageDefinitionFor(FleetServerPseudoKind)
-	b.Agent.Spec.Image = def.Image
-	b.Agent.Spec.Version = def.Version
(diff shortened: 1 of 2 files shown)


barkbay (approved):
LGTM

PR #5641, opened 2022-05-05
Update test jobs to use 8.2.0

pebrc:
Updates the stack versions test job to test the 8.2.0 release. Drops 7.10 from the list of tested versions and starts testing 8.3.0-SNAPSHOT.

Dropping 7.10 to keep the overall number of versions under test bounded.


thbkrkr:
There are also patch versions that we can upgrade: from 7.17.1 to 7.17.3 and from 8.1.0 to 8.1.3.

thbkrkr (approved):
LGTM

PR #5642, opened 2022-05-05
Refactor upgrade completion logic

### LINKED ISSUE #5632: Revisit node shutdown clean up and improve testing
#5629 implemented a minimal fix to address the issue with `auto_expand_replicas` and node shutdown. 

We should look into improving the testing around that e.g through a continuous check during mutation tests that ensure that at most as many shutdowns exist as the change budget allows. 

We could also improve the code a bit and for example move all clean up to the top of the upgrade function.

pebrc:
Fixes #5632 

Modest refactoring to keep all the upgrade completion logic in one function and just move it to the top.

Adds a watcher for node shutdowns to the e2e tests that do rolling upgrades to ensure we only have as many concurred node shutdowns as the change budget allows nodes to be unavailable. 

Fixes a subtle bug with the watchers caused by closing over the loop iteration variable which meant that we were executing only the last watch function in the list of watchers despite logging the correct watchers names! Deserves a separate explanation somewhere once I find the time, took me a while to figure out.

### DIFF
--- a/test/e2e/test/elasticsearch/checks_restart.go
+++ b/test/e2e/test/elasticsearch/checks_restart.go
@@ -15,8 +15,8 @@ import (
 
 	esv1 "github.com/elastic/cloud-on-k8s/pkg/apis/elasticsearch/v1"
 	"github.com/elastic/cloud-on-k8s/pkg/controller/common/version"
-	"github.com/elastic/cloud-on-k8s/pkg/controller/elasticsearch/shutdown"
 	esclient "github.com/elastic/cloud-on-k8s/pkg/controller/elasticsearch/client"
+	"github.com/elastic/cloud-on-k8s/pkg/controller/elasticsearch/shutdown"
 	"github.com/elastic/cloud-on-k8s/test/e2e/test"
 )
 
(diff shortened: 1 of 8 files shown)


test/e2e/test/elasticsearch/checks_restart.go
```diff
@@ -0,0 +1,63 @@
...
+			}
+			if len(restarts) > maxConcurrentRestarts {
+				observedErrors = append(observedErrors, fmt.Errorf("expected %d, got %d, restarts: %v", maxConcurrentRestarts, len(restarts), restarts))
+			}
+		},
+		func(k *test.K8sClient, t *testing.T) { //nolint:thelper
+			require.Empty(t, observedErrors)
+		},
+		func() bool {
+			return version.MustParse(es.Spec.Version).LT(version.MinFor(7, 15, 2))
```
pebrc, line 60: I realise this should probably defined only once in code and reused here.

pebrc:
run/e2e-tests tags=es

thbkrkr (approved):
LGTM

PR #5643, opened 2022-05-05
Allow small number of network errors during load+mutation testing.

naemono:
Due to the history of `TestMutationWhileLoadTesting` test randomly failing, we're reviving the possibility of allowing a small number of network errors during mutation e2e tests.

Breakdown of allowances
* Only http error codes of `401` (unauthorized), and `0` (network error) are allowed.  Anything outside of this will fail.
* Only 5 or less network errors (code `0`) are allowed.  Anything > 5 will fail.

### DIFF
--- a/test/e2e/es/mutation_test.go
+++ b/test/e2e/es/mutation_test.go
@@ -14,14 +14,15 @@ import (
 	"testing"
 	"time"
 
-	esv1 "github.com/elastic/cloud-on-k8s/pkg/apis/elasticsearch/v1"
-	"github.com/elastic/cloud-on-k8s/pkg/dev/portforward"
-	"github.com/elastic/cloud-on-k8s/test/e2e/test"
-	"github.com/elastic/cloud-on-k8s/test/e2e/test/elasticsearch"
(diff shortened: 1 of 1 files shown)


test/e2e/es/mutation_test.go
```diff
@@ -292,9 +293,24 @@ func TestMutationWhileLoadTesting(t *testing.T) {
...
-			assert.Equal(t, 1, len(metrics.StatusCodes), msgAndArgs)
-			if _, ok := metrics.StatusCodes["401"]; !ok {
-				assert.Fail(t, "all status codes should be 401", msgAndArgs)
+			switch len(metrics.StatusCodes) {
+			case 1:
+				if _, ok := metrics.StatusCodes["401"]; !ok {
+					assert.Fail(t, "all status codes should be 401", msgAndArgs)
+				}
+			case 2:
+				// allow 5 or less network errors during load testing, as we randomly see these during testing.
```
pebrc, line 302: Let's maybe create an issue to investigate the reason for these failures and link to it here, so that we don't accept this behaviour as given.

pebrc (approved):

PR #5644, opened 2022-05-05
Configurable allocation delay

### LINKED ISSUE #5582: Make shutdown API allocation_delay configurable and review default value
## Bug Report
After upgrading the operator to version 2+, we started noticing rolling upgrade procedures being much slower. Upon looking more into it, the utilization of the shutdown API instead of the legacy transient settings was causing Elasticsearch to start reallocation of indices before the restarted nodes were rejoining the cluster, thus taking a very long time to heal after each node restart.

This seems to be caused by the fact that the operator now relies on a timeout rather than waiting for the pod to be back up and manually restart allocations and therefore, if that timeout expires before the node rejoins the cluster, the shards start getting reallocated.

I believe all this is the expected behavior of the shutdown API, however, I think there might be a misalignment in terms of the default timeout. The shutdown API allows to specify a custom timeout (allocation_delay) when performing the shutdown call, but that doesn't seem to get used by the operator, which instead relies on the default allocation_delay of the indices as seen by the comment [here](https://github.com/elastic/cloud-on-k8s/blob/2.1.0/pkg/controller/elasticsearch/shutdown/node.go#L91). The comment seems to suggest the operator should rely on the default timeout of 5m, however, Elasticsearch documentation reports the default timeout to be 1m (see [here](https://www.elastic.co/guide/en/elasticsearch/reference/8.1/delayed-allocation.html).

1 minute definitely seems to be a very short timeout in Kubernetes where if anything in the lifecycle of the pod (termination, unmounting of volumes, scheduling of the new pod, image pulling, mounting of volumes, initcontainers....) takes a little too long to execute, there is a high risk of the node not rejoining the cluster in time (which is what has happened to us for multiple nodes).

To further confirm this is the issue, I have changed every index in the cluster to have a higher delayed_timeout (10m) and ran another upgrade and this time everything completed successfully without any delays. This solution, however, is very tedious because delayed_timeout is a per index setting (so it needs to be applied to every index and needs to be configured on every index template to ensure new indices inherit it). Furthermore, this setting is meant more for unplanned issues where possibly the 1m default response time makes sense, while it might make sense to have a higher one for planned maintenances if it is known that the maintenance could take longer.

My proposal is to set a custom allocation_delay in the shutdown call to override the default one and set a sane default of at least 5m in that call. Furthermore, possibly make that value configurable on the operator level so that, for clusters with particular needs that might expect an even higher timeout to be needed, it is possible to override it and set a proper timeout that matches the needs of that cluster.

pebrc:
Fixes #5582 

Adds a new `restart` section to the upgrade strategy to allow configuring the `allocation_delay` for node restarts. 

```yaml
kind: Elasticsearch
metadata:
  name: shutdown-test
spec:
  version: 8.1.0 
  updateStrategy:
    restart:
      allocationDelay: 10m
```

I am open to making this an annotation if we think tweaking this value is too much of an edge case to warrant such an API change. 

I opted for not trying to parse the Elasticsearch time unit values but relying on the user providing the correct string. This of course opens the door for misconfiguration that will be harder to detect because it will be Elasticsearch that will return an error on `PUT` node shutdown which is only visible in operator logs. By introducing user controlled input here we make the orchestration process susceptible to unexpected failures. 

The alternative would be to 
* try to parse Elasticsearch time-units and give immediate feedback through the validating web hook. 
* use a fixed time unit like `allocationDelaySeconds`which would be in line with what the Kubernetes API does in various places we can then reliably marshal to Elasticsearch time units.


pebrc:
Putting this back in draft as we have to do some more research on the right approach here: e.g. is a global parameter enough, should it be calculated dynamically etc.

pebrc:
The intention behind `allocation_delay` is not so much to account for slow startup of Elasticsearch node and recovery from a local copy of the shards on a potentially slow disk. Those recoveries are running async anyway and as long as the node reports it has a good copy of a shard no re-allocation will happen. The idea behind `allocation_delay` is rather to help with slow restarts from an orchestrator point of view. I.e. the time between actual shutdown of the node until it is able to rejoin the cluster again. 

What that means is that we do not have to include ES node size or number of shards on the node into the calculation. It probably also means that a global parameter as proposed in this PR is probably good enough. There will still be differences between types of nodes, ones on faster hardware vs ones on slower but I expect them to be closer or at least close enough together.

pebrc:
I am going to park this PR for the time being. I don't feel there is an urgent enough need for this right now.

PR #5646, opened 2022-05-06
E2E: retry check cluster UUID check

pebrc:
The "Cluster UUID should have been preserved" check sometimes fails if the connection to Elasticsearch times out. While this might point to an underlying issue with the Elasticsearch cluster, this step is not intended to test Elasticsearch availability. This PR thus wraps the step in `test.Eventually` to allow retries.


barkbay (approved):
LGTM

Issue #5467, opened 2022-03-14
Integrate k8s client request tracking into our APM integration


pebrc:
The current APM integration in the operator does track requests to Elasticsearch but it does not track requests made to the Kubernetes API server. 

#5450 uncovered that we do too many unnecessary requests to the API server. Part of the problem why this went unnoticed is that gap in our observation of the operator. Adding this capability to the operator creates the necessary visibility that will allow us to set up automated alerts if the number of requests goes above a baseline for individual test cases we run (e.g. a ES cluster creation).

pebrc:
Closed with https://github.com/elastic/cloud-on-k8s/pull/5651

PR #5649, opened 2022-05-08
Do not use context.Background() on write requests to the k8s API

pebrc:
This is preparatory work for #5467 

It introduces the correct context into all requests made with the k8s client-go library that do not use the cache. So the rule is 
* `Create`, `Update`, `Delete` need a correct context 
* `List`, `Get` do not because the cached client dismisses the context argument anyway. I added one where it was in scope but did not do additional refactoring to introduce one where there was not context in scope.

The main motivation for the distinction between reads and writes is to limit scope of the refactoring  somewhat.

This will allow us in a next step to add tracing for k8s client requests (sneak preview of how that will look like): 
Screenshot 2022-05-09 at 17 13 56

## Note for reviewers

This is probably bit of a nightmare to review given the repetitive nature of the change and large change set. Things of interest to focus on could be:
* I made changes to the `asyncTasks` that are run from the [manager/main.go](https://github.com/elastic/cloud-on-k8s/pull/5649/files#diff-435eecb1d2af40ee96747f88ab71eb2758da989c92f76dcb1f714fc1b300c633) to introduce context and start transactions. This probably worth a closer look and also worth following into the individual tasks like the license info reporter and the telemetry reporter etc. 
* test code is probably less interesting as it just adds contexts where I changed function signatures


pebrc:
Jenkins test this please

cmd/manager/main.go
```diff
@@ -663,41 +663,49 @@ func readOptionalCA(caDir string) (*certificates.CA, error) {
...
 		}()
 	}
 
 	// Garbage collect orphaned secrets leftover from deleted resources while the operator was not running
 	// - association user secrets
-	garbageCollectUsers(cfg, managedNamespaces)
+	gcCtx := tracing.NewContextTransaction(ctx, tracer, "garbage-collection", "on_operator_start", nil)
+	err := garbageCollectUsers(gcCtx, cfg, managedNamespaces)
+	if err != nil {
+		log.Error(err, "user garbage collector failed")
```
thbkrkr, line 703: Nit: we are going to log twice `user garbage collector failed`:

```
  2022-05-17T10:09:35.427+0200    ERROR   manager user garbage collector failed {
    "service.version": "2.3.0-SNAPSHOT+6b671559",
    "error": "user garbage collector creation failed: root_error_message"
  }
```

```suggestion
		log.Error(err)
```

thbkrkr (approved):
LGTM

Issue #5528, opened 2022-03-28
Call desired nodes API upon desired state changes


henningandersen:
## Proposal

We would like to have ECK call the Elasticsearch [desired nodes API](https://www.elastic.co/guide/en/elasticsearch/reference/8.1/update-desired-nodes.html) whenever a new desired state is available, before carrying out steps towards that new desired state.

This will enable Elasticsearch to make better decisions (like avoiding unnecessary and expensive shard movements), improving the user experience, safety, avoiding data loss and speeding up reaching the desired state.

henningandersen:
Thanks @barkbay for taking this on. I will try to respond to your questions below:

> My hunch is that we should continue whatever the error is

I'd say not to continue if it fails any validation etc. Invoking the desired nodes API is done at URGENT priority and bypasses circuit breakers in order to ensure it succeeds. If it cannot store its state, it is likely that a node will not be able to join either.

I think we could live with trying hard (with some retries) and then continuing if it fails. I'd hope it would be reconciled eventually then though? Also, it may leave the cluster in a less than ideal state (but likely not worse than today until we rely on this information more heavily).

> Would it however prevent new nodes from joining the cluster for example?

No, we consider that a health issue, not something we validate against.

>  would be tempted to use something that uniquely identifies the Elasticsearch resource, like the UID ... Upon a restore, even if the Elasticsearch resource is recreated with a same name, it would allow to start a new history

IIUC, this sounds exactly right. We should only switch history ID in extreme cases like restoring k8s data store (i.e., going backwards in version numbers).

> I was thinking about using the generation field, (generation is a metadata field incremented when the spec of a resource is changed). It should be fine as long as discontinuous versions are allowed?

Yes, I think generation is the right field to use here. There can be holes in the sequence, no problem. As long as it is always increasing when new changes happen.

> The resources allocated to a Pod may not be known to the operator

I believe we validate that cpu and memory is specified (@fcofdez keep me honest here). I think not having these could be problematic, at least to autoscaling and balancing. Is this a normal configuration or an atypical one?

> The actual storage capacity might be greater than the one requested ... Is it a problem if the capacity observed by Elasticsearch is greater than the one set in the desired API?

Yes, that could turn out problematic. For instance autoscaling might say that we need one more additional node until a new node from a previous autoscaling event actually joins.

fcofdez:
[...]

PR #5650, opened 2022-05-10
Add support for the desired nodes API

barkbay:
This PR is a draft implementation to add support for the [Elasticsearch desired API](https://www.elastic.co/guide/en/elasticsearch/reference/8.1/update-desired-nodes.html). I want to raise it now to allow anyone interested in this feature to try it out. There is an ongoing discussion regarding how the CPU capacity should be provided to Elasticsearch (refer to the [related ECK issue](https://github.com/elastic/cloud-on-k8s/issues/5528#issuecomment-1115145925) or this [Elasticsearch PR](https://github.com/elastic/elasticsearch/pull/86434)) which is not yet reflected in this draft. 

## Required resources

For the operator to be able to call the desired node API the following resources are expected in `nodeSets[*].podTemplate.spec.containers["elasticsearch"]`:

* memory: Both the `limit` and the `request` must be set to a same value.
* cpu:
  * `limit` is returned if set
  * if `limit` is not set then `request` is returned
* storage: Multiple storage data path is not supported. The storage path must be backed by a volume claim. The storage capacity is read from the `PersistentVolumeClaim` status, or from the `PersistentVolumeClaim` specification if it is not bound yet.

## What if resources are missing?

If the capacity of at least one resource (cpu,memory or storage) cannot be calculated the reconciliation is not stopped, the desired node API is however not called, it is also cleared. The Elasticsearch status subresource helps to understand why, here are a few examples:

```json
    {
      "lastTransitionTime": "2022-05-10T12:54:14Z",
      "message": "Cannot get compute and storage resources from Elasticsearch resource generation 2: cannot compute resources for NodeSet \"ml\": no CPU request or limit set",
      "status": "False",
      "type": "ResourcesAwareManagement"
    }
```

```json
    {
      "lastTransitionTime": "2022-05-10T12:59:29Z",
      "message": "Cannot get compute and storage resources from Elasticsearch resource generation 2: cannot compute resources for NodeSet \"hot\": Memory limit is set but value is 0",
      "status": "False",
      "type": "ResourcesAwareManagement"
    }
```

If the state of the desired node API can be calculated for a generation:

```json
    {
      "lastTransitionTime": "2022-05-10T13:03:44Z",
      "message": "Successfully calculated compute and storage resources from Elasticsearch resource generation 4",
      "status": "True",
      "type": "ResourcesAwareManagement"
    }
```

## Testing

Minimum Elasticsearch version to test this PR is `8.1.0`


pebrc (commented):
This looks really clean 👌. I left a few minor comments. In my experiments what I found confusing is the interaction between our resource defaults and resource limits/requests I set explicitly but that is nothing to do with your work here.

barkbay:
Commit [f511711](https://github.com/elastic/cloud-on-k8s/pull/5650/commits/f51171154a7d729bcfc62bbac60d149f3df008fe) relies on https://github.com/elastic/elasticsearch/pull/86434 to set the CPU resources, and makes Elasticsearch 8.3.0 the min. supported version.

pkg/controller/elasticsearch/driver/desired_nodes.go
```diff
@@ -0,0 +1,79 @@
...
+		// It is not possible to build the desired node spec because of the Elasticsearch specification
+		d.ReconcileState.ReportCondition(
+			esv1.ResourcesAwareManagement,
+			corev1.ConditionFalse,
+			fmt.Sprintf("Cannot get compute and storage resources from Elasticsearch resource generation %d: %s", d.ES.Generation, err.Error()),
+		)
+		// It is fine to continue, error is only reported through the condition.
+		// We should however clear the desired nodes API since we are a degraded (not resources aware) mode.
+		if esReachable {
+			return results.WithError(esClient.DeleteDesiredNodes(ctx))
```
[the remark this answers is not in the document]
barkbay: > Is there an optimisation to say we only delete if there are desired nodes to begin with? 

This is a very good question as I think it actually raises another one: should we always call `DELETE /_internal/desired_nodes` whatever is the current history ID?

@fcofdez May I have your thoughts on this suggestion? My understanding is that the only way to get the current desired nodes state is by calling `/_internal/desired_nodes/_latest`. However, what if the `_latest` stored state is related to a previous history/instance of the cluster? Said differently, what if the cluster has been restored and is now running with a new history? Calling `DELETE /_internal/desired_nodes` without being able to specify an history ID would also delete that previous state, which is maybe not what we want? (maybe it is, just want to be sure it makes sense from an Elasticsearch pov to delete all the histories at once)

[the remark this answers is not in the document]
fcofdez: > Is there an optimisation to say we only delete if there are desired nodes to begin with?

I think it's better to let Elasticsearch to handle that, it is almost a no-op when there are no desired nodes, so as long as we call that API occasionally it should be fine.

> This is a very good question as I think it actually raises another one: should we always call DELETE /_internal/desired_nodes whatever is the current history ID?

No, `DELETE /_internal/desired_nodes` is expected to be used when the orchestrator decides to move to the "unmanaged" mode, meaning that Elasticsearch would stop relying on the desired nodes information to make decisions.

> My understanding is that the only way to get the current desired nodes state is by calling /_internal/desired_nodes/_latest. 

You're correct, the only way to get the current desired nodes is by calling `GET /_internal/desired_nodes/_latest`.

> However, what if the _latest stored state is related to a previous history/instance of the cluster? Said differently, what if the cluster has been restored and is now running with a new history? Calling DELETE /_internal/desired_nodes without being able to specify an history ID would also delete that previous state, which is maybe not what we want? (maybe it is, just want to be sure it makes sense from an Elasticsearch pov to delete all the histories at once)

For the restore scenario that you're describing, if the cluster state is restored, the latest desired nodes might be a few versions behind from the latest desired nodes version that the operator is aware of. As long as the next update uses a version greater than the restored desired nodes, the update should proceed without any issues. In any case, Elasticsearch returns an error if the provided version is behind the latest stored version.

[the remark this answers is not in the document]
barkbay: > No, DELETE /_internal/desired_nodes is expected to be used when the orchestrator decides to move to the "unmanaged" mode, meaning that Elasticsearch would stop relying on the desired nodes information to make decisions.

Agreed. In ECK this will happen if the resources cannot be calculated (for example if no CPU has been provided). This however might be an involuntary situation, for example if the user makes a mistake, in that case all the "histories" will be deleted. We can revisit this mechanism once the API is more lenient regarding missing resources.

> I think it's better to let Elasticsearch to handle that, it is almost a no-op when there are no desired nodes, so as long as we call that API occasionally it should be fine.

👍 , thanks !

barkbay:
This PR is ready for review, however you'll need https://github.com/elastic/elasticsearch/pull/86434 to be merged in `8.3.0-SNAPSHOT` to test it or run the new E2E check. I'll post a comment here once it is the case.

fcofdez:
I've just merged https://github.com/elastic/elasticsearch/pull/86434. It should be available in the next `8.3.0-SNAPSHOT` build

barkbay:
Jenkins test this please

pkg/controller/elasticsearch/nodespec/desired_nodes.go
```diff
@@ -0,0 +1,313 @@
...
+				envPodName, nodeResource.nodeName,
+				envNamespace, sts.Namespace,
+				envHeadlessServiceName, resources.HeadlessService.Name,
+			)
+			var settings map[string]interface{}
+			if err := resources.Config.CanonicalConfig.Unpack(&settings); err != nil {
+				return nil, false, err
+			}
+			visit(nil, settings, func(s string) string {
+				return knownVariablesReplacer.Replace(s)
```
pebrc, line 122: I would like to better understand why we have to replace these variables for the desired nodes API. https://github.com/elastic/cloud-on-k8s/issues/5528#issuecomment-1118508705 it seems that you are worried some of the optimisations that ES does will be negatively affected if there are still variables in the settings. Can you maybe add a comment?

pkg/controller/elasticsearch/nodespec/desired_nodes.go
```diff
@@ -0,0 +1,313 @@
...
+		nodeResources[i].requeue = true
+		if storageInSpec, exists := pvc.Spec.Resources.Requests[corev1.ResourceStorage]; exists {
+			nodeResources[i].storage = storageInSpec.Value()
+		} else {
+			// PVC does exist, but Spec.Resources.Requests is empty, this is unlikely to happen for a PVC, fall back to claimed storage.
+			nodeResources[i].storage = *claimedStorage
+		}
+	}
+
+	return nodeResources, nil
```
pebrc, line 300: Nit: this last bit L269-300 could be the `Build()` method of your builder. But I guess it's fine given that this is just an internal abstraction and deals mostly with storage.

test/e2e/test/elasticsearch/checks_es.go
```diff
@@ -88,6 +91,96 @@ func (e *esClusterChecks) CheckESHealthGreen() test.Step {
...
+}
+
+// expectDesiredNodesAPI attempts to detect when the desired nodes state is expected to be set.
+func expectDesiredNodesAPI(es *esv1.Elasticsearch) bool {
+	if usesEmptyDir(*es) {
+		return false
+	}
+	for _, nodeSet := range es.Spec.NodeSets {
+		if nodeSet.Config != nil {
+			cfgPathData := nodeSet.Config.Data["path.data"]
```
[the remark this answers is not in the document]
barkbay: Good catch 👍 
Fixed in https://github.com/elastic/cloud-on-k8s/commit/a623f6fd543b0c5fe77c75dbd5043fccec5c8e4b

pkg/controller/elasticsearch/nodespec/desired_nodes.go
barkbay: I added a comment, the main motivation is to provide a value to the `node.name` parameter.

pkg/controller/elasticsearch/driver/desired_nodes_test.go
```diff
@@ -0,0 +1,809 @@
...
+			condition := d.ReconcileState.Index(esv1.ResourcesAwareManagement)
+			hasCondition := condition >= 0
+			assert.True(t, hasCondition, "ResourcesAwareManagement condition should be set")
+			if !hasCondition {
+				return
+			}
+			c := d.ReconcileState.Conditions[condition]
+			assert.Equal(t, tt.want.condition.status, c.Status)
+			for _, expectedMessage := range tt.want.condition.messages {
+				assert.True(t, strings.Contains(c.Message, expectedMessage), "expected message in condition: \"%s\", got \"%s\" ", expectedMessage, c.Message)
```
thbkrkr, line 531:
```suggestion
				assert.True(t, strings.Contains(c.Message, expectedMessage), "expected message in condition: %q, got %q", expectedMessage, c.Message)
```

thbkrkr (approved):
Left some very minor comments.
LGTM!

pkg/controller/elasticsearch/driver/desired_nodes_test.go
```diff
@@ -0,0 +1,809 @@
...
+						}),
+				),
+			want: want{
+				result:       wantResult{},
+				deleteCalled: true,
+				condition: &wantCondition{
+					status: corev1.ConditionFalse,
+					messages: []string{
+						"CPU request is set but value is 0",
+						"Memory limit is set but value is 0",
```
thbkrkr, line 396: Sorry, I missed this one that goes hand in hand.
```suggestion
						"memory limit is set but value is 0",
```

PR #5651, opened 2022-05-10
Add APM tracing for client-go requests to the Kubernetes API

pebrc:
Follow up to #5649 

Will rebase once the first PR is merged. Relevant commits is:

https://github.com/elastic/cloud-on-k8s/pull/5651/commits/15e7e872a24435c5c8a9df5be8e838146f5716c8

This adds a custom APM round tripper for requests to the Kubernetes API. This can potentially be contributed to [elastic/apm-agent-go ](https://github.com/elastic/apm-agent-go) as a new module. But I think it would be good to try it first in the context of the eck-operator. I structured the code in such a way to allow that if we decide it is worth contributing.

Why a custom round tripper? 

To get a meaningful span summary that is specific to k8s. The current implementation is not perfect in this regard especially sub-resource requests are represented somewhat arbitrarily by subsections of the request path. 

Screenshot 2022-05-10 at 16 01 22

The other thing implement here is to track requests done in the background by the client-go library to maintain the local cache. As far as I know there is no way to inject a context into the library that would enable tracing. This PR implements instead a default transaction mechanism to track those requests.

Screenshot 2022-05-10 at 16 00 48

### DIFF
--- a/.ci/Dockerfile
+++ b/.ci/Dockerfile
@@ -13,7 +13,7 @@ ENV EKSCTL_VERSION=0.74.0
 ENV ECK_DIAG_VERSION=1.0.3
 
 # golangci-lint
-RUN curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh \
+RUN curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh \
     | sh -s -- -b $(go env GOPATH)/bin v${GOLANGCILINT_VERSION}
 
 # kubebuilder to get required tools (etcd, apiserver, etc.)
--- a/config/crds/v1/all-crds.yaml
+++ b/config/crds/v1/all-crds.yaml
@@ -4063,10 +4063,9 @@ spec:
                         x-kubernetes-int-or-string: true
                       selector:
                         description: Label query over pods whose evictions are managed
-                          by the disruption budget. A null selector selects no pods.
-                          An empty selector ({}) also selects no pods, which differs
-                          from standard behavior of selecting all pods. In policy/v1,
-                          an empty selector will select all pods in the namespace.
+                          by the disruption budget. A null selector will match no
+                          pods, while an empty ({}) selector will select all pods
+                          within the namespace.
                         properties:
                           matchExpressions:
                             description: matchExpressions is a list of label selector
--- a/config/crds/v1/bases/elasticsearch.k8s.elastic.co_elasticsearches.yaml
+++ b/config/crds/v1/bases/elasticsearch.k8s.elastic.co_elasticsearches.yaml
@@ -8673,10 +8673,9 @@ spec:
                         x-kubernetes-int-or-string: true
                       selector:
                         description: Label query over pods whose evictions are managed
-                          by the disruption budget. A null selector selects no pods.
-                          An empty selector ({}) also selects no pods, which differs
-                          from standard behavior of selecting all pods. In policy/v1,
-                          an empty selector will select all pods in the namespace.
+                          by the disruption budget. A null selector will match no
+                          pods, while an empty ({}) selector will select all pods
+                          within the namespace.
                         properties:
                           matchExpressions:
                             description: matchExpressions is a list of label selector
--- a/deploy/eck-operator/charts/eck-operator-crds/templates/all-crds.yaml
+++ b/deploy/eck-operator/charts/eck-operator-crds/templates/all-crds.yaml
@@ -4093,10 +4093,9 @@ spec:
                         x-kubernetes-int-or-string: true
                       selector:
                         description: Label query over pods whose evictions are managed
-                          by the disruption budget. A null selector selects no pods.
-                          An empty selector ({}) also selects no pods, which differs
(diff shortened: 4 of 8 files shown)


pkg/controller/common/tracing/apmclientgo/client.go
```diff
@@ -0,0 +1,178 @@
...
+	b.WriteString(statement)
+	b.WriteRune(' ')
+	b.WriteString(path)
+	return b.String()
+}
+
+// ClientOption sets options for tracing client requests.
+type ClientOption func(*roundTripper)
+
+func WithDefaultTransaction(f func() *apm.Transaction) ClientOption {
```
barkbay, line 174: nit: maybe add a goc

pkg/controller/common/tracing/apmclientgo/client.go
```diff
@@ -0,0 +1,178 @@
...
+	r           http.RoundTripper
+	defaultTxFn func() *apm.Transaction
+}
+
+func (r roundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
+	ctx := req.Context()
+	tx := apm.TransactionFromContext(ctx)
+	if tx == nil && r.defaultTxFn != nil {
+		tx = r.defaultTxFn()
+		defer tx.End()
```
barkbay, line 45: It may panic here if there is no `tracer`:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x472950e]

goroutine 30 [running]:
go.elastic.co/apm/v2.(*Transaction).End(0x0)
        /Users/michael/go/pkg/mod/go.elastic.co/apm/v2@v2.0.0/transaction.go:277 +0x2e
github.com/elastic/cloud-on-k8s/pkg/controller/common/tracing/apmclientgo.roundTripper.RoundTrip({{0x5b1a720?, 0xc0002437c0?}, 0xc000449fd0?}, 0xc00060d400)
        /Users/michael/go/src/github.com/elastic/cloud-on-k8s/pkg/controller/common/tracing/apmclientgo/client.go:48 +0x8b2
golang.org/x/oauth2.(*Transport).RoundTrip(0xc000098840, 0xc00060d300)
[...]
```

Maybe we should set the wrapper only if `tracer` is non `nil`, for example:

```golang
	// set up APM  tracing if configured
	var tracer *apm.Tracer
	if viper.GetBool(operator.EnableTracingFlag) {
		tracer = tracing.NewTracer("elastic-operator")
		// set up APM tracing for client-go
		cfg.Wrap(tracing.ClientGoTransportWrapper(
			apmclientgo.WithDefaultTransaction(tracing.ClientGoCacheTx(tracer)),
		))
	}
```

barkbay (commented):
I'm still doing some tests, it works well so far 👍 , just wanted to leave a few comments.

I feel a bit out of depth because I'm not familiar with the `go.elastic.co/apm/v2` package, but it's a very interesting PR!

cmd/manager/main.go
```diff
@@ -457,9 +458,18 @@ func startOperator(ctx context.Context) error {
...
 
+	// set up APM  tracing if configured
+	var tracer *apm.Tracer
+	if viper.GetBool(operator.EnableTracingFlag) {
+		tracer = tracing.NewTracer("elastic-operator")
+	}
+	// set up APM tracing for client-go
+	cfg.WrapTransport = tracing.ClientGoTransportWrapper(
+		apmclientgo.WithDefaultTransaction(tracing.ClientGoCacheTx(tracer)),
+	)
```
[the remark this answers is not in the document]
pebrc: Nice! I missed that. So you can actually conveniently configure multiple wrappers this way.

pkg/controller/common/tracing/apmclientgo/client.go
```diff
@@ -0,0 +1,182 @@
...
+	if req.Method == "PUT" {
+		numSegments = 3
+	}
+
+	pathSegments := strings.Split(req.URL.Path, "/")
+	path := strings.Join(pathSegments[len(pathSegments)-numSegments:], "/")
+	// let's call out watch requests explicitly
+	if watch := req.URL.Query().Get("watch"); watch == "true" {
+		statement = "WATCH"
+	}
```
barkbay, line 164: Any reason to not just use `req.URL.Path`?

image

barkbay (approved):
LGTM

I'm not sure I'm the best person to review the `RoundTrip` function though, but given it seems inspired from the [APM http module](https://github.com/elastic/apm-agent-go/blob/f6b33856d836e91859d1919e5351495e6129adfe/module/apmhttp/client.go#L78-L136) I think it is fine (it might be an argument for having it as a new module as you suggested).

pkg/controller/common/tracing/apmclientgo/client.go
pebrc: I thought it got very long in the UI so I was looking for something shorter that still conveyed the most relevant information.

pebrc:
Jenkins test this please

>  /bin/bash: line 1: golangci-lint: command not found
I am confused

PR #5663, opened 2022-05-12
Allow non-IPs in service spec to avoid noop updates

### LINKED ISSUE #5657: Operator updates services on each reconciliation
While working on https://github.com/elastic/cloud-on-k8s/pull/5651 I noticed that we update the services created by the operator on every single reconciliation. 
Screenshot 2022-05-10 at 16 01 22

This seems to be caused by a diff in `ClusterIPs` where we ignore `ClusterIPs` when they are not a parseable IP as it is the case for `None`. 

I don't recall why we have this special casing of non-IPs, I suspect a validation in an older version of Kubernetes might have been at play as I don't think it is there without reason. In any case we should revisit this and see if we can avoid unnecessary updates. 

This needs to be tested across k8s versions probably to avoid any breakage.

pebrc:
Fixes #5657 

We are copying parts of the service spec populated on the service side into the `expected` service spec to avoid unnecessary updates of services. For some reason that is unfortunately not documented an exception was made for the `ClusterIP[s]` values that not valid IPs like `None`. 

By not including those we always update the service on each reconciliation.

Prior relevant work on in this area:
https://github.com/elastic/cloud-on-k8s/pull/2691
https://github.com/elastic/cloud-on-k8s/pull/4929

~I am still investigating~ It is unclear to me why we made this exclusion originally. I have not been able to create a situation where this became a problem. It seems `None` is perfectly acceptable to create a headless service, as expected.

### DIFF
--- a/pkg/controller/common/service_control.go
+++ b/pkg/controller/common/service_control.go
@@ -89,20 +89,17 @@ func applyServerSideValues(expected, reconciled *corev1.Service) {
 	if expected.Spec.Type == "" {
 		expected.Spec.Type = reconciled.Spec.Type
 	}
-	// ClusterIP might not exist in the expected service,
+	// ClusterIPs might not exist in the expected service,
 	// but might have been set after creation by k8s on the actual resource.
 	// In such case, we want to use these values for comparison.
 	// But only if we are not changing the type of service and the api server has assigned an IP
-	if expected.Spec.Type == reconciled.Spec.Type && expected.Spec.ClusterIP == "" {
-		expected.Spec.ClusterIP = reconciled.Spec.ClusterIP
-	}
-
-	// ClusterIPs also might not exist in the expected service,
-	// but might have been set after creation by k8s on the actual resource.
-	// In such case, we want to use these values for comparison.
-	// But only if we are not changing the type of service and the api server has assigned IPs
-	if expected.Spec.Type == reconciled.Spec.Type && len(expected.Spec.ClusterIPs) == 0 {
-		expected.Spec.ClusterIPs = reconciled.Spec.ClusterIPs
+	if expected.Spec.Type == reconciled.Spec.Type {
+		if expected.Spec.ClusterIP == "" {
+			expected.Spec.ClusterIP = reconciled.Spec.ClusterIP
+		}
+		if len(expected.Spec.ClusterIPs) == 0 {
+			expected.Spec.ClusterIPs = reconciled.Spec.ClusterIPs
+		}
 	}
 
 	// SessionAffinity may be defaulted by the api server


pkg/controller/common/service_control.go
```diff
@@ -142,15 +141,6 @@ func applyServerSideValues(expected, reconciled *corev1.Service) {
 	}
```
barkbay, line 141:
```suggestion
	}
	
	// InternalTrafficPolicy may be defaulted by the api server starting K8S v1.22
	if expected.Spec.InternalTrafficPolicy == nil {
		expected.Spec.InternalTrafficPolicy = reconciled.Spec.InternalTrafficPolicy
	}
```

barkbay (commented):
> It is unclear to me why we made this exclusion originally.

Same 🤷‍♂️ 

Doing some tests I noticed that  `InternalTrafficPolicy` is set by the api server starting K8S 1.22 ([documentation states that it is beta since v1.23](https://kubernetes.io/docs/concepts/services-networking/service-traffic-policy/) but I can clearly see it enabled by default on `v1.22.9-gke.1300`):

```diff
  v1.ServiceSpec{
        ... // 16 identical fields
        AllocateLoadBalancerNodePorts: nil,
        LoadBalancerClass:             nil,
-       InternalTrafficPolicy:         nil,
+       InternalTrafficPolicy:         &"Cluster",
  }
```

pkg/controller/common/service_control.go
pebrc: A good catch! I am running an older k8s version.

pkg/controller/common/service_control_test.go
```diff
@@ -285,7 +288,7 @@ func Test_applyServerSideValues(t *testing.T) {
 			}},
 		},
 		{
-			name: "Reconciled ClusterIP[s] is not used if the reconciled ClusterIP[s] are not valid IPs",
+			name: "Reconciled ClusterIP[s] is also used if the reconciled ClusterIP[s] are not valid IPs",
```
thbkrkr, line 291: Not sure what we are testing now with this case. It's equivalent to the previous case 'Reconciled ClusterIP/ClusterIPs/Type/SessionAffinity is used if expected ClusterIP/Type/SessionAffinity is empty'.

pkg/controller/common/service_control_test.go
```diff
@@ -515,6 +520,32 @@ func Test_applyServerSideValues(t *testing.T) {
 				IPFamilies:      []corev1.IPFamily{corev1.IPv6Protocol},
 			}},
 		},
+		{
+			name: "Reconciled InternalTrafficPolicy is used if empty",
```
thbkrkr, line 524: Nit:
```suggestion
			name: "Reconciled InternalTrafficPolicy is used if the expected one is empty",
```

pkg/controller/common/service_control_test.go
```diff
@@ -515,6 +520,32 @@ func Test_applyServerSideValues(t *testing.T) {
...
+				reconciled: corev1.Service{Spec: corev1.ServiceSpec{
+					InternalTrafficPolicy: ptr(corev1.ServiceInternalTrafficPolicyCluster),
+				}},
+			},
+			want: corev1.Service{Spec: corev1.ServiceSpec{
+				InternalTrafficPolicy: ptr(corev1.ServiceInternalTrafficPolicyCluster),
+			}},
+		},
+		{
+			name: "Provided InternalTrafficPolicy is used if not empty",
```
thbkrkr, line 536:
```suggestion
			name: "Expected InternalTrafficPolicy is used if not empty",
```

pkg/controller/common/service_control.go
```diff
@@ -94,15 +93,15 @@ func applyServerSideValues(expected, reconciled *corev1.Service) {
...
 
 	// ClusterIPs also might not exist in the expected service,
 	// but might have been set after creation by k8s on the actual resource.
 	// In such case, we want to use these values for comparison.
 	// But only if we are not changing the type of service and the api server has assigned IPs
-	if expected.Spec.Type == reconciled.Spec.Type && len(expected.Spec.ClusterIPs) == 0 && validClusterIPs(reconciled.Spec.ClusterIPs) {
+	if expected.Spec.Type == reconciled.Spec.Type && len(expected.Spec.ClusterIPs) == 0 {
 		expected.Spec.ClusterIPs = reconciled.Spec.ClusterIPs
 	}
```
thbkrkr, line 106: Nit: doesn't it make you want to combine the two?
```diff
diff --git a/pkg/controller/common/service_control.go b/pkg/controller/common/service_control.go
index 272ea82b0..f4ec0229e 100644
--- a/pkg/controller/common/service_control.go
+++ b/pkg/controller/common/service_control.go
@@ -89,20 +89,16 @@ func applyServerSideValues(expected, reconciled *corev1.Service) {
        if expected.Spec.Type == "" {
                expected.Spec.Type = reconciled.Spec.Type
        }
-       // ClusterIP might not exist in the expected service,
-       // but might have been set after creation by k8s on the actual resource.
-       // In such case, we want to use these values for comparison.
-       // But only if we are not changing the type of service and the api server has assigned an IP
-       if expected.Spec.Type == reconciled.Spec.Type && expected.Spec.ClusterIP == "" {
-               expected.Spec.ClusterIP = reconciled.Spec.ClusterIP
-       }
 
-       // ClusterIPs also might not exist in the expected service,
-       // but might have been set after creation by k8s on the actual resource.
-       // In such case, we want to use these values for comparison.
-       // But only if we are not changing the type of service and the api server has assigned IPs
-       if expected.Spec.Type == reconciled.Spec.Type && len(expected.Spec.ClusterIPs) == 0 {
-               expected.Spec.ClusterIPs = reconciled.Spec.ClusterIPs
+       // ClusterIP[s] might not exist in the expected service, but might have been set after creation by k8s on the actual resource.
+       // Use these values  for comparison, but only if we are not changing the type of service and the api server has assigned IP[s].
+       if expected.Spec.Type == reconciled.Spec.Type {
+               if expected.Spec.ClusterIP == "" {
+                       expected.Spec.ClusterIP = reconciled.Spec.ClusterIP
+               }
+               if len(expected.Spec.ClusterIPs) == 0 {
+                       expected.Spec.ClusterIPs = reconciled.Spec.ClusterIPs
+               }
        }
```

pebrc: Good point

thbkrkr (approved):
LGTM

PR #5665, opened 2022-05-12
Note to user when overwriting environment variables using .registry.env

naemono:
Attempting to push to docker registry using `make docker-push`, and having your environment variables for `DOCKER_LOGIN`, and `DOCKER_PASSWORD` overwritten with the contents of `.registry.env` file without warning is frustrating.

```
❯ make docker-push DOCKER_LOGIN=my-user DOCKER_PASSWORD=mypassword
>> Image == docker.elastic.co/eck-dev/eck-operator-mmontgomery:2.3.0-SNAPSHOT-bfed29a2
Authentication to docker.elastic.co...
Login Succeeded
Push docker.elastic.co/eck-dev/eck-operator-mmontgomery:2.3.0-SNAPSHOT-bfed29a2...
The push refers to repository [docker.elastic.co/eck-dev/eck-operator-mmontgomery]
denied: requested access to the resource is denied
```

Would a note to the console such as `potentially overwriting environment variables using contents of .registry.env file` be preferred?

### DIFF
--- a/hack/docker.sh
+++ b/hack/docker.sh
@@ -20,6 +20,7 @@ retry() { "$SCRIPT_DIR/retry.sh" 5 "$@"; }
 # source variables if present
 if [[ -f ${REGISTRY_ENV} ]]; then
     # shellcheck disable=SC2046
+    echo "potentially overwriting environment variables using contents of .registry.env file"
     export $(sed "s|[[:space:]]*=[[:space:]]*|=|g" "${REGISTRY_ENV}")
 fi
 
--- a/hack/docker.sh
+++ b/hack/docker.sh
@@ -19,8 +19,8 @@ retry() { "$SCRIPT_DIR/retry.sh" 5 "$@"; }
 
 # source variables if present
 if [[ -f ${REGISTRY_ENV} ]]; then
-    # shellcheck disable=SC2046
     echo "potentially overwriting environment variables using contents of .registry.env file"
+    # shellcheck disable=SC2046
     export $(sed "s|[[:space:]]*=[[:space:]]*|=|g" "${REGISTRY_ENV}")
 fi
 
--- a/hack/docker.sh
+++ b/hack/docker.sh
@@ -38,7 +38,7 @@ docker-login() {
 
     # Since this check doesn't work very well in Jenkins, and Jenkins sets
     # environment variable 'CI' only perform this check when not in Jenkins.
-    if [[ -z "${CI}" && -f ~/.docker/config.json && $(grep -q "${registry}" ~/.docker/config.json) ]]; then
+    if [[ -z "${CI}" && -f ~/.docker/config.json ]] && grep -q "${registry}" ~/.docker/config.json; then
         echo "not authenticating to ${registry} as configuration block already exists in ~/.docker/config.json"
         return
     fi
--- a/hack/docker.sh
+++ b/hack/docker.sh
@@ -28,6 +28,13 @@ docker-login() {
     local image=$1
     local registry=${image%%"/"*}
 
+    # Since this check doesn't work very well in Jenkins, and Jenkins sets
+    # environment variable 'CI' only perform this check when not in Jenkins.
+    if [[ -z "${CI}" && -f ~/.docker/config.json && `grep $registry ~/.docker/config.json` ]]; then
+        echo "not authenticating to ${registry} as configuration block already exists in ~/.docker/config.json"
+        return
+    fi
+
     case "$image" in
 
         docker.elastic.co/*)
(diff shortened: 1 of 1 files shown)


pebrc:
I wonder if we should have something like a local dev mode in this script where it does not try to authenticate against the Docker registry because in local mode you are typically already logged in? 

Something like

```diff
diff --git a/hack/docker.sh b/hack/docker.sh
index 0ad1d2aed..81818cd28 100755
--- a/hack/docker.sh
+++ b/hack/docker.sh
@@ -14,6 +14,7 @@ set -euo pipefail
 
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 REGISTRY_ENV="$SCRIPT_DIR/../.registry.env"
+SKIP_LOGIN="$SCRIPT_DIR/../.skip.docker.login"
 
 retry() { "$SCRIPT_DIR/retry.sh" 5 "$@"; }
 
@@ -27,6 +28,10 @@ docker-login() {
     local image=$1
     local registry=${image%%"/"*}
 
+    if [[ -f ${SKIP_LOGIN} ]]; then
+      return
+    fi
+
     case "$image" in
 
         docker.elastic.co/*)
```

naemono:
Any thoughts on something like this:

```
diff --git a/hack/docker.sh b/hack/docker.sh
index 9abcd91ff..a6afdbc5f 100755
--- a/hack/docker.sh
+++ b/hack/docker.sh
@@ -28,6 +28,11 @@ docker-login() {
     local image=$1
     local registry=${image%%"/"*}

+    if [[ -f ~/.docker/config.json && `grep $registry ~/.docker/config.json` ]]; then
+        echo "not authenticating to ${registry} as configuration block already exists in ~/.docker/config.json"
+        return
+    fi
+
     case "$image" in

         docker.elastic.co/*)
```

naemono:
> We had this kind of logic initially but it turned out it did not work reliably in CI where for some reason it assumed a log in was not necessary and builds in CI were failing. I don't remember the details but I am sure git history still has it somewhere.

I see it https://github.com/elastic/cloud-on-k8s/pull/4035/files#diff-bbee639ea7a1487300adf5687c0518330b529b9b014892b19f341f048e350646L29 but not notes around the reason for removing it.  

How about the latest update, which only does this check when it detects it's not in Jenkins @pebrc ?

thbkrkr:
> How about the latest update, which only does this check when it detects it's not in Jenkins?

How does it solve the issue? If you are already logged in using `.registry.env`, then you will not be logged in again using the provided `DOCKER_` env vars (`DOCKER_LOGIN=myusername DOCKER_PASSWORD=secretpassword make docker-push`).

AFAIU we want when the file `.registry.env` exists, either (1) the `DOCKER_` env vars takes precedence or (2) the script fails? I'm tempted for option (2) as it seems to me that we don't use very often `DOCKER_` env vars and I like the explicit side that requires you to remove `.registry.env` if you want to use `DOCKER_` env vars.

```diff
diff --git a/hack/docker.sh b/hack/docker.sh
index 0ad1d2aed..0e0a1b705 100755
--- a/hack/docker.sh
+++ b/hack/docker.sh
@@ -17,6 +17,11 @@ REGISTRY_ENV="$SCRIPT_DIR/../.registry.env"
 
 retry() { "$SCRIPT_DIR/retry.sh" 5 "$@"; }
 
+if [[ -f "${REGISTRY_ENV}" && "${DOCKER_LOGIN:-}" != "" ]]; then
+    echo "error: setting DOCKER_LOGIN/DOCKER_PASSWORD is incompatible the '.registry.env' file"
+    exit
+fi
+
```

naemono:
@thbkrkr I see your point.  I've set the block to fail, and left the piece where we're not logging in again if the login is already set, which is just wasteful and time consuming.

hack/docker.sh
```diff
@@ -27,6 +36,13 @@ docker-login() {
     local image=$1
     local registry=${image%%"/"*}
 
+    # Since this check doesn't work very well in Jenkins, and Jenkins sets
+    # environment variable 'CI' only perform this check when not in Jenkins.
+    if [[ -z "${CI}" && -f ~/.docker/config.json && $(grep -q "${registry}" ~/.docker/config.json) ]]; then
```
thbkrkr, line 41:
```suggestion
    if [[ -z "${CI}" && -f ~/.docker/config.json ]] && grep -q "${registry}" ~/.docker/config.json; then
```

thbkrkr: Not sure it's still relevant to keep this check with [a4c46c2](https://github.com/elastic/cloud-on-k8s/pull/5665/commits/a4c46c265028ae8fbb19a2ba111bee2a57dc3315) but if we want it, this should be the fix to make shellcheck happy.

thbkrkr (approved):

Issue #5668, opened 2022-05-13
[E2E] EKS jobs are failing 


barkbay:
EKS jobs are failing with the following error:

```
03:03:58  panic: can't determine Kubernetes version, err invalid character 'K' looking for beginning of value
03:03:58  
03:03:58  goroutine 1 [running]:
03:03:58  github.com/elastic/cloud-on-k8s/test/e2e/cmd/run.getKubernetesVersion(0x1aedc14?)
03:03:58  	/go/src/github.com/elastic/cloud-on-k8s/test/e2e/cmd/run/run.go:226 +0x2f4
03:03:58  github.com/elastic/cloud-on-k8s/test/e2e/cmd/run.(*helper).initTestContext(0xc00084c700)
03:03:58  	/go/src/github.com/elastic/cloud-on-k8s/test/e2e/cmd/run/run.go:185 +0x265
03:03:58  github.com/elastic/cloud-on-k8s/test/e2e/cmd/run.doRun({{0xc00070df60, 0x2, 0x2}, {0x7ffe2128aa2d, 0x37}, {0x7ffe2128aaf7, 0x5}, {0x7ffe2128ab14, 0x17}, {0x0, ...}, ...})
03:03:58  	/go/src/github.com/elastic/cloud-on-k8s/test/e2e/cmd/run/run.go:90 +0x663
03:03:58  github.com/elastic/cloud-on-k8s/test/e2e/cmd/run.Command.func2(0xc000838780?, {0x1adffdf?, 0x12?, 0x12?})
03:03:58  	/go/src/github.com/elastic/cloud-on-k8s/test/e2e/cmd/run/command.go:74 +0x85
03:03:58  github.com/spf13/cobra.(*Command).execute(0xc000838780, {0xc00056e5a0, 0x12, 0x12})
03:03:58  	/go/pkg/mod/github.com/spf13/cobra@v1.4.0/command.go:856 +0x67c
03:03:58  github.com/spf13/cobra.(*Command).ExecuteC(0xc000838500)
03:03:58  	/go/pkg/mod/github.com/spf13/cobra@v1.4.0/command.go:974 +0x3b4
03:03:58  github.com/spf13/cobra.(*Command).Execute(...)
03:03:58  	/go/pkg/mod/github.com/spf13/cobra@v1.4.0/command.go:902
03:03:58  main.main()
[...]

PR #5669, opened 2022-05-16
[E2E] Command.Execute: return stdout and stderr separately 

barkbay:
This is a quick and dirty hack to workaround https://github.com/elastic/cloud-on-k8s/issues/5668.
Definitely not an ideal solution, happy to close it without merging.


pebrc:
I wonder if we could simply distinguish between `stdout` and `stderr` to work around this issue. It looks like we are using `CombinedOutput` which mixes both and makes it hard to dismiss `stderr` for parsing. 

https://github.com/elastic/cloud-on-k8s/blob/20f522ff9f5fa04f8ccb22ce19edaa6d044a1aa6/test/e2e/test/command/command.go#L29

But the deprecation and warning messages are output to `stderr` in AWS CLI (which acts here as authentication plugin to `kubectl`  I believe) here: 

https://github.com/aws/aws-cli/blob/d0ab1c4864bb015349bf0f8bd34329855c0bf116/awscli/customizations/eks/get_token.py#L177

And the server/client difference warning in `kubectl` also goes to `stderr` here: 

https://github.com/kubernetes/kubernetes/blob/ea0764452222146c47ec826977f49d7001b0ea8c/staging/src/k8s.io/kubectl/pkg/cmd/version/version.go#L162

test/e2e/test/command/command.go
```diff
@@ -21,12 +22,20 @@ type Command struct {
...
 // Execute runs the command and returns the output.
-func (c *Command) Execute(ctx context.Context) ([]byte, error) {
+func (c *Command) Execute(ctx context.Context) ([]byte, []byte, error) {
 	cmd := exec.CommandContext(ctx, c.executable, c.args...) //nolint:gosec
 	cmd.Dir = c.workDir
 	cmd.Env = append(os.Environ(), c.env...)
 
-	return cmd.CombinedOutput()
+	var stderr bytes.Buffer
+	cmd.Stderr = &stderr
```
thbkrkr, line 31: To be sure I understand correctly, this will fix #5668 because the message `Kubeconfig user entry is using deprecated API version client.authentication.k8s.io/v1alpha1. Run 'aws eks update-kubeconfig' to update.` is written to `stderr`?

barkbay: I would say that it's more a workaround than a fix, but yes that's the idea. We would only try to unmarshal `stdout`, not the various warnings we may have in `stderr`

pebrc: Yes, that was the idea.

pebrc (approved):

thbkrkr (approved):

PR #5671, opened 2022-05-16
Skip node shutdown check for non-HA clusters

pebrc:
When I added the node shutdown check to the e2e tests in https://github.com/elastic/cloud-on-k8s/pull/5642 I ran all tests via our PR e2e job but this runs them on the current stack version i.e. 8.2.0 so all version upgrade tests were not run. This hid an incorrect assumption I had made about the legitimate number of restarts that can exist at given time. When we upgrade non-HA clusters we restart all nodes at once so it is OK to have more restarts in flight at the same time than the change budget allows. 

I opted to disable the check as a whole on small clusters instead of adjusting the formula to keep the calculation simple.

### DIFF
--- a/test/e2e/test/elasticsearch/checks_restart.go
+++ b/test/e2e/test/elasticsearch/checks_restart.go
@@ -58,7 +58,7 @@ func newNodeShutdownWatcher(es esv1.Elasticsearch) test.Watcher {
 			require.Empty(t, observedErrors)
 		},
 		func() bool {
-			return version.MustParse(es.Spec.Version).LT(shutdown.MinVersion)
+			return version.MustParse(es.Spec.Version).LT(shutdown.MinVersion) && es.Spec.NodeCount() > 2
 		},
 	)
 }
--- a/test/e2e/test/elasticsearch/checks_restart.go
+++ b/test/e2e/test/elasticsearch/checks_restart.go
@@ -58,6 +58,7 @@ func newNodeShutdownWatcher(es esv1.Elasticsearch) test.Watcher {
 			require.Empty(t, observedErrors)
 		},
 		func() bool {
+			// do not run this check on versions that do not support node shutdown or non-HA clusters were we restart all nodes at once
 			return version.MustParse(es.Spec.Version).LT(shutdown.MinVersion) && es.Spec.NodeCount() > 2
 		},
 	)


test/e2e/test/elasticsearch/checks_restart.go
```diff
@@ -58,7 +58,8 @@ func newNodeShutdownWatcher(es esv1.Elasticsearch) test.Watcher {
 			require.Empty(t, observedErrors)
 		},
 		func() bool {
-			return version.MustParse(es.Spec.Version).LT(shutdown.MinVersion)
+			// do not run this check on versions that do not support node shutdown or non-HA clusters where we restart all nodes at once
+			return version.MustParse(es.Spec.Version).LT(shutdown.MinVersion) && es.Spec.NodeCount() > 2
```
thbkrkr, line 62: Isn't it rather `|| es.Spec.NodeCount() < 3`?

We do a slightly more precise test for upgrades:
https://github.com/elastic/cloud-on-k8s/blob/eecd096008e7547526acd655da0ec48af3778899/test/e2e/test/elasticsearch/steps_mutation.go#L171-L177

```suggestion
			isNonHA := MustNumMasterNodes(es) < 3 || MustNumDataNodes(es) == 1
			return version.MustParse(es.Spec.Version).LT(shutdown.MinVersion) || isNonHA
```

pebrc: You are absolutely right! Monday morning brain not fully booted yet I guess.

barkbay: Would it make sense to introduce a new function and reuse it in both places?

```golang
// IsNonHA returns true if the provided Elasticsearch resource is not highly available by design. Either because it runs
// less than 3 master nodes or because there is only one data node.
func IsNonHA(es esv1.Elasticsearch) bool {
	return MustNumMasterNodes(es) < 3 || MustNumDataNodes(es) == 1
}
```

pebrc: Absolutely I will do that.

test/e2e/test/elasticsearch/steps_mutation.go
```diff
@@ -168,14 +168,21 @@ func (b Builder) MutationTestSteps(k *test.K8sClient) test.StepList {
 	return steps
 }
 
+// IsNonHASpec return true if the cluster specified as highly available.
```
thbkrkr, line 171:
```suggestion
// IsNonHASpec return true if the cluster specified as not highly available.
```

thbkrkr (approved):
LGTM

barkbay (approved):
LGTM

PR #5674, opened 2022-05-17
Be explicit about minimum TLS version

pebrc:
Go 1.18 defaults to TLS 1.2 anyway but explicitly configures the client to that minimum version. (also we are still on 1.17)


barkbay (approved):
LGTM

PR #5700, opened 2022-05-25
Update doc pages that use the repository-gcs plugin

### LINKED ISSUE #5457: Revise custom images documentation
We use the gcs-repository plugin to illustrate the process of creating custom Docker images here https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-custom-images.html

As of Elasticsearch 8.0 the common repository plugins (Azure, GCS, S3) are already bundled in the regular image so there is no need to create a custom image for that. We should use a different example as this might be misleading to users and I am not sure if the plugin installation of an already installed plugin would actually succeed.

thbkrkr:
- Update 'Create custom images' doc: replace `Google Cloud Storage Repository Plugin` by `ICU Analysis Plugin` in the examples
- Same in the 'Init containers for plugin downloads' and 'Services Meshes' doc
-  Add a note in 'Create automated snapshots' doc about  bundled plugins in ES from version 8.0

Resolves #5457.

### DIFF
--- a/docs/orchestrating-elastic-stack-applications/elasticsearch/snapshots.asciidoc
+++ b/docs/orchestrating-elastic-stack-applications/elasticsearch/snapshots.asciidoc
@@ -17,6 +17,8 @@ To set up automated snapshots for Elasticsearch on Kubernetes you have to:
 
 The examples below use the https://www.elastic.co/guide/en/elasticsearch/plugins/7.17/repository-gcs.html[Google Cloud Storage Repository Plugin].
 
+NOTE: Support for S3, GCS and Azure repositories is bundled in Elasticsearch by default from version 8.0.
+
 For more information on Elasticsearch snapshots, check https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-snapshots.html[Snapshot and Restore].
 
 [id="{p}-install-plugin"]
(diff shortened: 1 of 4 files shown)


docs/orchestrating-elastic-stack-applications/elasticsearch/snapshots.asciidoc
```diff
@@ -17,6 +17,8 @@ To set up automated snapshots for Elasticsearch on Kubernetes you have to:
 
 The examples below use the https://www.elastic.co/guide/en/elasticsearch/plugins/7.17/repository-gcs.html[Google Cloud Storage Repository Plugin].
 
+NOTE: Support for S3, GCS and Azure repositories is bundled in Elasticsearch by default from version 8.0.
```
pebrc, line 20:
```suggestion
NOTE: Support for S3, GCS and Azure repositories is bundled with Elasticsearch by default from version 8.0.
```
I know it's `in` in the official Elasticsearch documentation. But my gut feeling is it should be `with`, maybe a native speaker can help.

pebrc (approved):

docs/orchestrating-elastic-stack-applications/elasticsearch/snapshots.asciidoc
alaudazzi: Both are currently used in the official docs.

thbkrkr: Let's stay aligned with the ES doc :man_shrugging:

PR #5689, opened 2022-05-24
Upgrade PodDisruptionBudget API from v1beta1 to v1

barkbay:
This PR upgrades the API version used to manage `PodDisruptionBudget` from `v1beta1` to `v1` 

Fixes #4565 and #1773

PR #5709, opened 2022-05-30
Upgrade PodDisruptionBudget from v1beta1 to v1

barkbay:
Revives https://github.com/elastic/cloud-on-k8s/pull/5689 by adding a mechanism to detect which version of the `policy` group is available.

* `policy/v1beta1` API version of PodDisruptionBudget will no longer be served in `1.25`
* `policy/v1` version is available since `1.21`

The purpose of this new PR is then mostly for ECK `2.3.0` to be able to run on K8S `1.20` 

Note that the conversion code is not exported to `k8s.io/api`, which means that it's not possible to reuse the [native conversion mechanism](https://github.com/kubernetes/kubernetes/blob/ea0764452222146c47ec826977f49d7001b0ea8c/pkg/apis/policy/v1/conversion.go).


barkbay:
Previous build failed with:

```
#13 1.149 pkg/controller/elasticsearch/nodespec/desired_nodes.go:12:2: cannot find module providing package go.elastic.co/apm: import lookup disabled by -mod=readonly
```

🤔

pkg/controller/elasticsearch/pdb/version.go
```diff
@@ -0,0 +1,67 @@
...
+	v1beta1.Spec.MaxUnavailable = toConvert.Spec.MaxUnavailable
+	return v1beta1
+}
+
+func isPDBV1Available(k8sClient k8s.Client) (bool, error) {
+	isPDBV1Available := getPDBV1Available()
+	if isPDBV1Available != nil {
+		return *isPDBV1Available, nil
+	}
+	return initPDBV1Available(k8sClient)
```
thbkrkr, line 39: Did you consider to use `sync.Once.Do(f)`?

barkbay: > Did you consider to use sync.Once.Do(f)?

yes, I'm not sure to know how the error/retry can be handled with `sync.Once`?

thbkrkr: Good point. Me neither. I think it would make more sense if we move this in the startup of the operator but then we need to implement the retry which is free by being in a reconciliation loop and propagate the bool or make it fully global, it's not better.

thbkrkr (approved):
LGTM

Issue #5505, opened 2022-03-22
Investigate Helm charts for individual resources managed by the ECK operator


pebrc:
We currently have a Helm chart for the operator itself. However we do not have Helm charts that allow users to install individual Elastic Stack applications like Elasticsearch, Kibana etc through Helm. 

Historically we have been reluctant to built such a chart because we are afraid we will have to more or less recreate the full CRDs through Helm templating. 

Given that Helm charts are fairly easy to create we would like to create a few prototypes of different approaches to this problem so that we can make a more informed decision on how to proceed. 

Specifically it would be good to draft maybe one or two versions of a Helm chart, maybe starting with Elasticsearch which is the most complex application managed by ECK:

* a very simple wrapper around the CRD with templating just for the version and the name as these are standardised in Helm 
* a more opinionated approach with additional features (maybe with virtual memory setup, file realm users, more complex node topology)

An additional question to answer would be how we would maintain the Helm chart as we release new versions of the operator. For example is there an easy way to generate it from the latest CRD version by applying a few textual transformations. How big would the maintenance effort be if we do not do that.

noslowerdna:
@pebrc I was looking for this exact feature (specifically, a Helm chart wrapping the Elasticsearch CRD) a few months ago, and didn't find anything apart from a blog post recommending that approach (["Elastic on Elastic: How InfoSec deploys infrastructure and stays up-to-date with ECK"](https://www.elastic.co/blog/elastic-on-elastic-how-infosec-deploys-infrastructure-and-stays-up-to-date-with-eck)) and an old extremely generic project, [eck-helm-chart](https://github.com/8gears/eck-helm-chart). If it helps to have as a starting point or standalone example, [here's a gist](https://gist.github.com/noslowerdna/f8468b6c9b658cf599a1949392915c58) with the simple Helm chart I ended up creating to meet our needs (any Helm style feedback is welcome as I don't have much experience in this space). It's rather opinionated and somewhat customized for our environment, so I'm sure substantial adjustments would be necessary.

Since there are so many disparate and possibly complicated options, maybe a library of various examples accumulated from the community along with accompanying documentation providing guidance and consensus best practices would be a reasonable solution to satisfy the objectives of this issue. That could get most people 80% of the way there; I know it would have been very helpful in my case.

kaykhancheckpoint:
Any update on this?

pebrc:
First draft of how we could structure the `values.yaml` file, comments inline: 

```yaml
[...]

PR #5717, opened 2022-05-30
Helm Chart for ECK Resources V1

naemono:
Attached is the first iteration of solving #5505.  This version is simply exposing the nodesets themselves directly to the user, and attempting to simplify bits like zone awareness, which, as you will see, makes the templating much more complex.

This is intended to start a discussion on what the best approach would look like.

Another version of this is in the works to show how something such as the "topologies"

```
topologies:
  - "master"
  - "data_hot"
  - "data_cold"
```

route could work.


framsouza:
I would do it slightly differently, I think the key here is to keep it as simple as possible and give to our customers the building blocks to build their own charts.

I don't think we need/have to specify everything in the _values.yaml,_ I would keep the syntax/configuration as similar as possible to what we have today in ECK, without building everything again, by doing that we don't have to "learn" how to deal with this syntax as the idea is to keep the syntax we have today in ECK (except for some specific uses cases),

For ES `values.yaml`, I would do the following,
```
version: 8.2.0

esName: "quickstart"

labels: {}

annotations: {}

esSpec:
  nodeSets:
  - name: default
    count: 1
    config:
      node.roles: ["master", "data", "ingest"]
      node.store.allow_mmap: false
    podTemplate:
      spec:
        containers:
        - name: elasticsearch
          resources:
            requests:
              cpu: 1
            limits:
              memory: 1Gi
```
and `template/elasticsearch.yaml`
```
kind: Elasticsearch
metadata:
  name: {{ .Values.esName }}
  {{- if .Values.labels }}
  labels:
    {{ toYaml .Values.labels | indent 4 }}
  {{- end }}
  {{- if .Values.annotations }}
  annotations:
    {{ toYaml .Values.annotations | indent 4 }}
  {{- end }}
spec:
  version: {{ .Values.version }}
{{ toYaml .Values.esSpec | indent 2 }}
```

Using this configuration, the users will basically perform the "[ECK quickstart](https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-quickstart.html)" with helm 

Inside the elasticsearch helm, we could have an `example/` directory with some building blocks/examples (as we have [here](https://github.com/elastic/cloud-on-k8s/tree/main/config/recipes)),

```
├── elasticsearch
│   ├── Chart.yaml
│   ├── charts
│   ├── examples
│   │   ├── affinity-and-selector.yaml
│   │   ├── custom-persistent-volume-claim.yaml
│   │   ├── hot-warm-cold.yaml
│   │   ├── mounting-volumes.yaml
│   │   ├── saml.yaml
│   │   └── spread-constrains-allocation-awareness.yaml
│   ├── templates
│   │   ├── NOTES.txt
│   │   ├── _helpers.tpl
│   │   └── elasticsearch.yaml
│   └── values.yaml
```

One example is to setup saml, if customer what to deploy that instead of the ECK quickstart, they can run `helm install elasticsearch ./elasticsearch --values elasticsearch/examples/saml.yaml` 

And the file content, 
```
version: 8.2.0

esName: "quickstart"

labels: {}

annotations: {}

esSpec:
  nodeSets:
  - name: default
    count: 1
    config:
      node.roles: ["master", "data", "ingest"]
      node.store.allow_mmap: false
# Make sure to adjust the idp endpoints accordingly
      xpack.security.authc.realms:
        saml:
          saml1:
            attributes.principal: nameid
            idp.entity_id: urn:framsouza.eu.auth0.com
            idp.metadata.path: https://framsouza.eu.auth0.com/samlp/metadata/nbNudmA4FTPI3kO6im0o5KPV3gugxxQI
            order: 2
            sp.acs: https://kibana.framsouza.co/api/security/v1/saml
            sp.entity_id: https://framsouza.co
            sp.logout: https://kibana.framsouza.co/logout
```

Same for hot-warm, node-affinity and so one. I think we can build examples around these cases instead of put complexity in the `values.yaml`, I think it's something we can incorporate in the future if there's any specific reason to do so

Now, talking about Kibana, this is how I would do it:
```
version: 8.2.0

KibanaName: "quickstart"

labels: {}

annotations: {}

KibanaSpec:
  count: 1
  elasticsearchRef:
# Make sure to use your own elasticsearch name resource, see https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-deploy-kibana.html
    name: quickstart
    namespace: default

# Enable it if you are want to connect Kibana to an Elasticsearch that's NOT managed by ECK
esOutofECK:
  enabled: true
  elasticsearch_hosts: "https://es.example.com:9200"
  elasticsearch_username: elastic
# Create a secret to store the elasticsearch credentias, see https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-kibana-es.html#k8s-kibana-external-es
  secret_name: kibana-elasticsearch-credentials
```

and the `template/kibana.yaml`
```
kind: Kibana
metadata:
  name: {{ .Values.KibanaName }}
spec:
  version: {{ .Values.version }}
  count: 1
  {{- if .Values.esOutofECK.enabled }}
  config:
    elasticsearch.hosts:
      {{ .Values.esOutofECK.elasticsearch_hosts }}
#    elasticsearch.username: {{ .Values.esOutofECK.elasticsearch_username }}
  secureSettings:
    - secretName: {{ .Values.esOutofECK.secret_name }}
  {{- else }}
{{ toYaml .Values.KibanaSpec | indent 2 }}
  {{- end }}
```
I would like to know your thought about it @pebrc @jmlrt

pebrc (commented):
I left a few comments I am curious as to how you envisage the other stack applications to fit in. I guess there is two options: having individual charts for each application (an approach taken by #5730) and having a single chart with toggles to include/exclude certain templates. A third option would be to have individual charts per application plus a "stack" chart that conditionally depends on the individual app charts.

deploy/eck-resources/values.yaml
```diff
@@ -0,0 +1,431 @@
+# TODO:
+# 1) Verify Openshift support
+# 2) Verify support of PSPs/SCCs
+
+# Global settings that will be applied to all resources (Elasticsearch/Kibana/etc)
+#
+global:
```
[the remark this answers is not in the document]
naemono: Well, I was trying to avoid us having to manage a helm chart for each Elastic stack component, and keep things simpler with a single helm chart that supports all components (when their templates are added in a future release).  Really want other's opinions on single vs multiple helm charts for the individual components.

I added this `global` section (which could certain be renamed) so that you could have a set of labels/annotations that were applied to each resource managed by this instance of this chart, to easily identify the kibana/ES instances owned by this chart.

deploy/eck-resources/values.yaml
```diff
@@ -0,0 +1,431 @@
...
+          #     - bash
+          #     - -c
+          #     - /mnt/elastic-internal/scripts/readiness-probe-script.sh
+          #   failureThreshold: 3
+          #   initialDelaySeconds: 10
+          #   periodSeconds: 5
+          #   successThreshold: 1
+          #   timeoutSeconds: 5
+          env:
+          - name: ES_JAVA_OPTS
```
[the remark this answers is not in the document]
naemono: Sure 👍 .  This was just an example environment variable, which likely will be just removed after @framsouza 's comments, which seems to make sense having different values files for example configurations.

naemono:
@pebrc @framsouza Changes have been pushed that show the changes suggested by framsouza, along with keeping the chart as a single chart that supports multiple Elastic stack components/resources.  I'll now build an example that has multiple charts so we can compare/contrast that approach.

deploy/eck-resources/examples/elasticsearch/detailed.yaml
```diff
@@ -0,0 +1,458 @@
+# Global settings that will be applied to all resources (Elasticsearch/Kibana/etc)
+#
+global:
```
jmlrt, line 3: `global` is a keyword for values that can be accessed by all charts and subcharts (https://helm.sh/docs/chart_template_guide/subcharts_and_globals/#global-chart-values). I don't think of any side effect that could happen by re-using it for a different purpose here for now, but better check twice or use a different value name.

deploy/eck-resources/examples/elasticsearch/detailed.yaml
```diff
@@ -0,0 +1,458 @@
...
+            #   exec:
+            #     command:
+            #     - bash
+            #     - -c
+            #     - /mnt/elastic-internal/scripts/readiness-probe-script.sh
+            #   failureThreshold: 3
+            #   initialDelaySeconds: 10
+            #   periodSeconds: 5
+            #   successThreshold: 1
+            #   timeoutSeconds: 5
```
jmlrt, line 360: Maybe you could use [YAML Anchor](https://helm.sh/docs/chart_template_guide/yaml_techniques/#yaml-anchors) to avoid duplication for all these kinds of configurations that may be reused in multiple parts of the values.

jmlrt (approved):

deploy/eck-resources/Chart.yaml
```diff
@@ -0,0 +1,31 @@
...
+#
+# Library charts provide useful utilities or functions for the chart developer. They're included as
+# a dependency of application charts to inject those utilities and functions into the rendering
+# pipeline. Library charts do not define any templates and therefore cannot be deployed.
+type: application
+
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
+# Versions are expected to follow Semantic Versioning (https://semver.org/)
+version: 0.1.0
```
jmlrt, line 28: I would also recommand setting the `appVersion` field to the version of the operator.

deploy/eck-resources/examples/elasticsearch/detailed.yaml
naemono: I did initially try to use yaml anchors, and ran into helm issues with it, will try again though.

naemono:
Closing this, as we've had recent discussion on a path forward here, and will re-open with a more complete, reviewable PR.

deploy/eck-resources/examples/elasticsearch/detailed.yaml
jmlrt: >I did initially try to use yaml anchors, and ran into helm issues with it, will try again though.

TBH I never used them myself, but mostly saw them in other charts and PRs. It seems convenient in some use cases with YAML duplications like in umbrella charts or ECK spec but don't have more personal feedback on it.

PR #5723, opened 2022-06-02
Change loop into copy

jiuker:
// returns the number of elements copied, which will be the minimum of len(src) and len(dst).
	// won't handle "copy(dst, src []Type) int" nothing

copy method look better？
	// according to $GOPATH/src/builtin/builtin/copy() document
	// returns the number of elements copied, which will be the minimum of len(src) and len(dst).
	// so handle "copy(dst, src []Type) int" nothing，because len is the same!

### DIFF
--- a/pkg/controller/elasticsearch/driver/fixtures.go
+++ b/pkg/controller/elasticsearch/driver/fixtures.go
@@ -117,9 +117,7 @@ type upgradeTestPods []testPod
 
 func newUpgradeTestPods(pods ...testPod) upgradeTestPods {
 	result := make(upgradeTestPods, len(pods))
-	for i := range pods {
-		result[i] = pods[i]
-	}
+	copy(result, pods)
 	return result
 }
 


jiuker:
Jenkins test this please.

thbkrkr:
> Jenkins test this please.

Only Elastic maintainers can trigger this due to security reasons.

Jenkins test this please.

pkg/controller/elasticsearch/driver/fixtures.go
```diff
@@ -117,9 +117,7 @@ type upgradeTestPods []testPod
 
 func newUpgradeTestPods(pods ...testPod) upgradeTestPods {
```
barkbay, line 118: I think this function is useless, it can be replaced everywhere by:

```diff
index db1ec8669..5670ae57c 100644
--- a/pkg/controller/elasticsearch/driver/upgrade_predicates_test.go
+++ b/pkg/controller/elasticsearch/driver/upgrade_predicates_test.go
@@ -153,10 +153,10 @@ func TestUpgradePodsDeletion_WithNodeTypeMutations(t *testing.T) {
                        name: "Risky mutation with 7.x nodes",
                        fields: fields{
                                esVersion: "7.2.0",
-                               upgradeTestPods: newUpgradeTestPods(
+                               upgradeTestPods: upgradeTestPods{
                                        newTestPod("masterdata-0").withVersion("7.2.0").withRoles(esv1.MasterRole, esv1.DataRole).isHealthy(true).needsUpgrade(false).isInCluster(true).inStatefulset("masterdata"),
                                        newTestPod("other-master-0").withVersion("7.2.0").withRoles(esv1.MasterRole, esv1.DataRole).isHealthy(true).needsUpgrade(true).isInCluster(true).inStatefulset("other-master"),
-                               ),
+                               },
                                maxUnavailable: 1,
                                health:         client.Health{Status: esv1.ElasticsearchGreenHealth},
                                mutation:       removeMasterType("other-master"),
```

Maybe the original intent was to do something more involved than copying a slice, but I can't remember.

barkbay (approved):
LGTM, thanks.

I left a small comment about another simplification. Feel free to raise a subsequent PR if you want to.

pkg/controller/elasticsearch/driver/fixtures.go
jiuker: According to many test cases, pods are in different states, and the upgrade state is just one of them. Maybe new is better. However, according to the test cases, this function can be replaced by slice. For example:
```
[]testPod{
	{Pod1},
	{Pod2},
	{Pod3},
	…
}
```

jiuker: or just：
```
func newUpgradeTestPods(pods ...testPod) upgradeTestPods {
	return pods
}
```
I don't know which is better。

jiuker: > I think this function is useless, it can be replaced everywhere by:
> 
> ```diff
> index db1ec8669..5670ae57c 100644
> --- a/pkg/controller/elasticsearch/driver/upgrade_predicates_test.go
> +++ b/pkg/controller/elasticsearch/driver/upgrade_predicates_test.go
> @@ -153,10 +153,10 @@ func TestUpgradePodsDeletion_WithNodeTypeMutations(t *testing.T) {
>                         name: "Risky mutation with 7.x nodes",
>                         fields: fields{
>                                 esVersion: "7.2.0",
> -                               upgradeTestPods: newUpgradeTestPods(
> +                               upgradeTestPods: upgradeTestPods{
>                                         newTestPod("masterdata-0").withVersion("7.2.0").withRoles(esv1.MasterRole, esv1.DataRole).isHealthy(true).needsUpgrade(false).isInCluster(true).inStatefulset("masterdata"),
>                                         newTestPod("other-master-0").withVersion("7.2.0").withRoles(esv1.MasterRole, esv1.DataRole).isHealthy(true).needsUpgrade(true).isInCluster(true).inStatefulset("other-master"),
> -                               ),
> +                               },
>                                 maxUnavailable: 1,
>                                 health:         client.Health{Status: esv1.ElasticsearchGreenHealth},
>                                 mutation:       removeMasterType("other-master"),
> ```
> 
> Maybe the original intent was to do something more involved than copying a slice, but I can't remember.

Which's better?

PR #5730, opened 2022-06-03
Helm Charts for ECK Resources

### LINKED ISSUE #5505: Investigate Helm charts for individual resources managed by the ECK operator

framsouza:
Fix #5505 

This PR contains the first version of the ECK resources using helm-charts

The main idea is to keep the configuration as simple as possible, and as organized as possible. With this in mind, the default configuration will deploy the quickstart examples.

The main `values.yaml` in each folder/resource has a minimal version/definition. The `examples/` has a set of examples on how to custom the charts, I tried to cover all the options we have on this [page](https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-elasticsearch-specification.html).

The `recipes/` folder contains examples on how to set up observability (for instance), the idea here is to have recipes ready to handle some specific use cases.

TO BE DONE:

- [ ] Deploy Elastic Maps Server
- [ ] Deploy Enterprise Search
- [ ] Beats/Kibana/APM/Fleet Docs
- [ ] Add more `recipes/`

cc @jmlrt @Kushmaro

### DIFF
--- a/deploy/eck-resources/recipes/README.md
+++ b/deploy/eck-resources/recipes/README.md
@@ -1,0 +1,15 @@
+# ECK Resources Helm Charts Recipes
+
+This folder contains ECK commom use cases.
+
+##### Observability
+`observability/`
+
+If you wish to deploy ECK for observability purpose, you can run only one command, and we you will have:
+- Elasticsearch 
+- Kibana
+- Fleet
+- Kubernetes Integration
+
+To install:
+`helm install observability deploy/eck-resources/recipes/observability/`
--- a/deploy/eck-resources/recipes/observability/Chart.yaml
+++ b/deploy/eck-resources/recipes/observability/Chart.yaml
@@ -1,0 +1,28 @@
+apiVersion: v2
+name: observability
+description: A Helm chart for Kubernetes
+
+# A chart can be either an 'application' or a 'library' chart.
+#
+# Application charts are a collection of templates that can be packaged into versioned archives
+# to be deployed.
+#
+# Library charts provide useful utilities or functions for the chart developer. They're included as
+# a dependency of application charts to inject those utilities and functions into the rendering
+# pipeline. Library charts do not define any templates and therefore cannot be deployed.
+type: application
+
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
+# Versions are expected to follow Semantic Versioning (https://semver.org/)
+version: 0.1.0
+
+# This is the version number of the application being deployed. This version number should be
+# incremented each time you make changes to the application. Versions are not expected to
+# follow Semantic Versioning. They should reflect the version the application is using.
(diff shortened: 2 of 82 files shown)


pebrc (commented):
Thanks for opening the PR. I took a first look. I think the main thing that is no clear to me is how we want to structure the Chart with regards to the separate applications. You went for separate Charts for each application which makes sense. But I wonder if we also want a something like a stack chart with dependencies on the individual application charts. You have done something like that in the observability recipe but I don't fully understand the purpose of that one given that it seems to duplicate the charts.

deploy/eck-resources/recipes/README.md
```diff
@@ -0,0 +1,15 @@
+# ECK Resources Helm Charts Recipes
+
+This folder contains ECK commom use cases.
```
jmlrt, line 3:
```suggestion
This folder contains ECK common use cases.
```

deploy/eck-resources/recipes/observability/Chart.yaml
```diff
@@ -0,0 +1,28 @@
...
+
+# This is the version number of the application being deployed. This version number should be
+# incremented each time you make changes to the application. Versions are not expected to
+# follow Semantic Versioning. They should reflect the version the application is using.
+appVersion: 1.16.0
+
+dependencies:
+  - name: elasticsearch
+  - name: kibana
+  - name: fleet-managed-elastic-agent
```
jmlrt, line 28: I would recommend defining repositories here and not duplicate every subcharts in the `charts` directory.
See https://github.com/elastic/cloud-on-k8s/pull/5736#pullrequestreview-1001109226

jmlrt (commented):
Many thanks for providing an alternative with lighter values and multiple charts and adding so many examples of values for real use cases 👍🏻 .

naemono:
Closing this, as we've had recent discussion on a path forward here, and we will re-open with a more complete, reviewable PR.

PR #5736, opened 2022-06-06
Multi-chart Helm Chart for ECK Resources 

naemono:
2nd iteration of solving #5505 . This is a followup to #5717 from some discussion in that PR.  This shows a multi-chart example that also simply exposes the direct nodesets to the user (for Elasticsearch), and would also provide example values files for each ECK resource (Elasticsearch/Kibana/etc) that would provide more detailed/complex examples of building the intended resources.


deploy/eck-resources/charts/elasticsearch/values.yaml
```diff
@@ -0,0 +1,47 @@
+# TODO:
+# 1) Verify Openshift support
+# 2) Verify support of PSPs/SCCs
+
+# Whether to install an Elasticsearch cluster.
+#
+enabled: true
```
pebrc, line 7: Left over from previous version? IIUC this is only used in the parent chart in the condition and does not need to be included here?

jmlrt (commented):
If you decide to go with multiple chart and umbrella chart, I would rather create standalone charts for both product charts and umbrella chart with [Helm dependencies](https://helm.sh/docs/helm/helm_dependency/#helm) than embed all product charts as subchart into the umbrella's `./charts` directory. Example:

```console
➜  eck-resources tree
.
├── elasticsearch-eck
│   ├── Chart.yaml
│   ├── templates
│   │   └── elasticsearch.yaml
│   └── values.yaml
├── kibana-eck
│   ├── Chart.yaml
│   ├── templates
│   │   └── kibana.yaml
│   └── values.yaml
└── stack-eck
    ├── Chart.yaml
    └── values.yaml

5 directories, 8 files
➜  eck-resources cat stack-eck/Chart.yaml
apiVersion: v2
name: stack-eck
description: Deploy Elastic Stack (Elasticsearch & Kibana) via ECK operator
type: application
appVersion: 2.3.0
version: 0.1.2

dependencies:
- name: elasticsearch-eck
  version: "0.1.2"
  repository: "https://helm.elastic.co"
  condition: elasticsearch.enabled
- name: kibana-eck
  version: "0.1.2"
  repository: "https://helm.elastic.co.co"
  condition: kibana.enabled

➜  eck-resources cat stack-eck/values.yaml
globals:
  # ...

elasticsearch-eck:
  enabled: true
  # ...

kibana-eck:
  enabled: true
  # ...
```

I feel that this is cleaner, especially for people that want to use a product chart without the umbrella chart.

deploy/eck-resources/charts/elasticsearch/values.yaml
```diff
@@ -0,0 +1,47 @@
...
+
+# Annotations that will be applied to Elasticsearch.
+#
+annotations: {}
+  # key: value
+
+# Additional Elasticsearch specification.
+# Reference examples in examples/elasticsearch directory.
+#
+# Elasticsearch example values for a single node "all-in-one" cluster, with 1Gi storage.
```
pebrc, line 30: I assume support for all the other elements (`http`, `transport`, `secureSetting` etc)  in the spec is still coming given that this is a draft?

pebrc (commented):
I think one potential advantage of having individual charts for the stack applications and umbrella charts is that we could have more than one umbrella chart if we wanted. Let's say we want to have a default Elastic stack chart that just installs the stack and then also some form of all-in-one observability chart with production Elasticsearch + monitoring Elasticsearch clusters and Kibanas plus Agents etc that would potentially be easier to implement via a separate umbrella chart that depends multiple times on the stack product charts vs having just one chart with a ton of conditionals to toggle various applications on or off?

PR #5745, opened 2022-06-08
should have context

jiuker:
should have context


thbkrkr:
I'm not against this change because it's cleaner, although it is pretty useless. See https://github.com/elastic/cloud-on-k8s/pull/5649:

> List, Get do not because the cached client dismisses the context argument anyway. I added one where it was in scope but did not do additional refactoring to introduce one where there was not context in scope.

But I would rather if we do one, we do them all. Hence I'm closing this one.

```
> grep 'context.Background()' -r | grep -v _test.go | grep -c 'context.Background()'
255
```

If you have the time and motivation to do all the refactoring, please do it.

PR #5746, opened 2022-06-08
Set Pod Spec EnableServiceLinks to false by default

### LINKED ISSUE #2030: Set enableServiceLinks to false by default in pod specs
From Kubernetes 1.13 onwards, setting `enableServiceLinks` to `false` in the pod spec will prevent the injection of environment variables with service information into the pod. Since Elasticsearch, Kibana nor APM server depend on these injected variables, defaulting `enableServiceLinks` to `false` in all pod specs will help avoid issues with name clashes such as the one reported in #1987

thbkrkr:
Set `EnableServiceLinks` to `false` in the Pod spec to prevent the injection of environment variables with service information into the Pods.

Resolves #2030.

### DIFF
--- a/pkg/controller/common/defaults/pod_template_test.go
+++ b/pkg/controller/common/defaults/pod_template_test.go
@@ -32,6 +32,7 @@ func TestPodTemplateBuilder_setDefaults(t *testing.T) {
 			want: corev1.PodTemplateSpec{
 				Spec: corev1.PodSpec{
 					AutomountServiceAccountToken: &varFalse,
+					EnableServiceLinks: &varFalse,
 					Containers: []corev1.Container{
 						{
 							Name: "mycontainer",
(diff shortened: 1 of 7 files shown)


pkg/controller/common/defaults/pod_template_test.go
```diff
@@ -51,6 +52,27 @@ func TestPodTemplateBuilder_setDefaults(t *testing.T) {
...
+					Containers: []corev1.Container{
+						{
+							Name: "mycontainer",
+						},
+					},
+				},
+			},
+		},
+		{
+			name: "don't override user enable service links",
```
pebrc, line 65:
```suggestion
			name: "don't override user enabled service links",
```

pebrc (commented):
Code looks good. My only concern is that this causes a rolling restart of all resources. Do we have other changes that cause a restart already planned for that release?

PR #5757, opened 2022-06-09
Deployer: pick up cached vault auth token from env

pebrc:
Make deployer able to pick up cached vault auth Information from the environment. The intent here is to reduce friction for people setting up deployer for the first time. The second goal is to avoid having to store credentials, GitHub tokens or even pull secrets coming from vault in local unprotected configuration files on disk. Instead the only thing on disk will be the cached vault token (which already exists anyway once you logged in) 

I will update our documentation around this in a separate PR.


barkbay (approved):
LGTM

Issue #5713, opened 2022-05-30
TestReversalStatefulSetRename/ES_cluster_health_should_eventually_be_green is flaky


thbkrkr:
May 09th | 7.13.4        | https://devops-ci.elastic.co/job/cloud-on-k8s-e2e-tests-stack-versions/731//testReport
May 10th | 7.12.1        | https://devops-ci.elastic.co/job/cloud-on-k8s-e2e-tests-stack-versions/732/testReport/
May 12th | 7.12.1 7.14.2 | https://devops-ci.elastic.co/job/cloud-on-k8s-e2e-tests-stack-versions/734/testReport/
May 23th | 7.13.4        | https://devops-ci.elastic.co/job/cloud-on-k8s-e2e-tests-stack-versions/741/testReport/
May 30th | 7.11.2 7.14.2 | https://devops-ci.elastic.co/job/cloud-on-k8s-e2e-tests-stack-versions/746/testReport/

From the diagnostic generated with https://devops-ci.elastic.co/job/cloud-on-k8s-e2e-tests-stack-versions/746/testReport/junit/github/com_elastic_cloud-on-k8s_test_e2e_es/Run_tests_for_different_Elastic_Stack_versions_in_GKE___7_11_2___TestReversalStatefulSetRename_ES_cluster_health_should_eventually_be_green_01/.

Cluster is red:
```
# _cluster/health
epoch      timestamp cluster                        status node.total node.data shards pri relo init unassign pending_tasks max_task_wait_time active_shards_percent
1653889170 05:39:30  test-sset-rename-reversal-j9wt red             1         1      2   2    0    0        1             0                  -                 66.7%
```
Index `data-integrity-check` is red:
```
# _cat/indices
health status index                uuid                   pri rep docs.count docs.deleted store.size pri.store.size sth
red    open   data-integrity-check w2WIk1JyRIG6sfEwTe589g   3   0          4            0     14.6kb         14.6kb false
```
```
# _cat/shards
index                shard prirep state      docs  store ip            node
data-integrity-check 1     p      STARTED       4 14.4kb 10.74.241.123 test-sset-rename-reversal-j9wt-es-masterdata-0
data-integrity-check 2     p      UNASSIGNED                           
data-integrity-check 0     p      STARTED       0   208b 10.74.241.123 test-sset-rename-reversal-j9wt-es-masterdata-0
```
```
# allocation_explain.json
{
  "index" : "data-integrity-check",
  "shard" : 2,
[...]

PR #5758, opened 2022-06-09
Check shard activity before removing a node

barkbay:
This PR is an attempt to solve https://github.com/elastic/cloud-on-k8s/issues/5713
During a node removal operation, it attempts to ensure that there is no pending shard activity before fetching the shards using `_cat/shards`.

Note that recent versions of Elasticsearch which are relying on the shutdown API should not be affected.

See also https://github.com/elastic/cloud-on-k8s/issues/3070 for more context.


pebrc (approved):
I did a few simple tests last week on a 7.14.0 test cluster and looked good to me 👍

PR #5760, opened 2022-06-09
Helm Chart for ECK Resources - Enabled Topology option

naemono:
Attached is the another iteration of solving #5505. This version is using enabled topology option, discussed in #5505, which shows how this option also adds more complexity to the templating.


naemono:
Agreed in a discussion that the simpler approach of exposing the direct nodesets makes the templating much simpler, as described in #5736, #5717, and #5730, and this approach should not be taken.  Closing.

PR #5768, opened 2022-06-13
E2E: Do not leak continuous health check go routine

### LINKED ISSUE #5767: E2E error while creating the Elasticsearch client
We have error messages like the following in our e2e test recently, even long after the corresponding test `TestMutationNodeSetReplacementWithChangeBudget` has finished: 

```
error while creating the Elasticsearch client: secrets "test-1-change-budget-f27z-es-elastic-user" not found.
```

The underlying issue is that we are leaking a go routine that we spawn to monitor the cluster health during a mutation:

The go routine is created based on the result of 

https://github.com/elastic/cloud-on-k8s/blob/678092d5357c395751544c6c3489c49735986a79/test/e2e/test/elasticsearch/steps_mutation.go#L89

and it is terminated depending on evaluating the same function again a few lines below here:

https://github.com/elastic/cloud-on-k8s/blob/678092d5357c395751544c6c3489c49735986a79/test/e2e/test/elasticsearch/steps_mutation.go#L143

Now you could reasonably expect that function to return the same value for the same inputs but unfortunately it returns `false` on the first run and `true` on the second run thereby skipping the termination of the go routine.

The underlying reason took me a while to track down but it is caused because in between the two calls we update the Elasticsearch resource with the builder in this call 

https://github.com/elastic/cloud-on-k8s/blob/678092d5357c395751544c6c3489c49735986a79/test/e2e/test/elasticsearch/steps_mutation.go#L66

which has a sideeffect on the Elasticsearch object in that It apparently marshals/unmarshals the object. The `ucfg.Config` object does lose type information during this process (everything is an `interface` afterwards and the `DeepEqual` comparison here 
https://github.com/elastic/cloud-on-k8s/blob/cebc0f33f101555c7d94f1f3f4d5b9b9358d4e2d/test/e2e/test/elasticsearch/builder.go#L561

is no longer true. 

To fix this I think we should always make a defensive copy before using the object in the `Update` function of the k8s client and we should probably use a shared variable to control both the start and stop of the go routine.

pebrc:
Fixes #5767

### DIFF
--- a/test/e2e/test/elasticsearch/steps_mutation.go
+++ b/test/e2e/test/elasticsearch/steps_mutation.go
@@ -60,7 +60,9 @@ func (b Builder) UpgradeTestSteps(k *test.K8sClient) test.StepList {
 				for k, v := range b.Elasticsearch.Annotations {
 					curEs.Annotations[k] = v
 				}
-				curEs.Spec = b.Elasticsearch.Spec
+				// defensive copy as the spec struct contains nested objects like ucfg.Config that don't marshal/unmarshal
+				// without losing type information making later comparisons with deepEqual fail.
+				curEs.Spec = *b.Elasticsearch.Spec.DeepCopy()
 				// may error-out with a conflict if the resource is updated concurrently
 				// hence the usage of `test.Eventually`
 				return k.Client.Update(context.Background(), &curEs)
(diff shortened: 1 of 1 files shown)


naemono (approved):
Yeah not seeing any downside to this approach. Nice catch.

PR #5769, opened 2022-06-14
Update note about self stack-monitoring

thbkrkr:
Updates stack-monitoring doc since ECK 2.2.0 self-monitoring is possible.


barkbay (approved):
LGTM

PR #5777, opened 2022-06-15
[DOC] Update Network Policies

### LINKED ISSUE #5765: Update network policy documentation
[Network policy documentation](https://www.elastic.co/guide/en/cloud-on-k8s/master/k8s-network-policies.html#k8s-network-policies-operator-isolation) states the K8S control plane IP can be retrieved using `kubectl cluster-info | grep master` while it is not working anymore.

barkbay:
Fixes #5765

This PR also:
* Fixes the operator label name which is now `app.kubernetes.io/name`
* Adds a sample for Agent and Fleet Server

### DIFF
--- a/docs/advanced-topics/network-policies.asciidoc
+++ b/docs/advanced-topics/network-policies.asciidoc
@@ -37,9 +37,9 @@ kubectl label namespace team-b eck.k8s.elastic.co/tenant=team-b
 
 To set up the network policies correctly you must know the operator Pod selector and the Kubernetes API server IP. They may vary depending on your environment and how the operator has been installed.
 
-=== Operator Ood selector
+=== Operator Pod selector
 
-The operator Pod labels depends on how the operator has been installed, refer to the following table to know what label name should be used in the network policies.
+The operator Pod label depends on how the operator has been installed, refer to the following table to know what label name should be used in the network policies.
 
 [cols="1,1"]
 |===
--- a/docs/advanced-topics/network-policies.asciidoc
+++ b/docs/advanced-topics/network-policies.asciidoc
@@ -39,7 +39,7 @@ To set up the network policies correctly you must know the operator Pod selector
 
 === Operator Pod selector
 
-The operator Pod label depends on how the operator has been installed, refer to the following table to know what label name should be used in the network policies.
+The operator Pod label depends on how the operator has been installed. Check the following table to know which label name is used in the network policies.
 
 [cols="1,1"]
 |===
@@ -458,7 +458,7 @@ spec:
 == Isolating {agent} and {fleet}
 
 
-NOTE: Some {agent} policies may require additional access rules than what is listed here.
+NOTE: Some {agent} policies may require additional access rules other than those listed here.
 
 
 [cols="h,1"]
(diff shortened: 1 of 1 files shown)


thbkrkr (dismissed):
There are other references of the old `control-plane` label in the webhook documentation which we can fix here or in another PR.

```diff
diff --git a/docs/operating-eck/webhook.asciidoc b/docs/operating-eck/webhook.asciidoc
index 344317542..15ed8a004 100644
--- a/docs/operating-eck/webhook.asciidoc
+++ b/docs/operating-eck/webhook.asciidoc
@@ -219,7 +219,7 @@ metadata:
 spec:
   podSelector:
     matchLabels:
-      control-plane: elastic-operator
+      app.kubernetes.io/name: elastic-operator
   ingress:
   - from: []
     ports:
@@ -245,7 +245,7 @@ metadata:
 spec:
   podSelector:
     matchLabels:
-      control-plane: elastic-operator
+      app.kubernetes.io/name: elastic-operator
```

barkbay:
> So maybe let's keep that out of this PR?

Good catch! I added a comment in a "Prerequisites" section. The reason is that it is unfortunately not correct in both cases 😕
Let me know what you think.

@alaudazzi I would be glad to have your thoughts too, thanks 🙇

docs/advanced-topics/network-policies.asciidoc
```diff
@@ -32,7 +33,35 @@ kubectl label namespace team-a eck.k8s.elastic.co/tenant=team-a
 kubectl label namespace team-b eck.k8s.elastic.co/tenant=team-b
 ----
 
+== Prerequisites
 
+To set up the network policies correctly you must know the operator Pod selector and the Kubernetes API server IP. They may vary depending on your environment and how the operator has been installed.
+
+=== Operator Ood selector
```
thbkrkr, line 40:
```suggestion
=== Operator Pod selector
```

docs/advanced-topics/network-policies.asciidoc
```diff
@@ -32,7 +33,35 @@ kubectl label namespace team-a eck.k8s.elastic.co/tenant=team-a
...
+
+The operator Pod labels depends on how the operator has been installed, refer to the following table to know what label name should be used in the network policies.
```
thbkrkr, line 42:
```suggestion
The operator Pod label depends on how the operator has been installed, refer to the following table to know what label name should be used in the network policies.
```

thbkrkr:
> Good catch! I added a comment in a "Prerequisites" section. The reason is that it is unfortunately not correct in both cases confused. Let me know what you think.

Yes, that's already not correct, but I would have waited for #5788 to fix it properly, probably by also setting the label `app.kubernetes.io/name` when installing via the raw manifest so that we use only one label in the code and in the document.

But we can keep your doc update as an intermediate fix :+1:.

barkbay:
> Yes, that's already not correct, but I would have waited for https://github.com/elastic/cloud-on-k8s/issues/5788 to fix it properly

My concern is that the documentation in its current state might be misleading for users who have already deployed the operator. I agree we need to fix it, the intent here is to update the documentation for affected and supported versions.

thbkrkr (approved):
LGTM

PR #5781, opened 2022-06-15
ECK Resources Helm Chart

### LINKED ISSUE #5505: Investigate Helm charts for individual resources managed by the ECK operator

naemono:
closes #5505

This PR contains the first version of the ECK-managed resources helm charts including

1. eck-stack chart, which is a parent chart for installing all of the stack components
2. eck-elasticsearch chart for installing Elasticsearch
3. eck-kibana chart for installing Kibana

The main idea is to keep the configuration as simple as possible. With this in mind, the default configuration will deploy:

1. Elasticsearch quickstart example
2. Kibana will not be deployed unless enabled with `--set kibana.enabled=true`

The examples in `eck-stack/examples` only include the quickstart example, and an additional example for both Elasticsearch, and Kibana, for ease of initial review.  If preferred, we can add examples for all the options on this [page](https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-elasticsearch-specification.html).  Let us know.

Not included in this PR:
- Beats
- Elastic Maps Server
- Enterprise Search
- APM Server
- Agent

## TODO

- [x] Write documentation in the official ECK /docs directory.
- [x] Document using example values.
- [x] Potentially add tests.
- [x] ~~Add CI/CD to release chart(s)~~ (doesn't appear necessary, as existing process will release all charts in repo)
- [x] Documentation concerning this feature being Enterprise

## Open Questions

- [ ] How do we release this/these charts independent of the eck-operator?  Do we see the need for this initially?  (just create issue for this in the future?)
- [ ] How do we use the dev helm charts repo with this/these charts?  The dependencies on the Chart.yaml in eck-stack doesn't support this.
- [ ] Do we want to document every single option in eck-ES, and eck-Kibana readme, similar to other charts?  [See example](https://github.com/elastic/cloud-on-k8s/pull/5781/files#diff-3ca21b2db6174999d6cd02d1f6c51ceaa34a08edc9dd25350ed8698b2a33b367R70)


hack/helm/test.sh
```diff
@@ -0,0 +1,29 @@
...
+check() {
+    local TEST_DIR="$1"
+    cd "${TEST_DIR}"
+
+    docker run -ti --rm -v "$(pwd)":/apps quintush/helm-unittest:latest -3 -f 'templates/tests/*.yaml' .
+    cd -
+}
+
+# The following odd style is to resolve SC2045
+shopt -s dotglob; for i in "${SCRIPT_DIR}"/../../deploy/*; do
```
pebrc, line 22: I am not sure I follow here SC2045 is "Iterating over ls output is fragile. Use globs." you are not iterating over ls output? Also dotglob does include filenames starting with . in the glob expansion, is that really what we want here?

deploy/eck-stack/templates/tests/kibana_test.yaml
```diff
@@ -0,0 +1,48 @@
+suite: test kibana
+templates:
+  - charts/eck-kibana/templates/kibana.yaml
```
pebrc, line 3: Same here regarding dependency build as with the Elasticsearch test and I would move them to the Kibana chart and have only tests that test the stack chart here.

docs/orchestrating-elastic-stack-applications/stack-helm-chart.asciidoc
```diff
@@ -0,0 +1,70 @@
...
+[id="{p}-elasticsearch-helm-customize"]
+=== Customizing Elasticsearch using eck-stack Helm Chart
+
+There are many example Helm values files for installing and managing a more advanced Elasticsearch cluster link:{eck_github}/tree/{eck_release_branch}/deploy/eck-stack/examples/elasticsearch[in the project repository].
+
+To use one of these example configurations, use the `--values` Helm option, as seen in the following section.
+
+[source,sh]
+----
+# Install an eck-managed Elasticsearch cluster using the node roles example with hot, warm, and cold node groups.
```
pebrc, line 42:
```suggestion
# Install an eck-managed Elasticsearch cluster using the node roles example with hot, warm, and cold data tiers.
```
I think that's the official term.

docs/orchestrating-elastic-stack-applications/stack-helm-chart.asciidoc
```diff
@@ -0,0 +1,70 @@
...
+helm install es-quickstart elastic/eck-stack -n elastic-stack --create-namespace --values https://github.com/elastic/cloud-on-k8s/tree/main/deploy/eck-stack/examples/elasticsearch/hot-warm-cold.yaml
+----
+
+[float]
+[id="{p}-install-kibana-helm"]
+== Installing Kibana using the eck-stack Helm Chart
+
+Similar to the <<{p}-deploy-kibana,quickstart>>, the following section describe how to setup a simple Kibana instance, along with a simple Elasticsearch cluster managed by ECK, and how to customize a Kibana deployment using the eck-stack Helm chart.
```
pebrc, line 50:
```suggestion
Similar to the <<{p}-deploy-kibana,quickstart>>, the following section describes how to setup a simple Kibana instance, along with a simple Elasticsearch cluster managed by ECK, and how to customize a Kibana deployment using the eck-stack Helm chart.
```

docs/orchestrating-elastic-stack-applications/stack-helm-chart.asciidoc
```diff
@@ -0,0 +1,70 @@
...
+
+[source,sh]
+----
+# Install an eck-managed Elasticsearch and Kibana using the default values, which deploys the quickstart example.
+helm install es-kb-quickstart elastic/eck-stack -n elastic-stack --create-namespace --set eck-kibana.enabled=true
```
pebrc, line 55: Should the default be that Kibana is always enabled?

docs/orchestrating-elastic-stack-applications/stack-helm-chart.asciidoc
```diff
@@ -0,0 +1,70 @@
...
+[id="{p}-kibana-helm-customize"]
+=== Customizing Kibana using eck-stack Helm Chart
+
+There are many example Helm values files for installing and managing a more advanced Kibana instance link:{eck_github}/tree/{eck_release_branch}/deploy/eck-stack/examples/kibana[in the project repository].
+
+To use one of these example configurations, use the `--values` Helm option, as seen in the following section.
+
+[source,sh]
+----
+# Install an eck-managed Elasticsearch and Kibana using the Elasticsearch node roles example with hot, warm, and cold node groups, and the Kibana example customizing the http service.
```
pebrc, line 68:
```suggestion
# Install an eck-managed Elasticsearch and Kibana using the Elasticsearch node roles example with hot, warm, and cold data tiers, and the Kibana example customizing the http service.
```

deploy/eck-elasticsearch/templates/elasticsearch.yaml
```diff
@@ -0,0 +1,41 @@
+---
+apiVersion: elasticsearch.k8s.elastic.co/v1
+kind: Elasticsearch
+metadata:
+  name: {{ required "A name for the Elasticsearch instance is required" .Values.name }}
+  namespace: {{ .Release.Namespace }}
+  {{- if .Values.labels }}
```
pebrc, line 7: You are not including the fixed labels from the `elasticsearch.labels` helper template.

deploy/eck-elasticsearch/templates/_helpers.tpl
```diff
@@ -0,0 +1,51 @@
...
+{{- define "elasticsearch.name" -}}
+{{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" }}
+{{- end }}
+
+{{/*
+Create a default fully qualified app name.
+We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
+If release name contains chart name it will be used as a full name.
+*/}}
+{{- define "elasticsearch.fullname" -}}
```
pebrc, line 13: I don't see this being used anywhere.

deploy/eck-stack/examples/elasticsearch/hot-warm-cold.yaml
```diff
@@ -0,0 +1,135 @@
...
+        name: elasticsearch-data
+      spec:
+        accessModes:
+        - ReadWriteOnce
+        resources:
+          requests:
+            storage: 20Ti
+        # Adjust to your storage class name
+        #
+        storageClassName: local-storage
```
pebrc, line 135: Should we leave out the storage class names here and try to build a template that works out of the box?

deploy/eck-stack/examples/elasticsearch/quickstart.yaml
```diff
@@ -0,0 +1,84 @@
...
+              memory: 1Gi
+  # Uncomment these lines in case you want to increase vm.max_map_count and/or use initContainers.
+  # ref: https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-virtual-memory.html
+  #
+  #        initContainers: # 
+  #        - name: sysctl
+  #          securityContext:
+  #            privileged: true
+  #            runAsUser: 0
+  #          command: ['sh', '-c', 'sysctl -w vm.max_map_count=262144']
```
pebrc, line 84: This seems to duplicate more or less the `values.yaml` file of the chart. Do we need that? Also should the examples be placed in the corresponding charts?

deploy/eck-stack/examples/kibana/http-configuration.yaml
```diff
@@ -0,0 +1,42 @@
...
+        spec:
+          # Type of service to deploy for Kibana.
+          # This deploys a load balancer in a cloud service provider, where supported.
+          # 
+          type: LoadBalancer
+      tls:
+        selfSignedCertificate:
+          subjectAltNames:
+          - ip: 1.2.3.4
+          - dns: kibana.example.com
```
pebrc, line 42: Should we leave the `subjectAltNames` configuration  commented given that is not a working configuration?

deploy/eck-elasticsearch/templates/elasticsearch.yaml
```diff
@@ -0,0 +1,41 @@
...
+  labels:
+    {{- toYaml .Values.labels | indent 4 }}
+  {{- end }}
+  {{- if .Values.annotations }}
+  annotations:
+    {{- toYaml .Values.annotations | indent 4 }}
```
pebrc, line 13: I think we should also include the license annotation proposed here https://github.com/elastic/cloud-on-k8s/pull/5815

deploy/eck-kibana/Chart.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v2
+name: eck-kibana
+description: A Helm chart to deploy Kibana managed by the ECK Operator.
+type: application
+version: 0.1.0
```
jmlrt, line 5: also having `appVersion` field would be great.

deploy/eck-elasticsearch/values.yaml
```diff
@@ -0,0 +1,86 @@
...
+# - "data_content"
+# - "ml"
+# - "ingest"
+# ref: https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-node-configuration.html
+#
+nodeSets:
+- name: default
+  count: 1
+  config:
+    node.store.allow_mmap: false
```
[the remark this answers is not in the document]
naemono: I noted this, and removed the init container.

hack/helm/test.sh
naemono: Previously implementation was:
```
for i in $(ls "${SCRIPT_DIR}"/../../deploy/); do
```
hence me moving to the above style to resolve `SC2045`.

I've adjusted the glob to include alphanumeric, and removed the comment.  Hopefully it makes sense as is.

deploy/eck-stack/templates/tests/elasticsearch_test.yaml
```diff
@@ -0,0 +1,34 @@
+suite: test elasticsearch
+templates:
+  - charts/eck-elasticsearch/templates/elasticsearch.yaml
```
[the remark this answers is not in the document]
naemono: I added the dependency step to the `hack/helm/test.sh` script to fix this.  I'll address the tests, and examples and the chart directories they're in in another comment...

docs/orchestrating-elastic-stack-applications/stack-helm-chart.asciidoc
naemono: This has been changed, and documentation updated.

deploy/eck-elasticsearch/templates/_helpers.tpl
naemono: This is now used, it's use documented, and tested.

deploy/eck-elasticsearch/templates/elasticsearch.yaml
```diff
@@ -0,0 +1,41 @@
+---
+apiVersion: elasticsearch.k8s.elastic.co/v1
+kind: Elasticsearch
+metadata:
+  name: {{ required "A name for the Elasticsearch instance is required" .Values.name }}
```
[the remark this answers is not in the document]
naemono: I didn't think this initially made sense, but I've enabled it's use, documented it, and tested it to try and make it as clear as possible.

deploy/eck-elasticsearch/templates/elasticsearch.yaml
```diff
@@ -0,0 +1,41 @@
...
+  {{- if .Values.transport }}
+  transport:
+    {{- toYaml .Values.transport | indent 4 }}
+  {{- end }}
+  version: {{ required "An Elasticsearch version is required" .Values.version }}
+  {{- if .Values.monitoring }}
+  monitoring:
+    {{- toYaml .Values.monitoring | indent 4 }}
+  {{- end }}
+{{- if .Values.remoteClusters }}
```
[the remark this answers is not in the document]
naemono: Nope, missed it.  Fixed.

deploy/eck-kibana/templates/_helpers.tpl
```diff
@@ -0,0 +1,62 @@
...
+*/}}
+{{- define "kibana.selectorLabels" -}}
+app.kubernetes.io/name: {{ include "kibana.name" . }}
+app.kubernetes.io/instance: {{ .Release.Name }}
+{{- end }}
+
+{{/*
+Create the name of the service account to use
+*/}}
+{{- define "kibana.serviceAccountName" -}}
```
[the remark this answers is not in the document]
naemono: I removed this from both, as it's not used in either.  It was leftover from `helm create`.

deploy/eck-kibana/templates/kibana.yaml
```diff
@@ -0,0 +1,22 @@
...
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "kibana.labels" . | nindent 4 }}
+    {{- range $key, $value := .Values.labels }}
+    {{ $key }}: {{ $value | quote }}
+    {{- end }}
+  annotations:
+    {{- range $key, $value := .Values.annotations }}
+    {{ $key }}: {{ $value | quote }}
+    {{- end }}
```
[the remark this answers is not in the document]
naemono: Overlooked this.  Fixed.

naemono:
## Issue 1
Ok, here's one of the larger issues that both @pebrc and @jmlrt noticed, and hit upon.  I initially intended for the examples to only be in the `eck-stack` chart, but I see how it makes sense in either both locations, or in the single chart that the example supports.  Unfortunately, this brings up an issue about duplication/DRY principles, as the example values that will work in `eck-elasticsearch` chart, will not just "work" in `eck-stack` chart, and here's why.

To set version of Elasticsearch in `eck-elasticsearch` chart:

```
helm install quickstart elastic/eck-elasticsearch -n elastic-stack --create-namespace --set version=8.1.0
```

To set version of Elasticsearch in `eck-stack` chart:

```
helm install quickstart elastic/eck-stack -n elastic-stack --create-namespace --set eck-elasticsearch.version=8.1.0
```

Note that all sub-charts in the `eck-stack` chart have to have their values prefixed with the exact chart name.

So these values would work for `eck-elasticsearch` chart, if passed via `--values=my-elasticsearch-values.yaml`

```
# my-elasticsearch-values.yaml
version: 8.1.0
```

But would be ignored for the `eck-stack` chart  if passed via `--values=my-elasticsearch-values.yaml`, as there's no `eck-elasticsearch` prefix.

So if we want examples that work for both `eck-stack` Chart, and `eck-*` Charts, most of the values will have to be duplicated across directories.

## Issue 2
@jmlrt noted a couple times about `Chart.appVersion` in his comments in this PR, and in a previous PR and noted:

> I also recommend adding the [appVersion](https://helm.sh/docs/topics/charts/#the-appversion-field) set to the operator version.

I'd like to further discuss that, as I intentionally left it out, as it's usually tied to the version of the software that this chart is installing, and in our case, we're not "exactly" using the parent chart to install software.

* Do we want to tie these charts to a specific version of the operator?  If so, do we want to use that to check that some version `>=` `Chart.appVersion` is installed in the cluster prior to installation, and fail early if it's not installed?

jmlrt:
> @jmlrt noted a couple times about `Chart.appVersion` in his comments in this PR, and in a previous PR and noted:
> 
> > I also recommend adding the [appVersion](https://helm.sh/docs/topics/charts/#the-appversion-field) set to the operator version.
> 
> I'd like to further discuss that, as I intentionally left it out, as it's usually tied to the version of the software that this chart is installing, and in our case, we're not "exactly" using the parent chart to install software.
> 
> * Do we want to tie these charts to a specific version of the operator?  If so, do we want to use that to check that some version `>=` `Chart.appVersion` is installed in the cluster prior to installation, and fail early if it's not installed?

Well, I don't have a strong opinion on that, and not sure if there are best practices for charts interacting with operators. I would tend to think that `appVersion` should be tied to the operator version because the chart will not install the operator, but, it will interact with the operator to install Elastic product. If you have to add breaking changes to the custom resources specs someday this could help to figure what operator version is compatible with a specific chart version without having to add hard constraints or maintain a compatibility matrix.

docs/orchestrating-elastic-stack-applications/stack-helm-chart.asciidoc
```diff
@@ -0,0 +1,72 @@
...
+
+[source, sh]
+----
+helm repo add elastic https://helm.elastic.co
+helm repo update
+----
+
+NOTE: The minimum supported version of Helm is 3.2.0.
+
+NOTE: ECK Stack Helm Charts are currently being released as an Enterprise-level licensing feature.
```
pebrc, line 20:
```suggestion
NOTE: ECK Stack Helm Charts are currently being released as an Enterprise licensed feature.
```

deploy/eck-stack/README.md
```diff
@@ -0,0 +1,83 @@
...
+To install the ECK Operator using Helm.
+
+```sh
+# Add the Elastic Helm Repository
+helm repo add elastic https://helm.elastic.co && helm repo update
+# Install the ECK Operator cluster-wide
+helm install elastic-operator elastic/eck-operator -n elastic-system --create-namespace
+```
+
+Additional ECK Operator Helm installation options can be found within the [our documentation](https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-install-helm.html)
```
pebrc, line 36:
```suggestion
Additional ECK Operator Helm installation options can be found within [our documentation](https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-install-helm.html)
```

deploy/eck-stack/README.md
```diff
@@ -0,0 +1,83 @@
...
+
+- Kubernetes 1.20+
+- Elastic ECK Operator
+
+## Installing the Chart
+
+### Installing the ECK Operator
+
+Before using this chart, the Elastic ECK Operator is required to be installed within the Kubernetes cluster.
+Full installation instructions can be found within the [our documentation](https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-installing-eck.html)
```
pebrc, line 25:
```suggestion
Full installation instructions can be found within [our documentation](https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-installing-eck.html)
```

deploy/eck-stack/README.md
pebrc: This sentence is almost identical to the one on L25

docs/orchestrating-elastic-stack-applications/stack-helm-chart.asciidoc
```diff
@@ -0,0 +1,72 @@
...
+
+[source,sh]
+----
+# Install an eck-managed Elasticsearch cluster using the default values, which deploys the quickstart example.
+helm install es-quickstart elastic/eck-stack -n elastic-stack --create-namespace --set eck-kibana.enabled=false
+----
+
+[float]
+[id="{p}-elasticsearch-helm-customize"]
+=== Customizing Elasticsearch using eck-stack Helm Chart
```
pebrc, line 36:
```suggestion
=== Customizing Elasticsearch using the eck-stack Helm Chart
```

docs/orchestrating-elastic-stack-applications/stack-helm-chart.asciidoc
```diff
@@ -0,0 +1,72 @@
...
+
+[source,sh]
+----
+# Install an eck-managed Elasticsearch and Kibana using the default values, which deploys the quickstart examples.
+helm install es-kb-quickstart elastic/eck-stack -n elastic-stack --create-namespace
+----
+
+[float]
+[id="{p}-kibana-helm-customize"]
+=== Customizing Kibana using eck-stack Helm Chart
```
pebrc, line 62:
```suggestion
=== Customizing Kibana using the eck-stack Helm Chart
```

docs/orchestrating-elastic-stack-applications/stack-helm-chart.asciidoc
```diff
@@ -0,0 +1,72 @@
...
+
+There are many example Helm values files for installing and managing a more advanced Kibana instance link:{eck_github}/tree/{eck_release_branch}/deploy/eck-stack/examples/kibana[in the project repository].
```
pebrc, line 64: This almost verbatim repeats the Elasticsearch section. I wonder if a better structure for this page would be:
* Installing the Elastic stack using the eck-stack Helm chart (describes the default case Kibana+ES)
* Customising the Elastic stack using the eck-stack Helm chart (describes how to use the values files)
* Installing individual stack applications (describes how to turn off for example Kibana and/or how to use the sub- charts directly)

pebrc (commented):
I don't have a good answer for your issue no. 1 regarding the duplication of values files. There is an open issue in Helm to add prefixes that would allow users to use the same values file at different levels. I did not realise this limitation existed when I made the suggestion to move them to the individual charts. I think I am in favour of not duplicating them.

jmlrt:
For issue 1, maybe you can reuse [global values](https://helm.sh/docs/chart_template_guide/subcharts_and_globals/#helm) to avoid values duplication.

quick example with an umbrella chart and a subchart:
```console
➜  tree
.
├── Chart.yaml
├── charts
│   └── subchart
│       ├── Chart.yaml
│       ├── templates
│       │   └── configmap.yaml
│       └── values.yaml
├── templates
│   └── configmap.yaml
└── values.yaml

➜  cat templates/configmap.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: umbrella-configmap
data:
  version: {{ .Values.global.version }}
  app: {{ .Values.app }}

➜  cat values.yaml
global:
  version: 0.1.0
app: umbrella

➜  cat charts/subchart/templates/configmap.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: subchart-configmap
data:
  version: {{ .Values.global.version }}
  app: {{ .Values.app }}

➜  cat charts/subchart/values.yaml
global:
  version: 0.0.1
app: subchart
```

- install of the subchart alone:
```console
➜  cd charts/subchart

➜  helm template .
---
# Source: subchart/templates/configmap.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: subchart-configmap
data:
  version: 0.0.1
  app: subchart

➜  helm template . --set global.version=2.0.0 --set app=test
---
# Source: subchart/templates/configmap.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: subchart-configmap
data:
  version: 2.0.0
  app: test
```

- install of the umbrella chart:
```console
➜  subchart cd ../..

➜  helm template .
---
# Source: umbrella/charts/subchart/templates/configmap.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: subchart-configmap
data:
  version: 0.1.0
  app: subchart
---
# Source: umbrella/templates/configmap.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: umbrella-configmap
data:
  version: 0.1.0
  app: umbrella

➜  helm template . --set global.version=2.0.0 --set app=test
---
# Source: umbrella/charts/subchart/templates/configmap.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: subchart-configmap
data:
  version: 2.0.0
  app: subchart
---
# Source: umbrella/templates/configmap.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: umbrella-configmap
data:
  version: 2.0.0
  app: test

➜  helm template . --set global.version=2.0.0 --set app=test --set subchart.app=subtest
---
# Source: umbrella/charts/subchart/templates/configmap.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: subchart-configmap
data:
  version: 2.0.0
  app: subtest
---
# Source: umbrella/templates/configmap.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: umbrella-configmap
data:
  version: 2.0.0
  app: test
```

deploy/eck-elasticsearch/examples/hot-warm-cold.yaml
```diff
@@ -0,0 +1,134 @@
...
+      # ref: https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-advanced-node-scheduling.html#k8s-affinity-options
+      #
+      affinity:
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+            - matchExpressions:
+              - key: beta.kubernetes.io/instance-type
+                operator: In
+                # This should be adjusted to the instance type according to your setup
```
pebrc, line 28: One issue I see with comments like this is that they are not present in the rendered result. So a user blindly applying the values file an expecting a working configuratin to result from that would be somewhat surprised.

deploy/eck-elasticsearch/examples/hot-warm-cold.yaml
```diff
@@ -0,0 +1,134 @@
+---
+nodeSets:
+- name: hot
+  count: 1
+  config:
+    node.roles: ["data_hot", "data_content", "ingest", "master"]
```
pebrc, line 6: If we are doing dedicated data tiers I think it would make sense to have dedicated masters as well

pebrc (commented):
I am not sure if using global values will work well for the the number of configuraiton options we need would need to make effectively global to profit from this approach. I see two options 

 *  just accept the fact that examples are somewhat split between single app charts and stack chart. Maybe have examples that involve multiple apps in the stack chart e.g. a Kibana + Elasticsearch setup and examples that involve only the single app in the individual app chart. 
 * just have everything in the stack chart as you had it initially @naemono. My original comment asking for the examples to be moved to the individual app charts was made while I was unaware of the restriction in Helm that does not allow us to re-prefix values files for use in both places (there is an unresolved issue to fix that in Helm though)

naemono:
The largest issue I see with the current implementation (outside of the existing issues we are working through) is that multiple instances of any resource are not supported.  If we wanted this stack chart to implement the simplest stack monitoring configuration, which consists of multiple Elasticsearch clusters (and likely 2x kibana instances), one sending it's monitoring data to the other, we'd have to install the stack chart twice, each named differently.  I think this would also more fully come to light trying to implement many agent recipes we currently have in our code base, for example [fleet-kubernetes-integration](https://github.com/elastic/cloud-on-k8s/blob/main/config/recipes/elastic-agent/fleet-kubernetes-integration.yaml), which need 2x elastic-agents to function properly.

In the case of a parent/child chart scenario, it seems that `aliases` could be a solution to this (refer to the same child chart twice in requirements, with an alias), but it doesn't fully work per this (and many other) [issues.](https://github.com/helm/helm/issues/7093#issuecomment-701055080)

The only solution I see to this problem would be 

## Multi-chart solution
```yaml
# eck-stack values.yaml
eck-elasticsearch:
  instances:
  - name: "instance1"
    version: "8.2.3"
    nodeSets: []
  - name: "instance2"
    version: "8.2.3"
    nodeSets: []
```

## Single-chart solution
```yaml
# eck-stack values.yaml
elasticsearch:
  instances:
  - name: "instance1"
    version: "8.2.3"
    nodeSets: []
  - name: "instance2"
    version: "8.2.3"
    nodeSets: []
```

Do we want to support a list of Kibana/ES instances?

Kushmaro:
Thanks @naemono , understood. well, I don't really view "installing helm charts twice" for two clusters monitoring one another a problem, TBH, but I'm also aware I lack some context as to how other helm charts are doing this.

From the top of my mind, I'm thinking a single chart should represent "a resource" (using that vague term intentionally) where, we define such "a resource" as a an Elastic Stack. Thus, if you want to deploy "two stacks" monitoring each other, you'll have to install the two charts.. that makes sense to me but i'd be happy to get any further insight.

docs/orchestrating-elastic-stack-applications/stack-helm-chart.asciidoc
naemono: Most recent update reorganizes this documentation @pebrc .

naemono:
@Kushmaro I'll add another example where I think this issue I reference is relevant [fleet-k8s-recipe](https://github.com/elastic/cloud-on-k8s/blob/main/config/recipes/elastic-agent/fleet-kubernetes-integration.yaml).  This is a relatively simple agent example, with fleet, which involves 2x agents, which (as of today's style) wouldn't be supported.  (I'm aware eck-agent helm chart hasn't been built yet, but it'll likely follow these initial styles, which I why I bring it up)

naemono:
> * just accept the fact that examples are somewhat split between single app charts and stack chart. Maybe have examples that involve multiple apps in the stack chart e.g. a Kibana + Elasticsearch setup and examples that involve only the single app in the individual app chart.

@pebrc  I've followed this recommendation in my current changes, so let me know how you feel about this approach.

naemono:
> I don't see a change. The examples are still duplicated between stack and individual app charts. Did you maybe forget to push a commit?

I'm re-reading the comment, and I'm realizing I mis-interpreted what you said.  I'll update the PR.

> I am not sure how that solves the alias problem with Helm? Or are you suggesting to give up on the individual app charts and move all the templating into the stack chart?

I think there's some confusion here.  Aliases were mentioned as being a potential way to install 2x (or more) Elasticsearch clusters with a single chart installation, as the `eck-elasticsearch` chart currently doesn't support more than one ES cluster.  I don't think they are a good solution to that problem though, as it seems cumbersome to use.  I'm also not suggesting moving away from multiple charts, but showing that we could support a list of resources in both the single chart, and the multi-chart approach.

> But I am wondering whether you personally ran into the issue with overriding values in aliased charts during development?

I only read through the issues, and did not attempt to replicate the issue.  I'll do the same as you and replicate this, and update the PR with my findings.

naemono:
Aliases do seem to allow multiple instances of, say, Elasticsearch to be brought online with a single chart installation.  I do think the user experience is pretty poor though.

1. git clone eck-stack chart
2. manually edit `Chart.yaml`
3. install using `helm install mychartname .'

The below change does work though:
```
diff --git a/deploy/eck-stack/Chart.yaml b/deploy/eck-stack/Chart.yaml
index 557d641a1..0b7ea22cc 100644
--- a/deploy/eck-stack/Chart.yaml
+++ b/deploy/eck-stack/Chart.yaml
@@ -22,6 +22,11 @@ dependencies:
     repository: "file://../eck-elasticsearch"
     # TODO: Before merge
     # repository: "https://helm.elastic.co"
+  - name: eck-elasticsearch
+    alias: es02
+    condition: es02.enabled
+    version: "0.1.0"
+    repository: "file://../eck-elasticsearch"
   - name: eck-kibana
     condition: eck-kibana.enabled
     version: "0.1.0"
diff --git a/deploy/eck-stack/values.yaml b/deploy/eck-stack/values.yaml
index 3acf03159..3921a0835 100644
--- a/deploy/eck-stack/values.yaml
+++ b/deploy/eck-stack/values.yaml
@@ -12,3 +12,7 @@ eck-elasticsearch:
 #
 eck-kibana:
   enabled: true
+
+es02:
+  enabled: true
+  fullnameOverride: quickstart02
```

pebrc:
>  I do think the user experience is pretty poor though.

The intention behind my question was less that we expect users to manually edit the Chart to make it work but rather to get a sense of whether alias offer a way for us to build out the stack chart in the future for example with multiple instances of an Elastic Agent chart (Fleet Server and Agents). 

If based on your experiment we are somewhat confident that this is viable (despite the issue in the Helm GH repo you found) than that would help us make a call in favour of going public with what you have built so far and not redesigning now for a multi-instance version of the stack chart that is based on lists of Elasticsearches, Agents etc.

deploy/eck-kibana/templates/NOTES.txt
```diff
@@ -0,0 +1,9 @@
+
+1. Check Kibana status
+  $ kubectl get kibana --namespace={{ .Release.Namespace }} -L common.k8s.elastic.co/type=kibana
```
pebrc, line 3: See my comments to the Elasticsearch version of this. Same applies here.

deploy/eck-elasticsearch/templates/NOTES.txt
```diff
@@ -0,0 +1,9 @@
+
+1. Check Elasticsearch cluster status
+  $ kubectl get es --namespace={{ .Release.Namespace }} -L common.k8s.elastic.co/type=elasticsearch
+
+2. Check Elasticsearch pod status
+  $ kubectl get pods --namespace={{ .Release.Namespace }} -L common.k8s.elastic.co/type=elasticsearch
+
+3. Expose Elasticsearch service
+  $ kubectl port-forward svc/{{ .Values.name }}-es-http 9200
```
pebrc, line 9: `.Values.name` does not exist?  Also the http service naming convention is something like `{{ include "elasticsearch.fullname }}-es-http` I am not sure how important this instruction is though in practice maybe we can leave 3. off altogether?

deploy/eck-kibana/templates/_helpers.tpl
```diff
@@ -0,0 +1,54 @@
...
+{{- printf "%s-%s" .Chart.Name .Chart.Version | replace "+" "_" | trunc 63 | trimSuffix "-" }}
+{{- end }}
+
+{{/*
+Common labels
+*/}}
+{{- define "kibana.labels" -}}
+helm.sh/chart: {{ include "kibana.chart" . }}
+{{ include "kibana.selectorLabels" . }}
+{{- if .Chart.AppVersion }}
```
pebrc, line 39: Same as above for Elasticsearch: we don't define an `AppVersion`

deploy/eck-kibana/values.yaml
```diff
@@ -0,0 +1,44 @@
+---
+# Default values for eck-kibana.
+# This is a YAML-formatted file.
+
+# Overridable names of the Kibana instance.
+# By default, this is the Release name set for the chart,
+# followed by 'eck-kibana'.
+#
+# nameOverride will override the name of the Chart with the name set here,
+# so nameOverride: testing, would convert to '{{ Release.name }}-testing'
```
pebrc, line 10: Same as Elasticsearch I would try to be consistent in the examples

deploy/eck-kibana/values.yaml
```diff
@@ -0,0 +1,44 @@
...
+#
+# nameOverride: "quickstart"
+#
+# fullnameOverride will override both the release name, and the chart name,
+# and will name the Kibana instance exactly what is specified.
```
pebrc, line 15:
```suggestion
# and will name the Kibana instance exactly as specified.
```

deploy/eck-kibana/values.yaml
```diff
@@ -0,0 +1,44 @@
+---
+# Default values for eck-kibana.
+# This is a YAML-formatted file.
+
+# Overridable names of the Kibana instance.
```
pebrc, line 5: instance is a tricky word, here you mean Kibana later on in the comment on `count` you mean replicas when you say instance. We need to settle on consistent terminology. I personally would not mind if we called it just a Kibana resource in the context of Kubernetes, because that's what it is and name the replicas just replicas. That makes it easy for users familiar with K8s to map the concepts.

deploy/eck-kibana/values.yaml
```diff
@@ -0,0 +1,44 @@
...
+# Labels that will be applied to Kibana.
+#
+labels: {}
+
+# Annotations that will be applied to Kibana.
+#
+annotations: {}
+
+spec:
+  # Count of Kibana instances to create.
```
pebrc, line 32:
```suggestion
  # Count of Kibana replicas to create.
```

deploy/eck-kibana/values.yaml
```diff
@@ -0,0 +1,44 @@
...
+  #
+  count: 1
+
+  # Reference to ECK-managed Elasticsearch instance.
+  #
+  elasticsearchRef:
+    name: quickstart
```
pebrc, line 39: Should this default to `eck-elasticsearch` so that it works out-of-the-box.

deploy/eck-kibana/values.yaml
```diff
@@ -0,0 +1,44 @@
...
+    # Optional namespace reference to Elasticsearch instance.
+    # If not specified, then the namespace of the Kibana instance
```
pebrc, line 41:
```suggestion
    # If not specified, then the namespace of the Kibana resource
```
and so in other places.

deploy/eck-stack/Chart.yaml
```diff
@@ -0,0 +1,32 @@
+apiVersion: v2
+name: eck-stack
+description: |
+  A Parent Helm chart for all Elastic stack resources managed by the ECK Operator.
+  Currently supported:
+  * Elasticsearch
+  * Kibana
+  Future Roadmap:
```
pebrc, line 8: I would leave the Roadmap out of the Chart.

deploy/eck-stack/README.md
```diff
@@ -0,0 +1,84 @@
...
+
+```console
+$ helm delete my-release -n my-namespace
+```
+
+The command removes all the Elastic Stack resources associated with the chart and deletes the release.
+
+## Configuration
+
+The following table lists the configurable parameters of the cert-manager chart and their default values.
```
pebrc, line 67:
```suggestion
The following table lists the configurable parameters of the eck-stack chart and their default values.
```

deploy/eck-stack/templates/_helpers.tpl
```diff
@@ -0,0 +1,51 @@
...
+{{- printf "%s-%s" .Chart.Name .Chart.Version | replace "+" "_" | trunc 63 | trimSuffix "-" }}
+{{- end }}
+
+{{/*
+Common labels
+*/}}
+{{- define "eck-stack.labels" -}}
+helm.sh/chart: {{ include "eck-stack.chart" . }}
+{{ include "eck-stack.selectorLabels" . }}
+{{- if .Chart.AppVersion }}
```
pebrc, line 39: AppVersion not defined as with the product charts.

deploy/eck-stack/values.yaml
```diff
@@ -0,0 +1,14 @@
+---
+# Default values for eck-stack.
+# This is a YAML-formatted file.
+
+# If enabled, will use the eck-elasticsearch chart and deploy an Elasticsearch cluster.
+#
+eck-elasticsearch:
+  enabled: true
+  fullnameOverride: quickstart
```
pebrc, line 9: Should this be on?

deploy/eck-stack/templates/NOTES.txt
```diff
@@ -0,0 +1,6 @@
+Elasticsearch ECK-Stack {{ .Chart.Version }} has been deployed successfully!
+
+More information on the Elastic ECK Operator, and its Helm chart can be found
+within our documentation.
+
+https://www.elastic.co/guide/en/cloud-on-k8s/current/index.html
```
pebrc, line 6: We could have here an instruction like: 

```
k get elastic -l "app.kubernetes.io/instance"={{ .Release.Name}}
```

deploy/eck-stack/examples/custom-elasticsearch-kibana.yaml
```diff
@@ -0,0 +1,76 @@
...
+  version: 8.2.3
+
+  nodeSets:
+  - name: default
+    count: 1
+    # Comment out when setting the vm.max_map_count via initContainer, as these are mutually exclusive.
+    # ref: https://www.elastic.co/guide/en/cloud-on-k8s/master/k8s-virtual-memory.html
+    #
+    config:
+      node.store.allow_mmap: false
```
pebrc, line 18: Non production disclaimer needs to be added to this setting.

deploy/eck-stack/examples/custom-elasticsearch-kibana.yaml
```diff
@@ -0,0 +1,76 @@
+---
+eck-elasticsearch:
+  # Name of the Elasticsearch instance.
```
pebrc, line 3:
```suggestion
  # Name of the Elasticsearch cluster.
```

or resource?

deploy/eck-stack/examples/custom-elasticsearch-kibana.yaml
```diff
@@ -0,0 +1,76 @@
...
+  # Name of the Kibana instance.
+  #
+  fullnameOverride: quickstart
+  
+  # Version of Kibana.
+  #
+  version: 8.2.3
+  
+  spec:
+    # Count of Kibana instances to create.
```
pebrc, line 42: See my comments with regards to `instance` above for the single product chart. Applies here as well.

docs/orchestrating-elastic-stack-applications/stack-helm-chart.asciidoc
```diff
@@ -0,0 +1,69 @@
...
+----
+# Install an eck-managed Elasticsearch and Kibana using the default values, which deploys the quickstart examples.
+helm install es-kb-quickstart elastic/eck-stack -n elastic-stack --create-namespace
+----
+
+[float]
+[id="{p}-eck-stack-helm-customize"]
+=== Customizing Kibana and Elasticsearch using the eck-stack Helm Chart's example values
+
+There are many example Helm values files for installing and managing a more advanced Elasticsearch and/or Kibana link:{eck_github}/tree/{eck_release_branch}/deploy/eck-stack/examples[in the project repository].
```
pebrc, line 38:
```suggestion
There are example Helm values files for installing and managing a more advanced Elasticsearch and/or Kibana link:{eck_github}/tree/{eck_release_branch}/deploy/eck-stack/examples[in the project repository].
```

deploy/eck-elasticsearch/templates/_helpers.tpl
```diff
@@ -0,0 +1,54 @@
...
+{{- printf "%s-%s" .Chart.Name .Chart.Version | replace "+" "_" | trunc 63 | trimSuffix "-" }}
+{{- end }}
+
+{{/*
+Common labels
+*/}}
+{{- define "elasticsearch.labels" -}}
+helm.sh/chart: {{ include "elasticsearch.chart" . }}
+{{ include "elasticsearch.selectorLabels" . }}
+{{- if .Chart.AppVersion }}
```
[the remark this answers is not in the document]
naemono: This was leftover from our previous discussion about potentially including `AppVersion`.  Removed.

deploy/eck-kibana/values.yaml
naemono: I have gone through all files in the Charts looking for references to instance, or cluster `([k,K]ibana|[e,E]lasticsearch)\s(instance|cluster)` and adjusted them all to be consistently using `resource` instead.

deploy/eck-kibana/values.yaml
naemono: Here's why I initially left it this way.  When installing these charts using the parent `eck-stack` chart, I'm using `fullnameOverride: quickstart` to make these work out of the box, but I'm seeing now that installing Elasticsearch using `eck-elasticsearch` chart, it will be named `eck-elasticsearch`, by default, and if you subsequently install the `eck-kibana` chart, it will fail to work.  I'll get this changed to `eck-elasticsearch`, and adjust the `eck-stack` chart's values to use this as well.  👍

naemono: well, actually, upon further testing, the default is this:

```
{{ .Release.Name }}-{{ .Chart.Name }}
```

So installing `eck-elasticsearch` Chart using `eck-elasticsearch` as the chart name, you get:

`eck-elasticsearch-eck-elasticsearch` as the Elasticsearch name.

So you can't make the `eck-elasticsearch` chart, and the `eck-kibana` chart work together out of the box without knowing the release name up front, or using fullnameOverride by default, or not tying the ES/Kibana names to the release name at all.  @pebrc Let me know if you have a preference here.

naemono:
I'm working through why helm testing isn't wanting to work in CI...

deploy/eck-kibana/values.yaml
pebrc: I see the problem. I would in this case at least make it so that it works in the stack chart out of the box. Something like
```yaml
eck-elasticsearch:
  enabled: true
  # This is adjusting the full name here, so the default values of the `eck-kibana` chart will work
  # by default, as it references `eck-elasticsearch` in it's `elasticsearchRef`.
  fullnameOverride: elasticsearch

# If enabled, will use the eck-kibana chart and deploy a Kibana resource.
#
eck-kibana:
  enabled: true
  spec:
    elasticsearchRef:
      name: elasticsearch

Comment (naemono) [reply deploy/eck-kibana/values.yaml]: @pebrc thanks, sounds good.  That was certainly my initial intent when using `fullnameOverride` in the `eck-stack` chart's values files.  I've also added the `elasticsearchRef.name` as requested.  I think the helm unittests are finally working again 🤞 , so this should be good for the final bit of review.

File: .ci/Dockerfile [hunk 47]
```diff
@@ -82,6 +83,13 @@ RUN curl -fsSLO "https://github.com/weaveworks/eksctl/releases/download/v${EKSCT
     mv eksctl /usr/local/bin/eksctl && \
     rm eksctl_Linux_amd64.tar.gz
 
+RUN curl -fsSL -o get_helm.sh https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-3 && \
+    chmod 700 get_helm.sh && \ 
+    ./get_helm.sh -v v${HELM_VERSION} --no-sudo && \
+    rm get_helm.sh
+
+RUN helm plugin install https://github.com/quintush/helm-unittest --version 0.2.8
```
pebrc, line 91: These new requirements should be mentioned in the `dev-setup.md` file maybe in a new section about developing the Helm charts.

pebrc (approved):
LGTM!

.ci/Dockerfile
naemono: 👍 Added.  Will get this merged after latest ci tasks finished.

PR #5783, opened 2022-06-16
Move first ES cluster state observation out of go routine

### LINKED ISSUE #5812: The first ES cluster state observation is asynchronous
Opening this issue following PR #5783 to discuss about the potential issue.

Context:
- The Elasticsearch controller uses Observers to cache the ES clusters health.
- The PodDisruptionBudget is updated based on the ES cluster health ([`allowedDisruptions()`](https://github.com/elastic/cloud-on-k8s/blob/43588e5d6946367481ad0e4e4a34d7946c5b9b5c/pkg/controller/elasticsearch/pdb/reconcile.go#L186-L213))

Issue:
The issue occurs when the operator restarts. At the first ES reconciliation, the ES controller starts the Observer and retrieves the observed cluster health which is still equal to `unknown`. The reconciliation is performed with the last stored status in the ES resource, so all is good, except that the `unknown` value is used to update the ES status when the reconciliation is complete. At the second ES reconcilition, the reconciliation is performed with an `unknown` status which triggers an update of the PDB. The ES status is updated again with the last observed health when the reconciliation is complete. At the third reconciliation,  the reconciliation is performed with a known status and the PDB is updated again to return to normal.

I'm not sure that's a real issue. This translates from the operator point of view as until I don't see a green cluster I don't allow pod disruptions. It sounds ok for me.

Cause:
When an Observer is created, the cluster health is initialized to `unknown`. Then the Observer is started with a first observation and a timer to trigger next observations periodically. The whole start operation is executed asynchronously in a separated goroutine. This means there is no simple way to start the observer and get the first observed heath.

Solution proposed in #5783:
Move the first observation out of the goroutine to make it synchronous. Interestingly, it fails the [`TestManager_AddObservationListener`](https://github.com/elastic/cloud-on-k8s/blob/43588e5d6946367481ad0e4e4a34d7946c5b9b5c/pkg/controller/elasticsearch/observer/manager_test.go#L176-L212) unit test that tests the notification listeners mechanism. The test uses blocking channels to test the listeners like the ES controller does ([`WatchClusterHealthChange()`](https://github.com/elastic/cloud-on-k8s/blob/43588e5d6946367481ad0e4e4a34d7946c5b9b5c/pkg/controller/elasticsearch/observer/watch.go#L16-L30)). With the first observation synchronous that uses a listener which writes in a blocking channel, the test is stuck because the code that consumes the channel of the listener has not started. We could use asynchronous channels to fix this but...

I need a confirmation, I think that the first observation is intentionally asynchronous because we don't want to block the reconciliation if an ES cluster takes time to respond.

Let's take a step back. 
Is it a real issue?
If yes, could we consider to reconcile the PDB only when the health is not unknown? This would imply that at the creation, the PDB is created once the pods are ready.

jiuker:
There is a process to execute in Start, but the execution point is uncertain. As a result, the `lastHealth` in the observer structure is unknown, and pkg/controller/elasticsearch/pdb/reconcile.go:196 returns 0. The probability of restarting the operator will cause *v1beta1.PodDisruptionBudgetStatus `MinAvailable` to be updated from 1 (it is the correct 1 before operator restart) to 2. After the observe is executed, the `lastHealth` is green, and the `MinAvailable` is updated from 2 to 1. Therefore, restart the operator *v1beta1.PodDisruptionBudgetStatus is updated twice.

Repro steps:
1. Add `fmt.Println("update!") line 182 of `pkg/controller/common/reconciler/reconciler.go`:
2. start operator
3. kubectl apply -f config/samples/elasticsearch/elasticsearch.yaml。
4. wait all pod ready
5. restart operator
6. output:
```sh
update!
...
update!
```

Resolves #5812.


jiuker:
repro steps：
1.start operator。
2.kubctl apply -f config/samples/elasticsearch/elasticsearch.yaml。
3.wait all pod ready。
4.restart operator。
5.output:
 ```
update!
 ...
 update!
```

**Review by thbkrkr — CHANGES_REQUESTED** [review]
Nice catch! 

If you look at `runPeriodically()`, there was an intention to trigger a first observation before doing it periodically but it has been done before the go routine starts. So I would rather we fix `runPeriodically() / Start()`. I think we can merge the two because the indirection provided by `runPeriodically()` makes reading the flow a bit harder and let us make this bug exist, and then move the first observation out of the go routine, like this:

```diff
diff --git a/pkg/controller/elasticsearch/observer/observer.go b/pkg/controller/elasticsearch/observer/observer.go
index 1fb8bc8aa..31fd5d415 100644
--- a/pkg/controller/elasticsearch/observer/observer.go
+++ b/pkg/controller/elasticsearch/observer/observer.go
@@ -65,9 +65,26 @@ func NewObserver(cluster types.NamespacedName, esClient client.Client, settings
        return &observer
 }
 
-// Start the observer in a separate goroutine
+// Start starts the Observer in a separate goroutine.
+// The cluster state is observed periodically until the observer stop channel is closed.
 func (o *Observer) Start() {
-       go o.runPeriodically()
+       // initial synchronous observation
+       o.observe()
+       // periodic asynchronous observations
+       go func() {
+               ticker := time.NewTicker(o.settings.ObservationInterval)
+               defer ticker.Stop()
+
+               for {
+                       select {
+                       case <-ticker.C:
+                               o.observe()
+                       case <-o.stopChan:
+                               log.Info("Stopping observer for cluster", "namespace", o.cluster.Namespace, "es_name", o.cluster.Name)
+                               return
+                       }
+               }
+       }()
 }
 
 // Stop the observer loop
@@ -85,25 +102,6 @@ func (o *Observer) LastHealth() esv1.ElasticsearchHealth {
        return o.lastHealth
 }
 
-// runPeriodically triggers a state retrieval every tick,
-// until the given context is cancelled
-func (o *Observer) runPeriodically() {
-       o.observe()
-
-       ticker := time.NewTicker(o.settings.ObservationInterval)
-       defer ticker.Stop()
-
-       for {
-               select {
-               case <-ticker.C:
-                       o.observe()
-               case <-o.stopChan:
-                       log.Info("Stopping observer for cluster", "namespace", o.cluster.Namespace, "es_name", o.cluster.Name)
-                       return
-               }
-       }
-}
-
```

jiuker:
> Nice catch!
> 
> If you look at `runPeriodically()`, there was an intention to trigger a first observation before doing it periodically but it has been done before the go routine starts. So I would rather we fix `runPeriodically() / Start()`. I think we can merge the two because the indirection provided by `runPeriodically()` makes reading the flow a bit harder and let us make this bug exist, and then move the first observation out of the go routine, like this:
> 
> ```diff
> diff --git a/pkg/controller/elasticsearch/observer/observer.go b/pkg/controller/elasticsearch/observer/observer.go
> index 1fb8bc8aa..31fd5d415 100644
> --- a/pkg/controller/elasticsearch/observer/observer.go
> +++ b/pkg/controller/elasticsearch/observer/observer.go
> @@ -65,9 +65,26 @@ func NewObserver(cluster types.NamespacedName, esClient client.Client, settings
>         return &observer
>  }
>  
> -// Start the observer in a separate goroutine
> +// Start starts the Observer in a separate goroutine.
> +// The cluster state is observed periodically until the observer stop channel is closed.
>  func (o *Observer) Start() {
> -       go o.runPeriodically()
> +       // initial synchronous observation
> +       o.observe()
> +       // periodic asynchronous observations
> +       go func() {
> +               ticker := time.NewTicker(o.settings.ObservationInterval)
> +               defer ticker.Stop()
> +
> +               for {
> +                       select {
> +                       case <-ticker.C:
> +                               o.observe()
> +                       case <-o.stopChan:
> +                               log.Info("Stopping observer for cluster", "namespace", o.cluster.Namespace, "es_name", o.cluster.Name)
> +                               return
> +                       }
> +               }
> +       }()
>  }
>  
>  // Stop the observer loop
> @@ -85,25 +102,6 @@ func (o *Observer) LastHealth() esv1.ElasticsearchHealth {
>         return o.lastHealth
>  }
>  
> -// runPeriodically triggers a state retrieval every tick,
> -// until the given context is cancelled
> -func (o *Observer) runPeriodically() {
> -       o.observe()
> -
> -       ticker := time.NewTicker(o.settings.ObservationInterval)
> -       defer ticker.Stop()
> -
> -       for {
> -               select {
> -               case <-ticker.C:
> -                       o.observe()
> -               case <-o.stopChan:
> -                       log.Info("Stopping observer for cluster", "namespace", o.cluster.Namespace, "es_name", o.cluster.Name)
> -                       return
> -               }
> -       }
> -}
> -
> ```

Done

thbkrkr:
Jenkins test this please

thbkrkr (approved):

jiuker:
I modified the E2E. Becase our usage scenarios do not conform to the E2E. So there is a problem with the E2E test, and the program runs without problem. @thbkrkr

thbkrkr:
Jenkins test this please

jiuker:
It is possible that other test cases are affected. I should do more tests. Sorry.

jiuker:
I have tested package observer all test Fun。Make jenkins test again。 @thbkrkr

pkg/controller/elasticsearch/observer/observer.go
```diff
@@ -65,9 +65,26 @@ func NewObserver(cluster types.NamespacedName, esClient client.Client, settings
 	return &observer
 }
 
-// Start the observer in a separate goroutine
+// Start starts the Observer in a separate goroutine.
```
barkbay, line 68: godoc should be updated to mention that the first observation is synchronous

pkg/controller/elasticsearch/observer/observer_test.go
```diff
@@ -78,12 +78,17 @@ func TestNewObserver(t *testing.T) {
 	onObservation := func(cluster types.NamespacedName, previousHealth, newHealth esv1.ElasticsearchHealth) {
 		events <- cluster
 	}
+	doneCh := make(chan struct{})
+	go func() {
+		// let it observe at least 3 times
+		require.Equal(t, types.NamespacedName{Namespace: "ns", Name: "cluster"}, <-events)
+		require.Equal(t, types.NamespacedName{Namespace: "ns", Name: "cluster"}, <-events)
+		require.Equal(t, types.NamespacedName{Namespace: "ns", Name: "cluster"}, <-events)
+		doneCh <- struct{}{}
```
barkbay, line 87:
```suggestion
		close(doneCh)
```

pkg/controller/elasticsearch/observer/manager_test.go
```diff
@@ -197,18 +197,21 @@ func TestManager_AddObservationListener(t *testing.T) {
...
 	})
-
+	doneCh := make(chan struct{})
+	go func() {
+		// events should be propagated to both listeners
+		<-eventsCluster1
+		<-eventsCluster2
+		<-eventsCluster1
+		<-eventsCluster2
+		doneCh <- struct{}{}
```
barkbay, line 207:
```suggestion
		close(doneCh)
```

thbkrkr:
Jenkins test this please

jiuker:
review again @barkbay

pkg/controller/elasticsearch/observer/observer.go
```diff
@@ -65,9 +65,27 @@ func NewObserver(cluster types.NamespacedName, esClient client.Client, settings
...
+// First observation is synchronous,Set lastHealth immediately to prevent it from causing secondary updates to other resources during initialization.
+// The cluster state is observed periodically until the observer stop channel is closed.
```
thbkrkr, line 70:
```suggestion
// Start starts the Observer in a separate goroutine after a first synchronous observation.
// The first observation is synchronous to allow to retrieve the cluster state immediately after the start.
// Then, observations are performed periodically in a separate goroutine until the observer stop channel is closed.
```

jiuker:
review again@[thbkrkr](https://github.com/thbkrkr)

thbkrkr:
Jenkins test this please

thbkrkr (approved):

barkbay (approved):
LGTM

barkbay:
I think this PR has actually introduced some performance issues described here: https://github.com/elastic/cloud-on-k8s/issues/6078

PR #5785, opened 2022-06-16
[E2E] Upgrade Kind node versions

barkbay:
Upgrade Kind nodes images that rely on `runc` version `1.1.1`

### DIFF
--- a/.ci/packer_cache.sh
+++ b/.ci/packer_cache.sh
@@ -12,7 +12,7 @@ set -eou pipefail
 docker pull "$(make -C .ci --no-print-directory print-ci-image)"
 
 # Kind images (https://hub.docker.com/r/kindest/node/tags)
-# Pull exact images due to image incompatibility between kind 0.8 and 0.9 see https://github.com/kubernetes-sigs/kind/releases
+# Pull exact images to speed the Kind cluster startup and ensure images compatibility between kind versions, see https://github.com/kubernetes-sigs/kind/releases
 docker pull kindest/node:v1.20.15@sha256:6f2d011dffe182bad80b85f6c00e8ca9d86b5b8922cdf433d53575c4c5212248
 docker pull kindest/node:v1.21.12@sha256:f316b33dd88f8196379f38feb80545ef3ed44d9197dca1bfd48bcb1583210207
 docker pull kindest/node:v1.22.9@sha256:8135260b959dfe320206eb36b3aeda9cffcb262f4b44cda6b33f7bb73f453105
--- a/.ci/packer_cache.sh
+++ b/.ci/packer_cache.sh
@@ -13,8 +13,8 @@ docker pull "$(make -C .ci --no-print-directory print-ci-image)"
 
 # Kind images (https://hub.docker.com/r/kindest/node/tags)
 # Pull exact images due to image incompatibility between kind 0.8 and 0.9 see https://github.com/kubernetes-sigs/kind/releases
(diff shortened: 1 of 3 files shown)


.ci/packer_cache.sh
```diff
@@ -13,8 +13,8 @@ docker pull "$(make -C .ci --no-print-directory print-ci-image)"
 
 # Kind images (https://hub.docker.com/r/kindest/node/tags)
 # Pull exact images due to image incompatibility between kind 0.8 and 0.9 see https://github.com/kubernetes-sigs/kind/releases
```
barkbay, line 15: Not sure it is still needed...

hack/deployer/config/plans.yml
```diff
@@ -164,10 +164,10 @@ plans:
   clusterName: kind-ci
   clientVersion: 0.11.1
```
thbkrkr, line 165: Also bump the kind version to [v0.14.0](https://github.com/kubernetes-sigs/kind/releases/tag/v0.14.0) (duplicated in the `kind-dev` profile)?

.ci/packer_cache.sh
thbkrkr: Do you mean the comment, the use of exact images or the whole script?

Use exact images? I think we still need it. I see 2 different `v1.23.6` images for kind 0.13.0 and 0.14.0.
The comment? Still good to get some context why we use the ids in the image names.
The script? I think it's worth it, it speeds up the job a bit.

I would just update the comment to make it more generic:

```suggestion
# Pull exact images due to images incompatibility between kind versions see https://github.com/kubernetes-sigs/kind/releases
```

hack/deployer/config/plans.yml
```diff
@@ -162,12 +162,12 @@ plans:
...
   clusterName: kind-ci
-  clientVersion: 0.11.1
+  clientVersion: 0.14.0
   provider: kind
-  kubernetesVersion: 1.21.1
+  kubernetesVersion: 1.22.9
   kind:
     nodeCount: 3
-    nodeImage: kindest/node:v1.22.0@sha256:b8bda84bb3a190e6e028b1760d277454a72267a5454b57db34437c34a588d047
+    nodeImage: kindest/node:v1.22.9@sha256:8135260b959dfe320206eb36b3aeda9cffcb262f4b44cda6b33f7bb73f453105
```
thbkrkr, line 170: Let's also update `kind-dev` defined above?

.ci/packer_cache.sh
barkbay: > Use exact images? I think we still need it. I see 2 different v1.23.6 images for kind 0.13.0 and 0.14.0.

The whole script, I don't think we need to use 2 different versions of Kind anymore. I feel it increases a bit the maintenance complexity for what seems to be  more or less 30 seconds from GCP:

```
time docker pull kindest/node:v1.24.1
v1.24.1: Pulling from kindest/node
f500296146dd: Pull complete
3db0d32e06a6: Pull complete
Digest: sha256:11276ce4763d5600379c0104c4cb51bf7420c8c02937708e958ad68d08e0910c
Status: Downloaded newer image for kindest/node:v1.24.1
docker.io/kindest/node:v1.24.1

real	0m30.683s
user	0m0.073s
sys	0m0.063s
```

That being said it's not a strong feeling 😄 ,  I'm perfectly fine with keeping it. Let's just update the comment.

thbkrkr: I'm ok to delete it and see if we really perceive an increase in the average duration of execution of the pr job.

thbkrkr: The fewer bits on the network, the better for the :earth_africa:?

thbkrkr (approved):
LGTM

PR #5793, opened 2022-06-17
Increase default Beats guaranteed memory to 300Mi

### LINKED ISSUE #5036: Revisit Filebeat default memory requests/limits
We are currently defaulting to 200Mi. However the Filebeats ingesting e2e test cluster data into our monitoring were constantly running into OOM. We should reevaluate whether the current default is still enough with recent version of Filebeat/Metricbeat.

pebrc:
Fixes #5036 

We did test drive this in #5681, but that PR did not cover all e2e tests especially the recipe tests that use default settings and do not rely on the e2e config template. Consequently we still see OOMing on those tests. I have also reached out to the Beats team for guidance and would wait with merging until we get a response.


pebrc:
run/e2e-tests tags=beat

thbkrkr (approved):
LGTM

PR #5796, opened 2022-06-17
Stricter notion of esReacheable: require health response

### LINKED ISSUE #5776: Cannot verify license message is misleading
When a cluster is not available the error message that is most prominent is `cannot verify license` this is very misleading because what is actually happening is that the user either does not have a cluster (because it did not form) or the cluster is not reachable (because it is overloaded or there is a network issue) 

We should signal this condition as such and reserve the license check message to cases where the cluster is actually available but we cannot verify the license for other reasons.

pebrc:
Fixes #5776 

Simplest (?) possible fix without doing extra requests. The goal is avoid confusing error messages like "could not verify license" but instead have a meaningful status like "Elasticsearch is not reachable". The implication here is that `esReachable` is now tied to the first successful observation of cluster health. Those happen asynchronously in the observer mechanism and are not done inside the reconciliation loop. I think that is OK as the cluster is not available immediately but with a certain delay anyway, depending on hardware and the time it takes for the ES container in the Pods to become ready. If the timing is unfortunate we might lose an additional 10 seconds (which is the current observation interval) which seems acceptable to me.

### DIFF
--- a/pkg/controller/elasticsearch/driver/driver.go
+++ b/pkg/controller/elasticsearch/driver/driver.go
@@ -224,6 +224,8 @@ func (d *defaultDriver) Reconcile(ctx context.Context) *reconciler.Results {
 	defer esClient.Close()
 
 	esReachable, err := services.IsServiceReady(d.Client, *internalService)
+	// use unknown health as a proxy for a cluster not responding to requests
+	esReachable = esReachable && observedState() != esv1.ElasticsearchUnknownHealth
 	if err != nil {
 		return results.WithError(err)
 	}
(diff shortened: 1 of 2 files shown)


pebrc:
run/e2e-tests tags=es

barkbay (commented):
Thanks for raising this PR, I agree that the current status may be confusing.
I'm must confess however I'm a bit hesitant regarding the way to improve things as there can be 2 reasons for having an "unknown" state:
1. The operator has not attempted to reach the cluster yet, in which case I would expect the condition to be `corev1.ConditionUnknown`. We don't really no if Elasticsearch is not responding to requests.
2. The operator attempted to get the cluster state, but an error occurred (time out, credentials issue...). Health is still unknown but we now have an error to explain why. The condition should then be `corev1.ConditionFalse`, with maybe the error as the condition message.

That being said you can argue that the observer should attempt a first connection pretty quickly once the operator is started or when a cluster is created, so may be I'm overthinking...

barkbay (approved):
> If the timing is unfortunate we might lose an additional 10 seconds (which is the current observation interval) which seems acceptable to me.

Same. We could still consider it as a first step and improve it later if needed.

PR #5809, opened 2022-06-20
Migrate to /v2 for easier go modules import

### LINKED ISSUE #5095: Move to `vX.X.X` tags for easier go module imports
Go modules rely on a `vX.X.X` tag naming scheme. Currently we use an `X.X.X` scheme, which does not make it easy to import this project as a dependency in other projects.

Let's move to `vX.X.X` for the next tag.

Additionally, we should probably care about what we do once migrating to `v2.x`, which seems far from trivial: https://donatstudios.com/Go-v2-Modules.

naemono:
resolves #5095

Migrate to a /v2 import path to allow for applications to have an easier time importing contained packages.  ~~I don't *think* we can do this without migrating to `/v3`, as opposed to`/v2` since we already are in version 2.x.  Since we are already in the process of releasing `2.3.0`, we'd have to move to `3.0.0` in the next release cycle to incorporate this change.~~

We decided that would could move to /v2 instead of moving to /v3.


naemono:
There's some build issues in CI I'm working through, btw...

pebrc:
I feel like doing a major release just to introduce the `v3` module is maybe a step too far? What is the impact if we move to `v2` in a minor release? Is there any? I assume anyone consuming ECK as a library is currently pinning to individual commits due to the lack of a correctly versioned module path? But maybe I am missing something. Curious to hear everyone's thoughts.

thbkrkr:
>  I assume anyone consuming ECK as a library is currently pinning to individual commits

Yes, it looks like that, or a tag: [github.com/search: "filename:go.mod github.com/elastic/cloud-on-k8s"](https://github.com/search?q=filename%3Ago.mod+github.com%2Felastic%2Fcloud-on-k8s&type=code).

I also don't see the issue of moving to `/v2` and releasing it in the next minor.

naemono:
👍 sounds good.  It simply felt like a change we'd want to increment major version for.  I struggle to find a technical reason outside of imports, which shouldn't be an issue because of being locked to a commit hash.  I'll update this to /v2 and note when done

thbkrkr:
I took a quick look and it looks good! I found another occurrence to change.

```diff
diff --git a/Makefile b/Makefile
index 2289f381b..ac2ba12c0 100644
--- a/Makefile
+++ b/Makefile
@@ -54,10 +54,10 @@ OPERATOR_DOCKERHUB_IMAGE_UBI ?= docker.io/elastic/$(IMG_NAME)-ubi8:$(IMG_VERSION
 print-operator-image:
        @ echo $(OPERATOR_IMAGE)
 
-GO_LDFLAGS := -X github.com/elastic/cloud-on-k8s/pkg/about.version=$(VERSION) \
-       -X github.com/elastic/cloud-on-k8s/pkg/about.buildHash=$(TAG) \
-       -X github.com/elastic/cloud-on-k8s/pkg/about.buildDate=$(shell date -u +'%Y-%m-%dT%H:%M:%SZ') \
-       -X github.com/elastic/cloud-on-k8s/pkg/about.buildSnapshot=$(SNAPSHOT)
+GO_LDFLAGS := -X github.com/elastic/cloud-on-k8s/v2/pkg/about.version=$(VERSION) \
+       -X github.com/elastic/cloud-on-k8s/v2/pkg/about.buildHash=$(TAG) \
+       -X github.com/elastic/cloud-on-k8s/v2/pkg/about.buildDate=$(shell date -u +'%Y-%m-%dT%H:%M:%SZ') \
+       -X github.com/elastic/cloud-on-k8s/v2/pkg/about.buildSnapshot=$(SNAPSHOT)
```

thbkrkr (approved):
LGTM

barkbay:
@naemono Does this fully resolve #5095? My understanding is that we must also change the way we name the git `tags` (for example `2.4.0` -> `v2.4.0`)?

naemono:
@barkbay No, it does not appear to fully resolve #5095.  `elastic/cloud-on-k8s` is still imported as a git sha:

```
❯ mkdir cd ~/tmp/test && cd ~/tmp/test && go mod init github.com/naemono/mytest
❯ go get github.com/elastic/cloud-on-k8s/v2@2.4.0
go: added github.com/elastic/cloud-on-k8s/v2 v2.0.0-20220812092235-96282ca98df1

❯ cat go.mod
module github.com/naemono/mytest

go 1.19

require github.com/elastic/cloud-on-k8s/v2 v2.0.0-20220812092235-96282ca98df1 // indirect
```

I tested in a test git repo, and found that when prefixing the release with `v`, the import stopped using the SHA, and just used the version, ie `v2.0.1`.  

```
❯ cat go.mod
module github.com/naemono/mytest

go 1.19

require (
	github.com/organization/go-modules-test/v2 v2.0.1 // indirect
```

I'll open a final issue.

PR #5820, opened 2022-06-22
E2E: Use Open Policy Agent to enforce security constraints

### LINKED ISSUE #5726: [E2E] Remove PodSecurityPolicy dependency (K8S 1.25)
`PodSecurityPolicy` API is removed in K8S `1.25`. While this API is not actively used by the operator we rely on it to ensure that stack applications are running with restricted privileges during our e2e tests  (with the exception of `Beats` and `Agent`).

An alternative would be to migrate to [Pod Security Admission](/docs/concepts/security/pod-security-admission/). There is however a significant difference in that Pod Security Admission does not support sub-namespace policy granularity. Said differently: all the workloads in a namespace are bound to a same security constraints policy, it is not possible to mix restricted and privileged policies within a same namespace.

If we decide to rely on the PodSecurity admission controller we would have then to create a dedicated namespace for privileged workloads like Agent and Beats. It might require to update the associations (for example `Agent` deployed in a first namespace, while `Kibana` and `Elasticsearch` are deployed in a second one), I don't know for now the impact of such a change and if it can easily be done.

Another alternative would be to stop relying on any security policy constraints mechanism, assuming that [`DefaultSecurityContext`](https://github.com/elastic/cloud-on-k8s/blob/721cccfd905f43c345c7957fbfe173ea3ebb6bbf/test/e2e/test/default.go#L16-L36) is always set and unexpected privilege requirements would be detected by the OpenShift pipelines (only run during the week-end though, not ideal as a "reactive" feedback loop).

Related documentation:
* [Migrate from PSP](https://kubernetes.io/docs/tasks/configure-pod-container/migrate-from-psp/)
* [Pod Security Admission concept](https://kubernetes.io/docs/concepts/security/pod-security-admission/)

barkbay:
This PR is draft/test to replace the use of `PodSecurityPolicies` with [Gatekeeper](https://github.com/open-policy-agent/gatekeeper).

Goal is twofold:
1. Replace PSPs with a standard way to enforce security policies.
2. Make it (hopefully) easier for everyone to deploy security policies locally and ensure that Pods do not require additional privileges.

Fixes #5726


barkbay:
Closing in favor of https://github.com/elastic/cloud-on-k8s/pull/5901

PR #5824, opened 2022-06-23
Upgrade test harness: fix CRDs first install

barkbay:
If the `from` release contains a `crds.yaml` file then the upgrade test harness tool always attempts to replace CRDs, even if they have not be installed yet.

The problem can be reproduced using `go run main.go --from-release=v191 --to-release=upcoming` for example:

```
2022-06-23T10:31:07.039+0200    ERROR   eck-upgrade.fixture     [FAIL] InstallCRDs[v191]        {"step": "TestInstallOperator[v191]", "step": "InstallCRDs[v191]", "error": "[error when replacing \"testdata/v191/crds.yaml\": customresourcedefinitions.apiextensions.k8s.io \"agents.agent.k8s.elastic.co\" not found,
```

### DIFF
--- a/hack/upgrade-test-harness/fixture/eck.go
+++ b/hack/upgrade-test-harness/fixture/eck.go
@@ -51,16 +51,23 @@ func (tp TestParam) Suffixed(name string) string {
 }
 
 // TestInstallOperator is the fixture for installing an operator.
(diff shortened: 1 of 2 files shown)


thbkrkr (approved):
LGTM

PR #5826, opened 2022-06-23
E2E: Resize azure/aks vms from v3 -> v5

### LINKED ISSUE #5825: E2E tests in Azure/AKS are failing - pod never becomes ready
End-to-End tests in Azure/AKS are failing as the e2e pod that is attempting to be started, never becomes ready, and the whole process fails.  This seems to be associated with Disk Pressure:

```
20:00:11  {"reason":"Evicted","message":"The node had condition: [DiskPressure]. ","kind":"Pod","name":"eck-e2e-y17n6--1-kvv5n","namespace":"e2e-y17n6"}
20:00:11  {"reason":"BackoffLimitExceeded","message":"Job has reached the specified backoff limit","kind":"Job","name":"eck-e2e-y17n6","namespace":"e2e-y17n6"}
```

I suspect we need to adjust the disk size in Azure for these to become healthy again.

naemono:
Fixes #5825 

Since AKS E2E tests appear to be failing because of disk pressure, let's take the opportunity to upgrade the version type of our VMs, and leverage more local storage.

Comparison of proposed instance type change: 

| Size  | vCPU | Memory: GiB | TempStorage (ssd) GiB | Cost/hr | Architecture |
| ------------- | ------------- | ---- | ---- | --- | -- |
| Standard_D8s_v3  | 8  |  32 | 64 | 0.3840 | 3rd Generation Intel® Xeon® Platinum 8370C (Ice Lake), (Cascade Lake), (Broadwell) or (Haswell) |
| Standard_D8ds_v5  | 8 |  32 | 300 | 0.4520 | 3rd Generation Intel® Xeon® Platinum 8370C (Ice Lake) |

### DIFF
--- a/hack/deployer/config/plans.yml
+++ b/hack/deployer/config/plans.yml
@@ -46,7 +46,7 @@ plans:
   clusterName: ci
   provider: aks
   kubernetesVersion: 1.22.6
-  machineType: Standard_D8s_v3
+  machineType: Standard_D8d_v5
   serviceAccount: true
   psp: false
   diskSetup: kubectl apply -k hack/deployer/config/local-disks
@@ -59,7 +59,7 @@ plans:
   clusterName: dev
   provider: aks
   kubernetesVersion: 1.22.6
-  machineType: Standard_D8s_v3
+  machineType: Standard_D8d_v5
   serviceAccount: false
   psp: false
   aks:


pebrc (approved):
Let's try it but I find it odd that 64Gi of local storage are not enough for what we are doing in those test runs.

barkbay:
From the [Azure documentation](https://docs.microsoft.com/en-us/azure/virtual-machines/managed-disks-overview#temporary-disk):

> On Azure Linux VMs, the temporary disk is typically /dev/sdb

If I SSH into an Azure K8S node:

```
root@aks-nodepool1-39942109-vmss000002:/# df -m
Filesystem     1M-blocks  Used Available Use% Mounted on
/dev/sda1          29599 24157      5427  82% /
/dev/sda15           105     5       100   5% /boot/efi
/dev/sdb1         271116    61    257216   1% /mnt
```

I think it explains why changing the node type did not improve the situation: `containerd` stores containers data in `/dev/sda1` ... which is already `82%` full when the node is created, the "TempStorage", in `sdb`, is not used.

I can manually make room by cleaning up existing images and get back to `15%` used.

```
# crictl rmi -prune
Deleted: mcr.microsoft.com/oss/calico/kube-controllers:v3.23.1
Deleted: mcr.microsoft.com/k8s/flexvolume/blobfuse-flexvolume:1.0.15
[.... a lot of images ....]

# df -m
Filesystem     1M-blocks  Used Available Use% Mounted on
/dev/sda1          29599  4433     25151  15% /
```

Not a viable solution for our CI though...

Issue #5684, opened 2022-05-21
[Bug] Upgrading from ECK `2.1.0` to `2.2.0` Causes issues with Kibana and Fleet during Rolling Restart


BenB196:
## Bug Report
**What did you do?**

I upgraded ECK from `2.1.0` to `2.2.0`

**What did you expect to see?**

I expected a rolling upgrade to happen (related #5648).

**What did you see instead? Under which circumstances?**

I saw a rolling upgrade, but I saw Kibana and Fleet stuck in crash loops for a good duration of the rolling restart time.

**Environment**

* ECK version: `2.1.0` -> `2.2.0`

* Kubernetes information:

	* On premise ? yes
	* Cloud: GKE / EKS / AKS ? n/a
	* Kubernetes distribution: Openshift / Rancher / PKS ? Rancher RKE2 `v1.22.8+rke2r1`

* Resource definition:
Setup a "large" cluster across multiple availability zones (AZs), and the ClusterIP service point to a subset of nodes across the different AZs.

* Logs:

Relevant Kibana Log
```
License information could not be obtained from Elasticsearch due to {"error":{"root_cause":[{"type":"security_exception","reason":"failed to authenticate service account [elastic/kibana] with token name [kibana-prod_kibana-prod_3de51e32-e911-41a2-acd0-c74242b32687]","header":{"WWW-Authenticate":["Basic realm=\"security\" charset=\"UTF-8\"","Bearer realm=\"security\"","ApiKey"]}}],"type":"security_exception","reason":"failed to authenticate service account [elastic/kibana] with token name [kibana-prod_kibana-prod_3de51e32-e911-41a2-acd0-c74242b32687]","header":{"WWW-Authenticate":["Basic realm=\"security\" charset=\"UTF-8\"","Bearer realm=\"security\"","ApiKey"]}},"status":401} error
```

## Issue/Bug:

It appears that with the switch to service accounts for Kibana and Fleet server (#5468) an issue that happens is that it can take a decent amount of time in large clusters to be to have all of the ingress nodes in the cluster be available with the new service accounts to be able to successfully authenticate.

In large clusters where a rolling restart can take hours, this can leave Kibana and Fleet unusable for some time.

## Steps to Reproduce:

(Note: I'm providing a setup similar to mine as I know its reproducible with it, but it might also be reproducible with a smaller setup.

1. Have ECK operation on `2.1.0`
2. Have a relatively large cluster (something that takes a decent amount of time to restart).
[...]

PR #5830, opened 2022-06-24
Do not use service accounts until Elasticsearch nodes have been upgraded

barkbay:
This PR is a ~very early draft~ proposal to mitigate https://github.com/elastic/cloud-on-k8s/issues/5684

With this PR the Elasticsearch controller adds an orchestration hint on the Elasticsearch resource to let the association controllers know when service accounts can be used to authenticate applications.

Note that there is still a small disruption when the client (Kibana or Fleet Server) is restarted as only one authentication mechanism is available at a given point in time. The disruption should, however, be much smaller, for Kibana it should be similar to what happen during version upgrade: https://github.com/elastic/cloud-on-k8s/blob/e5e279c561ae174fe3c9484f61b5e23b61756a4b/pkg/controller/kibana/driver.go#L207-L209


barkbay:
Note that this PR should rollback the situation where the "application user" has been replaced by a service account token, but Elasticsearch is not fully reconciled yet (for example because of health being `yellow`), making Kibana available again.

pkg/utils/optional/Bool.go
```diff
@@ -0,0 +1,71 @@
...
+		return NewBool(b.value || other.value)
+	}
+	if b == nil {
+		return other
+	}
+	if other == nil {
+		return b
+	}
+	return nil
+}
```
pebrc, line 71: Maybe add a few unit tests

pkg/controller/elasticsearch/driver/nodes.go
```diff
@@ -212,13 +214,25 @@ func (d *defaultDriver) reconcileNodeSpecs(
...
 	// as of 7.15.2 with node shutdown we do not need transient settings anymore and in fact want to remove any left-overs.
-	if esReachable && d.isNodeSpecsReconciled(actualStatefulSets, d.Client, results) {
+	if esReachable && isNodeSpecsReconciled {
 		if err := d.maybeRemoveTransientSettings(ctx, esClient); err != nil {
 			return results.WithError(err)
 		}
 	}
 
+	// Set or update an orchestration hint to let the association controller know of service account are supported.
+	if isNodeSpecsReconciled {
```
pebrc, line 226: This looks out of place are we not setting the orchestration hint in `driver.go`?

barkbay:
> I take it that you would have addressed the few nits I found yourself given that you said this is an early draft.

I mentioned it was an "early draft" because I didn't do any manual test before submitting the PR 😄  After a few promising tests (upgrades from `2.1` and `2.2`) I decided to remove the draft status to have additional opinions.

pkg/controller/association/reconciler.go
```diff
@@ -330,10 +331,23 @@ func (r *Reconciler) reconcileAssociation(ctx context.Context, association commo
...
+	// Detect if we should use a service account.
+	var esHints hints.OrchestrationsHints
 	if len(serviceAccount) > 0 {
+		// We must first ensure that the relevant orchestration hint is set on the Elasticsearch cluster.
+		esHints, err = hints.NewFrom(es)
+		if err != nil {
+			return commonv1.AssociationPending, err
+		}
+		if !esHints.ServiceAccounts.IsSet() {
+			r.log(k8s.ExtractNamespacedName(association)).Info("Waiting for Elasticsearch to report if service accounts are supported")
```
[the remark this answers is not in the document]
barkbay: > IIUC we know they are supported based on the version and this is about the condition that they are deployed on all nodes?

Yes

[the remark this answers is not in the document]
barkbay: > IIUC we know they are supported based on the version and this is about the condition that they are deployed on all nodes?

It means that all the Elasticsearch nodes can read the tokens in `conf/service_tokens`, not that the token for the application have been generated and are actually available yet.

pkg/controller/association/reconciler.go
```diff
@@ -330,10 +331,23 @@ func (r *Reconciler) reconcileAssociation(ctx context.Context, association commo
...
+		}
+		if !esHints.ServiceAccounts.IsSet() {
+			r.log(k8s.ExtractNamespacedName(association)).Info("Waiting for Elasticsearch to report if service accounts are fully rolled out")
+			return commonv1.AssociationPending, nil
+		}
+	}
+
+	// If it is the case create the related Secrets and update the association configuration on the associated resource.
+	assocLabels := r.AssociationResourceLabels(k8s.ExtractNamespacedName(association.Associated()), assocRef.NamespacedName())
+	if len(serviceAccount) > 0 && esHints.ServiceAccounts.IsTrue() {
```
pebrc, line 350: Should this not be 
```diff
-       if len(serviceAccount) > 0 && esHints.ServiceAccounts.IsTrue() {
+       if len(serviceAccount) > 0 && esHints.ServiceAccounts.IsSet() {
                applicationSecretName := secretKey(association, r.ElasticsearchUserCreation.UserSecretSuffix)
                r.log(k8s.ExtractNamespacedName(association)).V(1).Info("Ensure service account exists", "sa", serviceAccount)
                err := ReconcileServiceAccounts(
@@ -367,8 +367,12 @@ func (r *Reconciler) reconcileAssociation(ctx context.Context, association commo
                expectedAssocConf.AuthSecretName = applicationSecretName.Name
                expectedAssocConf.AuthSecretKey = "token"
                expectedAssocConf.IsServiceAccount = true
-               // update the association configuration if necessary
-               return r.updateAssocConf(ctx, expectedAssocConf, association)
+
+               if esHints.ServiceAccounts.IsTrue() {
+                       // update the association configuration if necessary
+                       return r.updateAssocConf(ctx, expectedAssocConf, association)
+               }
+               return commonv1.AssociationPending, nil
        }
```
We can create the service token user immediately. There should be no need to check for the annotation at this point (but it does not hurt either). 

What needs to be predicated on the presence of the hint annotation is the actual update of the association config on the Kibana resource because that will trigger the rolling upgrade of the Kibana Pods. Updating the mounted secret before that with service account token information should be fine as Kibana does not currently hot reload its config (there is of course an edge case here if an unplanned restart of Kibana happens in while we are waiting for Elasticsearch to roll out the token)

pkg/controller/elasticsearch/driver/driver.go
```diff
@@ -359,6 +367,112 @@ func (d *defaultDriver) newElasticsearchClient(
...
+	if err != nil {
+		return err
+	}
+	if !supportServiceAccounts {
+		d.ReconcileState.UpdateOrchestrationHints(
+			d.ReconcileState.OrchestrationHints().Merge(hints.OrchestrationsHints{ServiceAccounts: optional.NewBool(false)}),
+		)
+		return nil
+	}
+
```
pebrc, line 409: Should this be: 

```diff
        if !supportServiceAccounts {
                d.ReconcileState.UpdateOrchestrationHints(
-                       d.ReconcileState.OrchestrationHints().Merge(hints.OrchestrationsHints{ServiceAccounts: optional.NewBool(false)}),
+                       d.ReconcileState.OrchestrationHints().Merge(hints.OrchestrationsHints{ServiceAccounts: nil}),
                )
                return nil
        }
 
+       d.ReconcileState.UpdateOrchestrationHints(d.ReconcileState.OrchestrationHints().Merge(hints.OrchestrationsHints{
+               ServiceAccounts: optional.NewBool(false),
+       }))
+
```

So basically saying: we interpret the orchestration hint as follows: 
* `nil` i.e. annotation absent. ES is running an older version that does not support service accounts or the operator version is older 
* `false` there are service accounts configured but they are not observable yet on all nodes via the `_security` API
* `true` there are service accounts and the token are reported from all nodes

pkg/controller/elasticsearch/driver/nodes.go
pebrc: I think this whole block should go away. It is true that `isNodeSpecReconciled` is kind of a proxy for "all nodes should have the service account token now" but it is not as accurate as observing it through the `_security` API which is what we are doing anyway.

pebrc (commented):
I did some more tests and maybe I misunderstood your intention with the `service_accounts` hint. I left some more comments to explain how I understood it.

pkg/controller/elasticsearch/driver/nodes.go
pebrc: I think finally start to grasp your idea. You are waiting for the symlink to be present even before rolling out any service account tokens!

pkg/controller/association/reconciler.go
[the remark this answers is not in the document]
pebrc: Yes! I misunderstood the implementation. Please disregard my comment. Maybe we can make this explicit in a comment somewhere and explain the connection to our symlinking and what the empty `service_tokens` file enables.

pebrc (approved):
I took a couple of wrong turns when reviewing this 😄 Sorry for spamming this PR a bit with comments. But now that I feel I have finally understood the subtleties of the implementation I think it is looking really good and would definitely unblock the problem #5684 describes. 

There might be potential for a further optimisation because we now have a way to observe which tokens Elasticsearch sees on the file system. We could build it so that even if Kubernetes were for some reason too slow to propagate the filesystem changes to all nodes we would still keep Kibana/Fleet server available. IIUC right now once we symlinked the tokens file into the Elasticsearch config directory we start rolling out the tokens controlled by the association controller. We then immediately rotate Kibana with the new credentials. If Kubernetes takes its time to propagate the tokens into the Pods filesystem Kibana might still be down. Because you introduced a three-valued boolean we could have it so that the absence of the hint means we are not ready for tokens. `false` could mean we are ready to roll out the tokens i.e we have the symlink but they are either not created yet or not observable yet and `true` would mean we have completed the rollout. 

But given that we did not have this functionality for traditional user/password accounts either and it does not seem to be a problem in practice, that might be going too far given that we under certain time constraints with this bug fix.

pkg/controller/association/reconciler.go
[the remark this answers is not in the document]
barkbay: > Maybe we can make this explicit in a comment somewhere and explain the connection to our symlinking and what the empty `service_tokens` file enables.

I added a comment here: https://github.com/elastic/cloud-on-k8s/pull/5830/files#diff-e8f049d805085e8bf5034827dc84fab0a658451828eca4aa65da40fbd214ae37R21-R27 Let me know if it helps.

pkg/controller/elasticsearch/driver/nodes.go
```diff
@@ -212,13 +214,25 @@ func (d *defaultDriver) reconcileNodeSpecs(
...
+		allNodesRunningServiceAccounts, err := esv1.AreServiceAccountsSupported(d.ES.Spec.Version)
```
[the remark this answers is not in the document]
barkbay: https://github.com/elastic/cloud-on-k8s/pull/5830/commits/d3a27a6b48b662b00152864ee179d493568c3335

barkbay:
@pebrc I should have added more comments to explain the idea, sorry about that. May I ask you a last review on the comments I just added?
Thanks 🙏

pebrc (approved):
Still LGTM. I also did a few more manual tests with different types of clusters in different states. Partially migrated to service account tokens, combing and operator upgrade with a version upgrade that introduces tokens, and so forth and all looked good.

PR #5836, opened 2022-06-28
Improve "Operator crashes on startup with `OOMKilled`" docs section

ppf2:
Add an example of an OOMKilled event and share best practice on setting memory limits: recommend setting `spec.containers[].resources.limits.memory` and `spec.containers[].resources.requests.memory` to the same value.

cc: @kunisen

### DIFF
--- a/docs/operating-eck/troubleshooting/common-problems.asciidoc
+++ b/docs/operating-eck/troubleshooting/common-problems.asciidoc
@@ -57,7 +57,7 @@ The default link:https://kubernetes.io/docs/concepts/configuration/manage-resour
 kubectl patch sts elastic-operator -n elastic-system -p '{"spec":{"template":{"spec":{"containers":[{"name":"manager", "resources":{"limits":{"memory":"2Gi"}}}]}}}}'
 ----
 
-NOTE: It is a best practice to set `spec.containers[].resources.limits.memory` and `spec.containers[].resources.requests.memory` to the same value.
+NOTE: It is recommended to set limits (`spec.containers[].resources.limits`) that match requests (`spec.containers[].resources.requests`) in order to prevent operator's Pod from being terminated during link:https://kubernetes.io/docs/concepts/scheduling-eviction/node-pressure-eviction/[node-pressure eviction].
 
 [id="{p}-{page_id}-webhook-timeout"]
 == Timeout when submitting a resource manifest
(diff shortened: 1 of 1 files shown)


barkbay:
> May I ask your opinion to see if we should also explicitly set that for the operator too, please? @pebrc

For the JVM it is indeed a best practice because:
* The JVM sets the Heap size according to the `limit`, not the `request`
* Elasticsearch is started with [`-XX:+AlwaysPreTouch`](https://github.com/elastic/elasticsearch/blob/master/distribution/tools/server-cli/src/main/java/org/elasticsearch/server/cli/SystemJvmOptions.java#L30-L31)

Therefore, memory pages of an Elasticsearch Pod are immediately mapped, up to the `limit`. Said differently, if the `request` is lower than the `limit` then an Elasticsearch Pod is _immediately_ a [candidate for eviction](https://kubernetes.io/docs/concepts/scheduling-eviction/node-pressure-eviction/#pod-selection-for-kubelet-eviction), a soon as it is started.

The situation is a bit different for the operator since it is not running in a virtual machine, with a fixed amount of memory mapped. There may be memory usage spikes, but memory may eventually be returned to the operating system. Hence, having different values might make sense. That being said, we can indeed remind the user that to avoid any disruption it is recommend to:
* set the `request` and the `limit` to a same value (or just set the `limit` without any `request`)
* monitor the RSS memory to adjust the allocated resources

docs/operating-eck/troubleshooting/common-problems.asciidoc
```diff
@@ -19,6 +50,7 @@ The default link:https://kubernetes.io/docs/concepts/configuration/manage-resour
 kubectl patch sts elastic-operator -n elastic-system -p '{"spec":{"template":{"spec":{"containers":[{"name":"manager", "resources":{"limits":{"memory":"2Gi"}}}]}}}}'
 ----
 
+NOTE: It is a best practice to set `spec.containers[].resources.limits.memory` and `spec.containers[].resources.requests.memory` to the same value.
```
barkbay, line 53:
```suggestion
NOTE: It is recommended to set limits (`spec.containers[].resources.limits`) that match requests (`spec.containers[].resources.requests`) in order to prevent operator's Pod from being terminated during link:https://kubernetes.io/docs/concepts/scheduling-eviction/node-pressure-eviction/[node-pressure eviction].
```

ppf2:
thx for the suggestions!

barkbay (approved):
LGTM

alaudazzi (approved):
Left a minor editing suggestion. Otherwise LGTM.

Issue #5779, opened 2022-06-15
Should the operator create enrolment tokens instead of configuring Agents to talk to Kibana


pebrc:
The current implementation of Elastic Agent with the ECK operator configures Agents to enrol themselves in Fleet by providing them with connection information to talk to Kibana themselves: 

https://www.elastic.co/guide/en/fleet/master/running-on-kubernetes-managed-by-fleet.html#_settings

```
- name: KIBANA_HOST
  value: ""
- name: KIBANA_FLEET_USERNAME
  value: ""
- name: KIBANA_FLEET_PASSWORD
  value: ""
```
That is not ideal as individual Agents have access to the Kibana API with more privileges than strictly necessary or more privileges than they would have if they were configured directly with an enrolment API key instead. 

The question is whether we should change the implementation so that: 
* the operator creates enrolment API keys for the Agents by talking to the Kibana API (a complication here is that we need to pick a policy in order to create such a key)
* the operator then creates the Agent deployment/daemonset with this API key obviating the need for a direct Kibana connection from individual Agents IIUC 

The Fleet API is not stable/internal but documented here: 
https://raw.githubusercontent.com/elastic/kibana/master/x-pack/plugins/fleet/common/openapi/bundled.json

pebrc:
We got some feedback from the Fleet team that enrolment token are indeed the preferred way. I think we should strive to refactor our Agent support accordingly before we remove the "experimental" flag.

pebrc:
We could 
* introduce a new field in the CRD called `agentPolicy` 
* if the field is populated by the user with an Agent policy name or ID we use that policy to enrol the agent/fleet server
* if the field is not set we call the `/api/fleet/agent_polices/`  endpoint in Kibana which produces output similar to this:
```
{
  "items": [
    {
      "id": "eck-agent",
      "namespace": "default",
      "monitoring_enabled": [
        "logs",
        "metrics"
      ],
      "is_default": true,
      "name": "Elastic Agent on ECK policy",
      "unenroll_timeout": 900,
      "is_preconfigured": true,
      "status": "active",
      "is_managed": false,
      "revision": 3,
      "updated_at": "2022-06-22T14:39:09.683Z",
      "updated_by": "system",
      "package_policies": [
        "b58bd4d8-3fa4-54a1-8776-a733732c8a3d",
        "a3b718bd-efec-54f4-b513-7711c744a8ec"
      ],
      "agents": 0
    },
    {
[...]

PR #5846, opened 2022-06-29
Create agent enrollment tokens in the operator 

pebrc:
Fixes https://github.com/elastic/cloud-on-k8s/issues/5779

See issue for detail on the approach taken. But to repeat the gist: do not expose highly privileged Kibana credentials to Elastic Agent Pods. Instead make the operator do all the Kibana API interactions with the more limited user (possible as of 8.1) and expose only the enrolment token to the Elastic Agent as it would be also the case on a typical bare metal installation. 

Tradeoffs:
*  Agent rollout needs to wait until Kibana is available (but in the old system the Agents would just crash loop until that was the case so I consider this actually an improvement) 
* Operator needs to stay up to date with Fleet API development. So far the relevant APIs have been fairly stable with one API being renamed for consistency. 
* Each Kibana association creates a Kibana user. Technically we could use a single user to do all the API interactions with Kibana but that would have broken out of the association mechanism and it seemed convenient and correct to keep using that. 

Some additional choices I made that are maybe of interest to reviewers: I am not reusing the default enrolment token but creating a new one. No particular strong reason for that. It allows us to clean the tokens up if something goes wrong and also allows us to run without explicitly calling `/api/fleet/setup` (which creates the default tokens). In practice I ended up calling the setup API anyway because we cannot rely on users configuring Kibana with pre-defined policies in `kibana.yml` (as do all our recipes). So if there are strong feelings against creating our own tokens we can revisit this choice.

I hope I did not miss anything fundamental. I did manual tests with `7.14` (first version of Fleet we support), `7.17` and a recent `8.x`.


pebrc:
run/e2e-tests tags=agent

pebrc:
run/e2e-tests tags=agent

pebrc:
Jenkins test this please

pkg/controller/agent/fleet.go
```diff
@@ -0,0 +1,308 @@
...
+		log:           logger,
+	}
+}
+
+func (f fleetAPI) request(
+	ctx context.Context,
+	method string,
+	pathWithQuery string,
+	requestObj,
+	responseObj interface{}) error {
```
david-kow, line 86:
```suggestion
	requestObj, responseObj interface{},
) error {
```

david-kow (approved):
LGTM! One thing I'd like to clarify is what happens if the user sets `FLEET_SERVER_POLICY_ID` env var separately from `policyID`. Should we document the precedence?

pkg/controller/agent/fleet.go
```diff
@@ -0,0 +1,307 @@
...
+func isKibanaReachable(ctx context.Context, client k8s.Client, kibanaNSN types.NamespacedName) (bool, error) {
+	var kb v1.Kibana
+	err := client.Get(ctx, kibanaNSN, &kb)
+	if err != nil {
+		return false, err
+	}
+	if kb.Status.Health != commonv1.GreenHealth {
+		return false, nil // requeue
+	}
+	return true, nil
```
pebrc, line 242: I realise this does not work well with external KibanaRefs. I will look into making an actuall HTTP request instead here.

pebrc: It turns out this is a rabbit hole. Making external refs work is more complicated. Other things required: communication between Fleet Server and Elasticsearch needs to use a service token otherwise Agent will try to talk to Kibana to generate one. Also I ran into some weird connectivity issues where Agent seemed to connect to an IP directly without host header and failed to connect through the ESS proxy. Still looking into that, but I might punt the external ref support to another PR if it makes sense at all.

pebrc:
run/e2e-tests tags=agent

PR #5850, opened 2022-07-04
Use static transaction names for APM

### LINKED ISSUE #5840: Use common name for ECK APM transactions
The name of the APM transaction is taken from the namespace/name of the reconciled resource. This causes transactions for different ES clusters to be grouped separately:
Screenshot 2022-06-28 at 09 14 32

I believe we should change it (or make it configurable) to provide a constant transaction name for each of the controllers. This way, we would be able to group transactions together in the APM UI. If we want to retain an option to filter by application (ES, KB, etc.) name, we can introduce label(s) for name/namespace.

pebrc:
Fixes #5840 

Instead of using names derived from the name of the resource being reconciled, use static names that represent the individual controllers. The idea is to facilitate comparability between similar transactions rather than creating an ever increasing number of transaction names to aggregate over. 

Also standardised the transaction types to three options: `periodic`, `run-once` and `reconciliation` to represent the three main types of work being done in the operator. `reconciliation` is the same term @david-kow has been using in his recent work.  Hope that make sense.

### DIFF
--- a/pkg/controller/common/tracing/transaction.go
+++ b/pkg/controller/common/tracing/transaction.go
@@ -10,35 +10,22 @@ import (
 	"go.elastic.co/apm/v2"
 )
 
-const ReconciliationTxType = "reconciliation"
-const PeriodicTxType = "periodic"
-const RunOnceTxType = "run-once"
+type TxType string
 
-// NewTransaction starts a new transaction and sets up a new context with that transaction that also contains the related
-// APM agent's tracer.
-func NewTransaction(ctx context.Context, t *apm.Tracer, txName string) (*apm.Transaction, context.Context) {
-	if t == nil {
-		return nil, ctx // apm turned off
-	}
-	tx := t.StartTransaction(txName, ReconciliationTxType)
-	return tx, apm.ContextWithTransaction(ctx, tx)
-}
-
-// EndTransaction nil safe version of APM agents tx.End()
-func EndTransaction(tx *apm.Transaction) {
-	if tx != nil {
-		tx.End()
-	}
-}
+const (
+	ReconciliationTxType TxType = "reconciliation"
+	PeriodicTxType       TxType = "periodic"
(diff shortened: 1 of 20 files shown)


pkg/controller/common/tracing/transaction.go
```diff
@@ -8,16 +8,19 @@ import (
 	"context"
 
 	"go.elastic.co/apm/v2"
-	"k8s.io/apimachinery/pkg/types"
 )
 
+const ReconciliationTxType = "reconciliation"
+const PeriodicTxType = "periodic"
+const RunOnceTxType = "run-once"
```
thbkrkr, line 15: Nit:
```suggestion
const (
	ReconciliationTxType = "reconciliation"
	PeriodicTxType       = "periodic"
	RunOnceTxType        = "run-once"
)
```

pkg/controller/common/tracing/transaction.go
```diff
@@ -8,16 +8,19 @@ import (
...
+
 // NewTransaction starts a new transaction and sets up a new context with that transaction that also contains the related
 // APM agent's tracer.
-func NewTransaction(ctx context.Context, t *apm.Tracer, name types.NamespacedName, txType string) (*apm.Transaction, context.Context) {
+func NewTransaction(ctx context.Context, t *apm.Tracer, txName string) (*apm.Transaction, context.Context) {
```
thbkrkr, line 19: Should we be more explicit?
```suggestion
func NewReconciliationTransaction(ctx context.Context, t *apm.Tracer, txName string) (*apm.Transaction, context.Context) {
```

pkg/controller/common/tracing/transaction.go
thbkrkr: Have you considered using a custom `TxType` type to avoid that we can invert the `txType` and the `txName` which are both of type `string`?

david-kow (approved):
LGTM. Thanks for working on this.

PR #5860, opened 2022-07-06
handle io copy error

panguicai008:

### DIFF
--- a/hack/operatorhub/main.go
+++ b/hack/operatorhub/main.go
@@ -218,7 +218,9 @@ func makeRequest(url string) (io.Reader, error) {
 
 	buf := new(bytes.Buffer)
 	_, err = io.Copy(buf, resp.Body)
-
+	if err != nil {
+		return nil, fmt.Errorf("failed to copy response body to buffer: %w", err)
+	}
 	return buf, err
 }
 


barkbay (commented):
Thanks for the PR.

IMHO error wrapping should be used to add some context to an error. I'm not sure that it is the case in your proposal, `failed to copy response body to buffer` feels a bit generic and I don't think it would help to understand the root cause of the error.

I think I would stick with the current implementation.

panguicai008:
> Thanks for the PR.
> 
> IMHO error wrapping should be used to add some context to an error. I'm not sure that it is the case in your proposal, `failed to copy response body to buffer` feels a bit generic and I don't think it would help to understand the root cause of the error.
> 
> I think I would stick with the current implementation.

I want to deal with this error. The `makeRequest` function caller uses error wrapping, so I should output error directly? I can replace `%w` with`%v`, do you think so?

panguicai008:
I just think that when an error may appear in a function, it should be handled. Like in other open source projects, this `io.Copy` error will also be polished. I don't pay much attention to what happens when error occurs. In addition, the purpose of reading this code is to generate bundle image from various versions of CSV, build catalog image, and prepare for privatization deployment without using the official operator hub

thbkrkr:
Closing because the proposed change to wrap the error does not provide any real improvement over the current implementation.

PR #5861, opened 2022-07-06
Add new operator flag to control Elasticsearch health observation intervals

### LINKED ISSUE #5839: Expose health check interval as operator-level configuration option
Today, ECK allows to override the default, 10s interval through annotation on ES cluster. It would be helpful to enable this as a operator level config option.

pebrc:
Fixes #5839 

Introduce new global flag to conrol Elasticsearch health observation interval. 
* annotations on individual Elasticsearch resources take precedence to avoid breaking existing customisations for users
* a non-positve value disables asynchronous observation completely. Only one synchronous observation happens during reconciliation. This disables also timely automatic pod disruption budget adjustment on health changes. As a side effect of client-go cache refreshes observations still happen every 10 hours due to reconciliation. This means disabling the asynchronous observation has the same effect as setting the observation interval to 10 hours.


thbkrkr:
I understand that we need to do something if the provided interval is negative but I'm not sure that we should disable the observer. What does it mean to manage an ES for ECK without an observer? Isnt it dangerous to reconcile ES without knowing the real health?
How about validating that the provided interval is in a range that makes sense (`5s < i < 1h`) and crashing the operator if it's not.

pkg/controller/elasticsearch/observer/observer.go
```diff
@@ -26,9 +26,9 @@ type Settings struct {
 	Tracer              *apm.Tracer
 }
 
-// defaultObservationInterval is the default interval of observation.
-// if the Elasticsearch cluster is unavailable, the actual interval would be observationInterval + requestTimeout.
-const defaultObservationInterval = 10 * time.Second
+// defaultObservationTimeout is the default timeout for an observation. The observer uses the observation interval as a timeout
+// the default applies if the observation interval is not positive to allow at least one successful observation.
```
thbkrkr, line 30:
```suggestion
// defaultObservationTimeout is the default timeout for an observation. The observer uses the observation interval as a timeout.
// The default applies if the observation interval is not positive to allow at least one successful observation.
```

pkg/controller/elasticsearch/observer/observer.go
```diff
@@ -26,9 +26,9 @@ type Settings struct {
...
+const defaultObservationTimeout = 10 * time.Second
```
thbkrkr, line 31: Another way to make it explicit that we use the interval for the timeout?
```suggestion
const (
	DefaultInterval = 10 * time.Second
	defaultimeout  = DefaultInterval
)
```

thbkrkr (commented):
The behaviour looks good to me.
I left some minor comments on names and constants.

pkg/controller/elasticsearch/observer/observer.go
pebrc: I was referring to this bit here https://github.com/elastic/cloud-on-k8s/blob/2033eaa17f277bba0d19eb74ea0c552f6a396069/pkg/controller/elasticsearch/observer/observer.go#L112

Timeout and default interval are independant in principle. I think the reason why by default the interval is used is that we want to avoid a situation where multiple requests are in flight.

cmd/manager/main.go
```diff
@@ -192,6 +192,11 @@ func Command() *cobra.Command {
 		3*time.Minute,
 		"Default timeout for requests made by the Elasticsearch client.",
 	)
+	cmd.Flags().Duration(
+		operator.ElasticsearchObservationIntervalFlag,
+		10*time.Second,
```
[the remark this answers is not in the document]
thbkrkr: The intention was to tie the default interval to the default timeout but this doesn't make sense, I didn't think very far.

thbkrkr (approved):

Issue #5853, opened 2022-07-05
[OperatorHub] ClusterServiceVersion: set image digest instead of tag


barkbay:
RedHat certification process expects the image digest instead of the tag in the `ClusterServiceVersion`.
For example in the case of ECK 2.3:

```diff
--- a/operators/elasticsearch-eck-operator-certified/2.3.0/manifests/elasticsearch-eck-operator-certified.clusterserviceversion.yaml
+++ b/operators/elasticsearch-eck-operator-certified/2.3.0/manifests/elasticsearch-eck-operator-certified.clusterserviceversion.yaml
@@ -5,7 +5,7 @@ metadata:
     capabilities: Full Lifecycle
     categories: Database
     certified: 'false'
-    containerImage: registry.connect.redhat.com/elastic/eck-operator-ubi8:2.3.0
+    containerImage: registry.connect.redhat.com/elastic/eck-operator@sha256:3ed9323e46ccbdd932a259bb0dd9f04010ec842b4cc7b4c4e7ee889760a90c1d
     createdAt: 2022-07-05 07:37:33
     description: Run Elasticsearch, Kibana, APM Server, Beats, Enterprise Search, Elastic Agent and Elastic Maps Server on Kubernetes and OpenShift
     repository: https://github.com/elastic/cloud-on-k8s
@@ -346,7 +346,7 @@ spec:
             spec:
               serviceAccountName: elastic-operator
               containers:
-              - image: registry.connect.redhat.com/elastic/eck-operator-ubi8:2.3.0
+              - image: registry.connect.redhat.com/elastic/eck-operator@sha256:3ed9323e46ccbdd932a259bb0dd9f04010ec842b4cc7b4c4e7ee889760a90c1d
                 name: manager
                 args:
                   - "manager"
@@ -365,7 +365,7 @@ spec:
                     fieldRef:
                       fieldPath: metadata.annotations['olm.operatorNamespace']
                 - name: OPERATOR_IMAGE
-                  value: registry.connect.redhat.com/elastic/eck-operator-ubi8:2.3.0
+                  value: registry.connect.redhat.com/elastic/eck-operator@sha256:3ed9323e46ccbdd932a259bb0dd9f04010ec842b4cc7b4c4e7ee889760a90c1d
                 resources:
                   limits:
                     cpu: 1
```

PR #5863, opened 2022-07-06
Certified operator: get certified operator image digest from RedHat catalog

barkbay:
Fixes https://github.com/elastic/cloud-on-k8s/issues/5853

This PR assumes that the API key and the project ID are provided as environment variables by the release job.

### DIFF
--- a/hack/operatorhub/config.yaml
+++ b/hack/operatorhub/config.yaml
@@ -40,6 +40,6 @@ packages:
     operatorRepo: registry.connect.redhat.com/elastic/eck-operator
     ubiOnly: true
     ## digestPinning should only be set to true for certified operator.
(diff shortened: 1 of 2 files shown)


hack/operatorhub/catalog.go
```diff
@@ -0,0 +1,76 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
```
pebrc, line 1: I think if the main package is split into multiple source files it cannot be run simply with `go run main.go` which is what is documented in the README and what the build automation will try to do.

pebrc (commented):
I was not able to succesfully test this, I always got a 404. But that could totally be me being too stupid to figure our the right project id.

barkbay:
Last commits changed the env. var name to `REDHAT_API_KEY` (instead of `REDHAT_API_TOKEN`)

pebrc (approved):
LGTM

Issue #5865, opened 2022-07-07
TestMutationSecondMasterSetDown is flaky


barkbay:
`TestMutationSecondMasterSetDown` failed 4 times today:

* [cloud-on-k8s-e2e-tests-gke-k8s-versions](https://devops-ci.elastic.co/job/cloud-on-k8s-e2e-tests-gke-k8s-versions/770/testReport/) on GKE `1.20` and `1.22`
* [cloud-on-k8s-e2e-tests-kind-k8s-versions](https://devops-ci.elastic.co/job/cloud-on-k8s-e2e-tests-kind-k8s-versions/790/testReport/) on Kind `1.23.6 IPV6`
* [cloud-on-k8s-e2e-tests-main](https://devops-ci.elastic.co/job/cloud-on-k8s-e2e-tests-main/384/testReport/)

```
=== RUN   TestMutationSecondMasterSetDown/Elasticsearch_cluster_health_should_not_have_been_red_during_mutation_process
    steps_mutation.go:150: Elasticsearch cluster health check failure at 2022-07-06 18:12:25.389331963 +0000 UTC m=+5386.055151684: elasticsearch client failed for https://test-mutation-2nd-master-set-mqcr-es-internal-http.e2e-f79bs-mercury.svc:9200/_cluster/health: Get "[https://test-mutation-2nd-master-set-mqcr-es-internal-http.e2e-f79bs-mercury.svc:9200/_cluster/health":](https://test-mutation-2nd-master-set-mqcr-es-internal-http.e2e-f79bs-mercury.svc:9200/_cluster/health) context deadline exceeded
    steps_mutation.go:152: 
        	Error Trace:	steps_mutation.go:152
        	Error:      	Not equal: 
        	            	expected: 0
        	            	actual  : 1
        	Test:       	TestMutationSecondMasterSetDown/Elasticsearch_cluster_health_should_not_have_been_red_during_mutation_process
```

barkbay:
It also failed yesterday https://devops-ci.elastic.co/job/cloud-on-k8s-e2e-tests-main/384/testReport/
The most notable recent change is #5844

pebrc:
The request tracing added in https://github.com/elastic/cloud-on-k8s/pull/5869 gives some additional clues.

[...]

PR #5871, opened 2022-07-12
Collect ES stack traces and report all health check errors

pebrc:
During the investigation of https://github.com/elastic/cloud-on-k8s/issues/5865 I saw a few of these 4xx errors that would like to better understand: 

```
[23][null][BAD_REQUEST][application/json;charset=utf-8][351] sent response to [Netty4HttpChannel{localAddress=/10.244.3.206:9200, remoteAddress=/10.244.2.3:46366}] success [true]
```
These are health check responses to our e2e test Pod. I would like to better understand what is going on there. Also `error_trace` is a good idea in general for tests and dev mode.


pkg/controller/elasticsearch/client/base.go
```diff
@@ -129,7 +131,18 @@ func (c *baseClient) request(
...
+		if err != nil {
+			return err
+		}
+		q := url.Query()
+		q.Set("error_trace", "true")
+		url.RawQuery = q.Encode()
+		rawURL = url.String()
+	}
+	request, err := http.NewRequest(method, rawURL, body) //nolint:noctx
 	if err != nil {
```
naemono, line 146: Why don't you wait to check to see if debug is enabled until after you generate the request? Then you won't have to parse the URL and have that additional error logic.  Something like

```
q := request.URL.Query()
q.Add("error_trace", "true")
request.URL.RawQuery = q.Encode()
```

test/e2e/test/elasticsearch/steps_mutation_test.go
```diff
@@ -22,13 +23,13 @@ func Test_clusterUnavailability(t *testing.T) {
 	require.False(t, cu.hasExceededThreshold())
 
 	// mark the cluster as available, we're still below the threshold
-	cu.markUnavailable()
+	cu.markUnavailable(errors.New("Connection refused"))
```
thbkrkr, line 26: It doesn't matter here but error strings should not be capitalized (unless they start with proper names or acronyms). To keep the habit?

```suggestion
	cu.markUnavailable(errors.New("connection refused"))
```

test/e2e/test/elasticsearch/steps_mutation_test.go
```diff
@@ -22,13 +23,13 @@ func Test_clusterUnavailability(t *testing.T) {
...
 	require.False(t, cu.start.IsZero())
 	require.False(t, cu.hasExceededThreshold())
 
 	// marking as unavailable again should not change the start time
 	initialStartTime := cu.start
-	cu.markUnavailable()
+	cu.markUnavailable(errors.New("Context deadline exceeded"))
```
thbkrkr, line 32:
```suggestion
	cu.markUnavailable(errors.New("context deadline exceeded"))
```

test/e2e/test/elasticsearch/steps_mutation_test.go
```diff
@@ -38,7 +39,7 @@ func Test_clusterUnavailability(t *testing.T) {
 	require.False(t, cu.hasExceededThreshold())
 
 	// simulate a lower threshold we should have exceeded
-	cu.markUnavailable()
+	cu.markUnavailable(errors.New("Something else"))
```
thbkrkr, line 42:
```suggestion
	cu.markUnavailable(errors.New("something else"))
```

thbkrkr (approved):
LGTM

naemono (approved):
Was waiting for checks to pass but 👍

PR #5878, opened 2022-07-18
Enable beats stack monitoring configuration

### LINKED ISSUE #5563: enable beats monitoring in ECK
With the [stack monitoring](https://www.elastic.co/guide/en/cloud-on-k8s/2.1/k8s-stack-monitoring.html#k8s-stack-monitoring) enabled for ECK cluster, this simplify the whole stack monitoring config.
 it would be good to have the similar function for beats monitoring enabled. 
Currently have to add the following in beats yaml file:
```
---
apiVersion: [beat.k8s.elastic.co/v1beta1](https://beat.k8s.elastic.co/v1beta1)
kind: Beat
metadata:
  name: aws
  namespace: eck
spec:
  type: filebeat
  version: 7.16.3
  config:
    name: beat-aws
  monitoring:
    enabled: true
    elasticsearch:
      hosts: ["[https://${MONITORING_CLUSTERNAME_ES_HTTP_SERVICE_HOST}:${MONITORING_CLUSTERNAME_ES_HTTP_SERVICE_PORT}]
      username: ${ES_USER}
      password: ${ES_PASSWORD}
      ssl.certificate_authorities: ["/etc/pki/root/tls.crt"]
      ssl.verification_mode: certificate
```

naemono:
Closes #5563

This change enables easy stack monitoring configuration for Beats, such as we already have for both Elasticsearch and Kibana, by adding the following configuration stanza in the Beats CRD

```
kind: Beat
spec:
  monitoring:
    elasticsearchRefs:
    - name: elasticsearch
```

## Notable differences to Elasticsearch/Kibana configuration

With both Elasticsearch and Kibana, `monitoring.metrics` and/or `monitoring.logs` are used to enable stack monitoring

Elasticsearch example:
```
kind: Elasticsearch
spec:
  monitoring:
    metrics:
      elasticsearchRefs:
      - name: metrics-es
    logs:
      elasticsearchRefs:
      - name: logs-es
```

This differs from [Beats internal stack monitoring configuration](https://www.elastic.co/guide/en/beats/filebeat/current/monitoring-internal-collection.html), where a single Elasticsearch clusters is configured, and only metrics monitoring data is sent.


test/e2e/test/beat/checks.go
```diff
@@ -87,3 +89,38 @@ func checkEvent(url string, check func(int) error) ValidationFunc {
 		return check(len(results.Hits.Hits))
 	}
 }
+
+func containsDocuments(esClient client.Client, indexPattern string) error {
```
pebrc, line 93: This is looks like a copy of existing code. I think we should be able to reuse the existing check steps for monitoring in the `checks` package. You just have to expose a new reduced step list for Beats without the side car checks.

pkg/controller/common/stackmon/validations/validations.go
```diff
@@ -19,6 +19,7 @@ const (
 	invalidElasticsearchRefsMsg = "Only one Elasticsearch reference is supported for %s Stack Monitoring"
 
 	InvalidKibanaElasticsearchRefForStackMonitoringMsg = "Kibana must be associated to an Elasticsearch cluster through elasticsearchRef in order to enable monitoring metrics features"
+	InvalidBeatsElasticsearchRefForStackMonitoringMsg  = "Beats must be associated to an Elasticsearch cluster through elasticsearchRef in order to enable monitoring metrics features"
```
pebrc, line 22: I think we would want an actual validation in the webhook/reconciler using this (in addition to the error handling you implemented during config generation)

pebrc (commented):
Had a quick first look, I haven't done any testing yet.

test/e2e/test/beat/checks.go
naemono: Understood.  I've implemented a similar exposed func in the `checks` package that just returns the only step needed for beats.  Lmk if this is what you intended.

pkg/controller/common/stackmon/validations/validations.go
naemono: Ok, additional validation added to the default webhook checks for Beat that validates the monitoring section of the specificiation.  *Note* that these are slightly different than the `common.stackmon.validations`, in that they break the "refs defined", and the length of the refs into 2 separate distinct validations+errors.

https://github.com/elastic/cloud-on-k8s/blob/main/pkg/controller/common/stackmon/validations/validations.go#L44

This doesn't seem right, as if the refs are not defined, you'll never get a validation error:
```
if monitoring.AreEsRefsDefined(refs) && len(refs) != 1 {
```

Should we create an issue, and get the stackmon validations updated as well?

pebrc: Not sure I fully understand. At the top of the common validations there is a `monitoring.IsDefined(resource)` call which is a logical OR or logs or metrics refs defined. So we know at least one is defined, otherwise we would exit the whole validation function. After that it is just a matter of which one or if both are defined and if they are defined correctly.

pebrc (commented):
Almost LGTM. I am having second thoughts though whether we should stick to the common 
```
monitoring:
    metrics:
      elasticsearchRefs:
      - name: metrics-es
    logs:
      elasticsearchRefs:
      - name: logs-es
 ```
schema just to keep the door open for future enhancements that would for example export logs as well and to keep the CRDs somewhat uniform. Can't quite make my mind up.  Maybe @thbkrkr has an opinion as well given that he wrote the original implementation.

File: pkg/apis/beat/v1beta1/validations.go [hunk 3]
```diff
@@ -24,13 +28,16 @@ var (
...
+		checkMonitoring,
 	}
 
 	updateChecks = []func(old, curr *Beat) field.ErrorList{
 		checkNoDowngrade,
 	}
 
 	typeRegex = regexp.MustCompile("^[a-zA-Z0-9-]+$")
+	// minStackVersion is the minimum Stack version to enable Stack Monitoring on an Elastic Beats.
+	minStackVersion = version.MustParse("7.2.0-SNAPSHOT")
```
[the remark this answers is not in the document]
naemono: As for the snapshot question, I was following the convention set out here https://github.com/elastic/cloud-on-k8s/commit/fb8205af22f9af286f1948e98678601a89fb8d83#diff-9f672c3af5fec0acba2a50cec9fa58743687d5f35d3f7e4c5591dcd63bca6dd4R27, and was assuming that this feature was enabled in the beats 7.2.0 snapshots prior to it's release.  The chances of someone running a 7.2.0 snapshot version at this stage is probably > .1%, but I was trying to be consistent with how we were handling this in the stackmon/validations package.

As for the internal monitoring being added in 7.2, I was looking at the [public monitoring documentation](https://www.elastic.co/guide/en/beats/filebeat/7.2/monitoring-internal-collection.html)
and this page doesn't exist prior to 7.2, but if you go back to 7.1 and look at the [configuration reference](https://www.elastic.co/guide/en/beats/filebeat/7.1/filebeat-reference-yml.html) you can see that monitoring existed, but the configuration was a different format:

```
# Uncomment to send the metrics to Elasticsearch. Most settings from the
# Elasticsearch output are accepted here as well. Any setting that is not set is
# automatically inherited from the Elasticsearch output configuration, so if you
# have the Elasticsearch output configured, you can simply uncomment the
# following line, and leave the rest commented out.
#xpack.monitoring.elasticsearch:
```

as opposed to 

```
monitoring:
  enabled: true
  elasticsearch:
```

So it technically existed prior to 7.2, but considering we'd have to support 2 different configuration formats, it seems sane to just support this >= 7.2.

test/e2e/test/checks/monitoring.go
```diff
@@ -51,6 +58,10 @@ func (c stackMonitoringChecks) Steps() test.StepList {
 	}
 }
 
+func (c stackMonitoringChecks) BeatsStep() test.Step {
+	return c.CheckMonitoringMetricsIndex()
```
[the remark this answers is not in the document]
naemono: Completely agree.  Will remove.

pkg/apis/beat/v1beta1/validations.go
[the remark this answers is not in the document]
pebrc: Ok makes sense. I think the snapshot versioning makes sense when we release a new feature at the same time as when the corresponding stack feature is released, so that we can test against pre-release versions.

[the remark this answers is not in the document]
naemono: Removed `-SNAPSHOT`, and added a comment around why 7.2 is the min supported version.

pkg/controller/common/stackmon/validations/validations.go
naemono: I hope this makes it more clear.  The following block only checks if the version is not supported, and only returns early if one is defined AND the version is not supported:
```
if monitoring.IsDefined(resource) {
		err := IsSupportedVersion(version)
		if err != nil {
			finalMinStackVersion, _ := semver.FinalizeVersion(MinStackVersion.String()) // discards prerelease suffix
			errs = append(errs, field.Invalid(field.NewPath("spec").Child("version"), version,
				fmt.Sprintf(UnsupportedVersionMsg, finalMinStackVersion)))
		}
	}
```

`IsDefined` checks the following, so if the monitoring definition just has `Namespace`, then it won't be technically `Defined`.
```
// IsDefined checks if the object selector is not nil and has a name or a secret name.
// Namespace is not mandatory as it may be inherited by the parent object.
func (o *ObjectSelector) IsDefined() bool {
	return o != nil && o.NameOrSecretName() != ""
}
```

Let's say someone does this
```
kind: Elasticsearch
spec:
  version: 8.2.3
  monitoring:
    metrics:
      elasticsearchRefs:
      - namespace: "test",
```

So the `IsSupportedVersion` check will be skipped, as this isn't `Defined` and won't return early, and the following check will be skipped `if monitoring.AreEsRefsDefined(refs) && len(refs) != 1 {` as this monitoring ref isn't `Defined` as well, and the subsequent logs check will be skipped as well, as the logs block isn't defined at all, and this `Validate` function will return no errors, although this clearly isn't a valid monitoring block...

naemono: Forgot to mention, I check this exact situation [here](https://github.com/elastic/cloud-on-k8s/pull/5878/files#diff-e391f05ef44e6228796152fb00ef6b2ff0c2766f2d330bb0e4e979b759b951e1R258) for beats monitoring

pebrc (changes requested):
I have thought about it some more and asked for opinions from others and I think now we should indeed stick with the same API schema we have for Kibana and Elasticsearch. The reasons are as originally stated: to be consistent across CRDs and to be able to add log extraction through Filebeat. It would be ok to do the log bit in a separate PR though if you prefer. It might also be good to reach out to the Observability team to get their take on the preferred way to do Beats monitoring and log extraction given that this leads to a kind of Russian doll problem where we add more Beats eg. Filebeat to monitor another Beat, which then raises the question where does one stop (but we already have that problem with the monitoring of Kibana and ES so maybe that is ok)

naemono:
@pebrc I've reached out to observability, and they also suggested we use sidecars to monitor these instead of internal collectors.  My changes to allow sidecars is nearly complete, with some unit tests for the functionality being taken care of, and I'll note that this is ready for another look.

pkg/controller/beat/common/stackmon/metricbeat.tpl.yml
```diff
@@ -0,0 +1,9 @@
+metricbeat.modules:
+- module: beat
+  metricsets:
+    - stats
+    - state
+  period: 10s
+  hosts: ["{{ .URL }}"]
+
```
pebrc, line 8: Don't we need `xpack.enabeld: true` here according to https://www.elastic.co/guide/en/beats/metricbeat/current/monitoring-metricbeat-collection.html

pkg/apis/beat/v1beta1/validations_test.go
```diff
@@ -207,3 +209,126 @@ func Test_checkNoDowngrade(t *testing.T) {
...
+								{
+									Name:      "elasticsearch",
+									Namespace: "test",
+								},
+							},
+						},
+					},
+				},
+			},
+			want: field.ErrorList{&field.Error{Type: field.ErrorTypeInvalid, Field: "spec.version", BadValue: "7.1.0", Detail: "Unsupported version for Stack Monitoring. Required >= 7.2.0."}},
```
pebrc, line 257: Would it make sense to have a test case that covers invalid version + invalid ref which as you mentioned is not properly covered in the common validation code?

pkg/controller/beat/common/config.go
```diff
@@ -109,6 +110,13 @@ func buildBeatConfig(
 		return nil, err
 	}
 
+	// if metrics monitoring is enabled, then enable the metrics http endpoint for metricsbeat sidecar.
+	if monitoring.IsMetricsDefined(&params.Beat) {
+		if err = cfg.MergeWith(settings.MustCanonicalConfig(map[string]bool{"http.enabled": true})); err != nil {
```
pebrc, line 115: Don't we also need to disable (self)monitoring and set a UUID according to this doc https://www.elastic.co/guide/en/beats/metricbeat/current/monitoring-metricbeat-collection.html

pkg/controller/beat/common/stackmon/filebeat.yml
```diff
@@ -0,0 +1,8 @@
+filebeat.inputs:
+- type: filestream
+  id: filebeatlogs
+  paths:
+    - /var/log/*.log
+    - /usr/share/filebeat/data/registry/filebeat/*.json
```
pebrc, line 6: Same as for metric beat: `xpack.enabled: true` https://www.elastic.co/guide/en/beats/filebeat/current/monitoring-metricbeat-collection.html

pkg/controller/beat/common/config_test.go
```diff
@@ -76,101 +76,135 @@ func Test_buildBeatConfig(t *testing.T) {
...
+		CASecretName:   "secret2",
+		URL:            "url",
+	})
 	assocConf, err := esAssocWithCA.AssociationConf()
 	require.NoError(t, err)
 	assocConf.CACertProvided = true
 
-	withAssocWithCAWithonfig := *withAssocWithCA.DeepCopy()
-	withAssocWithCAWithonfig.Spec.Config = userCfg
+	withAssocWithCAWithconfig := *withAssocWithCA.DeepCopy()
```
pebrc, line 90:
```suggestion
	withAssocWithCAWithConfig := *withAssocWithCA.DeepCopy()
```

pkg/controller/beat/common/pod_test.go
```diff
@@ -195,12 +305,72 @@ func assertConfiguration(t *testing.T, pod corev1.PodTemplateSpec) {
...
+		u, err := url.Parse(cfg.MetricBeat.Modules[0].Hosts[0])
+		require.NoError(t, err)
+		assert.Equal(t, "3033", u.Port())
+	}
+	if monitoring.IsLogsDefined(&beat) {
+		assert.True(t, containersContains(pod.Spec.Containers, "logs-monitoring-sidecar"))
+	}
+}
+
+// metricsbeatConfig represents the metricsbeat configuration for testing purposes.
```
pebrc, line 349:
```suggestion
// metricsbeatConfig represents the Metricbeat configuration for testing purposes.
```

pkg/controller/beat/common/pod_test.go
```diff
@@ -195,12 +305,72 @@ func assertConfiguration(t *testing.T, pod corev1.PodTemplateSpec) {
...
+//
+// example:
+//
+// metricbeat:
+//   modules:
+//     - hosts:
+//       - http://example.com:3033
+type metricsbeatConfig struct {
```
pebrc, line 357:
```suggestion
type metricbeatConfig struct {
```

pebrc (commented):
I left a few comments. In my tests the metrics never showed up in the "Stack monitoring" view in Kibana. It wasn't clear to me what was the root cause for this (e.g. I added the missing UUID)  but I guess it warrants further investigation.

naemono:
> I left a few comments. In my tests the metrics never showed up in the "Stack monitoring" view in Kibana. It wasn't clear to me what was the root cause for this (e.g. I added the missing UUID) but I guess it warrants further investigation.

Yeah, it's not showing up in stack monitoring for some reason for me either, but the data is definitely getting into the indices (same for filebeat, but not shown)
image

I'll figure out the reason for the missing `stack monitoring` kibana piece.

Also, I'm working through the comments....

pkg/controller/beat/common/stackmon/stackmon.go
```diff
@@ -0,0 +1,100 @@
...
+		return sideCar, errors.New("only one Elasticsearch reference is supported for Stack Monitoring")
+	}
+	assoc := associations[0]
+
+	credentials, err := association.ElasticsearchAuthSettings(client, assoc)
+	if err != nil {
+		return sideCar, err
+	}
+
+	username, password = credentials.Username, credentials.Password
```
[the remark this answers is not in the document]
naemono: til.  makes sense now.  thanks.

test/e2e/test/beat/steps.go
```diff
@@ -132,8 +135,8 @@ func (b Builder) CheckK8sTestSteps(k *test.K8sClient) test.StepList {
 					beat.Status.ExpectedNodes = 0
 					beat.Status.AvailableNodes = 0
 				}
-				if beat.Status != expected {
-					return fmt.Errorf("expected status %+v but got %+v", expected, beat.Status)
+				if !cmp.Equal(beat.Status, expected) {
+					return fmt.Errorf("expected status health %+v, got diff: %s", expected, cmp.Diff(beat.Status, expected))
```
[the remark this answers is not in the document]
naemono: leftover from some testing, and I missed it in my personal review.  removed.

pkg/controller/beat/common/stackmon/metricbeat.tpl.yml
naemono: Upon testing, yes it was needed.  Added.

pkg/apis/beat/v1beta1/validations_test.go
naemono: Invalid version isn't relevant since moving to sidecars; removed.

pkg/controller/beat/common/config.go
naemono: disabled self monitoring, and no UUID needed, as I testing with/without.

pkg/controller/beat/common/stackmon/filebeat.yml
naemono: For metricbeat, this setting moves the metrics from being written in the `*metricbeat` indices, to the `*monitoring` indices, and allows them to show up in stack monitoring page in kibana.  This isn't true for filebeat.  I reached out to the observability team and ran it by them.  They verified no setting will allow filebeat to write data to the `*monitoring` indices, and show up in stack monitoring.

pkg/controller/common/stackmon/validations/validations.go
naemono: PR Updated to use common validation

naemono:
I believe this is now ready for more 👀 .  I'm just verifying the e2e tests at this time.

naemono:
run/e2e-tests tags=beat

naemono:
run/e2e-tests tags=beat

naemono:
run/e2e-tests tags=beat

pkg/controller/beat/common/stackmon/filebeat.yml
pebrc: Hm odd. I wonder why the documentation recommends setting this then.

pebrc:
I did a quick test, haven't looked at the code yet. But somehow the UI still does not show content

Screenshot 2022-08-18 at 21 03 50



There is data flowing into my monitoring cluster though. Not sure what is going on with the stack monitoring UI. Going to take a closer look tomorrow.
```
green open .ds-metricbeat-8.3.2-2022.08.18-000001       qnmXL7hNQeKIecbdeJ3yhQ 1 1    99 0 214.2kb 106.4kb
green open .ds-filebeat-8.3.2-2022.08.18-000001         Wwi6mAX0T1Cczfs-dQDc4w 1 1 12178 0   2.8mb   1.3mb
green open .ds-.monitoring-beats-8-mb-2022.08.18-000001 851mItfaS7iiCbYwB5yQ2g 1 1   139 0 511.1kb 248.6kb
```

pkg/apis/beat/v1beta1/beat_types.go
```diff
@@ -99,6 +106,29 @@ type DeploymentSpec struct {
...
+	// +kubebuilder:validation:Required
+	ElasticsearchRefs []commonv1.ObjectSelector `json:"elasticsearchRefs,omitempty"`
+}
+
+type LogsMonitoring struct {
+	// ElasticsearchRefs is a reference to a list of monitoring Elasticsearch clusters running in the same Kubernetes cluster.
+	// Due to existing limitations, only a single Elasticsearch cluster is currently supported.
+	// +kubebuilder:validation:Required
+	ElasticsearchRefs []commonv1.ObjectSelector `json:"elasticsearchRefs,omitempty"`
+}
```
pebrc, line 130: I am thinking these types are probably identical between Elasticsearch, Kibana and Beats and could be shared? But maybe there was a reason not to as the original implementation chose to redefine them for each application?

pkg/controller/beat/common/config_test.go
```diff
@@ -75,114 +75,151 @@ func Test_buildBeatConfig(t *testing.T) {
...
+	withAssocWithCAWithConfig.Spec.Config = userCfg
 
 	withAssocWithConfig := *withAssoc.DeepCopy()
 	withAssocWithConfig.Spec.Config = userCfg
 
 	for _, tt := range []struct {
 		name          string
 		client        k8s.Client
-		beat          beatv1beta1.Beat
+		beat          func() beatv1beta1.Beat
```
pebrc, line 98: This does not need to be a function.

pkg/controller/beat/common/stackmon/filebeat.yml
pebrc: I just realised the documentation I mentioned covers Metricbeat monitoring Filebeat. Sorry my mistake 🤦

pkg/controller/beat/common/driver.go
```diff
@@ -99,6 +103,9 @@ func Reconcile(
 
 	podTemplate, err := buildPodTemplate(params, defaultImage, configHash)
 	if err != nil {
+		if errors.Is(err, beat_stackmon.ErrMonitoringClusterUUIDUnavailable) {
+			results.WithReconciliationState(reconciler.Requeue.WithReason("ElasticsearchRef UUID unavailable while configuring beats stack monitoring")).WithResult(reconcile.Result{Requeue: true, RequeueAfter: 10 * time.Second})
```
pebrc, line 107:
```suggestion
			results.WithReconciliationState(reconciler.RequeueAfter(10 * time.Second).WithReason("ElasticsearchRef UUID unavailable while configuring Beats stack monitoring"))
```

Reason is currently not used in Beats controller, but I guess it does not hurt either.

pkg/controller/beat/common/pod_test.go
```diff
@@ -158,6 +262,7 @@ func Test_buildPodTemplate(t *testing.T) {
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
+			tt.args.params.Context = context.Background()
```
pebrc, line 265: If we are defaulting this here, which is good, I think we should remove it where it is set in the test cases.

pkg/controller/beat/common/pod_test.go
```diff
@@ -198,12 +311,73 @@ func assertConfiguration(t *testing.T, pod corev1.PodTemplateSpec) {
 			break
 		}
 	}
-	assert.NotNil(t, configVolume)
+	require.NotNil(t, configVolume)
 	// Validate the mode
-	assert.NotNil(t, configVolume.DefaultMode, "default volume mode for beat configuration should not be nil")
+	require.NotNil(t, configVolume.DefaultMode, "default volume mode for beat configuration should not be nil")
 	assert.Equal(t, expectedConfigVolumeMode, *configVolume.DefaultMode)
```
pebrc, line 317: This should be `require` as well.

pkg/controller/beat/common/pod_test.go
```diff
@@ -198,12 +311,73 @@ func assertConfiguration(t *testing.T, pod corev1.PodTemplateSpec) {
...
+	// Validate that the Pod's volumes contain a Secret as a monitoring CA volume.
+	for i := range pod.Spec.Volumes {
+		if pod.Spec.Volumes[i].Name == "beat-monitoring-certs" {
+			monitoringVolume = &pod.Spec.Volumes[i]
+			break
+		}
+	}
+	require.NotNil(t, monitoringVolume)
+	require.NotNil(t, monitoringVolume.Secret)
+	assert.Equal(t, monitoringVolume.Secret.SecretName, "testbeat-es-testes-ns-monitoring-ca")
```
pebrc, line 332:
```suggestion
	require.Equal(t, monitoringVolume.Secret.SecretName, "testbeat-es-testes-ns-monitoring-ca")
```

pkg/controller/beat/common/pod_test.go
```diff
@@ -198,12 +311,73 @@ func assertConfiguration(t *testing.T, pod corev1.PodTemplateSpec) {
...
+		assert.True(t, ok)
+
+		var cfg metricbeatConfig
+		assert.NoError(t, yaml.Unmarshal(data, &cfg))
+		u, err := url.Parse(cfg.MetricBeat.Modules[0].Hosts[0])
+		require.NoError(t, err)
+		assert.Equal(t, "3033", u.Port())
+	}
+	if monitoring.IsLogsDefined(&beat) {
+		assert.True(t, containersContains(pod.Spec.Containers, "logs-monitoring-sidecar"))
```
pebrc, line 351: All these `assert.*` need to be `require`. `assert` returns a `bool` which is here simply ignored i.e. we would not even know if the test failed. `require` calls the `assert` variant and makes the test fail if the `assert` returns `false`.

pkg/controller/beat/common/stackmon/stackmon_test.go
```diff
@@ -0,0 +1,221 @@
...
+    username: es-user
+`
+	beatSidecarFixture := stackmon.BeatSidecar{
+		Container: containerFixture,
+		ConfigSecret: corev1.Secret{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "beat-beat-monitoring-metricbeat-config",
+				Namespace: "test",
+				Labels: map[string]string{
+					"common.k8s.elastic.co/type":                "elasticsearch",
```
pebrc, line 105: These labels seem wrong. Why is this of  type "elasticsearch". I understand this not something you introduced in this PR but this needs at least a follow up issue to clean this up.

pebrc (commented):
Looks good! A few more nits and I think we should call out the odd behaviour of the Kibana UI we observed during testing in the docs.

pkg/controller/beat/common/pod_test.go
pebrc: Turns out this is not true. Sorry about that. Test fails with `assert` just as well. TIL

pkg/apis/beat/v1beta1/beat_types.go
naemono: @pebrc I have updated this PR to move all of these *Monitoring objects to v1.common package, but this does change the generated docs to be less specific.  Lmk what you think.

pkg/controller/beat/common/stackmon/stackmon.go
```diff
@@ -0,0 +1,117 @@
...
+	ErrMonitoringClusterUUIDUnavailable = errors.New("beats stack monitoring cluster uuid is unavailable")
+)
+
+func Filebeat(ctx context.Context, client k8s.Client, resource monitoring.HasMonitoring, version string) (stackmon.BeatSidecar, error) {
+	filebeat, err := stackmon.NewFileBeatSidecar(ctx, client, resource, version, filebeatConfig, nil)
+	if err != nil {
+		return stackmon.BeatSidecar{}, err
+	}
+
+	return filebeat, nil
```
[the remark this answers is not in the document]
naemono: This was redundant in all iterations of this code, and I've removed it in all 3 places.

pkg/controller/beat/common/stackmon/stackmon_test.go
naemono: https://github.com/elastic/cloud-on-k8s/issues/5967

docs/advanced-topics/stack-monitoring.asciidoc
```diff
@@ -57,10 +57,28 @@ spec:
...
+        namespace: observability <3>
+    logs:
+      elasticsearchRefs:
+      - name: monitoring
+        namespace: observability <3>
 ----
 
 <1> The use of `namespace` is optional if the monitoring Elasticsearch cluster and the monitored Elasticsearch cluster are running in the same namespace.
 <2> The use of `namespace` is optional if the Elasticsearch cluster and the Kibana instance are running in the same namespace.
+<3> The use of `namespace` is optional if the Elasticsearch cluster and the Beats instance are running in the same namespace.
```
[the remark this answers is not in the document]
naemono: I've updated these docs.  Let me know if the wording is good.

pebrc (commented):
Sorry but I only now noticed that the log monitoring does not work for Beats. All that is ingested are the sidecars own logs. I think this is due to Beats logging to stdout on k8s as you should. For the other stack monitoring implementation for Elasticsearch and Kibana we switched to file based logging and a shared volume between the monitored Pod and the monitoring Pod to extract the logs.

naemono:
> All that is ingested are the sidecars own logs.

I'm working through this... seeing odd issue where something is listening on 5066, even though I am disabling this http endpoint in both filebeat, and metricbeat sidecars.  I'll update when it's resolved.

```
Exiting: could not start the HTTP server for the API: listen tcp 127.0.0.1:5066: bind: address already in use
```

naemono:
metricbeat, and filebeat (or any 2x beats) running on the same host, listening on `127.0.0.1:5066` seem to be the cause of this... I'm investigating a solution.
```
metricbeat-beat-metricbeat-4bs56            2/3     CrashLoopBackOff   10 (68s ago)   27m   10.142.0.41   gke-mmontgomery-testcluster01-pool-1-9b30cf4a-9eg2   <none>           <none>
filebeat-beat-filebeat-cx52x                2/2     Running            2 (27m ago)    27m   10.142.0.41   gke-mmontgomery-testcluster01-pool-1-9b30cf4a-9eg2   <none>           <none>
```

barkbay:
> I'm thinking through some sort of solution to this that's deterministic in which all ports,

Could you give more details about your solution?
* If deployed as a daemonset do we want to use the same port on all the nodes?
* How to not collide with ports already opened on the node?
* Is there any security consideration to have? 
* Any chance to have the Beats team working on a Unix socket based interface?

naemono:
@barkbay I don't currently have a solution to this problem.  I'll document here some potential (partial) solutions here that I've been working through, but none of them fully stop the potential for port conflicts when using `hostNetwork: true`, but that's true for ANY pod that uses this setting...

## Ensure distinct `http.port` setting for each Beat daemonset, with stack monitoring enabled, and `hostNetwork: true` set.

For each Beat that is both a daemonset, and has stack monitoring enabled with `hostNetwork: true` set, we could have a package that returns a distinct `http.port`, starting at `5066` and incrementing.

1.  We would have to 'save' the assigned stack monitoring port (likely) in the `status` of the Beat object for future queries.
2. This wouldn't fully resolve the issue, as a customer could have something running on 5066 on the host already, which would collide with the Beat.
3. Multiple operator instances running within the same cluster would potentially run into issues, as they wouldn't share this data, unless we persist this data in a shared configmap that multiple operators query.

## Adjust the `monitoring` section for Beat to require that the user assign a monitoring port when `hostNetwork: true` is set.

This solution move the onus to the user to ensure that the stack monitoring ports for Beat within the cluster do not collide. 

1. Again, multiple operators running within the same cluster could run into issues, as they may not have the RBAC rules to query Beats outside of their assigned namespaces for port assignments when validating the Beat object upon initial validation/creation.
2. Also has the potential problem where the k8s nodes may have something already running on the node, on the assigned port, that the user is unaware of.

## Document the potential port conflict, and remove any `hostNetwork: true` settings from Beat types that do not require it.

There are indications that `hostNetwork: true` is not fully required for Beats to function in certain Beat types, see [issue comment](https://github.com/elastic/beats/issues/15013#issuecomment-596428967).  If this is true, we could remove `hostNetwork: true` from our recipes/documentation for the Beats that do not require this setting (metricbeat/packetbeat seem to be the only ones that should 'require' this setting).  In this scenario, we could document the potential for port conflicts when hostnetwork is used, and leave it up to the user to avoid this scenario.

Let me know if you have strong feelings for/against these options.  I personally do not like assigning ports starting at `5066` and incrementing, as it seems likely run into a port conflict issue sooner or later, and the user would be pretty blind to the reason it's happening.  It also seems as though we should pursue removing `hostNetwork: true` from our recipes for Beats that do not require it regardless of the final solution we pursue here.

To answer the questions you specifically asked:

> * If deployed as a daemonset do we want to use the same port on all the nodes?

I think you're asking if 2x beat daemonsets should use the same port for each daemonset.  I'd say likely not if stack monitoring is enabled, and `hostNetwork: true` is set for both of them.

> * How to not collide with ports already opened on the node?

I don't know of any way to stop this from happening when a server on the k8s node is run outside of k8s that binds to a port.

> * Is there any security consideration to have?

We are opening port `5066` on the node's host network when stack monitoring is enabled, so any process on the host could query the Beat's metrics, so that's a potential security risk depending on the data that's served over the monitoring server.

> * Any chance to have the Beats team working on a Unix socket based interface?

I will bring this up with the Beats team to see their thoughts.

naemono:
@barkbay unix sockets/windows pipes appear to be a valid option, which could be named according to namespace/name, and solve this problem: https://www.elastic.co/guide/en/beats/metricbeat/current/http-endpoint.html.

I'll do some testing and see what comes of this...

naemono:
image

Beats stack monitoring now works over unix sockets, which solves the issues where we were getting port conflicts on host network.

naemono:
run/e2e-tests tags=beat

pkg/controller/beat/common/stackmon/filebeat.yml
```diff
@@ -0,0 +1,17 @@
...
+    id: filebeatlogs
+    paths:
+      - /var/log/*.log
+      - /var/log/beat/*
+      - /usr/share/filebeat/data/registry/filebeat/*.json
+
+path.home: /usr/share/filebeat-sidecar
+path.config: /usr/share/filebeat-sidecar/config
+path.data: /usr/share/filebeat-sidecar/data
+path.logs: /usr/share/filebeat-sidecar/logs
```
thbkrkr, line 13: Why do we need this?

thbkrkr (commented):
This looks good! I'm testing with this [manifest](https://gist.github.com/thbkrkr/10094ceca3fc6f148b5253223f8f678b) and I get a socket permission error for Filebeat:
```json
{
    "log.level": "error",
    "@timestamp": "2022-09-09T13:40:34.362Z",
    "log.origin": {
        "file.name": "module/wrapper.go",
        "file.line": 256
    },
    "message": "Error fetching data for metricset beat.state: error making http request:
 Get \"http://unix/state\": dial 
 unix /var/shared/filebeat-default-filebeat.sock: connect: permission denied",
    "service.name": "metricbeat",
    "ecs.version": "1.6.0"
}
```

naemono:
> This looks good! I'm testing with this [manifest](https://gist.github.com/thbkrkr/10094ceca3fc6f148b5253223f8f678b) and I get a socket permission error for Filebeat:
> 
> ```json
> {
>     "log.level": "error",
>     "@timestamp": "2022-09-09T13:40:34.362Z",
>     "log.origin": {
>         "file.name": "module/wrapper.go",
>         "file.line": 256
>     },
>     "message": "Error fetching data for metricset beat.state: error making http request:
>  Get \"http://unix/state\": dial 
>  unix /var/shared/filebeat-default-filebeat.sock: connect: permission denied",
>     "service.name": "metricbeat",
>     "ecs.version": "1.6.0"
> }
> ```

I'm investigating these permission issues...

naemono:
@thbkrkr meant to update yesterday.  The permission errors were from attempting to use `/usr/share/filebeat-sidecar` as the path for the filbeat sidecar, which was unnecessary.  I removed that from the config template, tested your manifest locally, and saw no issues.  This should be ready for 👀 again.

naemono:
@thbkrkr There were additional issues found surrounding sidecar permissions reading the unix socket.  These were resolved, and tests were added around this feature.  It's again ready for 👀

pkg/controller/beat/common/pod.go
```diff
@@ -146,11 +204,34 @@ func buildPodTemplate(
 		WithVolumes(volumes...).
 		WithVolumeMounts(volumeMounts...).
 		WithInitContainers(initContainers...).
-		WithInitContainerDefaults()
```
thbkrkr, line 149: IIUC the `-e` flag passed as an argument prevents the main beat process to write a log file on disk (warning in this section https://www.elastic.co/guide/en/beats/filebeat/current/configuration-logging.html#configuration-logging) despite we have configured the logging settings for this.

naemono: @thbkrkr I didn't update the comments here, but changes were pushed that move logging to a file when stack monitoring is enabled.

pebrc:
I tapped into a trap when testing this yesterday by using one of our existing manifests with the `-e` option and only with @thbkrkr 's help was able to identify why my logs would not show up in the monitoring cluster. 

I think we have a usability issue here with some of the existing recipes using the `-e` option. We can assume that some of our users also might have specified it when overriding the default beats command. I think we should at least comment this requirement and maybe even consider, as @thbkrkr suggested, to automatically remove the `-e` option if log delivery is configured.

naemono:
> I think we have a usability issue here with some of the existing recipes using the `-e` option. We can assume that some of our users also might have specified it when overriding the default beats command. I think we should at least document this requirement and maybe even consider, as @thbkrkr suggested, to automatically remove the `-e` option if log delivery is configured.

I didn't think of the scenario where the customer/recipe already had the `-e` option. Resolved in https://github.com/elastic/cloud-on-k8s/pull/5878/commits/52cd829cc6259cc1bdbd81158f845318bcd384f6.  I'll also get the documentation updated to note this.

naemono:
@pebrc documentation added.  Let me know how you feel about the wording.

thbkrkr (approved):
LGTM :rocket:

pebrc (approved):

PR #5887, opened 2022-07-22
Retry E2E job a single time.

naemono:
Since we continue to see odd failures in Azure/AKS for our CI jobs, let's allow the k8s job controlling the test run to retry the job a single time.

According to [k8s job documentation](https://kubernetes.io/docs/concepts/workloads/controllers/job/#pod-backoff-failure-policy), setting the following on the job should allow this:

```
job.spec.backoffLimit: 1
job.spec.template.spec.restartPolicy: OnFailure
```


naemono:
run/e2e-tests

pebrc:
Would that not just cover up the problem? Do we have a theory what causes the failures?

PR #5889, opened 2022-07-26
ECK resources Helm chart - Elastic Agent & Elastic Fleet Server Agent

naemono:
related to https://github.com/elastic/cloud-on-k8s/issues/5505

This PR contains the 2nd version of the ECK-managed resources helm charts including

eck-elastic-agent chart for installing elastic-agent
eck-fleet-server for managing fleet server as an Elastic Agent
minor updates to the eck-stack chart.

~~Note that this Chart allows multiple instances of Elastic Agent to be installed, along with allowing multiple instances of  serviceAccounts, clusterRoles, and clusterRoleBindings.~~

## ~~In Progress~~
- [x] additional manual testing
- [x] additional chart unit tests


deploy/eck-elastic-agent/Chart.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v2
+name: eck-elastic-agent
+description: A Helm chart to deploy Elastic Agent managed by the ECK Operator.
+kubeVersion: ">= 1.20.0-0"
+type: application
+version: 0.1.0
+sources:
+  - https://github.com/elastic/cloud-on-k8s
+  - https://github.com/elastic/elastic-agent
+icon: https://images.contentstack.io/v3/assets/bltefdd0b53724fa2ce/blt77c2da6e0198746e/620ac24e6662ca0a6f617114/icon-agent-32-color.svg
```
naemono, line 10: Anyone know a better icon location than this?

pebrc (commented):
LGTM for the approach taken. One thing I am wondering though is if we could make the whole  RBAC side of deploying agent easier with the Helm chart. But I understand that is quite difficult if you have list of agents where you do not know what it what.

naemono:
> make the whole RBAC side of deploying agent easier with the Helm chart

I would love to make this easier for the user, but I'm not sure the Helm chart is the right place to do it.  If we wanted to ensure that rbac rules were sufficient to the configuration, that logic would likely make more sense, and be easier to maintain in the operator itself, as complex logic in Helm charts just comes across as complex, ugly, and difficult to maintain.

I can certainly go down that road and attempt this within the Helm charts.  Please let me know and I'll make the changes, and see what they look like.

pebrc:
>  Please let me know and I'll make the changes, and see what they look like.

I don't have something that is ready to implement. I feel like Agent is special enough, compared to Kibana and Elasticsearch, to warrant doing a bit of design work for the Helm chart. 

For example one idea I had was to make two different Helm charts one for Fleet Server and one for Elastic Agents. Behind the scenes it is the same CRD of course still. 

That would allow us to configure two different sets of default RBAC permissions.

More far reaching permissions for Agent,  roughly oriented maybe on Kubernetes observability recipe or maybe [DaemonSet published by the Agents team](https://raw.githubusercontent.com/elastic/elastic-agent/master/deploy/kubernetes/elastic-agent-managed-kubernetes.yaml)

And less permissions for the Fleet server chart. 

User could then override the default RBAC permissions for each of these charts to either restrict them further (e.g. for an APM integration we don't need all the K8s API permissions that K8s observability needs) or expand them where necessary. 

With a Fleet server chart and an Elastic Agent chart we should be able to cover most if not all use cases we have currently in the recipes section of the repository. This is just an idea at this stage and not ready to implement. Probably worth discussing with the team. My motivation is to find a solution how we can keep the templating logic minimal in the Charts but still add value for our users by starting with sane defaults that cover the most common case.

naemono:
We met and had a discussion about the current implementation, vs a more specialized implementation for, say, Fleet, and comparing what the user experience was like between the two, we decided to implement the more specialized Helm chart for Fleet, which simplifies what the Agent helm chart looks like.  I'll move this back to a WIP, and update when there's more information.

naemono:
@pebrc Update: Both Elastic Agent, and Elastic Fleet Server Agent helm charts are now included in this PR, and are ready for review.

pebrc (commented):
This looks really promising. I think I would take it maybe one step further with the defaults in the Elastic Agent chart I left a few comments to explain.

deploy/eck-fleet-server/values.yaml
```diff
@@ -0,0 +1,115 @@
...
+  elasticsearchRefs:
+  - name: quickstart
+    # Optional namespace reference to Elasticsearch resource.
+    # If not specified, then the namespace of the Fleet Server Agent resource
+    # will be assumed.
+    #
+    # namespace: default
+  
+  mode: fleet
+  fleetServerEnabled: true
```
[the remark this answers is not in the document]
naemono: https://github.com/naemono/cloud-on-k8s/blob/f629f9344da1c44da64af90f1bf957f430d72f87/deploy/eck-fleet-server/templates/fleet-server.yaml#L16-L23

Note that there's validation around these that ensure that they are the right string/bool.  If we hard-coded them in the template, and someone added them, we'd get duplicative keys in this `spec` map.  I certainly could add in some template logic to remove them if they're in the default values though.   Let me know your thoughts.

deploy/eck-elastic-agent/values.yaml
```diff
@@ -0,0 +1,175 @@
...
+      spec:
+        containers:
+        - name: agent
+          securityContext:
+            runAsUser: 0
+
+  # Configuration of Agent, specifically used in Agent standalone mode.
+  # ref: https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-elastic-agent.html
+  #
+  config:
```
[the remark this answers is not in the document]
naemono: I did this intentionally to make the default agent installation something that was functional.  Even the above daemonSet block that's in the default values here is going to be affected by that bug, which seems to be moving towards a merging/resolution from discussion in [slack](https://kubernetes.slack.com/archives/C51E88VDG/p1659353024087259) 

If you'd prefer to leave this config block empty, I have no issues with it, so lmk.

deploy/eck-elastic-agent/values.yaml
```diff
@@ -0,0 +1,175 @@
...
+      streams:
+      - metricset: cpu
+        data_stream.dataset: system.cpu
+
+# ServiceAccount to be used by Elastic Agent. Some Elastic Agent features, such as the Kubernetes integration,
+# require that Agent Pods interact with Kubernetes APIs. This functionality requires specific permissions
+# ref: https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-elastic-agent-fleet-configuration.html#k8s-elastic-agent-fleet-configuration-role-based-access-control
+#
+serviceAccount: {}
+  # name: elastic-agent
```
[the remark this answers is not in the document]
naemono: IIUC, the default configuration I have in place here doesn't require k8s api interactions, which is why I left this block blank.  I'm testing this now to verify.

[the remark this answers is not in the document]
naemono: image

Yeah, I'm seeing the default configuration working without SA.

deploy/eck-elastic-agent/values.yaml
```diff
@@ -0,0 +1,175 @@
...
+clusterRoleBinding: {}
+  # name: fleet-server
+  # subjects:
+  # - kind: ServiceAccount
+  #   name: elastic-agent
+  #   namespace: default
+  # roleRef:
+  #   kind: ClusterRole
+  #   name: elastic-agent
+  #   apiGroup: rbac.authorization.k8s.io
```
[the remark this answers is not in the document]
naemono: Most recent commit adds a default for the subjects.namespace.  As for on by default, see previous comment, and lmk if you want this to change.

deploy/eck-elastic-agent/values.yaml
```diff
@@ -0,0 +1,175 @@
...
+  #   verbs:
+  #   - get
+  # - apiGroups:
+  #   - "batch"
+  #   resources:
+  #   - jobs
+  #   verbs:
+  #   - "get"
+  #   - "list"
+  #   - "watch"
```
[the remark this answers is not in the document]
naemono: Ok, I see the point here.  So, ignore previous comments about system-integration being on by default, and I'll add a commit to enable these shortly.

deploy/eck-elastic-agent/values.yaml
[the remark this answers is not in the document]
naemono: ^^ Ignore these comments, as https://github.com/elastic/cloud-on-k8s/pull/5889#discussion_r951477756 is the most recent... changes are coming here to enable RBAC by default...

deploy/eck-elastic-agent/values.yaml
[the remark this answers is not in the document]
naemono: This is now blank by default.

deploy/eck-elastic-agent/values.yaml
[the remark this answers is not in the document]
naemono: RBAC Enabled by default.  Also, tests updated to ensure default RBAC values.

deploy/eck-elastic-agent/values.yaml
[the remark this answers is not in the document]
naemono: This is now complete.

deploy/eck-fleet-server/templates/fleet-server.yaml
```diff
@@ -0,0 +1,25 @@
...
+  {{- if not (hasKey .Values.spec "mode") }}
+  {{- fail "spec.mode of 'fleet' is required for fleet-server" }}
+  {{- else if (not (eq .Values.spec.mode "fleet")) }}
+  {{- fail "spec.mode of 'fleet' is required for fleet-server" }}
+  {{- end }}
+  {{- if not (hasKey .Values.spec "fleetServerEnabled") }}
+  {{- fail "spec.fleetServerEnabled is required to be true for fleet-server" }}
+  {{- else if (not (.Values.spec.fleetServerEnabled)) }}
+  {{- fail "spec.fleetServerEnabled is required to be true for fleet-server" }}
+  {{- end }}
```
barkbay, line 24: Why do we allow the user to change these settings in deploy/eck-fleet-server/values.yaml if they have "fixed" and mandatory values?

```suggestion
  mode: "fleet"
  fleetServerEnabled: true
  {{- if (hasKey .Values.spec "mode") }}
  {{- fail "spec.mode cannot be changed" }}
  {{- end }}
  {{- if (hasKey .Values.spec "fleetServerEnabled") }}
  {{- fail "spec.fleetServerEnabled cannot be changed" }}
  {{- end }}
  {{- toYaml .Values.spec | nindent 2 }}
```

barkbay (commented):
Still testing...

deploy/eck-elastic-agent/examples/system-integration.yaml
```diff
@@ -0,0 +1,131 @@
+version: 8.2.3
```
barkbay, line 1: I would add a comment to explain that this should only be used in "standalone" mode. (or maybe add a `README.md` similar to this [one](https://github.com/elastic/cloud-on-k8s/blob/main/config/recipes/elastic-agent/README.asciidoc))

deploy/eck-elastic-agent/values.yaml
```diff
@@ -0,0 +1,165 @@
...
+      spec:
+        containers:
+        - name: agent
+          securityContext:
+            runAsUser: 0
+
+  # Configuration of Agent, specifically used in Agent standalone mode.
+  # ref: https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-elastic-agent.html
+  #
+  config: {}
```
barkbay, line 80: Should we let the user set `config` and `elasticsearchRefs ` explicitly? In fleet mode, using the provided example `deploy/eck-elastic-agent/examples/fleet-agents.yaml`,  it fails with:

```
Error: INSTALLATION FAILED: admission webhook "elastic-agent-validation-v1alpha1.k8s.elastic.co" denied the request: Agent.agent.k8s.elastic.co "my-agents-eck-elastic-agent" is invalid:
[spec.config: Invalid value: v1.Config{Data:map[string]interface {}{}}: remove config, it can't be set in fleet mode, spec.fleetServerEnabled: Invalid value: false: remove Elasticsearch reference, it can't be enabled in fleet mode when Fleet Server is not enabled as well]
```

deploy/eck-elastic-agent/values.yaml
```diff
@@ -0,0 +1,165 @@
...
+    # If not specified, then the namespace of the Agent instance
+    # will be assumed.
+    #
+    # namespace: default
+
+  # Daemonset, or Deployment specification for Agent.
+  # At least one is required of [daemonSet, deployment].
+  # ref: https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-elastic-agent-fleet-configuration.html#k8s-elastic-agent-fleet-configuration-chose-the-deployment-model
+  #
+  daemonSet:
```
barkbay, line 69: How do we switch to a `Deployment`?
I get this error when I try to replace `daemonSet`  with `deployment` in  `deploy/eck-elastic-agent/examples/fleet-agents.yaml`:

```
admission webhook "elastic-agent-validation-v1alpha1.k8s.elastic.co" denied the request: Agent.agent.k8s.elastic.co "my-agents-eck-elastic-agent" is invalid: [spec.daemonSet: Forbidden: Specify either daemonSet or deployment, not both,
```

deploy/eck-elastic-agent/.helmignore
```diff
@@ -0,0 +1,24 @@
+# Patterns to ignore when building packages.
```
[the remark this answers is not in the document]
naemono: Makes sense.  I've renamed it.

deploy/eck-fleet-server/templates/fleet-server.yaml
naemono: Makes sense, and updated.

deploy/eck-elastic-agent/examples/system-integration.yaml
naemono: I've added comments to both examples noting when they should be used.

deploy/eck-elastic-agent/values.yaml
naemono: I've disabled them explicitly in the examples now.  We have to allow these to be adjustable, as the agent chart supports both fleet, and standalone, so config is required in some cases.  Are you thinking that config should just be set to `null` in the default values instead of `{}`, and the ES ref left empty?

deploy/eck-elastic-agent/values.yaml
naemono: You'd have to set `spec.daemonSet: null` since it's defined in the default values.  This is unfortunate, as we're trying to make the default useful, but this is never really user friendly when making modifications, as you have to set default values to `null` to disable them.

deploy/eck-agent/templates/elastic-agent.yaml
```diff
@@ -0,0 +1,18 @@
...
+  name: {{ include "elasticagent.fullname" . }}
+  labels:
+    {{- include "elasticagent.labels" $ | nindent 4 }}
+  annotations:
+    eck.k8s.elastic.co/license: enterprise
+    {{- if .Values.annotations }}
+    {{- toYaml .Values.annotations | indent 4 }}
+    {{- end }}
+spec:
+  version: {{ required "An Elastic-Agent version is required" .Values.version }}
```
thbkrkr, line 14:
```suggestion
  version: {{ required "An Elastic Agent version is required" .Values.version }}
```

deploy/eck-fleet-server/Chart.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v2
+name: eck-fleet-server
+description: A Helm chart to deploy Elastic Fleet Server as an Agent managed by the ECK Operator.
```
thbkrkr, line 3: To keep it simple and consistent with its [doc](https://www.elastic.co/guide/en/fleet/current/fleet-overview.html), I think we should just write 'Fleet Server' and not 'Fleet Server Agent' or 'Elastic Fleet Server Agent'. It is not as ambiguous as Agent.

```suggestion
description: A Helm chart to deploy Fleet Server managed by the ECK Operator.
```

(`grep '\(Elastic \)*Fleet Server Agent' -ir deploy` to list occurrences to change)

naemono: Sounds good.  All have been updated.

deploy/eck-agent/templates/cluster-role-binding.yaml
```diff
@@ -0,0 +1,28 @@
+{{- if .Values.clusterRoleBinding }}
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: {{ .Values.clusterRoleBinding.name }}
+  labels:
+    {{- include "elasticagent.labels" $ | nindent 4 }}
+    {{- if .Values.clusterRoleBinding.labels }}
+    {{- toYaml .Values.clusterRoleBinding.labels | indent 4 }}
```
thbkrkr, line 10:
```suggestion
    {{- toYaml .Values.clusterRoleBinding.labels | nindent 4 }}
```

deploy/eck-agent/templates/cluster-role-binding.yaml
```diff
@@ -0,0 +1,28 @@
...
+    {{- end }}
+  {{- if .Values.clusterRoleBinding.annotations }}
+  annotations:
+    {{- toYaml .Values.clusterRoleBinding.annotations | indent 4 }}
```
thbkrkr, line 14:
```suggestion
    {{- toYaml .Values.clusterRoleBinding.annotations | nindent 4 }}
```

thbkrkr:
Is it intentional that labels/annotations are managed differently in the ClusterRoleBinding/ClusterRole/ServiceAccount between the two charts? The eck-agent chart supports dedicated labels/annotations but for labels we also get the labels of the agent, while for the eck-fleet-server no annotation at all and only the labels from the fleet-server.

naemono:
> Is it intentional that labels/annotations are managed differently in the ClusterRoleBinding/ClusterRole/ServiceAccount between the two charts? The eck-agent chart supports dedicated labels/annotations but for labels we also get the labels of the agent, while for the eck-fleet-server no annotation at all and only the labels from the fleet-server.

This absolutely was not intentional, and has been resolved.  Thanks for catching that @thbkrkr.  I think I'll also add additional tests around labels/annotations similar to https://github.com/elastic/cloud-on-k8s/pull/6004/files#diff-9612072d3cca4f0281e66e5bd8fdb713f35792862686355d89ff507348419028R41

naemono:
@thbkrkr While adding tests, I noticed that `.Values.[labels|annotations]`, and `.Values.[serviceaccount|clusterRole|clusterRoleBinding].[labels|annotations]` weren't both making it into the rendered templates, and this was fixed, and tests were added for all.

deploy/eck-agent/examples/fleet-agents.yaml
```diff
@@ -0,0 +1,98 @@
...
+        hostNetwork: true
+        dnsPolicy: ClusterFirstWithHostNet
+        automountServiceAccountToken: true
+        securityContext:
+          runAsUser: 0
+
+serviceAccount:
+  name: elastic-agent
+
+clusterRole:
```
pebrc, line 31: This seems unnecessary now that we default to a cluster role

deploy/eck-agent/examples/fleet-agents.yaml
```diff
@@ -0,0 +1,98 @@
...
+  - apiGroups:
+    - "batch"
+    resources:
+    - jobs
+    verbs:
+    - "get"
+    - "list"
+    - "watch"
+
+clusterRoleBinding:
```
pebrc, line 89: Same here: this is defaulted already

deploy/eck-agent/values.yaml
```diff
@@ -0,0 +1,165 @@
...
+      spec:
+        containers:
+        - name: agent
+          securityContext:
+            runAsUser: 0
+
+  # Configuration of Agent, specifically used in Agent standalone mode.
+  # ref: https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-elastic-agent.html
+  #
+  config: {}
```
pebrc, line 80: This will lead to errors when running in Fleet mode but due to the Helm issues it cannot be nullified. We could relax the validation in the operator to tread `config: nil` and `config: {}` the same.

deploy/eck-stack/examples/fleet-agents.yaml
```diff
@@ -0,0 +1,208 @@
...
+          hostNetwork: true
+          dnsPolicy: ClusterFirstWithHostNet
+          automountServiceAccountToken: true
+          securityContext:
+            runAsUser: 0
+  
+  serviceAccount:
+    name: elastic-agent
+
+  clusterRole:
```
pebrc, line 130: Same as the example in the Agent chart: the `clusterRole` and `clusterRoleBinding` is not necessary here because we have a sensible default for it.

deploy/eck-stack/Chart.yaml
```diff
@@ -24,3 +24,17 @@ dependencies:
     # the helm.elastic.co repository.
     # repository: "file://../eck-kibana"
     repository: "https://helm.elastic.co"
+  - name: eck-elastic-agent
```
pebrc, line 27: You renamed the chart but did not update the references here

deploy/eck-agent/values.yaml
naemono: > This will lead to errors when running in Fleet mode but due to the Helm issues it cannot be nullified via a values override. We could relax the validation in the operator to tread `config: nil` and `config: {}` the same.

I went ahead and tested this, as I thought I had run though this process a couple weeks ago, and I can't replicate this issue.  The validating webhook allows this; all pods come online; I don't see any errors in fleet-server, or agent pod logs.  Can you expound on where the errors from this might be @pebrc ?

```
❯ helm install eck-stack . -n default --values ./examples/fleet-agents.yaml
NAME: eck-stack
LAST DEPLOYED: Tue Sep 20 12:15:01 2022
NAMESPACE: default
STATUS: deployed
REVISION: 1
TEST SUITE: None
NOTES:
Elasticsearch ECK-Stack 0.1.0 has been deployed successfully!

To see status of all resources, run

kubectl get elastic -n default -l "app.kubernetes.io/instance"=eck-stack

More information on the Elastic ECK Operator, and its Helm chart can be found
within our documentation.

https://www.elastic.co/guide/en/cloud-on-k8s/current/index.html

❯ kc get MutatingWebhookConfiguration,ValidatingWebhookConfiguration -A | grep elastic
validatingwebhookconfiguration.admissionregistration.k8s.io/elastic-operator.elastic-system.k8s.elastic.co                    10         21m

❯ kc get pods -n default
NAME                                  READY   STATUS    RESTARTS   AGE
eck-stack-eck-agent-agent-9nhhf       1/1     Running   0          2m30s
eck-stack-eck-agent-agent-gvq4n       1/1     Running   0          2m30s
eck-stack-eck-agent-agent-s8p2k       1/1     Running   0          2m30s
elasticsearch-es-default-0            1/1     Running   0          8m8s
elasticsearch-es-default-1            1/1     Running   0          8m8s
elasticsearch-es-default-2            1/1     Running   0          8m8s
fleet-server-agent-6fddd99dcd-h8679   1/1     Running   0          2m31s
kibana-kb-6766b97bf7-k6z8w            1/1     Running   0          8m4s
```

pebrc: Just tried once more (I am using a fresh 2.4.0 install, yesterday I was using the operator code from your branch) 

```
helm install eck-stack . -n default --values ./examples/fleet-agents.yaml
Error: INSTALLATION FAILED: admission webhook "elastic-agent-validation-v1alpha1.k8s.elastic.co" denied the request: Agent.agent.k8s.elastic.co "eck-stack-eck-agent" is invalid: spec.config: Invalid value: v1.Config{Data:map[string]interface {}{}}: remove config, it can't be set in fleet mode
```

naemono: @pebrc Agent default value for `spec.config` moved to `null` instead of `{}` to resolve.  Let me know if you see any further issues.

pebrc (approved):

PR #5890, opened 2022-07-27
Configure host path volume to point to correct path

### LINKED ISSUE #4428: Adjust Elastic Agent controller to make use of new container state handling in 7.13
> By default the state of the Docker container goes to /usr/share/elastic-agent/state. This can changed with STATE_PATH environment. Elastic Agent logs now default to output stderr so they can be collected by the container runtime. In the case that LOGS_PATH variable is set then the Elastic Agent will log to $LOGS_PATH/elastic-agent.log instead of stderr.

from https://github.com/elastic/beats/pull/24817

We have to adjust the data path starting with 7.13 to benefit from this improvement.

pebrc:
Fixes #4428 

Mount the host path volume to point to the path where Elastic Agent stores its local state. This should make sure that while running on the same version no duplicate ingestion is happening. After version upgrades the internal sub-paths within the state directory still change see https://github.com/elastic/elastic-agent/issues/750 and events will be re-ingested. 

This is a potentially breaking change as I am configuring the host path volume now independently of the mode the agents are running in: Fleet or stand-alone. It will be breaking for users who have configured a deployment with multiple replicas where more than one replicas is placed by the Kubernetes scheduler on the same K8s node. 

Can we mitigate the effects of this? We could optimistically assume that Deployments with multiple replicas will be mainly used for Fleet Server and not use the host path volumes in this case.

I tried to think of a test case but it is tricky without restarting Elastic Agent Pods to see if duplicates are ingested or not and no clear indication of ingest state (i.e. have we waited long enough for duplicates to appear etc)

### DIFF
--- a/docs/orchestrating-elastic-stack-applications/agent-fleet.asciidoc
+++ b/docs/orchestrating-elastic-stack-applications/agent-fleet.asciidoc
@@ -535,3 +535,24 @@ To establish trust, the Pod needs to update the CA store through a call to `upda
 
 === Running Endpoint Security integration
 Running Endpoint Security link:https://www.elastic.co/guide/en/security/current/install-endpoint.html[integration] is not yet supported in containerized environments, like Kubernetes. This is not an ECK limitation, but the limitation of the integration itself. Note that you can use ECK to deploy Elasticsearch, Kibana and Fleet Server, and add Endpoint Security integration to your policies if Elastic Agents running those policies are deployed in non-containerized environments.
+
+=== Storing local state in host path volume
+Elastic Agent when managed by ECK stores local state in a host path volume. This ensures that integrations run by the agent can continue their work without duplicating work that has already been done after the Pod has been recreated for example because of a Pod configuration change.
+Multiple replicas of an agent, for example Fleet Servers, can not be deployed on the same underlying Kubernetes node as they would try to use the same host path. If local state storage in host path volumes is not desired this can be turned off by configuring an `emptyDir` volume instead:
+
+[source,yaml]
+----
+apiVersion: agent.k8s.elastic.co/v1alpha1
+kind: Agent
+metadata:
+  name: fleet-server
+spec:
+  deployment:
+    replicas: 3
+    podTemplate:
+      spec:
+        volumes:
+        - name: agent-data
+          emptyDir: {}
+...
+----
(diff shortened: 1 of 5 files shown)


barkbay (commented):
Since we are considering to flag this as a breaking change, and since Agent is still considered as "experimental", I'm wondering if we should not change the path on the host node's filesystem

```yaml
  - hostPath:
      path: /var/lib/<ns>/<name>/agent-data
      type: DirectoryOrCreate
```

I never realised that the first directory created by the operator is named according to the namespace name. However `/var/lib` is not exclusively used by ECK, I'm then wondering if we should not consider the case where there might be a conflict with an existing directory. For example on GKE nodes there is an existing directory named `/var/lib/metrics`. Creating an agent in a namespace named `metrics` would store Agent data in a directory that might have different purposes and lifecycle. Also it does not seem to be consistent with the naming from the Agent project which is `/usr/share/elastic-agent/...`

Should we change the directory path on the host to something like `/usr/share/elastic-agent/server/<ns>/<name>/state` ?

pebrc:
> Should we change the directory path

Yes I think that sounds reasonable. 

> /usr/share/elastic-agent/server/<ns>/<name>/state

Why the server bit?

pebrc:
Also `/var/lib` seems to the right place for [application state data](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch05s08.html)? 

Should we do `/var/lib/elastic-agent/<ns>/<name>/state`?

barkbay:
> Why the server bit?

I was wondering if it would make sense to add the Fleet "type" in the path, but you're right it's not required.

> Should we do `/var/lib/elastic-agent/<ns>/<name>/state`?

The intent was to be consistent with the path in the Pod. It is however definitely not a strong opinion and you're right, I think `/var/lib` is better suited.

barkbay:
> Can we mitigate the effects of this? We could optimistically assume that Deployments with multiple replicas will be mainly used for Fleet Server and not use the host path volumes in this case.

Maybe we could mention it in our documentation as a "[limitation](https://www.elastic.co/guide/en/cloud-on-k8s/2.2/k8s-elastic-agent-fleet-known-limitations.html)"?

barkbay (approved):
LGTM

pebrc:
run/e2e-tests tags=agent

pebrc:
run/e2e-tests tags=agent

PR #5773, opened 2022-06-14
Buildkite

thbkrkr:
This is the first Buildkite pipeline which:
- build the operator and the e2e-tests Docker images using buildah
- lint go code, run unit-tests, integration-tests, manifest-gen tests and Helm Charts tests
- run e2e-tests dynamically depending on the branch and the schedule (GitHub commit vs nightly schedule)
    - !main => only TestSmoke on Kind
    -  main && !nightly => all tests on GKE
    -  main && nightly  => all tests on all envs

It replaces almost all the Jenkins jobs. Only the release job is missing.
It can run alongside the existing Jenkins jobs.

2022-09-29-202208_1241x372_scrot

With Jenkins, we use GCP VM for the worker nodes and we run every steps using Makefile targets inside a CI Docker image (`make -C .ci TARGET="x y z" ci`). With Buildkite, we have the ability to provide a Docker image and use a lightweight Agent that runs on k8s using this image. This implies that the test artefacts downloaded before the CI Docker image was started must be done in the Agent.
The old mecanism is still possible on Buildkite using a `gcp` (VM) agent. It is kept for running e2e-tests on Kind because Kind requires Docker.

* f0eef959f  Hello Buildkite!
	- Pipeline definition
	- Buildkite Agent Dockerfile (this is roughly the CI Docker image defined in .ci/Dockerfile)
	- bash scripts to get the test artefacts independently from `.ci/Makefile`

* 9d75bec6e  Add buildah make targets
	- New targets to use buildah to build and push the operator and the e2e-tests Docker image from the lightweight k8s agent

* de9390a53  Add steps manifest-gen and helm
	- Dissociate the manifest-gen tests from unit-tests
	- Use `kubeval` binary instead of the Docker image
	- Add Helm tests to the pipeline

* 933983edc  Tweak lint and shellcheck commands
	- Add `--external-sources ` because `.buildkite/scripts/common/get-test-artifacts.sh` source `.env`

* 62499b402  Add --userns=host for Docker next Docker
	- To be able to run `docker run` with the socket of a Docker daemon host configured configured to run containers in user namespaces

* e5e026267  Default values for e2e/main
	- Set default values for TEST_TAGS and E2E_TAGS variables

* c85d5d76b  Make E2E_IMG_TAG constant
	- Make e2e tests image name independent of the e2e tests pipeline name. This allows e2e tests to be run without having to rebuild the e2e tests image.

* b47c8b211  Use new bk-agent image for CI
[...]

PR #5891, opened 2022-07-27
E2E: Run ECK diagnostics for any failed step

thbkrkr:
Currently we dump k8s events and run eck-diagnostics only if the `runTestsRemote` step fails. When for example `waitForOperatorToBeReady` fails, we remain blind. This commit moves the eck-diagnostics run so we have it for any failed step.

It helped me in #5773.
This may also help to diagnose https://github.com/elastic/cloud-on-k8s/issues/5825 and future errors when the operator pod is never ready.

### DIFF
--- a/test/e2e/cmd/run/run.go
+++ b/test/e2e/cmd/run/run.go
@@ -88,6 +88,8 @@ func doRun(flags runFlags) error {
 
 	for _, step := range steps {
 		if err := step(); err != nil {
+			helper.dumpEventLog()
+			helper.runECKDiagnostics()
 			return err
 		}
 	}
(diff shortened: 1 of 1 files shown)


thbkrkr:
> AFAIK it completely ignores the namespace in which the test job runs.

I'm mostly interested in debugging why the operator Pod doesn't start with this PR, not the job. You are right that for the current AKS issue, it is the Job Pod we are interested in. I checked and it seems ok.

The job is deployed in the `context.E2ENamespace`:

https://github.com/elastic/cloud-on-k8s/blob/721cccfd905f43c345c7957fbfe173ea3ebb6bbf/config/e2e/e2e_job.yaml#L14-L17

which is added to the list of namespaces passed to the `--resources-namespaces` flag:

https://github.com/elastic/cloud-on-k8s/blob/bf36c5d1ddb4b1af10d3e856aa22cff5c2971fb0/test/e2e/cmd/run/run.go#L826-L829

PR #5895, opened 2022-07-28
Fix Vault path handling on read

thbkrkr:
Moves the path handling in a single `read` method to avoid to append twice the root path when `GetMany` uses `Get`.

Following of #5892.


pebrc (approved):

thbkrkr:
```

thbkrkr:
Jenkins test this please

PR #5896, opened 2022-07-28
Reuse Fleet enrollment tokens where possible

pebrc:
I realised during testing that my new implementation of the Fleet enrolment creates enrolment tokens that are never cleaned up when the Elastic Agent is deleted. I cannot think of a reliable way of keeping track which tokens were created by the operator so instead I propose to change the implementation to try to reuse existing valid tokens where possible. 

This PR also fixes the APM tracing which was recording requests made to the Kibana Fleet API as Elasticsearch requests (and the same also for requests made to the Enterprise Search API)


go.sum
```diff
@@ -1306,16 +1401,23 @@ honnef.co/go/tools v0.0.1-2020.1.4/go.mod h1:X/FiERA/W4tHapMX5mGpAtMSVEeEUOyHaw9
...
 k8s.io/apimachinery v0.24.2 h1:5QlH9SL2C8KMcrNJPor+LbXVTaZRReml7svPEh4OKDM=
 k8s.io/apimachinery v0.24.2/go.mod h1:82Bi4sCzVBdpYjyI4jY6aHX+YCUchUIrZrXKedjd2UM=
+k8s.io/apiserver v0.24.0/go.mod h1:WFx2yiOMawnogNToVvUYT9nn1jaIkMKj41ZYCVycsBA=
 k8s.io/apiserver v0.24.2/go.mod h1:pSuKzr3zV+L+MWqsEo0kHHYwCo77AT5qXbFXP2jbvFI=
+k8s.io/client-go v0.24.0/go.mod h1:VFPQET+cAFpYxh6Bq6f4xyMY80G6jKKktU6G0m00VDw=
 k8s.io/client-go v0.24.2 h1:CoXFSf8if+bLEbinDqN9ePIDGzcLtqhfd6jpfnwGOFA=
 k8s.io/client-go v0.24.2/go.mod h1:zg4Xaoo+umDsfCWr4fCnmLEtQXyCNXCvJuSsglNcV30=
+k8s.io/code-generator v0.24.0/go.mod h1:dpVhs00hTuTdTY6jvVxvTFCk6gSMrtfRydbhZwHI15w=
 k8s.io/code-generator v0.24.2/go.mod h1:dpVhs00hTuTdTY6jvVxvTFCk6gSMrtfRydbhZwHI15w=
+k8s.io/component-base v0.24.0/go.mod h1:Dgazgon0i7KYUsS8krG8muGiMVtUZxG037l1MKyXgrA=
```
pebrc, line 1420: Not sure where these changes come from.

pkg/controller/agent/fleet.go
```diff
@@ -276,22 +308,33 @@ func reconcileEnrollmentToken(ctx context.Context, agent agentv1alpha1.Agent, cl
...
 
-CREATE:
-	key, err := api.createEnrollmentAPIKey(ctx, policyID)
+FindOrCreate:
+	// try to re-use an existing key
+	newkey := false
+	key, err := api.findEnrollmentAPIKey(ctx, policyID)
 	if err != nil {
-		return EnrollmentAPIKey{}, err
+		ulog.FromContext(ctx).Info("Could not find existing Fleet enrollment API keys, creating new one", "error", err.Error())
```
barkbay, line 316: Should we only create a new one if the error is `404`?

pkg/controller/agent/fleet.go
```diff
@@ -276,22 +308,33 @@ func reconcileEnrollmentToken(ctx context.Context, agent agentv1alpha1.Agent, cl
 		}
 	}
 
-CREATE:
-	key, err := api.createEnrollmentAPIKey(ctx, policyID)
+FindOrCreate:
+	// try to re-use an existing key
+	newkey := false
```
barkbay, line 313: nit
```suggestion
	newKey := false
```

pkg/controller/agent/fleet.go
pebrc: I think we also want to create a new one if there is a bunch of old tokens but they are either all for other policies or have been disabled by the user.

pkg/controller/agent/fleet.go
```diff
@@ -276,22 +308,33 @@ func reconcileEnrollmentToken(ctx context.Context, agent agentv1alpha1.Agent, cl
...
 	// on the top level but we would then potentially create redundant enrollment tokens in the Fleet API
 	if agent.Annotations == nil {
 		agent.Annotations = map[string]string{}
 	}
 	agent.Annotations[FleetTokenAnnotation] = key.ID
 	err = client.Update(ctx, &agent)
-	if err != nil {
+	if err != nil && newkey {
 		// we have failed to store the token id in an annotation let's try to remove the token again
 		return EnrollmentAPIKey{}, k8serrors.NewAggregate([]error{err, api.deleteEnrollmentAPIKey(ctx, key.ID)})
```
barkbay, line 333: Is it still required since we attempt to reuse existing keys?

pebrc: You mean deleting the key? Hm you are right, we would try to reuse it next time round. I can remove that.

pkg/controller/agent/fleet.go
barkbay: What if `findEnrollmentAPIKey` fails with a timeout for example? In that case we don't know if there is an existing and valid token, we may create a new one while it is not necessary.

pkg/controller/common/tracing/http.go
```diff
@@ -0,0 +1,23 @@
...
+	"strings"
+)
+
+// RequestName returns method and path to be used in APM span names.
+func RequestName(request *http.Request) string {
+	if request == nil {
+		return ""
+	}
+	var b strings.Builder
+	b.Grow(len(request.Method) + len(request.URL.Path))
```
barkbay, line 18:
```suggestion
	b.Grow(len(request.Method) + len(request.URL.Path) + 1)
```

pkg/controller/agent/fleet.go
pebrc: Good point. We can check for the specific error as you suggested in your updated comment above.

barkbay (approved):
LGTM, I did a few tests, removed the annotation or deleted the tokens manually using the `api/fleet` API and it works as expected.

PR #5897, opened 2022-07-28
Also set unenroll_timeout for Fleet Server

barkbay:
Not sure why `unenroll_timeout` is not set for the Fleet server in our examples. It has the same effect as on "regular" agents: It eventually removes former Fleet servers that have been replaced by rolling out new Pods.

image


barkbay:
jenkins test this please

pebrc (approved):

PR #5899, opened 2022-07-28
ECK resources Helm chart - Beats

naemono:
related to https://github.com/elastic/cloud-on-k8s/issues/5505

This PR contains the 3rd version of the ECK-managed resources helm charts including

eck-beats chart for installing the different types of Beats
minor updates to the eck-stack chart.


deploy/eck-stack/examples/metricbeat_hosts.yaml
```diff
@@ -0,0 +1,219 @@
...
+    # Reference to ECK-managed Elasticsearch resource, ideally from {{ "elasticsearch.fullname" }}
+    #
+    elasticsearchRef:
+      name: quickstart
+
+eck-beats:
+  enabled: true
+  name: metricbeat
+  spec:
+    type: metricbeat
```
barkbay, line 60: I don't think it is possible, but still asking 😄 : what if, as part of a same `eck-stack` Helm release, I want to deploy other Beats? Is it possible to manage, let say, `Filebeat` and `Metricbeat` as part of the same release? My understanding is that this kind of behaviour is possible with https://github.com/elastic/cloud-on-k8s/pull/5889

deploy/eck-beats/examples/auditbeat_hosts.yaml
```diff
@@ -0,0 +1,113 @@
+name: auditbeat
+type: auditbeat
+version: 8.2.3
+spec:
+  elasticsearchRef:
+    name: elasticsearch
+  kibanaRef:
+    name: kibana
+  config:
+    # Since filebeat is used in the default values, this needs to be removed with an empty list.
```
barkbay, line 10: This is mentioned in several places, does it not feel a bit odd that the user must explicitly disable a `Beat` to deploy another one?

deploy/eck-beats/values.yaml
```diff
@@ -0,0 +1,144 @@
...
+      spec:
+        automountServiceAccountToken: true
+        terminationGracePeriodSeconds: 30
+        dnsPolicy: ClusterFirstWithHostNet
+        hostNetwork: true # Allows to provide richer host metadata
+        containers:
+        - name: filebeat
+          securityContext:
+            runAsUser: 0
+            # If using Red Hat OpenShift uncomment this:
```
barkbay, line 75: nit: this comment about OpenShift is not preserved in the resulting Beat resource, not sure if there is a way to keep comments.

deploy/eck-beats/examples/auditbeat_hosts.yaml
naemono: Over the course of the previous Helm Chart PRs, we've tried to maintain that installing with the default values gives some sort of useful "default" installation.  Though just recently, in the agent Helm Charts PR, the `config` section in the default values was removed, so maybe this makes sense here as well, but the default will not be useful, but will make the user experience for overrides much better.  Thoughts?

deploy/eck-stack/examples/metricbeat_hosts.yaml
naemono: The only way to do this, with the current setup of this Chart (not allowing a list of beats), would be to clone this repo, and update the `eck-stack/Chart.yaml` and add additional `dependencies` on the eck-beats chart, named with `aliases`, and install locally

(note the '.' meaning use local files)
```
helm install eck-stack .  --values /path/to/user/values.yaml
```

deploy/eck-beats/Chart.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v2
+name: eck-beats
+description: A Helm chart to deploy Elastic Beats managed by the ECK Operator.
+kubeVersion: ">= 1.20.0-0"
```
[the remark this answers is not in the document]
naemono: I updated this to include the reasoning behind this requirement, as I think it's important.  This would need to be updated during the normal release process for the eck-operator, when the requirement changes.

deploy/eck-beats/Chart.yaml
```diff
@@ -0,0 +1,10 @@
...
+type: application
+version: 0.1.0
```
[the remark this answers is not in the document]
naemono: This should be increased when we make changes to the underlying Chart.

deploy/eck-stack/templates/tests/beats_test.yaml
```diff
@@ -0,0 +1,50 @@
+suite: test beats
```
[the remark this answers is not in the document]
naemono: I've updated README to include formation about how to run and develop the tests.

deploy/eck-beats/examples/auditbeat_hosts.yaml
```diff
@@ -0,0 +1,113 @@
+name: auditbeat
+type: auditbeat
+version: 8.2.3
```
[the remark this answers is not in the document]
naemono: I've updated this script, and made the appropriate changes to the versions in this PR, but am hesitant to update the other charts in this PR, as we'd need to bump the chart versions when this happens.

deploy/eck-beats/examples/auditbeat_hosts.yaml
barkbay: Sorry, for the very late answer, I need to have a look at what is proposed for Agent. Defaulting to `Filebeat` feels a bit "opinionated" to me. It means we assume that the "most useful" default is Filebeat, not Metricbeat, and I'm not sure I get the reason why.

deploy/eck-beats/Chart.yaml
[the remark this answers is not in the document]
barkbay: If a user is using a version of the operator that still supports a previous k8s version (for example ECK `2.2`  which still supports K8S `1.19`), but the user needs to update the Helm Chart: does it prevent the user to use that new version of the Chart on `1.19`? I'm just trying to understand how the min. k8s version supported by the operator and the min. k8s version supported by a Chart are tied.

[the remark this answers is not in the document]
naemono: If a user was running eck `2.2` on K8S `1.19` and attempted to use these Helm Charts, they would fail (helm would throw a failure) because of this K8S version requirement in the Chart.yaml.

deploy/eck-beats/examples/auditbeat_hosts.yaml
naemono: I see why defaulting to `Filebeat` may not make the most sense here.  I'll update this to use`Metricbeat` by default.

naemono: actually @barkbay, I remember why I defaulted to filebeat once I went back and looked at this.  By defaulting to filebeat, we add the minimum amount to the default values and still have a useful "default" Beat install.  If we move this to metricbeat, many more default values would be needed to be added to make a useful default Beat install.  Let me know your thoughts here.

deploy/eck-beats/examples/filebeat_no_autodiscover.yaml
```diff
@@ -0,0 +1,49 @@
+apiVersion: beat.k8s.elastic.co/v1beta1
```
pebrc, line 1: This should be a values file not a custom resource

deploy/eck-beats/examples/metricbeat_hosts.yaml
```diff
@@ -0,0 +1,160 @@
...
+
+serviceAccount:
+  name: metricbeat
+
+clusterRoleBinding:
+  name: metricbeat
+  subjects:
+  - kind: ServiceAccount
+    name: metricbeat
+    namespace: default
```
pebrc, line 156: Ideally this would be the release namespace

deploy/eck-stack/examples/metricbeat_hosts.yaml
```diff
@@ -0,0 +1,219 @@
+eck-elasticsearch:
+  enabled: true
+
+  # Name of the Elasticsearch resource.
+  #
+  fullnameOverride: quickstart
+
+  # Version of Elasticsearch.
+  #
+  version: 8.2.3
```
pebrc, line 10:
```suggestion
  version: 8.5.0
```

deploy/eck-stack/examples/metricbeat_hosts.yaml
```diff
@@ -0,0 +1,219 @@
...
+  
+  serviceAccount:
+    name: metricbeat
+  
+  clusterRoleBinding:
+    name: metricbeat
+    subjects:
+    - kind: ServiceAccount
+      name: metricbeat
+      namespace: default
```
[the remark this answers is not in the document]
naemono: We actually do default to the `Release.Namespace`.  This was an oversight keeping this in the example.  It's ben removed.

https://github.com/elastic/cloud-on-k8s/pull/5899/files#diff-7a54af9bb234c05618afe6d34207f8931b534d186a560cebe506c9584414fb3fR26
```
    namespace: {{ .namespace | default $.Release.Namespace | quote }}
```

deploy/eck-beats/examples/filebeat_no_autodiscover.yaml
naemono: This has been resolved.

deploy/eck-beats/examples/metricbeat_hosts.yaml
naemono: Oversight.  It will default to `Release.Namespace`.  Removed.

deploy/eck-beats/examples/auditbeat_hosts.yaml
```diff
@@ -0,0 +1,111 @@
+name: auditbeat
+version: 8.3.3
+spec:
+  type: auditbeat
+  elasticsearchRef:
+    name: elasticsearch
```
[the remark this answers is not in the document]
naemono: I've updated this per most comments, but want to check into this further tomorrow, as these names are dependent on the name of the chart itself, unless doing a full name override, so I'll want to ensure this is consistent with the other charts.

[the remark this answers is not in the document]
naemono: @pebrc I've updated all values and all examples across all charts to be consistent for naming and updated all versions to latest stack 8.5.0, along with updating all tests to match/pass.

I've also bumped the versions of all charts where stack version changes were made, and updated the eck-stack Chart.yaml's dependency versions to match.

This should be ready for another set of 👀

hack/update-stack-version.sh
```diff
@@ -34,7 +34,7 @@ for_all_yaml_do() {
 	local function="$1"
 	# Directories containing Yaml files with version references to replace
 	# Note: hack/operatorhub/config.yaml will need to be updated manually
-	local dirs=(config/samples config/recipes config/e2e test/e2e)
+	local dirs=(config/samples config/recipes config/e2e test/e2e deploy/eck-stack deploy/eck-beats deploy/eck-kibana deploy/eck-elasticsearch)
```
pebrc, line 37: Why not just `deploy`? Assuming you want to exclude the operator Helm Chart why is are the Agent charts missing?

deploy/eck-stack/examples/metricbeat_hosts.yaml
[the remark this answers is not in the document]
pebrc: I think it's still there. Forgot to push a commit?

pebrc (commented):
LGTM

deploy/eck-agent/values.yaml
```diff
@@ -45,7 +45,7 @@ spec:
   # Reference to ECK-managed Elasticsearch instance.
   #
   elasticsearchRefs:
-  - name: elasticsearch
+  - name: eck-elasticsearch
```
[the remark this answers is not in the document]
naemono: These are simply the default values, and are just placeholders for users to provide their own names, as the names of these follow these rules from the values files: (they, by default, are prefixed with the `Release` name that the user provides, so there's no way to predict it)

```
# Overridable names of the Elasticsearch resource.
# By default, this is the Release name set for the chart,
# followed by 'eck-elasticsearch'.
#
# nameOverride will override the name of the Chart with the name set here,
# so nameOverride: quickstart, would convert to '{{ Release.name }}-quickstart'
#
# nameOverride: "quickstart"
#
# fullnameOverride will override both the release name, and the chart name,
# and will name the Elasticsearch resource exactly as specified.
#
# fullnameOverride: "quickstart"
```

Since we are also incrementing the semver already, it may make sense to indicate that these are breaking changes, if any users are using this default behavior.  I'll do so.

naemono:
> I think it's still there. Forgot to push a commit?

@pebrc this namespace reference was also in an example in `eck-beats/examples/*`, which is where I removed the previous one from.  I've removed this one as well.   Ty

Also, to the `hack/update-stack-version.sh` comment, I believe the initial intent was to not include the operator chart.  I've updated this to include agent/fleet.

brsolomon-deloitte:
@naemono Has this chart been pushed to the https://helm.elastic.co repository?

naemono:
> @naemono Has this chart been pushed to the https://helm.elastic.co repository?

@brsolomon-deloitte apologies for the delays in response.  You should see the Helm chart within the repository in about 1 hour.

PR #5906, opened 2022-08-01
Set Agent container privileged on OpenShift if running as root

barkbay:
This is an attempt to run Agent on OpenShift with an hostPath volume. An alternative would be to disable E2E Agent tests on OpenShift as it is the case for standalone Agents: https://github.com/elastic/cloud-on-k8s/blob/d5af986aad19d84ceee1d36fe2b94fef30c23775/test/e2e/agent/recipes_test.go#L212-L216

Related to https://github.com/elastic/cloud-on-k8s/pull/5890

### DIFF
--- a/test/e2e/test/agent/builder.go
+++ b/test/e2e/test/agent/builder.go
@@ -15,10 +15,12 @@ import (
 	rbacv1 "k8s.io/api/rbac/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/rand"
+	"k8s.io/utils/pointer"
 	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
 
 	agentv1alpha1 "github.com/elastic/cloud-on-k8s/v2/pkg/apis/agent/v1alpha1"
 	commonv1 "github.com/elastic/cloud-on-k8s/v2/pkg/apis/common/v1"
+	"github.com/elastic/cloud-on-k8s/v2/pkg/controller/agent"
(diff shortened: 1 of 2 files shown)


barkbay:
Tests results for `github.com/elastic/cloud-on-k8s/v2/test/e2e/agent` run on OpenShift `4.10.20` from my laptop:

image

test/e2e/test/agent/builder.go
```diff
@@ -195,13 +197,31 @@ func (b Builder) WithRestrictedSecurityContext() Builder {
...
+			b.PodTemplate.Spec.Containers,
+			corev1.Container{
+				Name:            agent.ContainerName,
+				SecurityContext: &securityContext,
+			},
+		)
+		return b
 	}
 
+	b.PodTemplate.Spec.Containers[containerIdx].SecurityContext = &securityContext
```
thbkrkr, line 212: Shouldn't we also fix [`WithContainerSecurityContext`](https://github.com/elastic/cloud-on-k8s/blob/4233f7b8e6bc8dc90712cb65c12dc0c89bfcc8f1/test/e2e/test/beat/builder.go#L208) in the `Beat` builder?

barkbay: > Shouldn't we also fix [WithContainerSecurityContext](https://github.com/elastic/cloud-on-k8s/blob/4233f7b8e6bc8dc90712cb65c12dc0c89bfcc8f1/test/e2e/test/beat/builder.go#L208) in the Beat builder?

It does not seem to be used, I removed it.

thbkrkr (approved):
LGTM

naemono (approved):
Also LGTM; makes sense.

PR #5922, opened 2022-08-03
Telemetry improvements

david-kow:


naemono (approved):
lgtm

naemono:
Just for future sake, I wish we had a simpler way than just this
```
strings.HasPrefix(val, "eck-elasticsearch-") || strings.HasPrefix(val, "eck-kibana-")
```
to determine this that wouldn't rely on this being updated when we release future resource helm charts.  Only thing I could suggest is added a static label to the existing, and future helm charts to look for in telemetry.

PR #5925, opened 2022-08-04
[CI] Increase AKS OS disk to 120GB

### LINKED ISSUE #5825: E2E tests in Azure/AKS are failing - pod never becomes ready

barkbay:
Fixes #5825

* Before:

```
root@aks-nodepool1-39942109-vmss000002:/# df -m
Filesystem     1M-blocks  Used Available Use% Mounted on
overlay            29599 24157      5427  82% /
```

* After:

```
root@aks-nodepool1-23298563-vmss000002:/# df -m
Filesystem     1M-blocks  Used Available Use% Mounted on
overlay           118917 23873     95029  21% /
```

### DIFF
--- a/hack/deployer/config/plans.yml
+++ b/hack/deployer/config/plans.yml
@@ -46,7 +46,7 @@ plans:
   clusterName: ci
   provider: aks
   kubernetesVersion: 1.22.6
-  machineType: Standard_D8d_v5
+  machineType: Standard_D8s_v3
   serviceAccount: true
   psp: false
   diskSetup: kubectl apply -k hack/deployer/config/local-disks
@@ -59,7 +59,7 @@ plans:
   clusterName: dev
   provider: aks
   kubernetesVersion: 1.22.6
-  machineType: Standard_D8d_v5
+  machineType: Standard_D8s_v3
   serviceAccount: false
   psp: false
   aks:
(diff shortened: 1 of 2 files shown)


naemono:
@barkbay Great catch.  I assumed Azure would use the full disk size given to it, and was unaware of this particular option in AKS.  Don't see this for EKS.  GKE seems to have a similar option:

https://github.com/elastic/cloud-on-k8s/blob/main/hack/deployer/runner/gke.go#L178

We use a disk size of 30 `--disk-size 30` for GKE.  Should be go ahead and change this one as well, as I can only imagine we'll run into this issue in the future in GKE.

naemono (approved):

barkbay:
> We use a disk size of 30 --disk-size 30 for GKE. Should be go ahead and change this one as well, as I can only imagine we'll run into this issue in the future in GKE.

I _think_ this problem is specific to AKS: it seems that there are a lot of `mcr.microsoft.com/*` images loaded on the K8S nodes when it is created/initialised (see my comment about the effect of `crictl rmi -prune` [here](https://github.com/elastic/cloud-on-k8s/pull/5826#issuecomment-1204862007)).

PR #5927, opened 2022-08-04
Remove references to ioutil

barkbay:
`io/ioutil` is deprecated since Go 1.16 and reported as such when `golangci-lint` is upgraded to its latest version.

### DIFF
--- a/hack/deployer/runner/kind.go
+++ b/hack/deployer/runner/kind.go
@@ -96,7 +96,7 @@ func (k *KindDriver) create() error {
 	if err != nil {
 		return err
 	}
-	defer os.RemoveAll(tmpManifest.Name())
+	defer os.Remove(tmpManifest.Name())
 
 	// Delete any previous e2e kind cluster with the same name
 	err = k.delete()
--- a/hack/deployer/runner/kind.go
+++ b/hack/deployer/runner/kind.go
@@ -7,7 +7,6 @@ package runner
 import (
 	"fmt"
 	"io/fs"
-	"io/ioutil"
 	"log"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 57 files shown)


hack/deployer/runner/kind.go
```diff
@@ -97,7 +96,7 @@ func (k *KindDriver) create() error {
 	if err != nil {
 		return err
 	}
-	defer os.Remove(tmpManifest.Name())
+	defer os.RemoveAll(tmpManifest.Name())
```
thbkrkr, line 99: Why this change? Isn't `tmpManifest` a file?

pkg/utils/test/webhook.go
```diff
@@ -77,6 +76,7 @@ func ValidationWebhookFailed(causeRegexes ...string) func(*testing.T, *admission
 }
 
 // RunValidationWebhookTests runs a series of ValidationWebhookTestCases
+//
```
thbkrkr, line 79: Do we need this?

thbkrkr (approved):
LGTM

hack/deployer/runner/kind.go
barkbay: I used a script to generate this PR, something may not have worked as expected here 😬  I'll fix it.

pkg/utils/test/webhook.go
barkbay: I'm confused, this line is automatically added by my IDE, I'm trying to understand why 🤔

barkbay: This is actually added by `gofmt` 🤷

barkbay:
jenkins test this please

pkg/utils/test/webhook.go
barkbay: > This is actually added by gofmt 🤷

Just realised that it is because I'm using [Go 1.19](https://tip.golang.org/doc/go1.19#go-doc):

> Go 1.19 adds support for links, lists, and clearer headings in doc comments. As part of this change, [gofmt](https://tip.golang.org/cmd/gofmt) now reformats doc comments to make their rendered meaning clearer.

PR #5934, opened 2022-08-09
Elastic Agent: use hostPath only as of 7.15.0

pebrc:
Adds a version gate to allow the hostPath volume only on 7.15 or later. 

Testing is purely through our e2e tests, which I can change if reviewers feel a unit test would be in order. I avoided it because the `buildPodTemplate` function seemed a bit hairy to unit test.

Excluded from release notes because it is a bug fix for yet unreleased functionality.

### DIFF
--- a/pkg/controller/agent/pod.go
+++ b/pkg/controller/agent/pod.go
@@ -146,7 +146,7 @@ func buildPodTemplate(params Params, fleetCerts *certificates.CertificatesSecret
 
 	v, err := version.Parse(params.Agent.Spec.Version)
 	if err != nil {
-		return corev1.PodTemplateSpec{}, pkgerrors.Wrap(err, "unexpected error while parsing specified version")
+		return corev1.PodTemplateSpec{}, err // error unlikely and should have been caught during validation
 	}
 	// volume with agent data path if version > 7.15 (available since 7.13 but non-functional as agent tries to fork child
 	// processes in data path directory and hostPath volumes are always mounted non-exec)
--- a/pkg/controller/agent/pod.go
+++ b/pkg/controller/agent/pod.go
@@ -28,6 +28,7 @@ import (
 	"github.com/elastic/cloud-on-k8s/v2/pkg/controller/common/labels"
 	"github.com/elastic/cloud-on-k8s/v2/pkg/controller/common/reconciler"
 	"github.com/elastic/cloud-on-k8s/v2/pkg/controller/common/tracing"
+	"github.com/elastic/cloud-on-k8s/v2/pkg/controller/common/version"
 	"github.com/elastic/cloud-on-k8s/v2/pkg/controller/common/volume"
 	"github.com/elastic/cloud-on-k8s/v2/pkg/utils/k8s"
 	"github.com/elastic/cloud-on-k8s/v2/pkg/utils/maps"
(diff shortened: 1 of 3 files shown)


thbkrkr:
I tried to test the fix using [recipes/elastic-agent/kubernetes-integration.yaml](https://github.com/elastic/cloud-on-k8s/blob/main/config/recipes/elastic-agent/kubernetes-integration.yaml#L18-L20) (without the volumeMounts L18-20) and with versions 7.13/7.14 and didn't get the initial bug using the `main` branch. What am I missing?

pebrc:
> I tried to test the fix using [recipes/elastic-agent/kubernetes-integration.yaml](https://github.com/elastic/cloud-on-k8s/blob/main/config/recipes/elastic-agent/kubernetes-integration.yaml#L18-L20) (without the volumeMounts L18-20) and with versions 7.13/7.14 and didn't get the initial bug using the main branch. What am I missing?

That's a good question. I used `TestSystemIntegrationConfig` with and without the fix and saw a test failure without it. Based on my current understanding it should happen with any integration. Let me try it myself again.

pkg/controller/agent/pod.go
```diff
@@ -143,8 +144,16 @@ func buildPodTemplate(params Params, fleetCerts *certificates.CertificatesSecret
 			WithArgs("-e", "-c", path.Join(ConfigMountPath, ConfigFileName))
 	}
 
-	// volume with agent data path
-	vols = append(vols, createDataVolume(params))
+	v, err := version.Parse(params.Agent.Spec.Version)
+	if err != nil {
+		return corev1.PodTemplateSpec{}, pkgerrors.Wrap(err, "unexpected error while parsing specified version")
```
thbkrkr, line 149: An invalid version should be caught by the validation layer. I wouldn't wrap the error here as we already do in all places where we parse the version and just return the error if there is one.

pebrc: Valid point, I will fix that.

thbkrkr (approved):
Despite my test which does not work and my little comment about error wrapping, code LGTM. I don't want to block this if we want to run nightly e2e tests asap.

pebrc:
So the [kubernetes-integration recipe](https://github.com/elastic/cloud-on-k8s/blob/0d11cf4e016d1d542e74ad2690052ffd081fffb6/config/recipes/elastic-agent/kubernetes-integration.yaml#L129-L152) is missing the leases RBAC permission. Once I added it the test failed as expected on `main` with

```
 cbeat: permission denied: failed to start '/usr/share/elastic-agent/state/data/install/metricbeat-7.14.2-linux-x86_64/metricbeat': fork/exec /usr/share/elastic-agent/state/data/install/metricbea │
│ t-7.14.2-linux-x86_64/metricbeat: permission denied                                                                                                                                                │
│ 2022-08-09T19:34:13.776Z    ERROR    log/reporter.go:36    2022-08-09T19:34:13Z - message: Application: metricbeat--7.14.2[e242e3e5-d369-4dc2-933f-f9531a488c0d]: State changed to FAILED: failed  │
│ to start '/usr/share/elastic-agent/state/data/install
```

thbkrkr (approved):
:+1: :+1:

pebrc:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|2.4|[](https://github.com/elastic/cloud-on-k8s/pull/5936)|

### Questions ?

PR #5947, opened 2022-08-12
Default Pod annotations should be a local variable

### LINKED ISSUE #5946: Concurrent map writes while generating Es Pod annotations
The [DefaultAnnotation](https://github.com/elastic/cloud-on-k8s/blob/a4036bcd2a2372de2277d363fe1f1747650016a2/pkg/controller/elasticsearch/nodespec/podspec.go#L177) map is shared across all the goroutines/workers.
It may lead to:
* Pods not being annotated correctly
* Operator crash:

```
/go/src/github.com/elastic/cloud-on-k8s/pkg/controller/elasticsearch/nodespec/podspec.go:186 +0x255 fp=0xc003eb7f88 sp=0xc003eb7778 pc=0x17aabd5
github.com/elastic/cloud-on-k8s/pkg/controller/elasticsearch/nodespec.buildAnnotations({{{0x1860929, 0xd}, {0xc005dad320, 0x1f}}, {{0xc00712d998, 0x14}, {0x0, 0x0}, {0xc00712d9b0, 0x11}, ...}, ...}, ...)
/usr/local/go/src/runtime/map_faststr.go:294 +0x38b fp=0xc003eb7778 sp=0xc003eb7710 pc=0x41304b
runtime.mapassign_faststr(0x19f9a20, 0xc000230ed0, {0x1c9dbd0, 0x28})
/usr/local/go/src/runtime/panic.go:1198 +0x71 fp=0xc003eb7710 sp=0xc003eb76e0 pc=0x435ab1
runtime.throw({0x1c7c756, 0xc0000cdd40})
goroutine 1580 [running]:
fatal error: concurrent map writes
```

barkbay:
Fixes #5946 

I'm still trying to understand how we could detect this kind of bug earlier, as part of our testing framework and CI workflow.

### DIFF
--- a/pkg/controller/elasticsearch/nodespec/defaults.go
+++ b/pkg/controller/elasticsearch/nodespec/defaults.go
@@ -12,7 +12,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	commonv1 "github.com/elastic/cloud-on-k8s/v2/pkg/apis/common/v1"
-	"github.com/elastic/cloud-on-k8s/v2/pkg/controller/common/annotation"
 	"github.com/elastic/cloud-on-k8s/v2/pkg/controller/common/defaults"
 	"github.com/elastic/cloud-on-k8s/v2/pkg/controller/elasticsearch/label"
 	"github.com/elastic/cloud-on-k8s/v2/pkg/controller/elasticsearch/settings"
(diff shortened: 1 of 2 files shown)


pkg/controller/elasticsearch/nodespec/defaults.go
```diff
@@ -40,12 +40,14 @@ var (
 			corev1.ResourceMemory: DefaultMemoryLimits,
 		},
 	}
+)
 
-	// DefaultAnnotations are the default annotations for the Elasticsearch pods
-	DefaultAnnotations = map[string]string{
+// defaultAnnotations are the default annotations for the Elasticsearch pods
+func defaultAnnotations() map[string]string {
```
barkbay, line 46: I've been tempted to inline it since it is only used at a single place.

pebrc: I think that would be reasonable.

pebrc (approved):

barkbay:
> I'm still trying to understand how we could detect this kind of bug earlier, as part of our testing framework and CI workflow.

It can be somewhat detected using the following integration test: https://gist.github.com/barkbay/b72b86540093b30bedffafb10f62a323

Note however that it is not consistently detected. I'll try to improve the test and open an issue to discuss it.

PR #5951, opened 2022-08-12
Base Kind client image on official Docker-in-Docker image

thbkrkr:
We are doing Docker in Docker to create Kind k8s clusters but 
we just copy the Docker binary for it and it can be problematic
in certain environment.

This commit changes the Kind client image to base it on the official
 Docker in Docker image that does more than just copy the 
binary (see [dockerd-entrypoint.sh](https://github.com/docker-library/docker/blob/master/20.10/dind/dockerd-entrypoint.sh) and [docker/hack/dind](https://github.com/moby/moby/blob/master/hack/dind)).

Relates to #5773.

### DIFF
--- a/hack/deployer/clients/kind/Dockerfile
+++ b/hack/deployer/clients/kind/Dockerfile
@@ -1,17 +1,7 @@
-FROM buildpack-deps:bullseye-curl as builder
+FROM docker:20.10-dind
 
 ARG CLIENT_VERSION
-ENV DOCKER_VERSION=19.03.13
 
-# Docker client to build and push images
-RUN curl -fsSLO https://download.docker.com/linux/static/stable/x86_64/docker-${DOCKER_VERSION}.tgz && \
-    tar xzf docker-${DOCKER_VERSION}.tgz --strip 1 -C /usr/local/bin docker/docker && \
-    rm docker-${DOCKER_VERSION}.tgz
-# Kind to run k8s cluster locally in Docker
-RUN curl -fsSLO https://github.com/kubernetes-sigs/kind/releases/download/v${CLIENT_VERSION}/kind-linux-amd64 && \
-    mv kind-linux-amd64 /usr/local/bin/kind && chmod +x /usr/local/bin/kind
-
-FROM scratch
-
-COPY --from=builder /usr/local/bin/kind .
-COPY --from=builder /usr/local/bin/docker .
+RUN apk add --no-cache curl && \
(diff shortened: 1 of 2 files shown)


pebrc (commented):
A quick test on macOS with Docker Desktop did work for me so from that angle 👍 But I wonder if bind-mounting the docker socket as we did before was not the simpler solution as this is actual not  Docker in Docker but more like Docker next to Docker (all containers created are siblings to the original container). So I guess it would be good to better understand the problems you mentioned.

thbkrkr:
> But I wonder if bind-mounting the docker socket as we did before was not the simpler solution as this is actual not Docker in Docker but more like Docker next to Docker (all containers created are siblings to the original container). 

I lacked precision. I didn't have the intention to change that. It's Docker next to Docker that didn't work in the new environment.

> So I guess it would be good to better understand the problems you mentioned.

```
Got permission denied while trying to connect to the Docker daemon socket at unix:///var/run/docker.sock: Get http://%2Fvar%2Frun%2Fdocker.sock/v1.40/containers/json: dial unix /var/run/docker.sock: connect: permission denied
```

I then tried `--privileged` and got another somewhat cryptic error that I no longer reproduce. Now I get:
```
docker: Error response from daemon: privileged mode is incompatible with user namespaces.  You must run the container in the host namespace when running privileged mode.
```

Which is pretty clear. The host's Docker daemon is configured to run containers in user namespaces.

```
> docker info
...
 Security Options:
  apparmor
  seccomp
   Profile: default
  userns
```

So, a workaround to be able to do Docker next to Docker is to run the container with `--userns=host`.

One thing that is still weird though is that using the dind image, we don't need `--userns=host`.

:red_circle: `docker run -v /var/run/docker.sock:/var/run/docker.sock --entrypoint /docker docker.elastic.co/eck-ci/deployer-kind:0.11.1-e496a752325be8ae ps`
:heavy_check_mark: `docker run --userns host -v /var/run/docker.sock:/var/run/docker.sock --entrypoint /docker docker.elastic.co/eck-ci/deployer-kind:0.11.1-e496a752325be8ae ps`
:heavy_check_mark: `docker run -v /var/run/docker.sock:/var/run/docker.sock docker:20.10-dind docker ps -a`
:heavy_check_mark: `docker run --userns host -v /var/run/docker.sock:/var/run/docker.sock docker:20.10-dind docker ps -a`

Closing because we don't really need this change.

PR #5962, opened 2022-08-19
Inherit all environment variables from ES container in initContainers

### LINKED ISSUE #5577: Keystore init-container should inherit Elasticsearch environment variables
When referenced in the Elasticsearch configuration file, environment variables must not only be set in the elasticsearch container but also in the keystore init-container. Otherwise `/usr/share/elasticsearch/bin/elasticsearch-keystore` may not be able to initialize properly as described in https://github.com/elastic/cloud-on-k8s/issues/5558

A solution would be to have the keystore init-container automatically inherit the Elasticsearch environment variables.

pebrc:
Fixes #5577


pkg/controller/elasticsearch/nodespec/podspec.go
```diff
@@ -112,7 +112,8 @@ func BuildPodTemplateSpec(
 		WithVolumes(volumes...).
 		WithVolumeMounts(volumeMounts...).
 		WithInitContainers(initContainers...).
-		WithInitContainerDefaults(corev1.EnvVar{Name: settings.HeadlessServiceName, Value: headlessServiceName}).
+		// inherit all env vars from main containers to allow Elasticsearch tools that read ES config to work in initContainers
+		WithInitContainerDefaults(builder.MainContainer().Env...).
```
barkbay, line 116: nit: not sure if it's really important in that context but `builder.MainContainer()` may return `nil`.

pkg/controller/common/defaults/pod_template.go
```diff
@@ -57,8 +57,8 @@ func NewPodTemplateBuilder(base corev1.PodTemplateSpec, containerName string) *P
 	return builder.setDefaults()
 }
 
-// getContainer retrieves the main Container from the pod template
-func (b *PodTemplateBuilder) getContainer() *corev1.Container {
+// MainContainer retrieves the main Container from the pod template
```
barkbay, line 60:
```suggestion
// MainContainer retrieves the main Container from the pod template or nil if not found.
```

barkbay (approved):
I'm not sure if we should check for the returned value of `MainContainer`, which maybe `nil` but LGTM overall.

I think we should also update our documentation [here](https://github.com/elastic/cloud-on-k8s/blob/07bc2ab8144e9a9a0a2e21d06f5e019d4608f9c8/docs/orchestrating-elastic-stack-applications/elasticsearch/es-secure-settings.asciidoc?plain=1#L75) and [here](https://github.com/elastic/cloud-on-k8s/blob/07bc2ab8144e9a9a0a2e21d06f5e019d4608f9c8/docs/orchestrating-elastic-stack-applications/elasticsearch/advanced-node-scheduling.asciidoc?plain=1#L241-L248) to mention this improvement. We can also add the (expected) release version to our documentation, in order to mention that this PR will trigger a [rolling restart](https://www.elastic.co/guide/en/cloud-on-k8s/master/k8s-upgrading-eck.html#k8s-beta-to-ga-rolling-restart) of the managed Elasticsearch clusters. This will be the case even if user is not using secure settings or advanced topologies. `elastic-internal-init-filesystem`, I think, will not take advantage of this PR, but I guess it will be the case for `elastic-internal-suspend`, so I guess we cannot exclude this last one.

pkg/controller/elasticsearch/nodespec/podspec.go
pebrc: That's true but given that we construct the pod builder ourselves a few lines above and ensure through defaulting that we have a main container I would say we can maybe ignore it here?

PR #5966, opened 2022-08-22
Report incorrect license type in logs and events

### LINKED ISSUE #5963: Create a more helpful error message if wrong license is applied
Elastic has two types of licenses: single cluster licenses for Elasticsearch and orchestration licenses (which contain a bunch of cluster licenses) for orchestration products like ECK and ECE. 

If a user of the ECK operator accidentally applies a cluster license to the operator the error messages are a confusing and we do not clearly indicate that the wrong kind of license has been applied. We should be able to improve that by looking at the structure of the license the user tries to apply and give clear and actionable feedback to avoid this confusion.

pebrc:
Fixes #5963 

by checking for the license type when parsing the user submitted license files and logging a more helpful error message. Also creates events in the license controller based on the errors during parsing. The license parsing also happens during license checks outside the license controller and the custom error allows us to limit the event creation to the license controller.

### DIFF
--- a/pkg/controller/common/events/events.go
+++ b/pkg/controller/common/events/events.go
@@ -8,6 +8,8 @@ package events
 const (
 	// EventReasonDelayed describes events where a requested change was delayed e.g. to prevent data loss.
 	EventReasonDelayed = "Delayed"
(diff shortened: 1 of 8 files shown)


barkbay (approved):
LGTM

PR #5977, opened 2022-08-29
Temporary fixes for Redhat release automation

naemono:
The filtering that we're currently using in our Redhat automation is broken, and we have an open case working to resolve this.  In the interim, we can grab the most recent 4 published images, and loop through them to find the tag/version we are looking for.

### DIFF
--- a/hack/operatorhub/main.go
+++ b/hack/operatorhub/main.go
@@ -51,6 +51,7 @@ const (
 	yamlSeparator = "---\n"
 
 	ImagesEndpoint = "https://catalog.redhat.com/api/containers/v1/projects/certification/id/%s/images"
+	MAX_IMAGES     = 4
 )
 
 type cmdArgs struct {
@@ -600,12 +601,14 @@ func getImageDigest(apiKey, projectId, version string) (string, error) {
 	req.Header.Set("X-API-KEY", apiKey)
 
 	q := req.URL.Query()
-	// "repositories.tags.name" filter does not work any longer.  See https://github.com/elastic/cloud-on-k8s/issues/XXXX.
+	// "repositories.tags.name" filter does not work any longer.  See https://github.com/elastic/cloud-on-k8s/issues/5985.
 	// In the interim, we're going to get the most recent 4 published, undeleted images
 	// sorted by descending creation date, and loop through them.
-	// q.Add("filter", fmt.Sprintf("repositories.tags.name==%s;deleted==false", version))
+	//
+	// TODO: Revert to "q.Add("filter", fmt.Sprintf("repositories.tags.name==%s;deleted==false", version))",
+	// remove the page_size and sort_by query parameters and the MAX_IMAGES constant once resolved.
 	q.Add("filter", "repositories.published==true;deleted==false")
-	q.Add("page_size", "4")
+	q.Add("page_size", fmt.Sprintf("%d", MAX_IMAGES))
 	q.Add("sort_by", "creation_date[desc]")
 	req.URL.RawQuery = q.Encode()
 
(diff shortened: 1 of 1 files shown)


hack/operatorhub/main.go
```diff
@@ -595,11 +600,17 @@ func getImageDigest(apiKey, projectId, version string) (string, error) {
 	req.Header.Set("X-API-KEY", apiKey)
 
 	q := req.URL.Query()
-	q.Add("filter", fmt.Sprintf("repositories.tags.name==%s;deleted==false", version))
+	// This filter does not work any longer.  Redhat case is open to find working filter.
```
barkbay, line 603: Maybe track this problem with an issue, and add reference it here.
```suggestion
	// "repositories.tags.name" filter does not work any longer.  See https://github.com/elastic/cloud-on-k8s/issues/XXXX.
```

hack/operatorhub/main.go
```diff
@@ -595,11 +600,17 @@ func getImageDigest(apiKey, projectId, version string) (string, error) {
...
+	// In the interim, we're going to get the most recent 4 published, undeleted images
+	// sorted by descending creation date, and loop through them.
+	// q.Add("filter", fmt.Sprintf("repositories.tags.name==%s;deleted==false", version))
+	q.Add("filter", "repositories.published==true;deleted==false")
+	q.Add("page_size", "4")
```
barkbay, line 608: Maybe make it a constant and add it in the error message below?
```suggestion
	q.Add("page_size", MAX_IMAGES)
```

hack/operatorhub/main.go
```diff
@@ -615,19 +626,23 @@ func getImageDigest(apiKey, projectId, version string) (string, error) {
...
+				if tag.Name == version {
+					imageDigest = image.DockerImageDigest
+				}
+			}
+		}
 	}
-	imageDigest := images.Data[0].DockerImageDigest
 	if imageDigest == "" {
-		return "", fmt.Errorf("image digest for %s is empty", version)
+		return "", fmt.Errorf("no recent images found for %s", version)
```
barkbay, line 645: nit: It might be helpful to let the user know that only 4 images have been retrieved from the API.

```suggestion
		return "", fmt.Errorf("no published and undeleted image for version %s found in the %d most recent items returned by the catalog API", version, MAX_IMAGES)
```

hack/operatorhub/main.go
```diff
@@ -595,11 +600,17 @@ func getImageDigest(apiKey, projectId, version string) (string, error) {
 	req.Header.Set("X-API-KEY", apiKey)
 
 	q := req.URL.Query()
-	q.Add("filter", fmt.Sprintf("repositories.tags.name==%s;deleted==false", version))
+	// This filter does not work any longer.  Redhat case is open to find working filter.
+	// In the interim, we're going to get the most recent 4 published, undeleted images
+	// sorted by descending creation date, and loop through them.
+	// q.Add("filter", fmt.Sprintf("repositories.tags.name==%s;deleted==false", version))
```
barkbay, line 606:
```suggestion
	// TODO: Revert to "q.Add("filter", fmt.Sprintf("repositories.tags.name==%s;deleted==false", version))", remove the page_size filter and the MAX_IMAGES constant once resolved.
```

barkbay (approved):
LGTM

hack/operatorhub/main.go
```diff
@@ -595,11 +600,17 @@ func getImageDigest(apiKey, projectId, version string) (string, error) {
...
+	q.Add("filter", "repositories.published==true;deleted==false")
+	q.Add("page_size", "4")
+	q.Add("sort_by", "creation_date[desc]")
 	req.URL.RawQuery = q.Encode()
 
 	client := http.Client{
-		Timeout: 30 * time.Second,
+		Timeout: 60 * time.Second,
```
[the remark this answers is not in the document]
naemono: I ran into a timeout failure working through this yesterday because of the 30s timeout.  Unfortunately their api sometimes returns very very slowly.

naemono:
Closing this without merging, as Redhat fixed their filters in the api, so this is uneeded.

PR #5978, opened 2022-08-29
Autoscaling Elasticsearch: Introduce a dedicated custom resource

barkbay:
This PR introduces a dedicated Kubernetes Resource Definition, and an associated controller, to configure Elasticsearch autoscaling with ECK.

## Naming and group

This PR introduces a new resource named `ElasticsearchAutoscaler`. It lives in a group named `autoscaling.k8s.elastic.co`:

```yaml
apiVersion: autoscaling.k8s.elastic.co/v1alpha1
kind: ElasticsearchAutoscaler
```

 This is mostly to let the door open to any additional autoscaling resources that we may want to add in the future.

## ElasticsearchAutoscaler specification

The new CRD is very similar to the actual autoscaling annotation, with the obvious difference that a `elasticsearchRef` must be provided by the user:

```yaml
apiVersion: autoscaling.k8s.elastic.co/v1alpha1
kind: ElasticsearchAutoscaler
metadata:
  name: autoscaling-sample
spec:
  elasticsearchRef:
    name: elasticsearch-sample
  policies:
    - name: di
      roles: ["data", "ingest" , "transform"]
      ## Optional: section below can be used if fine-grain tuning of the Elasticsearch deciders is required.
      #deciders:
      #  proactive_storage:
      #    forecast_window: 5m
      resources:
        nodeCount:
          min: 3
          max: 8
        cpu:
          min: 2
          max: 8
        memory:
          min: 2Gi
          max: 16Gi
        storage:
          min: 64Gi
          max: 512Gi
    - name: ml
      roles:
        - ml
      resources:
        nodeCount:
          min: 1
          max: 9
        cpu:
          min: 1
          max: 4
        memory:
          min: 2Gi
          max: 8Gi
        storage:
          min: 1Gi
          max: 1Gi
```

Only one cluster can be managed by a given `ElasticsearchAutoscaler`. This is similar to the K8S `HorizontalPodAutoscaler` and `VerticalPodAutoscaler`. It also makes it easier to understand the autoscaler status and the relationship between the autoscaler and the Elasticsearch cluster.

## Status

The status consists of 2 main elements:
1. `conditions` provides an overall state of the reconciliation status.
2. `policies` holds the calculated resources and any error/important messages.

```yaml
status:
  conditions:
  - lastTransitionTime: "2022-08-29T11:11:43Z"
    status: "False"
    type: Limited
  - lastTransitionTime: "2022-08-27T17:02:32Z"
    status: "True"
    type: Healthy
  - lastTransitionTime: "2022-08-27T17:01:41Z"
    status: "True"
    type: Active
  - lastTransitionTime: "2022-08-27T17:02:32Z"
    message: Elasticsearch is available
    status: "True"
    type: Online
  observedGeneration: 3
  policies:
  - lastModificationTime: "2022-08-29T11:19:07Z"
    name: di
    nodeSets:
    - name: di
      nodeCount: 7
    resources:
      limits:
        cpu: "2"
        memory: 8Gi
      requests:
        cpu: "2"
        memory: 8Gi
        storage: 5Gi
  - lastModificationTime: "2022-08-29T11:19:07Z"
    name: ml
    nodeSets:
    - name: ml
      nodeCount: 0
    resources:
      limits:
        cpu: "1"
        memory: 2Gi
      requests:
        cpu: "1"
        memory: 2Gi
        storage: 1Gi
```

## Printed columns

The conditions `Limited`, `Active` and `Healthy` are printed as part of the output of `kubectl get elasticsearchautoscaler.autoscaling.k8s.elastic.co/<autoscaler_name>`:

```
NAME                                                                    TARGET                 ACTIVE   HEALTHY   LIMITED
elasticsearchautoscaler.autoscaling.k8s.elastic.co/autoscaling-sample   elasticsearch-sample   True     True      True
```

* `TARGET`: the name of Elasticsearch cluster which is autoscaled
* `ACTIVE`: True when the ElasticsearchAutoscaler resource is managed by the operator, and the target Elasticsearch cluster does exist.
* `HEALTHY`: True if resources have been calculated for all the autoscaling policies and no error has been encountered during the reconciliation process.
* `LIMITED`: True when a resource limit is reached

For each printed condition there is an additional message in the related condition.

## Noteworthy change

By default there is now a ratio of `1:1` between the CPU `limit` and the `request`. This is to comply with the current desired nodes API implementation.

## Testing

* Elasticsearch autoscaling is an Enterprise feature, both at the ECK and the Elasticsearch level, in dev mode [you can start a trial](https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-licensing.html#k8s-start-trial).
* Autoscaling events can be generated by using [the fixed decider](https://www.elastic.co/guide/en/elasticsearch/reference/current/autoscaling-fixed-decider.html), this is what is done in the e2e test for an example: https://github.com/elastic/cloud-on-k8s/blob/79ef072d4d21570cdfc407ab61feff9ef913cf50/test/e2e/es/autoscaling_test.go#L89-L90

TODO:
- [x] Add unit tests for the new controller
- [x] Add webhook validation
- [x] Add a new E2E Test
- [x] Update documentation, can be done in a dedicated PR: https://github.com/elastic/cloud-on-k8s/issues/5997


barkbay:
Jenkins test this please

barkbay:
`elasticsearch-ci/docs` fails because [`crd-ref-docs`](https://github.com/elastic/crd-ref-docs) duplicates the `Conditions` section, not sure why 🤔 

```asciidoc
[id="{anchor_prefix}-github-com-elastic-cloud-on-k8s-v2-pkg-apis-common-v1alpha1-conditions"]
=== Conditions 



.Appears In:
****
- xref:{anchor_prefix}-github-com-elastic-cloud-on-k8s-v2-pkg-apis-elasticsearch-v1-elasticsearchstatus[$$ElasticsearchStatus$$]
****

[cols="25a,75a", options="header"]
|===
| Field | Description
| *`type`* __xref:{anchor_prefix}-github-com-elastic-cloud-on-k8s-v2-pkg-apis-common-v1alpha1-conditiontype[$$ConditionType$$]__ | 
| *`status`* __link:https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#conditionstatus-v1-core[$$ConditionStatus$$]__ | 
| *`lastTransitionTime`* __link:https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#time-v1-meta[$$Time$$]__ | 
| *`message`* __string__ | 
|===


[id="{anchor_prefix}-github-com-elastic-cloud-on-k8s-v2-pkg-apis-common-v1alpha1-conditions"]
=== Conditions 



.Appears In:
****
- xref:{anchor_prefix}-github-com-elastic-cloud-on-k8s-v2-pkg-apis-elasticsearch-v1-elasticsearchstatus[$$ElasticsearchStatus$$]
****

[cols="25a,75a", options="header"]
|===
| Field | Description
| *`type`* __xref:{anchor_prefix}-github-com-elastic-cloud-on-k8s-v2-pkg-apis-common-v1alpha1-conditiontype[$$ConditionType$$]__ | 
| *`status`* __link:https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#conditionstatus-v1-core[$$ConditionStatus$$]__ | 
| *`lastTransitionTime`* __link:https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#time-v1-meta[$$Time$$]__ | 
| *`message`* __string__ | 
|===
```

barkbay:
Last commit enables advanced validation:
* Either by using the admission controller:

```
for: "config/recipes/autoscaling/elasticsearch.yaml": admission webhook "elastic-esa-validation-v1alpha1.k8s.elastic.co" denied the request:
ElasticsearchAutoscaler.autoscaling.k8s.elastic.co "autoscaling-sample" is invalid:
spec.policies[0].resources.nodeCount.min: Invalid value: -1: min count must be equal or greater than 0
``` 

* Or at the `conditions` level if the webhook is disabled:

```yaml
status:
  conditions:
  - lastTransitionTime: "2022-08-31T09:58:03Z"
    message: Autoscaler is unhealthy
    status: "True"
    type: Active
  - lastTransitionTime: "2022-08-31T09:58:03Z"
    message: 'ElasticsearchAutoscaler.autoscaling.k8s.elastic.co "autoscaling-sample"
      is invalid: spec.policies[0].resources.nodeCount.min: Invalid value: -1: min
      count must be equal or greater than 0'
    status: "False"
    type: Healthy
  - lastTransitionTime: "2022-08-31T09:58:03Z"
    message: Autoscaler is unhealthy
    status: "False"
    type: Online
  - lastTransitionTime: "2022-08-30T11:53:45Z"
    status: "False"
    type: Limited
  observedGeneration: 2
```

Note that the custom resource is considered as "unhealthy" in that case:

```
NAME                                                                    TARGET                 ACTIVE   HEALTHY   LIMITED
elasticsearchautoscaler.autoscaling.k8s.elastic.co/autoscaling-sample   elasticsearch-sample   True     False     False
```

Kushmaro:
Some comments due to some of my specific autoscaling (in Elasticsearch) ignorance :)

`deciders` - i'd like to mask this for UX, it sounds to me from a UX perspective like its a "type of scaler", not sure but I think we might want to find something a bit clearer here.
`proactive_storage` - i'd simplify this to "storage", since this is just storage based autoscaling.

Also, I'm less familiar with how CPU relates to storage based autoscaling, but I'm guessing this is a mandatory field, I'm just wondering - why? 

```yaml
apiVersion: autoscaling.k8s.elastic.co/v1alpha1
kind: ElasticsearchAutoscaler
metadata:
  name: autoscaling-sample
spec:
  elasticsearchRef:
    name: elasticsearch-sample
  policies:
    - name: di
      roles: ["data", "ingest" , "transform"]
      deciders:
        proactive_storage:
          forecast_window: 5m
      resources:
        nodeCount:
          min: 3
          max: 8
        cpu:
          min: 2
          max: 8
        memory:
          min: 2Gi
          max: 16Gi
        storage:
          min: 64Gi
          max: 512Gi
    - name: ml
      roles:
        - ml
      deciders:
        ml:
          down_scale_delay: 10m
      resources:
        nodeCount:
          min: 1
          max: 9
        cpu:
          min: 1
          max: 4
        memory:
          min: 2Gi
          max: 8Gi
        storage:
          min: 1Gi
          max: 1Gi
```

barkbay:
Each Elasticsearch autoscaling [decider](https://www.elastic.co/guide/en/elasticsearch/reference/current/autoscaling-deciders.html) comes with default settings, I set some of them in my example, but the only purpose was to show that they can be changed by the end user when needed.
It's fine to remove them if you want to rely on the default settings:

```yaml
apiVersion: autoscaling.k8s.elastic.co/v1alpha1
kind: ElasticsearchAutoscaler
metadata:
  name: autoscaling-sample
spec:
  elasticsearchRef:
    name: elasticsearch-sample
  policies:
    - name: di
      roles: ["data", "ingest" , "transform"]
      resources:
        nodeCount:
          min: 3
          max: 8
        cpu:
          min: 2
          max: 8
        memory:
          min: 2Gi
          max: 16Gi
        storage:
          min: 64Gi
          max: 512Gi
    - name: ml
      roles:
        - ml
      resources:
        nodeCount:
          min: 1
          max: 9
        cpu:
          min: 1
          max: 4
        memory:
          min: 2Gi
          max: 8Gi
        storage:
          min: 1Gi
          max: 1Gi
```

I also added a comment in the example, see this commit https://github.com/elastic/cloud-on-k8s/commit/79ef072d4d21570cdfc407ab61feff9ef913cf50
Let me know if you think it's clearer.

barkbay:
run/e2e-tests tags=es

naemono:
@barkbay the initial part that caught me when beginning to look at this, is that the cpu validation seems to differ from standard k8s `resources.*.cpu` validation.

```
❯ kc apply -f config/custom/elasticsearch/autoscaling/test.yaml -n default
The ElasticsearchAutoscaler "autoscaling-test" is invalid:
* spec.policies.resources.cpu.min: Invalid value: "number": spec.policies.resources.cpu.min in body must be of type integer,string: "number"
* <nil>: Invalid value: "": "spec.policies.resources.cpu.min" must validate at least one schema (anyOf)
* spec.policies.resources.cpu.min: Invalid value: "number": spec.policies.resources.cpu.min in body must be of type integer: "number"
```

Spec:
```
spec:
  elasticsearchRef:
    name: testing
  policies:
    - name: testing
      roles: ["master", "data"]
      resources:
        cpu:
          min: 0.1
          max: 2
```

Is this intentional?

pkg/apis/common/v1alpha1/resources.go
```diff
@@ -0,0 +1,301 @@
...
+// ResourceListInt64 is a set of (resource name, quantity) pairs.
+type ResourceListInt64 map[corev1.ResourceName]int64
+
+// NodeResourcesInt64 is mostly use in logs to print comparable values which can be used in dashboards.
+type NodeResourcesInt64 struct {
+	Requests ResourceListInt64 `json:"requests,omitempty"`
+	Limits   ResourceListInt64 `json:"limits,omitempty"`
+}
+
+// ToInt64 converts all the resource quantities to int64, mostly to be logged and build dashboard.
```
naemono, line 239:
```suggestion
// ToInt64 converts all the resource quantities to int64, mostly to be logged and to build dashboards.
```

pkg/apis/elasticsearch/v1/autoscaling_test.go
```diff
@@ -8,15 +8,96 @@ import (
 	"reflect"
 	"testing"
 
-	"github.com/stretchr/testify/assert"
 	"k8s.io/apimachinery/pkg/api/resource"
 
+	"github.com/blang/semver/v4"
+
+	"github.com/stretchr/testify/assert"
```
naemono, line 15: Strange addition of newlines here

pkg/apis/elasticsearch/v1/elasticsearch_types.go
```diff
@@ -459,17 +464,70 @@ func (es *Elasticsearch) ElasticServiceAccount() (commonv1.ServiceAccountName, e
...
+	}
+	return autoscalingSpecification.PollingPeriod, nil
+}
+
+func (es Elasticsearch) GetElasticsearchAutoscalerStatus() (v1alpha1.ElasticsearchAutoscalerStatus, error) {
+	autoscalingStatus, err := ElasticsearchAutoscalerStatusFrom(es)
+	if err != nil {
+		return v1alpha1.ElasticsearchAutoscalerStatus{}, err
+	}
+	return autoscalingStatus, nil
```
naemono, line 516:
```suggestion
	return ElasticsearchAutoscalerStatusFrom(es)
```

pkg/controller/autoscaling/elasticsearch/controller.go
```diff
@@ -9,25 +9,37 @@ import (
...
-	"github.com/elastic/cloud-on-k8s/v2/pkg/controller/common/reconciler"
 	"github.com/elastic/cloud-on-k8s/v2/pkg/controller/common/tracing"
 	"github.com/elastic/cloud-on-k8s/v2/pkg/controller/common/version"
 	esclient "github.com/elastic/cloud-on-k8s/v2/pkg/controller/elasticsearch/client"
 	"github.com/elastic/cloud-on-k8s/v2/pkg/controller/elasticsearch/services"
 	"github.com/elastic/cloud-on-k8s/v2/pkg/controller/elasticsearch/user"
-	"github.com/elastic/cloud-on-k8s/v2/pkg/controller/elasticsearch/validation"
 	"github.com/elastic/cloud-on-k8s/v2/pkg/dev"
 	"github.com/elastic/cloud-on-k8s/v2/pkg/utils/k8s"
 	logconf "github.com/elastic/cloud-on-k8s/v2/pkg/utils/log"
```
naemono, line 45: The ordering of these imports seems off.

pkg/controller/autoscaling/elasticsearch/controller.go
```diff
@@ -62,38 +75,82 @@ type ReconcileElasticsearch struct {
...
-	// Fetch the Elasticsearch instance
-	var es esv1.Elasticsearch
-	requeue, err := r.fetchElasticsearch(ctx, request, &es)
-	if err != nil || requeue {
+	log := logconf.FromContext(ctx)
+	// Fetch the ElasticsearchAutoscaler instance
+	var esa autoscalingv1alpha1.ElasticsearchAutoscaler
+	if err := r.Get(ctx, request.NamespacedName, &esa); err != nil {
+		if apierrors.IsNotFound(err) {
+			log.V(1).Info("ElasticsearchAutoscaler %s/%s not found")
```
naemono, line 121: Are you missing `fmt` here?  the `%s/%s` aren't doing anything.

pkg/controller/autoscaling/elasticsearch/controller.go
```diff
@@ -102,70 +159,219 @@ func (r *ReconcileElasticsearch) Reconcile(ctx context.Context, request reconcil
...
+	updateStatus, err := r.updateStatus(ctx, log, esa)
 	if err != nil {
-		return reconcile.Result{}, tracing.CaptureError(ctx, err)
+		return reconcile.Result{}, err
 	}
+	results.WithResult(updateStatus)
+
+	if reconciledEs == nil {
+		// no reconciled Elasticsearch has been returned, likely to be the case if a fatal error prevented
+		// to calculate the new compute and storage resources.
```
naemono, line 244: Might want to re-word this comment.

naemono (changes requested):
Not quite done with review, but some early comments.

Functionally so far I've had no issues testing this autoscaling functionality.

pkg/apis/elasticsearch/v1/elasticsearch_types.go
barkbay: It's done on purpose to be sure we do not return a half built/ corrupted status.

pkg/apis/common/v1alpha1/autoscaling_status.go
```diff
@@ -0,0 +1,358 @@
...
+		for eventType := range policyStateBuilder.states {
+			if eventType == VerticalScalingLimitReached || eventType == HorizontalScalingLimitReached {
+				asb.scalingLimitEvents.Add(policyStateBuilder.policyName)
+			} else {
+				asb.nonScalingLimitEvents.Add(policyStateBuilder.policyName)
+			}
+		}
+		policyStates[i] = policyStateBuilder.Build()
+		i++
+	}
```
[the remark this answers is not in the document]
barkbay: `policyStatusBuilder` is a `map[string]*AutoscalingPolicyStatusBuilder`, not a `slice`. Therefore `i` would be of type `string`.

pkg/apis/common/v1alpha1/autoscaling_status.go
```diff
@@ -0,0 +1,358 @@
...
+	EmptyResponse                 AutoscalingEventType = "EmptyResponse"
+	HorizontalScalingLimitReached AutoscalingEventType = "HorizontalScalingLimitReached"
+	MemoryRequired                AutoscalingEventType = "MemoryRequired"
+	NoNodeSet                     AutoscalingEventType = "NoNodeSet"
+	OverlappingPolicies           AutoscalingEventType = "OverlappingPolicies"
+	StorageRequired               AutoscalingEventType = "StorageRequired"
+	UnexpectedNodeStorageCapacity AutoscalingEventType = "UnexpectedNodeStorageCapacity"
+	VerticalScalingLimitReached   AutoscalingEventType = "VerticalScalingLimitReached"
+
+	NotOnlineErrorMessage string = "An error prevented the use of the autoscaling API"
```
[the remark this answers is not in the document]
barkbay: > Is this a default error message used when autoscaling errors?

This is only used in the `ElasticsearchAutoscalerHealthy` Condition to let the user know that the autoscaler was not able to connect, or use the result, of the [autoscaling API](https://www.elastic.co/guide/en/elasticsearch/reference/current/autoscaling-apis.html).

pkg/controller/autoscaling/elasticsearch/controller.go
barkbay: I tried to improve this comment, not sure if it's enough though... let me know if you have a suggestion 🙇

barkbay:
> @barkbay the initial part that caught me when beginning to look at this, is that the cpu validation seems to differ from standard k8s resources.*.cpu validation.

> * <nil>: Invalid value: "": "spec.policies.resources.cpu.min" must validate at least one schema (anyOf)
> * spec.policies.resources.cpu.min: Invalid value: "number": spec.policies.resources.cpu.min in body must be of type integer: "number"

> Is this intentional?

Unless I'm missing something there is nothing specific to the generation of the OpenAPIV3 schema of [Quantity](https://github.com/kubernetes/apimachinery/blob/6d854d747c21b06aa210af1a3187ca8182860339/pkg/api/resource/quantity.go#L100) values in this PR.
Note that my IDE expects either an integer or a string (using the `m` suffix):

image

With the `m` suffix:

image

With quotes:

image

barkbay:
@naemono see also:

* This validation which prevents the use of float in custom resources: https://github.com/kubernetes-sigs/controller-tools/blob/40db49591af1bf9c336c7c3cefb417af618df402/pkg/crd/schema.go#L433-L438
* This documentation, "[Specify a CPU request and a CPU limit](https://kubernetes.io/docs/tasks/configure-pod-container/assign-cpu-resource/#specify-a-cpu-request-and-a-cpu-limit)", which clearly use some quotes around the quantity value: 

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: cpu-demo
  namespace: cpu-example
spec:
  containers:
  - name: cpu-demo-ctr
    image: vish/stress
    resources:
      limits:
        cpu: "1"
      requests:
        cpu: "0.5"
    args:
    - -cpus
    - "2"
```

I don't understand how the example you mentioned is not rejected by the API server tbh.

naemono:
> 
> I don't understand how the example you mentioned is not rejected by the API server tbh.

I think I'm simply assuming that the `Elasticsearch.spec.nodeSets[].podTemplate.spec.containers[].resources` follows the same validation rules as `Pod.spec.containers[].resources`.  Strange, when I apply the following, it works without failure, even without quotes.

```
apiVersion: v1
kind: Pod
metadata:
  name: cpu-demo
spec:
  containers:
  - name: cpu-demo-ctr
    image: vish/stress
    resources:
      limits:
        cpu: 1.0
      requests:
        cpu: 0.5
    args:
    - -cpus
    - "2"
```

It's probably not worth digging into to find why the validation is slightly different, but it certainly seems to be.

pkg/apis/common/v1alpha1/autoscaling_status.go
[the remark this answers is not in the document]
naemono: My apologies, I simply assumed a slice.  🤦‍♂️

pkg/controller/autoscaling/elasticsearch/validation/webhook.go
```diff
@@ -0,0 +1,140 @@
...
+	admissionv1 "k8s.io/api/admission/v1"
+	ctrl "sigs.k8s.io/controller-runtime"
+
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+
+	"github.com/elastic/cloud-on-k8s/v2/pkg/controller/common/license"
+	"github.com/elastic/cloud-on-k8s/v2/pkg/utils/k8s"
+	ulog "github.com/elastic/cloud-on-k8s/v2/pkg/utils/log"
+	"github.com/elastic/cloud-on-k8s/v2/pkg/utils/set"
```
pebrc, line 26: Import grouping seems off

pkg/controller/autoscaling/elasticsearch/validation/webhook_test.go
```diff
@@ -0,0 +1,606 @@
...
+									Name:              "data_policy",
+									AutoscalingPolicy: commonv1alpha1.AutoscalingPolicy{Roles: []string{"data"}},
+								},
+								AutoscalingResources: defaultResources,
+							},
+						},
+					},
+				},
+				checker: yesCheck,
+			},
```
pebrc, line 482: Missing the `wantValidationError` here?

pkg/controller/common/autoscaling/association.go
```diff
@@ -0,0 +1,67 @@
...
+
+import (
+	"context"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/tools/record"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	ulog "github.com/elastic/cloud-on-k8s/v2/pkg/utils/log"
+
```
pebrc, line 15: extra import group

pebrc (commented):
I tried to look through the source code and I it looks really good, for me it is ready to merge. I only found a few nits here and there and one incorrect usage of the log API. I also ran some tests and found some issues. However I am not sure if they are related to this PR or whether we have the same already with annotation based autoscaling. 

What I am seeing is that the Elasticsearch resource never fully reconciles (or at least only briefly). This is because the desired nodes API integration throws errors like the following: 

```
2022-09-19T20:45:16.341+0200	ERROR	manager.eck-operator	Reconciler error	{"service.version": "2.5.0-SNAPSHOT+8831df33", "controller": "elasticsearch-controller", "object": {"name":"es","namespace":"default"}, "namespace": "default", "name": "es", "reconcileID": "896949ed-99d7-4f71-a02f-c839d853a9aa", "error": "elasticsearch client failed for https://es-es-internal-http.default.svc:9200/_internal/desired_nodes/651bb9ea-360b-462f-87a1-fe5415d78ac5/8?error_trace=true: 400 Bad Request: {Status:400 Error:{CausedBy:{Reason: Type:} Reason:Desired nodes with history [651bb9ea-360b-462f-87a1-fe5415d78ac5] and version [8] already exists with a different definition Type:
```

It seems that the same generation of the ES resource is reconciled multiple times (expected I would say) but that the desired nodes differ within one generation. ~I have not dug deeper yet why exactly this is~ It seems likely that this is related to changing resources for the autoscaled node sets (the part that confuses me here is that if the resources change I would also expect the generation to change)
```diff
<       "storage": "3221225472b",
---
>       "storage": "2147483648b",
```

I think this is because we take the desired storage form the volume claim which is being resized one by one while the Elasticsearch resource is not changing. 


Because the nodes reconciliation has already happened at this point the autoscaling itself is not negatively affected and the cluster keeps scaling correctly. But the Elasticsearch status is stuck in applying changes and the log is full of the reconciler errors.

pkg/apis/common/v1alpha1/resources.go
```diff
@@ -0,0 +1,301 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License 2.0;
+// you may not use this file except in compliance with the Elastic License 2.0.
+
+package v1alpha1
```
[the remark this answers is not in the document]
barkbay: Yes, I have spent quite a bit of time dealing with import cycles 😞

pkg/apis/common/v1alpha1/autoscaling_status.go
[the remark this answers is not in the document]
barkbay: I have updated the condition message with your proposal [here](https://github.com/elastic/cloud-on-k8s/pull/5978/commits/36083b690d6937ba5862ce0c936b6d890bd79b88#diff-3a468729ad38582666b9dc853e089fad4a69834ccf0f1f258f35242e24a77b09L31-R31).

barkbay:
> This is because the desired nodes API integration throws errors like the following:
> [...]
> But the Elasticsearch status is stuck in applying changes and the log is full of the reconciler errors.

I hit a similar issue in https://github.com/elastic/cloud-on-k8s/issues/5979 I think we need to revisit the way we use the desired nodes API for the next release 😕

pkg/controller/autoscaling/elasticsearch/validation/webhook_test.go
barkbay: We actually support the case where the user is not relying on the default volume claim. I added a comment in the unit test [here](https://github.com/elastic/cloud-on-k8s/pull/5978/commits/285485c5e70d3eece4e1544a5f0b7b21a49c3a24#diff-f8e73750d745a6d002f0e88a8c1cd42455668bc2374110c3a77105e43ef50da9R483).

pkg/utils/stringsutil/strings.go
```diff
@@ -93,3 +93,19 @@ func SortStringSlice(s []string) {
...
 }
+
+func Truncate(s string, n int) string {
+	if n <= 0 {
+		return ""
+	}
+	truncated := ""
+	count := 0
+	for _, char := range s {
+		truncated += string(char)
```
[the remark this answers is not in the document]
barkbay: Let me know if this is what you had in mind: https://github.com/elastic/cloud-on-k8s/pull/5978/commits/fe773064f6120c059822659bab6e575e3a912267

Makefile
```diff
@@ -97,7 +97,7 @@ go-generate:
 generate-crds-v1: go-generate controller-gen
 	# Generate webhook manifest
 	# Webhook definitions exist in both pkg/apis and pkg/controller/elasticsearch/validation
-	$(CONTROLLER_GEN) webhook object:headerFile=./hack/boilerplate.go.txt paths=./pkg/apis/... paths=./pkg/controller/elasticsearch/validation/...
+	$(CONTROLLER_GEN) webhook object:headerFile=./hack/boilerplate.go.txt paths=./pkg/apis/... paths=./pkg/controller/elasticsearch/validation/... paths=./pkg/controller/autoscaling/elasticsearch/validation/...
```
thbkrkr, line 100: Nit: why not just gives `./pkg/...`? It's faster on my machine than setting 3 paths.
```suggestion
	$(CONTROLLER_GEN) webhook object:headerFile=./hack/boilerplate.go.txt paths=./pkg/...
```

pkg/apis/common/v1alpha1/resources.go
```diff
@@ -0,0 +1,301 @@
...
+		if nr.HasLimit(resourceName) {
+			mergedResources.Limits[resourceName] = nr.GetLimit(resourceName)
+		} else {
+			delete(mergedResources.Limits, resourceName)
+		}
+	}
+	return *mergedResources
+}
+
+// MaxMerge merge the specified resource into the NodeResources only if its quantity is greater
```
thbkrkr, line 123:
```suggestion
// MaxMerge merges the specified resource into the NodeResources only if its quantity is greater
```

thbkrkr (commented):
Very nice work!

Nit: almost all k8s resources displayed via kubectl have an "AGE" column, not this one. No big deal but my eyes aren't used to it.

I spotted a small limitation I think, we can't create an autoscaler for an ES that has 1 nodeSet without explicit `node.roles, which seems acceptable to me.
```
  nodeSets:
  - name: default
    count: 3
    config:
      node.store.allow_mmap: false
```

barkbay:
@naemono I think I addressed your comments, please let me know if I missed something 🙇

pkg/apis/common/v1alpha1/autoscaling_status.go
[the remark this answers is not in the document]
naemono: I think the change makes this more clear, thank you.

naemono (approved):
This looks great.  Nice work @barkbay

PR #5580, opened 2022-04-13
4561 replace ubi with scratch

naemono:
Closes #4561 

This will update the default Dockerfile for ECK operator to use a scratch image instead of UBI.  An additional UBI dockerfile has been added, which is leveraged in CI tooling to build/push an additional image to both:

1. docker.elastic.co
2. docker.io

So, a total of 4 images (not including different platforms) will be built when building a new release (non-SNAPSHOT)

1. docker.elastic.co/elastic/eck-operator:x.y.z
2. docker.io/elastic/eck-operator:x.y.z
3. docker.elastic.co/elastic/eck-operator:x.y.z-ubi
4. docker.io/elastic/eck-operator:x.y.z-ubi

This will be paired with an additional PR for the internal redhat tooling leveraged during release.

Reason for Draft:

Additional testing is being done now to ensure:
- [ ] These new scratch images work in non-Openshift environment.
- [ ] The UBI images work in Openshift environments.

PR #5983, opened 2022-08-31
Restore operator binary installation in slash

thbkrkr:
This commit reverts the location of the operator binary in the Docker image to how it was before #5580 (`/` instead of `/home/nonroot`). This change avoids having to modify internal tools that use `/`.


barkbay (approved):
LGTM

PR #5970, opened 2022-08-24
Update Stack Version to 8.4.0

david-kow:

PR #5986, opened 2022-08-31
Update Stack version to 8.4.2

thbkrkr:
- Update `update-stack-version.sh` to bump Stack version in ECK Stack Helm chart
- Bump Stack version from 8.2.3 to 8.4.2 in ECK Stack Helm chart using ^
- Bump Stack version from 8.4.0 to 8.4.2 everywhere
- Align e2e-monitor cluster version

To be backported in branch `2.4`.

### DIFF
--- a/config/e2e/monitoring.yaml
+++ b/config/e2e/monitoring.yaml
@@ -174,7 +174,7 @@ spec:
       meta:
         package:
           name: system
-          version: 8.4.1
+          version: 8.4.2
       data_stream:
         namespace: default
       processors:
--- a/config/e2e/monitoring.yaml
+++ b/config/e2e/monitoring.yaml
@@ -5,7 +5,7 @@ metadata:
   name: e2e-agent
   namespace: {{ .E2ENamespace }}
 spec:
-  version: 8.2.3 # needs to less or equal to e2e-monitor version
+  version: 8.4.0 # needs to less or equal to e2e-monitor version
   elasticsearchRefs:
     - secretName: eck-{{ .TestRun }}
   daemonSet:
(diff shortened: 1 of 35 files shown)


config/e2e/monitoring.yaml
```diff
@@ -174,7 +174,7 @@ spec:
       meta:
         package:
           name: system
-          version: 8.4.0
+          version: 8.4.1
```
thbkrkr, line 177: I think this one is wrong since a while. The latest version of the system package is [`1.19.4`](https://github.com/elastic/integrations/blob/main/packages/system/changelog.yml). But I'd rather handle that in a separate PR.

config/recipes/elastic-agent/system-integration.yaml
```diff
@@ -31,7 +31,7 @@ spec:
       meta:
         package:
           name: system
-          version: 8.4.0
+          version: 8.4.1
```
thbkrkr, line 34: Same.

naemono:
run/e2e-tests

thbkrkr:
> In previous PRs we also updated `pkg/controller/common/stackmon/config_test.go` [ref pr](https://github.com/elastic/cloud-on-k8s/pull/5970/files#diff-084dcf6507e01ca5add09011aeff83185a7d6122b2edce389840b5355065c573R69), should we do that here as well?

No, it's in a test. It was mistakenly updated without consequences by the script last time.

pebrc (approved):
LGTM, do you want to raise an issue for the problem with the system package in the monitoring config?

PR #5998, opened 2022-09-05
Fix chart "annotations" field identation

JParisR:
Adding the `nindent` prepends a newline to the beginning of the string to prevent the new annotation from being on the same line.

**Example using `indent` (before):**
indent-example

**Example using `nindent` (after):**
nindent-example


naemono:
Closing this as #6004 is a more complete fix of the issues found, and noted in this PR.  Thanks @JParisR for the contribution, as it's greatly appreciated.

PR #6004, opened 2022-09-08
Fix helm chart rendering issues associated with indentation

naemono:
supercedes #5998 

Issues were found with the rendering of annotations/labels, and other specific Elasticsearch values.  This PR fixes all of the issues, and adds tests around the previously untested custom annotations, and labels values, and adds additional tests for custom Elasticsearch values.

# Notable change

This stops version control of the `eck-stack/Chart.lock` file, as it requires an unreleased version of the child charts, so it will not be able to be rendered accurately during child chart version upgrades.


thbkrkr (approved):
LGTM

PR #5999, opened 2022-09-05
Add license template for Helm Charts

zhekazuev:
At the moment, if you follow the official instructions, when you try to install the official [ElasticSearch](https://artifacthub.io/packages/helm/elastic/eck-elasticsearch) and [Kibana](https://artifacthub.io/packages/helm/elastic/eck-kibana) or [ECK Stack](https://artifacthub.io/packages/helm/elastic/eck-stack) Charts after installing the operator and CRDs, you will receive an error of mismatch between basic and enterprise licenses in Secret and Chart respectively.

Changes add:
- license templating

```yaml
# values.yaml
# Elastic license. basic or enterprise
#
license: basic
```

- default value for license - basic

```yaml
annotations:
  # eck.k8s.elastic.co/license: enterprise
  eck.k8s.elastic.co/license: {{ .Values.license | default "basic" | quote }}
 ```

PR #6019, opened 2022-09-19
Allow license type parametrization

bvis:
Signed-off-by: Basilio Vera <basilio.vera@softonic.com>

The helm charts that are provided by eck-stack for generating Elasticsearch clusters and its Kibana have "enterprise" license value hardcoded in the template.

With this you can use a parameter in your chart to use "basic" license instead. The default values remain the same maintaining backwards compatibility.


thbkrkr:
Hello, sorry but I'm closing for the same reason than https://github.com/elastic/cloud-on-k8s/pull/5999#issuecomment-1237818210.

PR #6028, opened 2022-09-21
Update k8s supported versions

### LINKED ISSUE #6024: Update K8S supported versions
We [state](https://github.com/elastic/cloud-on-k8s/blob/8662b116f8712176d1178e18c44d4f14eadec60a/README.md?plain=1#L20) that the operator supports Kubernetes 1.20-1.24, however:
* `1.21` has reached EOL on `2022-06-28`
* same for `1.20` 6 months ago
* `1.25.1` is already available on Kind ([`1.25.2` should be around the corner though](https://groups.google.com/a/kubernetes.io/g/dev/c/tA6LNOQTR4Q/m/zL73maPTAQAJ?utm_medium=email&utm_source=footer))

Let's target a support for K8S `1.21` - `1.25` for the next release?

thbkrkr:
- Updates kind version from 0.14.0 to 0.15.0
- Adds k8s 1.25.1 to the list of tested k8s versions, removes k8s 1.20
- Updates the supported k8s versions in the doc and the main README

Resolves #6024.


README.md
```diff
@@ -17,7 +17,7 @@ Current features:
 
 Supported versions:
 
-*  Kubernetes 1.20-1.24
+*  Kubernetes 1.21-1.25
```
barkbay, line 20: Maybe also update [hack/operatorhub/templates/csv.tpl](https://github.com/elastic/cloud-on-k8s/blob/d9addf1e1a540a0d1dfc70f3ff550593e3ba09aa/hack/operatorhub/templates/csv.tpl#L268-L270)?

barkbay (commented):
Should we also update the Helm Charts ([this one for example](https://github.com/elastic/cloud-on-k8s/blob/d9addf1e1a540a0d1dfc70f3ff550593e3ba09aa/deploy/eck-stack/Chart.yaml#L8))?

hack/manifest-gen/internal/manifest.go
```diff
@@ -72,6 +73,8 @@ func Generate(opts *GenerateFlags) error {
 	client.ReleaseName = "elastic-operator"
 	client.Namespace = opts.OperatorNamespace
 	client.PostRenderer = helmLabelRemover{}
+	fakeKubeVersion, _ := chartutil.ParseKubeVersion("v9.99.99")
+	client.KubeVersion = fakeKubeVersion
```
thbkrkr, line 77: Otherwise the version used is `1.20` coming from the default capacities which makes the generation fail if we want to start to require `>=1.21.0`.

barkbay: [IIUC](https://github.com/helm/helm/blob/b5e39c15ea4f2ac14b1e5b6635d46410ffb79fec/pkg/action/install.go#L216-L222) since Helm is not connected to a real K8S cluster (`ClientOnly == true`) then a K8S API version must be arbitrarily selected. This version is hardcoded in Helm, and as of today its value is `1.20`, which prevents Helm to render the manifests from the Charts if we request a higher value as it is the case in this PR (`1.21`).
I would maybe add a comment to explain the problem and why we are setting this value.

hack/manifest-gen/internal/manifest.go
```diff
@@ -72,6 +73,8 @@ func Generate(opts *GenerateFlags) error {
 	client.ReleaseName = "elastic-operator"
 	client.Namespace = opts.OperatorNamespace
 	client.PostRenderer = helmLabelRemover{}
+	fakeKubeVersion, _ := chartutil.ParseKubeVersion("v9.99.99")
```
barkbay, line 76:
```suggestion
	fakeKubeVersion, err := chartutil.ParseKubeVersion("v9.99.99")
	if err != nil {
		return fmt.Errorf("invalid fake kube version '%s': %s", fakeKubeVersion, err)
	}
```

barkbay (approved):
LGTM

PR #6037, opened 2022-09-23
Increment desired nodes version on each call

### LINKED ISSUE #5979: Upgrading from ECK 2.4.0 to latest version may fail
Got this error for one of my Elasticsearch cluster while upgrading from `2.4.0` to main:

```
400 Bad Request: {Status:400 Error:{CausedBy:{Reason: Type:}
Reason:
Desired nodes with history [d7bf9e8e-47a0-40ad-8156-400ae519eb6c] and version [2] already exists with a different definition
```

I _think_ (not 💯 sure yet) this is because of https://github.com/elastic/cloud-on-k8s/pull/5950: the Elasticsearch configuration has changed, but the `metadata.generation` of the Elasticsearch resource, used as the `version` field for the desired nodes API,  is still the same.

I wonder if we should have an "upgrade" e2e pipeline, something like our `upgrade-test-harness`, that should be run automatically when submitting a PR in order to detect this kind of issue.

pebrc:
Fixes  #5979

Update the version of the desired nodes history on each call by fetching the latest version before updating. 

The optimisation outlined in https://github.com/elastic/cloud-on-k8s/issues/5979#issuecomment-1255966670 is in a separate commit and can be backed out if we think it is too much.

### DIFF
--- a/pkg/controller/elasticsearch/driver/desired_nodes.go
+++ b/pkg/controller/elasticsearch/driver/desired_nodes.go
@@ -69,7 +69,14 @@ func (d *defaultDriver) updateDesiredNodes(
 		results.WithReconciliationState(defaultRequeue.WithReason("Storage capacity is not available in all PVC statuses, requeue to refine the capacity reported in the desired nodes API"))
 	}
 	if esReachable {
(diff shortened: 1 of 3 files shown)


barkbay (approved):
LGTM 👍

PR #6048, opened 2022-09-28
Update the base Docker image of OpenShift installer

thbkrkr:
This commit changes the base Docker image used for the OpenShift installer from `scratch` to `gcr.io/distroless/base-debian11:nonroot` because since the latest version of the installer (4.11.x), the Terraform binary that is downloaded by the installer is not statically linked.


thbkrkr:
Duplicates of #6049.

PR #6054, opened 2022-09-28
Use snapshot EPR for snapshot tests

### LINKED ISSUE #6053: TestNameValidation fails for 8.5.0-SNAPSHOT
```
Document produced by APM Server v8.5.0, which is newer than the installed APM integration (v8.4.0). The APM integration must be upgraded
```

The APM integration installed is `latest`. I assume there is no APM integration available for 8.5.0 yet in the package registry. I am not sure if we have a way to fix this until the new version is published.

pebrc:
Fixes #6053

The snapshot version of the EPR has newer versions of the integration packages that allow the test to pass. The risk is that the snapshot versions of integration packages introduce new instability into our tests.


barkbay (approved):
LGTM

PR #6063, opened 2022-10-03
Add latest APM fleet package in Kibana examples when using standalone APM server.

### LINKED ISSUE #5059: Update APM Server examples/samples for 8.0.0
Standalone APM Server examples should be updated with a change to Kibana config as below.

```yaml
xpack.fleet.packages:
- name: apm
  version: latest 
```

This is due to the fact APM Server will no longer create apm indices/templates and instead, templates will be installed during APM integration installation. The configuration above makes sure that Kibana will install it automatically during startup.

naemono:
resolves #5059 

This change updates all of our standalone APM Server examples to include the latest APM fleet package in the Kibana resource.

### DIFF
--- a/config/recipes/traefik/02-elastic-stack.yaml
+++ b/config/recipes/traefik/02-elastic-stack.yaml
@@ -41,12 +41,12 @@ metadata:
   labels:
     app: hulk
 spec:
+  version: 8.4.2
+  count: 1
   config:
     xpack.fleet.packages:
     - name: apm
       version: latest
-  version: 8.4.2
-  count: 1
   elasticsearchRef:
     name: hulk
 ---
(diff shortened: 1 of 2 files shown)


config/recipes/traefik/02-elastic-stack.yaml
```diff
@@ -41,6 +41,10 @@ metadata:
   labels:
     app: hulk
 spec:
+  config:
+    xpack.fleet.packages:
+    - name: apm
+      version: latest
   version: 8.4.2
   count: 1
```
thbkrkr, line 49: How about keeping th `version`, `count`, `config` fields in the same order:
```suggestion
  version: 8.4.2
  count: 1
  config:
    xpack.fleet.packages:
    - name: apm
      version: latest
```

thbkrkr (commented):
Should we also add a note in the APM documentation as this may still be difficult to find out?

thbkrkr (approved):
Since the standalone APM server is deprecated and this config is not mentioned in the APM documentation, I think we can leave it at that by just updating the samples.

PR #6071, opened 2022-10-05
Enable FIPS compliancy in Makefile

naemono:
This potential change will enable FIPS compliancy for the operator build by making the changes:

1. Add an `ENABLE_FIPS` environment variable which will add a build flag `goexperiment.boringcrypto` to the existing `GO_TAGS` variable which we use during `go build` and `docker build` operations.
2. Append `-fips` to all container build names that we push to existing registries.
3. Adjusts the release pipeline in Jenkins to run another docker build operation with `ENABLE_FIPS=true` environment variable set. 

~~This change does *not*~~

~~1. Add any Jenkins/Buildkite changes that would allow FIPS these changes to be triggered in our current/future CI system.~~

### DIFF
--- a/.ci/pipelines/build.Jenkinsfile
+++ b/.ci/pipelines/build.Jenkinsfile
@@ -34,11 +34,21 @@ pipeline {
                         sh 'make -C .ci TARGET=ci ci'
                     }
                 }
-                stage('Build and push Docker image') {
-                    steps {
-                        sh '.ci/setenvconfig build'
-                        sh 'make -C .ci license.key TARGET=ci-release ci'
-                        sh 'make -C .ci license.key TARGET=build-operator-multiarch-image ci ENABLE_FIPS=true'
+                stage('build') {
+                    failFast true
+                    parallel {
+                        stage("build and push operator image") {
+                            steps {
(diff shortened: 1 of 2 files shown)


.ci/pipelines/build.Jenkinsfile
```diff
@@ -38,6 +38,7 @@ pipeline {
                     steps {
                         sh '.ci/setenvconfig build'
                         sh 'make -C .ci license.key TARGET=ci-release ci'
+                        sh 'make -C .ci license.key TARGET=build-operator-multiarch-image ci ENABLE_FIPS=true'
```
naemono, line 41: Would it make sense to do a whole other step here, with a different target in `setenvconfig`?  It seems overkill when looking at this, but `setenvconfig` is used to set all env variables in all other steps, but since the `build` case block is split between `nightly` and `release` already, it seemed adding a 3rd would be ugly.

naemono: I think the parallel build makes perfect sense.  I've updated this PR with this new parallel build phase.

thbkrkr (approved):
LGTM

PR #6079, opened 2022-10-10
Run tests in AKS with k8s 1.24.6

pebrc:
The version we used up to now is no longer available. This bumps versions to the latest available.


barkbay (approved):
LGTM

Issue #6076, opened 2022-10-10
Impact of hashes comparisons in the Elasticsearch reconciliation performance


barkbay:
For each reconciliation the operator systematically  [performs some `bcrypt` operations](https://github.com/elastic/cloud-on-k8s/blob/56de716785d02c7b34467ecb3fc3fa2304e7449d/pkg/controller/elasticsearch/user/predefined.go#L166-L172) to ensure Elasticsearch users passwords hashes are up-to-date in the file realm. These cryptographic operations can have a significant impact in cpu constrained environments:

<img width="965" alt="image" src="https://user-images.githubusercontent.com/976373/194806914-60bdcacb-ed9b-415d-928f-9263c2b24fcc.png">

The use of `bcrypt.CompareHashAndPassword` is linearly proportional to the number of users managed in the file realm. I think there are 3 of them by default ([not counting the ones generated for some associations](https://github.com/elastic/cloud-on-k8s/blob/0d0aa0ca4758f2ba70ba7b236ba42f6ada97f302/pkg/controller/association/user.go#L150-L163)). Using a LRU cache like [github.com/hashicorp/golang-lru](https://github.com/hashicorp/golang-lru) can significantly improve the situation:

<img width="520" alt="image" src="https://user-images.githubusercontent.com/976373/194805686-60a6cbed-10a5-47d7-954d-0dcff94d89be.png">

_(in the example above I added the LRU cache and a couple of additional spans)_

The question is to find a good cache size, it can be the `max-concurrent-reconciles` value (with a multiplier?) or inferred from the available memory (a bit involved maybe).

PR #6080, opened 2022-10-10
Add password hash (cached) generator

barkbay:
Attempt to mitigate and close https://github.com/elastic/cloud-on-k8s/issues/6076

The cache size is arbitrarily set to 5 `x` the number of workers. It can be changed (and disabled) via a configuration flag. We could dynamically resize the cache according to the number of managed resources, but I thought it was a bit involved for a first implementation.

From a security point of view:
* It means that passwords and hashes are stored in memory for a longer time, I don't think it is a concern as they are already both in memory, and reading the process memory would require the server to be compromised.
* Password comparison is no longer "time consistent", but I can't a see how an attacker could take advantage of this behaviour in the context of a reconciliation loop.

### DIFF
--- a/NOTICE.txt
+++ b/NOTICE.txt
@@ -1287,6 +1287,378 @@ Exhibit B - “Incompatible With Secondary Licenses” Notice
 
 
 --------------------------------------------------------------------------------
+Module  : github.com/hashicorp/golang-lru
+Version : v0.5.4
+Time    : 2020-01-16T18:29:26Z
+Licence : MPL-2.0
+
+Contents of probable licence file $GOMODCACHE/github.com/hashicorp/golang-lru@v0.5.4/LICENSE:
+
+Mozilla Public License, version 2.0
+
+1. Definitions
+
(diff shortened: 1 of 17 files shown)


barkbay:
Jenkins test this please

pebrc (approved):
LGTM I was conflicted whether such a low-level implementation detail should be exposed through a dedicated flag it feels like something that should go into some form of internal performance tuning settings.  But I guess we don't have such a concept so I am good with the choice made here.

pebrc:
I just realised an entry for the new flag in the documentation page for the operator flags is missing.

barkbay:
> I just realised an entry for the new flag in the documentation page for the operator flags is missing.

It makes me realise that I forgot to add the parameter in the Helm chart generated configmap: https://github.com/elastic/cloud-on-k8s/pull/6080/commits/6673a77ed7ff31bc5aaaf39bd4cb0faef30db517 😄

pebrc (approved):

Issue #6078, opened 2022-10-10
Synchronous observation side effects


barkbay:
https://github.com/elastic/cloud-on-k8s/pull/5783 added a "synchronous" _(in the Elasticsearch controller itself)_ observation. 

This introduced the following bug / side effects:

## 1. We should not try to do a first observation if Elasticsearch service is not ready

If [`services.IsServiceReady`](https://github.com/elastic/cloud-on-k8s/blob/85aa75b7a7765330475c0b92a0851939ac3d9578/pkg/controller/elasticsearch/services/services.go#L144) is `false` then the observation is likely to timeout or failed, while unnecessarily blocking a worker. We should therefore skip the observation.

## 2. Slow observations can prevent other workers to make progress

If there is at least 1 Elasticsearch cluster that is slow to be observed then the [observerLock](https://github.com/elastic/cloud-on-k8s/blob/85aa75b7a7765330475c0b92a0851939ac3d9578/pkg/controller/elasticsearch/observer/manager.go#L94-L95) is likely to become a hot spot:

<img width="1339" alt="image" src="https://user-images.githubusercontent.com/976373/194813490-f8d78b40-7086-47f8-a65b-d3b8e12821f7.png">

In the screenshot above the worker has been blocked `2` x `10s` (in `get_observer` and `create_or_replace_observer`) because of the use of `m.observerLock.Lock()`, which was locked by another worker. It has then itself prevented other workers to make progress for `10s` while trying to observe the Elasticsearch cluster being reconciled. If we want to keep this feature then a workaround could be to use a custom, low(?), timeout for this first observation.

PR #6084, opened 2022-10-12
Observer improvements

barkbay:
This PR implements a few improvements on the observer:

* The first synchronous observation implemented in https://github.com/elastic/cloud-on-k8s/pull/5783 is no longer invoked if the Elasticsearch `Service` is not "ready"
* `observer.Start()` is invoked outside of any lock-protected section
* An APM span added around `observer.observe(...)`

I eventually decided to not introduce a custom, low, timeout as suggested in https://github.com/elastic/cloud-on-k8s/issues/6078 for the following reasons:
* Moving the first synchronous observation out of the lock-protected section should solve the main bottleneck.
* Adding an additional timeout value means adding a new flag or/and implement an implicit default value.
* If the call to `_cluster/health` is slow then chances are that it'll also be the case for subsequent API calls, not sure why we would treat this one as a special case.

Fixes https://github.com/elastic/cloud-on-k8s/issues/6078

### DIFF
--- a/pkg/controller/elasticsearch/observer/manager.go
+++ b/pkg/controller/elasticsearch/observer/manager.go
@@ -74,12 +74,12 @@ func (m *Manager) Observe(ctx context.Context, cluster esv1.Elasticsearch, esCli
 
 	switch {
 	case !exists:
-		return m.createOrReplaceObserver(nsName, settings, esClient, doFirstObservation)
+		return m.createOrReplaceObserver(ctx, nsName, settings, esClient, doFirstObservation)
 	case exists && (!observer.esClient.Equal(esClient) || observer.settings != settings):
-		return m.createOrReplaceObserver(nsName, settings, esClient, doFirstObservation)
+		return m.createOrReplaceObserver(ctx, nsName, settings, esClient, doFirstObservation)
 	case exists && settings.ObservationInterval <= 0:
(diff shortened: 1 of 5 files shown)


pkg/controller/elasticsearch/observer/manager.go
```diff
@@ -61,24 +67,36 @@ func (m *Manager) getObserver(key types.NamespacedName) (*Observer, bool) {
...
-		return m.createOrReplaceObserver(nsName, settings, esClient)
+		// This Elasticsearch resource is already being observed asynchronously, no need to do a first observation.
+		observer = m.createOrReplaceObserver(ctx, nsName, settings, esClient)
 	case exists && settings.ObservationInterval <= 0:
 		// in case asynchronous observation has been disabled ensure at least one observation at reconciliation time.
-		return m.getAndObserveSynchronously(nsName)
+		return m.getAndObserveSynchronously(ctx, nsName)
 	default:
+		// No change, close the provided Client and return the existing observer.
 		esClient.Close()
```
thbkrkr, line 89: I had forgotten that. It's not great. I opened:
- https://github.com/elastic/cloud-on-k8s/issues/6090

thbkrkr (approved):
I was about to suggest moving the blocking call outside the lock and you beat me :-)
I checked that it doesn't break the fix made for #5812.

LGTM!

PR #6086, opened 2022-10-12
Add container-suffix operator flag

### LINKED ISSUE #6064: Consider adding a --container-suffix startup parameter to the operator
We are currently setting `containerSuffix` if the `-ubi-only` flag is used which sets it to `-ubi`. However there is an additional use case for users who want to benefit from the image defaulting but need a different suffix than the default empty string (e.g. because they use a mirror image registry or need a custom suffix to disambiguate images), an additional parameter might be useful. 

It would be illegal to combine `--ubi-only` with `--container-suffix`.

pebrc:
Fixes #6064 

`--container-suffix` allows users to set a default suffix for container images that will be applied whenever the operator defaults to setting the image used. That is the case whenever the user does not explicitly specify the `image` attribute on a Elastic Stack resource manifest. 

It is illegal to combine both `--ubi-only` and `--container-suffix` because `--ubi-only` already implies a container suffix of `-ubi`

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -179,6 +179,11 @@ func Command() *cobra.Command {
 		"Container registry to use when downloading Elastic Stack container images",
 	)
 	cmd.Flags().String(
+		operator.ContainerSuffixFlag,
+		"",
+		fmt.Sprintf("Suffix to be appended to container images by default. Cannot be combined with %s", operator.UBIOnlyFlag),
+	)
+	cmd.Flags().String(
 		operator.DebugHTTPListenFlag,
 		"localhost:6060",
 		"Listen address for debug HTTP server (only available in development mode)",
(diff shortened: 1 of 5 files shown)


cmd/manager/main.go
```diff
@@ -178,6 +178,11 @@ func Command() *cobra.Command {
 		container.DefaultContainerRegistry,
 		"Container registry to use when downloading Elastic Stack container images",
 	)
+	cmd.Flags().String(
+		operator.ContainerSuffixFlag,
+		"",
+		fmt.Sprintf("Suffix to be appended to container images by default. Cannot be combined with %s", operator.UBIOnlyFlag),
```
barkbay, line 184: nit: maybe do the same for the UBI flag?

```golang
	cmd.Flags().Bool(
		operator.UBIOnlyFlag,
		false,
		fmt.Sprintf("Use only UBI container images to deploy Elastic Stack applications. UBI images are only available from 7.10.0 onward. Cannot be combined with %s", operator.ContainerSuffixFlag),
	)
```

And do a similar change in the doc?

barkbay (approved):
LGTM!

If we want this flag to be available with Helm I think we should also update the [ConfigMap template](https://github.com/elastic/cloud-on-k8s/blob/028636a11d26a9c90a67a1af6947d7d6cc9eebfd/deploy/eck-operator/templates/configmap.yaml#L10).

barkbay (approved):
LGTM

PR #6118, opened 2022-10-25
Draft: Extend existing reattach-pv tool

naemono:
The proposed change would extend the existing reattach-pv tool to support creating a new cluster (clusterB) using PVs from a previously deleted cluster (clusterA), while maintaining the new name (clusterB).

Why is this needed?
==

If a customer accidentally deletes clusterA, then immediately re-creates clusterA, it will be created with the same name, and new PersistentVolumes, becoming a new cluster. If that cluster begins ingesting data of some sort, then it will contain indexes with valid data that would need to be retained.

This change would allow the Released PVs of clusterA to be used to build a new clusterB, which contains the old data of clusterA, which would allow clusterA then to ingest the data of clusterB, and have the whole of the customer's data.

What is missing?
==
- [ ] Readme is not yet updated.

Testing Done
==
1. Create, delete, delete PVCs from prev cluster, then use tool to re-create existing cluster with existing PVs, verifying indexes with previous data exist. (previous functionality)
2. Create `kubectl apply -f es.yaml`, delete `kubectl delete es`, delete PVCs from prev cluster, create again with same name `kubectl apply -f es.yaml`, then create 2nd cluster using existing PVs, and verify indexes with previous data exist.

### DIFF
--- a/support/reattach-pv/main.go
+++ b/support/reattach-pv/main.go
@@ -173,7 +173,10 @@ func expectedVolumeClaims(es esv1.Elasticsearch, oldEsName string) map[types.Nam
 					claim = claimTemplate
 				}
 			}
-			claim.Name = claimName(es, oldEsName, nodeSet.Name, i)
+			claim.Name = fmt.Sprintf(
+				"%s-%s",
+				volume.ElasticsearchDataVolumeName,
+				sset.PodName(esv1.StatefulSet(es.Name, nodeSet.Name), i))
 			claim.Namespace = es.Namespace
 			if claim.Namespace == "" {
 				claim.Namespace = "default"
@@ -189,19 +192,6 @@ func expectedVolumeClaims(es esv1.Elasticsearch, oldEsName string) map[types.Nam
 		}
 	}
 	return claims
-}
-
-// claimName generates the name of the expected volume claim given an Elasticsearch cluster, a nodeSet name
-// an ordinal, and potentially an old Elasticsearch cluster name.
-func claimName(es esv1.Elasticsearch, oldEsName string, nodeSetName string, ordinal int32) string {
-	esName := es.Name
-	if oldEsName != "" {
-		esName = oldEsName
-	}
-	return fmt.Sprintf(
-		"%s-%s",
-		volume.ElasticsearchDataVolumeName,
-		sset.PodName(esv1.StatefulSet(esName, nodeSetName), ordinal))
 }
 
 // findReleasedPVs returns the list of Released PersistentVolumes.
@@ -233,22 +223,34 @@ func matchPVsWithClaim(pvs []v1.PersistentVolume, claims map[types.NamespacedNam
 		if pv.Spec.ClaimRef == nil {
 			continue
 		}
-		// The following regex defines the expected persistent volume claim name format
-		// The regex patters for both EsName, and NodeSetName groups are directly from apimachinery validation:
-		// https://github.com/kubernetes/apimachinery/blob/master/pkg/util/validation/validation.go#L178
-		r := regexp.MustCompile(`^elasticsearch-data-(?P<EsName>[a-z0-9]([-a-z0-9]*[a-z0-9])+)-es-(?P<NodeSetName>[a-z0-9]([-a-z0-9]*[a-z0-9])+)-(?P<Ordinal>[0-9]+)$`)
-		regexMatches := r.FindStringSubmatch(pv.Spec.ClaimRef.Name)
-		// 4 matches are expected here, as the first match is the full string, and the next are EsName, NodeSetName, and Ordinal.
-		if len(regexMatches) != 4 {
-			continue
+		expectedClaimName := pv.Spec.ClaimRef.Name
+		// if you're building a newly named cluster, from a previous cluster's PVs, we'll
+		// need to extract the nodeSetName, and ordinal from the PV's claimref.Name,
+		// and replace the expected claim name to be the newly generated cluster's name.
+		if oldEsName != "" {
+			// The following regex defines the expected persistent volume claim name format
+			// The regex patters for both EsName, and NodeSetName groups are directly from apimachinery validation:
+			// https://github.com/kubernetes/apimachinery/blob/master/pkg/util/validation/validation.go#L178
+			r := regexp.MustCompile(`^elasticsearch-data-([a-z0-9]([-a-z0-9]*[a-z0-9])?)-es-([a-z0-9]([-a-z0-9]*[a-z0-9])?)-([0-9])+$`)
+			regexMatches := r.FindStringSubmatch(pv.Spec.ClaimRef.Name)
+			// 6 matches are expected here, as the first match is the full string, and the next are EsName+internal match, NodeSetName+internal match, and Ordinal.
+			if len(regexMatches) != 6 {
+				fmt.Printf("expected 6 regex, got %d: %v", len(regexMatches), regexMatches)
+				continue
+			}
+			nodeSetName := regexMatches[3]
+			strOrdinal := regexMatches[5]
+			ordinal, err := strconv.Atoi(strOrdinal)
+			if err != nil {
+				continue
+			}
+			expectedClaimName = fmt.Sprintf(
+				"%s-%s",
+				volume.ElasticsearchDataVolumeName,
+				sset.PodName(esv1.StatefulSet(es.Name, nodeSetName), int32(ordinal)))
 		}
-		nodeSetName := regexMatches[2]
-		strOrdinal := regexMatches[3]
-		ordinal, err := strconv.Atoi(strOrdinal)
-		if err != nil {
-			continue
-		}
-		claim, expected := claims[types.NamespacedName{Namespace: pv.Spec.ClaimRef.Namespace, Name: claimName(es, oldEsName, nodeSetName, int32(ordinal))}]
+
+		claim, expected := claims[types.NamespacedName{Namespace: pv.Spec.ClaimRef.Namespace, Name: expectedClaimName}]
 		if !expected {
 			continue
 		}
@@ -278,6 +280,7 @@ func createAndBindClaims(c k8s.Client, volumeClaims []MatchingVolumeClaim, dryRu
 		// match.claim now stores the created claim metadata
 		// patch the volume spec to match the new claim
 		match.volume.Spec.ClaimRef.UID = match.claim.UID
+		match.volume.Spec.ClaimRef.Name = match.claim.Name
 		match.volume.Spec.ClaimRef.ResourceVersion = match.claim.ResourceVersion
 		if !dryRun {
 			if err := c.Update(context.Background(), &match.volume); err != nil {
--- a/support/reattach-pv/main.go
+++ b/support/reattach-pv/main.go
@@ -8,6 +8,8 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"regexp"
+	"strconv"
 	"strings"
 
 	"github.com/spf13/cobra"
@@ -28,6 +30,7 @@ import (
 
 const (
 	esManifestFlag = "elasticsearch-manifest"
+	oldEsNameFlag  = "old-elasticsearch-name"
 	dryRunFlag     = "dry-run"
 )
 
@@ -39,6 +42,8 @@ var Cmd = &cobra.Command{
 		if dryRun {
 			fmt.Println("Running in dry run mode")
 		}
+
+		oldEsName := viper.GetString(oldEsNameFlag)
 
 		err := esv1.AddToScheme(scheme.Scheme)
 		exitOnErr(err)
@@ -52,14 +57,14 @@ var Cmd = &cobra.Command{
 		err = checkElasticsearchNotFound(c, es)
 		exitOnErr(err)
 
-		expectedClaims := expectedVolumeClaims(es)
+		expectedClaims := expectedVolumeClaims(es, oldEsName)
 		err = checkClaimsNotFound(c, expectedClaims)
 		exitOnErr(err)
 
 		releasedPVs, err := findReleasedPVs(c)
 		exitOnErr(err)
 
-		matches, err := matchPVsWithClaim(releasedPVs, expectedClaims)
+		matches, err := matchPVsWithClaim(releasedPVs, expectedClaims, es, oldEsName)
 		exitOnErr(err)
 
 		err = createAndBindClaims(c, matches, dryRun)
@@ -75,6 +80,11 @@ func init() {
 		esManifestFlag,
 		"",
 		"path pointing to the Elasticsearch yaml manifest",
+	)
+	Cmd.Flags().String(
+		oldEsNameFlag,
+		"",
+		"name of previous Elasticsearch cluster (to use existing volumes)",
 	)
 	Cmd.Flags().Bool(
 		dryRunFlag,
@@ -152,8 +162,8 @@ func checkClaimsNotFound(c k8s.Client, claims map[types.NamespacedName]v1.Persis
 }
 
 // expectedVolumeClaims builds the list of PersistentVolumeClaim that we expect to exist for the given
-// Elasticsearch cluster.
-func expectedVolumeClaims(es esv1.Elasticsearch) map[types.NamespacedName]v1.PersistentVolumeClaim {
+// Elasticsearch cluster, or oldEsName, if given.
+func expectedVolumeClaims(es esv1.Elasticsearch, oldEsName string) map[types.NamespacedName]v1.PersistentVolumeClaim {
 	claims := make(map[types.NamespacedName]v1.PersistentVolumeClaim, es.Spec.NodeCount())
 	for _, nodeSet := range es.Spec.NodeSets {
 		for i := int32(0); i < nodeSet.Count; i++ {
@@ -163,10 +173,7 @@ func expectedVolumeClaims(es esv1.Elasticsearch) map[types.NamespacedName]v1.Per
 					claim = claimTemplate
 				}
 			}
-			claim.Name = fmt.Sprintf(
-				"%s-%s",
-				volume.ElasticsearchDataVolumeName,
-				sset.PodName(esv1.StatefulSet(es.Name, nodeSet.Name), i))
+			claim.Name = claimName(es, oldEsName, nodeSet.Name, i)
 			claim.Namespace = es.Namespace
 			if claim.Namespace == "" {
 				claim.Namespace = "default"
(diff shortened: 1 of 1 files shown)


barkbay (commented):
> This change would allow the Released PVs of clusterA to be used to build a new clusterB

I think an alternative would be to recreate `clusterA`, with the same name, but in a different namespace. The "only" manual step is to patch the `PersistentVolumes` to refer to the new namespace, with something along those lines:

1. get the released PVs

```
 k get pv | grep Released
NAME   CAPACITY  ACCESS MODES   RECLAIM POLICY   STATUS      CLAIM                                                      STORAGECLASS         AGE
pv1   2Gi        RWO            Retain           Released    ns1/elasticsearch-data-elasticsearch-sample-es-default-0   retain-sc            17m
pv2   2Gi        RWO            Retain           Released    ns1/elasticsearch-data-elasticsearch-sample-es-default-2   retain-sc            17m
pv3   2Gi        RWO            Retain           Released    ns1/elasticsearch-data-elasticsearch-sample-es-default-1   retain-sc            17m
```

2. patch the PVs with the new namespace

```
 kubectl patch pv pv1 pv2 pv3  -p '{"spec":{"claimRef": {"namespace": "ns2"} }}'
```

3. Ensure that the new namespace `ns2` does exist and is set in the `metadata.namespace` of the Elasticsearch manifest we want to recreate.

4. Run the `reattach-pv` tool as usual

It involves some manual steps, maybe a bit error prone...

support/reattach-pv/main.go
```diff
@@ -152,8 +162,8 @@ func checkClaimsNotFound(c k8s.Client, claims map[types.NamespacedName]v1.Persis
 }
 
 // expectedVolumeClaims builds the list of PersistentVolumeClaim that we expect to exist for the given
-// Elasticsearch cluster.
-func expectedVolumeClaims(es esv1.Elasticsearch) map[types.NamespacedName]v1.PersistentVolumeClaim {
+// Elasticsearch cluster, or oldEsName, if given.
+func expectedVolumeClaims(es esv1.Elasticsearch, oldEsName string) map[types.NamespacedName]v1.PersistentVolumeClaim {
```
barkbay, line 166: `oldEsName` is not used in this function

support/reattach-pv/main.go
```diff
@@ -207,13 +217,40 @@ type MatchingVolumeClaim struct {
...
 	for _, pv := range pvs {
 		if pv.Spec.ClaimRef == nil {
 			continue
 		}
-		claim, expected := claims[types.NamespacedName{Namespace: pv.Spec.ClaimRef.Namespace, Name: pv.Spec.ClaimRef.Name}]
+		expectedClaimName := pv.Spec.ClaimRef.Name
+		// if you're building a newly named cluster, from a previous cluster's PVs, we'll
+		// need to extract the nodeSetName, and ordinal from the PV's claimref.Name,
+		// and replace the expected claim name to be the newly generated cluster's name.
+		if oldEsName != "" {
```
barkbay, line 230: IIUC the value of `oldEsName` is not used, it could be a boolean.

support/reattach-pv/main.go
```diff
@@ -207,13 +217,40 @@ type MatchingVolumeClaim struct {
...
+			// The following regex defines the expected persistent volume claim name format
+			// The regex patters for both EsName, and NodeSetName groups are directly from apimachinery validation:
+			// https://github.com/kubernetes/apimachinery/blob/master/pkg/util/validation/validation.go#L178
+			r := regexp.MustCompile(`^elasticsearch-data-([a-z0-9]([-a-z0-9]*[a-z0-9])?)-es-([a-z0-9]([-a-z0-9]*[a-z0-9])?)-([0-9])+$`)
```
barkbay, line 234: `r` is loop invariant, could be extracted as a var:

```golang
var (
	r = regexp.MustCompile(`^elasticsearch-data-([a-z0-9]([-a-z0-9]*[a-z0-9])?)-es-([a-z0-9]([-a-z0-9]*[a-z0-9])?)-([0-9])+$`)
)
```

support/reattach-pv/main.go
```diff
@@ -207,13 +217,40 @@ type MatchingVolumeClaim struct {
...
 		}
-		claim, expected := claims[types.NamespacedName{Namespace: pv.Spec.ClaimRef.Namespace, Name: pv.Spec.ClaimRef.Name}]
+		expectedClaimName := pv.Spec.ClaimRef.Name
+		// if you're building a newly named cluster, from a previous cluster's PVs, we'll
+		// need to extract the nodeSetName, and ordinal from the PV's claimref.Name,
+		// and replace the expected claim name to be the newly generated cluster's name.
+		if oldEsName != "" {
+			// The following regex defines the expected persistent volume claim name format
+			// The regex patters for both EsName, and NodeSetName groups are directly from apimachinery validation:
+			// https://github.com/kubernetes/apimachinery/blob/master/pkg/util/validation/validation.go#L178
```
barkbay, line 233: nit: should be a permalink.

support/reattach-pv/main.go
```diff
@@ -207,13 +217,40 @@ type MatchingVolumeClaim struct {
...
 			continue
 		}
-		claim, expected := claims[types.NamespacedName{Namespace: pv.Spec.ClaimRef.Namespace, Name: pv.Spec.ClaimRef.Name}]
+		expectedClaimName := pv.Spec.ClaimRef.Name
+		// if you're building a newly named cluster, from a previous cluster's PVs, we'll
+		// need to extract the nodeSetName, and ordinal from the PV's claimref.Name,
+		// and replace the expected claim name to be the newly generated cluster's name.
+		if oldEsName != "" {
+			// The following regex defines the expected persistent volume claim name format
+			// The regex patters for both EsName, and NodeSetName groups are directly from apimachinery validation:
```
barkbay, line 232:
```suggestion
			// The regex patterns for both EsName, and NodeSetName groups are directly from apimachinery validation:
```

support/reattach-pv/main.go
```diff
@@ -207,13 +217,40 @@ type MatchingVolumeClaim struct {
...
 		if pv.Spec.ClaimRef == nil {
 			continue
 		}
-		claim, expected := claims[types.NamespacedName{Namespace: pv.Spec.ClaimRef.Namespace, Name: pv.Spec.ClaimRef.Name}]
+		expectedClaimName := pv.Spec.ClaimRef.Name
+		// if you're building a newly named cluster, from a previous cluster's PVs, we'll
+		// need to extract the nodeSetName, and ordinal from the PV's claimref.Name,
+		// and replace the expected claim name to be the newly generated cluster's name.
+		if oldEsName != "" {
+			// The following regex defines the expected persistent volume claim name format
```
barkbay, line 231:
```suggestion
			// The following regex defines the expected persistent volume claim name format.
```

support/reattach-pv/main.go
barkbay: Or maybe the original idea was to use `oldEsName` to get the right PVs?

```golang
previousClaimName := fmt.Sprintf(`^elasticsearch-data-%s-es-([a-z0-9]([-a-z0-9]*[a-z0-9])?)-([0-9])+$`, oldEsName)
r := regexp.MustCompile(previousClaimName)

regexMatches := r.FindStringSubmatch(pv.Spec.ClaimRef.Name)

if len(regexMatches) != 4 {
fmt.Printf("ignoring claim name %s, expected 4 regex, got %d: %v", previousClaimName, len(regexMatches), regexMatches)
  continue
}

nodeSetName := regexMatches[1]
strOrdinal := regexMatches[3]
```

support/reattach-pv/main.go
naemono: yeah it was used previously.  Removed, thanks.

support/reattach-pv/main.go
```diff
@@ -28,6 +30,7 @@ import (
 
 const (
 	esManifestFlag = "elasticsearch-manifest"
+	oldEsNameFlag  = "old-elasticsearch-name"
```
barkbay, line 33: I'm still a bit confused by this flag:
* Its value is never used, we could replace it by `viper.GetBool(oldEsNameFlag)`
* The 2nd parameter in `expectedVolumeClaims` is not used

naemono: Actually, I see your point now @barkbay , and this brings up a concern about matching exactly the PVs of the old cluster.  I'm going to revisit your comment about making the regex directly dependent on the old ES name.

naemono: Commit done, and verified with a new round of testing @barkbay .  The `oldEsName` is now used in the regex to limit the volumes that are used.

support/reattach-pv/main.go
```diff
@@ -207,13 +217,42 @@ type MatchingVolumeClaim struct {
 }
 
 // matchPVsWithClaim iterates over existing pvs to match them to an expected pvc.
-func matchPVsWithClaim(pvs []v1.PersistentVolume, claims map[types.NamespacedName]v1.PersistentVolumeClaim) ([]MatchingVolumeClaim, error) {
+func matchPVsWithClaim(pvs []v1.PersistentVolume, claims map[types.NamespacedName]v1.PersistentVolumeClaim, es esv1.Elasticsearch, oldESName string) ([]MatchingVolumeClaim, error) {
+	// The following regex defines the expected persistent volume claim name format.
+	// The regex patterns for both EsName, and NodeSetName groups are directly from apimachinery validation:
+	// https://github.com/kubernetes/apimachinery/blob/6809593a70331498e028baf9baa3b14b10c75214/pkg/util/validation/validation.go#L178
+	var (
+		r = regexp.MustCompile(fmt.Sprintf(`^elasticsearch-data-(%s)-es-([a-z0-9]([-a-z0-9]*[a-z0-9])?)-([0-9])+$`, oldESName))
```
thbkrkr, line 225: I don't really understand why this is `([a-z0-9]([-a-z0-9]*[a-z0-9])?)` and not just `([a-z0-9][-a-z0-9]*[a-z0-9])` to capture the nodesetName?

I like to use named capturing group to auto-document regexp: `^elasticsearch-data-(?P<clusterName>%s)-es-(?P<nodesetName>[a-z0-9][-a-z0-9]*[a-z0-9])-(?P<ordinal>[0-9])+$`.

support/reattach-pv/main.go
```diff
@@ -207,13 +217,42 @@ type MatchingVolumeClaim struct {
...
+			if err != nil {
+				continue
+			}
+			expectedClaimName = fmt.Sprintf(
+				"%s-%s",
+				volume.ElasticsearchDataVolumeName,
+				sset.PodName(esv1.StatefulSet(es.Name, nodeSetName), int32(ordinal)))
+		}
+
+		claim, expected := claims[types.NamespacedName{Namespace: pv.Spec.ClaimRef.Namespace, Name: expectedClaimName}]
```
thbkrkr, line 256: Instead of using a regexp, maybe we could just do a replacement since the PVC names prefix is quite specific (`elasticsearch-data-%s-es`).

```go
if oldEsName != "" {
	expectedClaimName = strings.ReplaceAll(expectedClaimName, pvcPrefix(oldEsName), pvcPrefix(newEsName))
}
```
```go
func pvcPrefix(clusterName string) string {
	return esv1.ESNamer.Suffix(fmt.Sprintf("%s-%s", volume.ElasticsearchDataVolumeName, clusterName))
}
```

support/reattach-pv/main.go
```diff
@@ -207,13 +217,42 @@ type MatchingVolumeClaim struct {
 }
 
 // matchPVsWithClaim iterates over existing pvs to match them to an expected pvc.
-func matchPVsWithClaim(pvs []v1.PersistentVolume, claims map[types.NamespacedName]v1.PersistentVolumeClaim) ([]MatchingVolumeClaim, error) {
+func matchPVsWithClaim(pvs []v1.PersistentVolume, claims map[types.NamespacedName]v1.PersistentVolumeClaim, es esv1.Elasticsearch, oldESName string) ([]MatchingVolumeClaim, error) {
```
thbkrkr, line 220:
```suggestion
func matchPVsWithClaim(pvs []v1.PersistentVolume, claims map[types.NamespacedName]v1.PersistentVolumeClaim, oldEsName string, newEsName string) ([]MatchingVolumeClaim, error) {
```

support/reattach-pv/main.go
```diff
@@ -207,13 +217,43 @@ type MatchingVolumeClaim struct {
...
```
barkbay, line 220:
```suggestion
func matchPVsWithClaim(pvs []v1.PersistentVolume, claims map[types.NamespacedName]v1.PersistentVolumeClaim, es esv1.Elasticsearch) ([]MatchingVolumeClaim, error) {
	oldESName := viper.GetString(oldEsNameFlag)
```

support/reattach-pv/main.go
```diff
@@ -40,6 +43,8 @@ var Cmd = &cobra.Command{
 			fmt.Println("Running in dry run mode")
 		}
 
+		oldEsName := viper.GetString(oldEsNameFlag)
```
barkbay, line 46: nit: it can be moved closer to where it is used, see my other suggestion.
```suggestion
```

support/reattach-pv/main.go
```diff
@@ -207,13 +217,43 @@ type MatchingVolumeClaim struct {
...
+		if oldESName != "" {
+			regexMatches := r.FindStringSubmatch(pv.Spec.ClaimRef.Name)
+			// 5 matches are expected here, as the first match is the full string, and the next is the ES name, NodeSetName+internal match, and Ordinal.
+			if len(regexMatches) != 5 {
+				continue
+			}
+			nodeSetName := regexMatches[2]
+			strOrdinal := regexMatches[4]
+			ordinal, err := strconv.Atoi(strOrdinal)
+			if err != nil {
```
barkbay, line 246: I would log this error to stderr since I don't think it is expected.

support/reattach-pv/main.go
```diff
@@ -207,13 +217,42 @@ type MatchingVolumeClaim struct {
...
 
 // matchPVsWithClaim iterates over existing pvs to match them to an expected pvc.
-func matchPVsWithClaim(pvs []v1.PersistentVolume, claims map[types.NamespacedName]v1.PersistentVolumeClaim) ([]MatchingVolumeClaim, error) {
+func matchPVsWithClaim(pvs []v1.PersistentVolume, claims map[types.NamespacedName]v1.PersistentVolumeClaim, es esv1.Elasticsearch, oldESName string) ([]MatchingVolumeClaim, error) {
+	// The following regex defines the expected persistent volume claim name format.
+	// The regex patterns for both EsName, and NodeSetName groups are directly from apimachinery validation:
+	// https://github.com/kubernetes/apimachinery/blob/6809593a70331498e028baf9baa3b14b10c75214/pkg/util/validation/validation.go#L178
+	var (
+		r = regexp.MustCompile(fmt.Sprintf(`^elasticsearch-data-(%s)-es-([a-z0-9]([-a-z0-9]*[a-z0-9])?)-([0-9])+$`, oldESName))
+	)
```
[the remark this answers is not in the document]
barkbay: On a second thought I'm wondering if the regexp compiles if `oldESName` is an empty string? 🤔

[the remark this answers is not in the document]
naemono: https://go.dev/play/p/_EdafBggdFK

It compiles fine.

support/reattach-pv/main.go
naemono: @thbkrkr I tested both, and you're right.  I initially thought something like `mycluster-thing-name01` would require the trailing `?`, but now I see it does not.

I also removed the named captures, as I wasn't using the names, only the indexes, but I see your point. 

After your below comment about simply replacing the prefix, this whole section is likely going to be removed.

support/reattach-pv/main.go
naemono: I simply removed the print, as it's unnecessary.

support/reattach-pv/main.go
naemono: I'm making these changes, and running some tests to ensure validity, but this certainly cleans up this section.  Thanks @thbkrkr .

naemono: Tested successfully.  This cleaned things up nicely @thbkrkr

support/reattach-pv/README.md
```diff
@@ -39,6 +51,34 @@ go build
 ./reattach-pv --elasticsearch-manifest elasticsearch.yml
 ```
```
thbkrkr, line 52: Instead of repeating the whole tool usage, we could just improve the example. I think we can skip the `go build`.

------------------------------

Example assuming `cluster-A` was accidently deleted:

```
# re-create the cluster with the same name
./reattach-pv --elasticsearch-manifest cluster-A.yml --dry-run

# re-create the cluster with a new name
reattach-pv --elasticsearch-manifest cluster-B.yml --old-elasticsearch-name cluster-A --dry-run

# if everything seems ok, then, execute again without the dry-run flag
```

support/reattach-pv/README.md
```diff
@@ -1,21 +1,32 @@
...
 * The Elasticsearch resource to re-create has the exact same specs as the deleted one. Same cluster name, same node sets, same count, etc.
 * The current default kubectl context targets the desired Kubernetes cluster.
 
+2. If building a new cluster, with a new name, from existing unused PVs from a previously deleted, and re-created cluster
+
+* The Elasticsearch resource to create does not exist in Kubernetes. (The previous cluster with the previous name can exist, using new PVs)
+* All PersistentVolumes of the previous cluster still exist with the status `Released`.
+* The Elasticsearch resource to create has the exact same specs (same node sets, same count, etc.) as the deleted, and re-created one, but with different cluster name.
+* The current default kubectl context targets the desired Kubernetes cluster.
+
```
thbkrkr, line 25: I think the shorter the doc, the more likely it is to be read. I find it a bit hard to read. It took me a while to understand the differences between the two. 

Only `* All PersistentVolumeClaims of the previous cluster do not exist anymore` doesn't apply for the second case. We could rephrase it in `* All PersistentVolumeClaims to re-create does not exist anymore` but I think `* The Elasticsearch resource to re-create does not exist in Kubernetes.` covers that.

So we could just keep one list of conditions without the one for the PersistentVolumeClaims and just mention that the cluster name can be the same or not?

```suggestion
* The Elasticsearch resource to re-create does not exist in Kubernetes.
* All PersistentVolumes of the deleted cluster still exist with the status `Released`.
* The Elasticsearch resource to re-create has the exact same nodeSet specifications as the deleted one (same nodeSet names and counts).
* The current default kubectl context targets the desired Kubernetes cluster.

The Elasticsearch resource to be recreated can have the same name as the deleted one or a new name. In the second case, you must provide the name of the deleted cluster through the flag `--old-elasticsearch-name`.

```

thbkrkr (commented):
Code LGTM. I left two comments to try to make the doc a little more clear and concise.

naemono:
README changes applied @thbkrkr .  Let me know if all looks well.  Thanks.

PR #6124, opened 2022-10-28
Remove vegeta use elasticseach-go for load test

pebrc:
Purpose of this change is to remove the flakiness of the TestMutationWhileLoadTesting test. 

Current status:
* 8.x test passes
* 7.x test passes
* 6.x test fails consistently with results like `duration: 7m21.356621542s, number of requests: 1354, requests/s 3.067814, errors: map[failed request: 503 Service Unavailable:56]`

I propose to skip the test for 6.x, alternatively we could introduce a version specific unavailability budget similar to the existing ContinuousHealthCheck to allow for different amounts of unavailability:

https://github.com/elastic/cloud-on-k8s/blob/d9addf1e1a540a0d1dfc70f3ff550593e3ba09aa/test/e2e/test/elasticsearch/steps_mutation.go#L32-L52

With 7.x and 8.x we still might see some test failures given the 3 retries the client does might not suffice in practice to cover the unavailability in the worst case scenario (the continuous health check allows for 40 seconds of downtime in 8.x) 

I have my doubts about the general usefulness of this test though. It never was a real load test to begin with, even with `vegeta` it only hit `/` and it never authenticated so from an Elasticsearch point of view all requests were relatively cheap error responses. No documents were ingested no queries were run. In essence this test is the same as the continuous health check we run during mutations anyway, just with a much higher resolution (10 reqs/s instead of 0.3 reqs/s)

I would be interested in the reviewers opinions on:

1. should we remove the test altogether 
2. should we have a unavailability budget if we keep it and keep testing 6.x
3. should we skip testing 6.x as proposed by me in the PR


pebrc:
run/e2e-test match=TestMutationWhileLoadTesting

pebrc:
run/e2e-test match=TestMutationWhileLoadTesting

pebrc:
run/e2e-tests match=TestMutationWhileLoadTesting

pebrc:
run/e2e-tests match=TestMutationWhileLoadTesting

pebrc:
run/e2e-tests match=TestMutationWhileLoadTesting

pebrc:
run/e2e-tests match=TestMutationWhileLoadTesting

PR #6143, opened 2022-11-04
Do not run multiple lint operations in parallel when releasing.

naemono:
We recently adjusted the release flow to add a `FIPS` compatible image, and added a parallel step which was running multiple `ci-check` make targets, which include `golangci-lint`, which we have set to not allow parallel runs.  This was causing some nightly builds to timeout with the following error:

```
ERRO Timeout exceeded: try increasing it by passing --timeout option 
```

This change adjusts the flow to:
1. Run the `ci-check` target prior to any build operations.
2. Run the 2x build targets in parallel.

### DIFF
--- a/.ci/pipelines/build.Jenkinsfile
+++ b/.ci/pipelines/build.Jenkinsfile
@@ -34,19 +34,24 @@ pipeline {
                         sh 'make -C .ci TARGET=ci ci'
                     }
                 }
+                stage('Run ci checks prior to build') {
+                    steps {
+                        sh 'make -C .ci TARGET=ci-check ci'
+                    }
+                }
                 stage('build') {
                     failFast true
                     parallel {
                         stage("build and push operator image") {
                             steps {
                                 sh '.ci/setenvconfig build'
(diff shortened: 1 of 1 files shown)


.ci/pipelines/build.Jenkinsfile
```diff
@@ -34,19 +34,24 @@ pipeline {
                         sh 'make -C .ci TARGET=ci ci'
                     }
                 }
+                stage('Run ci checks prior to build') {
+                    steps {
+                        sh 'make -C .ci TARGET=ci-check ci'
```
pebrc, line 39: I think this step is redundant we already have it on line 29 https://github.com/elastic/cloud-on-k8s/pull/6143/files#diff-68c3b03af5147c61c49f4b3316dcdd00fc6c3566986ddf0a12f30f3346cb95a5L29

naemono: Good point @pebrc ; removed.  Interesting, as it's included in `ci-release` target, so we were running it twice, it seems.

pebrc (approved):

PR #6146, opened 2022-11-07
Remove parallelism from fips build

naemono:
The nightly build, which is now including a FIPS build is failing without any specific error:

```
18:16:44  docker.elastic.co/eck-ci/eck-ci-tools:f2e52f24
18:16:45  >> Image == docker.elastic.co/eck-snapshots/eck-operator:2.6.0-SNAPSHOT-2022-11-05-0d0dbea3
18:16:45  >> Image == docker.elastic.co/eck-snapshots/eck-operator:2.6.0-SNAPSHOT-2022-11-05-0d0dbea3
18:16:53  make: *** [Makefile:238: build-operator-multiarch-image] Error 1
18:16:54  Makefile:48: recipe for target 'ci-internal' failed
18:16:54  make[1]: *** [ci-internal] Error 2
18:16:54  eck-multi-arch already exists
18:16:54  make: *** [Makefile:239: build-operator-multiarch-image] Terminated
```

I'm wondering if the parallelism is more trouble than it's worth here, and am just removing it.

PR #6149, opened 2022-11-08
Manifests uploads with FIPS mode

thbkrkr:
In #6146, I forgot to consider the upload yaml step. It needs to be moved after the build steps.

Recent issues with building the 2 versions in parallel stem from `OPERATOR_IMAGE` override not working when `ENABLE_FIPS=true` due to `?=`.

```diff
OPERATOR_IMAGE               ?= $(BASE_IMG):$(IMG_VERSION)

ifeq ($(ENABLE_FIPS),true)
-	OPERATOR_IMAGE           ?= $(BASE_IMG)-fips:$(IMG_VERSION)
+	OPERATOR_IMAGE           := $(BASE_IMG)-fips:$(IMG_VERSION)
endif
```

Also we can't just call `build-operator-image` instead of `ci-release` like we did with #6143 or we won't `generate` the manifests.

In FIPS mode, `-prefix` is appended to the operator image name but the upload yaml script only uses the operator image version to upload the manifests. If we want to upload the all-in-one manifests in FIPS mode the script needs to be adapted to push manifests with a different version.

Last thing, I think we can remove the indirection `VERSION=IMG_VERSION` for the yaml upload which makes it possible to remove `VERSION="${sh(returnStdout: true, script: '. ./.env; echo $IMG_VERSION').trim()}"` from the pipeline.


.ci/Makefile
```diff
@@ -113,10 +113,10 @@ yaml-upload: VAULT_AWS_CREDS = secret/cloud-team/cloud-ci/eck-release
...
 yaml-upload: AWS_SECRET_ACCESS_KEY = $(shell VAULT_TOKEN=$(VAULT_TOKEN) $(vault) read -address=$(VAULT_ADDR) -field=secret-access-key $(VAULT_AWS_CREDS))
 yaml-upload:
-ifndef VERSION
-	$(error VERSION not set to upload YAML to S3)
+ifndef IMG_VERSION
+	$(error IMG_VERSION not set to upload YAML to S3)
 endif
 	@ $(MAKE) \
-		DOCKER_OPTS="-e VERSION=$(VERSION) -e AWS_ACCESS_KEY_ID=$(AWS_ACCESS_KEY_ID) -e AWS_SECRET_ACCESS_KEY=$(AWS_SECRET_ACCESS_KEY)" \
+		DOCKER_OPTS="-e ENABLE_FIPS=$(ENABLE_FIPS) -e VERSION=$(IMG_VERSION) -e AWS_ACCESS_KEY_ID=$(AWS_ACCESS_KEY_ID) -e AWS_SECRET_ACCESS_KEY=$(AWS_SECRET_ACCESS_KEY)" \
```
naemono, line 120:
```suggestion
		DOCKER_OPTS="-e VERSION=$(IMG_VERSION) -e AWS_ACCESS_KEY_ID=$(AWS_ACCESS_KEY_ID) -e AWS_SECRET_ACCESS_KEY=$(AWS_SECRET_ACCESS_KEY)" \
```
Looking at the `ci-internal` target, it does 2 things:

1. calls ci-build-image, which builds `.ci/Dockerfile`, which is contains the needed ci tools, such as golangci-lint, shellcheck, gcloud, eksctl, helm, etc, which doesn't need/use the `ENABLE_FIPS` env var.
2. run `docker run` using the `DOCKER_CMD`, which will be `hack/manifest-upload.sh`, which also doesn't use the `ENABLE_FIPS` env var.

I think this can be removed here.

.ci/pipelines/build.Jenkinsfile
```diff
@@ -37,37 +37,31 @@ pipeline {
...
                     parallel {
-                        stage("build and push operator image") {
+                        stage("build and push operator image and manifests") {
                             agent {
                                 label 'linux'
                             }
                             steps {
                                 sh '.ci/setenvconfig build'
-                                sh 'make -C .ci license.key TARGET=build-operator-multiarch-image ci'
+                                sh 'make -C .ci license.key TARGET=ci-release ci'
```
naemono, line 46: `ci-release` runs this:

```
ci-release: clean ci-check build-operator-multiarch-image
```

`ci-check` runs this, which includes the `generate` step:

```
ci-check: check-license-header lint shellcheck generate check-local-changes check-predicates
```

Since we run `ci-check` here https://github.com/elastic/cloud-on-k8s/pull/6149/files#diff-68c3b03af5147c61c49f4b3316dcdd00fc6c3566986ddf0a12f30f3346cb95a5L29, prior to calling these build steps, I was trying to avoid re-running that step.

.ci/pipelines/build.Jenkinsfile
```diff
@@ -37,37 +37,31 @@ pipeline {
...
                                 label 'linux'
                             }
+                            environment {
+                                ENABLE_FIPS="true"
+                            }
                             steps {
                                 sh '.ci/setenvconfig build'
-                                sh 'make -C .ci license.key TARGET=build-operator-multiarch-image ci ENABLE_FIPS=true'
+                                sh 'make -C .ci license.key TARGET=ci-release ci'
+                                sh 'make -C .ci yaml-upload'
```
naemono, line 60: Could you please help me understand why we want to upload the manifests in fips mode?  Wasn't the intent of having a FIPS image/container simply to allow someone to do something like:

```
helm install elastic-operator -n elastic-system elastic/eck-operator --set=image.repository=docker.elastic.co/eck/eck-operator-fips --version 2.5.0
```

Or edit their eck-operator installation after install to update the image to be fips compliant?

Makefile
```diff
@@ -62,10 +62,10 @@ OPERATOR_DOCKERHUB_IMAGE_UBI ?= docker.io/elastic/$(IMG_NAME)-ubi8:$(IMG_VERSION
...
 ifeq ($(ENABLE_FIPS),true)
 	GO_TAGS += goexperiment.boringcrypto
-	OPERATOR_IMAGE               ?= $(BASE_IMG)-fips:$(IMG_VERSION)
-	OPERATOR_IMAGE_UBI           ?= $(BASE_IMG)-ubi8-fips:$(IMG_VERSION)
-	OPERATOR_DOCKERHUB_IMAGE     ?= docker.io/elastic/$(IMG_NAME)-fips:$(IMG_VERSION)
-	OPERATOR_DOCKERHUB_IMAGE_UBI ?= docker.io/elastic/$(IMG_NAME)-ubi8-fips:$(IMG_VERSION)	
+	OPERATOR_IMAGE               := $(BASE_IMG)-fips:$(IMG_VERSION)
+	OPERATOR_IMAGE_UBI           := $(BASE_IMG)-ubi8-fips:$(IMG_VERSION)
+	OPERATOR_DOCKERHUB_IMAGE     := docker.io/elastic/$(IMG_NAME)-fips:$(IMG_VERSION)
+	OPERATOR_DOCKERHUB_IMAGE_UBI := docker.io/elastic/$(IMG_NAME)-ubi8-fips:$(IMG_VERSION)
```
naemono, line 68: Thanks for catching this.

.ci/pipelines/build.Jenkinsfile
thbkrkr: Makes sense until we added the `agent`. So we can call `generate-crds-v1 build-operator-multiarch-image`.

thbkrkr:
Hm, I see another issue. When we call `make -C .ci TARGET=xxx ci ENABLE_FIPS=true`, it calls behind the scene `docker run docker.elastic.co/eck-ci/eck-ci-tools "make xxx"` without propagating the environment variable in the container. Currently, all environment variables are made available exclusively through the `.env` file generated by `setenvconfig` always called before `make -C .ci TARGET=xxx ci`.

We could use `IMG_SUFFIX` to generate the image name `-fips` variants and set `GO_TAGS` in `setenvconfig`.

```sh
# setenvconfig

  IMG_SUFFIX=""
  GO_TAGS="release"

  if [[ "${ENABLE_FIPS:-}" == "true" ]]; then
    IMG_SUFFIX=-fips
    GO_TAGS="release goexperiment.boringcrypto"
  fi

write_env <<ENV
REGISTRY_NAMESPACE=eck-snapshots
IMG_SUFFIX=${IMG_SUFFIX}

GO_TAGS=${GO_TAGS}
```

But given that with buildkite, we are going to remove the docker layer as we are already in a container, we can keep the logic in the root Makefile and just specifically propagate the env var. Not great but it will go away once we migrate this release part on buildkite.

```diff
--- a/.ci/Makefile
+++ b/.ci/Makefile
@@ -44,13 +44,13 @@ ci:
 ci-interactive:
        @ $(MAKE) DOCKER_OPTS=-i DOCKER_CMD=bash ci-internal
 
ci-internal: ci-build-image
        @ docker volume create --name $(ECK_CI_VOLUME) > /dev/null && \
          docker run --rm -t $(DOCKER_OPTS) \
                -v /var/run/docker.sock:/var/run/docker.sock \
                -v $(ECK_CI_VOLUME):/root/ \
                -v $(ROOT_DIR):$(GO_MOUNT_PATH) \
-               -e SHARED_VOLUME_NAME=$(ECK_CI_VOLUME) \
+               -e SHARED_VOLUME_NAME=$(ECK_CI_VOLUME) -e ENABLE_FIPS=$(ENABLE_FIPS) \
```

naemono (approved):
Lgtm now.

PR #6157, opened 2022-11-09
Expose K8S client QPS as a flag

barkbay:
This is a proposal to expose the QPS setting from the K8S Go client as a new operator flag.

FWIW here is a comparison of the throughput and latency with the default QPS and a QPS of 100 during a benchmark with a same workload:

## Default QPS of 5

image

## QPS of 100

image

With a QPS of 100 the latency is significantly reduced (~from `8 sec` to less than `1 sec` for the 2/3 of the test), even if the throughput drops at the end. First investigation indicates that this seems to be related to the CPU being throttled.

Burst is set to twice the QPS value to keep the default ratio from the Go client. I did not have the opportunity to play with it so far.

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -256,6 +256,11 @@ func Command() *cobra.Command {
 		60*time.Second,
 		"Timeout for requests made by the Kubernetes API client.",
 	)
+	cmd.Flags().Float32(
+		operator.KubeClientQPS,
+		0,
+		"Kubernetes API client queries per second.",
+	)
 	cmd.Flags().Bool(
 		operator.ManageWebhookCertsFlag,
 		true,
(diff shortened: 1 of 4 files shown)


pebrc (approved):
LGTM!

deploy/eck-operator/templates/configmap.yaml
```diff
@@ -25,6 +25,9 @@ data:
     {{- end }}
     set-default-security-context: {{ .Values.config.setDefaultSecurityContext }}
     kube-client-timeout: {{ .Values.config.kubeClientTimeout }}
+    {{- if .Values.config.kubeClientQPS }}
+    telemetry-interval: {{ .Values.config.kubeClientQPS }}
```
thbkrkr, line 29:
```suggestion
    kube-client-qps: {{ int .Values.config.kubeClientQPS }}
```

cmd/manager/main.go
```diff
@@ -256,6 +256,11 @@ func Command() *cobra.Command {
 		60*time.Second,
 		"Timeout for requests made by the Kubernetes API client.",
 	)
+	cmd.Flags().Float32(
+		operator.KubeClientQPS,
+		0,
+		"Kubernetes API client queries per second.",
```
thbkrkr, line 262: I'm nitpicking?
```suggestion
		"Maximum number of queries per second to the Kubernetes API.",
```

thbkrkr (approved):
:+1:

I bet at some point we will want to control the burst independently of the qps. Let's see :)

PR #6162, opened 2022-11-10
echo username and details of .env file for debugging ci failures.

naemono:
Currently building nightly releases is failing with:

```
16:31:26  + .ci/setenvconfig build
16:31:26  .ci/setenvconfig: line 16: .ci/../.env: Permission denied
16:31:26  [Pipeline] }
16:31:26  [Pipeline] // script
16:31:26  Error when executing success post condition:
```

This seems to indicate that writing/updating the .env file is failing, as if it was written by either
1. a different user
2. or with permissions that aren't allowing u+w

This change will at least help debug the issue, but not solve it.

### DIFF
--- a/.ci/pipelines/build.Jenkinsfile
+++ b/.ci/pipelines/build.Jenkinsfile
@@ -86,7 +86,7 @@ pipeline {
     post {
         success {
             script {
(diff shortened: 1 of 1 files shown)


naemono:
Going to merge this without approval, as it can't make the failures any worse, and can only help debugging the issue.  To be immediately removed upon resolution of current issues.

PR #6161, opened 2022-11-09
Attempting to fix ci pipeline

naemono:
Suggestions from @thbkrkr

PR #6163, opened 2022-11-10
Fix .env permission in CI pipelines

thbkrkr:
Creating the `.env` file inside the ci docker container just to pass shellcheck was a bad idea. The file is created as root and this prevents `setenvconfig` which is run outside the docker ci container from updating it.

```sh
08:33:22  + .ci/setenvconfig e2e/main
08:33:22  .ci/setenvconfig: line 16: .ci/../.env: Permission denied
```

This commit removes the creation of `.env` in the shellcheck make target and instead moves the `setenvconfig` call before `make ci-check`.

Following of #6161.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -35,6 +35,7 @@ steps:
 
       - label: ":go: checks"
         commands:
+          - .ci/setenvconfig pr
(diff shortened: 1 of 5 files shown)


pebrc (approved):
🤞

PR #6164, opened 2022-11-10
Get license key to run ci-check

thbkrkr:
`setenvconfig` is now called first and sets `GO_TAGS=release` and `LICENSE_PUBKEY=/go/src/github.com/elastic/cloud-on-k8s/.ci/license.key`. 
It causes an issue when we run `make ci-check` which executes `make go-generate` which depends on the `license.key` file when `GO_TAGS=release`.

```
# reminder
> grep go: pkg/controller/common/license/pubkey.go
//go:build release
//go:generate go run ../../../../cmd/license-initializer/main.go --out zz_generated.pubkey.go
```

Let's just recover the missing file. It's already fetched later via `get-test-artifacts` in the main/snapshot jobs but I want to avoid more changes.

### DIFF
--- a/.ci/pipelines/build.Jenkinsfile
+++ b/.ci/pipelines/build.Jenkinsfile
@@ -27,7 +27,7 @@ pipeline {
                 stage('Run checks') {
                     steps {
                         sh '.ci/setenvconfig build'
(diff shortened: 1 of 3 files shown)


naemono (approved):
sounds fine to me at this point.

PR #6169, opened 2022-11-10
Fix regression in cloud-on-k8s-e2e-tests-stack-versions job

thbkrkr:
Fixes a regression in the job `cloud-on-k8s-e2e-tests-stack-versions` introduced by the refactoring of the k8s cluster names created to run the e2e tests (770b4da). 

The generated cluster name (for buildkite) was used as the stack version:

```
--- FAIL: TestGlobalCA (1.26s)
panic: semver: Parse(eck-715-869-e2e): No Major.Minor.Patch elements found [recovered]
	panic: semver: Parse(eck-715-869-e2e): No Major.Minor.Patch elements found
```

The cluster name can be generated (buildkite) or provided as an argument (jenkins) and I forgot to remain compliant with the latter and I also missed the reorder of the two args.

### DIFF
--- a/.ci/pipelines/e2e-tests-stack-versions-gke.Jenkinsfile
+++ b/.ci/pipelines/e2e-tests-stack-versions-gke.Jenkinsfile
@@ -210,7 +210,7 @@ pipeline {
 }
 
 def runWith(lib, failedTests, clusterName, stackVersion) {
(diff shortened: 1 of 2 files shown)


thbkrkr:
There is another regression to clean ocp cluster:
```
21:13:14  .ci/setenvconfig: line 815: ocp_version: unbound variable
```

PR #6182, opened 2022-11-18
Minor doc update

DJRickyB:


barkbay (commented):
Thanks for the fix 🙇 

It seems however that "ad hoc" comes from Latin words (not from your PR, I just noticed it now). I know that we have tried to avoid latinisms in the past, see https://github.com/elastic/cloud-on-k8s/pull/5456 for example.
My feeling is that we should maybe change this title? Maybe `Taking JVM heap dumps` would be fine?
CC @alaudazzi 👋 who may confirm my statement.

barkbay (approved):
LGTM
Thanks 👍

PR #6194, opened 2022-11-25
Extend Elasticsearch pre-stop hooks

andreaskapfer:
This PR adds the possibility to add additional actions to the Elasticsearch pre-stop hook via the environment variable `PRE_STOP_ACTIONS`. The first (and at the moment only) action is `flush` see [Rolling restart](https://www.elastic.co/guide/en/elasticsearch/reference/current/restart-cluster.html#restart-cluster-rolling).

Also a bug in some tests is fixed (assertions where not called).

### DIFF
--- a/docs/orchestrating-elastic-stack-applications/elasticsearch/prestop.asciidoc
+++ b/docs/orchestrating-elastic-stack-applications/elasticsearch/prestop.asciidoc
@@ -36,3 +36,6 @@ spec:
             - name: PRE_STOP_ADDITIONAL_WAIT_SECONDS
               value: "5"
 ----
+
+It is possible to specify additional actions to be carried out before the `Pod` is stopped via the environment variable `PRE_STOP_ACTIONS` (different actions separated by a comma).
+The only supported action at the moment is `flush` which performs a flush as suggested in link:https://www.elastic.co/guide/en/elasticsearch/reference/current/restart-cluster.html#restart-cluster-rolling[Rolling restart].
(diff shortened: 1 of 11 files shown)


pebrc:
@andreaskapfer thanks for your contribution but the operator is already performing a flush during rolling upgrades/restrats

https://github.com/elastic/cloud-on-k8s/blob/85aa75b7a7765330475c0b92a0851939ac3d9578/pkg/controller/elasticsearch/driver/upgrade.go#L276-L303


I would rather try to minimise the amount of logic we run in a pre-stop hook if possible. Did you find the existing functionality lacking in some regard that we are not aware of?

Also we are still supporting older versions of Elasticsearch where the recommended operation is a [synced flush](https://www.elastic.co/guide/en/elasticsearch/reference/6.8/restart-upgrade.html) so this code would need to be version dependent.

andreaskapfer:
@pebrc Thank you for your response. I actually was not aware of the flush being performed directly by the operator. Never mind, I was just playing around and wanted to contribute something useful on the way.

Nevertheless, I found two bugs in `predefined_test.go` (existing assertions were not called). Should I include the fixes, which are also in this PR, in a separate PR?

pebrc:
> Nevertheless, I found two bugs in `predefined_test.go` (existing assertions were not called). Should I include the fixes, which are also in this PR, in a separate PR?

That would be great!

thbkrkr:
Looking forward to a separate PR that fixes the assertions in `predefined_test.go`. Thank you Andreas.

PR #6226, opened 2022-12-12
Helm Release Pipeline

naemono:
This adds a Buildkite pipeline for releasing of ECK Helm charts.
This also adds a custom Helm release tool.

Included:

1. Pipeline for releasing ECK resource Helm charts.
2. Pipeline for releasing ECK Operator Helm Chart.
3. Go tool to handle releasing Helm Charts to Elastic Helm Buckets/Repositories.

What's been manually tested:

* Releasing of ECK resources Helm charts to both dev and production Elastic Helm Buckets/Repositories. (resolving eck-stack chart dependency issue)
* Releasing of ECK Operator Helm chart to dev Elastic Helm Bucket/Repository.

In progress:

- [x] Currently testing Buildkite integration (will fully test during actual release).


hack/helm/release/cmd/root.go
```diff
@@ -0,0 +1,172 @@
...
+
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+
+	"github.com/elastic/cloud-on-k8s/hack/helm/release/internal/helm"
+)
+
+const (
+	// viper flags
+	bucketFlag          = "bucket"
```
thbkrkr, line 21:
```suggestion
	bucketFlag          = "bucket"
```

hack/helm/release/cmd/root.go
```diff
@@ -0,0 +1,172 @@
...
+
+func releaseCmd() *cobra.Command {
+	releaseCommand := &cobra.Command{
+		Use:     "release",
+		Short:   "Release ECK Helm Charts",
+		Long:    `This command will Release ECK Helm Charts to a given Environment.`,
+		Example: fmt.Sprintf("  %s", "release --charts-dir=./deploy --upload-index --dry-run=false"),
+		PreRunE: validate,
+		RunE: func(_ *cobra.Command, _ []string) error {
+			log.Printf("Releasing charts in (%s) to bucket (%s) in repo (%s)\n", viper.GetString("charts-dir"), viper.GetString("bucket"), viper.GetString("charts-repo-url"))
```
thbkrkr, line 61:
```suggestion
			log.Printf("Releasing charts in (%s) to bucket (%s) in repo (%s)\n", viper.GetString(chartsDirFlag), bucket, chartsRepoURL)
```

hack/helm/release/cmd/root.go
```diff
@@ -0,0 +1,172 @@
...
+			return helm.Release(
+				helm.ReleaseConfig{
+					ChartsDir:           viper.GetString("charts-dir"),
```
thbkrkr, line 64:
```suggestion
					ChartsDir:           viper.GetString(chartsDirFlag),
```

hack/helm/release/cmd/root.go
```diff
@@ -0,0 +1,172 @@
...
+					Bucket:              bucket,
+					ChartsRepoURL:       chartsRepoURL,
+					CredentialsFilePath: viper.GetString("credentials-file"),
```
thbkrkr, line 67:
```suggestion
					CredentialsFilePath: viper.GetString(credentialsFileFlag),
```

hack/helm/release/cmd/root.go
```diff
@@ -0,0 +1,172 @@
...
+					DryRun:              viper.GetBool("dry-run"),
```
thbkrkr, line 68:
```suggestion
					DryRun:              viper.GetBool(dryRunFlag),
```

hack/helm/release/cmd/root.go
```diff
@@ -0,0 +1,172 @@
...
+					Excludes:            viper.GetStringSlice("excludes"),
```
thbkrkr, line 69:
```suggestion
					Excludes:            viper.GetStringSlice(excludesFlag),
```

hack/helm/release/cmd/root.go
```diff
@@ -0,0 +1,172 @@
...
+	flags.String(
+		chartsDirFlag,
+		"./deploy",
+		"Charts directory which contain helm charts (env: HELM_CHARTS_DIR)",
+	)
+	_ = viper.BindPFlag(chartsDirFlag, flags.Lookup(chartsDirFlag))
+
+	flags.String(
+		credentialsFileFlag,
+		"",
```
thbkrkr, line 93:
```suggestion
		"/tmp/credentials.json",
```

hack/helm/release/cmd/root.go
```diff
@@ -0,0 +1,172 @@
...
+	flags.StringSlice(
+		excludesFlag,
+		[]string{},
+		"Names of helm charts to exclude from release. (env: HELM_EXCLUDES)",
+	)
+	_ = viper.BindPFlag(excludesFlag, flags.Lookup(excludesFlag))
+
+	flags.Bool(
+		enableVaultFlag,
+		false,
```
thbkrkr, line 114: I would enable vault by default.
```suggestion
		true,
```

hack/helm/release/cmd/root.go
naemono: What's the intention here?  I see no changes suggested.

thbkrkr: Oops, the constant can be deleted, it is no longer used.

hack/helm/release/readme.md
```diff
@@ -0,0 +1,48 @@
...
+releaser --env=dev --charts-dir=path/to/deploy --credentials-file=path/to/gcs-bucket-credentials.json --dry-run=false
+```
+
+## Configuration
+
+| Parameter           | Description                                                                                                    | Environment Variable    | Default                            |
+|---------------------|----------------------------------------------------------------------------------------------------------------|-------------------------|------------------------------------|
+| `--env`             | Environment in which to upload Helm chart packages.                                                            | `HELM_ENV`              | `dev`                              |
+| `--charts-dir`      | Full path to directory containing Helm charts to release.                                                      | `HELM_CHARTS_DIR`       | `./deploy`                         |
+| `--credentials-file`| Full path to credentials file to use for GCS bucket.                                                           | `HELM_CREDENTIALS_FILE` | `""`                               |
```
thbkrkr, line 46:
```suggestion
| `--credentials-file`| Full path to credentials file to use for GCS bucket.                                                           | `HELM_CREDENTIALS_FILE` | `"/tmp/credentials.json"`                               |
```

hack/helm/release/internal/helm/helm.go
```diff
@@ -0,0 +1,555 @@
...
+	if err := uploadCharts(ctx, tempDir, conf.noDeps, conf.releaseConf); err != nil {
+		return err
+	}
+	if err := updateIndex(ctx, tempDir, conf); err != nil {
+		return err
+	}
+	if conf.releaseConf.DryRun {
+		log.Printf("Not processing charts with dependencies (%v) as the Helm index isn't being updated, or dry-run is set", conf.withDeps)
+		return nil
+	}
```
thbkrkr, line 239: Why do we shortcut in the middle of `uploadChartsAndUpdateIndex` to not process charts with dependencies? IIUC the index of the charts with deps will be invalid because we didn't update the index of the parent charts. But I wonder if it is really that bad? With dry-run we want to test as much as possible of the process and we upload anything.

hack/helm/release/cmd/root.go
```diff
@@ -0,0 +1,172 @@
...
+		"Environment in which to release helm charts (env: HELM_ENV)",
+	)
+	_ = viper.BindPFlag(envFlag, flags.Lookup(envFlag))
+
+	flags.StringSlice(
+		excludesFlag,
+		[]string{},
+		"Names of helm charts to exclude from release. (env: HELM_EXCLUDES)",
+	)
+	_ = viper.BindPFlag(excludesFlag, flags.Lookup(excludesFlag))
```
thbkrkr, line 109: The trick with the `--char-dir` and the `--excludes` to only release the eck-resources Helm charts is a bit twisted I think. Maybe it would be cleaner if it's possible to move the eck-resources charts to a parent directory at the same level than the operator?

Instead of:
```
release --char-dir deploy/eck-operator
release --char-dir deploy --excludes eck-operator
```

We could do:
```
release --char-dir deploy/eck-operator
release --char-dir deploy/eck-resources
```

.buildkite/pipeline.yml
```diff
@@ -107,3 +107,34 @@ steps:
...
           - "e2e-tests-image-build"
+
+  - group: helm-release
+
+    steps:
+
+      - label: ":helm: release eck-operator helm chart - pipeline upload"
+        # From https://semver.org, prefixed with 'v'.
+        # The '$$' is required when formatting regular expressions: https://buildkite.com/docs/pipelines/conditionals#formatting-regular-expressions
+        if: build.tag =~ /^v(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-((?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$$/
```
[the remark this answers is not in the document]
naemono: Agreed @thbkrkr. Only thought here is that we're going to be releasing helm charts on these tags for versions that the `release` process isn't complete, as, I believe, that process still lives in jenkins, right?  I think I'd prefer to comment this section out for now, and only enable this when we can wait for the release process in BK to be complete.  Thoughts?

hack/helm/release/internal/helm/helm.go
naemono: Ok, I guess that makes sense.  It's been removed.

.buildkite/pipeline.yml
```diff
@@ -107,3 +107,45 @@ steps:
...
+        if: build.tag != null
+        commands:
+          - buildkite-agent pipeline upload .buildkite/pipeline-helm-release-eck-operator.yml
+
+      - label: ":helm: release eck-operator helm chart manually - pipeline upload"
+        if: build.message == "release eck-operator helm chart"
+        commands:
+          - buildkite-agent pipeline upload .buildkite/pipeline-helm-release-eck-operator.yml
+
+      - label: ":helm: helm eck-resources - Chart.yaml validation check (expected fail when no files have changed)"
```
naemono, line 127: This bothers me.  This seemed to be the `proper` way to do this via BK, but it's always failing in BK (which is 'expected'), which causes github 'checks' to always show failed.  I'm investigating another way to do this, or simply removing this `cd` step for now, and requiring manual release.  Thoughts?

naemono: removed per conversation with @thbkrkr

hack/helm/release/internal/helm/helm.go
```diff
@@ -0,0 +1,616 @@
...
+//     is set to another repository (such as a dev helm repo), the dependency's
+//     repository will be re-written to the given 'charts-repo-url' flag.
+//  4. Run 'Helm dependency update' for the Helm Chart.
+//  5. Run 'Helm package chart' for the Helm Chart to generate a tarball.
+//  6. Copy the tarball to the GCS bucket.
+func uploadCharts(ctx context.Context, tempDir string, charts []chart, conf ReleaseConfig) error {
+	type helmRepo struct {
+		name, url string
+	}
+	for _, r := range []helmRepo{{name: "stable", url: stableHelmChartsURL}, {name: conf.Bucket, url: conf.ChartsRepoURL}} {
```
thbkrkr, line 308: We add Helm repositories even if there are no charts to download? Must be harmless I guess?

hack/helm/release/internal/helm/helm.go
```diff
@@ -0,0 +1,616 @@
...
+		if err := addHelmRepository(r.name, r.url); err != nil {
+			return err
+		}
+	}
+	log.Printf("Done adding (%s) repository.\n", stableHelmChartsURL)
```
thbkrkr, line 313: The same message is already printed by `addHelmRepository(name, url)`.
```suggestion
```

hack/helm/release/internal/helm/helm.go
```diff
@@ -0,0 +1,616 @@
...
+			return err
+		}
+	}
+
+	return nil
+}
+
+// addHelmRepository will add the given Helm repository name + url
+// to the local Helm configuration.
+func addHelmRepository(name, url string) error {
```
thbkrkr, line 346: `name` and `url` arguments are not used. We always add "stable", "https://charts.helm.sh/stable" instead, normal?

hack/helm/release/internal/helm/helm.go
```diff
@@ -0,0 +1,616 @@
...
+	}
+	for _, r := range f.Repositories {
+		cr, err := repo.NewChartRepository(r, getter.All(cli.New()))
+		if err != nil {
+			return fmt.Errorf("while getting new chart repository: %w", err)
+		}
+		if _, err = cr.DownloadIndexFile(); err != nil {
+			return fmt.Errorf("while updating repository index for (%s): %w", cr.Config.Name, err)
+		}
+		log.Printf("Updated repository index for (%s)\n", cr.Config.Name)
```
thbkrkr, line 289: I've tested to release in `--env=dev` and I see that we are updating the 2 repositories, is it normal?:

```
2023/01/27 20:32:08 Updated repository index for (elastic) elastic
2023/01/27 20:32:08 Updated repository index for (elastic-dev) elastic-dev
```

.buildkite/scripts/helm/has-any-chartyml-changed.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
```
thbkrkr, line 1: Let's remove it for now because it's no longer in use?

hack/helm/release/internal/helm/helm.go
```diff
@@ -0,0 +1,616 @@
...
+
+	files, err := filepath.Glob(config.sourceGlob)
+	if err != nil {
+		return fmt.Errorf("while search for file glob (%s): %w", config.sourceGlob, err)
+	}
+	if len(files) == 0 {
+		return fmt.Errorf("couldn't file helm package with glob (%s)", config.sourceGlob)
+	}
+	// TODO is this 'helm' prefix always assumed?  I don't think so...
+	// This should be the trailing path of the given repo url, right?
```
thbkrkr, line 438: Can we check this and remove the TODO?

.buildkite/pipeline-helm-release-eck-resources.yml
```diff
@@ -0,0 +1,31 @@
...
+      - label: "release eck-resources helm charts dev"
+        key: helm-release-eck-resources-dev
+        commands:
+          - cd hack/helm/release
+          # ignoring stdout to avoid "go: downloading ..." output.
+          - make build 2> /dev/null
+          - cp bin/releaser /tmp/releaser
+          # if a branch is sent as an environment variable, change to that branch.
+          - if [[ ! -z "$$HELM_BRANCH" ]]; then echo "checking out $$HELM_BRANCH branch" && git checkout $$HELM_BRANCH ; fi
+          - VAULT_ROOT_PATH="secret/ci/elastic-cloud-on-k8s" HELM_CREDENTIALS_FILE=/tmp/credentials.json /tmp/releaser --env=dev --charts-dir=../../../deploy --excludes=eck-operator --enable-vault
```
thbkrkr, line 19: I think we can simplify a bit.
We can avoid the `cd` using `make -C`.
`VAULT_ROOT_PATH` isn't used, instead there is `HELM_VAULT_SECRET`. 
`HELM_CREDENTIALS_FILE` is set to `/tmp/credentials.jso` by default.
` --enable-vault` is true by default.
`--charts-dir` is set to `deploy` by default.

```suggestion
          - make -C hack/helm/release build 2> /dev/null
          - if [[ -n "$$HELM_BRANCH" ]]; then echo "checking out $$HELM_BRANCH branch" && git checkout $$HELM_BRANCH ; fi
          - hack/helm/release/bin/releaser --env=dev --excludes=eck-operator
```

hack/helm/release/Dockerfile
```diff
@@ -0,0 +1,29 @@
+# Build the operator binary
```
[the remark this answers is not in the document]
naemono: Nope, removed.

hack/helm/release/internal/helm/helm.go
naemono: I've updated this logic to break early.  It's harmless though.

hack/helm/release/internal/helm/helm.go
naemono: This was fixed.  I was initially only adding stable, and then realized that for dependencies sake, itself would be needed.  Resolved.

hack/helm/release/internal/helm/helm.go
naemono: Thanks for pointing this out.  I've verified, and since we're using constants now for dev/prod, we can assume we can use the `Path` parsed from "net/url".Parse, with the leading `/` trimmed.  👍

.buildkite/pipeline-helm-release-eck-resources.yml
naemono: Excellent point.  Adjusted all pipelines.

hack/helm/release/readme.md
```diff
@@ -0,0 +1,91 @@
+# ECK Helm Chart Releaser
```
thbkrkr, line 1: I'm a bit concerned about maintaining this readme with this high level of details. The tool already has a `--help` flag that covers a lot of information. We could stay more general.

hack/helm/release/cmd/root.go
```diff
@@ -0,0 +1,171 @@
...
+func init() {
+	cobra.OnInitialize(initConfig)
+}
+
+func releaseCmd() *cobra.Command {
+	releaseCommand := &cobra.Command{
+		Use:     "release",
+		Short:   "Release ECK Helm Charts",
+		Long:    `This command will Release ECK Helm Charts to a given Environment.`,
+		Example: fmt.Sprintf("  %s", "release --charts-dir=./deploy --upload-index --dry-run=false"),
```
thbkrkr, line 57: I think `--upload-index` has been removed.

```suggestion
		Example: fmt.Sprintf("  %s", "release --charts-dir=deploy --dry-run=false"),
```

.buildkite/pipeline.yml
```diff
@@ -107,3 +107,27 @@ steps:
...
+
+      - label: ":helm: release eck-resources helm charts manually - pipeline upload"
+        if: build.message == "release eck-resources helm charts"
+        commands:
+          - buildkite-agent pipeline upload .buildkite/pipeline-helm-release-eck-resources.yml
+
+      - label: ":helm: release all helm charts manually - pipeline upload"
+        if: build.message == "release all helm charts"
+        commands:
+          - buildkite-agent pipeline upload .buildkite/pipeline-helm-release-all.yml
```
thbkrkr, line 133: I think we said we will publish daily snapshots of all helm charts daily from `main` to the `dev` repository. There will be more change on this part when we introduce the pipeline-release.yml definition so I'm ok with merging the PR in this state and moving forward.

hack/helm/release/readme.md
```diff
@@ -0,0 +1,91 @@
...
+
+```
+  curl --request POST --url https://api.buildkite.com/v2/organizations/elastic/pipelines/cloud-on-k8s-operator/builds \
+    --header 'Authorization: Bearer your-bk-token' --header 'Content-Type: application/json' \
+    --data '{
+        "commit": "HEAD",
+        "branch": "main",
+        "env": {
+          "HELM_DRY_RUN": "false",
+          "HELM_BRANCH": "2.6"
```
thbkrkr, line 63: What the interest of using this `HELM_BRANCH` environment variable instead of the regular required properties `commit` and `branch`?

hack/helm/release/internal/helm/helm.go
naemono: I'm assuming you ran this locally?  It's loading your local configuration, which includes both elastic and elastic-dev repositories, and updating all the indexes, which is a harmless operation.  You won't see this in Buildkite.

```
func updateHelmRepositories() error {
	// simulates helm repo update
	f, err := repo.LoadFile(cli.New().RepositoryConfig)
	if err != nil {
		return fmt.Errorf("while loading repository file: %w", err)
	}
	for _, r := range f.Repositories {
		cr, err := repo.NewChartRepository(r, getter.All(cli.New()))
		if err != nil {
			return fmt.Errorf("while getting new chart repository: %w", err)
		}
		if _, err = cr.DownloadIndexFile(); err != nil {
			return fmt.Errorf("while updating repository index for (%s): %w", cr.Config.Name, err)
		}
		log.Printf("Updated repository index for (%s)\n", cr.Config.Name)
	}
	return nil
}
```

For me, it's even a bit more:
```
2023/01/30 13:58:20 Updated repository index for (grafana)
2023/01/30 13:58:21 Updated repository index for (metrics-server)
2023/01/30 13:58:22 Updated repository index for (prometheus-community)
2023/01/30 13:58:23 Updated repository index for (elastic)
2023/01/30 13:58:24 Updated repository index for (jetstack)
2023/01/30 13:58:24 Updated repository index for (cdf)
2023/01/30 13:58:25 Updated repository index for (argo)
2023/01/30 13:58:25 Updated repository index for (nginx-stable)
2023/01/30 13:58:25 Updated repository index for (aws-ebs-csi-driver)
2023/01/30 13:58:26 Updated repository index for (elastic-helm-dev)
```

hack/helm/release/readme.md
naemono: @thbkrkr readme updated to try and explain this.  Let me know if it's more clear.

thbkrkr: `branch` and `commit` properties select the branch from which to build the tool and `HELM_BRANCH` the branch from which to release? It's not clear for me why we couldn't just use `branch` and `commit` for the two. IIUC one interest to separate the two is to be able to release code where the tool didn't exist?

naemono: I see your point. I guess the only possible use case for this after merge of this PR would be if changes were needed to the tooling. But if that was the case then we could just merge the tooling changes prior to running the tooling. I will simply remove this to make things more clear.

thbkrkr (approved):
LGTM

We will make small adjustments when integrating into the release pipeline.

Issue #6236, opened 2022-12-14
ElasticsearchAutoscaler documentation is unclear on supported node roles


brsolomon-deloitte:
The [ECK documentation on `ElasticsearchAutoscaler`](https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-autoscaling.html) is unclear as to which node roles it supports.

First it states:

> Currently, autoscaling is supported for Elasticsearch [data tiers](https://www.elastic.co/guide/en/elasticsearch/reference/current/data-tiers.html) and machine learning nodes.

Perhaps I am misreading, but this seems to imply that the autoscaling API, and hence `ElasticsearchAutoscaler`, only supports data nodes or machine learning nodes, and does not support e.g. ingest nodes.

But then it gives this example:

```yaml
spec:
  ...
  policies:
    - name: data-ingest
      roles: ["data", "ingest" , "transform"]
```

...which seems to indicate that nodes other than data nodes are supported.

* ECK version:

	2.5.0

thbkrkr:
I admit that it is a bit ambiguous. The autoscaling policies only applies to nodes that have **at least** `data` or exactly `ml` role.

In the example, nodes with roles `["data", "ingest", "transform"]` are **data** nodes and also ingest and transform nodes, they are not just specialized `data`.

| nodes.roles |  supported? |
| --------------- | ------------------ |
| `["data"]` | :heavy_check_mark: 
| `["data", "ingest" , "transform"]` | :heavy_check_mark: 
| `["ml"]` | :heavy_check_mark: 
| `["ingest" , "transform"]` | :x: 
| `["data", "ml"]` | :x: 
| `["data", "ml", "ingest" , "transform"]` | :x: 
| `["ml", "ingest"]` | :x: 

Note that you can't autoscale nodes with roles `data` and `ml`. There is a validation rule to detect that ML nodes must be in a dedicated autoscaling policy.

I'll see to remove the ambiguity from the doc and add validation for this.

Hythloday-zero:
This is still catching folks off-guard.  Any movement on amending the docs page on this??

PR #6246, opened 2022-12-16
Add validation rule for ElasticsearchAutoscaler resource

thbkrkr:
- validate that all roles are valid
- ~validate that there is at least `ml` or a `data` role in the node roles~

Relates to #6236.

### DIFF
--- a/pkg/controller/common/autoscaling/validations.go
+++ b/pkg/controller/common/autoscaling/validations.go
@@ -161,16 +161,6 @@ func ValidateAutoscalingPolicies(
 			}
 		}
 
-		// at least ml or a data role is present
-		if !CanBeAutoscaled(autoscalingSpec.Roles) {
-			errs = append(
-				errs,
-				field.Invalid(
-					autoscalingSpecPath(i, "name"), strings.Join(autoscalingSpec.Roles, ","),
-					"autoscaling is only supported for data tiers and ML nodes"),
-			)
-		}
-
 		if !(autoscalingSpec.NodeCountRange.Min >= 0) {
 			errs = append(
 				errs,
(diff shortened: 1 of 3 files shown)


pkg/controller/common/autoscaling/validations.go
```diff
@@ -149,6 +149,28 @@ func ValidateAutoscalingPolicies(
...
+					errs,
+					field.Invalid(
+						autoscalingSpecPath(i, "name"), strings.Join(autoscalingSpec.Roles, ","),
+						fmt.Sprintf("node role %s is invalid", role)),
+				)
+			}
+		}
+
+		// at least ml or a data role is present
+		if !CanBeAutoscaled(autoscalingSpec.Roles) {
```
barkbay, line 165: I think I initially wanted to limit this kind of validation rules since it binds the life cycle of the Elasticsearch deciders with the life cycle of the operator itself. Said differently, if we follow this path, we might have to introduce rules depending on Elasticsearch versions when new deciders are introduced to allow them to be used. Instead, I would expect Elasticsearch to return an error if an autoscaling policy does not contain any role that can autoscaled.

thbkrkr: I see the point. In the other side, we already have plenty of es-feature version gates in the operator, which is one thing among others that makes it powerful. Once in place, it is something stable. Also, there shouldn't be many new deciders in the future.
I don't have a strong opinion but it seems to me that it's worth it to the user versus the cost to us, which seems to me pretty low.

thbkrkr: I tested and the UX is not so bad. I will see all the same to update the doc to remove the ambiguity.

PR #6272, opened 2023-01-03
Install gke-gcloud-auth-plugin binary in ci Docker image

thbkrkr:
Required to support k8s v1.26 (https://cloud.google.com/blog/products/containers-kubernetes/kubectl-auth-changes-in-gke).
`USE_GKE_GCLOUD_AUTH_PLUGIN` is set to have kubectl use the new binary plugin for authentication instead of using the default provider-specific code because the deployer relies on kubectl.

Relates to #6237.


naemono (approved):
lgtm

PR #6278, opened 2023-01-05
Try to reconcile license even in absence of known health status

### LINKED ISSUE #6274: Cluster license cannot be updated by ECK if it has already expired
Related https://github.com/elastic/cloud-on-k8s/pull/5796

With the changes from above mentioned PR it is no longer possible to update the cluster license if the currently installed license has already expired because the reachability criterion relies on a successful health observation:

https://github.com/elastic/cloud-on-k8s/blob/2.5/pkg/controller/elasticsearch/driver/driver.go#L243

We need to distinguish the two criteria (service has endpoints and health API responds).

However it seems with expired license health observations are no longer possible as the health API is blocked in the degraded state Elasticsearch enters into when the license expires.

One potential fix for this issue is therefore only apply the first criteria (service has endpoints) when deciding whether to try the license API call or not. 

Workarounds:
* manually install the license contained in the Kubernetes secret called `$CLUSTER_NAME-es-license` via `curl` https://www.elastic.co/guide/en/elasticsearch/reference/current/update-license.html (that's all the operator does anyway)
* revert to a Basic license which will re-enable the operator to apply the correct license https://www.elastic.co/guide/en/elasticsearch/reference/current/start-basic.html (CAVEAT: I don't know if that has negative non-reversible effects on features in use)

pebrc:
Fixes #6274

Change the logic so that license reconciliation is attempted as soon as we have endpoints in the K8s service. But only log an error if we think based on the observed cluster health that the license API call should succeed. 

* in case of an expired license the observed health will be 403, and we will try to reconcile the license
* attempts will be logged by the es client in debug mode
* if the cluster should respond to requests  based on the health status additional logging on errors will happen and a Kubernetes event will be generated. This restriction is motivated by the fact the users get confused by messages like "could not reconcile license" when in fact the underlying issue is that their cluster is down. So this preserves the functionality that https://github.com/elastic/cloud-on-k8s/pull/5796 introduced.

### DIFF
--- a/pkg/controller/elasticsearch/driver/driver.go
+++ b/pkg/controller/elasticsearch/driver/driver.go
@@ -279,13 +279,18 @@ func (d *defaultDriver) Reconcile(ctx context.Context) *reconciler.Results {
 	// controllers that may be waiting for the orchestration hint.
 	results.WithError(d.maybeSetServiceAccountsOrchestrationHint(ctx, esReachable, esClient, resourcesState))
 
-	// reconcile the Elasticsearch license
-	if esReachable {
+	// reconcile the Elasticsearch license (even if we assume the cluster might not respond to requests to cover the case of
+	// expired licenses where all health API responses are 403)
+	if isServiceReady {
 		err = license.Reconcile(ctx, d.Client, d.ES, esClient, currentLicense)
 		if err != nil {
 			msg := "Could not reconcile cluster license, re-queuing"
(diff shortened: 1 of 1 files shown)


thbkrkr (approved):
LGTM

I kind of tested it by tweaking the operator (observer health / license reconciliation error).

pebrc:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|2.6|[](https://github.com/elastic/cloud-on-k8s/pull/6290)|

### Questions ?

PR #6281, opened 2023-01-05
OperatorHub Release CLI / Moving OperatorHub release to Buildkite

naemono:
This adds a Buildkite pipeline for release operations pertaining to OperatorHub/Red Hat.
This also augments our existing OpertaorHub CLI and adds new commands/sub-commands.

~~What's been manually tested:~~

~~Full set of operatorhub release operations with a `BC` build of Operator.~~

~~In progress:~~

- [x] ~~Currently testing Buildkite integration (will fully test during actual release).~~

^^ All fully tested during release process for 2.6.x.


.buildkite/pipeline.yml
```diff
@@ -107,3 +107,48 @@ steps:
...
+        commands:
+           - cd hack/operatorhub
+           - make build
+           - ./operatorhub container push --dry-run=false
+
+      - label: "run container preflight"
+        key: "redhat-preflight"
+        if: build.message == "run operatorhub release"
+        agents:
+          image: "quay.io/opdev/preflight:1.2.1-linux-amd64@sha256:6f63033a934d9a9821cea2247858b0557f471b1f0fc9c4f691ae1c934d44d90b"
```
naemono, line 135: I wonder here if pulling the preflight binary while running with our default agent image (or adding the preflight binary to our agent) would be a better alternative than what's being done below.  Would love thoughts about this approach.

hack/operatorhub/internal/operatorhub/operatorhub.go
```diff
@@ -0,0 +1,647 @@
...
+				closer.Close()
+			}
+		}
+	}
+	for _, p := range manifestPaths {
+		r, err := os.Open(p)
+		if err != nil {
+			return nil, closer, fmt.Errorf("while opening %s: %w", manifestPaths, err)
+		}
+		rs = append(rs, r)
```
thbkrkr, line 273: The fix to ensure local YAML files have a proper "end of directives" marker (#5447) is missing:
```suggestion
		rs = append(rs, r)
		// if we're using local yaml files, ensure that they have a proper
		// end of directives marker between them.
		rs = append(rs, strings.NewReader(yamlSeparator))
```

.buildkite/pipeline.yml
```diff
@@ -107,3 +107,70 @@ steps:
...
+
+    steps:
+
+      - label: "build operatorhub tool"
+        key: "build-operatorhub-tool"
+        if: build.message == "run operatorhub release"
+        commands:
+           - cd hack/operatorhub
+           - make build
+           - buildkite-agent artifact upload "bin/operator*"
```
pebrc, line 121: Would it make sense to use a custom image for this to not have to build the tooling every time?

.buildkite/pipeline.yml
```diff
@@ -107,3 +107,70 @@ steps:
...
+           - chmod u+x /usr/local/bin/operatorhub
+           - /usr/local/bin/operatorhub container push --dry-run=false
+
+      - label: "run container preflight"
+        key: "redhat-preflight"
+        if: build.message == "run operatorhub release"
+        depends_on:
+          - "redhat-container-push"
+        commands:
+           - curl -o /tmp/jq https://stedolan.github.io/jq/download/linux64/jq && chmod +x /tmp/jq
```
pebrc, line 150: A custom image could also contain `jq`

naemono: Agreed.  I've opened a PR to add jq to our custom image we already use @pebrc

.buildkite/pipeline.yml
naemono: @pebrc the build process takes ~60 seconds in BK.  If we add this binary to the image we're already using, then the process to fix any new Red Hat changes in the release process (which are pretty common) would involve:

1. opening a PR in this repository, waiting for approval/merge of the tooling.
2. opening a PR in the image repository to bump the tooling version, waiting for approval/merge.
3. Re-running the bk job.

As opposed to:
1. opening a PR in this repository with the tooling changes
2. running the bk job, pointing to the open PR for the tooling changes.  (note the `--buildkite-pr-repository` and `--buildkite-pr-id` flags in the `operatorhub buildkite` command which exist for this scenario)

hack/operatorhub/internal/operatorhub/operatorhub.go
```diff
@@ -0,0 +1,650 @@
...
+	resp, err := client.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("while executing HTTP GET %s: %w", url, err)
+	}
+
+	defer func() {
+		if resp.Body != nil {
+			resp.Body.Close()
+		}
+	}()
```
thbkrkr, line 319: https://github.com/elastic/cloud-on-k8s/commit/1168b13c6606034d1d1f43fcce207918df2d2ce0

```suggestion
	defer resp.Body.Close()
```

hack/operatorhub/README.md
```diff
@@ -1,23 +1,224 @@
...
 ```shell
-go run main.go --conf=config.yaml 
+./bin/operatorhub generate-manifests -c config.yaml -t '2.5.0'
 ```
 
 To generate configuration based on yet unreleased YAML manifests:
 
 ```shell
-go run main.go --conf=config.yaml --yaml-manifest=../../config/crds.yaml --yaml-manifest=../../config/operator.yaml
+./bin/operatorhub generate-manifests -c config.yaml -y ../../config/crds.yaml -y ./../config/operator.yaml -s '8.6.0' -V '2.5.0' -t '2.6.0'
```
thbkrkr, line 127: Why asked `-s '8.6.0' -V '2.5.0' -t '2.6.0'`when it's already defined in `-c config.yaml`?

.buildkite/pipeline.yml
```diff
@@ -107,3 +107,69 @@ steps:
...
+           - "echo curl -s -G \"https://catalog.redhat.com/api/containers/v1/projects/certification/id/$$OHUB_PROJECT_ID/images?filter=repositories.tags.name==$$OHUB_TAG\" -H \"X-API-KEY: $$OHUB_API_KEY\" > /tmp/redhat.json"
+           - "curl -s -G \"https://catalog.redhat.com/api/containers/v1/projects/certification/id/$$OHUB_PROJECT_ID/images?filter=repositories.tags.name==$$OHUB_TAG\" -H \"X-API-KEY: $$OHUB_API_KEY\" > /tmp/redhat.json"
+           - cat /tmp/redhat.json
+           # The following command will fail if no data was returned, so preflight will not run if it's already run
+           - jq -e '.data[0]' /tmp/redhat.json && exit 0
+           - curl -L -o /tmp/preflight https://github.com/redhat-openshift-ecosystem/openshift-preflight/releases/download/1.2.1/preflight-linux-amd64
+           - chmod u+x /tmp/preflight
+           - vault read -format=json -field=data secret/ci/elastic-cloud-on-k8s/operatorhub-release-preflight > /tmp/auth.json
+           - /tmp/preflight check container quay.io/redhat-isv-containers/$$OHUB_PROJECT_ID:$$OHUB_TAG --pyxis-api-token=$$OHUB_API_KEY --certification-project-id=$$OHUB_PROJECT_ID --submit -d /tmp/auth.json
```
thbkrkr, line 160: When there are so many commands, I think it will be more appropriate to write a bash script (`.buildkite/scripts/release/operatorhub-preflight.sh`). This would avoid to think about the $$ for example.

Did you consider to move this in the tool? I feel like `/usr/local/bin/operatorhub container preflight` would have been nice. I'm good with this version too.

.buildkite/pipeline.yml
```diff
@@ -107,3 +107,69 @@ steps:
         depends_on:
           - "operator-image-build"
           - "e2e-tests-image-build"
+
+  - group: redhat-release
+
+    steps:
+
```
thbkrkr, line 114: I would move the steps in a new pipeline `.buildkite/pipeline-release-operatorhub.yml` that we invoke from the root pipeline `.buildkite/pipeline.yml`.

```suggestion

      - label: "redhat operatorhub"
        if: build.message == "run operatorhub release"
        commands:
            - buildkite-agent pipeline upload .buildkite/pipeline-release-operatorhub.yml
```

hack/operatorhub/internal/container/container.go
```diff
@@ -0,0 +1,355 @@
...
+		return nil
+	}
+	if c.Force {
+		log.Println("pushing image as force was set")
+	} else {
+		log.Println("✓")
+	}
+	if err = pushImageToRegistry(c); err != nil {
+		return fmt.Errorf("failed to push image: %w", err)
+	}
```
thbkrkr, line 92:
```suggestion
	}

	if err = pushImageToRegistry(c); err != nil {
		log.Println("x")	
		return fmt.Errorf("failed to push image: %w", err)
	}
	log.Println("✓")	
```

hack/operatorhub/README.md
naemono: yeah, I will get this simplified.  good point.  Unnecessary.

.buildkite/pipeline.yml
naemono: I'll get this contained within a script.  👍 

> Did you consider to move this in the tool? I feel like `/usr/local/bin/operatorhub container preflight` would have been nice. I'm good with this version too.

Oh yes I tried to get this fully into this tool, see https://github.com/redhat-openshift-ecosystem/openshift-preflight/issues/845

.buildkite/pipeline-release-operatorhub.yml
```diff
@@ -0,0 +1,56 @@
...
+
+steps:
+
+  - group: redhat-release
+
+    steps:
+
+      - label: "build operatorhub tool"
+        key: "build-operatorhub-tool"
+        if: build.message == "run operatorhub release"
```
thbkrkr, line 15: Since `pipeline-release-operatorhub.yml` is uploaded only `if: build.message == "run operatorhub release"` dans `pipeline.yml`, it seems redundant to repeat the conditional for all steps.

hack/operatorhub/README.md
```diff
@@ -1,23 +1,227 @@
...
 ```
 
 To generate configuration based on yet unreleased YAML manifests:
 
 ```shell
-go run main.go --conf=config.yaml --yaml-manifest=../../config/crds.yaml --yaml-manifest=../../config/operator.yaml
+# If using newVersion|prevVersion|stackVersion variables in config.yaml
+./bin/operatorhub generate-manifests -c config.yaml -y ../../config/crds.yaml -y ./../config/operator.yaml
+# If wanting to override newVersion|prevVersion|stackVersion variables in config.yaml
+./bin/operatorhub generate-manifests -c config.yaml -y ../../config/crds.yaml -y ./../config/operator.yaml -s '8.6.0' -V '2.5.0' -t '2.6.0'
```
thbkrkr, line 130: What is the use case that would require overriding the values defined in `config.yaml`?

.buildkite/scripts/operatorhub/operations.sh
```diff
@@ -0,0 +1,82 @@
...
+    fi
+
+    buildkite-agent artifact download "bin/operator*" /usr/local/
+    buildkite-agent artifact download "config/*.yaml" .
+    chmod u+x /usr/local/bin/operatorhub
+    /usr/local/bin/operatorhub container publish --dry-run=false
+    cd hack/operatorhub
+    /usr/local/bin/operatorhub generate-manifests --prev-version="$OHUB_PREV_VERSION" --stack-version="$OHUB_STACK_VERSION" --yaml-manifest=../../config/crds.yaml --yaml-manifest=../../config/operator.yaml
+    /usr/local/bin/operatorhub bundle generate --dir="$(pwd)"
+    /usr/local/bin/operatorhub bundle create-pr --dir="$(pwd)"
```
thbkrkr, line 54: I would have kept these release commands in the pipeline and just use this script for the preflight commands.

hack/operatorhub/cmd/buildkite/buildkite.go
```diff
@@ -0,0 +1,207 @@
...
+const (
+	url = "https://api.buildkite.com/v2/organizations/elastic/pipelines/cloud-on-k8s-operator/builds"
+)
+
+// Command will return the buildkite command
+func Command() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:          "buildkite",
+		Short:        "Start operatorhub release operation within Buildkite",
+		Long:         "Start operatorhub release operation within Buildkite.",
```
thbkrkr, line 31: I think
```bash
curl -XPOST "https://api.buildkite.com/v2/organizations/elastic/pipelines/cloud-on-k8s-operator-release-operatorhub/builds" \
-H "Authorization: Bearer $BK_TOKEN" \
-d '{
    "commit": "2.6.1",
    "branch": "2.6"
}'
```
could have been enough.

thbkrkr (commented):
Maybe I miss something but as discussed in https://github.com/elastic/cloud-on-k8s/pull/6281#discussion_r1080575799, I think we shouldn't duplicate what is defined in `./config.yaml` (`OHUB_CONF`). This configuration file is the source of truth that lives in the codebase and is updated only through pull requests.
This means that all the following flags / env vars should disappear:
* OHUB_SUPPORTED_OPENSHIFT_VERSIONS
* OHUB_STACK_VERSION
* OHUB_PREV_VERSION
* OHUB_TAG
* OHUB_BRANCH

It seems to me that having to adjust them is painful and above all prone to error.

hack/operatorhub/templates/package.tpl
```diff
@@ -1,5 +1,5 @@
 channels:
-- currentCSV: {{ .PackageName }}.v{{ .NewVersion }}
+- currentCSV: {{ .PackageName }}.{{ .NewVersion }}
```
thbkrkr, line 2: I think we need to keep the `v`.
```suggestion
- currentCSV: {{ .PackageName }}.v{{ .NewVersion }}
```

Otherwise, I get:

```
> OHUB_TAG=2.6.1 hack/operatorhub/bin/operatorhub bundle generate --dir=.
2023/02/10 15:19:38 Generating operator bundle image for publishing
2023/02/10 15:19:38 ⅹ
Error: during opm generate: error interpreting channels, please manually input channels instead
2023/02/10 15:19:38 failed to run command: during opm generate: error interpreting channels, please manually input channels instead
```

Because we use the `v`prefix to name the CSV:
```
# templates/csv.tpl
apiVersion: operators.coreos.com/v1alpha1
kind: ClusterServiceVersion
metadata:
  name: {{ .PackageName }}.v{{ .NewVersion }}
```

.buildkite/pipeline-release-operatorhub.yml
```diff
@@ -0,0 +1,56 @@
...
+        commands:
+          - cd hack/operatorhub
+          - make build
+          - buildkite-agent artifact upload "bin/operator*"
```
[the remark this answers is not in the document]
naemono: I swore I tried this, and it wasn't properly working, but I'll certainly try again, as it makes total sense.

.buildkite/scripts/operatorhub/operations.sh
```diff
@@ -0,0 +1,82 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
naemono: License header also added to `.buildkite/scripts/common/get-test-artifacts.sh`.

hack/operatorhub/README.md
naemono: I think I was wanting to ensure that the settings were absolutely correct when running in an automated fashion, but since the typical use case for this is to run this from a branch that already has `config.yaml` updated appropriately, then we can likely do without these, agreed.   They've been removed.

.buildkite/scripts/operatorhub/operations.sh
naemono: I figured since I was cleaning up the pipeline, I might as well keep it as clean as possible, hence moving all of this to a script.  Lmk if you need me to move it back.

hack/operatorhub/cmd/buildkite/buildkite.go
naemono: We discussed this privately, and decided that opening #6408 would be beneficial.

naemono:
> Maybe I miss something but as discussed in [#6281 (comment)](https://github.com/elastic/cloud-on-k8s/pull/6281#discussion_r1080575799), I think we shouldn't duplicate what is defined in `./config.yaml` (`OHUB_CONF`). This configuration file is the source of truth that lives in the codebase and is updated only through pull requests. This means that all the following flags / env vars should disappear:
> 
> * OHUB_SUPPORTED_OPENSHIFT_VERSIONS
> * OHUB_STACK_VERSION
> * OHUB_PREV_VERSION
> * OHUB_TAG
> * OHUB_BRANCH
> 
> It seems to me that having to adjust them is painful and above all prone to error.

@thbkrkr I've removed prev/stack, but removing the others is a much bigger change as, for example, tag is used throughout the tool.  I'll get these removed/updated soon.

thbkrkr:
> Unfortunately OHUB_SUPPORTED_OPENSHIFT_VERSIONS is required, as there's no setting for it in the configuration file. If you want, I can certainly add that as a new option there and remove the cli flag. LMK.

I think it's good to have all the configuration in the source files, so I feel like saying yes please.
Should it be more something like `minSupportedOpenShiftVersion` as we don't want the range here?

We already have the range in the template that we could half-generate from it or accept to have the info duplicated (we already have the minKubeVersion duplicated in the template).
https://github.com/elastic/cloud-on-k8s/blob/d8d38f5130725c2872c7aba6241ead5dc6ffba94/hack/operatorhub/templates/csv.tpl#L270

naemono:
> We already have the range in the template that we could half-generate from it or accept to have the info duplicated (we already have the minKubeVersion duplicated in the template).

I've kept it simple, and just added it to the configuration file, and updated the validation to ensure that it exists, and is not a range.

thbkrkr (approved):
LGTM!


The vault bits could benefit from #6418 but we shouldn't block for that.

PR #6298, opened 2023-01-10
Changes required to prefix versions with 'v'.

### LINKED ISSUE #6134: Prefix releases/tags with `v`.
See https://github.com/elastic/cloud-on-k8s/pull/5809#issuecomment-1298877962

Even after moving to a `/v2` go module path in `go.mod`, we are still being imported using a legacy format:

```
❯ go get github.com/elastic/cloud-on-k8s/v2@2.4.0
go: added github.com/elastic/cloud-on-k8s/v2 v2.0.0-20220812092235-96282ca98df1
```

We need to adjust our release process to ensure we are prefixing releases/tags with `v`, such as `v2.6.0`.

Things to consider:

- [ ] Redhat release process changes needed.
- [ ] Documentation noting the change, as not to confuse someone wanting to install a previous version, only to find it's not prefixed with `v`.
- [ ] ECK Helm chart changes.
- [ ] Tooling changes needed under `hack` directory? (release-notes, upgrade-test-harness, update-stack-version.sh)
- [ ] CI changes?
- [ ] Public facing documentation changes (kubectl install links; release notes links)

naemono:
closes #6134 

These are the changes that I am aware of that would need to happen to prefix versions with `v`.

I'm unsure this is something we want to do this `2.6` release, but these are here for future release(s) when we want to pursue this.

Other changes needed
- [ ] ECK Download page Document needs updated to update the `kubectl create -f https://download.elastic.co/downloads/eck/${VERSION}/crds.yaml` links.

### DIFF
--- a/.ci/setenvconfig
+++ b/.ci/setenvconfig
@@ -667,7 +667,7 @@ GO_TAGS=release
 export LICENSE_PUBKEY=${PROJECT_PATH}/.ci/license.key
 
 SNAPSHOT=true
-IMG_VERSION=v$(cat $WORKSPACE/VERSION)-$(date +%F)-$(git rev-parse --short --verify HEAD)
+IMG_VERSION=$(cat $WORKSPACE/VERSION)-$(date +%F)-$(git rev-parse --short --verify HEAD)
 ENV
 
   else
@@ -686,11 +686,7 @@ GO_TAGS=release
 export LICENSE_PUBKEY=${PROJECT_PATH}/.ci/license.key
 
 SNAPSHOT=false
-IMG_VERSION=$TAG_NAME
-# If IMG_VERSION doesn't begin with a 'v', add it.
-if [[ ! "${IMG_VERSION}" =~ ^v.* ]]; then
-    IMG_VERSION = "v${IMG_VERSION}"
-fi
+IMG_VERSION=${TAG_NAME#"v"}
 ENV
 
   fi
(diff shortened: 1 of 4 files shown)


thbkrkr:
The documentation would also be impacted by this change (https://github.com/elastic/cloud-on-k8s/pull/6284#discussion_r1064499311).
What do you think about my proposal to minimize changes (https://github.com/elastic/cloud-on-k8s/issues/6134#issuecomment-1377052219)?

pebrc:
Should we get the necessary changes for this in with enough time to spare before the next release so that we can convince ourselves that everything works (release automation etc)?

naemono:
Yes @pebrc I'll focus on this this upcoming week.

naemono:
Wasn't able to get to this last week, but will work on this first thing this week to get these changes moving forward.

naemono:
I've updated this PR with the discussed changes.

> The documentation would also be impacted by this change ([#6284 (comment)](https://github.com/elastic/cloud-on-k8s/pull/6284#discussion_r1064499311)).

@thbkrkr I don't think these documentation changes are needed if we're not going to update the `eck-attributes.asciidoc ` file with this `v` prefix:
```
:eck_version: v2.6.0
```

Right?

Also, I went through the [operatorhub changes](https://github.com/elastic/cloud-on-k8s/pull/6281), and they will work as is, as we actually use the `v` prefix already for the `name` field, but ensure it's stripped for the `version` field in the rendered CSV:

```
# template
name: {{ .PackageName }}.v{{ .NewVersion }}
# rendered
name: elastic-cloud-eck.v2.6.0

...
# template
version: {{ .NewVersion }}
# rendered
version: 2.6.0
```

thbkrkr (approved):
LGTM

thbkrkr:
> Right?
Yes, it was not clear but it was a OR between my two sentences not a AND :)

PR #6302, opened 2023-01-10
Do not overwrite deployment status when reconciling

pebrc:
Updating the deployment status should only be done by the deployment controller and not by the ECK operator.

### DIFF
--- a/pkg/controller/common/deployment/reconcile.go
+++ b/pkg/controller/common/deployment/reconcile.go
@@ -15,6 +15,7 @@ import (
 	"github.com/elastic/cloud-on-k8s/v2/pkg/controller/common/hash"
 	"github.com/elastic/cloud-on-k8s/v2/pkg/controller/common/reconciler"
 	"github.com/elastic/cloud-on-k8s/v2/pkg/utils/k8s"
+	"github.com/elastic/cloud-on-k8s/v2/pkg/utils/maps"
 )
 
 // Params to specify a Deployment specification.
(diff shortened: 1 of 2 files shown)


pkg/controller/common/deployment/reconcile.go
```diff
@@ -71,7 +72,12 @@ func Reconcile(
 			return hash.GetTemplateHashLabel(reconciled.Labels) != hash.GetTemplateHashLabel(expected.Labels)
 		},
 		UpdateReconciled: func() {
-			expected.DeepCopyInto(reconciled)
+			// set expected annotations and labels, but don't remove existing ones
```
barkbay, line 75: Maybe it's something I realised a while ago and forgot, but this change is not strictly required since the reconciler never updates the status: `params.Client.Update(params.Context, params.Reconciled)` [here](https://github.com/elastic/cloud-on-k8s/blob/8beeaec046abb16bb337bdd316d6174ddaf70f77/pkg/controller/common/reconciler/reconciler.go#L175) leaves the status untouched. I noticed it when I reverted the change to `expected.DeepCopyInto(reconciled)` and realised that the unit test still succeeded.

barkbay (commented):
LGTM

barkbay (approved):

pkg/controller/common/deployment/reconcile.go
pebrc: That's because my unit test was flawed 😞 I compared with a version of the Deployment that had not really been updated because the template hash was the same. With an additional assertion it fails as expected.

PR #6326, opened 2023-01-16
Prune EOL ECK versions in upgrade-test-harness

thbkrkr:
Prune EOL ECK versions in upgrade-test-harness lower than the oldest supported version (currently `1.6.x`, see https://www.elastic.co/fr/support/eol).


pebrc (commented):
We should also delete the `alpha` specific code in the test runner if we drop the upgrade from older versions (which I agree is reasonable to do)

pebrc (approved):

PR #6329, opened 2023-01-16
Fix test helm charts script exit code

thbkrkr:
This commit fixes the script to test Helm charts by enabling `set -eu` to exit at the first error, otherwise the script passes regardless of the status of the tests.

### DIFF
--- a/hack/helm/test.sh
+++ b/hack/helm/test.sh
@@ -1,16 +1,21 @@
 #!/usr/bin/env bash
-
+#
 # Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
 # or more contributor license agreements. Licensed under the Elastic License 2.0;
 # you may not use this file except in compliance with the Elastic License 2.0.
-
+#
 # Script to test helm charts
+#
+set -eu
 
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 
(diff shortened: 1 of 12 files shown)


hack/helm/test.sh
```diff
@@ -29,11 +32,14 @@ check() {
     helm lint --strict .
```
naemono, line 32: Thanks for catching this.  Would we want to also fail if linting fails?

```suggestion
    helm lint --strict . || STATUS=1
```

naemono (approved):

hack/helm/test.sh
thbkrkr: Yes indeed. Looking back, I wonder why we don't just `set -e` to fail on the first error?

naemono: 💯 .  We should likely just do that, and ` || true` anything that isn't fatal.

thbkrkr: Was the original intention precisely not to stop at the first error and reveal them all in a single execution? I'm not sure it's worth it because the script runs very fast.

naemono:
> Was the original intention precisely not to stop at the first error and reveal them all in a single execution? I'm not sure it's worth it because the script runs very fast.

Thanks @thbkrkr .  I was going to say that we probably should unittest all helm charts even if the previous failed, but I don't think it's critical here, as long as the whole script fails if one fails, which you've done here.

PR #6350, opened 2023-01-23
Do not test 8.5.3 due to APM issue

pebrc:
Out e2e test exposed https://github.com/elastic/apm-server/issues/10089 

As a quick fix to silence our test pipeline I suggest to go back to 8.5.2 until a new (fixed) patch version of 8.5 is available. Alternatively we can just skip the APM tests for 8.5.3. (but I don't have time for that right now, can PR that tomorrow)


naemono (approved):
I see no reason not to do this.  👍

PR #6352, opened 2023-01-24
Fix Stackmon Sidecar Secret having incorrect labels.

### LINKED ISSUE #5967: NewMetricBeatSidecar in common stackmon package defaults to label type Elasticsearch
https://github.com/elastic/cloud-on-k8s/blob/3d7bd7760e7801508ca0d71c88aa15bacd8c68fa/pkg/controller/common/stackmon/config.go#L91

Regardless of the type of resource sent to `NewMetricBeatSidecar`, the following labels are always set:

```
Labels: map[string]string{
	"common.k8s.elastic.co/type":                "elasticsearch",
	"elasticsearch.k8s.elastic.co/cluster-name": "associated_cluster",
}
```

It seems as though this should be set according to the `resource` argument, not hard-coded to `elasticsearch`.  Since both Kibana, and now Beat stack monitoring use this, it should be updated.

naemono:
Resolves #5967 

Currently the Secret for the stack monitoring sidecar *always* has the labels:

```
Labels: map[string]string{
	"common.k8s.elastic.co/type":                "elasticsearch",
	"elasticsearch.k8s.elastic.co/cluster-name": "associated_cluster_name",
}
```

This fixes the labels to be associated with the actual resource that is being monitored.

To do this in a manner to avoid a Switch/Case statement for each type that can have stack monitoring enabled, I instead created a new interface

```
// HasElasticLabels allows a return of Elastic assigned labels for any object.
type HasElasticLabels interface {
	client.Object
	GetElasticLabels() map[string]string
}
```

Such that any Elastic CRD object now can call `GetElasticLabels` and have the appropriate 2 labels returned

```
	"common.k8s.elastic.co/type":                "correct_type",
	"elasticsearch.k8s.elastic.co/(name|cluster-name)": "correct_name",
```

Why not `GetLabels` instead of `GetElasticLabels`?
1. `GetLabels` already exists on the `client.Object` interface.
2. `GetElasticLabels` indicates that we're returning the Elastic specific labels in use by our operator.

To make this consistent across all CRDs controllers, I've removed the `NewLabels`, or `Labels` func within each controller, and moved to using the new method on the CRD.

Also, `TypeLabelName` was moved from `pkg/controller/common/labels` => `pkg/apis/common/v1` to avoid an import cycle.

### DIFF
--- a/pkg/apis/agent/v1alpha1/labels.go
+++ b/pkg/apis/agent/v1alpha1/labels.go
@@ -1,3 +1,7 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License 2.0;
+// you may not use this file except in compliance with the Elastic License 2.0.
+
 package v1alpha1
 
 import (
--- a/pkg/apis/agent/v1alpha1/labels.go
+++ b/pkg/apis/agent/v1alpha1/labels.go
@@ -1,0 +1,13 @@
+package v1alpha1
+
+import (
+	commonv1 "github.com/elastic/cloud-on-k8s/v2/pkg/apis/common/v1"
+)
+
+// GetElasticLabels will return the common Elastic assigned labels for the Elastic Agent.
+func (a *Agent) GetElasticLabels() map[string]string {
+	return map[string]string{
+		commonv1.TypeLabelName:      "agent",
+		"agent.k8s.elastic.co/name": a.Name,
+	}
+}
--- a/pkg/apis/apm/v1/labels.go
+++ b/pkg/apis/apm/v1/labels.go
@@ -1,0 +1,13 @@
+package v1
+
+import (
+	commonv1 "github.com/elastic/cloud-on-k8s/v2/pkg/apis/common/v1"
+)
+
+// GetElasticLabels will return the common Elastic assigned labels for the APM Server.
+func (as *ApmServer) GetElasticLabels() map[string]string {
+	return map[string]string{
+		commonv1.TypeLabelName:    "apm-server",
+		"apm.k8s.elastic.co/name": as.Name,
+	}
+}
--- a/pkg/apis/apm/v1/labels.go
+++ b/pkg/apis/apm/v1/labels.go
@@ -1,3 +1,7 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License 2.0;
+// you may not use this file except in compliance with the Elastic License 2.0.
+
 package v1
 
 import (
--- a/pkg/apis/beat/v1beta1/labels.go
+++ b/pkg/apis/beat/v1beta1/labels.go
@@ -1,3 +1,7 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License 2.0;
+// you may not use this file except in compliance with the Elastic License 2.0.
+
 package v1beta1
 
 import (
--- a/pkg/apis/beat/v1beta1/labels.go
+++ b/pkg/apis/beat/v1beta1/labels.go
@@ -1,0 +1,13 @@
+package v1beta1
+
+import (
+	commonv1 "github.com/elastic/cloud-on-k8s/v2/pkg/apis/common/v1"
+)
+
+// GetElasticLabels will return the common Elastic assigned labels for the Beat.
+func (b *Beat) GetElasticLabels() map[string]string {
+	return map[string]string{
+		commonv1.TypeLabelName:     "beat",
+		"beat.k8s.elastic.co/name": b.Name,
+	}
+}
--- a/pkg/apis/common/v1/common.go
+++ b/pkg/apis/common/v1/common.go
@@ -353,6 +353,16 @@ type HasObservedGeneration interface {
 	GetObservedGeneration() int64
 }
 
+// TypeLabelName is used to represent a resource type in k8s resources
+const TypeLabelName = "common.k8s.elastic.co/type"
+
+// HasElasticLabels allows a return of Elastic assigned labels for any object.
+// +kubebuilder:object:generate=false
+type HasElasticLabels interface {
+	client.Object
+	GetElasticLabels() map[string]string
+}
+
 // DisableDowngradeValidationAnnotation allows circumventing downgrade/upgrade checks.
 const DisableDowngradeValidationAnnotation = "eck.k8s.elastic.co/disable-downgrade-validation"
 
--- a/pkg/apis/elasticsearch/v1/labels.go
+++ b/pkg/apis/elasticsearch/v1/labels.go
@@ -1,3 +1,7 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License 2.0;
+// you may not use this file except in compliance with the Elastic License 2.0.
+
 package v1
 
 import (
--- a/pkg/apis/elasticsearch/v1/labels.go
+++ b/pkg/apis/elasticsearch/v1/labels.go
@@ -1,0 +1,13 @@
+package v1
+
+import (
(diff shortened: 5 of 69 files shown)


pkg/controller/common/labels/labels.go
```diff
@@ -41,3 +39,8 @@ func (l TrueFalseLabel) AsMap(value bool) map[string]string {
 func AddCredentialsLabel(original map[string]string) map[string]string {
 	return maps.Merge(map[string]string{credentialsLabel: "true"}, original)
 }
+
+// New will return the common Elastic labels for any instance of the Elastic Stack.
+func New(obj commonv1.HasElasticLabels) (labels map[string]string) {
```
pebrc, line 44: I am not sure I see the point of this function (it is used only in one place where we could call the method from the interface directly as far as I can tell). It reads a tiny bit nicer because `labels.New` is pretty self-explanatory but maybe not worth it?

pkg/apis/common/v1/common.go
```diff
@@ -353,6 +353,16 @@ type HasObservedGeneration interface {
 	GetObservedGeneration() int64
 }
 
+// TypeLabelName is used to represent a resource type in k8s resources
+const TypeLabelName = "common.k8s.elastic.co/type"
+
+// HasElasticLabels allows a return of Elastic assigned labels for any object.
+// +kubebuilder:object:generate=false
+type HasElasticLabels interface {
```
pebrc, line 361: You mentioned it in the PR description already and naming is hard. I think `Elastic` as a name is maybe not ideal because everything we do is something `Elastic` I guess. I was trying to think of something more descriptive but all seemed pretty tortured like `TypeNameLabels` or incomplete like `TypeLabels`. One option I don't totally hate is `GetIdentityLabels` and `HasIdentityLabels` the rationale being that the name and type labels together identify the resource uniquely e.g. for label selectors and such. Wdyt?

pebrc (commented):
LGTM! The only thing is the name for the new interface maybe. I added a suggestion in the comments but maybe some else has a better idea too.

pkg/controller/common/labels/labels.go
naemono: Yeah, makes perfect sense.  Removed.  

*Note* I kept the test in this package, but renamed it to `TestGetIdentityLabels`, which tests all of the CRD's methods. It makes more sense from a package standpoint for this to go in `pkg/apis/common/v1`, but that would be an import cycle, so I left it here. The only other option would be to make 7 tests, one in each of the `pkg/apis/CRD/*` packages, which seemed overkill. Lmk your thoughts on that.

pkg/apis/common/v1/common.go
naemono: I can be fine with that change.  Implemented.

pkg/apis/agent/v1alpha1/labels.go
```diff
@@ -0,0 +1,17 @@
...
+// or more contributor license agreements. Licensed under the Elastic License 2.0;
+// you may not use this file except in compliance with the Elastic License 2.0.
+
+package v1alpha1
+
+import (
+	commonv1 "github.com/elastic/cloud-on-k8s/v2/pkg/apis/common/v1"
+)
+
+// GetElasticLabels will return the common Elastic assigned labels for the Elastic Agent.
```
thbkrkr, line 11:
```suggestion
// GetIdentityLabels will return the common Elastic assigned labels for the Elastic Agent.
```

pkg/apis/apm/v1/labels.go
```diff
@@ -0,0 +1,17 @@
...
+// or more contributor license agreements. Licensed under the Elastic License 2.0;
+// you may not use this file except in compliance with the Elastic License 2.0.
+
+package v1
+
+import (
+	commonv1 "github.com/elastic/cloud-on-k8s/v2/pkg/apis/common/v1"
+)
+
+// GetElasticLabels will return the common Elastic assigned labels for the APM Server.
```
thbkrkr, line 11:
```suggestion
// GetIdentityLabels will return the common Elastic assigned labels for the APM Server.
```

pkg/apis/beat/v1beta1/labels.go
```diff
@@ -0,0 +1,17 @@
...
+// or more contributor license agreements. Licensed under the Elastic License 2.0;
+// you may not use this file except in compliance with the Elastic License 2.0.
+
+package v1beta1
+
+import (
+	commonv1 "github.com/elastic/cloud-on-k8s/v2/pkg/apis/common/v1"
+)
+
+// GetElasticLabels will return the common Elastic assigned labels for the Beat.
```
thbkrkr, line 11:
```suggestion
// GetIdentityLabels will return the common Elastic assigned labels for the Beat.
```

pkg/apis/elasticsearch/v1/labels.go
```diff
@@ -0,0 +1,17 @@
...
+// or more contributor license agreements. Licensed under the Elastic License 2.0;
+// you may not use this file except in compliance with the Elastic License 2.0.
+
+package v1
+
+import (
+	commonv1 "github.com/elastic/cloud-on-k8s/v2/pkg/apis/common/v1"
+)
+
+// GetElasticLabels will return the common Elastic assigned labels for the Elasticsearch cluster.
```
thbkrkr, line 11:
```suggestion
// GetIdentityLabels will return the common Elastic assigned labels for the Elasticsearch cluster.
```

pkg/apis/enterprisesearch/v1/labels.go
```diff
@@ -0,0 +1,17 @@
...
+// or more contributor license agreements. Licensed under the Elastic License 2.0;
+// you may not use this file except in compliance with the Elastic License 2.0.
+
+package v1
+
+import (
+	commonv1 "github.com/elastic/cloud-on-k8s/v2/pkg/apis/common/v1"
+)
+
+// GetElasticLabels will return the common Elastic assigned labels for EnterpriseSearch.
```
thbkrkr, line 11:
```suggestion
// GetIdentityLabels will return the common Elastic assigned labels for the Enterprise Search.
```

docs/reference/api-docs.asciidoc
```diff
@@ -504,6 +504,8 @@ HTTPConfig holds the HTTP layer configuration for resources.
 
 
 
+
+
```
thbkrkr, line 508: Do you know why these new lines are added?

pkg/apis/maps/v1alpha1/labels.go
```diff
@@ -0,0 +1,17 @@
...
+// or more contributor license agreements. Licensed under the Elastic License 2.0;
+// you may not use this file except in compliance with the Elastic License 2.0.
+
+package v1alpha1
+
+import (
+	commonv1 "github.com/elastic/cloud-on-k8s/v2/pkg/apis/common/v1"
+)
+
+// GetElasticLabels will return the common Elastic assigned labels for EnterprisMapsServer.
```
thbkrkr, line 11:
```suggestion
// GetIdentityLabels will return the common Elastic assigned labels for the Elastic Maps Server.
```

pkg/apis/kibana/v1/labels.go
```diff
@@ -0,0 +1,17 @@
...
+// or more contributor license agreements. Licensed under the Elastic License 2.0;
+// you may not use this file except in compliance with the Elastic License 2.0.
+
+package v1
+
+import (
+	commonv1 "github.com/elastic/cloud-on-k8s/v2/pkg/apis/common/v1"
+)
+
+// GetElasticLabels will return the common Elastic assigned labels for Kibana.
```
thbkrkr, line 11:
```suggestion
// GetIdentityLabels will return the common Elastic assigned labels for the Kibana.
```

thbkrkr (commented):
Thanks for fixing this.
Just fix the name of `GetIdentityLabels` in the comments and this is good to merge.

docs/reference/api-docs.asciidoc
naemono: @thbkrkr 936cf510c specifically was to address this. Something is "requiring" those new spaces, and one of the checks (I think the "local changes" check) tagged this, so I specifically added those spaces to satisfy that. Lmk if you want me to dig further, or just accept it.

thbkrkr (approved):
LGTM

PR #6363, opened 2023-01-26
Don't failfast on e2e tests

naemono:
## Primary change this PR introduces

We currently do not continue on any failure of our e2e tests, so any potentially subsequent failures are never seen, and never investigated by our team, which can mask potentially critical failures, especially when testing new snapshot releases for an upcoming stack release. This change removes this `failfast` feature, and runs a full suite of e2e tests on each run regardless of failures.

### Changes to facilitate this removal of `failfast`

* ECK-diagnostics are run after *each* failure in our e2e tests.
* ECK-diagnostics binary is now built into the e2e container (it previously was only within the e2e test runner).
* ECK-Diagnostics output Zip file, named for the test failure, is uploaded to GS bucket, namespaced to Buildkite build number after each ECK-diagnotics run.
* If any e2e test fails, a final ECK-diagnostics is run by the e2e test runner container.
* All Zip files are downloaded from GS Bucket at the end of the test run by the e2e test runner.
* All Zip files are uploaded as BK artifacts.

#### Additional changes in this PR
* Kubernetes cluster build for e2e tests is run in parallel with container builds to speed up e2e test runs.
* ephemeralStorage: "20G" change is to allow the container builds to stop failing due to local storage limitations.


naemono:
run/e2e-tests

naemono:
buildkite test this

naemono:
To make the merge and use of this PR easier, this will await the merging of https://github.com/elastic/eck-diagnostics/pull/148.

test/e2e/test/step.go
```diff
@@ -32,23 +45,147 @@ func (l StepList) WithStep(testStep Step) StepList {
...
+	}
+	if !found {
+		env = append(env, "HOME=/tmp")
+	}
+	return env
+}
+
+// This simulates "kubectl delete elastic" in the e2e namespace.
+func deleteElasticResources() error {
+	cfg, err := rest.InClusterConfig()
```
pebrc, line 141: This won't work for local runs. `config.GetConfig()`

test/e2e/test/step.go
```diff
@@ -32,23 +45,147 @@ func (l StepList) WithStep(testStep Step) StepList {
...
+	clntset, err := kubernetes.NewForConfig(cfg)
+	if err != nil {
+		log.Error(err, "while getting clientset")
+		return err
+	}
+
+	type version string
+	groupVersionMap := map[string][]version{}
+
+	apiGroups, _, _ := clntset.Discovery().ServerGroupsAndResources()
```
pebrc, line 155: Error ignored. Also I think the second return  argument already contains all the information needed for the deletion and makes the `NewDiscoveryCategoryExpander` unnecessary. It is a list of all API resources for each API Group/Version each entry contains all resources and sub-resources for each version including the category.

test/e2e/test/context.go
```diff
@@ -64,7 +64,14 @@ func initializeContext() {
...
 
 	logutil.ChangeVerbosity(ctx.LogVerbosity)
+
+	// Do not print credentials in the following log output
+	backup := ctx.GCPCredentials
+	ctx.GCPCredentials = nil
+
 	log.Info("Test context initialized", "context", ctx)
+
+	ctx.GCPCredentials = backup
```
pebrc, line 74: This is very awkward and would go away if we re-used the existing `test-secrets` logic.

pebrc (commented):
I left a few comments. Mainly around integrating this with existing functionality and making sure that we can still do local e2e runs.

Makefile
```diff
@@ -573,8 +574,8 @@ ci-build-operator-e2e-run: setup-e2e build-operator-image e2e-run
 run-deployer: build-deployer
 	./hack/deployer/deployer execute --plans-file hack/deployer/config/plans.yml --config-file deployer-config.yml
 
-set-kubeconfig:
-	./hack/deployer/deployer get credentials --plans-file hack/deployer/config/plans.yml --config-file deployer-config.yml
+set-kubeconfig: build-deployer
+	./hack/deployer/deployer get credentials --authenticate --plans-file hack/deployer/config/plans.yml --config-file deployer-config.yml
```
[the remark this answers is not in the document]
naemono: I have added this, and additionally added a check that the existing account is within the correct Google project, as the default BK Google credentials exist, but are associated with another project, and fail.

.buildkite/scripts/common/run-main-e2e-tests.sh
```diff
@@ -0,0 +1,21 @@
...
+    # On Copy any artifacts from the test run locally to be picked up s buildkite artifacts
+    if gsutil ls "gs://eck-e2e-buildkite-artifacts/jobs/$BUILDKITE_PIPELINE_NAME/$BUILDKITE_BUILD_NUMBER/eck-diagnostic*.zip" ; then
+        gsutil cp "gs://eck-e2e-buildkite-artifacts/jobs/$BUILDKITE_PIPELINE_NAME/$BUILDKITE_BUILD_NUMBER/eck-diagnostic*.zip" .
+    fi
+}
+
+main() {
+    trap 'onError' ERR
+    # Read gcp bucket credentials to allow eck-diagnostics output writing to eck-e2e-buildkite-artifacts bucket.
+    vault read -field=service-account secret/ci/elastic-cloud-on-k8s/ci-gcp-k8s-operator > /tmp/auth.json
```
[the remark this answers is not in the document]
naemono: This makes sense. I have moved all of this to the test secret.

test/e2e/test/step.go
```diff
@@ -32,23 +45,147 @@ func (l StepList) WithStep(testStep Step) StepList {
...
+	// The following appends the test, and it's sub-test names together with a '-'.
+	// Example: For TestAutoscalingLegacy/Secrets_should_eventually_be_created:
+	// testName: TestAutoscalingLegacy, step.Name: Secrets_should_eventually_be_created
+	fullTestName := fmt.Sprintf("%s-%s", testName, step.Name)
+	// Convert any spaces to "_", and "/" to "-" in the test name.
+	normalizedTestName := strings.ReplaceAll(strings.ReplaceAll(fullTestName, " ", "_"), "/", "-")
+	cmd := exec.Command("eck-diagnostics", "--output-directory", "/tmp", "-n", fmt.Sprintf("eck-diagnostic-%s.zip", normalizedTestName), "-o", ctx.Operator.Namespace, "-r", strings.Join(otherNS, ","), "--run-agent-diagnostics") //nolint:gosec
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	cmd.Env = ensureTmpHomeEnv(cmd.Environ())
```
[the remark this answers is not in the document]
naemono: I have added a comment as to why this is needed. I also added a helper function. If you'd prefer me try to use the `command` package, I'll certainly give it a shot. Lmk.

test/e2e/test/step.go
naemono: Great point. I've adjusted this.

test/e2e/test/step.go
naemono: I'll work through testing this and update with more information.

naemono:
@pebrc all requested changes have been addressed and tested. This is ready for another set of 👀 when you can.

.buildkite/scripts/common/run-main-e2e-tests.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+# On any command failures, make sure that diagnostics are copied from bucket to local Buildkite environment
+# to be uploaded as artifacts.
+onError() {
+    # On Copy any artifacts from the test run locally to be picked up s buildkite artifacts
```
barkbay, line 8:
```suggestion
    # On error, copy any artifacts from the test run locally to be picked up as buildkite artifacts
```

hack/deployer/runner/gke.go
```diff
@@ -309,6 +309,22 @@ func (d *GKEDriver) bindRoles() error {
...
+	log.Println("Verifying gcloud authentication...")
+	// --verbosity flag here disables warnings, and survey output.
+	out, err := exec.NewCommand(`gcloud auth list --filter=status:ACTIVE --format="value(account)" --verbosity error`).StdoutOnly().OutputList()
+	if err != nil {
+		return fmt.Errorf("while retrieving list of credentialed gcloud accounts: %w", err)
+	}
+	// If there's no authenticated user, or the authenticated user doesn't exist in the elastic-cloud-dev account
+	// then we need to authenticate with what's within vault.
+	if len(out) == 0 || (len(out) > 0 && !strings.Contains(out[0], "elastic-cloud-dev")) {
+		if err = authToGCP(
```
barkbay, line 321:
```suggestion
		if err := authToGCP(
```

hack/deployer/runner/gke.go
```diff
@@ -309,6 +309,22 @@ func (d *GKEDriver) bindRoles() error {
...
 func (d *GKEDriver) GetCredentials() error {
+	log.Println("Verifying gcloud authentication...")
+	// --verbosity flag here disables warnings, and survey output.
+	out, err := exec.NewCommand(`gcloud auth list --filter=status:ACTIVE --format="value(account)" --verbosity error`).StdoutOnly().OutputList()
+	if err != nil {
+		return fmt.Errorf("while retrieving list of credentialed gcloud accounts: %w", err)
+	}
+	// If there's no authenticated user, or the authenticated user doesn't exist in the elastic-cloud-dev account
+	// then we need to authenticate with what's within vault.
+	if len(out) == 0 || (len(out) > 0 && !strings.Contains(out[0], "elastic-cloud-dev")) {
```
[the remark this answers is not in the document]
pebrc: I agree with Michael here that we should rely on the context here.

test/e2e/test/step.go
```diff
@@ -32,23 +51,165 @@ func (l StepList) WithStep(testStep Step) StepList {
...
+	// and a single run of diagnostics is incredibly slow.
+	cmd.Env = ensureTmpHomeEnv(cmd.Environ())
+	if err := cmd.Run(); err != nil {
+		log.Error(err, fmt.Sprintf("while running eck-diagnostics: %s", err))
+	}
+}
+
+func uploadDiagnosticsArtifacts() {
+	ctx := Ctx()
+	cmd := exec.Command("gsutil", "cp", "/tmp/*.zip", fmt.Sprintf("gs://eck-e2e-buildkite-artifacts/jobs/%s/%s/", ctx.JobName, ctx.BuildNumber)) //nolint:gosec
```
pebrc, line 136: Should this bucket be configuration rather than being hard-coded? Also who is cleaning the artefacts up once they have been copied out of this storage bucket

.buildkite/scripts/common/run-main-e2e-tests.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+# On any command failures, make sure that diagnostics are copied from bucket to local Buildkite environment
+# to be uploaded as artifacts.
+onError() {
+    # On error, copy any artifacts from the test run locally to be picked up as buildkite artifacts
+    if gsutil ls "gs://eck-e2e-buildkite-artifacts/jobs/$BUILDKITE_PIPELINE_NAME/$BUILDKITE_BUILD_NUMBER/eck-diagnostic*.zip" ; then
+        gsutil cp "gs://eck-e2e-buildkite-artifacts/jobs/$BUILDKITE_PIPELINE_NAME/$BUILDKITE_BUILD_NUMBER/eck-diagnostic*.zip" .
```
pebrc, line 10: Do we need to remove the artefacts after copying or to avoid filling up this bucket over time?

pebrc (commented):
Almost LGTM

hack/deployer/runner/gke.go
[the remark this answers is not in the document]
naemono: I've turned this into a constant and used this instead.

test/e2e/test/step.go
```diff
@@ -32,23 +51,165 @@ func (l StepList) WithStep(testStep Step) StepList {
...
 		if !t.Run(ts.Name, ts.Test) {
-			logf.Log.Error(errors.New("test failure"), "stopping early")
+			logf.Log.Error(errors.New("test failure"), "continuing with additional tests")
 			if ts.OnFailure != nil {
 				ts.OnFailure()
 			}
-			t.FailNow()
+			ctx := Ctx()
+			defer func() {
+				if err := deleteElasticResources(); err != nil {
```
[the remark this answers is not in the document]
naemono: I'm now removing all secrets, and pods in the e2e namespace as well.

test/e2e/test/step.go
naemono: Bucket name is now a configuration/flag, and the bucket is cleaned up in the bash script.

naemono:
Thanks @pebrc . I've implemented these suggested changes and am testing them to see if everything works as expected.

pebrc (approved):
👍 I think we should merge (after your successful test).

.ci/setenvconfig
```diff
@@ -296,6 +298,7 @@ MONITORING_SECRETS=${PROJECT_PATH}/.ci/monitoring-secrets.json
 
 PIPELINE=e2e/main
 BUILD_NUMBER=$BUILD_NUMBER
+JOB_NAME=$JOB_NAME
```
thbkrkr, line 301: IIUC, we look at `JOB_NAME` to know if we need retrive the gcp creds (for the bucket) and know if we can run the diagnostic.
Wouldn't `E2E_FAILFAST=true` be more explicit?

test/e2e/test/step.go
```diff
@@ -42,11 +49,18 @@ func (l StepList) RunSequential(t *testing.T) {
 			continue
 		}
 		if !t.Run(ts.Name, ts.Test) {
-			logf.Log.Error(errors.New("test failure"), "stopping early")
+			logf.Log.Error(errors.New("test failure"), "continuing with additional tests")
 			if ts.OnFailure != nil {
 				ts.OnFailure()
 			}
-			t.FailNow()
```
thbkrkr, line 49: What does it mean for nightly e2e tests where `JOB_NAME` is not defined? Shouldn't we continue to `t.FailNow()` for them?

.ci/setenvconfig
naemono: To your point, wouldn't `CI` environment variable work even better? If running in CI, we can run the diagnostics on each test failure...thoughts?

naemono: oh second thought, idk. Job name is used to build the namespace for the bucket upload as well:
```
	run(ctx, "gsutil", "cp", "/tmp/*.zip", fmt.Sprintf("gs://%s/jobs/%s/%s/", testCtx.GSBucketName, testCtx.JobName, testCtx.BuildNumber))
```
I don't know if I think `E2E_FAILFAST` is more clear, as we are never failing fast at this point, right?

test/e2e/test/step.go
naemono: https://github.com/elastic/cloud-on-k8s/pull/6363/files#diff-e9a5bfc47dbabe806572ddf5571346ffce93e837d675487f39da1a828fc74555R51
```
  JOB_NAME=${BUILDKITE_PIPELINE_NAME}
```

Am I misunderstanding something? Isn't `JOB_NAME` always going to be set, whether in jenkins, or in BK?

thbkrkr: I misread and thought it was only set for e2e/main to start. So we are enabling this for all e2e-tests for all providers.

Don't we need to update [pipeline-nightly-e2e-tests.yml](https://github.com/elastic/cloud-on-k8s/blob/2bf050d8fe19921f2bc770a1bb3840f9eb6acf33/.buildkite/pipeline-nightly-e2e-tests.yml) like for [pipeline.yml](https://github.com/elastic/cloud-on-k8s/pull/6363/files#diff-a3991ebf1475eb82acab13946ffc1eca02b43917f6d5bec3898f45c8b0b9bd53) where we split the cluster creation from the execution of the tests? I think https://github.com/elastic/cloud-on-k8s/pull/6525 would really help to do that.

test/e2e/test/diagnostics.go
```diff
@@ -0,0 +1,65 @@
...
+	})
+	otherNS := append([]string{testCtx.E2ENamespace}, testCtx.Operator.ManagedNamespaces...)
+	// The following appends the test, and it's sub-test names together with a '-'.
+	// Example: For TestAutoscalingLegacy/Secrets_should_eventually_be_created:
+	// testName: TestAutoscalingLegacy, step.Name: Secrets_should_eventually_be_created
+	fullTestName := fmt.Sprintf("%s-%s", testName, step.Name)
+	// Convert any spaces to "_", and "/" to "-" in the test name.
+	normalizedTestName := strings.ReplaceAll(strings.ReplaceAll(fullTestName, " ", "_"), "/", "-")
+	run(ctx, "eck-diagnostics", "--output-directory", "/tmp", "-n", fmt.Sprintf("eck-diagnostic-%s.zip", normalizedTestName), "-o", testCtx.Operator.Namespace, "-r", strings.Join(otherNS, ","), "--run-agent-diagnostics")
+	run(ctx, "gsutil", "cp", "/tmp/*.zip", fmt.Sprintf("gs://%s/jobs/%s/%s/", testCtx.GSBucketName, testCtx.JobName, testCtx.BuildNumber))
```
thbkrkr, line 64: if I resolve well, it gives:

`gs://bucket-name/jobs/job-name/123/eck-diagnostic-TestXyz/Step_abc.zip`

An execution of a job executes several times in parallel the e2e tests suite (true on Jenkins and worse in Buildkite where there is only one job which executes all the combinations). If the same test and subtests fail in all executions, isn't there a diagnostic name conflict?

We could instead use `CLUSTER_NAME` which is unique for each execution of the e2e tests suite.

.ci/setenvconfig
thbkrkr: > wouldn't CI environment variable work even better?

I thought we wanted to control whether we fail fast or not. If the goal is only to switch the feature in CI, yes, I think `CI` is a good candidate.

>  Job name is used to build the namespace for the bucket upload as well

We may not need it and use `CLUSTER_NAME`, see https://github.com/elastic/cloud-on-k8s/pull/6363#discussion_r1139367438.

test/e2e/test/step.go
naemono: Ok, I think I've got this configured for the nightly bits. Thanks for that. I'll work to trigger that to fully test it as soon as I can.

test/e2e/test/diagnostics.go
naemono: good point @thbkrkr .  I'll make some changes to address this.

.ci/setenvconfig
naemono: > I thought we wanted to control whether we fail fast or not. If the goal is only to switch the feature in CI, yes, I think `CI` is a good candidate.
> 
> > Job name is used to build the namespace for the bucket upload as well
> 
> We may not need it and use `CLUSTER_NAME`, see [#6363 (comment)](https://github.com/elastic/cloud-on-k8s/pull/6363#discussion_r1139367438).

I just re-read the issue. There's no mention of toggling on/off. I'll investigate and get back to you

pebrc: There is no need to toggle fail fast behaviour on and off IMO. Thus there is also no need for an environment variable with that name. 

The only distinction we need to introduce is between local runs and CI runs with access to the relevant bucket and where we know that eck-diagnostics is in scope and we want to execute it after failures. Whether we want to use `CI` or `JOB_NAME`  depends mostly if one of them implies the others. E.g if `CI` implies that `JOB_NAME` is set then  `CI` is more expressive as the main variable that control the different behaviour.

test/e2e/test/step.go
pebrc: > Don't we need to update [pipeline-nightly-e2e-tests.yml](https://github.com/elastic/cloud-on-k8s/blob/2bf050d8fe19921f2bc770a1bb3840f9eb6acf33/.buildkite/pipeline-nightly-e2e-tests.yml) like for [pipeline.yml](https://github.com/elastic/cloud-on-k8s/pull/6363/files#diff-a3991ebf1475eb82acab13946ffc1eca02b43917f6d5bec3898f45c8b0b9bd53) where we split the cluster creation from the execution of the tests? I think https://github.com/elastic/cloud-on-k8s/pull/6525 would really help to do that.

The splitting up of the pipelines is an optimisation that happened to be made in this PR but is actually not essential to the feature we are implementing here. But of course we can add it for the nightly pipeline as well for consistency.

naemono:
@thbkrkr @pebrc testing hasn't shown any additional issues after recent changes, including some local runs to ensure things are working properly there as well. This should be ready for another round of review.

pebrc:
> .buildkite/scripts/common/run-main-e2e-tests.sh: line 9: CLUSTER_NAME: unbound variable

naemono:
Some recent change is causing the below when a test fails, at the end of the e2e run:
```
CommandException: One or more URLs matched no objects.
--
  | make: *** [Makefile:520: e2e-run] Error 2
  | 🚨 Error: The command exited with status 2
```
I'm assuming it's something to do with the recent change to the pathing associated with `gsutil`. I'll fix and report back.

naemono:
Ok, @pebrc and @thbkrkr. Both e2e/main, and e2e/nightly tests were run with this, and everything worked as expected. I think one more round of 👀 would be great here when yall have the time. 👍

Makefile
```diff
@@ -517,6 +517,9 @@ e2e-buildah: go-generate buildah-login
 		$(E2E_IMG)
 
 e2e-run: go-generate
+	.buildkite/scripts/common/run-e2e-tests.sh
+
+e2e-run-actual:
```
naemono, line 522: Anyone got any suggestions on a better name here? Naming == hard.

.buildkite/pipeline-nightly-e2e-tests.yml
```diff
@@ -12,7 +12,7 @@ steps:
           - .ci/setenvconfig e2e/aks
           - make -C .ci TARGET="run-deployer e2e-run" ci
         artifact_paths:
-          - "eck-diagnostic-*.zip"
+          - "eck-diagnostics-*.zip"
```
thbkrkr, line 15:
```suggestion
          - "eck-diagnostic*.zip"
```


I think we create `eck-diagnostic-*.zip` in test/e2e/test/diagnostics.go:
```go
run(ctx, "eck-diagnostics", "--output-directory", "/tmp", "-n", fmt.Sprintf("eck-diagnostic-%s.zip", normalizedTestName), "-o", testCtx.Operator.Namespace, "-r", strings.Join(otherNS, ","), "--run-agent-diagnostics")
```

.buildkite/scripts/common/run-e2e-tests.sh
```diff
@@ -0,0 +1,27 @@
...
+    # On error, copy any artifacts from the test run locally to be picked up as buildkite artifacts
+    if gsutil ls "gs://eck-e2e-buildkite-artifacts/jobs/$CLUSTER_NAME/eck-diagnostic*.zip" 2> /dev/null ; then
+        gsutil cp "gs://eck-e2e-buildkite-artifacts/jobs/$CLUSTER_NAME/eck-diagnostic*.zip" .
+        gsutil rm "gs://eck-e2e-buildkite-artifacts/jobs/$CLUSTER_NAME/eck-diagnostic*.zip" .
+    fi
+}
+
+main() {
+    trap 'onError' ERR
+    make -C .ci get-test-artifacts
```
thbkrkr, line 23: What secret do we need to run this script?

.buildkite/scripts/common/run-e2e-tests.sh
```diff
@@ -0,0 +1,27 @@
...
+    make e2e-run-actual
```
thbkrkr, line 24: I wonder if we shouldn't instead keep only the gsutil commands in this script and use the bk mechanism `depends_on/allow_failure: true` to trigger it, as we clean up the cluster.

.buildkite/scripts/common/run-e2e-tests.sh
naemono: I'm not sure this is necessary now with your recent changes adding `pipeline-gen`. I'll remove, test and verify.

.buildkite/scripts/common/run-e2e-tests.sh
naemono: Interesting. Hadn't thought of that. Let me see what that would look like and report back.

naemono: We discussed and decided that having diagnostics files across BK steps wasn't a major issue. I'm working through the changes needed to have the copying of diagnostic files from bucket be triggered by BK cleanup step.

naemono: update on this: We discussed and decided that having artifacts across multiple BK steps wasn't an issue and adjusted the copying of diagnostics to happen in the cleanup phase of BK.

naemono:

.buildkite/scripts/common/run-e2e-tests.sh
naemono: Oh yeah, this was verified, and removed.

test/e2e/cmd/run/run.go
```diff
@@ -261,6 +262,20 @@ func (h *helper) initTestSecrets() error {
 		return err
 	}
 
+	// Only initialize gcp credentials when running in CI
+	if os.Getenv("CI") != "" {
```
thbkrkr, line 266: In theory, only the CI system sets `CI=true`, nobody sets `CI=false`, however it would be more correct to check:
```suggestion
	if os.Getenv("CI") == "true" {
```

test/e2e/test/diagnostics.go
```diff
@@ -0,0 +1,75 @@
...
+	// If we don't have credentials to write to bucket, then we can't run diagnostics
+	// after e2e test failures, as we upload the zip file to a bucket.
+	if _, err := os.Stat(ctx.GCPCredentialsPath); err != nil && errors.Is(err, fs.ErrNotExist) {
+		return false
+	} else if err != nil {
+		log.Error(err, "while checking for existence of %s", ctx.GCPCredentialsPath)
+		return false
+	}
+	// If we're not in CI, then don't run diagnostics on e2e test failures.
+	if os.Getenv("CI") == "" {
```
thbkrkr, line 37: Like https://github.com/elastic/cloud-on-k8s/pull/6363/files#r1151097338.
```suggestion
	if os.Getenv("CI") != "true" {
```

test/e2e/test/diagnostics.go
```diff
@@ -0,0 +1,75 @@
...
+	if !canRunDiagnostics(testCtx) {
+		return
+	}
+	log.Info("running eck-diagnostics job")
+	once.Do(func() {
+		run(ctx, "initialising gs-util", "gcloud", "auth", "activate-service-account", fmt.Sprintf("--key-file=%s", testCtx.GCPCredentialsPath))
+	})
+	otherNS := append([]string{testCtx.E2ENamespace}, testCtx.Operator.ManagedNamespaces...)
+	// The following appends the clustername, test name, and it's sub-test names together with a '-'.
+	// The cluster name is added tot he eck-diagnostics file name to avoid conflicts at the last step
```
thbkrkr, line 66: Typo:
```suggestion
	// The cluster name is added to the eck-diagnostics file name to avoid conflicts at the last step
```

PR #6364, opened 2023-01-27
Move to 8.6.1 and re-enable Fleet tests

pebrc:
Related: I think we need indeed a better mechanism for bumping Helm chart versions (something like `-SNAPSHOT`) otherwise it is going to be really hard to make sure we don't bump unnecessarily (or forget to do it)


naemono (approved):
I think this is what we must do.

Issue #6367, opened 2023-01-31
TestFleetAgentWithoutTLS is failing


thbkrkr:
`TestFleetAgentWithoutTLS` fails in `8.6.1`.

It started to fail with `8.6.0` and was disabled along with `TestFleetMode` (#6308).
`TestFleetMode`  seems fixed in `8.6.1`, but `TestFleetAgentWithoutTLS` still fails.

The test fails because index `metrics-elastic_agent.filebeat-default`is missing.
```
Test:   TestFleetAgentWithoutTLS/ES_data_should_pass_validations
Type:   index_not_found_exception
Reason: no such index [metrics-elastic_agent.filebeat-default]
```

All agents failed to connect to fleet-server:
```
Error: fail to enroll: fail to execute request to fleet-server: dial tcp 172.30.66.19:8220: connect: connection refused
```

Because Fleet-server binds to localhost:

```
> grep 'server listening'  e2e-n1jzw-mercury/pod/test-fleet-agent-notls-fs-kqwz-agent-8884c876d-qf9d5/logs.txt | jq .bind
"localhost:8220"
"localhost:8221"
"localhost:8220"
"localhost:8221"
"localhost:8220"
"localhost:8221"
```

This has been already identified in this closed PR: https://github.com/elastic/cloud-on-k8s/pull/6214.

The initial change in agent causing this regression:
- https://github.com/elastic/elastic-agent/pull/1867

Known issue in agent:
- https://github.com/elastic/elastic-agent/issues/2197

Which should be fixed by:
- https://github.com/elastic/elastic-agent/pull/2198.

thbkrkr:
Probably unrelated to the test failure, there are two panics in the fleet server log:

```json
{
    "log.level": "error",
    "@timestamp": "2023-01-31T02:00:09.850Z",
    "message": "Harvester crashed with: harvester panic with: close of closed channel
    goroutine 212 [running]:
    runtime/debug.Stack()
    	runtime/debug/stack.go:24 +0x65
    github.com/elastic/beats/v7/filebeat/input/filestream/internal/input-logfile.startHarvester.func1.1()
    	github.com/elastic/beats/v7/filebeat/input/filestream/internal/input-logfile/harvester.go:167 +0x78
    panic({0x55f8b504c1c0, 0x55f8b5613ae0})
    	runtime/panic.go:844 +0x258
[...]

Issue #6331, opened 2023-01-17
TestFleet* fail on 8.6.x


thbkrkr:
When updating the stack version to 8.6.0 (#6327), `TestFleetKubernetesIntegrationRecipe` fails.

https://devops-ci.elastic.co/job/cloud-on-k8s-pr-e2e-tests/98/

```
=== RUN   TestFleetKubernetesIntegrationRecipe/ES_data_should_pass_validations
Retries (30m0s timeout): .......
    step.go:44: 
        Error: Received unexpected error:
            elasticsearch client failed for https://elasticsearch-6qxw-es-internal-http.e2e-nru4o-mercury.svc:9200/_data_stream/logs-elastic_agent.fleet_server-default: 404 Not Found: 
            RootCause:[{
                Reason: no such index [logs-elastic_agent.fleet_server-default]
                Type: index_not_found_exception
             }]
    --- FAIL: TestFleetKubernetesIntegrationRecipe/ES_data_should_pass_validations (1800.00s)
```

Index `logs-elastic_agent.fleet_server-default`is missing because fleet server and agents are in error.  

I can reproduce the issue just by applying [`config/recipes/elastic-agent/fleet-kubernetes-integration.yaml`](https://github.com/elastic/cloud-on-k8s/blob/main/config/recipes/elastic-agent/fleet-kubernetes-integration.yaml) after updating the stack version to `8.6.0`.

The fleet-server agent pod is running, but the log shows that the process terminated as soon as it was enrolled.

```
> k logs fleet-server-agent-6b8bb8976f-hcz24 | grep '{' | tail -15 | jq .message -r
fleet-server routes set up
server hard connection limiter installed
Running on policy with Fleet Server integration: eck-fleet-server; missing config fleet.agent.id (expected during bootstrap process)
Unit state changed fleet-server-default-fleet-server (STARTING->DEGRADED): Running on policy with Fleet Server integration: eck-fleet-server; missing config fleet.agent.id (expected during bootstrap process)
Unit state changed fleet-server-default (STARTING->DEGRADED): Running on policy with Fleet Server integration: eck-fleet-server; missing config fleet.agent.id (expected during bootstrap process)
Fleet Server - Running on policy with Fleet Server integration: eck-fleet-server; missing config fleet.agent.id (expected during bootstrap process)
Starting enrollment to URL: https://fleet-server-agent-http.default.svc:8220/
Elastic Agent successfully enrolled
Elastic Agent successfully enrolled
[...]

Issue #6308, opened 2023-01-11
TestFleetMode is failing


barkbay:
`TestFleetMode` is [failing](https://devops-ci.elastic.co/job/cloud-on-k8s-e2e-tests-snapshot-versions/759/testReport/) in `cloud-on-k8s-e2e-tests-snapshot-versions`:

```
13:31:25  {"Time":"2023-01-04T12:31:25.72786109Z","Action":"output","Package":"github.com/elastic/cloud-on-k8s/v2/test/e2e/agent","Test":"TestFleetMode/ES_data_should_pass_validations","Output":"    --- FAIL: TestFleetMode/ES_data_should_pass_validations (1800.00s)\n"}
13:31:25  {"Time":"2023-01-04T12:31:25.727873411Z","Action":"fail","Package":"github.com/elastic/cloud-on-k8s/v2/test/e2e/agent","Test":"TestFleetMode/ES_data_should_pass_validations","Elapsed":1800}
13:31:25  {"Time":"2023-01-04T12:31:25.727880965Z","Action":"fail","Package":"github.com/elastic/cloud-on-k8s/v2/test/e2e/agent","Test":"TestFleetMode","Elapsed":1893.41}
13:31:25  {"Time":"2023-01-04T12:31:25.72789171Z","Action":"output","Package":"github.com/elastic/cloud-on-k8s/v2/test/e2e/agent","Output":"FAIL\n"}
```

barkbay:
I'm running all the Agent tests right now using ECK `2.6.0` to open a PR soonish. I might also need to disable `TestFleetAgentWithoutTLS` which is failing for no obvious reasons:

<img width="506" alt="image" src="https://user-images.githubusercontent.com/976373/211819782-1934b884-282b-4614-889b-b29a82f0991b.png">

barkbay:
`TestAgentVersionUpgradeToLatest8x` was eventually ✅ 

<img width="506" alt="image" src="https://user-images.githubusercontent.com/976373/211826986-7ac89546-f083-4242-a3be-157b70793b5c.png">

thbkrkr:
I took a quick look to see if this is similar to:
- #6331

I see the same pattern: "enrollment and shutdown" but then the agent is restarted.

```
>  k -n e2e-mercury logs test-agent-fleet-fs-nlnt-agent-696889fb48-8qj7g | grep '{' | jq .message | less
"Fleet Server - Starting"
"Updating running component model"
[...]

PR #6368, opened 2023-01-31
Disable TestFleetAgentWithoutTLS and TestFleetKubernetesIntegrationRecipe

thbkrkr:
We re-enabled them yesterday with the `8.6.1` update (#6364) thinking they were failing for the same reason as `TestFleetMode` (#6308) but that is not the case.

This disables:
- `TestFleetAgentWithoutTLS` due to a known issue 
- `TestFleetKubernetesIntegrationRecipe` for yet unknown issue 

I kept `TestFleetAPMIntegrationRecipe` and `TestFleetCustomLogsIntegrationRecipe` but I would not be surprised if they are affected by the same problem as `TestFleetKubernetesIntegrationRecipe`. I'm going to trigger a few runs from this PR to try to confirm that.

Relates to:
- #6331
- #6367

### DIFF
--- a/test/e2e/agent/recipes_test.go
+++ b/test/e2e/agent/recipes_test.go
@@ -88,6 +88,9 @@ func TestMultiOutputRecipe(t *testing.T) {
 }
 
 func TestFleetKubernetesIntegrationRecipe(t *testing.T) {
+	// This test is disabled until we understand why it is failing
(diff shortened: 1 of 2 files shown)


thbkrkr:
run/e2e-tests tags=agent

pebrc (approved):

PR #6388, opened 2023-02-08
Update git config of ci container for recent git versions

thbkrkr:
This commit configures git in the ci container to add an exception for the repository to avoid error `fatal: detected dubious ownership in repository` as recent versions of git require the .git folder to be owned by the current user.

Relates to https://github.blog/2022-04-12-git-security-vulnerability-announced/.


barkbay (approved):
LGTM

PR #6392, opened 2023-02-08
Delete migrated Jenkins jobs on Buildkite

thbkrkr:
Delete Jenkins jobs that are migrated on Buildkite (main and pr)
and others no longer or not used (ocp3 and custom).


thbkrkr:
It's normal that the Jenkins pr build failed because this PR deletes its pipeline definition but as long as it is not merged into main the job still exists and look for this definition.

naemono:
What about https://github.com/elastic/cloud-on-k8s/blob/main/.ci/jobs/pr-gke-e2e-tests.yml?

It's triggering `script-path: .ci/pipelines/e2e-tests-main-gke.Jenkinsfile`, which is being removed, and also is triggered by `trigger-phrase: '(run\sfull\spr\sbuild).*|(run/e2e-tests).*'`, which we talk about In our [Contributing guide](https://github.com/elastic/cloud-on-k8s/blob/main/CONTRIBUTING.md).

I think we should open an issue to modify our contributing guide for current BK environment, but the changes to it can come in a later PR.

thbkrkr:
Good point, the Contribution Guide needs to be updated. It's a bit wrong what it says there because today a community user can't use the comment trigger himself for security reasons.

naemono (approved):

PR #6404, opened 2023-02-10
Initial Commit of Basic Operator for Logstash

robbavey:
Initial commit of a simple operator.

The first operator will create:
* A Service exposing the logstash metrics API, so it can be used for monitoring purposes
* A secret holding logstash.yml
* A StatefulSet deploying the logstash pods
  * Pods deployed with a readiness probe testing against metrics API

The sample logstash yml file as located in config/samples/logstash/logstash.yaml will
create:

```
✗ kubectl tree logstash logstash-sample
NAMESPACE  NAME                                                  READY  REASON  AGE
default    Logstash/logstash-sample                              -              4m5s
default    ├─Secret/logstash-sample-ls-config                    -              4m4s
default    ├─Service/logstash-sample-ls-http                     -              4m5s
default    │ └─EndpointSlice/logstash-sample-ls-http-kwfsg       -              4m5s
default    └─StatefulSet/logstash-sample-ls                      -              4m4s
default      ├─ControllerRevision/logstash-sample-ls-79bd59f869  -              4m4s
default      ├─Pod/logstash-sample-ls-0                          True           3m59s
default      ├─Pod/logstash-sample-ls-1                          True           3m59s
default      └─Pod/logstash-sample-ls-2                          True           3m59s
```

And shows status:

```
✗ kubectl get logstash logstash-sample
NAME              AVAILABLE   EXPECTED   AGE    VERSION
logstash-sample   3           3          9m1s   8.6.1
```

This PR also includes basic tests:

* Unit tests for names and validations in `pkg/apis`
* End to end tests in e2e/logstash
  * These test a single logstash, multiple logstashes and upgrade to latest `8.x`
  * Tests that logstash is setup and torn down correctly, and returns status correctly
 * Integration into `TestSamples` and `TestVersionUpgradeOrdering` in the test suites.

 Still todo: 
* Unit tests for controller

  

Co-authored-by: Kaise Cheng <kaise.cheng@elastic.co>


robbavey:
run/e2e-tests

robbavey:
@thbkrkr, @kaisecheng, @naemono, @barkbay I think this is ready for review now. 

There are a few things that are still in progress - the logstash controller still needs unit tests, but the basics are there, and I've added a unit tests for the `api` package, and added some basic end to end tests, and update the `stack` and `sample` tests to include logstash tests too - I haven't added to the `Smoke` tests, as from where I'm looking, they appear to be a somewhat limited and targeted set of tests. I've tested the logstash, `smoke`, `stack` and `sample` e2e tests locally, but would like to test the full e2e test suite too.

pkg/controller/logstash/sset/sset.go
```diff
@@ -0,0 +1,97 @@
...
+		Spec: appsv1.StatefulSetSpec{
+			UpdateStrategy: appsv1.StatefulSetUpdateStrategy{
+				Type: appsv1.RollingUpdateStatefulSetStrategyType,
+			},
+			// we don't care much about pods creation ordering, and manage deletion ordering ourselves,
+			// so we're fine with the StatefulSet controller spawning all pods in parallel
+			PodManagementPolicy:  appsv1.ParallelPodManagement,
+			RevisionHistoryLimit: params.RevisionHistoryLimit,
+			// build a headless service per StatefulSet, matching the StatefulSet labels
+			ServiceName: params.ServiceName,
```
pebrc, line 49: The service referenced here is not a headless service but the HTTP service for Logstash which has a ClusterIP. This means the individual Logstash Pods are not resolvable via DNS (because that service always resolves to the ClusterIP of the service and the routing then happens with the `iptables` magic of kube proxy). 

The K8s docs ask for a headless service for each stateful set. The thing I am not sure about is whether we need it if we don't need the ability to resolve the individual Logstash Pods via DNS.

pebrc (commented):
Nice work! I left a few comments. But in general I think we could merge this already as a minimal viable version of Logstash support. The only thing that we should maybe decide is the API for the services/HTTP attribute in the CRD because that is really messy to change once we released it.

pkg/apis/logstash/v1alpha1/webhook.go
```diff
@@ -0,0 +1,88 @@
...
+	var errors field.ErrorList
+	if old != nil {
+		for _, uc := range updateChecks {
+			if err := uc(old, a); err != nil {
+				errors = append(errors, err...)
+			}
+		}
+
+		if len(errors) > 0 {
+			return apierrors.NewInvalid(groupKind, a.Name, errors)
```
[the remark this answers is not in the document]
kaisecheng: you are right. It should go through all checks.

pkg/controller/common/version/version.go
```diff
@@ -33,6 +33,7 @@ var (
 	// Due to bugfixes present in 7.14 that ECK depends on, this is the lowest version we support in Fleet mode.
 	SupportedFleetModeAgentVersions = MinMaxVersion{Min: MustParse("7.14.0-SNAPSHOT"), Max: From(8, 99, 99)}
 	SupportedMapsVersions           = MinMaxVersion{Min: From(7, 11, 0), Max: From(8, 99, 99)}
+	SupportedLogstashVersions       = MinMaxVersion{Min: From(8, 6, 0), Max: From(8, 99, 99)}
```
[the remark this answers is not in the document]
kaisecheng: Logstash needs to add some features to work well with operator and we have no plan to backport those features to 7.17

pkg/controller/logstash/config.go
```diff
@@ -0,0 +1,125 @@
...
+		return results.WithError(err)
+	}
+
+	_, _ = configHash.Write(cfgBytes)
+
+	return results
+}
+
+func buildConfig(params Params) ([]byte, error) {
+	existingCfg, err := getExistingConfig(params.Context, params.Client, params.Logstash)
```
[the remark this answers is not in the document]
kaisecheng: `existingCfg` should be removed. We have no reason to retrieve config from secret. If users want to store configuration in server, they should use central pipeline management. Thanks for raising this.

pkg/controller/logstash/sset/sset.go
kaisecheng: I changed it to headless service following the k8s doc. The default service in Logstash is only for metrics collection. Having load balancing with iptables sounds more than it needs.

pkg/apis/logstash/v1alpha1/name.go
```diff
@@ -0,0 +1,40 @@
...
+// you may not use this file except in compliance with the Elastic License 2.0.
+
+package v1alpha1
+
+import (
+	common_name "github.com/elastic/cloud-on-k8s/v2/pkg/controller/common/name"
+)
+
+const (
+	defaultServiceSuffix = "default"
```
robbavey, line 12: What do you think about using the name "api" rather than "default", to reflect that this service is for the Logstash API? So this would show up as `logstash-sample-ls-api`, rather than `logstash-sample-ls-default`?

pkg/controller/logstash/reconcile.go
```diff
@@ -0,0 +1,82 @@
...
+)
+
+func reconcileStatefulSet(params Params, podTemplate corev1.PodTemplateSpec) (*reconciler.Results, logstashv1alpha1.LogstashStatus) {
+	defer tracing.Span(&params.Context)()
+	results := reconciler.NewResult(params.Context)
+
+	s := sset.New(sset.Params{
+		Name:                 logstashv1alpha1.Name(params.Logstash.Name),
+		Namespace:            params.Logstash.Namespace,
+		ServiceName:          logstashv1alpha1.DefaultServiceName(params.Logstash.Name),
```
pebrc, line 34: You are setting the default service name here but you are only creating the default service if no custom service is defined.

pkg/controller/logstash/driver.go
```diff
@@ -0,0 +1,87 @@
...
+	status := logstash.Status
+	status.ObservedGeneration = logstash.Generation
+	return status
+}
+
+func internalReconcile(params Params) (*reconciler.Results, logstashv1alpha1.LogstashStatus) {
+	defer tracing.Span(&params.Context)()
+	results := reconciler.NewResult(params.Context)
+
+	_, err := reconcileServices(params)
```
pebrc, line 74: You are ignoring the return value here. Can it simply be removed from the function?

pebrc (commented):
I realise additional commits were pushed since you requested my review. Some of my comments might thus be outdated already.

pkg/apis/logstash/v1alpha1/logstash_types.go
```diff
@@ -0,0 +1,143 @@
...
+	// +kubebuilder:pruning:PreserveUnknownFields
+	Config *commonv1.Config `json:"config,omitempty"`
+
+	// ConfigRef contains a reference to an existing Kubernetes Secret holding the Logstash configuration.
+	// Logstash settings must be specified as yaml, under a single "logstash.yml" entry. At most one of [`Config`, `ConfigRef`]
+	// can be specified.
+	// +kubebuilder:validation:Optional
+	ConfigRef *commonv1.ConfigSource `json:"configRef,omitempty"`
+
+	// HTTP holds the HTTP layer configuration for the Logstash Metrics API
```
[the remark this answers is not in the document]
robbavey: Fixed

pkg/apis/logstash/v1alpha1/logstash_types.go
```diff
@@ -0,0 +1,143 @@
...
+	// +kubebuilder:validation:Optional
+	ServiceAccountName string `json:"serviceAccountName,omitempty"`
+}
+
+type LogstashService struct {
+	Name string `json:"name,omitempty"`
+	// Service defines the template for the associated Kubernetes Service object.
+	Service commonv1.ServiceTemplate `json:"service,omitempty"`
+	// TLS defines options for configuring TLS for HTTP.
+	TLS commonv1.TLSOptions `json:"tls,omitempty"`
```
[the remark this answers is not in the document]
robbavey: Yes, we would like to add support for [TLS](https://github.com/elastic/logstash/blob/main/config/logstash.yml#L149) for the metrics API in a follow-up commit.

config/samples/logstash/logstash_svc.yaml
```diff
@@ -0,0 +1,42 @@
...
+              targetPort: 9600
+    - name: beats
+      service:
+        spec:
+          type: ClusterIP
+          ports:
+            - port: 5044
+              name: "beats"
+              protocol: TCP
+              targetPort: 5044
```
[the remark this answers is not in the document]
robbavey: Good point, the name seems redundant here, will remove.

I do think we may need the flexibility to define different service types, given the variety of plugins we support, and the use case to support multiple of those in a single logstash config

pkg/controller/logstash/service.go
```diff
@@ -0,0 +1,92 @@
...
+	"github.com/elastic/cloud-on-k8s/v2/pkg/controller/common/defaults"
+	"github.com/elastic/cloud-on-k8s/v2/pkg/controller/logstash/network"
+)
+
+// reconcileServices reconcile Services defined in the spec.
+// When Services are empty, a default Service for port 9600 is created.
+// If api.http.port is customized, user is expected to config Services.
+// When Services exist, the port 9600 does not attach to any of Service.
+func reconcileServices(params Params) ([]corev1.Service, error) {
+	if len(params.Logstash.Spec.Services) == 0 {
```
robbavey, line 22: Rather than checking if the number of services is 0, we should probably check whether a service has been spec'd with the same name as the `default` service - I think it may make sense to rename this to `api`, or `internal`, rather than `default`, WDYT?

kaisecheng: we can make a default headless service regardless of user input similar to ES
```
NAME                                    TYPE        CLUSTER-IP      EXTERNAL-IP   PORT(S)    AGE
elasticsearch-sample-es-default         ClusterIP   None            <none>        9200/TCP   17h
elasticsearch-sample-es-http            ClusterIP   10.98.229.240   <none>        9200/TCP   17h
elasticsearch-sample-es-internal-http   ClusterIP   10.96.255.68    <none>        9200/TCP   17h
elasticsearch-sample-es-transport       ClusterIP   None            <none>        9300/TCP   17h
```
For user defined service, name it with suffix`api`.  If user doesn't provide input, create a `ClusterIP` service with suffix `http` for metrics.

robbavey: I was wondering if we could call the service for metrics `api`,`internal` or `http`, regardless of whether they have set values in the `spec`. And if they define a service with the same name, it overrides the default settings, allowing the user to change the cluster type from the default if they need to - say if they have an external monitoring tool that reads from the logstash API.

kaisecheng: yes. Using the name as an identifier to override default settings (eg. port) sounds good. 

My concern is this [comment](https://github.com/elastic/cloud-on-k8s/pull/6404#discussion_r1111119489) 
> K8s docs ask for a headless service for each stateful set

Regardless of user's setting in `spec`, I think we can always create headless service for Logstash

robbavey: Yes, I agree - I do think we need to always create a headless service

config/samples/logstash/logstash_svc.yaml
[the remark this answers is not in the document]
pebrc: That's a good point. I guess a ordinal system would be possible but there is also a `metadata` section in the service template that potentially allows users to specify a name (even though current docs say we will ignore it) so we have another redundancy. If it is easier to stick with the name attribute I think I would be OK with that too.

```
  services:
    - service:
        metadata:
          name: "stats"
        spec:
          type: ClusterIP
          ports:
            - port: 9600
              # N.B. https://istio.io/latest/docs/ops/configuration/traffic-management/protocol-selection/
              name: "http-stats"
              protocol: TCP
              targetPort: 9600
```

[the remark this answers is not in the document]
robbavey: On reflection, I think the name scheme works for us better than an ordinal scheme might do - it provides context when viewing service details, and also to allow the rules for the default API service to be overridden if necessary. I'd be happy to use the `metadata` method if that is more idiomatic, but I'd also like to be consistent with the rest of ECK wherever possible, and having a special case where `metadata` is meaningful seems like it might be a cause for confusion.

pkg/controller/logstash/driver.go
roaksoax: @kaisecheng or it could be removed and added in the upcoming PR that adds TLS support?

kaisecheng: We will add TLS support in the near future, so prefer to keep it. I can add a TODO to explain

robbavey:
@pebrc This should be ready for another round. Thanks!

barkbay (commented):
Do we also want to update:

* The memory aggregator: https://github.com/elastic/cloud-on-k8s/blob/48d6eed92fac68c86490e878a359195f00afa3f0/pkg/license/aggregator.go#L38-L56
* The telemetry reporter: https://github.com/elastic/cloud-on-k8s/blob/48d6eed92fac68c86490e878a359195f00afa3f0/pkg/telemetry/telemetry.go#L112-L135

?

config/samples/logstash/logstash.yaml
```diff
@@ -0,0 +1,27 @@
+apiVersion: elasticsearch.k8s.elastic.co/v1
+kind: Elasticsearch
+metadata:
+  name: elasticsearch-sample
```
barkbay, line 4: The purpose of this Elasticsearch resource is not clear.

hack/operatorhub/config.yaml
```diff
@@ -30,6 +30,9 @@ crds:
   - name: stackconfigpolicies.stackconfigpolicy.k8s.elastic.co
     displayName: Elastic Stack Config Policy
     description: Elastic Stack Config Policy
+  - name: logstashes.logstash.k8s.elastic.co
```
barkbay, line 33: We may also want to update [hack/operatorhub/templates/csv.tpl](https://github.com/elastic/cloud-on-k8s/blob/main/hack/operatorhub/templates/csv.tpl)

pkg/apis/logstash/v1alpha1/logstash_types.go
```diff
@@ -0,0 +1,142 @@
...
+)
+
+const (
+	// Kind is inferred from the struct name using reflection in SchemeBuilder.Register()
+	// we duplicate it as a constant here for practical purposes.
+	Kind = "Logstash"
+)
+
+// LogstashSpec defines the desired state of Logstash
+type LogstashSpec struct {
```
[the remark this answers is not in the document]
robbavey: Yes, there will be a follow up PR to add support for Elasticsearch references

pkg/controller/logstash/sset/sset.go
```diff
@@ -0,0 +1,97 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      params.Name,
+			Namespace: params.Namespace,
+			Labels:    params.Labels,
+		},
+		Spec: appsv1.StatefulSetSpec{
+			UpdateStrategy: appsv1.StatefulSetUpdateStrategy{
+				Type: appsv1.RollingUpdateStatefulSetStrategyType,
+			},
+			// we don't care much about pods creation ordering, and manage deletion ordering ourselves,
```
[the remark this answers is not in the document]
robbavey: I agree - I think we will want to expose this at a later date

hack/operatorhub/config.yaml
robbavey: Done

config/samples/logstash/logstash.yaml
robbavey: Done

pkg/controller/logstash/reconcile.go
```diff
@@ -0,0 +1,82 @@
...
+	if err != nil {
+		return results.WithError(err), params.Status
+	}
+
+	var status logstashv1alpha1.LogstashStatus
+	if status, err = calculateStatus(&params, reconciled.Status.ReadyReplicas, reconciled.Status.Replicas); err != nil {
+		err = errors.Wrap(err, "while calculating status")
+	}
+
+	return results.WithError(err), status
```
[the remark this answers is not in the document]
robbavey: Done

pkg/apis/logstash/v1alpha1/name.go
```diff
@@ -0,0 +1,40 @@
...
+
+// Namer is a Namer that is configured with the defaults for resources related to a Logstash resource.
+var Namer = common_name.NewNamer("ls")
+
+// ConfigSecretName returns the name of a secret used to storage Logstash configuration data.
+func ConfigSecretName(name string) string {
+	return Namer.Suffix(name, configSuffix)
+}
+
+func ConfigMapName(name string) string {
```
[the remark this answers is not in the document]
robbavey: Removed

robbavey:
@barkbay Thanks for the comments - I think I addressed the points from your review

barkbay (commented):
Overall LGTM,  a few other nits:
* Do we plan to add some unit tests? There is none at the moment in the `controller` package. I feel like it would make sense for functions like `reconcileServices`  or `buildPodTemplate`.
* Other ECK resources have a `Health` field in their status. I was wondering if it would make sense to add one for Logstash. The status already holds the number of `availableNodes`, I don't know if we could offer more to help the user understand the Logstash resource health?
* Should we already open an issue for each feature/task that it is still to be done or discussed? A few that I have in mind:
  * Adding the Elasticsearch reference
  * Exposing RollingUpdateStatefulSetStrategy
  * Implementing TLS
  * Update the documentation
  * Discussing telemetry and usage report

robbavey:
@barkbay Unit tests is a good point - I'll add. And I'll take a look at the health and see what makes sense for Logstash

As for follow up tasks - we have a plan to use this PR as a building block for additional functionality. The additional functionality includes stack monitoring, better support for handling logstash pipelines, Elasticsearch references, TLS support for monitoring, etc. 

See https://github.com/elastic/ingest-dev/issues/1441 for meta issue

robbavey:
@barkbay - given the LGTM, would you have any objections to having this committed, and adding unit tests later today/early tomorrow (especially as this PR is targeted to a `feature/logstash` branch) - I would like to unblock the in progress [stack monitoring PR](#6436 )

barkbay (approved):
LGTM!

Issue #6090, opened 2022-10-13
Create an ES client for the observer only if needed


thbkrkr:
Since #2916, we close the ES client at each observation to avoid memory leak, because at each reconciliation we create a new ES client even if we don't need it.

https://github.com/elastic/cloud-on-k8s/blob/85aa75b7a7765330475c0b92a0851939ac3d9578/pkg/controller/elasticsearch/observer/manager.go#L78-L80

It had been identified as a first non-optimal remediation that we had to improve, but ultimately we never did.

https://github.com/elastic/cloud-on-k8s/pull/2916#issuecomment-616603001:
> I think that instead of passing in a fully instantiated client that gets thrown away most of the time, we should pass a function that creates the client if needed only. This is because creating the client still requires some work like adding the certificates to a trust chain and creating a new HTTP client etc.

PR #6407, opened 2023-02-13
Create ES client for observer only if needed

andreaskapfer:
This PR implements the functionality to create a new ES client for the observer only if needed, i.e. if there are changes in the client config.

It fixes https://github.com/elastic/cloud-on-k8s/issues/6090


pkg/controller/elasticsearch/client/client.go
```diff
@@ -123,6 +123,10 @@ type Client interface {
 	// The Elasticsearch endpoint will be added automatically to the request URL which should therefore just be the path
 	// with a leading /
 	Request(ctx context.Context, r *http.Request) (*http.Response, error)
+	// BasicAuthUser returns the BasicAuthUser configured for this client
```
naemono, line 126:
```suggestion
	// BasicAuthUser returns the credentials for an Elasticsearch user.
~~~

**Review by naemono — COMMENTED** [review]
Changes look good. Minor comment changes requested.

File: pkg/controller/elasticsearch/client/mock.go [hunk 2]
```diff
@@ -33,6 +34,18 @@ func NewMockClientWithUser(v version.Version, u BasicAuth, fn RoundTripFunc) Cli
 	return versioned(baseClient, v)
 }
 
+func NewFullMockClient(v version.Version, u BasicAuth, fn RoundTripFunc, endpoint string, caCerts []*x509.Certificate) Client {
```
pebrc, line 37: Not sure if we need this. Could you not just use the regular client constructor function `esclient.NewElasticsearchClient` for the test where this is used?

pkg/controller/elasticsearch/driver/driver.go
```diff
@@ -523,3 +539,18 @@ func esReachableConditionMessage(internalService *corev1.Service, isServiceReady
 		return fmt.Sprintf("Service %s/%s has endpoints", internalService.Namespace, internalService.Name)
 	}
 }
+
+func isEsClientUpToDate(esClient esclient.Client, version version.Version, user esclient.BasicAuth, url string, caCerts []*x509.Certificate) bool {
```
pebrc, line 543: Feels like this is almost identical to the existing `Equals` function on the client and maybe should be defined in the client as well to avoid the two implementations from diverging and introducing subtle bugs i.e. this function returning the same client but the `Equals` function returning a different result.

pkg/controller/elasticsearch/client/client.go
andreaskapfer: method and comment not needed anymore

pkg/controller/elasticsearch/client/client.go
```diff
@@ -123,6 +123,10 @@ type Client interface {
...
+	BasicAuthUser() BasicAuth
+	// CaCerts returns the CaCerts configured for this client
```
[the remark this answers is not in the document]
andreaskapfer: method and comment not needed anymore

pkg/controller/elasticsearch/client/mock.go
andreaskapfer: Yes, we can. Done.

pkg/controller/elasticsearch/observer/manager.go
```diff
@@ -67,13 +67,20 @@ func (m *Manager) getObserver(key types.NamespacedName) (*Observer, bool) {
...
+func (m *Manager) Observe(ctx context.Context, cluster esv1.Elasticsearch, esClientProvider func(client.Client) client.Client, isServiceReady bool) *Observer {
 	defer tracing.Span(&ctx)()
 	nsName := k8s.ExtractNamespacedName(&cluster)
 	settings := m.extractObserverSettings(ctx, cluster)
 
 	observer, exists := m.getObserver(nsName)
 
+	var esClient client.Client
+	if exists {
+		esClient = esClientProvider(observer.esClient)
```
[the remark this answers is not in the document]
andreaskapfer: I think for readability it is better to carry out the comparison twice.

pkg/controller/elasticsearch/driver/driver.go
```diff
@@ -391,6 +391,22 @@ func (d *defaultDriver) newElasticsearchClient(
...
+	user esclient.BasicAuth,
+	v version.Version,
+	caCerts []*x509.Certificate,
+) func(existingEsClient esclient.Client) esclient.Client {
+	return func(existingEsClient esclient.Client) esclient.Client {
+		url := services.ElasticsearchURL(d.ES, state.CurrentPodsByPhase[corev1.PodRunning])
+		if existingEsClient != nil && existingEsClient.HasProperties(v, user, url, caCerts) {
+			return existingEsClient
+		}
+		return d.newElasticsearchClient(ctx, state, user, v, caCerts)
```
thbkrkr, line 406: Should we close the existing client not reused here?
```suggestion
		existingEsClient.Close()
		return d.newElasticsearchClient(ctx, state, user, v, caCerts)
```

pkg/controller/elasticsearch/observer/manager.go
```diff
@@ -85,8 +92,7 @@ func (m *Manager) Observe(ctx context.Context, cluster esv1.Elasticsearch, esCli
 		// in case asynchronous observation has been disabled ensure at least one observation at reconciliation time.
 		return m.getAndObserveSynchronously(ctx, nsName)
 	default:
-		// No change, close the provided Client and return the existing observer.
-		esClient.Close()
```
thbkrkr, line 89: This was preventing the memory leak (https://github.com/elastic/cloud-on-k8s/pull/2916/files). Where do close al clients that are not reused because connexion information has changed?

pkg/controller/elasticsearch/client/client.go
```diff
@@ -128,6 +128,8 @@ type Client interface {
 	Version() version.Version
 	// URL returns the Elasticsearch URL configured for this client
 	URL() string
+	// HasProperties checks whether this client has the indicated properties
```
pebrc, line 131:
```suggestion
	// HasProperties checks whether this client has the indicated properties.
```

pebrc (approved):
LGTM

pkg/controller/elasticsearch/observer/manager.go
```diff
@@ -67,13 +67,20 @@ func (m *Manager) getObserver(key types.NamespacedName) (*Observer, bool) {
 
 // Observe gets or create a cluster state observer for the given cluster
 // In case something has changed in the given esClient (eg. different caCert), the observer is recreated accordingly
-func (m *Manager) Observe(ctx context.Context, cluster esv1.Elasticsearch, esClient client.Client, isServiceReady bool) *Observer {
+func (m *Manager) Observe(ctx context.Context, cluster esv1.Elasticsearch, esClientProvider func(client.Client) client.Client, isServiceReady bool) *Observer {
```
thbkrkr, line 70: We already have an EsClientProvider type:

https://github.com/elastic/cloud-on-k8s/blob/b6eca617087885b672391bc889af66ad59154623/pkg/controller/autoscaling/elasticsearch/controller.go#L42

Can we consolidate your changes with this existing?

pkg/controller/elasticsearch/observer/manager.go
pebrc: https://github.com/elastic/cloud-on-k8s/blob/180210427177dbf1046417e08be4a8d92c2ce2f1/pkg/controller/elasticsearch/observer/observer.go#L99 closes the client when the observer is stopped. This one only existed to close client that were created unnecessarily.

pkg/controller/elasticsearch/observer/manager.go
pebrc: I am not sure it makes sense. I agree that the there is an unfortunate naming overlap. But here we actually have different function signature `func(esclient.Client) esclient.Client`. We can think about a better name maybe?

pkg/controller/elasticsearch/observer/manager.go
thbkrkr: Perfect.

pkg/controller/elasticsearch/observer/manager.go
thbkrkr: Hmm, I see. Let's keep it that way.

thbkrkr:
buildkite test this

thbkrkr:

naemono (approved):

PR #6418, opened 2023-02-15
Refactoring build and test secrets setup

thbkrkr:
This adds a new `vault` package to load secrets for the license generator used in go generate and the e2e tests runner.

This gives the ability to stop to always write secrets on disk, only if requested.

Current behaviour is not changed: if the file exists, nothing is done, if the file doesn't exist, it is retrieved from Vault and written on disk. A new mode is added that retrieves the file from Vault and keep it in memory.
- `TEST_LICENSE=`: no secret read
- `TEST_LICENSE=.ci/test-license.json`: secret read and written to the given path
- `TEST_LICENSE=in-memory`: secret read but not write to disk

This simplifies the setup of the secrets that needs a different path depending on the context (k8s agent or docker in gcp agent) because when in docker we set the path before being in docker.

Testing:
```sh
 GO_TAGS=release LICENSE_PUBKEY=$(pwd)/.ci/license-dev.pubkey make go-generate
 GO_TAGS=release BUILD_LICENSE_KEY=dev LICENSE_PUBKEY=$(pwd)/.ci/license.pubkey make go-generate
```


.ci/Dockerfile
```diff
@@ -92,6 +92,12 @@ RUN curl -fsSL -o get_helm.sh https://raw.githubusercontent.com/helm/helm/main/s
 
 RUN helm plugin install https://github.com/quintush/helm-unittest --version 0.2.8
 
+ENV VAULT_VERSION=1.12.3
```
pebrc, line 95: Do we also need to add this to the buildkite agent image?

pkg/utils/vault/vault.go
```diff
@@ -0,0 +1,173 @@
...
+		}
+		return field
+	}
+}
+
+func (f SecretFile) readFromVault() ([]byte, error) {
+	if os.Getenv(vaultAddrEnvVar) == "" {
+		return nil, fmt.Errorf("%s must be set", vaultAddrEnvVar)
+	}
+	if os.Getenv(vaultTokenEnvVar) == "" {
```
pebrc, line 88: Should we have similar functionality like https://github.com/elastic/cloud-on-k8s/blob/0207226ffbb03f8eff9f549ae103e96991a06f41/hack/deployer/vault/vault.go#L114 where it tries to use existing vault token cache when running on a developer machine?

pkg/utils/vault/vault.go
```diff
@@ -0,0 +1,173 @@
...
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	return bytes, nil
+}
+
+// LicensePubKeyPrefix prefixes the given field with the value of the build license public key environment variable.
+func LicensePubKeyPrefix(field string) func() string {
```
pebrc, line 74: This seems very specific to the license generator use case for being in a generic `utils/vault` package?

pkg/utils/vault/vault.go
```diff
@@ -0,0 +1,173 @@
...
+		return nil, fmt.Errorf("%s must be set", vaultAddrEnvVar)
+	}
+	if os.Getenv(vaultTokenEnvVar) == "" {
+		return nil, fmt.Errorf("%s must be set", vaultTokenEnvVar)
+	}
+
+	log.Printf("Read secret %s from vault", f.Path)
+
+	var secret *api.Secret
+	if err := retry.Do(
```
pebrc, line 95: I think we already have our own retry function `UntilSuccess`. Any particular reason not to use it with a timeout?

pkg/utils/vault/vault_test.go
```diff
@@ -0,0 +1,211 @@
...
+
+	// check that the file exists
+	_, err = os.Stat(f.Name)
+	assert.NoError(t, err)
+
+	// overwrite the file
+	err = os.WriteFile(f.Name, []byte(`new_content`), 0600)
+	assert.NoError(t, err)
+
+	// load the file to checlk we read the new content and don't read in vault
```
pebrc, line 49:
```suggestion
	// load the file to check we read the new content and don't read in vault
```

pebrc (commented):
LGTM. We have another vault util in deployer. Can we unify the two? E.g by using the `in-memory` type?

.ci/Dockerfile
thbkrkr: No, this is handled by an environment hook in the Buildkite agent itself, like vault login, whether it's a gcp or k8s agent.

pkg/utils/vault/vault.go
thbkrkr: My pierced memory that has trouble retaining all the good things that are present in this repository :) I will use our `UntilSuccess`.

pkg/utils/vault/vault.go
thbkrkr: Yes, I know but this is a logic shared between the license generator and the test runner.

pkg/utils/vault/vault.go
thbkrkr: Yes!

hack/deployer/runner/ocp3.go
```diff
@@ -1,196 +0,0 @@
-// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
```
thbkrkr, line 1: I created https://github.com/elastic/cloud-on-k8s/issues/6441 to properly clean ocp3 in a follow-up.

pkg/utils/vault/vault.go
thbkrkr: I moved it in a separate file to make it more specific.

thbkrkr:
I moved the code related to vault that was in the deployer into the utils package and refactored the deployer and the helm-releaser in order to they use the new utils/vault package.

pkg/utils/vault/client.go
```diff
@@ -0,0 +1,117 @@
...
+// dev mode and less at CI scenarios, so that users can log in with their vault credentials and deployer will pick up the
+// auth token
+func readCachedToken() (string, error) {
+	dir, err := os.UserHomeDir()
+	if err != nil {
+		return "", err
+	}
+	path := filepath.Join(dir, ".vault-token")
+	_, err = os.Stat(path)
+	if os.IsNotExist(err) {
```
naemono, line 105: https://pkg.go.dev/os#IsNotExist
`This function predates errors.Is. It only supports errors returned by the os package. New code should use errors.Is(err, fs.ErrNotExist).`

pkg/utils/vault/read.go
```diff
@@ -0,0 +1,74 @@
...
+)
+
+// Get fetches contents of a single field at a specified path in Vault
+func (c *Client) Get(secretPath string, fieldName string) (string, error) {
+	result, err := c.GetMany(secretPath, fieldName)
+	if err != nil {
+		return "", err
+	}
+
+	return result[0], nil
```
naemono, line 28: I think `.GetMany` could return a 0 length slice.

naemono (commented):
Mostly lgtm.  Minor things seen.  Not quite done with review.

pkg/utils/vault/read.go
thbkrkr: I don't think it's possible. I will add unit tests.

thbkrkr:
I did one more refactoring to add unit test for `Get` and `GetMany`.

I think it's cleaner like that.

```go
c, err := vault.NewClient()

value, err := vault.Get(c, path, field)

values, err := vault.GetMany(c, path, field1, field2)

bytes, err := vault.ReadFile(c, vault.SecretFile{
	Name:       fieldName,
	Path:       path,
})
```

pkg/utils/vault/client.go
```diff
@@ -0,0 +1,118 @@
...
+		method = "github"
+		data = map[string]interface{}{"token": ghToken}
+	default:
+		var err error
+		token, err = readCachedToken()
+		if err != nil {
+			return errors.Wrap(err, "failed to read cached token")
+		}
+		if token == "" {
+			return fmt.Errorf("set VAULT_TOKEN or VAULT_ROLE_ID/VAULT_SECRET_ID or GITHUB_TOKEN or run `vault login`")
```
naemono, line 77: Probably want to use the constants above.

pkg/utils/vault/client.go
```diff
@@ -0,0 +1,118 @@
...
+		method = "approle"
+		data = map[string]interface{}{"role_id": roleID, "secret_id": secretID}
+	case ghToken != "":
+		method = "github"
+		data = map[string]interface{}{"token": ghToken}
+	default:
+		var err error
+		token, err = readCachedToken()
+		if err != nil {
+			return errors.Wrap(err, "failed to read cached token")
```
naemono, line 74: nit: "while reading cached token"

pkg/utils/vault/client.go
```diff
@@ -0,0 +1,118 @@
...
+		}
+		if token == "" {
+			return fmt.Errorf("set VAULT_TOKEN or VAULT_ROLE_ID/VAULT_SECRET_ID or GITHUB_TOKEN or run `vault login`")
+		}
+	}
+
+	if token == "" {
+		resp, err := c.Logical().Write(fmt.Sprintf("auth/%s/login", method), data)
+		if err != nil {
+			return errors.Wrapf(err, "failed to vault login using method %s", method)
```
naemono, line 84:
```suggestion
			return errors.Wrapf(err, "while logging into vault using method %s", method)
```

pkg/utils/vault/client.go
```diff
@@ -0,0 +1,118 @@
...
+		}
+		if resp.Auth == nil {
+			return fmt.Errorf("failed to vault login: no auth info in response")
```
naemono, line 87:
```suggestion
			return fmt.Errorf("while logging into vault: no auth info in response")
```

pkg/utils/vault/secret_file.go
```diff
@@ -0,0 +1,120 @@
...
+	"fmt"
+	"log"
+	"os"
+	"path/filepath"
+	"time"
+)
+
+const (
+	// inMemory is the name of the file to use to not write it to disk and only keep it in memory
+	inMemory = "in-memory"
```
naemono, line 19: Feels like maybe this should be exported so someone outside of this package can `SecretFile{Name: vault.InMemoryFile}` or something?

Or maybe have `func NewSecretFile` and `func NewInMemorySecretfile` with `inMemory` just being `bool`?

pkg/utils/vault/secret_file.go
```diff
@@ -0,0 +1,120 @@
...
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	return bytes, nil
+}
+
+func (f SecretFile) readFromVault(c Client) ([]byte, error) {
+	log.Printf("Read %s from vault", f.Path)
```
naemono, line 66: Do we want to keep this?

test/e2e/cmd/run/run.go
```diff
@@ -254,28 +255,47 @@ func isOcp3Cluster(h *helper) bool {
...
 	if h.testLicense != "" {
-		bytes, err := os.ReadFile(h.testLicense)
+		bytes, err := vault.ReadFile(c, vault.SecretFile{
+			Name:          h.testLicense,
+			Path:          "test-licenses",
+			FieldResolver: vault.LicensePubKeyPrefix("enterprise"),
+		})
 		if err != nil {
-			return fmt.Errorf("reading %v: %w", h.testLicense, err)
+			return err
```
naemono, line 271: Looking at what errors may be returned from `ReadFile`, I think this should be a more specific wrapped error here for context.  Also the 2 other instances below.

naemono (commented):
Still not seeing anything major here, just a couple minor things.

pkg/utils/vault/secret_file.go
thbkrkr: Yes, we can export it but currently it is not used in other place. We want to keep the flexibility and be able to choose between the 3 modes (don't read, read in mem or read and write) depending on the context:
- in ci: `MY_SECRET_FILE=""` (pr) || `MY_SECRET_FILE=in-memory`
- dev: `MY_SECRET_FILE=""` || `MY_SECRET_FILE=/a/path/to/a/file (to reproduce ci but not get the secret at each try)

pkg/utils/vault/secret_file.go
thbkrkr: I wasn't sure, it's useful but it prints a log line which may seem to come from nowhere. Let's delete it.

naemono (approved):
Let's get this merged to fix the current broken ci. I've seen nothing to prevent merge.

PR #6433, opened 2023-02-20
Make `E2E_JSON` and `E2E_TAGS` usable at runtime

thbkrkr:
This commit changes how to configure the variables `E2E_JSON` and `E2E_TAGS` for the test runner.

Instead of injecting them at build time in the docker image, we let the variables be propagated in the job container and used at runtime by [run.sh](https://github.com/elastic/cloud-on-k8s/blob/6f819294421171759503cf82ca790fdf6aac9048/test/e2e/run.sh#L16-L18).

This gives the ability to build the image once and not 32 times to run nightly e2e-tests.

We already build once on Buildkite but without `eks-arm` which requires `E2E_TAGS="es kb apm ent beat agent"`.
Jenkins jobs are not migrated to take advantage of this.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -486,8 +486,7 @@ E2E_TAGS += $(GO_TAGS)
 export E2E_TAGS
 
 e2e-docker-build: go-generate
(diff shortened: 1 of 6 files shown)


naemono (approved):
makes sense to me.

PR #6434, opened 2023-02-20
Refactoring docker registry creds

thbkrkr:
- Stop writing credentials to disk by moving the vault command to `docker.sh`. The script is already smart enough to authenticate only if necessary (in dev when `CI` is not set).
- Move publishing images in DockerHub to a separate Jenkins job as there is no solution on Buildkite yet for this kind of shared Secret.
- Stop to include `.env` in `.ci/.Makefile` just to get `VAULT_ROOT_PATH` and instead duplicate the if in the Makefile.

### DIFF
--- a/.ci/Makefile
+++ b/.ci/Makefile
@@ -20,9 +20,9 @@ endif
 # BUILD_ID is present during run on Jenkins machine, but not on dev box, hence using it here to distinguish between those cases
 ifndef VAULT_TOKEN
 ifdef BUILD_ID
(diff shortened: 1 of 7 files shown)


naemono:
@thbkrkr when is the new Jenkins job for publishing to dockerhub intended to be run?  What's it's trigger?

thbkrkr:
This is a job to be triggered manually, once the ECK image has been released, by providing a value for the two parameters `ECK_VERSION` and `DRY_RUN`.

PR #6436, opened 2023-02-21
Stack monitoring for Logstash

kaisecheng:
This commit needs to rebase onto https://github.com/elastic/cloud-on-k8s/pull/6404

Following the prior art, the PR integrates stack monitoring for Logstash by adding metricsbeat and filebeat as sidecars to deliver metrics and log to monitoring es cluster

There is a sample resource in `config/samples/logstash/logstash_stackmonitor.yaml` to deploy and test. It will create 
```
> kubectl tree logstash logstash-sample
default    Logstash/logstash-sample                                      -              18m
default    ├─Secret/logstash-sample-default-monitoring-beat-ls-mon-user  -              11m
default    ├─Secret/logstash-sample-ls-config                            -              11m
default    ├─Secret/logstash-sample-ls-monitoring-default-monitoring-ca  -              11m
default    ├─Secret/logstash-sample-ls-monitoring-filebeat-config        -              11m
default    ├─Secret/logstash-sample-ls-monitoring-metricbeat-config      -              11m
default    ├─Service/logstash-sample-ls-api.                             -              11m
default    │ └─EndpointSlice/logstash-sample-ls-http-wfxfs               -              11m
default    └─StatefulSet/logstash-sample-ls                              -              11m
default      ├─ControllerRevision/logstash-sample-ls-5c5dc87d74          -              11m
default      ├─ControllerRevision/logstash-sample-ls-7844b4db48          -              11m
default      ├─ControllerRevision/logstash-sample-ls-84fc6d5f87          -              11m
default      └─Pod/logstash-sample-ls-0                                  True           10m
```

And show status:

```
> kubectl get logstash logstash-sample
NAME              AVAILABLE   EXPECTED   AGE   VERSION
logstash-sample   1           1          35m   8.6.1
```

This PR includes tests

- e2e `TestLogstashStackMonitoring`
- unit test `sidecar_test.go`
- webhook test

Still todo: 

- integrate metrics API with https and username password
- handle monitor that is from external ES cluster


config/samples/logstash/logstash.yaml
```diff
@@ -0,0 +1,27 @@
+apiVersion: elasticsearch.k8s.elastic.co/v1
```
roaksoax, line 1: @kaisecheng should this sample include how to use monitoring? And or, add a new sample that shows how to use monitoring?

roaksoax (commented):
@kaisecheng I think the PR is missing a sample showing how to use monitoring.

test/e2e/logstash/stack_monitoring_test.go
```diff
@@ -0,0 +1,44 @@
...
+
+	// create 1 monitored and 2 monitoring clusters to collect separately metrics and logs
+	metrics := elasticsearch.NewBuilder("test-ls-mon-metrics").
+		WithESMasterDataNodes(2, elasticsearch.DefaultResources)
+	logs := elasticsearch.NewBuilder("test-ls-mon-logs").
+		WithESMasterDataNodes(2, elasticsearch.DefaultResources)
+	monitored := logstash.NewBuilder("test-ls-mon-a").
+		WithNodeCount(1).
+		WithMonitoring(metrics.Ref(), logs.Ref()).
+		//TODO: remove command when Logstash has built with a monitor version of log4j2.properties
```
robbavey, line 35: I'm in two minds as to whether we should include support for log monitoring, given the current lack of upstream support in the docker image to supply a working log4j2.properties file, leading to this requirement to manually add a log4j2.properties file. 

In this case, I get nervous of having this dependency on downloading an upstream logstash log4j2.properties file in order for the e2e tests to pass.

robbavey (commented):
Great stuff! still working through, but I have a couple of questions/observations

test/e2e/logstash/stack_monitoring_test.go
kaisecheng: From the operator side, we don't limit the feature if users have a custom docker image. Customers may already have one in production. I am open to remove log monitoring.

The upstream log4j2.properties is under Logstash team control. I have fewer security concerns here.

robbavey: My concern was more around having a test dependency on an artifact from the `main` branch of a separate codebase, which may result in unintended, and difficult to track down, test failures in this codebase if we ever change the log4j2.properties file in that branch. 

What is the behaviour if the `log:` monitoring stanza is created without the log4j2.properties addition to the container? `filebeat` sidecar gets created but just doesn't do anything?

kaisecheng: Without the file, Logstash won't write log to file, so `filebeat` sideccar is created but does nothing.

Would it help if the download change to `v8.6.1` instead of `main`? The file won't change and the test result should be consistent.

robbavey: I think that's fine for now. I'd like to create a couple of issues to track this - on the logstash side, an issue to track the update to the docker container to support an environment variable to use the file-based log4j2.properties, and a linked `TODO` to to track removing it in this codebase

kaisecheng:
@robbavey I have added an [issue](https://github.com/elastic/logstash/issues/14941) to track file logging in docker on Logstash side. Also, updated this [PR](https://github.com/elastic/cloud-on-k8s/pull/6436/commits/7e38109561aaad8d79ac75e36bb9c3da864eb403) to support env variable $LOG_STYLE for the future. Once upstream adds the log support, operator can use it without upgrade.

kaisecheng:
@pebrc, @barkbay this is ready for review now

pkg/controller/logstash/stackmon/sidecar.go
```diff
@@ -0,0 +1,109 @@
...
+			return nil, err
+		}
+
+		volumes = append(volumes, b.Volumes...)
+		builder.WithContainers(b.Container)
+		configHash.Write(b.ConfigHash.Sum(nil))
+	}
+
+	if monitoring.IsLogsDefined(&logstash) {
+		// enable Stack logging to write Logstash logs to disk
```
robbavey, line 83:
```suggestion
		// Set environment variable to tell Logstash container to write logs to disk
```

pkg/controller/logstash/stackmon/sidecar.go
```diff
@@ -0,0 +1,109 @@
...
+		builder.WithEnv(fileLogStyleEnvVar())
+
+		b, err := Filebeat(ctx, client, logstash)
+		if err != nil {
+			return nil, err
+		}
+
+		// create a logs volume shared between Logstash and Filebeat
+		logsVolume := volume.NewEmptyDirVolume(logstashLogsVolumeName, logstashLogsMountPath)
```
robbavey, line 92: We may want to revisit this at a later point, as this creates an ephemeral logs volume that will go away when the pod disappears. When we start adding persistent volumes we might want to consider adding a logs volume. Maybe a comment or a `TODO:` here to remind us.

test/e2e/logstash/stack_monitoring_test.go
```diff
@@ -0,0 +1,45 @@
...
+import (
+	"github.com/elastic/cloud-on-k8s/v2/pkg/controller/common/stackmon/validations"
+	"github.com/elastic/cloud-on-k8s/v2/test/e2e/test"
+	"github.com/elastic/cloud-on-k8s/v2/test/e2e/test/checks"
+	"github.com/elastic/cloud-on-k8s/v2/test/e2e/test/elasticsearch"
+	"github.com/elastic/cloud-on-k8s/v2/test/e2e/test/logstash"
+	"testing"
+)
+
+// TestESStackMonitoring tests that when an Logstash is configured with monitoring, its log and metrics are
```
robbavey, line 18: Nit: outdated comment

robbavey (commented):
Great stuff! I have a few comments, mostly nits

barkbay (commented):
Still testing, I only left a few nits.

I have some mixed feelings about the fact that the user must use an external `log4j2.properties`. I'm pretty sure some users will want to try this feature soon, and I think it'll be a bit hard to understand why there is no log in ES.

A first step would be to update the documentation, not sure if it's enough though. We could also generate a k8s warning or print something in the operator logs, but as we have no good way to detect if `log4j2.properties` is correctly setup it might be hard to detect in which cases we have to do so.

test/e2e/logstash/stack_monitoring_test.go
```diff
@@ -0,0 +1,45 @@
...
+// correctly delivered to the referenced monitoring Elasticsearch clusters.
+func TestLogstashStackMonitoring(t *testing.T) {
+	// only execute this test on supported version
+	err := validations.IsSupportedVersion(test.Ctx().ElasticStackVersion)
```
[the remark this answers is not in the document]
kaisecheng: Good question! The support [matrix](https://www.elastic.co/support/matrix#matrix_compatibility) says Logstash 8.6.x is compatible with 8.6.x of "monitoring & management Elasticsearch cluster and any Elasticsearch clusters that metrics are routed through". From the test perspective, I think we can safely test it against version 8.0.0+ of elasticsearch.

kaisecheng:
@barkbay 
The issue of logging in upstream Logstash is fixed and backported to 8.7.0.
I have refactored the version validation, updated tests to fulfill min version requirement, updated the sample, and removed comment/todo/doc. It is ready for review.

barkbay (commented):
Sorry for the delay. I only reviewed the e2e test in this review. I'm still running a few other tests and I think I will ✅ soon.

test/e2e/test/logstash/checks.go
```diff
@@ -56,7 +56,10 @@ func CheckStatus(b Builder, k *test.K8sClient) test.Step {
 				AvailableNodes: b.Logstash.Spec.Count,
 				Version:        b.Logstash.Spec.Version,
 			}
-			if logstash.Status != expected {
+
+			if (logstash.Status.ExpectedNodes != expected.ExpectedNodes) ||
+				(logstash.Status.AvailableNodes != expected.AvailableNodes) ||
+				(logstash.Status.Version != expected.Version) {
 				return fmt.Errorf("expected status %+v but got %+v", expected, logstash.Status)
```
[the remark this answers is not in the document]
barkbay: However, it is involved in all cases, whether monitoring is activated or not? I was thinking about something along these lines:

```golang
func CheckStatus(b Builder, k *test.K8sClient) test.Step {
	return test.Step{
		Name: "Logstash status should have the correct status",
....
			expectedAssociationInStatus := len(logstash.Spec.Monitoring.Metrics.ElasticsearchRefs) + len(logstash.Spec.Monitoring.Metrics.ElasticsearchRefs)
			actualAssociationInStatus := len(logstash.Status.MonitoringAssociationStatus)
			if expectedAssociationInStatus != actualAssociationInStatus {
				return fmt.Errorf("expected %d associations in status but got %d", expectedAssociationInStatus, actualAssociationInStatus)
			}
			for a, s := range logstash.Status.MonitoringAssociationStatus {
				if s != v1.AssociationEstablished {
					return fmt.Errorf("association %s has status %s ", a, s)
				}
			}
```

barkbay (commented):
If there are more than 1 monitoring cluster referenced in `metrics` then the status states that all the associations are `Established`, while this configuration is not supported (and therefore, nothing is actually deployed):

```yaml
kind: Logstash
metadata:
  annotations:
    association.k8s.elastic.co/es-conf-272414756: '{....}'
    association.k8s.elastic.co/es-conf-3519008756: '{....}'
    association.k8s.elastic.co/es-conf-4222863958: '{....}'
  creationTimestamp: "2023-03-21T08:07:05Z"
  generation: 1
  labels:
    test-name: test-ls-mon-a
  name: test-ls-mon-a-5ptv
  namespace: e2e-mercury
  resourceVersion: "757992"
  uid: 74383387-d449-459f-a3fb-07279bef0368
spec:
  count: 1
  monitoring:
    logs:
      elasticsearchRefs:
      - name: test-ls-mon-logs-lp4x
        namespace: e2e-mercury
    metrics:
      elasticsearchRefs:
      - name: test-ls-mon-metrics1-rr2c
        namespace: e2e-mercury
      - name: test-ls-mon-metrics2-zmvs
        namespace: e2e-mercury
  podTemplate: {}
  version: 8.7.0-SNAPSHOT
status:
  monitoringAssociationStatus:
    e2e-mercury/test-ls-mon-logs-lp4x: Established
    e2e-mercury/test-ls-mon-metrics1-rr2c: Established
    e2e-mercury/test-ls-mon-metrics2-zmvs: Established
  observedGeneration: 1
```

I'm not sure this comes from your PR. Would you mind opening an issue if it's not the case as it seems wrong and misleading? Note that I was not running the webhook, which I think would have prevented this from happening, still looks wrong anyway.

pkg/apis/logstash/v1alpha1/validations.go
```diff
@@ -7,33 +7,42 @@ package v1alpha1
 import (
 	"k8s.io/apimachinery/pkg/util/validation/field"
 
+	"github.com/elastic/cloud-on-k8s/v2/pkg/controller/common/stackmon/validations"
+
 	commonv1 "github.com/elastic/cloud-on-k8s/v2/pkg/apis/common/v1"
 	"github.com/elastic/cloud-on-k8s/v2/pkg/controller/common/version"
```
barkbay, line 13:
```suggestion
	commonv1 "github.com/elastic/cloud-on-k8s/v2/pkg/apis/common/v1"
	"github.com/elastic/cloud-on-k8s/v2/pkg/controller/common/stackmon/validations"
	"github.com/elastic/cloud-on-k8s/v2/pkg/controller/common/version"
```

barkbay (commented):
Almost LGTM, left another comment on the tests.

barkbay (approved):
LGTM

test/e2e/test/k8s_client.go
```diff
@@ -383,10 +383,14 @@ func (k K8sClient) CreateOrUpdate(objs ...client.Object) error {
...
+	return StackPodListOptions(esNamespace, esName, label.Type)
+}
+
+func StackPodListOptions(namespace, name, typeLabel string) []k8sclient.ListOption {
+	ns := k8sclient.InNamespace(namespace)
 	matchLabels := k8sclient.MatchingLabels(map[string]string{
-		commonv1.TypeLabelName:     label.Type,
-		label.ClusterNameLabelName: esName,
+		commonv1.TypeLabelName:     typeLabel,
+		label.ClusterNameLabelName: name,
```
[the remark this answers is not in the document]
kaisecheng: Reverted the changes of CheckBeatSidecars and renamed it as suggested

test/e2e/test/logstash/checks.go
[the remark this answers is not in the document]
kaisecheng: I see. You are right. Added it to status check. Thank you!

kaisecheng:
> 
> I'm not sure this comes from your PR. Would you mind opening an issue if it's not the case as it seems wrong and misleading? Note that I was not running the webhook, which I think would have prevented this from happening, still looks wrong anyway.

@barkbay Create an issue for tracking https://github.com/elastic/cloud-on-k8s/issues/6558. I could not replicate the problem locally. Please update the issue if you have steps to reproduce it.

Issue #6369, opened 2023-02-01
Optional client authentication should be supported


barkbay:
Following up https://github.com/elastic/cloud-on-k8s/issues/5817 I think we should not print a warning if `xpack.security.http.ssl.client_authentication` is set to `optional` as I would expect the probes and operator to still work when this value is used. We should also update the documentation accordingly.

PR #6440, opened 2023-02-22
Allow optional SSL client authentication setting

rhr323:
Elasticsearch supports a few different [settings for client authentication settings for https](https://www.elastic.co/guide/en/elasticsearch/reference/current/security-settings.html). This PR allows `xpack.security.http.ssl.client_authentication
` to be set to either `optional` or `none`, but it will keep emitting an unsupported warning in case it is set to `required`.

Fixes: https://github.com/elastic/cloud-on-k8s/issues/6369


pkg/controller/elasticsearch/validation/warnings.go
```diff
@@ -26,10 +37,35 @@ func noUnsupportedSettings(es esv1.Elasticsearch) field.ErrorList {
...
+	unsupported := config.HasKeys(esv1.UnsupportedSettings)
+	for _, setting := range unsupported {
+		errs = append(errs, field.Forbidden(field.NewPath("spec").Child("nodeSets").Index(index).Child("config").Child(setting), unsupportedConfigErrMsg))
+	}
+	return errs
+}
+
+func validateClientAuthentication(config *common.CanonicalConfig, index int) field.ErrorList {
+	type ClientAuthSetting struct {
+		Value string `config:"xpack.security.http.ssl.client_authentication"`
```
rhr323, line 57: Let me know if there is an easier / generic way to access a setting given the key as string, so we don't need to define a struct for it.

pkg/controller/elasticsearch/validation/warnings.go
```diff
@@ -13,9 +13,20 @@ import (
...
+
 func noUnsupportedSettings(es esv1.Elasticsearch) field.ErrorList {
+	return checkNodeSets(es, validateSettings)
+}
+
+func validClientAuthentication(es esv1.Elasticsearch) field.ErrorList {
+	return checkNodeSets(es, validateClientAuthentication)
+}
+
+func checkNodeSets(es esv1.Elasticsearch, check nodeSetChecker) field.ErrorList {
```
rhr323, line 29: My attempt to reduce doing the same iteration in both validations. Let me know if there's a better way.

docs/orchestrating-elastic-stack-applications/elasticsearch/reserved-settings.asciidoc
```diff
@@ -22,7 +22,6 @@ The following Elasticsearch settings are managed by ECK:
 * `xpack.security.authc.reserved_realm.enabled`
 * `xpack.security.enabled`
 * `xpack.security.http.ssl.certificate`
-* `xpack.security.http.ssl.client_authentication`
```
rhr323, line 25: I wonder if it's best to leave this our from the list of ECK managed settings, or if we should explain what values are supported / not supported.

pkg/controller/elasticsearch/validation/warnings.go
thbkrkr: Better is pretty relative here :) Your approach is good. I think I would just have added the new validation in the existing `noUnsupportedSettings` func.

Maybe create two functions:
```go
func noUnsupportedSettings(es esv1.Elasticsearch) field.ErrorList {
	...
		errs = append(errs, validUnsupportedSettings(config)...)
		errs = append(errs, validPartiallyUnsupportedSettings(config)...)
	}
	return errs
}

func validUnsupportedSettings(config *common.CanonicalConfig) field.ErrorList
func validPartiallyUnsupportedSettings(config *common.CanonicalConfig) field.ErrorList
```

pkg/controller/elasticsearch/validation/validations.go
```diff
@@ -24,23 +24,24 @@ import (
...
+	parseStoredVersionErrMsg           = "Cannot parse current Elasticsearch version. String format must be {major}.{minor}.{patch}[-{label}]"
+	parseVersionErrMsg                 = "Cannot parse Elasticsearch version. String format must be {major}.{minor}.{patch}[-{label}]"
+	pvcImmutableErrMsg                 = "volume claim templates can only have their storage requests increased, if the storage class allows volume expansion. Any other change is forbidden"
+	pvcNotMountedErrMsg                = "volume claim declared but volume not mounted in any container. Note that the Elasticsearch data volume should be named 'elasticsearch-data'"
+	unsupportedConfigErrMsg            = "Configuration setting is reserved for internal use. User-configured use is unsupported"
+	unsupportedUpgradeMsg              = "Unsupported version upgrade path. Check the Elasticsearch documentation for supported upgrade paths."
+	unsupportedVersionMsg              = "Unsupported version"
+	notAllowedNodesLabelMsg            = "Node label not in the exposed node labels list"
+	unsupportedClientAuthenticationMsg = "Mandatory client authentication is not supported"
```
thbkrkr, line 44: I think this can be confusing and it would be good to be more specific.
Something like `Configuration setting 'xpack.security.http.ssl.client_authentication: required' is unsupported`.

```go
func partiallyUnsupportedConfigErrMsg(setting, value string) string {
	return fmt.Sprintf("Configuration setting `%s: %s` is unsupported", setting, value)
}
```

pkg/controller/elasticsearch/validation/warnings.go
```diff
@@ -26,10 +37,35 @@ func noUnsupportedSettings(es esv1.Elasticsearch) field.ErrorList {
...
+		Value string `config:"xpack.security.http.ssl.client_authentication"`
+	}
+	forbiddenValue := "required" // we allow 'none' and 'optional' but 'required' is not supported
+
+	var errs field.ErrorList
+	var setting ClientAuthSetting
+	if err := config.Unpack(&setting); err != nil {
+		return errs
+	}
+	if setting.Value == forbiddenValue {
```
thbkrkr, line 66: Maybe a bit overkill as we only have one setting and one forbidden value but we could create a map in `elasticsearch/v1/fields.go`:

```go
var PartiallyUnsupportedSettings = map[string][]string{
	XPackSecurityHttpSslClientAuthentication: {"required"}, // "optional" or "none" are allowed
}
```

And iterate:

```go
func validPartiallyUnsupportedSettings(config *common.CanonicalConfig) field.ErrorList {
	var errs field.ErrorList
	for setting, unsupportedVals := range esv1.PartiallyUnsupportedSettings {
		value, _ := config.String(setting)
		for _,  := range unsupportedVals {
			if value == unsupportedVal {
				errs = append(errs, field.Invalid(field.NewPath("spec").Child("nodeSets").Index(i).Child("config").Child(setting),
					value, partiallyUnsupportedConfigErrMsg(setting, unsupportedVal)))
			}
		}

	}
	return errs
}
```

docs/orchestrating-elastic-stack-applications/elasticsearch/reserved-settings.asciidoc
thbkrkr: +1 to explain what values are supported / not supported. It doesn't seem to be a big effort to maintain and that shouldn't be something that changes a lot.

Not sure about the wording:
* `xpack.security.http.ssl.client_authentication` (supported for `none` or `optional` but not `required`)

pkg/controller/elasticsearch/validation/warnings.go
pebrc: Does this resolve path separators correctly ? I.e. `xpack.security.http...` I don't remember.

docs/orchestrating-elastic-stack-applications/elasticsearch/reserved-settings.asciidoc
pebrc: I would probably add new section "Partially restricted settings" and say `xpack.security.http.ssl.client_authentication` (supported values `none`, `optional`)

docs/orchestrating-elastic-stack-applications/elasticsearch/reserved-settings.asciidoc
```diff
@@ -22,12 +22,15 @@ The following Elasticsearch settings are managed by ECK:
...
 * `xpack.security.http.ssl.enabled`
 * `xpack.security.http.ssl.key`
 * `xpack.security.transport.ssl.certificate`
 * `xpack.security.transport.ssl.enabled`
 * `xpack.security.transport.ssl.key`
 * `xpack.security.transport.ssl.verification_mode`
 
+The following Elasticsearch settings are not supported by ECK:
+
+ * `xpack.security.http.ssl.client_authentication`: `required`
```
pebrc, line 34: This is also good 👍

pkg/controller/elasticsearch/validation/warnings.go
rhr323: Looks like it works as expected! I added the wrapper and a few basic test cases.

pkg/controller/elasticsearch/validation/warnings.go
rhr323: Makes sense, thanks for the suggestion. I made the changes, I think it's more readable now.

pkg/controller/elasticsearch/validation/warnings.go
rhr323: Nice! I was thinking about the same thing initially 😄. We should probably have something like this in case we encounter another similar setting. But yeah, for one setting only, maybe the simpler is better (for now).

pkg/controller/elasticsearch/validation/validations.go
thbkrkr: I forgot that we display `Spec.nodesets [0] .config.xpack.security.http.ssl.client_authentication` with the message. Without this information, it would have been confused.
It's good like that, don't change anything.

thbkrkr (approved):
LGTM

pebrc (approved):
Nice work!

PR #6450, opened 2023-02-23
Remove migrated jobs jenkins

thbkrkr:
- delete migrated jobs in buildkite
- move parameters in the jjbb definition for publish-dockerhub job
- redirect setenvconfig stdout in build (follow-up of #6445)


naemono (approved):
seems good to me.

PR #6480, opened 2023-03-03
Logstash pipelines

kaisecheng:
This PR allows configuration of Logstash `pipeline.yml` in CRD. `pipeline.yml` is an array of map of string/interface{} defining multiple pipelines. `.` in between the key is treated as string.

```yml
- pipeline.id: stdin_stdout
  config.string: input { stdin {} } output { stdout{} }
- pipeline.id: demo
  pipeline.workers: 1
  path.config: "/usr/share/logstash/pipeline"
```
Pipelines can be set from either inline in `Pipelines` or secret referencing in`PipelinesRef`.

This PR includes tests

- e2e `TestPipelineConfigRefLogstash`, `TestPipelineConfigLogstash`
- unit test


pkg/apis/logstash/v1alpha1/logstash_types.go
```diff
@@ -39,6 +39,17 @@ type LogstashSpec struct {
 	// +kubebuilder:validation:Optional
 	ConfigRef *commonv1.ConfigSource `json:"configRef,omitempty"`
 
+	// Pipelines holds the Logstash Pipelines. At most one of [`Pipelines`, `PipelinesRef`] can be specified.
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:pruning:PreserveUnknownFields
+	Pipelines []map[string]string `json:"pipelines,omitempty"`
```
robbavey, line 45: We might need to rethink this a little - using a `[]map[string]string` here requires the values of numerical values in the yaml to be a string - 
eg

`pipeline.workers: 4` is invalid, and users must specify
`pipeline.workers: "4"`

and requires a flattened yaml structure - 

```
`pipeline:
    id: "demo"
    workers: "4"
```
will also fail due to the requirement for a single string.

From [this](https://github.com/kubernetes-sigs/controller-tools/issues/636) issue, `map[string]interface{}` doesn't appear to be allowed in CRD definitions - there are a couple of workarounds described there that might work for us, and we should investigate. Also, would defining this as a `[]commonv1.Config` work?

robbavey (changes requested):
Good stuff, but I do have concerns on the inline pipeline definition - the current implementation requires the pipeline yml to be defined as `<flattened_key>: <string_value>`, rather than straight yaml, and requires the individual types to be strings too, not integers. This is probably something we should try and fix

pkg/controller/logstash/pipeline.go
```diff
@@ -0,0 +1,79 @@
...
+// getUserPipeline extracts the pipeline either from the spec `pipeline` field or from the Secret referenced by spec
+// `pipelineRef` field.
+func getUserPipeline(params Params) (*PipelinesConfig, error) {
+	if params.Logstash.Spec.Pipelines != nil {
+		return NewPipelinesConfigFrom(params.Logstash.Spec.Pipelines)
+	}
+	return ParseConfigRef(params, &params.Logstash, params.Logstash.Spec.PipelinesRef, PipelineFileName)
+}
+
+func defaultPipeline() *PipelinesConfig {
```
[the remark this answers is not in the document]
kaisecheng: I think the default pipeline should be the same as docker and believe should keep the pipeline in Secret to have similar behavior as other configs.
Changed the pipeline to the following
```yaml
- "pipeline.id": "main"
  "path.config": "/usr/share/logstash/pipeline"
```

pkg/apis/logstash/v1alpha1/logstash_types.go
kaisecheng: Many thanks for raising this issue. I changed Pipelines to `[]commonv1.Config`, which can handle integer, string and boolean. I was trying to use `instr.IntOrString` but it fails to handle `true` as string and required `"true"`.

Logstash `pipelines.yml` does not take `pipeline.workers` as a flattened yaml structure 😕 It has to be a `.` in-between, otherwise raises error. Check [here](https://github.com/elastic/logstash/blob/main/logstash-core/lib/logstash/settings.rb#L45-L77) and [here](https://github.com/elastic/logstash/blob/main/logstash-core/lib/logstash/settings.rb#L168).
That's why I have to make [pipelines_config](https://github.com/elastic/cloud-on-k8s/pull/6480/files#diff-afedbf46a40edef9976bb3e3164244742b123e8c4f281242ec886e2769c6882cR22) to not flatten the keys as [canonical_config](https://github.com/elastic/cloud-on-k8s/blob/main/pkg/controller/common/settings/canonical_config.go#L25)

test/e2e/test/logstash/checks.go
```diff
@@ -65,29 +85,63 @@ func CheckStatus(b Builder, k *test.K8sClient) test.Step {
...
+				Name: "metrics",
+				Path: "/",
+			},
+			Want{
+				Status: "green",
+			}),
+		b.CheckMetricsRequest(k,
+			Request{
+				Name: "default pipeline",
+				Path: "/_node/pipelines/demo",
```
robbavey, line 100:
```suggestion
				Path: "/_node/pipelines/main",
```

This needs to be `main` to match the value in the samples file

robbavey (changes requested):
Mostly nits, but the `TestSample` e2e test appears to be currently broken

robbavey:
Something I think we'll need to figure out before GA is pipeline reloads. If we change a pipeline definition in `pipelines` or a secret pointed to in `pipelineRef`, it will trigger a complete restart of the pod, rather than reloading the pipeline - defining the pipeline in a volume mount pointed to by `path.config` enables individual pipelines to be reloaded if `config.reload.automatic` is enabled.

From a little digging, there are some complications - `secrets` or `configMaps` volumes defined as `subPath` will [not receive updates](https://kubernetes.io/docs/concepts/storage/volumes/#secret). This means that `pipelines.yml` as it is currently implemented will not be updated inside the container even if the definition is changed in either the secret, or inline in the logstash CRD. 

We may need to figure out a way to use `NewSecretVolumeWithMountPath` to mount the config directory and write the configs into there, rather than rely on the `subPath`. If we do that, we can choose to not update the [configHash](https://github.com/elastic/cloud-on-k8s/pull/6480/files#diff-19649d7d5b030ee703218ba84cdc1aac6595195e515a7f3f81ee11da419e3bd6R42) when secrets are updated to avoid triggering a restart on pipeline changes.

I don't think this should block this PR, but we should add a separate issue to deal with pipeline reloads

kaisecheng:
> We may need to figure out a way to use `NewSecretVolumeWithMountPath` to mount the config directory and write the configs into there, rather than rely on the `subPath`. If we do that, we can choose to not update the [configHash](https://github.com/elastic/cloud-on-k8s/pull/6480/files#diff-19649d7d5b030ee703218ba84cdc1aac6595195e515a7f3f81ee11da419e3bd6R42) when secrets are updated to avoid triggering a restart on pipeline changes.
> 
> I don't think this should block this PR, but we should add a separate issue to deal with pipeline reloads

I have a solution in mind similar to setup keystore with init container or kibana config init container.
Logstash will need a config init container, map Secrets (config.yml, pipelines.yml) to directories and map EmptyDir to `config-local` directory, and then make soft links from secret dir to `config-local`. In the logstash container, map `config-local` to `/usr/share/logstash/config/`. Hopefully, soft links get updates when secrets are updated.

pebrc:
> We may need to figure out a way to use NewSecretVolumeWithMountPath to mount the config directory and write the configs into there, rather than rely on the subPath. If we do that, we can choose to not update the [configHash](https://github.com/elastic/cloud-on-k8s/pull/6480/files#diff-19649d7d5b030ee703218ba84cdc1aac6595195e515a7f3f81ee11da419e3bd6R42) when secrets are updated to avoid triggering a restart on pipeline changes.

This should be the preferred way. Mounting the whole Secret without `subPath` will propagate changes automatically. Not adding the contents to the config hash is the way to go if Logstash can auto-reload the pipelines. This works as long as pipeline config can be in a separate directory on the container (which I think it can judging from Logstash docs) so that you can mount the whole secret without any projections. 

> I have a solution in mind similar to setup keystore with init container or kibana config init container.
Logstash will need a config init container, map Secrets (config.yml, pipelines.yml) to directories and map EmptyDir to config-local directory, and then make soft links from secret dir to config-local

If you can I would avoid that because of the added complexity. We are only doing that for Elasticsearch for example because we want to preserve some of the configuration that comes with the Docker image, so we selectively soft-link additional configuration into the Elasticsearch configuration directory. That is because Elasticsearch does not support additional configuration directories. IIUC you have that for Logstash pipelines which can be in a dedicated directory so you can just mount on top of that.

robbavey:
Unfortunately, I believe we have the same issues as Elasticsearch with the config directory. Logstash expects `pipelines.yml` to be in the same config directory as `logstash.yml` (which we also create in the operator from `config`/`configRef` in the CRD), and `log4j2.properties` and `jvm.options`, which we source from the docker image.

`pipelines.yml`, which is what we are creating here, contains settings about the pipelines that logstash is running (number of workers, batch sizes, pq information, etc) and either the pipeline definitions themselves, or a pointer to where the pipeline definitions can be found. A change to either a pipeline, _or its settings_, should trigger a reload of that pipeline, and that pipeline only

pkg/controller/logstash/pipeline_test.go
```diff
@@ -0,0 +1,132 @@
...
+		},
+		{
+			name: "pipeline populated",
+			pipelines: []commonv1.Config{
+				{Data: map[string]interface{}{"pipeline.id": "main"}},
+			},
+			want: MustParsePipelineConfig([]byte(`- "pipeline.id": "main"`)),
+		},
+		{
+			name: "configref populated - no secret",
```
robbavey, line 52: Nit:
```suggestion
			name: "pipelinesref populated - no secret",
```

pkg/controller/logstash/pipeline_test.go
```diff
@@ -0,0 +1,132 @@
...
+				SecretRef: commonv1.SecretRef{
+					SecretName: "my-secret-pipeline",
+				},
+			},
+			client:  k8s.NewFakeClient(),
+			want:    NewPipelinesConfig(),
+			wantErr: true,
+		},
+		{
+			name: "configref populated - no secret key",
```
robbavey, line 63: Nit: 
```suggestion
			name: "pipelinesref populated - no secret key",
```

pkg/controller/logstash/pipeline_test.go
```diff
@@ -0,0 +1,132 @@
...
+			client: k8s.NewFakeClient(&corev1.Secret{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "my-secret-pipeline",
+				},
+			}),
+			want:    NewPipelinesConfig(),
+			wantErr: true,
+		},
+		{
+			name: "configref populated - malformed config",
```
robbavey, line 78: Nit: 
```suggestion
			name: "pipelinesref populated - malformed config",
```

pkg/controller/logstash/pipeline_test.go
```diff
@@ -0,0 +1,132 @@
...
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "my-secret-pipeline-2",
+				},
+				Data: map[string][]byte{"pipelines.yml": []byte("something:bad:value")},
+			}),
+			want:    NewPipelinesConfig(),
+			wantErr: true,
+		},
+		{
+			name: "configref populated",
```
robbavey, line 94: Nit: 
```suggestion
			name: "pipelinesref populated",
```

robbavey (changes requested):
Looks great, I have a few nits, plus we'll need to add a task/follow up for less invasive pipeline reloads.

kaisecheng:
added an issue to follow up pipeline reload https://github.com/elastic/ingest-dev/issues/1661

pkg/controller/logstash/config.go
```diff
@@ -63,7 +63,7 @@ func buildConfig(params Params) ([]byte, error) {
 }
 
 // getUserConfig extracts the config either from the spec `config` field or from the Secret referenced by spec
-// `configRef` field.
+// `pipelinesRef` field.
```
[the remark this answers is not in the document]
robbavey: Are you sure? The code references `params.Logstash.Spec.ConfigRef`, not `pipelinesRef` - [this](https://github.com/elastic/cloud-on-k8s/pull/6480/files#diff-19649d7d5b030ee703218ba84cdc1aac6595195e515a7f3f81ee11da419e3bd6R62) appears to be the `pipelinesRef` source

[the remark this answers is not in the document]
kaisecheng: oh no 🤦 I thought it is pipeline_config.go. You are right

test/e2e/test/logstash/checks.go
```diff
@@ -83,29 +103,63 @@ func CheckStatus(b Builder, k *test.K8sClient) test.Step {
...
+			}),
+	}
+}
 
-				if status.Status != "green" {
-					return fmt.Errorf("expected green but got %s", status.Status)
+func (b Builder) CheckMetricsRequest(k *test.K8sClient, req Request, want Want) test.Step {
+	return test.Step{
+		Name: fmt.Sprintf("Logstash should respond to %s requests", req.Name),
+		Test: func(t *testing.T) {
```
robbavey, line 130: I think we should change this to a `test.Eventually` block, so it will be retried if the endpoint isn't up and running straight away

```

			// send request and parse to map obj
			client, err := NewLogstashClient(b.Logstash, k)
			if err != nil {
				return err
			}

			bytes, err := DoRequest(client, b.Logstash, "GET", req.Path)
			if err != nil {
				return err
			}

			var response map[string]interface{}
			err = json.Unmarshal(bytes, &response)
			if err != nil {
				return err
			}

			// parse response to ucfg.Config for traverse
			res, err := settings.NewCanonicalConfigFrom(response)
			if err != nil {
				return err
			}

			// check status
			status, err := res.String("status")
			if err != nil {
				return err
			}
			if status != want.Status {
				fmt.Errorf("expected %s but got %s", want.Status, status)
			}

			// check expected string
			for k, v := range want.Match {
				str, err := res.String(k)
				if err != nil {
					return err
				}
				if str != v {
					fmt.Errorf("expected %s to be %s but got %s", k, v, str)
				}
			}
			return nil
		}),
```

robbavey (commented):
I'm happy with this from the Logstash side - just a request to add some more robustness when testing the endpoints in the samples

pkg/controller/logstash/pipeline.go
```diff
@@ -0,0 +1,83 @@
...
+	pipelines := []map[string]string{
+		{
+			"pipeline.id": "main",
+			"path.config": "/usr/share/logstash/pipeline",
+		},
+	}
+
+	return MustPipelinesConfig(pipelines)
+}
```
barkbay, line 83: nit: could we use a `var` instead?
```suggestion
var (
	defaultPipeline = MustPipelinesConfig([]map[string]string{
		{
			"pipeline.id": "main",
			"path.config": "/usr/share/logstash/pipeline",
		},
	})
)
```

pkg/apis/logstash/v1alpha1/logstash_types.go
```diff
@@ -42,6 +42,17 @@ type LogstashSpec struct {
 	// +kubebuilder:validation:Optional
 	ConfigRef *commonv1.ConfigSource `json:"configRef,omitempty"`
 
+	// Pipelines holds the Logstash Pipelines. At most one of [`Pipelines`, `PipelinesRef`] can be specified.
```
barkbay, line 45: Would it make sense to add a link to https://www.elastic.co/guide/en/logstash/current/multiple-pipelines.html to explain the expected format for each pipeline? (I'm a complete noob wrt Logstash, I was just trying to understand where the format was documented, so feel free to ignore this comment 😄 )

test/e2e/test/logstash/builder.go
```diff
@@ -116,6 +117,38 @@ func (b Builder) WithServices(services ...logstashv1alpha1.LogstashService) Buil
...
+				VolumeMounts: mounts,
+			},
+		}
+	} else {
+		if b.Logstash.Spec.PodTemplate.Spec.Containers[0].VolumeMounts == nil {
+			b.Logstash.Spec.PodTemplate.Spec.Containers[0].VolumeMounts = []corev1.VolumeMount{}
+		}
+		b.Logstash.Spec.PodTemplate.Spec.Containers[0].VolumeMounts = append(b.Logstash.Spec.PodTemplate.Spec.Containers[0].VolumeMounts, mounts...)
+	}
+	return b
```
barkbay, line 149: nit: reduce nesting with early return
```suggestion
	if b.Logstash.Spec.PodTemplate.Spec.Containers == nil {
		b.Logstash.Spec.PodTemplate.Spec.Containers = []corev1.Container{
			{
				Name:         "logstash",
				VolumeMounts: mounts,
			},
		}
		return b
	}

	if b.Logstash.Spec.PodTemplate.Spec.Containers[0].VolumeMounts == nil {
		b.Logstash.Spec.PodTemplate.Spec.Containers[0].VolumeMounts = []corev1.VolumeMount{}
	}
	b.Logstash.Spec.PodTemplate.Spec.Containers[0].VolumeMounts = append(b.Logstash.Spec.PodTemplate.Spec.Containers[0].VolumeMounts, mounts...)

	return b
```

barkbay (commented):
first review, I'll do another one and some tests soonish

pkg/apis/logstash/v1alpha1/logstash_types.go
```diff
@@ -42,6 +42,17 @@ type LogstashSpec struct {
...
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:pruning:PreserveUnknownFields
+	Pipelines []commonv1.Config `json:"pipelines,omitempty"`
+
+	// PipelinesRef contains a reference to an existing Kubernetes Secret holding the Logstash Pipelines.
+	// Logstash pipelines must be specified as yaml, under a single "pipeline.yml" entry. At most one of [`Pipelines`, `PipelinesRef`]
```
barkbay, line 51:
```suggestion
	// Logstash pipelines must be specified as yaml, under a single "pipelines.yml" entry. At most one of [`Pipelines`, `PipelinesRef`]
```

test/e2e/logstash/pipeline_test.go
```diff
@@ -0,0 +1,141 @@
...
+	steps := test.StepsFunc(func(k *test.K8sClient) test.StepList {
+		return test.StepList{
+			b.CheckMetricsRequest(k,
+				logstash.Request{
+					Name: "pipeline [generator]",
+					Path: "/_node/pipelines/generator",
+				},
+				logstash.Want{
+					Status: "green",
+				}),
```
barkbay, line 65: nit, out of curiosity: why are we not checking `pipelines.generator.workers` as it is the case for the default pipeline here: https://github.com/elastic/cloud-on-k8s/pull/6480/files#diff-4c2ec435669b8407b1fef39e9ce03a145ae797732051e4bbcb42caf136d975c6R117-R120

```suggestion
				logstash.Want{
					Status: "green",
					Match:  map[string]string{"pipelines.generator.workers": "1"},
				}),
```

test/e2e/logstash/pipeline_test.go
```diff
@@ -0,0 +1,141 @@
...
+	steps := test.StepsFunc(func(k *test.K8sClient) test.StepList {
+		return test.StepList{
+			b.CheckMetricsRequest(k,
+				logstash.Request{
+					Name: "pipeline [split]",
+					Path: "/_node/pipelines/split",
+				},
+				logstash.Want{
+					Status: "green",
+				}),
```
barkbay, line 136: Same here?
```suggestion
				logstash.Want{
					Status: "green",
					Match:  map[string]string{"pipelines.split.workers": "2"},
				}),
```

test/e2e/logstash/pipeline_test.go
```diff
@@ -0,0 +1,141 @@
...
+	corev1 "k8s.io/api/core/v1"
+	"testing"
+
+	commonv1 "github.com/elastic/cloud-on-k8s/v2/pkg/apis/common/v1"
+	"github.com/elastic/cloud-on-k8s/v2/test/e2e/test"
+	"github.com/elastic/cloud-on-k8s/v2/test/e2e/test/logstash"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestPipelineConfigRefLogstash(t *testing.T) {
```
barkbay, line 19: nit: add a go doc to explain the purpose of this test

test/e2e/logstash/pipeline_test.go
```diff
@@ -0,0 +1,141 @@
...
+				logstash.Want{
+					Status: "green",
+				}),
+		}
+	})
+
+	test.Sequence(before, steps, b).RunSequential(t)
+}
+
+func TestPipelineConfigLogstash(t *testing.T) {
```
barkbay, line 72: nit: add a go doc to explain the purpose of this test

pkg/apis/logstash/v1alpha1/logstash_types.go
kaisecheng: Probably not, just like `kibana.yml` and `elasticsearch.yml` expect users to have basic context. Linking the pipeline-to-pipeline page may confuse them.

test/e2e/logstash/pipeline_test.go
```diff
@@ -0,0 +1,141 @@
...
+	secretName := "ls-split-pipe"
+
+	pipelineSecret := corev1.Secret{
```
barkbay, line 75: We should eventually delete this `Secret`.

test/e2e/logstash/pipeline_test.go
```diff
@@ -0,0 +1,141 @@
...
+	commonv1 "github.com/elastic/cloud-on-k8s/v2/pkg/apis/common/v1"
+	"github.com/elastic/cloud-on-k8s/v2/test/e2e/test"
+	"github.com/elastic/cloud-on-k8s/v2/test/e2e/test/logstash"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestPipelineConfigRefLogstash(t *testing.T) {
+	secretName := "ls-generator-pipeline"
+
+	pipelineSecret := corev1.Secret{
```
barkbay, line 22: We should eventually delete this `Secret`.

pkg/controller/logstash/pipelines_config.go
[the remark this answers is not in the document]
kaisecheng: Personally, having prefix "Pipeline" helps me to differentiate PipelineConfig, Config, and CanonicalConfig. I am more into keeping it and staying in the same package. @robbavey Any thoughts?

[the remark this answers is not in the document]
barkbay: > Personally, having prefix "Pipeline" helps me to differentiate PipelineConfig, Config

Outside of the package they would be prefixed by the package name, `PipelineConfig` would be `pipelines.Config`. I must confess that all these `pipelines_*` files seem a bit odd to me and a hint that they should be in a same package.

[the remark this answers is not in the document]
robbavey: I think splitting it out into a package makes sense - in code outside of the package, the method names will be prefixed by the package name `pipelines.fromConfig`, for example

test/e2e/logstash/pipeline_test.go
kaisecheng: Good catch! should check pipeline.* to ensure the pipeline works. The default number of workers is the number of CPU cores. I am going to replace `pipelines.*.workers: 2` with `pipelines.*.batch_size: 125`, which is a static number from `logstash.yml`

pkg/controller/logstash/pipelines_config.go
```diff
@@ -0,0 +1,134 @@
...
+			diff = append(diff, fmt.Sprintf("%s vs %s, ", m, m2))
+		}
+	}
+
+	if len(diff) > 0 {
+		return true, fmt.Errorf("there are %d differences. %s", len(diff), diff)
+	}
+
+	return false, nil
+}
```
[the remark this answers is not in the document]
kaisecheng: `pipelines_*.go` has moved to package. I put back these test functions as two test files in different packages referencing them.

pkg/controller/logstash/pipelines/configref.go
```diff
@@ -0,0 +1,37 @@
...
+)
+
+// ConfigRefWatchName returns the name of the watch registered on the secret referenced in `pipelinesRef`.
+func ConfigRefWatchName(resource types.NamespacedName) string {
+	return fmt.Sprintf("%s-%s-pipelinesref", resource.Namespace, resource.Name)
+}
+
+// ParseConfigRef retrieves the content of a secret referenced in `pipelinesRef`, sets up dynamic watches for that secret,
+// and parses the secret content into a PipelinesConfig.
+func ParseConfigRef(
```
robbavey, line 25: Nit:

```suggestion
func ParsePipelinesRef(
```

robbavey (commented):
Great stuff, just a couple of nits/suggestions around clarifying naming around `PipelinesRef` and `ConfigRef`

pkg/controller/logstash/pipelines/configref.go
```diff
@@ -0,0 +1,37 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+
+	commonv1 "github.com/elastic/cloud-on-k8s/v2/pkg/apis/common/v1"
+	"github.com/elastic/cloud-on-k8s/v2/pkg/controller/common"
+	"github.com/elastic/cloud-on-k8s/v2/pkg/controller/common/driver"
+)
+
+// ConfigRefWatchName returns the name of the watch registered on the secret referenced in `pipelinesRef`.
+func ConfigRefWatchName(resource types.NamespacedName) string {
```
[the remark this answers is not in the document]
kaisecheng: lint suggested not to use `Pipelines` prefix. Renamed to `ref.go` and `RefWatchName()`

kaisecheng:
@barkbay I think it is ready 2nd round

barkbay (approved):
LGTM! Nice work 👍

robbavey:
https://github.com/elastic/cloud-on-k8s/pull/6674 adds pipeline reload functionality

PR #6482, opened 2023-03-03
Fix handling of unmanaged esRef in Beats stack monitoring

### LINKED ISSUE #5880: Remote Associations from secrets without `namespace` don't fully work as intended
alluded to here:
https://github.com/elastic/cloud-on-k8s/pull/5240#discussion_r837274347
calling association.GetUnmanagedAssociationConnectionInfoFromSecret(params.Client, esRef)
with a ref such as this:
```
    elasticsearchRefs:
    - secretName: elasticsearch-es-beat-secret
```
without `namespace: namespace_name`

fails with an error such as this:
```
[elastic-operator-0] {"log.level":"error","@timestamp":"2022-06-30T13:10:33.770Z","log.logger":"manager.eck-operator","message":"Reconciler error","service.version":"2.4.0-SNAPSHOT+d62c461b","service.type":"eck","ecs.version":"1.4.0","controller":"beat-controller","object":{"name":"heartbeat","namespace":"default"},"namespace":"default","name":"heartbeat","reconcileID":"811351ae-8472-4c58-af24-3f4b076d8d03","error":"unable to get: /elasticsearch-es-beat-secret because of unknown namespace for the cache","errorCauses":[{"error":"unable to get: /elasticsearch-es-beat-secret because of unknown namespace for the cache"}],"error.stack_trace":"sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.12.1/pkg/internal/controller/controller.go:273\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.12.1/pkg/internal/controller/controller.go:234"}
```

as this: https://github.com/elastic/cloud-on-k8s/blob/main/pkg/controller/association/secret.go#L77
returns a namespace of `""`, which means `All Namespaces`, and this use case isn't tested in the unit tests, as we don't have a caching client... I suspect we could replicate this in the tests with an additional integration test that does use the caching client..

### LINKED ISSUE #6230: Cannot configure metrics monitoring on Beats CRD with `secretName` (ECK 2.5)
## Bug Report
**What did you do?**

I set up stack monitoring as per [ECK docs](https://www.elastic.co/guide/en/cloud-on-k8s/master/k8s-stack-monitoring.html) in a central, management k8s cluster which is intended to receive log and metric data from various environmental k8s clusters (i.e. sandbox, dev, prod), and this went well. I can view monitoring data for our production cluster (ES, Kibana, and Beats) in a new monitoring cluster. In this case I am populating `elasticsearchRefs` using the names of Elastic resources since the production and monitoring clusters are deployed in the same k8s cluster.

I then wanted to configure stack monitoring in Beats deployed to our environmental k8s clusters to ship monitoring data to the monitoring ES cluster that had been deployed while following the above doc link, by following the docs to [connect to an external monitoring cluster](https://www.elastic.co/guide/en/cloud-on-k8s/2.5/k8s_connect_to_an_external_monitoring_elasticsearch_cluster.html). Since these Beats are deployed in k8s clusters external to the one our production & monitoring ES clusters are deployed to, we have to use k8s-external URLs to connect.

I added the following snippet to Beats CRDs in these environmental k8s clusters:

```
spec:
  monitoring:
    logs:
      elasticsearchRefs:
      - secretName: elastic-monitoring-settings
    metrics:
      elasticsearchRefs:
      - secretName: elastic-monitoring-settings
```

`elastic-monitoring-settings` is a secret containing keys as per the example given in the doc link to connect to an external cluster, and there is no CA cert defined (since our endpoints are using certs issued by a well-known CA).

I had expected to see Beats daemonsets being updated to deploy 2 sidecars alongside the main pod - one for Metricbeat, and the other for Filebeat - each shipping monitoring log/metric data to our monitoring cluster, just as I had observed in our production cluster when I configured stack monitoring on it using the names of Elastic resources.

First of all, there were no changes made to the Beat daemonset at all. I noted that the Beat resource was annotated with this but otherwise, nothing happens:

```
association.k8s.elastic.co/es-conf-4009089187: '{"authSecretName":"elastic-monitoring-settings","authSecretKey":"password","isServiceAccount":false,"caCertProvided":false,"caSecretName":"","url":"https://URL:443","version":"8.5.3"}'
```

While debugging I tried removing the `monitoring.metrics` object in the Beats spec, and I saw that the daemonset was updated and a Filebeat sidecar was deployed into the Beat pod. I verified that it was able to ship Beat logs to the monitoring cluster (the logs for the sidecar confirmed it was connected to our monitoring cluster).

I then tried removing `monitoring.logs` and re-adding `monitoring.metrics`, and ECK does not change the Beat daemonset, i.e. the Filebeat sidecar remains (since I had just had logs monitoring configured) and no Metricbeat sidecar is added into the daemonset.

If I then remove the `monitoring` block entirely, ECK will remove the Filebeat sidecar and the association annotation. Then, finally, if I re-add `monitoring.metrics` (and not `monitoring.logs`), then the association annotation is re-added onto the Beat resource but the daemonset is not updated and no changes are made to the running Beat pods (but a Metricbeat sidecar should be added).

It seems like `monitoring.metrics` does not work with an unmanaged secret containing connection details, although `monitoring.logs` works fine, and additionally the presence of `monitoring.metrics` with a `secretName` will prevent ECK making any other updates to the Beat.

**Environment**

* ECK version: 2.5.0+642f9ecd

* Kubernetes information: GKE v1.23.12-gke.1600

* Resource definition: https://gist.github.com/scrossan/59cc60fb3b10d530c9bb140bb42f2af7#file-beat-yml

* Logs: Logs for the cases I described above can be found in this gist: https://gist.github.com/scrossan/59cc60fb3b10d530c9bb140bb42f2af7

pebrc:
Fixes #6230 
Fixes #5880

Beats stack monitoring assumed that the associated Elasticsearch cluster is a local cluster managed by ECK. However a Beats definition as follows would break the stack monitoring feature:
```yaml
kind: Beat
metadata:
  name: filebeat
spec:
  type: filebeat
  version: 8.6.1
  elasticsearchRef:
    secretName: es-ref
  kibanaRef:
    secretName: kibana-ref
  monitoring:
    metrics:
      elasticsearchRefs:
        - secretName: monitoring-metrics-es-ref
    logs:
      elasticsearchRefs:
        - secretName: monitoring-metrics-es-ref
```

Worse still we would silently swallow the error creating the appearance of normalcy when the situation was in fact not normal. 

This PR tries to address these issues by: 
* handling the error correctly
* special casing external references to Elasticsearch correctly by 
    1. distinguishing between external references and internal references when generating Beats configuration 
    2. making a request in case of a remote unmanaged cluster to find out the cluster UUID which we seem to need for stack monitoring configuration (I have not tested what happens if we don't set it ) 

Known issues: 
* if the external Elasticsearch cluster has stack monitoring for the cluster itself turned off then the Beats monitoring data will not show up in the Kibana UI (a known issue that we already document). However monitoring data will still flow into the cluster for use in custom dashboards etc.

### DIFF
--- a/pkg/controller/association/secret.go
+++ b/pkg/controller/association/secret.go
@@ -100,6 +100,22 @@ func GetUnmanagedAssociationConnectionInfoFromSecret(c k8s.Client, o commonv1.Ob
 	return &ref, nil
 }
 
+// Version performs an HTTP GET request to the unmanaged Elastic resource at the given path and returns a string extracted
+// from the returned result using the given json path and validates it is a valid semver version.
+func (r UnmanagedAssociationConnectionInfo) Version(path string, jsonPath string) (string, error) {
+	ver, err := r.Request(path, jsonPath)
+	if err != nil {
+		return "", err
+	}
+
+	// validate the version
+	if _, err := version.Parse(ver); err != nil {
+		return "", err
+	}
+
+	return ver, nil
+}
+
 // Request performs an HTTP GET request to the unmanaged Elastic resource at the given path and returns a string extracted
 // from the returned result using the given json path.
 func (r UnmanagedAssociationConnectionInfo) Request(path string, jsonPath string) (string, error) {
@@ -149,14 +165,7 @@ func (r UnmanagedAssociationConnectionInfo) Request(path string, jsonPath string
 	if err := j.Execute(buf, obj); err != nil {
 		return "", err
 	}
-	ver := buf.String()
-
-	// validate the version
-	if _, err := version.Parse(ver); err != nil {
-		return "", err
-	}
-
-	return ver, nil
+	return buf.String(), nil
 }
 
 // filterUnmanagedElasticRef returns those associations that reference using a Kubernetes secret an Elastic resource not managed by ECK.
--- a/pkg/controller/association/controller/agent_fleetserver.go
+++ b/pkg/controller/association/controller/agent_fleetserver.go
@@ -70,7 +70,7 @@ func referencedFleetServerStatusVersion(c k8s.Client, fsRef commonv1.ObjectSelec
 		if err != nil {
 			return "", err
 		}
-		ver, err := info.Request("/api/status", "{ .version.number }")
+		ver, err := info.Version("/api/status", "{ .version.number }")
 		if err != nil {
 			// version is in the status API from version 8.0
 			if err.Error() == "version is not found" {
--- a/pkg/controller/association/controller/apm_kibana.go
+++ b/pkg/controller/association/controller/apm_kibana.go
@@ -79,7 +79,7 @@ func referencedKibanaStatusVersion(c k8s.Client, kbRef commonv1.ObjectSelector) 
 		if err != nil {
 			return "", err
 		}
-		ver, err := info.Request("/api/status", "{ .version.number }")
+		ver, err := info.Version("/api/status", "{ .version.number }")
 		if err != nil {
 			return "", err
 		}
--- a/pkg/controller/association/controller/kibana_ent.go
+++ b/pkg/controller/association/controller/kibana_ent.go
@@ -70,7 +70,7 @@ func referencedEntStatusVersion(c k8s.Client, entRef commonv1.ObjectSelector) (s
 		if err != nil {
 			return "", err
 		}
-		ver, err := info.Request("/api/ent/v1/internal/version", "{ .number }")
+		ver, err := info.Version("/api/ent/v1/internal/version", "{ .number }")
 		if err != nil {
 			return "", err
 		}
--- a/pkg/controller/association/controller/kibana_es.go
+++ b/pkg/controller/association/controller/kibana_es.go
@@ -77,7 +77,7 @@ func referencedElasticsearchStatusVersion(c k8s.Client, esRef commonv1.ObjectSel
 		if err != nil {
 			return "", err
 		}
-		ver, err := info.Request("/", "{ .version.number }")
+		ver, err := info.Version("/", "{ .version.number }")
 		if err != nil {
 			return "", err
 		}
--- a/pkg/controller/association/reconciler_test.go
+++ b/pkg/controller/association/reconciler_test.go
@@ -75,7 +75,7 @@ var (
 				if err != nil {
 					return "", err
 				}
-				// Bypass: ver, err := ref.Request("/", "{ .version.number }") and just return the version
+				// Bypass: ver, err := ref.Version("/", "{ .version.number }") and just return the version
 				return stackVersion, nil
 			}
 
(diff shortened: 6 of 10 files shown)


pkg/controller/association/secret.go
```diff
@@ -100,6 +100,22 @@ func GetUnmanagedAssociationConnectionInfoFromSecret(c k8s.Client, o commonv1.Ob
 	return &ref, nil
 }
 
+// Version performs an HTTP GET request to the unmanaged Elastic resource at the given path and returns a string extracted
+// from the returned result using the given json path and validates it is a valid semver version.
+func (r UnmanagedAssociationConnectionInfo) Version(path string, jsonPath string) (string, error) {
```
naemono, line 105: So I understand what you're doing here, but I'm a bit unsure as to why we're doing it. We are wanting all communications to unmanaged ES clusters in this code path to be explicitly version validated. We were originally doing this for all requests to unmanaged ES clusters. Was this causing an issue?

pebrc: Not sure if I understand your question here. I just renamed this function because the name was misleading. It was called `Request` originally when in fact it was doing more than just "requesting" but also parsing the result of applying the JSON path expression as a semver version. 

I have not changed anything with regards to when we to version validation on associations. All I wanted was a way to do a more generic request through this mechanism so that we can extract the UUID from the remote cluster for the Beats configuration.

naemono: Let me try to better explain. We now have both Request, and Version, of which Version is exactly the same code path as Request, and Request now just doesn't parse the version. You also updated all association controllers to use Version, which doesn't change the code path, but makes what's happening a bit clearer. But now in `associatedESUUID ` you call `uuid, err := remoteES.Request("/", "{.cluster_uuid}")`, which now does not do version validation. I'm wondering why you're not calling `.Version` here instead, as do all of the other code paths, and if that's the case, do we want to un-export `Request`?

pebrc: I am not doing a version validation because what I am interested in is not the version but the cluster UUID for the Beats config. Version validation is still/already done by the association controller. All I want is to pull a piece of data out of the unmanaged cluster so that we can configure Beats in the same way we do it for ECK managed Elasticsearch clusters.

naemono: Okay, so the validation is already being done by the beats controller, which is the piece I was missing here. Thanks again for the explanation

naemono (approved):

thbkrkr (approved):
LGTM

PR #6491, opened 2023-03-07
Fix dev license public key setup

thbkrkr:
This fixes how the dev license public key is setup to run e2e-tests.

In https://github.com/elastic/cloud-on-k8s/pull/6418, `IS_SNAPSHOT_BUILD=Yes` was changed to `BUILD_LICENSE_PUBKEY=dev` but there is one place that was not updated.

This should fix `TestElasticMapsServerCrossNSAssociation`that is failing in job `cloud-on-k8s-e2e-tests-snapshot-versions`.


barkbay (approved):
LGTM

PR #6514, opened 2023-03-13
Use vault to docker login for pushing client image

thbkrkr:
Resolves #6513.

PR #6516, opened 2023-03-13
Make sure file exists before reading it

thbkrkr:
```
14:37:07  ./hack/deployer/deployer execute --plans-file hack/deployer/config/plans.yml --config-file deployer-config.yml
...
14:37:37  Error: while logging into docker registry: open /root/.docker/config.json: no such file or directory
```

This fixes the case where the file doesn't exist 🤦.

The interesting part is that CI was not set to true due to the fact that this runs inside the CI docker container and the env variable is not propagated. I've added it.

Follow-up of #6514.


naemono (approved):
Lgtm

PR #6517, opened 2023-03-13
[ci/jks] Move setenv in snapshot-versions Jenkins job

thbkrkr:
This fixes a side effect of #6465 which causes an issue in the Jenkins job `cloud-on-k8s-e2e-tests-snapshot-versions` by ensuring the environment is set before any call to `make -C .ci ci`.

The first step `make -C .ci TARGET=validate-jenkins-pipelines ci` causes the fetch of the prod public license key.
Next `.ci/setenvconfig dev/build` sets `BUILD_LICENSE_PUBKEY=dev`. Because the `.ci/license.key` already exists, the dev license public key to test snapshots is not retrieved.

This is more of a quick fix as this job is revisited on Buildkite.

### DIFF
--- a/.ci/pipelines/e2e-tests-snapshot-versions-gke.Jenkinsfile
+++ b/.ci/pipelines/e2e-tests-snapshot-versions-gke.Jenkinsfile
@@ -33,6 +33,11 @@ pipeline {
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


naemono (approved):
Nice catch on this.

Issue #6515, opened 2023-03-13
Failed to scheduling EKS pods: running PreBind plugin "VolumeBinding": binding volumes: timed out waiting for the condition


thbkrkr:
[e2e-tests-eks-arm/632](https://devops-ci.elastic.co/job/cloud-on-k8s-e2e-tests-eks-arm/632/testReport/github/com_elastic_cloud-on-k8s_v2_test_e2e_agent/TestSystemIntegrationRecipe_All_expected_Pods_should_eventually_be_ready/)

EKS pods cannot be scheduled because they are waiting for their volumes to be bound.

In #6469, we updated EKS to install k8s `1.24` but we now also need to install the Amazon EBS CSI driver:

> If you plan to deploy workloads that use Amazon EBS volumes in a new 1.23 cluster, install the Amazon EBS CSI driver in your cluster before deploying the workloads your cluster. 
(https://docs.aws.amazon.com/eks/latest/userguide/kubernetes-versions.html#kubernetes-1.23).

```json
{
    "apiVersion": "v1",
    "kind": "Event",
    "count": 3,
    "firstTimestamp": "2023-03-13T01:25:07Z",
    "lastTimestamp": "2023-03-13T01:45:11Z", 
   "involvedObject": {
        "apiVersion": "v1",
        "kind": "Pod",
        "name": "elasticsearch-wdrv-es-default-0",
        "namespace": "e2e-sdroq-mercury",
    },
    "message": "running PreBind plugin \"VolumeBinding\": binding volumes: timed out waiting for the condition",
    "reason": "FailedScheduling",
    "type": "Warning",
    "source": {
        "component": "default-scheduler"
    }
}
```

pebrc:
Are we not using local volumes for all tests? Did the disk setup not work?

thbkrkr:
I didn't question that, I'm very unsure but it seemed to me that the majority of the tests use local volumes but that we purposely kept some tests with EBS?

According to the log, the disk setup looks fine:
```
[2023-03-13T01:03:39.937Z] Added new context arn:aws:eks:eu-west-1:444732909647:cluster/eck-jks-e2e-eks-arm-632 to /root/.kube/config
[2023-03-13T01:03:43.046Z] storageclass.storage.k8s.io/e2e-default created
[2023-03-13T01:03:43.447Z] configmap/local-provisioner-config created
[2023-03-13T01:03:43.950Z] configmap/setup-disks created
[...]

PR #6518, opened 2023-03-13
[ci/eks] Force the use of the e2e storage class when testing recipes

thbkrkr:
This forces the use of our storage class and therefore stops using EBS when testing recipes for EKS.

It's a quick fix to solve #6515. It remains to be discussed later if we keep this or if we prefer to continue doing a few tests relying on network attached storage and thus do the work to install the Amazon EBS CSI driver in EKS.

Relates to #6515.


barkbay (approved):
LGTM

PR #6527, opened 2023-03-15
[ci/bk] Support release for tags prefixed with 'v'

thbkrkr:
This changes the tag detection function to support tags prefixed with 'v'. This function is used to set the build context and so and influences the name of artifacts ([code](https://github.com/elastic/cloud-on-k8s/blob/41b7b9df5dae5bb45976d9c40ef669a737d958f7/.buildkite/scripts/build/setenv.sh#L60-L88)). We could just check that `BUILDKITE_TAG` is not empty in this function, but it seemed to me that a basic check couldn't hurt. I was a bit wrong.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -21,7 +21,9 @@ steps:
         command: "make generate check-local-changes"
 
       - label: ":go: checks"
-        commands: "make check-license-header check-predicates shellcheck reattach-pv"
+        commands:
+          - "make check-license-header check-predicates shellcheck reattach-pv"
+          - .buildkite/scripts/build/setenv_test.sh
 
   - group: tests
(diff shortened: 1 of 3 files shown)


naemono (approved):
Seems good. I know it sounds crazy, but the more I see these types of changes in these bash scripts in our CI pipeline, I wonder if moving them to go would make more sense simply from just being able to actually test them.

thbkrkr:
> Seems good. I know it sounds crazy, but the more I see these types of changes in these bash scripts in our CI pipeline, I wonder if moving them to go would make more sense simply from just being able to actually test them.

bash does not prevent writing tests. I had written one at the beginning which I had finally abandoned and which I have just revived.

bash vs go is another discussion that we can have.

PR #6528, opened 2023-03-15
[ci/bk] Dry run Helm chart release only if test

thbkrkr:
This makes `DRY_RUN=true` when doing Helm chart release only when this was triggered with the build message `buildkite test release`. In other contexts, for a merge-main build, a tag build or a build message starting with `release eck`, the Helm chart release is performed.


naemono (approved):
Makes sense

PR #6529, opened 2023-03-15
Use Docker run cache in dockerfile to speed up "go mod download" step

naemono:
This change may not benefit CI, but it certainly makes a difference locally:
https://docs.docker.com/build/cache/#use-the-dedicated-run-cache

Before:
```
#11 [builder 4/7] RUN go mod download
#11 sha256:afb98e5b1dc5bdaece9b7d6cf4d492ad40326f5fd2bff431cb752685e0e644d8
#11 DONE 37.1s
```

After:
```
#11 [builder 4/7] RUN --mount=type=cache,mode=0755,target=/go/pkg/mod go mod download
#11 sha256:2209421556ce85002e395ad4af5cc292b67bb5fef026f4e418721b1890afb92e
#11 DONE 0.3s
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,14 +10,14 @@ WORKDIR /go/src/github.com/elastic/cloud-on-k8s
 # cache deps before building and copying source so that we don't need to re-download as much
 # and so that source changes don't invalidate our downloaded layer
 COPY ["go.mod", "go.sum", "./"]
-RUN go mod download
+RUN --mount=type=cache,mode=0755,target=/go/pkg/mod go mod download
 
 # Copy the go source
 COPY pkg/    pkg/
 COPY cmd/    cmd/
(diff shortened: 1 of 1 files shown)


thbkrkr (approved):
It's a bit better that what we tried to achieve with:
```Dockerfile
COPY ["go.mod", "go.sum", "./"]
RUN go mod download
````
which only caches the layer if `go.mod` and `go.sum` have not changed.

Besides, I realize that it never worked! I think because `ARG GO_TAGS` is declared before and embeds the date, it changes for each build and prevents layer caching.

Using the dedicated RUN cache will improve the build time even when these files change. It only improves subsequent builds so it will not improve build time in CI where each image build is performed once on a fresh worker but it is nice for local builds.

LGTM

PR #6377, opened 2023-02-02
Buildkite release steps

thbkrkr:
- add step to upload manifests
- refactor how to set build and test env vars

PR #6535, opened 2023-03-16
Refactoring ubi/dockerhub images release

thbkrkr:
I missed `SNAPSHOT=true` in #6377. We can just replicate this but it isn't very explicit to understand what's going on when `SNAPSHOT=true`. Currently here is the logic to build/push the UBI image and also push the images to Docker Hub:
- when `TAG_NAME` is not empty, `.ci/setenvconfig build` set `SNAPSHOT=true`
- `make docker-multiarch-build` relies on `SNAPSHOT` to build one or all images

This PR proposes the following changes:

- Keep `SNAPSHOT=true` just for information as it is used in `GO_LDFLAGS` (`-X github.com/elastic/cloud-on-k8s/v2/pkg/about.buildSnapshot=$(SNAPSHOT)`)

- Add `PUBLISH_IMAGE_UBI` and `PUBLISH_IMAGE_UBI` to be more explicit about what we want to build and separate the two operation to be more flexible. at the moment we don't publish to Docker Hub in Buildkite.

- Rename and rearrange the build make targets a bit

- Restrict the `FIPS` images to the `linux/amd64` platform if `BUILD_PLATFORM` isn't set

We can now easily choose which variant of the image we want to build:
```js
make publish-operator-multiarch-image

PUBLISH_IMAGE_UBI=true make publish-operator-multiarch-image

PUBLISH_IMAGE_UBI=true PUBLISH_IMAGE_DOCKERHUB=true make publish-operator-multiarch-image

ENABLE_FIPS=true PUBLISH_IMAGE_UBI=true PUBLISH_IMAGE_DOCKERHUB=true BUILD_PLATFORM=linux/amd64,linux/arm64 \
make publish-operator-multiarch-image

ENABLE_FIPS=true PUBLISH_IMAGE_UBI=true PUBLISH_IMAGE_DOCKERHUB=true BUILD_PLATFORM=linux/amd64,linux/arm64 \
make publish-operator-multiarch-image
```


pebrc (approved):

thbkrkr:
Buildkite test release

thbkrkr:
buildkite test release

Issue #6537, opened 2023-03-16
dry run Helm chart release


thbkrkr:
#6528 doesn't work as expected ([cloud-on-k8s-operator/builds/2284](https://buildkite.com/elastic/cloud-on-k8s-operator/builds/2284#0186e9a0-61a5-4178-9653-792611c2b68e)).

```
Error: invalid argument "" for "-d, --dry-run" flag: strconv.ParseBool: parsing "": invalid syntax
````

When doing:
```yaml
- command: buildkite-agent pipeline upload other-pipeline.yml
  env:
    K: 42
```

Steps in `pipeline.yml`doesn't get the env `K=42`.

I confused with the step `trigger` but this requires creating a separate pipeline:

```yaml
- trigger: "other-pipeline"
  build:
    env:
      K: 42
```

PR #6538, opened 2023-03-16
Fix dry run Helm chart release

thbkrkr:
Relates to #6537.

### DIFF
--- a/.buildkite/pipeline-release.yml
+++ b/.buildkite/pipeline-release.yml
@@ -7,18 +7,29 @@ steps:
           ( build.branch == "main" && build.source != "schedule" )
           || build.tag != null
           || build.message =~ /^buildkite test .*release.*/
+        depends_on:
+          - "operator-image-build"
         commands:
           - buildkite-agent artifact download 'config/*.yaml' config/
           - .buildkite/scripts/build/setenv.sh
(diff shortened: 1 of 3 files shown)


.buildkite/pipeline-release.yml
```diff
@@ -7,18 +7,29 @@ steps:
...
+        command: buildkite-agent pipeline upload .buildkite/pipeline-release-helm.yml
+
+      # live helm charts release
+      - label: ":buildkite:"
+        if: |
+          ( build.branch == "main" && build.source != "schedule" )
+          || build.tag != null
+          || build.message =~ /^release eck/
+        command: |
+          sed "s|DRY_RUN: true|DRY_RUN: false|" .buildkite/pipeline-release-helm.yml | buildkite-agent pipeline upload
```
naemono, line 29: weird, but I can't find details in the docs making this easier, or in the forums, so 👍

naemono (approved):

.buildkite/pipeline-release.yml
thbkrkr: 'm also perlpexed on this solution but I haven't found a better one other than moving the release in another dedicated pipeline and invoking it using the step `trigger`. I'm leaving the issue open so we can discuss it.

PR #6544, opened 2023-03-17
Call _nodes/shutdown from pre-stop hook

### LINKED ISSUE #6478: Consider calling _node/shutdown from a pre-stop hook
Packing complex logic (and API calls!) into a pre-stop hook is a bit of an anti-pattern. However the ECK operator currently does not handle k8s node maintenance in a graceful way. This is because evictions due to node maintenance are not orchestrated by the operator and none of the pre-shutdown logic is executed that we run on regular scale down or ES Pod upgrades. 

This becomes a problem on clusters with a lot of data where Pods are being evicted due to node maintenance and shard recovery kicks in almost immediately. 

A possible solution would be to add a pre-stop script that queries the ES API to find out whether a node shutdown is currently in progress. If so it does nothing more. If not is issues a `_node/shutdown` request to the ES API of type `restart` (which is a guess of course because we cannot know in the pre-stop hook what kind of shutdown is happening)

Downsides of this approach are:
* exposure of additional API credentials (`cluster_admin`) in the script
* implementing a solid retry logic and dealing with unavailbility  of ES (the overall pre-stop hook timeout helps here)
* implementing a loop to wait for the shutdown `complete` condition from the ES side

pebrc:
Fixes #6478 

Adds a more complicated shell script to the pre-stop lifecycle hook. This is covers the uses case described in the referenced issue which affect mostly larger clusters with lots of data. 

This is obviously a best effort attempt at orchestrating node shutdown more gracefully. Many things can go wrong here. I anticipate that the most common reason why this logic might fail is on slow or not responding ES clusters where we might run into the overall lifecycle hook timeout. 

There is some fragile grep'ing over JSON responses here as well because we don't have `jq` or similar tools available in the Elasticsearch image.


barkbay (approved):
I have not tested all the combinations but LGTM.

Note that this PR will trigger a cluster restart, we should also update https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-upgrading-eck.html#k8s-beta-to-ga-rolling-restart accordingly.

pkg/controller/elasticsearch/nodespec/lifecycle_hook.go
```diff
@@ -36,5 +42,180 @@ set -euo pipefail
...
+
+NODE_ID=$(grep $POD_NAME $resp_body | cut -f 1 -d ' ')
+
+# check if there is an ongoing shutdown request
+request -X GET $ES_URL/_nodes/$NODE_ID/shutdown $BASIC_AUTH
+if grep -q -v '"nodes":\[\]' $resp_body; then
+	delayed_exit      
+fi
+
+echo "initiating node shutdown"
```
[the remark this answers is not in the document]
pebrc: But should we then not make sure that they are JSON formatted? Otherwise we will probably never see them?

pkg/controller/elasticsearch/nodespec/lifecycle_hook.go
```diff
@@ -36,5 +42,180 @@ set -euo pipefail
...
+echo "retrieving node ID"
+retry 10 request -X GET "$ES_URL/_cat/nodes?full_id=true&h=id,name" $BASIC_AUTH
+if [ "$?"  -ne 0 ]; then
+	error_exit $status
+fi
+
+NODE_ID=$(grep $POD_NAME $resp_body | cut -f 1 -d ' ')
+
+# check if there is an ongoing shutdown request
+request -X GET $ES_URL/_nodes/$NODE_ID/shutdown $BASIC_AUTH
```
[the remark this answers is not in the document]
pebrc: Oh good idea to run shellcheck! I will incorporate those suggestions.

pkg/controller/elasticsearch/nodespec/lifecycle_hook.go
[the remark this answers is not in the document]
barkbay: Yes, we could do something similar to what already exists in the readiness script : https://github.com/elastic/cloud-on-k8s/blob/5e77a7f4e61af4448fa05dea632510faba9d7462/pkg/controller/elasticsearch/nodespec/readiness_probe.go#L36-L40

pkg/controller/elasticsearch/nodespec/lifecycle_hook.go
```diff
@@ -36,5 +42,191 @@ set -euo pipefail
...
+  PROBE_PASSWORD=$(<{{.PreStopUserPasswordPath}})
+  BASIC_AUTH="-u {{.PreStopUserName}}:${PROBE_PASSWORD}"
+else
+  BASIC_AUTH=''
+fi
+
+ES_URL={{.ServiceURL}}
+
+if is_master; then
+  retry 10 request -X POST "$ES_URL/_cluster/voting_config_exclusions?node_names=$POD_NAME" $BASIC_AUTH
```
pebrc, line 172: Voting configuration exclusions should not be set for stateful masters that will bounce back after a restart. Node removals need a voting configuration exclusion but that will be handled by the operator. 

Setting a voting configuration exclusion on every terminating Pod is actually dangerous as it effectively shrinks the quorum.

pkg/controller/elasticsearch/nodespec/lifecycle_hook.go
```diff
@@ -36,5 +42,181 @@ set -euo pipefail
...
+  local count=0
+  until "$@"; do
+    exit=$?
+    wait=$((2 ** count))
+    count=$((count + 1))
+    if [ $global_dns_error_cnt -gt "$max_dns_errors" ]; then
+      error_exit "too many DNS errors, giving up"
+    fi
+    if [ $count -lt "$retries" ]; then
+      printf "Retry %s/%s exited %s, retrying in %s seconds...\n" "$count" "$retries" "$exit" "$wait" >&2
```
barkbay, line 99: Should we also use the `log` function here?

barkbay (approved):
LGTM, I did a bit of manual testing and it seems to work as expected.

As you mentioned in a comment the only thing I'm afraid of would be an unexpected behaviour because the Pod's metadata would not be up to date in the client cache. This adds another "best-effort" layer 😄

pkg/controller/elasticsearch/nodespec/lifecycle_hook.go
```diff
@@ -36,5 +42,181 @@ set -euo pipefail
...
+}
+
+function log() {
+   local timestamp
+   timestamp=$(date --iso-8601=seconds)
+   echo "{\"@timestamp\": \"${timestamp}\", \"message\": \"$@\", \"ecs.version\": \"1.2.0\", \"event.dataset\": \"elasticsearch.pre-stop-hook\"}" | tee /proc/1/fd/2 2> /dev/null
+}
+
+function error_exit() {
+  log "$1"
```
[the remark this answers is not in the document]
pebrc: Good point!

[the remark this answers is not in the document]
pebrc: However the $resp_body variable is just a filename not worth logging. And actually logging the contents of that file is quite tricky because it is JSON itself and I don't know an easy way to escape that if we want to embed it into another JSON document as part of `log`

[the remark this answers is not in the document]
thbkrkr: In that case maybe just do `error_exit "Failed to call node shutdown API"` and keep `log "$1"`.

pkg/controller/elasticsearch/nodespec/lifecycle_hook.go
```diff
@@ -165,7 +165,7 @@ ES_URL={{.ServiceURL}}
 
 log "retrieving node ID"
 retry 10 request -X GET "$ES_URL/_cat/nodes?full_id=true&h=id,name" $BASIC_AUTH
-if [ "$?"  -ne 0 ]; then
+if [ "$?" -ne 0 ]; then
 	error_exit "$status"
```
thbkrkr, line 169: I think `$status` needs to be declared outside the function `request` to be used globally, as here.

pebrc: You are right, I did not test that error condition.

pebrc: I found more issues while fixing this. But now we should be good.

pkg/controller/elasticsearch/nodespec/lifecycle_hook.go
```diff
@@ -36,5 +42,178 @@ set -euo pipefail
...
+  local count=0
+  until "$@"; do
+    exit=$?
+    wait=$((2 ** count))
+    count=$((count + 1))
+    if [ $global_dns_error_cnt -gt "$max_dns_errors" ]; then
+      error_exit "too many DNS errors, giving up"
+    fi
+    if [ $count -lt "$retries" ]; then
+      log "Retry $count/$retries exited $exit, retrying in $wait seconds"
```
thbkrkr, line 99:
```suggestion
      log "retry $count/$retries exited $exit, retrying in $wait seconds"
```

pkg/controller/elasticsearch/nodespec/lifecycle_hook.go
```diff
@@ -36,5 +42,178 @@ set -euo pipefail
...
+      sleep $wait
+    else
+      log "Retry $count/$retries exited $exit, no more retries left."
```
thbkrkr, line 102: Nitpick: other log messages start with a lowercase letter and have no final stop.
```suggestion
      log "retry $count/$retries exited $exit, no more retries left"
```

thbkrkr (approved):
👍

PR #6551, opened 2023-03-20
Ensure when dry-run isn't set that 'latest' tag is set on pushed image.

### LINKED ISSUE #6511: Newest certified operatorhub/Redhat releases aren't being tagged as "latest"
See https://catalog.redhat.com/software/containers/elastic/eck-operator/5fabf6d1ecb52450895164be?tag=latest&push_date=1661651299000

Newest versions of eck-operator released by the operatorhub tooling isn't tagging the release as `latest`, so it's not showing up first in the above Redhat catalog. We need to adjust our tooling to also tag as `latest` upon release.

We maybe want an option to the `container` sub-command that performs another `crane.Push`

https://github.com/elastic/cloud-on-k8s/blob/be5287bd629ebd160c714712a5c2016949552f8a/hack/operatorhub/internal/container/container.go#L240

with a `latest` tag if flag is set.

naemono:
closes #6511 

Recent releases of ECK Operator into operatorhub/Redhat ecosystem hasn't had the image tagged as `latest`. This change ensure that when `dry-run` is set to `false` the image is tagged as `latest` in `quay.io`.

Local test run below on recent `2.6.1` image successfully added the `latest` tag
```
❯ make build && ./bin/operatorhub container push --redacted-flags-- --enable-vault=false -F --dry-run=false
go build -o bin/operatorhub main.go
2023/03/20 13:53:24 Determining if image already exists within project with tag: 2.6.1
2023/03/20 13:53:25 pushing image as force was set
2023/03/20 13:53:25 pulling image (docker.elastic.co/eck/eck-operator-ubi8:2.6.1) in preparation to push (quay.io/redhat-isv-containers/5fa1f9fc4bbec60adbc8cc94:2.6.1) to quay.io registry:
2023/03/20 13:53:27 pushing image (quay.io/redhat-isv-containers/5fa1f9fc4bbec60adbc8cc94:2.6.1) to quay.io registry:
2023/03/20 13:53:29 ✓
2023/03/20 13:53:29 tagging (quay.io/redhat-isv-containers/5fa1f9fc4bbec60adbc8cc94:2.6.1) as 'latest' in quay.io registry:
2023/03/20 13:53:30 ✓
```

image

### DIFF
--- a/hack/operatorhub/internal/container/container.go
+++ b/hack/operatorhub/internal/container/container.go
@@ -220,7 +220,7 @@ func getFirstUndeletedImage(images []Image) *Image {
 // container image locally, and push it to Quay.io registry using the
 // provided credentials.
 func pushImageToRegistry(c PushConfig) error {
-	if c.DryRun {
+	if c.DryRun && !c.Force {
 		log.Printf("not pushing image as dry-run is set.")
 		return nil
 	}
@@ -236,7 +236,8 @@ func pushImageToRegistry(c PushConfig) error {
 		log.Println("ⅹ")
 		return fmt.Errorf("while pulling (%s): %w", imageToPull, err)
 	}
-	log.Printf("pushing image (%s) to quay.io registry: ", formattedEckOperatorRedhatReference)
+	log.Println("✓")
+	log.Printf("pushing image (%s) to quay.io registry\n", formattedEckOperatorRedhatReference)
 	err = crane.Push(
 		image,
 		formattedEckOperatorRedhatReference,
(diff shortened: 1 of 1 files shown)


hack/operatorhub/internal/container/container.go
```diff
@@ -251,7 +252,19 @@ func pushImageToRegistry(c PushConfig) error {
 		log.Println("ⅹ")
 		return fmt.Errorf("while pushing (%s): %w", formattedEckOperatorRedhatReference, err)
 	}
-	log.Println("✓")
+	// Since we only push when dry-run isn't set, go ahead and
```
thbkrkr, line 255:
```suggestion
	log.Println("✓")
	// Since we only push when dry-run isn't set, go ahead and
```

hack/operatorhub/internal/container/container.go
```diff
@@ -220,7 +220,7 @@ func getFirstUndeletedImage(images []Image) *Image {
 // container image locally, and push it to Quay.io registry using the
 // provided credentials.
 func pushImageToRegistry(c PushConfig) error {
-	if c.DryRun {
+	if c.DryRun && !c.Force {
```
thbkrkr, line 223: Why did you change this? Isn't it strange to do `--dry-run --force`?

naemono: Because if the same image already exists, you can't force-push the image, or add latest tag, but that doesn't exactly make sense with dry-run, agreed. I've removed this. If dry-run is set it should never do anything.
```
	log.Printf("Determining if image already exists within project with tag: %s", c.Tag)
	exists, err := ImageExistsInProject(c.HTTPClient, c.RedhatCatalogAPIKey, c.ProjectID, c.Tag)
	if err != nil {
		log.Println("ⅹ")
		return fmt.Errorf("failed to determine if image exists: %w", err)
	}
	if exists && !c.Force {
		log.Println("✓")
		log.Println("not continuing as image was already found within redhat project")
		return nil
	}
	if c.Force {
		log.Println("pushing image as force was set")
	}

	if err = pushImageToRegistry(c); err != nil {
		log.Println("x")
		return fmt.Errorf("failed to push image: %w", err)
	}
````

thbkrkr (approved):
LGTM

PR #6556, opened 2023-03-21
Fix publish dockerhub

thbkrkr:
This fixes several bugs in the `publish-dockerhub.sh` script.

- The credentials stored in the `docker-registry-elastic` secret cannot push to the `eck-dev` namespace of the Elastic Docker registry, only to `eck-ci`. That is the reason of the error `server message: insufficient_scope: authorization failed`.
- The test to check if docker buildx should be installed is inverted.
- The docker buildx version `v0.8.2` does not support multiple repositories, we need to install at least `v0.9.0` to have it.
- The URL to install docker buildx is wrong, it is the one for arm instead of amd.
- The docker login command doesn't have the name of the registry, required in dry run mode to be authenticated to the Elastic Docker registry.
- The secret to use in dry-run mode to get the docker registry creds is inverted with the one to use in live mode.

### DIFF
--- a/hack/publish-dockerhub.sh
+++ b/hack/publish-dockerhub.sh
@@ -25,11 +25,11 @@ vault_login() {
 
 registry_login() {
     if [[ "${DRY_RUN:-}" == "false" ]]; then
+        username=$(vault read -field=username secret/release/docker-hub-eck)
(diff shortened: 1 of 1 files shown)


naemono (approved):
Looks good.

PR #6563, opened 2023-03-22
Fleet server QuickStart: use emptyDir for the agent-data volumes

daverick:
The default value is to use a host path Volume, that can be restricted by a podSecurityPolicy.
When there is such a podSecurityPolicy, the pod is not created and the cause of the error is lost in the event of the ECK created ReplicaSet. But in the `fleet-server-quickstart` events you can see a reconciliation error: `Reconciliation error: failed to request https://kibana-quickstart-kb-http.default.svc:5601/api/fleet/setup, status is 401` which is not the real problem.

This proposed change would avoid this situation by following the [workaround that we have when host path is not possible](https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-elastic-agent-fleet-known-limitations.html#k8s_storing_local_state_in_host_path_volume_2). This is not optimal for production but this is a QuickStart  ;-)

### DIFF
--- a/docs/orchestrating-elastic-stack-applications/agent-fleet.asciidoc
+++ b/docs/orchestrating-elastic-stack-applications/agent-fleet.asciidoc
@@ -44,6 +44,9 @@ spec:
         automountServiceAccountToken: true
         securityContext:
           runAsUser: 0
+        volumes:
+        - name: agent-data
+          emptyDir: {} # may not be suited for production, the default value is to use host path volume
 ---
 apiVersion: agent.k8s.elastic.co/v1alpha1
 kind: Agent
@@ -64,6 +67,9 @@ spec:
         automountServiceAccountToken: true
         securityContext:
           runAsUser: 0
+        volumes:
+        - name: agent-data
+          emptyDir: {} # may not be suited for production, the default value is to use host path volume
 ---
 apiVersion: kibana.k8s.elastic.co/v1
 kind: Kibana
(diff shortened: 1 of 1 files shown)


docs/orchestrating-elastic-stack-applications/agent-fleet.asciidoc
```diff
@@ -64,6 +67,9 @@ spec:
         automountServiceAccountToken: true
         securityContext:
           runAsUser: 0
+        volumes:
+        - name: agent-data
+          emptyDir: {} # may not be suited for production, the default value is to use host path volume
```
pebrc, line 72: I would only do this for Fleet server. Which afaik does not actually need to persist state. While regular Agents running integrations do need this.

barkbay: > I would only do this for Fleet server.

IIUC last commit removed the `emptyDir` on the Fleet server.

Also:
* The use of `emptyDir` is documented in the `Storing local state in host path volume` section, at the end of this document. Should we update it to mention security restrictions, and the fact that it is not required for Fleet Server (which is a TIL for me btw)?
* Should we just not try to use an `hostpath` if Fleet Server is enabled? (to be done in a separate issue/pr ofc)

pebrc: I have not been able to get an answer from the Fleet team on this. However given that we are running Fleet server as a deployment and the chance is high that the on eviction or deletion of the Pod we end up on a different k8s node, I think we are good to merge this as is.

pebrc (approved):

PR #6575, opened 2023-03-23
Controller tests

robbavey:
This commit adds extra unit tests for the Logstash Controller, including tests for config, controller, pods and services


pkg/controller/logstash/pod.go
```diff
@@ -127,3 +129,8 @@ func readinessProbe(useTLS bool) corev1.Probe {
 		},
 	}
 }
+
+// GetLogstashContainer returns the Logstash container from the given podSpec.
+func GetLogstashContainer(podSpec corev1.PodSpec) *corev1.Container {
+	return pod.ContainerByName(podSpec, "logstash")
```
kaisecheng, line 135: there is a constant for the container name https://github.com/elastic/cloud-on-k8s/blob/feature/logstash/pkg/apis/logstash/v1alpha1/logstash_types.go#L17

pkg/controller/logstash/pod_test.go
```diff
@@ -0,0 +1,205 @@
...
+					},
+				},
+			}},
+			assertions: func(pod corev1.PodTemplateSpec) {
+				assert.Len(t, pod.Spec.Volumes, 2)
+				assert.Len(t, GetLogstashContainer(pod.Spec).VolumeMounts, 2)
+			},
+		},
+	}
+	for _, tt := range tests {
```
kaisecheng, line 193: Can we add test coverage for custom ports, annotation and readinessProbe?

kaisecheng (commented):
Many thanks for adding the tests. Mostly looks good to me. I left three comments for you to consider.

pkg/controller/logstash/pod_test.go
robbavey: Done (I think...)

pkg/controller/logstash/logstash_controller_test.go
```diff
@@ -0,0 +1,224 @@
...
+}
+
+func TestReconcileLogstash_Reconcile(t *testing.T) {
+	defaultLabels := (&logstashv1alpha1.Logstash{ObjectMeta: metav1.ObjectMeta{Name: "testLogstash"}}).GetIdentityLabels()
+	tests := []struct {
+		name     string
+		objs     []runtime.Object
+		request  reconcile.Request
+		want     reconcile.Result
+		expected logstashv1alpha1.Logstash
```
barkbay, line 45: Would it make sense to also check that all the other objects have been reconciled? I think only validating that they were created would be fine as their "contents" are verified in other tests.

For example:

```golang
type expectedObject struct {
	t    client.Object
	name types.NamespacedName
}

type expectedObjects []expectedObject

func (e expectedObjects) assertExist(t *testing.T, k8s client.Client) {
	t.Helper()
	for _, o := range e {
		obj := o.t.DeepCopyObject().(client.Object)
		assert.NoError(t, k8s.Get(context.Background(), o.name, obj), "Expected object not found: %s", o.name)
	}
}
```

And then in a test:

```golang
		{
			name: "Logstash with ready StatefulSet and Pod updates status properly",
			[...]
			expectedObjects: []expectedObject{
				{
					t:    &corev1.Service{},
					name: types.NamespacedName{Namespace: "test", Name: "testLogstash-ls-api"},
				},
				{
					t:    &corev1.Secret{},
					name: types.NamespacedName{Namespace: "test", Name: "testLogstash-ls-config"},
				},
			},
			wantErr: false,
		},
	for _, tt := range tests {
[...]
			tt.expectedObjects.assertExist(t, r.Client)
[...]
}
```

pkg/controller/logstash/services_test.go
```diff
@@ -0,0 +1,250 @@
...
+			require.Equal(t, len(tc.wantSvc), len(haveSvc))
+
+			for i := range tc.wantSvc {
+				comparison.AssertEqual(t, &tc.wantSvc[i], &haveSvc[i])
+			}
+		})
+	}
+}
+
+func defaultServiceFromOwner(owner *logstashv1alpha1.Logstash) *corev1.Service {
```
barkbay, line 219: nit: why do we need a function to create the default service while it should be created automatically?

test/e2e/test/logstash/http_client.go
```diff
@@ -31,9 +32,16 @@ func NewLogstashClient(logstash v1alpha1.Logstash, k *test.K8sClient) (*http.Cli
 }
 
 func DoRequest(client *http.Client, logstash v1alpha1.Logstash, method, path string) ([]byte, error) {
-	scheme := "http"
+	var scheme = "http"
+	var port = network.HTTPPort
+	for _, service := range logstash.Spec.Services {
+		if service.Name == "api" {
```
barkbay, line 38: nit: maybe `"api"` would deserve to be a constant that could be reused here, in `pkg/controller/logstash/service.go` and `pkg/controller/logstash/pod.go`?

pkg/controller/logstash/pod.go
```diff
@@ -107,23 +109,32 @@ func getDefaultContainerPorts() []corev1.ContainerPort {
...
+				Port:   intstr.FromInt(port),
 				Path:   "/",
 				Scheme: scheme,
 			},
 		},
 	}
+	return probe
+}
+
+// GetLogstashContainer returns the Logstash container from the given podSpec.
```
barkbay, line 137: nit: this seems to be only used in unit tests, let's move it to `pkg/controller/logstash/pod_test.go`?

barkbay (approved):
lgtm!

pkg/controller/logstash/logstash_controller_test.go
robbavey: Great idea! Done

pkg/controller/logstash/services_test.go
robbavey: Good point - I've removed it

robbavey:
@barkbay Ready for another round

pkg/controller/logstash/pod.go
```diff
@@ -117,23 +118,27 @@ func getDefaultContainerPorts() []corev1.ContainerPort {
...
-func readinessProbe(useTLS bool) corev1.Probe {
-	scheme := corev1.URISchemeHTTP
-	if useTLS {
-		scheme = corev1.URISchemeHTTPS
+func readinessProbe(logstash logstashv1alpha1.Logstash) corev1.Probe {
+	var scheme = corev1.URISchemeHTTP
+	var port = network.HTTPPort
+	for _, service := range logstash.Spec.Services {
+		if service.Name == LogstashAPIServiceName {
+			port = int(service.Service.Spec.Ports[0].Port)
```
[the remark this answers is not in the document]
robbavey: Good catch! Fixed

robbavey:
@barkbay Nice find! I've pushed a fix and added a test to verify

barkbay (approved):
LGTM!

Issue #6583, opened 2023-03-24
e2e-tests failures


thbkrkr:
The first run of e2e tests since #6525 is merged failed for:

#### kind

```sh
Exited with status -1 (agent lost)
```

- mem/cpu resources issue

- I didn't report that for kind, we need at least a `n1-standard-16` gcp agent.

#### eks-arm

```sh
make run-deployer
/buildkite/builds/bk-agent-elastic-k8s-167966427351480
0400/elastic/cloud-on-k8s-operator-nightly/.buildkite/scripts/test/../../../.env: line 6: kb: command not found
```

- caused when `set-deployer-config.sh` sources `.env` that contains an env value with spaces

To reproduce:
```
echo "E2E_TAGS=es kb apm ent beat agent" >> .env
.buildkite/scripts/test/set-deployer-config.sh
```

#### eks

```
step.go:44:
Error Trace:    /go/src/github.com/elastic/cloud-on-k8s/test/e2e/test/utils.go:88
Error:          Received unexpected error:
    pod es-apm-sample-bcrc-es-default-0 is not Ready.
    Status:{
        "phase": "Pending",
        "conditions": [
            {
                "type": "PodScheduled",
                "status": "False",
                "lastProbeTime": null,
                "lastTransitionTime": "2023-03-24T14:04:15Z",
                "reason": "SchedulerError",
                "message": "running PreBind plugin \"VolumeBinding\": binding volumes: timed out waiting for the condition"
            }
        ],
        "qosClass": "Burstable"
    }
Test:           TestSamples/apm-apm_es_kibana/All_expected_Pods_should_eventually_be_ready
```

Looks like #6518 doesn't work.

Because `TestSamples` doesn't use the same builder than the one used for recipes changed in #6518.

#### stack-8.7.0-SNAPSHOT-dev

License issue on `TestElasticMapsServerCrossNSAssociation`.

```
FAIL: TestElasticMapsServerCrossNSAssociation/ElasticMapsServer_deployment_should_be_created (1800.00s)
```
```
{
    "reason": "ReconciliationError",
    "message": "Elastic Maps Server is an enterprise feature. Enterprise features are disabled",
    "kind": "ElasticMapsServer",
    "name": "test-cross-ns-ems-es-kt8z",
    "namespace": "e2e-rodno-venus"
}
```

[...]

PR #6586, opened 2023-03-24
Set kind gcp agents machine type to n1-standard-16

thbkrkr:
Relates to #6583.


pebrc (approved):
LGTM I wonder if we should make it more generic though? Like an option to override machine type as one of the `fixed` variables or so.

thbkrkr:
> LGTM I wonder if we should make it more generic though? Like an option to override machine type as one of the fixed variables or so.

If we have more like this, I think it will deserve it. It seems a bit overkill for the moment.

PR #6587, opened 2023-03-24
[ci] Stop using .env file in set-deployer-script.sh script

thbkrkr:
Quick fix the issue when an env var has spaces by stopping to use `.env` file in `set-deployer-script.sh` script.
Also switch to a comma separated list for the Go build tags as spaces separated list is deprecated.

Limitation: E2E_TAGS with multiple tags does not work with the pipeline-gen used with flags because the comma is already used to separate the k=v tuples in --mixed and --fixed.

Relates to #6596.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -469,18 +469,21 @@ E2E_REGISTRY_NAMESPACE     ?= eck-dev
 E2E_IMG_TAG                ?= $(IMG_VERSION)
 E2E_IMG                    ?= $(REGISTRY)/$(E2E_REGISTRY_NAMESPACE)/eck-e2e-tests:$(E2E_IMG_TAG)
 E2E_STACK_VERSION          ?= 8.6.1
-export TESTS_MATCH         ?= "^Test" #testing # can be overriden to eg. TESTS_MATCH=TestMutationMoreNodes to match a single test
+# regexp to filter tests to run
+export TESTS_MATCH         ?= "^Test"
 export E2E_JSON            ?= false
 TEST_TIMEOUT               ?= 30m
 E2E_SKIP_CLEANUP           ?= false
 E2E_DEPLOY_CHAOS_JOB       ?= false
-E2E_TAGS                   ?= e2e  # go build constraints potentially restricting the tests to run
-E2E_TEST_ENV_TAGS          ?= ""   # tags conveying information about the test environment to the test runner
+# go build constraints potentially restricting the tests to run
+E2E_TAGS                   ?= e2e
+# tags conveying information about the test environment to the test runner
+E2E_TEST_ENV_TAGS          ?= ""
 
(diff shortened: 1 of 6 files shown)


thbkrkr:
buildkite test this -f p=kind,t=TestSmoke,E2E_TAGS="es kb"

thbkrkr:
buildkite test this -f p=kind,t=TestSmoke,E2E_TAGS="es kb"

thbkrkr:
buildkite test this -f p=kind,t=TestSmoke,E2E_TAGS="es kb"

thbkrkr:
buildkite test this -f p=kind,t=TestSmoke,E2E_TAGS="es kb"

thbkrkr:
buildkite test this -f p=kind,t=TestSmoke,E2E_TAGS="es kb"

thbkrkr:
buildkite test this -f p=kind,t=TestSmoke,E2E_TAGS="es kb"

thbkrkr:
buildkite test this -f p=kind,t=TestSmoke,E2E_TAGS="es kb"

naemono:
@thbkrkr would moving to [comma separated](https://github.com/golang/go/commit/80e7832733fd245181af3394077f2df21303a4aa) make this any easier to handle? Are there historical reasons for us using space separated?

pebrc (commented):
LGTM but I am not sure I understand why we no longer need the `.env` file.

Makefile
```diff
@@ -469,18 +469,21 @@ E2E_REGISTRY_NAMESPACE     ?= eck-dev
...
+# regexp to filter tests to run
+export TESTS_MATCH         ?= "^Test"
 export E2E_JSON            ?= false
 TEST_TIMEOUT               ?= 30m
 E2E_SKIP_CLEANUP           ?= false
 E2E_DEPLOY_CHAOS_JOB       ?= false
-E2E_TAGS                   ?= e2e  # go build constraints potentially restricting the tests to run
-E2E_TEST_ENV_TAGS          ?= ""   # tags conveying information about the test environment to the test runner
+# go build constraints potentially restricting the tests to run
+E2E_TAGS                   ?= e2e
```
thbkrkr, line 479: I moved the comment otherwise all the spaces before the `#` are used for the variable value!

pebrc (approved):
LGTM I wonder if we should refactor `pipeline-gen` and maybe use a different separator? I know you had reservations about using something like `;` because of its special meaning in shells. But given that that can be escaped or single quoted it might be a good alternative to lift the restriction we are adding in this PR in a (non-urgent) follow-up.

naemono (approved):

thbkrkr:
To use `;` to separate args in pipeline-gen indeed seems a good alternative with this new constraint in mind. I added a note on this to follow up https://github.com/elastic/cloud-on-k8s/issues/6596#issuecomment-1485804120.

PR #6588, opened 2023-03-24
[ci/eks] Force the use of the e2e storage class when testing samples

thbkrkr:
Similar to #6518.
Relates to #6583, #6515.


thbkrkr:
buildkite test this -f p=eks,t=TestSamples

pebrc (approved):

Issue #6571, opened 2023-03-23
Test 1.25 and 1.26 on kind


pebrc:
Kind v0.17.0 support both 1.25 and 1.26 with the following node images:

kindest/node:v1.25.3@sha256:f52781bc0d7a19fb6c405c2af83abfeb311f130707a0e219175677e366cc45d1
kindest/node:v1.26.0@sha256:691e24bd2417609db7e589e1a479b902d2e209892a10ce375fab60a8407c7352

Source: https://github.com/kubernetes-sigs/kind/releases

PR #6593, opened 2023-03-27
Update k8s supported versions to 1.22-1.26

thbkrkr:
Update k8s supported versions to `1.22-1.26`.

Relates to #6571.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ Current features:
 
 Supported versions:
 
-*  Kubernetes 1.21-1.25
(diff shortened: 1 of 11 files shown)


thbkrkr:
> Should we be more liberal in the Helm chart and allow installation even on older not supported versions. Or do we have a strong reason to lock it down like this.

There is no strong reason to do that. We can keep the `minKubeVersion` of the Helm charts to the minimal k8s version which is tested daily with the e2e tests suite.

pebrc (approved):

PR #6595, opened 2023-03-27
Remove migrated Jenkins jobs

thbkrkr:
Remove last migrated Jenkins jobs, only keep the release job.


naemono:
@thbkrkr These make targets are still mentioned in the `.ci/README.md`, so we probably want to either remove or edit that file as well.

pebrc (approved):
LGTM, good point about the README though (can come in a follow up if you prefer)

PR #6597, opened 2023-03-27
Update OpenShift supported versions to 4.8-4.12

thbkrkr:
Update OpenShift supported versions to `4.8-4.12`.

Relates to #6592.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ Current features:
 Supported versions:
 
 *  Kubernetes 1.22-1.26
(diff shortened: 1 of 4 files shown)


pebrc:
Maybe I missed it, but do we want to run a full set of e2e tests before we officially declare support. I think https://github.com/elastic/cloud-on-k8s/pull/6592 only did a smoke test?

pebrc (approved):

PR #6604, opened 2023-03-28
Use Buildkite agent image for ci container on gcp agent

thbkrkr:
This avoids having to maintain two almost similar Dockerfiles in two different places.


naemono (approved):
Double checked the agent Dockerfile against this one, and I don't see any missing items, or reason not to proceed. I really like not having to duplicate across these 2x Dockerfiles. 👍

PR #6605, opened 2023-03-28
Add a new role for APM Server 8.7.0+

axw:
Add a new role for APM Server 8.7.0+, adding read privileges for the system indices `.apm-agent-configuration` and `.apm-source-map`.

### DIFF
--- a/pkg/controller/association/controller/apm_es.go
+++ b/pkg/controller/association/controller/apm_es.go
@@ -100,6 +100,14 @@ func getAPMElasticsearchRoles(associated commonv1.Associated) (string, error) {
 		return "", err
 	}
 
+	// 8.7.x and above
+	if v.GTE(version.MinFor(8, 7, 0)) {
+		return strings.Join([]string{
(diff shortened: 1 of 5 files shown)


pebrc:
@thbkrkr this probably still needs to go into 2.7!
@axw will APM server 8.7 work without this i.e. is this a blocker for ECK + APM server 8.7?

axw:
> will APM server 8.7 work without this i.e. is this a blocker for ECK + APM server 8.7?

It's not a blocker. There are two features that this affects. Those features previously required a `kibanaRef`, and they will continue to work with a `kibanaRef`. The reason behind this change is essentially to allow the features to work without a connection to Kibana.

pebrc:
> It's not a blocker. 

That's good to hear. @thbkrkr I think we should still aim to include this given that we have a bit more time.

thbkrkr:
> @thbkrkr I think we should still aim to include this given that we have a bit more time.

Yes, I think so too. I merge and take care of the backport.

Issue #6600, opened 2023-03-27
Beats: Using `secureSettings` with `hostPath` fails without `runAsUser: 0`


naemono:
related #6156 
and potentially related #6599 (If we want to try and automatically handle this scenario)

When using `secureSettings` with `hostPath` fails on `elastic-internal-init-keystore` `initContainer` without setting `runAsUser: 0`.

logs
```
❯ kc logs -n default eck-beats-beat-filebeat-xgkfx elastic-internal-init-keystore
Initializing keystore.
+ echo 'Initializing keystore.'
+ filebeat keystore create --force
error initializing beat: failed to create Beat meta file: open /usr/share/filebeat/data/meta.json.new: permission denied
```

Manifest
```
apiVersion: beat.k8s.elastic.co/v1beta1
kind: Beat
metadata:
  name: eck-beats
  namespace: default
spec:
  config:
    filebeat.inputs:
    - paths:
      - /var/log/containers/*.log
      type: container
    output.elasticsearch.hosts:
    - https://my-elasticsearch.gcp.elastic-cloud.com
    output.elasticsearch.username: elastic
    processors:
    - add_host_metadata: {}
    - add_cloud_metadata: {}
  daemonSet:
    podTemplate:
      spec:
        automountServiceAccountToken: true
        containers:
        - name: filebeat
          securityContext:
            runAsUser: 0
          volumeMounts:
          - mountPath: /var/log/containers
            name: varlogcontainers
          - mountPath: /var/log/pods
            name: varlogpods
          - mountPath: /var/lib/docker/containers
            name: varlibdockercontainers
        dnsPolicy: ClusterFirstWithHostNet
        hostNetwork: true
        terminationGracePeriodSeconds: 30
        volumes:
        - hostPath:
            path: /var/log/containers
          name: varlogcontainers
        - hostPath:
            path: /var/log/pods
          name: varlogpods
        - hostPath:
            path: /var/lib/docker/containers
          name: varlibdockercontainers
  secureSettings:
  - secretName: elasticsearch-credentials
  type: filebeat
  version: 8.6.1
```

ebuildy:
this is a serious security issue, most of good security practices prevent to run container as root.

A quick (and dirty) fix is to use an init container to chown:

```
initContainers:
- command:
  - sh
  - -c
  - chown -R 1000:1000 /usr/share/metricbeat/data
  image: docker.elastic.co/beats/metricbeat:{{ $values.version }}
[...]

PR #6599, opened 2023-03-27
Automatically adjust Elastic Agent hostPath permissions

naemono:
closes #6239 

## Background

Currently it is required to have the following set when running Elastic Agent with a `hostPath`:

```
    podTemplate:
      spec:
        containers:
          - name: agent
            securityContext:
              runAsUser: 0
```

The only way to avoid this is configuring an `emptyDir` volume instead of `hostPath`.

## What this proposes

Detailed further in #6239 we want to automatically add an `initContainer` that maintains the Agent permissions to avoid requiring the Agent to run perpetually as root.

If the following all are true, an `initContainer` is automatically added to Elastic Agent that maintains permissions
1. Agent volume is not set to `emptyDir`.
2. Agent version is above 7.15.
3. Agent spec is not configured to run as root.

## Testing

- [x] Local testing `make e2e-local` of all Agent e2e tests using `8.7.0-SNAPSHOT` and enabling all disabled Agent tests because of #6331 
- [ ] Full remote e2e test run using 7.x
- [ ] Full remote e2e test run using 8.6
- [ ] Adding of additional Unit tests

Issue #6280, opened 2023-01-05
Beats CRD does not support running as filebeat user


brsolomon-deloitte:
It appears (undocumented) that the `Beats` CRD / eck-beats does not support running as the `filebeat` user (UID 1000) from the default docker.elastic.co/beats/filebeat image, and must be run as UID 0 (bad security practice).

Config:

```
version: 8.5.3
spec:
  type: filebeat
  deployment:
    replicas: 1
    strategy:
      type: Recreate
    podTemplate:
      spec:
        containers:
          - name: filebeat
            securityContext:
              runAsUser: 1000
```

Error:

```
{"log.level":"info","@timestamp":"2023-01-05T13:59:39.964Z","log.origin":{"file.name":"instance/beat.go","file.line":708},"message":"Home path: [/usr/share/filebeat] Config path: [/usr/share/filebeat] Data path: [/usr/share/filebeat/data] Logs path: [/usr/share/filebeat/logs]","service.name":"filebeat","ecs.version":"1.6.0"}
{"log.level":"error","@timestamp":"2023-01-05T13:59:39.964Z","log.origin":{"file.name":"instance/beat.go","file.line":1057},"message":"Exiting: failed to create Beat meta file: open /usr/share/filebeat/data/meta.json.new: permission denied","service.name":"filebeat","ecs.version":"1.6.0"}
Exiting: failed to create Beat meta file: open /usr/share/filebeat/data/meta.json.new: permission denied
```

brsolomon-deloitte:
Within the container, the filebeat user is `uid=1000(filebeat) gid=1000(filebeat) groups=1000(filebeat),0(root)` and

```
$ ls -ld /usr/share/filebeat/data
drwxrwxr-x 2 root root 6 Dec  4 05:01 /usr/share/filebeat/data
```

This means the filebeat user _should_ have write access there since they are part of the 0 group and the group octal has a `w` bit set.

pknetbcn:
Same here. 

Is there any solution?

ebuildy:
As you noticed, the ECK operator create a hostPath volume for Beat pod.

The hostPath will be mounted as root (k8s / docker side), hence your user cant read it.

A workaround is to use an initContainer, run as root, to chown the volume before running the beat container:

In pod template, add the following initContainer:

```
command:
- sh
- -c
- chown -R 1000:1000 /usr/share/beat/data
image: docker.elastic.co/beats/filebeat:{{ .version }}
[...]

PR #6612, opened 2023-03-28
Discuss/Draft: Automatically adjust Elastic Beat hostPath permissions

naemono:
relates #6599, #6600, #6280

I would like to handle Beat `hostpath` permissions similar to Agent being done in #6599. Unfortunately Beats seems to be a bit more complicated and strict when it comes to permissions.

## Issues
1. Beat handles files in subdirectories in it's `data` directory, not just the root so recursive operations are needed when adjusting permissions.
2. Beat seems to try and `chmod` files that it doesn't like the ownership/permissions of (see details in `hostPathVolumeInitContainerCommand` func)
3. Beat is very strict about it's permissions on the keystore, and requires it be owned by the UID of Beat, and only be readable/writable by that user.

What this means is we need to know the UID that beat is running as, which isn't especially difficult in a standard Kubernetes environment, but when Openshift is involved, this is not predictable.

Does anyone see a path forward for handling `hostPaths` automatically for Beat in Openshift outside of detecting it, and doing nothing (as we do now). The current solution in this PR works (in limited testing) outside of Openshift.

### DIFF
--- a/pkg/controller/beat/common/pod.go
+++ b/pkg/controller/beat/common/pod.go
@@ -188,6 +188,8 @@ func buildPodTemplate(
 		sideCars = append(sideCars, sideCar.Container)
 	}
 
+	initContainers = append(initContainers, maybeBeatInitContainerForHostpathVolume(params.Beat)...)
+
 	labels := maps.Merge(params.Beat.GetIdentityLabels(), map[string]string{
 		VersionLabelName: spec.Version})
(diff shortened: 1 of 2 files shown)


naemono:
Closing for the same reasons noted https://github.com/elastic/cloud-on-k8s/pull/6599#issuecomment-1513366543

PR #6630, opened 2023-03-30
Fix indexing issues with Helm charts release tool

naemono:
## Core issue this is resolving
When releasing Helm Charts during a standard release, we release both the eck-operator, and the eck-resources helm charts at the same time, using the same tool in CI. Since both of these operations writes an `index.yaml` file to a Google Storage bucket, they can conflict with each other. GS has [request preconditions](https://cloud.google.com/storage/docs/request-preconditions#json-api) to help with this scenario.

## What's changing?
1. When we go to write a new Helm Index `index.yaml` file, since we've previously read the index file, we know it's `generation` id which is an `int64` number handled/maintained by Google Storage. We instruct the Storage Client to ensure that the underlying `index.yaml` file has a specified `generation` with a `precondition`, and if the underlying file isn't that `generation`, then fail the write operation. On `http.StatusPreconditionFailed` we delay for a random amount of time, and retry the full Helm release operation.
2. New cli flag `keep-temp-dir` to keep the temporary directories around for debugging purposes.
3. When packaging the Helm Chart, we write the destination directly to the temp directory, instead of performing a copy operation *to* the temp directory. 
4. Fixing some linter issues (using `os` vs `ioutil`; switch on error type vs `errors.As`)
5. Adding some unit tests surrounding the behavior of reading/writing google storage files and preconditions.


hack/helm/release/internal/helm/google_storage_test.go
```diff
@@ -0,0 +1,313 @@
...
+		}))
+	if err != nil {
+		t.Fatalf("getting new storage client: %s", err)
+		return nil, nil
+	}
+	return client, ts
+}
+
+func testServer() *httptest.Server {
+	return httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
naemono, line 151: The below set of if statements aren't very readable. Any suggestions on cleaning this up would be appreciated.

naemono:
buildkite test this

naemono:
buildkite test this

naemono:
buildkite test this

thbkrkr:
This doesn't solve the problem we are facing where the root cause is that we run `helm repo update` which overwrites our recent change on the index. It makes us retrieve a version of the index cached by the CDN which does not necessarily contain the recent update that we just added a few seconds ago.

From my understanding, using request preconditions when updating the index in the GCS bucket is good when you have two processes that may work concurrently on the index. In our case, we release the helm charts sequentially, so I don't think we are affecting by this.

As I suggested earlier, I think we should keep the tool as simple as possible, since we have full control over all the charts and their sub charts dependencies that we are releasing. I sketched it here (https://github.com/thbkrkr/cloud-on-k8s/commit/092ee18f90487084295da9cd03e45ca3a94cb5b1) removing everything I could and just kept a simple flow `readCharts; uploadCharts; uploadIndex`, also without separating charts with deps vs without deps for which I don't understand the need.

naemono:
@thbkrkr Just getting back to this. These releases are not sequential, as they just rely on the build, and the dev operation to complete and in the last release they overlapped by 2 seconds:
```
(operator prod helm chart)
Started Running Job
(+243ms)
Thu 30th Mar at 3:41:23.124 AM

Finished Running Job
(+14s)
Thu 30th Mar at 3:41:38.081 AM
Command Exit Status	0

(eck-resources prod helm charts)
Started Running Job
(+243ms)
Thu 30th Mar at 3:41:36.649 AM

 Finished Running Job
(+1m)
Thu 30th Mar at 3:42:52.148 AM
```

```
      - label: "operator prod helm chart"
        if: |
          build.tag =~ /^v[0-9]+\.[0-9]+\.[0-9]+\$/
          || build.message == "release eck-operator helm charts"
          || build.message == "release all helm charts"
        depends_on:
          - "build-helm-releaser-tool"
          - "eck-operator-dev-helm"
        commands:
          - buildkite-agent artifact download "bin/releaser" /usr/local/
          - chmod u+x /usr/local/bin/releaser
          - releaser --env=prod --charts-dir=deploy/eck-operator --dry-run=\$DRY_RUN

      - label: "eck-resources prod helm charts"
        if: |
          build.tag =~ /^v[0-9]+\.[0-9]+\.[0-9]+\$/
          || build.message == "release eck-resources helm charts"
          || build.message == "release all helm charts"
        depends_on:
          - "build-helm-releaser-tool"
          - "eck-resources-dev-helm"
        commands:
          - buildkite-agent artifact download "bin/releaser" /usr/local/
          - chmod u+x /usr/local/bin/releaser
          - releaser --env=prod --excludes=eck-operator --dry-run=\$DRY_RUN
```

I'm going to look further into the changes your branch introduces and make sure we are in agreement to the actual root cause of these failures and get this resolved. 👍

naemono:
closing this in favor of #6734

PR #6638, opened 2023-04-03
Update upgrade-test-harness to ECK 2.7.0 / 2.8.0-SNAPSHOT

thbkrkr:
* Prune EOL versions `1.7.0/7.14.x`, `1.7.1/7.15.x`
* Test only latest ECK `1.9.1/7.17.x`, remove `1.9.0/7.17.x`
* Update patch Stack versions
* Add ECK `2.7.0/8.7.0`
* Update upcoming to `2.8.0-SNAPSHOT/8.8.0-SNAPSHOT`


hack/upgrade-test-harness/testdata/v180/stack.yaml
```diff
@@ -4,7 +4,7 @@ kind: Elasticsearch
 metadata:
   name: es
 spec:
-  version: 7.15.0
+  version: 7.16.3
```
naemono, line 7: This was confusing, but now I see that these were just wrong... 👍

naemono (approved):
looks good.

PR #6640, opened 2023-04-03
Fix detecting multiarch image already pushed

### LINKED ISSUE #6624: eks-arm is failing: arm e2e tests image is not built in nightly pipeline
`eks-arm` is failing in the cloud-on-k8s-operator-nightly pipeline.

I haven't fully checked yet but I think the problem is that the arm64 image hasn't been built.

How is it possible?

The main cloud-on-k8s-operator pipeline builds only the amd64 image:
```sh
docker buildx build \
  --progress=plain \
  --file test/e2e/Dockerfile \
  --platform linux/amd64 \
  --push \
  -t docker.elastic.co/eck-ci/eck-e2e-tests:2.8.0-SNAPSHOT-348d5a6d .
```

When the cloud-on-k8s-operator-nightly pipeline runs, we build the image with `--platform linux/arm64,linux/amd64` and the exact same image name for the e2e tests image, but we skip the build:

```
OK: image docker.elastic.co/eck-ci/eck-e2e-tests:2.8.0-SNAPSHOT-348d5a6d already published
```

I added the "is already pushed" test recently for the e2e tests image. It only existed for the operator image before Buildkite.
https://github.com/elastic/cloud-on-k8s/blob/61b5cad8c98c9cc4c1ce22e52f3c0a75d7e642ec/Makefile#L497-L500

Also before #6535, the test (for the operator image) checked if there was an arm64 image pushed. 

`docker buildx imagetools inspect $(OPERATOR_IMAGE) | grep -q 'linux/arm64' 2>&1 >/dev/null`

But this doesn't work well when we are not building `linux/arm64`, for a PR for example. That's why I removed the `| grep -q 'linux/arm64'`.

Quickfix: use a different name for the e2e image in the nightly pipeline than in the main pipeline.

Better fix (?): Have a better test of whether image(s) have already been pushed depending on the platform(s) we want to build.

thbkrkr:
* Reuse images built on merge-main in nightly-main and also build the two platforms on merge-main: there is no point in rebuilding the images in the nightly build when they were built during the merge.

* Fix detection of multiarch image publishing for correctness: by adding a script that checks if an image has been published for given platforms.

Resolves #6624.

### DIFF
--- a/.buildkite/scripts/build/is_published.sh
+++ b/.buildkite/scripts/build/is_published.sh
@@ -25,7 +25,7 @@ main() {
   local platform=$2
 
   set +e
-  inspect_out="$(docker buildx imagetools inspect $image)"
+  inspect_out=$(docker buildx imagetools inspect "$image")
   set -e
 
   case "$platform" in
--- a/.buildkite/scripts/build/is_published.sh
+++ b/.buildkite/scripts/build/is_published.sh
@@ -1,0 +1,47 @@
+#!/usr/bin/env bash
+
+# Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+# or more contributor license agreements. Licensed under the Elastic License 2.0;
+# you may not use this file except in compliance with the Elastic License 2.0.
+
+# Script to check if a multi arch image has already been published.
+#
+# Usage: is_published.sh IMAGE PLATFORM
+
+set -eu
+
+is_amd64_build() {
+  grep -c "^Name:\s*$image$" <<< "$inspect_out" >/dev/null
+}
+
+is_amd64_arm64_build() {
+  grep -c "^Name:\s*$image$"        <<< "$inspect_out" >/dev/null && \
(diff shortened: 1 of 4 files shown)


.buildkite/scripts/build/is_published.sh
```diff
@@ -0,0 +1,47 @@
...
+set -eu
+
+is_amd64_build() {
+  grep -c "^Name:\s*$image$" <<< "$inspect_out" >/dev/null
+}
+
+is_amd64_arm64_build() {
+  grep -c "^Name:\s*$image$"        <<< "$inspect_out" >/dev/null && \
+  grep -c "Platform:\s*linux/amd64" <<< "$inspect_out" >/dev/null && \
+  grep -c "Platform:\s*linux/amd64" <<< "$inspect_out" >/dev/null
```
naemono, line 20:
```suggestion
  grep -c "Platform:\s*linux/arm64" <<< "$inspect_out" >/dev/null
```

naemono (changes requested):
I think arm64 is missing in the grep.

naemono (approved):
Looks good now

PR #6644, opened 2023-04-03
Better support dev operator build to test snapshot stack versions

### LINKED ISSUE #6567: Support operator build with dev license pubkey in PR builds and explicit image suffix
Follow-up of https://github.com/elastic/cloud-on-k8s/pull/6525#discussion_r1143851378.

There are two things to improve here:
1. better explicit why we use the value of `BUILD_LICENSE_PUBKEY` to suffix the operator image in the pipeline generator
2. support operator build with dev license pubkey in PR builds

For 1., we could add a more explicit `OPERATOR_IMAGE_SUFFIX` which will be a repeat of `BUILD_LICENSE_PUBKEY` during build and will be used for testing instead of `BUILD_LICENSE_PUBKEY` as suggested in https://github.com/elastic/cloud-on-k8s/pull/6525#discussion_r1145082455?

For 2., following the current logic, the problem is that when we build images, we don't look if there will be e2e-tests that test stack snapshot and need the -dev operator image.

For the context, in Jenkins, we had a separate pipeline only to build and test stack snapshot. Now, in Buildkite, we unified all tests in a single pipeline. It looks like this:
```
pipeline-build
  if nightly-build
    additional build-operator-image with BUILD_LICENSE_PUBKEY
      -> use BUILD_LICENSE_PUBKEY to get a special pubkey embedded in the operator (cmd/license-initializer)
      -> push image suffixed by -$BUILD_LICENSE_PUBKEY

pipeline-e2e
  pipeline-gen
    if BUILD_LICENSE_PUBKEY
      -> use image suffixed by -$BUILD_LICENSE_PUBKEY
```

thbkrkr:
- Explicit OPERATOR_VERSION_SUFFIX var for the build pipeline
- Automatically add special vars to test snapshot stack versions in pipeline-gen

Resolves #6567.

### DIFF
--- a/.buildkite/e2e/nightly-main-matrix.yaml
+++ b/.buildkite/e2e/nightly-main-matrix.yaml
@@ -15,8 +15,6 @@
     - E2E_STACK_VERSION: "8.3.3"
     - E2E_STACK_VERSION: "8.4.3"
     - E2E_STACK_VERSION: "8.7.0-SNAPSHOT"
-      BUILD_LICENSE_PUBKEY: dev
-      TEST_LICENSE_PKEY_PATH: "in-memory"
 
 - label: kind
   fixed:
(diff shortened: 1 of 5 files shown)


thbkrkr:
buildkite test this -f p=kind,s=8.8.0-SNAPSHOT,t=TestEnterpriseTrial

PR #6658, opened 2023-04-05
Improve operatorhub release

thbkrkr:
This fixes a few issues encountered when using the tool to release ECK `2.7.0`.

To run manually the tool, I added a few make targets and updated the preflight script.

Most important fixes:
- Do not fail fork if repo already up-to-date bea966c
- Fix path of additional file for 'git add' 3e99612
- Fix replacement without effect b0a796d

Important changes:
* Set default dir to ./ b659060
* Checks crds/operator resources have been extracted from manifests 551ef58
* Update pipeline definitions b920893

Other changes are less important.

Test:
- https://buildkite.com/elastic/cloud-on-k8s-operator-redhat-release/builds/18


.buildkite/pipeline-release-redhat.yml
```diff
@@ -1,50 +1,61 @@
...
+  - label: ":redhat: generate and create-pr"
+    key: "redhat-release"
+    depends_on:
+      - "build-operatorhub-tool"
+      - "redhat-publish"
+    commands:
+      - buildkite-agent artifact download "bin/operatorhub" /usr/local/ && chmod u+x /usr/local/bin/operatorhub
+      - |
+        cd hack/operatorhub
+        #operatorhub generate-manifests --yaml-manifest=../../config/crds.yaml --yaml-manifest=../../config/operator.yaml
```
naemono, line 58: I suspect you intended to remove this.

.buildkite/pipeline-release.yml
```diff
@@ -35,4 +35,7 @@ steps:
         if: |
           build.message =~ /^buildkite test .*operatorhub.*/
           || build.message =~ /^release eck .*operatorhub.*/
-        command: buildkite-agent pipeline upload .buildkite/pipeline-release-redhat.yml
+        trigger: cloud-on-k8s-operator-redhat-release
```
naemono, line 38: Ok, it took me a minute here to understand that this was triggering another pipeline. Gotcha. Hadn't dealt much with the `trigger` command.

hack/operatorhub/Makefile
```diff
@@ -2,5 +2,27 @@
...
 # you may not use this file except in compliance with the Elastic License 2.0.
 
+DRY_RUN ?= true
+TAG     ?= $$(cat ../../VERSION)
+
 build:
 	go build -o bin/operatorhub main.go
+
+container-push:
+	bin/operatorhub container push --dry-run=$(DRY_RUN)
```
naemono, line 12:
```suggestion
	bin/operatorhub container push --dry-run="$(DRY_RUN)"
```

hack/operatorhub/Makefile
```diff
@@ -2,5 +2,27 @@
...
+	../../.buildkite/scripts/release/redhat-preflight.sh $(TAG)
+
+container-publish:
+	bin/operatorhub container publish --dry-run="$(DRY_RUN)"
+
+generate-manifests:
+	bin/operatorhub generate-manifests --yaml-manifest=../../config/crds.yaml --yaml-manifest=../../config/operator.yaml
+
+bundle-generate:
+	bin/operatorhub bundle generate --dir="$$(pwd)"
```
naemono, line 24: Isn't the `$$` only needed in Buildkite?

```
❯ dir="$$(pwd)"; echo $dir
70487(pwd)

❯ dir="$(pwd)"; echo $dir
/Users/mmontgomery
```

hack/operatorhub/Makefile
```diff
@@ -2,5 +2,27 @@
...
+
+create-pr:
+	@ bin/operatorhub bundle create-pr --enable-vault=false --dir="$$(pwd)" --dry-run="$(DRY_RUN)" -D=false \
+		-f ""$(GITHUB_FULLNAME)"" -e "$(GITHUB_EMAIL)" -u "$(GITHUB_USERNAME)" -g "$(GITHUB_TOKEN)"
```
naemono, line 28: Is the `""` around `GITHUB_FULLNAME` intentional?

hack/operatorhub/config.yaml
```diff
@@ -1,4 +1,4 @@
-newVersion: 2.8.0-SNAPSHOT
+newVersion: 2.7.0
```
naemono, line 1: Was this intentional? Seems outside of the scope of this PR.

hack/operatorhub/Makefile
thbkrkr: Here we don't want the variable `$pwd` but run the shell command `$(pwd)` and `$` is a special character in make that needs to be escaped. I can use the other syntax `$(shell pwd)` which is less confusing.

hack/operatorhub/Makefile
thbkrkr: Nop.

hack/operatorhub/config.yaml
thbkrkr: It was just for testing, reverting.

.buildkite/pipeline-release-redhat.yml
thbkrkr: Yes, it seems to me more correct to download the released manifests rather than to regenerate them but this has the downside that we can't run the tool (even with `--dry-run=true`) on a version for which we have not yet published manifests.

naemono (approved):

PR #6662, opened 2023-04-05
Logstash add ElasticsearchRefs

kaisecheng:
This PR allows configuration of `elasticsearchRefs` in CRD to associate Elasticsearch clusters to Logstash. The connectivity info/ credentials are passed via environment variable to Logstash container for referencing in logstash-output-elasticsearch, logstash-input-elasticsearch and logstash-filter-elasticsearch.

Sample config
```yaml
kind: Logstash
metadata:
  name: logstash-sample
spec:
  elasticsearchRefs:
    - name: production
    - name: demo
      namespace: demo
    
```
Environment variables available in Logstash container
```bash
DEFAULT_PRODUCTION_ES_HOSTS
DEFAULT_PRODUCTION_ES_USERNAME
DEFAULT_PRODUCTION_ES_PASSWORD
DEFAULT_PRODUCTION_ES_SSL_CERTIFICATE_AUTHORITY # path of ca.crt of production cluster
DEMO_DEMO_ES_HOSTS
DEMO_DEMO_ES_USERNAME
DEMO_DEMO_ES_PASSWORD
DEMO_DEMO_ES_SSL_CERTIFICATE_AUTHORITY # path of ca.crt of demo cluster
```
Logstash pipelines.yml
```yaml
- pipeline.id: main
  config.string: |
    output { 
     elasticsearch{
       hosts => "${DEFAULT_PRODUCTION_ES_HOSTS}"
       user => "${DEFAULT_PRODUCTION_ES_USERNAME}"
       password => "${DEFAULT_PRODUCTION_ES_PASSWORD}" 
       cacert => "${DEFAULT_PRODUCTION_ES_SSL_CERTIFICATE_AUTHORITY}"
      } 
    }
```


pkg/apis/logstash/v1alpha1/logstash_types.go
```diff
@@ -31,6 +35,10 @@ type LogstashSpec struct {
 	// +kubebuilder:validation:Optional
 	Image string `json:"image,omitempty"`
 
+	// ElasticsearchRef is a reference to an Elasticsearch cluster running in the same Kubernetes cluster.
+	// +kubebuilder:validation:Optional
+	ElasticsearchRefs []commonv1.ObjectSelector `json:"elasticsearchRefs,omitempty"`
```
robbavey, line 40: Thinking forward a little to when we will have named clusters in Logstash, we talked about introducing the notion of a `clusterName` into the ElasticsearchRef definition along the lines of

```
elasticsearchRefs:
  - clusterName: k8sData
    name: local
    namespace: production               
```

with the `clusterName` being used as the id in the `cluster.yaml` equivalent that we envisage for Logstash.
If we think about a `cluster.yaml` looking something like:

```
elasticsearch_clusters:
  k8sData:
    host: "https://metaldata-es-http:9200"
    username: “logstash_system
    password: REDACTED
    ca_file: /mnt/elastic-internal/es-certs/ca.crt
```

I think we could do this by creating the `LogstashSpec` to look like:

```
type LogstashSpec struct {
  :
  // Named list of Elasticsearch Clusters
  ElasticsearchRefs[] ElasticsearchCluster `json:"elasticsearchRefs,omitempty"` 
  :
}

// Definition of type to define an Elasticsearch Cluster to be referred to in 
// Logstash pipeline
type ElasticsearchCluster struct {
  commonv1.ObjectSelector `json:",omitempty,inline"`
  ClusterName              string `json:"clusterName,omitempty"`
}
```

We could potentially create the env variables with the clusterName? (And if we did would we need the namespace here?)

There is some prior art with the elastic agent using an [outputName](https://github.com/elastic/cloud-on-k8s/blob/main/config/recipes/elastic-agent/multi-output.yaml#L8) to distinguish between multiple `elasticsearchRefs`.

Alternatively, we could stick with what we have here, and write `cluster.yaml` as:

```
elasticsearch_clusters:
  local_production:
    host: "https://metaldata-es-http:9200"
    username: “logstash_system
    password: REDACTED
    ca_file: /mnt/elastic-internal/es-certs/ca.crt
```

using `$namespace_$name` as the id.

WDYT?

robbavey (commented):
This looks *fantastic*. I do have a couple of broad questions before I dive in fully

pkg/apis/logstash/v1alpha1/logstash_types.go
kaisecheng: These are awesome suggestions. +1 to follow the prior art of agent to unify the experience. Also, +1 to bring cluster name in technical preview. 

I think we can use `outputName` instead of `clusterName` to further share the experience. For the env variable, let's replace `$namespace_$name` with `outputName`. The name needs to be unique in `elasticsearchRefs`. I will add validation.

robbavey: My thought on using `clusterName` over `outputName` is that we also use want to use `elasticsearchRef` in `input` and `filter` plugins, not just for outputs

kaisecheng: @robbavey Revisiting the usage of `clusterName`. I have doubts about its value. One drawback is the stack monitoring should understand `clusterName` and send metrics to the corresponding cluster, which does not feasible now because the operator only considers namespace + name in the association. 

From the design doc, things we want to achieve (now and future `cluster.yml`) with clusterName seem can be replaced by namespace + name.
```
    elasticsearch {
      cluster_id => namespace_name
    }
```
```
elasticsearch_clusters:
  namespace_name:
    host: "https://metaldata-es-http:9200"
    username: “logstash_system
    password: REDACTED
    ca_file: /mnt/elastic-internal/es-certs/ca.crt
  namespace_name2:
    cloud_id: XXXXXXX
     api_key: yyyyyy
```
`clusterName` works like a nickname that is handy when user wants to give a second name.
I wanna keep the current env variable naming convention, namespace + name, at least in the technical preview.

robbavey: Does stack monitoring need `clusterName`? Or can it just use the same settings that it has today? It is metricbeat sending the data, so I don't think it would need any changes? 

But, it does seem a little confusing having two different ways to specify `ElasticsearchRef` depending on whether it is a plugin usage or a stack monitoring usage (plus we'll need to consider central management soon)

I'm ok with keeping it as `namespace_name` for the technical preview. My main concern was that it was a little unwieldy, especially taking into account that users need to remember to change all their `-` to `_` when reading ENV variable, but if it turns out that it is a significant usability concern we can amend later.

kaisecheng: stack monitoring doesn't need `clusterName` to work, and it is perfectly fine to work the same way as of today with namespace + name. But when users give a nickname to a cluster, it is not nice that stack monitoring doesn't understand/ support nickname.

robbavey: ok - Let's keep it with `namespace_name` for technical preview, and change if it looks problematic

kaisecheng:
@barkbay I think this is ready for review

There is a [discussion](https://github.com/elastic/cloud-on-k8s/pull/6662#discussion_r1163350603) about bringing `clusterName` to `elasticsearchRefs`. The decision is not to add `clusterName` in the technical preview.

pkg/controller/elasticsearch/user/roles.go
```diff
@@ -173,6 +175,22 @@ var (
...
+				"read_ilm",
+				"manage_logstash_pipelines",
+				"manage_index_templates",
+				"cluster:admin/ingest/pipeline/get",
+			},
+			Indices: []esclient.IndexRole{
+				{
+					Names:      []string{"logstash-*", "ecs-logstash-*", "logs-*", "metrics-*", "synthetics-*", "traces-*"},
+					Privileges: []string{"manage", "read", "create_doc", "view_index_metadata", "create_index"},
+				},
```
robbavey, line 191: I think we need more permissions here, in terms of both index names, and privileges. 
I tried a few different simple scenarios, and there were a few issues - 
Removing `data_streams => false` will write to `logstash`, and `ecs_compatibility => disabled` will write to `ecs-logstash` by default, so we will need to add those to the list of `Names`
Additionally, I think we need to add additional privileges  - we get a 
```
[indices:data/write/index:op_type/index] is unauthorized for user [default-logstash-sample-default-elasticsearch-logstash-user] with effective roles [eck_logstash_user_role] on indices [ecs-logstash], this action is granted by the index privileges [create,index,write,all]"
``` 
error when attempting to write to this index. I suspect we will need the `write` privilege here - the elasticsearch output is able to do `create`, `index`, `update` and `delete` [actions](https://www.elastic.co/guide/en/logstash/current/plugins-outputs-elasticsearch.html#plugins-outputs-elasticsearch-action)
If we are going to limit the indices here, we will need to add instructions on how users can get around this

robbavey (changes requested):

pkg/controller/elasticsearch/user/roles.go
kaisecheng: Thanks for double check it. I have updated the setting. 
`ecs-logstash` and `logstash` are required by `ilm_rollover_alias`
The logstash [doc](https://www.elastic.co/guide/en/logstash/master/ls-security.html) regarding secure connection needs update

robbavey: Thank you! This now works as expected.

pkg/apis/logstash/v1alpha1/logstash_types.go
robbavey: What do you think about removing `DEFAULT_` if we are using the default namespace? And only including namespace if it is specified?

pkg/controller/logstash/env_test.go
```diff
@@ -0,0 +1,106 @@
...
+		},
+		{
+			name:   "es ref without tls",
+			params: params,
+			setAssocConfs: func(assocs []commonv1.Association) {
+				assocs[0].SetAssociationConf(&commonv1.AssociationConf{
+					AuthSecretName: "logstash-sample-default-elasticsearch-sample-logstash-user",
+					AuthSecretKey:  "default-logstash-sample-default-elasticsearch-sample-logstash-user",
+					CACertProvided: false,
+					CASecretName:   "logstash-sample-logstash-es-default-elasticsearch-sample-ca",
```
robbavey, line 85: Would we expect this to be created if we are not using tls?

robbavey (commented):
This looks good to me from a logstash perspective.

My only concern is around the experience when referring to an index that is not one of the "default" ones, and how we choose to document this.

barkbay:
> @barkbay I think this is ready for review

Sure, I'll do a first review today.

pkg/apis/logstash/v1alpha1/logstash_types.go
kaisecheng: k8s best practice is to not create any resources to the `default` namespace. I would expect users always create their namespaces, and keep `DEFAULT_` for consistency in case it is used.  An alternative could be creating two sets of env variables for the default namespace, one with `DEFAULT_` and one without a namespace.

Not sure if we will use `clusterName` or `namespace_name` in `cluster.yml` in the future. Having two sets of names may be confusing. I tend to believe users have knowledge of the name of the default namespace.

pkg/controller/elasticsearch/user/roles.go
```diff
@@ -150,6 +152,22 @@ var (
 				},
 			},
 		},
+		LogstashUserRole: esclient.Role{
```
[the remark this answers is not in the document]
barkbay: > The indices chosen here will restrict logstash to only being able to create the "standard" indices/templates, and not custom ones created by users.

That's a good point. New roles can be added using [the `roles` section](https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-users-and-roles.html#k8s_creating_custom_roles), and even if it is not ideal this can also be used to override predefined roles.

barkbay (commented):
I left a few comments, I still have to do some tests.

pkg/controller/logstash/env.go
```diff
@@ -0,0 +1,60 @@
...
+			envs = append(envs, createEnvVar(esRefName+"_ES_SSL_CERTIFICATE_AUTHORITY", caPath))
+		}
+	}
+
+	return envs, nil
+}
+
+func getEsRefNamespacedName(assoc commonv1.Association) string {
+	ref := assoc.AssociationRef()
+	return fmt.Sprintf("%s_%s", ref.Namespace, ref.Name)
```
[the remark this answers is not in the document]
kaisecheng: Great that you raise the issue. We can address the external ES connection in a follow-up. I am thinking of using `ref.NameOrSecretName()` for env. In the future, the operator will need to write managed and unmanaged cluster info to a `cluster.yml`. Logstash will need to map the `cluster.yml` to plugin config.

In the technical preview, we aim for managed clusters.

pkg/controller/logstash/env_test.go
kaisecheng: removed

pkg/controller/logstash/env.go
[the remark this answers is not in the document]
kaisecheng: add an issue for followup https://github.com/elastic/ingest-dev/issues/1764

kaisecheng:
@barkbay I have created a follow-up for connecting external / non ECK managed cluster and applied all suggestions. This is ready for the 2nd look.

pkg/apis/logstash/v1alpha1/logstash_types.go
barkbay: I was trying to understand why the proposal of using the `clusterName` has been ruled out. The main reason is that concatenating the namespace with the name seems a bit brittle, in the sense that it does not really prevent collisions (yeah I know, it's a bit "far-fetched" 😄 ), and also because I think this is the kind of situation where, as a user, I would favor an explicit naming over an implicit one. It allows for example to change the target cluster without updating the configuration.

> One drawback is the stack monitoring should understand clusterName and send metrics to the corresponding cluster, which does not feasible now because the operator only considers namespace + name in the association.

I think it makes sense to treat stack monitoring as a special case because the concept of `clusterName` has no meaning in that context?

barkbay: I just realized that this would also allow users to install the sample manifest into a new namespace without needing to update the configuration:

```
k apply -f config/samples/logstash/logstash_es.yaml -n foo
```

```
[2023-04-20T13:58:05,513][ERROR][logstash.agent           ] An exception happened when converging configuration {:exception=>RuntimeError, :message=>"Could not fetch the configuration, message: Cannot evaluate `${DEFAULT_ELASTICSEARCH_SAMPLE_ES_HOSTS}`. Replacement variable `DEFAULT_ELASTICSEARCH_SAMPLE_ES_HOSTS` is not defined in a Logstash secret store or as an Environment entry and there is no default value given."}
```

kaisecheng: My main concern about not using `clusterName` is as a user, I am happy to give an explicit name, but it is very confusing that I cannot use the name in stack monitoring. As a user, I could not understand why stack monitoring is a special case.

robbavey: I tend to agree with @barkbay that the lack of `clusterName` in `elasticsearchRef` in stack monitoring is probably ok - `clusterName` is added as an aid for the user to refer to an elasticsearch cluster pointed to by `elasticsearchRef` outside of the kubernetes CRD in pipeline configurations, without needing to deal with kubernetes semantics, such as namespaces. The ability to define/move namespaces outside of the CRD is an an interesting twist on this.

pebrc: Could this be an optional feature? If users define it it will be used to name the environment variables? Also could it then just be named `alias`? 

```yaml
elasticsearchRefs:
  - alias: o11y-cluster
    name: monitor
    namespace: production  
```
 I think I also agree that monitoring is different enough to allow a special case here. Because users have to understand the relationship between the `elasticsearchRef` and what they define in configuration there is a motivation for such a construct. Why would one want to use an alias for stack monitoring?

pkg/controller/elasticsearch/user/roles.go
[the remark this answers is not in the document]
pebrc: I feel like the uncommented/explained use of `superuser` roles by Agent and Enterprise Search is actually somewhat of a bug. Even if there is a motivation for it in order to create and ingest data into custom indices, it feels like something that should be called out in a comment/doc at the very least and probably a custom role appropriate privileges to create arbitrary indices would have still been better than `superuser`.

pkg/apis/logstash/v1alpha1/logstash_types.go
kaisecheng: In Logstash context, `clusterName` is a mapping of name to endpoint (host, username, password, cacert). The endpoint can be an input (source) or an output (destination). It will be referenced in pipeline.

```
input { 
  kafka { } 
}
output {
  elasticsearch { cluster_name => "production" }
  elasticsearch { cluster_name => "uat" }
}
```

In Logstash CRD, we allow users to define the name mapping
```
  elasticsearchRefs:
    - clusterName: production
      name: es1
    - clusterName: uat
       name:es2
    - clusterName: monitor
       name:es3
```

However, Stack Monitoring does not understand the name `production`, `uat` and `monitor`. If Stack Monitoring wants to insert metrics to `monitor`, it needs the following

```
  monitoring:
    metrics:
      elasticsearchRefs:
        - name: es3
```
As a user, I am confused why I can't use `monitor` that I declared earlier. The name works in pipeline but not in Stack Monitoring.

IMO, ultimately we will need `clusterName` to achieve `cluster.yml` which contains the mapping of alias to endpoint. In technical preview, it seems not ready. I hope the example can explain the confusion.

robbavey: I guess my confusion here, is what would be your expectation of using `clusterName` in the `monitoring` section? 

Are you expecting to be able to use `clusterName` instead of `name` + `namespace`? Or as well as? Is your expectation that you could use 
```
 monitoring:
    metrics:
      elasticsearchRefs:
        - clusterName: monitor
```

or that you would expect 
```
 monitoring:
    metrics:
      elasticsearchRefs:
         - clusterName: monitor
            name:es3
```
or that `clusterName` is arbitrary without the specific `cluster` or `clusterName` setting in `cluster.yml` when that comes to be?

I agree that it feels a little half baked without the `cluster.yml`, and we only refer to `clusterName` in the env variable, but I do feel that it will have utility when we start allowing named clusters, and maybe introducing the notion earlier is of some value.

Thoughts @flexitrev, @jsvd, @roaksoax?

kaisecheng: my expectation is to be able to use `clusterName` to output metrics to `monitor`
```
 monitoring:
    metrics:
      elasticsearchRefs:
        - clusterName: monitor
```
and won't stop user to use namespace + name to output
```
  monitoring:
    metrics:
      elasticsearchRefs:
        - name: es3
```

The declaration in `elasticsearchRefs` will all convert to `cluster.yml`. Only the `clusterName` in `cluster.yml` can be referenced in pipeline.

```
  elasticsearchRefs:
    - clusterName: production
      name: es1
    - clusterName: uat
       name:es2
    - clusterName: monitor
       name:es3
```
cluster.yml
```
- production:
    host:...
    username:...
    password:...
    cacert:...
- uat:
    host:...
    username:...
    password:...
    cacert:...
- monitor:
    host:...
    username:...
    password:...
    cacert:...
```

barkbay (commented):
FYI users can provide their own Elasticsearch certificates, signed by a custom or "well-known" CA, with a [Subject Alternative|Common]Name that does not include `<es>-es-http.<ns>.svc`.

I did a quick test using the [`CertManager`](https://cert-manager.io/):
1. I generated a custom CA, and used it to sign a certificate, omitting the `.svc` cluster name (as it would be the case for a "public" certificate signed by Let's Encrypt for example).
2. I referenced that certificate using our [documented ](https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-tls-certificates.html)`http` field:

```yaml
kind: Elasticsearch
metadata:
  name: elasticsearch-sample
spec:
  version: 8.7.0
  http:
    tls:
      certificate:
        secretName: es-crt-secret
```

As a result:

1. Logstash was no longer able to connect (note: it was an update, Elasticsearch certificates were updated on an already running Elasticsearch instance):

```
[2023-04-21T07:34:40,561][WARN ][logstash.outputs.elasticsearch][main] Attempted to resurrect connection to dead ES instance, but got an error {:url=>"https://foo-logstash-sample-foo-elasticsearch-sample-logstash-user:xxxxxx@elasticsearch-sample-es-http.foo.svc:9200/", :exception=>LogStash::Outputs::ElasticSearch::HttpClient::Pool::HostUnreachableError, :message=>"Elasticsearch Unreachable: [https://elasticsearch-sample-es-http.foo.svc:9200/][Manticore::UnknownException] Certificate for <elasticsearch-sample-es-http.foo.svc> doesn't match any of the subject alternative names: [example.com, foo.example.com]"}
```

2. Pod was still considered as `Available`:

```
k get logstashes.logstash.k8s.elastic.co,pod
NAME                                               AVAILABLE   EXPECTED   AGE   VERSION
logstash.logstash.k8s.elastic.co/logstash-sample   1           1          17h   8.7.0

NAME                                    READY   STATUS    RESTARTS   AGE
[...]
pod/logstash-sample-ls-0                1/1     Running   0          6m18s
```

 The fact that the `Pod` is marked as `Ready` is perhaps confusing from the user's point of view. But this point can be addressed later when working on a "health" API.

Regarding the validation issue, I believe this is one of the reason `ssl.verificationMode` is set to `certificate` for Kibana for example. I think there are 2 workarounds:
* Let the user define the name in the certificate in `hosts => [ "https://name-in-cert" ]`
* Change the way the certificate is validated (idk how it can be done in Logstash).

I just wanted to make you aware of that situation. FWIW we are not consistent re. how this situation is handled: https://github.com/elastic/cloud-on-k8s/issues/4812

config/samples/logstash/logstash_es.yaml
```diff
@@ -0,0 +1,35 @@
...
+  version: 8.7.0
+  elasticsearchRefs:
+    - name: elasticsearch-sample
+  pipelines:
+    - pipeline.id: main
+      config.string: |
+        input { exec { command => 'uptime' interval => 10 } } 
+        output { 
+          elasticsearch {
+            hosts => [ "${DEFAULT_ELASTICSEARCH_SAMPLE_ES_HOSTS}" ]
```
barkbay, line 28: nit: let's add a comment to explain that this example is expected to be deployed in the `default` namespace. The env. variables must be updated if it is not the case.

pkg/controller/logstash/volume.go
```diff
@@ -0,0 +1,75 @@
...
+// you may not use this file except in compliance with the Elastic License 2.0.
+
+package logstash
+
+import (
+	"fmt"
+	"path"
+
+	commonv1 "github.com/elastic/cloud-on-k8s/v2/pkg/apis/common/v1"
+
```
barkbay, line 12: nit: remove empty line

test/e2e/logstash/pipeline_test.go
```diff
@@ -7,13 +7,15 @@
 package logstash
 
 import (
-	corev1 "k8s.io/api/core/v1"
 	"testing"
 
+	corev1 "k8s.io/api/core/v1"
+
```
barkbay, line 13: nit: remove new line

barkbay (commented):
LGTM from an "ECK pov"!
I'll defer to @robbavey or @pebrc for the final ✅

pebrc (commented):
LGTM from the ECK perspective. I leave the decision on the `clusterName` feature to you, it feels like it can also be added later if necessary with the small downside that both ways have to be supported then.

pkg/apis/logstash/v1alpha1/validations.go
```diff
@@ -86,3 +90,19 @@ func checkSinglePipelineSource(a *Logstash) field.ErrorList {
...
+
+func checkESRefsNamed(l *Logstash) field.ErrorList {
+	var notNamed []string
+	for _, o := range l.Spec.ElasticsearchRefs {
+		if o.ClusterName == "" {
+			notNamed = append(notNamed, o.NamespacedName().String())
+		}
+	}
+	if len(notNamed) > 0 {
+		msg := fmt.Sprintf("when declaring multiple refs all have to be named, missing clusterName on %v", notNamed)
```
robbavey, line 102: I think the error message needs to state that `clusterName` is mandatory for any number of elasticsearchRefs, not just multiple

```suggestion
		msg := fmt.Sprintf("clusterName is a mandatory field - missing on %v", notNamed)
```

pkg/apis/logstash/v1alpha1/validations_test.go
```diff
@@ -229,3 +229,176 @@ func Test_checkSupportedVersion(t *testing.T) {
...
+	}{
+		{
+			name: "no ref: OK",
+			args: args{
+				b: &Logstash{},
+			},
+			wantErr: false,
+		},
+		{
+			name: "one unnamed ref: NOK",
```
robbavey, line 343:
```suggestion
			name: "one ref, missing clusterName: NOK",
```

pkg/apis/logstash/v1alpha1/validations_test.go
```diff
@@ -229,3 +229,176 @@ func Test_checkSupportedVersion(t *testing.T) {
...
+								ObjectSelector: commonv1.ObjectSelector{Name: "bla", Namespace: "blub"},
+							},
+						},
+					},
+				},
+			},
+			wantErr: true,
+		},
+		{
+			name: "multiple named refs: OK",
```
robbavey, line 358:
```suggestion
			name: "multiple refs, each with clusterName: OK",
```

pkg/apis/logstash/v1alpha1/validations_test.go
```diff
@@ -229,3 +229,176 @@ func Test_checkSupportedVersion(t *testing.T) {
...
+								ClusterName:    "blub",
+							},
+						},
+					},
+				},
+			},
+			wantErr: false,
+		},
+		{
+			name: "unamed within multiple: NOK",
```
robbavey, line 378:
```suggestion
			name: "multiple refs, one missing clusterName: NOK",
```

robbavey (changes requested):
This is great - really close, I have a couple of nits on clarifying the messaging on missing `clusterName`

robbavey (approved):
LGTM from the logstash side

kaisecheng:
@pebrc @barkbay I have added `clusterName` and gotten thumbs up from Rob. This is ready to review.

robbavey:
@kaisecheng One thing I noticed, that I should have noticed earlier 🤦 - we use `USERNAME` as the suffix in the env variable, but the plugin settings use `user` for the configuration settings. What do you think about having `USER` for the env variable?

kaisecheng:
@robbavey I have changed the suffix to `USER`

robbavey (approved):
LGTM

robbavey:
@pebrc @barkbay This is ready for final review by the ECK team - I'm happy with this from the logstash side

pkg/apis/logstash/v1alpha1/logstash_types.go
```diff
@@ -92,6 +96,12 @@ type LogstashService struct {
 	TLS commonv1.TLSOptions `json:"tls,omitempty"`
 }
 
+// ElasticsearchCluster definition of type to define an Elasticsearch Cluster to be referred to in Logstash pipeline
+type ElasticsearchCluster struct {
+	commonv1.ObjectSelector `json:",omitempty,inline"`
+	ClusterName             string `json:"clusterName,omitempty"`
```
barkbay, line 102:
```suggestion
	// +kubebuilder:validation:Required
	// +kubebuilder:validation:MinLength=1
	ClusterName             string `json:"clusterName,omitempty"`
```

pkg/apis/logstash/v1alpha1/validations.go
```diff
@@ -86,3 +90,19 @@ func checkSinglePipelineSource(a *Logstash) field.ErrorList {
...
+			notNamed = append(notNamed, o.NamespacedName().String())
+		}
+	}
+	if len(notNamed) > 0 {
+		msg := fmt.Sprintf("clusterName is a mandatory field - missing on %v", notNamed)
+		return field.ErrorList{
+			field.Forbidden(field.NewPath("spec").Child("elasticsearchRefs"), msg),
+		}
+	}
+	return nil
```
barkbay, line 107:
```suggestion
	var errorList field.ErrorList
	for i, esRef := range l.Spec.ElasticsearchRefs {
		if esRef.ClusterName == "" {
			errorList = append(
				errorList,
				field.Required(
					field.NewPath("spec").Child("elasticsearchRefs").Index(i).Child("clusterName"),
					fmt.Sprintf("clusterName is a mandatory field - missing on %v", esRef.NamespacedName())),
			)
		}
	}
	return errorList
```

barkbay (approved):
LGTM

pkg/apis/logstash/v1alpha1/logstash_types.go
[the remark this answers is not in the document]
kaisecheng: would like to improve it in future

PR #6670, opened 2023-04-10
Sync images tagged as latest when dry-run is not set.

### LINKED ISSUE #6665: Multiple operatorhub/quay.io releases are seen as "latest".
related #6511 

After resolving the issue where we were not tagging the `latest` release in operatorhub environment, and after the release of `2.7.0`, we are seeing that multiple versions are now tagged as `latest`. This seems contradictory as to how most container registries work where only one can be tagged as `latest`. We need to investigate why this is happening during release, and ensure only the latest release is tagged as `latest`.

image

naemono:
resolves #6665

We recently noticed that we were seeing multiple images tagged as `latest` after publishing a new operatorhub release. We appear to need to run a `sync-tags` operation on any tags that are not considered `latest` after the release of a new release. This handles that operation.

### DIFF
--- a/hack/operatorhub/internal/container/container.go
+++ b/hack/operatorhub/internal/container/container.go
@@ -28,6 +28,9 @@ const (
 	httpXAPIKeyHeader              = "X-API-KEY"
 	httpApplicationJSONHeaderValue = "application/json"
 	getImagesFilter                = "repositories.tags.name==%s"
+	publishOperation               = "publish"
+	syncTagsOperation              = "sync-tags"
+	latestTag                      = "latest"
 )
 
 var (
@@ -46,6 +49,7 @@ type CommonConfig struct {
 	HTTPClient          *http.Client
 	ProjectID           string
 	RedhatCatalogAPIKey string
+	APIURL              string
 	RegistryPassword    string
 	Tag                 string
 }
@@ -70,8 +74,14 @@ func PushImage(c PushConfig) error {
 		c.HTTPClient = defaultHTTPClient()
 	}
 
+	commonConfig := CommonConfig{
+		APIURL:              catalogAPIURL,
+		HTTPClient:          c.HTTPClient,
+		RedhatCatalogAPIKey: c.RedhatCatalogAPIKey,
+		ProjectID:           c.ProjectID,
+		Tag:                 c.Tag}
 	log.Printf("Determining if image already exists within project with tag: %s", c.Tag)
-	exists, err := ImageExistsInProject(c.HTTPClient, c.RedhatCatalogAPIKey, c.ProjectID, c.Tag)
+	exists, err := ImageExistsInProject(commonConfig)
 	if err != nil {
 		log.Println("ⅹ")
 		return fmt.Errorf("failed to determine if image exists: %w", err)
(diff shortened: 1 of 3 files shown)


hack/operatorhub/internal/container/container.go
```diff
@@ -342,22 +390,26 @@ func isImageScanned(httpClient *http.Client, apiKey, projectID, tag string) (ima
 	return nil, false, nil
 }
 
+func doPublish(image *Image, c CommonConfig) error {
+	return doOperationForImage(image, c, publishOperation)
+}
```
thbkrkr, line 395: Nit: Not sure of the interest of this minimal function which creates an indirection for not much.

hack/operatorhub/internal/container/types.go
```diff
@@ -40,10 +40,21 @@ type Repository struct {
...
 }
+
+func (t Tags) containsName(name string) bool {
+	for _, tag := range t {
+		if tag.Name == name {
+			return true
+		}
+	}
+	return false
+}
```
thbkrkr, line 60: We could take more advantage of the `Tag` type and have:

```suggestion
func (ts Tags) contains(tag Tag) bool {
	for _, t := range ts {
		if t.Name == tag.Name {
			return true
		}
	}
	return false
}
```

hack/operatorhub/internal/container/container.go
```diff
@@ -101,12 +116,17 @@ func PublishImage(c PublishConfig) error {
 		c.HTTPClient = defaultHTTPClient()
 	}
 
-	return publishImageInProject(c.HTTPClient, c.ImageScanTimeout, c.RedhatCatalogAPIKey, c.ProjectID, c.Tag, c.DryRun)
+	return publishImageInProject(CommonConfig{
+		HTTPClient:          c.HTTPClient,
+		RedhatCatalogAPIKey: c.RedhatCatalogAPIKey,
+		ProjectID:           c.ProjectID,
+		Tag:                 c.Tag,
+	}, c.ImageScanTimeout, c.DryRun)
```
thbkrkr, line 124: It's a bit strange having to recreate the `CommonConfig` here don't you think?

thbkrkr (commented):
Sounds good to me, except I found it a bit difficult to understand why we need to recreate `CommonConfig`. Wouldn't it be simpler to just remove the API URL and HTTP client from it or set those two once and for all at the beginning. I'm also not sure about the benefit to have the `PushConfig` and the `PublishConfig` for one more setting that can be passed as an argument to the `PushImage` and `PublishImage` functions. We could also take more advantage of the type struct `Tag`. I would see it like that: https://github.com/thbkrkr/cloud-on-k8s/commit/40baa3ab0e8f4b6add6c7d9952d4008102e58cf3.

naemono:
Thanks @thbkrkr .

I've gone with most of your suggestions, with a few minor changes:

If we're going to go with this
`func getImagesByTag(c CommonConfig, tag Tag) ([]Image, error) {`
then I've just removed the `Tag` field from the `CommonConfig` and passed it in as an argument to the exported funcs.

Lmk what you think.

thbkrkr (approved):
LGTM

PR #6674, opened 2023-04-10
Logstash - add ability to reload pipeline(s) without triggering full pod restart

robbavey:
This commit adds the ability to reload logstash pipelines when the pipeline changes, without triggering a full restart of the pod, leveraging Logstash's ability to watch pipeline definitions, and reload automatically if a change is discovered.

A logstash config directory typically includes `logstash.yml`, `pipelines.yml`, `jvm.options` and `log4j2.properties` required to run logstash - while logstash can store the contents of a pipeline definition in any location, the `pipelines.yml` definition file, which states where these definition files are must be in the same config directory as the other setup files. 

To enables us to have a mixture of copied and generated files in this config diretory, an `initContainer` is used, with a small script to prepare the config directory. 
* The script copies the contents of the `/usr/share/logstash/config` into the a shared config volume
* Additionally softlinks are created into this volume from the `pipelines.yml` and `logstash.yml` secrets created by the logstash operator. 

Additionally, we now do not include changes to pipelines in the configuration hash that triggers a pod reload, but instead write the `config.reload.automatic: true` setting in to `logstash.yml`

Note that triggering a reload is not immediate - there may be a delay measured in minutes between the pipeline definition being changed, and that being reflected in a pipeline reload.

### DIFF
--- a/pkg/controller/logstash/driver.go
+++ b/pkg/controller/logstash/driver.go
@@ -88,7 +88,10 @@ func internalReconcile(params Params) (*reconciler.Results, logstashv1alpha1.Log
 		return results.WithError(err), params.Status
 	}
 
-	if err := reconcilePipeline(params, configHash); err != nil {
+	// We intenetionally DO NOT pass the configHash here. We don't want to consider the pipeline definitions in the
+	// hash of the config to ensure that a pipeline change does not automatically trigger a restart
+	// of the pod, but allows Logstash's automatic reload of pipelines to take place
+	if err := reconcilePipeline(params); err != nil {
 		return results.WithError(err), params.Status
 	}
 
--- a/pkg/controller/logstash/driver.go
+++ b/pkg/controller/logstash/driver.go
@@ -88,7 +88,7 @@ func internalReconcile(params Params) (*reconciler.Results, logstashv1alpha1.Log
 		return results.WithError(err), params.Status
 	}
 
-	// We intenetionally DO NOT pass the configHash here. We don't want to consider the pipeline definitions in the
+	// We intentionally DO NOT pass the configHash here. We don't want to consider the pipeline definitions in the
 	// hash of the config to ensure that a pipeline change does not automatically trigger a restart
 	// of the pod, but allows Logstash's automatic reload of pipelines to take place
 	if err := reconcilePipeline(params); err != nil {
--- a/pkg/controller/logstash/initcontainer.go
+++ b/pkg/controller/logstash/initcontainer.go
@@ -16,7 +16,7 @@ const (
 
 	// InitConfigScript is a small bash script to prepare the logstash configuration directory
 	InitConfigScript = `#!/usr/bin/env bash
-set -eux
+set -eu
 
 init_config_initialized_flag=` + InitContainerConfigVolumeMountPath + `/elastic-internal-init-config.ok
 
--- a/pkg/controller/logstash/initcontainer.go
+++ b/pkg/controller/logstash/initcontainer.go
@@ -22,19 +22,14 @@ init_config_initialized_flag=` + InitContainerConfigVolumeMountPath + `/elastic-
 
 if [[ -f "${init_config_initialized_flag}" ]]; then
     echo "Logstash configuration already initialized."
-	exit 0
+    exit 0
 fi
 
 echo "Setup Logstash configuration"
 
 mount_path=` + InitContainerConfigVolumeMountPath + `
 
-for f in /usr/share/logstash/config/*.*; do
-	filename=$(basename $f)
-	if [[ ! -f "$mount_path/$filename" ]]; then
-		cp $f $mount_path
-	fi
-done
+cp -f /usr/share/logstash/config/*.* "$mount_path"
 
 ln -sf ` + InternalConfigVolumeMountPath + `/logstash.yml  $mount_path
 ln -sf ` + InternalPipelineVolumeMountPath + `/pipelines.yml  $mount_path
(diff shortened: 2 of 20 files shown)


kaisecheng:
@robbavey I tried to run logstash.yaml sample and update the Secret `logstash-sample-ls-pipeline` with `kubectl apply -f`, but unable to update the pipeline to test the reload. Do you have any trick to update the pipeline?

robbavey:
@kaisecheng It takes a surprisingly long time - I've seen it take up to 3 minutes for the change to get picked up. For manual testing, I typically make and apply the change then watch the pod logs.

pebrc:
> It takes a surprisingly long time - I've seen it take up to 3 minutes for the change to get picked up

This could be due to the kubelets sync interval which is 60 seconds plus some jitter and I guess there is probably also some delay on the Logstash side before the change is detected in the filesystem. We use a trick for Elasticsearch to force an immediate sync for the secrets where timely sync is important: we annotate the Pods (we use a timestamp but the actual annotation is irrelevant) which forces a sync. 

I am not sure if pipeline changes need such urgent propagation to warrant such extra complexity but I thought I mention it: 

https://github.com/elastic/cloud-on-k8s/blob/85aa75b7a7765330475c0b92a0851939ac3d9578/pkg/controller/elasticsearch/certificates/transport/reconcile.go#L166

kaisecheng (commented):
Thanks for taking care of the pipeline reload issue. It works as expected. I have tested pipeline updates through `pipelines` and `pipelinesRef`. Both take 1 ~ 2 minutes locally to reflect the change.  The  e2e test looks good. I run it with `export E2E_TAGS=e2e; make e2e-local TESTS_MATCH=TestPipelineConfigLogstash`.

The only question on my mind is whether we should enable `config.reload.automatic: true`  in ECK by default, because it is not enabled in docker nor any distribution

robbavey:
@kaisecheng It's a good question, and worth discussing - 

I think when a change is made to a `pipeline` or `pipelineRef` in the CRD, we can choose to 
* Restart the pipeline(s)
* Restart the Pod 
* Do nothing

My thoughts here are that a change to a definition in the CRD, such as a change to a `pipeline` or `pipelineRef` is a deliberate action, that a user might expect to be applied on without manual intervention - similar to a change in `config` or `configRef`, which would trigger a restart of a pod. 

This change allows that change to be acted upon with as little disruption as we can get away with - we can now limit the change to pipeline(s), rather than restarting the whole pod. If we don't set `config.reload.automatic`, a user would have to intervene manually, and delete the pod(s) to have the pipeline change come into effect, if we don't restart the pod.

But, let's discuss

kaisecheng:
Agree that having pipeline auto-reload is a better experience. Sadly, it is not the default in Logstash. When `config.reload.automatic` is set to `false`, changes in pipeline do not restart the pipeline nor restart the pod.

Users may expect pod restart when pipeline changes, just like changing in `config` `configRef` results in a pod restart.
I think making pod restart when reload is set to `false` and pipeline restart when reload is set to `true` is truly making reloading a configurable flag. It can be done by passing the boolean flag to `reconcilePipeline()` to update `configHash`.

I am still not sure whether changing the default reload behavior only in ECK is right, but open to this change.

pebrc:
buildkite test this -f p=gke,t=TestLogstashPipelineReload -m s=8.7.0

pebrc (approved):
LGTM

pkg/controller/logstash/initcontainer.go
```diff
@@ -0,0 +1,80 @@
...
+
+	logstashv1alpha1 "github.com/elastic/cloud-on-k8s/v2/pkg/apis/logstash/v1alpha1"
+)
+
+const (
+	InitConfigContainerName = "logstash-internal-init-config"
+
+	// InitConfigScript is a small bash script to prepare the logstash configuration directory
+	InitConfigScript = `#!/usr/bin/env bash
+set -eux
```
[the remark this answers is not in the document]
robbavey: Nope, there for debugging. Thanks!

thbkrkr (commented):
Damn, I had reviewed this two days ago but didn't submit it sorry.
Looks good! I spotted a few minor things, nothing blocking to merge.

pkg/controller/logstash/driver.go
```diff
@@ -88,7 +88,10 @@ func internalReconcile(params Params) (*reconciler.Results, logstashv1alpha1.Log
 		return results.WithError(err), params.Status
 	}
 
-	if err := reconcilePipeline(params, configHash); err != nil {
+	// We intentionally DO NOT pass the configHash here. We don't want to consider the pipeline definitions in the
+	// hash of the config to ensure that a pipeline change does not automatically trigger a restart
+	// of the pod, but allows Logstash's automatic reload of pipelines to take place
+	if err := reconcilePipeline(params); err != nil {
```
thbkrkr, line 94: Should we pass the `configHash` when `config.reload.automatic`equals false?

robbavey: It's a good question. I'm erring on the side of 'no' at the moment, but I think this is something that could change after the technical preview depending on feedback.

My reasoning on this is that the `false` (default) value of non-k8s logstash doesn't react to pipeline changes at all, and to change this semantic to restart logstash completely on pipeline changes feels like very different behaviour.

Thinking about how we could add flexibility, I wonder if we might want to introduce something for ECK here, along the lines of: 

`config.reload.restart_policy: detected_only|all|none`, which would either set `config.reload.automatic: true` for `detected_only`, and `false` for `all` or `none`, passing the `configHash` if the value is `all`, and not if it is `none`.

cc @flexitrev, @roaksoax, @jsvd

pebrc (approved):

PR #6688, opened 2023-04-13
Default to running Elasticsearch as `ReadOnlyRootFilesystem: true`

### LINKED ISSUE #6126: Cannot run elasticsearch with security best practice ``readOnlyRootFilesystem: true``
## Bug Report

**What did you do?**

As a good practice, we configure security context (here ``readOnlyRootFilesystem: true``):

```
kind: Elasticsearch
spec:
  nodeSets:
    - 
      count: 3
      name: master
      podTemplate:
        spec:
          containers:
            - 
              securityContext: &securityContext
                allowPrivilegeEscalation: false
                capabilities:
                  drop:
                    - ALL
                privileged: false
                readOnlyRootFilesystem: true
                runAsNonRoot: true
                runAsUser: 1000
          initContainers:
            - name: elastic-internal-init-filesystem
              securityContext: 
                <<: *securityContext
            - name: elastic-internal-suspend
              securityContext:
                <<: *securityContext
            - name: elastic-internal-init-keystore
              securityContext:
                <<: *securityContext
```

Operator create the STS, create pods, create container and init container can run fine.

The first initContainer cannot run because a folder is not writable. Diging the source code, we can see:

prepare-fs.sh script:

```
...
echo "Linking /mnt/elastic-internal/xpack-file-realm/users to /usr/share/elasticsearch/config/users"
		ln -sf /mnt/elastic-internal/xpack-file-realm/users /usr/share/elasticsearch/config/users
...
```
where the directory `` /usr/share/elasticsearch/config`` is not a volume, hence read-only.

We cant find any workaround, except changing polaris or OPA rules.

As a clean solution, the prepare-fs.sh script should use a temporary directory, mounted as ``emptyDir``.

References:

* https://github.com/elastic/cloud-on-k8s/issues/2108
* https://github.com/elastic/cloud-on-k8s/issues/4022

naemono:
## Primary Change

This enables running Elasticsearch with `readOnlyRootFilesystem: true` as this aligns with security best practices.

## Todo
- [ ] Full set of e2e tests on both 7.x and 8.x
- [ ] Documentation

related #6126

### DIFF
--- a/pkg/controller/elasticsearch/nodespec/podspec.go
+++ b/pkg/controller/elasticsearch/nodespec/podspec.go
@@ -128,6 +128,7 @@ func BuildPodTemplateSpec(
 		enableLog4JFormatMsgNoLookups(builder)
 	}
 
+	// default to running the "elasticsearch" container as "readOnlyRootFilesystem: true"
 	mainContainer := builder.MainContainer()
 	if mainContainer != nil {
 		mainContainer.SecurityContext = &corev1.SecurityContext{
--- a/pkg/controller/elasticsearch/nodespec/podspec.go
+++ b/pkg/controller/elasticsearch/nodespec/podspec.go
@@ -12,7 +12,6 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
-
 	"k8s.io/utils/pointer"
 
 	esv1 "github.com/elastic/cloud-on-k8s/v2/pkg/apis/elasticsearch/v1"
--- a/pkg/controller/elasticsearch/nodespec/podspec.go
+++ b/pkg/controller/elasticsearch/nodespec/podspec.go
@@ -13,6 +13,8 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
 
+	"k8s.io/utils/pointer"
+
 	esv1 "github.com/elastic/cloud-on-k8s/v2/pkg/apis/elasticsearch/v1"
 	"github.com/elastic/cloud-on-k8s/v2/pkg/controller/common/annotation"
 	"github.com/elastic/cloud-on-k8s/v2/pkg/controller/common/container"
@@ -28,7 +30,6 @@ import (
 	"github.com/elastic/cloud-on-k8s/v2/pkg/controller/elasticsearch/stackmon"
 	esvolume "github.com/elastic/cloud-on-k8s/v2/pkg/controller/elasticsearch/volume"
 	"github.com/elastic/cloud-on-k8s/v2/pkg/utils/k8s"
-	"github.com/elastic/cloud-on-k8s/v2/pkg/utils/pointer"
 )
 
 const (
@@ -125,6 +126,13 @@ func BuildPodTemplateSpec(
 	if ver.LT(version.From(7, 2, 0)) {
 		// mitigate CVE-2021-44228
 		enableLog4JFormatMsgNoLookups(builder)
+	}
+
+	mainContainer := builder.MainContainer()
+	if mainContainer != nil {
+		mainContainer.SecurityContext = &corev1.SecurityContext{
+			ReadOnlyRootFilesystem: pointer.Bool(true),
+		}
 	}
 
 	return builder.PodTemplate, nil
--- a/pkg/controller/elasticsearch/nodespec/podspec_test.go
+++ b/pkg/controller/elasticsearch/nodespec/podspec_test.go
@@ -15,6 +15,7 @@ import (
 	"github.com/stretchr/testify/require"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	k8s_pointer "k8s.io/utils/pointer"
 
 	commonv1 "github.com/elastic/cloud-on-k8s/v2/pkg/apis/common/v1"
 	esv1 "github.com/elastic/cloud-on-k8s/v2/pkg/apis/elasticsearch/v1"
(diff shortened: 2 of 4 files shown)


naemono:
buildkite test this -f p=gke,s=8.7.0,s=7.17.8

barkbay (commented):
> resolves https://github.com/elastic/cloud-on-k8s/issues/6126

nit: I think this change is a good thing, but unless I'm missing something it does not solve #6126. The requirement is to be able to run the whole `Pod` with `readOnlyRootFilesystem: true`, which is for now not possible for the init container `elastic-internal-init-filesystem` (this is something I missed during my initial review of the issue).

I'm also wondering if it would be a good thing to include other security settings like:

```yaml
securityContext:
  allowPrivilegeEscalation: false
  capabilities:
    drop:
      - ALL
  privileged: false
  readOnlyRootFilesystem: true
  runAsNonRoot: true
```

naemono:
Thanks for the heads up about wanting to run the whole pod as read-only @barkbay . 

> which is for now not possible for the init container

So I was reviewing this and am curious about the history of the approach we take. From what I see happening in this init container that would violate having it mounted as read-only are these copy operations:

```
# these are all from the xpack secret, and are separate keys within the secret itself:
Linking /mnt/elastic-internal/xpack-file-realm/users to /usr/share/elasticsearch/config/users
Linking /mnt/elastic-internal/xpack-file-realm/roles.yml to /usr/share/elasticsearch/config/roles.yml
Linking /mnt/elastic-internal/xpack-file-realm/users_roles to /usr/share/elasticsearch/config/users_roles
# This is from the 'config' secret, and is a key within the secret
Linking /mnt/elastic-internal/elasticsearch-config/elasticsearch.yml to /usr/share/elasticsearch/config/elasticsearch.yml
# This is a key within the unicast configmap
Linking /mnt/elastic-internal/unicast-hosts/unicast_hosts.txt to /usr/share/elasticsearch/config/unicast_hosts.txt
# This is another from the xpack secret
Linking /mnt/elastic-internal/xpack-file-realm/service_tokens to /usr/share/elasticsearch/config/service_tokens
File linking duration: 0 sec.
```

Why are we not simply mounting these keys as files from the secret/cm into this location?

For example (untested atm):
```
volumeMounts:
- mountPath: /usr/share/elasticsearch/config
   name: xpack-users
   readOnly: true
volumes:
  - name: xpack-users
     secret:
        secretName: elasticsearch-es-xpack-file-realm
        items:
        - key: users
          path: users
```

barkbay:
> Why are we not simply mounting these keys as files from the secret/cm into this location?

`mountPath` must be unique, and all the original content of `/usr/share/elasticsearch/config` would be wiped out. I think the idea is to keep the original content of `/usr/share/elasticsearch/config`, as it is provided by the Elasticsearch image, and update only what is necessary.

Also note that `subPath` cannot be used as the content would not be updated:

> Note: A container using a Secret as a [subPath](https://kubernetes.io/docs/concepts/storage/volumes#using-subpath) volume mount does not receive automated Secret updates.

naemono:
> Also note that `subPath` cannot be used as the content would not be updated:

So the solution to one of the 2 issues (that we know of) is having a separate secret for each of these files that are needed to be propagated into the config directory and cannot be renamed or pointed to a different file in the `elasticsearch.yml`. 

As for the issue of the configuration directory we can adjust this via `ES_PATH_CONF` environment variable and use a mount path outside of `root` but that would involve a full redesign of how we're handling ES configuration.  These are the files I see inside of existing `config` directory btw:

```
config/role_mapping.yml
config/service_tokens
config/transport-certs/
config/log4j2.properties
config/jvm.options
config/jvm.options.d
config/elasticsearch-plugins.example.yml
config/transport-remote-certs/
config/node-transport-cert/
config/elasticsearch.yml
config/elasticsearch.keystore
config/unicast_hosts.txt
config/users_roles
config/roles.yml
config/http-certs/
config/users
config/log4j2.file.properties
```

How would you feel about this proposal @barkbay :

1. solving the existing issues seen in the e2e test run with just running the main container as `readOnlyRootFilesystem: true`.
2. Investigating adding the additional security features you mentioned above.
3. adjusting the existing issue to be an issue about redesigning how we do ES configuration to allow `readOnlyRootFilesystem: true`
4. merging this as an incremental step in the direction of running the whole pod as readonly?

Jorricks:
> I'm also wondering if it would be a good thing to include other security settings like:
> 
> ```yaml
> securityContext:
>   allowPrivilegeEscalation: false
>   capabilities:
>     drop:
>       - ALL
>   privileged: false
>   readOnlyRootFilesystem: true
>   runAsNonRoot: true
> ```

Just jumping in here. That would be awesome. Currently we are enforcing our namespaces to have a [restricted pod security standard](https://kubernetes.io/docs/concepts/security/pod-security-standards/). However, this operator breaks under this enforcement without the above.

barkbay:
> Currently we are enforcing our namespaces to have a [restricted pod security standard](https://kubernetes.io/docs/concepts/security/pod-security-standards/).

Unfortunately `runAsNonRoot`  is not that straightforward as the Elasticsearch image is running as a non numeric user:

```
elasticsearch-sample-es-default-2   0/1     Init:CreateContainerConfigError   0          3m37s
```

```
Warning  Failed     2m10s (x12 over 3m48s)  kubelet            Error: container has runAsNonRoot and image has non-numeric user (elasticsearch), cannot verify user is non-root
```

But maybe this is something we can discuss with the Elasticsearch team.

naemono:
@barkbay https://github.com/elastic/elasticsearch/pull/95390 opened against Elasticsearch for this change after speaking with this and seeing no reason to not make this change.

thbkrkr:
Superseeded by #6703.

PR #6689, opened 2023-04-14
Use provided base path for stackconfigpolicy's snapshot repository

### LINKED ISSUE #6692: base path setting on stackconfigpolicy snapshot repository
For cloud repositories (azure, gcs, s3), ECK sets `base_path` to `snapshots/<namespace>-<esName>` and it's not possible to override it. This is problematic for example to configure a read-only repository where a user wants full control of the `base_path`.

https://github.com/elastic/cloud-on-k8s/blob/ae88fe1662ed9f04fabc063edfdaaa4984ddb6d1/pkg/controller/elasticsearch/filesettings/file_settings.go#L148

I missed the original intent that was to only add `<namespace>-<esName>` to the user provided `base_path` like it's done for `location` for `fs` repository and `path` for `hdfs` repository. But this is also a problem if the need is to fully control the `base_path`.

I could see 3 cases, but I'm not sure we need to deal with the third one, which makes everything more complex because we need to be able to distinguish it from the second one.

| | user provided `base_path` | configured `base_path` |
|---|---|---|
| 1) no base_path |  | `snapshots/ns1-es1`|
| 2) full base_path | `some/thing` | `some/thing` |
| 3) partial base_path | `some/thing` | `some/thing/ns1-es1` |

thbkrkr:
This changes how ECK managed the `base_path` setting for stackconfigpolicy's cloud snapshot repositories.

Instead of always setting it to `snapshots/<namespace>-<esName>`, it will only set it when the user hasn't provide one.

The documentation is updated to mention this specificity.

Resolves #6692.

### DIFF
--- a/docs/orchestrating-elastic-stack-applications/stack-config-policy.asciidoc
+++ b/docs/orchestrating-elastic-stack-applications/stack-config-policy.asciidoc
@@ -35,7 +35,7 @@ Elastic Stack configuration policies can be defined in a `StackConfigPolicy` res
 * `name` is a unique name used to identify the policy.
 * `spec.elasticsearch` describes the settings to configure and at least one setting must be defined. Each of the following fields except `clusterSettings` is an associative array where keys are arbitrary names and values are definitions:
   ** `clusterSettings` are the settings that go into the elasticsearch.yml file.
-  ** `snapshotRepositories` are snapshot repositories for defining an off-cluster storage location for your snapshots (check <<{p}-{page_id}-specifics-snap-repo,specifics>>).
+  ** `snapshotRepositories` are snapshot repositories for defining an off-cluster storage location for your snapshots. Check <<{p}-{page_id}-specifics-snap-repo>> for more information.
   ** `snapshotLifecyclePolicies` are snapshot lifecycle policies, to automatically take snapshots and control how long they are retained.
   ** `securityRoleMappings` are role mappings, to define which roles are assigned to each user by identifying them through rules.
   ** `ingestPipelines` are ingest pipelines, to perform common transformations on your data before indexing.
@@ -239,7 +239,8 @@ Important events are also reported through Kubernetes events, such as when two c
 [id="{p}-{page_id}-specifics-snap-repo"]
 == Specifics for snapshot repositories
 
-In order to avoid a conflict between multiple Elasticsearch clusters writing their snapshot to the same location, ECK automatically:
-- sets the `base_path` to `snapshots/<namespace><esName>` when it is not provided, for Azure, GCS and S3 repositories
-- appends `<namespace><esName>` to `location` for a FS repository
-- appends `<namespace><esName>` to `path` for an `HDFS` repository
+In order to avoid a conflict between multiple Elasticsearch clusters writing their snapshots to the same location, ECK automatically:
+
+- sets the `base_path` to `snapshots/<namespace>-<esName>` when it is not provided, for Azure, GCS and S3 repositories
+- appends `<namespace>-<esName>` to `location` for a FS repository
+- appends `<namespace>-<esName>` to `path` for an HDFS repository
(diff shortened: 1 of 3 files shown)


thbkrkr:
buildkite test this -f p=kind,t=TestStackConfigPolicy

pkg/controller/elasticsearch/filesettings/file_settings.go
```diff
@@ -145,7 +145,11 @@ func mutateSnapshotRepositorySettings(snapshotRepository map[string]interface{},
 	}
 	switch snapshotRepository["type"] {
 	case "azure", "gcs", "s3":
-		settings["base_path"] = fmt.Sprintf("snapshots/%s", suffix)
+		basePath, ok := settings["base_path"].(string)
```
barkbay, line 148: Should we also update the documentation to explain how `base_path` is updated by the operator?

thbkrkr: Yes.

thbkrkr:
This is good for another review. I changed to let the user fully control the `base_path` and updated the doc.

docs/orchestrating-elastic-stack-applications/stack-config-policy.asciidoc
```diff
@@ -234,3 +234,12 @@ Important events are also reported through Kubernetes events, such as when two c
...
 ----
+
+[float]
+[id="{p}-{page_id}-specifics-snap-repo"]
+== Specifics for snapshot repositories
+
+In order to avoid a conflict between multiple Elasticsearch clusters writing their snapshot to the same location, ECK automatically:
+- sets the `base_path` to `snapshots/<namespace><esName>` when it is not provided, for Azure, GCS and S3 repositories
+- appends `<namespace><esName>` to `location` for a FS repository
+- appends `<namespace><esName>` to `path` for an `HDFS` repository
```
barkbay, line 245: * Something is wrong in the final rendering, I think a `\n` is missing before the list.
image

* I think the hyphen is missing as well (`uniqSuffix := fmt.Sprintf("%s-%s", esNs, esName)`)?

* Should we make `snapshot` the word plural?

```suggestion
In order to avoid a conflict between multiple Elasticsearch clusters writing their snapshots to the same location, ECK automatically:

- sets the `base_path` to `snapshots/<namespace>-<esName>` when it is not provided, for Azure, GCS and S3 repositories.
- appends `<namespace>-<esName>` to `location` for a FS repository.
- appends `<namespace>-<esName>` to `path` for an `HDFS` repository.
```

docs/orchestrating-elastic-stack-applications/stack-config-policy.asciidoc
```diff
@@ -35,7 +35,7 @@ Elastic Stack configuration policies can be defined in a `StackConfigPolicy` res
 * `name` is a unique name used to identify the policy.
 * `spec.elasticsearch` describes the settings to configure and at least one setting must be defined. Each of the following fields except `clusterSettings` is an associative array where keys are arbitrary names and values are definitions:
   ** `clusterSettings` are the settings that go into the elasticsearch.yml file.
-  ** `snapshotRepositories` are snapshot repositories for defining an off-cluster storage location for your snapshots.
+  ** `snapshotRepositories` are snapshot repositories for defining an off-cluster storage location for your snapshots (check <<{p}-{page_id}-specifics-snap-repo,specifics>>).
```
barkbay, line 38: nit: I have a small preference for the link name to be automatically synchronized:

```suggestion
  ** `snapshotRepositories` are snapshot repositories for defining an off-cluster storage location for your snapshots. Check <<{p}-{page_id}-specifics-snap-repo>> for more information.
```

⬇️ 

image

barkbay (approved):
LGTM, I was wondering if this change should be mentioned in the release notes, as the behavior will change for users who attempted to set a `base_path` and inadvertently kept it in their configs.

docs/orchestrating-elastic-stack-applications/stack-config-policy.asciidoc
thbkrkr: Triple-yes. 🙇

thbkrkr:
> if this change should be mentioned in the release notes, as the behavior will change for users who attempted to set a base_path and inadvertently kept it in their configs.

Good point. I created an issue:

- https://github.com/elastic/cloud-on-k8s/issues/6702

PR #6704, opened 2023-04-19
Helm Release: Fail to write files to bucket in some scenarios

naemono:
We recently had a report of the `digest/sha256sum` of `2.7.0` release reported in the Helm index of Helm Chart didn't match the actual `tgz` file itself. This is happening because we are not overwriting existing `tgz` files if they already exist.

So we are
1. Creating a new .tgz file
2. Indexing the file(s) with their new digest
3. Attempting to upload the new .tgz file, but not explicitly overwriting it
4. Uploading the new index with the new digest which doesn't match what's in the bucket.

This change allows these 2 paths:
1. If the helm chart contains `SNAPSHOT` no precondition is set, which means google will not fail when overwriting a file.
2. Otherwise we assume this is a production/tagged release and we will set a precondition, which if encountered, we throw an error and fail the whole operation.

Resolves https://github.com/elastic/cloud-on-k8s/issues/6714.

### DIFF
--- a/hack/helm/release/internal/helm/helm.go
+++ b/hack/helm/release/internal/helm/helm.go
@@ -10,7 +10,6 @@ import (
 	"io"
 	"io/ioutil"
 	"log"
-	"net/http"
 	"net/url"
 	"os"
 	"path/filepath"
@@ -20,7 +19,6 @@ import (
 	"cloud.google.com/go/storage"
 	"github.com/avast/retry-go/v4"
 	"golang.org/x/exp/slices"
-	"google.golang.org/api/googleapi"
 	"gopkg.in/yaml.v3"
 	"helm.sh/helm/v3/pkg/action"
 	"helm.sh/helm/v3/pkg/cli"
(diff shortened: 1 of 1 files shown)


hack/helm/release/internal/helm/helm.go
```diff
@@ -479,6 +479,8 @@ func copyChartToGCSBucket(ctx context.Context, config copyChartToBucketConfig) e
 	if len(files) == 0 {
 		return fmt.Errorf("couldn't file helm package with glob (%s)", config.sourceGlob)
 	}
+	isSnapshot := strings.Contains(files[0], "SNAPSHOT")
```
thbkrkr, line 482: I'm a little paranoid and so this test seems a little weak to me, but in this context it should be more than enough.

thbkrkr (approved):
LGTM

hack/helm/release/internal/helm/helm.go
naemono: Yeah, in this block we only have access to the tarball, so it would be a bit more work to try and get access to the Chart.yaml and check if the `version` field ends in `-SNAPSHOT`. This seems sufficient for our use-case.

I've also updated the comments on this func, as they were incorrect with this current change, and will merged when CI is complete.

thbkrkr: We could search for `-SNAPSHOT.tgz` :man_shrugging:.

naemono: Easy enough. Updated.

PR #6710, opened 2023-04-20
E2E: retry setup of data integrity check

pebrc:
We have seen some test failures on builds where the cluster rejects connection attempts, even though it has previously already responded to health checks.  The setup of the data integrity check is one of the few places where we do not retry requests on failures. I looked at the logs and it seems the requests are made only a few seconds after the cluster reached ready state so I am wondering if we should be more lenient here and allow for retries.

### DIFF
--- a/test/e2e/test/elasticsearch/steps_mutation.go
+++ b/test/e2e/test/elasticsearch/steps_mutation.go
@@ -106,11 +106,16 @@ func (b Builder) MutationTestSteps(k *test.K8sClient) test.StepList {
 	//nolint:thelper
 	steps := test.StepList{
 		test.Step{
-			Name: "Add some data to the cluster before starting the mutation",
+			Name: "Set up a data integrity check",
 			Test: func(t *testing.T) {
 				dataIntegrityCheck = NewDataIntegrityCheck(k, b)
-				require.NoError(t, dataIntegrityCheck.Init())
 			},
+		},
+		test.Step{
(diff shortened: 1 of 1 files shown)


thbkrkr (approved):
Retrying makes sense to me, but I'd still like to have at least one hypothesis that explains why sometimes we can't reach the cluster while the k8s HTTP service is ready and Elasticsearch is green.

I took a quick look at the `Init` func and I see a little code smell, but that probably does no harm:

https://github.com/elastic/cloud-on-k8s/blob/5fc226963c7cf6d47232e850fff5c06c5786e515/test/e2e/test/elasticsearch/checks_data.go#L93-L97

IIUC, the index will usually not exist, so we get an error and we leak a connection by not closing the body.

pebrc:
> but I'd still like to have at least one hypothesis that explains why sometimes we can't reach the cluster while the k8s HTTP service is ready and Elasticsearch is green.

So are you arguing for not retrying? Otherwise we will probably hide this error. Given that it is relatively rare it will be hard to track down.

thbkrkr:
buildkite test this -f p=gke,t=TestCustomTransportCA -m s=8.5.3,s=8.6.2,s=8.7.1

pebrc:
buildkite test this -f p=gke,t=TestCustomTransportCA -m s=8.7.0

thbkrkr (approved):
👍

Issue #6695, opened 2023-04-17
eck.yaml outputs as single-line string since ECK v2.6


PixelOrange:
## Bug Report
**What did you do?**
Download https://download.elastic.co/downloads/eck/2.7.0/operator.yaml

**What did you expect to see?**
```
data:
  eck.yaml: |-
    log-verbosity: 0
    metrics-port: 0
(etc)
```

**What did you see instead? Under which circumstances?**

`eck.yaml: "log-verbosity: 0\nmetrics-port: 0\ncontainer-registry: docker.elastic.co\ncontainer-suffix: \nmax-concurrent-reconciles: 3\nca-cert-validity: 8760h\nca-cert-rotate-before: 24h\ncert-validity: 8760h\ncert-rotate-before: 24h\nexposed-node-labels: [topology.kubernetes.io/.*,failure-domain.beta.kubernetes.io/.*]\nset-default-security-context: auto-detect\nkube-client-timeout: 60s\nelasticsearch-client-timeout: 180s\ndisable-telemetry: false\ndistribution-channel: all-in-one\nvalidate-storage-class: true\nenable-webhook: true\nwebhook-name: elastic-webhook.k8s.elastic.co\nenable-leader-election: true\nelasticsearch-observation-interval: 10s"`

**Environment**

* ECK version: v2.6.0, v2.6.1, v2.6.2, v2.7.0

* Kubernetes information: N/A - issue exists in the text file itself

This makes it much more difficult to edit and adjust the values and flags set in the ConfigMap.

thbkrkr:
It's always been like that, isn't it? I'm not sure to understand the problem? Do you want to manually edit the eck configuration? Don't you have a tool / script / ... to install it?

It seems to me that it's quite simple to use a little bit of bash to update the config on the fly, like this for example:

```sh
> eckconf() { sed "s|$1: [^\]*|$1: $2|"; }

>  curl -s https://download.elastic.co/downloads/eck/2.7.0/operator.yaml \
	| eckconf max-concurrent-reconciles 10 \
	| eckconf kube-client-timeout 120s \
	| k apply -f-

> getcm() { kubectl -n elastic-system get cm elastic-operator -o json | jq '.data["eck.yaml"]' -r; }

> getcm | grep -E "max|kube-client"
max-concurrent-reconciles: 10
kube-client-timeout: 120s
```

or post-install:
```sh
> upcm() {
	local key=$1 val=$2
	getcm > eck.yaml;
[...]

PR #6711, opened 2023-04-20
Conditionally set container-suffix in ECK config

thbkrkr:
This commit ensures that the `config.containerSuffix` Helm value is not empty before writing it to the ECK configMap data to avoid the side effect of the configMap data being formatted on a single line.

I intentionally kept the PR scoped to the bug but we could extend this to all values used in the ECK config.

Resolves to #6695.

### DIFF
--- a/deploy/eck-operator/templates/configmap.yaml
+++ b/deploy/eck-operator/templates/configmap.yaml
@@ -11,7 +11,9 @@ data:
     log-verbosity: {{ int .Values.config.logVerbosity }}
     metrics-port: {{ int .Values.config.metricsPort }}
     container-registry: {{ .Values.config.containerRegistry }}
-    container-suffix: {{ .Values.config.containerSuffix }}
+    {{- with .Values.config.containerSuffix }}
+    container-suffix: {{ . }}
+    {{- end }}
     max-concurrent-reconciles: {{ int .Values.config.maxConcurrentReconciles }}
     {{- if .Values.config.passwordHashCacheSize }}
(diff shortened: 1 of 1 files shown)


deploy/eck-operator/templates/configmap.yaml
```diff
@@ -11,7 +11,9 @@ data:
     log-verbosity: {{ int .Values.config.logVerbosity }}
     metrics-port: {{ int .Values.config.metricsPort }}
     container-registry: {{ .Values.config.containerRegistry }}
-    container-suffix: {{ .Values.config.containerSuffix }}
+    {{- with .Values.config.containerSuffix }}
+    container-suffix: {{ . }}
+    {{- end }}
     max-concurrent-reconciles: {{ int .Values.config.maxConcurrentReconciles }}
     {{- if .Values.config.passwordHashCacheSize }}
```
naemono, line 18: Do we want to change all of these `if` blocks to `with` blocks for consistency?

https://helm.sh/docs/chart_template_guide/control_structures/
`the block after with only executes if the value of PIPELINE is not empty.`

naemono (approved):
lgtm but with one question about consistency

pebrc (approved):

deploy/eck-operator/templates/configmap.yaml
thbkrkr: I find the `with` syntax very nice for readability and figured it was worth starting to use it at the cost of a little break in consistency.

It would be nice to move all the `if` to `with` but this is a low hanging fruit: https://github.com/elastic/cloud-on-k8s/issues/6716.

PR #6717, opened 2023-04-21
Operator leader election: use Leases only

### LINKED ISSUE #4150: Switch leader election to use Leases
Since Kubernetes 1.14 a new leader election method based on the `coordination.k8s.io` API is available and it is the default method in controller-runtime since 0.7.0. We can switch to this new default once support for Kubernetes 1.13 is dropped from the operator.

https://github.com/kubernetes-sigs/controller-runtime/pull/1144

barkbay:
Fix #4150 

Note that this PR does not remove the `ConfigMap`, I wasn't sure if it's really worth making an API call every time the operator starts? (no strong opinion though...)

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -99,7 +99,7 @@ const (
 	DefaultWebhookName = "elastic-webhook.k8s.elastic.co"
 	WebhookPort        = 9443
 
-	LeaderElectionConfigMapName = "elastic-operator-leader"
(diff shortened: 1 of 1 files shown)


thbkrkr (approved):
> I wasn't sure if it's really worth making an API call every time the operator starts? (no strong opinion though...)

Same. Hard to decide. It's not super clean but it doesn't bother. It would imply an API call every time the operator starts, but there is a cache. I'm a little more on the side of it is not worth it.

PR #6720, opened 2023-04-21
Fix service tests

robbavey:
This test was failing, because the readiness probe was hitting the port defined by the service rather than the default port. This commit updates the e2e test to set the config to serve the logtash api from the same port that the service specifies.


test/e2e/logstash/logstash_test.go
```diff
@@ -87,6 +90,9 @@ func TestLogstashWithCustomServiceAndAmendedApi(t *testing.T) {
 
 	logstashBuilder := (logstash.NewBuilder(name).
 		WithNodeCount(1).
+		WithConfig(map[string]interface{}{
```
barkbay, line 93: nit: Could we add a comment/godoc on `TestLogstashWithReworkedApiService` and `TestLogstashWithCustomServiceAndAmendedApi`? It took me a few minutes to figure out why this change is not needed for `TestLogstashWithCustomService`.

test/e2e/test/logstash/builder.go
```diff
@@ -161,6 +161,14 @@ func (b Builder) GetMetricsIndexPattern() string {
 	return ".monitoring-logstash-8-mb"
 }
 
+func (b Builder) WithConfig(config map[string]interface{}) Builder {
+	b.Logstash.Spec.Config = &commonv1.Config{
+		Data: config,
+	}
+
```
barkbay, line 168: nit:
```suggestion
```

barkbay (approved):
LGTM

PR #6724, opened 2023-04-25
[Fleet] Deprecate is_default and is_default_fleet_server flags

### LINKED ISSUE #6678: Remove `is_default` and `is_default_fleet_server` flag from ECK recipes
Flags `is_default` and `is_default_fleet_server` are deprecated since 8.1.

Remove these from ECK recipes:
https://github.com/elastic/cloud-on-k8s/blob/main/config/recipes/elastic-agent/fleet-kubernetes-integration.yaml#L30

The change should be tested, so that Fleet on Kubernetes still works after removing these flags.

_Originally posted by @jen-huang in https://github.com/elastic/ingest-docs/issues/76#issuecomment-1499426546_

barkbay:
This PR:
* Returns a warning in the webhook response if `spec.PolicyID` is not set:

```json
{
	"kind": "AdmissionReview",
	"apiVersion": "admission.k8s.io/v1beta1",
	"response": {
		"uid": "705ab4f5-6393-11e8-b7cc-42010a800002",
		"allowed": true,
		"status": {
			"metadata": {},
			"code": 200
		},
		"warnings": [
			"spec.PolicyID is empty, spec.PolicyID will become mandatory in a future release"
		]
	}
}
```

* Removes `is_default` and `is_default_fleet_server` from our examples and our documentation, adds `spec.PolicyID` instead.

TODO:
- [x] Create an event/log if `is_default` or `is_default_fleet_server` is set.
- [x] There are still some references in:
  - [x]  The Helm Charts
  - [x]  **E2E tests**: I think I will not update [`E2EFleetPolicies`](https://github.com/elastic/cloud-on-k8s/blob/dd2ac2769d54bf008aa519d2253fbbf861d156d0/test/e2e/agent/config.go#L8-L46) as long as the flags are supported. `spec.PolicyID` will be tested as part of the updates in `config/recipes/elastic-agent`, for example with [`TestFleetKubernetesIntegrationRecipe`](https://github.com/elastic/cloud-on-k8s/blob/dd2ac2769d54bf008aa519d2253fbbf861d156d0/test/e2e/agent/recipes_test.go#L90-L136). _**TODO**: create an issue to track this effort once these 2 flags are actually removed_
- [x] For the next major release, should we already skip the default policy mechanism, and make a `spec.PolicyID` mandatory field for any stack version > 8.x.x ?

Fix #6678


barkbay:
This is what the `warning` looks like from a CLI perspective:

image

pkg/apis/agent/v1alpha1/agent_types.go
```diff
@@ -24,6 +24,10 @@ const (
 	FleetServerServiceAccount commonv1.ServiceAccountName = "fleet-server"
 )
 
+var (
+	MandatoryPolicyIDVersion = version.MustParse("9.0.0-SNAPSHOT")
```
pebrc, line 28: Do we know it is going away in 9.0.0 or is this only a placeholder? If so let's add a comment.

pkg/utils/test/events.go
```diff
@@ -0,0 +1,42 @@
...
+	for {
+		select {
+		case e := <-recorder.Events:
+			gotEvents = append(gotEvents, e)
+			gotEventCount++
+			if gotEventCount == minEventCount {
+				// We have read the min. expected number of events
+				return gotEvents
+			}
+		case <-c:
```
pebrc, line 37: Do we need the timeout given that this is always used in a unit test context and there is no expectation of concurrent use? 

```golang
	t.Helper()
	if minEventCount == 0 {
		return nil
	}
	gotEvents := make([]string, 0, minEventCount)
	gotEventCount := 0
	for i := 0; i < minEventCount; i++ {
		select {
		case e := <-recorder.Events:
			gotEvents = append(gotEvents, e)
			gotEventCount++
		default:
			t.Errorf("Timeout: expected at least %d events, got %d", minEventCount, gotEventCount)
		}
	}
	return gotEvents
```

also the name is could be then something along the lines of  `ReadNEvents`

pebrc (commented):
LGTM want to run a few tests still. I like the way you integrated warnings into the webhook tooling 👍

pkg/apis/agent/v1alpha1/agent_types.go
barkbay: Hi @juliaElastic 👋 , could you confirm that `is_default` and `is_default_fleet_server` flags should be removed in `9.0.0`?

juliaElastic: Yes, that is the plan.

pkg/utils/test/events.go
barkbay: > Do we need the timeout given that this is always used in a unit test context and there is no expectation of concurrent use?

I think we need a timeout so we eventually exit the function if there are less events than `minEventCount`. Otherwise, the function may wait forever on `case e := <-recorder.Events`.

pebrc: That's why I added the default clause in the `select` so that it immediately errors out or maybe I am missing your point.

barkbay: Sorry I'm missed your original point. I was assuming that we didn't want to depend on whether or not events were published synchronously in the recorder. But I agree that it is fair to say that it will be the case here.

PR #6727, opened 2023-04-26
Allow custom certificates on the transport layer

### LINKED ISSUE #6479: Consider allowing custom certificates on the transport layer
Existing functionality: 
* ECK will by default generate self-signed x509 certificates for each node of the transport layer in the [format Elasticsearch expects ](https://www.elastic.co/guide/en/elasticsearch/reference/current/security-basic-setup.html#generate-certificates)
* optionally ECK can use a user provided CA and private key to create those certificates https://www.elastic.co/guide/en/cloud-on-k8s/2.6/k8s-transport-settings.html#k8s-transport-ca

The user provided CA and the private key needs to be exposed to the ECK operator to issue certificates. It will create a Kubernetes secret to contain all certificates and private keys for each of the nodes in the Elasticsearch cluster. 

Pre-generating the certificates is impractical because the number of Elasticsearch nodes may grow and shrink as the cluster scales. The operator can of course deal with this by updating the certificate secret as needed. But a user would not.

To allow an even tighter integration with existing certificate infrastructure and decoupling of the certificate generation from the ECK operator we should investigate if we can support https://cert-manager.io/docs/projects/csi-driver/
This would allow us to mount node specific certificates without sharing keys. 

Trade-offs:
* user experience: user has to setup and configure cert-manager and cert-manager CSI driver
* we would no longer be able to use the otherName SAN required for trust restrictions in ESS (thorough testing of CCS/CCR with ESS would be part of this work) https://github.com/elastic/elasticsearch/pull/91946 relaxes the requirement on newer versions of ES somewhat by allowing SAN/dnsName as well. 
* where do the users responsibilities end: does the user create the issuer and reference it in the ES custom resource? Are we going to build an explicit (optional) dependency on cert-manager into ECK 
* cert-manager performance at scale is problematic (not an issue for majority of users)

Plan: 
1. Introduce `transport.tls.certificateAuthorities.configMapName` and `http.tls.certificateAuthorities.configMapName` (for symmetry) to allow users to specify additionally trusted CAs (enables tools like trust-manager, and allows ECK to uses those in remote cluster associations)
2. Remove the warning when users set `xpack.security.transport.ssl.key`, `xpack.security.transport.ssl.certificate` (to discuss: should this be only be allowed when they set `transport.tls.certificate.trust` ?) 
3. Add a page to our documentation on transport based on the example in https://github.com/elastic/cloud-on-k8s/issues/6479#issuecomment-1475337819 plus the `certificateAuthorities` attribute.

pebrc:
Fixes #6479 

Introduce a new attribute `transport.tls.certificateAuthorities.configMapName` which expects a config map with a `ca.crt` file that can contain one or more CAs in PEM format. Alternative name suggestion `transport.tls.trust`

Concatenate these additional CAs with the CA issued by ECK in the internal and public secrets. This ensures that all existing mechanisms including remote cluster support seamlessly pick up the additional trust. It also allows transitioning without downtime from self-issued certificates to third-party issued certificates. 

Adds a documentation section to illustrate how transport certificates can be configured with cert-manager csi-driver (arguably could also be a recipe but not one we auto-test because of the additional infrastructure requirements) 

Despite declaring my intention to add a similar attribute to the HTTP options I did not go through with it. I don't see the same need given that the HTTP certificate is effectively a server certificate and we have an existing mechanism through associations to establish trust. Also HTTP options are shared across all stack applications and thus this change has a much larger scope. We can follow up in a separate PR if reviewers think it useful.


docs/orchestrating-elastic-stack-applications/elasticsearch/remote-clusters.asciidoc
```diff
@@ -95,23 +95,13 @@ kind: Elasticsearch
 metadata:
   name: cluster-two
 spec:
+  transport:
+    tls:
+      certificateAuthorities:
+        configMapName: remote-certs
```
pebrc, line 101: I just realised that the example needs more adjustment will update in a little bit.

pebrc:
buildkite test this -f p=gke,t=TestCustomTransportCA -m s=8.7.1,s=8.8.0-SNAPSHOT^BUILD_LICENSE_PUBKEY=dev

pebrc:
Test failed due to Google outage

pebrc:
buildkite test this -f p=gke,t=TestCustomTransportCA -m s=8.8.0-SNAPSHOT^BUILD_LICENSE_PUBKEY=dev

pkg/controller/elasticsearch/certificates/transport/trust.go
```diff
@@ -0,0 +1,36 @@
...
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+
+	v1 "github.com/elastic/cloud-on-k8s/v2/pkg/apis/elasticsearch/v1"
+	"github.com/elastic/cloud-on-k8s/v2/pkg/controller/common/certificates"
+	"github.com/elastic/cloud-on-k8s/v2/pkg/utils/k8s"
+)
+
+func MaybeRetrieveAdditionalCAs(ctx context.Context, client k8s.Client, elasticsearch v1.Elasticsearch) ([]byte, error) {
```
barkbay, line 19: Do we need to watch the `configMap`?

pebrc: Oh yes! I changed the implementation. Originally I had the config map mounted directly into the Pod and thought Elasticsearch would take care of watching the contents changing. But now that I am concatenating it is actually necessary to watch it.

pkg/apis/elasticsearch/v1/elasticsearch_types.go
```diff
@@ -172,6 +172,9 @@ type TransportTLSOptions struct {
 	// - `ca.crt`: The CA certificate in PEM format.
 	// - `ca.key`: The private key for the CA certificate in PEM format.
 	Certificate commonv1.SecretRef `json:"certificate,omitempty"`
+	// CertificateAuthorities is a references to a config map that contains one or more x509 certificates for
+	// trusted authorities in PEM format. The certificates need to be in a file called `ca.crt`.
```
barkbay, line 176: I wonder if there are cases where the name of the key is not under the control of the user. For example the CA key name from the [OpenShift CA service](https://docs.openshift.com/container-platform/4.12/security/certificate_types_descriptions/service-ca-certificates.html#purpose) is `service-ca.crt`:

> Other services can request that the CA bundle for the service CA be injected into API service or config map resources by annotating with service.beta.openshift.io/inject-cabundle: true to support validating certificates generated from the service CA. In response, the Operator writes its current CA bundle to the CABundle field of an API service or **as service-ca.crt to a config map**

I don't think however that the OpenShift CA service can be used to manage the transport certificates, so it's not a valid argument. And we can offer a first implementation where the expected key is `ca.crt`, and later add a new field to select a different key:

```yaml
  transport:
    tls:
      certificateAuthorities:
        configMapName: trust
        keys: ## Optional, ca.crt by default
          - service-ca.crt
```

docs/orchestrating-elastic-stack-applications/elasticsearch/transport-settings.asciidoc
```diff
@@ -68,3 +68,93 @@ spec:
...
     count: 3
 ----
+
+== Issue node transport certificates with third-party tools
+
+When following the instructions in <<{p}-transport-ca>> the issuance of certificates is orchestrated by the ECK operator and the operator needs access to the CAs private key.
+If this is undesirable it is also possible to configure node transport certificates without involving the ECK operator. The following two pre-requisites apply:
+
+1. the tooling used must be able to issue individual certificates for each Elasticsearch node and dynamically add or remove certificates as the cluster scales up and down.
+2. the ECK operator must be configured to trust the CA in use.
```
barkbay, line 78:
```suggestion
2. The Elasticsearch nodes must be configured to trust the CA in use.
```

pkg/apis/elasticsearch/v1/elasticsearch_types.go
```diff
@@ -172,6 +172,9 @@ type TransportTLSOptions struct {
 	// - `ca.crt`: The CA certificate in PEM format.
 	// - `ca.key`: The private key for the CA certificate in PEM format.
 	Certificate commonv1.SecretRef `json:"certificate,omitempty"`
+	// CertificateAuthorities is a references to a config map that contains one or more x509 certificates for
```
thbkrkr, line 175:
```suggestion
	// CertificateAuthorities is a reference to a config map that contains one or more x509 certificates for
```

barkbay (approved):
LGTM! (at least for the non-docs part, I've never been very comfortable reviewing documentation 🙃 )
I have done some tests with the `cert-manager` and it seems to work as expected 👍

barkbay:
@pebrc I'll cut the 2.8 branch in a bit, let me know if you want to get this one into the next release 🙇

pebrc:
Sorry. I will merge this now. I could have sworn I had already hit the button a few days ago ...

PR #6731, opened 2023-04-26
Fix the sample tests for logstash service and stack monitoring

robbavey:
The stack monitoring tests were failing as the builder was not creating the monitoring elasticsearch references. This was added, as well as a minor refactoring to split the builder into adding metrics and logging monitoring separately.

This commit also calculates the the monitoring assocation counts correctly, as previously the metrics association was double counted.

This commit also fixes the logstash service sample end to end test - the pipeline id in the yaml definition did not match that in the test.

### DIFF
--- a/test/e2e/test/logstash/checks.go
+++ b/test/e2e/test/logstash/checks.go
@@ -143,17 +143,13 @@ func CheckStatus(b Builder, k *test.K8sClient) test.Step {
 	}
 }
 
-func uniqueAssociationCount(refs, otherRefs []v1.ObjectSelector) int {
-	uniqueAssociations := make(map[v1.ObjectSelector]bool)
-
-	for _, val := range refs {
-		uniqueAssociations[val] = true
+func uniqueAssociationCount(refsList ...[]v1.ObjectSelector) int {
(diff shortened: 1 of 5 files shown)


test/e2e/test/logstash/checks.go
```diff
@@ -143,6 +143,20 @@ func CheckStatus(b Builder, k *test.K8sClient) test.Step {
...
+	for _, val := range refs {
+		uniqueAssociations[val] = true
+	}
+
+	for _, val := range otherRefs {
+		uniqueAssociations[val] = true
+	}
+
+	return len(uniqueAssociations)
+}
```
barkbay, line 158: nit: using a `struct{}` instead of a `bool` is more optimal because the former doesn't use any additional memory (I do however agree that it doesn't make much difference in this case since we are dealing with a very small set 😄 )

```suggestion
func uniqueAssociationCount(refsList ...[]v1.ObjectSelector) int {
	uniqueAssociations := make(map[v1.ObjectSelector]struct{})
	for _, refs := range refsList {
		for _, val := range refs {
			uniqueAssociations[val] = struct{}{}
		}
	}
	return len(uniqueAssociations)
}
```

barkbay (approved):
LGTM!

PR #6737, opened 2023-04-27
Support global container repository

### LINKED ISSUE #6643: Allow users to specify custom repo when specifying `CONTAINER_REPOSITORY` for Elastic Operator
# Use Case

As my company's Elastic Administrator, I need to be able to specify a custom respository within my local `CONTAINER_REGISTRY`.

# Details

**Elastic Operator version:** 2.6.1

I am required to use my company's local registry to pull images from in our environment. Because of conventions, I am also required to use a specific naming schema for images within the registry:

```
registry.example.com/team/elastic-agent:8.7.0
```

When deploying the Elastic Operator, I am able to specify a default container registry, e.g.:

```
CONTAINER_REGISTRY=registry.example.com/team
```

The issue is that Elastic Operator appears to have the image naming schema hardcoded. E.g., when I deploy Elastic Agent with my custom registry and repository, Elastic Operator attempts to pull:

```
registry.example.com/team/beats/elastic-agent:8.7.0
```

As a workaround, I can specify the image in my deployment:

```
image: 'registry.example.com/team/elastic-agent:8.7.0
```

However, this partially defeats the purpose of using the Elastic Operator, which is to allow Elastic Operator to manage the image source.

# Solution

Either require the `CONTAINER_REGISTRY` to contain both registry and repository (`registry.example.com/team` in this example) **or** add another configuration option `CONTAINER_REPOSITORY`, which would default to the current value (`beats` in this example), which would mirror current behavior when pulling an image from the default registry (`docker.elastic.co`).

Since requiring `CONTAINER_REGISTRY` to contain the full registry and repository name would likely be a breaking change for users who pull their images directly from Elastic, the latter option would be better.

thbkrkr:
This adds a new flag `--container-repository` to the operator to be able to specify a global container repository.

With this it is now possible to use DockerHub images for example:

`--container-registry docker.io --container-repository=elastic`

Resolves #6643.

### DIFF
--- a/docs/operating-eck/air-gapped.asciidoc
+++ b/docs/operating-eck/air-gapped.asciidoc
@@ -39,13 +39,24 @@ To make use of your mirrored images you can either set the image for each applic
 
 When creating custom resources ({eck_resources_list}), the operator defaults to using container images pulled from the `docker.elastic.co` registry. If you are in an environment where external network access is restricted, you can configure the operator to use a different default container registry by starting the operator with the `--container-registry` command-line flag. Check <<{p}-operator-config>> for more information on how to configure the operator using command-line flags and environment variables.
 
-The operator expects container images to be located at specific paths in the default container registry. Make sure that your container images are stored at the right path and are tagged correctly with the stack version number. For example, if your private registry is `my.registry` and you wish to deploy components from stack version {version}, the following image paths should exist:
-
+The operator expects container images to be located at specific repositories in the default container registry. Make sure that your container images are stored at the right repositories and are tagged correctly with the Stack version number. For example, if your private registry is `my.registry` and you wish to deploy components from Stack version {version}, the following image names should exist:
 
 * +my.registry/elasticsearch/elasticsearch:{version}+
 * +my.registry/kibana/kibana:{version}+
 * +my.registry/apm/apm-server:{version}+
 
+[float]
+[id="{p}-container-registry-override"]
+== Override the default container repositories
+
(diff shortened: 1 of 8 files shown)


docs/operating-eck/air-gapped.asciidoc
```diff
@@ -39,13 +39,24 @@ To make use of your mirrored images you can either set the image for each applic
...
 
 * +my.registry/elasticsearch/elasticsearch:{version}+
 * +my.registry/kibana/kibana:{version}+
 * +my.registry/apm/apm-server:{version}+
 
+[float]
+[id="{p}-container-repository-override"]
+== Use a global container repository
+
+If you cannot follow the default Elastic image repositories, you can configure the operator to use a different container repository by starting the operator with the `--container-repository` command-line flag.
```
pebrc, line 52:
```suggestion
If you cannot follow the default Elastic image repositories naming scheme, you can configure the operator to use a different container repository by starting the operator with the `--container-repository` command-line flag.
```

pkg/controller/common/container/container_test.go
```diff
@@ -50,6 +51,21 @@ func TestImageRepository(t *testing.T) {
...
+			version:    "42.0.0",
+			repository: "elastic",
+			want:       testRegistry + "/elastic/elasticsearch:42.0.0",
+		},
+		{
+			name:       "Elasticsearch image with custom repository and suffix",
+			image:      ElasticsearchImage,
+			version:    "42.0.0",
+			repository: "elastic",
+			suffix:     "-obi1",
```
pebrc, line 66: 🕶️

pebrc (approved):
LGTM!

PR #6748, opened 2023-05-02
Update K8S versions

barkbay:
This PR updates various versions of K8S in our documentation and in our pipelines.


barkbay:
I did not update `kubeVersion` in `deploy/eck-*/Chart.yaml`. It is `>= 1.21.0-0` right now, while we state that we only support `1.22`. I don't know if this is an oversight or if we have a different policy.

pebrc:
> I did not update kubeVersion in deploy/eck-*/Chart.yaml. It is >= 1.21.0-0 right now, while we state that we only support 1.22. I don't know if this is an oversight or if we have a different policy.

The last time we [discussed this](https://github.com/elastic/cloud-on-k8s/pull/6593#issuecomment-1485125744) we decided to keep the `kubeVersion` at the lowest version still tested in CI even if it is lower than the officially supported version. The rationale was that we did not want to enforce the support policy as a hard technical limit that could cause friction for customers as long as there was not strong technical need to do that.

barkbay:
buildkite test this -f p=kind,t=TestSmoke

naemono (approved):
lgtm

Issue #6740, opened 2023-04-28
runAsNonRoot breaks stack monitoring on older versions of the stack


pebrc:
```
  state:                                                                                                                                                                                                                                                                                                               │
│       waiting:                                                                                                                                                                                                                                                                                                           │
│         message: 'container has runAsNonRoot and image has non-numeric user (metricbeat),                                                                                                                                                                                                                                │
│           cannot verify user is non-root (pod: "elasticsearch-es-mixed-2_beats-3064(51b07a17-26c4-4301-9380-ea6fe3e75aca)",    
```

I think the side car containers have the same issue as the main Elasticsearch container. I missed this when testing the original PR https://github.com/elastic/cloud-on-k8s/pull/6703

barkbay:
Good catch! I think I missed this one because we apply a default `PodSecurityContext` in our e2e tests: https://github.com/elastic/cloud-on-k8s/blob/e679b5f4beb19bc4e983b5a6834d75c20dc4fdb0/test/e2e/test/default.go#L14-L33

@naemono I did not realised that you self assigned this one (the page was not refreshed in my browser when I clicked the `assign yourself` link). FF is today, I don't think we have to wait for https://github.com/elastic/beats/pull/35272. I think it's also acceptable to let users manually set "RunAsNonRoot" when stack monitoring is enabled.

barkbay:
I think that we have an issue with any stack version prior to `8.0.0` as soon as there is no `PodSecurityContext` 🤦 

@naemono could take a look? To reproduce you can deploy the elasticsearch manifest in the `samples` directory using version `7.17.8`:

```
> k get es,pods
NAME                                                              HEALTH    NODES   VERSION   PHASE             AGE
elasticsearch.elasticsearch.k8s.elastic.co/elasticsearch-sample   unknown           7.17.8    ApplyingChanges   4m9s

NAME                                    READY   STATUS                  RESTARTS      AGE
pod/elasticsearch-sample-es-default-0   0/1     Init:CrashLoopBackOff   5 (41s ago)   4m6s
pod/elasticsearch-sample-es-default-1   0/1     Init:CrashLoopBackOff   5 (68s ago)   4m6s
pod/elasticsearch-sample-es-default-2   0/1     Init:CrashLoopBackOff   5 (50s ago)   4m6s
```

barkbay:
[...]

PR #6754, opened 2023-05-03
Remove `runAsNonRoot` for Beats.

naemono:
related #6740 

Until https://github.com/elastic/beats/pull/35272 is merged, we can't enable `runAsNonRoot` for Beats.


barkbay (commented):
I think we also need to update https://github.com/elastic/cloud-on-k8s/blob/a9a9b24c7091ee3e952109bf1684e2af1eeba840/test/e2e/test/elasticsearch/check_securitycontext.go#L52

naemono:
> test/elasticsearch/check_securitycontext.go

Thanks, had missed that this was also tested here.

barkbay (approved):
LGTM

PR #6755, opened 2023-05-03
Do not set capabilities for Elasticsearch < 8.x

naemono:
Elasticsearch v7 does not work well when `securityContext.Capabilities` is set to `Drop: "all"`. Both the initContainer that prepares the filesystem fails when attempting to preserve ownership, and the main Elasticsearch container fails with `chroot: cannot change root directory to '/': Operation not permitted` as the containers likely run as the `root` user pre v8.

### DIFF
--- a/pkg/controller/elasticsearch/securitycontext/securitycontext.go
+++ b/pkg/controller/elasticsearch/securitycontext/securitycontext.go
@@ -12,22 +12,30 @@ import (
 )
 
 var (
-	// MinStackVersion is the minimum Stack version to use RunAsNonRoot with the Elasticsearch image.
+	// RunAsNonRootMinStackVersion is the minimum Stack version to use RunAsNonRoot with the Elasticsearch image.
 	// Before 8.8.0 Elasticsearch image runs has non-numeric user.
 	// Refer to https://github.com/elastic/elasticsearch/pull/95390 for more information.
-	MinStackVersion = version.MustParse("8.8.0-SNAPSHOT")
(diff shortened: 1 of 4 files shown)


pkg/controller/elasticsearch/securitycontext/securitycontext.go
```diff
@@ -12,22 +12,30 @@ import (
...
-		Capabilities: &corev1.Capabilities{
-			Drop: []corev1.Capability{"ALL"},
-		},
 		Privileged:               ptr.Bool(false),
 		ReadOnlyRootFilesystem:   ptr.Bool(enableReadOnlyRootFilesystem),
 		AllowPrivilegeEscalation: ptr.Bool(false),
 	}
-	if ver.LT(MinStackVersion) {
+	if ver.LT(DropCapabilitiesMinStackVersion) {
+		return sc
```
barkbay, line 33: Maybe add a unit test to check that `Capabilities` are not set if < 8.0.0

naemono:
buildkite test this -f p=gke -m s=8.7.0,s=7.17.10

barkbay (approved):
LGTM

pkg/controller/elasticsearch/securitycontext/securitycontext.go
naemono: > Maybe add a unit test to check that `Capabilities` are not set if < 8.0.0

Will do. Gotta fix some other tests as well...

PR #6769, opened 2023-05-08
Fix incorrect auth secret_token key post 8.0

### LINKED ISSUE #6768: ApmServer secret token config is invalid for APM Server 8.0+
https://github.com/elastic/cloud-on-k8s/blob/f82671958ae2f529b1cc626112d0bddaf8b41205/pkg/controller/apmserver/config.go#L30

ECK is setting outdated configuration for the APM secret token, that was removed in APM Server 8.0 (https://www.elastic.co/guide/en/apm/guide/current/release-notes-8.0.html#_breaking_changes_7). This should be updated to `apm-server.auth.secret_token` for 8.0+

The outcome of this bug is that, by default, APM Server has no secret token auth configured.

pebrc:
Fixes #6768 

To discuss: whether we should change the e2e tests to check the case where a client tries to use an incorrect auth token to make a requests (which should be rejected). This would also IUC test for correct configuration of the APM server with a auth token. Because we did not do this check we missed the misconfiguration since 8.0 in our e2e testing. 

Marking this as breaking because we shipped versions of ECK that misconfigured 8.0+ APM servers and therefore there might be user installations out there that work only because of this misconfiguration. If we now ship the fix and their clients are not actually setting an auth token then APM server will start rejecting their requests.

### DIFF
--- a/pkg/controller/apmserver/config.go
+++ b/pkg/controller/apmserver/config.go
@@ -19,6 +19,7 @@ import (
 	"github.com/elastic/cloud-on-k8s/v2/pkg/controller/common/certificates"
 	"github.com/elastic/cloud-on-k8s/v2/pkg/controller/common/reconciler"
 	"github.com/elastic/cloud-on-k8s/v2/pkg/controller/common/settings"
+	"github.com/elastic/cloud-on-k8s/v2/pkg/controller/common/version"
 	"github.com/elastic/cloud-on-k8s/v2/pkg/utils/k8s"
 )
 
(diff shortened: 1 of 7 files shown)


pebrc:
buildkite test this -f p=gke,E2E_TAGS=apm  -m s=8.7.0,s=7.17.10

pebrc:
buildkite test this -f p=gke,E2E_TAGS=apm -m s=8.7.0,s=7.17.10

pebrc:
I think there is a problem with the recent changes to `securityContexts` not dropping capabilities on older versions of Elasticsearch and the e2e test checks not accounting for that. Will need to figure out why this is not being flagged up in the nightlies (maybe it is but burrowed under other failures)

barkbay:
```
{"Time":"2023-05-08T19:20:13.711514403Z","Action":"output","Package":"github.com/elastic/cloud-on-k8s/v2/test/e2e/apm","Test":"TestCrossNSAssociation/Elasticsearch_containers_SecurityContext_should_be_set","Output":"    --- FAIL: TestCrossNSAssociation/Elasticsearch_containers_SecurityContext_should_be_set (0.00s)\n"}
```

Should be fixed by https://github.com/elastic/cloud-on-k8s/pull/6772

pebrc:
buildkite test this -f p=gke,E2E_TAGS=apm -m s=8.7.0,s=7.17.10

barkbay (approved):
LGTM

pebrc:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|2.8|[](https://github.com/elastic/cloud-on-k8s/pull/6778)|

### Questions ?

PR #6775, opened 2023-05-09
Run full e2e tests matrix on any tag

thbkrkr:
This commit changes the test strategy and runs the full e2e tests matrix on any tag.


barkbay (approved):
Thanks! 🙇

PR #6782, opened 2023-05-10
Test max message size of a slack notification

thbkrkr:
Just an experiment to check the max message size of a slack notification from Buildkite.


thbkrkr:
✅ 

Capture d’écran 2023-05-11 à 22 06 32

Issue #6765, opened 2023-05-05
[Docs] Upgrading from to/through 2.6 causes a rolling restart of Elasticsearch cluster.


BenB196:
## Bug Report
**What did you do?**

Upgrade from 2.5 to 2.6 (or to 2.7 directly)

**What did you expect to see?**

Per the [docs](https://www.elastic.co/guide/en/cloud-on-k8s/2.7/k8s-upgrading-eck.html#k8s-beta-to-ga-rolling-restart) there shouldn't be a rolling restart for 2.6 or 2.7.

**What did you see instead? Under which circumstances?**

ECK Operator managed cluster does a rolling restart.

**Environment**

* ECK version:

	2.5 -> 2.6/2.7

* Kubernetes information:

	insert any information about your Kubernetes environment that could help us:
	* On premise N/A
	* Cloud: GKE / EKS / AKS N/A
	* Kubernetes distribution: Openshift / Rancher / PKS N?A
	
The documents should be updated to include 2.6 in the list of versions that will cause a rolling restart of a cluster.

PR #6786, opened 2023-05-10
[DOC] Add 2.6 and 2.7 to the triggered restart list

barkbay:
Fixes https://github.com/elastic/cloud-on-k8s/issues/6765

### DIFF
--- a/docs/operating-eck/upgrading-eck.asciidoc
+++ b/docs/operating-eck/upgrading-eck.asciidoc
@@ -97,7 +97,7 @@ This will update the ECK installation to the latest binary and update the CRDs a
 
 Upgrading the operator results in a one-time update to existing managed resources in the cluster. This potentially triggers a rolling restart of pods by Kubernetes to apply those changes. The following list contains the ECK operator versions that would cause a rolling restart after they have been installed.
 
- 1.6, 1.9, 2.0, 2.1, 2.2, 2.4, 2.5, 2.8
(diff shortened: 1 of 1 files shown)


pebrc (approved):
Surprised that we missed this.

thbkrkr:
Same, I swear I checked that there was no restart when updating from `2.6` to `2.7` with a very simple stack (quickstart). Maybe it happens with a more complex stack.

BenB196:
> Same, I swear I checked that there was no restart when updating from `2.6` to `2.7` with a very simple stack (quickstart). Maybe it happens with a more complex stack.

FYI, my issue might've been confusing, I've upgraded a few ~10 ECK operators, and didn't see an restart from 2.6 -> 2.7, but did see a restart for 2.5 -> 2.6 and 2.5 -> 2.7, so might only be 2.6 that causes a restart.

(Weird side note, 2.7 does cause some beats/agents to be restarted, but I don't think that is a documented thing anywhere)

barkbay:
> didn't see a restart from 2.6 -> 2.7,

I did a quick test before opening the PR. While Elasticsearch Pods were not restarted in that context, it was not the case for the Kibana ones which were actually restarted.

> Weird side note, 2.7 does cause some beats/agents to be restarted, but I don't think that is a documented thing anywhere

Same for Kibana, that's why I included 2.7 in that list, which I did not consider as restricted to Elasticsearch.

PR #6820, opened 2023-05-17
Use text/template instead of html/template

thbkrkr:
This replaces the use of `text/template` with `html/template` which has a lower attack surface (less CVE).


naemono (approved):
See no reason why we'd need html/template vs text/template. 👍

PR #6822, opened 2023-05-17
[buildkite] Limit notified failures to not exceed message longer than 3000 bytes

thbkrkr:
This prevents to send a Slack notification with a message greater than 3000 chars by limiting the number of notified test failures to 15 and adding the text `Go to build to see all xyz failures.` at the end when there are more.

To make this workable from the template, the map of list of short failures by env is flattened and the total count of short failures is injected in the template.

Not related to this, this also renames trimError in truncateError for consistency.

### DIFF
--- a/.buildkite/e2e/reporter/main.go
+++ b/.buildkite/e2e/reporter/main.go
@@ -228,17 +228,19 @@ func truncateNotifyMessage(yamlDef []byte, length int) ([]byte, error) {
 	}
 	err := yaml.Unmarshal(yamlDef, &pipeline)
 	if err != nil {
-		return nil, nil
+		return nil, err
 	}
 	for i, s := range pipeline.Steps {
 		for j, n := range s.Notify {
-			// truncate message and replace last 3 chars by '...'
-			pipeline.Steps[i].Notify[j].Slack.Message = n.Slack.Message[0:length-1-3] + "..."
+			if len(n.Slack.Message) >= length {
+				// truncate message and replace last 3 chars by '...'
+				pipeline.Steps[i].Notify[j].Slack.Message = n.Slack.Message[0:length-1-3] + "..."
+			}
 		}
 	}
 	yamlDefBytes, err := yaml.Marshal(pipeline)
 	if err != nil {
-		return nil, nil
+		return nil, err
 	}
 	return yamlDefBytes, nil
 }
--- a/.buildkite/e2e/reporter/main.go
+++ b/.buildkite/e2e/reporter/main.go
@@ -14,7 +14,6 @@ import (
 	"text/template"
 
 	"github.com/joshdk/go-junit"
-	"gopkg.in/yaml.v3"
 )
 
 const (
@@ -106,8 +105,8 @@ func main() {
 	if err != nil {
 		exitWith(err)
 	}
-	var output bytes.Buffer
-	err = tpl.Execute(&output, map[string]interface{}{
+	var outBytes bytes.Buffer
+	err = tpl.Execute(&outBytes, map[string]interface{}{
 		"FailuresCount":      failuresCount,
 		"ShortFailuresCount": shortFailuresCount,
 		"Tests":              tests,
@@ -116,12 +115,14 @@ func main() {
 		exitWith(err)
 	}
 
-	outBytes, err := truncateNotifyMessage(output.Bytes(), maxSlackMessageSizeBytes)
-	if err != nil {
-		exitWith(err)
+	out := outBytes.String()
+
+	// truncate notify slack message if it exceeds the max size
+	if outputFormat == notifyFailures && len(out) > maxSlackMessageSizeBytes {
+		out = out[0:maxSlackMessageSizeBytes]
 	}
 
-	fmt.Print(string(outBytes))
+	fmt.Println(out)
 
 	if failuresCount > 0 {
 		os.Exit(1)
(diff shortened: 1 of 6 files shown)


.buildkite/e2e/reporter/main.go
```diff
@@ -195,14 +208,41 @@ func sortByName(tests []junit.Test) {
...
+		} `yaml:"steps"`
+	}
+	err := yaml.Unmarshal(yamlDef, &pipeline)
+	if err != nil {
+		return nil, nil
+	}
+	for i, s := range pipeline.Steps {
+		for j, n := range s.Notify {
+			// truncate message and replace last 3 chars by '...'
+			pipeline.Steps[i].Notify[j].Slack.Message = n.Slack.Message[0:length-1-3] + "..."
```
naemono, line 236: Isn't this going to fail with `panic: runtime error: slice bounds out of range [:3000] with capacity x` when the message isn't at least of length 3000? A quick test in go playground seems to verify this.

naemono (changes requested):

thbkrkr:
I'm not sure about the best solution to limit the size of the message.

I thought we could pre compute the message before the templating or inject a function in the template that takes all the tests and is enough smart to stop writing when the size exceeds:

```diff
-       tpl, err := template.New("report").Parse(srcTpl)
+       tpl, err := template.New("report").Funcs(template.FuncMap{
+               "notifyMessage": func(t map[string]sortedTests) string {
+                       res := ""
+                       for envName, tests := range t {
+                               for _, test := range tests.ShortFailed {
+                                       failure := fmt.Sprintf("             🐞 `%s` %s\n", envName, test.Name)
+                                       if len(res)+len(failure) > maxSlackMessageSizeBytes {
+                                               break
+                                       }
+                                       res += failure
+                               }
+                       }
+                       return res
+               },
+       }).Parse(srcTpl)
```

I saw a problem with this approach. You need to know how many spaces to add before the 🐞 to correctly respect the indentation. Edit: in fact this is not a real issue when you stop writing a multiline string.

I tried another approach (the current proposal) which is to post-process the message after templating. The downside is that I put part of the schema used to define the notification pipeline in the go code to parse the message from the final rendered template.

.buildkite/e2e/reporter/main.go
```diff
@@ -195,14 +208,41 @@ func sortByName(tests []junit.Test) {
...
+			Notify []struct {
+				Slack struct {
+					Message string `yaml:"message"`
+				} `yaml:"slack"`
+			} `yaml:"notify"`
+		} `yaml:"steps"`
+	}
+	err := yaml.Unmarshal(yamlDef, &pipeline)
+	if err != nil {
+		return nil, nil
```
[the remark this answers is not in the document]
thbkrkr: Yes! It was late...

.buildkite/e2e/reporter/main.go
thbkrkr: You are right, it needs a length check before truncation.

naemono (approved):
Seems good now

.buildkite/e2e/reporter/main.go
```diff
@@ -195,14 +208,43 @@ func sortByName(tests []junit.Test) {
...
+func truncateNotifyMessage(yamlDef []byte, length int) ([]byte, error) {
+	var pipeline struct {
+		Steps []struct {
+			Notify []struct {
+				Slack struct {
+					Message string `yaml:"message"`
+				} `yaml:"slack"`
+			} `yaml:"notify"`
+		} `yaml:"steps"`
+	}
```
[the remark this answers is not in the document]
thbkrkr: This doesn't work well because it removes all fields that are not declared in the struct.

thbkrkr:
It doesn't work with yaml parsing without fully defining the whole structure. As a simple solution, I propose to just truncate the entire rendered template without optimization, so accepting to lose 191 chars corresponding to information other than the message.

```yaml
# The first "191 chars":
steps:
  - label: ""
    if: build.branch == "main"
    command: echo "107 failures to notify"; exit 1
    notify:
      - slack:
          channels:
            - "#eck"
          message: |
```

thbkrkr:
We discussed and concluded that a better approach would be to simply limit the number of reported tests and add a text "See more ..." if there are more than the limit. The limit will be pretty low as it is a Slack notification and the goal is not to be exhaustive, and also to be sure to not exceed the total message size of 3000 bytes.

One potential problem could come from that we don't control the size of the test name like we control the size of the env name ([getSlugName](https://github.com/elastic/cloud-on-k8s/blob/c18ece5ca165d522874075da3404f72dad1aa507/.buildkite/e2e/pipeline-gen/main.go#L337-L340)).

To stay under 3000 bytes, with a limit of 15 reported tests, a 'See more...' message of 40 bytes, this gives 197 bytes per line ("🐞 `TestName` env-name") where env-name cannot exceed 16 chars and "🐞 `` " takes 12 chars. This leaves 169 chars for the test name ( (3000-40)/15-12-16=169 ) ; the current longest name is 55 chars (TestEnterpriseSearchTLSDisabledVersionUpgradeToLatest8x) ; so we should be safe.

.buildkite/e2e/reporter/templates/notify-failures.tpl.yml
```diff
@@ -1,18 +1,23 @@
...
-            🐞 `{{ $test.Name }}` {{ $envName }}
+          {{- range $i, $test := .ShortFailures }}
+            🐞 `{{ $test.Name }}` {{ $test.EnvName }}
+            {{- if gt $i $.MaxNotifiedShortFailures }}
+              {{- break }}
             {{- end }}
           {{- end }}
+          {{- if gt .ShortFailuresCount $.MaxNotifiedShortFailures }}
+            Go to build to see all {{ .ShortFailuresCount }} failures.
+          {{- end }}
```
naemono, line 21: I think this makes a lot of sense here, as it will keep slack messages short as they should be. 👍

naemono (approved):
lgtm

Issue #6561, opened 2023-03-22
Make the deployer configurable via environment variables


thbkrkr:
Today, the deployer is configured via a YAML file.

Example:
```yaml
id: gke-ci
overrides:
  vaultInfo:
    address: https://vault.xyz:1234
    rootPath: secret/ci/abcd
  operation: create
  clusterName: eck-e2e-gke-abcd-0
  kubernetesVersion: 1.22
  gke:
    region: europe-west42
```

It is not convenient to use for CI where it's simpler to use environment variables. 

We had to write an intermediate bash script to write the file from environment variables.

https://github.com/elastic/cloud-on-k8s/blob/fcbc995b75a5f99ca7861c195d3b88b4a2b71065/.ci/setenvconfig#L348-L359

In #6525, I moved this into a new script to remove duplicates and isolate the deployer config from the config to build images and the test runner config. Before that, all configs were mixed in `.ci/setenvconfig`.

It would be better if the deployer could be configured directly via environment variables to avoid this bash script which adds an unnecessary extra layer.

I thought of something generic that would give a way to configure any configuration setting:

```sh
DEPLOYER_PLAN_ID=gke-ci
DEPLOYER_vaultInfo_address=https://vault.xyz:1234
DEPLOYER_vaultInfo_rootPath=ecret/ci/abcd
DEPLOYER_operation=create
DEPLOYER_clusterName=eck-e2e-gke-abcd-0
DEPLOYER_kubernetesVersion=1.22
DEPLOYER_gke_region=europe-west42
```

PR #6827, opened 2023-05-18
E2E Test Cluster Cleanup step

naemono:
## Proposed Change
~Adds a step after our nightly e2e tests that will always run, and cleanup any dangling clusters that are >= 3 days old.~
Adds a dedicated pipeline that we can run to cleanup dangling e2e test clusters that are >= 3 days old.

### Limitations
* ~Only looks at GKE clusters at this time.~
* ~Doesn't handle OCP and Tanzu at this time~

### Tests run
* Cleanup of existing GKE cluster
* Cleanup of existing AKS cluster
* Cleanup of existing EKS cluster
* Cleanup of existing Tanzu cluster
* Cleanup of OCP cluster

### TODO
- [x] GKE
- [x] AKS
- [x] EKS
- [x] OCP
- [x] Tanzu


.buildkite/scripts/test/cleanup-e2e-clusters.sh
```diff
@@ -0,0 +1,27 @@
...
+gcloud auth activate-service-account --key-file=/tmp/ci-gcp-k8s-operator.json
+rm -f /tmp/ci-gcp-k8s-operator.json
+gcloud config set project elastic-cloud-dev
+
+# Get a list of cluster names with a `createTime` < 3 days ago.
+CLUSTERS=$(gcloud container clusters list --region=europe-west6 --format="value(name)" --filter="createTime<${DATE} AND name~eck-e2e.*")
+
+for i in ${CLUSTERS} ; do
+    echo "Deleting cluster $i"
+    gcloud container clusters delete "${i}" --region=europe-west6
```
barkbay, line 26: Any reason not to use the deployer? It feels like we are duplicating the deletion logic, for example any improvements to the cleanup process in the deployer would be lost. This may also be needed for more advanced deployments like OpenShift.

.buildkite/pipeline-e2e-tests.yml
```diff
@@ -62,3 +63,17 @@ steps:
         command: |
           cd .buildkite/e2e/pipeline-gen && go build -o pipeline-gen
           cat ../nightly-main-matrix.yaml | ./pipeline-gen | buildkite-agent pipeline upload
+
+      # After nightly e2e tests, run a cleanup scripts to ensure dangling clusters are deleted
+      - label: ":broom: e2e test cluster cleanup"
```
barkbay, line 68: Would it be possible to have a dedicated pipeline? I was thinking of something that could be developed and tested as stand-alone, and triggered manually.

.buildkite/scripts/test/cleanup-e2e-clusters.sh
naemono: Yeah, I didn't think to run the deployer's delete functionality tbh. I'll look back at this and see what it would take to just use it. 💯

naemono: This is now using the deployer to delete GKE clusters @barkbay

hack/deployer/cmd/cleanup.go
```diff
@@ -0,0 +1,80 @@
...
+		if err != nil {
+			return err
+		}
+		clusters, err := client.Cleanup()
+		if err != nil {
+			return err
+		}
+		for _, cluster := range clusters {
+			log.Printf("deleted cluster: %s", cluster)
+		}
```
thbkrkr, line 77: Not sure about the interest of this log.More generally, I think we could lighten up what we log a bit because it's quite verbose.

thbkrkr: I commented the deletion to test. With the deletion on, there is a third log `deleting cluster eck-e2e-stack-8-5-3-cgno-4899`.
```sh
> go run main.go cleanup --provider gke
2023/07/20 13:15:20 Authenticating as service account...
Updated property [core/project].
Activated service account credentials for: [k8s-operator-ci@elastic-cloud-dev.iam.gserviceaccount.com]
eck-e2e-gke-all-gmvc-4912
eck-e2e-stack-8-5-3-cgno-4899
eck-e2e-stack-8-5-3-zbjv-239
2023/07/20 13:15:25 deleting cluster: eck-e2e-gke-all-gmvc-4912
2023/07/20 13:15:25 Deleting cluster...
2023/07/20 13:15:25 deleting cluster: eck-e2e-stack-8-5-3-cgno-4899
2023/07/20 13:15:25 Deleting cluster...
2023/07/20 13:15:25 deleting cluster: eck-e2e-stack-8-5-3-zbjv-239
2023/07/20 13:15:25 Deleting cluster...
2023/07/20 13:15:25 deleted cluster: eck-e2e-gke-all-gmvc-4912
2023/07/20 13:15:25 deleted cluster: eck-e2e-stack-8-5-3-cgno-4899
2023/07/20 13:15:25 deleted cluster: eck-e2e-stack-8-5-3-zbjv-239
```

With `eks`, we also log `null`. I think it's come from the execution of an eks command executed to get the list of clusters to delete.
```
> go run main.go cleanup --provider eks
2023/07/20 13:17:07 dry-run mode enabled
2023/07/20 13:17:15 dry-run mode enabled
eck-e2e-eks-arm-qdnz-5023
null
2023/07/20 13:17:21 deleting cluster eck-e2e-eks-arm-qdnz-5023
2023/07/20 13:17:21 Deleting cluster eck-e2e-eks-arm-qdnz-5023...
2023/07/20 13:17:21 deleted cluster: eck-e2e-eks-arm-qdnz-5023
```

hack/deployer/runner/eks.go
```diff
@@ -245,4 +246,35 @@ func (e *EKSDriver) writeAWSCredentials() error {
...
+	for _, cluster := range allClusters {
+		e.ctx["ClusterName"] = cluster
+		clustersToDelete, err := exec.NewCommand(describeClusterCmd).AsTemplate(e.ctx).Output()
+		if err != nil {
+			return nil, fmt.Errorf("while describing cluster %s: %w", cluster, err)
+		}
+		if clustersToDelete != "" {
+			deletedClusters = append(deletedClusters, cluster)
+			log.Printf("deleting cluster %s", cluster)
+			if err := e.newCmd("eksctl delete cluster -v 1 --name {{.ClusterName}} --region {{.Region}} --wait").Run(); err != nil {
```
thbkrkr, line 271: I think we should create `func (e *EKSDriver) delete() error {` and reuse it here and in `func (e *EKSDriver) Execute() error` where we also delete the cluster.

hack/deployer/runner/eks.go
```diff
@@ -245,4 +246,35 @@ func (e *EKSDriver) writeAWSCredentials() error {
 	return os.WriteFile(file, []byte(fileContents), 0600)
 }
 
+func (e *EKSDriver) Cleanup() ([]string, error) {
+	if err := e.auth(); err != nil {
+		return nil, err
+	}
+	daysAgo := time.Now().Add(-24 * 3 * time.Hour)
+	e.ctx["Date"] = daysAgo.Format(time.RFC3339)
+	allClustersCmd := `eksctl get cluster -r "{{.Region}}" -o json | jq -r 'map(select(.Name|test("eck-e2e")))| .[].Name'`
```
thbkrkr, line 255: We should put string `"eck-e2e"` used to filter ECK clusters in a constant.

hack/deployer/runner/gke.go
```diff
@@ -400,3 +400,30 @@ func (d *GKEDriver) deleteDisks(disks []string) error {
...
+	}
+	if err := authToGCP(
+		d.vaultClient, GKEVaultPath, GKEServiceAccountVaultFieldName,
+		d.plan.ServiceAccount, false, d.ctx[GoogleCloudProjectCtxKey],
+	); err != nil {
+		return nil, err
+	}
+	daysAgo := time.Now().Add(-24 * 3 * time.Hour)
+	d.ctx["Date"] = daysAgo.Format(time.RFC3339)
+	cmd := `gcloud container clusters list --verbosity error --region={{.Region}} --format="value(name)" --filter="createTime<{{.Date}} AND name~eck-e2e.*"`
```
thbkrkr, line 416: We should put string `"eck-e2e"` used to filter ECK clusters in a constant.

hack/deployer/runner/aks.go
```diff
@@ -181,9 +187,48 @@ func (d *AKSDriver) GetCredentials() error {
...
+
+func (d *AKSDriver) Cleanup() ([]string, error) {
+	// Do not use docker in the cleanup steps
+	d.withoutDocker = true
+	if err := d.auth(); err != nil {
+		return nil, err
+	}
+	daysAgo := time.Now().Add(-24 * 3 * time.Hour)
+	d.ctx["Date"] = daysAgo.Format(time.RFC3339)
+	clustersCmd := `az resource list -l {{.Location}} -g {{.ResourceGroup}} --resource-type "Microsoft.ContainerService/managedClusters" --query "[?tags.project == 'eck-ci']" | jq -r --arg d "{{.Date}}" 'map(select(.createdTime | . <= $d))|.[].name'`
```
thbkrkr, line 210: Shouldn't we use the same strategy as on gke and eks and filter ECK clusters by name? Imagine if we move the `e2e-monitor` cluster to aks, wouldn't it be deleted if the filter is on the tags.project?

hack/deployer/runner/azure/azure.go
```diff
@@ -38,10 +38,15 @@ func NewCredentials(c vault.Client) (Credentials, error) {
...
-		Run()
+func Login(creds Credentials, withoutDocker bool) error {
+	if !withoutDocker {
+		return Cmd("login", "--service-principal", "-u", creds.ClientID, "-p", creds.ClientSecret, "--tenant", creds.TenantID).
+			WithoutStreaming().
+			Run()
+	}
+	command := fmt.Sprintf(`az login --service-principal -u %s -p %s --tenant %s`, creds.ClientID, creds.ClientSecret, creds.TenantID)
+	_, err := exec.NewCommand(command).Output()
+	return err
```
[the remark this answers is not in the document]
naemono: Great optimization. Thanks!

.buildkite/pipeline-e2e-test-cleanup.yml
```diff
@@ -0,0 +1,43 @@
+agents:
+  image: docker.elastic.co/ci-agent-images/cloud-k8s-operator/buildkite-agent:9203b231
+  memory: 2G
+
+steps:
+
+  - label: ":go: build deployer"
+    key: "build-deployer"
+    if: |
+      build.message == "run ECK e2e test cleanup"
```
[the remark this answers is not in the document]
naemono: Leftover from my testing, removed.

hack/deployer/cmd/cleanup.go
```diff
@@ -0,0 +1,80 @@
...
+			case runner.KindDriverID:
+				return errUnimplemented
+			default:
+				return fmt.Errorf("unknown provider %s", provider)
+			}
+		},
+	}
+
+	cleanupCmd.Flags().StringVar(&plansFile, "plans-file", "config/plans.yml", "File containing execution plans.")
+	cleanupCmd.Flags().StringVar(&provider, "provider", "gke", "Provider to use.")
```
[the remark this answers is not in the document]
naemono: I just used a `DurationVar` instead, which removes having to parse the duration. Lmk your thoughts.

hack/deployer/runner/eks.go
```diff
@@ -245,4 +246,35 @@ func (e *EKSDriver) writeAWSCredentials() error {
...
+	allClusters, err := exec.NewCommand(allClustersCmd).AsTemplate(e.ctx).OutputList()
+	if err != nil {
+		return nil, err
+	}
+	var deletedClusters []string
+	describeClusterCmd := `aws eks describe-cluster --name "{{.ClusterName}}" --region "{{.Region}}" | jq -r --arg d "{{.Date}}" 'map(select(.cluster.createdAt | . <= $d))|.[].cluster.name'`
```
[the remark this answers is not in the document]
naemono: Well this one is a bit annoying tbh. You don't get much information at all from `eksctl get cluster`, certainly do not get the `cluster.createdAt` piece which we need so you have to run `describe-cluster`.

catalog-info.yaml
```diff
@@ -142,3 +142,32 @@ spec:
...
+    apiVersion: buildkite.elastic.dev/v1
+    kind: Pipeline
+    metadata:
+      name: cloud-on-k8s-operator-e2e-cluster-cleanup
+      description: Cleanup Elastic Cloud on Kubernetes (ECK) e2e tests
+    spec:
+      repository: elastic/cloud-on-k8s
+      pipeline_file: .buildkite/pipeline-e2e-test-cleanup.yml
+      provider_settings:
+        trigger_mode: none
```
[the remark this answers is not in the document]
naemono: I added a schedule about an hour before the e2e nightly tests run, which seems to make sense. lmk if that works.

hack/deployer/runner/eks.go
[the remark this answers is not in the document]
thbkrkr: I missed that we get the `createdAt` information with this call. It's good like that.

hack/deployer/runner/aks.go
naemono: Good point. I've added this check to the jq piece of this command.

hack/deployer/cmd/cleanup.go
naemono: I've adjusted to use `WithoutStreaming` to most of these commands, and removed any extraneous print statements.

hack/deployer/runner/gcloud.go
```diff
@@ -43,6 +45,10 @@ func authToGCP(
 			return err
 		}
 
+		if gCloudProject == "" {
+			gCloudProject = "elastic-cloud-dev"
+		}
```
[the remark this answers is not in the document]
naemono: I've moved this to be pulled from vault, as it's already there.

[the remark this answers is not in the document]
pebrc: I am confused: if we don't have `deployer-config.yaml` how can we have vault access? And why can't we have deployer config here as well? And why is duplicating the information we usually push from the buildkite pipeline now in vault an acceptable solution? And if vault is an acceptable solution why are we keeping the option to define the project via the deployer configuration. I think we should try to reduce complexity with every change instead of adding to it.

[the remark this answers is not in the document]
thbkrkr: > if we don't have deployer-config.yaml how can we have vault access?

By setting VAULT_ADDR and VAULT_ROOT_PATH. I know it's a bit confusing to have this info set via env var and the others via the deployer-config but it's not something cloud-specific, so I think it's okay. It's also aligned with https://github.com/elastic/cloud-on-k8s/issues/6561.

> why can't we have deployer config here as well?

I did not propose this because I did not find it natural but yes we could, by setting a fake `CLUSTER_NAME` or adapt set-deployer-config.sh to make it optional.
- `> E2E_PROVIDER=kind CLUSTER_NAME=n/a .buildkite/scripts/test/set-deployer-config.sh`

>  if vault is an acceptable solution why are we keeping the option to define the project via the deployer configuration.

I wasn't explicit but my idea was, if we decide to move it to Vault, to remove it from the deployer config. Otherwise it is confusing to have the two.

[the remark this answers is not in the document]
pebrc: Just to be clear I am fine with moving the project name to vault, but I think if we do it we should completely remove the option from the deployer config as you said it is otherwise confusing.

hack/deployer/runner/aks.go
```diff
@@ -181,9 +187,48 @@ func (d *AKSDriver) GetCredentials() error {
...
+func (d *AKSDriver) Cleanup() ([]string, error) {
+	// Do not use docker in the cleanup steps
+	d.withoutDocker = true
+	if err := d.auth(); err != nil {
+		return nil, err
+	}
+	daysAgo := time.Now().Add(-24 * 3 * time.Hour)
+	d.ctx["Date"] = daysAgo.Format(time.RFC3339)
+	clustersCmd := fmt.Sprintf(`az resource list -l {{.Location}} -g {{.ResourceGroup}} --resource-type "Microsoft.ContainerService/managedClusters" --query "[?tags.project == 'eck-ci']" | jq -r --arg d "{{.Date}}" 'map(select((.createdTime | . <= $d) and (.name|test("%s"))))|.[].name'`, e2eClusterNamePrefix)
+	clustersToDelete, err := exec.NewCommand(clustersCmd).AsTemplate(d.ctx).OutputList()
```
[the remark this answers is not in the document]
naemono: Good point. Fixed

hack/deployer/runner/driver.go
```diff
@@ -12,8 +12,9 @@ import (
 )
 
 const (
-	CreateAction = "create"
-	DeleteAction = "delete"
+	CreateAction         = "create"
+	DeleteAction         = "delete"
+	e2eClusterNamePrefix = "eck-e2e"
```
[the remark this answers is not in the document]
naemono: @pebrc I think I've done what you're asking. Could you verify when you get some time please?

hack/deployer/runner/aks.go
```diff
@@ -181,9 +187,48 @@ func (d *AKSDriver) GetCredentials() error {
...
 func (d *AKSDriver) delete() error {
 	log.Print("Deleting cluster...")
-	return azure.Cmd("aks",
+	args := []string{"aks",
 		"delete", "--yes",
 		"--name", d.plan.ClusterName,
-		"--resource-group", d.plan.Aks.ResourceGroup).
-		Run()
+		"--resource-group", d.plan.Aks.ResourceGroup}
+	if !d.withoutDocker {
```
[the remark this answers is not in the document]
naemono: Had to give a bit of history here to try and explain this approach.

Was building the deployer using this image
```
  image: docker.elastic.co/ci-agent-images/cloud-k8s-operator/buildkite-agent:abaeba8c
```

And was running as follows with another image:
```
agents:
      # docker is needed for running aks deployer
      provider: "gcp"
      image: "family/elastic-buildkite-agent-ubuntu-2004-lts"
      machineType: "n1-standard-16"
commands:
      - buildkite-agent artifact download "hack/deployer/deployer" /tmp
      - chmod u+x /tmp/hack/deployer/deployer
      - /tmp/hack/deployer/deployer cleanup --plans-file hack/deployer/config/plans.yml --provider aks

```
Note what this was doing:

https://buildkite.com/elastic/cloud-on-k8s-operator/builds/4818#018931c7-a218-476b-88da-01287449f0bb

```
2023-07-07 19:17:01 INFO   Successfully downloaded "hack/deployer/deployer" 24417992 bytes
--
  | /tmp/hack/deployer/deployer: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by /tmp/hack/deployer/deployer)
  | /tmp/hack/deployer/deployer: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by /tmp/hack/deployer/deployer)
  | 🚨 Error: The command exited with status 1
```

Which led me to just try and run the deployer for aks without docker, as the `az` binary was already contained within the buildkite agent image.

Regardless, what I turned out to be missing, which is how things are being done now, is wrapping this whole process in docker, which actually (re)builds the binary only for aks which isn't exactly optimal, but it works and we don't have the divergent paths with conditionals for Azure/aks.

```
- make -C .buildkite TARGET="run-deployer-cleanup-aks" ci
```

hack/deployer/runner/gcloud.go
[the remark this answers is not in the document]
naemono: I will create an issue to try and completely remove project names from code and into vault. This PR is now using deployer-config similar to other commands.

hack/deployer/cmd/cleanup.go
```diff
@@ -0,0 +1,79 @@
...
+		Use:   "cleanup",
+		Short: "Runs the cleanup operation to cleanup clusters older than 3 days in the given provider.",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			switch provider {
+			case runner.GKEDriverID:
+				return cleanup(plansFile, []string{"gke-ci"}, &runner.GKEDriverFactory{})
+			case runner.AKSDriverID:
+				return cleanup(plansFile, []string{"aks-ci"}, &runner.AKSDriverFactory{})
+			case runner.OCPDriverID:
+				return errUnimplemented
```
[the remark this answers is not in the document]
naemono: OCP/Tanzu have been added to the cleanup steps after some investigation into how that actually could be done. switch/case no longer needed after optimization so it's bee removed.

hack/deployer/runner/ocp.go
```diff
@@ -462,3 +462,29 @@ func (d *OCPDriver) baseDomain() string {
...
+		return nil, err
+	}
+	daysAgo := time.Now().Add(-olderThan)
+
+	params := d.bucketParams()
+	params["Date"] = daysAgo.Format(time.RFC3339)
+	params["E2EClusterNamePrefix"] = prefix
+	params["Region"] = d.plan.Ocp.Region
+
+	cmd := `gcloud compute instances list --verbosity error --zones={{.Region}}-a,{{.Region}}-b,{{.Region}}-c --filter="name~'^{{.E2EClusterNamePrefix}}-ocp.*-master.*' AND status=RUNNING" --format=json | jq -r --arg d "{{.Date}}" 'map(select(.creationTimestamp | . <= $d))|.[].name' | grep -o '{{.E2EClusterNamePrefix}}-ocp-[a-z]*-[0-9]*' | sort | uniq`
```
barkbay, line 477: Should we not list the zones in the region instead of hardcoding them? Maybe with something along those lines:
```bash
gcloud compute zones list --filter='region:https://www.googleapis.com/compute/v1/projects{{.GCloudProject}}regions/{{.Region}}' --format="value(selfLink.name())"
```

barkbay, line 477: Why do we filter on "masters" only? I wonder if we could hit a case where only the bootstrap vm has been created.
```suggestion
	cmd := `gcloud compute instances list --verbosity error --zones={{.Region}}-a,{{.Region}}-b,{{.Region}}-c --filter="name~'^{{.E2EClusterNamePrefix}}-ocp.*' AND status=RUNNING" --format=json | jq -r --arg d "{{.Date}}" 'map(select(.creationTimestamp | . <= $d))|.[].name' | grep -o '{{.E2EClusterNamePrefix}}-ocp-[a-z]*-[0-9]*' | sort | uniq`
```

barkbay (commented):
I think I'm fine listing the vms, one concern is that an OCP cluster is more than a set of VMs, there are also network objects (vpc, firewall rules, dns settings), buckets, service accounts ... I'm wondering if there is another way to ensure that they are all eventually garbage collected. That being said I'm 👍 about using the VM names for now.

hack/deployer/runner/ocp.go
naemono: There's probably no great reason. My thought was that if the cluster is online, then the master has to be online, so search for signs of it. I'll update and we can modify in the future if needed.

naemono:
> I think I'm fine listing the vms, one concern is that an OCP cluster is more than a set of VMs, there are also network objects (vpc, firewall rules, dns settings), buckets, service accounts ... I'm wondering if there is another way to ensure that they are all eventually garbage collected. That being said I'm 👍 about using the VM names for now.

Well, If I find signs of a vm, I'm running the ocp delete tool wrapped in docker, so it hopefully cleans up as much as it can? Honestly I'm not totally sure of the extend of cleanup the tool itself does. We can obviously modify this approach in the future if we find dangling objects from clusters.

hack/deployer/runner/ocp.go
naemono: I think this makes sense. I'll get that adjusted.

hack/deployer/runner/driver.go
```diff
@@ -31,6 +32,8 @@ type Driver interface {
 	Execute() error
 	// GetCredentials updates a kubeconfig file with appropriate credentials for the current environment.
 	GetCredentials() error
+	// Cleanup will attempt to delete e2e test clusters older than the given time.Duration in a cloud provider.
+	Cleanup(string, time.Duration) ([]string, error)
```
pebrc, line 36: Return value `[]string` seems unused.

hack/deployer/runner/eks.go
```diff
@@ -245,4 +246,48 @@ func (e *EKSDriver) writeAWSCredentials() error {
...
+		}
+	}
+
+	return deletedClusters, nil
+}
+
+func (e *EKSDriver) delete() error {
+	log.Printf("deleting cluster %s", e.ctx["ClusterName"])
+	// --wait to surface failures to delete all resources in the Cloud formation
+	if err := e.newCmd("eksctl delete cluster -v 1 --name {{.ClusterName}} --region {{.Region}} --wait").Run(); err != nil {
```
pebrc, line 287: if statement seems redundant

hack/deployer/runner/aks.go
```diff
@@ -180,10 +185,45 @@ func (d *AKSDriver) GetCredentials() error {
...
+	if err := d.auth(); err != nil {
+		return nil, err
+	}
+
+	daysAgo := time.Now().Add(-olderThan)
+	d.ctx["Date"] = daysAgo.Format(time.RFC3339)
+	d.ctx["E2EClusterNamePrefix"] = prefix
+
+	clustersCmd := `az resource list -l {{.Location}} -g {{.ResourceGroup}} --resource-type "Microsoft.ContainerService/managedClusters" --query "[?tags.project == 'eck-ci']" | jq -r --arg d "{{.Date}}" 'map(select((.createdTime | . <= $d) and (.name|test("{{.E2EClusterNamePrefix}}"))))|.[].name'`
+	clustersToDelete, err := exec.NewCommand(clustersCmd).AsTemplate(d.ctx).OutputList()
```
pebrc, line 206: This command still directly uses AKS CLI

hack/deployer/runner/ocp.go
```diff
@@ -462,3 +463,45 @@ func (d *OCPDriver) baseDomain() string {
...
+
+	cmd := `gcloud compute instances list --verbosity error --zones={{.Zones}} --filter="name~'^{{.E2EClusterNamePrefix}}-ocp.*' AND status=RUNNING" --format=json | jq -r --arg d "{{.Date}}" 'map(select(.creationTimestamp | . <= $d))|.[].name' | grep -o '{{.E2EClusterNamePrefix}}-ocp-[a-z]*-[0-9]*' | sort | uniq`
+	clustersToDelete, err := exec.NewCommand(cmd).AsTemplate(params).WithoutStreaming().OutputList()
+	if err != nil {
+		return nil, err
+	}
+
+	for _, cluster := range clustersToDelete {
+		d.plan.ClusterName = cluster
+		if err = d.delete(); err != nil {
```
pebrc, line 502: This is not going to work. You have to run through the setup steps as in `Execute` including `ensureClientImage`

hack/deployer/runner/tanzu.go
```diff
@@ -442,4 +443,45 @@ func (t *TanzuDriver) restoreInstallerState() error {
...
+		clustersToDelete, err := exec.NewCommand(clustersCmd).AsTemplate(params).WithoutStreaming().OutputList()
+		if err != nil {
+			return nil, fmt.Errorf("while running az resource list command: %w", err)
+		}
+
+		for _, cluster := range clustersToDelete {
+			t.plan.ClusterName = cluster
+			t.plan.Tanzu.ResourceGroup = cluster
+			log.Printf("deleting cluster %s\n", cluster)
+			if err = t.delete(); err != nil {
```
pebrc, line 478: Same here: this is not going to work. You need to either call `Execute` or run the `setup` function

naemono: Yeah, this is due to my unfamiliarity with these wrapped docker commands. I've updated all of these and am testing now with existing tanzu clusters that need to be deleted, and a custom ocp cluster I can fake needing to be deleted. This will verify these cleanup commands for both of these. Thanks for catching this.

thbkrkr (commented):
It's almost good. Thanks for your patience.

naemono:
OCP cleanup was also a success after the most recent changes. This should b ready for the next set of 👀 .

hack/deployer/runner/tanzu.go
```diff
@@ -442,4 +443,54 @@ func (t *TanzuDriver) restoreInstallerState() error {
...
+			"| jq -r --arg d", sinceDate.Format(time.RFC3339),
+			fmt.Sprintf(`'map(select((.createdTime | . <= $d) and (.name|test("%s-tanzu"))))|.[].name'`, prefix),
+			fmt.Sprintf("| grep -o '%s-tanzu-[a-z]*-[0-9]*' | sort | uniq", prefix)).OutputList()
+		if err != nil {
+			return fmt.Errorf("while running az resource list command: %w", err)
+		}
+
+		for _, cluster := range clustersToDelete {
+			t.plan.ClusterName = cluster
+			t.plan.Tanzu.ResourceGroup = cluster
```
pebrc, line 483:
```suggestion
			t.plan.Tanzu.ResourceGroup = rg
```

pebrc (approved):
LGTM one thing I wonder though if we want to continue with the deletion in the loops when one cluster deletion fails.

hack/deployer/runner/tanzu.go
naemono: I've updated this, even though resourcegroup == cluster in tanzu, this makes it more clear.

Also, I've continued on failure for all cluster types.

PR #6844, opened 2023-05-23
[DOC] Mention base_path change in 2.8.0 release notes

### LINKED ISSUE #6702: Mention #6689 in 2.8 release notes
Mention the change introduced in #6689 in the release notes, as the behavior will change for users who attempted to set a `base_path` and inadvertently kept it in their configs.

barkbay:
Fixes #6702


docs/release-notes/highlights-2.8.0.asciidoc
```diff
@@ -62,4 +62,11 @@ securityContext:
...
\ No newline at end of file
+It is now possible to fully delegate the generation of the transport certificates used by the Elasticsearch nodes. Refer to <<{p}-transport-third-party-tools>> for more information about the requirements as well as some examples using the link:https://cert-manager.io/docs/projects/csi-driver/[cert-manager csi-driver] and link:https://cert-manager.io/docs/projects/trust-manager/[trust-manager] projects.
+
+[float]
+[id="{p}-280-stack-config-base-path"]
+=== Provided base_path setting in snapshot repositories configuration is always applied
+
+Before ECK 2.8.0 the `base_path` setting was overridden by the operator in order to avoid conflict. The value provided for this setting is now always used when provided. Refer to <<{p}-stack-config-policy-specifics-snap-repo>> for more details.
```
pebrc, line 71:
```suggestion
Before ECK 2.8.0 the `base_path` setting was overridden by the operator in order to avoid conflicts. The value provided for this setting is now always used when provided. Refer to <<{p}-stack-config-policy-specifics-snap-repo>> for more details.
```

pebrc (approved):
LGTM. Is this a "breaking" change?

barkbay:
> LGTM. Is this a "breaking" change?

Only if user attempted to set it in the past, and forgot it is set while it's actually not used by the current version of ECK. In such a case, upgrading to `2.8.0` will enable this setting which was ignored until then. I don't _think_ it is a breaking change, @thbkrkr what do you think?

thbkrkr:
> I don't think it is a breaking change, @thbkrkr what do you think?

I'd say it's not a breaking change because we don't break anything if you don't make a mistake but if you do, it will break you. I'd therefore be on the side of caution and I'll present this change as a breaking change.

barkbay:

thbkrkr (commented):
Nit: Sorry if I'm a bit late but I think we should rename "Copy Elasticsearch configuration files before creating links" #6703 in something like "Hardened Security Context for Elasticsarch".

I think there are people who only read the release notes and not the highlights.

barkbay:
buildkite test this

thbkrkr (approved):

barkbay:

PR #6854, opened 2023-05-25
Add Selector to LogstashStatus

robbavey:
While running some tests with the VerticalPodAutoscaler, I was unable to make it work with the Logstash CRD, which appears to be due to the lack of a selector in the LogstashStatus object. This commit adds the missing Selector.

### DIFF
--- a/pkg/controller/logstash/reconcile.go
+++ b/pkg/controller/logstash/reconcile.go
@@ -62,15 +62,13 @@ func reconcileStatefulSet(params Params, podTemplate corev1.PodTemplateSpec) (*r
 func calculateStatus(params *Params, sset appsv1.StatefulSet) (logstashv1alpha1.LogstashStatus, error) {
 	logstash := params.Logstash
 	status := params.Status
-	labelSelector := sset.Spec.Selector
-
 	pods, err := k8s.PodsMatchingLabels(params.Client, logstash.Namespace, map[string]string{NameLabelName: logstash.Name})
 	if err != nil {
 		return status, err
 	}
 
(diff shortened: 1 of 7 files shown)


pkg/controller/logstash/reconcile.go
```diff
@@ -49,26 +51,34 @@ func reconcileStatefulSet(params Params, podTemplate corev1.PodTemplateSpec) (*r
...
 		return status, err
 	}
 
+	if labelSelector != nil {
+		selector, err := metav1.LabelSelectorAsSelector(labelSelector)
+		if err != nil {
+			return logstashv1alpha1.LogstashStatus{}, err
+		}
+		status.Selector = selector.String()
+	}
```
barkbay, line 78: nit: move the `labelSelector` variable closer to where it is used, I think we can omit it as well:

```suggestion
	pods, err := k8s.PodsMatchingLabels(params.Client, logstash.Namespace, map[string]string{NameLabelName: logstash.Name})
	if err != nil {
		return status, err
	}

	if sset.Spec.Selector != nil {
		selector, err := metav1.LabelSelectorAsSelector(sset.Spec.Selector)
		if err != nil {
			return logstashv1alpha1.LogstashStatus{}, err
		}
		status.Selector = selector.String()
	}
```

barkbay (commented):
LGTM

barkbay (approved):

robbavey:
Thank you @barkbay! I am unable to merge into this repo, so I would appreciate if someone could do this for me once CI is green

PR #6856, opened 2023-05-26
Fix rendering of autopilot recipes README

thbkrkr:
This fixes the rendering of the autopilot recipes README that is written in the AsciiDoc format by:
- renaming `README.md` to `README.asciidoc`

Also:
- add some line breaks to space the text
- move Kibana at the top in manifests where it is not
- move manifest names from the titles to a command


barkbay (approved):
LGTM

PR #6875, opened 2023-06-02
changing build-operator-image to publish-operator-image in make deploy

kvalliyurnatt:
This commit replaces references to build-operator-image with publish-operator-image in the makefile as build-operator-image is no longer a valid make target


thbkrkr:

PR #6885, opened 2023-06-06
Removing autoscale annotation and related code

kvalliyurnatt:
Since autoscaling annotation has been deprecated and we have a CRD for autoscaling now, removing code related to the autoscaling annotation.


thbkrkr:
Looks good!

We can also removed `pkg/controller/autoscaling/elasticsearch/testdata/annotation/` which is no longer used.

There may also be validation that annotation and Autoscaling CRD cannot be used at the same time but I'd like to get @barkbay's opinion first? It doesn't cost much to keep but likely rephrase the validation error message to say that the annotation support has been removed and only the CRD must be used.
I'm talking about these bits:
- `noAutoscalingAnnotation()` in `pkg/controller/autoscaling/elasticsearch/validation/validations.go`
- unit test "User should not use both the Autoscaling CRD and the annotation" in `pkg/controller/autoscaling/elasticsearch/controller_test.go`
- unit test "Using both the autoscaler and the autoscaling annotation" in `pkg/controller/autoscaling/elasticsearch/validation/webhook_test.go`
 - `pkg/controller/autoscaling/elasticsearch/testdata/custom_resource/annotation/`

barkbay:
buildkite test this -f p=gke,t=TestAutoscaling,s=8.8.0

barkbay (commented):
Looks good 👍 

> There may also be validation that annotation and Autoscaling CRD cannot be used at the same time but I'd like to get @barkbay's opinion first? It doesn't cost much to keep but likely rephrase the validation error message to say that the annotation support has been removed and only the CRD must be used.

I think I'm 👍 with Thibault's suggestion. The annotation should not be used at all now.

thbkrkr (approved):
LGTM!

barkbay (approved):
LGTM !

PR #6895, opened 2023-06-11
Logstash connect to external Elasticsearch

kaisecheng:
This PR adds doc for connecting external Elaticsearch

### DIFF
--- a/docs/orchestrating-elastic-stack-applications/logstash.asciidoc
+++ b/docs/orchestrating-elastic-stack-applications/logstash.asciidoc
@@ -380,7 +380,7 @@ kind: Secret
 metadata:
   name: external-es-ref
 stringData:
-  url: https://abcd-42.xyz.elastic-cloud.com <1>
+  url: https://abcd-42.xyz.elastic-cloud.com:443 <1>
   username: logstash_user <2>
   password: REDACTED <3>
   ca.crt: REDACTED <4>
@@ -409,6 +409,8 @@ spec:
 <3> The password of the user to be authenticated to the Elasticsearch cluster.
 <4> The CA certificate in PEM format to secure communication to the Elasticsearch cluster (optional).
 <5> The `secretName` and `name` attributes are mutually exclusive, you have to choose one or the other.
+
+NOTE: Please always specify the port in URL when connect to external Elasticsearch Cluster.
 
 [id="{p}-logstash-expose-services"]
 === Expose services
--- a/docs/orchestrating-elastic-stack-applications/logstash.asciidoc
+++ b/docs/orchestrating-elastic-stack-applications/logstash.asciidoc
@@ -288,10 +288,10 @@ NOTE: Logstash persistent queues (PQs) and dead letter queues (DLQs) are not cur
 [id="{p}-logstash-pipelines-es"]
 === Using Elasticsearch in Logstash pipelines
 
-The `spec.elasticsearchRefs` section provides a mechanism to help configure Logstash to estabish a secured connection to one or more managed Elasticsearch clusters. By default, each `elasticsearchRef` will target all nodes in its referenced Elasticsearch cluster. If you want to direct traffic to specific nodes of your Elasticsearch cluster, refer to <<{p}-traffic-splitting>> for more information and examples.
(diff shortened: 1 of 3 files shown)


pkg/controller/logstash/env_test.go
```diff
@@ -162,3 +163,60 @@ func Test_getEnvVars(t *testing.T) {
...
+			wantErr:  false,
+		},
+		{
+			name:     "Non-HTTPS URL",
+			url:      "http://some.gcp.cloud.es.io",
+			expected: "http://some.gcp.cloud.es.io",
+			wantErr:  false,
+		},
+		{
+			name:     "Non-HTTPS without port",
```
robbavey, line 193:
```suggestion
			name:     "Non-HTTPS with port",
```

docs/orchestrating-elastic-stack-applications/logstash.asciidoc
```diff
@@ -367,6 +367,48 @@ spec:
...
+You can reference a Secret instead of an Elasticsearch cluster in the `elasticsearchRefs` section through the `secretName` attribute:
+
+[source,yaml,subs="attributes,callouts"]
+----
+apiVersion: v1
+kind: Secret
+metadata:
+  name: external-es-ref
+stringData:
+  url: https://abcd-42.xyz.elastic-cloud.com <1>
```
robbavey, line 383: I think we should add a note somewhere that we add the default port of `443` to `https` URLs, as this is different behavior to standard logstash plugins, and different to how non-cloud Elasticsearch works out of the box

robbavey (commented):
LGTM - I have a documentation suggestion, and a nit on test naming

docs/orchestrating-elastic-stack-applications/logstash.asciidoc
```diff
@@ -367,6 +367,48 @@ spec:
...
+spec:
+  version: {version}
+  count: 1
+  elasticsearchRefs:
+    - clusterName: prod-es
+      secretName: external-es-ref <5>
+  monitoring:
+    metrics:
+      elasticsearchRefs:
+      - secretName: external-es-ref <5>
```
pebrc, line 401: Assuming a user specifies a URL without explicit port in the secret, then the changes in this PR append 443 by default for the environment. However Beats also assume 9200 by default and while the Logstash side is fixed, the monitoring will still not work until the user fixes the `url` field in the secret to include a port.

pebrc (commented):
I think I am slightly -1 on the port appending. As explained in my comment it does fix the problem only for Logstash and not for any of the other stack applications e.g. the Beats used for monitoring. I think I would favour calling out explicitly in the doc that the default port expected is 9200 and any other port has to be configured explicitly via the url.

kaisecheng:
Kibana assumes 443. Beats and Logstah assume 9200. Not sure what the default port is for others. The stack monitoring is not consistent.
+1 for calling out explicitly in doc. Instead of explaining the assumption of default port, having a note to ask for always add the port in URL may be simpler.

kaisecheng:
@robbavey @pebrc I have removed the port appending. Now the PR only updates the doc. It is ready for review.

docs/orchestrating-elastic-stack-applications/logstash.asciidoc
```diff
@@ -367,6 +367,50 @@ spec:
...
+----
+
+<1> The URL to reach the Elasticsearch cluster.
+<2> The username of the user to be authenticated to the Elasticsearch cluster.
+<3> The password of the user to be authenticated to the Elasticsearch cluster.
+<4> The CA certificate in PEM format to secure communication to the Elasticsearch cluster (optional).
+<5> The `secretName` and `name` attributes are mutually exclusive, you have to choose one or the other.
+
+NOTE: Please always specify the port in URL when connect to external Elasticsearch Cluster.
```
pebrc, line 413:
```suggestion
NOTE: Please always specify the port in URL when connecting to an external Elasticsearch Cluster.
```

pebrc (approved):

robbavey (approved):
LGTM

pebrc:
@kaisecheng should I merge this?

kaisecheng:
@pebrc Yes please!

PR #6905, opened 2023-06-14
Revisit how to get e2e tests results

thbkrkr:
This changes the way to get e2e test results by downloading them from the k8s job instead of writing them locally from the k8s job log stream.

What changes?

In the remote k8s job:
- the pod will never terminate (`config/e2e/e2e_job.yaml`)
- the pod is set as ready when `go test` is completed (`run.sh`)
- `go test` output is written to `/tmp/e2e-tests.json` in the pod (`run.sh`)

In the jobManager in the local test runner:
- change the file name to write the log stream
- `kubectl cp ns/name:/tmp/e2e-tests.json ./e2e-tests-$clusterName.json` when the pod is ready
- stop to `job.WaitForLogs()`, just `jobManager.Stop()`


test/e2e/cmd/run/job.go
```diff
@@ -49,29 +52,18 @@ func NewJob(podName, templatePath string, writer io.Writer, timestampExtractor t
...
-}
-
 // Stop is only a best effort to stop the streaming process
 func (j *Job) Stop() {
 	if j.stopRequested {
 		// Job already stopped
 		return
 	}
 	close(j.stopLogStream)
+	close(j.streamErrors)
```
thbkrkr, line 66: I'm not sure why we didn't close this channel.

test/e2e/run.sh
```diff
@@ -32,6 +32,8 @@ main() {
   else
     run_e2e_tests "$@"
   fi
+  touch /tmp/done
+  sleep infinity
 }
 
-main "$@"
+main "$@" | tee "$RESULT_FILE"
```
naemono, line 39: Would it be better to have the `tee` command living within the `run_e2e-tests` and `run_chaos` functions such that it closes the`RESULT_FILE` properly? Wouldn't we kinda be leaving an open file handle?

naemono (approved):
Mostly lgtm. bit of a question about the location of the `tee` command.

test/e2e/run.sh
thbkrkr: Yes, you're right, we can move the `tee` and avoid that, though I don't think it can cause an issue.

thbkrkr: I will take the opportunity to do it only for the e2e tests job, not the chaos job.

test/e2e/cmd/run/job_manager.go
```diff
@@ -128,24 +129,31 @@ func (jm *JobsManager) Start() {
...
 			switch newPod.Status.Phase {
 			case corev1.PodRunning:
 				job.onPodEvent(jm.Clientset, newPod)
-			case corev1.PodSucceeded:
-				if job.podSucceeded {
-					// already done, but the informer is not stopped yet so this code is still running
-					return
+
+				// download result file when pod is ready
+				if podutils.IsPodReady(newPod) {
```
pebrc, line 138: Is that really worth the extra `kubectl` dependency? I don't see a big difference to our existing `k8s.IsPodReady` (we also check for `ContainersReady`)

https://github.com/elastic/cloud-on-k8s/blob/b6b50758127d24c1e67b62799711864fcd138582/pkg/utils/k8s/k8sutils.go#L67-L75

pebrc (commented):
Approach LGTM. I remember now that the only reason I used this approach in eck-diagnostics (compared to running the work bit (here: the tests) in an init-container and the having the main container do nothing and wait for the download) was to avoid issues with Istio where init containers cannot do HTTP requests. But I think it makes sense here as well.

test/e2e/cmd/run/job_manager.go
thbkrkr: Really not, I'll use ours.

test/e2e/cmd/run/job.go
thbkrkr: I'll remove `stopRequested`.

What confused me is that:
In `WaitForLogs()`, we closed the 2 channels `stopLogStream` and `streamErrors`.
In `Stop()`, we closed only channel `stopLogStream`.
We almost always called `WaitForLogs()` and then `Stop()` except on pod deletion.

PR #6910, opened 2023-06-15
Remove last Jenkins bits

thbkrkr:
This removes the last bits related to Jenkins that are no longer needed since switching to buildkite.


pebrc (approved):
LGTM

PR #6917, opened 2023-06-15
Track job stops and file downloads

pebrc:
To avoid panic when `jobManager.Stop()` is called more than once if the informer triggers more than one Pod update event in a ready state.

### DIFF
--- a/test/e2e/cmd/run/job.go
+++ b/test/e2e/cmd/run/job.go
@@ -77,8 +77,7 @@ func (j *Job) WithResultFile(f string) *Job {
 // onPodEvent ensures that log streaming is started and also manages the internal state of the Job based on the events
 // received from the informer.
 func (j *Job) onPodEvent(client *kubernetes.Clientset, pod *corev1.Pod) {
-	switch pod.Status.Phase { //nolint:exhaustive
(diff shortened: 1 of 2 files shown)


test/e2e/cmd/run/job.go
```diff
@@ -73,8 +77,7 @@ func (j *Job) WithResultFile(f string) *Job {
 // onPodEvent ensures that log streaming is started and also manages the internal state of the Job based on the events
 // received from the informer.
 func (j *Job) onPodEvent(client *kubernetes.Clientset, pod *corev1.Pod) {
-	switch pod.Status.Phase { //nolint:exhaustive
-	case corev1.PodRunning:
+	if pod.Status.Phase == corev1.PodRunning {
```
naemono, line 80: Probably just want to `if pod.Status.Phase == corev1.PodRunning && !j.jobStarted {` at this point.

naemono (approved):
I think this fixes the issue, but can be simplified just a little more.

test/e2e/cmd/run/job.go
pebrc: Right!

PR #6918, opened 2023-06-16
Do not close channels, rem errexit

pebrc:
Avoid having to wait for log completion and avoid running into "send on closed channel" errors by not closing the error and log channel in the job abstraction we use for e2e tests. The rationale is that the program will exit anyway at this point and we do not need to manager those resources. 

Remove the `errexit` shell option from `run.sh` to avoid the script exiting early on errors and preventing the job manager to download the log file.

### DIFF
--- a/test/e2e/cmd/run/job.go
+++ b/test/e2e/cmd/run/job.go
@@ -22,8 +22,7 @@ type Job struct {
 	runningWg  *sync.WaitGroup // wait for the dependency to be started
 
 	// Job context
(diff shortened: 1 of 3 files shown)


thbkrkr:
buildkite test this -f p=gke,s=8.9.0-SNAPSHOT -m t=TestSmoke,t=TestCrossNSAssociation

thbkrkr (approved):
LGTM

PR #6923, opened 2023-06-19
Add pipeline to release Helm charts

thbkrkr:
This creates a new pipeline `cloud-on-k8s-operator-helm-release` to trigger only the release of the eck-operator and/or eck-stack Helm charts:

- Add an entry in `catalog-info.yaml` to create the new pipeline that triggers `pipeline-helm-release.yml`
- Keep the step to build the helm releaser tool on every build in `pipeline-build.yml` to make sure it compiles. 
- Add a step to build the helm releaser tool in `pipeline-helm-release.yml` to make it independent of `pipeline-build.yml` that will be not run when triggered by this new pipeline.

### DIFF
--- a/.buildkite/pipeline-build.yml
+++ b/.buildkite/pipeline-build.yml
@@ -64,8 +64,6 @@ steps:
           machineType: "n1-standard-16"
 
       - label: ":go: helm releaser tool"
-        key: "build-helm-releaser-tool"
         commands:
           - cd hack/helm/release
           - make build
-          - buildkite-agent artifact upload bin/releaser
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -51,4 +51,6 @@ steps:
     command: buildkite-agent pipeline upload .buildkite/pipeline-e2e-tests.yml
 
   - label: ":buildkite:"
+    depends_on:
+      - "operator-image-build"
     command: buildkite-agent pipeline upload .buildkite/pipeline-release.yml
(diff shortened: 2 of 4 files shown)


.buildkite/pipeline.yml
```diff
@@ -51,4 +51,6 @@ steps:
     command: buildkite-agent pipeline upload .buildkite/pipeline-e2e-tests.yml
 
   - label: ":buildkite:"
+    depends_on:
+      - "operator-image-build"
```
kvalliyurnatt, line 55: with this change, do we need this ?
https://github.com/elastic/cloud-on-k8s/blob/dcea160ed9540b9c1da3c63f43cdbb84705610b9/.buildkite/pipeline-release.yml#L10

thbkrkr: There is no longer a dependency between the build and the helm release, but we still don't want to release the helm charts before the docker images referenced therein are released.

kvalliyurnatt (approved):

.buildkite/pipeline-build.yml
```diff
@@ -64,8 +64,6 @@ steps:
           machineType: "n1-standard-16"
 
       - label: ":go: helm releaser tool"
-        key: "build-helm-releaser-tool"
         commands:
           - cd hack/helm/release
           - make build
```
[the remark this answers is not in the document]
thbkrkr: Yes for the time being. In particular, the hack/helm/release go module depends on eck and after an upgrade of a go dep upgrade in eck, it can fail to compile if we didn't `go mod tidy`.

PR #6926, opened 2023-06-20
[helm/eck-elasticsearch] Support disabling podDisruptionBudget

### LINKED ISSUE #6871: Unable to disable podDisruptionBudget
## Bug Report
**What did you do?**
I want to disable the podDisruptionBudget by setting podDisruptionBudget with an empty hash

I should have the following directive in the result of helm chart

```
eck-elasticksearch:
  ...
  podDisruptionBudget: {}
```

directive podDisruptionBudget is absent from the manifest generated by the Helm

**Environment**

* ECK version:

	2.8.0

* Kubernetes information:
	* On premise : deployed with kubespray 2.21.0
	
	for each of them please give us the version you are using
```
$ kubectl version
WARNING: This version information is deprecated and will be replaced with the output from kubectl version --short.  Use --output=yaml|json to get the full version.
Client Version: version.Info{Major:"1", Minor:"24", GitVersion:"v1.24.6", GitCommit:"b39bf148cd654599a52e867485c02c4f9d28b312", GitTreeState:"clean", BuildDate:"2022-09-21T13:19:24Z", GoVersion:"go1.18.6", Compiler:"gc", Platform:"linux/amd64"}
Kustomize Version: v4.5.4
Server Version: version.Info{Major:"1", Minor:"25", GitVersion:"v1.25.6", GitCommit:"ff2c119726cc1f8926fb0585c74b25921e866a28", GitTreeState:"clean", BuildDate:"2023-01-18T19:15:26Z", GoVersion:"go1.19.5", Compiler:"gc", Platform:"linux/amd64"}
```

* Resource definition:
```
eck-elasticsearch:
  enabled: true
  fullnameOverride: "sophie-mon"
  version: 8.7.0
  monitoring:
    metrics:
      elasticsearchRefs:
      - name: sophie-mon
        namespace: cdn-bigdata
    logs:
      elasticsearchRefs:
      - name: sophie-mon
        namespace: cdn-bigdata
  nodeSets:
  - name: "nodes"
    count: 3
    config:
      # No need to define specific role on nodes
      node.roles: ["master","data","ingest","transform","ml","remote_cluster_client"]
    podTemplate:
      spec:
        containers:
        - name: elasticsearch
          env:
          - name: ES_JAVA_OPTS
            value: -Xms8g -Xmx8g
          resources:
            limits:
              cpu: 8
              memory: 12Gi
    volumeClaimTemplates:
    - metadata:
        name: elasticsearch-data
      spec:
        accessModes:
        - ReadWriteOnce
        resources:
          requests:
            storage: 200Gi
        storageClassName: lyo0-oc9k-nfs
  podDisruptionBudget: {}
```

* Logs:
```
No error
```

I could also check this by generating the global manifest with helm template command and my values file.

As a workaround I do an ugly hack by defining a value disablePdb on value file with true value

and the following code on template for eck-elasticsearch
```
{{- if .Values.disablePdb }}
  podDisruptionBudget: {}
  {{- else }}
  {{- with .Values.podDisruptionBudget }}
  podDisruptionBudget:
    {{- toYaml . | nindent 4 }}
  {{- end }}
  {{- end }}
```

Thanks by advance for your replies.

--
Regards
Christophe

thbkrkr:
Adds a new explicit `podDisruptionBudget.disabled` value to be able to explicitely disable the PDB for Elasticsearch because we can't differentiate no `podDisruptionBudget` set (which means ECK sets a default PDB) from an explicit empty `podDisruptionBudget: {}` (which means I want to disable the PDB).

```sh
> test() { helm template . "$@" | yq .spec.podDisruptionBudget; }
echo "# default PDB"
test
echo "# custom PDB"
test --set=podDisruptionBudget.spec.minAvailable=2 --set='podDisruptionBudget.spec.selector.matchLabels.elasticsearch\.k8s\.elastic\.co\/cluster-name=quickstart'
echo "# no PDB"
test --set=podDisruptionBudget.disabled=true

# default PDB
null
# custom PDB
spec:
  minAvailable: 2
  selector:
    matchLabels:
      elasticsearch.k8s.elastic.co/cluster-name: quickstart
# no PDB
{}
```
Resolves #6871.

### DIFF
--- a/deploy/eck-stack/charts/eck-elasticsearch/values.yaml
+++ b/deploy/eck-stack/charts/eck-elasticsearch/values.yaml
@@ -126,8 +126,8 @@ remoteClusters: {}
 volumeClaimDeletePolicy: ""
 
 # Settings to limit the disruption when pods need to be rescheduled for some reason such as upgrades or routine maintenance.
-# Default budget selects all cluster pods and sets `maxUnavailable` to 1.
-# To disable, set `disabled` to true.
+# Defaults to a budget that selects all Elasticsearch cluster pods and sets `maxUnavailable` to 1 if unset.
+# To completely disable the pod disruption budget set `disabled` to true.
 #
 # podDisruptionBudget:
 #   spec:
--- a/deploy/eck-stack/charts/eck-elasticsearch/values.yaml
+++ b/deploy/eck-stack/charts/eck-elasticsearch/values.yaml
@@ -127,7 +127,7 @@ volumeClaimDeletePolicy: ""
 
 # Settings to limit the disruption when pods need to be rescheduled for some reason such as upgrades or routine maintenance.
 # Default budget selects all cluster pods and sets `maxUnavailable` to 1.
-# To disable, set to the empty value (`{}`).
+# To disable, set `disabled` to true.
 #
 # podDisruptionBudget:
 #   spec:
@@ -135,7 +135,7 @@ volumeClaimDeletePolicy: ""
 #     selector:
 #       matchLabels:
 #         elasticsearch.k8s.elastic.co/cluster-name: quickstart
-
+#   disabled: true
 
 # Used to check access from the current resource to a resource (for ex. a remote Elasticsearch cluster) in a different namespace.
 # Can only be used if ECK is enforcing RBAC on references.
--- a/deploy/eck-stack/charts/eck-elasticsearch/values.yaml
+++ b/deploy/eck-stack/charts/eck-elasticsearch/values.yaml
@@ -121,12 +121,12 @@ remoteClusters: {}
 
 # VolumeClaimDeletePolicy sets the policy for handling deletion of PersistentVolumeClaims for all NodeSets.
 # Possible values are DeleteOnScaledownOnly and DeleteOnScaledownAndClusterDeletion.
-# Defaults to DeleteOnScaledownAndClusterDeletion if unset.
+# By default, if not set or empty, the operator sets DeleteOnScaledownAndClusterDeletion.
 #
 volumeClaimDeletePolicy: ""
 
 # Settings to limit the disruption when pods need to be rescheduled for some reason such as upgrades or routine maintenance.
-# By default when it is unset, the operator sets a default budget that selects all Elastisearch cluster pods and sets `maxUnavailable` to 1.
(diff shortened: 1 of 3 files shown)


deploy/eck-stack/charts/eck-elasticsearch/values.yaml
```diff
@@ -127,15 +127,15 @@ volumeClaimDeletePolicy: ""
 
 # Settings to limit the disruption when pods need to be rescheduled for some reason such as upgrades or routine maintenance.
 # Default budget selects all cluster pods and sets `maxUnavailable` to 1.
-# To disable, set to the empty value (`{}`).
+# To disable, set `disabled` to true.
```
naemono, line 130: I think this now needs to be a little more clear, maybe. 

```
# By default when `podDisruptionBudget` is unset, or set to `null` the operator sets a default pod disruption budget and selects all cluster pods and sets `maxUnavailable` to 1.
# To have the operator set an empty pdb set to the empty value (`{}`).
# To completely disable the pdb set `disabled` to true.
```

naemono (approved):
Possible re-word needed in values, but lgtm.

deploy/eck-stack/charts/eck-elasticsearch/values.yaml
thbkrkr: I see but the important part is that "if unset, the default is X" and it is not so important whether the default is set by Helm or the operator. But let's try to avoid this misunderstanding.

Do you think we should rewrite the others?

https://github.com/elastic/cloud-on-k8s/blob/53e7f25aac53a5a5f4340bf897ac1edfed27239d/deploy/eck-stack/charts/eck-elasticsearch/values.yaml#L124

https://github.com/elastic/cloud-on-k8s/blob/53e7f25aac53a5a5f4340bf897ac1edfed27239d/deploy/eck-stack/charts/eck-elasticsearch/values.yaml#L146

naemono: Well, `revisionHistoryLimit` is a k8s default setting isn't it, not set by default by eck? I'm on the fence on adjusting that wording.

https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#clean-up-policy

And `volumeClaimDeletePolicy` is definitely defaulted by ECK, so I'd support slightly adjusting the wording, sure.

thbkrkr: e25f5d7

PR #6881, opened 2023-06-05
Multi-architecture container images native build

thbkrkr:
A new way to build multi-architecture images, basically with native workers.

WIP: I will create small PR from this one to reduce the surface of change and facilitate reviews.

PR #6927, opened 2023-06-20
Lazy vault client creation when reading vault file

thbkrkr:
This allows reading a Vault file without creating a Vault client when the file already exists, which was the main purpose of the `vault.ReadFile` (reads the file if it exists or else reads the Secret from Vault and writes it to on disk).

A new `ClientProvider` type is introduced with an associated constructor and the signature of the `vault.ReadFile` func is updated to use it:

```diff
# new type
+type ClientProvider func() (Client, error)

# update signature of vault.ReadFile
-func ReadFile(c Client, f SecretFile) ([]byte, error) {
+func ReadFile(clientProvider ClientProvider, f SecretFile) ([]byte, error) {

# new constructor
+func NewClientProvider() func() (Client, error) {
+	var err error
+	var client Client
+	var once sync.Once
+	return func() (Client, error) {
+		once.Do(func() {
+			client, err = NewClient()
+		})
+		return client, err
+	}
+}
```

Motivation for this change: make the operator build (with the `go generate` step) possible in Docker, which opens the door to use the elastic docker-builder buildkite agent. By fetching the public key first, we can `go generate` during the Docker build (in the Dockerfile) without the dependency to Vault, which is a pain to manage in a Dockerfile.

Relates to https://github.com/elastic/cloud-on-k8s/pull/6881.

### DIFF
--- a/pkg/utils/vault/client.go
+++ b/pkg/utils/vault/client.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"sync"
 
 	"github.com/hashicorp/vault/api"
 	"github.com/pkg/errors"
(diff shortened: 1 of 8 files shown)


pkg/utils/vault/client.go
```diff
@@ -23,10 +24,25 @@ const (
 	ghTokenEnvVar = "GITHUB_TOKEN" //nolint:gosec
 )
 
+var once sync.Once
```
pebrc, line 27: with the `sync.Once` at the package level and the variable that holds the client in a closure of the `NewClientProvider` won't you be able to generate a `nil` client if two pieces of code call `NewClientProvider`? So I would expect the `once` also to be in the closure or the `client` variable also be global. Or am I misremembering how the `sync` package works (totally possible)?

thbkrkr: You are completely right! 🤦

pebrc (approved):

PR #6930, opened 2023-06-20
Track in git the default eck config

thbkrkr:
This tracks in git the default config file of the operator `eck.yaml` that is embedded in the Docker images.

This frees the build of the Docker images from the generation of this file that requires `kubectl`, `hack/manifest-gen` and `hack/config-extractor/`.

Relates to #6881.


config/eck.yaml
```diff
@@ -0,0 +1,18 @@
...
+exposed-node-labels: [topology.kubernetes.io/.*,failure-domain.beta.kubernetes.io/.*]
+set-default-security-context: auto-detect
+kube-client-timeout: 60s
+elasticsearch-client-timeout: 180s
+disable-telemetry: false
+distribution-channel: image
+validate-storage-class: true
+enable-webhook: false
+enable-leader-election: true
+elasticsearch-observation-interval: 10s
```
naemono, line 18: nit: missing newline

naemono (approved):
I had to go back and look at the history of why we're handling the configuration file as we are with `manifest-gen | config-extrator`. I was wondering if we're going to run into a situation where these files are out of sync, as right now the config file is generated using helm template `deploy/eck-operator/templates/configmap.yaml`. It appears that the existing `check-local-changes` rule will catch this though, which is great. After saying all of that:

LGTM

PR #6814, opened 2023-05-16
Add script to call Buildkite API to trigger RedHat ECK release

thbkrkr:
This adds a script with the curl command to call the Buildkite API to trigger the pipeline `cloud-on-k8s-operator-redhat-release`that releases ECK for OperatoHub/RedHat. This also enables building branches for this pipeline.

Test:
```
BK_TOKEN=$(jq .graphql_token ~/.buildkite/config.json -r) USERNAME=thb \
  ECK_VERSION=v2.8.0-bc4 DRY_RUN=true ./trigger-bk-release.sh
```
https://buildkite.com/elastic/cloud-on-k8s-operator-redhat-release/builds/29

PR #6934, opened 2023-06-21
Update pipelines provider settings

thbkrkr:
This configures the `nightly`, `redhat-release` and `helm-release` pipelines to not be triggered when code is pushed but triggerable via the API for branches, PRs and tags.

I didn't understand how it works in https://github.com/elastic/cloud-on-k8s/pull/6814#discussion_r1195131276.

The thing is that we have some special code in the tooling that processes the `catalog-info.yml` file that made me think that `trigger_mode: none` was incompatible with `build_branches: true` but they are two different independent things.

```py
# These fields have been set in the past for many pipelines with no
# build trigger. Maintaining them here prevents excessive state
# mutation.
if provider_settings.get("trigger_mode") == "none":
    expanded |= {
        "build_branches": False,
        "build_pull_requests": False,
        "publish_commit_status": False,
        "skip_pull_request_builds_for_existing_commits": False,
    }
```

### DIFF
--- a/catalog-info.yaml
+++ b/catalog-info.yaml
@@ -39,8 +39,9 @@ spec:
     spec:
       repository: elastic/cloud-on-k8s
       provider_settings:
+        trigger_mode: "code"
+        build_pull_requests: true
         build_branches: true
-        build_pull_request_ready_for_review: true
         build_tags: true
         publish_blocked_as_pending: true
         publish_commit_status: true
@@ -74,6 +75,9 @@ spec:
       repository: elastic/cloud-on-k8s
       provider_settings:
         trigger_mode: none
+        build_pull_requests: true
+        build_branches: true
+        build_tags: true
       schedules:
         Every day:
           branch: main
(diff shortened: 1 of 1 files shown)


catalog-info.yaml
```diff
@@ -39,8 +39,9 @@ spec:
     spec:
       repository: elastic/cloud-on-k8s
       provider_settings:
+        trigger_mode: "code"
+        build_pull_requests: true
```
thbkrkr, line 43: Already set by default but let's be explicit.

catalog-info.yaml
```diff
@@ -39,8 +39,9 @@ spec:
...
         build_branches: true
-        build_pull_request_ready_for_review: true
```
thbkrkr, line 43: This consists of rebuilding pull requests when they are ready for review. I don't think we need it.

catalog-info.yaml
```diff
@@ -74,6 +75,9 @@ spec:
       repository: elastic/cloud-on-k8s
       provider_settings:
         trigger_mode: none
+        build_pull_requests: true
+        build_branches: true
+        build_tags: true
```
thbkrkr, line 80: Let's also make it possible to trigger the nightly pipeline via the API for anything.

pebrc (approved):
LGTM

PR #6936, opened 2023-06-21
Helm release pipeline

thbkrkr:
This fixes a bug where `.buildkite/pipeline-release-helm.yml` has no bk agent and prepares the last bits to release Helm charts. 

- Declare an agent in `.buildkite/pipeline-release-helm.yml` because it is now run without `.buildkite/pipeline.yml` that was setting the agent
- Remove the `DRY_RUN` env set at the pipeline level to make it overridable at the build level (REST API) (defaults to true if not set in the Helm releaser tool).
- Add a script to trigger the Helm charts release by creating a build via the Buildkite API


pebrc (approved):
Lgtm

PR #6937, opened 2023-06-21
[helm/eck-elasticsearch] Support disabling podDisruptionBudget #6926

thbkrkr:
Backport the following commit to `2.8`:
- #6926


pebrc (approved):
LGTM is this a breaking change compared to current behaviour?

thbkrkr:
> is this a breaking change compared to current behaviour?

No, if you set `podDisruptionBudget: {}` in the `values.yaml` file, the behavior does not change. No `podDisruptionBudget `is defined in the ES resource and therefore the operator defines a default PDB.

PR #6938, opened 2023-06-21
Validating policy ID only when running in fleet mode for Elastic Agent

### LINKED ISSUE #6903: Agent: "spec.PolicyID is empty" warning is raised despite Agent is running in standalone mode
The Agent webhook attempts to check if `PolicyID` is set whether or not the Agent is running in `fleet` or `standalone` mode:

```golang
func (a *Agent) GetWarnings() []string {
	if a == nil {
		return nil
	}
	if len(a.Spec.PolicyID) == 0 {
		return []string{fmt.Sprintf("%s %s/%s: %s", Kind, a.Namespace, a.Name, MissingPolicyIDMessage)}
	}
	return nil
}
```

This check should only be done in `fleet` mode.

kvalliyurnatt:
The Agent webhook attempts to check if PolicyID is set whether or not the Agent is running in fleet or standalone mode, This check should only be done in fleet mode. Fixes #6903

### DIFF
--- a/pkg/apis/agent/v1alpha1/webhook.go
+++ b/pkg/apis/agent/v1alpha1/webhook.go
@@ -38,7 +38,7 @@ func (a *Agent) GetWarnings() []string {
 	if a == nil {
 		return nil
 	}
-	if len(a.Spec.PolicyID) == 0 {
+	if a.Spec.Mode == AgentFleetMode && len(a.Spec.PolicyID) == 0 {
 		return []string{fmt.Sprintf("%s %s/%s: %s", Kind, a.Namespace, a.Name, MissingPolicyIDMessage)}
 	}
 	return nil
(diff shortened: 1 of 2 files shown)


pkg/apis/agent/v1alpha1/webhook.go
```diff
@@ -38,7 +38,7 @@ func (a *Agent) GetWarnings() []string {
 	if a == nil {
 		return nil
 	}
-	if len(a.Spec.PolicyID) == 0 {
+	if a.Spec.Mode == AgentFleetMode && len(a.Spec.PolicyID) == 0 {
```
naemono, line 41: I think this change makes perfect sense, but I also think it's a perfect opportunity for a set of unit tests around this function.

kvalliyurnatt: There are existing unit tests, will fix those and add a new one

naemono: I was only looking for a `webhooks_test.go` file and didn't see it. My apologies.

barkbay (approved):
LGTM

naemono (approved):
also lgtm now

PR #6944, opened 2023-06-22
Bump versions for eck-stack 0.6.0 release

thbkrkr:


barkbay (approved):
I think it LGTM, a bit hard to review tbh 😄 
I'm wondering if it would make sense to "align" all these versions? (can be discussed async later, let's merge this one for now)

Issue #6950, opened 2023-06-22
make integration-tests also runs the unit tests


thbkrkr:
When `go test -tags='integration' ./pkg/... ./cmd/...` is run, it also runs all the unit-tests. This is caused by the way go build constraints work and the fact that there is no build constraint for unit-tests

As per the golang doc https://pkg.go.dev/cmd/go#hdr-Build_constraints:

> A build constraint, also known as a build tag, is a condition under which a file should be included in the package.

So if you we want to run the tests only for the build tags `integration` then we need to provide a different tag for the other tests.

I see a few options to fix this:

- a) stop to run separately unit-tests and integration-tests: the downside is there is an interest to keep the pr build fast by running the two in parallel
- b) add a build constraint for all unit-tests (337 files `*_test.go`)
- c) rename all integ tests with the same prefix `TestInteg_` (12 tests)
- d) run `go test` only for packages that contain `go:build integration`

```sh
> grep 'go:build integration' -rl | grep _test.go | xargs -n1 dirname | uniq
./cmd/license-initializer
./pkg/utils/fs
./pkg/controller/license/trial
./pkg/controller/license
./pkg/controller/common/watches
./pkg/controller/common/certificates
```

a) too bad not to use this optimization, there is more than a minute to win
b) fastidious and we'll have to think about this for any new `_test.go` file
c) we'll have to think about this for any new integ test
d) still run unit tests in the packages where there are integ tests and kinda hacky in the sense that we're not using the build constraint for its original purpose but it does the job
c) and d) are not exclusive and can be combined

thbkrkr:
Closed by https://github.com/elastic/cloud-on-k8s/pull/6951.

PR #6951, opened 2023-06-23
Run only integration tests

thbkrkr:
This runs integration tests only in packages where there are integration tests, to not run all unit tests along with integration tests.

The duration of the step `integration-tests` which is the longest before running the e2e tests is reduced from `4m` to `2m`.

Relates to #6950.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -171,14 +171,18 @@ helm-test:
 	@hack/helm/test.sh
 
 integration: GO_TAGS += integration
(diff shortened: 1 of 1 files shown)


thbkrkr:
Before:
Capture d’écran 2023-06-23 à 14 11 50

After:
Capture d’écran 2023-06-23 à 14 12 41

pebrc (approved):

PR #6952, opened 2023-06-23
Run TestVersionUpgradeOrdering with/without Logstash

### LINKED ISSUE #6928: TestVersionUpgradeOrdering is always skipped since Logstash was added
`TestVersionUpgradeOrdering` is not run since it has been updated to test an upgrade of Logstash.

It is because the initial version used for the upgrade is too low to test Logstash.

https://github.com/elastic/cloud-on-k8s/blob/8e3beaf80aae6dec61e64371778b1d082f0dddc4/test/e2e/stack_test.go#L38-L40

The `SkipTest()` func of the Logstash builder returns `true` when the version is less than `8.6.0`, which is the case here as the LatestReleasedVersion7x is `7.17.8`.

https://github.com/elastic/cloud-on-k8s/blob/8e3beaf80aae6dec61e64371778b1d082f0dddc4/test/e2e/test/logstash/builder.go#L228-L232

https://github.com/elastic/cloud-on-k8s/blob/8e3beaf80aae6dec61e64371778b1d082f0dddc4/pkg/controller/common/version/version.go#L36

thbkrkr:
This fixes an issue where `TestVersionUpgradeOrdering` is not run since it has been updated
to test an upgrade with Logstash because the upgrade is from `7x` to `8x` and Logstash is 
supported by ECK from `8.6.0` by:
- removing Logstash from `TestVersionUpgradeOrdering`
- adding a new test `TestVersionUpgradeOrderingWithLogstash` that tests an upgrade from `8.6.0` to `8x`

This also:
- adds a call to `association.AllowVersion` at the start of the logstash reconciliation to prevent upgrading Logstash until Elasticsearch is upgraded
- configures an associated Elasticsearch to Logstash otherwise there is no reason not to wait to upgrade Logstash
- updates the `StackResourceVersions` methods that ensure a strict ordering is respected during the version upgrade

Resolves #6928.


thbkrkr:
buildkite test this -f p=gke,t=TestVersionUpgradeOrdering

thbkrkr:
buildkite test this -f p=gke,t=TestVersionUpgradeOrder

pebrc (approved):
LGTM. Should we also update the `IsValid` function to check for Logstash if it is defined?

thbkrkr:
> Should we also update the IsValid function to check for Logstash if it is defined?

Yes. I also added the `AllowVersion` check in the Logstash controller as I think this are the missing bits to ensure ES and Logstash are upgraded in the right order.

thbkrkr:
buildkite test this -f p=gke,t=TestVersionUpgradeOrder

thbkrkr:
I found the bug: Logstash has not associated ES, so there is no reason not to wait to upgrade Logstash.

pebrc:
> I found the bug: Logstash has not associated ES, so there is no reason not to wait to upgrade Logstash.

Ah 💥 probably worth a comment in that method!

test/e2e/stack_test.go
```diff
@@ -29,65 +29,107 @@ import (
...
+
+	ls := logstash.NewBuilder("ls").WithVersion(initialVersion).
+		WithElasticsearchRefs(
+			// associate logstash to the es ref stored in the stackVersions
+			logstashv1alpha1.ElasticsearchCluster{
+				ObjectSelector: commonv1.ObjectSelector{
+					Namespace: stackVersions.Elasticsearch.ref.Namespace,
+					Name:      stackVersions.Elasticsearch.ref.Name,
+				},
+				ClusterName: "es",
```
thbkrkr, line 111: @pebrc Would you say more than that?

thbkrkr:
buildkite test this -f p=gke,t=TestVersionUpgradeOrder

test/e2e/stack_test.go
pebrc: That's even better than what I had in mind. I thought we just document that we don't have an association. But you are right why not just create the association.

pebrc (approved):
LGTM

Issue #6946, opened 2023-06-22
Filebeat/autodiscover ES_data_should_pass_validations fails in 8.9.0-SNAPSHOT


thbkrkr:
[cloud-on-k8s-operator-nightly/builds/217@stack-8.9.0-SNAPSHOT](https://buildkite.com/elastic/cloud-on-k8s-operator-nightly/builds/217#0188dfec-e18a-4ab9-b938-04ce543b4953)

- TestFilebeatDefaultConfig/ES_data_should_pass_validations
- TestBeatSecureSettings/ES_data_should_pass_validations 
- TestBeatConfigRef/ES_data_should_pass_validations

```go
--- FAIL: TestFilebeatDefaultConfig (1887.47s)
    --- FAIL: TestFilebeatDefaultConfig/ES_data_should_pass_validations (1800.00s)
                Error Trace:    /go/src/github.com/elastic/cloud-on-k8s/test/e2e/test/utils.go:94
                Error:          Received unexpected error:
                                hit count should be more than 0 for /*beat*/_search?q=agent.type:filebeat

--- FAIL: TestBeatSecureSettings (1888.35s)
    --- FAIL: TestBeatSecureSettings/ES_data_should_pass_validations (1800.00s)
                Error Trace:    /go/src/github.com/elastic/cloud-on-k8s/test/e2e/test/utils.go:94
                Error:          Received unexpected error:
                                hit count should be more than 0 for /*beat*/_search?q=agent.type:filebeat

--- FAIL: TestBeatConfigRef (1896.88s)
    --- FAIL: TestBeatConfigRef/ES_data_should_pass_validations (1800.00s)
                Error Trace:    /go/src/github.com/elastic/cloud-on-k8s/test/e2e/test/utils.go:94
                Error:          Received unexpected error:
                                hit count should be more than 0 for /*beat*/_search?q=agent.type:filebeat
```

thbkrkr:
Running `TestFilebeatDefaultConfig` locally I see missing RBAC permissions to use Filebeat with autodiscover, not sure if this is the root cause:

```
E0627 12:15:14.396909       7 reflector.go:138] pkg/mod/k8s.io/client-go@v0.23.4/tools/cache/reflector.go:167: 
  Failed to watch *v1.ReplicaSet: failed to list *v1.ReplicaSet: replicasets.apps is forbidden: 
[...]

PR #6958, opened 2023-06-27
Fix RBAC to run e2e tests with Filebeat autodiscover in 8.9.0

thbkrkr:
Looks like Filebeat autodiscover `8.9.0` requires more permissions.

I will check why we need this (hence not closing #6946) but in the meantime, we need this to make our e2e tests build green again.

Relates to #6946.

### DIFF
--- a/config/e2e/roles.yaml
+++ b/config/e2e/roles.yaml
@@ -13,6 +13,14 @@ rules:
   - get
   - watch
   - list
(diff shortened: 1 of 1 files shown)


thbkrkr:
buildkite test this -f p=gke,s=8.9.0-SNAPSHOT,E2E_TAGS=beat

naemono (approved):
💯

PR #6879, opened 2023-06-05
Migrate how to publish ECK images on Docker Hub

thbkrkr:
This adds a step in the `pipeline-release.yml` definition to trigger the service pipeline [unified-release-copy-elastic-images-to-dockerhub](https://buildkite.com/elastic/unified-release-copy-elastic-images-to-dockerhub) ([code](https://github.com/elastic/unified-release/blob/4c7404756bd941642892e06415f8df433708ae0c/service/copy-elastic-images-to-dockerhub/copy.sh#L17-L18)) to copy ECK Docker images from Elastic registry to DockerHub, only for final tags (`^v[0-9]+\.[0-9]+\.[0-9]+\$`).

This replaces the `publish-dockerhub` Jenkins job which is deleted here.

PR #6964, opened 2023-06-28
Generate the trigger step to remove v prefix from BUILDKITE_TAG

thbkrkr:
This fixes a bug introduced in https://github.com/elastic/cloud-on-k8s/pull/6879 where we can't just trigger the pipeline that copies the images with
`IMAGES_TAG: "${BUILDKITE_TAG}"` because the tag contains the `v` prefix and ECK images are tagged 
without it.

The fix is to go through a bit of bash to generate the step with the variable set without the `v` using shell parameter expansion.


barkbay (approved):
Not tested but LGTM.

pebrc (approved):

PR #6959, opened 2023-06-27
Multi-arch images build via drivah and serverless-docker-builder

thbkrkr:
This change revisits the build of ECK's container images by making them drivah-able and use the `serverless-docker-builder`.

The main thing is:
```yaml
  - group: build operator
    key: "operator-image-build"
    steps:
      - label: ":buildkite:"
        agents:
          image: "docker.elastic.co/ci-agent-images/serverless-docker-builder:0.0.5"
        commands: make -C /agent generate-docker-images
        env:
          DOCKER_REGISTRY_VAULT_PATH: secret/ci/elastic-cloud-on-k8s/docker-registry-elastic
          PRE_BUILD_SCRIPT: ".buildkite/scripts/build/pre-build-operator.sh"
          DRIVAH_BUILD_PATH: "build"
          RECURSIVE: true

  - group: build e2e-tests
    key: "e2e-tests-image-build"
    steps:
      - label: ":buildkite:"
        agents:
          image: "docker.elastic.co/ci-agent-images/serverless-docker-builder:0.0.5"
        commands: make -C /agent generate-docker-images
        env:
          DOCKER_REGISTRY_VAULT_PATH: secret/ci/elastic-cloud-on-k8s/docker-registry-elastic
          DRIVAH_BUILD_PATH: "test/e2e"
```

The `serverless-docker-builder` agent step generates 3 steps to build the arm and the amd images and push the manifest. It executes `drivah build $DRIVAH_BUILD_PATH --push` which discovers all `Dockerfile` and `drivah.toml*` in the given path and build the associated container images using the buildah engine, in a `k8s` agent for amd and a `vm` agent for arm.
This means that we don't have `make` or `go` when `drivah build`.

In order to be able to use drivah to build eck, we now fetch the license public key first and then we `go-generate` in Docker during the build.

I made that we can still invoke the build of images from the `Makefile` to dev purposes.

```sh
# to install drivah (requires python 3.9+, git, buildah or docker)
pip install git+ssh://git@github.com/elastic/drivah@0.24.0
```

### e2e-tests image

In CI, `drivah.toml.sh` is directly called by `drivah build`.

In dev, you can:
```sh
# build without license public key
GO_TAGS= make drivah-build-e2e

# build with license public key (default) (this uses your current arch (arm if m1) and the docker engine)
make drivah-build-e2e 

# build/push amd/arm from an arm? machine using docker buildx
make docker-push-e2e
```

### operator image(s)

[...]

PR #6970, opened 2023-07-03
Fix tests run dependency

thbkrkr:
I got confused at the last moment in 445aee26 with the dependencies of the step to run e2e tests. We need to depend on the step that creates the cluster, the dependencies on the build steps were already removed.

It was not catch in the PR because we use `kind` for PRs which is a provider that doesn't use a remoteConfig.

Related to #6959.


thbkrkr:
buildkite test this p=gke,t=TestSmoke

thbkrkr:
buildkite test this -f p=gke,t=TestSmoke

pebrc (approved):

PR #6980, opened 2023-07-05
Make 'RUN go generate' specific between prod and dev builds

thbkrkr:
This fixes a regression introduced in #6959, where the `eck` flavor build uses the layer that generates the public key of the `eck-dev` flavor build, using the specific name of the public license key in the `RUN go-generate` instruction.


pebrc (approved):
LGTM (let's backport to 2.9 also)

PR #6981, opened 2023-07-05
Make 'RUN go generate' specific between prod and dev builds (#6980)

thbkrkr:
Backport the following commit to `2.9`:
- #6980


barkbay (approved):
lgtm

PR #6994, opened 2023-07-07
E2E set test storage class for Logstash data volume

pebrc:
Always set the storage class for Logstash to `e2e-default` in the e2e tests. This is a pragmatic deviation from the practice we have for Elasticsearch were we test with the e2e-default storage class in general but test with network attached volumes in samples and recipes to have both local and networked volume test paths. 

Adding something similar for Logstash seemed to complex with the current builders we have.

Without this change Logstash tests fail on EKS where the EBS CSI driver is not installed by default.

### DIFF
--- a/test/e2e/samples_test.go
+++ b/test/e2e/samples_test.go
@@ -116,7 +116,9 @@ func createBuilders(t *testing.T, decoder *helper.YAMLDecoder, sampleFile, testN
 				WithLogsMonitoring(logRefs...).
 				WithRestrictedSecurityContext().
 				WithLabel(run.TestNameLabel, fullTestName).
-				WithPodLabel(run.TestNameLabel, fullTestName)
+				WithPodLabel(run.TestNameLabel, fullTestName).
+				WithTestStorageClass()
+
 		default:
 			return b
 		}
--- a/test/e2e/es/topology_test.go
+++ b/test/e2e/es/topology_test.go
@@ -53,7 +53,7 @@ func TestResourcesRequirements(t *testing.T) {
 				},
 			},
 			VolumeClaimTemplates: []corev1.PersistentVolumeClaim{
-				newPVC("2Gi", elasticsearch.DefaultStorageClass),
+				newPVC("2Gi", test.DefaultStorageClass),
 			},
 		})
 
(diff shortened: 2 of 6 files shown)


thbkrkr:
buildkite test this -f p=eks,t=TestLogstashWith

test/e2e/samples_test.go
```diff
@@ -116,7 +116,9 @@ func createBuilders(t *testing.T, decoder *helper.YAMLDecoder, sampleFile, testN
 				WithLogsMonitoring(logRefs...).
 				WithRestrictedSecurityContext().
 				WithLabel(run.TestNameLabel, fullTestName).
-				WithPodLabel(run.TestNameLabel, fullTestName)
+				WithPodLabel(run.TestNameLabel, fullTestName).
+				WithTestStorageClass()
```
thbkrkr, line 120: We already have the same kind of exception for Elasticsearch on EKS with [#6588](https://github.com/elastic/cloud-on-k8s/pull/6588/files) to tests samples due to #6515.

Shouldn't we make it conditional to only do it on `eks`?

pebrc: So I think the problem is for the Elasticsearch builder we add the `e2e-default` storage class in the `WithNodeSet` builder method which is used in all but the samples and recipes tests IIUC. We don't have an equivalent method for Logstash. That's why I added `WithTestStorageClass` into the constructor and here because the samples sidestep the builder methods and the constructor call is too early to have the right effect. 

What we could do is parameterise `WithTestStorageClass` on the platform so that it only adds the storage class for EKS runs. That would mean all other runs always use network attached storage.

The other option would be to just always run the Logstash tests with `e2e-default` which is what I am proposing here. I think it might be a bit faster overall and requires less conditional logic.

naemono: I have no issues defaulting to the default storage class for logstash here. It can still be overridden conditionally per provider, so it isn't limiting us.

thbkrkr: > What we could do is parameterise WithTestStorageClass on the platform so that it only adds the storage class for EKS runs.
That's what I had in mind, but I'm ok with the current approach.

thbkrkr (approved):
LGTM

Issue #5795, opened 2022-06-17
Elasticsearch_cluster_health_should_not_have_been_red_during_mutation_process (revisited)


pebrc:
Related to https://github.com/elastic/cloud-on-k8s/issues/5040 but now with node shutdown API in use:

https://devops-ci.elastic.co/job/cloud-on-k8s-e2e-tests-gke-k8s-versions/754//testReport

Rolling upgrade from 8.2.0 to 8.2.2

Timeline: 
*  {"Time":**"2022-06-17T07:17:01.717730831Z"**,"Action":"output","Package":"github.com/elastic/cloud-on-k8s/test/e2e/kb","Test":"TestVersionUpgradeToLatest8x/Creating_an_Elasticsearch_cluster_should_succeed","Output":"=== RUN   TestVersionUpgradeToLatest8x/Creating_an_Elasticsearch_cluster_should_succeed\n"}
*  {"Time":**"2022-06-17T07:18:07.786852419Z"**,"Action":"run","Package":"github.com/elastic/cloud-on-k8s/test/e2e/kb","Test":"TestVersionUpgradeToLatest8x/Applying_the_Elasticsearch_mutation_should_succeed"} **this is were the test runner starts the rolling upgrade**
* Jun 17, 2022 @ **07:18:09.556** cluster state updated, source [auto create [ilm-history-5][org.elasticsearch.action.admin.indices.create.AutoCreateAction$TransportAction$CreateIndexTask@4f704a38]]
 -- index [[.ds-ilm-history-5-2022.06.17-000001/VSEhLPyNTpOL8_9xw-YaVQ]]
----shard_id [.ds-ilm-history-5-2022.06.17-000001][0]
--------[.ds-ilm-history-5-2022.06.17-000001][0], node[c4wCBppvQRWsqabiqMd1hg], [P], recovery_source[new shard recovery], s[INITIALIZING], a[id=5av7tzdTRfeS67wh1Qw-Ig], unassigned_info[[reason=INDEX_CREATED], at[2022-06-17T07:18:09.536Z], delayed=false, allocation_status[no_attempt]]
--------[.ds-ilm-history-5-2022.06.17-000001][0], node[null], [R], recovery_source[peer recovery], s[UNASSIGNED], unassigned_info[[reason=REPLICA_ADDED], at[2022-06-17T07:18:09.537Z], delayed=false, allocation_status[no_attempt]] **ILM history is auto-created**
[...]

PR #7000, opened 2023-07-11
Relax mutation health checks for TestAPMServerVersionUpgradeToLatest8x

thbkrkr:
This adds a new boolean `relaxMutationHealthChecks` in the ES builder in order to relax the continuous health checks performed during mutations.

When a new index is created at the same time as the mutation, the shutdown node API currently does not prevent shutting down a node which has a new uninitialized replica, resulting in a cluster with red health status for a few seconds while the node comes back.

Enabled for:
- `TestAPMServerVersionUpgradeToLatest8x`
- `TestAgentVersionUpgradeToLatest8x`
- `kb/TestVersionUpgradeToLatest8x`
- `beat/TestVersionUpgradeToLatest8x`

Relates to https://github.com/elastic/cloud-on-k8s/issues/5795.

### DIFF
--- a/test/e2e/test/elasticsearch/steps_mutation.go
+++ b/test/e2e/test/elasticsearch/steps_mutation.go
@@ -159,7 +159,9 @@ func (b Builder) MutationTestSteps(k *test.K8sClient) test.StepList {
 					for _, f := range continuousHealthChecks.Failures {
 						t.Errorf("Elasticsearch cluster health check failure at %s: %s", f.timestamp, f.err.Error())
 					}
-					assert.Equal(t, b.relaxMutationHealthChecks, continuousHealthChecks.FailureCount)
+					if !b.relaxMutationHealthChecks {
+						assert.Equal(t, 0, continuousHealthChecks.FailureCount)
+					}
 				},
 			},
 			test.Step{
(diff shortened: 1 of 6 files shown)


test/e2e/test/elasticsearch/steps_mutation.go
```diff
@@ -155,10 +155,13 @@ func (b Builder) MutationTestSteps(k *test.K8sClient) test.StepList {
...
 				Test: func(t *testing.T) {
 					continuousHealthChecks.Stop()
+
 					for _, f := range continuousHealthChecks.Failures {
 						t.Errorf("Elasticsearch cluster health check failure at %s: %s", f.timestamp, f.err.Error())
 					}
-					assert.Equal(t, 0, continuousHealthChecks.FailureCount)
+					if !b.relaxMutationHealthChecks {
+						assert.Equal(t, 0, continuousHealthChecks.FailureCount)
+					}
```
naemono, line 164: Primarily just a question. "relaxing" the mutation health checks to me seems like you'd allow a "couple" of failures, but this is allowing 1000s of failures. What are your thoughts on simply adjusting the max failures to be something like 10 or 5 when we're relaxing them? (I'm unsure how many we're seeing in these scenarios that we're trying to prevent.

thbkrkr: I started with a counter, then I wasn't sure what value to use, so I switched to a boolean. But yes, said like that, a counter seems safer.
We often see 9 or 10 failed health checks.

pebrc (approved):

naemono (approved):

thbkrkr:
In the last commits I adjusted the list of tests that uses this "relaxation" by looking at the failures history for `Elasticsearch_cluster_health_should_not_have_been_red_during_mutation_process`.

pebrc (approved):

PR #7001, opened 2023-07-11
[2.8][docs] Update Maps for 8.8.1 indication fix

jsanz:
Related to elastic/kibana#161572

This is a `8.8` documentation change proposal to address an issue in the Elastic Maps Server `8.8.2` that prevents the image from loading. After discussing with the Release Engineering Team we agreed on updating docs since `8.8.3` is not planned and we are effectively leaving the `8.8` branch for the Maps Server with this issue.

I understand `2.8` is the base branch at this repo where this change should be applied so this shows in the documentation for the `8.8` stack.

### DIFF
--- a/docs/orchestrating-elastic-stack-applications/maps.asciidoc
+++ b/docs/orchestrating-elastic-stack-applications/maps.asciidoc
@@ -41,9 +41,18 @@ kind: ElasticMapsServer
 metadata:
   name: quickstart
 spec:
+ifeval::["{version}"=="8.8.2"]
+  version: 8.8.1
+endif::[]
+ifeval::["{version}"!="8.8.2"]
   version: {version}
+endif::[]
   count: 1
 ----
+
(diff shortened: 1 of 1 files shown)


pebrc:
LGTM. Regarding the choice of branch. My suggestion would be to target `main` and add the version labels because I think we want this in the docs for any version starting with 2.8 e.g. a user on 2.9 which is coming soon might still run an older stack version and should see the warning in the docs.

jsanz:
> LGTM. Regarding the choice of branch. My suggestion would be to target `main` and add version labels for 2.8. and 2.9 (I did that already) because I think we want this in the docs for any version starting with 2.8 e.g. a user on 2.9 which is coming soon might still run an older stack version and should see the warning in the docs.

OK makes sense. I'll open a new PR from `main` picking this commit here to avoid history issues.

PR #7009, opened 2023-07-12
Fixes to upgrade-test-harness

pebrc:
Fixes to upgrade-test-harness for the 2.9.0 release. 
1. incorrectly named test files renamed
2. replace or create CRDs to support newly added CRDs
3. add new release 2.9.0


thbkrkr (approved):
I'm not against the exceptions mechanism, but removing Logstash from the tool until we promote it to no longer be a technical preview seems simpler and more logical to me.

Otherwise code LGTM. It would be nice if we had a pipeline to run this tool.

pebrc:
> but removing Logstash from the tool until we promote it to no longer be a technical preview seems simpler and more logical to me.

Agreed, I will do that.

hack/upgrade-test-harness/fixture/fixture.go
```diff
@@ -175,19 +175,17 @@ func replaceManifests(path string) func(ctx *TestContext) error {
...
 		}
 		ctx.AddCleanupFunc(deleteResources(manifests))
-		return ctx.Replace(manifests)
+		return ctx.ReplaceOrCreate(manifests)
 	}
 }
 
-// deleteManifests is a convenience test function that deletes the provided manifests to the cluster.
-func deleteManifests(path string) func(*TestContext) error {
+func deleteResourcesOfKind(path string, kinds ...string) func(ctx *TestContext) error {
```
thbkrkr, line 182: Unused function that could be removed?

hack/upgrade-test-harness/k8s/kubectl.go
```diff
@@ -256,6 +270,12 @@ func (h *Kubectl) Delete(resources *resource.Result, timeout time.Duration) erro
 			return err
 		}
 
+		for _, k := range kinds {
+			if k != info.Object.GetObjectKind().GroupVersionKind().Kind {
+				return nil
+			}
+		}
+
```
thbkrkr, line 278: The change of adding the argument "kinds" in this function is not used. Revert?

PR #7019, opened 2023-07-13
Add to air-gapped doc how to use a mirrored operator image

jeanfabrice:
Add a chapter how to run the ECK operator itself in an air-gapped environment


docs/operating-eck/air-gapped.asciidoc
```diff
@@ -33,6 +33,23 @@ To make use of your mirrored images you can either set the image for each applic
...
+
+Container registry vendors generally offer native solutions to mirror container images internally. As an alternative, you can use a Docker client which has access to both the public `docker.elastic.co` registry and to the private container registry as follows:
+
+[source,sh]
+----
+docker pull docker.elastic.co/eck/eck-operator:{eck_version}
+docker tag docker.elastic.co/eck/eck-operator:{eck_version} my-registry/eck/eck-operator:{eck_version}
+docker push my-registry/eck/eck-operator:{eck_version}
+----
+
```
kunisen, line 52: Last time I remember we went to the consensus that it's better to hide the concrete command for docker because normally people should know how to use container runtime to do things, i.e. pull/save/load/push container images.
I don't mind adding this to the doc because it explains things clearer, but just want to double check with you if it's really that you would like to add, or an accident...?

kunisen:
Thanks @jeanfabrice for raising the doc PR!! 🙇 
I built the docs and found `{eck_version}` is not reflected as the real version, as other sections in ECK docs. 
Maybe you need to tweak the asciidoc and copy from other places to see if it can show the correct version.

---

image

```bash
# you can clone elastic/docs repository and run this command to build the ECK docs and show it in browser
➜  docs git:(master) docbldeck --open
```

jeanfabrice:
Let me check. I naively thought this variable would be available cross doc page. Maybe the current doc page is missing some bits of configuration. Thanks!

docs/operating-eck/air-gapped.asciidoc
jeanfabrice: As I found the PR to be a bit "thin", I added this section for additional informations. However you are right we went to a consensus. Let's remove it

jeanfabrice:
@kunisen According to https://docs.asciidoctor.org/asciidoc/latest/text/monospace/, asciidoc attributes, like `{eck_version}`, should resolve properly when used inside monospace code. 
It doesn't work with our doc builder, and I don't know why. Note that the simple documentation example, `E=mc^2^`, doesn't work either.
I'm asking the doc team.

jeanfabrice:
@kunisen the issue with the attribute resolution is now fixed

kunisen:
Thank you Jean-Fabrice so much for the amazing effort!

docs/operating-eck/air-gapped.asciidoc
```diff
@@ -33,6 +33,14 @@ To make use of your mirrored images you can either set the image for each applic
 
 
 
+[float]
+[id="{p}-use-internal-operator-image"]
+== Use an internal image of the ECK operator
+
+To deploy the ECK operator in an air-gapped environment, you first have to mirror the operator image itself from `docker.elastic.co` to a private container registry, for example `my-registry`.
```
pebrc, line 40:
```suggestion
To deploy the ECK operator in an air-gapped environment, you first have to mirror the operator image itself from `docker.elastic.co` to a private container registry, for example `my.registry`.
```
All the other examples on the page use `my.registry`

docs/operating-eck/air-gapped.asciidoc
```diff
@@ -33,6 +33,14 @@ To make use of your mirrored images you can either set the image for each applic
 
 
 
+[float]
+[id="{p}-use-internal-operator-image"]
+== Use an internal image of the ECK operator
```
pebrc, line 38: Should call it a `mirrored` image instead, not sure. Internal sounds ambiguous.

pebrc (commented):
Looks good thanks for raising this.

docs/operating-eck/air-gapped.asciidoc
jeanfabrice: 👍

thbkrkr (approved):

PR #7025, opened 2023-07-17
Remove global bk agent declaration

thbkrkr:
This removes the global buildkite agent declaration and therefore declares an agent for each step, due to a bug with the docker-builder that just surfaced since the AWS provider accepts `image` as well as `imagePrefix`, which makes global/local merge behavior less predictable.


pebrc (approved):
Shall we merge this and back port as a temp fix? Until we have new version of the docker-builder?

Issue #7032, opened 2023-07-19
(Release) Build creates invalid install manifest


pebrc:
Build for rc5 as an example https://buildkite.com/elastic/cloud-on-k8s-operator/builds/4899#01894938-f453-4a20-85fc-e117253284a4

Evidence https://download.elastic.co/downloads/eck/2.9.0-bc5/operator.yaml

```
 serviceAccountName: eck-operator-root
 - image: "docker.elastic.co/eck-dev/eck-operator-root:2.9.0-3e40d5b9"
 ```

when it should be: 

```
 serviceAccountName: eck-operator
 - image: "docker.elastic.co/eck/eck-operator:2.9.0-bc5"
```

So it looks like the `OPERATOR_NAME`, `REGISTRY_NAMESPACE` and `IMAGE_TAG` environment variables are not set correctly when generating the manifest. `OPERATOR_NAME` takes `IMAGE_SUFFIX` as an input which needs to be setup https://github.com/elastic/cloud-on-k8s/blob/dc79edd0b2f1ef64742ea8269164773c864bd605/Makefile#L44

cc @thbkrkr

thbkrkr:
Closed by https://github.com/elastic/cloud-on-k8s/pull/7034.

PR #7034, opened 2023-07-19
Fix manifest image name

thbkrkr:
This correctly sets the operator image via `IMAGE_NAME` and `IMAGE_TAG` before generating the k8s manifests
and refactors the use of `OPERATOR_NAME` to make it independent from `IMAGE_SUFFIX`.

Relates to #7032.

### DIFF
--- a/.buildkite/scripts/release/k8s-manifests.sh
+++ b/.buildkite/scripts/release/k8s-manifests.sh
@@ -14,14 +14,21 @@ set -eu
 
 ROOT="$(cd "$(dirname "$0")"; pwd)/../../.."
 
+source "$ROOT/.buildkite/scripts/common/trigger.sh"
+source "$ROOT/.buildkite/scripts/common/operator-image.sh"
+
 retry() { "$ROOT/hack/retry.sh" 5 "$@"; }
 
-get_image_tag() {
-  buildkite-agent meta-data get operator-image --default "" | cut -d':' -f2
-}
-
 main() {
-  local version=${BUILDKITE_TAG:-$(get_image_tag)}
+  echo "--- generate-manifests"
+
+  # set IMAGE_NAME and IMAGE_TAG before generating manifests
+  operator::set_image_vars "${TRIGGER:-$(trigger::set_from_env)}"
+  make generate-manifests
+
+  echo "--- upload-manifests"
+
+  local version=${BUILDKITE_TAG:-$IMAGE_TAG}
   version=${version#v} # remove v prefix
 
   if [[ "$version" == "" ]]; then
--- a/Makefile
+++ b/Makefile
@@ -45,7 +45,7 @@ IMAGE_SUFFIX        ?= -$(subst _,,$(shell whoami))
 REGISTRY_NAMESPACE  ?= eck-dev
 OPERATOR_NAME       ?= eck-operator$(IMAGE_SUFFIX)
 
-export IMAGE_NAME   := $(REGISTRY)/$(REGISTRY_NAMESPACE)/$(OPERATOR_NAME)
+export IMAGE_NAME   ?= $(REGISTRY)/$(REGISTRY_NAMESPACE)/$(OPERATOR_NAME)
 export IMAGE_TAG    ?= $(VERSION)-$(SHA1)
 OPERATOR_IMAGE      ?= $(IMAGE_NAME):$(IMAGE_TAG)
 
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -117,7 +117,7 @@ generate-manifests: controller-gen
 	# -- generate  crds manifest
 	@ ./hack/manifest-gen/manifest-gen.sh -c -g > config/crds.yaml
 	# -- generate  operator manifest
-	@ ./hack/manifest-gen/manifest-gen.sh -g \
+	./hack/manifest-gen/manifest-gen.sh -g \
```
thbkrkr, line 120: Let's see the command here.

.buildkite/scripts/release/k8s-manifests.sh
```diff
@@ -14,14 +14,21 @@ set -eu
...
-  buildkite-agent meta-data get operator-image --default "" | cut -d':' -f2
-}
+retry() { "$ROOT/hack/retry.sh" 5 "$@"; }
 
 main() {
-  local version=${BUILDKITE_TAG:-$(get_image_tag)}
+  echo "--- generate-manifests"
+
+  # set IMAGE_NAME and IMAGE_TAG before generating manifests
+  operator::set_image_vars "${TRIGGER:-$(trigger::set_from_env)}"
```
thbkrkr, line 26: I think it's better to get the image names and tags using the two scripts than reading the operator image from meta-data.

pebrc, line 26: Where do we setup `OPERATOR_NAME` ?

pebrc: The problem I see is that `OPERATOR_NAME` is not only used in `IMAGE_NAME` but also influences the naming of the stateful set and the service accounts etc. Unless I am missing something the `OPERATOR_NAME` will be `eck-operator-root` now for a release build which is not what we want. It should be `elastic-operator` to be consistent with previous releases and also to avoid creating garbage on upgrades.

thbkrkr: You're right. The change was introduced in https://github.com/elastic/cloud-on-k8s/pull/6959.

Before #6959, we set `IMAGE_SUFFIX`="" via [`.buildkite/scripts/build/setenv.sh`](https://github.com/elastic/cloud-on-k8s/blob/696367a2f7106fe75be3d450639fa60154121ec6/.buildkite/scripts/build/setenv.sh). With #6959, I simplified the number of variables to set by only setting `IMAGE_NAME` and `IMAGE_TAG` via [`.buildkite/scripts/common/operator-image.sh`](https://github.com/elastic/cloud-on-k8s/blob/445aee26d9f6ba2371afe387c026be83faf561c5/.buildkite/scripts/common/operator-image.sh).

I didn't see that the intermediate OPERATOR_NAME is also used to generate the manifests.

I noticed when you started this exact thread that we try to set it again:
https://github.com/elastic/cloud-on-k8s/blob/696367a2f7106fe75be3d450639fa60154121ec6/Makefile#L87-L88

thbkrkr: I propose to keep only the last declaration and remove the first that uses the `IMAGE_SUFFIX`.

```diff
--- a/Makefile
+++ b/Makefile
@@ -43,9 +43,8 @@ export ARCH         ?= $(shell uname -m | sed -e "s|x86_|amd|" -e "s|aarch|arm|"
 # for dev, suffix image name with current user name
 IMAGE_SUFFIX        ?= -$(subst _,,$(shell whoami))
 REGISTRY_NAMESPACE  ?= eck-dev
-OPERATOR_NAME       ?= eck-operator$(IMAGE_SUFFIX)

-export IMAGE_NAME   ?= $(REGISTRY)/$(REGISTRY_NAMESPACE)/$(OPERATOR_NAME)
+export IMAGE_NAME   ?= $(REGISTRY)/$(REGISTRY_NAMESPACE)/eck-operator$(IMAGE_SUFFIX)
```

thbkrkr: From the last [build](https://buildkite.com/elastic/cloud-on-k8s-operator/builds/5013#01896e4c-4bd7-444c-ad26-b3b7dc7cc70b):
```
# -- generate  operator manifest
--
  | ./hack/manifest-gen/manifest-gen.sh -g \
  | --namespace=elastic-system \
  | --profile=global \
  | --set=installCRDs=false \
  | --set=telemetry.distributionChannel=all-in-one \
  | --set=image.tag=7034-2c25550a \
  | --set=image.repository=docker.elastic.co/eck-ci/eck-operator-pr \
  | --set=nameOverride=elastic-operator \
  | --set=fullnameOverride=elastic-operator > config/operator.yaml
```

pebrc (approved):

pebrc:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|2.9|[](https://github.com/elastic/cloud-on-k8s/pull/7038)|

### Questions ?

PR #7040, opened 2023-07-19
tidy helm releaser

pebrc:
It seems there is a difference between the one I raised for 2.9


thbkrkr:
This is just pruning useless entry, it is not really required.

thbkrkr (approved):

PR #7060, opened 2023-07-26
Make red-hat job not dependent on tag version of build

pebrc:
Take the version of the release branch `VERSION` instead of the `BUILDKITE_TAG` variable to allow releasing the OLM bits with fixes that came after the release version was tagged.

### DIFF
--- a/.buildkite/scripts/release/redhat-preflight.sh
+++ b/.buildkite/scripts/release/redhat-preflight.sh
@@ -9,6 +9,8 @@
 
 set -euo pipefail
 
+ROOT="$(cd "$(dirname "$0")"; pwd)/../../.."
+
 VAULT_ROOT_PATH=${VAULT_ROOT_PATH:-secret/ci/elastic-cloud-on-k8s}
 
 tmpDir=$(mktemp -d)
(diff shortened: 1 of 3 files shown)


.buildkite/scripts/release/redhat-preflight.sh
```diff
@@ -20,7 +22,8 @@ container_already_verified() {
 }
 
 main() {
-    local tag="${1#v}"
+    local tag
+    tag=$(cat "$ROOT"/VERSION)
```
thbkrkr, line 26: It's not you, but reading this code again, I don't like how I wrote it.
Your change will work, but the code suggests that `tag` is local but it still used by the other function `container_already_verified`.

Maybe remove the local work and use uppercase to suggest it's global? Rename it to TAG or ECK_VERSION?
```bash
    ECK_VERSION=$(cat "$ROOT/VERSION")
    export ECK_VERSION
```

Or don't use export but declare the vars outside main:

```
API_KEY=
PROJECT_ID=
ECK_VERSION=
```

thbkrkr (approved):
LGTM

Left a comment to improve the script a bit. I'll let you consider if it's worth it.

pebrc:
@thbkrkr I applied your suggestion

pebrc:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|2.9|[](https://github.com/elastic/cloud-on-k8s/pull/7062)|

### Questions ?

PR #7120, opened 2023-08-25
Fix autopilot e2e test

pebrc:
Use standard k8s labels. 
Add a name label that is not suffixed by a random string during e2e testing. 

Fixes failing autopilot e2e tests that fail to deploy the DaemonSet due to the kyverno policies that were targeting a fixed label while the DaemonSet was using randomised labels.

Alternative implementation: 
* do not use the random suffix on the DaemonSet to begin with (I am open to that as well)


barkbay (approved):
LGTM

PR #7156, opened 2023-09-06
feat(elasticsearch): Add more info to status

smcavallo:
Signed-off-by: smcavallo <smcavallo@hotmail.com>

Allow more info about the Elasticsearch cluster to be added to status field.

Changes the struct from `ElasticsearchHealth` which is a string to an object called `ElasticsearchState` which can contain whatever other information about the cluster that can be later used to populate status field.

In this case - the goal is to be able to expose `PercentDataUsed` for the cluster.  This is a key performance indicator for the cluster.
The eventual goal is to be able to enable autoscaling for data nodes.  Once the cluster approaches the high water mark for disk, it would be great if the operator could automatically scale up and add an additional data node.

Submitting this as WIP PR in order to get feedback.
Currently working through a circular dependency issue with imports.

### DIFF
--- a/pkg/apis/elasticsearch/v1/status.go
+++ b/pkg/apis/elasticsearch/v1/status.go
@@ -6,12 +6,20 @@ package v1
 
 import (
 	"fmt"
+	"strconv"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	commonv1 "github.com/elastic/cloud-on-k8s/v2/pkg/apis/common/v1"
 	"github.com/elastic/cloud-on-k8s/v2/pkg/apis/common/v1alpha1"
+	"github.com/elastic/cloud-on-k8s/v2/pkg/controller/elasticsearch/client"
 )
+
+// ElasticsearchState contains info from the cluster API which can be used to update status
+type ElasticsearchState struct {
+	ElasticsearchHealth ElasticsearchHealth
+	NodesStats          client.NodesStats
+}
 
 // ElasticsearchHealth is the health of the cluster as returned by the health API.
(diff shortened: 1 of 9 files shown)


barkbay:
> In this case - the goal is to be able to expose PercentDataUsed for the cluster.

I'm not sure we want to use the K8S API as a metric server. If we think some metrics are important then they should be collected in a relevant observability system. Using the status sub-resource to store metrics can also result in a lot of update calls which can either add unnecessary load to the K8S API server or cause throttling of the K8S client.

smcavallo:
@barkbay - that makes sense.  If I revise this PR to change ElasticsearchHealth from a string to a struct will that be acceptable?  It at least opens the door to additional metrics in the future and should result in net-zero change to the operator.

barkbay:
>  If I revise this PR to change ElasticsearchHealth from a string to a struct as the only change will that be acceptable? 

I'm not sure I understand what all the implications would be, and this would favor the idea that the k8s api can be used to store any metric/health information which, I think, is a not a good idea.

thbkrkr:
> The eventual goal is to be able to enable autoscaling for data nodes. Once the cluster approaches the high water mark for disk, it would be great if the operator could automatically scale up and add an additional data node.

This is already supported, check https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-autoscaling.html.

Given this and all the implications highlighted in previous comments, I am closing sorry. We must be vigilant about the developments we make to the operator, because it is today used in many critical systems. 

Please open an issue to explain in more detail the problem you are trying to solve and why the solution you are proposing is the right approach to achieve it, in order to start the discussion.

Thank you for your understanding and your interest in the project.

PR #7081, opened 2023-08-04
Add missing RBAC rules for *Beats as of 8.9.0

pebrc:
Fixes https://github.com/elastic/cloud-on-k8s/issues/7079

This was first reported in https://github.com/elastic/cloud-on-k8s/issues/6946#issuecomment-1620493601

I am unsure if we should add a warning to the docs. But I think we should back port this.

PR #7161, opened 2023-09-08
Update beat/agent doc with missing RBAC rules required from 8.9.0

thbkrkr:
Similar to:
- #7081

for the documentation.

Relates to:
- #6946


naemono (approved):
matches https://github.com/elastic/cloud-on-k8s/pull/7081/files
👍

PR #7184, opened 2023-09-26
Try to stabilise tests by checking for cluster health early

pebrc:
Attempt to fix https://github.com/elastic/cloud-on-k8s/issues/7172

PR #7189, opened 2023-09-28
Exclude naming test from health check

pebrc:
More fallout from https://github.com/elastic/cloud-on-k8s/pull/7184

I wonder if we should reverse the logic an only do the health check for specific tests that involve Beats and Agent?

Also I don't think the fix in #7184 has actually helped.

### DIFF
--- a/test/e2e/naming_test.go
+++ b/test/e2e/naming_test.go
@@ -59,7 +59,7 @@ func testLongestPossibleName(t *testing.T) {
 			Count: 1,
 		}).
 		WithRestrictedSecurityContext().
-		// TODO understand why this is necessary, test passes most of the time but sometimes is takes more than 15 minutes for Kibana to become avaialable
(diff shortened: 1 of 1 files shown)


pebrc:
buildkite test this -f p=gke,t=TestNameValidation

test/e2e/naming_test.go
```diff
@@ -58,7 +58,10 @@ func testLongestPossibleName(t *testing.T) {
 			Name:  nodeSpecName,
 			Count: 1,
 		}).
-		WithRestrictedSecurityContext()
+		WithRestrictedSecurityContext().
+		// TODO understand why this is necessary, test passes most of the time but sometimes is takes more than 15 minutes for Kibana to become avaialable
```
thbkrkr, line 62:
```suggestion
		// TODO understand why this is necessary, test passes most of the time but sometimes is takes more than 15 minutes for Kibana to become available
```

pebrc:
buildkite test this -f p=gke,t=TestFleet

PR #7200, opened 2023-10-05
Move soft_fail directive

thbkrkr:
This fixes where to define the `soft_fail` directive introduced in #7197.

PR #7212, opened 2023-10-10
soft_fail on any status code on k8s cluster deletion

thbkrkr:
I wanted to be very selective on the errors that we ignore and didn't look at carefully that the exit code was `2`.

Note that:
> By design, the exit status is two if `make` encounters any errors.
https://www.gnu.org/software/make/manual/html_node/Running.html 

You need to look at failure from the top, not the bottom:

Capture d’écran 2023-10-10 à 11 00 50

After this failed attempt to be selective, I propose to just soft fail in case of any error.

Relates to #7197, #7200.


thbkrkr:
buildkite test this -f p=gke,t=TestSmoke

barkbay (approved):
LGTM

PR #7240, opened 2023-10-16
Update versions for the 2.10 release

rhr323:
Updating versions as perparation for the new release.


hack/operatorhub/config.yaml
```diff
@@ -1,4 +1,4 @@
-newVersion: 2.10.0-SNAPSHOT
+newVersion: 2.10.0
 prevVersion: 2.9.0
 stackVersion: 8.10.2-SNAPSHOT
```
barkbay, line 3: I think we want the next stack release to be referenced here.

```suggestion
stackVersion: 8.11.0
```

thbkrkr (commented):
Looks good, just need to merge `2.10` in the PR branch to fix the conflict and get c5a17b4dc047, then make sure to remove `-SNAPSHOT` from: https://github.com/elastic/cloud-on-k8s/blob/c5a17b4dc047351c7e7f0363b6f4c609f9980510/deploy/eck-stack/charts/eck-logstash/values.yaml#L21

thbkrkr (approved):

hack/operatorhub/config.yaml
```diff
@@ -41,7 +41,7 @@ packages:
   - outputPath: certified-operators
     packageName: elasticsearch-eck-operator-certified
     distributionChannel: certified-operators
-    minSupportedOpenshiftVersion: v4.9
+    minSupportedOpenshiftVersion: v4.11
```
thbkrkr, line 44: Nit: should be in https://github.com/elastic/cloud-on-k8s/pull/7244 instead.

PR #7241, opened 2023-10-16
Version bump for next release

rhr323:
Updating ECK versions to 2.11.0-SNAPSHOT and stack to ~8.11.0-SNAPSHOT~ 8.12.0-SNAPSHOT


rhr323:
There's no 8.12-SNAPSHOT stack, so I guess 8.11 is fine for now?

hack/operatorhub/config.yaml
```diff
@@ -1,6 +1,6 @@
-newVersion: 2.10.0-SNAPSHOT
-prevVersion: 2.9.0
-stackVersion: 8.10.2-SNAPSHOT
+newVersion: 2.11.0-SNAPSHOT
+prevVersion: 2.10.0
+stackVersion: 8.12.0-SNAPSHOT
```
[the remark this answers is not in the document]
rhr323: I ended up merging a separate PR for this:
https://github.com/elastic/cloud-on-k8s/pull/7244/files

hack/upgrade-test-harness/conf.yaml
```diff
@@ -36,5 +36,5 @@ testParams:
     operatorVersion: 2.9.0
     stackVersion: 8.9.0
   - name: upcoming
-    operatorVersion: 2.10.0-SNAPSHOT
-    stackVersion: 8.10.0-SNAPSHOT
+    operatorVersion: 2.11.0-SNAPSHOT
+    stackVersion: 8.11.0-SNAPSHOT
```
thbkrkr, line 40: If we want to also update upgrade-test-harness in this PR, there are more files to update, see for example: https://github.com/elastic/cloud-on-k8s/pull/6865.

naemono: I've added the missing test-harness data.

pebrc (approved):
I would have updated the test harness only after the release. If we, theoretically, discover a bug in the install manifests that we fix in the 2.10 branch it would not be captured here unless someone actively remembers to forward port it to main.

naemono:
> I would have updated the test harness only after the release. If we, theoretically, discover a bug in the install manifests that we fix in the 2.10 branch it would not be captured here unless someone actively remembers to forward port it to main.

I can remove these today. Thanks.

naemono:
Actually my stack updates yesterday were incorrect, as the STACK version is not directly lined up with the ECK version at this time. I'll get these updated...

naemono:
> Actually my stack updates yesterday were incorrect, as the STACK version is not directly lined up with the ECK version at this time. I'll get these updated...

These have been updated.

naemono:
> I can remove these today. Thanks.

We discussed, and these are being left within this PR.

PR #7244, opened 2023-10-17
Update oldest supported Openshift version

rhr323:
According to our [policy](https://github.com/elastic/k8s-dev/issues/238) and [OpenShift lifecycle policy](https://access.redhat.com/support/policy/updates/openshift#dates), oldest supported versions are updated as follows:
- OpenShift: 4.9 -> 4.11

### DIFF
--- a/deploy/eck-operator/Chart.yaml
+++ b/deploy/eck-operator/Chart.yaml
@@ -14,7 +14,7 @@ version: 2.10.0-SNAPSHOT
 
 appVersion: 2.10.0-SNAPSHOT
 
-kubeVersion: ">=1.24.0-0"
+kubeVersion: ">=1.21.0-0"
 
 keywords:
   - Logstash
--- a/deploy/eck-operator/Chart.yaml
+++ b/deploy/eck-operator/Chart.yaml
@@ -14,7 +14,7 @@ version: 2.10.0-SNAPSHOT
 
 appVersion: 2.10.0-SNAPSHOT
 
-kubeVersion: ">=1.21.0-0"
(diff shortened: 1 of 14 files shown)


deploy/eck-operator/Chart.yaml
```diff
@@ -14,7 +14,7 @@ version: 2.10.0-SNAPSHOT
 
 appVersion: 2.10.0-SNAPSHOT
 
-kubeVersion: ">=1.21.0-0"
+kubeVersion: ">=1.24.0-0"
```
pebrc, line 17: Unless there is a strong technical reason to restrict to 1.24 or greater I would leave things in the Helm charts at 1.21. Even if officially unsupported I think we don't want to make users' life harder than necessary. But if there is for example an API we use that is not available in older versions that would be an argument to upgrade.

rhr323: Ok, so the docs should say 1.24+ only, but the helm charts will be more permissive.
  
How about this part? Use 1.21 there, too?
```
  # kubeVersion is the effective Kubernetes version we target when generating the all-in-one.yaml.
-  kubeVersion: 1.21.0
+  kubeVersion: 1.24.0
```

rhr323:
Btw, should this be merged to main, than backported to 2.10?

deploy/eck-operator/Chart.yaml
thbkrkr: Yes, I think we want to leave kubeVersion at 1.21 everywhere in the helm charts for now.

thbkrkr (approved):

PR #7252, opened 2023-10-19
Allow setting additional Operator flags.

### LINKED ISSUE #6091: Operator flags not available through Helm
Some operator flags are not available in the Helm chart via the ConfigMap:
- ca-dir
- config
- disable-config-watch
- ip-family
- operator-namespace
- ubi-only
- webhook-secret

We could add a CI check to ensure we don't forget to update the chart when there is a new operator flag. Not sure it's worth it because we rarely add operator flags and the bash commands below to extract the information for the comparison seem brittle. 

```bash
$ operator_flags() { grep = pkg/controller/common/operator/flags.go | grep -v development | grep -o '".*"' | sed 's/"//g' | sort; }
                                                                                                                                                              
$ config_map() { cat deploy/eck-operator/templates/configmap.yaml | grep -A99 "eck.yaml" | grep -vE "eck.yaml| if | end " | grep -o "[a-z\-]*:" | sed "s/://" | sort; }
                                                                                                                                                                                                          
$ diff <(config_map) <(operator_flags)
2a3
> ca-dir
4a6
> config
5a8
> disable-config-watch
14a18
> ip-family
20a25
> operator-namespace
22a28
> ubi-only
25a32
> webhook-secret
```

```diff
diff --git a/pkg/controller/common/operator/flags.go b/pkg/controller/common/operator/flags.go
index 0b770be75..b3ac75179 100644
--- a/pkg/controller/common/operator/flags.go
+++ b/pkg/controller/common/operator/flags.go
@@ -5,7 +5,7 @@
 package operator
 
 const (
-       AutoPortForwardFlag                  = "auto-port-forward"
+       AutoPortForwardFlag                  = "auto-port-forward" // for development purpose
        CADirFlag                            = "ca-dir"
        CACertRotateBeforeFlag               = "ca-cert-rotate-before"
        CACertValidityFlag                   = "ca-cert-validity"
@@ -13,7 +13,7 @@ const (
        CertValidityFlag                     = "cert-validity"
        ConfigFlag                           = "config"
        ContainerRegistryFlag                = "container-registry"
-       DebugHTTPListenFlag                  = "debug-http-listen"
+       DebugHTTPListenFlag                  = "debug-http-listen" // for development purpose
        DisableConfigWatch                   = "disable-config-watch"
        DisableTelemetryFlag                 = "disable-telemetry"
        DistributionChannelFlag              = "distribution-channel"
@@ -38,4 +38,5 @@ const (
        WebhookCertDirFlag                   = "webhook-cert-dir"
        WebhookNameFlag                      = "webhook-name"
        WebhookSecretFlag                    = "webhook-secret"
+       _                                    = "log-verbosity"
```

naemono:
closes #6091 

## What is this change?

Allows setting the following Operator flags via the Helm chart:
- `-disable-config-watch`
- `-ip-family`
- `-operator-namespace` # is set indirectly via `.Release.Namespace`
- `-ubi-only`
- `-webhook-secret`

## Notes for review
`-operator-namespace` isn't allowed to be set directly, but is set indirectly via whatever namespace the Helm chart is installed within {{ `.Release.Namespace` }}.
`-config` is not allowed to be set at all, as it doesn't make sense in the context of a Helm chart. (It's used in the statefulset template `--config=/conf/eck.yaml`.

### Testing done
```
❯ helm upgrade -i elastic-operator . -n elastic-system --create-namespace --set=config.disableConfigWatch=true --set=config.ipFamily=IPv4 --set=config.ubiOnly=true --set=image.tag=2.9.0

❯ kubectl get cm -n elastic-system elastic-operator -o yaml | yq '.data["eck.yaml"]'
log-verbosity: 0
metrics-port: 0
container-registry: docker.elastic.co
max-concurrent-reconciles: 3
ca-cert-validity: 8760h
ca-cert-rotate-before: 24h
cert-validity: 8760h
cert-rotate-before: 24h
disable-config-watch: true
exposed-node-labels: [topology.kubernetes.io/.*,failure-domain.beta.kubernetes.io/.*]
ip-family: IPv4
set-default-security-context: auto-detect
kube-client-timeout: 60s
elasticsearch-client-timeout: 180s
disable-telemetry: false
distribution-channel: helm
validate-storage-class: true
enable-webhook: true
webhook-name: elastic-operator.elastic-system.k8s.elastic.co
webhook-port: 9443
operator-namespace: elastic-system
enable-leader-election: true
elasticsearch-observation-interval: 10s
ubi-only: true
```

### DIFF
--- a/config/eck.yaml
+++ b/config/eck.yaml
@@ -6,6 +6,7 @@ ca-cert-validity: 8760h
 ca-cert-rotate-before: 24h
 cert-validity: 8760h
 cert-rotate-before: 24h
+disable-config-watch: false
 exposed-node-labels: [topology.kubernetes.io/.*,failure-domain.beta.kubernetes.io/.*]
 set-default-security-context: auto-detect
 kube-client-timeout: 60s
@@ -14,5 +15,7 @@ disable-telemetry: false
 distribution-channel: image
 validate-storage-class: true
 enable-webhook: false
+operator-namespace: elastic-system
 enable-leader-election: true
 elasticsearch-observation-interval: 10s
+ubi-only: false
--- a/deploy/eck-operator/values.yaml
+++ b/deploy/eck-operator/values.yaml
@@ -128,6 +128,8 @@ webhook:
   objectSelector: {}
   # port is the port that the validating webhook binds to.
   port: 9443
+  # secret specifies the Kubernetes secret to be mounted into the path designated by the certsDir value to be used for webhook certificates.
+  secret: ""
 
 # hostNetwork allows a Pod to use the Node network namespace.
 # This is required to allow for communication with the kube API when using some alternate CNIs in conjunction with webhook enabled.
@@ -184,8 +186,14 @@ config:
   # certificatesRotateBefore defines when to rotate a certificate that is due to expire.
   certificatesRotateBefore: 24h
 
+  # disableConfigWatch specifies whether the operator watches the configuration file for changes.
+  disableConfigWatch: false
+
   # exposedNodeLabels is an array of regular expressions of node labels which are allowed to be copied as annotations on Elasticsearch Pods.
   exposedNodeLabels: [ "topology.kubernetes.io/.*", "failure-domain.beta.kubernetes.io/.*" ]
+
+  # ipFamily specifies the IP family to use. Possible values: IPv4, IPv6 and "" (auto-detect)
+  ipFamily: ""
 
   # setDefaultSecurityContext determines whether a default security context is set on application containers created by the operator.
   # *note* that the default option now is "auto-detect" to attempt to set this properly automatically when both running
@@ -210,6 +218,10 @@ config:
 
   # Interval between observations of Elasticsearch health, non-positive values disable asynchronous observation.
   elasticsearchObservationInterval: 10s
+
+  # ubiOnly specifies whether the operator will use only UBI container images to deploy Elastic Stack applications. UBI images are only available from 7.10.0 onward.
+  # Cannot be combined with the containerSuffix value.
+  ubiOnly: false
 
 # Prometheus PodMonitor configuration
 # Reference: https://github.com/prometheus-operator/prometheus-operator/blob/master/Documentation/api.md#podmonitor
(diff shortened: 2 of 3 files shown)


config/eck.yaml
```diff
@@ -14,5 +15,7 @@ disable-telemetry: false
 distribution-channel: image
 validate-storage-class: true
 enable-webhook: false
+operator-namespace: elastic-system
```
pebrc, line 18: Default is empty string?

pebrc (commented):
Otherwise lgtm

config/eck.yaml
naemono: This is coming from `manifest-gen` which defaults to using the 'elastic-system' namespace when generating yaml from the helm charts. 
https://github.com/elastic/cloud-on-k8s/blob/20f11e756989d9811d3f22f638949a81409a4ef6/hack/manifest-gen/main.go#L121
And with the addition of `operator-namespace: {{ .Release.Namespace }}`, this is why it's now defaulting to setting this in the default configuration.

If preferred, I can just remove this from the template, as it really isn't getting us anything.

deploy/eck-operator/values.yaml
```diff
@@ -184,9 +186,15 @@ config:
...
   certificatesRotateBefore: 24h
 
+  # disableConfigWatch specifies whether the operator watches the configuration file for changes.
+  disableConfigWatch: false
+
   # exposedNodeLabels is an array of regular expressions of node labels which are allowed to be copied as annotations on Elasticsearch Pods.
   exposedNodeLabels: [ "topology.kubernetes.io/.*", "failure-domain.beta.kubernetes.io/.*" ]
 
+  # ipFamily specifies the IP family to use. Possible values: IPv4, IPv6 and "" (auto-detect)
+  ipFamily: ""
```
thbkrkr, line 196: The odd little thing is that it looks like we are setting `auto-detect` by default here, which we'll actually do, but not by relying on `ip-family: ""` in the configMap but rather via the default value set in [main.go](https://github.com/elastic/cloud-on-k8s/blob/20f11e756989d9811d3f22f638949a81409a4ef6/cmd/manager/main.go#L260-L263
).
But this is probably better than writing `ip-family:` empty in the configMap (by removing the `with` directive) or leave this commented.

thbkrkr (approved):

deploy/eck-operator/values.yaml
naemono: Yeah, I initially had `ip-family` set without a `with` directive, and changed it upon seeing how it was being represented in the configmap iirc.

naemono:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|2.10|[](https://github.com/elastic/cloud-on-k8s/pull/7259)|

### Questions ?

PR #7269, opened 2023-10-26
Ignore `remote_cluster_client` role when calculating autoscaling status.

naemono:
## What does this change?

This change ignores the `remote_cluster_client` role when calculating autoscaling status as it is likely to exist across multiple nodesets.

## Reasoning for this change

If multiple nodesets use the `remote_cluster_client` role, you will see the following in the autoscaling status:
```
  policies:
  - lastModificationTime: "2023-10-26T16:17:41Z"
    name: hot
...
    state:
    - messages:
      - role remote_cluster_client is declared in autoscaling policies hot,warm
      type: OverlappingPolicies
```

While this does not indicate a failure, or prevent autoscaling from functioning, it's confusing to the user.

Since the `remote_cluster_client` role doesn't control data storage, but only allows the node to act as a cross-cluster client, it's existence shouldn't be a concern for the autoscaling controller.

### Documentation
[Elasticsearch Remote Cluster Settings](https://www.elastic.co/guide/en/elasticsearch/reference/8.10/remote-clusters-settings.html#remote-clusters-settings)
[Kibana Settings](https://www.elastic.co/guide/en/kibana/current/monitoring-settings-kb.html#monitoring-general-settings) that note that to enable CCS, `remote_cluster_client` must exist on all nodes.


pebrc (commented):
I think validations need to be addressed here as well: 
We have a validation that says `ML nodes must be in a dedicated autoscaling policy` so a node role definition of [`ml`, `remote_cluster_client`] would be invalid I think.

naemono:
> I think validations need to be addressed here as well: We have a validation that says `ML nodes must be in a dedicated autoscaling policy` so a node role definition of [`ml`, `remote_cluster_client`] would be invalid I think.

This has been updated, and a unit test has been added @pebrc

pkg/controller/common/autoscaling/validations_test.go
```diff
@@ -0,0 +1,87 @@
...
+			Max: resource.MustParse("2"),
+		},
+		MemoryRange: &v1alpha1.QuantityRange{
+			Min: resource.MustParse("2Gi"),
+			Max: resource.MustParse("4Gi"),
+		},
+	}
+)
+
+func TestValidateAutoscalingPolicies(t *testing.T) {
```
pebrc, line 35: TestValidateElasticsearchAutoscaler was an existing test that could have been used?

pebrc (approved):

pkg/controller/common/autoscaling/validations_test.go
```diff
@@ -0,0 +1,87 @@
...
+								Roles:    []string{"ml", "remote_cluster_client"},
+								Deciders: nil,
+							},
+						},
+						AutoscalingResources: defaultAutoscalingResources,
+					},
+				},
+			},
+			want: nil,
+		},
```
barkbay, line 78: We still want the ml tier to be in a dedicated `nodeSet`, we should expect an error in the following test I think:

```suggestion
		},
		{
			name: "ML policy with roles [ml, remote_cluster_client] succeeds",
			args: args{
				autoscalingSpecPath: func(index int, child string, more ...string) *field.Path {
					return field.NewPath("spec").
						Child("policies").
						Index(index).
						Child(child, more...)
				},
				autoscalingPolicies: v1alpha1.AutoscalingPolicySpecs{
					{
						NamedAutoscalingPolicy: v1alpha1.NamedAutoscalingPolicy{
							Name: "data",
							AutoscalingPolicy: v1alpha1.AutoscalingPolicy{
								Roles:    []string{"data", "remote_cluster_client"},
								Deciders: nil,
							},
						},
						AutoscalingResources: defaultAutoscalingResources,
					},
					{
						NamedAutoscalingPolicy: v1alpha1.NamedAutoscalingPolicy{
							Name: "ml1",
							AutoscalingPolicy: v1alpha1.AutoscalingPolicy{
								Roles:    []string{"ml", "remote_cluster_client"},
								Deciders: nil,
							},
						},
						AutoscalingResources: defaultAutoscalingResources,
					},
					{
						NamedAutoscalingPolicy: v1alpha1.NamedAutoscalingPolicy{
							Name: "ml2",
							AutoscalingPolicy: v1alpha1.AutoscalingPolicy{
								Roles:    []string{"ml"},
								Deciders: nil,
							},
						},
						AutoscalingResources: defaultAutoscalingResources,
					},
				},
			},
			want: /* an error */,
		},
```

barkbay (commented):
I'm actually wondering if we should not ignore the remote cluster client in the whole validation logic. For example, should we not consider 2 node sets with `[data]` and `[data,remote_cluster_client]` a duplicated role set?

```diff
diff --git a/pkg/controller/common/autoscaling/validations.go b/pkg/controller/common/autoscaling/validations.go
index 809c2a02b..19fc23585 100644
--- a/pkg/controller/common/autoscaling/validations.go
+++ b/pkg/controller/common/autoscaling/validations.go
@@ -122,25 +122,32 @@ func ValidateAutoscalingPolicies(
                }
 
                // Validate the set of roles managed by this autoscaling policy.
-               if autoscalingSpec.Roles == nil || len(autoscalingSpec.Roles) == 0 {
-                       errs = append(errs, field.Required(autoscalingSpecPath(i, "roles"), "roles field is mandatory and must not be empty"))
+               // We remove the remote cluster role from the set of roles since it can be applied to some of the nodesets.
+               effectiveRoles := set.Make(autoscalingSpec.Roles...).Diff(set.Make(string(esv1.RemoteClusterClientRole))).AsSlice()
+
+               if len(effectiveRoles) == 0 {
+                       errs = append(
+                               errs,
+                               field.Required(autoscalingSpecPath(i, "roles"),
+                                       fmt.Sprintf("roles field is mandatory and must not be empty (ignoring %s)", esv1.RemoteClusterClientRole)),
+                       )
                } else {
-                       if containsStringSlice(rolesSet, autoscalingSpec.Roles) {
+                       if containsStringSlice(rolesSet, effectiveRoles) {
                                // A set of roles must be unique across all the autoscaling policies.
                                errs = append(
                                        errs,
                                        field.Invalid(
                                                autoscalingSpecPath(i, "name"),
                                                strings.Join(autoscalingSpec.Roles, ","),
-                                               "roles set is duplicated"),
+                                               fmt.Sprintf("roles set is duplicated (ignoring %s)", esv1.RemoteClusterClientRole)),
                                )
                        } else {
-                               rolesSet = append(rolesSet, autoscalingSpec.Roles)
+                               rolesSet = append(rolesSet, effectiveRoles)
                        }
                }
 
                // Machine learning nodes must be in a dedicated tier.
-               if stringsutil.StringInSlice(string(esv1.MLRole), autoscalingSpec.Roles) && len(autoscalingSpec.Roles) > 1 {
+               if stringsutil.StringInSlice(string(esv1.MLRole), effectiveRoles) && len(effectiveRoles) > 1 {
                        errs = append(
                                errs,
                                field.Invalid(
```

barkbay (approved):
* I did a quick test, scaling up and down from 0 for the ML tier* and it works as expected 👍 
* I'm fine merging it in `2.10`
* As mentioned in my last comment I'm wondering if we should not apply a similar logic to the other tiers regarding the way `remote_cluster_client` is handled in the validation logic, not a blocker for this PR though.

<details>
<summary>*</summary>
<p>

```yaml
---
apiVersion: autoscaling.k8s.elastic.co/v1alpha1
kind: ElasticsearchAutoscaler
metadata:
  name: autoscaling-sample
spec:
  elasticsearchRef:
    name: elasticsearch-sample
  policies:
    - name: di
      roles: ["data", "ingest" , "transform", "remote_cluster_client"]
      resources:
        nodeCount:
          min: 3
          max: 8
        cpu:
          min: 2
          max: 8
        memory:
          min: 2Gi
          max: 16Gi
        storage:
          min: 64Gi
          max: 512Gi
    - name: ml
      roles:
        - ml
        - remote_cluster_client
      deciders:
        ml:
          down_scale_delay: 2m ## scale down faster
      resources:
        nodeCount:
          min: 0
          max: 2
        cpu:
          min: 1
          max: 4
        memory:
          min: 2Gi
          max: 8Gi
        storage:
          min: 1Gi
          max: 1Gi

---
kind: Elasticsearch
metadata:
  name: elasticsearch-sample
spec:
  version: 8.10.4
  nodeSets:
    - name: master
      count: 1
      podTemplate:
        spec:
          containers:
            - name: elasticsearch
              # Specify both limits and requests to benefit from the desired nodes API / Elasticsearch resources aware management.
              resources:
                limits:
                  memory: 4Gi
                  cpu: 2
      config:
        node:
          roles: [ "master" ]
          store.allow_mmap: false
    - name: di
      config:
        node:
          roles: [ "data", "ingest", "transform", "remote_cluster_client" ]
          store.allow_mmap: false
    - name: ml
      config:
        node:
          roles: [ "ml", "remote_cluster_client" ]
          store.allow_mmap: false
---
kind: Kibana
metadata:
  name: kibana-sample
spec:
  version: 8.10.4
  count: 1
  elasticsearchRef:
    name: "elasticsearch-sample"
```

</p>
</details>

naemono:
> * As mentioned in my last comment I'm wondering if we should not apply a similar logic to the other tiers regarding the way `remote_cluster_client` is handled in the validation logic, not a blocker for this PR though.

Opened #7285 to address.

naemono:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|2.10|[](https://github.com/elastic/cloud-on-k8s/pull/7287)|

### Questions ?

PR #7284, opened 2023-11-03
`make e2e-local` Fixes

naemono:
## What is this changing
1. Fixes this issue running `make e2e-local`.
```
test/e2e/run.sh: line 19: E2E_TAGS: unbound variable
```
I'm honestly not 100% sure about this fix, but it seems to work locally for me. Please correct me if my syntax is slightly off.

2. `sleep infinity` doesn't work on `Darwin`, so if `uname` isn't `Linux` use `sleep 86400` which is `24 hours`.

### DIFF
--- a/test/e2e/run.sh
+++ b/test/e2e/run.sh
@@ -38,11 +38,9 @@ main() {
     run_e2e_tests "$@" | tee "$ARTEFACTS_DIR/e2e-tests-$CLUSTER_NAME.json"
   fi
   touch /tmp/done
-  if [ "$(uname)" == "Linux" ] ; then
-    sleep infinity
-  else
-    sleep 86400
-  fi
+  while true; do
+    sleep 60
+  done
 }
 
 main "$@"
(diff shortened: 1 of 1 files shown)


test/e2e/run.sh
```diff
@@ -37,7 +38,11 @@ main() {
     run_e2e_tests "$@" | tee "$ARTEFACTS_DIR/e2e-tests-$CLUSTER_NAME.json"
   fi
   touch /tmp/done
-  sleep infinity
+  if [ "$(uname)" == "Linux" ] ; then
+    sleep infinity
+  else
+    sleep 86400
```
rhr323, line 44: Out of curiosity, why not sleep 24h on all platforms?

rhr323 (approved):
LGTM 👍 !

test/e2e/run.sh
pebrc: +1 one for just one implementation that works cross platform. Naively I would think that 
```
while true; do
  sleep 1 
done
```
is the equivalent to `sleep infinity`?

thbkrkr: To give a bit of context, I wrote `sleep infinity` because it seemed like a clean and explicit way to wait forever but I missed that it wasn't cross-platform.

+1 for a more portable way, `while true; do sleep 42; done` or `tail -f /dev/null`.

naemono: This has been updated, lmk your thoughts.

thbkrkr (approved):

PR #7322, opened 2023-11-16
Update beats stack monitoring recipe

thbkrkr:
This configures beats in the stack monitoring recipe with `ssl.verification_mode: "certificate"` 
as https://github.com/elastic/beats/issues/8164 is resolved and removes the associated `TODO` explaining to do that.


barkbay (approved):
LGTM

PR #7324, opened 2023-11-17
Introduce Kibana config field in StackConfigPolicy

kvalliyurnatt:
## What does this change?
Introduce Kibana config Field in StackConfigPolicy. This field can be used to manage multiple kibana instances using stackConfigPolicy. The only configuration that can be configured for kibana using the StackConfigPolicy is the kibanaConfig which are the configurations that go into the kibana.yml file.

- [x] Add unit tests
- [x] Test locally in dev 
- [x] Introduce secure settings fields individually under ES and Kibana
- [x] Try to configure an auth policy

## Testing
I Have tested with all four auth config policies that we intend to support , JWT, OIDC, SAML and LDAP. I was able to successfully set these up with the StackConfigPolicy and also login into Kibana(make API calls against Elasticsearch in case of JWT) successfully.

However there are limitations here. With JWT and LDAP it is straightforward to configure multiple elasticsearch clusters with the same auth config. With OIDC we generate client secret and application ID and while doing so in the identity provider we enter a redirect url which is generally `KIBANA_ENDPOINT_URL/api/security/oidc/callback` now to be able to configure multiple KIbana instances to work with this, there are some identity providers like Google that allow you enter more than on redirect url. So it is possible to configure OIDC for multiple elasticsearch clusters as long as the provider being used allows entering multiple urls for the redirect url. However for SAML I could not find a similar option and to me it seems like we might not be able to configure multiple instances using the same SAML config as the SAML config is generated using a redirect url that points to a specific kibana instance. 

### JWT
I tested configuring JWT auth policy and it worked as expected. I add the jwks secret, elasticsearch config, role mapping and shared secret to the stackconfigpolicy, and then used the JWT and shared secret in headers to call Elasticsearch

```
curl -s -k -X GET -H "Authorization: Bearer $JWT" -H "ES-Client-Authentication: sharedsecret $SHAREDSECRET" https://localhost:9200/_security/_authenticate -H 'Content-Type: application/json'
```
Output
```
{"username":"elastic-agent","roles":["remote_monitoring_collector"],"full_name":null,"email":null,"metadata":{"jwt_claim_jti":"<removed>","jwt_token_type":"id_token","jwt_claim_iss":"https://es.credentials.controller.k8s.elastic.co","jwt_claim_aud":["elasticsearch"],"jwt_claim_sub":"elastic-agent"},"enabled":true,"authentication_realm":{"name":"jwt1","type":"jwt"},"lookup_realm":{"name":"jwt1","type":"jwt"},"authentication_type":"realm"}
```

I used the elastic-agent tokens so that I don't have to create a new role. But I found this was causing issues to Kibana due to the secure settings being applied to both Kibana and Elasticsearch. Working on a fix for this

### OIDC
StackConfigPolicy
```
kind: StackConfigPolicy
metadata:
  name: test-stack-config-policy
  # namespace: elastic-system or test-namespace
spec:
  resourceSelector:
    matchLabels:
      env: my-label
  elasticsearch:
    secureSettings:
    - secretName: oidc-secret
    securityRoleMappings:
      oidc_kibana:
        roles: [ "superuser" ]
        rules:
          all:
            - field: { realm.name: "oidc1" }
            - field: { username: "<username>" }
        enabled: true
    config:
       logger.org.elasticsearch.discovery: DEBUG
       xpack:
         security:
           authc:
             token.enabled: true
             realms:
               oidc:
                 oidc1:
                   order: 2
                   rp.client_id: "<Application Client ID>"
                   rp.response_type: "code"
                   rp.requested_scopes: ["openid", "email"]
                   rp.redirect_uri: "https://kibana.eck-ocp.elastic.dev/api/security/oidc/callback"
                   op.issuer: "https://accounts.google.com"
                   op.authorization_endpoint: "https://accounts.google.com/o/oauth2/v2/auth"
                   op.token_endpoint: "https://oauth2.googleapis.com/token"
                   op.userinfo_endpoint: "https://openidconnect.googleapis.com/v1/userinfo"
                   op.jwkset_path: "https://www.googleapis.com/oauth2/v3/certs"
                   claims.principal: email
                   claim_patterns.principal: "^([^@]+)@elastic\\.co$"
    secretMounts:
    - secretName: "test1234567891234567"
      mountPath: "test"
  kibana:
    config:
      server:
        publicBaseUrl: "https://kibana.eck-ocp.elastic.dev"
      xpack.security.authc.providers:
        oidc.oidc1:
          order: 0
          realm: oidc1
          description: "Log in with GCP"
        basic.basic1:
          order: 1
```

Screenshot 2023-12-06 at 7 41 29 PM

### SAML with Okta
StackConfigPolicy
```
kind: StackConfigPolicy
metadata:
  name: test-stack-config-policy
  # namespace: elastic-system or test-namespace
spec:
  resourceSelector:
    matchLabels:
      env: my-label
  elasticsearch:
    securityRoleMappings:
      cloud_saml_kibana:
        roles: [ "superuser" ]
        rules:
          all:
            - field: { realm.name: "test-saml" }
            - field: { username: "<email address>" }
        enabled: true
    config:
      logger.org.elasticsearch.discovery: DEBUG
      xpack.security.authc.realms:
        saml.test-saml:
          order: 2
          nameid_format: "urn:oasis:names:tc:SAML:2.0:nameid-format:transient"
          attributes.principal: "nameid:persistent"
          attributes.groups: "groups"
          idp.metadata.path: "<okta idp url>"
          idp.entity_id: "<okta entity id url>"
          sp.entity_id: "https://kibana.eck-ocp.elastic.dev"
          sp.acs: "https://kibana.eck-ocp.elastic.dev/api/security/saml/callback"
          sp.logout: "https://kibana.eck-ocp.elastic.dev/logout"
  kibana:
    config:
      server:
        publicBaseUrl: "https://kibana.eck-ocp.elastic.dev"
      xpack.security.authc.providers:
        saml.saml1:
          order: 0
          realm: test-saml
          description: "Log in with Okta"
          icon: "https://www.okta.com/themes/custom/okta_www_theme/images/logo.svg"
        basic.basic1:
          order: 1
```

Screenshot 2023-12-07 at 5 10 47 PM

### LDAP
I set up a LDAP server using the openLDAP helm chart on my dev kubernetes cluster and then configured two elasticsearch clusters via stack config policy to interact with it.  I was able to sucessfully login via kibana using a LDAP user.
stackConfigPolicy
```
kind: StackConfigPolicy
metadata:
  name: test-stack-config-policy
  # namespace: elastic-system or test-namespace
spec:
  resourceSelector:
    matchLabels:
      env: my-label
  elasticsearch:
    secureSettings:
    - secretName: ldap-secret
    securityRoleMappings:
      inline_ldap_role:
        roles: [ "superuser" ]
        rules:
          all:
            - field: { realm.name: "ldap1" }
            - field: { dn: "*,ou=users,dc=example,dc=org" }
        enabled: true
    config:
      xpack.security.authc.realms:
        ldap:
          ldap1:
            order: 0
            url: "ldap://openldap.default.svc.cluster.local:1389"
            bind_dn: "cn=admin,dc=example,dc=org"
            user_search:
              base_dn: "dc=example,dc=org"
              filter: "(cn={0})"
            group_search:
              base_dn: "dc=example,dc=org"
            unmapped_groups_as_roles: false
```


pkg/apis/stackconfigpolicy/v1alpha1/stackconfigpolicy_types.go
```diff
@@ -180,49 +195,77 @@ func (s *StackConfigPolicyStatus) setReadyCount() {
...
+		ResourceType: resourceType,
 	}
+	switch resourceType {
+	case ElasticsearchResourceType:
+		resourcePolicyStatus.ElasticsearchStatus = ElasticsearchPolicyStatus{
+			Phase: phase,
+			Error: PolicyStatusError{Message: msg},
+		}
+	case KibanaResourceType:
+		resourcePolicyStatus.KibanaStatus = KibanaPolicyStatus{
```
barkbay, line 212: It feels like we are missing an interface or common struct that would avoid this code duplication.

pkg/apis/stackconfigpolicy/v1alpha1/stackconfigpolicy_types.go
```diff
@@ -180,49 +195,77 @@ func (s *StackConfigPolicyStatus) setReadyCount() {
...
+	s.ResourcesStatuses[s.getResourceStatusKey(resource, resourceType)] = resourcePolicyStatus
 	s.Update()
 	return nil
 }
 
-func (s *StackConfigPolicyStatus) UpdateResourceStatusPhase(resource types.NamespacedName, status ResourcePolicyStatus, elasticsearchConfigAndMountsApplied bool) {
+func (s *StackConfigPolicyStatus) UpdateResourceStatusPhase(resource types.NamespacedName, status ResourcePolicyStatus, applicationConfigsApplied bool) {
 	defer func() {
-		s.ResourcesStatuses[resource.String()] = status
+		s.ResourcesStatuses[s.getResourceStatusKey(resource, status.ResourceType)] = status
```
barkbay, line 226: Should we check that this map not nil?

barkbay (commented):
```
\"test-stack-config-policy\" is invalid: spec.elasticsearch: Required value: Elasticsearch settings are mandatory and must not be empty"
```

I guess we should add something along those lines in the validation logic:

```golang
	if policy.Spec.Kibana.Config != nil {
		settingsCount += len(policy.Spec.Kibana.Config.Data)
	}
```

pkg/apis/stackconfigpolicy/v1alpha1/stackconfigpolicy_types.go
```diff
@@ -180,49 +195,77 @@ func (s *StackConfigPolicyStatus) setReadyCount() {
...
 		s.Update()
 	}()
+	switch status.ResourceType {
```
barkbay, line 229: Should we have a `default:`?

pkg/apis/stackconfigpolicy/v1alpha1/stackconfigpolicy_types.go
```diff
@@ -180,66 +190,97 @@ func (s *StackConfigPolicyStatus) setReadyCount() {
...
-		status.Phase = ErrorPhase
-		if status.ExpectedVersion > status.Error.Version {
+		if status.CurrentVersion == status.ExpectedVersion {
+			status.Phase = ReadyPhase
+			return nil
+		}
+		status.Phase = ApplyingChangesPhase
+	case KibanaResourceType:
+		if !applicationConfigsApplied {
+			// New KibanaConfig not yet applied to the Elasticsearch pod
```
naemono, line 248:
```suggestion
			// New KibanaConfig not yet applied to the Elasticsearch instance
```

pkg/controller/common/labels/labels.go
```diff
@@ -16,8 +16,8 @@ const (
 	StackConfigPolicyOnDeleteLabelName = "asset.policy.k8s.elastic.co/on-delete"
 	// OrphanSecretResetOnPolicyDelete is used to set the data field of a secret to an empty map when the associated StackConfigPolicy or Elasticsearch is deleted.
 	OrphanSecretResetOnPolicyDelete = "reset"
-	// OrphanSecretResetOnPolicyDelete is used to delete the  secret when the associated stackconfigpolicy or Elasticsearch is deleted.
-	OrphanObjectDeleteOnPolicyDelete = "delete"
+	// OrphanSecretDeleteOnPolicyDelete is used to delete the  secret when the associated stackconfigpolicy or Elasticsearch is deleted.
```
naemono, line 19:
```suggestion
	// OrphanSecretDeleteOnPolicyDelete is used to delete the secret when the associated StackConfigPolicy or Elasticsearch is deleted.
```

pkg/controller/kibana/label/label.go
```diff
@@ -17,3 +23,11 @@ const (
 	// Type represents the Kibana type
 	Type = "kibana"
 )
+
+// NewLabels constructs a new set of labels from an Elasticsearch definition.
```
naemono, line 27: Is it really from an Elasticsearch definition? Seems like a Kibana definition.

pkg/controller/stackconfigpolicy/controller.go
```diff
@@ -333,22 +368,104 @@ func (r *ReconcileStackConfigPolicy) doReconcile(ctx context.Context, policy pol
...
+			if err := status.AddPolicyErrorFor(kibanaNsn, policyv1alpha1.ConflictPhase, err.Error(), policyv1alpha1.KibanaResourceType); err != nil {
+				return results.WithError(err), status
+			}
+			continue
+		}
+
+		_, err = reconciler.ReconcileSecret(ctx, r.Client, expectedConfigSecret, &kibana)
+		if err != nil {
+			return results.WithError(err), status
+		}
```
naemono, line 445: might as well inline this since you don't use the result.

naemono (commented):
Haven't gotten fully through this, or tested locally, but some initial things I saw.

pkg/controller/stackconfigpolicy/controller.go
```diff
@@ -414,98 +530,161 @@ func (r *ReconcileStackConfigPolicy) updateStatus(ctx context.Context, scp polic
...
-		if err != nil && !apierrors.IsNotFound(err) {
-			return err
-		}
-		if apierrors.IsNotFound(err) {
-			// Elasticsearch has just been deleted
-			return nil
+			err = filesettings.ReconcileEmptyFileSettingsSecret(ctx, c, es, false)
+			if err != nil {
+				return err
+			}
```
naemono, line 615:
```suggestion
			if err := filesettings.ReconcileEmptyFileSettingsSecret(ctx, c, es, false); err != nil {
				return err
			}
```

pkg/controller/stackconfigpolicy/kibana_config_settings.go
```diff
@@ -0,0 +1,118 @@
...
+
+func newKibanaConfigSecret(policy policyv1alpha1.StackConfigPolicy, kibana kibanav1.Kibana) (corev1.Secret, error) {
+	kibanaConfigHash := getKibanaConfigHash(policy.Spec.Kibana.Config)
+	var configDataJSONBytes []byte
+	var err error
+	if policy.Spec.Kibana.Config != nil {
+		configDataJSONBytes, err = policy.Spec.Kibana.Config.MarshalJSON()
+		if err != nil {
+			return corev1.Secret{}, err
+		}
```
naemono, line 39: Since you declared the vars above, you can inline this as well...

pkg/controller/stackconfigpolicy/kibana_config_settings.go
```diff
@@ -0,0 +1,118 @@
...
+			Labels: kblabel.NewLabels(types.NamespacedName{
+				Name:      kibana.Name,
+				Namespace: kibana.Namespace,
+			}),
+			Annotations: map[string]string{
+				KibanaConfigHashAnnotation: kibanaConfigHash,
+			},
+		},
+		Data: map[string][]byte{
+			KibanaConfigKey: configDataJSONBytes,
```
naemono, line 54: Should we avoid creating this secret if the `policy.Spec.Kibana.Config == nil`? It's just going to be an empty secret..

pkg/controller/stackconfigpolicy/kibana_config_settings.go
```diff
@@ -0,0 +1,118 @@
...
+	kibanaConfigSecret := corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: kibana.Namespace,
+			Name:      GetPolicyConfigSecretName(kibana),
+			Labels: kblabel.NewLabels(types.NamespacedName{
+				Name:      kibana.Name,
+				Namespace: kibana.Namespace,
+			}),
+			Annotations: map[string]string{
+				KibanaConfigHashAnnotation: kibanaConfigHash,
```
naemono, line 50: And this would be an empty annotation because the config hash would be `""`, right? Seems odd.

naemono (commented):
Still yet to test this locally, but have gotten through looking at all of the changes...

pebrc (commented):
Code looks good to me. I have not gotten round to actually testing it though.

pkg/controller/stackconfigpolicy/kibana_config_settings.go
```diff
@@ -0,0 +1,118 @@
...
+	commonlabels "github.com/elastic/cloud-on-k8s/v2/pkg/controller/common/labels"
+	"github.com/elastic/cloud-on-k8s/v2/pkg/controller/common/reconciler"
+	"github.com/elastic/cloud-on-k8s/v2/pkg/controller/elasticsearch/filesettings"
+	kblabel "github.com/elastic/cloud-on-k8s/v2/pkg/controller/kibana/label"
+	"github.com/elastic/cloud-on-k8s/v2/pkg/utils/k8s"
+
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
```
thbkrkr, line 23: Nit: reorder imports.

pkg/controller/stackconfigpolicy/elasticsearch_config_settings.go
```diff
@@ -119,3 +121,23 @@ func getElasticsearchConfigAndMountsHash(elasticsearchConfig *commonv1.Config, s
 	}
 	return hash.HashObject(secretMounts)
 }
+
+// elasticsearchConfigAndSecretMountsApplied checks if the elasticsearch config and secret mounts from the stack config policy have been applied to the Elasticsearch cluster.
+func elasticsearchConfigAndSecretMountsApplied(ctx context.Context, c k8s.Client, policy policyv1alpha1.StackConfigPolicy, es esv1.Elasticsearch) (bool, error) {
+	// Get Pods for the given Elasticsearch
+	podList := corev1.PodList{}
+	if err := c.List(ctx, &podList, client.MatchingLabels{
```
thbkrkr, line 129: Could it be an issue if there is 2 ES in 2 namespaces and only one managed by a SCP?
In any case, it's better to reduce the spectrum of a list to the namespace when we can.
​
```suggestion
	if err := c.List(ctx, &podList, client.InNamespace(es.Namespace), client.MatchingLabels{
```

pkg/controller/kibana/policy_config_setting.go
```diff
@@ -0,0 +1,65 @@
...
+	kibanav1 "github.com/elastic/cloud-on-k8s/v2/pkg/apis/kibana/v1"
+	common "github.com/elastic/cloud-on-k8s/v2/pkg/controller/common/settings"
+	"github.com/elastic/cloud-on-k8s/v2/pkg/controller/stackconfigpolicy"
+	"github.com/elastic/cloud-on-k8s/v2/pkg/utils/k8s"
+)
+
+// PolicyConfig is a structure for storing Kibana config from the StackConfigPolicy
+type PolicyConfig struct {
+	KibanaConfig      *common.CanonicalConfig
+	PolicyAnnotations map[string]string
```
thbkrkr, line 24: Nit: `policyConfig.PodAnnotations` sounds better to me.
```suggestion
	PodAnnotations map[string]string
```

pkg/controller/kibana/config_settings.go
```diff
@@ -157,8 +157,8 @@ func NewConfigSettings(ctx context.Context, client k8s.Client, kb kbv1.Kibana, v
 		}
 	}
 
-	// merge the configuration with userSettings last so they take precedence
-	if err = cfg.MergeWith(userSettings); err != nil {
+	// Kibana config from the StackConfigPolicy takes precedence over userSettings, merge it last.
```
thbkrkr, line 160:
```suggestion
	// Kibana settings from a StackConfigPolicy takes precedence over user provided settings, merge them last.
```

pkg/controller/stackconfigpolicy/controller.go
```diff
@@ -184,6 +192,10 @@ func (r *ReconcileStackConfigPolicy) Reconcile(ctx context.Context, request reco
 // clusters configured by a StackConfigPolicy.
 type esMap map[types.NamespacedName]esv1.Elasticsearch
 
+// kbMap is a type alias for a Map of Kibana indexed by NamespaceName useful to manipulate the Kibana
+// clusters configured by a StackConfigPolicy.
```
thbkrkr, line 196:
```suggestion
// instances configured by a StackConfigPolicy.
```

pkg/controller/stackconfigpolicy/controller.go
```diff
@@ -333,22 +368,103 @@ func (r *ReconcileStackConfigPolicy) doReconcile(ctx context.Context, policy pol
...
+	if err := r.Client.List(ctx, &kibanaList, &listOpts); err != nil {
+		return results.WithError(err), status
+	}
+
+	configuredResources := kbMap{}
+	for _, kibana := range kibanaList.Items {
+		log.V(1).Info("Reconcile StackConfigPolicy", "policy_namespace", policy.Namespace, "policy_name", policy.Name, "kibana_namespace", kibana.Namespace, "kibana_name", kibana.Name)
+		kibana := kibana
+
+		// keep the list of ES to be configured
```
thbkrkr, line 415:
```suggestion
		// keep the list of Kibana to be configured
```

pkg/controller/stackconfigpolicy/controller.go
```diff
@@ -333,22 +368,103 @@ func (r *ReconcileStackConfigPolicy) doReconcile(ctx context.Context, policy pol
...
+	// delete Settings secrets for resources no longer selected by this policy
+	results.WithError(deleteOrphanSoftOwnedSecrets(ctx, r.Client, k8s.ExtractNamespacedName(&policy), nil, configuredResources, policyv1alpha1.KibanaResourceType))
+
 	return results, status
 }
 
-func newResourceStatus(currentSettings esclient.FileSettings, expectedVersion int64) policyv1alpha1.ResourcePolicyStatus {
+func newElasticsearchResourceStatus(currentSettings esclient.FileSettings, expectedVersion int64) policyv1alpha1.ResourcePolicyStatus {
 	status := policyv1alpha1.ResourcePolicyStatus{
+
```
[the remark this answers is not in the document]
kvalliyurnatt: wonder why the linter didn't complain.

pkg/controller/stackconfigpolicy/controller.go
```diff
@@ -414,98 +530,161 @@ func (r *ReconcileStackConfigPolicy) updateStatus(ctx context.Context, scp polic
...
-			reconciler.SoftOwnerKindLabel:                   policyv1alpha1.Kind,
-			commonlabels.StackConfigPolicyOnDeleteLabelName: commonlabels.OrphanObjectDeleteOnPolicyDelete,
-		},
+		matchLabels,
 	); err != nil {
 		return err
 	}
 
 	for i := range secrets.Items {
 		secret := secrets.Items[i]
```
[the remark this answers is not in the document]
kvalliyurnatt: I think it is because the linter will complain when we try send the memory address of the secret here
```
err := c.Delete(ctx, &secret)
		if err != nil && !apierrors.IsNotFound(err) {
			return err
		}
```
referencing memory of temporary loop variables

pkg/controller/kibana/policy_config_setting_test.go
```diff
@@ -0,0 +1,117 @@
...
+				"policy.k8s.elastic.co/kibana-config-hash": hashValue,
+			},
+		},
+		Data: map[string][]byte{
+			"kibana.json": []byte(`{"xpack.canvas.enabled":true}`),
+		},
+	}
+}
+
+func mkInvalidCluster(namespace string, owningPolicyName string, owningPolicyNamespace string, hashValue string) *corev1.Secret {
```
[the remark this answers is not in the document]
kvalliyurnatt: yeah no idea what I was thinking there 🤷

pkg/controller/stackconfigpolicy/kibana_config_settings.go
kvalliyurnatt: I thought it will be simpler to create it, even though it is empty, I can take another look at this.

pkg/controller/stackconfigpolicy/kibana_config_settings.go
kvalliyurnatt: I think even an empty config generates some hash.  I will double check

pkg/apis/stackconfigpolicy/v1alpha1/stackconfigpolicy_types.go
```diff
@@ -52,10 +54,12 @@ type StackConfigPolicyList struct {
 }
 
 type StackConfigPolicySpec struct {
-	ResourceSelector metav1.LabelSelector          `json:"resourceSelector,omitempty"`
-	SecureSettings   []commonv1.SecretSource       `json:"secureSettings,omitempty"`
-	Elasticsearch    ElasticsearchConfigPolicySpec `json:"elasticsearch,omitempty"`
-	Kibana           KibanaConfigPolicySpec        `json:"kibana,omitempty"`
+	ResourceSelector metav1.LabelSelector `json:"resourceSelector,omitempty"`
+	// SecureSettings applies only to Elasticsearch
+	// TODO: do we deprecate this and remove it in the future ?
```
barkbay, line 59: FWIW it should be possible to send a warning to the user when the resource is validated, with something along the lines of:

```diff
diff --git a/pkg/apis/stackconfigpolicy/v1alpha1/stackconfigpolicy_types.go b/pkg/apis/stackconfigpolicy/v1alpha1/stackconfigpolicy_types.go
index 475ae15c8..56165c67e 100644
--- a/pkg/apis/stackconfigpolicy/v1alpha1/stackconfigpolicy_types.go
+++ b/pkg/apis/stackconfigpolicy/v1alpha1/stackconfigpolicy_types.go
@@ -56,7 +56,7 @@ type StackConfigPolicyList struct {
 type StackConfigPolicySpec struct {
        ResourceSelector metav1.LabelSelector `json:"resourceSelector,omitempty"`
        // SecureSettings applies only to Elasticsearch
-       // TODO: do we deprecate this and remove it in the future ?
+       // Deprecated: SecureSettings only applies to Elasticsearch and will be removed in a next release. It must be set per application instead.
        SecureSettings []commonv1.SecretSource       `json:"secureSettings,omitempty"`
        Elasticsearch  ElasticsearchConfigPolicySpec `json:"elasticsearch,omitempty"`
        Kibana         KibanaConfigPolicySpec        `json:"kibana,omitempty"`
diff --git a/pkg/apis/stackconfigpolicy/v1alpha1/webhook.go b/pkg/apis/stackconfigpolicy/v1alpha1/webhook.go
index 2ca16658a..f73b0f455 100644
--- a/pkg/apis/stackconfigpolicy/v1alpha1/webhook.go
+++ b/pkg/apis/stackconfigpolicy/v1alpha1/webhook.go
@@ -5,6 +5,8 @@
 package v1alpha1
 
 import (
+       "fmt"
+
        apierrors "k8s.io/apimachinery/pkg/api/errors"
        "k8s.io/apimachinery/pkg/runtime"
        "k8s.io/apimachinery/pkg/runtime/schema"
@@ -36,6 +38,20 @@ var (
 
 var _ webhook.Validator = &StackConfigPolicy{}
 
+const (
+       SpecSecureSettingsDeprecated = "spec.SecureSettings is deprecated, secure settings must be set per application"
+)
+
+func (s *StackConfigPolicy) GetWarnings() []string {
+       if s == nil {
+               return nil
+       }
+       if len(s.Spec.SecureSettings) > 0 {
+               return []string{fmt.Sprintf("%s %s/%s: %s", Kind, s.Namespace, s.Name, SpecSecureSettingsDeprecated)}
+       }
+       return nil
+}
+
 // ValidateCreate is called by the validating webhook to validate the create operation.
 // Satisfies the webhook.Validator interface.
 func (p *StackConfigPolicy) ValidateCreate() (admission.Warnings, error) {
```

Not tested, but we do something similar for Agent: https://github.com/elastic/cloud-on-k8s/blob/5937685b9085a5b185dde3d31ad50ccb5cdc034b/pkg/apis/agent/v1alpha1/webhook.go#L37-L45

pkg/controller/stackconfigpolicy/controller.go
```diff
@@ -333,21 +374,107 @@ func (r *ReconcileStackConfigPolicy) doReconcile(ctx context.Context, policy pol
...
 
-	// requeue if not ready
-	if status.Phase != policyv1alpha1.ReadyPhase {
-		results.WithResult(defaultRequeue)
+	return results, status
+}
+
+func (r *ReconcileStackConfigPolicy) reconcileKibanaResources(ctx context.Context, policy policyv1alpha1.StackConfigPolicy, status policyv1alpha1.StackConfigPolicyStatus) (*reconciler.Results, policyv1alpha1.StackConfigPolicyStatus) {
+	log := ulog.FromContext(ctx)
+	log.V(1).Info("Reconcile Kibana Resources", "namespace", policy.Namespace, "policy_name", policy.Name)
```
barkbay, line 397: Something is wrong with the values, some of them are duplicated like `namespace` or `policy_name` are duplicated:

```
2023-12-07T15:38:42.467+0100    DEBUG   stackconfigpolicy-controller    Reconcile Kibana Resources      {"service.version": "0.0.0-SNAPSHOT+00000000", "iteration": "11", "namespace": "elastic-system", "policy_name": "test-stack-config-policy", "namespace": "elastic-system", "policy_name": "test-stack-config-policy"}
```

pkg/controller/elasticsearch/filesettings/reconciler.go
```diff
@@ -60,18 +61,18 @@ func ReconcileSecret(
...
 		Context:    ctx,
 		Client:     c,
-		Owner:      &es,
+		Owner:      owner,
 		Expected:   &expected,
 		Reconciled: reconciled,
 		NeedsUpdate: func() bool {
-			return !maps.IsSubset(expected.Labels, reconciled.Labels) ||
-				!maps.IsSubset(expected.Annotations, reconciled.Annotations) ||
+			return !reflect.DeepEqual(expected.Labels, reconciled.Labels) ||
```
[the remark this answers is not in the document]
kvalliyurnatt: @pebrc 
Two things why we can't/do not need to use IsSubset here
1) These are operator created settings not user created ones, so there are unlikely to be additional labels and annotations.
2) more importantly though, this leads to a bug that is currently present even with elasticsearch filesettings, consider this scenario
say currently the secret has the following annotations
```
annotations:
    policy.k8s.elastic.co/secure-settings-secrets: '[{"namespace":"default","secretName":"oidc-secret"}]'
    policy.k8s.elastic.co/settings-hash: "4284554765"
```
And we want to remove the secure settings annotations, because the secure setting is removed from the stackconfigpolicy,  so we want the new annotations to be the following
```
annotations:
    policy.k8s.elastic.co/settings-hash: "4284554765"
```
With the check that was in place
!maps.IsSubset(expected.Annotations, reconciled.Annotations)
expected annotations is a subset of reconciled annotations, so this is going to be false and we will never update the secret. So during my testing I found that removing secureSettings from the SCP had no effect

 instead if we check for equality here and then in the update function we merge the annotations and remove only the managed annotations (which is what we do) I think it still achieves what we want with preserving additional labels and annotations introduced by third party systems.

pkg/apis/stackconfigpolicy/v1alpha1/stackconfigpolicy_types.go
```diff
@@ -52,10 +54,11 @@ type StackConfigPolicyList struct {
 }
 
 type StackConfigPolicySpec struct {
-	ResourceSelector metav1.LabelSelector          `json:"resourceSelector,omitempty"`
-	SecureSettings   []commonv1.SecretSource       `json:"secureSettings,omitempty"`
-	Elasticsearch    ElasticsearchConfigPolicySpec `json:"elasticsearch,omitempty"`
-	Kibana           KibanaConfigPolicySpec        `json:"kibana,omitempty"`
+	ResourceSelector metav1.LabelSelector `json:"resourceSelector,omitempty"`
+	// Deprecated: SecureSettings only applies to Elasticsearch and will be removed in a future release. It must be set per application instead.
```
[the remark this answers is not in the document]
kvalliyurnatt: Adjusted the wording to just say this is deprecated.

pkg/controller/elasticsearch/filesettings/reconciler.go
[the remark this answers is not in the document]
pebrc: I agree with @barkbay and in this particular case could you not just use an empty array to reset the annotation instead of removing it? That would be compatible with `IsSubSet` as it also compares values.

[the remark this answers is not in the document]
kvalliyurnatt: @barkbay @pebrc I understand what we are trying to do here and that we want to preserve labels and annotations not added by the operator and not modify it, but the change does not change behaviour right ? even though we check for equality, in the upgrade function we do this 
```
reconciled.Labels = maps.Merge(reconciled.Labels, expected.Labels)
			// remove managed labels if they are no longer defined
			for _, label := range managedLabels {
				if _, ok := expected.Labels[label]; !ok {
					delete(reconciled.Labels, label)
				}
			}
			reconciled.Annotations = maps.Merge(reconciled.Annotations, expected.Annotations)
			// remove managed annotations if they are no longer defined
			for _, annotation := range managedAnnotations {
				if _, ok := expected.Annotations[annotation]; !ok {
					delete(reconciled.Annotations, annotation)
				}
			}
			reconciled.Data = expected.Data
```
which should preserve the user added labels etc right ? I did not fully understand the concern.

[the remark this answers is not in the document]
kvalliyurnatt: cleared this up offline with @pebrc and @thbkrkr , will add a new function that deals with only checking a specific set of labels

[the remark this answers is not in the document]
kvalliyurnatt: @pebrc @barkbay @thbkrkr I add a new function that you can pass in a set of managed keys and only those will be compared. Please take a look. TIA

kvalliyurnatt:
buildkite test this -f p=gke

pkg/utils/maps/maps.go
```diff
@@ -19,6 +19,16 @@ func IsSubset(toCheck, fullSet map[string]string) bool {
 	return true
 }
 
+// ManagedKeysChanged compares the values of the given keys managed by the elasticsearch operator between two maps and returns true if they are not equal.
+func ManagedKeysChanged(map1, map2 map[string]string, managedkeysToCompare []string) bool {
```
pebrc, line 23:
```suggestion
func IsEqualSubset(map1, map2 map[string]string, keySubSet []string) bool {
```
I think there is nothing "managed" about this on the level of the map util package.

pebrc (approved):
LGTM. Nice work! I did some very basic tests only. We already discussed that we need some documentation updates.  Are you also  going to adjust/extend our e2e tests for the stack config policy feature?

pkg/utils/maps/maps.go
kvalliyurnatt: Thanks, I was not happy with the name myself.

kvalliyurnatt:
> LGTM. Nice work! I did some very basic tests only. We already discussed that we need some documentation updates. Are you also going to adjust/extend our e2e tests for the stack config policy feature?

Yeah I can look into adding/ adjusting existing e2e tests, can that be a separate PR or should be part of this ?

pebrc:
> > LGTM. Nice work! I did some very basic tests only. We already discussed that we need some documentation updates. Are you also going to adjust/extend our e2e tests for the stack config policy feature?
> 
> Yeah I can look into adding/ adjusting existing e2e tests, can that be a separate PR or should be part of this ?

Can be separate

pkg/utils/maps/maps.go
```diff
@@ -19,6 +19,16 @@ func IsSubset(toCheck, fullSet map[string]string) bool {
 	return true
 }
 
+// IsEqualSubset compares the values of the given subset of keys between two maps and returns true if they are not equal.
```
barkbay, line 22: nit: it feels a bit odd to me that a function named `IsEqual` returns `true` when things are `not` equal.

pkg/controller/elasticsearch/filesettings/reconciler.go
```diff
@@ -60,18 +61,18 @@ func ReconcileSecret(
...
 		Context:    ctx,
 		Client:     c,
-		Owner:      &es,
+		Owner:      owner,
 		Expected:   &expected,
 		Reconciled: reconciled,
 		NeedsUpdate: func() bool {
-			return !maps.IsSubset(expected.Labels, reconciled.Labels) ||
-				!maps.IsSubset(expected.Annotations, reconciled.Annotations) ||
+			return !maps.IsEqualSubset(expected.Labels, reconciled.Labels, managedLabels) ||
```
barkbay, line 74: If `IsEqualSubset` returns `true` when *not* equal do we need the `! `?

I can observe the following messages every time I start the operator while the `Secret` `sample-es-file-settings` is not updated:

```
2023-12-13T08:42:46.849+0100    INFO    stackconfigpolicy-controller    Updating resource       {"service.version": "0.0.0-SNAPSHOT+00000000", "iteration": "1", "namespace": "elastic-system", "policy_name": "test-stack-config-policy", "kind": "Secret", "namespace": "test-ns", "name": "sample-es-file-settings"}
```

*restart*

```
2023-12-13T08:44:52.461+0100    INFO    stackconfigpolicy-controller    Updating resource       {"service.version": "0.0.0-SNAPSHOT+00000000", "iteration": "1", "namespace": "elastic-system", "policy_name": "test-stack-config-policy", "kind": "Secret", "namespace": "test-ns", "name": "sample-es-file-settings"}
```

pkg/utils/maps/maps.go
pebrc: Oh yes we need to change that. Or rename it so that it becomes clear.

pkg/controller/stackconfigpolicy/controller.go
```diff
@@ -414,98 +541,159 @@ func (r *ReconcileStackConfigPolicy) updateStatus(ctx context.Context, scp polic
...
 
-		esNsn := types.NamespacedName{
-			Namespace: s.Namespace,
-			Name:      s.Labels[eslabel.ClusterNameLabelName],
-		}
-		_, exist := configuredEs[esNsn]
-		if exist {
-			continue
-		}
+			log.V(1).Info("Reconcile empty file settings Secret for Elasticsearch",
```
barkbay, line 610: I don't think it comes from your PR but we should take care of extra keys like `namespace` which are already included in the logger fetched from `ctx`:

```json
{"service.version": "0.0.0-SNAPSHOT+00000000", "iteration": "1", "namespace": "elastic-system", "policy_name": "test-stack-config-policy", "namespace": "es-ns", "es_name": "sample", "owner_namespace": "elastic-system", "owner_name": "test-stack-config-policy"}
```

`namespace` key is set twice which means that the generated JSON document is not valid.

kvalliyurnatt: thanks let me make sure that is removed.

pkg/utils/maps/maps.go
kvalliyurnatt: sorry for the oversight I will fix this.

pkg/controller/elasticsearch/filesettings/reconciler.go
kvalliyurnatt: sorry 🤦  fixing this.

kvalliyurnatt: @barkbay @pebrc please take another look, I changed the condition a bit so this keeps existing behaviour and also checks if managed fields have changed/removed.

kvalliyurnatt: I don't see unnecessary updates being sent in my testing now

pkg/controller/stackconfigpolicy/kibana_config_settings.go
kvalliyurnatt: After thinking about this a little more I made the necessary changes to create the secret only when a config is present

pkg/controller/stackconfigpolicy/kibana_config_settings.go
kvalliyurnatt: This should be addressed with the change to only create the secret when config is present

pkg/controller/elasticsearch/filesettings/reconciler.go
```diff
@@ -51,27 +54,33 @@ func ReconcileEmptyFileSettingsSecret(
...
 		Reconciled: reconciled,
 		NeedsUpdate: func() bool {
-			return !maps.IsSubset(expected.Labels, reconciled.Labels) ||
-				!maps.IsSubset(expected.Annotations, reconciled.Annotations) ||
+			/*
+				log := ulog.FromContext(ctx)
+				log.Info("Reconciled is", "reconciled", *reconciled)
+				log.Info("Expected is", "expected", expected)
+			*/
+			// needs update if expected labels/annotations are not a subset of the existing labels/annotions or if managed labels/annotations have been removed/changed or if the data itself has changed.
```
barkbay, line 81:
```suggestion
			// Secrets must be updated in the following cases:
			// * the expected labels/annotations are not a subset of the existing labels/annotations,
			// * the managed labels/annotations have been removed/changed,
			// * the data itself has changed.
```

pkg/controller/stackconfigpolicy/controller.go
```diff
@@ -199,19 +218,41 @@ func (r *ReconcileStackConfigPolicy) doReconcile(ctx context.Context, policy pol
...
+	if status.Phase != policyv1alpha1.ReadyPhase {
+		results.WithResult(defaultRequeue)
+	}
+
+	return results, status
+}
+
+func (r *ReconcileStackConfigPolicy) reconcileElasticsearchResources(ctx context.Context, policy policyv1alpha1.StackConfigPolicy, status policyv1alpha1.StackConfigPolicyStatus) (*reconciler.Results, policyv1alpha1.StackConfigPolicyStatus) {
+	log := ulog.FromContext(ctx)
+	log.V(1).Info("Reconcile Elasticsearch Resources")
```
barkbay, line 252:
```suggestion
	log.V(1).Info("Reconcile Elasticsearch resources")
```

pkg/controller/stackconfigpolicy/controller.go
```diff
@@ -333,21 +373,111 @@ func (r *ReconcileStackConfigPolicy) doReconcile(ctx context.Context, policy pol
...
+			if err := status.AddPolicyErrorFor(kibanaNsn, policyv1alpha1.ConflictPhase, err.Error(), policyv1alpha1.KibanaResourceType); err != nil {
+				return results.WithError(err), status
+			}
+			continue
+		}
+
+		// create expected kibana config secret
+		if policy.Spec.Kibana.Config != nil {
+			// Only add to configured resources if kibana config is set
+			// This will help clean up the config secret if config gets removed from the stack config policy
```
barkbay, line 449:
```suggestion
		// Create the Secret that holds the Kibana configuration.
		if policy.Spec.Kibana.Config != nil {
			// Only add to configured resources if Kibana config is set.
			// This will help clean up the config secret if config gets removed from the stack config policy.
```

pkg/controller/elasticsearch/filesettings/reconciler.go
```diff
@@ -51,27 +54,33 @@ func ReconcileEmptyFileSettingsSecret(
...
+			return (!maps.IsSubset(expected.Labels, reconciled.Labels) || !maps.IsEqualSubset(expected.Labels, reconciled.Labels, managedLabels)) ||
+				(!maps.IsSubset(expected.Annotations, reconciled.Annotations) || !maps.IsEqualSubset(expected.Annotations, reconciled.Annotations, managedAnnotations)) ||
 				!reflect.DeepEqual(expected.Data, reconciled.Data)
```
barkbay, line 84: Not from your PR but I just realized that this comparison is not effective when keys are empty, because `DeepEquals` does make the difference between an empty slice and a `nil` slice:

image

⬆️ This continuously returns `true`

 Or another way of looking at things would be to consider that we should not try to set unnecessary keys in a Secret as it is done in [newElasticsearchConfigSecret](https://github.com/elastic/cloud-on-k8s/blob/45782fb3fdd06e95fbc30e0b5efa31233f798c52/pkg/controller/stackconfigpolicy/elasticsearch_config_settings.go#L39):

pkg/controller/stackconfigpolicy/controller.go
```diff
@@ -184,9 +199,13 @@ func (r *ReconcileStackConfigPolicy) Reconcile(ctx context.Context, request reco
 // clusters configured by a StackConfigPolicy.
 type esMap map[types.NamespacedName]esv1.Elasticsearch
 
+// kbMap is a type alias for a Map of Kibana indexed by NamespaceName useful to manipulate the Kibana
```
barkbay, line 202:
```suggestion
// kbMap is a type alias for a Map of Kibana indexed by NamespacedName useful to manipulate the Kibana
```

pkg/controller/stackconfigpolicy/controller.go
```diff
@@ -199,19 +218,41 @@ func (r *ReconcileStackConfigPolicy) doReconcile(ctx context.Context, policy pol
...
+	// requeue if not ready
+	if status.Phase != policyv1alpha1.ReadyPhase {
+		results.WithResult(defaultRequeue)
+	}
+
+	return results, status
+}
+
+func (r *ReconcileStackConfigPolicy) reconcileElasticsearchResources(ctx context.Context, policy policyv1alpha1.StackConfigPolicy, status policyv1alpha1.StackConfigPolicyStatus) (*reconciler.Results, policyv1alpha1.StackConfigPolicyStatus) {
+	log := ulog.FromContext(ctx)
```
barkbay, line 251:
```suggestion
	defer tracing.Span(&ctx)()
	log := ulog.FromContext(ctx)
```

pkg/controller/stackconfigpolicy/controller.go
```diff
@@ -333,21 +373,111 @@ func (r *ReconcileStackConfigPolicy) doReconcile(ctx context.Context, policy pol
...
+	results.WithError(handleOrphanSoftOwnedSecrets(ctx, r.Client, k8s.ExtractNamespacedName(&policy), configuredResources, nil, policyv1alpha1.ElasticsearchResourceType))
 
-	// requeue if not ready
-	if status.Phase != policyv1alpha1.ReadyPhase {
-		results.WithResult(defaultRequeue)
+	return results, status
+}
+
+func (r *ReconcileStackConfigPolicy) reconcileKibanaResources(ctx context.Context, policy policyv1alpha1.StackConfigPolicy, status policyv1alpha1.StackConfigPolicyStatus) (*reconciler.Results, policyv1alpha1.StackConfigPolicyStatus) {
+	log := ulog.FromContext(ctx)
```
barkbay, line 395:
```suggestion
	defer tracing.Span(&ctx)()
	log := ulog.FromContext(ctx)
```

pkg/controller/stackconfigpolicy/controller.go
```diff
@@ -405,107 +535,166 @@ func (r *ReconcileStackConfigPolicy) updateStatus(ctx context.Context, scp polic
...
-		"policy_name", scp.Name,
 		"status", status,
 	)
 	scp.Status = status
 	return common.UpdateStatus(ctx, r.Client, &scp)
 }
 
 func (r *ReconcileStackConfigPolicy) onDelete(ctx context.Context, obj types.NamespacedName) error {
-	return handleOrphanSoftOwnedSecrets(ctx, r.Client, obj, nil)
+	// Remove dynamic watches on secrets
```
barkbay, line 545:
```suggestion
	defer tracing.Span(&ctx)()
	// Remove dynamic watches on secrets
```

pkg/controller/stackconfigpolicy/secure_setings.go
```diff
@@ -0,0 +1,57 @@
...
+func GetSecureSettingsSecretSourcesForResources(ctx context.Context, kubeClient k8s.Client, resource metav1.Object, resourceKind string) ([]commonv1.NamespacedSecretSource, error) {
+	switch resourceKind {
+	case "Elasticsearch":
+		return filesettings.GetSecureSettingsSecretSources(ctx, kubeClient, resource)
+	case "Kibana":
+		return getKibanaSecureSettingsSecretSources(ctx, kubeClient, resource)
+	default:
+		// Just return empty since there are no other resource type monitored by the stack config policy
+		log := ulog.FromContext(ctx)
+		log.Info("Unknown resource kind, stackconfigpolicy only monitors Elasticsearch and Kibana resources", "resource kind", resourceKind)
```
barkbay, line 32: Should we return an error here?

pkg/controller/stackconfigpolicy/secure_setings.go
```diff
@@ -0,0 +1,57 @@
...
+
+	rawString, ok := secret.Annotations[commonannotation.SecureSettingsSecretsAnnotationName]
+	if !ok {
+		return []commonv1.NamespacedSecretSource{}, nil
+	}
+	var secretSources []commonv1.NamespacedSecretSource
+	err := json.Unmarshal([]byte(rawString), &secretSources)
+	if err != nil {
+		return nil, err
+	}
```
barkbay, line 54:
```suggestion
	if err := json.Unmarshal([]byte(rawString), &secretSources); err != nil {
		return nil, err
	}
```

barkbay (approved):
LGTM, great job 👍

pkg/controller/elasticsearch/filesettings/reconciler.go
kvalliyurnatt: I made changes to add keys only when data is present.

pkg/controller/stackconfigpolicy/secure_setings.go
kvalliyurnatt: This function is called from `secureSettingsVolume` in the common keystore package, which I think could be called from the apm, beat logstash etc, that is why I just decided to log and return.

PR #7349, opened 2023-11-28
Add retries in TestEnterpriseSearchConfigUpdate steps

thbkrkr:
Resolves https://github.com/elastic/cloud-on-k8s/issues/7348.

PR #7354, opened 2023-11-29
Fix TestEnterpriseSearchConfigUpdate

thbkrkr:
This fixes a regression introduced in #7349.


thbkrkr:
buildkite test this -f p=gke,t=TestEnterpriseSearchConfigUpdate

naemono (approved):
Makes sense from looking at what was changed in #7349 .

Issue #6468, opened 2023-03-01
Elastic agent crashloops after manifest update


boernd:
## Bug Report
**What did you do?**

I have an agent configured in `standalone` mode with the following config:
```
apiVersion: agent.k8s.elastic.co/v1alpha1
kind: Agent
metadata:
  labels:
    app.kubernetes.io/managed-by: pulumi
  name: xyz
  namespace: xyz
spec:
  config:
    inputs:
    - type: apm
      vars:
      - name: host
        value: 0.0.0.0:8200
  deployment:
    podTemplate:
      metadata:
        annotations:
          co.elastic.logs.apm/json.keys_under_root: "true"
          co.elastic.logs.apm/json.message_key: message
      spec:
        containers:
        - name: agent
          resources:
            limits:
              cpu: 500m
              memory: 250Mi
            requests:
              cpu: 150m
              memory: 250Mi
        securityContext:
          runAsUser: 0
        volumes:
[...]
    replicas: 1
  elasticsearchRefs:
  - name: xyz
[...]
  mode: standalone
  version: 8.6.2
```

If I do a change to the Agent manifest (e.g. changing resources) the new pod does not start but ends in a crashloop.
```
Error: another elastic-agent is already running
```
It looks like that this is due to the fact that the `agent-data` volume is host bound. For the APM use case I was able to work around this by overwriting the volume part but for e.g. logs I guess this is problematic:
```
        volumes:
        - emptyDir:
            sizeLimit: 100Mi
          name: agent-data
```

**What did you expect to see?**

Rolling update works.

**What did you see instead? Under which circumstances?**

New pod ends in a crashloop.

**Environment**

* ECK version:

	2.6.1

* Kubernetes information:

	EKS v1.24.8

thbkrkr:
I think it should be more explicit in the doc.

PR #7356, opened 2023-11-30
feat(helm): implementation of integrations chart

pkoutsovasilis:
I open this draft PR to capture the footprint of the changes required to satisfy https://github.com/elastic/elastic-agent/issues/3848 and kickoff further discussions for https://github.com/elastic/elastic-agent/issues/3847. So please don't merge this PR as nothing is concrete yet. With that said if any comments more than welcomed.


deploy/eck-integrations/README.md
```diff
@@ -0,0 +1,58 @@
+# ECK-Integrations
+
+ECK Integrations is a Helm chart to assist in the deployment of Elastic integrations, which are
+managed by the [ECK Operator](https://www.elastic.co/guide/en/cloud-on-k8s/current/index.html)
+
+## Supported Elastic Integrations
+
+The following Elastic integrations are currently supported. 
+
+- Kubernetes
```
joshdover, line 10: Cloud Defend?

pkoutsovasilis: yes you can set the cloudDefend.enabled = true helm chart flag but as of now it only adjusts and redeploys the agent of the daemonset with the appropriate mounts and permissions

deploy/eck-integrations/templates/eck-agent/elastic-agent-daemonset-secret.yaml
```diff
@@ -0,0 +1,41 @@
...
+        enabled: true
+        logs: true
+        metrics: true
+      providers:
+        kubernetes_leaderelection:
+          enabled: false
+        kubernetes:
+          node: ${NODE_NAME}
+          scope: node
+          hints.enabled: false
```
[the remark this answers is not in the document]
pkoutsovasilis: with what I have seen so far, ECK operator utilises the pure k8s Pod Template where init containers belong to so there shouldn't be anything that prohibits this. But I will have another look for sure!

deploy/eck-integrations/templates/eck-agent/_kubernetes/_kubernetes_metrics_apiserver.tpl
```diff
@@ -0,0 +1,41 @@
...
+
+{{/*
+Defaults for kube_apiserver input streams
+*/}}
+{{- define "agent.kubernetes.config.kube_apiserver.defaults" -}}
+hosts:
+{{- range dig "vars" "hosts" (list "https://${env.KUBERNETES_SERVICE_HOST}:${env.KUBERNETES_SERVICE_PORT}") . }}
+- {{. | quote}}
+{{- end }}
+period: {{ dig "vars" "period" "30s" . }}
```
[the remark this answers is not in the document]
pkoutsovasilis: short preview but I will update just one entry with comments in values.yaml to have this there as well

```yaml
kubernetes:
  objects:
    pods:
      state:
        vars:
          add_metadata: false
          hosts:
          - "localhost: 5050"
```

as an example above this will affect only the config of kube_state for pods and apply the values under `vars` instead of the default ones

deploy/eck-integrations/values.yaml
```diff
@@ -0,0 +1,93 @@
+
+
+elasticsearch:
+  host: ""
+  user: ""
+  pass: ""
```
joshdover, line 6: Need support for API keys here

deploy/eck-integrations/values.yaml
```diff
@@ -0,0 +1,93 @@
...
+      enabled: true
+    state:
+      enabled: true
+    logs:
+      enabled: true
+  
+  objects:
+    pods:
+      metrics:
+        enabled: true
```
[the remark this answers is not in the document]
gizas: I think yes that object should be removed. 
My take is that we should move at top level the metrics and logs. 
Then under metrics we should put all the objects that are relevant
Same for logs

So yes we should remove both object and control_plane

deploy/eck-integrations/values.yaml
```diff
@@ -0,0 +1,93 @@
...
+      state:
+        enabled: true
+    deployments:
+      state:
```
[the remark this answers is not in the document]
gizas: I can think only events for now.

Sorry what is state ? :)

joshdover (commented):
Tested and found the following issues on my cluster:
- Many errors from metricbeat in the form of `Error fetching data for metricset kubernetes.system: error doing HTTP request to fetch 'system' Metricset data: error making http request: Get \"https://jophia-01:10250/stats/summary\": lookup jophia-01 on 10.152.183.10:53: no such host`
- `host.name` on logs-elastic_agent.* was being reported as the pod name instead of the node name
- No kubectl or kube-state metrics were being ingested

I added these to the pod spec for the daemonset agent and it resolved most of the issues above:

```
hostNetwork: true
```

What remains after that fix:
- No container logs are ingested - it seems that the `/var/log` volume mounts were applied due to a bug with the `if` condition. I've included a fix for that in the PR linked below. For some reason though, I'm still not getting any files mounted on `/var/log/containers` for a reason I can't yet understand.
- System metrics are missing, but I think this just isn't in the helm chart yet.


Here's a PR with some of the fixes: https://github.com/elastic/cloud-on-k8s/pull/7363

deploy/eck-integrations/templates/eck-agent/elastic-agent-daemonset-secret.yaml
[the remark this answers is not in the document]
gizas: Indeed the hints will need to have the templates mounted to work. 
So we can add a new option in values.yaml sth like install_hint_templates that will install the specific initContainer

deploy/eck-integrations/templates/eck-agent/elastic-agent-daemonset-secret.yaml
```diff
@@ -0,0 +1,41 @@
+apiVersion: v1
+kind: Secret
```
gizas, line 2: Why is this a secret and not configmap? Just for my understanding

joshdover: Yeah it'd be best if we could use both a configmap for most of the config and reference the secrets for the ES connection details using Agent's k8s secret provider.

deploy/eck-integrations/templates/eck-agent/elastic-agent-daemonset-secret.yaml
```diff
@@ -0,0 +1,41 @@
...
+        sourceURI: 'https://artifacts.elastic.co/downloads/'
+      monitoring:
+        namespace: {{ .Values.kubernetes.namespace }}
+        use_output: default
+        enabled: true
+        logs: true
+        metrics: true
+      providers:
+        kubernetes_leaderelection:
+          enabled: false
```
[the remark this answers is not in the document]
gizas: Yes as long as we use only deployment for cluster wide metrics

deploy/eck-integrations/templates/eck-agent/elastic-agent-daemonset.yaml
```diff
@@ -0,0 +1,95 @@
...
+                - BPF # (since Linux 5.8) allows loading of BPF programs, create most map types, load BTF, iterate programs and maps.
+                - PERFMON # (since Linux 5.8) allows attaching of BPF programs used for performance metrics and observability operations.
+                - SYS_RESOURCE # Allow use of special resources or raising of resource limits. Used by 'Defend for Containers' to modify 'rlimit_memlock'
+            {{- end}}
+          resources:
+            limits:
+              memory: 700Mi
+            requests:
+              cpu: 100m
+              memory: 400Mi
```
gizas, line 41: I think also the limits should be configurable with default values

deploy/eck-integrations/templates/eck-agent/elastic-agent-daemonset-secret.yaml
```diff
@@ -0,0 +1,41 @@
...
+          node: ${NODE_NAME}
+          scope: node
+          hints.enabled: false
+          kubernetes_secrets:
+            enabled: true
+    inputs:
+    {{- include "agent.kubernetes.config.audit_logs_filestream.input" .Values.kubernetes | nindent 6 }}
+    {{- include "agent.kubernetes.config.container_logs.input" .Values.kubernetes | nindent 6 }}
+    {{- include "agent.kubernetes.config.kube_controller.input" .Values.kubernetes | nindent 6 }}
+    {{- include "agent.kubernetes.config.kube_scheduler.input" .Values.kubernetes | nindent 6 }}
```
gizas, line 39: There is a conditon here:
            condition: '${kubernetes.labels.component} == ''kube-scheduler'''

deploy/eck-integrations/templates/eck-agent/elastic-agent-daemonset-secret.yaml
```diff
@@ -0,0 +1,41 @@
...
+        kubernetes:
+          node: ${NODE_NAME}
+          scope: node
+          hints.enabled: false
+          kubernetes_secrets:
+            enabled: true
+    inputs:
+    {{- include "agent.kubernetes.config.audit_logs_filestream.input" .Values.kubernetes | nindent 6 }}
+    {{- include "agent.kubernetes.config.container_logs.input" .Values.kubernetes | nindent 6 }}
+    {{- include "agent.kubernetes.config.kube_controller.input" .Values.kubernetes | nindent 6 }}
```
gizas, line 38: Also there is a condition here:
            condition: '${kubernetes.labels.component} == ''kube-controller-manager'''

deploy/eck-integrations/values.yaml
```diff
@@ -0,0 +1,93 @@
...
+
+elasticsearch:
+  host: ""
+  user: ""
+  pass: ""
+
+cloudDefend:
+  enabled: false
+
+kubernetes:
```
gizas, line 11: Why is kubernetes a top level refernece here? If we refer to the package then I think we need to introduce the integration/ pckage as top and then kubernetes

deploy/eck-integrations/values.yaml
```diff
@@ -0,0 +1,93 @@
...
+  host: ""
+  user: ""
+  pass: ""
+
+cloudDefend:
+  enabled: false
+
+kubernetes: 
+  version: 1.51.0
+  enabled: false
```
gizas, line 13: Is this needed?

deploy/eck-integrations/values.yaml
```diff
@@ -0,0 +1,93 @@
...
+    proxy:
+      enabled: false
+    scheduler:
+      enabled: false
+    controller_manager:
+      enabled: false
+    audit_logs_filestream:
+      enabled: false
+
+eck_agent:
```
gizas, line 88: This should move to top

deploy/eck-integrations/values.yaml
```diff
@@ -0,0 +1,93 @@
...
+
+cloudDefend:
+  enabled: false
+
+kubernetes: 
+  version: 1.51.0
+  enabled: false
+  namespace: "default"
+      
+  system:
```
gizas, line 16: Is not this the system integration? If yes should move to the same level as Kubernetes

deploy/eck-integrations/values.yaml
```diff
@@ -0,0 +1,93 @@
...
+    metrics:
```
gizas, line 17: Also does not have logs?

deploy/eck-integrations/values.yaml
```diff
@@ -0,0 +1,93 @@
...
+  control_plane:
+    apiserver:
+      enabled: true
+    proxy:
+      enabled: false
+    scheduler:
+      enabled: false
+    controller_manager:
+      enabled: false
+    audit_logs_filestream:
```
gizas, line 85: This should move under container for sure

deploy/eck-integrations/templates/eck-agent/cluster-role-binding.yaml
```diff
@@ -0,0 +1,53 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
gizas, line 1: This file and cluster-role.yml and service-account.yml should start with `elastic-agent` prefix I think

Maybe we can put all of them inside elastic-agent folder for simplicity and be more clear

gizas:
@pkoutsovasilis really nice start here. For me everything worked right away. Adding some general comments here for further discussion:

-CCing  https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/values.yaml as an example. Shall we start using this naming convention and schema even from day1? I know is a big change for now but maybe worths the effort in order to start being compliant with Otel 
- We should add comments to all fields of value.yaml. Not for this POC of course but for later stable releases
- Adding the idea of schema from https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/values.schema.json. This can help us in some initial tests, validations and linting

- Right now I can not disable kube-state-metrics if  I dont want to. Are we thinking of doing so?
- There are two `fields` that users change: a) logs path and b) scraping period of metrics. Are those part of this POC?

deploy/eck-integrations/templates/eck-agent/elastic-agent-daemonset-secret.yaml
pkoutsovasilis: this is a secret guys because we use `configRef` from ECK operator and this associates the config from a secret with `secretName` property. We could switch to using `config` field of the CRD but this would require embedding the config yaml inside the agent CRD and it imo this would be overloaded. On the plus side having ElasticSearch creds/api_key saved in a secret file is more aligned with k8s practices

deploy/eck-integrations/values.yaml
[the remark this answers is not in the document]
joshdover: > My take is that we should move at top level the metrics and logs.
> Then under metrics we should put all the objects that are relevant
> Same for logs

@mlunadia Any input here? I think the mental model we're expecting users to have is "what do I want to monitor" followed by "which signals do I want for those things". This is what we have in the chart now. To me this makes sense, because as a user I'm thinking about the thing that is broken first (eg. pods) and then second which data I need to find the root cause of the issue.

@gizas, your preference is the other way: "which signals do I want" followed by "for which things do I want those signals".

deploy/eck-integrations/templates/eck-agent/elastic-agent-daemonset.yaml
```diff
@@ -0,0 +1,95 @@
...
+          env:
+          - name: NODE_NAME
+            valueFrom:
+              fieldRef:
+                fieldPath: spec.nodeName
```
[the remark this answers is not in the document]
gizas: Good catch. I think is needed only for security but does not harm to add it.

Please also **NETINFO** should be added:
https://github.com/elastic/elastic-agent/blob/main/deploy/kubernetes/elastic-agent-standalone-kubernetes.yaml#L732

deploy/eck-integrations/values.yaml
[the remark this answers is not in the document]
gizas: Yes indeed @joshdover . I am thinking also this from the UX, that if they (or dont) want  metrics they just disable one flag. The other way the need to go and disable all metirc references

deploy/eck-integrations/values.yaml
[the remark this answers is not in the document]
gizas: Hmm so even if we include the metrics under each object I think that a more suitable approach is:

```yaml
pods:
      metrics:
        enabled: true
        kubelet: 
            enabled: true
        kubestate:
           enabled: true
```
What do you think?

deploy/eck-integrations/values.yaml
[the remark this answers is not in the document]
pkoutsovasilis: my 2 cents, I resonate with @joshdover on the approach `"what do I want to monitor"`->`"which signals do I want for those things"`. But I hear your point as well @gizas, and we could support such setting by defining another field under `kubernetes`, e.g. `skip_metrics`, etc. that can act as an override knob for all other (sub-)metrics keys.

deploy/eck-integrations/values.yaml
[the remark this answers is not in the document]
joshdover: +1 on having the increasing levels of granularity. Only nit on your example is I think that maybe we should avoid using `kubestate` as it's specific to a specific implementation of these types of metrics. For instance, I don't think OTel Collector relies on these metrics (or they're missing - still tbd)

deploy/eck-integrations/templates/eck-agent/elastic-agent-daemonset.yaml
```diff
@@ -0,0 +1,95 @@
...
+        - name: var-lib
+          mountPath: /hostfs/var/lib
+          readOnly: true
+        - name: sys-kernel-debug
+          mountPath: /sys/kernel/debug
+        - name: elastic-agent-state
+          mountPath: /usr/share/elastic-agent/state
+        {{- end}}
+        {{- if eq .Values.cloudDefend.enabled true}}
+        volumes:
```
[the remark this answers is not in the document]
joshdover: @pkoutsovasilis I think we should use a different hostPath for this chart than is used in the other Agent manifests. That way if a user upgrades from managed to standalone agent (in this helm chart) the state gets reset.

Maybe `/var/lib/elastic-agent-eck/elastic-system/state`?

deploy/eck-integrations/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+# Install the ECK Operator cluster-wide
+helm install elastic-operator elastic/eck-operator -n elastic-system --create-namespace
+```
+
+Additional ECK Operator Helm installation options can be found within [our documentation](https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-install-helm.html)
+
+### Installing Kubernetes ECK Integration
+
+// TODO(panosk)
```
joshdover, line 40: Some possible examples to put here:

```suggestion

To enable all Kubernetes monitoring capabilities for metrics and logs:

helm install k8s-monitoring elastic/eck-integrations -n elastic-system --create-namespace \
  --set elasticsearch.host=https://elasticsearch:9200,elasticsearch.api_key=XXXX \
  --set kubernetes.enabled=true

To only enable logs:

helm install k8s-monitoring elastic/eck-integrations -n elastic-system --create-namespace \
  --set elasticsearch.host=https://elasticsearch:9200,elasticsearch.api_key=XXXX \
  --set kubernetes.enabled=true,kubernetes.metrics.enabled=false
  
To enable Cloud Defend:

helm install k8s-monitoring elastic/eck-integrations -n elastic-system --create-namespace \
  --set elasticsearch.host=https://elasticsearch:9200,elasticsearch.api_key=XXXX \
  --set cloudDefend.enabled=true
```

deploy/eck-integrations/values.yaml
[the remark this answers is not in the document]
joshdover: +1 on the override knob. I think it's reasonable to think that users may only want logs enabled but not metrics, for example. There should be an easy way to accomplish that.

Maybe just an overall `kubernetes.metrics.enabled` flag that takes precedence when it's `false`. Same for logs (and later traces).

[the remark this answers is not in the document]
mlunadia: > Any input here? I think the mental model we're expecting users to have is "what do I want to monitor" followed by "which signals do I want for those things". This is what we have in the chart now. To me this makes sense, because as a user I'm thinking about the thing that is broken first (eg. pods) and then second which data I need to find the root cause of the issue.

1. System that needs to be monitored: Kubernetes, nginx, redis, etc...
2. Resources in that system: Kubernetes nodes, pods, containers...
3. Signals: Logs, metrics, traces
4. Config: eg: Location> endpoint, directory, etc...

I'm in two minds on whether signals are nested into resources or the opposite but prefer it as above with an override to reduce user pain but I'm currently unsure how this could be done in yaml.

Have we considered ideas for how are we going to serve the corresponding values.yaml to users when we have 300+ integrations? A single values.yaml would be impossible to work with.

deploy/eck-integrations/values.yaml
mlunadia: We can't create top level references for specific integrations. If cloudDefend is the integration that we are going to install then this value should only surface on the corresponding values.yaml same for Kubernetes being a top-level reference

deploy/eck-integrations/values.yaml
mlunadia: Bitnami are a good point of reference for the above https://github.com/bitnami/charts/blob/main/bitnami/elasticsearch/values.yaml

deploy/eck-integrations/values.yaml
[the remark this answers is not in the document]
joshdover: > Have we considered ideas for how are we going to serve the corresponding values.yaml to users when we have 300+ integrations? A single values.yaml would be impossible to work with.

@mlunadia is this resolved by our discussion earlier? To summarize we said:
- Metrics integrations would be supported via hints-based autodiscovery
- Most log integrations are supported via routing container logs in ingest pipelines to more specific integrations. No Agent configuration is required. This is essentially how logs integrations work on k8s today.

I'd also add that if there are other integrations that users want to deploy on K8s, they can set up a managed agent to do that with just a regular ECK Agent resource, separate from the agent pods that are monitoring the k8s infrastructure and containers. The reason for this is that deployment mode, resource limits, and scaling of that integration have different properties than the integrations running on the main agent nodes doing k8s monitoring.

For example, if a user wants to use an AWS integration to pull data from S3, they can either create an ECK Agent CR with a Fleet enrollment token to a policy that has this integration, or they can create an [ECK Agent CR with the input configuration](https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-elastic-agent-quickstart.html) copied from the Integrations UI. They can then scale and manage that agent separate from the core agents monitoring k8s.

deploy/eck-integrations/values.yaml
joshdover: Yeah we'll definitely do that type of clean up - right now we're focused on a functional POC

deploy/eck-integrations/templates/eck-agent/elastic-agent-daemonset-secret.yaml
joshdover: Fine by me

deploy/eck-integrations/values.yaml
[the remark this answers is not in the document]
joshdover: If we want to simplify things for users so they only have to manage the helm chart and nothing else, we could add support for additional agents that get deployed separately. We wouldn't need to include every integration in the `values.yml` but instead have users copy the config from the Integrations UI for the integrations they want OR use a Fleet-managed agent.

So we could have something like this in the values.yml which is essentially just a list of Agent CRs:

```yaml
additionalAgents:
# key is `metadata.name` for the Agent CR
myAgent:
  # contents are the same as `spec` in the Agent CR
  version: 8.11.1
  deployment: {}
  config:
    inputs:
    # copy/pasted inputs from Integrations UI
```

deploy/eck-integrations/values.yaml
pkoutsovasilis: nope this `system` is this one https://docs.elastic.co/integrations/kubernetes/kubelet#system

deploy/eck-integrations/templates/eck-agent/cluster-role-binding.yaml
pkoutsovasilis: they are kinda already are inside the `eck-agent` folder 👍

deploy/eck-integrations/values.yaml
```diff
@@ -0,0 +1,348 @@
...
+## @param elasticsearch.host [default: ""] Elasticsearch host
+## @param elasticsearch.user [default: ""] Elasticsearch username
+## @param elasticsearch.pass [default: ""] Elasticsearch password
+## @param elasticsearch.apiKey [default: ""] Elasticsearch access key
+## You may specify an access key instead of username and password
+elasticsearch:
+  host: ""
+  user: ""
+  pass: ""
+  apiKey: ""
```
thbkrkr, line 11: A user may be required to provide a custom CA. Also, it's not great to have to configure sensitive data via Helm values. It would be nice to provide an alternative to avoid this.

If the Elasticsearch storing monitoring data is managed by ECK in the same k8s cluster, then ECK knows how to get all info about it, it just needs the namespace/name. Otherwise for any Elasticsearch, we could use a Secret.

The Elastic `Agent` custom resource has an `elasticsearchRefs` field of `ObjectSelector` type that could help as it supports a custom CA and the 2 usecases (ES managed by ECK and Secret to describe any ES), see the [ObjectSelector doc](https://www.elastic.co/guide/en/cloud-on-k8s/2.11/k8s-api-common-k8s-elastic-co-v1.html#k8s-api-github-com-elastic-cloud-on-k8s-v2-pkg-apis-common-v1-objectselector) and by using it, ECK configures the outputs in the `agent.yml` for you.

```yaml
kind: Agent
metadata:
  name: eck-agent-deployment
spec:
  version: 8.12.0

  # Elasticsearch managed by ECK
  elasticsearchRefs:
  - name: es-monitoring
    namespace: observability

  # Any Elasticsearch cluster
  # The referenced secret must contain the following:
  #  - `url`: the URL to reach the Elastic resource
  #  - `username`: the username of the user to be authenticated to the Elastic resource
  #  - `password`: the password of the user to be authenticated to the Elastic resource
  #  - `ca.crt`: the CA certificate in PEM format (optional).
  elasticsearchRefs:
  - secretName: es-monitoring-ref
```

pkoutsovasilis: thank you @thbkrkr this is more than a valid point and I like the approach of `elasticsearchRefs` for the reasons you mention but also because through that it is easier to support multiple outputs. One thing that I think it is not supported as of now and we need it is being able to utilise an `apikey` and not a pair of `(username, password)`

pkoutsovasilis: here is a [relevant PR](https://github.com/elastic/cloud-on-k8s/pull/7598) that introduces support for `apikey` only for `Agent->Unmanaged Elastic Search` association

deploy/eck-integrations/values.yaml
pkoutsovasilis: this system referes to this one https://github.com/elastic/integrations/blob/main/packages/kubernetes/data_stream/system/manifest.yml which if I am not mistaken there are no logs for it

deploy/eck-integrations/values.yaml
pkoutsovasilis: hmm after running some test I have to say apparently it is not 😕

deploy/eck-integrations/values.yaml
pkoutsovasilis: I think that this was already covered in one of our meetings, and as we discussed this helm chart is integration-focused thus the integrations are at the top level and then the helm chart templates transform all integrations and tight them to the specific agents. Please if you consider this as unresolved feel free to re-open 🙂

deploy/eck-integrations/templates/eck-agent/elastic-agent-daemonset-secret.yaml
```diff
@@ -0,0 +1,41 @@
...
+        username: {{ .Values.elasticsearch.user | quote  }}
+        password: {{ .Values.elasticsearch.pass | quote  }}
+    secret_references: []
+    agent:
+      download:
+        sourceURI: 'https://artifacts.elastic.co/downloads/'
+      monitoring:
+        namespace: {{ .Values.kubernetes.namespace }}
+        use_output: default
+        enabled: true
```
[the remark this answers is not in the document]
pkoutsovasilis: this is not configurable in values.yaml undert eck_agent.presets.presetName, and the user can change it at will

deploy/eck-integrations/templates/eck-agent/elastic-agent-daemonset.yaml
pkoutsovasilis: this is now configurable in values.yaml under eck_agent.presets.presetName, and the user can change it at will

deploy/eck-integrations/values.yaml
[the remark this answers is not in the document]
pkoutsovasilis: with the latest iteration this is now supported, users can specify custom integrations and tight them to custom agent presets [example](https://github.com/elastic/cloud-on-k8s/pull/7356/files#diff-8b91b0bf85c2ba0d1636bacc927e5b6952590b9aacecebcdaa9f2087bdbaa2cf).

deploy/eck-integrations/templates/eck-agent/elastic-agent-daemonset.yaml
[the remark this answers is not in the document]
pkoutsovasilis: good catch @joshdover! this is now fixed and the state of agents with `daemonset` mode is managed by ECK `/var/lib/elastic-agent/$namespace/$agent_name`. For the agents under `statefulset` or `daemonset` the state is set as an `EmptyDir` to avoid problems like this [one](https://github.com/elastic/cloud-on-k8s/issues/6468)

deploy/eck-integrations/values.yaml
```diff
@@ -0,0 +1,493 @@
...
+      agent:
+        monitoring:
+          namespace: default
+          use_output: default
+          enabled: true
+          logs: true
+          metrics: true
+      providers:
+        kubernetes_leaderelection:
+          enabled: false
```
cmacknz, line 491: This works to disable leader election properly via the agent CRD's config section?  Does this work to override the initial configurations for Fleet managed agents as well?

If so it is significantly easier than the hoops users are jumping through to do this without ECK today, see https://www.elastic.co/guide/en/fleet/current/advanced-kubernetes-managed-by-fleet.html.

cmacknz (commented):
I tried this following the multiple integration example and it worked as expected. 

I think it would be nice to have an example of how setup an equivalent set of Fleet managed agents with the separate daemonset and deployment with the initial agent container configuration overridden (e.g. with leader election disabled everywhere). Perhaps this is intended as a follow up, but users regularly ask for this and it would have simplified several recent support cases.

The only trouble I ran into was when experimenting with the `es-ref-secret`. If you omit the expected values (for example trying to use the not yet implemented `api-key` instead of `password`) the operator doesn't create anything and there's no obvious error. 

You can figure out what happened by looking at the operator logs, but users coming from the current deployment pattern aren't going to expect this or know to look for it. Is there a way to make these failures more immediately obvious? Or is this something we have to solve with documentation?

For example this is the error message the controller gave me, which was very clear, but you have to know to look at the controller logs to get it.

```json
{"log.level":"error","@timestamp":"2024-04-05T19:23:47.829Z","log.logger":"manager.eck-operator","message":"Reconciler error","service.version":"2.12.1+a56215f7","service.type":"eck","ecs.version":"1.4.0","controller":"agent-es-association-controller","object":{"name":"agent-pernode","namespace":"default"},"namespace":"default","name":"agent-pernode","reconcileID":"5ee3a9cb-cafb-4aaa-af40-d8a2e221711f","error":"username secret key doesn't exist in secret es-ref-secret","errorCauses":[{"error":"username secret key doesn't exist in secret es-ref-secret"}],"error.stack_trace":"sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.17.2/pkg/internal/controller/controller.go:329\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.17.2/pkg/internal/controller/controller.go:266\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.17.2/pkg/internal/controller/controller.go:227"}
```

deploy/eck-integrations/values.yaml
pkoutsovasilis: essentially via configRef which a secret that the chart populates accordingly and then the operator parses it and constructs properly the agent.yml but I think we mean the same here 🙂 I am not fully aware of how something similar could like for fleet-managed agents but definitely we can have a look

pkoutsovasilis:
> I tried this following the multiple integration example and it worked as expected.
> 
> I think it would be nice to have an example of how setup an equivalent set of Fleet managed agents with the separate daemonset and deployment with the initial agent container configuration overridden (e.g. with leader election disabled everywhere). Perhaps this is intended as a follow up, but users regularly ask for this and it would have simplified several recent support cases.
> 
> The only trouble I ran into was when experimenting with the `es-ref-secret`. If you omit the expected values (for example trying to use the not yet implemented `api-key` instead of `password`) the operator doesn't create anything and there's no obvious error.
> 
> You can figure out what happened by looking at the operator logs, but users coming from the current deployment pattern aren't going to expect this or know to look for it. Is there a way to make these failures more immediately obvious? Or is this something we have to solve with documentation?
> 
> For example this is the error message the controller gave me, which was very clear, but you have to know to look at the controller logs to get it.
> 
> ```json
> {"log.level":"error","@timestamp":"2024-04-05T19:23:47.829Z","log.logger":"manager.eck-operator","message":"Reconciler error","service.version":"2.12.1+a56215f7","service.type":"eck","ecs.version":"1.4.0","controller":"agent-es-association-controller","object":{"name":"agent-pernode","namespace":"default"},"namespace":"default","name":"agent-pernode","reconcileID":"5ee3a9cb-cafb-4aaa-af40-d8a2e221711f","error":"username secret key doesn't exist in secret es-ref-secret","errorCauses":[{"error":"username secret key doesn't exist in secret es-ref-secret"}],"error.stack_trace":"sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.17.2/pkg/internal/controller/controller.go:329\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.17.2/pkg/internal/controller/controller.go:266\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.17.2/pkg/internal/controller/controller.go:227"}
> ```
 
thanks for trying this out @cmacknz. The case you stumbled upon is coming from the operator itself. You are right in terms of plain k8s objects nothing appears to be created but Agent CRDs (this is what we deploy here) should be there and their status should indicate that something is happening but not getting ready. With that said ofc we will fortify this through documentation and have a look around how we could make such cases more "visible"

cmacknz:
> Agent CRDs (this is what we deploy here) should be there and their status should indicate that something is happening but not getting ready. 

Yes I actually forgot to check the agent CRDs and was just doing `kubectl get pods` when working since I'm not used to working with ECK by reflex yet.

deploy/eck-integrations/values.yaml
cmacknz: What I hope happens is the same mechanism overrides the initial configuration of the agent to seed it with the providers disabled, and then Fleet takes over from that starting base configuration.

The providers can't be configured from Fleet right now so this is a pain point.

cmacknz: Also, with the Helm chart being able to create both a Deployment and DaemonSet I don't think we need leaderelection at all anymore so it should be disabled everywhere by default (pending a closer review of the configurations to make sure we did this right).

The only reason leader election exists is so you can just deploy a DaemonSet and emulate having a dedicated deployment to simplify the configuration needed (AFAIK).

deploy/eck-integrations/values.yaml
```diff
@@ -0,0 +1,493 @@
...
+  ##        - `password`: the password of the user to be authenticated to the Elastic resource.
+  ##        - `ca.crt`: the CA certificate in PEM format (optional).
+  ##        - `api-key`: the key to authenticate against the Elastic resource instead of a username and password
+  ##                     (supported only for `elasticsearchRefs` in AgentSpec). This field cannot be used in
+  ##                     combination with the other fields name, namespace or serviceName.
+  ## E.g.
+  ## default:
+  ##   secretName: es-ref-secret [for *non* eck-managed elasticsearch]
+  ##   name: es-ref-name [for eck-managed elasticsearch]
+  ##   namespace: es-ref-namespace [for eck-managed elasticsearch]
```
pebrc, line 21: Should we have a default here? or at least an empty map so that the template renders?

deploy/eck-integrations/examples/multiple-integrations/README.md
```diff
@@ -0,0 +1,54 @@
...
+    kubectl create secret generic es-ref-secret \
+        --from-literal=username=... \
+        --from-literal=password=... \
+        --from-literal=url=...
+    ```
+    Note: specifying an `api-key`, instead of a `username` and `password`, is not supported at the moment but there is an already open PR to add support for it.
+
+3. `kubernetes` integration Assets are installed through Kibana
+4. `redis` integration Assets are installed through Kibana
+5. `nginx` integration Assets are installed through Kibana
```
pebrc, line 22:
```suggestion
4. `redis` integration assets are installed through Kibana
5. `nginx` integration assets are installed through Kibana
```
how?

deploy/eck-integrations/examples/nginx-custom-integration/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+In this example we define a `nginx` custom integration bound against a custom agent preset as shown in [agent-nginx.yaml](agent-nginx.yaml).
+
+## Prerequisites:
+1. Installed eck-operator helm chart
+   ```console
+   helm repo add elastic https://helm.elastic.co && helm repo update
+   helm install elastic-operator elastic/eck-operator --create-namespace
+   ```
+2. For **non** eck-managed ElasticSearch clusters you need a k8s secret that contains the connection details to it such as:
```
pebrc, line 11:
```suggestion
2. For **non** eck-managed Elasticsearch clusters you need a k8s secret that contains the connection details to it such as:
```

deploy/eck-integrations/examples/nginx-custom-integration/README.md
```diff
@@ -0,0 +1,43 @@
...
+3. `nginx` integration Assets are installed through Kibana
+
+## Run:
+1. For **non** eck-managed ElasticSearch clusters
+    ```console
+    helm install eck-integrations ../../ \
+         -f ./agent-nginx.yaml \
+         --set elasticsearchRefs.default.secretName=es-ref-secret 
+    ```
+    For eck-managed ElasticSearch clusters
```
pebrc, line 29:
```suggestion
    For eck-managed Elasticsearch clusters
```

deploy/eck-integrations/examples/nginx-custom-integration/README.md
```diff
@@ -0,0 +1,43 @@
...
+    ```
+
+2. Install the nginx deployment
+    ```console
+   kubectl apply -f ./nginx.yaml
+    ```
+   
+## Validate:
+
+1. The Kibana `nginx`-related dashboards should start showing up the respective info.
```
pebrc, line 43:
```suggestion
1. The Kibana `nginx`-related dashboards should start showing nginx related data.
```

test/e2e/helm/integrations/builder.go
```diff
@@ -0,0 +1,297 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License 2.0;
+// you may not use this file except in compliance with the Elastic License 2.0.
+
+//go:build chart_integrations || e2e
```
pebrc, line 5:
```suggestion
//go:build helm || e2e
```
should this just be a tag to target the helm chart tests in general?

test/e2e/helm/integrations/chart_test.go
```diff
@@ -0,0 +1,51 @@
...
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"k8s.io/apimachinery/pkg/util/rand"
+
+	"github.com/elastic/cloud-on-k8s/v2/test/e2e/test"
+)
+
+func TestDefaultKubernetes(t *testing.T) {
```
pebrc, line 19:
```suggestion
func TestIntegrationsChartDefaultKubernetes(t *testing.T) {
```

deploy/eck-integrations/examples/nginx-custom-integration/README.md
pkoutsovasilis: I always do that typo 🥲

deploy/eck-integrations/values.yaml
```diff
@@ -0,0 +1,493 @@
+## @section elasticsearchRef define outputs of elasticsearch type and connection parameters
```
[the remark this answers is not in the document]
pkoutsovasilis: this was inspired by bitnami charts, e.g. [this](https://github.com/bitnami/charts/blob/main/bitnami/apache/values.yaml#L4). In theory, hjaving such comment metadata should help in autogenerating the documentation for the values.yaml. That said, I am not sure if it adds value for us here

deploy/eck-integrations/examples/kubernetes-default/README.md
```diff
@@ -0,0 +1,39 @@
...
+   helm install elastic-operator elastic/eck-operator --create-namespace
+   ```
+2. For **non** eck-managed ElasticSearch clusters you need a k8s secret that contains the connection details to it such as:
+    ```console
+    kubectl create secret generic es-ref-secret \
+        --from-literal=username=... \
+        --from-literal=password=... \
+        --from-literal=url=...
+    ```
+    Note: specifying an `api-key`, instead of a `username` and `password`, is not supported at the moment but there is an already open PR to add support for it.
```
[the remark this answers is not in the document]
pkoutsovasilis: oh yes you are 100% right

deploy/eck-integrations/values.yaml
pkoutsovasilis: to be honest I was aiming for the template not to render if this is empty 😁 this is kinda crucial for defining the output and only the user knows the details of it. But if you have a better default idea here more than happy to incorporate it

deploy/eck-integrations/examples/multiple-integrations/README.md
pkoutsovasilis: thx for the question. Added links that show how to achieve this through Kibana 🙂

test/e2e/helm/integrations/builder.go
pkoutsovasilis: yep makes total sense

deploy/eck-integrations/examples/nginx-custom-integration/README.md
```diff
@@ -0,0 +1,43 @@
+# Example: Nginx Custom Integration
+
+In this example we define a `nginx` custom integration bound against a custom agent preset as shown in [agent-nginx.yaml](agent-nginx.yaml).
```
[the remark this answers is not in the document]
pkoutsovasilis: Gave it one more try to explain what I mean here 🙂

thbkrkr:
builkite test this -f p=gke,t=TestIntegrationsChartDefaultKubernetes,s=8.13.2

pkoutsovasilis:
sorry for keeping this open for so long, but based on some internal discussions it was decided that this Helm chart will found its home in `elastic-agent` repo. Expect that, ECK operator will still be supported by the "new" helm chart there, but there will be also an option to deploy elastic-agent as plain k8s objects. Thank you for all the help @thbkrk and @pebrc 🙏

PR #7357, opened 2023-12-01
Add StatefulSet as a deployment option for Elastic Agent

pkoutsovasilis:
This PR enriches the possible types of deploying agent by adding support for statefulSet. A use-case, that this PR adds support for, is deploying ksm with sharding (requires statefulset) with agent containers that extract only kube-state metrics, e.g. as being done [here](https://github.com/elastic/platform-observability-charts/blob/main/charts/elastic-agent/templates/ksm-elastic-agent-statefulset.yaml). Having this feature will also benefit this one https://github.com/elastic/elastic-agent/issues/3847.


barkbay (commented):
> A use-case, that this PR adds support for, is deploying ksm with sharding (requires statefulset) with agent containers that extract only kube-state metrics,

Do you think it would be possible to add an example in `config/recipes/elastic-agent`?

Also I believe `GetPodTemplate()` is not tested, but it's not from your PR. It could be interesting to check somewhere that the expected "native" resource (`daemonset`, `deployment` or `statefulet`) is created using the expected template.

pkg/controller/common/statefulset/reconcile.go
```diff
@@ -0,0 +1,107 @@
...
+			ServiceName:          params.ServiceName,
+			RevisionHistoryLimit: params.RevisionHistoryLimit,
+			// always RollingUpdate strategy as OnDelete would require the user
+			// to delete pods and this is not fully with the operator reconcile logic.
+			UpdateStrategy: appsv1.StatefulSetUpdateStrategy{
+				Type:          appsv1.RollingUpdateStatefulSetStrategyType,
+				RollingUpdate: nil,
+			},
+			// default statefulSet behaviour
+			PodManagementPolicy: appsv1.OrderedReadyPodManagement,
```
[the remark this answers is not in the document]
pkoutsovasilis: thanks for the comment @barkbay, tbh I went with the `OrderedReadyPodManagement` as it is the default one for `statefulset` and because the latter is mostly used for the constant pod names and the ordered pod management so you can update/rollback without downtime between versions. But I see your point, maybe have that as an option in the spec?

pkoutsovasilis:
> > A use-case, that this PR adds support for, is deploying ksm with sharding (requires statefulset) with agent containers that extract only kube-state metrics,
> 
> Do you think it would be possible to add an example in `config/recipes/elastic-agent`?
> 
> Also I believe `GetPodTemplate()` is not tested, but it's not from your PR. It could be interesting to check somewhere that the expected "native" resource (`daemonset`, `deployment` or `statefulet`) is created using the expected template.

@barkbay I added a recipe for ksm-sharding specifically that utilises the introduced statefulset type of the agent. Would you like me to try to tackle `GetPodTemplate()` in this PR?

pkg/controller/common/statefulset/reconcile.go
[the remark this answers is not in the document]
barkbay: My gut feeling is that unless it is necessary to start the Pods in a specific order I would go for `Parallel`, but I have to admit that I have no other argument than that this would allow Pods to start more quickly. 😄 

> But I see your point, maybe have that as an option in the spec?

👍

pkg/controller/agent/reconcile.go
```diff
@@ -36,25 +37,51 @@ func reconcilePodVehicle(params Params, podTemplate corev1.PodTemplateSpec) (*re
...
+				Name:      name,
+				Namespace: params.Agent.Namespace,
+			},
+		})
+		toDelete = append(toDelete, &v1.Deployment{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      name,
+				Namespace: params.Agent.Namespace,
+			},
+		})
```
barkbay, line 84: nit: `append`  is a variadic function
```suggestion
		toDelete = append(toDelete,
			&v1.DaemonSet{
				ObjectMeta: metav1.ObjectMeta{
					Name:      name,
					Namespace: params.Agent.Namespace,
				},
			},
			&v1.Deployment{
				ObjectMeta: metav1.ObjectMeta{
					Name:      name,
					Namespace: params.Agent.Namespace,
				},
			},
		)
```

config/recipes/elastic-agent/ksm-sharding.yaml
```diff
@@ -0,0 +1,471 @@
...
+            valueFrom:
+              fieldRef:
+                fieldPath: spec.nodeName
+          - name: POD_NAME
+            valueFrom:
+              fieldRef:
+                fieldPath: metadata.name
+  config:
+    id: 488e0b80-3634-11eb-8208-57893829af4e
+    revision: 2
```
barkbay, line 73: Can we remove these 2 lines? If not, maybe add a comment so our users know how to handle them?

pkg/controller/agent/reconcile.go
pkoutsovasilis: oh nice catch! sure this makes 100% sense and I can't tell how I missed it 😃

config/recipes/elastic-agent/ksm-sharding.yaml
pkoutsovasilis: this came from an existing agent recipe where I drew inspiration of [here](https://github.com/elastic/cloud-on-k8s/blob/main/config/recipes/elastic-agent/kubernetes-integration.yaml#L24). `revision` is not mandatory and can be removed and the same goes for `id` I think

pkoutsovasilis: ok I tested locally without these two fields and everything works as expected so I remove them

pkg/controller/common/statefulset/reconcile.go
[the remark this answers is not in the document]
pkoutsovasilis: @barkbay I pushed an update, now `PodManagementPolicy` defaults to `Parallel` but can be specified through yaml to `OrderedReadyPodManagement`

pkg/apis/agent/v1alpha1/validations.go
```diff
@@ -70,12 +71,25 @@ func checkPolicyID(a *Agent) field.ErrorList {
 }
 
 func checkAtMostOneDeploymentOption(a *Agent) field.ErrorList {
```
barkbay, line 73: Maybe add a quick unit test:

```golang
func Test_checkAtMostOneDeploymentOption(t *testing.T) {
	type args struct {
		a *Agent
	}
	tests := []struct {
		name string
		args args
		want field.ErrorList
	}{
		{
			args: args{
				a: &Agent{Spec: AgentSpec{
					StatefulSet: &StatefulSetSpec{},
				}},
			},
		},
		{
			args: args{
				a: &Agent{Spec: AgentSpec{
					Deployment:  &DeploymentSpec{},
					StatefulSet: &StatefulSetSpec{},
				}},
			},
			want: field.ErrorList{
				field.Forbidden(field.NewPath("spec.deployment"), "Specify at most one of [deployment, statefulSet]"),
				field.Forbidden(field.NewPath("spec.statefulSet"), "Specify at most one of [deployment, statefulSet]"),
			},
		},
		{
			args: args{
				a: &Agent{Spec: AgentSpec{
					Deployment: &DeploymentSpec{},
					DaemonSet:  &DaemonSetSpec{},
				}},
			},
			want: field.ErrorList{
				field.Forbidden(field.NewPath("spec.daemonSet"), "Specify at most one of [daemonSet, deployment]"),
				field.Forbidden(field.NewPath("spec.deployment"), "Specify at most one of [daemonSet, deployment]"),
			},
		},
		{
			args: args{
				a: &Agent{Spec: AgentSpec{
					Deployment:  &DeploymentSpec{},
					DaemonSet:   &DaemonSetSpec{},
					StatefulSet: &StatefulSetSpec{},
				}},
			},
			want: field.ErrorList{
				field.Forbidden(field.NewPath("spec.daemonSet"), "Specify at most one of [daemonSet, deployment, statefulSet]"),
				field.Forbidden(field.NewPath("spec.deployment"), "Specify at most one of [daemonSet, deployment, statefulSet]"),
				field.Forbidden(field.NewPath("spec.statefulSet"), "Specify at most one of [daemonSet, deployment, statefulSet]"),
			},
		},
	}
	for _, tt := range tests {
			if got := checkAtMostOneDeploymentOption(tt.args.a); !reflect.DeepEqual(got, tt.want) {
				t.Errorf("checkAtMostOneDeploymentOption() = \n%v, \nwant \n%v", got, tt.want)
			}
		})
	}
}
```

pkg/apis/agent/v1alpha1/agent_types.go
```diff
@@ -127,6 +132,32 @@ type DeploymentSpec struct {
...
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Enum=OrderedReady;Parallel
+	// +kubebuilder:default:=Parallel
+	PodManagementPolicy appsv1.PodManagementPolicyType `json:"podManagementPolicy,omitempty"`
+
+	// VolumeClaimTemplates is a list of persistent volume claims to be used by each Pod.
+	// Every claim in this list must have a matching volumeMount in one of the containers defined in the PodTemplate.
+	// Items defined here take precedence over any default claims added by the operator with the same name.
+	// +kubebuilder:validation:Optional
+	VolumeClaimTemplates []corev1.PersistentVolumeClaim `json:"volumeClaimTemplates,omitempty"`
```
barkbay, line 158: I wonder if we want to validate that there is no change applied to volume templates, as it's done for Logstash, something along:

```diff
diff --git a/pkg/apis/agent/v1alpha1/validations.go b/pkg/apis/agent/v1alpha1/validations.go
index 12c59c326..4f5f1365a 100644
--- a/pkg/apis/agent/v1alpha1/validations.go
+++ b/pkg/apis/agent/v1alpha1/validations.go
@@ -9,6 +9,7 @@ import (
        "reflect"
        "strings"
 
+       apiequality "k8s.io/apimachinery/pkg/api/equality"
        "k8s.io/apimachinery/pkg/util/validation/field"
 
        commonv1 "github.com/elastic/cloud-on-k8s/v2/pkg/apis/common/v1"
@@ -37,6 +38,7 @@ var (
 
        updateChecks = []func(old, curr *Agent) field.ErrorList{
                checkNoDowngrade,
+               checkPVCChanges,
        }
 )
 
@@ -137,6 +139,23 @@ func checkNoDowngrade(prev, curr *Agent) field.ErrorList {
        return commonv1.CheckNoDowngrade(prev.Spec.Version, curr.Spec.Version)
 }
 
+const (
+       pvcImmutableMsg = "Volume claim templates cannot be modified"
+)
+
+// checkPVCChanges ensures no PVCs are changed, as volume claim templates are immutable in StatefulSets.
+func checkPVCChanges(current, proposed *Agent) field.ErrorList {
+       var errs field.ErrorList
+       if current == nil || proposed == nil || current.Spec.StatefulSet == nil || proposed.Spec.StatefulSet == nil {
+               return errs
+       }
+       if !apiequality.Semantic.DeepEqual(current.Spec.StatefulSet.VolumeClaimTemplates, proposed.Spec.StatefulSet.VolumeClaimTemplates) {
+               errs = append(errs, field.Invalid(field.NewPath("spec").Child("volumeClaimTemplates"), proposed.Spec.StatefulSet.VolumeClaimTemplates, pvcImmutableMsg))
+       }
+
+       return errs
+}
+
 func checkSingleConfigSource(a *Agent) field.ErrorList {
        if a.Spec.Config != nil && a.Spec.ConfigRef != nil {
                msg := "Specify at most one of [`config`, `configRef`], not both"
```

barkbay (dismissed):
LGTM

pkg/apis/agent/v1alpha1/agent_types.go
pkoutsovasilis: oh nice catch, yes I think we want to validate this one. Only one minor change in the proposed commit above, I think we need to check if both current and proposed are both with a `Spec` of `StatefulSet`. Otherwise, we shouldn't check them

pkg/apis/agent/v1alpha1/validations.go
```diff
@@ -137,6 +143,27 @@ func checkNoDowngrade(prev, curr *Agent) field.ErrorList {
...
+	}
+
+	// need to check if current and proposed are both statefulsets
+	if current.Spec.StatefulSet == nil || proposed.Spec.StatefulSet == nil {
+		return errs
+	}
+
+	// checking semantic equality here allows providing PVC storage size with different units (eg. 1Ti vs. 1024Gi).
+	if !apiequality.Semantic.DeepEqual(current.Spec.StatefulSet.VolumeClaimTemplates, proposed.Spec.StatefulSet.VolumeClaimTemplates) {
+		errs = append(errs, field.Invalid(field.NewPath("spec").Child("volumeClaimTemplates"),
```
barkbay, line 160: nit:
```suggestion
		errs = append(errs, field.Invalid(field.NewPath("spec").Child("statefulSet.").Child("volumeClaimTemplates"),
```

pkoutsovasilis: oh my 😄 I adore the fact that I paid attention to the golang struct spec.Statefulset but not in the field path.. thanks @barkbay

barkbay (commented):
I did a test using the sample and Agent is stuck in a `red` state:

```
k get es,agent,sts,pods                  
NAME                                                       HEALTH   NODES   VERSION   PHASE   AGE
elasticsearch.elasticsearch.k8s.elastic.co/elasticsearch   green    3       8.10.4    Ready   118s

NAME                                       HEALTH   AVAILABLE   EXPECTED   VERSION   AGE
agent.agent.k8s.elastic.co/elastic-agent   red                  1          8.10.4    119s

NAME                                        READY   AGE
statefulset.apps/elastic-agent-agent        1/1     116s
statefulset.apps/elasticsearch-es-default   3/3     117s

NAME                             READY   STATUS    RESTARTS   AGE
pod/elastic-agent-agent-0        2/2     Running   0          116s
pod/elasticsearch-es-default-0   1/1     Running   0          116s
pod/elasticsearch-es-default-1   1/1     Running   0          116s
pod/elasticsearch-es-default-2   1/1     Running   0          116s
pod/kibana-kb-7d57f7cf77-qvkxj   1/1     Running   0          116s
```

I have the following ~error~ warning in Agent logs (`k logs pod/elastic-agent-agent-0 -c agent`):

```json
{
    "log.level": "warn",
    "@timestamp": "2023-12-19T07:30:40.848Z",
    "message": "Cannot index event publisher.Event ... (status=400): {\"type\":\"document_parsing_exception\",\"reason\":\"[1:2047] failed to parse: data stream timestamp field [@timestamp] is missing\",\"caused_by\":{\"type\":\"illegal_argument_exception\",\"reason\":\"data stream timestamp field [@timestamp] is missing\"}}, dropping event!",
    "component": {
        "binary": "filebeat",
        "dataset": "elastic_agent.filebeat",
        "id": "filestream-monitoring",
        "type": "filestream"
    },
    "log": {
        "source": "filestream-monitoring"
    },
    "log.origin": {
        "file.line": 446,
        "file.name": "elasticsearch/client.go"
    },
    "service.name": "filebeat",
    "ecs.version": "1.6.0",
    "log.logger": "elasticsearch"
}
```

Also `ecs.version` seems to be duplicated in the original document.

barkbay:
> I did a test using the sample and Agent is stuck in a red state:

Actually `Agent` became `green` once I restarted the ECK Pod: 

```
NAME                                       HEALTH   AVAILABLE   EXPECTED   VERSION   AGE
agent.agent.k8s.elastic.co/elastic-agent   green    1           1          8.10.4    15m
```

I think we may have a bug in `Agent` where reconciliations don't always happen when they should (in case of state change for example) 😕 .

barkbay (approved):
Re LGTM

PR #7377, opened 2023-12-11
fix secure settings link on stack config policy page

kvalliyurnatt:
The link to secure settings page was incorrect and was instead pointing to the custom images page.


naemono:

Issue #7170, opened 2023-09-18
Kibana service account token not recreated when following credential rotation documentation 


pebrc:
https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-rotate-credentials.html

instructs users to delete the relevant secret that has the  `eck.k8s.elastic.co/credentials=true` label. For Kibana that would be the `$KIBANA_NAME-kibana-user` secret. However if doing so the secret is not recreated with a new service account token. Only if we also delete the $NAMESPACE-$KIBANA_NAME-kibana-user` secret that contains the ES side of the token because this is the only one we are watching https://github.com/elastic/cloud-on-k8s/blob/25883251a99a02d8870729b376a4abf6c611cb21/pkg/controller/association/dynamic_watches.go#L89 IIUC

We should react to changes to the application side secret so that our credential rotation instructions work as expected. 

Workaround exists: restart ECK operator to re-reconcile the association an re-create the secret.

PR #7382, opened 2023-12-13
Allow Agent and Elastic stack in different namespaces.

### LINKED ISSUE #7352: Running Agent as non root in a dedicated namespace is not possible
Despite [what the documentation suggests](https://www.elastic.co/guide/en/cloud-on-k8s/2.10/k8s-elastic-agent-fleet-known-limitations.html#k8s_running_as_root_and_within_a_single_namespace_eck_2_10_0_and_agent_7_14_0):

> Until version 7.14.0 and ECK version 2.10.0, Elastic Agent in Fleet mode has to run as root and in the same namespace as the Elasticsearch cluster it connects to.

It is not possible to run Agent as non root and in a namespace other than the one where Elasticsearch is deployed. [This check](https://github.com/elastic/cloud-on-k8s/blob/ba648ba357fdb52e9b7e7bbe6c3f9ca045214a18/pkg/controller/agent/pod.go#L337-L343) is preventing this kind of setup to be deployed:

```golang
func applyRelatedEsAssoc(agent agentv1alpha1.Agent, esAssociation commonv1.Association, builder *defaults.PodTemplateBuilder) (*defaults.PodTemplateBuilder, error) {
	if esAssociation == nil {
		return builder, nil
	}

	esRef := esAssociation.AssociationRef()
	if !esRef.IsExternal() && !agent.Spec.FleetServerEnabled && agent.Namespace != esRef.Namespace {
		// check agent and ES share the same namespace
		return nil, fmt.Errorf(
			"agent namespace %s is different than referenced Elasticsearch namespace %s, this is not supported yet",
			agent.Namespace,
			esAssociation.AssociationRef().Namespace,
		)
	}
```

We should understand if it safe to remove this check and fix the documentation.

It might be an oversight of https://github.com/elastic/cloud-on-k8s/pull/6700

naemono:
resolves #7352 

## What does this change?

Both root and non-root, fleet and non-fleet Agent installations were being blocked from Agent running in a different namespace because of a legacy association check. This legacy check checked that the agent was running in the same namespace as Elasticsearch. The actual issue turned out to be when Agent is not running in the same namespace as Fleet server when fleet mode is enabled. This change enables Agent to run in any namespace regardless of where Fleet or Elasticsearch is deployed.

This supersedes #7353 as this turned out to be a much simpler changeset vs:
image

Changes:
- [x] The association controller for Agent -> Fleet can now retrieve a slice of secrets that should be copied to a target namespace to allow secure communications between Agent => Elasticsearch.

Testing Done/Verification
- [x] Secret(s) are deleted upon associated object being deleted.
- [x] Secret(s) are watched for changes.
- [x] Watches are removed on delete.
- [x] E2e tests updated to verify both same namespace, and cross-namespace associations work.
- [x] Verify when source secret changes, the copied secrets changes as well

### DIFF
--- a/docs/orchestrating-elastic-stack-applications/agent-fleet.asciidoc
+++ b/docs/orchestrating-elastic-stack-applications/agent-fleet.asciidoc
@@ -707,10 +707,10 @@ Until version 7.14.0 and ECK version 2.10.0, Elastic Agent and Fleet Server were
 
 As of Elastic Stack version 7.14.0 and ECK version 2.10.0 it is also possible to run Elastic Agent and Fleet as a non-root user. See <<{p}_storing_local_state_in_host_path_volume>> for instructions.
 
-=== Elastic Agent running in the same namespace as Elastic Fleet Server
-Until ECK version 2.11.0, Elastic Agent and Fleet Server were required to run within the same Namespace.
+=== Elastic Agent running in the same namespace as the Elastic stack.
+Until ECK version 2.11.0, Elastic Agent and Fleet Server were required to run within the same Namespace as Elasticsearch.
 
-As of ECK version 2.11.0, Elastic Agent and Fleet Server can be deployed in different Namespaces.
+As of ECK version 2.11.0, Elastic Agent, Fleet Server and Elasticsearch can all be deployed in different Namespaces.
 
 === Running Endpoint Security integration
 Running Endpoint Security link:https://www.elastic.co/guide/en/security/current/install-endpoint.html[integration] is not yet supported in containerized environments, like Kubernetes. This is not an ECK limitation, but the limitation of the integration itself. Note that you can use ECK to deploy Elasticsearch, Kibana and Fleet Server, and add Endpoint Security integration to your policies if Elastic Agents running those policies are deployed in non-containerized environments.
--- a/docs/orchestrating-elastic-stack-applications/agent-fleet.asciidoc
+++ b/docs/orchestrating-elastic-stack-applications/agent-fleet.asciidoc
@@ -702,12 +702,15 @@ Deploys single instance Elastic Agent Deployment in Fleet mode with APM integrat
 [id="{p}-elastic-agent-fleet-known-limitations"]
 == Known limitations
 
-=== Running as root and within a single namespace (ECK < 2.10.0 and Agent < 7.14.0)
-Until version 7.14.0 and ECK version 2.10.0, Elastic Agent in Fleet mode has to run as root and in the same namespace as the Elasticsearch cluster it connects to. 
+=== Running as root (ECK < 2.10.0 and Agent < 7.14.0)
+Until version 7.14.0 and ECK version 2.10.0, Elastic Agent and Fleet Server were required to run as root.
 
-This was due to configuration limitations in Fleet/Elastic Agent. ECK needed to establish trust between Elastic Agents and Elasticsearch. ECK was only able to fetch the required Elasticsearch CA correctly if both resources are in the same namespace.
 As of Elastic Stack version 7.14.0 and ECK version 2.10.0 it is also possible to run Elastic Agent and Fleet as a non-root user. See <<{p}_storing_local_state_in_host_path_volume>> for instructions.
-To establish trust, the Pod needs to update the CA store through a call to `update-ca-trust` before Elastic Agent runs. To call it successfully, the Pod needs to run with elevated privileges.
+
+=== Elastic Agent running in the same namespace as Elastic Fleet Server
+Until ECK version 2.11.0, Elastic Agent and Fleet Server were required to run within the same Namespace.
+
+As of ECK version 2.11.0, Elastic Agent and Fleet Server can be deployed in different Namespaces.
 
 === Running Endpoint Security integration
 Running Endpoint Security link:https://www.elastic.co/guide/en/security/current/install-endpoint.html[integration] is not yet supported in containerized environments, like Kubernetes. This is not an ECK limitation, but the limitation of the integration itself. Note that you can use ECK to deploy Elasticsearch, Kibana and Fleet Server, and add Endpoint Security integration to your policies if Elastic Agents running those policies are deployed in non-containerized environments.
--- a/pkg/apis/common/v1/association.go
+++ b/pkg/apis/common/v1/association.go
@@ -204,7 +204,7 @@ type AssociationConf struct {
 	IsServiceAccount      bool   `json:"isServiceAccount"`
 	CACertProvided        bool   `json:"caCertProvided"`
 	CASecretName          string `json:"caSecretName"`
-	AdditionalSecretsHash string `json:"additionalSecretsHash"`
+	AdditionalSecretsHash string `json:"additionalSecretsHash,omitempty"`
 	URL                   string `json:"url"`
 	// Version of the referenced resource. If a version upgrade is in progress,
 	// matches the lowest running version. May be empty if unknown.
--- a/pkg/apis/common/v1/association.go
+++ b/pkg/apis/common/v1/association.go
@@ -199,12 +199,13 @@ func FormatNameWithID(template string, id string) string {
 
 // AssociationConf holds the association configuration of a referenced resource in an association.
 type AssociationConf struct {
-	AuthSecretName   string `json:"authSecretName"`
-	AuthSecretKey    string `json:"authSecretKey"`
-	IsServiceAccount bool   `json:"isServiceAccount"`
-	CACertProvided   bool   `json:"caCertProvided"`
-	CASecretName     string `json:"caSecretName"`
-	URL              string `json:"url"`
+	AuthSecretName        string `json:"authSecretName"`
+	AuthSecretKey         string `json:"authSecretKey"`
+	IsServiceAccount      bool   `json:"isServiceAccount"`
+	CACertProvided        bool   `json:"caCertProvided"`
+	CASecretName          string `json:"caSecretName"`
+	AdditionalSecretsHash string `json:"additionalSecretsHash"`
+	URL                   string `json:"url"`
 	// Version of the referenced resource. If a version upgrade is in progress,
 	// matches the lowest running version. May be empty if unknown.
 	Version string `json:"version"`
--- a/pkg/apis/common/v1/association.go
+++ b/pkg/apis/common/v1/association.go
@@ -199,11 +199,13 @@ func FormatNameWithID(template string, id string) string {
 
 // AssociationConf holds the association configuration of a referenced resource in an association.
 type AssociationConf struct {
-	AuthSecretName        string `json:"authSecretName"`
-	AuthSecretKey         string `json:"authSecretKey"`
-	IsServiceAccount      bool   `json:"isServiceAccount"`
-	CACertProvided        bool   `json:"caCertProvided"`
-	CASecretName          string `json:"caSecretName"`
+	AuthSecretName   string `json:"authSecretName"`
+	AuthSecretKey    string `json:"authSecretKey"`
+	IsServiceAccount bool   `json:"isServiceAccount"`
+	CACertProvided   bool   `json:"caCertProvided"`
+	CASecretName     string `json:"caSecretName"`
+	// AdditionalSecretsHash is a hash of additional secrets such that when any of the underlying
+	// secrets change, the CRD annotation is updated and the pods are restarted.
 	AdditionalSecretsHash string `json:"additionalSecretsHash,omitempty"`
 	URL                   string `json:"url"`
 	// Version of the referenced resource. If a version upgrade is in progress,
--- a/pkg/controller/association/dynamic_watches.go
+++ b/pkg/controller/association/dynamic_watches.go
@@ -35,6 +35,12 @@ func serviceWatchName(associated types.NamespacedName) string {
 	return fmt.Sprintf("%s-%s-svc-watch", associated.Namespace, associated.Name)
 }
 
+// serviceWatchName returns the name of the watch setup on the custom service to be used to make requests to the
+// referenced resource.
+func additionalSecretWatchName(associated types.NamespacedName) string {
+	return fmt.Sprintf("%s-%s-secrets-watch", associated.Namespace, associated.Name)
+}
+
 // reconcileWatches sets up dynamic watches for:
 // * the referenced resource(s) managed or not by ECK (e.g. Elasticsearch for Kibana -> Elasticsearch associations)
 // * the CA secret of the referenced resource in the referenced resource namespace
@@ -93,7 +99,47 @@ func (r *Reconciler) reconcileWatches(associated types.NamespacedName, associati
 		}
 	}
 
+	if r.AdditionalSecrets != nil {
+		if err := ReconcileGenericWatch(associated, managedElasticRef, r.watches.Secrets, additionalSecretWatchName(associated), func() ([]types.NamespacedName, error) {
+			var toWatch []types.NamespacedName
+			for _, association := range associations {
+				secs, err := r.AdditionalSecrets(r.Client, association)
+				if err != nil {
+					return nil, err
+				}
+				toWatch = append(toWatch, secs...)
+			}
+			return toWatch, nil
+		}); err != nil {
+			return err
+		}
+	}
+
 	return nil
+}
+
+func ReconcileGenericWatch(
+	associated types.NamespacedName,
+	associations []commonv1.Association,
+	dynamicRequest *watches.DynamicEnqueueRequest,
+	watchName string,
+	watchedFunc func() ([]types.NamespacedName, error),
+) error {
+	if len(associations) == 0 {
+		// clean up if there are none
+		RemoveWatch(dynamicRequest, watchName)
+		return nil
+	}
+
+	watched, err := watchedFunc()
+	if err != nil {
+		return err
+	}
+	return dynamicRequest.AddHandler(watches.NamedWatch{
+		Name:    watchName,
+		Watched: watched,
+		Watcher: associated,
+	})
 }
 
 // ReconcileWatch sets or removes `watchName` watch in `dynamicRequest` based on `associated` and `associations` and
@@ -105,26 +151,17 @@ func ReconcileWatch(
 	watchName string,
 	watchedFunc func(association commonv1.Association) types.NamespacedName,
 ) error {
-	if len(associations) == 0 {
-		// clean up if there are none
-		RemoveWatch(dynamicRequest, watchName)
-		return nil
-	}
+	return ReconcileGenericWatch(associated, associations, dynamicRequest, watchName, func() ([]types.NamespacedName, error) {
+		emptyNamespacedName := types.NamespacedName{}
 
-	emptyNamespacedName := types.NamespacedName{}
-
-	toWatch := make([]types.NamespacedName, 0, len(associations))
-	for _, association := range associations {
-		watchedNamespacedName := watchedFunc(association)
-		if watchedNamespacedName != emptyNamespacedName {
-			toWatch = append(toWatch, watchedFunc(association))
+		toWatch := make([]types.NamespacedName, 0, len(associations))
+		for _, association := range associations {
+			watchedNamespacedName := watchedFunc(association)
+			if watchedNamespacedName != emptyNamespacedName {
+				toWatch = append(toWatch, watchedFunc(association))
+			}
 		}
-	}
-
-	return dynamicRequest.AddHandler(watches.NamedWatch{
-		Name:    watchName,
-		Watched: toWatch,
-		Watcher: associated,
+		return toWatch, nil
 	})
 }
 
@@ -142,4 +179,6 @@ func (r *Reconciler) removeWatches(associated types.NamespacedName) {
 	RemoveWatch(r.watches.Services, serviceWatchName(associated))
 	// - ES user secret
 	RemoveWatch(r.watches.Secrets, esUserWatchName(associated))
+	// - Additional secrets (typically in the case of Agent -> Fleet Server -> Elasticsearch)
+	RemoveWatch(r.watches.Secrets, additionalSecretWatchName(associated))
 }
--- a/pkg/controller/association/dynamic_watches.go
+++ b/pkg/controller/association/dynamic_watches.go
@@ -35,8 +35,8 @@ func serviceWatchName(associated types.NamespacedName) string {
 	return fmt.Sprintf("%s-%s-svc-watch", associated.Namespace, associated.Name)
 }
 
-// serviceWatchName returns the name of the watch setup on the custom service to be used to make requests to the
-// referenced resource.
+// additionalSecretWatchName returns the name of the watch setup on any additional secrets that
+// are copied during the association reconciliation.
 func additionalSecretWatchName(associated types.NamespacedName) string {
 	return fmt.Sprintf("%s-%s-secrets-watch", associated.Namespace, associated.Name)
 }
--- a/pkg/controller/association/dynamic_watches.go
+++ b/pkg/controller/association/dynamic_watches.go
@@ -5,6 +5,7 @@
 package association
 
 import (
+	"context"
 	"fmt"
 
 	"k8s.io/apimachinery/pkg/types"
@@ -49,7 +50,7 @@ func additionalSecretWatchName(associated types.NamespacedName) string {
 // * if there's an ES user to create, watch the user Secret in ES namespace
 // All watches for all given associations are set under the same watch name and replaced with each reconciliation.
 // The given associations are expected to be of the same type (e.g. Kibana -> Elasticsearch, not Kibana -> Enterprise Search).
-func (r *Reconciler) reconcileWatches(associated types.NamespacedName, associations []commonv1.Association) error {
+func (r *Reconciler) reconcileWatches(ctx context.Context, associated types.NamespacedName, associations []commonv1.Association) error {
 	managedElasticRef := filterManagedElasticRef(associations)
 	unmanagedElasticRef := filterUnmanagedElasticRef(associations)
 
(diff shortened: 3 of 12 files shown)


pebrc (commented):
The hash to rotate the agents on changes to the ES CA is missing, which means that the Agents won't pick up the new CA until the ECK operator restarts. We can remedy this by adding the hash to the association configuration annotation like this 

```diff
diff --git a/pkg/apis/common/v1/association.go b/pkg/apis/common/v1/association.go
index 8bcfaa2f2..5ecefd67b 100644
--- a/pkg/apis/common/v1/association.go
+++ b/pkg/apis/common/v1/association.go
@@ -199,12 +199,13 @@ func FormatNameWithID(template string, id string) string {
 
 // AssociationConf holds the association configuration of a referenced resource in an association.
 type AssociationConf struct {
-	AuthSecretName   string `json:"authSecretName"`
-	AuthSecretKey    string `json:"authSecretKey"`
-	IsServiceAccount bool   `json:"isServiceAccount"`
-	CACertProvided   bool   `json:"caCertProvided"`
-	CASecretName     string `json:"caSecretName"`
-	URL              string `json:"url"`
+	AuthSecretName        string `json:"authSecretName"`
+	AuthSecretKey         string `json:"authSecretKey"`
+	IsServiceAccount      bool   `json:"isServiceAccount"`
+	CACertProvided        bool   `json:"caCertProvided"`
+	CASecretName          string `json:"caSecretName"`
+	AdditionalSecretsHash string `json:"additionalSecretsHash"`
+	URL                   string `json:"url"`
 	// Version of the referenced resource. If a version upgrade is in progress,
 	// matches the lowest running version. May be empty if unknown.
 	Version string `json:"version"`
diff --git a/pkg/controller/association/controller/agent_fleetserver.go b/pkg/controller/association/controller/agent_fleetserver.go
index 91474ecad..ba51acf49 100644
--- a/pkg/controller/association/controller/agent_fleetserver.go
+++ b/pkg/controller/association/controller/agent_fleetserver.go
@@ -17,6 +17,7 @@ import (
 	"github.com/elastic/cloud-on-k8s/v2/pkg/controller/association"
 	"github.com/elastic/cloud-on-k8s/v2/pkg/controller/common/operator"
 	"github.com/elastic/cloud-on-k8s/v2/pkg/utils/k8s"
+	ulog "github.com/elastic/cloud-on-k8s/v2/pkg/utils/log"
 	"github.com/elastic/cloud-on-k8s/v2/pkg/utils/rbac"
 )
 
@@ -46,11 +47,12 @@ func AddAgentFleetServer(mgr manager.Manager, accessReviewer rbac.AccessReviewer
 	})
 }
 
-func addtionalSecrets(c k8s.Client, assoc commonv1.Association) ([]types.NamespacedName, error) {
+func addtionalSecrets(ctx context.Context, c k8s.Client, assoc commonv1.Association) ([]types.NamespacedName, error) {
+	log := ulog.FromContext(ctx)
 	associated := assoc.Associated()
 	var agent agentv1alpha1.Agent
 	nsn := types.NamespacedName{Namespace: associated.GetNamespace(), Name: associated.GetName()}
-	if err := c.Get(context.Background(), nsn, &agent); err != nil {
+	if err := c.Get(ctx, nsn, &agent); err != nil {
 		return nil, err
 	}
 	fleetServerRef := assoc.AssociationRef()
@@ -58,7 +60,7 @@ func addtionalSecrets(c k8s.Client, assoc commonv1.Association) ([]types.Namespa
 		return nil, nil
 	}
 	fleetServer := agentv1alpha1.Agent{}
-	if err := c.Get(context.Background(), fleetServerRef.NamespacedName(), &fleetServer); err != nil {
+	if err := c.Get(ctx, fleetServerRef.NamespacedName(), &fleetServer); err != nil {
 		return nil, err
 	}
 
@@ -74,14 +76,17 @@ func addtionalSecrets(c k8s.Client, assoc commonv1.Association) ([]types.Namespa
 
 	// if both agent and ES are same namespace no copying needed
 	if agent.GetNamespace() == esAssociation.GetNamespace() {
+		log.V(1).Info("no additional secrets because same namespace")
 		return nil, nil
 	}
 
 	conf, err := esAssociation.AssociationConf()
 	if err != nil {
+		log.V(1).Info("no additional secrets because no assoc conf")
 		return nil, err
 	}
 	if conf == nil || !conf.CACertProvided {
+		log.V(1).Info("no additional secrets because conf nil or no CA provided")
 		return nil, nil
 	}
 	return []types.NamespacedName{{
diff --git a/pkg/controller/association/dynamic_watches.go b/pkg/controller/association/dynamic_watches.go
index 2245543dd..a7dabb3eb 100644
--- a/pkg/controller/association/dynamic_watches.go
+++ b/pkg/controller/association/dynamic_watches.go
@@ -5,6 +5,7 @@
 package association
 
 import (
+	"context"
 	"fmt"
 
 	"k8s.io/apimachinery/pkg/types"
@@ -49,7 +50,7 @@ func additionalSecretWatchName(associated types.NamespacedName) string {
 // * if there's an ES user to create, watch the user Secret in ES namespace
 // All watches for all given associations are set under the same watch name and replaced with each reconciliation.
 // The given associations are expected to be of the same type (e.g. Kibana -> Elasticsearch, not Kibana -> Enterprise Search).
-func (r *Reconciler) reconcileWatches(associated types.NamespacedName, associations []commonv1.Association) error {
+func (r *Reconciler) reconcileWatches(ctx context.Context, associated types.NamespacedName, associations []commonv1.Association) error {
 	managedElasticRef := filterManagedElasticRef(associations)
 	unmanagedElasticRef := filterUnmanagedElasticRef(associations)
 
@@ -103,7 +104,7 @@ func (r *Reconciler) reconcileWatches(associated types.NamespacedName, associati
 		if err := ReconcileGenericWatch(associated, managedElasticRef, r.watches.Secrets, additionalSecretWatchName(associated), func() ([]types.NamespacedName, error) {
 			var toWatch []types.NamespacedName
 			for _, association := range associations {
-				secs, err := r.AdditionalSecrets(r.Client, association)
+				secs, err := r.AdditionalSecrets(ctx, r.Client, association)
 				if err != nil {
 					return nil, err
 				}
diff --git a/pkg/controller/association/reconciler.go b/pkg/controller/association/reconciler.go
index c72298b49..a552873d2 100644
--- a/pkg/controller/association/reconciler.go
+++ b/pkg/controller/association/reconciler.go
@@ -7,6 +7,7 @@ package association
 import (
 	"context"
 	"fmt"
+	"hash/fnv"
 	"reflect"
 	"time"
 
@@ -64,7 +65,7 @@ type AssociationInfo struct { //nolint:revive
 	// AdditionalSecrets are additional secrets to copy from an association's namespace to the associated resource namespace.
 	// Currently this is only used for copying the CA from an Elasticsearch association to the same namespace as
 	// an Agent referencing a Fleet Server.
-	AdditionalSecrets func(c k8s.Client, assoc commonv1.Association) ([]types.NamespacedName, error)
+	AdditionalSecrets func(context.Context, k8s.Client, commonv1.Association) ([]types.NamespacedName, error)
 	// Labels are labels set on all resources created for association purpose. Note that some resources will be also
 	// labelled with AssociationResourceNameLabelName and AssociationResourceNamespaceLabelName in addition to any
 	// labels provided here.
@@ -191,7 +192,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
 	}
 
 	// reconcile watches for all associations of this type
-	if err := r.reconcileWatches(associatedKey, associations); err != nil {
+	if err := r.reconcileWatches(ctx, associatedKey, associations); err != nil {
 		return reconcile.Result{}, tracing.CaptureError(ctx, err)
 	}
 
@@ -263,13 +264,14 @@ func (r *Reconciler) reconcileAssociation(ctx context.Context, association commo
 		return commonv1.AssociationPending, err // maybe not created yet
 	}
 
+	secretsHash := fnv.New32a()
 	if r.AdditionalSecrets != nil {
-		additionalSecrets, err := r.AdditionalSecrets(r.Client, association)
+		additionalSecrets, err := r.AdditionalSecrets(ctx, r.Client, association)
 		if err != nil {
 			return commonv1.AssociationPending, err // maybe not created yet
 		}
 		for _, sec := range additionalSecrets {
-			if err := copySecret(ctx, r.Client, r.AssociationInfo, k8s.ExtractNamespacedName(association.Associated()), association.GetNamespace(), sec); err != nil {
+			if err := copySecret(ctx, r.Client, secretsHash, k8s.ExtractNamespacedName(association.Associated()), association.GetNamespace(), sec); err != nil {
 				return commonv1.AssociationPending, err
 			}
 		}
@@ -294,10 +296,11 @@ func (r *Reconciler) reconcileAssociation(ctx context.Context, association commo
 
 	// construct the expected association configuration
 	expectedAssocConf := &commonv1.AssociationConf{
-		CACertProvided: caSecret.CACertProvided,
-		CASecretName:   caSecret.Name,
-		URL:            url,
-		Version:        ver,
+		CACertProvided:        caSecret.CACertProvided,
+		CASecretName:          caSecret.Name,
+		AdditionalSecretsHash: fmt.Sprint(secretsHash.Sum32()),
+		URL:                   url,
+		Version:               ver,
 	}
 
 	if r.ElasticsearchUserCreation == nil {
diff --git a/pkg/controller/association/reconciler_test.go b/pkg/controller/association/reconciler_test.go
index 678f47977..adcaab23a 100644
--- a/pkg/controller/association/reconciler_test.go
+++ b/pkg/controller/association/reconciler_test.go
@@ -341,7 +341,7 @@ func TestReconciler_Reconcile_NoESRef_Cleanup(t *testing.T) {
 	require.NotEmpty(t, kb.Annotations[kb.EsAssociation().AssociationConfAnnotationName()])
 	r := testReconciler(&kb, &kibanaUserInESNamespace, &kibanaUserInKibanaNamespace, &esCertsInKibanaNamespace)
 	// simulate watches being set
-	require.NoError(t, r.reconcileWatches(k8s.ExtractNamespacedName(&kb), []commonv1.Association{kb.EsAssociation()}))
+	require.NoError(t, r.reconcileWatches(context.Background(), k8s.ExtractNamespacedName(&kb), []commonv1.Association{kb.EsAssociation()}))
 	require.NotEmpty(t, r.watches.Secrets.Registrations())
 	require.NotEmpty(t, r.watches.ReferencedResources.Registrations())
 
diff --git a/pkg/controller/association/secret.go b/pkg/controller/association/secret.go
index 455a6c7e9..92fea27ce 100644
--- a/pkg/controller/association/secret.go
+++ b/pkg/controller/association/secret.go
@@ -11,6 +11,7 @@ import (
 	"crypto/x509"
 	"encoding/json"
 	"fmt"
+	"hash"
 	"net/http"
 
 	corev1 "k8s.io/api/core/v1"
@@ -19,11 +20,11 @@ import (
 
 	commonv1 "github.com/elastic/cloud-on-k8s/v2/pkg/apis/common/v1"
 	"github.com/elastic/cloud-on-k8s/v2/pkg/controller/common/certificates"
+	commonhash "github.com/elastic/cloud-on-k8s/v2/pkg/controller/common/hash"
 	"github.com/elastic/cloud-on-k8s/v2/pkg/controller/common/reconciler"
 	"github.com/elastic/cloud-on-k8s/v2/pkg/controller/common/version"
 	"github.com/elastic/cloud-on-k8s/v2/pkg/controller/elasticsearch/client"
 	"github.com/elastic/cloud-on-k8s/v2/pkg/utils/k8s"
-	"github.com/elastic/cloud-on-k8s/v2/pkg/utils/maps"
 )
 
 const (
@@ -194,16 +195,15 @@ func filterManagedElasticRef(associations []commonv1.Association) []commonv1.Ass
 }
 
 // copySecret will copy the source secret to the target namespace adding labels from the associated object to ensure garbage collection happens.
-func copySecret(ctx context.Context, client k8s.Client, info AssociationInfo, associated types.NamespacedName, targetNamespace string, source types.NamespacedName) error {
+func copySecret(ctx context.Context, client k8s.Client, secHash hash.Hash, associated types.NamespacedName, targetNamespace string, source types.NamespacedName) error {
 	var original corev1.Secret
 	if err := client.Get(ctx, source, &original); err != nil {
 		return err
 	}
 	var expected corev1.Secret
 	expected.Data = original.Data
-	// We merge the original labels with the associated object's association labels to ensure
-	// that this secret is garbage collected when the associated object is deleted.
-	expected.Labels = maps.Merge(original.Labels, info.Labels(associated))
+	commonhash.WriteHashObject(secHash, original.Data)
+	expected.Labels = original.Labels
 	expected.Annotations = original.Annotations
 	expected.Name = original.Name
 	expected.Namespace = targetNamespace
```

pkg/controller/association/secret.go
```diff
@@ -189,3 +193,21 @@ func filterManagedElasticRef(associations []commonv1.Association) []commonv1.Ass
...
+	if err := client.Get(ctx, source, &original); err != nil {
+		return err
+	}
+	var expected corev1.Secret
+	expected.Data = original.Data
+	commonhash.WriteHashObject(secHash, original.Data)
+	expected.Labels = original.Labels
+	expected.Annotations = original.Annotations
+	expected.Name = original.Name
+	expected.Namespace = targetNamespace
```
barkbay, line 209:
```suggestion
	expected := corev1.Secret{
		ObjectMeta: metav1.ObjectMeta{
			Name:        original.Name,
			Namespace:   targetNamespace,
			Labels:      original.Labels,
			Annotations: original.Annotations,
		},
		Data: original.Data,
		Type: original.Type,
	}
	commonhash.WriteHashObject(secHash, original.Data)
```

pkg/controller/association/secret.go
```diff
@@ -189,3 +193,21 @@ func filterManagedElasticRef(associations []commonv1.Association) []commonv1.Ass
...
+
+	_, err := reconciler.ReconcileSecret(ctx, client, expected, nil)
```
barkbay, line 211: I was wondering if we could use a soft owner to garbage collect this Secret.

```suggestion
	_, err := reconciler.ReconcileSecretNoOwnerRef(ctx, client, expected, softOwner)
```

`softOwner` being `referencedObj` IIUC, and maybe use `GarbageCollectSoftOwnedSecrets` to remove them (not tested, only an idea).

pkg/apis/common/v1/association.go
```diff
@@ -199,12 +199,13 @@ func FormatNameWithID(template string, id string) string {
...
-	IsServiceAccount bool   `json:"isServiceAccount"`
-	CACertProvided   bool   `json:"caCertProvided"`
-	CASecretName     string `json:"caSecretName"`
-	URL              string `json:"url"`
+	AuthSecretName        string `json:"authSecretName"`
+	AuthSecretKey         string `json:"authSecretKey"`
+	IsServiceAccount      bool   `json:"isServiceAccount"`
+	CACertProvided        bool   `json:"caCertProvided"`
+	CASecretName          string `json:"caSecretName"`
+	AdditionalSecretsHash string `json:"additionalSecretsHash"`
```
barkbay, line 207:
```suggestion
	AdditionalSecretsHash string `json:"additionalSecretsHash,omitempty"`
```

pkg/controller/association/dynamic_watches.go
```diff
@@ -93,41 +100,72 @@ func (r *Reconciler) reconcileWatches(associated types.NamespacedName, associati
...
+		}
+	}
+
 	return nil
 }
 
-// ReconcileWatch sets or removes `watchName` watch in `dynamicRequest` based on `associated` and `associations` and
-// `watchedFunc`. No watch is added if watchedFunc(association) refers to an empty namespaced name.
-func ReconcileWatch(
+func ReconcileGenericWatch(
```
barkbay, line 122:
```suggestion
func reconcileGenericWatch(
```

pkg/controller/association/reconciler.go
```diff
@@ -258,6 +264,19 @@ func (r *Reconciler) reconcileAssociation(ctx context.Context, association commo
...
 	}
 
+	secretsHash := fnv.New32a()
+	if r.AdditionalSecrets != nil {
+		additionalSecrets, err := r.AdditionalSecrets(ctx, r.Client, association)
+		if err != nil {
+			return commonv1.AssociationPending, err // maybe not created yet
+		}
+		for _, sec := range additionalSecrets {
+			if err := copySecret(ctx, r.Client, secretsHash, association.GetNamespace(), sec); err != nil {
```
barkbay, line 274: Are we supposed to watch this Secret? It's not reconciled if I delete it.

pkg/controller/association/reconciler.go
```diff
@@ -294,10 +296,11 @@ func (r *Reconciler) reconcileAssociation(ctx context.Context, association commo
 
 	// construct the expected association configuration
 	expectedAssocConf := &commonv1.AssociationConf{
-		CACertProvided: caSecret.CACertProvided,
-		CASecretName:   caSecret.Name,
-		URL:            url,
-		Version:        ver,
+		CACertProvided:        caSecret.CACertProvided,
+		CASecretName:          caSecret.Name,
+		AdditionalSecretsHash: fmt.Sprint(secretsHash.Sum32()),
```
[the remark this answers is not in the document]
pebrc: Actually I think we should always include the hash because we have a bug with the CA rotation with Agent even if Agent/FleetServer and ES are in the same namespace:
```diff
diff --git a/pkg/controller/association/controller/agent_fleetserver.go b/pkg/controller/association/controller/agent_fleetserver.go
index 7a3ebc4a9..89bda5264 100644
--- a/pkg/controller/association/controller/agent_fleetserver.go
+++ b/pkg/controller/association/controller/agent_fleetserver.go
@@ -74,12 +74,6 @@ func additionalSecrets(ctx context.Context, c k8s.Client, assoc commonv1.Associa
 		return nil, err
 	}
 
-	// if both agent and ES are same namespace no copying needed
-	if agent.GetNamespace() == esAssociation.GetNamespace() {
-		log.V(1).Info("no additional secrets because same namespace")
-		return nil, nil
-	}
-
 	conf, err := esAssociation.AssociationConf()
 	if err != nil {
 		log.V(1).Info("no additional secrets because no assoc conf")
diff --git a/pkg/controller/association/reconciler.go b/pkg/controller/association/reconciler.go
index d550d0c3a..a552873d2 100644
--- a/pkg/controller/association/reconciler.go
+++ b/pkg/controller/association/reconciler.go
@@ -271,7 +271,7 @@ func (r *Reconciler) reconcileAssociation(ctx context.Context, association commo
 			return commonv1.AssociationPending, err // maybe not created yet
 		}
 		for _, sec := range additionalSecrets {
-			if err := copySecret(ctx, r.Client, secretsHash, association.GetNamespace(), sec); err != nil {
+			if err := copySecret(ctx, r.Client, secretsHash, k8s.ExtractNamespacedName(association.Associated()), association.GetNamespace(), sec); err != nil {
 				return commonv1.AssociationPending, err
 			}
 		}
diff --git a/pkg/controller/association/secret.go b/pkg/controller/association/secret.go
index e8e746767..d0569c63b 100644
--- a/pkg/controller/association/secret.go
+++ b/pkg/controller/association/secret.go
@@ -195,14 +195,18 @@ func filterManagedElasticRef(associations []commonv1.Association) []commonv1.Ass
 }
 
 // copySecret will copy the source secret to the target namespace adding labels from the associated object to ensure garbage collection happens.
-func copySecret(ctx context.Context, client k8s.Client, secHash hash.Hash, targetNamespace string, source types.NamespacedName) error {
+func copySecret(ctx context.Context, client k8s.Client, secHash hash.Hash, associated types.NamespacedName, targetNamespace string, source types.NamespacedName) error {
 	var original corev1.Secret
 	if err := client.Get(ctx, source, &original); err != nil {
 		return err
 	}
+	commonhash.WriteHashObject(secHash, original.Data)
+	if targetNamespace == original.Namespace {
+		return nil
+	}
+
 	var expected corev1.Secret
 	expected.Data = original.Data
-	commonhash.WriteHashObject(secHash, original.Data)
 	expected.Labels = original.Labels
 	expected.Annotations = original.Annotations
 	expected.Name = original.Name
```

This is because we do not trigger a reconciliation in the Agent controller if the ES CA changes. Always including the hash annotation would make sure that the reconciliation happens (thanks to the annotation update) and the existing code takes then care of rotating the Agent pods because it looks at the secret contents of the ES CA secret.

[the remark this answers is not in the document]
pebrc: > Should we include the hash only if AdditionalSecrets is not empty?

I think you are right that we can omit the hash when there are not additional secrets. My comment was more aiming at the case where there are additional secrets but they are in the same namespace as the Agent.

pkg/controller/association/secret.go
pebrc: I think garbage collection is already tied to the association, in this case the Fleet Server->ES association. This has the odd side effect that the secret stays if the Agent->FleetServer association is severed but the Fleet Server->ES association still exists. I was thinking that this would be sufficient though because typically you would deploy Fleet server and Agent in tandem.

barkbay (commented):
Is this comment, https://github.com/elastic/cloud-on-k8s/pull/7353#pullrequestreview-1756750642, still valid? 

> I think we also need to adjust the documentation, ECK 2.10.0 does not allow Agent and ES to run in different namespaces.

pkg/controller/association/secret.go
barkbay: > This has the odd side effect that the secret stays if the Agent->FleetServer association is severed but the Fleet Server->ES association still exists.

Yes, this is what I was trying to "fix" with this suggestion. Not a strong opinion though, we can stick with the current behavior 👍

pkg/controller/association/reconciler.go
naemono: You're right, it doesn't return, but the `agent-es` association controller is triggered on delete. I wonder if this:
https://github.com/naemono/cloud-on-k8s/blob/5dd7bb85550fd0d0b23af1762803025e725bf2d3/pkg/controller/association/dynamic_watches.go#L103-L117
is associating the secret incorrectly? I'll investigate and update.

naemono: We were only watching the `source` secrets (the secrets we were copying into the target ns) and not the `target/created` secrets. It seems we would need to watch both:
1. if source secret changes, update target secret.
2. if target secret is changed/deleted, reconcile the secret.

I have updated and verified that both the target and source are re-created upon deletion.

naemono:
buildkite test this -f p=gke,s=8.11.0

pkg/controller/association/reconciler.go
pebrc: I noticed we have a similar issue with the fleet server es user/service token secret. Probably worth addressing in a separate PR though.

pebrc: I just realised we have captured my observation in an issue already https://github.com/elastic/cloud-on-k8s/issues/7170

pebrc:
https://github.com/elastic/cloud-on-k8s/pull/7382#discussion_r1426423982 could still be addressed but otherwise LGTM

naemono:
> [#7382 (comment)](https://github.com/elastic/cloud-on-k8s/pull/7382#discussion_r1426423982) could still be addressed but otherwise LGTM

Updated to only include the additional secrets Hash if there are additional secrets.

naemono:
Fixing unit tests...

naemono:
> Fixing unit tests...

Unit tests fixed.

docs/orchestrating-elastic-stack-applications/agent-fleet.asciidoc
```diff
@@ -702,12 +702,15 @@ Deploys single instance Elastic Agent Deployment in Fleet mode with APM integrat
...
-=== Running as root and within a single namespace (ECK < 2.10.0 and Agent < 7.14.0)
-Until version 7.14.0 and ECK version 2.10.0, Elastic Agent in Fleet mode has to run as root and in the same namespace as the Elasticsearch cluster it connects to. 
+=== Running as root (ECK < 2.10.0 and Agent < 7.14.0)
+Until version 7.14.0 and ECK version 2.10.0, Elastic Agent and Fleet Server were required to run as root.
 
-This was due to configuration limitations in Fleet/Elastic Agent. ECK needed to establish trust between Elastic Agents and Elasticsearch. ECK was only able to fetch the required Elasticsearch CA correctly if both resources are in the same namespace.
 As of Elastic Stack version 7.14.0 and ECK version 2.10.0 it is also possible to run Elastic Agent and Fleet as a non-root user. See <<{p}_storing_local_state_in_host_path_volume>> for instructions.
-To establish trust, the Pod needs to update the CA store through a call to `update-ca-trust` before Elastic Agent runs. To call it successfully, the Pod needs to run with elevated privileges.
+
+=== Elastic Agent running in the same namespace as Elastic Fleet Server
```
pebrc, line 710: The original requirement was that any Agent so both Fleet server and Agents needed to run in the **same namespace as Elasticsearch**. The new behaviour is that you can deploy everything wherever you want.

pkg/controller/association/reconciler_test.go
```diff
@@ -1048,11 +1050,320 @@ func TestReconciler_Reconcile_MultiRef(t *testing.T) {
...
+				"agentassociation.k8s.elastic.co/namespace": associated.Namespace,
+				"agentassociation.k8s.elastic.co/type":      commonv1.FleetServerAssociationType,
+			}
+		},
+		AssociationConfAnnotationNameBase:     commonv1.FleetServerConfigAnnotationNameBase,
+		AssociationResourceNameLabelName:      "agent.k8s.elastic.co/name",
+		AssociationResourceNamespaceLabelName: "agent.k8s.elastic.co/namespace",
+		ElasticsearchUserCreation:             nil,
+		AdditionalSecrets: func(ctx context.Context, c k8s.Client, assoc commonv1.Association) ([]types.NamespacedName, error) {
+			log.Printf("in additionalsecrets")
```
pebrc, line 1116: left over debugging?

pkg/controller/association/reconciler_test.go
```diff
@@ -1048,11 +1050,320 @@ func TestReconciler_Reconcile_MultiRef(t *testing.T) {
...
+			if err != nil {
+				log.Printf("additionalsecrets: single assoc error: %s", err)
+				return nil, err
+			}
+
+			// if both agent and ES are same namespace no copying needed
+			if agent.GetNamespace() == esAssociation.GetNamespace() {
+				log.Printf("additionalsecrets: same ns")
+				return nil, nil
+			}
```
pebrc, line 1151: we actually decided against this optimisation in production code

pebrc (approved):
LGTM just the docs needs an update I think to call out Elasticsearch specifically

docs/orchestrating-elastic-stack-applications/agent-fleet.asciidoc
naemono: > The original requirement was that any Agent so both Fleet server and Agents needed to run in the same namespace as Elasticsearch. The new behaviour is that you can deploy everything wherever you want.

The wording has been updated. Let me know your thoughts.

pebrc (approved):

Issue #6971, opened 2023-07-03
Enabled by default SSL for Logstash API


thbkrkr:
We should enabled by default SSL for Logstash API and make it possible to configure it through `spec.http` like it's possible for other stack apps (see [k8s-tls-certificates doc](https://www.elastic.co/guide/en/cloud-on-k8s/master/k8s-tls-certificates.html)).

```yaml
spec:
  http:
  
    service:
      spec:
        type: LoadBalancer
    
    tls:  
      selfSignedCertificate:
        disabled: true

      selfSignedCertificate:
        subjectAltNames:
          - ip: 160.46.176.15
          - dns: hulk.example.com
      
      certificate:
        secretName: my-cert
```

There is currently a TODO related to this:

https://github.com/elastic/cloud-on-k8s/blob/3a2ce8d4a2ca64c915b56e0b2e73ad0bc68e6eea/pkg/controller/logstash/pod.go#L104-L110

PR #7408, opened 2023-12-20
Logstash adds TLS support to API server

kaisecheng:
This PR adds TLS/ HTTPS and basic authentication integration to Logstash [API server](https://www.elastic.co/guide/en/logstash/current/monitoring-logstash.html#monitoring-api-security). The minimum support version changes from `8.6.0` to `8.12.0`.

Sample logstash.yml
```
api.ssl.enabled: "true"
api.ssl.keystore.path: "/path/to/keystore.p12"
api.ssl.keystore.password: "${SSL_KEYSTORE_PASSWORD}"
api.auth.type: basic
api.auth.basic.username: "${API_USERNAME}"
api.auth.basic.password: "${API_PASSWORD}"
```

HTTPS is on by default meaning `api.ssl.enabled`, `api.ssl.keystore.path` and `api.ssl.keystore.password` is set in config `logstash.yml`. The API server (puma jruby) only supports HTTPS with p12 keystore and java keystore. Therefore, [InitContainer](https://github.com/elastic/cloud-on-k8s/pull/7408/files#diff-000e81cb01c6f6b546ab205bc72599d2cc662ddcb8c5df9106eb7a2dd316c25aR38) needs to covert CA and TLS certs to the format puma accepts. If `api.ssl.enabled` set to true and the API Service is set to [disable](https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-tls-certificates.html#k8s-disable-tls) TLS `tls.selfSignedCertificate.disabled`, reconcile config [fails](https://github.com/elastic/cloud-on-k8s/pull/7408/files#diff-f2238a0d916b12187fca471853c77565a5d549079202cfe69199cd31b0139525R140). If API Service is set to disable and `api.ssl.enabled` is unset, server will disable TLS.

Logstash resolves `${VAR}` from [ENV](https://www.elastic.co/guide/en/logstash/current/environment-variables.html) and [Keystore](https://www.elastic.co/guide/en/logstash/current/keystore.html). When the same key is declared in both places, keystore takes the precedence. As Logstash allows setting HTTP basic authentication with `api.auth.type`, `api.auth.basic.username` and `api.auth.basic.password` in `logstash.yml`, this PR has integrated ReadinessProbe and Stack Monitoring by passing the resolved value of username password. The value of the variable comes from the following [sources](https://github.com/elastic/cloud-on-k8s/pull/7408/files#diff-f2238a0d916b12187fca471853c77565a5d549079202cfe69199cd31b0139525R202-R255) in the order of priority: Env, Env from ConfigMap, Env from Secret, Keystore from Secure Settings . The later sources take precedence.

Sample config
```yaml
kind: Elasticsearch
metadata:
  name: monitoring
spec:
  version: 8.12.0
  nodeSets:
    - name: default
      count: 1
      config:
        node.store.allow_mmap: false
---
apiVersion: v1
kind: Secret
metadata:
  name: logstash-secure-settings
stringData:
  API_USERNAME: batman
  API_PASSWORD: i_am_rich
---
kind: Logstash
metadata:
  name: logstash-sample
spec:
  count: 1
  version: 8.12.0
  config:
    api.auth.type: basic
    api.auth.basic.username: "${API_USERNAME}"
    api.auth.basic.password: "${API_PASSWORD"
  secureSettings:
    - secretName: logstash-secure-settings
  monitoring:
    metrics:
      elasticsearchRefs:
        - name: monitoring
    logs:
      elasticsearchRefs:
        - name: monitoring
  pipelines:
    - pipeline.id: main
      pipeline.workers: 2
      config.string: |
        output { 
          stdout {}
        }
---
```

The sample config creates following resources
```yaml
default    Logstash/logstash-sample                                      -              11m
default    ├─Secret/logstash-sample-default-monitoring-beat-ls-mon-user  -              11m
default    ├─Secret/logstash-sample-ls-config                            -              11m
default    ├─Secret/logstash-sample-ls-http-ca-internal                  -              11m
default    ├─Secret/logstash-sample-ls-http-certs-internal               -              11m
default    ├─Secret/logstash-sample-ls-monitoring-default-monitoring-ca  -              11m
default    ├─Secret/logstash-sample-ls-monitoring-filebeat-config        -              11m
default    ├─Secret/logstash-sample-ls-monitoring-metricbeat-config      -              11m
default    ├─Secret/logstash-sample-ls-pipeline                          -              11m
default    ├─Service/logstash-sample-ls-api                              -              11m
default    │ └─EndpointSlice/logstash-sample-ls-api-nh5w6                -              11m
default    └─StatefulSet/logstash-sample-ls                              -              11m
default      ├─ControllerRevision/logstash-sample-ls-5f77b6b9ff          -              11m
default      └─Pod/logstash-sample-ls-0                                  True           11m
```

In the past, Secret/logstash-sample-ls-config only stored the `logstash.yml` content. Now it stores the resolved value of api.ssl.keystore.password under the Secret key `API_KEYSTORE_PASS` for not exposing the password in plain text in initConfigContainer

e2e test
- TestLogstashStackMonitoring
- TestLogstashResolvingDollarVariableInStackMonitoring

fix: https://github.com/elastic/cloud-on-k8s/issues/6971, https://github.com/elastic/ingest-dev/issues/1591


pkg/controller/logstash/config_test.go
```diff
@@ -110,6 +139,29 @@ log:
...
+				logstash: v1alpha1.Logstash{
+					Spec: v1alpha1.LogstashSpec{
+						Config: &commonv1.Config{Data: map[string]interface{}{
+							"api.ssl.enabled": "false",
+						}},
+						Services: []v1alpha1.LogstashService{{
+							Name: LogstashAPIServiceName,
+							TLS: commonv1.TLSOptions{
+								SelfSignedCertificate: &commonv1.SelfSignedCertificate{
+									Disabled: false,
```
robbavey, line 155: This test passes regardless of whether the value here has the TLS block, or whether `selfSignedCertificate.Disabled` value is `true` or `false`, as only the `useTLS` flag is relevant here. 

I think we are missing a layer of testing, of how the `useTLS` setting is derived from the `TLSOptions`

robbavey (commented):
I left a comment on testing. 

Also - what is the effort to make the `api.*` settings support setting with secrets and ENV variables, rather than putting the password in clear in the `Config` object? 

Running through the example and seeing:
```
Spec:
  Config:
    api.auth.basic.password:  i_am_rich
    api.auth.basic.username:  batman
    api.auth.type:            basic
```
when running `k describe logstashes.logstash.k8s.elastic.co logstash-sample` is something we should fix before we GA this feature

kaisecheng:
@robbavey  Thanks for testing this PR.
I am making api.* related configs to support ${VAR} from ENV and keystore that defined in Logstash Spec. The key value pairs of keystore are only taken from Spec.SecureSettings. When the same key defined in both places, keystore takes precedence. As Logstash supports ${VAR:default_value}, operator also needs to support in the same way

robbavey (approved):
LGTM on what is up so far.

pkg/controller/logstash/config.go
```diff
@@ -85,3 +116,158 @@ func defaultConfig() *settings.CanonicalConfig {
...
+		Password:         pw,
+	}
+}
+
+// patchWithPatternValue matches the pattern ${VAR:default_value} against config and assign the default value
+// It gives a map of string (VAR) and config pointer that need to further resolve
+// VAR is the variable name that expect to be defined in Keystore or Env
+//
+//	The variable name can consist of digit, underscores and letters
+//	The default value is optional, ${VAR:} and ${VAR} are valid.
```
robbavey, line 181: I wonder if we need to support the `default` case, and if we can just stick with the standard `${VAR}`? If we do support the `default` case, we should log that we are using the default

As far as supporting default values, I think that allowing defaults will only work when there are already ENV variables set in the `podTemplate` container definition as it stands, so the use case for supporting defaults is potentially narrow, and potentially a source of confusion. 

Also, I'm not sure if allowing defaults is typical in Kubernetes - I don't believe that `envFrom` supports it's use

pkg/controller/logstash/config.go
```diff
@@ -85,3 +116,158 @@ func defaultConfig() *settings.CanonicalConfig {
...
+	}
+
+	return data
+}
+
+// getKeystoreEnvKeyValue gives a map that consolidate all key value pairs from user defined environment variables
+// and Keystore from SecureSettings. If the same key defined in both places, keystore takes the precedence.
+func getKeystoreEnvKeyValue(params Params) (map[string]string, error) {
+	data := make(map[string]string)
+	c := getLogstashContainer(params.Logstash.Spec.PodTemplate.Spec.Containers)
```
robbavey, line 206: The return value from `getLogstashContainer` should be checked - the value will be `nil` with a CRD definition not containing an explicit container definition in the `podTemplate` spec, which will cause a crash in subsequent checks of the `Env` and `EnvFrom`, etc below.

docs/orchestrating-elastic-stack-applications/logstash.asciidoc
```diff
@@ -699,6 +700,97 @@ spec:
...
+        - name: logstash
+          envFrom:
+            - secretRef:
+                name: logstash-api-secret <2>
+----
+
+<1> Store the username and password in a Secret.
+<2> Map the username and password to the environment variables of the Pod.
+<3> At Logstash startup, `${API_USERNAME}` and `${API_PASSWORD}` are replaced by the value of environment variables. Check links:https://www.elastic.co/guide/en/logstash/current/environment-variables.html[using environment variables] for more details.
+
```
robbavey, line 746: Maybe a reference somewhere to using the logstash keystore?

robbavey (changes requested):
Still working through it, but adding my thoughts as I go

pkg/controller/logstash/config.go
kaisecheng: Supporting default value is mainly for keeping consistent of the feature that Logstash provide. The default value is used when the ${VAR} is missing in env and keystore. It allows Logstash starts without raising error.

```yaml
kind: Logstash
metadata:
  name: logstash-sample
spec:
  count: 1
  version: 8.11.1
  config:
    api.auth.type: basic
    api.auth.basic.username: "${API_USERNAME:batman}"
    api.auth.basic.password: "${API_PASSWORD:i_am_rich}"
```

Supporting default is not a thing for Kubernetes, but for Stack Monitoring and ReadinessProbe query monitoring endpoint.

pkg/controller/logstash/config.go
```diff
@@ -5,65 +5,96 @@
...
 	"hash"
+	"regexp"
+	"strconv"
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 
-	logstashv1alpha1 "github.com/elastic/cloud-on-k8s/v2/pkg/apis/logstash/v1alpha1"
+	"github.com/elastic/cloud-on-k8s/v2/pkg/apis/logstash/v1alpha1"
```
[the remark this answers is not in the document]
kaisecheng: revert it.

pkg/controller/logstash/config.go
robbavey: Yeah, that makes sense, but I'd still like to see a log if we do revert to `default` values (not logging the default values...)

But I'm still on the fence whether we need the complexity of adding support for default values for ENV variables for this feature only, as I suspect adding default values for usernames and passwords is not going to be a massively used feature, and may be something we can document our way around.

robbavey (commented):
I think we're close to where I'm happy with it - I have a nit on the docs, and a question on the default values for ENV substitution. 

I also wonder if we need to document the keystore password?

pkg/controller/logstash/config.go
kaisecheng: Agree having default value for username password is a small use case and not a good practice. This makes me think why Logstash supports this feature in `logstash.yml` at the first place. For pipelines config, this is a handy features for business logic. However, for logstash config, all settings are important, so, if I expect a ${VAR} and it is missing, I only want to see error and stop Logstash.

Probably, the default value is just a fallback setup to disable basic authentication and SSL in testing phase.

```yaml
api.ssl.enabled: "${SSL_ENABLED:false}"
api.auth.basic.username: "${API_USERNAME:}"
api.auth.basic.password: "${API_PASSWORD:}"
```

I think it is acceptable to remove the feature. Do you see any downside to keep it? I am open to remove the feature.

kaisecheng:
> I also wonder if we need to document the keystore password?

Do you mean documenting that the operator sets the default password for TLS keystore? Or explain the whole picture of `api.ssl.keystore.password` to init container setting the keystore with openssl command with password?

robbavey:
> > I also wonder if we need to document the keystore password?
> 
> Do you mean documenting that the operator sets the default password for TLS keystore? Or explain the whole picture of `api.ssl.keystore.password` to init container setting the keystore with openssl command with password?

I think documenting that there is a default password for the TLS keystore, and that it can be changed by setting the `api.ssl.keystore.password` value in the configuration would be sufficient. There is prior art of default values for keystore default passwords with `keytool` using `changeit` by default when creating keystores

pkg/controller/logstash/config.go
robbavey: Yeah, for business logic in pipelines, it's very useful, but it feels a little less so in this case, and silently falling back to defaults for passwords is something I worry about.

How much effort is it to remove? And how much would it simplify?

kaisecheng: Let's remove it then. I will need to update doc, test cases and the config.go

robbavey (approved):
LGTM

kaisecheng:

barkbay (commented):
I wanted to try it using https://github.com/elastic/cloud-on-k8s/blob/df9bf399fc7fcf0e4e987c8f8ff30b41ac6fc126/config/samples/logstash/logstash_es.yaml but I got the following error:

```
k logs logstash-sample-ls-0 -c logstash-internal-init-config
Setup Logstash keystore for API server
bash: line 15: openssl: command not found
```

Is there any version restriction?

pkg/controller/logstash/driver.go
```diff
@@ -37,6 +40,8 @@ type Params struct {
 
 	OperatorParams    operator.Parameters
 	KeystoreResources *keystore.Resources
+	UseTLS            bool               // Service of API Server uses TLS
+	APIServerConfig   *configs.APIServer // resolved API server config
```
pebrc, line 44: I was wondering whether we really need both `APIServerConfig` and `UseTLS`  especially given that we have a `SSLEnabled` field in `configs.APIServer`. I think it absolutely correct to validate the service TLS config from the CRD and the Logstash config don't contradict each other, as you do. But once that is done could we not just rely on `APIServerConfig`?

pkg/controller/logstash/config.go
```diff
@@ -85,3 +116,161 @@ func defaultConfig() *settings.CanonicalConfig {
...
+	sslEnabled := config.SSLEnabled
+	if (svcUseTLS == sslEnabled) || (!useTLS && sslEnabled == "") {
+		return nil
+	}
+
+	return fmt.Errorf("API Service `spec.services.tls.selfSignedCertificate.disabled` is set to `%t`, but logstash config `api.ssl.enabled` is set to `%s`", !useTLS, sslEnabled)
+}
+
+// resolveAPIServerConfig gives ExpectedAPIServer with the resolved ${VAR} value
+func resolveAPIServerConfig(cfg *settings.CanonicalConfig, params Params) (*configs.APIServer, error) {
```
pebrc, line 144: Why does this return a pointer? Initially when I read this I thought the APIServer config might be optional but it seems its not. As far as I can see it tis always populated except for the error case (where we could return an empty struct).

pebrc (commented):
I started to go through it but need to take a break. It looks good so far, just a few cosmetic comments. I have not started with testing.

kaisecheng:
> Is there any version restriction?

@barkbay Are you getting the docker image released by Elastic? openssl command is there since 7.17.0 and I can see it is in your testing version 8.11.1

Running `openssl version` in the container returns `OpenSSL 1.1.1f  31 Mar 2020`
```
docker run --rm -it --entrypoint openssl docker.elastic.co/logstash/logstash:8.11.1 version
```

barkbay (commented):
> @barkbay Are you getting the docker image released by Elastic? openssl command is there since 7.17.0 and I can see it is in your testing version 8.11.1

Not in the UBI images until version `8.12.0` IIUC:

## :x: docker.elastic.co/logstash/logstash-ubi8:8.11.1
```
docker run --rm --entrypoint openssl  docker.elastic.co/logstash/logstash-ubi8:8.11.1 version
docker: Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: exec: "openssl": executable file not found in $PATH: unknown.
```

## ✅  docker.elastic.co/logstash/logstash:8.11.1
```
docker run --rm --entrypoint openssl  docker.elastic.co/logstash/logstash:8.11.1 version 
OpenSSL 1.1.1f  31 Mar 2020
```

## ✅  docker.elastic.co/logstash/logstash-ubi:8.12.0
```
docker run --rm --entrypoint openssl  docker.elastic.co/logstash/logstash-ubi:8.12.0 version 
OpenSSL 1.1.1k  FIPS 25 Mar 2021
```

Which means that this feature is likely to break Logstash deployments for users on OpenShift until `8.12.0`.

kaisecheng:
@barkbay You are right and thanks for testing it! Sadly, this is the case. I would update the minimum [stack version](https://github.com/elastic/cloud-on-k8s/blob/main/pkg/apis/logstash/v1alpha1/validations.go#L26) to 8.12.0, believe TLS is essential for GA.

pebrc (approved):
LGTM, we probably need a big note in the release notes that we are restricting the version to 8.12 especially for early adopters who are already running Logstash in earlier versions.

test/e2e/logstash/stack_monitoring_test.go
```diff
@@ -42,3 +57,108 @@ func TestLogstashStackMonitoring(t *testing.T) {
...
 	test.Sequence(nil, steps, metrics, logs, monitored).RunSequential(t)
 }
+
+// TestLogstashResolvingDollarVariableInStackMonitoring tests that the dollar sign variable is resolved in correct sequence, and
+// Logstash API server setup correctly, and Beats is able to collect metrics using username password.
+func TestLogstashResolvingDollarVariableInStackMonitoring(t *testing.T) {
+	// only execute this test on supported version
+	if version.MustParse(test.Ctx().ElasticStackVersion).LT(logstashv1alpha1.MinStackMonVersion) {
+		t.SkipNow()
+	}
```
pebrc, line 67: I think this is enforce by the log stash builder already

barkbay (commented):
nit: I noticed this error message while running `TestLogstashResolvingDollarVariableInStackMonitoring`:

```
k logs -n e2e-mercury pod/test-ls-mon-dollar-5hq9-ls-0 elastic-internal-init-keystore
+ keystore_initialized_flag=/usr/share/logstash/config/elastic-internal-init-keystore.ok
+ [[ -f /usr/share/logstash/config/elastic-internal-init-keystore.ok ]]
+ echo 'Initializing keystore.'
Initializing keystore.
+ echo y
+ /usr/share/logstash/bin/logstash-keystore create
Using bundled JDK: /usr/share/logstash/jdk
/usr/share/logstash/vendor/bundle/jruby/3.1.0/gems/concurrent-ruby-1.1.9/lib/concurrent-ruby/concurrent/executor/java_thread_pool_executor.rb:13: warning: method redefined; discarding old to_int
/usr/share/logstash/vendor/bundle/jruby/3.1.0/gems/concurrent-ruby-1.1.9/lib/concurrent-ruby/concurrent/executor/java_thread_pool_executor.rb:13: warning: method redefined; discarding old to_f
ERROR: Failed to load settings file from "path.settings". Aborting... path.setting=/usr/share/logstash/config, exception=LogStash::ConfigurationError, message=>Cannot evaluate `${API_PASSWORD}`. Replacement variable `API_PASSWORD` is not defined in a Logstash secret store or as an Environment entry and there is no default value given.
Sending Logstash logs to /usr/share/logstash/logs which is now configured via log4j2.properties

WARNING: The keystore password is not set. Please set the environment variable `LOGSTASH_KEYSTORE_PASS`. Failure to do so will result in reduced security. Continue without password protection on the keystore? [y/N] [2024-01-22T08:27:35,934][INFO ][org.logstash.secret.store.backend.JavaKeyStore] Created Logstash keystore at /usr/share/logstash/config/logstash.keystore
Created Logstash keystore at /usr/share/logstash/config/logstash.keystore
...
```

The e2e test is eventually ✅ , I was wondering if it's relevant.

kaisecheng:
@barkbay The error message is potentially an item to be improved in Logstash side.  The keystore command notices an unresolved `${VAR}` which should have printed warning instead of error. The keystore is created successfully with correct value, so the test case should be fine.

barkbay (approved):
LGTM

pkg/controller/logstash/configs/api_server.go
```diff
@@ -0,0 +1,24 @@
...
+type APIServer struct {
+	SSLEnabled       string
+	KeystorePassword string
+	AuthType         string
+	Username         string
+	Password         string
+}
+
+func (server APIServer) UseTLS() bool {
+	switch server.SSLEnabled {
```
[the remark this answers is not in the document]
kaisecheng: empty string need to convert to true, but ParseBool() gives false

pkg/controller/logstash/config.go
```diff
@@ -85,3 +117,152 @@ func defaultConfig() *settings.CanonicalConfig {
...
+}
+
+func getEnvKeyValues(params Params, c *corev1.Container, data map[string]string) error {
+	for _, env := range c.Env {
+		data[env.Name] = env.Value
+	}
+
+	for _, envFrom := range c.EnvFrom {
+		// from ConfigMap
+		if envFrom.ConfigMapRef != nil {
```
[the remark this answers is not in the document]
kaisecheng: Yes they are not watched. I think it is okay to let user manages the pod restart to get latest value

kaisecheng:
@pebrc This is ready to merge. Thank you!

PR #7427, opened 2024-01-02
Update NOTICE for new year 2024

thbkrkr:
This updates the `NOTICE.txt` file to reflect the new year (result of `make generate-notice-file`).


pebrc (approved):

barkbay:
FWIW this is already included in https://github.com/elastic/cloud-on-k8s/pull/7417 which is blocked because docs ci is down.

PR #7466, opened 2024-01-15
Logstash adds update strategy

kaisecheng:
Logstash adds `updateStrategy` in CRD which is a [StatefulSetUpdateStrategy](https://pkg.go.dev/k8s.io/api/apps/v1#StatefulSetUpdateStrategy)
The default type is "RollingUpdate"

sample resource

```yaml
kind: Logstash
metadata:
  name: logstash-sample
spec:
  count: 3
  version: 8.11.3
  pipelines:
    - pipeline.id: main
      pipeline.workers: 2
      config.string: |
        output { 
          stdout {}
        }
  updateStrategy: 
    type: "RollingUpdate"
    rollingUpdate:
      partition: 1
      maxUnavailable: 1
```


robbavey (approved):
LGTM!

kaisecheng:

pkg/controller/logstash/logstash_controller_test.go
```diff
@@ -201,6 +210,14 @@ func TestReconcileLogstash_Reconcile(t *testing.T) {
...
 					Version: "8.6.1",
 					Count:   1,
+					UpdateStrategy: appsv1.StatefulSetUpdateStrategy{
+						Type: appsv1.RollingUpdateStatefulSetStrategyType,
+						RollingUpdate: &appsv1.RollingUpdateStatefulSetStrategy{
+							MaxUnavailable: &intstr.IntOrString{
+								IntVal: 1,
+							},
+						},
+					},
```
pebrc, line 220: Not from this PR but I am a bit confused by this test. Not sure what it actually verifies. It seems it it is just retrieving the Logstash resource again that we just put in as an expected object.

pebrc (approved):

pkg/controller/logstash/logstash_controller_test.go
kaisecheng: It verifies the value of LogstashStatus is up-to-date

pebrc: I see. But for the addition of the `UpdateStrategy` a test would be more useful that verifies that the `UpdateStrategy` specified in the Logstash resource is actually applied to the `StatefulSet` that is being created.

kaisecheng: test is updated

pebrc (approved):

Issue #7450, opened 2024-01-09
Logstash crashes with env NODE_NAME


kaisecheng:
## Proposal

**Use case. Why is this important?**
Logstash container crashes before log4j be able to write logstash-plain.log.

## Bug Report
**What did you do?**

```yaml
apiVersion: logstash.k8s.elastic.co/v1alpha1
kind: Logstash
metadata:
  name: logstash-sample
spec:
  count: 1
  version: 8.11.1
  podTemplate:
    spec:
      containers:
        - name: logstash
          env:
            - name: "NODE_NAME"
              value: "Crash!"
```

**What did you expect to see?**

Logstash starts without error

**What did you see instead? Under which circumstances?**

Logstash crashes with CrashLoopBackOff

$ kubectl get pod
```
NAME                      READY   STATUS     RESTARTS     AGE
logstash-sample-ls-0      0/1     Error      1 (2s ago)   3s
```

$ kubectl logs -f logstash-sample-ls-0
```
Defaulted container "logstash" out of: logstash, logstash-internal-init-config (init)
2024/01/09 12:08:42 Setting 'node.name' from environment.
2024/01/09 12:08:42 error: open /usr/share/logstash/config/logstash.yml: read-only file system
```

**Environment**

* ECK version:

	Logstash technical preview. ECK 2.8

**Root cause**

This is an issue relates to logstash image startup process env2yaml, which overwrites the `logstash.yml` regardless of the existing settings in the file.  The source of `logstash.yml` is from Secret which is readonly [mapping](https://github.com/elastic/cloud-on-k8s/blob/v2.8.0/pkg/controller/logstash/initcontainer.go#L34) to mount path. Changing `ln -sf` to `cp` could solve the issue.

PR #7475, opened 2024-01-17
Fix Logstash crashes with env

kaisecheng:
Fixed: https://github.com/elastic/cloud-on-k8s/issues/7450

Logstash crashes when env variable name is in the list of [env2yaml](https://github.com/elastic/logstash/blob/main/docker/data/logstash/env2yaml/env2yaml.go#L50-L155)
This PR changes the config init container to copy the `logstash.yml` to `config` Volume to allow updating the file.

The following resource should start without error
```yaml
kind: Logstash
metadata:
  name: logstash-sample
spec:
  count: 1
  version: 8.11.1
  podTemplate:
    spec:
      containers:
        - name: logstash
          env:
            - name: "NODE_NAME"
              value: "No_Crash!"
```

### DIFF
--- a/pkg/controller/logstash/initcontainer.go
+++ b/pkg/controller/logstash/initcontainer.go
@@ -16,6 +16,8 @@ const (
 	InitConfigContainerName = "logstash-internal-init-config"
 
 	// InitConfigScript is a small bash script to prepare the logstash configuration directory
+	// Logstash rewrites the configuration file (logstash.yml) in the presence of ${VAR} environment variable,
+	// therefore the file is copied to the path instead of creating symbolic link.
 	InitConfigScript = `#!/usr/bin/env bash
 set -eu
 
--- a/pkg/controller/logstash/initcontainer.go
+++ b/pkg/controller/logstash/initcontainer.go
@@ -32,7 +32,7 @@ mount_path=` + volume.InitContainerConfigVolumeMountPath + `
 
 cp -f /usr/share/logstash/config/*.* "$mount_path"
 
-ln -sf ` + volume.InternalConfigVolumeMountPath + `/` + ConfigFileName + ` $mount_path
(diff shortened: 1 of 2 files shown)


robbavey (approved):
Tested - observed previous `CrashLoopBackoff`, now fixed

LGTM

kaisecheng:

pkg/controller/logstash/initcontainer.go
```diff
@@ -32,7 +32,7 @@ mount_path=` + volume.InitContainerConfigVolumeMountPath + `
 
 cp -f /usr/share/logstash/config/*.* "$mount_path"
 
-ln -sf ` + volume.InternalConfigVolumeMountPath + `/` + ConfigFileName + ` $mount_path
+cp ` + volume.InternalConfigVolumeMountPath + `/` + ConfigFileName + ` $mount_path
```
pebrc, line 35: Let's add a comment explaining that Logstash rewrites the configuration file in the presence of environment variable and that that's why we have to copy it.

pkg/controller/logstash/initcontainer.go
```diff
@@ -32,7 +32,7 @@ mount_path=` + volume.InitContainerConfigVolumeMountPath + `
...
 ln -sf ` + volume.InternalPipelineVolumeMountPath + `/` + PipelineFileName + ` $mount_path
```
pebrc, line 36: Are pipeline files not rewritten?

pebrc (approved):

pkg/controller/logstash/initcontainer.go
kaisecheng: pipeline files are not rewritten

kaisecheng:
@pebrc This is ready to merge. Thank you.

PR #7477, opened 2024-01-17
Fix dry-run of helm charts during release.

naemono:
We've seen multiple issues in this area: https://github.com/elastic/cloud-on-k8s/pull/6953, https://github.com/elastic/cloud-on-k8s/pull/7045

It appears we can't pass environment variables when uploading buildkite pipelines. This explicitly sets the `dry-run` flag to `false` during the release pipeline.

### DIFF
--- a/.buildkite/pipeline-release-helm.yml
+++ b/.buildkite/pipeline-release-helm.yml
@@ -24,7 +24,7 @@ steps:
     commands:
       - buildkite-agent artifact download "bin/releaser" /usr/local/
       - chmod u+x /usr/local/bin/releaser
-      - releaser --env=dev --charts-dir=deploy/eck-operator
+      - releaser --env=dev --charts-dir=deploy/eck-operator --dry-run=false
     agents:
       image: docker.elastic.co/ci-agent-images/cloud-k8s-operator/buildkite-agent:bfddf2b3
       memory: "2G"
@@ -45,7 +45,7 @@ steps:
     commands:
       - buildkite-agent artifact download "bin/releaser" /usr/local/
       - chmod u+x /usr/local/bin/releaser
-      - releaser --env=dev --charts-dir=deploy/eck-stack
+      - releaser --env=dev --charts-dir=deploy/eck-stack --dry-run=false
     agents:
       image: docker.elastic.co/ci-agent-images/cloud-k8s-operator/buildkite-agent:bfddf2b3
       memory: "2G"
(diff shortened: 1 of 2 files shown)


.buildkite/pipeline-release-helm.yml
```diff
@@ -24,7 +24,7 @@ steps:
     commands:
       - buildkite-agent artifact download "bin/releaser" /usr/local/
       - chmod u+x /usr/local/bin/releaser
-      - releaser --env=dev --charts-dir=deploy/eck-operator
+      - releaser --env=dev --charts-dir=deploy/eck-operator --dry-run=false
```
thbkrkr, line 27: A downside is that we will no longer be able to test helm chart releases with `--dry-run`.

Maybe we could add `env: HELM_DRY_RUN: ${HELM_DRY_RUN:-true}` at the top of the `.buildkite/pipeline-release-helm.yml` pipeline. This way we can still trigger a build via the API ([trigger-helm-release.sh](https://github.com/elastic/cloud-on-k8s/blob/84e69cd50a81c96314e058f8d61e0eac4ee01431/hack/helm/release/trigger-helm-release.sh#L36-L45)).

And uses `sed` to set `HELM_DRY_RUN` to `false` in the parent pipeline .buildkite/pipeline-release.yml`.

```diff
diff --git a/.buildkite/pipeline-release-helm.yml b/.buildkite/pipeline-release-helm.yml
index 031f35385..696627e7f 100644
--- a/.buildkite/pipeline-release-helm.yml
+++ b/.buildkite/pipeline-release-helm.yml
@@ -1,3 +1,6 @@
+env:
+  HELM_DRY_RUN: ${HELM_DRY_RUN:-true}
+
 steps:

   - label: ":go: helm releaser tool"
diff --git a/.buildkite/pipeline-release.yml b/.buildkite/pipeline-release.yml
index 025b76c23..eb2e1c22f 100644
--- a/.buildkite/pipeline-release.yml
+++ b/.buildkite/pipeline-release.yml
@@ -29,6 +29,7 @@ steps:
         if: | # merge-main or tags
           ( build.branch == "main" && build.source != "schedule" )
           || build.tag != null
-        command: buildkite-agent pipeline upload .buildkite/pipeline-release-helm.yml
-        env:
-          HELM_DRY_RUN: false
+        command: |
+          sed "s|HELM_DRY_RUN:.*|HELM_DRY_RUN: false|" .buildkite/pipeline-release-helm.yml \
+            | buildkite-agent pipeline upload
+
```

naemono: This makes sense. I've adjusted to this approach @thbkrkr

thbkrkr: I did a quick test. It works as expected. I still spotted a trap if we would want to upload the pipeline several times with different values. If the two pipelines are uploaded in //, the value of the env var is unpredictable and overwritten by the last upload. 

https://buildkite.com/elastic/cloud-on-k8s-operator/builds/7014 gives false and false.
https://buildkite.com/elastic/cloud-on-k8s-operator/builds/7025 without // (uses a wait) gives true and false.

thbkrkr (approved):

PR #6391, opened 2023-02-08
Minor documentation fixes

thbkrkr:
Fixes two small documentation issues submitted by community users in PRs that have been pending for some time.
* Fix install command in Helm chart readme (#6375)
* Update default value for set-default-security-context flag (#6347)

Resolves #6335.

PR #7486, opened 2024-01-18
Use eck-operator UBI image when ubiOnly=true.

### LINKED ISSUE #7485: ECK Operator helm chart should use UBI image for operator when `ubiOnly` is `true`
Currently the helm chart for the ECK Operator only sets the `ubi-only` flag in the ConfigMap when the `config.ubiOnly` value is set. It doesn't control the image that is run by the Operator itself. We should use `docker.elastic.co/eck/eck-operator-ubi:version` when `ubiOnly` is `true`.

naemono:
resolves #7485 

Ensure we use the `-ubi` operator image when `ubiOnly` is set to `true`.

```
❯ helm template eck-operator . -s templates/statefulset.yaml | yq '.spec.template.spec.containers[0].image'
docker.elastic.co/eck/eck-operator:2.12.0-SNAPSHOT

❯ helm template eck-operator . -s templates/statefulset.yaml --set=config.ubiOnly=true | yq '.spec.template.spec.containers[0].image'
docker.elastic.co/eck/eck-operator-ubi:2.12.0-SNAPSHOT
```

### DIFF
--- a/deploy/eck-operator/templates/statefulset.yaml
+++ b/deploy/eck-operator/templates/statefulset.yaml
@@ -43,7 +43,7 @@ spec:
         {{- toYaml . | nindent 8 }}
       {{- end }}
       containers:
-        - image: "{{ .Values.image.repository }}:{{ default .Chart.AppVersion .Values.image.tag }}"
+        - image: "{{ .Values.image.repository }}{{- if .Values.config.ubiOnly -}}-ubi{{- end -}}:{{ default .Chart.AppVersion .Values.image.tag }}"
           imagePullPolicy: {{ .Values.image.pullPolicy }}
           name: manager
           args:


thbkrkr:
We don't build a ubi snapshot today (`docker.elastic.co/eck/eck-operator-ubi:2.12.0-SNAPSHOT`. We need to identify the type of trigger that requires the image and update `BUILD_FLAVORS` in the `operator-image.sh` script based on that.

Example if it is only for nightly `main` builds:
```diff
diff --git a/.buildkite/scripts/common/operator-image.sh b/.buildkite/scripts/common/operator-image.sh
index 64d62b073..4c4e81ba3 100644
--- a/.buildkite/scripts/common/operator-image.sh
+++ b/.buildkite/scripts/common/operator-image.sh
@@ -50,7 +50,8 @@ operator::set_build_flavors_var() {
     if [[ "${BUILD_FLAVORS:-}" == "" ]]; then
         case $trigger in
             tag-*)           BUILD_FLAVORS="eck,eck-dev,eck-fips,eck-ubi,eck-ubi-fips" ;;
-            *-main)          BUILD_FLAVORS="eck,eck-dev" ;;
+            merge-main)      BUILD_FLAVORS="eck,eck-dev" ;;
+            nightly-main)    BUILD_FLAVORS="eck,eck-dev,eck-ubi" ;;
             *-test-snapshot) BUILD_FLAVORS="eck,eck-dev" ;;
             pr-*|merge-xyz)  BUILD_FLAVORS="eck" ;;
             dev)             BUILD_FLAVORS="dev" ;;

```

naemono:
> We don't build `-ubi` snapshot of the operator image today (`docker.elastic.co/eck/eck-operator-ubi:2.12.0-SNAPSHOT`). We need to identify the type of trigger that requires the image and update `BUILD_FLAVORS` in the `operator-image.sh` script based on that.

Thank you. Upon looking through these options, I suspect `nighty-main`, `tag-final`, and `tag-bc` probably make the most sense, as we want to do this testing during BC iterations, after a release, and are discussing adding something in the nightly. I'll get these updated.

naemono:
> We don't build `-ubi` snapshot of the operator image today

This just happened to be me testing on a branch where snapshot was set, and no way was intended to mean that we should build these on for all SNAPSHOT builds...

thbkrkr (commented):
Code LGTM but are we okay with changing the existing behavior by forcing the use of the `-ubi` image when the Helm `config.ubiOnly` value is enabled.

barkbay (commented):
Let's maybe update the  comment here if this setting does not only control the stack images: https://github.com/elastic/cloud-on-k8s/blob/3ab46dd999b0cab60ae2bf82df718a2912821e1d/deploy/eck-operator/values.yaml#L226

Let's also create an issue so we don't forget to mention this change in the next release notes?

Other than that, I think I'm ok with this change, but @thbkrkr's comment is relevant, curious to know if there are other opinions.

naemono:
> Let's also create an issue so we don't forget to mention this change in the next release notes?

#7485 exists and is tagged with the next release. Is this what you were intending?

deploy/eck-operator/values.yaml
```diff
@@ -223,7 +223,7 @@ config:
   # Interval between observations of Elasticsearch health, non-positive values disable asynchronous observation.
   elasticsearchObservationInterval: 10s
 
-  # ubiOnly specifies whether the operator will use only UBI container images to deploy Elastic Stack applications. UBI images are only available from 7.10.0 onward.
+  # ubiOnly specifies whether the operator will use only UBI container images to deploy Elastic Stack applications as well as its own deployment. UBI images are only available from 7.10.0 onward.
```
thbkrkr, line 226: Also update: https://github.com/elastic/cloud-on-k8s/blob/c1182524a6874f04f27df7b12496f889173d2d65/cmd/manager/main.go#L308 and regenerate the api-docs to synchronize `./docs/operating-eck/operator-config.asciidoc`?

barkbay:
> #7485 exists and is tagged with the next release. Is this what you were intending?

Sorry, I meant an issue to maybe add a short sentence to mention this change in the highlights.

deploy/eck-operator/values.yaml
naemono: @thbkrkr I don't think these are auto-generated. I see previous manual updates https://github.com/elastic/cloud-on-k8s/pull/6391/files#diff-be6fe99c5b6e0b3cfcd8302de12ec7855fc827001d2496a7257095af8bad66d5R42. I just manually updated this.

naemono:
> Sorry, I meant an issue to maybe add a short sentence to mention this change in the highlights.

I've added the `release-highlight` tag to that issue as well.

deploy/eck-operator/values.yaml
thbkrkr: Sorry, thinking a bit more, it doesn't really make sense as you don't use the operator to deploy the operator. It only makes sense in the context of the Helm chart.

Also, rereading it, I wonder if "deployment" is not confusing because the operator is deployed by default in a Statefulset. IIRC a deployment is used only when deploying on OpenShift using OperatorHub.

naemono: > Also, rereading it, I wonder if "deployment" is not confusing because the operator is deployed by default in a Statefulset.

Good point, I've re-worded it. I also removed the comment from the flag, as I agree, it didn't make sense in that context.

barkbay:
> I've added the release-highlight tag to that issue as well.

I have to confess that as a release manager I'm not sure I've ever used it 😬 . We should maybe update our release process.

naemono:
> > I've added the release-highlight tag to that issue as well.
> 
> I have to confess that as a release manager I'm not sure I've ever used it 😬 . We should maybe update our release process.

I just updated the documentation/template mentioning this tag during writing of the release highlights.

deploy/eck-operator/values.yaml
thbkrkr: I think I still see the comment in the flag (file `cmd/manager/main.go`): https://github.com/elastic/cloud-on-k8s/pull/7486/files#diff-435eecb1d2af40ee96747f88ab71eb2758da989c92f76dcb1f714fc1b300c633.

naemono: > I think I still see the comment in the flag (file `cmd/manager/main.go`): https://github.com/elastic/cloud-on-k8s/pull/7486/files#diff-435eecb1d2af40ee96747f88ab71eb2758da989c92f76dcb1f714fc1b300c633.

My apologies, this has been removed.

thbkrkr (approved):

PR #7491, opened 2024-01-22
Temporarily disable data stream check in TestFleet*IntegrationRecipe e2e tests in 8.12.0

thbkrkr:
This disables the check of the existence of the data stream `metrics-elastic_agent.filebeat-default` with stack version `8.12.x and 8.13.x` in the following e2e tests:
- TestFleetAgentWithoutTLS
- TestFleetMode
- TestFleetKubernetesIntegrationRecipe
- TestFleetAPMIntegrationRecipe

Relates to #7389.

Issue #7389, opened 2023-12-18
TestFleet* seems to be flaky (again)


naemono:
See #6494 and #6331 

TestFleet* seems to be flaky again.  Seeing this again:

```
elasticsearch client failed for 
https://test-agent-fleet-hh26-es-internal-http.e2e-lkeiy-mercury.svc:9200/_data_stream/metrics-elastic_agent.filebeat-default: 
404 Not Found: {Status:404 Error:{CausedBy:{Reason: Type:} 
Reason:no such index [metrics-elastic_agent.filebeat-default] 
Type:index_not_found_exception StackTrace: 
RootCause:[{Reason:no such index [metrics-elastic_agent.filebeat-default] Type:index_not_found_exception}]}}
```

The errors from Fleet agent that seem relevant:

```json
{
  "log.level": "warn",
  "@timestamp": "2023-12-17T23:31:24.333Z",
  "message": "Filebeat is unable to load the ingest pipelines for the configured modules because the Elasticsearch output is not configured/enabled. If you have already loaded the ingest pipelines or are using Logstash pipelines, you can ignore this warning.",
  "component": {
    "binary": "filebeat",
    "dataset": "elastic_agent.filebeat",
    "id": "filestream-monitoring",
    "type": "filestream"
  },
  "log": {
    "source": "filestream-monitoring"
  },
  "log.origin": {
    "file.line": 331,
    "file.name": "beater/filebeat.go",
    "function": "github.com/elastic/beats/v7/filebeat/beater.(*Filebeat).Run"
  },
  "service.name": "filebeat",
  "ecs.version": "1.6.0"
}
```

https://buildkite.com/elastic/cloud-on-k8s-operator-nightly/builds/396#018c79bc-99cf-48d8-af3e-17fb84b4176c

thbkrkr:
I wonder if the culprit would not be:

```json
{
    "log.level": "error",
    "@timestamp": "2023-12-22T00:23:04.168Z",
    "message": "add_cloud_metadata: received error failed with http status code 404",
    "component": {
        "binary": "filebeat",
        "dataset": "elastic_agent.filebeat",
        "id": "filestream-monitoring",
        "type": "filestream"
    },
    "log": {
        "source": "filestream-monitoring"
    },
    "log.logger": "add_cloud_metadata",
    "log.origin": {
        "file.line": 173,
        "file.name": "add_cloud_metadata/providers.go",
[...]

PR #7497, opened 2024-01-24
Factorization by introducing WithFleetAgentDataStreamsValidation()

thbkrkr:
`TestAgentVersionUpgradeToLatest8x` is also affected by #7389. 

This commit refactors all validations done in Fleet E2E tests with disabling the problematic validation based on version, so that all tests benefit from it.

Relates to #7491, #7389.

### DIFF
--- a/test/e2e/test/agent/builder.go
+++ b/test/e2e/test/agent/builder.go
@@ -181,6 +181,23 @@ func (b Builder) WithESValidation(validation ValidationFunc, outputName string) 
 	return b
 }
 
+func (b Builder) WithFleetAgentDataStreamsValidation() Builder {
+	v := version.MustParse(test.Ctx().ElasticStackVersion)
+	b = b.
+		WithDefaultESValidation(HasWorkingDataStream(LogsType, "elastic_agent", "default")).
+		WithDefaultESValidation(HasWorkingDataStream(LogsType, "elastic_agent.filebeat", "default")).
+		WithDefaultESValidation(HasWorkingDataStream(LogsType, "elastic_agent.fleet_server", "default")).
+		WithDefaultESValidation(HasWorkingDataStream(LogsType, "elastic_agent.metricbeat", "default")).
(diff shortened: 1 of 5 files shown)


thbkrkr:
buildkite test this -f p=gke,E2E_TAGS=agent -m s=8.11.0,s=8.12.0,s=8.13.0-SNAPSHOT

test/e2e/test/agent/builder.go
```diff
@@ -181,6 +181,23 @@ func (b Builder) WithESValidation(validation ValidationFunc, outputName string)
...
+func (b Builder) WithFleetAgentDataStreamsValidation() Builder {
+	v := version.MustParse(test.Ctx().ElasticStackVersion)
+	b = b.
+		WithDefaultESValidation(HasWorkingDataStream(LogsType, "elastic_agent", "default")).
+		WithDefaultESValidation(HasWorkingDataStream(LogsType, "elastic_agent.filebeat", "default")).
+		WithDefaultESValidation(HasWorkingDataStream(LogsType, "elastic_agent.fleet_server", "default")).
+		WithDefaultESValidation(HasWorkingDataStream(LogsType, "elastic_agent.metricbeat", "default")).
+		WithDefaultESValidation(HasWorkingDataStream(MetricsType, "elastic_agent.elastic_agent", "default")).
+		WithDefaultESValidation(HasWorkingDataStream(MetricsType, "elastic_agent.fleet_server", "default")).
+		WithDefaultESValidation(HasWorkingDataStream(MetricsType, "elastic_agent.metricbeat", "default"))
```
thbkrkr, line 193: Note that the two following:
```
		WithDefaultESValidation(HasWorkingDataStream(LogsType, "elastic_agent", "default")).
		WithDefaultESValidation(HasWorkingDataStream(MetricsType, "elastic_agent.fleet_server", "default")).
```
were not always done in all Fleet* E2E tests, but I don't think there is a good reason why we shouldn't do them.

naemono (approved):
I think this makes sense, and when we have to make changes to fleet e2e tests (common) they are more centralized. ty

PR #7509, opened 2024-01-26
Update version for 2.11.1 release.

naemono:
Updating versions files in preparation for 2.11.1 release.


docs/eck-attributes.asciidoc
```diff
@@ -1,4 +1,4 @@
-:eck_version: 2.11.0
+:eck_version: 2.11.1
 :eck_crd_version: v1
 :eck_release_branch: 2.11
```
naemono, line 3: I double checked some old tags (2.6.1 for example) to ensure this release branch stayed the same..

hack/operatorhub/config.yaml
```diff
@@ -1,4 +1,4 @@
-newVersion: 2.11.0
+newVersion: 2.11.1
 prevVersion: 2.10.0
```
naemono, line 2: I double checked some old tags (2.6.1 for example) to ensure this also was not incremented to 2.11.0.

barkbay:
```suggestion
prevVersion: 2.11.0
```

> I double checked some old tags (2.6.1 for example) to ensure this also was not incremented to 2.11.0.

Is it not because `2.6.0` was never released in the operator hub?  See this PR for example https://github.com/elastic/cloud-on-k8s/pull/6322/files.

image

barkbay (approved):
LGTM

PR #7535, opened 2024-02-07
Update Elastic Stack version to v8.12.1

thbkrkr:
This bumps the current Elastic Stack version from `8.12.0` to `8.12.1`

using `hack/update-stack-version.sh 8\.12\.0 8.12.1`.

Also `mv config/samples/logstash/logstash_pv.yml config/samples/logstash/logstash_pv.yaml` so it is updated by the above script.

And finally updates `hack/upgrade-test-harness/conf.yaml`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -421,7 +421,7 @@ drivah-build-e2e:
 
 # -- run
 
-E2E_STACK_VERSION          ?= 8.12.0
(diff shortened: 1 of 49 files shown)


barkbay (approved):
LGTM

PR #7536, opened 2024-02-07
Update upgrade-test-harness to use latest patch Elastic Stack versions

thbkrkr:
This updates `upgrade-test-harness` to use latest patch versions: `8.7.1`, `8.8.2`, `8.9.2` and `8.11.4`.

`8.12.x` will be updated in #7535.

### DIFF
--- a/hack/upgrade-test-harness/conf.yaml
+++ b/hack/upgrade-test-harness/conf.yaml
@@ -28,16 +28,16 @@ testParams:
     stackVersion: 8.6.2
   - name: v270
     operatorVersion: 2.7.0
-    stackVersion: 8.7.0
+    stackVersion: 8.7.1
   - name: v280
     operatorVersion: 2.8.0
-    stackVersion: 8.8.0
+    stackVersion: 8.8.2
   - name: v290
     operatorVersion: 2.9.0
-    stackVersion: 8.9.0
+    stackVersion: 8.9.2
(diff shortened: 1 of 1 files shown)


barkbay (approved):
LGTM

hack/upgrade-test-harness/conf.yaml
```diff
@@ -28,16 +28,16 @@ testParams:
...
-    stackVersion: 8.7.0
+    stackVersion: 8.7.1
   - name: v280
     operatorVersion: 2.8.0
-    stackVersion: 8.8.0
+    stackVersion: 8.8.2
   - name: v290
     operatorVersion: 2.9.0
-    stackVersion: 8.9.0
+    stackVersion: 8.9.2
```
barkbay, line 37: I missed that we should also update the stack manifests: https://github.com/elastic/cloud-on-k8s/blob/39fae91af962ddf1ec3d03afff4999f2dbe54794/hack/upgrade-test-harness/testdata/v290/stack.yaml#L7

Current tests are failing with:

```
[FAIL] CheckElasticsearch[v290] {"step": "TestStatusOfResources[v290]", "step": "CheckElasticsearch[v290]", "error": "status mismatch: want={health:green nodes:3 version:8.9.2} have={health:green nodes:3 version:8.9.0}: retriable error"}
```

PR #7560, opened 2024-02-14
Allow HTTPs/TLS for metrics endpoint

naemono:
## What is this change?

This allows users to configure HTTPs on the exposed Prometheus metrics port.

### Details

* This is disabled by default, but we note the intent to enable by default in v2.14.0.
* This uses a self-signed certificate for this initial implementation (which is handled purely within controller-runtime). If we want to potentially allow for providing a certificate we can add that as a feature in a future PR prior to enabling by default in v2.14.0 (or future decided release).

### Local testing

Target configuration:
image

Metrics data:
image

```
❯ cat ~/tmp/kube-prometheus.values.yaml
prometheus:
  prometheusSpec:
    serviceMonitorNamespaceSelector: {}
    podMonitorNamespaceSelector: {}
    podMonitorSelectorNilUsesHelmValues: false # allows not having to set release labels on podmonitor to have it scraped
    serviceMonitorSelectorNilUsesHelmValues: false

❯ helm upgrade -i elastic-operator -n elastic-system . --reuse-values --set=installCRDs=true --set=image.repository=---cut--- --set=image.tag=---cut--- --set=config.metricsPort=9090 --set=config.enableSecureMetrics=true --set=podMonitor.enabled=true
```

Podmonitor:
```
❯ kc get podMonitor -n elastic-system elastic-operator -o yaml
apiVersion: monitoring.coreos.com/v1
kind: PodMonitor
metadata:
  annotations:
    meta.helm.sh/release-name: elastic-operator
    meta.helm.sh/release-namespace: elastic-system
  creationTimestamp: "2024-02-14T16:38:34Z"
  generation: 1
  labels:
    app.kubernetes.io/instance: elastic-operator
    app.kubernetes.io/name: elastic-operator
    app.kubernetes.io/version: 2.12.0-SNAPSHOT
    helm.sh/chart: eck-operator-2.12.0-SNAPSHOT
  name: elastic-operator
  namespace: elastic-system
  resourceVersion: "184783956"
  uid: b3d5e92a-b3a8-4a10-9271-5bd58d58e5f8
spec:
  namespaceSelector:
    matchNames:
    - elastic-system
  podMetricsEndpoints:
  - interval: 5m
    path: /metrics
    port: metrics
    scheme: https
    scrapeTimeout: 30s
    tlsConfig:
      insecureSkipVerify: true
  selector:
    matchLabels:
      app.kubernetes.io/instance: elastic-operator
      app.kubernetes.io/name: elastic-operator
```

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -288,6 +288,11 @@ func Command() *cobra.Command {
 		DefaultMetricPort,
 		"Port to use for exposing metrics in the Prometheus format (set 0 to disable)",
 	)
+	cmd.Flags().Bool(
+		operator.EnableSecureMetricsFlag,
+		false,
+		"Enables serving metrics over HTTPS (Will be enabled by default in v2.14.0)",
+	)
 	cmd.Flags().StringSlice(
 		operator.NamespacesFlag,
 		nil,
@@ -581,7 +586,8 @@ func startOperator(ctx context.Context) error {
 		log.Info("Exposing Prometheus metrics on /metrics", "port", metricsPort)
 	}
 	opts.Metrics = metricsserver.Options{
-		BindAddress: fmt.Sprintf(":%d", metricsPort), // 0 to disable
+		BindAddress:   fmt.Sprintf(":%d", metricsPort), // 0 to disable
+		SecureServing: viper.GetBool(operator.EnableSecureMetricsFlag),
 	}
 
 	webhookPort := viper.GetInt(operator.WebhookPortFlag)
--- a/config/eck.yaml
+++ b/config/eck.yaml
@@ -1,5 +1,6 @@
 log-verbosity: 0
 metrics-port: 0
+enable-secure-metrics: false
 container-registry: docker.elastic.co
 max-concurrent-reconciles: 3
 ca-cert-validity: 8760h
--- a/deploy/eck-operator/templates/configmap.yaml
+++ b/deploy/eck-operator/templates/configmap.yaml
@@ -10,6 +10,7 @@ data:
   eck.yaml: |-
     log-verbosity: {{ int .Values.config.logVerbosity }}
     metrics-port: {{ int .Values.config.metricsPort }}
+    enable-secure-metrics: {{ .Values.config.enableSecureMetrics }}
     container-registry: {{ .Values.config.containerRegistry }}
     {{- with .Values.config.containerSuffix }}
     container-suffix: {{ . }}
--- a/deploy/eck-operator/templates/podMonitor.yaml
+++ b/deploy/eck-operator/templates/podMonitor.yaml
@@ -28,6 +28,11 @@ spec:
       {{- with .Values.podMonitor.podMetricsEndpointConfig }}
         {{- toYaml . | nindent 6 }}
       {{- end }}
+      {{- if .Values.config.enableSecureMetrics }}
+      scheme: https
+      tlsConfig:
+        insecureSkipVerify: true
+      {{- end }}
   namespaceSelector:
     matchNames:
       - {{ .Release.Namespace }}
--- a/deploy/eck-operator/values.yaml
+++ b/deploy/eck-operator/values.yaml
@@ -162,6 +162,9 @@ config:
   # metricsPort defines the port to expose operator metrics. Set to 0 to disable metrics reporting.
   metricsPort: "0"
 
+  # enableSecureMetrics specifies whether to enable secure metrics. When enabled, the operator will serve metrics over HTTPS. (Will be enabled by default in v2.14.0)
+  enableSecureMetrics: false
+
   # containerRegistry to use for pulling Elasticsearch and other application container images.
   containerRegistry: docker.elastic.co
 
--- a/docs/operating-eck/operator-config.asciidoc
+++ b/docs/operating-eck/operator-config.asciidoc
@@ -37,6 +37,7 @@ ECK can be configured using either command line flags or environment variables.
 |manage-webhook-certs |true |Enables automatic webhook certificate management.
 |max-concurrent-reconciles |3 | Maximum number of concurrent reconciles per controller (Elasticsearch, Kibana, APM Server). Affects the ability of the operator to process changes concurrently.
 |metrics-port |0 |Prometheus metrics port. Set to 0 to disable the metrics endpoint.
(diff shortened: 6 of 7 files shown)


barkbay (commented):
👋 , 2 questions regarding this change:
* Do we want to introduce something like `--metrics-bind-address` to restrict the traffic to the localhost interface and enforce the use of something like kube rbac proxy?
* Any reason for postponing the use of custom certificates instead of the self signed ones?

naemono:
> * Do we want to introduce something like `--metrics-bind-address` to restrict the traffic to the localhost interface and enforce the use of something like kube rbac proxy?

Well, this change is coming in #7567, which are the RBAC additions. Since working on that option, I've found that enabling auth/rbac forces TLS/HTTPs, which bleeds into this feature a bit, so we likely need to consider both it seems.



> Any reason for postponing the use of custom certificates instead of the self signed ones?

Primarily for simplicity sake. Since feature freeze is 7 days away, this seemed like a simple change we could get into the release by then, and if we wanted to work on allowing custom certificates at a future date it seems to be supported by [metricsserver](sigs.k8s.io/controller-runtime/pkg/metrics/server).Options.TLSOpts:
```
	// TLSOpts is used to allow configuring the TLS config used for the server.
	// This also allows providing a certificate via GetCertificate.
	TLSOpts []func(*tls.Config)
```

also kube-rbac-proxy has these options pertaining to certificates (it also will generate a self-signed certificate if none exists):
```
      --tls-cert-file string                        File containing the default x509 Certificate for HTTPS. (CA cert, if any, concatenated after server cert)
      --tls-private-key-file string                 File containing the default x509 private key matching --tls-cert-file.
```

It feels like we need something similar to this (Helm chart specific)

```
metrics:
  port: 0 // difficult since we already have `metricsPort`
  enableRBAC: false
  certificateSecret: // would allow for self-signed if not populated.
    name: ""
    namespace: ""
```

barkbay:
>> Do we want to introduce something like --metrics-bind-address to restrict the traffic to the localhost interface and enforce the use of something like kube rbac proxy?

> Well, this change is coming in https://github.com/elastic/cloud-on-k8s/pull/7567, which are the RBAC additions.

I'm not sure to understand how #7567 helps to restrict the traffic on the localhost interface. This is how the metrics endpoint is protected in a project generated by a recent version of `kubebuilder`:

```yaml
# This patch inject a sidecar container which is a HTTP proxy for the
# controller manager, it performs RBAC authorization against the Kubernetes API using SubjectAccessReviews.
apiVersion: apps/v1
kind: Deployment
metadata:
  name: controller-manager
  namespace: system
spec:
  template:
    spec:
      containers:
      - name: kube-rbac-proxy
        securityContext:
          allowPrivilegeEscalation: false
          capabilities:
            drop:
            - "ALL"
        image: gcr.io/kubebuilder/kube-rbac-proxy:v0.15.0
        args:
        - "--secure-listen-address=0.0.0.0:8443"
        - "--upstream=http://127.0.0.1:8080/"
        - "--logtostderr=true"
        - "--v=0"
        ports:
        - containerPort: 8443
          protocol: TCP
          name: https
        resources:
          limits:
            cpu: 500m
            memory: 128Mi
          requests:
            cpu: 5m
            memory: 64Mi
      - name: manager
        args:
        - "--health-probe-bind-address=:8081"
        - "--metrics-bind-address=127.0.0.1:8080"
        - "--leader-elect"
 ```

The important part here is the fact that the operator listens only on  `127.0.0.1` thanks to `--metrics-bind-address=127.0.0.1:8080`.  This ensures that there is no traffic outside of the `Pod`, because only the containers in the `Pod` can connect to `127.0.0.1`. My point is that as long as we don't offer a similar flag, and [the operator listens on all the network interfaces](https://github.com/elastic/cloud-on-k8s/blob/109566c90d0ea29b7217fc4b0eaa9a1d9f1f4255/cmd/manager/main.go#L584), it is not possible for a user to enforce this restriction.

Should we:

1. Implement `--metrics-bind-address` instead of `--enable-secure-metrics`
2. Document how the `/metrics` endpoint can be secured using:
   1.  `--metrics-bind-address=127.0.0.1:XXXX`
   2. The RBAC proxy configured with `--upstream=http://127.0.0.1:XXXX/"` to expose the `/metrics` endpoint on the K8s cluster using either their own or self signed certificates.

This way the metrics service security is fully delegated to the RBAC proxy, and there is no need for a `enable-secure-metrics` flag (which is still something we can implement later).

pebrc:
I like @barkbay proposal

naemono:
> I'm not sure to understand how #7567 helps to restrict the traffic on the localhost interface.

Correct, I was referring to #7567 introducing kube-rbac-proxy, not restricting the traffic. I'll investigate this suggestion as well.

naemono:
@barkbay @pebrc I'll pursue the options proposed here, but I think it would be best to pursue this with changes to #7567 , since most of the `kube-rbac-proxy` work is already completed there. The hesitation I had to implementing the `kube-rbac-proxy` in this PR is because you can’t use it without enabling RBAC, so this basically ties the “tls/https” issue into the “auth/rbac” issue.  So if we want to implement just the `auth/rbac` proposal while securing the localhost interface with a `metrics-bind-address` option, I’d be fine with that, but it doesn’t allow `TLS` without `RBAC` and I just wanted to explicitly note that.

If we all agree on the above, let me know and I'll close this PR, and concentrate on this work in #7567

naemono:
> Document how the `/metrics` endpoint can be secured using:
> 2. The RBAC proxy configured with `--upstream=http://127.0.0.1:XXXX/"` to expose the `/metrics` endpoint on the K8s cluster using either their own or self signed certificates.

And for clarification, do we want to just document how to do this, or enable this in the helm chart, as I am pretty positive we can't even do this with the current state of the helm chart, as it requires an additional sidecar, which our helm chart doesn't support, among other missing pieces, such as additional clusterroles/clusterrolebindings, along with needing a `ServiceMonitor`, etc (see #7567 for details)

pebrc:
I think we have two groups of users. Those who use our all-in-one YAML manifest and those that use Helm (I think the split is roughly 50/50, with a few percent using OLM or the certified offering). For the ones using the manifest the `--metrics-bind-address` option + documentation how to kustomize a secure setup would be relevant. 

In #7567 we can then offer a much more convenient solution for our Helm users that builds on top of `kube-rbac-proxy` to offer TLS+RBAC and prevents access to the manager metrics port as described by @barkbay just by setting a few values when running Helm.

naemono:
closing in favor of #7567

PR #7582, opened 2024-02-27
Remove noisy stack config policy log line

### LINKED ISSUE #7529: Remove noisy stack config policy log line
https://github.com/elastic/cloud-on-k8s/blob/1fa362a198e38cc000730e09b7cb8424569c167e/pkg/controller/stackconfigpolicy/secure_setings.go#L32

This turns out to be quite a noise log line that we probably should remove or adjust since Logstash can have secure settings as well and we will end up with many instances of this message in the logs that might alert users even though there is no problem.

thbkrkr:
Resolves #7529.


naemono (approved):
👍

PR #7587, opened 2024-02-29
Ensure statefulset name label is set for Logstash pods

### LINKED ISSUE #7545: Logstash resources are not reconciled when upgrading from ECK 2.11.x to 2.12.0-SNAPSHOT
## How to reproduce

1. Deploy ECK 2.11.1
2. Apply the sample resource [`config/samples/logstash/logstash_es.yaml`](https://raw.githubusercontent.com/elastic/cloud-on-k8s/main/config/samples/logstash/logstash_es.yaml)
3. Wait for the Logstash Pod to be created
4. Upgrade ECK to `main`

The following log message is printed endlessly:

```
2024-02-09T09:10:46.459+0100    DEBUG   logstash-controller     Some pods still need to be created/deleted      {"service.version": "0.0.0-SNAPSHOT+00000000", "iteration": "12", "namespace": "default", "logstash_name": "logstash-sample", "namespace": "default", "statefulset_name": "logstash-sample-ls", "pending_creations": ["logstash-sample-ls-0"], "pending_deletions": []}
```

The root cause is that expectations for Logstash are never considered as "met", it's because label `logstash.k8s.elastic.co/statefulset-name` is originally missing on `Pods` created with ECK `2.11.1`:

```
> k get pods
NAME                                READY   STATUS    RESTARTS   AGE
elasticsearch-sample-es-default-0   1/1     Running   0          8m22s
elasticsearch-sample-es-default-1   1/1     Running   0          8m22s
logstash-sample-ls-0                1/1     Running   0          8m21s ### Created with ECK 2.11.1

> k get pods -l logstash.k8s.elastic.co/statefulset-name=logstash-sample-ls
No resources found in default namespace.
```

Therefore existing Pods are never returned by [GetActualPodsForStatefulSet](https://github.com/elastic/cloud-on-k8s/blob/309ab6d5cb8193550e0138af22a98f4d9a75c5d5/pkg/controller/common/statefulset/pod.go#L41).

thbkrkr:
Resolves #7545.

### DIFF
--- a/pkg/controller/logstash/driver.go
+++ b/pkg/controller/logstash/driver.go
@@ -208,6 +208,9 @@ func ensureStsNameLabelIsSetOnPods(params Params) error {
 		return nil
 	}
 	// add the missing label and update the sts resource
+	if sts.Spec.Template.Labels == nil {
+		sts.Spec.Template.Labels = map[string]string{}
+	}
 	sts.Spec.Template.Labels[labels.StatefulSetNameLabelName] = params.Logstash.Name
 	return params.Client.Update(params.Context, &sts)
 }
--- a/pkg/controller/logstash/driver.go
+++ b/pkg/controller/logstash/driver.go
@@ -91,7 +91,7 @@ func internalReconcile(params Params) (*reconciler.Results, logstashv1alpha1.Log
 
 	// ensure that the label used by expectations is set as it is not in place if the logstash
 	// resource was created with ECK < 2.12
-	if err := ensureStsNameLabelIsSetOnPods(params); err != nil {
+	if err := ensureSTSNameLabelIsSetOnPods(params); err != nil {
 		return results.WithError(err), params.Status
 	}
 
@@ -194,7 +194,7 @@ func isPendingReconciliation(sset appsv1.StatefulSet) bool {
 	return sset.Generation != sset.Status.ObservedGeneration
 }
 
-func ensureStsNameLabelIsSetOnPods(params Params) error {
+func ensureSTSNameLabelIsSetOnPods(params Params) error {
 	sts, err := retrieveActualStatefulSet(params.Client, params.Logstash)
 	if apierrors.IsNotFound(err) {
 		// maybe the sts doesn't exist yet or was deleted, let it be (re)created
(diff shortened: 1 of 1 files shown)


pkg/controller/logstash/driver.go
```diff
@@ -187,3 +193,21 @@ func podReconciliationDone(ctx context.Context, c k8s.Client, sset appsv1.Statef
 func isPendingReconciliation(sset appsv1.StatefulSet) bool {
 	return sset.Generation != sset.Status.ObservedGeneration
 }
+
+func ensureStsNameLabelIsSetOnPods(params Params) error {
```
pebrc, line 197:
```suggestion
func ensureSTSNameLabelIsSetOnPods(params Params) error {
```

pkg/controller/logstash/driver.go
```diff
@@ -187,3 +193,21 @@ func podReconciliationDone(ctx context.Context, c k8s.Client, sset appsv1.Statef
...
+	if err != nil {
+		return err
+	}
+	if _, ok := sts.Spec.Template.Labels[labels.StatefulSetNameLabelName]; ok {
+		// label is already in place, great
+		return nil
+	}
+	// add the missing label and update the sts resource
+	sts.Spec.Template.Labels[labels.StatefulSetNameLabelName] = params.Logstash.Name
+	return params.Client.Update(params.Context, &sts)
```
pebrc, line 212: I was wondering if we should requeue due to cache lag but I guess if would do that automatically as long as the label is missing in the cache.

pebrc (approved):

pkg/controller/logstash/driver.go
```diff
@@ -89,6 +89,12 @@ func internalReconcile(params Params) (*reconciler.Results, logstashv1alpha1.Log
 	defer tracing.Span(&params.Context)()
 	results := reconciler.NewResult(params.Context)
 
+	// ensure that the label used by expectations is set as it is not in place if the logstash
+	// resource was created with ECK < 2.12
+	if err := ensureStsNameLabelIsSetOnPods(params); err != nil {
```
thbkrkr, line 94:
```suggestion
	if err := ensureSTSNameLabelIsSetOnPods(params); err != nil {
```

PR #7589, opened 2024-03-01
Adjusting the version check to disable fleet e2e test validation

kvalliyurnatt:
Disable the fleet e2e test validation for version 8.14.0 as well, since ware seeing test failures https://elastic.slack.com/archives/C8RU9J1C2/p1709282906128319

Like the original issue 
https://github.com/elastic/cloud-on-k8s/issues/7389

### DIFF
--- a/test/e2e/test/agent/builder.go
+++ b/test/e2e/test/agent/builder.go
@@ -192,7 +192,7 @@ func (b Builder) WithFleetAgentDataStreamsValidation() Builder {
 		WithDefaultESValidation(HasWorkingDataStream(MetricsType, "elastic_agent.fleet_server", "default")).
 		WithDefaultESValidation(HasWorkingDataStream(MetricsType, "elastic_agent.metricbeat", "default"))
 	// https://github.com/elastic/cloud-on-k8s/issues/7389
-	if v.LT(version.MinFor(8, 12, 0)) || v.GE(version.MinFor(8, 14, 0)) {
+	if v.LT(version.MinFor(8, 12, 0)) || v.GE(version.MinFor(8, 15, 0)) {
 		b = b.WithDefaultESValidation(HasWorkingDataStream(MetricsType, "elastic_agent.filebeat", "default"))
 	}
 	return b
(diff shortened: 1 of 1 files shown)


test/e2e/test/agent/builder.go
```diff
@@ -192,7 +192,7 @@ func (b Builder) WithFleetAgentDataStreamsValidation() Builder {
 		WithDefaultESValidation(HasWorkingDataStream(MetricsType, "elastic_agent.fleet_server", "default")).
 		WithDefaultESValidation(HasWorkingDataStream(MetricsType, "elastic_agent.metricbeat", "default"))
 	// https://github.com/elastic/cloud-on-k8s/issues/7389
-	if v.LT(version.MinFor(8, 12, 0)) || v.GE(version.MinFor(8, 14, 0)) {
+	if v.LT(version.MinFor(8, 12, 0)) || v.GE(version.MinFor(8, 15, 0)) {
```
naemono, line 195: I'm sorry, I could be sleep deprived, but where is 8.15.x coming from? From reading through #7389, shouldn't we just say:

```
if v.LT(version.MinFor(8, 12, 0)) {
```

Isn't it going to be this way going forward? We don't want to have to keep bumping the `GE` version, right?

thbkrkr (approved):
👍

test/e2e/test/agent/builder.go
kvalliyurnatt: That is a good question, I had the same question as well. But I left it in so we could discuss in this PR. I am not sure  why we have an upper bound on this.

thbkrkr, line 195: You're right, this is the new behaviour!

```suggestion
	if v.LT(version.MinFor(8, 12, 0)) {
```

kvalliyurnatt:
buildkite test this -f p=gke,E2E_TAGS=agent -m s=8.11.0,s=8.12.0,s=8.13.0-SNAPSHOT,s=8.14.0-SNAPSHOT

naemono (approved):
💯

PR #7640, opened 2024-03-18
Fix namespace logged twice

barkbay:
Fix a log in which the namespace is logged twice (which generates an invalid json document):

```json
{"log.level":"info","@timestamp":"2024-03-16T16:00:00.042Z","log.logger":"agent-es","message":"Deleting secret","service.version":"2.10.0+foo","service.type":"eck","ecs.version":"1.4.0","iteration":"14","namespace":"my-ns","agent_name":"my-fleet-server","namespace":"my-ns","secret_name":"my-fleet-server-my-ns-elasticsearch-prod-agent-user","associated_name":"my-fleet-server"}
```


naemono (approved):
lgtm

Issue #7027, opened 2023-07-17
Improve Logstash keystore command performance


kaisecheng:
Logstash keystore init container slows down the startup process. To update the keystore, `logstash-keystore` command starts and stops jvm for each key. The process is proportion to the number of keys. In the local env, adding 10 keys takes six minutes to get Logstash ready.

Logstash keystore command should allow adding keys in a batch.

**Config to reproduce the issue** 

```yml
apiVersion: v1
kind: Secret
metadata:
  name: logstash-secure-settings
stringData:
  A: a
  B: b
  C: c
  D: d
  E: e
  F: f
  G: g
  H: h
  I: i
  J: j
---
apiVersion: logstash.k8s.elastic.co/v1alpha1
kind: Logstash
metadata:
  name: logstash-sample
spec:
  version: 8.8.0
  count: 1
  pipelines:
    - pipeline.id: main
      config.string: |-
        input { exec { command => 'uptime' interval => 10 } }
        filter {
          if ("${HELLO:}" != "") {
            mutate { add_tag => ["awesome"] }
          }
        }
  secureSettings:
    - secretName: logstash-secure-settings
```

relates: https://github.com/elastic/logstash/issues/15186

PR #7642, opened 2024-03-19
Fix Logstash keystore performance

kaisecheng:
Fixes: #7027

Logstash has performance issue when init container creates keystore with a couple of keys. 
Since Logstash 8.12.0, `logstash-keystore` command supports adding multiple keys in one operation. 
This PR uses the command to add keys in a batch. The minimum version of Logstash currently running in ECK is 8.12.0

The steps to create keystore in Logstash
```bash
# create keystore without password
echo 'y' | bin/logstash-keystore create 
# add keys one and two
bin/logstash-keystore add one two
Enter value for one: # input ASCII and non-empty string, and then press <enter>
Added 'one' to the Logstash keystore.
Enter value for two:
Added 'two' to the Logstash keystore.
```

The following resources should start approximately in a minute
```yaml
apiVersion: v1
kind: Secret
metadata:
  name: logstash-secure-settings
stringData:
  HELLO: HALLO
  A: "1"
  B: "2"
  C: "3"
  D: "4"
  E: "5"
  F: "6"
  G: "7"
  H: "8"
  I: "9"
  J: "10"
---
kind: Logstash
metadata:
  name: logstash-sample
spec:
  version: 8.12.0
  count: 1
  pipelines:
    - pipeline.id: main
      config.string: |-
        filter {
          if ("${HELLO:}" != "") {
            mutate { add_tag => ["${A}", "${B}", "${C}", "${D}", "${E}", "${F}", "${G}", "${H}", "${I}", "${J}"] }
          }
        }
  secureSettings:
    - secretName: logstash-secure-settings
```


pkg/controller/common/keystore/initcontainer.go
```diff
@@ -27,6 +27,8 @@ type InitContainerParameters struct {
 	KeystoreAddCommand string
 	// Keystore create command
 	KeystoreCreateCommand string
+	// ContainerCommand is the bash script to run in container
+	ContainerCommand string
```
robbavey, line 31:
```suggestion
	CustomScript string
```

This may be easier to follow, if we call this `CustomScript`, and comment to state that this is a script to run that overrides the default Keystore script?

pkg/controller/logstash/keystore.go
```diff
@@ -22,10 +22,52 @@ const (
...
+touch {{ .KeystoreVolumePath }}/elastic-internal-init-keystore.ok
+{{ end -}}
+
+echo "Keystore initialization successful."
+`
+
 	initContainersParameters = keystore.InitContainerParameters{
-		KeystoreCreateCommand:         keystoreCommand + " create",
-		KeystoreAddCommand:            keystoreCommand + ` add "$key" < "$filename"`,
+		KeystoreCreateCommand:         "echo 'y' | /usr/share/logstash/bin/logstash-keystore create",
```
robbavey, line 69: For consistency, we might want to think about using both `KeystoreCreateCommand` and `KeystoreAddCommand` here, and in the script, or just calling out to `logstash-keystore` directly in the script.

robbavey (changes requested):
Works as advertised, just a couple of naming nits

pkg/controller/logstash/keystore.go
kaisecheng: added the command for consistency

pkg/controller/logstash/keystore.go
```diff
@@ -22,10 +22,53 @@ const (
...
+		KeystoreAddCommand:            `echo -e "$vals" | bin/logstash-keystore add "${keys[@]}"`,
```
robbavey, line 70:
```suggestion
		KeystoreAddCommand:            `echo -e "$vals" | /usr/share/logstash/bin/logstash-keystore add "${keys[@]}"`,
```

Nit: Inconsistent invocation of `logstash-keystore`

robbavey (commented):
LGTM other than small nit about using the same executable path consistently

kaisecheng:
@pebrc @barkbay this is ready for ECK team to review

pebrc (approved):
LGTM

kaisecheng:
@pebrc thanks for the review. This PR is ready to merge

pebrc:
@robbavey can you approve to override your previous "change requested"?

robbavey (approved):
LGTM

robbavey:
Approved

Issue #7664, opened 2024-03-27
Reconciler error - elasticsearch client failed - [update_desired_nodes_request] failed to parse field [nodes]


thbkrkr:
After upgrading ECK 2.12.0, upgrading Elasticsearch from 8.12.2 to 8.13.0 is stalled on the following error:
```go
2024-03-27T11:48:26.479+0100    ERROR   manager.eck-operator    Reconciler error        {
  "service.version": "2.12.0-SNAPSHOT+dbdcc1c5", 
  "controller": "elasticsearch-controller", 
  "object": {"name":"q","namespace":"test"}, "namespace": "test", "name": "q", 
  "reconcileID": "f7ab9df2-ab31-4a34-b07e-c0c56c36e5f0", 
  "error": "elasticsearch client failed for https://q-es-internal-http.test.svc:9200/_internal/desired_nodes/4859b6c6-4d16-41a2-a59b-563dcb9cf34c/3?error_trace=true: 
  400 Bad Request: {Status:400 Error:{CausedBy:{Reason:Required [node_version] Type:illegal_argument_exception}
  Reason:[1:1387] [update_desired_nodes_request] failed to parse field [nodes] Type:x_content_parse_exception
  StackTrace:org.elasticsearch.xcontent.XContentParseException: [1:1387] [update_desired_nodes_request] failed to parse field [nodes]
...
```

thbkrkr:
The bug was introduced in:
- https://github.com/elastic/cloud-on-k8s/pull/7628

The Elasticsearch version from the `Spec` is used to check if the deprecated field needs to be removed or not, instead of the minimum Elasticsearch version in the cluster.

thbkrkr:
Manifest to reproduce:

```yaml
apiVersion: elasticsearch.k8s.elastic.co/v1
kind: Elasticsearch
metadata:
  name: test
spec:
  version: 8.12.0
  nodeSets:
  - name: master
    count: 1
    config:
      node.store.allow_mmap: false
    podTemplate:
      spec:
        containers:
        - name: elasticsearch
          resources:
            requests:
              memory: 4Gi
              cpu: 2
            limits:
              memory: 4Gi
```

thbkrkr:
We should have caught this bug with the e2e-tests suite. We need to update at least one `TestVersionUpgrade` with container CPU resources, so that the desired node API is used during this upgrade.

[...]

PR #7663, opened 2024-03-27
Move desired nodes version gate in the client

thbkrkr:
This moves the version gate for removing the deprecated nodes field when using the desired nodes API in the es client,
so it uses the minimum version of Elasticsearch in the cluster instead of the Elasticsearch version in the Spec.

Resolves #7664.

PR #7661, opened 2024-03-26
Update current Stack version to 8.13.0

thbkrkr:
This updates the current Stack version to `8.3.0`.

PR #7679, opened 2024-03-28
Update e2e test to cover upgrade using desired nodes API

thbkrkr:
This updates `TestVersionUpgradeSingleToLatest8x` to use CPU requests and memory limits, so the desired nodes API is used during an upgrade.

Why update juste one test? I think there are two reasons why we don't set CPU resources by default for all tests: 
- we don't want to have flaky tests due to scheduling issue because of a lack of resources
- we want to have fast tests with the max of CPU

Relates to #7664.

This first updates the test:
* 6a67480ae  Update TestVersionUpgradeSingleToLatest8x with container resources

For testing, I reverted the fix to validate that this change in the test effectively fails without the https://github.com/elastic/cloud-on-k8s/pull/7663 fix:
* a4111029a  Revert "Move desired nodes version gate in the client (7663)"

🟡 The [build](https://buildkite.com/elastic/cloud-on-k8s-operator/builds/7948) succeeded instead of failing? Normal, the test was not doing an ES upgrade to `8.13.0`. Let's update the latest 8x. (Not smart, I could have simply set `t=8.13.0` when triggering the test.) 
* 056437176  -> 8.13.0

🔴 Now the [build](https://buildkite.com/elastic/cloud-on-k8s-operator/builds/7951) failed and we have the expected error (`elasticsearch client failed - [update_desired_nodes_request] failed to parse field [nodes] `) in the operator log in the diagnostics. 

Reverting the revert to now validate that the fix already comitted fixes the new test:
* 209d82ce2 Revert "Revert "Move desired nodes version gate in the client (7663)"

🟢 The [build](https://buildkite.com/elastic/cloud-on-k8s-operator/builds/7955) succeeded.

### DIFF
--- a/test/e2e/es/version_upgrade_test.go
+++ b/test/e2e/es/version_upgrade_test.go
@@ -9,6 +9,9 @@ package es
 import (
 	"testing"
 
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+
 	esv1 "github.com/elastic/cloud-on-k8s/v2/pkg/apis/elasticsearch/v1"
 	"github.com/elastic/cloud-on-k8s/v2/test/e2e/test"
 	"github.com/elastic/cloud-on-k8s/v2/test/e2e/test/elasticsearch"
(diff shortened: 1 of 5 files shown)


test/e2e/test/version.go
```diff
@@ -18,7 +18,7 @@ const (
 	// LatestReleasedVersion7x is the latest released version for 7.x
 	LatestReleasedVersion7x = "7.17.8"
 	// LatestReleasedVersion8x is the latest release version for 8.x
-	LatestReleasedVersion8x = "8.12.1"
+	LatestReleasedVersion8x = "8.13.0"
```
thbkrkr, line 21: Same change in https://github.com/elastic/cloud-on-k8s/pull/7661.

test/e2e/es/version_upgrade_test.go
```diff
@@ -182,13 +185,22 @@ func TestVersionUpgradeSingleToLatest8x(t *testing.T) {
 
 	test.SkipInvalidUpgrade(t, srcVersion, dstVersion)
 
+	resources := corev1.ResourceRequirements{
```
pebrc, line 188: Maybe add a comment explaining your motivation for adding the explicit resource requirement.

pebrc (approved):

PR #7690, opened 2024-04-05
fix(eck-operator): make automountServiceAccountToken configurable

stefan-caraiman:
Keeps the `automountServiceAccountToken` defaulting to `true` just as before for ServiceAccount/Statefulset

It makes it configurable though for cases such as with Azure AKS Security policy which scans for any pod/SA that have it set to `true` and instead recommended as part of their security benchmarks to disable automounting and injecting the service account tokens as volumes explicitly for each workload that requires it: [Kubernetes clusters should disable automounting API credentials](https://portal.azure.com/#blade/Microsoft_Azure_Policy/PolicyDetailBlade/definitionId/%2Fproviders%2FMicrosoft.Authorization%2FpolicyDefinitions%2F423dd1ba-798e-40e4-9c4d-b6902674b423)

A similar [thread](https://github.com/cert-manager/cert-manager/issues/5254) on this from `cert-manager` and [hardened values](https://cert-manager.io/docs/installation/best-practice/#best-practice-helm-chart-values) for such setups & [rationale](https://cert-manager.io/docs/installation/best-practice/#restrict-auto-mount-of-service-account-tokens) behind it.


stefan-caraiman:
@thbkrkr could you please have a look 👀  thanks!

deploy/eck-operator/values.yaml
```diff
@@ -188,6 +193,8 @@ config:
       #       serviceMonitorNamespaceSelector: {}
       #       serviceMonitorSelectorNilUsesHelmValues: false
       enabled: false
+      # additional volume mounts for the kube-rbac-proxy container.
+      volumeMounts: []
```
thbkrkr, line 197: IIUC, if you want to disable automount service account tokens and secure the metrics entrypoint, you will need to define `volumeMounts` twice?

```yaml
volumeMounts:
- mountPath: /var/run/secrets/kubernetes.io/serviceaccount
  name: serviceaccount-token
  readOnly: true

config:
  metrics:
    secureMode:
      volumeMounts:
      - mountPath: /var/run/secrets/kubernetes.io/serviceaccount
        name: serviceaccount-token
        readOnly: true
  ```

**Review by thbkrkr — COMMENTED** [review]
Looks very good overall! I left one question.

Could you please add some unit tests that covers the different cases? Here's a start to get you started:

```yaml
> cat deploy/eck-operator/templates/tests/statefulset_test.yaml
suite: test operator statefulset
templates:
  - statefulset.yaml
  - configmap.yaml
tests:
  - it: should have automount service account tokens set by default
    asserts:
      - template: statefulset.yaml
        equal:
          path: spec.template.spec.automountServiceAccountToken
          value: true
  - it: should disable automount service account tokens
    set:
      automountServiceAccountToken: false
    asserts:
      - template: statefulset.yaml
        equal:
          path: spec.template.spec.automountServiceAccountToken
          value: false
```

deploy/eck-operator/values.yaml
thbkrkr: I'm missing something. In which case, would we have two different `volumeMounts`?

stefan-caraiman: If `automountServiceAccountToken: false` and both the `kube-rbac-proxy` and `manager` containers require to talk to the K8s api we would need to explicitly set in each one of the containers the extra `volumeMounts` with the `serviceaccounttoken` mounted in order to reach K8s api.

Unless I misunderstood and `kube-rbac-proxy` container actually does not require to communicate with the K8s API, in which case the extra `volumeMount` is not required for it 😅 

TLDR: Extra `volumeMounts` should be added to any container still requiring K8s API access in case `automountServiceAccountToken` is `false`

stefan-caraiman:
@thbkrkr added some minimal unit tests, let me know if i should add something else 👍

thbkrkr (approved):

thbkrkr:
buildkite test this

deploy/eck-operator/values.yaml
thbkrkr: I was a bit stuck in my mind with a single definition of volumesMounts for all containers, but there may be cases where you want different additional volumesMounts for a given container, which makes the need to have separate definitions valid per container.

thbkrkr:
buildkite test this

thbkrkr:
Thank you Stephan for this contribution!

PR #7708, opened 2024-04-12
Beats index permissions compatibility with ECS index naming

### LINKED ISSUE #7692: Feature request: Beats predefined role permissions compatibility with ECS guidelines
### Context

The Elastic Common Schema, besides field names and types, also defines a [common index naming schema](https://www.elastic.co/guide/en/ecs/8.13/ecs-data_stream.html) as part of the `data_stream` fields definition.
Elastic has previously published a blog post on this here: https://www.elastic.co/blog/an-introduction-to-the-elastic-data-stream-naming-scheme
These index naming conventions are also followed by Elastic Agent integrations. With the current ECS versions the definition for `data_stream.type` [states](https://www.elastic.co/guide/en/ecs/8.13/ecs-data_stream.html#field-data-stream-type):

> Currently allowed values are "logs" and "metrics". We expect to also add "traces" and "synthetics" in the near future.

### Current problem

Currently the predefined roles definition for beats [here](https://github.com/elastic/cloud-on-k8s/blob/c65eeafc65342d77becb83c6f9d81be14cd0da0f/pkg/controller/elasticsearch/user/roles.go#L203) sets the beats user role to allow index permissions only for that beat types name e.g. `filebeat-*`. This means using ECS compliant index names with beats requires custom role permissions overrides, which are further complicated with the versioning implemented for the beats i.e. overriding that role can break during future cluster upgrades.
Meanwhile logstash already has a predefined role that allows creating indexes with the `logs-*` prefix.

### Proposed solution

Duplicate the current index permissions for `*beat-*`to `logs-*` here: https://github.com/elastic/cloud-on-k8s/blob/c65eeafc65342d77becb83c6f9d81be14cd0da0f/pkg/controller/elasticsearch/user/roles.go#L203

An example on how the defined role could look like:

```
eck_beat_es_filebeat_role_v77:
    cluster:
        - monitor
        - manage_ilm
        - manage_ml
        - read_ilm
        - cluster:admin/ingest/pipeline/get
    indices:
        - names:
            - filebeat-*
            - logs-*
          privileges:
            - manage
            - read
            - create_doc
            - view_index_metadata
            - create_index
    applications: []
```

Would this be a change the maintainers find acceptable? If so I can raise a PR to implement the change.

haam3r:
Grant beats additional default permissions allowing to create indexes that follow the Elastic Common Schema.

Closes #7692 which contains context for the change.

### DIFF
--- a/pkg/controller/elasticsearch/user/roles.go
+++ b/pkg/controller/elasticsearch/user/roles.go
@@ -61,7 +61,7 @@ const (
 	V77 = "v77"
 
 	// Additional index permissions for Beats users
-	BeatsAdditionalPermissions = [string]{"logs-*"}
+	BeatsAdditionalPermissions = []string{"logs-*"}
 )
 
 var (
--- a/pkg/controller/elasticsearch/user/roles.go
+++ b/pkg/controller/elasticsearch/user/roles.go
@@ -200,7 +200,7 @@ func init() {
 			Cluster: []string{"monitor", "manage_ilm", "manage_ml", "read_ilm", "cluster:admin/ingest/pipeline/get"},
 			Indices: []esclient.IndexRole{
 				{
-					Names:      []string{fmt.Sprintf("%s-*", beat)},
+					Names:      []string{fmt.Sprintf("%s-*", beat), "logs-*"},
 					Privileges: []string{"manage", "read", "create_doc", "view_index_metadata", "create_index"},
 				},
 			},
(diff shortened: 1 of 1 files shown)


haam3r:
@pebrc I currently only added permissions for `logs-*`, however the ECS naming schema also includes `"metrics-*"`. Should I add a metrics prefix too?

pebrc:
@haam3r that's a good question. In principle we want to run with minimal permissions for each Beat, but I am not sure if we can come up with a clear mapping for each Beat. I will try to find out more.

pebrc:
@haam3r we would want this indeed to be specific to the beat involved: 
* Filebeat is “logs”
* Metricbeat is “metrics”
* Packetbeat is “logs”
* Auditbeat is “logs”
* Heartbeat is “synthetics”

haam3r:
@pebrc Gotcha, I'll see about reworking the PR to make the permissions beat specific

haam3r:
@pebrc Does that look better?

pebrc:
buildkite test this

pebrc:
buildkite test this -f p=gke,E2E_TAGS=beat

pebrc (approved):
LGTM, it would be nice to have a unit test for this given that we have a bit more complexity with the lookup now.

PR #7709, opened 2024-04-12
Update elasticsearchRef when elastic is external instance in kibana.asciidoc

kvashninselectel:
Set elasticsearchRef to "". Otherwise, operator won't create an instance, cause by default the instance has ref name "elasticsearch".

### DIFF
--- a/docs/orchestrating-elastic-stack-applications/kibana.asciidoc
+++ b/docs/orchestrating-elastic-stack-applications/kibana.asciidoc
@@ -81,6 +81,8 @@ spec:
     elasticsearch.hosts:
       - https://elasticsearch.example.com:9200
     elasticsearch.username: elastic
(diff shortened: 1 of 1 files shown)


thbkrkr:
Hello,

Could you explain what you are trying to solve with this change?

When you find a bug, we really appreciate it if you start by opening an issue first.

> Set elasticsearchRef to "". Otherwise, operator won't create an instance, cause by default the instance has ref name "elasticsearch".

If `elasticsearchRef` is not set, the operator should not set it to `elasticsearch`. This behavior is not expected. Do you have a YAML manifest to reproduce this?

thbkrkr:
Please open an issue to provide context for this change. In the meantime, I'm closing it and we'll see if it needs to be reopened depending on the issue. Thank you.

PR #7723, opened 2024-04-17
Allow disabling of elastic user.

### LINKED ISSUE #7719: Potentially allow disabling the creation of the `elastic` user.
Currently the ECK Operator always creates the public-facing `elastic` user with `superuser` role for any Elasticsearch cluster. This issue is created to discuss if we should allow for this to be disabled on a per-cluster basis, and if so what would need to change for this to happen.

Known changes that would have to occur:
1. eck-diagnostics: This uses the `elastic` user by default. We would need to add the option of specifying a username/password combination to use that had sufficient privileges.
2. documentation: We reference this user in our [docs](https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-users-and-roles.html#k8s-default-elastic-user). We would need to augment these to note that this user can be disabled.
3. e2e tests: We would likely want to add an e2e test that disables this user and verifies the lack of the k8s secret.

To discuss:
1. Any concerns for pursuing this?
2. Are there any further integrations that we are unaware of that would cause issues when disabling this user?

naemono:
resolves #7719 

### Primary changes

This change allows a user to disable the `elastic` user from being created upon Elasticsearch creation. A use case for this would be when an organization/user would prefer to manage all users/roles via SSO (SAML/IDP/LDAP/etc).

This adds a new field to the Elasticsearch CRD:

```
kind: Elasticsearch
metadata:
  name: elasticsearch-sample
spec:
  auth:
    disableElasticUser: true
```

This also updates the documentation to reference the new change in multiple sections where the `elastic` user is mentioned.

### Additional changes

Since one of the primary uses of the `elastic` user is to allow our ECK diagnostics to run, a new user with minimal privileges has been created to allow this functionality: `elastic-internal-diagnostics`.

### Testing done

1. Simple elasticsearch cluster creation
2. Full stack creation using ES/Kibana/Fleet server/Agents with file-realm user added; Kibana login/functionality worked without issues, along with Fleet server and all agents worked perfectly.
3. ECK diagnostics run using this version of ECK.

### Open issues
- [ ] The eck-diagnostics call to `/api/exception_lists/items/_find?list_id=endpoint_host_isolation_exceptions&namespace_type=agnostic` is still non-functional. I'm working with the relevant team to try and find the right permissions to adjust.

TODO: 
- [x] Documentation updates
- [x] ~e2e testing~; unit testing seems sufficient for this, as we check for the non-existence of the elastic secret.
- [x] Fix unit tests
- [x] Fix e2e smoke tests


kvalliyurnatt (approved):
LGTM from a code perspective, have not tested it myself.

pkg/controller/elasticsearch/user/reconcile.go
```diff
@@ -106,9 +106,12 @@ func aggregateFileRealm(
 	}
 
 	// reconcile predefined users
-	elasticUser, err := reconcileElasticUser(ctx, c, es, existingFileRealm, userProvidedFileRealm, passwordHasher)
-	if err != nil {
-		return filerealm.Realm{}, esclient.BasicAuth{}, err
+	var elasticUser users
+	if !es.Spec.Auth.DisableElasticUser {
+		elasticUser, err = reconcileElasticUser(ctx, c, es, existingFileRealm, userProvidedFileRealm, passwordHasher)
```
pebrc, line 111: If you move the conditional check into the `reconcileElasticUser` function then non of the code here needs to change. `reconcileElasticUser` would return and empty  `users` slice and that would return an empty realm.

docs/orchestrating-elastic-stack-applications/security/users-and-roles.asciidoc
```diff
@@ -25,6 +25,25 @@ kubectl get secret quickstart-es-elastic-user -o go-template='{{.data.elastic |
 
 To rotate this password, refer to: <<{p}-rotate-credentials>>.
 
+=== Disabling the default `elastic` user
+
+If your prefer to manage all users via SSO, for example using <<{p}-saml-authentication>> or OpenID Connect, you can disable the default `elastic` user by setting the `auth.disableElasticUser` field in the Elasticsearch resource to `true`:
```
pebrc, line 30:
```suggestion
If your prefer to manage all users via SSO, for example using <<{p}-saml-authentication>> or OpenID Connect, you can disable the default `elastic` superuser by setting the `auth.disableElasticUser` field in the Elasticsearch resource to `true`:
```

pkg/controller/elasticsearch/user/roles.go
```diff
@@ -66,6 +69,51 @@ var (
...
+			},
+			Applications: []esclient.ApplicationRole{
+				{
+					Application: "kibana-.kibana",
+					Resources:   []string{"*"},
+					// Unfortunately the following privileges are not sufficient access all Kibana APIs
+					// that are required for the ECK diagnostics. In the future we should try again to
+					// generate a more fine-grained role and not use "*".
+					Privileges: []string{
+						"*",
```
pebrc, line 89: Are you still investigating this or is this the least privileges we can give?

pebrc (commented):
Looks good! I was trying to think if the `disableElasticUser` attribute is the right name but could not come up with something more succinct 👍

pkg/controller/elasticsearch/user/reconcile.go
```diff
@@ -122,12 +122,12 @@ func aggregateFileRealm(
...
-		elasticUser.fileRealm(),
-		associatedUsers.fileRealm(),
-		userProvidedFileRealm,
-	)
+	fileRealmUsers := []filerealm.Realm{internalUsers.fileRealm(), associatedUsers.fileRealm(), userProvidedFileRealm}
+	if !es.Spec.Auth.DisableElasticUser {
+		// not using append here as order matters.
+		fileRealmUsers = []filerealm.Realm{internalUsers.fileRealm(), elasticUser.fileRealm(), associatedUsers.fileRealm(), userProvidedFileRealm}
+	}
+	fileRealm := filerealm.MergedFrom(fileRealmUsers...)
```
pebrc, line 130:
```suggestion
	fileRealm := filerealm.MergedFrom(
		internalUsers.fileRealm(),
		elasticUser.fileRealm(),
		associatedUsers.fileRealm(),
		userProvidedFileRealm,
	)
```

pebrc: No need for the conditional now that you have refactored the `reconcileElasticUser` method

pkg/controller/elasticsearch/user/roles.go
naemono: @pebrc update: I've found that some of these endpoints require platinum licenses, which is why I'm getting unauthorized errors when attempting to call them, even with privileges that should allow me to call them, as I'm running a dev version of ECK `make go-run` which doesn't support enterprise features.

I'm going to try and test this a bit differently and update.

naemono: Ok @pebrc with the current roles, which are as limited as I can get them, I can successfully run eck-diagnostics with no issues. This should be ready for final reviews. An upcoming PR to eck-diagnostics will be upcoming once this is merged.

pebrc (approved):
LGTM code wise. Did a quick test disabling/enabling the user. I also got the errors on the `api/exception_lists` in the Kibana diagnostics, I am approving nonetheless assuming you will update the permissions as needed.

naemono:
> LGTM code wise. Did a quick test disabling/enabling the user. I also got the errors on the `api/exception_lists` in the Kibana diagnostics, I am approving nonetheless assuming you will update the permissions as needed.

I'm taking one more look at this...

PR #7743, opened 2024-04-24
Bump GKE version in tests to 1.26, AKS 1.29

pebrc:
1.25 is no longer available.

### DIFF
--- a/hack/deployer/config/plans.yml
+++ b/hack/deployer/config/plans.yml
@@ -3,7 +3,7 @@ plans:
   operation: create
   clusterName: ci
   provider: gke
-  kubernetesVersion: 1.28
+  kubernetesVersion: 1.26
   machineType: n1-standard-8
   serviceAccount: true
   enforceSecurityPolicies: true
@@ -18,7 +18,7 @@ plans:
   operation: create
   clusterName: ci-autopilot
   provider: gke
-  kubernetesVersion: 1.28
+  kubernetesVersion: 1.26
   serviceAccount: true
   enforceSecurityPolicies: true
   # this is disabled in autopilot: container provisioner is privileged; not allowed in Autopilot
@@ -32,7 +32,7 @@ plans:
   operation: create
   clusterName: dev
   provider: gke
-  kubernetesVersion: 1.28
+  kubernetesVersion: 1.26
   machineType: n1-standard-8
   serviceAccount: false
   enforceSecurityPolicies: true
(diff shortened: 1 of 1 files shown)


kvalliyurnatt (approved):

thbkrkr (approved):
Let's also update AKS to `1.28.7`?

thbkrkr:
buildkite test this -f t=TestSmoke -m p=gke,p=eks,p=aks^k=1.28.7

pebrc:
Is it not 1.28.5 for AKS?
```
 az aks get-versions --location westeurope --output table
KubernetesVersion    Upgrades
-------------------  -----------------------
1.29.2               None available
1.29.0               1.29.2
1.28.5               1.29.0, 1.29.2
1.28.3               1.28.5, 1.29.0, 1.29.2
1.27.9               1.28.3, 1.28.5
1.27.7               1.27.9, 1.28.3, 1.28.5
1.26.12              1.27.7, 1.27.9
1.26.10              1.26.12, 1.27.7, 1.27.9
```

pebrc:
Or we could even go to 1.29.2

thbkrkr:
`1.29.2` :+1: 

`1.28.7` came from https://learn.microsoft.com/en-us/azure/aks/supported-kubernetes-versions?tabs=azure-cli#azure-portal-and-cli-versions.

Nit: seeing the update title, I wonder if you also inadvertently updated the eks version to `1.28`, which is good (not the inadvertent :) but the update).

pebrc:
> Nit: seeing the update title, I wonder if you also inadvertently updated the eks version to 1.28, which is good (not the inadvertent :) but the update).

Well spotted. We can actuall go to 1.29 there as well.

thbkrkr:
Maybe because your cli is outdated?

pebrc:
buildkite test this -f t=TestSmoke -m p=gke,p=eks,p=aks

pebrc:
Going back to 1.26 for GKE seems to work. Trying 1.27 now to see if that still works too. I don't have a good solution. I tried various things to make the `kubectl replace -f ` work on 1.28 but the sc is recreated too fast and removing the `addonmanager.kubernetes.io/mode` annotation does not seem to have any effect. Going to create a follow up issue. The approach we have chosen to labeling these disks seems quite hacky.

pebrc:
Merging this now. This is not where I wanted it to be but at least it should give us a working CI for the time being (see the follow up issues)

PR #7750, opened 2024-04-26
Do not assume cgroup mem info present

pebrc:
It seems that on k8s > 1.26 (GKE)  the Elasticsearch node_stats response no longer contains cgroup information on 6.8.23. 

Older versions of Elasticsearch < 7.16  cannot parse cgroup v2 structures and have no cgroup information. 
https://github.com/elastic/elasticsearch/blob/4f67856884ff580d8b48a858411b8c17cb0f8cdb/server/src/main/java/org/elasticsearch/monitor/os/OsProbe.java#L499-L513

### DIFF
--- a/test/e2e/test/elasticsearch/checks_es.go
+++ b/test/e2e/test/elasticsearch/checks_es.go
@@ -349,6 +349,12 @@ func compareCgroupMemoryLimit(topologyElement esv1.NodeSet, nodeStats client.Nod
 		// no expected memory, consider it's ok
 		return nil
 	}
+
+	if len(nodeStats.OS.CGroup.Memory.LimitInBytes) == 0 {
+		// no cgroup info (e.g. on GKE 1.26+) consider it ok.
+		return nil
+	}
+
 	// ES returns a string, parse it as an int64, base10
 	actualCgroupMemoryLimit, err := strconv.ParseInt(
 		nodeStats.OS.CGroup.Memory.LimitInBytes, 10, 64,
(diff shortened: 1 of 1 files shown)


pebrc:
buildkite test this -f t=TestSmoke -m p=gke,s=6-8-23

pebrc:
buildkite test this -f t=TestSmoke -m p=gke,s=6.8.23

pebrc:
buildkite test this -f t=TestSmoke,p=gke,s=6.8.23

test/e2e/test/elasticsearch/checks_es.go
```diff
@@ -349,6 +349,13 @@ func compareCgroupMemoryLimit(topologyElement esv1.NodeSet, nodeStats client.Nod
 		// no expected memory, consider it's ok
 		return nil
 	}
+
+	if len(nodeStats.OS.CGroup.Memory.LimitInBytes) == 0 {
+		// Elasticsearch versions before 7.16 cannot parse cgroup v2 information and
+		// will have no information in this field. Considering it ok.
+		return nil
+	}
```
thbkrkr, line 357: If I'm being paranoid, I'd say we should add a version gate, otherwise we might miss a regression in the future where a newer ES version might fail to parse the cgroup info?

pebrc: We can do that but have to look at the version for each pod, because we can have mixed versions during an upgrade where some nodes have the info and some not.

pebrc:
buildkite test this -f t=TestVersionUpgradeTwoNodes68xTo7x,p=gke,s=7.17.8

thbkrkr (approved):
LGTM

PR #7754, opened 2024-04-26
Helm chart changes to allow CRD prevention webhook

### LINKED ISSUE #5117: Prevent unintentional removal of all clusters if CRD is removed
## Proposal

We probably want to try and prevent users from accidentally purging all ES clusters (or any Elastic custom resource for that matter) from their cluster by accidentally deleting the custom resource itself, as the below scenario shows is possible.  This could be caused by a user during maintenance thinking that they could remove, reinstall without issues, or potentially from issues during a `helm upgrade` where the user is wanting to completely remove ECK operator, and reinstall it, not understanding the consequences of their action.  

Example of the issue to prevent:
```
❯ kind create cluster --config config/samples/elasticsearch/custom/kind-cluster.yaml --wait 10m --retain
Creating cluster "kind" ...

❯ helm install elastic-operator elastic/eck-operator -n elastic-system --create-namespace -f ./eck-operator-values.yaml
NAME: elastic-operator
LAST DEPLOYED: Sun Dec  5 14:56:39 2021
NAMESPACE: elastic-system
STATUS: deployed
REVISION: 1
TEST SUITE: None
NOTES:
1. Inspect the operator logs by running the following command:
   kubectl logs -n elastic-system sts/elastic-operator

❯ cat ./eck-operator-values.yaml
webhook:
  enabled: true
managedNamespaces:
- elastic

❯ kubectl apply -f config/samples/elasticsearch/custom/elasticsearch-small-1-member.yaml -n elastic

❯ cat config/samples/elasticsearch/custom/elasticsearch-small-1-member.yaml
kind: Elasticsearch
metadata:
  name: testing
spec:
  version: 7.15.1
  nodeSets:
  - name: masters
    count: 1
    config:
      node.roles: ["master", "data"]
      node.store.allow_mmap: false

❯ kubectl get es -n elastic
NAME      HEALTH   NODES   VERSION   PHASE   AGE
testing   green    1       7.15.1    Ready   96s

❯ kc delete crd elasticsearches.elasticsearch.k8s.elastic.co
customresourcedefinition.apiextensions.k8s.io "elasticsearches.elasticsearch.k8s.elastic.co" deleted

❯ kubectl get pod -n elastic
NAME                   READY   STATUS        RESTARTS   AGE
testing-es-masters-0   1/1     Terminating   0          2m3s
```

### Possible prevention path

We likely could add an additional validating webhook that routes all CRD deletion requests to a validating webhook such as:

```
apiVersion: admissionregistration.k8s.io/v1beta1
kind: ValidatingWebhookConfiguration
metadata:
  name: elastic-crd-deletion-prevention.k8s.elastic.co
webhooks:
- name: elastic-crd-deletion-prevention.k8s.elastic.co
  rules:
  - apiGroups:
    - apiextensions.k8s.io
    apiVersions:
    - v1
    operations:
    - DELETE
    resources:
    - customresourcedefinitions
  clientConfig:
    service:
      namespace: webhooks
      name: eck-webhook-server
      path: /crd
```

Where we would
1. ignore all non-elastic CRDs
2. Query for any instances of the CRD type in the cluster, and if the number is > 0, refuse the request with a detailed explanation of the reasoning.

naemono:
## What are these changes?

resolves #5117

- Since the accidental removal of Elastic CRDs also removes the underlying deployments (removing the Elasticsearch CRD removes all Elasticsearch deployments), this operation can be extremely painful for a user. This change adds an additional webhook to the existing `ValidatingWebhookConfiguration` that will prevent a CRD from being deleted when it detects any existing deployment inside the Kubernetes cluster. If no deployment is detected, then the CRD deletion can proceed.
- This also adds an annotation to existing CRDs installed via Helm `helm.sh/resource-policy: keep` which allows the CRDs to persistent even after a Helm deletion/uninstall happens.

## Testing

Tested both installation via Helm chart, and installation via Yaml manifests, and both procedures described in the documentation successfully block a crd deletion:

```
❯ kubectl delete crd kibanas.kibana.k8s.elastic.co
Error from server (Forbidden): admission webhook "elastic-prevent-crd-deletion.k8s.elastic.co" denied the request: deletion of Elastic CRDs is not allowed while in use
```

Also tested uninstalling the Helm chart, which allows the CRDs to persist:
```
❯ helm uninstall elastic-operator -n elastic-system
These resources were kept due to the resource policy:
[CustomResourceDefinition] kibanas.kibana.k8s.elastic.co
[CustomResourceDefinition] logstashes.logstash.k8s.elastic.co
[CustomResourceDefinition] stackconfigpolicies.stackconfigpolicy.k8s.elastic.co
[CustomResourceDefinition] agents.agent.k8s.elastic.co
[CustomResourceDefinition] apmservers.apm.k8s.elastic.co
[CustomResourceDefinition] beats.beat.k8s.elastic.co
[CustomResourceDefinition] elasticmapsservers.maps.k8s.elastic.co
[CustomResourceDefinition] elasticsearchautoscalers.autoscaling.k8s.elastic.co
[CustomResourceDefinition] elasticsearches.elasticsearch.k8s.elastic.co
[CustomResourceDefinition] enterprisesearches.enterprisesearch.k8s.elastic.co
```

TODO:
- [x] testing
- [x] handle manifest installation method
- [x] documentation for this feature
- [x] documentation for installing/enabling using both helm + manifests.

Open Questions to answer:
* If the helm chart is removed, and the CRDs are a dependency (as is the case in ECK), does this prevent that scenario?
  * It did allow the CRD removal, but the new annotation disallows this.
* In OLM managed installations, is there an option to conditionally install this?
  * ~Investigation in progress.~ I see no options to have conditional parts of an OLM installation See [docs](https://olm.operatorframework.io/docs/tasks/creating-operator-manifests/)


deploy/eck-operator/values.yaml
```diff
@@ -141,6 +141,11 @@ webhook:
   port: 9443
   # secret specifies the Kubernetes secret to be mounted into the path designated by the certsDir value to be used for webhook certificates.
   secret: ""
+  crdDeletionProtection:
+    # determines whether the operator should prevent deletion of Elastic CRDs.
+    # This is a separate configuration because enabling this webhook requires additional RBAC permissions
+    # to read all CRDs within the cluster.
+    enabled: false
```
naemono, line 148: I struggled a bit on the location of this value. Any issues with this?

docs/operating-eck/eck-permissions.asciidoc
```diff
@@ -71,6 +71,7 @@ These permissions are needed by the Service Account that ECK operator runs as.
 |PodDisruptionBudget|policy|no|Ensuring update safety for Elasticsearch. Check link:https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-pod-disruption-budget.html[docs] to learn more.
 |StorageClass|storage.k8s.io|yes|Validating storage expansion support. Check link:https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-volume-claim-templates.html#k8s_updating_the_volume_claim_settings[docs] to learn more.
 |coreauthorization.k8s.io|SubjectAccessReview|yes|Controlling access between referenced resources. Check link:https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-restrict-cross-namespace-associations.html[docs] to learn more.
+|apiextensions.k8s.io|CustomResourceDefinition|yes|Preventing accidental CRD Deletion. Check <<{p}-enabling-crd-prevention-webhook,docs>> to learn more.
```
naemono, line 74: This style of doc link is different than the previous, as I couldn't get a link such as `link:https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-webhook.html#k8s_enabling_the_crd_prevention_webhook[docs]` to work, as the docs build would say "no link found". Any pointers, or is it a non-issue?

pkg/controller/common/webhook/custom_resource_deletion.go
```diff
@@ -0,0 +1,110 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
```
naemono, line 1: I also worry a bit about the location of this file? Any better locations for this? Maybe a new package?

docs/operating-eck/eck-permissions.asciidoc
pebrc: I might be wrong but I think the way you did it with the template is more correct. I think we want to avoid hard-coding absolute URLs in our docs where possible.

pkg/controller/common/webhook/custom_resource_deletion.go
```diff
@@ -0,0 +1,110 @@
...
+	if req.Operation != admissionv1.Delete {
+		return admission.Allowed("")
+	}
+	crd := &extensionsv1.CustomResourceDefinition{}
+	err := wh.decoder.DecodeRaw(req.OldObject, crd)
+	if err != nil {
+		return admission.Errored(http.StatusBadRequest, err)
+	}
+
+	if isElasticCRD(crd) && wh.isInUse(crd) {
```
pebrc, line 69: I wonder if we could limit the impact of this webhook further by using something like [matchConditions](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#matching-requests-matchconditions) in the webhook definition to filter out non-Elastic CRDs to begin with.

pkg/controller/common/webhook/custom_resource_deletion_test.go
```diff
@@ -0,0 +1,103 @@
...
+									{Name: "v1"},
+								},
+							},
+						}),
+					},
+				},
+			},
+			want: admission.Allowed(""),
+		},
+	}
```
pebrc, line 89: Test case with a non-Elastic CRD?

naemono:
> > If the helm chart is removed, and the CRDs are a dependency (as is the case in ECK), does this prevent that scenario?
> 
> No, it appears to allow the complete removal of everything without complaint. I'm investigating our reasoning in the past of using a separate chart (I suspect it was because [this method](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#method-1-let-helm-do-it-for-you) didn't even exist)

After some investigation, I wouldn't recommend changing our approach to how we install CRDs, as we have made additions to our CRDs (which is allowed) but moving them to the `/crds` directory wouldn't allow us to do that in the future as Helm hasn't provided a path forward: (see https://github.com/helm/helm/issues/7735)

I'm investigating another path to ensuring that CRDs are maintained.

https://helm.sh/docs/howto/charts_tips_and_tricks/#tell-helm-not-to-uninstall-a-resource

```yaml
metadata:
  annotations:
    helm.sh/resource-policy: keep
```

pkg/controller/common/webhook/custom_resource_deletion.go
naemono: This is a great idea, I've implemented this, and am testing it now.

naemono: Yeah, I've added this but:

`FEATURE STATE: Kubernetes v1.30 [stable] `

In older versions, this is just ignored in the final webhook:
```
- admissionReviewVersions:
  - v1
  clientConfig:
    service:
      name: elastic-operator-webhook
      namespace: elastic-system
      path: /validate-prevent-crd-deletion-k8s-elastic-co
      port: 443
  failurePolicy: Ignore
  matchPolicy: Exact
  name: elastic-prevent-crd-deletion.k8s.elastic.co
  namespaceSelector: {}
  objectSelector: {}
  rules:
  - apiGroups:
    - apiextensions.k8s.io
    apiVersions:
    - v1
    operations:
    - DELETE
    resources:
    - customresourcedefinitions
    scope: '*'
  sideEffects: None
  timeoutSeconds: 10
```

pebrc:
> helm.sh/resource-policy: keep

Why do we need the webhook if we add this annotation? I think we identified Helm users as the ones to be susceptible to accidental CRD deletion. With the annotation that risk seems to be mitigated. My suggestion would be to only rely on this annotation and scrap the webhook part.

naemono:
> > helm.sh/resource-policy: keep
> 
> Why do we need the webhook if we add this annotation? I think we identified Helm users as the ones to be susceptible to accidental CRD deletion. With the annotation that risk seems to be mitigated. My suggestion would be to only rely on this annotation and scrap the webhook part.

Having the webhook prevents anyone from accidentally running this:

```
❯ kubectl delete crd elasticsearches.elasticsearch.k8s.elastic.co
```

And wiping all of their Elasticsearch instances within a Kubernetes cluster. The addition of the Helm annotation only applies to Helm installs, and not installations using the Yaml manifests. Additionally any Helm user could run the above command, even with the annotation, and wipe all Elasticsearch instances.

Regardless, since the reports of this (#5117) have only come from Helm users, I do understand simply wanting to put the annotation in place for now. @pebrc Let me know if you'd just prefer a new PR with the simple annotation, and I'll get that done and we can shelve this for another time (potentially if/when more reports come in from accidental deletion of CRD with non-Helm users)

naemono:
> > Let me know if you'd just prefer a new PR with the simple annotation, and I'll get that done and we can shelve this for another time (potentially if/when more reports come in from accidental deletion of CRD with non-Helm users)
> 
> I think this would be my preference indeed given the complexity of adding a webhook vs the simplicity of the annotation. My reasoning is that this problem is indeed restricted to Helm users. Of course someone can do a `kubectl delete` but that is an explicit choice. The problem with our operator Helm chart is that we chose to embed the CRDs as a subchart so that if you do `helm uninstall eck-operator` you would accidentally uninstall the CRDs as well. That is a non-obvious implementation detail of the chart that can indeed be easy to overlook especially given that Helm has support for a `crds` subdirectory that we don't use (for good reasons at the time).
> 
> All other distribution channels should be less prone to this kind of operator error:
> 
> * our YAML manifests come in two files, you have to explicitly do `kubectl delete -f crds.yaml` to delete the CRDs
> * everything installed via OLM will keep the CRDs around on uninstall (OpenShift certified operator and OLM community operator via operatorhub)
> * user employing GitOps approaches with tools like ArgoCD will also be immune given that in this example Argo CD also does not delete the CRDs
> 
> When we initially discussed our options to protect against that kind of mistake I was not aware of the existence of this Helm feature, otherwise that would have been the preferred solution from the beginning.
> 
> I am a happy to hear other opinions as well of course but for now the annotation seems the best compromise to address the problem that has been reported.

Sound reasoning. I'll get that PR opened, and we'll close this one once that is open.

naemono:
closing in favor of #7811

PR #7755, opened 2024-04-28
Extend cgroup version gate to all cgroup comparisons

pebrc:
https://github.com/elastic/cloud-on-k8s/pull/7750 had a few flaws: 

* I did allow the CPU related cgroup comparisons to do ahead even if we had no cgroup data
* On k8s versions running cgroup v1 the cgroup based comparison was unnecessarily skipped

This PR tries to address these problems by:

* modelling the cgroup section as optional in the API to check for its presence
* gating all cgroup based comparisons on the presence of the cgroup info in node stats if running ES < 7.16
* but make the absence of cgroup information an error for ES versions that should support it

### DIFF
--- a/pkg/controller/elasticsearch/client/model.go
+++ b/pkg/controller/elasticsearch/client/model.go
@@ -111,16 +111,18 @@ type NodesStats struct {
 type NodeStats struct {
 	Name string `json:"name"`
 	OS   struct {
-		CGroup struct {
-			Memory struct {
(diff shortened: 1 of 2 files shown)


pebrc:
buildkite test this -f t=TestMutationResizeMemoryDown,p=gke -m s=6.8.23,s=7.17.8

thbkrkr (approved):
LGTM

PR #7763, opened 2024-04-30
update version to 8.15 and 2.14

kvalliyurnatt:
Update stack version to 8.15.0-SNAPSHOT and operator version to 2.14.0-SNAPSHOT


naemono (changes requested):
I think you're missing the `deploy/eck-stack/*Chart.yaml` updates in here as well, see https://github.com/elastic/cloud-on-k8s/commit/c6f79eace7c673085006d42072d0be4b82264d27

hack/upgrade-test-harness/conf.yaml
```diff
@@ -42,5 +42,5 @@ testParams:
     operatorVersion: 2.12.1
     stackVersion: 8.13.2
   - name: upcoming
-    operatorVersion: 2.13.0-SNAPSHOT
-    stackVersion: 8.14.0-SNAPSHOT
+    operatorVersion: 2.14.0-SNAPSHOT
+    stackVersion: 8.15.0-SNAPSHOT
```
naemono, line 46: You probably want to make these change in a separate PR. Check https://github.com/elastic/cloud-on-k8s/commit/0b9cf09b091d822d045e7be22faeda7ceeebbd32

naemono (approved):
One change, otherwise g2g.

hack/upgrade-test-harness/conf.yaml
kvalliyurnatt: @naemono so this must be done after the release ?

naemono: No, there are just additional changes that go along with it. You can add them to this PR if you'd like, but if you just remove this, it's good to go. The upgrade-test-harness pieces are a bit more involved.

kvalliyurnatt: aah got it, will open a separate PR for that then.

PR #7784, opened 2024-05-07
Fix diagnostic role pre-8.5

### LINKED ISSUE #7776: elastic-internal-diagnostics may not allow to detect all stack versions
I think the new diagnostic user introduced in #7723 does not always allow https://github.com/elastic/eck-diagnostics to run ES or Kibana diagnostic tools.

Also refer to my comment here: https://github.com/elastic/cloud-on-k8s/commit/8420c9c03764305e2fb3e10273832f774a7cc292#r141719077

naemono:
resolves #7776 

The Elasticsearch role `read_security` did not exist prior to Elasticsearch 8.5. This instead uses the `manage_security` role in previous versions of 8.x, and uses `superuser` role in 7.x which appears to be required.

Testing:
- [x] 8.5+
- [x] < 8.5 (8.4.2)
- [x] 7.17.21 (with superuser privileges)

### DIFF
--- a/pkg/controller/elasticsearch/user/predefined.go
+++ b/pkg/controller/elasticsearch/user/predefined.go
@@ -97,7 +97,11 @@ func reconcileInternalUsers(
 		return nil, fmt.Errorf("while parsing Elasticsearch version (%s): %w", es.Spec.Version, err)
 	}
 	if ver.LT(version.From(8, 5, 0)) {
-		users[4].Roles = []string{DiagnosticsUserRoleV7}
+		// Diagnostics user needs superuser role in 7.x.
+		users[4].Roles = []string{SuperUserBuiltinRole}
+		if ver.GTE(version.From(8, 0, 0)) {
+			users[4].Roles = []string{DiagnosticsUserRoleV80}
+		}
 	}
 	return reconcilePredefinedUsers(
 		ctx,
--- a/pkg/controller/elasticsearch/user/predefined.go
+++ b/pkg/controller/elasticsearch/user/predefined.go
@@ -97,10 +97,15 @@ func reconcileInternalUsers(
 		return nil, fmt.Errorf("while parsing Elasticsearch version (%s): %w", es.Spec.Version, err)
 	}
 	if ver.LT(version.From(8, 5, 0)) {
-		// Diagnostics user needs superuser role in 7.x.
-		users[4].Roles = []string{SuperUserBuiltinRole}
+		// Diagnostics user needs Superuser role in 7.x.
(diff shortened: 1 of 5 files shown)


naemono:
There may be another issue with Kibana privileges in 7.x specific to Fleet. I'm investigating...

pkg/controller/elasticsearch/user/predefined.go
```diff
@@ -83,18 +85,26 @@ func reconcileInternalUsers(
...
+		{Name: ProbeUserName, Roles: []string{ProbeUserRole}},
+		{Name: MonitoringUserName, Roles: []string{RemoteMonitoringCollectorBuiltinRole}},
+		{Name: DiagnosticsUserName, Roles: []string{DiagnosticsUserRoleV85}},
+	}
+	ver, err := version.Parse(es.Spec.Version)
+	if err != nil {
+		return nil, fmt.Errorf("while parsing Elasticsearch version (%s): %w", es.Spec.Version, err)
+	}
+	if ver.LT(version.From(8, 5, 0)) {
+		users[4].Roles = []string{DiagnosticsUserRoleV7}
```
barkbay, line 100: Using an index (`4`) while we want to use a role for a specific user feels error prone to me. If the order of users is changed in a later PR, it would assign the role to another user without being noticed in unit tests in all the cases.

pkg/controller/elasticsearch/user/predefined.go
```diff
@@ -83,18 +85,26 @@ func reconcileInternalUsers(
...
+		{Name: PreStopUserName, Roles: []string{ClusterManageRole}},
+		{Name: ProbeUserName, Roles: []string{ProbeUserRole}},
+		{Name: MonitoringUserName, Roles: []string{RemoteMonitoringCollectorBuiltinRole}},
+		{Name: DiagnosticsUserName, Roles: []string{DiagnosticsUserRoleV85}},
+	}
+	ver, err := version.Parse(es.Spec.Version)
+	if err != nil {
+		return nil, fmt.Errorf("while parsing Elasticsearch version (%s): %w", es.Spec.Version, err)
+	}
+	if ver.LT(version.From(8, 5, 0)) {
```
barkbay, line 99: Let's maybe add a unit test to cover that case.

pkg/controller/elasticsearch/user/predefined.go
naemono: I added a func to replace the roles. lmk if this is the direction you intended.

barkbay (approved):
LGTM

PR #7806, opened 2024-05-13
Remove dependency exclusions no longer necessary

thbkrkr:
This removes dependencies exclusions no longer necessary since the parent dependencies were updated.


naemono (approved):
Seems fine.

Issue #7745, opened 2024-04-24
Storage class parameter update stopped working with k8s 1.27 on GKE


pebrc:
We need to find an alternative way of attribution. Or make sure none of our environments use the built-in storage classes anymore. 

```
storageclass.storage.k8s.io "premium-rwo" deleted
Error from server (AlreadyExists): storageclasses.storage.k8s.io "premium-rwo" already exists
storageclass.storage.k8s.io "premium-rwo" deleted
error: timed out waiting for the condition
storageclass.storage.k8s.io "premium-rwo" deleted
storageclass.storage.k8s.io/premium-rwo replaced
storageclass.storage.k8s.io "standard-rwo" deleted
Error from server (AlreadyExists): storageclasses.storage.k8s.io "standard-rwo" already exists
storageclass.storage.k8s.io "standard-rwo" deleted
Error from server (AlreadyExists): storageclasses.storage.k8s.io "standard-rwo" already exists
storageclass.storage.k8s.io "standard-rwo" deleted
Error from server (AlreadyExists): storageclasses.storage.k8s.io "standard-rwo" already exists
storageclass.storage.k8s.io "standard-rwo" deleted
error: timed out waiting for the condition
storageclass.storage.k8s.io "standard-rwo" deleted
Error from server (AlreadyExists): storageclasses.storage.k8s.io "standard-rwo" already exists
```
Looks like our storage class labeling does not work anymore in GKE 1.27

_Originally posted by @pebrc in https://github.com/elastic/cloud-on-k8s/issues/7743#issuecomment-2075099145_

pebrc:
Idea: install kyverno in all dev/CI environments (should be mostly already the case), add a policy that prevents us from using the built-in GKE storage classes. Create our own storage classes. We already have `e2e-default` but duplicate the netword attched built-ins from GKE and use those with correct attribution and labelling.

PR #7817, opened 2024-05-16
Stop recreating GKE storage classes

pebrc:
Fixes https://github.com/elastic/cloud-on-k8s/issues/7745

This is a pre-requisite for starting to test Kubernetes >= 1.27 on GKE

* create copies of built-in classes with the right labels for disk GC
* create a kyverno policy to prevent the use of the built-in ones
* updates bk agent 

TODO
- [x]  testing with samples that use the built-in classes
- [x] testing with the autopilot recipe which uses a built-in class as an ephemeral volume (should be OK as those autodelete)
- [x] thinking about making kyverno a pre-req on GKE clusters to avoid leaking GCE disks in local dev mode.


pebrc:
buildkite test this -f p=gke-autopilot,t=TestSamples,s=8.13.2

pebrc:
Looks like the test runner got evicted: 

```
cat events.json | jq -r '.items[] |"\(.involvedObject.name): \(.message) \(.lastTimestamp)"'
...
 eck-e2e-h88kj-d54c8: deleting pod for node scale down 2024-05-16T18:48:32Z
```

pebrc:
buildkite test this -f p=gke-autopilot,t=TestSamples,s=8.13.2,k=1.27

pebrc:
buildkite test this -f p=gke,t=TestSamples,s=8.13.2

pebrc:
buildkite test this -f p=gke,t=TestSamples,s=8.13.2

pebrc:
buildkite test this -f p=gke-autopilot,t=TestSamples,s=8.13.2,k=1.27

hack/deployer/runner/gke.go
```diff
@@ -193,38 +202,60 @@ func (d *GKEDriver) setupLabelsForGCEProvider() error {
...
-		if err := retry.OnError(
-			wait.Backoff{Duration: 10 * time.Millisecond, Steps: 5},
-			func(err error) bool { return true },
-			func() error {
-				return exec.NewCommand(fmt.Sprintf(`cat <<EOF | kubectl replace --force -f -
-%s
-EOF`, string(storageClassYaml))).Run()
-			},
-		); err != nil {
+		// start by removing the default storage class label so that our copy can take over that role
```
thbkrkr, line 222: Nit: we don't actually remove the default storage class label, but set it to false?
```suggestion
		// start by making the existing gce storage classes non-default so that our copy can support this role
```

thbkrkr (approved):
Smart, LGTM.

hack/deployer/runner/kyverno/install/gke.yaml
```diff
@@ -0,0 +1,52 @@
+---
```
barkbay, line 1: nit: let's maybe use a more explicit file name? For example`hack/deployer/runner/kyverno/install/disallow-gke-storageclasses.yaml`?

hack/deployer/runner/kyverno/install/gke.yaml
```diff
@@ -0,0 +1,52 @@
...
+            value: 
+            - premium-rwo
+            - standard 
+            - standard-rwo
+  - name: ss-storageclass
+    match:
+      any:
+      - resources:
+          kinds:
+          - StatefulSet
```
barkbay, line 41: Should we not (also) apply this rule at the Pod level?

barkbay:
> thinking about making kyverno a pre-req on GKE clusters to avoid leaking GCE disks in local dev mode

Do we have any reason to not set `enforceSecurityPolicies: true` in `gke-dev`?

pebrc:
> Do we have any reason to not set enforceSecurityPolicies: true in gke-dev?

I think that is already the case

barkbay:
> I think that is already the case

🤦 , just realized that I was creating GKE cluster with `enforceSecurityPolicies: false` 😐

barkbay (approved):
LGTM

PR #7829, opened 2024-05-19
Remove explicit GKE storage class

pebrc:
The recipe specified an explicit storage class. But we are no longer allowing the use of the default classes in e2e tests because we cannot label the disks created by them. I thought that the kyverno policy introduced in https://github.com/elastic/cloud-on-k8s/pull/7817 would ignore ephemeral volumes but I was wrong. This tries to work around the problem by just removing the explicit class name to fall back to the default class (not ideal but 🤷 ).


pebrc:
buildkite test this -f p=gke-autopilot,t=TestAutopilot,s=8.13.2,k=1.27

barkbay (approved):
LGTM

PR #7834, opened 2024-05-22
Route apiVersion must be route.openshift.io/v1

mulbc:
OpenShift Route apiVersion is now route.openshift.io/v1 

If set to v1 we will get a Fatal error:
```
error: resource mapping not found for name: "kibana-sample" namespace: "" from "STDIN": no matches for kind "Route" in version "v1"
```


mulbc:
I signed the CLA now

thbkrkr:
buildkite test this

thbkrkr:
Thank you @mulbc!

PR #7839, opened 2024-05-22
Added support for Kibana Ingress

jamesagarside:
Added support for creating an Ingress resource when using the Kibana Helm chart.


deploy/eck-stack/charts/eck-kibana/templates/ingress.yml
```diff
@@ -0,0 +1,61 @@
+{{- if and .Values.ingress.enabled}}
```
naemono, line 1:
```suggestion
{{- if .Values.ingress.enabled }}
```

deploy/eck-stack/charts/eck-kibana/templates/ingress.yml
```diff
@@ -0,0 +1,61 @@
...
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  name: {{ include "kibana.fullname" . }}-ingress
+  labels:
+    {{- include "kibana.labels" . | nindent 4 }}
+  {{- with .Values.ingress.annotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+    {{- end }}
```
naemono, line 11:
```suggestion
  {{- end }}
```

deploy/eck-stack/charts/eck-kibana/values.yaml
```diff
@@ -42,3 +42,69 @@ spec:
     # will be assumed.
     #
     # namespace: default
+
+ingress:
+  enabled: false
+  labels: {}
+  annotations:
```
naemono, line 49:
```suggestion
  annotations: {}
```

naemono (changes requested):
Took a quick look and saw a couple of things that initially stood out. Will take a longer look later. Would definitely prefer some test cases added for this ingress functionality specific to different types of ingress, like CSP-specific ingress types, and whether or not this format of `values` supports all the standard ingress types.

deploy/eck-stack/charts/eck-kibana/values.yaml
```diff
@@ -42,3 +42,69 @@ spec:
...
+    #           name: ''
+    #           port:
+    #             name: ''
+
+  # -- Additional TLS configuration
+  # @default -- `[]` (See [values.yaml])
+  extraTls: []
+    # - hosts:
+    #   - kibana.example.com
+    #   secretName: your-certificate-name
```
naemono, line 110: Sorry for the delays @jamesagarside in getting back to this. I'm assigned to this now and will follow it through to completion. I needed a better understanding of ingress as a whole, and how CSPs handle them and went down a bit of a rabbit hole working through getting ES and Kibana working in all cloud providers and opened #7941 as a discussion item. I wanted to discuss the differences between your implementation of the values, and what I had come up with https://github.com/elastic/cloud-on-k8s/pull/7941/files#diff-19d8910e9e220072251887cb93679712a3b08bf3dfc24151edaa6304aa72989aR79-R121.

I feel that this implementation in this PR may be overly complex for what our use case is, which is simply getting Kibana exposed to the public internet. If you could, compare the 2 implementations and let me know what you think of the suggestions from my PR. Specifically I'm wondering what the use case is for the `extra*` fields in the values file here in the context of Kibana.

Thanks, and looking forward to your input.

naemono: @jamesagarside and I discussed this, and agreed to go ahead and go with what exists in #7941. We'll close this. Thanks much for kicking off this process to get ingress in place!

jamesagarside:
Thank you @naemono for getting this feature added so quickly. Happy to close

PR #7843, opened 2024-05-23
Use hash for secure settings secret updates

### LINKED ISSUE #7842: Use secret hash instead of resource version to recreate pods when updating secure settings secrets
Problem: If an external tool updates the `-es-secure-settings-` Secret resource version without updating the Secret data, the operator restarts the Elasticsearch nodes.

Cause: When using secure settings Secrets, the operator uses the resource version of the Secrets to compute the final hash set on the Elasticsearch pods label in order to recreate Pods when the Secret changes.

https://github.com/elastic/cloud-on-k8s/blob/86a25bf06e1038855f390e9e4fcc21ebe1baaa26/pkg/controller/common/keystore/user_secret.go#L88-L90

Proposal: use a hash of the Secret data instead of the resource version.

thbkrkr:
This uses the hash of the `-es-secure-settings` Secret data instead of its resource version to know if Pods need to be recreated to update the Keystore.
This will avoid unnecessary restarts of Elasticsearch restarts if the Secret resource version changes without changing the data.

Note: upgrading to the eck version including this new change will result in a graceful restart.

Resolves #7842.

### DIFF
--- a/pkg/controller/apmserver/deployment.go
+++ b/pkg/controller/apmserver/deployment.go
@@ -110,7 +110,7 @@ func buildConfigHash(c k8s.Client, as *apmv1.ApmServer, params PodSpecParams) (s
 
 	// - in the APMServer keystore
 	if params.keystoreResources != nil {
-		_, _ = configHash.Write([]byte(params.keystoreResources.Version))
+		_, _ = configHash.Write([]byte(params.keystoreResources.Hash))
 	}
 
 	// - in the APMServer TLS certificates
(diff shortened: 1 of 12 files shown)


naemono:
@thbkrkr I understand the reasoning behind this, but is this going to cause a rolling-restart for all customers who upgrade to this version? It seems as though it will.

pebrc (approved):
LGTM, I think the rolling restart is unavoidable.

thbkrkr:
Yes, upgrading to the eck version including this new change will result in an unavoidable graceful restart. For users experiencing this problem, this should be the last one without real data change :)

thbkrkr:
I created https://github.com/elastic/cloud-on-k8s/issues/7846 to not forget to update the docs.

PR #7847, opened 2024-05-24
Use readiness port

### LINKED ISSUE #7841: Use Elasticearch readiness port instead of / for Pod readiness
https://github.com/elastic/elasticsearch/issues/50187 introduces a new readiness TCP endpoint we should use to determine pod readiness as it will take cluster membership into account https://www.elastic.co/guide/en/elasticsearch/reference/current/advanced-configuration.html#readiness-tcp-port

Our current readiness probe does call `/` which returns 200 before the node has joined a cluster. When rotating master nodes in quick succession through external means (so not through the rolling upgrade the ECK opererator orchestrates) for example a k8s cluster upgrade or cluster-autoscaler activity, the cluster might become unavailable for a few seconds because of that.

pebrc:
Fixes #7841

As of 8.2.0 use the new `readiness.port` setting to enable a TCP readiness check that is sensitive to cluster membeship of the node. This should improve cluster availablity during external disruptions e.g. due to node upgrades, as the PDB will disallow furhter upgrades until the most recently upgraded node has rejoined the cluster. 

I choose to create a separate file mounted next to the existing script to facilitate upgrades. Specifically I wanted to avoid a situation where during the upgrade we update the scripts configmap overwriting the old probe script with the new one and thereby breaking the not yet upgraded nodes in the cluster. 

The approach taken here creates some techical debt in form of the extra script hanging around when not needed. But I imaging we can drop it when we stop supporting ES < 8.2.0. 

An alternative approach would be to integrate a version check in the existing script which seemed more complicated to reason about to me. 

Marked as bug because a cluster might be unavailable if master nodes are deleted while the previously deleted ones are not yet back in the cluster, which violates our promise of interruption free ES operations when running in HA mode with multiple master nodes.


pebrc:
buildkite test this -f p=gke,E2E_TAGS: "es" -m s=8.1.3,s=8.13.2

pebrc:
buildkite test this -f p=gke,E2E_TAGS=es -m s=8.1.3,s=8.13.2

pkg/controller/elasticsearch/nodespec/readiness_probe.go
```diff
@@ -9,12 +9,30 @@ import (
...
 
-func NewReadinessProbe() *corev1.Probe {
+// as of 8.2.0 a simplified unauthenticated readiness port is available which takes cluster membership into account
+// see https://www.elastic.co/guide/en/elasticsearch/reference/current/advanced-configuration.html#readiness-tcp-port
+
+const (
+	ReadinessPortProbeScriptConfigKey = "readiness-port-script.sh"
+	// ReadinessPortProbeScript is the simplified readiness check for ES >= 8.2.0 which supports a dedicated TCP check
+	ReadinessPortProbeScript = `#!/usr/bin/env bash
+nc -z -v -w5 127.0.0.1 8080
```
thbkrkr, line 26: Nit naming suggestion: `ReadinessTCPProbeXYZ`and `OldReadinessHTTPProbeXYZ`

thbkrkr (approved):
LGTM

thbkrkr:
buildkite test this p=gke,s=8.1.3,t=TestVersionUpgradeToLatest8x

pkg/controller/elasticsearch/nodespec/readiness_probe.go
```diff
@@ -9,12 +9,30 @@ import (
...
+`
+)
+
+func NewReadinessProbe(v version.Version) *corev1.Probe {
+	scriptKey := ReadinessPortProbeScriptConfigKey
+	if v.LE(esv1.MinReadinessPortVersion) {
```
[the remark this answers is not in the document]
pebrc: I added more coverage in https://github.com/elastic/cloud-on-k8s/pull/7847/commits/2652e0ff461e848453668e42d39bf98d3a6a7bc1

pkg/controller/elasticsearch/nodespec/__snapshots__/podspec_test.snap
```diff
@@ -0,0 +1,1852 @@
...
+     }
+    ],
+    "image": "docker.elastic.co/elasticsearch/elasticsearch:7.2.0",
+    "lifecycle": {
+     "preStop": {
+      "exec": {
+       "command": [
+        "bash",
+        "-c",
+        "/mnt/elastic-internal/scripts/pre-stop-hook-script.sh"
```
pebrc, line 124: pre 8.2.0

pkg/controller/elasticsearch/nodespec/__snapshots__/podspec_test.snap
```diff
@@ -0,0 +1,1852 @@
...
+      "value": "/mnt/elastic-internal/pod-mounted-users/elastic-internal-probe"
+     },
+     {
+      "name": "PROBE_USERNAME",
+      "value": "elastic-internal-probe"
+     },
+     {
+      "name": "READINESS_PROBE_PROTOCOL",
+      "value": "https"
+     },
```
pebrc, line 1099: We could remove those

pkg/controller/elasticsearch/nodespec/__snapshots__/podspec_test.snap
```diff
@@ -0,0 +1,1852 @@
...
+      "name": "transport",
+      "protocol": "TCP"
+     }
+    ],
+    "readinessProbe": {
+     "exec": {
+      "command": [
+       "bash",
+       "-c",
+       "/mnt/elastic-internal/scripts/readiness-port-script.sh"
```
pebrc, line 1139: New script post 8.2.0

pebrc:
buildkite test this p=gke,s=8.1.3,t=TestVersionUpgradeToLatest8x

pebrc:
buildkite test this -f p=kind,s=8.1.3 -m t=TestVersionUpgradeSingleToLatest8x,t=TestVersionUpgradeTwoNodesToLatest8x,t=TestExternalESStackMonitoring,t=TestForceUpgradePendingPodsInOneStatefulSet,t=TestKillSingleNodeReusePV,t=TestPodTemplateValidation,t=TestRedClusterCanBeModifiedByDisablingPredicate,t=TestStackConfigPolicy,t=TestVolumeRetention

pebrc:
There is an edge case with single node clusters that I did not consider:
* the readiness.port reports an ES node as not ready if a shutdown record exists for the ES node
* an unready node is not listed in the (internal) k8s service we use and assumed not reacheable
* this means in turn that the operator is not able to call the `DELETE _nodes/$NODE_ID/shutdown` API to remove the shutdown record and the cluster remains unready forever

The problem does not present itself with multinode clusters where at least one node is available at all times in the same way but the edge case applies to those as well if for example all nodes are deleted at the same time or some other external factor.

pebrc:
buildkite test this -f p=kind,E2E_TAGS=es -m s=8.1.3,s=8.13.2

pebrc:
buildkite test this -f p=kind,E2E_TAGS=es -m s=8.1.3,s=8.13.2

pebrc:
I am thinking about the different options to address the problem:
1. consider this a bug in the Elasticsearch implementation of the readiness probe? (I leaning towards no, as the contract for the shutdown API is indeed for the orchestrator to clean up the shutdown records)
2. Allow unready Pods in the service (maybe only as of ES >= 8.2.0), this is what I have been testing in https://github.com/elastic/cloud-on-k8s/pull/7847/commits/23589474ba130ae6ade23a02abe70c7fbcc1d786 and following commits. While this addresses the issue at hand it also introduces extra errors when the operator tries to a node that is not ready and actually also not serving requests. 
3. Rewrite the Elasticsearch client logic to not go through the k8s service but instead to pick a Pod at random from a priority list which is listing all ready pods and then all running (but not ready ) pods. 
4. A variant of 3 that does still go throught the service but uses a different label that the operator puts onto the pods when they become ready to serve but not necessarily ready in the k8s sense (Njal's idea from 2019) I don't like that too much due to the extra API requests required for labeling the pods and prefer 3. 

@barkbay @thbkrkr given that you have reviewed this PR I would be curious about your thoughts.

barkbay:
Could we rely on the service unless there is no `ready` endpoint, in which case we would pick a random Pod? (mostly to be conservative and only use a new logic for some edge cases like single node clusters, all Pods deleted at a same time...)

pebrc:
buildkite test this -f p=kind,E2E_TAGS=es -m s=8.1.3,s=8.13.2

pebrc:
The test run from the comment above completed successful in 2h instead of the usual 4h, which is suspicious. I need to take a closer look if this is a bug or if the fact the we optmistically make connections to non-ready pods speeds up the tests so much.

pebrc:
buildkite test this -f p=gke,TESTS_MATCH=* -m s=8.1.3,s=8.13.2

pebrc:
buildkite test this -f p=gke,t=Test -m s=8.1.3,s=8.13.2

pkg/controller/elasticsearch/services/services.go
```diff
@@ -181,26 +183,59 @@ func getServiceByName(c k8s.Client, es esv1.Elasticsearch, serviceName string) (
...
-// otherwise this delegates to ExternalServiceURL.
-func ElasticsearchURL(es esv1.Elasticsearch, pods []corev1.Pod) string {
-	var schemeChange bool
-	for _, p := range pods {
-		scheme, exists := p.Labels[label.HTTPSchemeLabelName]
-		if exists && scheme != es.Spec.HTTP.Protocol() {
-			// scheme in existing pods does not match scheme in spec, user toggled HTTP(S)
-			schemeChange = true
+type urlProvider struct {
+	pods   func() []corev1.Pod
```
barkbay, line 187: Is there any reason not to return early in case of an error? 
```suggestion
	pods   func() ([]corev1.Pod, error)
```

pkg/controller/elasticsearch/driver/driver.go
```diff
@@ -184,22 +184,20 @@ func (d *defaultDriver) Reconcile(ctx context.Context) *reconciler.Results {
 		minVersion = &d.Version
 	}
 
-	isServiceReady, err := services.IsServiceReady(d.Client, *internalService)
```
barkbay, line 187: Just realized that we should maybe do a similar change in the [autoscaling client](https://github.com/elastic/cloud-on-k8s/blob/634ccd111feddf70f3bcaf21db9ecce990bc9d94/pkg/controller/autoscaling/elasticsearch/driver.go#L101) and remove `services.IsServiceReady`?

pkg/controller/elasticsearch/client/url.go
```diff
@@ -0,0 +1,40 @@
...
+// or more contributor license agreements. Licensed under the Elastic License 2.0;
+// you may not use this file except in compliance with the Elastic License 2.0.
+
+package client
+
+type URLProvider interface {
+	// PodURL is a url for a random pod (falls back to ServiceURL).
+	PodURL() string
+	// ServiceURL is the url for the Kubernetes service related to the Pod URLs provided.
+	ServiceURL() string
```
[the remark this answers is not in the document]
pebrc: Good idea.

pkg/controller/elasticsearch/services/services.go
```diff
@@ -181,26 +183,59 @@ func getServiceByName(c k8s.Client, es esv1.Elasticsearch, serviceName string) (
...
+	return len(k8s.RunningPods(u.pods())) > 0
+}
+
+func NewElasticsearchURLProvider(ctx context.Context, es esv1.Elasticsearch, client k8s.Client) client.URLProvider {
+	return &urlProvider{
+		pods: func() []corev1.Pod {
+			log := ulog.FromContext(ctx)
+			pods, err := k8s.PodsMatchingLabels(client, es.Namespace, label.NewLabelSelectorForElasticsearch(es))
+			if err != nil {
+				log.Error(err, "while fetching pods from cache in URL provider")
```
pebrc, line 228: @barkbay with https://github.com/elastic/cloud-on-k8s/pull/7847/files/1649df75c87bdd0033acd79e2530661fc3c48e7e#r1623879698 you mean return here in case of an error?

barkbay: Yes

pkg/utils/k8s/k8sutils.go
```diff
@@ -90,6 +91,7 @@ func TerminatingPods(pods []corev1.Pod) []corev1.Pod {
 	return terminating
 }
 
+// RunningPods filters pods for Pods are in running (and not terminating).
```
barkbay, line 94: nit: feels like something is missing
```suggestion
// RunningPods filters pods for Pods that are in the running phase (and not terminating).
```

pkg/controller/elasticsearch/client/url.go
```diff
@@ -0,0 +1,46 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License 2.0;
+// you may not use this file except in compliance with the Elastic License 2.0.
+
+package client
+
+type URLProvider interface {
+	// URL return a URL to route traffic to (can fall back to a k8s service URL).
```
barkbay, line 8:
```suggestion
	// URL returns a URL to route traffic to (can fall back to a k8s service URL).
```

pkg/controller/elasticsearch/client/url.go
```diff
@@ -0,0 +1,46 @@
...
+	URL() (string, error)
+	// Equals returns true if the other URLProvider is equal to the one in the receiver.
+	Equals(other URLProvider) bool
+	// HasEndpoints return true if the provider has currently any endpoints/URLs to return.
```
barkbay, line 12:
```suggestion
	// HasEndpoints returns true if the provider has currently any endpoints/URLs to return.
```

barkbay (approved):
LGTM

pebrc:
I tried to compare memory usage against a baseline from the last nightly run:
Screenshot 2024-06-04 at 12 07 55

with the one of the longer runs triggered from this PR here https://github.com/elastic/cloud-on-k8s/pull/7847#issuecomment-2138311920
Screenshot 2024-06-04 at 12 08 29

I did not spot any significant change, but the comparison leaves much to be desired.

PR #7859, opened 2024-05-29
Update supported-versions.asciidoc

ocuil:
Following the Support Matrix (https://www.elastic.co/support/matrix#matrix_kubernetes) ECK is supported on VMware Tanzu ... 

I think we need to review the both documents (Support Matrix and this one)

### DIFF
--- a/docs/supported-versions.asciidoc
+++ b/docs/supported-versions.asciidoc
@@ -2,6 +2,7 @@
 * OpenShift 4.11-4.14
 * Google Kubernetes Engine (GKE), Azure Kubernetes Service (AKS), and Amazon Elastic Kubernetes Service (EKS)
 * Helm: 3.2.0+
+* VMware Tanzu Kubernetes Grid Versions: 	1.4, 1.5.4
 * Elasticsearch, Kibana, APM Server: 6.8+, 7.1+, 8+
 * Enterprise Search: 7.7+, 8+
 * Beats: 7.0+, 8+


docs/supported-versions.asciidoc
```diff
@@ -2,6 +2,7 @@
 * OpenShift 4.11-4.14
 * Google Kubernetes Engine (GKE), Azure Kubernetes Service (AKS), and Amazon Elastic Kubernetes Service (EKS)
 * Helm: 3.2.0+
+* VMware Tanzu Kubernetes Grid Versions: 	1.4, 1.5.4
```
pebrc, line 5: I am not sure we want to highlight here that we support a very old version of Tanzu K8s that is no longer supported by VMWare and that is based on a Kubernetes version that is no longer supported (1.22). If anything we should strike Tanzu from the support matrix.

ocuil: Morning @pebrc thanks for sharing !!! you are right, as I'm not sure whats of both it's the most important (I think Support Matrix becase our Support is based on it ...) I don't know what and were we have update it, I just want to share because some coworkers and I use this documentation in from our customers and I feel it's not so clear.

thbkrkr:
The explicit mention of 'VMware Tanzu' will be removed from the [matrix](https://www.elastic.co/fr/support/matrix#matrix_kubernetes) page, hence there is no need to update the README or the docs.

Using ECK on Tanzu is still possible, as long as it is certified as compliant by the CNCF, as indicated at the bottom of the matrix page:

> We currently support ECK on Kubernetes installations that are conformant according to the list provided by the [Cloud Native Computing Foundation](https://www.cncf.io/certification/software-conformance/). We only support Kubernetes versions installed via installers as listed [here](https://docs.google.com/spreadsheets/d/1uF9BoDzzisHSQemXHIKegMhuythuq_GL3N1mlUUK2h0/edit#gid=0). Kubernetes distributions listed [here](https://docs.google.com/spreadsheets/d/1uF9BoDzzisHSQemXHIKegMhuythuq_GL3N1mlUUK2h0/edit#gid=0) are only supported on a discretionary basis.

PR #7874, opened 2024-06-07
Allow unassigned nodes by REPLICA_ADDED during shutdown

mrkm4ntr:
As talked in https://github.com/elastic/cloud-on-k8s/issues/3867, current implementation is conservative.
We are using `index.routing.allocation.total_shards_per_node` and sometimes there are unassigned shards by this after increasing the number of replicas. We are changing the number of nodes and the number of replicas automatically, so unassigned shards that have unassigned reason `REPLICA_ADDED` blocks other nodeSet to be scale in. The reason `REPLICA_ADDED` is for the shards that have never been assigned so we can ignore these during shutdown.

### DIFF
--- a/pkg/controller/elasticsearch/migration/migrate_data.go
+++ b/pkg/controller/elasticsearch/migration/migrate_data.go
@@ -74,8 +74,8 @@ func nodeMayHaveShard(ctx context.Context, es esv1.Elasticsearch, shardLister es
 		if shard.NodeName == podName {
 			return true, nil
 		}
-		// shard node undefined (likely unassigned)
-		if shard.NodeName == "" {
+		// shard node undefined (likely unassigned by restarting the node)
+		if shard.NodeName == "" && shard.UnassignedReason != "REPLICA_ADDED" {
 			ulog.FromContext(ctx).Info("Found orphan shard, preventing data migration",
 				"namespace", es.Namespace, "es_name", es.Name,
 				"index", shard.Index, "shard", shard.Shard, "shard_state", shard.State)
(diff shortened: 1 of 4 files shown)


pkg/controller/elasticsearch/migration/migrate_data.go
```diff
@@ -74,8 +74,8 @@ func nodeMayHaveShard(ctx context.Context, es esv1.Elasticsearch, shardLister es
 		if shard.NodeName == podName {
 			return true, nil
 		}
-		// shard node undefined (likely unassigned)
-		if shard.NodeName == "" {
+		// shard node undefined (likely unassigned by restarting the node)
+		if shard.NodeName == "" && shard.UnassignedReason != "REPLICA_ADDED" {
```
barkbay, line 78: This code path is not supposed to be used [since Elasticsearch `7.15.2`](https://github.com/elastic/cloud-on-k8s/blob/4f89b715f7dfe270b56227953986a98e4d64e090/pkg/controller/elasticsearch/shutdown/interface.go#L14). It has been superseded by the [shutdown API](https://www.elastic.co/guide/en/elasticsearch/reference/7.17/put-shutdown.html). I would suggest to upgrade to a more recent version of Elasticsearch that supports this API. If you have some issues during upgrades feel free to open a topic in https://discuss.elastic.co/c/orchestration/eck/79

barkbay: If we merge this change then you will have to update ECK, but since you are using ES 8.x this code path is not going to be used 🙃  
Any reason for not upgrading ECK? 1.9.1 is pretty old (released at the end of 2021)

mrkm4ntr: Ah, yes. We will upgrade ECK.
We applied this patch to 1.9.1 by ourselves for hot fix and the patched version is running in our cluster. I'd like to avoid patch this for v2 by ourselves but now I know this patch is not needed for us. 
This PR might help someone who is using older Elasticsearch and newer ECK but there is no such person. Please feel free to close this.

Issue #7790, opened 2024-04-30
ECK TestFleet* is failing


barkbay:
https://buildkite.com/elastic/cloud-on-k8s-operator-nightly/builds/530:

I had a closer look at `TestFleetMode/Fleet_in_same_namespace_as_Agent`:

* `2024-04-29T23:03:21.220427957Z` - `TestFleetMode/Fleet_in_same_namespace_as_Agent/ES_data_should_pass_validations` starts
* `2024-04-29T23:18:21.222389355Z` - Failure with `{Status:404 Error:{CausedBy:{Reason: Type:} Reason:no such index [logs-elastic_agent-default]`

### Agents

The 3 Agents have the same exact same final log lines:

```json
{
    "log.level": "error",
    "@timestamp": "2024-04-29T23:10:26.530Z",
    "log.origin": {
        "file.name": "fleet/fleet_gateway.go",
        "file.line": 206
    },
    "message": "Could not communicate with fleet-server Checking API will retry, error: fail to checkin to fleet-server: Post \"https://test-agent-fleet-fs-w6xh-agent-http.e2e-fo9kn-mercury.svc:8220/api/fleet/agents/e94192f6-77db-431e-929e-98a11f267d6b/checkin?\": dial tcp 10.56.235.142:8220: connect: connection refused",
    "ecs.version": "1.6.0"
}
```

```json
{
    "log.level": "error",
    "@timestamp": "2024-04-29T23:15:23.323Z",
    "log.origin": {
        "file.name": "fleet/fleet_gateway.go",
        "file.line": 206
    },
    "message": "Could not communicate with fleet-server Checking API will retry, error: fail to checkin to fleet-server: Post \"https://test-agent-fleet-fs-w6xh-agent-http.e2e-fo9kn-mercury.svc:8220/api/fleet/agents/bc83885c-c297-44a9-95db-b5471fba15e8/checkin?\": dial tcp 10.56.235.142:8220: connect: connection refused",
    "ecs.version": "1.6.0"
}
```

```json
{
    "log.level": "error",
    "@timestamp": "2024-04-29T23:16:30.150Z",
    "log.origin": {
        "file.name": "fleet/fleet_gateway.go",
        "file.line": 206
    },
[...]

PR #7789, opened 2024-05-09
Increase default memory for agent

kvalliyurnatt:
In our e2e tests we noticed test failures related to agent and fleet. Debugging that we found the agent pods were using about 366Mi of memory vs a limit of 350Mi, this was causing the processes to be OOM killed. Increasing the default memory allocation for agent while we figure out the increase in memory usage with the latest release.

PR #7884, opened 2024-06-10
[E2E] Increase Agent memory on OpenShift

barkbay:
Similar to https://github.com/elastic/cloud-on-k8s/pull/7789, however Agent's memory is only increased on OpenShift.

The rational is that, since there are more components running on a OpenShift cluster, it requires more memory than other K8s distributions. Therefore OpenShift is considered as a special case.

* Relates to https://github.com/elastic/cloud-on-k8s/issues/7790
* Should be revert once https://github.com/elastic/elastic-agent/issues/4729 is fixed

### DIFF
--- a/test/e2e/test/agent/builder.go
+++ b/test/e2e/test/agent/builder.go
@@ -10,9 +10,11 @@ import (
 
 	ghodssyaml "github.com/ghodss/yaml"
 	"github.com/stretchr/testify/require"
+
 	appsv1 "k8s.io/api/apps/v1"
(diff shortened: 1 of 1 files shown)


barkbay:
Converted as a draft while I'm doing some tests...

barkbay:
I ran the other failing tests (`TestMultipleOutputConfig` and `TestSystemIntegrationConfig`) locally, and there are now also passing.

pebrc (approved):

PR #7892, opened 2024-06-11
fix: cleanup pre-stop-hook-script.sh and wait for NODE_ID

BobVanB:
  Maybe, probably have somewhere.
  No, take it or leave it, we have a workaround.
  Yes its against main.

# The main problem:

When upgrading a image with some other plugin, the cluster will restart each pod and try to remove it.

This piece of code can be empty:

```shell
NODE_ID=$(grep "$POD_NAME" "$resp_body" | cut -f 1 -d ' ')
```

## Result

- There is no NODE_ID and the request is broken between `_nodes` and `shutdown`
   ```log
   {"@timestamp": "2024-06-11T09:37:05+00:00", "message": "400 http://cps2-logging-es-internal-http.ams-system-logging.svc:9200/_nodes//shutdown", "ecs.version": "1.2.0", "event.dataset": "elasticsearch.pre-stop-hook"}
   ```
- The pod that is restarting will crash each time with `error_exit "failed to call node shutdown API"` and a shutdown is never called. Thus resulting in recreating the same pod again and starting allover from the top.

# What i still want to know

When the node is terminated and the pre-stop-hook-script.sh is called, is it possible that the node is already removed from the cluster?

This piece of code returns the list of nodes and i wonder if the pod is terminated the node is actually already not present in this list from active nodes. Still no basis for this claim, but i have not confirmed if the NODE_ID is empty because the other nodes in the cluster don't see the node that is terminated.

```shell
request -X GET "${ES_URL}/_cat/nodes?full_id=true&h=id,name"
```

# What has been done

- After some debugging and trying to understand the code, i ended up cleaning it up with shellcheck.
- TODO: build a retry loop to get the NODE_ID


pkg/controller/elasticsearch/nodespec/lifecycle_hook.go
```diff
@@ -146,62 +146,68 @@ function supports_node_shutdown() {
...
+  error_exit "failed to retrieve nodes"
+fi
+
+if ! NODE_ID="$(grep -oP "\K\w+(?= ${POD_NAME})" "${resp_body}")"
+then
+  error_exit "failed to extract node id"
 fi
 
-NODE_ID=$(grep "$POD_NAME" "$resp_body" | cut -f 1 -d ' ')
+if ! request -X GET "${ES_URL}/_nodes/${NODE_ID}/shutdown" "${BASIC_AUTH[@]}"
```
thbkrkr, line 182: Let's keep the comment.
```suggestion
# check if there is an ongoing shutdown request
if ! request -X GET "${ES_URL}/_nodes/${NODE_ID}/shutdown" "${BASIC_AUTH[@]}"
```

pkg/controller/elasticsearch/nodespec/lifecycle_hook.go
```diff
@@ -146,62 +146,68 @@ function supports_node_shutdown() {
...
 fi
 
-ES_URL={{.ServiceURL}}
+ES_URL="{{.ServiceURL}}"
 
 log "retrieving node ID"
-retry 10 request -X GET "$ES_URL/_cat/nodes?full_id=true&h=id,name" $BASIC_AUTH
-if [ "$?" -ne 0 ]; then
-	error_exit "failed to retrieve node ID"
+if ! retry 8 request -X GET "${ES_URL}/_cat/nodes?full_id=true&h=id,name" "${BASIC_AUTH[@]}"
```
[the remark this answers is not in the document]
BobVanB: This is explained with the title `Why is terminationGracePeriodSeconds way less then possible script run time?`
retry 9 and retry 10 will be after the `terminationGracePeriodSeconds` and the scripts will be terminated and no other functions like `error_exit` can be called.

Better to stop cleanly, then to finished suddenly.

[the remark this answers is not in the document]
thbkrkr: Makes sense, I missed that, sorry.

thbkrkr:
buildkite test this -p gke,E2E_TAGS=es -v s=7.17.8,s=8.14.1,s=8.15.0-SNAPSHOT

thbkrkr:

thbkrkr:
This breaks two e2e tests:
```
TestMutationResizeMemoryDown/Stopping_to_watch_for_correct_node_shutdown_API_usage
TestMutationResizeMemoryUp/Stopping_to_watch_for_correct_node_shutdown_API_usage
```

I don't know exactly what's going on yet.
The failure is because during the mutation, `/_nodes/shutdown` returned more than one entry:

```
[
	{Q3SwszElnHxaJg RESTART pre-stop hook 1719495839993 COMPLETE {COMPLETE 0 no shard relocation is necessary for a node restart} {COMPLETE} {COMPLETE}}
	{eUcnfdK-Q3SwszElnHxaJg RESTART 70382 1719495839494 COMPLETE {COMPLETE 0 no shard relocation is necessary for a node restart} {COMPLETE} {COMPLETE}}
]
```

thbkrkr:
The pre-stop hook incorrectly extracted the node id, which created 2 shutdown records with different ids (Q3SwszElnHxaJg and eUcnfdK-Q3SwszElnHxaJg).

pkg/controller/elasticsearch/nodespec/lifecycle_hook.go
```diff
@@ -146,62 +146,69 @@ function supports_node_shutdown() {
...
+then
+  error_exit "failed to retrieve nodes"
 fi
 
-NODE_ID=$(grep "$POD_NAME" "$resp_body" | cut -f 1 -d ' ')
+if ! NODE_ID="$(grep -oP "\K\w+(?= ${POD_NAME})" "${resp_body}")"
```
thbkrkr, line 177: This doesn't work when the node ID has `-`.

The simple thing I can think of to do the extraction is: `grep "${POD_NAME}" "${resp_body}" | awk '{print $1}'`.

I tested with this script:
```
#!/usr/bin/env bash

set -uo pipefail


req() {
  echo 'eb0GwvNWTkCuQngXptksnA pod-0
ACJTNKxESkes23TYiX6eZA pod-1
t1yM7brtQB256e-BVmM-Vw pod-2'
}

resp_body=tmpfile
req > $resp_body

test() {
  echo "-- test with POD_NAME=$POD_NAME"

  if ! NODE_ID="$(grep -oP "\K\w+(?= ${POD_NAME})" "${resp_body}")"; then
    echo no id
  else 
    echo id: $NODE_ID
  fi

  extract_node_id() {
    grep "${POD_NAME}" "${resp_body}" | awk '{print $1}'
  }

  if ! NODE_ID="$(extract_node_id  ${POD_NAME})"; then
    echo no id
  else 
    echo id: $NODE_ID
  fi

  echo
}

POD_NAME=pod-xyz
test

POD_NAME=pod-1
test

POD_NAME=pod-2
test
```

thbkrkr: Output:
```
> ./test.sh
-- test with POD_NAME=pod-xyz
no id 🟢
no id 🟢

-- test with POD_NAME=pod-1
id: ACJTNKxESkes23TYiX6eZA 🟢
id: ACJTNKxESkes23TYiX6eZA 🟢

-- test with POD_NAME=pod-2
id: Vw 🔴
id: t1yM7brtQB256e-BVmM-Vw 🟢
```

BobVanB: Easy fix, \w is one word, use a different regex to match all before '<space><name>'

```shell
grep -oP ".*?(?=\ pod-1321)"
```

BobVanB:
```suggestion
if ! NODE_ID="$(grep -oP ".*?(?=\ ${POD_NAME})" "${resp_body}")"
```

thbkrkr: > Are there any more corner cases?

Sorry but I don't know.

Do you see an issue with using `grep | awk`? I would like to avoid using `grep --perl-regexp` if possible because regexes are often tricky.

BobVanB: Nope, i don't see any issues, by habit i tend to make only one command and if possible ignore the pipe into other commands.
This leads to 2 commands that needs to function and checked.

Personal preference i think.

thbkrkr:

pkg/controller/elasticsearch/nodespec/lifecycle_hook.go
thbkrkr: Traditionally, I like regular expressions, but when it comes to filtering and at the same time extracting with a grouping, it starts to get a little trickier. Here, we prefer to prioritize readability for maintainability in case an operator is required to read this script.
So, yes please return to `grep | cut`.

BobVanB: Yes i will push it tomorrow.

BobVanB:

thbkrkr:

thbkrkr (approved):

thbkrkr:
Thank you @BobVanB!

pebrc (approved):

pkg/controller/elasticsearch/nodespec/lifecycle_hook.go
```diff
@@ -146,62 +146,69 @@ function supports_node_shutdown() {
...
 fi
 
-ES_URL={{.ServiceURL}}
+ES_URL="{{.ServiceURL}}"
 
 log "retrieving node ID"
-retry 10 request -X GET "$ES_URL/_cat/nodes?full_id=true&h=id,name" $BASIC_AUTH
-if [ "$?" -ne 0 ]; then
-	error_exit "failed to retrieve node ID"
+if ! retry 8 request -X GET "${ES_URL}/_cat/nodes?full_id=true&h=id,name" "${BASIC_AUTH[@]}"
```
barkbay, line 172: `8`  feels a bit like a magic number, should it be a variable with a comment to explain why we selected that value and why it can be greater?

thbkrkr: Yes, something like that?

```diff
diff --git a/pkg/controller/elasticsearch/nodespec/lifecycle_hook.go b/pkg/controller/elasticsearch/nodespec/lifecycle_hook.go
index cf91a1433..590dd6adb 100644
--- a/pkg/controller/elasticsearch/nodespec/lifecycle_hook.go
+++ b/pkg/controller/elasticsearch/nodespec/lifecycle_hook.go
@@ -83,6 +83,9 @@ function request() {
   return 0
 }

+# number of retries to try not to last more than default terminateGracePeriodSeconds (0 + 1 + 2 + 4 + 8 + 16 + 32 + 64 < 180s)
+retries_count=8
+
 function retry() {
   local retries=$1
   shift
@@ -169,7 +172,7 @@ fi
 ES_URL="{{.ServiceURL}}"

 log "retrieving node ID"
-if ! retry 8 request -X GET "${ES_URL}/_cat/nodes?full_id=true&h=id,name" "${BASIC_AUTH[@]}"
+if ! retry "$retries_count" request -X GET "${ES_URL}/_cat/nodes?full_id=true&h=id,name" "${BASIC_AUTH[@]}"
 then
   error_exit "failed to retrieve nodes"
 fi
@@ -191,7 +194,7 @@ if grep -q -v '"nodes":\[\]' "$resp_body"; then
 fi

 log "initiating node shutdown"
-if ! retry 8 request -X PUT "${ES_URL}/_nodes/${NODE_ID}/shutdown" "${BASIC_AUTH[@]}" -H 'Content-Type: application/json' -d"
+if ! retry "$retries_count" request -X PUT "${ES_URL}/_nodes/${NODE_ID}/shutdown" "${BASIC_AUTH[@]}" -H 'Content-Type: application/json' -d"
```

"try" because there are two `retry $n request` in the script. If we consumes the budget in the first, but reach the second, we can still last longer than `terminateGracePeriodSeconds`.

BobVanB: In my opinion i would not make this change and make this a new pull request after this script is fixed and use one feature request at a time. I tried to not change the script functionally and only made changes for problematic scripting errors and stopped making changes after the script finally worked without errors.

In that regard it can be said that it also should return to 10 and break each change apart:

1. Fix: `error_exit` if NODE_ID is empty.
2. Fix: Style fixes and errors found by `shellcheck`
3. Fix: More sensible usage of the retry counter.

I also had a few that i didn't do:

1. Use a library for all for scripts and reuse the same code.
2. Test the code using `shellspec` and test the library every time.
3. Add more logging lines to debug.
4. Attach it to log4j and differentiate between debug and info lines.

And got more, but for now i only want NODE_ID fixed.

pkg/controller/elasticsearch/nodespec/lifecycle_hook.go
```diff
@@ -113,25 +113,25 @@ function log() {
...
+  log "delaying termination for ${remaining} seconds"
   sleep $remaining
-  exit ${1-0}
+  exit "${1-0}"
 }
 
 function supports_node_shutdown() {
-  local version="$1"
-  version=${version#[vV]}
+  local version=$1
```
BobVanB, line 133:
```suggestion
  local version="$1"
```

pkg/controller/elasticsearch/nodespec/lifecycle_hook.go
thbkrkr: I think it's better to change to `8` as you pointed out and fine to do it in this PR. I push [3285537] with the above suggestion. I'm going to rerun the tests as a matter of principle and it will be good to be merged.

thbkrkr:

thbkrkr:
buildkite test this

thbkrkr:
Thank you very much for the contribution and for your patience @BobVanB.

PR #7893, opened 2024-06-11
[E2E] Fix Monitoring Agent on OCP and add journald integration

barkbay:
This PR fixes a permission issue for the E2E monitoring Agents on OpenShift (the ones which are collecting the logs, not the ones being tested)

It also adds the `journald` integration to capture OOM events:

image

An unfortunate side effect is that I need to increase the Agent memory from `768Mi` to `1Gi`, not sure if it is related to OpenShift or the the `journald` integration though (or both).

### DIFF
--- a/config/e2e/monitoring.yaml
+++ b/config/e2e/monitoring.yaml
@@ -17,6 +17,9 @@ spec:
           - name: agent
             securityContext:
               runAsUser: 0
+  {{ if .OcpCluster }}
+              privileged: true
+  {{ end }}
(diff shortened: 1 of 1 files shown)


pebrc (approved):

barkbay:
buildkite test this -f p=gke,s=8.14.0,t=TestSmoke

barkbay:
buildkite test this -f p=gke,s=8.14.0 -m t=TestSmoke

PR #7896, opened 2024-06-12
Handle Serverless version in association versions check

barkbay:
This PR attempts to detect when a referenced resource is an Elasticsearch project, and skip version check if it's the case.

### DIFF
--- a/pkg/controller/association/reconciler.go
+++ b/pkg/controller/association/reconciler.go
@@ -77,7 +77,8 @@ type AssociationInfo struct { //nolint:revive
 	// base is used to recognize annotations eligible for removal when association is removed.
 	AssociationConfAnnotationNameBase string
 	// ReferencedResourceVersion returns the currently running version of the referenced resource.
-	// It may return an empty string if the version is unknown.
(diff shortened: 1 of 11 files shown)


barkbay:
buildkite test this -f p=gke,s=8.14.0 -m t=TestVersionUpgradeOrdering*

pkg/controller/association/reconciler.go
```diff
@@ -78,7 +78,7 @@ type AssociationInfo struct { //nolint:revive
 	AssociationConfAnnotationNameBase string
 	// ReferencedResourceVersion returns the currently running version of the referenced resource.
 	// It may return an empty string if the version is unknown.
-	ReferencedResourceVersion func(c k8s.Client, association commonv1.Association) (string, error)
+	ReferencedResourceVersion func(c k8s.Client, association commonv1.Association) (string, bool, error)
```
pebrc, line 81: Document the boolean return value?

pebrc (commented):
LGTM in general I wonder if we should support the serverless flavour for Kibana as well.

pebrc (approved):

PR #7921, opened 2024-06-27
Minor fixes in Logstash docs

thbkrkr:
This fixes the name of the entry in the Secret to use to define pipelines configuration: `s/logstash.yaml/pipelines.yaml/`.

Thanks @jeanfabrice for the report.


karenzone (approved):
Thanks for these corrections, @thbkrkr.  LGTM

Issue #6954, opened 2023-06-23
Data Transport Cert Secret Size Overrun With Big Scale Out


rtluckie:
## Bug Report

**What did you do?**

- Attempted to scale the data replicas to 250.

**What did you expect to see?**

- successful scale up

**What did you see instead? Under which circumstances?**

- It appears that the ECK operator will overflow the max k8s secret size (1MB) for the transport certs if you scale the data nodes to >250.
- The operator gets stuck in a scale up loop while it tries to reconcile the cert secret. Even after scaling down the operator does not seem to recover.

```
"Secret "elasticsearch-XXX-es-data-es-transport-certs" is invalid: data: Too long: must have at most 1048576 bytes" error
```

**Failed remediations**

- issue transport cert as documents [here](https://www.elastic.co/guide/en/cloud-on-k8s/master/k8s-transport-settings.html#k8s-transport-third-party-tools)
- issue wildcard transport certs as documents [here](https://www.elastic.co/guide/en/cloud-on-k8s/master/k8s-transport-settings.html#k8s-transport-third-party-tools)

**Environment**

- ECK version: 2.8.0

- Kubernetes information:

  - Cloud: GKE v1.26.3-gke.1000

- kubectl version: v1.27.2

- Resource definition:

```
apiVersion: elasticsearch.k8s.elastic.co/v1
kind: Elasticsearch
metadata:
  name: elasticsearch-myapp
spec:
  version: 8.6.1
  http:
    tls:
      selfSignedCertificate:
        disabled: true
  nodeSets:
  - config:
      action:
        auto_create_index: false
      node.roles:
      - master
    count: 3
    name: election
    podTemplate:
      metadata:
        annotations:
          linkerd.io/inject: enabled
        labels:
          ec.ai/component: elasticsearch
          ec.ai/component_group: myapp-service
      spec:
        affinity:
          nodeAffinity:
            requiredDuringSchedulingIgnoredDuringExecution:
              nodeSelectorTerms:
              - matchExpressions:
                - key: cloud.google.com/gke-spot
                  operator: DoesNotExist
          podAntiAffinity:
            preferredDuringSchedulingIgnoredDuringExecution:
            - podAffinityTerm:
                labelSelector:
                  matchExpressions:
                  - key: elasticsearch.k8s.elastic.co/cluster-name
                    operator: In
                    values:
                    - elasticsearch-myapp
                topologyKey: topology.kubernetes.io/zone
              weight: 100
[...]

Issue #1841, opened 2019-10-02
Review Certificate Reconciliation Logic


charith-elastic:
As noted in #1832, the `certificates.Reconcile` method has a large footprint in terms of memory usage. This is probably due to parsing of certificates over and over. We should review the logic to see if we can avoid some of this work.

Alloc Space:
![image](https://user-images.githubusercontent.com/52399125/66025146-a5b51180-e4ed-11e9-8e8c-57a23efd2bf9.png)

anyasabo:
Some specific examples called out in this PR: https://github.com/elastic/cloud-on-k8s/pull/2541

charith-elastic:
In a single iteration of the Elasticsearch reconciliation loop, the certificates are parsed and encoded multiple times.

- HTTP CA
    - Check that parsing succeeds 
    - Check expiry time to create a tentative requeue request
    - Return parsed CA certificates in case the HTTP certificate needs to be regenerated
- HTTP certificate
    - If user-provided, parse to ensure that the secret is valid
    - If self-signed, parse to check whether the certificate should be updated 
    - Return either the user-provided or self-signed certificate as a secret which then gets parsed again to check expiry time and create a new tentative requeue request
    - Encode the certificate again to create the public secret
- Transport CA
    - Check that parsing succeeds 
    - Check expiry time to create a tentative requeue request
    - Return parsed CA certificates to be used to generated the xpack configuration
    - Encode the certificate again to create the public secret
- Transport certificates
    - For each node in the Elasticsearch cluster:
        - Parse the transport certificate entry to ensure it is valid (and regenerate if necessary)
        - Parse the certificate again to check for expiry and create tentative requeue request
- HTTP certificate chain
    - Parse the certificate chain to pass to the Elasticsearch client



A synthetic benchmark (using the fake client) of `certificates.Reconcile` produced the following:

```
goos: linux
goarch: amd64
pkg: github.com/elastic/cloud-on-k8s/pkg/controller/elasticsearch/certificates
BenchmarkReconcile          500  11719931 ns/op 1246609 B/op      19199 allocs/op
BenchmarkReconcile-2        500  13337516 ns/op 1268568 B/op      19227 allocs/op
BenchmarkReconcile-4        500  10793754 ns/op 1275032 B/op      19219 allocs/op
BenchmarkReconcile-8        500  11304709 ns/op 1279162 B/op      19223 allocs/op
```

CPU and memory profiles are attached. 

[...]

PR #7925, opened 2024-07-01
Add option to disable self-signed transport certs

pebrc:
Related to https://github.com/elastic/cloud-on-k8s/issues/6954

It offers users a workaround for the problem with too many certificates in the transport certificate secret. They can configure external transport cert provisioning and disable self-signed transport certificates. When using a solution eg. like cert-manager's csi-driver as [documented here ](https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-transport-settings.html#k8s-transport-third-party-tools) this should allow for larger node sets of more than 250 nodes. 

The large cluster scenario is certainly an an edge case but on smaller clusters the disabling of certificate provisioning might still be attractive [reducing the amount of work the operator has to do in this area.](https://github.com/elastic/cloud-on-k8s/issues/1841)

Note the new option to disable the self-signed transport certificates below:

```yaml
kind: Elasticsearch
metadata:
  name: es
spec:
  version: 8.6.2 
  transport:
    tls:
      certificateAuthorities:
        configMapName: trust 
      selfSignedCertificates:
        disabled: true # <<<< new option
  nodeSets:
  - name: mixed
    count: 3
    config:
      xpack.security.transport.ssl.key: /usr/share/elasticsearch/config/cert-manager-certs/tls.key
      xpack.security.transport.ssl.certificate: /usr/share/elasticsearch/config/cert-manager-certs/tls.crt
      node.store.allow_mmap: false
    podTemplate:
      spec:
        containers:
        - name: elasticsearch
          env:
          - name: PRE_STOP_ADDITIONAL_WAIT_SECONDS
            value: "5"
          volumeMounts: 
          - name: transport-certs
            mountPath: /usr/share/elasticsearch/config/cert-manager-certs
        volumes: 
        - name: transport-certs
          csi: 
            driver: csi.cert-manager.io
            readOnly: true
            volumeAttributes: 
              csi.cert-manager.io/issuer-name: ca-cluster-issuer
              csi.cert-manager.io/issuer-kind: ClusterIssuer
              csi.cert-manager.io/dns-names: "${POD_NAME}.${POD_NAMESPACE}.svc.cluster.local"
```

The option does not remove existing certificates from the secret so that the cluster keeps working during the transition if this option is turned on on an existing cluster. 

I also opted to remove the symlinking of certificates into the `emptyDir` config volume. I tried to figure out why we did this in the first place and am not sure. The reason I could think of was that we wanted to have static and predictable certificate and key file names across all nodes (`transport.tls.crt` and `transport.tls.key`) But we can just use the `POD_NAME` environment variable to link directly into the mounted certificate secret volume. 

The reason to change this behaviour now is again to support the transition between externally provisioned certs and self-signed certs provisioned by ECK: if a user flips the switch to disable and then re-enable the self-signed certs, but does this accidentally without also configuring the config settings for the transport layer there is an edge case where an Elasticsearch pod will crashloop and cannot recover if we use symlinking: 
1. disable self-signed transport certs
2. scale the cluster up by one or more nodes
3. new nodes won't come up because certs are missing (user error) 
4. user tries to recover by re-enabling self-signed certs
5. ES keeps bootlooping on the new nodes because the symlink is missing 

By removing the symlinking the node can recover as soon as the certificates appear in the filesystem.


pebrc:

pebrc:
I just realised I forgot to update the docs. Will push another commit.

pkg/controller/elasticsearch/certificates/transport/reconcile.go
```diff
@@ -89,23 +89,62 @@ func reconcileNodeSetTransportCertificatesSecrets(
...
 
 	secret, err := ensureTransportCertificatesSecretExists(ctx, c, es, ssetName)
 	if err != nil {
-		return err
+		return results.WithError(err)
 	}
 	// defensive copy of the current secret so we can check whether we need to update later on
 	currentTransportCertificatesSecret := secret.DeepCopy()
+	if es.Spec.Transport.TLS.SelfSignedEnabled() {
+		if results.WithResults(reconcilePodTransportCertificates(ctx, es, ca, secret, pods, rotationParams)); results.HasError() {
```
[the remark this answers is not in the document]
pebrc: I think this is a good idea. If I understand correctly we would: 
1. Add the annotation when the transport certs are disabled: this forces are restart of the pods
2. once we observe the annotation present a given pod we can remove its transport cert from the secret

In practice users will probably combine turning off the transports certs with other spec changes that configure an alternative source for the certs but we need the explicit marker. 

Would you keep the transport CA around once all certs have been removed?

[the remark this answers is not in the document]
barkbay: > 1. Add the annotation when the transport certs are disabled: this forces are restart of the pods
> 2. once we observe the annotation present on a given pod we can remove its transport cert from the secret

Yes

> Would you keep the transport CA around once all certs have been removed?

I think it is fine to remove it in that case, only the new CA from `transport.tls.certificateAuthorities.configMapName` is useful once all the Pods are running with the externally provided certs.

pkg/controller/elasticsearch/certificates/transport/reconcile.go
```diff
@@ -89,20 +89,20 @@ func reconcileNodeSetTransportCertificatesSecrets(
 	es esv1.Elasticsearch,
 	ssetName string,
 	rotationParams certificates.RotationParams,
-) error {
+) *reconciler.Results {
```
pebrc, line 92: This "bugfix" leads to the clusters constantly being in state `ApplyingChanges` because we always have a requeue 🤔

pkg/controller/elasticsearch/certificates/transport/reconcile_test.go
```diff
@@ -7,22 +7,21 @@ package transport
...
 	"bytes"
 	"context"
-	"reflect"
 	"testing"
 
+	esv1 "github.com/elastic/cloud-on-k8s/v2/pkg/apis/elasticsearch/v1"
+	"github.com/elastic/cloud-on-k8s/v2/pkg/controller/common/certificates"
+	"github.com/elastic/cloud-on-k8s/v2/pkg/controller/common/comparison"
+	"github.com/elastic/cloud-on-k8s/v2/pkg/controller/elasticsearch/label"
+	"github.com/elastic/cloud-on-k8s/v2/pkg/utils/k8s"
```
naemono, line 16: I suspect the moving of these imports wasn't intentional.

pkg/controller/elasticsearch/certificates/transport/reconcile_test.go
```diff
@@ -188,13 +191,49 @@ func TestReconcileTransportCertificatesSecrets(t *testing.T) {
...
+				t.Helper()
+				// Check that there is 1 Secret per StatefulSet
+				assert.Equal(t, 1, len(secrets.Items))
+
+				transportCerts1 := getSecret(secrets, "test-es-name-es-sset1-es-transport-certs")
+				assert.NotNil(t, transportCerts1)
+				// Transport certs CA is still there
+				assert.Contains(t, transportCerts1.Data, "ca.crt")
+				// disabled marker is there
+				assert.Contains(t, transportCerts1.Data, "transport.certs.disabled")
```
naemono, line 221: I think you can use the constant here.

pkg/controller/elasticsearch/certificates/transport/reconcile_test.go
```diff
@@ -425,3 +464,95 @@ func Test_ensureTransportCertificateSecretExists(t *testing.T) {
...
+		{
+			name: "default settings",
+			args: args{
+				secret:        &corev1.Secret{Data: map[string][]byte{}},
+				ca:            testRSACA,
+				additionalCAs: nil,
+			},
+			assertSecrets: func(t *testing.T, secret map[string][]byte) {
+				t.Helper()
+				assert.Contains(t, secret, "ca.crt")
```
naemono, line 488: Lots of `"ca.crt"` in this file; should we just use the `certificates.CAFileName` in `pkg/controller/common/certificates`?

naemono (commented):
Initial walk through code, looks good. Couple of minor things. Haven't tested as of yet.

pkg/controller/elasticsearch/certificates/transport/reconcile_test.go
pebrc: No some autoformat thingy. Will reverse.

pkg/controller/elasticsearch/certificates/transport/reconcile_test.go
pebrc: I try to avoid to use constants in tests, otherwise defintion and check are the same thing and nothing is tested. Not a big deal here as the value is effectively arbitrary, but as a matter of principle.

docs/orchestrating-elastic-stack-applications/elasticsearch/transport-settings.asciidoc
```diff
@@ -111,12 +113,13 @@ spec:
...
             volumeAttributes:
-              csi.cert-manager.io/issuer-name: ca-cluster-issuer
+              csi.cert-manager.io/issuer-name: ca-cluster-issuer <2>
               csi.cert-manager.io/issuer-kind: ClusterIssuer
               csi.cert-manager.io/dns-names: "${POD_NAME}.${POD_NAMESPACE}.svc.cluster.local"
 ----
+<1> Disables the self-signed certificates generated by ECK for the transport layer.
 
-The example assumes that a `ClusterIssuer` by the name of `ca-cluster-issuer` exists and a PEM encoded version of the CA certificate is available in a ConfigMap (in the example named `trust`).  The CA certificate must be in a file called  `ca.crt` inside the ConfigMap in the same namespace as the Elasticsearch resource.
+<2> The example assumes that a `ClusterIssuer` by the name of `ca-cluster-issuer` exists and a PEM encoded version of the CA certificate is available in a ConfigMap (in the example named `trust`).  The CA certificate must be in a file called `ca.crt` inside the ConfigMap in the same namespace as the Elasticsearch resource.
```
naemono, line 123: I think the below `Certificate.spec.issuerRef.name` needs to be updated to `ca-cluster-issuer`, not `selfsigned-issuer`.

docs/orchestrating-elastic-stack-applications/elasticsearch/transport-settings.asciidoc
```diff
@@ -111,12 +113,13 @@ spec:
             driver: csi.cert-manager.io
             readOnly: true
             volumeAttributes:
-              csi.cert-manager.io/issuer-name: ca-cluster-issuer
+              csi.cert-manager.io/issuer-name: ca-cluster-issuer <2>
```
naemono, line 116: Should we mention that [self-signed issuers](https://cert-manager.io/docs/configuration/selfsigned/) cannot be used with the CSI driver? (The `certificaterequests` created by the csi driver do not include `cert-manager.io/private-key-secret-name` annotation)

naemono (approved):
@pebrc  Overall I approve this, but I am curious about transitioning between the 2 options of remote-issued transport certificates, and ECK self-signed certificates. I attempted to go from cert-manager issues transport => ECK transport and it just fails with:

```
exception caught on transport layer:
javax.net.ssl.SSLHandshakeException: PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target
```

Since you mention in the original message `to support the transition between externally provisioned certs and self-signed certs provisioned by ECK`, it sounds as though this transition would be supported. Is that indeed the case?

Is there some needed documentation to transition between the 2 states?

docs/orchestrating-elastic-stack-applications/elasticsearch/transport-settings.asciidoc
pebrc: So the idea was that we have a self-signed issuer which issues a certificate (`root-ca-secret`) which is then used in secod issuer the `ca-cluster-issuer` to work around the restriction of the CSI driver that it cannot work with self-signed issuers

pebrc:
> Is there some needed documentation to transition between the 2 states?

I think the important part when transitioning from one state to the other is to keep the CA ie. the trust bundle around until all nodes have rolled over to the self-signed mode and vice versa. 

For the transition to the externally created certificates we have included extra logic to keep the self-signed CA around until there is no longer  a pod requiring them. 

For the transiition form externally provided certificates to the ECK self-signed ones one has to apply multiple iterations: 

```yaml
kind: Elasticsearch
metadata:
  name: es
spec:
  version: 8.6.2 
  transport:
    tls:
      certificateAuthorities:
        configMapName: trust 
      selfSignedCertificates:
        disabled: false
  nodeSets:
  - name: mixed
    count: 3
    config:
      node.store.allow_mmap: false
```

which switches back to self-signed ECK certs and removes the driver volume and the config attributes and then in a second step also remove the trust in the extra CA in `.spec.transport.tls.certificateAuthorities`

barkbay (commented):
Not from your PR but the example in the documentation cannot be applied, the reason is that the cluster issuer `selfsigned-issuer` is not provided: https://github.com/elastic/cloud-on-k8s/blob/f91321445748b421d889e034cb6d45856eef75f2/docs/orchestrating-elastic-stack-applications/elasticsearch/transport-settings.asciidoc?plain=1#L151

```yaml
  - lastTransitionTime: "2024-07-22T07:51:15Z"
    message: 'Referenced "ClusterIssuer" not found: clusterissuer.cert-manager.io
      "selfsigned-issuer" not found'
    reason: Pending
    status: "False"
    type: Ready
```

In order to document a working example should we add that issuer to the snippet?

```yaml
apiVersion: cert-manager.io/v1
kind: ClusterIssuer
metadata:
  name: selfsigned-issuer
spec:
  selfSigned: {}
```

pebrc:
> In order to document a working example should we add that issuer to the snippet?

Yes I think that's a good idea.

barkbay (approved):
LGTM 👍

pebrc:
I added a bit to the docs to make the self-signed issuer explicit in https://github.com/elastic/cloud-on-k8s/pull/7925/commits/ef8386bf13d5cb53fb4debe78223068eb9809468

PR #7927, opened 2024-07-03
Add metadata field to ES role model

n1v0lg:
Elasticsearch roles also include a `metadata` field ([docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/security-api-put-role.html)). 

This PR updates the `Role` model accordingly. 

Relates: ES-8909


pkg/controller/elasticsearch/client/client.go
```diff
@@ -50,9 +50,10 @@ type ApplicationRole struct {
 
 // Role represents an Elasticsearch role.
 type Role struct {
-	Cluster      []string          `json:"cluster,omitempty"`
-	Indices      []IndexRole       `json:"indices,omitempty"`
-	Applications []ApplicationRole `json:"applications,omitempty"`
+	Cluster      []string               `json:"cluster,omitempty"`
+	Indices      []IndexRole            `json:"indices,omitempty"`
+	Applications []ApplicationRole      `json:"applications,omitempty"`
+	Metadata     map[string]interface{} `json:"metadata,omitempty"`
```
naemono, line 56: This seems reasonable. The original PR that implemented this:

https://github.com/elastic/cloud-on-k8s/pull/426/files#diff-8529b163021d25abbacbdccef966e4be6e2430b357fa6eafa299e23efd1f184bR48

had `Metadata` as a slightly different field. What was the intent of the previous design @thbkrkr  (since you made the change).

Looking back over v6, v7, and v8 of Elasticsearch, this `metadata` field has existed in the same form `object`, so I see no backwards compatibility issues with this change.

thbkrkr: AFAIR, there was no real intention, it was unnecessary premature anticipation. I had left in comments, because not used, different fields seen in an ES response that we could potentially have used one day. We removed them in #2777.
I think the small change of this PR is ok. Nit: we could use the `any` alias which is more pleasant to read.
```suggestion
	Metadata     map[string]any `json:"metadata,omitempty"`
```

n1v0lg: Thanks both. I pushed the [readibility improvement](https://github.com/elastic/cloud-on-k8s/pull/7927/commits/50a2fbacc103a304dbe8e31874847cc8086c0a72)

thbkrkr (approved):

thbkrkr:

n1v0lg:

n1v0lg:
@thbkrkr thanks for the review! Could you merge this once CI is green? Turns out I don't have permissions to merge.

PR #7941, opened 2024-07-15
Elasticsearch/Kibana Helm ECK-Stack Ingress support

naemono:
This change adds Helm Ingress support to both ECK-Elasticsearch and ECK-Kibana Helm Charts. There's still an open question [here](https://github.com/elastic/cloud-on-k8s/pull/7839#discussion_r1679433899), but after some internal discussions we feel that the simplified version here is a good starting point. If there's demand for the ability to add additional hosts/tls sections to ES/KB ingress, then we can add that support in a future relase. Removing draft status to enable review.

TODO:
- [x] documentation in `/docs`.


deploy/eck-stack/charts/eck-elasticsearch/templates/ingress.yaml
```diff
@@ -0,0 +1,46 @@
...
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  name: {{ include "elasticsearch.fullname" . }}
+  labels:
+    {{- with .Values.ingress.labels }}
+    {{- toYaml . | nindent 4 }}
+    {{- end }}
+  annotations:
+    eck.k8s.elastic.co/license: basic
```
kvalliyurnatt, line 12: why is the annotation being set as basic always ?

deploy/eck-stack/charts/eck-kibana/templates/ingress.yaml
```diff
@@ -0,0 +1,46 @@
...
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  name: {{ include "kibana.fullname" . }}
+  labels:
+    {{- with .Values.ingress.labels }}
+    {{- toYaml . | nindent 4 }}
+    {{- end }}
+  annotations:
+    eck.k8s.elastic.co/license: basic
```
kvalliyurnatt, line 12: same question here as well

deploy/eck-stack/charts/eck-kibana/examples/ingress/kibana-eks.yaml
```diff
@@ -0,0 +1,44 @@
...
+  className: alb
+  annotations:
+    alb.ingress.kubernetes.io/scheme: "internet-facing"
+    alb.ingress.kubernetes.io/listen-ports: '[{"HTTPS":443}]'
+    alb.ingress.kubernetes.io/backend-protocol: "HTTPS"
+    alb.ingress.kubernetes.io/target-type: "ip"
+    # To use an ALB with ECK, you must provide a valid ACM certificate ARN or use certificate discovery:
+    # ref: https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.8/guide/ingress/cert_discovery/
+    #
+    # alb.ingress.kubernetes.io/certificate-arn: "arn:aws:acm:us-east-1:00000000000:certificate/b65be571-8220-4f2e-8cb1-94194535d877"
```
kvalliyurnatt, line 28: why is this commented out if this is required ?

naemono: An acm certificate is absolutely required. Apparently Discovery can attempt to find it automatically, or you can statically configure it here. Since this isn't a valid Arn, I commented it out. Happy to uncomment it, or add additional documentation to make it clearer. Let me know

kvalliyurnatt: aaah I see, maybe we can add a note to say this needs to be uncommented out and updated with a valid ARN for the ingress to work correrctly.

deploy/eck-stack/charts/eck-elasticsearch/templates/ingress.yaml
naemono: Good point, this isn't required, and I removed it, but it also made me think about the fact that we should be added the "default" helm labels as well, which I've done, and added tests for both es/kb ingress.

pebrc (approved):
LGTM however I only tested the GKE classic application lb option.

kvalliyurnatt (approved):

Issue #7953, opened 2024-07-23
[E2E] Upgrading to Kibana 7.17.21 may fail


barkbay:
Likely related to https://github.com/elastic/kibana/pull/158182 which is not available in 7.x until `7.17.23`.

Kibana Pod never comes ready, following error appears in the logs:

```
{"type":"log","@timestamp":"2024-07-22T17:19:10+00:00","tags":["info","savedobjects-service"],"pid":7,"message":"[.kibana] Migration completed after 319ms"}
{"type":"log","@timestamp":"2024-07-22T17:21:10+00:00","tags":["error","savedobjects-service"],"pid":7,"message":"[.kibana_task_manager] Action failed with 'Request timed out'. Retrying attempt 1 in 2 seconds."}
{"type":"log","@timestamp":"2024-07-22T17:21:10+00:00","tags":["info","savedobjects-service"],"pid":7,"message":"[.kibana_task_manager] REINDEX_SOURCE_TO_TEMP_INDEX_BULK -> REINDEX_SOURCE_TO_TEMP_INDEX_BULK. took: 120008ms."}
...
{"type":"log","@timestamp":"2024-07-22T17:47:32+00:00","tags":["error","savedobjects-service"],"pid":7,"message":"[.kibana_task_manager] Action failed with 'Request timed out'. Retrying attempt 11 in 64 seconds."}
{"type":"log","@timestamp":"2024-07-22T17:47:32+00:00","tags":["info","savedobjects-service"],"pid":7,"message":"[.kibana_task_manager] REINDEX_SOURCE_TO_TEMP_INDEX_BULK -> REINDEX_SOURCE_TO_TEMP_INDEX_BULK. took: 184003ms."}
==== END logs for e2e-kktsf-mercury/test-version-upgrade-to-7x-tpcr-kb-599484746c-zw4kh ====
```

Kibana is actually waiting for some shards to be assigned:

```
.kibana_task_manager_7.17.21_reindex_temp                     0     p      STARTED       0    227b 10.37.129.117 test-version-upgrade-to-7x-g597-es-masterdata-0
.kibana_task_manager_7.17.21_reindex_temp                     0     r      UNASSIGNED  
```

Which is never going to happen since tests like `TestVersionUpgradeToLatest7x` or `TestVersionUpgradeAndRespecToLatest7x` are run with a 1 node ES cluster.

One workaround would be to add a version gate to create ES clusters with at least 2 nodes (credits to @pebrc for the idea)

[...]

PR #7954, opened 2024-07-23
[E2E] Use 2 nodes clusters for Kibana upgrade tests to 7.x

barkbay:
Workaround for https://github.com/elastic/cloud-on-k8s/issues/7953


pebrc (approved):
Thanks for fixing what I should have fixed days ago 🙇

Issue #7742, opened 2024-04-23
Logstash StatefulSet initContainers command is not properly reconciliated


Kiliandeca:
## Bug Report
**What did you do?**
I recently upgraded ECK from version `2.11.1` to `2.12.1`.

**What did you expect to see?**
I expected the upgrade to work transparently.

**What did you see instead? Under which circumstances?**
My new Logstash pods being rolled out started to fail with the following error message:
```
Your settings are invalid. Reason: File "/usr/share/logstash/config/api_keystore.p12" must exist but was not found.
```

After investigating what could be root cause I found out it's related to this change https://github.com/elastic/cloud-on-k8s/pull/7408
To summarize what I think is happening, the [PR properly changed](https://github.com/elastic/cloud-on-k8s/pull/7408/files#diff-000e81cb01c6f6b546ab205bc72599d2cc662ddcb8c5df9106eb7a2dd316c25aR26-R39) the `initContainers` to generate a Keystore for Logstash so it would not be breaking but my Statefulset `initContainers` didn't include the new script it was stuck on the old script which explain why my pod is failing.

In the end, I solved my problem by deleting the Logstash Statefulset without deleting the pods that were still running if this can help anyone. The operator immediately recreated the Statefulset with the new `initContainers` script as expected.
```
kubectl delete statefulset mylogstash-ls --cascade=orphan
```

I would expect ECK to update the `initContainers` command script during the reconciliation and I think that's a bug? 
Here's the instructions to reproduce this behavior:
- Creates a Logstash resource and wait for it to be deployed
- Manually update the Statefulset `initContainers` command script by adding anything
- The ECK operator never replace the change you made

```
apiVersion: logstash.k8s.elastic.co/v1alpha1
kind: Logstash
metadata:
  name: quickstart
spec:
  count: 1
  version: 8.13.2
  pipelines:
    - pipeline.id: main
      config.string: |
        input {
          beats {
            port => 5044
          }
        }
  services:
    - name: beats
      service:
        spec:
          type: NodePort
          ports:
            - port: 5044
              name: "filebeat"
              protocol: TCP
              targetPort: 5044
```

`quickstart-ls` Statefulset:
```
      initContainers:
        - name: logstash-internal-init-config
          image: docker.elastic.co/logstash/logstash:8.13.2
          command:
            - /usr/bin/env
            - bash
            - '-c'
[...]

PR #7982, opened 2024-08-01
Upgrade test harness fixes

barkbay:
This PR fixes a couple of issues wrt our upgrade test harness framework:

* Only test Logstash starting ECK `2.13` since it was in tech. preview before, and also because upgrading from a version below `2.12` is not trivial because of:
  *  this bug: https://github.com/elastic/cloud-on-k8s/issues/7742
  *  the `health` field which did not exit in the first releases
* Fix a bug in the `ReplaceOrCreate` function where objects where actually not replaced (more specifically they were replaced by their current versions, not by the one in the manifest)
* Fixes a few consistencies in the expected stack versions VS what versions are actually in the stack manifests.

### DIFF
--- a/hack/upgrade-test-harness/k8s/kubectl.go
+++ b/hack/upgrade-test-harness/k8s/kubectl.go
@@ -117,7 +117,7 @@ func (h *Kubectl) GetPods(namespace string, labelSelector string) ([]corev1.Pod,
 
 // LoadResources loads manifests from the given file path.
 func (h *Kubectl) LoadResources(filePath string) (*resource.Result, error) {
-	validator, err := h.factory.Validator(true)
+	validator, err := h.factory.Validator(metav1.FieldValidationStrict)
 	if err != nil {
 		return nil, fmt.Errorf("failed to obtain validator: %w", err)
 	}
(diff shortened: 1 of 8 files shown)


hack/upgrade-test-harness/k8s/kubectl.go
```diff
@@ -230,6 +230,10 @@ func (h *Kubectl) ReplaceOrCreate(resources resource.Visitor) error {
 			return err
 		}
 
+		// Create a copy of the object read from the manifest. This is required because info.Get() has the side effect
+		// of replacing the content of info.Object with what is *already* deployed, leading to a no-op operation if the
+		// object must be replaced later using the Replace() function.
+		object := info.Object.DeepCopyObject()
```
pebrc, line 236: Great finding 🤦

hack/upgrade-test-harness/testdata/v2100/install.yaml
```diff
@@ -451,7 +451,7 @@ spec:
       securityContext:
         runAsNonRoot: true
       containers:
-      - image: "docker.elastic.co/eck-dev/eck-operator-mmontgomery:2.10.0-63c18ccf"
```
pebrc, line 454: Oh wow

pebrc (approved):

PR #7990, opened 2024-08-05
runner: preallocate memory

florianl:
`toList()` knows upfront the number of tags that will end up in the list. Preallocating the exact size upfront reduces memory and GC operations.


barkbay (approved):
LGTM, thanks.

PR #8020, opened 2024-08-20
Upgrade to controller-runtime 0.19.0

naemono:
## Changes Detailed

There are additional breaking changes with controller-runtime 0.19.0, which mostly come down to this [breaking PR](https://github.com/kubernetes-sigs/controller-runtime/pull/2799). This integrates those changes into the existing codebase.

### Additional changes

Was forced to update `controller-tools`, see [failures](https://buildkite.com/elastic/cloud-on-k8s-operator/builds/9034#01917092-f659-461b-a279-6c1a111d9f96)
Also had to bump the memory from 2Gi to 4Gi (I tried 3Gi):
image


adamkasztenny (approved):
LGTM

test/e2e/cmd/run/eventlog.go
```diff
@@ -106,7 +106,7 @@ func (el *eventLogger) runEventProcessor() {
 			return
 		}
 
-		evtObj, exists, err := el.eventInformer.GetIndexer().GetByKey(key.(string)) //nolint:forcetypeassert
+		evtObj, exists, err := el.eventInformer.GetIndexer().GetByKey(key) //nolint:forcetypeassert
```
adamkasztenny, line 109: I think this can be removed since there isn't a type assertion anymore

```suggestion
		evtObj, exists, err := el.eventInformer.GetIndexer().GetByKey(key)
```

pkg/controller/common/watches/state.go
```diff
@@ -7,25 +7,26 @@ package watches
...
-		ConfigMaps:          NewDynamicEnqueueRequest[*corev1.ConfigMap](),
-		Secrets:             NewDynamicEnqueueRequest[*corev1.Secret](),
-		Services:            NewDynamicEnqueueRequest[*corev1.Service](),
-		Pods:                NewDynamicEnqueueRequest[*corev1.Pod](),
-		ReferencedResources: NewDynamicEnqueueRequest[client.Object](),
+		ConfigMaps:          NewDynamicEnqueueRequest[*corev1.ConfigMap, reconcile.Request](),
+		Secrets:             NewDynamicEnqueueRequest[*corev1.Secret, reconcile.Request](),
+		Services:            NewDynamicEnqueueRequest[*corev1.Service, reconcile.Request](),
+		Pods:                NewDynamicEnqueueRequest[*corev1.Pod, reconcile.Request](),
+		ReferencedResources: NewDynamicEnqueueRequest[client.Object, reconcile.Request](),
```
pebrc, line 20: You don't need the extra type parameter if it is always going to be a `reconcile.Request`
```diff
diff --git a/pkg/controller/association/dynamic_watches.go b/pkg/controller/association/dynamic_watches.go
index 66a93cd48..bc0927291 100644
--- a/pkg/controller/association/dynamic_watches.go
+++ b/pkg/controller/association/dynamic_watches.go
@@ -132,7 +132,7 @@ func (r *Reconciler) reconcileWatches(ctx context.Context, associated types.Name
 func reconcileGenericWatch[T client.Object, R reconcile.Request](
 	associated types.NamespacedName,
 	associations []commonv1.Association,
-	dynamicRequest *watches.DynamicEnqueueRequest[T, reconcile.Request],
+	dynamicRequest *watches.DynamicEnqueueRequest[T],
 	watchName string,
 	watchedFunc func() ([]types.NamespacedName, error),
 ) error {
@@ -158,7 +158,7 @@ func reconcileGenericWatch[T client.Object, R reconcile.Request](
 func ReconcileWatch[T client.Object, R reconcile.Request](
 	associated types.NamespacedName,
 	associations []commonv1.Association,
-	dynamicRequest *watches.DynamicEnqueueRequest[T, reconcile.Request],
+	dynamicRequest *watches.DynamicEnqueueRequest[T],
 	watchName string,
 	watchedFunc func(association commonv1.Association) types.NamespacedName,
 ) error {
@@ -177,7 +177,7 @@ func ReconcileWatch[T client.Object, R reconcile.Request](
 }
 
 // RemoveWatch removes `watchName` watch from `dynamicRequest`.
-func RemoveWatch[T client.Object, R comparable](dynamicRequest *watches.DynamicEnqueueRequest[T, R], watchName string) {
+func RemoveWatch[T client.Object](dynamicRequest *watches.DynamicEnqueueRequest[T], watchName string) {
 	dynamicRequest.RemoveHandlerForKey(watchName)
 }
 
diff --git a/pkg/controller/common/watches/handler.go b/pkg/controller/common/watches/handler.go
index b039c32a2..f5b653960 100644
--- a/pkg/controller/common/watches/handler.go
+++ b/pkg/controller/common/watches/handler.go
@@ -25,29 +25,29 @@ var (
 
 // HandlerRegistration is the event handler registration that can be added or removed
 // from DynamicEnqueueRequest.
-type HandlerRegistration[T client.Object, R comparable] interface {
+type HandlerRegistration[T client.Object] interface {
 	// Key identifies the transformer
 	Key() string
 	// EventHandler handles CRUD events and turns them into reconcile.Request if relevant.
-	EventHandler() handler.TypedEventHandler[T, R]
+	EventHandler() handler.TypedEventHandler[T, reconcile.Request]
 }
 
 // NewDynamicEnqueueRequest creates a new DynamicEnqueueRequest
-func NewDynamicEnqueueRequest[T client.Object, R reconcile.Request]() *DynamicEnqueueRequest[T, R] {
-	return &DynamicEnqueueRequest[T, R]{
-		registrations: make(map[string]HandlerRegistration[T, R]),
+func NewDynamicEnqueueRequest[T client.Object]() *DynamicEnqueueRequest[T] {
+	return &DynamicEnqueueRequest[T]{
+		registrations: make(map[string]HandlerRegistration[T]),
 	}
 }
 
 // DynamicEnqueueRequest is an EventHandler that allows addition and removal of
 // event handler registrations at runtime allowing dynamic reconciliation based on specific resources.
-type DynamicEnqueueRequest[T client.Object, R comparable] struct {
+type DynamicEnqueueRequest[T client.Object] struct {
 	mutex         sync.RWMutex
-	registrations map[string]HandlerRegistration[T, R]
+	registrations map[string]HandlerRegistration[T]
 }
 
 // AddHandlers adds the new event handlers to this DynamicEnqueueRequest.
-func (d *DynamicEnqueueRequest[T, R]) AddHandlers(handlers ...HandlerRegistration[T, R]) error {
+func (d *DynamicEnqueueRequest[T]) AddHandlers(handlers ...HandlerRegistration[T]) error {
 	for _, h := range handlers {
 		if err := d.AddHandler(h); err != nil {
 			return err
@@ -57,7 +57,7 @@ func (d *DynamicEnqueueRequest[T, R]) AddHandlers(handlers ...HandlerRegistratio
 }
 
 // AddHandler adds a new event handler to this DynamicEnqueueRequest.
-func (d *DynamicEnqueueRequest[T, R]) AddHandler(handler HandlerRegistration[T, R]) error {
+func (d *DynamicEnqueueRequest[T]) AddHandler(handler HandlerRegistration[T]) error {
 	d.mutex.Lock()
 	defer d.mutex.Unlock()
 
@@ -70,19 +70,19 @@ func (d *DynamicEnqueueRequest[T, R]) AddHandler(handler HandlerRegistration[T,
 }
 
 // RemoveHandler removes the handler defined by the transformer.
-func (d *DynamicEnqueueRequest[T, R]) RemoveHandler(handler HandlerRegistration[T, R]) {
+func (d *DynamicEnqueueRequest[T]) RemoveHandler(handler HandlerRegistration[T]) {
 	d.RemoveHandlerForKey(handler.Key())
 }
 
 // RemoveHandlerForKey removes the handler identified by the given key.
-func (d *DynamicEnqueueRequest[T, R]) RemoveHandlerForKey(key string) {
+func (d *DynamicEnqueueRequest[T]) RemoveHandlerForKey(key string) {
 	d.mutex.Lock()
 	defer d.mutex.Unlock()
 	delete(d.registrations, key)
 }
 
 // Registrations returns the list of registered handler names.
-func (d *DynamicEnqueueRequest[T, R]) Registrations() []string {
+func (d *DynamicEnqueueRequest[T]) Registrations() []string {
 	d.mutex.RLock()
 	defer d.mutex.RUnlock()
 	keys := make([]string, 0, len(d.registrations))
@@ -93,10 +93,10 @@ func (d *DynamicEnqueueRequest[T, R]) Registrations() []string {
 }
 
 // DynamicEnqueueRequest implements TypedEventHandler
-var _ handler.TypedEventHandler[client.Object, reconcile.Request] = &DynamicEnqueueRequest[client.Object, reconcile.Request]{}
+var _ handler.TypedEventHandler[client.Object, reconcile.Request] = &DynamicEnqueueRequest[client.Object]{}
 
 // Create is called in response to a create event - e.g. Pod Creation.
-func (d *DynamicEnqueueRequest[T, R]) Create(ctx context.Context, evt event.TypedCreateEvent[T], q workqueue.TypedRateLimitingInterface[R]) {
+func (d *DynamicEnqueueRequest[T]) Create(ctx context.Context, evt event.TypedCreateEvent[T], q workqueue.TypedRateLimitingInterface[reconcile.Request]) {
 	d.mutex.RLock()
 	defer d.mutex.RUnlock()
 	for _, v := range d.registrations {
@@ -105,7 +105,7 @@ func (d *DynamicEnqueueRequest[T, R]) Create(ctx context.Context, evt event.Type
 }
 
 // Update is called in response to an update event -  e.g. Pod Updated.
-func (d *DynamicEnqueueRequest[T, R]) Update(ctx context.Context, evt event.TypedUpdateEvent[T], q workqueue.TypedRateLimitingInterface[R]) {
+func (d *DynamicEnqueueRequest[T]) Update(ctx context.Context, evt event.TypedUpdateEvent[T], q workqueue.TypedRateLimitingInterface[reconcile.Request]) {
 	d.mutex.RLock()
 	defer d.mutex.RUnlock()
 	for _, v := range d.registrations {
@@ -114,7 +114,7 @@ func (d *DynamicEnqueueRequest[T, R]) Update(ctx context.Context, evt event.Type
 }
 
 // Delete is called in response to a delete event - e.g. Pod Deleted.
-func (d *DynamicEnqueueRequest[T, R]) Delete(ctx context.Context, evt event.TypedDeleteEvent[T], q workqueue.TypedRateLimitingInterface[R]) {
+func (d *DynamicEnqueueRequest[T]) Delete(ctx context.Context, evt event.TypedDeleteEvent[T], q workqueue.TypedRateLimitingInterface[reconcile.Request]) {
 	d.mutex.RLock()
 	defer d.mutex.RUnlock()
 	for _, v := range d.registrations {
@@ -124,7 +124,7 @@ func (d *DynamicEnqueueRequest[T, R]) Delete(ctx context.Context, evt event.Type
 
 // Generic is called in response to an event of an unknown type or a synthetic event triggered as a cron or
 // external trigger request - e.g. reconcile Autoscaling, or a Webhook.
-func (d *DynamicEnqueueRequest[T, R]) Generic(ctx context.Context, evt event.TypedGenericEvent[T], q workqueue.TypedRateLimitingInterface[R]) {
+func (d *DynamicEnqueueRequest[T]) Generic(ctx context.Context, evt event.TypedGenericEvent[T], q workqueue.TypedRateLimitingInterface[reconcile.Request]) {
 	d.mutex.RLock()
 	defer d.mutex.RUnlock()
 	for _, v := range d.registrations {
diff --git a/pkg/controller/common/watches/state.go b/pkg/controller/common/watches/state.go
index 805baa39b..d5f8db646 100644
--- a/pkg/controller/common/watches/state.go
+++ b/pkg/controller/common/watches/state.go
@@ -7,26 +7,25 @@ package watches
 import (
 	corev1 "k8s.io/api/core/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
 // NewDynamicWatches creates an initialized DynamicWatches container.
 func NewDynamicWatches() DynamicWatches {
 	return DynamicWatches{
-		ConfigMaps:          NewDynamicEnqueueRequest[*corev1.ConfigMap, reconcile.Request](),
-		Secrets:             NewDynamicEnqueueRequest[*corev1.Secret, reconcile.Request](),
-		Services:            NewDynamicEnqueueRequest[*corev1.Service, reconcile.Request](),
-		Pods:                NewDynamicEnqueueRequest[*corev1.Pod, reconcile.Request](),
-		ReferencedResources: NewDynamicEnqueueRequest[client.Object, reconcile.Request](),
+		ConfigMaps:          NewDynamicEnqueueRequest[*corev1.ConfigMap](),
+		Secrets:             NewDynamicEnqueueRequest[*corev1.Secret](),
+		Services:            NewDynamicEnqueueRequest[*corev1.Service](),
+		Pods:                NewDynamicEnqueueRequest[*corev1.Pod](),
+		ReferencedResources: NewDynamicEnqueueRequest[client.Object](),
 	}
 }
 
 // DynamicWatches contains stateful dynamic watches. Intended as facility to pass around stateful dynamic watches and
 // give each of them an identity.
 type DynamicWatches struct {
-	ConfigMaps          *DynamicEnqueueRequest[*corev1.ConfigMap, reconcile.Request]
-	Secrets             *DynamicEnqueueRequest[*corev1.Secret, reconcile.Request]
-	Services            *DynamicEnqueueRequest[*corev1.Service, reconcile.Request]
-	Pods                *DynamicEnqueueRequest[*corev1.Pod, reconcile.Request]
-	ReferencedResources *DynamicEnqueueRequest[client.Object, reconcile.Request]
+	ConfigMaps          *DynamicEnqueueRequest[*corev1.ConfigMap]
+	Secrets             *DynamicEnqueueRequest[*corev1.Secret]
+	Services            *DynamicEnqueueRequest[*corev1.Service]
+	Pods                *DynamicEnqueueRequest[*corev1.Pod]
+	ReferencedResources *DynamicEnqueueRequest[client.Object]
 }
```

pebrc (changes requested):

pkg/controller/common/watches/state.go
naemono: @pebrc Ok, this finally sunk into my head this time, and I get this. Sorry for the overdone typing. I can see when these are not needed now. Sorry, generics hadn't 'clicked' yet for me, I guess. I went through and corrected the others that weren't needed as well.

pebrc (commented):
Codewise looks good. There seems to be an issue though with the `required` annotation on the `elasticsearchRef`s in the monitoring subsection in the Elasticsearch CRD, probably something has changed in controller-tools, we need to double check what is going on there and if other CRDs are affected as well.

config/crds/v1/bases/elasticsearch.k8s.elastic.co_elasticsearches.yaml
```diff
@@ -566,6 +560,8 @@ spec:
                               type: string
                           type: object
                         type: array
+                    required:
+                    - elasticsearchRefs
```
pebrc, line 564: This is unexpected

naemono: Just so we are on the same page, the top level field `Monitoring` is set to optional:
https://github.com/elastic/cloud-on-k8s/blob/e9c6a259f46610fb5e14dc30c3e566b09d3ec8df/pkg/apis/elasticsearch/v1/elasticsearch_types.go#L135
The `Metrics` and `Logs` fields are also set to optional:
https://github.com/elastic/cloud-on-k8s/blob/e9c6a259f46610fb5e14dc30c3e566b09d3ec8df/pkg/apis/common/v1/monitoring.go#L11
And the `ElasticsearchRefs` for both of these are set to required:
https://github.com/elastic/cloud-on-k8s/blob/e9c6a259f46610fb5e14dc30c3e566b09d3ec8df/pkg/apis/common/v1/monitoring.go#L23

So maybe this is a bug fix that it's now doing what it's supposed to do? I'm still investigating, and looking through issues upstream, and will test to see if this causes any issues.

naemono: Yep, here we go:
https://github.com/kubernetes-sigs/controller-tools/issues/943
https://github.com/kubernetes-sigs/controller-tools/pull/944
Here's what we have:
```
	// +kubebuilder:validation:Required
	ElasticsearchRefs []ObjectSelector `json:"elasticsearchRefs,omitempty"`
```
Which is both `omitempty` and `Required`, which used to be treated as optional, is now explicitly required.

naemono: With something like the following, we get an expected failure:
```
spec:
  version: 8.15.0
  monitoring:
    metrics: {}
    logs: {}
```
Failure:
```
❯ kc apply -n default -f config/custom/elasticsearch-with-monitoring.yaml
elasticsearch.elasticsearch.k8s.elastic.co/elasticsearch-monitor-sink created
The Elasticsearch "elasticsearch-monitor-source" is invalid:
* spec.monitoring.logs.elasticsearchRefs: Required value
* spec.monitoring.metrics.elasticsearchRefs: Required value
```
And adjusting to this:
```
spec:
  version: 8.15.0
  monitoring:
    metrics:
      elasticsearchRefs:
      - name: elasticsearch-monitor-sink
        namespace: default
    logs:
      elasticsearchRefs:
      - name: elasticsearch-monitor-sink
        namespace: default
```
Gets the expected results:
```
❯ kc apply -n default -f config/custom/elasticsearch-with-monitoring.yaml
elasticsearch.elasticsearch.k8s.elastic.co/elasticsearch-monitor-sink unchanged
elasticsearch.elasticsearch.k8s.elastic.co/elasticsearch-monitor-source created
```

naemono: Oh wait, from the failure in the smoke tests:
```
        	            	Elasticsearch.elasticsearch.k8s.elastic.co "es-apm-sample-n9bk" is invalid: [spec.monitoring.logs.elasticsearchRefs: Required value, spec.monitoring.metrics.elasticsearchRefs: Required value]
```
There must be another permutation that could cause issues... checking.

naemono: So it must be the webhook doing this, as I'm running with just the built operator, without the webook, but see this in the logs when using the `es-apm-sample` we provide:
```
2024-08-21T08:57:46.026-0500    ERROR   manager.eck-operator    Reconciler error        {"service.version": "2.15.0-SNAPSHOT+def426fb", "controller": "elasticsearch-controller", "object": {"name":"es-apm-sample","namespace":"default"}, "namespace": "default", "name": "es-apm-sample", "reconcileID": "f1beddda-68ce-4b41-ad07-2459b3dede4e", "error": "[adjust resources: adjust discovery config: Elasticsearch.elasticsearch.k8s.elastic.co \"es-apm-sample\" is invalid: [spec.monitoring.logs.elasticsearchRefs: Required value, spec.monitoring.metrics.elasticsearchRefs: Required value], Elasticsearch.elasticsearch.k8s.elastic.co \"es-apm-sample\" is invalid: [spec.monitoring.logs.elasticsearchRefs: Required value, spec.monitoring.metrics.elasticsearchRefs: Required value]]", "errorCauses": [{"error": "adjust resources: adjust discovery config: Elasticsearch.elasticsearch.k8s.elastic.co \"es-apm-sample\" is invalid: [spec.monitoring.logs.elasticsearchRefs: Required value, spec.monitoring.metrics.elasticsearchRefs: Required value]"}, {"error": "Elasticsearch.elasticsearch.k8s.elastic.co \"es-apm-sample\" is invalid: [spec.monitoring.logs.elasticsearchRefs: Required value, spec.monitoring.metrics.elasticsearchRefs: Required value]"}]}
```

naemono: @pebrc I've found 2 paths to resolve this, none of which I particularly like:
1. Set `monitoring.*.elasticsearchRefs` explicitly to optional, which was how it was being treated previously because of it also having `omitEmpty` in the struct tags.
2. Set `Elasticsearch.Spec.Monitoring` to be a pointer. Now this brings up another whole can of worms, as we're talking about changing APIs that are `V1`, which has major red flags *but* it works (along with a slew of other code changes checking for `nil`). Now note the [k8s api conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required) concerning optional fields, which they indicate should be a pointer, or have built-in nil values [maps/slices], so should these have been a pointer from the beginning?

I've tried other options, such as the following, to no avail:
* adding  `+optional` to the parent optional fields.
* removing `omitEmpty` from ElasticsearchRefs.

pebrc: I think making it `+optional` sounds about right to me, if this was effectively the previous behaviour. Making `Monitoring` a pointer would also be possible I think as long as the serialisation of the struct stays the same (which I think is the case) But a pointer to struct containing a slice feels kind of awkward, the nil/empty slice is the more idiomatic way of expresssing the absence of value. 

I do suggest however to run the full set of e2e tests on this PR before we merge this in case we missed another case.

naemono: @kvalliyurnatt brought up a good point. In addition to setting this to optional, we could add some new validation such that the webhook could catch these being missing, and reject the change. I will also add this.

naemono: Actually, this is more difficult than initially anticipated as we use an non-pointer for the monitoring struct, so there's no difference between the user defining this without any logs/metrics associations, and it just being empty, so this is a check we can't do @kvalliyurnatt .

naemono:
buildkite test this -f p=gke -m s=7.17.8,s=8.15.0,s=8.16.0-SNAPSHOT

config/crds/v1/bases/elasticsearch.k8s.elastic.co_elasticsearches.yaml
pebrc: I vote for keeping the behaviour identical to the status quo (ignoring the difficulty you mentioned).

config/crds/v1/bases/elasticsearch.k8s.elastic.co_elasticsearches.yaml
```diff
@@ -534,6 +528,10 @@ spec:
                         description: |-
                           ElasticsearchRefs is a reference to a list of monitoring Elasticsearch clusters running in the same Kubernetes cluster.
                           Due to existing limitations, only a single Elasticsearch cluster is currently supported.
+
+                          This is not optional but was treated as such in previous version of kubernetes-sigs/controller-tools as we also specify
+                          `omitempty` here. See https://github.com/elastic/cloud-on-k8s/pull/8020 and https://github.com/kubernetes-sigs/controller-tools/issues/943
+                          for more details.
```
pebrc, line 534: I don't think we want that in the public API docs. I think there is a way to make it not show up there.

naemono: I'll investigate and update, thanks...

naemono: Well, I've partially solved this. Adding the `---` removes them from the rendered yaml, but unfortunately I don't think that the crd-ref-docs tool supports this, so it's in the rendered `api-docs` file....

pebrc: Why don't you put the comment above the monitoring struct, basically free-floating so that it doesn not get included. We can then in a second step modify the `crd-ref-docs` tool to follow the k8s convention of ignoring all comments after `---`

naemono: @pebrc this has been done, and the changes look correct now. The previous full e2e test run: https://buildkite.com/elastic/cloud-on-k8s-operator/builds/9053 only had 2 failures, and they appear to be associated with the recent changes you fixed in #8021.

naemono: I also created https://github.com/elastic/crd-ref-docs/issues/93

pebrc (approved):
LGTM!

PR #8022, opened 2024-08-22
[DOC] Fix request endpoint step flow

shainaraskas:
This PR solves some confusion around the order of two steps which represent two paths rather than steps that need to be performed in order

also fixed a numbering issue

original [here](https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-deploy-elasticsearch.html#k8s_request_elasticsearch_access)

image


pebrc (approved):

shainaraskas:
hi @pebrc - looks like I can't merge this on my own. would you be able to hit the button? thanks.

PR #8041, opened 2024-09-04
Fix eck-stack kibana examples

naemono:
While testing out the new eck-stack release, I found that the example was still incorrect. This fixes the example manifest.


barkbay (approved):
LGTM

PR #8051, opened 2024-09-13
Emphasize the importance of having snapshot

kunisen:
**:: Reason for the PR**

We have seen many master pods got lost thus having to repair the whole cluster use cases, and we'd like to emphasize the importance about having snapshot to call out it explicitly clear.

**:: After merge:** 

Orange part will be added

image

Text is something similar from ESS - https://www.elastic.co/guide/en/cloud/current/ec-snapshot-restore.html
```
Snapshots are an efficient way to ensure that your Elasticsearch indices can be recovered in the event of an accidental deletion, or to migrate data across deployments.
```
But we don't say "deployment" in ECK so I used clusters.


docs/orchestrating-elastic-stack-applications/elasticsearch/snapshots.asciidoc
```diff
@@ -8,6 +8,8 @@ endif::[]
 [id="{p}-{page_id}"]
 = Create automated snapshots
 
+NOTE: Snapshots are important to ensure that your Elasticsearch indices can be recovered in the event of an accidental deletion, or to migrate data across clusters.
```
thbkrkr, line 11: It's up to you to decide, but here's another suggestion that's a little more direct and concise.
```suggestion
NOTE: Snapshots are essential for recovering Elasticsearch indices in case of accidental deletion or for migrating data between clusters.
```

thbkrkr (approved):

docs/orchestrating-elastic-stack-applications/elasticsearch/snapshots.asciidoc
barkbay, line 11: Should we have this note as a header of the "Run Elasticsearch on ECK" section [here](https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-elasticsearch-specification.html). I'm asking because my feeling is that if someone is looking at the snapshot documentation then it may mean that this person is _already_ aware that it's an important topic (or maybe because it's too late 🙃 )

kunisen: Thanks @barkbay for being patient. 
Sure. I added this piece into https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-elasticsearch-specification.html too :) 

(Using @thbkrkr 's suggestion above).

image

kunisen: I think it is worth to repeat important things so it makes logical sense to put it in both below pages.
- https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-elasticsearch-specification.html
- https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-snapshots.html

But I will leave the final decision with you in case you feel we should avoid duplicated description. 

Thanks again!

kunisen:
It seems like I don't have permission to merge. Please help in advance and thanks! 🙏 
image

docs/orchestrating-elastic-stack-applications/elasticsearch/snapshots.asciidoc
kunisen: Thanks!

thbkrkr: 👌

Issue #7909, opened 2024-06-18
Fleet server pod not created when Kibana is set with `SERVER_BASEPATH`


msapitree:
## Bug Report
**What did you do?**

Following quickstart 
1) Deployed ECK
2) Deployed Elastic resource
3) Deployed Kibana resource
4) Redeployed Kibana resource with config modification (to setup public access to Kibana via k8s ingress on `https://xxx.mydomain.com/monitoring/kibana`):
5) Deployed Fleet Server Agent (together with  ClusterRole, ClusterRoleBinding and ServiceAccount)

**What did you expect to see?**
- pod is created for the fleet-server agent
- no error in operator logs
- agent is reported as green

**What did you see instead? Under which circumstances?**
- error in operator log  `failed to request http://monitoring-kb-http.monitoring-at.svc:5601/api/fleet/setup, status is 404)`

  Due to the change to server context, the URL path `/api/fleet/setup` is no longer valid and should have become `/monitoring/kibana/api/fleet/setup` considering the `server.basePath` config parameter (or it's env variable equivalent).

- no health information on elastic resource 
```
$kubectl -n monitoring-at get elastic
NAME                                      HEALTH   AVAILABLE   EXPECTED   VERSION   AGE
agent.agent.k8s.elastic.co/fleet-server                                             109m

NAME                                      HEALTH   NODES   VERSION   AGE
apmserver.apm.k8s.elastic.co/monitoring   green    1       8.14.0    5d21h

NAME                                                    HEALTH   NODES   VERSION   PHASE   AGE
elasticsearch.elasticsearch.k8s.elastic.co/monitoring   green    3       8.14.0    Ready   5d22h

NAME                                      HEALTH   NODES   VERSION   AGE
kibana.kibana.k8s.elastic.co/monitoring   green    1       8.14.0    114m
```

**Environment**

* ECK version: 2.13.0

* Kubernetes information: AKS
```
Client Version: v1.29.3
Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
Server Version: v1.28.9
```

* Resource definition:
- Kibana with `SERVER_BASEPATH`

```yaml
apiVersion: kibana.k8s.elastic.co/v1
kind: Kibana
metadata:
  name: monitoring
  namespace: monitoring-at
spec:
  version: 8.14.0
  count: 1
  elasticsearchRef:
    name: monitoring
  http:
    tls:
      selfSignedCertificate:
        disabled: true
[...]

PR #8053, opened 2024-09-18
Handle Kibana basepath updates for fleet

kvalliyurnatt:
fixes: #7909 
Currently the operator does not look at the kibana basePath config and accordingly alter path for fleet, in this PR we have the operator checking the basePath config. Currently this only looks at the config in the KIbana spec, the basePath can also be set in the ENV of the KIbana pod, this is currently not being handled

TODO: 

- [x] Test changes locally


pkg/controller/agent/fleet.go
```diff
@@ -265,6 +277,30 @@ func maybeReconcileFleetEnrollment(params Params, result *reconciler.Results) En
...
 
+func getKibanaBasePath(ctx context.Context, client k8s.Client, kibanaNSN types.NamespacedName) (string, error) {
+	var kb v1.Kibana
+	err := client.Get(ctx, kibanaNSN, &kb)
+	if err != nil {
+		return "", err
+	}
+
+	kibanaRawConfig := kb.Spec.Config
+	if kibanaRawConfig != nil {
```
kvalliyurnatt, line 288: I am not sure if there is a better way to read this config, this seems a little complicated.

kvalliyurnatt: I changed this, I think now it is a little better, thanks @naemono for the suggestion

pkg/controller/agent/fleet.go
```diff
@@ -265,6 +285,29 @@ func maybeReconcileFleetEnrollment(params Params, result *reconciler.Results) En
...
 }
 
+func getKibanaBasePath(ctx context.Context, client k8s.Client, kibanaNSN types.NamespacedName) (string, error) {
+	var kb v1.Kibana
+	err := client.Get(ctx, kibanaNSN, &kb)
+	if err != nil {
+		return "", err
+	}
+
+	if kb.Spec.Config != nil {
```
naemono, line 295: Can we flip this and just return early if `== nil`?  Then the remainder of the function is cleaner, and not in the middle of an if block.

kvalliyurnatt: done.

naemono:
@kvalliyurnatt I'm going to take one final look at this today and it should be g2g.

pkg/controller/agent/fleet.go
```diff
@@ -265,6 +285,30 @@ func maybeReconcileFleetEnrollment(params Params, result *reconciler.Results) En
 	return token
 }
 
+func getKibanaBasePath(ctx context.Context, client k8s.Client, kibanaNSN types.NamespacedName) (string, error) {
+	var kb v1.Kibana
+	err := client.Get(ctx, kibanaNSN, &kb)
```
barkbay, line 290: Does it mean that the Agent controller should watch Kibana resources? (I don't think it is the case, I might be wrong though)

pkg/controller/agent/fleet.go
```diff
@@ -265,6 +285,30 @@ func maybeReconcileFleetEnrollment(params Params, result *reconciler.Results) En
...
+	if err != nil {
+		return "", err
+	}
+
+	if kb.Spec.Config == nil {
+		return "", nil
+	}
+
+	kbucfgConfig, err := ucfg.NewFrom(kb.Spec.Config.Data)
```
barkbay, line 299: I think that one downside of that approach is that we are reading the _expected_ config, not necessarily the one currently applied in the Kibana config Secret.

kvalliyurnatt: That is a great point ! We probably should look at the secret which would be the current applied config. I can make that change

pkg/controller/agent/fleet.go
kvalliyurnatt: @barkbay this gets called inside the Fleet reconcile loop, and on every reconcile we will end up checking the basePath and updating if needed, in this case do we need a watch on Kibana resources ? would agent be using the API outside of the reconciliation loop, just trying to understand the workflow since I am not familiar with this code.

barkbay: 🤦 , you're right, sorry for the noise.

pkg/controller/agent/fleet.go
naemono: @barkbay but isn't the source of truth the kibana crd? @kvalliyurnatt and I were discussing and I'm not seeing much benefit of reading the secret vs the crd. Don't we treat the crd as the source of truth in nearly all of the controller, as we're talking eventual consistency here, right? Are there other places in the controller where we treat the secret as what should be applied across the stack? I'm trying to ensure we are consistent in how we handle this type of thing. Is there a situation you had in mind where one is better than the other?

barkbay: > @barkbay but isn't the source of truth the kibana crd

I consider the custom resources as the _desired_ state. The _current_ state can be different. It can be a temporary difference, the controller is making some progress and is implementing the desired state, or the controller can be "stuck" (bug in the controller, issue reaching the k8s api, problem in the Kibana resource itself..).

> [...] I'm not seeing much benefit of reading the secret

If the Kibana controller can't reconcile the configuration we may use an incorrect view of the configuration, and then we may also fail to reconcile the Agent resource. As mention before it can be a temporary failure (Agent controller is faster than Kibana controller), or it can a permanent one (bug in the Kibana controller, error in the Kibana resource...). In all the cases it can be confusing to see Agent controller errors while the root cause could be the Kibana controller, this is the kind of situation I wanted to avoid by challenging the proposed implementation.

kvalliyurnatt: I am not strongly leaning towards one implementation over the other, I have made the changes to read form secret, but @naemono  let me know if you feel strongly  to go with reading from the CR

I think this point below makes me lean more towards reading from the secret.
>  In all the cases it can be confusing to see Agent controller errors while the root cause could be the Kibana controller, this is the kind of situation I wanted to avoid by challenging the proposed implementation.

pkg/controller/agent/fleet.go
```diff
@@ -265,6 +285,34 @@ func maybeReconcileFleetEnrollment(params Params, result *reconciler.Results) En
...
+	err := client.Get(ctx, kibanaNSN, &kb)
+	if err != nil {
+		return "", err
+	}
+
+	if kb.Spec.Config == nil {
+		return "", nil
+	}
+
+	// Get Kibana Config secret
```
naemono, line 299:
```suggestion
	// Get Kibana config secret to extract the basepath. We are not using the Kibana CRD here for the basepath to optimize for the case where the desired and current state may differ, so we're choosing the current state to minimize any transient errors.
```

pkg/controller/agent/fleet.go
```diff
@@ -265,6 +285,34 @@ func maybeReconcileFleetEnrollment(params Params, result *reconciler.Results) En
...
+	kbSecretName := kibana.SecretName(kb)
+	var kbConfigsecret corev1.Secret
+	err = client.Get(ctx, types.NamespacedName{Namespace: kb.Namespace, Name: kbSecretName}, &kbConfigsecret)
+	if err != nil {
+		return "", err
+	}
```
naemono, line 305: Not asking for a change, but wanting to start the discussion on whether we want to handle `not found` any differently here, or in the calling block?

pkg/controller/agent/fleet.go
```diff
@@ -265,6 +285,34 @@ func maybeReconcileFleetEnrollment(params Params, result *reconciler.Results) En
...
+
+	kbCfg := kibanaConfig{}
+	err = yaml.Unmarshal(kbConfigsecret.Data[kibana.SettingsFilename], &kbCfg)
+	if err != nil {
+		return "", err
+	}
```
naemono, line 311: This, and the 2 error blocks preceding: I'd suggest we wrap the errors with context of some sort to identify their source/reason.

naemono (approved):
> I think this point below makes me lean more towards reading from the secret.

I don't disagree, but I wanted to have the discussion, and ensure that we are noting why we are doing so in the code, as it feels to me a slight (but not wrong) diversion for our SOP.

I'm good with the current approach, but do have some additional optimizations that are nice to have.

pkg/controller/agent/fleet.go
kvalliyurnatt: I am thinking, if the secret is not found as well we want to error out and reconcile again and there shouldn't be where the secret doesn't exist right ?

naemono: I guess what I was asking is if it's preferable to do something like this in the calling func to minimize noise in the logs:
```diff
diff --git a/pkg/controller/agent/fleet.go b/pkg/controller/agent/fleet.go
index 63cee6cc4..4f61d7ef2 100644
--- a/pkg/controller/agent/fleet.go
+++ b/pkg/controller/agent/fleet.go
@@ -19,6 +19,7 @@ import (
        "go.elastic.co/apm/module/apmhttp/v2"
        "gopkg.in/yaml.v3"
        corev1 "k8s.io/api/core/v1"
+       apierrors "k8s.io/apimachinery/pkg/api/errors"
        "k8s.io/apimachinery/pkg/types"
        "k8s.io/client-go/tools/record"
        "sigs.k8s.io/controller-runtime/pkg/reconcile"
@@ -266,7 +267,10 @@ func maybeReconcileFleetEnrollment(params Params, result *reconciler.Results) En
        }

        kibanaBasePath, err := getKibanaBasePath(params.Context, params.Client, params.Agent.Spec.KibanaRef.WithDefaultNamespace(params.Agent.Namespace).NamespacedName())
-       if err != nil {
+       if err != nil && apierrors.IsNotFound(err) {
+               result.WithResult(reconcile.Result{Requeue: true})
+               return EnrollmentAPIKey{}
+       } else if err != nil {
                result.WithError(err)
                return EnrollmentAPIKey{}
        }
```

kvalliyurnatt: aah I see what you mean, that makes sense.

kvalliyurnatt: done

kvalliyurnatt:
I tested it locally and I was able to verify that fleet tries to use the configured basepath
```
 "Post \"https://kibana-kb-http.kvalliy.svc:5601/monitoring/kibana/api/fleet/setup\": x509: certificate signed by unknown authority (possibly because of \"crypto/rsa: verification error\" while trying to verify candidate authority certificate \"kibana-http\")"}]
 ```
 I had some certificate issues locally, but I was able to check that the Kibana url included the basepath

**Review by pebrc — CHANGES_REQUESTED** [review]
I think the approach needs to be a bit more general as this PR fixes only the Fleet setup problem (which is the one where this issue was reported)

My intuition would be to address this on the level of the association configuration where we convey the Kibana URL to the associated objects:

https://github.com/elastic/cloud-on-k8s/blob/49e40caf6a75059cb17b84609ff151d1d0f215c1/pkg/controller/association/controller/apm_kibana.go#L58-L73

Other associated applications like Beats and APM in older versions also talk to the Kibana API to install dashboards IIRC and that functionality would be broken with a base path that is non-empty as well. 

Finally I think we should try to configure the readiness probe for the Kibana pods to respect the basePath setting: 

https://github.com/elastic/cloud-on-k8s/blob/49e40caf6a75059cb17b84609ff151d1d0f215c1/pkg/controller/kibana/pod.go#L56-L75

In the orginal issue I said we should only look at the Kibana configuration but I wonder if it would be much more effort to cover the case where users have set up the basePath via an environment variable (not recommended but 🤷 )

**Comment by kvalliyurnatt** [comment]
> I think the approach needs to be a bit more general as this PR fixes only the Fleet setup problem (which is the one where this issue was reported tbf)
> 
> My intuition would be to address this on the level of the association configuration where we convey the Kibana URL to the associated objects:
> 
> https://github.com/elastic/cloud-on-k8s/blob/49e40caf6a75059cb17b84609ff151d1d0f215c1/pkg/controller/association/controller/apm_kibana.go#L58-L73
> 
> Other associated applications like Beats and APM in older versions also talk to the Kibana API to install dashboards IIRC and that functionality would be broken with a base path that is non-empty as well.
> 
> Finally I think we should try to configure the readiness probe for the Kibana pods to respect the basePath setting:
> 
> https://github.com/elastic/cloud-on-k8s/blob/49e40caf6a75059cb17b84609ff151d1d0f215c1/pkg/controller/kibana/pod.go#L56-L75
> 
> In the orginal issue I said we should only look at the Kibana configuration but I wonder if it would be much more effort to cover the case where users have set up the basePath via an environment variable (not recommended but 🤷 )

@pebrc I have made the requested changes, please take another look when you get a chance

File: pkg/controller/association/controller/apm_kibana.go [hunk 8]
```diff
@@ -69,7 +81,58 @@ func getKibanaExternalURL(c k8s.Client, assoc commonv1.Association) (string, err
...
+	if kbBasePath != "" {
+		return kbBasePath, nil
+	}
+
+	if kb.Spec.Config == nil {
+		return "", nil
+	}
+
+	// Get Kibana config secret to extract the basepath.
+	// We are not using the Kibana CRD here for the basepath to optimize for the case where the desired and current state may differ, so we're choosing the current state to minimize any transient errors.
```
pebrc, line 109: I my testing I still see a fair amount of reconciliation errors with this approach. I believe this is due to the configuration being generated almost at the same time as the extraction of the association configuration, while you still might have old Kibana pods running which won't re-read their configuration. 

This makes me think that we should just go with what is in the Kibana custom resource. This would also address another issue I have with the current approach which is that we duplicate the implementation between the association configuration and the readiness probe. I would suggest to just use one implemenation shared between the two use cases (I think it should be possible to import the one from the Kibana controller here, which is based on the CR as the source of truth)

pkg/controller/kibana/pod.go
```diff
@@ -115,6 +129,51 @@ func GetKibanaContainer(podSpec corev1.PodSpec) *corev1.Container {
...
+
+	if kbCfg.Server.BasePath != "" {
+		// We give preference to base path set in the spec
+		return kbCfg.Server.BasePath, nil
+	}
+
+	// Check for a flattened structure
+	if kbucfgConfig.HasField("server.basePath") {
+		return kbucfgConfig.String("server.basePath", -1)
+	}
```
pebrc, line 176: This is unnecessary if you either re-use `CanoncialConfig` or alternatively pass the `PathSep(".")` option

```golang
var Options = []ucfg.Option{ucfg.PathSep("."), ucfg.AppendValues}
```
when parsing the config.

pkg/controller/kibana/pod.go
```diff
@@ -115,6 +129,51 @@ func GetKibanaContainer(podSpec corev1.PodSpec) *corev1.Container {
...
+		}
+	}
+	return ""
+}
+
 func getDefaultContainerPorts(kb kbv1.Kibana) []corev1.ContainerPort {
 	return []corev1.ContainerPort{{Name: kb.Spec.HTTP.Protocol(), ContainerPort: int32(network.HTTPPort), Protocol: corev1.ProtocolTCP}}
 }
+
+func getKibanaBasePath(kb kbv1.Kibana) (string, error) {
```
pebrc, line 149: In my testing and based on [the Kibana docs](https://www.elastic.co/guide/en/kibana/current/settings.html#server-rewriteBasePath) the basePath setting is only in effect and needs to be handled by ECK when combined with the `server.rewriteBasePath: true` option.

pebrc (commented):
I think we need to look at `server.rewriteBasePath` as well. Let me know if I missed something important here, but otherwise Kibana just ignores the basePath. 

Also I wonder if would make sense to either modify an existing e2e test or add a new one to exercise this functionality.

pkg/controller/kibana/pod.go
kvalliyurnatt: Thanks, I missed this completely, but I encountered multiple redirect issues when I tried adding based path without this config in my testing, will add a check for this

pkg/controller/association/controller/apm_kibana.go
kvalliyurnatt: I changed it to use a single method for both

pkg/controller/kibana/pod.go
kvalliyurnatt: Thanks, I made this change.

pkg/controller/kibana/pod.go
kvalliyurnatt: I made the change to look at rewriteBasePath

kvalliyurnatt:
> I think we need to look at `server.rewriteBasePath` as well. Let me know if I missed something important here, but otherwise Kibana just ignores the basePath.
> 
> Also I wonder if would make sense to either modify an existing e2e test or add a new one to exercise this functionality.

I modified the apm association test to test the Kibana basepath functionality

test/e2e/apm/association_test.go
```diff
@@ -85,6 +85,92 @@ func TestAPMKibanaAssociation(t *testing.T) {
...
+		WithKibanaRef(kbBuilder.Ref()).
+		WithNodeCount(1).
+		WithRestrictedSecurityContext().
+		WithConfig(map[string]interface{}{
+			"apm-server.ilm.enabled":                           false,
+			"setup.template.settings.index.number_of_replicas": 0, // avoid ES yellow state on a 1 node ES cluster
+		})
+
+	test.Sequence(nil, test.EmptySteps, esBuilder, kbBuilder, apmBuilder).RunSequential(t)
+}
```
pebrc, line 172: This is fine. I think we shoudl try to keep e2e execution times low though. I wonder if it would be quicker to use https://github.com/elastic/cloud-on-k8s/blob/e74296fa4cb39f58280d25c8f0bde6ed1404b0c1/test/e2e/test/run_mutation.go#L14

So basically have just one test and change the Kibana to use env vars, which might be faster than recreating the whole stack twice. Something along the lines of (untested!): 

```golang
	kbWithEnv := kbBuilder.DeepCopy().WithEnv([]corev1.EnvVar{
		{
			Name:  "SERVER_BASEPATH",
			Value: "/monitoring/kibana",
		},
		{
			Name:  "SERVER_REWRITEBASEPATH",
			Value: "true",
		},
	}).WithConfig(nil)

	test.RunMutations(t, []test.Builder{esBuilder, kbBuilder, apmBuilder}, []test.Builder{kbWithEnv})
```

pebrc (commented):
LGTM! a few nits and a finding around truthy values in Kibana conf. Otherwise good to go imo.

pebrc:
buildkite test this -f p=gke,t=TestAPMKibanaAssociation

pkg/controller/kibana/pod.go
```diff
@@ -115,6 +132,56 @@ func GetKibanaContainer(podSpec corev1.PodSpec) *corev1.Container {
...
+	envMap := make(map[string]string)
+	for _, envVar := range kbContainer.Env {
+		if envVar.Name == KibanaBasePathENVName || envVar.Name == KibanaRewriteBasePathENVName {
+			envMap[envVar.Name] = envVar.Value
+		}
+	}
+
+	// If SERVER_REWRITEBASEPATH is set to true, we should use the value of SERVER_BASEPATH
+	if rewriteBasePath, ok := envMap[KibanaRewriteBasePathENVName]; ok {
+		if rewriteBasePath == "true" {
```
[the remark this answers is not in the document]
kvalliyurnatt: Seems like both Cases work so I change this to be a ParseBool

test/e2e/apm/association_test.go
kvalliyurnatt: Thanks, I was not aware of this. I have made this change, this makes it much quicker and simpler

kvalliyurnatt:
buildkite test this -f p=gke,t=TestAPMKibanaAssociation

pebrc (approved):
LGTM nice work!

Issue #6617, opened 2023-03-29
Do not skip CI for .asciidoc at the PR bot level


thbkrkr:
I didn't think deeply when setting the Buildkite PR bot in #6121.

https://github.com/elastic/cloud-on-k8s/blob/d8704ec52a7c82a5c0606fc3068a0c43359da13b/.buildkite/pull-requests.json#L14

this causes no build to be triggered for a PR that just updates the documentation and therefore it waits for the GitHub status check:
- `buildkite/cloud-on-k8s-operator Expected — Waiting for status to be reported`.

I looked at what we were doing in Jenkins ([pr-gke.Jenkinsfile](https://github.com/elastic/cloud-on-k8s/blob/e440baef28c3c7d9ace857f5e7d86818161bae0c/.ci/pipelines/pr-gke.Jenkinsfile)). We triggered the build but by deactivating absolutely all the steps when it was only about docs.

Neither seems correct to me. I don't see an option of having a label-conditioned GitHub status check.
I have no better idea than to replicate what we were doing on Jenkins for now.

thbkrkr:
I think that with the new independent? `docs-build-pr` build, we can leverage `"skip_ci_on_only_changed": ["\\.asciidoc$"] `.

PR #8067, opened 2024-09-27
Skip CI if only ^docs/ changed

thbkrkr:
This is an attempt to not run the main CI pipeline when there are only documentation changes.

Compared to the last time I tried this, the docs build has been completely revised.

Relates to #6617.


pebrc (approved):
Let's try it.

PR #8075, opened 2024-09-30
fix(deploy): move '$leading_path' variable definition in eck-operator

devzizu:
## Description

This PR fixes an error that is thrown by Helm when templating the `eck-operator` chart. In this case, consider the following variables file:

```bash
config:
  metrics:
    port: "6000"
    secureMode:
      enabled: true
      tls:
        certificateSecret: "my-metrics-cert-secret"
        caSecret: "my-metrics-cert-secret"
        insecureSkipVerify: false
```

Running `helm template` with the above values generates the following error:

```bash
Error: template: eck-operator/templates/serviceMonitor.yaml:25:49: executing "eck-operator/templates/serviceMonitor.yaml" at <.Values.config.metrics.secureMode.tls.caMountDirectory>: can't evaluate field Values in type string

Use --debug flag to render out invalid YAML
```

This error occurs because when modifying the scope using `with`, the reference for `.Values`, for example, is lost ([helm docs](https://helm.sh/docs/chart_template_guide/control_structures/#modifying-scope-using-with)).

Also, unit tests pass with `make helm-test` (using Helm `v3.14` and unittest `v0.2.8`).

## PR requirements

- [x] Have you signed the [contributor license agreement](https://www.elastic.co/contributor-agreement)?
- [x] Have you followed the [contributor guidelines](https://github.com/elastic/cloud-on-k8s/tree/main/CONTRIBUTING.md)?
- [x] If you submit code, is your pull request against main? We recommend pull requests against main. We will backport them as needed.

### DIFF
--- a/deploy/eck-operator/templates/serviceMonitor.yaml
+++ b/deploy/eck-operator/templates/serviceMonitor.yaml
@@ -21,8 +21,8 @@ spec:
     tlsConfig:
       insecureSkipVerify: {{ .Values.config.metrics.secureMode.tls.insecureSkipVerify | default false }}
       {{- if (not .Values.config.metrics.secureMode.tls.insecureSkipVerify) }}
-      {{- with .Values.config.metrics.secureMode.tls.caSecret }}
       {{- $leading_path := trimSuffix "/" .Values.config.metrics.secureMode.tls.caMountDirectory }}
+      {{- with .Values.config.metrics.secureMode.tls.caSecret }}
       caFile: "{{ $leading_path }}/{{ . }}/ca.crt"
       {{- end }}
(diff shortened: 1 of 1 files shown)


thbkrkr:
buildkite test this

thbkrkr:
Thank you!

thbkrkr:
buildkite test this

thbkrkr:
> @thbkrkr Thanks for the review! Can we backport this to prior versions of the chart as well?

Thank you for the suggestion! We generally release prior versions only for critical bug fixes, so we might not backport this change to earlier versions unless it is deemed necessary.

thbkrkr (approved):

PR #8087, opened 2024-10-08
Fix logstash templating issue.

### LINKED ISSUE #8000: Incorrect templating of Logstash object when using pipelinesRef parameter
## Bug Report
**What did you do?**
We wanted to define our pipeline via secret, so we have configured pipelinesRef in values:
```yaml
eck-logstash:
    enabled: true
    pipelinesRef:
      secretName: logstash-pipelines-secret
```

We expected helmchart to be deployed successfully and see the Logstash object containing only _pipelinesRef_ parameter.

Templated Logstash object contains both _pipelines_ and _pipelinesRef_ parameters:
```yaml
kind: Logstash
metadata:
  name: release-name-eck-logstash
  labels:
    helm.sh/chart: eck-logstash-0.12.0
    app.kubernetes.io/name: eck-logstash
    app.kubernetes.io/instance: release-name
  annotations:
    eck.k8s.elastic.co/license: basic
spec:
  version: 8.15.0
  count: 1
  pipelinesRef:
    secretName: logstash-pipelines-secret

  pipelines:
    []
  volumeClaimTemplates:
    - metadata:
        name: logstash-data
      spec:
        accessModes:
        - ReadWriteOnce
        resources:
          requests:
            storage: 2Gi
        storageClassName: storage-del
  elasticsearchRefs:
    []
```
Helmchart cannot be deployed since validating webhook returns this error:
```
one or more synchronization tasks completed unsuccessfully, reason: admission webhook "elastic-logstash-validation-v1alpha1.k8s.elastic.co" denied the request: Logstash.logstash.k8s.elastic.co "ecn-monitoring-eck-logstash" is invalid: [spec.pipelines: Forbidden: Specify at most one of [`pipelines`, `pipelinesRef`], not both, spec.pipelinesRef: Forbidden: Specify at most one of [`pipelines`, `pipelinesRef`], not both] (retried 5 times).
```

**Environment**

* ECK version:

	v2.14.0

naemono:
Fixes #8000 

This change fixes a bug in the helm templating of Logstash and does not add `.spec.pipelines` when `.spec.pipelinesRef` is also defined. In addition this adds a test to ensure that the field `pipelines` is missing in `spec` when `pipelinesRef` is also defined.

### DIFF
--- a/deploy/eck-stack/charts/eck-logstash/templates/logstash.yaml
+++ b/deploy/eck-stack/charts/eck-logstash/templates/logstash.yaml
@@ -23,16 +23,15 @@ spec:
   revisionHistoryLimit: {{ . }}
   {{- end }}
 
+  {{- if and .Values.config .Values.configRef }}
+  {{- fail "config and configRef are mutually exclusive!" }}
+  {{- end }}
   {{- with .Values.config }}
   config:
     {{- toYaml . | nindent 4 }}
   {{- end }}
   {{- with .Values.configRef }}
   configRef:
-    {{- toYaml . | nindent 4 }}
-  {{- end }}
-  {{- with .Values.pipelinesRef }}
-  pipelinesRef:
     {{- toYaml . | nindent 4 }}
   {{- end }}
   {{- with .Values.podTemplate }}
@@ -43,8 +42,14 @@ spec:
   monitoring:
     {{- toYaml . | nindent 4 }}
   {{- end }}
-
-  {{- if not (.Values.pipelinesRef)  }}
+  {{- if and .Values.pipelines .Values.pipelinesRef }}
+  {{- fail "pipelines and pipelinesRef are mutually exclusive!" }}
+  {{- end }}
+  {{- with .Values.pipelinesRef }}
+  pipelinesRef:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+  {{- if .Values.pipelines  }}
   pipelines: {{ toYaml .Values.pipelines | nindent 4 }}
   {{- end }}
   volumeClaimTemplates: {{ toYaml .Values.volumeClaimTemplates | nindent 4 }}
--- a/deploy/eck-stack/charts/eck-logstash/templates/logstash.yaml
+++ b/deploy/eck-stack/charts/eck-logstash/templates/logstash.yaml
@@ -44,7 +44,9 @@ spec:
     {{- toYaml . | nindent 4 }}
   {{- end }}
 
+  {{- if not (.Values.pipelinesRef)  }}
   pipelines: {{ toYaml .Values.pipelines | nindent 4 }}
+  {{- end }}
   volumeClaimTemplates: {{ toYaml .Values.volumeClaimTemplates | nindent 4 }}
   elasticsearchRefs: {{ toYaml .Values.elasticsearchRefs | nindent 4 }}
   services: {{ toYaml .Values.services | nindent 4 }}
--- a/deploy/eck-stack/charts/eck-logstash/templates/tests/logstash_test.yaml
+++ b/deploy/eck-stack/charts/eck-logstash/templates/tests/logstash_test.yaml
@@ -157,11 +157,19 @@ tests:
       - equal:
           path: spec.configRef.secretRef.secretName
           value: configLogstashSecretName
-  - it: should render pipelinesRef properly
+  - it: should render pipelinesRef properly without pipelines field
     set:
       pipelinesRef:
         secretName: pipelineLogstashSecretName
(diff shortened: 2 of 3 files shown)


deploy/eck-stack/charts/eck-logstash/templates/logstash.yaml
```diff
@@ -44,7 +44,9 @@ spec:
     {{- toYaml . | nindent 4 }}
   {{- end }}
 
+  {{- if not (.Values.pipelinesRef)  }}
   pipelines: {{ toYaml .Values.pipelines | nindent 4 }}
+  {{- end }}
```
thbkrkr, line 49: Imagine if the user sets both `pipelines` and `pipelinesRef` fields, then only `pipelinesRef` will be rendered without error. What do you think about adding a validation to prevent setting both fields?
Nit: also group pipeline related bits together.
```diff
diff --git a/deploy/eck-stack/charts/eck-logstash/templates/logstash.yaml b/deploy/eck-stack/charts/eck-logstash/templates/logstash.yaml
index 3c4deb1bc..162ec317a 100644
--- a/deploy/eck-stack/charts/eck-logstash/templates/logstash.yaml
+++ b/deploy/eck-stack/charts/eck-logstash/templates/logstash.yaml
@@ -31,10 +31,6 @@ spec:
   configRef:
     {{- toYaml . | nindent 4 }}
   {{- end }}
-  {{- with .Values.pipelinesRef }}
-  pipelinesRef:
-    {{- toYaml . | nindent 4 }}
-  {{- end }}
   {{- with .Values.podTemplate }}
   podTemplate:
     {{- toYaml . | nindent 4 }}
@@ -43,8 +39,14 @@ spec:
   monitoring:
     {{- toYaml . | nindent 4 }}
   {{- end }}
-
-  {{- if not (.Values.pipelinesRef)  }}
+  {{- if and .Values.pipelines .Values.pipelinesRef }}
+  {{- fail "pipelines and pipelinesRef are mutually exclusive!" }}
+  {{- end }}
+  {{- with .Values.pipelinesRef }}
+  pipelinesRef:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+  {{- if .Values.pipelines  }}
   pipelines: {{ toYaml .Values.pipelines | nindent 4 }}
   {{- end }}
   volumeClaimTemplates: {{ toYaml .Values.volumeClaimTemplates | nindent 4 }}
diff --git a/deploy/eck-stack/charts/eck-logstash/values.yaml b/deploy/eck-stack/charts/eck-logstash/values.yaml
index a2d1d289a..e68694852 100644
--- a/deploy/eck-stack/charts/eck-logstash/values.yaml
+++ b/deploy/eck-stack/charts/eck-logstash/values.yaml
@@ -78,11 +78,11 @@ monitoring: {}
   #     namespace: observability
 
 # The Logstash pipelines, the ECK equivalent to pipelines.yml
+# NOTE: The `pipelines` and `pipelinesRef` fields are mutually exclusive. Only one of them should be defined at a time, as using both may cause conflicts.
 # ref: https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-logstash-configuration.html#k8s-logstash-pipelines
 #
 pipelines: []
-
-pipelinesRef: {}
+# pipelinesRef:
 #  secretRef:
 #    secretName: ''

```

naemono: Thanks @thbkrkr , this makes a lot of sense. Upon further looking at the documentation in the values file, I'm noticing some other issues I'm going to double check/fix:

This doesn't appear to be correct, according to the documentation:
```
-pipelinesRef: {}
-#  secretRef:
-#    secretName: ''
```

It should be:
```
# pipelinesRef:
#   secretName: ''
```

Also, do we have the same situation for `config` and `configRef`? I believe we do. I'm going to double check that as well, as those seem to be mutually exclusive also.

thbkrkr: Well seen, yes `pipelinesRef. secretRef` is not valid

and `config` / `configRef` are also mutually exclusive. 

https://github.com/elastic/cloud-on-k8s/blob/e415ac01e2526fc0cb38f78df69e0a66e7df597a/pkg/controller/logstash/validation/validations.go#L75

naemono: Ok @thbkrkr , the values file has been updated, the template will now fail on this as well, and there are 2x tests to ensure this functionality. Ready for next set of :eyes: when you can. Thanks!

thbkrkr (approved):
Good improvements! ✨

Issue #7818, opened 2024-05-16
Support remote clusters using API keys


pebrc:
Elasticsearch starting with version 8.10 supports a new security model for remote cluster connections using API keys (currently in beta): 
https://www.elastic.co/guide/en/elasticsearch/reference/current/remote-clusters-api-key.html

We should support this in ECK for the fully managed remote cluster connection setup documented here: 

https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-remote-clusters.html#k8s-remote-clusters-connect-internal

* dedicated API key API /_security/cross_cluster/api_key
* CA trust is required  (as before) 
* additional port for the remote connections (default 9443)

There are few open questions: 
* creating the API key requires user input to tailor the permissions to the specific use case. Automating a wildcard api key that allows access to everything would be giving a way the main benefit of the new security model

* our current API does not allow to specify the permissions
```yaml
apiVersion: elasticsearch.k8s.elastic.co/v1
kind: Elasticsearch
metadata:
  name: cluster-one
  namespace: ns-one
spec:
  nodeSets:
  - count: 3
    name: default
  remoteClusters:
  - name: cluster-two
    elasticsearchRef:
      name: cluster-two
      namespace: ns-two 
  version: 8.13.4
```

But we would need to express the cross-cluster API key permission model somehow:

```
POST /_security/cross_cluster/api_key
{
  "name": "my-cross-cluster-api-key",
  "expiration": "1d",   
  "access": {
    "search": [  
      {
        "names": ["logs*"]
      }
    ],
    "replication": [  
      {
        "names": ["archive*"]
      }
    ]
  },
  "metadata": {
    "description": "phase one",
    "environment": {
       "level": 1,
       "trusted": true,
       "tags": ["dev", "staging"]
    }
  }
}
```

Do we want to inline the `access` section into our CRD? Do we want to leave this as a step to be executed by the user (not very attractive if we want to facilitate automatic setup of these remote cluster relationships) Can we leverage stack config policies in a novel way to do this. 

Another question is how migration would work (if we support that at all from old world to new world). The stack documentation is to be found here: https://www.elastic.co/guide/en/elasticsearch/reference/current/remote-clusters-migrate.html

* do we want to try to fully automated this? 
* do we want to not automate but allow users to manually transition 
* do we do nothing and let both mechanisms exist next to each other?

PR #8089, opened 2024-10-09
Support remote clusters using API keys

barkbay:
This PRs allows to define API keys for remote cluster connections.

I'm creating this PR as a draft to get a first feedback on the overall architecture and how the API keys are managed.

The examples in this comment refer to the recipe available in [`config/recipes/remoteclusters/elasticsearch.yaml`](https://github.com/elastic/cloud-on-k8s/pull/8089/files#diff-a4022373d5ddf04ab075c5a9d98fb24a7ba1cceff3b57c775da0f67d45b80942).

## No new controller, ~`remoteca`~  ➡️   `remotecluster-controller`

The API key reconciliation has been included in the existing `remoteca` controller that was taking care of the certificate management.
It has been therefore renamed `remotecluster-controller` and is now also handling API keys reconciliation in both:
* The Elasticsearch API of the remote cluster
* The secure settings of the client cluster

### In the remote Elasticsearch cluster

API keys created in the remote cluster by the operator follow the following naming convention: `eck-<client_cluster_ns>-<client_cluster_name>-<remote_cluster_name>`.
Some metadata, like the client cluster name or the hash of the expected key specification, are attached to the API keys in order to facilitate their conciliation:

```json
{
  "api_keys": [
    {
      "id": "-yvMb5IBPP7lgi3Dtn2m",
      "name": "eck-ns1-cluster1-to-ns2-cluster2",
      "metadata": {
        "elasticsearch.k8s.elastic.co/config-hash": "660014403",
        "elasticsearch.k8s.elastic.co/name": "cluster1",
        "elasticsearch.k8s.elastic.co/namespace": "ns1",
        "elasticsearch.k8s.elastic.co/uid": "831b535e-c031-4b4d-aba5-3ad30271f5cc",
        "elasticsearch.k8s.elastic.co/managed-by": "eck"
      },
      "access": {
        "search": [
          {
            "names": [
              "kibana_sample_data_ecommerce"
            ]
          }
        ]
      }
    }
  ]
}
```

### In the client Elasticsearch cluster

Encoded keys used by the client cluster to authenticate against the remote one are stored in a new dedicated `Secret` on the client cluster (this `Secret` is handled using the `APIKeyStore` in `pkg/controller/remotecluster/keystore.go`):

```shell
> kubectl get secret -l common.k8s.elastic.co/type=remote-cluster-api-keys -A
NAMESPACE   NAME                          TYPE     DATA   AGE
ns1         cluster1-es-remote-api-keys   Opaque   1      13m
```

We are tracking the API key IDs in an annotation, this is used to detect when a key has been externally invalidated:

```shell
>kubectl get secret cluster1-es-remote-api-keys -n ns1 -o=jsonpath='{.metadata.annotations.elasticsearch\.k8s\.elastic\.co\/remote-cluster-api-keys}'  | jq
```

```json
{
  "to-ns2-cluster2": {
    "namespace": "ns2",
    "name": "cluster2",
    "id": "-yvMb5IBPP7lgi3Dtn2m"
  }
}
```

This Secret is:
*  Loaded as part of the existing secure setting mechanism (which triggers a restart when updated, refer to section "To be discussed in dedicated issues/prs").
* **Not** created until there are some keys to be stored.

## Is it possible to go back to the "legacy" remote cluster mode once API keys are enabled?

Yes, you just have to delete the api key from the custom resource. However, this may result in downtime.

## Todo

There are still a few things I'm not happy with. For example I'm wondering if we need an expectation mechanism for the Secrets that hold the encoded key: if a key is created and the Secret is created but not observed in the next reconciliation we may invalidate the key that has just been created.

Other things to complete before merging or going ga:

- [ ] Documentation
- [x] E2E Tests
- [x] Evaluate how this works with custom certificates on the transport layer, if there is any new requirement.

## To be discussed in dedicated issues/prs

- Avoid the restart of the whole cluster when a new API key is generated.

## Testing

As previously mentioned there is an example in [`config/recipes/remoteclusters/elasticsearch.yaml`](https://github.com/elastic/cloud-on-k8s/pull/8089/files#diff-a4022373d5ddf04ab075c5a9d98fb24a7ba1cceff3b57c775da0f67d45b80942).

Here are a few API calls that you may find useful while testing this PR:

* `GET /_remote/info` to get the current remote cluster status. Note that the `connected` field may not be immedately refreshed/updated)
* `GET /_security/api_key?active_only=true&name=eck-*` to get the active keys created by the operator.
* `GET /to-ns2-cluster2:kibana_sample_data_ecommerce/_search` , when using the manifest in the recipe and run on `ns1/cluster1` this should return the data from `ns2/cluster2`

Fixes https://github.com/elastic/cloud-on-k8s/issues/7818


barkbay:
Please do not spend some time on the case where node transport certificates are provided by a third-party tool (https://www.elastic.co/guide/en/cloud-on-k8s/master/k8s-transport-settings.html#k8s-transport-third-party-tools). This needs some adjustments, I'm working on it.

barkbay:
buildkite test this -f p=gke,t=TestRemoteClusterWithAPIKeys -m s=8.9.2,s=8.15.2,s=8.16.0-SNAPSHOT

pkg/controller/remotecluster/apikey.go
```diff
@@ -0,0 +1,190 @@
...
+	aliases := clientClusterAPIKeyStore.ForCluster(reconciledES.Namespace, reconciledES.Name)
+	for existingAlias := range aliases {
+		if expectedAliases.Has(existingAlias) {
+			continue
+		}
+		clientClusterAPIKeyStore.Delete(existingAlias)
+	}
+
+	// Save the generated keys in the keystore.
+	if err := clientClusterAPIKeyStore.Save(ctx, c, clientES); err != nil {
```
pebrc, line 108: If I understand correctly we reconcile from the server's perspective which means if a cluster is client to multiple remotes multiple threads of reconciliation will try to update the same secret? Potentially at the same time depending on the parallelism the controller is configured with? Do you think this will be problem in practice? Did you consider doing it the other way round?

pkg/controller/remotecluster/keystore.go
```diff
@@ -0,0 +1,232 @@
...
+		Name:      esv1.RemoteAPIKeysSecretName(owner.Name),
+		Namespace: owner.Namespace,
+	}
+	// Attempt to read the Secret
+	keyStoreSecret := &corev1.Secret{}
+	if err := c.Get(ctx, secretName, keyStoreSecret); err != nil {
+		if errors.IsNotFound(err) {
+			ulog.FromContext(ctx).V(1).Info("No APIKeyStore Secret found",
+				"namespace", owner.Namespace,
+				"es_name", owner.Name,
```
pebrc, line 79: I think this leads to duplicate `namespace` and `es_name` keys as those are already populated with the reconciled objects values.

pebrc (commented):
I spent a bit of time today testing your PR and it is really impressive how well it worked. 🚀   I am not sure if an expectation mechanism on the secrets is need IIUC the worst case is we re-create an API key? 

What I found slightly difficult to reason about is the reconciliation in the remote cluster controller. I know that this is to large parts existing code into which the new functionality was inserted. But maybe we can do something with the naming. IIUC the local cluster is always the remote server and I think this threw me off because in my head local meant the client that makes calls to a remote using an API key.

pkg/controller/remotecluster/apikey.go
barkbay: > Did you consider doing it the other way round?

Yes, since we need an Elasticsearch client to reconcile the API keys it seemed more "natural" to me to create the ES client "once", and then reconcile/propagates the API keys for all the client clusters in the same loop, when the remote cluster is reconciled (as opposed to create _n_ Elasticsearch clients for each of the remote clusters to reconcile the API keys). My feeling was that, in case of a conflict on the `Secret`, the consequence would be that we would "just" need to create a new API key during the next iteration, and try to store it again, which seemed acceptable.

In any case one thing we may want to solve is when the API key is not immediately observed in the Secret (because of the client's cache). This would require a kind of "expectation mechanism": the generated keys would be stored in memory until they are persisted. This would prevent the conflict side effect when the `Secret` is created/updated since we would no longer need to generate a new key because we still have it in memory.

pebrc: Makes sense. It is probably computationally more expensive to create many ES clients with certificate pools than the current approach which minimises the number of clients created at the cost of potential conflicts.

pkg/apis/elasticsearch/v1/remote_cluster.go
```diff
@@ -0,0 +1,91 @@
...
+	// +kubebuilder:validation:Optional
+	Replication *Replication `json:"replication,omitempty"`
+}
+
+type Search struct {
+	// +kubebuilder:validation:Required
+	Names []string `json:"names,omitempty"`
+
+	// +kubebuilder:validation:Optional
+	FieldSecurity *FieldSecurity `json:"field_security,omitempty"`
```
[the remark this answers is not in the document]
barkbay: Added in https://github.com/elastic/cloud-on-k8s/pull/8089/commits/b6bf462cde84ad69d211a313d2e7f7ca8f9721c9 and https://github.com/elastic/cloud-on-k8s/pull/8089/commits/322530726199504488eeef9eb4b1a033e0bac97e

barkbay:
I removed the draft status. I still want to work on the documentation and add some additional unit tests, but happy to get additional feedback in the meantime.

pkg/controller/remotecluster/controller.go
```diff
@@ -0,0 +1,449 @@
...
+		}
+		// Create a new client
+		newEsClient, err := r.esClientProvider(ctx, r.Client, r.Dialer, *localEs)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+		// Check that the API is available
+		esClient = newEsClient
+		// Get all the API Keys, for that specific client, on the reconciled cluster.
+		getCrossClusterAPIKeys, err := esClient.GetCrossClusterAPIKeys(ctx, "eck-*")
```
pebrc, line 195: Should we try to handle the case where a new cluster is not reachable yet to avoid noisy error logs like this: 
```
manager.eck-operator	Reconciler error	{"service.version": "2.15.0-SNAPSHOT+65769109", "controller": "remotecluster-controller", "object": {"name":"cluster3","namespace":"ns2"}, "namespace": "ns2", "name": "cluster3", "reconcileID": "8aebea3f-118d-474e-b4c8-4b55efc26141", "error": "elasticsearch client failed for https://cluster3-es-default-0.cluster3-es-default.ns2:9200/_security/api_key?active_only=true&error_trace=true&name=eck-%2A: Get \"https://cluster3-es-default-0.cluster3-es-default.ns2:9200/_security/api_key?active_only=true&error_trace=true&name=eck-%2A\": EOF"}
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).reconcileHandler
	/Users/pebrc/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.19.0/pkg/internal/controller/controller.go:316
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).processNextWorkItem
	/Users/pebrc/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.19.0/pkg/internal/controller/controller.go:263
```

pkg/controller/remotecluster/keystore/keystore.go
```diff
@@ -0,0 +1,293 @@
...
+			Name:      secretName.Name,
+			Namespace: secretName.Namespace,
+			Annotations: map[string]string{
+				aliasesAnnotationName: string(aliases),
+			},
+			Labels: expectedLabels,
+		},
+		Data: data,
+	}
+	if _, err := reconciler.ReconcileSecret(ctx, c, expected, owner); err != nil {
```
pebrc, line 247: I was seeing a few noisy conflicts during testing I assume because `Save` is called in two places during one reconciliation?

```
2024-10-20T10:29:43.425+0200	ERROR	manager.eck-operator	Reconciler error	{"service.version": "2.15.0-SNAPSHOT+65769109", "controller": "remotecluster-controller", "object": {"name":"cluster1","namespace":"ns1"}, "namespace": "ns1", "name": "cluster1", "reconcileID": "316f64b0-454e-4285-8e89-34c18f0da6a2", "error": "Operation cannot be fulfilled on secrets \"cluster1-es-remote-api-keys\": the object has been modified; please apply your changes to the latest version and try again", "errorCauses": [{"error": "Operation cannot be fulfilled on secrets \"cluster1-es-remote-api-keys\": the object has been modified; please apply your changes to the latest version and try again"}]}
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).reconcileHandler
	/Users/pebrc/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.19.0/pkg/internal/controller/controller.go:316
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).processNextWorkItem
	/Users/pebrc/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.19.0/pkg/internal/controller/controller.go:263
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).Start.func2.2
	/Users/pebrc/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.19.0/pkg/internal/controller/controller.go:224
```

pkg/controller/remotecluster/apikey.go
pebrc: One downside of the current approach that occurred to me during testing is that the reconciliation that optimises for creating fewer clients for the remote server clusters is more disruptive to the client clusters as every new remote cluster connection requires a full restart of the cluster to update the keystores. I don't think this a problem in practice, it I can only think of very rare scenarios where this might be observable (adding remote clusters within very short intervals but not at once)

pkg/controller/remotecluster/controller.go
barkbay: This should be the case: https://github.com/elastic/cloud-on-k8s/blob/657691095e762897de05878900ea5a98ed9baa42/pkg/controller/remotecluster/controller.go#L181-L186

Maybe there's a bug, this `EOF` seems a bit odd though 🤔 , I would expect a connection refused or another error, `EOF` feels like the connection was initiated but unexpectedly closed.

barkbay: Could you check if this error comes from our [port-forwarder "hack"](https://github.com/elastic/cloud-on-k8s/tree/b5de7248728cd9c1b81cb71be01d44ac51bd4189/pkg/dev/portforward) used in dev mode please, maybe by looking at the full stacktrace?

pebrc: Yes that was it

pebrc (approved):
Nice work! 🚢 

Small aside: I noticed we don't watch rolebindings in the ES namespaces so when using the RBAC association restrictions we don't re-reconcile if those change. Not sure if there is an easy fix though because users could also use cluster role bindings. Also it has been like this for a very long time and was not introduced in your PR.

pkg/controller/elasticsearch/client/v8.go
```diff
@@ -35,6 +35,28 @@ func (c *clientV8) GetClusterState(ctx context.Context) (ClusterState, error) {
...
+	return response, err
+}
+
+func (c *clientV8) GetCrossClusterAPIKeys(ctx context.Context, name string) (CrossClusterAPIKeyList, error) {
+	var response CrossClusterAPIKeyList
+	err := c.get(ctx, fmt.Sprintf("/_security/api_key?active_only=true&name=%s", name), &response)
+	return response, err
+}
+
+func (c *clientV8) InvalidateCrossClusterAPIKey(ctx context.Context, name string) error {
```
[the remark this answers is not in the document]
barkbay: Because the key is actually not deleted:

> This API invalidates API keys created by the [create API key](https://www.elastic.co/guide/en/elasticsearch/reference/current/security-api-create-api-key.html) or [grant API key](https://www.elastic.co/guide/en/elasticsearch/reference/current/security-api-grant-api-key.html) APIs. Invalidated API keys fail authentication, but they can still be viewed using the [get API key information](https://www.elastic.co/guide/en/elasticsearch/reference/current/security-api-get-api-key.html) and [query API key information](https://www.elastic.co/guide/en/elasticsearch/reference/current/security-api-query-api-key.html) APIs, for at least the [configured retention period](https://www.elastic.co/guide/en/elasticsearch/reference/current/security-settings.html#api-key-service-settings-delete-retention-period), until they are automatically deleted.

https://www.elastic.co/guide/en/elasticsearch/reference/current/security-api-invalidate-api-key.html

pkg/controller/remotecluster/secret.go
```diff
@@ -39,12 +39,12 @@ func createOrUpdateCertificateAuthorities(
 	localClusterKey := k8s.ExtractNamespacedName(local)
 	remoteClusterKey := k8s.ExtractNamespacedName(remote)
 
-	// Add watches on the CA secret of the local cluster.
+	// AddKey watches on the CA secret of the local cluster.
```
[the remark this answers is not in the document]
barkbay: Yes 😐

barkbay:
buildkite test this -f p=gke,t=TestRemoteClusterWithAPIKeys -m s=8.9.2,s=8.15.2,s=8.16.0-SNAPSHOT,s=8.17.0-SNAPSHOT

pkg/controller/elasticsearch/client/v8.go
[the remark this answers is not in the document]
pebrc: TIL

barkbay:
> Small aside: I noticed we don't watch rolebindings in the ES namespaces so when using the RBAC association restrictions we don't re-reconcile if those change. Not sure if there is an easy fix though because users could also use cluster role bindings. Also it has been like this for a very long time and was not introduced in your PR.

Sorry, just realized I forgot to reply to your comment. We are actually reconciling every 15 minutes in the current implementation:

https://github.com/elastic/cloud-on-k8s/blob/ae1f62cd0e74f4ca80887bbcbdac21fcdec498e0/pkg/controller/remotecluster/controller.go#L323

```golang
// RequeueRbacCheck returns a reconcile result depending on the implementation of the AccessReviewer.
// It is mostly used when using the subjectAccessReviewer implementation in which case a next reconcile loop should be
// triggered later to keep the association in sync with the RBAC roles and bindings.
// See https://github.com/elastic/cloud-on-k8s/issues/2468#issuecomment-579157063
func RequeueRbacCheck(accessReviewer rbac.AccessReviewer) reconcile.Result {
	switch accessReviewer.(type) {
	case *rbac.SubjectAccessReviewer:
		return reconcile.Result{RequeueAfter: 15 * time.Minute}
	default:
		return reconcile.Result{}
	}
}
```

pebrc:
> We are actually reconciling every 15 minutes in the current implementation:

I had completely forgotton about that. I did not wait long enough during testing obviously.

PR #8094, opened 2024-10-09
Log when k8s resources are created/updated/deleted successfully

adamkasztenny:
It was not clear from the logs when k8s operations actually completed successfully.

### DIFF
--- a/pkg/controller/common/reconciler/reconciler.go
+++ b/pkg/controller/common/reconciler/reconciler.go
@@ -122,8 +122,7 @@ func ReconcileResource(params Params) error {
 	}
 
 	if params.NeedsRecreate != nil && params.NeedsRecreate() {
-		log.Info("Resource cannot be updated, hence will be deleted and then recreated")
-		log.Info("Deleting resource")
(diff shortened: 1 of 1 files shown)


kvalliyurnatt (approved):

naemono (approved):
also lgtm.

pkg/controller/common/reconciler/reconciler.go
```diff
@@ -119,8 +122,8 @@ func ReconcileResource(params Params) error {
 	}
 
 	if params.NeedsRecreate != nil && params.NeedsRecreate() {
-		log.Info("Resource cannot be updated, hence will be deleted and then recreated", "kind", kind, "namespace", namespace, "name", name)
-		log.Info("Deleting resource", "kind", kind, "namespace", namespace, "name", name)
+		log.Info("Resource cannot be updated, hence will be deleted and then recreated")
+		log.Info("Deleting resource")
```
thbkrkr, line 126: Not from your PR but I think we could combine these two log in one:
```suggestion
		log.Info("Deleting resource as it cannot be updated, it will be recreated")
```

thbkrkr (approved):

Issue #8042, opened 2024-09-04
Readiness probe not working with custom container port


ibalat:
Hi, I have upgraded 2.14 and it uses new readiness script checks 8080 port. But, we use 9200 port for elasticsearch container. How can I customize the script or is there another solution to solve the problem? I had updated readiness-port-script.sh with old readiness-probe-script.sh but it is not working because of missing env. At last status, pod is NotReady.

**readiness-port-script.sh**: `nc -z -v -w5 127.0.0.1 8080`

![Image](https://github.com/user-attachments/assets/69a96170-5bc4-429f-9916-393a7605962a)

Old version had used probe script but it is not working now because of missing environment variables ([#8009](https://github.com/elastic/cloud-on-k8s/pull/8009))
![Image](https://github.com/user-attachments/assets/eaca0707-996f-401e-8c8d-c7f295ac7b1f)

ashuec90:
I am also facing this issue just after disabling the xpack.security.http.ssl.enabled. and the readiness probe is failing with connection error.
```nc: connect to 127.0.0.1 port 8080 (tcp) failed: Connection refused```

**Comment by pebrc** [comment]
@ibalat the container ports are irrelevant for the readiness check. It uses port 8080 on localhost. Don't update the readiness port script, don't override the readiness probe.  See the guidance in our docs https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-readiness.html#k8s_elasticsearch_versions_8_2_0_and_later

@ashuec90 this likely just means that your Elasticsearch pod is not ready. Check the logs.

**Comment by ibalat** [comment]
@pebrc but no exposed 8080 port in container, how it check non-exists port? 8080 check return connection refused

**Comment by Mason0920** [comment]
@pebrc 
[The document](https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-readiness.html#k8s_elasticsearch_versions_8_2_0_and_later) says not to override the readiness probe, but I resolved the issue by overriding it as shown below.

```
        - name: elasticsearch
          readinessProbe:
            exec:
              command:
                - bash
                - -c
                - |
                  nc -z -v -w5 127.0.0.1 9200 #port change
```

pebrc:
@Mason0920 this is not doing what you think it is doing and you are putting the safe operation of the ECK operator at risk with this change. 

[...]

PR #8098, opened 2024-10-12
Do not prepopulate file-based settings with empty objects

pebrc:
A operator `settings.json` even with empty settings for each supported setting requires the appropriate handlers in Elasticsearch to be loaded. 

```
{
  "metadata": {
    "version": "1728032913938346943",
    "compatibility": "8.6.1"
  },
  "state": {
    "cluster_settings": {},
    "snapshot_repositories": {},
    "slm": {},
    "role_mappings": {},
    "ilm": {},
    "ingest_pipelines": {},
    "index_templates": {
      "component_templates": {},
      "composable_index_templates": {}
    }
  }
}
```

This causes problems when Elasticsearch is configured in a way that certain handlers are not available. In https://github.com/elastic/cloud-on-k8s/issues/8042 a user was running with `xpack.security.enabled: false` which meant that the security related handlers for role mappings were not available and the cluster never became ready. 

While we do not support running with security disabled in ECK it still seems appropriate to orchestrate ES in such a way that this problem is avoided. This PR suggests to move the following default file instead: 

```
{
  "metadata": {
    "version": "1728032913938346943",
    "compatibility": "8.6.1"
  },
  "state": {}
}
```

### DIFF
--- a/pkg/controller/elasticsearch/filesettings/file_settings.go
+++ b/pkg/controller/elasticsearch/filesettings/file_settings.go
@@ -51,6 +51,20 @@ type IndexTemplates struct {
 	ComposableIndexTemplates *commonv1.Config `json:"composable_index_templates,omitempty"`
 }
 
+func (st *SettingsState) SetComponentTemplates(t *commonv1.Config) {
+	if st.IndexTemplates == nil {
+		st.IndexTemplates = &IndexTemplates{}
+	}
+	st.IndexTemplates.ComponentTemplates = t
+}
+
+func (st *SettingsState) SetComposableIndexTemplates(t *commonv1.Config) {
+	if st.IndexTemplates == nil {
+		st.IndexTemplates = &IndexTemplates{}
+	}
+	st.IndexTemplates.ComposableIndexTemplates = t
+}
+
 // hash returns the hash of the Settings, considering only the State without the Metadata.
 func (s *Settings) hash() string {
 	return hash.HashObject(s.State)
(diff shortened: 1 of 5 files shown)


thbkrkr:
buildkite test this -f p=gke,t=TestStackConfigPolicy -m s=8.14.3,s=8.15.1,s=8.16.0-SNAPSHOT

thbkrkr (dismissed):

thbkrkr:
buildkite test this -f p=gke,t=TestStackConfigPolicy -m s=8.14.3,s=8.15.1,s=8.16.0-SNAPSHOT

pebrc:
There seems to be an issue with an empty settings file in that it does not correctly delete the snapshot repos created via file-based settings. 

Just consider the failing e2e test: 
1.  ES cluster with a snapshot repo installed through SCP
```
k8surl GET  _snapshot
{"repo_test":{"type":"gcs","settings":{"bucket":"my-bucket","base_path":"snapshots/default-elasticsearch-sample"}}}⏎  
```
2. Delete SCP
3. Settings file is emtpy:
```
kubectl exec elasticsearch-sample-es-default-0 -- cat config/operator/settings.json | jq
Defaulted container "elasticsearch" out of: elasticsearch, elastic-internal-init-filesystem (init), elastic-internal-suspend (init)
{
  "metadata": {
    "version": "1728897840507149000",
    "compatibility": "8.6.1"
  },
  "state": {}
}
```
5. Snapshot repo remains configured in the ES custer
```
k8surl GET  _snapshot
{"repo_test":{"type":"gcs","settings":{"bucket":"my-bucket","base_path":"snapshots/default-elasticsearch-sample"}}}⏎  
```

pebrc:
I am closing this for the time being we can reconsider when the issue with snapshots is fixed

Issue #8102, opened 2024-10-14
[E2E] TestPipelineConfig*Logstash are failing on 8.16.0-SNAPSHOT


barkbay:
We are observing some failures for some Logstash e2e tests since Thursday, October 10th.

Failures seem to be always due to **`expected status to be green but got yellow`** (or `red`) when calling the `/` API:

* `TestPipelineConfigLogstash/Logstash_should_respond_to_metrics_requests`

```
=== RUN   TestPipelineConfigLogstash/Logstash_should_respond_to_metrics_requests
Retries (15m0s timeout):
    step.go:51: 
        	Error Trace:	/go/src/github.com/elastic/cloud-on-k8s/test/e2e/test/utils.go:94
        	Error:      	Received unexpected error:
        	            	expected status to be green but got yellow
        	Test:       	TestPipelineConfigLogstash/Logstash_should_respond_to_metrics_requests
```

* `TestPipelineConfigRefLogstash/Logstash_should_respond_to_metrics_requests`

```
=== RUN   TestPipelineConfigRefLogstash/Logstash_should_respond_to_metrics_requests
Retries (15m0s timeout): .........................................................................................................................................................................................................................................................................................................
    step.go:51: 
        	Error Trace:	/go/src/github.com/elastic/cloud-on-k8s/test/e2e/test/utils.go:94
        	Error:      	Received unexpected error:
        	            	expected status to be green but got red
        	Test:       	TestPipelineConfigRefLogstash/Logstash_should_respond_to_metrics_requests
```

This is fairly easy to reproduce locally, I can confirm that the Logstash response is:

```json
{
  "host" : "test-split-pipeline-j6pk-ls-0",
  "version" : "8.16.0",
  "http_address" : "0.0.0.0:9600",
  "id" : "02115e04-db10-4c04-8263-0426d5c94a5b",
  "name" : "test-split-pipeline-j6pk-ls-0",
  "ephemeral_id" : "4f94107d-edb7-4fed-86b3-8f14caeeebf2",
  "snapshot" : true,
  "status" : "yellow",
  "pipeline" : {
    "workers" : 2,
    "batch_size" : 125,
    "batch_delay" : 50
  },
  "build_date" : "2024-10-13T02:35:37+00:00",
  "build_sha" : "dc0739bdaf2fbab416aaa5ab91ccab9895cd7c3d",
  "build_snapshot" : true
}
```

[...]

PR #8110, opened 2024-10-15
[E2E] Update Logstash health check with the new health API standard

kaisecheng:
This commit updated Logstash health check to align with the new definition for green status.
Only red status is considered as a failure.

Fix: https://github.com/elastic/cloud-on-k8s/issues/8102

### DIFF
--- a/test/e2e/test/logstash/checks.go
+++ b/test/e2e/test/logstash/checks.go
@@ -176,7 +176,12 @@ func (b Builder) CheckStackTestSteps(k *test.K8sClient) test.StepList {
 				Password: password,
 			},
 			Want{
-				Match: map[string]string{"status": "green"},
+				MatchFunc: map[string]func(string) bool{
+					// only red is considered as failure in health API
+					"status": func(status string) bool {
+						return status != "red"
+					},
+				},
 			}),
 		b.CheckMetricsRequest(k,
 			Request{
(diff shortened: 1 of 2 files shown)


kaisecheng:
buildkite test this -f p=kind -m t=TestLogstashPipelineReload,t=TestPipelineConfigLogstash,t=TestPipelineConfigRefLogstash

kaisecheng:
buildkite test this -f p=kind -m t=TestLogstashPipelineReload,t=TestPipelineConfigLogstash,t=TestPipelineConfigRefLogstash

test/e2e/test/logstash/checks.go
```diff
@@ -176,7 +176,12 @@ func (b Builder) CheckStackTestSteps(k *test.K8sClient) test.StepList {
 				Password: password,
 			},
 			Want{
-				Match: map[string]string{"status": "green"},
+				MatchFunc: map[string]func(string) bool{
+					// only red is considered as failure in health API
+					"status": func(status string) bool {
+						return status != "red"
+					},
```
barkbay, line 183: Is there not a chance that we miss another value, like an empty string for example? I was wondering if we could explicitly expect the `green` or `yellow` values, and reuse a function.
```suggestion
						"status": isGreenOrYellow,
```

With `isGreenOrYellow` being:

```golang
// isGreenOrYellow returns true if the status is either green or yellow, red is considered as failure in health API.
func isGreenOrYellow(status string) bool {
	return status == "green" || status == "yellow"
}
```

barkbay (approved):
LGTM, thanks for the fix!
I only left a question/suggestion.

test/e2e/test/logstash/checks.go
kaisecheng: The status only gives one of the color, no empty string

kaisecheng:
buildkite test this -f p=kind -m t=TestLogstashPipelineReload,t=TestPipelineConfigLogstash,t=TestPipelineConfigRefLogstash

kaisecheng:
@barkbay Thanks for reviewing and the suggestion. I think this is ready to merge

PR #8151, opened 2024-10-22
Clarify high availability recommendations in Elasticsearch orchestration docs

stefnestor:
👋 howdy, team! Can we call out that our recommendation for updates applying to [highly available](https://www.elastic.co/guide/en/elasticsearch/reference/current/high-availability.html) clusters should be multiple nodes per [data tier](https://www.elastic.co/guide/en/elasticsearch/reference/current/data-tiers.html) and not just multiple nodes. (Multiple nodes would potentially keep responsiveness as long as master-eligible, but this avoids `status:red` search/ingestion issues.)

### DIFF
--- a/docs/orchestrating-elastic-stack-applications/elasticsearch/orchestration.asciidoc
+++ b/docs/orchestrating-elastic-stack-applications/elasticsearch/orchestration.asciidoc
@@ -149,7 +149,7 @@ Due to relying on Kubernetes primitives such as StatefulSets, the ECK orchestrat
 ** Single-node clusters
 ** Clusters containing indices with no replicas
 
-If an {es} node holds the only copy of a shard, this shard becomes unavailable while the node is upgraded. link:{ref}/high-availability.html[Highly available] clusters are recommended, including having more than one node per link:{ref}/data-tiers.html[data tier] and at least one replica per index.
+If an {es} node holds the only copy of a shard, this shard becomes unavailable while the node is upgraded. link:{ref}/high-availability-cluster-design.html[Highly available] clusters are recommended, including having more than one node per link:{ref}/data-tiers.html[data tier] and at least one replica per index.
 
 * Elasticsearch Pods may stay `Pending` during a rolling upgrade if the Kubernetes scheduler cannot re-schedule them back. This is especially important when using local PersistentVolumes. If the Kubernetes node bound to a local PersistentVolume does not have enough capacity to host an upgraded Pod which was temporarily removed, that Pod will stay `Pending`.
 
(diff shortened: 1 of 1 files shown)


docs/orchestrating-elastic-stack-applications/elasticsearch/orchestration.asciidoc
```diff
@@ -149,7 +149,7 @@ Due to relying on Kubernetes primitives such as StatefulSets, the ECK orchestrat
 ** Single-node clusters
 ** Clusters containing indices with no replicas
 
-If an Elasticsearch node holds the only copy of a shard, this shard becomes unavailable while the node is upgraded. Clusters with more than one node and at least one replica per index are recommended.
+If an {es} node holds the only copy of a shard, this shard becomes unavailable while the node is upgraded. link:{ref}/high-availability.html[Highly available] clusters are recommended, including having more than one node per link:{ref}/data-tiers.html[data tier] and at least one replica per index.
```
stefnestor, line 152:
```suggestion
If an {es} node holds the only copy of a shard, this shard becomes unavailable while the node is upgraded. link:{ref}/high-availability-cluster-design.html[Highly available] clusters are recommended, including having more than one node per link:{ref}/data-tiers.html[data tier] and at least one replica per index.
```
FWIW noting a [highly available](https://www.elastic.co/guide/en/elasticsearch/reference/current/high-availability-cluster-design.html) alternative link which may be better.

kilfoyle (approved):
LGTM! 🚀

stefnestor:
Thanks, @kilfoyle ! While I believe this is true (as it is true for other Elastic platforms like self-hosted and ESS/ECE), my team had some uncertainty to confirm this is also true for ECK. Could we flag this requiring Dev sign-off? 🙏🏽

kilfoyle:
> Could we flag this requiring Dev sign-off?

Yup! If my memory serves, I don't think I can merge anything in this repo without the ECK team first giving it a thumbs up. :-)

kilfoyle:
@elasticmachine run docs-build

docs/orchestrating-elastic-stack-applications/elasticsearch/orchestration.asciidoc
```diff
@@ -149,7 +149,7 @@ Due to relying on Kubernetes primitives such as StatefulSets, the ECK orchestrat
 ** Single-node clusters
 ** Clusters containing indices with no replicas
 
-If an Elasticsearch node holds the only copy of a shard, this shard becomes unavailable while the node is upgraded. Clusters with more than one node and at least one replica per index are recommended.
+If an {es} node holds the only copy of a shard, this shard becomes unavailable while the node is upgraded. link:{ref}/high-availability-cluster-design.html[Highly available] clusters are recommended, including having more than one node per link:{ref}/data-tiers.html[data tier] and at least one replica per index.
```
pebrc, line 152:
```suggestion
If an {es} node holds the only copy of a shard, this shard becomes unavailable while the node is upgraded. To ensure link:{ref}/high-availability-cluster-design.html[high availability] it is recommended to configure clusters with three master nodes, more than one node per link:{ref}/data-tiers.html[data tier] and at least one replica per index.
```
Just having multiple nodes per data tier does not make a cluster highly available. ECK starts treating a cluster as HA only if it also has at least three master nodes and this also what the doc says that you propose to link to.

thbkrkr (approved):

Issue #7833, opened 2024-05-21
Missing ECK network policy for Isolating Elasticsearch Nodes (Ingress 9300)


datadingbat:
I believe this doc page:
https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s_prerequisites.html

is missing an important network policy that needs to be specified for the "Isolating Elasticsearch" section. The Egress section lists "TCP port 9300 to other Elasticsearch nodes in the namespace (transport port)." but the Ingress section is missing "TCP port 9300 from other Elasticsearch nodes in the namespace (transport port)." The YAML file that's referenced in this section shows an Ingress rule for TCP 9300, but it's not provided in the documentation text.

Thanks,
Jesse

PR #8164, opened 2024-10-29
[docs] Add explanation line for es ingress 9300 port

TimBosman:
As described in the #7833 :

The network policy page refers does not specify in text that Elasticsearch needs to accept incoming (ingress) traffic from other elasticsearch nodes on port 9300. While the code does configure that.


thbkrkr (approved):

thbkrkr:
Thank you.

Issue #8161, opened 2024-10-28
Service not reset when set then removed from definition


thbkrkr:
ECK manages a set of services per Elasticsearch cluster that users can configure. By default, the operator configures the http service with type`ClusterIP`. If a user configures it with type `LoadBalancer` and later completely deletes it from the es resource definition, the service is never reconfigured to the default type`ClusterIP`.

### Step to reproduce:
- Deploy an Elasticsearch with a `LoadBalancer` service:

```yaml
apiVersion: elasticsearch.k8s.elastic.co/v1
kind: Elasticsearch
metadata:
  name: c0
spec:
  version: 8.15.3
  http:
    service:
      spec:
        type: LoadBalancer
  nodeSets:
  - name: default
    count: 1
    config:
      node.store.allow_mmap: false
```

- Remove the service from the definition and redeploy:

```yaml
apiVersion: elasticsearch.k8s.elastic.co/v1
kind: Elasticsearch
metadata:
  name: c0
spec:
  version: 8.15.3
  # http:
  #   service:
  #     metadata:
  #       annotations:
  #         answer: "42"
  #     spec:
  #       type: LoadBalancer
  nodeSets:
  - name: default
    count: 1
    config:
      node.store.allow_mmap: false
```

- Check that the service type is still `LoadBalancer`:

```sh
NAME                  TYPE           CLUSTER-IP      EXTERNAL-IP    PORT(S)          AGE
c0-es-http            LoadBalancer   10.1.2.3   192.1.2.3   9200:31142/TCP   3m24s
```

### Why?

We have a function to set the defaults, but we don't set the `type` there: https://github.com/elastic/cloud-on-k8s/blob/645f75022f5117960381b7e52c95c16409a5bcd3/pkg/controller/common/defaults/service.go#L14.

We have a function to `applyServerSideValues` for several reasons, to decide if we need to recreate or update the service, where we retrieve and restore the current reconciled type.
https://github.com/elastic/cloud-on-k8s/blob/645f75022f5117960381b7e52c95c16409a5bcd3/pkg/controller/common/service_control.go#L53-L54

thbkrkr:
I wonder what other default value than `ClusterIP` may be set (comes from https://github.com/elastic/cloud-on-k8s/pull/1011)?
https://github.com/elastic/cloud-on-k8s/blob/645f75022f5117960381b7e52c95c16409a5bcd3/pkg/controller/common/service_control.go#L91-L94

[...]

PR #8165, opened 2024-10-29
Fix reset service type

thbkrkr:
These changes ensure that when deleting the type of an http(s) external service declaration in an elastic resource definition, the service type is properly reset to a default value.

To test: after creating a `LoadBalancer` service, then reapplying without the service type, the http service should be reset to `ClusterIP`.

```yaml
kind: Elasticsearch
metadata:
  name: c0
spec:
  version: 8.15.3
  http:
    service:
      metadata:
        annotations:
          answer: "42"
      spec:
         #type: LoadBalancer
  nodeSets:
  - name: default
    count: 1
    config:
      node.store.allow_mmap: false
```

Relates to #8161.

-------

This can be broken down into ~3~ 2 changes:
- When the type of an external service is not set, it defaults to `ClusterIP`. Thats seems mandatory to me to be able to detect to reset ClusterIP when no service is configured while there is still the version of the service typed LoadBalancer on the server side. It's ok to do this change because the default value of the `Type` field is this value ([v1/types.go](https://github.com/kubernetes/kubernetes/blob/f01e0d64db0eaff171626c5fd210e83be7708588/pkg/apis/core/types.go#L4523-L4524)). So, this will not change the current type for existing clusters that have not set the type.
- When a resource is reconciled and needs to be recreated, reconciliation is skipped if the resource is being deleted. This is because the deletion of the service can take many seconds (e.g.: gcp lb deletion takes >30s). Without this, the reconciliation fails with error `object is being deleted: services "mycluster-es-http" already exists` many times (10 in 30s with the exponetial backoff) until the resource is deleted.

```scala
2024-10-29T20:25:33.611+0100    ERROR   manager.eck-operator    Reconciler error       {
    "service.version": "2.16.0-SNAPSHOT+1dbb7572",  "controller": "elasticsearch-controller",
    "object": {"name": "c0", "namespace": "lab"},
    "namespace": "lab", "name": "c0", "reconcileID": "bb3f9bf3-0ef4-45f6-be9d-a73c94f01d93",
    "error": "object is being deleted: services \"c0-es-http\" already exists",
    "errorCauses": [{"error": "object is being deleted: services \"c0-es-http\" already exists"}]
}
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).reconcileHandler
        /Users/krkr/dev/pkg/mod/sigs.k8s.io/controller-runtime@v0.19.0/pkg/internal/controller/controller.go:316
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).processNextWorkItem
        /Users/krkr/dev/pkg/mod/sigs.k8s.io/controller-runtime@v0.19.0/pkg/internal/controller/controller.go:263
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).Start.func2.2
        /Users/krkr/dev/pkg/mod/sigs.k8s.io/controller-runtime@v0.19.0/pkg/internal/controller/controller.go:224
```

Edit 1: I reverted (d38519e2f):
- When the type of a service changes, server side changes are not applied. We may lose some case where we could update instead of recreate but I think it will avoid some cases where certain fields have no sense with the new type.

Not sure that we need it. In fact, a service is already recreated when type changes from LB->CIP and CIP->LB because we don't apply ClusterIP coming from the server side when the type changes.

Edit 2: Still, if you do the switch CIP->LB->CIP enough quickly before getting a ClusterIP (ok, this is stupid, but) and skipping the recreating, you could see some weird transient error like:

```scala
    "error": "Service \"c0-es-http\" is invalid: [
        spec.allocateLoadBalancerNodePorts: Forbidden: may only be used when `type` is 'LoadBalancer',
        spec.externalTrafficPolicy: Invalid value: \"Cluster\": may only be set for externally-accessible services]"
```


thbkrkr:
I see another bug. `NodePort` -> `ClusterIP`. We detect that the service needs to be recreated but we apply the server side values to the expected service, that is then copied into the object service to be created: 

https://github.com/elastic/cloud-on-k8s/blob/645f75022f5117960381b7e52c95c16409a5bcd3/pkg/controller/common/reconciler/reconciler.go#L100-L107

This results in trying to create a `ClusterIP` service with `externalTrafficPolicy` and boom:

```
2024-10-29T23:46:48.632+0100    INFO    elasticsearch-controller        Deleting resource as it cannot be updated, it will be recreated {"service.version": "2.16.0-SNAPSHOT+ecaf23ef", "iteration": "46", "namespace": "lab", "es_name": "c0", "kind": "Service", "namespace": "lab", "name": "c0-es-http"}
2024-10-29T23:46:48.827+0100    INFO    elasticsearch-controller        Deleted resource successfully   {"service.version": "2.16.0-SNAPSHOT+ecaf23ef", "iteration": "46", "namespace": "lab", "es_name": "c0", "kind": "Service", "namespace": "lab", "name": "c0-es-http"}
2024-10-29T23:46:48.827+0100    INFO    elasticsearch-controller        Creating resource       {"service.version": "2.16.0-SNAPSHOT+ecaf23ef", "iteration": "46", "namespace": "lab", "es_name": "c0", "kind": "Service", "namespace": "lab", "name": "c0-es-http"}
2024-10-29T23:46:49.186+0100    INFO    elasticsearch-controller        Ending reconciliation run       {"service.version": "2.16.0-SNAPSHOT+ecaf23ef", "iteration": "46", "namespace": "lab", "es_name": "c0", "took": "554.2575ms"}
2024-10-29T23:46:49.186+0100    ERROR   manager.eck-operator    Reconciler error        {"service.version": "2.16.0-SNAPSHOT+ecaf23ef", "controller": "elasticsearch-controller", "object": {"name":"c0","namespace":"lab"}, "namespace": "lab", "name": "c0", "reconcileID": "f456219b-cdab-4138-945f-00fc1da75b87", "error": "Service \"c0-es-http\" is invalid: spec.externalTrafficPolicy: Invalid value: \"Cluster\": may only be set for externally-accessible services", "errorCauses": [{"error": "Service \"c0-es-http\" is invalid: spec.externalTrafficPolicy: Invalid value: \"Cluster\": may only be set for externally-accessible services"}]}
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).reconcileHandler
        /Users/krkr/dev/pkg/mod/sigs.k8s.io/controller-runtime@v0.19.0/pkg/internal/controller/controller.go:316
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).processNextWorkItem
        /Users/krkr/dev/pkg/mod/sigs.k8s.io/controller-runtime@v0.19.0/pkg/internal/controller/controller.go:263
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).Start.func2.2
        /Users/krkr/dev/pkg/mod/sigs.k8s.io/controller-runtime@v0.19.0/pkg/internal/controller/controller.go:224
2024-10-29T23:46:49.187+0100    INFO    elasticsearch-controller        Starting reconciliation run     {"service.version": "2.16.0-SNAPSHOT+ecaf23ef", "iteration": "47", "namespace": "lab", "es_name": "c0"}
2024-10-29T23:46:49.188+0100    INFO    elasticsearch-controller        Creating resource       {"service.version": "2.16.0-SNAPSHOT+ecaf23ef", "iteration": "47", "namespace": "lab", "es_name": "c0", "kind": "Service", "namespace": "lab", "name": "c0-es-http"}
2024-10-29T23:46:49.390+0100    INFO    elasticsearch-controller        Created resource successfully   {"service.version": "2.16.0-SNAPSHOT+ecaf23ef", "iteration": "47", "namespace": "lab", "es_name": "c0", "kind": "Service", "namespace": "lab", "name": "c0-es-http"}
2024-10-29T23:46:50.223+0100    INFO    elasticsearch-controller        Ensuring no voting exclusions are set   {"service.version": "2.16.0-SNAPSHOT+ecaf23ef", "iteration": "47", "namespace": "lab", "es_name": "c0", "namespace": "lab", "es_name": "c0"}
2024-10-29T23:46:51.098+0100    INFO    elasticsearch-controller        Ending reconciliation run       {"service.version": "2.16.0-SNAPSHOT+ecaf23ef", "iteration": "47", "namespace": "lab", "es_name": "c0", "took": "1.911611416s"}
2024-10-29T23:46:51.099+0100    INFO    elasticsearch-controller        Starting reconciliation run     {"service.version": "2.16.0-SNAPSHOT+ecaf23ef", "iteration": "48", "namespace": "lab", "es_name": "c0"}
```

We could return after the deletion instead of creating directly but it causes a downtime of the service until the next reconciliation happens to do the creation, not great.

I think this gives a good reason to skip applying server side values when the type changes, no?

```go
diff --git a/pkg/controller/common/service_control.go b/pkg/controller/common/service_control.go
index 551552b23..12a957888 100644
--- a/pkg/controller/common/service_control.go
+++ b/pkg/controller/common/service_control.go
@@ -88,6 +88,11 @@ func needsUpdate(expected *corev1.Service, reconciled *corev1.Service) bool {

 // applyServerSideValues applies any default that may have been set from the reconciled version.
 func applyServerSideValues(expected, reconciled *corev1.Service) {
+       // skip if the service type changes from something different to the default ClusterIP value.
+       if reconciled.Spec.Type != corev1.ServiceTypeClusterIP && expected.Spec.Type != reconciled.Spec.Type {
+               return
+       }
+
```

pkg/controller/common/reconciler/reconciler.go
```diff
@@ -88,8 +88,21 @@ func ReconcileResource(params Params) error {
...
 
 	create := func() error {
+		if deleted {
+			// check if not being deleted
+			err = params.Client.Get(params.Context, types.NamespacedName{Name: name, Namespace: namespace}, params.Reconciled)
+			if err != nil && !apierrors.IsNotFound(err) {
+				return err
+			}
+			if !params.Reconciled.GetDeletionTimestamp().IsZero() {
+				log.Info("Waiting for resource to be created because the old one is being deleted")
```
barkbay, line 101: "Waiting" seems a bit misleading to me. With this change it is up to the caller to detect that the resource has actually not been reconciled, and try again later. By "swallowing"/"hiding" the conflict and returning `nil` here we are assuming that the caller is going to do another attempt, but we can't be sure of that? This makes me feel that it should be up to the caller to decide what to do/log in case of a conflict, by using `apierrors.IsAlreadyExists()`

barkbay:
> I think this gives a good reason to skip applying server side values when the type changes, no?

Yes, I think it makes sense.

pkg/controller/common/reconciler/reconciler.go
thbkrkr: You are right, we can't be 100% sure that the caller will retry. You proposal looks a good idea.

This means we will delete and (try) to create until the deletion is effective.

The only thing that tickles me is that with the current log messages it can be a little confusing.

From LoadBalancer to ClusterIP:
```scala
Deleting resource as it cannot be updated, it will be recreated
Deleted resource successfully
Creating resource
Deleting resource as it cannot be updated, it will be recreated
Deleted resource successfully
Creating resource
Deleting resource as it cannot be updated, it will be recreated
Deleted resource successfully
Creating resource
Deleting resource as it cannot be updated, it will be recreated
Deleted resource successfully
Creating resource
Deleting resource as it cannot be updated, it will be recreated
Deleted resource successfully
Creating resource
Deleting resource as it cannot be updated, it will be recreated
Deleted resource successfully
Creating resource
Deleting resource as it cannot be updated, it will be recreated
Deleted resource successfully
Creating resource
Creating resource
Created resource successfully
Ensuring no voting exclusions are set
Ensuring no voting exclusions are set
Ensuring no voting exclusions are set
```

barkbay: > The only thing that tickles me is that with the current log messages it can be a little confusing.

Is it when we requeue immediately, or when we just ignore the error in the driver? Maybe we could add an additional log in that specific case.

pkg/controller/elasticsearch/driver/driver.go
```diff
@@ -140,6 +141,9 @@ func (d *defaultDriver) Reconcile(ctx context.Context) *reconciler.Results {
 
 	externalService, err := common.ReconcileService(ctx, d.Client, services.NewExternalService(d.ES), &d.ES)
 	if err != nil {
+		if apierrors.IsAlreadyExists(err) {
+			return results.WithReconciliationState(defaultRequeue.WithReason("Could not reconcile external service"))
```
thbkrkr, line 145: Open to any suggestions for this message that is used in the `ReconciliationComplete` status condition:

```yaml
> k get es c0 -o yaml | yq .status
availableNodes: 1
conditions:
  - lastTransitionTime: "2024-10-30T11:40:06Z"
    message: Could not reconcile external service  # 👈 👈 👈 👈 👈
    status: "False"
    type: ReconciliationComplete
  - lastTransitionTime: "2024-10-30T11:36:02Z"
    message: All nodes are running version 8.15.3
    status: "True"
    type: RunningDesiredVersion
  - lastTransitionTime: "2024-10-30T11:39:49Z"
    message: Service default/c0-es-internal-http has endpoints
    status: "True"
    type: ElasticsearchIsReachable
  - lastTransitionTime: "2024-10-30T11:38:57Z"
    message: 'Cannot get compute and storage resources from Elasticsearch resource generation 10: cannot compute resources for NodeSet "default": no CPU request or limit set'
    status: "False"
    type: ResourcesAwareManagement
health: unknown
inProgressOperations:
  downscale:
    lastUpdatedTime: "2024-10-30T11:35:31Z"
  upgrade:
    lastUpdatedTime: "2024-10-30T11:35:31Z"
  upscale:
    lastUpdatedTime: "2024-10-30T11:35:31Z"
observedGeneration: 11
phase: ApplyingChanges
version: 8.15.3
```

pkg/controller/common/reconciler/reconciler.go
thbkrkr: It is when we ignore the error and then requeue in the driver (at the caller level), see 0218fd6db.

pkg/controller/elasticsearch/driver/driver.go
barkbay, line 145: Maybe something along the lines of
```suggestion
			return results.WithReconciliationState(defaultRequeue.WithReason("Pending external Service recreation"))
```

pebrc: Or maybe be explicit about the name of the service, because I think the concept of "external" vs "internal" service is not something users are aware of. We don't mention it in the docs afaik. It is a code level concept only. 

```diff
diff --git a/pkg/controller/elasticsearch/driver/driver.go b/pkg/controller/elasticsearch/driver/driver.go
index df3defd16..434636b04 100644
--- a/pkg/controller/elasticsearch/driver/driver.go
+++ b/pkg/controller/elasticsearch/driver/driver.go
@@ -142,7 +142,7 @@ func (d *defaultDriver) Reconcile(ctx context.Context) *reconciler.Results {
 	externalService, err := common.ReconcileService(ctx, d.Client, services.NewExternalService(d.ES), &d.ES)
 	if err != nil {
 		if apierrors.IsAlreadyExists(err) {
-			return results.WithReconciliationState(defaultRequeue.WithReason("Could not reconcile external service"))
+			return results.WithReconciliationState(defaultRequeue.WithReason(fmt.Sprintf("Pending %s service recreation", services.ExternalServiceName(d.ES.Name))))
 		}
 		return results.WithError(err)
 	}
```

barkbay (approved):
LGTM

PR #8169, opened 2024-10-31
Fix resetting service type to default when not specified (#8165)

thbkrkr:
Backport the following commit to `2.15`:
- #8165


pebrc (approved):
LGTM, not sure what is going on with the build.

thbkrkr:
In the backport commit, the new code uses an import (k8serrors) added by ae1f62cd0e74 which is not in the `2.15` branch. I will fix the import.

PR #8175, opened 2024-11-01
Ensure esclient is closed after each reconcile.

### LINKED ISSUE #8174: Potential memory leak in the ECK Autoscaler.
Image

There seems to be a potential memory leak in the ECK operator, that seems to be associated with autoscaling being enabled. Upon testing memory utilization of the operator since #357, utilization was stable (as can be seen in the graph), until autoscaling was enabled on the cluster. Since that time (~3 days) heap usage has grown from ~20MB up to ~120MB.

naemono:
Resolves #8174 

While investigating #8174 a couple pprof heap dumps were taken over the course of a couple of days, and compared using the diff capabilities of pprof:
```
go tool pprof -http=:8082 -no_browser -diff_base dump-1.out.gz dump-2.out.gz dump-3.out.gz
```
This brought to light the following function in the autoscaling controller (`attemptOnlineReconciliation`), which seems to have had it's heap allocations grow by about %2000 since the first dump:
image

Upon further investigation, we don't appear to be closing idle connections in the underlying http client, which could be leaking goroutines, and therefore leading to this memory increase.

### DIFF
--- a/pkg/controller/autoscaling/elasticsearch/driver.go
+++ b/pkg/controller/autoscaling/elasticsearch/driver.go
@@ -114,6 +114,7 @@ func (r *baseReconcileAutoscaling) attemptOnlineReconciliation(
 	if err != nil {
 		return nil, err
 	}
+	defer esClient.Close()
 
 	// Update Machine Learning settings
 	mlNodes, maxMemory := esv1.GetMLNodesSettings(autoscalingSpec)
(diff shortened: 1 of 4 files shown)


pkg/controller/autoscaling/elasticsearch/driver.go
```diff
@@ -114,6 +114,7 @@ func (r *baseReconcileAutoscaling) attemptOnlineReconciliation(
 	if err != nil {
 		return nil, err
 	}
+	defer esClient.Close()
```
pebrc, line 117: I think we would need the same here https://github.com/elastic/cloud-on-k8s/blob/fa9234eaab21d76597706d8300d153d9ea631997/pkg/controller/stackconfigpolicy/controller.go#L714

pebrc (approved):

naemono:
Signs from initial testing of the latest build show good signs, as the memory seems stable. I'm going to let it run the rest of the day to ensure before merging.

kvalliyurnatt (approved):
Great Find!

naemono:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|2.15|[](https://github.com/elastic/cloud-on-k8s/pull/8180)|

### Questions ?

PR #8184, opened 2024-11-06
[WIP] Make `image` setting consistent accross all ECK Helm charts

jeanfabrice:
We have 4 Helm charts that implement the image setting and copy it to `.spec.image`
We have 3 others Helm charts (Kibnana, Elastic-Agent and Beat) that just copy the entire spec object from `values.yaml` to the custom resource `.spec`
The idea of this PR is to make all Helm Charts consistent while exposing the `image` setting for clarity

It implies breaking changes. I don't know (yet) how we version / handle Helm charts breaking changes at the Helm charts level.

### DIFF
--- a/deploy/eck-stack/charts/eck-agent/examples/fleet-agents.yaml
+++ b/deploy/eck-stack/charts/eck-agent/examples/fleet-agents.yaml
@@ -3,24 +3,23 @@
 #
 version: 8.17.0-SNAPSHOT
 
-spec:
-  # This must match the name of an Agent policy.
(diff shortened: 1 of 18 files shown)


naemono:
@jeanfabrice thanks for making this consistent. I'll look at this and see if there's a way to make it a non-breaking change. Maybe only use `Values.image` if `.Values.spec.image` is missing? I'll see what I can do.

thbkrkr:
Superseded by:
- #8192 
- #8246
- #8248

Issue #8204, opened 2024-11-14
eck/eck-operator fips and ubi


ke5C2Fin:
Default image is `eck/eck-operator:2.15.0`.
https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s_installation.html#k8s_fips_compliant_installation_using_helm
This document says to set `image.repository=docker.elastic.co/eck/eck-operator-fips`, which gives us this image `eck/eck-operator-fips:2.15.0`.
If we do not do that, but we set `config.ubiOnly=true`, we get this image `eck/eck-operator-ubi:2.15.0`.
However, if we set both fips and ubi we get this image `eck/eck-operator-fips-ubi:2.15.0`, which doesn't exist because the actual image is `eck/eck-operator-ubi-fips:2.15.0`.

thbkrkr:
Thank you for reporting this. The issue is indeed a bug in how the Helm chart constructs the image name when it contains `-fips` and `config.ubiOnly` is enabled. We’ll work on resolving this. In the meantime, I think you can use the following workaround:

```sh
helm install elastic-operator elastic/eck-operator \
  -n elastic-system --create-namespace \
  --set=image.repository=docker.elastic.co/eck/eck-operator-ubi-fips \
  --set=config.containerSuffix=-ubi
```

PR #8272, opened 2024-11-28
[Helm] Add a setting to deploy FIPS compliant ECK image

barkbay:
Fixes https://github.com/elastic/cloud-on-k8s/issues/8204

I'm not sure there are other (simple) ways than adding a new setting if we want to be compatible with `config.ubiOnly=true`?


thbkrkr (approved):
This seems to me to be a reasonable small change to solve the problem compared to producing `-fips-ubi` instead of `-ubi-fips`, which I think would be close to a breaking change involving many more changes. It's also nice to have an explicit setting to enable fips, rather than knowing that you have to use a specific image.

LGTM

PR #8278, opened 2024-11-29
[Helm] Updated pod labels to include all standard k8s labels and updated hel…

jimmybchopps:
- Updated labels to use common kubernetes labels on pods (previously it was only on statefulset and several were missing from pod) to increase traceability throughout the application.
- Added helm unittests to check for updated labels in addition to the podLabels that can be provided via values.yaml file
- Ran `make helm-test` to validate changes and updates to test

### DIFF
--- a/deploy/eck-operator/templates/statefulset.yaml
+++ b/deploy/eck-operator/templates/statefulset.yaml
@@ -31,7 +31,7 @@ spec:
         {{- toYaml . | nindent 8 }}
         {{- end }}
       labels:
-        {{- include "eck-operator.selectorLabels" . | nindent 8 }}
+        {{- include "eck-operator.labels" . | nindent 8 }}
         {{- with .Values.podLabels }}
         {{- toYaml . | nindent 8 }}
         {{- end }}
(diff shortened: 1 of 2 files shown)


thbkrkr:
We use the operator Helm chart to generate the YAML manifests that allow the operator to be installed without Helm.

Check https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-installing-eck.html:
```sh
# Install without Helm
kubectl create -f https://download.elastic.co/downloads/eck/2.15.0/crds.yaml
kubectl apply -f https://download.elastic.co/downloads/eck/2.15.0/operator.yaml // <-- operator YAML manifest
```

You can call the generation via `make generate`. The operator YAML manifest is then written in `./config/operator.yaml`.

One issue with your change is that we end up with "Helm" bits in this YAML manifest.

```sh
> make generate
> cat config/operator.yaml | grep -i helm
        helm.sh/chart: eck-operator-2.16.0-SNAPSHOT
```

jimmybchopps:
@thbkrkr I see, I'll rework the solution and test it out using make generate to make sure this is resolved.

barkbay:
Closing due to lack of activity (feel free to reopen though).
I created https://github.com/elastic/cloud-on-k8s/issues/8584 to track this improvement.

PR #4894, opened 2021-09-29
Update cert manager.io examples in the doc and recipes

thbkrkr:
- Bump to `cert-manager.io/v1` 
- Set `subject.organizations` to not generate a certificate that is rejected as malformed by Java, due to lack of an issuer DN
- Add an example to issue multiple certificates from the same self-signed CA

Resolves #4845.

PR #8317, opened 2024-12-11
Fix cert-manager API version

nitishaggarwal49:
It looks like cert-manager.io/v1alpha2 has been replaced with cert-manager.io/v1.
cc @jeanfabrice


thbkrkr (approved):

thbkrkr:
Thanks. We missed this one in:
- https://github.com/elastic/cloud-on-k8s/pull/4894.

PR #8327, opened 2024-12-12
Refactor stack monitoring templating

### LINKED ISSUE #8314: Make the stack monitoring templating more modular
We are adding more an more application specific attributes to the common templating logic the latest example is https://github.com/elastic/cloud-on-k8s/pull/8311

* We should investigate to make the code a bit more modular, to have for example the Kibana module specifics only in the Kibana controller.
* Futhermore it would be good if they code was a bit easier to test. Currently we never test the actualy rendered configuration files but a test template in the common code is used for testing. The only real test are e2e tests which are expensive and slow.

pebrc:
Fixes #8314

This PR introduces the following changes: 
* separate template rendering from the shared stack monitoring code as we have have introduced a few application specific config settings in the templates and Beat was not using the shared rendering to begin with
* test the actual templates used in production code instead of a synthetic template in the shared code
* modify existing tests to use snapshots instead of counting volumes/volume mounts (more verbose but easier to verify in my opinion, better diffs on test failures) 

Not addressed:
* `stackmon.Metricbeat|Filebeat` are called twice: once for generating the config secret and once to generate the side cars (duplicated work)

### DIFF
--- a/pkg/controller/common/stackmon/sidecar.go
+++ b/pkg/controller/common/stackmon/sidecar.go
@@ -102,6 +102,8 @@ func NewBeatSidecar(ctx context.Context, client k8s.Client, beatName string, ima
 	}, nil
 }
 
+// CAVolume returns a volume containing the CA certificate for the monitored resource if TLS is enabled.
+// If TLS is not enabled or no (self-signed) CA is in use, it returns nil.
 func CAVolume(
 	c k8s.Client,
 	nsn types.NamespacedName,
@@ -121,4 +123,14 @@ func CAVolume(
 		fmt.Sprintf("%s-local-ca", string(associationType)),
 		fmt.Sprintf("/mnt/elastic-internal/%s/%s/%s/certs", string(associationType), nsn.Namespace, nsn.Name),
 	), nil
+}
+
+// TemplateParams are commonly used parameters to render a Beats configuration template.
+// Stack monitoring implementations can choose to implement their own template parameters if needed.
+type TemplateParams struct {
+	URL      string
+	Username string
+	Password string
+	IsSSL    bool
+	CAVolume volume.VolumeLike
 }
--- a/pkg/controller/common/stackmon/sidecar.go
+++ b/pkg/controller/common/stackmon/sidecar.go
@@ -6,11 +6,14 @@ package stackmon
 
 import (
 	"context"
+	"fmt"
 	"hash"
 
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
 
 	commonv1 "github.com/elastic/cloud-on-k8s/v2/pkg/apis/common/v1"
+	"github.com/elastic/cloud-on-k8s/v2/pkg/controller/common/certificates"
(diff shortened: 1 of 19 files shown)


pebrc:
buildkite test this -f p=gke,t=Test.*StackMonitoring -m s=8.16.0,s=7.17.25,s=8.15.4

pkg/controller/elasticsearch/stackmon/sidecar.go
```diff
@@ -35,20 +36,37 @@ func Metricbeat(ctx context.Context, client k8s.Client, es esv1.Elasticsearch) (
...
-		password,
-		es.Spec.HTTP.TLS.Enabled(),
-	)
+
+	v, err := version.Parse(es.Spec.Version)
+	if err != nil {
+		return stackmon.BeatSidecar{}, err
+	}
+
+	caVolume, err := stackmon.CAVolume(client, k8s.ExtractNamespacedName(&es), esv1.ESNamer, commonv1.KbMonitoringAssociationType /*???*/, es.Spec.HTTP.TLS.Enabled())
```
barkbay, line 45: I was just doing a first read of your PR and noticed the `/*???*/`, I agree that `KbMonitoringAssociationType` feels a bit odd here but did not dig enough to be sure.

pebrc: I wanted to investigate this myself too and forgot about it. It just has an effect on the name of the volume mount. Changing it would mean a restart for the cluster. But maybe we should do it nonetheless

pkg/controller/common/stackmon/sidecar.go
```diff
@@ -125,3 +101,24 @@ func NewBeatSidecar(ctx context.Context, client k8s.Client, beatName string, ima
 		Volumes:      podVolumes,
 	}, nil
 }
+
+func CAVolume(
```
barkbay, line 105: nit: for new readers, let's maybe add a godoc to explain which CA we are referring to here as there might be 2 CAs involved in the process (the one to connect to monitored component and the one from the association).

pkg/controller/kibana/stackmon/sidecar.go
```diff
@@ -60,20 +61,38 @@ func Metricbeat(ctx context.Context, client k8s.Client, kb kbv1.Kibana, basePath
...
-	)
+	v, err := version.Parse(kb.Spec.Version)
+	if err != nil {
+		return stackmon.BeatSidecar{}, err // error unlikely and should have been caught during validation
+	}
+	caVol, err := stackmon.CAVolume(client, k8s.ExtractNamespacedName(&kb), kbv1.KBNamer, commonv1.KbMonitoringAssociationType, kb.Spec.HTTP.TLS.Enabled())
+	if err != nil {
+		return stackmon.BeatSidecar{}, err
+	}
+	type inputConfigData struct {
```
barkbay, line 72: cosmetic nit, we are also sometimes using an anonymous struct, I was wondering if we should make this type a first class citizen in `stackmon`.

barkbay (approved):
LGTM 👍

PR #8343, opened 2024-12-17
Only conditionally set Kibana default security context.

naemono:
While preparing for upcoming release during OCP testing, it was found that the default security context for Kibana fails:

```
Error creating: pods "quickstart-kb-78dccccd95-" is forbidden: unable to validate against any security context constraint: [provider "anyuid": Forbidden: not usable by user or serviceaccount, provider restricted-v2: .spec.securityContext.fsGroup: Invalid value: []int64{1000}: 1000 is not an allowed group, provider restricted-v2: .initContainers[0].runAsUser: Invalid value: 1000: must be in the ranges: [1000700000, 1000709999], provider restricted-v2: .containers[0].runAsUser: Invalid value: 1000: must be in the ranges: [1000700000, 1000709999], ...]
```

The Kibana default security context was not following the previous pattern set with Elasticsearch where we only set this when the relevant flag `set-default-security-context`. This adjusts this to follow that pattern.

### DIFF
--- a/pkg/controller/kibana/pod.go
+++ b/pkg/controller/kibana/pod.go
@@ -40,6 +40,7 @@ const (
 	KibanaBasePathEnvName        = "SERVER_BASEPATH"
 	KibanaRewriteBasePathEnvName = "SERVER_REWRITEBASEPATH"
 	defaultFSGroup               = 1000
+	defaultFSUser                = 1000
 )
 
 var (
--- a/pkg/controller/kibana/pod.go
+++ b/pkg/controller/kibana/pod.go
@@ -12,6 +12,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/utils/ptr"
 	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/elastic/go-ucfg"
@@ -39,6 +40,7 @@ const (
 	TempVolumeMountPath          = "/tmp"
 	KibanaBasePathEnvName        = "SERVER_BASEPATH"
 	KibanaRewriteBasePathEnvName = "SERVER_REWRITEBASEPATH"
+	defaultFSGroup               = 1000
 )
 
 var (
@@ -108,6 +110,7 @@ func NewPodTemplateSpec(
 	keystore *keystore.Resources,
 	volumes []volume.VolumeLike,
 	basePath string,
+	setDefaultSecurityContext bool,
 ) (corev1.PodTemplateSpec, error) {
 	labels := kb.GetIdentityLabels()
 	labels[kblabel.KibanaVersionLabelName] = kb.Spec.Version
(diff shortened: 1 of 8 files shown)


pkg/controller/kibana/securitycontext.go
```diff
@@ -19,8 +19,6 @@ var (
 		},
 		Privileged:             ptr.To(bool(false)),
 		ReadOnlyRootFilesystem: ptr.To(bool(true)),
-		RunAsUser:              ptr.To(int64(1000)),
-		RunAsGroup:             ptr.To(int64(1000)),
```
pebrc, line 23: Why remove that?

pkg/controller/kibana/pod.go
```diff
@@ -39,6 +39,7 @@ const (
 	TempVolumeMountPath          = "/tmp"
 	KibanaBasePathEnvName        = "SERVER_BASEPATH"
 	KibanaRewriteBasePathEnvName = "SERVER_REWRITEBASEPATH"
+	defaultFSGroup               = 1000
```
pebrc, line 42: Unused?

pebrc (commented):
We also need to update the description of the flag in `main.go` and the docs. It currently says "Enables setting the default security context with fsGroup=1000 for Elasticsearch 8.0+ Pods. Ignored pre-8.0. Possible values: true, false, auto-detect","

pkg/controller/kibana/pod.go
naemono: Updated to use this.

pkg/controller/kibana/securitycontext.go
```diff
@@ -19,10 +19,10 @@ var (
...
+		RunAsUser:              ptr.To(int64(defaultFSGroup)),
```
pebrc, line 22: technically this is not the fsGroup...

pebrc (approved):

pkg/controller/kibana/securitycontext.go
naemono: I added another const.

pebrc:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|2.16|[](https://github.com/elastic/cloud-on-k8s/pull/8344)|

### Questions ?

pkg/controller/kibana/pod.go
```diff
@@ -137,9 +140,9 @@ func NewPodTemplateSpec(
 	// Limiting to 7.10.0 here as there was a bug in previous versions causing rebuilding
 	// of browser bundles to happen on plugin install, which would attempt a write to the
 	// root filesystem on restart.
-	if v.GTE(version.From(7, 10, 0)) {
-		builder.WithPodSecurityContext(defaultPodSecurityContext).
-			WithContainersSecurityContext(defaultSecurityContext).
+	if v.GTE(version.From(7, 10, 0)) && setDefaultSecurityContext {
```
barkbay, line 143: A bit late but I think this is not consistent with Elasticsearch for which only the addition of the [Pod SecurityContext](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.32/#podsecuritycontext-v1-core) is conditioned by this variable:
https://github.com/elastic/cloud-on-k8s/blob/59aada97b49774a9ac0de712822067525973c6ad/pkg/controller/elasticsearch/nodespec/podspec.go#L91-L94

While we always set a [SecurityContext for the containers](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.32/#securitycontext-v1-core) to set things like `readOnlyRootFilesystem` or `allowPrivilegeEscalation`:
https://github.com/elastic/cloud-on-k8s/blob/59aada97b49774a9ac0de712822067525973c6ad/pkg/controller/elasticsearch/nodespec/podspec.go#L134

Not a blocker for `2.16` but we may want to make it consistent.

PR #8346, opened 2024-12-18
Chown for elasticsearch folder recursively

vara2504:
RedHat Linux is causing a 'Permission denied' error for subfolders and files because chown was not applied recursively. As a result, when changing the Elasticsearch UID/GID, the Elasticsearch container fails with an 'Access Denied' error for the data/* files.

### DIFF
--- a/pkg/controller/elasticsearch/initcontainer/prepare_fs_script.go
+++ b/pkg/controller/elasticsearch/initcontainer/prepare_fs_script.go
@@ -130,7 +130,7 @@ var scriptTemplate = template.Must(template.New("").Parse(
 	if [[ $EUID -eq 0 ]]; then
 		{{range .ChownToElasticsearch}}
 			echo "chowning {{.}} to elasticsearch:elasticsearch"
-			chown -v elasticsearch:elasticsearch {{.}}
+			chown -vR elasticsearch:elasticsearch {{.}}
 		{{end}}
 	fi
 	echo "chown duration: $(duration $chown_start) sec."


pkg/controller/elasticsearch/initcontainer/prepare_fs_script.go
```diff
@@ -130,7 +130,7 @@ var scriptTemplate = template.Must(template.New("").Parse(
 	if [[ $EUID -eq 0 ]]; then
 		{{range .ChownToElasticsearch}}
 			echo "chowning {{.}} to elasticsearch:elasticsearch"
-			chown -v elasticsearch:elasticsearch {{.}}
+			chown -vR elasticsearch:elasticsearch {{.}}
```
barkbay, line 133: This requires the container to run as root which is not the case for the Elasticsearch image (by default). Shouldn't this problem be solved by https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#configure-volume-permission-and-ownership-change-policy-for-pods?

vara2504: @barkbay : ECK runs init containers as[ root already]( https://github.com/elastic/cloud-on-k8s/blob/7792ac9facd3ebe1524bab70d0c4f6c6e8577d9c/pkg/controller/elasticsearch/initcontainer/prepare_fs_script.go#L130), which is necessary to perform actions like chown. Since we are already performing the chown operation, why not make it recursive.

barkbay: As mentioned in my previous message the ES container is no longer running as root by default (IIRC since Elasticsearch 8.x). I think the `if` statement you are referring to is not run most of the time.

vara2504: @barkbay Thank you. I could see now, In elastic 8 where elastic container is set to run as non root . Will close the PR.

PR #8380, opened 2024-12-27
Add additional volume for Kibana logs when hardened security context is enabled

naemono:
related: #8378 

This adds an additional volume to the Kibana pod when the hardened security context is enabled as `readOnlyRootFilesystem` is set to `true`.

### DIFF
--- a/pkg/controller/kibana/pod.go
+++ b/pkg/controller/kibana/pod.go
@@ -55,7 +55,7 @@ var (
 	// the Kibana pod has readOnlyRootFilesystem set to true.
 	PluginsVolume = volume.NewEmptyDirVolume(PluginsVolumeName, PluginsVolumeMountPath)
 
-	// LogsVolume can be used to persist logs after installation when
+	// LogsVolume can be used to persist logs even when
 	// the Kibana pod has readOnlyRootFilesystem set to true.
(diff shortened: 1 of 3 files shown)


pkg/controller/kibana/pod.go
```diff
@@ -53,6 +55,10 @@ var (
 	// the Kibana pod has readOnlyRootFilesystem set to true.
 	PluginsVolume = volume.NewEmptyDirVolume(PluginsVolumeName, PluginsVolumeMountPath)
 
+	// LogsVolume can be used to persist logs after installation via an init container when
```
pebrc, line 58: Not sure how the initcontainer is relevant here?

pebrc:
buildkite test this -f p=gke,E2E_TAGS=kb

pkg/controller/kibana/pod.go
```diff
@@ -53,6 +55,10 @@ var (
 	// the Kibana pod has readOnlyRootFilesystem set to true.
 	PluginsVolume = volume.NewEmptyDirVolume(PluginsVolumeName, PluginsVolumeMountPath)
 
+	// LogsVolume can be used to persist logs after installation when
```
pebrc, line 58:
```suggestion
	// LogsVolume can be used to persist logs even when
```

pebrc (approved):

Issue #8391, opened 2025-01-03
[E2E] Resources are not cleaned up when test fails


barkbay:
While investigating an E2E test failure I realized that the related Elasticsearch cluster was never deleted:

<img width="1666" alt="Image" src="https://github.com/user-attachments/assets/7a1abbca-6279-468d-ac34-5e19489a5e3e" />

On the screenshot above we can see that the Elasticsearch Pods were still running while the test failed several hours ago.

This might explain the scheduling issues reported in https://github.com/elastic/cloud-on-k8s/pull/8377

PR #8392, opened 2025-01-03
[E2E] Fix deleteTestResources

barkbay:
Creating a dynamic client from `clntset.RESTClient()` does not seem to work.
It was also already reported a while back in https://github.com/openshift/client-go/issues/92#issuecomment-464269221, this PR implements the same fix: using `NewForConfig(cfg)` instead.

Fix https://github.com/elastic/cloud-on-k8s/issues/8391


test/e2e/test/utils.go
```diff
@@ -214,8 +214,11 @@ func deleteTestResources(ctx context.Context) error {
 		}
 	}
 
+	dynamicClient, err := dynamic.NewForConfig(cfg)
```
pebrc, line 217: Shall we add a comment to explain the workaround?

pebrc (approved):

PR #8412, opened 2025-01-15
[2.16.1] Release notes, highlights, known issues

naemono:
To be backported to `2.16.1-wip` branch upon merge.

### DIFF
--- a/docs/release-notes/highlights-2.16.0.asciidoc
+++ b/docs/release-notes/highlights-2.16.0.asciidoc
@@ -5,8 +5,8 @@
 [id="{p}-2160-known-issues"]
 === Known issues
 
-- Enabling a Kibana hardened security context as well as enabling Kibana audit logs will cause Kibana to go into a `CrashLoopBackoff` state with the error `read-only file system, open '/usr/share/kibana/logs/audit.log'`.
-- Enabling a Kibana hardened security context as well as utilizing Kibana plugins would cause the plugins to fail to load.
+- Enabling a Kibana hardened security context (only supported in Kibana 7.10+) as well as enabling Kibana audit logs will cause Kibana to go into a `CrashLoopBackoff` state with the error `read-only file system, open '/usr/share/kibana/logs/audit.log'`.
+- Enabling a Kibana hardened security context (only supported in Kibana 7.10+) as well as utilizing Kibana plugins would cause the plugins to fail to load.
 
 [float]
 [id="{p}-2160-new-and-notable"]
--- a/docs/release-notes/highlights-2.16.0.asciidoc
+++ b/docs/release-notes/highlights-2.16.0.asciidoc
@@ -2,6 +2,13 @@
 == 2.16.0 release highlights
 
 [float]
+[id="{p}-2160-known-issues"]
+=== Known issues
+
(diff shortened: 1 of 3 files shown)


kvalliyurnatt (approved):

docs/release-notes/highlights-2.16.0.asciidoc
```diff
@@ -1,6 +1,13 @@
 [[release-highlights-2.16.0]]
 == 2.16.0 release highlights
 
+[float]
+[id="{p}-2160-known-issues"]
+=== Known issues
+
+- Enabling a Kibana hardened security context as well as enabling Kibana audit logs will cause Kibana to go into a `CrashLoopBackoff` state with the error `read-only file system, open '/usr/share/kibana/logs/audit.log'`.
```
barkbay, line 8: Since users do not set explicitly this security context, should we instead mention which Kibana versions[^1] are affected when upgrading to `2.16` ?

[^1]: I'm not sure myself actually, should it be `7.9` as mentioned [here](https://github.com/elastic/cloud-on-k8s/blob/0efeb943cd04bfaa252f2950a7fff410d40085c4/pkg/controller/kibana/pod.go#L146-L149), or `7.10` as tested [here](https://github.com/elastic/cloud-on-k8s/blob/0efeb943cd04bfaa252f2950a7fff410d40085c4/pkg/controller/kibana/initcontainer/configmap.go#L27-L28)?

pebrc:
Are we also updating the supported versions to include k8s 1.32 or are we doing this in a separate change?

docs/release-notes/highlights-2.16.0.asciidoc
pebrc: I think we should clean this up. It looks like 7.9 works after a quick smoke test I did with the security context in place with one error/warning:
```
Babel could not write cache to file: /usr/share/kibana/optimize/.babel_register_cache.json 
because it resides in a readonly filesystem. Cache is disabled.
```

naemono:
> Are we also updating the supported versions to include k8s 1.32 or are we doing this in a separate change?

@pebrc where was this discussed? I would think this change would be in a minor release, not a patch release.

naemono:
> > Are we also updating the supported versions to include k8s 1.32 or are we doing this in a separate change?
> 
> @pebrc where was this discussed? I would think this change would be in a minor release, not a patch release.

Discussed internally. Added in https://github.com/elastic/cloud-on-k8s/pull/8412/commits/9aecc5729eee2475c468816f32d08fbf9d26ad24

docs/release-notes/highlights-2.16.0.asciidoc
naemono: Fixed in https://github.com/elastic/cloud-on-k8s/pull/8413. Will update the wording.

naemono: Updated in https://github.com/elastic/cloud-on-k8s/pull/8412/commits/40b51b3ce39895743f29712bef98c9331361357f

pebrc:
> > Are we also updating the supported versions to include k8s 1.32 or are we doing this in a separate change?
> 
> @pebrc where was this discussed? I would think this change would be in a minor release, not a patch release.

https://github.com/elastic/cloud-on-k8s/pull/8404 mentioned it as a possibility.

pebrc (approved):

PR #8415, opened 2025-01-16
Add support in documentation for k8s 1.32.

naemono:
Do we also want to update this: https://github.com/elastic/cloud-on-k8s/blob/8b2cf421bc370a81d08bf3cd2c605614077824d1/hack/api-docs/config.yaml#L19

And this seems to have been missed in `main`: https://github.com/elastic/cloud-on-k8s/blob/8b2cf421bc370a81d08bf3cd2c605614077824d1/hack/operatorhub/templates/csv.tpl#L287

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ Current features:
 
 Supported versions:
 
-*  Kubernetes 1.27-1.31
+*  Kubernetes 1.27-1.32
 *  OpenShift 4.12-4.17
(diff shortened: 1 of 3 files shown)


pebrc (approved):

pebrc:
Should this target main first and the be backported?

naemono:
> Should this target main first and the be backported?

Was wondering that myself after I opened this... Let's do that instead

PR #8431, opened 2025-01-23
[Post 2.16.1] Update upgrade test harness with v2.16.1

naemono:
This PR updates the versions in the upgrade test harness following the release of ECK 2.16.1

### DIFF
--- a/hack/upgrade-test-harness/testdata/v2161/install.yaml
+++ b/hack/upgrade-test-harness/testdata/v2161/install.yaml
@@ -1,0 +1,797 @@
+# Source: eck-operator/templates/operator-namespace.yaml
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: elastic-system
+  labels:
+    name: elastic-system
+---
+# Source: eck-operator/templates/service-account.yaml
+apiVersion: v1
+kind: ServiceAccount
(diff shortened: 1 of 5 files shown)


hack/upgrade-test-harness/testdata/v2161/install.yaml
```diff
@@ -0,0 +1,797 @@
...
+      labels:
+        control-plane: elastic-operator
+    spec:
+      terminationGracePeriodSeconds: 10
+      serviceAccountName: elastic-operator
+      automountServiceAccountToken: true
+      securityContext:
+        runAsNonRoot: true
+      containers:
+      - image: "docker.elastic.co/eck-dev/eck-operator-mmontgomery:2.16.1-034e7d45"
```
barkbay, line 457: Something looks wrong here.

barkbay (changes requested):

hack/upgrade-test-harness/testdata/v2161/install.yaml
naemono: It appears to have also been this way in the 2.16.0 testdata. I fixed that as well. Thanks!

barkbay (approved):
LGTM

PR #8451, opened 2025-02-01
Check EntSearch version within range

pebrc:
Recent test failures made me realise that the Enterprise search tests should not run at all against 9.0.0 

> 🐞 TestEntSearchAssociation/Creating_Enterprise_Search_should_succeed ~ stack-9-0-0-snap
> 🐞 TestEnterpriseSearchConfigUpdate/Creating_Enterprise_Search_should_succeed ~ stack-9-0-0-snap
> 🐞 TestEnterpriseSearchCrossNSAssociation/Creating_Enterprise_Search_should_succeed ~ stack-9-0-0- snap

### DIFF
--- a/test/e2e/test/enterprisesearch/builder.go
+++ b/test/e2e/test/enterprisesearch/builder.go
@@ -47,7 +47,7 @@ func (b Builder) DeepCopy() *Builder {
 
 var _ test.Builder = Builder{}
 
-// SkipTest returns true if the version is not  in supported range, or if the version is incompatible with Openshift.
+// SkipTest returns true if the version is not in the supported range, or if the version is incompatible with Openshift.
 func (b Builder) SkipTest() bool {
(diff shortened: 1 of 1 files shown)


test/e2e/test/enterprisesearch/builder.go
```diff
@@ -48,11 +47,10 @@ func (b Builder) DeepCopy() *Builder {
 
 var _ test.Builder = Builder{}
 
-// SkipTest returns true if the version is not at least 7.7.0, or if the version is incompatible with Openshift.
+// SkipTest returns true if the version is not  in supported range, or if the version is incompatible with Openshift.
```
thbkrkr, line 50:
```suggestion
// SkipTest returns true if the version is not in the supported range, or if the version is incompatible with Openshift.
```

thbkrkr (approved):

PR #8452, opened 2025-02-02
Move away from deprecated container input

pebrc:
Another test failure in stack 9.0

> 🐞 TestBeatSecureSettings/ES_data_should_pass_validations ~ stack-9-0-0-snap

My theory is that that we are using depracated input that makes this test fail. Testing it out with this PR.


pebrc:
buildkite test this -f p=gke,t=TestBeatSecureSettings -m s=7.17.26,s=8.18.0-SNAPSHOT,s=9.0.0-SNAPSHOT

PR #8460, opened 2025-02-06
Logstash e2e: ssl => ssl_enabled and fix supported versions listing

pebrc:
Between 8.7 and 8.8 `ssl` has been renamed to `ssl_enabled` as of 9.0.0 `ssl` is not longer allowed. ~This removes the attribute from the test config as it is implied by the `https` scheme in the hosts URL.~

This updates the property to `ssl_enabled`. Also we do incorrectly state that we support Logstash 8.7+ while in code we exclude anything before 8.12! https://github.com/elastic/cloud-on-k8s/blob/0548959d68de23b7086a56a98bfa8a3876a62a55/pkg/controller/common/version/version.go#L36

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ Supported versions:
 *  Beats: 7.0+, 8+
 *  Elastic Agent: 7.10+ (standalone), 7.14+, 8+ (Fleet)
 *  Elastic Maps Server: 7.11+, 8+
-*  Logstash 8.7+
+*  Logstash 8.12+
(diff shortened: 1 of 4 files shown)


pebrc:
buildkite test this -f p=kind,t=TestLogstashEsOutput -m s=9.0.0-SNAPSHOT,s=8.12.0,s=8.17.1

barkbay (approved):
lgtm

Issue #8462, opened 2025-02-07
[Doc] Add central pipeline management example to Logstash


kaisecheng:
CPM is a popular feature in Logstash. Customers want guidelines on how to config in ECK Logstash

```yaml
apiVersion: elasticsearch.k8s.elastic.co/v1
kind: Elasticsearch
metadata:
  name: elasticsearch-sample
spec:
  version: 9.0.0-SNAPSHOT
  nodeSets:
    - name: default
      count: 3 
      config:
        node.store.allow_mmap: false
        xpack.license.self_generated.type: trial # for CPM
  monitoring:
    metrics:
      elasticsearchRefs:
        - name: elasticsearch-sample
---
apiVersion: logstash.k8s.elastic.co/v1alpha1
kind: Logstash
metadata:
  name: logstash-sample
spec:
  count: 1
  version: 9.0.0-SNAPSHOT
  elasticsearchRefs:
    - clusterName: production
      name: elasticsearch-sample
  podTemplate:
    spec:
      containers:
        - name: logstash
  config:
    xpack.management.enabled: true
    xpack.management.elasticsearch.hosts: "${PRODUCTION_ES_HOSTS}"
    xpack.management.elasticsearch.username: "${PRODUCTION_ES_USER}"
    xpack.management.elasticsearch.password: "${PRODUCTION_ES_PASSWORD}"
    xpack.management.elasticsearch.ssl.certificate_authority: "${PRODUCTION_ES_SSL_CERTIFICATE_AUTHORITY}"
    xpack.management.pipeline.id: ["*somekeys*"]
  monitoring:
    metrics:
      elasticsearchRefs:
        - name: elasticsearch-sample
    logs:
      elasticsearchRefs:
        - name: elasticsearch-sample
---
apiVersion: kibana.k8s.elastic.co/v1
kind: Kibana
metadata:
  name: kibana-sample
spec:
  version: 9.0.0-SNAPSHOT
  elasticsearchRef:
    name: elasticsearch-sample
  count: 1
  monitoring:
    metrics:
      elasticsearchRefs:
        - name: elasticsearch-sample
```

Issue #7616, opened 2024-03-11
Doc: Break up source file for Logstash ECK docs into smaller files 


karenzone:
## Proposal

**Use case. Why is this important?**
Content for Logstash ECK continues to grow. Having a single source file is becoming increasingly unwieldy.  
Proposal: 
* Move Logstash content to a new folder, and break the monolithic source file into smaller units to align with approach used for Elasticsearch and Security content.
* After restructuring files, make sure that content builds and renders properly with no regressions or loss of content. 

### GitHub history
Ideally, we preserve GitHub attributions and history in the process. I need some guidance on how to approach this.

karenzone:
Go forward plan: 
* Rename original file to preserve history and attributions for the project.
* Manually add contributors to subsequent files.

PR #8471, opened 2025-02-11
Break-up Logstash docs + Add Centralized Pipeline Management documentation

TimBosman:
I did three things in this PR:

- Go from one asciidoc file for Logstash documentation to a asciidoc file for every page
- Replaced some product names with attributes (e.g. `Logstash` -> `{ls}`)
- Added a documentation page for Centralized Pipeline Management

Only thing I'm not completely sure the following section is recommended:

```yaml
xpack.management.elasticsearch.username: "${PRODUCTION_ES_USER}"
xpack.management.elasticsearch.password: "${PRODUCTION_ES_PASSWORD}"
```

If I recall correctly this will setup a connection with the `elastic` user, which would be an escalation of permissions.

Related issues:
- #8462
- #7616


pebrc:
The documentation content is being moved to https://github.com/elastic/docs-content starting with the next release ECK 3.0. I wonder if t would make sense to make these changes there instead.

pebrc:
Closing this as the contents has moved to a different repo.

PR #8472, opened 2025-02-11
Fix diskSizeGb parameter

pebrc:
The type of the parameter is `int64` and the unit is implied in the name.


pebrc:
Screenshot 2025-02-11 at 13 42 07

This seems to work ^^

naemono (approved):
:100:

Issue #8479, opened 2025-02-21
Elastic Stack 9.0 will drop the -ubi suffix


pebrc:
Starting with release 9.0 stack container images will no longer come with a `-ubi` suffix. Instead the default image will be UBI based and there will be wolfi variants with a dedicated suffix. 

We need to: 
* adjust the defaulting to `-ubi` when the `--ubi-only` flag is set
* test the upgrade path on OLM throrogly
* revisit the usefulness of the `--ubi-only` for stack 9.0 (maybe at least log a warnign)

https://github.com/elastic/cloud-on-k8s/blob/3cf131dba124ba9bf56a4861816ee8706960b7c5/cmd/manager/main.go#L516-L519

https://github.com/elastic/cloud-on-k8s/blob/3cf131dba124ba9bf56a4861816ee8706960b7c5/pkg/controller/common/container/container.go#L84-L87

alpar-t:
@pebrc While the default images are becoming UBI based, we are not planning on dropping the `-ubi` suffix, e.g. `elasticsearch` and `elasticsearch-ubi` will both be available and be the exact same image.

pebrc:
Thanks @alpar-t I think what we need to decide then if we should move the new suffixless version when a customer is requesting UBI images with ECK 3.0 so that you can eventually stop producing the ones with the explicit `-ubi` suffix.

pebrc:
@kvalliyurnatt I see you assigned yourself. I think #8480 would be one approach that we could generalise a bit for all images (with exception for the ones that no longer supported in 9.0: enterprisesearch and journalbeat)

barkbay:
@kvalliyurnatt I added the `v3.0.0` label to this one, I believe this has been addressed in https://github.com/elastic/cloud-on-k8s/pull/8509?

barkbay:
@kvalliyurnatt do you know if we can close this? (this is labeled with `3.0.0` and still open)

kvalliyurnatt:
Hey sorry I missed your earlier ping, yes this can be closed.

PR #8480, opened 2025-02-21
feat: update elastic-agent image

pkoutsovasilis:
in 9.x ubi is the default for elastic-agent and beats and although -ubi suffixed images are still getting produced they are practically the same as the default ones.

Also for 9.0 and onwards elastic-agent image repo changes from `beats/elastic-agent` to `elastic-agent/elastic-agent`

Relates https://github.com/elastic/ingest-dev/issues/3987, https://github.com/elastic/cloud-on-k8s/issues/8479


pkg/controller/common/container/container.go
```diff
@@ -88,7 +89,14 @@ func ImageRepository(img Image, ver version.Version) string {
...
 	if useUBISuffix && ver.Major >= 9 {
-		if img == APMServerImage {
+		switch img {
+		case APMServerImage,
+			AgentImageSince9,
+			FilebeatImage,
+			MetricbeatImage,
+			HeartbeatImage,
+			AuditbeatImage,
+			PacketbeatImage:
```
pebrc, line 99: Should this not affect all stack images? Are there ones that keep the `-ubi` suffix?

pkoutsovasilis: I am 100% sure only for elastic-agent and beats 🙂 That said I can't find anything about Journalbeat 

```
$ docker manifest inspect docker.elastic.co/beats/journalbeat:9.0.0-SNAPSHOT
no such manifest: docker.elastic.co/beats/journalbeat:9.0.0-SNAPSHOT
```

pebrc: Journalbeat was removed in 7.16 to my knowledge

pkoutsovasilis: ok that makes sense, then everything I am 100% sure that needs to change is in this PR 🙂

kvalliyurnatt:
@pkoutsovasilis the feature freeze for next release is coming up (March 11th) is there anything blocking from merging this ?

pebrc:
@pkoutsovasilis  we are going with a more radical approach for the 9.0 release and will remove the `-ubi` suffix wholesale. See https://github.com/elastic/cloud-on-k8s/pull/8509

Can you maybe update this PR to only contain the Agent image change so we can merge it?

pkoutsovasilis:
> @pkoutsovasilis we are going with a more radical approach for the 9.0 release and will remove the `-ubi` suffix wholesale. See #8509
> 
> Can you maybe update this PR to only contain the Agent image change so we can merge it?

ok I will adjust it only for Agent and dropping support for `-ubi` images.  ty both @kvalliyurnatt @pebrc

kvalliyurnatt (approved):

PR #8481, opened 2025-02-24
Update docker.elastic.co/ci-agent-images/serverless-docker-builder to 0.4.0

barkbay:
Update docker-builder image.

### DIFF
--- a/.buildkite/scripts/build/pre-build-operator.sh
+++ b/.buildkite/scripts/build/pre-build-operator.sh
@@ -9,7 +9,7 @@
 
 set -euo pipefail
 
-ROOT="$(cd "$(dirname "$0")"; pwd)/../../.."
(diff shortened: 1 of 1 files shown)


.buildkite/scripts/build/pre-build-operator.sh
```diff
@@ -9,7 +9,7 @@
 
 set -euo pipefail
 
-ROOT="$(cd "$(dirname "$0")"; pwd)/../../.."
+ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")"; pwd)/../../.."
```
barkbay, line 12: This is required because the script is now "sourced", therefore `dirname "$0"` is the directory from which `pre-build-operator.sh` is sourced, not the directory of the script itself.

pebrc (approved):

Issue #8498, opened 2025-02-28
Adjust docs tooling for docs v3


pebrc:
- [x] 1.  disable the master/main docs build in favour of the new CI checks merged on https://github.com/elastic/cloud-on-k8s/pull/8493 (done in https://github.com/elastic/docs/pull/3176)

- [x] 2. adjust the repository tooling (which will otherwise fail all builds)

- [x] 3. merge this PR.

_Originally posted by @eedugon in https://github.com/elastic/cloud-on-k8s/issues/8496#issuecomment-2688562939_

eedugon:
@pebrc , have we finished all tasks of this issue?

PR #8496, opened 2025-02-27
[docs] Migrate docs from AsciiDoc to Markdown

eedugon:
Migrates docs from AsciiDoc to Markdown

This PR should substitute https://github.com/elastic/cloud-on-k8s/pull/8494 created by @colleenmcginnis (I didn't have permissions to merge the latest changes in main, and I might need to add a few extra changes to the content).

 > Note: When you get the build working and click the View deployment button, you'll need to add /reference to the end of the URL you're sent to in order to see a rendered page.

PR #8501, opened 2025-02-28
Migrate API reference to markdown

pebrc:
Part of #8498

https://docs-v3-preview.elastic.dev/elastic/cloud-on-k8s/pull/8501/reference/

Adjust the templating for the API reference. Markdown was modelled after https://github.com/elastic/cloud-on-k8s/pull/8496

Differences to the manually migrated files: API reference tool creates a single file not a file per group, we could however adjust that if needed.


hack/api-docs/templates/gv_list.tpl
```diff
@@ -1,22 +1,20 @@
 {{- define "gvList" -}}
 {{- $groupVersions := . -}}
+---
+mapped_pages:
+  - https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-api-reference.html
+navigation_title: API reference
```
barkbay, line 6: nit, for consistency?

```suggestion
navigation_title: API Reference
```

(not 100% sure)

barkbay (approved):
LGTM

eedugon:
@pebrc :

> Differences to the manually migrated files: API reference tool creates a single file not a file per group, we could however adjust that if needed.

I believe that's totally up to you to decide.

In the `asciidoc` system ([here](https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-api-reference.html)) they appear 
 to be separate documents, that's why our automatic migration created different markdown files for each item.

If you think it's better to show all the content in the same page (like k8s api docs) I'm almost sure that's totally fine.

pebrc:
> @pebrc :
> 
> > Differences to the manually migrated files: API reference tool creates a single file not a file per group, we could however adjust that if needed.
> 
> I believe that's totally up to you to decide.
> 
> In the `asciidoc` system ([here](https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-api-reference.html)) they appear to be separate documents, that's why our automatic migration created different markdown files for each item.
> 
> If you think it's better to show all the content in the same page (like k8s api docs) I'm almost sure that's totally fine.

I spent a bit more time on breaking up the single file. The tooling we have makes it hard to generate the grouping into files the way the migration has created them. I think I stick with a single file if that's OK for now. If we have more time to optimise this later we can think about breaking it up into single files per API kind.

PR #8507, opened 2025-03-03
Remove support for 6.x stack version

kvalliyurnatt:
With the Release of ECK 3.0.0 and Elasticsearch 9.0.0 we will no longer support 6.x versions. Removing support for this in the code


README.md
```diff
@@ -19,7 +19,7 @@ Supported versions:
 
 *  Kubernetes 1.28-1.32
 *  OpenShift 4.13-4.18
-*  Elasticsearch, Kibana, APM Server: 6.8+, 7.1+, 8+
+*  Elasticsearch, Kibana, APM Server: 7.1+, 7.1+, 8+
```
kvalliyurnatt, line 22: should this be 7.17 ?

naemono: Aren't we both removing support for 6.x, and adding support for 9? I suspect that we'll be supporting just 7.17, yes.

naemono, line 22:
```suggestion
*  Elasticsearch, Kibana, APM Server: 7.17, 8+
```

kvalliyurnatt: Thanks yes you are correct. Would there be a version of 8 we will be supporting or would be just 8+ ?

naemono: I think we support all versions of 8 in this scenario, similar to how we supported all versions of 7 previously.

pkg/controller/common/version/version.go
```diff
@@ -25,7 +25,7 @@ var GlobalMinStackVersion Version
 var (
 	SupportedAPMServerVersions        = MinMaxVersion{Min: From(6, 2, 0), Max: From(9, 99, 99)}
 	SupportedEnterpriseSearchVersions = MinMaxVersion{Min: From(7, 7, 0), Max: From(8, 99, 99)}
```
barkbay, line 27: If `7.17.x` is the new min. supported version should we also update all of this?
```suggestion
	SupportedAPMServerVersions        = MinMaxVersion{Min: From(7, 17, 0), Max: From(9, 99, 99)}
	SupportedEnterpriseSearchVersions = MinMaxVersion{Min: From(7, 17, 0), Max: From(8, 99, 99)}
```

README.md
kvalliyurnatt: spoke with Peter and he mentioned that while the official supported version will be 7.17.x we should be a little lenient in the operator and allow any 7.x versions, I will make changes for this

kvalliyurnatt:
> Do we also want to remove the code that will no longer be invoked? [Like using only the shutdown API?](https://github.com/elastic/cloud-on-k8s/blob/34797314eeded5694caec31cfa1a13c95713db25/pkg/controller/elasticsearch/shutdown/interface.go#L14)

I am going to do this separately in a different PR, we at least want to block users from creating 6.x versions of the stack for this release, Peter mentioned we can do the code clean up even after the release

README.md
```diff
@@ -19,7 +19,7 @@ Supported versions:
 
 *  Kubernetes 1.28-1.32
 *  OpenShift 4.13-4.18
-*  Elasticsearch, Kibana, APM Server: 6.8+, 7.1+, 8+
+*  Elasticsearch, Kibana, APM Server: 7.0+, 7.1+, 8+
```
kvalliyurnatt, line 22: each stack application is at a different version, should we standardize to a specific minimum version ?

pebrc: I think the README should say 7.17 which is in line with our EOL policy

pkg/controller/common/version/version.go
barkbay: I think this is expected given that we have gradually added support for them over time.

hack/operatorhub/templates/csv.tpl
```diff
@@ -290,7 +290,7 @@ spec:
 
     * Google Kubernetes Engine (GKE), Azure Kubernetes Service (AKS), and Amazon Elastic Kubernetes Service (EKS)
 
-    * Elasticsearch, Kibana, APM Server: {{ if .UbiOnly }}7.10+{{ else }}6.8+, 7.1+{{ end }}, 8+
+    * Elasticsearch, Kibana, APM Server: {{ if .UbiOnly }}7.17+{{ else }}7.0+, 7.1+{{ end }}, 8+
```
pebrc, line 293:
```suggestion
    * Elasticsearch, Kibana, APM Server: 7.17+, 8+
```

README.md
pebrc: The user facing docs should still say 7.17.x and we should even add a validation warning if someone is submitting a lower version in a manifest. My argument for leniency was more along the lines of avoiding a hard rejection if a user missed our communications about the EOL policy and accidentally upgraded to ECK 3.0 before upgrading their stack.

kvalliyurnatt: I have added in a deprecation warning for the stacks

pkg/apis/common/v1/validations.go
```diff
@@ -76,6 +76,20 @@ func CheckSupportedStackVersion(ver string, supported version.MinMaxVersion) fie
...
 }
 
+// CheckDeprecatedStackVersion checks that the given version is not deprecated.
+func CheckDeprecatedStackVersion(ver string, deprecatedVersion version.MinMaxVersion) (string, field.ErrorList) {
+	v, err := ParseVersion(ver)
+	if err != nil {
+		return "", err
+	}
+
+	if v.Major >= deprecatedVersion.Min.Major && v.Major <= deprecatedVersion.Max.Major && v.Minor >= deprecatedVersion.Min.Minor && v.Minor <= deprecatedVersion.Max.Minor {
```
barkbay, line 86:
```suggestion
	if v.GTE(version.DeprecatedVersions.Min) && v.LT(version.DeprecatedVersions.Max) {
```

pkg/apis/common/v1/validations.go
```diff
@@ -76,6 +76,20 @@ func CheckSupportedStackVersion(ver string, supported version.MinMaxVersion) fie
 	return nil
 }
 
+// CheckDeprecatedStackVersion checks that the given version is not deprecated.
+func CheckDeprecatedStackVersion(ver string, deprecatedVersion version.MinMaxVersion) (string, field.ErrorList) {
```
barkbay, line 80: It seems that the second parameter is always `version.DeprecatedVersions`?
```suggestion
func CheckDeprecatedStackVersion(ver string) (string, field.ErrorList) {
```

pkg/apis/agent/v1alpha1/webhook.go
```diff
@@ -95,6 +95,15 @@ func (a *Agent) validate(old *Agent) (admission.Warnings, error) {
 		}
 	}
 
+	// depreciation check
```
barkbay, line 98: Should we move this at the beginning of the function to ensure that the warning is there even if one of the `updateChecks` has failed?

pkg/controller/common/version/version.go
```diff
@@ -23,9 +23,12 @@ var GlobalMinStackVersion Version
 
 // supported Stack versions. See https://www.elastic.co/support/matrix#matrix_compatibility
 var (
-	SupportedAPMServerVersions        = MinMaxVersion{Min: From(6, 2, 0), Max: From(9, 99, 99)}
+	DeprecatedVersions = MinMaxVersion{Min: From(7, 0, 0), Max: From(7, 16, 0)}
+	// Some of the versions below might be EOL but we do not want to block users from usign ECK with those versions
```
barkbay, line 27: nit, typo:
```suggestion
	// Some of the versions below might be EOL but we do not want to block users from using ECK with those versions.
```

pkg/controller/elasticsearch/validation/validations.go
```diff
@@ -98,14 +98,20 @@ func validNodeLabels(proposed esv1.Elasticsearch, exposedNodeLabels NodeLabels)
...
 	var errs field.ErrorList
 	for _, val := range validations {
 		if err := val(es); err != nil {
 			errs = append(errs, err...)
 		}
 	}
-	return errs
+
+	// check if Elasticsearch version is deprecated
+	warnings, deprecatedErrors := checkDeprecatedVersion(&es)
```
barkbay, line 110: I think we can remove `checkDeprecatedVersion(...)`?
```suggestion
warnings, deprecatedErrors := commonv1.CheckDeprecatedStackVersion(es.Spec.Version, version.DeprecatedVersions)
```

pkg/utils/test/webhook.go
```diff
@@ -74,6 +74,26 @@ func ValidationWebhookFailed(causeRegexes ...string) func(*testing.T, *admission
 	}
 }
 
+func ValidationWebhookSuceededWithWarnings(warningsRegexes ...string) func(*testing.T, *admissionv1beta1.AdmissionResponse) {
```
barkbay, line 77:
```suggestion
func ValidationWebhookSucceededWithWarnings(warningsRegexes ...string) func(*testing.T, *admissionv1beta1.AdmissionResponse) {
```

pkg/utils/test/webhook.go
```diff
@@ -74,6 +74,26 @@ func ValidationWebhookFailed(causeRegexes ...string) func(*testing.T, *admission
...
+			t.Logf("Checking for existence of: %s", wr)
+			for _, warning := range response.Warnings {
+				match, err := regexp.MatchString(wr, warning)
+				require.NoError(t, err, "Match '%s' returned error: %v", wr, err)
+				if match {
+					found = true
+					break
+				}
+			}
+			require.True(t, found, "[%s] is not present in cause list", wr)
```
barkbay, line 92:
```suggestion
			require.True(t, found, "[%s] is not present in warning list", wr)
```

barkbay (approved):
LGTM

pkg/apis/apm/v1beta1/webhook_test.go
```diff
@@ -122,32 +122,45 @@ func TestWebhook(t *testing.T) {
...
+			Name:      "deprecated version",
+			Operation: admissionv1beta1.Create,
+			Object: func(t *testing.T, uid string) []byte {
+				t.Helper()
+				apm := mkApmServer(uid)
+				apm.Spec.Version = "7.4.0"
+				return serialize(t, apm)
+			},
+			Check: test.ValidationWebhookSuceededWithWarnings(
+				`Version 7.4.0 is EOL and will be removed in a future release of the ECK operator`,
```
pebrc, line 148:
```suggestion
				`Version 7.4.0 is EOL and support for it will be removed in a future release of the ECK operator`,
```

pkg/apis/common/v1/validations.go
```diff
@@ -76,6 +76,20 @@ func CheckSupportedStackVersion(ver string, supported version.MinMaxVersion) fie
...
+	v, err := ParseVersion(ver)
+	if err != nil {
+		return "", err
+	}
+
+	if v.Major >= deprecatedVersion.Min.Major && v.Major <= deprecatedVersion.Max.Major && v.Minor >= deprecatedVersion.Min.Minor && v.Minor <= deprecatedVersion.Max.Minor {
+		return fmt.Sprintf("Version %s is EOL and will be removed in a future release of the ECK operator", ver), nil
```
pebrc, line 87:
```suggestion
		return fmt.Sprintf("Version %s is EOL and support for it will be removed in a future release of the ECK operator", ver), nil
```

pkg/controller/common/version/version.go
pebrc, line 27:
```suggestion
	// Some of the versions below might be EOL but we do not want to block users from using ECK with those versions yet.
```

pebrc (approved):

pkg/apis/agent/v1alpha1/webhook.go
kvalliyurnatt: good point, done

PR #8509, opened 2025-03-06
Remove ubi suffix for 9.x images

kvalliyurnatt:
With 9.x release all images should be based off of ubi, so we should no longer need the -ubi suffix

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -322,7 +322,7 @@ func Command() *cobra.Command {
 	cmd.Flags().Bool(
 		operator.UBIOnlyFlag,
 		false,
-		fmt.Sprintf("Use only UBI container images to deploy Elastic Stack applications. UBI images are only available from 7.10.0 onward. Cannot be combined with %s", operator.ContainerSuffixFlag),
+		fmt.Sprintf("Use only UBI container images to deploy Elastic Stack applications. UBI images are only available from 7.10.0 onward. Ignored from 9.x as default images are based on UBI. Cannot be combined with %s", operator.ContainerSuffixFlag),
 	)
 	cmd.Flags().Bool(
 		operator.ValidateStorageClassFlag,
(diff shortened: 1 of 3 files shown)


barkbay (commented):
Change technically LGTM but should we document that it is intended that the suffix will no longer be set for 9.x?

Either here:

```diff
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -322,7 +322,7 @@ func Command() *cobra.Command {
        cmd.Flags().Bool(
                operator.UBIOnlyFlag,
                false,
-               fmt.Sprintf("Use only UBI container images to deploy Elastic Stack applications. UBI images are only available from 7.10.0 onward. Cannot be combined with %s", operator.ContainerSuffixFlag),
+               fmt.Sprintf("Use only UBI container images to deploy Elastic Stack applications. UBI images are only available from 7.10.0 onward. Ignored from 9.x as default images are based on UBI. Cannot be combined with %s", operator.ContainerSuffixFlag),
        )
        cmd.Flags().Bool(
                operator.ValidateStorageClassFlag,
```

Or in the documentation?

pebrc:
> Change technically LGTM but should we document that it is intended that the suffix will no longer be set for 9.x?
> 
> Either here:
> 
> ```diff
> --- a/cmd/manager/main.go
> +++ b/cmd/manager/main.go
> @@ -322,7 +322,7 @@ func Command() *cobra.Command {
>         cmd.Flags().Bool(
>                 operator.UBIOnlyFlag,
>                 false,
> -               fmt.Sprintf("Use only UBI container images to deploy Elastic Stack applications. UBI images are only available from 7.10.0 onward. Cannot be combined with %s", operator.ContainerSuffixFlag),
> +               fmt.Sprintf("Use only UBI container images to deploy Elastic Stack applications. UBI images are only available from 7.10.0 onward. Ignored from 9.x as default images are based on UBI. Cannot be combined with %s", operator.ContainerSuffixFlag),
>         )
>         cmd.Flags().Bool(
>                 operator.ValidateStorageClassFlag,
> ```
> 
> Or in the documentation?

+1 to that. I think if you add it as suggested by Michael and here https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-operator-config.html we should be good.

barkbay (approved):
LGTM

Issue #8499, opened 2025-02-28
Elastic operator incorrectly reconciles Kibana config into secret


Fiki80:
We are deploying Elastic stack via cloud-on-k8s helmcharts. If we use the "simplified" object syntax to configure xpack fleet agent policy in Kibana values, then Kibana config secret is incorrectly interpreted.

This is a part of Kibana object in yaml:
```
apiVersion: kibana.k8s.elastic.co/v1
kind: Kibana
metadata:
  annotations:
    association.k8s.elastic.co/es-conf: xxx
    eck.k8s.elastic.co/license: basic
  generation: 34
  labels:
    app.kubernetes.io/instance: ecn-monitoring
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/name: eck-kibana
    argocd.argoproj.io/instance: ecn-monitoring
    helm.sh/chart: eck-kibana-0.14.1
  name: ecn-monitoring-eck-kibana
  namespace: sdlc-monitoring
spec:
  config:
    server:
      publicBaseUrl: https://xxx.xxx
    xpack:
      fleet:
        agentPolicies:
        - id: eck-external-agent-policy
          monitoring_enabled:
          - logs
          - metrics
          name: Elastic Agent on ECK policy - external
          namespace: extern
          package_policies:
          - id: kubernetes-3
            name: kubernetes-3
            package:
              name: kubernetes
            inputs:
              kube-proxy-kubernetes/metrics:
                enabled: true
                streams:
                  '[kubernetes.proxy]':
                    enabled: true
                    vars:
                      hosts:
                      - ecn-monitoring-kube-promet-kube-proxy.kube-system:10249
                vars: {}
          unenroll_timeout: 900
```

And this is how Kibana configuration secret is created by the operator:
```
xpack:
    fleet:
        agentPolicies:
            - id: eck-external-agent-policy
              monitoring_enabled:
                - logs
                - metrics
              name: Elastic Agent on ECK policy - external
              namespace: extern
              package_policies:
                - id: kubernetes-3
                  inputs:
                    kube-proxy-kubernetes/metrics:
                        enabled: true
                        streams:
                            '[kubernetes':
                                proxy]:
                                    enabled: true
                                    vars:
                                        hosts:
                                            - ecn-monitoring-kube-promet-kube-proxy.kube-system:10249
                        vars: null
              unenroll_timeout: 900
```

We can see 2 issues with the created secret. 

[...]

PR #8512, opened 2025-03-07
Config: Allow escaping dots in keys via '[unsplit.key]' syntax

barkbay:
This PR allows users to escape dots in paths, similar to what Kibana supports since https://github.com/elastic/kibana/pull/178841

Relates to https://github.com/elastic/cloud-on-k8s/issues/8499


barkbay:
buildkite test this -f p=gke,s=8.17.3,t=TestFleet

pkg/controller/common/settings/canonical_config.go
```diff
@@ -22,7 +22,7 @@ import (
 type CanonicalConfig ucfg.Config
 
 // Options are config options for the YAML file. Currently contains only support for dotted keys.
-var Options = []ucfg.Option{ucfg.PathSep("."), ucfg.AppendValues}
+var Options = []ucfg.Option{ucfg.PathSep("."), ucfg.AppendValues, ucfg.EscapePath()}
```
pebrc, line 25: Should this generally be enabed or only in Kibana that supports the escape syntax?

barkbay: I'm not sure how to make this specific to Kibana tbh, this `Options` is used in a lot of places, including receivers like `MergeWith` or `HasKeys` ... I can change all these functions so they accept an "optional `Options`". On the other hand I think it should be fine to enable this for other applications, even if they are not compatible with this syntax?

pebrc: Maybe you are right.

PR #8521, opened 2025-03-10
update the net package

kvalliyurnatt:
Update the net library to 0.37.0


pebrc (approved):

pebrc:
Why do you want to exclude this from the release notes? I would label it as `>renovate` so that it shows up as a dependecy upgrade in the release notes.

PR #8532, opened 2025-03-11
update operator and stack version

kvalliyurnatt:
Update stack version to 9.0.0 and operator version to 3.0.0

### DIFF
--- a/pkg/apis/agent/v1alpha1/agent_types.go
+++ b/pkg/apis/agent/v1alpha1/agent_types.go
@@ -24,10 +24,6 @@ const (
 	FleetServerServiceAccount commonv1.ServiceAccountName = "fleet-server"
 )
 
-var (
-	MandatoryPolicyIDVersion = version.MustParse("9.0.0")
-)
-
 // AgentSpec defines the desired state of the Agent
 type AgentSpec struct {
 	// Version of the Agent.
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ Current features:
 Supported versions:
 
 *  Kubernetes 1.28-1.32
-*  OpenShift 4.13-4.18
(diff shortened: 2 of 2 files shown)


hack/operatorhub/config.yaml
```diff
@@ -41,7 +41,7 @@ packages:
   - outputPath: certified-operators
     packageName: elasticsearch-eck-operator-certified
     distributionChannel: certified-operators
-    minSupportedOpenshiftVersion: v4.12
+    minSupportedOpenshiftVersion: v4.14
```
kvalliyurnatt, line 44: 4.12 and 4.13 seem to be EOL however 4.12 still seems to have some sort of support
https://access.redhat.com/product-life-cycles?product=OpenShift%20Container%20Platform%204

kvalliyurnatt: I need to update the README, but I thought I will update that main and do a backport. But if it is okay to just do it directly here I can do that as well

kvalliyurnatt: Spoke with Peter and we will set the min version to 4.14 according our policy of supporting N-4 versions, Peter also confirmed that there were only 10 clusters running 4.12.

pebrc (approved):

pkg/apis/agent/v1alpha1/agent_types.go
```diff
@@ -25,7 +25,7 @@ const (
 )
 
 var (
-	MandatoryPolicyIDVersion = version.MustParse("9.0.0-SNAPSHOT")
+	MandatoryPolicyIDVersion = version.MustParse("9.0.0")
```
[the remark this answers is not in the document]
kvalliyurnatt: removed it

PR #8542, opened 2025-03-17
Fix enterprise search version in chart and other stack versions in readme and operator hub

kvalliyurnatt:
Enterprise search  does not have a 9.X version, updating charts and also fixing README to add the correct supported versions

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,12 +19,12 @@ Supported versions:
 
 *  Kubernetes 1.28-1.32
 *  OpenShift 4.14-4.18
-*  Elasticsearch, Kibana, APM Server: 7.17+, 8+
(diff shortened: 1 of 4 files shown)


barkbay:
PR's title is `fix enterprise search version in chart` but is it not actually fixing the versions for the other components?

kvalliyurnatt:
> PR's title is `fix enterprise search version in chart` but is it not actually fixing the versions for the other components?

updated the title

kvalliyurnatt:
> Let's maybe also update the operator hub versions:
> 
> https://github.com/elastic/cloud-on-k8s/blob/88f069153c127a9065068b987b2da3920223ab60/hack/operatorhub/templates/csv.tpl#L293-L303

Thanks! fixed this

barkbay (approved):

PR #8587, opened 2025-03-28
Update aks cluster

barkbay:
This PR updates the AKS cluster version and disables the http_application_routing add-on (it's [deprecated](https://learn.microsoft.com/en-us/azure/aks/app-routing-migration) and I'm not sure we actually use it)


barkbay:
buildkite test this -f t=TestSmoke -m p=aks

pebrc (approved):

Issue #916, opened 2019-05-23
Ready checks are incompatible with Service usage and PDB usage


nkvoll:
Ready checks are currently used to decide whether a `Service` should include the pods, so pods are considered ready as soon as they can respond to HTTP requests.

Once we start to use `PodDisruptionBudget`s, readiness means that the pod is not considered disrupted. In this scenario, the safest bet is cluster state: green. Yellow might be tolerated in some scenarios (e.g 1 node cluster), but it seems this should be configurable by the user.

These two are not immediately compatible, but since readiness is the only way to indicate to the k8s admins that the cluster is disrupted, this takes precedence.

To facilitate this, we can tolerate unready pods in the service. This causes a problem: requests may be forwarded to pods that are not ready to receive requests at all. To fix this, we can use some kind of new `ready` label on the pods in the Service selector. This label may be controlled by the operator (e.g pinging individual nodes every X seconds), effectively moving the service-related readiness probe to the operator and separating it from the PDB-related readiness check.

Relates to: https://github.com/elastic/cloud-on-k8s/issues/914

pebrc:
> This label may be controlled by the operator (e.g pinging individual nodes every X seconds)

One disadvantage of this approach is that unavailable pods might be listed as endpoints of the service and available pods might be missing from the endpoints if the operator is not running. The current implementation allows clusters to run (and fail) independently from the failure domain of the operator (unless structural changes are being applied to the cluster via the operator, in which case they are somewhat tied together)

anyasabo:
Just to be sure I understand the concerns correctly:

Current ready check:
    Does the pod respond to http requests

Both the pod disruption budget and service use the pod's ready status to decide if pod is disrupted *and* if the pod is ready to receive requests.

So if the pod gets disrupted, it will get marked as ready as soon as that pod is up and responding to requests. This might cause a problem because the cluster may still be yellow and be recovering (say it is replicating from a primary to get back in sync since it was last up), but from the pod disruption budget it thinks everything is fine and may disrupt the pod hosting the primary shard, which will make the cluster even sadder.

If we change the readiness check to both:
- the pod responds to http requests
- the cluster is green

Then essentially one pod going down (which would turn the cluster yellow), marks the whole cluster as not ready and removes them all from the service, which is Not Good.


Would it maybe make sense to mark a node as not ready if it is a target node for any active recoveries? I think that might address this in a more idiomatic way unless there's other reasons for not readiness I might be missing.

----

[...]

PR #8590, opened 2025-04-03
Use MaxUnavailable for PDB

mrkm4ntr:
We should use MaxUnavailable for simplicity.

```
// MaxUnavailable can only be used if the selector matches a builtin controller selector
// (eg. Deployments, StatefulSets, etc.). We cannot use it with our own cluster-name selector.
```

This comment was written by [this reason](https://github.com/elastic/cloud-on-k8s/issues/916#issuecomment-533135010).  According to this, maxUnavailable can be used only when selected pods are managed by only 1 StatefulSet. But it's not true. We can use maxUnavailable for the case multiple StatefulSets are managing selected pods. See [here](https://github.com/kubernetes/kubernetes/blob/v1.32.3/pkg/controller/disruption/disruption.go#L846-L850).

If this change is valid, I will change the document as well.

### DIFF
--- a/pkg/controller/elasticsearch/pdb/reconcile.go
+++ b/pkg/controller/elasticsearch/pdb/reconcile.go
@@ -162,12 +162,7 @@ func expectedPDB(es esv1.Elasticsearch, statefulSets sset.StatefulSetList) (*pol
 // buildPDBSpec returns a PDBSpec computed from the current StatefulSets,
 // considering the cluster health and topology.
 func buildPDBSpec(es esv1.Elasticsearch, statefulSets sset.StatefulSetList) policyv1.PodDisruptionBudgetSpec {
-	// compute MinAvailable based on the maximum number of Pods we're supposed to have
-	nodeCount := statefulSets.ExpectedNodeCount()
-	// maybe allow some Pods to be disrupted
-	minAvailable := nodeCount - allowedDisruptions(es, statefulSets)
-
-	minAvailableIntStr := intstr.IntOrString{Type: intstr.Int, IntVal: minAvailable}
+	maxUnavailableIntStr := intstr.IntOrString{Type: intstr.Int, IntVal: allowedDisruptions(es, statefulSets)}
 
 	return policyv1.PodDisruptionBudgetSpec{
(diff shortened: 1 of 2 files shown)


barkbay:
Interesting, thanks for your suggestion! While I think you're right, I'm a bit on the fence with that change for the following reasons:
* I'm wondering if we want to keep that idea that Elasticsearch Pods are actually controlled by the ECK operator, and using `StatefulSets` should be an implementation detail which may change in the future. 
* Do we know if that case (1 pdb with several built-in controllers) is covered by some unit tests in the pdb controller? You mentioned a comment, but I would like to know if we have any guarantee that this logic is not going to change (The K8s documentation is not really clear, so I'm trying to understand if we are not in a corner case that may no longer be supported at some point).

mrkm4ntr:
> I'm wondering if we want to keep that idea that Elasticsearch Pods are actually controlled by the ECK operator, and using StatefulSets should be an implementation detail which may change in the future.

If ECK will stop to use StatefulSet, I think we will create new CRD for managing each NodeSet like current StatefulSet does. If  so, that CRD can have scale subresource then we can use maxUnavailable.

> Do we know if that case (1 pdb with several built-in controllers) is covered by some unit tests in the pdb controller? You mentioned a comment, but I would like to know if we have any guarantee that this logic is not going to change (The K8s documentation is not really clear, so I'm trying to understand if we are not in a corner case that may no longer be supported at some point).

Unfortunately no. But I feel current implementation is really straightforward. I hope they won't introduce a breaking change in this...

pebrc:
Hi @mrkm4ntr,

Thank you for this contribution and for pointing out that the original limitation around `maxUnavailable` was overly conservative! Your analysis of the Kubernetes disruption controller code was correct.

However, this PR has been superseded by #8780, which was merged in September 2025 as part of ECK v3.2.0. That PR introduced granular PodDisruptionBudgets based on node roles, and as part of that implementation, we switched to using `maxUnavailable` — the same change you proposed here.

Since the core improvement you identified is now part of the codebase, I'm going to close this PR. We appreciate you taking the time to investigate and contribute!

PR #8606, opened 2025-04-15
Add logs volume for file/metricbeat when stack monitoring is enabled with readOnlyRootFilesystem.

### LINKED ISSUE #8605: File/Metricbeat need logs volume when readOnlyRootFilesystem is enabled with stack monitoring.
When `readOnlyRootFilesystem` is enabled for Elasticsearch or Kibana and stack monitoring is enabled both filebeat and metricbeat need an `emptyDir` volume to write logs otherwise they receive an error when attempting to write any logs to their logs directory as can happen when indexing errors occur, and a new event file is attempted to be written:

```
write error: failed to open new log file for writing: failed to open new file '/usr/share/filebeat/logs/filebeat-events-data-20250328.ndjson': open /usr/share/filebeat/logs/filebeat-events-data-20250328.ndjson: read-only file system
```

naemono:
resolves #8605

When `readOnlyRootFilesystem` is enabled for Elasticsearch or Kibana and stack monitoring is enabled both filebeat and metricbeat need an `emptyDir` volume to write logs otherwise they receive an error when attempting to write any logs to their logs directory as can happen when indexing errors occur, and a new event file is attempted to be written.

### DIFF
--- a/pkg/controller/elasticsearch/nodespec/podspec.go
+++ b/pkg/controller/elasticsearch/nodespec/podspec.go
@@ -134,7 +134,7 @@ func BuildPodTemplateSpec(
 		WithContainersSecurityContext(securitycontext.For(ver, enableReadOnlyRootFilesystem)).
 		WithPreStopHook(*NewPreStopHook())
 
-	builder, err = stackmon.WithMonitoring(ctx, client, builder, es)
+	builder, err = stackmon.WithMonitoring(ctx, client, builder, es, enableReadOnlyRootFilesystem)
 	if err != nil {
 		return corev1.PodTemplateSpec{}, err
 	}
--- a/pkg/controller/elasticsearch/stackmon/__snapshots__/sidecar_test.snap
+++ b/pkg/controller/elasticsearch/stackmon/__snapshots__/sidecar_test.snap
@@ -903,3 +903,240 @@ setup.template.settings:
  }
 }
 ---
+
+[TestWithMonitoring/with_metrics_and_logs_monitoring_and_read_only_root_filesystem - 1]
+{
+ "metadata": {
+  "annotations": {
+   "elasticsearch.k8s.elastic.co/monitoring-config-hash": "1805099278"
+  },
+  "creationTimestamp": null
+ },
+ "spec": {
+  "automountServiceAccountToken": false,
+  "containers": [
+   {
+    "env": [
+     {
+      "name": "ES_LOG_STYLE",
+      "value": "file"
+     }
+    ],
+    "name": "elasticsearch",
+    "resources": {}
+   },
+   {
+    "args": [
+     "-c",
+     "/etc/metricbeat-config/metricbeat.yml",
+     "-e"
+    ],
+    "env": [
+     {
+      "name": "POD_IP",
+      "valueFrom": {
+       "fieldRef": {
+        "apiVersion": "v1",
+        "fieldPath": "status.podIP"
+       }
+      }
+     },
+     {
+      "name": "POD_NAME",
+      "valueFrom": {
+       "fieldRef": {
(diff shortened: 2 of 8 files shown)


naemono:
> Change looks good. I wonder if we should do something for Beats and Logstash as well. We do not configure a security context on those but if a user does we should be able to detect that and configure the additional volumes for stack monitoring. Wdyt?

Ok, this has been added for logstash as well, if any containers are found within the pod are found to have readOnlyRootFilesystem set.

pkg/controller/elasticsearch/nodespec/podspec.go
```diff
@@ -134,7 +134,7 @@ func BuildPodTemplateSpec(
 		WithContainersSecurityContext(securitycontext.For(ver, enableReadOnlyRootFilesystem)).
 		WithPreStopHook(*NewPreStopHook())
 
-	builder, err = stackmon.WithMonitoring(ctx, client, builder, es)
+	builder, err = stackmon.WithMonitoring(ctx, client, builder, es, enableReadOnlyRootFilesystem)
```
barkbay, line 137: nit: The way we decide to include those empty volumes is not the same between Elasticsearch and Logstash. For Logstash we try to infer that decision using `isReadOnlyRootFilesystem(...)` on the pod spec. While for Elasticsearch we are only using our custom logic, however `enableReadOnlyRootFilesystem` does not mean that the file system is not going to be "read-only". Maybe this has been enforced by the user in the ES pod template.
Should we be consistent and use `isReadOnlyRootFilesystem(...)` in all the controllers? (or just add the empty volumes by default as suggested in my previous comment)

pebrc: > or just add the empty volumes by default as suggested in my previous comment

+1 to that proposal. It removes all the conditional logic and makes the code simpler and it is never a good idea to write into a containers filesystem no matter what the security settings are.

naemono:
> For logs I just realized that in the case of Beats both the main container and the side car share the same `/usr/share/filebeat/logs`:

I have created #8622 to address the duplicate volume mount issue.

pebrc:
@naemono I think you are still missing the volume for metricbeat logs in the case of beats monitoring.

naemono:
> @naemono I think you are still missing the volume for metricbeat logs in the case of beats monitoring.

This has been resolved. Thanks for noting this.

pebrc:
buildkite test this -f t=Test.*StackMonitoring -m p=gke,p=ocp

pkg/controller/beat/common/stackmon/stackmon.go
```diff
@@ -103,6 +103,19 @@ func MetricBeat(ctx context.Context, client k8s.Client, beat *v1beta1.Beat) (sta
 		},
 	})
 
+	// Add shared volume for logs consumption.
+	sidecar.Container.VolumeMounts = append(sidecar.Container.VolumeMounts, corev1.VolumeMount{
+		Name:      "metricbeat-logs",
```
pebrc, line 108: `const` ?

pebrc: Any reason not use  `volume.NewEmptyDirVolume("metricbeat-logs", "/usr/share/metricbeat/logs")`

naemono: I added a constant to `beat/common/stackmon` and updated all controllers to use this for consistency, and updated all to use NewEmptyDirVolume.

pebrc (approved):
LGTM

pebrc:
This was not labelled with the target release and is missing from the release notes. Also this triggeres a restart of the affected pods which we also do not document at the moment.

PR #8607, opened 2025-04-15
update testdata for upgrade harness test to include the latest ECK version

kvalliyurnatt:
As part of the post release work updating the testdata for the upgrade harness test.
The update is a little different than usual since we have a limitation that we have to first upgrade to 8.18.0 stack version before upgrading to 9.0.0 version, i have added two folders `v300` and `v300-stack9`. Additionally we do not have Enterprise search in 9.0.0, so I have not included that in `v300-stack9` stack.yaml.


barkbay (approved):
LGTM

PR #8614, opened 2025-04-17
add pre-release flag to container push command in operatorhub

kvalliyurnatt:
For build candidates we now push the operator image to  `docker.elastic.co/eck-snapshots`. Adding an option in the `hack/operatorhub` tool to be able to pull images from that repo to test before a release

### DIFF
--- a/hack/operatorhub/cmd/container/container.go
+++ b/hack/operatorhub/cmd/container/container.go
@@ -68,6 +68,14 @@ func Command() *cobra.Command {
 		"force will force the attempted pushing of remote images, even when the exact version is found remotely. (OHUB_FORCE)",
 	)
 
+	cmd.PersistentFlags().BoolVarP(
+		&flags.PreRelease,
+		flags.PreReleaseFlag,
+		"P",
+		false,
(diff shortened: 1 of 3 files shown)


naemono (approved):
Looks fine here.

hack/operatorhub/cmd/container/container.go
```diff
@@ -68,6 +68,14 @@ func Command() *cobra.Command {
 		"force will force the attempted pushing of remote images, even when the exact version is found remotely. (OHUB_FORCE)",
 	)
 
+	cmd.PersistentFlags().BoolVarP(
+		&flags.PreRelease,
+		flags.PreReleaseFlag,
+		"P",
```
pebrc, line 74:
```suggestion
		"p",
```

kvalliyurnatt: @pebrc `p` is already being used for project id
https://github.com/elastic/cloud-on-k8s/blob/9a72e9a8c6740eabaa67a6815536faeda797d5b6/hack/operatorhub/cmd/root.go#L55

Issue #8615, opened 2025-04-21
eck-operator PDB template generates corrupt YAML with bad labels placement


den-is:
This issue was caught by ArgoCD automatically during upgrade 2.16.1 -> 3.0.0.
I was able to replicate the same issue on my workstations using plain `helm template ...`
```sh
helm template -n elastic-system \
elastic-operator elastic/eck-operator \
--kube-version 1.30 \
--values values.yaml \
--version=3.0.0 --debug > eck-rendered.yaml
```

values.yaml
```yaml
installCRDs: true
replicaCount: 2
podDisruptionBudget:
  enabled: true
  minAvailable: 1
serviceMonitor:
  enabled: false
podLabels:
  env: staging
```

Resulting PodDisruptionBudget yaml:
```yaml
---
# Source: eck-operator/templates/pdb.yaml

apiVersion: policy/v1
kind: PodDisruptionBudget
metadata:
  name: elastic-operator
  namespace: elastic-system
  labels:    app.kubernetes.io/name: elastic-operator
    app.kubernetes.io/instance: elastic-operator
    app.kubernetes.io/version: "3.0.0"
    helm.sh/chart: eck-operator-3.0.0
    app.kubernetes.io/managed-by: Helm
spec:
  minAvailable: 1
  selector:
    matchLabels:      app.kubernetes.io/name: elastic-operator
      app.kubernetes.io/instance: elastic-operator
```

PR #8616, opened 2025-04-21
fix: Correct labels injection in eck-operator's PDB template

den-is:
Correctly render labels in PDB template

Fixes https://github.com/elastic/cloud-on-k8s/issues/8615


den-is:
I have signed CLA

stklcode:
Confirmed working.

Actually, the minimal change `s/indent/nindent/g` would suffice here while keeping the style of other templates (label expressions are always on the single line). `{{-` already takes care of the leading newline, if necessary. Result is the same.

den-is:
@stklcode I used to write compact template when there is no extra with/if logic, and fit everything in one line for.

But,... I've updated the commit with only `nindent` change

pebrc:
buildkite test this

pebrc (approved):

Issue #2652, opened 2020-03-02
Configure labels or annotations on child resources


spencergilbert:
## Proposal
Allow users to set labels on the "top level" resources created from the CRDs

**Use case. Why is this important?**
Today users can label the pods created by the ECK operator, but the statefulset for elasticsearch doesn't have any means of being labelled. Presumably the kibana would have the same issue - we use labels in monitoring, logs, generic discoverability, etc.

It'd be very helpful to have specifications in the CRD to apply labels to the deployments/statefulsets created from them.

spencergilbert:
I suppose also labelling the created services and all would be nice too

sebgl:
Would it make sense to have all StatefulSets/Deployment inherit the labels applied to the Elasticsearch/Kibana resource?


Example:
```yaml
apiVersion: elasticsearch.k8s.elastic.co/v1
kind: Elasticsearch
metadata:
  name: quickstart
  labels:
    foo: bar
spec:
  version: 7.6.0
  nodeSets:
  - name: default
    count: 1
```
would label the underlying StatefulSet with `foo: bar`. Same for Kibana Deployment. And all other resources: services, secrets, etc.

It feels like an easy way of labeling everything at once. Would that cover your use case @spencergilbert? Or would you expect to be able to specify per-NodeSet/StatefulSet labels?

spencergilbert:
That would definitely solve my use case :+1:

charith-elastic:
After prototyping how this would work, I think that it would be better to let users decide which labels or annotations they want to propagate to child objects. This is because some labels/annotations are probably meant for the CR only and having them propagated to children could result in confusion and/or unexpected behaviour (e.g. deleting by label selector).

I am proposing that we introduce two annotations that can be applied to CRs:
```yaml
eck.k8s.elastic.co/propagate-annotations: "comma,separated,list,of,annotations,to,propagate"
eck.k8s.elastic.co/propagate-labels: "comma,separated,list,of,labels,to,propagate"
```

Setting the annotation value to `*` would propagate all annotations/labels in the CR.

sebgl:
Similar issue for setting headless services labels: https://github.com/elastic/cloud-on-k8s/issues/3990.
As Charith suggested above, I think it makes sense to label all child resources, and not just StatefulSets.

tschroeder-zendesk:
[...]

PR #8673, opened 2025-05-26
Propagate metadata to children

barkbay:
Revives https://github.com/elastic/cloud-on-k8s/pull/3186 with some adjustments.

Fixes https://github.com/elastic/cloud-on-k8s/issues/2652

Still a wip, I might revisit the API a bit, and there's still some work to be done wrt e2e testing.

- [x] E2E Test
- [ ] Update documentation


prodsecmachine:

:white_check_mark: **security/snyk** check is complete. No issues have been found. [(View Details)](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/890b40e9-2d30-40da-9e3e-739027073037?refs=d4227ac8225bda58dba773c56c1a4b5f3937853e&source=prChecksComment)

:white_check_mark: **license/snyk** check is complete. No issues have been found. [(View Details)](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/890b40e9-2d30-40da-9e3e-739027073037/license?refs=d4227ac8225bda58dba773c56c1a4b5f3937853e&source=prChecksComment)

barkbay:
buildkite test this -f p=kind,t=TestMetadataPropagation

barkbay:
buildkite test this -f p=kind,t=TestMetadataPropagation

barkbay:
buildkite test this -f p=kind,t=TestMetadataPropagation

pkg/controller/agent/pod.go
```diff
@@ -170,16 +172,13 @@ func buildPodTemplate(params Params, fleetCerts *certificates.CertificatesSecret
...
 
-	agentLabels := maps.Merge(params.Agent.GetIdentityLabels(), map[string]string{
-		VersionLabelName: spec.Version})
-
-	annotations := map[string]string{
-		ConfigHashAnnotationName: fmt.Sprint(configHash.Sum32()),
-	}
-
+	podMeta := params.Meta.Merge(metadata.Metadata{
+		Labels:      maps.Merge(params.Agent.GetIdentityLabels(), map[string]string{VersionLabelName: spec.Version}),
```
pebrc: Were the identity labels not already included when the `params.Meta` was created in controller.go?

pkg/controller/common/annotation/metadata_propagation.go
```diff
@@ -0,0 +1,86 @@
...
+	"regexp"
+	"strings"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	// PropagateAnnotationsAnnotation is the annotation used to indicate which annotations should be propagated to child objects.
+	PropagateAnnotationsAnnotation = "eck.k8s.alpha.elastic.co/propagate-annotations"
```
pebrc, line 18: Not sure about the `alpha` bit here. I think once we ship this we are commited? But on second thought it might allow us to change the behavior later if need be. I guess I am on the fence.

pkg/controller/stackconfigpolicy/elasticsearch_config_settings.go
```diff
@@ -49,17 +51,21 @@ func newElasticsearchConfigSecret(policy policyv1alpha1.StackConfigPolicy, es es
 		}
 		data[ElasticSearchConfigKey] = configDataJSONBytes
 	}
+	meta := metadata.Propagate(&es, metadata.Metadata{
+		Labels: eslabel.NewLabels(types.NamespacedName{
+			Name:      es.Name,
+			Namespace: es.Namespace,
+		}),
```
pebrc:
```suggestion
		Labels: eslabel.NewLabels(k8s.ExtractNamespacedName(&es)),
```

pkg/controller/stackconfigpolicy/kibana_config_settings.go
```diff
@@ -38,17 +40,21 @@ func newKibanaConfigSecret(policy policyv1alpha1.StackConfigPolicy, kibana kiban
 			return corev1.Secret{}, err
 		}
 	}
+	meta := metadata.Propagate(&kibana, metadata.Metadata{
+		Labels: kblabel.NewLabels(types.NamespacedName{
+			Name:      kibana.Name,
+			Namespace: kibana.Namespace,
+		}),
```
pebrc:
```suggestion
		Labels: kblabel.NewLabels(k8s.ExtractNamespacedName(&kibana)),
```

pkg/controller/common/annotation/metadata_propagation.go
barkbay: This was mainly about leaving the door open to changes in behavior or in the expected content of the annotations. We may for example propagate the metadata to PVC or PV in a next release, or eventually expect the annotations to be a more "structured" string instead of a comma separated values (json?). The idea behind `alpha` was to deliver something, get some first feedback and adjust the final state of the feature if required. I'm happy to remove it if we think this is not needed.

barkbay:
buildkite test this -f p=kind,t=TestMetadataPropagation

pkg/controller/common/annotation/metadata_propagation.go
pebrc: Let's keep it as you proposed!

pebrc (commented):
I did a few tests and everything looks good. From my perspective we can merge this as is. 

 I wonder about two things though: 
* some objects are currently excluded from metadata propagation: 
  * assocation ES user secrets 
  * CA secrets seem to be missing metadata propagation (both the `-ca-internal` ones and the copied ones through the association mechanism)
  * you already mentioned PVCs are missing too 

I guess the question is where we draw the line?

* should we add global flags to control the behaviour in addition to the per resource annotations? 

For users that need to enforce certain metadata attributes across all resources created by the operator this might be relevant. Wdyt? 

OTOH they could also use a form of policy enforcement to achieve the same with kyverno or similar tools.

barkbay:
> CA secrets seem to be missing metadata propagation

This should work for new resources, IIRC the problem is that we don't reconcile theses `Secrets` [until the CA must be renewed](https://github.com/elastic/cloud-on-k8s/blob/4f714469aed46faf5fb0d2331384cc79ccf85246/pkg/controller/common/certificates/http_reconcile.go#L120C5-L120C16).

> assocation ES user secrets

For the associations I think the question is "from which resources the metadata must be propagated?"

* Should we only use the metadata from the application that creates the association on all the `Secrets` created? (for example metadata from Kibana in the Kibana to ES case)
* Should we only propagate the metadata to the resources loaded by the application that creates the association (the Secret that holds the API key in the Kibana example).

I was not really sure about the best strategy here.

pebrc:
> > CA secrets seem to be missing metadata propagation
> 
> This should work for new resources, IIRC the problem is that we don't reconcile theses `Secrets` [until the CA must be renewed](https://github.com/elastic/cloud-on-k8s/blob/4f714469aed46faf5fb0d2331384cc79ccf85246/pkg/controller/common/certificates/http_reconcile.go#L120C5-L120C16).

Ah yes that makes sense.

> 
> > assocation ES user secrets
> 
> For the associations I think the question is "from which resources the metadata must be propagated?"
> 
>  Should we only use the metadata from the application that creates the association on all the `Secrets` created? (for example metadata from Kibana in the Kibana to ES case)
> 

That's a good question. My intution was that it would make sense to use the metadata from the side of the association that creates the association e.g. Kibana in the KB->ES case. So the one that has the `*Ref` 

>   Should we only propagate the metadata to the resources loaded by the application that creates the association (the Secret that holds the API key in the Kibana example).
> I was not really sure about the best strategy here.

Same. It is somewhat arbitrary which side we pick. But I guess for me the main argument is to try to cover all objects that ECK creates. So that means also propagating to objects that the "other" side of the association consumes.

pebrc:
We could also merge this as a first step and follow up in separate PRs with the discussed extensions. Let me know what you prefer.

barkbay:
> That's a good question. My intution was that it would make sense to use the metadata from the side of the association that creates the association e.g. Kibana in the KB->ES case. So the one that has the *Ref

Makes sense 👍 , let me try to do this in this pr.

barkbay:
Still need to adjust the e2e test...

barkbay:
buildkite test this -f p=kind,t=TestMetadataPropagation

pebrc (approved):
Nice work! I did a few more simple tests and all looked good to me including the association resources.

PR #8757, opened 2025-07-18
[WIP] Use subpages for third party deps and API

barkbay:
Mostly for testing and getting a preview, please ignore.


prodsecmachine:

:white_check_mark: **security/snyk** check is complete. No issues have been found. [(View Details)](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/88af76f8-5407-44a8-bba4-8d90c6db78c5?refs=b512ca001a9f533f753e147dccab3af5f3429418&source=prChecksComment)

:white_check_mark: **license/snyk** check is complete. No issues have been found. [(View Details)](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/88af76f8-5407-44a8-bba4-8d90c6db78c5/license?refs=b512ca001a9f533f753e147dccab3af5f3429418&source=prChecksComment)

barkbay:
Landing page does not look great with

```yaml
toc:
  - file: index.md
  - file: api-docs.md
  - folder: third-party-dependencies
  - file: eck-configuration-flags.md
  - file: upgrade-predicates.md
```

⬇️ 

image

I don't understand why `previous` is "third-party deps" while I would expect it to be the next since it is a child?

eedugon:
> I don't understand why previous is "third-party deps" while I would expect it to be the next since it is a child?

I think our `previous` views have some limitations, and maybe that's one of them.

> Landing page does not look great with

What part you don't like and what would you like to improve? Maybe I can help you with that!

hack/licence-detector/generate-notice.sh
```diff
@@ -11,21 +11,38 @@ set -euo pipefail
...
     (
+        version="$(get_current_version)"
+        # Remove dots from the version string for compatibility with the doc web site.
+        outFile="${version//./_}.md"
         cd "$PROJECT_DIR"
         go mod download
         go list -m -json all | "${TEMP_DIR}"/go-licence-detector \
             -depsTemplate="${SCRIPT_DIR}"/templates/dependencies.md.tmpl \
-            -depsOut="${PROJECT_DIR}"/docs/reference/third-party-dependencies.md \
+            -template-value=eckVersion="${version}" \
```
barkbay, line 36: This relies on https://github.com/elastic/go-licence-detector/pull/29

barkbay:
Preview with the new template:
image

barkbay:
This PR is ready for review 🎉 , the preview is available here: https://docs-v3-preview.elastic.dev/elastic/cloud-on-k8s/pull/8757/reference

docs/reference/api-docs.md
```diff
@@ -1,2073 +1,13 @@
 ---
 mapped_pages:
   - https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-api-reference.html
-navigation_title: API Reference
+navigation_title: API Reference (moved)
```
barkbay, line 4: I had to preserve that page because it is referenced in the `docs-content` repo, doc generation fails if I remove it. I believe we can fix that in a follow up PR once this one is merged.

pebrc (approved):
LGTM! Nice work! As we release more versions of ECK we might have to revisit this strategy especially if there is an empty diff between versions it might make sense to think about grouping multiple versions in one document or something along those lines. But for now I am super happy we have a path forward.

docs/reference/api-reference/3_0_0.md
```diff
@@ -4,7 +4,7 @@ mapped_pages:
 navigation_title: API Reference for 3.0.0
 applies_to:
   deployment:
-    eck: ga 3.0.0
+    eck: ga 3.0.0, ga 3.1.0
```
barkbay: @pebrc I tested what we discussed out of band, here is the result:

image

pebrc: Very nice! This opens then indeed the door to collapse documentation for mulitple versions into one page.

eedugon:
@barkbay : should we remove the API reference and 3PP references docs identified as `from the main branch`? What's the value of them when we'll have the `3.0.0` and `3.0.1` versions published?

docs/reference/api-reference/3_1_0.md
shainaraskas, line 1: I just quickly diffed the two api references and I'm not seeing any substantive differences (one line about fleet advanced config). do we need to generate these as separate pages right now? do you expect substantive development to come to this area soon? 

I also wonder how the versioning approach for the API packages (e.g. v1, v1beta1, v1alpha1) interact with the versioning for ECK itself. it feels possibly like the API reference needs to follow its own versioning scheme based on these hints.

if that's the case, maybe we could state which api versions are compatible with which stack versions as an alternative

docs/reference/third-party-dependencies/index.md
```diff
@@ -0,0 +1,12 @@
...
+  - https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-dependencies.html
+navigation_title: Third-party dependencies 
+applies_to:
+  deployment:
+    eck: all
+---
+
+# Third-party dependencies [k8s-reference-dependencies]
+
+This section contains reference information about third-party dependencies used by {{eck}}.
```
shainaraskas, line 12: this file should ideally link to the children in the section

docs/reference/toc.yml
```diff
@@ -1,6 +1,7 @@
 toc:
   - file: index.md
   - file: api-docs.md
-  - file: third-party-dependencies.md
+  - folder: api-reference
+  - folder: third-party-dependencies
```
shainaraskas, line 5: I think that because we're using this folder model rather than a file/children model, these are being listed in a weird order. ideally, we'd list these dependency files from newest to oldest:

image

docs/reference/third-party-dependencies/3_1_0.md
```diff
@@ -1,16 +1,16 @@
 ---
 mapped_pages:
   - https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-dependencies.html
-navigation_title: Third-party dependencies 
+navigation_title: Third-party dependencies for 3.1.0
```
shainaraskas: because these sit under a "third-party dependencies" heading, we can trim this to just the version #

```suggestion
navigation_title: 3.1.0
```

shainaraskas:
>should we remove the API reference and 3PP references docs identified as from the main branch? What's the value of them when we'll have the 3.0.0 and 3.0.1 versions published?

if we still want to generate them for internal/dev use, we could consider making them [hidden files](https://elastic.github.io/docs-builder/configure/content-set/navigation/#hidden-files) so they can be visited but not consumed by readers who might be confused by them

shainaraskas (commented):
overall, this looks really great, and thank you for taking on the effort of refactoring the generation so quickly. my key piece of feedback is that we should think about the long-term vision for the API reference before we build a bunch of pages we need to maintain long term.

docs/reference/api-reference/3_1_0.md
barkbay: > one line about fleet advanced config

This is something new in ECK 3.1.0, which was not supported in 3.0.0. So I think I would keep 2 separate files.

> I also wonder how the versioning approach for the API packages (e.g. v1, v1beta1, v1alpha1) interact with the versioning for ECK itself.

Each K8S API version stage can be used to represent a different maturity and stability level. When a new version is introduced it also requires a change in the operator. They are not related to stack versions.

barkbay:
> What's the value of them when we'll have the 3.0.0 and 3.1.0 versions published?

IIRC we have to document third-party dependencies for compliance purposes, not sure if this includes the ones currently referenced by the project in its `main` branch though. I think it is also helpful to ensure that we can always generate these files, without waiting for a new release branch to be created. So maybe we can hide them as suggested, @pebrc do you have an opinion?

docs/reference/toc.yml
barkbay: yes 😕 , I think it would be a bit more complex to automate the generation of these files with the file/children model.

barkbay:
Preview of the new navigation titles:

image

pebrc:
> > What's the value of them when we'll have the 3.0.0 and 3.1.0 versions published?
> 
> IIRC we have to document third-party dependencies for compliance purposes, not sure if this includes the ones currently referenced by the project in its `main` branch though. I think it is also helpful to ensure that we can always generate these files, without waiting for a new release branch to be created. So maybe we can hide them as suggested, @pebrc do you have an opinion?

Hiding the file for the `main` branch sounds like a good option to me. And I agree we should keep generating the file for `main` to make sure your our generation tooling is functional.

barkbay:
> if we still want to generate them for internal/dev use, we could consider making them [hidden files](https://elastic.github.io/docs-builder/configure/content-set/navigation/#hidden-files) so they can be visited but not consumed by readers who might be confused by them

Are hidden files compatible with the fact that we want to avoid the file/children model?

docs/reference/api-reference/3_1_0.md
shainaraskas: >This is something new in ECK 3.1.0, which was not supported in 3.0.0. So I think I would keep 2 separate files.

this is the substance of the change: 

image
image

the change does not seem to a change to the API functionality (unless this is missing a related value for the new config model). Given this, I'd rather rely on the applicability information [in the core docs](https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/2019/deploy-manage/deploy/cloud-on-k8s/configuration-fleet#k8s-elastic-agent-fleet-configuration-custom-configuration) than generate a whole new API reference for a clarifying sentence.

> Each K8S API version stage can be used to represent a different maturity and stability level. When a new version is introduced it also requires a change in the operator. They are not related to stack versions.

it sounds like the version stage doesn't actually guarantee specific behavior independently of the eck version, so it still makes sense to use the ECK version as the primary versioning model 👍

shainaraskas:
> Are hidden files compatible with the fact that we want to avoid the file/children model?

unfortunately, I don't think so. if we needed to, we could ship this and revisit a more granular toc separately.

PR #8086, opened 2024-10-08
Kibana: Set default hardened security context

naemono:
Closes: #7787

## What is this change?

This sets the security context for Kibana to be more secure/hardened by default.

### Testing/Todo

- [ ] Test v8 Kibana in e2e tests
- [ ] Test v7 Kibana in e2e tests

PR #8811, opened 2025-09-04
Improve config/recipes/elastic-agent/fleet-kubernetes-integration-nonroot.yaml with drop capabilities

barkbay:
This PR improves `config/recipes/elastic-agent/fleet-kubernetes-integration-nonroot.yaml` to explain how to run Fleet server with dropped capabilities and ro file system.

(I think ECK should ideally do that by default as it is already the case for [Elasticsearch](https://github.com/elastic/cloud-on-k8s/pull/6703) and [Kibana](https://github.com/elastic/cloud-on-k8s/pull/8086))

### DIFF
--- a/config/recipes/elastic-agent/fleet-kubernetes-integration-nonroot.yaml
+++ b/config/recipes/elastic-agent/fleet-kubernetes-integration-nonroot.yaml
@@ -155,6 +155,23 @@ spec:
       spec:
         serviceAccountName: fleet-server
         automountServiceAccountToken: true
+        ## Uncomment the following lines to run fleet-server with restricted privileges and drop capabilities.
+#        containers:
+#        - name: agent
+#          securityContext:
+#            allowPrivilegeEscalation: false
+#            readOnlyRootFilesystem: true
+#            capabilities:
+#              drop: ["ALL"]
+#          env:
+#            - name: STATE_PATH
+#              value: /usr/share/elastic-agent/state
+#            - name: CONFIG_PATH
+#              value: /usr/share/elastic-agent/state
+#          args:
+#            - -e
+#            - -c
+#            - /etc/agent/elastic-agent.yml
 ---
 apiVersion: agent.k8s.elastic.co/v1alpha1
 kind: Agent


prodsecmachine:

:white_check_mark: **security/snyk** check is complete. No issues have been found. [(View Details)](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/61eaa404-24cd-471e-88a9-51bd4b6a711a?refs=c6b1a364c55f98402f79b3716b66c14ab9cbaf69&source=prChecksComment)

:white_check_mark: **license/snyk** check is complete. No issues have been found. [(View Details)](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/61eaa404-24cd-471e-88a9-51bd4b6a711a/license?refs=c6b1a364c55f98402f79b3716b66c14ab9cbaf69&source=prChecksComment)

config/recipes/elastic-agent/fleet-kubernetes-integration-nonroot.yaml
```diff
@@ -155,6 +155,23 @@ spec:
       spec:
         serviceAccountName: fleet-server
         automountServiceAccountToken: true
+        ## Uncomment the following lines to run fleet-server with restricted privileges and drop capabilities.
```
barkbay, line 158: 2 things I'm not sure:
1. Should it be the default in this recipe? (Said differently, should we remove the `#`?)
2. Do we want to apply the same changes to the Agents below? I think we don't as it could break some packages which may require some capabilities to run correctly?

pchila: For a managed elastic-agent, the important bits are:
- have the CONFIG_PATH and the STATE_PATH being writable and persisted across restarts (hence the 2 env vars below)
- the `-c /etc/agent/elastic-agent.yml` which will cause the config file to be copied in the state path and used allowing for some initial config injection from secrets/config maps.

The securityContext part is optional and depends on how locked down the user wants the agent containers to be and it's more a case-by-case choice so I would leave that commented and let the user be stricter if he needs to be.

The changes I mentioned above need to be replicated to the `elastic-agent` Agent as well.

barkbay:
I'm closing this PR since what we are trying to solve here is not related to whether the container is running as `root` or not. This needs to be addressed at the operator level and therefore requires an issue to discuss how to solve this.

PR #8871, opened 2025-10-22
Upgrade masters last when upgrading ES clusters

### LINKED ISSUE #8429: [Bug] ECK should only Update Elasticsearch StatefulSet version when attempting to upgrade the StatefulSet Pods
(ECK version 2.15.0)

**Background:**

I was recently upgrading a rather large Elasticsearch cluster from 8.16.2 to 8.17.1, but ran into an issue where one of the dedicated Master pods was recreated part way through the upgrade process.

**Issue**

The problem appears that ECK when it gets an upgrade of the Elasticsearch version, it will automatically update _all_ statefulset versions right away, and then perform the rolling restart. The problem is that if a pod gets killed/recreated part way through the process, there is no longer an "order of operations" applied and things can be upgraded in the wrong order.

**Reproduction:**

1. Create an Elasticsearch cluster with dedicated masters
2. Upgrade the Elasticsearch cluster
3. Recreate one of the master pods _while_ the upgrade is still working on non-master nodes
4. Once the new master node gets created, create an index
5. The index will get assigned the new Elasticsearch index version, and won't be allocatable on the lower version non-master nodes
6. Observe that the upgrade managed via ECK deadlocks on a yellow state because of allocation issues from step 5.

**Expectation:**

ECK should only upgrade the statefulset version when its ready to perform the rolling restart of that statefulset, and _not_ so far before in the upgrade process.

**Workaround:**

To workaround the deadlock, I had to manually (and carefully) delete/recreate each of the remaining non-master pods to allow them to pick up the new version.

naemono:
Fixes #8429 

## What is changing?

This ensure that the master StatefulSets are always upgraded last when doing a version upgrade of Elasticsearch.

## Validation
1. Manual testing
2. Unit test for behavior
3. e2e test validating behavior


prodsecmachine:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/18868b71-ab6d-462c-afa0-1786a4376c02?refs=906c863ad0012801aac8339137a6c5a88634ee9a&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/18868b71-ab6d-462c-afa0-1786a4376c02/license?refs=906c863ad0012801aac8339137a6c5a88634ee9a&source=prChecksComment) |

naemono:
buildkite test this -f p=kind,t=TestNonMasterFirstUpgradeComplexTopology -m s=9.1.2

naemono:
buildkite test this -f p=kind,t=TestHandleUpscaleAndSpecChanges_VersionUpgradeDataFirstFlow -m s=9.1.2

pkg/controller/elasticsearch/driver/upscale.go
```diff
@@ -166,3 +207,88 @@ func adjustStatefulSetReplicas(
...
+	targetVersion version.Version,
+) (bool, error) {
+	for _, statefulSet := range actualStatefulSets {
+		// Skip master StatefulSets
+		if label.IsMasterNodeSet(statefulSet) {
+			continue
+		}
+
+		// If the StatefulSet is not at the target version, it is not upgraded
+		// so don't even bother looking at the state/status of the StatefulSet.
```
barkbay, line 341: I think it is possible to enter a kind of "deadlock" when the `master` role is added to an existing `nodeSet`.
In that case:
*  The updated sts is detected here as a "not (yet) master sst"
*  But it is never going to be upgraded because it is considered as a `masterResources` in `adjusted`

Simple manifest to replicate the deadlock:

```yaml
kind: Elasticsearch
metadata:
  name: elasticsearch-sample
spec:
  version: 9.1.6
  nodeSets:
  - name: ns1
    config:
      node.roles: ["master","data"]
      node.store.allow_mmap: false
    count: 1
  - name: ns2
    config:
      node.roles: ["data", "ingest", "ml"]
      node.store.allow_mmap: false
    count: 3
```

Then upgrade to `9.2.0` with:

```yaml
kind: Elasticsearch
metadata:
  name: elasticsearch-sample
spec:
  version: 9.2.0
  nodeSets:
  - name: ns1
    config:
      node.roles: ["master","data"]
      node.store.allow_mmap: false
    count: 1
  - name: ns2
    config:
      node.roles: ["master", "data", "ingest", "ml"]
      node.store.allow_mmap: false
    count: 3
```

barkbay (dismissed):

barkbay (commented):
I think we also have a problem when we increase the master replicas:

## Initial manifest

```yaml
kind: Elasticsearch
metadata:
  name: elasticsearch-sample
spec:
  version: 9.1.6
  nodeSets:
  - name: ns1
    config:
      node.roles: ["master"]
      node.store.allow_mmap: false
    count: 1
  - name: ns2
    config:
      node.roles: ["data", "ingest", "ml"]
      node.store.allow_mmap: false
    count: 3
```

## Updated manifest

```yaml
kind: Elasticsearch
metadata:
  name: elasticsearch-sample
spec:
  version: 9.2.0
  nodeSets:
  - name: ns1
    config:
      node.roles: ["master"]
      node.store.allow_mmap: false
    count: 3
  - name: ns2
    config:
      node.roles: ["data", "ingest", "ml"]
      node.store.allow_mmap: false
    count: 3
```

Operator is stuck in:

```
    inProgressOperations:
      downscale:
        lastUpdatedTime: "2025-11-03T10:04:05Z"
      upgrade:
        lastUpdatedTime: "2025-11-03T10:04:05Z"
      upscale:
        lastUpdatedTime: "2025-11-03T10:13:51Z"
        nodes:
        - message: Creating master node
          name: elasticsearch-sample-es-ns1-1
          status: EXPECTED
        - message: Limiting master nodes creation to one at a time
          name: elasticsearch-sample-es-ns1-2
          status: PENDING
    observedGeneration: 3
    phase: ApplyingChanges
    version: 9.1.6
```

pkg/controller/elasticsearch/driver/upscale.go
```diff
@@ -166,3 +207,88 @@ func adjustStatefulSetReplicas(
...
+		}
+
+		for _, pod := range pods {
+			// Check if pod revision matches StatefulSet update revision
+			if statefulSet.Status.UpdateRevision != "" && sset.PodRevision(pod) != statefulSet.Status.UpdateRevision {
+				// This pod still needs to be upgraded
+				return false, nil
+			}
+		}
+	}
```
barkbay, line 377: I think this whole logic assumes that the StatefulSet controller has updated the status, which is not necessarily true from my experiments. I hit several cases where the StatefulSet spec has been updated but not its status. A non master sts is then detected as reconciled while it is not the case.

pkg/controller/elasticsearch/driver/upscale.go
```diff
@@ -66,33 +68,72 @@ func HandleUpscaleAndSpecChanges(
...
+		return results, nil
+	}
+
+	targetVersion, err := version.Parse(ctx.es.Spec.Version)
+	if err != nil {
+		return results, fmt.Errorf("while parsing Elasticsearch upgrade target version: %w", err)
+	}
+
+	// Check if all non-master StatefulSets have completed their upgrades before proceeding with master StatefulSets
+	if len(masterResources) > 0 {
```
barkbay: I was wondering in which cases we are expecting this to be `false`?
```suggestion
```

naemono: There are no cases. This has been removed.

pkg/controller/elasticsearch/driver/upscale.go
naemono: I've resolved a previous issue, but am still working on the full fix for this issue, btw. Should likely have a resolution tomorrow to move this forward.

pkg/controller/elasticsearch/driver/upscale.go
naemono: I believe I've also solved this case with a generation vs observedGeneration check. Lmk if you feel I've missed something.

naemono:
@barkbay All of the noted cases have been resolved in my testing. Lmk if you find any other issues.

pkg/controller/elasticsearch/driver/upscale.go
```diff
@@ -66,33 +69,70 @@ func HandleUpscaleAndSpecChanges(
...
+
+	// First, reconcile all non-master resources
+	actualStatefulSets, requeue, err := reconcileResources(ctx, actualStatefulSets, nonMasterResources)
+	if err != nil {
+		return results, fmt.Errorf("while reconciling non-master resources: %w", err)
+	}
+	if requeue {
+		results.Requeue = true
+		results.ActualStatefulSets = actualStatefulSets
+		return results, nil
```
barkbay, line 119: Not sure how much it is a problem, but I think this prevents the masters from being scaled until all the other nodes have been upgraded (this can be done today). I don't know if this could be a problem on large clusters, if a user wants to give more capacity to the ES control plane during an upgrade.

pkg/controller/elasticsearch/driver/upscale.go
```diff
@@ -66,33 +69,70 @@ func HandleUpscaleAndSpecChanges(
...
+	pendingNonMasterSTS, err := findPendingNonMasterStatefulSetUpgrades(ctx.k8sClient, actualStatefulSets, expectedResources.StatefulSets(), targetVersion)
+	if err != nil {
+		return results, fmt.Errorf("while checking non-master upgrade status: %w", err)
+	}
+
+	if len(pendingNonMasterSTS) > 0 {
+		// Non-master StatefulSets are still upgrading, skipping master StatefulSets temporarily.
+		// This will cause a requeue in the caller, and master StatefulSets will attempt to be processed in the next reconciliation
+		ctx.upscaleReporter.RecordPendingNonMasterSTSUpgrades(pendingNonMasterSTS)
+		results.ActualStatefulSets = actualStatefulSets
```
barkbay: Why do these 2 lines depend on `len(pendingNonMasterSTS) > 0`? 
My understanding is that:
* `results.ActualStatefulSets = actualStatefulSets` should be set immediately after the call to `reconcileResources(...)`
* `ctx.upscaleReporter.RecordPendingNonMasterSTSUpgrades(pendingNonMasterSTS)` can be called whether or not `len(pendingNonMasterSTS)` is `0`

(I'm just trying to make this function easier to read/process)

pkg/controller/elasticsearch/driver/upscale.go
```diff
@@ -166,3 +206,109 @@ func adjustStatefulSetReplicas(
...
+			if err != nil {
+				return actualStatefulSets, false, fmt.Errorf("handle volume expansion: %w", err)
+			}
+			if recreateSset {
+				ulog.FromContext(ctx.parentCtx).Info("StatefulSet is scheduled for recreation, requeuing", "name", res.StatefulSet.Name)
+				// The StatefulSet is scheduled for recreation: let's requeue before attempting any further spec change.
+				requeue = true
+				continue
+			}
+		} else if !exists {
```
barkbay: nit: `exists` has only 2 possible values
```suggestion
		} else {
```

pkg/controller/elasticsearch/driver/upscale.go
```diff
@@ -166,3 +206,109 @@ func adjustStatefulSetReplicas(
...
+		}
+		// update actual with the reconciled ones for next steps to work with up-to-date information
+		actualStatefulSets = actualStatefulSets.WithStatefulSet(reconciled)
+	}
+	ulog.FromContext(ctx.parentCtx).Info("Resources reconciled", "actualStatefulSets_size", len(actualStatefulSets), "requeue", requeue)
+	return actualStatefulSets, requeue, nil
+}
+
+// findPendingNonMasterStatefulSetUpgrades finds all non-master StatefulSets that have not completed their upgrades
+func findPendingNonMasterStatefulSetUpgrades(
```
barkbay, line 312: Should we use expectations here? I'm asking because IIUC we have (maybe) updated the `StatfulSets` just before.

```golang
// findPendingNonMasterStatefulSetUpgrades finds all non-master StatefulSets that have not completed their upgrades
func findPendingNonMasterStatefulSetUpgrades(
	client k8s.Client,
	actualStatefulSets es_sset.StatefulSetList,
	expectedStatefulSets es_sset.StatefulSetList,
	targetVersion version.Version,
	expectations *expectations.Expectations,
) ([]appsv1.StatefulSet, error) {

	pendingStatefulSet, err := expectations.ExpectedStatefulSetUpdates.PendingGenerations()
	if err != nil {
		return nil, err
	}
...
	pendingNonMasterSTS := make([]appsv1.StatefulSet, 0)
	for _, actualStatefulSet := range actualStatefulSets {
		expectedSset, _ := expectedStatefulSets.GetByName(actualStatefulSet.Name)

		// Skip master StatefulSets. We check both here because the master role may have been added
		// to a non-master StatefulSet during the upgrade spec change.
		if label.IsMasterNodeSet(actualStatefulSet) || label.IsMasterNodeSet(expectedSset) {
			continue
		}

		if slices.Contains(pendingStatefulSet, actualStatefulSet.Name) {
			pendingNonMasterSTS = append(pendingNonMasterSTS, actualStatefulSet)
			continue
		}
...

File: pkg/controller/elasticsearch/driver/upscale.go (hunk 5 above)
Comment (naemono) [reply pkg/controller/elasticsearch/driver/upscale.go]: Makes sense now that you've pointed it out. Updated.

File: pkg/controller/elasticsearch/driver/upscale.go (hunk 7 above)
Comment (naemono) [reply pkg/controller/elasticsearch/driver/upscale.go:312]: This also makes sense. I've also removed the later check in the code specifically checking generation/observedGeneration.

File: pkg/controller/elasticsearch/driver/upscale.go (hunk 4 above)
Comment (naemono) [reply pkg/controller/elasticsearch/driver/upscale.go:119]: I have added the functionality now to allow this, and updated the unit test to show this behavior, but I haven't manually tested this yet, but will do and update.

**Comment by barkbay** [comment]
I'll take another look today, sorry for the lag.

File: pkg/controller/elasticsearch/driver/upscale.go [hunk 8]
```diff
@@ -66,35 +72,116 @@ func HandleUpscaleAndSpecChanges(
...
-			if recreateSset {
-				// The StatefulSet is scheduled for recreation: let's requeue before attempting any further spec change.
-				results.Requeue = true
+	}
+
+	// The only adjustment we want to make to master statefulSets before ensuring that all non-master
+	// statefulSets have been reconciled is to scale up the replicas to the expected number.
+	// The only adjustment we want to make to master statefulSets before ensuring that all non-master
+	// statefulSets have been reconciled is to potentially scale up the replicas
+	// which should happen 1 at a time as we adjust the replicas early.
```
pebrc: Duplicate comment

pkg/controller/elasticsearch/driver/upscale.go
```diff
@@ -166,3 +253,114 @@ func adjustStatefulSetReplicas(
...
+		if err != nil {
+			return pendingNonMasterSTS, err
+		}
+
+		for _, pod := range pods {
+			// Check if pod revision matches StatefulSet update revision
+			if actualStatefulSet.Status.UpdateRevision != "" && sset.PodRevision(pod) != actualStatefulSet.Status.UpdateRevision {
+				// This pod still needs to be upgraded
+				pendingNonMasterSTS = append(pendingNonMasterSTS, actualStatefulSet)
+				continue
```
pebrc:
```suggestion
				break
```

once you found one pod that is not on the update revision and you added the stateful set we can break out to the next sts, instead of continuing to loop through the pods and potentially adding the same statefulSet multiple times

pkg/controller/elasticsearch/driver/upscale.go
```diff
@@ -166,3 +253,114 @@ func adjustStatefulSetReplicas(
...
+// reconcileResources handles the common StatefulSet reconciliation logic
+// It returns:
+// - the updated StatefulSets
+// - whether a requeue is needed
+// - any errors that occurred
+func reconcileResources(
+	ctx upscaleCtx,
+	actualStatefulSets es_sset.StatefulSetList,
+	resources []nodespec.Resources,
+) (es_sset.StatefulSetList, bool, error) {
```
barkbay: nit: Should we return an `UpscaleResults` directly, I have the feeling that we are always rebuilding a new one from the calling function?

```suggestion
func reconcileResources(
	ctx upscaleCtx,
	actualStatefulSets es_sset.StatefulSetList,
	resources []nodespec.Resources,
) (UpscaleResults, error) {
```

pkg/controller/elasticsearch/driver/upscale.go
```diff
@@ -66,35 +72,116 @@ func HandleUpscaleAndSpecChanges(
...
 		}
-		reconciled, err := es_sset.ReconcileStatefulSet(ctx.parentCtx, ctx.k8sClient, ctx.es, res.StatefulSet, ctx.expectations)
-		if err != nil {
-			return results, fmt.Errorf("reconcile StatefulSet: %w", err)
+
+		actualReplicas := sset.GetReplicas(actualSset)
+		targetReplicas := sset.GetReplicas(res.StatefulSet)
+
+		if actualReplicas < targetReplicas {
+			actualSset.Spec.Replicas = ptr.To[int32](targetReplicas)
```
pebrc:
```suggestion
			actualSset.Spec.Replicas = ptr.To(targetReplicas)
```

barkbay: ~I think this is bypassing the existing policy for creating master nodes:~ https://github.com/elastic/cloud-on-k8s/blob/cb0d001d077a3779c04d203642f76dd99f189ebd/pkg/controller/elasticsearch/driver/upscale_state.go#L216

barkbay: Sorry, while reading the code in my IDE I realized that the replicas are adjusted before 👍

pkg/controller/elasticsearch/driver/upscale.go
```diff
@@ -66,35 +72,116 @@ func HandleUpscaleAndSpecChanges(
...
+		return results, fmt.Errorf("while parsing Elasticsearch upgrade target version: %w", err)
+	}
+
+	// Check if all non-master StatefulSets have completed their upgrades before proceeding with master StatefulSets
+	pendingNonMasterSTS, err := findPendingNonMasterStatefulSetUpgrades(
+		ctx.k8sClient,
+		actualStatefulSets,
+		expectedResources.StatefulSets(),
+		targetVersion,
+		expectations.NewExpectations(ctx.k8sClient),
```
barkbay: Why are we creating new expectations?
```suggestion
		ctx.expectations,
```

pkg/controller/elasticsearch/driver/upscale.go
```diff
@@ -166,3 +253,114 @@ func adjustStatefulSetReplicas(
...
+		// so don't even bother looking at the state/status of the StatefulSet.
+		actualVersion, err := es_sset.GetESVersion(actualStatefulSet)
+		if err != nil {
+			return pendingNonMasterSTS, err
+		}
+		if actualVersion.LT(targetVersion) {
+			pendingNonMasterSTS = append(pendingNonMasterSTS, actualStatefulSet)
+			continue
+		}
+
```
barkbay, line 350: We should not rely on the status until the sts controller has observed the new generation.

```suggestion
		if actualStatefulSet.Status.ObservedGeneration < actualStatefulSet.Generation {
			// The StatefulSet controller has not yet observed the latest generation.
			pendingNonMasterSTS = append(pendingNonMasterSTS, actualStatefulSet)
			continue
		}
```

barkbay: or:
*  use `actualStatefulSets.PendingReconciliation()` before that loop
* create a common function to reuse the logic in https://github.com/elastic/cloud-on-k8s/blob/cb0d001d077a3779c04d203642f76dd99f189ebd/pkg/controller/elasticsearch/driver/expectations.go#L23-L46

pkg/controller/elasticsearch/driver/upscale.go
```diff
@@ -166,3 +253,114 @@ func adjustStatefulSetReplicas(
...
+// findPendingNonMasterStatefulSetUpgrades finds all non-master StatefulSets that have not completed their upgrades
+func findPendingNonMasterStatefulSetUpgrades(
+	client k8s.Client,
+	actualStatefulSets es_sset.StatefulSetList,
+	expectedStatefulSets es_sset.StatefulSetList,
+	targetVersion version.Version,
+	expectations *expectations.Expectations,
+) ([]appsv1.StatefulSet, error) {
+	pendingStatefulSet, err := expectations.ExpectedStatefulSetUpdates.PendingGenerations()
```
pebrc:
```suggestion
	pendingStatefulSets, err := expectations.ExpectedStatefulSetUpdates.PendingGenerations()
```

pkg/controller/elasticsearch/driver/upscale_test.go
```diff
@@ -569,3 +570,371 @@ func Test_adjustResources(t *testing.T) {
...
+	require.NoError(t, k8sClient.Get(context.Background(), types.NamespacedName{Namespace: "ns", Name: "data-sset"}, &dataSset))
+	dataSset.Status.UpdatedReplicas = 0                // No replicas updated yet
+	dataSset.Status.Replicas = 2                       // Total replicas
+	dataSset.Status.UpdateRevision = "data-sset-12345" // New revision (different from old)
+	require.NoError(t, k8sClient.Status().Update(context.Background(), &dataSset))
+
+	// Call HandleUpscaleAndSpecChanges and verify that both data upgrade has begun and master STS is not updated
+	fmt.Println("Calling HandleUpscaleAndSpecChanges which should upscale the master replicas to 4")
+	_, err = HandleUpscaleAndSpecChanges(ctx, actualStatefulSets, expectedResourcesUpgrade)
+	require.NoError(t, err)
```
barkbay: I would expect `fmt.Println ` to be caught by the linter.
```suggestion
	_, err = HandleUpscaleAndSpecChanges(ctx, actualStatefulSets, expectedResourcesUpgrade)
	require.NoError(t, err, "HandleUpscaleAndSpecChanges to upscale the master replicas to 4 should not return an error")
```

pkg/controller/elasticsearch/driver/upscale_test.go
```diff
@@ -569,3 +570,371 @@ func Test_adjustResources(t *testing.T) {
...
+	// Update the existing data pods to have the new revision
+	var pod0 corev1.Pod
+	require.NoError(t, k8sClient.Get(context.Background(), types.NamespacedName{Namespace: "ns", Name: "data-sset-0"}, &pod0))
+	pod0.Labels["controller-revision-hash"] = "data-sset-12345"
+	require.NoError(t, k8sClient.Update(context.Background(), &pod0))
+
+	var pod1 corev1.Pod
+	require.NoError(t, k8sClient.Get(context.Background(), types.NamespacedName{Namespace: "ns", Name: "data-sset-1"}, &pod1))
+	pod1.Labels["controller-revision-hash"] = "data-sset-12345"
+	require.NoError(t, k8sClient.Update(context.Background(), &pod1))
```
barkbay:
```suggestion
	for i := 0; i < int(dataSset.Status.Replicas); i++ {
		var pod corev1.Pod
		require.NoError(t, k8sClient.Get(context.Background(), types.NamespacedName{Namespace: "ns", Name: fmt.Sprintf("data-sset-%d", i)}, &pod))
		pod.Labels["controller-revision-hash"] = "data-sset-12345"
		require.NoError(t, k8sClient.Update(context.Background(), &pod))
	}
```

pkg/controller/elasticsearch/driver/upscale.go
```diff
@@ -66,35 +72,116 @@ func HandleUpscaleAndSpecChanges(
...
-		reconciled, err := es_sset.ReconcileStatefulSet(ctx.parentCtx, ctx.k8sClient, ctx.es, res.StatefulSet, ctx.expectations)
-		if err != nil {
-			return results, fmt.Errorf("reconcile StatefulSet: %w", err)
+
+		actualReplicas := sset.GetReplicas(actualSset)
+		targetReplicas := sset.GetReplicas(res.StatefulSet)
+
+		if actualReplicas < targetReplicas {
+			actualSset.Spec.Replicas = ptr.To[int32](targetReplicas)
+			if err := ctx.k8sClient.Update(ctx.parentCtx, &actualSset); err != nil {
```
pebrc: This is updating the stateful sets without recording an expectation (props go to claude for finding this one)

Image

barkbay:
I will take another look at it today.

_Edit: was not able to make it today, will do tomorrow first thing_

pkg/controller/elasticsearch/driver/upscale.go
```diff
@@ -66,35 +72,112 @@ func HandleUpscaleAndSpecChanges(
...
 		}
-		reconciled, err := es_sset.ReconcileStatefulSet(ctx.parentCtx, ctx.k8sClient, ctx.es, res.StatefulSet, ctx.expectations)
-		if err != nil {
-			return results, fmt.Errorf("reconcile StatefulSet: %w", err)
+
+		actualReplicas := sset.GetReplicas(actualSset)
+		targetReplicas := sset.GetReplicas(res.StatefulSet)
+
+		if actualReplicas < targetReplicas {
+			actualSset.Spec.Replicas = ptr.To(targetReplicas)
```
barkbay: I think we should use `UpdateReplicas(...)` instead of updating directly `Replicas`, to also update `common.k8s.elastic.co/template-hash`
```suggestion
			nodespec.UpdateReplicas(&actualSset, ptr.To[int32](targetReplicas))
```

pkg/controller/elasticsearch/driver/upscale.go
```diff
@@ -66,35 +72,112 @@ func HandleUpscaleAndSpecChanges(
...
+			if err := ctx.k8sClient.Update(ctx.parentCtx, &actualSset); err != nil {
```
barkbay: Should we not use `es_sset.ReconcileStatefulSet()` instead of calling `k8sClient.Update(..)` directly? It already includes the call to `expectations.ExpectGeneration(reconciled)`.

barkbay: And maybe a follow-up question is should we update `UpscaleResults` with the result of `es_sset.ReconcileStatefulSet()` to work with a consistent view in the rest of the driver logic.

pkg/controller/elasticsearch/driver/upscale.go
```diff
@@ -66,35 +72,112 @@ func HandleUpscaleAndSpecChanges(
...
-			}
-			if recreateSset {
-				// The StatefulSet is scheduled for recreation: let's requeue before attempting any further spec change.
-				results.Requeue = true
+	}
+
+	// The only adjustment we want to make to master statefulSets before ensuring that all non-master
+	// statefulSets have been reconciled is to potentially scale up the replicas
+	// which should happen 1 at a time as we adjust the replicas early.
+	if err = maybeUpscaleMasterResources(ctx, masterResources); err != nil {
```
barkbay:
```suggestion
	if err := maybeUpscaleMasterResources(ctx, masterResources); err != nil {
```

pkg/controller/elasticsearch/driver/upscale.go
```diff
@@ -66,35 +72,112 @@ func HandleUpscaleAndSpecChanges(
...
+func maybeUpscaleMasterResources(ctx upscaleCtx, masterResources []nodespec.Resources) error {
+	// Upscale master StatefulSets using the adjusted resources and read the current StatefulSet
+	// from k8s to get the latest state.
+	for _, res := range masterResources {
+		stsName := res.StatefulSet.Name
+
+		// Read the current StatefulSet from k8s to get the latest state
+		var actualSset appsv1.StatefulSet
+		if err := ctx.k8sClient.Get(ctx.parentCtx, k8s.ExtractNamespacedName(&res.StatefulSet), &actualSset); err != nil {
+			if apierrors.IsNotFound(err) {
```
barkbay, line 172: I was trying to understand in which cases we can get a `404` IIUC one of them is when the user is attempting to scale up the masters with a new nodeset? Let's maybe add a godoc to explain that we are only scaling up existing master statfulsets, new ones are ignored.

pkg/controller/elasticsearch/driver/upscale.go
barkbay: I just realized that calling this when `len(nonMasterResources) == 0` (or more generally, when all non-master nodesets have already been upgraded?) can be slightly suboptimal.

Assuming that the initial state is:

```yaml
kind: Elasticsearch
metadata:
  name: elasticsearch-sample
spec:
  version: 9.1.0
  nodeSets:
  - name: default
    config:
      node.roles: ["master", "data", "ingest", "ml"]
      node.store.allow_mmap: false
    count: 3
```

If we update and upgrade to:

```yaml
kind: Elasticsearch
metadata:
  name: elasticsearch-sample
spec:
  version: 9.1.2
  nodeSets:
  - name: default
    config:
      node.roles: ["master", "data", "ingest", "ml"]
      node.store.allow_mmap: false
    count: 4
```

Then we are going to scale up the `9.1.0` statefulset, leading to the creation of `elasticsearch-sample-es-default-3`, but immediately in the next reconciliation we are going to delete `elasticsearch-sample-es-default-3` to upgrade it to `9.1.2`

barkbay: My previous comment made me wonder if `!isVersionUpgrade` is actually the only reason we might want to reconcile everything at once.

barkbay:
buildkite test this -f p=kind,t=TestNonMasterFirstUpgradeComplexTopology -m s=8.15.2

test/e2e/es/non_master_first_upgrade_test.go
```diff
@@ -0,0 +1,103 @@
...
+func newNonMasterFirstUpgradeWatcher(es esv1.Elasticsearch) test.Watcher {
+	var violations []string
+
+	return test.NewWatcher(
+		"watch StatefulSet upgrade order: non-master StatefulSets should upgrade before master StatefulSets",
+		2*time.Second,
+		func(k *test.K8sClient, t *testing.T) {
+			statefulSets, err := essset.RetrieveActualStatefulSets(k.Client, k8s.ExtractNamespacedName(&es))
+			if err != nil {
+				t.Logf("failed to get StatefulSets: %s", err.Error())
```
barkbay, line 39: Will this test fail if we consistently get an error here? (my feeling is that it's not going to be the case because `violations` is always empty in that case, but maybe I'm missing something)

test/e2e/es/non_master_first_upgrade_test.go
```diff
@@ -0,0 +1,103 @@
...
+		WithVersion(srcVersion).
+		WithESMasterNodes(3, elasticsearch.DefaultResources).
+		WithESDataNodes(2, elasticsearch.DefaultResources).
+		WithESCoordinatingNodes(1, elasticsearch.DefaultResources)
+
+	mutated := initial.WithNoESTopology().
+		WithVersion(dstVersion).
+		WithESMasterNodes(3, elasticsearch.DefaultResources).
+		WithESDataNodes(2, elasticsearch.DefaultResources).
+		WithESCoordinatingNodes(1, elasticsearch.DefaultResources)
```
barkbay: Without `WithMutatedFrom` the data integrity index has no replicas, which means that the cluster is going to be red as soon as one Pod is recreated, and the test is going to fail.
```suggestion
	mutated := initial.WithVersion(dstVersion).WithMutatedFrom(&initial)
```

_Edit: confirmed by the e2e test I started in my previous buildkite comment_

```json
{
    "Time": "2025-12-03T09:12:11.122554149Z",
    "Action": "output",
    "Package": "github.com/elastic/cloud-on-k8s/v3/test/e2e/es",
    "Test": "TestNonMasterFirstUpgradeComplexTopology/Elasticsearch_cluster_health_should_not_have_been_red_during_mutation_process",
    "Output": "{\"log.level\":\"error\",\"@timestamp\":\"2025-12-03T09:12:11.117Z\",\"message\":\"continuing with additional tests\",\"service.version\":\"0.0.0-SNAPSHOT+00000000\",\"service.type\":\"eck\",\"ecs.version\":\"1.4.0\",\"error\":\"test Elasticsearch cluster health should not have been red during mutation process failed\",\"error.stack_trace\":\"github.com/elastic/cloud-on-k8s/v3/test/e2e/test.StepList.RunSequential\n\t/go/src/github.com/elastic/cloud-on-k8s/test/e2e/test/step.go:52\ngithub.com/elastic/cloud-on-k8s/v3/test/e2e/test.RunMutationsWhileWatching\n\t/go/src/github.com/elastic/cloud-on-k8s/test/e2e/test/run_mutation.go:77\ngithub.com/elastic/cloud-on-k8s/v3/test/e2e/es.runNonMasterFirstUpgradeTest\n\t/go/src/github.com/elastic/cloud-on-k8s/test/e2e/es/non_master_first_upgrade_test.go:76\ngithub.com/elastic/cloud-on-k8s/v3/test/e2e/es.TestNonMasterFirstUpgradeComplexTopology\n\t/go/src/github.com/elastic/cloud-on-k8s/test/e2e/es/non_master_first_upgrade_test.go:102\ntesting.tRunner\n\t/usr/local/go/src/testing/testing.go:1792\"}\n"
}
```

pkg/controller/elasticsearch/driver/upscale.go
naemono: I have done both in https://github.com/elastic/cloud-on-k8s/pull/8871/commits/5142908ff05098170fd00b6b2852321f52daacae

pkg/controller/elasticsearch/driver/upscale.go
naemono: Outdated after https://github.com/elastic/cloud-on-k8s/commit/5142908ff05098170fd00b6b2852321f52daacae

naemono: Is this https://github.com/elastic/cloud-on-k8s/pull/8871/commits/4b966db0cfc4d8ae414c0631cfb4bed412ae480b what are were intending from this comment @barkbay ?

test/e2e/es/non_master_first_upgrade_test.go
naemono: I have updated the e2e test appropriately. Thanks for catching this.

naemono: buildkite test this -f p=kind,t=TestNonMasterFirstUpgradeComplexTopology -m s=8.15.2

naemono:
buildkite test this -f p=kind,t=TestNonMasterFirstUpgradeComplexTopology -m s=8.15.2

test/e2e/es/non_master_first_upgrade_test.go
naemono: No it's not. I wonder what would be the limit we set on errors in the watcher where it would fail? There are 4 instances I can find where we don't fail in these watchers in the e2e tests currently:

https://github.com/elastic/cloud-on-k8s/blob/b43765bb45afddc011cf80d61a987ba7ee6965e1/test/e2e/kb/version_upgrade_test.go#L130
https://github.com/elastic/cloud-on-k8s/blob/b43765bb45afddc011cf80d61a987ba7ee6965e1/test/e2e/kb/version_upgrade_test.go#L240
https://github.com/elastic/cloud-on-k8s/blob/b43765bb45afddc011cf80d61a987ba7ee6965e1/test/e2e/kb/version_upgrade_test.go#L275
https://github.com/elastic/cloud-on-k8s/blob/b43765bb45afddc011cf80d61a987ba7ee6965e1/test/e2e/test/elasticsearch/checks_budget.go#L33

test/e2e/es/non_master_first_upgrade_test.go
```diff
@@ -0,0 +1,99 @@
...
+// runNonMasterFirstUpgradeTest runs the complete test for non-master first upgrade behavior
+func runNonMasterFirstUpgradeTest(t *testing.T, initial, mutated elasticsearch.Builder) {
+	watcher := newNonMasterFirstUpgradeWatcher(initial.Elasticsearch)
+
+	test.RunMutationsWhileWatching(
+		t,
+		[]test.Builder{initial},
+		[]test.Builder{mutated},
+		[]test.Watcher{watcher},
+	)
```
barkbay: nit: Is there a specific reason for having this in a dedicated function? (it feels to me that it is easier to read if this is inlined in `TestNonMasterFirstUpgradeComplexTopology`)

test/e2e/es/non_master_first_upgrade_test.go
barkbay: > There are 4 instances I can find where we don't fail in these watchers in the e2e tests currently

Yes, good point, I never realized this 😕  I'm not sure that there is a good solution. I would maybe just check that the logic runs successfully at least once? Far from being bulletproof, but that would cover the case where something is completely broken or wrong.

pkg/controller/elasticsearch/driver/upscale.go
```diff
@@ -66,34 +72,124 @@ func HandleUpscaleAndSpecChanges(
...
-		if _, err := common.ReconcileService(ctx.parentCtx, ctx.k8sClient, &res.HeadlessService, &ctx.es); err != nil {
-			return results, fmt.Errorf("reconcile service: %w", err)
+	}
+
+	// Only attempt this upscale of master StatefulSets if there are any non-master StatefulSets to reconcile,
+	// otherwise you will immediately upscale the masters, and then delete the new node you added to upgrade.
+	if len(nonMasterResources) > 0 {
+		// The only adjustment we want to make to master statefulSets before ensuring that all non-master
+		// statefulSets have been reconciled is to potentially scale up the replicas
+		// which should happen 1 at a time as we adjust the replicas early.
```
barkbay: I think this is no longer true since https://github.com/elastic/cloud-on-k8s/pull/8940?

test/e2e/es/non_master_first_upgrade_test.go
naemono: I've done this. Lmk if the manner I did this isn't what you were thinking.

naemono:
buildkite test this -f p=kind -m s=8.15.2,s=9.2.0

naemono:
> buildkite test this -f p=kind -m s=8.15.2,s=9.2.0

The failures have nothing to do with my PR, but I'm going to try and run it one more time to be doubly sure...

naemono:
buildkite test this -f p=kind,E2E_TAGS=beat -m s=8.15.2

naemono:
> buildkite test this -f p=kind,E2E_TAGS=beat -m s=8.15.2

Why beat? Because the failures were all here:

```
🐞 TestMetricbeatDefaultConfig/with_stack_monitoring ~ kind-8-15-2
🐞 TestMetricbeatDefaultConfig/with_stack_monitoring/ES_data_should_pass_validations ~ kind-8-15-2
🐞 TestMetricbeatDefaultConfig/without_stack_monitoring ~ kind-8-15-2
🐞 TestMetricbeatDefaultConfig/without_stack_monitoring/ES_data_should_pass_validations ~ kind-8-15-2
🐞 TestMetricbeatHostsRecipe/ES_data_should_pass_validations ~ kind-8-15-2
```

naemono:
My local testing for both 9.2.x and 8.15.2 in my gke cluster are not showing the same error as above, and I'm not really seeing how these changes would affect Beat e2e tests. I'm going ahead and getting this merged.

PR #8892, opened 2025-11-05
Service template support for remoteClusterServer interface

eedugon:
## Add service customization support for remoteClusterServer

This PR adds the ability to customize the Kubernetes Service manifest for Elasticsearch's remote cluster server, similar to how `spec.http.service` and `spec.transport.service` work.

For your consideration, if you believe there's no need to support this please ignore. The main use case where this is needed is remote clusters functionality from external clusters trying to reach the ECK-managed cluster.

For example in ECH and ECE, port 9443 is accessible in the platform proxies, and it would be great to offer a way to open it up for ECK managed clusters, same as HTTP.

### What changed

- Added `service` field to `RemoteClusterServer` struct using `ServiceTemplate`
- Updated `NewRemoteClusterService` to use the user-provided service template
- Maintains backward compatibility: if `service` is not specified, the service is created with the same defaults as before (ClusterIP with ClusterIP=None)

### Example Usage

```
spec:
  remoteClusterServer:
    enabled: true
    service:
      spec:
        type: LoadBalancer
```

- Added unit tests to verify default behavior and user-provided template customization
- All existing tests continue to pass


prodsecmachine:

|---|---|---|---|---|---|---|
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/8b76f68a-3685-4138-b30e-c3e8066c228a/license?refs=32bdfc00d021312cb09c9cf239180299224f7e54&source=prChecksComment) |
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/8b76f68a-3685-4138-b30e-c3e8066c228a?refs=32bdfc00d021312cb09c9cf239180299224f7e54&source=prChecksComment) |

pkg/controller/elasticsearch/services/services.go
```diff
@@ -161,29 +161,31 @@ func NewInternalService(es esv1.Elasticsearch, meta metadata.Metadata) *corev1.S
...
+	svc := corev1.Service{
+		ObjectMeta: es.Spec.RemoteClusterServer.Service.ObjectMeta,
+		Spec:       es.Spec.RemoteClusterServer.Service.Spec,
 	}
-	selector := label.NewLabels(k8s.ExtractNamespacedName(&es))
+
+	svc.ObjectMeta.Namespace = es.Namespace
+	svc.ObjectMeta.Name = RemoteClusterServiceName(es.Name)
+	// Allow connections to pods that are not yet ready
+	svc.Spec.PublishNotReadyAddresses = true
```
eedugon, line 173: Not totally sure why we do `PublishNotReadyAddresses = true` in this SVC, neither why it is headless by default, instead of a normal ClusterIP, but anyway that's not in the scope of this PR.

barkbay: > Not totally sure why we do PublishNotReadyAddresses = true in this SVC

I applied the same logic as for the transport protocol, my understanding is that a Pod doesn’t need to be Ready before it’s discoverable (could be wrong though 😄 )

> neither why it is headless by default

The DNS name is used as a seed, there is no load balancing:

```json
    "mode": "sniff",
    "seeds": [
      "cluster2-es-remote-cluster.ns2.svc:9443"
    ],
```

The DNS lookup will return a list of Pods, which is I believe enough (in the original use case) for ES to discover the other nodes.

pkg/apis/elasticsearch/v1/elasticsearch_types.go
```diff
@@ -155,6 +155,8 @@ type ElasticsearchSpec struct {
 type RemoteClusterServer struct {
 	// +kubebuilder:validation:Optional
 	Enabled bool `json:"enabled,omitempty"`
+	// Service defines the template for the associated Kubernetes Service object.
```
barkbay: nit
```suggestion
	// Service defines the template for the remote cluster server Service object.
```

barkbay (approved):
LGTM

pkg/controller/elasticsearch/services/services.go
eedugon: So, for this functionality we use direct pod to pod connectivity and `sniff` mode. It makes sense as all it's within the same ECK managed environment. And it adds less overhead than `proxy` mode with a  standard `ClusterIP` svc, so probably better from performance point of view.

I thought we were going to use `proxy` mode, but when feasible, I think sniff mode is actually better (it needs direct connectivity from the client to the discovered `publish_addresses` of the nodes).

Thanks for letting me know!

>  PublishNotReadyAddresses = true
> I applied the same logic as for the transport protocol, my understanding is that a Pod doesn’t need to be Ready before it’s discoverable

From remote clusters (RCS2.0) point of view, probably a pod needs to be ready before it can really execute searches, but maybe not to be used as a `seed` to discover the remote gateways / nodes. Anyway not a big deal IMO!

If we were using `proxy` mode with kube-proxy and standard services, then I'd suggest for RCS2.0 to not publish unready pods, as it could cause certain requests to fail during start and stop of nodes.

PR #8918, opened 2025-11-21
Set persistentVolumeClaimRetentionPolicy per default delete

barmettlerl:
With this change I want to make sure that the PVC getting deleted once the StatefulSet is deleted or scaled down. In the context of our Organisation, we'd like to have a policy in place that enfoces `persistentVolumeClaimRetentionPolicy` to be set. More information about persistentVolumeClaimRetentionPolicy you can find under https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#persistentvolumeclaim-retention.

I think having `whenDeleted` and `whenScaled` with the `Delete` option is the most sensible choice, since how I understood it the eck-operator does not take usage of the persistence property we would get if we have the `Retain` option. 

I've saw that https://github.com/elastic/cloud-on-k8s/issues/8478 already mentioned this issue.

Issue #8478, opened 2025-02-21
Add option to configure logstash statefulset persistentVolumeClaimRetentionPolicy


martin-ilavsky:
Logstash statefulset has current default:
```
 spec:
  persistentVolumeClaimRetentionPolicy:
    whenDeleted: Retain
    whenScaled: Retain    
```

I want to be able to change `whenScaled` to `Delete` with logstash CRD. Currently when I patch it, operator will change it to the original values. Have I just missed where is it supposed to be configured ?

barmettlerl:
They haven't implemented it yet

PR #8920, opened 2025-11-21
Set persistentVolumeClaimRetentionPolicy per default delete #8918

barmettlerl:
With this change I want to make sure that the PVC getting deleted once the StatefulSet is deleted or scaled down. In the context of our Organisation, we'd like to have a policy in place that enfoces persistentVolumeClaimRetentionPolicy to be set. More information about persistentVolumeClaimRetentionPolicy you can find under https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#persistentvolumeclaim-retention.

I think having whenDeleted and whenScaled with the Delete option is the most sensible choice, since how I understood it the eck-operator does not take usage of the persistence property we would get if we have the Retain option.

I've saw that https://github.com/elastic/cloud-on-k8s/issues/8478 already mentioned this issue.

signed agreement
[Complete_with_Docusign_Elastic_Contributor_L.pdf](https://github.com/user-attachments/files/23675423/Complete_with_Docusign_Elastic_Contributor_L.pdf)


prodsecmachine:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/d3b87f7e-590f-4c31-be45-4331c046a0a2?refs=a923991ff9861932421cac896817e72b9dd35dcf&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/d3b87f7e-590f-4c31-be45-4331c046a0a2/license?refs=a923991ff9861932421cac896817e72b9dd35dcf&source=prChecksComment) |

deploy/eck-operator/templates/statefulset.yaml
```diff
@@ -20,6 +20,9 @@ spec:
       {{- include "eck-operator.selectorLabels" . | nindent 6 }}
   serviceName: {{ include "eck-operator.fullname" . }}
   replicas: {{ .Values.replicaCount }}
+  persistentVolumeClaimRetentionPolicy:
```
barkbay, line 23: The operator has no `volumeClaimTemplates`, so I'm a bit unsure about the purpose of this.

pebrc:
I am closing this as there there has been no response to the code review.

deploy/eck-operator/templates/statefulset.yaml
barmettlerl: not sure, if having a `volumeClaimTemplates` would make a difference here. The `persistentVolumeClaimRetentionPolicy` is more about what happens to the storage when the workload scales or is deleted

barmettlerl:
ah sorry, I was in vacation

deploy/eck-operator/templates/statefulset.yaml
pebrc: I think the point Michael was trying to make here is that the operator does not have any persistent storage that would need a retention policy to begin with.

barmettlerl: fair enough, but in our case we have a policy in our cluster that enforces `persistentVolumeClaimRetentionPolicy` on all StatefullSets and it's cumbersome to make an exception to all namespaces we deploy the elastic stack. So making it more explicit would already help :)

barkbay: > it's cumbersome to make an exception to all namespaces we deploy the elastic stack

Why would you need to create an exception to all these namespaces? This manifest defines how the _operator_ is deployed, not the stack.

barmettlerl: yes, but the operator creates Statefulsets, no ? and those need to have `persistentVolumeClaimRetentionPolicy`

barkbay: How does this change relate to the StatefulSets created by the operator? `persistentVolumeClaimRetentionPolicy` cannot be currently customized in the [`nodeSet`](https://www.elastic.co/docs/reference/cloud-on-k8s/api-reference/3_2_0#nodeset). This would require an API change.

pebrc:
Apologies, but I am closing this again as there seems to be a misunderstanding about the effect of this change, I don't think it solves the problem it tries to solve.

Issue #8621, opened 2025-04-24
Review Kibana's recommended resources


afharo:
I noticed this config example, where we set Kibana's memory limits to 1GB:

https://github.com/elastic/cloud-on-k8s/blob/83ecf97cb6f1d68c49f8d31933ee24ccf3ae70dd/config/samples/kibana/kibana_es.yaml#L38-L41

I wonder if we should change that to memory **requests** instead.

I'd also like to ask: does the ECK operator change Node.js' Max Heap Size based on the requests? AFAIK, [V8 limits the heap size to half of the available memory](https://stackoverflow.com/questions/64119135/what-is-the-default-value-of-available-memory-when-max-old-space-size-flag-i/64120041#64120041), so we may be unnecessarily choking Kibana.

For context, ECH sets the max heap size to 80% of the available memory (apologies for [the private link](https://github.com/elastic/cloud-assets/blob/master/stackpack/kibana/kibana.sh#L60-L64)). Maybe this repo benefits from those Docker images already?

pebrc:
Thanks @afharo that's a good point we currently do not set `max-old-space-size` at all in ECK. And we are not using the ECH stack packs but rather the default Kibana container images. 

Would this not be something that should be already set in the Kibana images? For example that when running in a container a Kibana bootstrap script could check `/sys/fs/cgroup/memory/memory.limit_in_bytes` or `/sys/fs/cgroup/memory.max` for cgroups v2 and set the node options accordingly? This way all users would benefit from the recommeded defaults?

afharo:
> Would this not be something that should be already set in the Kibana images?

Fair point! It would benefit all users, from ECK to "plain" Docker ones.
@elastic/kibana-operations, WDYT?

PR #8951, opened 2025-12-10
Add NODE_OPTIONS environment variable to Kibana pod template

pebrc:
- Introduced a new environment variable `NODE_OPTIONS` with a default value of `--max-old-space-size=75% of container memory limit` in the Kibana pod template.
- `--max-old-space-size-percentage` which would be simpler is not allowed in `NODE_OPTIONS`
- Existing customer provided values in `NODE_OPTIONS` and the `--max-old-space-size` value will be merged. 
- Added a helper function `assertEnvValue` to validate the presence and value of environment variables in container tests.
- Updated unit tests to check for the default `NODE_OPTIONS` and ensure user-provided overrides are respected.

Fixes https://github.com/elastic/cloud-on-k8s/issues/8621 as a temporary workaround until Kibana includes a more sensible default in the container image.

This can be considered a **breaking change**: existing users that have not manually configured the Kibana heap and are currently relying on the 50% heap size for off-heap use cases like PDF reporting might run into issues. Discussion with the Kibana team is still ongoing. I will old of merging this until we have clarity. 

The 75% value is taken from https://github.com/elastic/kibana/issues/245742

### DIFF
--- a/pkg/controller/kibana/pod.go
+++ b/pkg/controller/kibana/pod.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"fmt"
 	"strconv"
+	"strings"
 
 	"github.com/elastic/cloud-on-k8s/v3/pkg/controller/common/metadata"
 
@@ -43,6 +44,8 @@ const (
 	rewriteBasePathEnvName = "SERVER_REWRITEBASEPATH"
 	// maxOldSpacePercentage is the percentage of the container's memory limit to allocate to Node.js old space (heap)
 	maxOldSpacePercentage = 75
+	// maxOldSpaceSizeOption is the Node.js option name for setting the max old space size
+	maxOldSpaceSizeOption = "--max-old-space-size"
 )
 
 var (
@@ -101,7 +104,34 @@ func calculateMaxOldSpace(resources corev1.ResourceRequirements) string {
 	memoryMB := memoryLimit.Value() / (1024 * 1024)
 	maxOldSpaceSizeMB := memoryMB * maxOldSpacePercentage / 100
 
-	return fmt.Sprintf("--max-old-space-size=%d", maxOldSpaceSizeMB)
+	return fmt.Sprintf("%s=%d", maxOldSpaceSizeOption, maxOldSpaceSizeMB)
+}
+
+// mergeNodeOptions merges the calculated maxOldSpaceOption with any existing NODE_OPTIONS value
+// from the container's environment. If the user has already specified --max-old-space-size in their
+// NODE_OPTIONS, it will be preserved and not overwritten. Returns the updated environment slice.
+func mergeNodeOptions(containerEnv []corev1.EnvVar, maxOldSpaceOption string) []corev1.EnvVar {
+	// Look for existing NODE_OPTIONS in the container's env
+	for i, env := range containerEnv {
+		if env.Name == EnvNodeOptions {
+			// User has NODE_OPTIONS defined
+			if strings.Contains(env.Value, maxOldSpaceSizeOption) {
+				// User already specified max-old-space-size, don't modify it
+				return containerEnv
+			}
+			// Append the calculated max-old-space-size to the existing NODE_OPTIONS
+			mergedValue := strings.TrimSpace(env.Value)
+			if mergedValue != "" {
+				mergedValue = mergedValue + " " + maxOldSpaceOption
+			} else {
+				mergedValue = maxOldSpaceOption
+			}
+			containerEnv[i].Value = mergedValue
+			return containerEnv
+		}
+	}
+	// No existing NODE_OPTIONS, append it
+	return append(containerEnv, corev1.EnvVar{Name: EnvNodeOptions, Value: maxOldSpaceOption})
 }
 
 // readinessProbe is the readiness probe for the Kibana container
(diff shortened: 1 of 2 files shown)


prodsecmachine:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/50ea954e-4cce-4c05-b3c2-777b764fba7a?refs=f3847f7f6ef448e6df5b15cfbb575b5b043e666d&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/50ea954e-4cce-4c05-b3c2-777b764fba7a/license?refs=f3847f7f6ef448e6df5b15cfbb575b5b043e666d&source=prChecksComment) |

pebrc:
buildkite test this -f p=gke,E2E_TAGS=agent,s=9.3.0-SNAPSHOT

pkg/controller/kibana/pod.go
```diff
@@ -139,7 +139,9 @@ func NewPodTemplateSpec(
 		WithReadinessProbe(readinessProbe(kb.Spec.HTTP.TLS.Enabled(), basePath)).
 		WithVolumes(scriptsConfigMapVolume.Volume()).WithVolumeMounts(scriptsConfigMapVolume.VolumeMount()).
 		WithVolumes(PluginsVolume.Volume()).WithVolumeMounts(PluginsVolume.VolumeMount()).
-		WithPorts(ports)
+		WithPorts(ports).
+		// Temporary fix to expand the usable memory to 75% of the total memory for Kibana until https://github.com/elastic/kibana/issues/245742 is implemented.
```
barkbay: On 9.2.0:

```
/usr/share/kibana/bin/../node/glibc-217/bin/node: --max-old-space-size-percentage= is not allowed in NODE_OPTIONS
```

Not sure what I'm doing wrong...

pebrc: I might be the node JS version. I just took the recommendation from the Kibana issue without double checking.

pebrc: OK found it: https://nodejs.org/api/cli.html#node_optionsoptions

Only `--max-old-space-size` is allowed in NODE_OPTIONS. I am not sure I understand why the percentage value is considered a security issue but the absolute value is OK

pebrc:
buildkite test this -f p=gke,E2E_TAGS=agent,s=9.3.0-SNAPSHOT

pkg/controller/kibana/pod.go
```diff
@@ -85,6 +87,23 @@ type basePathConfig struct {
 	}
 }
 
+// calculateMaxOldSpace calculates the --max-old-space-size value based on the container's memory limit.
+// It returns a NODE_OPTIONS string with --max-old-space-size set to maxOldSpacePercentage of the memory limit (in MB).
+// If no memory limit is set, it returns an empty string.
```
pebrc, line 95: Not sure if we should have a default if no limits are set. But I guess the current default of ~500MB would be in effect in this case.

pkg/controller/kibana/pod.go
```diff
@@ -141,6 +160,15 @@ func NewPodTemplateSpec(
 		WithVolumes(PluginsVolume.Volume()).WithVolumeMounts(PluginsVolume.VolumeMount()).
 		WithPorts(ports)
 
+	// Calculate and set NODE_OPTIONS based on the container's memory limit (after resources have been merged)
+	// This is a temporary fix to expand the usable memory to 75% of the total memory for Kibana
+	// until https://github.com/elastic/kibana/issues/245742 is implemented.
+	if kbContainer := builder.MainContainer(); kbContainer != nil {
+		if nodeOptions := calculateMaxOldSpace(kbContainer.Resources); nodeOptions != "" {
+			builder = builder.WithEnv(corev1.EnvVar{Name: EnvNodeOptions, Value: nodeOptions})
```
barkbay: I think this is a no-op if the user has already set something for `NODE_OPTIONS`, for example:

```yaml
  podTemplate:
    metadata:
      labels:
        foo: bar
    spec:
      containers:
      - name: kibana
        env:
        - name: NODE_OPTIONS
          value: "--max-semi-space-size=256"
```

Unfortunately I'm not sure there is an easy way to detect/merge this. There is a precedent in https://github.com/elastic/cloud-on-k8s/blob/3cf9803bbc6239839f12c366b0e1933bac44b55a/pkg/controller/elasticsearch/nodespec/podspec.go#L236-L262

But it might be debatable whether we need this logic if it's just a temporary fix.

We could also document this, or send a warning if NODE_OPTIONS is already set, but it might be less "user-friendly".

pebrc: I was thinking if a user starts messing around with NODE_OPTIONS they are likely knowledgeable enough to configure the heap as well. 

But I can look into a merging user provided NODE_OPTIONS with this. If we get confirmation from the Kibana team that this will be fixed in 9.3.0 then we probably also want a version gate for this.

barkbay: > I was thinking if a user starts messing around with NODE_OPTIONS they are likely knowledgeable enough to configure the heap as well.

But the user would have to know that `max-old-space-size` was implicitly set, and it is their responsibility to manage it manually as soon as `NODE_OPTIONS` is set in the template? Also, if this is a "dynamic" value, I feel that it would be easy to change the limit in the resources but forget to update the max-old-space-size accordingly.

pebrc: Fair points. Let me see how messy this gets but I will give it a try.

pebrc: Merging implemented in f3847f7f6ef448e6df5b15cfbb575b5b043e666d

PR #8958, opened 2025-12-16
Support reloadable keystore for Elasticsearch 9.3 and later

pebrc:
## Summary

Implements reloadable keystore support for Elasticsearch 9.3+ clusters, enabling secure settings updates without pod restarts.

## Motivation

Previously, updating Elasticsearch secure settings (credentials, API keys, etc.) required recreating pods because the keystore was generated by an init container at pod startup. This caused unnecessary disruption for credential rotations.

Elasticsearch 9.3+ introduces an enhanced `_nodes/reload_secure_settings` API that returns keystore digests, enabling the operator to detect when all nodes have reloaded the updated keystore.

### Architecture

Instead of using init containers to create keystores at pod startup, this feature:

1. **Keystore Job**: Runs a Kubernetes Job in the operator namespace that creates the keystore file using the `keystore-uploader` subcommand
2. **Staging Secret**: The Job writes the keystore to a staging Secret in the operator namespace
3. **Final Secret**: The operator copies the staging Secret to the Elasticsearch namespace with proper owner references
4. **Volume Mount**: The keystore Secret is mounted directly into ES pods at `/usr/share/elasticsearch/config/elasticsearch.keystore`
5. **Hot Reload**: When secure settings change, a new Job creates an updated keystore, and the operator calls the reload API

### Security considerations

- Jobs run in the operator namespace with minimal RBAC (only permission to create the staging Secret)
- Secure settings are temporarily copied to the operator namespace during keystore creation. Not ideal.
- The final keystore Secret in the ES namespace is owned by the Elasticsearch resource

### Alternatives considered

- Run the job alongside Elasticsearch: needs dynamic creation of a service account. This seemed worse from a security perspective than running in the operator namespace
- Reimplement the keystore building in Go: not recommended by the ES team
- Use a seed pod in the ES cluster to create the keystore. Requires exec'ing into the pod. Seems brittle. Security posture is potentially worse by giving the operator exec permissions.

### New components

- **`cmd/keystore-uploader`**: New subcommand that creates keystore files and uploads them to Kubernetes Secrets
- **`pkg/controller/elasticsearch/keystorejob`**: Package for reconciling keystore Jobs, staging Secrets, and coordinating the workflow
- **`--operator-image` flag**: New operator configuration to specify the image used for keystore Jobs

### Modified components

- **Elasticsearch client**: `ReloadSecureSettings` now returns the response with keystore digests for convergence detection
- **ES driver**: Routes to Job-based or init-container-based keystore creation based on ES version
- **Pod spec**: Mounts keystore Secret as a volume for ES 9.3+ clusters
- **RBAC**: Added permissions for Job management in the operator namespace

### Feature controls

- Minimum version: Elasticsearch 9.3.0
- Can be disabled via annotation: `eck.k8s.elastic.co/disable-reloadable-keystore: "true"`
- Automatically falls back to init containers if operator image is not configured

## Testing

- Added e2e tests covering:
  - Initial keystore creation via Job
  - Secure settings updates with hot reload
  - Convergence detection across all nodes
  - Fallback behavior for older ES versions

Created using Cursor + claude-4.5-opus-high.

### DIFF
--- a/pkg/controller/elasticsearch/driver/driver.go
+++ b/pkg/controller/elasticsearch/driver/driver.go
@@ -699,7 +699,7 @@ func (d *defaultDriver) reconcileReloadableKeystore(
 	// If the keystore job is still running or the secret isn't ready yet, request a requeue
 	// so we can check again and eventually call the reload API.
 	if !keystoreReady {
-		results.WithReconciliationState(defaultRequeue.WithReason("Keystore job not yet complete"))
+		results.WithReconciliationState(defaultRequeue.WithReason("Keystore Job not yet complete"))
 	}
 	keystoreConfig := nodespec.KeystoreConfig{
 		SecretName: esv1.KeystoreSecretName(d.ES.Name),


prodsecmachine:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/25de3ff0-a489-4003-a89a-9aa19027bf56?refs=3774598fbbdae0c62c2d8afc415d149d9f022ee2&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/25de3ff0-a489-4003-a89a-9aa19027bf56/license?refs=3774598fbbdae0c62c2d8afc415d149d9f022ee2&source=prChecksComment) |

pebrc:
buildkite test this -f t=TestReloadableKeystor,p=gke,s=9.3.0-SNAPSHOT

pebrc:
buildkite test this -f t=TestReloadableKeystore,p=gke,s=9.3.0-SNAPSHOT

pebrc:
buildkite test this -f t=TestReloadableKeystore,s=9.3.0-SNAPSHOT -m p=gke,p=ocp,p=eks,p=aks

pebrc:
buildkite test this -f t=TestReloadableKeystore|TestUpdateESSecureSettings,p=gke,s=9.3.0-SNAPSHOT

pebrc:
buildkite test this -f t=TestUpdateESSecureSettings,p=gke,s=9.3.0-SNAPSHOT

pkg/controller/elasticsearch/driver/driver.go
```diff
@@ -661,6 +675,84 @@ func warnUnsupportedDistro(pods []corev1.Pod, recorder *events.Recorder) {
...
+	// Reconcile the keystore job
+	keystoreReady, err := d.reconcileKeystoreJob(ctx, keystoreResources, meta)
+	if err != nil {
+		return nodespec.KeystoreConfig{}, "", err
+	}
+
+	// If the keystore job is still running or the secret isn't ready yet, request a requeue
+	// so we can check again and eventually call the reload API.
+	if !keystoreReady {
+		results.WithReconciliationState(defaultRequeue.WithReason("Keystore job not yet complete"))
```
barkbay:
```suggestion
		results.WithReconciliationState(defaultRequeue.WithReason("Keystore Job not yet complete"))
```

pkg/controller/elasticsearch/driver/driver.go
```diff
@@ -661,6 +675,84 @@ func warnUnsupportedDistro(pods []corev1.Pod, recorder *events.Recorder) {
...
+	if esImage == "" {
+		esImage = container.ImageRepository(container.ElasticsearchImage, d.Version)
+	}
+
+	// Extract pod template params from the first node set for the Job
+	var podTemplateParams keystorejob.JobPodTemplateParams
+	if len(d.ES.Spec.NodeSets) > 0 {
+		podSpec := d.ES.Spec.NodeSets[0].PodTemplate.Spec
+		podTemplateParams = keystorejob.JobPodTemplateParams{
+			ImagePullSecrets:   podSpec.ImagePullSecrets,
```
barkbay, line 739: We are referencing a Secret that is namespaced to the Elasticsearch cluster. IIUC there is no guarantee that this Secret also exists in the operator namespace? (we also do not have a guarantee that the credentials are the same in all the namespaces)

pkg/controller/elasticsearch/driver/driver.go
```diff
@@ -661,6 +675,84 @@ func warnUnsupportedDistro(pods []corev1.Pod, recorder *events.Recorder) {
...
+	}
+
+	return keystoreConfig, keystoreSecretMountPath, nil
+}
+
+// reconcileKeystoreJob reconciles the keystore creation Job for Elasticsearch 9.3+ clusters.
+// It returns true if the keystore Secret is ready (job completed), false if still pending.
+func (d *defaultDriver) reconcileKeystoreJob(ctx context.Context, keystoreResources *keystore.Resources, meta metadata.Metadata) (bool, error) {
+	// Get the Elasticsearch image, preferring spec.image if set, otherwise using the default
+	esImage := d.ES.Spec.Image
```
barkbay, line 729: We are creating a Job using the desired Elasticsearch version, not necessarily the one that's actually running. I'm not sure if it's a problem in practice, but I was wondering if we always create 'backward compatible' keystores.

pkg/controller/elasticsearch/driver/driver.go
```diff
@@ -661,6 +675,84 @@ func warnUnsupportedDistro(pods []corev1.Pod, recorder *events.Recorder) {
...
+		esImage = container.ImageRepository(container.ElasticsearchImage, d.Version)
+	}
+
+	// Extract pod template params from the first node set for the Job
+	var podTemplateParams keystorejob.JobPodTemplateParams
+	if len(d.ES.Spec.NodeSets) > 0 {
+		podSpec := d.ES.Spec.NodeSets[0].PodTemplate.Spec
+		podTemplateParams = keystorejob.JobPodTemplateParams{
+			ImagePullSecrets:   podSpec.ImagePullSecrets,
+			PodSecurityContext: podSpec.SecurityContext,
```
barkbay, line 740: Certainly an edge case but IIRC, on OpenShift, [restricted workloads are not allowed to use a same user ID (`runAsUser`) across different namespaces](https://docs.redhat.com/en/documentation/openshift_container_platform/4.10/html/authentication_and_authorization/managing-pod-security-policies).

> The restricted SCC:
> [...]
> * Requires that a pod is run as a user in a pre-allocated range of UIDs

My recollection is that this "pre-allocated range of UIDs" is the one in `openshift.io/sa.scc.uid-range`, which would mean that we may not always reuse the `PodSecurityContext` in 2 different namespaces.

cmd/keystore-uploader/main.go
```diff
@@ -0,0 +1,223 @@
...
+		timeout         time.Duration
+	)
+
+	cmd := &cobra.Command{
+		Use:   "keystore-uploader",
+		Short: "Upload a keystore file to a Kubernetes Secret",
+		Long: `Upload a keystore file to a Kubernetes Secret.
+
+This command is used internally by ECK to upload keystore files created by
+the keystore init container to a staging Secret in the operator namespace.
```
barkbay, line 62: It took me a moment to understand which "init container" we were referring to 😄 Let's maybe explain that this is the one in the Job, not the one in the ES Pod?

cmd/keystore-uploader/main.go
```diff
@@ -0,0 +1,223 @@
...
+The operator then copies this Secret to the target ES namespace.`,
```
barkbay, line 63:
```suggestion
The operator then copies this Secret to the target Elasticsearch namespace.`,
```

deploy/eck-operator/templates/statefulset.yaml
```diff
@@ -70,6 +70,8 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: metadata.namespace
+            - name: OPERATOR_IMAGE
```
barkbay, line 73: I have to admit, I was quite happy [when we removed it](https://github.com/elastic/cloud-on-k8s/pull/3320) 😄  Having everything in a single process makes things easier in dev mode.

deploy/eck-operator/templates/_helpers.tpl
```diff
@@ -198,6 +198,18 @@ updating docs/operating-eck/eck-permissions.asciidoc file.
   - update
   - patch
   - delete
+- apiGroups:
+  - batch
+  resources:
+  - jobs
```
barkbay, line 204: I think we also need to update https://github.com/elastic/eck-diagnostics to include the Jobs.

pkg/controller/elasticsearch/keystorejob/keystorejob.go
```diff
@@ -0,0 +1,684 @@
...
+func ReconcileJob(ctx context.Context, params Params) (done bool, err error) {
+	span, ctx := apm.StartSpan(ctx, "reconcile_keystore_job", tracing.SpanTypeApp)
+	defer span.End()
+
+	log := ulog.FromContext(ctx)
+	es := params.ES
+
+	// Names for resources in the operator namespace (staging area)
+	stagingKeystoreSecretName := esv1.StagingKeystoreSecretName(es.Namespace, es.Name)
+	stagingSecureSettingsName := esv1.StagingSecureSettingsSecretName(es.Namespace, es.Name)
```
barkbay, line 142: Maybe I'm missing something but I can't find where the resources created here are "garbage collected" if the parent resource has been deleted while the operator was not running.

pkg/controller/elasticsearch/keystorejob/keystorejob.go
```diff
@@ -0,0 +1,684 @@
...
+		},
+	}
+	if err := client.Delete(ctx, &secret); err != nil && !apierrors.IsNotFound(err) {
+		return err
+	}
+	return nil
+}
+
+// DeleteStagingKeystoreSecret deletes the staging keystore Secret from the operator namespace if it exists.
+func DeleteStagingKeystoreSecret(ctx context.Context, client k8s.Client, operatorNamespace string, es esv1.Elasticsearch) error {
```
barkbay, line 681: This does not seem to be used, `deleteStagingSecret` seems to be invoked directly in several places instead.

pkg/controller/elasticsearch/keystorejob/keystorejob.go
```diff
@@ -0,0 +1,684 @@
...
+	for _, c := range job.Status.Conditions {
+		if c.Type == batchv1.JobFailed && c.Status == corev1.ConditionTrue {
+			return true
+		}
+	}
+	return false
+}
+
+// DeleteJob deletes the keystore Job from the operator namespace if it exists.
+func DeleteJob(ctx context.Context, client k8s.Client, operatorNamespace string, es esv1.Elasticsearch) error {
```
barkbay, line 651: This function does not seem to be used.

pkg/controller/elasticsearch/keystorejob/keystorejob.go
```diff
@@ -0,0 +1,684 @@
...
+		},
+	}
+	if err := client.Delete(ctx, &job); err != nil && !apierrors.IsNotFound(err) {
+		return err
+	}
+	return nil
+}
+
+// DeleteKeystoreSecret deletes the keystore Secret from the ES namespace if it exists.
+func DeleteKeystoreSecret(ctx context.Context, client k8s.Client, es esv1.Elasticsearch) error {
```
barkbay, line 666: Same as for `DeleteStagingKeystoreSecret`

pkg/controller/elasticsearch/keystorejob/keystorejob.go
```diff
@@ -0,0 +1,684 @@
...
+	// Job is for the current hash - check its status
+	if isJobComplete(existingJob) {
+		// Job completed - the staging secret should now exist
+		// Return to let the main reconciliation loop handle copying it
+		log.V(1).Info("Keystore job completed, staging secret should be ready")
+		return nil
+	}
+
+	if isJobFailed(existingJob) {
+		log.Info("Keystore job failed, deleting for retry")
```
barkbay, line 397: Would it deserve to be logged at the `error` level? Should we also include the `Message` field from the condition in the message?

barkbay (commented):
I spent a couple of hours on it but I still feel I’ve only scratched the surface 😅 , there are a lot of moving parts, but here are my initial thoughts/concerns:

* Job resource consumption: Elasticsearch clusters are going to "consume" CPU and memory in the operator namespace. This may interfere with Kubernetes cluster operators who rely on per-namespace resource limits to control resource usage (https://kubernetes.io/docs/concepts/policy/resource-quotas/). They now have to account for the Jobs that can be created for each Elasticsearch cluster.
* There is indeed a security concern in the fact that a service account which is able to read the Secrets in the operator namespace may have the opportunity to read any secure setting living in other namespaces.
* I think users cannot customize the Job (for example, adding more resources to the init container, set specific image pull secrets, customize the security context...). We try to infer the Pod template from one of the `nodeSet`, but I'm not sure if it is enough. Also `eck.k8s.alpha.elastic.co/propagate-*` do not seem to be honored.
* Non happy path: Because the Job is running in the operator namespace a user may have to involve the K8s cluster admins to understand why it is failing.

My first overall impression is that this introduces quite a bit of complexity to transform the contents of one Secret into another. I’d be interested in a comparison of the benefits versus drawbacks of this Job-based approach, relative to generating the keystore directly in Go (even if I can imagine some of them: duplicated implementation to be maintained, new interdependence/connection between the two projects...)

pkg/controller/elasticsearch/driver/driver.go
pebrc: Excellent point. I think we need to use the image of the running pod with the lowest running version.

pkg/controller/elasticsearch/driver/driver.go
pebrc: Another good point. I don't have a good solution for that at the moment other than copying pull secrets, which feels like an overreach for the operator to do.

pkg/controller/elasticsearch/keystorejob/keystorejob.go
pebrc: Good point. I think we need a GC routine for that.

pebrc:
> My first overall impression is that this introduces quite a bit of complexity to transform the contents of one Secret into another. I’d be interested in a comparison of the benefits versus drawbacks of this Job-based approach, relative to generating the keystore directly in Go (even if I can imagine some of them: duplicated implementation to be maintained, new interdependence/connection between the two projects...)

Thanks for your thorough review. I agree with your assessment and I should have probably left this PR in draft mode, given that this emerged from a vibe-coding experiment. However there is real user demand for the functionality so we should indeed discuss the alternatives seriously.

I think the main motivator for not attempting to reimplement the keystore format in Go was two-fold:
* the explicit recommendation by the Elasticsearch team not to attempt it
* the potential loss of forward compatibility with ES versions if we have our own implementation. Right now a new release of Elasticsearch can almost always be run on older ECK versions.


**ES Keystore File Format Stability (Last ~3 Years)**

[KeyStoreWrapper.java
Lines 147-154](https://github.com/elastic/elasticsearch/blob/899cfc85b878d68753f627bd1ff31e993b3b5d45/server/src/main/java/org/elasticsearch/common/settings/KeyStoreWrapper.java#L147-L154)
```
// format version changelog:
// 1: initial version, ES 5.3    
// 2: file setting, ES 5.4    
// 3: FIPS compliant algos, ES 6.3   
// 4: remove distinction between string/files, ES 6.8/7.1    
// 5: Lucene directory API changed to LE, ES 8.0    
// 6: increase KDF iteration count, ES 8.14    
// 7: increase cipher key length to 256 bits, ES 9.0
```

**Recent Changes:**
June 2022: Removed support for v1 and v2 formats entirely (min version = 3)
April 2024: Version 6 - Increased KDF iterations from 10,000 to 210,000 (ES 8.14)
January 2025: Version 7 - Increased cipher key from 128 to 256 bits (ES 9.0)

It seems the format changes not only on major version releases but also sometimes in between. 

Having said that, I assume that newer versions are still backwards compatible with older versions of the keystore. More critical for an ECK reimplementation would be the removal of support for older versions. This seems to happen very conservatively though (v1/v2 go back at least to 2017 and support for those was removed in 2022).

However, the drawback remains that the ECK team will have to track changes to the file format actively and adjust the implementation/maintain potentially multiple implementations in parallel. 

Judging from the version change log we would have to support at least versions 4-7.

barkbay:
> I agree with your assessment and I should have probably left this PR in draft mode, given that this emerged from a vibe-coding experiment.

That's fine, it is an interesting problem, and as you said, a real use case.

> Run the job alongside Elasticsearch: needs dynamic creation of a service account. This seemed worse from a security perspective than running in the operator namespace

Trying to challenge this, if we know the `Secret` name in advance the ECK operator could just create a `ServiceAccount` and the related `RoleBinding`:

1. A "human" operator would create the role when ECK is installed:

```yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: elasticsearch-keystore-creator
rules:
- apiGroups: [""]
  resources: ["secrets"]
  verbs: ["create", "get", "update"]
  resourceNames: ["elasticsearch-keystore"]  # Restrict to keystore Secret
```

2. The ECK operator would create the `ServiceAccount` and `RoleBinding` to create or update the Secret that holds the keystore**s** (plural, since there would be one Secret for multiple clusters).

```yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  name: elasticsearch-keystore-creator
  namespace: my-project
---
apiVersion: rbac.authorization.k8s.io/v1
kind: RoleBinding
metadata:
  name: elasticsearch-keystore-creator
  namespace: my-project
subjects:
- kind: ServiceAccount
  name: elasticsearch-keystore-creator
  namespace: my-project
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: elasticsearch-keystore-creator
```

One obvious drawback is that we need to extend the operator's permissions to create the `elasticsearch-keystore-creator` `ServiceAccount` and `RoleBinding`. However since K8s 1.28 (stable since 1.30) I think we can restrict the `RoleBinding` created by the operator, maybe with something along the line of:

```yaml
apiVersion: admissionregistration.k8s.io/v1
kind: ValidatingAdmissionPolicy
metadata:
  name: restrict-operator-rolebinding
spec:
  failurePolicy: Fail
  matchConstraints:
    resourceRules:
    - apiGroups: ["rbac.authorization.k8s.io"]
      apiVersions: ["v1"]
      resources: ["rolebindings"]
      operations: ["CREATE", "UPDATE"]
  matchConditions:
  - name: match-elastic-operator
    expression: "request.userInfo.username == 'system:serviceaccount:elastic-system:elastic-operator'"
  validations:
  - expression: "object.roleRef.kind == 'ClusterRole' && object.roleRef.name == 'elasticsearch-keystore-creator'"
    message: "RoleBinding must reference ClusterRole 'elasticsearch-keystore-creator'"
  - expression: "object.roleRef.apiGroup == 'rbac.authorization.k8s.io'"
    message: "RoleBinding must reference rbac.authorization.k8s.io apiGroup"
---
apiVersion: admissionregistration.k8s.io/v1
kind: ValidatingAdmissionPolicyBinding
metadata:
  name: restrict-operator-rolebinding-binding
spec:
  policyName: restrict-operator-rolebinding
  validationActions:
  - Deny
```

That way, I think it would mean that the operator can only create `rolebindings` to `ClusterRole` `elasticsearch-keystore-creator` which has a very limited scope of permissions (read/write Secrets named `elasticsearch-keystore`).

A last drawback is that the `Secret` `elasticsearch-keystore` would hold keystores for several Elasticsearch (for example one key per keystore), which means that the ECK operator would have to extract it into the "real" Elasticsearch keystore (for example `elasticsearch-sample-es-keystore`) from it.

We still need a Job, but the secrets stay in the same namespace, and it solves the `imagePullSecrets` problem.

pebrc:
> ```yaml
> apiVersion: admissionregistration.k8s.io/v1
> kind: ValidatingAdmissionPolicy
> metadata:
>   name: restrict-operator-rolebinding
> spec:
>   failurePolicy: Fail
>   matchConstraints:
>     resourceRules:
>     - apiGroups: ["rbac.authorization.k8s.io"]
>       apiVersions: ["v1"]
>       resources: ["rolebindings"]
>       operations: ["CREATE", "UPDATE"]
>   matchConditions:
>   - name: match-elastic-operator
>     expression: "request.userInfo.username == 'system:serviceaccount:elastic-system:elastic-operator'"
>   validations:
>   - expression: "object.roleRef.kind == 'ClusterRole' && object.roleRef.name == 'elasticsearch-keystore-creator'"
>     message: "RoleBinding must reference ClusterRole 'elasticsearch-keystore-creator'"
>   - expression: "object.roleRef.apiGroup == 'rbac.authorization.k8s.io'"
>     message: "RoleBinding must reference rbac.authorization.k8s.io apiGroup"

That's an interesting idea. I am not sure we can use it in OLM and we probably need some templating for things like namespaces, but it is definitely worth exploring.

pebrc:
Closing this because of the drawbacks discussed.

PR #8960, opened 2025-12-17
Add recipe for individual Logstash pipeline management via ConfigMaps

LolloneS:


prodsecmachine:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/b7b79ee3-51d2-400d-a042-317c4be02604?refs=27df1fe6d81a11daf173afdd7189563a7df8a408&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/b7b79ee3-51d2-400d-a042-317c4be02604/license?refs=27df1fe6d81a11daf173afdd7189563a7df8a408&source=prChecksComment) |

config/recipes/logstash/logstash-multi-pipelines-as-configmaps.yaml
```diff
@@ -0,0 +1,262 @@
...
+kind: Elasticsearch
+metadata:
+  name: elasticsearch
+spec:
+  version: 9.2.2
+  nodeSets:
+    - name: default
+      count: 3
+      config:
+        node.store.allow_mmap: false
```
pebrc, line 13: Missing the explanatory comment that other recipes include:

```yaml
config:
  # This setting has performance implications. See the README for more details.
  node.store.allow_mmap: false
```

config/recipes/logstash/logstash-multi-pipelines-as-configmaps.yaml
```diff
@@ -0,0 +1,262 @@
...
+kind: Secret
+metadata:
+  name: elasticsearch-credentials
+stringData:
+  es_username: "logstash_writer"
+  es_password: "changeme"
+---
+# Kafka credentials
+apiVersion: v1
+kind: Secret
```
pebrc, line 76: 🐛 **Bug: Secret name mismatch**

This Secret is named `kafka-credentials`, but the Logstash pod template below (lines 237-238) references `kafka-pipeline-secret`. This will cause the pod to fail to start.

config/recipes/logstash/logstash-multi-pipelines-as-configmaps.yaml
```diff
@@ -0,0 +1,262 @@
...
+      pipeline { address => "global_output_address" }
+    }
+    output {
+      # example: conditional output
+      if [pipeline_source] == "beats" {
+        elasticsearch {
+          hosts => ["${ES_HOSTS}"]
+          cacert => "${ES_SSL_CERTIFICATE_AUTHORITY}"
+          user => "${OUTPUT_ES_USER}"
+          password => "${OUTPUT_ES_PASSWORD}"
```
pebrc: 🐛 **Bug: Incorrect environment variable names**

Since `elasticsearchRefs` uses `clusterName: eck`, the auto-injected env vars are prefixed with `ECK_`. Also, the correct option name is `ssl_certificate_authorities`, not `cacert`.

```diff
- hosts => ["${ES_HOSTS}"]
- cacert => "${ES_SSL_CERTIFICATE_AUTHORITY}"
+ hosts => ["${ECK_ES_HOSTS}"]
+ ssl_certificate_authorities => "${ECK_ES_SSL_CERTIFICATE_AUTHORITY}"
```

See other recipes (e.g., `logstash-eck.yaml`) for the correct pattern.

config/recipes/logstash/logstash-multi-pipelines-as-configmaps.yaml
```diff
@@ -0,0 +1,262 @@
...
+              key: es_password
+        - name: KAFKA_BROKER_USER
+          valueFrom:
+            secretKeyRef:
+              name: kafka-pipeline-secret
+              key: kafka_username
+        - name: KAFKA_BROKER_PASSWORD
+          valueFrom:
+            secretKeyRef:
+              name: kafka-pipeline-secret
```
pebrc: 🐛 This references `kafka-pipeline-secret`, but the Secret created above (line 73) is named `kafka-credentials`.

config/recipes/logstash/logstash-multi-pipelines-as-configmaps.yaml
```diff
@@ -0,0 +1,262 @@
...
+  services:
+    - name: beats
+      service:
+        spec:
+          type: ClusterIP
+          ports:
+            - port: 5044
+              name: "filebeat"
+              protocol: TCP
+              targetPort: 5044
```
pebrc: Nit: Missing newline at end of file.

config/recipes/logstash/logstash-multi-pipelines-as-configmaps.yaml
LolloneS: Thanks! Addressed.

config/recipes/logstash/logstash-multi-pipelines-as-configmaps.yaml
LolloneS: Thanks! Addressed.

config/recipes/logstash/logstash-multi-pipelines-as-configmaps.yaml
LolloneS: Thanks! Addressed.

LolloneS:
Hi @pebrc and thank you for your comments. I addressed most of them and will update the README ASAP.

I'll try to answer your questions as well as I can.

> Is there a compelling reason to use ConfigMaps over Secrets here? Pipeline configs typically don't contain secrets since 
credentials are injected via env vars.

Reasons:

1. Simply said, pipelines are not secrets. Indeed, even in virtual machines, they are stored in plaintext. They might _reference_ some secrets such as credentials, but they are not anything that should be hidden. Actually, I believe we should have a `Pipeline` CR that can reference Secrets for credentials, but that's another story :) long story short, there is a mismatch between the "classic" Logstash experience and the ECK one.

2. ConfigMaps are easier to edit and manage (think GitOps approaches), and using them allows for separation of concerns. Why would you hide an entire pipeline rather than only hiding the credentials it uses? I would rather have a setup where the pipeline is represented as a ConfigMap managed e.g. via ArgoCD and the credentials are managed e.g. via the External Secrets Operator rather than having a massive, single Secret that I can't easily and dynamically edit. 

> If the goal is "modular, separate resources per pipeline," wouldn't multiple Secrets with pipelinesRef achieve the same with less complexity? 

No. Again, think about a GitOps setup. Why would you store the pipeline as an encrypted, unreadable object in Git (think SOPS encryption etc.) rather than having it readable and just loading the few required secrets dynamically? The approach with Secrets leads to _more_ complexity, not less. But I agree with the general point that we are introducing an overhead (ConfigMap + Secret) to make up for the absence of a `Pipeline` CR.

However, to your point, having multiple ConfigMaps (or `Pipeline` CRs) also indeed helps having a modular concept that's similar to what customers are used to when running Logstash on VMs rather than on K8s.

> If there's a strong use case for this pattern (e.g., company policies requiring ConfigMaps for non-sensitive data), please add a note in the recipe explaining when users should choose this approach over pipelinesRef.

I honestly believe the use-case is simply using the most appropriate tool for the job. This very point came up during a consulting engagement with a customer yesterday. They are heavy Kubernetes users and were pretty surprised that we don't allow for a custom CR or a ConfigMap to be used for configuring pipelines, because Secrets are a pain to work with and our current setup does not integrate well with GitOps workflows and external operators handling _real_ Secrets (i.e. credentials etc.).

config/recipes/logstash/logstash-multi-pipelines-as-configmaps.yaml
```diff
@@ -0,0 +1,263 @@
...
+  99-elasticsearch-output.conf: |
+    input {
+      # example: aggregates data from both Kafka (direct) and Beats (processed)
+      pipeline { address => "global_output_address" }
+    }
+    output {
+      # example: conditional output
+      if [pipeline_source] == "beats" {
+        elasticsearch {
+          hosts => ["${ES_HOSTS}"]
```
pebrc, line 162: I tihnk this incorrect an should be `ECK_ES_HOSTS`?

PR #8971, opened 2025-12-23
[Docs] Increase vm.max_map_count suggested value

shainaraskas:
We want to recommend a higher `vm.max_map_count` for Elasticsearch environments - additional context in issue

## What issues does this PR fix?

Part of https://github.com/elastic/docs-content-internal/issues/592

### DIFF
--- a/NOTICE.txt
+++ b/NOTICE.txt
@@ -1,4 +1,4 @@
-Copyright 2018-2025 Elasticsearch BV
+Copyright 2018-2026 Elasticsearch BV
 
 This product includes software developed by The Apache Software
 Foundation (http://www.apache.org/).


prodsecmachine:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/1d58337a-3ab0-4b8b-a940-fd6904d0d7a3?refs=aa2a2e3e9fbe846f90ad18153fa52912058dde9c&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/1d58337a-3ab0-4b8b-a940-fd6904d0d7a3/license?refs=aa2a2e3e9fbe846f90ad18153fa52912058dde9c&source=prChecksComment) |

marciw (approved):

pebrc:
I see another ~30 references to the old value in the examples in the code base that we should probably address https://github.com/search?q=repo%3Aelastic%2Fcloud-on-k8s%20262144&type=code

shainaraskas:
@pebrc let me know if editing those recipes is overkill. I'll have to recreate the changes in main and backport them to the 3.x branches as well

config/recipes/autopilot/elasticsearch.yaml
```diff
@@ -30,7 +30,7 @@ spec:
           securityContext:
             privileged: true
             runAsUser: 0
-          command: ['/usr/local/bin/bash', '-e', '-c', 'echo 262144 > /proc/sys/vm/max_map_count']
+          command: ['/usr/local/bin/bash', '-e', '-c', 'echo 1048576 > /proc/sys/vm/max_map_count']
```
pebrc: My apologies, but after testing this new setting. I think we have to stick with the old value for now. GKE Autopilot has a whitelist for this DaemonSet that is based on the exact hash/contents. Which means we have to stick with the old value until we talked to Google about whitelisting the version with the new value as well.

config/recipes/autopilot/elasticsearch.yaml
```diff
@@ -60,7 +60,7 @@ spec:
               cpu: 500m
         initContainers:
         - name: max-map-count-check
-          command: ['sh', '-c', "while true; do mmc=$(cat /proc/sys/vm/max_map_count); if [ ${mmc} -eq 262144 ]; then exit 0; fi; sleep 1; done"]
+          command: ['sh', '-c', "while true; do mmc=$(cat /proc/sys/vm/max_map_count); if [ ${mmc} -eq 1048576 ]; then exit 0; fi; sleep 1; done"]
```
pebrc: Same here.

config/recipes/autopilot/fleet-kubernetes-integration.yaml
```diff
@@ -30,7 +30,7 @@ spec:
           securityContext:
             privileged: true
             runAsUser: 0
-          command: ['/usr/local/bin/bash', '-e', '-c', 'echo 262144 > /proc/sys/vm/max_map_count']
+          command: ['/usr/local/bin/bash', '-e', '-c', 'echo 1048576 > /proc/sys/vm/max_map_count']
```
pebrc: This also needs to stay.

config/recipes/autopilot/fleet-kubernetes-integration.yaml
```diff
@@ -60,7 +60,7 @@ spec:
               cpu: 500m
         initContainers:
         - name: max-map-count-check
-          command: ['sh', '-c', "while true; do mmc=$(cat /proc/sys/vm/max_map_count); if [ ${mmc} -eq 262144 ]; then exit 0; fi; sleep 1; done"]
+          command: ['sh', '-c', "while true; do mmc=$(cat /proc/sys/vm/max_map_count); if [ ${mmc} -eq 1048576 ]; then exit 0; fi; sleep 1; done"]
```
pebrc: So does this.

docs/orchestrating-elastic-stack-applications/elasticsearch/virtual-memory.asciidoc
```diff
@@ -89,7 +91,7 @@ spec:
           securityContext:
             privileged: true
             runAsUser: 0
-          command: ['/usr/local/bin/bash', '-e', '-c', 'echo 262144 > /proc/sys/vm/max_map_count']
+          command: ['/usr/local/bin/bash', '-e', '-c', 'echo 1048576 > /proc/sys/vm/max_map_count']
```
pebrc: Image



I think one issue we have here is that the GKE Autopilot docs point to this guide. After we merge this PR the virtual memory docs will recommend a DaemonSet that is no longer working on GKE Autopilot due to the strict whitelisting GKE does, where any change compared to the approved DaemonSet will lead to rejection (see my earlier comment). 

I am not sure how to best address this. One idea is that we could duplicate the virtual memory DaemonSet example in the Autopilot docs page with the old value. We should also add an explanatory comment, saying that on GKE Autopilot only the DaemonSet spec previously approved by Google is accepted and therefore a lower value for `vm.max_map.count` has to be used. 

We can also open a follow-up issue to engage with Google to get a newer version of the DaemonSet approved.

pebrc (commented):
I did some testing on GKE Autopilot with the new values recommended in this PR. We have an issue here in that only the previously approved DaemonSet with the old value will be accepted. I think for this PR it means that we have to keep all the autopilot examples on the old value for the time being and make some adjustments to the virtual memory page to call out the special case. 

I am also wondering why we are backporting this change to the 2.16 branch instead of just updating the main branch/docs-content with the new values. BTW for the current documentation in docs-content we have the same issue with GKE Autopilot.

deploy/eck-stack/examples/elasticsearch/ingress/elasticsearch-ingress-gke.yaml
shainaraskas, line 1: is this GKE autopilot, or just regular GKE, or both?

deploy/eck-stack/examples/kibana/ingress/kibana-gke.yaml
shainaraskas, line 1: is this GKE autopilot, or just regular GKE, or both?

deploy/eck-stack/charts/eck-elasticsearch/examples/ingress/elasticsearch-ingress-gke.yaml
shainaraskas, line 1: is this GKE autopilot, or just regular GKE, or both?

docs/orchestrating-elastic-stack-applications/elasticsearch/virtual-memory.asciidoc
shainaraskas: @pebrc thanks for spotting this. 

I've made [the following changes](https://github.com/elastic/cloud-on-k8s/pull/8971/commits/575b7355e4dcca2cbaca3cea3fbdb1bcfc644c81) to this PR:

* in the virtual memory docs, I've called out that GKE autopilot requires a different value. we can go further here and give an error message in the GKE autopilot doc if you have one.
* reverted the autopilot recipe changes
* reverted the max map count change in the code snippet on the gke autopilot page

we do have three more examples that refer to GKE - not sure if these should also be reverted or have a comment added:
https://github.com/elastic/cloud-on-k8s/pull/8971/files#diff-77f1bb9410a4ea3c6099f4fbd5be3691caac04e5679bddf1f34a1b1b99842a13
https://github.com/elastic/cloud-on-k8s/pull/8971/files#diff-102d7a34c4c9589e03cb65318e05792acdf4f516677aadacee82093de7d4ea35
https://github.com/elastic/cloud-on-k8s/pull/8971/files#diff-992bd46742b2dfffd13f95ce73af64b9cdd867e18b9d7ae88a762394d8538195

do you think using the updated value for the other examples is safe? I think if we expand on the autopilot doc with the error message, it might be ok, but since this is a nice-to-have thing, we could revert the examples and just keep the recommended value in the guidance. let me know what you think.

shainaraskas:
> I am also wondering why we are backporting this change to the 2.16 branch instead of just updating the main branch/docs-content with the new values. BTW for the current documentation in docs-content we have the same issue with GKE Autopilot.

Fixing the issue in docs-content now (draft while we agree on language here: https://github.com/elastic/docs-content/pull/4490).

Backporting was requested [here](https://github.com/elastic/docs-content-internal/issues/592), because ECK 2.16 supports the impacted versions of ES, 2.16 is still in support, and this setting is controlled at the ECK level. Not exactly sure why 2.16 was the only version targeted - @kunisen, can you provide some insight on that?

kunisen:
Thanks @shainaraskas and @pebrc !

> Backporting was requested https://github.com/elastic/docs-content-internal/issues/592, because ECK 2.16 supports the impacted versions of ES, 2.16 is still in support, and this setting is controlled at the ECK level. Not exactly sure why 2.16 was the only version targeted - @kunisen, can you provide some insight on that?

I don't have strong opinion here and this is my thought:
* ECK 2.16 and 3.x are the two versions we are actively maintaining, means for both code and docs. See [EOL website page here](https://www.elastic.co/support/eol#:~:text=2.x-,15%2DJan%2D2027,-15%2DJul%2D2027).
* Other versions, although it's great to change the value, it might be hard for us to do all of them. (It happens to ECK that being able to run all 8.16+ versions, and that could also mean ECK 1.x version right? IIRC, there's no such [ECK version and stack version compatibility limitation](https://www.elastic.co/support/matrix) like we have in ECE)

Based on the above thoughts, the most cost-effective and also following EOL policy way is to do 2.16 and 3.x, IMHO.

---

FWIW, why 2.16 is the only maintained version? is because we only maintain that minor. Per [EOL policy](https://www.elastic.co/support/eol#:~:text=Elastic%20will%20provide%20Maintenance%20to%20the%20most%20recent%20two%20Minor%20Releases%20of%20the%20then%2Dcurrent%20Major%20Release%2C%20and%20the%20final%20Minor%20Release%20of%20the%20previous%20Major%20Release.),
* For ECK 2.x, we maintain 2.16 (latest minor)
* For ECK 3.x, we maintain 3.2 and 3.1 (latest 2 minors)

For the version that is under support but not under maintenance per [our EOL policy](https://www.elastic.co/support/eol#:~:text=15%2DApr%2D2025-,15%2DJan%2D2026,-APM%20Agents), such as ECK 1.9, I didn't include it as target. 

---

@shainaraskas @pebrc if doc team or ECK team think we should expand it to more targets, I think the support team is more than happy to have that. But that might be an overkill as you said.

I will leave the decision to you since you are more authoritative teams than us :) thank you!

docs/orchestrating-elastic-stack-applications/elasticsearch/virtual-memory.asciidoc
barkbay: Since this setting can now be configured using a GKE `ComputeClass`, let's maybe remove the `DaemonSet` from our samples and update the documentation accordingly? https://github.com/elastic/cloud-on-k8s/pull/8982

pebrc: This is a great finding @barkbay I did not realise this was possible. It seems that custom compute classes are a relatively new addition ([August 2024](https://docs.cloud.google.com/kubernetes-engine/docs/release-notes#August_13_2024)) and they are only supported for GKE 1.30+. 

This PR targets ECK 2.16 where we still need to support Kubernetes 1.27 where this feature is not available yet. 

Here is what I think we should do: 
- for this PR (2.16) if we really need to back port this vm.max_map_count change exclude the Autopilot examples. 
- for the current documentation 3.2. and newer let's switch to compute classes as illustrated by @barkbay 's PR

deploy/eck-stack/examples/elasticsearch/ingress/elasticsearch-ingress-gke.yaml
pebrc: We can assume that all examples that do not explicitly mention Autopilot are standard GKE only.

deploy/eck-stack/examples/kibana/ingress/kibana-gke.yaml
pebrc: Regular GKE

deploy/eck-stack/charts/eck-elasticsearch/examples/ingress/elasticsearch-ingress-gke.yaml
pebrc: We can assume that all examples that do not explicitly mention Autopilot are standard GKE only.

docs/orchestrating-elastic-stack-applications/elasticsearch/virtual-memory.asciidoc
shainaraskas: I see that @barkbay's PR still has `vm.max_map_count` set at 262144. Given the context of this PR, is that something that can be bumped to 1048576 now that `kind` is `ComputeClass`? 

regardless, will look into editing the latest docs to reflect using a ComputeClass for autopilot.

shainaraskas: if you're using a `ComputeClass` rather than a `DaemonSet `to set the value, can you still use the `initContainer` described in this section to check the value?

https://www.elastic.co/docs/deploy-manage/deploy/cloud-on-k8s/deploy-eck-on-gke-autopilot#k8s-autopilot-deploy-elasticsearch

barkbay: > I see that @barkbay's PR still has vm.max_map_count set at 262144. Given the context of this PR, is that something that can be bumped to 1048576 now that kind is ComputeClass?

🤦 , good catch, sorry, I was more focused on the "how" than on the "what". Here is the fix: https://github.com/elastic/cloud-on-k8s/pull/8986

> if you're using a ComputeClass rather than a DaemonSet to set the value, can you still use the initContainer described in this section to check the value?

During my experiments this was set before Pods were scheduled, so the init container is not needed.

shainaraskas: I've made the required changes to this PR, and I've added instructions for using a custom ComputeClass in 3.2+ over here: https://github.com/elastic/docs-content/pull/4490

would love a review on both when time allows. thanks for your help so far.

docs/orchestrating-elastic-stack-applications/elasticsearch/virtual-memory.asciidoc
```diff
@@ -9,9 +9,11 @@ endif::[]
...
+Default values for virtual address space on Linux distributions can be too low for Elasticsearch to work properly, which may result in out-of-memory exceptions. This is why link:k8s-quickstart.html[the quickstart example] disables `mmap` through the `node.store.allow_mmap: false` setting. For production workloads, we recommend that you increase the kernel setting `vm.max_map_count` to `1048576` and leave `node.store.allow_mmap` unset.
 
-The kernel setting `vm.max_map_count=262144` can be set on the host directly, by a dedicated init container which must be privileged, or a dedicated Daemonset.
+The kernel setting `vm.max_map_count=1048576` can be set on the host directly, by a dedicated init container which must be privileged, or a dedicated Daemonset.
+
+IMPORTANT: For {{es}} version 8.16 and later, set the `vm.max_map_count` kernel setting to `1048576`; for {{es}} version 8.15 and earlier, set `vm.max_map_count` to `262144`. The exception is in GKE Autopilot environments, where `vm.max_map_count` must always be set to `262144`.
```
pebrc: Should we explain why? (DaemonSet approved by Google has the old value)

shainaraskas: is this copy accurate or have I warped it? :/

```
IMPORTANT: For {{es}} version 8.16 and later, set the `vm.max_map_count` kernel setting to `1048576`; for {{es}} version 8.15 and earlier, set `vm.max_map_count` to `262144`. The exception is in GKE Autopilot environments, where `vm.max_map_count` must always be set to `262144`. This is because any custom DaemonSet in GKE Autopilot must match the DaemonSet pre-approved by Google.
```

docs/orchestrating-elastic-stack-applications/elasticsearch/virtual-memory.asciidoc
```diff
@@ -9,9 +9,11 @@ endif::[]
...
+Default values for virtual address space on Linux distributions can be too low for Elasticsearch to work properly, which may result in out-of-memory exceptions. This is why link:k8s-quickstart.html[the quickstart example] disables `mmap` through the `node.store.allow_mmap: false` setting. For production workloads, we recommend that you increase the kernel setting `vm.max_map_count` to `1048576` and leave `node.store.allow_mmap` unset.
 
-The kernel setting `vm.max_map_count=262144` can be set on the host directly, by a dedicated init container which must be privileged, or a dedicated Daemonset.
+The kernel setting `vm.max_map_count=1048576` can be set on the host directly, by a dedicated init container which must be privileged, or a dedicated Daemonset.
+
+IMPORTANT: For {{es}} version 8.16 and later, set the `vm.max_map_count` kernel setting to `1048576`; for {{es}} version 8.15 and earlier, set `vm.max_map_count` to `262144`. The exception is in GKE Autopilot environments, where `vm.max_map_count` must always be set to `262144`. This is because any custom DaemonSet in GKE Autopilot must match the DaemonSet pre-approved by Google.
```
pebrc:
```suggestion
IMPORTANT: For {{es}} version 8.16 and later, set the `vm.max_map_count` kernel setting to `1048576`; for {{es}} version 8.15 and earlier, set `vm.max_map_count` to `262144`. The exception is in GKE Autopilot environments, where `vm.max_map_count` must always be set to `262144`. This is because any privileged DaemonSet in GKE Autopilot must match the DaemonSet pre-approved by Google.
```

shainaraskas:
@pebrc I'm sorry to bug you again, but I'm blocked from merging as well (not sure if it's due to insufficient permissions, or some other conflict between my branch and this repo)

image

I've [reopened the PR from a fork](https://github.com/elastic/cloud-on-k8s/pull/9004), but am running into the same issue

I also [reached out in the channel](https://elastic.slack.com/archives/C8RU9J1C2/p1767899552433949) because it looks like perhaps permissions to the repo have been changed

pebrc:
Apologies for the delay. I merged your PR. We have additional branch protections on release branches, that prevent you from merging.

PR #8978, opened 2025-12-30
feat: add logstash ingress

hoodadt:
Add ingress support to Logstash Helm chart with default ls-api backend and extra path backends; document values.yaml options


prodsecmachine:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/264308a6-27e2-4380-ab6b-2b7c9b392abc?refs=6e1f3281705a8602d16e000174153acf088c8145&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/264308a6-27e2-4380-ab6b-2b7c9b392abc/license?refs=6e1f3281705a8602d16e000174153acf088c8145&source=prChecksComment) |

deploy/eck-stack/charts/eck-logstash/templates/ingress.yaml
pebrc, line 1: The eck-elasticsearch and eck-kibana charts both have templates/tests/ingress_test.yaml. This PR should include one too imo.

deploy/eck-stack/charts/eck-logstash/values.yaml
```diff
@@ -113,3 +113,79 @@ services: []
...
+  #
+  # *NOTE* Many Cloud Service Providers handle TLS in a custom manner, and as such, it is recommended to consult their documentation.
+  # Notably GKE and Nginx Ingress Controllers seems to respect the Ingress TLS settings, AKS and EKS ignore it.
+  #
+  # - AKS:   https://learn.microsoft.com/en-us/azure/aks/app-routing-dns-ssl
+  # - GKE:   https://cloud.google.com/kubernetes-engine/docs/concepts/ingress#options_for_providing_ssl_certificates
+  # - EKS:   https://aws.amazon.com/blogs/containers/serve-distinct-domains-with-tls-powered-by-acm-on-amazon-eks/
+  # - Nginx: https://kubernetes.github.io/ingress-nginx/user-guide/tls/
+  #
+  # Kubernetes ingress TLS documentation:
```
pebrc, line 188: Both ES and Kibana have examples/ingress/ directories with cloud-specific examples. Consider adding:
examples/ingress/logstash-ingress-gke.yaml
examples/ingress/logstash-ingress-aks.yaml
examples/ingress/logstash-ingress-eks-alb.yaml

deploy/eck-stack/charts/eck-logstash/templates/ingress.yaml
```diff
@@ -0,0 +1,57 @@
...
+  {{- range .Values.ingress.hosts }}
+  {{- $hostPath := .path | default "/" }}
+  - host: {{ .host | quote }}
+    http:
+      paths:
+      - path: {{ $hostPath }}
+        pathType: {{ $pathType }}
+        backend:
+          service:
+            name: {{ include "logstash.fullname" $ }}-ls-api
```
pebrc: I find this default backend problematic: it routes to -ls-api on port 9600 (monitoring API)
Users likely want to only expose data pipelines (beats input, http input).
The monitoring API should probably be used via the k8s-internal service, not public ingress. At least there should be an option to turn this off. There is also a security aspect: The API exposes metrics and node info that some users may not want public and this API can be used without authentication.

deploy/eck-stack/charts/eck-logstash/templates/ingress.yaml
hoodadt: agreed. I’ve added an ingress test under templates/tests/ingress_test.yaml

deploy/eck-stack/charts/eck-logstash/values.yaml
hoodadt: I’ve added GKE, AKS, and EKS (ALB - NLB)

deploy/eck-stack/charts/eck-logstash/templates/ingress.yaml
hoodadt: I’ve adjusted things so the ingress focuses on exposing pipeline inputs by default.

hoodadt:
Could you please review the latest changes and let me know if there’s anything else you’d like adjusted or improved?

deploy/eck-stack/charts/eck-logstash/examples/ingress/logstash-ingress-gke.yaml
```diff
@@ -0,0 +1,57 @@
...
+          # This is required for `ClusterIP` services (which are the default ECK service type) to be used with Ingress in GKE clusters.
+          cloud.google.com/neg: '{"ingress": true}'
+          # Inform the GKE Ingress Controller that the backend speaks TCP (Beats).
+          cloud.google.com/app-protocols: '{"filebeat":"TCP"}'
+      spec:
+        type: ClusterIP
+        ports:
+          - name: filebeat
+            port: 5044
+            protocol: TCP
```
pebrc, line 47: I don't think this example works: 
```
Translation failed: invalid port application protocol: TCP
```
GKE Ingress is HTTP only. Gateway API is the way to go here. I wonder if we actually should continue adding ingress support the Helm chart.

deploy/eck-stack/charts/eck-logstash/examples/ingress/logstash-ingress-eks-alb.yaml
```diff
@@ -0,0 +1,67 @@
...
+            targetPort: 5044
+
+ingress:
+  enabled: true
+  className: alb
+  annotations:
+    alb.ingress.kubernetes.io/scheme: "internet-facing"
+    alb.ingress.kubernetes.io/listen-ports: '[{"HTTPS":443}]'
+    alb.ingress.kubernetes.io/backend-protocol: "TCP"
+    alb.ingress.kubernetes.io/target-type: "ip"
```
pebrc, line 52: I don't this will work either ALBs are HTTP only. NLB should work though

deploy/eck-stack/charts/eck-logstash/examples/ingress/logstash-ingress-aks.yaml
```diff
@@ -0,0 +1,52 @@
...
+
+services:
+  - name: beats
+    service:
+      spec:
+        type: ClusterIP
+        ports:
+          - name: filebeat
+            port: 5044
+            protocol: TCP
```
pebrc, line 41: Same here TCP won't work

pebrc (commented):
Looking at the examples you added I realised that the old ingress API is probably not a good fit for many Logstash use cases. I have doubts that added ingress to this Helm chart is very useful especially given that we have the Gateway API now that covers both HTTP and TCP use cases.

hoodadt:
It took me quite some time to figure out how to move to TCP, but I realize now that I was wrong.
Switching to HTTP should help resolve the issue. The Gateway API is a great option for TCP, but adding ingress support for Helm chart would still be valuable, especially for legacy on-prem clusters. Additionally, Fluentd which is CNCF certified works well with HTTP. I think it would be beneficial to continue with that approach.

pkoutsovasilis:
I agree with @pebrc's point here, and I'd actually take it a step further: I believe neither Ingress nor Gateway API resources should be part of our Helm charts directly. The challenge is that networking setups vary significantly across different environments and infrastructures. What works for one user's cluster (legacy on-prem with  traditional Ingress controllers) won't work for another (modern clusters with Gateway API implementations). Baking either approach into our charts means we're making opinionated choices that won't fit everyone's needs. That said, I do hear the valid need users have to be able to specify these resources alongside their deployments. My answer to that is the extraObjects pattern that  #9069 introduces. Once merged, any user will be able to define their own Ingress, Gateway, HTTPRoute, TCPRoute, or whatever networking resources their infrastructure requires - with full Helm templating support - and have them managed as part of the same Helm release lifecycle.

This gives users complete flexibility to, use traditional Ingress for legacy clusters, use Gateway API for modern setups, etc. @pebrc thoughts on this approach?

hoodadt:
I think we’re actually very aligned on the core problem: networking is highly environment-specific, and hard-coding assumptions into charts is a losing game long-term.

Where I’d slightly push back is that adding Ingress doesn’t have to mean we’re endorsing a single, opinionated networking model. Framed correctly, it’s more about providing a well-documented, supported option for the large set of users who do rely on traditional Ingress today, while still making it clear that it’s optional and overridable.

The extraObjects pattern you mention is a great escape hatch, and I see it as complementary rather than mutually exclusive. It gives power users full flexibility (Gateway API, custom HTTPRoutes, bespoke setups), but it also asks every user to roll their own networking from scratch, even for the most common case. That’s a pretty big ergonomics hit for newcomers and “just make it work” installs.

My view is Ingress support as an optional, values-driven configuration for users who need it and extraObjects for advanced/custom setups

As long as the Ingress bits are optional, values-driven, and not blocking other approaches, I think they strike a good balance between usability and flexibility. Removing them entirely feels like optimizing for the most complex clusters at the expense of the majority.

pebrc:
> This gives users complete flexibility to, use traditional Ingress for legacy clusters, use Gateway API for modern setups, etc. @pebrc thoughts on this approach?

I agree with your assessment @pkoutsovasilis. 

I am sorry @hoodadt for asking you for changes on this PR only to then realise later that the approach itself is not a direction we want to go in. 

I think the benefit in templating that the explicit Ingress support offers is negligible. The cognitive load of a templated ingress vs. a hand rolled one in `extraObjects` is very similar. Users will still have to understand the Ingress concept and meaning of all attributes in both cases. `extraObjects` will be a bit more verbose but we can add example configurations that allow a quick start for common use cases.  But on the other hand `extraObjects` comes with the massive benefit of not being opinionated about the network setup of the user. Especially for applications like Logstash Ingress will not be the right choice for many. We also have to keep the maintanability of the Helm chart itself in mind and try to the keep the number of possible ways of doing the same thing in check.

hoodadt:
thanks for the changes you asked me to make even though we’re not moving forward with this approach, going through that exercise was genuinely helpful. It helped me learn how to look at the problem space more broadly, so I really appreciate that.

hoodadt:
thanks for the changes you asked me to make even though we’re not moving forward with this approach, going through that exercise was genuinely helpful. It helped me learn how to look at the problem space more broadly, so I really appreciate that.

PR #9031, opened 2026-01-13
[deployer]: fix macos local kind execution

moukoublen:
Fix local kind execution. 

Mac docker desktop (since a specific release) uses a symlink in `/var/run/docker.sock` that points to `${HOME}/.docker/run/docker.sock`.

With the previous check, the initial link `/var/run/docker.sock` is considered valid and is passed into the docker commands `-v /var/run/docker.sock:/var/run/docker.sock`, but inside the container, the d2d tries to resolve the link and fails. 

I added a safeguard follow-link logic to cover other versions of Mac Docker Desktop, since I am not sure which version `/var/run/docker.sock` became a link to the home and vice versa. 

This PR updates the docker cli version of `docker.elastic.co/eck-ci/deployer-kind` because it was no longer working with the latest docker engine.

### DIFF
--- a/hack/deployer/runner/docker.go
+++ b/hack/deployer/runner/docker.go
@@ -7,7 +7,6 @@ package runner
 import (
 	"errors"
 	"os"
-	"runtime"
 )
 
 const defaultDockerSocket = "/var/run/docker.sock"
@@ -15,26 +14,17 @@ const defaultDockerSocket = "/var/run/docker.sock"
 var homeDockerSocket = os.ExpandEnv("${HOME}/.docker/run/docker.sock")
 
 func getDockerSocket() (string, error) {
-	if runtime.GOOS == "darwin" {
-		sck, err := followLink(defaultDockerSocket)
-		if err != nil {
-			hsc, hErr := followLink(homeDockerSocket)
-			if hErr != nil {
-				return "", errors.Join(err, hErr)
-			}
-
-			return hsc, nil
-		}
-
+	sck, err := followLink(defaultDockerSocket)
+	if err == nil { // if *not* error, return the socket
 		return sck, nil
 	}
 
-	_, err := os.Stat(defaultDockerSocket)
-	if err != nil {
-		return "", err
+	hsc, hErr := followLink(homeDockerSocket)
+	if hErr != nil {
+		return "", errors.Join(err, hErr)
 	}
 
-	return defaultDockerSocket, nil
+	return hsc, nil
 }
 
 func followLink(path string) (string, error) {


prodsecmachine:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/79adf2aa-ffa1-44b8-a0c8-9d8012f5e49d?refs=5a5e51301f9f1da80340b7c82397fc46d5616753&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/79adf2aa-ffa1-44b8-a0c8-9d8012f5e49d/license?refs=5a5e51301f9f1da80340b7c82397fc46d5616753&source=prChecksComment) |

hack/deployer/runner/docker.go
```diff
@@ -11,18 +11,35 @@ import (
...
 	return defaultDockerSocket, nil
 }
+
+func followLink(path string) (string, error) {
+	info, err := os.Lstat(path)
+	if err != nil {
+		return path, err
+	}
+
+	if isLink := info.Mode()&os.ModeSymlink != 0; !isLink {
```
naemono: I think you can adjust this to the below, but honestly I think the way you wrote makes it easier to understand... wouldn't mind other's input.
```
if info.Mode()&os.ModeSymlink == 0 {
```

pkoutsovasilis: I see the readability improvement by having `isLink` variable but since this isn't actually used I would resort to a comment and go without it
```suggestion
    // check if file is a symlink
	if info.Mode()&os.ModeSymlink != 0 {
```

hack/deployer/runner/docker.go
```diff
@@ -11,18 +11,35 @@ import (
 
 const defaultDockerSocket = "/var/run/docker.sock"
 
+var homeDockerSocket = os.ExpandEnv("${HOME}/.docker/run/docker.sock")
+
 func getDockerSocket() (string, error) {
+	if runtime.GOOS == "darwin" {
+		sck, err := followLink(defaultDockerSocket)
+		if err != nil {
```
pkoutsovasilis: should we fallback to `homeDockerSocket` for any err or just for `errors.Is(err, fs.ErrNotExist)`, I am thinking it should be the latter.

hack/deployer/runner/docker.go
moukoublen: Yes, the initial thought was to have a "named" variable for the bit operations magic, so to be clear to someone what this section checks. But I could simplify it.

hack/deployer/runner/docker.go
moukoublen: My only concern is why not have a "best effort" fallback? Like, if anything goes wrong trying to resolve/follow the `/var/run/docker.sock`, at least let's try the home one before we give up.

pkoutsovasilis: IMO this isn't a best effort fallback, this should be a fallback only when `defaultDockerSocket` doesn't exist. In any other case the initial err, which in this PR is completely masked, should be returned so we can clearly see what when wrong. my 2 cents

moukoublen: Sure, that can be solved by joining the errors. How about doing that? 
My reasoning is that it seems the "real" socket is inside `${HOME}/.docker/run/docker.sock` in Mac Docker Desktop. I cannot find a clear indication when that happened or if it is always the case. But due to that I think we should at least try to resolve that before we give up. 

Or we could reverse the order. Try the home first, then fall back to the root.

hack/deployer/runner/docker.go
```diff
@@ -5,24 +5,48 @@
...
 			}
-		} else {
-			// Handle other errors
-			return "", err
+
+			return hsc, nil
 		}
+
+		return sck, nil
+	}
```
pkoutsovasilis: the nested ifs are getting kinda verbose. I am also doubting if guarding this only for `darwin` is necessary; there is no reason this won't work as expected on `linux` as well

```suggestion
    sck, err := followLink(defaultDockerSocket)
	if err == nil {
	    return sck, nil
	}
	hsc, hErr := followLink(homeDockerSocket)
	if hErr != nil {
	    return "", errors.Join(err, hErr)
	}
	return hsc, nil
```

moukoublen: Somehow, the suggestion block cannot be merged (most probably because it contains both removed and added lines),

> Applying suggestions on deleted lines is currently not supported.


I committed separately and added a clarification comment because sometimes `err == nil` can be misread by people (I got caught in that before 😅).

naemono:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|3.3|[](https://github.com/elastic/cloud-on-k8s/pull/9097)|

### Questions ?

PR #9051, opened 2026-01-19
Add generic expectation mechanism

barkbay:
### Summary

Refactors the expectations mechanism to work with any Kubernetes resource type, not just `StatefulSets`. This is achieved using a prototype object pattern instead of Go generics, making the code more reusable while maintaining backward compatibility.

### Changes

- **Renamed** `ExpectedStatefulSetUpdates` → `ExpectedGenerations` to reflect the generic nature
- **Updated** `NewExpectedGenerations()` to accept a `client.Object` prototype that defines the resource type to track
- **Modified** all call sites to pass the appropriate object type (currently `&appsv1.StatefulSet{}`)
- **Generalized** the implementation to use `metav1.Object` and `client.Object` interfaces throughout
- **Updated** documentation and comments to remove StatefulSet-specific references

### Approach

The implementation uses a prototype object pattern:
1. Store a `client.Object` instance that represents the type to track
2. Use `DeepCopyObject()` to create new instances when checking the cache
3. Leverage interface methods (`GetGeneration()`, `GetUID()`, etc.) for type-agnostic operations

This approach avoids Go generics while achieving the same flexibility, similar to how controller-runtime's fake client works internally.

I decided not to use Go generics here as the type parameter would propagate through many layers (ClustersExpectation, controller structs, etc.) while the type is only needed for a single internal call in `generationSatisfied`. Happy to revisit it though.

### Future Work

This foundation allows us to easily extend expectations to other resource types (for example `Deployments`) by simply passing a different prototype object during initialization.


prodsecmachine:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/118c34e6-f68c-4444-a8d6-c7add522247a?refs=a982396c05ca96adf59e96da5cf30e5b34f237b1&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/118c34e6-f68c-4444-a8d6-c7add522247a/license?refs=a982396c05ca96adf59e96da5cf30e5b34f237b1&source=prChecksComment) |

pkg/controller/common/expectations/generations.go
```diff
@@ -6,90 +6,102 @@ package expectations
...
 	Generation int64
 }
 
-// NewExpectedStatefulSetUpdates returns an initialized ExpectedStatefulSetUpdates.
-func NewExpectedStatefulSetUpdates(client k8s.Client) *ExpectedStatefulSetUpdates {
-	return &ExpectedStatefulSetUpdates{
+// NewExpectedGenerations returns an initialized ExpectedGenerations.
+func NewExpectedGenerations(client k8s.Client, object client.Object) *ExpectedGenerations {
+	return &ExpectedGenerations{
+		object:      object,
```
pkoutsovasilis, line 38: here we essentially pass `object` just for calling `object, ok := e.object.DeepCopyObject().(client.Object)` [here](https://github.com/elastic/cloud-on-k8s/pull/9051/changes#diff-965fc7830598c317ab3ac3d3768f4865edb0c0059f33b53192fdc1825f81fd16R76). Wouldn't be more efficient in terms of memory consumption (we don't keep referencing any ptr to a k8s object) and computation (no need to DeepClone) to adjust the func signature to something like below

```suggestion
func clientObjectAlloc[T any, PT interface {
	*T
	client.Object
}]() client.Object {
	var t T
	return PT(&t)
}

func NewExpectedGenerations(client k8s.Client, clientObjectAlloc func() client.Object) *ExpectedGenerations {
	return &ExpectedGenerations{
		objectAlloc:  clientObjectAlloc,
```

in this way we can instantiate NewExpectedGenerations like this
```
NewExpectedGenerations(k8sClient, clientObjectAlloc[appsv1.StatefulSet])
```

I don't have any strong feelings about that, and the rest of the code checks out. Let me know what you think

pebrc: I think the trade-off here is between the micro optimisation of saving memory for one empty struct and the less clear method signature. If we want to follow the recommendation which I am not opposed to, I would tweak the naming: 

```
NewExpectedGenerations(k8sClient, forObject[appsv1.StatefulSet])
```

and call the argument maybe `newObjectFn`

pebrc (approved):

pkg/controller/common/expectations/generations.go
pkoutsovasilis: yes when this is an empty struct this is a micro optimisation; however, there is nothing preventing the caller to pass a full struct as done in [here](https://github.com/elastic/cloud-on-k8s/blob/7496152b76555ad24240d63aafbdfa95b35345cf/pkg/controller/elasticsearch/driver/downscale.go#L305). So should we update that part of the code to pass a `&appsv1.StatefulSet{}` and maybe add a godoc that the caller should pass empty structs?

PS: no hard opinions on the naming 🙂

barkbay: > yes when this is an empty struct this is a micro optimisation; however, there is nothing preventing the caller to pass a full struct as done in [here](https://github.com/elastic/cloud-on-k8s/blob/7496152b76555ad24240d63aafbdfa95b35345cf/pkg/controller/elasticsearch/driver/downscale.go#L305).
```golang
	// Expect the updated statefulset in the cache for next reconciliation.
	downscaleCtx.expectations.ExpectGeneration(&downscale.statefulSet)
```

Sorry, maybe I'm missing something, `ExpectGeneration` only sets an expectation, without creating a copy of that object. We only create a copy of the empty struct passed in `NewExpectedGenerations` when checking the expectation.

pkoutsovasilis: oops I messed up the func calls. Either way the fact that a caller can pass a full struct is valid, so at least extend the godoc to capture that the caller should pass empty structs

pkoutsovasilis (approved):

barkbay:
Thanks for the review @pebrc @pkoutsovasilis 🙇

PR #9083, opened 2026-01-28
Go modernize

moukoublen:
Run [modernize](https://pkg.go.dev/golang.org/x/tools@v0.41.0/go/analysis/passes/modernize) `go run golang.org/x/tools/go/analysis/passes/modernize/cmd/modernize@latest -fix -test ./...` for each go module.

Golangci-lint linter [modernizer](https://golangci-lint.run/docs/linters/configuration/#modernize) added to config.

## Update 
Separate commits were forced-pushed for easier review. 

1. add go-modernize recipe and script [b8d7e08](https://github.com/elastic/cloud-on-k8s/pull/9083/commits/b8d7e085848d18d44b037cee17e77a7198179a81)
2. run go modernize with `-any=true` [5dbbe58](https://github.com/elastic/cloud-on-k8s/pull/9083/commits/5dbbe58e31fa2d2118bf5e2afc71e669222f326a) (change `interface{}` -> `any` everywere; big and noisy) 
3. run go modernize [1195cbc](https://github.com/elastic/cloud-on-k8s/pull/9083/commits/1195cbcdd59012e6728529d3f43cadd86e02380d) (all the other analyzers)
4. clean-ups [bcf0adc](https://github.com/elastic/cloud-on-k8s/pull/9083/commits/bcf0adce6a1da78bd4ca08fe17b102bf47fcb0a7)


prodsecmachine:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/f605d454-26b4-4d63-82a1-6dcf41b828ca?refs=e51639b60256d8bf227b96e452529f298c3dd489&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/f605d454-26b4-4d63-82a1-6dcf41b828ca/license?refs=e51639b60256d8bf227b96e452529f298c3dd489&source=prChecksComment) |

pkg/apis/common/v1/validations.go
```diff
@@ -30,8 +30,8 @@ func NoUnknownFields(dest runtime.Object, meta metav1.ObjectMeta) field.ErrorLis
 		if err := d.Decode(dest.DeepCopyObject()); err != nil {
 			errString := err.Error()
 			unknownPrefix := "json: unknown field "
-			if strings.HasPrefix(errString, unknownPrefix) {
-				fld := strings.TrimPrefix(errString, unknownPrefix)
+			if after, ok0 := strings.CutPrefix(errString, unknownPrefix); ok0 {
```
pkoutsovasilis:
```suggestion
			if after, exists := strings.CutPrefix(errString, unknownPrefix); exists {
```

pkg/controller/autoscaling/elasticsearch/autoscaler/recommender/recommender.go
```diff
@@ -109,18 +109,13 @@ func getResourceValue(
...
-	if nodeResource < quantityRange.Min.Value() {
-		nodeResource = quantityRange.Min.Value()
-	}
-
-	// Resource has been rounded up or scaled up to meet the tier requirements. We need to check that those operations
-	// do not result in a resource quantity which is greater than the max. limit set by the user.
-	if nodeResource > maxCapacity {
-		nodeResource = maxCapacity
-	}
+			math.RoundUp(nodeResource, v1alpha1.GiB), quantityRange.Min.Value()), maxCapacity)
```
pkoutsovasilis: this needs some love

pkoutsovasilis (commented):
The changes look safe to me, left a couple minor comments. Let's make the linter happy

hack/go-modernize
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+GO_MODERNIZE="go run golang.org/x/tools/go/analysis/passes/modernize/cmd/modernize@latest -fix -test ./..."
+
+find . -name "go.mod" -type f | while read -r modfile; do
+  dir=$(dirname "${modfile}")
+  echo -e "running for \e[0;35m${dir}\e[0m"
+  (cd "${dir}" && ${GO_MODERNIZE})
+done
```
barkbay: nit: add a "check only" mode for CI and bash lint.
```suggestion
# Exit on error, undefined variables, and pipe failures
set -euo pipefail

# Allow overriding version via environment variable
: "${GO_MODERNIZE_VERSION:=latest}"

# Check mode (--check): report issues without fixing, exit non-zero if changes needed, to be used by CI
# Fix mode (default): automatically apply fixes
if [[ "${1:-}" == "--check" ]]; then
  echo "Running in check mode (no fixes will be applied)"
  GO_MODERNIZE_ARGS=(-test ./...)
else
  GO_MODERNIZE_ARGS=(-fix -test ./...)
fi

# Store command as array to preserve argument (handles spaces, quotes correctly)
GO_MODERNIZE_CMD=(go run "golang.org/x/tools/go/analysis/passes/modernize/cmd/modernize@${GO_MODERNIZE_VERSION}" "${GO_MODERNIZE_ARGS[@]}")

# Track failures to report at the end (allows processing all modules even if some fail)
failed=0

# Use -print0 and read -d '' to safely handle paths with spaces or special characters
# Process substitution < <(...) avoids subshell, ensuring exit codes propagate correctly
while IFS= read -r -d '' modfile; do
  dir=$(dirname "${modfile}")
  echo -e "running for \e[0;35m${dir}\e[0m"
  if ! (cd "${dir}" && "${GO_MODERNIZE_CMD[@]}"); then
    echo "Error: modernize failed for ${dir}" >&2
    failed=1
  fi
done < <(find . -name "go.mod" -type f -print0)

exit "${failed}"
```

barkbay:
Can we not ignore generated files 🤔 
```
go/src/github.com/elastic/cloud-on-k8s/pkg/apis/common/v1/zz_generated.deepcopy.go:34:4: Replace m[k]=v loop with maps.Copy
```

hack/go-modernize
moukoublen: That's very nice! 

Should'n we do like that: `make go-modernize check-local-changes`, like we do in generate and lint?

pkoutsovasilis:
> Can we not ignore generated files 🤔
> 
> ```
> go/src/github.com/elastic/cloud-on-k8s/pkg/apis/common/v1/zz_generated.deepcopy.go:34:4: Replace m[k]=v loop with maps.Copy
> ```

actually I think that this is something we should handle in this PR. The only way I think that we can achieve this is by introducing `modernize` in golangci-lint and not as a Makefile target. Something like the following should do the trick?!

```patch
--- a/.golangci.yml	(revision 5850956a3eed8085d6d7867f844ce14e2713c1a0)
+++ b/.golangci.yml	(date 1769754317098)
@@ -122,6 +122,9 @@
         # revive naming rules are not that happy with our packages/var names in common/utils/shared packages
         path: '.*/(common|utils|shared)/.*.go'
         text: 'var-naming: avoid (package names that conflict with Go standard library package names|meaningless package names)'
+      - linters:
+          - modernize
+        path: '.*zz_generated.*\.go'
     paths:
       - third_party$
       - builtin$
@@ -147,6 +150,7 @@
     - ineffassign
     - makezero
     - misspell
+    - modernize
     - nakedret
     - nestif
     - nilerr
```

moukoublen:
> I think that we can achieve this by introducing `modernize` in golangci-lint and not as a Makefile target.

Yes, I will add it to the Golangci config. I haven't noticed that it is included in the Golangci-lint. 


@pkoutsovasilis @barkbay 
Regarding the autogenerated files, perhaps we need to think it through, I'd suggest.

If the generator generates the files into one form (let's say `A`) and we transform them into a second form `B` then these are some tricky things:

1. What we do in the CI is `make generate check-local-changes`, it will fail. So we will have to do `make generate fmt-modernize check-local-changes` 
2. Everyone that works localy and runs `make generate` during development, will have unstaged changed (because of the restore of `A` form) unrelated to their work, and they will have to run the `fmt-modernize` and it can get messy during development and PRs (at least I've been caught in that loop in the past and it wasn't a nice experience)

pkoutsovasilis:
> * What we do in the CI is `make generate check-local-changes`, it will fail. So we will have to do `make generate fmt-modernize check-local-changes`

In my thinking we should never apply `modernize` to any generated files; my proposal was to completely remove any make target that does modernize (there will be no `fmt-modernize`). We rely solely on golanci-lint to apply modernize skipping generated files with appropriate rules. In the case we get new class of generated files (different path), then the existing CI linter step will be successful but the generate one will fails. This will be an indication that we need to update the skipping rules in golangci-lint. What am I missing?

moukoublen:
> What am I missing?

I think I got it, @barkbay asks if we can include generated files in the modernize ( _**not** ignore_ ), and by your response ( _The only way I think that we can achieve this_ ), I thought you were talking about that.

Now I see that you added an exclusion rule, probably because you thought the comment was about excluding auto-generated files.

So, we don't need any extra rule to do that because auto-generated files are excluded by default. 


We can add the `modernize` flag to golangci-lint and then run `golangci-lint run` or even `golangci-lint run --fix` whenever we want to apply changes. 


And the existing lint ci step will cover that.

pkoutsovasilis:
> I think I got it, @barkbay asks if we can include generated files in the modernize ( not ignore ), and by your response ( The only way I think that we can achieve this ), I thought you were talking about that.

double negation always gets me 🙂 Yes I am actually proposing of not modernizing generated files since these files aren't supposed to be edited (even if the manual edits come from tool). To this end, combining this with golangci-lint will allow us to not add any extra script and make target. @barkbay WDYT?

barkbay:
> Yes I am actually proposing of not modernizing generated files since these files aren't supposed to be edited

Yes, this is what I meant 👍

pkoutsovasilis:
> Yes, this is what I meant 👍

Thanks for the ACK @barkbay, in the .golagci.yml I see the following

```yaml
linters:
  exclusions:
    generated: lax
```

which according to the documentation means 
>     # - `lax`: sources are excluded if they contain lines like `autogenerated file`, `code generated`, `do not edit`, etc.

so I think we are covered

PR #9139, opened 2026-02-13
Make `pkg/apis/common/v1` thinner

moukoublen:
Remove dependencies from `pkg/apis/common/v1` (logs, apm, and controller-runtime). 

This specific package `pkg/apis/common/v1` is widely used in downstream projects. 

Removing some dependencies will help by reducing the number of dependencies for consumers who only use this package (`github.com/elastic/cloud-on-k8s/v3/pkg/apis/common/v1`) from `cloud-on-k8s`.

### DIFF
--- a/pkg/controller/common/name/name.go
+++ b/pkg/controller/common/name/name.go
@@ -73,7 +73,7 @@ func (n Namer) Suffix(ownerName string, suffixes ...string) string {
 	suffixedName, err := n.SafeSuffix(ownerName, suffixes...)
 	// we should never encounter an error at this point because the names should have been validated.
 	if err != nil {
-		fmt.Fprintf(os.Stderr, "Invalid name. This could prevent the operator from functioning correctly. ownerName: %s, suffixedName: %s, error: %s", ownerName, suffixedName, err.Error())
+		fmt.Fprintf(os.Stderr, "Invalid name. This could prevent the operator from functioning correctly. ownerName: %s, suffixedName: %s, error: %s\n", ownerName, suffixedName, err.Error())
 	}
 
 	return suffixedName


prodsecmachine:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/1739f6c7-f99a-4b6e-a0ac-9b5abc69a574?refs=8a653862dba0d24dd41456540177e780335a8e71&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/1739f6c7-f99a-4b6e-a0ac-9b5abc69a574/license?refs=8a653862dba0d24dd41456540177e780335a8e71&source=prChecksComment) |

pkg/controller/common/name/name.go
```diff
@@ -72,11 +69,8 @@ func (n Namer) WithDefaultSuffixes(defaultSuffixes ...string) Namer {
 
 // Suffix generates a resource name by appending the specified suffixes.
 func (n Namer) Suffix(ownerName string, suffixes ...string) string {
-	suffixedName, err := n.SafeSuffix(ownerName, suffixes...)
-	if err != nil {
-		// we should never encounter an error at this point because the names should have been validated
-		log.Error(err, "Invalid name. This could prevent the operator from functioning correctly", "name", suffixedName)
-	}
+	suffixedName, _ := n.SafeSuffix(ownerName, suffixes...)
+	// we should never encounter an error at this point because the names should have been validated
```
barkbay: Assuming this is true and future-proof, [I’m not sure we should hide the error completely](https://en.wikipedia.org/wiki/Error_hiding). Maybe we could log it so we still have visibility? _(edit: sorry I've been a bit fast on this one 😄 ,I just realized this is what the original code was doing, I would be in favor of keeping this)_

pebrc: Agreeing with @barkbay's comment -- silently discarding the error here removes the last safety net for diagnosing issues if the "should never happen" invariant is ever violated. The original code logged it, and I think we should preserve some form of observability.

Two lightweight alternatives that avoid pulling in the heavy `ulog` dependency (which transitively brings in APM + controller-runtime):

**Option A: `fmt.Fprintf(os.Stderr, ...)`**

Zero dependencies, still visible in pod logs, just not structured. Good enough for a "should never happen" path:

```go
suffixedName, err := n.SafeSuffix(ownerName, suffixes...)
if err != nil {
	// we should never encounter an error at this point because the names should have been validated
	fmt.Fprintf(os.Stderr, "ERROR: Invalid name %q, this could prevent the operator from functioning correctly: %v\n", suffixedName, err)
}
```

**Option B: Bridge `log/slog` to the existing logr pipeline**

Since we're on Go 1.25 and logr v1.4.3, we could set up global slog-to-logr bridging in `pkg/utils/log/log.go`'s `setLogger()`:

```go
slog.SetDefault(slog.New(logr.ToSlogHandler(crlog.Log)))
```

Then this code could use `slog.Error(...)` and messages would flow through the existing structured logging pipeline (ECS format, APM tracing, etc.) without importing `ulog` directly. 

I'd lean towards **Option B** for this PR as it is low effort and cleaner.

pkg/controller/common/scheme/scheme.go
```diff
@@ -83,3 +87,5 @@ func SetupV1beta1Scheme() {
 	}
 	mustAddSchemeOnce(&addToSchemeV1beta1, schemes)
 }
```
pebrc, line 91: Nit: `CommonV1` is only referenced in `SetupScheme()` within this same file. Consider making it unexported (`commonV1`) to prevent external callers from coupling to it.

Also, convention in this file (and generally in Go) is to declare package-level vars near the top -- consider moving this into the existing `var (...)` block alongside `addToScheme` and `addToSchemeV1beta1`.

pkg/controller/common/name/name.go
moukoublen: Actually, I am all a slog fanatic to be honest 😅  but in this case I believe we should select option A, and separately check if we can implement a lighter-weight flow of adding the logger (slog with bridge or whatever) to the context and pass it downstream to calls that need log.

moukoublen: Same response with that: https://github.com/elastic/cloud-on-k8s/pull/9139#discussion_r2804493527


The initial reason was to remove the ulog dependency, which brings many other big dependencies (e.g., apm) along with it. So, a custom log for now using fmt might be enough. Later, we can examine a proper "pass logger downstream through context" approach.

pkg/controller/common/scheme/scheme.go
moukoublen: The initial thought was that the previous variables (GroupVersion, SchemeBuilder) were public, so any downstream consumer might potentially access them. In that case, I considered offering a migration path rather than completely removing the scheme Builder.

moukoublen: Move was done in [bcb7b50](https://github.com/elastic/cloud-on-k8s/pull/9139/commits/bcb7b508475da6a4e190ae430423f9e71af7c77a)

pkg/controller/common/name/name.go
moukoublen: Done in [bcb7b50](https://github.com/elastic/cloud-on-k8s/pull/9139/commits/bcb7b508475da6a4e190ae430423f9e71af7c77a)

moukoublen: custom fmt log done in [bcb7b50](https://github.com/elastic/cloud-on-k8s/pull/9139/commits/bcb7b508475da6a4e190ae430423f9e71af7c77a)

barkbay: Can we not use the logger from the controller runtime instead of using `fmt`?

```diff
diff --git a/pkg/controller/common/name/name.go b/pkg/controller/common/name/name.go
index e0968d5de..e5506986e 100644
--- a/pkg/controller/common/name/name.go
+++ b/pkg/controller/common/name/name.go
@@ -11,10 +11,13 @@ import (
 
        "github.com/hashicorp/go-multierror"
        "k8s.io/apimachinery/pkg/util/validation"
+       ctrl "sigs.k8s.io/controller-runtime"
 
        "github.com/elastic/cloud-on-k8s/v3/pkg/utils/stringsutil"
 )
 
+var log = ctrl.Log.WithName("name")
```

moukoublen: The whole purpose of this pr is to avoid using the `"sigs.k8s.io/controller-runtime"` in this package, so the fmt seems good enough for the possibility of this error (which is nearly zero). 

If there is any possibility of this error occurring, the signature should change to return the error.

kvalliyurnatt: sorry I did not understand this part, why do we want to remove controller-runtime dependency ?  I guess to make the package as light as possible ?

moukoublen: Yes, this specific package is widely used in downstream projects. Trying to remove some of the dependencies will help by carrying fewer dependencies for consumers who only use this specific package.

barkbay: >The whole purpose of this pr is to avoid using the "sigs.k8s.io/controller-runtime" in this package, so the fmt seems good enough for the possibility of this error (which is nearly zero).

Can we update the main comment to explain why we are doing this?

pkg/controller/common/name/name.go
```diff
@@ -73,9 +71,9 @@ func (n Namer) WithDefaultSuffixes(defaultSuffixes ...string) Namer {
 // Suffix generates a resource name by appending the specified suffixes.
 func (n Namer) Suffix(ownerName string, suffixes ...string) string {
 	suffixedName, err := n.SafeSuffix(ownerName, suffixes...)
+	// we should never encounter an error at this point because the names should have been validated.
 	if err != nil {
-		// we should never encounter an error at this point because the names should have been validated
-		log.Error(err, "Invalid name. This could prevent the operator from functioning correctly", "name", suffixedName)
+		fmt.Fprintf(os.Stderr, "Invalid name. This could prevent the operator from functioning correctly. ownerName: %s, suffixedName: %s, error: %s", ownerName, suffixedName, err.Error())
 	}
```
pebrc, line 77: Nit: missing trailing `\n` -- without it the next line written to stderr will be concatenated onto this message.

Issue #6461, opened 2023-02-27
Validation webhooks should return warnings


rhr323:
While working on https://github.com/elastic/cloud-on-k8s/issues/6369, we realized that currently [warnings are only emitted as log messages from the reconciliation loop](https://github.com/elastic/cloud-on-k8s/blob/main/pkg/controller/elasticsearch/elasticsearch_controller.go#L268-L276), but those logs can easily be overlooked. [Admission webhooks](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#response) allow returning [warnings](https://github.com/kubernetes/api/blob/master/admission/v1/types.go#L149) with the response, which would make an improved user experience.

naemono:
I think I'm going to combine this with https://github.com/elastic/cloud-on-k8s/issues/8442. Moving to the validator in controller-runtime gives us warnings.

PR #9148, opened 2026-02-18
Simplify zone awareness

naemono:
## Summary

Resolves https://github.com/elastic/cloud-on-k8s/issues/3933

This PR introduces `spec.nodeSets[].zoneAwareness` as first-class API support for Elasticsearch zone awareness in ECK. The goal is to make the common zone-aware deployment path concise and safe by having the operator automatically wire:

- topology spread constraints (`maxSkew=1`, `whenUnsatisfiable=DoNotSchedule`),
- required node affinity (when `zones` are specified),
- downward node labels and `ZONE` env var injection,
- `node.attr.zone` and `cluster.routing.allocation.awareness.attributes` Elasticsearch config defaults.

This replaces the error-prone manual wiring across `config`, `podTemplate`, and the `eck.k8s.elastic.co/downward-node-labels` annotation while preserving user override behavior where appropriate.

### Minimal example

```yaml
spec:
  nodeSets:
  - name: default
    count: 3
    zoneAwareness: {}
```

### With explicit zone restriction

```yaml
spec:
  nodeSets:
  - name: hot
    count: 3
    zoneAwareness:
      zones: ["us-east-1a", "us-east-1b", "us-east-1c"]
```

To customize `maxSkew` or `whenUnsatisfiable`, provide a matching topology spread constraint via `podTemplate.spec.topologySpreadConstraints` with the same `topologyKey`. The operator will preserve the user-provided constraint and not inject its own default for that key.

## Risks

- **One-time rolling restart on upgrade for affected NodeSets.** Enabling `zoneAwareness` on any NodeSet changes the StatefulSet Pod template (topology spread constraints, `ZONE` env var, node affinity rules, `node.attr.zone`/`cluster.routing.allocation.awareness.attributes` config defaults). StatefulSet template changes trigger a controlled rolling restart.
- **Cluster-wide config impact in mixed clusters.** When _any_ NodeSet enables `zoneAwareness`, _all_ NodeSets in the cluster receive zone-related Elasticsearch config (`node.attr.zone: ${ZONE}`, `cluster.routing.allocation.awareness.attributes: k8s_node_name,zone`) and the `ZONE` env var. This ensures shard awareness is consistent across the cluster but means non-zoneAware NodeSets also gain these settings on first adoption.

### API / CRD

- Added `nodeSets[].zoneAwareness` field (`ZoneAwareness` struct) with two optional sub-fields:
  - `topologyKey` (defaults to `topology.kubernetes.io/zone`)
  - `zones` (optional list; `+listType=set`, `MinItems=1` when specified)
- Added helper methods: `TopologyKeyOrDefault()`, `HasZoneAwareness()`, `ZoneAwarenessTopologyKey()`.
- Extended `DownwardNodeLabels()` and `DownwardNodeLabelsHashInput()` to include zone-awareness-derived topology keys alongside the existing annotation-based labels. The hash input preserves the raw annotation value for backward compatibility to avoid spurious rolling restarts on operator upgrade.

### Scheduling

- Zone-aware NodeSets receive:
  - A `TopologySpreadConstraint` with `maxSkew=1`, `whenUnsatisfiable=DoNotSchedule` for the configured topology key. User-provided constraints for the same key are preserved (the operator skips its default).
  - A required node affinity `In` expression when `zones` are specified, restricting placement to the listed zone values.
- Non-zoneAware NodeSets in a cluster where other NodeSets enable zone awareness receive:
  - A required node affinity `Exists` guard on the cluster topology key, ensuring pods land only on nodes that carry the label (prevents init-container deadlock from missing annotations).

### Elasticsearch config

- When `clusterHasZoneAwareness` is true, the operator merges zone awareness defaults (`node.attr.zone: ${ZONE}`, `cluster.routing.allocation.awareness.attributes: k8s_node_name,zone`) into the ES config for all NodeSets, before user config is applied. Explicit user config takes precedence.

### Init container / scripts ConfigMap

- `DownwardNodeLabels()` now includes the zone-awareness topology key. The shared init script waits for all expected annotations (including the zone topology key annotation) before the Elasticsearch container starts. The Downward API volume and operator-side annotation patching provide the values.

### `PodTemplateBuilder` additions

- `WithTopologySpreadConstraints(...)`: appends constraints only when no existing constraint shares the same `topologyKey`. Fills in a missing `LabelSelector` on an existing constraint when the provided constraint has one.
- `WithRequiredNodeAffinityMatchExpressions(...)`: appends required node selector requirements by key without duplicating. For `Exists` operator, skips only when an existing expression on the same key already guarantees label existence (`In`, `Exists`, `Gt`, `Lt`).
- `WithPreferredNodeAffinityMatchExpressions(...)`: appends preferred scheduling terms by key.

### Validation

- **All zone-aware NodeSets must use the same topology key.** The operator generates a single shared init script that waits for all topology key annotations. Mixed keys would cause a pod deadlock (the init script waits for an annotation the operator never sets because the node lacks the other key's label). The `validZoneAwarenessTopologyKeys` validation enforces this.
- **Zone-awareness topology keys are exempt from `--exposed-node-labels` when the flag is empty** (the default). When `--exposed-node-labels` is explicitly configured, zone-awareness topology keys must also be allowed by that policy. This allows zone awareness to work out of the box without requiring flag configuration.

### Sample manifest

- Updated `config/samples/elasticsearch/elasticsearch.yaml` with commented-out `zoneAwareness` example.

### Non-zoneAware NodeSets when others are zoneAware

This is allowed. When at least one NodeSet enables zone awareness and another does not:

- All NodeSets receive zone env/config (`ZONE` env var, `node.attr.zone`, allocation awareness attributes) for cluster-wide consistency.
- Non-zoneAware NodeSets get a required node affinity `Exists` guard on the cluster topology key, ensuring pods only schedule on nodes that carry the label.
- Non-zoneAware NodeSets do **not** get topology spread constraints automatically.

### Mixed topology keys across NodeSets

**Not allowed.** All zone-aware NodeSets must share the same topology key. This is enforced by the validating webhook. The restriction exists because the operator uses a single shared init script and annotation set for the entire cluster.

### User overrides

- Existing user-provided topology spread constraints for the same key are preserved; the operator does not inject a duplicate.
- Existing required node affinity expressions for the same key are not overwritten.
- User-provided `ZONE` env var in `podTemplate` is preserved.
- Explicit user Elasticsearch config (`node.attr.zone`, `cluster.routing.allocation.awareness.attributes`) takes precedence over operator defaults.
- Legacy manual zone-awareness configurations (annotation + manual config/podTemplate) remain supported without modification.

## Backward Compatibility

- Existing manual zone-aware specs (annotation + manual config/podTemplate) are supported without modification.
- Migration to `zoneAwareness` is incremental and optional.
- Enabling `zoneAwareness` on any NodeSet triggers a one-time rolling restart of all NodeSets in the cluster due to the addition of zone-related config and env vars.
- The `DownwardNodeLabelsHashInput()` method preserves the raw annotation value for hash computation, avoiding spurious rolling restarts for existing clusters that already use the `eck.k8s.elastic.co/downward-node-labels` annotation without `zoneAwareness`.

## AI Usage

Cursor/gpt-5.3 codex used during:
- Planning/Spec creation
- Code writing
- Test plan writing/execution
- PR Description generation
- PR review process (ran this through tons of review using differing models to catch any edge cases missed)

Significant manual cleanup was required after AI-generated code and tests.

### DIFF
--- a/pkg/controller/elasticsearch/settings/merged_config.go
+++ b/pkg/controller/elasticsearch/settings/merged_config.go
@@ -20,16 +20,19 @@ import (
 	netutil "github.com/elastic/cloud-on-k8s/v3/pkg/utils/net"
 )
 
-// the name of the ES attribute indicating the pod's current k8s node
-const nodeAttrK8sNodeName = "k8s_node_name"
+const (
+	// the name of the ES attribute indicating the pod's current k8s node
+	nodeAttrK8sNodeName = "k8s_node_name"
+	// the name of the ES attribute indicating the pod's current zone
+	nodeAttrZone = "zone"
+	// the format of the environment variable reference
+	envVarReferenceFormat = "${%s}"
+)
 
-var nodeAttrNodeName = fmt.Sprintf("%s.%s", esv1.NodeAttr, nodeAttrK8sNodeName)
-
-const nodeAttrZone = "zone"
-
-var nodeAttrZoneName = fmt.Sprintf("%s.%s", esv1.NodeAttr, nodeAttrZone)
-
-const envVarReferenceFormat = "${%s}"
+var (
+	nodeAttrNodeName = fmt.Sprintf("%s.%s", esv1.NodeAttr, nodeAttrK8sNodeName)
+	nodeAttrZoneName = fmt.Sprintf("%s.%s", esv1.NodeAttr, nodeAttrZone)
+)
 
 // NewMergedESConfig merges user provided Elasticsearch configuration with configuration derived from the given
 // parameters. The user provided config overrides have precedence over the ECK config.


prodsecmachine:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/856f4076-2d93-47ae-b1eb-af52b90fa7e4?refs=e00d658686e6b3fe74029eebbddfdaa0de733809&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/856f4076-2d93-47ae-b1eb-af52b90fa7e4/license?refs=e00d658686e6b3fe74029eebbddfdaa0de733809&source=prChecksComment) |

pkg/controller/elasticsearch/nodespec/podspec.go
```diff
@@ -236,6 +242,199 @@ func buildAnnotations(
...
+		// when a nodeSet has no configured zone awareness but another nodeSet has zone awareness.
+		mergeRequiredNodeAffinityForTopologyExists(podSpec, clusterTopologyKey)
+		return
+	}
+	if nodeSet.ZoneAwareness == nil {
+		return
+	}
+
+	topologyKey := nodeSet.ZoneAwareness.TopologyKeyOrDefault()
+	if !hasTopologySpreadConstraintForKey(podSpec.TopologySpreadConstraints, topologyKey) {
```
kvalliyurnatt: IIUC this means if someone added both the zone awareness config and a topology spread constraint in the pod spec template section. the one in pod spec template section takes precedence ?

pkg/controller/elasticsearch/nodespec/podspec.go
```diff
@@ -236,6 +242,199 @@ func buildAnnotations(
...
+	// that the nodeSet is only scheduled on nodes that have the zone values.
+	if len(nodeSet.ZoneAwareness.Zones) > 0 {
+		mergeRequiredNodeAffinityForTopology(podSpec, topologyKey, nodeSet.ZoneAwareness.Zones)
+	}
+}
+
+// hasTopologySpreadConstraintForKey returns true when a spread constraint already exists
+// for the provided topology key.
+func hasTopologySpreadConstraintForKey(constraints []corev1.TopologySpreadConstraint, topologyKey string) bool {
+	for _, constraint := range constraints {
```
kvalliyurnatt: nit: I think we can use the slices.Contains method instead of a separate function

pkg/controller/elasticsearch/nodespec/podspec.go
kvalliyurnatt: Mentioned already in description that this is the expected behavior

pkg/controller/elasticsearch/nodespec/podspec.go
naemono: `slices.ContainsFunc` will do the job here. Good call. ty

pkg/apis/elasticsearch/v1/elasticsearch_types.go
```diff
@@ -454,12 +498,28 @@ type Elasticsearch struct {
...
+			continue
+		}
+		expectedAnnotations.Add(nodeSet.ZoneAwareness.TopologyKeyOrDefault())
+	}
+
+	if expectedAnnotations.Count() == 0 {
 		return nil
 	}
-	return strings.Split(expectedAnnotations, ",")
+	return expectedAnnotations.AsSortedSlice()
```
pebrc: The refactored DownwardNodeLabels() now returns a sorted, deduplicated slice (via set.Make().AsSortedSlice()), and the config hash computation switched from hashing the raw annotation string to hashing the joined sorted result:

For any existing deployment with multi-value eck.k8s.elastic.co/downward-node-labels in non-alphabetical order (e.g. topology.kubernetes.io/zone,topology.kubernetes.io/region), the config hash changes, triggering a spurious rolling restart even if zone awareness is not used. The test at podspec_test.go line 384 confirms this: the expected hash changed from "909392898" to "3356455586".

The PR's risk section mentions restarts for zone-awareness-affected NodeSets, but this additional restart vector for existing non-zone-aware deployments using multi-value downward-node-labels isn't called out. 
We should investigate if it is feasible to avoid the restart for users that do not use the feature.

pkg/controller/elasticsearch/nodespec/podspec.go
```diff
@@ -236,6 +243,196 @@ func buildAnnotations(
...
+	return esv1.DefaultZoneAwarenessTopologyKey
+}
+
+// topologyKeyOrDefault returns the provided key or the default zone topology key
+// when the provided value is empty.
+func topologyKeyOrDefault(topologyKey string) string {
+	if topologyKey == "" {
+		return esv1.DefaultZoneAwarenessTopologyKey
+	}
+	return topologyKey
```
pebrc: **Duplicate default logic**: This `topologyKeyOrDefault` checks `topologyKey == ""`, while `ZoneAwareness.TopologyKeyOrDefault()` in `elasticsearch_types.go:393` checks `strings.TrimSpace(za.TopologyKey) == ""`. The two behave differently for whitespace-only input (e.g. `" "`), which the CRD's `MinLength=1` technically permits.

Consider either:
- Delegating to a single shared function, or
- Having this function also trim, or
- Adding a CRD validation pattern (e.g. `^\S+$`) to disallow whitespace-only values

This is a minor maintenance hazard but worth aligning.

pkg/controller/elasticsearch/nodespec/podspec.go
```diff
@@ -236,6 +243,196 @@ func buildAnnotations(
...
+			},
+		)
+	}
+
+	// If the nodeSet has zone awareness and zones are configured, we want to ensure
+	// that the nodeSet is only scheduled on nodes that have the zone values.
+	if len(nodeSet.ZoneAwareness.Zones) > 0 {
+		mergeRequiredNodeAffinityForTopology(podSpec, topologyKey, nodeSet.ZoneAwareness.Zones)
+	}
+}
```
pebrc: **`PodTemplateBuilder` usage**: This function reaches into `builder.PodTemplate.Spec` directly to manipulate `TopologySpreadConstraints` and `Affinity.NodeAffinity`. While this works correctly and respects user overrides (via the `has*` guard functions), it bypasses the builder's `With*` pattern that the rest of the codebase uses idiomatically (e.g. `WithContainers`, `WithEnv`, `WithAnnotations`, `WithVolumes`).

The builder's contract is "apply defaults without overriding user-provided values", and each `With*` method encapsulates that logic. By operating directly on the spec, this function takes on responsibility for the "don't override" semantics itself (which it does, but in an ad-hoc way).

This is justified today because the builder lacks `WithTopologySpreadConstraints` and `WithRequiredNodeAffinity` methods. Worth considering adding those to `PodTemplateBuilder` to keep the pattern consistent, especially if more controllers need similar scheduling defaults in the future.

cmd/manager/main.go
```diff
@@ -252,8 +252,8 @@ func Command() *cobra.Command {
 	)
 	cmd.Flags().StringSlice(
 		operator.ExposedNodeLabels,
-		[]string{},
-		"Comma separated list of node labels which are allowed to be copied as annotations on Elasticsearch Pods, empty by default",
+		[]string{"topology.kubernetes.io/.*", "failure-domain.beta.kubernetes.io/.*"},
```
pebrc: **Changing the CLI default**: Previously this defaulted to `[]string{}` (empty), meaning only Helm users (whose `values.yaml` already set these patterns) had topology labels exposed. Now all operator deployments — including those installed via `make deploy`, static manifests, or OLM — get `topology.kubernetes.io/.*` and `failure-domain.beta.kubernetes.io/.*` exposed by default.

This is a good alignment and is needed for zone awareness to work out of the box. However, it's a behavior change for non-Helm users who previously had no exposed node labels. The upgrade notes in the PR description mention this implicitly but might benefit from calling it out explicitly as a default change.

pkg/controller/elasticsearch/settings/merged_config.go
```diff
@@ -53,6 +62,20 @@ func NewMergedESConfig(
...
+		cfg := NewCanonicalConfig()
+		return &cfg
+	}
+	cfg := map[string]any{}
+	zoneEnvVarRef := fmt.Sprintf(envVarReferenceFormat, EnvZone)
+	cfg[nodeAttrZoneName] = zoneEnvVarRef
+	if clusterHasZoneAwareness {
+		cfg[esv1.ShardAwarenessAttributes] = fmt.Sprintf("%s,%s", nodeAttrK8sNodeName, nodeAttrZone)
+	}
+	return &CanonicalConfig{common.MustCanonicalConfig(cfg)}
```
pebrc: **`node.attr.zone` set for non-zone-aware NodeSets**: When `clusterHasZoneAwareness` is true but `zoneAwareness` is nil for a specific NodeSet, this function still sets `node.attr.zone = ${ZONE}`. The ZONE env var is also injected (via `zoneAwarenessEnv`) for such NodeSets.

This means a non-zone-aware NodeSet in a cluster where another NodeSet has zone awareness will get `node.attr.zone` referencing a downward API annotation. If the node label is properly propagated, this works. But if the topology key for the zone-aware NodeSet differs from the default and there's no per-NodeSet zone awareness on this one, the attribute might reference the wrong annotation key.

The validation at `validZoneAwarenessTopologyKeys` guards against the ambiguous multi-key case, but it might be worth a comment here explaining that the default topology key is intentionally used as the fallback for non-zone-aware NodeSets in a zone-aware cluster.

pkg/controller/elasticsearch/validation/validations.go
```diff
@@ -98,6 +101,46 @@ func validNodeLabels(proposed esv1.Elasticsearch, exposedNodeLabels NodeLabels)
...
 }
 
+// validZoneAwarenessTopologyKeys rejects only this configuration:
+//
+//	multiple differing topology keys across zone-aware NodeSets
+//	and at least one NodeSet fails to have zone awareness configured.
+//
+// In that case, the non-zoneAware NodeSet needs a fallback topology key, but with
+// multiple candidate keys there is no clear choice.
+func validZoneAwarenessTopologyKeys(es esv1.Elasticsearch) field.ErrorList {
```
pebrc: **Good validation**: This is a well-thought-out validation that catches the genuinely ambiguous case (multiple differing topology keys with partial zone awareness). The choice to be permissive (allow identical keys, allow single-key mixed) while rejecting only the truly ambiguous configuration is the right trade-off for usability.

One edge case to consider: if a user explicitly sets `topologyKey: "topology.kubernetes.io/zone"` on one NodeSet and uses the default (empty) on another, they happen to resolve to the same key, but `TopologyKeyOrDefault()` would be needed here to normalize before comparison. Looking at the code, this is already handled correctly since the comparison uses `TopologyKeyOrDefault()`. 👍

pkg/controller/elasticsearch/nodespec/podspec.go
naemono: I actually went back and forth with this one. AI initially did it similar to this style, and I tried forcing it into the `With` pattern, only to see that some domain-specific knowledge of Elasticsearch was getting leaked into the `common/defaults` package. 

Let me see if I can extract TSC and NA in the ES package, and add them to the podspec with methods in the defaults package...

cmd/manager/main.go
naemono: It seemed a correct change to align with the Helm chart. Do you feel we should call this out outside of possibly the release notes/breaking changes?

pkg/controller/elasticsearch/nodespec/podspec.go
naemono: I've updated this to better match this style.

pkg/controller/elasticsearch/nodespec/podspec.go
naemono: Fixed matching TrimSpace

pkg/apis/elasticsearch/v1/elasticsearch_types.go
naemono: This was a good find. I've fixed this as well.

pkg/apis/elasticsearch/v1/elasticsearch_types.go
```diff
@@ -353,6 +357,46 @@ type NodeSet struct {
...
+	MaxSkew *int32 `json:"maxSkew,omitempty"`
+
+	// WhenUnsatisfiable configures how to handle unsatisfied spread constraints.
+	// Defaults to DoNotSchedule when not set.
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Enum=DoNotSchedule;ScheduleAnyway
+	WhenUnsatisfiable *corev1.UnsatisfiableConstraintAction `json:"whenUnsatisfiable,omitempty"`
+}
+
+func (za ZoneAwareness) TopologyKeyOrDefault() string {
```
barkbay, line 384: There's an inconsistency in how default values are accessed for `ZoneAwareness` fields:

| Field | Current approach | Location |
|-------|------------------|----------|
| `TopologyKey` | Receiver method `za.TopologyKeyOrDefault()` | `elasticsearch_types.go` |
| `MaxSkew` | Standalone function `zoneAwarenessMaxSkewOrDefault(za)` | `podspec.go` |
| `WhenUnsatisfiable` | Standalone function `zoneAwarenessWhenUnsatisfiableOrDefault(za)` | `podspec.go` |

### Recommendation

Move all default-value logic to receiver methods on `ZoneAwareness` for consistency:

```go
func (za ZoneAwareness) TopologyKeyOrDefault() string { ... }
func (za ZoneAwareness) MaxSkewOrDefault() int32 { ... }
func (za ZoneAwareness) WhenUnsatisfiableOrDefault() corev1.UnsatisfiableConstraintAction { ... }
```

**Benefits:**
- More idiomatic Go — defaults are behavior intrinsic to the type
- Better discoverability via autocomplete
- Cleaner call sites: `nodeSet.ZoneAwareness.MaxSkewOrDefault()` vs `zoneAwarenessMaxSkewOrDefault(nodeSet.ZoneAwareness)`
- Consistent pattern already established by `TopologyKeyOrDefault()`

### Optional: Extract to dedicated file

Consider moving all zone awareness code to `pkg/apis/elasticsearch/v1/zone_awareness.go`:

```go
// zone_awareness.go
package v1

import (
    corev1 "k8s.io/api/core/v1"
    "strings"
)

const (
    DefaultZoneAwarenessTopologyKey = "topology.kubernetes.io/zone"
)

type ZoneAwareness struct {
    TopologyKey       string                                `json:"topologyKey,omitempty"`
    Zones             []string                              `json:"zones,omitempty"`
    MaxSkew           *int32                                `json:"maxSkew,omitempty"`
    WhenUnsatisfiable *corev1.UnsatisfiableConstraintAction `json:"whenUnsatisfiable,omitempty"`
}

func (za ZoneAwareness) TopologyKeyOrDefault() string {
    if strings.TrimSpace(za.TopologyKey) == "" {
        return DefaultZoneAwarenessTopologyKey
    }
    return za.TopologyKey
}

func (za ZoneAwareness) MaxSkewOrDefault() int32 {
    if za.MaxSkew != nil {
        return *za.MaxSkew
    }
    return 1
}

func (za ZoneAwareness) WhenUnsatisfiableOrDefault() corev1.UnsatisfiableConstraintAction {
    if za.WhenUnsatisfiable != nil {
        return *za.WhenUnsatisfiable
    }
    return corev1.DoNotSchedule
}
```

This keeps `elasticsearch_types.go` focused and groups all zone awareness logic together.

pkg/apis/elasticsearch/v1/elasticsearch_types.go
```diff
@@ -353,6 +357,46 @@ type NodeSet struct {
 	VolumeClaimTemplates []corev1.PersistentVolumeClaim `json:"volumeClaimTemplates,omitempty"`
 }
 
+const (
+	// DefaultZoneAwarenessTopologyKey is used when topologyKey is not specified.
+	DefaultZoneAwarenessTopologyKey = "topology.kubernetes.io/zone"
```
pebrc: Why not use the existing const from  "k8s.io/api/core/v1"

pkg/controller/elasticsearch/nodespec/podspec.go
```diff
@@ -236,6 +255,158 @@ func buildAnnotations(
...
+		derived = append(derived, label)
+	}
+	if len(derived) == 0 {
+		return hashInput
+	}
+	if hashInput != "" {
+		hashInput += ","
+	}
+	return hashInput + strings.Join(derived, ",")
+}
```
pebrc: `downwardNodeLabelsHashInput` re-parses the `DownwardNodeLabelsAnnotation` (split by comma, trim whitespace) — the same logic that `DownwardNodeLabels()` in `elasticsearch_types.go` already implements. If the annotation format ever changes, both places would need updating.

Consider either:
- Moving this to a method on `Elasticsearch` (e.g. `DownwardNodeLabelsHashInput()`) next to `DownwardNodeLabels()` so the parsing logic stays in one place, or
- Refactoring so this function delegates to `DownwardNodeLabels()` for the dedup check and reads the raw annotation through a single accessor, avoiding the re-parsing.

pkg/controller/elasticsearch/nodespec/podspec.go
```diff
@@ -141,6 +145,20 @@ func BuildPodTemplateSpec(
...
+		ssetName,
+		clusterHasZoneAwareness,
+		clusterZoneAwarenessTopologyKey,
+	)
+	if spreadConstraint != nil {
+		builder.WithTopologySpreadConstraints(*spreadConstraint)
+	}
+	if requiredMatchExpression != nil {
+		builder.WithRequiredNodeAffinityMatchExpressions(*requiredMatchExpression)
+	}
```
pebrc: The nil checks here can be eliminated by having `zoneAwarenessSchedulingDirectives` return slices instead of pointers:

```go
func zoneAwarenessSchedulingDirectives(...) ([]corev1.TopologySpreadConstraint, []corev1.NodeSelectorRequirement)
```

Then the caller becomes:

```go
constraints, requirements := zoneAwarenessSchedulingDirectives(...)
builder.
    WithTopologySpreadConstraints(constraints...).
    WithRequiredNodeAffinityMatchExpressions(requirements...)
```

Both `With*` methods already handle empty input gracefully (`WithRequiredNodeAffinityMatchExpressions` returns early on `len(requirements) == 0`, `WithTopologySpreadConstraints` iterates over nothing). This also lets the calls flow naturally with the builder chain above instead of being a separate block.

pkg/controller/elasticsearch/nodespec/podspec.go
```diff
@@ -236,6 +255,158 @@ func buildAnnotations(
...
+}
+
+// topologyKeyOrDefault returns the provided key or the default zone topology key
+// when the provided value is empty.
+func topologyKeyOrDefault(topologyKey string) string {
+	if strings.TrimSpace(topologyKey) == "" {
+		return esv1.DefaultZoneAwarenessTopologyKey
+	}
+	return topologyKey
+}
```
pebrc: This duplicates `ZoneAwareness.TopologyKeyOrDefault()` in `elasticsearch_types.go:393` (same trim, same default, same constant).

Moreover, the callers of this function (`zoneAwarenessEnv` and `zoneAwarenessSchedulingDirectives`) only ever pass `clusterTopologyKey`, which is already the result of `zoneAwarenessTopologyKey()` at line 299 — which itself calls `TopologyKeyOrDefault()`. So by the time the value reaches here, it's already been defaulted and can never be empty. This function is effectively a no-op in practice.

Consider removing it and using the already-defaulted `clusterTopologyKey` directly in the callers.

pkg/controller/elasticsearch/nodespec/podspec.go
```diff
@@ -236,6 +255,158 @@ func buildAnnotations(
...
+// without zone awareness when cluster-wide awareness is enabled.
+func zoneAwarenessTopologyKey(nodeSets []esv1.NodeSet) string {
+	for _, nodeSet := range nodeSets {
+		if nodeSet.ZoneAwareness == nil {
+			continue
+		}
+		return nodeSet.ZoneAwareness.TopologyKeyOrDefault()
+	}
+	return esv1.DefaultZoneAwarenessTopologyKey
+}
```
pebrc: **Nit: misleading comment.** "cluster-wide awareness" suggests a global cluster-level setting, but `zoneAwareness` is a per-NodeSet field. This function returns the topology key from the first zone-aware NodeSet, to be used as a fallback for non-zone-aware NodeSets.

Suggested rewording:

```go
// zoneAwarenessTopologyKey returns the topology key from the first zone-aware NodeSet,
// used as a fallback for non-zone-aware NodeSets in the same cluster.
```

pkg/apis/elasticsearch/v1/elasticsearch_types.go
```diff
@@ -353,6 +357,46 @@ type NodeSet struct {
...
+	// MaxSkew controls how unevenly Pods may be distributed across topology domains.
+	// Defaults to 1 when not set.
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Minimum=1
+	MaxSkew *int32 `json:"maxSkew,omitempty"`
+
+	// WhenUnsatisfiable configures how to handle unsatisfied spread constraints.
+	// Defaults to DoNotSchedule when not set.
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Enum=DoNotSchedule;ScheduleAnyway
```
barkbay: The `ZoneAwareness` struct re-defines a subset of `TopologySpreadConstraint` fields with explicit kubebuilder validations:

```go
// +kubebuilder:validation:Enum=DoNotSchedule;ScheduleAnyway
WhenUnsatisfiable *corev1.UnsatisfiableConstraintAction `json:"whenUnsatisfiable,omitempty"`
```

This creates two potential maintenance issues:

1. **Enum constraints block future values** — If Kubernetes adds a new value (e.g., `ScheduleWithPenalty`), our CRD validation would reject it at admission time, even though the underlying K8s API would accept it. Users on newer K8s versions couldn't use valid new options until we update ECK.

2. **Missing fields require ECK updates** — Fields like `minDomains`, `nodeAffinityPolicy`, `nodeTaintsPolicy`, and `matchLabelKeys` were added in K8s 1.25-1.27. If a user wants to use these (or future fields), they'd need to fall back to `podTemplate` and lose the benefits of `zoneAwareness`.

### Proposal

Consider embedding `corev1.TopologySpreadConstraint` directly:

```go
type ZoneAwareness struct {
    // Embed the K8s type - all fields available, forward compatible
    corev1.TopologySpreadConstraint `json:",inline"`
    
    // Zone-awareness specific (maps to NodeAffinity, not TSC)
    Zones []string `json:"zones,omitempty"`
}
```

The controller would auto-fill `LabelSelector` if the user doesn't provide one (the main value-add of this feature), while giving users access to the full TSC API:

```yaml
# Simple case - labelSelector auto-generated
zoneAwareness:
  topologyKey: topology.kubernetes.io/zone

# Advanced case - minDomains works without ECK changes
zoneAwareness:
  topologyKey: topology.kubernetes.io/zone
  minDomains: 2
```

This gives us forward compatibility with K8s API evolution while preserving the simplified UX for the common case.

pkg/controller/elasticsearch/nodespec/podspec.go
```diff
@@ -236,6 +255,158 @@ func buildAnnotations(
...
+
+	if nodeSet.ZoneAwareness == nil && clusterHasZoneAwareness {
+		clusterTopologyKey = topologyKeyOrDefault(clusterTopologyKey)
+		// We want to ensure that the nodeSet is only scheduled on nodes that have the cluster topology key
+		// when a nodeSet has no configured zone awareness but another nodeSet has zone awareness.
+		requiredMatchExpression = &corev1.NodeSelectorRequirement{
+			Key:      clusterTopologyKey,
+			Operator: corev1.NodeSelectorOpExists,
+		}
+		return spreadConstraint, requiredMatchExpression
```
pebrc: **Implicit `Exists` affinity guard on non-zone-aware NodeSets — is this the right policy?**

This silently adds a *hard* scheduling constraint (`requiredDuringScheduling`) to NodeSets that never opted into zone awareness. The scenario: a user adds `zoneAwareness: {}` to their data NodeSet, and their master NodeSet — running on dedicated nodes that deliberately don't carry `topology.kubernetes.io/zone` labels — suddenly can't schedule. The only signal is a `FailedScheduling` event on the pod.

The rationale is sound: `node.attr.zone` and shard awareness are injected cluster-wide, so if the annotation is missing, `node.attr.zone` resolves to an empty string, causing shard allocation issues. The guard prevents that inconsistency.

But the user experience concern is real. Two suggestions:

1. **Emit a warning event** on the Elasticsearch resource when this implicit guard is applied, so the user understands why their non-zone-aware NodeSet can't schedule.
2. **Consider whether `preferredDuringScheduling` (soft preference) would be more conservative** for the non-zone-aware case, avoiding breakage at the cost of potentially running with an empty zone attribute on some nodes.

pkg/controller/elasticsearch/validation/validations.go
```diff
@@ -64,6 +66,7 @@ func validations(ctx context.Context, checker license.Checker, exposedNodeLabels
 		func(proposed esv1.Elasticsearch) field.ErrorList {
 			return validNodeLabels(proposed, exposedNodeLabels)
```
pebrc, line 69: **Validation error field path is misleading for zone-awareness-sourced labels.**

`validNodeLabels` iterates over `DownwardNodeLabels()`, which now returns labels from both the annotation *and* `zoneAwareness.topologyKey`. But the error path always points to `metadata.annotations/eck.k8s.elastic.co/downward-node-labels`.

A user who sets `zoneAwareness: { topologyKey: "custom.io/rack" }` on an operator without `custom.io/rack` in `--exposed-node-labels` will see an error pointing at an annotation that isn't even present in their resource.

The validation should distinguish between annotation-sourced and zone-awareness-sourced labels, and point to `spec.nodeSets[i].zoneAwareness.topologyKey` in the latter case.

pkg/controller/elasticsearch/settings/merged_config.go
pebrc: **The `zoneAwareness` parameter is logically redundant.**

It is never read past the early-return guard on line 66. And that guard itself is redundant: when `!clusterHasZoneAwareness`, no NodeSet has `zoneAwareness` set, so `zoneAwareness == nil` is always true in that case. The two conditions collapse to just `!clusterHasZoneAwareness`.

The function could simply take `clusterHasZoneAwareness bool` alone and the behavior would be identical. As written, the unused parameter is misleading — it suggests per-NodeSet config differentiation that doesn't actually happen.

Either remove the parameter (simplification), or use it to produce per-NodeSet config differences (e.g., different awareness attributes per NodeSet) if that's intended for the future.

cmd/manager/main.go
pebrc: This needs be called out at least as breaking change.

pkg/controller/elasticsearch/nodespec/podspec.go
```diff
@@ -236,6 +255,158 @@ func buildAnnotations(
...
+		trimmed := strings.TrimSpace(label)
+		if trimmed == "" {
+			continue
+		}
+		annotationLabels[trimmed] = struct{}{}
+	}
+
+	var derived []string
+	for _, label := range es.DownwardNodeLabels() {
+		if _, exists := annotationLabels[label]; exists {
```
barkbay:
```suggestion
	annotationLabels := set.Make()
	for label := range strings.SplitSeq(annotationValue, ",") {
		trimmed := strings.TrimSpace(label)
		if trimmed == "" {
			continue
		}
		annotationLabels.Add(trimmed)
	}

	var derived []string
	for _, label := range es.DownwardNodeLabels() {
		if annotationLabels.Has(label) {
```

pkg/controller/elasticsearch/nodespec/podspec_test.go
```diff
@@ -463,6 +464,67 @@ func Test_buildAnnotations(t *testing.T) {
...
+							Name: "with-za",
+							ZoneAwareness: &esv1.ZoneAwareness{
+								TopologyKey: "topology.custom.io/rack",
+							},
+						},
+					},
+				},
+			},
+			want: "topology.custom.io/rack",
+		},
```
barkbay, line 556:
```suggestion
		},
		{
			name: "zone awareness key matching annotation does not change hash",
			es: esv1.Elasticsearch{
				ObjectMeta: metav1.ObjectMeta{
					Annotations: map[string]string{
						esv1.DownwardNodeLabelsAnnotation: "topology.kubernetes.io/zone",
					},
				},
				Spec: esv1.ElasticsearchSpec{
					NodeSets: []esv1.NodeSet{
						{
							Name:          "default",
							ZoneAwareness: &esv1.ZoneAwareness{},
						},
					},
				},
			},
			want: "topology.kubernetes.io/zone",
		},
```

pkg/controller/elasticsearch/nodespec/resources.go
```diff
@@ -106,6 +118,19 @@ func BuildExpectedResources(
...
+// This allows us to ensure consistency and that each nodeSet has the
+// allocation awareness setting in place.
+// (cluster.routing.allocation.awareness.attributes)
+func hasZoneAwareness(nodeSets []esv1.NodeSet) bool {
+	for _, nodeSet := range nodeSets {
+		if nodeSet.ZoneAwareness != nil {
+			return true
+		}
+	}
+	return false
```
pebrc: Consider making this a receiver method on `NodeSetList` in `elasticsearch_types.go`, where the type and its `Names()` method already live:

```go
func (nsl NodeSetList) HasZoneAwareness() bool {
    for _, ns := range nsl {
        if ns.ZoneAwareness != nil {
            return true
        }
    }
    return false
}
```

Callers would use `es.Spec.NodeSets.HasZoneAwareness()` instead of `hasZoneAwareness(es.Spec.NodeSets)`. The same argument applies to `zoneAwarenessTopologyKey` in `podspec.go` — it could be `NodeSetList.ZoneAwarenessTopologyKey()`.

This keeps zone-awareness queries on the type that owns the data rather than scattered across controller packages.

pkg/controller/elasticsearch/nodespec/podspec.go
naemono: It feels as though `preferredDuringScheduling` is the right path here, and that's what I've implemented.

pkg/controller/elasticsearch/validation/validations.go
naemono: This logic has been updated, and `ParseDownwardNodeLabels` exported such that it can be used in this package.

pkg/controller/elasticsearch/nodespec/podspec.go
naemono: Resolved when `DownwardNodeLabelsHashInput` was exported.

pkg/apis/elasticsearch/v1/elasticsearch_types.go
naemono: So this is actually causing issues which I found during some testing:

```
❯ kc apply -n elastic -f ~/tmp/test-zone-a.yaml
The Elasticsearch "testing-default-az-keys" is invalid:
* spec.nodeSets[0].zoneAwareness.maxSkew: Required value
* spec.nodeSets[0].zoneAwareness.topologyKey: Required value
* spec.nodeSets[0].zoneAwareness.whenUnsatisfiable: Required value

✗ cat ~/tmp/test-zone-a.yaml
kind: Elasticsearch
metadata:
  name: testing-default-az-keys
spec:
  version: 9.3.0
  nodeSets:
  - name: default
    count: 1
    zoneAwareness: {}
```

So it seem that `corev1.TopologySpreadConstraint` has `maxSkew`, `topologyKey`, and `whenUnsatisfiable` tagged as `required` in its upstream OpenAPI schema. I'm not sure there's a way to avoid in-lining these fields. Suggestions welcome here.

barkbay: Indeed, good point. If I try to sum up the tradeoffs:

### Option A: Custom Fields

```go
type ZoneAwareness struct {
    TopologyKey       string                                `json:"topologyKey,omitempty"`
    MaxSkew           *int32                                `json:"maxSkew,omitempty"`
    WhenUnsatisfiable *corev1.UnsatisfiableConstraintAction `json:"whenUnsatisfiable,omitempty"`
    Zones             []string                              `json:"zones,omitempty"`
}
```

| Pros | Cons |
|------|------|
| `zoneAwareness: {}` works with defaults | We must manually track upstream API evolution |
| Minimal YAML for users | `+kubebuilder:validation:Enum` blocks future values (e.g., `ScheduleWithPenalty`) - but maybe we should remove it anyway?|
| Full control over validation (optional or not) | Missing advanced fields (`minDomains`, `matchLabelKeys`, `nodeAffinityPolicy`, etc.) |

### Option B: Embed `corev1.TopologySpreadConstraint`

```go
type ZoneAwareness struct {
    corev1.TopologySpreadConstraint `json:",inline"`
    Zones []string `json:"zones,omitempty"`
}
```

| Pros | Cons |
|------|------|
| All upstream fields available automatically | `maxSkew`, `topologyKey`, `whenUnsatisfiable` become **required** |
| Forward-compatible with Kubernetes evolution | Cannot use `zoneAwareness: {}` (not sure if it is a concern?) |
| No enum validation blocking future values |  |
| Advanced features like `minDomains` available | |

```yaml
zoneAwareness:
  topologyKey: topology.kubernetes.io/zone
  maxSkew: 1
  whenUnsatisfiable: DoNotSchedule
```

I would have a slight preference for embedding `corev1.TopologySpreadConstraint`, feels like the better long-term choice at a first glance. The UX inconvenience of 3 extra lines is outweighed by the maintenance and compatibility benefits. But I would be curious to get others' opinions on this.

_There are also the options of patching the generated CRD to remove required constraints or using the webhook to inject default values, both feel a bit hacky/fragile._

cmd/manager/main.go
pebrc: Actually I think we should try to avoid changing the default. I am not sure it is necessary.

pkg/apis/elasticsearch/v1/elasticsearch_types.go
```diff
@@ -353,6 +357,46 @@ type NodeSet struct {
...
 }
 
+const (
+	// DefaultZoneAwarenessTopologyKey is used when topologyKey is not specified.
+	DefaultZoneAwarenessTopologyKey = corev1.LabelTopologyZone
+)
+
+// ZoneAwareness configures topology-aware scheduling and shard-awareness defaults for a NodeSet.
+type ZoneAwareness struct {
+	corev1.TopologySpreadConstraint `json:",inline"`
```
pebrc: **Critical: CRD `required` fields from `TopologySpreadConstraint` embedding**

Because `corev1.TopologySpreadConstraint` is embedded inline, the generated CRD schema inherits its `required` markers — making `maxSkew`, `topologyKey`, and `whenUnsatisfiable` **required** in the `zoneAwareness` object (see lines 9846-9849 of the generated CRD YAML).

Consequences:
1. `zoneAwareness: {}` is **rejected** by the Kubernetes API server — contradicting the PR description which shows this as a valid use case
2. Users **must** always specify all three fields, making `MaxSkewOrDefault()`, `TopologyKeyOrDefault()`, and `WhenUnsatisfiableOrDefault()` effectively unreachable for CRD-validated resources
3. This defeats the stated goal of making zone awareness "concise and safe"
4. **API surface bloat**: `labelSelector`, `matchLabelKeys`, `minDomains`, `nodeAffinityPolicy`, `nodeTaintsPolicy` are exposed to users but should not be user-configurable — the operator auto-generates `labelSelector` for topology spread constraints

**Recommendation**: Don't embed `TopologySpreadConstraint` inline. Declare only the relevant fields directly in `ZoneAwareness`:

```go
type ZoneAwareness struct {
    // TopologyKey is the key of node labels used for zone-aware placement.
    // Defaults to "topology.kubernetes.io/zone".
    TopologyKey string `json:"topologyKey,omitempty"`

    // MaxSkew describes the degree to which pods may be unevenly distributed.
    // Defaults to 1.
    // +kubebuilder:validation:Minimum=1
    MaxSkew *int32 `json:"maxSkew,omitempty"`

    // WhenUnsatisfiable indicates how to deal with a pod that doesn't satisfy
    // the spread constraint. Defaults to DoNotSchedule.
    // +kubebuilder:validation:Enum=DoNotSchedule;ScheduleAnyway
    WhenUnsatisfiable corev1.UnsatisfiableConstraintAction `json:"whenUnsatisfiable,omitempty"`

    // Zones optionally restrict scheduling to the listed topology values.
    // +kubebuilder:validation:MinItems=1
    // +listType=set
    Zones []string `json:"zones,omitempty"`
}
```

This gives full control over optionality, enables `zoneAwareness: {}` to work, avoids exposing fields the operator manages internally, and allows adding proper kubebuilder validation markers.

cmd/manager/main.go
pebrc: **Avoid changing the CLI default for `exposed-node-labels`**

This changes the default from `[]string{}` to `[]string{"topology.kubernetes.io/.*", "failure-domain.beta.kubernetes.io/.*"}` for **all non-Helm operator deployments** (static manifests, `make deploy`, OLM). While it aligns with the Helm chart defaults, it's a silent behavior change.

**Suggestion**: Keep the default empty and instead make the zone awareness validation (`validNodeLabels`) aware of zone-awareness-derived topology keys. Since the operator itself injects the downward node label when zone awareness is configured, it controls both sides of the equation — so it doesn't need the user to separately authorize it via the CLI flag.

Concretely, in `validNodeLabels`, when checking a topology key from `zoneAwareness`, also allow well-known topology keys (like `topology.kubernetes.io/zone`) even if they don't match the `exposed-node-labels` flag. Or skip the exposed-node-labels check entirely for zone-awareness-derived topology keys.

This would:
1. **Keep the CLI default empty** — no behavior change for existing non-Helm deployments
2. **Still make `zoneAwareness: {}` work** without requiring users to also configure `--exposed-node-labels`
3. **Avoid a breaking change** that silently affects all non-Helm operator installations

pkg/controller/elasticsearch/nodespec/podspec.go
```diff
@@ -236,6 +254,84 @@ func buildAnnotations(
 	return annotations
 }
 
+// zoneAwarenessEnv returns the zone environment variable when zone awareness is enabled
+// at either cluster or NodeSet level. NodeSet-level configuration takes precedence.
+func zoneAwarenessEnv(nodeSet esv1.NodeSet, clusterHasZoneAwareness bool, clusterTopologyKey string) []corev1.EnvVar {
+	if !clusterHasZoneAwareness {
+		return nil
```
pebrc, line 260: **Implicit behavior for non-zone-aware NodeSets in zone-aware clusters**

When `clusterHasZoneAwareness` is true but a specific NodeSet has no `zoneAwareness`, this function still injects the `ZONE` env var. Combined with the ES config changes in `zoneAwarenessConfig()` (which adds `node.attr.zone: ${ZONE}` and `cluster.routing.allocation.awareness.attributes: k8s_node_name,zone`), **all NodeSets in the cluster are modified**, not just the ones with `zoneAwareness` configured.

The PR risk section mentions restarts for "affected NodeSets" but this means adding `zoneAwareness` to *one* NodeSet triggers rolling restarts on *every* NodeSet in the cluster. This could surprise users who expect only the configured NodeSet to be affected.

Suggestion: consider documenting this more explicitly in the PR description and release notes — "adding zoneAwareness to any NodeSet triggers a one-time rolling restart of **all** NodeSets in the cluster."

pkg/apis/elasticsearch/v1/elasticsearch_types.go
```diff
@@ -364,6 +408,28 @@ func (nsl NodeSetList) Names() []string {
...
+}
+
+// ZoneAwarenessTopologyKey returns the topology key from the first zone-aware NodeSet,
+// used as a fallback for non-zone-aware NodeSets in the same cluster.
+func (nsl NodeSetList) ZoneAwarenessTopologyKey() string {
+	for _, nodeSet := range nsl {
+		if nodeSet.ZoneAwareness == nil {
+			continue
+		}
+		return nodeSet.ZoneAwareness.TopologyKeyOrDefault()
```
pebrc, line 422: **Minor: fallback topology key depends on NodeSet ordering**

The fallback topology key for non-zone-aware NodeSets is determined by the **first** zone-aware NodeSet in the list. If a user reorders NodeSets in the spec (which should be semantically irrelevant), the fallback key could change, triggering a rolling restart on non-zone-aware NodeSets.

The validation at `validZoneAwarenessTopologyKeys` guards against the multi-key ambiguous case, but when there's a single topology key used by all zone-aware NodeSets, the ordering still matters for this function's return. Consider either:
- Documenting this ordering sensitivity, or
- Iterating all NodeSets to verify they agree on the key and using that consensus value

pkg/apis/elasticsearch/v1/elasticsearch_types.go
pebrc: **Revised recommendation: drop `maxSkew`/`whenUnsatisfiable` from the API entirely**

Building on the earlier comment about the CRD `required` fields problem — I think we can go further than cherry-picking individual fields. The `maxSkew` and `whenUnsatisfiable` toggles may not need to be in the `ZoneAwareness` API at all.

The `WithTopologySpreadConstraints` builder method already supports the escape hatch: if a user provides a spread constraint for the same `topologyKey` in their `podTemplate`, the operator's constraint is skipped and the user's values take full precedence. The operator even fills in the `labelSelector` on user-provided constraints that omit it (lines 156-158 of `pod_template.go`). This is already tested in the `"existing user spread constraint is preserved"` snapshot test.

So the API could be as simple as:

```go
type ZoneAwareness struct {
    // TopologyKey is the key of node labels used for zone-aware placement.
    // Defaults to "topology.kubernetes.io/zone".
    TopologyKey string `json:"topologyKey,omitempty"`

    // Zones optionally restrict scheduling to the listed topology values.
    // +kubebuilder:validation:MinItems=1
    // +listType=set
    Zones []string `json:"zones,omitempty"`
}
```

The `zoneAwareness` field captures the **intent** ("I want zone-aware scheduling"), while `podTemplate` handles **tuning** ("I want different skew/unsatisfiable behavior") — which is how every other scheduling concern works in ECK. Expert users who need `maxSkew: 3` or `whenUnsatisfiable: ScheduleAnyway` would write:

```yaml
nodeSets:
- name: data
  zoneAwareness:
    topologyKey: topology.kubernetes.io/zone
  podTemplate:
    spec:
      topologySpreadConstraints:
      - topologyKey: topology.kubernetes.io/zone
        maxSkew: 3
        whenUnsatisfiable: ScheduleAnyway
```

Benefits:
- Eliminates the CRD `required` fields problem entirely
- Avoids the drift concern raised in the earlier review (no re-implementation of upstream fields)
- Keeps the API surface minimal and focused on zone-awareness-specific concepts
- `zoneAwareness: {}` works as documented
- Consistent with how ECK handles every other pod template concern
- The operator auto-fills `labelSelector` even on user-provided constraints, so users don't need to know internal label details

barkbay: > API surface bloat: labelSelector, matchLabelKeys, minDomains, nodeAffinityPolicy, nodeTaintsPolicy are exposed to users but should not be user-configurable — the operator auto-generates labelSelector for topology spread constraints

Why `minDomains` should not be user-configurable? (there is also another thread [here](https://github.com/elastic/cloud-on-k8s/pull/9148#discussion_r2840958139))

pkg/apis/elasticsearch/v1/elasticsearch_types.go
barkbay: Resolving in favor of https://github.com/elastic/cloud-on-k8s/pull/9148#discussion_r2853373569 which feels more elegant.

pkg/apis/elasticsearch/v1/elasticsearch_types.go
naemono: I have to disagree here. Validation ensures that, in the case of a nodeSet without zone-awareness set, that all nodeSets agree on the same topology key. In that case, order is irrelevant, right? In any case, I've updated the comments to be more clear on this point if that helps.

pkg/controller/elasticsearch/nodespec/podspec.go
naemono: I've updated the description. I believe the upcoming release notes will fall on the ECK release manager. Correct me if I'm wrong here.

pkg/apis/elasticsearch/v1/elasticsearch_types.go
pebrc: Fair point. I think a comment is a good compromise here to make the invariant clear to future readers.

pebrc:
**Potential pod deadlock when zone-aware NodeSets use different topology keys**

`DownwardNodeLabels()` collects topology keys from ALL zone-aware NodeSets into a single union set. This set is used to generate one shared init script (via `ReconcileScriptsConfigMap`) that ALL pods use, regardless of which NodeSet they belong to. The init script busy-waits until every annotation in the set exists on the pod.

However, `getPodAnnotations()` in `node_labels.go` returns an error when the node is missing any expected label -- and a node hosting NodeSet A's pods will only have A's topology key, not B's. This means the operator never sets the annotations, and the init containers wait forever.

Example configuration that triggers this (passes validation):

```yaml
nodeSets:
- name: hot
  zoneAwareness:
    topologyKey: "custom.io/rack"
- name: warm
  zoneAwareness:
    topologyKey: "custom.io/room"
```

The chain is:

- [`DownwardNodeLabels()`](https://github.com/elastic/cloud-on-k8s/blob/59ed556115541e5e1754c50e1d6a4555a4a6a7e7/pkg/apis/elasticsearch/v1/elasticsearch_types.go#L535-L550) merges all topology keys into one set
- [`ReconcileScriptsConfigMap`](https://github.com/elastic/cloud-on-k8s/blob/59ed556115541e5e1754c50e1d6a4555a4a6a7e7/pkg/controller/elasticsearch/configmap/configmap.go#L33-L35) generates one shared script for all pods using that merged set
- [Init script](https://github.com/elastic/cloud-on-k8s/blob/59ed556115541e5e1754c50e1d6a4555a4a6a7e7/pkg/controller/elasticsearch/initcontainer/prepare_fs_script.go#L155-L162) waits for ALL annotations
- [`getPodAnnotations`](https://github.com/elastic/cloud-on-k8s/blob/59ed556115541e5e1754c50e1d6a4555a4a6a7e7/pkg/controller/elasticsearch/driver/shared/node_labels.go#L92-L116) errors when a node doesn't have all expected labels
- [Validation allows this config](https://github.com/elastic/cloud-on-k8s/blob/59ed556115541e5e1754c50e1d6a4555a4a6a7e7/pkg/controller/elasticsearch/validation/validations_test.go#L699)

Possible fixes: either tighten validation to require all zone-aware NodeSets share the same topology key, or make the init script / annotation system per-NodeSet-aware.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

config/samples/elasticsearch/elasticsearch.yaml
```diff
@@ -2,24 +2,22 @@
...
 spec:
   version: 9.3.0
   nodeSets:
   - name: default
+    # Uncomment zoneAwareness to spread this NodeSet across zones.
+    # zoneAwareness:
+    #   topologyKey: topology.kubernetes.io/zone
+    #   zones: ["us-east-1a", "us-east-1b", "us-east-1c"] # optional
+    #   maxSkew: 1 # optional, defaults to 1
+    #   whenUnsatisfiable: DoNotSchedule # optional, defaults to DoNotSchedule
```
pebrc: `maxSkew`, `whenUsatisfiable`. These are no longer supported. We should rather have either a sample or documentation to illustrate the `podTemplate` approach.

pkg/controller/elasticsearch/nodespec/resources.go
```diff
@@ -106,6 +117,10 @@ func BuildExpectedResources(
 	return nodesResources, nil
 }
 
+// hasZoneAwareness returns true when any NodeSet enables zone awareness.
+// This allows us to ensure consistency and that each nodeSet has the
+// allocation awareness setting in place.
```
pebrc: This seems stale?

cmd/manager/main.go
```diff
@@ -253,7 +253,7 @@ func Command() *cobra.Command {
 	cmd.Flags().StringSlice(
 		operator.ExposedNodeLabels,
 		[]string{},
-		"Comma separated list of node labels which are allowed to be copied as annotations on Elasticsearch Pods, empty by default",
+		"Comma separated list of node labels which are allowed to be copied as annotations on Elasticsearch Pods",
```
pebrc: restore the original comment?

pkg/apis/elasticsearch/v1/elasticsearch_types.go
```diff
@@ -452,14 +511,66 @@ type Elasticsearch struct {
 	AssocConfs map[commonv1.ObjectSelector]commonv1.AssociationConf `json:"-"`
 }
 
+// downwardNodeLabelsAnnotationValue returns the raw downward node labels annotation value.
+func (es Elasticsearch) downwardNodeLabelsAnnotationValue() string {
+	if es.Annotations == nil {
+		return ""
+	}
+	return es.Annotations[DownwardNodeLabelsAnnotation]
+}
```
pebrc, line 518: Nit: accessing a key on a `nil` map retuns the zero value (empty string for string) so this function seems redundant.

naemono:
> Possible fixes: either tighten validation to require all zone-aware NodeSets share the same topology key, or make the init script / annotation system per-NodeSet-aware.
> 

I required the same topology key here.

pkg/controller/elasticsearch/nodespec/podspec.go
```diff
@@ -236,6 +253,86 @@ func buildAnnotations(
...
+					label.StatefulSetNameLabelName: statefulSetName,
+				},
+			},
+		},
+	}
+
+	// If the nodeSet has zone awareness and zones are configured, we want to ensure
+	// that the nodeSet is only scheduled on nodes that have the zone values.
+	var requiredMatchExpressions []corev1.NodeSelectorRequirement
+	if len(nodeSet.ZoneAwareness.Zones) > 0 {
```
pebrc, line 327: When a NodeSet has zoneAwareness: {} (enabled but without listing specific zones), the function returns a TopologySpreadConstraint but does not add a NodeSelectorOpExists node affinity requirement for the topology key.

  This is a problem because:
  1. DownwardNodeLabels() adds the topology key (e.g. topology.kubernetes.io/zone) to the list of expected annotations
  2. The prepare-fs init script blocks forever waiting for those annotations to appear on the pod
  3. The operator sets those annotations from the node's label — if the node lacks the label, the annotation never appears
  4. A TopologySpreadConstraint with DoNotSchedule does not prevent scheduling on nodes that lack the topology key label — such nodes are simply excluded from the skew calculation

  The code already correctly handles the two other cases:
  - Non-zone-aware NodeSets in a zone-aware cluster get OpExists (lines 295-299, with an explicit deadlock warning comment)
  - Zone-aware NodeSets with explicit Zones get OpIn (which implies existence)

  Only the zoneAwareness: {} (no zones) case is missing the safeguard.

pkg/apis/elasticsearch/v1/elasticsearch_types.go
```diff
@@ -353,6 +357,37 @@ type NodeSet struct {
...
+	// If empty, Pods can be scheduled in any topology value for the selected topologyKey.
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:MinItems=1
+	// +listType=set
+	Zones []string `json:"zones,omitempty"`
+}
+
+// TopologyKeyOrDefault returns the configured topology key or the default.
+func (za ZoneAwareness) TopologyKeyOrDefault() string {
+	if strings.TrimSpace(za.TopologyKey) == "" {
```
pebrc: Nit: I find it a bit odd that we trim space for the comparison but then return the untrimmed value. Don't think it makes much difference in practice, but still odd.

pkg/controller/common/defaults/pod_template.go
```diff
@@ -142,6 +144,76 @@ func (b *PodTemplateBuilder) WithAffinity(affinity *corev1.Affinity) *PodTemplat
...
+// constraint already exists for their topology keys. If a constraint for a
+// topology key already exists and has no label selector, it is filled from the
+// provided constraint when one is available.
+func (b *PodTemplateBuilder) WithTopologySpreadConstraints(constraints ...corev1.TopologySpreadConstraint) *PodTemplateBuilder {
+	for _, constraint := range constraints {
+		if idx := slices.IndexFunc(b.PodTemplate.Spec.TopologySpreadConstraints, func(c corev1.TopologySpreadConstraint) bool {
+			return c.TopologyKey == constraint.TopologyKey
+		}); idx >= 0 {
+			existing := &b.PodTemplate.Spec.TopologySpreadConstraints[idx]
+			if (existing.LabelSelector == nil || k8s.IsLabelSelectorEmpty(*existing.LabelSelector)) &&
```
pebrc: Nit: Why not make k8s.IsLabelSelectorEmpty take a pointer instead of forcing each caller to do the nil check? AFAICT the number of call sites is small so I would be no big refactoring.

pkg/controller/common/defaults/pod_template.go
```diff
@@ -142,6 +144,76 @@ func (b *PodTemplateBuilder) WithAffinity(affinity *corev1.Affinity) *PodTemplat
...
+// pods from landing on nodes that lack the label.
+func (b *PodTemplateBuilder) WithRequiredNodeAffinityMatchExpressions(requirements ...corev1.NodeSelectorRequirement) *PodTemplateBuilder {
+	if len(requirements) == 0 {
+		return b
+	}
+
+	nodeSelector := ensureRequiredNodeSelector(&b.PodTemplate.Spec)
+	copiedRequirements := make([]corev1.NodeSelectorRequirement, 0, len(requirements))
+	for _, requirement := range requirements {
+		copied := corev1.NodeSelectorRequirement{
```
pebrc: Wny the manual copying? 

```golang
	for i := range requirements {
		copied := requirements[i].DeepCopy()
```

pkg/controller/common/defaults/pod_template_test.go
```diff
@@ -1409,3 +1409,399 @@ func TestPodTemplateBuilder_WithContainers(t *testing.T) {
...
+			requirements: []corev1.NodeSelectorRequirement{
+				{Key: "topology.kubernetes.io/zone", Operator: corev1.NodeSelectorOpExists},
+			},
+			wantAffinity: &corev1.Affinity{
+				NodeAffinity: &corev1.NodeAffinity{
+					RequiredDuringSchedulingIgnoredDuringExecution: &corev1.NodeSelector{
+						NodeSelectorTerms: []corev1.NodeSelectorTerm{
+							{MatchExpressions: []corev1.NodeSelectorRequirement{
+								{Key: "topology.kubernetes.io/zone", Operator: corev1.NodeSelectorOpDoesNotExist},
+								{Key: "topology.kubernetes.io/zone", Operator: corev1.NodeSelectorOpExists},
```
pebrc, line 1700: This looks like an unsatisfiable condition that could lead to pending pods? Could this happen in practice? It requires a user to define a pod template with a RequiredDuringSchedulingIgnoredDuringExecution node affinity that includes DoesNotExist (or NotIn) on the topology key (e.g. topology.kubernetes.io/zone), while also enabling zone awareness on their cluster. That's a contradictory configuration — "I don't want nodes with zone labels, but also please spread me across zones" — so it's unlikely but not impossible. The consequence would be silently unschedulable pods, which is a bad failure mode.

Can we think of ways to disarm this footgun? At least a log message or event? Maybe a validation is possible (not sure how many cases we would have to cover here)

pkg/controller/common/defaults/pod_template_test.go
```diff
@@ -1409,3 +1409,399 @@ func TestPodTemplateBuilder_WithContainers(t *testing.T) {
...
+			},
+			requirements: []corev1.NodeSelectorRequirement{
+				{Key: "topology.kubernetes.io/zone", Operator: corev1.NodeSelectorOpIn, Values: []string{"a", "b"}},
+			},
+			wantAffinity: &corev1.Affinity{
+				NodeAffinity: &corev1.NodeAffinity{
+					RequiredDuringSchedulingIgnoredDuringExecution: &corev1.NodeSelector{
+						NodeSelectorTerms: []corev1.NodeSelectorTerm{
+							{MatchExpressions: []corev1.NodeSelectorRequirement{
+								{Key: "topology.kubernetes.io/zone", Operator: corev1.NodeSelectorOpDoesNotExist},
```
pebrc, line 1765: The user has DoesNotExist on the zone key. The operator tries to inject In ["a", "b"] but hasNodeSelectorRequirementKey sees the key already exists and silently skips the In. The result is just DoesNotExist — pods will only land on nodes without the zone label, completely defeating zone awareness. Unlike the unsatisfiable case, this silently does the wrong thing without making the problem visible. This produce pods running in the wrong zones without any user feedback.

pkg/controller/elasticsearch/nodespec/podspec.go
naemono: That's a great catch. I've updated this logic, and updated the tests to ensure the behavior moving forward.

pkg/controller/common/defaults/pod_template.go
naemono: Just simplify forgetting about this function on occasions.

pkg/controller/common/defaults/pod_template_test.go
naemono: It seems possible, and it can't hurt to avoid it, and test the exact case. I've updated the logic, and while doing so, found some other odd edge cases which I've either updated the logic and/or added tests.

pkg/controller/common/defaults/pod_template_test.go
naemono: I've solved this issue [here](https://github.com/elastic/cloud-on-k8s/pull/9148/changes/822041692009b77535c3f94fbbaf5f207b773637#diff-e0b41f58404fb3608247e1e16a04b3b27a1abcafd46e1507d5b86b8c05aa30ffR212-R214), and augmented this [test](https://github.com/elastic/cloud-on-k8s/pull/9148/changes/822041692009b77535c3f94fbbaf5f207b773637#diff-d95bd5417282e8eb4398d4a167a435dd4f6009e457a0dcb2306be08039f20713R1741) to show the new functionality.

pkg/controller/elasticsearch/validation/warnings.go
```diff
@@ -5,14 +5,21 @@
...
 )
 
+const (
+	zoneAwarenessAffinityDoesNotExistWarningMsg = "Zone awareness injects an Exists requirement for the topology key; DoesNotExist on the same key makes this node selector term unsatisfiable, though other OR'd terms may still allow scheduling"
+	zoneAwarenessAffinityNotInWarningMsg        = "Zone awareness may conflict with required node affinity using NotIn on the topology key; this can make pods unschedulable depending on node labels"
+)
+
 var warnings = []validation{
 	noUnsupportedSettings,
+	validZoneAwarenessAffinityWarnings,
```
pebrc, line 22: I think we have an issue here in that the warnings are not surfaced in the webhook. I think that because when I added the warnings mechanism in November 2019 there was just no way to do that. I think it would be OK to fix that in a follow up PR though, because we now can return warnings in webhooks. [Update]: we already have an issue. Let's maybe make sure that https://github.com/elastic/cloud-on-k8s/issues/6461 is implement in the same release.

naemono: @pebrc I've assigned to me, and will begin investigation to have a follow-up pr.

pkg/controller/elasticsearch/settings/merged_config.go
```diff
@@ -25,6 +25,12 @@ const nodeAttrK8sNodeName = "k8s_node_name"
 
 var nodeAttrNodeName = fmt.Sprintf("%s.%s", esv1.NodeAttr, nodeAttrK8sNodeName)
 
+const nodeAttrZone = "zone"
+
+var nodeAttrZoneName = fmt.Sprintf("%s.%s", esv1.NodeAttr, nodeAttrZone)
+
+const envVarReferenceFormat = "${%s}"
```
pebrc: Nit: group consts and vars together

barkbay:
@naemono I added the `v3.4.0` label, looks like it was missing.

naemono:
> @naemono I added the `v3.4.0` label, looks like it was missing.

Thank you!

PR #9151, opened 2026-02-19
Fix ES shutdown node request: correct allocation delay type and expose it in client API

moukoublen:
The Elasticsearch [PUT shutdown API](https://www.elastic.co/docs/api/doc/elasticsearch/v8/operation/operation-shutdown-put-node#operation-shutdown-put-node-body-application-json-allocation_delay) expects allocation_delay as a time-value string like "5m" or "300s", but the previous implementation used a bare time.Duration which Go's encoding/json serializes as an integer (nanoseconds). This would produce invalid requests.

  * Fixed AllocationDelay type in ShutdownRequest: Changed from time.Duration (which serializes as an integer nanosecond count) to a custom *Duration type that serializes as an Elasticsearch-compatible seconds string (e.g. "300s"). This fixes incorrect JSON being sent in the PUT shutdown request body.
  * Added allocationDelay parameter to PutShutdown client function: The Client.PutShutdown interface and its clientV7 implementation now accept an optional *time.Duration parameter, allowing callers to specify a custom allocation delay when initiating a node shutdown. Passing nil preserves the previous behavior (Elasticsearch default of 5 minutes for restart-type shutdowns).
  * Implement formatter and parser functionality for elasticsearch [time values](https://www.elastic.co/docs/reference/elasticsearch/rest-apis/api-conventions#time-units).
  * Added a custom Duration JSON type: Introduced a Duration wrapper around time.Duration with MarshalJSON/UnmarshalJSON methods that use the new parser and formater to produce Elasticsearch-compatible duration strings (e.g. "20m" for 20 minutes).

### DIFF
--- a/pkg/controller/elasticsearch/client/model.go
+++ b/pkg/controller/elasticsearch/client/model.go
@@ -606,7 +606,7 @@ func ParseTimeValue(s string) (time.Duration, error) {
 			break
 		}
 	}
-	if sep == len(s) {
+	if sep == 0 || sep == len(s) {
 		return 0, fmt.Errorf("invalid elasticsearch duration: %q", orig)
 	}
 
--- a/pkg/controller/elasticsearch/client/model_test.go
+++ b/pkg/controller/elasticsearch/client/model_test.go
@@ -545,6 +545,11 @@ func TestParseTimeValue(t *testing.T) {
 			wantErr: true,
 		},
 		{
+			name:    "just numbers",
+			input:   "123",
+			wantErr: true,
+		},
+		{
 			name:    "no unit",
 			input:   "42",
 			wantErr: true,
@@ -590,36 +595,6 @@ func TestParseTimeValue(t *testing.T) {
 			require.NoError(t, err)
 			assert.Equal(t, tt.want, got)
 		})
-	}
-}
-
-func BenchmarkParseTimeValue(b *testing.B) {
-	inputs := []string{"0s", "3d", "5h", "15m", "42s", "750ms", "300micros", "123nanos", "-10s"}
-	b.ResetTimer()
-	for b.Loop() {
-		for _, s := range inputs {
-			_, _ = ParseTimeValue(s)
-		}
-	}
-}
-
-func BenchmarkFormatTimeValue(b *testing.B) {
-	inputs := []time.Duration{
-		0,
-		3 * 24 * time.Hour,
-		5 * time.Hour,
-		15 * time.Minute,
-		42 * time.Second,
-		750 * time.Millisecond,
-		300 * time.Microsecond,
-		123 * time.Nanosecond,
-		-10 * time.Second,
-	}
-	b.ResetTimer()
-	for b.Loop() {
-		for _, d := range inputs {
-			_ = FormatTimeValue(d)
-		}
 	}
 }
 


prodsecmachine:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/79a1b533-8e73-4d85-a5cb-1aaf55fc0a59?refs=5d942af59b34862c113b27847b2680e57a6d3e23&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/79a1b533-8e73-4d85-a5cb-1aaf55fc0a59/license?refs=5d942af59b34862c113b27847b2680e57a6d3e23&source=prChecksComment) |

moukoublen:
@pkoutsovasilis, personally, I am leaning towards having a distinct type for such established patterns as the Elasticsearch duration string format. Both for potential reusability, and because structs that use those types are more "self-described" and type-safe than when having a generic string. Unless you see any concerns with it, I’d prefer to continue with this approach.

P.S. The `formatAsSeconds` function is not ideal; perhaps a better Elasticsearch duration string formatter should be implemented, ~~but that's something to create a ticket for, perhaps~~. 

**Update:** Implemented elasticsearch time value formatter and parser according to the [documentation](https://www.elastic.co/docs/reference/elasticsearch/rest-apis/api-conventions#time-units).

pkg/controller/elasticsearch/client/model.go
```diff
@@ -526,3 +527,115 @@ type FileSettingsErrors struct {
 	ErrorKind string   `json:"error_kind"`
 	Errors    []string `json:"errors"`
 }
+
+// Duration wraps time.Duration to support JSON serialization as a Go duration string (e.g. "5m0s").
```
pkoutsovasilis: I believe that this comment needs to be updated, compound duration strings like `5m0s` aren't supported by elasticsearch

pkg/controller/elasticsearch/client/model_test.go
```diff
@@ -182,29 +319,308 @@ func TestShutdownResponse(t *testing.T) {
...
+	inputs := []string{"0s", "3d", "5h", "15m", "42s", "750ms", "300micros", "123nanos", "-10s"}
+	b.ResetTimer()
+	for b.Loop() {
+		for _, s := range inputs {
+			_, _ = ParseTimeValue(s)
+		}
+	}
+}
+
+func BenchmarkFormatAsElasticsearchDuration(b *testing.B) {
```
pkoutsovasilis:
```suggestion
func BenchmarkFormatTimeValue(b *testing.B) {
```

pkg/controller/elasticsearch/client/model_test.go
```diff
@@ -182,29 +319,308 @@ func TestShutdownResponse(t *testing.T) {
...
+				require.Error(t, err)
+				return
+			}
+			require.NoError(t, err)
+			assert.Equal(t, tt.want, got)
+		})
+	}
+}
+
+func BenchmarkParseElasticsearchDuration(b *testing.B) {
```
pkoutsovasilis:
```suggestion
func BenchmarkParseTimeValue(b *testing.B) {
```

pkg/controller/elasticsearch/client/model.go
```diff
@@ -526,3 +527,115 @@ type FileSettingsErrors struct {
...
+		n = ns / int64(time.Microsecond)
+		unit = "micros"
+	}
+
+	return strconv.FormatInt(n, 10) + unit
+}
+
+// ParseTimeValue parses a duration string using Elasticsearch time unit conventions
+// (d, h, m, s, ms, micros, nanos) and returns the equivalent time.Duration.
+// It is the inverse of FormatAsElasticsearchDuration.
```
pkoutsovasilis: I believe this should be `FormatTimeValue` instead of `FormatAsElasticsearchDuration`?

moukoublen: Yes (same for your other comments). I decided to rename the initial names I gave the functions and forgot to change them everywhere. Done in [03bde51](https://github.com/elastic/cloud-on-k8s/pull/9151/commits/03bde5189064da832f07ba588cce4585c6994cb7) 😅

pkg/controller/elasticsearch/client/model.go
moukoublen: Done in [03bde51](https://github.com/elastic/cloud-on-k8s/pull/9151/commits/03bde5189064da832f07ba588cce4585c6994cb7) and also specified that it is the Elasticsearch time value string format.

pkoutsovasilis (approved):
Code changes look good and address the underlying issue. However, the complexity of the solution still has me on the fence - we're adding a custom `Duration` type with `UnmarshalJSON` that will never be used, when we could just use `FormatTimeValue` directly in `PutShutdown`. I don't aim to block this PR, hence the approval, but I'd appreciate another review from @pebrc, @naemono before merging.

moukoublen:
> we're adding a custom Duration type with UnmarshalJSON that will never be used

`allocation_delay` is also present in the [Get the shutdown status](https://www.elastic.co/docs/api/doc/elasticsearch/v8/operation/operation-shutdown-get-node) that we currently parse, we just don't parse this specific field atm. 

I am not sure about the never here.

But if `UnmarshalJSON` is a big issue, we can remove that function and keep only the custom marshaller. 

> when we could just use FormatTimeValue directly in PutShutdown

Using a generic string removes any guarantees about what values are valid. With `Duration`, the compiler enforces that only proper duration values can be assigned. That gives us type safety, better self-documentation, and avoids pushing validation to runtime.

pkoutsovasilis:
> `allocation_delay` is also present in the [Get the shutdown status](https://www.elastic.co/docs/api/doc/elasticsearch/v8/operation/operation-shutdown-get-node) that we currently parse, we just don't parse this specific field atm.

https://github.com/elastic/cloud-on-k8s/blob/main/pkg/controller/elasticsearch/client/model.go#L478 do we parse it, though? Because if the field is missing then we don't need the UnmarshalJSON. If we do need it, then this PR should change also this struct?

pkg/controller/elasticsearch/client/model_test.go
```diff
@@ -182,29 +319,308 @@ func TestShutdownResponse(t *testing.T) {
...
+				require.Error(t, err)
+				return
+			}
+			require.NoError(t, err)
+			assert.Equal(t, tt.want, got)
+		})
+	}
+}
+
+func BenchmarkParseTimeValue(b *testing.B) {
```
pebrc: This seems a bit overkill for code that is not on any hot path.

pkg/controller/elasticsearch/client/model.go
```diff
@@ -526,3 +527,115 @@ type FileSettingsErrors struct {
 	ErrorKind string   `json:"error_kind"`
 	Errors    []string `json:"errors"`
 }
+
+// Duration wraps time.Duration to support JSON serialization as an elasticsearch time value string (e.g. "5m").
+type Duration time.Duration
```
pebrc, line 532: The scope of this change depends on whether we need to support parsing ES time values from GET shutdown responses (i.e. `UnmarshalJSON`). If we do, the full `FormatTimeValue`/`ParseTimeValue` implementation makes sense and I'm supportive of this change.

If we only need to *produce* valid time value strings for the PUT request, this could be significantly simpler: the existing `formatAsSeconds` function in `client.go` already formats a `time.Duration` as a valid ES time value (e.g. `"300s"`). The `MarshalJSON` could just reuse that pattern — ES is equally happy with `"300s"` as `"5m"`. That would avoid the need for `ParseTimeValue`, the unit-selection logic in `FormatTimeValue`, and ~90 lines of new code.

pkg/controller/elasticsearch/client/model.go
```diff
@@ -526,3 +527,115 @@ type FileSettingsErrors struct {
...
+	}
+
+	sep := 0
+	for i, c := range s {
+		sep = i
+		if c < '0' || c > '9' {
+			break
+		}
+	}
+	if sep == len(s) {
```
pebrc: Minor: if we keep `ParseTimeValue`, the separator-finding loop has a subtle flaw for all-digit inputs (e.g. `"42"`). After the loop, `sep` is `len(s)-1` (the last loop index), not `len(s)`, so the `sep == len(s)` guard doesn't catch it. The function still returns an error — but via the "unknown unit" branch with a misleading message (`unknown elasticsearch duration unit "2" in "42"`).

Suggested fix:
```go
sep := len(s)
for i, c := range s {
    if c < '0' || c > '9' {
        sep = i
        break
    }
}
if sep == 0 || sep == len(s) {
    return 0, fmt.Errorf("invalid elasticsearch duration: %q", orig)
}
```

This correctly handles: empty string (caught earlier), no digits (`sep == 0`), all digits / no unit (`sep == len(s)`), and valid input (`0 < sep < len(s)`).

pebrc (commented):
I don't want to block this PR but I think it is very likely overkill for what we need to achieve. We could just reuse formatSeconds and save ourselves 80% of the code. 

@moukoublen do you need to parse the duration from a node shutdown request for some of your future work on rolling restarts or why do you think we need to support the full format?

moukoublen:
@pebrc 
> do you need to parse the duration from a node shutdown request for some of your future work on rolling restarts or why do you think we need to support the full format?

> The scope of this change depends on whether we need to support parsing ES time values from GET shutdown responses (i.e. UnmarshalJSON). If we do, the full FormatTimeValue/ParseTimeValue implementation makes sense and I'm supportive of this change.

To be perfectly honest, I am not entirely sure right now whether I will use the parse functionality during the rolling restart (it is still forming). 

But since having a concrete type has the benefits I mentioned earlier, I thought it was a good opportunity to implement Elasticsearch TimeValue properly. 

my pov is that since the code is pretty simple, I feel it is worth having a full implementation. 


But I could also remove that functionality.

pkg/controller/elasticsearch/client/model_test.go
moukoublen: I created the benchmarks during implementation to compare some ideas. I could remove them.

pkoutsovasilis:
What has me on the fence is that by using concrete types with custom `UnmarshalJSON` we tightly couple unmarshalling and validation, which I'm not in favor of. As an example, we use `GetShutdown` in several places already in our code. With this approach, if `Duration` is eventually used in `NodeShutdown` (since `allocation_delay` is present in the GET response), and Elasticsearch changes how this string is encoded or our code has a bug, every call to `GetShutdown` would fail at the unmarshalling stage - even if the caller doesn't care about `allocation_delay` at all (as the ones we already have). 

The safest approach IMO is to leave duration fields as plain strings in response structs and introduce a method like `GetAllocationDelay() (time.Duration, error)` on `NodeShutdown` where we utilise `ParseTimeValue`, letting the caller decide if and how to handle parsing errors. This wouldn't affect any existing calls and gives us more flexibility. For the PutShutdown my initial proposal still remains we can use `FormatTimeValue` inside `PutShutdown` to transition from time.Duration to string.

moukoublen:
I think the conversation is starting to expand into a broader discussion about type aliases and custom marshal/unmarshal behavior, which is genuinely interesting (at least to me 😊). However, I feel that the scope of this PR is narrower than that. So I am sorry in advance if I am becoming too broad 😅.

From my point of view, this isn’t about validation, it’s about parsing. We have a concrete duration type that, in JSON, can only be represented as a string because JSON doesn’t have a native duration type.

For example, Kubernetes apimachinery follows the same approach here:
https://github.com/kubernetes/apimachinery/blob/eae5e9edd98282a6b222186b932c853aeff4c47f/pkg/apis/meta/v1/duration.go#L27-L29

Lazy parsing is certainly an option, but in my opinion, it makes sense only in specific cases. As you mentioned, it can help in situations where there’s a protocol mismatch, allowing calls that don’t touch the problematic fields to proceed without error. That said, it also risks pushing parsing errors deeper into the system, making them harder to trace and reason about.

More generally, if we defer parsing, we end up handling what is essentially a wire-level error much later in the application layer. That can lead to repetitive and potentially unnecessary error handling across the codebase, which I think we should try to avoid. If we take that idea a bit further (and exaggerate a bit to make my point clearer), we could even lazy-load the JSON objects instead of parsing them straight into structs. That can be useful sometimes (and I faced a situation where this was really helpful in unstable rest APIs), but I think it only really makes sense in pretty specific scenarios (e.g. unstable APIs).

(Update) From my perspective, it comes down to whether TimeValue is an unstable data type or one that’s likely to change in the future.


Having said that, if you both still have concerns about that, I will completely remove the code. @pebrc @pkoutsovasilis

pebrc:
@moukoublen Just to be clear: you have an approval from Panos and I said I won't block this. So from my perspective you are good to go. If you feel strongly about including the full  format support, I'm happy to go with that — the ES time value format is very stable, I don't think it has changed in the last 10 years, so the coupling concerns don't carry much     
  weight here in my opinion. Don't let this discussion hold you up.

Issue #9079, opened 2026-01-28
Support rolling restarts of Elasticsearch clusters


pebrc:
## Problem
Users need to trigger graceful rolling restarts without making spec changes. Current workarounds are hacky:

* Adding dummy environment variables
* Modifying secure settings or node settings
* Manually deleting pods (bypasses ECK's graceful orchestration)

## Use Cases
* Recovering from degraded states - Clearing stuck threads, connection pool issues
* Forcing a fresh start - After prolonged operation, clearing accumulated state
* Troubleshooting - Controlled restart as part of debugging

## Proposed Solution
New annotation that triggers a graceful rolling restart when its value changes:

```yaml
apiVersion: elasticsearch.k8s.elastic.co/v1
kind: Elasticsearch
metadata:
  name: my-cluster
  annotations:
    eck.k8s.elastic.co/restart-trigger: "2026-01-14T12:00:00Z"
```

Implementation: Include the annotation value in the pod template config hash. When it changes, the hash changes, triggering a rolling restart through existing upgrade orchestration.

## Behavior:

**Graceful**: Uses rolling upgrade logic with changeBudget, health checks, shard reallocation
**Idempotent**: Same value = no restart; only value changes trigger restart
**Auditable**: Value can encode timestamps or ticket numbers

## Open Questions

Especially for larger clusters shard relocation during rolling restarts is undesirable. Should we disable them?

```
PUT _cluster/settings
{
  "persistent": {
    "cluster.routing.allocation.enable": "primaries"
  }
}
```
Or should we follow up on https://github.com/elastic/cloud-on-k8s/issues/5582 and make the shard allocation delay on restarts configurable?

See also #5305 for the same problem in the context of host restarts

moukoublen:
# ECK Support Rolling Restarts of Elasticsearch Clusters

## Purpose of this Document
This document describes the need for a supported way to trigger a graceful rolling restart of an Elasticsearch cluster managed by ECK without changing the Elasticsearch or pod spec. It outlines the problem, user use cases, and design principles that a solution should follow. It is intended to align stakeholders and guide the design of a future feature (e.g. an annotation, a field on the Elasticsearch resource, or a separate API) that reuses ECK's existing upgrade and node-shutdown orchestration.


## Design Principles

[...]

PR #9172, opened 2026-02-26
Support rolling restarts of Elasticsearch clusters

moukoublen:
## Overview

This PR adds a supported, spec-free way to trigger a graceful rolling restart of an Elasticsearch cluster managed by ECK. Users can set or change an annotation on the Elasticsearch resource; the operator reuses the existing rolling upgrade path (node shutdown API, predicates, one-at-a-time deletion) so no new restart mechanism is introduced and no desired spec changes are required.

Design rationale and principles are documented [here](https://github.com/elastic/cloud-on-k8s/issues/9079#issuecomment-3952920071).

### New annotations (API)

- **`eck.k8s.elastic.co/restart-trigger`**  
  Setting or changing this annotation (e.g. to a timestamp like `2026-01-14T12:00:00Z`) triggers a rolling restart. The value is propagated to pod annotations, which changes the StatefulSet template hash so all pods become upgrade candidates. The existing upgrade flow then runs unchanged.

- **`eck.k8s.elastic.co/restart-allocation-delay`**  
  Optional duration string (e.g. `"5m"`, `"20m"`) used as `allocation_delay` for every node shutdown API call during rolling restarts and upgrades. If unset, Elasticsearch’s default (e.g. 5m) is used. Invalid values are logged and ignored.

### Implementation

- **Pod template (`nodespec/podspec.go`)**  
  When building the pod template, if `restart-trigger` is set on the Elasticsearch resource, it is merged into pod annotations (same key). If the annotation is removed from the resource but was previously present on pods, the last value is kept on the template to avoid triggering another restart when the user deletes the annotation.  
  `ConfigHashAnnotationName` is exported for use in the upgrade logic.

- **Node shutdown (`shutdown/node.go`)**  
  `NodeShutdown` gains an optional `allocationDelay *time.Duration`. `NewNodeShutdown` accepts it and passes it to `PutShutdown` instead of `nil`, so the node shutdown API receives the configured allocation delay when set.

- **Upgrade path (`driver/stateful/upgrade.go`, `upgrade_pods_deletion.go`)**  
  - **Reason and allocation delay:** `shutdownReasonAndAllocationDelay()` resolves the shutdown reason (restart-trigger value if set, otherwise resource version) and parses `restart-allocation-delay` from the Elasticsearch resource; both are passed into `NodeShutdown`.
  - **Annotation-only restarts:** `isAnnotationTriggeredRestart()` detects when the only difference from desired state is the restart-trigger (all pods have the same config hash as the expected StatefulSet template but some pods don’t yet have the current trigger value). This is used to tailor status messages.
  - **Observability:** The existing `UpgradeOperation` in status is reused. When the restart is annotation-triggered, node messages say “Deleting pod for rolling restart”, “Not all Pods are ready for a cluster rolling restart”, etc., instead of “rolling upgrade” wording.

### Tests

- **Unit:** `podspec` and `shutdown/node` tests updated for new behavior; upgrade tests extended (e.g. annotation-triggered restart detection, reason and allocation delay resolution).
- **E2E:** `TestRestartTriggerAnnotationCausesRollingRestart` creates an HA cluster, sets `restart-trigger` and `restart-allocation-delay`, and asserts a rolling restart completes (pods ready, cluster green, data intact).

### E2E / test helpers

- Elasticsearch builder and mutation helpers updated as needed to support annotation-based flows and the new e2e test.

## Usage example

```yaml
kind: Elasticsearch
metadata:
  name: my-cluster
  annotations:
    eck.k8s.elastic.co/restart-trigger: "2026-01-14T12:00:00Z"
    eck.k8s.elastic.co/restart-allocation-delay: "20m"
spec:
  # ... unchanged spec
```

To trigger another restart later, update the `restart-trigger` value (e.g. to a new timestamp). Removing the annotation does not trigger a new restart; the last trigger value is retained on the pod template.

### DIFF
--- a/pkg/controller/elasticsearch/driver/stateful/upgrade_pods_deletion.go
+++ b/pkg/controller/elasticsearch/driver/stateful/upgrade_pods_deletion.go
@@ -279,7 +279,7 @@ func (ctx *upgradeCtx) pendingPodsMessage() string {
 
 func (ctx *upgradeCtx) deleteAllMessage() string {
 	if ctx.isAnnotationTriggeredRestart {
-		return "Deleting pod for cluster rolling restart"
+		return "Deleting pod for full cluster restart"
 	}
 
 	return "Deleting pod for full cluster upgrade"


prodsecmachine:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/58caaf0f-3ec3-49ec-ba05-7a35bf3847ad?refs=1770cffe77c865b591dd1aba0b5324d3bdace96f&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/58caaf0f-3ec3-49ec-ba05-7a35bf3847ad/license?refs=1770cffe77c865b591dd1aba0b5324d3bdace96f&source=prChecksComment) |

docs/reference/rolling-restart.md
```diff
@@ -0,0 +1,42 @@
+---
+navigation_title: Elasticsearch rolling restart
+applies_to:
+  deployment:
+    eck: all
+---
+
+# Elasticsearch rolling restart
```
barkbay: It feels like a product feature documentation that should go somewhere [here](https://github.com/elastic/docs-content/tree/main/deploy-manage/deploy/cloud-on-k8s).

moukoublen: Thanks @barkbay, I will open a PR there as well.

Should I leave that `rolling-restart.md` in this repo or remove it?

pkg/controller/elasticsearch/nodespec/resources.go
```diff
@@ -77,6 +77,12 @@ func BuildExpectedResources(
 		return nil, err
 	}
 
+	// we retrieve the current pods restart trigger annotation.
+	podsRestartTriggerAnnotation, err := es_sset.GetActualPodsRestartTriggerAnnotationForCluster(client, es)
+	if err != nil {
+		return nil, err
+	}
```
moukoublen: I don't particularly love passing this value transparently through `BuildStatefulSet` -> `BuildPodTemplateSpec` to `buildAnnotations`, where it is actually used, but it was either that or using the k8s client to fetch the pods again and again per NodeSet inside the `BuildPodTemplateSpec`.

docs/reference/rolling-restart.md
barkbay: Products documentation lives in https://github.com/elastic/docs-content, so you should create a PR there. Note though that changes in docs-content are published almost immediately, so we'll need to wait for this feature to be released before merging the doc PR. `docs/reference` is mainly for generated files (with the exception of `docs/reference/eck-configuration-flags.md`, do you happen to remember why, @eedugon?).

moukoublen:
buildkite test this -f p=kind,t=TestRestartTriggerAnnotationCausesRollingRestart -m s=9.3.1

docs/reference/rolling-restart.md
eedugon: @barkbay , `docs/reference/eck-configuration-flags.md` was left in cloud-on-k8s because its nature is purely a reference doc (the list configuration flags), regardless of it being auto-generated or not. It's not a guide to configure ECK but more a reference of configuration flags that can be used, similar to Elasticsearch or Kibana configuration settings that are kept in product repositories.

If you feel it was a bad call because you prefer to keep `cloud-on-k8s` docs just for automatically generated docs we could definitely change it. That shouldn't be a problem at all. Just ping me in private for anything you need.

The nature of this PR and feature requires a narrative doc explaining the feature and how to use it, in `docs-content` for sure, as you have mentioned.

If you want my help to find a good spot (or spots) for the new feature (whenever it's implemented) to appear in the docs feel free to ping me in private @moukoublen (or Michael) and we can create a docs issue for it, with some suggestions.

At first sight I'd say this feature (rolling restarts through ECK) should be mentioned in:

- Somewhere within the ECK specific docs area --> I'd suggest to create a new doc called Rolling restarts in the `Managed deployments` section of the doc, maybe right after `Accessing services` and before `Configure deployments`.

image

  - [Nodes orchestration](https://www.elastic.co/docs/deploy-manage/deploy/cloud-on-k8s/nodes-orchestration) describes with details how ECK performs the orchestration of nodeSets in multiple use cases, and maybe it's a good place to add a note or comment about rolling restarts (for your consideration).

- In other areas:
  -  [Start and stop services](https://www.elastic.co/docs/deploy-manage/maintenance/start-stop-services) --> In that doc we have a note about ECK, so this new feature would fit perfectly fine in that section of the docs, I would mention it there too and link wherever we decide to document it fully.

docs/reference/rolling-restart.md
```diff
@@ -0,0 +1,42 @@
...
+## Annotations
+
+Set these annotations on the Elasticsearch resource:
+
+| Annotation | Description |
+|------------|-------------|
+| `eck.k8s.elastic.co/restart-trigger` | **Required to trigger.** Set or change this value (for example to a timestamp) to start a rolling restart. The value is propagated to pod annotations and is visible in Elasticsearch's node shutdown API as the shutdown reason. |
+| `eck.k8s.elastic.co/restart-allocation-delay` | Optional. Duration string (for example `"5m"`, `"20m"`) used as `allocation_delay` for the node shutdown API during rolling restarts and upgrades. If unset, Elasticsearch's default (for example 5m) is used. Invalid values or negative values, are logged and ignored. |
+
+To trigger another rolling restart later, update the `restart-trigger` value (for example to a new timestamp). Removing the annotation does **not** trigger a new restart; the operator retains the last trigger value on the pod template.
```
barkbay: We should maybe mention that removing the annotation does not stop/cancel the cluster restart.

pkg/controller/elasticsearch/nodespec/podspec.go
```diff
@@ -227,8 +229,21 @@ func buildAnnotations(
 		annotations[esv1.TransportCertDisabledAnnotationName] = "true"
 	}
 
+	// if the restart annotation is set, add it to pods' annotations.
+	restartTrigger := es.Annotations[esv1.RestartTriggerAnnotation]
+	if restartTrigger == "" {
+		// if restart annotation is missing but it was previously present on pods,
```
barkbay, line 252: Does this mean a user cannot reuse the same annotation value? (setting it, removing it, then setting the same value again would be a no-op?)

pkg/controller/elasticsearch/nodespec/podspec.go
```diff
@@ -63,6 +63,7 @@ func BuildPodTemplateSpec(
 	setDefaultSecurityContext bool,
 	policyConfig PolicyConfig,
 	meta metadata.Metadata,
+	podsRestartTriggerAnnotation string,
```
barkbay:
```suggestion
	actualPodsRestartTriggerAnnotationValue string,
```

pkg/controller/elasticsearch/nodespec/podspec.go
```diff
@@ -201,6 +202,7 @@ func buildAnnotations(
 	keystoreResources *keystore.Resources,
 	scriptsContent string,
 	policyAnnotations map[string]string,
+	podsRestartTriggerAnnotation string,
```
barkbay:
```suggestion
	actualPodsRestartTriggerAnnotationValue string,
```

pkg/controller/elasticsearch/nodespec/resources.go
```diff
@@ -77,6 +77,12 @@ func BuildExpectedResources(
 		return nil, err
 	}
 
+	// we retrieve the current pods restart trigger annotation.
+	podsRestartTriggerAnnotation, err := es_sset.GetActualPodsRestartTriggerAnnotationForCluster(client, es)
```
barkbay:
```suggestion
	actualPodsRestartTriggerAnnotationValue, err := es_sset.GetActualPodsRestartTriggerAnnotationForCluster(client, es)
```

pkg/controller/elasticsearch/driver/stateful/upgrade.go
```diff
@@ -64,12 +66,15 @@ func (d *Driver) handleUpgrades(
 		return results.WithError(err)
 	}
 
+	triggeredRestart := isAnnotationTriggeredRestart(d.ES, expectedResources, podsToUpgrade)
```
barkbay:
```suggestion
	isAnnotationTriggeredRestart := isAnnotationTriggeredRestart(d.ES, expectedResources, podsToUpgrade)
```

pkg/controller/elasticsearch/nodespec/podspec.go
moukoublen: Perhaps `current` prefix instead of `actual` is even more descriptive, wdyt?

moukoublen:
@barkbay 

> To help users avoid these pitfalls I was wondering if we should consider improving our validation mechanism. Maybe by adding an admission webhook warnings (non-blocking) for both cases. The webhook already supports WithWarnings. We could also add a log, or generate a Warning event.

I will explore that. We need to retreive the pods and checks the annotations on validation. 

> I think reusing the same annotation value (setting it, removing it, then setting it again) is a no-op since pods already carry that value

Indeed, it is. It was a choice between that and, when removing the restart trigger, triggering a new restart. The second felt "unnatural", so the current solution was followed.

pkg/controller/elasticsearch/nodespec/podspec.go
barkbay: This was to be consistent with `GetActualPodsRestartTriggerAnnotationForCluster` and with the `GetActual*` functions we have in the code (`GetActualPodsForStatefulSet`, `GetActualPodsForCluster`, `GetActualMastersForCluster`, `GetActualPods`). No strong opinion though, let's just use something more "explicit".

pkg/controller/elasticsearch/driver/stateful/upgrade.go
moukoublen: made `isTriggeredRestart` so it doesn't collide with the function name.

pkg/controller/elasticsearch/nodespec/resources.go
moukoublen: done in [0ecd467](https://github.com/elastic/cloud-on-k8s/pull/9172/commits/0ecd467a93daa47131f929f72b934b5761f30d63)

pkg/controller/elasticsearch/nodespec/podspec.go
moukoublen: done in [0ecd467](https://github.com/elastic/cloud-on-k8s/pull/9172/commits/0ecd467a93daa47131f929f72b934b5761f30d63)

pkg/controller/elasticsearch/nodespec/podspec.go
moukoublen: done in [0ecd467](https://github.com/elastic/cloud-on-k8s/pull/9172/commits/0ecd467a93daa47131f929f72b934b5761f30d63)

docs/reference/rolling-restart.md
moukoublen: Done in [6a926e6](https://github.com/elastic/cloud-on-k8s/pull/9172/commits/6a926e68a86b7e2a74f8ea3100f3a22a3a2e7e93)

pkg/controller/elasticsearch/nodespec/podspec.go
moukoublen: Yes, mentioned in [6a926e6](https://github.com/elastic/cloud-on-k8s/pull/9172/commits/6a926e68a86b7e2a74f8ea3100f3a22a3a2e7e93)

See also: https://github.com/elastic/cloud-on-k8s/pull/9172#issuecomment-4030628854

docs/reference/rolling-restart.md
moukoublen: Thanks both, 

I opened this [PR](https://github.com/elastic/docs-content/pull/5492) to add the two new annotations as new sections in node orchestration.

moukoublen:
@barkbay warning was added in valdation webhook
Screenshot 2026-03-16 at 8 02 07 PM
Screenshot 2026-03-16 at 7 57 36 PM

pkg/controller/elasticsearch/driver/stateful/upgrade_pods_deletion.go
```diff
@@ -268,3 +268,27 @@ func runPredicates(
 	// All predicates passed!
 	return nil, nil
 }
+
+func (ctx *upgradeCtx) pendingPodsMessage() string {
+	if ctx.isAnnotationTriggeredRestart {
+		return "Not all pods are ready for a cluster rolling restart"
```
pebrc, line 274: This message is inaccurate. In the deleteAll path there is no rolling restart. It is a full cluster restart.

pkg/controller/elasticsearch/validation/validations.go
```diff
@@ -549,3 +552,47 @@ func validLicenseLevel(ctx context.Context, es esv1.Elasticsearch, checker licen
 	}
 	return errs
 }
+
+func validateRestartTriggerWarnings(ctx context.Context, k8sClient k8s.Client, oldCR, newCR esv1.Elasticsearch) string {
```
pebrc, line 565: Should we also issue a warning if an illegal allocation delay value is specified?

pkg/controller/elasticsearch/nodespec/podspec.go
```diff
@@ -244,8 +246,21 @@ func buildAnnotations(
...
+	restartTrigger := es.Annotations[esv1.RestartTriggerAnnotation]
+	if restartTrigger == "" {
+		// if restart annotation is missing but it was previously present on pods,
+		// keep the old one, to avoid restarting when the annotation is deleted by user.
+		if actualPodsRestartTriggerAnnotationValue != "" {
+			restartTrigger = actualPodsRestartTriggerAnnotationValue
+		}
+	}
+	if restartTrigger != "" {
+		annotations[esv1.RestartTriggerAnnotation] = restartTrigger
```
pebrc, line 261: Maybe add a comment here that we are relying on the fact that this annotation is not part of the config hash to distinguish between spec changes and pod rotation caused by the restart annotation later in `isAnnotationTriggeredRestart`

pkg/apis/elasticsearch/v1/elasticsearch_types.go
```diff
@@ -44,6 +44,14 @@ const (
 	// TransportCertDisabledAnnotationName is the annotation that indicates that ECK-managed transport certs have been disabled for the Pod.
 	TransportCertDisabledAnnotationName = "elasticsearch.k8s.elastic.co/self-signed-transport-cert-disabled"
 
+	// RestartTriggerAnnotation allows users to trigger a graceful rolling restart by setting or changing this
+	// annotation value on the Elasticsearch resource. The value is propagated to pod annotations, causing the
+	// StatefulSet template hash to change and all pods to become upgrade candidates.
+	RestartTriggerAnnotation = "eck.k8s.elastic.co/restart-trigger"
+	// RestartAllocationDelayAnnotation configures the allocation_delay passed to the Elasticsearch node shutdown
+	// API during rolling restarts and upgrades. The value must be a valid Go duration string (e.g. "5m", "1h").
```
pebrc, line 54: I think this deserves separate call out in the docs, because it applies to any restart. This is almost like a separate little mini-feature in this rolling restart feature.

moukoublen: Yes, I added two separate sections for those.

https://github.com/elastic/docs-content/blob/eck-3.4.0/deploy-manage/deploy/cloud-on-k8s/nodes-orchestration.md?plain=1#L21-L22

pkg/controller/elasticsearch/nodespec/podspec.go
moukoublen: Done in [10db8c7](https://github.com/elastic/cloud-on-k8s/pull/9172/commits/10db8c76bd9e46ce4b610ca215ffc6fb4a0d66ec)

pkg/controller/elasticsearch/validation/validations.go
moukoublen: Done in [10db8c7](https://github.com/elastic/cloud-on-k8s/pull/9172/commits/10db8c76bd9e46ce4b610ca215ffc6fb4a0d66ec)

Added a warning both on create and update

pkg/controller/elasticsearch/driver/stateful/upgrade_pods_deletion.go
moukoublen: @pebrc I’m not sure I fully understand this part, could you please add a bit more explanation or context? Thanks 🙏

moukoublen: Ok I got 😅 fixed in [1770cff](https://github.com/elastic/cloud-on-k8s/pull/9172/commits/1770cffe77c865b591dd1aba0b5324d3bdace96f)

PR #9179, opened 2026-02-27
avoid empty reconcile requests in StackConfigPolicy secret watch

pkoutsovasilis:
## Summary
Fixes a bug where the StackConfigPolicy controller logs errors when `managedNamespaces` is configured:
```
  error: unable to get: / because of unknown namespace for the cache
```
The issue was in `reconcileRequestForSoftOwnerPolicy()` which pre-allocated a slice of reconcile requests and used `continue` to skip non-StackConfigPolicy owners, leaving zero-value entries with empty namespace/name in the returned slice.

## Steps to reproduce

1. run ECK operator with `managedNamespaces`:
```
make CGO_ENABLED=0 LICENSE_PUBKEY=$PWD/license.key GO_TAGS='release' MANAGED_NAMESPACES='elastic-system' run
```
Note: you can go with the helm installation if you prefer

2. Deploy an Elasticsearch cluster and a StackConfigPolicy:
```yaml
kind: Elasticsearch
metadata:
  name: elasticsearch
  namespace: elastic-system
  labels:
    env: test
  spec:
    version: 9.3.0
    nodeSets:
    - name: default
       count: 1
---
kind: StackConfigPolicy
metadata:
  name: test-policy
  namespace: elastic-system
spec:
  resourceSelector:
    matchLabels:
      env: test
    elasticsearch:
      clusterSettings:
        indices.recovery.max_bytes_per_sec: "100mb"
```

3. Trigger a modification on an Elasticsearch-owned secret:
```console
kubectl annotate secret -n elastic-system elasticsearch-es-elastic-user test=trigger
```

4. Observe the error in operator logs

5. Repeat the above with the code of this PR

### DIFF
--- a/pkg/controller/stackconfigpolicy/controller.go
+++ b/pkg/controller/stackconfigpolicy/controller.go
@@ -111,25 +111,31 @@ func addWatches(mgr manager.Manager, c controller.Controller, r *ReconcileStackC
 
 func reconcileRequestForSoftOwnerPolicy() handler.TypedEventHandler[*corev1.Secret, reconcile.Request] {
 	return handler.TypedEnqueueRequestsFromMapFunc[*corev1.Secret](func(ctx context.Context, secret *corev1.Secret) []reconcile.Request {
-		softOwners, err := reconciler.SoftOwnerRefs(secret)
-		if err != nil {
-			ulog.Log.Error(err, "Fail to get soft-owner policies of secret", "secret_name", secret.GetName(), "secret_namespace", secret.GetNamespace())
-			return nil
-		}
-
-		if len(softOwners) == 0 {
-			return nil
-		}
-
-		var requests []reconcile.Request
-		for _, nsn := range softOwners {
-			if nsn.Kind != policyv1alpha1.Kind {
-				continue
-			}
-			requests = append(requests, reconcile.Request{NamespacedName: types.NamespacedName{Namespace: nsn.Namespace, Name: nsn.Name}})
-		}
-		return requests
+		return toReconcileRequests(secret)
 	})
+}
+
+// toReconcileRequests returns reconcile requests for StackConfigPolicy owners of the given secret.
+// Returns nil if the secret has no soft owners or if none of the owners are StackConfigPolicy resources.
+func toReconcileRequests(secret *corev1.Secret) []reconcile.Request {
+	softOwners, err := reconciler.SoftOwnerRefs(secret)
+	if err != nil {
+		ulog.Log.Error(err, "Fail to get soft-owner policies of secret", "secret_name", secret.GetName(), "secret_namespace", secret.GetNamespace())
+		return nil
+	}
+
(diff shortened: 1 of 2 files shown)


prodsecmachine:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/2ff77edd-dcfb-4b3f-87b5-30ec847f0328?refs=adf5fdd70a03ceb98c1cd46309a62c635ffa0d3b&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/2ff77edd-dcfb-4b3f-87b5-30ec847f0328/license?refs=adf5fdd70a03ceb98c1cd46309a62c635ffa0d3b&source=prChecksComment) |

pkg/controller/stackconfigpolicy/controller.go
```diff
@@ -111,25 +111,31 @@ func addWatches(mgr manager.Manager, c controller.Controller, r *ReconcileStackC
...
-		for idx, nsn := range softOwners {
-			if nsn.Kind != policyv1alpha1.Kind {
-				continue
-			}
-			requests[idx] = reconcile.Request{NamespacedName: types.NamespacedName{Namespace: nsn.Namespace, Name: nsn.Name}}
+	if len(softOwners) == 0 {
+		return nil
+	}
+
+	var requests []reconcile.Request
```
moukoublen, line 131:
```suggestion
	requests := make([]reconcile.Request, 0, len(softOwners))
```
[totally nit / can be avoided]: since we know the "max" possible length, we could initialize with capacity (instead of length like it was before)

This avoids the first couple of reallocations of the internal buffer as we append elements to the slice. `prealloc` linter of golangci-lint should suggest that I don't know why it does not (I will check).

_Anyway, this is not a hot path, so an optimization like that does not really move the needle._

pkoutsovasilis: as a thumb rule I have that if we filter in a loop then don't apply prealloc (even if I fail to apply it always 🙂). I think that recent versions the `prealloc` linter have adopted the same rule. That said, I get what you mean about the performance, but since I know also that all the soft-owners (at least as of now) must be of the same kind, I prefer to get out without any allocations since we have a handful of secrets that are soft-owned rather than pre-allocating something that we may not need. If you have strong feelings lemme know

pkoutsovasilis:
buildkite test this -f p=gke,t=TestStackConfigPolicy*

pkg/controller/stackconfigpolicy/controller.go
pebrc: I agree pre-allocating when filtering does not make sense, especially because this is not a hot path.

PR #9188, opened 2026-03-02
Skip per-shard replica checks for GREEN clusters in require_started_replica predicate

### LINKED ISSUE #9187: `require_started_replica` predicate incorrectly blocks rolling upgrades
### Bug Summary
  The `require_started_replica` upgrade predicate incorrectly blocks rolling upgrades when:
  1. The cluster health is GREEN but has shards in `RELOCATING` state
  2. The cluster health is YELLOW but replicas are in `RELOCATING` state (which are in-sync)

  ### Root Cause
  The predicate treats `RELOCATING` shards as "not in-sync", but this is incorrect. According to clarification from the Elasticsearch team, the shard lifecycle is:
```
  UNASSIGNED → INITIALIZING → STARTED → RELOCATING
```
  A shard becomes in-sync during the `INITIALIZING` phase. When a shard is `RELOCATING`, it was previously `STARTED` and is being moved - **the source copy remains fully synchronized**. The `INITIALIZING` copy on the target node is not yet in-sync, but the `RELOCATING` source copy is safe. Additionally, when cluster health is GREEN, all shards are in-sync by definition, so the predicate should immediately allow deletion without checking individual shard states.

**Reference**: https://www.elastic.co/docs/deploy-manage/upgrade/deployment-or-cluster/elasticsearch#upgrade-process (step 9: "Wait for the status column to switch to green")

### Current Behavior
  ECK blocks node deletion when:
  - Cluster is GREEN but has `RELOCATING` shards
  - Cluster is YELLOW and replicas are `RELOCATING` (even though they're in-sync)
  Operator logs show repeated messages like:

  "failed_predicates":{"require_started_replica":["es-node-0"]}

  ### Expected Behavior
  - **GREEN cluster**: Always allow deletion (GREEN means all shards are in-sync)
  - **YELLOW cluster**: Allow deletion if replicas are `STARTED` or `RELOCATING` (both are in-sync). Only block if replicas are `INITIALIZING` or `UNASSIGNED`

  ### Impact
  This bug can cause rolling upgrades to stall indefinitely on clusters where the desired balance allocator is actively rebalancing shards (e.g., to balance disk usage across nodes). Since relocations can be continuous in such clusters, the upgrade never proceeds.

pkoutsovasilis:
## Summary
This PR fixes a bug in the `require_started_replica` upgrade predicate that was incorrectly blocking rolling upgrades even when the Elasticsearch cluster was GREEN.

### Issue
The `require_started_replica` predicate was evaluating per-shard replica states even for GREEN clusters, instead of trusting the cluster health status. This caused rolling upgrades to stall unnecessarily when shards were relocating, even though GREEN status indicates all replicas are fully synchronised.

### Root Cause
The predicate was evaluating per-shard replica states even for GREEN clusters, instead of trusting the cluster health status.


prodsecmachine:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/81fa1153-37f5-4e64-9a6d-ebd786b02f8f?refs=4f5b7ebdb14f426faf8262d272bd43fb12cf2b93&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/81fa1153-37f5-4e64-9a6d-ebd786b02f8f/license?refs=4f5b7ebdb14f426faf8262d272bd43fb12cf2b93&source=prChecksComment) |

pkoutsovasilis:
buildkite test this -f p=kind,E2E_TAGS=es -m s=9.3.0,s=8.19.0

pkoutsovasilis:
buildkite test this -f p=gke,E2E_TAGS=es -m s=9.3.0,s=8.19.0

pkg/controller/elasticsearch/client/model.go
```diff
@@ -46,16 +46,21 @@ type Health struct {
...
+// Note: in this safety model, INITIALIZING shards are considered not yet in-sync and
+// therefore unsafe for upgrades, so their presence is treated as blocking activity.
+// RELOCATING shards, on the other hand, are NOT considered blocking activity because
+// the source shard is already in-sync; the shard lifecycle is:
+// UNASSIGNED -> INITIALIZING -> STARTED -> RELOCATING.
 func (h Health) HasShardActivity() bool {
-	return h.TimedOut || // make sure request did not time out (i.e. no pending events)
-		h.NumberOfInFlightFetch > 0 || // no shards being fetched
-		h.InitializingShards > 0 || // no shards initializing
-		h.RelocatingShards > 0 // no shards relocating
```
barkbay, line 58: Let's remove it from `Health`. It's unused after this change I think, and unlikely to be needed for future orchestration logic.

pkg/controller/elasticsearch/client/model.go
```diff
@@ -46,16 +46,21 @@ type Health struct {
...
-// It can be the case if some shards are being fetched, relocated or initialized.
+// HasShardActivity indicates that there is some shard activity in the cluster that
+// could affect data safety during rolling upgrades.
 // It's only reliable if Health result was created with wait_for_events=languid
 // so that there are no pending initialisations in the task queue.
 // It returns true if the status request has timed out.
+//
+// Note: in this safety model, INITIALIZING shards are considered not yet in-sync and
+// therefore unsafe for upgrades, so their presence is treated as blocking activity.
+// RELOCATING shards, on the other hand, are NOT considered blocking activity because
```
barkbay: Thanks for digging into this and explaining it here, that's great 👍

barkbay (dismissed):
LGTM

pkg/controller/elasticsearch/driver/stateful/upgrade_predicates.go
```diff
@@ -365,6 +365,11 @@ var predicates = [...]Predicate{
 			if err != nil {
 				return false, err
 			}
+			// If the cluster health is reported as GREEN, consider it sufficiently healthy
+			// to skip the additional per-shard replica checks in this predicate.
+			if health.Status == esv1.ElasticsearchGreenHealth {
```
naemono, line 370: AI find:
**GREEN early-return skips per-shard checks when initializing shards may exist.** Elasticsearch can report GREEN with `initializing_shards > 0`. The early-return in `require_started_replica` bypasses the shard-level safety check, potentially allowing deletion of a node that holds the only STARTED copy of a shard whose replica is still INITIALIZING.

pkg/controller/elasticsearch/driver/stateful/upgrade_predicates.go
```diff
@@ -376,21 +381,25 @@ var predicates = [...]Predicate{
 			}
 			// We maintain two data structures to record:
 			// * The total number of replicas for a shard
-			// * How many of them are STARTED
-			startedReplicas := make(map[string]int)
+			// * How many of them are STARTED or RELOCATING (both are considered in-sync here)
+			// Shard lifecycle (general): UNASSIGNED -> INITIALIZING -> STARTED -> RELOCATING
+			// In this predicate a shard is only considered in-sync once it reaches STARTED, and it
+			// remains in-sync while in the RELOCATING state.
+			inSyncReplicas := make(map[string]int)
```
naemono: An additional AI find:

**RELOCATING replicas on the candidate node itself are not counted.** The `replicas` map in `require_started_replica` is populated by iterating shards *not* on the candidate. A shard relocating *away* from the candidate is excluded entirely, so the predicate may still block deletion of a node whose shard is mid-relocation away from it even though the destination already has a STARTED copy. The asymmetry is subtle but real.

naemono (commented):
Initial pass with AI found 2 things of note.

pkg/controller/elasticsearch/driver/stateful/upgrade_predicates.go
pkoutsovasilis: The GREEN early-return is intentional and correct. As clarified by the elasticsearch team, GREEN means that all replicas are fully synchronized and the cluster has sufficient redundancy.

pkg/controller/elasticsearch/driver/stateful/upgrade_predicates.go
pkoutsovasilis: I am not entirely sure I get the point of this AI find but maybe you @naemono can help me understand it since you posted it. That said, I believe that the current logic is correct. The predicate only checks PRIMARIES on the candidate node so relocating replicas on the candidate don't affect the decision. For a PRIMARY relocating away from the candidate: the source is RELOCATING (on candidate) and the target is INITIALIZING (on destination) - as documented in [ShardRoutingState.java#L42-L46](https://github.com/elastic/elasticsearch/blob/main/server/src/main/java/org/elasticsearch/cluster/routing/ShardRoutingState.java#L42-L46). Since we only count STARTED/RELOCATING as in-sync the INITIALIZING target is not counted, and deletion is correctly blocked.

pkg/controller/elasticsearch/client/model.go
pkoutsovasilis: removed it here 03a3d4aa4a3957fed5ba7f5afb059a54fedd4199

pkoutsovasilis:
buildkite test this -f p=gke,E2E_TAGS=es -m s=9.3.0,s=8.19.0

pkg/controller/elasticsearch/driver/stateful/upgrade_predicates.go
naemono: Ok, I'm no expert in this area, but I further looked into this claim and how the code handles this, and it's false. It stated that the destination "already has a STARTED copy." That's not really possible during relocation as the destination is always INITIALIZING until the relocation finishes. So, sorry for the confusion/delays in getting this merged. I'll resolve this.

naemono (dismissed):
👍

pebrc (dismissed):
LGTM. I was wondering where this original implementation was coming from and it turns out it is verbatim from the official rolling upgrade docs: https://www.elastic.co/docs/deploy-manage/upgrade/deployment-or-cluster/elasticsearch 

>During a rolling upgrade, primary shards assigned to a node running the new version cannot have their replicas assigned to a node with the old version. The new version might have a different data format that is not understood by the old version.

> If it is not possible to assign the replica shards to another node (there is only one upgraded node in the cluster), the replica shards remain unassigned and status stays yellow.

> In this case, you can proceed once there are no initializing or relocating shards (check the init and relo columns).

> As soon as another node is upgraded, the replicas can be assigned and the status will change to green.

However I believe, this guidance is probably a bit conservative and written for manual rolling upgrades.  We still honor this guidance in substance if not verbatim with our shard level predicates `require_started_replica` in combination with the cluster level `if_yellow_only_restart_upgrading_nodes_with_unassigned_replicas` predicate.

pebrc (changes requested):
After talking to David Turner once more I think we are OK with the short-circuiting for green but we should keep the conservative waiting on relocating shards in yellow state.

pebrc (approved):
LGTM!

PR #9193, opened 2026-03-03
Fix race condition in ClustersExpectation.create

barkbay:
When multiple goroutines call `ForCluster` for the same cluster simultaneously, they could both pass the `get()` check (finding no existing expectations) and then both call `create()`. Without double-checked locking, the second goroutine would overwrite the first one's Expectations instance, potentially losing state.

This fix moves the lock acquisition to the beginning of `create()` and adds a second check after acquiring the lock to return any existing `Expectations` created by another goroutine while waiting for the lock.

This PR also adds a deterministic unit test that verifies the fix by simulating the exact interleaving that would trigger the race condition.

Note that in practice this is not a problem for ECK, but since we aim to share this package with other services I think we should fix it for correctness.

### DIFF
--- a/pkg/controller/common/expectations/per_cluster.go
+++ b/pkg/controller/common/expectations/per_cluster.go
@@ -58,9 +58,13 @@ func (c *ClustersExpectation) get(cluster types.NamespacedName) (*Expectations, 
 }
 
 func (c *ClustersExpectation) create(cluster types.NamespacedName) *Expectations {
-	expectations := NewExpectations(c.client, c.object)
 	c.lock.Lock()
 	defer c.lock.Unlock()
+	// Check again in case another goroutine created it while we were waiting for the lock.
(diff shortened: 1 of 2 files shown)


prodsecmachine:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/1bd03089-a61e-4cc1-8470-e5b257c69903?refs=a60a48c8f26074ac820176516621e4e293043159&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/1bd03089-a61e-4cc1-8470-e5b257c69903/license?refs=a60a48c8f26074ac820176516621e4e293043159&source=prChecksComment) |

Issue #9056, opened 2026-01-20
Migrate from deprecated record.EventRecorder to events.EventRecorder (events.k8s.io/v1)


pkoutsovasilis:
## Context

With the upgrade to controller-runtime v0.23.0 in #9054, the `mgr.GetEventRecorderFor()` method has been deprecated in favor of the new events API introduced in kubernetes-sigs/controller-runtime#3262.

### What needs to change:

1. **Code changes**: Replace `mgr.GetEventRecorderFor()` with `mgr.GetEventRecorder()`
2. **RBAC updates**: Update event permissions from the core (`""`) apiGroup to `events.k8s.io` apiGroup
3. **Interface migration**: 
   - Old: `record.EventRecorder` → New: `events.EventRecorder`
   - Note: The new interface does **not** include the `AnnotatedEventf()` method

### Impact:

- Loss of ability to produce annotated events (need to evaluate if this is acceptable or if we need a workaround)
- RBAC manifests need updates across all controllers
- Changes to how we produce events

## References

- Deprecation PR: kubernetes-sigs/controller-runtime#3262
- Related PR: #9054

PR #9195, opened 2026-03-03
Migrate from `record.EventRecorder` to `events.EventRecorder`

dtuck9:
Fixes issue: https://github.com/elastic/cloud-on-k8s/issues/9056

### Changes made
1. Replace `mgr.GetEventRecorderFor()` with `mgr.GetEventRecorder()`
2. Replace `record.EventRecorder` with `events.EventRecorder` -- aliased as `toolsevents.EventRecord` following the existing pattern and to avoid collision with the `events` package in cloud-on-k8s.
3. Remove `annotations.go`, as there is no longer a way to write annotations to the event metadata.
4. Added `EmitEvent` and `EmitEventf` functions to wrap `Eventf` w/ `fmt.Sprintf` -- This allows `go vet` to catch a mismatch in the format parameters and provided arguments

Unfortunately there wasn't a great way to break this up into smaller PRs due to the following shared dependencies and the inherent requirement to use the same `EventRecord` throughout any given controller:
1. `k8sutil.MaybeEmitErrorEvent`
2. `association.AllowVersion`
3. `association.IsConfiguredIfSet`
6. `driver.Interface` -- `Recorder()` function

### Background
The deprecated `record.EventRecorder` interface is defined as:
```
type EventRecorder interface {
	// Event constructs an event from the given information and puts it in the queue for sending.
	// 'object' is the object this event is about. Event will make a reference-- or you may also
	// pass a reference to the object directly.
	// 'eventtype' of this event, and can be one of Normal, Warning. New types could be added in future
	// 'reason' is the reason this event is generated. 'reason' should be short and unique; it
	// should be in UpperCamelCase format (starting with a capital letter). "reason" will be used
	// to automate handling of events, so imagine people writing switch statements to handle them.
	// You want to make that easy.
	// 'message' is intended to be human readable.
	//
	// The resulting event will be created in the same namespace as the reference object.
	Event(object [runtime](https://pkg.go.dev/k8s.io/apimachinery/pkg/runtime).[Object](https://pkg.go.dev/k8s.io/apimachinery/pkg/runtime#Object), eventtype, reason, message [string](https://pkg.go.dev/builtin#string))

	// Eventf is just like Event, but with Sprintf for the message field.
	Eventf(object [runtime](https://pkg.go.dev/k8s.io/apimachinery/pkg/runtime).[Object](https://pkg.go.dev/k8s.io/apimachinery/pkg/runtime#Object), eventtype, reason, messageFmt [string](https://pkg.go.dev/builtin#string), args ...interface{})

	// AnnotatedEventf is just like eventf, but with annotations attached
	AnnotatedEventf(object [runtime](https://pkg.go.dev/k8s.io/apimachinery/pkg/runtime).[Object](https://pkg.go.dev/k8s.io/apimachinery/pkg/runtime#Object), annotations map[[string](https://pkg.go.dev/builtin#string)][string](https://pkg.go.dev/builtin#string), eventtype, reason, messageFmt [string](https://pkg.go.dev/builtin#string), args ...interface{})
}
```

Whereas `events.EventRecorder` is defined as:
```
type EventRecorder interface {
	// Eventf constructs an event from the given information and puts it in the queue for sending.
	// 'regarding' is the object this event is about. Event will make a reference-- or you may also
	// pass a reference to the object directly.
	// 'related' is the secondary object for more complex actions. E.g. when regarding object triggers
	// a creation or deletion of related object.
	// 'type' of this event, and can be one of Normal, Warning. New types could be added in future
	// 'reason' is the reason this event is generated. 'reason' should be short and unique; it
	// should be in UpperCamelCase format (starting with a capital letter). "reason" will be used
	// to automate handling of events, so imagine people writing switch statements to handle them.
	// You want to make that easy.
	// 'action' explains what happened with regarding/what action did the ReportingController
	// (ReportingController is a type of a Controller reporting an Event, e.g. k8s.io/node-controller, k8s.io/kubelet.)
	// take in regarding's name; it should be in UpperCamelCase format (starting with a capital letter).
	// 'note' is intended to be human readable.
	Eventf(regarding [runtime](https://pkg.go.dev/k8s.io/apimachinery/pkg/runtime).[Object](https://pkg.go.dev/k8s.io/apimachinery/pkg/runtime#Object), related [runtime](https://pkg.go.dev/k8s.io/apimachinery/pkg/runtime).[Object](https://pkg.go.dev/k8s.io/apimachinery/pkg/runtime#Object), eventtype, reason, action, note [string](https://pkg.go.dev/builtin#string), args ...interface{})
}
```
So the `Event` and `Eventf` functions were effectively consolidated to just `Eventf`, the signature for which was also updated to include 2 additional parameters:
1. `related`: the secondary object for more complex actions. E.g. when regarding object triggers a creation or deletion of related object.; **This did not immediately appear relevant for our current use cases.**
7. `action string`: explains what happened with `regarding`/what action did the controller take in `regarding`'s name; it should be in UpperCamelCase format (starting with a capital letter).

Lastly, the `Annotatedf` function was entirely removed, and there is now no way to add annotations to the generated events.

### DIFF
--- a/pkg/controller/common/events/events.go
+++ b/pkg/controller/common/events/events.go
@@ -51,7 +51,7 @@ const (
 	// EventActionCertificateReconciliation describes the certificate reconciliation step the controller was taking
 	// when the event was triggered.
 	EventActionCertificateReconciliation = "CertificateReconciliation"
-	// EventActionStatusUpdate s used when the resource health has changed or the controller fails to update the status sub-resource.
+	// EventActionStatusUpdate is used when the resource health has changed or the controller fails to update the status sub-resource.
 	EventActionStatusUpdate = "StatusUpdate"
 	// EventActionVersionCheck describes the version check step the controller was taking when the event was triggered.
 	EventActionVersionCheck = "VersionCheck"


prodsecmachine:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/e0305b22-7053-4d24-a154-8950abc1d52a?refs=ecdf8abcf345503c60f3018510634a055fceec85&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/e0305b22-7053-4d24-a154-8950abc1d52a/license?refs=ecdf8abcf345503c60f3018510634a055fceec85&source=prChecksComment) |

pkg/controller/common/configref.go
```diff
@@ -82,16 +82,16 @@ func ParseConfigRefToConfig(
 	}
 	data, exists := secret.Data[secretKey]
 	if !exists {
-		msg := fmt.Sprintf("unable to parse configRef secret %s/%s: missing key %s", namespace, configRef.SecretName, secretKey)
-		driver.Recorder().Event(resource, corev1.EventTypeWarning, events.EventReasonUnexpected, msg)
+		msg := fmt.Sprintf("unable to retrieve configRef secret %s/%s: missing key %s", namespace, configRef.SecretName, secretKey)
```
dtuck9, line 86: This was a drive-by fix since in the event we need to troubleshoot something, it would not be possible to differentiate it from the event triggered on line 93.

pkg/controller/common/configref_test.go
```diff
@@ -138,7 +138,7 @@ func TestParseConfigRef(t *testing.T) {
 			},
 			wantErr:     true,
 			wantWatches: []string{watchName},
-			wantEvent:   "Warning Unexpected unable to parse configRef secret ns/my-secret: missing key configFile.yml",
+			wantEvent:   "Warning Unexpected unable to retrieve configRef secret ns/my-secret: missing key configFile.yml",
```
dtuck9, line 141: Result of the drive-by fix.

pkg/controller/logstash/pipelines/ref_test.go
```diff
@@ -137,7 +137,7 @@ func TestParsePipelinesRef(t *testing.T) {
 			},
 			wantErr:     true,
 			wantWatches: []string{watchName},
-			wantEvent:   "Warning Unexpected unable to parse configRef secret ns/my-secret: missing key configFile.yml",
+			wantEvent:   "Warning Unexpected unable to retrieve configRef secret ns/my-secret: missing key configFile.yml",
```
dtuck9, line 140: A result of the drive-by fix.

test/e2e/apm/association_test.go
```diff
@@ -213,13 +213,14 @@ func TestAPMAssociationWhenReferencedESDisappears(t *testing.T) {
...
 
-							if prevStatusString == establishedString && currStatusString != prevStatusString {
+							// evt.Message defined as fmt.Sprintf("Association status changed from [%s] to [%s]")
+							statusIndex := strings.Index(evt.Message, establishedString)
+							if statusIndex == 32 {
 								assocLostEventSeen = true
 							}
 
-							if currStatusString == establishedString {
+							if statusIndex > 32 {
```
dtuck9: Because annotations on events are no longer supported, this was my best approximation of ensuring the appropriate event was sent, without refactoring the whole test.

pkg/controller/agent/fleet.go
```diff
@@ -248,7 +248,7 @@ func maybeReconcileFleetEnrollment(params Params, result *reconciler.Results) En
 		// we requeue if Kibana is unavailable: surface this condition to the user
 		message := "Delaying deployment of Elastic Agent in Fleet Mode as Kibana is not available yet"
 		log.Info(message)
-		params.EventRecorder.Event(&params.Agent, corev1.EventTypeWarning, events.EventReasonDelayed, message)
+		params.EventRecorder.Eventf(&params.Agent, nil, corev1.EventTypeWarning, events.EventReasonDelayed, events.EventActionAccessCheck, message)
```
naemono:
```suggestion
		params.EventRecorder.Eventf(&params.Agent, nil, corev1.EventTypeWarning, events.EventReasonDelayed, events.EventActionAccessCheck, "%s", message)
```

I believe all of these `Event(` -> `Eventf(` changes likely need this fix to avoid something like `%!s(MISSING)` in the output if for some reason an event contains a `%` (such as `50%`)

pkg/controller/agent/fleet.go
```diff
@@ -271,12 +271,12 @@ func maybeReconcileFleetEnrollment(params Params, result *reconciler.Results) En
 		message := "ECK cannot setup Fleet enrollment. Waiting for Kibana credentials. This should be a transient issue."
 		log.V(1).Info(err.Error())
 		log.Info(message)
-		params.EventRecorder.Event(&params.Agent, corev1.EventTypeWarning, events.EventReasonDelayed, message)
+		params.EventRecorder.Eventf(&params.Agent, nil, corev1.EventTypeWarning, events.EventReasonDelayed, "Authorization", message)
```
naemono:
```suggestion
		params.EventRecorder.Eventf(&params.Agent, nil, corev1.EventTypeWarning, events.EventReasonDelayed, events.EventActionAuthorization, message)
```

Some of these calls use the new constants defined, and some use raw strings. We should be consistent and define all of the constants instead.

dtuck9: Yeah I was following an existing pattern where single-use strings were defined in-line and the more frequently used were defined as constants, but I'm fine making this change.

pkg/controller/agent/fleet.go
dtuck9: I unfortunately think you are correct. `go vet` seems to miss that in this case, so it seems like there's not really another way to avoid a runtime error.

That also means that any mismatch in the number of args and the number of `%` could trigger that so that would always be a risk.

pkg/controller/kibana/driver.go
pebrc: Are we missing the action parameter here?

  The signature of MaybeEmitErrorEvent if I am not mistaken is:
  func MaybeEmitErrorEvent(r toolsevents.EventRecorder, err error, obj runtime.Object, reason, action, message string, args ...any)

  Here the format string "Invalid version '%s': %v" lands in the action position, kb.Spec.Version becomes the message, and err becomes a stray variadic arg. This compiles but produces a garbled event at runtime.

pkg/controller/association/reconciler.go
```diff
@@ -251,7 +250,7 @@ func (r *Reconciler) reconcileAssociation(ctx context.Context, association commo
 		// external reference, update association conf to associate the unmanaged resource
 		expectedAssocConf, err := r.ExpectedConfigFromUnmanagedAssociation(association)
 		if err != nil {
-			r.recorder.Eventf(association.Associated(), corev1.EventTypeWarning, events.EventAssociationError, "Failed to reconcile external resource %q: %v", assocRef.NameOrSecretName(), err.Error())
+			r.recorder.Eventf(association.Associated(), nil, corev1.EventTypeWarning, events.EventAssociationError, "GetExpectedConfig", "Failed to reconcile external resource %q: %v", assocRef.NameOrSecretName(), err.Error())
```
pebrc: I think inline strings are OK if there is only one occurence, but I think we need to keep in mind that they are user-facing. So they need to be tied to something the user can understand. Something like "AssociationReconciliation" would be maybe more appropriate here.

pkg/controller/elasticsearch/driver/shared/reconcile.go
```diff
@@ -283,7 +283,7 @@ func ReconcileSharedResources(
 		msg := "Could not update remote clusters in Elasticsearch settings, re-queuing"
 		if err != nil {
 			log.Info(msg, "err", err, "namespace", es.Namespace, "es_name", es.Name)
-			params.ReconcileState.AddEvent(corev1.EventTypeWarning, events.EventReasonUnexpected, msg)
+			params.ReconcileState.AddEvent(corev1.EventTypeWarning, events.EventReasonUnexpected, "RemoteClusterUpdateSettings", msg)
```
pebrc: `RemoteClusterUpdateSettings` is a bit of a mouthful: `RemoteClusterConfiguration` is maybe a more general term that users can understand?

pkg/controller/autoops/state.go
```diff
@@ -127,7 +127,7 @@ func (s *State) Finalize(isReconciled bool, reconciliationMessage string) {
 	switch {
 	case !isReconciled:
 		s.UpdateWithPhase(autoopsv1alpha1.ApplyingChangesPhase)
-		s.AddEvent(corev1.EventTypeWarning, events.EventReasonDelayed, reconciliationMessage)
+		s.AddEvent(corev1.EventTypeWarning, events.EventReasonDelayed, "Finalize", reconciliationMessage)
```
pebrc: `Finalize` is very vague? Also overloaded with K8s terminology: maybe `AutoOpsReconciliation` (not sure myself)

pkg/controller/agent/fleet.go
pebrc: As mentioned elsewhere one-off strings are OK'ish (if there is a single use). What we need to strive for though is that they make sense for a user seeing them in an event. Mabye the existing `EventActionAccessCheck` or `AssociationPreconditionCheck` or something along those lines would be a better fit.

pkg/controller/elasticsearch/elasticsearch_controller.go
```diff
@@ -279,7 +279,7 @@ func (r *ReconcileElasticsearch) internalReconcile(
 			"namespace", es.Namespace,
 			"es_name", es.Name,
 		)
-		reconcileState.AddEvent(corev1.EventTypeWarning, events.EventReasonValidation, err.Error())
+		reconcileState.AddEvent(corev1.EventTypeWarning, events.EventReasonValidation, "WarningCheck", err.Error())
```
pebrc:
```suggestion
		reconcileState.AddEvent(corev1.EventTypeWarning, events.EventReasonValidation, "PreReconciliationValidation", err.Error())
```
or reuse `EventActionValidation`

pkg/controller/agent/fleet.go
dtuck9: I may have gone a bit overboard. But I've updated everything so that `go vet` will appropriately catch any mismatches between the format and the arguments passed to eliminate the risk of `%!s(MISSING)` or `%!(EXTRA` at runtime.

pkg/controller/kibana/driver.go
dtuck9: I did catch that too and should be in the last commit I just pushed.

pkg/controller/autoops/state.go
dtuck9: Yeah I was struggling with some of these especially because I don't have the entire flow figured out yet.

dtuck9:
buildkite test this

pkg/controller/agent/fleet.go
pkoutsovasilis: I think that the `fmt.Sprintf` approach is a necessary evil to have proper `go vet` support. However having to call Eventf in this "special" way every time we want to emit an event IMO is easy to miss. Thus I would vote to exploit it in another way by first embedding fmt.Sprintf within the `MaybeEmitErrorEvent` and leaving the args ...any param and then by introducing a new k8s util func for the rest of Event emission we need, namely `EmitEvent`.

```go
// MaybeEmitErrorEvent emits an event if the error is report-worthy
func MaybeEmitErrorEvent(r toolsevents.EventRecorder, err error, obj runtime.Object, reason, action, message string, args ...any) {
	// ignore nil errors and conflict issues
	if err == nil || apierrors.IsConflict(err) {
		return
	}

	r.Eventf(obj, nil, corev1.EventTypeWarning, reason, action, "%s", fmt.Sprintf(message, args...))
}

// EmitEvent emits an event
func EmitEvent(r toolsevents.EventRecorder, obj runtime.Object, eventtype, reason, action, message string, args ...any) {
	r.Eventf(obj, nil, eventtype, reason, action, "%s", fmt.Sprintf(message, args...))
}
```
**Note**: don't daisy chain the above, because go vet works only when it detects the `fmt.Sprintf` within the func body

the following calls now are properly detected by go vet and we also reduce the code verbosity of emitting events
```go
k8s.MaybeEmitErrorEvent(r.recorder, err, as, events.EventReconciliationError, events.EventActionCertificateReconciliation, "Certificate reconciliation error: %v")
k8s.EmitEvent(r.recorder, as, corev1.EventTypeWarning, events.EventReconciliationError, events.EventActionCertificateReconciliation, "Certificate reconciliation error: %v")
```

Of course this means that every call of `.Eventf` would have now to become `k8s.EmitEvent` but you already change these in this PR.

No strong feeling on the proposal above, I just think it would make things clearer, wdyt?

naemono: I think I like the approach suggested by @pkoutsovasilis .

dtuck9: Yeah that seems reasonable.

dtuck9:
buildkite test this

pkg/controller/agent/controller.go
```diff
@@ -223,7 +223,7 @@ func (r *ReconcileAgent) validate(ctx context.Context, agent agentv1alpha1.Agent
 	// Run create validations only as update validations require old object which we don't have here.
 	if _, err := agent.ValidateCreate(); err != nil {
 		logconf.FromContext(ctx).Error(err, "Validation failed")
-		k8s.MaybeEmitErrorEvent(r.recorder, err, &agent, events.EventReasonValidation, err.Error())
+		k8s.MaybeEmitErrorEvent(r.recorder, err, &agent, events.EventReasonValidation, events.EventActionValidation, "%s", err.Error())
```
dtuck9: Having the `"%s"` here is still required -- without it, `go vet` outputs the following error:

```
pkg/controller/agent/controller.go:226:112: non-constant format string in call to github.com/elastic/cloud-on-k8s/v3/pkg/utils/k8s.MaybeEmitErrorEvent
```

pkg/controller/agent/fleet.go
```diff
@@ -246,9 +246,9 @@ func maybeReconcileFleetEnrollment(params Params, result *reconciler.Results) En
 	}
 	if !reachable {
 		// we requeue if Kibana is unavailable: surface this condition to the user
-		message := "Delaying deployment of Elastic Agent in Fleet Mode as Kibana is not available yet"
+		const message = "Delaying deployment of Elastic Agent in Fleet Mode as Kibana is not available yet"
 		log.Info(message)
-		params.EventRecorder.Event(&params.Agent, corev1.EventTypeWarning, events.EventReasonDelayed, message)
+		k8s.EmitEvent(params.EventRecorder, &params.Agent, corev1.EventTypeWarning, events.EventReasonDelayed, events.EventActionAccessCheck, message)
```
dtuck9, line 251: Again because of the `non-constant format string in call`, I've opted to make `message` a constant here to avoid having to define the `"%s"` format.

test/e2e/apm/association_test.go
pkoutsovasilis: I think that statusIndex here must be 33? Either way this feels kinda fragile should we go with something like the following

```go
  re := regexp.MustCompile(`Association status changed from \[(.+?)\] to \[(.+?)\]`)

  // ...inside the loop:
  case evt.Type == corev1.EventTypeNormal && evt.Reason == events.EventAssociationStatusChange:
      establishedString := commonv1.NewSingleAssociationStatusMap(commonv1.AssociationEstablished).String()
      m := re.FindStringSubmatch(evt.Message)
      if m != nil {
          if m[1] == establishedString && m[2] != establishedString {
              assocLostEventSeen = true
          }
          if m[2] == establishedString {
              assocEstablishedEventSeen = true
          }
      }
```

pkg/controller/agent/controller.go
pkoutsovasilis: this quirk here almost makes me think that we should refactor the `k8s.MaybeEmitErrorEvent` and `k8s.EmitEvent` funcs to `k8s.MaybeEmitErrorEventf` and `k8s.EmitEventf`, respectively. Then introduce two new helper funcs to utilise for cases like this one

```go
// MaybeEmitErrorEvent emits an event if the error is report-worthy
func MaybeEmitErrorEvent(r toolsevents.EventRecorder, err error, obj runtime.Object, reason, action, message string) {
	// ignore nil errors and conflict issues
	if err == nil || apierrors.IsConflict(err) {
		return
	}

	r.Eventf(obj, nil, corev1.EventTypeWarning, reason, action, "%s", message))
}

// EmitEvent emits an event with the given parameters
func EmitEvent(r toolsevents.EventRecorder, obj runtime.Object, eventType, reason, action, message string) {
	r.Eventf(obj, nil, eventType, reason, action, "%s", message)
}
```

pkg/controller/common/annotation/association.go
```diff
@@ -1,30 +0,0 @@
...
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-
-	commonv1 "github.com/elastic/cloud-on-k8s/v3/pkg/apis/common/v1"
-)
-
-// ForAssociationStatusChange constructs the annotation map for an association status change event.
-func ForAssociationStatusChange(prevStatus, currStatus commonv1.AssociationStatusMap) (map[string]string, error) {
-	return map[string]string{
-		PrevAssocStatusAnnotation: prevStatus.String(),
-		CurrAssocStatusAnnotation: currStatus.String(),
```
pebrc, line 17: Let's clean up the constants they are not longer used IIUC

pkg/controller/common/events/events.go
```diff
@@ -42,10 +42,72 @@ const (
...
+	EventActionAssociationConfiguration = "AssociationConfiguration"
+	// EventActionElasticsearchRetrieval describes the Elasticsearch retrieval step the controller was taking when the event was triggered
+	EventActionElasticsearchRetrieval = "ElasticsearchRetrieval"
+	// EventActionAnnotateResource describes the annotation step the controller was taking when the event was triggered
+	EventActionAnnotateResource = "AnnotateResource"
+	// EventActionDeploymentReconciliation describes the deployment reconciliation step the controller was taking when the event was triggered
+	EventActionDeploymentReconciliation = "DeploymentReconciliation"
+	// EventActionAssociationReconciliation describes the association reconciliation step the controller was taking when the event was triggered
+	EventActionAssociationReconciliation = "AssociationReconciliation"
+	// EventActionAutoOpsReconciliation = "AutoOpsReconciliation"
```
pebrc: Copy/paste mistake?

pkg/controller/common/events/events.go
```diff
@@ -42,10 +42,72 @@ const (
 	EventReconciliationError = "ReconciliationError"
 )
 
+// Event actions for common controller actions
+const (
+	// EventActionValidation describes the validation step the controller was taking when the event was triggered
```
pebrc: Nit: all godocs should end with a period

pkg/controller/elasticsearch/user/user_provided.go
```diff
@@ -213,25 +213,25 @@ func realmFromBasicAuthSecret(secret corev1.Secret, existing filerealm.Realm, pa
...
+func handleSecretNotFound(log logr.Logger, recorder toolsevents.EventRecorder, es esv1.Elasticsearch, secretName string, action string) {
 	msg := "referenced secret not found"
 	// logging with info level since this may be expected if the secret is not in the cache yet
 	log.Info(msg, "namespace", es.Namespace, "es_name", es.Name, "secret_name", secretName)
-	recorder.Event(&es, corev1.EventTypeWarning, events.EventReasonUnexpected, msg+": "+secretName)
+	k8s.EmitEvent(recorder, &es, corev1.EventTypeWarning, events.EventReasonUnexpected, action, "%s: %s", msg, secretName)
 }
 
-func handleInvalidSecretData(log logr.Logger, recorder record.EventRecorder, es esv1.Elasticsearch, secretName string, err error) {
+func handleInvalidSecretData(log logr.Logger, recorder toolsevents.EventRecorder, es esv1.Elasticsearch, secretName string, err error, action string) {
```
pebrc: action is always ` events.EventActionConfiguration`?

pkg/controller/elasticsearch/user/user_provided.go
```diff
@@ -213,25 +213,25 @@ func realmFromBasicAuthSecret(secret corev1.Secret, existing filerealm.Realm, pa
 	return user.fileRealm(), nil
 }
 
-func handleSecretNotFound(log logr.Logger, recorder record.EventRecorder, es esv1.Elasticsearch, secretName string) {
+func handleSecretNotFound(log logr.Logger, recorder toolsevents.EventRecorder, es esv1.Elasticsearch, secretName string, action string) {
```
pebrc: Action is always ` events.EventActionConfiguration`? Omit the extra arg like in `handlePotentialMisconfiguration`

pebrc:
buildkite test this -f p=gke,t=Test.*Association.*

pebrc:
I think we also may have an issue with the `GetEvents` method in the e2e test. Verifying whether we need to migrate to something like:

```golang
import eventsv1 "k8s.io/api/events/v1"

  func (k *K8sClient) GetEvents(opts ...k8sclient.ListOption) ([]eventsv1.Event, error) {
      var eventList eventsv1.EventList
      if err := k.Client.List(context.Background(), &eventList, opts...); err != nil {
          return nil, err
      }
      return eventList.Items, nil
  }
```

pebrc:
Confirming the `GetEvents` issue from my earlier comment. The [e2e build](https://buildkite.com/elastic/cloud-on-k8s-operator/builds/13531) shows failures in the event-checking tests:

- `TestAPMAssociationWithNonExistentES/Non_existent_backend_should_generate_event` — timed out after 900s: `event did not fire: AssociationError`
- `TestAPMAssociationWhenReferencedESDisappears/Lost_Elasticsearch_association_should_generate_events` — timed out after 15m: `expected events did not fire: assocEstablished=false assocLost=false noBackend=false`

All event flags are `false` — the tests never see any events because `GetEvents` lists `corev1.EventList` (core/v1) while the new `events.EventRecorder` creates events via the `events.k8s.io/v1` API. These are separate resources and don't show up when listing from the other API group.

In addition to switching `GetEvents` to use `eventsv1.EventList`, the field names differ: `evt.Message` → `evt.Note`, `evt.InvolvedObject` → `evt.Regarding`. The `eventlog.go` helper would need updating too.

dtuck9:
buildkite test this -f p=gke,t=Test.Association.

dtuck9:
buildkite test this -f p=gke,t=Test.Association.

pkg/controller/agent/controller.go
dtuck9: Good call

dtuck9:
buildkite test this -f p=gke,t=Test.Association.

dtuck9:
buildkite test this -f p=gke,t=Test.Association.

dtuck9:
buildkite test this -f p=gke,t=Test.Association.

dtuck9:
buildkite test this -f p=gke,t=Test.Association.

test/e2e/apm/association_test.go
dtuck9: Fair

dtuck9:
> One thing that is missing from the PR is updating the operator RBAC manifests to include event resources with the corresponding verbs for events.k8s.io apiGroup

@pkoutsovasilis I've updated the 2 rbac that I found, could you confirm I found everything?

**ETA: :lolsob: there are a lot more than 2 files. As if this PR could get any bigger!** Never mind, I'll just see myself out.

dtuck9:
buildkite test this -f p=gke,t=Test.Association.

dtuck9:
Screenshot 2026-03-13 at 1 11 57 PM
Is this normal output when the e2e tests are successful? The pipeline is still marked as successful. Or does that actually mean none of the tests were run?

**ETA: I think I figured it out... copy/paste didn't include the wrapping `*`**

dtuck9:
buildkite test this -f p=gke,t=Test.*Association.*

dtuck9:
buildkite test this -f p=gke,t=Test.*Association.*

pkoutsovasilis (commented):
The actual code changes look good to me overall. However one note:
* because of [this](https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/leaderelection/leader_election.go#L128) I believe that we need our RBAC to allow the correct verbs for event resources for both  `events.k8s.io` and `""` (core) apiGroups - this PR replaces the latter with the former.

deploy/eck-operator/templates/_helpers.tpl
```diff
@@ -171,7 +171,6 @@ updating docs/operating-eck/eck-permissions.asciidoc file.
   - ""
   resources:
   - pods
```
barkbay, line 173: @pkoutsovasilis made a good point in https://github.com/elastic/cloud-on-k8s/pull/9195#pullrequestreview-3952110712. But I think there's also a more general upgrade concern: during a Helm upgrade, RBAC resources are applied before the operator pod is restarted. This means there's a window where the currently running operator (still using `GetEventRecorderFor` / core events API) has already lost its RBAC permission for core events (I've been bitten by this in the past 🙃 ).

While on `events` it is maybe not that problematic, to be safe, we can keep both RBAC rules for a few releases:
  - Keep events in the core "" apiGroup for backward compatibility during upgrades (and for controller-runtime leader election)
  - Add events in events.k8s.io for the new code

The core events RBAC can then be removed in a follow-up release, once we can reasonably assume most of running operators have been upgraded and it is no longer needed.

```suggestion
  - pods
  - events
```

pkg/controller/common/events/events.go
```diff
@@ -42,10 +42,72 @@ const (
...
 )
 
+// Event actions for common controller actions
+const (
+	// EventActionValidation describes the validation step the controller was taking when the event was triggered.
+	EventActionValidation = "Validation"
+	// EventActionAggregation describes the aggregation step the controller was taking when the event was triggered.
+	EventActionAggregation = "Aggregation"
+	// EventActionCertificateReconciliation describes the certificate reconciliation step the controller was taking.
+	// when the event was triggered
```
pebrc:
```suggestion
	// EventActionCertificateReconciliation describes the certificate reconciliation step the controller was taking
	// when the event was triggered.
```

deploy/eck-operator/templates/_helpers.tpl
dtuck9: I had thought about that but wasn't sure how big of a deal it would be in this context. Thank you for confirming, I'll re-add the core events to the RBAC for now.

dtuck9:
buildkite test this -f p=gke,t=Test.*Association.*

dtuck9:
buildkite test this -f p=gke,t=Test.*Association.*

barkbay:
buildkite test this -f E2E_TAGS=agent -m p=gke,p=ocp

barkbay:
### Code review

Found 1 issue:

1. `EventListOptions` uses `regarding.name` and `regarding.namespace` as field selectors, but the [Kubernetes API server only registers `involvedObject.name` and `involvedObject.namespace` as selectable fields for events](https://github.com/kubernetes/kubernetes/blob/d71ba0f013e24a3ca7a277c34c8c89639fb1f840/pkg/registry/core/event/strategy.go#L119-L126) (both core/v1 and events.k8s.io/v1 share the same field selector registration). Using `regarding.name` will cause the API server to either reject the request or silently return no results, making e2e tests that assert on events vacuously pass with empty event lists.

https://github.com/elastic/cloud-on-k8s/blob/3128c4493bf76a51dd671777f34d76c8719fcf2d/test/e2e/test/k8s_client.go#L536-L543

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

--- 

**Update**: The issue flagged in my previous comment is incorrect. `regarding.name` and `regarding.namespace` are valid field selectors for `events.k8s.io/v1`. Kubernetes registers an explicit field label conversion mapping (in pkg/apis/events/v1/conversion.go) that translates regarding.name → involvedObject.name and regarding.namespace → involvedObject.namespace before matching. The PR code is correct.

barkbay:
buildkite test this -f E2E_TAGS=agent,p=ocp

dtuck9:
Thanks @barkbay, Cursor missed those and apparently so did I.

dtuck9:
buildkite test this -f E2E_TAGS=agent,p=ocp

barkbay:
> buildkite test this -f E2E_TAGS=agent,p=ocp

@dtuck9 Something seems wrong with the OCP deployer, I think this is going to fail, but not because of your PR. I'll try to investigate tomorrow and I don't think this should prevent your PR from being merged though. (I wanted to test your PR on OCP since it's a fairly opinionated K8s distro and things that usually work on regular distros can sometimes break there.) _(edit: related issue: https://github.com/elastic/cloud-on-k8s/issues/9238)_

hack/upgrade-test-harness/testdata/v191/install.yaml
```diff
@@ -231,6 +231,18 @@ rules:
...
+    resources:
+      - events
+    verbs:
+      - get
+      - list
+      - watch
+      - create
+      - update
+      - patch
+      - delete
```
dtuck9: Am I right to interpret this as ECK version 1.9.1?

In other words, is this change actually necessary here seeing as the `events.k8s.io` api didn't exist until k8s 1.19 and ECK 1.9 only supported k8s 1.18? @elastic/cloud-k8s-operator

barkbay: > Am I right to interpret this as ECK version 1.9.1?

Yes, this is the ECK 1.9.1 manifest. The upgrade test harness tests upgrading from old versions to the current one, so the old version's manifest should reflect what was actually deployed. I think we should revert this change. `upcoming/install.yaml` is usually updated by the release manager.

dtuck9: Thank you

barkbay:
👀

pkg/controller/common/events/events.go
```diff
@@ -42,10 +42,72 @@ const (
...
+	// EventActionGetSecret describes the get secret step the controller was taking when the event was triggered.
+	EventActionGetSecret = "GetSecret"
+	// EventActionParseSecret describes the parse secret step the controller was taking when the event was triggered.
+	EventActionParseSecret = "ParseSecret"
+	// EventActionShutdown describes the shutdown step the controller was taking when the event was triggered.
+	EventActionShutdown = "Shutdown"
+	// EventActionUpscale describes the upscale step the controller was taking when the event was triggered.
+	EventActionUpscale = "Upscale"
+	// EventActionConfiguration describes the configuration step the controller was taking when the event was triggered.
+	EventActionConfiguration = "Configuration"
```
barkbay: `Configuration` feels maybe quite broad, IIUC right now it's only used for Elasticsearch file-realm user secret issues in `user_provided.go`. Something more specific like `UserConfiguration` might be more accurate?

```suggestion
	// EventActionUserConfiguration describes the configuration step the controller was taking when the event was triggered.
	EventActionUserConfiguration = "UserConfiguration"
```

pkg/controller/common/events/events.go
```diff
@@ -42,10 +42,72 @@ const (
...
+const (
+	// EventActionValidation describes the validation step the controller was taking when the event was triggered.
+	EventActionValidation = "Validation"
+	// EventActionAggregation describes the aggregation step the controller was taking when the event was triggered.
+	EventActionAggregation = "Aggregation"
+	// EventActionCertificateReconciliation describes the certificate reconciliation step the controller was taking
+	// when the event was triggered.
+	EventActionCertificateReconciliation = "CertificateReconciliation"
+	// EventActionStatusUpdate describes the status update step the controller was taking when the event was triggered.
+	EventActionStatusUpdate = "StatusUpdate"
```
barkbay: nit: the godoc says "the status update step the controller was taking" which is a bit generic. Since this is used both when health degrades and when writing the status sub-resource fails, maybe something like:

```suggestion
	// EventActionStatusUpdate is used when the resource health has changed or the controller fails to update the status sub-resource.
	EventActionStatusUpdate = "StatusUpdate"
```

pkg/controller/common/events/events.go
```diff
@@ -42,10 +42,72 @@ const (
 	EventReconciliationError = "ReconciliationError"
 )
 
+// Event actions for common controller actions
+const (
+	// EventActionValidation describes the validation step the controller was taking when the event was triggered.
+	EventActionValidation = "Validation"
+	// EventActionAggregation describes the aggregation step the controller was taking when the event was triggered.
+	EventActionAggregation = "Aggregation"
```
barkbay: Nit: I find `Aggregation` a bit hard to interpret from the event consumer's perspective. It refers to the internal `results.Aggregate()` step in the reconciler. Someone reading Action: "Aggregation" on an event might not know what operation actually failed. I'm not sure I have a good suggestion though 😐 , `Reconciliation` maybe?

```suggestion
	// EventActionReconciliation is used when the overall reconciliation loop fails after aggregating sub-results.
	EventActionReconciliation = "Reconciliation"
```

pkg/controller/common/events/events.go
```diff
@@ -42,10 +42,72 @@ const (
...
+	// EventActionRemoteClusterAssociation describes the remote cluster association step the controller was taking when the event was triggered.
+	EventActionRemoteClusterAssociation = "RemoteClusterAssociation"
+	// EventActionAssociationPreconditionCheck describes the association precondition check step the controller was taking when the event was triggered.
+	EventActionAssociationPreconditionCheck = "AssociationPreconditionCheck"
+	// EventActionConfigSecretValidation describes the config secret validation step the controller was taking when the event was triggered.
+	EventActionConfigSecretValidation = "ConfigSecretValidation"
+	// EventActionOnlineReconciliation describes the online reconciliation step the controller was taking when the event was triggered.
+	EventActionOnlineReconciliation = "OnlineReconciliation"
+	// EventActionOfflineReconciliation describes the offline reconciliation step the controller was taking when the event was triggered.
+	EventActionOfflineReconciliation = "OfflineReconciliation"
```
barkbay: Nit: "online" and "offline" are specific to the autoscaling domain here: "online" means the Elasticsearch autoscaling API is reachable, "offline" means it's not. Would something like `AutoscalingOnlineReconciliation` / `AutoscalingOfflineReconciliation` be clearer? Either way the godoc
  could be more specific:

```suggestion
	// EventActionOnlineReconciliation is used when the autoscaling controller reconciles with a reachable Elasticsearch autoscaling API.
	EventActionOnlineReconciliation = "OnlineReconciliation"
	// EventActionOfflineReconciliation is used when the autoscaling controller reconciles without a reachable Elasticsearch, falling back to minimum resource requirements.
	EventActionOfflineReconciliation = "OfflineReconciliation"
```

pkg/controller/common/events/events.go
```diff
@@ -42,10 +42,72 @@ const (
...
+	// EventActionPolicyRetrieval describes the policy retrieval step the controller was taking when the event was triggered.
+	EventActionPolicyRetrieval = "PolicyRetrieval"
+	// EventActionAssociationConfiguration describes the association configuration step the controller was taking when the event was triggered.
+	EventActionAssociationConfiguration = "AssociationConfiguration"
+	// EventActionElasticsearchRetrieval describes the Elasticsearch retrieval step the controller was taking when the event was triggered.
+	EventActionElasticsearchRetrieval = "ElasticsearchRetrieval"
+	// EventActionAnnotateResource describes the annotation step the controller was taking when the event was triggered.
+	EventActionAnnotateResource = "AnnotateResource"
+	// EventActionDeploymentReconciliation describes the deployment reconciliation step the controller was taking when the event was triggered.
+	EventActionDeploymentReconciliation = "DeploymentReconciliation"
```
barkbay, line 85: Side note, and definitely not something from your PR: `EventActionDeploymentReconciliation` is only used by the APM controller, which suggests we're not consistent in how we handle reconciliation errors across Deployment based controllers.

pkg/controller/common/events/events.go
pebrc: 💯 agreed, actually I think this is more than a nit. This will be very confusing. `results.Aggregate` by definition combines results from different parts of the code so the error could be many things and `Aggregation` has no meaning beyond the code. 👍 for something generic like "Reconciliation" it is the best we can do here.

pkg/controller/common/events/events.go
```diff
@@ -42,10 +42,72 @@ const (
...
+// Event actions for common controller actions
+const (
+	// EventActionValidation describes the validation step the controller was taking when the event was triggered.
+	EventActionValidation = "Validation"
+	// EventActionReconciliation describes the reconciliation step the controller was taking when the event was triggered.
+	EventActionReconciliation = "Reconciliation"
+	// EventActionCertificateReconciliation describes the certificate reconciliation step the controller was taking
+	// when the event was triggered.
+	EventActionCertificateReconciliation = "CertificateReconciliation"
+	// EventActionStatusUpdate s used when the resource health has changed or the controller fails to update the status sub-resource.
```
barkbay:
```suggestion
	// EventActionStatusUpdate is used when the resource health has changed or the controller fails to update the status sub-resource.
```

Issue #9186, opened 2026-03-02
Immutable Configuration Resources for Stateless Elasticsearch Rolling Updates


barkbay:
## Problem

This issue is somewhat similar to https://github.com/elastic/cloud-on-k8s/issues/8633 but is specific to stateless.

During a rolling update triggered by a configuration change, both old and new ReplicaSets should not reference the same configuration Secret (or ConfigMap). If the ECK operator updates this resource **in-place** before updating the Deployment spec, this creates a race condition window where:

1. **Replacement pods created by the old ReplicaSet** (due to OOM kill, node failure, eviction, etc.) would boot with the **new configuration**
2. The new configuration may reference volume mounts, secrets, or paths that **do not exist** in the old ReplicaSet's pod template
3. There would be **no guarantee of consistency** between a pod's spec (volumes, mounts, env vars, image) and the configuration it reads at startup

**Worst case:** Old pods could enter a crash loop because the configuration they pick up references resources that only exist in the new pod spec.

## Affected Resources

For stateless Elasticsearch deployments:

| Resource | Type | Contents |
|----------|------|----------|
| `{cluster}-es-{tier}-es-config` | Secret | `elasticsearch.yml` |
| `{cluster}-es-scripts` | ConfigMap | `prepare-fs.sh`, `pre-stop-hook-script.sh` and `readiness-port-script.sh`

## Expected Behavior

Each ReplicaSet should be permanently bound to the configuration version it was created with. Configuration changes should produce new resources with unique names, ensuring old pods always read configuration consistent with their pod spec.

## Proposed Solution

Split configuration into:

### 1. Immutable, revision-specific resources

- Name pattern: `{baseName}-{contentHash}` (e.g., `es-es-index-es-config-a1b2c3d4`)
- Created with `immutable: true` for Secrets
- Each configuration change produces a **new** resource with a unique hash
- Referenced by the Deployment pod template, so the name becomes part of the ReplicaSet spec
- Content hash is computed from SHA-256 of the rendered configuration data

### 2. Hot-reloadable resources (unchanged)

The file settings secret (`{cluster}-es-file-settings`) containing `settings.json` remains hot-reloadable and is **not affected** by this change. Elasticsearch watches this file and applies changes without requiring pod replacement.

### Naming and Labeling

**Resource name:** `{baseName}-{shortHash}`
- `shortHash` is the first 8 characters of SHA-256 hash computed over the configuration data
[...]

Issue #8799, opened 2025-08-20
Optimize memory usage by configuring the controller-runtime client's cache options.


naemono:
We should be able to minimize the memory utilization of the operator by optimizing what objects that are cached by the controller-runtime's client: https://github.com/elastic/cloud-on-k8s/blob/ddf28f918446a50eb5794f606d8df0d0e7481d5f/cmd/manager/main.go#L565-L572

We should:
1. Understand current memory usage prior to optimization.
2. Only cache object types used by the ECK operator itself.
3. Only cache objects in namespaces managed by the operator.
4. Understand memory usage after optimization.

pebrc:
## Implementation Plan: Optimize Memory Usage via Controller-Runtime Cache Options

### Context

The operator currently uses a near-default `cache.Options` in [`cmd/manager/main.go`](https://github.com/elastic/cloud-on-k8s/blob/main/cmd/manager/main.go#L576-L609). The only customization is `DefaultNamespaces` for namespace-scoped caching. All watched types and any type accessed via `mgr.GetClient().Get/List` are cached with full objects including managed fields, and deep-copied on every read.

Controller-runtime v0.23.1 provides several levers in `cache.Options` and `cache.ByObject`:

- **`DefaultTransform`** -- strip fields from objects before committing to cache
- **`DefaultUnsafeDisableDeepCopy`** -- skip deep copy on `Get`/`List` (caller must not mutate)
- **`ByObject`** -- per-GVK cache configuration (selectors, transforms, namespaces)
- **`ReaderFailOnMissingInformer`** -- prevent accidental implicit informer creation

### Proposed Changes

#### 1. Strip managed fields globally

Add `DefaultTransform: cache.TransformStripManagedFields()` to `cache.Options`. Managed fields are metadata written by server-side apply and field managers. They are substantial (often the largest part of an object's metadata) and ECK never reads them. This is the single highest-impact, lowest-risk change.

```go
opts.Cache = cache.Options{
    DefaultNamespaces: map[string]cache.Config{},
    DefaultTransform:  cache.TransformStripManagedFields(),
}
```

**File:** `cmd/manager/main.go`, around line 606.

#### 2. Enable `ReaderFailOnMissingInformer`

[...]

PR #9196, opened 2026-03-04
Add immutableconfig package for content-addressed configuration

barkbay:
This PR introduces the `immutableconfig` package - a reusable library for managing **immutable, content-addressed Kubernetes configuration resources** (Secrets and ConfigMaps).

> **Note**: This package is not used yet in ECK. It will be integrated as part of the upcoming **stateless Elasticsearch** feature, where immutable configuration helps ensure safe rolling updates across stateless deployment tiers.

## Key Features

- **Content-addressed naming**: Resources get names like `my-config-a1b2c3d4` where the suffix is derived from the content hash. This ensures old and new pod templates reference different resources during rolling updates, preventing race conditions.

- **Unified RevisionManager API**: A single `RevisionManager` object manages the full lifecycle:
  - `Reconcile()` - Creates immutable resources with owner references
  - `PatchVolumes()` - Updates pod volume references to point to the new resource
  - `GC()` - Cleans up stale resources while protecting those still in use

- **Safe garbage collection**: GC protects resources that are:
  1. Reconciled in the current cycle
  2. Referenced by existing pod templates (via `PodTemplateExtractor`)

- **Extensible pod template extraction**: The `PodTemplateExtractor` interface allows GC to work with any pod-owning controller, not just ReplicaSets. The package provides `ReplicaSetExtractor` out of the box, but custom extractors can be implemented for StatefulSets or other controllers.

- **Classification system**: `MapClassifier` categorizes config files/volumes as `Immutable` or `Dynamic`, enabling mixed configurations where some settings are content-addressed and others are hot-reloadable.

## Example Usage

```go
revisions, _ := immutableconfig.NewRevisions(client, owner, namespace).
    WithConfigResourceSelector(resourceLabels).
    WithPodTemplateSource(immutableconfig.NewReplicaSetExtractor(rsLabels)).
    Build()

secretRev := revisions.ForSecretVolumes(volumeClassifier)
name, _ := secretRev.Reconcile(ctx, &secret)
secretRev.PatchVolumes(podSpec.Volumes, name)
secretRev.GC(ctx)
```

## Files

| File | Purpose |
|------|---------|
| `revision.go` | Core `Revision` API for reconcile, patch, and GC |
| `pod_template.go` | `PodTemplateExtractor` interface and `ReplicaSetExtractor` |
| `classification.go` | `Classifier` interface and `MapClassifier` |
| `resource.go` | `BuildImmutableSecret`/`BuildImmutableConfigMap` helpers |
| `hash.go` | Content hashing utilities |
| `doc.go` | Package documentation with full example |

## Test Plan

- [x] Unit tests for all public APIs
- [x] Tests for Secret and ConfigMap reconciliation
- [x] Tests for GC with ReplicaSet protection
- [x] Tests for multi-volume scenarios
- [x] Tests for builder validation

Fixes https://github.com/elastic/cloud-on-k8s/issues/9186

Props to @simitt  for the original implementation.


prodsecmachine:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/99e863e3-5e16-4cb6-b9b8-1dc999fff0c8?refs=339d81be05f49adbb9dd6346e6ec49237774e571&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/99e863e3-5e16-4cb6-b9b8-1dc999fff0c8/license?refs=339d81be05f49adbb9dd6346e6ec49237774e571&source=prChecksComment) |

pkoutsovasilis:
buildkite test this

pkg/controller/common/immutableconfig/classification.go
```diff
@@ -0,0 +1,86 @@
...
+
+// MapClassifier is a simple map-based implementation of Classifier.
+type MapClassifier map[string]Classification
+
+// Classify returns the classification for the given filename.
+func (m MapClassifier) Classify(filename string) Classification {
+	if c, ok := m[filename]; ok {
+		return c
+	}
+	return Unclassified
```
pebrc: Nit: this could just be: 
```
	return m[filename]
``` 
as the zero value is Unclassified.

pkg/controller/common/immutableconfig/resource.go
```diff
@@ -0,0 +1,86 @@
...
+// Note: ConfigMaps don't have an immutable field in older K8s versions, but we use the same naming pattern.
+// The returned ConfigMap has:
+//   - Name: "{baseName}-{shortHash}" where shortHash is the first 8 chars of the content hash
+//   - Labels: includes ConfigTypeLabelName=immutable and ConfigHashLabelName with truncated hash
+//   - Immutable: true (for K8s 1.21+)
+func BuildImmutableConfigMap(baseName, namespace string, data map[string]string, labels map[string]string) corev1.ConfigMap {
+	fullHash := ComputeStringContentHash(data)
+	name := ImmutableName(baseName, fullHash)
+
+	cmLabels := maps.Merge(labels, map[string]string{
```
pebrc: Similar: labels are mutated?

pkg/controller/common/immutableconfig/resource.go
```diff
@@ -0,0 +1,86 @@
...
+// BuildImmutableSecret creates an immutable Secret with a content-hash suffix in its name.
+// The returned Secret has:
+//   - Name: "{baseName}-{shortHash}" where shortHash is the first 8 chars of the content hash
+//   - Labels: includes ConfigTypeLabelName=immutable and ConfigHashLabelName with truncated hash
+//   - Immutable: true
+func BuildImmutableSecret(baseName, namespace string, data map[string][]byte, labels map[string]string) corev1.Secret {
+	fullHash := ComputeContentHash(data)
+	name := ImmutableName(baseName, fullHash)
+
+	secretLabels := maps.Merge(labels, map[string]string{
```
pebrc: Mutates the original `labels`? Swap order or merge into empty map?

pkg/controller/common/immutableconfig/revision.go
```diff
@@ -0,0 +1,356 @@
...
+	},
+	newList: func() client.ObjectList { return &corev1.ConfigMapList{} },
+	newObj:  func() client.Object { return &corev1.ConfigMap{} },
+	kind:    "ConfigMap",
+}
+
+// Revision manages the lifecycle of immutable content-addressed resources (Secrets or ConfigMaps):
+// create-only reconciliation with owner references, volume patching, and
+// garbage collection that protects resources referenced by existing pod templates.
+type Revision struct {
```
pebrc: I find the naming here slightly misleading as this does not represent a single revision. I am struggling to come up with a good name though,  but it feels more like a `RevisionManager` to me. Another naming option would be rename `resourceOps` to `resourceAdapter` (which also matches better its function IIUC) and the we have the `Ops` name available for the exported structs: `immutableconfig.Ops` 
```golang
secretOps := revisions.ForSecretVolumes(classifier)
name, _ := secretOps.Reconcile(ctx, &secret)
secretOps.PatchVolumes(podSpec.Volumes, name)
secretOps.GC(ctx)
```
This reads reasonable well, but the we would have to rename everyting: `NewOps`, `Revisions` would have to become something like `OpsFactory` which is very awkward.

So maybe RevisionManager is slightly better.

pkg/controller/common/immutableconfig/revision.go
```diff
@@ -0,0 +1,356 @@
...
+	}
+	if !apierrors.IsNotFound(err) {
+		return "", err
+	}
+	if err := r.client.Create(ctx, obj); err != nil && !apierrors.IsAlreadyExists(err) {
+		return "", err
+	}
+	r.reconciled.Insert(obj.GetName())
+	return obj.GetName(), nil
+}
```
pkoutsovasilis: question: I can see that BuildImmutableConfigMap and BuildImmutableSecret use the Immutable field of Kubernetes, but this immutability only protects data. Object metadata (labels, annotations, ownerReferences) can still be mutated. Now Build() uses ConfigTypeLabelName=immutable to set the resource selector used by GC. If this label is removed from a resource, GC won't find it. Should Reconcile() verify and restore this label, or is this the caller's responsibility?

barkbay: You're right that Kubernetes `Immutable` only protects `data`, not metadata. However, I think the current design is sufficient:

* `Reconcile()` sets owner references via `SetControllerReference()`, so even if a resource becomes invisible to GC, Kubernetes will clean it up when the owner is deleted. Functionality wouldn't break.

* The operator generally trusts labels to remain stable on resources it manages. This assumption will become even more fundamental as we implement cache optimizations (#8799) that rely on label selectors. Adding defensive label restoration here would be inconsistent with that trust model.

Given that label tampering would require deliberate action on operator-managed resources, I'd prefer to keep the current design rather than add complexity for this edge case.

pkg/controller/common/immutableconfig/resource.go
barkbay: Good catch 👍

pkg/controller/common/immutableconfig/revision.go
barkbay: > So maybe RevisionManager is slightly better.

Renamed in https://github.com/elastic/cloud-on-k8s/pull/9196/changes/2fd566fbde17e56d67d723ce08e38e06d9955a9e

PR #9197, opened 2026-03-04
Detect stale CA in certificate chain and trigger certificates reissuance

### LINKED ISSUE #9198: Certificates not reissued when CA rotates with same private key
### Description

When ECK renews a CA certificate using the same private key, leaf certificates signed by the old CA are not detected as needing reissuance. This occurs because the certificate reconciliation logic relies on x509.Verify(), which validates the cryptographic signature but does not detect that the CA certificate embedded in the chain has changed.

### Impact
After CA rotation, the certificate chain contains the old CA certificate while the cluster's trust store contains the new CA certificate. This mismatch causes TLS handshake failures with stricter validators (e.g., Java's PKIX):
```
  PKIX path validation failed: Path does not chain with any of the trust anchors
```

This can lead to:
  - Elasticsearch nodes unable to rejoin the cluster
  - Deadlock scenarios when combined with rolling upgrade constraints

### Root Cause
- `shouldIssueNewCertificate()` (transport certs) and `getHTTPCertificate()` (HTTP certs) only validate signatures via `x509.Verify()`
- When CA is renewed with the same private key, signature validation passes
- The CA certificate embedded in the PEM chain becomes stale but is not detected

### Fix
Update certificate reconciliation to extract both the leaf and CA certificates from the chain, and trigger reissuance when the CA certificate in the chain differs from the current CA.

pkoutsovasilis:
## Overview
This PR updates certificate reconciliation to detect CA rotations where the CA is reissued with the same private key. In this scenario, `x509.Verify()` alone still succeeds (valid signature), but the CA certificate embedded in the chain becomes stale.

### Changes
  - **Transport cert reconciliation**: Update `shouldIssueNewCertificate()` to extract both leaf and CA certs from the chain and trigger reissuance when the CA cert in the chain differs from the current CA.
  - **HTTP cert reconciliation**: Update `getHTTPCertificate()` to invalidate and reissue when the CA cert in the chain differs from the current CA.


prodsecmachine:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/b33b8fec-5122-44a1-aa4a-62783cd8ac4a?refs=1534b599237e3516da3979a0f28147a55446475e&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/b33b8fec-5122-44a1-aa4a-62783cd8ac4a/license?refs=1534b599237e3516da3979a0f28147a55446475e&source=prChecksComment) |

pkg/controller/common/certificates/http_reconcile.go
```diff
@@ -310,18 +309,31 @@ func getHTTPCertificate(
...
 		return nil
 	}
 
+	if !caCertInChain.Equal(ca.Cert) {
+		log.Info("CA certificate in chain does not match current CA, should issue new certificate",
+			"namespace", secret.Namespace,
+			"secret_name", secret.Name,
+		)
+		return nil
+	}
```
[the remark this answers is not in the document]
pkoutsovasilis:
```go
func (c *Certificate) Equal(other *Certificate) bool {
	if c == nil || other == nil {
		return c == other
	}
	return bytes.Equal(c.Raw, other.Raw)
}
```

so nope if caCertInChain is nil it won't panic

pkg/controller/elasticsearch/certificates/transport/pod_secret.go
```diff
@@ -123,11 +124,19 @@ func shouldIssueNewCertificate(
...
 		return true
 	}
 
+	if !caCertInChain.Equal(ca.Cert) {
+		log.Info("CA certificate in chain does not match current CA, should issue new certificate",
+			"namespace", pod.Namespace,
+			"pod_name", pod.Name,
+		)
+		return true
+	}
```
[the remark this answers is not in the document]
pkoutsovasilis:
```go
func (c *Certificate) Equal(other *Certificate) bool {
	if c == nil || other == nil {
		return c == other
	}
	return bytes.Equal(c.Raw, other.Raw)
}
```

so nope if caCertInChain is nil it won't panic

pebrc:
Thanks for the fix — the root cause analysis (Go 1.25 changing SKI computation from SHA-1 to truncated SHA-256 breaking Java's AKI→SKI PKIX validation) is clear and the approach is solid.

One thing I wanted to raise: have you considered pinning the Subject Key Identifier (SKI) during CA renewal instead of (or in addition to) triggering full leaf cert reissuance?

In `NewSelfSignedCA`, the certificate template doesn't set `SubjectKeyId`, so Go computes it automatically — which is where the cross-version breakage comes from. But `x509.CreateCertificate` uses `SubjectKeyId` as-is if it's set on the template. So `renewCAFromExisting` could carry forward the old CA's SKI:

```go
// In CABuilderOptions:
SubjectKeyId []byte

// In NewSelfSignedCA, on the template:
SubjectKeyId: options.SubjectKeyId,

// In renewCAFromExisting:
SubjectKeyId: oldCA.Cert.SubjectKeyId,
```

This would keep the SKI stable across Go versions, so existing leaf certs' AKI→SKI chains remain valid and no reissuance is needed. You would obviously also have to verify that SKI has not changed when checking certificates for validity.

The reason I bring this up is the transient disruption window during hot-reload. With the current PR, when the CA rotates and all leaf certs are reissued simultaneously, there's a window (up to kubelet sync interval, ~1 minute) where some pods have reloaded the new certs and others haven't. During that window, pods that have reloaded (new leaf cert with AKI→new_SKI, trusting new CA) and pods that haven't (old leaf cert with AKI→old_SKI, trusting old CA) cannot validate each other's certs in either direction — both A→B and B→A fail Java's PKIX check. With the byte-swap approach at least one direction still worked during that window.

Pinning the SKI avoids this entirely since all existing certs remain valid throughout the transition.

The tradeoff is that the byte-swap leaves a technically imprecise state (leaf signed by old CA, chain contains new CA), but that state is valid from both Go's and Java's perspective as long as AKI→SKI chains correctly and the key hasn't changed — which SKI pinning guarantees.

Thoughts?

pkoutsovasilis:
Thanks @pebrc, these are good points to consider.

## On SKI pinning
I looked at the code and `renewCAFromExisting` only applies to ECK-managed self-signed CAs. For **custom CA secrets** (provided via `spec.transport.tls.certificate` or `spec.http.tls.certificate`), ECK doesn't manage rotation at all - it just uses whatever CA the user provides. If a user rotates their custom CA (even with the same private key - although unlikely), we have no control over the SKI. The fix of this PR handles both scenarios uniformly: whether the CA changed because ECK rotated it, or because the user updated their custom CA secret, we detect the mismatch and reissue. AFAICT SKI pinning would only address the ECK-managed case.

## On relying on validator behaviour
This is my main concern with the byte-swap approach. We'd be leaving a state where the leaf cert's issuer (old CA) differs from the CA cert in the chain (new CA), relying on validators to only check cryptographic signatures rather than certificate identity. While the SKI pinning would work for Go and Java today, should we bet on it?! (maybe I am being too defensive here)

## On the transient disruption window
This is a valid concern only SKI pinning solves it

All the above said, maybe a solution is a hybrid approach? SKI pinning for ECK-generated certificate **in addition to** the CA matching checks for user provided secrets. Thoughts?

pebrc:
> While the SKI pinning would work for Go and Java today, should we bet on it?! (maybe I am being too defensive here)

I think your concern is fair but AKI/SKI chaining is defined in the RFC so it is not an implemenation quirk but something we can build upon? I would expect any conformant PKIX implementation to accept it if the identifiers match.

So in my mind the minimal fix is maybe the extended validation with AKI/SKI comparison. 

SKI pinning is not strictly necessary as the extended validation would already cover the bug (but maybe cheap IIUC it's 3 loc ), wdyt? 

_If we want to keep the full comparison we should probably remove the old byte-swap approach completely._

pkoutsovasilis:
Thanks @pebrc , I've updated the PR to incorporate your suggestions:

1. SKI Pinning - renewCAFromExisting now passes the existing CA's SubjectKeyId to preserve SKI stability during ECK-managed CA renewals
2. SKI Comparison - Added CAMatch function that compares Subject Key Identifiers to validate certificate chains. Falls back to full certificate equality if SKI is not available
3. Removed byte-swap - Removed the code that updated the CA in the certificate chain without reissuing the leaf certificate

AFAICT this approach:
* Avoids the transient disruption window for ECK-managed CA rotations (SKI remains stable)
* Triggers reissuance only when SKI doesn't match (custom CA rotation where SKI changed)
* Applies consistently to both transport and HTTP certificates

wdyt?

pkg/controller/elasticsearch/certificates/transport/pod_secret.go
```diff
@@ -194,39 +203,46 @@ func shouldIssueNewCertificate(
...
+		switch {
+		case leafCert != nil && caCert != nil:
+			return leafCert, caCert
+		case leafCert == nil && c.Subject.CommonName == commonName:
+			leafCert = c
+		case caCert == nil && c.IsCA:
+			caCert = c
 		}
 		names = append(names, c.Subject.CommonName)
 	}
```
moukoublen, line 232: [ this is for both functions ] 

Both extractTransportCertAndCA (transport) and getHTTPCertificate (HTTP) assume the TLS secret contains a chain that matches how we build it today: leaf first, then signer CA. That’s how we create them (e.g. EncodePEMCert(certificate, ca.Cert.Raw) in http_reconcile.go and EncodePEMCert(certData, ca.Cert.Raw) in pod_secret.go). 

I am wondering whether it makes sense to not blindly have these two assumptions (~~specific order~~, CA always present, no other CA in chain), and to add a check in both that the chain has the expected shape (e.g. two certs, one signs the other) and treat unexpected shapes as invalid.

Or perform 2 loops, one to pick `c.Subject.CommonName == commonName` (~~to not assume that the leaf will always be first~~) and then `leafCert.CheckSignatureFrom(c) == nil `  to get the direct signer of the certificate.

Perhaps it is not worth it, but I feel that those two assumptions, running in parallel, could lead to issues in the future if something changes in how we create and store the TLS secret.

pebrc:
One concern with the `CAMatch` check: it compares the **CA cert in the chain's SKI** against the **current CA's SKI**. This answers "was the CA in the chain updated?" but not "is the leaf cert compatible with the current CA?"

These diverge in the pre-existing byte-swapped state — i.e. clusters already affected by the original bug:

1. Old ECK code (pre-PR) ran on Go 1.25, CA was renewed, byte-swap executed
2. State is now: `tls.crt` = [leaf (`AKI=old_SKI`), new_CA (`SKI=new_SKI`)], `ca.crt` = [new_CA]
3. Java PKIX is already broken (leaf's AKI doesn't match trust anchor's SKI)

When the PR is deployed on this cluster:

- `caCertInChain` = new_CA (placed there by the old byte-swap)
- `CAMatch(new_CA, new_CA)` → `true` (literally the same cert)
- `x509.Verify` passes (Go is lenient)
- **No reissuance** → the broken state persists until the leaf cert naturally expires

The fix would be to check the **leaf cert's `AuthorityKeyId`** against the **current CA's `SubjectKeyId`** instead of (or in addition to) comparing the two CA certs:

```go
if len(certificate.AuthorityKeyId) > 0 && len(ca.Cert.SubjectKeyId) > 0 {
    if !bytes.Equal(certificate.AuthorityKeyId, ca.Cert.SubjectKeyId) {
        // AKI→SKI chain is broken, leaf was signed by a CA with a different SKI
        return nil
    }
}
```

This directly checks whether the leaf cert is compatible with the current CA. In the byte-swapped state, `certificate.AuthorityKeyId` (old_SKI) ≠ `ca.Cert.SubjectKeyId` (new_SKI) → reissuance triggered → broken state remediated immediately.

The `CAMatch` check could still be useful as a secondary safeguard (e.g. detecting when the chain CA cert itself is stale before it gets byte-swapped), but the AKI→SKI check on the leaf cert is the one that catches the actual incompatibility that breaks Java's PKIX validation.

So the recommendation would be: **AKI→SKI check on the leaf cert** (primary detection) + **SKI pinning** (prevention for future rotations). The `CAMatch` CA-to-CA comparison can be dropped or kept as defense-in-depth — it doesn't hurt, but it doesn't catch the critical case on its own.

pkoutsovasilis:
Thanks for catching that @pebrc - I had zoned out and was clearly focusing only on comparing the two CAs rather than what actually matters: whether the leaf cert is compatible with the current CA. Back on track now.

I've updated the implementation to follow your recommendation:
 * Primary detection (AKI→SKI check on leaf cert)
 * Replaced CAMatch with CertIsSignedByCA(leafCert, currentCA) which:
   1. Verifies the cryptographic signature via CheckSignatureFrom()
   2. Validates that leafCert.AuthorityKeyId matches currentCA.SubjectKeyId

This catches the pre-existing byte-swapped state immediately - leaf's AKI (old_SKI) ≠ current CA's SKI (new_SKI) → reissuance triggered. 

* Prevention (SKI pinning): renewCAFromExisting() continues to pass the existing CA's SubjectKeyId to keep SKI stable across ECK-managed rotations.
* Removed:
  * The CA-to-CA comparison (CAMatch)
  * CA extraction from the certificate chain - we now validate the leaf directly against the current CA

IMO the last approach is cleaner and directly addresses the incompatibility that breaks PKIX validation, wdyt?

pkg/controller/elasticsearch/certificates/transport/pod_secret.go
pkoutsovasilis: After a little back and forth (look below) the current PR took an alternative path forward so the previous assumptions no longer apply - we no longer extract the CA from the chain at all. So I am gonna resolve this now

pkoutsovasilis:
buildkite test this -f p=gke,t=TestCustom.*CA.* -m s=9.3.0,s=8.19.0

barkbay (commented):
I didn’t spend a lot of time reviewing this PR end-to-end, so I may be off here, so bear with me 🙂

While using an LLM to better understand the change set and edge cases, I ended up writing this test:

```go
func Test_ensureInternalSelfSignedCertificateSecretContents_UpdatesCAChainWhenLeafIsReused(t *testing.T) {
	// Build a rotated CA that keeps the same private key + same SKI.
	// In this setup the existing leaf can still be considered valid by AKI->SKI checks,
	// so reconcile may decide to reuse the leaf certificate bytes.
	rotatedCA, err := NewSelfSignedCA(CABuilderOptions{
		Subject:      pkix.Name{CommonName: "test-common-name"},
		PrivateKey:   testRSAPrivateKey,
		SubjectKeyID: testCA.Cert.SubjectKeyId,
	})
	require.NoError(t, err)
	require.False(t, testCA.Cert.Equal(rotatedCA.Cert), "rotated CA should be a different certificate")
	assert.Equal(t, testCA.Cert.SubjectKeyId, rotatedCA.Cert.SubjectKeyId, "rotated CA should preserve SKI")

	privateKey, err := EncodePEMPrivateKey(testRSAPrivateKey)
	require.NoError(t, err)

	// Seed the secret with OLD CA material:
	// - ca.crt points to old CA
	// - tls.crt contains a leaf signed by old CA and old CA in chain
	secret := &corev1.Secret{
			Name:      InternalCertsSecretName(esv1.ESNamer, testES.Name),
			Namespace: testES.Namespace,
		},
		Data: map[string][]byte{
			KeyFileName:  privateKey,
			CAFileName:   EncodePEMCert(testCA.Cert.Raw),
			CertFileName: pemTLS, // leaf signed by testCA with testCA in chain
		},
	}

	// Reconcile against the rotated CA.
	// Expected behavior:
	// - leaf may be reused (same key + same SKI path is still valid),
	// - but CA material in secret should still be refreshed to rotated CA.
	changed, err := ensureInternalSelfSignedCertificateSecretContents(
		context.Background(),
		secret,
		k8s.ExtractNamespacedName(&testES),
		esv1.ESNamer,
		testES.Spec.HTTP.TLS,
		nil,
		[]corev1.Service{testSvc},
		rotatedCA,
		RotationParams{
			Validity:     DefaultCertValidity,
			RotateBefore: DefaultRotateBefore,
		},
	)
	require.NoError(t, err)

	// ca.crt and tls.crt chain should be rewritten with the current CA.
	assert.True(t, changed, "CA data should be refreshed even when leaf certificate is reused")
	assert.Equal(t, EncodePEMCert(rotatedCA.Cert.Raw), secret.Data[CAFileName], "ca.crt should be updated to rotated CA")

	chain, err := ParsePEMCerts(secret.Data[CertFileName])
	require.NoError(t, err)
	require.GreaterOrEqual(t, len(chain), 2, "tls.crt should include leaf + CA chain")
	assert.True(t, chain[1].Equal(rotatedCA.Cert), "tls.crt chain should include rotated CA")
}
```

With current PR code, this test fails. Is that expected behaviour, or does it point to a gap in the "same SKI" rotation path?

pkg/controller/common/certificates/http_reconcile.go
```diff
@@ -254,21 +255,6 @@ func ensureInternalSelfSignedCertificateSecretContents(
 		secret.Data[CertFileName] = EncodePEMCert(certificate, ca.Cert.Raw)
 	}
 
-	// Ensure that the CA certificate is up-to-date.
-	expectedCaPem := EncodePEMCert(ca.Cert.Raw)
```
pebrc, line 258: I think this code needs to stay. I think I was maybe expressing myself not clearly in an earlier comment. What I meant was either we try to be smart and do the extended SKI/AKI validation OR we remove all the cert swapping and always renew all certs. I think we should try to do the smart thing.

pebrc: With the code restored @barkbay 's test passes.

pkoutsovasilis: changed in bde2b839e8370a2e43f2164fb6dd1a063cd445de

pkoutsovasilis:
thanks for catching that @barkbay! I re-introduced the byte-swapping and added your test in bde2b839e8370a2e43f2164fb6dd1a063cd445de which is now passes. 

Evidently the boundaries of re-issuing or just updating the chain were blurry in my head

pkg/controller/common/certificates/ca.go
```diff
@@ -55,6 +56,11 @@ type CABuilderOptions struct {
 	PrivateKey *rsa.PrivateKey
 	// ExpireIn defines in how much time will the CA expire (defaults to DefaultCertValidity if not provided).
 	ExpireIn *time.Duration
+	// SubjectKeyID to use for the CA certificate. If provided, this value is used as-is instead of being
+	// auto-generated by Go's x509.CreateCertificate. This is useful when renewing a CA with the same private
```
pebrc:
```suggestion
	// auto-generated by Go's x509.CreateCertificate. This is useful when renewing a CA with the same private key
```

pkg/controller/common/certificates/ca_reconcile_test.go
```diff
@@ -422,3 +492,164 @@ func Test_buildCAFromSecret(t *testing.T) {
...
+				// Mock a CA without SKI but with same public key for signature verification
+				PublicKey:               ca1.Cert.PublicKey,
+				SubjectKeyId:            nil, // No SKI
+				RawTBSCertificate:       ca1.Cert.RawTBSCertificate,
+				Signature:               ca1.Cert.Signature,
+				SignatureAlgorithm:      ca1.Cert.SignatureAlgorithm,
+				PublicKeyAlgorithm:      ca1.Cert.PublicKeyAlgorithm,
+				Raw:                     ca1.Cert.Raw,
+				RawSubjectPublicKeyInfo: ca1.Cert.RawSubjectPublicKeyInfo,
+			},
```
pebrc, line 637: This mock x509.Certificate is missing IsCA: true, BasicConstraintsValid: true, KeyUsage, and Version. In Go 1.20+, CheckSignatureFrom checks parent.IsCA and will fail on this mock, so the test passes for the wrong reason (signature check fails due to malformed mock rather than AKI→SKI mismatch). The test assertion is still correct, but it's not testing what the comment says. To properly test the AKI/SKI path, the mock would need at minimum IsCA: true, BasicConstraintsValid: true, Version: 3, and KeyUsage: x509.KeyUsageCertSign.

pkg/controller/common/certificates/ca.go
```diff
@@ -145,3 +152,40 @@ func PublicCertsHasCACert(client k8s.Client, namer name.Namer, namespace string,
 	_, ok := certsSecret.Data[CAFileName]
 	return ok, nil
 }
+
+// CertIsSignedByCA checks if the leaf certificate was signed by the current CA.
+// It validates both the cryptographic signature and the AKI→SKI chain relationship
+// (required by PKIX validation, e.g., Java).
```
pebrc: Maybe add more color as to why it is necessary: 
```
// Go's x509.Verify() does not use AKI/SKI during chain resolution, but Java's
// PKIX path builder uses AKI as a hard filter: trust anchors
// whose SKI does not match the leaf's AKI are discarded, causing path building
// to fail even when the signature is valid. ECK must detect these mismatches
// since Elasticsearch delegates TLS to the JDK's built-in PKIX implementation.
```
Or a slightly shorter version of that

pkg/controller/common/certificates/http_reconcile.go
```diff
@@ -322,6 +324,16 @@ func getHTTPCertificate(
 		return nil
 	}
 
+	if !CertIsSignedByCA(certificate, ca.Cert) {
+		log.Info("Certificate AKI does not match current CA SKI (AKI→SKI chain broken), should issue new certificate",
```
pebrc: CertIsSigned does more than AKI->SKI path validation. It also checks the cryptographic signature: Something more general like "Certificate is not valid for current CA, should issue new certificate" would be more appropriate?

pkg/controller/elasticsearch/certificates/transport/pod_secret.go
```diff
@@ -128,6 +129,16 @@ func shouldIssueNewCertificate(
 		return true
 	}
 
+	if !certificates.CertIsSignedByCA(cert, ca.Cert) {
+		log.Info("Certificate AKI does not match current CA SKI (AKI→SKI chain broken), should issue new certificate",
```
pebrc: Same here: log is too specific. See previous comment. I would keept the SKI/AKI k/v for debugging purposes though.

test/e2e/es/certs_test.go
```diff
@@ -428,6 +429,153 @@ func TestCustomTransportCA(t *testing.T) {
 	test.RunMutations(t, []test.Builder{withBuiltinCA}, []test.Builder{withBogusCert, withCustomCert, withBuiltinCA})
 }
 
+// TestCustomTransportCARotationWithSameSKI tests that when a user rotates their custom transport CA
+// while keeping the same private key and Subject Key Identifier (SKI), the cluster remains healthy
+// without requiring reissuance of all transport certificates. This is the expected behavior because:
+// 1. The cryptographic signature remains valid (same private key)
+// 2. The SKI/AKI chain validation passes (same SKI)
+// 3. ECK's CAMatch function recognizes that the SKI matches and doesn't trigger unnecessary reissuance
```
pebrc:
```suggestion
// 3. ECK's CertIsSignedByCA function recognizes that the SKI matches and doesn't trigger unnecessary reissuance
```

?

pkg/controller/common/certificates/ca_reconcile_test.go
```diff
@@ -422,3 +492,164 @@ func Test_buildCAFromSecret(t *testing.T) {
...
+		NotAfter:              time.Now().Add(24 * time.Hour),
+		PublicKeyAlgorithm:    x509.RSA,
+		SignatureAlgorithm:    x509.SHA256WithRSA,
+		SerialNumber:          big.NewInt(2),
+		KeyUsage:              x509.KeyUsageDigitalSignature,
+		BasicConstraintsValid: true,
+		IsCA:                  false,
+		// AuthorityKeyId intentionally not set
+	}
+	leafNoAKIDER, err := x509.CreateCertificate(cryptorand.Reader, &leafNoAKITemplate, &leafNoAKITemplate, leafKey.Public(), leafKey)
```
pebrc: I think the intent here is to test the broken validation path AKI->SKI, but you signing the leaf with the leaf key not with the ca key. So when CertIsSignedByCA(leafNoAKI, ca1.Cert) is called, CheckSignatureFrom(ca1.Cert) fails immediately (wrong key) — the function never reaches the AKI check at lines 187-188. The test passes and returns false, but for the wrong reason (signature failure, not missing AKI). To properly test the "no AKI" path, the leaf would need to be signed by a key matching ca1 but without AKI set — which is tricky because Go's x509.CreateCertificate auto-populates AKI from the parent's SKI.

pkg/controller/common/certificates/ca_reconcile.go
```diff
@@ -117,23 +115,32 @@ func renewCAFromExisting(
 	meta metadata.Metadata,
 	expireIn time.Duration,
 	caType CAType,
-	signer crypto.Signer,
+	existingCA *CA,
 ) (*CA, error) {
 	log := ulog.FromContext(ctx)
-	privateKey, ok := signer.(*rsa.PrivateKey)
+	if existingCA == nil || existingCA.Cert == nil {
```
pebrc: Nit: This seems redundant? The only call site already ensures this is not nil? Maybe I am missing something. But we can keep it as an extra defensive measure even if it is effectively dead code right now.

pebrc (commented):
LGTM just a few nits and testing improvements.

pkg/controller/common/certificates/ca_reconcile.go
pkoutsovasilis: changed in 0f0c3df19defff0d2dc40764debf2e0922533a4a

pkg/controller/common/certificates/ca_reconcile_test.go
pkoutsovasilis: fixed in ea54c9fbaed3efd82b720ccbd94987d9395b7422

test/e2e/es/certs_test.go
pkoutsovasilis: changed in 12e659e5c3e55f8b16a87ece3ba85cb5ec1f5ce5

pkg/controller/elasticsearch/certificates/transport/pod_secret.go
pkoutsovasilis: fixed in bfe01e69fbd62e236c261f428d0eefce072126eb

pkg/controller/common/certificates/http_reconcile.go
pkoutsovasilis: fixed in bfe01e69fbd62e236c261f428d0eefce072126eb

pkg/controller/common/certificates/ca.go
pkoutsovasilis: fixed in 12e659e5c3e55f8b16a87ece3ba85cb5ec1f5ce5

pkg/controller/common/certificates/ca_reconcile_test.go
pkoutsovasilis: fixed in ea54c9fbaed3efd82b720ccbd94987d9395b7422

pkg/controller/common/certificates/ca.go
pkoutsovasilis: fixed in 12e659e5c3e55f8b16a87ece3ba85cb5ec1f5ce5

pkg/controller/common/certificates/ca_reconcile.go
pebrc: Sorry I think my comment was misleading I wanted to say: keep it. Now the godoc is out of date in stating that you can pass in nil.

pebrc (approved):
LGTM

pkg/controller/common/certificates/ca_reconcile.go
pkoutsovasilis: re-added it here 1534b599237e3516da3979a0f28147a55446475e

Issue #8744, opened 2025-07-14
Empty volumeClaimTemplates still create volume claims


Breee:
We want to deploy an elasticsearch instance  

```
apiVersion: elasticsearch.k8s.elastic.co/v1
kind: Elasticsearch
metadata:
  name: elasticsearch
  namespace: test
spec:
  version: 8.18.3
  nodeSets:
  - config:
      node.store.allow_mmap: false
    count: 3
    name: default
    volumeClaimTemplates: []
    podTemplate:
      spec:
        containers:
        - name: elasticsearch
          resources:
            limits:
              cpu: "2"
              memory: 8Gi
          volumeMounts:
          - mountPath: /usr/share/elasticsearch/data
            name: elasticsearch-data
            subPathExpr: $(POD_NAME)
        volumes:
        - name: elasticsearch-data
          persistentVolumeClaim:
            claimName: test-elasticsearch-data
```

We bring our own PVCs and PVs due to the nature of our storage.  

The `volumeClaimTemplates: []` should lead to a statefulset that does not have any volumeClaims,  however, the operator will generate the default one per replica anyway,  ` elasticsearch-data-elasticsearch-es-default-0`,  ` elasticsearch-data-elasticsearch-es-default-1`, ` elasticsearch-data-elasticsearch-es-default-2`

pebrc:
Thank you for reporting this. We've confirmed the behavior and understand the issue.

When `volumeClaimTemplates` is empty or omitted, the operator adds a default `elasticsearch-data` volume claim template. Currently, only non-PVC volumes (like `emptyDir`) with the same name in `podTemplate.spec.volumes` suppress this default. A PVC volume reference — like the one you're using to bring your own pre-existing claim — does not suppress it, resulting in the unwanted extra PVCs being created.

We plan to fix this by treating any volume in `podTemplate.spec.volumes` with the matching name as a signal to suppress the default, regardless of volume type. This aligns with how our Logstash controller already handles the same scenario. In the meantime, as a workaround, you can use an `emptyDir` volume named `elasticsearch-data` in your pod template to suppress the default VCT, and then separately mount your pre-existing PVC under a different volume name — though we acknowledge that's not ideal.

PR #9199, opened 2026-03-05
elasticsearch controller: skip default PVC if volume with same name exists

moukoublen:
AppendDefaultPVCs used to skip default PVCs only for non-PVC volumes (emptyDir, hostPath, etc.). It now skips whenever any volume with the same name exists in the pod spec, including user-provided PVCs, so we no longer add redundant default PVCs when the user supplies their own claim. Logic is simplified to a single set of volume names.

fixes: https://github.com/elastic/cloud-on-k8s/issues/8744

____

## Summary

This change improves how the Elasticsearch controller applies default PersistentVolumeClaims (PVCs) when building Elasticsearch StatefulSets. The operator now skips adding a default volume claim template whenever a volume with the same name already exists in the pod spec, regardless of that volume’s type. This prevents redundant or conflicting volume claim templates when users supply their own PVCs (or other volume types) for the same volume name.

---

## Previous behavior

When no user-defined `volumeClaimTemplates` were specified for an Elasticsearch node set, the operator appended default volume claim templates (for example, for the default data volume). The operator did **not** add a default PVC only when the pod spec contained a **non-PVC** volume (such as `emptyDir` or `hostPath`) with the same name.

If a user defined a **PersistentVolumeClaim** volume with the same name as a default (e.g. by referencing an existing PVC in `podTemplate.spec.volumes`), the operator still appended the corresponding default volume claim template. That could result in:

- Redundant volume claim templates for the same logical volume name  
- Confusion or conflicts when the user’s intent was to use their own PVC for that volume

---

## New behavior

The default logic is unified: the operator skips appending a default PVC with that name whenever **any** volume with that name is already present in the pod spec. It no longer distinguishes between PVC and non-PVC volumes for this decision.

Specifically:

- If `podTemplate.spec.volumes` contains a volume whose name matches a default volume claim template (e.g. the default data volume name `elasticsearch-data`), the operator **does not** add that default PVC to `volumeClaimTemplates`, whether the existing volume is:
  - a user-provided PersistentVolumeClaim  
  - an `emptyDir`  
  - a `hostPath`  
  - or any other volume type  

- Only default PVCs whose names do **not** collide with an existing volume name are appended. If the user has defined no `volumeClaimTemplates` and has provided volumes for some (but not all) default names, only the remaining defaults are added.

- If the user has defined one or more `volumeClaimTemplates` explicitly, the operator does not add any default PVCs (this behavior is unchanged).

---

## Impact for users

- **Users who supply their own PVC volumes** (e.g., pre-provisioned or custom storage) in `podTemplate.spec.volumes`, for the same volume name as a default will no longer get an extra default volume claim template. The operator respects the existing volume and does not add a duplicate default PVC for that name.

- **Users who use non-PVC volumes** (e.g. `emptyDir`, `hostPath`) for a default volume name continue to see the same behavior: no default PVC is added for that name.

- **Behavior is consistent**: "a volume with this name already exists" always means "do not add the default PVC for this name", regardless of volume type. This makes the operator's behavior easier to reason about and avoids redundant volume claim templates.

---

## Breaking impact on ECK upgrade

If a user defined empty `volumeClaimTemplates` and provided custom PVC volumes in `podTemplate.spec.volumes` with the same names as the default volumes (for example elasticsearch-data), those custom volumes were not used and the default volumes were provisioned instead.

After upgrading, the operator will attempt to provision Elasticsearch nodes using the user-provided PVC volumes rather than the default ones, which results in the following warning being displayed:

```
Failed to apply spec change: while reconciling resources: reconcile StatefulSet: StatefulSet.apps "elasticsearch-c1-es-default" is invalid: spec: Forbidden: updates to statefulset spec for fields other than 'replicas', 'ordinals', 'template', 'updateStrategy', 'revisionHistoryLimit', 'persistentVolumeClaimRetentionPolicy' and 'minReadySeconds' are forbidden
```

A workaround is to remove the custom PVCs defined in `podTemplate.spec.volumes` that overlap with the default volume names. Since these volumes are not actually used, the default volumes can remain in place without issue.

**The likelihood that a user configured this scenario is extremely low.**

---

## Affected components

- **Elasticsearch**: Default volume claim template handling in the Elasticsearch controller (StatefulSet construction). The logic lives in the shared defaults package used when building node set specs.

No changes are required to the Elasticsearch or ElasticsearchNodeSet resource APIs; this is a behavioral refinement of existing defaulting logic.

### DIFF
--- a/pkg/controller/common/defaults/pvc_test.go
+++ b/pkg/controller/common/defaults/pvc_test.go
@@ -79,6 +79,26 @@ func TestAppendDefaultPVCs(t *testing.T) {
 			want: nil,
 		},
 		{
+			name: "skip only the default pvc whose name collides with a user-provided pvc volume",
+			args: args{
+				existing: nil,
+				podSpec: v1.PodSpec{
+					Volumes: []v1.Volume{
+						{
+							Name: foo.Name,
+							VolumeSource: v1.VolumeSource{
+								PersistentVolumeClaim: &v1.PersistentVolumeClaimVolumeSource{
+									ClaimName: "my-pre-existing-pvc",
+								},
+							},
+						},
+					},
+				},
+				defaults: []v1.PersistentVolumeClaim{foo, bar},
+			},
+			want: []v1.PersistentVolumeClaim{bar},
+		},
+		{
 			name: "do not add a default pvc when volumeClaimTemplates is empty and user provides own pvc volume",
 			args: args{
 				existing: []v1.PersistentVolumeClaim{},


prodsecmachine:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/fa330a63-8dd8-45ac-b5c4-942e2e5580fd?refs=cdbce6515a71f0c49a3a28d45b7f742ff13abd5e&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/fa330a63-8dd8-45ac-b5c4-942e2e5580fd/license?refs=cdbce6515a71f0c49a3a28d45b7f742ff13abd5e&source=prChecksComment) |

pkg/controller/common/defaults/pvc.go
```diff
@@ -25,18 +25,13 @@ func AppendDefaultPVCs(
 		return existing
 	}
```
moukoublen, line 26: Even though this part does not exist in [logstash](https://github.com/elastic/cloud-on-k8s/blob/2dad60da4493a80ed13be2094a45cbfba9f7da9a/pkg/controller/logstash/volume/volumes.go#L22-L41) logic, if this block were removed, then clusters that already have any user-defined volumeClaimTemplates (e.g., with custom names like my-data instead of elasticsearch-data) could still be passed in with length > 0. Without the shortcut, the code would fall through and append default PVCs. So after an upgrade, users who had only custom-named PVCs could suddenly get default-named PVCs as well. Keeping the `if len(existing) > 0 { return existing }` prevents that.

moukoublen:
Test before/after using `kind` and these resources:

```yaml
kind: Elasticsearch
metadata:
  name: elasticsearch-c1
  namespace: ns1
spec:
  version: 9.3.1
  nodeSets:
  - config:
      node.store.allow_mmap: false
    count: 3
    name: default
    volumeClaimTemplates: []
    podTemplate:
      spec:
        containers:
        - name: elasticsearch
          resources:
            limits:
              cpu: "1"
              memory: 1Gi
          volumeMounts:
          - mountPath: /usr/share/elasticsearch/data
            name: elasticsearch-data
            subPathExpr: $(POD_NAME)
        volumes:
        - name: elasticsearch-data
          persistentVolumeClaim:
            claimName: custom-pvc
---
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: custom-pvc
  namespace: ns1
spec:
  accessModes:
    - ReadWriteOnce
  resources:
    requests:
      storage: 10Gi
```

#### Before
```plain
$ kubectl --namespace ns1 get pvc
NAME                                               STATUS    VOLUME                                     CAPACITY   ACCESS MODES   STORAGECLASS   VOLUMEATTRIBUTESCLASS   AGE
custom-pvc                                         Pending                                                                        e2e-default    <unset>                 74s
elasticsearch-data-elasticsearch-c1-es-default-0   Bound     pvc-8ca142b4-83aa-4cc0-ad83-372090285fb8   1Gi        RWO            e2e-default    <unset>                 47s
elasticsearch-data-elasticsearch-c1-es-default-1   Bound     pvc-e405b740-fde0-4d7b-97e4-e9574528e41e   1Gi        RWO            e2e-default    <unset>                 47s
elasticsearch-data-elasticsearch-c1-es-default-2   Bound     pvc-a55ee24f-b255-4ce5-bc4a-825f527a0351   1Gi        RWO            e2e-default    <unset>                 47s

###### Example of node 0:

Name:             elasticsearch-c1-es-default-0
Namespace:        ns1
Init Containers:
  elastic-internal-init-filesystem:
    Mounts:
      /usr/share/elasticsearch/data from elasticsearch-data (rw)
  elastic-internal-suspend:
    Mounts:
      /usr/share/elasticsearch/data from elasticsearch-data (rw)
Containers:
  elasticsearch:
    Mounts:
      /usr/share/elasticsearch/data from elasticsearch-data (rw)
Volumes:
  elasticsearch-data:
    Type:       PersistentVolumeClaim (a reference to a PersistentVolumeClaim in the same namespace)
    ClaimName:  elasticsearch-data-elasticsearch-c1-es-default-0
    ReadOnly:   false
```


#### After

```plain
$ kubectl --namespace ns1 get pvc
NAME         STATUS   VOLUME                                     CAPACITY   ACCESS MODES   STORAGECLASS   VOLUMEATTRIBUTESCLASS   AGE
custom-pvc   Bound    pvc-bb7bd603-6bea-49ab-ac4b-6f121b0ebfc0   10Gi       RWO            e2e-default    <unset>                 46s


###### Example of node 0:

Name:             elasticsearch-c1-es-default-0
Namespace:        ns1
Init Containers:
  elastic-internal-init-filesystem:
    Mounts:
      /usr/share/elasticsearch/data from elasticsearch-data (rw)
  elastic-internal-suspend:
    Mounts:
      /usr/share/elasticsearch/data from elasticsearch-data (rw)
Containers:
  elasticsearch:
    Mounts:
      /usr/share/elasticsearch/data from elasticsearch-data (rw)
Volumes:
  elasticsearch-data:
    Type:       PersistentVolumeClaim (a reference to a PersistentVolumeClaim in the same namespace)
    ClaimName:  custom-pvc
    ReadOnly:   false
```

pkg/controller/common/defaults/pvc_test.go
```diff
@@ -59,22 +59,44 @@ func TestAppendDefaultPVCs(t *testing.T) {
...
 							},
 						},
 					},
 				},
 				defaults: []v1.PersistentVolumeClaim{bar},
 			},
-			want: []v1.PersistentVolumeClaim{bar},
+			want: nil,
+		},
+		{
```
pkoutsovasilis, line 101: nit: I would add this extra partial overlap unit-test case to have a "contract" in the code that we don't over do it with the skipping 

```suggestion
		},
		{
			name: "skip only the default pvc whose name collides with a user-provided volume",
			args: args{
				existing: nil,
				podSpec: v1.PodSpec{
					Volumes: []v1.Volume{
						{
							Name: foo.Name,
							VolumeSource: v1.VolumeSource{
								PersistentVolumeClaim: &v1.PersistentVolumeClaimVolumeSource{
									ClaimName: "my-pre-existing-pvc",
								},
							},
						},
					},
				},
				defaults: []v1.PersistentVolumeClaim{foo, bar},
			},
			want: []v1.PersistentVolumeClaim{bar},
		},
		{
```

pkoutsovasilis (approved):
LGTM

pkg/controller/common/defaults/pvc_test.go
moukoublen: Done in 4d9df72

Issue #9041, opened 2026-01-16
Remove Elastic Stack 7.17 version support from code.


naemono:
# Remove support for Stack 7.17

## Summary

Following PR #9038 which removes 7.17 support from documentation, we need to remove 7.17 support from the codebase itself.

## Potential areas to update

- **Version validation**: Update supported version ranges in `pkg/controller/common/version/version.go` to exclude 7.17
- **Upgrade path logic**: Update `pkg/controller/elasticsearch/version/supported_versions.go` where 7.17.0 is referenced as the minimum upgrade path to 8.0
- **Version constants**: Update constants like `ServiceAccountMinVersion` and `KibanaServiceAccountMinVersion` in `pkg/apis/elasticsearch/v1/elasticsearch_types.go` and `pkg/apis/kibana/v1/kibana_types.go`
- **Test cases**: Remove or update test cases that reference 7.17.0 versions
- **Deprecated versions**: Update `DeprecatedVersions` range if needed
- **E2E tests**: Update any e2e test configurations that use 7.17

## Note
- Enterprise search 8.x should still be supported as it's listed as End of Support on 15-Jul-2027 according to https://www.elastic.co/support/eol.

dtuck9:
Is this good to include in 3.4? How many releases do we typically wait between the deprecation notice (or more accurately, the removal of the support from the documentation) and removing the functionality?

dtuck9:
It looks like version.go has `DeprecatedVersions` that currently has a max of 7.16.

Could you confirm that this task is to remove all functionality pertaining to 7.x (as opposed to marking references to 7.17 as deprecated)?

pebrc:
@dtuck9 these are good questions. Maybe complete removal is a bit aggressive given that we haven't even emitted a deprecation warning for 7.17. 

For 6.x. we stopped support in 3.0 with a validation and removed the code in 3.3. I believe.

I think I would probably start with a warning and push the actual code removal to a future release 3.5 or 3.6.

dtuck9:
From my understanding, https://github.com/elastic/cloud-on-k8s/pull/9201 covers the case of marking 7.17.x as deprecated. I will keep this task open until we're ready to remove the functionality altogether, though.

PR #9201, opened 2026-03-05
Show all 7.X versions as deprecated

dtuck9:
This is the first step toward removing functionality specific to 7.X for https://github.com/elastic/cloud-on-k8s/issues/9041, and marks all 7.X versions as deprecated.


prodsecmachine:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/abca1df2-b118-4c9c-b816-6bd93374f0f8?refs=20b73110a1cb45c240f6b6080a9bbf0ba1b3eaa3&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/abca1df2-b118-4c9c-b816-6bd93374f0f8/license?refs=20b73110a1cb45c240f6b6080a9bbf0ba1b3eaa3&source=prChecksComment) |

pkg/controller/common/version/version.go
```diff
@@ -23,7 +23,7 @@ var GlobalMinStackVersion Version
 
 // supported Stack versions. See https://www.elastic.co/support/matrix#matrix_compatibility
 var (
-	DeprecatedVersions = MinMaxVersion{Min: From(7, 0, 0), Max: From(7, 16, 0)}
+	DeprecatedVersions = MinMaxVersion{Min: From(7, 0, 0), Max: From(8, 0, 0)}
```
dtuck9: I opted to add 8.0.0 here as this is currently an exclusive max (see [here](https://github.com/elastic/cloud-on-k8s/blob/2dad60da4493a80ed13be2094a45cbfba9f7da9a/pkg/apis/common/v1/validations.go#L86)) and that will ensure all 7.X minors and patches are accounted for.

I think there may be a bug here, though, as in one of the usage of `DeprecatedVersions`, we have an inclusive min and exclusive max:
https://github.com/elastic/cloud-on-k8s/blob/2dad60da4493a80ed13be2094a45cbfba9f7da9a/pkg/apis/common/v1/validations.go#L86-L88

While the `WithinRange` function has both min and max exclusive:
https://github.com/elastic/cloud-on-k8s/blob/2dad60da4493a80ed13be2094a45cbfba9f7da9a/pkg/controller/common/version/version.go#L62-L71

Which is used in the AutoOps reconciler to determine whether the version is EOL:
https://github.com/elastic/cloud-on-k8s/blob/2dad60da4493a80ed13be2094a45cbfba9f7da9a/pkg/controller/autoops/reconcile.go#L124-L127

**Semantically I would expect the `DeprecatedVersions` to be an inclusive range rather than half-open.** but it is at least problematic that we're using it in an exclusive way in one place and a half-open way in another place to define effectively the same thing.

pkoutsovasilis: good catch on the inconsistency @dtuck9 I agree with your assessment. To be honest changing the semantics of WithinRange although feels appropriate can be rather impactful. Maybe in this PR it's worth aligning AutoOps reconciler to use the same approach as we do in validations to be semantically aligned and re-visited in a future PR the semantics of WithinRange?

dtuck9: I agree that changing `WithinRange` seems too impactful for this context of this PR. My preference would be to update validations.go to have an inclusive range (as the semantics of `DeprecatedVersions` would imply all versions contained therein are deprecated), and also update the AutoOps reconciler to use an inclusive range.

Perhaps it's actually worth changing `WithinRange` to `WithinExclusiveRange` and also adding `WithinInclusiveRange`. What do you think?

pkoutsovasilis: yes introducing `WithinExclusiveRange` and using it sounds like a good plan 👍

dtuck9: I now realize I misread the `WithinRange` function and it is in fact already inclusive 🤦‍♀️ But my point still stands! I'm leaving `WithinRange` as-is, and updating validations.go to use it -- the AutoOps reconciler in other words was already doing what I would expect.

dtuck9:
buildkite test this

pkg/controller/common/version/version.go
```diff
@@ -23,7 +23,7 @@ var GlobalMinStackVersion Version
 
 // supported Stack versions. See https://www.elastic.co/support/matrix#matrix_compatibility
 var (
-	DeprecatedVersions = MinMaxVersion{Min: From(7, 0, 0), Max: From(7, 16, 0)}
+	DeprecatedVersions = MinMaxVersion{Min: From(7, 0, 0), Max: From(7, 17, 29)}
```
pebrc: This is fine because there will be no further versions of 7.x afaik. But I wonder if semantically we would want the counterpart to `version.MinFor` so something like `version.MaxFor(7)` to include future theoretical patches as well.

pkoutsovasilis (approved):

pkg/controller/common/version/version.go
dtuck9: I did consider changing this to 7.17.99 for that reason. I will look into adding `version.MaxFor(7)` though

PR #9219, opened 2026-03-12
Introduce AssociationRef interface to abstract ObjectSelector in associations

pkoutsovasilis:
## Summary
  This PR introduces the `AssociationRef` interface to abstract the `ObjectSelector` type behind the `Association.AssociationRef()` method, enabling different reference types for different association kinds.

  ### Key Changes
  - **New `AssociationRef` interface**: Defines a common contract for association references with methods like `IsSet()`, `IsExternal()`, `GetName()`, `GetNamespace()`, `GetSecretName()`, `GetServiceName()`, and `GetNamespacedName()`
  - **`ObjectSelector` methods renamed**: Changed from `NamespacedName()` to `GetNamespacedName()` and `NameOrSecretName()` to `GetNameOrSecretName()` to align with the  interface
  - **`LocalObjectSelector` now implements `AssociationRef`**: Added necessary methods to satisfy the interface, enabling its use for associations that don't support external references (via `secretName`)
  - **Updated all `AssociationRef()` implementations**: Changed return type from `ObjectSelector` to `AssociationRef` across all resource types (Agent, APM, Beat, Elasticsearch, EnterpriseSearch, Kibana, Logstash, Maps)
  - **Kibana `PackageRegistryRef` now uses `LocalObjectSelector`**: Since Package Registry doesn't support external references, using `LocalObjectSelector` enforces this at the type level rather than via runtime validation
 
  ### Benefits
  - Type-safe distinction between references that support external resources (`ObjectSelector`) and those that don't (`LocalObjectSelector`)
  - Extensibility for future reference types with different fields

### DIFF
--- a/pkg/apis/common/v1/common.go
+++ b/pkg/apis/common/v1/common.go
@@ -96,8 +96,13 @@ func (o LocalObjectSelector) NamespacedName() types.NamespacedName {
 
 // IsDefined checks if the local object selector is not nil and has a name.
 // Namespace is not mandatory as it may be inherited by the parent object.
+//
+// Deprecated: Use IsSet instead.
 func (o *LocalObjectSelector) IsDefined() bool {
-	return o != nil && o.Name != ""
+	if o == nil {
+		return false
+	}
+	return o.IsSet()
 }
 
 // IsSet checks if the local object selector has a name.
@@ -205,8 +210,13 @@ func (o ObjectSelector) NamespacedName() types.NamespacedName {
 
 // IsDefined checks if the object selector is not nil and has a name or a secret name.
 // Namespace is not mandatory as it may be inherited by the parent object.
+//
+// Deprecated: Use IsSet instead.
 func (o *ObjectSelector) IsDefined() bool {
-	return o != nil && o.NameOrSecretName() != ""
+	if o == nil {
+		return false
+	}
+	return o.IsSet()
 }
 
 // IsExternal returns true when the object selector references a Kubernetes secret describing an external


prodsecmachine:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/ee561de6-923a-43c4-967f-103abb1b03a4?refs=19408da521c5ae8234733f063c78255c89c6ec84&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/ee561de6-923a-43c4-967f-103abb1b03a4/license?refs=19408da521c5ae8234733f063c78255c89c6ec84&source=prChecksComment) |

docs/reference/api-reference/main.md
```diff
@@ -466,6 +466,8 @@ Package v1 contains API schema definitions for common types used by all resource
 
 
 
+
+
```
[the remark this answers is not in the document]
pkoutsovasilis: this is a generated file

test/e2e/test/beat/builder.go
```diff
@@ -343,14 +343,14 @@ func (b Builder) GetMetricsCluster() *types.NamespacedName {
...
 	return &metricsCluster
 }
 
 func (b Builder) GetLogsCluster() *types.NamespacedName {
 	if len(b.Beat.Spec.Monitoring.Metrics.ElasticsearchRefs) == 0 {
 		return nil
 	}
-	metricsCluster := b.Beat.Spec.Monitoring.Logs.ElasticsearchRefs[0].NamespacedName()
+	metricsCluster := b.Beat.Spec.Monitoring.Logs.ElasticsearchRefs[0].GetNamespacedName()
 	return &metricsCluster
```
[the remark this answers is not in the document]
pkoutsovasilis: ha! this PR didn't cause this one but we did have an issue there

pkg/apis/kibana/v1/webhook.go
```diff
@@ -153,13 +152,6 @@ func checkAssociations(k *Kibana) field.ErrorList {
 	err2 := commonv1.CheckAssociationRefs(monitoringPath.Child("logs"), k.GetMonitoringLogsRefs()...)
 	err3 := commonv1.CheckAssociationRefs(field.NewPath("spec").Child("elasticsearchRef"), k.Spec.ElasticsearchRef)
 	err4 := commonv1.CheckAssociationRefs(field.NewPath("spec").Child("enterpriseSearchRef"), k.Spec.EnterpriseSearchRef)
-	err5 := commonv1.CheckAssociationRefs(field.NewPath("spec").Child("packageRegistryRef"), k.Spec.PackageRegistryRef)
+	err5 := commonv1.CheckLocalAssociationRefs(field.NewPath("spec").Child("packageRegistryRef"), k.Spec.PackageRegistryRef)
 	return append(err1, append(err2, append(err3, append(err4, err5...)...)...)...)
```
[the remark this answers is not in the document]
pkoutsovasilis: we want this to be pruned and not rejected. Back in the original implementation of EPR we had to go with ObjectSelector as the necessary abstraction wasn't in place to go with LocalObjectSelector

pkg/controller/common/stackmon/name.go
```diff
@@ -24,7 +24,7 @@ func configVolumeName(name string, beatName string) string {
 }
 
 func caVolumeName(assoc commonv1.Association) string {
-	nsn := assoc.AssociationRef().Namespace + assoc.AssociationRef().NameOrSecretName()
+	nsn := assoc.AssociationRef().GetNamespace() + assoc.AssociationRef().GetNameOrSecretName()
 	nsnHash := fmt.Sprintf("%x", sha256.Sum256([]byte(nsn)))[0:6]
```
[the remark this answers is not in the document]
pkoutsovasilis: this would change pre-existing behaviour so not gonna do it

pkoutsovasilis:
buildkite test this -f p=gke,s=9.3.1,t=TestEPRAssociation

pkg/apis/common/v1/association.go
```diff
@@ -150,6 +151,27 @@ type Associated interface {
...
+	// GetNamespace returns the namespace of the referenced resource.
+	GetNamespace() string
+	// GetSecretName returns the secret name (useful for external/unmanaged associations).
+	GetSecretName() string
+	// GetServiceName returns the service name of the referenced resource.
+	GetServiceName() string
+	// GetNamespacedName returns the NamespacedName of the referenced resource.
+	GetNamespacedName() types.NamespacedName
+	// GetNameOrSecretName returns the name or the secret name of the reference.
+	GetNameOrSecretName() string
```
moukoublen: One design concern:

Unless I am missing something, I feel that having separate methods like `GetName()` and `GetSecretName()` (and possibly `GetNameOrSecretName()`) pushes the interface toward specialization (callers branch on "name vs secret name") instead of a single abstraction. As more association types are added, we'd likely keep adding more getters (GetXName(), etc.), which makes the interface harder to implement and to use in a generic way.

A more abstract approach could be: a single identifier accessor plus a type, e.g. `GetName() string` and `GetRefType() RefType` (or similar), so callers get "the" name for the reference and can branch on type when needed. That would keep the interface stable as new reference types are introduced and avoid method proliferation.

```golang
type RefType int

const (
	RefTypeManaged RefType = iota
	RefTypeSecret
)
```
And more types can be added (or put more specialized types if needed) as we go, without API changes. 

(PS ofc I asked AI to remind me of the term "method proliferation" and most probably I will forget it later 😅)

pkoutsovasilis: This interface is intentionally a direct abstraction of the existing ObjectSelector struct - in a sense all these methods already exist (either in the form of fields or methods) and are called throughout the codebase. The goal isn't to introduce a new abstraction layer but to enable returning different concrete types (e.g., LocalObjectSelector for PackageRegistry which doesn't support secretName) while keeping the same API surface that callers already use, e.g. association controller. 

That said, the RefType approach is an interesting idea for simplifying the interface but IMO it should be follow-up PR once we have more concrete use cases that would benefit from it. wdyt?

moukoublen: To be honest, since we do introduce an interface (and thus an opportunity for abstractions) and since we do introduce breaking changes (for example, renaming of ObjectSelector receiver functions), it seems to me like the perfect opportunity to introduce a proper abstraction (and communicate downstream) rather than introducing two separate breaking PRs and having to communicate 2 separate sets of breaking changes downstream. 

After all, the exported data from the ObjectSelector (which is used downstream) and the LocalObjectSelector will not change. 

But I would also appreciate thoughts from other fellas. 

Update: Alternatively, I would keep the ObjectSelector receiver functions as is and name them accordingly in the common interface to avoid introducing any downstream breaking changes.

pkoutsovasilis: no more breaking changes 8f5ef66e34e3b3318badd096cfae3bdeae311976 but sure let's wait for input from the rest of the team

moukoublen: On second thought, I agree, let's not do the proper abstractation in this PR, but I would suggest avoiding introducing breaking changes to ObjectSelector. Wdyt?

pkg/apis/common/v1/common.go
```diff
@@ -156,10 +197,10 @@ func (o ObjectSelector) NamespacedName() types.NamespacedName {
...
 }
 
-// IsDefined checks if the object selector is not nil and has a name or a secret name.
+// IsDefined checks if the object selector has a name or a secret name set.
 // Namespace is not mandatory as it may be inherited by the parent object.
-func (o *ObjectSelector) IsDefined() bool {
-	return o != nil && o.NameOrSecretName() != ""
+func (o ObjectSelector) IsDefined() bool {
+	return o.NameOrSecretName() != ""
 }
```
moukoublen: Unfortunatelly this can introduce a silent panic downstream. 

Even though we don't use `*ObjectSelector` anywhere in this repo (though this is heavily used, I didn't search all usages), if anyone downstream does use a pointer of ObjectSelector, after upgrading, a panic can occur in runtime.


See this example: https://go.dev/play/p/9F3HU1an68_0

Even though I'd love to get rid of that pointer receiver, I think it's a tricky situation for downstream projects.

pkoutsovasilis: ok I cannot maintain this as a pointer-receiver func and satisfy the interface at the same time, thus I had to introduce a new func `IsSet` as value receiver to satisfy the interface 3f04a13f39807b856b12bae71dea58f8dfeeb822

pkg/apis/kibana/v1/kibana_types.go
```diff
@@ -79,7 +79,7 @@ type KibanaSpec struct {
 	ElasticsearchRef commonv1.ObjectSelector `json:"elasticsearchRef,omitempty"`
 
 	// PackageRegistryRef is a reference to an Elastic Package Registry running in the same Kubernetes cluster.
-	PackageRegistryRef commonv1.ObjectSelector `json:"packageRegistryRef,omitempty"`
+	PackageRegistryRef commonv1.LocalObjectSelector `json:"packageRegistryRef,omitempty"`
```
pebrc, line 82: We are effectively narrowing the interface here. I am trying to think of scenarios where this would break. The only thing I can think of are users disabling the webhook and putting something into `secretName` It would be a non-functional configuration and would now just be pruned IIUC. So I think we are OK here.

test/e2e/test/beat/builder.go
```diff
@@ -348,9 +348,9 @@ func (b Builder) GetMetricsCluster() *types.NamespacedName {
 }
 
 func (b Builder) GetLogsCluster() *types.NamespacedName {
-	if len(b.Beat.Spec.Monitoring.Metrics.ElasticsearchRefs) == 0 {
+	if len(b.Beat.Spec.Monitoring.Logs.ElasticsearchRefs) == 0 {
```
pebrc, line 351: Nice catch!

pkg/apis/common/v1/common.go
```diff
@@ -100,6 +100,53 @@ func (o *LocalObjectSelector) IsDefined() bool {
 	return o != nil && o.Name != ""
```
pebrc: Should delegate to `IsSet`. Should be marked deprecated?

pkg/apis/common/v1/common.go
pebrc: Same here, should delegate to `IsSet`, mark deprecated?

pkg/apis/kibana/v1/kibana_types.go
pkoutsovasilis: Agreed, I don't see any issue here either. `secretName` was never a functional field for packageRegistryRef - it was only exposed because we had to use `ObjectSelector` at the time since we couldn't use `LocalObjectSelector` yet. AFAICT with the field now removed from the CRD schema, the API server will prune it on any new create/update, so anyone who had it set with the webhook disabled would just have their no-op configuration silently stripped on the next write.

pkg/apis/common/v1/common.go
pkoutsovasilis: changed in 19408da521c5ae8234733f063c78255c89c6ec84

pkg/apis/common/v1/common.go
pkoutsovasilis: changed in 19408da521c5ae8234733f063c78255c89c6ec84

Issue #9081, opened 2026-01-28
ECK-Generated Client Certificates


pebrc:
* ECK generates client certificates for associations (similar to self-signed server certs)
* Reuse existing per-application CA infrastructure
* ECK configures ES to trust client certificates signed by this CA
* ECK configures mTLS between stack applications where applicable (e.g. Fleet Server and Elastic Agent) 
* ECK configures ES to trust client certificates signed by a new ECK operator CA
* Users can optionally provide their own client certificates
* Opt-in via new `clientTLS` configuration section (Naming TBD)

Needs detailed technical design.

pkoutsovasilis:
All PRs that enable mTLS support for both Elasticseach and Fleet-server are merged, thus this is completed.

PR #9012, opened 2026-01-09
Set seccompProfile to RuntimeDefault

barkbay:
This PR adds a Seccomp Profile to the existing **Pod** security context.

| Controller | Seccomp Profile | Condition |
|------------|-----------------|-----------|
| **Elasticsearch** | `RuntimeDefault` | Version > 8.8 |
| **Kibana** | `RuntimeDefault` | Version > 7.10 |
| **Logstash** | `RuntimeDefault` | Always |
| **APM Server** | `RuntimeDefault` | Always |
| **Maps** | `RuntimeDefault` | Always |
| **PackageRegistry** | `RuntimeDefault` | Always |
|**AutoOpsAgentPolicy**| `RuntimeDefault` | Always |
| **Beats** | Not set | No security context by default |
| **Agent** | Not set | No security context by default |

PR #9229, opened 2026-03-16
Implement client certificate required support for Elasticsearch

pkoutsovasilis:
## Summary

  This PR implements the Elasticsearch side of the mTLS client certificate authentication feature ([#9081](https://github.com/elastic/cloud-on-k8s/issues/9081)). It adds a new `spec.http.tls.client.authentication` field to the Elasticsearch CRD that, when enabled, configures Elasticsearch to require client certificates on its HTTP  interface.

### Changes

  - **New API types**: `HTTPConfigWithClientOptions`, `TLSWithClientOptions`, and `ClientOptions` to support `spec.http.tls.client.authentication` on Elasticsearch
  - **Client certificate trust bundle**: Aggregates client certificates labeled with `eck.k8s.elastic.co/client-certificate: true` and the corresponding es cluster soft-owner labels into a trust bundle mounted into Elasticsearch pods
  - **Elasticsearch configuration**: Automatically sets `xpack.security.http.ssl.client_authentication: required` and configures the client CA path when client authentication is enabled and the manual configuration specified by the user either in the CRD or from referencing StackConfigPolicies doesn't explicitly set it.
  - **Operator client certificate**: ECK generates and manages a client certificate for the operator to continue communicating with Elasticsearch when client authentication is enabled
  - **Annotation-based signaling**: Sets a `client-authentication-required` annotation on the Elasticsearch CRD to signal that client certificates are needed
  - **Graceful transitions**: Defers cleanup of mTLS resources until all pods have rolled over, ensuring connectivity throughout configuration changes

### API

```yaml
...
  spec:
    http:
      tls:
        client:
          authentication: true  # Enable client certificate authentication
...
```

###  Test plan
  - Unit tests pass for new types, annotations, certificate reconciliation, ES config merging, and volume mounts
  - E2E tests validate enabling mTLS, operator connectivity, and disabling mTLS on Elasticsearch
  - Verify CRD schema is correct and backward-compatible

### DIFF
--- a/pkg/controller/elasticsearch/settings/merged_config_test.go
+++ b/pkg/controller/elasticsearch/settings/merged_config_test.go
@@ -569,6 +569,14 @@ func TestHasClientAuthenticationRequired(t *testing.T) {
 			cfg:  map[string]any{},
 			want: false,
 		},
+		{
+			name: "required but ssl disabled returns false",
+			cfg: map[string]any{
+				esv1.XPackSecurityHttpSslClientAuthentication: "required",
+				esv1.XPackSecurityHttpSslEnabled:              "false",
+			},
+			want: false,
+		},
 	} {
 		t.Run(tt.name, func(t *testing.T) {
 			cfg := CanonicalConfig{common.MustCanonicalConfig(tt.cfg)}


prodsecmachine:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/0b8e1615-5499-4af1-a9ff-2a5d43dcc3af?refs=4cb12548110194c1e18bebcdced7863503a7c655&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/0b8e1615-5499-4af1-a9ff-2a5d43dcc3af/license?refs=4cb12548110194c1e18bebcdced7863503a7c655&source=prChecksComment) |

pkoutsovasilis:
buildkite test this -f p=gke,t=TestClientAuthRequired.* -m s=9.3.1,s=8.19.12,s=9.2.6

pkoutsovasilis:
buildkite test this -f p=gke,t=.*,E2E_TAGS=es -m s=9.3.1

pkoutsovasilis:
buildkite test this -f p=gke,t=.*,E2E_TAGS=es -m s=9.3.1

pkg/apis/common/v1/common.go
```diff
@@ -310,6 +326,25 @@ func (tls TLSOptions) Enabled() bool {
...
+}
+
+// Enabled returns true when TLS is enabled based on this option struct.
+func (tls TLSWithClientOptions) Enabled() bool {
+	return tls.TLSOptions.Enabled()
+}
+
+// ClientOptions configures client certificate authentication for incoming connections.
+type ClientOptions struct {
+	// Authentication enables client authentication.
```
pebrc: Should we mention that this is enterprise only?

pkoutsovasilis: good point! changed in 246cb8b3f76c9402171d68833e57c954128dc094

pkg/controller/elasticsearch/driver/stateful/nodes.go
```diff
@@ -84,7 +85,8 @@ func (d *Driver) reconcileNodeSpecs(
 		return results.WithError(err)
 	}
 
-	expectedResources, err := nodespec.BuildExpectedResources(ctx, d.Client, d.ES, keystoreResources, actualStatefulSets, d.OperatorParameters.IPFamily, d.OperatorParameters.SetDefaultSecurityContext, meta)
+	// Build expected resources using pre-computed configs from ResolvedConfig.
+	expectedResources, err := nodespec.BuildExpectedResources(ctx, d.Client, d.ES, keystoreResources, actualStatefulSets, d.OperatorParameters.SetDefaultSecurityContext, meta, resolvedConfig.NodeSetConfigs, resolvedConfig.ClientAuthenticationRequired, resolvedConfig.PolicyConfig)
```
pebrc: Very long line? 10 params to the function? Pass `resolvedConfig` as a whole instead?

pkg/controller/elasticsearch/nodespec/resources.go
```diff
@@ -54,59 +52,43 @@ func (l ResourcesList) ExpectedNodeCount() int32 {
...
-
-	// Get policy config from StackConfigPolicy
-	policyConfig, err := getPolicyConfig(ctx, client, es)
-	if err != nil {
-		return nil, err
-	}
 
 	// we retrieve the current pods restart trigger annotation.
 	actualPodsRestartTriggerAnnotationValue, err := es_sset.GetActualPodsRestartTriggerAnnotationForCluster(client, es)
 	if err != nil {
```
pebrc, line 91: This positional lookup is brittle — it relies on `nodeSetConfigs` being in the same order as `es.Spec.NodeSets`, which is an implicit contract between two functions that's easy to break silently. The name check on line 80 converts a silent bug into a runtime error but doesn't eliminate the fragility. A `map[string]NodeSetConfig` keyed by `NodeSetName` would remove the ordering dependency entirely.

pkg/controller/elasticsearch/driver/shared/reconcile.go
```diff
@@ -462,3 +477,28 @@ func esReachableConditionMessage(internalService *corev1.Service, isServiceReady
 		return fmt.Sprintf("Service %s/%s has endpoints", internalService.Namespace, internalService.Name)
 	}
 }
+
+// DeleteClientCertResources deletes the client certificate resources (operator client cert secret,
+// trust bundle secret, and the client-authentication-required annotation) when client certificate
+// authentication is no longer required. This should only be called after all pods have rolled
+// to the new configuration without client authentication.
+func DeleteClientCertResources(ctx context.Context, c k8s.Client, es *esv1.Elasticsearch) error {
```
pebrc: `DeleteClientCertResources` is only called from one place (`stateful/nodes.go`) and its logic — removing an annotation and deleting two secrets — is certificate lifecycle management. It would sit more naturally in `pkg/controller/common/certificates/client_cert_reconcile.go`, alongside the code that creates those same resources. The current placement in `shared/reconcile.go` looks like it may have been driven by import cycle avoidance rather than cohesion.

pkg/controller/common/certificates/client_cert_reconcile.go
```diff
@@ -0,0 +1,405 @@
...
+	}
+
+	reconciledSecret, err := reconciler.ReconcileSecret(ctx, r.K8sClient, *secret, r.Owner)
+	if err != nil {
+		return nil, err
+	}
+	return &CertificatesSecret{
+		Secret: reconciledSecret,
+	}, nil
+}
```
pebrc, line 111: The manual label/annotation setup and nil-checks here are redundant — `reconciler.ReconcileSecret` already merges labels and annotations from `expected` into the reconciled secret (preserving any existing ones). The function could be simplified to just build an `expected` secret with the correct metadata and cert data, then delegate everything else to `ReconcileSecret`.

pkg/controller/common/certificates/client_cert_reconcile.go
```diff
@@ -0,0 +1,405 @@
...
+// The output is sorted by secret namespace/name for deterministic results.
+func buildTrustBundleFromSecrets(ctx context.Context, secrets []corev1.Secret) []byte {
+	log := ulog.FromContext(ctx)
+
+	sort.Slice(secrets, func(i, j int) bool {
+		if secrets[i].Namespace != secrets[j].Namespace {
+			return secrets[i].Namespace < secrets[j].Namespace
+		}
+		return secrets[i].Name < secrets[j].Name
+	})
```
pebrc: This can be simplified using `slices.SortFunc` and `cmp.Or` (available since Go 1.21):

```go
slices.SortFunc(secrets, func(a, b corev1.Secret) int {
    return cmp.Or(
        strings.Compare(a.Namespace, b.Namespace),
        strings.Compare(a.Name, b.Name),
    )
})
```

pkg/controller/common/certificates/client_cert_reconcile.go
```diff
@@ -0,0 +1,405 @@
...
+	}
+
+	delete(secret.Data, CAFileName)
+
+	return nil
+}
+
+// getClientCertificate returns the client certificate from the provided Secret if it's valid.
+// Returns nil if the certificate needs to be re-issued.
+func getClientCertificate(
```
pebrc: The name `getClientCertificate` reads like a simple accessor, but the function validates expiry, CN, and key match — returning `nil` to signal reissuance is needed. `clientCertificateIfValid` would be more idiomatic and better convey the intent.

pkg/controller/common/certificates/client_cert_reconcile.go
```diff
@@ -0,0 +1,405 @@
...
+			"namespace", secret.Namespace, "secret_name", secret.Name)
+		return nil
+	}
+
+	return cert.Raw
+}
+
+// getExistingPrivateKey parses the existing private key from the secret.
+// Returns nil if the key is missing or cannot be parsed.
+func getExistingPrivateKey(ctx context.Context, secret *corev1.Secret) crypto.Signer {
```
pebrc: Same as `getClientCertificate` — this parses and validates rather than just fetching. `privateKeyIfValid` would be more idiomatic.

pkg/controller/elasticsearch/configmap/configmap_test.go
```diff
@@ -108,6 +109,43 @@ func TestReconcileScriptsConfigMap(t *testing.T) {
...
+				var cm corev1.ConfigMap
+				err := client.Get(context.Background(), types.NamespacedName{Namespace: namespace, Name: configMapName}, &cm)
+				assert.NoError(t, err)
+				script := cm.Data[nodespec.PreStopHookScriptConfigKey]
+				assert.NotContains(t, script, "CLIENT_CERT=")
+				assert.NotContains(t, script, `"${CLIENT_CERT[@]}"`)
+				assert.NotContains(t, script, "--cert")
+				assert.NotContains(t, script, "--key")
+			},
+		},
```
pebrc: The tests here check for presence/absence of specific strings but don't do a full content comparison against a golden file. The podspec snapshot test doesn't help either — it uses an empty scripts configmap so the config-hash annotation doesn't reflect actual script content.

A golden file test rendering the script with `clientAuthenticationRequired=false` and comparing byte-for-byte against the expected output would guard against accidental template changes triggering unnecessary rolling restarts on operator upgrade.

pkg/controller/elasticsearch/settings/merged_config.go
```diff
@@ -190,3 +205,86 @@ func xpackConfig(ver version.Version, httpCfg commonv1.HTTPConfig, remoteCluster
...
+		esv1.XPackSecurityHttpSslCertificateAuthorities: casToMerge,
+	})
+	if err != nil {
+		return fmt.Errorf("failed to create trust bundle config: %w", err)
+	}
+	if err := config.MergeWith(trustBundleConfig); err != nil {
+		return fmt.Errorf("failed to merge trust bundle config: %w", err)
+	}
+	return nil
+}
```
pebrc: Since ucfg is configured with `AppendValues` (see `canonical_config.go`), the entire read-modify-write pattern here is unnecessary. There's no need to read the existing CAs, type-switch on them, or reconstruct the full list — `MergeWith` will append automatically. The idempotency check is also unnecessary since duplicate CA paths are harmless to Elasticsearch. The whole function could be:

```go
func appendClientTrustBundle(config *common.CanonicalConfig) error {
    trustBundleConfig, err := common.NewCanonicalConfigFrom(map[string]any{
        esv1.XPackSecurityHttpSslCertificateAuthorities: []string{
            path.Join(volume.ClientCertificatesTrustBundleMountPath, certificates.ClientCertificatesTrustBundleFileName),
        },
    })
    if err != nil {
        return err
    }
    return config.MergeWith(trustBundleConfig)
}
```
Prerequisite is that we switch the default xpack HTTP CA config to a slice (which we should anyway). This comes at the price of a full cluster restart for all customers when they upgrade (not sure if this is unavoidable anyway with this feature or whether we can opportunistically leverage the fact that other features we are planning to ship require a restart already) but might be a tradeoff worth making.

pkg/controller/elasticsearch/driver/stateful/nodes.go
pkoutsovasilis: changed in b7bd7d536c9004355617d2f65d641ab71724cfa5

pkg/controller/elasticsearch/nodespec/resources.go
pkoutsovasilis: changed in b7bd7d536c9004355617d2f65d641ab71724cfa5

pkg/controller/elasticsearch/driver/shared/reconcile.go
pkoutsovasilis: changed in f42728ce1bdc2f4eb23a8c11e11234b71b3e14bf

pkg/controller/common/certificates/client_cert_reconcile.go
pkoutsovasilis: good point! changed in 1ceda5dce5564e2e7d1041cbf5b48974b11322ef

pkg/controller/common/certificates/client_cert_reconcile.go
pkoutsovasilis: idiomatic sorting here 0609f65e1668555098116695d05e5b181ed20ed6

pkg/controller/common/certificates/client_cert_reconcile.go
pkoutsovasilis: renamed here b6c7e76c2c222d2d225ba51453910eff98eb9d7d

pkg/controller/common/certificates/client_cert_reconcile.go
pkoutsovasilis: renamed here b6c7e76c2c222d2d225ba51453910eff98eb9d7d

pkg/controller/elasticsearch/configmap/configmap_test.go
pkoutsovasilis: added here b48d69963a1e07e9cc80a82740dd24eced643f9f and here 821a02d69230efdaf6c25ba6c55e58dc05166924

pkg/controller/elasticsearch/settings/merged_config.go
pkoutsovasilis: what you say sounds like music to my ears but there is another scenario that I had in mind and I wrote this complicated logic, what if a user has a manual entry there as a string? If we go with your proposal we will override that as well. However, I do like it and if you consider the former acceptable more than happy to reform this func

pkg/controller/elasticsearch/nodespec/lifecycle_hook_test.go
```diff
@@ -0,0 +1,55 @@
...
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := RenderPreStopHookScript(svcURL, tt.clientAuthenticationRequired)
+			require.NoError(t, err)
+
+			goldenPath := filepath.Join("testdata", tt.goldenFile)
+
+			if *updateGolden {
+				require.NoError(t, os.WriteFile(goldenPath, []byte(got), 0644))
```
pebrc: This hand-rolls a golden file test (custom `-update-golden` flag, manual `os.ReadFile`/`os.WriteFile`) while the rest of the codebase uses the `snaps` library for the same pattern. Consider using `snaps` here for consistency.

pkg/controller/common/annotation/client_cert_test.go
```diff
@@ -0,0 +1,122 @@
...
+
+	"github.com/stretchr/testify/require"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+
+	"github.com/elastic/cloud-on-k8s/v3/pkg/utils/k8s"
+)
+
+func TestHasClientAuthenticationRequired(t *testing.T) {
```
pebrc, line 19: nit: `TestHasClientAuthenticationRequired` is a pure input/output function with no side effects — a table-driven test would be more idiomatic here. The `Set`/`Remove` tests are a closer call given the per-case fake client setup.

pkg/controller/common/certificates/client_cert_reconcile.go
```diff
@@ -0,0 +1,421 @@
...
+// Returns nil (without error) if the secret does not exist.
+func LoadOperatorClientCertIfExists(ctx context.Context, c k8s.Client, namer name.Namer, namespace, ownerName string) (*tls.Certificate, error) {
+	secretName := OperatorClientCertSecretName(namer, ownerName)
+	secret, err := k8s.GetSecretIfExists(ctx, c, types.NamespacedName{Namespace: namespace, Name: secretName})
+	if err != nil {
+		return nil, err
+	}
+	if secret == nil {
+		return nil, nil
+	}
```
pebrc: nit: the two early returns can be compressed:
```go
if err != nil || secret == nil {
    return nil, err
}
```

pkg/controller/common/certificates/client_cert_reconcile.go
```diff
@@ -0,0 +1,421 @@
...
+	privateKey := privateKeyIfValid(ctx, secret)
+	if privateKey == nil {
+		generatedPrivateKey, err := rsa.GenerateKey(cryptorand.Reader, 2048)
+		if err != nil {
+			return err
+		}
+		privateKey = generatedPrivateKey
+
+		// Use PKCS#8 encoding for client certificate keys to ensure compatibility with
+		// all elastic stack components
```
pebrc: nit: missing period and "Elastic Stack" should be capitalised (it's a product name):
```
// Use PKCS#8 encoding for client certificate keys to ensure compatibility with
// all Elastic Stack components.
```

test/e2e/es/client_auth_test.go
```diff
@@ -0,0 +1,268 @@
...
+	// Use the builder's actual ES name (includes random suffix from NewBuilder).
+	actualESName := initialBuilder.Elasticsearch.Name
+	clientCertSecretName := certificates.OperatorClientCertSecretName(esv1.ESNamer, actualESName)
+	trustBundleSecretName := certificates.ClientCertTrustBundleSecretName(esv1.ESNamer, actualESName)
+
+	// Helper steps to assert client cert resources exist
+	verifyClientCertResourcesExist := test.StepList{
+		{
+			Name: "Verify client-certificate-required annotation is set",
+			Test: test.Eventually(func() error {
```
pebrc: The annotation check and client cert secret verification steps here are duplicated from `TestClientAuthRequired` (lines ~51-74). Consider extracting into a helper function, e.g.:
```go
func verifyClientCertResourcesExist(k *test.K8sClient, namespace, esName, clientCertSecretName, trustBundleSecretName string) test.StepList
```
Both tests could then share the same verification logic.

pkg/utils/k8s/k8sutils.go
```diff
@@ -211,16 +211,30 @@ func DeleteSecretMatching(ctx context.Context, c Client, opts ...client.ListOpti
...
+// DeleteSecretIfExists deletes the secret identified by key if exists.
+func DeleteSecretIfExists(ctx context.Context, c Client, key types.NamespacedName) error {
+	secret, err := GetSecretIfExists(ctx, c, key)
+	if err != nil {
 		return err
 	}
-	err = c.Delete(ctx, &secret)
+	if secret == nil {
+		return nil
+	}
```
pebrc: nit: can be contracted to:
```go
if err != nil || secret == nil {
    return err
}
```

pkg/utils/k8s/k8sutils.go
```diff
@@ -211,16 +211,30 @@ func DeleteSecretMatching(ctx context.Context, c Client, opts ...client.ListOpti
...
+
+// DeleteSecretIfExists deletes the secret identified by key if exists.
+func DeleteSecretIfExists(ctx context.Context, c Client, key types.NamespacedName) error {
+	secret, err := GetSecretIfExists(ctx, c, key)
+	if err != nil {
 		return err
 	}
-	err = c.Delete(ctx, &secret)
+	if secret == nil {
+		return nil
```
pebrc: `DeleteSecretIfExists` duplicates the logic of the existing `DeleteResourceIfExists` elsewhere in this file. It could delegate to it:
```go
func DeleteSecretIfExists(ctx context.Context, c Client, key types.NamespacedName) error {
    return DeleteResourceIfExists(ctx, c, &corev1.Secret{ObjectMeta: metav1.ObjectMeta{
        Namespace: key.Namespace,
        Name:      key.Name,
    }})
}
```
Or callers could use `DeleteResourceIfExists` directly, making `DeleteSecretIfExists` redundant.

pkg/controller/elasticsearch/validation/validations.go
```diff
@@ -605,3 +607,23 @@ func validateRestartTriggerWarnings(ctx context.Context, k8sClient k8s.Client, o
 
 	return ""
 }
+
+func validClientAuthentication(ctx context.Context, es esv1.Elasticsearch, checker license.Checker) field.ErrorList {
+	if !es.Spec.HTTP.TLS.Client.Authentication {
+		return nil
+	}
```
pebrc, line 619: The license check intentionally only covers `spec.http.tls.client.authentication` (Option B) and does not gate the raw config path (`xpack.security.http.ssl.client_authentication: required`). This is by design for two reasons:
1. Extending the check to the raw config path would be a breaking change for users who manually configured client auth before this feature existed.
2. StackConfigPolicy-driven client auth (the only way to centrally enable mTLS across multiple clusters) relies on the raw config path, and StackConfigPolicy is itself an enterprise feature.

Do not add the raw config path to this check.

pkg/controller/elasticsearch/validation/validations.go
```diff
@@ -605,3 +607,23 @@ func validateRestartTriggerWarnings(ctx context.Context, k8sClient k8s.Client, o
...
+	if err != nil {
+		ulog.FromContext(ctx).Error(err, "while checking enterprise features during client authentication validation")
+		return nil
+	}
+	if !enabled {
+		return field.ErrorList{
+			field.Forbidden(
+				field.NewPath("spec").Child("http", "tls", "client", "authentication"),
+				"client certificate authentication is an enterprise feature. Enterprise features are disabled",
+			),
```
pebrc, line 639: nit: the message repeats "enterprise feature/features" in two consecutive sentences. Consider: `"client certificate authentication requires an enterprise license"`

pkg/controller/elasticsearch/settings/merged_config.go
pebrc: Good point — you're right that `AppendValues` only applies when the existing value is already an array. When the existing value is a scalar string, `mergeValues` replaces it rather than appending, so the simplified approach would silently drop a user's manual CA entry. The current logic is justified for that reason.

That said, the pattern of "append a string to a key that might currently be a scalar or array" is generic enough to belong on `CanonicalConfig` as a helper, e.g.:
```
func (c *CanonicalConfig) AppendString(key, value string) error
```

This would encapsulate the type-switching and read-modify-write logic in one place, keep `appendClientTrustBundle` simple, and be reusable for any future cases where ECK needs to append to a user-configurable string/array setting.

pkg/controller/elasticsearch/driver/shared/reconcile.go
```diff
@@ -156,6 +158,15 @@ func ReconcileSharedResources(
...
 	}
 
+	// Reconcile operator client certificate and trust bundle of client certificates.
+	operatorClientCert, clientCertResults := certificates.ReconcileOperatorClientCertAndTrustBundle(
+		ctx, d, &es, clientAuthenticationRequired, params.OperatorParameters.CertRotation, meta,
+	)
+	results.WithResults(clientCertResults)
+	if clientCertResults.HasError() {
+		return nil, results
+	}
```
pebrc, line 173: The early return here may be overly conservative. Unlike the HTTP cert reconciliation above (where failure prevents building the trusted CA pool and makes subsequent steps meaningless), a failure here just means the operator client cert is unavailable — the ES client is still created (without a client cert), `esReachable` becomes false if client auth is required, and subsequent steps handle that gracefully with a requeue.

The results aggregator was designed to accumulate errors across steps rather than gate on each one. Removing the early return and letting the error accumulate would be consistent with that intent and wouldn't endanger cluster stability.

pkg/controller/elasticsearch/driver/stateful/driver.go
```diff
@@ -30,16 +39,43 @@ var _ commondriver.Interface = (*Driver)(nil)
...
+
+	// Resolve configuration first. This computes merged configs for all NodeSets
+	// (including StackConfigPolicy) and detects clientAuthenticationRequired early,
+	// before we create the ES client.
+	resolvedConfig, err := ResolveConfig(ctx, d.Client, d.ES, d.OperatorParameters.IPFamily)
+	if err != nil {
+		return results.WithError(err)
+	}
+
+	if resolvedConfig.ClientAuthenticationRequired && !enabled {
```
pebrc: This runtime license check blocks reconciliation for users who have `xpack.security.http.ssl.client_authentication: required` set via raw config (Option A), regardless of whether they set it before this feature existed. On operator upgrade, these users would find their cluster stuck in a 5-minute requeue loop even though nothing changed on their side — a breaking change.

The validation layer correctly avoids this by only gating `spec.http.tls.client.authentication` (Option B). The same reasoning applies here: the license check should only block when the user has explicitly opted into ECK-managed client auth via the spec field.

Beyond the breaking change, the check is also redundant for Option B: `ValidateElasticsearch` (including `validClientAuthentication`) is called at the start of every reconciliation in `elasticsearch_controller.go`, not just in the admission webhook — so Option B is already blocked before reaching this code. And Option A via StackConfigPolicy is implicitly covered since StackConfigPolicy itself requires an enterprise license.

The check should either be removed entirely or scoped to es.Spec.HTTP.TLS.Client.Authentication only.

pkg/controller/common/certificates/client_cert_reconcile.go
```diff
@@ -0,0 +1,421 @@
...
+	return ParseTLSCertificate(*secret)
+}
+
+// DeleteClientCertResources deletes the client certificate resources (operator client cert secret,
+// trust bundle secret, and the client-authentication-required annotation) when client certificate
+// authentication is no longer required. This should only be called after all pods have rolled
+// to the new configuration without client authentication.
+func DeleteClientCertResources(ctx context.Context, c k8s.Client, owner client.Object, namer name.Namer) error {
+	// Remove the client-authentication-required annotation
+	if err := annotation.RemoveClientAuthenticationRequiredAnnotation(ctx, c, owner); err != nil {
```
pebrc: The cleanup ordering here is wrong: the annotation is removed first, then the secrets are deleted. If either `DeleteSecretIfExists` call fails, the annotation is already gone — so the next reconciliation sees no annotation, takes the early-return path, and never retries the secret deletion. The orphaned secrets persist indefinitely.

The fix is to reverse the order: delete both secrets first, then remove the annotation. That way, a transient failure leaves the annotation intact and the next reconciliation will retry the full cleanup sequence.

pkg/controller/elasticsearch/nodespec/lifecycle_hook_test.go
pkoutsovasilis: thanks for keeping me consistent here, changed in eca07cb36ad065aeef69b3f20e5bede61bceeeb2

pkg/controller/common/annotation/client_cert_test.go
pkoutsovasilis: changed in 084be00d91a83bdc1a81fdd87b929f9931b73455

pkg/controller/common/certificates/client_cert_reconcile.go
pkoutsovasilis: done in 023a545d55937070510543b4f5a683286420e788

pkg/utils/k8s/k8sutils.go
pkoutsovasilis: done in 023a545d55937070510543b4f5a683286420e788

pkg/controller/common/certificates/client_cert_reconcile.go
pkoutsovasilis: changed in 0aac18547f0e3e24c3af16a9bca76a38d89f6b9f

test/e2e/es/client_auth_test.go
pkoutsovasilis: So I realised that TestClientAuthRequired is essentially a subset of TestClientAuthRequiredTransition e2e test. Thus I have removed the former and re-structured a bit the latter to improve readability in eb5646000f938aec53ca615ed9409d1d05f1cb1a

pkg/utils/k8s/k8sutils.go
pkoutsovasilis: good point! went with the delegation case, just to maintain the intend of the caller to delete a secret in 7557e30f3459a8408504c878a8c0f9b3577770bc. Lemme know if you prefer removing completely DeleteSecretIfExists

pkg/controller/elasticsearch/validation/validations.go
pkoutsovasilis: IIUC I don't have a raw config path check here so I interpreted this comment as a request for adding a doc explaining just that 9b38842db904e7cff6705c54dfb844e827484a46

pkg/controller/elasticsearch/validation/validations.go
pkoutsovasilis: fixed in 9b38842db904e7cff6705c54dfb844e827484a46

pkg/controller/elasticsearch/driver/shared/reconcile.go
pkoutsovasilis: makes sense, changed in a62f75beb2247c351c13b36fe906c0e37e422dc5

pkg/controller/common/certificates/client_cert_reconcile.go
pkoutsovasilis: nice catch! fixed in e4cf7319e61d5d59740bbb54a51f7e42fe76e4f5

pkg/controller/elasticsearch/driver/stateful/driver.go
pkoutsovasilis: yep in my original thinking I was trying to avoid causing a roll out due to disabling the ECK-managed client certificates so I paused the world, but I see your point and I agree. Changed the behaviour here 12cfef259cbcc64811e8652a7dbf89f314272138

pkg/controller/elasticsearch/settings/merged_config.go
pkoutsovasilis: makes sense, changed in 6300ea5400bd0dfc509542665065287fea4f5a89

pkg/controller/common/annotation/client_cert.go
```diff
@@ -0,0 +1,61 @@
...
+// No-op if the annotation is already present.
+func SetClientAuthenticationRequiredAnnotation(ctx context.Context, k8sClient k8s.Client, obj client.Object) error {
+	if HasClientAuthenticationRequired(obj) {
+		return nil // already set correctly
+	}
+
+	ulog.FromContext(ctx).Info(
+		"Setting client-authentication-required annotation",
+		"namespace", obj.GetNamespace(),
+		"name", obj.GetName(),
```
barkbay: Duplicated fields:

```json
{"service.version": "0.0.0-SNAPSHOT+00000000", "iteration": "312", "namespace": "default", "es_name": "test-mtls2", "namespace": "default", "name": "test-mtls2"}
```

```suggestion
```

pkg/controller/common/certificates/client_cert_reconcile.go
```diff
@@ -0,0 +1,423 @@
...
+		Name:      secretName,
+	}
+
+	// Build the expected secret with correct metadata
+	expected := corev1.Secret{
+		ObjectMeta: k8s.ToObjectMeta(secretNSN),
+		Data:       make(map[string][]byte),
+	}
+	expected.Labels = utilmaps.Merge(maps.Clone(r.Metadata.Labels), extraLabels)
+	expected.Annotations = r.Metadata.Annotations
```
barkbay: This assigns pointers, not maps. If anything downstream mutates `expected.Annotations`, it also mutates `r.Metadata.Annotations`, should we clone this?

pkg/controller/common/certificates/client_cert_reconcile.go
```diff
@@ -0,0 +1,423 @@
...
+			strings.Compare(a.Name, b.Name),
+		)
+	})
+
+	var buf bytes.Buffer
+	for _, secret := range secrets {
+		certData := secret.Data[CertFileName]
+		if len(certData) == 0 {
+			log.V(1).Info("Skipping secret with no certificate data",
+				"namespace", secret.Namespace, "name", secret.Name)
```
barkbay: `namespace` is duplicated

```json
{"service.version": "0.0.0-SNAPSHOT+00000000", "iteration": "1", "namespace": "default", "es_name": "test-mtls-usercert", "namespace": "default", "name": "test-mtls-usercert-es-internal-client-cert"}
```

I think we need to check all the places where `namespace` is set again.

pkg/controller/common/certificates/client_cert_reconcile.go
```diff
@@ -0,0 +1,423 @@
...
+			continue
+		}
+		buf.Write(certData)
+		if len(certData) > 0 && certData[len(certData)-1] != '\n' {
```
barkbay: If we are here it means `certData` is not empty I think?
```suggestion
		if certData[len(certData)-1] != '\n' {
```

pkg/controller/common/certificates/client_cert_reconcile.go
```diff
@@ -0,0 +1,423 @@
...
+		log.V(1).Info("Certificate CommonName mismatch, will re-issue",
+			"expected", commonName, "actual", cert.Subject.CommonName)
+		return nil
+	}
+
+	if !CertIsValid(ctx, *cert, certReconcileBefore) {
+		return nil
+	}
+
+	privateKey := privateKeyIfValid(ctx, secret)
```
barkbay: nit: we already did this line 217

pkg/controller/common/certificates/client_cert_reconcile.go
```diff
@@ -0,0 +1,423 @@
...
+	publicKey crypto.PublicKey,
+	validity time.Duration,
+) *ValidatedCertificateTemplate {
+	template := ValidatedCertificateTemplate(x509.Certificate{
+		Subject: pkix.Name{
+			CommonName:         commonName,
+			OrganizationalUnit: []string{orgUnit},
+		},
+		NotBefore: time.Now().Add(-10 * time.Minute),
+		NotAfter:  time.Now().Add(validity),
```
barkbay, line 352: IIRC best practice for client certs is to include `BasicConstraintsValid: true` and `IsCA: false`
```suggestion
		NotAfter:  time.Now().Add(validity),
		BasicConstraintsValid: true,
		IsCA:                  false,
```

pkg/controller/common/certificates/client_cert_reconcile.go
```diff
@@ -0,0 +1,423 @@
...
+}
+
+// clientCertificateIfValid returns the client certificate from the provided Secret if it's valid.
+// Returns nil if the certificate needs to be re-issued.
+func clientCertificateIfValid(
+	ctx context.Context,
+	secret *corev1.Secret,
+	commonName string,
+	certReconcileBefore time.Duration,
+) []byte {
```
barkbay: Could we just return a boolean?

pkg/controller/elasticsearch/certificates/reconcile.go
```diff
@@ -78,7 +80,71 @@ func ReconcileHTTP(
...
+		K8sClient:    driver.K8sClient(),
+		Owner:        es,
+		Namer:        esv1.ESNamer,
+		Metadata:     meta,
+		CertRotation: certRotation,
+	}
+
+	operatorClientCertSecret, err := certReconciler.ReconcileOperatorCertificate(ctx)
+	if err != nil {
+		k8s.MaybeEmitErrorEvent(driver.Recorder(), err, es, events.EventReconciliationError, "Client certificate reconciliation error: %v", err)
```
barkbay: Any reason for generating an event here and not when handling other errors?

pkg/controller/elasticsearch/driver/shared/reconcile.go
```diff
@@ -77,7 +79,7 @@ func ReconcileSharedResources(
 	meta := metadata.Propagate(&es, metadata.Metadata{Labels: label.NewLabels(k8s.ExtractNamespacedName(&es))})
 
 	// Reconcile the scripts ConfigMap.
-	if err := configmap.ReconcileScriptsConfigMap(ctx, client, es, meta); err != nil {
+	if err := configmap.ReconcileScriptsConfigMap(ctx, client, es, meta, clientAuthenticationRequired); err != nil {
```
barkbay: Should we keep the client cert flags in the pre-stop hook script as long as the annotation is present, rather than tying it to the current desired state? During the disable transition, kubelet can hot-reload the updated `ConfigMap` (without client cert flags) into old pods before they're terminated. Their pre-stop hook then would fail against old nodes that still require client auth?

```suggestion
	// Keep client cert flags in the pre-stop hook script during disable transitions:
	// old pods (still requiring client auth) may get the updated ConfigMap via kubelet
	// volume refresh before being terminated. Without client cert flags, their pre-stop
	// hook would fail against old nodes that still require client authentication.
	scriptClientAuth := clientAuthenticationRequired || annotation.HasClientAuthenticationRequired(&es)
	if err := configmap.ReconcileScriptsConfigMap(ctx, client, es, meta, scriptClientAuth); err != nil {
```

pkg/controller/elasticsearch/certificates/reconcile.go
pkoutsovasilis: valid point moved the event emitting to the caller here e572c3061b642fd0c8cd5f396b2f7426f861eebe

pkg/controller/common/certificates/client_cert_reconcile.go
pkoutsovasilis: yep we can and it is done in 8e46a5080ad91eee9e5f06b2e3d92a4c1f35fa97

pkg/controller/common/certificates/client_cert_reconcile.go
pkoutsovasilis: valid point, I re-worked this in 8e46a5080ad91eee9e5f06b2e3d92a4c1f35fa97

pkg/controller/common/certificates/client_cert_reconcile.go
pkoutsovasilis: changed in ee8112f9cd25d92072823d0d7c07c9f3c6500943

pkg/controller/common/certificates/client_cert_reconcile.go
pkoutsovasilis: nice catch! I checked all of them I think and did the necessary changes in 59bd7b49cd11137e77422df0d77636e9127495f1

pkg/controller/common/annotation/client_cert.go
pkoutsovasilis: nice catch! changed in 59bd7b49cd11137e77422df0d77636e9127495f1

pkg/controller/common/certificates/client_cert_reconcile.go
pkoutsovasilis: cloned the maps in 510d21503bd4f6b1dbf93979b7e55ffc772601e0

pkg/controller/common/certificates/client_cert_reconcile.go
pkoutsovasilis: always happy to follow best practices, changed in 988b21e6c81280cd12ec23451dfb782f5c938fce

pkg/controller/elasticsearch/driver/shared/reconcile.go
pkoutsovasilis: Initially I had classified this as the same kind of an edge-case that exists for protocol changes (http↔https), documented at lifecycle_hook.go:239-241. But your point is valid and unlike the protocol change case, here we can actually eliminate the race without trade-offs.

That said, I believe that the proposal (scriptClientAuth := clientAuthenticationRequired || annotation.HasClientAuthenticationRequired(&es)) would cause an extra rolling restart on the disable transition:

  1. User disables client auth → clientAuthenticationRequired=false, but annotation still present → script keeps client cert flags → ConfigMap unchanged
  2. Pods roll due to ES config change (removing xpack.security.http.ssl.client_authentication: required)
  3. All pods rolled → DeleteClientCertResources removes annotation → script drops client cert flags → ConfigMap hash changes → second rolling restart

So I resorted to orchestration hints in 3245a10c8c6b4f8226e88b8be06bce47bc3520b5:
  - Added a client_certificate_in_scripts hint that controls whether the pre-stop hook script includes client certificate flags.
  - For new clusters (no existing hints annotation), the hint defaults to true - the script includes client cert flags from the start, so the race never applies.
  - For existing clusters, the hint is set to true when client authentication is first enabled, and is never cleared. This means the script retains client cert flags even after client auth is disabled, avoiding the race without triggering an extra rolling restart.
  - The file-existence checks in the script ([ -f "..." ]) make the flags a no-op when certificate files are not mounted, so there's no functional impact on clusters that don't use client authentication.

Lemme know what you think

pkoutsovasilis:
> * **Client auth + TLS disabled**: ES ignores `client_authentication: required` when `ssl.enabled: false`, but the operator still creates unnecessary cert resources — might be worth a validation check but ✅

good point @barkbay, I did add a validation check and made the operator to skip creating unnecessary cert resources in c7e6efe20497a6a9102efbff271ec8025305673c

pkg/controller/elasticsearch/driver/shared/reconcile.go
pkoutsovasilis: I didn't like the previous way of detecting new cluster so I did another follow up change to use established code patterns here ae7360b9d0bc50f7107e6a958b324f580d8d34fb 🙂

pkg/controller/elasticsearch/driver/shared/reconcile.go
```diff
@@ -77,8 +80,22 @@ func ReconcileSharedResources(
...
+	// Set the client certificate in scripts orchestration hint so the pre-stop hook script permanently
+	// includes client certificate flags. This avoids a race condition during the disable transition
+	// where kubelet may hot-reload the updated ConfigMap (without client cert flags) into old pods
+	// before they are terminated. The file-existence checks in the script make the flags a no-op
+	// when certificate files are not mounted.
+	// For new clusters (not yet bootstrapped), always set the hint. For existing clusters, set it
+	// only when client certificate authentication is enabled.
+	if clientAuthenticationRequired || !bootstrap.AnnotatedForBootstrap(es) {
+		params.ReconcileState.UpdateOrchestrationHints(
+			params.ReconcileState.OrchestrationHints().Merge(hints.OrchestrationsHints{ClientCertificateInScripts: true}),
```
pebrc: The merging here seems redundant? `UpdateOrchestrationHints` already does that?

pkg/controller/elasticsearch/hints/hints.go
```diff
@@ -26,14 +26,21 @@ type OrchestrationsHints struct {
 	// controllers should then rely on regular users until the value is true.
 	ServiceAccounts *optional.Bool    `json:"service_accounts,omitempty"`
 	DesiredNodes    *DesiredNodesHint `json:"desired_nodes,omitempty"`
+
+	// ClientCertificateInScripts controls whether the pre-stop hook script includes client certificate
+	// flags (with file-existence fallbacks) to avoid a race condition during the disable transition where
+	// kubelet may hot-reload the updated ConfigMap before pods are terminated. Defaults to true for new clusters.
+	// For existing clusters, it is set to true only when client certificate authentication is enabled.
+	ClientCertificateInScripts bool `json:"client_certificate_in_scripts,omitempty"`
```
pebrc: The comment conflates two distinct reasons for this hint:

1. **Why the flag is sticky** (once `true`, never cleared via OR-based `Merge`): to avoid a race condition during the disable transition where kubelet may hot-reload the updated ConfigMap into old pods before they are terminated. The `[ -f ... ]` guards in the script make the flags a no-op when cert files are not mounted, so keeping them is safe.

2. **Why existing clusters default to `false`**: to avoid a fleet-wide rolling restart on operator upgrade. If existing clusters (which have the hints annotation but not this field) defaulted to `true`, every cluster would see a ConfigMap hash change on the first reconciliation with the new operator. The per-cluster promotion to `true` (when mTLS is first enabled) is what limits the restart to only clusters that actually use the feature.

The current comment only explains #1. A reader would reasonably ask: *"if the script is safe either way due to file-existence checks, why not just default to `true` globally?"* — and get no answer.

pebrc: nit: the name `ClientCertificateInScripts` describes an implementation detail (how the pre-stop hook script is rendered) rather than a cluster state, which is inconsistent with the other hints. `NoTransientSettings`, `ServiceAccounts`, and `DesiredNodes` all describe cluster capability or feature state — consumers decide what to do with that state. `ClientCertificateInScripts` conflates the state with the action.

There is only one consumer (`RenderPreStopHookScript`) so the leak is contained, but a more semantic name would survive a future refactor better. Some alternatives:

- `MTLSOnboarded` — has this cluster ever gone through mTLS enablement
- `ClientAuthEverEnabled` — explicit about the one-way nature
- `MTLSAware` — the cluster's scripts should be treated as mTLS-aware

pebrc: I wonder if we need this optimization if https://github.com/elastic/cloud-on-k8s/pull/9012 will trigger a rolling restart for 3.4 for many/most  ES clusters (> 8.0 that have `set-default-security-context` to `auto-detect` or `true` I believe). That means we could piggy back this restart onto that and avoid the restart avoidance logic.

Leaving this out would simplify the code quite a bit imo. Thoughs @barkbay @pkoutsovasilis ?

test/e2e/es/client_auth_test.go
```diff
@@ -0,0 +1,212 @@
...
+				certs, err := certificates.ParsePEMCerts(secret.Data[certificates.CertFileName])
+				if err != nil {
+					return err
+				}
+				if len(certs) == 0 {
+					return fmt.Errorf("no certificates found in client cert secret")
+				}
+				cert := certs[0]
+				require.Contains(t, cert.ExtKeyUsage, x509.ExtKeyUsageClientAuth)
+				require.Equal(t, cert.Subject.CommonName, cert.Issuer.CommonName)
```
pebrc: require.Contains and require.Equal inside the Eventually callback will call t.FailNow() on first failure, aborting immediately instead of allowing Eventually to retry. These should use assert (which returns a bool) or return an error, consistent with the Eventually retry pattern.

pkg/controller/elasticsearch/hints/hints.go
pkoutsovasilis: yep I saw this one https://github.com/elastic/cloud-on-k8s/pull/9012 and I was thinking the same, but I can't quantify how small the impact would be for users that don't have set-default-security-context to auto-detect or true (I believe we don't support  < 8.0 in the new releases). If you say it's minimal then, echoing the same message, yes I vote to go without the orchestration hints as it would simplify a lot the code here

pkoutsovasilis: I struggled a bit with this name 😅 I am gonna wait for a decision on [this one](https://github.com/elastic/cloud-on-k8s/pull/9229#discussion_r3008834186) and if we stay with orchestration hints I am gonna go with `MTLSAware`

test/e2e/es/client_auth_test.go
pkoutsovasilis: nice catch! changed in 6a799cff24e0c71cbf75849ac35b192c9be75dad

pkg/controller/elasticsearch/driver/shared/reconcile.go
pkoutsovasilis: good point! fixed in 4ac32ed220681f8ba44c457aebbaca56eff30978

pkg/controller/elasticsearch/hints/hints.go
pkoutsovasilis: Updated the comment to cover both cases in c6e29e74897a801bef7e47e0d80052fa9c59eb40. On OpenShift, `set-default-security-context` auto-detects to false, so #9012 wouldn't trigger a rolling restart there - those users would get an unexpected restart if we dropped the hints. So keeping the orchestration hints seems as the safer option. @barkbay thoughts so we can move forward?

pkoutsovasilis: renamed in c6e29e74897a801bef7e47e0d80052fa9c59eb40

barkbay: I think #9012 will already trigger a rolling restart for most ES clusters on 3.4. I'd rather piggyback on that restart, than carry the extra orchestration complexity.

pebrc: Sounds good to me too. Lets accept a restart for 3.4

pkoutsovasilis: ok I changed it in ab5332e0fcc1b6b37bedb3f647d9b0d41f1f7c21 and now we always render the pre-stop hook script with the client certificates

pkoutsovasilis:
@pebrc @barkbay is there something else holding this PR from merging?

barkbay (approved):
LGTM

pkg/controller/common/certificates/client_cert_reconcile.go
```diff
@@ -0,0 +1,421 @@
...
+		return nil
+	}
+	key, err := ParsePEMPrivateKey(keyPEM)
+	if err != nil {
+		ulog.FromContext(ctx).Error(err, "Failed to parse private key",
+			"secret_namespace", secret.Namespace, "secret_name", secret.Name)
+		return nil
+	}
+	return key
+}
```
pebrc, line 320: The annotation is removed *after* secrets are deleted. If annotation removal fails (transient API error) after secrets are already gone, the next reconcile sees the annotation, enters the client-cert reconciliation path, and recreates the secrets — which is self-healing. However, the reverse is more problematic: the annotation incorrectly signals to association controllers that client certs are required when the resources backing that contract have been removed.

Swapping the order (remove annotation first, then delete secrets) is safer: orphaned secrets are harmless and will be GC'd via owner references, but an orphaned annotation actively misleads other controllers.

This was also flagged in earlier manual testing ("Cleanup ordering bug confirmed — annotation removal before secret deletion causes orphaned secrets on transient failures").

pkg/controller/elasticsearch/driver/stateful/config.go
```diff
@@ -0,0 +1,166 @@
...
+			es.Name, ver, ipFamily, es.Spec.HTTP, userCfg, policyConfig.ElasticsearchConfig,
+			es.Spec.RemoteClusterServer.Enabled, es.HasRemoteClusterAPIKey(), clusterHasZoneAwareness,
+			false,
+		)
+		if err != nil {
+			return false, "", err
+		}
+		if settings.HasClientAuthenticationRequired(cfg) {
+			return true, overrideWarning, nil
+		}
```
pebrc, line 121: nit: If NodeSet 0 has `client_authentication: required` in its raw config, the function returns `true` immediately without checking remaining NodeSets for override warnings. A NodeSet 1 with a per-NodeSet override to `optional` would have its warning silently swallowed. Unlikely in practice (mixed per-NodeSet client auth settings are unusual) but the behavior is surprising.

pkg/controller/elasticsearch/settings/merged_config_test.go
```diff
@@ -303,9 +417,162 @@ func TestNewMergedESConfig(t *testing.T) {
...
+			name: "key absent returns false",
+			cfg:  map[string]any{},
+			want: false,
+		},
+	} {
+		t.Run(tt.name, func(t *testing.T) {
+			cfg := CanonicalConfig{common.MustCanonicalConfig(tt.cfg)}
+			require.Equal(t, tt.want, HasClientAuthenticationRequired(cfg))
+		})
+	}
```
pebrc, line 585: nit: `HasClientAuthenticationRequired` has an explicit `ssl.enabled=false` early-return path (line 206 in `merged_config.go`) that isn't directly tested here. A one-line table entry like `{name: "required but ssl disabled", cfg: map[string]any{XPackSecurityHttpSslClientAuthentication: "required", XPackSecurityHttpSslEnabled: "false"}, want: false}` would close the gap.

pkg/controller/elasticsearch/driver/stateful/config.go
pkoutsovasilis: the "ineffective" warning wouldn't be accurate in the mixed case: if NodeSet 0 has required and NodeSet 1 overrides to optional, the spec is still effective for NodeSet 0 and the operator still needs to create client certificates. The warning saying "spec is ineffective" would be misleading. I think this edge case (mixed per-NodeSet client auth settings within the same cluster) is unusual enough that we can leave it for now - if we want to handle it properly, a validation error rejecting the misconfiguration would be more appropriate than a warning, wdyt?

pkg/controller/elasticsearch/settings/merged_config_test.go
pkoutsovasilis: gap closed in 4cb12548110194c1e18bebcdced7863503a7c655

pkg/controller/common/certificates/client_cert_reconcile.go
pkoutsovasilis: ok reverted the order here 49a16e04e739525c2511b78dd847d1d9b01a4e7c, however from this [previous comment](https://github.com/elastic/cloud-on-k8s/pull/9229#discussion_r2969437901) I got the feeling that leaving orphaned secrets is more critical

PR #9127, opened 2026-02-10
[WIP] Ensure fleet outputs block is added when fleet is enabled.

naemono:
Resolves: https://github.com/elastic/cloud-on-k8s/issues/9112

## Summary

Configuring Fleet-Server with ECK has historically been difficult for the end-user with some edge cases that, without reading through all of the documentation, will cause the deployment to fail. Specifically trusting the necessary Certificate Authorities when using self-signed certificates (which ECK does by default) has not been simple with the different ways that Agent/Fleet-Server can be configured (root/non-root/etc).

### Current State

If Agent/Fleet-server is run as root, we inject the CA directly into the containers OS-level CA trust, which allows the CA to be trusted explicitly. When running as non-root, which is required in many secure environments, one must add the following configuration block in Kibana, or Fleet-server will fail ([docs reference](https://www.elastic.co/docs/deploy-manage/deploy/cloud-on-k8s/configuration-fleet#k8s-elastic-agent-running-as-a-non-root-user)):

```yaml
xpack.fleet.outputs:
- id: eck-fleet-agent-output-elasticsearch
  is_default: true
  name: eck-elasticsearch
  type: elasticsearch
  hosts:
  - "https://elasticsearch-es-http.default.svc:9200"
    ssl:
      certificate_authorities: ["/mnt/elastic-internal/elasticsearch-association/default/elasticsearch/certs/ca.crt"]
```

In addition, the following field must *not* exist in the Kibana configuration when this outputs block exists:

`xpack.fleet.agents.elasticsearch.hosts`

### Future State

This change is an attempt to do the following
1. Remove the complexity of dealing with an init container modifying the root CA store of the Fleet-Server container.
2. Automatically inject the `xpack.fleet.outputs` block when appropriate
3. Ensure the `xpack.fleet-agents.elasticsearch.hosts` field is missing when the `outputs` block exists.

This will allow the following Kibana configuration block to "just work"

```yaml
apiVersion: kibana.k8s.elastic.co/v1
kind: Kibana
metadata:
  name: kibana
spec:
  version: 9.3.0
  count: 1
  elasticsearchRef:
    name: elasticsearch
  config:
    xpack.fleet.agents.fleet_server.hosts: ["https://fleet-server-agent-http.default.svc:8220"]
    xpack.fleet.packages:
... cut ...
    xpack.fleet.agentPolicies:
... cut ...
---

```

#### Manual

- [ ] Manual testing - in progress

```yaml
# Kibana spec.config
[...]

PR #8917, opened 2025-11-20
feat: multi SCP composition

pkoutsovasilis:
## Overview

This PR implements support for multiple StackConfigPolicies (SCPs) targeting the same Elasticsearch cluster or Kibana instance, using a weight-based priority system for deterministic policy composition.

### Weight-Based Priority System
- Policies are merged in order of weight (lower weight takes precedence)
- Default weight: `0`

### Conflict Detection

Conflicts are detected across multiple dimensions and will prevent policy application:

| Conflict Type | Condition                                                                        |  Result |
|---------------|----------------------------------------------------------------------------------|--------|
| **Weight Conflict** | Two or more policies with identical weights target the same Elasticsearch/Kibana |  ❌ Conflict |
| **SecretMount Name Conflict** | Different policies define `SecretMount` with same `SecretName`                   |  ❌ Conflict |
| **SecretMount Path Conflict** | Different policies define `SecretMount` with same `MountPath`                    |  ❌ Conflict |
| **Different Weights** | Policies have different weights and none of the above applies                    |  ✅ Pass - lower weight wins |

**Important:** Even if two policies with the same weight have non-overlapping resources, they still conflict because the weight collision makes the merge order ambiguous.

### Configuration Merging Behaviour

Different merge strategies are applied based on the configuration type:

- **Deep Merge** (recursive merging):
    - `ClusterSettings`
    - `Config`
    - `SnapshotLifecyclePolicies`
    - `SecurityRoleMappings`
    - `IndexLifecyclePolicies`
    - `IngestPipelines`
    - `IndexTemplates.ComposableIndexTemplates`
    - `IndexTemplates.ComponentTemplates`

- **Top-Level Key Replacement** (entire keys replaced):
    - `SnapshotRepositories` - each repository configuration is treated atomically

- **Union Merge** (with conflict detection):
    - `SecretMounts` - conflicts on duplicate `SecretName` OR duplicate `MountPath`
    - `SecureSettings` - merges by `SecretName`+`Key`, lower weight wins (no conflicts)

### Multi-Soft-Owner Secret Management

**File Settings and Policy Config Secrets:**
- Now support **multiple soft owners**
- Secrets are only deleted when all referencing soft-owners are removed
- Uses `eck.k8s.elastic.co/owner-refs` annotation with JSON-encoded map of owner namespaced names

**Secret Sources:**
- Remain **single soft owner** (existing behaviour unchanged)

This prevents secret leakage while enabling proper cleanup when policies are deleted.

## Related Issues

[...]

PR #9234, opened 2026-03-16
Implement fleet-server and elastic-agent support for presenting client certificates to elasticsearch

pkoutsovasilis:
## Summary

This PR implements client certificate support for Fleet Server and Elastic Agent (both fleet-managed and standalone) when connecting to an Elasticsearch that has client authentication enabled. This PR handles direct ES associations and transitive associations (fleet-managed agents connecting to ES through Fleet Server).

Relates to https://github.com/elastic/cloud-on-k8s/issues/9081

### Changes

  - **Agent `ElasticsearchSelector`**: Changes `Output.ObjectSelector` to `Output.ElasticsearchSelector`, adding `clientCertificateSecretName` support to Agent's `elasticsearchRefs`
  - **`TransitiveESRef` in `AssociationConf`**: New struct to propagate transitive Elasticsearch association state (e.g., the client cert secret name for the ES that Fleet Server connects to) through to fleet-managed agents
  - **Fleet Server association controller**: Extended to reconcile client certificates for the transitive ES association - when a fleet-managed agent associates with a Fleet Server, the controller looks up the Fleet Server's ES association conf and, if client authentication is required, creates a client certificate secret in the agent's namespace
  - **Orphaned secret cleanup**: Deletes stale transitive client certificate secrets when the Fleet Server's ES association changes or client authentication is disabled
  - **Standalone Agent pod spec**: Mounts client certificate volumes and configures `ELASTICSEARCH_CERT` / `ELASTICSEARCH_CERT_KEY` environment variables for standalone agents with direct ES associations
  - **Fleet-managed Agent pod spec**: Mounts the transitive client certificate secret and configures `FLEET_SERVER_ELASTICSEARCH_CERT` / `FLEET_SERVER_ELASTICSEARCH_CERT_KEY` for Fleet Server pods, and client cert volumes for fleet-managed agents
  - **Fleet Server env vars**: Extracted ES config population into a helper function (`populateFleetServerESConfig`) to reduce nesting complexity
  - **Kibana Fleet output injection**: Injects `ssl.certificate` and `ssl.key` into Kibana's `xpack.fleet.outputs` for fleet-managed agent policies when mTLS is enabled. Similarly it does the same for `ssl.certificate_authorities` when a CA is provided.

### API

```yaml
  kind: Agent
  spec:
    elasticsearchRefs:
      - name: elasticsearch
        outputName: default
        # Optional: use a custom client certificate
        clientCertificateSecretName: my-custom-client-cert
```


prodsecmachine:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/63e804f9-9ed7-4f9d-bcbd-211b3c9e5e02?refs=aa36b6261b65a28da684c32a17e04f352dc50da6&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/63e804f9-9ed7-4f9d-bcbd-211b3c9e5e02/license?refs=aa36b6261b65a28da684c32a17e04f352dc50da6&source=prChecksComment) |

pkoutsovasilis:
buildkite test this -f p=gke,t=TestClientAuthRequired.*,E2E_TAGS=agent -m s=9.3.1,s=8.19.2,s=9.2.6

pkoutsovasilis:
buildkite test this -f p=gke,t=TestClientAuthRequired.*,E2E_TAGS=agent -m s=9.3.2,s=8.19.13,s=9.2.7

pkoutsovasilis:
buildkite test this -f p=gke,t=TestClientAuthRequired.*,E2E_TAGS=agent -m s=9.3.2,s=8.19.13,s=9.2.7

pkoutsovasilis:
buildkite test this -f p=gke,t=TestClientAuthRequired.*,E2E_TAGS=agent -m s=9.3.3,s=8.19.14,s=9.2.8

pkg/controller/association/controller/agent_fleetserver.go
```diff
@@ -80,14 +93,160 @@ func additionalSecrets(ctx context.Context, c k8s.Client, assoc commonv1.Associa
...
+	}
+	if conf == nil {
+		log.V(1).Info("Transitive ES association conf is nil")
+		if err := deleteOrphanedTransitiveClientCertSecrets(ctx, c, assocMeta, associated, ""); err != nil {
+			return nil, results.WithError(err)
+		}
+		return nil, nil
+	}
+
+	ref := esAssociation.AssociationRef()
```
kvalliyurnatt, line 180: please ignore if this does not make sense, will this ref be a namespaced Name of Elasticsearch object or can it also return a secret (external reference, unmanaged Elasticsearch connection secret) ? asking because below we assume this returns an Elasticsearch object

pkg/controller/association/controller/agent_fleetserver.go
```diff
@@ -80,14 +93,160 @@ func additionalSecrets(ctx context.Context, c k8s.Client, assoc commonv1.Associa
...
+
+	var es esv1.Elasticsearch
+	if err := c.Get(ctx, ref.NamespacedName(), &es); err != nil {
+		return nil, results.WithError(err)
+	}
```
pebrc, line 193: Echoing @kvalliyurnatt's question — `ref` here comes from `esAssociation.AssociationRef()`. If the Fleet Server's ES ref is an external reference (using `secretName` instead of `name`), `ref.NamespacedName()` points to the connection secret, not an Elasticsearch object. The `c.Get` on line 181 would fail with NotFound, causing the reconciler to return an error and requeue indefinitely.

Should this check `ref.IsExternal()` and either skip client cert reconciliation for external refs (since the operator can't know whether the external ES requires client auth) or handle it differently?

pkg/controller/association/controller/agent_fleetserver.go
```diff
@@ -80,14 +93,160 @@ func additionalSecrets(ctx context.Context, c k8s.Client, assoc commonv1.Associa
...
+
+	secrets := make([]types.NamespacedName, 0, 2)
+	if conf.CACertProvided {
+		log.V(1).Info("additional secret because CA provided")
+		secrets = append(secrets, types.NamespacedName{
+			Namespace: fleetServer.Namespace,
+			Name:      conf.CASecretName,
+		})
+	}
+	if conf.ClientCertIsConfigured() && fleetServer.Spec.ElasticsearchRefs[0].GetClientCertificateSecretName() != "" {
```
pebrc: Nit: `fleetServer.Spec.ElasticsearchRefs[0]` is safe here because of the `len == 0` guard at line 73, but the guard and this access are ~35 lines apart. The same pattern appears at line 194. Consider extracting the ES ref into a local variable near the guard to make the safety more obvious, especially since this function is growing in complexity.

pkg/controller/agent/pod.go
```diff
@@ -85,9 +85,15 @@ const (
 	FleetServerElasticsearchUsername = "FLEET_SERVER_ELASTICSEARCH_USERNAME"
 	FleetServerElasticsearchPassword = "FLEET_SERVER_ELASTICSEARCH_PASSWORD" //nolint:gosec
 	FleetServerElasticsearchCA       = "FLEET_SERVER_ELASTICSEARCH_CA"
+	FleetServerElasticsearchCert     = "FLEET_SERVER_ES_CERT"
+	FleetServerElasticsearchCertKey  = "FLEET_SERVER_ES_CERT_KEY"
```
pebrc: Nit: the Go constant names (`FleetServerElasticsearchCert` / `FleetServerElasticsearchCertKey`) follow the `FleetServerElasticsearch*` pattern of the neighboring constants, but the actual env var values are `FLEET_SERVER_ES_CERT` / `FLEET_SERVER_ES_CERT_KEY` — note `ES` vs `ELASTICSEARCH`. All the other constants in this block have the Go name mirror the env var name (e.g., `FleetServerElasticsearchCA` → `FLEET_SERVER_ELASTICSEARCH_CA`). Consider `FleetServerESCert` / `FleetServerESCertKey` to match the values, or keep as-is with a short comment noting the naming difference.

pkg/controller/association/controller/agent_fleetserver.go
```diff
@@ -80,14 +93,160 @@ func additionalSecrets(ctx context.Context, c k8s.Client, assoc commonv1.Associa
...
+	return secrets, nil
+}
+
+// clientCertSecretName returns a deterministic secret name for a transitive client certificate.
+// The name includes a hash of the referenced resource's namespace and name to avoid collisions.
+func clientCertSecretName(associated commonv1.Associated, ref commonv1.AssociationRef, associationName string) string {
+	associatedName := associated.GetName()
+	refHash := hash.HashObject(ref.NamespacedName())
+	return associatedName + "-" + associationName + "-" + refHash + "-client-cert"
+}
```
pebrc, line 126: This function constructs the secret name via raw string concatenation rather than using the `Namer` tooling used elsewhere in the codebase (e.g., `OperatorClientCertSecretName` goes through `namer.Suffix()`). The `Namer` provides truncation safety and logging when names exceed Kubernetes limits.

In practice this is safe today — worst case is ~68 chars (`36 + 1 + 8 + 1 + 10 + 12`), well under the 253-char secret name limit — but it silently depends on the webhook enforcing `MaxResourceNameLength = 36` for Agent names. The same pattern exists in `association/client_cert.go:38` which predates this PR, so this isn't new, but worth noting.

pkg/controller/kibana/config_settings.go
```diff
@@ -401,3 +413,97 @@ func packageRegistrySettings(kb kbv1.Kibana) map[string]any {
...
+		if has, _ := output.Has("ssl.key", -1, opts...); !has {
+			if err := output.SetString("ssl.key", -1, path.Join(agent.FleetManagedAgentClientCertDir,
+				certificates.KeyFileName), opts...); err != nil {
+				return err
+			}
+		}
+	}
+
+	return nil //nolint:nilerr
+}
```
pebrc, line 457: suggestion (style): Both `injectFleetOutputClientCerts` and `injectFleetOutputCertificateAuthorities` use raw `(*ucfg.Config)` pointer casts with manual `Child()`/`Has()`/`SetString()` calls. This is the only place in `config_settings.go` that manipulates ucfg at this level — every other config section uses the idiomatic pattern of typed Go structs + `Unpack`/`MustCanonicalConfig`/`MergeWith`.

Consider aligning with the existing pattern, e.g.:

```go
type fleetOutputSSL struct {
    Certificate            string   `config:"certificate,omitempty"`
    Key                    string   `config:"key,omitempty"`
    CertificateAuthorities []string `config:"certificate_authorities,omitempty"`
}

type fleetOutput struct {
    Type  string          `config:"type"`
    Hosts []string        `config:"hosts"`
    SSL   *fleetOutputSSL `config:"ssl,omitempty"`
}

type fleetOutputsWrapper struct {
    Outputs []fleetOutput `config:"xpack.fleet.outputs"`
}
```

Then unpack, mutate, merge back. This would also make the URL matching more obvious and testable on plain Go structs, and would be consistent with the approach PR #9127 already takes for its fleet output structs.

pebrc (dismissed):
The PR adds a well-structured mTLS client certificate system for fleet-server and elastic-agent connections to Elasticsearch. The core certificate lifecycle (generation, rotation, trust bundle assembly, cleanup) builds cleanly on top of the existing `certificates.Reconciler` infrastructure. The transitive association chain (Agent → Fleet Server → ES) is handled through a clear callback mechanism (`ReconcileTransitiveESSecrets`) that avoids coupling the generic association reconciler to agent-specific logic.

Integration testing on a live cluster confirmed all three main paths work correctly: standalone agent with managed certs, fleet-managed agent with transitive cert propagation + Kibana ssl injection, and clean teardown when client auth is disabled.

**Key suggestions:**

1. **External ES ref safety** — `fleetManagedAgentTransitiveESRef` does a `c.Get` on the ES `NamespacedName` without checking `ref.IsExternal()`, which would fail for external ES references. (medium)

2. **Kibana config injection style** — The `injectFleetOutput*` functions use raw ucfg manipulation rather than the typed struct Unpack/MergeWith pattern used everywhere else in `config_settings.go`. This works but is inconsistent and harder to maintain. (low)

3. **URL matching** — `outputHostsContain` uses exact string equality. A trailing slash or any other equivalent-but-not-identical URL form would silently skip injection with no log output. (low)

pkg/controller/kibana/config_settings.go
pkoutsovasilis: I don't think that the typed approach is gonna work out of the box here. Specifically when round-tripping through unpack → mutate → `MergeWith`, nil pointer fields leak as `null` and nil slices as `[]` in the rendered YAML. This is the same class of issue we hit in #8917 where [ucfg replaces empty maps with nil during Unpack](https://github.com/elastic/cloud-on-k8s/pull/8917#discussion_r2571905667), causing downstream config rejection.

It's worth noting that #9127 takes the typed struct approach but for constructing new outputs from scratch, whereas this PR needs to selectively inject fields into existing user-provided outputs. I think the typed struct approach is what we want in general, but given the challenges of utilizing ucfg without introducing errors when modifying existing configs, I propose deferring this to a follow-up and sticking with raw ucfg here. 

unit-test that proves the nil leaking 
```golang
func Test_setting(t *testing.T) {
	type SSL struct {
		Certificate            *string  `config:"certificate"`
		Key                    *string  `config:"key"`
		CertificateAuthorities []string `config:"certificate_authorities"`
	}

	type Output struct {
		Type  string         `config:"type"`
		Hosts []string       `config:"hosts"`
		SSL   *SSL           `config:"ssl"`
		Rest  map[string]any `config:",inline"`
	}

	type Wrapper struct {
		Outputs []Output `config:"xpack.fleet.outputs"`
	}

	// Start with a config that has extra fields (id, name, is_default)
	cfg := settings.MustCanonicalConfig(map[string]any{
		"xpack.fleet.outputs": []any{
			map[string]any{
				"type":                  "elasticsearch",
				"hosts":                 []any{"https://es.default.svc:9200"},
				"id":                    "eck-fleet-agent-output-elasticsearch",
				"name":                  "eck-elasticsearch",
				"is_default":            true,
				"is_default_monitoring": true,
			},
			map[string]any{
				"type":                  "elasticsearch",
				"hosts":                 []any{"https://es.default.svc:9200"},
				"id":                    "eck-fleet-agent-output-elasticsearch2",
				"name":                  "eck-elasticsearch",
				"is_default":            true,
				"is_default_monitoring": true,
			},
		},
	})

	// Unpack into typed struct
	var wrapper Wrapper
	require.NoError(t, cfg.Unpack(&wrapper))

	// Verify unknown fields are captured in Rest
	require.Equal(t, "elasticsearch", wrapper.Outputs[0].Type)
	require.Contains(t, wrapper.Outputs[0].Rest, "id")
	require.Contains(t, wrapper.Outputs[0].Rest, "name")
	require.Contains(t, wrapper.Outputs[0].Rest, "is_default")
	require.Contains(t, wrapper.Outputs[0].Rest, "is_default_monitoring")
	require.Nil(t, wrapper.Outputs[0].SSL)

	// Set only cert+key, leave CertificateAuthorities nil
	certPath := "/path/to/cert"
	_ = certPath
	keyPath := "/path/to/key"
	wrapper.Outputs[0].SSL = &SSL{Key: &keyPath}

	// Remove old outputs and merge replacement
	ucfgCfg := (*ucfg.Config)(cfg)
	_, err := ucfgCfg.Remove("xpack.fleet.outputs", -1, settings.Options...)
	require.NoError(t, err)
	require.NoError(t, cfg.MergeWith(settings.MustCanonicalConfig(wrapper)))

	render, err := cfg.Render()
	require.NoError(t, err)
	fmt.Println(string(render))
}
```
PS: if I change `Certificate            *string  `config:"certificate"` to Certificate            string  `config:"certificate"` this will now render as an empty string `""` 🙂

pkg/controller/association/controller/agent_fleetserver.go
pkoutsovasilis: nice catch!! yes that was a gap and I fixed it in 4a3c2bc953e15ffd7914faf17e3c78450de75d7b thx

pkg/controller/association/controller/agent_fleetserver.go
pkoutsovasilis: yep that was a valid issue, see my reply [here](https://github.com/elastic/cloud-on-k8s/pull/9234#discussion_r3085328039)

pkg/controller/association/controller/agent_fleetserver.go
pkoutsovasilis: changed in 74629105e02c2a69667f15bfc6a3352fadcc6f9a

pkg/controller/agent/pod.go
pkoutsovasilis: renamed in a6e7c9f002b80fd17b3dfc802d257bf342952c4a

pkg/controller/association/controller/agent_fleetserver.go
pkoutsovasilis: transitioned to the Namer approach in 2c1fbe0b044c12895806abda4540cdfc6a31cb18 but I did introduce a Namer for secrets

pkg/controller/kibana/config_settings.go
pkoutsovasilis: I have improved the url matching in c6da4e1b1add2009ee1ba4016002fb3507d071f8

pkoutsovasilis:
buildkite test this -f p=gke,t=TestClientAuthRequired.*,E2E_TAGS=agent -m s=9.3.3,s=8.19.14,s=9.4.0-SNAPSHOT

pkg/controller/association/reconciler.go
```diff
@@ -327,6 +333,15 @@ func (r *Reconciler) reconcileAssociation(ctx context.Context, association commo
 		ClientCertSecretName: clientCertSecretName,
 	}
 
+	if r.ReconcileTransitiveESSecrets != nil {
+		transitiveES, ret := r.ReconcileTransitiveESSecrets(ctx, r.Client, association, assocMeta)
```
kvalliyurnatt, line 337: Potential panic in transitive ES secret reconciliation path.
we unconditionally call `ret.HasError()` after `ReconcileTransitiveESSecrets(...)`, but `fleetManagedAgentTransitiveESRef` (`pkg/controller/association/controller/agent_fleetserver.go`) returns `nil, nil` in several valid no-op paths (for example when fleet-server ref is unset, no ES ref exists, or mTLS is not required).
That means `ret` can be nil here, which can panic reconciliation.
Could we make this contract consistent by either:
1. always returning a non-nil `*reconciler.Results` from `fleetManagedAgentTransitiveESRef`, or
2. guarding in caller with `if ret != nil && ret.HasError()`?

pkoutsovasilis: ok maybe I am missing something, but the panic claim isn't accurate since we have a nil guard inside the func HasError() https://github.com/naemono/cloud-on-k8s/blob/main/pkg/controller/common/reconciler/results.go#L92-L94?

test/e2e/agent/client_auth_test.go
```diff
@@ -0,0 +1,308 @@
...
+	// 1 client certificate; elastic-agent
+	esWithLicense.PostCheckSteps = func(k *test.K8sClient) test.StepList {
+		return test.StepList{clientauth.CheckClientCertificatesCountStep(k, namespace, esBuilder.Elasticsearch.Name, 1)}
+	}
+
+	// Transition ES to client auth disabled.
+	esMutated := esBuilder.DeepCopy().WithMutatedFrom(&esBuilder)
+	esMutated.Elasticsearch.Spec.HTTP.TLS.Client.Authentication = false
+
+	esMutatedWrapped := test.WrappedBuilder{
```
pebrc, line 65:
```suggestion
	esWithoutLicense := test.WrappedBuilder{
```

pkg/controller/association/controller/agent_fleetserver_test.go
```diff
@@ -0,0 +1,633 @@
...
+			wantRef: &commonv1.TransitiveESRef{
+				ClientCertSecretName: "agent1-agent-es-" + hash.HashObject(esSelector.NamespacedName()) + "-client-cert",
+			},
+			wantCreatedSecret: &types.NamespacedName{
+				Namespace: "ns",
+				Name:      "agent1-agent-es-" + hash.HashObject(esSelector.NamespacedName()) + "-client-cert",
+			},
+		},
+		{
+			name: "conf is nil returns nil and cleans up",
```
pebrc, line 537: Are we missing one case for the external ref?

pkg/controller/kibana/config_settings.go
```diff
@@ -401,3 +414,108 @@ func packageRegistrySettings(kb kbv1.Kibana) map[string]any {
...
+			continue
+		}
+
+		if !outputHostsContain(output, esURL) {
+			continue
+		}
+
+		outputCanonical := (*settings.CanonicalConfig)(output)
+
+		if err := outputCanonical.AppendString("ssl.certificate_authorities", caPath); err != nil {
```
pebrc, line 489: nit: Should we make this idempotent?  To avoid repeated appending? The client cert version has guards. I know that appending the same CA multiple times is benign here.

pkg/controller/agent/pod.go
```diff
@@ -85,9 +85,15 @@ const (
 	FleetServerElasticsearchUsername = "FLEET_SERVER_ELASTICSEARCH_USERNAME"
 	FleetServerElasticsearchPassword = "FLEET_SERVER_ELASTICSEARCH_PASSWORD" //nolint:gosec
 	FleetServerElasticsearchCA       = "FLEET_SERVER_ELASTICSEARCH_CA"
+	FleetServerESCert                = "FLEET_SERVER_ES_CERT"
+	FleetServerESCertKey             = "FLEET_SERVER_ES_CERT_KEY"
 	FleetServerPolicyID              = "FLEET_SERVER_POLICY_ID"
 	FleetServerServiceToken          = "FLEET_SERVER_SERVICE_TOKEN" //nolint:gosec
 
+	// FleetManagedAgentClientCertDir is the stable mount path for client certificates inside
```
pebrc, line 93: Should we add a comment that this works because Fleet server is effectively constrained to a single ES output?

pebrc (approved):
LGTM 🤞

pkg/controller/association/reconciler.go
kvalliyurnatt: sorry I missed that completly.

kvalliyurnatt (approved):

pkg/controller/association/controller/agent_fleetserver_test.go
pkoutsovasilis: nice catch, added in 5f4e32da6ed752a5d1634433ab2d5fbc1e9c37ce

pkg/controller/agent/pod.go
pkoutsovasilis: added comment in 68315d54b63db3322baa61a9aed98c7612e00d0a

test/e2e/agent/client_auth_test.go
pkoutsovasilis: Unless I am missing something, the mutation only disables `spec.http.tls.client.authentication` on ES - the enterprise license installed by LicenseTestBuilder remains active throughout the mutation phase. IMO renaming to `esWithoutLicense` would be misleading - happy to rename if you had something else in mind though.

pkg/controller/kibana/config_settings.go
pkoutsovasilis: We've been here in the past; specifically, [your comment on #9229](https://github.com/elastic/cloud-on-k8s/pull/9229#discussion_r2965468043) - the necessary code to guard against duplicate CAs increases the complexity without any actual gains and as you already mentioned appending the same CA multiple times is benign here. So I'm going to keep it the same just to be consistent.

PS: the client cert version doesn't have uniqueness guards per se, it checks if it's already set by the user otherwise it sets it.

pkoutsovasilis:
buildkite test this -f p=gke,t=TestClientAuthRequired.*,E2E_TAGS=agent -m s=9.3.3,s=8.19.14,s=9.4.0-SNAPSHOT

pkg/controller/kibana/config_settings.go
pebrc: Not to nit-pick my own nit-pick but it looks like AppendString has in-built idempotency. In any case we are good.

pkoutsovasilis:
buildkite test this -f p=gke,t=TestClientAuthRequired.*,E2E_TAGS=agent -m s=9.3.3,s=8.19.14,s=9.4.0-SNAPSHOT

pkoutsovasilis:
## Update: Fleet Agent mTLS e2e tests - expected long runtime during rolling restarts

Heads up on something I investigated regarding the new fleet agent client auth transition tests (`TestClientAuthRequiredTransition_FleetAgent`).

[CI results](https://buildkite.com/elastic/cloud-on-k8s-operator/builds/14441)
| Test | Time |
|---|---|
| `TestClientAuthRequiredCustomCertificate_FleetAgent` | 6m57s |
| `TestClientAuthRequiredCustomCertificate_StandaloneAgent` | 2m19s |
| `TestClientAuthRequiredTransition_FleetAgent` | -> **12m18s** <- |
| `TestClientAuthRequiredTransition_StandaloneAgent` | 4m19s |

The `TestClientAuthRequiredTransition_FleetAgent` test is significantly slower than the other  at ~12 minutes. AFAICT this is not due to anything in this PR — it's due to how ES handles JVM shutdown with many indices during the rolling restart triggered by the mTLS → no-mTLS mutation.

### What's happening

When an ES node receives SIGTERM, `IndicesService.doStop()` flushes and closes all local shards using a fixed 5-thread pool. Each shard calls `engine.flushAndClose()` which commits translog data to Lucene segments and fsyncs to disk sequentially within each index.

A Fleet deployment creates ~22 data streams from agent self-monitoring (`logs-elastic_agent.*`, `metrics-elastic_agent.*`, `metrics-system.*`, etc.), plus Kibana system indices and alert indices - totaling ~71 indices. Each ES node hosts ~47 shards that must all be flushed on shutdown.

I confirmed this with DEBUG engine logging — the shutdown is entirely spent in flush operations across 4 threads spanning ~80 seconds per node. Even with `PRE_STOP_ADDITIONAL_WAIT_SECONDS=0` (as e2e tests set), the ES JVM shutdown alone takes ~130s per node — and on subsequent pods it reaches the 180s `terminationGracePeriodSeconds` limit, meaning **pods are being SIGKILLed** rather than shutting down gracefully. With 3 pods restarting sequentially (each waiting for the previous to rejoin and the cluster to go green), the rolling restart portion alone accounts for several minutes.

The standalone agent tests are faster because they don't involve Fleet Server + Kibana + fleet-managed agents, so fewer data streams are created and the overall stack is simpler.

### What I tried locally

- **Enabling mmapfs (removing `allow_mmap: false`)**: No difference - the bottleneck is not storage I/O type, it's the sequential shard flush.
- **Reducing index count by disabling `monitoring_enabled` on Fleet agent policies**: Tried this to see if fewer indices would reduce shutdown time. The idea is sound (fewer shards to flush = faster shutdown), but Fleet defaults monitoring to enabled when the field is omitted - it needs to be explicitly set to `monitoring_enabled: []`. Also, removing monitoring means `WithFleetAgentDataStreamsValidation()` can't be used, so validation needs to target other data streams (`system.cpu` from the `system` package).


As shown here the e2e tests pass but I am unsure about the impact on the CI run time, @pebrc thoughts? 🙂

pebrc:
> As shown here the e2e tests pass but I am unsure about the impact on the CI run time, @pebrc thoughts? 🙂

Lets accept the runtime increase in this PR and create dedicated issue that looks at our e2e tests holistically to see how we can speed up the runs.

pkoutsovasilis:
> Lets accept the runtime increase in this PR and create dedicated issue that looks at our e2e tests holistically to see how we can speed up the runs.

ok I created [this issue](https://github.com/elastic/cloud-on-k8s/issues/9393)

Issue #8442, opened 2025-01-28
Refactor validating webhooks to use CustomValidator


pebrc:
Maybe create an issue for this?

_Originally posted by @thbkrkr in https://github.com/elastic/cloud-on-k8s/pull/8400#discussion_r1932318913_
            
We made some compromises to merge #8400. The goal of the refactoring would be to align ourselves again with controller-runtimes recommended implementation and even be able to use the controller-runtime provided handler. 

One problem we would have to solve is to implement the namespace restriction our handler currently provides which ignores admission requests that are not for one of the managed namespaces. One possible idea would be to wrap the controller-runtime handler with a kind of middleware that does just that while allowing us to keep using the official implementation.

naemono:
Resolved by #9235

PR #9235, opened 2026-03-16
Surface webhook warnings; Refactor webhooks to use controller-runtime's Validator.

### LINKED ISSUE #6461: Validation webhooks should return warnings

naemono:
## Summary

Resolves #6461 and #8442

Validation webhooks now return admission warnings for (all of the Elastic stack) deprecated stack versions, deprecated fields, and unsupported Elasticsearch settings (Elasticsearch only). Previously these were only surfaced through operator logs, which users rarely noticed. With this change, `kubectl apply` displays warnings directly to the user, and warnings are also emitted as Kubernetes events (`reason: Validation`) during reconciliation.

### What changed

**Webhook refactor:**
- Replaced the custom ECK `admission.Validator` interface and handler (`pkg/controller/common/webhook/admission/validator.go`, `pkg/controller/common/webhook/webhook.go`) with controller-runtime's generic `admission.Validator[T]` and `admission.WithValidator[T]`.
- Introduced `ResourceValidator[T]` (`pkg/controller/common/webhook/resource_validator.go`) as a thin wrapper that handles namespace filtering and license annotation checks before delegating to the inner validator.
- Extracted webhook registration from `cmd/manager/main.go` into `cmd/manager/validation.go`.

**Warning propagation:**
- Agent, APM, Beat, Enterprise Search, Kibana, Maps, Package Registry, StackConfigPolicy, and ES v1beta1 now return deprecation warnings through their `Validate` functions.
- Elasticsearch v1, Logstash, ElasticsearchAutoscaler, and AutoOps use inner `admission.Validator[T]` implementations that return warnings alongside errors — including on denied requests, so users always see relevant warnings.
- Added deprecation warnings for PackageRegistry (version deprecation) and wired StackConfigPolicy's existing `GetWarnings()` (SecureSettings deprecation) into its admission response.
- Controller `validate()` functions for all affected resources now emit warnings returned by `Validate` as Kubernetes events (`reason: Validation`, type: `Warning`) instead of silently discarding them.

**Bug fixes:**
- Elasticsearch `ValidateUpdate` no longer swallows warnings when update validations (e.g. downgrade check) fail. Warnings are now collected unconditionally and returned with the denial.

### License checking

The license annotation check (`eck.k8s.elastic.co/license`) is handled differently depending on whether the resource's validation function is also called from the reconciler:

- **ES, ESA**: License check stays inside `validations()` (not the wrapper) because the reconciler calls `ValidateElasticsearch` / `ValidateElasticsearchAutoscaler` directly without going through the wrapper.
- **AutoOps**: License checker is passed to the inner validator for license-dependent version range selection; the wrapper receives `nil`.
- **Logstash**: No license-dependent validation; wrapper receives `nil`.
- **All other resources**: License check is performed by the wrapper via `preValidate`.

## Other fixes
- See https://github.com/elastic/cloud-on-k8s/pull/9235/commits/3bc13b05fe92a76b33ea61162ba27ed1a6a0c402. Wrong resource name in webhook manifest for packageRegistry, and webhook path wasn't correct across operator + helm so webhook calls were failing open. (Noticed this when warnings weren't produced for deprecated packageregistry)

<details>
  <summary>Testing</summary>

Scenarios Tested (both automated, and manual)

- Validating webhooks run through controller-runtime’s generic validator pattern; `kubectl` shows admission warnings where expected and rejects invalid specs cleanly.
- Deprecation (EOL stack versions) surfaces as non-blocking warnings on create/update across several resource types, including Elasticsearch and PackageRegistry.
- Elasticsearch-specific behavior: deprecation, unsupported settings, and warnings that still appear when an update is denied (e.g. downgrade).
- Other product-specific warnings (Agent, StackConfigPolicy, etc.) show up at apply time when configured.
- Downgrades stay blocked; deletes stay unrestricted; unmanaged namespaces are skipped when the operator limits watched namespaces.
- Older v1beta1 API versions still validate. Elasticsearch and Logstash are smoke-tested as examples of resources with their own validating-webhook code; most other CRDs share a single common registration path.
- Admission warnings are expected to line up with **Warning** events in the namespace after reconcile.
- Edge cases: operator upgrade, webhook cert rotation, a burst of parallel creates (sanity-check for deadlocks or webhook errors under concurrent traffic), license denial, webhook downtime (`failurePolicy: Ignore`), server-side apply / dry-run.

</summary>
</details>

## AI Usage

AI leveraged for planning the changes, implemented the changes, building a test plan for testing the changes, running the test plan, and performing code reviews. Manual reviews done after implementation, again with a multitude of findings of mistakes made by the AI tools. Manual testing of many of the automated test plan steps done as well.

- [x] Cursor used with multiple models (composer-2, Opus 4.6)
- [x] Claude Opus 4.6

### DIFF
--- a/pkg/apis/maps/v1alpha1/webhook.go
+++ b/pkg/apis/maps/v1alpha1/webhook.go
@@ -30,7 +30,7 @@ var (
 	}
 )
 
-// +kubebuilder:webhook:path=/validate-ems-k8s-elastic-co-v1alpha1-mapsservers,mutating=false,failurePolicy=ignore,groups=maps.k8s.elastic.co,resources=mapsservers,verbs=create;update,versions=v1alpha1,name=elastic-ems-validation-v1alpha1.k8s.elastic.co,sideEffects=None,admissionReviewVersions=v1,matchPolicy=Exact
+// +kubebuilder:webhook:path=/validate-ems-k8s-elastic-co-v1alpha1-mapsservers,mutating=false,failurePolicy=ignore,groups=maps.k8s.elastic.co,resources=elasticmapsservers,verbs=create;update,versions=v1alpha1,name=elastic-ems-validation-v1alpha1.k8s.elastic.co,sideEffects=None,admissionReviewVersions=v1,matchPolicy=Exact
 
 // Validate is called by the validating webhook to validate an ElasticMapsServer resource. There's no update-specific checks, so the old parameter is ignored.
 func Validate(m *ElasticMapsServer, _ *ElasticMapsServer) (admission.Warnings, error) {


prodsecmachine:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/28c30fa1-3be6-434c-b888-42fca9105c87?refs=fcce409d544395ad5b9353f149e7b14287569480&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/28c30fa1-3be6-434c-b888-42fca9105c87/license?refs=fcce409d544395ad5b9353f149e7b14287569480&source=prChecksComment) |

config/webhook/manifests.yaml
```diff
@@ -191,7 +191,7 @@ webhooks:
     - CREATE
     - UPDATE
     resources:
-    - packageregistry
+    - packageregistries
```
naemono, line 194: This was a previous bug in the packageregistry webhook.

deploy/eck-operator/templates/webhook.yaml
```diff
@@ -483,7 +483,7 @@ webhooks:
     service:
       name: {{ include "eck-operator.webhookServiceName" . }}
       namespace: {{ .Release.Namespace }}
-      path: /validate-epr-k8s-elastic-co-v1alpha1-elasticpackageregistry
+      path: /validate-epr-k8s-elastic-co-v1alpha1-packageregistry
```
naemono, line 486: This was a previous bug in the packageregistry webhook (path inconsistency)

pkg/apis/packageregistry/v1alpha1/webhook.go
```diff
@@ -32,49 +29,36 @@ var (
 	}
 )
 
-// +kubebuilder:webhook:path=/validate-epr-k8s-elastic-co-v1alpha1-packageregistry,mutating=false,failurePolicy=ignore,groups=packageregistry.k8s.elastic.co,resources=packageregistry,verbs=create;update,versions=v1alpha1,name=elastic-epr-validation-v1alpha1.k8s.elastic.co,sideEffects=None,admissionReviewVersions=v1,matchPolicy=Exact
+// +kubebuilder:webhook:path=/validate-epr-k8s-elastic-co-v1alpha1-packageregistry,mutating=false,failurePolicy=ignore,groups=packageregistry.k8s.elastic.co,resources=packageregistries,verbs=create;update,versions=v1alpha1,name=elastic-epr-validation-v1alpha1.k8s.elastic.co,sideEffects=None,admissionReviewVersions=v1,matchPolicy=Exact
```
naemono, line 32: This was a previous bug in the packageregistry webhook.

pkg/controller/logstash/validation/webhook.go
```diff
@@ -30,93 +28,71 @@ const (
...
 func RegisterWebhook(mgr ctrl.Manager, validateStorageClass bool, managedNamespaces []string) {
-	wh := &validatingWebhook{
+	inner := &validator{
 		client:               mgr.GetClient(),
-		decoder:              admission.NewDecoder(mgr.GetScheme()),
 		validateStorageClass: validateStorageClass,
-		managedNamespaces:    set.Make(managedNamespaces...),
 	}
+	// License checks run inside validations(), so we pass nil here
+	// (the reconciler calls validate directly).
```
pebrc: Nit: This comment appears to be copied from the Elasticsearch webhook, but Logstash's `validations()` doesn't contain any license checks (confirmed by reading `validations.go`). The PR description correctly says "Logstash: No license-dependent validation; wrapper receives nil."

Suggested rewording:
```go
// Logstash has no license-dependent validation, so we pass nil here.
```

config/webhook/manifests.yaml
pebrc, line 1: **Pre-existing bug (not introduced by this PR):** The ElasticMapsServer webhook registers `resources: ["mapsservers"]` (line 131) but the CRD's plural name is `elasticmapsservers`. This means the Maps validating webhook has never actually fired — requests always fail open via `failurePolicy: Ignore`.

I confirmed this by deploying the PR build and testing:
```
$ kubectl apply --dry-run=server -f - <<< '{"apiVersion":"maps.k8s.elastic.co/v1alpha1","kind":"ElasticMapsServer","metadata":{"name":"test","namespace":"default"},"spec":{"version":"0.0.0","count":1}}'
elasticmapsserver.maps.k8s.elastic.co/test created (server dry run)
```
Version `0.0.0` should be rejected but was allowed — the webhook isn't being called.

Same issue exists in `deploy/eck-operator/templates/webhook.yaml` (line 292). Since this PR already fixes the analogous PackageRegistry resource name bug, this would be a natural place to fix the Maps one too. The fix would be changing `mapsservers` → `elasticmapsservers` in both files.

pkg/controller/elasticsearch/validation/warnings.go
```diff
@@ -82,14 +95,40 @@ func validZoneAwarenessAffinityWarnings(es esv1.Elasticsearch) field.ErrorList {
...
+		case field.ErrorTypeForbidden:
+			admissionWarnings = append(admissionWarnings, fmt.Sprintf("%s: %s", e.Field, e.Detail))
+		case field.ErrorTypeInvalid:
+			blocking = append(blocking, e)
+		default:
+			blocking = append(blocking, e)
+		}
+	}
+	return admissionWarnings, blocking
+}
```
pkoutsovasilis, line 121: with this change the previously only-warning `xpack.security.http.ssl.client_authentication: required` is now blocking admission. Although this will change when https://github.com/elastic/cloud-on-k8s/pull/9229 is merged

pkg/apis/agent/v1alpha1/webhook.go
```diff
@@ -43,62 +39,30 @@ func (a *Agent) warnings() []string {
...
 	if old != nil {
 		for _, uc := range updateChecks {
 			if err := uc(old, a); err != nil {
 				errors = append(errors, err...)
 			}
 		}
-
-		if len(errors) > 0 {
-			return warnings, apierrors.NewInvalid(groupKind, a.Name, errors)
-		}
```
pkoutsovasilis, line 99: this is incosistent from what kibana validate does which still has early return after the updateChecks, should these two be aligned?

pkg/controller/elasticsearch/validation/webhook.go
```diff
@@ -31,137 +28,97 @@ const (
...
+	return nil, nil
+}
 
-	// If this Elasticsearch instance is not within the set of managed namespaces
-	// for this operator ignore this request.
-	if wh.managedNamespaces.Count() > 0 && !wh.managedNamespaces.Has(es.Namespace) {
-		eslog.V(1).Info("Skip Elasticsearch resource validation", "name", es.Name, "namespace", es.Namespace)
-		return admission.Allowed("")
+// ValidateElasticsearch validates an Elasticsearch instance against a set of validation funcs returning warnings and an error if validation fails.
+func ValidateElasticsearch(ctx context.Context, es esv1.Elasticsearch, checker license.Checker, exposedNodeLabels NodeLabels) (admission.Warnings, error) {
```
pebrc, line 87: The PR description says warnings are "emitted as Kubernetes events during reconciliation", and all other controllers (Agent, APM, Beat, Kibana, etc.) were updated to emit warnings from their `Validate` calls. However, the ES reconciler in `elasticsearch_controller.go:262` still discards the warnings:

```go
_, err := validation.ValidateElasticsearch(ctx, es, r.licenseChecker, r.ExposedNodeLabels)
```

There's a separate `CheckForWarnings` call at line 276, but it doesn't fully overlap with `ValidateElasticsearch`'s warnings. Specifically, `validateRestartAllocationDelayWarnings` is only returned by `ValidateElasticsearch` — never by `CheckForWarnings`. The settings warnings also use different formatting between the two paths.

Worth either emitting the returned warnings as events (matching the other controllers), or documenting that ES intentionally relies on `CheckForWarnings` alone and accepting the gap.

config/webhook/manifests.yaml
naemono: So @pebrc since this is the 2nd instance of this, and we run blind to these failures, what would you think about opening an issue where we have a ci-run tool that automatically validates all of these for us and fails CI when the CRD plural vs the webhook plural do not match?

pkg/controller/elasticsearch/validation/warnings.go
naemono: Good catch, I just updated the code, and the test for this.

pkg/apis/agent/v1alpha1/webhook.go
naemono: Good catch. I've made this consistent.

pkg/controller/elasticsearch/validation/webhook.go
naemono: I've adjusted this to match other controllers, and just removed the `CheckForWarnings` func.

config/webhook/manifests.yaml
pkoutsovasilis: having a ci-run tool that automatically validates all of these sounds like a good plan and I am in favour of it. In the meantime since you did address packageregistry here @naemono , maybe we should tackle `mapsservers`?

naemono: @pkoutsovasilis I had updated this here: https://github.com/elastic/cloud-on-k8s/pull/9235/changes#diff-ce1f294298cca0df2433c4c382d0a6da5be1c6e6ab363cca97ba3edf01fd6aa9, but not in `config/webhook/manifests.yaml`. I've updated both now, thanks.

pkoutsovasilis: I think the `generate` complains because of this 
```
// +kubebuilder:webhook:path=/validate-ems-k8s-elastic-co-v1alpha1-mapsservers,mutating=false,failurePolicy=ignore,groups=maps.k8s.elastic.co,resources=mapsservers,verbs=create;update,versions=v1alpha1,name=elastic-ems-validation-v1alpha1.k8s.elastic.co,sideEffects=None,admissionReviewVersions=v1,matchPolicy=Exact
```
which should become
```
// +kubebuilder:webhook:path=/validate-ems-k8s-elastic-co-v1alpha1-mapsservers,mutating=false,failurePolicy=ignore,groups=maps.k8s.elastic.co,resources=elasticmapsservers,verbs=create;update,versions=v1alpha1,name=elastic-ems-validation-v1alpha1.k8s.elastic.co,sideEffects=None,admissionReviewVersions=v1,matchPolicy=Exact
```

naemono: I had totally missed that the step was failing. I think I've fixed this... verifying...

naemono: Seems good now. 👍

Issue #8450, opened 2025-01-31
Validate that failure to create certificate secret for remote CA does not block reconciliation of Elasticsearch itself


kvalliyurnatt:
We need to validate that the reconciling the Elasticsearch cluster is not blocked by the failure to create secret for remote CA.

kvalliyurnatt:
This comment gives some more context

The user created 4 secrets with ["common.k8s.elastic.co/type"] == "remote-ca" label

project-01-elastic-example-master-master-cluster-elastic-es-remote-ca
project-01-elastic-example-master-master-cluster-elastic-es-remote-ca-bak
project-01-elastic-example-metrics-example-metrics-elastic-es-remote-ca
project-01-elastic-example-metrics-example-metrics-elastic-es-remote-ca-bak

each of them having ~300 certificates. What ECK does is it collates certificates from all the secrets with the remote-ca label and creates one single secret to mount to the Elasticsearch pod, this is the step that was failing. ECK by itself does not create so many certificates. It just collates all the user provided certificates into one secret. So we need to check with the customer to see why they added so many secrets. Continuous failures to create/update secrets could affect the ability of the operator to prevent es from going into an unhealthy state. we are going to audit the behavior to ensure that mission-critical pieces are reconciled first so that they do not get blocked by the failure to create this secret for the remote CA.

pebrc:
So the idea would be to make the remote CA aggregation failure non-blocking, similar to how other non-critical operations are handled? 

An alternative would be to have a clearer distinction between critical and non-critical operations in the operator and make sure the critical ones run first or at least cannot be bypassed by errors. This would probably be a larger refactoring.

kvalliyurnatt:
> So the idea would be to make the remote CA aggregation failure non-blocking, similar to how other non-critical operations are handled?

Yeah that makes sense, also like you mentioned it might be worth going through and identifying other non critical things in the critical path and separating them out.

dtuck9:
## Halt-on-error steps in `ReconcileSharedResources` 

The `certificates.ReconcileTransport` in particular is where the call is to `remoteca.Reconcile` where the steps to coalesce the `remote-ca`-labeled secrets happens.                                                                                                                          
                                                                                                                                                                                   
  ### Halt-on-error (early return)
✅  = critical
❌  = non-critical
❓ = needs discussion                                                                                                                                                  
   
  | Critical Path | Line | Step | Notes |                                                                                                                                                            
[...]

PR #9236, opened 2026-03-16
Move `URLProvider` to `Params` to inject into `Reconcile` functions

dtuck9:
Moving `URLProvider` to `Params` for the Elasticsearch controller's reconcile functions to enable better testability.

I've run the entire e2e test suite and all 175 tests passed.

Screenshot 2026-03-17 at 8 48 01 AM

A bit of a chicken and the egg scenario -- the `ReconcileSharedResources` function wasn't really testable as a result of the embedded `URLProvider`, so we don't have any test coverage on this function (outside of the e2e tests afaict) but in turn moving the `URLProvider` to `Params` was a bit risky despite being theoretically the same.

I opted to push the refactoring change, run the e2e tests, and then add an integration test as a follow-up commit to add at least happy path test coverage.

This is a precursor to reproducing the error in https://github.com/elastic/cloud-on-k8s/issues/8450 and a prerequisite for refactoring the Elasticsearch reconciliation. 

### As a unit test - coverage is increased
Prior to this change, the test coverage was:
```
ok      github.com/elastic/cloud-on-k8s/v3/pkg/controller/elasticsearch/driver/shared   1.515s  coverage: 29.8% of statements
```

And with just the single happy path test case:
```
ok      github.com/elastic/cloud-on-k8s/v3/pkg/controller/elasticsearch/driver/shared   (cached)        coverage: 66.2% of statements
```

### Test profiling
Prior to refactoring and skipping expected certificate generation, this test took 6+ seconds to run.

Skipping the expected certificate generation reduced that to < 3 seconds, which is still more than one might expect.

I ran CPU, memory, and block profiling on the test:

[mem.prof.txt](https://github.com/user-attachments/files/26289678/mem.prof.txt)
[cpu.prof.txt](https://github.com/user-attachments/files/26289679/cpu.prof.txt)
[block.prof.txt](https://github.com/user-attachments/files/26289680/block.prof.txt)

Where the majority of the block was spent in the HTTP server as well as the the test scaffolding itself, the memory was expectedly in the fake K8s client, and the CPU was in the certificate generation during reconciliation. I did not see much that I could improve within the scope of this task.

Additional refactoring will be done to `ReconcileSharedResources` to make this more testable, and potentially eliminate the need for the HTTP server.


prodsecmachine:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/e483f8f6-44a3-4108-9fd2-7701c901de30?refs=9932466b32d3bf8cdc8068c5a3716f88eaf67ec9&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/e483f8f6-44a3-4108-9fd2-7701c901de30/license?refs=9932466b32d3bf8cdc8068c5a3716f88eaf67ec9&source=prChecksComment) |

dtuck9:
buildkite test this -f p=gke

pkg/controller/elasticsearch/driver/shared/reconcile_integration_test.go
```diff
@@ -0,0 +1,207 @@
...
+			},
+			Data: map[string][]byte{},
+		}
+	}
+
+	return secretsToCreate
+}
+
+func mustGetParams(t *testing.T, esServer *httptest.Server, numRemoteCAs int, initK8sObjects ...client.Object) driver.Parameters {
+	defer time.Sleep(2 * time.Minute)
```
moukoublen: Why do we need this sleep here? 

If I am not mistaken, this sleep will "freeze" the function `mustGetParams` 2 minutes before it returns, right? 

Which means the test struct initialization will sleep for 2 minutes.

pebrc: As @moukoublen already noted, this `defer time.Sleep(2 * time.Minute)` will cause `mustGetParams` to block for 2 minutes before returning. This looks like a debugging leftover and should be removed.

pkg/controller/elasticsearch/driver/shared/reconcile_integration_test.go
pebrc, line 1: Consider whether this test needs to be an integration test at all. It uses `k8s.NewFakeClient` (not envtest) and a mock HTTP server — there's nothing here that requires the `integration` build tag.

A few suggestions:

1. **Drop the `integration` build tag** and make this a regular unit test. It'll run in CI without needing a special build flag and provide faster feedback.

2. **Mock at the ES client interface level** instead of standing up an `httptest.Server`. This would simplify the test setup significantly.

3. **Assert on side effects.** The current assertions only check the returned `ReconcileState` (labels, `ESReachable`, `ESClient != nil`). The main value of testing `ReconcileSharedResources` is verifying the K8s resources it creates/updates — secrets, services, certificates, configmaps, user credentials. Use `client.Get()` on the fake client to verify those were created with correct content and ownership.

4. **Add more test cases.** Right now there's only "happy path". Cases like no endpoints, missing pods, or version mismatches would make this test much more useful as a safety net for the upcoming reconciliation refactoring.

pkg/controller/elasticsearch/driver/shared/reconcile_integration_test.go
```diff
@@ -0,0 +1,207 @@
...
+	tests := []struct {
+		name                   string
+		params                 driver.Parameters
+		expectedState          *ReconcileState
+		reconciliationExpected bool
+	}{
+		{
+			name: "happy path",
+			params: mustGetParams(t, esServer, 0,
+				&corev1.Pod{Status: corev1.PodStatus{Phase: corev1.PodRunning, Conditions: []corev1.PodCondition{{Status: corev1.ConditionTrue, Type: corev1.ContainersReady}, {Status: corev1.ConditionTrue, Type: corev1.PodReady}}}, ObjectMeta: metav1.ObjectMeta{Name: "test-pod", Namespace: "test-ns", Labels: map[string]string{label.VersionLabelName: "9.3.1", label.ClusterNameLabelName: "test-es", label.HTTPSchemeLabelName: "http", label.StatefulSetNameLabelName: esv1.StatefulSet("test-es", "sset1")}}, Spec: corev1.PodSpec{HostAliases: []corev1.HostAlias{{IP: ip, Hostnames: []string{"test-pod.test-es-es-sset1.test-ns"}}}}},
```
pebrc: These struct literals are very hard to read as single lines. Consider breaking the Pod and EndpointSlice into named variables above the test table for readability.

pkg/controller/elasticsearch/driver/shared/reconcile_integration_test.go
```diff
@@ -0,0 +1,207 @@
...
+	watches2 "github.com/elastic/cloud-on-k8s/v3/pkg/controller/common/watches"
+	client2 "github.com/elastic/cloud-on-k8s/v3/pkg/controller/elasticsearch/client"
+	"github.com/elastic/cloud-on-k8s/v3/pkg/controller/elasticsearch/driver"
+	"github.com/elastic/cloud-on-k8s/v3/pkg/controller/elasticsearch/label"
+	"github.com/elastic/cloud-on-k8s/v3/pkg/controller/elasticsearch/observer"
+	"github.com/elastic/cloud-on-k8s/v3/pkg/controller/elasticsearch/reconcile"
+	"github.com/elastic/cloud-on-k8s/v3/pkg/controller/elasticsearch/version"
+	"github.com/elastic/cloud-on-k8s/v3/pkg/utils/cryptutil"
+	"github.com/elastic/cloud-on-k8s/v3/pkg/utils/k8s"
+	"github.com/elastic/cloud-on-k8s/v3/pkg/utils/pointer"
```
pebrc: Nit: The test mixes `k8s.io/utils/ptr` (line 24) and `pkg/utils/pointer` (here) for the same purpose. Consider using `ptr.To[int32](9200)` to keep it consistent.

pkg/controller/elasticsearch/driver/shared/reconcile_integration_test.go
dtuck9: Oh that was a relic of local debugging, I'll remove it.

dtuck9: Yup that's exactly what it was -- debugging the second test for the excessive number of remote CAs which I've completely omitted from this PR. Removed.

pkg/controller/elasticsearch/driver/shared/reconcile_integration_test.go
dtuck9: > **Drop the `integration` build tag and make this a regular unit test.** It'll run in CI without needing a special build flag and provide faster feedback.

Sounds good -- I'm moving it to the existing reconcile_test.go file then. (for context, I was following the guidelines of "any test that takes more than a few milliseconds is an integration test" guidance -- can't remember where exactly I saw that in the documentation -- and just for a single test case this is at 1.5s):

```
$ go test -run TestReconcileSharedResources
PASS
ok      github.com/elastic/cloud-on-k8s/v3/pkg/controller/elasticsearch/driver/shared   1.530s
```

> **Mock at the ES client interface level instead of standing up an `httptest.Server.`** This would simplify the test setup significantly.

The ES client is created within the function as well and depends on resources created earlier in the function, so refactoring to allow for mocking would be more complex, unless I am missing something?

> **Assert on side effects.** The current assertions only check the returned `ReconcileState` (labels, `ESReachable`, `ESClient != nil`). The main value of testing `ReconcileSharedResources` is verifying the K8s resources it creates/updates — secrets, services, certificates, configmaps, user credentials. Use `client.Get()` on the fake client to verify those were created with correct content and ownership.

Will do.

> **Add more test cases.** Right now there's only "happy path". Cases like no endpoints, missing pods, or version mismatches would make this test much more useful as a safety net for the upcoming reconciliation refactoring.

Will do.

dtuck9: As discussed, moving the test back to reconcile_integration_test.go and leaving the mocking in `httptest.Server`

I'll also look into why this test is so slow and will either update the test, the PR description, or both depending on the outcome.

pkg/controller/elasticsearch/driver/shared/reconcile_integration_test.go
```diff
@@ -0,0 +1,1942 @@
...
+					Namespace: "test-ns",
+				},
+				Spec: corev1.ServiceSpec{
+					Ports: []corev1.ServicePort{{Name: "http", Port: 9200}},
+				}},
+		},
+		Version: "9.3.1"},
+}
+
+func TestReconcileSharedResources(t *testing.T) {
```
pebrc, line 98: The test helper functions (`getExpectedService`, `getPortForService`, `mustBuildExpectedCertificateSecrets`, `mustGetExpectedConfigMaps`, `getPod`, etc.) reimplement logic that already exists in production code rather than reusing it:

- **Services**: `getExpectedService` / `getPortForService` manually reconstruct what `services.NewTransportService()`, `services.NewExternalService()`, etc. already produce. If production code is intentionally changed (e.g. different port name), the test helper silently drifts. Consider calling the production constructors to build expected objects.

- **Certificates**: `mustBuildExpectedCertificateSecrets` reproduces CA generation and leaf cert signing. Could instead read the secrets back from the fake client after reconciliation and validate their *properties* (CA signed the leaf, correct DNS SANs, valid time range) using `certificates.NewCertificatesSecret()` which already exists and validates structure.

- **Pods**: `getPod` hand-rolls labels. There is an existing `testPod` builder in `pkg/controller/elasticsearch/driver/stateful/fixtures.go` that uses `label.NewPodLabels()` to match production label construction. Consider extracting it to a shared test fixtures package, or at minimum use `label.NewPodLabels()` here.

- **Roles**: `defaultRoles` is a ~970-line verbatim copy of what `user.PredefinedRoles.FileBytes()` produces at runtime. Any role change will silently break this test. Consider generating the expected value at test time:
```go
expectedRoles, err := user.PredefinedRoles.FileBytes()
require.NoError(t, err)
```

The broader concern: the test compares reconciler output against an independent reimplementation of the same logic. If both have the same bug, the test passes. Using production constructors for expected values would catch actual behavioral regressions.

pkg/controller/elasticsearch/driver/shared/reconcile_integration_test.go
```diff
@@ -0,0 +1,1942 @@
...
+	if isCertOrCaSecret(expected.Name) {
+		actualCertData := make([]byte, 0, 1000)
+		expectedCertData := make([]byte, 0, 1000)
+		for name, cert := range actual.Data {
+			expectedCert, ok := expected.Data[name]
+			require.Truef(t, ok, "actual has Data key %q missing from expected secret %s", name, expected.Name)
+			// This should put the certs in the same order for comparison later
+			actualCertData = append(actualCertData, cert...)
+			expectedCertData = append(expectedCertData, expectedCert...)
+		}
```
pebrc: This loop only iterates over `actual.Data` and looks up matching keys in `expected.Data`. This catches extra keys in actual ("actual has Data key missing from expected"), but if `expected` has a key that `actual` is missing (e.g., if the reconciler stops producing `tls.key`), the loop never visits it and no assertion fires.

Consider adding a key-set or length check before the loop:
```go
assert.Equalf(t, len(expected.Data), len(actual.Data), "secret %s has wrong number of Data keys", actual.Name)
```

pkg/controller/elasticsearch/driver/shared/reconcile_integration_test.go
```diff
@@ -0,0 +1,1942 @@
...
+	// unless we set the URL to the esServer URL
+	urlProvider := esclient.NewStaticURLProvider(defaultURL)
+	if listeningServer != nil {
+		urlProvider = esclient.NewStaticURLProvider(listeningServer.URL) // required to get a response from the mock server
+	}
+
+	baselineObjects := append(mustGenerateRemoteCASecrets(t, baseElasticsearch.Namespace, baseElasticsearch.Name, numRemoteCAs), getServiceAccountTokenSecret(&baseElasticsearch))
+	initK8sObjects = append(initK8sObjects, baselineObjects...)
+
+	return driver.Parameters{
```
pebrc, line 998: `params.Version` is not set here, so it defaults to the zero value. In `ReconcileSharedResources` (reconcile.go:166), when `minVersion == nil` (no pods case), the code falls back to `&params.Version` — which is `0.0.0`. This happens to work because the no-pods test case doesn't assert on version-dependent behavior, but it could mask bugs if more assertions are added later.

pkg/controller/elasticsearch/driver/shared/reconcile_integration_test.go
```diff
@@ -0,0 +1,1942 @@
...
+			Data: map[string][]byte{
+				certificates.CAFileName: ca.Cert.Raw,
+			},
+		}
+	}
+
+	return secretsToCreate
+}
+
+func getExpectedService(es *esv1.Elasticsearch, st serviceType, resourceVersion string) corev1.Service {
```
pebrc: Nit: Naming is inconsistent across the helper functions. `get*` (e.g. `getExpectedService`, `getPod`, `getLabels`) implies retrieval but these are pure constructors. `must*` (e.g. `mustGetParams`, `mustBuildExpectedCertificateSecrets`) follows Go convention for "panics/fatals on error", but some `get*` functions also have no error path at all.

Consider a consistent convention:
- `new*` or `expected*` for pure builders (no error)
- `must*` for builders that call `require` / can fail the test

`get*` is not idiomatic in Go (I know we have not been consistent here in the code base)

pkg/controller/elasticsearch/driver/shared/reconcile_integration_test.go
```diff
@@ -0,0 +1,1942 @@
...
+	remote
+)
+
+var baseElasticsearch = esv1.Elasticsearch{
+	ObjectMeta: metav1.ObjectMeta{
+		Name: "test-es", Namespace: "test-ns",
+		ResourceVersion: "1",
+		Annotations:     map[string]string{observer.ObserverIntervalAnnotation: "10s"},
+		Labels:          map[string]string{label.VersionLabelName: "9.3.1", label.ClusterNameLabelName: "test-es"},
+		OwnerReferences: []metav1.OwnerReference{
```
pebrc, line 76: Nit: `baseElasticsearch` has an `OwnerReference` pointing back to itself (`Name: "test-es"`, `Kind: "Elasticsearch"`). Real Elasticsearch resources don't own themselves — owner references point to a *different* parent resource. This doesn't affect test outcomes today but is misleading for readers trying to understand production behavior.

pkg/controller/elasticsearch/driver/shared/reconcile_integration_test.go
dtuck9: Regarding `certificates.NewCertificatesSecret()` -- I did actually look to use that one initially, but it requires the existence of `tls.key` which the secrets for the public certs do not have:

https://github.com/elastic/cloud-on-k8s/blob/4f43a28920aa9f9e937d4fda5c1e0f6b8d722e81/pkg/controller/common/certificates/secret.go#L152-L155

Where `KeyFileName` is `tls.key`.

I will look into the other ones though.

pkg/controller/common/certificates/x509_othername.go
```diff
@@ -19,6 +19,9 @@ var (
 
 	// CommonNameObjectIdentifier is the OID for a CommonName field in x509
 	CommonNameObjectIdentifier = asn1.ObjectIdentifier{2, 5, 4, 3}
+
+	// OrganizationalUnitIdentifier is the OID for an OrganizationalUnit field in x509
+	OrganizationalUnitIdentifier = asn1.ObjectIdentifier{2, 5, 4, 11}
```
dtuck9, line 24: This is currently only used in this new test but I opted to define it here for consistency. LMK if you'd rather I limit it to the test.

pkg/controller/elasticsearch/driver/shared/reconcile_integration_test.go
dtuck9: > - **Services**: `getExpectedService` / `getPortForService` manually reconstruct what `services.NewTransportService()`, `services.NewExternalService()`, etc. already produce. If production code is intentionally changed (e.g. different port name), the test helper silently drifts. Consider calling the production constructors to build expected objects.

I've updated these. 

> - **Certificates**: `mustBuildExpectedCertificateSecrets` reproduces CA generation and leaf cert signing. Could instead read the secrets back from the fake client after reconciliation and validate their *properties* (CA signed the leaf, correct DNS SANs, valid time range) using `certificates.NewCertificatesSecret()` which already exists and validates structure.

I've implemented a different approach to ensuring the expected issuer, subject, and DNS names are valid. LMK if it resolves your concerns to a satisfactory standard.

> - **Pods**: `getPod` hand-rolls labels. There is an existing `testPod` builder in `pkg/controller/elasticsearch/driver/stateful/fixtures.go` that uses `label.NewPodLabels()` to match production label construction. Consider extracting it to a shared test fixtures package, or at minimum use `label.NewPodLabels()` here.

I opted to just call `label.NewPodLabels()` for now as the existing `testPod` fixture is stateful-specific, and ultimately we'll want something comparable for stateless, but that is outside of the scope of my intentions here, and I would like progress over perfection at this point.

> - **Roles**: `defaultRoles` is a ~970-line verbatim copy of what `user.PredefinedRoles.FileBytes()` produces at runtime. Any role change will silently break this test. Consider generating the expected value at test time:
> ```go
> expectedRoles, err := user.PredefinedRoles.FileBytes()
> require.NoError(t, err)
> ```

Fixed, thank you for pointing me to that.

pkg/controller/elasticsearch/driver/shared/reconcile_integration_test.go
```diff
@@ -0,0 +1,1041 @@
...
+	require.NoError(t, err, "error parsing internal HTTP certificate secret")
+
+	clientCerts, err := certificates.ParsePEMCerts(internalCert.CertChain())
+	require.NoError(t, err, "error parsing internal HTTP certificate chain")
+	return clientCerts
+}
+
+// isCertOrCaSecret returns true for secrets whose Data is generated at reconciliation time
+// and cannot be matched byte-for-byte, such as certificate secrets.
+func isCertOrCaSecret(secretName string) bool {
```
pebrc: Nit: `isCertOrCaSecret` is defined but never called. Looks like a leftover from a previous iteration of the assertion logic — can be removed.

pkg/controller/elasticsearch/driver/shared/reconcile_integration_test.go
```diff
@@ -0,0 +1,1041 @@
...
+	labels["eck.k8s.elastic.co/owner-name"] = es.Name
+	labels["eck.k8s.elastic.co/owner-namespace"] = es.Namespace
+
+	// Non-certificate secrets (users, roles, service account token)
+	serviceAccountTokenSecret := newServiceAccountTokenSecret(es)
+	fileSettings := filesettings.NewEmptySettings(time.Now().Unix())
+	fileSettingsData, err := json.Marshal(fileSettings)
+	require.NoError(t, err, "error marshalling file-settings")
+
+	filerealm.New()
```
pebrc: Nit: Return value is discarded — this is a no-op. Leftover from refactoring?

pkg/controller/elasticsearch/driver/shared/reconcile_integration_test.go
```diff
@@ -0,0 +1,1041 @@
...
+		},
+		Type: corev1.SecretTypeServiceAccountToken,
+		Data: map[string][]byte{
+			user.ServiceAccountTokenNameField: []byte("token-name"),
+			user.ServiceAccountHashField:      []byte("hash"),
+		},
+	}
+}
+
+func mustGetExpectedConfigMaps(t *testing.T, es *esv1.Elasticsearch, resourceVersion string, esServer *httptest.Server) map[string]corev1.ConfigMap {
```
pebrc: Nit: This is a constructor, not a retriever — `mustBuildExpectedConfigMaps` would be more consistent with `mustBuildExpectedSecrets` and `mustBuildParams`.

pkg/controller/elasticsearch/driver/shared/reconcile_integration_test.go
dtuck9: Good catch

pkg/controller/elasticsearch/driver/shared/reconcile_integration_test.go
dtuck9: Oh yeah, I missed that one, thanks

pkg/controller/elasticsearch/driver/shared/reconcile_integration_test.go
dtuck9: Good catch

PR #9256, opened 2026-03-23
fix: align DaemonSet reconciler with Deployment, handle stale cache

### LINKED ISSUE #9246: Agent/Beat DaemonSet reconciliation loop: operator uses Update instead of Create for non-existent DaemonSet
## Description

ECK operator cannot create DaemonSets for Agent and Beat resources. The operator's `ReconcileResource` function always takes the `Update` code path instead of `Create`, even when the DaemonSet does not exist in the API server. This results in an infinite reconciliation loop with `daemonsets.apps "..." not found` errors.

## Environment

- **ECK operator**: tested on both 3.0.0 (controller-runtime v0.20.3) and 3.3.1 (controller-runtime v0.22.4) — same behavior
- **Kubernetes**: v1.33.2
- **Helm chart**: eck-stack 0.17.0, eck-operator 3.3.1
- **Agent mode**: fleet (with `daemonSet` spec)
- **Cluster**: OVH managed Kubernetes

## Symptoms

1. Operator logs show a tight reconciliation loop (~10-20 iterations/second):
```
[info]  agent-controller: Updating resource  kind=DaemonSet name=eck-stack-eck-agent-agent
[info]  agent-controller: Ending reconciliation run
[error] manager.eck-operator: Reconciler error  error="daemonsets.apps \"eck-stack-eck-agent-agent\" not found"
[info]  agent-controller: Starting reconciliation run
```

2. The operator **never** logs "Creating resource" for DaemonSets — always "Updating resource"
3. This happens from **iteration 1** on a fresh operator start with 0 DaemonSets in the namespace
4. Deployments (fleet-server) and StatefulSets (elasticsearch, logstash) managed by the same operator work correctly
5. Both Agent and Beat DaemonSets are affected

### Confirmed via experiments:

1. **DaemonSet survives when operator is stopped** — manually created DaemonSet persists indefinitely with operator scaled to 0
2. **API server has 0 DaemonSets** — verified via `kubectl get --raw '/apis/apps/v1/namespaces/elastic-stack/daemonsets'` → 0 items
3. **RBAC is correct** — operator SA has full CRUD on daemonsets.apps (verified via `kubectl auth can-i`)
4. **No webhooks affect DaemonSets** — checked ValidatingWebhookConfigurations
5. **No resource quotas** — namespace has no quotas
6. **Fresh operator start reproduces the issue** — even after deleting Agent CR, all secrets, restarting operator, waiting 60s for cache sync, then creating Agent CR via ArgoCD — same result from iteration 1

### When DaemonSet is pre-created manually:

- Operator switches from "not found" to **conflict errors**: `"Operation cannot be fulfilled on daemonsets.apps: the object has been modified"`
- The operator can never successfully update the DaemonSet (generation stays at 1)
- This is because the tight reconciliation loop (~10/sec) causes constant resourceVersion conflicts with the DaemonSet controller

## Root Cause Analysis

Looking at the code in `pkg/controller/common/reconciler/reconciler.go`:

```go
err = params.Client.Get(ctx, types.NamespacedName{...}, params.Reconciled)
if err != nil && apierrors.IsNotFound(err) {
    return create()  // "Creating resource"
} else if err != nil {
    return fmt.Errorf("failed to get ...")
}
// NeedsUpdate check → "Updating resource" → client.Update()
```

**`client.Get` returns `nil` (found)** even though the DaemonSet doesn't exist in the API server. Since controller-runtime's `client.Get` reads from the informer cache, this suggests the cache contains a phantom DaemonSet entry.

This behavior is consistent from a fresh operator start (iteration 1) where the informer cache should be empty after the initial LIST returns 0 DaemonSets.

## Agent CR

```yaml
kind: Agent
metadata:
  name: eck-stack-eck-agent
  namespace: elastic-stack
spec:
  mode: fleet
  policyID: eck-agent
  version: 9.2.0
  fleetServerRef:
    name: fleet-server
  kibanaRef:
    name: kibana
  serviceAccountName: elastic-agent
  daemonSet:
    podTemplate:
      spec:
        containers:
        - name: agent
          resources:
            limits:
              cpu: 500m
              memory: 2Gi
            requests:
              cpu: 200m
              memory: 1Gi
        hostNetwork: true
        securityContext:
          runAsUser: 0
        serviceAccountName: elastic-agent
```

## Expected Behavior

The operator should create the DaemonSet when it doesn't exist (using `client.Create`) and then update it as needed.

## Actual Behavior

The operator always attempts `client.Update` (logs "Updating resource"), which fails with 404 because the DaemonSet doesn't exist.

ipiton:
## Summary

make DaemonSet `UpdateReconciled` to perform selective merge instead of `DeepCopyInto`

Aligns `pkg/controller/common/daemonset/reconcile.go` with the pattern already used by:
- **Deployment** (fixed in #6302, Jan 2023)
- **StatefulSet** (introduced in #7357, Dec 2023)

`DeepCopyInto` overwrites the entire reconciled object. The selective merge only updates Labels, Annotations, and Spec — preserving Status and any externally-set metadata.

Fixes #9246


prodsecmachine:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/ea6320ef-6263-4b85-aebb-a03550df9954?refs=948e317ecc584cf98d50024a5f4d299d759f4e0c&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/ea6320ef-6263-4b85-aebb-a03550df9954/license?refs=948e317ecc584cf98d50024a5f4d299d759f4e0c&source=prChecksComment) |

pkg/controller/common/reconciler/reconciler.go
```diff
@@ -121,6 +121,15 @@ func ReconcileResource(params Params) error {
...
 	}
 
+	// If the resource is being deleted (has a deletionTimestamp), treat it as non-existent
+	// and recreate it. The informer cache may return a terminating resource that the API server
+	// will reject updates on, leading to perpetual "not found" or conflict errors.
+	reconciledAccessor, accessorErr := meta.Accessor(params.Reconciled)
+	if accessorErr == nil && reconciledAccessor.GetDeletionTimestamp() != nil {
+		log.Info("Resource is being deleted, will recreate", "deletionTimestamp", reconciledAccessor.GetDeletionTimestamp())
+		return create()
+	}
```
pkoutsovasilis: I don't think the operator should be handling this. A stale informer cache is an infrastructure-level problem - if the cache is returning terminating resources after an extended API server outage, the right fix is at the cluster level, not silently working around it in the reconciler.

Additionally, calling create() here while the old resource is still terminating will likely hit an AlreadyExists error from the API server, potentially trading one reconcile loop for another?!

This also changes ReconcileResource which is shared across all resource types - that's a wide blast radius for an edge case observed on a single managed Kubernetes provider.

pkg/controller/common/reconciler/reconciler.go
```diff
@@ -177,6 +186,12 @@ func ReconcileResource(params Params) error {
 
 		err = params.Client.Update(params.Context, params.Reconciled)
 		if err != nil {
+			// If the resource was deleted between Get (from cache) and Update (to API server),
+			// fall back to creating it. This handles stale informer cache entries.
+			if apierrors.IsNotFound(err) {
+				log.Info("Resource disappeared between get and update, falling back to create")
+				return create()
+			}
```
pkoutsovasilis: Same concern as above - this is compensating for a stale informer cache at the wrong layer. If Get returns a resource from cache but Update gets a NotFound from the API server, the cache is out of sync and that's the problem to address, not something the reconciler should silently try to recover from.

ipiton:
Thanks for the review @pkoutsovasilis! You're right — the stale cache handling was overreaching.

I've stripped the PR down to just the DaemonSet `UpdateReconciled` fix, aligning it with the Deployment reconciler pattern (merge labels/annotations, overwrite Spec, leave Status intact). The `reconciler.go` changes are removed.

Will sign the CLA shortly.

pkoutsovasilis:
buildkite test this

pkoutsovasilis:
buildkite test this -f p=gke,t=TestBeatVersionUpgradeToLatest

pkoutsovasilis:
@ipiton thanks for addressing my comments, the code changes look good to me overall and after the CI reports green I am gonna approve and merge this PR. In the meantime please do update the PR description to reflect that it changes only the Daemonset UpdateReconciled method, ty

pkoutsovasilis:
buildkite test this

PR #9263, opened 2026-03-24
Fix FIPS build and add native Go FIPS 140-3 support

moukoublen:
# Add support for native Go FIPS 140-3 mode

## Summary

This PR migrates the ECK operator FIPS build from the BoringCrypto/CGO-based approach to Go's native FIPS 140-3 support (introduced in Go 1.24). The existing BoringCrypto build path is preserved as a fallback until the native Go FIPS module receives official certification. Additionally, this PR fixes the existing BoringCrypto build which was incorrectly passing `boringcrypto` as a Go build tag instead of setting `GOEXPERIMENT=boringcrypto` as an environment variable.

Key changes:

- **New Makefile build targets**: Added `go-build-fips` (native Go FIPS) and `go-build-fips-boringcrypto` (legacy BoringCrypto) targets alongside the existing `go-build`. The build recipe is now parameterized via `MAKE_BUILD_RECIPE` so that Dockerfiles and drivah configs can select the appropriate build variant.
- **Native FIPS build flags**: The `go-build-fips` target sets `GOFIPS140=v1.0.0` and injects the linker flag `-X runtime.godebugDefault=fips140=on` to enable FIPS 140-3 mode at runtime.
- **FIPS binary verification script** (`.buildkite/scripts/build/verify-fips.sh`): A new script that verifies the built binary contains the expected FIPS build settings. It supports both `native` and `boringcrypto` verification modes.
- **Runtime FIPS logging** (`cmd/manager/main.go`): At startup, the operator now logs whether it is running in FIPS mode and which type (`native` or `boringcrypto`). Detection uses `crypto/fips140.Enabled()` for native mode and checks for the `boringcrypto` GOEXPERIMENT build setting for the legacy path.
- **Dockerfile updates**: Both `Dockerfile` and `Dockerfile-ubi` now accept `MAKE_BUILD_RECIPE` as a build arg and copy the FIPS verification script into the build context, removing the hardcoded `CGO_ENABLED=0` so the build recipe controls it.
- **Fix BoringCrypto build**: The BoringCrypto FIPS build now correctly uses `GOEXPERIMENT=boringcrypto` as an environment variable during compilation (via `CGO_ENABLED=1`) instead of passing it as a Go build tag. This ensures the BoringCrypto experiment is properly activated by the Go toolchain.
- **Drivah config generation** (`build/gen-drivah.toml.sh`): Updated to pass the appropriate `MAKE_BUILD_RECIPE` to each image flavor. FIPS flavors currently use `go-build-fips-boringcrypto` with a TODO to switch to `go-build-fips` once the native Go FIPS module is certified.

### `GOFIPS140` (environment variable at build time)

Controls which version of the Go FIPS 140-3 cryptographic module is bundled into the binary.

| Value | Description |
|-------|-------------|
| *(unset / empty)* | No FIPS module is included. Standard Go crypto is used. |
| `latest` | Uses the latest available FIPS 140-3 module snapshot shipped with the Go toolchain. |
| `inprocess` | Uses the FIPS module currently being developed/validated (not yet certified). Intended for module developers. |
| `v1.0.0` | **Used in this PR.** Pins the certified FIPS 140-3 module at version v1.0.0. This is the first release of the Go native FIPS module. |

### `runtime.godebugDefault=fips140=...` (linker flag, runtime behavior)

Set via `-ldflags "-X runtime.godebugDefault=fips140=<value>"`. Controls the FIPS 140-3 runtime enforcement policy when the binary starts.

| Value | Description |
|-------|-------------|
| `off` | FIPS mode is disabled at runtime even if a FIPS module was compiled in. Crypto operations use the standard code paths. |
| `on` | **Used in this PR.** Enables FIPS 140-3 mode. All crypto operations are routed through the FIPS module. If the module's integrity checks fail at startup the program exits. |
| `only` | Strict FIPS mode. Same as `on`, but additionally *restricts* the program to only FIPS-approved algorithms. Any attempt to use a non-approved algorithm (e.g. ChaCha20-Poly1305) returns an error. |
| `debug` | Enables FIPS mode like `on`, but logs detailed diagnostics about FIPS module loading and self-tests. Intended for troubleshooting. |

_____

# Fix `boringcrypto` images

## Summary
The FIPS binary built with `boringcrypto` is now dynamically linked, so its base image has been changed from `wolfi/static` to `chainguard/glibc-dynamic`.

A single multi-stage Dockerfile is now used for all image variants. Each `drivah` toml (or `docker build` command) specifies `--target` to select the desired image stage.

The table below shows the base image and Dockerfile target for each variant:

| ECK Image                   | Base Image                     | Dockerfile stage | ECK Binary type |
|-----------------------------|--------------------------------|------------------|-----------------|
| `eck/eck-operator`          | `wolfi/static` _(unchanged)_   | `static`         | `static` _(unchanged)_ |
| `eck/eck-operator-fips`     | `chainguard/glibc-dynamic` _(new)_     | `dynamic`        | `dynamically linked` (CGO) _(new)_ |
| `eck/eck-operator-ubi`      | `ubi9/ubi-micro` _(unchanged)_ | `ubi`            | `static` _(unchanged)_ |
| `eck/eck-operator-ubi-fips` | `ubi9/ubi-micro` _(unchanged)_ | `ubi`            | `dynamically linked` (CGO) _(new)_ |

> Once the transition to native is complete, the CGO will be disabled, and a static binary can be provided in FIPS images as well. When this happens, we can drop `chainguard/glibc-dynamic` entirely. 

A new makefile recipe was added to build the images locally without pushing to registry.

Example:

```shell
make \
  DOCKER_BUILD_TARGET='static' \
  BUILD_PLATFORM='linux/arm64' \
  OPERATOR_IMAGE='eck:latest' \
  LICENSE_PUBKEY=eck-pubkey.key \
  MAKE_BUILD_RECIPE=go-build \
  docker-build-operator

make \
  DOCKER_BUILD_TARGET='static' \
  BUILD_PLATFORM='linux/arm64' \
  OPERATOR_IMAGE='eck:fips' \
  LICENSE_PUBKEY=eck-pubkey.key \
  MAKE_BUILD_RECIPE=go-build-fips \
  docker-build-operator

make \
  DOCKER_BUILD_TARGET='dynamic' \
  BUILD_PLATFORM='linux/arm64' \
  OPERATOR_IMAGE='eck:fips-boringcrypto' \
  LICENSE_PUBKEY=eck-pubkey.key \
  MAKE_BUILD_RECIPE=go-build-fips-boringcrypto \
  docker-build-operator

make \
  DOCKER_BUILD_TARGET='ubi' \
  BUILD_PLATFORM='linux/arm64' \
  OPERATOR_IMAGE='eck:ubi' \
  LICENSE_PUBKEY=eck-pubkey.key \
  MAKE_BUILD_RECIPE=go-build \
  docker-build-operator

make \
  DOCKER_BUILD_TARGET='ubi' \
  BUILD_PLATFORM='linux/arm64' \
  OPERATOR_IMAGE='eck:fips-ubi' \
  LICENSE_PUBKEY=eck-pubkey.key \
  MAKE_BUILD_RECIPE=go-build-fips \
  docker-build-operator

make \
  DOCKER_BUILD_TARGET='ubi' \
  BUILD_PLATFORM='linux/arm64' \
  OPERATOR_IMAGE='eck:fips-boringcrypto-ubi' \
  LICENSE_PUBKEY=eck-pubkey.key \
  MAKE_BUILD_RECIPE=go-build-fips-boringcrypto \
  docker-build-operator
```

### DIFF
--- a/.buildkite/scripts/common/operator-image.sh
+++ b/.buildkite/scripts/common/operator-image.sh
@@ -63,7 +63,7 @@ operator::set_build_flavors_var() {
             merge-xyz)       BUILD_FLAVORS="eck,eck-dev,eck-fips,eck-ubi,eck-ubi-fips" ;;
             merge-main)      BUILD_FLAVORS="eck,eck-dev" ;;
             *-test-snapshot) BUILD_FLAVORS="eck,eck-dev" ;;
-            pr-*)            BUILD_FLAVORS="eck,eck-dev,eck-fips,eck-ubi,eck-ubi-fips" ;;
+            pr-*)            BUILD_FLAVORS="eck" ;;
             dev)             BUILD_FLAVORS="dev" ;;
             *)               echo "error: trigger '$trigger' not supported"; exit ;;
         esac


prodsecmachine:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/3527c7db-a9d7-44a6-9d18-7d0e8678104a?refs=364f73bc1bd8d5e6eb66e090c131e0b88be5c0bf&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/3527c7db-a9d7-44a6-9d18-7d0e8678104a/license?refs=364f73bc1bd8d5e6eb66e090c131e0b88be5c0bf&source=prChecksComment) |

moukoublen:
@simitt you are right, I kept boringcrypt (after a fix), and we can switch to native once the golang native FIPS gets certified. Switching should take just changing a string that reflects the makefile recipe. I added a to-do to the specific script that we need to change to native when the time comes.

build/Dockerfile
```diff
@@ -6,6 +6,7 @@ ARG SHA1
 ARG SNAPSHOT
 ARG GO_TAGS
 ARG LICENSE_PUBKEY
+ARG MAKE_BUILD_RECIPE
```
pebrc: `MAKE_BUILD_RECIPE` has no default value. If someone builds this Dockerfile locally without passing `--build-arg MAKE_BUILD_RECIPE=...`, line 28 expands to just `make` which runs the default `all` target (build + lint + test + ...) instead of just compiling the binary.

Suggestion: add a safe default:
```dockerfile
ARG MAKE_BUILD_RECIPE=go-build
```

Same applies to `Dockerfile-ubi`.

.buildkite/scripts/build/verify-fips.sh
```diff
@@ -0,0 +1,62 @@
+#!/usr/bin/env bash
+
+# Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+# or more contributor license agreements. Licensed under the Elastic License 2.0;
+# you may not use this file except in compliance with the Elastic License 2.0.
+
+set -euo pipefail
+
+if [[ $# -lt 1 || $# -gt 2 ]]; then
```
pebrc: This guard allows exactly 1 argument (`$# -lt 1 || $# -gt 2` passes when count is 1), but the script unconditionally uses both `$1` and `$2`. With 1 argument, `binary="${2}"` is empty and the file-existence check produces a confusing `Error: file not found: ` message.

Since both arguments are always required, this should be:
```bash
if [[ $# -ne 2 ]]; then
```

.buildkite/scripts/common/operator-image.sh
```diff
@@ -63,7 +63,7 @@ operator::set_build_flavors_var() {
             merge-xyz)       BUILD_FLAVORS="eck,eck-dev,eck-fips,eck-ubi,eck-ubi-fips" ;;
             merge-main)      BUILD_FLAVORS="eck,eck-dev" ;;
             *-test-snapshot) BUILD_FLAVORS="eck,eck-dev" ;;
-            pr-*)            BUILD_FLAVORS="eck" ;;
+            pr-*)            BUILD_FLAVORS="eck,eck-dev,eck-fips,eck-ubi,eck-ubi-fips" ;;
```
pebrc: Commit message says "temporary build all flavors on pr-*" — this should be reverted to `BUILD_FLAVORS="eck"` before merge to avoid building 5 flavors on every PR branch.

cmd/manager/main.go
```diff
@@ -1138,3 +1141,29 @@ func reconcileWebhookCertsAndAddController(ctx context.Context, mgr manager.Mana
...
+				return
+			}
+			break
+		}
+	}
+
+	if fips140.Enabled() {
+		fipsType = "native"
+	}
+}
```
pebrc, line 1167: nit: The `defer` pattern here (declare var → defer the log → set var and return) works but is a bit unusual for a function this small. Since there are only two code paths that set `fipsType`, it would read more directly to just log inline:

```go
func fipsLog() {
	if buildInfo, ok := debug.ReadBuildInfo(); ok {
		for _, s := range buildInfo.Settings {
			if s.Key == "GOEXPERIMENT" {
				if strings.Contains(s.Value, "boringcrypto") {
					log.Info("operator runs in FIPS mode", "type", "boringcrypto")
					return
				}
				break
			}
		}
	}

	if fips140.Enabled() {
		log.Info("operator runs in FIPS mode", "type", "native")
	}
}
```

Also worth considering: logging when FIPS is *not* enabled (e.g. `log.Info("operator runs without FIPS mode")`) so operators can always confirm the FIPS status from startup logs.

build/Dockerfile
moukoublen: Fixed in [9cb9f7f](https://github.com/elastic/cloud-on-k8s/pull/9263/commits/9cb9f7f6daedafe03deb53c17c0d4d2578c618a4)

cmd/manager/main.go
moukoublen: Done in [9cb9f7f](https://github.com/elastic/cloud-on-k8s/pull/9263/commits/9cb9f7f6daedafe03deb53c17c0d4d2578c618a4)

.buildkite/scripts/common/operator-image.sh
moukoublen: Yes that was for testing, which is now done. Restored in [74fa254](https://github.com/elastic/cloud-on-k8s/pull/9263/commits/74fa2546d5b76ffa46ed2cd75585b2db3bdb3f57)

build/Dockerfile
```diff
@@ -53,3 +57,65 @@ COPY *.txt /licenses/
...
+      org.opencontainers.image.version="$VERSION" \
+      org.opencontainers.image.url="https://github.com/elastic/cloud-on-k8s/"
+
+COPY --from=builder /go/src/github.com/elastic/cloud-on-k8s/elastic-operator /elastic-operator
+COPY --from=builder /go/src/github.com/elastic/cloud-on-k8s/eck.yaml         /conf/eck.yaml
+
+# Copy NOTICE.txt and LICENSE.txt into the image
+COPY *.txt /licenses/
+
+USER 65532
```
moukoublen, line 89: This is because `wolfi-base` does not set the user (and thus is root). The value is `65532` to be on a par with the `wolfi/static` image.

pebrc: Should be a comment in the docker file?

build/Dockerfile
```diff
@@ -53,3 +57,65 @@ COPY *.txt /licenses/
...
+# dynamic - Image for non-static binary (FIPS boringcrypto)
+FROM docker.elastic.co/wolfi/wolfi-base:latest@sha256:b2134dbfbcfb987eec9c249fc81111963de596ff76098775c2444a2869401d9c AS dynamic
+
+ARG VERSION
+
+# Add common ECK labels and OCI annotations to image
+LABEL io.k8s.description="Elastic Cloud on Kubernetes automates the deployment, provisioning, management, and orchestration of Elasticsearch, Kibana, APM Server, Beats, and Enterprise Search on Kubernetes" \
+      io.k8s.display-name="Elastic Cloud on Kubernetes" \
+      org.opencontainers.image.authors="eck@elastic.co" \
+      org.opencontainers.image.base.name="gcr.io/distroless/static:nonroot" \
```
pebrc: This seems stale?

build/Dockerfile
```diff
@@ -53,3 +57,65 @@ COPY *.txt /licenses/
 
 ENTRYPOINT ["/elastic-operator"]
 CMD ["manager"]
+# ---------------------------------------------
+
+# ---------------------------------------------
+# dynamic - Image for non-static binary (FIPS boringcrypto)
+FROM docker.elastic.co/wolfi/wolfi-base:latest@sha256:b2134dbfbcfb987eec9c249fc81111963de596ff76098775c2444a2869401d9c AS dynamic
```
pebrc: I wonder if `wolfi/glibc-dynamic` would be a better fit. But I don't see it in our mirror. So maybe it is not available to us.

pebrc (commented):
LGTM (just a few nits)

pebrc:
Not a problem introduced in  your PR but I think it would be good to have the different image flavours we build to at least go through some smoke test, either our e2e `TestSmoke` or even more minimal `docker run <image> --version` to see if the build artifact is actually runnable and in case of the glibc based images the dynamic libraries resolve etc. Right now we are pretty much flying blind. This could be follow up work I don't want to complicate things in this PR.

moukoublen:
@pebrc a `docker run <image> --version` for each image should be easy to implement. I will add it.

build/Dockerfile
moukoublen: I replaced it with `cgr.dev/chainguard/glibc-dynamic:latest` in  [8b012e3](https://github.com/elastic/cloud-on-k8s/pull/9263/commits/8b012e315ccc731c748269eeaf4be2dfa97c5381)

build/Dockerfile
moukoublen: Done in [8b012e3](https://github.com/elastic/cloud-on-k8s/pull/9263/commits/8b012e315ccc731c748269eeaf4be2dfa97c5381)

build/Dockerfile
moukoublen: Fixed in [8b012e3](https://github.com/elastic/cloud-on-k8s/pull/9263/commits/8b012e315ccc731c748269eeaf4be2dfa97c5381)

moukoublen:
@pebrc I added a step in the pipeline that verifies the images. ([example](https://buildkite.com/elastic/cloud-on-k8s-operator/builds/13800#019d29b6-c84d-477f-9c0e-c4aa15ca04d1))

PR #9270, opened 2026-03-27
fix(deployer): fix local kind/k3d execution on macOS

pebrc:
## Summary
- Replace hardcoded Docker socket discovery with `docker info` availability check — the resolved host-side socket path was incorrectly used as the container-side mount target, where it has no meaning
- Fix both kind and k3d drivers to hardcode `/var/run/docker.sock` on both sides of the bind mount and set `DOCKER_HOST` explicitly inside the container
- Replace `kubectl config view --flatten` kubeconfig merge with `k8s.io/client-go/tools/clientcmd` struct-level merge — new entries take precedence over stale ones (e.g. recreated clusters on different ports), and writes are atomic via temp file + rename
- Add `GetCredentials()` call after kind cluster creation so the kubeconfig is immediately available

## Test plan
- [x] CI `kind/TestSmoke` passes (automatically triggered by this PR)
- [x] Local `kind` cluster creation works on macOS
- [x] Local `kind` cluster re-creation correctly updates kubeconfig with new API server port

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pkg/controller/association/secret.go
+++ b/pkg/controller/association/secret.go
@@ -183,7 +183,7 @@ func (r UnmanagedAssociationConnectionInfo) Request(path string, out any) error 
 		for _, c := range caCerts {
 			certPool.AddCert(c)
 		}
-		httpClient.Transport = &http.Transport{TLSClientConfig: &tls.Config{RootCAs: certPool}}
+		httpClient.Transport = &http.Transport{TLSClientConfig: &tls.Config{RootCAs: certPool}} //nolint:gosec
 	}
 
 	resp, err := httpClient.Do(req)
--- a/pkg/controller/common/statefulset/pod.go
+++ b/pkg/controller/common/statefulset/pod.go
@@ -69,10 +69,10 @@ func PodReconciliationDone(ctx context.Context, c k8s.Client, statefulSet appsv1
 
 		var reason strings.Builder
 		if len(pendingCreations) > 0 {
-			fmt.Fprintf(&reason, ", creations: %s", pendingCreations)
+			reason.WriteString(fmt.Sprintf(", creations: %s", pendingCreations))
 		}
 		if len(pendingDeletions) > 0 {
-			fmt.Fprintf(&reason, ", deletions: %s", pendingDeletions)
+			reason.WriteString(fmt.Sprintf(", deletions: %s", pendingDeletions))
 		}
 
 		return false, reason.String(), nil
--- a/test/e2e/test/http_client.go
+++ b/test/e2e/test/http_client.go
@@ -29,6 +29,7 @@ func NewHTTPClient(caCerts []*x509.Certificate) *http.Client {
 		certPool.AddCert(c)
 	}
 
+	//nolint:gosec  // [G402: TLS MinVersion too low] is not a concern here as it is test code.
 	transport.TLSClientConfig = &tls.Config{
 		RootCAs: certPool,
 	}


prodsecmachine:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/79260291-38a7-4f2c-a745-f68221aa6354?refs=d3b4d84db6b3663a86c678d6beaabd5e71627c1e&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/79260291-38a7-4f2c-a745-f68221aa6354/license?refs=d3b4d84db6b3663a86c678d6beaabd5e71627c1e&source=prChecksComment) |

pebrc:
buildkite test this -f E2E_PROVIDER=k3d,TESTS_MATCH=TestSmoke

hack/deployer/runner/kubeconfig.go
kvalliyurnatt, line 20: nit: could we add a small unit test for mergeKubeconfig function to make sure it does not corrupt an existing config.

pebrc: Added in c3b614f88 — extracted `mergeKubeconfigFiles` for testability and added table-driven tests covering stale entry override, preservation of unrelated clusters, current-context update, and file permission verification.

hack/deployer/runner/kind.go
```diff
@@ -231,13 +231,19 @@ func (k *KindDriver) cmd(args ...string) *exec.Command {
...
+	// We need the docker socket so that kind can bootstrap.
+	// The source path /var/run/docker.sock is evaluated by the Docker daemon (inside the Colima/Docker VM),
+	// so it always resolves to the actual daemon socket regardless of the macOS-visible socket path.
+	// DOCKER_HOST is explicitly set to override any host env var pointing at a macOS-side socket path.
 	// --userns=host to support Docker daemon host configured to run containers only in user namespaces
 	cmd := exec.NewCommand(`docker run --rm \
 		--userns=host \
 		-v {{.SharedVolume}}:/home \
-		-v /var/run/docker.sock:` + k.dockerSocket + ` \
+		-v /var/run/docker.sock:/var/run/docker.sock \
```
moukoublen, line 242: [Question]: since the `/var/run/docker.sock` is a symlink (in most of macos docker desktop) to something like `/Users/username/.docker/run/docker.sock`, how does linking a symlink inside the container work?

pebrc: The source path in a `-v` bind mount is not resolved on the macOS host filesystem — it is resolved by the Docker daemon, which runs inside the Linux VM (Colima or Docker Desktop). Inside that VM, `/var/run/docker.sock` is the actual Unix socket, not a symlink, so the mount works correctly regardless of what the macOS-side socket path looks like. This also holds regardless of the virtualisation backend (HyperKit, QEMU, Apple VZ) since the Docker daemon always listens on `/var/run/docker.sock` inside the VM by default — it's a daemon configuration, not something that varies with the hypervisor.

On CI (Linux) the same path is evaluated directly on the Linux host where the daemon runs, and it exists there as a real socket too.

We verified this works locally with both Colima and Docker Desktop, and the CI kind/TestSmoke and k3d/TestSmoke jobs both passed.

moukoublen: In general any `-v` mounts the macos filesystem, I thought the only expection is when the [docker.sock.raw](https://docs.docker.com/extensions/extensions-sdk/guides/use-docker-socket-from-backend/#:~:text=you%20can%20use-,/var/run/docker.sock.raw,-.%20Docker%20Desktop%20mounts) is used which is mac specific and in that case docker Desktop mounts the socket that lives in the Desktop VM, and not from the host.

Apparently, it works with the `/var/run/docker.sock` as well? I am wondering why. I remember it crashing two months ago, especially because of the symlink on macos. 

But no worries -the doc is about the extensions, perhaps they overwrite the `/var/run/docker.sock` in cli-, if it works locally on the macos, then we don't need to bother.

pebrc: Good point — the behaviour actually differs between runtimes. With Colima, `/var/run/docker.sock` doesn't exist on the macOS host at all, so the source path is resolved inside the VM where it is the real daemon socket. With Docker Desktop, it does exist on macOS as a symlink, and Docker follows it on the host side to mount the proxied socket that Docker Desktop uses to forward to the VM daemon. Either way, the container gets a working daemon socket at `/var/run/docker.sock`.

Worth noting: the bug in the old code wasn't in the source path (that was always hardcoded to `/var/run/docker.sock`). It was in using the resolved macOS path (e.g. `~/.colima/default/docker.sock`) as the **container-side destination**, which meant kind/k3d couldn't find the socket at the expected `/var/run/docker.sock` path inside the container.

pebrc: I did a bit more digging and the Docker daemon source code confirms the behaviour. In [`moby/moby daemon/volume/mounts/linux_parser.go`](https://github.com/moby/moby/blob/master/daemon/volume/mounts/linux_parser.go), the daemon calls `os.Stat()` on the bind mount source path in its own filesystem — the CLI never checks if the source exists, it just sends the path string over the API. This is also confirmed in [moby/moby#16860](https://github.com/moby/moby/issues/16860): *"the left hand side of the bind mount is always on the host machine, e.g. where the docker daemon is running."*

So for **Colima**: the daemon runs inside the Lima VM, it resolves `/var/run/docker.sock` in the VM filesystem where the socket exists. The path doesn't exist on macOS and doesn't need to — macOS is never consulted.

For **Docker Desktop**: the daemon also runs in a VM, but Docker Desktop shares the macOS filesystem into the VM via virtiofs. So `/var/run/docker.sock` (the macOS symlink) is visible inside the VM through that file share, and the daemon follows it.

Both paths lead to the same outcome: the daemon's own socket gets bind-mounted into the container, giving kind/k3d access to the Docker API to create the k8s node containers as siblings.

moukoublen: oh I totally missed that the arguments of `-v` were in the opposite direction 🤣 

It's `<host>:<container>`, and in our code, it was the opposite. Now I understand the issue we had before (because we mount the whole home directory), everything makes sense now.

PR #9287, opened 2026-03-31
Fips - keystore password support

### LINKED ISSUE #9171: ECK to natively support password-protected keystore creation for FIPS deployments
When FIPS mode is enabled, ECK should:

1. Generate a FIPS-compliant keystore password (>= 14 characters) and store it in an operator-managed Secret
2. Create the keystore with the -p flag using the generated password
3. Configure `ES_KEYSTORE_PASSPHRASE_FILE` in the Elasticsearch container
4. Mount the password Secret appropriately

This would provide a seamless FIPS experience without requiring any manual podTemplate overrides or user-managed secrets.

naemono:
# Add operator-managed FIPS keystore password support for Elasticsearch

Fixes #9171

## Summary

This PR adds native support for FIPS-compliant Elasticsearch keystore password management in ECK. When FIPS mode is enabled in Elasticsearch configuration (`xpack.security.fips_mode.enabled: true`), the operator now automatically generates, stores, and configures a password-protected keystore — eliminating the need for manual `podTemplate` overrides.

## Key Changes

- **Auto-detection**: Detects FIPS mode from `xpack.security.fips_mode.enabled` in merged ES config (ES 9.4.0+)
- **Operator-managed secret**: Creates `{es}-es-fips-keystore-password` Secret with a 24-character password (exceeds FIPS minimum of 14)
- **Secret mounting**: Mounts the password Secret read-only on both keystore init and main Elasticsearch containers
- **Runtime handling**: Sets `KEYSTORE_PASSWORD_FILE` environment variable pointing to the mounted Secret file
- **User override support**: Respects existing `KEYSTORE_PASSWORD`, `KEYSTORE_PASSWORD_FILE`, or `ES_KEYSTORE_PASSPHRASE_FILE` if user provides them
- **Version-gated**: Only activates for ES 9.4.0+ (required for `0440/0640` file mode support with `fsGroup`)
- **Upgrade-safe**: Non-FIPS clusters experience zero changes (byte-identical pod specs)

## Test Plan

<details>
<summary>Click to expand: Edge cases and scenarios tested</summary>

### Core Functionality
- Managed FIPS path — operator generates secret, mounts volumes, injects `KEYSTORE_PASSWORD_FILE`, renders FIPS-aware init script
- FIPS with no secure settings — init container still created for empty password-protected keystore (required for FIPS compliance)

### User Override Scenarios
- User provides `KEYSTORE_PASSWORD_FILE` — operator skips managed secret entirely
- User provides `KEYSTORE_PASSWORD` — operator skips managed secret entirely

### Configuration Edge Cases
- Mixed NodeSets (some FIPS, some not) — emits admission warning, applies cluster-wide FIPS behavior
- ES version < 9.4.0 with FIPS enabled — feature stays disabled (version gate)
- Upgrade ES from <9.4.0 to >=9.4.0 — managed FIPS behavior activates post-upgrade

### Operator Upgrade Safety
- Non-FIPS cluster upgrade — no rolling restart, no secret creation
- Existing manual workaround clusters — operator respects user overrides, no changes

### Automated Testing
- Unit/controller tests for `pkg/apis/elasticsearch/v1`, `pkg/controller/common/keystore`, `pkg/controller/elasticsearch/fips`, validation, nodespec, and driver packages
- E2E tests in `test/e2e/es` with FIPS focus

</details>

## Usage Example

```yaml
kind: Elasticsearch
metadata:
  name: fips-cluster
spec:
  version: 9.4.0
  nodeSets:
  - name: default
    count: 3
    config:
      node.roles: ["master", "data", "ingest"]
      xpack.security.fips_mode.enabled: true
```

No additional configuration required. The operator automatically manages the keystore password.

## Testing with Elasticsearch SNAPSHOT Versions

This feature requires ES 9.4.0+ and has been tested against SNAPSHOT builds. To test with SNAPSHOT versions, you must build and deploy a `release-dev` operator flavor:

```bash
# Fetch dev public key from Vault (base64 decode required)
vault read -field=dev-pubkey secret/ci/elastic-cloud-on-k8s/license | base64 --decode > build/dev-license.key

# Build  and push a release-dev operator image
BUILD_FLAVORS=release-dev drivah build ./build --push

Ensure you apply the dev-enterprise-license.json to ECK.
```


prodsecmachine:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/71820560-066b-4fb8-9705-ff696804a314?refs=67f1358fe391ac7442cfd17ec08fad5bd65520df&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/71820560-066b-4fb8-9705-ff696804a314/license?refs=67f1358fe391ac7442cfd17ec08fad5bd65520df&source=prChecksComment) |

naemono:
buildkite test this -f p=gke,E2E_TAGS=es,t=TestFIPS -m s=9.4.0-SNAPSHOT

pebrc:
@naemono this PR is still in draft mode is this intentional? Is it ready for reviews?

pkg/controller/elasticsearch/keystorepassword/keystore_password.go
```diff
@@ -0,0 +1,153 @@
...
+	"github.com/elastic/cloud-on-k8s/v3/pkg/utils/maps"
+)
+
+const (
+	// KeystorePasswordKey is the key used in the FIPS keystore password secret.
+	KeystorePasswordKey = "keystore-password"
+
+	// VolumeName is the source Secret volume name used for init-container consumption.
+	VolumeName = "fips-keystore-password"
+	// MountPath is the source Secret mount path used for init-container consumption.
```
pebrc: **Naming: `fips-*` couples the trigger to the mechanism.**

Password-protecting the Elasticsearch keystore is a general capability — FIPS just happens to mandate it. Naming the Secret (`fips-keystore-password`), volume, mount path, and package all with the `fips-` prefix precludes reuse for non-FIPS password-protected keystores and is potentially misleading at the API boundary (a user sees `{es}-es-fips-keystore-password` in their namespace).

More importantly, the Secret name becomes a compatibility constraint once released — renaming it later would require a migration.

Consider more neutral names: `keystore-password` for the Secret/volume/path, with FIPS-specific logic as a thin detection layer that decides *when* to activate the general keystore password machinery.

pkg/controller/elasticsearch/driver/shared/reconcile.go
```diff
@@ -467,6 +488,35 @@ func apiKeyStoreSecretSource(ctx context.Context, es *esv1.Elasticsearch, c k8s.
...
+	}
+	if !fipsEnabled {
+		return nil, fips.DeleteKeystorePasswordSecret(ctx, client, es)
+	}
+	if esVersion.LT(esversion.FIPSKeystorePasswordMinVersion) {
+		return nil, nil
+	}
+	userOverride, err := settings.AnyNodeSetHasUserProvidedKeystorePassword(ctx, client, es.Namespace, es.Spec.NodeSets)
+	if err != nil || userOverride {
+		return nil, err
```
pebrc: **Bug: Orphaned Secret when user adds an override after operator already created one.**

When `userOverride` is true, this returns `nil, nil` — it neither creates nor deletes the operator-managed Secret. Scenario:

1. FIPS enabled, no user override → operator creates `{es}-es-fips-keystore-password`
2. User adds `KEYSTORE_PASSWORD_FILE` env var → next reconcile: `userOverride` is true → returns `nil, nil` → Secret is orphaned

The `!fipsEnabled` path correctly calls `DeleteKeystorePasswordSecret` (line 508), but this path doesn't. Suggest:

```go
if userOverride {
    return nil, fips.DeleteKeystorePasswordSecret(ctx, client, es)
}
if err != nil {
    return nil, err
}
```

pkg/controller/elasticsearch/validation/warnings.go
```diff
@@ -79,6 +84,68 @@ func validateClientAuthentication(config *common.CanonicalConfig, index int) fie
...
+	parseableNodeSetCount           int
+	hasUserProvidedKeystorePassword bool
+}
+
+// fipsWarnings emits warnings when FIPS mode is inconsistent across NodeSets or
+// when managed keystore passwords are unsupported for the requested version.
+// It uses the same keystore password override detection as reconciliation
+// (explicit env and envFrom), which may perform API reads against c.
+func fipsWarnings(ctx context.Context, c k8s.Client, es esv1.Elasticsearch) (field.ErrorList, error) {
+	state, err := collectFIPSState(ctx, c, es)
```
pebrc, line 103: **Missing `envFrom` Secret/ConfigMap can block entire admission.**

`fipsWarnings` → `collectFIPSState` → `AnyNodeSetHasUserProvidedKeystorePassword` does live API reads against Secrets/ConfigMaps referenced in `envFrom`. If a referenced Secret doesn't exist and isn't marked `optional: true`, the error propagates up through `ValidateElasticsearch` (webhook.go:98-99) and **blocks the entire admission request**.

This is overly harsh for a warning-only code path. Users with external-secrets operators or GitOps patterns commonly reference Secrets that don't exist yet at admission time.

Consider degrading gracefully: treat "not found" (even for non-optional refs) as "can't determine override status, skip the managed-keystore warning" rather than failing admission.

pkg/controller/elasticsearch/fips/fips_keystore_password.go
```diff
@@ -0,0 +1,153 @@
...
+	KeystorePasswordKey = "keystore-password"
+
+	// VolumeName is the source Secret volume name used for init-container consumption.
+	VolumeName = "fips-keystore-password"
+	// MountPath is the source Secret mount path used for init-container consumption.
+	MountPath = "/mnt/elastic-internal/fips-keystore-password"
+	// PasswordFile is the mounted Secret file path read by the keystore init container.
+	PasswordFile = MountPath + "/keystore-password"
+
+	generatedPasswordLength = 24
```
barkbay: The FIPS keystore password length is hardcoded to 24, but the operator already has a `--password-length` flag (`operator.PasswordLengthFlag`, default 24) used by `password.NewGenerator` for other generated secrets (e.g. the `elastic` user password).

Should this use the same flag (or at least `max(flagValue, 14)` to honor the FIPS minimum) so that users who customize `--password-length` get consistent behavior across all operator-managed passwords?

pkg/controller/elasticsearch/keystorepassword/keystore_password.go
```diff
@@ -0,0 +1,153 @@
...
+		passwordBytes, err = password.RandomBytesWithoutSymbols(generatedPasswordLength)
+		if err != nil {
+			return nil, fmt.Errorf("while generating fips keystore password: %w", err)
+		}
+	}
+
+	expected := corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace:   secretName.Namespace,
+			Name:        secretName.Name,
```
barkbay, line 82: This secret contains a password but does not carry the `eck.k8s.elastic.co/credentials: "true"` label that other sensitive secrets use (via `labels.AddCredentialsLabel()`). Adding it would keep the convention consistent and let users/tooling identify all credential-bearing secrets uniformly.

```go
Labels: labels.AddCredentialsLabel(maps.Merge(label.NewLabels(k8s.ExtractNamespacedName(&es)), meta.Labels)),
```

pkg/controller/elasticsearch/nodespec/podspec.go
```diff
@@ -241,6 +245,9 @@ func buildAnnotations(
 	if keystoreResources != nil {
 		// resource version of the secure settings secret to rotate the pod on secure settings change
 		_, _ = configHash.Write([]byte(keystoreResources.Hash))
+		if keystoreResources.FIPSKeystorePasswordSecretResourceVersion != "" {
+			_, _ = configHash.Write([]byte(keystoreResources.FIPSKeystorePasswordSecretResourceVersion))
```
barkbay: The existing secure settings hash (`keystoreResources.Hash`) uses a **content hash** (`hash.HashObject(secureSettingsSecret.Data)` in `user_secret.go:95`), meaning only actual data changes trigger a rolling restart.

The FIPS password uses `ResourceVersion` instead, which changes on **any** secret mutation — including metadata-only changes like adding an annotation or label. This means an external controller touching the secret, or a delete/recreate with the same password, would trigger an unnecessary rolling restart.

Should this use `hash.HashObject(secret.Data)` to be consistent with the existing pattern and avoid spurious restarts?

pkg/controller/common/keystore/initcontainer.go
```diff
@@ -55,6 +57,23 @@ fi
 {{ end -}}
 echo "Initializing keystore."
 
+{{ if .FIPSKeystorePasswordPath -}}
+# FIPS: remove any existing keystore to avoid interactive "Overwrite?" prompt
+rm -f {{ .KeystoreVolumePath }}/elasticsearch.keystore
+
```
pebrc: **ES-specific assumptions in the shared `common/keystore` package.**

This line hardcodes `elasticsearch.keystore`, and line 67 appends `-p` to `KeystoreCreateCommand`. Both are Elasticsearch-specific — Kibana's keystore file is `kibana.keystore` and its password flag is `--password`, not `-p`.

Since `KeystoreCreateCommand` and `KeystoreAddCommand` are already parameterized per-caller, this password-protected branch breaks the abstraction by baking in ES conventions.

Two options:
1. Move the password-protected branch out of the shared template into the ES package tree (simpler, honest about current scope).
2. Parameterize properly: add a `KeystoreFileName` field and a `KeystoreCreatePasswordCommand` (or similar) instead of appending `-p` to the existing command.

pkg/controller/elasticsearch/fips/fips_keystore_password.go
```diff
@@ -0,0 +1,153 @@
...
+	if err := client.IgnoreNotFound(c.Get(ctx, secretName, &existingSecret)); err != nil {
+		return nil, err
+	}
+
+	passwordBytes := existingSecret.Data[KeystorePasswordKey]
+	if len(passwordBytes) == 0 {
+		var err error
+		passwordBytes, err = password.RandomBytesWithoutSymbols(generatedPasswordLength)
+		if err != nil {
+			return nil, fmt.Errorf("while generating fips keystore password: %w", err)
```
barkbay: FIPS is an acronym — should be uppercase in error messages for consistency with the rest of the codebase.

```suggestion
			return nil, fmt.Errorf("while generating FIPS keystore password: %w", err)
```

pkg/controller/elasticsearch/driver/shared/reconcile.go
```diff
@@ -467,6 +488,35 @@ func apiKeyStoreSecretSource(ctx context.Context, es *esv1.Elasticsearch, c k8s.
...
+	esVersion version.Version,
+	meta metadata.Metadata,
+	policyConfig nodespec.PolicyConfig,
+) (*corev1.Secret, error) {
+	fipsEnabled, err := settings.AnyNodeSetFIPSEnabled(es.Spec.NodeSets, policyConfig.ElasticsearchConfig)
+	if err != nil {
+		return nil, err
+	}
+	if !fipsEnabled {
+		return nil, fips.DeleteKeystorePasswordSecret(ctx, client, es)
```
barkbay: There's a potential ordering issue when FIPS is disabled. The secret is deleted eagerly here, but the StatefulSet pod template is updated much later in the reconciliation (after `ReconcileClusterUUID`, `stackmon`, associations, then `reconcileNodeSpecs` → `BuildPodTemplateSpec` → StatefulSet update).

If anything fails between the secret deletion and the StatefulSet update, the StatefulSet still has the old pod template referencing the now-deleted FIPS secret volume. Running pods are fine (kubelet cache), but if a pod is evicted, OOM-killed, or a node fails during that window, the replacement pod will be stuck in `ContainerCreating` because the secret volume mount references a non-existent secret.

Low probability but plausible on busy clusters with node maintenance or memory pressure. Consider deferring the secret deletion until after the StatefulSet has been updated with a pod spec that no longer references it — or returning a marker that triggers deletion later in the reconciliation.

naemono:
> @naemono this PR is still in draft mode is this intentional? Is it ready for reviews?

~~No, I moved it back to draft mode yesterday, as I noticed issues I hadn't previously seen and wanted more time to ensure it was in good shape.~~

Since reviews have already begun on this, and I've made my changes, I'm opening this back up for everyone.

## Known issues

1. e2e tests are failing, and I'll investigate and resolve tomorrow.

pkg/controller/elasticsearch/driver/shared/reconcile.go
naemono: Ok, I added a `MaybeGarbageCollectKeystorePasswordSecret` that is called near the end of the reconciliation, after the pod spec changes. Lmk if this is what you were thinking.

pkg/controller/elasticsearch/fips/fips_keystore_password.go
naemono: Ok, I've implemented this with an additional `Length` func on the interface, but I feel there may be another way you had in mind. Lmk if this doesn't work.

pkg/controller/elasticsearch/keystorepassword/keystore_password.go
naemono: Good point here. I've completely adjusted this.

pkg/controller/common/keystore/initcontainer.go
naemono: I went with option 1. :+1:

pkg/controller/elasticsearch/keystorepassword/keystore_password.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+	userOverride, err := esettings.AnyNodeSetHasUserProvidedKeystorePassword(ctx, c, es.Namespace, es.Spec.NodeSets)
+	if err != nil {
+		return err
+	}
+	if !fipsEnabled || userOverride {
+		return DeleteKeystorePasswordSecret(ctx, c, es)
+	}
+	return nil
+}
```
barkbay, line 185: The "should the operator manage a keystore password?" decision logic is duplicated between `reconcileManagedKeystorePasswordSecret` (in `reconcile.go`) and `MaybeGarbageCollectKeystorePasswordSecret` here. Both check `AnyNodeSetFIPSEnabled` and `AnyNodeSetHasUserProvidedKeystorePassword`, but with slightly different conditions (e.g. the version gate is only in the reconcile path).

If a new condition is added in one place but not the other, the create and delete paths could disagree — one creating the secret while the other deletes it. Consider extracting a shared `shouldManageKeystorePassword(...)` function that both paths call, ensuring the decision is always consistent.

pkg/controller/elasticsearch/validation/warnings.go
```diff
@@ -79,6 +88,127 @@ func validateClientAuthentication(config *common.CanonicalConfig, index int) fie
...
+	if err != nil {
+		return nil, err
+	}
+
+	var warnings field.ErrorList
+	if state.fipsEnabledCount > 0 && state.fipsEnabledCount < state.parseableNodeSetCount {
+		// Attach the warning to the nodeSets path since the inconsistency is per-NodeSet.
+		warnings = append(warnings, field.Invalid(field.NewPath("spec").Child("nodeSets"), es.Spec.NodeSets, inconsistentFIPSModeWarningMsg))
```
barkbay, line 111: The `NewCanonicalConfigFrom` error is silently skipped (`continue`), but the `MergeWith` error a few lines later returns a hard failure that propagates up and blocks admission. Since both are in a warning-only code path, should `MergeWith` follow the same pattern and skip the NodeSet rather than blocking admission?

pkg/controller/elasticsearch/validation/warnings.go
```diff
@@ -79,6 +88,127 @@ func validateClientAuthentication(config *common.CanonicalConfig, index int) fie
...
+	if verErr != nil {
+		// Surface the parse failure as a warning for callers that evaluate warnings in isolation;
+		// ValidateElasticsearch runs supportedVersion first, so this duplicates admission only if
+		// validations are skipped.
+		warnings = append(warnings, field.Invalid(field.NewPath("spec").Child("version"), es.Spec.Version, parseVersionErrMsg))
+		return warnings, nil //nolint:nilerr // version parse failures are admission warnings, not returned as validation errors
+	}
+
+	if state.fipsEnabledCount > 0 &&
+		!state.hasUserProvidedKeystorePassword &&
```
barkbay, line 124: This duplicates the policy config secret reading logic from `nodespec.GetPolicyConfig` with a hardcoded `"elasticsearch.json"` key constant. The comment mentions an import cycle with `nodespec`, but the actual cycle is `common/operator` → `validation` → `stackconfigpolicy` → `common/operator`.

Could this be extracted into `elasticsearch/settings` instead? `validation` already imports that package, and `settings` doesn't depend on `stackconfigpolicy` or `common/operator`. That would let both `nodespec` and `validation` share the same reading logic and avoid the duplicated constant.

pkg/controller/common/password/password.go
```diff
@@ -63,6 +64,20 @@ func (r *randomPasswordGenerator) Generate(ctx context.Context) ([]byte, error)
...
+func (r *randomPasswordGenerator) Length(ctx context.Context) int {
+	useLength, err := r.useLength(ctx)
+	if err != nil {
+		return 24
+	}
+	if useLength {
+		return r.length
+	}
+	// Keep Length aligned with Generate fallback behavior above when enterprise
+	// features are disabled which simply returns 24.
```
barkbay, line 77: `Generate` returns the error when `useLength` (license check) fails, but `Length` silently falls back to 24. If the license check is flaky, the caller gets a silent default with no signal that something went wrong — while `Generate` in the same reconcile loop would fail.

Should the signature be `Length(ctx context.Context) (int, error)` to match `Generate` and let the caller decide how to handle the failure?

pkg/controller/elasticsearch/keystorepassword/keystore_script.go
```diff
@@ -0,0 +1,58 @@
...
+// you may not use this file except in compliance with the Elastic License 2.0.
+
+package keystorepassword
+
+import "github.com/elastic/cloud-on-k8s/v3/pkg/controller/common/keystore"
+
+// #nosec G101 -- this is a shell template variable name, not a hardcoded credential.
+const elasticsearchPasswordProtectedKeystoreScript = `#!/usr/bin/env bash
+
+set -eux
```
barkbay: The existing non-FIPS scripts use `set -eux` safely because secret values flow through files, never shell variables — `set -x` only traces command arguments and file paths, not file contents.

This script is different: the password lives in a shell variable (`KEYSTORE_PASSWORD`) and is passed through `printf` and `echo -n`. `set -x` traces commands after variable expansion, so the password would appear in plain text in logs. The `set +x`/`set -x` toggle below mitigates this, but it's fragile — a future edit in the wrong place silently re-enables tracing over sensitive lines.

Since `set -x` adds no real diagnostic value here (the script is short, has explicit `echo` statements, and `set -e` already surfaces failures), could this just be `set -eu`? That removes the need for the `set +x`/`set -x` toggling below entirely.

pkg/controller/elasticsearch/keystorepassword/keystore_script.go
```diff
@@ -0,0 +1,58 @@
...
+KEYSTORE_PASSWORD=$(cat "{{ .KeystorePasswordPath }}")
+
+# create a password-protected keystore; printf supplies password twice (new + confirmation)
+printf "%s\n%s\n" "$KEYSTORE_PASSWORD" "$KEYSTORE_PASSWORD" | {{ .KeystoreCreateCommand }} -p
+
+# add all existing secret entries into it
+for filename in  {{ .SecureSettingsVolumeMountPath }}/*; do
+	[[ -e "$filename" ]] || continue # glob does not match
+	key=$(basename "$filename")
+	echo "Adding "$key" to the keystore."
```
barkbay: nit: shellcheck SC2027/SC2086 — the quotes around `$key` break the outer quoting, leaving `$key` effectively unquoted (vulnerable to globbing/word splitting):

```suggestion
	echo "Adding ${key} to the keystore."
```

(Same pattern exists in the non-FIPS default script in `initcontainer.go`.)

pkg/controller/common/keystore/initcontainer.go
```diff
@@ -34,6 +32,9 @@ type InitContainerParameters struct {
 	// SkipInitializedFlag when true do not use a flag to ensure the keystore is created only once. This should only be set
 	// to true if the keystore can be forcibly recreated.
 	SkipInitializedFlag bool
+	// KeystorePasswordPath can be referenced by a custom script when the keystore
+	// password is provided through a mounted file.
+	KeystorePasswordPath string
```
barkbay, line 37: What does "custom script" refer to here? The field is used by `ApplyPasswordProtectedKeystoreScript` which sets `CustomScript`, but the comment doesn't make that connection clear. A reader unfamiliar with the FIPS/keystore-password flow would not understand what "custom" means in this context.

pkg/controller/elasticsearch/validation/warnings.go
naemono: Yeah, I fought with this yesterday, hence the comment about it. I didn't see this path to resolve it. :+1:

pkg/controller/common/keystore/initcontainer.go
naemono: The TDD for this changed 3 times, so this comment was from a previously implementation, and wasn't updated when the new changes. It's been updated to be more accurate now.

pkg/controller/elasticsearch/keystorepassword/keystore_script.go
naemono: I don't think the `x` adds real value. I've removed it.

naemono:
buildkite test this -f p=gke,E2E_TAGS=es,t=TestFIPS -m s=9.4.0-SNAPSHOT

naemono:
buildkite test this -f p=gke,E2E_TAGS=es,t=TestFIPS -m s=9.4.0-SNAPSHOT

naemono:
buildkite test this -f p=gke,E2E_TAGS=es,t=TestFIPS -m s=9.4.0-SNAPSHOT

test/e2e/es/fips_keystore_test.go
```diff
@@ -0,0 +1,536 @@
...
+	}
+}
+
+func TestFIPSKeystoreManagedResources(t *testing.T) {
+	skipUnlessFIPSKeystoreStackVersion(t)
+	k := test.NewK8sClientOrFatal()
+	b := elasticsearch.NewBuilder("test-fips-ks-managed").
+		WithESMasterDataNodes(1, elasticsearch.DefaultResources).
+		WithEmptyDirVolumes().
+		WithAdditionalConfig(map[string]map[string]interface{}{
```
barkbay:
```suggestion
		WithAdditionalConfig(map[string]map[string]any{
```

test/e2e/es/fips_keystore_test.go
```diff
@@ -0,0 +1,536 @@
...
+		passphraseMountPath  = "/mnt/elastic-internal/e2e-user-fips-keystore-passphrase"
+		passphraseVolumeName = "e2e-user-fips-keystore-passphrase"
+	)
+	passphraseSecretName := fmt.Sprintf("test-fips-ks-user-override-pass-%s", rand.String(4))
+	userPassphraseFile := fmt.Sprintf("%s/%s", passphraseMountPath, passphraseSecretKey)
+
+	b := elasticsearch.NewBuilder("test-fips-ks-user-override").
+		WithESMasterDataNodes(1, elasticsearch.DefaultResources).
+		WithEmptyDirVolumes().
+		WithAdditionalConfig(map[string]map[string]interface{}{
```
barkbay:
```suggestion
		WithAdditionalConfig(map[string]map[string]any{
```

test/e2e/es/fips_keystore_test.go
```diff
@@ -0,0 +1,536 @@
...
+	steps.RunSequential(t)
+}
+
+func TestFIPSKeystoreSecretDeletedWhenFIPSDisabled(t *testing.T) {
+	skipUnlessFIPSKeystoreStackVersion(t)
+	k := test.NewK8sClientOrFatal()
+	b := elasticsearch.NewBuilder("test-fips-ks-off-cleans-secret").
+		WithESMasterDataNodes(1, elasticsearch.DefaultResources).
+		WithEmptyDirVolumes().
+		WithAdditionalConfig(map[string]map[string]interface{}{
```
barkbay:
```suggestion
		WithAdditionalConfig(map[string]map[string]any{
```

test/e2e/es/fips_keystore_test.go
```diff
@@ -0,0 +1,536 @@
...
+
+	k := test.NewK8sClientOrFatal()
+	selectorValue := rand.String(4)
+	b := elasticsearch.NewBuilder("test-fips-ks-scp").
+		WithESMasterDataNodes(1, elasticsearch.DefaultResources).
+		WithEmptyDirVolumes().
+		WithLabel("fips-ks-scp", selectorValue)
+	bWithLicense := test.LicenseTestBuilder(b)
+
+	config := commonv1.NewConfig(map[string]interface{}{
```
barkbay:
```suggestion
	config := commonv1.NewConfig(map[string]any{
```

test/e2e/es/fips_keystore_test.go
```diff
@@ -0,0 +1,536 @@
...
+// TestFIPSKeystoreWithSecureSettings validates that secure settings are added
+// through the password-protected keystore flow when FIPS mode is enabled.
+func TestFIPSKeystoreWithSecureSettings(t *testing.T) {
+	skipUnlessFIPSKeystoreStackVersion(t)
+
+	k := test.NewK8sClientOrFatal()
+	b := elasticsearch.NewBuilder("test-fips-ks-secure-settings").
+		WithESMasterDataNodes(1, elasticsearch.DefaultResources).
+		WithEmptyDirVolumes().
+		WithAdditionalConfig(map[string]map[string]interface{}{
```
barkbay:
```suggestion
		WithAdditionalConfig(map[string]map[string]any{
```

test/e2e/es/fips_keystore_test.go
```diff
@@ -0,0 +1,536 @@
...
+func checkFIPSPodTemplateUserOverride(pod *corev1.PodSpec, esContainer *corev1.Container) error {
+	hasFIPSVolume := slices.ContainsFunc(pod.Volumes, func(v corev1.Volume) bool {
+		return v.Name == esvolume.KeystorePasswordSecretVolumeName
+	})
+	mainHasFIPSMount := slices.ContainsFunc(esContainer.VolumeMounts, func(vm corev1.VolumeMount) bool {
+		return vm.Name == esvolume.KeystorePasswordSecretVolumeName && vm.MountPath == esvolume.KeystorePasswordSecretVolumeMountPath
+	})
+
+	initHasFIPSManagedArtifacts := false
+	if keystoreInit := containerByName(pod.InitContainers, commonkeystore.InitContainerName); keystoreInit != nil {
```
barkbay:
```suggestion
	if keystoreInit := pod.InitContainerByName(*podSpec, commonkeystore.InitContainerName); keystoreInit != nil {
```

with 

```golang
	"github.com/elastic/cloud-on-k8s/v3/pkg/controller/common/pod"
```

test/e2e/es/fips_keystore_test.go
```diff
@@ -0,0 +1,536 @@
...
+
+func checkFIPSPodTemplateOperatorManaged(pod *corev1.PodSpec, esContainer *corev1.Container) error {
+	hasFIPSVolume := slices.ContainsFunc(pod.Volumes, func(v corev1.Volume) bool {
+		return v.Name == esvolume.KeystorePasswordSecretVolumeName
+	})
+	mainHasFIPSMount := slices.ContainsFunc(esContainer.VolumeMounts, func(vm corev1.VolumeMount) bool {
+		return vm.Name == esvolume.KeystorePasswordSecretVolumeName && vm.MountPath == esvolume.KeystorePasswordSecretVolumeMountPath
+	})
+
+	keystoreInitContainer := containerByName(pod.InitContainers, commonkeystore.InitContainerName)
```
barkbay:
```suggestion
func checkFIPSPodTemplateOperatorManaged(podSpec *corev1.PodSpec, esContainer *corev1.Container) error {
	hasFIPSVolume := slices.ContainsFunc(podSpec.Volumes, func(v corev1.Volume) bool {
		return v.Name == esvolume.KeystorePasswordSecretVolumeName
	})
	mainHasFIPSMount := slices.ContainsFunc(esContainer.VolumeMounts, func(vm corev1.VolumeMount) bool {
		return vm.Name == esvolume.KeystorePasswordSecretVolumeName && vm.MountPath == esvolume.KeystorePasswordSecretVolumeMountPath
	})

	keystoreInitContainer := pod.InitContainerByName(*podSpec, commonkeystore.InitContainerName)
```

test/e2e/es/fips_keystore_test.go
```diff
@@ -0,0 +1,536 @@
...
+}
+
+func containerByName(containers []corev1.Container, name string) *corev1.Container {
+	for i := range containers {
+		if containers[i].Name == name {
+			return &containers[i]
+		}
+	}
+	return nil
+}
```
barkbay:
```suggestion
```

barkbay (commented):
I think we don't need `containerByName`

test/e2e/es/fips_keystore_test.go
```diff
@@ -0,0 +1,536 @@
...
+					return fmt.Errorf("expected at least one nodeset")
+				}
+				if es.Spec.NodeSets[0].Config == nil {
+					es.Spec.NodeSets[0].Config = b.Elasticsearch.Spec.NodeSets[0].Config.DeepCopy()
+				}
+				es.Spec.NodeSets[0].Config.Data[fipsSetting] = false
+				return k.Client.Update(context.Background(), &es)
+			}),
+		}).
+		WithStep(checkFIPSKeystoreSecretDeletedStep(k, b.Elasticsearch)).
```
barkbay, line 192: Should we check that the cluster is healthy after the FIPS -> non-FIPS transition?

barkbay (approved):
LGTM

naemono:
buildkite test this -f p=gke,E2E_TAGS=es,t=TestFIPS -m s=9.4.0-SNAPSHOT

naemono:
> LGTM

Thanks. One more e2e test and I'll get this merged. @pkoutsovasilis for visibility.

pkoutsovasilis:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|3.4|[](https://github.com/elastic/cloud-on-k8s/pull/9347)|

### Questions ?

pkg/controller/elasticsearch/driver/shared/reconcile.go
```diff
@@ -334,6 +338,16 @@ func ReconcileSharedResources(
 	keystoreSecurityContext := securitycontext.For(params.Version, true)
 	keystoreParams.SecurityContext = &keystoreSecurityContext
 
+	keystorePasswordSecret, err := reconcileManagedKeystorePasswordSecret(ctx, client, es, params.Version, params.OperatorParameters.PasswordGenerator, meta)
```
barkbay, line 341: I did not realize that this should not be part of the shared driver, the password is not going to be used in stateless.

PR #9307, opened 2026-04-03
Implement apm support for presenting client certificates to elasticsearch

pkoutsovasilis:
## Summary

  This PR implements client certificate support for APM Server when connecting to an Elasticsearch cluster that has client authentication enabled.

Relates to https://github.com/elastic/cloud-on-k8s/issues/9081

  ### Changes

  - **APM `ElasticsearchSelector`**: Changes `ElasticsearchRef` from `ObjectSelector` to `ElasticsearchSelector`, adding `clientCertificateSecretName` support to APM Server's `elasticsearchRef`
  - **APM Server config**: Configures `output.elasticsearch.ssl.certificate` and `output.elasticsearch.ssl.key` when client authentication is required, pointing to the  mounted client certificate files
  - **APM Server pod spec**: Mounts client certificate volumes alongside existing CA certificate volumes for associations that have client certificates configured. Refactored `withAssociationCACertsVolumes` into `withAssociationCertsVolumes` to handle both CA and client certificate volume mounts
  - **Webhook validation**: Updated to use `CheckElasticsearchSelectorRefs` for the APM Server's `elasticsearchRef`
  - **APM e2e tests**: Added `TestClientAuthRequiredTransition` (verifies APM Server remains healthy when ES transitions from client auth required to disabled, and that client certificate secrets are cleaned up) and `TestClientAuthRequiredCustomCertificate` (verifies APM Server works with a user-provided client certificate)

  ### API

  ```yaml
  apiVersion: apm.k8s.elastic.co/v1
  kind: ApmServer
  spec:
    elasticsearchRef:
      name: elasticsearch
      # Optional: use a custom client certificate
      clientCertificateSecretName: my-custom-client-cert
   ```

### DIFF
--- a/test/e2e/test/apmserver/builder.go
+++ b/test/e2e/test/apmserver/builder.go
@@ -101,10 +101,7 @@ func (b Builder) WithNodeCount(count int32) Builder {
 }
 
 func (b Builder) WithElasticsearchRef(ref commonv1.ObjectSelector) Builder {
-	b.ApmServer.Spec.ElasticsearchRef = commonv1.ElasticsearchSelector{
-		ObjectSelector:              ref,
-		ClientCertificateSecretName: b.ApmServer.Spec.ElasticsearchRef.ClientCertificateSecretName,
-	}
+	b.ApmServer.Spec.ElasticsearchRef.ObjectSelector = ref
 	return b
 }
 


prodsecmachine:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/6ec1f09b-1e40-40d3-8138-c108cc599a82?refs=8d57479252c52a6d7a7d183a0cc4d7060d8a4480&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/6ec1f09b-1e40-40d3-8138-c108cc599a82/license?refs=8d57479252c52a6d7a7d183a0cc4d7060d8a4480&source=prChecksComment) |

pkoutsovasilis:
buildkite test this -f p=gke,t=TestClientAuthRequired.*,E2E_TAGS=apm -m s=9.3.2,s=8.19.13,s=9.2.7

pkoutsovasilis:
buildkite test this -f p=gke,t=TestClientAuthRequired.*,E2E_TAGS=apm -m s=9.3.3,s=8.19.14,s=9.2.8

test/e2e/test/apmserver/builder.go
```diff
@@ -101,7 +101,15 @@ func (b Builder) WithNodeCount(count int32) Builder {
...
 func (b Builder) WithElasticsearchRef(ref commonv1.ObjectSelector) Builder {
-	b.ApmServer.Spec.ElasticsearchRef = ref
+	b.ApmServer.Spec.ElasticsearchRef = commonv1.ElasticsearchSelector{
+		ObjectSelector:              ref,
+		ClientCertificateSecretName: b.ApmServer.Spec.ElasticsearchRef.ClientCertificateSecretName,
+	}
+	return b
+}
+
+func (b Builder) WithClientCertificateSecret(secretName string) Builder {
```
simitt, line 108: This implicitly requires that `WithElasticsearchRef` is also called, but it's not enforced anywhere afaics. I'd add at least a comment here highlighting the dependency.

simitt (approved):
Overall, LGMT, just some nits - but feel free to ignore. 

nit: I wonder whether a test scenario where the ES cluster uses a publicly trusted cert but mTLS is configured is worth testing in the unit tests.

test/e2e/test/apmserver/builder.go
pkoutsovasilis: The ordering doesn't actually matter here - `WithElasticsearchRef` explicitly preserves any previously set ClientCertificateSecretName, so calling `WithClientCertificateSecret` before or after `WithElasticsearchRef` both work correctly. Now if a developer that writes an e2e test and doesn't set an ElasticsearchRef but calls `WithClientCertificateSecret` the result will be an invalid spec that would be caught by [CRD validation](https://github.com/tehbooom/cloud-on-k8s/blob/20f5e8be212737f9966640c174c912b62ac095ce/pkg/apis/common/v1/common.go#L316-L319) which is correct IMO. Am I missing something?

pkoutsovasilis:
> nit: I wonder whether a test scenario where the ES cluster uses a publicly trusted cert but mTLS is configured is worth testing in the unit tests.

A couple of thoughts on this:
1. I'm not sure how we'd fabricate a publicly trusted cert in our e2e CI environment.
2. A publicly trusted cert affects the client-side validation of the server (ES in our case) certificate - it would be trusted without needing a defining a custom trusted CA. But the server-side validation of client certificates (the mTLS part) remains the same regardless of how the server's own cert is issued. The client still needs to present a valid certificate that the server trusts.
3. I'd also expect mTLS with publicly trusted server certificates to be fairly uncommon in practice - In my head, although I can be off here, mTLS is typically used in internal/private zero-trust environments where self-signed or internal CAs are the norm.

Is there a particular edge case you have in mind @simitt where the combination would behave differently? Happy to discuss and fabricate a test if there's a concrete scenario worth covering.

test/e2e/test/apmserver/builder.go
simitt: >Now if a developer that writes an e2e test and doesn't set an ElasticsearchRef but calls WithClientCertificateSecret the result will be an invalid spec that would be caught by [CRD validation](https://github.com/tehbooom/cloud-on-k8s/blob/20f5e8be212737f9966640c174c912b62ac095ce/pkg/apis/common/v1/common.go#L316-L319) which is correct IMO. Am I missing something?

True; my concern was about discoverability when used without the `WithElasticsearchRef` - but it seems that this is covered.

simitt:
@pkoutsovasilis , looking at this [`pod_test`](https://github.com/elastic/cloud-on-k8s/pull/9307/changes#diff-e9507a63998a57580b14cc974d5c1cfcb6eb4677023f1f220a5118b0896a8c93R222-R231), my concern actually is already covered - client cert only, no CA. So you can safely drop my previous comment. 

>Is there a particular edge case you have in mind

No particular use case, just something I thought missing in test coverage.

PR #9316, opened 2026-04-06
Fix StackConfigPolicy controller performing unnecessary file-settings secret updates on every reconciliation

### LINKED ISSUE #9175: StackConfigPolicy controller may perform unnecessary secret updates on every reconciliation
## Description

The StackConfigPolicy controller may update secrets it manages (file-settings secret, ES config secret, and copied secret mounts) on every reconciliation even when nothing has actually changed. Because these secrets are owned by the Elasticsearch resource, each update triggers a watch event that causes the Elasticsearch controller to re-reconcile, which may in turn trigger further StackConfigPolicy reconciliations, creating a reconciliation amplification loop.

## Expected Behavior

When neither the StackConfigPolicy nor the Elasticsearch resource has changed, the StackConfigPolicy reconciliation should be a no-op: it should read the existing secrets, determine they are up-to-date, and skip the update.

## Actual Behavior

The `NeedsUpdate` check in `filesettings.ReconcileSecret` (`pkg/controller/elasticsearch/filesettings/reconciler.go`) and `reconciler.ReconcileSecret` (`pkg/controller/common/reconciler/secret.go`) may return `true` even when the logical content of the secret has not changed, causing an unnecessary API `Update` call.

Potential causes for spurious `NeedsUpdate == true`:

1. **Managed annotation drift**: The `NeedsUpdate` function checks `managedAnnotations` (which includes `SecureSettingsSecretsAnnotationName`, `SettingsHashAnnotationName`, `ElasticsearchConfigAndSecretMountsHashAnnotation`, `KibanaConfigHashAnnotation`, `SoftOwnerRefsAnnotation`) using `maps.IsEqualSubset`. If any of these annotations are absent from the expected secret but present on the reconciled secret (or vice versa), an update is triggered. The same set of `managedAnnotations` is used for both the file-settings secret and the ES config secret, but not all annotations are relevant to both — this could cause flip-flopping where one reconciliation adds an annotation and the next removes it.

2. **Non-deterministic JSON serialization of secure settings annotation**: `setSecureSettings` in `pkg/controller/elasticsearch/filesettings/secret.go` marshals `[]commonv1.NamespacedSecretSource` into JSON for the `SecureSettingsSecretsAnnotationName` annotation. If the slice order of secret sources varies between reconciliations (e.g. due to policy merge order), the JSON string would differ even though the logical content is the same.

3. **Metadata accumulation in `reconcileSecretMounts`**: In `pkg/controller/stackconfigpolicy/elasticsearch_config_settings.go`, the `meta` variable in `reconcileSecretMounts` is reassigned inside the loop via `meta = meta.Merge(...)`, causing metadata to accumulate across iterations. While `SourceSecretAnnotationName` is overwritten each iteration (same key), other accumulated metadata could cause annotation differences on subsequent reconciliations.

## Impact

- **Reconciliation amplification**: Each unnecessary secret update triggers an Elasticsearch controller reconciliation (via `OwnerWatch`), which may trigger further StackConfigPolicy reconciliations.
- **Increased API server load**: Unnecessary `Update` calls increase etcd writes and API server load.
- **Memory pressure**: The reconciliation storm consumes operator memory, contributing to OOM risk especially for operators managing many namespaces or large clusters.

## Relevant Code

- `pkg/controller/elasticsearch/filesettings/reconciler.go` — `ReconcileSecret` and `NeedsUpdate` logic (lines 79-86)
- `pkg/controller/elasticsearch/filesettings/secret.go` — `setSecureSettings` and `NewSettingsSecretWithVersion`
- `pkg/controller/common/reconciler/secret.go` — `ReconcileSecret` (general) and `SetMultipleSoftOwners`
- `pkg/controller/stackconfigpolicy/controller.go` — main reconciliation loop calling `ReconcileSecret` for each targeted ES cluster
- `pkg/controller/stackconfigpolicy/elasticsearch_config_settings.go` — `reconcileSecretMounts` with `meta` accumulation

## Suggested Fix

1. Add logging (at debug/trace level) to `ReconcileResource` when `NeedsUpdate` returns `true`, including which specific field (labels, annotations, or data) triggered the update. This would help diagnose exactly when and why unnecessary writes occur.
2. Ensure `managedAnnotations` lists are scoped per secret type rather than shared, so annotations irrelevant to a specific secret don't cause spurious updates.
3. Consider normalizing slice-based annotations (like `SecureSettingsSecretsAnnotationName`) by sorting before JSON serialization.
4. Fix the `meta` accumulation in `reconcileSecretMounts` by using a fresh copy of `meta` for each iteration.

pkoutsovasilis:
## Summary

  Fixes #9175

  The StackConfigPolicy controller performs an unnecessary `Update` API call on the `<cluster>-es-policy-config` secret on every reconciliation cycle when the policy uses  only file-settings fields (e.g. `clusterSettings`, `snapshotRepositories`, `securityRoleMappings`) without `elasticsearch.config` or `elasticsearch.secretMounts`.

  ### Root cause

  `newElasticsearchConfigSecret` initializes `data := make(map[string][]byte)`. When neither `Config` nor `SecretMounts` are set, `data` stays as an empty map. The Kubernetes API server normalizes empty maps to `nil` (the `Data` field uses `omitempty`), so on the next reconciliation `reflect.DeepEqual(map[string][]byte{}, nil)` returns `false`, causing `NeedsUpdate` to return `true` every cycle.

  ### Changes

  - **Fix nil-vs-empty-map mismatch**: Use `var data map[string][]byte` (nil) instead of `make(map[string][]byte)` in `newElasticsearchConfigSecret`, only allocating when there are entries to add.
  - **Scope managed annotations per secret type**: Replace the shared `managedAnnotations` list with per-type lists (`fileSettingsManagedAnnotations`, `esConfigManagedAnnotations`, `kibanaConfigManagedAnnotations`) and expose typed functions (`ReconcileFileSettingsSecret`, `ReconcileESConfigSecret`, `ReconcileKibanaConfigSecret`) so callers cannot pass the wrong annotation scope. The underlying `reconcileSecret` is now unexported.
  - **Add debug logging**: Log which field (labels, annotations, or data) triggered `NeedsUpdate` at `V(1)` level to aid future diagnosis.
  - **Avoid meta variable reuse in `reconcileSecretMounts`**: Use a fresh `secretMeta` per loop iteration instead of re-assigning `meta`.


prodsecmachine:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/f4e21df1-8be8-4949-a296-dffc46934ec0?refs=ddd4cd4ebb78396677b140264223e42f5fba0c7a&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/f4e21df1-8be8-4949-a296-dffc46934ec0/license?refs=ddd4cd4ebb78396677b140264223e42f5fba0c7a&source=prChecksComment) |

pkg/controller/stackconfigpolicy/elasticsearch_config_settings.go
moukoublen, line 73: [nit / personal preference]: I am wondering if keeping the initial approach and having something like this:

```golang
if len(data) == 0 { data = nil }
```

before setting `Secret.Data`, would be better, than to conditionally initializing the `data` each time we need to set a field.

pkoutsovasilis: In general I prefer up-front approaches but I feel this proposal improves the ergonomics of this func so I changed it in 3a0e939b5c6537934915e205cb16779fee90d96a

Issue #9295, opened 2026-04-02
Deprecate managed=false and remove legacy pause annotation


pebrc:
Part of #9250. Depends on #9292 (ES implementation).

### 1. Remove `common.k8s.elastic.co/pause` (deprecated since 2020)

Remove the legacy pause annotation support from `IsUnmanaged()`. This is a breaking change and must be called out in release notes.

### 2. Deprecate `eck.k8s.elastic.co/managed=false`

- Emit a deprecation warning (log line + Kubernetes Event) when `managed=false` is detected, pointing users to `pause-orchestration=true`
- Documentation stops recommending `managed=false` for maintenance scenarios
- `managed=false` continues to work unchanged for backwards compatibility during the deprecation period (at least 3 release cycles)

### 3. Handle cross-cutting controllers

The following controllers use `IsUnmanaged` but don't manage workload pods directly. They need consideration for what `pause-orchestration` means in their context:
- Association reconciler (`pkg/controller/association/reconciler.go`)
- Remote Cluster controller (`pkg/controller/remotecluster/controller.go`)
- Stack Config Policy controller (`pkg/controller/stackconfigpolicy/controller.go`)
- Autoscaling controller (`pkg/controller/autoscaling/elasticsearch/controller.go`)

### 4. Update documentation

- Update the [managed annotation docs](https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-upgrading-eck.html#k8s-beta-to-ga-rolling-restart) to recommend `pause-orchestration`
- Add documentation for the new `pause-orchestration` annotation

## Code pointers
- `pkg/controller/common/unmanaged.go`

PR #9330, opened 2026-04-08
[#9292] Elasticsearch changes for new `pause-orchestration` annotation

### LINKED ISSUE #9292: Pause orchestration: Elasticsearch implementation
Part of #9250.

### Annotation

```
eck.k8s.elastic.co/pause-orchestration: "true"
```

Separate from `eck.k8s.elastic.co/managed=false` which bypasses everything. Users can keep using `managed=false` for at least 3 release cycles; but we will mark it as deprecated.

Scope: per-cluster initially. Per-nodeSet is deferred — it would require annotating generated StatefulSets or a new spec field and adds significant complexity.

---

### Cut Point in the Reconciliation Loop

The reconciliation loop already has a clean boundary between housekeeping and spec-driven orchestration:

```
shared.ReconcileSharedResources()   ← housekeeping: certificates, users, unicast_hosts, health
        ↓
reconcileNodeSpecs()                ← spec-driven: StatefulSet updates, upscale, downscale, rolling upgrades
```

When `pause-orchestration` is set, skip `reconcileNodeSpecs()` entirely and return early after shared resources have been reconciled. This avoids threading a flag through every sub-function and is the minimal safe change.

---

### Changes Required

**1. Annotation constant** — `pkg/controller/common/unmanaged.go` (or a new file)

```go
// PauseOrchestrationAnnotation pauses spec-driven orchestration (rolling upgrades,
// StatefulSet spec changes, scale up/down) while keeping housekeeping running
// (certificate rotation, unicast hosts, user/secret reconciliation, health monitoring).
PauseOrchestrationAnnotation = "eck.k8s.elastic.co/pause-orchestration"

func IsOrchestrationPaused(object metav1.Object) bool {
    v, exists := object.GetAnnotations()[PauseOrchestrationAnnotation]
    return exists && v == "true"
}
```

**2. Early return in `driver/stateful/driver.go`**

After `ReconcileSharedResources` returns, check the annotation. Before returning, PDB reconciliation must still run (see below):

```go
if common.IsOrchestrationPaused(&d.ES) {
    actualSets, err := es_sset.RetrieveActualStatefulSets(d.Client, k8s.ExtractNamespacedName(&d.ES))
    if err != nil {
        return results.WithError(err)
    }
    if err := pdb.Reconcile(ctx, d.Client, d.ES, d.OperatorParameters.OperatorNamespace, actualSets, meta); err != nil {
        return results.WithError(err)
    }
    // emit condition + event (see §4)
    d.ReconcileState.UpdateWithPhase(esv1.ElasticsearchApplyingChangesPhase)
    return results
}
```

Note: `reconcileSuspendedPods` is called before this check and is **not** skipped — it acts on a separate user-facing annotation for debugging and is orthogonal to spec-driven orchestration.

**2a. Why PDB reconciliation must continue**

PDB dynamically sets `MinAvailable`/`MaxUnavailable` based on `es.Status.Health`, which stays live because the health observer keeps running. Freezing the PDB creates two failure modes:

- **Too permissive:** cluster green when paused → PDB allows 1 disruption → health drops to yellow after first drain → PDB should drop to 0 but doesn't → second drain risks data loss.
- **Too restrictive:** cluster yellow/red when paused → PDB stuck at 0 disruptions → `kubectl drain` stalls indefinitely, defeating the purpose of the annotation.

Since no upscale/downscale runs during pause, `actualStatefulSets` is stable and the PDB call is cheap and idempotent.

**3. Detect pending spec changes**

Used to populate the condition message. Reuse existing machinery:

```go
func (d *Driver) hasPendingSpecChanges(ctx context.Context, meta metadata.Metadata) (bool, error) {
    actualSets, err := es_sset.RetrieveActualStatefulSets(d.Client, k8s.ExtractNamespacedName(&d.ES))
    if err != nil {
        return false, err
    }
    expectedResources, err := nodespec.BuildExpectedResources(ctx, d.Client, d.ES, nil, actualSets,
        d.OperatorParameters.IPFamily, d.OperatorParameters.SetDefaultSecurityContext, meta)
    if err != nil {
        return false, err
    }
    return sset.StatefulSetsDiffer(actualSets, expectedResources.StatefulSets()), nil
}
```

**4. Condition and events** — `pkg/apis/elasticsearch/v1/status.go` + `reconcile/state.go`

New condition type:

```go
OrchestrationPaused v1alpha1.ConditionType = "OrchestrationPaused"
```

Set when paused:

```go
// No pending changes
d.ReconcileState.ReportCondition(esv1.OrchestrationPaused, corev1.ConditionTrue,
    "Orchestration paused via annotation; no pending spec changes detected")

// Pending changes exist
d.ReconcileState.ReportCondition(esv1.OrchestrationPaused, corev1.ConditionTrue,
    "Orchestration paused via annotation; spec changes are pending and will be applied on resume")
d.ReconcileState.AddEvent(corev1.EventTypeWarning, events.EventReasonUnhealthy,
    "Spec changes pending but orchestration is paused — remove annotation to apply")
```

Clear when annotation is absent:

```go
d.ReconcileState.ReportCondition(esv1.OrchestrationPaused, corev1.ConditionFalse, "")
```

A condition is preferred over a plain event because events are ephemeral — the condition is visible via `kubectl get elasticsearch` and persists until the state changes.

---

### What Continues Running

| Operation | Continues? | Notes |
|---|---|---|
| Certificate rotation (HTTP + transport) | ✅ | In `ReconcileSharedResources` |
| `unicast_hosts.txt` ConfigMap | ✅ | In `ReconcileSharedResources`; tracks pod IP churn from k8s-level drains |
| User and role reconciliation | ✅ | In `ReconcileSharedResources` |
| Health observation and status updates | ✅ | In `ReconcileSharedResources` |
| Service reconciliation | ✅ | In `ReconcileSharedResources` |
| Keystore reconciliation | ✅ | In `ReconcileSharedResources` |
| PDB reconciliation | ✅ | Runs before early return, using current actual StatefulSets |
| Pod suspension (debug) | ✅ | `reconcileSuspendedPods` runs before the pause check |
| StatefulSet spec updates | ⏸ | Skipped |
| Rolling upgrades (pod deletions) | ⏸ | Skipped |
| Upscale (new pods) | ⏸ | Skipped |
| Downscale (node removal) | ⏸ | Skipped |
| Force upgrade of bootlooping pods | ⏸ | Skipped |
| Desired nodes API | ⏸ | Skipped |

---

### Sequence Diagram

```
operator reconcile loop
        │
        ├─► reconcileSuspendedPods()                ← always runs (debug feature)
        │
        ├─► shared.ReconcileSharedResources()        ← always runs
        │       certificates, unicast_hosts,
        │       users, health observer
        │
        ├─► IsOrchestrationPaused?
        │       YES ─► pdb.Reconcile()               ← runs with current actual StatefulSets
        │               hasPendingSpecChanges?
        │                   ├─ NO  ─► ReportCondition(OrchestrationPaused, True, "no pending changes")
        │                   └─ YES ─► ReportCondition(OrchestrationPaused, True, "pending changes")
        │                             AddEvent(Warning, "spec changes pending")
        │               return
        │
        └─► reconcileNodeSpecs()                     ← skipped when paused
                upscale, downscale, rolling upgrades
```

---

### Resume Behaviour

When the annotation is removed, the annotation change triggers a reconcile. ECK computes the diff between desired and actual state and applies it via the normal upgrade/upscale/downscale path. The `OrchestrationPaused` condition is set to `False`. No special resume logic needed.

---

### Alternatives Considered

**A: Thread a flag through sub-functions** — check `IsOrchestrationPaused` inside `HandleUpscaleAndSpecChanges`, `HandleDownscale`, `handleUpgrades` individually. No benefit over cutting at the `reconcileNodeSpecs` boundary since we want to pause all topology changes. More code, more places for regressions.

**B: Reuse `managed=false` with a narrower bypass** — modify `IsUnmanaged` to accept phases to skip. Changes semantics of an existing annotation in a backwards-incompatible way.

**C: New field in the `Elasticsearch` spec** — `spec.orchestration.paused: true`. More formal, but requires a CRD version bump. An annotation is lower friction for an operational escape hatch and consistent with the existing `managed=false` pattern.

---

### Open Questions

1. **`GarbageCollectPVCs`**: inside `reconcileNodeSpecs`, runs after downscale. Since downscale is paused, no new PVCs become garbage, so skipping it is safe.
2. **Interaction with autoscaler**: if an autoscaler is active, the spec may keep changing while paused. The warning event covers this; no additional code change needed.

### Unit tests
- `IsOrchestrationPaused` with annotation present/absent/wrong value
- `hasPendingSpecChanges` with matching and differing StatefulSets
- `Driver.Reconcile` with annotation: verify `reconcileNodeSpecs` is not called, condition is set, shared resources ARE called

### Integration/e2e tests
- Set annotation on a running cluster; issue a spec change (e.g. bump node count); verify pods are not modified, condition shows `OrchestrationPaused=True` with warning event
- Remove annotation; verify rolling upgrade proceeds and condition clears

## Code pointers
- `pkg/controller/common/unmanaged.go`
- `pkg/controller/elasticsearch/driver/stateful/driver.go`
- `pkg/controller/elasticsearch/hints/hints.go`
- `pkg/apis/elasticsearch/v1/status.go`

dtuck9:
## What this PR does                                                                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                                            
Introduces a new `eck.k8s.elastic.co/pause-orchestration` annotation that pauses spec-driven orchestration (rolling upgrades, StatefulSet spec changes, scale up/down) while keeping housekeeping running (certificate rotation, unicast hosts, user/secret reconciliation, health monitoring).                                                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                            
This is distinct from the existing `eck.k8s.elastic.co/managed=false` annotation, which skips reconciliation entirely and is marked as `Deprecated` with this change. The new annotation allows operators to freeze cluster topology during maintenance windows without losing certificate rotation or  health monitoring.

  ## Changes                                                                                                                                                                                                                                                                                                                
   
  - **`pkg/controller/common/unmanaged.go`** — New `PauseOrchestrationAnnotation`                                                                                                                                                                                                                                           
    constant and `IsOrchestrationPaused` helper. `ManagedAnnotation` and `IsUnmanaged`
    are marked deprecated in favour of the new annotation (tracked in #9295).
                                                                                                                                                                                                                                                                                                                            
  - **`pkg/controller/elasticsearch/driver/stateful/driver.go`** — When
    `pause-orchestration: "true"` is set, `Reconcile` skips `reconcileNodeSpecs` and                                                                                                                                                                                                                                        
    calls `reconcileCriticalSteps` instead. `reconcileCriticalSteps` reconciles the PDB,                                                                                                                                                                                                                                    
    detects pending spec changes via `hasPendingSpecChanges`, and sets the
    `OrchestrationPaused` condition + phase accordingly. A warning event is emitted if                                                                                                                                                                                                                                      
    spec changes are pending while paused.                                                                                                                                                                                                                                                                                  
   
  - **`pkg/apis/elasticsearch/v1/status.go`** — New `ElasticsearchOrchestrationPaused`                                                                                                                                                                                                                                      
    phase (`"Paused"`) and `OrchestrationPaused` condition type.
                                                                                                                                                                                                                                                                                                                            
  - **`pkg/controller/common/events/events.go`** — New `EventReasonPaused` and
    `EventActionPendingOrchestrationChanges` constants.                                                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                                                            
  ## Testing
                                                                                                                                                                                                                                                                                                                            
  - Unit tests for `IsOrchestrationPaused`, `hasPendingSpecChanges`, `hasSpecDiff`, and
    `reconcileCriticalSteps` (condition wiring, event emission, phase, error propagation).
  - E2e test `TestPauseOrchestration` cycles through: create → enable annotation → update
    spec while paused (verifies scale-out is held back) → disable annotation (verifies                                                                                                                                                                                                                                      
    scale-out applies).
  - Full run of e2e tests against GKE
                                                                                                                                                                                                                                                                                                                            
  ## Relates to   

  Closes #9292

## AI Usage Disclaimer
Claude used for unit test case generation.


prodsecmachine:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/39159e32-552c-49b4-82de-d3f0d5d97453?refs=30c5cea0bdba18d7a851ea1ffac53462c00c60e6&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/39159e32-552c-49b4-82de-d3f0d5d97453/license?refs=30c5cea0bdba18d7a851ea1ffac53462c00c60e6&source=prChecksComment) |

dtuck9:
buildkite test this -f p=gke,t=TestPauseOrchestration*

dtuck9:
buildkite test this -f p=gke,t=TestPauseOrchestration*

test/e2e/es/pause_orchestration_test.go
```diff
@@ -0,0 +1,156 @@
...
+		WithSteps(elasticsearch.AnnotatePodsWithBuilderHash(initialBuilder, k)).
+		WithSteps(enabledBuilder.MutationTestSteps(k)).
+		WithSteps(verifyPauseOrchestrationResourcesExist(k, esNamespace, actualESName, 1)).
+		// Phase 2: update Elasticsearch spec
+		WithSteps(elasticsearch.AnnotatePodsWithBuilderHash(updatedBuilder, k)).
+		WithSteps(updatedBuilder.UpgradeTestSteps(k)).
+		WithSteps(verifyPauseOrchestrationResourcesExist(k, esNamespace, actualESName, 1)).
+		// Phase 3: disable pause-orchestration
+		WithSteps(elasticsearch.AnnotatePodsWithBuilderHash(*disabledBuilder, k)).
+		WithSteps(disabledBuilder.UpgradeTestSteps(k)).
```
dtuck9, line 79: Using `MutationTestSteps` here exposes a potential issue where the pending change fails as a result of the minimum expected nodes not yet running (i.e., the plan fails because the plan hasn't even been applied). E.g.

```
Error:          change budget violated
                Test:           TestPauseOrchestration/Stopping_to_pod_count_for_change_budget:_expect_to_stay_within_the_change_budget#01
                Messages:       ready pod count 1 when allowed min was 2
```

I will circle back to that.

test/e2e/test/utils.go
```diff
@@ -78,7 +78,12 @@ func ExitOnErr(err error) {
...
 // Eventually runs the given function until success with a default timeout.
 func Eventually(f func() error) func(*testing.T) {
-	return UntilSuccess(f, ctx.TestTimeout)
+	return EventuallyWithTimeout(f, ctx.TestTimeout)
+}
+
+// EventuallyWithTimeout runs the given function until success with the given timeout.
+func EventuallyWithTimeout(f func() error, timeout time.Duration) func(*testing.T) {
+	return UntilSuccess(f, timeout)
 }
```
dtuck9, line 82: Drive-by change: the `ctx.TestTimeout` value is used in `Eventually`, and set by the env var `TEST_TIMEOUT`, which is also used for the total test timeout value here: https://github.com/elastic/cloud-on-k8s/blob/9d41a77837a26026603313d81f94aa29d48b25d8/pkg/utils/retry/retry.go#L28-L61

If this value is surpassed, the context is canceled which kills the test and results in the following error:

```
{"log.level":"info","@timestamp":"2026-04-14T11:08:34.419-0700","log.logger":"e2e","message":"Test cancelled","service.version":"0.0.0-SNAPSHOT+00000000","service.type":"eck","ecs.version":"1.4.0","kill_error":null}
{"log.level":"info","@timestamp":"2026-04-14T11:08:34.420-0700","log.logger":"e2e","message":"Deleting resources","service.version":"0.0.0-SNAPSHOT+00000000","service.type":"eck","ecs.version":"1.4.0","file":"/tmp/eck-e2e/e2e/roles.yaml1768399112"}
{"log.level":"info","@timestamp":"2026-04-14T11:08:35.997-0700","log.logger":"e2e","message":"Deleting resources","service.version":"0.0.0-SNAPSHOT+00000000","service.type":"eck","ecs.version":"1.4.0","file":"/tmp/eck-e2e/e2e/rbac.yaml2711912485"}
{"log.level":"info","@timestamp":"2026-04-14T11:08:49.810-0700","log.logger":"e2e","message":"Cleaning up the scratch directory","service.version":"0.0.0-SNAPSHOT+00000000","service.type":"eck","ecs.version":"1.4.0","directory":"/tmp/eck-e2e/e2e"}
{"log.level":"error","@timestamp":"2026-04-14T11:08:49.812-0700","log.logger":"e2e","message":"Failed to run e2e tests","service.version":"0.0.0-SNAPSHOT+00000000","service.type":"eck","ecs.version":"1.4.0","error":"signal: killed","error.stack_trace":"github.com/elastic/cloud-on-k8s/v3/test/e2e/cmd/run.Command.func2\n\t/Users/diana/go/src/github.com/elastic/cloud-on-k8s/test/e2e/cmd/run/command.go:80\ngithub.com/spf13/cobra.(*Command).execute\n\t/Users/diana/go/pkg/mod/github.com/spf13/cobra@v1.10.2/command.go:1015\ngithub.com/spf13/cobra.(*Command).ExecuteC\n\t/Users/diana/go/pkg/mod/github.com/spf13/cobra@v1.10.2/command.go:1148\ngithub.com/spf13/cobra.(*Command).Execute\n\t/Users/diana/go/pkg/mod/github.com/spf13/cobra@v1.10.2/command.go:1071\nmain.main\n\t/Users/diana/go/src/github.com/elastic/cloud-on-k8s/test/e2e/cmd/main.go:32\nruntime.main\n\t/opt/homebrew/opt/go/libexec/src/runtime/proc.go:290"}
Error: signal: killed
exit status 1
```

As a result of this: https://github.com/elastic/cloud-on-k8s/blob/9d41a77837a26026603313d81f94aa29d48b25d8/test/e2e/cmd/run/run.go#L514-L531

**In other words, the original error that led to the timeout is swallowed.** 

I initially added this for debugging purposes, but it does highlight a larger issue in the e2e tests that makes debugging more difficult. Addressing this throughout the e2e tests is out of scope for this PR, and I will look into either getting the original error to bubble up correctly or replacing the usages of the pre-existing `Eventually` function with the `EventuallyWithTimeout` approach depending on further investigation.

dtuck9:
buildkite test this -f p=gke,t=TestPauseOrchestration*

dtuck9:
buildkite test this -f p=gke

pkg/controller/elasticsearch/driver/stateful/driver.go
```diff
@@ -69,12 +77,81 @@ func (d *Driver) Reconcile(ctx context.Context) *reconciler.Results {
...
 }
 
+func (d *Driver) reconcileCriticalSteps(ctx context.Context, state *shared.ReconcileState, resolvedConfig nodespec.ResolvedConfig) error {
+	actualSets, err := es_sset.RetrieveActualStatefulSets(d.Client, k8s.ExtractNamespacedName(&d.ES))
+	if err != nil {
+		return err
+	}
+	if err = pdb.Reconcile(ctx, d.Client, d.ES, d.OperatorParameters.OperatorNamespace, actualSets, state.Meta); err != nil {
+		return err
+	}
```
pebrc, line 103: **Bug: Phase never transitions to `Paused`**

`UpdateWithPhase(ElasticsearchOrchestrationPaused)` is called here, but the ES controller's post-reconcile logic in `elasticsearch_controller.go:209-214` always overwrites it:

```go
if isReconciled, message := results.IsReconciled(); !isReconciled {
    state.UpdateWithPhase(esv1.ElasticsearchApplyingChangesPhase)
    ...
} else {
    state.UpdateWithPhase(esv1.ElasticsearchReadyPhase)
}
```

Since `reconcileCriticalSteps` returns no error and no requeue, `IsReconciled()` returns true, and the phase is clobbered to `Ready`. Confirmed in manual testing: the phase stays `Ready` at all times while paused.

The fix likely needs to happen in the controller — either by checking for the paused state before overwriting the phase, or by having `reconcileCriticalSteps` signal through the results that it shouldn't be overwritten.

pkg/controller/elasticsearch/driver/stateful/driver.go
```diff
@@ -69,12 +77,81 @@ func (d *Driver) Reconcile(ctx context.Context) *reconciler.Results {
...
+
+	message := "Orchestration paused via annotation; no pending spec changes detected"
+	if hasPendingChanges {
+		message = "Orchestration paused via annotation; spec changes are pending and will be applied on resume"
+		d.ReconcileState.AddEvent(corev1.EventTypeWarning, events.EventReasonPaused,
+			events.EventActionPendingOrchestrationChanges, "Spec changes pending but orchestration is paused — remove annotation to apply")
+	}
+	d.ReconcileState.ReportCondition(esv1.OrchestrationPaused, corev1.ConditionTrue, message)
+	return nil
```
pebrc, line 134: **Bug: Pod deletion while paused leaves pods permanently not-ready**

Tested on both 8.17.0 and 9.3.3: deleting a pod while orchestration is paused leaves it stuck at `0/1 Running` indefinitely.

**Root cause:** The pre-stop hook registers a `RESTART` node shutdown with ES. ES's `ReadinessService` checks whether a shutdown metadata entry *exists* for the local node — it does not check the shutdown status. The readiness port (8080) stays closed until the shutdown entry is explicitly `DELETE`d via `_nodes/<id>/shutdown`. This is by design in ES (confirmed by `NodeShutdownReadinessIT` integration test and the `ReadinessService.checkReadyState()` implementation).

Normally the operator clears completed shutdowns during `reconcileNodeSpecs`. With orchestration paused, that path is skipped, so the shutdown is never cleared and the pod stays not-ready.

**Impact:** Pod evictions, node drains, or OOM kills during a maintenance window (the primary use case for this annotation) would leave the cluster degraded with not-ready pods.

**Fix:** `reconcileCriticalSteps` should clear completed RESTART shutdowns, similar to what `reconcileNodeSpecs` does.

pkg/controller/elasticsearch/driver/stateful/driver.go
```diff
@@ -69,12 +77,81 @@ func (d *Driver) Reconcile(ctx context.Context) *reconciler.Results {
...
+	actualSets es_sset.StatefulSetList,
+	state *shared.ReconcileState,
+	resolvedConfig nodespec.ResolvedConfig,
+) (bool, error) {
+	expectedResources, err := nodespec.BuildExpectedResources(ctx, d.Client, d.ES, state.KeystoreResources, actualSets,
+		d.OperatorParameters.SetDefaultSecurityContext, state.Meta, resolvedConfig)
+	if err != nil {
+		return false, err
+	}
+	return hasSpecDiff(actualSets, expectedResources.StatefulSets()), nil
```
pebrc: **`hasPendingSpecChanges` false positive on unchanged spec**

Tested: immediately after enabling the annotation (without any spec change), the condition reports "spec changes are pending and will be applied on resume" and a warning event is emitted.

The `equality.Semantic.DeepEqual` comparison between the actual StatefulSet (with server-side defaults applied by the API server) and the freshly computed expected StatefulSet (which may not include those defaults) produces a false diff.

This is misleading — the warning event and condition message tell the operator something is pending when nothing actually is.

pkg/controller/elasticsearch/driver/stateful/driver.go
```diff
@@ -69,12 +77,81 @@ func (d *Driver) Reconcile(ctx context.Context) *reconciler.Results {
 		return results.WithError(err)
 	}
 
+	if common.IsOrchestrationPaused(&d.ES) {
+		return results.WithError(d.reconcileCriticalSteps(ctx, sharedState, resolvedConfig))
+	}
+
```
pebrc, line 85: Nit: this sets `OrchestrationPaused: False` on every reconcile for every ES resource, even ones that have never used the annotation. All ES resources will accumulate this condition in their status. Consider only reporting the `False` condition when transitioning from paused (e.g. when the condition was previously `True`).

test/e2e/test/utils.go
pebrc: This introduces a second way to do the same thing as `Eventually`, used only by the new test file. There are 351 existing uses of `test.Eventually` across 76 files that all suffer from the same `TEST_TIMEOUT` error-swallowing problem described in your earlier comment.

Adding a per-test workaround for one test while leaving the other 351 call sites with the same issue creates a maintenance inconsistency without fixing the underlying problem. I'd prefer to either:
1. Fix the root cause (timeout/cancellation handling in the test runner), or
2. Drop this from the current PR and use `Eventually` for now — the e2e test doesn't seem to specifically need custom timeouts and fix it properly in a separate PR.

pebrc (changes requested):
Tested manually on a kind cluster against both ES 8.17.0 and 9.3.3. The core design is sound — the pause check is well-placed in the reconciliation flow (after shared resources/certs but before node spec changes), housekeeping correctly continues while paused (certs, secrets, health monitoring all verified), and scale-out is properly held back. Resuming applies pending changes immediately.

Found three bugs during testing:

1. **Phase never transitions to `Paused`** — the controller's post-reconcile logic always overwrites it to `Ready`.
2. **Pod deletion while paused leaves pods permanently not-ready** — the pre-stop hook registers a RESTART shutdown, but `reconcileCriticalSteps` never clears it. ES keeps the readiness port closed until the shutdown entry is explicitly deleted (this is by design in ES — `ReadinessService` checks for shutdown entry existence, not status). Reproduced on both 8.17.0 and 9.3.3. This is the most critical issue since pod disruptions during maintenance windows are exactly the scenario this feature is meant to handle.
3. **`hasPendingSpecChanges` false positive** — reports pending changes immediately after enabling the annotation even with no spec change, due to server-side defaulting drift in the actual-vs-computed StatefulSet comparison.

pkg/controller/elasticsearch/driver/stateful/driver.go
dtuck9: Is it preferred to set this to `Unknown` when it hasn't been previously set, or to leave the condition off entirely?

Also for my own understanding, could you explain why this is undesirable? I had deliberately done this to denote that orchestration was expected since the condition is set to `False`.

pkg/controller/elasticsearch/driver/stateful/driver.go
dtuck9: Could you confirm that adding the `Paused` phase is the approach we want?

I may be being too literal, but https://github.com/elastic/cloud-on-k8s/issues/9292#issue-4193936599 shows `ApplyingChanges` phase in §2: `d.ReconcileState.UpdateWithPhase(esv1.ElasticsearchApplyingChangesPhase)`

pkg/controller/elasticsearch/driver/stateful/driver.go
dtuck9: Do you have a suggestion on how to address this other than directly comparing each field of the `StatefulSetSpec`?

dtuck9: @moukoublen pointed me to https://github.com/elastic/cloud-on-k8s/blob/0b8582a1ab5f4f68404fbba826a247ad51becf51/pkg/controller/elasticsearch/sset/reconcile.go#L60-L62

And that seems to solve this, so I'm proceeding with that approach.

dtuck9: I also updated the e2e test to check for the expected message.

pkg/controller/elasticsearch/driver/stateful/driver.go
dtuck9: Along the same lines, does it make sense to explicitly requeue if there are pending changes?

pebrc: >  may be being too literal, but #9292 (comment) shows ApplyingChanges phase in §2: 

Maybe I did not consider this in enough detail in this earlier comment and I think it is right for you to challenge the need for the phase (IIUC). I think the condition is probably enough and might even be more useful in combination with the existing phase semantics: e.g. `Applying change` + `Condition: Paused` vs `Ready` + `Condition: Paused`. So I think its fine to drop and it also simplifies the logic.

pebrc: On second thought the requeue implied by the `ApplyingChanges` phase would send the controller into an unnecessarily tight loop in a paused state. So I think I would at least set a longer requeue interval if hasPendingChanges is true in paused state. It kind of duplicates the information in the k8s event that you are generating but maybe that is still useful during longer pauses and events expire (default is 1 hour afaik).

pkg/controller/elasticsearch/driver/stateful/driver.go
```diff
@@ -89,15 +92,33 @@ func (d *Driver) Reconcile(ctx context.Context) *reconciler.Results {
...
+	logger := log.WithValues("namespace", d.ES.Namespace, "es_name", d.ES.Name)
+
+	nodeShutdown := shutdown.NewNodeShutdown(state.ESClient, nodeNameToID, esclient.Restart, "", nil, logger)
+	actualPods, err := es_sset.GetActualPodsForCluster(d.Client, d.ES)
+	if err != nil {
+		return err
+	}
+
+	terminatingNodes := k8s.PodNames(k8s.TerminatingPods(actualPods))
+	if err = nodeShutdown.Clear(ctx, nodeShutdown.OnlyNonTerminatingNodes(terminatingNodes)); err != nil {
```
pebrc: Should this not be: 

```
nodeShutdown.Clear(ctx,
      esclient.ShutdownComplete.Applies,
      nodeShutdown.OnlyNonTerminatingNodes(terminating),
  )
```
to make sure we only clear actually complete node shutdowns and not other potential shutdowns in progress predating the pause that might require data migration i.e non-restart shutdowns.

pkg/controller/elasticsearch/driver/stateful/driver.go
```diff
@@ -89,15 +92,33 @@ func (d *Driver) Reconcile(ctx context.Context) *reconciler.Results {
 		*sharedState.ResourcesState, sharedState.KeystoreResources, sharedState.Meta, resolvedConfig))
 }
 
-func (d *Driver) reconcileCriticalSteps(ctx context.Context, state *shared.ReconcileState, resolvedConfig nodespec.ResolvedConfig) error {
+func (d *Driver) reconcileCriticalStepsWhilePaused(ctx context.Context, state *shared.ReconcileState, resolvedConfig nodespec.ResolvedConfig) error {
+	log := ulog.FromContext(ctx)
+	d.ReconcileState.UpdateWithPhase(esv1.ElasticsearchOrchestrationPaused)
```
pebrc: I think we also need to respect the expectations mechanism that tries to compensate for the cached client we are using. So a call to `expecationsSatisfied()` would be appropriate here I think.

The concrete risk: if pause is enabled right after (or during) a rolling upgrade, there may be outstanding expectations from that upgrade cycle. The cache could still show a pod as running that was just deleted, which means:
  - terminatingNodes is wrong → wrong shutdown filter → wrong shutdowns cleared
  - actualSets is stale → hasPendingSpecChanges gives a wrong answer

pkg/controller/elasticsearch/driver/stateful/driver.go
```diff
@@ -89,15 +92,33 @@ func (d *Driver) Reconcile(ctx context.Context) *reconciler.Results {
...
 	}
-	d.ReconcileState.UpdateWithPhase(esv1.ElasticsearchOrchestrationPaused)
+	esState := NewMemoizingESState(ctx, state.ESClient)
+	nodeNameToID, err := esState.NodeNameToID()
+	if err != nil {
+		return err
+	}
+	logger := log.WithValues("namespace", d.ES.Namespace, "es_name", d.ES.Name)
+
+	nodeShutdown := shutdown.NewNodeShutdown(state.ESClient, nodeNameToID, esclient.Restart, "", nil, logger)
```
pebrc: This unconditionally calls the implementation of the interface that is supported as of 7.15.2. This code will fail for earlier versions of Elasticsearch. So there is a cross dependency on https://github.com/elastic/cloud-on-k8s/issues/9041 here. Until 9041 is implemented the old code is still load bearing. I am not sure how you would like to handle it but I would at least put a version guard in here so that the code is functional as it stands.

pkg/controller/elasticsearch/driver/stateful/driver.go
pebrc: Adding a condition to every ES resource in every cluster creates status noise. The Kubernetes convention splits here:

  - Always-present conditions (`Ready`, `Available`, `Progressing`) — core lifecycle states that every consumer cares about
  - Appear-when-relevant conditions (`Disrupted` on PDBs, `MemoryPressure` on Nodes) — unusual states that only matter when active

`OrchestrationPaused` fits the second category. Its absence is unambiguous because the default state of every ES resource is "not paused." And the ECK version question is a non-concern in practice if ECK supports the annotation, it supports the condition.

I'd suggest: only set the condition once the annotation has been seen (i.e., set it to `True` when paused, set it to `False` when un-paused, and don't set it at all if the annotation was never used). That way users who use the feature get the full condition lifecycle, and everyone else doesn't accumulate a condition they never asked for.

pkg/controller/elasticsearch/driver/stateful/driver.go
dtuck9: Ok, I removed `Paused` and added a requeue after 5 minutes when paused w/ pending changes. There was precedent for 5 minutes, and I also thought that would provide decent balance between the retry interval frequency and applying the changes once unpaused.

Please let me know if you'd prefer a different interval.

pkg/controller/elasticsearch/driver/stateful/driver.go
dtuck9: How is this different than: https://github.com/elastic/cloud-on-k8s/blob/2e99d12a46e7bb7b367968926d16194f77c4df2f/pkg/controller/elasticsearch/driver/stateful/upgrade.go#L77

Or does that also have the same risk?

dtuck9: Never mind -- I found the client version check nested within a different function called a few lines later:

https://github.com/elastic/cloud-on-k8s/blob/2e99d12a46e7bb7b367968926d16194f77c4df2f/pkg/controller/elasticsearch/driver/stateful/upgrade.go#L367-L374

pkg/controller/elasticsearch/driver/stateful/driver.go
dtuck9: I also added `nodeShutdown.OnlyNodesInCluster,` based on https://github.com/elastic/cloud-on-k8s/blob/2e99d12a46e7bb7b367968926d16194f77c4df2f/pkg/controller/elasticsearch/driver/stateful/upgrade.go#L367-L374

dtuck9:
buildkite test this -f p=gke,t=TestPauseOrchestration*

dtuck9:
buildkite test this -f p=gke,t=TestPauseOrchestration*

pkg/controller/elasticsearch/driver/stateful/driver_test.go
```diff
@@ -6,16 +6,635 @@ package stateful
...
+		name                 string
+		k8sObjects           []crclient.Object // pre-existing StatefulSets beyond scriptsConfigMap
+		resolvedConfig       nodespec.ResolvedConfig
+		failK8sClient        bool
+		restartedNodeIDs     []string
+		removedNodeIDs       []string
+		wantErr              bool
+		wantRequeue          bool
+		wantCondStatus       corev1.ConditionStatus
+		wantCondMsgSubstr    string
```
pebrc, line 340: `wantRequeue`, `wantCondStatus`, and `wantCondMsgSubstr` are declared and populated in every test case but never asserted in the test body. The condition status/message and the 5-minute requeue are effectively untested at the unit level.

Something like this after the `Apply()` call should cover it:

```go
if tt.wantCondStatus != "" {
	idx := updatedES.Status.Conditions.Index(esv1.OrchestrationPaused)
	require.GreaterOrEqual(t, idx, 0, "expected OrchestrationPaused condition")
	assert.Equal(t, tt.wantCondStatus, updatedES.Status.Conditions[idx].Status)
	assert.Contains(t, updatedES.Status.Conditions[idx].Message, tt.wantCondMsgSubstr)
}
if tt.wantRequeue {
	requeue, _ := results.Aggregate()
	assert.NotZero(t, requeue.RequeueAfter, "expected requeue")
}
```

pkg/controller/elasticsearch/driver/stateful/driver.go
```diff
@@ -97,12 +108,125 @@ func (d *Driver) Reconcile(ctx context.Context) *reconciler.Results {
...
+		return results.WithError(err)
+	}
+
+	if supportsNodeShutdown(state.ESClient.Version()) {
+		esState := NewMemoizingESState(ctx, state.ESClient)
+		nodeNameToID, err := esState.NodeNameToID()
+		if err != nil {
+			return results.WithError(err)
+		}
+		logger := log.WithValues("namespace", d.ES.Namespace, "es_name", d.ES.Name)
```
pebrc: Nit: `namespace` and `es_name` are already in the log context — they're set by `NewReconciliationContext` at the top of the reconciliation loop. `ulog.FromContext(ctx)` on line 133 already carries them, so this `WithValues` call duplicates the fields. You can just pass `log` directly:

```suggestion
		logger := log
```

(Same pattern exists in `upgrade.go:75` but that's pre-existing.)

pkg/controller/elasticsearch/driver/stateful/driver.go
```diff
@@ -97,12 +108,125 @@ func (d *Driver) Reconcile(ctx context.Context) *reconciler.Results {
...
 	}
 
+	if common.IsOrchestrationPaused(&d.ES) {
+		return results.WithResults(d.reconcileCriticalStepsWhilePaused(ctx, sharedState, resolvedConfig, keystoreResources))
+	}
+
+	orchestrationPausedIndex := d.ES.Status.Conditions.Index(esv1.OrchestrationPaused)
+	if orchestrationPausedIndex >= 0 {
+		d.ReconcileState.ReportCondition(esv1.OrchestrationPaused, corev1.ConditionFalse, "Orchestration has resumed normally")
+	}
```
pebrc: The intent of this block — "if we were previously paused, clear the condition" — is hidden behind the `Index >= 0` idiom. Consider extracting a small helper to make the intent self-documenting:

```go
func (d *Driver) maybeResetPausedCondition() {
	if d.ES.Status.Conditions.Index(esv1.OrchestrationPaused) >= 0 {
		d.ReconcileState.ReportCondition(esv1.OrchestrationPaused, corev1.ConditionFalse, "Orchestration has resumed normally")
	}
}
```

Then this becomes just:
```go
d.maybeResetPausedCondition()
```

pebrc (approved):
Tested manually on a local k3d cluster against ES 8.17.0: pause with no changes, upscale while paused (blocked correctly), resume (upscale proceeds), pod deletion while paused (readiness recovers via shutdown cleanup), and managed=false precedence. Phases, conditions, and k8s events all behave as expected at every step.

A few minor suggestions inline — nothing blocking.

pkg/controller/elasticsearch/driver/stateful/driver_test.go
dtuck9: Good catch

pkg/controller/common/unmanaged.go
```diff
@@ -14,12 +14,27 @@ import (
...
+// IsOrchestrationPaused returns true if the PauseOrchestrationAnnotation exists and is set to true on the given resource
+// when non-critical orchestration steps should be skipped.
+func IsOrchestrationPaused(object metav1.Object) bool {
+	return object.GetAnnotations()[PauseOrchestrationAnnotation] == "true"
+}
+
 // IsUnmanaged checks if a given resource is currently unmanaged.
+//
+// Deprecated: Migrate to IsOrchestrationPaused instead. See https://github.com/elastic/cloud-on-k8s/issues/9295.
 func IsUnmanaged(ctx context.Context, object metav1.Object) bool {
```
barkbay, line 38: Should we emit a validating webhook warning when `eck.k8s.elastic.co/managed=false` (same pattern as `deprecatedStackVersionWarning` in [validation/warnings.go](https://github.com/elastic/cloud-on-k8s/blob/main/pkg/controller/elasticsearch/validation/warnings.go))?

```go
const managedFalseDeprecationWarningMsg = "eck.k8s.elastic.co/managed is deprecated, use eck.k8s.elastic.co/pause-orchestration instead"

func managedFalseDeprecationWarning(es esv1.Elasticsearch) field.ErrorList {
	if es.Annotations[common.ManagedAnnotation] != "false" {
		return nil
	}
	return field.ErrorList{field.Invalid(
		field.NewPath("metadata").Child("annotations").Key(common.ManagedAnnotation),
		"false", managedFalseDeprecationWarningMsg)}
}
```

pkg/controller/elasticsearch/driver/stateful/driver.go
```diff
@@ -97,12 +108,128 @@ func (d *Driver) Reconcile(ctx context.Context) *reconciler.Results {
...
 }
 
+func (d *Driver) maybeResetPausedCondition() {
+	orchestrationPausedIndex := d.ES.Status.Conditions.Index(esv1.OrchestrationPaused)
+	if orchestrationPausedIndex >= 0 {
+		d.ReconcileState.ReportCondition(esv1.OrchestrationPaused, corev1.ConditionFalse, "Orchestration has resumed normally")
+	}
+}
+
+func (d *Driver) reconcileCriticalStepsWhilePaused(
```
barkbay, line 145:
```suggestion
// reconcileCriticalStepsWhilePaused runs when pause-orchestration is enabled.
//
// Motivation (see https://github.com/elastic/cloud-on-k8s/issues/9250): during node-level maintenance,
// a concurrent Elasticsearch spec change can otherwise trigger rolling upgrades on top of the drain,
// compounding disruption. Pausing orchestration narrows the freeze to spec-driven StatefulSet work
// instead of using managed=false, which would skip all reconciliation (including certs and discovery)
// and risk cluster degradation over long windows.
//
// Shared reconciliation (certs, services, users, health, etc.) has already completed in Reconcile;
// this path skips reconcileNodeSpecs but still performs StatefulSet-adjacent work: wait on expectations,
// reconcile the PDB, clear completed restart shutdown records when supported, detect pending spec changes
// versus the live cluster, and report OrchestrationPaused (with a warning event and periodic requeue
// when changes are pending).
func (d *Driver) reconcileCriticalStepsWhilePaused(
```

pkg/controller/elasticsearch/driver/stateful/driver.go
```diff
@@ -97,12 +108,128 @@ func (d *Driver) Reconcile(ctx context.Context) *reconciler.Results {
...
+	ctx context.Context,
+	state *shared.ReconcileState,
+	resolvedConfig nodespec.ResolvedConfig,
+	keystoreResources *keystore.Resources,
+) *reconciler.Results {
+	log := ulog.FromContext(ctx)
```
barkbay, line 152:
```suggestion
	defer tracing.Span(&ctx)()
	log := ulog.FromContext(ctx)
```

pkg/controller/elasticsearch/driver/stateful/driver.go
```diff
@@ -97,12 +108,128 @@ func (d *Driver) Reconcile(ctx context.Context) *reconciler.Results {
...
+
+	for _, thisSet := range actualSets {
+		thatSet, exists := otherSetsByName[thisSet.Name]
+		if !exists {
+			log.V(1).Info("statefulset does not exist in other sets", "name", thisSet.Name)
+			return true
+		}
+
+		if !es_sset.EqualTemplateHashLabels(thatSet, thisSet) {
+			return true
```
barkbay, line 245: nit, for consistency?
```suggestion
			ssetLogger(ctx, thisSet).V(1).Info("statefulset template hash differs",
				"expected_hash", hash.GetTemplateHashLabel(thatSet.Labels),
				"actual_hash", hash.GetTemplateHashLabel(thisSet.Labels))
			return true
```

pkg/controller/elasticsearch/driver/stateful/driver.go
```diff
@@ -97,12 +108,128 @@ func (d *Driver) Reconcile(ctx context.Context) *reconciler.Results {
...
+
+	otherSetsByName := make(map[string]appsv1.StatefulSet, len(expectedSets))
+	for _, otherSet := range expectedSets {
+		otherSetsByName[otherSet.Name] = otherSet
+	}
+
+	for _, thisSet := range actualSets {
+		thatSet, exists := otherSetsByName[thisSet.Name]
+		if !exists {
+			log.V(1).Info("statefulset does not exist in other sets", "name", thisSet.Name)
```
barkbay: nit: let's use [`ssetLogger(ctx, thisSet)`](https://github.com/elastic/cloud-on-k8s/blob/a1491a1054d561ebea551f831c0dbb94fb163b43/pkg/controller/elasticsearch/driver/stateful/log.go#L16-L20) to log with key `statefulset_name`
```suggestion
			ssetLogger(ctx, thisSet).V(1).Info("statefulset does not exist in other sets")
```

pkg/controller/elasticsearch/driver/stateful/driver.go
```diff
@@ -97,12 +108,128 @@ func (d *Driver) Reconcile(ctx context.Context) *reconciler.Results {
...
+		return false, err
+	}
+	return hasSpecDiff(ctx, actualSets, expectedResources.StatefulSets()), nil
+}
+
+func hasSpecDiff(ctx context.Context, actualSets, expectedSets es_sset.StatefulSetList) bool {
+	log := ulog.FromContext(ctx)
+
+	if len(actualSets) != len(expectedSets) {
+		log.V(1).Info("Different number of statefulsets found")
```
barkbay:
```suggestion
		log.V(1).Info("Different number of statefulsets found",
			"actual_count", len(actualSets), "expected_count", len(expectedSets))
```

barkbay (approved):
LGTM

pkg/controller/common/unmanaged.go
dtuck9: That makes sense.

dtuck9: For my own understanding, why wouldn't we emit this warning when `eck.k8s.elastic.co/managed=true`?

barkbay: > For my own understanding, why wouldn't we emit this warning when eck.k8s.elastic.co/managed=true?

You are right, we should clearly state that this is deprecated, regardless of its value.

pkg/controller/common/unmanaged.go
```diff
@@ -14,12 +14,27 @@ import (
...
 
 const (
-	// ManagedAnnotation annotation
 	LegacyPauseAnnoation = "common.k8s.elastic.co/pause"
-	ManagedAnnotation    = "eck.k8s.elastic.co/managed"
+	// ManagedAnnotation is the annotation set in order to signal to the operator to skip reconciliation entirely for
+	// the given resource.
+	//
+	// Deprecated: Use PauseOrchestrationAnnotation instead. See https://github.com/elastic/cloud-on-k8s/issues/9295.
+	ManagedAnnotation = "eck.k8s.elastic.co/managed"
```
naemono, line 22: Ok, I was worried about the other controllers here, but I see the issue noted above, and see it will all be handled there. 👍

pkg/controller/elasticsearch/driver/stateful/driver.go
```diff
@@ -97,12 +110,146 @@ func (d *Driver) Reconcile(ctx context.Context) *reconciler.Results {
...
+
+	otherSetsByName := make(map[string]appsv1.StatefulSet, len(expectedSets))
+	for _, otherSet := range expectedSets {
+		otherSetsByName[otherSet.Name] = otherSet
+	}
+
+	for _, thisSet := range actualSets {
+		thatSet, exists := otherSetsByName[thisSet.Name]
+		if !exists {
+			ssetLogger(ctx, thisSet).V(1).Info("statefulset does not exist in other sets", "statefulset_name", thisSet.Name)
```
naemono: `statefulset_name` is duplicative here, as `ssetLogger` automatically adds it. Also on line 242

naemono (approved):
lgtm. One small thing noted.

pkg/controller/elasticsearch/driver/stateful/driver.go
dtuck9: Ah ok, I misunderstood Morello's earlier comment. Thanks!

naemono (approved):
👍

Issue #9165, opened 2026-02-25
TestAutoOpsAgentPolicyEnterprise: EDOT collector metricbeatreceiver does not retry on transient ES connection failure


pebrc:
## Summary

`TestAutoOpsAgentPolicyEnterprise/AutoOpsAgentPolicy_status_should_be_ready` fails consistently across multiple nightly runs because the EDOT collector's `metricbeatreceiver` does not retry when it fails to connect to Elasticsearch at startup. A brief window where the ES service endpoint is not yet accepting connections causes a permanent failure.

## Observed on

- **Nightly build**: 1196 (2026-02-25)
- **Affected runs**: `eck-e2e-gke-dsgy-1196` (GKE) and `eck-e2e-resilience-hqgh-1196` (resilience)
- **Operator**: `docker.elastic.co/eck-snapshots/eck-operator:3.4.0-SNAPSHOT-87a508f2`
- **EDOT collector**: `docker.elastic.co/elastic-agent/elastic-otel-collector-wolfi:9.3.0`
- **Kubernetes**: v1.33.8-gke.1026000

## Root Cause

The EDOT collector pod starts immediately after the operator creates its deployment. On its first connection attempt to the ES service, it gets `connection refused` — the ES cluster has just been reported as ready but the K8s service endpoint isn't fully routing yet (a ~2 second race window).

The `metricbeatreceiver` fails permanently on this initial error and never retries. The `healthcheckv2` extension (configured with `include_permanent_errors: true`) then returns HTTP 500 on every readiness probe for the remaining 15 minutes until the test times out.

### EDOT collector logs (from e2e o11y cluster)

```
2026-02-25T00:29:53Z info  Everything is ready. Begin running and processing data.
2026-02-25T00:29:53Z error Error dialing dial tcp 34.118.227.203:9200: connect: connection refused
2026-02-25T00:29:53Z error error starting metricbeat receiver:
    beat receiver run error: failed to register Cloud Connected cluster:
    failed to load cluster info: failed to send request:
    Get "https://es-es-internal-http.e2e-pxlwh-mercury.svc:9200/":
    dial tcp 34.118.227.203:9200: connect: connection refused
2026-02-25T00:30:23Z info  No non-zero metrics in the last 30s
... (repeats every 30s for 15 minutes until pod is killed)
```

### Timeline

| Time | Event |
|---|---|
| 00:29:48 | ES cluster `es` in `e2e-pxlwh-mercury` reported as ready |
[...]

PR #9333, opened 2026-04-09
Implement autoops agent support for presenting client certificates to elasticsearch

pkoutsovasilis:
This PR implements client certificate support for the AutoOps agent when connecting to Elasticsearch clusters that have client authentication enabled. Unlike the association-based resources (Kibana, APM, etc.), the AutoOps agent manages its own secrets directly - it auto-generates and reconciles a self-signed client certificate per policy+ES pair, labeled for trust bundle discovery by the ES controller.

Relates https://github.com/elastic/cloud-on-k8s/issues/9081

## Changes

  - **Client certificate secret naming**: Added `ClientCertSecret` naming function in `autoopsv1alpha1` to generate deterministic secret names per policy+ES pair (using the `cc` suffix)
  - **Client certificate reconciliation**: Added `reconcileAutoOpsESClientCertSecret` which generates a self-signed client certificate using the common `certificates.Reconciler`, labeled with soft-owner and `ClientCertificateLabelName` labels so the ES trust bundle controller discovers and trusts it. Includes certificate
   rotation scheduling via requeue
  - **Client certificate cleanup**: Added `deleteAutoOpsESClientCertSecret` to remove the client certificate secret when client authentication is no longer required on the  ES cluster. The GC also handles orphaned client cert secrets during selector changes
  - **AutoOps config template**: Extended both the 10s and 24h metricbeat receiver blocks in the config template to conditionally include `ssl.certificate` and `ssl.key` directives when client authentication is required
  - **Deployment pod spec**: When `HasClientAuthenticationRequired` is true for an ES cluster, the deployment builder mounts the client certificate secret as an additional volume alongside the existing CA volume
  - **AutoOps e2e test**: Added `TestClientAuthRequiredTransition` which verifies the AutoOps agent can connect to an ES cluster with client authentication enabled. 

For the e2e tests this PR  includes a workaround for the agent's lack of connection retry which should be removed as soon as https://github.com/elastic/cloud-on-k8s/issues/9165 is resolved


prodsecmachine:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/61cac02b-a5ec-4c76-9402-804967ab3a27?refs=d6fa8382255bdeb8276834249bec1b56235e371f&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/61cac02b-a5ec-4c76-9402-804967ab3a27/license?refs=d6fa8382255bdeb8276834249bec1b56235e371f&source=prChecksComment) |

pkoutsovasilis:
buildkite test this -f p=gke,t=TestClientAuthRequired.*,E2E_TAGS=autoops -m s=9.3.3

pkoutsovasilis:
buildkite test this -f p=gke,t=TestClientAuthRequired.*,E2E_TAGS=autoops -m s=9.3.3,s=9.4.0-SNAPSHOT

pkoutsovasilis:
buildkite test this -f p=gke,t=TestClientAuthRequired.*,E2E_TAGS=autoops -m s=9.3.3,s=9.4.0-SNAPSHOT

pkoutsovasilis:
buildkite test this -f p=gke,t=TestClientAuthRequired.*,E2E_TAGS=autoops -m s=9.3.3,s=9.4.0-SNAPSHOT

pkoutsovasilis:
buildkite test this -f p=gke,t=TestClientAuthRequired.*,E2E_TAGS=autoops -m s=9.3.3,s=9.4.0-SNAPSHOT

pkoutsovasilis:
buildkite test this -f p=gke,t=TestClientAuthRequired.*,E2E_TAGS=autoops -m s=9.3.3,s=9.4.0-SNAPSHOT

pkoutsovasilis:
buildkite test this -f p=gke,t=TestClientAuthRequired.*,E2E_TAGS=autoops -m s=9.3.3,s=9.4.0-SNAPSHOT

pkoutsovasilis:
buildkite test this -f p=gke,t=TestClientAuthRequired.*,E2E_TAGS=autoops -m s=9.4.0,s=9.5.0-SNAPSHOT

pkg/controller/autoops/secret.go
```diff
@@ -103,6 +106,79 @@ func (r *AgentPolicyReconciler) reconcileAutoOpsESCASecret(
...
+		Owner:        &policy,
+		Metadata:     meta,
+		CertRotation: certRotation,
+	}
+
+	commonName := policy.GetName()
+	orgUnit := policy.GetName()
+
+	results := reconciler.NewResult(ctx)
+	clientCertSecret, err := certReconciler.ReconcileClientCertificate(ctx, secretName, commonName, orgUnit, labels)
```
pebrc, line 172: This function is a near-verbatim copy of `association.ReconcileManagedClientCert` (pkg/controller/association/client_cert.go:88-127): same `certificates.Reconciler` construction, same `RotationParams`, same `GetPrimaryCertificate` / `ShouldRotateIn` requeue. The only thing blocking direct reuse is that `ReconcileManagedClientCert` is typed against `commonv1.Association`, which AutoOps doesn't implement.

Not sure if there is a good way to work around the different interfaces in use here. E.g. by using primitives in the parameter list directly in a third shared function of by looking at another interface. Not a blocker.

pkg/controller/autoops/configmap.go
```diff
@@ -58,6 +63,10 @@ const autoOpsESConfigTemplate = `receivers:
 {{- if .SSLEnabled}}
           ssl.verification_mode: certificate
           ssl.certificate_authorities: ["{{ .CACertPath }}"]
+{{- if .ClientCertPath}}
```
pebrc: This 5-line `ssl.*` conditional is now duplicated verbatim across the 10s and 24h metricbeat module blocks. Two is the threshold where it's worth extracting — the next module that needs SSL config (or the next time the SSL keys change shape) will require touching both copies in lockstep.

A `{{define "ssl"}}…{{end}}` block invoked via `{{template "ssl" .}}` would collapse the duplication. Cheap to do now while the template is small.

pkg/controller/autoops/secret.go
```diff
@@ -103,6 +106,79 @@ func (r *AgentPolicyReconciler) reconcileAutoOpsESCASecret(
...
+		return results.WithError(err)
+	}
+
+	// Schedule requeue for certificate rotation.
+	primaryCert, err := certificates.GetPrimaryCertificate(clientCertSecret.CertPem())
+	if err != nil {
+		return results.WithError(err)
+	}
+	results.WithReconciliationState(
+		reconciler.RequeueAfter(certificates.ShouldRotateIn(time.Now(), primaryCert.NotAfter, certRotation.RotateBefore)).ReconciliationComplete(),
```
pebrc, line 191: On rotation, the cert is re-issued into the same secret and the configmap content (and therefore the deployment's `configHash` annotation) doesn't change, so the deployment is not rolled. The pod picks up the new cert via kubelet's projected-secret refresh (~60s by default).

This matches the CA secret's behaviour (also no rollout on rotation) so it's not new — but worth either a brief comment here pointing out the assumption, or factoring a cert-data fingerprint into the configHash if you want a stronger guarantee. Low priority.

pkg/controller/autoops/secret.go
```diff
@@ -103,6 +106,79 @@ func (r *AgentPolicyReconciler) reconcileAutoOpsESCASecret(
...
+// The certificate is unique to this policy+ES pair and is labeled so the ES trust
+// bundle controller discovers it and includes it in the client CA trust bundle.
+// Returns a requeue duration for certificate rotation.
+func (r *AgentPolicyReconciler) reconcileAutoOpsESClientCertSecret(
+	ctx context.Context,
+	policy autoopsv1alpha1.AutoOpsAgentPolicy,
+	es esv1.Elasticsearch,
+) *reconciler.Results {
+	ulog.FromContext(ctx).V(1).Info("Reconciling AutoOps ES client cert secret",
+		"es_namespace", es.Namespace, "es_name", es.Name)
```
pebrc, line 140: `reconcileAutoOpsESCASecret` above gates on `es.Status.Phase != esv1.ElasticsearchReadyPhase` because it depends on the ES `http-certs-public` secret being populated. The client cert reconciler has no such dependency, so generating eagerly is defensible.

The wrinkle is sequencing: this cert can land before the ES trust bundle controller has discovered and trusted it, so the agent's first connection attempt may fail. Combined with the missing retry called out in #9165, that's the gap the e2e test is working around. Worth a one-line comment here referencing #9165 so the next reader doesn't wonder why the readiness gate is missing here but not above.

pkg/controller/autoops/configmap_test.go
```diff
@@ -179,6 +180,93 @@ service:
       encoding: json
 `)
 
+var defaultConfigWithSSLAndClientCert = []byte(`
```
pebrc, line 183: This fixture is ~95% identical to `defaultConfigWithSSL`. Since the only meaningful delta is the four `ssl.certificate` / `ssl.key` lines, a smaller assertion that targets those keys (e.g. assert presence and value of `receivers.metricbeatreceiver.metricbeat.modules[*].ssl.certificate`) would be less to maintain when unrelated fields change.

Not blocking — the golden-yaml pattern is already in use for the other cases. Just flagging that the second copy doubles the maintenance surface for the SSL block.

test/e2e/test/autoops/builder.go
```diff
@@ -16,6 +16,9 @@ import (
 	"github.com/elastic/cloud-on-k8s/v3/test/e2e/test"
 )
 
+// PolicyNameLabelKey is re-exported from the controller package for use in e2e tests.
+const PolicyNameLabelKey = controllerautoops.PolicyNameLabelKey
```
pebrc: Nit: since the controller package is already imported here as `controllerautoops`, callers in e2e code can reference `controllerautoops.PolicyNameLabelKey` directly. Re-exporting it as a value copy adds a parallel symbol that could drift. Up to you — happy to drop this if there's a reason for the indirection.

pebrc (approved):
LGTM. Nits below are courtesy of Claude. I did run a simple local test.

pkoutsovasilis:
buildkite test this -f p=gke,t=TestClientAuthRequired.*,E2E_TAGS=autoops -m s=9.3.4,s=9.4.0,s=9.5.0-SNAPSHOT

test/e2e/test/autoops/builder.go
pkoutsovasilis: changed in e0458b701514ded91dd56d08607466afeef3b932

pkg/controller/autoops/secret.go
pkoutsovasilis: added godoc why this func shouldn't gate on `es.Status.Phase != esv1.ElasticsearchReadyPhase` in ae4fbf81448137d316aa0e5ec09331156c57abcd

pkg/controller/autoops/secret.go
pkoutsovasilis: changed in 65f3387a979645c895d4005b095bed82385584f4

pkg/controller/autoops/configmap.go
pkoutsovasilis: implemented the "ssl" template in 5e2ab9d1c8a387b6ca1e2306149adf61711f9cbd

pkg/controller/autoops/secret.go
pkoutsovasilis: actually I reverted this in 5283b78c7cf1e19215bb00cd02cf2cf65c7fd39a to address [this comment](https://github.com/elastic/cloud-on-k8s/pull/9333#discussion_r3209908764) which is a real issue

pkg/controller/autoops/secret.go
pkoutsovasilis: ok this is a real issue we should rollout the autoops agent deployment on client cert changes since certificates are not hot reloaded. fixed in d49c8c002dea12b962b9f166477afb7bb9b21de8

pkg/controller/autoops/configmap_test.go
pkoutsovasilis: Thanks - I think I am gonna stick with duplication for now, mainly because the targeted-assertion alternative splits validation: four cases compare full golden YAML, one asserts a subset. I'd rather pay the duplication cost than fork the strategy for one case. Happy to revisit if the SSL block grows or another module needs SSL config.

pkoutsovasilis:
buildkite test this -f p=gke,t=.*,E2E_TAGS=autoops -m s=9.3.4,s=9.4.0,s=9.5.0-SNAPSHOT

pkoutsovasilis:
buildkite test this -f p=gke,t=.*,E2E_TAGS=autoops -m s=9.3.4,s=9.4.0,s=9.5.0-SNAPSHOT

pkg/controller/autoops/gc.go
```diff
@@ -240,7 +240,7 @@ func cleanupOrphanedConfigMaps(
 	return nil
 }
 
-// cleanupOrphanedSecrets removes secrets (CA and API key) for ES clusters that no longer match the selector.
+// cleanupOrphanedSecrets removes secrets (CA, client cert, and API key) for ES clusters that no longer match the selector.
```
pebrc, line 243: When the resource selector changes such that an ES cluster is no longer matched, `cleanupOrphanedSecrets` deletes the orphaned client cert (or CA) secret but doesn't `r.dynamicWatches.Secrets.RemoveHandlerForKey(secret.Name)`. The watch entry registered in `reconcileAutoOpsESClientCertSecret` / `reconcileAutoOpsESCASecret` leaks until policy deletion — at which point `onDelete` enumerates labeled secrets to clean up watches, but those secrets are already gone here, so the entries stick around.

Slow leak, proportional to selector churn — not blocking. Same pattern already applies to CA secrets, so the fix would touch both. Flagging for a future cleanup.

pkoutsovasilis: created [this issue](https://github.com/elastic/cloud-on-k8s/issues/9433), I am working on the fix

PR #9346, opened 2026-04-13
Simplified container resources spec for all ECK CRDs

### LINKED ISSUE #9220: Simplified container resource specification across all CRDs
Add a top-level `resources` field to all pod-owning CRD specs as a shorthand for setting container resource requests and limits on the main application container. This avoids the deeply nested `podTemplate.spec.containers[name=<main>].resources` path that users currently need to navigate.

## Motivation

Setting resource requests and limits is one of the most common customizations users make, yet it requires navigating 4 levels of nesting through `podTemplate`:

```yaml
# Current: verbose and error-prone
nodeSets:
  - name: default
    count: 3
    podTemplate:
      spec:
        containers:
          - name: elasticsearch
            resources:
              requests:
                memory: 8Gi
                cpu: 4
              limits:
                memory: 8Gi
                cpu: 4
```

This verbosity applies equally to Kibana, Beats, Agent, and all other CRDs. The nesting depth is a frequent source of user errors (see #3021) and makes simple deployments unnecessarily complex.

## Proposed Schema

A new `resources` field of type `corev1.ResourceRequirements` on the relevant spec level — `NodeSet` for Elasticsearch, top-level spec for all other CRDs.

**Elasticsearch:**

```yaml
kind: Elasticsearch
metadata:
  name: my-cluster
spec:
  version: 9.4.0
  nodeSets:
    - name: default
      count: 3
      resources:
        requests:
          memory: 8Gi
          cpu: 4
        limits:
          memory: 8Gi
          cpu: 4
```

**Kibana:**

```yaml
kind: Kibana
metadata:
  name: my-kibana
spec:
  version: 9.4.0
  resources:
    requests:
      memory: 1Gi
      cpu: 500m
```

## Scope

This should be implemented across all pod-owning CRDs:

- **Elasticsearch** — `nodeSets[].resources` (applies to the `elasticsearch` container)
- **Kibana** — `spec.resources` (applies to the `kibana` container)
- **APM Server** — `spec.resources` (applies to the `apm-server` container)
- **Beats** — `spec.resources` (applies to the main Beat container)
- **Elastic Agent** — `spec.resources` (applies to the `agent` container)
- **Enterprise Search** — `spec.resources` (applies to the `enterprise-search` container)
- **Logstash** — `spec.resources` (applies to the `logstash` container)
- **Elastic Maps Server** — `spec.resources` (applies to the `maps-server` container)

## Precedence

The top-level `resources` field takes precedence over any resource specification in `podTemplate`. If both are set, the value in `resources` wins. This ensures that the simplified field is always authoritative and users don't need to worry about conflicting values buried in `podTemplate`.

## Suggested Tasks

*These are suggestions; the assignee may adapt as needed.*

- [ ] Add `resources` field to `NodeSet` type
- [ ] Add `resources` field to all other pod-owning CRD spec types
- [ ] Implement resource merging logic (top-level trumps podTemplate)
- [ ] Regenerate CRDs
- [ ] Update API docs
- [ ] Add unit tests for precedence behavior
- [ ] Add e2e test
- [ ] Update user-facing documentation and examples

naemono:
# Simplified container resource spec across all CRDs

Closes #9220

## Summary

Adds a top-level `resources` field (of type `commonv1.Resources`) to all pod-owning CRD spec types as a shorthand for setting CPU and memory requests/limits on the main application container. Users no longer need to navigate the four-level nested `podTemplate.spec.containers[name=<main>].resources` path for the most common customization.

The shorthand applies to all eight pod-owning CRDs:

- `Elasticsearch` — `spec.nodeSets[].resources` (per-NodeSet, applied to the `elasticsearch` container)
- `Kibana` — `spec.resources` (applied to the `kibana` container)
- `APM Server` — `spec.resources` (applied to the `apm-server` container)
- `Beats` — `spec.resources` (applied to the main Beat container)
- `Elastic Agent` — `spec.resources` (applied to the `agent` container)
- `Enterprise Search` — `spec.resources` (applied to the `enterprise-search` container)
- `Logstash` — `spec.resources` (applied to the `logstash` container)
- `Elastic Maps Server` — `spec.resources` (applied to the `maps-server` container)
- `Elastic AutoOps Agent` — `spec.resources` (applied to the `autoops-agent` container)
- `Elastic Package Registry` — `spec.resources` (applied to the `package-registry` container)

Precedence: shorthand CPU/memory keys override values set on the main container in `podTemplate`. Non-CPU/memory resource keys in `podTemplate` (e.g. `ephemeral-storage`, `claims`) are preserved.

### API types — add `Resources` field to every pod-owning spec

**`pkg/apis/elasticsearch/v1/elasticsearch_types.go`**
- Adds `Resources commonv1.Resources` to `NodeSet` (per-NodeSet, not top-level spec).

**`pkg/apis/kibana/v1/kibana_types.go`, `pkg/apis/apm/v1/apmserver_types.go`, `pkg/apis/beat/v1beta1/beat_types.go`, `pkg/apis/agent/v1alpha1/agent_types.go`, `pkg/apis/enterprisesearch/v1/enterprisesearch_types.go`, `pkg/apis/logstash/v1alpha1/logstash_types.go`, `pkg/apis/maps/v1alpha1/maps_types.go`**
- Adds top-level `Resources commonv1.Resources` to each spec type.
- Regenerated deepcopy for each API group.

### Merge/override logic — shared helper for all controllers

**`pkg/apis/common/v1alpha1/resources.go`**
- Adds `WithResourcesAndOverrides(base, override)` that merges the shorthand onto a base container spec. Preserves non-CPU/memory keys (e.g. `ephemeral-storage`) and existing `ResourceClaims`.
- Keeps existing `WithResources` semantics intact for the no-shorthand path.

**`pkg/controller/common/defaults/pod_template.go`**
- Threads the new merge helper through `PodTemplateBuilder` so each controller opts in by passing the shorthand value at build time.
- Preserves the existing LimitRange escape hatch: an empty (non-nil) `requests` / `limits` map in `podTemplate` still suppresses operator defaults.
- Ensures CPU/memory defaults are not dropped when the main container has only `ResourceClaims` set (no `requests`/`limits`).

### Autoscaling — route through the new merge path

**`pkg/controller/autoscaling/elasticsearch/reconcile.go`**
**`pkg/controller/elasticsearch/driver/stateful/autoscaling.go`**
**`pkg/controller/elasticsearch/autoscaling/elasticsearch/resources/resources.go`**
- Updates the autoscaler so computed NodeSet resources flow through `WithResourcesAndOverrides` instead of being written directly into `podTemplate`. Without this, the autoscaler would overwrite the user's shorthand with its own podTemplate-level resources.
- Refreshes the `testdata/.../*-expected.yml` fixtures to reflect the merged output.

### Validation — non-blocking admission warning on overlap

**`pkg/apis/common/v1/resources_warnings.go`** (new)
**`pkg/controller/elasticsearch/validation/warnings.go`**
**`pkg/apis/*/webhook.go`** (all CRDs)
- Adds a warning (`shorthandResourcesOverrideWarning`) emitted by the validating webhook when `spec.resources` (or `spec.nodeSets[i].resources`) sets a CPU or memory value that is **also** set on the main container in `podTemplate`.
- Warning is per-NodeSet for Elasticsearch (identifies `spec.nodeSets[i].podTemplate`), per-resource for all other CRDs.
- Non-blocking: `kubectl apply` succeeds and the CR is persisted. Shorthand still wins at reconcile time.
- Does not fire when shorthand and podTemplate set *disjoint* keys (e.g. shorthand CPU + podTemplate memory).

### CRDs, schema, docs

**`config/crds/v1/**`, `deploy/eck-operator-crds/templates/all-crds.yaml`**
- Regenerated CRDs for all eight types with the new `resources` field.

**`docs/reference/api-reference/main.md`**, **`README.md`**
- API reference and README updated with the new shorthand.

### Tests

**`pkg/apis/common/v1alpha1/resources_test.go`**, **`pkg/apis/common/v1/resources_warnings_test.go`** (new)
**`pkg/controller/common/defaults/pod_template_test.go`** (new, 317 lines)
- Unit coverage for the merge helper: empty base, empty override, disjoint keys, overlapping keys, ResourceClaims preservation, LimitRange escape hatch, nil-to-empty no-op guard.

**`pkg/controller/elasticsearch/nodespec/nodeset_resources_test.go`** (new, 473 lines)
**`pkg/controller/elasticsearch/validation/warnings_test.go`** (new, 146 lines)
**`pkg/controller/elasticsearch/autoscaling/elasticsearch/reconcile_test.go`** (new, 373 lines)
**`pkg/controller/{kibana,apmserver,beat,agent,enterprisesearch,logstash,maps}/resources_test.go`** (new)
- Per-controller tests asserting the shorthand is honored, podTemplate preservation rules hold, and the autoscaler respects the new field.

**`test/e2e/es/autoscaling_test.go`**, **`test/e2e/es/topology_test.go`**, **`test/e2e/es/stackconfigpolicy_test.go`**, others
- E2E tests updated to exercise `spec.nodeSets[].resources` on at least one scenario and to remain correct under the new merge path.

## Behavior

Existing manifests are unaffected. When the shorthand is unset, the pod template is built exactly as before:

```yaml
# Before and after this PR — identical result
spec:
  nodeSets:
  - name: default
    count: 3
    podTemplate:
      spec:
        containers:
        - name: elasticsearch
          resources:
            requests: { memory: 8Gi, cpu: 4 }
            limits:   { memory: 8Gi, cpu: 4 }
```

With the shorthand set:

```yaml
spec:
  nodeSets:
  - name: default
    count: 3
    resources:
      requests: { memory: 8Gi, cpu: 4 }
      limits:   { memory: 8Gi, cpu: 4 }
```

Produces the same StatefulSet container resources — without the four levels of nesting.

When **both** are set, shorthand CPU/memory wins and `podTemplate` non-CPU/memory keys are preserved:

```yaml
spec:
  nodeSets:
  - name: default
    count: 3
    podTemplate:
      spec:
        containers:
        - name: elasticsearch
          resources:
            requests:
              cpu: 100m              # overridden by shorthand
              memory: 2Gi            # overridden by shorthand
              ephemeral-storage: 5Gi # preserved
    resources:
      requests: { cpu: 750m, memory: 6Gi }
```

Effective container `requests`: `{cpu: 750m, memory: 6Gi, ephemeral-storage: 5Gi}`. An admission warning is emitted naming `spec.nodeSets[0].podTemplate`.

## Test Plan

<details>
<summary>Click to expand: edge cases and scenarios tested</summary>

### Core Functionality
- Fresh create with `nodeSets[].resources` setting both CPU and memory.
- Default resources applied when neither shorthand nor podTemplate resources are set (ECK `memory=2Gi` default preserved).
- Shorthand overrides CPU/memory from podTemplate; `ephemeral-storage` and other keys preserved.
- Backward compatibility: podTemplate-only manifests work exactly as before.
- Partial override — memory-only via shorthand does not introduce CPU keys.
- Partial override — single limit key (e.g. `limits.cpu`) does not wipe the other keys from the base.

### User Override / Escape Hatches
- LimitRange escape hatch preserved: empty (non-nil) maps in podTemplate still suppress operator defaults under `WithResourcesAndOverrides`.
- User-defined init container resources in `podTemplate` are not overridden by the shorthand.

### Init Container Inheritance
- Operator-managed init containers with no resources set inherit the merged main-container resources.

### Multi-NodeSet
- Multiple NodeSets with different `resources` each produce a StatefulSet with its own independent resource values.
- Mixed NodeSets — one with `nodeSets[].resources`, one without — both reconcile correctly.

### Lifecycle
- Removing `nodeSets[].resources` reverts the cluster to ECK defaults on the next reconcile.
- Changing `nodeSets[].resources` triggers a rolling restart.
- Scale-up with `nodeSets[].resources` set works without surprises.

### Version Upgrade
- Elasticsearch version upgrade completes cleanly with `nodeSets[].resources` set.

### Configuration Edge Cases
- `requests`-only (no `limits`) accepted; ECK default `limits.memory=2Gi` base preserved.
- Zero-value quantities (`cpu: "0"`) accepted; operator does not panic.
- `kubectl explain elasticsearch.spec.nodeSets.resources` (and `.requests`/`.limits`) documents the new field.

### Upgrade Safety
- ECK operator upgrade (3.3.2 → this snapshot): existing clusters stay healthy, no unexpected restarts; new field becomes usable after CRD upgrade.
- CRD mismatch: without CRD upgrade, `nodeSets[].resources` is silently pruned by the API server (known k8s behavior, documented).
- Operator rollback (snapshot → 3.3.2) with shorthand set: old operator ignores the unknown field, StatefulSet keeps its last-set resources, no crash.

### Validation Warnings
- Warning fires exactly once per NodeSet when shorthand and podTemplate overlap on the same CPU or memory key.
- No warning when shorthand and podTemplate set disjoint keys (e.g. shorthand CPU + podTemplate memory).
- Warnings are per-NodeSet — offending index is identified in the message, clean NodeSets produce no warning.
- Warning is non-blocking: `kubectl apply` exits 0, the CR is persisted, shorthand wins at reconcile time.

### Autoscaling
- Autoscaler-computed resources flow through `WithResourcesAndOverrides` instead of being written directly into `podTemplate`, so a user-set shorthand is not clobbered by autoscaler output (updated `reconcile.go` + e2e `autoscaling_test.go`).

### Automated Testing
- Unit tests for `pkg/apis/common/v1alpha1`, `pkg/apis/common/v1`, `pkg/controller/common/defaults`, per-controller `resources_test.go` files for all eight CRDs, and `pkg/controller/elasticsearch/validation`.
- E2E tests in `test/e2e/es` covering autoscaling, topology, stackconfigpolicy, forced-upgrade, reversal, mutation, and volume scenarios with the new field exercised.

</details>

## Usage Example

Elasticsearch:

```yaml
kind: Elasticsearch
metadata:
  name: my-cluster
spec:
  version: 9.4.0
  nodeSets:
  - name: default
    count: 3
    resources:
      requests: { memory: 8Gi, cpu: 4 }
      limits:   { memory: 8Gi, cpu: 4 }
```

Kibana (and all other non-ES CRDs — `spec.resources` at the top level):

```yaml
kind: Kibana
metadata:
  name: my-kibana
spec:
  version: 9.4.0
  count: 1
  elasticsearchRef: { name: my-cluster }
  resources:
    requests: { memory: 1Gi, cpu: 500m }
```

No migration required for existing manifests — the new field is opt-in.


prodsecmachine:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/db22891f-05ee-4cc1-a351-7affc9f6d168?refs=336c726c6bcf6788ed6f8e4471a93ba702250fef&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/db22891f-05ee-4cc1-a351-7affc9f6d168/license?refs=336c726c6bcf6788ed6f8e4471a93ba702250fef&source=prChecksComment) |

naemono:
buildkite test this -f p=gke,s=9.3.2,t=TestAutoscaling

pkg/controller/autoscaling/elasticsearch/reconcile_test.go
```diff
@@ -0,0 +1,373 @@
...
+	assert.Equal(t, before.Spec, es.Spec)
+}
+
+// getMainContainer returns the Elasticsearch main container from a NodeSet pod template.
+func getMainContainer(nodeSet esv1.NodeSet) *corev1.Container {
+	for i := range nodeSet.PodTemplate.Spec.Containers {
+		container := nodeSet.PodTemplate.Spec.Containers[i]
+		if container.Name == esv1.ElasticsearchContainerName {
+			return &container
+		}
```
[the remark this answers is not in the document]
naemono: This is a test helper; not a concern.

pkg/controller/autoscaling/elasticsearch/testdata/custom_resource/ml/elasticsearch-expected.yml
```diff
@@ -29,17 +29,23 @@ spec:
         store.allow_mmap: false
     count: 1
     name: ml
+    resources:
+      limits:
+        cpu: "2"
+        memory: "4Gi"
+      requests:
+        cpu: "2"
+        memory: "4Gi"
```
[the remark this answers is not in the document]
naemono: Testdata; no concern.

pkg/apis/common/v1/resources_warnings.go
```diff
@@ -0,0 +1,44 @@
...
+		}
+	}
+	if main == nil {
+		return ""
+	}
+	// define a small closure here to help with readability
+	has := func(list corev1.ResourceList, name corev1.ResourceName) bool {
+		_, ok := list[name]
+		return ok
+	}
```
pkoutsovasilis, line 35: nit: I would grasp this opportunity to "modernise" a bit the pkg/utils/maps/maps.go package.

```patch
 // ContainsKeys determines if a set of label (keys) are present in a map of labels (keys and values).
-func ContainsKeys(m map[string]string, labels ...string) bool {
+func ContainsKeys[K comparable, V any](m map[K]V, labels ...K) bool {
 	for _, label := range labels {
 		if _, exists := m[label]; !exists {
 			return false
```

and I would use maps.ContainsKeys instead of `has` here

pkg/apis/common/v1/resources_warnings.go
```diff
@@ -0,0 +1,44 @@
...
+
+// PodTemplateResourcesOverrideWarning returns an admission warning when shorthand
+// CPU/memory values overlap with resources set on the main container in pt.
+// Returns "" when the main container is missing or there is no overlap.
+//
+// This check intentionally covers only the main container; init containers and
+// sidecars are not evaluated because the shorthand resources field only targets
+// the main container's CPU and memory.
+func PodTemplateResourcesOverrideWarning(shortPath, templatePath, mainContainer string, shorthand Resources, pt corev1.PodTemplateSpec) string {
+	var main *corev1.Container
```
pkoutsovasilis, line 21: nit: feel free to ignore but the var name `main` doesn't sit well with me 
```suggestion
func PodTemplateResourcesOverrideWarning(shortPath, templatePath, mainContainerName string, shorthand Resources, pt corev1.PodTemplateSpec) string {
	var mainContainer *corev1.Container
```

pkg/controller/autoscaling/elasticsearch/reconcile_test.go
```diff
@@ -0,0 +1,373 @@
...
+	}
+	require.NotNil(t, current)
+	assert.True(t, resource.MustParse(expected).Equal(*current))
+}
+
+// quantityPtr returns a pointer to a copy of the provided quantity value.
+func quantityPtr(quantity resource.Quantity) *resource.Quantity {
+	q := quantity
+	return &q
+}
```
pkoutsovasilis, line 373: this is redundant you can use `ptr.To` which is also consistent with the rest of the code base

pkg/controller/autoscaling/elasticsearch/resources/resources_test.go
```diff
@@ -217,11 +323,27 @@ func (nsb *nodeSetBuilder) withStorageRequest(qs string) *nodeSetBuilder {
...
+
 func (nsb *nodeSetBuilder) build() esv1.NodeSet {
 	nodeSet := esv1.NodeSet{
 		Name:   nsb.name,
 		Config: nil,
 		Count:  nsb.count,
+		Resources: commonv1.Resources{
+			Requests: commonv1.ResourceAllocations{},
+			Limits:   commonv1.ResourceAllocations{},
+		},
```
pkoutsovasilis, line 346: nit: These are empty-but-non-nil values, while the K8s API server normalises empty maps to nil on storage. The distinction doesn't affect the current tests (since `Match` checks individual keys, not `DeepEqual`), but could trip up a future test that compares against an API-server-round-tripped object. Consider initialising these only when entries are actually added - e.g., create the ResourceList inside the `if nsb.podTemplateMemoryRequest != nil` / `if nsb.podTemplateCPURequest != nil` branches below.

pkoutsovasilis (dismissed):
LGTM! The merge logic is clean and correctly centralised, the autoscaling integration is solid, and the test coverage is thorough. Left a couple of minor nits, nothing blocking.

pkg/controller/autoscaling/elasticsearch/resources/resources.go
```diff
@@ -42,15 +42,12 @@ func Match(ntr v1alpha1.NodeSetsResources, containerName string, nodeSet esv1.No
...
-		container := getContainer(containerName, nodeSet.PodTemplate.Spec.Containers)
-		if container == nil {
-			return false, nil
-		}
-		return ResourceEqual(corev1.ResourceMemory, ntr.NodeResources.Requests, container.Resources.Requests) &&
-			ResourceEqual(corev1.ResourceCPU, ntr.NodeResources.Requests, container.Resources.Requests) &&
-			ResourceEqual(corev1.ResourceMemory, ntr.NodeResources.Limits, container.Resources.Limits) &&
-			ResourceEqual(corev1.ResourceCPU, ntr.NodeResources.Limits, container.Resources.Limits), nil
+		currentRequests := resourceAllocationsToResourceList(nodeSet.Resources.Requests)
+		currentLimits := resourceAllocationsToResourceList(nodeSet.Resources.Limits)
```
barkbay: Could you please explain the migration path for an already-autoscaled cluster? If the source of truth changed, what happens to resources set at the PodTemplate level?

pkg/controller/autoscaling/elasticsearch/resources/resources.go
```diff
@@ -77,13 +74,19 @@ func ResourceEqual(resourceName corev1.ResourceName, expected, current corev1.Re
...
+func resourceAllocationsToResourceList(allocations commonv1.ResourceAllocations) corev1.ResourceList {
+	if allocations.CPU == nil && allocations.Memory == nil {
+		return nil
+	}
+	resources := make(corev1.ResourceList, 2)
+	if allocations.CPU != nil {
+		resources[corev1.ResourceCPU] = *allocations.CPU
+	}
+	if allocations.Memory != nil {
+		resources[corev1.ResourceMemory] = *allocations.Memory
```
barkbay: nit, future-proofs against any caller that decides to mutate the result, and the cost is negligible (one extra shallow `*inf.Dec` allocation only when the decimal form is in use, which is rare for CPU/memory quantities).

```suggestion
		resources[corev1.ResourceCPU] = allocations.CPU.DeepCopy()
	}
	if allocations.Memory != nil {
		resources[corev1.ResourceMemory] = allocations.Memory.DeepCopy()
```

barkbay (commented):
The PR adds the shorthand to the CRD schema and API reference, but none of the user-facing examples are updated to use (or even mention) it:

1. `config/recipes/**` still use `podTemplate.spec.containers[].resources`.
2. `deploy/eck-stack/charts/*/values.yaml` and `deploy/eck-stack/charts/*/examples/*.yaml` do the same.

Given the stated goal of the PR is to simplify the user experience, maybe it's worth updating some recipes or Helm examples using `spec.resources` (or per-NodeSet `resources` for ES) in this PR, so the feature is actually discoverable in this repo?

pkg/apis/agent/v1alpha1/webhook.go
```diff
@@ -33,10 +36,30 @@ func (a *Agent) warnings() []string {
...
+	if pt, path := a.activePodTemplate(); path != "" {
+		if w := commonv1.PodTemplateResourcesOverrideWarning("spec.resources", path, AgentContainerName, a.Spec.Resources, pt); w != "" {
+			warnings = append(warnings, w)
+		}
+	}
+	return warnings
+}
+
+// activePodTemplate returns the configured pod template and its spec path.
+// checkSpec ensures at most one deployment mode is set; returns ("", empty) when none.
```
barkbay: nit, not sure I understand the `returns ("", empty) when none`
```suggestion
// checkSpec ensures at most one deployment mode is set.
// If no deployment mode (DaemonSet, Deployment, or StatefulSet) is set, it returns an empty PodTemplateSpec and an empty path.
```

pkg/controller/autoscaling/elasticsearch/resources/resources.go
naemono: This is an area I didn't test as well as I should have. Upon testing this morning, I see there's room for improvement, as the podTemplate values are left in place, causing constant validation warnings in the logs, and events. Let me work through handling this more cleanly and update this.

naemono: I've updated this logic. The podTemplate resources are removed during the migration now.

pkg/controller/common/defaults/pod_template.go
```diff
@@ -389,6 +390,72 @@ func (b *PodTemplateBuilder) WithInitContainers(
...
+		merged = *resources.DeepCopy()
+		merged.Claims = claims
+	}
+	applyCPUAndMemoryOverrides(&merged.Limits, overrides.Limits)
+	applyCPUAndMemoryOverrides(&merged.Requests, overrides.Requests)
+
+	main.Resources = merged
+	b.setContainerDefaulter()
+	return b
+}
```
barkbay, line 428: `resourceRequirementsMergeBase` rebuilds `corev1.ResourceRequirements` from an empty struct and re-copies `Claims` by name. Any field added to `ResourceRequirements` upstream (as happened with `Claims` in 1.26) would be silently dropped here for users who set it in `podTemplate`.
We can start from a deep copy of the existing struct and only overwrite `Requests`/`Limits`. The helper and its returned bool go away:


```suggestion
// WithResourcesAndOverrides merges main-container resources from three sources:
//   - main container resources from the pod template (merge base; nil-vs-empty map shape for
//     Requests/Limits is preserved, including explicit empty maps for
//     [LimitRange](https://kubernetes.io/docs/concepts/policy/limit-range))
//   - resources: operator default ResourceRequirements (used for Requests and Limits only, and
//     only when the pod template sets neither; all other fields of the pod template's
//     ResourceRequirements, e.g. ResourceClaims, are preserved as-is)
//   - overrides: CRD spec.resources CPU/memory values (applied only for non-nil override pointers)
//
// If the main container does not exist, this method returns the builder unchanged; that should not
// happen if NewPodTemplateBuilder is called prior to this method.
func (b *PodTemplateBuilder) WithResourcesAndOverrides(resources corev1.ResourceRequirements, overrides commonv1.Resources) *PodTemplateBuilder {
	main := b.MainContainer()
	if main == nil {
		return b
	}
	merged := *main.Resources.DeepCopy()
	// When the pod template contributed no CPU/memory base, fall back to the operator defaults for
	// Requests/Limits only. Every other field of ResourceRequirements (ResourceClaims today, any
	// future additions upstream) stays as set by the user via the deep copy above.
	if merged.Requests == nil && merged.Limits == nil {
		defaults := resources.DeepCopy()
		merged.Requests = defaults.Requests
		merged.Limits = defaults.Limits
	}
	applyCPUAndMemoryOverrides(&merged.Limits, overrides.Limits)
	applyCPUAndMemoryOverrides(&merged.Requests, overrides.Requests)

	main.Resources = merged
	b.setContainerDefaulter()
	return b
}
```

barkbay:
For the record I tested the webhook, seems to work as expected 👍 
```
Warning: spec.nodeSets[0].resources overrides CPU/memory set in spec.nodeSets[0].podTemplate.spec.containers[] for container "elasticsearch"; remove one source of truth
Warning: spec.nodeSets[1].resources overrides CPU/memory set in spec.nodeSets[1].podTemplate.spec.containers[] for container "elasticsearch"; remove one source of truth
Warning: spec.nodeSets[2].resources overrides CPU/memory set in spec.nodeSets[2].podTemplate.spec.containers[] for container "elasticsearch"; remove one source of truth
elasticsearch.elasticsearch.k8s.elastic.co/es-sample configured
```

barkbay (commented):
Two pod-owning CRDs appear to be missing the new `resources` shorthand:

* `ElasticPackageRegistry` (`pkg/apis/packageregistry/v1alpha1`)
* `AutoOpsAgentPolicy` (`pkg/apis/autoops/v1alpha1`)

Both expose a `spec.podTemplate` of type `corev1.PodTemplateSpec` and their controllers already build pods via `defaults.NewPodTemplateBuilder`, so plugging in `WithResourcesAndOverrides` plus the validating warning would be a one liner each.

Was this deliberate (e.g. scope, single container assumption, internal only components) or just an oversight?

barkbay (commented):
Small behavioural asymmetry between `podTemplate` and the new `resources` shorthand worth flagging.

When a user sets only `limits` through `podTemplate`, the API server's builtin `limit → request` defaulting kicks in at Pod creation time, so the container ends up with `requests == limits` and `Guaranteed` QoS.

With the shorthand, `WithResourcesAndOverrides` takes a different path: if `main.Resources` on the `podTemplate` is empty, operator defaults are written into `merged` before the shorthand overrides are layered on top. That means `requests.memory` gets the operator default (e.g. `2Gi` for Elasticsearch) and the shorthand only writes to `limits`, so the final container is `limits={cpu:X, memory:Y}, requests={memory:2Gi}` and QoS is `Burstable`.

Repro on a clean `Elasticsearch` with just:

```yaml
resources:
  limits:
    cpu: 1
    memory: 3Gi
```

→ StatefulSet template carries `limits={cpu:1, memory:3Gi}, requests={memory:2Gi}`. But same CR rewritten under `podTemplate` with `limits` only → `limits={cpu:1, memory:3Gi}, requests=nil`, API server copies it, Pod is `Guaranteed`.

Two consequences:

1. Users migrating from `podTemplate` to the shorthand with the intent of a pure simplification can silently lose `Guaranteed` QoS.
2. Documenting the "set only `limits` to get the best practice memory config" pattern doesn't work with this shorthand.

Fix options I see:

1. In `WithResourcesAndOverrides`, widen the "use operator defaults" guard to also consider the shorthand: only inject defaults when `podTemplate` and shorthand both contribute nothing. Otherwise layer the shorthand directly on top of `podTemplate` (empty when absent) and leave untouched sides nil so the API server can default them.
2. Call this out loudly in the field godoc and release notes so users know to always set both `requests` and `limits` via the shorthand if they want `Guaranteed`.

naemono:
> Two pod-owning CRDs appear to be missing the new `resources` shorthand:
> 
> * `ElasticPackageRegistry` (`pkg/apis/packageregistry/v1alpha1`)
> * `AutoOpsAgentPolicy` (`pkg/apis/autoops/v1alpha1`)
> 
> Both expose a `spec.podTemplate` of type `corev1.PodTemplateSpec` and their controllers already build pods via `defaults.NewPodTemplateBuilder`, so plugging in `WithResourcesAndOverrides` plus the validating warning would be a one liner each.
> 
> Was this deliberate (e.g. scope, single container assumption, internal only components) or just an oversight?

It was just an oversight. These (and the corresponding webhook validation warnings) have been added, in addition to tests.

naemono:
> Small behavioural asymmetry between `podTemplate` and the new `resources` shorthand worth flagging.
> 
> When a user sets only `limits` through `podTemplate`, the API server's builtin `limit → request` defaulting kicks in at Pod creation time, so the container ends up with `requests == limits` and `Guaranteed` QoS.
> 
> With the shorthand, `WithResourcesAndOverrides` takes a different path: if `main.Resources` on the `podTemplate` is empty, operator defaults are written into `merged` before the shorthand overrides are layered on top. That means `requests.memory` gets the operator default (e.g. `2Gi` for Elasticsearch) and the shorthand only writes to `limits`, so the final container is `limits={cpu:X, memory:Y}, requests={memory:2Gi}` and QoS is `Burstable`.
> 
> Repro on a clean `Elasticsearch` with just:
> 
> ```yaml
> resources:
>   limits:
>     cpu: 1
>     memory: 3Gi
> ```
> 
> → StatefulSet template carries `limits={cpu:1, memory:3Gi}, requests={memory:2Gi}`. But same CR rewritten under `podTemplate` with `limits` only → `limits={cpu:1, memory:3Gi}, requests=nil`, API server copies it, Pod is `Guaranteed`.
> 
> Two consequences:
> 
> 1. Users migrating from `podTemplate` to the shorthand with the intent of a pure simplification can silently lose `Guaranteed` QoS.
> 2. Documenting the "set only `limits` to get the best practice memory config" pattern doesn't work with this shorthand.
> 
> Fix options I see:
> 
> 1. In `WithResourcesAndOverrides`, widen the "use operator defaults" guard to also consider the shorthand: only inject defaults when `podTemplate` and shorthand both contribute nothing. Otherwise layer the shorthand directly on top of `podTemplate` (empty when absent) and leave untouched sides nil so the API server can default them.
> 2. Call this out loudly in the field godoc and release notes so users know to always set both `requests` and `limits` via the shorthand if they want `Guaranteed`.

Really great edge case caught here. It really felt like more of a bug that something we want to document, and I've handled it as such (option 1).

naemono:
buildkite test this -f p=gke,s=9.3.2,t=TestAutoscaling

barkbay:
Almost LGTM, testing the autoscaler, will complete my review today.

barkbay (commented):
While testing the autoscaler upgrade path (3.3.2 → this PR), I noticed that NodeSets that are **not** managed by the autoscaler (in my case a dedicated `master` NodeSet using the legacy `podTemplate.spec.containers[].resources`) end up with an empty `spec.nodeSets[i].resources: { limits: {}, requests: {} }` block in the stored CR after the first reconcile, even though the user never set the shorthand and the autoscaler does not touch that NodeSet.

```yaml
nodeSets:
  name: master
  resources:
    limits: {}
    requests: {}
  podTemplate:
    spec:
      containers:
        name: elasticsearch
        resources:
          limits:   { cpu: 1, memory: 2Gi }
          requests: { cpu: 1, memory: 2Gi }
```

Functionally this is harmless: `Resources.IsEmpty()` returns true and `WithResourcesAndOverrides` skips the empty allocations, so the effective container resources and Pod QoS are unchanged. It is a CR-shape side effect of `commonv1.Resources` being a value-type struct: Go's `encoding/json` does not honor `omitempty` on non-pointer struct fields, so when the operator updates a sibling NodeSet (autoscaler write on `di`/`ml`) and the CR is round-tripped, master's zero-valued `Resources` is persisted as `{ limits: {}, requests: {} }` instead of being omitted.

It can also be confusing for the user: within a single CR we now have two different ways of expressing container resources side by side (the new shorthand on the autoscaled NodeSets, the legacy `podTemplate` block on master), **plus** the empty shorthand stub on the non-autoscaled NodeSets. When inspecting the CR after upgrade it looks like the master NodeSet is partially using the new field, while nothing was ever set there. Dropping the empty block (via either of the follow-ups below) would also make the "one source of truth" story easier to explain.

pkg/apis/common/v1/resources.go
```diff
@@ -0,0 +1,40 @@
...
+	CPU *resource.Quantity `json:"cpu,omitempty"`
+	// Memory overrides the main container's memory request/limit when the parent
+	// Resources is merged into a PodTemplate. A nil value means "do not override":
+	// any memory value already set on the main container in the PodTemplate is
+	// passed through unchanged. Setting this field to nil does not unset a memory
+	// value present in the PodTemplate; to remove it, edit the PodTemplate's
+	// container resources. Setting this field to the literal "0" sets the override
+	// to a zero quantity and does not clear the PodTemplate's memory.
+	Memory *resource.Quantity `json:"memory,omitempty"`
+}
```
barkbay, line 66: Feels like there is a bit of duplicated code between `resourceAllocationsToResourceList` and `applyCPUAndMemoryOverrides`: both walk the same two `*resource.Quantity` fields and produce a `corev1.ResourceList` keyed by CPU/memory, just with slightly different "write" semantics.

I wonder if we should introduce a `ToResourceList()` method on `commonv1.ResourceAllocations` in `pkg/apis/common/v1/resources.go` and have both call sites use it:

```suggestion
}

// ToResourceList converts the CPU/memory allocations into a corev1.ResourceList.
// Returns nil when both CPU and memory are unset so callers can treat absent
// allocations as "no value". Quantities are deep-copied so the returned map
// shares no internal state with the original ResourceAllocations.
func (ra *ResourceAllocations) ToResourceList() corev1.ResourceList {
	if ra == nil || (ra.CPU == nil && ra.Memory == nil) {
		return nil
	}
	out := make(corev1.ResourceList, 2)
	if ra.CPU != nil {
		out[corev1.ResourceCPU] = ra.CPU.DeepCopy()
	}
	if ra.Memory != nil {
		out[corev1.ResourceMemory] = ra.Memory.DeepCopy()
	}
	return out
}

```

Then:

```diff
diff --git a/pkg/controller/autoscaling/elasticsearch/resources/resources.go b/pkg/controller/autoscaling/elasticsearch/resources/resources.go
index c510cc2a2..3d36d0b21 100644
--- a/pkg/controller/autoscaling/elasticsearch/resources/resources.go
+++ b/pkg/controller/autoscaling/elasticsearch/resources/resources.go
@@ -7,7 +7,6 @@ package resources
 import (
        "fmt"

-       commonv1 "github.com/elastic/cloud-on-k8s/v3/pkg/apis/common/v1"
        "github.com/elastic/cloud-on-k8s/v3/pkg/apis/common/v1alpha1"

        corev1 "k8s.io/api/core/v1"
@@ -42,8 +41,8 @@ func Match(ntr v1alpha1.NodeSetsResources, nodeSet esv1.NodeSet) (bool, error) {
                        return false, fmt.Errorf("only 1 volume claim template is allowed when autoscaling is enabled, got %d in nodeSet %s", len(nodeSet.VolumeClaimTemplates), nodeSet.Name)
                }

-               currentRequests := resourceAllocationsToResourceList(nodeSet.Resources.Requests)
-               currentLimits := resourceAllocationsToResourceList(nodeSet.Resources.Limits)
+               currentRequests := nodeSet.Resources.Requests.ToResourceList()
+               currentLimits := nodeSet.Resources.Limits.ToResourceList()
                return ResourceEqual(corev1.ResourceMemory, ntr.NodeResources.Requests, currentRequests) &&
                        ResourceEqual(corev1.ResourceCPU, ntr.NodeResources.Requests, currentRequests) &&
                        ResourceEqual(corev1.ResourceMemory, ntr.NodeResources.Limits, currentLimits) &&
@@ -74,19 +73,3 @@ func ResourceEqual(resourceName corev1.ResourceName, expected, current corev1.Re
        return expectedValue.Equal(currentValue)
 }

-// resourceAllocationsToResourceList converts NodeSet shorthand CPU/memory allocations into a ResourceList.
-// It intentionally returns nil when both CPU and memory are unset so ResourceEqual can treat
-// absent shorthand allocations as "no expected value".
-func resourceAllocationsToResourceList(allocations commonv1.ResourceAllocations) corev1.ResourceList {
-       if allocations.CPU == nil && allocations.Memory == nil {
-               return nil
-       }
-       resources := make(corev1.ResourceList, 2)
-       if allocations.CPU != nil {
-               resources[corev1.ResourceCPU] = *allocations.CPU
-       }
-       if allocations.Memory != nil {
-               resources[corev1.ResourceMemory] = *allocations.Memory
-       }
-       return resources
-}
diff --git a/pkg/controller/common/defaults/pod_template.go b/pkg/controller/common/defaults/pod_template.go
index 9d287c2e5..a20e50516 100644
--- a/pkg/controller/common/defaults/pod_template.go
+++ b/pkg/controller/common/defaults/pod_template.go
@@ -430,17 +430,15 @@ func (b *PodTemplateBuilder) WithResourcesAndOverrides(resources corev1.Resource
 // If a write is required and dst is nil, the destination map is initialized.
 // With no override values, dst is left unchanged.
 func applyCPUAndMemoryOverrides(dst *corev1.ResourceList, overrides commonv1.ResourceAllocations) {
-       if overrides.CPU == nil && overrides.Memory == nil {
+       src := overrides.ToResourceList()
+       if src == nil {
                return
        }
        if *dst == nil {
                *dst = corev1.ResourceList{}
        }
-       if overrides.CPU != nil {
-               (*dst)[corev1.ResourceCPU] = *overrides.CPU
-       }
-       if overrides.Memory != nil {
-               (*dst)[corev1.ResourceMemory] = *overrides.Memory
+       for name, quantity := range src {
+               (*dst)[name] = quantity
        }
 }
```

pkg/apis/common/v1/resources.go
```diff
@@ -0,0 +1,40 @@
...
+// or more contributor license agreements. Licensed under the Elastic License 2.0;
+// you may not use this file except in compliance with the Elastic License 2.0.
+
+package v1
+
+import (
+	"k8s.io/apimachinery/pkg/api/resource"
+)
+
+type Resources struct {
```
barkbay, line 20:
```suggestion
// Resources is a shorthand for setting CPU and memory requests and limits on
// the main application container of a pod-owning CRD. It is a simplified
// alternative to the nested podTemplate.spec.containers[name=<main>].resources
// path: any non-nil CPU or memory value in Requests or Limits is written to the
// main container's resources at reconcile time, overriding any value already
// present in the PodTemplate for the same key. Non-CPU/memory resource keys
// (for example ephemeral-storage) and all other container fields set via
// PodTemplate are preserved as-is.
type Resources struct {
```

pkg/apis/common/v1/resources.go
```diff
@@ -0,0 +1,40 @@
...
+}
+
+// IsEmpty reports whether the shorthand contains no CPU or memory values on
+// either Requests or Limits.
+func (r Resources) IsEmpty() bool {
+	return r.Limits.CPU == nil && r.Limits.Memory == nil &&
+		r.Requests.CPU == nil && r.Requests.Memory == nil
+}
+
+type ResourceAllocations struct {
```
barkbay, line 50:
```suggestion
// ResourceAllocations holds optional CPU and memory quantities that the
// shorthand Resources field applies to the main container's Requests or Limits.
// Using pointers lets callers distinguish "no override" (nil) from "override
// with a zero quantity" (non-nil pointing to a zero value).
type ResourceAllocations struct {
```

pkg/apis/common/v1/resources_warnings.go
```diff
@@ -0,0 +1,41 @@
...
+		return ""
+	}
+	overlap := (shorthand.Requests.CPU != nil && maps.ContainsKeys(main.Resources.Requests, corev1.ResourceCPU)) ||
+		(shorthand.Requests.Memory != nil && maps.ContainsKeys(main.Resources.Requests, corev1.ResourceMemory)) ||
+		(shorthand.Limits.CPU != nil && maps.ContainsKeys(main.Resources.Limits, corev1.ResourceCPU)) ||
+		(shorthand.Limits.Memory != nil && maps.ContainsKeys(main.Resources.Limits, corev1.ResourceMemory))
+	if !overlap {
+		return ""
+	}
+	return fmt.Sprintf("%s overrides CPU/memory set in %s.spec.containers[] for container %q; remove one source of truth", shortPath, templatePath, mainContainer)
```
barkbay: nit: `containers[]` suffix is neither valid JSONPath nor particularly intuitive, should we use an indexed filter (`spec.podTemplate.spec.containers[name="kibana"]`, which makes the container explicit and removes the separate `%q` arg?
```suggestion
	return fmt.Sprintf("%s overrides CPU/memory set in %s.spec.containers[name=%q]; remove one source of truth", shortPath, templatePath, mainContainer)
```

pkg/apis/elasticsearch/v1/elasticsearch_types.go
```diff
@@ -385,6 +385,10 @@ type NodeSet struct {
 	// +kubebuilder:validation:Optional
 	Count int32 `json:"count"`
 
+	// Resources specifies the resource requests and limits (CPU and Memory only) for the Elasticsearch nodes in this NodeSet. When set these override the resource requests and limits set in the PodTemplate for the primary Elasticsearch container. To set the resources for other containers, use the PodTemplate.Spec.Containers[].Resources field.
```
barkbay: nit: ES godoc is missing a comma where every other CRD has one.

```suggestion
	// Resources specifies the resource requests and limits (CPU and Memory only) for the Elasticsearch nodes in this NodeSet. When set, these override the resource requests and limits set in the PodTemplate for the primary Elasticsearch container. To set the resources for other containers, use the PodTemplate.Spec.Containers[].Resources field.
```

pkg/apis/beat/v1beta1/beat_types.go
```diff
@@ -71,6 +71,12 @@ type BeatSpec struct {
 	// +kubebuilder:validation:Optional
 	ServiceAccountName string `json:"serviceAccountName,omitempty"`
 
+	// Resources provides a shorthand to set CPU and Memory resources on the main Beat container. When set, these
+	// values override any CPU or memory resource settings specified in the DaemonSet or Deployment PodTemplate for
+	// the primary Beat container. To set resources on other containers, use the PodTemplate.
```
barkbay: nit: godoc says "main Beat container" while every other CRD says "primary X container". Pick one for consistency.

pkg/apis/common/v1/resources.go
```diff
@@ -0,0 +1,40 @@
...
+	// CPU overrides the main container's CPU request/limit when the parent Resources
+	// is merged into a PodTemplate. A nil value means "do not override": any CPU
+	// value already set on the main container in the PodTemplate is passed through
+	// unchanged. Setting this field to nil does not unset a CPU value present in
+	// the PodTemplate; to remove it, edit the PodTemplate's container resources.
+	// Setting this field to the literal "0" sets the override to a zero quantity
+	// and does not clear the PodTemplate's CPU.
```
barkbay: Not sure the "does not clear the PodTemplate" is clear, but I don't think I have a better alternative, maybe:
```suggestion
	// A non-nil value wins over the PodTemplate's CPU, including an explicit zero quantity.
```

barkbay (approved):
LGTM, left a few additional nits.

naemono:
@barkbay would you mind one more look in the morning? I went with `omitZero` to handle the noted issue about an empty `resources` field, and things seem to work for me. I wanted to see if you had any further comments on the approach. Thanks!

barkbay:
> @barkbay would you mind one more look in the morning?

Will do

config/crds/v1/resources/agent.k8s.elastic.co_agents.yaml
```diff
@@ -17885,6 +17885,88 @@ spec:
...
+                          passed through unchanged. Setting this field to nil does not unset a memory
+                          value present in the PodTemplate; to remove it, edit the PodTemplate's
+                          container resources.
+                          A non-nil value wins over the PodTemplate's memory, including an explicit zero quantity.
+                        pattern: ^(\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))(([KMGTPE]i)|[numkMGTPE]|([eE](\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))))?$
+                        x-kubernetes-int-or-string: true
+                    type: object
+                required:
+                - limits
+                - requests
```
barkbay: I think `limits` and `requests` should not be required? (see my other comment)

pkg/apis/common/v1/resources.go
```diff
@@ -0,0 +1,76 @@
...
+type Resources struct {
+	// Use omitzero so a zero-valued Resources (no CPU or memory set) does not
+	// serialize as `{"limits":{},"requests":{}}`. encoding/json's omitempty does
+	// not treat a non-pointer struct value as empty, even when all of its fields
+	// are nil/zero, so an unset shorthand on one NodeSet would otherwise be
+	// persisted as an empty stub when the operator round-trips the CR after
+	// updating a sibling NodeSet (for example via the autoscaler).
+	Limits   ResourceAllocations `json:"limits,omitzero"`
+	Requests ResourceAllocations `json:"requests,omitzero"`
+}
```
barkbay, line 37: `omitzero` on `Resources.Limits` / `Resources.Requests` (+ on the top-level `resources` field for each CRD) fixes the empty-stub issue nicely, but controller-tools infers "optional" from `omitempty` or from an explicit `+optional` / `+kubebuilder:validation:Optional` marker — not from `omitzero`. With `omitempty` dropped and no marker, the generated schemas now mark both sub-fields as required, which rejects two common shapes users naturally write with the shorthand:

- requests only (Burstable QoS, floor without cap)
- limits only (cap without reservation)

Repro against the PR's generated Kibana CRD (same effect on the other 9 CRDs):

```console
$ cat <<'EOF' | kubectl apply --dry-run=server -f -
kind: Kibana
metadata:
  name: kb-requests-only
spec:
  version: 9.2.0
  count: 1
  elasticsearchRef:
    name: none
  resources:
    requests:
      memory: 1Gi
      cpu: "500m"
EOF
The Kibana "kb-requests-only" is invalid: spec.resources.limits: Required value
```

```suggestion
type Resources struct {
	// The omitzero JSON tag below is used (instead of omitempty) because
	// encoding/json's omitempty does not treat a non-pointer struct value as
	// empty even when all of its fields are nil. Without omitzero, an unset
	// shorthand on one NodeSet would be persisted as
	// `{"limits":{},"requests":{}}` when the operator round-trips the CR after
	// updating a sibling NodeSet (for example via the autoscaler). Both fields
	// must also carry +kubebuilder:validation:Optional because controller-gen
	// infers "optional" from omitempty (or an explicit marker), not from
	// omitzero.

	// Limits is the shorthand for the main container's CPU and memory limits.
	// +kubebuilder:validation:Optional
	Limits ResourceAllocations `json:"limits,omitzero"`
	// Requests is the shorthand for the main container's CPU and memory requests.
	// +kubebuilder:validation:Optional
	Requests ResourceAllocations `json:"requests,omitzero"`
}
```

PR #9348, opened 2026-04-14
Add Mergify configuration

moukoublen:
https://docs.mergify.com/configuration/file-format/

Summary
  - Introduce .github/mergify.yml to automate common PR workflows via Mergify

What it does
  - Auto-assign PRs to their author when no assignee is set (irrelevant to backport but useful 😅).
  - Conflict notification: comments on PRs with merge conflicts asking the author to resolve them
  - Backport automation: label-driven backports to 3.3 and 3.4 release branches
    - Adding label `backport-v3.3` will open a backport pr to `3.3` branch (after merge).
    - Adding label `backport-v3.4` will open a backport pr to `3.4` branch (after merge).
  - Forward-port to main: PRs merged in release branches can be forwarded to main via the forwardport-main label
  - Restricts the backport command to users with write permission and Renovate bot

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -1,0 +1,91 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/Mergifyio/docs/main/public/mergify-configuration-schema.json
+# https://docs.mergify.com/configuration/file-format/
+
+commands_restrictions:
+  backport:
+    conditions:
+      - or:
+          - sender-permission>=write
+          - sender=elastic-renovate-prod[bot]
+
+defaults:
+  actions:
+    backport:
+      title: "[{{ destination_branch }}] (backport #{{ number }}) {{ title }}"
+      assignees:
+        - "{{ author }}"
+      labels:
+        - "backport"
+
+pull_request_rules:
+  # Automatically assigns the PR author as the assignee when no one is assigned yet.
+  - name: self-assign PRs
+    conditions:
+      - -merged
+      - -closed
+      - "#assignee=0"
+    actions:
+      assign:
+        add_users:
+          - "{{ author }}"
+
+  # Comments on PRs that have merge conflicts, asking the author to resolve them.
+  - name: ask to resolve conflict
+    conditions:
+      - -merged
+      - -closed
+      - conflict
+    actions:
+      comment:
+        message: |
+          This pull request is now in conflicts. Could you fix it? 🙏
+          To fixup this pull request, you can check out it locally. See documentation: https://help.github.com/articles/checking-out-pull-requests-locally/
+          ```
+          git fetch upstream
+          git checkout -b {{head}} upstream/{{head}}
+          git merge upstream/{{base}}
+          git push upstream {{head}}
+          ```
+
+  # Forward-ports merged PRs labeled "forwardport-main" to the main branch via backport. In case we merge something in a release branch first, and then we need it in main branch.
+  - name: forward-port patches to main branch
+    conditions:
+      - merged
+      - label=forwardport-main
+    actions:
+      backport:
+        assignees:
+          - "{{ author }}"
+        branches:
+          - "main"
+        labels:
+          - "backport"
+        title: "[{{ destination_branch }}](backport #{{ number }}) {{ title }}"
+
+  # Backports merged PRs labeled "backport-v3.3" to the 3.3 release branch.
+  - name: backport patches to 3.3 branch
+    conditions:
+      - merged
+      - label=backport-v3.3
+    actions:
+      backport:
+        assignees:
+          - "{{ author }}"
+        branches:
+          - "3.3"
+        labels:
+          - "backport"
+
+  # Backports merged PRs labeled "backport-v3.4" to the 3.4 release branch.
+  - name: backport patches to 3.4 branch
+    conditions:
+      - merged
+      - label=backport-v3.4
+    actions:
+      backport:
+        assignees:
+          - "{{ author }}"
+        branches:
+          - "3.4"
+        labels:
+          - "backport"
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -18,17 +18,6 @@ defaults:
         - "backport"
 
 pull_request_rules:
-  # Automatically assigns the PR author as the assignee when no one is assigned yet.
-  - name: self-assign PRs
-    conditions:
-      - -merged
-      - -closed
-      - "#assignee=0"
-    actions:
-      assign:
-        add_users:
-          - "{{ author }}"
-
 
   # Forward-ports merged PRs labeled "forwardport-main" to the main branch via backport. In case we merge something in a release branch first, and then we need it in main branch.
   - name: forward-port patches to main branch
@@ -37,6 +26,7 @@ pull_request_rules:
       - label=forwardport-main
     actions:
       backport:
+        title: "[{{ destination_branch }}] (forward-port #{{ number }}) {{ title }}"
         branches: ["main"]
 
   # Backports merged PRs labeled "backport-v3.3" to the 3.3 release branch.
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -29,23 +29,6 @@ pull_request_rules:
         add_users:
           - "{{ author }}"
 
-  # Comments on PRs that have merge conflicts, asking the author to resolve them.
-  - name: ask to resolve conflict
-    conditions:
-      - -merged
-      - -closed
-      - conflict
-    actions:
-      comment:
-        message: |
-          This pull request is now in conflicts. Could you fix it? 🙏
-          To fixup this pull request, you can check out it locally. See documentation: https://help.github.com/articles/checking-out-pull-requests-locally/
-          ```
-          git fetch upstream
-          git checkout -b {{head}} upstream/{{head}}
-          git merge upstream/{{base}}
-          git push upstream {{head}}
-          ```
 
   # Forward-ports merged PRs labeled "forwardport-main" to the main branch via backport. In case we merge something in a release branch first, and then we need it in main branch.
   - name: forward-port patches to main branch
@@ -54,13 +37,7 @@ pull_request_rules:
       - label=forwardport-main
     actions:
       backport:
-        assignees:
-          - "{{ author }}"
-        branches:
-          - "main"
-        labels:
-          - "backport"
-        title: "[{{ destination_branch }}](backport #{{ number }}) {{ title }}"
+        branches: ["main"]
 
   # Backports merged PRs labeled "backport-v3.3" to the 3.3 release branch.
   - name: backport patches to 3.3 branch
@@ -69,12 +46,7 @@ pull_request_rules:
       - label=backport-v3.3
     actions:
       backport:
-        assignees:
-          - "{{ author }}"
-        branches:
-          - "3.3"
-        labels:
-          - "backport"
+        branches: ["3.3"]
 
   # Backports merged PRs labeled "backport-v3.4" to the 3.4 release branch.
   - name: backport patches to 3.4 branch
@@ -83,9 +55,4 @@ pull_request_rules:
       - label=backport-v3.4
     actions:
       backport:
-        assignees:
-          - "{{ author }}"
-        branches:
-          - "3.4"
-        labels:
-          - "backport"
+        branches: ["3.4"]


prodsecmachine:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/9c5c44bb-e21b-4424-8cc8-5ad640a64ceb?refs=8bfc8e9fe4e28a0970b114851287efa1a19a36d8&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/9c5c44bb-e21b-4424-8cc8-5ad640a64ceb/license?refs=8bfc8e9fe4e28a0970b114851287efa1a19a36d8&source=prChecksComment) |

.github/mergify.yml
```diff
@@ -0,0 +1,91 @@
...
+      - merged
+      - label=backport-v3.4
+    actions:
+      backport:
+        assignees:
+          - "{{ author }}"
+        branches:
+          - "3.4"
+        labels:
+          - "backport"
```
barkbay: Personally the existing backport flow has not been a big pain point for me. I am a bit concerned this adds another recurring task for the release managers, keeping Mergify labels/rules aligned with supported branches?

.github/mergify.yml
```diff
@@ -0,0 +1,91 @@
...
+      - -merged
+      - -closed
+      - "#assignee=0"
+    actions:
+      assign:
+        add_users:
+          - "{{ author }}"
+
+  # Comments on PRs that have merge conflicts, asking the author to resolve them.
+  - name: ask to resolve conflict
```
barkbay: Could you help me understand the purpose of this rule? GitHub already marks the PR as conflicted and prevents merging until it is resolved, so I am curious what extra value the automated comment adds?

.github/mergify.yml
```diff
@@ -0,0 +1,91 @@
...
+    backport:
+      title: "[{{ destination_branch }}] (backport #{{ number }}) {{ title }}"
+      assignees:
+        - "{{ author }}"
+      labels:
+        - "backport"
+
+pull_request_rules:
+  # Automatically assigns the PR author as the assignee when no one is assigned yet.
+  - name: self-assign PRs
```
barkbay: Similar question here: is there a concrete problem this solves for our workflow? I don't think we have relied on PR assignment in the past.

.github/mergify.yml
moukoublen: > keeping Mergify labels/rules aligned

We don't need to do any proactive work. We can create the label on demand the moment we need to backport something. Rules can be added automatically by some script (other teams use this approach as well). 

(For example instead of creating the release branch ourself, a single github action can create the branch, add the mergify rule to main in a new PR, and change what ever is needed inside the release branch. But this is another discussion about automating the release processes that we will have at some point.)

> Personally the existing backport flow has not been a big pain point for me.

I think we can all agree that backporting by simply adding a label to a PR is far easier than handling it manually. We had a discussion about it and we decided to give it a try.

.github/mergify.yml
moukoublen: This message is used in other mergify config throughout Elastic, so I thought of adding it as an extra help/indication for the author. We can remove it if we feel it's too much.

.github/mergify.yml
moukoublen: > I don't think we have relied on PR assignment in the past.

Right now, some people do use "put the author as assignee", and some do not. I think some teams use this rule in general, and some do not. To ensure consistency and because it doesn't introduce any harm, this rule unifies it without requiring any manual intervention. Do you think we should avoid that?

.github/mergify.yml
moukoublen: I removed that rule in  757c6bd

.github/mergify.yml
barkbay: > We don't need to do any proactive work. We can create the label on demand the moment we need to backport something. Rules can be added automatically by some script (other teams use this approach as well).

I think we would need to extend this for each new release (IIRC we almost always end up backporting something from `main` to the current `X.Y` branch). I am not opposed to that extra moving part, I am trying to understand which concrete problem it solves for us 🙂  I can see it being very useful when you need the same change on **many** release branches at once (for example all `9.x` branches), but that is uncommon for ECK, where we typically backport from `main` to **one** current release branch only.

I have not used Mergify enough to judge: next to a small CLI backport flow, I am not sure where the big win shows up for us. Happy to be corrected if there is an angle I am not seeing. Curious what others think either way, I am fine following the consensus.

moukoublen: > I have not used Mergify enough to judge: next to a small CLI backport flow, I am not sure where the big win shows up for us

I think the comparison is:  having every engineer set up (once) the cli with the appropriate access and then running the CLI manually every time they want to backport a PR versus adding a label in the PR itself, and then the packport PR just pops when the main PR is merged, doesn't seem like the second one is much easier, less time consuming, less friction?

.github/mergify.yml
pebrc: Let's maybe focus on the backports only for now. I don't think we have an established workflow for what "assigning" a PR means at the moment..

.github/mergify.yml
```diff
@@ -0,0 +1,58 @@
...
+        branches: ["3.3"]
+
+  # Backports merged PRs labeled "backport-v3.4" to the 3.4 release branch.
+  - name: backport patches to 3.4 branch
+    conditions:
+      - merged
+      - label=backport-v3.4
+    actions:
+      backport:
+        branches: ["3.4"]
```
pebrc, line 48: Could we collapse the three near-duplicate blocks (forward-port + the two backport rules) into a single regex-driven rule? Mergify supports named-capture in `label~=` and exposes the capture as a Jinja2 variable, e.g.:

```yaml
- name: backport via label
  conditions:
    - merged
    - label~=^backport-v(?P<branch>\d+\.\d+)$
  actions:
    backport:
      branches: ["{{ branch }}"]
```

That would directly address the recurring-maintenance concern raised earlier in this review: adding support for a new release branch becomes "create the `backport-vX.Y` label" with no edit to this file. The forward-port case can stay as a small dedicated rule (different action semantics/title), or be merged into the same scheme with a second pattern.

.github/mergify.yml
pebrc, line 1: `CONTRIBUTING.md` (section "Backports") still describes the `sqren/backport` CLI flow as the way to backport. If this label-driven flow becomes canonical, that doc should be updated — either here or as a tracked follow-up — so contributors don't have two competing instructions.

.github/mergify.yml
```diff
@@ -0,0 +1,58 @@
...
+
+
+  # Forward-ports merged PRs labeled "forwardport-main" to the main branch via backport. In case we merge something in a release branch first, and then we need it in main branch.
+  - name: forward-port patches to main branch
+    conditions:
+      - merged
+      - label=forwardport-main
+    actions:
+      backport:
+        branches: ["main"]
```
pebrc, line 30: Nit: this rule reuses the shared `defaults.actions.backport.title`, so a forward-ported PR will be titled `[main] (backport #N) …` — slightly misleading since the operation is a forward-port. Consider overriding `title:` on this action to say `forward-port` for clarity.

.github/mergify.yml
moukoublen: Are you sure something like that should work? I remember checking and couldn't find a way to generalize the rules, I am not sure if this regex group approach will work in mergify, I will take a look.

pebrc: I am not sure. This is based on a AI suggestion.

pkoutsovasilis: AFAIK Mergify does not work that way: label~= only uses the regex to decide whether the condition matches. Named capture groups from that condition are not exposed as template variables, so branches: ["{{ branch }}"] would not be populated from (?P<branch>…), action templates are built from [pull request attributes](https://docs.mergify.com/configuration/data-types/#template) (https://docs.mergify.com/configuration/data-types/#template), not from regex captures on conditions).

But we can do it in a different way, as other repos do (e.g. in beats), using github actions. 

Example below:

1. First extend the command restrictions
  ```yaml
  ‎# .github/mergify.yml
  commands_restrictions:
    backport:
      conditions:
        - or:
            - sender-permission >= write
            - sender = elastic-renovate-prod[bot]
            - sender = github-actions[bot]
  ```

2. Define the github action
  ```yaml
  # .github/workflows/mergify-backport-from-labels.yml
  name: Mergify backport from labels
  on:
    pull_request_target:
      types: [closed]
      branches: [main]
  # Token only used to comment on the PR (Mergify reads the comment).
  permissions:
    pull-requests: write
  concurrency:
    group: mergify-backport-pr-${{ github.event.pull_request.number }}
    cancel-in-progress: true
  jobs:
    backport:
      # Closed-but-not-merged must never post a backport command.
      if: >-
        github.event.pull_request.merged == true &&
        github.event.pull_request.base.ref == 'main' &&
        github.event.pull_request.head.repo.full_name == github.repository
      runs-on: ubuntu-latest
      steps:
        - name: Post @Mergifyio backport command
          # Pin to commit SHA from https://github.com/actions/github-script/releases
          uses: actions/github-script@xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
          with:
            script: |
              const pr = context.payload.pull_request;
              const labels = (pr.labels || []).map((l) => l.name);
              const labelRe = /^backport-v(\d+\.\d+)$/;
              const branchRe = /^\d+\.\d+$/;
              const branches = [];
              const seen = new Set();
              for (const name of labels) {
                const m = name.match(labelRe);
                if (!m) continue;
                const b = m[1];
                if (!branchRe.test(b)) {
                  core.setFailed(`Invalid branch segment from label ${name}: ${b}`);
                  return;
                }
                if (!seen.has(b)) {
                  seen.add(b);
                  branches.push(b);
                }
              }
              if (branches.length === 0) {
                core.info("No backport-vX.Y labels; skipping.");
                return;
              }
              // Single space-separated list; each segment is strictly X.Y
              const body = `@Mergifyio backport ${branches.join(" ")}`;
              await github.rest.issues.createComment({
                owner: context.repo.owner,
                repo: context.repo.repo,
                issue_number: pr.number,
                body,
              });
              core.info(`Posted: ${body}`);
  ```

moukoublen: @pkoutsovasilis  indeed, I see this approach in oblt-actions backport-active, we could perhaps change the `on` to `types: [closed,labeled]` to emulate the functionality that mergify offers that opens PR even if we add the label to a merged PR after it got merged.  

I think we need to consider whether to maintain the "label"->"branch" mapping in the mergify config, or the code in the GitHub action (even though once it's done, not much maintenance will be required). 

I am good with both to be honest. Perhaps exporting the code to a standalone action `./.github/actions/mergify-labels` for better maintainability.

moukoublen: On the other hand, since there are some steps that we do on main, each time we create a release branch, creating a github action to do all of those and also add a section to the mergify.yaml with the new label and backport branch is pretty trivial (this is how we did it on cloudbeat). 

But this discussion overlaps a bit with the release automation that we will have in a separate space. 

Perhaps merging the config in its simpler form and then, as part of the release automation discussion, we can decide on a more generic approach to the automations that we will implement for the release process, what do you all think?

.github/mergify.yml
moukoublen: done in [8bfc8e9](https://github.com/elastic/cloud-on-k8s/pull/9348/commits/8bfc8e9fe4e28a0970b114851287efa1a19a36d8)

.github/mergify.yml
moukoublen: Done in [8bfc8e9](https://github.com/elastic/cloud-on-k8s/pull/9348/commits/8bfc8e9fe4e28a0970b114851287efa1a19a36d8)

.github/mergify.yml
pkoutsovasilis: I’m in favour of implementing the GitHub Action approach now instead of keeping static per-branch Mergify rules, for a few reasons:
* We should start removing manual release steps, not adding another one (even if it’s trivial). 
* Static branch mappings in .mergify.yml are harder to validate safely: testing against non-release branches usually means temporary/invalid rule edits. With the workflow + github-script approach, we can point to temporary branches and test both clean and conflicting backports without reshaping Mergify rules each time.
* There’s no urgency that forces us to merge a partial solution now.

No strong feelings either way though - this is just my current thinking process, and I’m happy to go with what the group prefers.

.github/mergify.yml
pkoutsovasilis: Since this isn't settled yet and they are gonna be changes, I would vote for changing the `CONTRIBUTING.md` in a follow-up when we have solidified the new backport flow

pkoutsovasilis (approved):

PR #9359, opened 2026-04-16
Label Based Resource Discovery

### LINKED ISSUE #9026: Label selector for created secrets
[KEP 4601](https://github.com/kubernetes/enhancements/issues/4601) has recently been added to Kubernetes. It allows to improve security by limiting access to resources based on labels (and also fields).

Please improve support for this KEP in you operator by:

- Adding a label to all created resources, specific to this operator
- Add an option to only watch for secrets using this specific label (opt-in, for backward compatibility

Thanks

NB: Potential places to change: [`repo:elastic/cloud-on-k8s  SecretList -path:/_test.go$/`](https://github.com/search?q=repo%3Aelastic%2Fcloud-on-k8s++SecretList+-path%3A%2F_test.go%24%2F&type=code)

moukoublen:
## Summary

Introduces an opt-in operator flag, `--label-based-discovery`, that narrows the controller-runtime cache for `Secret`, `Service`, and `ConfigMap` resources to those carrying the new `eck.k8s.elastic.co/watched=true` label. When the flag is enabled, unlabeled resources of these types are excluded from the cache entirely, reducing the operator's memory footprint and API server load in clusters with large numbers of user-managed Secrets, Services, and ConfigMaps.

To make this safe to enable, every Secret, Service, and ConfigMap that ECK itself creates or reconciles is now stamped with the watch label, so flipping the flag on does not drop any resource the operator needs to see.

## Changes

- **New flag**: `--label-based-discovery` in `cmd/manager/main.go` (default `false`); flag constant in `pkg/controller/common/operator/flags.go`.
- **New label**: `eck.k8s.elastic.co/watched=true` (`commonv1.LabelBasedDiscoveryLabelName` / `LabelBasedDiscoveryLabelValue`) in `pkg/apis/common/v1/common.go`.
- **Cache wiring**: cache `ByObject` construction has been extracted into a new `buildByObject` helper. When the flag is enabled, `Secret`, `Service`, and `ConfigMap` are added to the cache's `ByObject` map with a label selector on `eck.k8s.elastic.co/watched=true`, alongside the existing ECK-owned filters on `Pod`, `Deployment`, `StatefulSet`, `DaemonSet`, and `PodDisruptionBudget` (which filter by `common.k8s.elastic.co/type`).
- **Secret labeling**: `ReconcileSecret` and `ReconcileSecretNoOwnerRef` in `pkg/controller/common/reconciler/secret.go` now add the watch label to every reconciled Secret.
- **Service labeling**: `ReconcileService` in `pkg/controller/common/service_control.go` now adds the watch label to every reconciled Service.
- **ConfigMap labeling**: the watch label is added to ECK-managed ConfigMaps in `pkg/controller/elasticsearch/configmap/configmap.go`, `pkg/controller/kibana/initcontainer/configmap.go`, `pkg/controller/autoops/configmap.go`, and `pkg/controller/common/immutableconfig/resource.go` (the `cmLabels` map pre-size was bumped from `+2` to `+3` to account for the new label).
- **Operator UUID ConfigMap**: `pkg/about/info.go` now stamps `eck.k8s.elastic.co/watched=true` on the operator's UUID ConfigMap when it is first created, so the operator can still read it back from the cache when the flag is enabled.
- **License Secrets**: `pkg/controller/common/license/crud.go` (`CreateTrialLicense`, `UpdateEnterpriseLicense`) and `pkg/controller/common/license/trial.go` (`ExpectedTrialStatus`) now stamp the watch label on trial-license, trial-status, and Enterprise-license Secrets.
- **File-settings Secret**: `pkg/controller/elasticsearch/filesettings/file_settings_secret.go` (`buildSecret`) and `pkg/controller/elasticsearch/filesettings/reconciler.go` (`reconcileSecret`) now stamp the watch label on file-settings Secrets.
- **AutoOps ES CA Secret**: `pkg/controller/autoops/secret.go` (`buildAutoOpsESCASecret`) now stamps the watch label on the AutoOps-managed CA Secret.
- **Elasticsearch roles/file-realm Secret**: `pkg/controller/elasticsearch/user/reconcile.go` (`reconcileRolesFileRealmSecret`) now stamps the watch label on the internal users Secret.
- **Helm chart / static config**:
  - `deploy/eck-operator/values.yaml`: new `config.labelBasedDiscovery` value (default `false`) with inline documentation.
  - `deploy/eck-operator/templates/configmap.yaml`: renders the value into the operator ConfigMap as `label-based-discovery`.
  - `config/eck.yaml`: default operator config now includes `label-based-discovery: false`.
- **Documentation**: new row in `docs/reference/eck-configuration-flags.md` describing the flag.
- **Tests**: unit tests updated across association, autoops, beat, certificates, keystore, filesettings, license (trial), logstash, remote cluster, stack-config-policy, user, and service-control packages to account for the new label; integration test for ES shared driver reconcile adjusted.

## Reconcile changes for existing resources

Several code paths previously short-circuited reconciliation when a Secret already existed and its payload was still valid. After an upgrade, those short-circuits would prevent the new `eck.k8s.elastic.co/watched=true` label from ever being stamped onto pre-existing Secrets. These paths have been updated so that they still reconcile the Secret's metadata (labels and annotations) while preserving the original "don't rotate or regenerate the payload" behavior:

- **`pkg/controller/common/certificates/ca_reconcile.go`**: `ReconcileCAForOwner` used to return the existing CA immediately when `CanReuseCA` returned true, skipping any write. It now calls `reconciler.ReconcileSecret` on the existing CA Secret before returning, so missing labels and annotations (notably the discovery label) are reconciled on upgrade without regenerating the CA.
- **`pkg/controller/autoops/api_key.go`**: `maybeUpdateAPIKey` used to return the existing Secret unchanged whenever the stored API key was still valid and its config hash matched. It now routes through `storeAPIKeyInSecret`, which reconciles the Secret's metadata (including the discovery label) without invalidating or rotating the key. The expected label set built in `buildAutoOpsESAPIKeySecret` now also includes `commonv1.LabelBasedDiscoveryLabelName`.
- **`pkg/controller/elasticsearch/driver/shared/reconcile.go`**: `MaybeReconcileEmptyFileSettingsSecret` used to return early with `(false, nil)` whenever the file-settings Secret already existed. It now calls `fs.Save(ctx, c, es)` on the existing Secret so any newly-managed labels and annotations are propagated; `Save` is a no-op when nothing has actually changed, so this does not introduce spurious writes.
- **`pkg/controller/common/certificates/http_reconcile.go`**: `ReconcileInternalHTTPCerts` already reconciled labels, but was not aware of the new one. It now explicitly ensures `eck.k8s.elastic.co/watched=true` is present on the internal HTTP certs Secret and flips `needsUpdate` when it has to add it.
- **`pkg/controller/elasticsearch/certificates/transport/reconcile.go`**: `ensureTransportCertificatesSecretExists` now stamps the discovery label alongside the existing `StatefulSetNameLabelName` on transport certificate Secrets, so those Secrets remain visible to the cache when the flag is enabled.
- **`pkg/controller/license/trial/trial_controller.go`**: the `trialLicensePopulated && validLicense && IsTrialStarted` branch used to be a no-op, which meant pre-existing trial-license Secrets would never receive the new label. It now calls `reconciler.ReconcileSecret` on that branch so labels/annotations are propagated without changing any license payload. `reconcileTrialStatus` also gained a `maps.IsSubset` check on labels and annotations so that a trial-status Secret missing the new label triggers an update (merging existing labels/annotations rather than overwriting them).
- **`pkg/controller/logstash/service.go`**: `reconcileService` previously discarded the reconciled Service returned by `common.ReconcileService`, so `apiSvc` and the returned slice reflected the pre-reconcile object without the watch label. It now returns the reconciled `*corev1.Service` and `reconcileServices` uses that value, so downstream consumers (and the cache) see the labeled Service.

Together these changes guarantee that after an operator upgrade (with the flag off), every Secret, Service, and ConfigMap that ECK owns will be re-labeled on its next reconcile pass, the prerequisite for the flag to be safely flipped on afterwards.

## Not enabled by default

When `--label-based-discovery` is on, the controller-runtime cache is configured with a label selector that only lists and watches Secrets, Services, and ConfigMaps already carrying `eck.k8s.elastic.co/watched=true`. Pre-existing resources without the label are invisible to the operator's cached client, so the reconcilers never see them, which means the "missing label → merge on update" path in `NeedsUpdate` / `UpdateReconciled` never fires. Only resources the operator creates from scratch (or that someone labels manually) ever enter the cache and get reconciled.

Because of this, the flag cannot ship on by default: an upgrade with the flag on would leave every pre-existing ECK-managed Secret, Service, and ConfigMap outside the cache, with no mechanism to reconcile them back in.

> Even setting that issue aside, we would still want to ship this off-by-default: any user who relies on unlabeled, custom Secrets, Services, or ConfigMaps (for example, referenced from an `Elasticsearch` or `Kibana` spec) would suddenly find them invisible to the operator after an upgrade, potentially breaking their clusters until they manually labeled every such resource.

Shipping the flag off by default avoids both problems: the operator keeps fetching all Secrets, Services, and ConfigMaps as before, and on each reconcile it stamps `eck.k8s.elastic.co/watched=true` onto the ones it owns. Once enough time has passed for every ECK-managed resource in a cluster to have been reconciled at least once, an operator can safely flip the flag on.

## E2E changes

The e2e test harness gained first-class support for running the full suite against an operator deployed with `--label-based-discovery` enabled, so the same test matrix can be exercised in both modes without ad-hoc overrides.

- **New runner flag**: `test/e2e/cmd/run/command.go` adds `--label-based-discovery` (default `false`) on the `e2e` command, bound via Viper so it can also be driven by the `E2E_LABEL_BASED_DISCOVERY` environment variable.
- **Test context propagation**: `test/e2e/cmd/run/run.go` passes `h.labelBasedDiscovery` into the rendered test `Context`; `test/e2e/test/context.go` exposes it as a new `LabelBasedDiscovery bool` field (`json:"label_based_discovery"`) so individual tests can branch on it if needed.
- **Helm deployment**: `config/e2e/helm-operator-under-test.yaml` now templates `config.labelBasedDiscovery: {{ .LabelBasedDiscovery }}` so the operator-under-test chart installation picks up the flag from the test context.
- **Makefile**: `E2E_LABEL_BASED_DISCOVERY` variable (defaulting to the `LABEL_BASED_DISCOVERY` env var from Buildkite, otherwise `false`) is forwarded to both the `e2e-run` and `e2e-local` targets as `--label-based-discovery=$(E2E_LABEL_BASED_DISCOVERY)`.
- **Buildkite pipeline generator**: `.buildkite/e2e/pipeline-gen/main.go` adds a new `EnvVarLabelBasedDiscovery` (`LABEL_BASED_DISCOVERY`) with the `lbd` shortcut, and `.buildkite/e2e/pipeline-gen/README.md` documents the shortcut and adds an example: `pipeline-gen -f p=kind,lbd=true,t=TestSmoke`.
- **Unrelated cleanup in the local runner**: `runTestsLocally` in `test/e2e/cmd/run/run.go` no longer wraps the local run in a `context.WithTimeout(h.testTimeout, ...)` — it now uses `context.WithCancel`, so a local run is bounded by Ctrl+C rather than a single global timeout. The `--test-timeout` help string in `command.go` is clarified to reflect that it only bounds individual retry loops (`Eventually`), not the overall run.


prodsecmachine:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/1952389e-7492-4190-bb50-56c19ba47403?refs=a0f78fe1234f891397993f4413477089f4d0c631&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/1952389e-7492-4190-bb50-56c19ba47403/license?refs=a0f78fe1234f891397993f4413477089f4d0c631&source=prChecksComment) |

cmd/manager/main.go
```diff
@@ -363,6 +363,11 @@ func Command() *cobra.Command {
 		"auto-detect",
 		"Enables setting the default security context with fsGroup=1000 for Elasticsearch 8.0+ Pods and Kibana 7.10+ Pods. Possible values: true, false, auto-detect",
 	)
+	cmd.Flags().Bool(
+		operator.LabelBasedDiscovery,
+		false,
+		"Restrict resource discovery (secrets and services) to labeled resources only. Unlabeled resources are skipped.",
```
pkoutsovasilis: confimaps is missing from the resource discovery parenthesis, maybe we should add also the label here?

cmd/manager/main_test.go
```diff
@@ -337,3 +343,56 @@ func newFakeK8sClientsetWithDiscovery(resources []*metav1.APIResourceList, disco
...
+				kind := reflect.TypeOf(obj).String()
+				query := cnf.Label.String()
+
+				expected, found := tc.expectedLabelQueryByTypeName[kind]
+				assert.True(t, found, "object not found in expected, %s", kind)
+				assert.Equal(t, expected, query)
+			}
+		})
+	}
+}
```
pkoutsovasilis, line 399: since we can't get away of using reflect here let's use it in a less fragile way (remove the use of `reflect.TypeOf(obj).String()`)


```suggestion
func TestBuildByObject(t *testing.T) {
	labelScopedExpected := commonv1.LabelBasedDiscoveryLabelName + `=true`

	tests := []struct {
		name                         string
		labelBasedDiscovery          bool
		expectedLabelQueryByTypeName map[reflect.Type]string
	}{
		{
			name:                "without label scope",
			labelBasedDiscovery: false,
			expectedLabelQueryByTypeName: map[reflect.Type]string{
				reflect.TypeFor[*corev1.Pod]():                   commonv1.TypeLabelName,
				reflect.TypeFor[*policyv1.PodDisruptionBudget](): commonv1.TypeLabelName,
				reflect.TypeFor[*appsv1.Deployment]():            commonv1.TypeLabelName,
				reflect.TypeFor[*appsv1.StatefulSet]():           commonv1.TypeLabelName,
				reflect.TypeFor[*appsv1.DaemonSet]():             commonv1.TypeLabelName,
			},
		},
		{
			name:                "with label scope",
			labelBasedDiscovery: true,
			expectedLabelQueryByTypeName: map[reflect.Type]string{
				reflect.TypeFor[*corev1.Pod]():                   commonv1.TypeLabelName,
				reflect.TypeFor[*policyv1.PodDisruptionBudget](): commonv1.TypeLabelName,
				reflect.TypeFor[*appsv1.Deployment]():            commonv1.TypeLabelName,
				reflect.TypeFor[*appsv1.StatefulSet]():           commonv1.TypeLabelName,
				reflect.TypeFor[*appsv1.DaemonSet]():             commonv1.TypeLabelName,
				reflect.TypeFor[*corev1.Secret]():                labelScopedExpected,
				reflect.TypeFor[*corev1.Service]():               labelScopedExpected,
				reflect.TypeFor[*corev1.ConfigMap]():             labelScopedExpected,
			},
		},
	}

	for _, tc := range tests {
		t.Run(tc.name, func(t *testing.T) {
			got, err := buildByObject(tc.labelBasedDiscovery)
			require.NoError(t, err)
			require.Len(t, got, len(tc.expectedLabelQueryByTypeName), "not expected number of items")

			for obj, cnf := range got {
				kind := reflect.TypeOf(obj)
				query := cnf.Label.String()

				expected, found := tc.expectedLabelQueryByTypeName[kind]
				assert.True(t, found, "object not found in expected, %s", kind)
				assert.Equal(t, expected, query)
			}
		})
	}
}
```

pkg/controller/common/immutableconfig/resource.go
```diff
@@ -61,10 +63,11 @@ func BuildImmutableConfigMap(baseName, namespace string, data map[string]string,
 	fullHash := ComputeStringContentHash(data)
 	name := ImmutableName(baseName, fullHash)
 
-	cmLabels := make(map[string]string, len(labels)+2)
+	cmLabels := make(map[string]string, len(labels)+3)
 	maps.Copy(cmLabels, labels)
 	cmLabels[ConfigTypeLabelName] = ConfigTypeImmutable
 	cmLabels[ConfigHashLabelName] = ShortHash(fullHash, hashLabelLen)
+	cmLabels[commonv1.LabelBasedDiscoveryLabelName] = commonv1.LabelBasedDiscoveryLabelValue
```
pkoutsovasilis: the same needs to change in `BuildImmutableSecret` above?

pkg/controller/common/reconciler/secret.go
```diff
@@ -43,6 +44,11 @@ func WithPostUpdate(f func()) func(p *Params) {
 // ReconcileSecret creates or updates the actual secret to match the expected one.
 // Existing annotations or labels that are not expected are preserved.
 func ReconcileSecret(ctx context.Context, c k8s.Client, expected corev1.Secret, owner client.Object, opts ...func(*Params)) (corev1.Secret, error) {
+	if expected.Labels == nil {
+		expected.Labels = make(map[string]string)
+	}
+	expected.Labels[commonv1.LabelBasedDiscoveryLabelName] = commonv1.LabelBasedDiscoveryLabelValue
```
pkoutsovasilis: look at `ReconcileSecretNoOwnerRef` 

```
// don't mutate expected (no side effects), make a copy
	expected = *expected.DeepCopy()
	if expected.Labels == nil {
		expected.Labels = make(map[string]string)
	}
```

since we mutate, let's deepCopy() here as well?

pkoutsovasilis (commented):
did a first pass of this PR, one more extra point
* `pkg/controller/common/license/crud.go:110`: calls c.Update directly without going through any reconciler. The operator-namespace license secret will never get the label via this path.

I need to do some more manual testing here. In addition, I think it would be useful if we were able to invoke some e2e tests with this flag enabled through a gh comment

moukoublen:
> pkg/controller/common/license/crud.go:110: calls c.Update directly without going through any reconciler. The operator-namespace license secret will never get the label via this path.

Good catch. It's also the trial status `reconcileTrialStatus` that calls c.Update directly and need this path (and I also need to check for any short-circuited reconciliation, like the ones I mentioned in the description).

cmd/manager/main.go
moukoublen: Done in [890d77d](https://github.com/elastic/cloud-on-k8s/pull/9359/commits/890d77dfb55db058683085ef164ef6adb7cf78dd) and 
0231577

pkg/controller/common/reconciler/secret.go
moukoublen: Then I think we need to do the same in all `reconcile secret` type of functions. I will take a look.

cmd/manager/main_test.go
moukoublen: Done

pkg/controller/common/immutableconfig/resource.go
moukoublen: I couldn't find a non-test usage of `BuildImmutableSecret,` so I left it as is, and actually, it's the opposite; this change should be removed since there is no non-test usage of `BuildImmutableConfigMap` as well.

(Ideally, the plan is not to add the label in the builders rather than in the "bottom" step (before reconciling) -whenever is possible-, but I will write a note about it in the description)

pkg/controller/common/reconciler/secret.go
moukoublen: I changed the side effect of any function that accepts a resource.

moukoublen:
buildkite test this -f p=kind,lbd=true,t=TestSmoke -m s=9.3.1

pkg/controller/license/trial/trial_controller.go
```diff
@@ -149,10 +150,14 @@ func (r *ReconcileTrials) reconcileTrialStatus(ctx context.Context, licenseName
...
-	if reflect.DeepEqual(expected.Data, trialStatus.Data) {
+	if reflect.DeepEqual(expected.Data, trialStatus.Data) &&
+		maps.IsSubset(expected.Labels, trialStatus.Labels) &&
+		maps.IsSubset(expected.Annotations, trialStatus.Annotations) {
 		return nil
 	}
 	trialStatus.Data = expected.Data
+	expected.Labels = maps.Merge(expected.Labels, trialStatus.Labels)
+	expected.Labels = maps.Merge(expected.Annotations, trialStatus.Annotations)
 	return r.Update(ctx, &trialStatus)
```
pebrc, line 161: Two bugs in this block:

**1. Copy-paste error (line 160)** — `expected.Labels` is assigned twice. Line 160 should be:
```go
expected.Annotations = maps.Merge(expected.Annotations, trialStatus.Annotations)
```

**2. Merged result is applied to the wrong variable** — the merge is computed into `expected`, but `r.Update` operates on `trialStatus`. The reconciled labels/annotations are silently discarded: the secret is updated with new `Data` but its labels remain unchanged, so the watched label is never written.

The correct fix:
```go
trialStatus.Data = expected.Data
trialStatus.Labels = maps.Merge(expected.Labels, trialStatus.Labels)
trialStatus.Annotations = maps.Merge(expected.Annotations, trialStatus.Annotations)
return r.Update(ctx, &trialStatus)
```

pkg/controller/license/trial/trial_controller.go
pebrc, line 1: Unrelated to the above: the `reconcileTrialStatus` function calls `r.Update(ctx, &trialStatus)` directly rather than going through `ReconcileSecret`. This means the overall "avoid side effects via DeepCopy" pattern adopted in this PR doesn't apply here, and any future label/annotation changes added to `ExpectedTrialStatus` would again need manual wiring into this update path. Worth a note or a follow-up to route through `ReconcileSecret` consistently.

pebrc (changes requested):
Second-pass review after the latest round of fixes. The previously raised issues are in good shape — the `DeepCopy` side-effect fixes, the `reflect.TypeFor` test rewrite, and the license/trial label backfill are all solid. Two remaining items:

**Blocker — bug in `reconcileTrialStatus` (introduced in `28d10190b`):** The label merge in lines 158–161 of `trial_controller.go` has a copy-paste error and applies the result to the wrong variable (`expected` instead of `trialStatus`). The trial status secret will be updated with new `Data` but labels/annotations are silently discarded — the watched label never reaches the stored object. See inline comment.

**Minor — `pkg/about/info.go` operator UUID ConfigMap:** `getOperatorUUID` creates and updates the operator UUID ConfigMap via the raw `clientset.CoreV1().ConfigMaps()` typed client, bypassing all ECK reconciliation helpers. This ConfigMap won't carry the `eck.k8s.elastic.co/watched=true` label. If `--label-based-discovery` is enabled and the operator namespace is a managed namespace, the cache will never see this ConfigMap. This is unlikely to be a practical problem today (the UUID ConfigMap is read once at startup, before the cache is active), but it's worth either adding the label or leaving an explicit comment explaining why this ConfigMap is intentionally excluded.

pkg/controller/license/trial/trial_controller.go
moukoublen: I am planning to create a ticket in general to route any reconciliation of `Secret`, `ConfigMap`, `Service` through the common ones (and if there is not one, to create a common one). 

Unfortunately, we don't have a way to enforce that anyone who creates a new secret, configmap, or service uses the new label, unless we create generic reconcilers to do so (like the current common Secret reconciler) and document that any resource creation should use them. 

Do you see any reason not to do that?

pkg/controller/license/trial/trial_controller.go
moukoublen: Done in [08a3ac2](https://github.com/elastic/cloud-on-k8s/pull/9359/commits/08a3ac24f381fda9ab86cb8ea6e8658f14e7b406)
and
[c05f3cd](https://github.com/elastic/cloud-on-k8s/pull/9359/commits/c05f3cd9e4dacb02424b5cacd148dcafdfb91e1f)

pkg/controller/common/immutableconfig/resource.go
moukoublen: Removed in 58bf69d

pkg/controller/common/operator/flags.go
```diff
@@ -47,4 +47,5 @@ const (
 	WebhookNameFlag                      = "webhook-name"
 	WebhookSecretFlag                    = "webhook-secret"
 	WebhookPortFlag                      = "webhook-port"
+	LabelBasedDiscovery                  = "label-based-discovery"
```
pebrc: Nit on naming: "discovery" is overloaded in Kubernetes (typically refers to API discovery — `/apis`, resource type enumeration). What this flag actually controls is **cache/watch scope**. A user scanning `--help` might not connect "discovery" to "which Secrets/Services/ConfigMaps the operator watches."

The name also describes the mechanism ("label-based") rather than the effect. Compare with existing flags like `--enforce-rbac-on-refs` or `--validate-storage-class` which describe what they do.

Something like `--restrict-watched-resources` or `--scoped-resource-cache` would be more self-explanatory. Acknowledge this is wired through Helm values, e2e, Buildkite, and docs at this point so renaming has a cost — but worth considering before this ships as a user-facing flag name.

pkg/controller/common/immutableconfig/resource.go
```diff
@@ -61,7 +61,7 @@ func BuildImmutableConfigMap(baseName, namespace string, data map[string]string,
 	fullHash := ComputeStringContentHash(data)
 	name := ImmutableName(baseName, fullHash)
 
-	cmLabels := make(map[string]string, len(labels)+2)
+	cmLabels := make(map[string]string, len(labels)+3)
```
pebrc: Nit: the capacity hint was bumped from `+2` to `+3`, but the function body still only adds 2 extra labels (`ConfigTypeLabelName`, `ConfigHashLabelName`). Since this function is only called from tests (no production callers), the `+3` is misleading — it suggests a third label was intended here. Should revert to `+2`.

pkg/controller/elasticsearch/filesettings/reconciler.go
```diff
@@ -55,6 +56,13 @@ func reconcileSecret(
 	owner client.Object,
 	managedAnnotations []string,
 ) error {
+	// don't mutate expected (no side effects), make a copy
+	expected = *expected.DeepCopy()
+	if expected.Labels == nil {
+		expected.Labels = make(map[string]string)
+	}
+	expected.Labels[commonv1.LabelBasedDiscoveryLabelName] = commonv1.LabelBasedDiscoveryLabelValue
```
pebrc: Nit: this function handles ES config and Kibana config Secrets (not the file-settings Secret itself, which goes through `buildSecret` in `file_settings_secret.go`). Both paths stamp the discovery label independently. This is correct but at first glance looks like redundant stamping. A brief comment clarifying which Secret types flow through this `reconcileSecret` vs. `Save`/`buildSecret` would help future readers.

pkg/about/info.go
```diff
@@ -132,6 +132,9 @@ func getOperatorUUID(ctx context.Context, clientset kubernetes.Interface, operat
 			ObjectMeta: metav1.ObjectMeta{
 				Namespace: operatorNs,
 				Name:      UUIDCfgMapName,
+				Labels: map[string]string{
+					commonv1.LabelBasedDiscoveryLabelName: commonv1.LabelBasedDiscoveryLabelValue,
+				},
```
pebrc, line 137: The Create path correctly stamps the label here. However, the Update path further down (~line 153–159, not in diff) does not: when the UUID key is missing from an existing ConfigMap, `reconciledCfgMap` is updated via `c.Update` without adding the discovery label.

This is likely harmless (the UUID ConfigMap is read via raw clientset at startup, before the cache is active), but it's inconsistent. Either add the label on the Update path too, or add a comment there explaining why this ConfigMap is intentionally excluded from label-based discovery.

cmd/manager/main.go
```diff
@@ -1073,3 +1066,38 @@ func fipsLog() {
...
+		return byObject, nil
+	}
+	log.Info("label based discovery is enabled")
+
+	labelBasedSelector := labels.SelectorFromSet(labels.Set{
+		commonv1.LabelBasedDiscoveryLabelName: commonv1.LabelBasedDiscoveryLabelValue,
+	})
+	byObject[&corev1.Secret{}] = cache.ByObject{Label: labelBasedSelector}
+	byObject[&corev1.Service{}] = cache.ByObject{Label: labelBasedSelector}
+	byObject[&corev1.ConfigMap{}] = cache.ByObject{Label: labelBasedSelector}
```
pebrc: **Bug — `elastic-licensing` ConfigMap invisible to cache when flag is on**

Tested locally: with `label-based-discovery=true`, the `elastic-licensing` ConfigMap in the operator namespace becomes invisible to the cached client. It carries `common.k8s.elastic.co/type=elastic-usage` but not `eck.k8s.elastic.co/watched=true`.

The resource reporter in `pkg/license/license.go:188` uses `reconciler.ReconcileResource` with the cached client. Since the cache can't see the ConfigMap, it sees "not found" → tries to Create → gets "already exists" from the API server. This error repeats every 2 minutes (the reporting frequency), and licensing data stops being updated.

Reproduction:
1. Install released ECK, deploy ES+Kibana, upgrade to PR image with flag=false (labels propagate)
2. Upgrade again with flag=true
3. `kubectl -n elastic-system logs elastic-operator-0 | grep "Failed to report licensing"` — error repeats every ~2min
4. `kubectl -n elastic-system get cm elastic-licensing -o jsonpath='{.data.timestamp}'` — timestamp frozen at the last successful update before the flag was enabled

Fix: add the watched label to the `elastic-licensing` ConfigMap in `pkg/license/license.go:180`, and backfill it on existing ConfigMaps (same upgrade-path pattern used elsewhere in this PR).

pebrc (changes requested):
Third-pass review after the latest round of fixes. All previously raised issues are resolved — the `reconcileTrialStatus` copy-paste bug, `DeepCopy` side-effect fixes, `reflect.TypeFor` test rewrite, and license/trial label backfill are all solid.

## Blocker

**`elastic-licensing` ConfigMap missing the watched label** — When `label-based-discovery=true`, the `elastic-licensing` ConfigMap in the operator namespace becomes invisible to the controller-runtime cache. The resource reporter (`pkg/license/license.go:188`) uses `reconciler.ReconcileResource` with the cached client, which returns "not found". It then tries to Create the ConfigMap and hits "already exists" from the API server. This error repeats every 2 minutes (the reporting interval), and licensing data stops being updated.

Verified locally: installed ECK 3.3.2 → deployed ES+Kibana → upgraded to `docker.elastic.co/eck-ci/eck-operator:pr-9359-69c5c0fa` with flag=false (labels propagated correctly to all 19 Secrets, 5 Services, 3 ConfigMaps) → upgraded with flag=true → `elastic-licensing` ConfigMap invisible, reporter error-loops every 2min. See inline comment for full reproduction steps.

Fix: add `eck.k8s.elastic.co/watched=true` to the `elastic-licensing` ConfigMap in `pkg/license/license.go:180` and backfill on existing ConfigMaps (same upgrade-path pattern used elsewhere in this PR).

## Non-blocking observations

- **Flag naming**: `--label-based-discovery` describes the mechanism, not the effect, and "discovery" is overloaded in Kubernetes. Something like `--restrict-watched-resources` would be more self-explanatory. See inline comment.
- **`info.go` Update path**: The UUID ConfigMap Create path stamps the label, but the Update path does not. Harmless (read via raw clientset) but inconsistent. A comment would help.
- **ConfigMap label stamping is fragmented**: No central `ReconcileConfigMap` helper — each call site stamps individually. Increases risk of future omissions (as the `elastic-licensing` miss demonstrates). The planned ticket to route reconciliation through common helpers would address this.
- **`BuildImmutableConfigMap` pre-size `+3`**: Function body only adds 2 extra labels; the `+3` hint is misleading (test-only function, cosmetic).

## Positive highlights

- The upgrade-path handling is thorough — short-circuit paths that previously skipped metadata are now fixed
- `DeepCopy`-before-mutation pattern consistently applied
- E2E harness support with Buildkite integration is well done
- Test coverage across all affected packages is comprehensive

pkg/about/info.go
moukoublen: Fixed in 036a2e6

pkg/controller/elasticsearch/filesettings/reconciler.go
moukoublen: Done in 036a2e6

pkg/controller/common/immutableconfig/resource.go
moukoublen: It was a leftover. Introduced when the watch label was added here, then the label was removed and that was a leftover. Done in  036a2e6

cmd/manager/main.go
moukoublen: Done in [14e69bb](https://github.com/elastic/cloud-on-k8s/pull/9359/commits/14e69bbff2236bc0167c6beb0cc060ed51789538)

pkg/controller/common/operator/flags.go
moukoublen: Done in [f53c757](https://github.com/elastic/cloud-on-k8s/pull/9359/commits/f53c757ce61d30a6d8f72a6fd240bb1751e6692c)

moukoublen:
buildkite test this -f p=kind,rwr=true,t=TestSmoke -m s=9.3.1

pebrc (approved):
LGTM — all previously raised issues are resolved (`elastic-licensing` blocker fix verified end-to-end, `info.go` label backfill, `BuildImmutableConfigMap` capacity hint reverted, flag renamed to `--restrict-watched-resources`, filesettings reconciler comment added). Approving.

## One follow-up observation (non-blocking)

**Webhook cert Secret + Service in operator namespace lack the watched label**

The webhook controller (`pkg/controller/webhook/webhook_certificates_controller.go:114`) sets up a cache-backed Watch on the operator's webhook cert Secret. But that Secret and the webhook Service are managed by the install method (Helm chart, raw YAML, or OLM), not created at runtime by ECK. In a fresh install of the eck-operator Helm chart, neither carries `eck.k8s.elastic.co/watched=true`:

```
elastic-operator-webhook-cert: app.kubernetes.io/managed-by=Helm  (no watched label)
elastic-operator-webhook (svc): app.kubernetes.io/managed-by=Helm  (no watched label)
```

With `--restrict-watched-resources=true`, the cache filter hides this Secret from the controller's Watch. The reconciler still reads/writes the Secret directly via `clientset` (`pkg/controller/webhook/reconcile.go:40,68`), and periodic requeue still fires from `ShouldRotateIn`, so cert rotation on schedule still works. But external/manual changes to the cert won't trigger a reconcile until the next periodic tick.

The webhook Service has the same gap, but no ECK controller reconciles it (the apiserver uses it to call back into the webhook), so it's a consistency issue rather than a functional one.

### Suggested fix

Stamp the watched label in `pkg/controller/webhook/reconcile.go:64` when the operator updates the webhook Secret. This is the cleanest because it covers all install methods uniformly (Helm, raw YAML, OLM, and any future packaging) and self-heals existing installs on first reconcile after upgrade — no install-method-specific change needed.

For the Service, since it's not functionally required and is outside the operator's reconciliation paths, options are: leave it (acceptable), or bake the label unconditionally into install manifests (Helm `templates/webhook.yaml`, raw `operator.yaml`, OLM bundle) — the label is harmless when the flag is off, so no conditional templating is needed.

.buildkite/e2e/pipeline-gen/README.md
```diff
@@ -64,11 +64,15 @@ pipeline-gen -e -f E2E_PROVIDER=gke,DEPLOYER_K8S_VERSION=1.23,E2E_STACK_VERSION=
 | `s` | `E2E_STACK_VERSION` |
 | `t` | `TESTS_MATCH` |
 | `sl` | `STATELESS` |
+| `rwr` | `RESTRICT_WATCHED_RESOURCES` |
```
pebrc, line 67: Should we have a nightly e2e run with this flag set?

pebrc: I think this would also require additional changes to the e2e tests.

pebrc:
buildkite test this -f p=kind,rwr=true,t=TestEnterprise -m s=9.3.1

.buildkite/e2e/pipeline-gen/README.md
moukoublen: Yes, e.g adding the watch label to resources created in e2e codebase like in this [e0d55ad](https://github.com/elastic/cloud-on-k8s/pull/9359/commits/e0d55ad3f743c604887b00c3bd67adecda29d179)

moukoublen:
buildkite test this -f p=kind,rwr=true,t=TestEnterprise -m s=9.3.1

pebrc (commented):
Follow-up on the e2e label-stamping (non-blocking, but worth tracking before the `lbd=true` Buildkite matrix is taken seriously)

The PR description states the suite can be exercised against an operator with `--restrict-watched-resources=true`. Today only two helpers stamp the label — `CreateEnterpriseLicenseSecret` (this file) and `CreateEnterpriseTrialLicenseSecret`. Most user-supplied Secrets/ConfigMaps that ECK is expected to read are still unlabeled, so they will be invisible to the cache and reconciliation will fail under flag=true.

Spot-checked sites (non-exhaustive, ~23 in total):

- Keystore / secureSettings: `test/e2e/es/keystore_test.go:30,40`, `test/e2e/kb/keystore_test.go:33`, `test/e2e/apm/configuration_test.go:50`, `test/e2e/logstash/keystore_test.go:40,99,109`, `test/e2e/beat/config_test.go:150`
- File realm / roles: `test/e2e/es/auth_test.go:73,84`, `test/e2e/kb/standalone_test.go:34`
- Logstash pipelines: `test/e2e/logstash/pipeline_test.go:24,89`
- StackConfigPolicy: `test/e2e/es/stackconfigpolicy_test.go:108,130`, `test/e2e/kb/stackconfigpolicy_test.go:74,225`
- Stack monitoring: `test/e2e/es/stack_monitoring_test.go:168`, `test/e2e/logstash/stack_monitoring_test.go:63`
- App config: `test/e2e/agent/config_test.go:64`, `test/e2e/ent/configuration_test.go:56`
- FIPS keystore: `test/e2e/es/fips_keystore_test.go:120` (+ ~276)
- Client TLS: `test/e2e/test/client-auth/checks.go:100,122`
- Global CA: `test/e2e/global_ca_test.go:208`

Suggested approach: rather than stamping at every call site, add a small helper (e.g. `func withWatchedLabel(s *corev1.Secret)` or wrap `K8sClient.CreateOrUpdate` for `*corev1.Secret`/`*corev1.ConfigMap`) so every user-supplied object the suite creates picks up the label uniformly. That mirrors the operator-side approach (one helper, broad coverage) and keeps future tests safe by default.

This can be a follow-up PR if you'd prefer to merge first and chase the Buildkite matrix afterwards — but flagging here so it isn't lost.

moukoublen:
> or wrap K8sClient.CreateOrUpdate for *corev1.Secret/*corev1.ConfigMap) so every user-supplied object the suite creates picks up the label uniformly. 

That would be my preference, to be honest. It's relatively simple, and I could add it to this pr.

PR #9367, opened 2026-04-17
Add downward-node-labels propagation to Kibana, Logstash and Elastic …

### LINKED ISSUE #9366: Feature: add downward-node-labels to Logstash / Kibana / Agent
**Is your feature request related to a problem? Please describe.**
I want to use node labels to set https://www.elastic.co/docs/reference/logstash/plugins/plugins-inputs-kafka#plugins-inputs-kafka-client_rack but downard-node-labels are only propogated for elasticsearch resources. 

**Describe the resource you would like to have implemented.**

Mirrors the existing Elasticsearch behavior: users can set the
eck.k8s.elastic.co/downward-node-labels annotation on the CR with a
comma-separated list of node labels, and the operator copies those labels
as annotations on the managed Pods. Pods can then read them via the
Downward API.

I primarily need this for Logstash but we might as well implement it for kibana / agent

orangecola:
Fixes #9366

I attest that I have reviewed the output


prodsecmachine:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/75cb328d-aa59-4938-975d-acde7ff1ccc5?refs=afde375183b9515a39d475786bd3088e493a8cf8&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/75cb328d-aa59-4938-975d-acde7ff1ccc5/license?refs=afde375183b9515a39d475786bd3088e493a8cf8&source=prChecksComment) |

barkbay (commented):
Elasticsearch runs a prepare-fs init container before the main Elasticsearch container. [That init script waits](https://github.com/elastic/cloud-on-k8s/blob/1679aab5ce833d32bc07360633226c7001d46189/pkg/controller/elasticsearch/initcontainer/prepare_fs_script.go#L156-L163) until the operator has written the requested node-derived keys onto the Pod’s metadata.annotations ([it polls the downward API annotations file](https://github.com/elastic/cloud-on-k8s/blob/1679aab5ce833d32bc07360633226c7001d46189/pkg/controller/elasticsearch/initcontainer/prepare_fs_script.go#L61)), so the main container does not start with those annotations missing.

This change adds the same operator patch behavior for Kibana, Logstash, and Elastic Agent **but not** that startup gate, so anything that consumes those annotations at first container start (for example `valueFrom.fieldRef` on `metadata.annotations[...]`) can still race on first start compared to Elasticsearch.

orangecola:
Have added to the startup gate for Kibana, Logstash, and Elastic Agent

orangecola:
@barkbay bump

barkbay:
> @barkbay bump

I'll do a review by the end of the week.

barkbay (commented):
Any reason to exclude Beat, APM Server, Maps or the Package Registry ?

barkbay (commented):
I think we need to reuse this validation: https://github.com/elastic/cloud-on-k8s/blob/d0d9521002a053c4e77bad390678c1d23ccd8b6b/pkg/controller/elasticsearch/validation/validations.go#L118

(and update https://github.com/elastic/cloud-on-k8s/blob/d0d9521002a053c4e77bad390678c1d23ccd8b6b/docs/reference/eck-configuration-flags.md)

barkbay (commented):
Is it worth extracting an interface? For example this one in `pkg/controller/common/nodelabels/nodelabels.go` :

```go
// AnnotationTarget is implemented by ECK custom resources whose managed Pods should have
// Kubernetes node labels copied to their annotations via AnnotatePods. Any ECK CR that already
// implements metav1.Object and GetIdentityLabels satisfies this interface once it exposes a
// DownwardNodeLabels accessor.
type AnnotationTarget interface {
	metav1.Object
	// DownwardNodeLabels returns the node labels expected to be copied as annotations on the
	// Pods managed by the resource. An empty result disables node-label propagation.
	DownwardNodeLabels() []string
	// GetIdentityLabels returns the label set identifying Pods managed by the resource.
	GetIdentityLabels() map[string]string
}

// AnnotatePods copies the expected node labels as annotations on all Pods managed by the given
// target. Missing node labels are reported as errors but do not stop the reconciliation of
// other Pods. The call is a no-op when the target has no downward node labels configured.
func AnnotatePods(ctx context.Context, c k8s.Client, t AnnotationTarget) *reconciler.Results {
	span, ctx := apm.StartSpan(ctx, "annotate_pods_with_node_labels", tracing.SpanTypeApp)
	defer span.End()
	expectedLabels := t.DownwardNodeLabels()
	if len(expectedLabels) == 0 {
		return results
	}
	pods, err := k8s.PodsMatchingLabels(c, t.GetNamespace(), t.GetIdentityLabels())
	if err != nil {
		return results.WithError(err)
	}
	for _, pod := range pods {
		results.WithError(annotatePod(ctx, c, pod, expectedLabels, t.GetName()))
	}
	return results
}
```

And then from the drivers:

```go
	// Patch the Pods to add the expected node labels as annotations. Record the error, if any, but do not stop the
	// reconciliation loop as we don't want to prevent other updates from being applied.
	statefulSetResults.WithResults(nodelabels.AnnotatePods(params.Context, params.Client, &params.Logstash))
```

orangecola:
**[Any reason to exclude Beat, APM Server, Maps or the Package Registry?](https://github.com/elastic/cloud-on-k8s/pull/9367#pullrequestreview-4360943025)**
extended support to all four. Each CR now has `DownwardNodeLabels()` / `HasDownwardNodeLabels()` / `DownwardNodeLabelsAnnotation`, drivers call `nodelabels.AnnotatePods`, pod templates add the `WaitForAnnotationsInitContainer` + downward-API volume when labels are configured, and the config hash mixes the annotation value so changes roll the Pods.

**[Reuse `validNodeLabels` and update the configuration docs](https://github.com/elastic/cloud-on-k8s/pull/9367#pullrequestreview-4360957591)**
Moved `NodeLabels` and `NewExposedNodeLabels` into `pkg/controller/common/nodelabels` and added a shared `ValidateAnnotation` helper. `esvalidation.NodeLabels` is now a type alias so existing imports keep working. The non-Elasticsearch webhooks are wrapped with a new `commonwebhook.WithNodeLabelsValidation` generic helper at registration so they all enforce the same `exposed-node-labels` policy, and the same check runs in each controller's `validate()` to cover the webhook-disabled path (matching the Elasticsearch behavior). `docs/reference/eck-configuration-flags.md` now lists all the CRDs the flag applies to.

**[Extract an `AnnotationTarget` interface](https://github.com/elastic/cloud-on-k8s/pull/9367#pullrequestreview-4361100361)**
added the interface you sketched in `pkg/controller/common/nodelabels/nodelabels.go`. Each driver call collapses to `nodelabels.AnnotatePods(ctx, client, &resource)`, and any ECK CR that already implements `metav1.Object` and `GetIdentityLabels` satisfies the interface once it exposes `DownwardNodeLabels()`.

pkg/controller/common/nodelabels/initcontainer.go
```diff
@@ -0,0 +1,74 @@
...
+import (
+	"bytes"
+	"fmt"
+	"strings"
+	"text/template"
+
+	corev1 "k8s.io/api/core/v1"
+
+	commonvolume "github.com/elastic/cloud-on-k8s/v3/pkg/controller/common/volume"
+	esvolume "github.com/elastic/cloud-on-k8s/v3/pkg/controller/elasticsearch/volume"
```
pkoutsovasilis: A common package should not depend on an ES-specific package

pkg/utils/nodelabels/nodelabels.go
```diff
@@ -0,0 +1,40 @@
...
+	if labels.Count() == 0 {
+		return nil
+	}
+	return labels.AsSortedSlice()
+}
+
+// FromAnnotations returns the list of downward node labels declared via DownwardNodeLabelsAnnotation.
+func FromAnnotations(annotations map[string]string) []string {
+	return Parse(annotations[DownwardNodeLabelsAnnotation])
+}
```
pkoutsovasilis: DownwardNodeLabelsAnnotation is a user-facing API annotation that applies to every ECK resource type. Placing the canonical definition in pkg/utils/nodelabels and then re-exporting it once per CRD type (DownwardNodeLabelsAnnotation = nodelabels.DownwardNodeLabelsAnnotation) is unnecessary indirection. I'd suggest moving the constant and the two parsing helpers to pkg/apis/common/v1, importing from there in all the CRD type files, and deleting pkg/utils/nodelabels entirely.

pkg/controller/common/nodelabels/initcontainer.go
```diff
@@ -0,0 +1,74 @@
...
+const WaitForAnnotationsContainerName = "elastic-internal-wait-for-node-labels"
+
+var waitScriptTemplate = template.Must(template.New("").Parse(
+	`#!/usr/bin/env bash
+set -eu
+expected_annotations=({{ .ExpectedAnnotations }})
+annotations_file={{ .AnnotationsFile }}
+function annotations_exist() {
+  for expected_annotation in "${expected_annotations[@]}"; do
+    if ! grep -qE "^${expected_annotation}=" "${annotations_file}"; then
```
pkoutsovasilis: With `-E` (ERE mode) the dots in label names like `topology.kubernetes.io/zone` are regex metacharacters matching any character. The existing Elasticsearch prepare_fs_script.go uses grep -c "^${expected_annotation}=" (BRE, no -E flag) - dots are still technically metacharacters there, but the same accepted behaviour. For consistency, let's align with the ES script.

pkg/controller/common/nodelabels/nodelabels.go
```diff
@@ -0,0 +1,140 @@
...
+	// Pods managed by the resource. An empty result disables node-label propagation.
+	DownwardNodeLabels() []string
+	// GetIdentityLabels returns the label set identifying Pods managed by the resource.
+	GetIdentityLabels() map[string]string
+}
+
+// AnnotatePods copies the expected node labels as annotations on all Pods managed by the given
+// target. Missing node labels are reported as errors but do not stop the reconciliation of
+// other Pods. The call is a no-op when the target has no downward node labels configured.
+func AnnotatePods(ctx context.Context, c k8s.Client, t AnnotationTarget) *reconciler.Results {
```
pkoutsovasilis, line 45: The new nodelabels.AnnotatePods in pkg/controller/common/nodelabels was designed as the single implementation for all resource types, but the Elasticsearch driver still calls its own local annotatePodsWithNodeLabels / annotatePodWithNodeLabels / isPodScheduled / getPodAnnotations. These are now duplicated with the common implementation. The ES driver should be migrated to call nodelabels.AnnotatePods and shared/node_labels.go deleted (or at minimum its internals replaced by calls to the common package)

cmd/manager/validation.go
```diff
@@ -87,26 +88,31 @@ func setupWebhook(
...
+		commonwebhook.WithNodeLabelsValidation(kbv1.Validate, exposedNodeLabels, schema.GroupKind{Group: kbv1.GroupVersion.Group, Kind: kbv1.Kind}), "Kibana")
 	commonwebhook.RegisterResourceWebhook(mgr, kbv1beta1.WebhookPath, checker, managedNamespaces, kbv1beta1.Validate, "Kibana")
-	commonwebhook.RegisterResourceWebhook(mgr, emsv1alpha1.WebhookPath, checker, managedNamespaces, emsv1alpha1.Validate, "Elastic Maps Server")
-	commonwebhook.RegisterResourceWebhook(mgr, eprv1alpha1.WebhookPath, checker, managedNamespaces, eprv1alpha1.Validate, "Package Registry")
+	commonwebhook.RegisterResourceWebhook(mgr, emsv1alpha1.WebhookPath, checker, managedNamespaces,
+		commonwebhook.WithNodeLabelsValidation(emsv1alpha1.Validate, exposedNodeLabels, schema.GroupKind{Group: emsv1alpha1.GroupVersion.Group, Kind: emsv1alpha1.Kind}), "Elastic Maps Server")
+	commonwebhook.RegisterResourceWebhook(mgr, eprv1alpha1.WebhookPath, checker, managedNamespaces,
+		commonwebhook.WithNodeLabelsValidation(eprv1alpha1.Validate, exposedNodeLabels, schema.GroupKind{Group: eprv1alpha1.GroupVersion.Group, Kind: eprv1alpha1.Kind}), "Package Registry")
```
pkoutsovasilis: WithNodeLabelsValidation here combined with the separate ValidateAnnotation call in each controller's validate() method creates a split: the webhook path enforces the policy through this wrapper, while the reconciler path does it through a standalone call bolted onto the existing validate(). As a result if somebody wants to understand or disable validation for these resources, now has to look in two places instead of one. Elasticsearch avoids this by passing exposedNodeLabels into ValidateElasticsearch directly, so one function covers both paths. IMO the affected types of this PR should migrate to the same paradigm (introduce RegisterWebhook) instead of using RegisterResourceWebhook + WithNodeLabelsValidation.

pkg/apis/agent/v1alpha1/agent_types.go
```diff
@@ -24,6 +25,9 @@ const (
 	AgentContainerName = "agent"
 	// FleetServerServiceAccount is the Elasticsearch service account to be used to authenticate.
 	FleetServerServiceAccount commonv1.ServiceAccountName = "fleet-server"
+	// DownwardNodeLabelsAnnotation holds an optional comma-separated list of expected node labels
+	// to be set as annotations on the Elastic Agent Pods.
+	DownwardNodeLabelsAnnotation = nodelabels.DownwardNodeLabelsAnnotation
```
pkoutsovasilis: if you move DownwardNodeLabelsAnnotation from utils to common/api/v1 this is redundant

pkg/controller/agent/pod.go
```diff
@@ -227,6 +234,21 @@ func buildPodTemplate(params Params, fleetCerts *certificates.CertificatesSecret
...
+	if params.Agent.HasDownwardNodeLabels() {
+		downwardAPIVolume := commonnodelabels.DownwardAPIVolume()
+		waitInit, err := commonnodelabels.WaitForAnnotationsInitContainer(params.Agent.DownwardNodeLabels())
+		if err != nil {
+			return corev1.PodTemplateSpec{}, err
+		}
+		builder = builder.
+			WithVolumes(downwardAPIVolume.Volume()).
+			WithInitContainers(waitInit).
+			WithInitContainerDefaults()
```
pkoutsovasilis: `WithInitContainerDefaults()` is called only inside the HasDownwardNodeLabels() branch. Every other resource type (Kibana, Logstash, APM, Beat, Maps, PackageRegistry) calls it unconditionally at the end of buildPodTemplate. Let's move the call outside when returning the PodTemplate?

pkoutsovasilis (commented):
Thanks for the contribution @orangecola! I did my pass on this PR and left some comments. Please while you address my comments merge main, so I can invoke a full run of our CI including the latest code that sits on main.

pkoutsovasilis:
buildkite test this

pkoutsovasilis:
buildkite test this -f p=gke,E2E_TAGS=e2e -m s=8.19.16,s=9.5.0-SNAPSHOT

[BK link](https://buildkite.com/elastic/cloud-on-k8s-operator/builds/15145)

pkg/apis/elasticsearch/v1/elasticsearch_types.go
```diff
@@ -33,7 +33,7 @@ const (
 	//       eck.k8s.elastic.co/disable-upgrade-predicates="if_yellow_only_restart_upgrading_nodes_with_unassigned_replicas"
 	DisableUpgradePredicatesAnnotation = "eck.k8s.elastic.co/disable-upgrade-predicates"
 	// DownwardNodeLabelsAnnotation holds an optional list of expected node labels to be set as annotations on the Elasticsearch Pods.
-	DownwardNodeLabelsAnnotation = "eck.k8s.elastic.co/downward-node-labels"
+	DownwardNodeLabelsAnnotation = commonv1.DownwardNodeLabelsAnnotation
```
pkoutsovasilis, line 36: Redundant re-export, DownwardNodeLabelsAnnotation now leaves in common api so it can be imported directly

pkg/controller/apmserver/deployment.go
```diff
@@ -117,6 +118,12 @@ func buildConfigHash(c k8s.Client, as *apmv1.ApmServer, params PodSpecParams) (s
 	// - in the APMServer configuration file content
 	_, _ = configHash.Write(params.ConfigSecret.Data[ApmCfgSecretKey])
 
+	// Changes to the downward-node-labels annotation must roll the APM Server Pods so the new annotations
+	// are re-applied on scheduling.
+	if as.HasDownwardNodeLabels() {
+		_, _ = configHash.Write([]byte(as.Annotations[commonv1.DownwardNodeLabelsAnnotation]))
```
pkoutsovasilis, line 124: Rreordering the annotation (e.g. "annotation-b,annotation-a" → "annotation-a,annotation-b") produces a different hash and triggers a spurious pod restart. I would propose here and in all other controllers that follow the same pattern to use `[]byte(strings.Join(resource.DownwardNodeLabels(), ","))` instead of `[]byte(resource.Annotations[commonv1.DownwardNodeLabelsAnnotation])`

pkg/controller/common/nodelabels/nodelabels.go
```diff
@@ -0,0 +1,136 @@
...
+	podAnnotations, err := getPodAnnotations(&pod, expectedLabels, node.Labels)
+	if err != nil {
+		return err
+	}
+	if len(podAnnotations) == 0 {
+		return nil
+	}
+	ulog.FromContext(ctx).Info(
+		"Setting Pod annotations from node labels",
+		"namespace", pod.Namespace,
```
pkoutsovasilis, line 81: namespace is injected into the context logger at pkg/controller/common/controller.go:49 via NewReconciliationContext, so here it is duplicated

pkg/controller/common/nodelabels/exposed.go
```diff
@@ -0,0 +1,66 @@
...
+			continue
+		}
+		errs = append(errs, field.Invalid(
+			field.NewPath("metadata").Child("annotations", commonv1.DownwardNodeLabelsAnnotation),
+			nodeLabel,
+			NotAllowedNodesLabelMsg,
+		))
+	}
+	return errs
+}
```
pkoutsovasilis, line 66: let's add exposed_test.go with our expectations of this func:
- annotation set, no exposed-node-labels configured → validation error
- annotation set, label matches the policy → no error
- annotation set, label does not match the policy → validation error
- no annotation set → no error (no-op)

pkg/controller/common/volume/downward_api.go
```diff
@@ -65,3 +67,9 @@ func (d DownwardAPI) Volume() corev1.Volume {
 func (DownwardAPI) VolumeMount() corev1.VolumeMount {
 	return downwardAPIVolumeMount
 }
+
+// AnnotationsFilePath returns the absolute path of the file exposing the Pod annotations within the
+// downward API volume mount. It is only populated when the volume is built WithAnnotations(true).
+func (DownwardAPI) AnnotationsFilePath() string {
+	return fmt.Sprintf("%s/%s", volume.DownwardAPIMountPath, volume.AnnotationsFile)
+}
```
pkoutsovasilis, line 75: I understand that what I am about to say predates this PR, but now that more controllers (not only the elasticsearch one) utilise the DownwardAPI I believe that `DownwardAPIVolumeName` and `DownwardAPIMountPath` consts need to migrate in this package and not be defined in `elasticsearch/volume`

pkg/controller/elasticsearch/validation/node_labels.go
```diff
@@ -5,32 +5,14 @@
...
-}
-
-func (n NodeLabels) IsAllowed(nodeLabel string) bool {
-	for _, r := range n {
-		if r.MatchString(nodeLabel) {
-			return true
-		}
-	}
-	return false
+	return commonnodelabels.NewExposedNodeLabels(exposedNodeLabels)
```
pkoutsovasilis, line 17: do we need this file? can't we use directly the commonnodelabels package?

pkg/controller/maps/pod.go
```diff
@@ -115,6 +116,20 @@ func newPodSpec(ems emsv1alpha1.ElasticMapsServer, configHash string, meta metad
...
+	if ems.HasDownwardNodeLabels() {
+		downwardAPIVolume := commonnodelabels.DownwardAPIVolume()
+		waitInit, err := commonnodelabels.WaitForAnnotationsInitContainer(ems.DownwardNodeLabels())
+		if err != nil {
+			return corev1.PodTemplateSpec{}, err
+		}
+		builder = builder.
+			WithVolumes(downwardAPIVolume.Volume()).
+			WithInitContainers(waitInit)
+	}
```
pkoutsovasilis, line 131: WithInitContainerDefaults() is called way above, which means that the Wait init container won't inherit the image

pkg/controller/packageregistry/pod.go
```diff
@@ -128,6 +129,20 @@ func newPodSpec(epr eprv1alpha1.PackageRegistry, configHash string, meta metadat
...
+	if epr.HasDownwardNodeLabels() {
+		downwardAPIVolume := commonnodelabels.DownwardAPIVolume()
+		waitInit, err := commonnodelabels.WaitForAnnotationsInitContainer(epr.DownwardNodeLabels())
+		if err != nil {
+			return corev1.PodTemplateSpec{}, err
+		}
+		builder = builder.
+			WithVolumes(downwardAPIVolume.Volume()).
+			WithInitContainers(waitInit)
+	}
```
pkoutsovasilis, line 144: WithInitContainerDefaults() is called way above, which means that the Wait init container won't inherit the image

pkg/controller/elasticsearch/validation/node_labels.go
pkoutsovasilis, line 17: do we need this file at all?

pkoutsovasilis (commented):
Thanks for working through my comments @orangecola , I did a second pass and left some new ones. More importantly, I did some manual testing and I have found that apm-server container image doesn't have grep installed so this approach will never work. Even worse, for some image variants of apm-server, namely wolfi, not even a shell is available within the image. For that reason I have two requests:
* Drop apm server support from this PR
* We need some e2e tests to validate that the downward-node-labels mechanism is working and thus make sure that future component image changes won't break it (We can handle this in a follow-up PR)

pkoutsovasilis:
Hi again @orangecola 👋  thank you for your patience with the review process and for working through all the feedback. After discussing the latest findings internally, we've realized that supporting this feature properly requires a deeper refactoring - specifically around the init container approach and its silent dependency on the component images having a shell and certain utilities available (which breaks for lean/wolfi variants). Thus we'll take it from here and complete the extra work required. We are really grateful and appreciate the effort you put into this! 🙇

pkoutsovasilis:
as described above, closing this one in favour of https://github.com/elastic/cloud-on-k8s/pull/9583. Once again thanks for the contribution @orangecola 🙇

PR #9369, opened 2026-04-17
Remove `context.WithTimeout` in favor of `context.WithCancel` for local e2e test runs

dtuck9:
## The Problem
When running the e2e tests with the `--local` flag and if/when a persistent error is encountered while running an e2e test within the `Eventually` function that ultimately surpasses the configured test timeout, the root cause error is lost and only the timeout is output. This makes troubleshooting failing tests extremely difficult.

## Why?
Most tests use the `Eventually` function, which uses the `ctx.TestTimeout` as a default timeout:
https://github.com/elastic/cloud-on-k8s/blob/7da7d3806837e52de0f3c1ec46788b9d72e949b0/test/e2e/test/utils.go#L79-L82

The `ctx.TestTimeout` is the value of the `--test-timeout` parameter

https://github.com/elastic/cloud-on-k8s/blob/7da7d3806837e52de0f3c1ec46788b9d72e949b0/test/e2e/cmd/run/run.go#L178

which is set by the `TEST_TIMEOUT` env var:

https://github.com/elastic/cloud-on-k8s/blob/7da7d3806837e52de0f3c1ec46788b9d72e949b0/Makefile#L496

Using this timeout, the `Eventually` function calls `UntilSuccess`:

https://github.com/elastic/cloud-on-k8s/blob/7da7d3806837e52de0f3c1ec46788b9d72e949b0/test/e2e/test/utils.go#L85-L96

which in turn calls `retry.UntilSuccess`:

https://github.com/elastic/cloud-on-k8s/blob/7da7d3806837e52de0f3c1ec46788b9d72e949b0/pkg/utils/retry/retry.go#L28-L61

Giving each individual test the duration of `TestTimeout` to succeed. This is where the persistent error gets surfaced and _ideally_ logged.

**However, the `ctx.TestTimeout` is also used for the entire test run timeout**  
https://github.com/elastic/cloud-on-k8s/blob/7da7d3806837e52de0f3c1ec46788b9d72e949b0/test/e2e/cmd/run/run.go#L501

Once the timeout is surpassed here, the `ctx.Done()` channel receives the an event, and then sends a kill:

https://github.com/elastic/cloud-on-k8s/blob/7da7d3806837e52de0f3c1ec46788b9d72e949b0/test/e2e/cmd/run/run.go#L521-L527

**Which in turn short-circuits the running test whose error never gets output.**

### Proposed
Update the `Eventually` function to use a hard-coded `DefaultTestTimeout` of 5 minutes as the simplest/quickest approach. 5 minutes is shorter than the overall test run, while still being long enough for each individual test step to complete. Because this is only an issue when running the e2e tests locally, this should be sufficient in most cases, though it does not completely eliminate the risk of this edge case occurring. It could also impact the nightly run if we find that 5 minutes is not enough for individual test steps running remotely, and could potentially cause more failures.

### Other options considered
- Configure an error channel in the test context to be read from when the context is cancelled in run.go. This would add some complexity and surface the error, but the original error would not be easily tracked to a particular test. The minimal benefit does not warrant the complexity and would imo cause different troubleshooting pains.
- Pass the `context.Context` into the `UntilSuccess` function, so that it can also know if the timeout has expired and handle it more gracefully to return the `lastErr` as expected. This would touch a lot of files including installing kyverno and reading from vault. It may be the most robust way of solving this problem, but again, the complexity does not seem warranted unless the 5 minute timeout proves to be problematic.
- Introduce a separate env var and parameter for individual test timeouts.


prodsecmachine:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/858dae2c-0806-42d9-8dfb-c95fc7ce294f?refs=a7231588e516e79050a31fc77fdbb63ac8542109&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/858dae2c-0806-42d9-8dfb-c95fc7ce294f/license?refs=a7231588e516e79050a31fc77fdbb63ac8542109&source=prChecksComment) |

dtuck9:
buildkite test this -f p=gke

barkbay:
buildkite test this -f p=gke-autopilot,t=TestAutopilot

[ [build](https://buildkite.com/elastic/cloud-on-k8s-operator/builds/14446) ]

barkbay (commented):
Makes sense as a minimal fix. However I think the same race is even more present for the callers that compute their step timeout as `ctx.TestTimeout * 2`:

- `test/e2e/test/checks/common.go`: https://github.com/elastic/cloud-on-k8s/blob/d17eb8d31bca4df8679d76151d5fcb64b3b3510f/test/e2e/test/checks/common.go#L50
- `test/e2e/test/agent/steps.go`, only on Autopilot: https://github.com/elastic/cloud-on-k8s/blob/d17eb8d31bca4df8679d76151d5fcb64b3b3510f/test/e2e/test/agent/steps.go#L90-L93
- `test/e2e/stack_test.go`: https://github.com/elastic/cloud-on-k8s/blob/d17eb8d31bca4df8679d76151d5fcb64b3b3510f/test/e2e/stack_test.go#L138-L142

In `--local` mode with the default `TEST_TIMEOUT=15m`, those steps have a 30 min retry deadline against a 15 min watchdog. The watchdog always fires first, so when one of those steps is the failing one, `lastErr` is always lost, which is exactly the symptom this PR is trying to remove for `Eventually`. After this PR, plain `Eventually()` is fixed but those three callers regress further relative to it (5 min cap vs 30 min that is never reached anyway).

I wonder if it would make sense to migrate those callers in the same PR? A flat `DefaultLongStepTimeout = 2 * DefaultStepTimeout` (10 min) is tempting but probably too tight: the `* 2` was introduced in #3943 to give stack upgrades 60 min, and is 30 min today (since `TEST_TIMEOUT` was later reduced to 15 min). Going to a flat 10 min would tighten it further with no measurement to back the choice, and stack upgrades (rolling ES restarts + version bumps) can legitimately be slow on cold nodes.

A safer option that still fixes the local-mode race is to derive the long budget from `ctx.TestTimeout` while keeping it strictly below the watchdog with a small buffer:

```go
const (
    DefaultRetryDelay  = 3 * time.Second
    DefaultStepTimeout = 5 * time.Minute
    // EventuallyBuffer is reserved at the end of ctx.TestTimeout so that any
    // per-step retry deadline fires before the local-mode watchdog SIGKILLs the
    // run.sh process group, ensuring lastErr is surfaced.
    EventuallyBuffer = 2 * time.Minute
)

// LongStepTimeout returns a retry deadline for slow steps (e.g. CheckPods,
// "wait for ES green", stack upgrades). It is guaranteed to be strictly less
// than ctx.TestTimeout so the local-mode race cannot lose lastErr, and it
// scales with TEST_TIMEOUT so users on slow clusters can still raise the
// ceiling via --test-timeout.
func LongStepTimeout(ctx Context) time.Duration {
    long := ctx.TestTimeout - EventuallyBuffer
    if long < 2*DefaultStepTimeout {
        long = 2 * DefaultStepTimeout
    }
    return long
}
```

Then `ctx.TestTimeout * 2` in `checks/common.go`, `agent/steps.go` and `stack_test.go` becomes `test.LongStepTimeout(test.Ctx())`. The properties this gives us:

- Default `TEST_TIMEOUT=15m`: long steps get 13 min, plain `Eventually` 5 min, both strictly under the 15 min watchdog. The local-mode race is fixed for all callers.
- Today's CI behaviour with `TEST_TIMEOUT=30m`: long steps get 28 min (close to the historical 30 min), so we do not silently tighten budgets that older commits chose carefully.
- Slow environments (Copilot, Autopilot under load, stateless): bumping `--test-timeout` lifts the long-step budget proportionally, without ever breaking the inner-fires-first invariant.

More broadly (and outside the scope of this PR), I think the root cause is that `ctx.TestTimeout` is overloaded: it is both the local-mode wall-clock watchdog and the per-step retry deadline, with the same value. Any caller that picks a per-step deadline `>= ctx.TestTimeout` will swallow its own error in local mode.

dtuck9:
> More broadly (and outside the scope of this PR), I think the root cause is that ctx.TestTimeout is overloaded: it is both the local-mode wall-clock watchdog and the per-step retry deadline, with the same value. Any caller that picks a per-step deadline >= ctx.TestTimeout will swallow its own error in local mode.

@barkbay Agreed. I had also considered introducing a separate parameter to decouple the individual test timeout from the overall `TestTimeout` but had opted for the easier approach initially. 

It sounds like that may be worthwhile though and I can look into that -- there could be some simple validation to ensure this value is always set to something less than `ctx.TestTimeout`.

dtuck9:
After thinking about it some more and discussing with @pebrc, I think I will actually refactor the `runTestsLocally` function to avoid the timeout in the first place for local test runs, in favor of just listening for `SIGINT` or `SIGTERM`. This will keep any changes scoped to local runs rather than potentially impacting CI, and will remove the race entirely while still being a relatively simple and straightforward change.

dtuck9:
buildkite test this -f p=gke-autopilot,t=TestAutopilot

[[build](https://buildkite.com/elastic/cloud-on-k8s-operator/builds/14459#_)]

barkbay (approved):
LGTM 👍 

One minor nit: now that `--test-timeout` only bounds a single retry loop and no longer the whole run, the current help text ("Timeout before failing a test") is a bit misleading. Maybe something like:

```diff
diff --git a/test/e2e/cmd/run/command.go b/test/e2e/cmd/run/command.go
index cf4c94657..3f2fdd781 100644
--- a/test/e2e/cmd/run/command.go
+++ b/test/e2e/cmd/run/command.go
@@ -100,7 +100,7 @@ func Command() *cobra.Command {
        cmd.Flags().StringVar(&flags.scratchDirRoot, "scratch-dir", "/tmp/eck-e2e", "Path under which temporary files should be created")
        cmd.Flags().StringVar(&flags.testRegex, "test-regex", "", "Regex to pass to the test runner")
        cmd.Flags().StringVar(&flags.testRunName, "test-run-name", randomTestRunName(), "Name of this test run")
-       cmd.Flags().DurationVar(&flags.testTimeout, "test-timeout", 15*time.Minute, "Timeout before failing a test")
+       cmd.Flags().DurationVar(&flags.testTimeout, "test-timeout", 15*time.Minute, "Maximum duration of a single retry loop inside a test step (e.g. Eventually); does not bound the overall test or run duration")
        cmd.Flags().StringVar(&flags.pipeline, "pipeline", "", "E2E test pipeline name")
        cmd.Flags().StringVar(&flags.buildNumber, "build-number", "", "E2E test build number")
        cmd.Flags().StringVar(&flags.provider, "provider", "", "E2E test infrastructure provider")
```

(Renaming the flag to something like `--step-timeout` would make this even clearer, but it would touch the Makefile, CLI flag, `test.Context` field and every caller for a purely cosmetic change ... so probably not worth doing in this PR.)

PR #9391, opened 2026-04-21
[AutoOps] Add `sending_queue` configuration to configmap

pickypg:
Adding the sending queue configuration allows the exporter to batch the request so that it does not exceed 4 MiB.

`otlp_http` has a default 4 MB limit, so large clusters can exceed that and cause their entire batch to be rejected.

### DIFF
--- a/pkg/controller/autoops/configmap.go
+++ b/pkg/controller/autoops/configmap.go
@@ -78,7 +78,7 @@ exporters:
     headers:
       Authorization: "AutoOpsToken ${env:AUTOOPS_TOKEN}"
     endpoint: ${env:AUTOOPS_OTEL_URL}
-	sending_queue:
+    sending_queue:
       batch:
         flush_timeout: 11s
         min_size: 1048576
@@ -87,7 +87,7 @@ exporters:
       block_on_overflow: true
       enabled: true
       queue_size: 52428800
-	  sizer: bytes
+      sizer: bytes
 service:
   extensions: [healthcheckv2]
   pipelines:
--- a/pkg/controller/autoops/configmap.go
+++ b/pkg/controller/autoops/configmap.go
@@ -78,6 +78,16 @@ exporters:
     headers:
       Authorization: "AutoOpsToken ${env:AUTOOPS_TOKEN}"
     endpoint: ${env:AUTOOPS_OTEL_URL}
+	sending_queue:
+      batch:
+        flush_timeout: 11s
+        min_size: 1048576
+        max_size: 4194304
+        sizer: bytes
+      block_on_overflow: true
+      enabled: true
+      queue_size: 52428800
+	  sizer: bytes
 service:
   extensions: [healthcheckv2]
   pipelines:
--- a/pkg/controller/autoops/configmap_test.go
+++ b/pkg/controller/autoops/configmap_test.go
@@ -23,6 +23,16 @@ exporters:
     endpoint: ${env:AUTOOPS_OTEL_URL}
     headers:
       Authorization: AutoOpsToken ${env:AUTOOPS_TOKEN}
+    sending_queue:
+      batch:
+        flush_timeout: 11s
+        min_size: 1048576
+        max_size: 4194304
+        sizer: bytes
+      block_on_overflow: true
+      enabled: true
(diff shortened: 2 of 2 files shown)


prodsecmachine:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/5b5a6207-63b2-47af-a7e5-b47e0be6c28a?refs=0bd40095dbfa385531b843d55a59972e78100bd5&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/5b5a6207-63b2-47af-a7e5-b47e0be6c28a/license?refs=0bd40095dbfa385531b843d55a59972e78100bd5&source=prChecksComment) |

pkoutsovasilis:
@pickypg this feels to me more like an `enhancement` than a `bug`, any particular reason you deem this as a `bug`?

pkoutsovasilis:
buildkite test this -f p=gke,t=.*,E2E_TAGS=autoops -m s=9.3.3,s=9.4.0-SNAPSHOT

pkoutsovasilis (approved):
LGTM

@pickypg one more question; this configuration of `otlp_http` is supported in all versions later versions, starting from `9.2.1`, right?

pickypg:
> this feels to me more like an enhancement than a bug, any particular reason you deem this as a bug?

Because without this change, an unpredictable number of payloads get rejected from Elastic Agent and AutoOps won't work for some end users.

> this configuration of otlp_http is supported in all versions later versions, starting from 9.2.1, right?

No, but I did not actually change that here (it's still `otlphttp`). I just used the new name out of habit. The new name didn't become available until 9.2.6+ and 9.3.1+ (I checked). For anyone interested, this can be validated easily by running OTel mode with this exporter set via the CLI: `--set 'exporters::otlp_http::endpoint=http://localhost:4318'` (whether an OTel Collector is really running there or not, the `otlp_http` will either work or not immediately).

pkoutsovasilis:
> Because without this change, an unpredictable number of payloads get rejected from Elastic Agent and AutoOps won't work for some end users.

if that's the case then I would expect this to not be configurable thus it never breaks?! 🤔 either way bug here means that you address a bug in ECK operator which still I don't think it falls under this case (@pebrc thoughts on the label?)


> No, but I did not actually change that here (it's still otlphttp). I just used the new name out of habit. The new name didn't become available until 9.2.6+ and 9.3.1+ (I checked). For anyone interested, this can be validated easily by running OTel mode with this exporter set via the CLI: --set 'exporters::otlp_http::endpoint=http://localhost:4318' (whether an OTel Collector is really running there or not, the otlp_http will either work or not immediately).

ok sorry I may have phrased that poorly, let me give it another shot 🙂 Are the config changes introduced here valid for all versions starting from `9.2.1`?

pkoutsovasilis:
buildkite test this -f p=gke,t=.*,E2E_TAGS=autoops -m s=9.2.1

pickypg:
> if that's the case then I would expect this to not be configurable thus it never breaks?! 🤔 either way bug here means that you address a bug in ECK operator which still I don't think it falls under this case

If you don't want to consider this a bug for the ECK operator itself, that's fine by me. It's just semantics. It's a bug for the ECK-defined configuration of Elastic Agent running AutoOps though.

> ok sorry I may have phrased that poorly, let me give it another shot 🙂 Are the config changes introduced here valid for all versions starting from 9.2.1?

Ah, yes! We actually use a variation of this all the way back in 9.2.1:

https://github.com/elastic/elastic-agent/blob/1079d6f853dd008cce88f3cd3027aa253f33e3bd/internal/pkg/otel/samples/linux/autoops_es.yml#L39-L47

pkoutsovasilis:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|3.4|[](https://github.com/elastic/cloud-on-k8s/pull/9396)|

### Questions ?

PR #9398, opened 2026-04-23
[#9293] Stateless/Agent/Beats changes for new `pause-orchestration` annotation

### LINKED ISSUE #9293: Pause orchestration: stateless controllers
Part of #9250. Depends on #9292 (ES implementation, which defines the annotation and helper).

## Scope

Add `pause-orchestration` support to stateless Deployment-based controllers:
- Kibana
- APM Server
- EnterpriseSearch
- Agent
- Beats
- Maps
- PackageRegistry

## Approach

These controllers have simpler reconciliation loops than Elasticsearch. The pause should skip Deployment spec updates while continuing housekeeping (certificate rotation, secret reconciliation, service reconciliation, status updates).

The exact cut point will vary per controller but the pattern is the same: after shared/housekeeping reconciliation, check `IsOrchestrationPaused()` and return early before applying Deployment spec changes.

Since these are stateless Deployments, Kubernetes handles pod rescheduling during node drains gracefully. The main benefit here is preventing spec-driven rolling restarts (e.g. from image version bumps) from compounding with node maintenance, and ensuring certificate rotation continues.

## Code pointers

All controllers currently use `IsUnmanaged()` as the first check:
- `pkg/controller/kibana/controller.go`
- `pkg/controller/apmserver/controller.go`
- `pkg/controller/enterprisesearch/enterprisesearch_controller.go`
- `pkg/controller/agent/controller.go`
- `pkg/controller/beat/controller.go`
- `pkg/controller/maps/controller.go`
- `pkg/controller/packageregistry/controller.go`

dtuck9:
## Pause orchestration: stateless controllers
                                                                                                                                                                                                                                         
  Closes #9293.                                                                                                                                                                                                                          
                                       
  Depends on #9330, https://github.com/elastic/cloud-on-k8s/pull/9408, and #9417 
                                                                                                                                                                                                                                         
  ### Summary                                                                                                                                                                                                                            
                                                                                                                                                                                                                                         
Extends the `eck.k8s.elastic.co/pause-orchestration` annotation introduced for Elasticsearch in #9330 to the five Deployment-based stateless controllers: Kibana, APM Server, Enterprise Search, Elastic Maps Server, and Package Registry.                                                                                                                                                      
                                                                                                                                                                                                                                       
  When the annotation is set to `"true"`, the controller completes all housekeeping steps (certificate rotation, service reconciliation, config secret updates) but exits before applying any changes to the Deployment spec. This prevents spec-driven rolling restarts — such as those triggered by an image version update — from disrupting nodes during maintenance, while keeping TLS certificates and configuration secrets current.                                                                                                                                                                                                         
                                                                                                                                                                                                                                         
  Agent and Beats are out of scope for this PR: their status types do not embed `commonv1.DeploymentStatus`, and they support multiple pod vehicles (`DaemonSet`, `StatefulSet`) that would each need equivalent handling.                                                                                                                                                                                 
                                                                                                                                                                                                                                         
  ### Changes                                                                                                                                                                                                                                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                         
  **Shared helpers (`pkg/controller/common/deployment/reconcile.go`)**                                                                                                                                                                   
                                                                                                                                                                                                                                         
  - `HasPendingChanges` — compares the template hash of the expected Deployment against the cluster state to determine whether a spec change is pending.                                                                                                                                                           
  - `SetPausedConditionAndEmitEvent` — calls `HasPendingChanges`, writes an `OrchestrationPaused: True` condition to the provided `*Conditions`, and emits a Kubernetes warning event. Takes `*commonv1alpha1.Conditions` so each controller passes the address of its own status field with no interface boilerplate.                                                                                                                                                                                                                         
  - `MaybeResetPausedCondition` — sets `OrchestrationPaused: False` on resume, but only if the condition already exists (no-op otherwise).                                                                                                                                                                              
                                                                                                                                                                                                                                         
  **Per-controller cut points**                                                                                                                                                                                                        
                                                                                                                                                                                                                                         
  Each controller checks `common.IsOrchestrationPaused` after building the expected `Deployment` but before calling `deployment.Reconcile`. On the unpaused path, `MaybeResetPausedCondition` clears any lingering condition from a previous pause.
                                                                                                                                                                                                                                         
  | Controller | File | Cut point |
  |---|---|---|                                                                                                                                                                                                                          
  | Kibana | `pkg/controller/kibana/driver.go` | After `deployment.New`, before `deployment.Reconcile` |                                                                                                                               
  | APM Server | `pkg/controller/apmserver/deployment.go` | After `deployment.New`, before `deployment.Reconcile` |                                                                                                                      
  | Enterprise Search | `pkg/controller/enterprisesearch/deployment.go` | After `deployment.New`, before `deployment.Reconcile` |
  | Elastic Maps Server | `pkg/controller/maps/controller.go` | After `deployment.New`, before `deployment.Reconcile` |                                                                                                                  
  | Package Registry | `pkg/controller/packageregistry/controller.go` | After `deployment.New`, before `deployment.Reconcile` |                                                                                                                 
  | Agent | `pkg/controller/agent/reconcile.go` | Separate resource build from reconcile, then cut point after building the expected resource, before `reconciliationFunc` |                                                                                                                 
  | Beats | `pkg/controller/agent/reconcile.go` | Similar to agent, after building the expected resource, before `reconciliationFunc` |
                                                                                                                                                                                                                                         
  **CRDs**                                                                                                                                                                                                                             
                                                                                                                                                                                                                                         
  Regenerated schema for the five affected CRDs to include the new `conditions` field in the status subresource.

  ### Testing

  - [X] Unit tests for `HasPendingChanges`, `SetPausedConditionAndEmitEvent`, and `MaybeResetPausedCondition`
  - [X] e2e test -- extension of `TestPauseOrchestration`
  - [ ] Manual test on k3d: set annotation → trigger a spec change → verify Deployment unchanged and condition visible via `kubectl get <resource> -o yaml` → remove annotation → verify spec change is applied and condition cleared

### AI Disclaimer
Claude Code used for design.


prodsecmachine:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/1977071d-6617-4d6f-b364-e0e214049a5a?refs=60a57bbd0f8a3deb6cb07c33ff62a301b3443971&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/1977071d-6617-4d6f-b364-e0e214049a5a/license?refs=60a57bbd0f8a3deb6cb07c33ff62a301b3443971&source=prChecksComment) |

dtuck9:
buildkite test this -f p=gke,t=TestPauseOrchestration

pkg/apis/apm/v1/labels.go
```diff
@@ -8,10 +8,15 @@ import (
 	commonv1 "github.com/elastic/cloud-on-k8s/v3/pkg/apis/common/v1"
 )
 
+const (
+	// Type represents the APM Server type
+	Type = "apm-server"
+)
```
dtuck9: Drive-by fix for an implementation inconsistency.

test/e2e/es/pause_orchestration_test.go
```diff
@@ -1,232 +0,0 @@
-// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
```
dtuck9, line 1: Moved to test/e2e and relabeled with `mixed` as it now covers all application types.

test/e2e/test/apmserver/checks_apm.go
```diff
@@ -380,7 +380,7 @@ func (c *apmClusterChecks) CheckAgentConfiguration(apm apmv1.ApmServer, k *test.
 					return err
 				}
 
-				uri := "/api/apm/settings/agent-configuration"
+				uri := "/api/apm/settings/agent-configuration?overwrite=true"
```
dtuck9, line 383: This test was designed to only call this once, but `TestPauseOrchestration` calls the `CheckTestSteps` function multiple times for each app. Added this to avoid getting a 400 (failure) on secondary calls.

dtuck9:
buildkite test this -f p=gke,t=TestPauseOrchestration -m s=8.19.14,s=9.3.4,s=7.17.29

[[build](https://buildkite.com/elastic/cloud-on-k8s-operator/builds/14774)]

pkg/controller/beat/common/reconcile_test.go
```diff
@@ -25,7 +25,7 @@ func Test_reconcileDaemonSet(t *testing.T) {
 		ServiceAccountName: "my-service-account",
 		HostNetwork:        true,
 	}
-	int10 := intstr.FromInt(10)
+	int10 := intstr.FromInt32(10)
```
dtuck9, line 28: Drive-by fix: `FromInt` was deprecated in favor of `FromInt32`.

dtuck9:
buildkite test this -f p=gke,t=TestPauseOrchestration -m s=8.19.14,s=9.3.4,s=7.17.29

dtuck9:
buildkite test this -f p=gke,t=TestPauseOrchestration -m s=8.19.14,s=9.3.4,s=7.17.29

[[build](https://buildkite.com/elastic/cloud-on-k8s-operator/builds/14780)]

dtuck9:
The e2e tests are flaky I think because of resource issues. Trying to see how to increase resources on the cluster, otherwise I may need to break these tests up (which would also be ok, I suppose, just a little more work).

dtuck9:
buildkite test this -f p=gke,t=TestPauseOrchestration_APM -m s=8.19.14,s=9.3.4,s=7.17.29

[[build](https://buildkite.com/elastic/cloud-on-k8s-operator/builds/14789)]

dtuck9:
I am going to go ahead and submit this for review, specifically asking for recommendations on the resource issue -- even a test with just Elasticsearch, Kibana, and APM is flaky particularly around pod deletion and recreation.

pebrc:
## Flakiness root cause analysis (Phase 5 pod deletion)

Looked at the diagnostics from the failing `TestPauseOrchestration_APM` run. Here's what we found.

### What happens

The failing step is `APM-ES services should have endpoints` — `es-master-0` is Running but never passes its Kubernetes readiness probe (port 8080 check).

The chain of events in the failing run:

1. **Phase 3** (rolling restart while unpaused): ECK uses the Node Shutdown API to coordinate the restart of `es-master-0`, leaving a RESTART shutdown record with status `COMPLETE` in ES cluster state for node ID `MNUwmR-6QuG3eGkbsQ9aAg`.

2. **Phase 5** (pod deleted while paused): the pod is deleted. The StatefulSet controller recreates it. The operator reconciles actively (iters 122–131), issuing a new cert and updating unicast hosts. In each of those iterations, `reconcileCriticalStepsWhilePaused` calls `nodeShutdown.Clear` with the `OnlyNodesInCluster` filter — but `es-master-0`'s node ID is not in `nodeNameToID` yet (old pod is gone, new pod hasn't joined ES yet), so the RESTART COMPLETE record is not cleared.

3. **Reconciliations stop at 22:31:39** (iter 131). All Kubernetes resources have settled: cert issued, unicast hosts updated, pod annotations stamped. The pod is now `Running/NotReady` — no further Kubernetes status transitions occur. The cluster health stays `green` (2 data nodes hold all shards), so the health-change observer fires no event either.

4. **22:32:01**: the new `es-master-0` pod joins the ES cluster — with the **same** node ID (`MNUwmR-6QuG3eGkbsQ9aAg`, reused from the PVC). Upon receiving the cluster state, ES finds its own RESTART COMPLETE record and immediately starts `"Starting node shutdown sequence for ML"`.

5. ES 8.x keeps port 8080 closed while in shutdown mode. The readiness probe fails indefinitely. No new Kubernetes resource changes occur (pod was already `Ready=False`), cluster stays green — so no further reconciliations run to call `Clear` and remove the now-clearable record.

The `node_shutdown_metadata: {}` in the collected cluster state confirms the record was eventually cleared (likely by the next periodic reconciliation or a later watch event), but by then ES had already started the ML shutdown sequence and doesn't automatically reopen port 8080.

### Why it's flaky

The window between "old pod deleted" and "new pod joins cluster" must overlap with a reconciliation that can clear the record. If all post-deletion reconciliations complete before the new pod rejoins, the `OnlyNodesInCluster` filter blocks every clearing attempt, and by the time the node ID re-enters the cluster there's no trigger left to run `Clear`.

### Suggested fix

In `reconcileCriticalStepsWhilePaused`, after calling `Clear`, check whether any StatefulSet pods are not yet ready. If so, return a requeue result (e.g. 10–30s) so the operator keeps running. This ensures that when the new pod rejoins the cluster, the next reconciliation can clear the now-valid COMPLETE record before ES has a chance to act on it.

Currently the only requeue in that path is the 5-minute requeue for pending spec changes (line 201). A not-ready-pod check would cover the pod-lifecycle case without requiring a spec change.

dtuck9:
Thanks @pebrc, definitely an interesting edge case.

dtuck9:
buildkite test this -f p=gke,t=TestPauseOrchestration_APM -m s=8.19.14,s=9.3.4,s=7.17.29

[[build](https://buildkite.com/elastic/cloud-on-k8s-operator/builds/14810)]

pkg/controller/common/pause_orchestration_test.go
```diff
@@ -73,3 +91,419 @@ func TestIsOrchestrationPaused(t *testing.T) {
...
+			} else {
+				c = k8s.NewFakeClient(tt.existingObjs...)
+			}
+			recorder := toolsevents.NewFakeRecorder(10)
+
+			err := SetPausedConditionAndEmitEvent(context.Background(), c, recorder, tt.parentObject, tt.expectedVehicle)
+
+			if tt.wantError {
+				assert.Error(t, err)
+				assert.Empty(t, tt.parentObject, "no condition should be set on error")
```
pebrc: `tt.parentObject` is `nil` in the `client error` test case (struct default zero value), so `assert.Empty(t, nil)` is a tautology — this doesn't verify "no condition should be set on error." Either:

- Pass a non-nil empty Kibana as `parentObject` and assert `assert.Empty(t, tt.parentObject.Conditions())`, or
- Drop the assertion and just check `assert.Error(t, err)`.

Separately, when `parentObject` is nil and `hasPendingChanges` happens to succeed, `parent.MergeConditions(...)` at line 73 would nil-deref — the current test passes only because the failing client errors first. A defensive nil check at the top of `SetPausedConditionAndEmitEvent` might be warranted.

pkg/controller/elasticsearch/driver/stateful/driver.go
```diff
@@ -186,16 +186,22 @@ func (d *Driver) reconcileCriticalStepsWhilePaused(
 			nodeShutdown.OnlyNodesInCluster,
 			nodeShutdown.OnlyNonTerminatingNodes(terminatingNodes),
 		))
+
+		for _, pod := range actualPods {
+			if !k8s.IsPodReady(pod) {
+				return results.WithRequeue(reconciler.DefaultRequeue)
+			}
+		}
```
pebrc: This requeue returns before `hasPendingSpecChanges` runs, so the `OrchestrationPaused` condition message can go stale (still showing "no pending changes" when there are pending changes, or vice versa) while pods are recovering. Also only inside the `supportsNodeShutdown` branch — older ES versions skip this requeue entirely.

The commit message says "Check that ES pods are ready; if not requeue" but doesn't motivate *why*. A short comment explaining the scenario (e.g. "after `nodeShutdown.Clear`, wait for pods to come back ready before evaluating pending changes — otherwise we could observe stale state") would help future readers.

pkg/controller/enterprisesearch/deployment.go
```diff
@@ -11,27 +11,46 @@ import (
...
+		// annotation is enabled, we return the existing deployment to avoid later setting the status health to "red"
+		// in the call to the common.DeploymentStatus function.
+		err = common.SetPausedConditionAndEmitEvent(ctx, r.K8sClient(), r.recorder, ent, &deploy)
+		if err != nil {
+			return appsv1.Deployment{}, err
+		}
+		var existing = appsv1.Deployment{}
+		if err := r.Client.Get(ctx, k8s.ExtractNamespacedName(&deploy), &existing); err != nil {
+			return appsv1.Deployment{}, err
+		}
```
pebrc: **Bug — pause on a never-reconciled EnterpriseSearch errors indefinitely.**

If the user sets `eck.k8s.elastic.co/pause-orchestration: "true"` on a brand-new EnterpriseSearch (or on one whose deployment was manually deleted), this `Get` returns `NotFound` and the function returns an error. Back in `doReconcile`:

```go
deploy, err := r.reconcileDeployment(ctx, &ent, configHash, meta)
if err != nil {
    return results.WithError(...), status  // <-- status is from newStatus(ent), pre-mutation
}
```

The `status` snapshot was taken before `SetPausedConditionAndEmitEvent` mutated `ent.Status.Conditions`, so the `OrchestrationPaused=True` condition is never persisted. Every requeue repeats the loop and the user sees only reconciliation errors.

Suggestion: treat `IsNotFound` here as "nothing to preserve" — return an empty deployment (or skip the deployment-status calculation entirely when paused) rather than an error.

pkg/controller/maps/controller.go
```diff
@@ -352,19 +352,36 @@ func buildConfigHash(c k8s.Client, ems emsv1alpha1.ElasticMapsServer, configSecr
...
+		// annotation is enabled, we return the existing deployment to avoid later setting the status health to "red"
+		// in the call to the common.DeploymentStatus function.
+		err = common.SetPausedConditionAndEmitEvent(ctx, r.K8sClient(), r.recorder, ems, &deploy)
+		if err != nil {
+			return appsv1.Deployment{}, err
+		}
+		var existing = appsv1.Deployment{}
+		if err := r.Client.Get(ctx, k8s.ExtractNamespacedName(&deploy), &existing); err != nil {
+			return appsv1.Deployment{}, err
+		}
```
pebrc: Same shape as the EnterpriseSearch deployment.go comment: `NotFound` here on a never-reconciled Maps resource turns the paused branch into a perpetual error loop, and the condition update is lost because the caller returns the pre-mutation `status` snapshot.

pkg/controller/packageregistry/controller.go
```diff
@@ -229,6 +229,13 @@ func (r *ReconcilePackageRegistry) doReconcile(ctx context.Context, epr eprv1alp
 	}
 
 	expectedDp := deployment.New(deployParams)
+	if common.IsOrchestrationPaused(&epr) {
+		err = common.SetPausedConditionAndEmitEvent(ctx, r.K8sClient(), r.recorder, &epr, &expectedDp)
+		status.Conditions = epr.Status.Conditions
+		return results.WithError(err), status
+	}
+
+	common.MaybeResetPausedCondition(&epr.Status.Conditions)
```
pebrc: After `MaybeResetPausedCondition` reassigns `epr.Status.Conditions` to a new slice (paused=False), the local `status.Conditions` from `newStatus(epr)` still points to the pre-reset slice. On the happy path this is corrected at line 252 by `status.DeploymentStatus = deploymentStatus` (which deep-copies the updated `epr.Status.DeploymentStatus`), but if any intermediate step errors (`deployment.Reconcile`, `PodsMatchingLabels`, `common.DeploymentStatus`) the function returns the stale `status` and the reset never reaches the persisted status.

Consider mirroring the paused branch's explicit `status.Conditions = epr.Status.Conditions` here for symmetry — small and self-documenting.

pkg/controller/common/pause_orchestration.go
```diff
@@ -5,18 +5,141 @@
...
+	parent.MergeConditions(commonv1.Condition{
+		Type:               commonv1.OrchestrationPaused,
+		Status:             corev1.ConditionTrue,
+		LastTransitionTime: metav1.Now(),
+		Message:            msg,
+	})
+
+	k8s.EmitEvent(recorder, parent, corev1.EventTypeWarning,
+		events.EventReasonPaused, events.EventActionReconciliation, msg)
+	return nil
```
pebrc: This emits a `Warning Paused` event on every reconcile while paused, regardless of whether the condition or message actually changed. With the default requeue cadence a paused resource fills the event log over time.

Compare with the ES driver (`pkg/controller/elasticsearch/driver/stateful/driver.go:202-205`), which only emits when `hasPendingChanges` is true. Worth aligning — either emit only on transitions, or only when there are pending changes.

pkg/controller/common/pause_orchestration.go
```diff
@@ -5,18 +5,141 @@
...
+// These type-checks are defined here to avoid an import cycle between the API packages and this package, but also
+// provides a clearly defined expectation as far as the implementors. Elasticsearch is notably and intentionally absent
+// as its pause-orchestration handling is very specific to Elasticsearch as a result of the different node tiers.
+var _ ObjectWithConditions = (*kbv1.Kibana)(nil)
+var _ ObjectWithConditions = (*apmv1.ApmServer)(nil)
+var _ ObjectWithConditions = (*v1beta1.Beat)(nil)
+var _ ObjectWithConditions = (*eprv1alpha1.PackageRegistry)(nil)
+var _ ObjectWithConditions = (*maps.ElasticMapsServer)(nil)
+var _ ObjectWithConditions = (*agentv1alpha1.Agent)(nil)
+var _ ObjectWithConditions = (*entv1.EnterpriseSearch)(nil)
```
pebrc, line 178: **Architecture / N+1 pattern.** Adding a new ECK resource type that supports pause-orchestration today requires:

1. Adding `Conditions` to the status type, plus `MergeConditions`/`Conditions()` methods.
2. Open-coding ~3 lines (`IsOrchestrationPaused` check, `SetPausedConditionAndEmitEvent`, `MaybeResetPausedCondition`) inside its reconcile function.
3. Adding a line to this `var _ ObjectWithConditions = ...` block.

And the existing seven implementations diverge into five shapes (Kibana state-pointer, APM state-pointer, EnterpriseSearch existing-deployment workaround, PackageRegistry explicit `status.Conditions = epr.Status.Conditions`, Agent/Beats `params.Status` bridging). That's a lot of surface area for a feature that should be a one-liner per controller.

A single helper would consolidate this:

```go
func ReconcileDeploymentPausedAware(
    ctx context.Context, c k8s.Client, recorder ...,
    parent ObjectWithConditions, expected *appsv1.Deployment,
    owner client.Object,
) (appsv1.Deployment, error) {
    if IsOrchestrationPaused(parent) {
        return appsv1.Deployment{}, SetPausedConditionAndEmitEvent(ctx, c, recorder, parent, expected)
    }
    conds := parent.Conditions()
    MaybeResetPausedCondition(&conds)
    // (need to write back to parent — interface tweak)
    return deployment.Reconcile(ctx, c, *expected, owner)
}
```

Not a blocking concern, but worth considering before the divergent shapes calcify.

pkg/controller/common/pause_orchestration.go
```diff
@@ -5,18 +5,141 @@
...
+	orchestrationPausedIndex := conditions.Index(commonv1.OrchestrationPaused)
+	if orchestrationPausedIndex >= 0 {
+		*conditions = conditions.MergeWith(commonv1.Condition{
+			Type:               commonv1.OrchestrationPaused,
+			Status:             corev1.ConditionFalse,
+			LastTransitionTime: metav1.Now(),
+			Message:            PausedOrchestrationResumed,
+		})
+	}
+}
```
pebrc, line 139: **Asymmetric observability — no resume event.** Pause emits a `Warning Paused` event but resume emits nothing. Operators tailing events see only one side of the transition. Consider emitting a `Normal` event (e.g. `EventReasonResumed`) when the condition actually transitions from True → False here.

pkg/controller/common/pause_orchestration_test.go
pebrc: `tt.parentObject` is `nil` in the `client error` test case (struct default zero value), so `assert.Empty(t, nil)` is a tautology — this doesn't verify "no condition should be set on error." Either:

- Pass a non-nil empty Kibana as `parentObject` and assert `assert.Empty(t, tt.parentObject.Conditions())`, or
- Drop the assertion and just check `assert.Error(t, err)`.

Separately, when `parentObject` is nil and `hasPendingChanges` happens to succeed, `parent.MergeConditions(...)` at line 73 would nil-deref — the current test passes only because the failing client errors first. A defensive nil check at the top of `SetPausedConditionAndEmitEvent` might be warranted.

pkg/controller/elasticsearch/driver/stateful/driver.go
pebrc, line 198: The intent here is right but the placement deserves a comment, and there are three implementation concerns worth addressing.

**Why this check is needed (and the non-paused path doesn't need an equivalent).** In the non-paused case the rolling upgrade machinery in `upgrade.go` uses `IsPodReady` at every per-pod decision (lines 244, 434, 448) — readiness is intrinsic to the inner loop. The paused path is an observer with no inner loop, so once `expectationsSatisfied` returns true (which ignores pod *status* — only checks pod *names* per `PodReconciliationDone`'s docstring), the next thing is `hasPendingSpecChanges` comparing SSet template hashes. That misses two scenarios: (1) pause set mid-spec-roll — the SSet template already matches expected so it returns false while pods are still cycling, and (2) a pod deleted externally while paused — no spec change, but the replacement is mid-startup. The `IsPodReady` loop catches both. Worth a comment here capturing this rationale — a future reader sees a bare `IsPodReady` loop after `nodeShutdown.Clear` and is going to assume it's coupled to node shutdown.

**Concerns with the implementation:**

1. **Scope is wrong.** Both scenarios above apply to all ES versions, but this check sits inside `supportsNodeShutdown` (>= 7.15) because that's where `actualPods` happens to be in scope. Pre-7.15 clusters still hit the original problem. The pod fetch + check should be hoisted above this `if`.

2. **Silent during the wait.** Returns before `hasPendingSpecChanges` / condition update, so while pods are recovering the user sees the previous condition message — possibly stale. A condition update like `OrchestrationPaused: True, message: "waiting for pods to stabilize"` would give the operator something to look at. Distinct from the "pending changes" message — the user paused because they want stability, so the two states are worth distinguishing.

3. **Unbounded on stuck pods.** A CrashLoopBackOff pod (or any permanently unhealthy pod) pins this in a `DefaultRequeue` loop indefinitely with no event, no condition change. Pause + sick-pod becomes invisible.

A possible consolidation: rather than a bare `IsPodReady` loop, lift the steadiness check to the StatefulSet level — something like

```go
func ssetIsSteady(s appsv1.StatefulSet) bool {
    return s.Status.ObservedGeneration == s.Generation &&
        s.Status.CurrentRevision == s.Status.UpdateRevision &&
        s.Status.ReadyReplicas == ptr.Deref(s.Spec.Replicas, 0)
}
```

— and surface unsteady SSets via the condition (with a message naming the offending SSet) before requeueing. Covers both scenarios the current code does, doesn't need pod fetches, and is version-agnostic.

test/e2e/pause_orchestration_test.go
```diff
@@ -0,0 +1,657 @@
...
+	kbUpdated := kbEnabled.DeepCopy().WithNodeCount(2).WithMutatedFrom(&kbEnabled)
+	phase2Builder.AddBuilder(kbUpdated)
+	apmUpdated := apmEnabled.DeepCopy().WithNodeCount(2).WithMutatedFrom(&apmEnabled)
+	phase2Builder.MaybeAddBuilder(apmlabel.Type, apmUpdated)
+	emsUpdated := emsEnabled.DeepCopy().WithNodeCount(2).WithMutatedFrom(&emsEnabled)
+	phase2Builder.MaybeAddBuilder(emslabels.Type, emsUpdated)
+	// TODO how to perform spec change for beat and agent?
+	beatUpdated := beatEnabled.DeepCopy().WithMutatedFrom(&beatEnabled)
+	phase2Builder.MaybeAddBuilder(beatcommon.TypeLabelValue, beatUpdated)
+	agentUpdated := agentEnabled.DeepCopy().WithMutatedFrom(&agentEnabled)
```
pebrc: Acknowledged in the PR description, but worth flagging that without a real spec change in phase 2 for Beat/Agent, the assertion at line 119 (`CheckTestSteps(phase1[i], k)` — "topology unchanged while paused") is comparing unchanged-against-unchanged. The Beat/Agent paths through `pause_orchestration.go` and `pkg/controller/{beat,agent}/...` aren't really being exercised end-to-end.

Low-effort options:
- Bump a container resource limit on `Spec.DaemonSet.PodTemplate` / `Spec.Deployment.PodTemplate` — that's enough to change the template hash.
- Toggle a benign config key.

Either would let `hasPendingChanges` light up.

test/e2e/test/apmserver/checks_k8s.go
```diff
@@ -142,6 +142,7 @@ func CheckStatus(b Builder, k *test.K8sClient) test.Step {
 					AvailableNodes: b.ApmServer.Spec.Count,
 					Version:        b.ApmServer.Spec.Version,
 					Health:         "green",
+					Conditions:     as.Status.Conditions, // Ignore Conditions whose LastTransitionTime is unpredictable
```
pebrc, line 145: Minor: comment says "Ignore Conditions whose LastTransitionTime is unpredictable" but `Conditions: as.Status.Conditions` actually ignores the entire Conditions field — every condition delta becomes invisible.

`cmpopts.IgnoreFields(commonv1.Condition{}, "LastTransitionTime")` would still detect changes in `Type` / `Status` / `Message` and missing/extra conditions. Same shape in EnterpriseSearch, Kibana, EPR, Maps checks files.

pebrc (commented):
AI generated review. Most of the concerns below are about error paths and consistency between the per-controller implementations rather than the happy-path wiring, which I verified works correctly for each controller.

**Summary of concerns**

1. **Error paths drop the just-set condition.** Most visible: pause-orchestration set on a never-reconciled EnterpriseSearch or Maps resource loops with `NotFound` errors and never persists `OrchestrationPaused=True`. Similar shape in PackageRegistry's unpaused path if any later step errors after `MaybeResetPausedCondition`. Inline notes below.

2. **Hardcoded N+1 across seven controllers, five different shapes.** Same three-step pattern (check / set / reset) is open-coded everywhere with subtly different conventions (state-pointer vs explicit copy vs `params.Status` bridging in Agent/Beats). A single helper wrapping `deployment.Reconcile` would collapse ~21 call sites and remove the divergence.

3. **Spurious events.** A `Warning Paused` event is emitted on every reconcile while paused, not only on transitions or pending-changes detection. The ES implementation only emits when there are pending changes — worth aligning.

4. **Agent/Beats e2e doesn't actually test the pause guarantee.** TODO acknowledged — phase 2 builders for Beat/Agent don't mutate spec, so `CheckTestSteps(phase1[i], k)` is checking unchanged-against-unchanged. A simple resource-limit bump or config tweak would exercise the template-hash path.

**Smaller issues mentioned in the summary**

- `MergeConditions` nil-check duplicated across 7 types is redundant — `Conditions.MergeWith` already handles nil receivers (status.go:46).
- `CheckStatus` in test/e2e/test/*/checks_k8s.go sets `Conditions: x.Status.Conditions` and comments that this ignores `LastTransitionTime` — actually ignores all conditions. `cmpopts.IgnoreFields(commonv1.Condition{}, "LastTransitionTime")` would preserve more signal.
- `DeepCopy()` on the test/e2e builders silently drops `Validations`, `ValidationsOutputs`, `AdditionalObjects`, `PodTemplate`, `Suffix`. Fine for this test but the name is misleading — maybe `Clone()` with a doc comment, or copy the dropped fields.
- No resume event — pause emits a Warning, resume emits nothing. Asymmetric observability for operators tailing events.

**Verification**

I traced the condition flow empirically for EnterpriseSearch and PackageRegistry — the happy path persists conditions correctly via `status.DeploymentStatus = common.DeploymentStatus(...)` deep-copying the embedded `DeploymentStatus`. So the concern is specifically the *error paths* where this overwrite doesn't happen, plus consistency across the new controllers.

test/e2e/test/apmserver/checks_k8s.go
dtuck9: The `Conditions` field is checked for expectations in `pause_orchestration_test.go`, so checking them here would:
1. be redundant
2. require refactoring every builder a little bit to decouple the `Conditions` from the object the builder is building, because we don't want the object to be created with the expected Conditions the same way we would want the `Spec.Count` or `Spec.Version` to be created.

Because the `Conditions` field is currently limited to the `OrchestrationPaused` condition, I opted for the path of least resistance by containing the `Conditions` checks in `pause_orchestration_test.go`.

If you think it's worth refactoring for future `Conditions` extensibility, I can do that too, though.

test/e2e/pause_orchestration_test.go
dtuck9: These are now implemented.

pkg/controller/elasticsearch/driver/stateful/driver.go
dtuck9: Another edge case I had realized is that `if !k8s.IsPodReady(pod)` would end up re-queuing on terminating pods, which we also wouldn't want to do.

If I understand the `ssetIsSteady` suggestion correctly, I think that edge case would also be covered by the `ReadyReplicas` check, correct?

pkg/controller/common/pause_orchestration.go
dtuck9: That makes sense, and the Elasticsearch implementation will also need to be updated.

pkg/controller/common/pause_orchestration.go
dtuck9: I will evaluate this and agree it would be more ideal, but this is highlighting a pre-existing drift in the implementations of each of these, which makes consolidation more difficult without additional refactoring of the existing behavior.

dtuck9: I think I would rather decouple that effort from the initial implementation since it has a wider spread impact.

dtuck9:
buildkite test this -f p=gke,E2E_TAGS=mixed -m s=8.19.14,s=9.3.4,s=7.17.29

[[build](https://buildkite.com/elastic/cloud-on-k8s-operator/builds/14821)]

dtuck9:
buildkite test this -f p=gke,TESTS_MATCH="^TestPauseOrchestration*" -m s=8.19.14,s=9.3.4,s=7.17.29

[[build](https://buildkite.com/elastic/cloud-on-k8s-operator/builds/14822)]

dtuck9:
buildkite test this -f p=gke,TESTS_MATCH=^TestPauseOrchestration* -m s=8.19.14,s=9.3.4,s=7.17.29

[[build](https://buildkite.com/elastic/cloud-on-k8s-operator/builds/14823)]

dtuck9:
buildkite test this -f p=gke,TESTS_MATCH=^TestPauseOrchestration* -m s=8.19.14,s=9.3.4,s=7.17.29

[[build](https://buildkite.com/elastic/cloud-on-k8s-operator/builds/14825)]

dtuck9:
buildkite test this -f p=gke,TESTS_MATCH=^TestPauseOrchestration* -m s=8.19.14,s=9.3.4,s=7.17.29

dtuck9:
buildkite test this -f p=gke,TESTS_MATCH=^TestPauseOrchestration* -m s=8.19.14,s=9.3.4,s=7.17.29

[[build](https://buildkite.com/elastic/cloud-on-k8s-operator/builds/14832)]

dtuck9:
The Agent test is still flaky -- yellow health instead of the expected green -- even with the requeue in place. I suspect the Beat is as well, though it's harder to reproduce now.  I think I'll need to replicate the Elasticsearch path's

```
func ssetIsSteady(s appsv1.StatefulSet) bool {
	return s.Status.ObservedGeneration == s.Generation &&
		s.Status.CurrentRevision == s.Status.UpdateRevision &&
		s.Status.ReadyReplicas == ptr.Deref(s.Spec.Replicas, 0)
}
```
which looks at the state of the actual statefulset.

Instead of having a redundant lookup on the existing resource in addition to the `hasPendingChanges` function, I will look into refactoring and possibly include the suggestion from https://github.com/elastic/cloud-on-k8s/pull/9398#discussion_r3268519068.

pkg/controller/common/pause_orchestration.go
dtuck9: I've done this to hopefully resolve the flakiness and consolidate the functionality. 

One disclaimer, I did put it in the pause_orchestration file, regrettably, because trying to put it alongside the respective object's `Reconcile` function, e.g., `deployment.Reconcile`, caused an import cycle.

dtuck9:
buildkite test this -f p=gke,TESTS_MATCH=^TestPauseOrchestration* -m s=8.19.14,s=9.3.4,s=7.17.29

[[build](https://buildkite.com/elastic/cloud-on-k8s-operator/builds/14852)]

dtuck9:
buildkite test this -f p=gke,TESTS_MATCH=^TestPauseOrchestration* -m s=8.19.14,s=9.3.4,s=7.17.29

[[build](https://buildkite.com/elastic/cloud-on-k8s-operator/builds/14854)]

pkg/controller/beat/common/reconcile.go
```diff
@@ -104,18 +122,24 @@ func reconcileDeployment(rp ReconciliationParams) (int32, int32, error) {
...
+		return v1.Deployment{}, err
 	}
+	return deployment.WithTemplateHash(d), nil
+}
 
-	reconciled, err := deployment.Reconcile(rp.ctx, rp.client, d, &rp.beat)
+func reconcileDeployment(rp ReconciliationParams, obj client.Object) (beatv1beta1.BeatStatus, int32, int32, error) {
+	expected, ok := obj.(*v1.Deployment)
+	if !ok {
+		return rp.beat.Status, 0, 0, fmt.Errorf("%T is not a DaemonSet", obj)
```
barkbay:
```suggestion
		return rp.beat.Status, 0, 0, fmt.Errorf("%T is not a Deployment", obj)
```

pkg/controller/elasticsearch/driver/stateful/driver_test.go
```diff
@@ -375,6 +375,23 @@ func TestDriver_reconcileCriticalStepsWhilePaused(t *testing.T) {
 				},
 			},
 		},
+		{
+			name:                 "restarted pods that have not yet joined trigger a requeue",
```
barkbay, line 379: I think the test name doesn't match what it tests: k8sObjects is empty, so actualSets is also empty?

pkg/apis/apm/v1/apmserver_types.go
```diff
@@ -207,6 +207,16 @@ func (as *ApmServer) GetObservedGeneration() int64 {
 	return as.Status.ObservedGeneration
 }
 
+// MergeConditions provides a nil-safe way to merge the ApmServerStatus's Conditions with the new Condition(s).
```
barkbay, line 210: This comment (on all the types) feels a bit misleading as only the `conditions` can be `nil`, not the receiver.
```suggestion
// MergeConditions merges the given Conditions into the ApmServerStatus's Conditions, handling a nil Conditions slice.
```

Or we can also handle the case where `as` is `nil`.

pkg/controller/common/pause_orchestration.go
```diff
@@ -5,18 +5,222 @@
...
 
 // IsOrchestrationPaused returns true if the PauseOrchestrationAnnotation exists and is set to true on the given resource
 // when non-critical orchestration steps should be skipped.
 func IsOrchestrationPaused(object metav1.Object) bool {
 	return object.GetAnnotations()[PauseOrchestrationAnnotation] == "true"
 }
+
+// setPausedConditionAndEmitEvent adds the OrchestrationPaused condition with a value of True and emits an event. The parent
+// is the ECK CR, such as v1.Kibana, to be updated with the appropriate v1.Conditions, while the expected client.Object
+// is the underlying kubernetes resource that is created as a result of the ECK object's Spec.
```
barkbay: nit


```suggestion
// setPausedConditionAndEmitEvent adds the OrchestrationPaused condition with a value of True on the parent, and emits a
// Warning event when spec changes are pending (no event is emitted on the no-changes path). The parent is the ECK CR,
// such as v1.Kibana, to be updated with the appropriate v1.Conditions, while the expected client.Object is the
// underlying kubernetes resource that is created as a result of the ECK object's Spec.
```

barkbay (commented):
Looks good overall. I left a few nits. I'll run some tests today and report back here.

pkg/controller/common/pause_orchestration.go
```diff
@@ -5,18 +5,222 @@
...
+			if apierrors.IsNotFound(err) {
+				return appsv1.Deployment{}, nil
+			}
+			return appsv1.Deployment{}, err
+		}
+
+		setPausedConditionAndEmitEvent(recorder, owner, &expected, &actual)
+
+		return actual, nil
+	}
```
barkbay: There is a bit of duplicated code, I was wondering if we could introduce a new function that could be shared, something along the lines of:

```go
// reconcilePauseAware wraps the given reconcileFn with pause-orchestration handling: when paused, it fetches the
// existing resource, sets the OrchestrationPaused condition (with a Warning event if spec changes are pending), and
// returns the live object without applying any spec change. When not paused, it clears any stale OrchestrationPaused
// condition and delegates to reconcileFn.
//
// PT exists so the function can take the address of T values and pass them to client.Object-typed APIs (Get,
// setPausedConditionAndEmitEvent). It is constrained to be a pointer to T that also implements client.Object.
func reconcilePauseAware[T any, PT interface {
	*T
	client.Object
}](
	ctx context.Context,
	c k8s.Client,
	recorder toolsevents.EventRecorder,
	expected T,
	owner ObjectWithConditions,
	reconcileFn func(context.Context, k8s.Client, T, client.Object) (T, error),
) (T, error) {
	if IsOrchestrationPaused(owner) {
		var actual T
		if err := c.Get(ctx, k8s.ExtractNamespacedName(PT(&expected)), PT(&actual)); err != nil {
				return *new(T), nil
			}
			return *new(T), err
		}
		setPausedConditionAndEmitEvent(recorder, owner, PT(&expected), PT(&actual))
		return actual, nil
	}

	maybeResetPausedCondition(recorder, owner)

	return reconcileFn(ctx, c, expected, owner)
}
```

And then:

```go
// ReconcileDeployment reconciles the given Deployment for the specified owner, taking into account
// whether orchestration has been paused by the eck.k8s.elastic.co/pause-orchestration annotation.
func ReconcileDeployment(ctx context.Context, c k8s.Client, recorder toolsevents.EventRecorder, expected appsv1.Deployment, owner ObjectWithConditions) (appsv1.Deployment, error) {
	return reconcilePauseAware[appsv1.Deployment](ctx, c, recorder, expected, owner, deployment.Reconcile)
}

// ReconcileDaemonSet reconciles the given DaemonSet for the specified owner, taking into account
// whether orchestration has been paused by the eck.k8s.elastic.co/pause-orchestration annotation.
func ReconcileDaemonSet(ctx context.Context, c k8s.Client, recorder toolsevents.EventRecorder, expected appsv1.DaemonSet, owner ObjectWithConditions) (appsv1.DaemonSet, error) {
	return reconcilePauseAware[appsv1.DaemonSet](ctx, c, recorder, expected, owner, daemonset.Reconcile)
}

// ReconcileStatefulSet reconciles the given StatefulSet for the specified owner, taking into account
// whether orchestration has been paused by the eck.k8s.elastic.co/pause-orchestration annotation.
func ReconcileStatefulSet(ctx context.Context, c k8s.Client, recorder toolsevents.EventRecorder, expected appsv1.StatefulSet, owner ObjectWithConditions) (appsv1.StatefulSet, error) {
	return reconcilePauseAware[appsv1.StatefulSet](ctx, c, recorder, expected, owner, statefulset.Reconcile)
}
```

barkbay: Following up my previous comment I realized that if the resource has never been reconciled the status is `red` (maybe expected?) and status is not updated, so a fix could be:

```go
// reconcilePauseAware wraps the given reconcileFn with pause-orchestration handling: when paused, it fetches the
// existing resource, sets the OrchestrationPaused condition (with a Warning event if spec changes are pending), and
// returns the live object without applying any spec change. When not paused, it clears any stale OrchestrationPaused
// condition and delegates to reconcileFn.
//
// PT exists so the function can take the address of T values and pass them to client.Object-typed APIs (Get,
// setPausedConditionAndEmitEvent). It is constrained to be a pointer to T that also implements client.Object.
func reconcilePauseAware[T any, PT interface {
	*T
	client.Object
}](
	ctx context.Context,
	c k8s.Client,
	recorder toolsevents.EventRecorder,
	expected T,
	owner ObjectWithConditions,
	reconcileFn func(context.Context, k8s.Client, T, client.Object) (T, error),
) (T, error) {
	if IsOrchestrationPaused(owner) {
		var actual T
		var actualForDiff client.Object // nil when the resource doesn't exist yet — treated as "pending changes" by hasPendingChanges
		err := c.Get(ctx, k8s.ExtractNamespacedName(PT(&expected)), PT(&actual))
		switch {
		case err == nil:
			actualForDiff = PT(&actual)
		case apierrors.IsNotFound(err):
			// first-reconcile-while-paused: leave actualForDiff nil so the condition is still reported with PendingChanges
		default:
			return *new(T), err
		}
		setPausedConditionAndEmitEvent(recorder, owner, PT(&expected), actualForDiff)
		return actual, nil
	}

	maybeResetPausedCondition(recorder, owner)

	return reconcileFn(ctx, c, expected, owner)
}
```

barkbay: Confirmed with a quick test:

## Without that change

```yaml
status:
  count: 0
  health: red
  observedGeneration: 1
```

## With the change just above

```yaml
status:
  conditions:
  - lastTransitionTime: "2026-05-22T10:00:08Z"
    message: Orchestration paused via annotation; spec changes are pending and will
      be applied on resume
    status: "True"
    type: OrchestrationPaused
  count: 0
  health: red
  observedGeneration: 1
```

pkg/controller/agent/reconcile.go
```diff
@@ -114,15 +135,14 @@ func reconcilePodVehicle(params Params, podTemplate corev1.PodTemplateSpec) (*re
 		}
 	}
```
barkbay: This is not gated by `IsOrchestrationPaused`, so a paused vehicle swap deletes the existing Deployment/DaemonSet/StatefulSet while the paused reconcile path correctly skips creating the new one, result is zero pods while paused.

## Repro
1. Apply an `Agent`, wait for the pod.
2. Set eck.k8s.elastic.co/pause-orchestration: "true".
3. Patch the spec to remove deployment and add statefulSet.

Within ~2s: Deployment gone, pod terminated, no StatefulSet created.

## Fix

Wrap the `toDelete` loop with `if !common.IsOrchestrationPaused(&params.Agent)` (and `&params.Beat` in the beat file). Cleanup then runs naturally on the next reconcile after the annotation is removed.

Worth a test in controller_test.go: existing Deployment + Agent spec'd as StatefulSet with the annotation → Deployment preserved, no StatefulSet, condition reported.

barkbay (changes requested):

barkbay (commented):
Small nit from my testing: when a user sets a typo'd value like `True` / `yes` / `1`, ECK silently does nothing. Someone who fat-fingers the annotation may think the feature is broken. In `IsOrchestrationPaused`, if the annotation key is present but the value isn't "true", maybe we could emit a `Warning` event with text like `pause-orchestration annotation value '%s' ignored — expected literal 'true'`? Or do it in the validating webhook?

pkg/controller/agent/reconcile.go
dtuck9: Thank you. I seem to have a smaller bug in addition to this as that does not render the expected `Orchestration paused via annotation; spec changes are pending and will be applied on resume` message and `Warning` event in this edge case either.

Working on both fixes.

pkg/apis/apm/v1/apmserver_types.go
dtuck9: Ah, this is a product of removing the following check that was marked as redundant:

```
	if as.Status.Conditions == nil {
		as.Status.Conditions = make([]commonv1.Condition, 0)
	}
```

Thinking defensively, I'm inclined to re-add that check along with a `nil` check on `as`. Thoughts?

dtuck9:
@barkbay re: 

> Small nit from my testing: when a user sets a typo'd value like `True` / `yes` / `1`, ECK silently does nothing. Someone who fat-fingers the annotation may think the feature is broken. In `IsOrchestrationPaused`, if the annotation key is present but the value isn't "true", maybe we could emit a `Warning` event with text like `pause-orchestration annotation value '%s' ignored — expected literal 'true'`? Or do it in the validating webhook?

I had the same thought when I initially implemented the related Elasticsearch changes, but talked myself out of it because the existing `unmanaged` annotation worked the same way. Happy to add this -- from the looks of it I'd need to add a validation step for each of the respective CRs' webhook.go?

And separately, I think I'd rather do it as a separate PR at this point because this one is already a bit bloated.

pkg/controller/common/pause_orchestration.go
dtuck9: Great suggestion, thanks

pkg/controller/agent/reconcile.go
dtuck9: Wait... the secondary bug was actually resolved with the other code change you recommended in https://github.com/elastic/cloud-on-k8s/pull/9398#discussion_r3287500772 because of the `nil` check in `hasPendingChanges`. I wish there were an elegant way to check the zero-value of the 3 different potential types, but I think your suggestion is probably the most elegant way.

Regarding, `red`, though -- Ideally that wouldn't be the case and instead should reflect the state of the pre-paused spec. What's happening though is the "actual" (i.e., the pending resource that hasn't been created yet) returns 0 for `reconciled.Status.ReadyReplicas` because the "actual" doesn't even exist yet, and it will automatically get set to `red` as a result:

https://github.com/elastic/cloud-on-k8s/blob/3b6812b7580439d3c03ecf77647eeebebd3b5e03/pkg/controller/agent/health.go#L28-L29

Trying to think through ways to resolve that without going too far into the weeds of diminishing returns.

dtuck9:
buildkite test this -f p=gke,TESTS_MATCH=^TestPauseOrchestration* -m s=8.19.14,s=9.3.4,s=7.17.29

[[build](https://buildkite.com/elastic/cloud-on-k8s-operator/builds/14874)]

pkg/controller/elasticsearch/driver/stateful/driver_test.go
dtuck9: I think I need to reevaluate all of these test cases tbh. The intent here was that the ES client doesn't return the particular node that hasn't rejoined in the `GetNodes` call and to requeue in that event, but you're right that `actualSets` shouldn't be empty here.

barkbay (approved):
Both bug fixes look good, verified them live against this branch 👍 

LGTM

pkg/controller/agent/reconcile.go
barkbay: Yes, I think a red status is reasonable here, the `OrchestrationPaused=True` / `spec changes are pending and will be applied on resume` condition is the canonical signal, health is downstream of that.

dtuck9:
I am going to address the remaining smaller items in a follow up PR.

Issue #9082, opened 2026-01-28
Full PKI Realm Integration


pebrc:
* ECK configures Elasticsearch's PKI realm
* Map certificate Subject DNs to roles
* Full certificate lifecycle management
* Not supported by all stack apps (Fleet server, Fleet-managed agents, Beats)

pkoutsovasilis:
Here is a proposal for how we approach the full PKI realm integration tracked in this issue. The core idea is to allow stack components to authenticate to Elasticsearch using their ECK-managed TLS client certificate instead of a username/password or API key. The design centres on a single new field on `spec.elasticsearchRef`, shared across all compatible stack component CRDs, with the Elasticsearch controller and association controller each taking on a well-scoped piece of the work. Details below.

## 1. CRD Changes - `authenticationMode` on `spec.elasticsearchRef`

The `spec.elasticsearchRef` field present on Kibana, APM Server, Beats, Logstash, Elastic Agent, Enterprise Search, and Elastic Maps Server gains a new optional sub-field:

```yaml
elasticsearchRef:
  name: my-es
  authenticationMode: PKI   # new; omitting it is equivalent to Default
```

- **`Default`** (the zero value, backward-compatible) - standard ECK credential flow: ECK creates an Elasticsearch user or service-account token, stores it in a Secret, and injects it into the component pod. Exactly the behavior that exists today.

- **`PKI`** - ECK skips credential creation entirely. The component's ECK-managed TLS client certificate is presented at the TLS layer, and Elasticsearch's PKI realm maps the certificate's Subject DN to an Elasticsearch role. No username, password, or API key is distributed to the component. This requires Elasticsearch to have mTLS enabled (`spec.http.tls.client.authentication: true`).

> [!NOTE]
> This field is also a valid baseline for exposing finer-grained credential modes in the future - for example `BasicAuthUsernamePassword`, `BasicAuthAPIKey`, or `ServiceAccountToken` - for components that support them. Today `Default` encapsulates all of that under a single value; the enum can be extended without breaking the existing API.

### Behavior when mTLS is not active

[...]

PR #9399, opened 2026-04-24
Implement mTLS support for fleet-server and elastic-agent connections

pkoutsovasilis:
## Summary

This PR implements the Fleet Server side of the mTLS client certificate authentication feature ([#9081](https://github.com/elastic/cloud-on-k8s/issues/9081)). It adds a new `spec.http.tls.client.authentication` field to the Agent CRD (when Fleet Server is enabled) that configures Fleet Server to require client certificates from connecting Elastic Agents.

### Changes

- **New API types**: Extended `HTTPConfig` with `HTTPConfigWithClientOptions` and `TLSWithClientOptions` to support `spec.http.tls.client.authentication` on Agent resources with Fleet Server enabled
- **FleetServerSelector**: New `FleetServerSelector` type with `ClientCertificateSecretName` field allowing users to specify custom client certificates for Agent-to-Fleet Server connections
- **Client certificate trust bundle**: Aggregates client certificates from Elastic Agents connecting to a Fleet Server into a trust bundle mounted into Fleet Server pods
- **Fleet Server configuration**: Automatically sets `FLEET_SERVER_CLIENT_AUTH=required` environment variable when client authentication is enabled
- **Internal client certificate**: ECK generates and manages an internal client certificate that is mounted in the Fleet Server pod, allowing the internal Elastic Agent to authenticate to the bootstrapped Fleet Server when client authentication is enabled
- **Agent client certificates**: When Elastic Agents connect to a Fleet Server with client authentication enabled, ECK auto-generates client certificates for them (or users can specify their own via `fleetServerRef.clientCertificateSecretName`)
- **Annotation-based signaling**: Sets a `client-authentication-required` annotation on the Fleet Server Agent resource to signal to connecting Agents that client certificates are required
- **License validation**: Client authentication requires an Enterprise license (validated via webhook)
- **ValueFrom support detection**: When `FLEET_SERVER_CLIENT_AUTH` is set via `valueFrom` in pod templates, ECK emits a warning since it cannot determine the actual value

### API

Enable client authentication on Fleet Server:

```yaml
kind: Agent
metadata:
  name: fleet-server
spec:
  mode: fleet
  fleetServerEnabled: true
  http:
    tls:
      client:
        authentication: true  # Enable client certificate authentication (enterprise feature)
  # ...
```

Connect an Elastic Agent to a Fleet Server with client auth enabled:

```yaml
kind: Agent
metadata:
  name: elastic-agent
spec:
  mode: fleet
  fleetServerRef:
    name: fleet-server
    # Optional: specify a custom client certificate secret
    # clientCertificateSecretName: my-custom-client-cert
  # ...
```

## Known Limitation: Elastic Agent mTLS Transition Bug

> **Important:** There is a known bug in Elastic Agent ([elastic/elastic-agent#13810](https://github.com/elastic/elastic-agent/issues/13810)) that affects mTLS transitions.

When transitioning a Fleet Server from mTLS-enabled (`FLEET_SERVER_CLIENT_AUTH=required`) to mTLS-disabled (or vice versa), existing Elastic Agent pods will crashloop with:

```
Error: failed to read from disk store: open /mnt/elastic-internal/fleetserver-association/default/fleet-server/client-certs/tls.crt: no such file or directory
```

**Root cause:** The Elastic Agent stores its TLS configuration in an encrypted state file (`fleet.enc`). When mTLS is disabled, the certificate files are removed, but the agent still tries to use the stored paths.

**Workaround:** When transitioning mTLS on or off, users must clear the Elastic Agent state directory (`/usr/share/elastic-agent/state/`).


prodsecmachine:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/aafa8b07-ad3d-4b77-87af-800b210667cc?refs=ee9124593164463e24fb2c3f97113d4c666a3ee7&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/aafa8b07-ad3d-4b77-87af-800b210667cc/license?refs=ee9124593164463e24fb2c3f97113d4c666a3ee7&source=prChecksComment) |

pkoutsovasilis:
buildkite test this -f p=gke,t=TestClientAuthRequiredCustomCertificate_FleetServerToAgent,E2E_TAGS=agent -m s=9.3.3,s=8.19.14,s=9.4.0-SNAPSHOT

pkoutsovasilis:
buildkite test this -f p=gke,t=TestClientAuthRequiredCustomCertificate_FleetServerToAgent,E2E_TAGS=agent -m s=9.3.3,s=8.19.14,s=9.4.0,s=9.5.0-SNAPSHOT

pkoutsovasilis:
buildkite test this -f p=gke,t=TestClientAuthRequiredCustomCertificate_FleetServerToAgent,E2E_TAGS=agent -m s=9.3.4,s=8.19.15,s=9.2.5,s=9.4.0,s=9.5.0-SNAPSHOT

pkg/controller/agent/driver.go
```diff
@@ -215,6 +240,123 @@ func newService(agent agentv1alpha1.Agent, meta metadata.Metadata) *corev1.Servi
...
+	}
+	enabled, err := params.LicenseChecker.EnterpriseFeaturesEnabled(params.Context)
+	if err != nil {
+		return false, "", err
+	}
+	if !enabled {
+		return false, "", nil
+	}
+
+	specEnabled := params.Agent.Spec.HTTP.TLS.Client.Authentication
```
pebrc: **Silent feature downgrade when an Enterprise license is not present.**

When the user sets `spec.http.tls.client.authentication: true` but the cluster lacks an Enterprise license, `EnterpriseFeaturesEnabled` returns `false` and the function silently returns `false, ""` — Fleet Server is then provisioned **without** client authentication and no event is emitted.

This becomes a real risk because the webhook uses `failurePolicy: ignore` (`pkg/apis/agent/v1alpha1/webhook.go:28`): if the webhook is unavailable when the manifest is applied, validation is skipped and the resource lands in the API. The user reasonably believes they have mTLS, but they silently don't.

Consider emitting a warning event here (similar to the TLS-disabled / `valueFrom` branches below) when `specEnabled && !enabled`, e.g. `"spec.http.tls.client.authentication is ineffective because an Enterprise license is not enabled"`.

pkg/controller/agent/driver.go
```diff
@@ -215,6 +240,123 @@ func newService(agent agentv1alpha1.Agent, meta metadata.Metadata) *corev1.Servi
...
+				continue
+			}
+			// The env var is set via valueFrom (ConfigMap/Secret): we cannot determine the actual value
+			// at reconcile time, so we treat it as not required and emit a warning.
+			if env.ValueFrom != nil {
+				return false, "spec.http.tls.client.authentication is ineffective because FLEET_SERVER_CLIENT_AUTH is " +
+					"set via valueFrom and the operator cannot determine the actual value", nil
+			}
+			// The env var is set to required: client auth is required regardless of the spec value.
+			if env.Value == "required" {
```
pebrc: **Magic string `"required"`.**

This value also appears in `pkg/controller/agent/pod.go:696` (`fleetServerCfg[FleetServerClientAuth] = "required"`) and in tests. Consider a single constant (e.g. `FleetServerClientAuthRequired = "required"` next to the `FleetServerClientAuth` env-var name in `pod.go`) so the env-var name and its sentinel value live together and there's no risk of a typo drift between writer and reader.

pkg/controller/agent/driver.go
```diff
@@ -136,6 +140,27 @@ func internalReconcile(params Params) (*reconciler.Results, agentv1alpha1.AgentS
...
+			return results.WithError(err), params.Status
+		}
+		if clientAuthWarning != "" {
+			k8s.EmitEvent(params.EventRecorder, &params.Agent, corev1.EventTypeWarning, events.EventReasonValidation, events.EventActionValidation, clientAuthWarning)
+		}
+		clientCertResults := reconcileFleetServerClientAuth(params, clientAuthRequired, fleetCerts, configHash)
+		if clientCertResults.HasError() {
+			return results.WithResults(clientCertResults), params.Status
+		}
+		results.WithResults(clientCertResults)
```
pebrc: Style nit: this `results.WithResults(clientCertResults)` discards the return value. `WithResults` does mutate the receiver so the behavior is correct, but two lines above we use `return results.WithResults(...)` and elsewhere in this function (`line 138`, `170`) the same pattern is used as a return value. Either drop this call (the error path at line 159 already merged the only errors that can flow in) or assign it explicitly for symmetry.

pkg/controller/agent/pod.go
```diff
@@ -619,13 +657,19 @@ func getFleetSetupFleetEnvVars(client k8s.Client, fleetToken EnrollmentAPIKey, f
 			if assocConf.GetCACertProvided() {
 				fleetCfg[FleetCA] = path.Join(CertificatesDir(assoc), CAFileName)
 			}
+
+			if assocConf.ClientCertIsConfigured() {
+				clientCertDir := standaloneAgentClientCertificatesDir(assoc)
```
pebrc: `standaloneAgentClientCertificatesDir` is now used for the **fleet-managed** agent case too (this new call site for the FleetServer association's client cert). The name implied a standalone-only path, which is no longer accurate. Suggest renaming to `associationClientCertificatesDir` (or similar) so future readers don't assume scope it doesn't have.

pkg/controller/agent/webhook.go
```diff
@@ -0,0 +1,81 @@
...
+// validClientAuthentication checks that client certificate authentication requires an enterprise license.
+func validClientAuthentication(ctx context.Context, a *agentv1alpha1.Agent, checker commonlicense.Checker) field.ErrorList {
+	if !a.Spec.HTTP.TLS.Client.Authentication {
+		return nil
+	}
+	enabled, err := checker.EnterpriseFeaturesEnabled(ctx)
+	if err != nil {
+		ulog.FromContext(ctx).Error(err, "while checking enterprise features during client authentication validation")
+		return nil
+	}
```
pebrc, line 73: The webhook **fails-open** on `EnterpriseFeaturesEnabled` errors (returns `nil`, admits the resource), while the controller in `driver.go:isFleetServerClientAuthRequired` **fails-closed** on the same error (returns it, reconcile errors out).

Both are defensible — webhook avoids blocking writes during a transient license-checker outage; controller avoids silently disabling security — but the asymmetry is non-obvious. Worth a one-line comment here noting the deliberate divergence, so a future reader doesn't "fix" the inconsistency in the wrong direction.

pkg/apis/agent/v1alpha1/validations.go
```diff
@@ -302,6 +303,34 @@ func checkSingleESRefInFleetMode(a *Agent) field.ErrorList {
 func checkAssociations(a *Agent) field.ErrorList {
 	err1 := commonv1.CheckAssociationRefs(field.NewPath("spec").Child("elasticsearchRefs"), a.ElasticsearchRefs()...)
 	err2 := commonv1.CheckAssociationRefs(field.NewPath("spec").Child("kibanaRef"), a.Spec.KibanaRef)
-	err3 := commonv1.CheckAssociationRefs(field.NewPath("spec").Child("fleetServerRef"), a.Spec.FleetServerRef)
+	var err3 field.ErrorList
+	if err := a.Spec.FleetServerRef.IsValid(); err != nil {
+		err3 = field.ErrorList{field.Forbidden(field.NewPath("spec").Child("fleetServerRef"), fmt.Sprintf("Invalid association reference: %s", err))}
```
pebrc: Minor consistency nit: `CheckElasticsearchSelectorRefs` and `CheckLocalAssociationRefs` in `pkg/apis/common/v1/validations.go` already wrap this exact "iterate refs, run `IsValid`, return `field.Forbidden`" boilerplate. Consider adding `CheckFleetServerSelectorRefs(path, refs...) field.ErrorList` next to them and calling it here, so all three selector types validate through the same helper.

pkg/apis/common/v1/common.go
```diff
@@ -333,6 +333,48 @@ func (e ElasticsearchSelector) WithDefaultNamespace(defaultNamespace string) Ela
...
+// FleetServerSelector defines a reference to a Fleet Server managed by the operator
+// or a Secret describing an external Fleet Server not managed by the operator.
+type FleetServerSelector struct {
+	ObjectSelector `json:",inline"`
+
+	// ClientCertificateSecretName is the name of an existing Kubernetes secret containing a client certificate
+	// (tls.crt) and private key (tls.key) for client authentication to the referenced Fleet Server.
+	// This field is only relevant when the referenced Fleet Server has client authentication enabled.
+	// If not specified and the referenced Fleet Server requires client authentication, ECK will auto-generate a
+	// client certificate.
```
pebrc, line 347: `FleetServerSelector.IsValid()` requires `Name` to be set when `ClientCertificateSecretName` is non-empty, which means an externally-referenced Fleet Server (`secretName` form, supported by the embedded `ObjectSelector`) can never have a custom client cert.

That seems like the intended behavior (the certs are tied to ECK-managed Fleet Servers), but it's not currently spelled out for the user. Suggest one more sentence in the field doc: `"This field is only meaningful when referencing a Fleet Server managed by ECK (via name); it cannot be combined with secretName."`

pkg/apis/agent/v1alpha1/agent_types.go
```diff
@@ -88,7 +88,7 @@ type AgentSpec struct {
 
 	// HTTP holds the HTTP layer configuration for the Agent in Fleet mode with Fleet Server enabled.
 	// +kubebuilder:validation:Optional
-	HTTP commonv1.HTTPConfig `json:"http,omitempty"`
+	HTTP commonv1.HTTPConfigWithClientOptions `json:"http,omitempty"`
```
pebrc, line 93: The user-facing comment on `HTTP` doesn't mention the new `tls.client.authentication` subfield. Worth one extra sentence pointing users at the enterprise-only feature, e.g.:

`"Set tls.client.authentication: true to require connecting Elastic Agents to present a client certificate (requires an Enterprise license)."`

Otherwise the only place this feature surfaces in the API docs is the `Authentication` field comment buried in `commonv1.ClientOptions`.

test/e2e/agent/client_auth_test.go
```diff
@@ -273,6 +273,91 @@ func TestClientAuthRequiredCustomCertificate_FleetAgent(t *testing.T) {
 	test.BeforeAfterSequence(before, after, test.LicenseTestBuilder(esBuilder), kbBuilder, fleetServerBuilder, agentWrapped).RunSequential(t)
 }
 
+// TestClientAuthRequiredCustomCertificate_FleetServerToAgent tests that a fleet-managed Agent uses a user-provided
+// client certificate when connecting to a Fleet Server that has client authentication enabled.
+func TestClientAuthRequiredCustomCertificate_FleetServerToAgent(t *testing.T) {
```
pebrc, line 278: Only the **custom-cert** path is covered in e2e (`TestClientAuthRequiredCustomCertificate_FleetServerToAgent`). The **auto-generated** cert path — when the user does not set `fleetServerRef.clientCertificateSecretName` — is the recommended default and the more commonly exercised path in practice, but is currently only verified via unit tests + the managed-mTLS recipe.

A `TestClientAuthRequired_FleetServerToAgent_AutoGenerated` analog (no `WithFleetServerRefWithClientCert`, just `WithFleetServerRef`) would close the gap. Skipping the transition variant given the upstream elastic-agent bug noted in the PR description is reasonable.

pkg/controller/agent/driver.go
```diff
@@ -215,6 +240,123 @@ func newService(agent agentv1alpha1.Agent, meta metadata.Metadata) *corev1.Servi
...
+	}
+
+	if err := annotation.SetClientAuthenticationRequiredAnnotation(params.Context, params.Client, &params.Agent); err != nil {
+		return results.WithError(err)
+	}
+
+	certReconciler := certificates.Reconciler{
+		K8sClient:    params.Client,
+		Owner:        &params.Agent,
+		Namer:        Namer,
```
pebrc, line 335: **Architecture / N+1 follow-up (not a blocker for this PR).**

`reconcileFleetServerClientAuth` (~30 lines) and `isFleetServerClientAuthRequired` (~60 lines) are written specifically for this CRD. The API-level type `HTTPConfigWithClientOptions` is already shared (used by ES and now Agent), but extending this feature to a third resource (Kibana, APM…) would mean duplicating both helpers in another controller.

Worth considering a follow-up to lift these into a shared helper colocated with `client_cert_reconcile.go`, taking the owner, namer, license checker, and a callback that locates the relevant container's env vars. Flagging now since two instances is the natural moment to generalize — easier than refactoring after the third.

pkg/apis/common/v1/common.go
```diff
@@ -333,6 +333,48 @@ func (e ElasticsearchSelector) WithDefaultNamespace(defaultNamespace string) Ela
...
+// WithDefaultNamespace adds a default namespace to a given FleetServerSelector if none is set.
+func (f FleetServerSelector) WithDefaultNamespace(defaultNamespace string) FleetServerSelector {
+	if len(f.Namespace) > 0 {
+		return f
+	}
+	return FleetServerSelector{
+		ObjectSelector:              f.ObjectSelector.WithDefaultNamespace(defaultNamespace),
+		ClientCertificateSecretName: f.ClientCertificateSecretName,
+	}
+}
```
pebrc, line 378: **Consider collapsing `FleetServerSelector` and `ElasticsearchSelector` into a single shared type.**

The two are structurally identical: both embed `ObjectSelector`, both add a `ClientCertificateSecretName string` field, and `GetClientCertificateSecretName` / `IsValid` / `WithDefaultNamespace` are character-for-character the same modulo the receiver name. The error string in `IsValid` is literally identical. The only meaningful difference is one word in each doc comment.

A shared `AssociationRefWithClientCert` (or similar) type would remove ~40 lines plus the matching deepcopy, and would give the next association that needs client certs (Kibana? APM?) a drop-in option rather than another sibling type.

Two copies is the cheapest moment to deduplicate — once a third resource needs the same thing, the refactor touches more CRDs and more controllers. I'd much rather see this generalized now than carry the duplication forward.

Not a blocker for this PR, but worth deciding deliberately rather than by inertia from how `ElasticsearchSelector` happened to be introduced in #9234.

pkg/controller/agent/webhook.go
```diff
@@ -0,0 +1,81 @@
...
+	ulog "github.com/elastic/cloud-on-k8s/v3/pkg/utils/log"
+)
+
+// RegisterWebhook registers the Agent validating webhook with license-aware validation.
+func RegisterWebhook(mgr ctrl.Manager, checker commonlicense.Checker, managedNamespaces []string) {
+	inner := &webhookValidator{licenseChecker: checker}
+	// Pass nil for licenseChecker: the inner validator performs license enforcement itself.
+	v := commonwebhook.NewResourceValidator[*agentv1alpha1.Agent](nil, managedNamespaces, inner)
+	wh := admission.WithValidator[*agentv1alpha1.Agent](mgr.GetScheme(), v)
+	mgr.GetWebhookServer().Register(agentv1alpha1.WebhookPath, wh)
```
pebrc, line 27: **Passing `nil` here silently drops the annotation-based license check.**

The previous registration in `cmd/manager/validation.go` passed `checker` to `RegisterResourceWebhook`, which means `NewResourceValidator` ran `hasRequestedLicenseLevel` in its `preValidate` — i.e., a user setting `eck.k8s.elastic.co/license: enterprise` on the Agent resource forced an Enterprise license check at admission time. That gate is now gone for Agent unless the user also sets `spec.http.tls.client.authentication: true`.

The wrapper's checker and `validClientAuthentication` aren't redundant — they enforce different things:
- Wrapper: **annotation-driven** ("user marked this resource Enterprise → require Enterprise")
- `validClientAuthentication`: **feature-driven** ("this specific field is set → require Enterprise")

The accompanying comment "_the inner validator performs license enforcement itself_" is misleading because the inner validator only performs the feature-driven half.

The annotation mechanism is largely vestigial (legacy Helm-chart tagging), but it's still part of the API surface and I don't think we should remove it as a side effect of this PR. Suggest:

```go
inner := &webhookValidator{licenseChecker: checker}
v := commonwebhook.NewResourceValidator[*agentv1alpha1.Agent](checker, managedNamespaces, inner)
```

so the wrapper continues to enforce the annotation check and the inner validator continues to enforce the feature gate — both fire, comment can be deleted.

(Worth noting: the autoops webhook at `pkg/controller/autoops/validation/webhook.go:38` has the same pattern and the same regression — a separate cleanup PR could fix both.)

pkg/controller/agent/driver.go
```diff
@@ -215,6 +240,123 @@ func newService(agent agentv1alpha1.Agent, meta metadata.Metadata) *corev1.Servi
...
+	specEnabled := params.Agent.Spec.HTTP.TLS.Client.Authentication
+	tlsEnabled := params.Agent.Spec.HTTP.TLS.Enabled()
+
+	// Client auth is meaningless without TLS.
+	if !tlsEnabled {
+		if specEnabled {
+			return false, "spec.http.tls.client.authentication is ineffective because TLS is disabled on the Fleet Server HTTP interface", nil
+		}
+		return false, "", nil
+	}
```
pebrc: The `specEnabled && !tlsEnabled` branch is unreachable.

`checkClientAuthentication` in `pkg/apis/agent/v1alpha1/validations.go:313` rejects exactly this combination, and that check runs not only at the webhook but also at reconcile entry: `pkg/controller/agent/controller.go:203` calls `r.validate(ctx, agent)` → `agentv1alpha1.Validate(&agent, nil)` → all `defaultChecks`, including `checkClientAuthentication`. If validation fails the reconcile bails out at line 205 and `internalReconcile` is never reached (it has a single caller).

So by the time this code runs, `specEnabled=true` implies `tlsEnabled=true`. The inner branch and its warning string can never fire.

Suggest collapsing the block to just:

```go
if !tlsEnabled {
    return false, "", nil
}
```

Same reasoning applies to the implicit `FleetServerEnabled` precondition (validation rejects `client.authentication=true && !FleetServerEnabled`, and the caller at L145 already gates on `FleetServerEnabled` anyway). Worth a once-over for any other "double-check what the validator already enforced" leftovers in this function.

pebrc (commented):
Solid PR — the trust-bundle / soft-owner-label / annotation-signaling pattern from #9216 / #9234 extends cleanly to the Fleet-Server-as-server case. Most of my comments are polish (magic strings, naming, doc) and architecture-leaning suggestions for the next time this pattern lands on another CRD.

A few that are worth a closer look before merge:

- **`webhook.go:28`** — passing `nil` to `NewResourceValidator` silently drops the annotation-based license check (`eck.k8s.elastic.co/license: enterprise`). The mechanism is largely vestigial but still part of the API surface; suggest passing `checker` to both the wrapper and the inner validator.
- **`driver.go:262`** — when `authentication: true` is set on a non-Enterprise cluster (e.g., webhook bypassed via `failurePolicy: ignore`), the controller silently provisions Fleet Server without mTLS and emits no event. Worth a warning event mirroring the other ineffective-config branches.
- **`driver.go:266`** — the `specEnabled && !tlsEnabled` branch is unreachable: `controller.go:203` re-runs `defaultChecks` at reconcile entry, so validation has already rejected this combination by the time we get here. Same applies to the implicit `FleetServerEnabled` check.
- **`common.go:336`** — `FleetServerSelector` and `ElasticsearchSelector` are character-for-character duplicates modulo type names. Two copies is the cheapest moment to collapse them into one shared `AssociationRefWithClientCert` type.
- **e2e coverage** (`client_auth_test.go:278`) — only the user-provided cert path is covered. The auto-generated path (the recipe default) is the more common UX and is currently only verified by unit tests.

None are blockers; happy to chat through any of them.

pkg/controller/agent/driver.go
pebrc: Manual-test follow-up — narrowing the scope of this comment based on what actually happens.

Tested the license-revocation scenario on kind with the PR snapshot image: created the recipe with an Enterprise license, then deleted the license secret while the Fleet Server was running with `authentication: true`. Observed:

- License controller detects the change → `eck-licensing` configmap drops to `basic`
- On the next agent reconcile, `isFleetServerClientAuthRequired` correctly returns `(false, "", nil)` and the driver calls `DeleteClientCertResources`
- That helper's first step (`RemoveClientAuthenticationRequiredAnnotation`) makes a `client.Update()` on the Agent
- The operator's **own** validating webhook rejects the update because the spec still has `client.authentication: true` on Basic, surfacing a clear `ReconciliationError` event:
  ```
  admission webhook "elastic-agent-validation-v1alpha1.k8s.elastic.co" denied the request:
  spec.http.tls.client.authentication: Forbidden: client certificate authentication requires an enterprise license
  ```
- mTLS stays continuously active (pod not rolled, secrets not deleted) — no silent security downgrade

So the silent-downgrade concern only applies to the **create-bypass** path: webhook unavailable via `failurePolicy: ignore` + user applies a new Agent with `authentication: true` + no Enterprise license. In that case the annotation doesn't yet exist, `reconcileFleetServerClientAuth` short-circuits without any Agent update to be blocked, and the FS is silently provisioned without mTLS.

Suggest scoping the warning event to that specific case — i.e., emit it only when the annotation is *absent* and the spec asks for client auth without a license, rather than on every reconcile of an established resource. That keeps the noisy/safe license-revocation path quiet (it already surfaces via `ReconciliationError`) while still catching the genuine silent-downgrade scenario at create time.

pkg/controller/agent/driver.go
pkoutsovasilis: Good point but IMO these aren't exactly generic helper funcs and they are CRD specific:

For example:

Fleet Server-specific handling:
* Updates the configHash so pods are restarted (not hot-reloaded) when certificates change
* Includes the HTTP CA certificate in the trust bundle to allow the local Elastic Agent that bootstraps Fleet Server to communicate with it
* Detection logic (isFleetServerClientAuthRequired) checks FLEET_SERVER_CLIENT_AUTH env var in pod templates

Elasticsearch-specific handling:
* Different detection logic in detectClientAuthenticationRequired that reads from ES config/settings
* No configHash involvement (ES handles cert reload differently)
* Trust bundle composition differs

While at a high level both enable mTLS detection and management, the CRD-specific requirements make a shared helper awkward - it would need callbacks or configuration  for each divergent behavior. Keeping these in their respective controller packages feels more natural since the logic is tightly coupled to each CRD's specifics. So yes if a new server CRD supports mTLS in the future I would say it needs to tailor these two according to its specific needs. Does that make sense?

pkg/apis/common/v1/common.go
pkoutsovasilis: I'd prefer to keep these as separate types. For context, we previously had a single shared selector and ran into issues where ES-specific fields were exposed on associations that didn't need them.

Looking ahead, full PKI realm integration #9082 may involve mapping certificate Subject DNs to roles and other ES-specific configuration. Having distinct selectors gives us flexibility to add resource-specific fields without "polluting" unrelated associations - ElasticsearchSelector will likely need PKI-specific options that don't apply to FleetServerSelector.

IMO the ~40 lines of duplication feels acceptable given the flexibility it provides for future divergence but no strong feeling, let me know what you think

pkoutsovasilis:
buildkite test this -f p=gke,t=TestClientAuthRequiredCustomCertificate_FleetServerToAgent,E2E_TAGS=agent -m s=9.3.4,s=8.19.15,s=9.2.5,s=9.4.0,s=9.5.0-SNAPSHOT

pkg/controller/agent/driver.go
pebrc: Yes, that makes sense. I agree the detection logic should remain controller-specific, and if the reconciliation helper needs callbacks for config hash updates, trust bundle extras, reload semantics, and cleanup timing, it’s probably not buying much.

I think the current split is reasonable: common cert primitives in pkg/controller/common/certificates, with each controller deciding how to compose them. We can revisit later based on the actual overlap rather than abstracting around two divergent implementations now.

pkg/controller/agent/driver.go
pkoutsovasilis: ok makes sense went implemented the scoped warning in 2a6e5023c430835e2cb0d0defef43180668ab89d

pkg/controller/agent/driver.go
pkoutsovasilis: magic removed in 2a6e5023c430835e2cb0d0defef43180668ab89d

pkg/controller/agent/driver.go
pkoutsovasilis: symmetry augmented in 2a6e5023c430835e2cb0d0defef43180668ab89d

pkg/controller/agent/pod.go
pkoutsovasilis: renamed in 2a6e5023c430835e2cb0d0defef43180668ab89d

pkg/controller/agent/webhook.go
pkoutsovasilis: added comment  in 2a6e5023c430835e2cb0d0defef43180668ab89d

pkg/apis/agent/v1alpha1/validations.go
pkoutsovasilis: nice catch, introduced `CheckFleetServerSelectorRefs` in 2a6e5023c430835e2cb0d0defef43180668ab89d

pkg/apis/common/v1/common.go
pkoutsovasilis: added doc in 2a6e5023c430835e2cb0d0defef43180668ab89d

pkg/apis/agent/v1alpha1/agent_types.go
pkoutsovasilis: added extra sentence in 2a6e5023c430835e2cb0d0defef43180668ab89d

pkg/controller/agent/webhook.go
pkoutsovasilis: fixed in 2a6e5023c430835e2cb0d0defef43180668ab89d, I will open a separate PR for autoops

pkg/controller/agent/driver.go
pkoutsovasilis: unreachable branch removed in 2a6e5023c430835e2cb0d0defef43180668ab89d

pkoutsovasilis:
buildkite test this -f p=gke,t=.*_FleetServerToAgent,E2E_TAGS=agent -m s=9.3.4,s=8.19.15,s=9.2.5,s=9.4.0,s=9.5.0-SNAPSHOT

test/e2e/agent/client_auth_test.go
pkoutsovasilis: I added an e2e test with eck managed client certificates in 892a1d57b30bd76a61866a1943f1ee067a7c7635. I have also added the test steps to transition to disabled mTLS and when the agent issue is fixed I am gonna open a follow-up PR to uncomment them and version gate also the feature. Lemme know what you think on that

pkoutsovasilis: Created also this issue https://github.com/elastic/cloud-on-k8s/issues/9443 do keep track of the follow-up

pkoutsovasilis:
buildkite test this -f p=gke,t=.*_FleetServerToAgent,E2E_TAGS=agent -m s=9.3.4,s=8.19.15,s=9.2.5,s=9.4.0,s=9.5.0-SNAPSHOT

pkg/controller/agent/pod.go
```diff
@@ -644,6 +689,17 @@ func getFleetSetupFleetServerEnvVars(ctx context.Context, client k8s.Client) fun
 			fleetServerCfg[FleetServerPortEnv] = fmt.Sprintf("%d", FleetServerPort)
 		}
 
+		if clientAuthRequired {
+			// FLEET_CA is passed as --certificate-authorities to enrollment, which populates
+			// Server.TLS.CAs for client cert verification. Point it to the trust bundle so
+			// fleet-server can verify connecting agents' client certificates.
+			fleetServerCfg[FleetCA] = path.Join(FleetServerClientTrustBundleMountPath, certificates.ClientCertificatesTrustBundleFileName)
+			fleetServerCfg[FleetServerClientAuth] = "required"
```
pebrc: Magic-string cleanup from 2a6e5023c missed this call site — use the `FleetServerClientAuthRequired` constant defined at `pod.go:103` (already used in `driver.go:290`) instead of the literal `"required"`.

```suggestion
			fleetServerCfg[FleetServerClientAuth] = FleetServerClientAuthRequired
```

pkg/controller/agent/driver.go
```diff
@@ -215,6 +240,123 @@ func newService(agent agentv1alpha1.Agent, meta metadata.Metadata) *corev1.Servi
...
+		CertRotation: params.OperatorParams.CertRotation,
+	}
+
+	operatorClientCertSecret, err := certReconciler.ReconcileOperatorCertificate(params.Context)
+	if err != nil {
+		return results.WithError(err)
+	}
+
+	// Include the fleet HTTP CA in the trust bundle so the internal agent can verify the
+	// fleet-server's server certificate when FLEET_CA points to the trust bundle.
```
pebrc: Comment-vs-code mismatch: the comment says "fleet HTTP **CA**", but `ReconcileTrustBundle` reads `secret.Data[CertFileName]` (i.e. `tls.crt` — the fleet-server's leaf cert + intermediates), not `ca.crt`. The same wording is at line ~312 in the function header.

Functionally this is OK because the trust bundle is rebuilt every reconcile so the leaf stays current, and Go's TLS verification will accept a leaf-only trust store. But the comment should reflect what the code actually does — say "fleet HTTP cert" — otherwise a future reader will think `ca.crt` is being bundled and may break the behavior when refactoring. If we want true CA-based trust (less brittle across cert reissuance) we'd need to pass the CA cert instead, but I don't think that's needed for this PR.

pebrc (approved):
LGTM!

pkg/controller/agent/pod.go
pkoutsovasilis: oops missed that, nice catch fixed in c42843a0a6a90a6a4d170fbcfb8c51607a60b1d9

pkg/controller/agent/driver.go
pkoutsovasilis: changed in e8015d33e3932b96ca0e945eefbc282501b64746

pkoutsovasilis:
buildkite test this -f p=gke,t=.*_FleetServerToAgent,E2E_TAGS=agent -m s=9.3.4,s=8.19.15,s=9.2.5,s=9.4.0,s=9.5.0-SNAPSHOT

PR #9402, opened 2026-04-26
fix expectDesiredNodesAPI to also check nodeSet.Resources shorthand

pkoutsovasilis:
## Summary

Fixes `TestResourcesRequirements/Check_desired_nodes_API_state` E2E test failure introduced by #9346.

The `expectDesiredNodesAPI` function only checked for resources in `nodeSet.PodTemplate.Spec.Containers[].Resources`, but #9346 updated the test to use the new `nodeSet.Resources` shorthand. This caused the test to incorrectly expect no Desired Nodes API state (because PodTemplate resources were empty), while the operator correctly used the shorthand and called the API.

## Changes

- Added `getNodeSetResources()` helper that merges resources from both the PodTemplate and the new `Resources` shorthand field
- Updated `expectDesiredNodesAPI` to use this helper

### DIFF
--- a/test/e2e/test/elasticsearch/checks_es.go
+++ b/test/e2e/test/elasticsearch/checks_es.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"maps"
 	"strconv"
 	"strings"
 
@@ -177,17 +178,7 @@ func expectDesiredNodesAPI(es *esv1.Elasticsearch) (bool, error) {
 			}
 		}
 
-		var esResources *corev1.ResourceRequirements
-		for _, c := range nodeSet.PodTemplate.Spec.Containers {
-			if c.Name == "elasticsearch" {
-				c := c
-				esResources = &c.Resources
-			}
-		}
-		if esResources == nil {
-			// Elasticsearch container not found, very unlikely to happen in an E2E test.
-			return false, nil
-		}
+		esResources := getNodeSetResources(nodeSet)
 		memReq, hasMemReq := esResources.Requests[corev1.ResourceMemory]
 		memLimit, hasMemLimit := esResources.Limits[corev1.ResourceMemory]
 		if !hasMemLimit {
@@ -205,6 +196,35 @@ func expectDesiredNodesAPI(es *esv1.Elasticsearch) (bool, error) {
 		}
 	}
 	return true, nil
+}
+
+// getNodeSetResources returns the effective ResourceRequirements for the elasticsearch container,
+// checking both the Resources shorthand and the PodTemplate container resources.
+// The shorthand takes precedence over PodTemplate values for CPU and memory.
+func getNodeSetResources(nodeSet esv1.NodeSet) corev1.ResourceRequirements {
+	var result corev1.ResourceRequirements
+	for _, c := range nodeSet.PodTemplate.Spec.Containers {
+		if c.Name == esv1.ElasticsearchContainerName {
+			result = *c.Resources.DeepCopy()
+			break
+		}
+	}
+	if nodeSet.Resources.IsEmpty() {
+		return result
+	}
+	if limits := nodeSet.Resources.Limits.ToResourceList(); limits != nil {
+		if result.Limits == nil {
+			result.Limits = corev1.ResourceList{}
+		}
+		maps.Copy(result.Limits, limits)
+	}
+	if requests := nodeSet.Resources.Requests.ToResourceList(); requests != nil {
+		if result.Requests == nil {
+			result.Requests = corev1.ResourceList{}
+		}
+		maps.Copy(result.Requests, requests)
+	}
+	return result
 }
 
 func (e *esClusterChecks) CheckESNodesTopology() test.Step {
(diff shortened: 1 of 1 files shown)


prodsecmachine:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/6204e78f-079b-49ce-b7c5-e319129fa540?refs=6cef97a6889efc617d636abc42291eb7d2b64536&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/6204e78f-079b-49ce-b7c5-e319129fa540/license?refs=6cef97a6889efc617d636abc42291eb7d2b64536&source=prChecksComment) |

pkoutsovasilis:
buildkite test this -f p=gke,t=TestResourcesRequirements,E2E_TAGS=es

pkoutsovasilis:
buildkite test this -f p=gke,t=TestResourcesRequirements

test/e2e/test/elasticsearch/checks_es.go
```diff
@@ -207,6 +198,35 @@ func expectDesiredNodesAPI(es *esv1.Elasticsearch) (bool, error) {
 	return true, nil
 }
 
+// getNodeSetResources returns the effective ResourceRequirements for the elasticsearch container,
+// checking both the Resources shorthand and the PodTemplate container resources.
+// The shorthand takes precedence over PodTemplate values for CPU and memory.
+func getNodeSetResources(nodeSet esv1.NodeSet) corev1.ResourceRequirements {
```
pebrc: Naming nit: this returns the *effective* container resources after merging two sources, so something like `effectiveContainerResources(nodeSet)` would communicate intent better than the generic `get`. Not blocking.

test/e2e/test/elasticsearch/checks_es.go
```diff
@@ -207,6 +198,35 @@ func expectDesiredNodesAPI(es *esv1.Elasticsearch) (bool, error) {
...
+	var result corev1.ResourceRequirements
+	for _, c := range nodeSet.PodTemplate.Spec.Containers {
+		if c.Name == esv1.ElasticsearchContainerName {
+			result = *c.Resources.DeepCopy()
+			break
+		}
+	}
+	if nodeSet.Resources.IsEmpty() {
```
pebrc: Production's `WithResourcesAndOverrides` (`pkg/controller/common/defaults/pod_template.go`) injects `DefaultResources` (2Gi memory request+limit) when both PodTemplate and shorthand are empty. This helper does not mirror that. Today both paths still converge on `expectDesiredNodesAPI=false` for the empty/empty case (no CPU set), so it isn't a bug — but if `DefaultResources` ever gains a CPU value the two will diverge silently. A short comment here noting "operator defaults are intentionally not applied because they don't change the boolean outcome" would protect against that.

test/e2e/test/elasticsearch/checks_es.go
```diff
@@ -207,6 +198,35 @@ func expectDesiredNodesAPI(es *esv1.Elasticsearch) (bool, error) {
...
+// checking both the Resources shorthand and the PodTemplate container resources.
+// The shorthand takes precedence over PodTemplate values for CPU and memory.
+func getNodeSetResources(nodeSet esv1.NodeSet) corev1.ResourceRequirements {
+	var result corev1.ResourceRequirements
+	for _, c := range nodeSet.PodTemplate.Spec.Containers {
+		if c.Name == esv1.ElasticsearchContainerName {
+			result = *c.Resources.DeepCopy()
+			break
+		}
+	}
```
pebrc, line 221: The previous inline implementation had an explicit `return false, nil` when no `elasticsearch` container existed, with a comment explaining the case. The new helper silently returns a zero `ResourceRequirements`, which still resolves to `false` downstream via the `hasMemLimit` check — so functionally equivalent — but the explicit-intent guard was useful documentation. Consider adding a short comment that the missing-container case relies on the downstream `hasMemLimit` check to short-circuit.

pebrc (approved):
LGTM. The merge logic in `getNodeSetResources` correctly mirrors production's `WithResourcesAndOverrides` semantics (shorthand layered on top of PodTemplate, only CPU/memory keys touched), and the `DeepCopy` at line 208 is necessary because the helper later mutates the result maps via `maps.Copy`. Left a few non-blocking observations inline.

test/e2e/test/elasticsearch/checks_es.go
pkoutsovasilis: renamed in ae4b1e4acfb059529ca24d22b5af5508bbb8b94a

test/e2e/test/elasticsearch/checks_es.go
pkoutsovasilis: add a comment about that in ae4b1e4acfb059529ca24d22b5af5508bbb8b94a

test/e2e/test/elasticsearch/checks_es.go
pkoutsovasilis: changed the signature to return a boolean that reflects if the container was found to maintain the explicit return in ae4b1e4acfb059529ca24d22b5af5508bbb8b94a

PR #9408, opened 2026-04-27
Move Condition types from common/v1alpha1 to common/v1

dtuck9:
### Summary
                                                                                                                                                                                                                                                                                                                            
  The `Condition`, `Conditions`, and `ConditionType` types were living in `pkg/apis/common/v1alpha1/status.go` alongside the autoscaling API types, but they are general-purpose status primitives used across multiple v1 APIs. **This PR graduates the `v1alpha1.Conditions` to the GA, stable `pkg/apis/common/v1` package in preparation for extending the `Conditions` concept to stateless types.** 

### Reasoning
As part of https://github.com/elastic/cloud-on-k8s/issues/9293, the `Conditions` will be included on stateless types as well as stateful types. Elasticsearch was already completed in #9330, and the corresponding changes for stateless types are ongoing in #9398.

However, unrelated changes already merged to main, in combination with the changes in #9398, now create an import cycle:

```
$ go vet ./...
package github.com/elastic/cloud-on-k8s/v3/cmd
        imports github.com/elastic/cloud-on-k8s/v3/cmd/manager from main.go
        imports github.com/elastic/cloud-on-k8s/v3/pkg/apis/agent/v1alpha1 from main.go
        imports github.com/elastic/cloud-on-k8s/v3/pkg/apis/common/v1 from agent_types.go
        imports github.com/elastic/cloud-on-k8s/v3/pkg/apis/common/v1alpha1 from common.go
        imports github.com/elastic/cloud-on-k8s/v3/pkg/apis/common/v1 from resources.go: import cycle not allowed
```

The `Condition` type was originally introduced in 2022, and it seems like a reasonable time to graduate it to `v1` with the extension to stateless types.

I have opted to do this in a separate PR from #9398 for reviewability.
                                                                                                                                                                                                                                                                                                                            
  ### Changes                                                                                                                                                                                                                                                                                                               
                                                            
  **Core move**
  - Adds `Condition`, `Conditions`, `ConditionType`, and the `OrchestrationPaused` constant to `pkg/apis/common/v1/status.go`
  - Deletes `pkg/apis/common/v1alpha1/status.go`
  - Adds generated deepcopy methods for the new types in `pkg/apis/common/v1/zz_generated.deepcopy.go`                                                                                                                                                                                                                      
  - Removes the now-redundant deepcopy methods from `pkg/apis/common/v1alpha1/zz_generated.deepcopy.go`
                                                                                                                                                                                                                                                                                                                            
  **`common/v1alpha1` autoscaling types** (same package as the deleted file — required updates)
  - `autoscaling_status.go`: adds `commonv1` import; updates `ConditionType` constant declarations, the `ElasticsearchAutoscalerStatus.Conditions` field, and all `Condition{}`/`Conditions` literals to use the `commonv1.` qualifier                                                                                      
  - `autoscaling_status_test.go`: same qualification for test assertions                                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                                                            
  **Reference updates across the codebase**                                                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                            
  | File | Change |                                         
  |---|---|
  | `pkg/apis/elasticsearch/v1/status.go` | `v1alpha1.ConditionType` → `commonv1.ConditionType` for ES-specific constants |
  | `pkg/controller/elasticsearch/reconcile/status.go` | `commonv1alpha1` import → `commonv1` |                                                                                                                                                                                                                             
  | `pkg/controller/elasticsearch/reconcile/status_test.go` | same |
  | `pkg/controller/elasticsearch/reconcile/state_test.go` | same |                                                                                                                                                                                                                                                         
  | `pkg/controller/elasticsearch/elasticsearch_controller_test.go` | same |                                                                                                                                                                                                                                                
  | `pkg/controller/elasticsearch/driver/stateful/driver.go` | `esv1.OrchestrationPaused` → `commonv1.OrchestrationPaused` |
  | `pkg/controller/elasticsearch/driver/stateful/driver_test.go` | same |                                                                                                                                                                                                                                                  
  | `pkg/controller/autoscaling/elasticsearch/controller.go` | `v1alpha1.Condition{` → `commonv1.Condition{` |
  | `test/e2e/es/pause_orchestration_test.go` | `esv1.OrchestrationPaused` → `commonv1.OrchestrationPaused` |                                                                                                                                                                                                               
  | `test/e2e/test/elasticsearch/autoscaling/steps.go` | `v1alpha1.ConditionType` → `commonv1.ConditionType` |                                                                                                                                                                                                              
                                                                                                                                                                                                                                                                                                                            
  ### Notes                                                                                                                                                                                                                                                                                                                 
                                                            
  - The autoscaling-specific `ConditionType` constants (`ElasticsearchAutoscalerActive`, `ElasticsearchAutoscalerHealthy`, `ElasticsearchAutoscalerLimited`, `ElasticsearchAutoscalerOnline`) remain in `common/v1alpha1` — only the base types moved.                                                                      
  - The Elasticsearch-specific constants (`ElasticsearchIsReachable`, `ReconciliationComplete`, `ResourcesAwareManagement`, `RunningDesiredVersion`, `OrchestrationPaused`) remain in their respective packages; only their type annotation changes from `v1alpha1.ConditionType` to `commonv1.ConditionType`.
  - No API schema changes — the JSON serialization is identical. 

### AI disclaimer
Claude code used to perform the majority of the migration.


prodsecmachine:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/87efcbfd-5bac-4ed7-9928-6619ec3c99ee?refs=df2798af06f395467d4d96c8680849d96aa77c72&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/87efcbfd-5bac-4ed7-9928-6619ec3c99ee/license?refs=df2798af06f395467d4d96c8680849d96aa77c72&source=prChecksComment) |

kvalliyurnatt (approved):
LGTM code wise. IIUC this has no impact for existing users for upgrade since we are not changing any actual CRDs.

dtuck9:
> LGTM code wise. IIUC this has no impact for existing users for upgrade since we are not changing any actual CRDs.

@kvalliyurnatt I believe that's correct as the `Condition` should serialize/deserialize identically. Is there a way to test the upgrade path though by any chance?

pkg/apis/elasticsearch/v1/status.go
```diff
@@ -106,11 +104,10 @@ func (es *Elasticsearch) SetAssociationStatusMap(typ commonv1.AssociationType, s
 }
 
 const (
-	ElasticsearchIsReachable v1alpha1.ConditionType = "ElasticsearchIsReachable"
-	ReconciliationComplete   v1alpha1.ConditionType = "ReconciliationComplete"
-	ResourcesAwareManagement v1alpha1.ConditionType = "ResourcesAwareManagement"
-	RunningDesiredVersion    v1alpha1.ConditionType = "RunningDesiredVersion"
-	OrchestrationPaused      v1alpha1.ConditionType = "OrchestrationPaused"
```
naemono, line 113: So this is being removed from a "V1" package. I don't think that's allowed, even if it's moved to another v1 package.

naemono (changes requested):
I'm sorry I wasn't able to take a long look at this, but something stood out quickly where changes to a v1 api are not allowed. 

Overall I think the `common/v1alpha1` promotion to v1 makes sense, as many of it's constructs are 3+ years old.

pkg/apis/elasticsearch/v1/status.go
```diff
@@ -68,8 +67,7 @@ type ElasticsearchStatus struct {
 
 	// +optional
 	// Conditions holds the current service state of an Elasticsearch cluster.
-	// **This API is in technical preview and may be changed or removed in a future release.**
-	Conditions v1alpha1.Conditions `json:"conditions"`
+	Conditions commonv1.Conditions `json:"conditions"`
```
pebrc, line 70: Worth calling out explicitly in response to @naemono's concern: this changes the Go type of `ElasticsearchStatus.Conditions` from `v1alpha1.Conditions` to `commonv1.Conditions`, but the on-the-wire JSON shape is identical (verified via the CRD diff — only the technical-preview prose is removed from the description fields). So from a Kubernetes API perspective there is no breaking change to the `elasticsearch.k8s.elastic.co/v1` schema.

The technical-preview comment was effectively the stability disclaimer for this field. Removing it here is the explicit graduation of `Conditions` to a stable v1 API. Worth saying that in the PR description so reviewers don't have to infer it.

There is still a Go-level break for any external consumer that imported `v1alpha1.Condition`, `v1alpha1.Conditions`, `v1alpha1.ConditionType`, or `esv1.OrchestrationPaused` directly — but `v1alpha1` carries no stability promise. A release-note entry would be a courtesy for downstream Go importers.

pkg/controller/autoscaling/elasticsearch/controller.go
```diff
@@ -22,6 +22,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	autoscalingv1alpha1 "github.com/elastic/cloud-on-k8s/v3/pkg/apis/autoscaling/v1alpha1"
+	commonv1 "github.com/elastic/cloud-on-k8s/v3/pkg/apis/common/v1"
```
pebrc, line 25: Minor: this file now imports both `commonv1` (for the `Condition` struct) and `v1alpha1` (for `ElasticsearchAutoscalerActive` etc., which intentionally stay in v1alpha1 since they're autoscaler-specific). Acceptable, just slightly noisy — readers will see `commonv1.Condition{Type: v1alpha1.ElasticsearchAutoscalerActive, ...}` and may wonder if the split is intentional. A short comment near one of these literals (or in the PR description) clarifying that the autoscaler-specific condition types deliberately stayed in v1alpha1 would help future readers.

pkg/apis/common/v1/status.go
pebrc, line 1: Nit: now that `Conditions.Index` and `Conditions.MergeWith` live in `common/v1`, a focused unit test next to this file would be cleaner. Today they're only exercised transitively via `pkg/apis/common/v1alpha1/autoscaling_status_test.go`, which leaves the new package without its own test coverage. Not a blocker — coverage is preserved — but worth doing while we're touching this.

pebrc (commented):
Mechanical refactor that does what it says — JSON wire format is identical (verified via CRD diff: only the technical-preview prose is removed from the description fields), so the Kubernetes-level v1 API is unchanged.

A few non-blocking comments inline. The main one is around @naemono's outstanding concern: I think it's worth replying with the on-the-wire equivalence reasoning and explicitly framing this PR as the graduation of `Conditions` from technical-preview to stable v1, rather than leaving reviewers to infer it. The Go-level break (for external importers of `v1alpha1.Condition*` or `esv1.OrchestrationPaused`) deserves a release-note entry as a courtesy.

Tests pass locally on the affected packages (`pkg/apis/common/...`, `pkg/apis/elasticsearch/...`, `pkg/controller/elasticsearch/reconcile/...`, `pkg/controller/elasticsearch/driver/stateful/...`, `pkg/controller/autoscaling/...`).

pkg/apis/common/v1alpha1/status.go
pebrc, line 1: Following up on my own earlier comment about the Go-level break — on reflection I think it deserves more than a release-note entry.

The population affected here isn't really "people who imported v1alpha1 directly" — it's "people who consume `Elasticsearch.Status.Conditions`," which is a stable v1 API field. Until this PR, the only way to name that field's type was `v1alpha1.Conditions`. So anyone building an external controller, kubectl plugin, or programmatic consumer that reads ES status was effectively forced through v1alpha1, even though they were consuming a v1 API. The "v1alpha1 has no stability promise" defense is technically correct but practically weak in this specific case.

Suggestion: keep `pkg/apis/common/v1alpha1/status.go` as a thin shim of type aliases plus a deprecation comment, rather than deleting it outright:

```go
package v1alpha1

import commonv1 "github.com/elastic/cloud-on-k8s/v3/pkg/apis/common/v1"

// Deprecated: use commonv1.Condition. Will be removed in a future release.
type Condition = commonv1.Condition
// Deprecated: use commonv1.Conditions. Will be removed in a future release.
type Conditions = commonv1.Conditions
// Deprecated: use commonv1.ConditionType. Will be removed in a future release.
type ConditionType = commonv1.ConditionType
```

And similarly in `pkg/apis/elasticsearch/v1/status.go`:

```go
// Deprecated: use commonv1.OrchestrationPaused. Will be removed in a future release.
const OrchestrationPaused = commonv1.OrchestrationPaused
```

Cost: ~15 LoC and one extra import. Benefits: existing external importers keep compiling, `// Deprecated:` lights up in `staticcheck`/IDE tooling so they get a clear migration signal, and we can drop the shim in a later minor release (e.g. 3.5 or 3.6) once consumers have had a window to migrate. Type aliases don't require regenerating deepcopy and don't change the on-the-wire behavior.

Happy to be talked out of this if there's a reason the hard break is preferable, but my default would be to ship the shim.

Props to @naemono to challenge my original assertion.

pebrc (commented):
Follow-up on my earlier review: I want to revise the "release-note entry is enough" advice. The Go-level break here is more impactful than I first framed it, because the only path to naming the type of `Elasticsearch.Status.Conditions` (a stable v1 field) was via v1alpha1. Proposing a small type-alias shim with deprecation markers as a softer migration path — see inline.

pkg/apis/common/v1alpha1/status.go
pebrc: Actually after some more thought I don't think the shimming will help. As we still have the breakage in the controller status. Maybe a breaking change notification in the release notes is acceptable though given that this is an edge case of what k8s API compatibility demands.

dtuck9: Worth also pointing out that the alias for `Condition` to `commonv1.Condition` would still reference the `commonv1.ConditionType` in the `ConditionType`, and is therefore still a breaking change. **I imagine if users _are_ referencing these `Conditions`, the `ConditionType` is most likely the field they're referencing.**

Re: 
> so they get a clear migration signal, and we can drop the shim in a later minor release (e.g. 3.5 or 3.6) once consumers have had a window to migrate

Additionally, they wouldn't actually be able to migrate because the Elasticsearch CRD would still have the `v1alpha.Conditions` and we would basically just be giving them advanced notice that there will be a breaking change in a few releases without giving them anything to actually migrate to.

dtuck9: We can also resolve the import cycle if we move `ToNodeSetResourcesWith`:
https://github.com/elastic/cloud-on-k8s/blob/f468e123175ea6c38556e1b985f2d85f327e1750/pkg/apis/common/v1alpha1/resources.go#L95-L116


 to the `v1` package in v1/ resources.go with some slight refactoring, e.g.
```
// ToNodeSetResourcesWith builds NodeSet shorthand resources by overriding existing CPU/memory values
// with the values present in this NodeResources recommendation.
// If a recommendation for a given resource is missing, the existing value is preserved.
func ToNodeSetResourcesWith(nr v1alpha1.NodeResources, existing Resources) Resources {
	merged := *existing.DeepCopy()

	if nr.HasRequest(corev1.ResourceCPU) {
		request := nr.GetRequest(corev1.ResourceCPU)
		merged.Requests.CPU = ptr.To(request)
	}
	if nr.HasRequest(corev1.ResourceMemory) {
		request := nr.GetRequest(corev1.ResourceMemory)
		merged.Requests.Memory = ptr.To(request)
	}
	if nr.HasLimit(corev1.ResourceCPU) {
		limit := nr.GetLimit(corev1.ResourceCPU)
		merged.Limits.CPU = ptr.To(limit)
	}
	if nr.HasLimit(corev1.ResourceMemory) {
		limit := nr.GetLimit(corev1.ResourceMemory)
		merged.Limits.Memory = ptr.To(limit)
	}

	return merged
}
```

LMK if that would be the preferred approach (it's certainly a lot smaller), and I'll update this PR accordingly.

pkg/apis/elasticsearch/v1/status.go
dtuck9: Are you referring specifically to the `OrchestrationPaused` condition type? If so, it hasn't been released yet -- it will be released in 3.5.0 -- so I don't see a problem in moving it now before this feature is released (and imo is necessary now before it's released because stateless controllers will also reference this `ConditionType`, as the concept of `Conditions` is being extended to the stateless controllers as a part of this feature).
 
If you're referring to the overall `Conditions` promotion -- I don't think there's a way around making that some degree of breaking aside from having redundant `Conditions` on the `ElasticsearchStatus`, such that `v1alpha1.Conditions` and `v1.Conditions` coexist for a period of time _or_ creating a v2 `Elasticsearch` CRD that has the new `v1.Conditions` on it and leaving the existing v1 CRD untouched. 

The latter may be the more technically correct way to manage this, but it seems a bit heavy handed for a situation in which the on-the-wire format hasn't actually changed.

pkg/apis/elasticsearch/v1/status.go
dtuck9: Do you have suggestions on how to make the PR description more explicit? I have this verbiage in the top **Summary** section:

> This PR graduates them to pkg/apis/common/v1 in preparation for extending the Conditions concept to stateless types.

dtuck9: I've updated the PR description to bold the above phrase with some additional verbiage tweaks. LMK if that is sufficient.

pkg/apis/common/v1/status.go
```diff
@@ -43,6 +43,11 @@ func (c Conditions) Index(conditionType ConditionType) int {
 }
 
 func (c Conditions) MergeWith(nextConditions ...Condition) Conditions {
+	if c == nil {
+		// deliberately set this to 0 rather than len(nextConditions) so that items are not prematurely
+		// set to empty Conditions
+		c = make(Conditions, 0)
+	}
```
dtuck9, line 48: Drive-by fix: I don't think `c` is ever nil in practice, but I did add a test case for that condition which of course caused a panic, so better safe than sorry.

pkg/controller/autoscaling/elasticsearch/controller.go
dtuck9: Added the explanation for autoscaling-specific `ConditionType` [here](https://github.com/elastic/cloud-on-k8s/pull/9408/changes/1ab7e98703eecc01c54374c56ebf1900dc6d0c68#diff-3a468729ad38582666b9dc853e089fad4a69834ccf0f1f258f35242e24a77b09R36-R37)

pkg/apis/common/v1/status.go
dtuck9: Added test cases, discovered and fixed pre-existing nil-pointer risk.

dtuck9:
buildkite test this -f p=gke,t=TestPauseOrchestration*

pkg/apis/elasticsearch/v1/status.go
naemono: Sorry, I missed this reply. `OrchestrationPaused` is what I was referring to, but since it hasn't been released, I see this is fine.

naemono (approved):
This seems good to me now, after the explanation concerning `OrchestrationPaused`.

Issue #2973, opened 2020-04-28
Consul Connect and ECK


charith-elastic:
Consul Connect seems to operate in a different way from other service meshes as it requires applications to opt-in to the mesh. As far as I can tell, there are no built-in provisions for transparent proxying of _all_ traffic like Istio or Linkerd.  The implementation patterns appear to be the following:

- [Use Consul as the cluster DNS resolver](https://www.consul.io/docs/platform/k8s/dns.html):  Existing Kubernetes services can be added to Consul automatically via [catalog sync](https://www.consul.io/docs/platform/k8s/service-sync.html) but as far as I can tell, they still need to be accessed with the special suffix `service.consul` instead of the `cluster.local` suffix. [needs more research to confirm]
- [Use the Connect sidecar](https://www.consul.io/docs/platform/k8s/connect.html): This requires the applications to declare the services they need access to by using the `consul.hashicorp.com/connect-service-upstreams` annotation. Each upstream service must have a unique port that the application can then access by sending the requests to `127.0.0.1:<port>`.  Services not declared in the annotation cannot be accessed through the Connect proxy. [Dynamic upstreams require native integration with Connect](https://www.consul.io/docs/connect/proxies.html#dynamic-upstreams-require-native-integration).
- [Native integration](https://www.consul.io/docs/connect/native.html): Connect Go library can be used instead of the proxy sidecar to access services. This obviously requires applications to be built specifically with the knowledge that they are being deployed to a Consul Connect environment.

The above integration methods make it difficult for the ECK operator to be fully integrated into the mesh like it does with Istio and Linkerd. The operator needs to access the Elastic Stack applications that it creates -- which are not known beforehand. Declaring the upstream services using the `consul.hashicorp.com/connect-service-upstreams` annotation is therefore not possible. Furthermore, the operator accesses Stack applications using the Kubernetes DNS names of the respective services. So far, it seems that in order to go through the mesh, the operator needs to use the Consul service DNS name (or in the case of the proxy, `127.0.0.1:<port>` -- which becomes even more complicated if there is more than one deployment of an application as each one will need to be manually configured to use a unique port number instead of the application default). 

Users who wish to have full enmeshment (all application communications going through the mesh) will have a hard time configuring ECK to do so. Even if they know the exact set of Stack applications they are going to deploy and can list them using the above-mentioned annotation, the operator itself needs to be changed to use the special DNS name or the port number to access the Stack applications through the mesh. 

[...]

Issue #2939, opened 2020-04-23
Green health is reported while it should be unknown


barkbay:
I got into a situation where the reported status was always `green`:

```
NAME                                              HEALTH   NODES   VERSION   PHASE             AGE
elasticsearch.elasticsearch.k8s.elastic.co/es-0   green    5       7.6.2     ApplyingChanges   7m42s
```

While my cluster was broken with only 1 master available out of 3, which means that the status should have been `unknown` I guess:

```
# curl -k -v -u elastic-internal-probe:redacted https://127.0.0.1:9200/_cluster/health
< HTTP/1.1 503 Service Unavailable
```

anyasabo:
I can reproduce this. In my 3 node GKE cluster, I set hard anti affinity rules on the masters, let it build. I then tainted two of the nodes and killed two of the pods. The ES resource stayed green/Ready afterwards, even though it was reconciling it

>{"log.level":"info","@timestamp":"2020-06-09T19:36:16.712Z","log.logger":"generic-reconciler","message":"Recoverable error during step, continuing","service.version":"1.2.0-2bc08ef2","service.type":"eck","ecs.version":"1.4.0","step":"reconcile-cluster-license","error":"failed to revert to basic: 503 Service Unavailable: unknown","errorVerbose":"503 Service Unavailable


>{"log.level":"error","@timestamp":"2020-06-09T19:35:16.311Z","log.logger":"driver","message":"Could not update remote clusters in Elasticsearch settings","service.version":"1.2.0-2bc08ef2","service.type":"eck","ecs.version":"1.4.0","namespace":"default","es_name":"elasticsearch-aff","error":"503 Service Unavailable: ","error.stack_trace":"github.com/go-logr/zapr.(*zapLogger).Error\n\t/go/pkg/mod/github.com/go-logr/zapr@v0.1.0/zapr.go:128\ngithub.com/elastic/cloud-on-k8s/pkg/controller/elasticsearch/driver.(*defaultDriver).Reconcile\n\t/go/src/github.com/elastic/cloud-on-k8s/pkg/controller/elasticsearch/driver/driver.go:228

[...]

PR #9428, opened 2026-05-08
observer: add retry with random pod selection for health checks

sealemar:
## Problem

When the ECK operator runs behind a service mesh sidecar (Istio/Envoy, Consul Connect),
the health observer's `_cluster/health` calls route through the proxy. During rolling
upgrades, transient failures (sidecar returning 503 for a terminated-but-not-yet-removed
endpoint, or `URLProvider` picking a pod whose IP is stale in the informer cache) cause
the observer to reset cluster health to `unknown`.

Because `esReachable = hasEndpoints && hasKnownHealthState`, a single failed observation
sets `esReachable=false`, blocking the rolling upgrade until the next observation tick
(~10s default).

## Changes

A ~16 line change in `observe()`:

- Single context with the full `ObservationInterval` budget is shared across two attempts.
- If the first attempt returns `ElasticsearchUnknownHealth` AND the budget hasn't been
  exhausted (`totalCtx.Err() == nil`), retry once.
- `URLProvider.URL()` selects a fresh random pod on each call, so the retry typically
  targets a different endpoint.

No interface changes. No behavior change when the budget is exhausted (slow timeouts →
retry is automatically skipped).

## Why a retry, given URLProvider already randomizes?

`URL()` does select a fresh random pod, but only at the next observation tick (~10s later).
The retry collapses that ~10s upgrade-pause window to milliseconds for transient failures.

For service-mesh deployments (e.g., Istio), the structural argument is more subtle: the
operator's traffic is routed via per-pod-DNS to an Envoy `ORIGINAL_DST` cluster with one
effective endpoint. Envoy's default route retry policy (`num_retries: 2`,
`retry_host_predicate: previous_hosts`) cannot help — there's no alternative host within
a single-endpoint cluster. The operator-side retry escapes by going to a different
pod-DNS → different cluster.

See the [response to review](https://github.com/elastic/cloud-on-k8s/pull/9428#issuecomment-4491767130)
for production telemetry and a live Envoy config dump backing this analysis.

## Tests

Unit tests: 3 new tests in `observer_test.go` covering retry-succeeds-on-second-attempt,
both-attempts-fail, and retry-skipped-after-timeout.

## Related Issues

- #2973 — User reported hours-long `ApplyingChanges` with Consul Connect service mesh
  (same root cause, different mesh)
- #2939 / #4938 — The `esReachable` check was introduced to fix health reported as
  green when ES was actually down. This PR addresses the sensitivity of that check to
  transient failures during rolling upgrades.
- #9165 — Similar retry-resilience gap in the EDOT collector component
- elastic/docs-content#6577 — documents the Istio sidecar port exclusion (mesh-side mitigation that complements this PR)

### DIFF
--- a/pkg/controller/elasticsearch/client/base.go
+++ b/pkg/controller/elasticsearch/client/base.go
@@ -45,15 +45,6 @@ func (c *baseClient) Close() {
 	}
 }
 
-// CloseIdleConnections closes idle connections in the underlying HTTP transport
-// without fully closing the client. Useful for discarding stale connections to
-// terminated pods before retrying a request.
-func (c *baseClient) CloseIdleConnections() {
-	if c.HTTP != nil {
-		c.HTTP.CloseIdleConnections()
-	}
-}
-
 func (c *baseClient) equal(c2 *baseClient) bool {
 	// handle nil case
 	if c2 == nil {
--- a/pkg/controller/elasticsearch/client/base.go
+++ b/pkg/controller/elasticsearch/client/base.go
@@ -45,6 +45,15 @@ func (c *baseClient) Close() {
 	}
 }
 
+// CloseIdleConnections closes idle connections in the underlying HTTP transport
+// without fully closing the client. Useful for discarding stale connections to
+// terminated pods before retrying a request.
+func (c *baseClient) CloseIdleConnections() {
+	if c.HTTP != nil {
+		c.HTTP.CloseIdleConnections()
+	}
+}
+
 func (c *baseClient) equal(c2 *baseClient) bool {
 	// handle nil case
 	if c2 == nil {
--- a/pkg/controller/elasticsearch/client/client.go
+++ b/pkg/controller/elasticsearch/client/client.go
@@ -68,6 +68,10 @@ type Client interface {
 	SecurityClient
 	// Close idle connections in the underlying http client.
 	Close()
+	// CloseIdleConnections closes idle connections in the underlying HTTP transport
+	// without fully closing the client. This is useful to discard stale connections
+	// to terminated pods before retrying a health check.
+	CloseIdleConnections()
 	// Equal returns true if other can be considered as the same client.
 	Equal(other Client) bool
 	// GetClusterInfo get the cluster information at /
--- a/pkg/controller/elasticsearch/client/client.go
+++ b/pkg/controller/elasticsearch/client/client.go
@@ -68,10 +68,6 @@ type Client interface {
 	SecurityClient
 	// Close idle connections in the underlying http client.
 	Close()
-	// CloseIdleConnections closes idle connections in the underlying HTTP transport
-	// without fully closing the client. This is useful to discard stale connections
-	// to terminated pods before retrying a health check.
-	CloseIdleConnections()
 	// Equal returns true if other can be considered as the same client.
 	Equal(other Client) bool
 	// GetClusterInfo get the cluster information at /
--- a/pkg/controller/elasticsearch/observer/manager_test.go
+++ b/pkg/controller/elasticsearch/observer/manager_test.go
@@ -162,8 +162,9 @@ func TestManager_ObserveSync(t *testing.T) {
 			manager: NewManager(-1*time.Second, nil),
 			expectedHealth: []esv1.ElasticsearchHealth{
 				esv1.ElasticsearchGreenHealth,
-				// the flapping client returns an error on the second request
-				esv1.ElasticsearchUnknownHealth,
+					// the flapping client returns an error on the second request,
+					// but the grace period keeps the previous known health
+					esv1.ElasticsearchGreenHealth,
 			},
 		},
 		{
--- a/pkg/controller/elasticsearch/observer/manager_test.go
+++ b/pkg/controller/elasticsearch/observer/manager_test.go
@@ -162,9 +162,9 @@ func TestManager_ObserveSync(t *testing.T) {
 			manager: NewManager(-1*time.Second, nil),
 			expectedHealth: []esv1.ElasticsearchHealth{
 				esv1.ElasticsearchGreenHealth,
-					// the flapping client returns an error on the second request,
-					// but the grace period keeps the previous known health
-					esv1.ElasticsearchGreenHealth,
+				// the flapping client returns an error on the second request,
+				// but the retry with fresh URL selection recovers on the third call
+				esv1.ElasticsearchGreenHealth,
 			},
 		},
 		{
--- a/pkg/controller/elasticsearch/observer/observer.go
+++ b/pkg/controller/elasticsearch/observer/observer.go
@@ -35,18 +35,25 @@ const defaultObservationTimeout = 10 * time.Second
 // OnObservation is a function that gets executed when a new state is observed
 type OnObservation func(cluster types.NamespacedName, previousHealth, newHealth esv1.ElasticsearchHealth)
 
+// consecutiveFailureThreshold is the number of consecutive health check failures required
+// before the observer downgrades the cluster health to unknown. This prevents transient
+// network issues (e.g., envoy sidecar returning 503 during pod termination) from immediately
+// resetting the cluster health and stalling rolling upgrades.
+const consecutiveFailureThreshold = 3
+
 // Observer regularly requests an ES endpoint for cluster state,
 // in a thread-safe way
 type Observer struct {
-	cluster       types.NamespacedName
-	esClient      esclient.Client
-	settings      Settings
-	creationTime  time.Time
-	stopChan      chan struct{}
-	stopOnce      sync.Once
-	onObservation OnObservation
-	lastHealth    esv1.ElasticsearchHealth
-	mutex         sync.RWMutex
+	cluster             types.NamespacedName
+	esClient            esclient.Client
+	settings            Settings
+	creationTime        time.Time
+	stopChan            chan struct{}
+	stopOnce            sync.Once
+	onObservation       OnObservation
+	lastHealth          esv1.ElasticsearchHealth
+	consecutiveFailures int
+	mutex               sync.RWMutex
 }
 
 // NewObserver creates and starts an Observer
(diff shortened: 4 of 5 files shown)


prodsecmachine:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/95771cc3-fe14-4729-bc98-d3b1173e7348?refs=fe9ed316337db38e8617b8b5c6e5c0913046c05b&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/95771cc3-fe14-4729-bc98-d3b1173e7348/license?refs=fe9ed316337db38e8617b8b5c6e5c0913046c05b&source=prChecksComment) |

pkoutsovasilis (commented):
Thanks for the contribution @sealemar. I have to be honest and say that I am skeptical about this PR. Let me explain my concerns:

#### 1. Delaying the `Unknown` health state weakens an intentional safety gate

The `consecutiveFailureThreshold` masking has a concrete downstream effect. `LastHealth()` feeds `hasKnownHealthState` in `driver/shared/reconcile.go:238-239`:

```go
hasKnownHealthState := observedState() != esv1.ElasticsearchUnknownHealth
esReachable        := hasEndpoints && hasKnownHealthState
```

`esReachable` is the gate at `driver/stateful/nodes.go:168` that stops rolling upgrades, downscales and voting-exclusion clearing from running while the cluster API is unreachable. With this PR, two consecutive genuinely-unreachable observations report the *previous* known health, so `esReachable` stays `true` while ES is in fact unreachable. Downstream destructive paths each issue live ES calls and will individually fail, so there's no silent corruption - but the calm "Elasticsearch cannot be reached yet, re-queuing" short-circuit is replaced with a stream of error events, the `ElasticsearchIsReachable` condition is stale for up to `2 × ObservationInterval`, and a slowly-degrading cluster that flaps Green → Unknown → Green → Unknown will never surface its Unknown state because the counter resets on each Green observation. The existing binary "we don't know, don't proceed" signal is deliberate and I'd rather not weaken it globally to paper over one mesh-specific failure mode.

#### 2. `CloseIdleConnections()` duplicates the existing `Close()`

Looking at `pkg/controller/elasticsearch/client/base.go`, the two methods are byte-for-byte identical:

```go
func (c *baseClient) Close() {
    if c.HTTP != nil { c.HTTP.CloseIdleConnections() }
}

func (c *baseClient) CloseIdleConnections() {
    if c.HTTP != nil { c.HTTP.CloseIdleConnections() }
}
```

Adding `CloseIdleConnections()` to the public `Client` interface for one caller is not justified when `Close()` already does the same thing. If the close-before-retry is truly needed, `o.esClient.Close()` suffices with no interface change.

#### 3. `services.go:urlProvider.URL()` already selects a random pod

`services.go:urlProvider.URL()` already picks a random ready pod on every call - that's the existing behavior before this PR. So on the retry the HTTP transport opens a fresh connection to whichever pod `URL()` returns on its own, making `CloseIdleConnections()` between attempts redundant. If we land on the same pod again, closing idle connections doesn't help either. It's not clear what problem the retry is actually solving that the existing random selection and the next observation tick don't already cover.

---

I'm not convinced this is an operator problem, it feels more like compensating for a service-mesh misbehavior/misconfiguration. Even if it is I don't think this is how we should solve it in this way.

sealemar:
@pkoutsovasilis , thank you for the thorough review. I am running additional validations and preparing a response.

sealemar:
Thanks @pkoutsovasilis for the review. Below is a response to each concern, with supporting evidence from production telemetry, code, and a live Envoy config dump.

## Concerns 1 & 2: addressed by removal

The latest commit removes both:

- **Grace period / `consecutiveFailureThreshold`** — agreed it weakens the `esReachable` safety gate. A single failed observation now still sets `Unknown` immediately.
- **`CloseIdleConnections()` interface method** — agreed it duplicates `Close()`. Removed from the interface; the retry no longer touches the connection pool.

The remaining change is ~16 lines in `observe()`: a single context with the full `ObservationInterval` budget shared across two attempts. Retry is skipped if the budget is exhausted (`totalCtx.Err() != nil`).

## Concern 3: what does the retry actually solve?

You wrote:
> `services.go:urlProvider.URL()` already picks a random ready pod on every call ... It's not clear what problem the retry is actually solving that the existing random selection and the next observation tick don't already cover.

From a live Envoy config dump on an operator pod with Istio sidecar:

```
cluster.outbound|9200||elasticsearch-es-default.es-prod.svc.cluster.local:
  connect_timeout: 10s
  type: ORIGINAL_DST

route.retry_policy:
  num_retries: 2
  retry_on: connect-failure,refused-stream,unavailable,...
  retry_host_predicate: previous_hosts
  host_selection_retry_max_attempts: 5
```

When the operator hits `elasticsearch-es-default-N....:9200`, that pod-specific DNS name maps to its own `ORIGINAL_DST` cluster with one effective endpoint (the IP from DNS). Envoy's default retry policy attempts 3 connections per request, but the `previous_hosts` predicate has no alternative host within a single-endpoint cluster — all 3 attempts target the same stale IP. The slow-fail population in production has a tight ~9.23s distribution (p50=9236ms, p99=9289ms, stdev=64ms across 1.1M events), consistent with a deterministic timer rather than network jitter.

The operator-side retry is structurally different: each `URL()` call selects a fresh random pod. With probability `(N-1)/N` (where N is the ready pod count) the retry targets a different pod-DNS, resolving to a different IP. With Envoy in the path that means a different `ORIGINAL_DST` cluster entirely. The existing Envoy retries cannot do this because they're scoped to one cluster.

The next observation tick (~10s later) would also use a different random pod, but during those 10s `lastHealth=Unknown` triggers `WatchClusterHealthChange` (`watch.go:35-49`) → reconcile → `esReachable=false` (`reconcile.go:240`) → upgrade re-queued (`stateful/nodes.go:168`). The retry collapses this window to milliseconds.

## Production telemetry

In 90 days of production telemetry (~81M operator health-check requests through Istio sidecars):

| Bucket | Errors | % | Cause |
|---|---|---|---|
| Slow (8-10s) | 2,676,245 | 98.9% | `503` with `delayed_connect_error_113` (EHOSTUNREACH) |
| Fast (<100ms) | 27,734 | 1.0% | `no_healthy_upstream` / `via_upstream` 503 |

When the first attempt returns 503 at ~9.23s, `totalCtx` (10s) still has ~770ms remaining → retry fires. Successful health-check durations across 836,276 events have p50=7ms and p99=96ms; 99.999% complete in under 770ms.

## Port exclusion and retry are complementary

Slow-fail is addressed mesh-side by `excludeOutboundPorts: '9200'` on the operator's Istio sidecar, which removes the 9.2s Envoy retry tax. Documentation for that mitigation will be contributed in a separate PR.

The retry in this PR provides same-tick recovery via fresh random pod selection. Its value is independent of whether the path goes through Envoy: with Envoy, it escapes the single-host `ORIGINAL_DST` cluster; without Envoy, it recovers from an operator-side stale-cache race (`urlProvider` picks a pod the informer believes is Ready but whose IP is gone). Neither port exclusion nor retry alone fully covers both cases.

## On "operator problem vs mesh misbehavior"

The slow-fail population we measured is mesh-induced; mesh-side documentation will be contributed separately. The 16-line retry in this PR — no interface change, no behavior change when budget is exhausted — provides same-tick recovery for stale-cache races regardless of whether the mesh is in the path.

sealemar:
I have also updated the title of this PR and the opening message to match the overall change introduced in this PR.

pkoutsovasilis:
Thanks for the updated PR and the detailed telemetry @sealemar, I appreciate the effort. The two main concerns are addressed and the change is now much leaner - no health state masking, no interface widening.

However I still don't think we should add the retry, and the reason is rooted in how the operator is designed.

#### The reconcile loop is already the retry

When a health check fails, `LastHealth()` changes to `Unknown`. `healthChangeListener` (`observer/watch.go:34-49`) fires an immediate reconcile event on every health state change. On top of that, `DefaultRequeue` is 10 seconds (`pkg/controller/common/reconciler/results.go:17`). So a failed observation already triggers: an immediate reconcile, followed by another one 10 seconds later - each picking a random ready pod via `URL()`. Eventual convergence through the reconcile loop is the operator's design contract. Adding a second call within the same `observe()` tick is doing what the loop does anyway, just without the event boundary.

#### The retry doesn't fix staleness

Both calls draw from the same informer cache. If the cache shows a terminating pod as Ready, both attempts can pick the same bad endpoint. The retry doesn't fix staleness - it spins the same roulette wheel twice.

#### The Envoy config proves this is a mesh configuration - and Istio already has the retry

The config dump you shared actually reinforces this. Istio already ships a default retry policy:

```
num_retries: 2
retry_on: connect-failure, refused-stream, unavailable (503)
retry_host_predicate: previous_hosts
host_selection_retry_max_attempts: 5
```

`previous_hosts` is explicitly designed to route retries to a different upstream instance. It fails here only because `ORIGINAL_DST` gives each pod-specific DNS name its own single-endpoint cluster - so there is no alternative host for Envoy to select within that cluster. The operator's retry hits the exact same wall for the same reason when it lands on the same pod.

The root cause is the `ORIGINAL_DST` cluster type, not a missing retry. That is a mesh-side configuration and the separate documentation PR, about `excludeOutboundPorts: '9200'`, you mentioned feels the sole right resolution - explaining how to configure the mesh correctly when running ECK, not having the operator issue a subsequent call to work around it.

---

I'd close this PR and let the mesh-side documentation be the resolution.

sealemar:
Thanks @pkoutsovasilis. The concerns about the grace period and interface change are fully addressed.

On the retry: you described it as "spinning the same roulette wheel twice." That's accurate — and it's why it helps. One random draw per tick has P=1/N of hitting a stale pod. Two draws have P=1/N². For a 3-node cluster, that's 33% → 11%. For a 10-node cluster, 10% → 1%.

The reconcile loop does converge, but each tick is an independent 1/N draw. Consecutive misses are not unlikely — and each costs ~10s of upgrade pause. The retry doesn't fix staleness; it reduces the per-tick miss rate by a factor of N, within milliseconds, at 16 lines of code with no interface change.

Production data from the last 10 days: clusters with 97-99% health-check success rates still show scattered slow-fail events, each causing a 10s upgrade pause. The retry would collapse most of those to milliseconds.

pkoutsovasilis:
Thanks @sealemar. 

I am not so sure about the "1/N² in milliseconds vs 1/N per 10s tick" framing. Two draws inside the same observe() tick aren't independent: both URL() calls read the same informer cache snapshot. If staleness is to be considered here, the second draw queries the same view of the world - it can re-pick the same bad endpoint, or pick a different one that's stale for the same cache-lag reason. AFAICT the factor-of-N reduction needs independent draws against ground truth; in-tick we have two draws against one snapshot.

More importantly, setting probabilities aside, with this PR we are adding operator code that only earns its keep when the mesh is misconfigured (as shown by your own telemetry the slow-fail population is mesh-induced), and paying that cost on every cluster that hits such a failure, mesh or no mesh.

I'd still close this PR and focus on the mesh-side documentation one.

sealemar:
Closing-the-loop with production data: across 14 upgrades between 2026-05-22 and 2026-05-26 with the `excludeOutboundPorts: '9200'` annotation on the ECK operator chart, telemetry shows zero slow-503s, zero `remote_connection_failure` events, and zero observer-driven `Unknown` health transitions. In our deployment, **the mesh-induced failure mode is fully resolved by the annotation alone**; we have no production evidence of the operator-side stale-cache scenario to argue for independently.

Closing this PR and focusing on [elastic/docs-content#6577](https://github.com/elastic/docs-content/pull/6577) as the canonical resolution.

Thanks for the thorough review on both rounds.

PR #9437, opened 2026-05-12
[#9435] Update `WithTemplateHash` functions to hash based on the `Spec` and not the whole object

### LINKED ISSUE #9435: Incorrect value of common.k8s.elastic.co/template-hash label causes unnecessary reconciliation
## Bug Description

Elasticsearch sets the \`common.k8s.elastic.co/template-hash\` label based on \`StatefulSet.Spec\`:
https://github.com/elastic/cloud-on-k8s/blob/f734f12737106633a2b1f06ae00cd1b99ce4a70e/pkg/controller/elasticsearch/nodespec/statefulset.go#L143

However, stateless Deployment resources set the label's value based on the **entire Deployment** object:
https://github.com/elastic/cloud-on-k8s/blob/f734f12737106633a2b1f06ae00cd1b99ce4a70e/pkg/controller/common/deployment/reconcile.go#L93

This means the hash can differ from the stored value:
https://github.com/elastic/cloud-on-k8s/blob/f734f12737106633a2b1f06ae00cd1b99ce4a70e/pkg/controller/common/deployment/reconcile.go#L72-L77

## Impact

Because the expected Deployment is always constructed fresh from CR state (never from the existing resource), the hash stabilises after one update — there is no perpetual churn. However, hashing the full object rather than just the \`Spec\` makes the label an unreliable spec-only change indicator: it also captures \`ObjectMeta\` fields that are not strictly spec changes.

In practice the impact is narrow. The `NeedsUpdate` checks that follow the hash comparison use `!maps.IsSubset` to cover those same fields, so any hash mismatch is always redundant with those checks and never an independent trigger for an unintended update. The fix aligns the implementation with the documented intent of the label and removes the inconsistency with the Elasticsearch `StatefulSet` path.

## Expected Behavior

The `template-hash` label for `Deployment`-based resources should be computed from a stable subset of the `Deployment` spec (e.g. `Deployment.Spec`), consistent with how Elasticsearch computes it from `StatefulSet.Spec`.

dtuck9:
Fixes: #9435

---                                                                                                                                                                                                                                                                                                                       
  ## Summary                                                                                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                            
  Fixes https://github.com/elastic/cloud-on-k8s/issues/9435.                                                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                                                            
  The `WithTemplateHash` functions for `Deployment`, `DaemonSet`, and `StatefulSet` were computing the `common.k8s.elastic.co/template-hash` label by hashing the entire object, including its `Labels` field. Because the label being written is itself part of the object being hashed, this creates a circular dependency: the hash  
  computed on reconciliation always differs from the hash already stored on the resource, causing an unnecessary update on every reconciliation loop.
                                                                                                                                                                                                                                                                                                                            
  The Elasticsearch-specific `StatefulSet` path has always hashed only `StatefulSet.Spec`, which is the correct approach. This PR aligns the common reconcilers with that pattern by changing all three `WithTemplateHash` functions to hash only the Spec field.                                                                 
   
  ## Changes                                                                                                                                                                                                                                                                                                                   
                  
  - `pkg/controller/common/deployment/reconcile.go`: hash `Deployment.Spec` instead of the full `Deployment` object.                                                                                                                                                                                                              
  - `pkg/controller/common/daemonset/reconcile.go`: hash `DaemonSet.Spec` instead of the full `DaemonSet` object.
  - `pkg/controller/common/statefulset/reconcile.go`: hash `StatefulSet.Spec` instead of the full `StatefulSet` object.                                                                                                                                                                                                           
  - Tests updated/added for all three to assert that calling `WithTemplateHash` on an already-hashed object produces the same hash (idempotency), which was the failing case before this fix.
                                                                                                                                                                                                                                                                                                                            
  ## Testing         
                                                                                                                                                                                                                                                                                                                            
  - Unit tests for `WithTemplateHash` in all three packages now include an idempotency assertion: applying `WithTemplateHash` to an object that already carries the label must produce the same hash value.                                                                                                                     
  - The existing assertions (hash is set, original object is unmodified, hash changes when spec changes) are preserved.

### DIFF
--- a/pkg/controller/common/daemonset/reconcile_test.go
+++ b/pkg/controller/common/daemonset/reconcile_test.go
@@ -1,3 +1,7 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License 2.0;
+// you may not use this file except in compliance with the Elastic License 2.0.
+
 package daemonset
 
 import (
--- a/pkg/controller/common/daemonset/reconcile_test.go
+++ b/pkg/controller/common/daemonset/reconcile_test.go
@@ -1,0 +1,45 @@
+package daemonset
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	appsv1 "k8s.io/api/apps/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/elastic/cloud-on-k8s/v3/pkg/controller/common/hash"
+)
+
+func TestWithTemplateHash(t *testing.T) {
+	d := appsv1.DaemonSet{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "daemon",
+			Namespace: "ns",
+		},
+		Spec: appsv1.DaemonSetSpec{
+			UpdateStrategy: appsv1.DaemonSetUpdateStrategy{},
+		},
+	}
+
+	withHash := WithTemplateHash(d)
+	// the label should be set
+	require.NotEmpty(t, withHash.Labels[hash.TemplateHashLabelName])
+	// original object should be kept unmodified
+	require.Empty(t, d.Labels)
+
+	// label should be consistent
+	withSameHash := WithTemplateHash(d)
+	require.Equal(t, withHash.Labels[hash.TemplateHashLabelName], withSameHash.Labels[hash.TemplateHashLabelName])
+
+	// label should be the same if no spec changed
+	withSameHash = WithTemplateHash(withSameHash)
+	require.Equal(t, withHash.Labels[hash.TemplateHashLabelName], withSameHash.Labels[hash.TemplateHashLabelName])
+
+	// label should be different if the spec changed
+	d.Spec.UpdateStrategy = appsv1.DaemonSetUpdateStrategy{
+		Type: appsv1.RollingUpdateDaemonSetStrategyType,
+	}
+	withDifferentHash := WithTemplateHash(d)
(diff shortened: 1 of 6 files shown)


prodsecmachine:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/71981a56-d611-48ae-abb7-3d811e43cdc7?refs=4e808080b557b5dbd08d5c2d3e174d6850f2bb1a&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/elastic-cloud-on-k8s/pr-checks/71981a56-d611-48ae-abb7-3d811e43cdc7/license?refs=4e808080b557b5dbd08d5c2d3e174d6850f2bb1a&source=prChecksComment) |

pkg/controller/common/daemonset/reconcile_test.go
```diff
@@ -0,0 +1,45 @@
+package daemonset
```
kvalliyurnatt, line 5: this is missing the Elasticsearch copyright

dtuck9: Yup working on it.

kvalliyurnatt (approved):

dtuck9:
buildkite test this -f p=gke,t=.*,E2E_TAGS=kb

[[build](https://buildkite.com/elastic/cloud-on-k8s-operator/builds/14755)]

